Issue #2564, opened 2026-07-08
Research API: one evidence spine — facts and diffs as a single stream


richlander:
## Goal

Collapse the Research library's **two parallel evidence models** into one
subject/anchor-keyed evidence spine:

- the offset-keyed **fact** substrate (`ResearchViews` / `ResearchFactRegistry`,
  `Annotation` with a typed `Descriptor`), and
- the subject-keyed **diff** substrate (`ResearchDiff`,
  `ResearchDiffEvidence` / `ResearchDiffRow`).

Once they share a row model, a diff stops being a separate code path and becomes
`facts(new) − facts(old)` — the same producers at two arities. This is the
integration-seam workstream under the harness north star (#2561).

## Why

Research is correctly the dependency-inverting **join** across
Analysis / Decompiler / Instructions / Metadata: the siblings don't depend on each
other; Research is where they meet. That decision is right. But the "unified view"
is today two unrelated evidence worlds — different keys, different row types,
different composition, and no shared descriptor:

| | Fact side | Diff side |
| --- | --- | --- |
| Key | IL offset | `ResearchSubjectKey` |
| Row | `Annotation` | `ResearchDiffEvidence` **and** `ResearchDiffRow` |
| Id | typed `Descriptor` (Id/Category/Conditionality) | stringly `api.*` / `il.operation.*` / `analysis.signal.*` |
| Display | neutral; `ResearchViews` renders | baked into the row (`*DisplayRow`) |

The framing that unlocks it (from discussion): **a fact stream is one column of
state (`Present`); a diff stream is the same rows with the old/new column reduced
to a polarity tag (`Added`/`Removed`/`Changed`).** Same rows, one extra field.

## Core model

```text
FactRow = (subject, anchor, descriptor, detail, provenance, polarity)

source stream : polarity = Present                     (one column)
diff  stream : polarity ∈ {Added, Removed, Changed}    (old/new → polarity)

diff-facts(old, new) = facts(new) − facts(old)         keyed by stable fact identity
```

Every producer emits `(subject, anchor, descriptor, detail)` over **one**
assembly; a single generic differ then yields "unsafe added / allocation added /
throw introduced / reflection removed" for **every** producer at zero per-producer
cost. A new fact producer ships its diff dimension automatically.

Today this is half-done and inconsistent:

- **unsafe** is diffed by offset identity
  (`BodySignalDiff.CompareUnsafe` → added/removed) — a real fact diff;
- **allocations** are diffed only by count
[...]

PR #2627, opened 2026-07-10
Add extension member Finding producers

richlander:
Adds typed Metadata Finding producers for extension methods and C# 14 extension properties. Findings use canonical structured identity, preserve the legacy scanner record contract, fail the whole census when shape is unavailable, and keep rendered signatures out of payloads.

Extension-property identity includes receiver and index parameters, deduplicates getter/setter accessors, distinguishes same-named properties across receivers, filters hidden declaration metadata, and excludes ordinary property accessors and standalone `set_*` methods.

Also addresses late review findings from #2621: rank-zero array formatting cannot throw `ArgumentOutOfRangeException`, generic method fallback parameters use `!!`, and the intentional classified-method equality boundary is documented.

Validation: full solution build; 450 Metadata tests; 25 targeted product scanner tests. Final adversarial reviews on this exact rebased head are CLEAN from Claude Opus 4.8 and Gemini 3.1 Pro.

PR #2623, opened 2026-07-10
Enforce Research change payload ownership

richlander:
## Conclusion

**PASS** — `ResearchChange` now rejects native payloads that disagree with its declared mechanism, closing the interim tagged-slot hazard without turning the migration projection into a permanent union.

Follow-up to #2620 and its architecture review. Advances #2614 and #2564.

## Change

- Validate every API, body-signal, IL, and C# native payload slot against `ResearchChangeMechanism`.
- Reject mixed-family construction such as an API-tagged change carrying an IL row.
- Preserve common-field-only status, failure, and aggregate rows, including ReturnToSender status and IL failure-string rows.
- Treat default or empty display arrays as no payload; constrain nonempty arrays to their native family.
- Document the invariant as temporary and retiring with the projection fields.

## Validation

```text
dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config

ILInspector.Research.Tests: 77 passed
ReturnToSenderEvidenceTests: 3 passed
Total: 80 passed

markdownlint: clean
```

The solution build retains three pre-existing nullable warnings in Metadata tests.

## Adversarial review

Claude Opus 4.8 and Gemini 3.1 Pro are reviewing `5f2839d3` in separate clean worktrees. Results will be reconciled in a PR comment.

PR #2622, opened 2026-07-10
Consume Metadata Findings in library inspection

richlander:
## Summary

- retain typed `FindingInspection<T>` results for Metadata-owned library observations across acquisition, section discovery, views, and package rollups
- replace duplicate CLI observation models and per-integration collection fanout with projections from producer payloads
- consume classified-method Findings for semantic counts and eligibility while keeping rendered signatures as presentation-only rows
- preserve the existing raw JSON field schema through JSON-only compatibility projections; custom-attribute array order now follows the producer's deterministic Finding order

Research remains the aggregation boundary for views that genuinely join multiple producers; DotnetInspector consumes single-domain Metadata Findings directly.

## Validation

- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config`
- `dotnet run --project src/dotnet-inspect.Tests -c Release` (1,706 passed, 10 tool-dependent skips)
- `npx markdownlint-cli docs/architecture.md`
- base/head `library --json` A/B on the same assembly: identical schema and values; only custom-attribute array ordering changed to deterministic producer order

Related to #2564.

PR #2638, opened 2026-07-11
Add Analysis allocation Findings

richlander:
Allocation occurrences now have one Analysis-owned Finding census for single-version facts and two-version changes.

- add total `AnalysisFindings.InspectAllocations` and conservative `CompareAllocations` correspondence over source, kind, and allocated type
- retain IL offsets and tokens as typed version-local provenance while classifying semantic facet changes
- project Research annotations from the census and retain the typed allocation comparison on `ResearchChange`
- replace the count-only allocation diff lane while preserving count and hot/cold DotnetInspector output
- surface equal-count remove/add replacements that were previously invisible

Validation: solution build; 416 Analysis tests; 81 Research tests; 1,730 DotnetInspector tests (10 expected ilasm/ildasm skips); markdownlint.

Advances #2564.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -135,7 +135,9 @@ separate axes; see [Finding Coordinates](design/finding-coordinates.md).
 - **R1** is the lower, metadata/IL representation. `ILInspector.Analysis` reads
   assemblies with `System.Reflection.Metadata`, builds whole-assembly indexes,
   and produces method signals, direct-call evidence, allocation occurrences, and
-  leverage data without depending on the decompiler IR.
+  leverage data without depending on the decompiler IR. `AnalysisFindings`
+  projects allocation occurrences into one typed, IL-ordered census used by
+  both single-version and two-version consumers.
 - **R2** is the higher decompiler representation. `ILInspector.Decompiler`
   imports one method into IR, raises/lowers it for C# printing, projects raw or
   annotated IL, and supplies IR anchors for source-line placement.
@@ -148,6 +150,10 @@ separate axes; see [Finding Coordinates](design/finding-coordinates.md).
   computed as a view rather than stored as duplicate state. API comparisons
   retain Metadata's `ApiFindingComparison` (type/member `FindingComparison<T>`
   envelopes plus the classified `ApiDiff`) alongside the Research projection.
+  Allocation changes retain Analysis's
+  `FindingComparison<AllocationOccurrence>` on their Research projection;
+  count, hot/cold, and ranking fields are derived from that comparison rather
+  than from a separate count-only diff.
 
 `ResearchFactRegistry` is the dogfooded analyzer registry for the overlay.
 Producers implement `IResearchFactProducer` with a stable name, produced fact
@@ -158,7 +164,7 @@ includes:
 
 | Producer | Source | Facts |
 | -------- | ------ | ----- |
-| `AllocationOccurrenceFactProducer` | `ILInspector.Analysis` `LibraryBodyIndex` allocation occurrences | `alloc.box`, `alloc.array`, `alloc.new`, `alloc.closure`, `alloc.statemachine`, `alloc.delegate`, `alloc.enumerator` |
+| `AllocationOccurrenceFactProducer` | `ILInspector.Analysis` allocation Finding census | `alloc.box`, `alloc.array`, `alloc.new`, `alloc.closure`, `alloc.statemachine`, `alloc.delegate`, `alloc.enumerator` |
 | `DecompilerHiddenFactProducer` | existing decompiler annotation classifier | `unsafe.*`, `lifetime.*` |
 
 The important boundary is that Analysis remains SRM-only, NativeAOT-friendly,
--- a/docs/design/finding-coordinates.md
+++ b/docs/design/finding-coordinates.md
@@ -17,13 +17,16 @@ different questions and do not share one generic string slot.
 `Ordinal` does not control alignment or move classification. It lets an ordered
 producer retain the source-stream location after observations have been paired,
 materialized, or projected. Ordered IL, C#, and text observations populate it;
-Metadata identity sets leave it null.
+allocation observations also populate it after ordering by IL offset. Metadata
+identity sets leave it null.
 
 This distinction is intentional:
 
 - an API member does not gain semantic position because its inventory was
   sorted for deterministic output;
 - an IL operation's ordinal is an operation-array index, not its IL offset;
+- an allocation's ordinal is its position in the allocation census, while its
+  IL offset remains typed payload provenance;
 - a text line's ordinal is a logical line index, not a cross-document identity;
 - changing retained ordinals does not turn an order-preserving match into a
   move.
--- a/docs/design/finding-producers.md
+++ b/docs/design/finding-producers.md
@@ -161,6 +161,7 @@ Before approving a producer, answer:
 - Does comparison reuse the shared matcher and fold?
 - Which bespoke producer, adapter, or comparison path will this replace?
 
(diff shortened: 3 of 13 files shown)


richlander:
## Review — the Analysis proof the producers doc asked for, delivered as specified; one fail-open arm to close

The finding-producers doc set a concrete bar: "the first end-to-end Analysis proof should use one allocation census for both single-version audit and old/new comparison, replacing the count-only Research allocation path rather than adding another adapter." This PR meets it literally — `AllocationOccurrenceFactProducer` (single-version) and `AddAllocationRow` (two-version) both consume `AnalysisFindings.InspectAllocations`, the count-only `"allocations"` lane in `AddCountRows` is deleted rather than shadowed, and the doc flips from "should" to "uses." That's replacement, not accretion, and it's the pattern I'd want the *next* Analysis producer to copy.

The doctrine mechanics check out:

- **Identity vs provenance is drawn exactly right.** Correspondence identity is `Source:Kind:AllocatedType`; MVID, token, IL offset, and operand token are version-local provenance and never establish cross-version identity. `CompareAllocations_IgnoresVersionLocalProvenance` pins it. The `AllocationOccurrence` doc-comment update ("the offset is version-local provenance, not cross-version correspondence identity") says the rule out loud where the next reader will see it.
- **The ordinal story is the ordered-producer case, applied correctly.** Allocations are a source-stream ordering (IL order is semantic), so the census orders by IL offset and assigns ordinals post-ordering — which is what lets the shared Ordered matcher align duplicate identity keys (two `new object()` in one method) conservatively by stream position. The finding-coordinates doc addition ("an allocation's ordinal is its position in the allocation census, while its IL offset remains typed payload provenance") keeps the Ordinal-doesn't-classify-moves invariant intact.
- **Envelope forging goes through the only legal gate.** `FindingComparison.Compare` + `TransformPairs(ClassifyFacetChanges)` — no bespoke fold, and the Present→Changed facet reclassification is exactly what `TransformPairs` exists for.
- **`SameSemanticFacets` is the best trick in the PR.** Normalizing away the three provenance fields via `with`-copy and comparing record equality means *every other field is a semantic facet by construction* — a new field added to `AllocationOccurrence` automatically participates in change detection instead of being silently ignored. Fail-visible by default. The corollary is a small drift hazard: `DescribeFacetChanges` hand-enumerates fourteen facets, so that same new field would produce a `Changed` pair with an *empty* detail string until someone remembers the list. Cheap insurance: when the changes list comes back empty, emit `"other facets changed"` (or assert) so detection and description can't silently diverge.
- **The `ResearchChange` typed payload follows the #2623 disposition exactly** — `AllocationComparison` validated to `BodySignals` via `ValidatePayloadMechanism`, with the cross-mechanism rejection test. Consistent with the interim-hazard remedy we settled there, no union refactor smuggled in.

Two things to address, one of which I'd fix before merge:

**1. `AddAllocationRow` swallows `Failed` — the new lane can fail silently where the old one couldn't.**

```csharp
var complete = comparison switch
{
    FindingComparison<AllocationOccurrence>.Complete value => value,
    FindingComparison<AllocationOccurrence>.Failed => null,
};
if (complete is null || complete.IsExact)
    return;
```

The count-only lane this replaces had no failure mode — it always reported counts. The new lane routes through the matcher with `acceptanceThreshold: 100`, and the `Failed` arm maps to *no row at all*. If `Failed` is reachable for two `Complete` inputs (threshold rejection on a method with a large allocation census), then precisely the methods with the most allocation churn — the ones this lane exists to catch — vanish from the diff with no signal. That's the #2622 lesson in a new spot: an operation failure rendering as "nothing to report." The fix is small either way: on `Failed`, fall back to the count-only row (counts are still computable from the raw occurrences you already hold) or emit a failure-marked row; and if `Failed` is genuinely unreachable here, replace the null-swallow with a throw so the assumption is enforced rather than assumed.

**2. The identity fallback chain ends in display text, then a wildcard.**

```csharp
string allocatedType = occurrence.AllocatedType?.ToQualifiedDisplayString()
    ?? occurrence.RuntimeAllocationType
    ?? occurrence.Detail
    ?? "?";
```

By the standard we applied to the Metadata producers (#2627 fails closed when `CanonicalExtendedType` is null), this is a softening: identity degrades structured → semi-structured → rendered text → `"?"`, and two structurally-unknown allocations of the same source+kind become interchangeable under the wildcard. I think the softening is *defensible* here — Analysis occurrences are heuristic evidence with no canonical anchor to demand, and failing the whole census over one odd occurrence would be the worse trade — but that reasoning should live in a comment on `GetAllocationIdentityKey`. Right now it reads as the exact pattern the Metadata producers were told not to do, with no marker that the difference is deliberate.

Minor, no action needed: the hot/cold flip branch gates on `IsHotAllocation` (heap + in-loop + non-throw-path) while the equal-count `changed` branch computes `inLoop` from raw `InLoop` — a small definitional asymmetry between adjacent branches; and the equal-count remove/add replacement surfacing (`Object` → `byte[]` at the same count, previously invisible) is a genuine capability gain and has the right test.

Endorse with item 1 fixed. The census/consumer structure is exactly what the doc demanded, and the `with`-copy facet trick deserves to become the house pattern for provenance-vs-semantics splits.

---

richlander:
Adversarial review reconciliation for `e59b6c46`:

- **Claude Opus 4.8: CLEAN.** Verified the exact provenance mask, source/kind/type correspondence, all semantic facets, equal-count remove/add behavior, non-allocation lane isolation, count/hot/cold compatibility, native payload ownership, fact projection, and self-diff determinism. Ran focused Analysis, Research, and DiffCommand tests plus a clean solution build.
- **Gemini 3.1 Pro: CLEAN.** Independently verified clean-worktree identity/duplicate semantics, record-facet comparison, count-lane isolation, empty census handling, non-heap occurrence behavior, and all 81 Research tests.

The reviewers agreed: no high-confidence correctness, security, or compatibility findings. No follow-up or resolution commit applies because both reviews were clean on the current PR head.

richlander:
Ready to merge.

All merge-blocking local validation and CI are green, and Claude Opus 4.8 plus Gemini 3.1 Pro both returned CLEAN adversarial reviews on the current head.

richlander:
Addressed the review feedback in `8281ed93`:

- replaced the unreachable `Failed => null`/early-return path with a fail-visible invariant exception; a future non-total allocation inspection can no longer silently erase a diff row
- documented why unresolved Analysis allocation identity deliberately falls back through runtime/detail evidence and then ordered unknown candidates
- added `other facets changed` as a drift-safe description when record equality detects a future semantic facet not yet named by the detail formatter

The existing focused Analysis, Research, hot/cold, and self-diff tests remain green. No change was made for the noted hotness asymmetry because it was explicitly non-actionable and reflects the existing heap/non-throw hotness definition.

richlander:
Fixed-head adversarial rereview reconciliation for `8281ed93`:

- **Claude Opus 4.8: CLEAN.** Confirmed the `Failed` case is unreachable with two total `Complete` censuses, the new exception is correctly fail-visible, unknown candidates are conservatively stream-aligned, future unnamed facets retain a non-empty detail, and allocation/count/hotness/native-payload behavior remains correct. Focused Analysis and Research suites passed.
- **Gemini 3.1 Pro: CLEAN.** Independently confirmed the invariant, fallback matching, equal-count replacement, payload ownership, empty-census semantics, future-facet fallback, and body-signal hotness equivalence.

Both reviewers agree the maintainer feedback is resolved with no outstanding correctness, security, or compatibility findings.

richlander:
Ready to merge.

The maintainer feedback is resolved in `8281ed93`; updated CI is green, and Claude Opus 4.8 plus Gemini 3.1 Pro both returned CLEAN fixed-head rereviews with no outstanding findings.

PR #2640, opened 2026-07-11
Replace project grounding with skills

richlander:
## Summary

- Replace `project -S Grounding` with `project -S Skills` for direct dependency package `skills/**/SKILL.md` discovery.
- Change `project --readme <package-id>` to resolve root `README.md`, then root `PROJECT.md`.
- Update project command docs, tips, and command execution tests for the new behavior.

## Validation

- `dotnet run --project src/dotnet-inspect.Tests -c Release -- -method "*Project_*"`
- `npx markdownlint-cli README.md docs/architecture.md docs/design/rendering-model.md docs/workflows/core/package-inspection.md`
- `dotnet build src/dotnet-inspect/dotnet-inspect.csproj -c Release --configfile nuget.config -p:PublishAot=false`

## Review

Behavior change; adversarial review requested before marking ready.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ the .NET 11 daily feed and nuget.org: most projects target the nightly
 | Source | Examples | Notes |
 | ------ | -------- | ----- |
 | NuGet packages | `package System.Text.Json`, `type --package Markout` | Supports versions, custom sources, `nuget.config`, TFMs, package layout, dependencies, and vulnerabilities. |
-| Restored projects | `type Command --project ./src/App`, `project ./src/App -S Grounding --print` | Uses an existing `project.assets.json` as restored-assets context for API lookup, relationship search, and package grounding; restore/build first if dependencies changed. No restore/build/MSBuild evaluation is run. |
+| Restored projects | `type Command --project ./src/App`, `project ./src/App -S Skills --print` | Uses an existing `project.assets.json` as restored-assets context for API lookup, relationship search, and dependency package skills; restore/build first if dependencies changed. No restore/build/MSBuild evaluation is run. |
 | Platform libraries | `library System.Private.CoreLib`, `library System.Text.Json --version 10.0.0`, `diff --platform System.Runtime@9.0.0..10.0.0` | Resolves installed SDK/runtime assemblies, including runtime-only implementation assemblies with no NuGet package. |
 | Local assets | `library ./bin/MyLib.dll`, `package ./pkg/MyLib.nupkg` | Useful for auditing builds before publishing. |
 
@@ -107,7 +107,7 @@ context for copied DLLs. A future `--deps` source can represent runtime
 | Capability | Commands | Highlights |
 | ---------- | -------- | ---------- |
 | Package inventory | `package` | Metadata, versions, TFMs, file layout, dependency tree, metadata audit, vulnerability data, custom feeds, NuGet config support. |
-| Project grounding | `project` | Direct dependency `Grounding` rows and printable best package docs from restored projects. |
+| Project skills | `project` | Direct dependency `Skills` rows from package `skills/**/SKILL.md` files, plus version-resolved package README/PROJECT docs from restored projects. |
 | Library audit | `library` | Assembly identity, public key token, trim/AOT metadata, unsafe/interoperability signals, OpenTelemetry support, symbols/PDBs, SourceLink and determinism audit, references, resources, async method classification. |
 | API discovery | `type`, `member`, `find` | Type search, member tables, docs, overload selection, generics, obsolete-member markers, direct calls and callers, source/decompiled/IL drill-in. Add `--project` to resolve type/member queries in the project's restored dependency context. |
 | API compatibility | `diff` | Version ranges, package or platform diffs, breaking/additive/potentially-breaking classification, type and member filters. |
@@ -122,7 +122,7 @@ context for copied DLLs. A future `--deps` source can represent runtime
 | Command | Purpose |
 | ------- | ------- |
 | `package X` | Inspect NuGet metadata, versions, dependencies, TFMs, layout, and vulnerabilities. |
-| `project [path]` | Inspect restored direct package references for grounding indexes and package docs. |
+| `project [path]` | Inspect restored direct package references for skill files and package docs. |
 | `library X` | Inspect assembly metadata, symbols, SourceLink, references, resources, and async methods. |
 | `type X` | Discover types or render a single type shape. |
 | `member X` | Inspect members, docs, overloads, decompiled/lowered C#, SourceLink-backed original source, and IL. |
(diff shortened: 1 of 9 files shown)


richlander:

- Claude Opus 4.8: CLEAN. Verified mode selection, discovery schema, shape projections, print behavior, JSON contexts, README/PROJECT fallback, unchanged package Grounding behavior, and no version metadata changes. Non-blocking notes only: root-only README/PROJECT narrowing is intentional; exact-case `SKILL.md` discovery matches the stated contract.
- Gemini 3.1 Pro: CLEAN. Verified recursive `skills/**/SKILL.md` discovery, shape projections, schema discovery, JSON serialization, print output, README/PROJECT fallback, intact project dependency versions, and unchanged PackageCommand behavior.

No actionable findings from either review, so no follow-up commit applies. CI still has one test job in progress; I will post Ready to merge after it completes successfully.

richlander:
## Review — right direction; two behavior losses to confirm before marking ready

The core move is doctrinally sound and worth stating plainly: **`Grounding` manufactured a "best," `Skills` enumerates what exists.** The old lens ran a tool-side ranking ladder (`AGENTS.md` > `README.md` > `PACKAGE.md` > declared readme) and presented the winner as the row — curation the tool performed on the consumer's behalf. The new lens is a census of a purpose-built artifact (`skills/**/SKILL.md` with name/description frontmatter), where the *convention* defines membership and the tool just reports it. That's the same shift the Finding work keeps making — stop synthesizing a "best," report the census — and it aligns the lens with the agent-skills ecosystem convention it's actually for. The `--readme` change has the same honest-labeling virtue: the old test literally asserted that `--readme` returned `AGENTS.md` content in preference to the README; a flag named `--readme` now returns the readme.

Mechanics that check out: `Directory.EnumerateFiles(..., "SKILL.md", AllDirectories).Order(StringComparer.Ordinal)` gives deterministic row order; the project section name becomes a local `"Skills"` const instead of borrowing `PackageSections.PackageReadme` (the old lens literally reused the package readme section identifier — good riddance); package-side `-S Grounding` is untouched and `rendering-model.md` now correctly distinguishes the two; the test conversion is thorough across jsonl/paths/value/print/print-all/bare/row/count/columns and adds `PROJECT.md` fallback coverage.

Three things to settle, ranked:

**1. `--readme` silently dropped the nuspec-declared readme — confirm this is intended.** Old resolution consulted `NuspecParser.FindAndParse(...).ReadmeFile` and could resolve a readme anywhere in the package layout. New resolution is *top-level `README.md`, then top-level `PROJECT.md`*, filename-convention only. A package that declares `<readme>docs/README.md</readme>` — a real and common layout — now resolves to nothing and is skipped. That's a curious inversion: the nuspec declaration is the package author's *own* statement of "this is my readme," which makes it the most producer-owned signal available, and the new code replaces it with a filename convention. Dropping `AGENTS.md`/`PACKAGE.md` from `--readme` is right (they're not readmes); dropping the declared readme loses real ones. Suggest keeping declared-readme as the *final* fallback (README.md > PROJECT.md > declared), unless there's a reason it was cut that the PR body doesn't state. Related: `PROJECT.md` is a convention with no definition anywhere in the repo — one sentence somewhere (architecture.md already mentions it in passing) saying what's expected to produce it would keep the next reader from wondering if it's a typo for `PACKAGE.md`.

**2. Absence became invisible — fine as redesigned, but say it out loud in the row-semantics docs.** `Grounding` emitted one row per direct dependency, *including an empty row* for a dependency with nothing — absence was visible, and `--count` counted dependencies. `Skills` yields zero rows for a dependency without skills, and `--count` now counts files (the updated count test: 2 skills in one dep + a dep with none = 2, where the old lens would have said 2 deps). Rows-are-artifacts is the coherent model for this lens, but it means "which of my dependencies have *no* skills" is no longer answerable from it, and any script consuming `--count` changes meaning. The rendering-model line ("renders every direct dependency package `skills/**/SKILL.md` file") implies it; the `--count` semantics change deserves an explicit mention since it's the one output a script is most likely parsing.

**3. One stale prose line.** `docs/workflows/core/package-inspection.md` line ~220 still reads "`project` grounding and API commands with `--project` both use an existing `project.assets.json`…" — the command block above it was converted but this sentence wasn't. (`release-notes.md`'s historical `--agents-index` mention is fine to leave; it describes a past release.)

Trivial, no action: `skills`/`SKILL.md` matching is exact-case on Linux — correct for the convention, just inherently case-sensitive where the old readme resolution was case-insensitive; and the mode error message still advertises `--agents-index`, which is retained-but-de-advertised — consistent with keeping it as legacy.

Endorse the design; item 1 is the one I'd want answered (and ideally fixed with the declared-readme fallback) before this is marked ready.

---

richlander:
Ready to merge

Final status:
- CI is green / merge state CLEAN.
- Two adversarial reviews are clean (Claude Opus 4.8 and Gemini 3.1 Pro), with reconciliation already posted.
- Additional project.assets.json check: a restored local package with `skills/query/SKILL.md` lists that file in `project.assets.json`, but this PR intentionally does not depend on that listing; `project -S Skills` resolves from the package root path and printed the skill row successfully.

richlander:
Update after Ready to merge:

Pushed a follow-up commit (`7836ff11`) per review direction from @richlander: `project -S Skills` now uses a DotnetInspector.Services project.assets.json file-entry API (`ParsePackageFileEntries` / `HasPackageFileEntries`) for `skills/**/SKILL.md` instead of enumerating package directories in the CLI.

The previous Ready to merge comment is superseded. I am waiting for fresh CI and fresh clean adversarial reviews on the new head before posting Ready to merge again.

Additional local validation on the new commit:
- `dotnet run --project src/DotnetInspector.Services.Tests -c Release -- -class DotnetInspector.Services.Tests.ProjectAssetsParserTests`
- `dotnet run --project src/dotnet-inspect.Tests -c Release -- -method "*Project_*"`
- `dotnet build src/dotnet-inspect/dotnet-inspect.csproj -c Release --configfile nuget.config -p:PublishAot=false`

richlander:
Follow-up after R2 review:

Pushed `19cc5924` to address Gemini R2 findings and Opus non-blocking notes:

- Skip non-string `libraries[*].files` entries before `GetString()` so malformed file arrays do not abort parsing.
- Match both `skills/SKILL.md` and `skills/**/SKILL.md`, preserving the previous recursive discovery behavior for a top-level file under `skills/`.
- Suppress the second project-assets TFM log in project Skills mode; `ParsePackageReferences` already logs the selected TFM.

Local validation on the new head:
- `dotnet run --project src/DotnetInspector.Services.Tests -c Release -- -class DotnetInspector.Services.Tests.ProjectAssetsParserTests`
- `dotnet run --project src/dotnet-inspect.Tests -c Release -- -method "*Project_*"`
- `dotnet build src/dotnet-inspect/dotnet-inspect.csproj -c Release --configfile nuget.config -p:PublishAot=false`

Waiting for fresh CI and fresh clean adversarial reviews before Ready to merge.

richlander:
Final adversarial review reconciliation for head `19cc5924`:

- Claude Opus 4.8 R3: CLEAN. Verified the malformed `files` entry guard, `skills/SKILL.md` top-level match, duplicate verbose TFM log removal, direct-dependency semantics, TFM/path resolution, projections/print/count/JSON behavior, README/PROJECT fallback, and package Grounding isolation. Non-blocking notes only: `HasPackageFileEntries` intentionally reports assets-entry presence, while CLI rows require file existence; README root enumeration cost is negligible.
- Gemini 3.1 Pro R3: CLEAN. Verified clean worktree, local build/tests, R2 fixes, direct dependency traversal, path resolution, JSON contexts, README fallback, and unchanged package/version behavior. Non-blocking note only: `FileSystemName.MatchesSimpleExpression` is sufficient for this use case with the explicit `skills/SKILL.md` pattern plus nested `skills/**/SKILL.md`.

No outstanding review findings remain. Follow-up commits that resolved R2 feedback: `19cc5924`.

richlander:
Ready to merge

Issue #2639, opened 2026-07-11
diff silently drops sections from multi-section selections


richlander:
## Problem

The `diff` command registers `Changes`, `Analysis Diff`, and now
`Implementation Diff` as selectable sections, but its execution path treats the
body-level sections as alternate early-return lenses. Requests that resolve to
multiple sections silently render only the first matching lens:

```bash
dotnet-inspect diff --library old/Foo.dll..new/Foo.dll \
  -S "Changes,Implementation Diff"
```

This renders only `Implementation Diff`. `-S @All` resolves all registered
sections but renders only `Analysis Diff`, preserving the pre-existing
Analysis-over-Changes precedence while omitting the other selected sections.

This conflicts with the general section-selection contract in
`docs/design/section-model.md`, where multiple Markdown sections compose and
explicit single-table formats diagnose multi-section requests.

## Expected

Choose and document one coherent command contract:

1. Compose selected diff sections in Markdown/JSON and diagnose multiple
   sections for table/TSV/JSONL, matching the general section model; or
2. Model API, analysis, and implementation as mutually exclusive named lenses
   rather than independently selectable sections, with an explicit diagnostic
   for mixed selection and `@All`.

The command should not silently omit selected sections.

Found during adversarial review of #2543. This behavior predates the
`Implementation Diff` section: `Changes,Analysis Diff` already renders only
`Analysis Diff`.

hiSandog:
Should `diff` reject section sets it cannot compare one-to-one, rather than silently omitting unmatched sections? An explicit diagnostic listing the dropped section names would make this behavior much safer, especially when the command is used in scripts where a partial diff can look complete.

Issue #2637, opened 2026-07-11
Discovery parity test fails whenever dotnet-inspect.Tests sources are modified


richlander:
## Problem

`CommandExecutionTests.CliDiscoverySections_AreSelectable` inspects the currently
built `dotnet-inspect.Tests.dll`, discovers `SourceLink Integrity`, then requires
selecting every discovered section to exit 0. In a feature worktree that
modifies any test source, the assembly PDB checksum reflects the local source
while its SourceLink URL resolves the committed base revision.
`library -S "SourceLink Integrity"` correctly reports `Mismatch` and exits 1, so
the full product suite fails for normal test changes.

This will affect every agent/PR that edits `dotnet-inspect.Tests` source.

## Reproduction

1. Modify `src/dotnet-inspect.Tests/DiffCommandTests.cs`.
2. Rebuild the test project.
3. Run:

```bash
dotnet run --project src/dotnet-inspect.Tests -c Release --no-build --no-restore -- \
  -method "*CliDiscoverySections_AreSelectable*"
```

Observed failure:

```text
library -S 'SourceLink Integrity' failed after being listed by -D
```

Direct inspection reports the modified test source under `Mismatched Files` and
exits 1. The same focused test passes on clean `origin/main` (`7dfef7aa`).

## Expected

Discovery-selection parity should remain valid in a dirty feature worktree. The
test can exclude a status-bearing section whose nonzero exit is expected for the
intentionally modified self-assembly, inspect a stable committed fixture
instead, or assert that selection executes and returns the expected integrity
status rather than requiring exit 0 for this section.

Found during #2543 validation.

Issue #2635, opened 2026-07-11
ImplementationDiff reports metadata-token comparison failures for identical assemblies


richlander:
## Problem

`ImplementationDiff.Compare` reports changed IL rows when an assembly is compared with itself whenever method bodies contain metadata-token operands. This prevents an exact assembly self-diff from being empty and will surface false `changed` evidence in the new CLI consumer.

Example evidence:

```text
IL diff failed: old body metadata token operand at IL_0002 requires a MetadataReader-backed comparison
```

## Reproduction

At `7dfef7aa`, after building the fixture/product graph:

```bash
dotnet artifacts/bin/dotnet-inspect/release/dotnet-inspect.dll diff \
  --library artifacts/bin/DiffFixtures.V1/release/DiffFixtureSample.dll..artifacts/bin/DiffFixtures.V1/release/DiffFixtureSample.dll \
  -S "Implementation Diff" --table -t DiffSample
```

Observed summary: `27 changed members; 0 C# and 27 IL evidence rows.` Every row is a metadata-token comparison failure, despite identical inputs.

## Expected

The IL implementation comparison should use the available metadata readers for token operands and report an exact diff for identical inputs. Comparison failures should remain visible when comparison is genuinely unavailable; they should not occur for this reader-backed assembly path.

Found while implementing #2543.

PR #2641, opened 2026-07-11
Expose ImplementationDiff as an opt-in diff section

### LINKED ISSUE #2543: Expose ImplementationDiff as an opt-in diff CLI section
## Goal

After #2542 lands, add an opt-in `dotnet-inspect diff` section over the product `ImplementationDiff` component so users can inspect merged C# + IL/body implementation evidence without changing default API diff output.

## Proposed shape

- Add a non-default section such as `Implementation Diff` or `Research Diff` to `diff` discovery.
- Use `ImplementationDiff.Compare(...)` rather than invoking `CSharpBodyDiff`, `IlBodyDiff`, or `ResearchDiff` ad hoc from the CLI.
- Preserve existing default `diff` output and existing `Analysis Diff` behavior.
- Support `--library`, package/platform ranges, `--type`, `--member`, `--rows`, `--table`/`--tsv`/`--jsonl` through the normal section/schema machinery.
- Render member-centric rows that can carry C# display rows, IL display rows, or both.

## Constraints

- Keep this opt-in until verbosity, cost, and row shape are settled.
- Do not duplicate producer wording; use the typed display rows/unified lines exposed by `ImplementationDiff`.
- Keep API compatibility rows separate unless a later UX explicitly joins API + implementation evidence.

Related: #2326, #2437, #2542.

richlander:
## Summary

- Add an explicit-only `Implementation Diff` section to `diff` discovery.
- Project Research-owned C# and IL unified evidence as
  `Member | Mechanism | Change | Evidence` rows.
- Reuse package, platform, and local-library acquisition plus existing
  type/member selectors and Markdown/table/TSV/JSONL output.
- Preserve the default API compatibility view and existing `Analysis Diff`
  behavior.

Closes #2543.

## Validation

- `ILInspector.Research.Tests`: 78 passed
- `DiffCommandTests`: 66 passed
- `CommandLineTests`: 174 passed
- Full `dotnet-inspect.Tests`: 1,725 passed, 10 skipped, 1 known
  dirty-worktree failure tracked by #2637
- `markdownlint` passed for all changed Markdown
- Live fixture-pair output checked in Markdown, table, TSV, and JSONL

## Adversarial review

Exact head `ff95a3c2` was reviewed in isolated clean worktrees by Claude Opus
4.8 and Gemini 3.1 Pro.

- Gemini's first pass was clean.
- Opus reproduced metadata-token IL comparison failures and pre-existing
  multi-section first-lens precedence. These are tracked by #2635 and #2639;
  neither is introduced by this CLI projection. Failure rows remain visible
  rather than being hidden.
- Final exact-head re-reviews by both Opus and Gemini were clean with no
  outstanding findings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ context for copied DLLs. A future `--deps` source can represent runtime
 | `type X` | Discover types or render a single type shape. |
 | `member X` | Inspect members, docs, overloads, decompiled/lowered C#, SourceLink-backed original source, and IL. |
 | `find X` | Search for types across packages, frameworks, projects, and local assets. |
-| `diff X` | Compare API surfaces between versions. |
+| `diff X` | Compare API surfaces by default; opt into analysis or C# + IL implementation evidence. |
 | `extensions X` | Find extension methods and C# extension properties for a type. |
 | `implements X` | Find concrete implementors or subclasses. |
 | `depends X` | Walk type, package, or library dependency graphs; emits Mermaid diagrams. |
@@ -301,6 +301,7 @@ dotnet-inspect member JsonSerializer --package System.Text.Json Serialize:1 -S "
 dotnet-inspect member MyType MyMethod:1 --library MyLib.dll -S "Unsafe*"
 dotnet-inspect library System.Text.Json --il-offset 0x06000004+0x15
 dotnet-inspect diff --package System.Text.Json@9.0.0..10.0.0 --breaking
+dotnet-inspect diff --library old/Foo.dll..new/Foo.dll -S "Implementation Diff" -m MyType.HotPath
 dotnet-inspect depends Stream --markdown --mermaid
 dotnet-inspect implements IEquatable --project ./src/App -v:q
 dotnet-inspect extensions string --project ./src/App -v:n
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -36,7 +36,7 @@ The tool is organized around source inspection, API lookup, relationship, and ut
 │  Search for types across packages, platform, and assets      │
 ├─────────────────────────────────────────────────────────────┤
 │                         diff                                 │
-│  Compare API surfaces between package/platform versions      │
+│  Compare API, analysis, or implementation evidence           │
 ├─────────────────────────────────────────────────────────────┤
 │            depends / extensions / implements                 │
 │  Relationship discovery for APIs, packages, and libraries    │
@@ -85,11 +85,14 @@ Extract public API surface using metadata:
 
 ### diff
 
-Compares API surfaces between two package versions:
+Compares two package, platform, or local-library versions:
 
-- Added, removed, and modified types
-- Member-level changes within types
-- Version range syntax: `Package@v1..v2`
+- API compatibility is the default lens: added, removed, and modified types and
+  members.
+- `-S "Analysis Diff"` selects body-signal changes.
+- `-S "Implementation Diff"` selects Research-composed C# + IL evidence grouped
+  by member.
+- Version range syntax is `Package@v1..v2` or `old/Foo.dll..new/Foo.dll`.
 
 ### find
 
--- a/docs/design/implementation-diff.md
+++ b/docs/design/implementation-diff.md
@@ -2,7 +2,7 @@
 
 `ImplementationDiff` is the product-side C# + IL/body diff projection in
 `ILInspector.Research`. It is the reusable implementation-diff component for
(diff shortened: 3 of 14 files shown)


richlander:
Adversarial review reconciliation for exact head `ff95a3c2`:

- **Gemini 3.1 Pro, first pass:** clean. Verified default API diff
  preservation, explicit section discovery, all four output formats,
  type/member filtering, empty/failure rendering, and Markout source-generation
  registration.
- **Claude Opus 4.8, first pass:** reproduced two limitations:
  1. metadata-token IL comparisons produce explicit failure rows;
  2. mixed diff-section requests use first-lens precedence.
- **Disposition:** both findings were verified, but neither is introduced by
  this commit. The first is a Research producer limitation now tracked by
  #2635; this CLI intentionally preserves producer failure evidence rather than
  hiding it. The second predates this section (`Analysis Diff` already
  superseded `Changes`) and is tracked as command-level composition debt in
  #2639. No resolution commit applies in this PR because changing either would
  expand #2543 into a producer or section-architecture redesign.
- **Final exact-head re-review:** Claude Opus 4.8 and Gemini 3.1 Pro both
  returned **CLEAN**, with no outstanding findings.

richlander:
Ready to merge

richlander:
## Review — the CLI wiring the design doc reserved, done the way it prescribed

The implementation-diff design doc drew an unusual line when the component landed: "intentionally not a CLI section yet; CLI wiring should be a separate usability change that chooses section names, verbosity, table schema, and row limits without changing the product change primitive." This PR is that change, and it stays inside the line on every axis I checked:

- **The CLI is a pure consumer.** `BuildImplementationDiff` calls `ImplementationDiff.Compare` over `ResearchDiffInput.FromAssemblies`; the view builder calls `ImplementationDiff.UnifiedLines(change)` — which the doc restricts to *presentation boundaries*, and a CLI formatter is exactly that. No producer is invoked or reconstructed CLI-side, no third row family is introduced; the durable model stays producer-owned and the CLI projects it.
- **Explicit-only is the right default posture.** `ExplicitOnly => true`, `IsExpensive => true`, and the early-return path means the default API compatibility view is untouched — the test pins `DoesNotContain("API Diff")` under the section and the docs repeat "omit `-S` to retain the default API diff" everywhere the section is mentioned.
- **Honest labeling, three times.** The status callout ("body-level evidence, not public API compatibility"), the workflow doc, and the compatibility skill all say the same thing: evidence, not a compatibility classification or semantic-equivalence proof. That's the information-type discipline applied to prose, and it's what keeps this section from being misread as a breaking-change detector.
- **Fail-visible in the projection.** A change whose `UnifiedLines` come back empty still emits a row carrying `Detail ?? Descriptor.Title` rather than vanishing, and an unexpected mechanism falls through to `Mechanism.ToString()` rather than being dropped. Both arms fail toward visibility. Same for the known #2635 metadata-token IL failures — they render as rows instead of being filtered, which is the correct disposition while the underlying bug is tracked.
- **Reuse where reuse was owed.** The member-target resolver is shared with Analysis Diff via a `bodySectionName` parameter rather than a copy — so the fail-closed "method-like target" rejection behaves identically across both body-level sections, with the message correctly attributing the section (test-pinned). Acquisition, `--type`/`--member`, `--table`/`--tsv`/`--jsonl`, row limits — all existing machinery, no parallel path.

Also appreciated: the api-diff workflow renumbering quietly fixes the pre-existing duplicate section numbers (there were two `## 7`s) while inserting the new section.

**One interaction to keep on the radar — the dual-selection precedence.** `SelectsImplementationDiff(options) && !SelectsAnalysisDiff(options)` means `-S "Analysis Diff" -S "Implementation Diff"` deterministically renders Analysis and *silently* drops Implementation. Deterministic-with-a-known-winner is the right interim behavior (better than order-dependence), and I see it's tracked as the pre-existing first-lens-precedence issue in #2639 — but when that issue is resolved, the resolution should include a user-visible note when a requested section is suppressed, not just a documented precedence. A user who asked for two sections and got one, with no signal, is the same fail-silent shape this codebase has been systematically eliminating; it just lives at the CLI layer this time.

Minor, no action: `BuildImplementationDiff`'s `MergeSurfaces(...) ?? new ApiSurface()` fallback for the member-filter path is safe because an empty surface makes the resolver throw "did not resolve in either diff input" — fail-closed by composition, though it takes a moment to see; a one-line comment would save the next reader that moment.

Endorse. Both adversarial reviews were clean on exact head, the two reproduced findings are genuinely pre-existing and tracked, and the section closes #2543 exactly as scoped.

---

PR #2625, opened 2026-07-10
Add caller-driven package Finding onset

richlander:
## Summary

- resolve `Package@A..B` as an inclusive, caller-directed version vector without acquiring package payloads
- require an explicit `--at <version|#N|first|last>` cell for `type` and `member`, then acquire only that pinned package
- add sparse version-labelled Finding correlation with distinct present, missing, absent, failed, and unevaluated semantics
- document the agent-owned onset/bisect workflow

## Validation

- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config`
- `dotnet run --project src/ILInspector.Instructions.Tests -c Release --no-build` (103 passed)
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-build` (1719 run; the sole dirty-source SourceLink mismatch passed after rebuilding the committed head)
- focused range/correlation/parser tests (179 passed across targeted runs)
- live `System.Text.Json@8.0.0..8.0.5` vector, type, and member probes
- markdownlint on all changed Markdown

## Review

Two isolated adversarial reviews completed cleanly: Claude Opus 4.8 and Gemini 3.1 Pro.

PR #2642, opened 2026-07-11
Expose native API Finding transitions

richlander:
## Summary

- add an explicit `Finding Transitions` diff section for caller-selected API boundaries
- project Metadata's native type/member `PairFinding<T>` cases with endpoint labels and side presence
- preserve caller-owned version traversal while documenting the vector, point-probe, final-pair workflow

## Example

```bash
dotnet-inspect diff \
  --package System.Text.Json@8.0.6..9.0.0 \
  -t System.Text.Json.Schema.JsonSchemaExporter \
  -S "Finding Transitions"
```

The focused result is `PairFinding.Added`, with `8.0.6` absent and `9.0.0`
present. `PairFinding.Present`, `Removed`, and `Changed` use the same native
comparison path.

## Validation

- `dotnet run --project src/dotnet-inspect.Tests -c Release`: 1,743 total, 0 failed, 10 environment skips
- real package, local-library, table, JSONL, discovery, and target-validation proofs

## Adversarial review

- **Gemini 3.1 Pro:** CLEAN. Verified native `PairFinding` provenance, polarity,
  overload identities, endpoint labels, formats, discovery, and option guards.
- **Claude Opus 4.8:** initially questioned absent-both member handling. No
  resolution commit applies: `diff` is the final pair confirmation, while
  `type/member --at` own search probes; a member must exist at one endpoint to
  establish its stable identity. After clarification, the clean re-review
  verified the full feature and returned CLEAN.


richlander:
Ready to merge

richlander:
## Review — the confirmation step the bisect workflow was missing, surfaced without a costume

This closes the loop the correlation work opened. #2625 gave the caller probes (`@A..B` vectors, `--at` point cells) to *locate* a candidate boundary; the nomenclature doc has always insisted the probes are not themselves the introduction result. This section is the missing final move: run Metadata's real comparison on the chosen pair and report the **native union case by name** — `PairFinding.Added`, `Present`, `Removed`, `Changed` — with endpoint labels. No reclassification into breaking/additive, and the guard makes that explicit: combining with `--breaking`/`--additive`/`--changed`/`--name-only` is an error, because the section "reports the exact PairFinding kind." That's the information-type discipline applied at the CLI: the native transition is the good; classification is a different lens you must ask for separately.

What I verified and like:

- **Fail-closed at both ends.** `CompletePairs` throws on a `Failed` comparison rather than rendering an empty section; and all three option guards (target required, no classification filters, selected alone) run *before acquisition* — the tests pin that a bad flag combination against nonexistent DLLs produces the option error, never "not found." Validation-before-work is the right ordering and it's test-enforced.
- **Dual-selection errors instead of silently preceding.** "Finding Transitions must be selected by itself" is exactly the fail-visible behavior I flagged on #2641, where `-S "Analysis Diff" -S "Implementation Diff"` silently drops one section. When #2639 resolves the first-lens precedence question, this section's approach — refuse loudly rather than pick quietly — is the model I'd unify on.
- **The absent-both asymmetry is inherent and correctly documented.** A type target missing at both endpoints yields no row plus the "exists at neither endpoint" note (exit 0), while a member target missing at both endpoints errors from the resolver. That's not an inconsistency: a type name is nominal (the caller supplied a complete identity, so absence is representable), while a member identity must be *resolved against a surface* — absent-both means there is no stable identity to report on, which is the same reasoning that closed Opus's question. The skill doc states both readings ("no row means it exists at neither"; member targets need one endpoint). Good.
- **Endpoint labels come from pair sides, not re-derivation** — `OldSide`/`NewSide` project the union case; `absent`/`present` fall out of null-ness. The `Detail` column carries the native `Changed` detail (`sealed: false -> true` in the test) untranslated.
- **The docs land in all three places that matter** — nomenclature (probes locate, the native pair confirms), version-resolution (the workflow), and the compatibility skill (with the exact-reading table an agent needs). The three tellings agree with each other and with the code.

**One coordination flag: this collides with #2641 head-on.** Both PRs renumber `api-diff.md` sections 6–8 → 7–10 by inserting a new `## 6`, both insert an early-return block at the same spot in `DiffCommand.ExecuteAsync`, and both register a new section in `DiffSections` pipeline/schema and a new view pair in `DiffViews`/`DiffOutputFormatter`. Whichever lands second has a real (if mechanical) rebase, and the api-diff workflow will need one more renumbering pass so both new sections get stable numbers. Worth deciding the order now.

Minor, no action: `Diff_FindingTransitions_ConfirmsPackageTypeIntroduction` downloads real System.Text.Json versions — consistent with this suite's existing live-package tests, just noting it joins the network-dependent set.

Endorse. Small surface, correct boundaries, and it makes the bisect workflow honest end-to-end: probes to find the pair, native comparison to prove it.

---

richlander:
Conflict and review follow-up is resolved in `2fa5c585`.

- Merged latest `origin/main` after #2641.
- Preserved `Implementation Diff` unchanged as workflow section 6 and added
  `Finding Transitions` as section 7; subsequent workflow sections are
  renumbered through 11.
- Preserved both command branches and both schema/view registrations.
- Added a regression test proving
  `-S "Finding Transitions" -S "Implementation Diff"` fails before endpoint
  acquisition instead of silently choosing a lens.
- The live-package test remains intentional and consistent with the existing
  suite; no action applies to that minor observation.

Local validation after the merge:

- solution build succeeded
- full CLI suite: 1,751 total, 0 failed, 10 environment skips
- changed Markdown passes markdownlint

Post-merge adversarial re-reviews:

- Claude Opus 4.8: **CLEAN**
- Gemini 3.1 Pro: **CLEAN**

Both reviewers specifically verified lens precedence, dual-selection failure,
native `PairFinding` polarity, schema/source-generation registration, endpoint
disposal, and the workflow renumbering. Refreshed CI is the remaining
merge-readiness gate.

richlander:
Ready to merge

Issue #2614, opened 2026-07-10
Finding spine: harden contracts before broad adoption


richlander:
## Conclusion

Completed. The Finding spine's public contracts were hardened before broad
Analysis adoption, and the first producer migrations retired bespoke paths
rather than adding parallel adapters.

The remaining soft-matching implementation and collection-value equality work
now have focused owners in #2585 and #2669. They are follow-up capabilities,
not blockers for exact hard-tier producers.

## P0 — contract integrity

- [x] Make `FindingComparison<T>.Complete` non-forgeable.
  - #2617 made completed/failed envelopes constructible only through the shared
    comparison operation.
  - Generic inspection projection, matching, folding, failure propagation, and
    envelope construction now have one orchestration path.
  - Producer-owned classification is constrained to preserve matched atoms and
    order.
- [x] Fail closed on malformed identities and invalid Finding primitives.
  - #2617 rejects uninitialized keys and validates subjects, descriptors,
    payloads, censuses, ordinals, matches, and transformations at construction
    boundaries.
  - Empty identity content remains valid for exact empty text lines; null or
    uninitialized identity does not.

## P1 — model evolution

- [x] Tighten type semantics.
  - #2620 keeps `IFinding` observation-only, retains `IPairFinding` as the
    heterogeneous transition contract, and keeps inspection failures outside
    the observation hierarchy.
- [x] Separate ordering, correspondence, subject identity, and provenance.
  - #2624 made producer ordinal optional and explicitly non-authoritative for
    matching.
  - Metadata identity sets do not fabricate positions.
  - The evidence did not justify one universal structural anchor:
    `FindingSubject.Key`, `FindingKey`, typed member anchors, IL offsets, and
    annotation ranges retain their distinct contracts.
- [x] Define the rung-2 soft-matching shape before implementation.
  - #2585 owns the structured identity projections, typed deltas, match-tier
    provenance, strict-to-loose ladder, false-positive discipline, and scaling
    design. Hard matching remains authoritative and runs first.
- [x] Give collection value equality a focused owner before cache/dedup use.
  - #2669 now owns the equality/hash contract for Finding records and payloads
    carrying `ImmutableArray` / `ImmutableHashSet`.

## Research adoption gate

- [x] Preserve heterogeneous Research capability rather than migrating
  `ResearchDiffEvidence` / `ImplementationDiffEvidence` wholesale.
- [x] Land one end-to-end semantic proof that retires a bespoke lane.
[...]

PR #2643, opened 2026-07-11
Add Analysis call-site and unsafety Finding producers

richlander:
Adds the remaining high-value Analysis Finding producers while preserving the raw observation boundaries.

## Summary

- add total call-site and definite-unsafety ordered censuses plus an identity-multiset census for broader unsafe evidence
- keep exact correspondence separate from IL offsets and metadata-token provenance; decode `calli` signatures structurally
- route Research call-site cost/semantics and unsafety annotations through the Analysis censuses
- replace `BodySignalDiff` correspondence with shared Finding comparisons while preserving its compatibility count and offset-attribution projection
- document the producer ownership, ordering, and cross-family suppression boundaries

Advances #2614.

## Validation

- `dotnet build dotnet-inspect.slnx -c Release --configfile ./nuget.config`
- `dotnet run --project src/ILInspector.Analysis.Tests -c Release --no-build` (429 passed)
- `dotnet run --project src/ILInspector.Research.Tests -c Release --no-build` (81 passed)
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-build` (1721 passed, 10 tool-dependent skips)
- `dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class ILInspector.Decompiler.Tests.UnsafetyOccurrenceFactTests` (6 passed)
- `npx markdownlint-cli docs/architecture.md docs/overview.md docs/design/finding-producers.md`

## Review

Isolated Opus and Gemini adversarial reviews are pending.


richlander:
## Review — three producers, each in its native family; the #2638 dispositions adopted; one unreachability claim to confirm

This completes the Analysis producer set the way the doc update describes it, and the family choices are the headline: each observation kind gets the result shape its coordinates actually support, rather than one shape stretched over all three.

- **Call sites and definite unsafety are IL-ordered censuses** — offsets order the stream, ordinals are census positions, offsets/tokens stay provenance. Same rules as allocations.
- **Unsafe evidence is an identity multiset with `MatchMode: IdentitySet` and *no ordinals*** — some evidence has no IL coordinate, so fabricating stream positions would violate the coordinates doc's identity-set rule. The census is still sorted for deterministic output, and the test pins `Ordinal == null` on every finding — sorted-for-determinism without gaining semantic position, which is precisely the distinction the doc draws.
- **The `member`/`body` discriminator inside the evidence identity key** fixes a real conflation: declaration-level evidence can no longer match against body-level evidence, and `CompareUnsafe_DoesNotMatchMemberEvidenceToBodyEvidence` pins the regression.

Two of my #2638 asks show up here, both done right:

1. **`DescribeFacetChanges` gained the `"other facets changed"` fallback** — detection and description can no longer silently diverge for allocations.
2. **The `Failed` arm is now a throw with the assumption stated** — `"A comparison of total Analysis censuses cannot fail"` in `BodySignalDiff.AddComparisonRows` is exactly the shape I asked for on `AddAllocationRow`: enforce the unreachability rather than swallow. **The one thing I want confirmed before merge: is it actually unreachable?** These comparisons pass `acceptanceThreshold: 100` (call sites, unsafety). If threshold rejection is a reachable outcome for two `Complete` inspections — say a method with a very large unsafe-operation census — this throw turns a diff of that method into a crash. Fail-visible beats fail-silent, but if the threshold can fire, the projection should degrade (e.g., count-only rows) rather than throw; and if it can't fire for Complete/Complete inputs, a one-line comment saying *why* would let the next reader trust the message. Either answer is fine; the code should show which one is true.

The `BodySignalDiff` rewrite itself is the quiet win. The bespoke count-based matcher is gone — comparison now goes through the shared producers, and `BodySignalDiff` keeps only its compatibility projection (per-identity add/remove *count deltas* with offset attribution). Two properties fall out and both are test-pinned: reordered operations cancel to zero rows (the LCS's Removed+Added for a moved operation share an identity key, so the delta nets out — the old count-only semantics preserved by construction), and the richer transition data still exists underneath for consumers that want more than counts. The doc line drawing the boundary — cross-family suppression "is consumer projection policy, not a reason for either producer to return an incomplete census" — is the correct division and it's now written down.

**The calli fix removes a version-local leak from identity.** Previously an indirect call's `Unsupported` detail embedded the standalone-signature *token*, so identical function-pointer calls in different builds compared as different. Now the signature is decoded structurally (behind `SignatureBlobGuard`, with a narrow catch → `Unsupported`), and the fallback identity is the bare `calli-unsupported` bucket — the test pins that two calli differing only by token compare exact. One consequence to note in a comment: two *genuinely different* undecodable calli also compare as `Present` — the fallback bucket is deliberately coarse, quiet in the conservative direction. Rare (the guard only rejects malformed blobs), but it's the same tiered-identity softening as the allocation key and deserves the same one-line justification I asked for there.

Two smaller observations:

- **The two facet-detection patterns now coexist in one file, with opposite defaults.** Allocations use the `with`-copy structural trick — a new payload field becomes a semantic facet automatically (fail-visible). Call sites use a whitelist (`Kind`, `InLoop`, `Opcode`) — a new `DirectCall` field is silently *not* a facet. The whitelist is defensible (most `DirectCall` fields are identity or provenance, so the normalize-away set would be large), but the divergence is invisible to a reader; a comment on `SameCallSiteFacets` saying the whitelist is deliberate — and that new fields must opt in — prevents someone "fixing" it toward the allocation pattern or, worse, adding a field and assuming it participates.
- **The new composite keys are length-prefixed (`{len}:{part}`) and collision-safe** — strictly better than the allocation key's plain colon-join from #2638. Follow-up material: migrate `GetAllocationIdentityKey` onto `CompositeKey` so all four producers share one key discipline.

Verified against the code, not just the diff: the architecture table's producer names (`UnsafetyOccurrenceFactProducer`, `CallSiteCostFactProducer`, `CallSiteSemanticsFactProducer`, `MethodHeaderLeverageFactProducer`, `DecompilerLifetimeFactProducer`) all exist — the table is a correction of stale docs, not aspiration. And the Research producers now consume the censuses (with the session-index preference from #2638's pattern), so single-version annotations and two-version comparisons genuinely share one source.

Endorse pending the adversarial reviews you already have queued, with the `Failed`-unreachability question answered and ideally the two comments (calli fallback bucket, facet whitelist) added.

---

richlander:

PR head `23e975ce` has two clean isolated reviews:

- **Gemini 3.1 Pro — CLEAN.** Verified structural call and `calli` identity, token/offset exclusion, guarded malformed-signature handling, Finding ordering and multiplicity, the two unsafety families, `BodySignalDiff` count/offset preservation, and Research consumer compatibility.
- **Claude Opus 4.8 — CLEAN.** Independently traced the same identity and migration paths, confirmed `FindingComparison.Failed` is unreachable for the total `Complete` inputs, ran the focused Analysis and Research tests, and reproduced four SDK-sensitive Analysis-suite failures identically at the merge base rather than attributing them to this PR.

Both reviewers used separate clean detached worktrees at `23e975ce`; neither reported a correctness, security, or compatibility finding.

### Reconciliation

- The earlier Gemini review found that structural `calli` identity omitted its calling convention. Commit `23e975ce` adds the raw signature header and generic arity to `MemberRef`, populates them for normal methods and `calli`, includes them in correspondence identity, and adds a calling-convention regression test.
- The maintainer review asked whether the `Failed` arm can be reached, and requested rationale for malformed-`calli` fallback identity and the call-site facet whitelist. Commit `23e975ce` documents that failure only follows a `Failed` input inspection (these producers construct total `Complete` inspections), that acceptance threshold affects fringe folding rather than outcome state, that malformed `calli` identity is deliberately coarse and token-free, and that new `DirectCall` fields must explicitly opt into transition classification.
- No follow-up applies to the suggested allocation-key migration; it was identified as separate follow-up material rather than a defect in these producers.

Local solution build, Analysis tests (430/430), Research tests (81/81), CLI tests (1728 passed, 10 expected tool-dependent skips), and changed-Markdown lint completed on the rebased head. PR CI is green and GitHub reports the PR mergeable.

**Ready to merge.**

richlander:
## Follow-up @ `23e975ce` — gate cleared, all three review items landed, plus a real identity strengthening

Verified the new commit against my review:

- **The `Failed`-unreachability question is answered in code, not just in a reply**: the comment on the throw arm states the invariant precisely — `FindingComparison` fails only when an input inspection is `Failed`; both inputs are total `Complete` censuses; `acceptanceThreshold` affects fringe folding, not the comparison outcome. That's the one-line "why" I asked for, and it makes the throw trustworthy. Gate cleared.
- **The calli fallback bucket got its justification comment** ("token-free and deliberately coarse so version-local token churn does not fabricate a removal/addition") and **the facet whitelist got its opt-in note** ("new DirectCall fields are identity or provenance unless explicitly opted into transition classification here"). Both are exactly the markers I wanted against future "fixes."
- **Bonus that goes beyond my review**: `MemberRef` now carries the raw ECMA-335 `SignatureHeader` and `GenericArity`, both keyed into call-site identity. Previously two function-pointer calls differing only in calling convention (or `explicitthis`/`generic` flags — facts the parameter and return types genuinely cannot encode) compared identical; now they're a Removed/Added transition, test-pinned. That's a real identity hole closed, populated at all three decode sites (calli resolution, MethodDefinition, MemberReference), so definition-based and reference-based identities stay aligned.

Nothing further from me — endorsement stands, now unconditional on my side, pending your queued adversarial reviews.

---

Issue #2545, opened 2026-07-08
C2: CFG-backed stack-slot copy propagation


richlander:
## Summary

Track a C2 follow-up for CFG/lifetime-backed stack-slot copy propagation: safely collapse simple synthetic copies such as `S_dst = S_src; use(S_dst)` into direct uses of `S_src` when the copied value is provably stable.

## Context

PR #2539 reduced printer-owned split slots by allowing unresolved enum-like constants to share the same typed carrier when the printer can already spell the value as `(T)0`. In the Roslyn `Binder::GetMethodGroupOrLambdaDelegateType` witness, the output now keeps the source slot typed correctly:

```csharp
RefKind S_20;
ScopedKind S_22;

S_20 = V_1[V_22];
S_20 = (RefKind)0;
S_21 = S_20;

S_22 = V_5[V_22];
S_22 = (ScopedKind)0;
S_23 = S_22;

S_16.Add(new AnonymousTypeField(S_17, S_18, S_19, S_21, S_23, S_24, S_26, S_27));
```

The remaining opportunity is not another printer typing tweak. `S_21` and `S_23` are copy carriers. They could potentially collapse to direct uses:

```csharp
S_16.Add(new AnonymousTypeField(S_17, S_18, S_19, S_20, S_22, S_24, S_26, S_27));
```

## Required proof

This needs CFG/lifetime evidence, not printer-local text matching. A safe rule should prove at least:

- destination slot has a single use;
- destination slot is assigned exactly from a source slot;
- the source slot is definitely assigned on every path to the copy;
- no write to the source slot can occur between the copy and the destination use on any path;
- the source/destination slots are in the same body scope (do not cross lambdas/local functions);
- the rewrite does not move side effects or change evaluation order;
- render A/B and, where checkable, fidelity evidence confirm the collapse is safe.

## Starting witness

- `Microsoft.CodeAnalysis.CSharp.Binder::GetMethodGroupOrLambdaDelegateType`
- After #2539, this method still reports 2 un-unified split slots in `DecompilerHarness --slot-unifier-census` and has the simple copy-carrier shape around `S_21 = S_20` and `S_23 = S_22`.

## Non-goals

- Do not add printer-only copy folding based on names or local output text.
- Do not assume a copy is removable without CFG/lifetime proof.
- Do not broaden type rules; this is a value lifetime/copy propagation slice.

richlander:
[...]

PR #2645, opened 2026-07-11
Add CFG-backed stack slot copy propagation

richlander:
- Advances #2545
- Changes the decompiler pipeline to remove proof-backed synthetic stack-slot copy carriers before slot materialization.
- Card revision: `701bd10f`

## Change

**Conclusion:** **REVIEW** — the targeted CFG/lifetime rule removes five copy-carrier render artifacts with no valid→invalid render regressions; adversarial reviews are pending.

Before:

```csharp
S_20 = parameterRefKinds[i];
...
S_21 = S_20;
...
S_22 = parameterScopes[i];
...
S_23 = S_22;
...
S_16.Add(new AnonymousTypeField(S_17, S_18, S_19, S_21, S_23, S_24, S_26, S_27));
```

After:

```csharp
S_20 = parameterRefKinds[i];
...
S_22 = parameterScopes[i];
...
S_16.Add(new AnonymousTypeField(S_17, S_18, S_19, S_20, S_22, S_24, S_26, S_27));
```

The new pass only rewrites `S_dst = S_src` when the destination has exactly one store/use, the source is the deferred multi-block residual class, a same-container CFG walk proves no path rewrites the source before the use, and no outside edge can enter the use region. It declines nested scopes, multiple destination uses, source rewrites, non-slot copy values, and lock receiver carriers.

## Evidence

| Check | Result |
| --- | --- |
| Build | `dotnet build src/ILInspector.Decompiler.Tests/ILInspector.Decompiler.Tests.csproj -c Release --configfile nuget.config --no-restore` passed |
| Focused tests | `StackSlotCopyPropagationPassTests` passed (8 tests) |
| Census harness test | `SlotResidualCensus_PrintsPostF2MeasurementCard` passed |
| Pipeline ledger/stage tests | `LoweringCoverageTests` + `PipelineStageTests` passed (20 tests) |
| Real witness | `Microsoft.CodeAnalysis.CSharp.Binder::GetMethodGroupOrLambdaDelegateType` fixed |
| Full decompiler executable | Not used as a blocker: `CoreLibStaticFieldLock_PreservesCopiedReceiverLocal` fails on this SDK/runtime and reproduces on exact base `/tmp/ab-base-2545` |

### C2 slot-unifier census

Explicit merge-base `de0337f4` vs PR head:

| Metric | Base | PR | Delta |
| --- | ---: | ---: | ---: |
| StoreStackSlot nodes reaching printer | 49,803 | 49,797 | -6 |
| LoadStackSlot nodes reaching printer | 77,032 | 77,026 | -6 |
| Distinct slots considered by printer unifier | 43,715 | 43,709 | -6 |
| Methods with residual stack slots | 7,975 | 7,975 | 0 |
| Single-candidate slots | 43,202 | 43,198 | -4 |
| Multi-candidate slots unified by printer | 357 | 357 | 0 |
| Un-unified split slots | 208 | 206 | -2 |
| Stack-slot declarations emitted | 19,117 | 19,114 | -3 |

Pass impact:

```text
Microsoft.CodeAnalysis.CSharp.Binder::GetMethodGroupOrLambdaDelegateType
Microsoft.CodeAnalysis.CSharp.Symbols.SourceMemberContainerTypeSymbol::AfterMembersChecks
Microsoft.CodeAnalysis.CSharp.CodeGen.CodeGenerator::TryEmitOptimizedReadonlySpan
DotnetInspector.Output.ApiOutputFormatter::ToUnsafeMemberRow
DotnetInspector.Output.ApiOutputFormatter.<>c::<PopulateIndexSections>b__29_21

stack-slot-copy-propagation changed 5/89065 methods (0.01%); 0 pass bugs
```

## Decompiler quality

**Conclusion:** **PASS** — deterministic explicit-base corpus comparison matched baseline tolerances; render A/B found no valid→invalid regressions.

### Render A/B

Explicit merge-base render snapshot, `--workers 20`:

```text
Render A/B Check: 89065 methods evaluated
Changed: 5 (structural: 3, paren-equivalent: 0, unparsed: 2)
Semantic: valid->valid: 2, invalid->valid: 0, valid->invalid: 0, invalid->invalid: 3
Added:   0
Removed: 0
```

### Generated quality-diff card

```markdown
### Decompiler quality diff

Correctness coverage: validity compiled 25,183 methods (compile-cap 4,000; per-sample, not corpus-wide); fidelity sampled 36 / 89,065 (0.04%)

| Semantic defects (-) | 69 (0.27%) → 69 (0.27%) (neutral) | 0 |
| Fidelity opcode diffs (-) | 5 (13.89%) → 5 (13.89%) (neutral) | 0 |
| Fidelity exact (+) | 31 (86.11%) → 31 (86.11%) (neutral) | 0 |
| Fidelity recompile failures (-) | 0 (0.00%) → 0 (0.00%) (neutral) | 0 |
| Fidelity context failures (-) | 0 (0.00%) → 0 (0.00%) (neutral) | 0 |

Pinned-subset gate (PR quick rate/count regressions evaluated here): Fully raised 88.03% -> 88.03% (0 pp); conditional residual 2.70% -> 2.70% (0 pp); Full malformed 0 -> 0 (0); opcode diffs 5 -> 5 (0).

```

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | Pending | Adversarial review requested after PR open |
| Gemini Pro | Pending | Adversarial review requested after PR open |

**Review conclusion:** **REVIEW** — pending two clean adversarial reviews.

## Validation

```bash
dotnet build src/ILInspector.Decompiler.Tests/ILInspector.Decompiler.Tests.csproj -c Release --configfile nuget.config --no-restore
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class ILInspector.Decompiler.Tests.StackSlotCopyPropagationPassTests
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -method '*SlotResidualCensus_PrintsPostF2MeasurementCard*'
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class ILInspector.Decompiler.Tests.LoweringCoverageTests -class ILInspector.Decompiler.Tests.PipelineStageTests

dotnet run --project tools/DecompilerHarness -c Release -- "${assemblies[@]}" --slot-unifier-census
dotnet run --project tools/DecompilerHarness -c Release -- "${assemblies[@]}" --pass-impact stack-slot-copy-propagation
dotnet run --project /tmp/ab-base-2545/tools/DecompilerHarness -c Release -- "${assemblies[@]}" --workers 20 --emit-render-ab /tmp/c2-copy-base.renderab
dotnet run --project tools/DecompilerHarness -c Release -- "${assemblies[@]}" --workers 20 --render-ab /tmp/c2-copy-base.renderab
dotnet run --project tools/DecompilerHarness -c Release -- "${assemblies[@]}" --diff-corpus-baseline /tmp/c2-copy-explicit-base-corpus-baseline-seq.json --quality-diff-card --corpus-fidelity-cap 3 --max-examples 3 --sequential
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/StackSlotCopyPropagationPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/StackSlotCopyPropagationPassTests.cs
@@ -252,6 +252,67 @@ public class StackSlotCopyPropagationPassTests
     }
 
     [Fact]
+    public void DoesNotPropagateWhenNestedBranchCanEnterUseRegion()
+    {
+        var body = new BlockContainer();
+        var entry = new Block(0);
+        entry.Add(new ConditionalBranch(new LoadArgument(1, "flag", Bool), 8));
+        body.Add(entry);
+        var firstSourceStore = new Block(4);
+        firstSourceStore.Add(new StoreStackSlot(0, new LoadArgument(0, "x", Int32)));
+        firstSourceStore.Add(new Branch(12));
+        body.Add(firstSourceStore);
+        var secondSourceStore = new Block(8);
+        secondSourceStore.Add(new StoreStackSlot(0, new LoadArgument(2, "y", Int32)));
+        secondSourceStore.Add(new ConditionalBranch(new LoadArgument(1, "flag", Bool), 16));
+        body.Add(secondSourceStore);
+        var copyJoin = new Block(12);
+        copyJoin.Add(new StoreStackSlot(1, new LoadStackSlot(0, Int32)));
+        copyJoin.Add(new Branch(24));
+        body.Add(copyJoin);
+        var nestedBranch = new Block(16);
+        var thenArm = new Block(17);
+        thenArm.Add(new Branch(24));
+        nestedBranch.Add(new IfStatement(new LoadArgument(1, "flag", Bool), thenArm, null));
+        nestedBranch.Add(new Return(null));
+        body.Add(nestedBranch);
+        var use = new Block(24);
+        use.Add(UseSlot(1));
+        use.Add(new Return(null));
+        body.Add(use);
+        var function = Function(body);
+
+        new StackSlotCopyPropagationPass().Run(function, PassContext.None);
+        function.CheckInvariant();
+
+        Assert.Contains(function.Descendants.OfType<StoreStackSlot>(), store => store.Slot == 1);
+        Assert.Contains(function.Descendants.OfType<LoadStackSlot>(), load => load.Slot == 1);
+    }
+
+    [Fact]
+    public void DoesNotPropagateWhenNestedLoopTerminatorIsPresent()
+    {
+        var body = MultiBlockSourceBeforeCopy();
+        var copy = new Block(12);
+        copy.Add(new StoreStackSlot(1, new LoadStackSlot(0, Int32)));
+        body.Add(copy);
+        var use = new Block(16);
+        use.Add(UseSlot(1));
+        var thenArm = new Block(17);
+        thenArm.Add(new Continue());
+        use.Add(new IfStatement(new LoadArgument(1, "flag", Bool), thenArm, null));
+        use.Add(new Return(null));
+        body.Add(use);
+        var function = Function(body);
+
+        new StackSlotCopyPropagationPass().Run(function, PassContext.None);
+        function.CheckInvariant();
+
+        Assert.Contains(function.Descendants.OfType<StoreStackSlot>(), store => store.Slot == 1);
+        Assert.Contains(function.Descendants.OfType<LoadStackSlot>(), load => load.Slot == 1);
+    }
+
+    [Fact]
     public void DoesNotCrossNestedBodyScope()
     {
         var lambdaBody = new BlockContainer();
--- a/src/ILInspector.Decompiler.Tests/StackSlotCopyPropagationPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/StackSlotCopyPropagationPassTests.cs
@@ -1,0 +1,262 @@
+using ILInspector.Decompiler.Pipeline;
+
+namespace ILInspector.Decompiler.Tests;
+
+public class StackSlotCopyPropagationPassTests
+{
+    static readonly TypeRef Holder = TypeRef.Definition("Synthetic", "Samples", "Holder", ValueTypeHint.ReferenceType);
+    static readonly TypeRef Void = TypeRef.CoreLib("System", "Void");
+    static readonly TypeRef Int32 = TypeRef.CoreLib("System", "Int32");
+    static readonly TypeRef Bool = TypeRef.CoreLib("System", "Boolean");
+    static readonly TypeRef Object = TypeRef.CoreLib("System", "Object");
+
+    [Fact]
+    public void PropagatesSingleUseCopyAcrossFallthroughBlock()
+    {
+        var body = MultiBlockSourceBeforeCopy();
+        var copy = new Block(12);
+        copy.Add(new StoreStackSlot(1, new LoadStackSlot(0, Int32)));
+        body.Add(copy);
+        var use = new Block(16);
+        use.Add(UseSlot(1));
+        use.Add(new Return(null));
+        body.Add(use);
+        var function = Function(body);
+
+        new StackSlotCopyPropagationPass().Run(function, PassContext.None);
+        function.CheckInvariant();
+
+        Assert.DoesNotContain(function.Descendants.OfType<StoreStackSlot>(), store => store.Slot == 1);
+        Assert.DoesNotContain(function.Descendants.OfType<LoadStackSlot>(), load => load.Slot == 1);
+        Assert.Single(function.Descendants.OfType<LoadStackSlot>(), load => load.Slot == 0);
+    }
+
+    [Fact]
+    public void PropagatesOnlyWhenEveryBranchPathKeepsSourceStable()
+    {
+        var body = MultiBlockSourceBeforeCopy();
+        var copy = new Block(12);
+        copy.Add(new StoreStackSlot(1, new LoadStackSlot(0, Int32)));
+        copy.Add(new ConditionalBranch(new LoadArgument(1, "flag", Bool), 20));
+        body.Add(copy);
+        var fallthrough = new Block(16);
+        fallthrough.Add(new Branch(24));
+        body.Add(fallthrough);
+        var target = new Block(20);
+        target.Add(new Branch(24));
+        body.Add(target);
+        var join = new Block(24);
+        join.Add(UseSlot(1));
+        join.Add(new Return(null));
+        body.Add(join);
+        var function = Function(body);
+
+        new StackSlotCopyPropagationPass().Run(function, PassContext.None);
+        function.CheckInvariant();
+
+        Assert.DoesNotContain(function.Descendants.OfType<StoreStackSlot>(), store => store.Slot == 1);
+        Assert.DoesNotContain(function.Descendants.OfType<LoadStackSlot>(), load => load.Slot == 1);
+        Assert.Single(function.Descendants.OfType<LoadStackSlot>(), load => load.Slot == 0);
+    }
+
+    [Fact]
+    public void DoesNotPropagateWhenSourceIsRewrittenBeforeUse()
+    {
+        var function = RewriteSourceBeforeUse();
+
+        new StackSlotCopyPropagationPass().Run(function, PassContext.None);
+        function.CheckInvariant();
+
+        Assert.Contains(function.Descendants.OfType<StoreStackSlot>(), store => store.Slot == 1);
+        Assert.Contains(function.Descendants.OfType<LoadStackSlot>(), load => load.Slot == 1);
+    }
+
+    [Fact]
+    public void DoesNotPropagateWhenDestinationHasMultipleUses()
+    {
+        var body = MultiBlockSourceBeforeCopy();
+        var copy = new Block(12);
+        copy.Add(new StoreStackSlot(1, new LoadStackSlot(0, Int32)));
+        body.Add(copy);
+        var firstUse = new Block(16);
+        firstUse.Add(UseSlot(1));
+        body.Add(firstUse);
+        var secondUse = new Block(20);
+        secondUse.Add(UseSlot(1));
+        secondUse.Add(new Return(null));
+        body.Add(secondUse);
+        var function = Function(body);
+
+        new StackSlotCopyPropagationPass().Run(function, PassContext.None);
+        function.CheckInvariant();
+
+        Assert.Contains(function.Descendants.OfType<StoreStackSlot>(), store => store.Slot == 1);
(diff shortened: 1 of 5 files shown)


richlander:
Adversarial review reconciliation for the current head `07441e5e`:

- First-pass Opus 4.8 and Gemini 3.1 both found the same soundness issue: `StackSlotCopyPropagationPass.Successors()` omitted `Leave` and `SwitchBranch` edges. Fixed in `72d665c3` by modeling `Leave.TargetOffset`, every `SwitchBranch.TargetOffsets` entry, and switch fallthrough.
- Second-pass Gemini found that nested structured control flow could hide branches/leaves from the flat CFG proof; second-pass Opus found nested `Break`/`Continue` bypassed the loop-terminator guard. Fixed in `07441e5e` by making the pass flat-CFG-only: it declines candidate containers with nested `Block`/`BlockContainer`, `Break`, or `Continue`, while leaving lambda/local-function scopes to the existing separate scope walk.
- Third-pass Gemini 3.1 review on `07441e5e`: CLEAN. Evidence checked included flat-CFG enforcement, successor modeling, domination/source-stability, lock/increment/nested-scope hazards, and the census F2 lookup.
- Third-pass Opus 4.8 review on `07441e5e`: CLEAN. Evidence checked included structured-node coverage through nested `Block`/`BlockContainer`, all offset-carrying flat terminators, adversarial flat-loop probes, copy-store deletion invariants, and `SlotResidualCensus.F2PassIndex`.

Validation refreshed on `07441e5e`:

- `StackSlotCopyPropagationPassTests`: 15/15 passed.
- `SlotResidualCensus_PrintsPostF2MeasurementCard`: 1/1 passed.
- `LoweringCoverageTests`: 6/6 passed.
- `PipelineStageTests`: 14/14 passed.
- Pass impact: `stack-slot-copy-propagation changed 1/89065 methods (0.00%); 0 pass bugs`.
- Slot census: un-unified split slots 206; stack-slot declarations 19,115.
- Render A/B: 1 changed method; 0 valid→invalid; 1 pre-existing invalid→invalid.
- Explicit-base quality card matched tolerances; semantic defects improved 69 → 54, fidelity unchanged.

CI is green on `07441e5e`. Ready to merge.

richlander:
## Adversarial review — one constructible unsound rewrite; the guard set is otherwise real. Gate on fixing the back-edge case.

Since the PR is explicitly in REVIEW status pending adversarial passes, here is one. I verified every claimed decline has a real guard in code (non-slot/self-copy, single dst store+load, multi-block source residual, nested scopes via three independent layers, lock receiver carrier, flat-container restriction), the DFS terminates by construction (`visited` over `(block, startStatement)` pairs — at most 2·|blocks| states, back-edges can't spin it), it fails closed on out-of-range indices, the pipeline placement is correct (after the final slots-only inlining, before `SlotMaterializationPass`), and the `NativePasses` ledger registration is load-bearing against `LoweringCoverageTests`. The census delta (−6 loads/stores/slots, 0 change to multi-candidate unification) matches the five-witness claim.

### The hole: a back-edge from the use block defeats both proofs at once

Three facts compose, all verified directly in `TryGetCopyPath`/`SourceStableOnEveryPath`/`RegionDominatedByCopy`:

1. The region walk **stops at the use block** — `if (state.BlockIndex == useBlockIndex) continue;` — so the use block's successors are never traversed.
2. The stability scan **caps the use block at the use statement** — `int end = blockIndex == path.UseBlockIndex ? path.UseStatement.ChildIndex : ...` — so a write to `S_src` *after* the use in the same block is never seen.
3. The predecessor check asks only that region blocks' predecessors be **in the region** — and the use block itself is in the region.

So in a flat container: `B4: S0 = x` / `B8: S0 = y` (two blocks — multi-block residual ✓) / `B12: S1 = S0` (single store ✓) / `B16: Use(S1); S0 = z; if (cond) goto B16; return` (single load ✓, no unmodeled flow ✓). Region = {B12, B16}; B16's predecessors are {B12, B16} ⊆ region ✓; the `S0 = z` sits past the scan cap ✓. The pass rewrites `Use(S1)` → `Use(S0)`: the original reads the stale copy on every iteration; the rewrite reads `z` from iteration 2 onward. **And this shape is reachable at this stage by the pipeline's own admission** — `StructuringPass` declines backward conditional branches (`Record("cond-backward-branch"); // backward = loop (later slice)`), leaving exactly these flat-with-back-edge containers for late passes to meet. There is no back-edge-shaped test in the suite.

Two sound fixes, either works: **(a)** walk *through* the use block (drop the early `continue`) and scan every region block fully (drop the `end` cap) — the cap exists to tolerate src-rewrites on paths that never return, but with a back-edge those paths do return, and full-scan only declines cases that were unsound anyway; or **(b)** cheaper, require the region to be acyclic — decline when any region block's successor lands back in the region. Both preserve all five current witnesses (forward/diamond shapes).

### Two smaller soundness notes, one question

- **Cross-container edges are invisible to the dominance proof.** `Predecessors` maps only `container.Blocks`, but a catch handler is a sibling subtree, and ECMA-335 permits `leave` from a catch back into its associated try region. If copy+use live in a flat `TryBody` container and a catch rewrites `S_src` then re-enters, neither check can see it. Question for you rather than a gate: can the importer/structurer actually produce catch-re-entry into a mid-region offset at this stage? If provably not, a comment saying so; if unknown, decline candidates whose container is a try body.
- **A nested container's entry block has an implicit external predecessor** (loop-body containers re-enter block 0 every iteration) that the in-container predecessor map doesn't record. Exempting only `CopyBlockIndex` doesn't cover a region that includes block 0 by another route. Low severity, but the "region entry = copy block" invariant is unchecked for index 0.

### Test gaps — the report card for the guard set

- **The lock-guard test is non-discriminating**: the `Lock` node in the fixture contains a `BlockContainer`, so `HasUnmodeledControlFlow` declines the candidate even with `FeedsLockReceiverCarrier` deleted. A mutant without the lock guard passes the suite. The fixture needs the lock in a *sibling* container so only the guard fires.
- **Completely untested guards**: `use.Parent is IncrementDecrement` (zero tests — and its neighboring-pass rationale, invalid `1++`, is exactly the kind of thing that regresses silently), destination multi-*store* (only multi-*load* is tested), self-copy, and `IsMultiBlockSourceResidual`'s negative directions. And nothing exercises a back-edge, i.e., the machinery where the real hole lives.
- The three "nested" negative tests all actually exercise the single blunt `HasUnmodeledControlFlow` check rather than the edge analysis their names suggest — fine as written, but the names over-claim.

Also noting, no action needed: the lock guard's `Root(store).Descendants` scan crosses into lambda bodies with independent local numbering — over-decline in the fail-safe direction, but wrong-scope reasoning worth a comment; unresolved `Branch` targets are silently treated as container exits (under-approximates reachability — only matters if a malformed container reaches this pass); and `use.Type ?? source.Type` can stamp the propagated load with a type that disagrees with other `S_src` loads, which `SlotMaterializationPass`'s one-type rule then vetoes — a residual-stays-residual risk, not correctness.

### Verdict

The architecture is right — proof-backed, decline-by-default, flat-region-only, correctly placed, honestly measured. But hole A is a real miscompile-class defect in the exact container shape this pass was built to visit, so: **fix the back-edge case (option a or b) plus a regression test with a back-edge fixture before merge**. The discriminating-lock-test and untested-guard items should ride along in the same commit.

---

richlander:
Addressed the latest PR feedback on the current head `e03cd14c`:

- Fixed the flat back-edge soundness hole by declining copied-value regions with cycles, so `Use(S_dst); S_src = z; if (cond) goto use` cannot rewrite to `Use(S_src)` across later iterations.
- Limited candidates to independent method/lambda/local-function body containers, avoiding nested try/lock/loop/switch body implicit-entry and cross-container concerns.
- Added regression/guard coverage for cyclic regions, destination multi-store, increment/decrement targets, self-copy, source-residual negative cases, and nested structured-container decline. `StackSlotCopyPropagationPassTests` now has 22 passing tests.

Refreshed evidence on `e03cd14c`:

- `StackSlotCopyPropagationPassTests`: 22/22 passed.
- `SlotResidualCensus_PrintsPostF2MeasurementCard`: 1/1 passed.
- `LoweringCoverageTests`: 6/6 passed.
- `PipelineStageTests`: 14/14 passed.
- Pass impact: `stack-slot-copy-propagation changed 1/89065 methods (0.00%); 0 pass bugs`.
- Slot census: un-unified split slots 206; stack-slot declarations 19,115.
- Render A/B: 1 changed method; 0 valid→invalid; 1 pre-existing invalid→invalid.
- Explicit-base quality card matched tolerances; semantic defects improved 69 → 52 and fidelity opcode diffs improved 5 → 4.

Fourth-pass adversarial reviews on `e03cd14c`:

- Gemini 3.1: CLEAN. Checked cycle detection/dominance interaction, independent-container scope, increment/decrement interaction, flat CFG successor modeling, and `SlotResidualCensus.F2PassIndex`.
- Opus 4.8: CLEAN. Checked the use-block back-edge fix, container restriction, flat-CFG guard, successor over-approximation, dominance/source-stability, test discrimination, and `F2PassIndex`.

CI is green and merge state is CLEAN on `e03cd14c`. Ready to merge.

richlander:
## Re-review @ `e03cd14c` — gate cleared

The fix takes my option (b) and then goes further:

- **`RegionHasCycle`** — a three-color DFS over the region using the *same* `Successors` edge model — declines any cyclic copied-value region. Hole A's construction (use-block back-edge hiding a post-use `S_src` rewrite) is now unrepresentable: the self/back-edge makes the region cyclic, the candidate is declined, and `DoesNotPropagateAcrossCyclicCopiedValueRegion` pins exactly that fixture.
- **`IsIndependentBodyScope`** (container's parent must be `IrFunction`/`Lambda`/`LocalFunctionStatement`) closes holes B and C in a single guard: try-body containers (catch re-entry blindness) and nested loop-body containers (implicit entry-block predecessor) are simply no longer candidates. That's a stronger answer than the comment I asked for — the question of whether those shapes are reachable no longer matters. `DoesNotPropagateInsideNestedStructuredContainer` pins it.
- **Every untested guard I named now has a discriminating test**: destination multi-store, self-copy, single-store source, same-block multi-store source, and `IncrementDecrement`.

One item from my review I don't see addressed: the **lock-guard test is still non-discriminating** (its `Lock` fixture lives in the same container, so `HasUnmodeledControlFlow` — and now also `IsIndependentBodyScope` — would decline even with `FeedsLockReceiverCarrier` deleted). In fact, with the independent-body-scope guard in place, it's worth asking whether the lock guard is now dead code: a `lock` statement's receiver carrier inside an independent body still trips `HasUnmodeledControlFlow` via the `Lock`'s nested `BlockContainer`. Either construct a fixture where only the lock guard fires, or delete the guard with a comment explaining why the other two guards subsume it. Follow-up grade, not a gate — redundant guards are safe, just misleading about what's load-bearing.

Soundness objection withdrawn. With the cyclic-region decline and the independent-scope restriction, every path from copy to use is acyclic, same-container, and fully scanned — the proof obligations now actually hold. Endorse.

---

PR #2626, opened 2026-07-10
Add CSharp formatter seam

richlander:
## Summary

- add a concrete, immutable `CSharpFormatter` facade in `ILInspector.CSharp` for member, type, and declaration-unit formatting over Metadata shapes
- route `CSharpTypePrinter`, dotnet-inspect API declaration rows, and Decompiler's body-independent declaration spelling through the facade
- cache translated Metadata options and formatter variants, reuse type-printer formatters by namespace, and avoid redundant member-list materialization
- keep Metadata's existing declaration writer as the compatibility implementation during the structured-spelling migration

## Design

This establishes the composition boundary without changing Metadata producers while #2574 is still hardening that surface:

```text
Metadata shapes
    -> consumer-owned API/body selection
        -> CSharpFormatter for declaration text
            -> consumer-owned source/Markout composition
```

`CSharpFormatter` does not query or group APIs, render Markdown, depend on Decompiler/Research, or own body policy. `CSharpTypePrinter` remains the higher namespace/type composition layer. Decompiler combines formatter-produced declarations with reconstructed bodies.

The facade intentionally omits a public identifier-spelling API: Metadata's current helper is signature-text-specific, so exposing it as a general qualified-name formatter would create the wrong primitive contract.

## Compatibility

- `type StringBuilder --markdown -v:n --tips q` is byte-identical before/after on the same SDK/runtime.
- Final-head render A/B against merge base `11198494` evaluated 89,065 methods: 0 changed, 0 added, 0 removed.
- Existing Metadata C# compatibility APIs remain available.
- The default `--shape` compatibility path is unchanged.

## Validation

- `ILInspector.CSharp.Tests`: 28 passed
- `dotnet-inspect.Tests`: 1,710 passed, 10 tool-dependent skips
- `TypeSourceCheckTests`: 33 passed
- `TypeBindCheckTests`: 5 passed
- Release solution build: passed
- targeted net10 CSharp build: passed with zero warnings/errors
- markdownlint: passed

The exhaustive Decompiler run has reported four failures that each reproduce at clean `origin/main`; they are tracked in #2629, #2630, #2631, and #2632.

### Decompiler quality diff

Corpus: #1166 real-world decompiler corpus sensor: #1150 pinned NuGet assemblies plus dotnet-inspect managed assemblies. 14 assemblies, 89,065 methods
Correctness coverage: validity compiled 350 methods (compile-cap 25; per-sample, not corpus-wide); fidelity sampled 36 / 89,065 (0.04%)

[...]

Issue #1166, opened 2026-06-22
Real-world corpus regression baseline (continuous CI sensor)


richlander:
## Goal

Stand up a fixed real-world corpus measured as a **continuous CI baseline-diff**
on decompiler PRs — a *sensor*, not a manual re-baseline step. Follow-up to
recommendation #2 of the [architecture review](https://gist.github.com/richlander/f3ed4a639670133b9105b245081b3604)
and the #1150 design spike; tracked under #1148.

## Why

- **CoreLib is unrepresentative.** The #1150 spike measured fully-raised at
  **96.46% (CoreLib) → 89.71% (NuGet) → 82.68% (dotnet-inspect itself)**, and the
  forward-merge structural gap at **1.51% → 2.74% → 2.11%**. The tool is materially
  weaker on the code users actually decompile, and that signal is invisible to a
  CoreLib-only view.
- **Floors are not baselines.** Today's corpus floors tolerate slow regressions
  (a sub-floor slip ships silently). A baseline-diff catches per-method/pattern
  regressions.
- **It dissolves the "wait on the next re-baseline" trap.** When the number is
  ambient (always current in CI), re-baselining stops being a discrete event that
  gates decisions. This is the prerequisite the #1150 step-4 start-trigger leans on.
- **It makes #1150's numbers machine-checkable** (review rec #5): the spike's NuGet
  table is currently prose; checking the corpus in converts it to a reproducible
  baseline.

## Scope

- **Seed corpus (pin exact versions):** Newtonsoft.Json 13, Roslyn 5.0
  (`Microsoft.CodeAnalysis[.CSharp]`), System.CommandLine 3.0-preview,
  NuGet.Versioning 7.3, ApplicationInsights 2.23, plus dotnet-inspect's own
  assemblies. (These are the #1150 set.)
- **Metrics per run:** per-method fully-raised %, `structuring: conditional-branch`
  residual, forward-merge candidates (`--structuring-stops`), and fidelity/validity
  defect counts.
- **Mechanism:** check in a baseline; CI diffs each decompiler PR against it;
  regressions beyond a documented tolerance fail or block merge.

## Acceptance

- Baseline committed and reproduces the #1150 corpus table.
- CI job runs the diff on decompiler-touching PRs.
- Documented tolerance thresholds and how to read the diff.

## Guardrails

- Keep the product path SRM-only, NativeAOT-friendly, Roslyn-free.
- This is a standing sensor (continuous), not a manual deliberation gate.
[...]

PR #1150, opened 2026-06-22
Design spike (#1148): sequence the dominator/retained-label rewrite

richlander:
Design spike for #1148 — the considered response to recommendation #1 of the
[adversarial architecture review](https://gist.github.com/richlander/f3ed4a639670133b9105b245081b3604)
("finish the dominator-driven structurer is the top priority"). Per the issue, this
is a **docs-only design spike** and the durable record of the decision. Adds a
`## Design spike (#1148)` section to `docs/design/control-flow-structuring.md`.

## Decision: sequence step 4, don't shelve it

An earlier draft of this spike concluded "don't start step 4 yet," sized entirely on
CoreLib. The [PR review](https://gist.github.com/richlander/a6f12e0ca8c426ee034be29a01b3f7a2)
correctly flagged that CoreLib is the one corpus the team already knows is
unrepresentative. Re-baselining on a **real-world corpus** reversed the conclusion.

| Corpus | Methods | Fully raised | Fwd-merge candidates |
| --- | --- | --- | --- |
| CoreLib (preview.5) | 41,012 | 96.46% | 1.51% |
| NuGet (Roslyn, Newtonsoft, System.CommandLine, NuGet.Versioning, App Insights) | 67,448 | 89.71% | **2.74%** |
| dotnet-inspect itself | 20,083 | 82.68% | **2.11%** |

The forward-branch-to-common-exit shape step 4 targets is the **dominant structural
gap on every corpus and denser off CoreLib** — not a shrinking remainder. Combined
with the mechanism argument (the range model provably cannot name an out-of-range
post-dominator) and the exhausted cheap normalizer route (+3 from the last terminator
slice), the step-4 go/no-go is effectively decided. What remains is sequencing:

1. **Land the bounded return-tail-aware guard combiner first** — protects the #640
   canary, unblocks #921, no invariant change. Filed as **#1167**.
2. **Wire up the real-world corpus regression sensor before the rewrite lands** —
   review rec #2; a tooling prerequisite / regression sensor, *not* an authorization
   gate on starting. Filed as **#1166**.
3. **Then begin the non-terminator retained-label prototype** when the slope has
   flattened (a diamond/guard normalizer PR moves the real-world residual < ~0.1pp)
   and a `--dump` confirms single-merge readability. If readability fails, the shape
   stays flat *by stated policy*. This keeps step 4 last/riskiest while ensuring
   "gated" cannot quietly become "never."

## What the spike also corrected (per review)

[...]

Issue #2631, opened 2026-07-10
Class-union not-pattern test emits lowered nullable expression


richlander:
## Failure

`TypeSourceComposerUnionTests.ClassUnionSwitchExpression_RendersTypePatternArms` no longer raises the `IsNotCat` fixture to the expected class-union pattern.

## Reproduction

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
  -method "*ClassUnionSwitchExpression_RendersTypePatternArms*"
```

Expected:

```csharp
return pet is not null && pet is not Cat;
```

Actual begins:

```csharp
return !(pet is null ? false : (pet.Value is Cat))...
```

The failure reproduces in a clean worktree at `origin/main` (`11198494c848881121638dcd536c27cd472a8e6d`). It makes the full Decompiler suite red for every agent.

PR #2649, opened 2026-07-11
Preserve structured labels and align quality samples

richlander:
Incorporates the quality-card sampling feedback from #2626.

- Preserves branch-target labels when structuring and boolean-folding replace statements.
- Keeps externally targeted infinite-loop labels outside the synthesized loop scope.
- Makes semantic/fidelity card outcomes and pinned gates neutral when sampled method identities differ.
- Card revision: `13671f5c`

## Change

**Conclusion:** **PASS** — exhaustive validity removes 14 `CS0159` defects with
zero regressions, and render A/B has no valid-to-invalid transition.

Before (`CSharpSpellability.TypeReason`):

```csharp
if (__dotnet_inspect_switch0 == 0) goto IL_003C;
// ...
if (IsCoreLibPrimitive(type))
```

After:

```csharp
if (__dotnet_inspect_switch0 == 0) goto IL_003C;
// ...
IL_003C:
if (IsCoreLibPrimitive(type))
```

The same sibling defect affected statement replacements in
`BooleanFoldingPass`; all replacement families now inherit the replaced
statement's offset. Infinite-loop synthesis also transfers a targeted loop-head
offset to the `while` statement instead of leaving its label inside the loop,
where an external `goto` would be illegal.

## Evidence

| Check | Result |
| --- | --- |
| Product/solution build | Pass |
| Focused tests | 25 passed |
| Decompiler fast suite | 2,250 passed |
| Decompiler exhaustive suite | 2,384 passed, 1 skipped, 2 failures reproduced identically on clean `main` |
| Exhaustive validity | 73,064 `Full` methods compiled; 14 `CS0159` improvements; 0 regressions |
| Render A/B | 89,065 methods; 18 changed; 14 invalid-to-valid, 0 valid-to-invalid, 4 invalid-to-invalid |
| Real witness | `CSharpSpellability.TypeReason` restores `IL_003C`; `System.Xml.XmlSubtreeReader.Read` restores `IL_00F8` and `IL_0109` |

The 18 render changes are 17 structural and one unparsed classification. The
changed-method compile-back lane attempted all 18: 5 existing opcode diffs, 12
recompile failures, and 1 generated-member context failure; no row was
`NotFull`, but none produced an exact opcode proof. These are reported as
current outcomes, not passing evidence. The semantic render lane and exhaustive
validity differential are the decisive evidence for this label-validity fix.

## Decompiler quality

**Conclusion:** **ADVISORY** — structural pinned gates are neutral; semantic and
fidelity outcome gates are intentionally ungated because their sampled method
identities differ. The remaining verdict items are only the quick run's lower
validity cap/coverage versus the daily baseline.

### Decompiler quality diff

| Semantic defects (sampling differs) | 54 (0.21%) → 0 (0.00%) | n/a |
| Fidelity opcode diffs (sampling differs) | 4 (11.11%) → 4 (11.11%) | n/a |
| Fidelity exact (sampling differs) | 32 (88.89%) → 32 (88.89%) | n/a |

Pinned-subset gate (PR quick rate/count regressions evaluated here): Fully raised 88.03% -> 88.03% (0 pp); conditional residual 2.70% -> 2.70% (0 pp); Full malformed 0 -> 0 (0); fidelity ungated (sampling differs; rely on changed-method fidelity).

Regressions:

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Gemini 3.1 Pro | Clean on `13671f5c` | Found the infinite-loop scope hole; fix and focused regression verified clean |
| Claude Opus 4.8 | Clean on `13671f5c` | Mutation-tested loop-label ownership and checked ordinary/EH loop siblings |

**Review conclusion:** **PASS** — two independent exact-head reviews are clean.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.SwitchBranchRenderingTests -class ILInspector.Decompiler.Tests.BooleanFoldingSourceOffsetTests -class ILInspector.Decompiler.Tests.CorpusSensorComparisonTests
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/BooleanFoldingSourceOffsetTests.cs
+++ b/src/ILInspector.Decompiler.Tests/BooleanFoldingSourceOffsetTests.cs
@@ -14,6 +14,31 @@ public class BooleanFoldingSourceOffsetTests
     static readonly TypeRef Holder = TypeRef.CoreLib("Synthetic", "Holder");
 
     [Fact]
+    public void NestedGuardFold_PreservesSourceOffset()
+    {
+        var innerThen = new Block(0);
+        innerThen.Add(new Return(null));
+        var inner = new IfStatement(new LoadArgument(1, "b", Boolean), innerThen, null);
+        var outerThen = new Block(0);
+        outerThen.Add(inner);
+        var outer = new IfStatement(new LoadArgument(0, "a", Boolean), outerThen, null);
+        outer.SetSourceOffset(0x08);
+
+        var block = new Block(0);
+        block.Add(outer);
+        var function = Function(
+            block,
+            Void,
+            [new Parameter("a", Boolean), new Parameter("b", Boolean)]);
+
+        new BooleanFoldingPass().Run(function, PassContext.None);
+
+        var folded = Assert.IsType<IfStatement>(Assert.Single(block.Children));
+        Assert.IsType<LogicalBinary>(folded.Condition);
+        Assert.Equal(0x08, folded.SourceOffset);
+    }
+
+    [Fact]
     public void GuardReturnFold_PreservesSourceOffset()
     {
         var then = new Block(0);
--- a/src/ILInspector.Decompiler.Tests/BooleanFoldingSourceOffsetTests.cs
+++ b/src/ILInspector.Decompiler.Tests/BooleanFoldingSourceOffsetTests.cs
@@ -7,7 +7,10 @@ namespace ILInspector.Decompiler.Tests;
 public class BooleanFoldingSourceOffsetTests
 {
     static readonly TypeRef Boolean = TypeRef.CoreLib("System", "Boolean");
+    static readonly TypeRef Byte = TypeRef.CoreLib("System", "Byte");
     static readonly TypeRef Int32 = TypeRef.CoreLib("System", "Int32");
+    static readonly TypeRef Object = TypeRef.CoreLib("System", "Object");
+    static readonly TypeRef Void = TypeRef.CoreLib("System", "Void");
     static readonly TypeRef Holder = TypeRef.CoreLib("Synthetic", "Holder");
 
     [Fact]
@@ -53,6 +56,80 @@ public class BooleanFoldingSourceOffsetTests
 
         var folded = Assert.IsType<Return>(Assert.Single(block.Children));
         Assert.Equal(0x20, folded.SourceOffset);
+    }
+
+    [Fact]
+    public void BooleanElementStoreRetype_PreservesSourceOffset()
+    {
+        var store = new StoreElement(
+            Byte,
+            new LoadArgument(0, "values", TypeRef.SzArray(Boolean)),
+            new Constant(0, Int32),
+            new Constant(1, Int32));
+        store.SetSourceOffset(0x30);
+
+        var block = new Block(0);
+        block.Add(store);
+        var function = Function(
+            block,
+            Void,
+            [new Parameter("values", TypeRef.SzArray(Boolean))]);
+
+        new BooleanFoldingPass().Run(function, PassContext.None);
+
+        var folded = Assert.IsType<StoreElement>(Assert.Single(block.Children));
+        Assert.Equal(0x30, folded.SourceOffset);
+    }
+
+    [Fact]
+    public void CoalesceStoreFold_PreservesSourceOffset()
+    {
+        var initial = new StoreLocal(0, Object, new LoadArgument(0, "value", Object));
+        initial.SetSourceOffset(0x40);
+        var then = new Block(0);
+        then.Add(new StoreLocal(0, Object, new LoadArgument(1, "fallback", Object)));
+        var guard = new IfStatement(
+            new Comparison(
+                ComparisonKind.Equal,
+                isUnsigned: false,
+                new LoadArgument(0, "value", Object),
+                new Constant(null, Object)),
+            then,
+            null);
+
+        var block = new Block(0);
+        block.Add(initial);
+        block.Add(guard);
+        var function = Function(
+            block,
+            Void,
+            [new Parameter("value", Object), new Parameter("fallback", Object)]);
+
+        new BooleanFoldingPass().Run(function, PassContext.None);
+
+        var folded = Assert.IsType<StoreLocal>(Assert.Single(block.Children));
+        Assert.Equal(0x40, folded.SourceOffset);
+    }
+
+    [Fact]
+    public void SlotDiamondFold_PreservesSourceOffset()
+    {
+        var then = new Block(0);
+        then.Add(new StoreStackSlot(0, new Constant(1, Int32)));
+        var @else = new Block(0);
+        @else.Add(new StoreStackSlot(0, new Constant(2, Int32)));
+        var diamond = new IfStatement(new LoadArgument(0, "c", Boolean), then, @else);
+        diamond.SetSourceOffset(0x50);
+
+        var block = new Block(0);
+        block.Add(diamond);
+        block.Add(new Return(new LoadStackSlot(0, Int32)));
+        var function = Function(block, Int32, [new Parameter("c", Boolean)]);
+
+        new BooleanFoldingPass().Run(function, PassContext.None);
+
+        var folded = Assert.IsType<StoreStackSlot>(block.Children[0]);
+        Assert.Equal(0x50, folded.SourceOffset);
     }
 
     static IrFunction Function(
--- a/src/ILInspector.Decompiler.Tests/BooleanFoldingSourceOffsetTests.cs
+++ b/src/ILInspector.Decompiler.Tests/BooleanFoldingSourceOffsetTests.cs
@@ -1,0 +1,72 @@
+using System.Collections.Immutable;
+
+using ILInspector.Decompiler.Pipeline;
+
+namespace ILInspector.Decompiler.Tests;
+
+public class BooleanFoldingSourceOffsetTests
+{
+    static readonly TypeRef Boolean = TypeRef.CoreLib("System", "Boolean");
+    static readonly TypeRef Int32 = TypeRef.CoreLib("System", "Int32");
+    static readonly TypeRef Holder = TypeRef.CoreLib("Synthetic", "Holder");
+
+    [Fact]
+    public void GuardReturnFold_PreservesSourceOffset()
+    {
+        var then = new Block(0);
+        then.Add(new Return(new Constant(true, Boolean)));
+        var guard = new IfStatement(new LoadArgument(0, "c", Boolean), then, null);
+        guard.SetSourceOffset(0x10);
+
+        var block = new Block(0);
+        block.Add(guard);
+        block.Add(new Return(new Constant(false, Boolean)));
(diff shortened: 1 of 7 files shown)


richlander:

Exact reviewed head: `13671f5c` against `f30d3611`.

| Reviewer | Final result | Reconciliation |
| --- | --- | --- |
| Claude Opus 4.8 | **CLEAN** | Initial review found the guard-return sibling of the label-provenance defect. `acbbd372` fixed return folds and `f04aec02` completed the statement-replacement sibling audit. Final review mutation-tested the loop-label fix and checked ordinary/EH loop shapes. |
| Gemini 3.1 Pro | **CLEAN** | Found missing direct nested-guard coverage, resolved by `53d4feca`, then found the infinite-loop scope hole, resolved by `13671f5c`. Final review verified the label belongs before the synthesized `while`, with the recursive body stamp suppressed. |

Both reviewers independently re-reviewed the full final diff in isolated clean
worktrees and reported no outstanding correctness, compatibility, or material
coverage findings.

One Gemini first-round item requested a committed
`System.Xml.XmlSubtreeReader.Read` canary. No resolution commit applies: that
method is manual corpus evidence, not a claimed committed fixture; the committed
real-importer canary is `CSharpSpellability.TypeReason`. The PR body now states
that distinction explicitly.

richlander:
## Final readiness

**Ready to merge.**

- PR CI is green.
- Claude Opus 4.8 and Gemini 3.1 Pro both returned **CLEAN** on exact head
  `13671f5c`.
- Exhaustive decompiler suite: 2,384 passed, 1 skipped, 2 failed. Both failures
  are the same `PointerStoreUsesOriginalAddress` fidelity-docket mismatch and
  reproduce identically on clean `main` (`f30d3611`) in both
  `FidelityGateTests` and `LoweredFidelityGateTests`.
- `PointerStoreUsesOriginalAddress` is absent from the 18-method render A/B
  delta, so those failures are not introduced by this PR.
- Exhaustive corpus validity has 14 `CS0159` improvements and zero regressions;
  render A/B has 14 invalid-to-valid and zero valid-to-invalid transitions.

No merge performed.

richlander:
## Review — the structuring fix is architectural, the card fix closes my #2626 complaint; one same-class gap left open in the folds

Three changes in one PR, so taking them in order of how much I trust each.

### The StructuringPass fix is the right kind of fix

The old code stamped the branch-target offset on the *raw detached statement* before the structuring switch ran — and when the switch consumed that statement into an `IfStatement`/`WhileLoop`/`Break`, the stamped node was discarded and the label vanished. The new code stamps the first node **actually emitted** for the block, after the switch runs (`result.Children[resultStart]`). That kills the vanish-by-consumption class at its root for block starts rather than patching the observed instance. The infinite-loop variant is equally sound: `suppressStartTargetLabel` keeps the stamp off the body's first inner statement and `SetSourceOffset` puts it on the `WhileLoop` node — and the correctness argument is worth writing down somewhere, because it's subtle: moving a label *outward* is always scope-safe (an enclosing-scope label is reachable from nested statements, and `goto` to the head of a `while (true)` is semantically re-entry), and the printer recomputes `_labelTargets` from gotos that *survive* structuring, so heads whose back-edges were all consumed get no label at all. The dangerous direction — label trapped inside a scope — is exactly what's removed. The evidence matches the mechanism: 73,064 methods compiled, 14 `CS0159` fixed, 0 regressions, 0 valid→invalid in the 89K render A/B, with real witnesses (`CSharpSpellability.TypeReason`, `XmlSubtreeReader.Read`) and the end-to-end test decompiling the actual product method.

### The quality-card fix is the resolution of the #2626 sampling artifact, done properly

On #2626 I flagged that "Semantic defects 54→0 (good)" was the same sampling-cap artifact as the flagged regression, wearing a favorable face. This PR fixes the *mechanism*, not the label: `HaveSameMethodSample` compares the **full sampled identity sets** (assembly!type::method+signature), not cap values — the comment even nails the insight ("equal caps do not guarantee equal samples"). When populations differ, the row renders as "Semantic defects (sampling differs)" with no `(good)`/`(bad)` verdict and an `n/a` count delta, and pinned regressions gate only on matching populations. Nine tests pin both directions, including the null-per-method-detail case failing closed toward neutral. The rate movement still *displays* while neutralized — correct choice; hiding it would be its own dishonesty.

One asymmetry to fix while you're here: when samples differ, the **fidelity** summary visibly flips to "ungated (sampling differs; rely on changed-method fidelity)", but the **pinned count gates are skipped silently** — no failure and no marker. A genuine semantic-defect regression accompanied by any sample churn (corpus growth, cap landing differently, a method flipping into `full-malformed` and thereby *exiting the population*) produces nothing. The trade toward no-false-positives is deliberate and comment-documented, but silent-skip is the shape this repo keeps eliminating: give the count gates the same visible "ungated" line the fidelity summary already has, so a skipped gate is greppable in the run output.

### The BooleanFolding coverage leaves one instance of the same defect class open

The seven fold sites now inherit the **replaced** statement's offset — but the folds also *consume* statements, and those offsets are silently dropped. `FoldGuardedReturn` detaches `tailReturn` and inherits only from `guard`; the nested-guard fold discards the inner `IfStatement`'s offset. If a consumed statement was itself a live goto target, its label still vanishes — precisely the CS0159 class this PR fixes, one node over. Nothing in the pass checks the consumed statement's branch-target status, and `InheritSourceOffset` no-ops silently when the replacement already carries an offset. Either the folds should decline (or double-stamp) when the consumed statement is a branch target, or a comment should state why consumed tails can't be live targets at this pipeline stage — right now that's an unproven assumption sitting where the last bug lived. Related residue, minor: the exhaustive validity run is strong evidence this doesn't fire *in the current corpus*, which is why this is an ask rather than a gate.

Smaller items:

- **Conditional `while` loops are the un-fixed sibling**: `BuildWhileCondition` consumes the condition block's statements and the new `WhileLoop` at that site gets no offset — an external goto targeting the condition block's offset still loses its label. The PR fixed the infinite-loop family only; worth a tracking issue so it's a known frontier rather than a surprise.
- **The while-stamp now depends on the printer's first-wins label dedup** (`_emittedLabels`): the loop node and its first inner statement legitimately share an offset, and the label lands outside only because the outer node prints first. The ordering is test-pinned (`IndexOf("IL_0010:") < IndexOf("while (true)")`), which is good — but first-wins is order-dependent, not scope-aware, and the `Clone()` path in terminator inlining copies offsets, so same-offset pairs exist elsewhere too. A one-line comment on `AppendLabel` naming the invariant ("outer occurrence must print first; the structuring stamp relies on this") would protect it.
- Missing witness: the `StoreStackSlot` branch of the coalesce fold has no direct test (the `StoreLocal` branch does); and the `Branch → Break` site doesn't inherit while the three `Leave →` sites do — covered by the block-start stamp, but the inconsistency invites a wrong copy-paste later.

Endorse — the A/B is decisive, the structuring rearchitecture is the correct fix shape, and the card change is exactly what #2626's review asked for. The consumed-statement offset drop and the silent count-gate skip are the two things I'd address before this merges.

---

richlander:
## Follow-up review resolution

Addressed the latest feedback in `51db6610`:

- Boolean folds now decline before mutation when any consumed statement is a
  live branch target. Coverage includes nested guards, both guarded-return
  arms, ternary returns, coalesce fallback stores, slot diamonds, and
  local-function target-scope collisions.
- The pinned-subset summary now marks skipped Full-malformed and semantic-defect
  count gates as `ungated (sampling differs)` instead of silently omitting
  them. The predicates mirror the hard-gate sample predicates.
- Added the missing `StoreStackSlot` coalesce source-offset witness.
- `Branch -> Break` now inherits source provenance consistently with the
  existing `Leave -> Break` paths.
- Documented the printer's outer-first label ownership invariant.
- Filed conditional-while label provenance as #2664 with synthetic/EH acceptance
  criteria.

Evidence on the rebased exact head:

- solution build passed;
- focused source-offset and corpus-sensor tests passed;
- exhaustive decompiler suite: 2,451 passed, 2 failed; both failures reproduce
  on clean current `main` and remain tracked by #2630 and #2631;
- feedback-only render A/B: 89,065 methods, **0 changed**;
- full PR render A/B remains 18 changed methods, 14 invalid-to-valid, zero
  valid-to-invalid;
- updated quality card prints
  `semantic defects ungated (sampling differs)` in the pinned-subset line. Its
  only verdict failures are the documented quick-run cap mismatch against the
  exhaustive baseline.

The prior `Ready to merge` status is superseded while exact-head adversarial
re-review and CI run.

richlander:
## Exact-head adversarial review reconciliation

Exact reviewed head: `51db6610` against `59565b7f`.

| Reviewer | Final result | Reconciliation |
| --- | --- | --- |
| Claude Opus 4.8 | **CLEAN** | Verified every statement-consuming Boolean fold, nested-function target isolation, outer-first printer label ownership, structuring provenance, pinned-gate predicate alignment, and non-vacuous regressions. The observed `CoreLibStaticFieldLock_PreservesCopiedReceiverLocal` failure reproduced identically on clean `main` and remains tracked by #2630; no resolution commit applies. |
| Gemini 3.1 Pro | **CLEAN** | Independently verified consumed-target preservation across fold siblings, `EnclosingBodyScope` isolation, `Branch -> Break` provenance, printer ordering, and visible pinned-gate `ungated` output. No actionable findings; no resolution commit applies. |

Both reviewers examined isolated clean worktrees at the exact PR head and reported no outstanding findings. The prior feedback is resolved by `51db6610`; the conditional-while sibling frontier is explicitly tracked by #2664.

## Final readiness

**Ready to merge.**

- PR CI is green at exact head `51db6610` and GitHub reports the PR mergeable.
- Claude Opus 4.8 and Gemini 3.1 Pro both returned **CLEAN** exact-head reviews.
- Exhaustive decompiler suite: 2,451 passed, 2 failed; both failures reproduce on clean current `main` and are tracked by #2630 and #2631.
- Feedback-only render A/B: 89,065 methods, **0 changed**.
- Full PR render A/B remains 18 changed, 14 invalid-to-valid, and zero valid-to-invalid.

No merge performed.

richlander:
## Follow-up @ `51db6610` — both holds released

- **The consumed-statement gap is closed by declining, which is the stronger option.** `ConsumesBranchTarget` checks the statements a fold would discard (tail returns, inner guards) against live branch targets collected from the enclosing body scope, and the fold declines rather than silently dropping a targeted statement's label. That removes the unproven "consumed tails are never live targets" assumption I flagged — the pass now *checks* instead of assumes, and the new tests pin both directions (fold proceeds when the tail is untargeted, declines when it's a target).
- **Skipped count gates are now visibly ungated**: "Full malformed ungated (sampling differs)" and "semantic defects ungated (sampling differs)" join the fidelity summary's existing marker, test-pinned — a skipped gate is now greppable in run output rather than silent.
- Bonus: the `Branch → Break` site gained its `InheritSourceOffset(branch)`, closing the inconsistency with the three `Leave →` sites that I flagged as copy-paste bait.

Both pre-merge asks met, with the label fix taken in the fail-closed direction. Endorsement now unconditional — this is safe to merge. (The conditional-`while` sibling remains the tracked frontier from my original review; unchanged and fine as follow-up.)

---

PR #2650, opened 2026-07-11
Compose selected diff sections

### LINKED ISSUE #2639: diff silently drops sections from multi-section selections

richlander:
## Summary

- compose every selected diff section in Markdown and JSON, including empty-state summaries
- reject explicit table, TSV, and JSONL output when multiple sections are selected
- register `diff --json`/`--markdown` and preserve single-section/default rendering

## Validation

- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-restore -- -class DotnetInspector.Tests.DiffCommandTests`
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-restore -- -class DotnetInspector.Tests.CommandLineTests`
- `npx markdownlint-cli docs/architecture.md`

## Adversarial review

Claude Opus 4.8 and Gemini 3.1 Pro both identified the missing real CLI JSON registration in the first round. Commit c8437879 wires JSON/Markdown through the shared format resolver and adds end-to-end CLI coverage. Both reviewers returned CLEAN on the final head.

Fixes #2639

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -92,6 +92,8 @@ Compares two package, platform, or local-library versions:
 - `-S "Analysis Diff"` selects body-signal changes.
 - `-S "Implementation Diff"` selects Research-composed C# + IL evidence grouped
   by member.
+- Multiple selected sections compose in Markdown and JSON. Explicit
+  table/TSV/JSONL output requires exactly one selected section.
 - Version range syntax is `Package@v1..v2` or `old/Foo.dll..new/Foo.dll`.
 
 ### find
--- a/src/dotnet-inspect.Tests/DiffCommandTests.cs
+++ b/src/dotnet-inspect.Tests/DiffCommandTests.cs
@@ -803,6 +803,119 @@ public class DiffCommandTests
         Assert.DoesNotContain("API Diff", output, StringComparison.Ordinal);
     }
 
+    [Fact]
+    public async Task ExecuteAsync_MultipleSelectedSections_ComposesMarkdown()
+    {
+        var v1 = FixtureCatalog.DiffPair.OldAssemblyPath();
+        var v2 = FixtureCatalog.DiffPair.NewAssemblyPath();
+
+        var (exitCode, output, error) = await ConsoleCapture.RunAsync(() =>
+            DiffCommand.ExecuteAsync(new DiffOptions
+            {
+                LibraryVersionRange = $"{v1}..{v2}",
+                Select = ["Analysis Diff", "Implementation Diff"],
+                TypeFilter = ["DiffSample"],
+                ChangedOnly = true
+            }));
+
+        Assert.Equal(0, exitCode);
+        Assert.Empty(error);
+        Assert.Contains("## Analysis Diff", output, StringComparison.Ordinal);
+        Assert.Contains("## Implementation Diff", output, StringComparison.Ordinal);
+        Assert.Contains("RegressesAllocInLoop", output, StringComparison.Ordinal);
+        Assert.Contains("ConstantValue", output, StringComparison.Ordinal);
+    }
+
+    [Fact]
+    public async Task ExecuteAsync_MultipleSelectedSections_ExplicitTableReturnsError()
+    {
+        var v1 = FixtureCatalog.DiffPair.OldAssemblyPath();
+        var v2 = FixtureCatalog.DiffPair.NewAssemblyPath();
+
+        var (exitCode, output, error) = await ConsoleCapture.RunAsync(() =>
+            DiffCommand.ExecuteAsync(new DiffOptions
+            {
+                LibraryVersionRange = $"{v1}..{v2}",
+                Select = ["Changes", "Implementation Diff"],
+                OneLine = true,
+                OneLineExplicitlySet = true
+            }));
+
+        Assert.Equal(1, exitCode);
+        Assert.Empty(output);
+        Assert.Contains("Selection matches 2 sections", error, StringComparison.Ordinal);
+        Assert.Contains("--table, --tsv, and --jsonl display one section at a time", error, StringComparison.Ordinal);
+    }
+
+    [Fact]
+    public async Task ExecuteAsync_MultipleSelectedSections_ComposesJson()
+    {
+        var v1 = FixtureCatalog.DiffPair.OldAssemblyPath();
+        var v2 = FixtureCatalog.DiffPair.NewAssemblyPath();
+
+        var (exitCode, output, error) = await ConsoleCapture.RunAsync(() =>
+            DiffCommand.ExecuteAsync(new DiffOptions
+            {
+                LibraryVersionRange = $"{v1}..{v2}",
+                Select = ["Analysis Diff", "Implementation Diff"],
+                JsonOutput = true,
+                TypeFilter = ["DiffSample"],
+                ChangedOnly = true
+            }));
+
+        Assert.Equal(0, exitCode);
+        Assert.Empty(error);
+        using var document = System.Text.Json.JsonDocument.Parse(output);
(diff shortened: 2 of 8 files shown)


richlander:

- **Claude Opus 4.8:** first round found that the new JSON path was unreachable because `diff` did not register `--json`. The finding was reproduced through the real CLI.
- **Gemini 3.1 Pro:** independently found the same registration gap and additionally noted that direct flag parsing bypassed `DOTNET_INSPECT_FORMAT=json`.
- **Resolution:** c8437879 registers `--json` and `--markdown`, resolves JSON through the shared format resolver, treats JSON as raw output, and adds an end-to-end command-line regression test. No findings were dismissed.
- **Final re-review:** both Claude Opus 4.8 and Gemini 3.1 Pro returned **CLEAN** on c8437879. They verified `@All`, empty selected sections, explicit tabular diagnostics, implicit allocation-analysis selection, real CLI/environment JSON routing, NativeAOT source generation, row limiting, and disposal paths.

richlander:
Ready to merge

richlander:
## Review — #2639 resolved in the strongest direction; one dropped disclaimer to restore

When I flagged the silent first-lens precedence on #2641, the options were: document the precedence, warn on suppression, or compose. This picks **compose** — a user selecting two sections gets both, in one document, with empty states rendered as summaries rather than vanishing. That's better than the warning I asked for, and the constraint it imposes is exactly right: table/TSV/JSONL demand one section (`ValidateSingleSectionForTabular`, checked up front, fail-visible error) because a projected table cannot carry two schemas. Markdown and JSON compose; tabular selects. Clean rule, stated in the docs.

Also right:

- **JSON uses `WhenWritingNull`** so unselected sections are *absent* from the document, not `null` placeholders — the JSON shape mirrors the selection.
- **The section-name string literals became `DiffSections.*.Name` references** — small, but it removes the three copies of `"Analysis Diff"` that could drift.
- **The adversarial-review round demonstrably worked**: both reviewers caught that `--json`/`--markdown` weren't actually registered on the command, and the fix came with end-to-end CLI coverage rather than just the missing two lines. The single-section and default paths are untouched (the composed path is entered only for `--json` or multi-select), so existing behavior is preserved by construction.

**One should-fix: composition drops the per-section Status callouts.** `DiffDocumentView` carries each section's `Summary` but not its `Status`. The Implementation Diff section's callout — "C# and IL implementation evidence is body-level evidence, not public API compatibility" — was the honest-labeling line #2641 put in three places, and it disappears in exactly the scenario where it matters most: when Implementation Diff renders *directly beneath the API compatibility section* in one document, where a reader is most likely to conflate the two. Carrying the callout text into the composed section (a paragraph under the summary would do) restores it.

Two smaller notes:

- `RenderDocumentView` hand-encodes the column arrays for all three sections — a third copy of the schema alongside `DiffSections.CreateSchema()` and the per-section Markout views. Adding a column to `ImplementationDiffRow` now requires remembering this spot too. Not blocking, but it's the kind of drift that shows up six months later as a column missing only from composed output.
- **Coordination with #2642**: Finding Transitions ships with a "must be selected by itself" guard that predates composition. Once both land, `-S "Finding Transitions" -S Changes` errors while every other combination composes. That may be the right call (it's a focused confirmation lens, and its schema is tabular-ish), but it should be a conscious decision recorded in the guard's error message or docs — otherwise it reads as the last un-migrated section rather than a deliberate exemption.

Endorse with the callout restored. Nice resolution of #2639 — composing beats both silent precedence and warn-and-drop.

---

richlander:
**PR feedback and post-merge review reconciliation**

- **Status callouts:** resolved in [`ccca00b7`](https://github.com/richlander/dotnet-inspect/commit/ccca00b7). Composed Analysis Diff and Implementation Diff sections now retain their body-level-evidence disclaimers without duplicating empty-state summaries.
- **Presentation-free JSON:** follow-up review found Analysis Diff member identities could carry Markdown `<code>` markup and HTML entities into JSON. Resolved in [`256f9cf6`](https://github.com/richlander/dotnet-inspect/commit/256f9cf6) by keeping raw member data at the model boundary and applying code decoration only while rendering Markdown.
- **Composed schema duplication:** no resolution commit applies. This was a non-blocking maintainability note rather than a current correctness defect; the manual composed renderer necessarily defines each heterogeneous table shape, and focused tests now cover its columns and raw JSON values.
- **Finding Transitions coordination:** resolved deliberately in merge commit [`31106a1f`](https://github.com/richlander/dotnet-inspect/commit/31106a1f). It remains a focused endpoint-confirmation lens selected by itself; `-S @All` now explains the exception and directs callers to select `Changes`, `Analysis Diff`, and `Implementation Diff` explicitly. The workflow documents the contract, and single-section Finding Transitions JSON is supported through the composed document model.

**Final adversarial review on `31106a1f`:**

- **Claude Opus 4.8:** **CLEAN** — reviewed the full `origin/main...HEAD` diff and verified the 1,759-test suite, Release build, multi-section Markdown/JSON, tabular rejection, empty states, disclaimers, raw identities, NativeAOT source generation, and Finding Transitions routing.
- **Gemini 3.1 Pro:** **CLEAN** — independently verified composition, explicit tabular failures, empty-section preservation, presentation-free JSON/JSONL, source-generated serialization, focused Finding Transitions behavior, and disclaimer rendering.

No final findings were dismissed. Refreshed CI is green.

richlander:
Ready to merge

richlander:
## Follow-up @ `31106a1f` — should-fix landed, plus two things I didn't ask for and am glad to see

- **The disclaimers survive composition**: `DistinctStatusMessage` carries each section's Status callout into composed Markdown and JSON, with the equality guard correctly suppressing it when the status merely repeats the empty-state summary — so the "body-level evidence, not public API compatibility" line now renders exactly where Implementation Diff sits beneath the API table. That was my one should-fix.
- **Composed JSON is now presentation-free** (`256f9cf`): Analysis Diff member identities stored raw, code decoration applied only in Markdown-oriented renderers — HTML tags and entities no longer leak into the JSON document. Right layering; good catch.
- **The #2642 interaction got the conscious decision I asked for**: the merged Finding Transitions guard message now states *why* it's exempt from composition ("a focused endpoint-confirmation lens; select comparison sections explicitly instead of using @All"). It reads as a deliberate exemption rather than an un-migrated section.

Endorsement now unconditional. One coordination note: #2661 extends the Finding Transitions guard block in `DiffCommand.cs` from a pre-composition base, so whichever of these merges second has a small conflict to reconcile in that region — the guard-message wording and the `--finding` selector logic both live there.

---

PR #2651, opened 2026-07-11
Classify pointer store compile-back diff

### LINKED ISSUE #2644: PointerStoreUsesOriginalAddress fails both fidelity dockets on main
## Problem

Both Stage 8 opcode bosses report `PointerStoreUsesOriginalAddress` as a new
compile-back opcode diff:

- `FidelityGateTests.NoNewOpcodeDiffsBeyondKnownDocket`
- `LoweredFidelityGateTests.NoNewOpcodeDiffsBeyondKnownDocket`

The failure reproduces on clean current `origin/main` at `de0337f4`, so it is a
repository baseline failure rather than a feature-branch regression.

The fixture is:

```csharp
public static unsafe int PointerStoreUsesOriginalAddress(int* ptr, int* next)
{
    *(ptr) = (ptr = next) == null ? 5 : 5;
    return *next;
}
```

`Rung6PointerStore_PreservesOriginalAddressAcrossArgumentStore` separately
asserts that the raised source stores through a preserved temporary address,
not the reassigned `ptr`. The fixture was introduced by `252f07c0` / #2465,
but it is not present in either fidelity known-diff docket.

## Reproduction

```bash
  -method "*NoNewOpcodeDiffsBeyondKnownDocket*"
```

Result on `de0337f4`:

```text
Total: 2, Failed: 2
New fidelity check opcode diffs: PointerStoreUsesOriginalAddress
New lowered fidelity check opcode diffs: PointerStoreUsesOriginalAddress
```

The same two failures also appeared in the exhaustive Decompiler run associated
with the C# formatter work; that run had 2,367 tests, with the other four
failures already tracked by #2629-#2632.

## Expected

Determine whether this is:

1. a real compile-back fidelity regression in original-address preservation,
   which should be fixed and pinned exact; or
2. an intentional residual introduced with #2465, which should be explicitly
   classified in both dockets with the opcode delta documented.

Do not silently leave both opcode bosses red.

richlander:
Fixes #2644

## Change

**Conclusion:** **PASS** - `PointerStoreUsesOriginalAddress` is an intentional compile-back opcode residual, not an unpinned product regression.

The fixture still renders through the preserved original pointer address (`*S_...`, not `*ptr =`) while Roslyn compile-back introduces locals around the conditional value/store, producing a canonical opcode diff.

## Evidence

| Check | Result |
| --- | --- |
| Stage 8 opcode boss | `NoNewOpcodeDiffsBeyondKnownDocket` passed for sugared and lowered gates |
| Pointer checkability pins | Sugared and lowered `PointerStoreUsesOriginalAddress_StaysCompileBackCheckable` passed |
| Address-preservation guards | Raised and lowered `Rung6*PointerStore_PreservesOriginalAddressAcrossArgumentStore` passed |

Opcode evidence:

```text
orig : ldarg ldarg dup starg ldc.i4 conv.u beq ldc.i4 br ldc.i4 stind.i4 ldarg ldind.i4 ret
recmp: ldarg ldarg stloc dup starg ldc.i4 conv.u beq ldc.i4 br ldc.i4 stloc ldloc ldloc stind.i4 ldarg ldind.i4 ret
```

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | Clean after fix | First pass requested explicit checkability and lowered address-preservation guards; `33445884` added them. Second pass reported CLEAN with only a non-blocking assertion wording note. |
| Gemini 3.1 Pro | Clean | Confirmed both dockets are updated, both gates stay green, and the new guards cover the silent `RecompileFail` and lowered-shape risks. |

**Review conclusion:** **PASS** - two clean second-pass adversarial reviews; the only first-pass finding was resolved by `33445884`.

## Validation

```bash
dotnet restore dotnet-inspect.slnx --configfile nuget.config
dotnet run --no-restore --project src/ILInspector.Decompiler.Tests -c Release -- \
  -method "*NoNewOpcodeDiffsBeyondKnownDocket*" \
  -method "*PointerStoreUsesOriginalAddress_StaysCompileBackCheckable*" \
  -method "*Rung6PointerStore_PreservesOriginalAddressAcrossArgumentStore*" \
  -method "*Rung6LoweredPointerStore_PreservesOriginalAddressAcrossArgumentStore*"
```

Result: `Total: 6, Errors: 0, Failed: 0, Skipped: 0`.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
@@ -102,6 +102,13 @@ public class FidelityGateTests
         // over-render. Pre-existing slow-docket gap surfaced by running the gate
         // locally — the lowered/sugared gates are Speed=Slow (Deep Inspect / publish only).
         "CharConditionalElementStore",
+        // PointerStoreUsesOriginalAddress (#2644): the raised source preserves
+        // the original pointer address across the argument reassignment
+        // (`*S_... =`, never `*ptr =`), but compile-back introduces locals around
+        // the conditional value before the indirect store. Valid C#, not
+        // opcode-exact; the address-preservation shape is pinned separately by
+        // Rung6PointerStore_PreservesOriginalAddressAcrossArgumentStore.
+        "PointerStoreUsesOriginalAddress",
         // Unmasked by the in/out skeleton-parameter fix (#1931): these CfgSampleClass
         // methods were RecompileFail before — the whole reconstructed type failed to
         // compile because InOperatorVec's in-parameter operators rendered as illegal
--- a/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
@@ -107,7 +107,8 @@ public class FidelityGateTests
         // (`*S_... =`, never `*ptr =`), but compile-back introduces locals around
         // the conditional value before the indirect store. Valid C#, not
         // opcode-exact; the address-preservation shape is pinned separately by
-        // Rung6PointerStore_PreservesOriginalAddressAcrossArgumentStore.
+        // Rung6PointerStore_PreservesOriginalAddressAcrossArgumentStore and
+        // checkability is pinned below.
         "PointerStoreUsesOriginalAddress",
         // Unmasked by the in/out skeleton-parameter fix (#1931): these CfgSampleClass
         // methods were RecompileFail before — the whole reconstructed type failed to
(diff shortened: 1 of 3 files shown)


richlander:
## Review — docket growth done the way docket growth should be done

Adding a fixture to a known-diff docket is normally where test coverage goes to die: the entry suppresses the failure, and nothing thereafter notices if the fixture rots from "known benign diff" into "doesn't compile at all." This PR adds the entry *and* the anti-rot machinery in the same change, which is the pattern worth naming:

- **The classification is evidenced, not asserted.** The opcode streams show exactly what compile-back does — introduces `stloc`/`ldloc` locals around the conditional value before the `stind.i4` — and what it doesn't do: the store target still goes through the preserved original address. Roslyn manufacturing temporaries is a canonical opcode residual, not an address-semantics change.
- **Checkability is pinned in both views.** The new `PointerStoreUsesOriginalAddress_StaysCompileBackCheckable` gates (sugared and lowered) require the fixture to stay `Exact` or `OpcodeDiff` forever — if it ever degrades to `RecompileFail` or drops out of evaluation, the gate fires with the opcode streams in the message. That's the silent-`RecompileFail` risk closed: the docket entry can't quietly stop checking anything.
- **The semantic property is pinned independently of the opcode question.** The address-preservation shape (`*S_` present, `*ptr =` absent) now has a lowered-view gate alongside the existing raised one, so the thing the fixture actually exists to protect is asserted directly, in both pipelines, regardless of how the compile-back diff evolves.
- **The docket comments carry the why**, with the issue number and the cross-reference to the pinning tests — the next person auditing the docket doesn't have to re-derive the investigation.

The `Assert.True(matches.Count > 0, ...)` guard inside the checkability gates is the quiet load-bearing detail: without it, the fixture disappearing from evaluation entirely would render both gates vacuously green. With it, evaluation coverage itself is pinned.

Nothing to flag. This closes #2644 as a correctly-classified residual and leaves the docket *more* protective than before the entry was added. Endorse.

---

Issue #2646, opened 2026-07-11
RTS stabilization tranche: safety floor, structured closure growth, and corpus parity


richlander:
## Goal

Track the next execution tranche under #2560 after the C# type-printer boundary landed in #2633. This issue coordinates existing implementation issues; it does not replace their acceptance criteria.

The tranche makes RTS safe to use as the fidelity vertical before expanding or retiring the legacy path.

## Ordered work

- [ ] Refresh the current `origin/main` RTS/legacy A/B over the pinned 500-target property-getter population and record the result on #2560 and #2280.
- [ ] Complete #2558: restore the monotonic fidelity floor with explicit fallback provenance and prove `Worse: 0`.
- [ ] Verify #2633 fixed #2629 and #2632 on merged `main`; close them with focused evidence or preserve them with an updated reproducer.
- [ ] Complete #2559: replace localized diagnostic-message scraping with structured closure-growth evidence.
- [ ] Advance #2280: wire RTS into corpus fidelity, broaden the retirement population, and establish a stable parity boss.
- [ ] Run the same-cap corpus comparison and one daily/manual RTS-fidelity snapshot required by #2560.

## Parallel oracle work

#2452 may proceed independently once the shared RTS front half is stable. It should classify fixture output as source-equivalent, valid-but-different, invalid, or unavailable without becoming a prerequisite for the monotonic safety floor.

## Exit criteria

- Current A/B evidence is recorded and reproducible.
- RTS reports `Worse: 0` on the retirement population.
- No legacy `Exact`/`OpcodeDiff` row becomes RTS `RecompileFail`/`ContextFail`.
- Closure growth no longer depends on localized diagnostic text.
- Corpus fidelity can execute through RTS with checked denominators and stable failure buckets.
- #2560 has enough current evidence to decide whether the legacy vertical can be retired.

Related: #2560, #2558, #2559, #2280, #2452, #2629, #2632.

richlander:
Current-main kickoff evidence (`f30d3611`):

```text
RETURNTOSENDER A/B over 500 property getters

  Rescued       : 1
  Same          : 499
  Changed       : 0
  Worse         : 0
  CurrentMissing: 0
```

[...]

Issue #2560, opened 2026-07-08
RTS goes live: replace the compile-back vertical (whole-module skeleton + CB_CLUSTER fidelity oracle)


richlander:
## Goal

Make **ReturnToSender (RTS)** the compile-back / fidelity vertical, and retire
compile-back's **architecture** — the whole-module string skeleton plus the
`CB_CLUSTER` escalation used as the corpus fidelity oracle. This is an
architecture replacement, not a rewrite of everything Roslyn-backed: the useful
*components* of compile-back are explicitly **reused**, only the vertical
reconstruction strategy is retired.

This issue is the single go-live tracker. It coordinates the existing RTS
trackers (#2049 ladder, #2050 Rubicon, #2280 parity burndown) into one ordered
path and adds the reused-loop shape fixes surfaced by code review.

## Scope

**Retire (the vertical):** generic whole-module skeleton emission as the
reconstruction strategy, and `CB_CLUSTER` as the standalone fidelity oracle,
once RTS is at parity.

**Reuse (the components), do not retire:**

- the compiler-as-oracle closure-**membership** loop (grow same-assembly roots
  from Roslyn missing-symbol diagnostics) — *"Do not replace compiler-driven
  closure membership with speculative static closure"*;
- the whole-module / generic cluster path itself, retained as the monotonic
  **fallback floor** during transition (see #2558);
- opcode canonicalization and the IL/body diff primitives;
- the existing `Exact` / `OpcodeDiff` / `RecompileFail` / `ContextFail` status
  buckets.

**Out of scope:** the other Roslyn-backed harness oracles — `--validity-check`,
`--type-check`, `--bind-check`, and the generated-fixture shape oracle — stay as
they are. This tracker is only about the compile-back/fidelity vertical.

## Why it is not done yet (current gate)

Committed A/B (#2280, measured at `d414f390`):

```text
RETURNTOSENDER A/B over 500 property getters
  Rescued: 0   Same: 244   Changed: 0   Worse: 256   CurrentMissing: 0
```

RTS currently **regresses** coverage: it rescues nothing, is worse on 256/500
property getters, and loses 5 previously-`Exact` rows (dominant causes:
`CS0641` attribute base chain, `CS0266` enum-constant coercion). Parity — not new
capability — is the front of this work.

## Design note: what makes compile-back capable, and how RTS keeps it

Compile-back's capability is three mechanisms:

1. **compiler-as-oracle closure loop** — Roslyn computes the exact transitive
[...]

PR #2652, opened 2026-07-11
Use structured diagnostics for RTS closure growth

### LINKED ISSUE #2559: RTS closure loop: grow membership from structured diagnostics, not localized message scraping
## Problem

The compiler-driven closure-membership loop grows roots by parsing the
**human-readable diagnostic message text**: `AddClosureRoots` reads
`diagnostic.GetMessage()` and extracts quoted identifiers via `QuotedNames`
(`tools/DecompilerHarness/ReturnToSender.cs:1768`, `1922`).

This is locale- and format-fragile: it assumes the English
`'Foo' does not exist in the namespace 'Bar'` phrasing, and any message-format
change silently stalls the closure (growth returns `false` → `closure-stalled`
→ `RecompileFail`) with no signal that the *parse*, not the *membership*,
failed. The weakness is inherited from legacy compile-back, so fixing it hardens
both paths.

## Fix

Drive membership growth from Roslyn's **structured** diagnostic data rather than
message text:

- use `Diagnostic.Descriptor.Id` + structured `Arguments`, and/or resolve the
  unresolved symbols through the `SemanticModel`, instead of scraping quoted
  spans from the localized string;
- keep the same diagnostic-code coverage the loop handles today
  (`CS0246`/`CS0234`/`CS0103`/`CS0122` for types, `CS1061`/`CS0117` for
  members) but source the names structurally.

## Done

- Membership growth no longer depends on English message phrasing.
- Closure-growth tests cover the type and member diagnostic families via
  structured data, including the `CS0234` namespace-segment case.
- No behavior regression in the RTS A/B checked population.

Part of the RTS compile-back vertical-replacement umbrella (linked as parent).
Related: #2280, #2050.

richlander:
- Advances #2646 and #2560
- Closes #2559
- Changes harness-only closure membership; product C# output is unchanged
- Evidence revision: `bea0570d`

## Change

**Conclusion:** **PASS** — RTS and legacy `CB_CLUSTER` now derive closure
membership from structured Roslyn syntax/semantic evidence instead of localized
diagnostic message text, with the 500-target parity population unchanged at
`Worse: 0`.

Before, both loops parsed quoted spans from `Diagnostic.GetMessage()`. A locale
or message-format change could silently turn a resolvable closure into
`closure-stalled`.

After, one shared extractor uses diagnostic IDs, source locations, syntax,
semantic receiver types, and inaccessible candidate symbols for
`CS0246`/`CS0234`/`CS0103`/`CS0122`/`CS1061`/`CS0117`.
Supported extraction misses are now explicit
`closure-stalled-unextracted[CS####,...]` evidence rather than ordinary silent
stalls.

## Scope and safety

- **Root cause:** closure growth treated human-readable compiler prose as a
  structured membership contract.
- **Fix shape:** shared tools-only evidence extraction consumed by both RTS and
  the retained cluster compatibility path.
- **Product impact:** none; no shipped product path or rendered C# changed.
- **Identity handling:** raw metadata names feed simple/namespace indexes;
  escaped C# identities feed full-type indexes, including contextual-keyword
  namespace segments such as `await`.
- **Syntax coverage:** qualified namespaces, widened expression diagnostics,
  inaccessible candidates, nested and generic receiver identities, keyword
  identifiers, null-conditional member binding, object/collection initializers,
  and await patterns.
- **Implicit member evidence:** Roslyn await/collection symbol information
  distinguishes genuine `GetAwaiter`/`Add` gaps from explicit missing members
  nested inside the same syntax.
- **RTS boundary:** RTS still owns closure orchestration; product/shared code
  still owns C# artifact rendering.
- **Miss provenance:** RTS retains the reason in result `Detail`; legacy cluster
  fallback retains it in structured `CaptureDetail` and summary output.

## Evidence

| Check | Result |
| --- | --- |
| Product/test solution build | Pass, 0 errors |
| Structured diagnostic tests | 32/32 passed |
| RTS + forced-cluster integration | 3/3 passed |
| ReturnToSender A/B | 1 rescued, 499 same, 0 changed, 0 worse |
| Markdown lint | Pass |
| Adversarial review | Opus CLEAN; Gemini CLEAN on exact head `bea0570d` |

The full Decompiler executable ran 2,375 tests: 2,371 passed and four failed.
All four reproduce identically on a clean detached `origin/main` at `f30d3611`:

- `LockSugarTests.CoreLibStaticFieldLock_PreservesCopiedReceiverLocal`
- `TypeSourceComposerUnionTests.ClassUnionSwitchExpression_RendersTypePatternArms`
- raised and lowered `NoNewOpcodeDiffsBeyondKnownDocket`
  (`PointerStoreUsesOriginalAddress`)

### ReturnToSender A/B

Fixed corpus from `eng/prepare-decompiler-corpus.sh`, cap 500:

```text
RETURNTOSENDER A/B over 500 property getters

  Rescued       : 1
  Same          : 499
  Changed       : 0
  Worse         : 0
  CurrentMissing: 0
```

The rescue remains
`Newtonsoft.Json.Converters.XmlDocumentWrapper::get_DocumentElement`:
legacy `RecompileFail CS0266` to RTS `Exact`.

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | CLEAN | Exact head `bea0570d`; traced all RTS/legacy propagation paths and ran 32 structured-evidence tests |
| Gemini 3.1 Pro | CLEAN | Exact head `bea0570d`; verified reason isolation, capture provenance, escaping parity, and internal API compatibility |

First-round findings were reproduced and resolved in `6a1756b7`:

- widened-span expression `CS0234`;
- object/collection initializer and await implicit members.

Second-round explicit-vs-implicit ambiguity was resolved in `ff669f8b` using
Roslyn await/collection symbol information and nearest-context ordering. The
attributed dispositions and non-actions are recorded in
[round one](https://github.com/richlander/dotnet-inspect/pull/2652#issuecomment-4945065647)
and
[round two](https://github.com/richlander/dotnet-inspect/pull/2652#issuecomment-4945214527).

Owner feedback on extractor-miss observability and named test gaps was resolved
in `bea0570d`. It adds durable miss provenance, direct CS0122 member-kind and
null-path coverage, nested/generic receiver coverage, and shared namespace
escaping. Follow-up #2657 tracks remaining `ValidityCheck` message-text
decisions. Follow-up #2658 tracks receiver-compatible narrowing for the legacy
static extension-method index.

**Review conclusion:** **PASS** — both fresh isolated reviewers returned CLEAN
on the exact owner-feedback-fix head with no outstanding findings.


richlander:
First-round adversarial review disposition on head `fdaf19f9`:

- **Claude Opus 4.8:** found widened-span `CS0234` diagnostics in expression/static-member syntax selected the leftmost namespace identifier instead of the rightmost missing segment.
- **Gemini 3.1 Pro:** found `CS0117` object-initializer members and implicit `CS1061` `Add`/`GetAwaiter` diagnostics had no ordinary member-access receiver and were dropped.

All findings were reproduced as four failing real-Roslyn regression rows, then resolved in `6a1756b7`:

- diagnostic-specific rightmost-name selection plus member-access namespace reconstruction;
- object/collection initializer type recovery;
- implicit `Add` and `GetAwaiter` evidence;
- 18 structured diagnostic rows now pass.

Post-fix solution build succeeds and the fixed 500-target A/B remains **Rescued 1 / Same 499 / Changed 0 / Worse 0 / CurrentMissing 0**. Fresh isolated exact-head re-reviews are starting now; this PR is not ready to merge until both are CLEAN.

richlander:
Second-round review disposition:

- **Claude Opus 4.8** found implicit `Add`/`GetAwaiter` synthesis could hijack an explicit missing member nested inside `await` or a collection initializer.
- **Gemini 3.1 Pro** independently reported the same explicit-vs-implicit ambiguity on the superseded `6a1756b7` head.
- A follow-up adversarial probe found the converse control: a valid explicit member can be the element/operand of a genuinely missing `Add`/`GetAwaiter`, and a valid outer `await` can enclose a missing-`Add` initializer.

Resolved in `ff669f8b` using Roslyn semantic evidence:

- unresolved explicit member symbols take precedence;
- `GetAwaitExpressionInfo(...).GetAwaiterMethod` identifies genuine await-pattern failures;
- `GetCollectionInitializerSymbolInfo(...)` identifies genuine missing `Add` methods;
- nearest syntax context breaks contaminated-binding ties.

The structured matrix is now 24/24, including both positive and inverse await/collection controls. Build passes and the fixed 500-target A/B remains **Rescued 1 / Same 499 / Changed 0 / Worse 0 / CurrentMissing 0**. Fresh exact-head re-reviews are starting on `ff669f8b`; the PR remains not ready to merge pending two CLEAN results.

richlander:
Final exact-head review and merge-gate status for `ff669f8b`:

- **Claude Opus 4.8: CLEAN** — verified a clean worktree and exercised namespace, initializer, await, ambiguity, and iterative closure shapes with real Roslyn runs.
- **Gemini 3.1 Pro: CLEAN** — verified a clean worktree, explicit/implicit precedence, contaminated-binding iteration, identity handling, and no branch-only test regressions.
- First- and second-round findings are resolved by `6a1756b7` and `ff669f8b`; no outstanding review findings remain.
- Solution build passes; structured diagnostic matrix is 24/24; markdownlint passes.
- Fixed 500-target A/B: **Rescued 1 / Same 499 / Changed 0 / Worse 0 / CurrentMissing 0**.
- CI is green, GitHub reports CLEAN/MERGEABLE, and the branch remains based on current `origin/main` (`f30d3611`).

**Ready to merge.**

richlander:
## Review — the right cure for message-text coupling; one observability ask and three named test gaps

The core replacement is exactly what the root-cause statement demands. The old loops treated localized compiler prose as a membership contract; the new extractor is anchored on things Roslyn actually guarantees: diagnostic IDs, source spans (`Location.IsInSource` + `FindNode` + span-intersecting `SimpleNameSyntax`), syntax shape, and bound/candidate symbols. The per-ID split is principled — CS0246/CS0103/CS0234 stay pure syntax; CS1061/CS0117/CS0122 use the semantic model where the syntax genuinely can't answer (receiver types via `GetTypeInfo` with the `GetSymbolInfo` fallback that makes *static* receivers work; CS0122 reading the actual inaccessible symbol from `CandidateSymbols` rather than guessing from text). And the `preferRightmost` selection for member/segment diagnostics vs leftmost for type/name diagnostics is a small decision that quietly carries the whole span-anchoring approach.

**The best piece is the explicit-vs-implicit disambiguation.** `IsMissingExplicitMemberReference` doesn't just check syntax position — it re-checks `GetSymbolInfo(simpleName).Symbol is null`, so `new Receiver { item.Value }` (where `item.Value` binds fine and the real failure is the pattern-based `Add`) routes to implicit evidence, while `new() { value.Missing }` routes to the explicit member. The implicit side is then *semantic*, not name-guessed: `GetAwaitExpressionInfo(...).GetAwaiterMethod is null` and `GetCollectionInitializerSymbolInfo(...).Symbol is null`, with nesting depth arbitrating when both enclose the span — and both nestings are tested. That's the difference between detecting a missing pattern member and pattern-matching on its name.

The A/B is the right harness-level proof (500-target parity, 1 rescued, 0 worse — the rescue itself demonstrating the structured path *grows* where text parsing stalled), and `QuotedNames` is genuinely deleted from both loops with no message-parsing fallback left behind.

**One ask I'd make before merge — make extractor misses observable.** Both loops do `if (reference is null) continue;`, and an iteration that grows nothing exits as `closure-stalled`. That means every future extractor blind spot — a non-source-located diagnostic, a span the simple-name heuristics miss, a compound expression where `preferRightmost` picks the wrong name — converts to the *same* `closure-stalled` outcome as a genuinely unresolvable closure, with nothing distinguishing them. The PR's own framing of the old bug was "a format change could **silently** turn a resolvable closure into closure-stalled"; the structured path is enormously less likely to miss, but when it does, it fails silently in the identical way. Cheap fix: count extraction nulls per loop and split the stall reason (`closure-stalled` vs `closure-stalled-unextracted` or a diagnostic-ID breakdown in the reason string). That turns the next blind spot into a greppable signal instead of a parity mystery.

**Named gaps, all small:**

1. **`InaccessibleReference`'s member-symbol branches are untested.** The CS0122 test covers only the `INamedTypeSymbol` branch; the `IMethodSymbol`/`IPropertySymbol`/`IFieldSymbol`/`IEventSymbol` arms (member name + escaped containing type) have zero coverage. Same for nested-type and generic *receivers* driving `TypeName` through the symbol path — the existing coverage is string-level `NormalizeTypeName` only. And there are no negative tests at all for the null-return paths, which per the above are exactly the paths that matter.
2. **Escaper divergence: `await`.** `TypeName` builds namespace text with Roslyn's `ToDisplayString()`, but the `FullTypes` index keys go through `CSharpNaming.SafeIdentifier`, which escapes `await` (not a C# reserved keyword, so Roslyn won't). A namespace segment named `await` produces `await.Foo` from the extractor vs `@await.Foo` in the index — a `FullTypes` miss, mitigated but not fixed by the leaf fallback into the raw `Types` index. Any future divergence between the two escapers has the same failure mode; worth either a shared escaping helper or a comment pinning the assumption next to `TypeName`.
3. **The FidelityCheck CS1061 arm looks up the member name module-wide** (it discards `ContainingType`), so an implicit-evidence miss injects `GetAwaiter`/`Add` into an index where *any* static class with an `Add` pulls in. Bounded by `maxRoots`/`maxIterations`, so it can't run away — but it can burn the root budget and flip a fixable row to `closure-root-budget`. The RTS side uses the receiver type; the cluster side deserves the same narrowing, even as a follow-up.

Out-of-scope but worth a tracking note since this PR's thesis is "no decisions on message text": `ValidityCheck.cs` still content-matches `GetMessage().Contains("__Shell")` — the last locale-sensitive decision point in the harness family.

Endorse — harness-only, A/B-proven, and the extractor is well-designed where it counts. The observability ask is the only thing I'd want in before this merges, precisely because this PR's success will make the next silent stall *look* like a real one.

---

richlander:
Addressed the owner feedback in `bea0570d`:

- supported structured-evidence misses now surface as `closure-stalled-unextracted[CS####,...]` instead of ordinary stalls; RTS stores this in `Detail`, while legacy cluster fallback preserves durable `CaptureDetail` and reports it;
- added direct real-Roslyn coverage for CS0122 method/property/field/event candidates, nested/generic receiver identities, non-source null extraction, deterministic miss reasons, and the contextual namespace segment `await`;
- namespace symbol identities now use raw segments through the same `CompileBackCSharpNames.EscapeNamespace` path as `FullTypes`;
- kept the legacy CS1061 static-class lookup broad because its job is locating extension-method declaring roots rather than receiver roots; receiver-compatible signature narrowing is tracked in #2658;
- remaining `ValidityCheck` message-text decisions are tracked in #2657.

Evidence on this head: focused diagnostic tests 32/32, cluster integration 1/1 (33 total in the combined run), solution build clean, markdownlint clean, and the pinned 500-target A/B remains 1 rescued / 499 same / 0 worse. Fresh isolated Opus and Gemini exact-head reviews are running; the PR is not being marked ready until both are CLEAN.

richlander:
## Final exact-head adversarial review — CLEAN / CLEAN

Both isolated reviewers examined `bea0570dd9bb1ce47e0543c84e75192b378d26d1` after the owner-feedback fix.

- **Claude Opus 4.8: CLEAN.** Traced all three RTS loops and both legacy ForceAll/Escalate paths, verified current-iteration-only miss attribution, supported-ID fencing, reason distinctness, `CaptureDetail` preservation, and namespace escaping parity; ran all 32 structured diagnostic tests successfully.
- **Gemini 3.1 Pro: CLEAN.** Independently verified the same propagation and provenance contracts, deterministic ID output, contextual-keyword identity parity, explicit/implicit member behavior, and internal record compatibility; ran the 32 focused tests successfully.

The reviewers agreed with no divergent or outstanding findings. No additional resolution commit applies because both reviewed the fixed exact head and returned CLEAN. Owner feedback is resolved by `bea0570d`; the intentionally separate follow-ups are #2657 (`ValidityCheck` message-text decisions) and #2658 (receiver-compatible legacy extension-root narrowing).

Local evidence remains 32/32 focused diagnostics, forced-cluster integration clean, solution build clean, markdownlint clean, and 500-target A/B at 1 rescued / 499 same / 0 worse.

richlander:
Ready to merge

richlander:
## Follow-up @ `bea0570d` — all three review items landed

- **Extraction misses are now observable**: `FailureReason` yields `closure-stalled-unextracted[CS0117,CS1061]` (sorted, deduped, test-pinned), and the census output groups and counts the unextracted-stall rows. The next extractor blind spot is now a greppable signal instead of a parity mystery — exactly the ask.
- **The CS0122 member-symbol branches got their tests**: method, property, field, and event candidates, each pinning member name + escaped containing type.
- **The `await` escaper divergence is fixed at the source**: `TypeName` now routes namespace text through `CompileBackCSharpNames.EscapeNamespace` instead of Roslyn's `ToDisplayString()`, so the extractor and the `FullTypes` index share one escaping authority rather than two that happened to agree.

The FidelityCheck CS1061 module-wide member lookup (receiver type discarded on the cluster side) remains as the known over-inclusion, budget-bounded — fine as the named follow-up. Endorsement stands, now with the observability condition met.

---

PR #2647, opened 2026-07-11
Add RTS artifact request envelope

richlander:
- Advances #2529
- Changes harness-only request envelope around RTS compile-back composition; product output is unchanged
- Evidence revision: `f6782105`

## Change

**Conclusion:** **PASS** — this adds the first tools-only `ArtifactRequest` / `ProductArtifact` envelope and routes getter, setter, and method source composition through one request-shaped entry without changing closure derivation, Roslyn oracle growth, or rendering policy.

Before:

```text
RTS called ComposePropertyGetter/ComposePropertySetter/ComposeMethod directly with repeated positional argument lists.
```

After:

```text
RTS constructs ArtifactRequest and calls CompileBackSourceComposer.Compose(request), which wraps the existing compose overloads and returns ProductArtifact.
```

## Scope and safety

- **Root cause:** #2529 still lacked the formal request/result boundary after the Research pass-through shim was removed.
- **Fix shape:** Tools-only orchestration envelope; behavior-preserving wrapper around existing source composition.
- **Product impact:** None intended; this does not change product source rendering, closure seeding, Roslyn compile-back, or IL/opcode comparison.
- **Scope boundary:** Moving closure derivation behind the provider and reducing the three duplicated compile loops are follow-up slices.
- **RTS boundary:** RTS remains the tools-only planner/oracle; Research remains fact/projection-only per current docs.

## Evidence

| Check | Result |
| --- | --- |
| DecompilerHarness build | `dotnet build tools/DecompilerHarness/DecompilerHarness.csproj -c Release --configfile nuget.config --verbosity quiet` passed |
| Focused tests | `dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.ReturnToSenderPrototypeTests` passed: 96/96 |
| ReturnToSender A/B | Not applicable; wrapper-only, no behavior change intended |
| Broad compile-back | Not run; wrapper-only, focused harness tests cover the linked planner path |

## Compile-back quality

**Conclusion:** **PASS** — no checkable population change is claimed; this is an API-boundary slice for future RTS closure/request work.

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| Move closure derivation behind provider | Left as frontier | Requires request mutation/oracle-addition design beyond this wrapper slice |
| Collapse duplicated compile loops | Left as frontier | Safer after the request/result shape lands |
[...]

PR #2655, opened 2026-07-11
Use structural RTS artifact requests

richlander:
- Follows up #2647 feedback
- Changes harness-only artifact request shape; product output is unchanged
- Evidence revision: `fddad52a`

## Change

**Conclusion:** **PASS** — replaces the nullable `TargetProperty` validated slot with structural method/getter/setter request records, so property handles are required by construction for accessor artifacts and absent from method artifacts.

Before:

```text
ArtifactRequest carried ArtifactKind plus PropertyDefinitionHandle? TargetProperty; getter/setter dispatch validated the nullable slot at runtime.
```

After:

```text
MethodArtifactRequest has no property slot; PropertyGetterArtifactRequest and PropertySetterArtifactRequest require TargetProperty in their constructors.
```

## Scope and safety

- **Root cause:** PR #2647 intentionally used the validated-slot pattern for a transitional wrapper; review feedback noted a structural union is the natural end state.
- **Fix shape:** Internal harness-only request variants and type-pattern dispatch in `CompileBackSourceComposer.Compose`.
- **Product impact:** None intended; the underlying `ComposePropertyGetter`, `ComposePropertySetter`, and `ComposeMethod` calls receive the same values as before.
- **Scope boundary:** `ProductArtifact.SourceFacts` remains pre-positioned and unconsumed; assertions can be added when the first consumer lands.

## Evidence

| Check | Result |
| --- | --- |
| DecompilerHarness build | `dotnet build tools/DecompilerHarness/DecompilerHarness.csproj -c Release --configfile nuget.config --verbosity quiet` passed |
| Focused tests | `dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.ReturnToSenderPrototypeTests` passed: 96/96 |

## Review

Adversarial review requested after PR creation; results will be reconciled before marking ready.

### DIFF
--- a/tools/DecompilerHarness/ReturnToSender.cs
+++ b/tools/DecompilerHarness/ReturnToSender.cs
@@ -809,13 +809,13 @@ static class ReturnToSender
 
         for (int iteration = 0; iteration < maxIterations; iteration++)
         {
-            var sourceResult = CompileBackSourceComposer.Compose(new ArtifactRequest(
-                Kind: ArtifactKind.PropertyGetter,
+            var sourceResult = CompileBackSourceComposer.Compose(new PropertyGetterArtifactRequest(
                 AssemblyPath: assemblyPath,
                 Reader: reader,
                 Function: function,
                 TargetType: typeHandle,
                 TargetMethod: getterHandle,
+                TargetProperty: propertyHandle,
                 TargetBody: printed.Output,
                 FullType: fullType,
                 MethodName: methodName,
@@ -823,8 +823,7 @@ static class ReturnToSender
                 SignatureText: CorpusMethodIdentity.SignatureText(function.Signature),
                 ClosureRoots: closureRoots,
                 ClosureFacts: closureFacts,
-                ClosureMemberRequirements: closureMemberRequirements,
-                TargetProperty: propertyHandle));
+                ClosureMemberRequirements: closureMemberRequirements));
             var plan = sourceResult.Plan;
 
             if (plan.Diagnostics.FirstOrDefault(diagnostic => diagnostic.Layer == "type identity") is { } identityDiagnostic)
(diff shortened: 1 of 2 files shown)


richlander:
Adversarial review reconciliation for head `fddad52a`:

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | CLEAN | Verified clean isolated worktree, structural request forwarding, getter/setter/method handle mapping, property propagation, switch dispatch, no stale `ArtifactKind`/`RequiredProperty`, build, and `ReturnToSenderPrototypeTests` 96/96. |
| Gemini 3.1 Pro | CLEAN | Verified positional forwarding through the record hierarchy, handle assignment, `TargetProperty` propagation, record inheritance/equality behavior, switch dispatch, and build/link success. |

No outstanding review findings.

richlander:
Ready to merge.

Merge-blocking status is complete: CI is green, focused local validation passed, and both required adversarial reviews are clean with reconciliation posted above.

richlander:
## Review — the structural union I sketched on #2647, delivered

This is the follow-up my #2647 review asked for: the `ArtifactKind` enum + nullable `TargetProperty` + runtime `RequiredProperty` throw is replaced by a request hierarchy where the property requirement is *structural* — `MethodArtifactRequest` has no property slot at all, and the abstract `PropertyAccessorArtifactRequest` carries a non-nullable `PropertyDefinitionHandle` that both accessor variants inherit. The invalid state (accessor request without a property) is now unrepresentable rather than validated, `RequiredProperty` is deleted, and the `Compose` dispatch switches on request *type* with each accessor arm reading `getter.TargetProperty`/`setter.TargetProperty` directly. Call sites read better too — the property handle now sits with the other target handles instead of trailing as a nullable afterthought.

- The shared 13-parameter positional record repeated across four declarations (base, method, accessor base, two leaves) is the C# tax on positional-record inheritance — verbose but mechanical, and the compiler enforces the threading. If the parameter list grows again, that's the cue to consider a grouped sub-record (e.g., a `ClosureInputs` bundle) rather than a fifth copy.
- The `_ => throw new ArgumentException(...)` arm remains reachable only by a new unmatched subclass of the abstract record — correct belt-and-suspenders given the hierarchy is internal and open.

Wrapper-on-wrapper, still behavior-preserving by inspection (same arguments, same overloads, same six call sites). Endorse.

---

PR #2656, opened 2026-07-11
Fix reader-backed implementation IL diff

### LINKED ISSUE #2635: ImplementationDiff reports metadata-token comparison failures for identical assemblies

richlander:
## Summary

Fix assembly-wide `ImplementationDiff` so metadata-token operands use the
`MetadataReader` instances already owned by Research.

- compare `MethodBodyBlock` values through the reader-backed `IlBodyDiff`
  overload
- preserve fail-closed behavior for genuinely readerless comparisons
- cover exact token-bearing self-diff and symbolic changed-token evidence in
  Research and CLI tests

The issue reproduction now reports:

```text
Summary   No implementation differences detected.
```

Cross-version `CallToken` evidence resolves symbolically from
`System.Math::Abs(int32)` to `System.Math::Sign(int32)`.

## Validation

- `dotnet run --project src/ILInspector.Research.Tests -c Release --no-restore`
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-restore -- -class "*DiffCommandTests*"`
- exact CLI self-diff reproduction from #2635

Fixes #2635

### DIFF
--- a/src/ILInspector.Research.Tests/ResearchDiffTests.cs
+++ b/src/ILInspector.Research.Tests/ResearchDiffTests.cs
@@ -1002,6 +1002,39 @@ public class ResearchDiffTests
     }
 
     [Fact]
+    public void ImplementationDiff_CompareAssemblies_SameTokenBearingAssemblyIsExact()
+    {
+        var path = FixtureCatalog.DiffPair.OldAssemblyPath();
+
+        var diff = ImplementationDiff.CompareAssemblies(
+            path,
+            path,
+            new ImplementationDiffOptions(
+                ImplementationDiffMechanism.IlBody,
+                TypeFilters: new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "DiffSample" }));
+
+        Assert.True(diff.IsEmpty);
+    }
+
+    [Fact]
+    public void ImplementationDiff_CompareAssemblies_ResolvesChangedTokenOperandsSymbolically()
+    {
+        var diff = ImplementationDiff.CompareAssemblies(
+            FixtureCatalog.DiffPair.OldAssemblyPath(),
+            FixtureCatalog.DiffPair.NewAssemblyPath(),
+            new ImplementationDiffOptions(
+                ImplementationDiffMechanism.IlBody,
+                TypeFilters: new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "DiffSample" }));
+
+        var changed = Assert.Single(diff.Members, member => member.Subject.MemberName == "CallToken");
+        Assert.DoesNotContain(changed.Changes, change => change.IlDisplayFailureRow is not null);
+        Assert.Contains(changed.Changes, change =>
+            ImplementationDiff.UnifiedLines(change).Any(line => line.Contains("::Abs(", StringComparison.Ordinal)));
+        Assert.Contains(changed.Changes, change =>
+            ImplementationDiff.UnifiedLines(change).Any(line => line.Contains("::Sign(", StringComparison.Ordinal)));
+    }
+
+    [Fact]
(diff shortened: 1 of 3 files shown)


richlander:

Exact head reviewed: `67018c8d`

- **Claude Opus 4.8 — CLEAN.** Reconstructed the assembly-wide call chain,
  verified the new tests are non-vacuous, and checked reader/body lifetime,
  decode validation, token canonicalization, assembly pairing, malformed-input
  behavior, and the SRM-only/NativeAOT boundary. Focused Research and CLI tests
  passed.
- **Gemini 3.1 Pro — CLEAN.** Independently reconstructed the reader-backed
  comparison path and verified eager symbolic token resolution, preserved
  `MethodInstructions` decode validation, reader disposal safety, unchanged
  pairing/filtering, and no runtime assembly loading. Full Research, Analysis,
  and CLI test sweeps passed.

The reviewers agreed that retaining `MethodBodyBlock` until the
`MetadataReader`-backed `IlBodyDiff.Compare` call fixes the false failure rows
without weakening readerless fail-closed behavior. Neither reviewer reported an
actionable finding, so no resolution commit applies.

richlander:
Ready to merge

richlander:
## Review — closes #2635 at the root, not at the symptom

The #2635 failures existed because IL token operands are version-local: comparing two assemblies' raw token integers is meaningless, so token-bearing instructions produced failure rows ("requires a MetadataReader-backed comparison"). This fix threads both sides' `MetadataReader`s into `IlBodyDiff.Compare` so token operands resolve **symbolically** — the test pins that a `Math.Abs` → `Math.Sign` call-target change renders as `::Abs(`/`::Sign(` evidence lines rather than a failure row.

The two test directions are the right pair:

- **Same assembly vs itself is now `IsExact`** — the previous behavior treated identical token-bearing methods as un-comparable, which is the fail-closed-turned-false-positive case; a self-diff producing anything but empty was the tell that the comparison was operating on the wrong identity space.
- **Changed token operands resolve to symbol names** — provenance (the token) stays version-local while the comparison runs on the resolved identity, which is the same identity-vs-provenance split the Finding producers settled on (#2638/#2643), now applied inside the IL body differ.

The plumbing change — `TryDecode` returning the raw `MethodBodyBlock` and exposing the reader, with decoding moved inside the differ where reader context lives — is the correct ownership move: decoding without the reader was precisely how the token-blindness happened.

CLI-level assertion (`DoesNotContain("requires a MetadataReader-backed comparison")`) confirms the failure rows are gone from the composed view, not just from the unit under test. Small, targeted, and it converts a tracked known-failure into pinned behavior. Endorse.

---

Issue #2653, opened 2026-07-11
DotnetInspector consumer burndown: replace CLI-owned paths with Findings and lower services


richlander:
## Goal

Make `dotnet-inspect` a thinner consumer of typed lower-layer Findings and
services. Retire CLI-owned domain computation and duplicate acquisition paths
without moving presentation policy into producers.

This is a claimable queue from the CLI layering audit at `f30d3611`. Claim one
row by commenting `Claiming C<n>` before starting. A claim should end in a PR,
an explicit blocker, or a comment releasing the row.

## Boundaries

- Metadata owns metadata observations and identity.
- Analysis owns typed method-body observations.
- Research owns genuine joins across producers.
- Services/Packages own reusable acquisition and resolution.
- The CLI owns option interpretation, section selection, ranking for a user
  lens, and rendering.
- A new Finding producer must have at least two real consumers and retire a
  bespoke path. Do not add adapters that only rename an existing query.
- Preserve output and failure visibility unless a row explicitly changes
  behavior.

## Claimable work

- [ ] **C1 — Consume the existing extension-member Finding census.**
  Route both the library Extension Methods section and the `extensions` command
  through `MetadataFindings.InspectExtensionMembers` and one typed Metadata
  inventory/query path. Retain `FindingInspection<ExtensionMemberObservation>`
  until the CLI projection boundary, preserve target/reachability filtering and
  output, surface acquisition failures, and remove the duplicate
  `ExtensionMethodSummary`/scanner projection path.
  **Primary areas:** `LibraryMetadataService`, `LibraryInspection`,
  `LibraryInspectionView`, `ExtensionsCommand`,
  `MetadataFindings.Extensions`.
  **Ready:** yes.

- [ ] **C2 — Add a neutral assembly-set acquisition service.**
  Move the reusable package/platform/project/local assembly collection and
  lifetime contract behind a lower `DotnetInspector.Services`/Packages API.
  Inputs and results must not depend on CLI option, logger, or view types.
  Migrate the current `AssemblyCollector` consumers and make cleanup ownership
  explicit.
  **Primary areas:** `AssemblyCollector`, `FindCommand`, `ImplementsCommand`,
  `ExtensionsCommand`, `TypeFindIfMissResolver`, `CallerScopeResolver`,
  `DependencyGraphService`, `DotnetInspector.Services`.
  **Ready:** yes. Coordinate with #2122.

[...]

PR #2628, opened 2026-07-10
Preserve Finding failures and JSON compatibility

richlander:
## Summary

- retain library scanner acquisition exceptions as typed failed Finding inspections instead of collapsing them to absence
- restore omission of null custom-attribute values and preserve pre-migration custom-attribute/resource JSON element ordering
- cover all migrated Metadata Finding scanners with acquisition-failure and compatibility tests

Follow-up to #2622.

## Evidence

- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config`
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-build -- -class '*LibraryFindingConsumerTests*'` — 8 passed
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-build -- -method- '*CliDiscoverySections_AreSelectable*'` — 1,700 passed, 10 tool-dependent skips
- pre-#2622 versus fixed-head `library ... --json -v:d --references` output is byte-identical for the same `dotnet-inspect.dll`

The excluded discovery matrix case is a self-inspection SourceLink-integrity check: it passes on clean `origin/main` and reports the expected checksum mismatch after tracked source is modified and rebuilt.

PR #2660, opened 2026-07-11
Add neutral assembly set resolver

richlander:
## Summary

Implements #2653 C2 by moving reusable assembly acquisition into a neutral `DotnetInspector.Services` resolver and migrating CLI consumers off `AssemblyCollector`.

## Changes

- Added `AssemblySetResolver` plus neutral request/result/diagnostic/lifetime types in `DotnetInspector.Services`.
- Migrated find/type lookup/find-if-miss, implements, extensions, member caller scope, and dependency graph paths to the owned `AssemblySet` contract.
- Kept CLI option adaptation in `dotnet-inspect` via `AssemblySourceOptionsExtensions`.
- Retired the CLI-owned `AssemblyCollector`.
- Added focused resolver and CLI regression tests for package cleanup ownership, directory diagnostics/source metadata, and dependency graph compatibility.

## Validation

- `dotnet build src/dotnet-inspect/dotnet-inspect.csproj -c Release --configfile nuget.config --nologo -v:q`
- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config --nologo -v:q`
- `dotnet run --project src/DotnetInspector.Services.Tests -c Release --no-restore -- -class DotnetInspector.Services.Tests.AssemblySetResolverTests`
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-restore -- -class DotnetInspector.Tests.TypeSearchServiceTests -class DotnetInspector.Tests.DependencyGraphServiceTests`
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-restore -- -class DotnetInspector.Tests.CommandExecutionTests`
- `npx markdownlint-cli docs/overview.md`
- CLI smokes: `find --bin`, `implements --library`, `extensions --library`, and `depends` type/library modes with `--count`.

## Adversarial review

In progress with isolated worktrees:

- Claude Opus 4.8
- Gemini 3.1 Pro


richlander:
Adversarial review reconciliation for head `264ed0c0`:

- Gemini 3.1 Pro r1: CLEAN. Checked resolver lifetime ownership, diagnostics, layering, migrated consumers, and local build/tests.
- Claude Opus 4.8 r1: Found one compatibility issue: mixed-source `find` collection order changed project assets before platform sources, which could change display and `--limit` results.
- Resolution: pushed `264ed0c0` (`Preserve find assembly source order`), adding `AssemblySetRequest.SourceOrder`, keeping the neutral resolver default aligned with old `AssemblyCollector` consumers, and making `TypeSearchService` request the old `find` order: packages, assemblies, platform assemblies, platform frameworks, projects, directories.
- Gemini 3.1 Pro r2: CLEAN on `264ed0c0`; verified the source-order fix plus lifetime, behavior, layering, diagnostics, migration coverage, and package dependency graph separation.
- Claude Opus 4.8 r2: CLEAN on `264ed0c0`; verified the prior finding is fixed and no remaining correctness/security/compatibility issues were found.

Local validation after the fix:

- `dotnet build src/dotnet-inspect/dotnet-inspect.csproj -c Release --configfile nuget.config --nologo -v:q`
- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config --nologo -v:q`
- `dotnet run --project src/DotnetInspector.Services.Tests -c Release --no-restore -- -class DotnetInspector.Services.Tests.AssemblySetResolverTests`
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-restore -- -class DotnetInspector.Tests.TypeSearchServiceTests -class DotnetInspector.Tests.DependencyGraphServiceTests`
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-restore -- -class DotnetInspector.Tests.CommandExecutionTests`
- `npx markdownlint-cli docs/overview.md`
- CLI smokes for `find --bin`, `implements --library`, `extensions --library`, and `depends` type/library modes with `--count`.

Ready to merge.

richlander:
## Review — the unification earns its place; hold for the four unpinned drifts and the Warning-only diagnostic model

First, what's right. This kills a real two-way duplication: `AssemblyCollector` and TypeSearchService's hand-rolled six-source loop had already drifted from each other, and five consumers convert in this PR (extensions ×2, implements, depends ×3, caller scope, and the find/type/lookup stack through `CollectTypesAsync`) — well past the earns-a-place bar. The layering is genuinely neutral (`DotnetInspector.Services`, `Action<string>` not `VerboseLogger`, diagnostics as data, CLI adapter kept CLI-side). The `SourceOrder` knob is the standout design decision: it *acknowledges* that the two old stacks disagreed on iteration order and preserves find's historical order explicitly instead of silently unifying semantics that were never the same. And the caller-scope conversion fixes a real pre-existing fail-open bug — the old code deleted caller-scope package temp dirs *before* the caller scan read them, silently yielding empty results for `--caller-scope-package`; the new `CallerScopeAssemblySet` lifetime is correct.

But a consolidation PR's one job is to be boring, and this one ships **four unpinned behavior changes** that inherited whichever old stack's semantics the resolver happened to pick, with no tests and no mention:

1. **extensions/implements now pass a platform-framework hint** to `PlatformResolver.ResolveAssemblyAsync` (old `AssemblyCollector` passed none) — they silently adopted find's semantics for `--platform-assembly` + framework.
2. **`find --package` no longer scans `runtimes/**` dlls** when TFM resolution falls back to whole-extract scanning — the reverse direction: find adopted `AssemblyCollector`'s exclusion, and results can shrink.
3. **`depends <package>` can pick a different root assembly** — the old lexicographic `/lib/` sort is replaced by real TFM priority with unrestricted-filesystem-order fallback, and the specific "No libraries found in package" error collapsed into the generic could-not-resolve message. `DependencyGraphServiceTests` has exactly one test; none of this is pinned.
4. **`find --limit` no longer short-circuits resolution** — everything downloads/extracts up front and the limit only gates scanning, so capped runs now do network work (and can emit warnings) for sources the old code never touched; parallel scanning also widened from platform-frameworks-only to all sources.

Some of these may be *improvements* (2 and 3 plausibly are; the caller-scope fix definitely is), but each is a decision, and right now they're accidents of which implementation won. Each needs either a pinning test that asserts the new choice on purpose, or a revert to the old behavior for that consumer — plus a line in the PR body so the change is on the record. The `MemberCommand` caller-scope fix deserves its own test most of all, since it's the one *intentional* behavior change and nothing pins it.

**Second hold: the diagnostic model codifies fail-open.** `AssemblySetDiagnosticSeverity` has exactly one member — `Warning` — so total resolution failure returns a successful empty set, and fail-closed is delegated entirely to consumer discipline. That discipline is already failing inside this very PR: `DependencyGraphService` renders diagnostics only when `Assemblies.Count > 0`, so on the empty platform branch it disposes the set and the resolution error *vanishes* — the exact absence-indistinguishable-from-emptiness shape #2628/#2662 just eliminated elsewhere. And the rendering is five verbatim copies of `Console.Error.WriteLine($"Warning: {diagnostic.Message}")` across five files, so when a second severity arrives, the handling drifts per-consumer. Two fixes: give the enum an `Error` severity now (or an `IsEmpty`+`Diagnostics` contract note stating callers MUST render before treating empty as empty), and give diagnostic rendering a single owner the five call sites share.

Smaller items: `SourceOrder` silently drops sources not listed (a request with packages but `SourceOrder = [Directory]` ignores the packages, no diagnostic) and duplicate kinds resolve twice — guard or document; `TypeSearchService.CollectFromPath` is now dead public code; `DependencyGraphService` re-privatized a copy of `CleanupTempDirs` because `BuildPackageDependencyTreeAsync` wasn't converted — a partial migration worth a TODO naming the follow-up.

**Also: this conflicts with #2662**, which rewrites the same `ExtensionsCommand` scan path this PR converts. #2662 is clean and ready; I'd land it first and rebase this on top, reconciling the census-based scan with the resolver-based collection.

Verdict: right architecture, right ownership, real bug fixed — but hold until the four drifts are pinned-or-reverted and the diagnostic contract can't silently eat a total failure. As it stands, the PR trades two visible duplicated stacks for one neutral stack with invisible seams.

---

richlander:
Final reconciliation for head `e2ec3731`:

- Follow-up feedback was addressed in `fde18794`:
  - pinned consumer-specific source behavior for platform hints, package runtime inclusion, `depends` package root selection, and capped `find --limit` resolution;
  - added shared `AssemblySetDiagnosticWriter` plus `Error` diagnostics;
  - preserved #2662 extension finding/census consumption while moving acquisition to `AssemblySetResolver`;
  - added focused tests for resolver source-order guards, runtime package inclusion, `depends` package roots/no-lib errors, capped `find`, and caller-package lifetime.
- Final Opus finding was addressed in `e2ec3731`: speculative platform misses in `depends <System.*|Microsoft.* package>` no longer emit a spurious warning before package fallback.
- Gemini 3.1 Pro final review: CLEAN on `e2ec3731`.
- Claude Opus 4.8 final review: CLEAN on `e2ec3731`.

Validation after feedback fixes:

- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config --nologo -v:q`
- `dotnet run --project src/DotnetInspector.Services.Tests -c Release --no-restore -- -class DotnetInspector.Services.Tests.AssemblySetResolverTests`
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-restore -- -class DotnetInspector.Tests.TypeSearchServiceTests -class DotnetInspector.Tests.DependencyGraphServiceTests -class DotnetInspector.Tests.CallerScopeResolverTests -class DotnetInspector.Tests.ExtensionsCommandTests`
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-restore -- -class DotnetInspector.Tests.CommandExecutionTests`
- `npx markdownlint-cli docs/overview.md`
- CLI smokes for `find --bin`, `implements --library`, `extensions --library`, and `depends` type/library modes with `--count`.

CI is green. Ready to merge.

richlander:
## Follow-up @ `e2ec3731` — both holds released

Verified `fde1879` against the hold list:

- **All four drifts are now pinned as explicit, tested decisions instead of accidents.** The platform-framework hint became an opt-in (`PlatformAssemblyFrameworkHint` on the request — old `AssemblyCollector` consumers don't set it, find does); `runtimes/**` inclusion/exclusion is consumer-pinned with zip-fixture tests on both sides; `depends`' historical root selection survives as an explicit `AssemblySetPackageSelectionMode` (lib-filtered descending order pinned by test, and the specific "No libraries found in package" error restored — now as an `Error`-severity diagnostic with its own test); and capped `find` short-circuiting is restored.
- **The diagnostic model is no longer structurally fail-open.** The severity enum gained `Error`, and rendering has a single owner (`AssemblySetDiagnosticWriter`) with severity-aware prefixes replacing the five copy-pasted `Console.Error` lines — the two structural fixes I asked for. The caller-package lifetime fix also got its focused tests (`CallerScopeResolverTests`), so the one intentional behavior change is pinned.
- `e2ec373`'s `includeErrors: false` for the speculative `System.*`/`Microsoft.*` platform probe in `depends` is the right scoping — a probe that's *expected* to miss shouldn't print an error before the package fallback runs, and that matches the old auto-detection behavior rather than reintroducing the silent-drop (the fallback path's own failures still render).

Hold released; endorse. One reminder from the original review: this still overlaps #2662's `ExtensionsCommand` rewrite — #2662 merged first as suggested, so just make sure the census-based scan and the resolver-based collection reconciled cleanly in the rebase (the extensions tests updated in `fde1879` suggest they did).

---

PR #2661, opened 2026-07-11
Expose allocation Finding transitions

richlander:
## Summary

- add `diff --finding analysis.allocation` as the member-scoped final-pair confirmation for allocation onset
- project Analysis-owned `PairFinding<AllocationOccurrence>` values as `Added`, `Present`, `Removed`, or `Changed` without inferring onset from aggregate counts
- retain exact allocation comparisons only when the focused Finding lens requests them; ordinary `Analysis Diff` behavior and cost remain unchanged
- preserve the existing API Finding transition defaults and public Research constructor signatures

The caller still owns version enumeration and probing. This command compares only the supplied endpoints and requires exactly one resolved method target.

## Evidence

The V1/V2 allocation fixtures produce native occurrence transitions:

- `RegressesAllocInLoop`: `PairFinding.Added` plus a changed existing occurrence
- `ImprovesAlloc`: `PairFinding.Present` plus two `PairFinding.Removed` occurrences
- `SameAllocationCountBecomesHot`: `PairFinding.Changed` with loop/path/multiplicity facet detail
- `Stable`: successful empty censuses and no manufactured pair

## Validation

- `dotnet run --project src/ILInspector.Analysis.Tests -c Release --no-build` — 416 passed
- `dotnet run --project src/ILInspector.Research.Tests -c Release --no-build` — 83 passed
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-build` — 1,748 passed, 10 environment skips

## Adversarial review

- Claude Opus 4.8 found that adding optional positional parameters removed existing CLR constructor signatures. Commit `e6903d9c` restored the original signatures, moved retention to an init property, guarded default immutable arrays, and added reflection coverage. Opus rereviewed the corrected head: **CLEAN**.
- Gemini 3.1 Pro reviewed the original implementation and corrected head: **CLEAN**. It independently exercised descriptor/section errors, native allocation pair projection, exact comparison retention, and API fallback behavior.


richlander:
Ready to merge. Merge-blocking CI is green, and the corrected head has two clean adversarial rereviews (Claude Opus 4.8 and Gemini 3.1 Pro).

richlander:
## Review — the confirmation lens generalizes correctly to its second producer; coordinate the #2650 conflict

#2642 built Finding Transitions as an API-pair confirmation lens; this PR proves the design generalizes: `--finding analysis.allocation` selects Analysis-owned allocation pairs through the *same* section, same row schema, same native-union-case reporting. That's the arity the lens was designed for — descriptor-addressed, producer-owned pairs — and getting a second producer through it without a schema change is the evidence the abstraction was right.

The mechanics hold up:

- **Descriptor validation is fail-closed and enumerates its support**: unknown `--finding` values error with the supported list (`api.type`, `api.member`, `analysis.allocation`); the default inference (member filter present → `api.member`, else `api.type`) preserves #2642's behavior exactly; and each descriptor gets its own target-shape validation (allocation requires exactly one `--member`, `api.type` rejects `--member`, `api.member` requires it) — all before acquisition, keeping #2642's validate-before-work discipline.
- **`RetainAllocationComparisons` is the right mechanism for Present.** The #2638 lane only attaches comparisons to *change rows*, so an exact comparison — which produces no change row — would be invisible, and a confirmation lens must be able to say `PairFinding.Present`. Retaining the full comparison envelope (including `Exact`) on an opt-in flag, stored as a separate `AllocationFindingComparison` list on `ResearchComparison`, gets Present without bloating the default path. The nomenclature-doc addition says this out loud ("retains exact as well as changed allocation comparisons … so it can report `PairFinding.Present`, not only candidate boundaries"), which is exactly the sentence a future reader needs.
- **Target resolution reuses the resolver with `bodySectionName: "Finding Transitions"`** — the method-like-target rejection stays consistent across all three body-level lenses; no parallel resolution path.
- Allocation row targets are built from the member subject plus the producer's canonical allocation identity (same tiered fallback as `GetAllocationIdentityKey`), so the row's `Target` column speaks the producer's identity language rather than a display invention.

Two flags:

1. **This conflicts with #2650 in `DiffCommand.cs`** — both PRs rewrite the Finding Transitions guard block (this one adds the `--finding` selector logic; #2650's merge reworded the guard message and added the composition dispatch just above it). Mechanical to reconcile, but whichever lands second should make sure the `--finding`-implies-FT selection (`SelectsFindingTransitions` now firing on `options.Finding is not null`) is checked *before* #2650's `WriteSelectedDocument` multi-section dispatch — otherwise `--finding` plus an explicit `-S` combination could route into composition instead of erroring.
2. Minor: in the dispatch, `TryResolveFindingDescriptor(options, out var findingDescriptor, out _)` discards the return value — safe because validation already ran in `ExecuteAsync`, but a one-line comment (or `Debug.Assert`) saying so would keep someone from "handling" it later and diverging the two call sites.

Also noting the small duplication now standing: allocation comparisons reach consumers two ways — change-attached (`ResearchChange.AllocationComparison`, non-exact only) and the new opt-in retained list (all outcomes). The rationale is sound (Present has no change row to ride), but when a third consumer arrives it's worth deciding whether the change-attached copy still earns its slot.

Endorse. The docs/skills updates (nomenclature, version-resolution, compatibility + performance skills) keep all tellings in sync with the code, which has become one of this repo's better habits.

---

richlander:
Processed the review feedback in `e0c00d9e` after merging current `origin/main` / #2650.

- **#2650 conflict:** resolved. `--finding` now implies the standalone `Finding Transitions` section before discovery/output validation; focused-lens validation still runs before acquisition and before composed-document dispatch. Both standalone rendering and #2650's JSON document path use the same API/allocation dispatcher.
- **Regression coverage:** added `--finding analysis.allocation --json` coverage proving the document contains `finding_transitions`, not default `changes`, plus a missing-library test proving `--finding` with `-S \"Analysis Diff\"` is rejected before acquisition.
- **Discarded resolver result:** addressed by replacing the unchecked second call with a defensive `ResolveFindingDescriptor` helper; dispatch cannot silently continue after failed resolution.
- **Two allocation-comparison access paths:** no change. `ResearchChange.AllocationComparison` remains the native payload for non-exact change consumers, while the opt-in retained list is required for exact `Present` comparisons that have no change row. No resolution commit applies beyond documenting and testing that distinction in this PR.

Post-merge adversarial rereviews: **CLEAN** from Claude Opus 4.8 and Gemini 3.1 Pro. Full local CLI (1,764 passed / 10 environment skips), Research (85), and Analysis (430) suites pass. The prior `Ready to merge` status is superseded while refreshed CI runs.

richlander:
Ready to merge. The #2650 conflict and review flags are resolved in e0c00d9e; refreshed CI is green, and Claude Opus 4.8 plus Gemini 3.1 Pro both returned CLEAN on the merged head.

richlander:
## Post-merge check @ `e0c00d9e` — the ordering hazard I flagged is now live; one small fix before merge

The rebase onto composed-sections main landed mostly clean, but the interaction I warned about materialized in one specific shape: **`--json` now intercepts Finding Transitions requests.**

In `ExecuteAsync`, the composition dispatch runs before the transitions dispatch:

```csharp
if (options.JsonOutput || options.IncludeSections is { Count: > 1 })
{
    WriteSelectedDocument(inputs, options);
    return 0;
}

if (SelectsFindingTransitions(options))
{ ... }
```

Two silent misroutes follow, both new since #2650's merge made `--json` a registered diff option:

1. `diff --package X@a..b --finding analysis.allocation --json` — the guard block up top passes (no `-S` sections to trip the selected-by-itself check, and it doesn't examine `JsonOutput`), then `JsonOutput` routes into `WriteSelectedDocument`, which defaults the selection to **Changes** — the `--finding` request is silently ignored and the user gets a composed API-diff JSON document instead of allocation transitions.
2. `diff ... -S "Finding Transitions" --json` — same interception; `WriteSelectedDocument` checks only Changes/Analysis/Implementation membership, finds none, and emits a document with a title, versions, and **no sections at all** — an empty answer with exit 0.

Both are the fail-silent shape (a request quietly answered with something other than what was asked). Two ways to fix, either fine:

- **Error in the FT guard block**: when `SelectsFindingTransitions(options) && options.JsonOutput`, reject with "Finding Transitions does not support --json; use --jsonl or --table" — consistent with the section's existing refuse-loudly posture; or
- **Route FT before the composition dispatch** (move the `SelectsFindingTransitions` dispatch above the `JsonOutput || Count > 1` check), if FT-as-JSON should work — the view already serializes through `WriteProjectedTable` for `--jsonl`, so erroring is probably the cheaper and more honest option.

Either way, one regression test pinning `--finding … --json` (error or transitions output, not a Changes document) closes the class. Everything else about the rebase reconciled correctly — the `--finding` selector logic and #2650's reworded guard coexist, and the select-alone check still fires for section combinations. Endorse once this one dispatch interaction is fixed and pinned.

---

richlander:
This finding does not reproduce on the cited head `e0c00d9e`; no new code
change is needed.

The conflict-resolution commit already includes the two missing pieces:

- `DiffCommand.cs:31` normalizes bare `--finding` to the standalone
  `Finding Transitions` selection before discovery/output validation.
- `DiffCommand.cs:613-615` handles `Finding Transitions` inside
  `WriteSelectedDocument` and calls the same `BuildSelectedFindingTransitions`
  dispatcher used by the non-document path.

Reproduced directly from clean `e0c00d9e`:

- `--finding analysis.allocation --json` emits only `title`, `versions`,
  `finding_transitions_summary`, and `finding_transitions`; the rows include
  `PairFinding.Added` / `analysis.allocation` / `absent -> present`. There is
  no `changes` property.
- `-S "Finding Transitions" --json` emits the same Finding document shape with
  API `PairFinding.Present` rows; it is not empty.

`ExecuteAsync_AllocationFindingTransitions_ComposesJson` pins the first case,
and `ExecuteAsync_FindingTransitions_ComposesJson` pins the second. Both passed
locally and in refreshed CI. This comment appears to have evaluated the
pre-resolution conflict shape rather than the pushed resolution, so no
resolution commit applies. PR #2661 remains `Ready to merge`.

richlander:
## Retraction — my `--json` misroute finding was wrong; hold released

I re-verified against the actual head `e0c00d9e`, and the rebuttal is correct: neither misroute reproduces. My mistake, and it's worth stating exactly where I went wrong so the finding's death is as documented as its birth:

- **Shape 1** (`--finding analysis.allocation -m X --json`): I claimed the `--finding` request would be silently ignored because `WriteSelectedDocument` defaults a null selection to Changes. But `ExecuteAsync` **materializes the selection at entry** — `if (options.Finding is not null && options.IncludeSections is null)` sets `IncludeSections = { "Finding Transitions" }` — so the composed document contains the allocation transitions section, exactly as requested.
- **Shape 2** (`-S "Finding Transitions" --json`): I claimed an empty document because `WriteSelectedDocument` only knew Changes/Analysis/Implementation. At this head it doesn't — the merge added a `selected.Contains(DiffSections.FindingTransitions.Name)` branch that builds the transitions view into the document.

The root of my error: I verified the *dispatch order* with targeted greps (which is as I described — composition before the FT early-return) but pattern-matched `WriteSelectedDocument`'s body and the entry normalization against #2650's pre-merge version instead of re-reading them at this head. The merge commit had already extended both, which is precisely the reconciliation my original review asked for — done, and done more completely than I noticed. The dispatch-order interaction I warned about was real as a hazard; the merge resolved it by making composition FT-aware rather than by reordering, which is arguably the better design (`--json` uniformly means "document," and FT rides in it).

#2661 is clear from my side: endorse, safe to merge. The rest of my review (descriptor validation, `RetainAllocationComparisons` for Present, resolver reuse) stands as written.

---

PR #2662, opened 2026-07-11
Consume extension member Findings in DotnetInspector

richlander:
## Summary

- consume Metadata extension-member Findings in both library inspection and the `extensions` command
- retain display signatures as CLI compatibility data while carrying typed inspection state through projection
- scan each assembly once for direct and reachable queries, with visible per-assembly failure isolation
- preserve successful output and harden generic receivers and duplicate metadata anchors

Addresses #2653 (C1).

## Validation

- solution build
- `dotnet-inspect.Tests`: 1,756 passed, 10 expected skips
- `ILInspector.Metadata.Tests`: 452 passed
- byte-identical successful library/direct/reachable JSON output
- architecture Markdown lint

## Adversarial review

Claude Opus 4.8 and Gemini 3.1 Pro independently reviewed isolated clean worktrees. Initial findings covered mixed-input failure isolation, generic receiver matching, and duplicate stable anchors; commits `4f2aff7a` and `a22f9028` resolve them with regressions. Both reviewers returned CLEAN on final head `d0d83ce7`.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -188,6 +188,11 @@ producers or preserve producer-native structural evidence. Failed censuses stay
 distinct from empty results and render as inspection diagnostics without
 suppressing unaffected sections.
 
+Extension discovery follows the same rule for both `library` and `extensions`:
+each assembly is scanned once into Metadata's extension-member Finding census,
+then CLI target, reachability, overload, and display projections consume that
+shared inventory.
+
 ### Metadata Extraction
 
 ```text
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -3972,6 +3972,40 @@ public class CommandExecutionTests
     }
 
     [Fact]
+    public async Task Extensions_FailedAssemblyWarnsAndPreservesSuccessfulResults()
+    {
+        var corruptPath = Path.Combine(
+            Path.GetTempPath(),
+            $"dotnet-inspect-corrupt-{Guid.NewGuid():N}.dll");
+        try
+        {
+            await File.WriteAllTextAsync(
+                corruptPath,
+                "not a PE image",
+                TestContext.Current.CancellationToken);
+
+            var (exit, output, error) = await RunAppAsync(
+                "extensions",
+                "MetadataReader",
+                "--library", typeof(MetadataFindings).Assembly.Location,
+                "--library", corruptPath,
+                "--count",
+                "--tips", "q");
+
+            Assert.Equal(0, exit);
+            Assert.True(int.Parse(output.Trim()) > 0);
+            Assert.Contains(
+                "Warning: Extension member inspection failed",
+                error,
+                StringComparison.Ordinal);
+        }
+        finally
+        {
+            File.Delete(corruptPath);
+        }
+    }
+
+    [Fact]
     public async Task Extensions_JsonlAfterPackage_RendersJsonlAndDoesNotWarnAboutPackageFlag()
     {
         var (exit, output, error) = await RunAppAsync(
--- a/src/dotnet-inspect.Tests/ExtensionsCommandTests.cs
+++ b/src/dotnet-inspect.Tests/ExtensionsCommandTests.cs
@@ -139,7 +139,7 @@ public class ExtensionsCommandTests
     }
 
     [Fact]
-    public void CommandProjection_ConsumesFindingCensusAndPreservesDisplayShape()
+    public void CommandProjection_ConsumesFindingCensusAndHandlesEdgeCases()
     {
         var assembly = new AssemblyCollector.AssemblyInfo(
             typeof(ExtensionsCommandTests).Assembly.Location,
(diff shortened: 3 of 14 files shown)


richlander:

- **Claude Opus 4.8** initially found that one failed assembly aborted the multi-assembly `extensions` command and discarded valid results. `4f2aff7a` now warns once, skips only that assembly, and retains successful projections; the mixed-input command regression covers this.
- **Gemini 3.1 Pro** initially found generic receiver matching silently returned no results because canonical `` `N<T>`` names were normalized twice. It also verified the failure-isolation correction. `a22f9028` restores matching through the prior display receiver type with generic coverage.
- **Claude Opus 4.8** then found duplicate stable anchors from legal-but-obfuscated metadata could throw in `ToDictionary`. `a22f9028` uses a collision-tolerant lookup and verifies exact typed payload correspondence; duplicate-anchor coverage is included.
- `d0d83ce7` keeps those edge cases in a metadata-token-stable fixture.

Both Claude Opus 4.8 and Gemini 3.1 Pro independently re-reviewed final head `d0d83ce7` from clean isolated worktrees and returned **CLEAN** with no outstanding findings.

richlander:
Ready to merge. Local validation, CI, and two clean fixed-head adversarial reviews are complete.

richlander:
## Review — the extension producer gets its consumer, with a cross-validation move worth copying

This wires #2627's extension-member producer into the CLI the same way #2621→#2622 wired the library sections, and it carries the two disciplines those PRs established — plus one new move I'd like to see become standard.

- **Fail-visible replaces fail-quiet.** The old `ScanForExtensions` swallowed acquisition exceptions into `logger.Log` (invisible unless verbose) and returned an empty list — the exact absence-indistinguishable-from-emptiness pattern #2628 eliminated on the library side. Now `InspectExtensionAssembly` converts the exception into a `Failed` inspection with subject + descriptor + reason, and the command surfaces it on stderr unconditionally while excluding the failed census from results. An unreadable assembly is now a visible warning, not a silently smaller answer.
- **The cross-validation is the standout.** `ProjectExtensions` requires every scanner inventory member to correspond to a census observation — anchor, kind, canonical extended type, return type, assembly, all matched — and throws if the census and inventory diverge. During the migration window where both representations exist, that turns "the Finding census is decorative" into "the Finding census is *checked against* the inventory it will replace, on every run." Divergence means a producer bug, and it fails loudly instead of shipping whichever copy happened to be rendered. When the legacy inventory is eventually deleted, this check has already proven the census carries identical information.
- **The reachable path stops rescanning.** Previously every (assembly × reachable type) pair re-opened and re-scanned the assembly; now one census per assembly is built once and filtered per target with the same `TypeMatcher.Normalize`/`Matches` predicate the scanner used internally — behavior-preserving (same match semantics) with the quadratic file I/O gone.
- **Kind is projected from the validated observation** (`observation.Kind == ExtensionMemberKind.Method ? "method" : "property"`) rather than trusted from the raw string — small, but it means rendered output speaks the census's typed vocabulary.
- The library side follows the established section pattern: `ScanExtensionMembers` populates the inspection on `LibraryInspection`, and "Extension Methods" joins the Library Info failure-section allowlist so acquisition failures surface in the failures table exactly like Resources/Custom Attributes/Type Forwarders do.

One consideration, not a blocker: the correspondence mismatch **throws**, taking the whole command down. For an invariant violation that indicates a producer bug, loud is correct — but if a single assembly in a multi-assembly scan trips it, the user loses results for every other assembly. Converting that specific case to a per-assembly `Failed`-census path (warning + exclusion, same as acquisition failures) would keep the invariant enforcement while containing the blast radius to the offending assembly. Either way, the invariant stays checked.

Test coverage is broad in the right places — command-level warning surfacing, correspondence assertions, the section-pipeline and formatter updates, and `FindingTestData` gaining extension fixtures. Endorse.

---

richlander:
## Proposal — turn this PR's migration approach into the missing design doc

The Finding docs now cover three of the four layers well: `finding-nomenclature.md` (what the types mean), `finding-producers.md` (how to build a producer), `finding-coordinates.md` (identity vs order vs provenance). What has no home is the thing this PR just demonstrated: **how a producer gets adopted** — the consumer-side migration playbook. That knowledge currently lives only in PR threads, and every future consumer migration will re-derive it or get it wrong.

Suggested shape: a new `docs/design/finding-adoption.md` (linked from overview.md alongside the other three), organized as *rules, each anchored to the merged implementation that models it* — so the doc teaches by pointing at real code rather than describing an ideal:

1. **Migration runs dual representations, cross-validated on every execution.** The census is not decorative while the legacy inventory survives — every member must correspond (anchor, kind, canonical identity), and divergence fails loudly as a producer bug. *Model: this PR's `ProjectExtensions` correspondence check.* Corollary rule: the legacy representation is deleted only after the check has run at scale, and the check is what proves the census carries identical information.
2. **Acquisition failure is retained, not collapsed.** `Absent` ≠ `Failed` ≠ successful-empty, end to end, with failures surfaced unconditionally (not verbose-gated) and scoped per-source so one bad assembly doesn't erase the rest. *Models: #2628's `Failed`-inspection retention + failures table; this PR's per-assembly `Failed` census with stderr warning.*
3. **Replacement, not accretion.** The consumer deletes the bespoke lane it replaces in the same PR; no parallel adapter left running. *Models: #2638 (count-only allocation lane deleted), #2640 (Grounding lens replaced, not shadowed).*
4. **Envelopes are consumed fail-visibly.** A `Failed` comparison never renders as "nothing to report" — it either falls back to a visible degraded row or throws with the unreachability invariant stated in a comment. *Models: #2643's `BodySignalDiff` throw-with-invariant; #2641's empty-evidence fallback rows.*
5. **Confirmation lenses report the native case.** CLI surfaces expose `PairFinding.{Kind}` verbatim, refuse classification-filter costumes, and validate targets before acquisition. *Models: #2642, generalized by #2661.*
6. **Sensor and docket integrity ride along.** Quality gates compare sampled *identity sets*, not caps; docket entries ship with checkability pins so they can't rot. *Models: #2649's `HaveSameMethodSample`; #2651's stays-checkable gates.*

Two things make this doc cheap to write now and expensive to skip: the exemplars are all merged and stable (the doc is an index, not new design work), and the next wave of consumers is already visible — classified methods, call sites, unsafety, and extension members each need CLI/Research adoption, and each will face exactly these six decisions. A one-page rules-plus-models doc turns "recognized as well constructed" from a review outcome into a spec the next PR can be written against.

Happy to draft it if useful — the content above is essentially the outline.

---

PR #2663, opened 2026-07-11
Stabilize static lock receiver canary

### LINKED ISSUE #2630: Lock sugar test fails for AppContext.GetData on .NET 11 daily
## Failure

`LockSugarTests.CoreLibStaticFieldLock_PreservesCopiedReceiverLocal` fails against the current .NET 11 daily corelib. The rendered `System.AppContext.GetData` body no longer contains the expected `lock (V_1)`.

## Reproduction

```bash
  -method "*CoreLibStaticFieldLock_PreservesCopiedReceiverLocal*"
```

Observed with SDK `11.0.100-preview.7.26325.14`:

```text
Assert.Contains() Failure: Sub-string not found
Not found: "lock (V_1)"
```

The failure reproduces in a clean worktree at `origin/main` (`11198494c848881121638dcd536c27cd472a8e6d`). It makes the full Decompiler suite red for every agent using the current daily SDK.

richlander:
Closes #2630

## Change

**Conclusion: PASS** — keep the importer-level static-field lock canary while removing its dependency on the mutable implementation of `System.AppContext.GetData`.

The test now compiles a local static-field lock fixture and verifies all three relevant boundaries:

- the importer produces the compiler lowering;
- `LockSugarPass` retains the copied receiver as a `LoadLocal` with its defining store;
- the printer renders `lock (V_n)` rather than inlining the static field.

This is test-only. Product raising and rendering behavior are unchanged, so render A/B and corpus quality cards are not applicable.

## Evidence

| Check | Result |
| --- | --- |
| Product/test solution build | Pass |
| `LockSugarTests` | 4 passed |
| `LockSugarSoundnessTests` | 7 passed |
| Full Decompiler suite before final main rebase | 2370 passed, 1 SDK-capability skip |

The focused lock suites and solution build were rerun after rebasing onto current `origin/main`.

## Review

Adversarial review will run against final head `fb5adaba` in isolated worktrees.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -4184,11 +4184,18 @@ public sealed class CtorChainSamples : CtorChainBase
     public CtorChainSamples(long value) : this(value.ToString()) { }
 }
 
-/// <summary>Lock shapes the lock-sugar pass must raise: a void lock, a lock with a value body, and a lock on a parameter.</summary>
+/// <summary>Lock shapes the lock-sugar pass must raise, including static-field, instance-field, and parameter receivers.</summary>
 public sealed class LockFixtureSamples
 {
+    static readonly object s_staticRoot = new();
+    static int s_staticValue;
     readonly object _root = new();
     int _value;
+
+    public static void IncrementStaticUnderLock()
+    {
+        lock (s_staticRoot) { s_staticValue++; }
+    }
 
     public void IncrementUnderLock()
     {
--- a/src/ILInspector.Decompiler.Tests/IrImporterTests.cs
+++ b/src/ILInspector.Decompiler.Tests/IrImporterTests.cs
@@ -5124,18 +5124,18 @@ public class LockSugarTests
     }
 
     [Fact]
-    public void CoreLibStaticFieldLock_PreservesCopiedReceiverLocal()
+    public void StaticFieldLock_PreservesCopiedReceiverLocal()
     {
-        using var source = MetadataSource.Open(typeof(object).Assembly.Location);
-        var function = IrImporter.Import(source, "System.AppContext", "GetData");
-        Assert.NotNull(function);
-        IrPasses.Run(function!);
+        var function = IrImportFor(nameof(LockFixtureSamples.IncrementStaticUnderLock));
 
-        var output = CSharpPrinter.Print(function!).Output;
+        var lockNode = Assert.Single(function.Descendants.OfType<Pipeline.Lock>());
+        var lockObject = Assert.IsType<LoadLocal>(lockNode.LockObject);
+        Assert.Contains(function.Descendants.OfType<StoreLocal>(), store => store.Index == lockObject.Index);
 
+        var output = CSharpPrinter.Print(function).Output;
         Assert.NotNull(output);
-        Assert.Contains("lock (V_1)", output);
-        Assert.DoesNotContain("lock (AppContext.s_dataStore)", output);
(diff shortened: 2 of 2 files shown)


richlander:
## Review — the canary now owns its input; assertions got stronger, not just stabler

The old test's flaw was that it asserted decompiler behavior against a *mutable external implementation* — whatever `System.AppContext.GetData` happens to compile to in the SDK's CoreLib du jour. Any runtime refactor of that method broke the canary spuriously, which is exactly why it was sitting in everyone's "reproduces on clean main" list (#2626, #2652 evidence tables). Compiling a local static-field lock fixture makes the canary self-contained, consistent with the fixture-governance direction the extension-member work established: compiler-emitted fixtures the repo controls, not shipped binaries it doesn't.

What makes this better than a like-for-like swap:

- **The assertions moved from textual to structural.** The old test checked two strings. The new one asserts all three boundaries independently: the importer produced the lowering (a single `Lock` node exists), `LockSugarPass` kept the copied receiver as a `LoadLocal` *with its defining store present* (the semantic that actually matters — the receiver copy is load-bearing, per the same reasoning that put the lock-receiver-carrier guard in #2645), and the printer renders through the local. A regression in any one layer now fails with a targeted assertion instead of a string mismatch.
- **`lock \(V_\d+\)` instead of hard-coded `V_1`** — robust to local-numbering churn from unrelated pass changes, while the negative assertion (no `lock (LockFixtureSamples.s_staticRoot)` inlining) still pins the actual defect mode from #2630.
- The fixture doc-comment update keeps `LockFixtureSamples`' inventory honest (static-field, instance-field, parameter receivers).

One trade worth stating for the record, not changing: the old test incidentally verified against IL *as actually shipped* in CoreLib — an archaeology check on real-world compilation. The new fixture is compiled by the current Roslyn, so if a future compiler changes the static-lock lowering shape, the canary follows it automatically. That's the right default (the corpus sensor is the tool for real-world IL coverage; unit canaries should be hermetic), but it means this test now guards "our passes handle the lowering Roslyn emits" rather than "…the lowering CoreLib happens to contain" — which is what you want from a *canary*.

Test-only, A/B correctly declared not-applicable, and merging it retires one of the four standing clean-main failures — the known-failure docket shrinks. Endorse.

---

richlander:

PR head `bec48d6d` has two clean isolated reviews:

- **Gemini 3.1 Pro — CLEAN.** Its first pass found that the redundant negative render assertion still depended on static-field qualification spelling. Commit `bec48d6d` removes that assertion; Gemini then re-ran the focused evidence and returned clean.
- **Claude Opus 4.8 — CLEAN.** It independently verified the compiler-emitted receiver-copy lowering in Debug and Release, the structural `LoadLocal`/`StoreLocal` proof, fixture-inventory isolation, and test-only scope. Its fixed-head confirmation found no outstanding issue after `bec48d6d`.

Both fixed-head reviews used separate clean detached worktrees.

### Reconciliation

- The maintainer review endorsed replacing the mutable CoreLib witness with a hermetic importer fixture and strengthening the proof from text-only to structural assertions.
- That review described the then-present negative assertion as an additional defect-mode pin. Gemini correctly observed that it relied on qualification spelling and was redundant: `Assert.IsType<LoadLocal>(lockNode.LockObject)` rejects field inlining structurally, while `lock \(V_\d+\)` proves the printer consumed the local. `bec48d6d` removes the weaker assertion rather than preserving a second spelling-sensitive check.
- No product code changed. Render A/B and corpus quality cards remain not applicable.

The solution build, focused lock suites, full Decompiler suite, and PR CI are green. GitHub reports the PR mergeable.

**Ready to merge.**

Issue #2529, opened 2026-07-08
RTS boundary: finish moving residual C# knowledge into product IR (post-#2520)


richlander:
## Summary

Follow-up to the RTS "cross the river" work (#2050) and the typed-evidence move in #2520 ("Carry initializer member evidence in IR").

The RTS ⇄ product boundary is healthy and directionally correct: product code (`ILInspector.Decompiler` — `CompileBackSourceComposer`, `IrImporter`, `CSharpPrinter`, IR passes) owns C# artifact production, and RTS (`tools/DecompilerHarness/ReturnToSender.cs`) is the tools-only orchestrator/oracle (target selection, Roslyn compile, compiler-driven closure growth, IL/opcode diff, failure classification, source-fidelity probe). Dependency edges match the design doc (`RTS → Research, Analysis, Decompiler, Roslyn`; no product project references RTS).

#2520 is the canonical exemplar of the intended move: it deleted RTS's name-string `AddNamedMemberRequirement` and the Composer's ~43-line string-name `TryCreateClosureMemberRequirement` overload, replacing them with typed `InitializerEntry.ConsumedMethod`/`ConsumedField` evidence populated by `ObjectInitializerPass`. C# knowledge moved from harness → product IR.

This issue captures the **residual C# knowledge still in RTS** and two **architectural gaps vs `docs/design/fact-planned-compile-back-harness.md`**, with suggested resolutions to finish the #2520 pattern.

> References are file/line as of the #2520 head (`feature/rts-ir-evidence-initializers`, `5de73d48`).

## Residual C# knowledge in RTS

All inside `SeedTypedClosureRoots` (`tools/DecompilerHarness/ReturnToSender.cs:1400–1666`), graded worst-first.

### 1. Operator-name synthesis (worst)

`IncrementDecrement` carries only booleans (`IsIncrement`/`IsChecked`/`IsUserDefined`); RTS *constructs* the `op_Increment`/`op_CheckedIncrement`/… `MethodRef` from those booleans + `Target.ResultType` (`ReturnToSender.cs:1584–1592`), and `UncheckedOperatorName` (`:1656`) hardcodes the `op_Checked* → op_*` sibling mapping (also invoked at `:1513`). This is pure C# operator-naming knowledge living in the harness — exactly the class of knowledge #2520 removed for initializers.

[...]

PR #2665, opened 2026-07-11
Move RTS closure seed behind artifact provider

richlander:
- Advances #2529
- Changes harness-only RTS/provider boundary: initial typed closure roots/facts/member requirements are now derived inside `CompileBackSourceComposer.Compose(ArtifactRequest)`; RTS carries only target/oracle closure roots and Roslyn feedback facts across iterations.
- Evidence revision: `0dba1240`

## Change

**Conclusion:** **REVIEW** — provider-owned initial closure seeding is now behind the artifact request envelope, while RTS still owns Roslyn oracle closure growth and opcode comparison.

## Scope and safety

- **Root cause:** RTS was still deriving typed closure inputs directly before artifact creation, keeping product-shaped declaration knowledge in the planner/oracle loop.
- **Fix shape:** `ArtifactRequest` no longer carries closure member requirements; `CompileBackSourceComposer` derives initial typed closure inputs from the raised `IrFunction` and overlays RTS oracle roots/facts on every compose.
- **Product impact:** Harness-only; product CLI output is unchanged.
- **RTS boundary:** RTS still orchestrates closure iteration, Roslyn diagnostics, and opcode compare; provider-side composition owns initial same-assembly typed closure derivation.

## Evidence

| Check | Result |
| --- | --- |
| Harness build | `dotnet build tools/DecompilerHarness/DecompilerHarness.csproj -c Release --configfile nuget.config --verbosity quiet` passed |
| Focused tests | `dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.ReturnToSenderPrototypeTests` passed, 96/96 |
| Markdown | `npx markdownlint-cli docs/design/fact-planned-compile-back-harness.md` passed |

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| Roslyn diagnostic/oracle closure growth | Not moved | It depends on Roslyn `Diagnostic`/`SemanticModel` and remains RTS/tools-only feedback, not provider-owned initial derivation. |
| Product/shared API extraction | Future slice | This PR keeps the behavior-preserving boundary inside the current harness provider path. |

## Review

Adversarial review requested separately per repo policy.

### DIFF
--- a/docs/design/fact-planned-compile-back-harness.md
+++ b/docs/design/fact-planned-compile-back-harness.md
@@ -262,6 +262,11 @@ The artifact provider should return structured diagnostics when it cannot
 produce an artifact. RTS should surface those diagnostics; it should not patch
 the artifact with local C#.
 
+Initial typed closure derivation belongs with artifact production: the provider
+can derive same-assembly roots, source facts, and member requirements from the
+raised `IrFunction`. RTS remains responsible for Roslyn oracle growth and passes
+only the target/oracle closure roots and facts discovered by compile feedback.
+
 The product/shared side should own truthful declaration facts that are useful
 beyond ReturnToSender: namespaces, type/member signatures, base/interface
 relationships, constructor signatures, generic constraints, explicit interface
--- a/tools/DecompilerHarness/ReturnToSender.cs
+++ b/tools/DecompilerHarness/ReturnToSender.cs
@@ -799,10 +799,6 @@ static class ReturnToSender
             targetRoot,
         };
         var closureFacts = new Dictionary<TypeDefinitionHandle, List<CompileBackFact>>();
-        var closureMemberRequirements = new Dictionary<TypeDefinitionHandle, List<CompileBackMemberRequirement>>();
-        // PrintRaised mutates the imported function in place, so raised evidence
-        // such as WithExpression is available to typed shell seeding here.
-        SeedTypedClosureRoots(reader, function, typeHandle, targetRoot, closureRoots, closureFacts, closureMemberRequirements);
         const int maxRoots = 200;
         const int maxIterations = 80;
         Diagnostic? firstError = null;
@@ -822,8 +818,7 @@ static class ReturnToSender
                 Overload: overload,
                 SignatureText: CorpusMethodIdentity.SignatureText(function.Signature),
                 ClosureRoots: closureRoots,
-                ClosureFacts: closureFacts,
-                ClosureMemberRequirements: closureMemberRequirements));
+                ClosureFacts: closureFacts));
             var plan = sourceResult.Plan;
 
             if (plan.Diagnostics.FirstOrDefault(diagnostic => diagnostic.Layer == "type identity") is { } identityDiagnostic)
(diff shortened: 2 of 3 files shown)


richlander:
Adversarial review reconciliation for updated head `e808cc47`:

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | CLEAN | Verified effective closure semantics/budget preservation, source facts/member requirements, no mutable-state leakage, SRM-only provider seed, and local build/tests. |
| Gemini 3.1 Pro | CLEAN | Verified `EffectiveClosureRootCount` set-union behavior, isolated oracle/provider state, fact/requirement preservation, SRM-only boundary, and local build/tests/markdownlint. |

First-round Opus noted a non-blocking budget semantic edge: after moving initial seed roots behind `Compose`, RTS's `maxRoots` check counted oracle roots only. Addressed in `e808cc47` by exposing effective provider closure roots on `ProductArtifact` and using `EffectiveClosureRootCount(...)` so the budget applies to provider-seeded plus oracle-grown roots. Gemini first round was clean.

No outstanding review findings remain.

richlander:
Ready to merge

richlander:
## Review — the first #2647 frontier delivered, with the budget semantics carefully preserved

This is the "move closure derivation behind the provider" slice that #2647's frontier table named, and the ownership split lands exactly where the design doc update says it should: **the provider derives what's knowable from the artifact itself** (same-assembly roots, source facts, member requirements — from the raised `IrFunction` inside `Compose` via `CreateClosureInputs`), while **RTS keeps only what compile feedback teaches** (Roslyn oracle growth). The ~200-line `SeedTypedClosureRoots` — including its subtle corelib-facade canonicalization comment — moves wholesale rather than being reimplemented, and `ClosureMemberRequirements` disappears from the request contract entirely since callers no longer have anything to say about it. The request shrinks toward "target + oracle evidence," which is the direction #2529 wants.

The detail that shows care: **`EffectiveClosureRootCount`**. The `maxRoots` budget used to be checked against a set that contained both seeded and oracle roots; after the move, RTS's set holds only target + oracle roots, so a naive count would have quietly *raised* the effective budget by however many roots the provider derives. Counting the union (artifact-derived roots plus oracle roots not already among them) keeps the stall/budget behavior aligned with the old semantics instead of letting the refactor loosen a limit nobody decided to loosen.

1. **The seeding now runs on every `Compose` call instead of once per target.** The old code seeded before the iteration loop; the new derivation executes per iteration (up to `maxIterations` = 80 times), re-walking the function's descendants each round. It's deterministic — same inputs, same derived roots — so this is pure repeated work, not a correctness issue, but on stubborn closures it's an O(iterations × function-size) cost the old shape didn't pay. If the provider ever caches per-`(reader, function, targetType)`, this is the first candidate.
2. **A subtle stall-timing shift, bounded and harmless**: a compile diagnostic can now "grow" the oracle set with a root the provider had already derived (previously that root was pre-seeded into the same set, so the addition was a no-op and the loop stalled a round earlier). The composed source is identical either way, so the loop stalls on the following iteration — worst case one extra compile per overlapping root. Worth knowing if closure-iteration counts in harness telemetry shift slightly after this merges.

Harness-only, contract-shrinking, and the doc paragraph in `fact-planned-compile-back-harness.md` now describes the boundary the code actually implements. Endorse.

---

PR #2667, opened 2026-07-12
Add shared Research member projection

richlander:
Refs #2653 (C6).

## Summary
- Add `ResearchViews.ProjectMember`, a per-member projection request/result that imports one selected method, builds one `ResearchAssemblyContext`, and shares collected facts/header facts across annotated source, cost overlay, semantics overlay, and structured fact rows.
- Keep existing individual `ResearchViews` APIs as wrappers.
- Migrate `MemberCodeProvider` to use one Research projection for the selected member-backed views.
- Add a focused Research test that asserts fact/header producers run once and see the same assembly context across requested views.

## Validation
- `dotnet run --project src/ILInspector.Research.Tests -c Release --no-restore -- -class ILInspector.Research.Tests.ResearchFactRegistryTests`
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-restore -- -class DotnetInspector.Tests.CommandExecutionTests -method "*Member_SelectedOverload_Select*" -method "*Member_SingleOverload_SourceCategory_IncludesSourceDiff*" -method "*Member_SelectedOverload_SourceCategory_IncludesIlAndNoLoweredSource*" -method "*Member_Facts_IsExplicitOnly_NotShownAtDetailed*"`
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-restore -- -class DotnetInspector.Tests.IndexBuildInvariantTests`
- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config --nologo -v:q`

## Review status
Adversarial reviews are required for this Research/analysis behavior change and are pending. This PR is not ready to merge yet.

### DIFF
--- a/src/ILInspector.Research.Tests/ResearchFactRegistryTests.cs
+++ b/src/ILInspector.Research.Tests/ResearchFactRegistryTests.cs
@@ -63,6 +63,33 @@ public class ResearchFactRegistryTests
     }
 
     [Fact]
+    public void MemberProjection_CollectsFactsOnceForRequestedViews()
+    {
+        using var source = MetadataSource.Open(typeof(ResearchFixture).Assembly.Location);
+        var producer = new CountingProducer();
+        var registry = new ResearchFactRegistry(producer);
+
+        var projection = ResearchViews.ProjectMember(new ResearchViews.MemberProjectionRequest(
+            source,
+            typeof(ResearchFixture).FullName!,
+            nameof(ResearchFixture.BoxInt),
+            AnnotatedSource: true,
+            CostOverlay: true,
+            SemanticsOverlay: true,
+            FactRows: true,
+            Registry: registry));
+
+        Assert.Equal(1, producer.FactCollectCount);
+        Assert.Equal(1, producer.HeaderFactCollectCount);
+        Assert.NotNull(producer.AssemblyContext);
+        Assert.Same(producer.AssemblyContext, producer.HeaderAssemblyContext);
+        Assert.Contains("cost.test", projection.AnnotatedSource!.Output);
+        Assert.Contains("cost.test", projection.CostOverlay!.Body.Output);
+        Assert.Contains("semantics.test", projection.SemanticsOverlay!.Output);
+        Assert.Contains(projection.Facts!, row => row.Id == "cost.header.test");
+    }
+
+    [Fact]
     public void DefaultAllocationProducer_ProjectsAnalysisFindingCensus()
     {
         using var source = MetadataSource.Open(typeof(ResearchFixture).Assembly.Location);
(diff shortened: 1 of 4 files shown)


richlander:
Adversarial review reconciliation for fixed head `d74de525`:

- Gemini final review: CLEAN. Re-reviewed `ResearchViews.cs`, `MemberCodeProvider.cs`, `ResearchFactRegistryTests.cs`, and `CommandExecutionTests.cs`; verified the prior trace-overwrite finding is fixed by preserving `researchProjection.Trace ?? decompileTrace`; reported no remaining correctness/security/compatibility findings.
- Opus final review: CLEAN. Re-reviewed the fixed diff against `origin/main`; verified view parity, member-provider wiring, trace precedence, and shared fact/header collection; reported no high-confidence correctness/security/compatibility regressions.

First-round Gemini and Opus both found the same trace-clobber regression for Decompiled Source + Facts. Resolved in commit `d74de525` by preserving the existing decompile trace when a Facts-only projection has no render trace, with regression coverage in `Member_SelectDecompiledSourceAndFacts_RequestTraceKeepsDecompileOutcome`. No dismissed blocking findings remain.

CI is green on the fixed head. Ready to merge.

richlander:
## Review — one import, one collection, N views; the consolidation is right, two seams to note

The duplication this removes was real: `RenderMixed`, `RenderCostOverlayWithHeaderFacts`, `RenderSemanticsOverlay`, and `CollectFactRows` each independently imported the method and ran the fact registry, so a CLI member view wanting source + cost + semantics + facts paid four imports and four collections — and, more subtly, computed its views from four *separate* imported functions. `ProjectMember(MemberProjectionRequest)` imports once, collects once, and projects whatever combination the request asks for; every view is now derived from the same `IrFunction` and the same fact set by construction. The legacy APIs survive as thin delegations (behavior-compatible for existing callers), the request/result envelope follows the house verb-nominalization, and `MemberCodeProvider` becomes the first multi-facet consumer.

Two seams worth a comment or a small fix, neither blocking:

1. **The combined-request error semantics are inherited, not chosen.** The catch block preserves each legacy API's posture — fact-rows requests re-throw (as `CollectFactRows` always did), render requests degrade to a `DecompilerResult.Failure` (as `Run` always did). But a *combined* request (`FactRows: true, AnnotatedSource: true`) now throws and discards the renders it could have produced — the strictest constituent wins, silently. Probably right (facts are the machine-consumable output; a caller asking for them wants hard failure), but it's a decision the new API makes without saying so; one comment on the `if (request.FactRows) throw;` line settles it. Related nit: the `request.FactRows ? [] : null` in the failure result below it is dead — unreachable past that throw.
2. **A small behavior change for no-assembly-path sources.** The old overlay paths returned *unannotated* output when `AssemblyPath` was empty (they skipped `Collect` entirely); `ProjectMember` always runs the registry (with a null assembly context) and annotates with whatever the producers yield — so decompiler-derived facts (e.g. lifetime annotations) now appear on overlays for path-less sources where they previously didn't. The producers all self-guard on missing assembly context, so this is contained and arguably *more* consistent — but it's a behavioral delta the delegation shims don't preserve, worth a line in the PR body if any path-less `MetadataSource` exists in practice.

Also good: `headerFacts` are collected only when a requesting view needs them; the result-level `Trace` picks the first available per-view trace rather than fabricating one; and the `!` assertions in the delegation shims are sound because each shim sets exactly the flag whose result it dereferences.

Endorse. This is the right shape for the member-view pipeline, and the single-import guarantee quietly removes a whole class of could-the-views-disagree questions.

---

PR #2668, opened 2026-07-12
Fix package dependency cache serialization

richlander:
## Summary

- replace delimiter-based dependency-group cache entries with structured JSON
- invalidate lossy `pkg-index-v9` entries by advancing the cache category to v10
- cover bounded and open NuGet version ranges that contain commas

## Validation

- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-build`
- `dotnet publish src/dotnet-inspect/dotnet-inspect.csproj -c Release -r linux-x64 --self-contained -p:PublishAot=true`

This is a focused cache-format correctness fix with no public output or API changes.

### DIFF
--- a/src/dotnet-inspect.Tests/CacheCommandTests.cs
+++ b/src/dotnet-inspect.Tests/CacheCommandTests.cs
@@ -1,5 +1,6 @@
 using DotnetInspector.Commands;
 using DotnetInspector.Core;
+using DotnetInspector.Inspectors;
 using DotnetInspector.Options;
 using DotnetInspector.Packages;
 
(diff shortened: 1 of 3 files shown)


richlander:
The first CI run exposed a process-wide cache test-isolation race: the new regression fixture called `CoreCache.Initialize` while the suite runs cache consumers in parallel, interrupting `CacheCommandTests` maintenance.

Resolved in `0fd62569` by testing the dependency-group serializer directly, removing all global cache initialization from the new fixture. The CI-equivalent fast CLI suite now passes locally (1,264 tests; 10 existing tool-availability skips).

richlander:
## Adversarial review — fixed head `0fd62569`

- **Claude Opus 4.8: CLEAN.** Verified the full `Set` → `FieldDocument` → `TryGet` path with bounded/open ranges, empty versions, and framework-less groups; confirmed malformed JSON becomes a cache miss, v9 cleanup is correct, and AOT/trim analyzers remain clean.
- **Gemini 3.1 Pro: CLEAN.** Independently verified comma/null preservation, single-line JSON compatibility with `FieldDocument`, invalid-cache containment, v10 invalidation, NativeAOT safety, and removal of the process-wide test side effect.

Both reviewers agreed there are no correctness, compatibility, security, or test-isolation findings. No review-resolution commit is needed. The earlier CI race was resolved separately in `0fd62569`; the corrected head has green CI and both isolated review worktrees remained clean.

**Ready to merge.**

richlander:
## Review — classic delimiter collision, fixed at the right layer

The bug: dependency groups were cached as `"tfm|name@ver,name@ver"`, but NuGet version *ranges* legally contain commas — `[1.0.0,2.0.0)` — so the comma-split shredded any ranged dependency on read-back, silently corrupting the cached dependency graph. The fix replaces the hand-rolled compact format with structured JSON per group, and the round-trip theory covers the three range shapes that break the old format (bounded, min-unbounded, max-bounded).

The two details that make this complete rather than just correct:

- **The cache category bump (`pkg-index-v9` → `pkg-index-v10`)** means every entry written in the corrupting format becomes unreachable rather than being read back wrong, and the existing obsolete-category maintenance cleans it up — poisoned data is evicted, not reinterpreted.
- **The test now references `PackageIndexCache.Category`** instead of hardcoding the version string, so the next format bump can't leave the cleanup test silently pinning a stale category name — the test follows the constant.

Round-trip parity is preserved for the null-TFM case (both old and new collapse null → empty string on read), and `SerializeDependencyGroup`/`DeserializeDependencyGroup` being `internal` + directly tested is the right testing seam.

One observation, no action: the old format also couldn't represent a dependency id containing `@` or `,` or a TFM containing `|` — all now free fixes from the structured encoding. Cheap insurance if you ever want it: one more theory case with an exotic-but-legal id. Endorse; safe to merge.

---

PR #2670, opened 2026-07-12
Route RTS target body rendering through provider

richlander:
- Advances #2529
- Changes harness-only RTS/provider boundary: target-body C# rendering is now represented as a provider-produced `ProductTargetBody` carried through `ArtifactRequest` / `ProductArtifact`.
- Evidence revision: `009c2f1e`

## Change

**Conclusion:** **REVIEW** — RTS still selects/imports the target method and owns Roslyn compile feedback, but it no longer calls `CSharpPrinter.PrintRaised` directly for ReturnToSender artifacts.

## Scope and safety

- **Root cause:** The previous envelope still accepted raw target-body source text from RTS, so the planner called the product C# output printer directly before constructing an artifact request.
- **Fix shape:** Add `ProductTargetBody` and `CompileBackSourceComposer.CreateTargetBody(...)`; artifact requests carry the typed provider-produced body/provenance, and `Compose(...)` consumes its source text.
- **Product impact:** Harness-only boundary cleanup; emitted target body, decisions, and compile-back source should be unchanged.
- **RTS boundary:** RTS orchestrates target selection/import, Roslyn diagnostics, closure growth, compilation, and opcode comparison. Product/provider path owns target-body rendering and artifact composition.

## Evidence

| Check | Result |
| --- | --- |
| Harness build | `dotnet build tools/DecompilerHarness/DecompilerHarness.csproj -c Release --configfile nuget.config --verbosity quiet` passed |
| Focused tests | `dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.ReturnToSenderPrototypeTests` passed, 96/96 |
| Markdown | `npx markdownlint-cli docs/design/fact-planned-compile-back-harness.md` passed |

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| Importing the target `IrFunction` | Not moved | RTS still selects and imports the target for this behavior-preserving slice. |
| Roslyn diagnostic/oracle closure growth | Not moved | It remains RTS/tools-only compile feedback. |

## Review

Adversarial review requested separately per repo policy.

### DIFF
--- a/docs/design/fact-planned-compile-back-harness.md
+++ b/docs/design/fact-planned-compile-back-harness.md
@@ -266,6 +266,9 @@ Initial typed closure derivation belongs with artifact production: the provider
 can derive same-assembly roots, source facts, and member requirements from the
 raised `IrFunction`. RTS remains responsible for Roslyn oracle growth and passes
 only the target/oracle closure roots and facts discovered by compile feedback.
+Target-body rendering is also provider-owned: RTS selects the target and carries
+the provider-produced body/provenance, but should not call the C# output printer
+directly.
 
 The product/shared side should own truthful declaration facts that are useful
 beyond ReturnToSender: namespaces, type/member signatures, base/interface
--- a/tools/DecompilerHarness/ReturnToSender.cs
+++ b/tools/DecompilerHarness/ReturnToSender.cs
@@ -777,9 +777,7 @@ static class ReturnToSender
 
         var function = IrImporter.Import(source, fullType, methodName, overload)
             ?? throw new InvalidOperationException($"Could not import {fullType}::{methodName}.");
-        var printed = CSharpPrinter.PrintRaised(function, method => IrImporter.Import(source, method));
-        if (printed.Output is null)
-            throw new InvalidOperationException($"Could not print {fullType}::{methodName}.");
+        var targetBody = CompileBackSourceComposer.CreateTargetBody(source, function, fullType, methodName);
 
         var original = ILInstructionPrinter.Disassemble(pe, reader, getter)
             ?? throw new InvalidOperationException($"Could not disassemble {fullType}::{methodName}.");
@@ -812,7 +810,7 @@ static class ReturnToSender
                 TargetType: typeHandle,
                 TargetMethod: getterHandle,
                 TargetProperty: propertyHandle,
-                TargetBody: printed.Output,
+                TargetBody: targetBody,
                 FullType: fullType,
                 MethodName: methodName,
                 Overload: overload,
@@ -830,9 +828,9 @@ static class ReturnToSender
                     string.Join(" ", originalOps),
                     "",
                     $"{identityDiagnostic.Reason}: {identityDiagnostic.Detail}",
-                    TargetBody: printed.Output,
+                    TargetBody: targetBody.Source,
                     MemberAnchor: memberAnchor,
-                    Decisions: printed.Decisions);
+                    Decisions: targetBody.Decisions);
             }
 
             string unit = sourceResult.Source;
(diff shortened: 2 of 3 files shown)


richlander:
Adversarial review reconciliation for head `009c2f1e`:

- Opus 4.8: CLEAN. Verified behavior preservation, unchanged `PrintRaised` mutation timing, unchanged result fields, intact request/result shape, no broken external call sites, and no provider-boundary violation.
- Gemini 3.1 Pro: CLEAN. Verified the same target-body rendering call and null-output exception behavior, unchanged mutation timing across getter/method/setter paths, unchanged external result surface, and no Roslyn/product-boundary leakage into the provider path.

No follow-up commits were needed. Local validation and PR CI are green.

Ready to merge.

richlander:
## Review — behavior-preserving by inspection; the ×3 diff is the loudest argument yet for the loop collapse

Verified mechanically: `CreateTargetBody` is the *verbatim* code RTS previously ran inline — same `CSharpPrinter.PrintRaised` call with the same import callback, same null-output throw with the identical message — and the threading swap (`printed.Output` → `targetBody.Source`, `printed.Decisions` → `targetBody.Decisions`) is applied symmetrically across all three loops, with the body still rendered once per target before the iteration loop rather than per-iteration. Pure move, no behavior or perf delta; the focused 96/96 is adequate evidence for a slice this shape.

The boundary itself is the right next step in the #2529 sequence: after #2665 moved closure derivation behind the provider, the raw `string TargetBody` on the request was the last place RTS handed the provider *pre-rendered product output* instead of typed material. `ProductTargetBody(Source, Decisions)` fixes the type of the conversation — RTS selects and imports; the provider renders and owns the decisions provenance — and the doc paragraph lands the rule where the next contributor will find it. The frontier table honestly scopes what did *not* move (target import, oracle growth), which keeps this slice reviewable.

1. **`ProductArtifact.TargetBody` is currently an echo** — it carries `request.TargetBody` back to a caller who necessarily already has it. That's sensible pre-positioning for the next slice (when the provider derives the body internally and the request stops carrying it, the artifact becomes the source of truth) — but until then it's an unexercised field, same status as `SourceFacts` was in #2647. When the first real consumer arrives, pin it then.
2. **This diff is the strongest evidence yet for the compile-loop collapse.** Every hunk in `ReturnToSender.cs` appears exactly three times, character-identical modulo handle names — this PR had to make the same edit in twelve places across the getter/method/setter loops. Each boundary slice is now paying a ×3 tax, and the tax collector is visible in the diff itself. I'd promote the loop unification from "named frontier" to "next slice" before further envelope work multiplies through it again.

Minor: the new `using DotnetInspector.Services;` in the planner — if it's load-bearing for `MetadataSource`/`DecompilerDecision` resolution, fine; if it's a leftover from an earlier draft, drop it.

Endorse. Harness-only, verifiable by reading, and the boundary it draws is the one the design doc says comes next.

---

richlander:
Updated reconciliation for head `9383d452` after addressing review feedback:

- Follow-up change: removed the unused `DotnetInspector.Services` import from `ReturnToSenderTypePlanner.cs` in commit `9383d452`.
- Opus 4.8 r2: CLEAN. Verified clean checkout at exact head, successful DecompilerHarness build, no symbol-binding dependency on `DotnetInspector.Services`, unchanged target-body rendering behavior/mutation timing, unchanged `Result.TargetBody` and decisions flow, and no Roslyn/oracle responsibility moved into the provider.
- Gemini 3.1 Pro r2: CLEAN. Verified clean checkout at exact head, successful DecompilerHarness build, no binding/extension-method issue from the import removal, preserved `CreateTargetBody` behavior, preserved artifact/request data flow, and intact provider seam boundaries.

PR CI is green on this head.

Ready to merge.

richlander:
Re-verified @ `9383d452`: the delta is exactly the stray-import removal from my review — the `using` was a leftover, not load-bearing, and nothing else moved. Endorsement stands unchanged: behavior-preserving by inspection, safe to merge. (The loop-collapse recommendation and the echo-field note remain follow-up material for the next #2529 slice.)

---

PR #2671, opened 2026-07-12
Expose call-site Finding transitions

richlander:
## Conclusion

`diff --finding analysis.call-site` now confirms when one caller method gained,
lost, retained, or changed a direct-call occurrence using Analysis's native
`PairFinding<DirectCall>` comparison.

## What changed

- Added `analysis.call-site` to the existing focused `Finding Transitions`
  descriptor selector.
- Requires exactly one caller method and compares only the supplied endpoints.
- Retains complete call-site comparisons in Research only when requested, so
  exact `Present` pairs remain visible without increasing ordinary Body Signals
  work.
- Preserves Analysis-owned exact instantiated-callee identity and
  dispatch/opcode/loop facet classification.
- Supports standalone table/Markdown output and composed JSON documents.
- Documents the allocation-onset to call-cause workflow in design and agent
  guidance.

```bash
dotnet-inspect diff --package Foo@1.4.0..1.5.0 \
  -t Foo.Parser -m Parse \
  --finding analysis.call-site
```

`PairFinding.Added` is the authoritative new-call occurrence. `Changed` can
show that a retained call moved into a loop or changed dispatch/opcode facets.

## Validation

- Solution build succeeded.
- Analysis: 430 passed.
- Research: 87 passed.
- CLI: 1,765 passed, 10 environment skips.
- Changed Markdown passes markdownlint.

## Adversarial review

- **Claude Opus 4.8: CLEAN.** Verified producer authority, structural callee
  identity, opt-in default-path cost, parameter threading, public constructor
  compatibility, pre-acquisition validation, JSON routing, and real focused
  tests.
- **Gemini 3.1 Pro: CLEAN.** Verified generic/overload identity, duplicate
  occurrence matching, complete native pair retention, lazy call-index
  acquisition, descriptor normalization, Combine propagation, and full
  Research/CLI diff tests.

The reviewers agreed and reported no actionable findings, so no review
resolution commit applies.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -160,7 +160,9 @@ separate axes; see [Finding Coordinates](design/finding-coordinates.md).
   Allocation changes retain Analysis's
   `FindingComparison<AllocationOccurrence>` on their Research projection;
   count, hot/cold, and ranking fields are derived from that comparison rather
-  than from a separate count-only diff.
+  than from a separate count-only diff. Focused endpoint-confirmation consumers
+  can also request complete allocation or direct-call comparisons, including
+  exact pairs that have no `ResearchChange` row.
 
 `ResearchFactRegistry` is the dogfooded analyzer registry for the overlay.
 Producers implement `IResearchFactProducer` with a stable name, produced fact
--- a/docs/design/finding-nomenclature.md
+++ b/docs/design/finding-nomenclature.md
@@ -141,6 +141,21 @@ Research retains exact as well as changed allocation comparisons when this lens
 requests them, so it can report `PairFinding.Present`, not only candidate
 boundaries.
 
+The same contract identifies a likely cause after an allocation onset. The
+caller method is the subject; each exact instantiated callee is a call-site
+observation:
+
+```bash
+dotnet-inspect diff --package Foo@1.4.0..1.5.0 \
+  -t Foo.Parser -m Parse \
+  --finding analysis.call-site
+```
+
+`PairFinding.Added` confirms a new direct-call occurrence. `Changed` can report
+that a retained call changed dispatch, opcode, or loop context. IL offsets and
+metadata tokens remain endpoint-local provenance rather than correspondence
+identity.
+
 ## Research composition
 
 Research is the join layer for sibling producers. It may retain a producer's
--- a/docs/design/finding-producers.md
+++ b/docs/design/finding-producers.md
@@ -174,11 +174,14 @@ projection from the same comparison path; there is no separate count-only
 allocation diff path. The CLI selects and retains the native comparison with
 `diff --finding analysis.allocation`.
 
+Direct calls are the second end-to-end Analysis proof. Research's call-site
+fact producers and `diff --finding analysis.call-site` consume the same
+IL-ordered `Finding<DirectCall>` census. Research retains the complete native
+comparison only when requested, so `Present` remains observable without
+increasing the ordinary Body Signals result.
+
 The remaining Analysis body observations preserve their native families:
 
-- direct calls use an IL-ordered `Finding<DirectCall>` census keyed by exact
-  instantiated callee identity; dispatch, opcode, and loop context remain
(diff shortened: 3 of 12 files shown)


richlander:
Adversarial review reconciliation for `66ab15bb`:

- **Claude Opus 4.8 — CLEAN.** Verified Analysis remains the sole producer authority, exact instantiated-callee identity excludes endpoint-local token/offset provenance, call-index acquisition stays opt-in, public Research constructors remain available, and validation/JSON routing reproduce correctly.
- **Gemini 3.1 Pro — CLEAN.** Verified generic/overload and duplicate-occurrence matching, complete `Present` retention, lazy default behavior, descriptor normalization, `Combine` propagation, and real Research/CLI tests.

The reviewers agreed and reported no actionable findings. No review-resolution commit applies.

richlander:
Ready to merge. Merge-blocking CI is green, the current head `66ab15bb` has clean Claude Opus 4.8 and Gemini 3.1 Pro adversarial reviews, and no actionable review feedback remains.

richlander:
## Review — third descriptor on the lens, template held; time to name the generalization the fourth one will demand

This lands `analysis.call-site` exactly on the rails #2661 laid for `analysis.allocation`, and every load-bearing decision carries over correctly:

- **Producer authority is preserved end-to-end.** The retained comparison is built by `AnalysisFindings.CompareCallSites` over raw occurrence arrays — the #2643 identity (exact instantiated callee, calling convention and arity included) and the deliberate facet whitelist (dispatch/opcode/loop) do the classification; the CLI only renders `PairFinding.{Kind}` verbatim. No parallel matcher, no costume.
- **The opt-in cost discipline is genuinely opt-in**: `BuildAnalysisSnapshot` fetches `GetDirectCallsByCaller()` only when `RetainCallSiteComparisons` is set (`callsByToken = includeCallSites ? ... : null`), so ordinary Body Signals runs pay nothing — the exact property both adversarial reviewers verified, and the right one to have checked.
- **Validation stays pre-acquisition** and the exactly-one-member guard generalizes to both analysis descriptors with a parameterized message instead of a second copy.
- **`CallSiteTarget` renders structural callee identity** (declaring type, name, type arguments, parameters, constructor special-case) with the `Unsupported`/calli fallback degrading to the declaring-type display — the Target column speaks the producer's identity language, and display stays out of correspondence.
- **The docs tell the workflow story, not just the flag**: the nomenclature and performance-skill additions position this as the *cause step after an allocation onset* — confirm the allocation boundary, then compare the same caller's call census, with "use the single-version `Calls` section while probing; use this final adjacent comparison as the onset proof" restating the probe-vs-proof discipline in agent guidance. That's the composed workflow the lens family was built for.

Two forward-looking items, neither blocking:

1. **The retained-comparison plumbing just got its second copy, and the third is predictable.** `CallSiteFindingComparison` replicates the allocation shape exactly: a record, a ctor overload (the public ctor chain is now three deep), a `ResultBuilder` list, a `Retain*` flag, an `Add` overload, and a `SelectMany` merge in `Compare`. When `analysis.unsafety` arrives — and the #2643 producer means it will — that's the earns-a-place threshold for one generalized retained-comparisons container (descriptor-keyed, or a small generic wrapper) replacing the per-payload plumbing. Worth a note in the code or the tracking issue now so the fourth descriptor is the refactor trigger rather than the fourth copy.
2. **Coordinate with #2672** — both PRs are based on the same main and both rewrite `DiffCommand.cs` (this one in the guard/builder region, #2672 restructuring the whole acquisition layer around `DiffEndpoint`, including the surface fallbacks your new builder calls into). Same drill as #2650/#2661: whichever lands second has a mechanical but real rebase; #2672's `MergeSurfaces` → `AssemblySetSurfaceBuilder` swap touches lines this PR's member-target resolution sits next to.

Both adversarial reviews CLEAN with verification lists that cover the properties that matter (identity, opt-in cost, ctor compatibility, JSON routing). Endorse — safe to merge, with the #2672 ordering decided first.

---

PR #2672, opened 2026-07-12
Move diff acquisition onto assembly sets

richlander:
## Summary

- Route package, platform, and local diff endpoints through `AssemblySetResolver`.
- Retain selected package TFM and centralize deterministic multi-assembly API surface construction in `DotnetInspector.Services`.
- Let owned `AssemblySet`s control extraction lifetime; remove `DiffCommand.MergeSurfaces`, temp-directory bookkeeping, and the stale duplicate merge path.
- Keep range parsing, compatibility filtering, ranking, and rendering in the CLI.

`AssemblySetPackageSelectionMode.TargetFramework` now uses the historical diff selector for every assembly-set consumer, not only `diff`. Existing consumer coverage remains green. The intentional observable delta is that satellite resource assemblies are excluded before consumers search or merge the set; they do not contribute API types, and the resolver test pins that behavior.

## Demo

```console
$ dotnet-inspect diff --package System.CommandLine@2.0.0-beta4.22272.1..2.0.0 -n 8
# API Diff: System.CommandLine

| Field | Value |
| ----- | ----- |
| Versions | **2.0.0-beta4.22272.1** -> **2.0.0** |
| Summary | **Summary:** 182 breaking, 92 additive, 1 potentially breaking across 83 types |

## Breaking Changes
```

## Evidence

- `dotnet build dotnet-inspect.slnx -c Release`: succeeded.
- `DotnetInspector.Services.Tests`: 254 passed.
- `dotnet-inspect.Tests`: 1760 passed, 10 skipped.
- Package, platform, and local success output A/B against `origin/main`: byte-for-byte identical.
- Platform resolution errors retain the prior fatal wording without the collection-only `skipping` suffix.
- `markdownlint` passes for the changed design doc.

Refs #2653 (C3).

### DIFF
--- a/docs/design/assembly-inspection-query.md
+++ b/docs/design/assembly-inspection-query.md
@@ -196,6 +196,14 @@ assembly that *defines* the selected member (via the type-lookup path), so a mem
 query over a package resolves to one reference, not many. Fan-out therefore happens only for
 assembly-level inspection without a selector.
 
+**Incremental acquisition bridge.** `AssemblySetResolver` is the current lower-layer primitive
+for that fan-out. It returns an owned `AssemblySet`: entries retain source, version, source kind,
+and selected TFM, while the set owns package-extraction directories until disposal.
+`AssemblySetSurfaceBuilder` composes an acquired set into one deterministic `ApiSurface` when a
+consumer, such as `diff`, needs package-level API comparison. The CLI still owns endpoint-range
+parsing, compatibility filtering, ranking, and rendering; it does not select package TFMs, merge
+assembly surfaces, or manage extraction directories.
+
 ### 3. `AssemblyInspectionSession` — one PE-lifetime owner, composing `PdbContext`
 
 Opened from a `ResolvedAssemblyReference`, it owns the `PEReader`/`MetadataReader`, opens once,
--- a/src/DotnetInspector.Services.Tests/AssemblySetResolverTests.cs
+++ b/src/DotnetInspector.Services.Tests/AssemblySetResolverTests.cs
@@ -1,4 +1,5 @@
 using System.IO.Compression;
+using ILInspector.Metadata;
 
 namespace DotnetInspector.Services.Tests;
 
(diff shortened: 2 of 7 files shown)


richlander:
## Review — diff joins the resolver cleanly; one cross-consumer change needs a sentence, one writer-consistency nit

The C3 slice does for `diff` what #2660 did for the find/extensions family, and the lifetime cleanup is the best part: `DiffEndpoint : IDisposable` owning its `AssemblySet` deletes the hand-rolled temp bookkeeping (`FromTempDir`/`ToTempDir`/`DeleteTempDir`) entirely, and I traced the error paths — `from` is disposed when `to` fails, failed endpoint resolution disposes its own set, and success hands both to `DiffInputs.Dispose`. No leak window. `AssemblySetSurfaceBuilder` centralizing the merge also quietly kills a fail-silent spot: the old `MergeSurfaces` skipped a null per-assembly surface with a bare `continue`; the builder logs `! {file}: API surface unavailable`. And recording the selected TFM on `AssemblySetEntry` turns what was a tuple side-channel into typed provenance, with the builder's single-package/single-TFM attribution rules being conservative in the right way (name only when *all* entries share one package source; TFM only when exactly one distinct).

The byte-for-byte A/B on all three endpoints plus the full suites is the right evidence for the diff-side claim. Two things to settle:

**1. The resolver's default package selection changed for *every* consumer, not just diff — say so and own it.** The `TargetFramework` mode swaps `TfmResolver.ResolvePackagePath` + directory scan for `TfmSelector.SelectHighestAssembliesFromPackage` (diff's historical selector). The evidence that this is safe for the find/extensions family is real but implicit: the #2660 consumer pins (runtimes opt-in, TFM selection, source order) pass **unmodified**, and the new resolver tests pin the selection directly. But the PR body scopes its parity claim to diff A/B, while the blast radius is all six consumer families — this is exactly the accidental-inheritance class #2660 just spent a cycle pinning, so the body should state the intended equivalence explicitly. One observable delta I can identify: `TfmSelector` filters **satellite resource assemblies**, which the old TFM-directory scan included — so `find --package`/`extensions --package` may stop surfacing `*.resources.dll` entries. That's noise removal (resource assemblies carry no API types worth matching), but it's a behavior change for non-diff consumers and deserves a line, ideally a pin.

**2. The diff path bypasses the single-owner diagnostic writer.** `ResolveDiffEndpointAsync` consumes `Diagnostics.LastOrDefault()?.Message` as the error when the set is empty — but when the set is *non-empty with warnings*, nothing renders on stderr. #2660 established `AssemblySetDiagnosticWriter` precisely so severity handling wouldn't drift per-consumer; diff should call it like the other five families do. Related micro-nit: `LastOrDefault()` on the empty-set path discards all but one diagnostic — first-plus-count or a join keeps multi-source failures whole.

Minor, no action needed: `DiffEndpoint.Paths` re-materializes a list on every access and it's read several times per command (Build* fallbacks, analysis lanes) — a field or lazy would do; and the new TFM-aware empty-package error messages ("No assemblies found for target framework 'X' in package 'Y'") are strictly better than the old generic ones — error-path text drift in the improving direction is fine by me.

Endorse with the cross-consumer statement added (and ideally the satellite-assembly pin); item 2 is a three-line consistency fix worth taking in the same round.

---

richlander:
Addressed the review feedback in 94a69402:

- The PR body now explicitly owns the cross-consumer `TargetFramework` selector change and the intentional satellite-resource exclusion. The resolver test name and fixture pin that exclusion directly.
- `diff` now uses `AssemblySetDiagnosticWriter` for warnings on non-empty sets and preserves every empty-set diagnostic instead of selecting only the last one.
- The same commit closes exception-path ownership gaps in both `AssemblySetResolver` and the two-endpoint diff flow, with a cleanup regression test.
- Platform fatal errors remove the collection-only `, skipping.` suffix and match the prior CLI behavior.

Full Services/CLI suites, solution build, and the package/platform/local output A/B pass on this head.

richlander:

Reviewed the exact final head `94a69402` in separate clean worktrees with **Claude Opus 4.8** and **Gemini 3.1 Pro**.

### First round

- **Opus** found exception-path ownership gaps: `AssemblySetResolver` could lose tracked extraction directories if collection threw, and an already-resolved `from` endpoint could leak if `to` resolution threw.
- **Gemini** found a platform error-text regression: fatal diff errors inherited the collection-only `, skipping.` suffix.
- The PR feedback also requested that the cross-consumer package-selector change be explicit, satellite exclusion be pinned, and non-empty-set diagnostics flow through `AssemblySetDiagnosticWriter` without dropping multi-diagnostic failures.

All actionable guidance was resolved in 94a69402: exception cleanup and endpoint disposal are now guarded and regression-tested; platform error wording matches the prior CLI path; satellite filtering is explicitly pinned and documented; non-empty warnings use the shared writer; and empty-set failures retain every diagnostic.

### Clean re-review

- **Claude Opus 4.8:** CLEAN — verified disposal/idempotence, diagnostics, TFM/runtime/satellite selection, surface composition, local error classification, cross-consumer modes, and full suites.
- **Gemini 3.1 Pro:** CLEAN — independently verified all exception paths, deterministic ordering, package `all` semantics, surface counts/name/TFM, local missing-vs-invalid behavior, and platform suffix handling.

No findings remain open on the final head.

richlander:
Ready to merge.

Final head `94a69402` has green CI, the full local Services/CLI suites and solution build pass, package/platform/local success outputs remain byte-for-byte identical to `origin/main`, and both isolated exact-head adversarial re-reviews are CLEAN.

richlander:
## Follow-up @ `94a69402` — all asks landed, plus exception-path hardening; endorsement now unconditional

- **The diff path now renders through `AssemblySetDiagnosticWriter`** — the single-owner rendering consistency with the other five consumer families, as asked.
- **Empty-set errors join *all* diagnostics** instead of `LastOrDefault` — multi-source failures survive whole.
- **The satellite-assembly delta got its pin**: `CollectAsync_PackageSelectsStableHighestTfmAssemblySetAndFiltersSatellites` turns the one observable cross-consumer behavior change I identified into an asserted decision rather than an accident.
- Bonus beyond my asks: `CollectAsync_ExceptionAfterPackageExtractionCleansOwnedDirectory` — an exception between extraction and set construction no longer leaks the temp directory, which closes the one lifetime window the `DiffEndpoint` ownership model couldn't reach.

Safe to merge from my side.

---

Issue #2280, opened 2026-07-04
RTS parity burndown for retiring compile-back coverage


richlander:
## Goal

Retire the legacy compile-back fidelity harness only after ReturnToSender (RTS)
has equal-or-better coverage as the corpus fidelity oracle.

Compile-back is currently still the coverage gate. RTS is the strategic
replacement, but it is not yet at coverage parity on the committed corpus sample.

## Current evidence

Measured at `d414f390` in a clean detached worktree:

```bash
git worktree add --detach /tmp/dotnet-inspect-rts-coverage d414f390
dotnet restore tools/DecompilerHarness/DecompilerHarness.csproj --configfile nuget.config --verbosity quiet
dotnet build tools/DecompilerHarness/DecompilerHarness.csproj -c Release --no-restore --nologo --verbosity quiet
```

Current committed real-world compile-back baseline:

- `tools/DecompilerHarness/corpus/real-world-baseline.json`
- 14 assemblies, 89,065 methods
- validity cap: 4000
- fidelity cap: 3 per assembly
- fidelity sample: 36 checked methods, 31 `Exact`, 5 `OpcodeDiff`, 0 `RecompileFail`, 0 `ContextFail`
- sampled method kinds: 13 getters, 1 setter, 9 constructors, 13 ordinary/operator/source-gen methods

The PR quick corpus deliberately does not run compile-back fidelity:

```text
--compile-cap 0
--corpus-fidelity-cap 0
```

so the daily real-world corpus is the active aggregate compile-back coverage
surface.

RTS current positive coverage:

```bash
dotnet run --project tools/DecompilerHarness -c Release --no-build -- \
  --return-to-sender-catalog --max-examples 20
```

```text
RETURNTOSENDER GENERATED FIXTURE FRONTIER over 41 fixture(s), 119 target(s)

Fixtures:
  Passed : 41
  Skipped: 0
  Failed : 0

Targets:
  Passed : 119
  Skipped: 0
  Failed : 0
```

```bash
dotnet run --project tools/DecompilerHarness -c Release --no-build -- \
  --return-to-sender-catalog record --max-examples 20
```

```text
RETURNTOSENDER GENERATED FIXTURE FRONTIER over 7 fixture(s), 12 target(s)

Fixtures:
  Passed : 7
  Skipped: 0
  Failed : 0

Targets:
  Passed : 12
  Skipped: 0
  Failed : 0
```

Catalogue inventory:

- 50 addressable fixtures
[...]

Issue #2673, opened 2026-07-12
SourceLink as a second decompiler oracle: fault isolation, gap mining, integrity triangulation, and ground-truth diff evidence


richlander:
## Premise

RTS and SourceLink are oracles for **different questions**, and the value is in triangulating them — never blending them.

- **RTS (Roslyn round-trip)** answers *"is the decompiled C# correct?"* — recompile, compare opcodes, compiler as judge. It cannot judge shape: a method rendered as goto-soup that recompiles exactly passes at 100%.
- **SourceLink** provides the *authored source* — ground truth for the question RTS can't ask: *"how far is our raising from what a human wrote?"*

Design invariant (the honest-labeling rule applied to oracles): **each lane reports its own verdict.** The moment SourceLink evidence *changes* a compile-back outcome rather than sitting beside it, the round-trip oracle is contaminated and RTS loses the property that makes it trustworthy. SourceLink lanes are diagnostic, comparative, or product evidence — never inputs to what RTS proves about the decompiler, except in the explicitly-labeled fault-isolation mode below.

## Slice 1 — Fault isolation inside RTS (smallest, do first)

Today a `RecompileFail` conflates two very different defects: *our decompiled body is wrong* vs *our reconstructed shell/closure is wrong*.

With SourceLink coverage, on failure re-run the **same `ArtifactRequest` with the authored body substituted**:

- original body compiles in our shell, decompiled body doesn't → **body defect**
- neither compiles → **shell/closure defect**

This splits the largest failure bucket into two actionable dockets, and independently validates the shell machinery — the authored body is the one input known to be legitimate C#.

Two facts make this nearly free now:

1. **#2670's `ProductTargetBody` boundary** — the target body is a typed provider-produced input on the request, so substitution is literally constructing a second request with a differently-sourced body. No printer plumbing changes.
2. **The synergy runs both ways**: per-method compilation of SourceLink source is normally infeasible (original files need whole-project context — `#if`, partial classes, generated source). RTS's shell reconstruction is exactly what makes per-method original-body compilation possible. Substitute *body text into our shell*; never attempt to compile original files standalone.

## Slice 2 — Gap mining: a source-correspondence census (Finding producer)

For corpus methods with SourceLink coverage, **structurally** compare decompiled output against authored source (Roslyn syntax trees, normalized — never text diff) and bucket divergences:

[...]

PR #2675, opened 2026-07-12
Measure native RTS parity in the corpus sensor

richlander:
- Advances #2280
- Changes harness-only corpus fidelity orchestration; product output is unchanged
- Evidence revision: `f353a139`

## Change

**Conclusion:** **PASS** — `CorpusSensor` can now evaluate its exact
compile-back checked sample through native ReturnToSender outcomes, with
explicit mode, reference-status, capture-provenance, failure-bucket, and parity
evidence.

## Scope and safety

- **Root cause:** RTS fidelity was available only through standalone commands,
  so the corpus sensor and baseline schema could not exercise or track the
  strategic replacement oracle.
- **Fix shape:** Add opt-in
  `--corpus-fidelity-oracle rts-parity`, align RTS outcomes to the exact
  compile-back `Exact`/`OpcodeDiff` target sample, and persist rescued/same/worse
  parity plus per-method reference/capture provenance.
- **Oracle isolation:** Corpus parity disables RTS's compile-back floor. The
  compile-back result remains beside each native RTS result as reference
  evidence and cannot rewrite RTS metrics or parity.
- **Product impact:** None; this is harness-only and compile-back remains the
  default corpus oracle.
- **Scope boundary:** This does not retire compile-back or regenerate the
  committed baseline.
- **RTS boundary:** RTS still owns orchestration and proof only. This PR does
  not modify artifact requests, provider-owned target-body rendering,
  `CompileBackSourceComposer`, or closure composition. It only adds an explicit
  no-floor option to the existing target evaluator for corpus parity.

## Evidence

| Check | Result |
| --- | --- |
| Solution build | Pass |
| Focused tests | 117 corpus-sensor and RTS tests pass |
| Merged RTS integration | 80 fixture-catalog + 16 generated-catalog tests pass |
| Prior full-suite failure | #2685 is fixed on current main; targeted regression passes |
| Fixed real-world corpus | 14 assemblies / 89,065 methods / sequential cap 3 |
| RTS parity | 0 rescued / 36 same / 0 worse |
| Native RTS outcomes | 31 Exact / 5 OpcodeDiff / 0 RecompileFail / 0 ContextFail |
| Capture provenance | 36 native RTS / 0 compile-back-floor rows |
| Deep NuGet probe | 48 Exact / 21 OpcodeDiff / 36 RecompileFail; 0 rescued / 48 same / 57 worse |
| Product output | Unchanged; harness-only |

The fixed-corpus run completed in 7:08.80 with a maximum RSS of 2,472,384 KB.
The deterministic 105-target NuGet.Versioning probe demonstrates the review fix:
36 rows previously hidden by the floor now report native `RecompileFail`, and
all 57 native regressions count as worse. The 21 constructor opcode gaps remain
visible under #2678.

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| Make RTS the committed baseline oracle | Not changed | Requires same-cap daily evidence and the broader retirement decision under #2280/#2560. |
| Remove compile-back sampling | Not changed | Compile-back still defines the bounded parity population. |
| Compile-back floor | Disabled for parity | Native RTS statuses exclusively feed parity metrics; the floor remains the default for other RTS callers. |
| Constructor parity | Exposed, not fixed | A deterministic NuGet.Versioning probe finds 21 constructor `Exact` → `OpcodeDiff` rows (#2678). |
| Parallel cap sampling | Exposed, not fixed | Identical capped runs can select different methods; comparisons gate on sample identity (#2679). |
| Full-suite union fixture | Resolved on main | #2685's targeted regression passes after rebasing over #2683. |
| #2670 provider rendering | Not changed | Rebased over merged #2670; provider rendering remains outside this PR's diff. |

## Review

Exact-head adversarial re-review is in progress for `f353a139`.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CorpusSensorComparisonTests.cs
+++ b/src/ILInspector.Decompiler.Tests/CorpusSensorComparisonTests.cs
@@ -138,7 +138,7 @@ public class CorpusSensorComparisonTests
 
         Assert.Contains(
             regressions,
-            regression => regression == "fidelity oracle differs (baseline compile-back, current return-to-sender)");
+            regression => regression == "fidelity oracle differs (baseline compile-back, current rts-parity)");
         string report = CorpusSensor.QualityMetricChangesForTesting(baseline, current);
         Assert.Contains("Fidelity exact (oracle differs)", report);
     }
--- a/src/ILInspector.Decompiler.Tests/CorpusSensorComparisonTests.cs
+++ b/src/ILInspector.Decompiler.Tests/CorpusSensorComparisonTests.cs
@@ -114,6 +114,115 @@ public class CorpusSensorComparisonTests
     }
 
     [Fact]
+    public void Compare_RejectsFidelityOracleMismatch()
+    {
+        var baseline = Snapshot(
+            totalMethods: 1,
+            fullyRaisedMethods: 1,
+            fullyRaisedBasisPoints: 10_000,
+            pinnedMethods: FidelityMethods(("One", "Exact")),
+            fidelityCompileCap: 1,
+            fidelityCheckedMethods: 1,
+            fidelityExactMethods: 1);
+        var current = Snapshot(
+            totalMethods: 1,
+            fullyRaisedMethods: 1,
+            fullyRaisedBasisPoints: 10_000,
+            pinnedMethods: FidelityMethods(("One", "Exact")),
+            fidelityCompileCap: 1,
+            fidelityCheckedMethods: 1,
+            fidelityExactMethods: 1,
+            fidelityOracle: CorpusFidelityOracle.ReturnToSender);
+
+        var regressions = CorpusSensor.Compare(baseline, current, [], gateAggregateRates: false);
+
+        Assert.Contains(
+            regressions,
+            regression => regression == "fidelity oracle differs (baseline compile-back, current return-to-sender)");
+        string report = CorpusSensor.QualityMetricChangesForTesting(baseline, current);
+        Assert.Contains("Fidelity exact (oracle differs)", report);
+    }
+
+    [Fact]
+    public void AlignReturnToSenderResults_ReportsUnavailableTarget()
+    {
+        var target = new FidelityCheck.CompileBackResult(
+            "Fixture",
+            "Method",
+            1,
+            "(corelib:System.Int32) -> corelib:System.Int32",
+            FidelityCheck.CompileBackStatus.Exact,
+            "ldarg.1 ret",
+            "ldarg.1 ret",
+            Detail: null);
+
+        var result = Assert.Single(
+            CorpusSensor.AlignReturnToSenderResultsForTesting(
+                [target],
+                Array.Empty<ReturnToSender.Result>()));
+
+        Assert.Equal(FidelityCheck.CompileBackStatus.ContextFail, result.Status);
+        Assert.Equal("return-to-sender-target-unavailable", result.Detail);
+        Assert.Equal("return-to-sender", result.CaptureDetail);
+        var buckets = FidelityCheck.SummarizeFailures(
+            [result],
+            FidelityCheck.CompileBackStatus.ContextFail);
+        Assert.Equal(1, buckets["return-to-sender target unavailable"].Count);
+    }
+
+    [Fact]
+    public void SummarizeReturnToSenderParity_ClassifiesRescuedSameAndWorse()
+    {
+        var exact = CompileBackResult("Exact", FidelityCheck.CompileBackStatus.Exact);
+        var rescued = CompileBackResult("Rescued", FidelityCheck.CompileBackStatus.OpcodeDiff);
+        var worse = CompileBackResult("Worse", FidelityCheck.CompileBackStatus.Exact);
+
+        var parity = CorpusSensor.SummarizeReturnToSenderParityForTesting(
+            [exact, rescued, worse],
+            [
+                exact,
+                rescued with { Status = FidelityCheck.CompileBackStatus.Exact },
+                worse with { Status = FidelityCheck.CompileBackStatus.OpcodeDiff },
+            ]);
+
+        Assert.Equal(1, parity.RescuedMethods);
+        Assert.Equal(1, parity.SameMethods);
+        Assert.Equal(1, parity.WorseMethods);
+        Assert.Equal(3, parity.ComparedMethods);
+    }
+
+    [Fact]
+    public void Compare_GatesReturnToSenderParityWhenSampleMatches()
+    {
+        var baseline = Snapshot(
+            totalMethods: 1,
+            fullyRaisedMethods: 1,
+            fullyRaisedBasisPoints: 10_000,
+            pinnedMethods: FidelityMethods(("One", "Exact")),
+            fidelityCompileCap: 1,
+            fidelityCheckedMethods: 1,
+            fidelityExactMethods: 1,
+            fidelityOracle: CorpusFidelityOracle.ReturnToSender,
+            returnToSenderParity: new ReturnToSenderParityMetrics(0, 1, 0));
+        var current = Snapshot(
+            totalMethods: 1,
+            fullyRaisedMethods: 1,
+            fullyRaisedBasisPoints: 10_000,
+            pinnedMethods: FidelityMethods(("One", "OpcodeDiff")),
+            fidelityCompileCap: 1,
+            fidelityCheckedMethods: 1,
+            fidelityOpcodeDiffMethods: 1,
+            fidelityOracle: CorpusFidelityOracle.ReturnToSender,
+            returnToSenderParity: new ReturnToSenderParityMetrics(0, 0, 1));
+
+        var regressions = CorpusSensor.Compare(baseline, current, [], gateAggregateRates: false);
+
+        Assert.Contains(
+            regressions,
+            regression => regression.StartsWith("RTS parity worse methods increased", StringComparison.Ordinal));
+    }
+
+    [Fact]
     public void Compare_DoesNotGateSemanticCountsWhenPinnedSamplesDiffer()
     {
         var baseline = Snapshot(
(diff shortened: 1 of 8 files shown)


richlander:

Exact reviewed head: `5a9ec6e2`

- **Claude Opus 4.8:** CLEAN
- **Gemini 3.1 Pro:** CLEAN

Both reviewers independently verified that RTS evaluates and preserves the exact
compile-back checked population, aligns type/method/overload/signature
identities, emits missing targets as explicit `ContextFail` rows, classifies
rescued/same/worse transitions correctly, rejects cross-oracle comparisons,
gates parity on identical method samples, preserves legacy snapshot/default
compile-back behavior, executes real RTS/provider paths after #2670, and remains
harness-only.

### Reconciled non-actions

- #2678 tracks the constructor `Exact` -> `OpcodeDiff` rows surfaced by the new
  oracle. They reproduce identically before and after #2670; this PR correctly
  exposes them rather than masking them.
- #2679 tracks pre-existing parallel cap-selection nondeterminism. This PR's
  comparison and parity gates require identical method samples, so incompatible
  runs are reported ungated instead of compared.
- #2685 tracks the sole full-suite failure. It reproduces unchanged on exact
  `origin/main`; this PR does not touch that test or rendering surface.
- Opus noted that a pre-existing RTS-native context-failure row can lose its
  native detail when its empty signature does not align. The replacement remains
  an explicit, stably bucketed `ContextFail`, so there is no success masking or
  blocking defect.

No resolution commit applies: neither exact-head review found an actionable PR
defect, and the disclosed follow-up work is independently tracked above.

richlander:
Ready to merge

richlander:
## Review — useful parity mode, but the compile-back floor contaminates the very verdict this PR exists to measure. Gate on un-blending it.

The comparison discipline is right where it's cross-snapshot: oracle-mismatched baselines hard-fail (`fidelity oracle differs`), all fidelity gates are guarded on same-oracle, card rows get `(oracle differs)` labels, and old-format baselines correctly deserialize to compile-back so cross-oracle diffs trip the failure. Unavailable targets fail closed as `ContextFail` with their own bucket, and missing/failed rows count as `worse` in parity. Good bones.

**The gate: `CompileBackTargets` unconditionally ends with `ApplyCompileBackFloor`, which *rewrites the status* — so in RTS mode, the lane never reports its own verdict.** Consequences, all traced:

- An RTS closure failure with usable payload is silently re-verdicted to compile-back's `Exact`/`OpcodeDiff`. The aggregate `FidelitySensorMetrics` counts it as an RTS success with **no floored-row counter anywhere in snapshot metrics** — provenance survives only in per-method `CaptureDetail`.
- **Parity is corrupted in both directions**: reference=`OpcodeDiff`, RTS fails, floor=`Exact` → counted *"rescued"* — a rescue produced by compile-back, credited to RTS. Reference=`Exact`, RTS fails, floor=`Exact` → *"same"* — an RTS regression completely masked from the `worse` gate, which is the one gate this mode adds.
- The snapshot's headline fidelity numbers under `FidelityOracle: return-to-sender` are therefore floor-blended, not an RTS verdict.

This is the exact invariant #2673 wrote down — "the moment [another oracle's] evidence changes a compile-back outcome rather than sitting beside it, the oracle is contaminated" — violated in mirror image: here compile-back evidence changes RTS outcomes. And the fix is already in the codebase: the getter-census path calls with `applyCompileBackFloor: false`. **Corpus mode should do the same** — the compile-back result is already held separately as the parity reference, so the floor adds nothing here except contamination. Native RTS statuses feed the metrics and parity; if a blended convenience view is wanted, add it as a separate labeled row with a floored-row count. One test pinning "RTS failure + floor-rescuable payload → parity `worse`, not `rescued`" closes the class.

Three more, non-gating:

1. **One bad assembly aborts the whole snapshot**: `EvaluateReturnToSender` calls `CompileBackTargets` with no try/catch, while `FidelityCheck.Evaluate` swallows per-assembly open failures and RTS's own `Run` catches these exact exception types. Per-assembly catch + a visible skipped-assembly line matches the sensor's existing posture.
2. **Name the mode what it is.** Selection runs compile-back first and keeps only its `Exact`/`OpcodeDiff` sample, so this can demonstrate RTS *rescues on compile-back's population*, never RTS *coverage* wins. The README says this honestly ("a parity population"); the flag value `return-to-sender` and snapshot field oversell it. `rts-parity` (or a body sentence in the snapshot) keeps future readers from misreading headline numbers. Related quiet shift: in RTS mode `NotFullMethods` is structurally 0 and the failure buckets describe only the selected population — same-named report fields, different semantics; worth a doc line.
3. The tolerance-0 parity gate silently disarms whenever the nondeterministic parallel sampler produces a different population (only a `(sampling differs)` card label). Consistent with the existing fidelity-gate philosophy, but for a zero-tolerance gate it means the gate rarely bites off-baseline — if parity becomes a CI signal, the sampler needs a deterministic mode first.

The direction is right — a standing parity lane between the two oracles is exactly what #2673's triangulation needs, and the cross-snapshot discipline shows the lessons landing. But merge this after the floor is out of the RTS lane's verdict path; as-is, the mode's headline numbers measure "RTS-or-compile-back, whichever succeeded," which is neither oracle.

---

richlander:
Resolved the review gate in `9ef22560`.

- Corpus mode now calls `CompileBackTargets(..., applyCompileBackFloor: false)`.
  Native RTS statuses exclusively feed headline metrics and rescued/same/worse
  parity; compile-back remains separate reference evidence.
- Other RTS callers retain the existing floor-enabled default.
- Added a real attribute-shell regression proving a floor-rescuable native RTS
  failure remains a failure and counts as `worse`, never `same` or `rescued`.
- Canonicalized the mode and snapshot identity to `rts-parity`; the existing
  `return-to-sender` and `rts` CLI spellings remain aliases.
- Added per-assembly RTS parity fault isolation with a visible diagnostic.
- Documented that this is a compile-back-selected parity population, not
  independent RTS coverage, and that `NotFullMethods` is structurally zero.

Updated evidence:

- focused tests: 115/115;
- fixed corpus, sequential cap 3: 31 Exact / 5 OpcodeDiff, 0 rescued / 36 same /
  0 worse, all 36 rows native;
- deterministic NuGet.Versioning cap 500: 48 Exact / 21 OpcodeDiff /
  36 RecompileFail, 0 rescued / 48 same / 57 worse, all 105 rows native.

The deep probe is the expected honesty change: 36 floor-hidden failures and all
57 native regressions are now visible. Parallel sample determinism remains
separate work under #2679.

The earlier `Ready to merge` status is withdrawn pending two clean exact-head
re-reviews of `9ef22560`.

richlander:
## Follow-up @ `9ef22560` — gate cleared; the lane now reports its own verdict

- **The floor is out of the corpus lane**: `EvaluateReturnToSender` calls `CompileBackTargets(..., applyCompileBackFloor: false)`, so native RTS statuses feed the metrics and parity. The compile-back reference still exists where it belongs — as the parity baseline — and nowhere else.
- **The exact contamination case I specified is pinned**: the new end-to-end test drives a target where RTS fails, proves the floored call *would* have rescued it (`UsedCompileBackFloor` true), then asserts the native run stays failed and parity classifies it **`Worse: 1`, `Rescued: 0`, `Same: 0`**. The rescue-credited-to-the-wrong-oracle bug and the masked-regression bug are both now unrepresentable without failing this test.
- **The mode is named what it is**: `rts-parity` (with the old spellings as aliases), threaded through the snapshot field, the oracle-differs failure message, and a README section that describes the transition semantics honestly.
- **The crash guard landed too**: a bad corpus assembly now logs `RTS parity skipped {path}` and degrades per-assembly instead of aborting the whole snapshot.

That's all four items, including both non-gating ones. The sampling-determinism note stays as the known caveat if parity ever becomes a CI gate — consistent with the existing fidelity-gate philosophy, so nothing further from me. Endorse; safe to merge.

---

richlander:
Rebased onto current `origin/main` and resolved the RTS conflict in `f353a139` by preserving main's new source-index/fault-isolation overloads while adding the explicit no-floor corpus path.

Post-rebase evidence: solution build passes; focused RTS/corpus tests 117/117; fixture catalogs 80/80; generated catalog 16/16; fixed corpus remains 31 Exact / 5 OpcodeDiff with parity 0 rescued / 36 same / 0 worse; deterministic NuGet probe remains 48 Exact / 21 OpcodeDiff / 36 RecompileFail with 57 worse. Exact-head reviews restarted.

richlander:
## Rebase check @ `f353a13` — clean; endorsement carries over unchanged

Verified the rebase over the merged #2686/#2683/#2682 stack point-by-point:

- The floor opt-out survived at the corpus call site (`CompileBackTargets(..., applyCompileBackFloor: false)`), and the parameterized overload chain reconciled cleanly with #2686's `FaultIsolation` addition to `Result` — the two features compose rather than collide, and the getter-census path correctly keeps its own `applyCompileBackFloor: false`.
- `rts-parity` naming is intact end-to-end, now including a `JsonStringEnumMemberName("rts-parity")` on the snapshot enum — the serialized field matches the CLI vocabulary.
- The per-assembly crash guard (`RTS parity skipped {path}`) survived, and `CorpusSensor.cs` at head is content-identical to the version I verified at `9ef22560`.

Nothing new to review; safe to merge.

---

richlander:
## Final exact-head review reconciliation

Merged head `f353a139` received two clean isolated reviews:

- **Claude Opus 4.8:** CLEAN — verified native no-floor parity, the combined source-index overload graph, 117 focused tests, 96 catalog tests, JSON/cross-mode behavior, and docs.
- **Gemini 3.1 Pro:** CLEAN — independently verified floor isolation, identity/failure-closed alignment, targeted exception visibility, snapshot compatibility, provider stability, and the same green test/build surface.

Both reviewers agreed the rebase preserved #2686 source-body fault isolation while keeping corpus parity native. No actionable findings or resolution commit apply. Opus noted two non-blocking pre-existing/low-risk observations: the `InvalidOperationException` skip filter is somewhat broad, and malformed PE input can fail earlier in oracle-independent completeness analysis.

PR #2676, opened 2026-07-12
Document Finding consumer adoption

richlander:
## Summary

- define the consumer-side Finding adoption contract: cross-validation, retained acquisition outcomes, replacement, fail-visible envelope handling, native transition cases, and quality-gate integrity
- anchor every rule to merged implementations from #2628 through #2671
- link the new design note from the repository overview

Follow-up to #2662.

## Validation

- `markdownlint` on changed Markdown
- documentation-only; no measured behavior change and no adversarial review required

### DIFF
--- a/docs/design/finding-adoption.md
+++ b/docs/design/finding-adoption.md
@@ -1,0 +1,155 @@
+# Finding Adoption
+
+A Finding producer is not adopted when a consumer merely calls it. Adoption is
+complete when the consumer retains the producer's outcome semantics, replaces
+the bespoke path that the producer supersedes, and exposes the producer's
+native cases without disguising them as legacy classifications.
+
+This document is the consumer-side companion to
+[Finding Producer Design](finding-producers.md). The observation and transition
+vocabulary comes from [Finding Nomenclature](finding-nomenclature.md), while
+identity, order, and provenance follow
+[Finding Coordinates](finding-coordinates.md).
+
+## 1. Cross-validate dual representations
+
+**Rule.** A migration may temporarily retain a legacy inventory for display or
+compatibility, but the Finding census must not be decorative. Build both
+representations from the same acquisition and cross-validate every consumed
+legacy item against a Finding observation using producer-owned identity and the
+domain fields needed to disambiguate collisions. A count comparison is not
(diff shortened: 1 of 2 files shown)


richlander:
## Review — the missing fourth doc, and it improves on the proposal

This realizes the `finding-adoption.md` proposal from #2662 faithfully — six rules, each anchored to a merged model rather than an ideal — and sharpens it in several places worth calling out:

- **The opening definition is the best sentence in the doc**: adoption is complete when the consumer "retains the producer's outcome semantics, replaces the bespoke path that the producer supersedes, and exposes the producer's native cases without disguising them as legacy classifications." That's the whole doctrine in one test.
- **Rule 1 adds two things my sketch lacked**: "a count comparison is not enough" (the exact shortcut a lazy cross-validation would take), and the explicit framing that the dual-representation period *is the proof*, not a license for two permanent lanes — which pre-empts the accretion failure mode rule 3 then forbids.
- **Rule 4 states both legal postures precisely** — visible degraded row when `Failed` is reachable, throw-with-stated-invariant when provably not — matching how #2643 and #2641 actually diverged, rather than mandating one.
- **Rule 6's docket rationale names the escape hatch** ("a regression can escape by falling into an excluded failure bucket") that makes checkability pins non-optional rather than nice-to-have.
- **The closing "Adoption review" checklist** turns the doc into a review rubric — which is what makes it cheap to enforce on the next consumer PR instead of re-litigating in threads.

I verified every model citation against the merged PRs: #2662's correspondence check, #2628's failure retention, #2638/#2640's replacements, #2643's throw-with-invariant, #2641's fallback rows, the #2642→#2661→#2671 lens lineage, #2649's identity-set sampling, #2651's checkability pins — all accurate as described. File-level source links (rather than line links) are the right durability trade.

One optional addition for a future revision, not this PR: a seventh rule or a sentence under rule 3 covering the *retained-comparison* pattern (#2661/#2671's opt-in `Retain*Comparisons` lists) — when a confirmation lens needs `Present` pairs that produce no change rows, retention is opt-in and the third consumer triggers generalization. That pattern now has two instances and will recur.

Endorse — merge it and start holding PRs to the checklist.

---

Issue #2575, opened 2026-07-08
Harden TypeNodeProvider (tree signature provider) against nested-TypeSpec StackOverflow (#2490 follow-up)


richlander:
Follow-up from #2574 (which completed the string-producing `SignatureDecoder` guard surface for #2490).

## Gap

`ILInspector.Metadata.TypeNodeProvider` is a second `ISignatureTypeProvider` (produces `TypeNode` trees, used by `ApiSurfaceExtractor` for field/method/property surface). Its `GetTypeFromSpecification` re-decodes nested TypeSpecs **unguarded**:

```csharp
public TypeNode GetTypeFromSpecification(MetadataReader reader, GenericContext? context, TypeSpecificationHandle handle, byte rawTypeKind)
{
    var typeSpec = reader.GetTypeSpecification(handle);
    return typeSpec.DecodeSignature(this, context);   // nested cmod TypeSpec cycle → uncatchable SO
}
```

This is the same cross-handle TypeSpec cycle / deep-blob vector fixed for `SignatureDecoder` in #2574 and `TypeRefDecoder` in #2489/#2550, but on a **different provider** (`this` = `TypeNodeProvider`, so the `SignatureDecoder` internal guard does not cover it).

Additionally, the three top-level `TypeNodeProvider.Instance` decodes in `ApiSurfaceExtractor` (field ~571, method ~746, property ~1284) need a `SignatureBlobGuard` prescan for the single-blob deep-nest / huge-count vector.

## Fix pattern (mirror #2574)

1. Internal guard on `TypeNodeProvider.GetTypeFromSpecification`: `[ThreadStatic]` depth (256) + per-blob length (1024) + cumulative (4096) caps, then `SignatureBlobGuard.IsSafeToDecode(..., Kind.TypeSpecification)`, fail closed to a placeholder `TypeNode` (e.g. `NamedTypeNode("object", isRef: true)`).
2. A `GuardedTypeNode` helper (internal, mirroring `GuardedSignatureText`) with field/method/property wrappers returning safe `TypeNode` / `MethodSignature<TypeNode>` fallbacks; route the `ApiSurfaceExtractor` top-level sites through it.
3. Regression tests mirroring `SignatureDecoderRecursionTests` (self-referential cmod TypeSpec + 100k-SZARRAY) through the `TypeNodeProvider` path.

[...]

Issue #2490, opened 2026-07-06
Harden all signature-decode entry points against over-deep blobs (SRM native-stack StackOverflow beyond TypeRefDecoder)


richlander:
## Summary

PR #2489 (fixing #2159) hardened `TypeRefDecoder` against **cyclic** malformed
metadata (self-referential resolution scope / nested type / TypeSpec) and added a
blob-length cap to `GetTypeFromSpecification` for **over-deep single TypeSpecification
blobs**. During adversarial review, two independent reviewers (GPT-5.5 and Gemini 3.1
Pro) verified a related, still-open crash vector that PR #2489 deliberately scoped out:

**SRM's own `SignatureDecoder<T>.DecodeType` recurses on the native stack** once per
nested structural element (`PTR`/`SZARRAY`/`BYREF`/`PINNED`/`GENERICINST`/`CMOD`/`ARRAY`)
within a *single* signature blob, **before any provider callback runs**. A sufficiently
deep blob overflows the stack with an **uncatchable `StackOverflowException`**, bypassing
the provider-level `s_recursionDepth` guard entirely.

Verified repros (both aborted the process, exit code 134):
- Gemini: TypeSpec blob of 50,000 × `ELEMENT_TYPE_PTR` (`0x0f`) + `I4` — ~16k `DecodeType` frames.
- GPT-5.5: TypeSpec blob of 100,000 × `ELEMENT_TYPE_SZARRAY` (`0x1d`) + `I4` — ~87k frames.

PR #2489 closes the vector **through `TypeRefDecoder.GetTypeFromSpecification`** (the method
both reviewers reproduced through) with a blob-length cap. This issue tracks the **tool-wide**
remainder.

## Scope of remaining work

The other signature-decode entry points that use `TypeRefDecoder.Instance` as the provider
are still exposed, because they call SRM decode helpers directly (not via
`GetTypeFromSpecification`):

- `src/ILInspector.Analysis/MemberResolver.cs` — `DecodeSignature`, `DecodeMethodSignature`
  (method / member-ref / method-spec).
- `src/ILInspector.Analysis/LibraryBodyIndex.cs` — `DecodeSignature`, `DecodeLocalSignature`,
  `DecodeMethodSignature`.
- `src/ILInspector.Analysis/LeakTriage.cs` — `DecodeSignature`.
- The corresponding Decompiler decode call sites.

## Why a length cap alone is insufficient here

[...]

PR #2574, opened 2026-07-08
Close SignatureDecoder guard bypasses: nested TypeSpec + MetadataDeclarationQuery delegation (#2490)

richlander:
## Why

Follow-up to #2566 (which merged the composer-only `SignatureDecoder` guard). Two-model adversarial review of #2566 found the guard was **bypassable two ways**, so `main` currently still has an uncatchable `StackOverflowException` reachable on malformed metadata through the string-producing `SignatureDecoder`:

1. **Composer → MetadataDeclarationQuery delegation** (converged Critical / GPT Issue 1): the compile-back / type-source composers delegate to `MetadataDeclarationQuery`, which decoded `SignatureDecoder.Instance` **unguarded** — so a deep property/method sig bypasses the composer guard entirely.
2. **Nested TypeSpec via cmod** (GPT Issue 2): `SignatureDecoder.GetTypeFromSpecification` re-decodes nested TypeSpecs reached by a custom-modifier handle **unguarded**. A shallow (guard-passing) outer blob whose cmod references a deep/cyclic TypeSpec re-enters SRM's native-recursive decoder → uncatchable SO. This is the string-decoder parallel of #2489 / #2550.

## What

- **Relocate `SignatureBlobGuard`** `ILInspector.Metadata` → `ILInspector.MetadataPrimitives` (namespace unchanged, pure SRM) so the lowest layer — `SignatureDecoder` + `TypeResolver` — can reach it. No consumer churn.
- **Harden `SignatureDecoder.GetTypeFromSpecification`** with the #2489/#2550 caps: per-blob length `1024`, `[ThreadStatic]` cumulative `4096`, re-entry depth `256`, then `SignatureBlobGuard`; fail closed to `object`. Closes the nested-cycle vector for **every** `SignatureDecoder.Instance` caller at once.
- **Guard `TypeResolver.GetTypeNameFromSpecification`** (top-level TypeSpec decode).
- **Route `MetadataDeclarationQuery` (5 sites)** through `GuardedSignatureText` (fixes finding 1).
[...]

PR #2677, opened 2026-07-12
Guard Metadata string signature decoding

richlander:
Advances #2490.

## Change

**Conclusion:** **PASS** — Metadata's string-producing signature decoder now rejects
top-level structural recursion/count hazards before entering SRM and bounds nested
TypeSpec re-entry without changing valid CoreLib output.

This is slice A of the fixed mechanism matrix posted on #2490:

- route every top-level Metadata `SignatureDecoder` decode through
  `GuardedSignatureText`, which applies the iterative `SignatureBlobGuard` prescan;
- bound nested TypeSpec depth, per-blob length, and cumulative bytes in
  `SignatureDecoder.GetTypeFromSpecification`;
- move `SignatureBlobGuard` to MetadataPrimitives so both layers can use it, while
  preserving its old `ILInspector.Metadata` assembly identity with a type forwarder;
- add process-isolated deep-signature/cyclic-TypeSpec regressions and a source census
  that rejects new raw Metadata string-decoder entry points.

## Frozen scope

This PR covers only:

1. SRM native recursion/count preallocation in top-level Metadata string signatures.
2. Cross-handle nested TypeSpec re-entry in the same string provider.
3. Compatibility of the moved `SignatureBlobGuard` public type identity.

Different mechanisms remain separate follow-ups: other
`ISignatureTypeProvider` implementations (#2575), TypeRef/TypeDef graph cycles,
name/arity/output amplification, Analysis/Decompiler provider closure, and
custom-attribute decoding.

## Evidence

| Check | Result |
| --- | --- |
| Solution build | Pass |
| Metadata tests | 452 passed |
| Product tests | 1,768 passed, 10 skipped |
| Analysis tests | 430 passed |
| Decompiler fast suite | 2,333 passed |
| CoreLib valid-output A/B | Byte-identical on all three commands |

Exact merge base: `2b3db7ee9ad405031885fe71872cde7ffe64136a`  
Exact reviewed head: `f774dd5de5be74a6ead9b7a7d04e314297385208`

| CoreLib command | Base/head SHA-256 |
| --- | --- |
| `type System.String --json -v:d` | `0cb3c8d6e7ed6f354839b27853102ee96838980d8170e50ab96b2d8166b8f5e7` |
| `library --json` | `6e48206952b47467883b424f53419ac27d81acf1a609e2df6986e6fc2ab8efc9` |
| `extensions System.String --json` | `cea7b70da91baaf75da00a3f95f8f5542a76bafb341fb7fc4fc9ef95469f414f` |

The decompiler quality card is not applicable: this changes malformed-metadata
containment rather than raising, structuring, or printing, and valid product output
is byte-identical.

## Adversarial review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | CLEAN | Verified gateway closure, iterative grammar bounds, balanced TypeSpec state, child-process regressions, and A/B on the exact head. |
| Gemini 3.1 Pro | CLEAN | Verified all A1/A2/A3 mechanisms, census closure, cyclic/deep fixtures, type forwarding, and A/B on the exact head. |

Both reviewers independently identified only the already-scoped, non-blocking
provider-closure follow-up tracked by #2575. No resolution commit applies because
that mechanism is deliberately excluded from this slice.

### DIFF
--- a/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
+++ b/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
@@ -1455,7 +1455,7 @@ public static class ApiSurfaceExtractor
     private static string? GetFirstParameterType(MetadataReader reader, TypeDefinition typeDef, MethodDefinition method)
     {
         var context = GenericContext.ForMethod(reader, typeDef, method);
-        var signature = method.DecodeSignature(SignatureDecoder.Instance, context);
+        var signature = GuardedSignatureText.MethodText(reader, method, context);
         return signature.ParameterTypes.Length > 0 ? signature.ParameterTypes[0] : null;
     }
 
--- a/src/ILInspector.Metadata/EcosystemIntegrationScanner.cs
+++ b/src/ILInspector.Metadata/EcosystemIntegrationScanner.cs
@@ -614,7 +614,7 @@ public static class EcosystemIntegrationScanner
             MethodSignature<string> signature;
             try
             {
-                signature = method.DecodeSignature(SignatureDecoder.Instance, context);
+                signature = GuardedSignatureText.MethodText(reader, method, context);
             }
             catch (BadImageFormatException)
             {
--- a/src/ILInspector.Metadata/ExtensionMethodScanner.cs
+++ b/src/ILInspector.Metadata/ExtensionMethodScanner.cs
@@ -125,7 +125,7 @@ public static class ExtensionMethodScanner
                 // Decode signature to get first parameter type
                 {
                     var context = GenericContext.ForMethod(reader, typeDef, method);
-                    var signature = method.DecodeSignature(SignatureDecoder.Instance, context);
+                    var signature = GuardedSignatureText.MethodText(reader, method, context);
                     if (signature.ParameterTypes.Length == 0) continue;
 
                     var extendedType = signature.ParameterTypes[0];
@@ -222,7 +222,7 @@ public static class ExtensionMethodScanner
 
                 {
                     var context = GenericContext.ForMethod(reader, typeDef, method);
-                    var signature = method.DecodeSignature(SignatureDecoder.Instance, context);
+                    var signature = GuardedSignatureText.MethodText(reader, method, context);
                     if (signature.ParameterTypes.Length == 0) continue;
 
                     var extendedType = signature.ParameterTypes[0];
@@ -331,7 +331,7 @@ public static class ExtensionMethodScanner
 
                     try
                     {
-                        var sig = prop.DecodeSignature(SignatureDecoder.Instance, context);
+                        var sig = GuardedSignatureText.PropertyText(reader, prop, context);
                         var propType = UnwrapAsyncType(sig.ReturnType);
 
                         if (!string.IsNullOrEmpty(propType) && !IsPrimitiveType(propType) && visited.Add(propType))
@@ -360,7 +360,7 @@ public static class ExtensionMethodScanner
                     try
                     {
                         var methodContext = GenericContext.ForMethod(reader, typeDef, method);
-                        var sig = method.DecodeSignature(SignatureDecoder.Instance, methodContext);
+                        var sig = GuardedSignatureText.MethodText(reader, method, methodContext);
                         var retType = UnwrapAsyncType(sig.ReturnType);
 
                         if (!string.IsNullOrEmpty(retType) && retType != "void" &&
@@ -487,15 +487,17 @@ public static class ExtensionMethodScanner
                     continue;
 
                 var displayContext = GenericContext.ForType(reader, markerType);
-                var markerSignature = markerMethod.DecodeSignature(
-                    SignatureDecoder.Instance,
+                var markerSignature = GuardedSignatureText.MethodText(
+                    reader,
+                    markerMethod,
                     displayContext);
                 if (markerSignature.ParameterTypes.Length != 1)
                     throw new BadImageFormatException(
                         "An extension marker must have exactly one receiver parameter.");
 
-                var propertySignature = property.DecodeSignature(
-                    SignatureDecoder.Instance,
+                var propertySignature = GuardedSignatureText.PropertyText(
+                    reader,
+                    property,
                     displayContext);
                 string propertyName = reader.GetString(property.Name);
                 string parameterText = propertySignature.ParameterTypes.Length == 0
@@ -563,7 +565,7 @@ public static class ExtensionMethodScanner
         GenericContext? context)
     {
         string name = reader.GetString(method.Name);
-        var signature = method.DecodeSignature(SignatureDecoder.Instance, context);
+        var signature = GuardedSignatureText.MethodText(reader, method, context);
 
         // First parameter is the extension receiver, rendered with a 'this ' prefix.
         return SignatureRenderer.RenderDecodedSignature(reader, method, name, signature, extensionThis: true);
--- a/src/ILInspector.Metadata/GuardedSignatureText.cs
+++ b/src/ILInspector.Metadata/GuardedSignatureText.cs
@@ -1,0 +1,71 @@
+using System.Collections.Immutable;
+using System.Reflection.Metadata;
+
+namespace ILInspector.Metadata;
+
+/// <summary>
+/// Top-level prescan gateway for string-producing signature decodes in Metadata.
+/// Nested TypeSpec re-entry is guarded by <see cref="SignatureDecoder"/>.
+/// </summary>
+internal static class GuardedSignatureText
+{
+    const string Unresolved = "object";
+    static readonly MethodSignature<string> UnresolvedMethod = new(default, Unresolved, 0, 0, []);
+
+    public static MethodSignature<string> MethodText(
+        MetadataReader reader,
+        MethodDefinition method,
+        GenericContext? context)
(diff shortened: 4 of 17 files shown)


richlander:
Ready to merge

Merge-blocking validation is complete on exact head `f774dd5d`: CI is green, CoreLib A/B is byte-identical, and Claude Opus 4.8 plus Gemini 3.1 Pro both returned CLEAN. The provider-closure observation is non-blocking and remains tracked separately in #2575.

richlander:
## Review — the frozen-scope slice is the right process answer to #2574; three asks, one about a crash that's still live

First, the process call deserves endorsement on its own: #2574 died of a moving claim — six adversarial rounds, each "clean except one new Critical," because the scope kept absorbing every adjacent mechanism the reviews surfaced. Closing it and re-cutting a fixed mechanism matrix with this as slice A is the correct fix for *that* failure, and the slice discipline shows in the diff: one gateway (`GuardedSignatureText`, six fail-closed wrappers), one nested-re-entry guard (`TypeSpecGuard` on `SignatureDecoder` + `TypeResolver`), 22 call sites rerouted mechanically, and a census test to keep new raw decodes from sneaking in.

The test approach is the strongest part: the **process-isolated** crash regressions (child `dotnet` process, assert exit 0) genuinely prove stack-overflow containment in a way no in-process try/catch can — an uncatchable SO fails the outer test as exit 134. And the `MetadataDeclarationQuery` 100k-depth worker directly pins #2574's original finding.

Three asks, ranked:

**1. The known-Critical crash on `TypeNodeProvider` should be pinned in-repo before this merges, even though its fix is correctly deferred.** The census *allowlists* `TypeNodeProvider.cs`, whose `NameDecoder` path has the same nested-TypeSpec re-entry hole this PR fixes elsewhere — and #2574's round-6 history classified that hole Critical *with a working SO repro*, reachable from `ApiSurfaceExtractor`. Deferring the fix to #2575 is legitimate slice discipline; letting the reproduced crash live only in a closed PR's comment history is not. Add the repro as an explicitly-skipped test (skip reason: #2575) or a tracked known-failure entry now, so the deferred hole is greppable and #2575's completion criterion is executable. That converts "scoped follow-up" from a promise into a pin — the same standard #2651 set for docket entries.

**2. The degraded sentinel is invisible — `"object"` is a plausible lie.** Rejection renders as a legal-looking signature (`object M()`), indistinguishable from a genuine `System.Object` member, with no diagnostic, counter, or marker anywhere — and `UnresolvedMethod` carries `default(SignatureHeader)`, so a caller reading `Header.IsInstance` on a degraded value gets fabricated data. Contrast your own Analysis layer: `TypeRefDecoder` degrades to `TypeRef.Unsupported("type-specification recursion depth exceeded")` — a carrier of *why*. The Metadata layer deserves the same: an unambiguous sentinel (or at minimum a debug-visible counter) so a malformed assembly doesn't produce silently-wrong inspection output. Related inconsistency worth one sweep: rejection consequences are per-site accidental — most sites silently skip, `ILTokenResolver` renders `Name<>`, and `ExtensionMethodScanner` *throws* `BadImageFormatException` when the degraded zero-param marker signature trips its arity check. Fail-closed should be one designed contract, not three emergent ones.

**3. Two acceptance envelopes for the same operation.** A top-level TypeSpec decode through `TypeResolver.GetTypeNameFromSpecification` gets the full `TypeSpecGuard` (1024-byte cap, depth, cumulative), while the same decode through `GuardedSignatureText.TypeSpecText` gets prescan only — a 2 KB wide-but-shallow TypeSpec succeeds via one door and degrades via the other. Also note the 1024-byte cap on the `TypeResolver` path is a **new hard limit on legal input** (previously unguarded; CoreLib's largest is 57 bytes, but the three-command CoreLib A/B can't falsify wide generic instantiations). Pick one envelope, and state the legal-input bound it implies. Smaller residue in the same family: `TypeSpecGuard` copy-pastes `TypeRefDecoder`'s constants and shape rather than sharing them (defensible — separate thread-statics per provider — but say so), and `TryEnter` itself can throw on a bogus handle row, so the guard's never-throw posture isn't total.

None of these reopen the slice's core: every path it touches is strictly safer than main, the containment proofs are real, and the census — while a tripwire rather than a proof — is the right cheap enforcement. Endorse with ask 1 addressed pre-merge (it's a test addition, not scope creep — the frozen-scope rule cuts against new *fixes*, not against pinning known holes), and 2–3 as the named follow-ups for the #2575/#2490 matrix.

---

richlander:
## Review feedback addressed

Thanks for the endorsement of the frozen-scope restart and the three ranked asks. Disposition below.

### Ask 1 — pin the known `TypeNodeProvider` crash before merge (accepted, done)

Added a **test-only** pin in `tests/ILInspector.Metadata.Tests/SignatureDecoderSafetyTests.cs` (commit `47de0f26`):

- `CyclicTypeSpecThroughApiSurfaceExtractorWorker` builds a synthetic assembly whose static field signature and a TypeSpec both carry a self-referential `CMOD_REQD → TypeSpec row 1` cycle, then calls `ApiSurfaceExtractor.Extract`, which routes through the still-unguarded `TypeNodeProvider`. Directly selected, it reproduces the known uncatchable `StackOverflowException` (child process aborts, exit 134).
- `CyclicTypeSpec_ThroughApiSurfaceExtractor_IsContainedInChildProcess` is `[Fact(Skip = "Known TypeNodeProvider crash tracked by #2575.")]`. It stays greppable and executable, so completing #2575 flips it from skipped to a passing containment test — the pin becomes the executable completion criterion, exactly as asked. This is a test addition only; no product path changed.

### Ask 2 — invisible degraded sentinel (`"object"`), and Ask 3 — two acceptance envelopes for the same TypeSpec (accepted as follow-ups, not in this slice)

Both are real and worth a designed fix, but they change the fail-closed *contract* rather than close a same-mechanism crash, so per the frozen-scope rule they do not widen slice A. Filed together as **#2689**, covering:

- an explicit, inspectable rejection result instead of a plausible `object` signature, with one designed fail-closed contract across field/method/property/method-spec/TypeSpec (replacing the current per-site skip / `Name<>` / `BadImageFormatException` divergence);
- reconciling the top-level TypeSpec acceptance envelope between `GuardedSignatureText.TypeSpecText` (prescan only) and `TypeResolver.GetTypeNameFromSpecification` (prescan + 1,024-byte / 4,096-byte caps), plus stating the legal-input bound the cap implies and adding wide-but-shallow boundary fixtures above 1,024 bytes;
- documenting or centralizing the shared cap constants that `TypeSpecGuard` currently copies from `TypeRefDecoder`, and the `TryEnter`-on-bogus-handle throw margin.

### Validation on the pinned head

- Rebased onto `origin/main` `96320e8a`; exact head `47de0f26`.
- `ILInspector.Metadata.Tests`: 462 passed, 1 skipped (the new pin).
- CoreLib A/B (`type System.String -v:d --json`, `library --json`, `extensions System.String --json`) byte-identical, base `96320e8a` vs head `47de0f26`.
- Re-running the two isolated adversarial reviews on exact head `47de0f26`; the added commit is test-only.

richlander:
## Adversarial re-review on pinned head — two CLEAN

Both reviewers re-ran on exact head `47de0f26` (base `96320e8a`), focused on the delta since the prior two clean reviews (a rebase plus the test-only pin commit).

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | CLEAN | `git diff b198896c..47de0f26` touches only the one test file; `src/*` diff empty. Suite shows SKIP + all pass, runner survives; directly-selected worker aborts exit 134 through `TypeNodeProvider.GetTypeFromSpecification` → `ApiSurfaceExtractor.Extract` (the #2575 path). A1/A2/A3 untouched. |
| Gemini 3.1 Pro | CLEAN | Confirmed test-only scope, guards untouched, skipped parent does not down the runner, worker reproduces the known crash (exit 134), workspace clean. |

No blocking findings. The reproduced crash is the deliberately-deferred #2575 path, not a slice-A regression.

richlander:
Ready to merge (updated)

Exact head `47de0f26` is fully validated: CI green (build + test), CoreLib A/B byte-identical vs base `96320e8a`, and two fresh CLEAN adversarial reviews (Claude Opus 4.8 + Gemini 3.1 Pro) on this exact head. Review ask 1 is resolved in-repo (skipped `#2575` crash pin); asks 2–3 are tracked as follow-up #2689. Not merging without your approval.

richlander:
## Follow-up @ `47de0f26` — the pin landed in the strongest form; safe to merge

`CyclicTypeSpec_ThroughApiSurfaceExtractor_IsContainedInChildProcess` is better than what I asked for:

- It's a **real repro, not a placeholder**: it synthesizes the cyclic-TypeSpec assembly with `MetadataBuilder` and drives it through `ApiSurfaceExtractor.Extract` — the actual reachable product path — under the same process-isolated worker harness as the live containment tests.
- The `Skip` reason names #2575, so the deferred hole is greppable and tracked in-repo rather than in closed-PR memory.
- Most importantly, **the worker stays executable**, so #2575's completion criterion is mechanical: delete the `Skip` attribute and the pin becomes a containment test. That's the docket-with-checkability-pin pattern (#2651) applied to a deferred crash — the fix can't be claimed without the test proving it.

Pre-merge ask met; the degraded-sentinel and envelope-unification items remain the named follow-ups for the #2490 matrix as agreed. Endorse — safe to merge.

---

PR #2633, opened 2026-07-10
Route RTS type plans through CSharpTypePrinter

richlander:
Fixes #2607

## Change

**Conclusion: PASS** — ReturnToSender owns closure/type-cluster planning without
carrying a parallel C# declaration renderer. `CSharpTypePrinter` composes exact
typed requests, and all declaration spelling flows through `CSharpFormatter`.

- Extend the CSharp request contract with structured full/stub bodies,
  independent property accessors, explicit constructor initializers, primary
  constructors, nested types, enums, delegates, and block namespaces.
- Move the compile-back planner from `ILInspector.Decompiler` into
  `tools/DecompilerHarness`.
- Retire `CompileBackTypeDeclaration`, `CompileBackMemberDeclaration`, and the
  Research forwarding facade.
- Keep typed RTS requirements as planner evidence while rendering separate
  `CSharpTypePrintRequest` values.

## Scope and safety

- **Root cause:** RTS duplicated declaration spelling and source composition in
  a compile-back-specific model inside product assemblies.
- **RTS boundary:** RTS owns closure growth, membership, synthesis,
  accessibility flattening, body-policy selection, accessor bodies, and
  constructor initializer shape.
- **CSharp boundary:** CSharp validates typed requests fail-closed and owns their
  legal spelling through `CSharpFormatter`; it does not invent missing member
  contents.
- **Research impact:** Research returns to a pure evidence aggregator; it no
  longer forwards compile-back shell composition.

## Evidence

| Check | Result |
| --- | --- |
| Solution build | Pass |
| `CSharpTypePrinterTests` | 47 passed |
| Metadata tests | 452 passed |
| RTS prototype + catalog tests | 176 passed |
| Markdown lint | Pass |
| Product render A/B vs. `de0337f4` | 89,065 evaluated; 0 changed, 0 added, 0 removed |
| RTS source A/B | 12/12 Ladder property-getter plans byte-identical |

The product render A/B used the explicit post-rebase merge base and the
14-assembly real-world corpus with 20 workers. It covers the shared Metadata
writer path, including the structured metadata-only default-parameter change.

## Owner feedback disposition

- Abstract members now reject `Full` and `Stub` policies instead of silently
  dropping requested bodies.
- Stub properties require explicit accessor bodies; unsupported field-like
  event body policies fail closed; primary-constructor chaining is an explicit
  planner-provided request shape.
- Type, member, property, delegate, identifier, generated-name, and namespace
  spelling route through the merged `CSharpFormatter` facade from #2626.
[...]

PR #2680, opened 2026-07-12
Finish C# spelling ownership and remove RTS transition alias

richlander:
## Change

**Conclusion: PASS** — C# declaration spelling now lives behind
`ILInspector.CSharp.CSharpFormatter`; Metadata retains structured facts and
narrow compatibility signature text only. The obsolete RTS
`TypeRequirements => Types` transition alias is removed.

- Moves `CSharpDeclarationWriter` and its focused tests from Metadata to CSharp
  and makes the implementation internal.
- Moves parameter declaration/list spelling to `CSharpFormatter` and removes
  `ApiParameter.Declaration` / `ApiSignature.ParameterDeclarationsSummary`.
- Removes the RTS `TypeRequirements => Types` transition alias and updates all
  consumers to the canonical `Types` property.
- Carries structured type accessibility through extraction and CSharp snapshots,
  replacing the type/delegate `public` hardcode.
- Hardens keyword spelling at the CSharp/legacy-compatibility boundary and keeps
  the two keyword tables synchronized.
- Evidence revision: `cf7b19328311ae765097f0753928bb3587aac8ed`

## Scope and safety

- **Layering:** Metadata does not depend on CSharp; production declaration
  composition now has one CSharp-owned public seam.
- **Product path:** SRM-only, NativeAOT-friendly, Roslyn-free, and does not load
  inspected assemblies.
- **Accessibility rendering:** declarations for non-public types and delegates
  selected by `--all` now render their actual accessibility instead of the
  incorrect hardcoded `public`.
- **JSON schema:** detailed type JSON now includes `accessibility` for extracted
  non-public types. Public and older serialized `ApiType` values remain null;
  CSharp rendering deliberately treats null as `public` for compatibility.
- **Keyword compatibility:** CSharp and Metadata's narrow legacy-signature
  escaper use the same conservative keyword set. Function-pointer syntax remains
  syntax, while genuine types named `delegate`, `readonly`, or `scoped` escape.
- **RTS boundary:** harness orchestration consumes `Types`; shared CSharp owns
  source spelling.

## Declaration-surface A/B

Targeted A/B compares merge base `96320e8a` with `cf7b1932` against the same
head-built `dotnet-inspect.Tests.dll` fixture.

```diff
-public class InternalTopLevelSurfaceFixture
+internal class InternalTopLevelSurfaceFixture
```

Detailed CLI JSON has the single expected schema addition:

```diff
   "metadata_name": "InternalTopLevelSurfaceFixture",
+  "accessibility": "internal",
   "kind": "class",
```

The `type --all --markdown -v:d` member surface is byte-identical
(`sha256 83454f51b184bc88ca11945621d0ed1f19c8403c7b14227e53a3353e9ff30df1`
on both sides). Focused type-printer tests cover the corresponding non-public
delegate declaration.

## Evidence

| Check | Result |
| --- | --- |
| `ILInspector.CSharp.Tests` | 90 passed |
| `ILInspector.Metadata.Tests` | 418 passed |
| `dotnet-inspect.Tests` | 1,777 passed, 10 tool-dependent skips |
| `ReturnToSenderPrototypeTests` | 96 passed |
| Render-text A/B | 89,065 methods, 0 changed, 0 added, 0 removed |

### Decompiler quality diff

| Semantic defects (sampling differs) | 54 (0.21%) → 0 (0.00%) | n/a |
| Fidelity opcode diffs (sampling differs) | 4 (11.11%) → 3 (8.33%) | n/a |
| Fidelity exact (sampling differs) | 32 (88.89%) → 33 (91.67%) | n/a |

Regressions:

The generated advisory is the documented cap mismatch: all pinned rate/count
gates are neutral, pass bugs remain zero, and the explicit merge-base render A/B
has no changed methods.

## Review

Earlier adversarial rounds found compatibility-signature keyword escaping,
quoted defaults, global keyword-type, and metadata-expression edge cases; all
were resolved with compiled fixtures and integration coverage. Owner feedback
on declaration evidence, keyword-policy skew, syntax/type-identity ambiguity,
and the null accessibility fallback is addressed by `cf7b1932`. Exact-head
signoff re-review is in progress.

### DIFF
--- a/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
@@ -1,6 +1,6 @@
 using ILInspector.Metadata;
 
-namespace ILInspector.Metadata.Tests;
+namespace ILInspector.CSharp.Tests;
 
 public sealed class CSharpDeclarationWriterTests
 {
--- a/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
@@ -92,13 +92,27 @@ public sealed class CSharpFormatterTests
                 DefaultValueText = "default"
             },
             new() { Type = "class", Name = "value" },
-            new() { Type = "System.Collections.Generic.List<class>", Name = "items" }
+            new() { Type = "System.Collections.Generic.List<class>", Name = "items" },
+            new() { Type = "delegate", Name = "delegateValue" },
+            new() { Type = "readonly", Name = "readonlyValue" },
+            new() { Type = "scoped", Name = "scopedValue" },
+            new() { Type = "delegate*<ref int, void>", Name = "callback" }
         };
 
         Assert.Equal(
-            "([System.Runtime.InteropServices.Optional] System.@event.MyClass @event = default, @class value, System.Collections.Generic.List<@class> items)",
+            "([System.Runtime.InteropServices.Optional] System.@event.MyClass @event = default, @class value, System.Collections.Generic.List<@class> items, @delegate delegateValue, @readonly readonlyValue, @scoped scopedValue, delegate*<ref int, void> callback)",
             CSharpFormatter.FormatParameterList(parameters));
     }
+
+    [Theory]
+    [InlineData("await")]
+    [InlineData("file")]
+    [InlineData("init")]
+    [InlineData("record")]
+    [InlineData("required")]
+    [InlineData("scoped")]
+    public void EscapesConservativeContextualKeywordSet(string identifier)
+        => Assert.Equal($"@{identifier}", CSharpFormatter.EscapeIdentifier(identifier));
 
     [Fact]
     public void FormatsDelegateWithStructuredAccessibility()
--- a/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
@@ -79,6 +79,55 @@ public sealed class CSharpFormatterTests
     }
 
     [Fact]
+    public void FormatsParameterListsWithAttributesDefaultsAndEscapedKeywords()
+    {
+        var parameters = new ApiParameter[]
+        {
+            new()
+            {
+                Attributes = ["System.Runtime.InteropServices.Optional"],
+                Type = "System.event.MyClass",
+                Name = "event",
+                HasDefault = true,
+                DefaultValueText = "default"
+            }
+        };
+
+        Assert.Equal(
+            "([System.Runtime.InteropServices.Optional] System.@event.MyClass @event = default)",
+            CSharpFormatter.FormatParameterList(parameters));
+    }
+
+    [Fact]
+    public void FormatsDelegateWithStructuredAccessibility()
+    {
+        var type = new ApiType
+        {
+            Namespace = "Samples",
+            Name = "Callback",
+            Kind = "delegate",
+            Accessibility = "private",
+            Members =
+            [
+                new ApiMember
+                {
+                    Name = "Invoke",
+                    Kind = "method",
+                    SignatureModel = new ApiSignature
+                    {
+                        ReturnType = "void",
+                        MemberName = "Invoke"
+                    }
+                }
+            ]
+        };
+
+        Assert.Equal(
+            "private delegate void Callback();",
+            new CSharpFormatter().FormatDelegate(type, type.Members.Single()));
+    }
+
+    [Fact]
     public void RejectsUndefinedPolicies()
     {
         Assert.Throws<ArgumentOutOfRangeException>(
--- a/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
@@ -90,11 +90,13 @@ public sealed class CSharpFormatterTests
                 Name = "event",
                 HasDefault = true,
                 DefaultValueText = "default"
-            }
+            },
+            new() { Type = "class", Name = "value" },
+            new() { Type = "System.Collections.Generic.List<class>", Name = "items" }
         };
 
         Assert.Equal(
-            "([System.Runtime.InteropServices.Optional] System.@event.MyClass @event = default)",
+            "([System.Runtime.InteropServices.Optional] System.@event.MyClass @event = default, @class value, System.Collections.Generic.List<@class> items)",
             CSharpFormatter.FormatParameterList(parameters));
     }
 
--- a/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
@@ -130,6 +130,17 @@ public sealed class CSharpFormatterTests
     }
 
     [Fact]
+    public void KnownIdentifierEscapingIsIdempotent()
+    {
+        Assert.Equal(
+            "System.Action<@event>",
+            CSharpFormatter.EscapeKnownIdentifiers("System.Action<@event>", ["event"]));
+        Assert.Equal(
+            "System.Action<@event>",
+            CSharpFormatter.EscapeKnownIdentifiers("System.Action<event>", ["event"]));
+    }
+
+    [Fact]
     public void RejectsUndefinedPolicies()
     {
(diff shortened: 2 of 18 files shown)


richlander:

Final reviewed head: `d520b6b945f1fd5a8ad011b9ad1b46755cfccdc9`.

| Reviewer | Final result | Evidence |
| --- | --- | --- |
| Claude Opus 4.8 | CLEAN | Exact-head build and CSharp/Metadata/CLI suites; keyword type positions, literal safety, layering, accessibility, and alias removal checked |
| Gemini 3.1 Pro | CLEAN | Exact-head producer-level attribute/default checks; idempotent complex-signature escaping; layering, accessibility, and runtime constraints checked |

Earlier rounds found real edge cases:

- Gemini found that localizing legacy `ApiMember.Signature` parameter spelling
  dropped qualified-keyword escaping. `a7b42d65` restored it with a compiled
  regression fixture.
- Gemini found malformed quoted string defaults and unescaped global keyword
  type identities. `42244d39` fixed the producers and CSharp type spelling.
- Full CLI validation then caught an `@@event` interaction with the existing
  generic-identifier pass. `1f89216c` made that pass idempotent and added direct
  plus integration coverage.
- Gemini found the same keyword-type problem in real enum-default and
  `System.Type`/enum attribute-argument producers. `daa108fc` fixed those
  producer boundaries; `d520b6b9` kept the compiled fixture warning-free.

Opus and Gemini agreed on the final behavior and reported no outstanding
findings. No findings were dismissed or left without action.

Final evidence: Release solution build passes; CSharp 84/84, Metadata 418/418,
CLI 1,775/1,775 (10 environment-dependent skips), RTS 96/96; CI is green; and
the explicit merge-base render A/B evaluated 89,065 methods with 0 changed,
added, or removed.

richlander:
Ready to merge

richlander:
## Review — the move is real and the layering is finally clean; but this PR is three PRs, and the two hidden ones need their own evidence

**The headline claim checks out.** `CSharpDeclarationWriter` physically moves (git rename, no copy left in Metadata, no forwarding shim, zero stragglers by grep), all five writer types demote to `internal` in their new home, `ApiParameter.Declaration`/`ParameterDeclarationsSummary` — the last Metadata members composing C# text — are deleted with all consumers moved to `CSharpFormatter`, and the dependency direction stays clean (Metadata references no CSharp; the facade and writer are now same-assembly siblings). The "temporarily delegates to Metadata's compatibility writer" era from #2626 is genuinely over, and the residual Metadata spelling is scoped by an honest permanent comment to the legacy `ApiMember.Signature` compatibility string only. That scoping decision — duplicate the escapers rather than create a Metadata→CSharp edge — is the right layering call.

**But two more PRs are riding inside the move, and neither gets the evidence its nature demands:**

1. **An output-changing accessibility fix.** `RenderTypeDeclarationCore` and `FormatDelegate` stop hardcoding `public`; internal/private types inspected with `--all` now render `internal class Foo` (new `ApiType.Accessibility`, populated in both extractors), and `--json` gains an `accessibility` field — a schema addition. These are *improvements* — the hardcoded `public` on non-public types was a lie, and my #2633 review flagged the delegate hardcode — but they're user-visible rendering and schema changes shipped under a "move" title with unit-test-only evidence, in a repo whose culture settles output changes with A/B or corpus runs. Ask: state both changes in the PR body as deliberate, and run whatever A/B the `type`/`member` surfaces support (the 89K decompiler render A/B doesn't cover declaration formatting) — or explicitly enumerate the expected diffs.
2. **Five keyword-escaping commits** (qualified segments, nested generic args, `typeof()`/enum-cast attribute arguments, string-literal quotes, `EscapeKnownIdentifiers` idempotence) — each a real fix with a test, fine on the merits.

**The duplication needs one hardening pass before it fossilizes:**

- `IsTypeSyntaxKeyword` is duplicated **character-for-character** on both sides of the boundary, and the two escaper truth tables *already disagree*: Metadata's set includes `record`/`required`/`init`/`file`/`scoped` but lacks `await`; CSharp's is the reverse. A parameter named `record` renders `@record` in the legacy signature but bare `record` from the facade — the skew is pre-existing, but this PR chose to fork `ParameterDeclaration` into Metadata rather than leave one implementation, which entrenches it. If the two sets *should* differ (compat string vs. modern rendering), a comment on each naming its counterpart and the intentional delta; if not, unify the tables even while the scanners stay separate.
- `IsTypeSyntaxKeyword`'s allow list is also lossy in the fail-open direction: a genuine type actually named `delegate`/`readonly`/`scoped` won't be escaped and yields invalid C#, while the tested cases (`class`, `event`) work. Rare, but it's the pattern this repo usually annotates.
- `Accessibility ?? "public"` conflates null-means-public with null-means-unknown — an old serialized surface or a construction site that forgets the field silently renders `public` with no diagnostic. Three sites were updated; nothing enforces the fourth. A doc line on the property (mirroring `MetadataName`'s "null in older serialized surfaces") plus the fallback choice stated would do.

Process note for next time rather than this PR: a pure-move commit followed by behavior commits would have made the 95%-similarity rename reviewable as a move; as structured, verifying byte-preservation required manually separating edits from relocation. And the `TypeRequirements => Types` alias removal riding along is welcome (it was on the residue list) but belongs in the message's headline, not buried.

Endorse the architecture — this completes the spelling-ownership migration the #2626/#2633 arc set up. Pre-merge asks: the accessibility/JSON changes named in the body with declaration-surface A/B or an expected-diff list, and the escaper truth-table skew either unified or explicitly annotated.

---

richlander:
Addressed in `cf7b1932` and reflected in the revised PR title/body.

- **Accessibility/schema disclosure:** the body now calls out actual
  non-public type/delegate rendering and the detailed JSON `accessibility`
  addition. `ApiType.Accessibility` documents that null includes older
  serialized surfaces and deliberately renders as `public`.
- **Declaration A/B:** merge base `96320e8a` versus `cf7b1932`, inspecting the
  same head-built fixture, changes
  `public class InternalTopLevelSurfaceFixture` to `internal class ...` and adds
  only `"accessibility": "internal"` to detailed JSON. The Markdown member
  surface is byte-identical on both sides.
- **Escaper policy:** the CSharp and Metadata compatibility keyword sets are now
  synchronized and cross-referenced. Type-syntax recognition distinguishes
  function-pointer syntax from genuine types named `delegate`, `readonly`, and
  `scoped`; compiled Metadata fixtures and focused CSharp tests cover both.
- **Alias removal:** `TypeRequirements => Types` is now in the title and opening
  summary.

Current evidence: Release build; CSharp 90/90, Metadata 418/418, CLI
1,777/1,777 (10 environment-dependent skips), RTS 96/96; explicit merge-base
render A/B 89,065 methods with zero changed, added, or removed.

The previous `Ready to merge` comment is superseded by this new head. Two
exact-head adversarial signoffs are being rerun before readiness is reposted.

richlander:
## Follow-up @ `cf7b1932` — all three asks landed, and the allow list fix is better than what I asked for

- **The hidden PRs are hidden no more.** The title now declares both riders, and the body states the accessibility-rendering change and the JSON `accessibility` schema addition as deliberate — with the **declaration-surface A/B** I asked for: the exact expected diff (`public class` → `internal class` on the fixture), the single JSON addition, and a byte-identical hash on the unaffected member surface. That's an expected-diff list *and* the proof the diff is exactly that list. The quality card also correctly shows the #2649-era `(sampling differs)` neutral rows rather than claiming credit.
- **The escaper truth tables are unified**: `s_csharpReservedKeywords` absorbs `await`/`record`/`required`/`init`/`file`/`scoped` (superset of both old sets), with the sync comment pointing at Metadata's legacy escaper and a theory pinning the contextual set. `record` and `await` now escape identically on both sides of the boundary.
- **The lossy allow list got the contextual fix rather than the annotation I'd have settled for**: `IsTypeSyntaxKeyword` now uses position — primitives only when not qualified, `delegate` as syntax only before `*`, `ref`/`in`/`out`/`readonly`/`unmanaged` only inside `delegate*<...>`, modifiers only leading-position-before-whitespace. A genuine type named `delegate` or `scoped` now escapes (test-pinned: `@delegate delegateValue` alongside an unescaped `delegate*<ref int, void>` in the same parameter list) — the fail-open hole closed instead of documented.
- The null-accessibility fallback is documented as the compatibility choice for older serialized surfaces, and the `TypeRequirements => Types` alias removal — residue item from the #2633 sweep — rides along, now declared in the headline.

Endorse; safe to merge once your in-progress exact-head signoff completes. This closes the spelling-ownership migration the #2626/#2633 arc opened — Metadata now holds structured facts plus one narrow, comment-fenced legacy escaper, and every declaration renders through the CSharp seam.

---

PR #2666, opened 2026-07-11
Add decompiler fidelity-cause Finding producer

richlander:
## Change

**Conclusion:** **PASS** — decompiler fidelity lowering now has one complete,
typed, identity-bearing cause census shared by grading, Findings, compatibility
remarks, and harness consumers.

- Adds `DecompilerFindings.InspectFidelityCauses` and
  `CompareFidelityCauses`.
- Adds typed unknown/signature/IL/local provenance without using version-local
  coordinate values as Finding identity.
- Derives `IrFunction.Fidelity` from the same cause enumerator.
- Adds `DEC0013` for residual `continue` and `DEC0014` for referenced unraised
  pinned locals, closing cases where `Partial` previously had no remark.
- Keeps `FidelityRemarks.Collect` as a compatibility projection.
- Migrates harness reports, corpus sensing, `--remarks`, and DEC0009
  classification away from diagnostic-message parsing.

Card revision: `0d88f230`

## Evidence

| Check | Result |
| --- | --- |
| Release solution build | Pass |
| Non-AOT product build | Pass |
| Focused producer/consumer tests | 19 passed |
| Decompiler fast suite | 2,300 passed after excluding two failures reproduced on merge-base `59565b7f` |
| Markdownlint | Pass |
| Real imported witness | `StackallocInitializerResiduals::StackallocPointerInitializer` reports typed `DEC0004` at `IL_001E` |
| Full-path allocation probe | 744 -> 664 bytes per `Fidelity` call on a 100-node Full function after final review hardening |

This does not change raised/printer C# text, so render-text A/B is not
applicable. The highest relevant boss is the Stage 9 PR quick corpus sensor.

### Decompiler quality diff

Corpus: #1166 real-world decompiler corpus sensor: #1150 pinned NuGet assemblies plus dotnet-inspect managed assemblies. 15 assemblies, 1,486 methods
Sample: hash-stable 100 methods per assembly
Correctness coverage: validity not run; fidelity not run

Baseline staleness: corpus drifted from the pinned baseline (generated 2026-06-26). Aggregate count deltas below include this corpus change and are not a clean PR signal; rebaseline against current main (--emit-corpus-baseline) before trusting count-based regressions.
- total methods 1,461 -> 1,486 (+25)
- ILInspector.MetadataPrimitives: 61 -> 86 methods (+25)

| Metric | Change | Count delta |
| ------ | ------ | ----------- |
[...]

PR #2681, opened 2026-07-12
Add product fidelity cause section

richlander:
Follows #2666 by adding its first envelope-retaining product consumer.

## Change

**Conclusion:** **PASS** - the opt-in member `Fidelity Causes` section now
answers why a method is not Full fidelity without collapsing producer outcomes.

```bash
dotnet-inspect member <type> <method> -S "Fidelity Causes"
```

The section retains `FindingInspection<DecompilerFidelityCause>` through
acquisition and renders distinct outcomes for:

- `Complete` with a Full-fidelity empty census
- `Complete` with ordered typed causes and locations
- `Absent` when the selected method has no IL body
- `Failed` when import, context, or projection fails

When selected with `Decompiled Source`, both sections share one imported and
raised `IrFunction`. The section is explicit-only, is not part of `@Source`,
and participates in single-overload auto-selection.

## Evidence

| Check | Result |
| --- | --- |
| Non-AOT product build | Pass |
| Product suite | 1,776 passed, 10 skipped |
| Focused Debug fidelity tests | Pass |
| Changed Markdown | Pass |
| Decompiler quality corpus | Not applicable; no raise, typing, or printer behavior changed |

The compiled `TypedReferenceType` fixture provides a real importer canary for
`DEC0004` / `mkrefany`; an empty-body fixture verifies that valid raised IR
remains `Complete` and renders as valid empty C#. A malformed-token synthetic
PE verifies the command-level `Failed` path.

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | Clean on `48a3e372` | Verified aligned empty-body output, envelope states, malformed-input safety, and AOT constraints |
| Gemini 3.1 Pro | Clean on `48a3e372` | Verified importer/printer failure paths, synthetic PE determinism, routing, and output modes |

Earlier rounds found four actionable issues:

1. A null import could be misclassified as `Absent`; `13fd71e0` now uses the
   selected method's metadata body state and preserves failures.
2. The `mkrefany` test assumed a Release-only IL offset; `13fd71e0` now checks
   typed IL-location shape in both configurations.
3. A valid empty method body's fidelity inspection could be misclassified as
   `Failed`; `28da09a1` initially corrected the fidelity outcome.
4. Decompiled Source still reported that empty body as `DEC0003` while Fidelity
   Causes reported Full; `48a3e372` removes the invalid body-only C# failure
   synthesis so both sections agree, and adds command-level `Failed`, TSV, and
   discovery coverage.

Both reviewers re-reviewed the exact fixed head with no outstanding findings.

## Validation

```bash
dotnet build dotnet-inspect.slnx -c Release --no-restore --verbosity quiet
dotnet build src/dotnet-inspect -c Release -p:PublishAot=false --no-restore --verbosity quiet
dotnet run --project src/dotnet-inspect.Tests -c Debug -- \
  -method "*Member_SelectedOverload_SelectFidelityCauses_EmptyBodyIsComplete*" \
  -method "*Member_SelectedOverload_SelectFidelityCauses_RendersTypedCause*"
npx markdownlint-cli README.md docs/architecture.md \
  docs/design/finding-producers.md skills/decompiler/SKILL.md
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ context for copied DLLs. A future `--deps` source can represent runtime
 | Relationships | `depends`, `extensions`, `implements` | Type hierarchies, package dependencies, library reference graphs, extension methods/properties, implementors and subclasses. Add `--project` to search project-referenced packages. |
 | Source mapping | `type`/`library`/`package -S "Source Files"`, `member -S "Source Locations"` / `"Original Source"` | SourceLink URLs, member file/line locations, source fetching, URL verification, token+IL-offset to source-line resolution. |
 | Performance analysis *(experimental)* | `library`/`type`/`member -S "Top Leverage"`, `"Performance Triage"`, `"Call Graph"`, `"Caller Graph"` | Whole-assembly call-graph leverage ranking — direct callers, root reach, fanout, depth, loop calls — with opt-in per-node cost signals (alloc, copy, unsafe, reflection, throw/exception, catch/finally) and actionable rewrite-shape detection. |
-| Decompiler *(experimental)* | `member -S @Source` (`Decompiled Source`, `Annotated Source`, `Original Source`, `Source Diff`, `IL`) | Raises method bodies to C#, interleaves IL and hidden-fact annotations, diffs SourceLink-backed source against decompiled source, and degrades honestly with `DEC####` fidelity diagnostics rather than emitting plausible-but-wrong source. |
+| Decompiler *(experimental)* | `member -S @Source` (`Decompiled Source`, `Annotated Source`, `Original Source`, `Source Diff`, `IL`); `member -S "Fidelity Causes"` | Raises method bodies to C#, interleaves IL and hidden-fact annotations, diffs SourceLink-backed source against decompiled source, and exposes typed `DEC####` fidelity causes rather than emitting plausible-but-wrong source. |
 | Agent-friendly output | global flags | Markdown by default, compact `--table`, normalized `--tsv`, `--jsonl`, `--plaintext`, `--json`, Mermaid diagrams, section/field projection, `--count`, table row limiting, built-in head/tail limiting. |
 
 ## Command inventory
@@ -226,13 +226,16 @@ honestly: IL with no faithful C# spelling renders as a visible comment and
 lowers the result's fidelity level (`Full` → `Partial` → `StructuredOnly` →
 `IlOnly` → `Failed`) instead of emitting plausible-but-wrong source, with a
 stable `DEC####` diagnostic on every degradation. If output looks wrong,
-capture all four sections; maintainers diagnose pipeline state with
+capture all four sections; select `Fidelity Causes` for the typed cause census
+behind the grade. That section distinguishes Full fidelity, an absent method
+body, and inspection failure. Maintainers diagnose pipeline state with
 DecompilerHarness.
 
 ```bash
 dotnet-inspect member JsonSerializer --package System.Text.Json Serialize:1 -S @Source
 dotnet-inspect member MyType Method:1 --library MyLib.dll -S "Decompiled Source"
 dotnet-inspect member MyType Method:1 --library MyLib.dll -S "Annotated Source"
+dotnet-inspect member MyType Method:1 --library MyLib.dll -S "Fidelity Causes"
 dotnet-inspect library MyLib.dll --il-offset 0x06000001+0x5
 ```
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -80,7 +80,9 @@ Inspects .NET library files (PE/COFF format):
 Extract public API surface using metadata:
 
 - `type` renders type shape, summaries, members, and `--shape` declarations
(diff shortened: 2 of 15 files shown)


richlander:
Ready to merge

richlander:
## Review — the #2666 product consumer, done inspect-first; one cross-section contradiction to reconcile

This answers the "≥2 consumers" question my #2666 review left open, and it answers the *other* residue from that review too: there is **no `Fidelity == Full` pre-gate anywhere** — a succeeded projection flows into `InspectFidelityCauses` regardless of fidelity, so the CLI inspects-then-folds where the harness consumers still fold-then-inspect. The envelope consumption is the adoption doc's rule 2 realized: Complete-empty, Complete, Absent, and Failed each render as their own row shape, the section renders *even when Failed* (with the joined diagnostics verbatim), and nothing is collapsed. The IR handed to the producer is genuinely the final raised tree (assigned to the `out` param before printing, so even a print-throw path can't hand a stale one), and the subject identity's absolute-path internals never leak into output. Section wiring is complete across both registries with the ExplicitOnly and not-in-`@Source` choices pinned by tests, and the docs land in all four places.

**The one thing to reconcile before merge: the DEC0003 carve-out creates an observable cross-section contradiction today.** For `void EmptyBody() {}`:

- `-S "Decompiled Source"` renders `Failure(EmptyOutput, "projection produced no output…")`
- `-S "Fidelity Causes"` says "No fidelity causes; decompiler fidelity is Full."

Same method, two sections disagreeing about whether something failed. The excusal (`projection.Diagnostics.All(d => d.Id == EmptyOutput)` → treat as valid) is a deliberate choice per the second commit — and it's arguably the *correct* verdict, since an empty body genuinely has Full fidelity — but then Decompiled Source's story is the wrong one, and the two should agree. The subtler cost: projection-level `EmptyOutput` existed to catch a raise pass that wrongly drops every statement, and the excusal can't distinguish that bug class from a legitimately empty body (both yield whitespace output and an empty census). Options: give the empty-body case its own visible wording here ("Complete — empty body") rather than the generic Full line, fix Decompiled Source to render empty bodies as success-with-note so the sections agree, and add a comment on the excusal naming the traded-away bug class. Any of these; the current silent disagreement is the only shape I'd hold on. (Mitigations already in place and verified: IR-*attached* `EmptyOutput` still fail-closes in the producer, so only the projection-level variant is excused.)

One concern from the deep-read I can partially defuse: the "fidelity is Full" wording is an inference from census emptiness, never cross-checked against `IrFunction.Fidelity` — but since #2666, `Fidelity` *is* a fold over this same cause enumeration, so they cannot disagree **by construction today**. That construction is load-bearing, though, and lives in another assembly: one test pinning `census empty ⇔ Fidelity == Full` through the product path would keep a future `HasAny`/`CollectCauses` divergence from silently breaking the section's headline sentence.

Test gaps to fill (cheap): no **command-level Failed** test — the Failed row is only exercised via hand-built envelopes, and the `pipelineSource is null → Failed` branch is entirely untested end-to-end; no `--json`/`--tsv` projection test for the new rows; no `-D` discovery listing test. Follow-up grade: `BuildFidelityCauseInspection` re-implements operation-failure classification parallel to the producer's `IsOperationFailure`, and `SelectedMethodHasBody` duplicates the importer's overload-matching rules — two divergence seams that want either a shared helper or a cross-pinning test.

Endorse with the empty-body story reconciled and the Failed e2e pinned. The envelope discipline here is the best of any product consumer so far — it just needs its neighbors to tell the same story about the same method.

---

richlander:
Addressed the review hold in `48a3e372`: empty method bodies now render as
successful empty C# projections, so Decompiled Source and Fidelity Causes
agree.

Added command-level coverage for a diagnosed importer crash (`Failed`, not
`Absent`), plus TSV and scoped discovery coverage. Re-running exact-head
adversarial review and CI before restoring `Ready to merge`.

richlander:
## Follow-up @ `da91c557` — reconciled in the right direction, all gaps pinned; safe to merge

- **The cross-section contradiction is resolved the way it should have been**: Fidelity Causes' verdict was the correct one, and Decompiled Source now agrees — a legitimately empty body renders as a successful source projection, extended consistently across annotated/cost/semantics overlays (`emptyOutputIsFailure: false`), while **annotated IL keeps its empty-output failure semantics** (IL genuinely can't be empty for a method with a body). One story per method across all sections. The traded-away bug class (a raise pass wrongly dropping every statement) is now guarded by the lanes actually suited to it — the IL section, the corpus render A/B, and compile-back — rather than by a projection-level heuristic that couldn't tell the two cases apart. That trade deserves its comment, but the direction is right.
- **All three test gaps closed**: the command-level Failed path now has a real end-to-end test driving a synthesized malformed fixture through the importer to a rendered `Failed` row; TSV projection and `-D --schema` discovery are covered.
- The inspection condition also *simplified* as a consequence — with empty bodies legitimately succeeding, the EmptyOutput carve-out I flagged disappears entirely (`projection?.Succeeded == true`), which is the best possible fate for a carve-out.

Endorsement unconditional; safe to merge.

---

PR #2682, opened 2026-07-12
Preserve typed decompiler results through CLI formatting

richlander:
Advances #2653

- Preserves complete `DecompilerResult` values through Research and CLI composition.
- Removes constructor-chain and `await` source-text protocols from `MemberCodeProvider`.
- Card revision: `a83a8c4a`

## Change

**Conclusion:** **PASS** — the CLI now consumes typed constructor, async,
fidelity, diagnostic, and trace evidence without changing decompiler body
rendering.

Before, a method returning the literal `"await"` could acquire a false `async`
modifier because the CLI scanned rendered text:

```csharp
public async string AwaitText()
{
    return "await";
}
```

After:

```csharp
public string AwaitText()
{
    return "await";
}
```

The same typed path now composes real constructor initializers from
`DecompilerResult.ConstructorChain`; body text resembling `: base(...)` is no
longer reparsed as structural metadata.

## Evidence

| Check | Result |
| --- | --- |
| Product/test build | Pass |
| Research tests | 89 passed |
| CLI tests | 1,765 passed, 10 environment skips; fixed-head formatter tests: 15 passed, 1 environment skip |
| NativeAOT publish | Pass |
| Decompiler fast suite | 2,332 / 2,333 passed; the one failure reproduces unchanged on clean `origin/main` |
| Render A/B | 89,065 evaluated; 0 changed, 0 added, 0 removed |

The pre-existing fast-suite failure is
`TypeSourceComposerUnionTests.ClassUnionSwitchExpression_RendersTypePatternArms`;
the identical expected/actual mismatch reproduces at clean merge-base
`2b3db7ee`.

## Decompiler quality

**Conclusion:** **PASS** — the pinned PR quick gate matched baseline tolerances.
Aggregate movement is advisory baseline drift from 25 additional
`ILInspector.MetadataPrimitives` methods, while the explicit render A/B is
empty.

### Decompiler quality diff

Corpus: #1166 real-world decompiler corpus sensor: #1150 pinned NuGet assemblies plus dotnet-inspect managed assemblies. 15 assemblies, 1,486 methods

Baseline staleness: corpus drifted from the pinned baseline (generated 2026-06-26). Aggregate count deltas below include this corpus change and are not a clean PR signal; rebaseline against current main (--emit-corpus-baseline) before trusting count-based regressions.
- total methods 1,461 -> 1,486 (+25)
- ILInspector.MetadataPrimitives: 61 -> 86 methods (+25)

| Fully raised (+) | 1,213 (83.03%) → 1,232 (82.91%) (bad) | +19 |
| Conditional-branch residual (-) | 35 (2.40%) → 44 (2.96%) (bad) | +9 |
| Forward-merge stops (-) | 27 (1.85%) → 32 (2.15%) (bad) | +5 |

Pinned-subset gate (PR quick rate/count regressions evaluated here): Fully raised 82.17% -> 82.33% (+0.16 pp); conditional residual 3.00% -> 3.00% (0 pp); fidelity ungated (sampling differs; rely on changed-method fidelity).

Advisory aggregate rate movement (not a PR quick hard gate; review for decompiler-risky changes):
- fully-raised dropped 0.12 pp (baseline 83.03%, PR 82.91%, tolerance 0.10 pp)
- conditional-branch residual increased 0.56 pp (baseline 2.40%, PR 2.96%, tolerance 0.10 pp)
- forward-merge stop increased 0.30 pp (baseline 1.85%, PR 2.15%, tolerance 0.10 pp)

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | CLEAN | Full fixed-head diff, diagnostics, async, constructor, and live output checks |
| Gemini 3.1 Pro | CLEAN | Fixed-head extraction, diagnostics, async, cost, trace, and API checks |

**Review conclusion:** **PASS** — both isolated reviewers found no actionable
issues on exact head `a83a8c4a`; the
[final reconciliation comment](https://github.com/richlander/dotnet-inspect/pull/2682#issuecomment-4951705465)
records their evidence and why no resolution commit applies.

### DIFF
--- a/src/ILInspector.Research.Tests/ResearchFactRegistryTests.cs
+++ b/src/ILInspector.Research.Tests/ResearchFactRegistryTests.cs
@@ -1,3 +1,4 @@
+using ILInspector.Decompiler;
 using ILInspector.Decompiler.Annotations;
 using ILInspector.Decompiler.Pipeline;
 
@@ -13,6 +14,33 @@ public class ResearchFactRegistryTests
             new TestProducer("source"));
 
         Assert.Equal(["source", "consumer"], registry.ProducerNames);
+    }
+
+    [Fact]
+    public void MemberProjection_PreservesConstructorChainMetadata()
+    {
+        using var source = MetadataSource.Open(typeof(ResearchConstructorFixture).Assembly.Location);
+
+        var projection = ResearchViews.ProjectMember(new ResearchViews.MemberProjectionRequest(
+            source,
+            typeof(ResearchConstructorFixture).FullName!,
+            ".ctor",
+            AnnotatedSource: true,
+            CostOverlay: true,
+            SemanticsOverlay: true));
+
+        var results = new[]
+        {
+            Assert.IsType<DecompilerResult>(projection.AnnotatedSource),
+            Assert.IsType<DecompilerResult>(projection.CostOverlay?.Body),
+            Assert.IsType<DecompilerResult>(projection.SemanticsOverlay)
+        };
+        Assert.All(results, result =>
+        {
+            Assert.True(result.Succeeded, string.Join(Environment.NewLine, result.Diagnostics));
+            Assert.StartsWith("this(", result.ConstructorChain);
+            Assert.DoesNotContain(": this(", result.Output);
+        });
     }
 
     [Fact]
@@ -397,4 +425,17 @@ public static class ResearchFixture
     public static int LoopCaller18(int count) { int total = 18; for (int i = 0; i < count; i++) total += HighLoopLeverageCallee(i); return total; }
     public static int LoopCaller19(int count) { int total = 19; for (int i = 0; i < count; i++) total += HighLoopLeverageCallee(i); return total; }
     public static int LoopCaller20(int count) { int total = 20; for (int i = 0; i < count; i++) total += HighLoopLeverageCallee(i); return total; }
+}
+
+public sealed class ResearchConstructorFixture
+{
+    public ResearchConstructorFixture()
+        : this(42)
+    {
+    }
+
+    public ResearchConstructorFixture(int value)
+    {
+        GC.KeepAlive(value);
+    }
 }
--- a/src/ILInspector.Research.Tests/ResearchFactRegistryTests.cs
+++ b/src/ILInspector.Research.Tests/ResearchFactRegistryTests.cs
@@ -44,6 +44,22 @@ public class ResearchFactRegistryTests
     }
 
     [Fact]
+    public void RunProjection_PreservesFailureDiagnostics()
+    {
+        var failure = DecompilerResult.Failure(
+            DiagnosticIds.ContextUnavailable,
+            "overlay context unavailable");
+
+        var result = ResearchViews.RunProjection(
+            () => failure,
+            emptyOutputIsFailure: true);
+
+        Assert.Same(failure, result);
+        Assert.Equal(DiagnosticIds.ContextUnavailable, Assert.Single(result.Diagnostics).Id);
+        Assert.DoesNotContain(result.Diagnostics, diagnostic => diagnostic.Id == DiagnosticIds.EmptyOutput);
+    }
+
+    [Fact]
     public void EmptyRegistry_ProducesNoOverlayFacts()
     {
         using var source = MetadataSource.Open(typeof(ResearchFixture).Assembly.Location);
--- a/src/ILInspector.Research/ILInspector.Research.csproj
+++ b/src/ILInspector.Research/ILInspector.Research.csproj
@@ -14,6 +14,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <InternalsVisibleTo Include="ILInspector.Research.Tests" />
+  </ItemGroup>
+
+  <ItemGroup>
     <ProjectReference Include="..\ILInspector.Analysis\ILInspector.Analysis.csproj" />
     <ProjectReference Include="..\ILInspector.Decompiler\ILInspector.Decompiler.csproj" />
     <ProjectReference Include="..\ILInspector.Findings\ILInspector.Findings.csproj" />
--- a/src/ILInspector.Research/ResearchViews.cs
+++ b/src/ILInspector.Research/ResearchViews.cs
@@ -68,7 +68,7 @@ public static class ResearchViews
             if (request.AnnotatedSource)
             {
                 annotatedSource = WithTrace(
-                    Run(() => RenderMixedCore(
+                    RunProjection(() => RenderMixedCore(
                         request.Source,
                         request.Type,
                         request.Method,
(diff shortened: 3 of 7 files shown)


richlander:

Exact reviewed head: `fe3e108a`

| Reviewer | Result | Evidence |
| --- | --- | --- |
| Claude Opus 4.8 | CLEAN | Verified the complete `DecompilerResult` contract through all four source sections and live-checked constructor-only, field-assigning constructor, real async, `"await"` literal, and full-type output. |
| Gemini 3.1 Pro | CLEAN | Verified empty constructor results, per-result fidelity/diagnostic/trace preservation, both async flags, cost headers, expression bodies, and declaration composition. |

The reviewers agreed that:

- constructor-only empty `Output` remains successful when
  `ConstructorChain` is present;
- Research and CLI projections retain typed fidelity, diagnostics, trace,
  constructor, and async evidence;
- `ContainsAwaitExpression || RequiresAsyncBodyModifier` is the correct
  declaration rule;
- body text is no longer misinterpreted as constructor or async metadata;
- decompiled, annotated, cost-overlay, and semantics-overlay sections use the
  same typed composition path without losing section-specific headers.

No actionable or disputed findings were reported, so no resolution commit
applies.

richlander:
Ready to merge

richlander:
## Review — three erasure points closed, and the type system now carries what a text scan used to guess

This is billed as formatting plumbing, but it fixes a real rendering bug and a real evidence-integrity hole:

- **The `async` inference was a text scan, and it was wrong.** The old `ContainsAwaitKeyword` matched the *word* "await" in rendered output — including inside string literals — so a method containing `"await"` as data rendered with a spurious `async` modifier. Now the declaration uses the printer's own typed evidence (`ContainsAwaitExpression`/`RequiresAsyncBodyModifier`), and the end-to-end test pins exactly the poisonous case (`"await"` literal present, ` async ` absent). That's a user-visible correctness fix riding in a refactor, properly pinned.
- **`ResearchViews.Run` was laundering fidelity.** Every annotated/cost/semantics render was re-wrapped via `DecompilerResult.Success(output)` — which fabricates `Full` fidelity and empty diagnostics — so research traces claimed Full-with-no-diagnostics for degraded renders, and a real inner failure collapsed to a generic `EmptyOutput` with the true diagnostic IDs discarded. Now pipelines return the typed result and overlays graft with `result with { Output = ... }`, so fidelity, diagnostics, constructor chains, and async flags survive to the boundary. This is the same disease #2666's review called out in the harness (verdicts fabricated at the seam), cured in the CLI.
- **The constructor-chain stringify/reparse cycle is gone** — the chain stays typed from printer to declaration composition instead of being serialized into body text and re-parsed by `StartsWith(": base(")`. Per-section traces replace the shared-trace overwrite (the research trace used to clobber the decompiled section's breadcrumb), and failed sections still render `// DECxxxx` comments — fail-visible confirmed at the `FormatCSharpResult` seam.

**Two pinning asks — both for behavior changes the PR makes deliberately but doesn't test:**

1. **Failed overlay sections now show real diagnostics** instead of old `Run`'s generic EmptyOutput substitution — the improvement most worth having, and no test drives a failed annotated/cost/semantics render through the new path to assert the `// DEC...` text.
2. **`forceAsync` now applies to annotated/cost/semantics declarations** (previously decompiled-source only) — likely correct, but an async state-machine's annotated view renders differently than before and nothing pins it.

Two notes for the record, no action: the removed body-text chain-lift was a safety net, and the new test *enshrines* the degraded output on regression (a leaking `: base(42)` stays in the body as an invalid statement) — defensible as fail-visible-at-the-source, but it's a deliberate robustness removal, so it's good the test documents the choice; and `MemberCodeProvider` unconditionally rebuilds `decompiledResult.Trace`, which is harmless today but will silently discard any richer printer-supplied trace if one ever arrives — a one-line `??=` future-proofs it. The pre-existing `FieldInitializers` gap (chain-only exemption in `Run` doesn't cover initializer-only ctor bodies) stays pre-existing; noting it so it lands on someone's list rather than nobody's.

Endorse — with the two pins added, this is exactly the "typed evidence to the presentation boundary" shape the rendering model doc prescribes.

---

richlander:
## Feedback resolution

Addressed both pinning requests in
[`a83a8c4a`](https://github.com/richlander/dotnet-inspect/commit/a83a8c4a):

- `RunProjection_PreservesFailureDiagnostics` proves a failed Research
  projection retains its original diagnostic ID instead of becoming
  `DEC0003`.
- `PopulateCSharpSections_PreservesOverlayFailureDiagnostics` drives failed
  annotated, cost, and semantics results through the actual shared formatter
  path and pins the rendered `// DECxxxx` comments.
- `PopulateCSharpSections_AppliesTypedAsyncEvidenceToAllOverlays` pins `async`
  declaration composition for annotated, cost, and semantics overlays while
  also retaining cost header comments.

The formatter's existing four-section block was extracted unchanged into
`PopulateCSharpSections` so these tests exercise the production path rather
than a test-only approximation.

Non-actions:

- The removed body-text constructor parser remains intentionally removed; its
  near-miss test documents that typed evidence, not body spelling, owns
  declaration structure.
- `decompiledResult.Trace` remains host-created because this seam adds the
  actual symbol source. A future product-supplied trace needs an explicit
  provenance merge rather than `??=`, which could retain stale/empty symbol
  evidence.
- The pre-existing initializer-only constructor gap is now tracked in #2688.

Refreshed exact-head evidence: solution build passed; 89 Research tests and 15
formatter tests passed (one environment skip); render A/B evaluated 89,065
methods with 0 changed/added/removed; the PR quick corpus gate matched baseline
tolerances.

richlander:
## Follow-up @ `a83a8c4a` — both pins landed; safe to merge

- `PopulateCSharpSections_PreservesOverlayFailureDiagnostics` drives failed annotated/cost/semantics renders through the shared formatter path and asserts the real `// DEC000x` text in each section — the improvement most worth having is now guarded.
- `PopulateCSharpSections_AppliesTypedAsyncEvidenceToAllOverlays` pins the deliberate `forceAsync` spread to all overlay declarations, and `RunProjection_PreservesFailureDiagnostics` pins the deeper property — a projection failure passes through with its original diagnostic ID (`ContextUnavailable` asserted, `EmptyOutput` asserted absent), so the old fidelity-laundering can't silently return.

Endorsement now unconditional.

---

richlander:
## Final adversarial review reconciliation

Exact reviewed head: `a83a8c4a`

| Reviewer | Result | Evidence |
| --- | --- | --- |
| Claude Opus 4.8 | CLEAN | Reviewed both commits and all seven changed files; reproduced constructor-only and async output; verified original diagnostics, section extraction, cost headers, trace, and API boundaries. |
| Gemini 3.1 Pro | CLEAN | Verified unchanged `hasCode` semantics, real production-path tests, diagnostic rendering, async overlays, constructor-only output, cost formatting, and internal-only test exposure. |

Both reviewers confirmed that feedback commit
[`a83a8c4a`](https://github.com/richlander/dotnet-inspect/commit/a83a8c4a)
fully resolves the requested pins:

- failed Research projections retain their original diagnostics;
- failed annotated, cost, and semantics sections render those exact
  `// DECxxxx` comments;
- typed async evidence composes declarations across all three overlays;
- the extracted shared formatter block preserves production behavior and cost
  header comments.

No actionable or disputed findings remain, so no further resolution commit
applies. The deliberate non-actions and #2688 follow-up are documented in the
[feedback-resolution comment](https://github.com/richlander/dotnet-inspect/pull/2682#issuecomment-4951679012).

richlander:
Ready to merge

PR #2684, opened 2026-07-12
Replace message-based validity classification

### LINKED ISSUE #2657: Replace ValidityCheck diagnostic message-text decisions
## Problem

`ValidityCheck` still makes a classification decision with localized Roslyn prose (`diagnostic.GetMessage().Contains("__Shell")`). PR #2652 removes the same kind of message-text coupling from compile-back closure membership; this remaining decision can still change under localization or compiler wording changes.

## Scope

- Inventory `ValidityCheck` decisions that inspect `Diagnostic.GetMessage()`.
- Replace classification logic with structured evidence: diagnostic ID, source location/span, syntax, and semantic symbols as appropriate.
- Keep message text only for human-readable diagnostics, not control flow.
- Add focused positive and near-miss tests for the replacement discriminator and null/non-source locations.

## Acceptance criteria

- No validity classification decision depends on localized diagnostic message text.
- Existing validity buckets remain stable on the fixture suite unless an intentional correction is documented.
- Structured-evidence misses are observable rather than silently classified as ordinary failures.

Tracked from #2646 and follow-up feedback on #2652.

richlander:
Fixes #2657

- Replaces localized-message validity decisions with structured Roslyn and
  decompiler-model evidence.
- Card revision: `83dc4c7213f885b35bfca6c730c0d7c9a421872e`

## Change

**Conclusion: PASS** - validity classification now uses diagnostic IDs, source
syntax, semantic symbols/types, and decompiler model types; missing structured
evidence remains visible as `VLD0001`.

- Centralizes semantic diagnostic classification for `ValidityCheck` and
  `LibraryReport`.
- Anchors shell-`this` decisions to diagnostic-specific operands, including
  structurally equivalent array/tuple/generic shapes without hiding unrelated
  sibling mismatches.
- Resolves static-name collisions from diagnostic-specific type syntax,
  including qualified names.
- Intentionally exposes genuine method-level `CS0161` and protected-access
  defects that the former `__Shell` message substring filter hid.

## Evidence

| Check | Result |
| --- | --- |
| Product/test/fixture solution build | Pass |
| Focused validity and shell-noise tests | 38 passed |
| Decompiler fixture validity boss | 5,350/5,350 checked; 10 newly visible `CS0161`; no other delta |
| CoreLib validity boss | 38,434/38,434 checked; 25 newly visible `CS0161` rows and one genuine `CS1540`; no `VLD0001` |
| Markdown lint | Pass |
| PR CI | Pass |
| Full decompiler suite | 2,486 passed; one unrelated `TypeSourceComposerUnionTests` failure reproduces on `origin/main` |

The intentional fixture-bucket correction is documented in the harness README.

### Decompiler quality diff

Correctness coverage: validity compiled 25,183 methods (compile-cap 4,000; per-sample, not corpus-wide); fidelity sampled 36 / 89,065 (0.04%)

| Semantic defects (sampling differs) | 54 (0.21%) → 101 (0.40%) | n/a |
| Fidelity opcode diffs (sampling differs) | 4 (11.11%) → 3 (8.33%) | n/a |
| Fidelity exact (sampling differs) | 32 (88.89%) → 33 (91.67%) | n/a |

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | CLEAN at `83dc4c72` | Reproduced direct/structural positives, mixed/sibling negatives, static collisions, `CS1540`, and `VLD0001` |
| Gemini 3.1 Pro | CLEAN at `83dc4c72` | Verified recursive structural equivalence, operand symmetry, qualified names, fail-visible fallback, and report parity |

**Review conclusion: PASS** - both isolated exact-head reviewers are clean.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ValidityCoverageReportingTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ValidityCoverageReportingTests.cs
@@ -1,4 +1,5 @@
 using ILInspector.DecompilerHarness;
+using ILInspector.Decompiler.Pipeline;
 
 namespace ILInspector.Decompiler.Tests;
 
@@ -7,6 +8,7 @@ public class ValidityCoverageReportingTests
 {
     static readonly object ConsoleGate = new();
     static string FixturePath => typeof(LadderRung1.Foundation).Assembly.Location;
+    static string DecompilerPath => typeof(IrImporter).Assembly.Location;
 
     [Fact]
     public void ValidityCheck_CappedRunLabelsSemanticFindingsAsPerSample()
@@ -27,6 +29,32 @@ public class ValidityCoverageReportingTests
         Assert.Contains("No compilation performed", output);
         Assert.Contains("conditional-arm-numeric-join-cast", output);
         Assert.Contains("conditional-target-numeric-cast", output);
+    }
+
+    [Fact]
+    [Trait("Speed", "Slow")]
+    public void DecompilerAssembly_MissingReturnPopulationIsPinned()
+    {
+        var actual = ValidityCheck.Evaluate(DecompilerPath)
+            .Where(result => result.SemanticDiagnostics.Any(diagnostic => diagnostic.Id == "CS0161"))
+            .Select(result => result.Id)
+            .Order(StringComparer.Ordinal)
+            .ToArray();
+
+        string[] expected =
+        [
+            "ILInspector.Decompiler.Pipeline.BooleanFoldingPass::IsNullableCoalesceExpressionContext",
+            "ILInspector.Decompiler.Pipeline.CSharpPrinter::ForLoopIncrementText",
+            "ILInspector.Decompiler.Pipeline.DeconstructionAssignmentPass::TryMatchTupleSeed",
+            "ILInspector.Decompiler.Pipeline.FixedArrayRaising::SameLoadPlace",
+            "ILInspector.Decompiler.Pipeline.IndexFromEndPass::LengthReceiver",
+            "ILInspector.Decompiler.Pipeline.InlineArrayCollectionPass::PlaceFromAddress",
+            "ILInspector.Decompiler.Pipeline.IsPatternPass::IsPatternLocalNull",
+            "ILInspector.Decompiler.Pipeline.NullConditionalPass::MemberReceiver",
+            "ILInspector.Decompiler.Pipeline.UnionSwitchExpressionPass::SameTailNode",
+            "ILInspector.Decompiler.TypeSourceComposer::DecompileBody",
+        ];
+        Assert.Equal(expected, actual);
     }
 
     static string CaptureConsole(Func<int> action)
--- a/src/ILInspector.Decompiler.Tests/ValidityCoverageReportingTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ValidityCoverageReportingTests.cs
@@ -42,6 +42,11 @@ public class ValidityCoverageReportingTests
             .ToArray();
 
         string[] expected =
+#if DEBUG
+        [
+            "ILInspector.Decompiler.TypeSourceComposer::DecompileBody",
+        ];
+#else
         [
             "ILInspector.Decompiler.Pipeline.BooleanFoldingPass::IsNullableCoalesceExpressionContext",
             "ILInspector.Decompiler.Pipeline.CSharpPrinter::ForLoopIncrementText",
@@ -54,6 +59,7 @@ public class ValidityCoverageReportingTests
             "ILInspector.Decompiler.Pipeline.UnionSwitchExpressionPass::SameTailNode",
             "ILInspector.Decompiler.TypeSourceComposer::DecompileBody",
         ];
+#endif
         Assert.Equal(expected, actual);
     }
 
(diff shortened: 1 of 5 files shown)


richlander:

- **Claude Opus 4.8 (round 1):** found that scanning complete ancestor
  expression subtrees could hide a genuine conversion/operator defect when an
  unrelated sibling used synthetic `this`. Resolved in
  [`40fc0343`](https://github.com/richlander/dotnet-inspect/commit/40fc0343015c58c595b47a6b2bcc911669da61cf)
  by anchoring evidence to diagnostic-specific operands and adding sibling
  near-miss regressions.
- **Gemini 3.1 Pro (round 1):** found that a qualified `CS0721` parameter type
  could be misidentified as the method return type. Resolved in
  [`40fc0343`](https://github.com/richlander/dotnet-inspect/commit/40fc0343015c58c595b47a6b2bcc911669da61cf)
  with diagnostic-ID-specific type syntax extraction and a qualified-name
  regression.
- **Gemini 3.1 Pro (round 2):** found that exact top-level `__Shell` type checks
  missed array/tuple structural shell artifacts. Resolved in
  [`83dc4c72`](https://github.com/richlander/dotnet-inspect/commit/83dc4c7213f885b35bfca6c730c0d7c9a421872e)
  with recursive shell-type substitution plus a mixed structural mismatch
  negative.
- **Final exact-head reviews:** Opus and Gemini both returned **CLEAN** at
  `83dc4c7213f885b35bfca6c730c0d7c9a421872e`.

All actionable guidance was implemented; there were no dismissed or
non-actioned findings.

richlander:
Ready to merge.

richlander:
## Review — the right cure, applied to the right disease; the parity story needs its evidence committed

The mechanism is exactly what the #2652 review asked to happen to this lane. Every deleted decision was message-or-text-based (`GetMessage().Contains("__Shell")`, the CS1605 source-line `Contains("this")`, the locale-sensitive quoted-name scrape), and every replacement is structural: an enumerated ID gate, per-ID syntax extraction of the diagnosed operand, symbolic shell identity (`Name: "__Shell"` at global scope — a symbol test, not a string-in-message test), and recursive structural equivalence that substitutes `__Shell → declaring type` so composite filters only fire when the *non-shell* structure also matches. The `VLD0001` fail-visible marker for unextractable evidence is the pattern #2652's `closure-stalled-unextracted` established, correctly transplanted. And routing `LibraryReport` through the shared classifier kills a copy-paste lane that could drift — real dedup. The negative tests are genuinely adversarial (shell-`this` in a sibling subexpression must not launder an unrelated CS0266; kept-count parity when one operand legitimately filters).

**Grep verdict on the thesis**: zero `GetMessage()` *decisions* remain in the validity lane — with one footnote worth adding to the README: `TypeBindCheck.cs` still extracts the ambiguous name from CS0104 message text. It's invariant-culture-pinned and reporting-only, so it's not a filtering decision — but "the last message-text decision in the harness" should cite it as the known extraction remnant, or the next audit rediscovers it.

**Four asks, first two pre-merge:**

1. **Commit the evidence for "ten previously hidden missing-return defects."** That's the PR's only quantitative parity claim and it exists solely as a README sentence — no baseline change, no pinned-count test, nothing reproducible. A fixture-count test (or the sensor artifact) turns the intentional CS0161 change from an assertion into a fact. This repo's own standard — every other bucket-changing PR shipped its A/B — applies doubly to a PR whose whole point is classification integrity.
2. **Document (and ideally pin) the *undocumented* parity change: unenumerated shell-noise codes now surface as ordinary defects.** The old blanket message-match suppressed *any* error naming `__Shell`; the new gate covers nine IDs, so shell-`this` diagnostics outside the set — CS0039 (`this as Derived`, the faithful `isinst` render), CS1579, CS1674 are plausible corpus shapes — now appear as real defects with **no VLD marker**, indistinguishable from decompiler bugs. That may be the correct disposition (surface-then-enumerate beats silently-suppress), but the README's "evidence gaps remain visible" currently overstates the guarantee to cover only the nine, and no test pins the kept-by-default behavior. One theory over CS0039-style shapes plus a README sentence closes it.
3. **A third of the ID set has no positive filter test** — CS0266, CS8129, and especially CS1605, which replaced the crudest text heuristic in the old code. Cheap theories.
4. **Unify the evidence-miss disposition**: `ClassifySyntheticShellThis` returns bare `Keep` when the expected syntax shape is absent, while the CS1540 path returns `Unextracted` (→ VLD0001) for the analogous miss — the same failure class gets two different visibilities depending on which arm it enters.

For the record, follow-up grade: `MetadataFullName` compares without assembly or generic arity (over-filters full-name collisions and arity siblings — the collision family is half-acknowledged in a comment, the arity blindness isn't), and the new CS1540 exactness means protected access through a subclass-typed receiver — legal in the real type — is now reported (fail-safe direction, but it's new noise worth a docket note if the corpus hits it).

Endorse the mechanism unconditionally — this and #2652 together retire message-text as a decision input across the harness. The asks are all about making the classification *changes* as evidence-backed as the classification *mechanism* now is.

---

richlander:
## Follow-up @ `9050bca6` — all four asks landed, with the count claim made properly falsifiable; safe to merge

- **The "ten" is now executable**: a ten-method CS0161 sensor asserts the exact population, and `9050bca` makes it configuration-aware — explicit expectations for both Release and Debug IL with the documented claim scoped to Release. That's the difference between a README sentence and a fact: the claim now breaks the build if it drifts, in either build flavor.
- **The undocumented parity change is pinned**: unenumerated shell diagnostics are asserted kept-by-default, so the surface-then-enumerate disposition is a tested contract rather than an accident of the ID gate.
- **The remaining third of the ID set** (CS1605 — the one that replaced the crudest old heuristic — plus CS0266/CS8129) now has positive filter tests over real Roslyn diagnostics.
- **Evidence-miss disposition unified**: syntax-shape misses consistently fail visible instead of splitting between bare `Keep` and `VLD0001` by which arm they entered.

With this, the classification *changes* are as evidence-backed as the classification *mechanism* — and between this PR and #2652, message text is now fully retired as a decision input across the harness (the invariant-pinned CS0104 extraction in `TypeBindCheck` remains the one documented reporting-only remnant). Endorsement unconditional.

---

Issue #2632, opened 2026-07-10
RTS closure source mangles keyword namespace in using directive


richlander:
## Failure

`ReturnToSenderPrototypeTests.CompileBackFirstPropertyGetter_EscapesKeywordNamespacesInClosureRoots` emits a mangled `using` directive for a keyword namespace, causing compile-back to fail.

## Reproduction

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
  -method "*CompileBackFirstPropertyGetter_EscapesKeywordNamespacesInClosureRoots*"
```

Expected:

```csharp
using My.@event;
```

Actual source contains:

```csharp
using My.__event;
```

while the namespace declaration itself is correctly emitted as `namespace My.@event`. Roslyn reports `CS0234` because `My.__event` does not exist.

The failure reproduces in a clean worktree at `origin/main` (`11198494c848881121638dcd536c27cd472a8e6d`). It makes both the RTS prototype class run and the full Decompiler suite red for every agent.

richlander:
Fixed by #2633 through formatter-owned namespace spelling. Verified on merged `origin/main` (`f30d3611`): `CompileBackFirstPropertyGetter_EscapesKeywordNamespacesInClosureRoots` passes and emits `using My.@event;`.

Issue #2629, opened 2026-07-10
RTS closure plan omits body-type evidence for return type


richlander:
## Failure

`ReturnToSenderPrototypeTests.CompileBackPropertyGetters_AddsSameAssemblyReturnTypeClosureRoot` adds `Helper` to `Plan.Types` but does not retain the expected Metadata `body-type` source fact in `Plan.TypeRequirements` / closure evidence.

## Reproduction

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
  -method "*CompileBackPropertyGetters_AddsSameAssemblyReturnTypeClosureRoot*"
```

The first compile-back result is still `Exact` and includes `Helper`, but this assertion fails:

```csharp
requirement.SourceFacts.Any(fact => fact.Id == "body-type"
    && fact.Producer == "metadata"
    && fact.Detail == "Helper")
```

The failure reproduces in a clean worktree at `origin/main` (`11198494c848881121638dcd536c27cd472a8e6d`). It makes both the RTS prototype class run and the full Decompiler suite red for every agent.

richlander:
Fixed by #2633. Verified on merged `origin/main` (`f30d3611`): `CompileBackPropertyGetters_AddsSameAssemblyReturnTypeClosureRoot` passes and retains the Metadata `body-type` source fact.

PR #2686, opened 2026-07-12
Add RTS source-body fault isolation

richlander:
- Advances #2673
- Changes harness-only RTS reporting: `RecompileFail` rows with authored source coverage now get a sidecar body-vs-shell fault-isolation result.
- Evidence revision: `3eac610a`

## Change

**Conclusion:** **REVIEW** — RTS now retries the same typed `ArtifactRequest` with only the authored target body substituted, preserving the top-level `RecompileFail` verdict while classifying the fault as `body-defect` or `shell-or-closure-defect`.

Before:

```text
RecompileFail conflated invalid product/decompiler bodies with RTS shell, closure, reference, or standalone source-consumption failures.
```

After:

```text
RecompileFail remains the oracle verdict; `FaultIsolation` and generated-catalog reason refine source-covered rows into body-defect vs shell-or-closure-defect.
```

## Scope and safety

- **Root cause:** RTS had only Roslyn diagnostics from the product artifact, so a recompile failure could not distinguish body defects from shell/closure defects.
- **Fix shape:** Reuse the existing fixture/source body index, pass source paths into generated fixture RTS runs, and retry only the target body in the same request shell.
- **Product impact:** none; changes are under `tools/DecompilerHarness` and tests/docs.
- **Scope boundary:** this slice uses existing fixture/source-path coverage; remote SourceLink fetching remains future work.
- **RTS boundary:** RTS only orchestrates the sidecar retry and reporting; product/shared code still owns C# body/artifact generation.

## Evidence

| Check | Result |
| --- | --- |
| Product/test/fixture build | Pass: `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config --nologo -v:q` |
| Focused fault-isolation tests | Pass: `dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -method "*TryIsolateRecompileFailure*"` |
| Existing source-probe regressions | Pass: `dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -method "*ReturnToSenderSourceProbe*"` |
| Generated RTS catalog tests | Pass: `dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -method "*ReturnToSenderCatalog*"` |
| Markdown | Pass: `npx markdownlint-cli docs/design/fact-planned-compile-back-harness.md` |
[...]

PR #2687, opened 2026-07-12
Fix decompiler and harness residue tail

### LINKED ISSUE #2658: Narrow legacy cluster extension roots by receiver compatibility
## Problem

The legacy `FidelityCheck` reconstruction closure resolves `CS1061` by looking up the missing member name across all static classes in the target module. That broad index is intentional for extension methods—the declaring root is not the receiver type—but common names such as `Add` and `GetAwaiter` can pull unrelated static roots into the closure and consume the root budget.

PR #2652 now provides structured receiver-type evidence for these diagnostics.

## Scope

- Enrich the legacy cluster method index with extension receiver compatibility (including generic and inherited receiver shapes).
- Use `ClosureDiagnosticReference.ContainingType` to select compatible extension-declaring roots without losing valid extension-method discovery.
- Preserve a safe fallback for metadata shapes that cannot be classified structurally, with explicit provenance.
- Add adversarial fixtures with same-named extension methods for unrelated receiver types, including implicit `Add` and `GetAwaiter` patterns.

## Acceptance criteria

- A `CS1061` adds only receiver-compatible extension roots when compatibility can be proven.
- Existing extension-method cluster rescues remain checkable.
- Root-budget and parity evidence show no regression.

Follow-up from review of #2652; tracked under #2646.

### LINKED ISSUE #2664: Preserve live labels when structuring conditional while loops
## Problem

`StructuringPass` preserves branch-target provenance when it synthesizes an
infinite `while (true)` loop, but the conditional-while path remains a sibling
frontier.

When `FindWhileShape` succeeds, `BuildWhileCondition` consumes the condition
block's branch nodes and the synthesized `WhileLoop` receives no source offset.
If an unconsumed external goto still targets that condition block, the printer
can retain the goto while losing the target label, producing `CS0159`.

Relevant path:

- `StructuringPass.BuildRegion` constructs `new WhileLoop(condition, body)`
  from `BuildWhileCondition(loop)`.
- `BuildWhileCondition` detaches the back/exit branch conditions.
- Unlike the infinite-loop path, the resulting loop is not stamped with the
  targeted block's offset.

## Expected

Conditional-loop structuring should preserve any live external target label at
a C#-legal enclosing scope, without stamping an inner statement or emitting a
label when all incoming branches were consumed.

## Acceptance

- Add a synthetic external-target fixture for a conditional `while`.
- Pin the label before the synthesized loop and verify every surviving goto has
  a label.
- Include ordinary and EH-adjacent near misses so moving the label outward does
  not cross an illegal scope.
- Run render-text A/B and classify all changed methods.

Found during follow-up review of #2649.

richlander:
Follow-up to #2631
Closes #2664
Closes #2658
Follow-up to #2666

## Change

**Conclusion: PASS** — closes the small residue tail without corpus render
movement:

- raises the .NET 11 class-union not-pattern lowering back to
  `return pet is not Cat;`, while a non-union `pet.Value is not Cat` stays flat;
- preserves externally targeted conditional-loop labels and declines an
  unrepresentable externally targeted second compound latch;
- makes fidelity-cause inspection fail honest before Full/Partial folding;
- narrows CS1061 closure expansion to actual, receiver-compatible extension
  methods, including inherited and generic-interface receivers, with explicit
  fallback provenance when metadata is insufficient;
- moves allocation Finding identity to the length-prefixed `CompositeKey`.

Before:

```csharp
return !(!pet ? false : pet.Value is Cat);
```

After:

```csharp
return pet is not Cat;
```

The #2629 and #2632 baseline repros already pass current `main`; no product
change was needed for those rows.

## Evidence

| Check | Result |
| --- | --- |
| Solution build | Pass |
| Non-AOT product build | Pass |
| Analysis suite | 422 passed on reconciled head |
| Decompiler suite | 2,489 passed before the final main-only reconciliation |
| Fast decompiler suite | 2,377 passed on reconciled head |
| Shape proof | Union positive/non-union near miss; live/consumed/compound/EH loop-label fixtures |
| Harness proof | Complete/Absent/Failed census, structured shell evidence, Add/GetAwaiter receiver adversaries |
| Render-text A/B | Explicit base `81ba0eec`; 89,065 methods; 0 changed, 0 added, 0 removed |
| Changed-method fidelity | No fixed-corpus render changes. The #2631 compiled fixture selects its exact expected spelling from the imported IL shape; a synthetic IR fixture separately pins the conditional matcher and provenance. Compile-back is not currently checkable because the reconstructed shell omits the union declaration context. |

### Decompiler quality diff

Corpus: #1166 real-world decompiler corpus sensor: #1150 pinned NuGet assemblies plus dotnet-inspect managed assemblies. 15 assemblies, 1,486 methods

Baseline staleness: corpus drifted from the pinned baseline (generated 2026-06-26). Aggregate count deltas below include this corpus change and are not a clean PR signal; rebaseline against current main (--emit-corpus-baseline) before trusting count-based regressions.
- total methods 1,461 -> 1,486 (+25)
- ILInspector.MetadataPrimitives: 61 -> 86 methods (+25)

| Fully raised (+) | 1,213 (83.03%) → 1,233 (82.97%) (bad) | +20 |
| Conditional-branch residual (-) | 35 (2.40%) → 44 (2.96%) (bad) | +9 |
| Forward-merge stops (-) | 27 (1.85%) → 32 (2.15%) (bad) | +5 |

Pinned-subset gate (PR quick rate/count regressions evaluated here): Fully raised 82.17% -> 82.33% (+0.16 pp); conditional residual 3.00% -> 3.00% (0 pp); fidelity ungated (sampling differs; rely on changed-method fidelity).

Advisory aggregate rate movement (not a PR quick hard gate; review for decompiler-risky changes):
- conditional-branch residual increased 0.56 pp (baseline 2.40%, PR 2.96%, tolerance 0.10 pp)
- forward-merge stop increased 0.30 pp (baseline 1.85%, PR 2.15%, tolerance 0.10 pp)

Per-method delta artifact: `/tmp/residue-pr-delta.json`

### Decompiler quality diff

Risk warning: thin correctness coverage (validity below 1.00% floor; fidelity below 0.10% floor). Add targeted improved examples and still-flat near misses; aggregate numbers are not sufficient alone.

| Semantic defects (sampling differs) | 54 (0.21%) → 2 (0.57%) | n/a |
| Fidelity opcode diffs (sampling differs) | 4 (11.11%) → 4 (11.11%) | n/a |
| Fidelity exact (sampling differs) | 32 (88.89%) → 32 (88.89%) | n/a |

Regressions:

Per-method delta artifact: `/tmp/residue-corpus-delta.json`

The full-card verdict is the documented sampling-cap artifact: current
`--compile-cap 25` is compared with a baseline captured at 4,000. Structural,
malformed, fidelity, and pass-bug rows are neutral; the hash-stable PR quick
pinned gate passes.

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | Clean | Final head `12a2c5ad`, isolated checkout |
| Gemini 3.1 Pro | Clean | Final head `12a2c5ad`, isolated checkout |

**Review conclusion: PASS** — both final-head adversarial reviews are clean.


richlander:

Final head: `07c9150a`

| Reviewer | Result |
| --- | --- |
| Claude Opus 4.8 | CLEAN |
| Gemini 3.1 Pro | CLEAN |

Earlier Gemini rounds found two actionable issues:

- The raised class-union outer negation could lose source provenance. `951b9e5d`
  now prefers the consumed outer-negation offset and falls back to the stamped
  type test.
- Broad synthetic-shell suppression could hide a real CS0029 over
  `this.GetInt()`. `7a232bc7` now suppresses only diagnostics located on the
  synthetic global `__Shell` `this`; invalid arguments and enclosing conversion
  diagnostics remain visible.

Gemini re-reviewed those fixes and returned clean. The final CI SDK exposed a
compiler-lowering split: preview.5 emits a guarded class-union form while
preview.7 emits the conditional-slot form. `07c9150a` accepts both compiler
spellings in the compiled fixture and adds deterministic synthetic IR coverage
for the preview.7 matcher and provenance. The exact preview.5 fast decompiler
suite passed 2,347 tests.

Both reviewers then independently reviewed final head `07c9150a` in clean,
isolated worktrees and found no outstanding correctness, security, or
compatibility issues. Opus also noted a pre-existing SDK-sensitive hardcoded IL
offset in `Structuring_PreservesFlattenedSwitchTargetLabels`; no resolution
commit applies because that assertion is unchanged from the merge base and the
structural goto/label invariant passes.

All required CI checks pass. **Ready to merge.**

richlander:
## Review — five residue items genuinely retired, but two product pass changes without corpus A/B and one untested closure path make this a hold; and it collides with three sibling PRs

Scoring against the residue catalog first: **(a) fold-then-inspect is fixed everywhere** — CorpusSensor, LibraryReport, and both Program loops now inspect the census before folding, inspection failures land in `passBugs`/`crashes` with `fidelity-inspection-failed` instead of masquerading as `fidelity: DEC000x` residuals, and `PrimaryBucket` now *throws* rather than fabricating a bucket for a failed inspection. **(b) Absent≠Failed** gets its `CensusState` enum with distinct error codes, tested. **(c) CS1061 receiver narrowing** lands with real machinery (extension-attribute-gated index, receiver-compatible root selection, provenance-carrying fallback). **(d) the conditional-while label sibling from #2649 is fixed** — the loop is stamped with the condition block's offset, the compound-latch fold correctly *declines* when the second test is externally targeted (no label point exists mid-`&&`), and the negative control pins that consumed edges emit no label. **(f) the allocation key moves onto `CompositeKey`.** Items (g) transition aliases and (h) subject-key overload suffix remain open; (e) was already fixed on main.

Three things before this merges:

**1. Two product output changes with synthetic-only evidence.** The StructuringPass changes and the new `TryMatchClassUnionConditionalNotTypeTest` raise both alter rendered C#, and the diff contains no corpus render A/B against the merge base — the evidence bar `docs/decompiler-raise-discipline.md` itself calls non-negotiable and every other raise PR in this series has met (#2683's 89K/0-delta being the sibling example from the same day). The semantics check out by hand (the null path is preserved through the conjunction), but hand-checking is what the A/B exists to replace.

**2. The CS1061 fix's *production* path is untested and has a silent-degradation hole.** Both new selection tests pass `compatibleReceiverTypes: null`, exercising only the metadata closure — the semantic-closure branch production actually feeds is uncovered. And that branch treats a semantic closure as authoritative (`complete = true`) even though `CompatibleReceiverTypes` is computed from the *partial cluster compilation*, whose base types may be unresolved error symbols — that's why the closure is still growing. A truncated hierarchy then silently drops the needed extension root with `UsedFallback == false`: the provenance mechanism built to make fallback visible is precisely the thing that doesn't fire in this case, and the row degrades to `ClusterBailed` with no signal. Either don't set `complete` when the receiver's hierarchy contains error symbols, or route the truncated case through the fallback path — plus one selection-level test through the semantic branch.

**3. The fixture assertion weakening quietly unpins the new matcher.** `IsNotCat`'s exact assertion became a two-spelling disjunction for SDK portability — same move as #2683, same cost: the compiled fixture no longer proves the new raise fires on *any* real SDK's IL; only the hand-built IR test does. If the matcher's shape guess drifts from what csc emits, this suite stays green. Same fix as I asked there: branch the assertion on the detectable IL shape, or state which SDK produces which spelling.

Smaller, ride-along grade: the collision half of the allocation-key test is theater (the asserted pair also produced distinct keys under the old format — the hardening is fine, the test just doesn't demonstrate a collision); the `SimpleNameFromDiagnosticMessage` deletion narrows CS0712/21/22/23 filtering with no test or count; and the `--dump` failure path now prints `ErrorCode` where it used to print the actual reason (the `Detail` is right there — print it like the other two paths do).

**Coordination — this PR collides with three open siblings, one of them semantically:**
- **#2684 rewrites the same `IsShellArtifact` logic in `ValidityCheck.cs`, more comprehensively** (enumerated ID gate + per-ID syntax + structural equivalence vs. this PR's this-token symbol check). These aren't just textual conflicts — they're two different answers to the same question. Land #2684 and drop this PR's shell commit in the rebase.
- **#2683 touches the same `TypeSourceComposerUnionTests` assertions** and raises the *slot-diamond* lowering of the same source shape this PR's matcher handles in the *conditional* lowering — complementary, but the shared test file needs one reconciled story.
- **#2674/#2686 vs. the RTS-harness files**: this PR's `FidelityCheck`/`CorpusSensor` changes sit next to #2675's, and the ReturnToSender loop consolidation (#2674) should land before anything else touches those loops.

The residue work itself is exactly what was asked and mostly well-executed — the fold-then-inspect fix alone makes the fidelity metrics honest. Hold for the A/B, the semantic-closure fix, and the rebase order; all three are one-round items.

---

richlander:
## Maintainer feedback resolution

Resolved in `66217fc3` and the final sibling reconciliation `1a79f362`.

1. **Render A/B:** the PR body already carried the original explicit-base
   89,065-method zero-delta result. I regenerated it after each main update; the
   final run is against merge base `81ba0eec` and reports 89,065 evaluated,
   0 changed, 0 added, 0 removed.
2. **Semantic receiver closure:** semantic evidence now carries an explicit
   completeness bit. Roslyn error types mark the hierarchy incomplete;
   incomplete evidence is supplemented from metadata, and when receiver
   metadata is unavailable unmatched extension roots use explicit fallback
   provenance rather than disappearing silently. Tests cover unresolved bases,
   semantic-incomplete/metadata-complete inherited and generic-interface
   selection, and semantic-incomplete/metadata-unavailable fallback.
3. **SDK-portable union fixture:** merged #2683 replaced the two-spelling
   disjunction with an assertion selected from the detected imported IL shape.
   This PR retains the separate synthetic proof for the post-fold conditional
   matcher and its source provenance.

Ride-along items:

- The allocation test no longer claims a collision the old format did not have;
  it now verifies only the requested `CompositeKey` serialization.
- `--dump --remarks` now includes the inspection `Detail` with its error code.
- Merged #2684 fully supersedes this PR's narrower validity implementation and
  tests; the merge keeps #2684's shared structured classifier and drops the old
  path from this branch.

Coordination is complete: #2683, #2684, #2675, and #2686 are all merged, and
`1a79f362` reconciles their overlapping files. Current evidence on that head:
solution and non-AOT builds pass; focused reconciliation tests pass 121; fast
Decompiler tests pass 2,377; Analysis passes 422; final render A/B is zero
delta. Exact-head Opus/Gemini re-review and CI are running, so this PR is not
yet re-marked ready.

richlander:
## Additional adversarial finding resolved

Opus and Gemini independently found the same receiver-closure gap while
reviewing `66217fc3`: array and type-parameter receivers were marked complete
even though their implicit `object`/base/interface hierarchy was not enumerated.
That could silently exclude a valid extension root.

Resolved in `0a38006f`:

- type-parameter and other non-named receiver closures are explicitly
  incomplete;
- metadata supplementation or explicit fallback now owns those missing
  hierarchy edges;
- array receiver provenance retains a readable rank-qualified type name;
- direct CS1061 evidence tests cover unconstrained `T` and `int[]`.

Targeted closure/selection tests pass 56/56, and the exact-head render A/B remains
neutral at 89,065 methods with zero changed, added, or removed. CI and two clean
exact-head reviews are pending.

richlander:
## Mixed compatible/unknown fallback finding resolved

Gemini found that an incomplete receiver closure could still lose a required
extension root when a universal/generic candidate was already classified
compatible: the compatible root won and unknown candidates were discarded.

Resolved in `53361f11`:

- when no unknown candidates remain, compatible roots return normally;
- when any candidate remains unknown, compatible and unknown roots are
  deduplicated and returned together;
- the combined selection is explicitly marked fallback with its reason, so
  uncertainty never becomes a silent compatibility claim;
- an `int[]` fixture with both `Add<T>(this T, string)` and
  `Add(this System.Array, int)` pins that both declaring roots are retained.

Targeted closure/selection tests pass 57/57. Exact-head render A/B remains
neutral at 89,065 methods with zero changed, added, or removed. CI and two clean
exact-head reviews are pending.

richlander:
## Interface receiver closure finding resolved

Opus found the final named-type sibling: Roslyn's interface hierarchy omits the
implicit conversion to `System.Object`, so a complete interface closure could
silently exclude a `this object` extension root.

Resolved in `12a2c5ad`:

- semantic interface receiver closures include `System.Object`;
- metadata interface receiver closures include the same implicit base;
- a direct semantic-evidence test pins the complete interface closure;
- selection tests prove both metadata-only and semantic-complete paths retain an
  object extension without fallback.

Targeted closure/selection tests pass 59/59. Exact-head render A/B remains
neutral at 89,065 methods with zero changed, added, or removed. CI and two clean
exact-head reviews are pending.

richlander:
## Final review reconciliation

Final head: `12a2c5ad`

| Reviewer | Result |
| --- | --- |
| Claude Opus 4.8 | CLEAN |
| Gemini 3.1 Pro | CLEAN |

The maintainer feedback and subsequent adversarial rounds produced four
actionable receiver-evidence fixes:

- `66217fc3` adds explicit semantic closure completeness, metadata
  supplementation, and visible fallback provenance.
- `0a38006f` marks arrays, type parameters, and other non-named receivers
  incomplete and preserves readable array provenance.
- `53361f11` retains unknown roots alongside compatible roots whenever
  uncertainty remains, marking the combined selection as fallback.
- `12a2c5ad` adds the implicit `System.Object` edge to semantic and metadata
  interface closures.

The sibling coordination is also complete: `1a79f362` merged the landed #2683,
#2684, #2675, and #2686 changes, dropped this PR's superseded validity
implementation in favor of #2684, retained inspect-before-fold in
`LibraryReport`, and reconciled the class-union fixture with #2683's
imported-shape-aware assertion.

Explicit non-actions:

- The original "missing A/B" claim required no code fix because the PR already
  carried a 89,065-method zero-delta A/B; it was regenerated after each main
  update. Final base `81ba0eec`: 89,065 evaluated, 0 changed, 0 added, 0 removed.
- Gemini's params-array allocation note for `CompositeKey` is a non-blocking
  performance optimization, not a correctness/compatibility defect. No
  resolution commit applies; the shared helper intentionally exposes a
  `params` contract today and any allocation optimization belongs in measured
  performance triage rather than this residue correctness PR.

Final evidence:

- solution and non-AOT builds pass;
- focused reconciliation tests pass 121;
- receiver closure/selection tests pass 59;
- fast Decompiler suite passes 2,377;
- Analysis suite passes 422;
- all required CI checks pass;
- both exact-head isolated reviews are clean.

**Ready to merge.**

richlander:
Re-reviewed at head 12a2c5a. **All three held asks landed — hold cleared, safe to merge.** CI is green at this head.

**Ask 1, corpus A/B: landed.** The PR body's Evidence table now carries the render-text A/B against the explicit merge base `81ba0eec` — 89,065 methods, 0 changed/added/removed — with quality-diff cards (quick gate + full fixed corpus) and re-assertion at each new head, which is the regeneration cadence the doctrine wants. The quick-gate card's +9 conditional-residual / +5 forward-merge rows are corroborated as baseline drift (+25 corpus methods from MetadataPrimitives) by the neutral 89K fixed-corpus card and the 0 pp hash-stable pinned subset. One honest caveat, not blocking: zero delta means the corpus exercises *neither* changed product pass — the StructuringPass and UnionSwitchExpressionPass changes rest on synthetic IR fixtures, which the PR itself discloses.

**Ask 2, CS1061 semantic-closure fail-open: genuinely fixed.** The old `bool complete = semanticClosure is { Count: > 0 } || ...` — where any non-empty closure built from a partial cluster compilation was authoritative — is gone. `CompatibleReceiverTypes` now computes an explicit completeness bit: error types (recursing through type arguments, array elements, pointers), type parameters, and all non-named receivers mark the closure incomplete; incomplete semantic evidence is supplemented from original-assembly metadata; and residual uncertainty returns `UsedFallback: true` with concrete provenance (`"receiver hierarchy incomplete for {receiver}"`) surfaced on the result row. All four paths are pinned by tests. Truncated-hierarchy-as-complete no longer exists.

**Ask 3, rebase vs #2683/#2684: landed with zero delta.** `git diff origin/main..head` over the `IsShellArtifact` files is empty, the whole 16-file diff has no shell-related lines, and this PR's narrower `7a232bc` suppression plus its `ValidityShellNoiseTests` addition were dropped in the merge in favor of main's #2684 version, exactly as your resolution comment stated.

**The three post-hold commits are real hardening, not motion.** "Keep non-named receiver fallback visible" correctly routes arrays/type-parameters through metadata supplement or visible fallback (they can never prove a complete closure semantically). "Retain unknown extension fallback roots" fixes a fail-open in the PR's own earlier design — a universal `Add<T>(this T, ...)` candidate could silently eclipse an unproven `Add(this System.Array, ...)` root; now uncertainty never becomes a silent compatibility claim, pinned with the mixed `int[]` fixture. "Include object in interface receiver closure" is correct C# applicability, applied symmetrically on the semantic and metadata sides, and matters because complete closures suppress the fallback safety net — the claim is now actually true for interfaces.

Follow-up material (none blocking, all harness-side and visibly degrading at the row level):

1. **Arity-erased receiver-type index, first-wins** — `receiverTypes.TryAdd(NormalizeReceiverTypeName(...), handle)` collapses `Result` and `Result<T>` to one arbitrary entry, so `BuildReceiverClosure` can walk the wrong type's hierarchy and return `complete=true`, which suppresses the very fallback provenance built to catch exclusions. The one residual silent-exclusion shape left; worth an arity-aware key or a pinning test.
2. `complete = metadataComplete` overwrites rather than ANDs the semantic bit — defensible (original metadata is authoritative for a named receiver), but it's what lets item 1's wrong-typedef walk override a semantically-flagged-incomplete closure.
3. The `TryMatchClassUnionConditionalNotTypeTest` raise has synthetic-IR-only positive coverage in this PR; whether real IL exercises it depends on the ambient SDK's lowering. Disclosed in the PR — fine — but if csc's slot-diamond shape drifts, no compiled test here turns red.

---

Issue #2604, opened 2026-07-10
Finding spine: arity ladder + verb-nominalization envelopes (retire *FindingsResult, one generic FindingComparison<T>)


richlander:
Refs #2585. Naming/shape contract for the Finding spine, settled in design discussion; filing so it lands as doctrine before the producer fleet finishes rebasing. Touches #2597/#2598/#2601 and every future producer.

*Amended per the review discussion below: `FindingComparison<T>` is a closed union (not a flat record), `CorrelatedFinding` requires version-labeled pairs, and the bisect semantics distinguish occurrence-level from subject-level disappearance.*

## The model: two type families

**1. Information types — the arity ladder (the goods).** Facts about the *subject*. Durable currency: serialized, stored, folded, aggregated. Named bare, noun-first, prime names:

| Type | Arity | Is |
| --- | --- | --- |
| `Finding<T>` | N = 1 | one observation |
| `PairFinding<T>` | N = 2 | one classified transition, composed of Findings |
| `CorrelatedFinding<T>` (future) | N > 2 | one subject's timeline: current + older + older…, composed of **version-labeled** PairFindings |

Boundary discipline: a 2-version diff is never wrapped in a `CorrelatedFinding`; a lone transition is just a `PairFinding`.

**2. Operation outcomes — the envelopes (the receipts).** Facts about an *invocation*: what a caller needs to interpret one API call. Per-call, disposable; no fold consumes them, no report row is one. Named as the **nominalization of their verb**:

| Verb | Envelope | Carries (the goods) |
| --- | --- | --- |
| `Match` | `FindingMatch` | edges + fringe |
| `Inspect` | `FindingInspection<T>` | `Complete(Finding<T>[])` \| `Absent` \| `Failed` |
| `Compare` | **`FindingComparison<T>`** | `Complete(pairs, match, inspections)` \| `Failed(inspections)` |
| `Correlate` (future) | `FindingCorrelation<T>` | `CorrelatedFinding<T>[]` + per-feed inspections |

**Dependency invariant:** envelopes carry ladder types; the ladder never references an envelope.

### 1. Retire `*FindingsResult` → one generic `FindingComparison<T>` union in the leaf

`Result` is a filler suffix (same family as `Engine`/`Manager` — says *an operation happened*, not *which*), and the per-producer result types are converging on an identical, fully domain-free shape anyway.

[...]

PR #2691, opened 2026-07-12
Define CLI command transition model

richlander:
## Conclusion

Define the command-transition doctrine before adding correlation UX.

## Model

- Separate source context, structural focus/zoom, operation/arity, lens,
  traversal policy, and output projection.
- Change commands when subject identity or operation arity changes; keep lens
  and rendering changes within the current command.
- Treat `type -> member` as structural zoom.
- Treat `inspect -> diff -> timeline` as operation/arity transitions over the
  same source and focus.
- Define `Package@A..B` as an address space. The active operation owns payload
  cardinality: enumerate zero cells, inspect one addressed cell, compare two
  endpoints, or correlate N explicit probes.
- Preserve explicit acquisition backpressure and distinguish unevaluated,
  missing, absent, and failed cells.

The document evaluates where ranges make sense today on `package`, `type`, and
`member`; records the current `library` limitation; and concludes that a future
`timeline` should be a peer to `diff`, not a mode or section under unary
inspection.

## Scope

Docs only. This PR does not add `timeline`, change command syntax, or alter
range resolution.

## Validation

- `markdownlint` passed for all changed Markdown.
- Architecture review verified current claims against the command parsers,
  package-vector implementation, and Finding correlation semantics.

Context: #2625 and #2604.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -49,6 +49,13 @@ The tool is organized around source inspection, API lookup, relationship, and ut
 └─────────────────────────────────────────────────────────────┘
 ```
 
+The command surface has independent source, focus, operation, lens, traversal,
+and rendering axes. Noun-first commands (`package`, `type`, `member`) select a
+structural focus for unary inspection; operation-first commands (`diff`, and a
+future `timeline`) change arity while retaining source/focus selectors. See
+[Command Transition Model](design/command-transition-model.md) for the decision
+rules and version-range cardinality contract.
+
 ### package
 
 Inspects NuGet package metadata without extracting libraries:
--- a/docs/design/command-transition-model.md
+++ b/docs/design/command-transition-model.md
@@ -200,6 +200,7 @@ package payloads. Once selected, the normal output-shape rules apply:
 | `--urls` | Project URL-bearing rows to a URL Vector. | None. Valid only if the version-row schema exposes a URL. |
 | `--print` | Resolve a printable payload already referenced by one selected row. | May fetch that declared payload at the same evaluated address; must not add or evaluate another source address. |
 | `--print-all` | Resolve every declared printable row payload in stable row order. | Explicitly authorizes payload fan-out only for already selected/evaluated rows; must not evaluate source addresses. |
+| `--head N` / `--tail N` | Clip rendered output lines after projection. | None. They do not select printable rows or limit payload fetches. |
 
 Shape reducers do not revise operation arity. In particular:
 
@@ -211,6 +212,12 @@ Shape reducers do not revise operation arity. In particular:
 - `--print` is exactly-one, not implicit-first: one printable row prints
   directly, multiple printable rows require `--row N` or `--print-all`, and zero
   printable rows reject;
+- `--head N` and `--tail N` run after print selection and fetching, so
+  `--print --head 1` does not select the first printable row and
+  `--print-all --head 1` still authorizes all declared payload fetches;
+- `--rows --head N` and the symmetric `--rows --tail N` are first/last
+  table-row rendering windows and remain incompatible with `--print` and
+  `--print-all`; `--row N|first|last` selects exactly one printable row;
 - a plain version string has no printable document. `--print` and `--print-all`
   must report that the selected shape is not printable rather than silently
   transition from version-address rows to package artifact inspection. The
--- a/docs/design/command-transition-model.md
+++ b/docs/design/command-transition-model.md
@@ -1,0 +1,329 @@
+# Command Transition Model
+
+`dotnet-inspect` has two kinds of top-level commands today:
+
+- noun-first inspection commands such as `package`, `library`, `type`, and
+  `member`;
+- operation-first commands such as `diff`.
+
+That split is useful, but only if transitions between commands follow an
+explicit model. A user should be able to predict whether a new gesture changes
+the subject, the operation, the lens, or only the rendering.
+
+The governing rule is:
+
+> One transition should change one axis. Change commands when subject identity
+> or operation arity changes; use options and sections for context, lenses, and
+> projection.
+
+## Independent axes
+
+| Axis | Question | Examples | CLI shape |
+| --- | --- | --- | --- |
+| Source context | Where is the subject acquired from? | package, platform, local library, restored project, TFM | Named options such as `--package`, `--platform`, `--library`, `--project`, `--tfm` |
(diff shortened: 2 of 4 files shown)


richlander:
## Review — the doc that makes `timeline` derivable instead of ad hoc; endorse

This is the right document at the right moment: the CLI has been growing gestures one PR at a time (`--at`, `@A..B`, `--finding`, composed sections, Finding Transitions' select-alone rule), each locally justified — this doc extracts the model they were all implicitly following and makes the next operation *derivable*. Things I checked and value:

- **The six axes are genuinely independent and match shipped behavior.** Every merged decision I've reviewed lands cleanly on one axis: `--finding` descriptors are lens, `--at` is traversal, `-S` composition is lens+projection, the unaddressed-unary-range error and validate-before-acquisition are the cardinality contract. I couldn't find a merged behavior the model retrodicts wrongly — which is the test of a descriptive-turned-normative doc.
- **"A version range is an address space" with a per-operation cardinality table (Enumerate 0 / Inspect 1 / Compare 2 / Correlate N)** is the doc's best contribution: the same syntax participates in every command without changing meaning, and each range-consuming path must state how many payload cells it may acquire. That single rule kills both failure modes we've seen adjacent lenses flirt with — implicit aggregate views and silent endpoint-picking.
- **The timeline section preserves every hard-won invariant** from #2604/#2625: the five cell states with `Unevaluated` explicitly a *presentation state* formed by joining the vector with the sparse correlation ("not fabricated as a Finding or inspection outcome" — the manufactured-match rule applied to a new surface before anyone could violate it); probe order not defining timeline order (positions from the resolved vector); recommend-don't-acquire bisect preserving the caller-owned probe budget; and "a timeline locates a candidate boundary; it does not claim introduction" — the nomenclature doc's probe-vs-proof line carried forward, so `timeline` composes with Finding Transitions rather than competing with it.
- **The decision checklist ends with the doctrine question** ("are unevaluated, absent, missing, and failed states kept distinct?") — command design and Finding semantics converging on one rubric, and the question-4 pairing rule (schema change alone discriminates nothing) heads off the most common category error.
- **The non-goals are as load-bearing as the goals**: no `inspect` command until composition demands it, no session state, no automatic range scans, and — critically — deferring `timeline` *syntax* until its producer/view contract is designed, while making that contract's constraints explicit now.

Two small suggestions, neither blocking: cross-link `finding-adoption.md` rule 5 from the pairwise-confirmation example (the two docs state the same native-case rule from opposite sides, and the link makes the pairing discoverable); and the `library --package` paragraph's "must adopt the same explicit one-cell address rule" reads as a latent requirement with no tracking hook — one sentence naming it as a checked constraint (or an issue reference) keeps it from becoming the un-migrated corner the next audit finds.

Endorse. With this merged, the correlation-UX item on the project list stops being "design a timeline" and becomes "implement the contract this doc already wrote."

---

richlander:
Follow-up on the output-shape feedback is in 65ce1dda.

The architecture now classifies `--versions` as a package lens selecting a zero-package-payload Version Vector, not as an operation peer of `diff` or `timeline`. `--count`, `--urls`, and `--print` are downstream shape operations: count reduces the selected Vector, URLs require a URL-bearing row schema without extra package acquisition, and print cannot reinterpret a version-address row as permission to inspect/download that package. The same boundary prevents timeline reducers/projectors from probing unevaluated cells.

The doc also records the current specialized `--versions` early-exit writer—which bypasses some shared reducers/projectors—as an implementation gap against the output-shape model, not architectural precedent.

richlander:
The type/member distinction exposed one more independent axis, addressed in 6bf445bd.

A type-focused pairwise operation can select either (1) a type-identity census, answering whether the type was added or removed, or (2) a type-scoped member census, answering which members were added or removed. Both retain type focus and differ by observation producer. Member focus instead addresses one particular member and asks for that member transition; it is not a substitute for the collection census.

The architecture now names **Observation / census** separately from focus/zoom, operation/arity, and representation. This also means diff and future timeline operate over the selected producer census rather than assuming one Finding identity per focused subject.

richlander:
Captured in d2152e4c. The architecture now identifies a type-focused member census across a package range as the most informative initial timeline composition: the type is the stable scope, members are observation identities, and adjacent complete censuses yield native Added/Removed transitions for each member track.

It also separates dense timeline traversal from bisect. An explicit bounded full traversal may authorize all N package acquisitions and produce a true transition timeline, including removal and re-addition. Sparse traversal evaluates only selected cells, brackets candidate boundaries, and uses bisect only as a recommendation policy. A range alone still authorizes no package scan.

richlander:
Captured in 93defe06. The model now separates a focus domain from a point selector/coordinate. IL-offset facts are reachable from multiple structural scopes: library uses a composite MethodDef+offset coordinate that discovers the containing member, while member-focused body views already carry member identity and expose peer offset-scoped facts.

This makes focus a graph rather than a mandatory package -> library -> type -> member -> offset ladder. Commands represent independently navigable focus domains; a complete selector may refine an established domain directly without requiring an offset command or every intermediate zoom step.

richlander:
All architecture feedback is reconciled at 93defe06, and the docs-only checks are green. This is low-risk documentation work; no adversarial review is required. Ready to merge.

richlander:
Vector printing is now locked down in ed10b05b and aligned with the existing tested projection behavior.

`--print` is exactly-one, never implicit-first: one printable row prints; multiple printable rows require `--row N` or `--print-all`; zero printable rows reject. `--print-all` is the explicit fan-out gesture. Printability belongs to each row rather than to Vector shape, so neither flag may reinterpret version addresses as package artifacts or evaluate unevaluated timeline cells.

richlander:
The agreed line/row-window grammar is documented in f27d5d50:

- `--head N` / `--tail N`: rendered-line windows after projection and fetching.
- `--rows --head N` / `--rows --tail N`: first/last data-row windows per table.
- `--row N|first|last`: exactly one printable-row selector.

Row windows remain incompatible with `--print` and `--print-all`, so they cannot resolve print cardinality or limit fan-out. Implementation follow-up is tracked in #2694.

richlander:
Type transition observation scope is now explicit in 496f3ff7.

A type-focused operation may compare three different relationships: self presence (was the type added/removed), owned children (which declared members changed), or attached facets (which applied attribute occurrences changed). These remain the same type focus and diff/timeline operation; the observation producer changes. Human output may compose clearly labelled sections, while focused/machine output must identify one census explicitly. Member focus cannot substitute for the type-scoped member census, and attributes do not imply a new attribute command.

PR #2696, opened 2026-07-12
Stabilize corpus sensor sampling

richlander:
## Summary
- make corpus semantic validity sampling hash-stable after syntax filtering instead of consuming the cap by parallel race
- make corpus compile-back and RTS parity use hash-stable target attempts and preserve target-attempt order before applying caps
- update the real-world quality-card command to match the checked-in baseline compile cap

## Validation
- dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.CorpusSensorComparisonTests
- dotnet build tools/DecompilerHarness/DecompilerHarness.csproj -c Release --configfile nuget.config --verbosity quiet
- npx markdownlint-cli tools/DecompilerHarness/README.md
- dotnet run --project tools/DecompilerHarness -c Release -- artifacts/bin/ILInspector.Decompiler/release/ILInspector.Decompiler.dll --emit-corpus-snapshot /tmp/corpus-sensor-health-smoke.json --compile-cap 1 --corpus-fidelity-cap 1 --corpus-method-cap 10 --max-examples 1

## Notes
This removes the documented cap mismatch that produced the recurring real-world quality-card advisory and gives tolerance-0 rows such as RTS parity a deterministic target population when corpus inputs and caps match.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CorpusSensorComparisonTests.cs
+++ b/src/ILInspector.Decompiler.Tests/CorpusSensorComparisonTests.cs
@@ -223,6 +223,26 @@ public class CorpusSensorComparisonTests
     }
 
     [Fact]
+    public void DeterministicCompileBackTargetAttempts_UsesStableSampleRegardlessOfInputOrder()
+    {
+        string assemblyPath = Path.Combine(Environment.CurrentDirectory, "pinned.dll");
+        var methods = Enumerable.Range(0, 105)
+            .Select(i => SnapshotMethod($"M{i:000}", assemblyPath: "pinned.dll", fidelityCheck: "not-sampled"))
+            .Append(SnapshotMethod("<Generated>", assemblyPath: "pinned.dll", fidelityCheck: "not-sampled"))
+            .ToArray();
+
+        var forward = CorpusSensor.DeterministicCompileBackTargetAttemptsForTesting(methods, assemblyPath, cap: 1);
+        var reversed = CorpusSensor.DeterministicCompileBackTargetAttemptsForTesting(methods.Reverse().ToArray(), assemblyPath, cap: 1);
+
+        Assert.Equal(100, forward.Count);
+        Assert.Equal(
+            forward.Select(target => $"{target.Type}::{target.Method}{target.Signature}"),
+            reversed.Select(target => $"{target.Type}::{target.Method}{target.Signature}"));
+        Assert.DoesNotContain(forward, target => target.Method.Contains('<', StringComparison.Ordinal));
+        Assert.All(forward, target => Assert.Equal(assemblyPath, target.AssemblyPath));
+    }
+
+    [Fact]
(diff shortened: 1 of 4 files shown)


richlander:
Adversarial review reconciliation for `d2965bd6`:

- Claude Opus 4.8 initial review: CLEAN at `a31e6e9b`; verified deterministic validity sampling, path identity, target ordering, RTS parity alignment, cap edge cases, and local focused tests.
- Gemini 3.1 Pro initial review: found two actionable compile-back sampling issues: the deterministic target path could evaluate the full inflated attempt pool, and compile-back failure buckets used that full pool.
- Follow-up: fixed both in `d2965bd6` by evaluating deterministic target attempts only until the useful sample is filled and by scoping `AllResults`/failure buckets to that evaluated prefix.
- Claude Opus 4.8 re-review: CLEAN at `d2965bd6`; confirmed both Gemini findings resolved, target order preserved across batches, cap edge cases safe, and RTS parity still aligned.
- Gemini 3.1 Pro re-review: CLEAN at `d2965bd6`; confirmed prefix isolation, order preservation, RTS parity, cap bounds, sparse multi-batch behavior, and deterministic semantic-validity cap handling.

CI is green and merge state is clean. Ready to merge.

richlander:
Reviewed at head d2965bd6. **Endorse — safe to merge.**

This lands priority item "corpus sensor health" cleanly on both fronts:

**Deterministic sampling.** `DeterministicCompileBackTargetAttempts` ordering by `StableMethodHash` (FNV-1a over both UTF-16 bytes) then ordinal key replaces enumeration-order sampling, and `EvaluateTargetsInAttemptOrderUntilUseful` makes the sample a function of the corpus rather than of thread scheduling. The ValidityCheck change is the stronger one: the old design let early-finishing evaluation threads consume the cap, so the semantic subset depended on runtime timing. The two-phase replacement — syntax for all candidates in parallel, then a hash-stable `OrderBy(StableValidityHash).ThenBy(StableValidityKey).Take(cap)` semantic subset — removes the race by construction instead of trying to synchronize it. Right fix.

**`AllResults` unification.** Having a single `FidelityCheck.Evaluate` pass feed both the useful sample and the all-results bucket fixes the #2675(d) asymmetry (useful and all buckets previously came from separate evaluations that could disagree) and deletes a duplicate evaluation. Good.

**Cap re-alignment.** The documented validity invocation cap moving 25→4000 resolves the standing "validity cap below baseline comparability" advisory — sensor runs are now comparable to baseline again.

Two nits, neither blocking:

1. The FNV-1a implementation now exists in both `CorpusSensor.cs` and `ValidityCheck.cs`. Two instances is tolerable; per the repo's third-instance rule, the next consumer should trigger extraction of a shared `StableHash` helper rather than a third copy.
2. Attempt-pool underfill (fewer eligible targets than cap) behaves as before — fine, just noting it was checked.

---

PR #2698, opened 2026-07-12
Replace effective-discovery text parsing with render manifests

richlander:
## Summary

- capture section headings, table columns, and field rows directly from typed Markout formatter events
- share type/member document composition between normal rendering and effective discovery
- use one render manifest across type/member, package, and library discovery, removing Markdown heading/table parsing and whole-document substring matching

This keeps discovery coupled to the serializer contract rather than Markdown syntax while preserving the existing CLI output. Part of #2653.

## Validation

- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-build` — 1,798 passed, 10 environment skips
- `dotnet publish src/dotnet-inspect/dotnet-inspect.csproj -c Release -r linux-x64`
- `npx markdownlint-cli docs/design/schema-query.md`
- byte-identical stdout/stderr and exit codes against the merge-base for seven effective-discovery cases spanning library, package, type, and member commands


richlander:

Reviewed independently by **Claude Opus 4.8** and **Gemini 3.1 Pro** on isolated worktrees.

### Round one (`1e09b906`)

- **Opus:** CLEAN after broad base/head command-output comparisons across type, member, package, and library discovery.
- **Gemini:** identified that the reusable manifest formatter retained its last heading between the two type probe documents. I reproduced this against the clean exact head: a headless `Kind` field from document two was incorrectly attributed to the last section from document one (`Methods` fields became `MethodOnly,Kind`).
- **Resolution:** fixed in [`b4141340`](https://github.com/richlander/dotnet-inspect/commit/b414134065826d13b582ac10755dfa45d1e37c26) by resetting document-scoped heading and streaming-table state before each serialization while retaining accumulated manifest data. Added a regression test proving a headless field in the next document cannot leak into the prior section.
- **Gemini also suggested** unioning columns from every table emitted under one section. No change was made: the removed Markdown parser intentionally kept the first table under each heading, current same-section view alternatives are mutually exclusive, repeated two-document probes use the same active column options, and no incorrect discovery command/repro was found.

### Exact fixed-head re-review (`b4141340`)

- **Claude Opus 4.8:** CLEAN. Verified document reset, formatter dispatch/scoping, package/library fields, type/member opt-in and empty sections, first-table compatibility, builds, and focused discovery tests.
- **Gemini 3.1 Pro:** CLEAN. Confirmed the leak is sealed without manifest data loss and confirmed no current serializer path requires multi-table column unioning.

Both reviewers therefore agree the fixed head has no outstanding correctness, security, or compatibility findings. The union suggestion has no resolution commit because it would change the preserved first-table contract without a demonstrated failing path.

richlander:
Ready to merge

richlander:
Reviewed at head b4141340. **One small fix, then merge.**

This is the right retirement done at the right layer, continuing the #2652/#2684 line of removing rendered text as a decision input. All three text parsers are deleted: the heading/table markdown re-parsing in `DiscoverOutput`, the pipe-cell splitting in `LibraryCommand` (`GetRenderedFieldLabels`/`ExtractSection`/`IsMarkdownSeparatorCell`), and the crudest — `PackageCommand.FilterSchemaToEffectiveFields`'s whole-document case-insensitive substring match. The replacement runs the same serializer through a capturing `RenderManifestFormatter` into a typed `RenderedSectionManifest`, so discovery consumes structured events with no markdown round-trip.

The headline honesty win is properly pinned: renaming the fixture to `Test.Published.PackageInfoDiscovery` so the package ID contains a schema field name, with `Assert.DoesNotContain("| Published | field |", ...)`, is a real regression pin against the old substring matcher's false positive. The "Reset manifest context between documents" head commit fixes genuine (if mostly latent) cross-document heading leakage on the two-document type-discovery path and pins it at the formatter level. No statics anywhere in the new code — per-capture formatter instances make parallel test runs safe by construction.

**The pre-merge ask:** `EffectiveCategory = "effective-v10"` in `LibraryCommand` is unchanged. Cached field lists computed by the deleted pipe-cell parser (keyed on path + file size) will keep being served after the parser is gone. If the PR's premise is that the old parser could be wrong, its cached outputs shouldn't outlive it — bump to `effective-v11` so the manifest path recomputes everything.

Non-blocking, flagged for follow-up (both are parity with the old parser, but the typed manifest was the moment to fix them):

1. `FormatHeading` ignores `level`, so a table under a sub-heading is attributed to the sub-heading, `HasTable(section)` goes false, and the section gate fail-closed drops it. The old `StartsWith('#')` parser had the same flaw; a typed formatter can distinguish h2 from h4.
2. `IsFieldTable` decides by `headers[0].Equals("Field", OrdinalIgnoreCase)` — a name heuristic surviving inside a text-heuristic-retirement PR. A future field-style section whose first column isn't literally "Field" silently records no fields and falls open to the full schema.
3. `RecordTable`'s first-wins policy across the two merged type-discovery documents is fine as parity, but deserves a one-line comment stating it's deliberate.

---

richlander:
Addressed the pre-merge cache invalidation feedback in [`cf691cbd`](https://github.com/richlander/dotnet-inspect/commit/cf691cbd): `LibraryCommand.EffectiveCategory` is now `effective-v11`, so field lists produced by the retired Markdown parser cannot be served by the typed-manifest implementation. The existing versioned-category registration will retire `effective-v10` on cache maintenance.

richlander:
Re-reviewed at new head cf691cb. The cache invalidation landed exactly as asked — `EffectiveCategory` bumped `effective-v10` → `effective-v11`, so field lists computed by the deleted pipe-cell parser can no longer be served from cache. **Clear to merge.** The three non-blocking items (heading-level attribution, `IsFieldTable` name heuristic, first-table-wins comment) remain follow-up material.

---

richlander:
## Final exact-head review addendum

After the cache-category feedback was resolved in [`cf691cbd`](https://github.com/richlander/dotnet-inspect/commit/cf691cbd067ce72b5247241b5bdd06d29dbfd8de), both required reviewers re-reviewed that exact head in fresh isolated worktrees:

- **Claude Opus 4.8:** CLEAN. Confirmed all reads/writes use `effective-v11`, parser-era `effective-v10` entries necessarily miss, and `RegisterVersionedCategory("effective-v", "effective-v11")` safely retires older directories.
- **Gemini 3.1 Pro:** CLEAN. Confirmed the same invalidation/cleanup path and audited other caches; `PackageIndexCache` stores typed package metadata rather than discovery schemas, and no other cache persists render-manifest output.

There are no outstanding correctness, security, or compatibility findings on `cf691cbd`.

richlander:
Ready to merge

PR #2697, opened 2026-07-12
Expose analysis.unsafety transitions with descriptor-keyed retention

richlander:
## Conclusion

`diff --finding analysis.unsafety` now confirms whether one method gained,
retained, or lost definite unsafe IL operations using Analysis's native
`PairFinding<UnsafetyOccurrence>` comparison.

This is the fourth descriptor on the focused Finding Transitions lens and the
agreed refactor trigger from the #2671 review: retained comparisons now use one
typed, descriptor-keyed container instead of adding a third payload-specific
copy.

## What changed

- Added `analysis.unsafety` descriptor selection for exactly one method target.
- Renders unsafe operation kind/detail while keeping IL offsets as
  endpoint-local provenance.
- Replaced allocation/call-site retention records, flags, result lists,
  constructor branches, builder overloads, and merge branches with
  `RetainedFindingComparison<T>` plus `RetainedFindingComparisonSet`.
- Selects retained comparisons by descriptor ID, so future producers extend the
  container rather than the plumbing shape.
- Preserves opt-in cost: direct-call indexing is still acquired only when the
  call-site descriptor is retained; ordinary Body Signals output is unchanged.
- Updated Finding design, version-resolution, compatibility, and correctness
  guidance.

```bash
dotnet-inspect diff --package Foo@1.4.0..1.5.0 \
  -t Foo.Parser -m Parse \
  --finding analysis.unsafety
```

`PairFinding.Added` is the unsafe-operation introduction proof. `Present` and
`Removed` distinguish persistence from disappearance.

## Evidence

| Check | Result |
| --- | --- |
| Release solution build | Pass |
| `ILInspector.Analysis.Tests` | 421 passed |
| `ILInspector.Research.Tests` | 99 passed |
| `dotnet-inspect.Tests` | 1,803 passed, 10 tool-dependent skips |
| Changed Markdown | markdownlint clean |

Focused coverage includes native Added/Present/empty unsafety censuses,
descriptor/payload type enforcement, multi-descriptor merge, pre-acquisition
validation, table output, composed JSON, and real CLI execution against the
compiled V1/V2 fixture pair.

## Review

Two isolated exact-head adversarial reviews are clean:

| Reviewer | Result |
| --- | --- |
| Claude Opus 4.8 | CLEAN |
| Gemini 3.1 Pro | CLEAN |

Both verified descriptor/payload type safety, bucket-preserving Combine
semantics, native unsafety identity, duplicate multiset behavior, opt-in
[...]

PR #2699, opened 2026-07-12
Add type-focused package timelines

richlander:
## Summary

- add top-level `timeline` correlation over inclusive package version vectors
- support type presence, owned-member, and applied-attribute Finding censuses
- preserve unevaluated, missing, subject-absent, failed, dense, and sparse-gap semantics
- align pairwise `diff --finding` with the same type-scoped member and attribute observations

## Acquisition and output

- no `--at` resolves version metadata but acquires zero package payloads
- repeated `--at` evaluates only the selected sparse cells; `--at all` explicitly authorizes dense traversal
- `Evaluations` and `Transitions` compose with selection, count, table, TSV, JSONL, JSON, and row windows
- gap-spanning transitions are qualified and never claim an exact onset/removal version

## Validation

- solution build
- full Metadata executable suite: 432 total, 0 failed, 1 known skip
- full CLI executable suite: 1806 total, 0 failed, 10 tool-availability skips
- live zero-payload, sparse, dense, reverse-range, missing-type, count, and JSONL runs

## Review

Two isolated adversarial reviews completed cleanly on `285f66c6`: Gemini 3.1
Pro and Claude Opus 4.8. The attributed reconciliation is in the PR comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,6 +304,7 @@ dotnet-inspect member JsonSerializer --package System.Text.Json Serialize:1 -S "
 dotnet-inspect member MyType MyMethod:1 --library MyLib.dll -S "Unsafe*"
 dotnet-inspect library System.Text.Json --il-offset 0x06000004+0x15
 dotnet-inspect diff --package System.Text.Json@9.0.0..10.0.0 --breaking
+dotnet-inspect timeline --package System.Text.Json@8.0.0..9.0.0 --type System.Text.Json.JsonSerializer --members --at all
 dotnet-inspect diff --library old/Foo.dll..new/Foo.dll -S "Implementation Diff" -m MyType.HotPath
 dotnet-inspect depends Stream --markdown --mermaid
 dotnet-inspect implements IEquatable --project ./src/App -v:q
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -38,6 +38,9 @@ The tool is organized around source inspection, API lookup, relationship, and ut
 │                         diff                                 │
 │  Compare API, analysis, or implementation evidence           │
 ├─────────────────────────────────────────────────────────────┤
+│                       timeline                               │
+│  Correlate type-owned Findings across a package range         │
+├─────────────────────────────────────────────────────────────┤
 │            depends / extensions / implements                 │
 │  Relationship discovery for APIs, packages, and libraries    │
 ├─────────────────────────────────────────────────────────────┤
@@ -51,8 +54,8 @@ The tool is organized around source inspection, API lookup, relationship, and ut
 
 The command surface has independent source, focus, operation, lens, traversal,
 and rendering axes. Noun-first commands (`package`, `type`, `member`) select a
-structural focus for unary inspection; operation-first commands (`diff`, and a
-future `timeline`) change arity while retaining source/focus selectors. See
+structural focus for unary inspection; operation-first commands (`diff` and
+`timeline`) change arity while retaining source/focus selectors. See
 [Command Transition Model](design/command-transition-model.md) for the decision
 rules and version-range cardinality contract.
 
@@ -104,6 +107,21 @@ Compares two package, platform, or local-library versions:
 - Multiple selected sections compose in Markdown and JSON. Explicit
   table/TSV/JSONL output requires exactly one selected section.
 - Version range syntax is `Package@v1..v2` or `old/Foo.dll..new/Foo.dll`.
+
+### timeline
+
+Correlates one type-focused Finding census across an inclusive package version
+vector:
+
+- `--finding api.type` observes the type's own presence and facets.
+- `--finding api.member` observes all members owned by the type.
+- `--finding api.attribute` observes applied attribute occurrences.
+- No `--at` evaluates zero package payloads; repeated `--at` selectors perform
+  sparse probes; `--at all` explicitly authorizes dense traversal.
+- `Evaluations` preserves `Present`/`Missing` self-presence and
+  `Complete`/`SubjectAbsent` owned censuses, plus `Failed` and `Unevaluated`
+  cells. `Transitions` compares adjacent evaluated cells; a
+  gap-spanning row is qualified and never claims an exact transition version.
 
 ### find
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -118,9 +118,11 @@ vector:
 - `--finding api.attribute` observes applied attribute occurrences.
 - No `--at` evaluates zero package payloads; repeated `--at` selectors perform
   sparse probes; `--at all` explicitly authorizes dense traversal.
-- `Evaluations` preserves `Present`/`Missing` self-presence and
-  `Complete`/`SubjectAbsent` owned censuses, plus `Failed` and `Unevaluated`
-  cells. `Transitions` compares adjacent evaluated cells; a
+- `Evaluations` preserves `Present`/`Missing` self-presence through an exact
+  `FindingCorrelation<T>` and `Complete`/`SubjectAbsent` owned censuses through
+  `FindingCensusCorrelation<T>`, plus `Failed` and `Unevaluated` cells. Exact
+  identity tracks derive from the census correlation rather than bypassing it.
+  `Transitions` compares adjacent evaluated cells; a
   gap-spanning row is qualified and never claims an exact transition version.
 
 ### find
--- a/docs/design/command-transition-model.md
+++ b/docs/design/command-transition-model.md
@@ -242,7 +242,7 @@ Shape reducers do not revise operation arity. In particular:
   transition from version-address rows to package artifact inspection. The
   explicit transition remains `package Package@version`.
 
-The same rule applies to a future timeline. `--count` can reduce an already
+The same rule applies to `timeline`. `--count` can reduce an already
 assembled Timeline table; it cannot probe additional cells. `--print` and
 `--print-all` can print only payloads already carried or explicitly referenced
 by evaluated rows; they cannot turn unevaluated rows into implicit acquisition.
--- a/docs/design/command-transition-model.md
+++ b/docs/design/command-transition-model.md
@@ -437,18 +437,24 @@ is deferred, but the architecture permits it.
 ### Sparse timeline and bisect
 
 A sparse traversal evaluates only caller-selected cells. It preserves the
-Finding correlation semantics:
+Finding census-correlation semantics:
 
-- `Present`: a completed census contains the exact identity;
-- `Missing`: a completed census does not contain it;
+- `Complete`: the focused census completed, including when it contains zero
+  observations;
 - `SubjectAbsent`: the producer has no applicable subject input;
 - `Failed`: inspection did not complete;
 - `Unevaluated`: the address exists in the resolved vector but was not supplied
-  to `FindingCorrelation`.
+  to `FindingCensusCorrelation`.
 
 `Unevaluated` is a presentation state formed by joining the version vector with
 the sparse correlation. It is not fabricated as a Finding or inspection
 outcome.
+
+When a projection selects one exact observation identity,
+`FindingCensusCorrelation.Correlate` produces its `FindingCorrelation` track:
+`Present` means the completed census contains that identity and `Missing` means
+it does not. Whole-census and exact-identity states must not be combined into
+one shadow cell-state model.
 
 Probe order does not define timeline order. Positions come from the resolved
 version vector in the caller's range direction; evaluated inspections retain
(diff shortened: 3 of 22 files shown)


richlander:

Both required reviewers evaluated the same clean rebased head, `feca9bce`, in
separate detached worktrees and reported **CLEAN**.

- **Claude Opus 4.8:** verified zero-payload unaddressed vectors, sparse and
  dense acquisition, duplicate selector handling, all three observation
  censuses, gap qualification, reverse ranges, diff parity, output rules,
  endpoint lifetime, source-generated serialization, and the full CLI suite
  (1806 passed, 10 tool-availability skips).
- **Gemini 3.1 Pro:** independently exercised sparse recommendations, reverse
  ranges, missing versus subject-absent semantics, count behavior, attribute
  occurrence identity, format composition, endpoint disposal, and existing
  diff/timeline tests. It reported no correctness, compatibility, or security
  findings.

The reviews agreed on the acquisition, Finding-state, exact-versus-gap,
resource-lifetime, output-composition, and NativeAOT/SRM contracts. They had no
conflicting findings. Opus recorded two non-blocking observations: mixed
positional/`--package` input reports a generic extra-positional diagnostic, and
`api.type` may report producer-owned facet `Changed` rows in addition to pure
presence. Both behaviors are safe and consistent with existing parser and
Metadata Finding semantics, so no resolution commit applies.

Before these requested reviews, preflight review found and fixed two semantic
issues in `ee5ddebc`: missing type self-presence now renders `Missing` rather
than `SubjectAbsent`, and applied-attribute argument changes now retain
occurrence identity and classify as `Changed`. Both reviewers evaluated those
fixes and signed off cleanly.

richlander:
Ready to merge

richlander:
Correction: the first Gemini reviewer workspace contained a modified test and an untracked probe after reporting clean status, so that Gemini sign-off is invalidated. PR #2699 is not currently ready to merge. A fresh isolated clean-head Gemini re-review is now running; the Opus CLEAN review remains valid.

richlander:
Resolved the valid clean-head Gemini finding in 285f66c6. Timeline transitions now preserve owned-census envelope changes: missing type → existing empty census emits `SubjectAvailable`, and the reverse emits `SubjectUnavailable`; sparse instances retain the unknown-boundary qualification. Added regression coverage and reran the solution build plus the full CLI executable suite (1807 total, 0 failed, 10 tool-availability skips). Both Opus and Gemini are re-reviewing this fixed head in fresh isolated worktrees; the PR remains not ready pending two clean fixed-head reviews.

richlander:
## Fixed-head adversarial re-review

Commit `285f66c6` resolves Gemini's clean-head finding by preserving
`FindingInspection` envelope transitions for owned censuses:

- `Absent → Complete` emits `SubjectAvailable`;
- `Complete → Absent` emits `SubjectUnavailable`;
- sparse instances retain the unknown-boundary qualification;
- native member/attribute atom transitions remain alongside the envelope row;
- `None` is suppressed when subject availability changed.

Both required reviewers re-reviewed this exact fixed head in fresh isolated
worktrees and reported **CLEAN**:

- **Gemini 3.1 Pro:** verified empty and non-empty member/attribute censuses,
  forward/reverse and sparse/dense traversal, gap qualification, output rules,
  endpoint lifetime, and an empty final `git status --short`.
- **Claude Opus 4.8:** verified eight adversarial envelope/atom combinations,
  including the original empty-census failure, reverse removal, failed
  inspection, attribute census, and self-presence separation; it also reran the
  full affected suites and ended with an empty worktree.

The reviewers agree that the finding is resolved with no remaining correctness,
security, or compatibility issues. No guidance was dismissed in this round.
The earlier Gemini sign-off remains explicitly invalidated; these two clean
fixed-head reviews supersede it.

richlander:
Ready to merge

richlander:
Reviewed at head 285f66c6. **Hold — two structural items and one behavioral fix, on top of an excellently executed behavioral contract.**

First, credit where due: the command-transition-model.md timeline contract is genuinely honored. Zero `--at` performs zero acquisition (pinned), `--at all` is the sole dense opt-in, `Unevaluated` is formed purely in the `BuildView` presentation join (`vector.Addresses` vs the evaluated-by-position dictionary) and never touches acquisition, gaps are qualified honestly ("Observed across a gap; the exact transition version is unknown" — the timeline locates a candidate boundary, it does not claim introduction), `RecommendProbe` is a pure midpoint-of-longest-unevaluated-run function that recommends rather than acquires, adjacent-version transitions use native `PairFinding.Added` vocabulary, and endpoint disposal is clean. The `DiffEndpoint` → shared `ApiSurfaceEndpointResolver` extraction is real reuse, and `InspectApiType` returning `Complete([])` for a missing type's self-presence census (vs `Absent` for member/attribute censuses of a missing type) gets the nomenclature exactly right.

But two things breach the spine:

**1. The command bypasses the machinery built for this exact operation.** #2625 landed `FindingCorrelation<T>`/`CorrelatedFinding<T>`/`FindingCorrelationPoint<T>` as the N-version tier, and finding-nomenclature.md's table says "Correlate → `FindingCorrelation<T>`". This command — the first shipped Correlate operation — instead defines a private shadow union `InspectionProjection { Complete(int Count); Absent(string?); Failed(string) }` and never touches the correlation types. The doc now describes machinery the only Correlate command in the product doesn't use. Two acceptable resolutions: (a) route the timeline through the correlation tier, or (b) make a deliberate, documented case in finding-nomenclature.md for a census-correlation tier (count-per-position rather than identity-per-position) and say why the timeline is one. Silence is the one option that isn't acceptable — otherwise the next N-version consumer faces two competing precedents and no rule.

**2. A sixth cell state, legitimized by amending the doc in the same PR.** `"Complete"` was added to architecture.md's cell-state list as part of this PR while finding-nomenclature.md stays silent. That's doc-rewriting to fit the code, not doc-following — exactly the move the adoption checklist exists to catch. If the state earns its place, both docs should say so as a deliberate decision; resolving item 1 will likely settle this too.

**The behavioral fix:** the `catch (Exception ex)` in `ExecuteAsync` aborts the entire dense traversal, discarding already-paid evaluations (probe 40 of 50 throws → all 39 completed evaluations are thrown away). That's the wrong failure topology for Correlate — the envelope already has per-position `Failed`; a mid-traversal fault should degrade that position and keep the rest.

Smaller items, batchable with the above:
- The `GetPairs is null → "Finding comparison failed."` branch is unreachable via `FindingComparison.Compare` and swallows detail if it ever fires — throw with the stated invariant instead.
- `TryResolveTypeName` gives exact matches no precedence over suffix matches, so a type whose full name is also another type's suffix can be ambiguous when it shouldn't be.
- `RecommendProbe` prints an `--at #N` fragment; a full copy-pasteable command line would close the loop.
- Probe-order independence is implemented but not pinned — one test evaluating {A,C} then {B} vs {B} then {A,C} asserting identical timelines would lock it.

Also note: this will want a rebase over #2697, since it extracts `DiffEndpoint` out of `DiffCommand` and #2697 rewrites the transition builders in the same file.

---

richlander:
Addressed the maintainer hold on rebased head `eb990b00` (now based on merged #2697):

- Added shared `FindingCensusCorrelation<T>` in `ILInspector.Findings` for N-address whole-census outcomes. It preserves version-labelled `FindingInspection<T>` values, exposes shared pair comparison, and derives exact `FindingCorrelation<T>` tracks through `Correlate`. The type-presence timeline now uses that exact track; member/attribute timelines use the census tier, so `Complete([])` survives without becoming a fifth exact-identity state.
- Reconciled `finding-nomenclature.md`, command-transition architecture, overview, and architecture docs around census versus exact-identity correlation and the meaning of `Complete`.
- Moved exception containment to the per-address evaluation boundary. A thrown acquisition/extraction fault becomes that cell's `Failed` inspection and later addresses continue; vector resolution, validation, correlation invariants, and rendering remain operation-level failures.
- Removed the generic comparison-failure fallback. Producer comparisons must match the correlated inspections; an unknown or inconsistent outcome now throws with descriptor and address context.
- Exact full type names now take precedence over short/suffix matches.
- Probe recommendations now print a complete reusable `dotnet-inspect timeline --package ... --type ... --finding ... --at '#N'` command.
- Added regression coverage for per-cell thrown exceptions with later success, exact-name precedence, full probe commands, and `{A,C} then {B}` versus `{B} then {A,C}` probe-order independence.

Local validation on this head: solution build; 112 Instructions tests; 432 Metadata tests (1 known skip); 1817 CLI tests (10 tool-availability skips); changed-Markdown lint. Fresh isolated Gemini and Opus reviews are now being requested; the PR remains on hold pending two clean fixed-head reviews and CI.

richlander:
## Fixed-head adversarial review reconciliation

Both required reviewers evaluated the same rebased head, `eb990b00`, in separate detached worktrees and reported **CLEAN**. Both worktrees remained empty under `git status --short`.

- **Claude Opus 4.8:** verified the shared census-correlation validation/ordering, exact type-presence bridge, empty-census subject transitions, per-cell failure continuation, producer-comparison invariants, exact-name precedence, shell-safe probe commands, reverse/probe ordering, #2697 descriptor integration, disposal, and SRM/AOT constraints. It ran targeted Timeline/Diff, Metadata Finding, and Finding pilot suites with no failures.
- **Gemini 3.1 Pro:** independently attacked inspection value equality across producer re-inspection, `Complete([])` versus `Absent`, reverse-range topology, full-name/suffix collisions, `#N` shell quoting, sparse midpoint selection, exception containment, #2697 integration, and endpoint lifetime. It found no correctness, compatibility, security, or architectural-contract issues.

The reviewers agree that the maintainer hold is fully resolved and reported no divergent or dismissed guidance. No additional resolution commit applies after `eb990b00`.

CI is green: .NET 10 build, Linux test suite, markdownlint, and change classification all passed. Local evidence remains: solution build; 112 Instructions tests; 432 Metadata tests (1 known skip); 1817 CLI tests (10 tool-availability skips); changed-Markdown lint.

richlander:
Ready to merge

richlander:
Re-reviewed at head eb990b0 (rebased onto current main, so the #2697 `DiffCommand` overlap is already resolved). **Hold cleared — safe to merge.** Every item landed, and the structural resolution is better than either option I offered.

**1. The correlation-spine breach: resolved by making the census tier first-class.** Instead of routing a whole-census timeline through the exact-identity `FindingCorrelation<T>` (which would have forced a fake identity key per cell) or leaving the private shadow union in place, the PR adds `FindingCensusCorrelation<T>` to the spine itself: the N-address outcome that preserves each `FindingInspection<T>` so `Complete([])`, `SubjectAbsent`, and `Failed` stay distinguishable — with `Correlate(key)` as the explicit bridge down to the exact-identity tier, and `FindingCorrelation<T>` refactored to *compose* it (shared `FindingCorrelationValidation.Order` keeps the duplicate-key/position rejection; `Create`/`Compare` delegate — replacement, not accretion). The command's `InspectionProjection` shadow union is gone; rows are built by pattern-matching the native `FindingInspection<T>` and `FindingCorrelationPoint<T>` cases directly.

**2. The "Complete" cell state: now doc-following, not doc-rewriting.** finding-nomenclature.md carries both operation rows — "Correlate census → `FindingCensusCorrelation<T>`" and "Correlate identity → `FindingCorrelation<T>`" — and states the rule I wanted stated: "`Complete` is therefore a census inspection state, not a fifth exact-identity correlation state," with command-transition-model.md adding "Whole-census and exact-identity states must not be combined into one shadow cell-state model." That's the deliberate, documented case, phrased as a rule the next N-version consumer can follow. The earlier `SubjectAvailable`/`Added` double-vocabulary concern dissolves under the same split: `api.type` renders the identity track (`Present`/`Missing`), member/attribute render the census track (`Complete`/`SubjectAbsent`), and subject-availability transitions are a distinct axis with their own rows rather than a costume for `Added`.

**3. Failure topology: fixed.** `EvaluateCellsAsync` catches per-address, so a throwing evaluation becomes that cell's `Failed` (with `ExType: message` provenance) and later cells still evaluate — pinned by `CellException_BecomesFailureAndLaterCellsStillEvaluate`. Failed-endpoint transitions route through the spine's `correlation.Compare` fold and render visible `Failed` rows.

All the smaller items landed too: the detail-swallowing "Finding comparison failed." branch is replaced by a visible `Failed` row plus throw-with-stated-invariant on unknown outcomes (and there's a new integrity guard I didn't ask for — the producer comparison's inspections are checked reference-identical to the correlated censuses, so a drifting producer fails loudly); `TryResolveTypeName` gives exact full-name matches precedence (pinned); `RecommendProbe` now emits a full shell-quoted copyable command; `--at all` refuses to combine with other selectors; and `ProbeOrder_DoesNotChangeTimelineOrder` pins the order-independence contract. `EmptyOwnedCensus_PreservesSubjectAvailabilityTransitions` pinning that an empty owned census is `Complete(0)` rather than a fabricated absence is exactly the fail-closed distinction the census tier exists to preserve.

The behavioral contract from the first review (zero `--at` = zero acquisition, `--at all` sole dense opt-in, `Unevaluated` as pure presentation join, qualified gap rows) is intact at this head. Good landing — this PR now *is* the exemplar for the Correlate rows in the adoption doc.

---

Issue #2689, opened 2026-07-12
Define explicit Metadata signature degradation and TypeSpec acceptance


richlander:
Follow-up from review of #2677, within the #2490 mechanism matrix.

## Problem

The malformed-signature fallback is currently a plausible value:

- type/field decodes return `"object"`;
- method/property decodes return a zero-parameter `MethodSignature<string>` with a
  default header and `"object"` return type;
- method-spec decodes return an empty argument list.

Callers cannot distinguish rejection from genuine metadata. Consequences are
currently accidental: some sites skip, some render degraded names, and some can
throw after interpreting the fallback as a real signature.

The same top-level TypeSpec also has two acceptance envelopes:

- `GuardedSignatureText.TypeSpecText` applies the iterative structural prescan;
- `TypeResolver.GetTypeNameFromSpecification` additionally applies
  `TypeSpecGuard`'s 1,024-byte per-blob and 4,096-byte cumulative limits.

That makes legal wide-but-shallow input route-dependent.

## Acceptance

- Define an explicit, inspectable rejection result for guarded Metadata signature
  decoding; do not represent rejection as a genuine `object` signature.
- Give consumers one designed fail-closed contract for rejected field, method,
  property, method-spec, and TypeSpec signatures.
- Make the top-level TypeSpec acceptance envelope consistent across
  `GuardedSignatureText` and `TypeResolver`, and document any legal-input bound.
- Add valid wide-but-shallow TypeSpec boundary fixtures above 1,024 bytes and
  malformed-handle coverage.
- Keep provider-specific re-entry state isolated where required, but centralize or
  document shared cap values so they cannot drift accidentally.

This is separate from #2575, which closes the known crash paths in the other
`ISignatureTypeProvider` implementations.

PR #2700, opened 2026-07-12
Guard remaining Metadata signature providers against decode StackOverflow (#2575)

### LINKED ISSUE #2575: Harden TypeNodeProvider (tree signature provider) against nested-TypeSpec StackOverflow (#2490 follow-up)

richlander:
## Summary

Slice B of #2490 (fixes #2575). Closes the last unguarded
`ISignatureTypeProvider` implementations in `ILInspector.Metadata` against SRM
native-stack `StackOverflowException` from over-deep or cyclic signature blobs.
This is the frozen-scope follow-up to the abandoned #2574: the complete provider
census is enumerated up front and frozen by a completeness test, so the scope
cannot ratchet round-by-round.

## Providers closed (all 5)

| Provider | File | Nested guard | Top-level gateway |
| --- | --- | --- | --- |
| `PointerDetector` | PointerDetector.cs | ✓ `TypeSpecGuard` | ✓ `GuardedProviderDecode` |
| `ILSignatureTypeProvider` | CanonicalIL.cs | ✓ | ✓ |
| `TypeNodeProvider` | TypeNodeProvider.cs | ✓ | ✓ |
| `AnchorSignatureTypeProvider` | ApiMemberIdentity.cs | ✓ | ✓ |
| `InaccessibleTypeDetector` | SignatureSpellability.cs | ✓ | ✓ |

## Mechanism (both vectors per provider)

1. **Top-level deep blob** — SRM's `DecodeSignature` recurses on the native
   stack for every nested element *before* the first provider callback, so a
   deep single blob overflows the stack uncatchably. The new
   `GuardedProviderDecode` gateway prescans every top-level provider decode with
   `SignatureBlobGuard.IsSafeToDecode` and fails closed to the provider's
   degraded sentinel (`false` / `"object"` / `"System.Object"` /
   `NamedTypeNode("object")`) when the blob is too deep or malformed.
2. **Nested cross-handle TypeSpec cycle** — each provider's own
   `GetTypeFromSpecification` now bounds re-entry through `TypeSpecGuard`
   (made `public` to mirror the already-public `SignatureBlobGuard`).

This mirrors slice A (#2677): `SignatureDecoder` nested guard + `GuardedSignatureText`
top-level gateway.

## Completeness (anti-ratchet)

`ProviderSignatureDecodeBoundaryTests` freezes the contract:

- **No raw top-level provider decode** may bypass `GuardedProviderDecode`.
- **Every provider** must bound nested re-entry with `TypeSpecGuard`.

A newly added provider or an un-gated decode fails this test rather than
shipping a same-mechanism hole. This is the fix for what killed #2574.

## Crash-containment proofs

Process-isolated child-process regressions (exit-0 = contained):

- `CyclicTypeSpec_ThroughApiSurfaceExtractor_*` — the previously-**skipped**
  #2575 pin, now **un-skipped and passing** (real `ApiSurfaceExtractor.Extract`
  entry, TypeNodeProvider nested cycle).
- `DeepFieldSignature_ThroughApiSurfaceExtractor_*` — TypeNodeProvider top-level.
- `DeepTypeSpec_ThroughCanonicalIl_*` — ILSignatureTypeProvider via `ResolveTypeHandle`.
- `DeepMethodSignature_ThroughPointerDetector_*` — via `ScanPresenceFlags`.
- `DeepMethodSignature_ThroughAnchorProvider_*` — via `CreateMethodAnchorInfo`.
- `DeepMethodSignature_ThroughSpellability_*` — via `CanSpellMethod`.

The shared `TypeSpecGuard`/`SignatureBlobGuard` are additionally crash-proven by
slice A's `TypeResolver`/`MetadataDeclarationQuery` workers.

## Validation

- Build clean (`dotnet build dotnet-inspect.slnx -c Release`).
- Tests: Metadata **440**, product **1797** (10 pre-existing ilasm skips),
  Analysis **421**, targeted Decompiler IL-projection/importer/printer **113** —
  all green, 0 failures.
- **CoreLib A/B byte-identical** (base 81ba0eec vs head) for `type`, `library`,
  `extensions`, `member -S IL` (String, List), and `member --unsafe`. Valid
  signatures pass the prescan, so guarded output equals direct output.

## Non-goals (separate #2490 slices)

TypeRef/TypeDef graph cycles (slice C), amplification (slice D),
custom-attribute `DecodeValue` (slice E), and the #2689 degradation-contract
follow-up.

## Adversarial review

Requesting two cross-model reviews per AGENTS.md (isolated worktrees, reproduce
on clean head) before Ready to merge.

### DIFF
--- a/src/ILInspector.Metadata/ApiMemberIdentity.cs
+++ b/src/ILInspector.Metadata/ApiMemberIdentity.cs
@@ -94,7 +94,18 @@ public static class ApiMemberIdentity
             => TypeResolver.GetTypeNameFromReference(reader, handle).Replace('+', '.');
 
         public string GetTypeFromSpecification(MetadataReader reader, GenericContext? context, TypeSpecificationHandle handle, byte rawTypeKind)
-            => reader.GetTypeSpecification(handle).DecodeSignature(this, context);
+        {
+            if (!TypeSpecGuard.TryEnter(reader, handle, out int blobLength))
+                return "System.Object";
+            try
+            {
+                return reader.GetTypeSpecification(handle).DecodeSignature(this, context);
+            }
+            finally
+            {
+                TypeSpecGuard.Exit(blobLength);
+            }
+        }
 
         public string GetSZArrayType(string elementType) => $"{elementType}[]";
 
@@ -238,11 +249,11 @@ public static class ApiMemberIdentity
         PropertyDefinition property)
     {
         var context = GenericContext.ForType(reader, markerType);
-        var markerSignature = markerMethod.DecodeSignature(AnchorSignatureTypeProvider.Instance, context);
+        var markerSignature = GuardedProviderDecode.Method(reader, markerMethod, AnchorSignatureTypeProvider.Instance, context, "System.Object");
         if (markerSignature.ParameterTypes.Length != 1)
             throw new BadImageFormatException("An extension marker must have exactly one receiver parameter.");
 
-        var propertySignature = property.DecodeSignature(AnchorSignatureTypeProvider.Instance, context);
+        var propertySignature = GuardedProviderDecode.Property(reader, property, AnchorSignatureTypeProvider.Instance, context, "System.Object");
         string propertyName = reader.GetString(property.Name);
         string typeFullName = FormatDefinitionName(reader, extensionClassHandle);
         string extendedType = markerSignature.ParameterTypes[0];
@@ -268,9 +279,12 @@ public static class ApiMemberIdentity
     {
         var type = reader.GetTypeDefinition(typeHandle);
         string methodName = reader.GetString(method.Name);
-        var signature = method.DecodeSignature(
+        var signature = GuardedProviderDecode.Method(
+            reader,
+            method,
             AnchorSignatureTypeProvider.Instance,
-            GenericContext.ForMethod(reader, type, method));
+            GenericContext.ForMethod(reader, type, method),
+            "System.Object");
         string typeFullName = FormatDefinitionName(reader, typeHandle);
         string memberName = MethodMemberName(reader, methodName, method);
         // Route the SRM-direct producer through the single full-name grammar core so it
--- a/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
+++ b/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
@@ -569,7 +569,7 @@ public static class ApiSurfaceExtractor
         byte typeNullableContext)
     {
         var context = GenericContext.ForType(reader, typeDef);
-        var fieldNode = field.DecodeSignature(TypeNodeProvider.Instance, context);
+        var fieldNode = GuardedProviderDecode.Field(reader, field, TypeNodeProvider.Instance, context, (TypeNode)new NamedTypeNode("object", isReferenceType: true));
         var fieldBytes = NullabilityReader.GetNullableBytes(reader, field.GetCustomAttributes());
         int pos = 0;
         fieldNode.ApplyNullability(fieldBytes, ref pos, typeNullableContext);
@@ -744,7 +744,7 @@ public static class ApiSurfaceExtractor
     {
         string name = reader.GetString(method.Name);
         var context = GenericContext.ForMethod(reader, typeDef, method);
-        var treeSignature = method.DecodeSignature(TypeNodeProvider.Instance, context);
+        var treeSignature = GuardedProviderDecode.Method(reader, method, TypeNodeProvider.Instance, context, (TypeNode)new NamedTypeNode("object", isReferenceType: true));
 
         // Determine the effective nullable default: method overrides type
         byte methodContext = NullabilityReader.GetNullableContext(reader, method.GetCustomAttributes());
@@ -1282,7 +1282,7 @@ public static class ApiSurfaceExtractor
     {
         string name = reader.GetString(prop.Name);
         var context = GenericContext.ForType(reader, typeDef);
-        var treeSignature = prop.DecodeSignature(TypeNodeProvider.Instance, context);
+        var treeSignature = GuardedProviderDecode.Property(reader, prop, TypeNodeProvider.Instance, context, (TypeNode)new NamedTypeNode("object", isReferenceType: true));
 
         // Apply nullability to the property type
         var propBytes = NullabilityReader.GetNullableBytes(reader, prop.GetCustomAttributes());
--- a/src/ILInspector.Metadata/AssemblyDetailScanner.cs
+++ b/src/ILInspector.Metadata/AssemblyDetailScanner.cs
@@ -462,7 +462,7 @@ public static class AssemblyDetailScanner
                     {
                         try
                         {
-                            var sig = method.DecodeSignature(PointerDetector.Instance, null);
+                            var sig = GuardedProviderDecode.Method(reader, method, PointerDetector.Instance, (object?)null, false);
                             if (sig.ReturnType || sig.ParameterTypes.Any(p => p))
                                 flags.HasUnsafeCode = true;
                         }
--- a/src/ILInspector.Metadata/CanonicalIL.cs
+++ b/src/ILInspector.Metadata/CanonicalIL.cs
@@ -46,7 +46,18 @@ public sealed class ILSignatureTypeProvider : ISignatureTypeProvider<string, Gen
         => Prefix(rawTypeKind) + CanonicalIL.QualifiedName(reader, handle);
 
     public string GetTypeFromSpecification(MetadataReader reader, GenericContext? context, TypeSpecificationHandle handle, byte rawTypeKind)
-        => reader.GetTypeSpecification(handle).DecodeSignature(this, context);
+    {
+        if (!TypeSpecGuard.TryEnter(reader, handle, out int blobLength))
+            return "object";
+        try
+        {
+            return reader.GetTypeSpecification(handle).DecodeSignature(this, context);
+        }
+        finally
+        {
+            TypeSpecGuard.Exit(blobLength);
+        }
+    }
 
     public string GetSZArrayType(string elementType) => $"{elementType}[]";
     public string GetArrayType(string elementType, ArrayShape shape) => $"{elementType}[{new string(',', Math.Max(shape.Rank - 1, 0))}]";
@@ -110,8 +121,8 @@ public static class CanonicalIL
     {
         HandleKind.TypeDefinition => QualifiedName(reader, (TypeDefinitionHandle)handle),
         HandleKind.TypeReference => QualifiedName(reader, (TypeReferenceHandle)handle),
-        HandleKind.TypeSpecification => reader.GetTypeSpecification((TypeSpecificationHandle)handle)
-            .DecodeSignature(ILSignatureTypeProvider.Instance, null),
+        HandleKind.TypeSpecification => GuardedProviderDecode.TypeSpec(
+            reader, (TypeSpecificationHandle)handle, ILSignatureTypeProvider.Instance, (GenericContext?)null, "object"),
         _ => $"0x{MetadataTokens.GetToken(handle):X8}"
     };
 
@@ -269,7 +280,7 @@ public static class CanonicalIL
         var method = reader.GetMethodDefinition(handle);
         string name = reader.GetString(method.Name);
         string parent = QualifiedName(reader, method.GetDeclaringType());
-        var sig = method.DecodeSignature(ILSignatureTypeProvider.Instance, genericContext: null);
+        var sig = GuardedProviderDecode.Method(reader, method, ILSignatureTypeProvider.Instance, (GenericContext?)null, "object");
         return FormatCall(sig, parent, name, genericArgs: null);
     }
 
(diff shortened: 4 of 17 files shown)


richlander:
Reviewed at head c55aee7, including the commit-by-commit evolution. **One pre-merge ask; otherwise this is the right slice B and I endorse it.**

**What's right.** The product changes are mechanically consistent with slice A — no new guard type invented, just `SignatureBlobGuard` prescan gateways (`GuardedProviderDecode`, one ternary shape everywhere) plus the identical `TypeSpecGuard.TryEnter`/`Exit` try/finally stamp for nested TypeSpec re-entry in all five providers, same constants, no per-site overrides. The containment tests follow doctrine: process-isolated child workers per provider entry path, against synthesized adversarial PEs, and the long-standing `CyclicTypeSpec_ThroughApiSurfaceExtractor` skip is finally un-skipped — that closes the #2575 item that's been on my priority list since the skip landed.

**The census evolution is the best part of this PR.** v1's inverted allow-list (only flagged a raw decode if the line carried a literal provider token — a local-variable alias evaded it entirely, and it never scanned MetadataPrimitives, so the AttributeDecoder hole was structurally invisible to it) → v2's deny-list regex with whole-file gateway exemptions → v3's Roslyn AST scan where gateway files get **no** exemption and every decode must individually sit in an `IsSafeToDecode` ternary or be a `GetTypeSpecification(...)` re-entry. Comment-spoofing, newline-splitting, and aliasing are all dead as evasions. The two-model adversarial review process (#2693) demonstrably earned its keep here.

**The pre-merge ask: a runtime pin for the AttributeDecoder fix.** Commit 2's headline — `GetUnderlyingEnumType` prescanning the enum field signature and degrading to `Int32` — is the one fixed site with *zero* runtime coverage. The census pins the guard's **shape**, not its **strength**: `IsSafeToDecode(..., maxDepth: 1_000_000)` or a wrong `Kind` argument fully satisfies the AST classifier while restoring the crash. Every other guarded path has a process-isolated pin that would catch a strength regression; this one has nothing. The machinery already exists (`RunWorker` + a synthesized deep enum-field blob) — one worker asserting containment and the `Int32` degrade closes it. This matters doubly because the `Int32` fallback is the most dangerous sentinel in the slice: a real I8-underlying enum degraded to `Int32` shifts subsequent attribute-blob byte parsing, producing *wrong* attribute values rather than visibly degraded ones.

Non-blocking, for the tracker (#2575/#2490):

1. **Scope honesty.** The census proves closure for ILInspector.Metadata + MetadataPrimitives only. There are ~20 raw `Decode*Signature` sites with the identical uncatchable mechanism in ILInspector.Instructions (`IlBodyDiff`, `StackTypeSignatureProvider` — including an unguarded `GetTypeSpecification(handle).DecodeSignature(this, ...)` re-entry), ILInspector.Analysis (`TypeRefDecoder`, `MemberResolver`), and the Decompiler's `IrImporter`. Presumably slice C — but nothing in-repo ratchets that. Please enumerate the residual sites on the tracking issue so the docket entry is checkable.
2. **Fact 2 is file-granular**: one `TypeSpecGuard.TryEnter` anywhere in a file sanctions every nested re-entry in that file — the same whole-file weakness the AST rewrite fixed for prescans. A per-invocation check (enclosing block is the `try` of a `TryEnter`) would close it.
3. **`TypeSpecGuard` went `internal`→`public`** with raw thread-static accounting and an unpaired `Exit(int)` — an external caller's pairing mistake silently skews `s_depth`/`s_cumulativeBytes` thread-wide. A disposable token (or `InternalsVisibleTo` instead of `public`) is the safer surface.
4. **Degradation is silent by convention** (inherited from slice A): `HasUnsafeCode` can be hidden by an over-deep pointer-bearing signature, spellability fails open to "spellable", and `FallbackSignature`'s `default` header reads as static. Matching slice A's convention is the right call *for this slice*, but a "degraded-decode is visible" pass (diagnostic or marker) deserves its own line on #2490 rather than staying unstated.
5. The manual verification claims (census fails on all three reviewer evasions; CoreLib A/B byte-identical) are prose-only. Encoding the evasion snippets as negative self-tests of the census would make its strength durable through refactors.

---

richlander:
## Adversarial review reconciliation — Ready to merge

Two isolated cross-model adversarial reviews (I'm Opus, so reviewers are **GPT-5.5** and **Gemini 3.1 Pro**), each in its own worktree, both **CLEAN on the exact head `223b7c45`**. CI (build + test) is green on that head.

### Product code: CLEAN (both reviewers, every round)

Both reviewers repeatedly audited both assemblies and confirmed **no unguarded top-level signature decode remains**, and that valid output is unchanged — CoreLib CLI hashes byte-identical to the pre-PR base for `type`, `library`, `extensions`, `member -S IL`, `member --unsafe`.

### Findings — all verified and resolved

|---|----------|---------|------------|
| 1 | Gemini | `AttributeDecoder.GetUnderlyingEnumType` performed an **unguarded** top-level `field.DecodeSignature` (deep enum-field blob → uncatchable native StackOverflow). Reviewer reproduced the crash. | Prescan with `SignatureBlobGuard.IsSafeToDecode`, fail closed to `Int32` — `f1c8594a`. Gemini reproduced it is now **contained**. |
| 2 | Gemini + GPT | The frozen census was a **fragile string/regex scan** — whole-file gateway exemption (GPT); newline-split + comment-spoof (Gemini). | Rewrote as a **Roslyn AST census** — `c55aee79`. |
| 3 | GPT | Prescan ternary accepted **any** `IsSafeToDecode` in its condition, not one bound to the decoded blob (nil-handle / unrelated-blob laundering). | Bind the guard to the decode's **same receiver `.Signature`** — `6820356d`. |
| 4 | Gemini | Classifier matched only `MemberAccessExpressionSyntax`, missing **null-conditional `?.`** (`MemberBinding`) and **method-group / delegate** forms. | Enumerate **every decode-name occurrence** and require each to be the invoked member of a sanctioned call — `60ddc831`. |
| 5 | (proactive) | Denylist covered only the three currently-used decoders. | Added **`DecodeLocalSignature` / `DecodeMethodSpecificationSignature`** — `223b7c45`. |

### Convergence

On `223b7c45` the census provably **flags** every evasion the reviewers constructed: gateway-file, local-alias, newline-split, comment-spoof, nil-handle guard, unrelated-blob guard, null-conditional, method-group/delegate, lambda, and `DecodeLocalSignature`. It **passes** on the real tree (all 21 decode sites classify).

**Residual (both reviewers, explicitly non-blocking):** purely theoretical *test-gaming* shapes — reflection-by-string (`GetMethod("DecodeSignature")` is a string literal, not a `SimpleNameSyntax`), `dynamic` dispatch (separately blocked by AOT analyzers `IL2026`/`IL3050`), and deliberately-broken guard control-flow (`!guard`, `|| true`). None is an accidental-regression shape; both reviewers state they are out of the native-stack-SO threat model.

Metadata suite 440 green; product/analysis/targeted-decompiler suites green; CoreLib A/B byte-identical.

**Status: Ready to merge.** (Not merging without maintainer approval.)

richlander:
## Addressed @richlander's blocking pre-merge ask: runtime containment pin

New head: `be0e21a1` (one test-only commit on top of the two-CLEAN head `223b7c45`).

The census pinned the AttributeDecoder guard's *shape*, not its *strength* —
`GetUnderlyingEnumType` had zero runtime coverage, and its `Int32` fallback is
the most dangerous sentinel (a real I8-underlying enum degraded to `Int32`
mis-sizes the value and shifts subsequent attribute-blob byte parsing).

Added a process-isolated worker (reusing `RunWorker`) that builds a synthetic
custom-attribute PE — enum `N.E` with an over-deep first instance-field
signature, attribute `N.A` with `.ctor(N.E)`, applied to `N.A` — so decoding
that attribute drives `ArgTypeProvider.GetUnderlyingEnumType` through the deep
enum-field blob, the exact guarded path. The `[Fact]` asserts child exit 0
(containment) **and** a non-null decode with the single fixed argument (proving
the graceful `Int32` degrade executed rather than a throw).

**Reproduction check:** temporarily reverting the guard to an unguarded
`field.DecodeSignature(...)` makes the containment `[Fact]` fail — the child
StackOverflows (nonzero exit) — confirming the worker genuinely reproduces the
crash and the guard is what contains it. Restored; Metadata suite **442 green**
(+2). Product code is unchanged, so the CoreLib A/B remains byte-identical.

The five non-blocking tracker items are recorded on #2575
(https://github.com/richlander/dotnet-inspect/issues/2575#issuecomment-4953939440),
including the concrete slice-C scope (residual unguarded `Decode*Signature`
sites in `ILInspector.Instructions`/`ILInspector.Analysis` + `IrImporter`).

Re-review of the new head dispatched to both cross-model reviewers on isolated
worktrees; will post the reconciliation once they report.

richlander:
## Re-review reconciliation on head `be0e21a1` (runtime pin)

Both cross-model reviewers re-reviewed the single test-only commit that adds the
runtime containment pin. **Both CLEAN**, each with an independent crash-on-revert
reproduction.

**GPT-5.5 — CLEAN.** Reproduced the unguarded path in an isolated script: output
reached `before DecodeValue` and `GetUnderlyingEnumType N.E`, then the process
died with `Stack overflow`, exit `134`; with the guard in place the worker exits
0 and decodes one fixed argument. Confirmed `100_000` SZARRAY depth is
sufficient here and the TypeDef row-2 coded index is single-byte (asserted).

**Gemini 3.1 Pro — CLEAN.** Independently reverted the guard → child crashed with
uncatchable native StackOverflow, parent detected nonzero exit and failed the
`[Fact]`. Audited the MetadataBuilder shape (field/method ownership, `0x11`
VALUETYPE + `CodedIndex.TypeDefOrRefOrSpec`, blob prolog + 4-byte value + zero
named args) — no gaps. Noted the `Assert.NotNull` + `Assert.Single` pair proves
the guard fails closed to `Int32` and *completes* the decode, rather than a
managed throw that `TryDecode` would swallow to null.

**Agreement / divergence:** No divergence — both verified the same route, the same
crash-on-revert, and the same graceful-degrade semantics. No findings; nothing
dismissed.

This is the fourth CLEAN pair across the slice (`223b7c45` product+census, now
`be0e21a1` runtime pin). CI: `build-net10` green, `test` running — will confirm.

**Ready to merge** once `test` is green. Not self-merging — awaiting your
go-ahead, @richlander.

richlander:
CI green on `be0e21a1`: `test` and `build-net10` pass; smoke/pack/markdownlint correctly skipped (no product/config/doc changes). Two CLEAN cross-model reviews on this exact head. **Ready to merge** — awaiting your go-ahead, @richlander (not self-merging).

richlander:
Re-reviewed at head be0e21a. **The pre-merge ask landed, and the census got stronger than I asked for — clear to merge.**

**The runtime pin is exactly right.** `DeepEnumField_ThroughAttributeDecoder_IsContainedInChildProcess` drives the real path — `AttributeDecoder.TryDecode` → `CustomAttribute.DecodeValue` → `ArgTypeProvider.GetUnderlyingEnumType` — against a synthesized PE whose enum carries a 100,000-deep field signature blob, in a process-isolated worker per doctrine. Crucially it pins both halves: containment (a strength regression crashes the child, nonzero exit) *and* graceful degrade (`decoded` non-null with the single fixed argument read as Int32). This is the strength pin the AST census structurally cannot provide, and its absence was my one blocker.

The three census-hardening commits close strength gaps I'd flagged as non-blocking, which is better than I asked:

- **Blob binding (6820356)**: `ConditionGuardsReceiverSignature` now requires the `IsSafeToDecode` prescan to take the *decoded receiver's own* `.Signature` as an argument — an unrelated or nil-handle guard can no longer launder an unguarded decode past the census.
- **Occurrence-level flagging (60ddc83)**: the census now enumerates every occurrence of a decode method *name* and requires each to be the invoked member of a sanctioned call — method-group/delegate references and null-conditional forms are violations rather than invisible, closing the deferred-invocation escape class.
- **Denylist completion (223b7c4)**: `DecodeLocalSignature` and `DecodeMethodSpecificationSignature` added with the right rationale comment — freezing entry points the product doesn't currently call is precisely what an anti-ratchet gate is for, and the transitive-coverage note for the sub-decoders answers the "why not DecodeType" question before it's asked.

The remaining follow-ups from my first review (residual raw-decode sites in Instructions/Analysis/Decompiler enumerated on the tracker, file-granular Fact 2, `TypeSpecGuard`'s public unpaired `Exit`, census negative self-tests) stay non-blocking. Nice slice.

---

PR #2701, opened 2026-07-12
Define value equality for Finding collection records

### LINKED ISSUE #2669: Define value equality for Finding collection records
## Conclusion

Define the equality contract for Finding records that carry immutable
collections before those envelopes are used as cache keys, deduplication
identities, or change detectors.

Today several public record shapes contain `ImmutableArray<T>` values, including
`FindingInspection<T>.Complete`, `FindingMatch`, and
`FindingComparison<T>.Complete` through its pairs and match. Record-generated
equality composes the collection type's equality, which does not imply
element-wise semantic equality for independently materialized arrays. Payload
records may have the same problem with `ImmutableArray<T>` or
`ImmutableHashSet<T>`.

This is not a matcher-identity problem. `FindingKey` remains the authority for
cross-version correspondence; this issue is about the .NET equality and hash
contract exposed by the value-shaped API.

## Work

- Inventory public Finding envelopes and producer payload records that contain
  immutable collections.
- Decide which shapes promise semantic value equality and which are operation
  results whose identity should not be inferred from record syntax.
- For value-equal shapes, define collection-aware equality and matching hash
  codes.
- For non-value-equal shapes, make that contract explicit and avoid accidental
  record equality suggesting otherwise.
- Document whether and where payload equality participates; do not make generic
  Finding matching depend on arbitrary `T.Equals`.

## Acceptance criteria

- Two independently materialized but semantically identical values have the
  documented equality result.
- Equal values produce equal hash codes.
- Reordered ordered streams and reordered identity sets follow their respective
  documented semantics rather than one universal collection rule.
- Tests cover empty/default rejection, duplicate elements, ordering, and
  independently allocated collections.
- Finding correspondence behavior remains key-driven and unchanged.

Follow-up from #2614.

richlander:
## Conclusion

Finding collection records now expose semantic .NET value equality instead of
immutable-collection instance equality, without changing `FindingKey`-driven
correspondence.

Ordered streams compare by sequence and multiplicity. Equivalence allow-lists
compare as sets. Analysis payload leaves that carry immutable arrays now own
the same explicit sequence contract.

## What changed

- Added sequence-aware equality and hashes for completed inspections, matches,
  completed comparisons, and correlated durable findings.
- Added set-aware equality and hashes for `FindingEquivalence`; allow-lists
  normalize custom collection comparers to enum identity semantics.
- Preserved positional-record `init`/`with` behavior while validating default,
  null, and null-element collection states.
- Added ordered equality for `MethodIdentity` and `MemberRef`, which makes
  `DirectCall`, `UnsafetyOccurrence`, and `UnsafeEvidence` compose correctly.
- Documented the payload boundary: `Finding<T>` uses producer-owned payload
  equality, while matching remains exclusively key-driven.

## Behavior

```csharp
var first = new FindingMatch(
    [new(FindingEdgeKind.Matched, 0, 0, 100)],
    []);
var independent = new FindingMatch(
    [new(FindingEdgeKind.Matched, 0, 0, 100)],
    []);

Debug.Assert(first == independent);
Debug.Assert(first.GetHashCode() == independent.GetHashCode());
var changed = independent with
{
    Edges = [new(FindingEdgeKind.Matched, 1, 0, 100)],
};
Debug.Assert(first != changed);
```

Set construction order does not affect `FindingEquivalence`; ordered array
reordering and duplicate-count changes do affect inspection, match, comparison,
and payload values.

## Compatibility

- Public positional record properties retain their `init` setters.
- Required immutable arrays reject `default`; optional generic-parameter names
  normalize omission to an initialized empty array.
- No matcher or fold path calls payload `Equals` or `GetHashCode`.
- Open Analysis/Research PR files are untouched.

## Evidence

| Check | Result |
| --- | --- |
| `ILInspector.Instructions.Tests` | 111 passed |
| `ILInspector.Analysis.Tests` | 426 passed |
| `ILInspector.Research.Tests` | 96 passed |
| Focused Decompiler member-identity tests | 27 passed |
| Changed Markdown | markdownlint clean |

## Review

| Reviewer | Exact head | Result |
| --- | --- | --- |
| Claude Opus 4.8 | `f98497c0` | CLEAN |
| Gemini 3.1 Pro | `f98497c0` | CLEAN |

Gemini's first pass found that set equality accidentally selected LINQ's
quadratic comparer overload. Commit `f98497c0` switched to native
`ImmutableHashSet.SetEquals`; both reviewers independently verified the fix and
the complete final diff with real tests and throwaway probes.

Closes #2669.

### DIFF
--- a/docs/design/finding-value-equality.md
+++ b/docs/design/finding-value-equality.md
@@ -1,0 +1,73 @@
+# Finding Value Equality
+
+Finding correspondence and .NET value equality are separate contracts.
+`FindingKey` controls cross-version correspondence. Equality answers whether
+two already-materialized information or outcome values represent the same
+content.
+
+## Core collection contracts
+
+Public Finding records use the semantics of the collection they expose:
+
+| Shape | Collection meaning | Equality |
+| --- | --- | --- |
+| `FindingInspection<T>.Complete.Findings` | Ordered census | Sequence equality |
+| `FindingMatch.Edges` | Ordered committed alignment | Sequence equality |
+| `FindingMatch.MoveCandidates` | Ordered candidate evidence | Sequence equality |
+| `FindingComparison<T>.Complete.Pairs` | Ordered transition stream | Sequence equality |
+| `CorrelatedFinding<T>.Occurrences` | Version-position order | Sequence equality |
+| `FindingEquivalence` allow-lists | Identity sets | Set equality |
+
+Sequence equality is order- and multiplicity-sensitive. Independently allocated
+arrays with equal elements are equal and produce equal hash codes. Reordering
+elements, adding a duplicate, or removing a duplicate changes the value.
+Default `ImmutableArray<T>` values are rejected at construction; empty
+initialized arrays remain valid values.
+
+Set equality is independent of enumeration and construction order. Duplicate
+inputs are normalized by `ImmutableHashSet<T>` and do not change the value.
+Null sets are rejected at construction. Allow-lists normalize custom collection
+comparers to the enum's default identity comparer so equal policies cannot
+behave differently.
+
+The union wrappers compose the active case's equality. Consequently,
+independently materialized inspections and comparisons are equal when their
+cases and complete ordered content are equal. `FindingCorrelation<T>` remains a
+reference-identity operation object; its durable `CorrelatedFinding<T>` value
+has collection-aware equality.
+
+## Payload boundary
+
+`Finding<T>` and `PairFinding<T>` compose `EqualityComparer<T>.Default`.
+Payload types therefore own their .NET equality contract. A producer record
+that contains `ImmutableArray<T>` or `ImmutableHashSet<T>` must define
+collection-aware equality if it promises semantic value equality; the generic
+Finding layer does not reinterpret opaque payloads.
+
+Payload equality never controls correspondence. `FindingMatcher` consumes
+`FindingKey` streams and does not call `T.Equals` or `T.GetHashCode`. Two
+findings may therefore correspond by key even when their producer-owned
+payload equality reports them as different.
+
+## Producer inventory
+
+The current Finding payloads fall into three groups:
+
+- scalar-only value records such as `CanonicalIlOperation`,
+  `CSharpCanonicalLine`, and `DecompilerFidelityCause` already have correct
+  generated equality;
+- `MethodIdentity` and `MemberRef` carry ordered type/name arrays and define
+  sequence equality and matching hashes. `DirectCall`, `UnsafetyOccurrence`,
+  and `UnsafeEvidence` compose those leaf contracts;
+- Metadata's union payload comparison supplies its own set-oriented comparer
+  instead of relying on the payload record's generated equality.
+
+Generic producer promotion may use `EqualityComparer<T>.Default` when no
+producer comparer is supplied. A collection-bearing payload must therefore
+either define semantic equality itself or supply the producer operation with
+an explicit comparer.
+
+Collection-bearing Analysis summaries and Research diff/composition results
+are operation or presentation outputs rather than Finding payload values. They
+are not cache or correspondence identities and retain their existing
+operation-result semantics.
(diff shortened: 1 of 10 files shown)


richlander:

Both isolated reviewers completed clean reviews against exact head `f98497c0`.

| Reviewer | Final result | Evidence |
| --- | --- | --- |
| Claude Opus 4.8 | **CLEAN** | Verified native set complexity, all equality/hash members, positional-record clone/`init`/`Deconstruct` behavior, default/null guards, union composition, key-only matching, 111 Instructions tests, 426 Analysis tests, and a 33-assertion independent probe. |
| Gemini 3.1 Pro | **CLEAN** | Verified native set complexity, commutative set hashing, comparer normalization, positional-record backing fields, default/empty behavior, all payload members, union composition, key-only matching, 111 Instructions tests, and 426 Analysis tests. |

Gemini's first pass on `8f20b1e8` found that
`right.Contains(value, comparer)` bound to LINQ and made set equality
quadratic. I verified the finding and fixed it in
[`f98497c0`](https://github.com/richlander/dotnet-inspect/commit/f98497c0)
by using native `ImmutableHashSet.SetEquals`. Both reviewers then independently
re-reviewed that exact fixed head and reported no outstanding findings.

Opus's first pass on the old head was otherwise clean. The final reviews agree
on the equality/hash contract, compatibility of the preserved positional
record surface, validation behavior, payload ownership, and unchanged
`FindingKey` correspondence. No findings were dismissed and no further
resolution commit applies.

All required CI checks are green. **Ready to merge.**

richlander:
Reviewed at head f98497c0. **Endorse — safe to merge.**

This fixes a real, latent defect: C#'s generated record equality over `ImmutableArray<T>` members compares backing-array *references*, so independently materialized records with identical content were unequal and hashed differently. Every spine record carrying collections had this — `FindingInspection<T>.Complete`, `FindingMatch`, `FindingComparison<T>.Complete`, `CorrelatedFinding<T>`, `MethodIdentity`, `MemberRef`. The PR gives them genuine sequence/set value equality with matching hash codes, and the design doc makes the semantics deliberate rather than incidental.

What makes this well-constructed:

- **The doctrine boundary is stated and pinned.** finding-value-equality.md separates the two contracts precisely: `FindingKey` controls cross-version correspondence; .NET equality answers whether two materialized values have the same content — and payload equality is producer-owned and *never* drives matching. `FindingPayloadEquality_IsProducerOwnedButMatchingRemainsKeyDriven` is the load-bearing test: two findings with equal keys but reference-unequal collection payloads still match key-driven while `Assert.NotEqual` holds on the findings themselves. That test is what prevents .NET equality from becoming a shadow identity channel.
- **Validation now survives `with` expressions.** The backing-field + init-accessor pattern means `first with { Findings = default }` throws where the old positional records silently accepted a default array (`FindingMatch` previously had *no* construction validation at all — `new FindingMatch(default, [])` was representable). Pinned for every record.
- **The comparer-normalization hole is closed adversarially.** `FindingEquivalence` rebuilding its sets with the default comparer — pinned by the `AllPairKindsEqualComparer` test — means two policies that print identically can no longer behave differently. Nice catch; that's exactly the "equal policies cannot behave differently" invariant the doc states.
- **`FindingCorrelation<T>` stays a reference-identity operation object while its durable `CorrelatedFinding<T>` value gets collection-aware equality** — the right split, and the doc says so explicitly.
- The set hash (count + xor + sum of element hashes) is order-independent and sound; sequence hashes prefix the length; empty vs default is handled consistently with construction rules.

Two nits, neither blocking:

1. `ImmutableArrayValueEquality` (ILInspector.Analysis) and `FindingValueEquality` (ILInspector.Findings) are near-duplicate helpers. Their default-handling deliberately differs (Analysis rejects defaults at construction; Findings tolerates them defensively), and they live in different assemblies, so two instances is defensible — but a third copy should trigger extraction into a shared internal source or a public helper.
2. `GetHashCode` is O(n) per call and uncached; if a large `Complete` census ever becomes a hot dictionary key, a cached hash is the follow-up. Not needed now.

---

PR #2703, opened 2026-07-13
Remove oneline compatibility alias

richlander:
## Summary
- remove the hidden `--oneline` compatibility alias, exact `--terse` runtime/help references, and `DOTNET_INSPECT_FORMAT=oneline`/`one-line` aliases
- document the alias retirement in current release notes while restoring the historical shipped note that introduced the hidden compatibility alias
- rename product row-output plumbing from OneLine to Tabular/Table terminology, including timeline command wiring after merging current `origin/main`
- update type-command help, docs, smoke scripts, parser tests, and the source-built PR corpus baseline to use compact table / `--table` terminology
- leave `real-world-baseline.json` pinned to its published `dotnet-inspect.any` 0.14.0 self-corpus identities so Deep Inspect baseline comparison stays comparable
- treat env-selected table/tsv/jsonl as explicit row output for normal rendering, but ignore env row formats under `--bare` unless a row-format flag is explicitly passed
- preserve explicit non-row CLI formats like `--json` over env row formats even when `--bare` is present
- pin recognized non-tabular env formats (`markdown`, `json`, `plaintext`, `mermaid`) as explicit format selections
- keep generic English phrases like “one-line descriptions” where they are not the removed `--oneline` flag/token

## Validation
- git grep -n -- `--oneline` -- . `:(exclude)src/dotnet-inspect/release-notes.md` `:(exclude)tools/DecompilerHarness/corpus/real-world-baseline.json` => no matches
- git grep -n -- `--terse` -- . => no matches
- git grep -n -i -- `oneline` -- . `:(exclude)src/dotnet-inspect/release-notes.md` `:(exclude)tools/DecompilerHarness/corpus/real-world-baseline.json` => no matches
- git grep -n `terse output\|terse, no docs` -- src docs README.md scripts => no matches
- dotnet run --project src/dotnet-inspect -c Release -- type --help | grep "compact table"
- git diff --exit-code origin/main -- tools/DecompilerHarness/corpus/real-world-baseline.json
- dotnet build src/dotnet-inspect/dotnet-inspect.csproj -c Release --configfile nuget.config --verbosity quiet
- dotnet build src/dotnet-inspect/dotnet-inspect.csproj -c Release --configfile nuget.config --verbosity quiet -p:DefaultTargetFramework=net10.0
- dotnet run --project src/dotnet-inspect.Tests -c Release -- -class DotnetInspector.Tests.Parsers.MemberOptionsParserTests
- dotnet run --project src/dotnet-inspect.Tests -c Release -- -class DotnetInspector.Tests.Parsers.TypeOptionsParserTests
- dotnet run --project src/dotnet-inspect.Tests -c Release -- -class DotnetInspector.Tests.TimelineCommandTests
- dotnet run --project src/dotnet-inspect.Tests -c Release -- -method "*Package_Content_Bare_IgnoresEnvironmentRowFormat*"
- dotnet run --project src/dotnet-inspect.Tests -c Release
- npx markdownlint-cli src/dotnet-inspect/release-notes.md
- npx markdownlint-cli docs/workflows/features.md
- npx markdownlint-cli docs/design/usability-improvements.md
- npx markdownlint-cli backlog.md docs/decompiler-quality.md docs/design/output-format-analysis.md docs/design/output-shapes.md docs/design/section-model.md docs/design/section-pipeline.md docs/design/service-model-refactoring.md docs/workflows/core/member-lookup-docs-code.md docs/workflows/core/type-queries.md docs/workflows/discovery/api-diff.md docs/workflows/discovery/type-shape-and-implements.md
- dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.AnnotationStructuredViewTests


richlander:
Adversarial review reconciliation for current head `bbd3c048`:

- **MAI-Code-1-Flash r9:** CLEAN. Verified clean worktree, `--oneline` rejected, `DOTNET_INSPECT_FORMAT=oneline` ignored, and source/docs/scripts no longer expose relevant `--oneline`, `--terse`, `OneLine`, or `ResolveOneLine` compatibility residue.
- **Claude Opus 4.8 r9:** CLEAN. Verified clean worktree, `--oneline` rejected, env `oneline` ignored, type help no longer says terse output, env table/tsv/jsonl explicitness works, `--bare` suppresses env row formats while preserving explicit non-row formats like `--json`, source-built PR baseline identities exist, and `real-world-baseline.json` remains pinned/unchanged.

Earlier review findings were addressed in follow-up commits: env tabular explicitness and source-built baseline names (`8368e694`), pinned real-world baseline restore (`af5e31d6`), exact `--terse` removal (`c25a02b2` / `0dd1feca`), shared `--bare` env-row suppression (`2cdab3da`), merge-result timeline tabular wiring (`04237e76`), generic release-note wording restore (`e358d72a`), and explicit non-row format precedence under `--bare` (`bbd3c048`).

CI is green and merge state is clean. Ready to merge.

richlander:
Reviewed at head bbd3c04, all ten commits. The removal itself is done right — zero `OneLine` identifiers survive anywhere in `src/`, the rename is semantics-preserving (`ResolveTabular` has the identical Table/Tsv/Jsonl body), the just-merged timeline command was correctly re-covered after the main merge, and the real-world corpus baseline ends the PR **byte-identical to merge base** after the mid-PR sed corruption was caught and restored (`af5e31d` — good recovery; the file pins published-binary identities that no source edit can change). The `--bare` × env-format interaction that review feedback shook loose is properly landed: suppression generalized into the shared resolver for all commands, with `bbd3c04` preserving explicit `--json`/`--markdown` under bare suppression, pinned end-to-end plus three parser-level tests. **One fix round before merge:**

**1. The two removals are unpinned, and one of them fails open.**
- `--oneline` now dies with a generic unknown-option parse error — acceptable for a hidden alias, but there's no negative pin. One test (`--oneline` → parse error) locks the removal against accidental resurrection.
- `DOTNET_INSPECT_FORMAT=oneline` (and `one-line`) previously selected Table and now silently falls through `_ => null` to the command default — different output, no diagnostic, no pin. That's a fail-open behavior change of exactly the kind this repo's doctrine exists to catch. Either keep the env alias mapping to Table (it costs one switch arm and breaks no one), or make the drop visible (warn on unrecognized/retired values) — but decide it explicitly and pin it.

**2. `IsFormatExplicitlySet` gained broader semantics than the review asked for.** The feedback was about env *tabular* modes counting as explicit; the implementation makes **every** recognized env value (markdown, json, plaintext, mermaid too) count as explicit format, which quiets tips and suppresses the example-command epilogues. Only the env=table consequences are pinned; env=markdown/json behavior changes ride along untested. Either narrow the check to the tabular values or pin the broader semantics as deliberate.

**3. Release notes are edited backwards but not forwards.** The PR deletes the *shipped* release's line documenting the compatibility alias (rewriting history) while adding no entry for the removal itself — a breaking CLI change. Restore the historical line and add the removal to the current notes instead.

Small residue, take or leave: a few sed scars where prose was renamed by pattern rather than meaning — `SkillCommand.cs:22` "The table description is sourced from the skill's YAML frontmatter" (was "one-line description" — a skill description isn't a table), `Tsv_HasAHeaderAndTabularPerFact` (original meant "one line **per fact**", which was the assertion's semantic content, not the alias), and two "Table table" comments in `OfflineVerbosityTests`. Also noting: env=table + multi-section `-S` now exits 1 where main auto-promoted to markdown — intentional per the review feedback and fine, but only pinned at the parser-property level; an end-to-end pin would be cheap. And the env-override cache removal means the variable is live-read — env-mutating tests are serialized in `[Collection("Console")]`, but any format resolution *outside* that collection during parallel runs could transiently observe a test's env value; low but nonzero flake risk worth keeping in mind if CI ever flakes there.

---

richlander:
Amending my review — the ask list was partly too high, and the right test exposed it: **what distinguishes removed `--oneline` from never-existent `--twoline`?** Once the option is deleted, nothing — both are unknown tokens handled by the parser's generic rejection. Applying that test:

**Withdrawn: the negative pin for `--oneline` rejection.** A test asserting the parser rejects `--oneline` would be testing (a) that unknown options error — generic parser behavior, not this PR's surface — and (b) that nobody has defined an option by that name — testing the absence of code. Neither is what the pins doctrine exists for, and the doctrine I cited doesn't apply anyway: `--oneline` fails *loud* (hard error, nonzero exit), and "no silent behavior changes without pins" targets silent ones. The removal needs no rejection test, same as `--twoline` never did.

**Narrowed: the env-var ask.** The `--twoline` symmetry holds at steady state here too — `DOTNET_INSPECT_FORMAT=twoline` was always silently ignored, and `oneline` now joins it. What's different is only the *transition*: environments that already have `DOTNET_INSPECT_FORMAT=oneline` set (CI configs, dotfiles) silently change output, with no loud failure to prompt migration — unlike the flag, which announces itself. But a pin doesn't fix a transition; a decision does. Any of these settles it, in increasing order of effort: (a) a release-note line documenting the retirement and its env-var effect — no code, and probably right for a hidden alias's env twin; (b) keep the env alias one more release (one switch arm); (c) warn on the retired value. Only (b) or (c) create product surface that would then deserve a pin; if you pick (a), there's nothing to pin and nothing more to ask.

**Unchanged, because they're live behavior rather than absence:**
1. `IsFormatExplicitlySet` returning true for *every* recognized env value (markdown/json/plaintext/mermaid) is new semantics beyond the env-tabular case review asked for — it quiets tips and suppresses epilogues, and only the table case is pinned. Narrow it to the tabular values or pin the broader behavior deliberately.
2. Release notes: restore the shipped release's line (history shouldn't be edited) and add an entry for the removal in the current notes — that entry is also the natural home for the env-var transition note from (a), which could make this whole thing one docs change.

Net: this PR is one small commit from merge — release-notes fix (carrying the env retirement note) plus the `IsFormatExplicitlySet` narrowing-or-pin.

---

richlander:
Re-reviewed at head 625e232. **Both amended asks landed — clear to merge.**

**Release notes: resolved exactly as asked.** The shipped release's historical line ("Keeps `--oneline` as a hidden compatibility alias") is restored, and the current notes gain the removal entry carrying the env retirement note with the migration path (`--table` / `DOTNET_INSPECT_FORMAT=table`). That's option (a) from my amendment — the transition is documented, nothing left to pin.

**`IsFormatExplicitlySet`: the pin-the-broader-semantics path was chosen, and it paid for itself.** The theory covers all four non-tabular env values (markdown/json/plaintext/mermaid → `ResolveFormat` matches, `IsFormatExplicitlySet` true, `IsTableExplicitlySet` false — the boundary I wanted stated), plus the downstream tip-suppression consequence pinned for env=markdown. And writing those pins exposed a real incoherence: `DOTNET_INSPECT_FORMAT=json` counted as an explicit format (quieting tips) **without actually selecting JSON output**, because `JsonOutput` read the raw `--json` flag. The `625e232` fix routes `JsonOutput` through `ResolveFormat` uniformly across the member/type/find/package parsers and the search/depends definitions — bringing JSON in line with how Mermaid already resolved, so a recognized env value now both counts as explicit *and* takes effect. That's the pin-first discipline doing exactly what it's for: the deliberate-semantics test forced the semantics to actually be coherent before they could be pinned.

The branch is also current with main (merge base at the #2699 merge tip), so the timeline terminology commits are consistent with what landed. Nothing outstanding.

---

richlander:
Adversarial review reconciliation for final head `64e99a3502914c12bd0815d836ca1badd3d7d801`:

- Claude Opus 4.8 r14: CLEAN. Verified removal of executable `--oneline`/`--terse`/`OneLine` surfaces, env-format resolution including `DOTNET_INSPECT_FORMAT=json`, baseline correctness, markdownlint smoke config, RowWindow/Tabular merge resolutions, and the shared-state test serialization fix.
- MAI Flash r14: CLEAN. Verified the same final head in an isolated clean worktree and reproduced the fast CLI test suite with 1340 total, 0 failed, 10 skipped.

Actionable prior feedback was resolved by follow-up commits:
- `625e2328` routes `JsonOutput` through the effective resolved output format so env `json` is honored.
- `a08c851a` makes pack smoke markdownlint use the repo config and documents generated padded tables.
- `235390d2` resolves merge conflicts by combining origin/main's `RowWindow` row-selection model with this PR's `Tabular` naming.
- `64e99a35` serializes tests that mutate global cache/environment state to remove the CI full-suite race.

CI is green on the final head. Ready to merge.

richlander:
Re-verified at head 64e99a3 (post-main-merge). The two new commits are housekeeping and both responsive: `[Collection("Console")]` on the env-reading test classes closes the parallel-flake window I noted when the env-override cache was removed, and the markdownlint config change is CI plumbing. **Clearance stands — merge when ready.**

---

PR #2704, opened 2026-07-13
Add --rows --tail and --row first|last selectors (#2694)

### LINKED ISSUE #2694: Add symmetric row-tail and printable endpoint selectors
PR #2691 locks the output-shape grammar but leaves two implementation follow-ups.

## Desired behavior

- `--head N` / `--tail N` remain post-projection rendered-line windows.
- `--rows --head N` keeps the first N data rows per rendered table.
- Add `--rows --tail N` to keep the last N data rows per rendered table while preserving headings and table headers.
- Extend printable-row selection from numeric `--row N` to `--row N|first|last`.
- Keep both row-window forms incompatible with `--print` and `--print-all`; only `--row` resolves printable-row cardinality.
- Line and row windows must not constrain payload fetching or operation acquisition.

## Current gaps

- The CLI rejects `--rows --tail N` in Program.cs.
- Shared `--row` parsing accepts only integers.

Add focused xUnit v3 executable tests for Markdown, TSV/JSONL where supported, print endpoint selection, incompatibility diagnostics, and preservation of post-projection line-window behavior.

richlander:
Closes #2694. Follow-up to #2691 (grammar lock).

## What

Completes the symmetric output-shape grammar:

1. **`--rows --tail N`** — keep the **last** N data rows per rendered table (Markdown, TSV, JSONL), mirroring the existing `--rows --head N`. `--rows` now requires exactly one of head/tail and rejects both together.
2. **`--row N|first|last`** — extend printable/projected-row selection with symbolic endpoints. `first`→row 1, `last`→last row, resolved against the actual row count at render time. An invalid token is rejected by an option validator.

Both row-window forms remain incompatible with `--print`/`--print-all`; only `--row` resolves printable-row cardinality. Line/tail **line** windows (`--tail N` without `--rows`) are unchanged and still run post-projection.

## How

New value types in `Output/RowSelection.cs`:

- `RowWindow(Count, FromEnd)` — head (`--head`) vs tail (`--tail`) data-row window. Implicit `int→RowWindow` (bare count = head) keeps count-only call sites concise.
- `RowSelector` (`First`/`Last`/`FromIndex`, `Resolve(count)`, `TryParse`) — the `--row` selection, resolved lazily at the single render site that knows the row count.

`int? Rows`→`RowWindow?` and `int?/int? PrintRow`→`RowSelector?` thread through the options records and the three `--row` resolution sites (`ShapeProjectionOutput`, `PrintProjectionOutput`, `ApiCommand`); call sites like `Apply(markdown, options.Rows)` stay textually identical.

## Tests

- `OutputFormatterTests`: Markdown/TSV/JSONL tail windows (+ tail-wider-than-table), `RowSelector.TryParse` keyword/index/reject-invalid, and `Resolve` first/last/index. Full `dotnet-inspect.Tests` suite green (1832 total, 0 failed, 10 pre-existing skips for ilasm/ildasm).
- CLI: manually verified `--rows --head/--tail`, both-together and neither diagnostics, `--tsv/--jsonl --rows --tail`, and the `--row` invalid-token validator against CoreLib. `RowSelector.Resolve` endpoints covered by unit tests.

## Docs

`docs/design/output-shapes.md` and `docs/design/rendering-model.md` updated to reflect the now-implemented tail rows and `first|last` endpoints. `markdownlint` clean.

## Review

Non-trivial CLI grammar change → requesting two cross-model adversarial reviews per AGENTS.md.

### DIFF
--- a/docs/design/command-transition-model.md
+++ b/docs/design/command-transition-model.md
@@ -229,7 +229,7 @@ Shape reducers do not revise operation arity. In particular:
 - `--urls` may expose registry URLs if version rows gain such a field, but it
   must not download package contents to manufacture them;
 - `--print` is exactly-one, not implicit-first: one printable row prints
-  directly, multiple printable rows require `--row N` or `--print-all`, and zero
+  directly, multiple printable rows require `--row N|first|last` or `--print-all`, and zero
   printable rows reject;
 - `--head N` and `--tail N` run after print selection and fetching, so
   `--print --head 1` does not select the first printable row and
--- a/docs/design/output-shapes.md
+++ b/docs/design/output-shapes.md
@@ -139,9 +139,8 @@ windows:
 - `--rows --tail N` keeps the last N data rows.
 
 Both row-window forms are incompatible with `--print` and `--print-all`;
-`--row N|first|last` is the explicit printable-row selector. The current CLI
-implements head rows and rejects tail rows; tail-row support is a follow-up to
-this symmetric contract.
+`--row N|first|last` is the explicit printable-row selector. The CLI implements
+both head and tail data-row windows symmetrically.
 
 This policy deliberately rejects implicit-first behavior. Row order may change
 with filtering, producer evolution, or package versions, and choosing the first
@@ -173,7 +172,7 @@ payloads on already evaluated rows; it cannot probe missing cells.
 | `-n N` / `--head N` / numeric shorthand such as `-20` | keep the first N rendered output lines unless `--rows` is active |
 | `--tail N` | keep the last N rendered output lines unless `--rows` is active |
 | `--rows --head N` | keep the first N **data rows per table**, across Markdown, TSV, and JSONL |
-| `--rows --tail N` | keep the last N **data rows per table**; target symmetry, not yet implemented |
+| `--rows --tail N` | keep the last N **data rows per table**, across Markdown, TSV, and JSONL |
 | `--bare` | render the selected payload without document decoration; it changes presentation only, not the selected shape |
 | `--plaintext` | render a whole-document plain-text view; distinct from `--bare` |
 
--- a/docs/design/rendering-model.md
+++ b/docs/design/rendering-model.md
@@ -56,10 +56,10 @@ The `package` command inspects a NuGet package. Its default view is *package ide
 | `--path` | File resolution | Table of package-relative file paths and sizes; repeatable with `--match all` or `--match first` |
 | `--content` | File content | Contents for files selected by `--path`, with separator blocks or `--jsonl` rows |
 | `--readme` | README content | Compatibility alias for best package grounding content (`AGENTS.md` > `README.md` > `PACKAGE.md` > declared readme); supports `--frontmatter`/`--body` |
-| `--value` | Scalar projection | Prints one scalar cell or field from a selected section; use `--row N` when multiple rows match |
+| `--value` | Scalar projection | Prints one scalar cell or field from a selected section; use `--row N\|first\|last` when multiple rows match |
 | `--urls` | URL projection | Prints URL-bearing selected-section rows as a URL list, JSONL rows, or a JSON array |
 | `--paths` | Path projection | Prints path-bearing selected-section rows as a path list, JSONL rows, or a JSON array |
-| `--print` | Row payload | Prints one document behind a selected printable section row; use `--row N` to choose the Nth printable row when multiple printable rows exist |
+| `--print` | Row payload | Prints one document behind a selected printable section row; use `--row N\|first\|last` to choose the printable row when multiple printable rows exist |
 | `--print-all` | Row payloads | Prints every printable row from one selected section, using separators for text, JSONL rows, or one JSON array with `--json-array` |
 | `--versions` | Version history | Available versions from nuget.org |
 | `--library` | Library metadata | Delegates to library inspection |
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -399,7 +399,16 @@ public class CommandExecutionTests
         {
             args = CommandLineBuilder.PreprocessArgs(args);
             var root = CommandLineBuilder.CreateRootCommand();
-            return await root.Parse(args).InvokeAsync();
+            try
+            {
+                return await root.Parse(args).InvokeAsync();
+            }
+            catch (RowWindowValidationException ex)
+            {
+                // Mirror Program.cs so --rows window validation surfaces as exit 1 + stderr.
+                Console.Error.WriteLine($"Error: {ex.Message}");
+                return 1;
+            }
         });
     }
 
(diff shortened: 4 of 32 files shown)


richlander:
## Adversarial review reconciliation (round 1)

Two cross-model reviews (I authored as Claude Opus 4.8 → reviewers from the other families), each in an isolated worktree at the PR head.

| Reviewer | Result | Finding |
| --- | --- | --- |
| **GPT-5.5** | 1 Medium | `--rows --tail 0` / `--head 0` on a headerless format (JSONL, `--no-header` TSV) emitted a phantom blank line — an invalid empty JSONL record. |
| **Gemini 3.1 Pro** | CLEAN | Cosmetic note: `--rows` incompatibility diagnostics still said `--row N`. |

Agreement: both exercised tail/head boundaries (N=0/1/count/>count), TSV-no-header, JSONL, separator handling, the implicit `int→RowWindow`, the `--row` validator + downstream range checks at all three sites, and Program.cs validation. Gemini also ran a byte-level A/B against the merge-base (`-S "Member Index" --rows -n 3` for `--tsv`/`--jsonl`) → identical output. No divergence on correctness.

### Fixes (commit `1d515c90`)

1. **Zero-width headerless window** (GPT-5.5): `OutputFormatter.LimitRenderedTableRows` now returns `string.Empty` when the kept header+window is empty, instead of re-adding the trailing newline. This was **pre-existing for `--head 0`**; the fix corrects both head and tail. Verified live: JSONL/`--no-header` TSV zero window → 0 bytes; TSV *with* header zero window → header row only. Added `OutputFormatterTests` regressions (`*ZeroWindow*`).
2. **Message consistency** (Gemini): the 8 `--rows` incompatibility diagnostics in Api/Library/Package/Project now say `--row N|first|last`.

Full `dotnet-inspect.Tests` suite green (1834 total, 0 failed, 10 pre-existing ilasm/ildasm skips). Re-review of the fixed head with both reviewers is in progress; will confirm two CLEAN sign-offs before marking Ready to merge.

richlander:
## Ready to merge

Both re-reviews of the fixed head (`1d515c90`) came back **CLEAN**:

- **GPT-5.5** — CLEAN. Re-ran its original repro: JSONL / `--no-header` TSV zero window → 0 bytes; TSV-with-header zero window → header only; non-zero head/tail windows unchanged; changed diagnostics now say `--row N|first|last`; `OutputFormatterTests` 101/0.
- **Gemini 3.1 Pro** — CLEAN. Confirmed the zero-window fix (headerless → empty, header → header-only, non-zero unchanged) and the 8 message edits, no regressions.

CI green: `build-net10` pass, `markdownlint` pass, `test (ubuntu-26.04, linux-x64)` pass.

No outstanding findings. Not merging — awaiting @richlander's explicit approval.

richlander:
Reviewed at head 1d515c9. The core is right — **one fix round: close the `=`-syntax gate bypass, and give the new grammar end-to-end pins.**

**What's right.** Both selectors are defined once in `SharedOptions` and consumed at the single render sites that know the row count — no per-command reimplementation, no unconverted `int? Rows` stragglers. Shape-reducer purity holds everywhere I care about it: `RowWindow` is applied strictly after serialization, `RowSelector.Resolve` takes counts from already-assembled row lists, and on the one acquisition-adjacent path (api URL printing) `--row last` narrows fetches to one payload — fewer, never more, exactly what command-transition-model.md's reducer table requires. `--tail` preserves the markdown header+separator (pinned), and the review-fix commit is a genuine catch: a zero-width window over a headerless format (JSONL, `--no-header` TSV) used to emit a phantom blank line — an invalid empty JSONL record — and now returns empty, pinned in both head and tail forms including the header-only TSV case.

**Ask 1 — the two-parser split defeats the "exactly one window" gate.** Program.cs gates on `ArgumentPreprocessor`'s exact-token scan (`args[i] == "--tail"` + separate next token), but the actual window comes from System.CommandLine's `ParseRows`, which prefers head. Two consequences:
- `--rows --head=3 --tail 2` slips past the both-set check (the preprocessor doesn't see `=`-syntax) and **silently applies a head window of 3**, directly contradicting the PR's own "cannot combine" error;
- `--rows --tail=2` falsely errors with "requires -n/--head N or --tail N" even though a valid tail was given.

The (b) half mirrors a pre-existing `--head=N` quirk, but (a) is a new silent-precedence path in this PR's grammar. Moving the conflict check into `ParseRows` (where both parse results are visible) closes both.

**Ask 2 — zero end-to-end coverage of the new grammar.** All fourteen new tests are unit-level on the formatter/selector. No `CommandExecutionTests` exercises `--rows --tail N` or `--row first|last` through a real command, and neither of the two new Program.cs error messages is pinned. The existing e2e `--rows` tests are all head-form, so a command that wires `ParseRows` but mishandles the tail branch regresses invisibly. A few cases (one command × tail window, one × `--row first`, the dual-window error) would close it.

Non-blocking:

1. **Mid-table separator lines are reordered** in the rewritten `MarkdownTableRowLimiter`: interior separators used to be emitted in place and are now hoisted after the windowed data rows — an unpinned behavior change in a shared limiter that now applies to the *head* form too. Probably no current producer emits interior separators, but either restore in-place emission or pin the new order deliberately.
2. `cache` accepts `--rows --tail 5` and silently does nothing (`CacheOptions` never consumes it) — pre-existing accepted-but-inert surface, widened by this PR without comment. Worth an issue if not a fix.
3. The implicit `int → RowWindow` conversion keeps call sites tidy today but silently bakes head semantics into any future `int` assignment — a small foot-gun to remember when the next window feature lands.
4. Doc drift: command-transition-model.md still says "require `--row N` or `--print-all`" in one sentence while everything around it says `--row N|first|last`.

---

richlander:
Re-reviewed at head ce65cfd (rebased onto current main). **Both blocking asks landed, plus all three non-blocking items — clear to merge.**

**Ask 1: fixed at the right layer.** The token-scan gate in Program.cs is deleted (not doubled) and validation moved into `SharedOptions.BuildRowWindow`, which reads the real System.CommandLine parse values — so `=`-syntax and concatenated forms are covered by construction. Both-set throws, neither-set throws, and `RowWindowValidationException` surfaces as the same one-line CLI error. Both bypass shapes I cited are e2e-pinned: `--head= + --tail` now exits 1 with the conflict message (`Rows_RejectsBothHeadAndTailWithEqualsSyntax`), and `--tail=N` — the falsely-erroring form — now applies a tail window (`Rows_TailEqualsSyntaxAppliesTailWindow`). The comment at the old site explaining *why* validation can't live in the preprocessor is exactly the kind of constraint note that earns its place.

**Ask 2: the grammar has real end-to-end pins now** — head-vs-tail windows differing through a real command, both error paths with exit codes and message text, and `--row first`/`--row last` verified to fetch the first/last source respectively (pinning selector semantics at the acquisition-adjacent site, which is the one that matters).

**Non-blocking items, all landed:** interior separators are emitted in place (order preserved, with a unit pin — the unpinned behavior change is gone), the implicit `int → RowWindow` operator is removed with call sites made explicit (the foot-gun is dead), and the command-transition-model doc drift is fixed.

One residual wrinkle from the validation move, noting for the record rather than asking: `cache` accepts `--rows`/`--tail` but never calls `ParseRows`, so it now skips window validation entirely — `cache --rows` (previously a Program.cs-level exit 1) and even `cache --rows -n 5 --tail 3` now run silently with the options ignored. That's the pre-existing accepted-but-inert surface I flagged, made marginally wider by moving validation into the consumer. Fine as-is for this PR; the clean fix is removing the output-options group from `cache` (or wiring it), which deserves its own small issue.

---

richlander:
Post-merge follow-up: after this PR squash-merged at `ce65cfd0`, the two
adversarial reviewers (GPT-5.5 + Gemini 3.1 Pro) both independently found that
the `--rows` head/tail validation threw `RowWindowValidationException` from
inside the command `SetAction`, so `System.CommandLine`'s invocation pipeline
swallowed it into an **"Unhandled exception" stack trace** (exit 1) instead of a
clean `Error: ...` line — the `Program.cs` catch was dead code. The e2e tests
here passed falsely because the stack-trace text contains the message string.

Fix opened as #2714: validation moved to a parse-time command validator in
`SharedOptions.AddOutputOptionsTo` (covers `=`-syntax and concatenated forms),
plus a `RunAppAsync` fidelity fix. Both reviewers re-reviewed the fix and
returned CLEAN with real-run evidence. Full suite 1852/0-failed.

Also filed #2712 (non-blocking): `cache` accepts a valid single `--rows` window
inertly now that validation is per-command rather than a global gate.

PR #2705, opened 2026-07-13
Compose timelines with analysis findings

richlander:
## Conclusion

`timeline` now composes package version correlation with Analysis descriptors:

```bash
dotnet-inspect timeline --package Foo@1.0.0..2.0.0 \
  --type Foo.Parser --member Parse \
  --finding analysis.unsafety --at all
```

The same member focus supports `analysis.allocation` and
`analysis.call-site`. `api.member --member M` now selects one exact member
identity track rather than the full type-owned member census.

## Contract

- No `--at` still acquires zero package payloads.
- Sparse and dense traversal preserve `Complete([])`, `SubjectAbsent`,
  `Failed`, and `Unevaluated` cells.
- Analysis decodes only the resolved method body in the package assembly.
- Adjacent transitions use each Analysis producer's native comparison;
  gap-spanning rows remain qualified candidate boundaries.
- Metadata resolves structural focus, Analysis owns occurrences, and Findings
  owns N-address correlation; no parallel timeline model was added.

## Evidence

| Check | Result |
| --- | --- |
| Focused timeline tests | 15 passed |
| Full CLI executable suite | 1,821 total, 0 failed, 10 tool-dependent skips |
| Live package timelines | `api.member` plus all three Analysis descriptors |
| Changed Markdown | markdownlint clean |

## Review

Two isolated exact-head adversarial reviews completed cleanly on `1213cc49`:
Claude Opus 4.8 and Gemini 3.1 Pro. Both exercised selector identity,
multi-assembly/body-token scoping, cell-state preservation, package lifetime,
specialized comparisons, zero-acquisition behavior, and output routing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,7 @@ dotnet-inspect member MyType MyMethod:1 --library MyLib.dll -S "Unsafe*"
 dotnet-inspect library System.Text.Json --il-offset 0x06000004+0x15
 dotnet-inspect diff --package System.Text.Json@9.0.0..10.0.0 --breaking
 dotnet-inspect timeline --package System.Text.Json@8.0.0..9.0.0 --type System.Text.Json.JsonSerializer --members --at all
+dotnet-inspect timeline --package MyLib@1.0.0..2.0.0 --type MyType --member Parse --finding analysis.unsafety --at all
 dotnet-inspect diff --library old/Foo.dll..new/Foo.dll -S "Implementation Diff" -m MyType.HotPath
 dotnet-inspect depends Stream --markdown --mermaid
 dotnet-inspect implements IEquatable --project ./src/App -v:q
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -39,7 +39,7 @@ The tool is organized around source inspection, API lookup, relationship, and ut
 │  Compare API, analysis, or implementation evidence           │
 ├─────────────────────────────────────────────────────────────┤
 │                       timeline                               │
-│  Correlate type-owned Findings across a package range         │
+│  Correlate API or member-body Findings across a package range │
 ├─────────────────────────────────────────────────────────────┤
 │            depends / extensions / implements                 │
 │  Relationship discovery for APIs, packages, and libraries    │
@@ -110,12 +110,15 @@ Compares two package, platform, or local-library versions:
 
 ### timeline
 
-Correlates one type-focused Finding census across an inclusive package version
-vector:
+Correlates one type- or member-focused Finding census across an inclusive
+package version vector:
 
 - `--finding api.type` observes the type's own presence and facets.
 - `--finding api.member` observes all members owned by the type.
+- `--finding api.member --member M` selects one exact member identity track.
 - `--finding api.attribute` observes applied attribute occurrences.
+- `--finding analysis.allocation`, `analysis.call-site`, or
+  `analysis.unsafety` observes one selected method's Analysis census.
 - No `--at` evaluates zero package payloads; repeated `--at` selectors perform
   sparse probes; `--at all` explicitly authorizes dense traversal.
 - `Evaluations` preserves `Present`/`Missing` self-presence through an exact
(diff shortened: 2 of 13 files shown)


richlander:
Adversarial review reconciliation for exact head `1213cc49`:

| Reviewer | Result |
| --- | --- |
| Claude Opus 4.8 | CLEAN |
| Gemini 3.1 Pro | CLEAN |

Both reviewers independently verified the same attack surface in isolated
worktrees: exact `api.member` key selection and sibling filtering, overload and
method-like target handling, multi-assembly resolution, selected MethodDef
scoping, AssemblySet temporary-directory lifetime, sparse/failure/absence/empty
cell semantics, specialized Analysis comparisons, validation-before-acquisition,
zero-`--at` behavior, output routing, and SRM/NativeAOT boundaries.

They agreed that no high-confidence correctness, compatibility, security, or
contract issue remained. There were no divergent or dismissed findings and no
review-resolution commit applies.

richlander:
Ready to merge

richlander:
Reviewed at head 1213cc4. **Endorse — safe to merge.** This is the cross-family composition proof done the way the docs promised, and the docs say so in the right places.

**The composition is genuinely three-family with no new model.** Metadata resolves the structural member focus (`MemberTargetSelector`/`MemberTargetResolver`, the same machinery the `member` command uses), Analysis supplies producer-native censuses (`InspectAllocations`/`InspectCallSites`/`InspectUnsafety` + their `Compare*` folds at threshold 100), and Findings owns the N-address correlation — both paths converge on one shared `BuildCorrelatedView<T>` over `FindingCensusCorrelation<T>`, with the old Metadata-only builder refactored into a delegating adapter rather than left as a sibling. command-transition-model.md's "does not introduce a Research-owned timeline model" is true in the code. The exact-identity track for `api.member --member M` goes through `Correlate(key)` per rule 7, and I verified the key shape: `ResolveMemberCorrelationKey` builds `FindingKey(handle.CanonicalSignature ?? handle.Identity, type.FullName)` via the same `ApiMemberIdentity.CreateHandle` path `InspectApiMembers` uses, so Present/Missing can't silently diverge — and `MemberSelector_CorrelatesOneExactIdentity` pins it end-to-end (Present→Missing cells, and the transition filter shows only Run's removal, not Stop's facet change).

**The contract discipline held under extension:**
- Validate-before-acquisition extended correctly: `--finding analysis.*` requires `--member` and `--member` is rejected for `api.type`/`api.attribute`, both checked before any acquisition — pinned by `AnalysisTimeline_RequiresMemberBeforeAcquisition`.
- Acquisition cardinality unchanged: analysis inspection happens only on caller-selected cells, and each cell decodes *only* the selected method body (`LibraryBodyIndex.Open(..., bodyScope: ImmutableHashSet.Create(token))`) — pinned by `AnalysisTimeline_InspectsOnlyResolvedMethodBody`, which is a nice test (it runs against the test assembly's own unsafe method).
- Outcome semantics are honest at every fork in `InspectAnalysisAssemblies`: type/member absent → `Absent` with which-one detail; ambiguous selector or multi-assembly resolution → `Failed` with reason; non-method-like target → `Failed` naming the resolved kind; bodiless member → `Absent`. Per-cell analysis exceptions become that cell's `Failed` — the rule-7 failure topology carries over.
- The census-consistency invariant now holds through value equality (#2701's contribution — the producer comparison's rebuilt inspections compare equal to the correlated censuses by content), which is exactly the composition the equality PR was supposed to enable.
- All eleven prior pins retained; the skills docs keep the two-step honest ("a gap-spanning `Added` row locates a candidate boundary... Confirm the adjacent pair" → `diff`).

Two non-blocking items:

1. **`using ILInspector.Research;` is unused** — nothing in the file references Research, which is good (it's the doc claim), so the import should go before it invites a dependency.
2. **Endpoint lifetime widened from per-cell to per-command.** #2699 disposed each `ApiSurfaceEndpoint` immediately after surface extraction; this PR retains all endpoints until the command's `finally` because analysis inspection reads `Endpoint.Paths` after the evaluation loop. For `--at all` on a long vector that's N extracted packages held on disk simultaneously where one at a time would do — restructuring to run the analysis inspection inside the per-cell loop (and dispose each endpoint there) would restore the tighter footprint, or a one-line comment stating the tradeoff is deliberate. Dense traversal is opt-in and bounded, so this is a follow-up, not a blocker.

---

PR #2707, opened 2026-07-13
Add portable PDB Finding producers

richlander:
## Summary

- add Metadata Finding producers and comparisons for portable-PDB source documents, member-to-document mappings, compilation options, and compilation references
- support complete source-document censuses with an optional canonical-path substring query, while retaining `Complete([])`, `Absent`, and `Failed`
- migrate SourceLink availability/integrity to the source-document census and member source-location enrichment to metadata-token-scoped mappings
- document producer inputs/outputs, identities, consumer boundaries, and APIs eligible for retirement after migration

## Product boundaries

The new producers are local, SRM-only, NativeAOT-compatible, Roslyn-free, and network-free. Source fetching, checksum verdicts, authored/decompiled correspondence, and rebuild fidelity remain downstream operations.

Low-level PE/PDB extraction remains the authoritative substrate. Once remaining consumers migrate, convenience APIs such as `GetTrackedFiles`, `GetEmbeddedFiles`, and overload-ordinal `ResolveMethodSource` can be removed or made internal; source audit/integrity summaries and the public-type source view remain distinct product projections.

Advances #2673 without closing its SourceLink acquisition or oracle slices.

## Validation

- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release --no-build` (459 passed)
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-build` (1,817 passed, 10 tool-dependent skips)
- `npx markdownlint-cli docs/design/source-finding-producers.md docs/design/finding-producers.md docs/overview.md docs/architecture.md`
- CI: green

## Review

- Claude Opus 4.8: CLEAN on exact head `bdccf1b5`
- Gemini 3.1 Pro: CLEAN on exact head `bdccf1b5`
- Owner feedback and adversarial findings are reconciled in the PR comments.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -231,6 +231,13 @@ producers or preserve producer-native structural evidence. Failed censuses stay
 distinct from empty results and render as inspection diagnostics without
 suppressing unaffected sections.
 
+Portable-PDB observations use four Metadata Finding families: source documents,
+member-to-document mappings, compilation options, and compilation references.
+Source audit and integrity consume the document census; member source-location
+enrichment consumes the mapping census by metadata token. Network acquisition
+and checksum agreement remain operation outcomes rather than Metadata Findings.
+See [Source Finding Producers](design/source-finding-producers.md).
+
 Extension discovery follows the same rule for both `library` and `extensions`:
 each assembly is scanned once into Metadata's extension-member Finding census,
 then CLI target, reachability, overload, and display projections consume that
--- a/docs/design/finding-producers.md
+++ b/docs/design/finding-producers.md
@@ -193,6 +193,13 @@ path. The member command's opt-in `Fidelity Causes` section keeps the complete
 projects it only when rendering. A complete empty census, absent method body,
 and failed inspection therefore remain distinct product outcomes.
 
+Portable-PDB source observations follow the same product-first path.
+`MetadataFindings` exposes source documents, member-to-document mappings,
+compilation options, and compilation references as separate identity-set
+censuses. Source reachability, checksum verification, authored/decompiled
+correspondence, and rebuild fidelity remain consumer operations over those raw
+observations. See [Source Finding Producers](source-finding-producers.md).
+
 The remaining Analysis safety observations preserve their native families:
 
 - declaration, signature, local, call, and opcode evidence use an identity
--- a/docs/design/source-finding-producers.md
+++ b/docs/design/source-finding-producers.md
@@ -1,0 +1,87 @@
+# Source Finding Producers
+
+Metadata owns the local, SRM-only observations available from PE and portable
+PDB data. Network source acquisition, checksum verdicts, decompiler source
+correspondence, and old/new product interpretation remain separate consumers.
+
+## Producer inventory
+
+| Producer | Input | Output |
+| -------- | ----- | ------ |
+| `InspectSourceDocuments` | A `SourceLinkService` with a loaded portable PDB, or an already-extracted `IEnumerable<SourceDocument>` | `FindingInspection<SourceDocumentObservation>` with one identity-set Finding per PDB document |
+| `InspectMemberSources` | A loaded PE/PDB pair, or extracted `IEnumerable<MemberSourceInfo>`; an optional metadata-token query narrows the census | `FindingInspection<MemberSourceObservation>` with one Finding per member/document relationship |
+| `InspectCompilationOptions` | Portable-PDB module `CompilationOptions` custom debug information, or extracted option rows | `FindingInspection<CompilationOptionInfo>` with one Finding per option |
+| `InspectCompilationReferences` | Portable-PDB module `CompilationMetadataReferences` custom debug information, or extracted reference rows | `FindingInspection<CompilationReferenceInfo>` with one Finding per compiler reference |
+
+The service overloads return `Absent` when no portable PDB is loaded and
+`Failed` when the PDB cannot produce a complete census. The enumerable overloads
+are total over an already-acquired inventory, so an empty input is
+`Complete([])`.
+
+`InspectSourceDocuments` accepts an optional `SourceDocumentQuery`. Its
+`PathContains` value performs an ordinal-ignore-case substring match over the
+canonical source path. A null query returns every PDB document; a query with no
+matches returns `Complete([])`.
+
+The document producer does not restrict file extensions: every portable-PDB
+document is observable. Existing SourceLink availability and integrity sections
+retain their narrower C#, Visual Basic, and F# population when folding the
+census, preserving their established product metrics.
+
+## Identities and coordinates
+
+Source-document identity is the canonical path recovered from the SourceLink
+wildcard mapping when available. The original PDB path, document row, resolved
+URL, storage kind, and expected checksum remain typed payload data. Document-row
+renumbering does not make a comparison `Changed`.
+
+Member-source identity is the canonical `MemberAnchor` signature. Metadata token
+and document row are same-version coordinates; canonical document path and line
+range are the compared relationship. This lets consumers locate the exact PDB
+method without using overload ordinals.
+
+Compilation-option identity is the option name. Compilation-reference identity
+is its normalized reference name; aliases, image kind, embedding, timestamp,
+image size, MVID, and any currently-reserved flag bits are changeable payload
+facets.
+
+All four families use identity-set matching and leave `Finding.Ordinal` null.
+
+## Consumer boundaries
+
+`SourceAuditService` and `SourceIntegrityService` consume the source-document
+census. Their reachability and checksum statuses are operation results and
+presentation folds, not additional Metadata Findings.
+
+`MemberSourceLocationCollector` consumes member-source Findings by metadata
+token. The decompiler SourceLink oracle can later consume the same mappings,
+then join a complete source acquisition with the selected member observation.
+
+Compilation options and references describe available rebuild context. They do
+not claim that the context is complete enough to reproduce the original build.
+
+## Migration and API retirement
+
+The Findings initially sit over the existing PDB extraction substrate. Adoption
+does not require deleting the SRM readers that authoritatively produce document,
+sequence-point, option, and reference rows.
+
+Once all consumers use the censuses, these convenience APIs can be removed or
+made internal:
+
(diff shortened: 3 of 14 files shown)


richlander:

- **Claude Opus 4.8, first round:** found that an empty/whitespace portable-PDB document name could throw `ArgumentException` outside the expected malformed-metadata classification. Verified and fixed in `7c9024a4`: empty paths now throw `BadImageFormatException`, which produces a failed inspection.
- **Gemini 3.1 Pro, first round:** found that SourceLink mapping needed case-insensitive matching and that one malformed selected method could suppress other selected-member results. Verified and fixed in `7c9024a4`: matching is ordinal-ignore-case, and the consumer inspects each selected metadata token independently.
- **Agreement/divergence:** the reviewers found separate issues; there were no conflicting or dismissed findings, so all first-round guidance has a resolution commit.
- **Exact-head re-review:** Claude Opus 4.8 and Gemini 3.1 Pro both returned **CLEAN** on `7c9024a4`, using isolated clean worktrees and focused builds/tests/repros.

Local validation and CI are green. **Ready to merge.**

richlander:
Reviewed at head 7c9024a, both commits. The producer design is doctrinally sound and this is the right substrate for #2673's remaining slices — **one fix round: pin the `Failed` path, fix the per-member scan regression, and settle the multi-document selection change.**

**What's right, and it's a lot.** Four real three-outcome producers (`metadata.source-document`, `metadata.member-source`, `metadata.compilation-option`, `metadata.compilation-reference`) with the outcome forks in the correct places: no PDB → `Absent` (pinned across all four); PDB-without-SourceLink → `Complete` with `Storage = Unmapped` (correct — SourceLink absence is not census absence, and `HasSourceLink` gating stays in consumers); malformed metadata → `Failed` via typed `BadImageFormatException` guards; empty and filtered-to-empty → `Complete([])` (pinned). I found no exception-to-empty-Complete conversion anywhere. Identity discipline is exemplary: canonical paths and `CanonicalSignature` anchors as identity, with `DocumentRowId`/`MetadataToken`/line coordinates kept as provenance only — and *proven*, not just claimed, by the row-renumber→`Present` and token-renumber→`Present` pins. The consumer story is genuine migration: `SourceAuditService`/`SourceIntegrityService` now take the census (direct `GetTrackedFiles()` calls deleted, one census feeding two folds), the overload-ordinal `ResolveMethodSource` lane is retired from the collector (a version-local-coordinate correspondence gone — doctrine win), and the residual legacy lanes are named in source-finding-producers.md's retirement section rather than left silent. The CDI readers are flat iterative loops with guaranteed forward progress, and anchor creation routes through `GuardedProviderDecode.Method` — so the #2490 crash class is covered by the existing guard, verified. The first-round adversarial fixes (empty-path classification, case-insensitive SourceLink matching, per-member fault isolation replacing all-or-nothing batch failure) are each correct and pinned.

**Ask 1 — the `Failed` outcome ships unpinned.** The hardening commit exists to make malformation land as `Failed`, yet no test drives a producer to `Failed`: the empty-path test hits `Canonicalize` directly, and `PdbCompilationInfoReader`'s three `BadImageFormatException` guards (truncated reference row, non-null-terminated blob, empty name) are entirely unexercised. One malformed-PDB fixture through `InspectX(service, …)` asserting `Failed` with the reason closes the gap — same standard as the AttributeDecoder runtime pin on #2700: the outcome the commit is about needs a test that fails if it regresses.

**Ask 2 — the fault-isolation fix is O(selected members × assembly method count).** `MemberSourceLocationCollector` now calls `InspectMemberSources` once per member, and each call walks every `MethodDefinition` in the assembly (the token filter short-circuits the PDB read, not the walk). For dozens of members on a BCL-sized assembly that's a real regression vs both the batched pre-hardening call and the old direct resolve. Fault isolation doesn't require per-call enumeration — inspect per-token inside one enumeration pass keyed off the token set, keeping the per-member `Failed` isolation while walking once.

**Ask 3 — multi-document member selection changed semantics, unpinned.** The old lane used `debugInfo.Document` (the method's root document); the new code picks `OrderBy(DocumentRowId).First()`, which for partial/source-generated/`#line` methods can select the generated file where the old code selected the authored one. Either restore root-document preference or pin the new rule with a multi-document fixture and a comment saying it's deliberate.

Non-blocking, for follow-up:

1. **Two SourceLink-matching implementations must agree and nothing proves it**: identity canonicalization uses the new in-repo `SourceDocumentPath` (ordinal-ignore-case, longest-pattern-first) while `ResolvedUrl`/`Storage` classification comes from the external SourceLinkFetch resolver. A divergence makes the payload internally incoherent (prefix-stripped identity but `Unmapped` storage). A cross-check test — or deriving `Storage` from the same local match — would close it.
2. `IsPdbInspectionFailure` classifying `InvalidOperationException` as inspection failure launders future producer bugs into polite `Failed` rows, and the `Failed` helper keeps only `ex.Message` (type and stack discarded — triage suffers). Meanwhile `IOException` escapes the filter and lands in `LibraryMetadataService`'s whole-inspection `catch { return null; }`, nuking the library inspection instead of one census. Worth tightening in both directions.
3. The substrate returns `[]`/yield-break when `_pdbReader is null` — safe today because producers gate on `HasPdb` first, but it's the exact absence→empty shape the doctrine warns about if a future caller reaches the enumerable overloads directly. A one-line comment at those substrate sites would inoculate.
4. A `Failed` census leaves audit counters at defaults (`AllSourcesAccessible = false` indistinguishable from "audited and inaccessible" for a consumer that skips the failures projection) — pre-existing view risk now reachable via the census path; worth a consumer-level test.

---

richlander:
Addressed the review round in `5e22bddf`:

1. **`Failed` is now pinned through the public producer path.** `MalformedPortablePdb_ProducesFailedInspection` builds a matching portable PDB with a truncated compilation-reference CDI row, opens it through `SourceLinkService`, and asserts `InspectCompilationReferences` returns `FindingInspection.Failed` with the malformed-row reason.
2. **Selected-member inspection no longer scans the assembly once per member.** Token-scoped enumeration resolves requested MethodDef handles directly. The collector performs one batched inspection in the healthy path; if malformed metadata fails that batch, it retries per token for fault isolation, with each retry still O(1) handle lookup rather than a method-table scan. A real-PDB multi-token query test pins the batch path.
3. **Multi-document selection is explicit and pinned.** Relationships carry `IsPrimaryDocument`; the collector prefers the PDB root document. When a multi-document method omits the root, the first visible sequence-point document is deliberately primary. A real `#line` fixture pins that rule.

The non-blocking SourceLink resolver cross-check, exception-taxonomy/diagnostics tightening, substrate comments, and failed-audit presentation distinction are useful follow-up hardening, but are outside these three producer/collector corrections; no resolution commit applies to those suggestions in this PR.

Local build, 459 Metadata tests, 1,817 CLI tests (10 tool-dependent skips), and Markdown lint pass. Fresh exact-head adversarial reviews are in progress before this returns to ready-to-merge state.

richlander:
Follow-up `bdccf1b5` stabilizes the multi-document fixture identified by both reviewers: the `#line` directive now precedes the method declaration, so the first visible sequence point is deterministically the generated document in both Debug and repository-required Release builds. Clean focused runs pass in both configurations. Fresh exact-head reviews are running on `bdccf1b5`.

richlander:
Final review reconciliation for exact head `bdccf1b5`:

- The owner feedback was resolved in `5e22bddf`: public-path malformed-PDB `Failed` coverage, O(selected) token-keyed member inspection with isolated fallback, and explicit multi-document primary selection.
- Claude Opus 4.8 and Gemini 3.1 Pro both identified that the initial `#line` fixture behaved differently in Debug because the directive followed the method declaration. This was verified and fixed in `bdccf1b5` by moving the directive before the declaration; raw PDB checks and focused tests now confirm the same root-omitted/first-visible behavior in Debug and Release.
- Claude Opus 4.8: **CLEAN** on `bdccf1b5` after Release Metadata and CLI suites plus raw SRM fixture checks.
- Gemini 3.1 Pro: **CLEAN** on `bdccf1b5` after the full Release Metadata suite and focused edge-case review.
- No reviewer guidance remains dismissed or unresolved. The earlier Debug-only disagreement was superseded by the configuration-independent fixture fix.

Local validation and CI are green. **Ready to merge.**

richlander:
Re-reviewed at head bdccf1b. **All three asks landed — clear to merge.**

**Ask 1, the `Failed` fixture: landed properly.** `MalformedPortablePdb_ProducesFailedInspection` builds a real malformed portable PDB — a truncated `CompilationMetadataReferences` CDI blob, associated with the copied assembly via its actual CodeView GUID so `SourceLinkService.Open` genuinely binds it — and drives the producer end-to-end: `HasPdb` true, `InspectCompilationReferences` → `Failed` with "truncated" in the reason. That exercises the `PdbCompilationInfoReader` truncation guard through the full outcome-classification path, which is what the census-vs-strength distinction required. (The non-null-terminated and empty-name guards remain unexercised — fine as residue; the outcome path itself is now pinned.)

**Ask 2, the O(members × methods) scan: fixed at both layers, better than asked.** `PdbContext.EnumerateSelectedMethods` makes token-scoped queries direct handle lookups (token-type mask, row-bounds check, ordered enumeration) instead of scanning the method table — so even the fallback path is O(selected). And the collector restores the batched call as the primary path, keeping the per-token isolation only as the fallback when the batch fails, with the complexity claim stated in a comment. Fault isolation preserved, performance regression gone, and the unscoped census still walks the full table as it should.

**Ask 3, multi-document selection: resolved by making the rule explicit rather than accidental.** `IsPrimaryDocument` carries the portable PDB's `MethodDebugInformation.Document` (or the first visible sequence point's document when the root is omitted), the collector prefers primary with row order as deterministic fallback — explicitly restoring the legacy resolver's preference, per the comment — and the design doc states the rule. The `#line`-based multi-document fixture is a genuine reproduction of the shape I was worried about, and pinning `IsPrimaryDocument` into `MemberSourcesEqual` means a primary-document flip across versions surfaces as `Changed` rather than silently reordering presentation — a semantic widening, but the right one now that primary-ness is part of the compared relationship, and the doc says so.

The follow-up items from the first review (dual SourceLink-matching coherence, `IsPdbInspectionFailure`'s IOE breadth, substrate absence→empty comments, consumer-level Failed-census counters test) remain follow-up material — none blocking. With this, the #2673 slice 3–4 substrate is complete and merge-ready.

---

PR #2702, opened 2026-07-13
Fix shadowed type qualifiers in decompiler output

richlander:
- Fixes #2583
- Changes expression-side type qualifiers so static member references use `global::...` when the simple type name is shadowed by an in-scope parameter/local.
- Card revision: `2ccaca30`

## Change

**Conclusion:** **PASS** — shadowed type qualifiers now bind to the intended type instead of a same-named value.

Before:

```csharp
return TypeNameShadow.M(M);
```

After:

```csharp
return global::TypeNameShadow.M(M);
```

The normal no-shadow spelling remains unchanged, e.g. `Other.Helper(1)`.

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass: `dotnet build src/dotnet-inspect -c Release -p:PublishAot=false --configfile nuget.config` |
| Focused tests | Pass: `CSharpPrinterReceiverTests`, including parameter and local shadow canaries |
| Adjacent name tests | Pass: `NestedScopeNameCollisionTests`, `MemberNameCollisionRenderingTests`, `ForeignNestedTypeSpellingTests` |
| Decompiler suite | Pass: `dotnet run --project src/ILInspector.Decompiler.Tests -c Release` |
| Render A/B | Pass at `2ccaca30`: merge-base `81ba0eec`, 89,065 methods, 0 changed / 0 added / 0 removed |
| Real witness | Synthetic compile-back canaries for the #2583 binding failure |

## Decompiler quality

**Conclusion:** **PASS** — the pinned corpus gate matched baseline tolerances; semantic/fidelity rows are sampled and ungated as reported by the harness.

### Generated quality-diff card

```markdown
### Decompiler quality diff

Corpus: #1166 real-world decompiler corpus sensor: #1150 pinned NuGet assemblies plus dotnet-inspect managed assemblies. 14 assemblies, 89,065 methods
Correctness coverage: validity compiled 25,183 methods (compile-cap 4,000; per-sample, not corpus-wide); fidelity (compile-back) sampled 36 / 89,065 (0.04%)

| Metric | Change | Count delta |
| ------ | ------ | ----------- |
| Fully raised (+) | 78,406 (88.03%) → 78,406 (88.03%) (neutral) | 0 |
| Conditional-branch residual (-) | 2,401 (2.70%) → 2,401 (2.70%) (neutral) | 0 |
| Forward-merge stops (-) | 2,286 (2.57%) → 2,286 (2.57%) (neutral) | 0 |
[...]

PR #2708, opened 2026-07-13
Fix RTS constructor-chain opcode loss (#2678)

### LINKED ISSUE #2678: RTS constructor parity: empty bodies lose constructor-chain opcodes
Follow-up to #2280 and evidence surfaced by #2675.

## Problem

ReturnToSender emits valid but empty constructor bodies for a family of
`NuGet.Versioning` targets that the existing compile-back oracle checks as
`Exact`. RTS therefore reports `OpcodeDiff` and loses constructor chaining/body
opcodes.

The fixed-corpus cap-3 run at #2675 head `5a9ec6e2` surfaced three worse rows:

- `NuGet.Versioning.NuGetVersion::.ctor(string)`
- `NuGet.Versioning.SemanticVersion::.ctor(int, int, int)`
- `NuGet.Versioning.SemanticVersion::.ctor(int, int, int, string)`

A deterministic full useful-sample probe over `NuGet.Versioning` 7.3.0 finds
21 `Exact -> OpcodeDiff` rows, all constructors. The same 21 rows occur before
and after #2670 (`d46af1ba` and `5a9ec6e2`), so this is an existing RTS parity
gap rather than a provider-rendering regression.

Example for `NuGetVersion::.ctor(string)`:

```text
compile-back reference: Exact
RTS:                    OpcodeDiff
original opcodes:       ldarg ldarg call call ret
recompiled opcodes:     ldarg call ret
```

RTS source:

```csharp
public NuGetVersion(string version)
{
}
```

## Deterministic reproduction

```bash
  "$HOME/.nuget/packages/nuget.versioning/7.3.0/lib/net8.0/NuGet.Versioning.dll" \
  --emit-corpus-snapshot /tmp/rts-nuget-cap500.json \
  --compile-cap 0 \
  --corpus-fidelity-cap 500 \
  --corpus-fidelity-oracle return-to-sender \
  --sequential \
  --max-examples 3
```

Current result:

```text
Fidelity (return-to-sender, cap 500): 73 exact, 32 opcode diffs,
0 recompile failures, 0 context failures over 105 checked
RTS parity: 0 rescued, 84 same, 21 worse over 105 compile-back targets
```

## Acceptance

- Preserve constructor chaining and target-body semantics for these RTS targets,
  or explicitly route unsupported constructor reconstruction through the
  transition floor rather than silently compiling an empty body.
- Add a real compiled-fixture regression covering a constructor whose original
  body has `ldarg ldarg call call ret`.
- The deterministic `NuGet.Versioning` probe reports no constructor
  `Exact -> OpcodeDiff` rows.

richlander:
- Closes #2678
- Changes ReturnToSender to preserve `this(...)` constructor-chain opcodes; product change to `ConsumedMemberEvidence` is RTS-only (decompiler render output unchanged)
- Evidence revision: `229436cd`

## Change

**Conclusion:** **PASS** — RTS reconstructed target ctors with empty bodies, dropping the chain call the printer lifts into `DecompilerResult.ConstructorChain`; re-emitting it as a `this(...)` initializer, gated by a sound overload-safety predicate, moves 18 ctor rows `OpcodeDiff -> Exact` with zero regressions.

Before (recompiled `NuGetVersion(string)` ctor):

```text
ldarg call ret            # empty body, chain dropped -> Exact -> OpcodeDiff
```

After:

```text
ldarg ldarg call call ret # : this(Parse(version)) preserved -> Exact
```

## Scope and safety

- **Root cause:** `CompileBackSourceComposer.CreateTargetBody` kept only `printed.Output` and discarded `printed.ConstructorChain`, so the target ctor requirement rendered a bare `CSharpBlockBody` with no initializer.
- **Fix shape:** RTS orchestration — thread `ConstructorChain` through `ProductTargetBody` and the target `CompileBackMemberRequirement`, re-emit via `CSharpConstructorInitializer`. Restrict to `this(...)` self-chains (flat shells have no reconstructed base class, so `base(args)` -> CS1729).
- **Overload-safety gate (the core of adversarial rounds 1–7):** the printer can drop disambiguating casts (`CSharpPrinter.Numerics.cs` `CoerceText`, unmodified here), so a re-emitted `: this(args)` could bind to the wrong constructor (a sibling, or the target itself -> CS0516). The initializer is emitted **only** when the chained-to ctor's parameter types are all supported, a same-arity non-target ctor is present, **and either**:
  1. **Faithful args** — every argument is a non-lossy, non-target-typed expression whose result type structurally equals its parameter type (excludes `Box`/`Coerce`/`Convert`/`IsInstance`/`Constant`, which the printer can spell at a different static type, and the target-typed `Lambda`/`CollectionExpression`/`TupleExpression`/`InterpolatedStringExpression`, whose `ResultType` proves nothing about binding); **or**
  2. **Unique applicable candidate** — exactly one constructor in the whole shell (target included) is *applicable* to an N-argument call, modelling `params` absorption and optional-parameter omission (`ConstructorCouldBindToArgCount`), not just declared arity.
  The failure direction is conservative throughout: can't prove the bind -> drop the initializer -> the pre-PR empty-body OpcodeDiff, never a wrong bind.
- **Product impact:** `ConsumedMemberEvidence` now seeds `.ctor` Call callees with `AllowTargetRoot` so the flat shell reconstructs the chained-to sibling ctor. This layer is consumed only by ReturnToSender — decompiler render output is unchanged.
- **RTS boundary:** RTS only orchestrates closure + Roslyn + opcode compare; the printer/shared code owns chain lifting and C# source generation.

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass |
| Focused tests | `ReturnToSenderPrototypeTests` (107), `ConsumedMemberEvidenceTests` (8) passed |
| Regression tests | `PreservesThisConstructorChainOpcodes`, `PreservesConstructorChainWhenArityIsUnique` (Exact); `DropsInitializerWhenChainedConstructorUnreconstructable` / `...IsOverloadAmbiguous` / `...ChainBindsToTargetConstructor` / `...CrossArityParamsSiblingCanBind` / `...ChainArgumentIsAmbiguousLambda` (strip) passed |
| Deterministic probe | `OpcodeDiff -> Exact: 18`; total `OpcodeDiff` 30 -> 12; `RecompileFail` 63 -> 63 (no regressions) |

## Compile-back quality

**Conclusion:** **PASS** — 18 ctor `OpcodeDiff -> Exact`, zero new `RecompileFail`.

Run at head `229436cd`: `NuGet.Versioning.dll` (7.3.0 net8.0), `--corpus-fidelity-cap 500 --corpus-fidelity-oracle return-to-sender --sequential`.

```text
BASE: Exact 118, OpcodeDiff 30, RecompileFail 63, not-sampled 98
FIX : Exact 136, OpcodeDiff 12, RecompileFail 63, not-sampled 98

ctor BASE: OpcodeDiff 24, Exact 12, RecompileFail 3, not-sampled 14
ctor FIX : OpcodeDiff  6, Exact 30, RecompileFail 3, not-sampled 14

Transitions: OpcodeDiff -> Exact: 18   (no other transitions)
```

### Why 18, not the 21 of the first head

The first head (`1c121952`) emitted the chain from a weaker guard and scored 21 wins. Adversarial rounds 1–7 proved that guard could bind to the wrong overload (sibling hijack, self-referential CS0516, cross-arity `params` hijack, typeless-lambda ambiguity), so the gate was tightened to the sound faithful-args / unique-applicable-candidate predicate above. That stricter proof preserves **18** of the 21 (their arguments are provably faithful, e.g. `this(Parse(version))` where `Parse` returns the parameter type exactly, or their arity is uniquely applicable) and conservatively declines **3** `SemanticVersion` chains that pass a `string[]` argument to an `IEnumerable<string>` parameter alongside a same-arity `(Version, string, string)` sibling — not faithful (`string[]` != `IEnumerable<string>`) and not unique-applicable. Those 3 stay at the **pre-PR OpcodeDiff baseline** (they were OpcodeDiff before, still OpcodeDiff now) — a documented sound miss, **not a regression**. Proving `string[]` cannot bind to the sibling's `string` parameter would need a full type model.

Remaining ctor OpcodeDiffs at head (all pre-existing / by design / sound-miss — none are regressions):

```text
NuGet.SimplePool`1::.ctor                                                    # field-initializer lift (out of scope)
NuGet.StringBuilderPool::.ctor                                              # field-initializer lift (out of scope)
NuGet.Versioning.NuGetVersion(Version, IEnumerable<string>, string, string) # base(...) chain, flat shell -> left empty
NuGet.Versioning.SemanticVersion(int, int, int, string, ...)  x2            # string[] -> IEnumerable<string> sound miss
NuGet.Versioning.SemanticVersion(Version, string, string)                   # string[] -> IEnumerable<string> sound miss
```

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| `base(...)` chains in flat shells | Left as frontier | RTS shells are object-based; no base class to bind `base(args)` to (CS1729). Left empty as before. |
| Field-initializer ctors (`SimplePool`, `StringBuilderPool`) | Not changed | Different lifting (`FieldInitializers`); would need separate support. |
| `string[]` -> `IEnumerable<string>` chains (3 `SemanticVersion` ctors) | Conservative miss | Not provably faithful and not unique-applicable; stays at pre-PR OpcodeDiff baseline. No correctness risk; proving safety needs a full type model. |

## Review

Adversarial review per policy — two reviewers from model families other than my own (Claude Opus 4.8). Both **CLEAN on the final head `229436cd`**; every finding across rounds 1–7 was verified and resolved by a commit (`fc4065dc`, `fdc86627`, `9e4f805a`, `1128d5c8`, `d2a50bd9`, `229436cd`). See the round-5→7 reconciliation comment for the full trail.

| Reviewer | Result | Notes |
| --- | --- | --- |
| GPT-5.5 | CLEAN | Forced rebuilds; 107 RTS + 8 CME tests; probe `OpcodeDiff -> Exact: 18`, RF 63; stackalloc/span + target-typed probes clean |
| Gemini 3.1 Pro | CLEAN | Verified target-typed exclusion set complete; adversarial `decimal.One` (preserved Exact) + lossy `short` (stripped) fixtures; both branches sound |

**Review conclusion:** **PASS** — two CLEAN reviews on the final head.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class "ILInspector.Decompiler.Tests.ConsumedMemberEvidenceTests"
dotnet run --project tools/DecompilerHarness -c Release --no-build -- "$HOME/.nuget/packages/nuget.versioning/7.3.0/lib/net8.0/NuGet.Versioning.dll" --emit-corpus-snapshot /tmp/out.json --compile-cap 0 --corpus-fidelity-cap 500 --corpus-fidelity-oracle return-to-sender --sequential --max-examples 3
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -571,11 +571,11 @@ public class ReturnToSenderPrototypeTests
     {
         // Issue #2678 guard: the printer can render chain arguments without a
         // disambiguating cast (a `box` to `object` prints without `(object)`), so
-        // when two same-arity siblings are reconstructed the emitted
-        // `: this(args)` could silently bind to the wrong overload. The guard
-        // requires a unique same-arity sibling; here `C(object)` and `C(int)` are
-        // both present (the latter pulled in by `new C(2)`), so the initializer
-        // must be stripped rather than emit a wrong-binding chain.
+        // any second reconstructed constructor could silently hijack the
+        // `: this(args)` binding. The guard emits the initializer only when the
+        // chained-to constructor is the sole reconstructed sibling; here
+        // `C(object)` and `C(int)` are both present (the latter pulled in by
+        // `new C(2)`), so the initializer must be stripped.
         var assemblyPath = CompileFixture("""
             public class C
             {
@@ -590,6 +590,47 @@ public class ReturnToSenderPrototypeTests
                 public C(string z) : this((object)1)
                 {
                     _ = new C(2);
+                }
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("C", ".ctor", 2)]));
+
+            Assert.NotEqual(FidelityCheck.CompileBackStatus.RecompileFail, result.Status);
+            Assert.DoesNotContain(": this(", result.Source);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
+    public void CompileBackTargets_DropsInitializerWhenSiblingConstructorCrossesArity()
+    {
+        // Issue #2678 guard: overload resolution can bind a chain-call argument
+        // list to a constructor of a DIFFERENT IL arity (a `params` array absorbs
+        // extra arguments). A same-arity check alone would miss this, so the guard
+        // requires the chained-to constructor to be the sole reconstructed
+        // sibling. Here `C(params int[])` and `C(int, int)` are both present, so
+        // `: this(1, 2)` could bind to either; the initializer must be stripped.
+        var assemblyPath = CompileFixture("""
+            public class C
+            {
+                public C(params int[] xs)
+                {
+                }
+
+                public C(int a, int b)
+                {
+                }
+
+                public C(string z) : this(1, 2)
+                {
+                    _ = new C(9);
                 }
             }
             """);
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -482,6 +482,47 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
+    public void CompileBackTargets_PreservesThisConstructorChainOpcodes()
+    {
+        // Issue #2678: RTS used to reconstruct target constructors with empty
+        // bodies, dropping the `: this(...)` chain call. The recompiled ctor then
+        // emitted `ldarg call ret` instead of the original `ldarg ldarg call call
+        // ret`, producing an OpcodeDiff. The chain must be preserved so the ctor
+        // round-trips Exact.
+        var assemblyPath = CompileFixture("""
+            public class Versioned
+            {
+                public Versioned(string text) : this(Parse(text))
+                {
+                }
+
+                public Versioned(int value)
+                {
+                    Value = value;
+                }
+
+                public int Value { get; }
+
+                private static int Parse(string text) => text.Length;
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("Versioned", ".ctor", 0,
+                    "(corelib:System.String) -> corelib:System.Void")]));
+
+            Assert.Equal(FidelityCheck.CompileBackStatus.Exact, result.Status);
+            Assert.Contains(": this(", result.Source);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
     public void CompileBackTargets_UsesTargetBackingFieldWriteForConstructorAssignment()
     {
         var assemblyPath = CompileFixture("""
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -570,12 +570,12 @@ public class ReturnToSenderPrototypeTests
     public void CompileBackTargets_DropsInitializerWhenChainedConstructorIsOverloadAmbiguous()
     {
         // Issue #2678 guard: the printer can render chain arguments without a
-        // disambiguating cast (a `box` to `object` prints without `(object)`), so
-        // any second reconstructed constructor could silently hijack the
-        // `: this(args)` binding. The guard emits the initializer only when the
-        // chained-to constructor is the sole reconstructed sibling; here
-        // `C(object)` and `C(int)` are both present (the latter pulled in by
-        // `new C(2)`), so the initializer must be stripped.
+        // disambiguating cast (a `box` to `object` prints as its inner value), and
+        // C# overload resolution — including the target constructor — can then
+        // re-bind the call. A boxed argument is never faithful, and here three
+        // same-arity constructors (`C(object)`, `C(int)`, the target `C(string)`)
+        // share the callee's arity, so unique-arity binding does not hold either.
+        // The initializer must be stripped rather than emit a wrong-binding chain.
         var assemblyPath = CompileFixture("""
             public class C
             {
@@ -609,28 +609,25 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
-    public void CompileBackTargets_DropsInitializerWhenSiblingConstructorCrossesArity()
+    public void CompileBackTargets_DropsInitializerWhenChainBindsToTargetConstructor()
     {
-        // Issue #2678 guard: overload resolution can bind a chain-call argument
-        // list to a constructor of a DIFFERENT IL arity (a `params` array absorbs
-        // extra arguments). A same-arity check alone would miss this, so the guard
-        // requires the chained-to constructor to be the sole reconstructed
-        // sibling. Here `C(params int[])` and `C(int, int)` are both present, so
-        // `: this(1, 2)` could bind to either; the initializer must be stripped.
+        // Issue #2678 guard: overload resolution also considers the target
+        // constructor itself. The printer drops the `(object)` cast from
+        // `: this((object)text)`, printing `: this(text)`; with the target
+        // `C(string)` in the shell, `text` (a string) binds back to the target
+        // constructor — `C(string)` calling itself (CS0516). The argument is not
+        // faithful (its printed string type differs from the `object` parameter)
+        // and the callee shares the target's arity, so the initializer must be
+        // stripped.
         var assemblyPath = CompileFixture("""
             public class C
             {
-                public C(params int[] xs)
+                public C(object value)
                 {
                 }
 
-                public C(int a, int b)
+                public C(string text) : this((object)text)
                 {
-                }
-
-                public C(string z) : this(1, 2)
-                {
-                    _ = new C(9);
                 }
             }
             """);
@@ -638,10 +635,49 @@ public class ReturnToSenderPrototypeTests
         {
             var result = Assert.Single(ReturnToSender.CompileBackTargets(
                 assemblyPath,
-                [new ReturnToSender.RequestedTarget("C", ".ctor", 2)]));
+                [new ReturnToSender.RequestedTarget("C", ".ctor", 1)]));
 
             Assert.NotEqual(FidelityCheck.CompileBackStatus.RecompileFail, result.Status);
             Assert.DoesNotContain(": this(", result.Source);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
+    public void CompileBackTargets_PreservesConstructorChainWhenArityIsUnique()
+    {
+        // Issue #2678: a chain whose arguments the printer cannot type precisely (a
+        // bare `null`) is still safe to emit when exactly one constructor in the
+        // shell has the chain's argument count — a normal-form exact-arity match
+        // has no competing overload, so `: this(1, 2, null)` binds unambiguously to
+        // the sole three-argument constructor. The initializer must be preserved so
+        // the constructor round-trips Exact.
+        var assemblyPath = CompileFixture("""
+            public class C
+            {
+                public C(int a, int b, string s)
+                {
+                    S = s;
+                }
+
+                public C(string z) : this(1, 2, null)
+                {
+                }
+
+                public string S { get; }
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("C", ".ctor", 1)]));
+
+            Assert.Equal(FidelityCheck.CompileBackStatus.Exact, result.Status);
+            Assert.Contains(": this(", result.Source);
         }
         finally
         {
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -686,6 +686,50 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
+    public void CompileBackTargets_DropsInitializerWhenCrossArityParamsSiblingCanBind()
+    {
+        // Issue #2678 guard (Gemini R5): unique *declared* arity is not enough — a
+        // cross-arity `params` (or optional) sibling can absorb an N-argument call
+        // and, for a lossy argument the printer cannot type precisely, offer a
+        // better conversion that steals the bind. Here `C()` chains
+        // `: this((object)null)`; the printer drops the `(object)` cast, printing
+        // `: this(null)`. `C(object)` is the only one-parameter constructor, but
+        // `C(string, params int[])` is also applicable to a one-argument call and
+        // `null` binds to `string` (more derived than `object`) via params
+        // expansion — the wrong overload. The initializer must be stripped.
+        var assemblyPath = CompileFixture("""
+            public class C
+            {
+                public C(object x)
+                {
+                }
+
+                public C(string x, params int[] y)
+                {
+                }
+
+                public C() : this((object)null)
+                {
+                    _ = new C("hello");
+                }
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("C", ".ctor", 2)]));
+
+            Assert.NotEqual(FidelityCheck.CompileBackStatus.RecompileFail, result.Status);
+            Assert.DoesNotContain(": this(", result.Source);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
     public void CompileBackTargets_UsesTargetBackingFieldWriteForConstructorAssignment()
     {
         var assemblyPath = CompileFixture("""
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -730,6 +730,54 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
+    public void CompileBackTargets_DropsInitializerWhenChainArgumentIsAmbiguousLambda()
+    {
+        // Issue #2678 guard (Gemini R6): a lambda argument prints typeless
+        // (`() => ...`) and relies on C# target-typing, so its IR result type
+        // matching the parameter does NOT prove an unambiguous bind. Here `C()`
+        // chains `: this((Func<int>)(() => 1))`; the shell also contains
+        // `C(Expression<Func<int>>)` (pulled in by the body). Both constructors
+        // accept `() => 1`, so the printed `: this(() => 1)` is ambiguous (CS0121).
+        // The lambda must not be treated as faithful; with a same-arity sibling in
+        // the shell, unique-arity does not hold either, so the initializer is
+        // stripped and the body (which references the sibling through a typed local,
+        // so it stays unambiguous) still compiles.
+        var assemblyPath = CompileFixture("""
+            using System;
+            using System.Linq.Expressions;
+            public class C
+            {
+                public C(Func<int> x)
+                {
+                }
+
+                public C(Expression<Func<int>> y)
+                {
+                }
+
+                public C() : this((Func<int>)(() => 1))
+                {
+                    Expression<Func<int>> e = () => 2;
+                    _ = new C(e);
+                }
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("C", ".ctor", 2)]));
+
+            Assert.NotEqual(FidelityCheck.CompileBackStatus.RecompileFail, result.Status);
+            Assert.DoesNotContain(": this(", result.Source);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
     public void CompileBackTargets_UsesTargetBackingFieldWriteForConstructorAssignment()
     {
         var assemblyPath = CompileFixture("""
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -567,6 +567,48 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
+    public void CompileBackTargets_DropsInitializerWhenChainedConstructorIsOverloadAmbiguous()
+    {
+        // Issue #2678 guard: the printer can render chain arguments without a
+        // disambiguating cast (a `box` to `object` prints without `(object)`), so
+        // when two same-arity siblings are reconstructed the emitted
+        // `: this(args)` could silently bind to the wrong overload. The guard
+        // requires a unique same-arity sibling; here `C(object)` and `C(int)` are
+        // both present (the latter pulled in by `new C(2)`), so the initializer
+        // must be stripped rather than emit a wrong-binding chain.
+        var assemblyPath = CompileFixture("""
+            public class C
+            {
+                public C(object x)
+                {
+                }
+
+                public C(int x)
+                {
+                }
+
+                public C(string z) : this((object)1)
+                {
+                    _ = new C(2);
+                }
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("C", ".ctor", 2)]));
+
+            Assert.NotEqual(FidelityCheck.CompileBackStatus.RecompileFail, result.Status);
+            Assert.DoesNotContain(": this(", result.Source);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
     public void CompileBackTargets_UsesTargetBackingFieldWriteForConstructorAssignment()
     {
         var assemblyPath = CompileFixture("""
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -523,6 +523,50 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
+    public void CompileBackTargets_DropsInitializerWhenChainedConstructorUnreconstructable()
+    {
+        // Issue #2678 guard: when the chained-to constructor has an unsupported
+        // signature (a function pointer), the planner drops it from the shell. A
+        // same-arity sibling ctor pulled in by another dependency must NOT be
+        // mistaken for the chained-to ctor: emitting `: this(args)` would bind to
+        // the wrong overload and fail with CS1503. The initializer must be
+        // stripped, falling back to an (empty) body that still compiles.
+        var assemblyPath = CompileFixture(
+            """
+            public unsafe class Chained
+            {
+                public Chained(int value) : this((delegate*<void>)value)
+                {
+                    _ = new Chained(true);
+                }
+
+                public Chained(delegate*<void> callback)
+                {
+                }
+
+                public Chained(bool flag)
+                {
+                }
+            }
+            """,
+            allowUnsafe: true);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("Chained", ".ctor", 0,
+                    "(corelib:System.Int32) -> corelib:System.Void")]));
+
+            Assert.NotEqual(FidelityCheck.CompileBackStatus.RecompileFail, result.Status);
+            Assert.DoesNotContain(": this(", result.Source);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
     public void CompileBackTargets_UsesTargetBackingFieldWriteForConstructorAssignment()
     {
         var assemblyPath = CompileFixture("""
--- a/src/ILInspector.Decompiler/Pipeline/Ir/ConsumedMemberEvidence.cs
+++ b/src/ILInspector.Decompiler/Pipeline/Ir/ConsumedMemberEvidence.cs
@@ -44,7 +44,12 @@ public readonly record struct ConsumedMemberEvidence(
         switch (node)
         {
             case Call call:
-                evidence.Add(new(Method: call.Callee));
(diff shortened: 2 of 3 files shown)


richlander:
Reviewed at head 1c12195. **Endorse — safe to merge.** A well-scoped fix with the evidence lane chosen correctly.

**The fix is right.** The printer lifts `: this(...)`/`: base(...)` out of the body into `ConstructorChain` (chain calls are invalid as body statements), and RTS was dropping it on the floor — an empty reconstructed body that recompiled to `ldarg call ret` against an original `ldarg ldarg call call ret`. Threading it through `ProductTargetBody` → `CompileBackMemberRequirement.ConstructorInitializer` → `CSharpConstructorInitializer` re-emission is the minimal correct path, and the two restrictions are both principled and documented in place:

- `this(...)`-only emission (flat object-based shells have no base ctor; `base(args)` would CS1729) keeps the prior behavior for base chains rather than trading OpcodeDiff for RecompileFail.
- The sibling-present guard strips the initializer when the chained-to overload wasn't reconstructed, again preferring the old visible OpcodeDiff over an uncompilable shell.

**Evidence lane verified.** The commit's claim that `ConsumedMemberEvidence` is consumed only by ReturnToSender is true — I grepped: its only consumers are `ReturnToSenderTypePlanner` and its own tests, so decompiler render output is unchanged by construction and the deterministic NuGet.Versioning RTS probe (21 ctor OpcodeDiff → Exact, RecompileFail stable at 63) is the right A/B for this change; no corpus render A/B is owed. The `AllowTargetRoot: call.Callee.Name is ".ctor"` seeding is the correct generalization — a chain call is a construction context exactly like `NewObject`, and the comment says so.

Non-blocking notes:

1. **The sibling-present guard matches by parameter count only** (`member.Parameters.Count == chainCtor.ParameterTypes.Length`). If the chained-to ctor was dropped but a *different* same-arity overload was reconstructed, the initializer survives and either fails to bind (RecompileFail) or binds to the wrong overload via implicit conversion — which surfaces as OpcodeDiff since the call token differs. So the failure mode is visible, never a false `Exact`, which is why this is a note and not a hold — but a type-aware match (or a comment stating arity-match is intentional and why wrongness stays visible) would harden it.
2. **Two documented paths are unpinned**: the base-chain-left-empty behavior and the sibling-missing initializer strip. One fixture each would lock the deliberate conservatism against future "helpful" regressions.
3. `ConstructorChainCallee` mirrors the printer's chain detection by hand; if the printer's shape drifts, the guard misses and the outcome is a visible RecompileFail — fail-visible drift, acceptable, but worth remembering if the printer's chain handling ever changes.

---

richlander:
## Adversarial review — round 1 (both reviewers found the same issue)

Two independent reviewers in isolated worktrees:

| Reviewer | Finding | Severity |
| --- | --- | --- |
| **Gemini 3.1 Pro** | Sibling-present guard matched by arity only | High |
| **GPT-5.5** | Same: arity-only guard binds wrong overload | Medium |

**Agreed finding (verified, valid):** the guard checked only `member.Parameters.Count == chainCtor.ParameterTypes.Length`. When a target chains to a sibling ctor with an unsupported signature (a `delegate*<void>` function-pointer param), the planner drops that ctor; if a *different* same-arity sibling was pulled into the shell by another dependency, the arity check passed and RTS emitted `: this(args)` binding to the wrong overload → **CS1503 → RecompileFail** (a regression from the prior empty-body `OpcodeDiff`). Both reviewers reproduced with a function-pointer fixture.

**Resolution — commit `fc4065dc`:** the chained-to ctor now counts as reconstructed only when **both** (a) a same-arity non-target `Constructor` requirement is present **and** (b) all of the callee's parameter types are supported surface types — checked structurally on the callee's `TypeRef`s via a new `IsUnsupportedChainParameterType` helper that mirrors `MethodRequirement`'s `IsUnsupportedSurfaceSignature` drop (`delegate*`/`<>`/`{`). When the chained-to ctor is not reconstructed, the initializer is stripped and the body falls back to empty. Structural `TypeRef` matching avoids display-string format fragility, so it cannot regress the 21 real wins.

Added `CompileBackTargets_DropsInitializerWhenChainedConstructorUnreconstructable` covering the reviewers' shape. NuGet.Versioning probe unchanged: **21 `OpcodeDiff -> Exact`, zero regressions** (RecompileFail stays 63).

Round-2 re-review of the fixed head requested from both reviewers.

richlander:
Re-reviewed at head fc4065d. **Fix is correct and properly pinned — clear to merge.** Credit to the adversarial reviewers here, and a calibration note on my own first pass: I called the arity-only guard non-blocking on the grounds that wrong-bind stays visible (OpcodeDiff/RecompileFail, never a false Exact). That was true but under-weighted — the reviewers' shape shows the guard could convert the prior *empty-body OpcodeDiff* into a *RecompileFail*, which is a regression, not just a missed improvement. Blocking was the right call.

**The fix.** The guard now requires the chained-to ctor's parameter types to all be supported surface shapes (mirroring the planner's own `IsUnsupportedSurfaceSignature` drop predicate) in addition to a same-arity non-target sibling being present. That's the right indirect formulation given the representation gap — the sibling requirements carry rendered parameter strings while `chainCallee` carries `TypeRef`s, so directly matching them would reintroduce exactly the format fragility the commit avoids. Instead it asks "would the planner have dropped the chained-to ctor?" — and if yes, strips the initializer. The pin reproduces the reviewers' trap precisely: a `delegate*<void>` chained-to ctor dropped from the shell, plus `_ = new Chained(true);` pulling in the same-arity `bool` sibling as bait, asserting both `NotEqual RecompileFail` and no `: this(` in the source. And the NuGet.Versioning probe re-run confirms the strengthened guard cost none of the 21 real wins.

Two notes, neither blocking:

1. **One wording correction for the record:** the commit message says the check is "structural on the callee's TypeRefs," but `IsUnsupportedChainParameterType` matches on `type.ToDisplayString()` substrings — the same display-string heuristic family as `IsUnsupportedSurfaceSignature` it mirrors. That's *consistent* (mirroring a string heuristic with a string heuristic is the faithful choice), just not structural. Worth knowing the next time someone audits text-heuristic surfaces.
2. **The mirror is a second hand-maintained copy of the drop predicate.** If `IsUnsupportedSurfaceSignature` ever gains a new unsupported shape, this mirror misses it and the wrong-bind path returns — visibly (CS1503 RecompileFail), so it degrades loudly, but extracting a shared `ContainsUnsupportedSurfaceShape(string)` core both predicates call would remove the drift risk for one line of code. Same for the residual approximation: the guard still verifies "a same-arity sibling exists," not "the sibling *is* the chained-to ctor" — a supported chained-to ctor dropped for some other reason with a same-arity sibling present would wrong-bind again. `AllowTargetRoot` seeding makes that hard to reach today, and the failure stays visible, so a comment stating the approximation suffices.

---

richlander:
## Adversarial review — round 2

Re-reviewed the hardened head (`fc4065dc`) with both reviewers:

- **GPT-5.5: CLEAN.** Confirmed the round-1 finding resolved (function-pointer shape → OpcodeDiff, no `: this(`), 21 wins intact, RTS 102/102 + evidence 8/8 pass.
- **Gemini 3.1 Pro: new finding (High)** — overload-resolution hijack: the printer can render chain args without a disambiguating cast (a `box` to `object` prints without `(object)`), so with two same-arity siblings in the shell, `: this(args)` can silently bind to the wrong overload.

**Verification (merge-base attribution):** I reproduced the fixture on `fc4065dc` → `STATUS=OpcodeDiff`, `: this(1)` binding to `C(int)`, RECOMP missing `box`. The harness **correctly reports OpcodeDiff — no false Exact**. Root cause is `CoerceText` dropping the `(object)` cast in `CSharpPrinter.Numerics.cs`, which **this PR does not modify** (`git diff 145e7d44 HEAD -- CSharpPrinter.Numerics.cs` is empty). Pre-PR this shape was already OpcodeDiff (empty body, no chain), so it is **not a bucket regression** — the probe shows transitions are exactly `OpcodeDiff -> Exact: 21`, zero `Exact -> OpcodeDiff`.

**Resolution — commit `fdc86627`:** although not a regression, emitting semantically-wrong source is undesirable, so I hardened the guard *within RTS scope*: the chained-to ctor now counts as reconstructed only when **exactly one** same-arity non-target sibling exists (`.Any` → `.Count == 1`). When two same-arity siblings exist, the initializer is stripped and the body falls back to empty. Added `CompileBackTargets_DropsInitializerWhenChainedConstructorIsOverloadAmbiguous`. NuGet probe unchanged (`OpcodeDiff -> Exact: 21`, RecompileFail 63) — no real win depends on ambiguous multi-sibling arity.

**Frontier (follow-up, out of scope):** boxed-arg chains with ambiguous overloads still stay OpcodeDiff (not rescued to Exact) because the printer drops the `(object)` box cast — a pre-existing `CSharpPrinter` limitation that would need its own render A/B to fix.

Round-3 confirmation of `fdc86627` requested from both reviewers.

richlander:
## Round-3 → Round-4 reconciliation (sole-sibling guard)

Round-3 produced **divergent** findings on the arity-based guard:

- **Gemini (High)**: overload resolution can bind a chain-call argument list to a constructor of a **different IL arity** (`params` arrays, optional parameters), so a same-arity check misses cross-arity hijacks.
- **GPT-5.5 (Medium)**: the same-arity uniqueness check **over-strips** valid unambiguous chains.

These pull in opposite directions. Root insight: **arity fundamentally cannot predict overload binding** when the printer renders chain arguments without a disambiguating cast (a `box` to `object` prints without `(object)`) — the drop is in `CSharpPrinter.Numerics.cs` `CoerceText`, which this PR does not touch.

**Resolution (`9e4f805a`):** replace the arity heuristic with a bulletproof rule — emit `: this(args)` **only when the chained-to constructor is the sole reconstructed (non-target) constructor in the shell**. Any second reconstructed ctor of *any* arity → strip the initializer to an empty body.

- Closes **every** cross-arity hijack (Gemini) — a second ctor of any arity forces a strip.
- Never emits a wrong-binding chain. It over-strips some theoretically-safe chains (GPT), but that only ever leaves an **OpcodeDiff** (the pre-PR baseline), never a wrong result or RecompileFail.

**Evidence (unchanged across all guard variants):**

- Deterministic NuGet.Versioning probe: `OpcodeDiff → Exact: 21`, **zero regressions** — every real win chains to the sole ctor in its shell.
- `ReturnToSenderPrototypeTests`: 104 passing (added `CompileBackTargets_DropsInitializerWhenSiblingConstructorCrossesArity` reproducing Gemini's params exploit).
- `ConsumedMemberEvidenceTests`: 8 passing.

Round-4 re-review requested from both reviewers on `9e4f805a`.

richlander:
## Round-4 → Round-5 reconciliation (faithful-args ∪ unique-arity)

Round-4 produced a **convergent** finding — **both reviewers independently** hit the same defect on the sole-sibling head (`9e4f805a`):

- **Gemini (High)** / **GPT-5.5 (Medium)**: the sole-sibling guard counts only *non-target* constructors, but C# overload resolution also considers the **target constructor itself**. When the printer drops a disambiguating cast, `: this((object)text)` prints `: this(text)` and binds back to the target `C(string)` → **CS0516** (constructor cannot call itself) → RecompileFail (worse than the pre-PR OpcodeDiff baseline).

This — together with Round-3's opposing findings (cross-arity hijack vs. over-strip) — confirms that **constructor counting cannot decide overload safety**. Binding correctness depends on whether the **printed arguments are faithful** (no printer-dropped disambiguating cast), not on how many constructors exist. The cast-drop lives in `CSharpPrinter.Numerics.cs` `CoerceText`, which this PR does not modify.

**Resolution (`1128d5c8`):** replace the constructor-count heuristic with a **sound** predicate — emit `: this(args)` only when **either**:

1. **Faithful args** — every argument is a non-lossy expression whose result type structurally equals its parameter type (excludes `Box`/`Coerce`/`Convert`/`IsInstance`/`Constant`, which the printer can spell at a different static type). Faithful args bind to the exact callee regardless of the other constructors — **including the target constructor itself**.
2. **Unique-arity binding** — exactly one constructor in the whole shell (target included) has the chain's argument count. A normal-form exact-arity match beats any `params`/optional expansion, so even a bare `null` binds unambiguously.

Both branches are provably sound: **no wrong binding, no RecompileFail**. This closes the self-referential hijack (CS0516), the box/null-object hijacks, and the cross-arity hijacks, and — addressing GPT's Round-3 over-strip finding — **re-admits unambiguous constant chains** the sole-sibling rule dropped.

**Evidence:**

- Deterministic NuGet.Versioning probe: `OpcodeDiff → Exact: 18`, **zero regressions**, RecompileFail unchanged at 63.
- `ReturnToSenderPrototypeTests`: 105 passing. New: `DropsInitializerWhenChainBindsToTargetConstructor` (self-hijack strip), `PreservesConstructorChainWhenArityIsUnique` (unique-arity constant chain → Exact); retained box-ambiguity and function-pointer strip tests.
- `ConsumedMemberEvidenceTests`: 8 passing.

**Sound conservative miss (not a bug, no regression):** three real NuGet chains pass a `string[]` argument to an `IEnumerable<string>` parameter alongside a same-arity `(Version, string, string)` sibling. `string[]` ≠ `IEnumerable<string>` (not faithful) and the arity is shared (not unique), so they stay at the pre-PR OpcodeDiff baseline. Soundly proving `string[]` cannot bind to the sibling's `string` parameter would require a full type model; the miss costs 3 potential wins (21 → 18) with **zero correctness risk**.

Round-5 confirmation requested from both reviewers on `1128d5c8`.

richlander:
## Round-5 → Round-7 reconciliation — two CLEAN on final head `229436cd`

Both adversarial reviewers now report **CLEAN on the final head `229436cd`**. Rounds 5–7 hardened the two-branch `: this(args)` predicate against two further overload-resolution holes both reviewers helped surface. Reviewers: **GPT-5.5** and **Gemini 3.1 Pro** (I am Claude Opus 4.8, so both reviewers are from other model families per the adversarial-review policy).

### Round 5 — Gemini (High): cross-arity `params`/optional hijack in the unique-arity branch — VERIFIED, FIXED

The unique-arity branch counted competing constructors by **declared** arity (`Parameters.Count == N`). C# overload resolution also considers **cross-arity** candidates via `params`/optional expansion. Reproduced on the prior head: `C() : this((object)null)` with siblings `C(object)` and `C(string, params int[])` — the printer drops the `(object)` cast, printing `: this(null)`; `C(object)` was the only one-parameter ctor so the arity looked unique, but `null` binds to `C(string, params int[])` in expanded form (`string` is more derived than `object`) — the wrong overload (OpcodeDiff here; a CS0121 shape could regress to RecompileFail).

**Fix (`d2a50bd9`):** replaced the declared-arity count with `ConstructorCouldBindToArgCount(member, N)`, which counts a ctor as a competing candidate when it is *applicable* to an N-argument call under `params`/optional expansion (`hasParams ⇒ max = ∞`; `required` = leading non-optional non-`params` parameters; applicable iff `required ≤ N ≤ max`). GPT independently confirmed this head CLEAN; Gemini confirmed its fixture now strips. Gemini also noted the shell strips `[Optional]`, so required-count is exact for the generated shell.

### Round 6 — Gemini (High): typeless-lambda ambiguity in the faithful-args branch — VERIFIED, FIXED

`ChainArgumentIsFaithful` assumed every non-lossy expression prints with a syntax that carries its static type. A **lambda** `() => ...` prints typeless and relies on target-typing, so a matching IR result type does not prove an unambiguous bind. `C() : this((Func<int>)(() => 1))` with a sibling `C(Expression<Func<int>>)` emits `: this(() => 1)`, which both ctors accept → CS0121 → RecompileFail.

Attribution note: Gemini's original fixture body `new C((Expression<Func<int>>)null)` independently prints `new C(null)` (ambiguous) and RecompileFails on `main` too — a pre-existing body-side issue. I isolated the **chain** behavior with a typed-local body (`Expression<Func<int>> e = () => 2; _ = new C(e);`) that prints unambiguously (`new C(Expression.Lambda<Func<int>>(...))`): without the fix that isolated shape newly RecompileFails vs `main`'s chainless OpcodeDiff, confirming a real PR-introduced regression.

**Fix (`229436cd`):** `ChainArgumentIsFaithful` also excludes target-typed nodes lacking an intrinsic static type — `Lambda`, `CollectionExpression`, `TupleExpression`, `InterpolatedStringExpression`. They fall back to the unique-arity branch (or strip). Excluding is strictly safe: it never emits a wrong or uncompilable chain.

### Round 7 — both CLEAN on `229436cd`

- **GPT-5.5 — CLEAN**: forced rebuilds; `ReturnToSenderPrototypeTests` 107/107, `ConsumedMemberEvidenceTests` 8/8; NuGet.Versioning probe `OpcodeDiff → Exact: 18`, RecompileFail `63 → 63`; stackalloc/span and conditional-lambda target-typed probes produced no wrong-binding initializer or new RecompileFail.
- **Gemini 3.1 Pro — CLEAN**: verified the target-typed exclusion set is complete — `default` is covered by the `Constant` exclusion; `DelegateCreation` prints explicitly typed (`new T(M)`), method groups / `AddressOfMethod` map to function-pointer parameters already dropped as unsupported; anonymous objects and target-typed `new()` are printed with explicit shape/type. Ran adversarial fixtures: `this((decimal)1)` prints `decimal.One` (intrinsic type) → correctly preserved Exact; `this((short)1)` lossy → correctly stripped to OpcodeDiff, no CS0121. Both branches sound.

### Final predicate (sound)

Emit `: this(args)` only when the chained-to ctor's parameter types are all supported, a same-arity non-target ctor is present, and **either**:

1. **Faithful args** — every argument is a non-lossy, non-target-typed expression whose result type structurally equals its parameter type (excludes `Box`/`Coerce`/`Convert`/`IsInstance`/`Constant` and the target-typed `Lambda`/`CollectionExpression`/`TupleExpression`/`InterpolatedStringExpression`); **or**
2. **Unique-arity** — exactly one constructor in the whole shell (target included) is *applicable* to an N-argument call under `params`/optional expansion.

### Evidence on `229436cd`

- NuGet.Versioning deterministic probe: `OpcodeDiff → Exact: 18`, **zero regressions**, RecompileFail unchanged at 63.
- `ReturnToSenderPrototypeTests` 107/107, `ConsumedMemberEvidenceTests` 8/8.
- Documented sound conservative miss (no regression): three NuGet chains passing `string[]` to an `IEnumerable<string>` parameter alongside a same-arity `(Version, string, string)` sibling stay at the pre-PR OpcodeDiff baseline (not faithful, not unique-arity).

No dismissed findings this cycle — every reviewer finding across rounds 1–7 was verified and resolved by a commit (`fc4065dc`, `fdc86627`, `9e4f805a`, `1128d5c8`, `d2a50bd9`, `229436cd`). RTS-scoped only; the shared `CSharpPrinter` is unchanged.

richlander:
**Ready to merge.**

- Two CLEAN adversarial reviews on the final head `229436cd`: **GPT-5.5** and **Gemini 3.1 Pro** (both from model families other than my own, Claude Opus 4.8). Every reviewer finding across rounds 1–7 was verified and resolved by a commit.
- CI green on the final head: `build-net10`, `test (ubuntu-26.04, linux-x64)`, `csharp-diff-smoke`, and `changes` all pass; `il-diff-smoke` / `markdownlint` / `pack` correctly skipped (no matching changes).
- NuGet.Versioning deterministic probe: `OpcodeDiff → Exact: 18`, zero regressions, RecompileFail unchanged at 63. `ReturnToSenderPrototypeTests` 107/107, `ConsumedMemberEvidenceTests` 8/8.
- Scope: RTS/DecompilerHarness only; the shared `CSharpPrinter` is unchanged.

I will **not** merge — awaiting @richlander's explicit approval.

richlander:
Re-reviewed at head 229436c — the five gate revisions since my last clearance. **The gate is now genuinely sound; one ask: re-stamp the probe evidence at this head.**

The final formulation closes the residual approximation I noted ("a same-arity sibling exists" ≠ "the sibling *is* the chained-to ctor") properly: emission now requires the drop-predicate pass *and* a present same-arity sibling *and* — the new part — either **faithful arguments** (every argument's printed form carries its parameter's exact type, accounting for the printer dropping disambiguating casts on `Box`/`Coerce`/`Convert`/`IsInstance`/bare constants, and excluding target-typed forms — lambdas, collection expressions, tuples, interpolated strings — whose `ResultType` proves nothing about binding) or **unique applicable candidate**, where applicability correctly models `params` absorption and optional-parameter omission rather than declared arity. That second branch catching the cross-arity `params` sibling that could steal a lossy-argument bind is exactly the kind of case this gate needed. The failure direction remains conservative throughout: can't prove the bind → drop the initializer → the old visible empty-body OpcodeDiff, never a wrong-bind. And the in-place comments documenting *why* each expression class is unfaithful are constraint documentation at its best.

**The ask:** the PR body's evidence table still says "Evidence revision: `1c121952`" — the first head, five gate revisions ago. The gate has strictly tightened since, so the question the probe needs to answer changed: does the faithful-args/unique-arity gate preserve the 21 real `OpcodeDiff → Exact` wins (their arguments must be provably faithful, e.g. `this(Parse(version))` where `Parse` returns the parameter type exactly), or did the stricter proof strip some back to empty-body? A re-run of the deterministic NuGet.Versioning probe at 229436c with the transition counts settles it — same exact-head discipline as the #2702 re-stamp. If the 21 hold and `RecompileFail` stays at 63, this is clear to merge.

---

richlander:
Re-stamped the PR body evidence at head `229436cd` (was `1c121952`, five gate revisions stale). Fresh deterministic NuGet.Versioning probe, exact-head discipline:

```text
BASE: Exact 118, OpcodeDiff 30, RecompileFail 63, not-sampled 98
FIX : Exact 136, OpcodeDiff 12, RecompileFail 63, not-sampled 98

ctor BASE: OpcodeDiff 24, Exact 12, RecompileFail 3, not-sampled 14
ctor FIX : OpcodeDiff  6, Exact 30, RecompileFail 3, not-sampled 14

Transitions: OpcodeDiff -> Exact: 18   (no other transitions)
RecompileFail: 63 -> 63
```

**Direct answer to your ask:** the stricter faithful-args / unique-applicable-candidate gate preserves **18** of the first head's 21 wins and conservatively strips **3** back to their pre-PR OpcodeDiff baseline — it did *not* leave a wrong-bind anywhere, and `RecompileFail` holds at 63.

The 3 it declines are all `SemanticVersion` ctors chaining `this(version, ParseReleaseLabels(releaseLabel), metadata)` where `ParseReleaseLabels` returns `string[]` but the parameter is `IEnumerable<string>`, alongside a same-arity `(Version, string, string)` sibling. That argument is neither faithful (`string[]` != `IEnumerable<string>` structurally) nor unique-applicable (arity shared with the sibling), so the gate drops the initializer. Those rows were `OpcodeDiff` on the base and remain `OpcodeDiff` — **no regression**; the win count moved 21 -> 18 purely because the proof got sound, not because anything got worse. Proving `string[]` cannot bind to the sibling's `string` parameter would require a full type model (out of scope). This is documented in the PR body's "Why 18, not the 21 of the first head" section and the frontiers table.

The wins that *are* preserved are provably faithful — e.g. `NuGetVersion(string) : this(Parse(version))` where `Parse` returns the parameter type exactly — or uniquely applicable by arity.

PR body updated: evidence revision, conclusion (18), tests (107 RTS + 8 CME), remaining-OpcodeDiff list (now 6: the original 2 field-init + 1 `base(...)` flat-shell, plus the 3 sound misses), and the review table (both CLEAN). Ready to merge — still awaiting your explicit approval to actually merge.

richlander:
## Independent final-head restamp

Reproduced the deterministic `NuGet.Versioning` probe in isolated base/head worktrees (`145e7d44` → `229436cd`). Method identities match exactly.

```text
BASE: Exact 118, OpcodeDiff 30, RecompileFail 63, not-sampled 98
HEAD: Exact 136, OpcodeDiff 12, RecompileFail 63, not-sampled 98

Transitions: OpcodeDiff -> Exact 18
All other statuses unchanged; zero regressions.

Ctor BASE: Exact 12, OpcodeDiff 24, RecompileFail 3, not-sampled 14
Ctor HEAD: Exact 30, OpcodeDiff 6,  RecompileFail 3, not-sampled 14
```

Exact-head tests also pass: `ReturnToSenderPrototypeTests` 107/107 and `ConsumedMemberEvidenceTests` 8/8. This resolves the latest evidence-restamp ask; the current PR body correctly names `229436cd` and explains why the hardened gate retains 18 rather than the first head's 21 wins.

**One closure concern:** `Closes #2678` does not satisfy #2678's current acceptance criterion that the deterministic probe report no constructor `Exact -> OpcodeDiff` rows. Six remain:

- two field-initializer lifts (`SimplePool`, `StringBuilderPool`);
- one `base(...)` chain (`NuGetVersion(Version, IEnumerable<string>, string, string)`);
- three conservative `string[] -> IEnumerable<string>` chain misses in `SemanticVersion`.

The code/evidence are mergeable, but either change the linkage to `Advances #2678` and leave the issue open, or file explicit follow-ups for all six rows and revise #2678's closure contract before merging.

richlander:
Re-verified at head 229436c. **The probe re-stamp landed with exactly the analysis the ask was for — clear to merge.** The "Why 18, not 21" section is model evidence writing: the stricter gate preserves the 18 provably-faithful/unique-arity wins, and the 3 `SemanticVersion` chains it declines (`string[]` argument to an `IEnumerable` parameter beside a same-arity sibling) stay at their **pre-PR OpcodeDiff baseline** — a documented sound miss, not a regression — with the honest note that proving that bind safe needs a full type model. `RecompileFail` stable at 63, no other transitions, remaining ctor OpcodeDiffs enumerated and attributed. Both reviewers CLEAN at this exact head. Nothing outstanding.

---

richlander (commented):
## Independent review — VERDICT: READY (minor residuals, all fail-visible)

The fix is sound, conservative, and correctly scoped. The overload-safety gate survives the attacks I ran; failures are fail-visible (`RecompileFail` / stays-`OpcodeDiff`), never a false `Exact`.

**Verified:**
- **RTS-only / product render unchanged** — `AllowTargetRoot`/`EffectiveAllowTargetRoot` and `ConsumedMemberEvidence.AddFrom` have exactly one non-test consumer (`ReturnToSenderTypePlanner`). No product render path reads it.
- **`this(...)`-only, base left empty** — initializer is set only when `constructorChain.StartsWith("this(")`, off the printer's own `callee.DeclaringType == function.DeclaringType` classification (not a text heuristic), so `base(...)` → string starts `base(` → stays null → CS1729 unreachable.
- **Faithful-args branch is sound** — identity-conversion arguments make the chained-to ctor the unique best overload; target-typed (`Lambda`/`CollectionExpression`/`TupleExpression`/`InterpolatedStringExpression`) and static-type-shiftable (`Box`/`Coerce`/`Convert`/`IsInstance`/`Constant`) families are excluded; `ConstructorCouldBindToArgCount` reads the same `Modifier`/`HasDefault` fields the shell renders. No path emits on uncertainty.

**Two low residuals (worth a code comment; neither is a false-`Exact`):**
1. `ReturnToSenderTypePlanner.cs` ~1215 — the gate requires "a same-arity non-target sibling present", which is not identical to "the chained-to ctor present". If the chained-to ctor were dropped for a reason *not* captured by `IsUnsupportedChainParameterType` while a faithful-args sibling exists, the args wouldn't match → CS1503 `RecompileFail` (empty-body→RecompileFail direction). Mitigated because `AllowTargetRoot` seeding guarantees the chained-to ctor is reconstructed.
2. ~1244 — `chainedConstructorReconstructed` is computed *before* the `includeRecordSurface`/`typedEqualsSibling` additions, so a record-surface-injected same-arity ctor isn't counted as a competitor → theoretical false-unique. Niche (needs an explicit `this`-chain inside record-surface reconstruction), worth a comment.

**Process nit:** `Closes #2678` may overclaim — 6 ctor `OpcodeDiff` rows remain at head (2 field-init lifts, 1 flat-shell `base(...)`, 3 `string[]→IEnumerable<string>` sound misses). If #2678's bar is "zero ctor `OpcodeDiff`", consider `Advances #2678` or revise the issue's contract.

`mergeable_state: clean`; low staleness risk (only `CSharpPrinter.cs` moved on main since base, via #2702; the IR-based gate is robust to its qualifier-text changes).

---
*Independent review by Claude Code — static (diff + source at head `229436c`; no local build/test/harness run, so the "18 wins / 0 regressions" probe is unverified here — though a reviewer comment reproduced it in isolated worktrees).*

---

PR #2709, opened 2026-07-13
Harden render-manifest attribution and expose degraded signatures

richlander:
## Summary

- attribute rendered tables only to logical level-2 Markout sections, honoring heading offsets and nested headings
- classify field-layout tables from the active product schema instead of the `Field` display-name heuristic
- surface guarded signature fallback as `Decode: degraded` across ApiMember-based type/member views and JSON
- disambiguate degraded placeholders from genuine `object` signatures in API diff and stable member identity

Follow-up to #2698.

## Validation

- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release --no-build` — 452 passed
- `dotnet run --project src/ILInspector.CSharp.Tests -c Release --no-build` — 90 passed
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-build` — 1,819 passed, 10 environment skips
- NativeAOT linux-x64 publish
- `npx markdownlint-cli docs/architecture.md docs/design/schema-query.md`

## Adversarial review

- First pass: Gemini 3.1 Pro was CLEAN. Claude Opus 4.8 found that degraded `object` placeholders could collide with genuine signatures in diff/anchor identity and that `MetadataDeclarationQuery` discarded decode status.
- Resolved both findings in 183b0ed7, including nested/cyclic TypeSpec propagation, ordinary nested-generic stability, attribute-only diff matching, and focused regressions.
- Exact-head re-review of 183b0ed7: Claude Opus 4.8 CLEAN; Gemini 3.1 Pro CLEAN. No outstanding findings or non-actions.


richlander:
Ready to merge. CI is green, local build/tests/NativeAOT validation passed, and Claude Opus 4.8 plus Gemini 3.1 Pro both returned CLEAN on exact head 183b0ed7.

richlander:
Reviewed at head 183b0ed. **Endorse — safe to merge.** This one PR closes four items from my open ledger, each done the way I'd have asked:

**1. Degraded decode is now fail-visible on the API surface** (the #2490 follow-up from the #2700 review). `DegradedTypeNode` replaces the bare `NamedTypeNode("object")` fallback — renders identically (including correct nullability-byte consumption for `object?`) but carries `IsDegraded`, propagated through every composite node: generic arguments, arrays, pointers, byref, function pointers, modified types, passthrough. That flows into `ApiMember.SignatureDecodeStatus` and renders as a `[MarkoutSkipNull]` `Decode: degraded` cell — affected rows only, so clean tables gain no schema noise. The null-means-clean contract is documented with the back-compat rationale for older serialized surfaces. The architecture.md sentence — "fail-closed... but not presented as ordinary metadata" — is exactly the doctrine statement this convention needed.

**2. The degraded-anchor collision is closed** (concern 5 from #2700: a degraded signature reading `object M()` could mint an anchor colliding with a genuine `object M()`). `ComputeFingerprint(signature, isDegraded)` prefixes `D:` before hashing, so fingerprint and stable selector diverge while the canonical signature text stays shared — the right split, since the grammar is still the grammar and the fingerprint is where identity lives. The diff analyzer's `MemberKey(Signature, DecodeStatus)` makes degraded-vs-genuine surface as `MemberSignatureChanged` with an explicit `[decode: degraded]` marker instead of silently pairing — and the attribute-only scope exemption (decode status excluded from the key when signatures aren't compared) is a thoughtful touch, pinned by `DegradedSignature_DoesNotPreventAttributeOnlyComparison`.

**3. Both #2698 follow-ups landed.** Heading attribution is now level-aware (`_sectionHeadingLevel` from the writer options' offset; nested subheadings can't steal tables — pinned by the `Method Details` null-columns assertion), and `IsFieldTable`'s first-column-named-"Field" heuristic is replaced by the active schema's `field` ItemKind — schema-driven, with pins in both directions: renamed headers (`Property`/`Contents`) still record fields, and a table with a coincidental `Field` column can't create them. The name heuristic that survived the text-heuristic-retirement PR is now retired too.

**4. The containment workers got stronger than #2700 left them.** They now assert not just no-crash but `SignatureDecodeStatus.Degraded` end-to-end through both `ApiSurfaceExtractor` and `MetadataDeclarationQuery`, and a method-signature worker joins the field and cycle workers. That converts the visibility feature itself into a pinned contract — a regression back to silent `object` substitution fails the suite.

Checked and fine: the non-`Result` `GuardedSignatureText` lanes remain because TypeSourceComposer and EcosystemIntegrationScanner have no member-status surface to thread degradation into — justified coexistence, not accretion; migrate them if those paths ever grow visibility. `MemberKey`'s default record equality is ordinal-equivalent to the old `StringComparer.Ordinal` dictionaries.

Two notes, neither blocking:

1. **Cross-epoch diff skew**: a surface serialized before this PR carries null status; re-inspecting the same malformed assembly now yields `Degraded`, so a cached-old vs fresh-new diff reports a spurious `MemberSignatureChanged`. Only reachable for malformed assemblies in cached-surface scenarios — rare — but worth one sentence in schema-query.md or the release notes so the first person to hit it recognizes it.
2. The `Decode` column presumably participates in `--columns`/discovery via the manifest now that capture is schema-aware — if it's meant to be selectable, a one-line schema registration check is worth doing; if it's presentation-only, current state is fine.

---

Issue #2710, opened 2026-07-13
Harden Instructions type-name graph climbs against cyclic metadata


richlander:
## Summary

While closing #2490's remaining signature-decoder recursion surface, the
`ILInspector.Instructions` census found a separate native-stack recursion
mechanism in `IlAssemblyDiff`:

- `IlAssemblyDiff.TypeName` recursively follows TypeDef declaring types.
- `SignatureIdentityProvider.TypeDefinitionName` does the same.
- `SignatureIdentityProvider.GetTypeFromReference` recursively follows TypeRef
  resolution scopes.

Malformed metadata can make those graphs cyclic or pathologically deep, causing
an uncatchable `StackOverflowException`. `IlBodyDiff` already bounds equivalent
name climbs, but `IlAssemblyDiff` does not.

This is intentionally separate from #2490: signature-blob structural recursion
and metadata graph recursion require different guards and fixtures.

## Acceptance

- Replace or bound every recursive TypeDef declaring-type and TypeRef
  resolution-scope climb in `ILInspector.Instructions`.
- Add process-isolated cyclic/deep graph fixtures through public
  `IlAssemblyDiff.CompareStreams` / `CompareMembers` paths.
- Add an anti-ratchet census for equivalent recursive metadata graph climbs.
- Preserve valid-assembly diff identities and CoreLib output.

PR #2711, opened 2026-07-13
Harden Instructions signature decoding against deep blobs

### LINKED ISSUE #2490: Harden all signature-decode entry points against over-deep blobs (SRM native-stack StackOverflow beyond TypeRefDecoder)

richlander:
## Summary

Closes #2490's final signature-decoder exposure by routing every
`ILInspector.Instructions` provider decode through a shared structural prescan
and every nested TypeSpec re-entry through `TypeSpecGuard`.

- unsafe method/calli signatures fail unresolved instead of fabricating stack effects
- rejected diff identities remain stable across metadata-token shifts
- the provider-decode AST census now covers `ILInspector.Instructions`
- process-isolated fixtures cover deep method, field, member-ref, local,
  method-spec, standalone/calli, and cyclic TypeSpec paths

The separate cyclic type-name graph mechanism found during the census is tracked
in #2710.

## Validation

- `ILInspector.Instructions.Tests`: 116 passed
- signature safety + provider census: 23 passed
- CoreLib return-address A/B: normalized snapshots byte-identical
  (36,846 matched; 8,980 agree; 27,866 diverge; 7,015 unmatched)

## Adversarial review

Exact head `7a463017` received two clean isolated-checkout reviews:

- Claude Opus 4.8: **CLEAN**
- Gemini 3.1 Pro: **CLEAN**

Both verified the complete decode census, guard-kind mapping, balanced TypeSpec
re-entry, fail-unresolved stack behavior, stable rejected identities, duplicate
row handling, process-isolated fixtures, and SRM/NativeAOT layering. Neither
reported an actionable finding, so no review-resolution commit applies.

Closes #2490

### DIFF
--- a/src/ILInspector.Instructions.Tests/SignatureDecoderSafetyTests.cs
+++ b/src/ILInspector.Instructions.Tests/SignatureDecoderSafetyTests.cs
@@ -1,0 +1,409 @@
+using System.Diagnostics;
+using System.Reflection;
+using System.Reflection.Metadata;
+using System.Reflection.Metadata.Ecma335;
+using System.Reflection.PortableExecutable;
+
+namespace ILInspector.Instructions.Tests;
+
+public class SignatureDecoderSafetyTests
+{
+    const string WorkerVariable = "DOTNET_INSPECT_INSTRUCTIONS_SIGNATURE_WORKER";
+
+    [Fact]
+    public void DeepAndCyclicSignatures_ThroughAssemblyDiff_AreContainedInChildProcess()
+        => RunWorker(nameof(AssemblyDiffWorker));
+
+    [Fact]
+    public void DeepAndCyclicSignatures_ThroughStackResolver_AreContainedInChildProcess()
+        => RunWorker(nameof(StackResolverWorker));
+
+    [Fact]
+    public void AssemblyDiffWorker()
+    {
+        if (!IsSelectedWorker(nameof(AssemblyDiffWorker)))
+            return;
+
+        var image = BuildAdversarialImage();
+        using var oldStream = new MemoryStream(image);
+        using var newStream = new MemoryStream(image);
+
+        var result = IlAssemblyDiff.CompareStreams(
+            oldStream,
+            "old.dll",
+            newStream,
+            "new.dll");
+
+        Assert.True(result.Diff.ComparedBodyCount > 0);
+        Assert.Equal(0, result.Diff.FailureCount);
+
+        using var oneOverload = new MemoryStream(BuildRejectedOverloadImage(methodCount: 1));
+        using var twoOverloads = new MemoryStream(BuildRejectedOverloadImage(methodCount: 2));
+        var overloadResult = IlAssemblyDiff.CompareStreams(
+            oneOverload,
+            "one.dll",
+            twoOverloads,
+            "two.dll");
+
+        Assert.Equal(1, overloadResult.Diff.FailureCount);
+
+        var unshifted = BuildShiftedGuardedMethodImage(addPrefixMethod: false);
+        var shifted = BuildShiftedGuardedMethodImage(addPrefixMethod: true);
+        using var unshiftedStream = new MemoryStream(unshifted);
+        using var shiftedStream = new MemoryStream(shifted);
+        var shiftedResult = IlAssemblyDiff.CompareStreams(
+            unshiftedStream,
+            "unshifted.dll",
+            shiftedStream,
+            "shifted.dll");
+
+        Assert.Equal(1, shiftedResult.Diff.FailureCount);
+        Assert.Equal(1, shiftedResult.Diff.PairExactCount);
+
+        using var oldPe = new PEReader(new MemoryStream(unshifted));
+        using var newPe = new PEReader(new MemoryStream(shifted));
+        var oldReader = oldPe.GetMetadataReader();
+        var newReader = newPe.GetMetadataReader();
+        var memberResult = IlAssemblyDiff.CompareMembers(
+            oldPe,
+            oldReader,
+            FindMethod(oldReader, "Caller"),
+            newPe,
+            newReader,
+            FindMethod(newReader, "Caller"));
+
+        Assert.True(memberResult.Diff.IsExact);
+    }
+
+    [Fact]
+    public void StackResolverWorker()
+    {
+        if (!IsSelectedWorker(nameof(StackResolverWorker)))
+            return;
+
+        var image = BuildAdversarialImage();
+        using var pe = new PEReader(new MemoryStream(image));
+        var reader = pe.GetMetadataReader();
+        var entryHandle = FindMethod(reader, "Entry");
+        var entry = reader.GetMethodDefinition(entryHandle);
+        var body = pe.GetMethodBody(entry.RelativeVirtualAddress);
+        var resolver = MetadataStackTypeResolver.Create(reader, entryHandle, body);
+
+        Assert.Equal(StackType.Unknown, resolver.Field(MetadataTokens.GetToken(MetadataTokens.FieldDefinitionHandle(1))));
+        Assert.Equal(StackType.Unknown, resolver.Field(MetadataTokens.GetToken(MetadataTokens.MemberReferenceHandle(2))));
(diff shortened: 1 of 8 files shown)


richlander:

Exact head `7a463017` was reviewed in separate detached worktrees:

- **Claude Opus 4.8 — CLEAN**
- **Gemini 3.1 Pro — CLEAN**

The reviewers agreed that every Instructions `Decode*Signature` entry is bound
to the correct `SignatureBlobGuard.Kind`, all nested TypeSpec re-entry is
balanced through `TypeSpecGuard`, unsafe stack-effect signatures fail
unresolved, rejected identities remain stable across token shifts, and the
process-isolated fixtures cover the claimed surface. There was no divergence
and no dismissed guidance.

Earlier implementation review findings were resolved in this head: explicit
stack-signature rejection, stable/lazy rejected-signature digests, O(n)
duplicate-row correlation, member-ref field coverage, and concurrent child
stdout/stderr draining. The exact-head reviewers found no remaining actionable
guidance, so no separate resolution commit applies.

richlander:
Ready to merge

richlander:
Reviewed at head 7a46301. **Endorse — safe to merge.** This closes the #2490 census exactly as scoped, and the identity design for rejected signatures is better than the recipe I wrote.

**Coverage is complete.** All sixteen sites from the #2490 slice-C census are routed: both unguarded nested TypeSpec re-entries (`StackTypeSignatureProvider.GetTypeFromSpecification`, `IlAssemblyDiff`'s `SignatureIdentityProvider`) now go through `TryTypeSpec` with `TypeSpecGuard` bracketing; the nine `IlBodyDiff` format sites, `IlAssemblyDiff.MethodKey`, and all seven `MetadataStackTypeResolver` decodes get prescan gateways. `ProviderSignatureDecodeBoundaryTests.AssemblyRoots` gains `ILInspector.Instructions`, so closure is now mechanically enforced — the census passing *is* the proof, which beats my grep. The new gateway is a justified sibling rather than a copy-paste: Instructions needs `Try*` forms (reject-detection) alongside fallback-value forms, plus `StandaloneMethod`/`LocalVariables` shapes Metadata's gateway doesn't have, and the csproj gains only MetadataPrimitives (where `SignatureBlobGuard`/`TypeSpecGuard` live) — layering respected.

**The `RejectedIdentity` design is the standout.** Identity lanes fall back to `<unsupported-signature:{SHA256(blob)}>` — content-derived, not positional and not a shared sentinel. That gets three things right at once:
- distinct malformed signatures get distinct identities (no degraded-collision class — the problem #2709 had to solve with fingerprint disambiguation is avoided here by construction);
- the same malformed blob pairs with itself across versions, so self-diff of an adversarial assembly is exact rather than spuriously changed (pinned: `FailureCount == 0` on self-diff);
- identity is independent of metadata row position — pinned by the shifted-image test, where inserting a `Dummy` method shifts every row and the guarded method still pairs `Exact` and `CompareMembers` on the caller stays `IsExact`. That's the row-renumber discipline from the source-Finding producers applied to degraded identities.

The `MethodMapKey(Identity, Occurrence)` change also fixes a latent silent drop I hadn't flagged: `MethodMap` previously `TryAdd`'d first-wins, so a second method with an identical rendered key silently vanished from the diff. Now occurrences are disambiguated, deterministically ordered, and an unmatched occurrence surfaces as a failure — pinned by the one-vs-two rejected-overloads test asserting `FailureCount == 1` instead of silence.

**Fallback semantics are honest per lane.** Identity/rendering lanes show the visible sentinel in diff text; stack-lattice lanes fall back to `StackType.Unknown` — the lattice's native "don't know" — and the interpreter surfaces it as `IsComplete == false` with "Unresolved Call signature" (pinned); `TryResolveCall`/`TryResolveCalli` return false rather than guessing; and `MetadataStackTypeResolver.Create` throws a typed `BadImageFormatException` for an undecodable root signature rather than fabricating an argument frame. Both workers are process-isolated per doctrine, and the adversarial image is thorough — deep field/method/member-ref/method-spec/local/calli blobs plus a cyclic TypeSpec, all exercised through one entry body.

Two residual notes for #2490's closeout, neither blocking this PR:

1. The census now covers the three SRM-only assemblies (Metadata, MetadataPrimitives, Instructions). ILInspector.Analysis and the Decompiler are guarded — I verified every site at main during the #2490 census — but their guard shapes (if-statement + depth/cumulative counters in the two `TypeRefDecoder`s) don't match the census's sanctioned ternary form, so they're protected today but not anti-ratcheted. Either teach the census the if-guard shape someday or record on #2490 that those assemblies rely on review rather than the gate.
2. Cosmetic: a stray blank line added before the closing brace in `IlBodyDiff.cs`, and `MethodMap`'s switch from first-wins `TryAdd` to occurrence keys is a (more honest) behavior change for well-formed assemblies carrying duplicate rendered keys — worth one line in the commit lore if IL-diff counts ever shift on a real corpus.

With this merged, #2490's acceptance criteria are met — every `Decode*Signature` call site in the product is either gateway-guarded or census-enforced, with process-isolated containment pins in all three guarded assemblies. The issue can close.

---

PR #2714, opened 2026-07-13
Fix --rows window validation leaking a stack trace (follow-up to #2704)

richlander:
## Summary

Follow-up fix to #2704 (merged as d1d6222c). That PR shipped a real user-facing
regression that adversarial review caught **after** the squash-merge: the
`--rows` head/tail window validation threw `RowWindowValidationException` from
`SharedOptions.BuildRowWindow`, which runs inside the command `SetAction` during
`System.CommandLine`'s `InvokeAsync`. SCL's invocation pipeline catches that
throw and prints a raw **"Unhandled exception" stack trace** (exit 1) instead of
the intended clean `Error: ...` one-liner. The `catch` in `Program.cs` was
therefore dead code.

### Before (on current `main`)

```console
$ dotnet-inspect type System.String -S "Member Index" --rows --head=3 --tail 2
Unhandled exception: DotnetInspector.Output.RowWindowValidationException: --rows cannot combine -n/--head with --tail; choose one row window.
   at DotnetInspector.Services.SharedOptions.BuildRowWindow(...)
   ...
   at System.CommandLine.Invocation.InvocationPipeline.InvokeAsync(...)
```

### After

```console
$ dotnet-inspect type System.String -S "Member Index" --rows --head=3 --tail 2
Error: --rows cannot combine -n/--head with --tail; choose one row window.   # stderr, exit 1, empty stdout, no stack trace

$ dotnet-inspect type System.String -S "Member Index" --rows
Error: --rows requires -n/--head N or --tail N.                              # stderr, exit 1
```

## Fix

Validate the `--rows` head/tail window at **parse time** via a command validator
in `SharedOptions.AddOutputOptionsTo` (`command.Validators.Add(result => ...
result.AddError(...))`). It reads the real `System.CommandLine` parse values
(Rows/Limit/Tail), so it covers `=`-syntax (`--head=3`) and concatenated forms
(`-n3`) that the arg-preprocessor token scan misses. `Program.cs` already checks
`result.Errors` after `Parse` and prints clean `Error: ...` lines before
`InvokeAsync`, so no code path change is needed there.

`BuildRowWindow`'s throws remain as a defensive, unit-tested invariant guard (the
validator gates the CLI path first); the `Program.cs` catch is likewise
defensive.

Test-harness fidelity: `RunAppAsync` now mirrors `Program.cs` by surfacing
`result.Errors` as clean `Error: ...` stderr lines before `InvokeAsync`, so the
e2e grammar tests reflect the real CLI contract instead of SCL's usage-help dump
(this is why the original #2704 e2e tests passed falsely — the stack-trace text
happened to contain the message string).

## Validation

- `dotnet build src/dotnet-inspect -c Release` — succeeds.
- Full test suite (`dotnet run --project src/dotnet-inspect.Tests -c Release`):
  **1852 total, 0 failed, 10 skipped**.
- Manual CLI checks across forms: `--head=3 --tail 2`, `-n3 --tail 2`,
  `-3 --tail 2` (all rejected, exit 1, clean error); `--tail=2`, `-n 2` (valid,
  exit 0); cross-command probes (`type`, `member`, `cache`, `timeline`, `diff`,
  …) all emit the clean parse-time error with no stack trace.

## Adversarial review

Both **GPT-5.5** and **Gemini 3.1 Pro** independently found the swallowed-exception
bug on the pre-fix head, and both re-reviewed this fix at head `0eb0f233`
(cherry-picked here as `d49466b5`) and returned **CLEAN** with real-run evidence,
including confirmation that `Rows`/`Tail` options are only ever added through
`AddOutputOptionsTo` (so no command has the options without the validator).

### DIFF
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -1406,6 +1406,9 @@ public class CommandExecutionTests
         Assert.Equal(1, exit);
         Assert.Empty(output);
         Assert.Contains("--rows cannot combine -n/--head with --tail", error);
+        // Pin the failure shape, not just the message: a swallowed BuildRowWindow
+        // throw would dump a stack trace that also contains the message and exits 1.
+        Assert.DoesNotContain("Unhandled exception", error);
     }
 
     [Fact]
@@ -1417,6 +1420,8 @@ public class CommandExecutionTests
         Assert.Equal(1, exit);
         Assert.Empty(output);
         Assert.Contains("--rows requires -n/--head N or --tail N", error);
+        // Pin the failure shape, not just the message (see above).
+        Assert.DoesNotContain("Unhandled exception", error);
     }
 
     [Fact]
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -399,13 +399,28 @@ public class CommandExecutionTests
         {
             args = CommandLineBuilder.PreprocessArgs(args);
             var root = CommandLineBuilder.CreateRootCommand();
+            var result = root.Parse(args);
+            // Mirror Program.cs: surface parse/validation errors (including the --rows
+            // head/tail window validator) as a clean "Error: ..." line on stderr with
+            // exit 1, instead of letting InvokeAsync print usage help.
+            if (result.Errors.Count > 0)
+            {
+                foreach (var error in result.Errors)
+                {
+                    var message = error.Message.StartsWith("Error:", StringComparison.OrdinalIgnoreCase)
+                        ? error.Message
+                        : $"Error: {error.Message}";
+                    Console.Error.WriteLine(message);
+                }
+                return 1;
+            }
             try
             {
-                return await root.Parse(args).InvokeAsync();
+                return await result.InvokeAsync();
             }
             catch (RowWindowValidationException ex)
             {
-                // Mirror Program.cs so --rows window validation surfaces as exit 1 + stderr.
+                // Defensive: matches the Program.cs safety-net catch.
                 Console.Error.WriteLine($"Error: {ex.Message}");
                 return 1;
             }
--- a/src/dotnet-inspect/Program.cs
+++ b/src/dotnet-inspect/Program.cs
@@ -114,10 +114,11 @@ if (showTraceMermaid && args.Length > 0 && argsBeforePreprocess.FirstOrDefault()
     RequestTelemetry.Breadcrumb("preprocess", $"{string.Join(' ', argsBeforePreprocess)} -> {string.Join(' ', args)}");
 
 // Install line-limiting writer when -NN shorthand was used (e.g. -30).
-// With --rows, -n/-NN is interpreted by commands as per-table row limits, and the
-// head/tail conflict is validated by SharedOptions.BuildRowWindow against the real
-// System.CommandLine parse (see the RowWindowValidationException catch below), not
-// the token scan here — the scan misses =-syntax and concatenated forms.
+// With --rows, -n/-NN is interpreted by commands as per-table row limits. The
+// head/tail conflict is validated at parse time by the command validator in
+// SharedOptions.AddOutputOptionsTo against the real System.CommandLine parse (so it
+// covers =-syntax and concatenated forms the arg-preprocessor token scan misses),
+// which is why no --rows gate remains here.
 var rowLimitMode = args.Any(a => a == "--rows");
 // Line/tail windows apply only outside --rows mode; in --rows mode the count is a
 // per-table data-row window rendered by the commands, not an output-line window.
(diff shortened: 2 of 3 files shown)


richlander:
Reviewed at head d49466b. **The fix is correct and at the right layer — one one-line ask, then merge.** And an accounting note first: this regression shipped through my #2704 clearance. I verified the `catch (RowWindowValidationException)` existed in Program.cs and took it as evidence of the UX, without checking that SCL's invocation pipeline intercepts handler exceptions *before* they can reach that catch — making it dead code — and I accepted e2e pins whose assertions (`Contains(message)` + exit 1) were satisfiable by the broken stack-trace output. Two review lessons I'm banking: a catch around `InvokeAsync` is not evidence of the error contract, and error-path assertions must reject the failure *shape*, not just contain the success text.

**The fix.** Parse-time command validator in `AddOutputOptionsTo` reading the real parse values (`=`-syntax and concatenated forms covered, same as before), surfacing through `result.AddError` → the pre-existing Program.cs parse-error path, which I verified prints clean `Error:` lines and exits before `InvokeAsync` ever runs. Both defensive layers (the `BuildRowWindow` throws, the Program.cs catch) are retained with comments correctly demoting them to safety nets. Since `Rows`/`Tail` are only ever added via `AddOutputOptionsTo`, every command gets the validator — which also quietly fixes the wrinkle I noted on #2704: `cache --rows` errors at parse time again instead of running silently. And the `RunAppAsync` harness change fixes the false-pass mechanism at its root by mirroring the real CLI's parse-error contract.

**The ask: pin the failure shape.** `Rows_RejectsBothHeadAndTailWithEqualsSyntax` and `Rows_RequiresHeadOrTailWindow` still assert only `Contains(message)` + exit 1 + empty stdout — exactly the assertions that passed against the broken output. If the validator were ever removed, `BuildRowWindow`'s defensive throw inside `InvokeAsync` would reproduce the stack-trace dump, and both tests would pass again (the dump contains the message text and exits 1). One line in each — `Assert.DoesNotContain("Unhandled exception", error)` — closes the loophole, and the suite already uses that exact idiom (`Type_PlatformPrefixBrowse_AllMissProjection_ReportsCleanError`). Without it, the bug class this PR exists to fix remains assertable-around.

Credit to both adversarial reviewers for catching this post-merge — the "found on pre-fix head independently by both models, re-reviewed CLEAN at head with real-run evidence" loop is the process working as designed.

---

richlander:
Addressed in `f8d5aa71`: added `Assert.DoesNotContain("Unhandled exception", error)` to both `Rows_RejectsBothHeadAndTailWithEqualsSyntax` and `Rows_RequiresHeadOrTailWindow`, so the tests now reject the failure *shape* (the stack-trace dump) rather than just containing the success message. Verified both still pass; the assertion would fail against the old swallowed-exception output (which begins with `Unhandled exception:`).

Thanks for the accounting note — banking the same two lessons: a `catch` around `InvokeAsync` isn't evidence of the error contract, and error-path assertions must reject the failure shape, not just contain the success text.

richlander:
Re-verified at head f8d5aa7. **The failure-shape pins landed** — both error-path tests now carry `Assert.DoesNotContain("Unhandled exception", error)` with a comment stating exactly why the message-only assertion was insufficient. The false-pass loophole is closed. **Clear to merge.**

---

richlander (commented):
## Independent review — VERDICT: READY (with a merge-coordination flag vs #2721)

The core fix is correct and at the right layer.

- **Validator logic** (`SharedOptions.cs`): `if (!GetValue(Rows)) return;` then rejects both-head-and-tail ("cannot combine") and neither ("requires"), accepts exactly one. Not inverted, no off-by-one — an exact behavioral mirror of `BuildRowWindow`. Because it reads the real SCL `ParseResult`, the `=`-syntax (`--head=3`) and concatenated (`-n3`) forms are covered.
- **Coverage claim holds**: `Options.Add(Rows|Tail|Limit)` appears only inside `AddOutputOptionsTo`, and all call sites register the validator alongside the options — so no command can have Rows/Tail without the validator.
- **Dead-catch reasoning holds**: `Program.cs` returns on `result.Errors` before `InvokeAsync`, so the invalid combo never reaches the action; `BuildRowWindow`'s throw remains a defensive invariant and the `catch (RowWindowValidationException)` is genuinely dead for the CLI path (not masking a live path).
- **Harness fix is right**: `RunAppAsync` surfacing `result.Errors` before `InvokeAsync`, plus the new `DoesNotContain("Unhandled exception")` asserts, closes the "#2704 tests passed falsely" loophole.

**Minor:** the harness formats errors inline (`StartsWith("Error:") ? msg : $"Error: {msg}"`) rather than reusing `Program.cs`'s `FormatParseError`, which additionally reshapes SCL's `Cannot parse argument` messages. Identical for this PR's `--rows` messages, but a future `-n abc`-style e2e test through `RunAppAsync` would observe the raw SCL text and could mask a `FormatParseError` regression. Cheap fix: reuse the same helper.

### ⚠️ Merge coordination with #2721
Both PRs insert a validator at the same spot in `AddOutputOptionsTo`. #2721 adds a `bool supportsRowWindows = true` param and, for `cache`, a "--rows not supported" rejection. Guaranteed textual conflict; and a mechanical keep-both yields **two contradictory errors on bare `cache --rows`**. Resolve by gating this PR's head/tail window validator behind `if (supportsRowWindows)` so an opted-out command gets only #2721's rejection.

---
*Independent review by Claude Code — static (diff + source at head `f8d5aa7`; no local build/test run).*

---

PR #2715, opened 2026-07-13
Make signature limit degradation visible

richlander:
## Summary

- carry guarded-decode degradation through unsafe pointer detection instead of collapsing it to `HasUnsafeCode = false`
- fail signature spellability closed and expose `SignatureSpellabilityResult.DecodeStatus`
- render unresolved unsafe signature presence as a `Decode degraded` diagnostic
- preserve pointer and required-modifier evidence through custom-modifier combinators

## Evidence

- process-isolated 100,000-level pointer-bearing and spellability fixtures report `Degraded`
- real TypeSpec-backed `modopt(pointer)` and external-internal `modreq` fixtures pin modifier behavior
- ordinary pointer and accessible/inaccessible signatures retain their existing results with no decode marker
- Metadata tests: 468 passed
- CLI tests: 1,844 passed, 10 tool-dependent skips
- linux-x64 NativeAOT publish succeeded
- changed Markdown passes markdownlint

Follow-up to #2700 and #2709.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -98,7 +98,10 @@ Extract public API surface using metadata:
 Guarded metadata signature rejection remains fail-closed (`object`/empty
 signature shape), but it is not presented as ordinary metadata:
 `ApiMember.SignatureDecodeStatus` is `Degraded`, and member tables add a
-`Decode: degraded` cell only for affected rows.
+`Decode: degraded` cell only for affected rows. The same status contract covers
+non-rendering signature inspections: spellability rejects a degraded signature
+instead of treating it as spellable, and unsafe-member discovery emits a
+diagnostic row when its pointer-signature scan is incomplete.
 
 ### diff
 
--- a/src/ILInspector.Metadata/AssemblyDetailScanner.cs
+++ b/src/ILInspector.Metadata/AssemblyDetailScanner.cs
@@ -471,10 +471,15 @@ public static class AssemblyDetailScanner
                             var detection = PointerDetection.Combine(
                                 decoded.Value.ReturnType,
                                 decoded.Value.ParameterTypes);
-                            if (decoded.IsDegraded || detection.IsDegraded)
-                                flags.UnsafeSignatureDecodeStatus = SignatureDecodeStatus.Degraded;
                             if (detection.HasPointer)
+                            {
                                 flags.HasUnsafeCode = true;
+                                flags.UnsafeSignatureDecodeStatus = null;
+                            }
+                            else if (decoded.IsDegraded || detection.IsDegraded)
+                            {
+                                flags.UnsafeSignatureDecodeStatus = SignatureDecodeStatus.Degraded;
+                            }
                         }
                         catch (Exception ex) when (ex is BadImageFormatException or InvalidOperationException or ArgumentException)
                         {
--- a/src/ILInspector.Metadata/AssemblyDetailScanner.cs
+++ b/src/ILInspector.Metadata/AssemblyDetailScanner.cs
@@ -462,12 +462,24 @@ public static class AssemblyDetailScanner
                     {
                         try
                         {
-                            var sig = GuardedProviderDecode.Method(reader, method, PointerDetector.Instance, (object?)null, false);
-                            if (sig.ReturnType || sig.ParameterTypes.Any(p => p))
+                            var decoded = GuardedProviderDecode.MethodResult(
+                                reader,
+                                method,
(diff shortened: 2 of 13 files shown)


richlander:
## Adversarial review reconciliation — Ready to merge

Exact head: `1af95079`

Two isolated cross-model reviews are **CLEAN on the exact head**:

- **Claude Opus 4.8:** CLEAN; verified all top-level/nested degradation propagation, modifier semantics, aggregate scanner ordering, JSON/Markdown visibility, concurrency, SRM/NativeAOT constraints, Metadata 468/468, and focused CLI coverage.
- **Gemini 3.1 Pro:** CLEAN; reproduced both modifier fixtures, traced every provider combinator, and verified degraded/pointer ordering semantics and output visibility.

### Round-one findings

Gemini identified two actionable modifier near misses on `acf00832`:

1. `PointerDetector.GetModifiedType` dropped pointer evidence in a custom modifier. Fixed in [`1af95079`](https://github.com/richlander/dotnet-inspect/commit/1af95079): pointer evidence now combines modifier and unmodified types, pinned by a TypeSpec-backed `modopt(pointer)` fixture.
2. Spellability dropped inaccessible required-modifier evidence. Fixed in [`1af95079`](https://github.com/richlander/dotnet-inspect/commit/1af95079): inaccessible modifier evidence propagates for `modreq`, pinned by an external-internal modifier fixture.

Gemini also questioned skipping later signature decodes after a definite pointer. No exhaustive-scan change applies: this is an aggregate presence query, so once `HasUnsafeCode` is definitively true, later signatures cannot make presence uncertain. The follow-up clarifies that contract and clears aggregate degradation when a pointer establishes presence. Both reviewers verified this resolution as truthful in round two.

CI is green (`test`, `build-net10`, `markdownlint`, `changes`); local solution build, Metadata 468/468, CLI 1,844 passed with 10 tool-dependent skips, and linux-x64 NativeAOT publish are clean.

**Ready to merge.**

richlander:
Reviewed at head 1af9507. **Endorse — safe to merge.** This completes the fail-visible sweep from the #2700 review: the two remaining silent lanes (an over-deep pointer-bearing signature could *hide* `HasUnsafeCode`; spellability failed open to "spellable") are now honest, and the pattern from #2709 is applied consistently rather than reinvented.

**HasUnsafeCode.** `PointerDetector` now carries `(HasPointer, IsDegraded)` through every combinator, and the scanner's semantics are exactly right: a definite pointer answers the presence question (marker cleared), degraded-without-pointer sets `UnsafeSignatureDecodeStatus = Degraded`, and I verified the state is order-independent — the `if (!flags.HasUnsafeCode)` guard means no later degraded method can re-set the marker after a definite answer, and the explicit `= null` handles the degraded-then-pointer order. The old blanket `catch { }` (a silent-skip I'd flagged) is narrowed to typed decode exceptions that record degradation. The product surface is a visible Diagnostic row in Unsafe Members ("unsafe-code presence may be incomplete") with `CanRender` extended so the section appears even when the member list is empty — pinned at both the view and section-pipeline layers.

**Spellability.** `InspectField/Property/Method` return `SignatureSpellabilityResult(CanSpell, DecodeStatus)`; both the guard-rejection path and the decode-exception path now yield `CanSpell: false` + `Degraded` — the fail-open-to-"spellable" behavior is dead, and shell planning consuming the retained bool wrappers now drops degraded signatures instead of trusting them.

**Two genuine semantics fixes ride along, both pinned:** `GetModifiedType` previously ignored the modifier entirely, so (a) a pointer inside a custom modifier didn't count toward `HasUnsafeCode` — now it does (`PointerInCustomModifier_IsDetectedWithoutDegradation`), and (b) an *inaccessible required* modifier didn't make a signature unspellable — now it does, correctly scoped to `isRequired` only since modopt can be dropped at recompile (`InspectMethod_RejectsInaccessibleRequiredModifier`, with a synthesized modreq to a hidden external type). Both are the right answers, and the near-miss tests (valid pointer, pointer-in-modifier → detected with *no* degradation marker) pin the boundary from the other side.

**Test upgrades follow the #2711 standard:** the process-isolated workers now assert the degraded status end-to-end rather than just no-crash, so the visibility feature itself is a pinned contract.

One structural note, not an ask: this adds a third degradation-tracking shape to the gateways (`GuardedSignatureText.DecodeResult`, now `GuardedProviderDecode.DecodeResult`, plus Instructions' `Try*` pattern). All three are justified by their distinct consumer needs, but if a fourth appears, it's worth consolidating the result shape.

---

Issue #2059, opened 2026-07-01
Performance Triage: add static metadata for dynamic trace correlation


richlander:
## Summary

`runfaster` can now read raw `.nettrace` GC allocation ticks and join allocation stacks back to `dotnet-inspect` Performance Triage rows. The next precision blocker is on the static side: triage rows need richer machine-readable metadata so a dynamic correlator can move from **method+shape hot** to **specific row likely hot**.

This is a static-output request for Performance Triage, not a request to put dynamic trace parsing into `dotnet-inspect`.

## Why

Current triage JSONL rows are enough for a human:

- `member`
- `shape`
- `evidence`
- `fix`
- `confidence`
- `loop`
- `il`

They are not quite enough for a correlator. With raw `.nettrace`, `runfaster` can observe allocation ticks like:

- allocated type (`System.String`, `System.Text.StringBuilder`, `System.Char[]`, `System.Numerics.BigInteger`, `Func<>`, arrays, etc.)
- sampled bytes/count
- managed stack frames

But when one method has several static rows of the same shape, dynamic evidence still lands at “this method has matching allocation types under it” rather than “this exact row is the likely source.”

Examples from the prototype:

- Humanizer `ArabicNumberToWordsConverter.Convert(...)`: 9 `string-build-in-loop` rows in one hot method. `.nettrace` confirms 187,627 matching string-build allocation ticks / 18.63 GB sampled under the method, but cannot distinguish which IL rows contributed how much.
- Newtonsoft.Json `JsonWriter.WriteValue(JsonWriter, PrimitiveTypeCode, object)`: one `box System.Numerics.BigInteger` row at `IL_03BF`. `.nettrace` sees 10 matching `System.Numerics.BigInteger` allocation ticks / 1.02 MB sampled under the method, plus much larger method-level string/StringBuilder allocation context. This is much stronger because the static row has a specific expected allocated type.

## Requested static metadata

Add machine-readable fields to Performance Triage JSONL / JSON rows where possible:

### Expected allocation identity

Fields that let a dynamic trace match allocation ticks to static rows without parsing prose:

- `expectedAllocatedTypes`: array of runtime type names, e.g.
  - `System.Numerics.BigInteger` for `box System.Numerics.BigInteger`
  - `System.String`, `System.Text.StringBuilder`, `System.Char[]` for `string-build-in-loop`
  - `T[]` / concrete array type for `small-array` / `newarr`
[...]

PR #2716, opened 2026-07-13
Retain Finding provenance in Performance Triage

richlander:
## Summary

- retain exact `analysis.allocation` / `analysis.call-site` provenance on Performance Triage rows
- add compact `Candidate` IDs plus machine-readable `Finding`, `Operation`, and `Token` fields across library/type/member output
- make the fields filterable/sortable and document runtime/static versus cross-version join semantics

This intentionally does not pull correctness-only or PDB/network-bound Finding families into performance ranking. Aggregate and cross-method rows keep a candidate ID without claiming an exact source Finding.

Partial progress on #2059.

## Evidence

- `ILInspector.Analysis.Tests`: 427 passed
- `dotnet-inspect.Tests`: 1811 passed, 10 skipped
- tool-assembly census: 1,238 rows / 1,238 unique Candidate IDs; 1,205 allocation-backed, 26 call-site-backed, 7 intentionally aggregate/cross-method
- interleaved CPU measurements showed no material Performance Triage regression versus `origin/main`
- markdownlint clean on all changed Markdown

## Adversarial review

Two isolated review passes completed against commit `3594e919`:

- Claude Opus 4.8: CLEAN
- Gemini 3.1 Pro: CLEAN

Both checked producer/shape attribution, candidate stability and collision risk, aggregate-row behavior, output/query consistency, performance, SRM/NativeAOT boundaries, and focused real command output. No follow-up commit was required.


richlander:

Reviewed the exact validated head `3594e919` in separate clean worktrees.

- **Claude Opus 4.8 — CLEAN.** Verified allocation-vs-call-site attribution at shared IL instruction classes, candidate identity inputs and within-build scope, all output/query projections, aggregate-row null provenance, CoreLib timing parity, and SRM/NativeAOT compatibility.
- **Gemini 3.1 Pro — CLEAN.** Independently checked shape-to-producer attribution on real output, candidate uniqueness and cross-scope stability, Markdown/TSV/JSONL plus library/type/member consistency, aggregate/cross-method behavior, memoization/performance, API compatibility, and the new tests.

The reviewers agreed that exact rows preserve the correct producer while `allocation-hotspot` / `scan-method-in-loop-call` rows do not overclaim exact provenance. They found no conflicting conclusions and no actionable findings. Gemini observed two Analysis tests that also fail on the parent under its preview-SDK environment; those were verified as pre-existing and unrelated to this PR, so no resolution commit applies. No review follow-up commit was required.

richlander:
Ready to merge. Local validation, two clean adversarial reviews, and all merge-blocking CI checks are complete.

richlander:
Reviewed at head 3594e91. **Doctrinally sound — the design got the hard boundaries right — but one fix round: the documented behaviors need pins, and the ID namespace should be self-owned before these become user-visible currency.**

**What's right, and it's the part that matters:** triage is not promoted to a Finding. The `Finding` field is the producer's `FindingDescriptor.Id` verbatim; producer identity is fingerprinted into the Candidate ID rather than re-rendered onto the row; `Token`/`IL` are documented as version-local provenance with cross-version correspondence explicitly delegated back to `diff --finding`/`timeline --finding` — the join-semantics doc ("`Candidate` is exact within one assembly build… IL offsets and metadata tokens remain provenance rather than correspondence identity") is exactly the right statement. The scope claim checks out too: only allocation/call-site families are consumed; correctness and PDB/network-bound families stay out of ranking. And finding-adoption.md rule 9 is a faithful generalization of the boundary.

**Ask 1 — pin what the docs assert.** Three documented behaviors have zero coverage: (a) **aggregate rows** — SKILL.md and both design docs state that `allocation-hotspot`/`scan-method-in-loop-call` rows carry a `pt~` id with empty `Finding`/`Operation`/`Token`, and no test exercises it; (b) the **type/member output path** — only the library JSONL path is tested, and `OptimizationOpportunityRow` is a positional record, so a column transposition in `ApiOutputFormatter` or the markout view would compile and ship silently; (c) **ID stability** — determinism is implied by the inputs but never asserted (open the index twice, compare). The 1,238/1,238 uniqueness claim is empirical PR-body evidence with no structural guarantee — a 40-bit truncated hash with no collision handling; at minimum an assert-unique in the builder (or lengthen-on-collision) turns the prose claim into a property.

**Ask 2 — give the fingerprint a self-owned namespace.** `CandidateId` hashes under `MemberAnchor.FingerprintPrefix = "dotnet-inspect.member-index.v1\n"`, passing a non-signature composite into a parameter named `canonicalSignature`. That couples every `pt~` id to a foreign versioning contract: if member-index rotates to v2, all triage ids silently change for an unrelated reason. A `performance-triage.v1` prefix is a two-line change now and a breaking change later.

Non-blocking, worth a line each:
1. The allocation-over-call-site tie-break (an offset in both censuses always attributes to `analysis.allocation`) is correct for enumerator allocations but undocumented — one comment.
2. `FirstOrDefault` on the offset join silently picks the lower-identity occurrence if two ever share an offset — latent today (one instruction, one occurrence); an invariant comment or `Single` would make the assumption checkable.
3. `--where "Token=0x2000001"` won't match the zero-padded render (`0x02000001`) — doc the 8-digit form or normalize in the filter.
4. Aggregate vs provenance-missing are distinguishable only by the incidental absence of `IL` — if an exact row's offset can ever miss both censuses, the docs' "empty Finding ⇒ aggregate" equation breaks; either state the invariant or add an explicit marker.
5. `Candidate` is advertised as sortable but sorts in hash order — semantically vacuous; fine to keep for stable pagination, worth a doc word so nobody reads meaning into it.

---

richlander:
## Feedback resolution and refreshed adversarial review

Resolved the owner feedback in [`3c9ad412`](https://github.com/richlander/dotnet-inspect/commit/3c9ad412):

- moved `pt~` hashing to the self-owned `dotnet-inspect.performance-triage.v1` namespace, included the complete `FindingKey`, added unambiguous length-prefixed encoding, expanded the initial digest to 16 hex characters, and deterministically lengthen truncated-prefix collisions;
- replaced first-match producer joins with a checked single-occurrence invariant and documented allocation precedence at shared offsets;
- added explicit `exact` / `aggregate` / `unmatched` provenance across library, type, member, JSONL, TSV, Markdown, filtering, sorting, docs, and the performance skill;
- normalized exact metadata-token predicates and documented Candidate hash ordering as stable pagination only;
- pinned aggregate rows, type/member positional projections, candidate determinism across index openings, and candidate uniqueness in tests.

Refreshed adversarial review used separate isolated clones at the exact fixed head:

- **Claude Opus 4.8 — CLEAN.** It exercised the full .NET 9 shared-framework set (169 assemblies, 5,514 triage rows), confirmed candidate uniqueness/determinism, exact/aggregate invariants, token predicate behavior, cross-scope/output consistency, and SRM/NativeAOT constraints.
- **Gemini 3.1 Pro — CLEAN.** It independently verified the identity namespace/canonical inputs and collision path, checked joins and allocation precedence, traced every output projection, and ran the affected Analysis and CLI test suites.

The reviewers agreed on every attack point and reported no actionable or conflicting findings. No findings were dismissed and no additional resolution commit applies. Refreshed merge-blocking CI is green.

richlander:
Ready to merge. Owner feedback is resolved at `3c9ad412`; two refreshed adversarial reviews are CLEAN and all merge-blocking CI checks pass.

richlander:
Re-reviewed at head 3c9ad41. **Both asks landed and the fix round went well past them — clear to merge.**

- **Self-owned namespace**: `dotnet-inspect.performance-triage.v1`, decoupled from the member-index fingerprint contract. Done as asked.
- **Uniqueness is now structural, not empirical**: 16-hex IDs, collision-checked at index build with deterministic lengthening on a truncated-prefix collision — the prose claim became a property, and `CandidateIdsAreStableAndUnique` pins stability (open twice, compare) plus distinctness.
- **The aggregate/provenance-missing conflation I called non-blocking got a real fix**: an explicit `Provenance=exact|aggregate|unmatched` field, so an instruction-level row that fails its census join is distinguishable from a deliberate aggregate — pinned by the aggregate-provenance theory over both aggregate shapes.
- The token filter normalization (`Token=0x2000001` now matching the zero-padded render) closes non-blocking item 3 too.

Nothing outstanding.

---

richlander (commented):
## Independent review — VERDICT: MINOR-CONCERNS (effectively mergeable)

Functionally correct and well-tested; the two flagged items are advisory, low severity.

**Verified:**
- **Candidate ID uniqueness** is a structural guarantee, not just empirical: `AttachFindingProvenance` enforces a global `HashSet<string>` with deterministic lengthen-or-throw. ID = `pt~` + 16-hex SHA-256 prefix over length-prefixed (namespace, descriptor+`FindingKey`, method token+IL offset, ordinal, shape).
- **Stability / no cross-run drift** — deterministic iteration order; IDs assigned once on the whole-assembly index and merely filtered for type/member scope, so cross-scope IDs are identical. Cross-*version* IDs change by design (documented; delegated to `diff --finding`/`timeline --finding`).
- **Provenance correctness** — join is by IL offset (a unique instruction) so exact-by-construction; allocation-over-call-site precedence at shared offsets is enforced; aggregate/cross-method rows correctly carry `Provenance=aggregate` with null `Finding`/`Operation`/`Token`; `unmatched` correctly distinguishes a failed instruction-level join from a deliberate aggregate.
- **Query consistency** — all new fields (+`Weight`) are in both `FilterableFields` and `SortableFields` and wired through the triage field/compare/predicate paths; JSONL omits nulls consistently.
- **Verbosity contract OK** — Performance Triage is scanner-opt-in (`IncludeOpportunities`), not a default `-v:m` section, so the always-on columns aren't default-view noise.

**Advisory (low severity):**
1. The candidate-exhaustion `throw` (`LibraryBodyIndex.cs` ~158-167, and `SingleFindingAtOffset` ~186-189) converts a producer-side duplicate-row data anomaly into an **assembly-wide exception** for the whole `OptimizationOpportunities` property — which also backs Top Leverage. Practically unreachable (IL offset is unique; aggregates dedup per token), but a defensive skip/dedup would be safer than converting a data anomaly into a section-wide failure.
2. Column-ordering taste: the machine-join columns (`Candidate`/`Finding`/`Provenance`) are inserted ahead of the human-relevant `Shape`/`Evidence`, widening an already-wide table.

`mergeable_state: clean`. Neither item blocks merge.

---
*Independent review by Claude Code — static (diff + source at head `3c9ad41`; no local build/test run, so the "1,238 unique IDs" census and no-regression perf claim are unverified here).*

---

richlander:
## Follow-up review reconciliation

Reviewed the later `MINOR-CONCERNS` advisory and the subsequent owner re-review at `3c9ad412`, which concluded **clear to merge / nothing outstanding**.

No code change applies to either advisory:

1. **Invariant failures:** truncated candidate-prefix collisions are already handled by deterministic lengthening. The remaining throws represent a duplicate full candidate identity or multiple producer occurrences at one MethodDef+IL offset—violations of the producer/index contract. Silently skipping or deduplicating those rows would hide corrupt provenance and undo the requested checked-single/structural-uniqueness guarantees, so the explicit failure is intentional.
2. **Column order:** `Candidate` / `Finding` / `Provenance` are intentionally adjacent to `Member` as the row's machine identity/provenance block, consistently across library/type/member and Markdown/TSV/JSONL. Reordering is a taste tradeoff rather than a correctness or compatibility fix and would churn the newly pinned positional contract.

No findings were dismissed as incorrect; both were considered valid design observations but non-actionable for this PR. No resolution commit applies. PR remains ready to merge.

PR #2717, opened 2026-07-13
Add Finding-backed implementation and authored-source evidence

richlander:
Advances #2673

- Adopts native C# and IL Finding comparisons through Research and product
  `ImplementationDiff`.
- Adds checksum-gated authored-source rebuild fidelity beside the independent
  decompiler lane.
- Adds opt-in authored `Source` as a peer product mechanism.
- Evidence revision: `f8ebc0b2`

## Change

> Should we accept these implementation-evidence lanes?

**Conclusion:** **PASS** — the product and harness now retain producer-owned
C#, IL, and authored Source evidence without blending checksum integrity,
decompiler fidelity, authored rebuild fidelity, determinism, or build-context
drift.

Before:

```text
ImplementationDiff reconstructed C#/IL rows after discarding native Finding
envelopes; authored SourceLink bytes could prove integrity but could not be
rebuilt or compared as product implementation evidence.
```

After:

```text
DiffFixtureSample.DiffSample.ConstantValue()
  IL      changed  ldc.i4 1 -> ldc.i4 2
  C#      changed  return 1 -> return 2
  Source  removed  public static int ConstantValue() => 1;
  Source  added    public static int ConstantValue() => 2;
```

## Scope and safety

- **Finding adoption:** Research retains native `csharp.line` and `il.op`
  comparisons, descriptor-gates them, and preserves one-sided, bodiless,
  `Absent`, and `Failed` outcomes.
- **Authored acquisition:** `DotnetInspector.Services` fetches and caches exact
  bytes, verifies SHA-1/SHA-256 portable-PDB checksums before decoding, and
  returns token-scoped member/document Finding envelopes.
- **Harness oracle:** authored A→IL substitutes only the target body into the
  same final RTS `ArtifactRequest`; the independent decompiled B→IL verdict is
  unchanged.
- **Product impact:** `diff -S "Implementation Diff" --authored-source` adds
  checksum-verified authored `Source` beside decompiled `CSharp` and shipped
  `IL`. Network acquisition is explicit; without the flag, existing C#/IL-only
  output is unchanged.
- **Boundary:** Services and Research remain network/Roslyn separated; Roslyn
  rebuild and body extraction stay in `DecompilerHarness`; inspected
  assemblies use SRM rather than runtime loading.

## Evidence

| Check | Result |
| --- | --- |
| Release solution build | Pass; 3 pre-existing Metadata nullable warnings |
| Product suite | 1,865 total; 1,855 passed, 10 expected tool skips |
| Services suite | 259 passed |
| Research suite | 106 passed |
| Authored rebuild focus | 16 passed |
| Product diff focus | 286 passed |
| Decompiler fast suite | 2,403 passed |
| Markdownlint | Pass |
| Real product Source witness | `ConstantValue`: 2 authored Source rows beside 3 decompiled C# and 2 IL rows |

## Oracle separation

Real SourceLink-backed run over
`DiffFixtureSample.ConstructorSample::get_Value`:

| Evidence lane | Verdict |
| --- | --- |
| PDB checksum integrity | `Exact` |
| Decompiled B→IL | `Exact` |
| Authored A→IL | `IlDifferent` |
| Deterministic build | `False` |
| Build context | `Drift` — compiler version differs from the current RTS compiler |

The authored mismatch is therefore reported as build/compiler drift, not a
decompiler defect and not an integrity failure.

## Compile-back quality

> Did this add a second oracle without weakening the first?

**Conclusion:** **PASS** — the authored body reuses the final typed RTS request
while keeping its verdict and context labels separate from decompiler
compile-back. This PR does not change importer, raise, typing, or printer
behavior, so render-text A/B and a decompiler quality-diff card are not
applicable.

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| Nondeterministic or compiler-drifted A→IL | Explicit context verdict | Not clean evidence of authored-source fidelity |
| Missing SourceLink coverage | `Absent` | Missing evidence is not a failure |
| Source acquisition/checksum error | `Failed` | Never rendered as a semantic change |
| Assembly-wide source-only census | Not added | Product enrichment is bounded to members already changed by C#/IL |

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | Clean on `f8ebc0b2` | Verified the whole PR, real SRM generic-interface names, 24 identity probes, cache repair, boundaries, and targeted tests |
| Gemini 3.1 Pro | Clean on `f8ebc0b2` | Verified constructed-generic, alias, ambiguity, cache, comparison, and lifetime adversarial cases |

Earlier rounds found and resolved:

1. Valid-Base64 cache poisoning could persist after checksum mismatch;
   `6aa335b9` now bypasses invalid memory/disk bytes, repairs from valid network
   bytes, and never caches an invalid network payload.
2. Source caches were not concurrency-safe; `deae7342` uses
   `ConcurrentDictionary`.
3. Authored extraction could miss or select neighboring explicit-interface
   declarations; `d66a637e` searches all members and preserves qualifiers.
4. Generic, qualified, alias, and same-name interface identities needed ranked
   matching; `3672851a` and `f8ebc0b2` add full-identity scoring, conservative
   ambiguity rejection, CLR generic-name normalization, and adversarial tests.

## Validation

```bash
dotnet run --project src/DotnetInspector.Services.Tests -c Release --no-build
dotnet run --project src/ILInspector.Research.Tests -c Release --no-build
  -trait- "Speed=Slow"
npx markdownlint-cli README.md docs/design/implementation-diff.md \
  docs/design/source-finding-producers.md \
  docs/design/fact-planned-compile-back-harness.md \
  skills/compatibility/SKILL.md

  artifacts/bin/DiffFixtures.V1/release/DiffFixtureSample.dll \
  --authored-rebuild-fidelity --cap 1 --max-examples 3

dotnet artifacts/bin/dotnet-inspect/release/dotnet-inspect.dll diff \
  --library artifacts/bin/DiffFixtures.V1/release/DiffFixtureSample.dll..artifacts/bin/DiffFixtures.V2/release/DiffFixtureSample.dll \
  -S "Implementation Diff" --authored-source \
  -t DiffSample -m ConstantValue
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ context for copied DLLs. A future `--deps` source can represent runtime
 | Project skills | `project` | Direct dependency `Skills` rows from package `skills/**/SKILL.md` files, plus version-resolved package README/PROJECT docs from restored projects. |
 | Library audit | `library` | Assembly identity, public key token, trim/AOT metadata, unsafe/interoperability signals, OpenTelemetry support, symbols/PDBs, SourceLink and determinism audit, references, resources, async method classification. |
 | API discovery | `type`, `member`, `find` | Type search, member tables, docs, overload selection, generics, obsolete-member markers, direct calls and callers, source/decompiled/IL drill-in. Add `--project` to resolve type/member queries in the project's restored dependency context. |
-| API compatibility | `diff` | Version ranges, package or platform diffs, breaking/additive/potentially-breaking classification, type and member filters. |
+| API compatibility | `diff` | Version ranges, package or platform diffs, breaking/additive/potentially-breaking classification, type and member filters, plus opt-in decompiled C#/IL/checksum-verified authored Source evidence. |
 | Relationships | `depends`, `extensions`, `implements` | Type hierarchies, package dependencies, library reference graphs, extension methods/properties, implementors and subclasses. Add `--project` to search project-referenced packages. |
 | Source mapping | `type`/`library`/`package -S "Source Files"`, `member -S "Source Locations"` / `"Original Source"` | SourceLink URLs, member file/line locations, source fetching, URL verification, token+IL-offset to source-line resolution. |
 | Performance analysis *(experimental)* | `library`/`type`/`member -S "Top Leverage"`, `"Performance Triage"`, `"Call Graph"`, `"Caller Graph"` | Whole-assembly call-graph leverage ranking — direct callers, root reach, fanout, depth, loop calls — with opt-in per-node cost signals (alloc, copy, unsafe, reflection, throw/exception, catch/finally) and actionable rewrite-shape detection. |
@@ -127,7 +127,7 @@ context for copied DLLs. A future `--deps` source can represent runtime
 | `type X` | Discover types or render a single type shape. |
 | `member X` | Inspect members, docs, overloads, decompiled/lowered C#, SourceLink-backed original source, and IL. |
 | `find X` | Search for types across packages, frameworks, projects, and local assets. |
-| `diff X` | Compare API surfaces by default; opt into analysis or C# + IL implementation evidence. |
+| `diff X` | Compare API surfaces by default; opt into analysis or peer decompiled C#, IL, and checksum-verified authored Source implementation evidence. |
 | `extensions X` | Find extension methods and C# extension properties for a type. |
 | `implements X` | Find concrete implementors or subclasses. |
 | `depends X` | Walk type, package, or library dependency graphs; emits Mermaid diagrams. |
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ dotnet-inspect library System.Text.Json --il-offset 0x06000004+0x15
 dotnet-inspect diff --package System.Text.Json@9.0.0..10.0.0 --breaking
 dotnet-inspect timeline --package System.Text.Json@8.0.0..9.0.0 --type System.Text.Json.JsonSerializer --members --at all
 dotnet-inspect timeline --package MyLib@1.0.0..2.0.0 --type MyType --member Parse --finding analysis.unsafety --at all
-dotnet-inspect diff --library old/Foo.dll..new/Foo.dll -S "Implementation Diff" -m MyType.HotPath
+dotnet-inspect diff --library old/Foo.dll..new/Foo.dll -S "Implementation Diff" --authored-source -m MyType.HotPath
 dotnet-inspect depends Stream --markdown --mermaid
 dotnet-inspect implements IEquatable --project ./src/App -v:q
 dotnet-inspect extensions string --project ./src/App -v:n
--- a/docs/design/fact-planned-compile-back-harness.md
+++ b/docs/design/fact-planned-compile-back-harness.md
@@ -24,6 +24,20 @@ The important distinction is artifact production vs proof:
 - ReturnToSender owns artifact requests, compilation, product diff invocation,
   and failure reporting.
 
+The harness also has a strict two-oracle invariant:
+
+- decompiled **B→IL** answers whether product-decompiled C# rebuilds to shipped
+  IL;
+- checksum-verified authored **A→IL** measures authored-source rebuild fidelity;
+- neither verdict changes, rescues, or declassifies the other.
+
+PDB checksum verification proves that fetched authored bytes correspond to the
+PDB. It is not the A→IL verdict. Determinism, compilation options, references,
+compiler version, generators, and project context are reported separately as
+build-context evidence. `--authored-rebuild-fidelity` substitutes the acquired
+authored body into the same final `ArtifactRequest`, invokes product IL diff,
+and prints both oracle results plus checksum, determinism, and context status.
+
 The existing `CB_CLUSTER=1` path already has a strong generic answer for closure
 membership: compile, read the compiler's missing-symbol diagnostics, add the
 named same-assembly roots, and repeat until the closure stops growing or hits a
--- a/docs/design/implementation-diff.md
+++ b/docs/design/implementation-diff.md
@@ -28,7 +28,8 @@ family.
   Source comparisons retain `FindingComparison<string>` with the `text.line`
   descriptor. Research
   cross-checks their exactness against the richer semantic projections for
-  members present on both sides.
+  members present on both sides. A disagreement is retained as a per-member
+  `Failed` diagnostic; it does not abort healthy members in the same diff.
 
 ## Research comparison model
 
@@ -38,7 +39,7 @@ type-centric groups from that collection; grouped and flat consumers therefore
 cannot observe divergent copies of the same result.
 
 Each `ResearchChange` carries one mechanism, a `FindingDescriptor`, an
-added/removed/changed classification, its subject, and any native producer
+added/removed/changed/failed classification, its subject, and any native producer
 payload needed for typed presentation. It is deliberately not a
 `PairFinding<T>`. Metadata now exposes genuine API type/member comparisons and
 `ResearchComparison.ApiComparison` retains that producer-owned envelope. C#,
@@ -50,6 +51,24 @@ their semantic rows remain because they carry richer producer-owned evidence,
 while retained comparisons expose the exact census transitions. `Source` never
 replaces or changes the meaning of `CSharp`: one describes checksum-verified
 authored text and the other describes product-decompiled text.
+
+### Deliberate dual-representation decision
+
+This design revises the earlier plan to retire the C# and IL semantic
+projections immediately after Finding adoption. The two retained
+representations have different durable payloads:
+
+- native Finding comparisons own inspection outcomes, stable census identity,
+  and added/removed/present/changed transitions;
+- `CSharpBodyDiff` and `IlBodyDiff` own aligned hunks, typed display failures,
+  old/new offsets, and richer producer-formatted evidence.
+
+The semantic projections therefore remain deliberately rather than by
+accretion. Every overlapping member is cross-validated against the native
+comparison, and divergence becomes a visible per-member `Failed` diagnostic.
+If the Finding producers later carry equivalent aligned hunk and typed display
+payloads, the semantic projections should be deleted rather than matched a
+third time.
 
 ## Consumer contract
 
@@ -69,6 +88,10 @@ Use `CompareMembersWithAuthoredSource` when the caller also has old/new
 `WithAuthoredSourceComparisons` to enrich an assembly comparison. These APIs
 preserve `Complete`, `Absent`, and `Failed` independently and retain the native
 line comparison. Research does not fetch source.
+
+Finding acquisition and cross-validation failures use
+`ResearchChangeKind.Failed`; they are operational diagnostics, never semantic
+`Changed` rows in table, TSV, JSONL, or programmatic consumers.
 
 The `diff --finding csharp.line` and `diff --finding il.op` focused lenses read
 those retained comparisons and render native `PairFinding` cases. Missing
@@ -97,6 +120,10 @@ endpoint PDB and
 SourceLink body, verifies the document checksum, and adds a separately labeled
 `Source` lane. Missing mappings and acquisition failures remain visible rather
 than falling back to decompiled C#.
+The authored A→IL lane reuses the final RTS shell/request but compiles with
+portable-PDB-recorded options when available; the decompiled B→IL lane uses the
+RTS compile context. `BuildContext` and determinism verdicts therefore remain
+part of interpreting any Exact/IlDifferent disagreement.
 Package, platform, and local-library ranges use the same acquisition path as the
 default API diff; `--type`, `--member`, row limits, table, TSV, and JSONL
 projection continue to apply. The CLI consumes this product component and does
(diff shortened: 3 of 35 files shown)


richlander:
Adversarial review reconciliation for exact head `f8ebc0b2`:

- **Gemini 3.1 Pro:** initially found persistent valid-Base64 cache poisoning,
  unsafe source caches, and authored-member identity gaps. Resolved by
  `6aa335b9`, `deae7342`, `d66a637e`, `3672851a`, and `f8ebc0b2`. Final
  whole-PR review is clean after adversarial constructed-generic, alias,
  neighboring-member, ambiguity, cache-repair, comparison, and lifetime
  probes.
- **Claude Opus 4.8:** independently reviewed the whole PR, verified real SRM
  explicit-interface metadata names, ran 24 authored-identity probes, and
  rechecked cache repair, concurrency, Finding retention, envelope states,
  reader lifetimes, and product/harness boundaries. Final exact-head review is
  clean.

Two Gemini findings were dismissed rather than changed:

- Normal property getter extraction was already present in the continuation of
  the reviewed switch; later commits nevertheless strengthened complete member
  identity and neighboring-member handling.
- The harness `HttpClient` is used by `AcquirePdbAsync`; authored source bytes
  intentionally use the SSRF-hardened shared untrusted client.

All actionable findings have resolution commits. No findings remain open.

richlander:
Both exact-head adversarial reviews are clean on `f8ebc0b2`, all actionable
findings have resolution commits, CI is green, the branch is based on current
`origin/main`, and GitHub reports the PR mergeable with a clean merge state.

**Ready to merge.**

richlander:
Reviewed at head f8ebc0b, all nine commits. **Hold — one substantive fix round.** The acquisition and oracle machinery is excellent; the asks are all in how the Finding lanes surface and in one doctrinal move that needs to be made explicit.

**What's verified and strong:** the checksum gate is genuinely sound — fetch-time validation plus re-verification in `FromContent` before any decode, so unverified bytes can never become `Complete` evidence or reach extraction; the cache-poisoning fix bypasses/evicts invalid entries and never caches invalid network bytes, pinned by the poison-then-repair test with request counts. The five verdict axes stay structurally separate (`DecompilerLane` passed through untouched, determinism deliberately outside `Status`, exit code keyed only to authored-lane operational failures), the `--authored-source` gate holds (all network operations live below the early return, scope bounded to already-changed members), and the ranked identity matching with tie rejection is the right shape.

**The fix round:**

1. **The replacement-not-accretion deviation needs to be a decision, not a rewrite.** The bespoke `ToCSharpChanges`/`ToIlChanges` row reconstruction survives as the Implementation Diff's row source, with native comparisons running as a parallel shadow oracle — and implementation-diff.md's prior commitment ("retain those producer-owned values and *retire the corresponding projection fields*") was rewritten in this same PR to "their semantic rows remain." That's the doc-rewriting move the adoption checklist exists to catch (same class as #2699's first round). Two acceptable resolutions, per adoption rule 1: declare this a cross-validation period with a tracked deletion plan, or make the permanent-two-representations case explicitly — acknowledging the prior commitment was revised and *why* the semantic rows aren't superseded (richer hunk evidence is a real argument; make it in the doc, deliberately).
2. **The divergence cross-check has the wrong failure topology.** `throw new InvalidOperationException("C# Finding comparison diverged…")` in `ImplementationDiff`/`ResearchDiff` crashes the entire `diff` command if the two lanes ever disagree on one member — and lane disagreement on real-world canonicalization is exactly the thing `PromoteRawTextChanges` already had to patch once. Degrade to a per-member Failed/diagnostic row; a cross-check exists to make divergence *visible*, not fatal.
3. **Failed C#/IL Finding comparisons render as `"changed"`.** `FindingFailureChange` maps `Failed` → `ResearchChangeKind.Changed`, and the formatter remaps only Source-lane failures to `"failed"`. That's Failed leaking into the semantic-change column — the exact thing the spine forbids — and the Source lane in this same PR shows the correct pattern. Align the C#/IL lanes (and note JSONL/TSV consumers see the raw `Changed` today, so the fix belongs in the model, not the formatter).
4. **Missing checksum is `Failed`, should be `Absent`.** A PDB that records no checksum is missing evidence, not a failed inspection — and the mislabel has CI consequences: it becomes `SourceFailed` (exit 1) in the harness where `SourceAbsent` (exit 0) is the honest state. Conservative-never-Complete is right; the axis label isn't.
5. **Pin the ambiguity rejection.** The rank ladder's tie-rejection path has no test (all existing tests exercise rank *preference*), and the matcher can silently pick a wrong declaration in absence scenarios — erased-generic rank 2/1 winning alone when the exact declaration is missing from the snippet, and `.ctor` matching any constructor with no arity check. One equal-rank tie test plus an arity check on constructor matching (or a comment bounding the blast radius via the checksum-verified sequence-point window) closes the sharpest edges.

Non-blocking, for the record: the A-lane compiles under PDB-recorded options while the B-lane uses RTS defaults — defensible and Drift-reported, but "substitutes into the same final ArtifactRequest" undersells it; an Exact-vs-IlDifferent disagreement is not attributable to source alone, worth one doc sentence. `FetchSourceBytesCoreAsync` still accepts `http://` — checksums neutralize tampering but cleartext leaks which members are inspected; https-only is cheap. Verified-identical authored source produces no row, indistinguishable from not-attempted — a positive witness ("Source verified, identical") would strengthen the lane's value. And with retention on, every member decompiles twice — fine now, worth remembering when someone profiles `diff`.

---

richlander (commented):
## Independent review — VERDICT: BLOCKING

The security-critical machinery is **sound** — I verified it and it holds. What blocks merge is a live conflict plus failure-topology on the new Finding surfaces (items 2–3 match this PR's own newest, unaddressed review).

**Verified good:**
- **Checksum-verified-before-decode** — `AuthoredSourceAcquisition.FromContent` re-verifies the portable-PDB checksum and returns `Failed` on `Unavailable/Unsupported/Mismatch` *before* any `Encoding.UTF8.GetString`/`ExtractMethodBody`.
- **Cache-poisoning resistance (6aa335b9 holds)** — `SourceFetcher` gates every `CoreCache.Set` behind the validator; invalid memory bytes are removed and bypassed, invalid disk bytes fall through to network, and an invalid network payload is never cached. A poisoned cache still can't produce `Complete` evidence.
- **Concurrency (deae7342 holds)** — both caches are `ConcurrentDictionary`; values are immutable and validator-gated, so the check-then-act paths are benign. (Minor: no request coalescing — same URL may be fetched N× concurrently; perf only.)
- **Opt-in network** — byte-identical without `--authored-source` (early return before any fetch; formatter branches on `hasSourceLane`); a CLI guard rejects the flag without the Implementation Diff section. Boundary respected: Services is SRM-only, Roslyn lives only in `tools/DecompilerHarness`.
- **Member matching is conservatively tie-rejecting** — searches all members, preserves explicit-interface specifiers, ranks by identity, and a top-score tie sets `ambiguous` → returns false rather than guessing.

**Blockers:**
1. **Active merge conflict** — `mergeable_state: dirty` against `main` (contradicts the "clean/ready" comment). Needs resolution + re-review.
2. **Lane-divergence `throw` aborts the command.** `ResearchDiff.AddCSharpDiff`/`AddIlBodyDiff` and `ImplementationDiff.CompareMembers` throw `InvalidOperationException` when the native Finding comparison disagrees with the semantic diff. Not reachable on plain `-S "Implementation Diff"` or `--authored-source`, but **is** reachable on the new user-invocable `diff --finding csharp.line` / `--finding il.op` surface. Any real-world canonicalizer disagreement crashes the whole diff instead of degrading to a `Failed` row. Aggravating: the C# cross-check compares `Anchor.StableSelector` across two independently-produced row sets, which can itself throw spuriously.
3. **`Failed` C#/IL comparisons render as `"changed"`.** `FindingFailureChange`/`AddFindingComparisonFailure` emit `ResearchChangeKind.Changed`; `DiffOutputFormatter` remaps to `"failed"` **only** for `Mechanism == Source`. A decode/decompile failure on the C#/IL lanes leaks into the semantic-change column (and raw `Changed` in JSONL/TSV) — violates this PR's own "Failed is never a semantic change" contract. The fix belongs in the model, not the formatter — the Source lane already shows the correct pattern.

**Minor non-blockers:** missing-PDB-checksum maps to `Failed` where `Absent` (missing evidence, exit 0) is more honest; `http://` is accepted (prefer https-only — checksums neutralize tampering but cleartext still leaks which members are inspected).

---
*Independent review by Claude Code — static (diff + source at head `f8ebc0b`; no local build/test run).*

---

richlander:
Feedback round resolved on exact head `f060942d`:

1. `implementation-diff.md` now explicitly revises the earlier projection-retirement plan and records the permanent dual-representation decision: native Finding comparisons own inspection/census transitions; semantic C#/IL projections retain aligned hunks, typed display failures, offsets, and producer-formatted evidence. Equivalent future Finding payloads are the deletion criterion.
2. C#/IL cross-validation disagreement now emits a scoped per-member `Failed` diagnostic instead of throwing and erasing healthy members.
3. Finding acquisition failures are model-level `ResearchChangeKind.Failed`; table, TSV, JSONL, and programmatic consumers therefore no longer receive semantic `Changed` rows.
4. Missing checksum metadata is now `Absent` with mapping/document provenance and `Unavailable` verification; acquisition preflights it before any fetch. Unsupported algorithms, mismatch, fetch, and extraction failures remain `Failed`.
5. Tests now pin equal-rank generic ambiguity and multiple-constructor rejection. The constructor matcher comment documents the checksum-verified sequence-point-window boundary and rejects ambiguous multi-constructor windows.

The non-blocking context note is also documented: authored A→IL may use PDB-recorded options while decompiled B→IL uses RTS context, so build-context and determinism verdicts remain necessary when interpreting disagreement.

The branch is rebased onto current `origin/main`; local build, focused Research/product/Services/authored-rebuild suites, and changed-Markdown lint pass. Two fresh exact-head adversarial reviews are pending before this returns to ready state.

richlander:
Final adversarial review reconciliation for exact head `d249a966`:

- **Gemini 3.1 Pro:** the first feedback-head review found that operational semantic failures could still emit `Changed` plus synthetic add/remove rows beside a native Finding `Failed` row. Resolved by `d249a966`: operational semantic failures are now model-level `Failed`; duplicate generic Finding failures are suppressed in favor of richer typed semantic failures; C# rows sharing an operational-failure hunk are omitted; genuine body absence and partial decoded IL evidence remain visible. Gemini's fresh exact-head review is clean after checking these paths and running all 111 Research tests.
- **Claude Opus 4.8:** independently reviewed the whole exact head, including native/semantic failure deduplication, subject/descriptor scoping, failure-hunk filtering, divergence topology, checksum/cache ordering, reader lifetimes, identity ambiguity, and rebased CLI wiring. Its fresh exact-head review is clean; it also ran the Research suite and focused decompiler/authored-rebuild coverage.

The reviewers agreed that the five maintainer requests are fully resolved and found no remaining actionable issues. The only actionable final-round finding has resolution commit `d249a966`; no dismissed findings or non-actions remain.

richlander:
Current-main exact-head review update for `86a24928`:

- **Claude Opus 4.8:** clean on the whole rebased head. It verified the failure-dedup and hunk-filtering fix, envelope/one-sided cases, checksum/cache ordering, SourceLink security boundary, identity ambiguity, reader lifetimes, and current CLI wiring, with solution/focused test runs.
- **Gemini 3.1 Pro:** initially re-raised the injected-client concern. This was dismissed because SourceLink URLs are attacker-controlled PDB data and intentionally use `SharedUntrustedFetch`; substituting the package/PDB client would weaken the SSRF boundary. Its related duplicate-network claim was also disproved by `SourceFetcher`'s persistent exact-byte cache across instances. Gemini then re-reviewed the same exact head in a fresh clean worktree and returned clean, classifying memory-cache reuse as optional only.

Both permitted model families are now clean on exact head `86a24928`, based directly on current `origin/main` (`6c041f97`). The only dismissed item has no resolution commit because the current client split is the required security behavior; no findings remain open.

richlander:
Latest-base exact-head review update for `20ca5259`:

`main` gained #2720's signature-guard/receiver-indexing changes and #2721, so the branch was rebased again before the ready signal. **Claude Opus 4.8** and **Gemini 3.1 Pro** independently reviewed the whole new exact head in fresh isolated worktrees; both are clean. Their targeted checks included #2720 interactions with canonical member identities, `LibraryBodyIndex.DeclaredMethods`, C#/IL retained comparison keys, explicit/constructed-generic matching, failure deduplication, checksum/cache security, and reader lifetimes. Metadata (483), Research (111), product diff (112), Services focus (5), and authored-rebuild focus (18) also pass locally.

No new findings arose from the rebase. The earlier SourceLink client concern remains intentionally dismissed: attacker-controlled URLs require `SharedUntrustedFetch`, and the persistent byte cache prevents the claimed cross-instance network refetch. No resolution commit applies.

richlander:
All maintainer feedback is resolved, both exact-head adversarial reviews are clean on `20ca5259`, CI is green, the branch is based on `fe9d04a0`, and GitHub reports a clean mergeable state.

**Ready to merge.**

PR #2718, opened 2026-07-13
Make Metadata signature rejection explicit

### LINKED ISSUE #2689: Define explicit Metadata signature degradation and TypeSpec acceptance

richlander:
Closes #2689.

Guarded Metadata string-signature decodes now return a typed `Decoded` or `Rejected` outcome instead of silently presenting `object`, an empty method-spec argument list, or a default method/property signature as genuine metadata.

This preserves #2709 and #2715 provenance behavior: API census consumers project rejected signatures as `SignatureDecodeStatus.Degraded`, while discovery consumers skip rejected rows and direct rendering boundaries throw through their existing row-level containment. Nested TypeSpec rejection propagates to the enclosing decode.

TypeSpec acceptance is now route-independent. The legacy 1,024-byte per-blob cap is removed; both typed gateways use the documented 4,096-byte active-closure budget, 256 cross-handle depth limit, and 512 structural-depth guard. A valid 1,500-argument shallow TypeSpec proves the widened legal envelope. The existing public `TypeResolver.GetTypeNameFromSpecification` signature remains compatible; `DecodeTypeNameFromSpecification` exposes the typed outcome.

Validation:
- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release` (474 passed after the final main rebase)
- `dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.GuardedSignatureTextTests"` (1 passed)
- full Decompiler suite before the final main rebases (2,528 passed, 1 SDK-dependent skip)
- `npx markdownlint-cli docs/metadata-primitives.md`
- GitHub Actions: all merge-blocking jobs passed

Adversarial review: Gemini 3.1 Pro and Claude Opus 4.8 both reported CLEAN on fixed head `a879869a`; round-1 findings and their resolution are reconciled in PR comments.

### DIFF
--- a/docs/metadata-primitives.md
+++ b/docs/metadata-primitives.md
@@ -79,6 +79,41 @@ helper is SRM-only. `Analysis` stays independent of `Metadata`.
   metadata *product* surface
 - each consumer's own `TypeRef`/IR model and its `ISignatureTypeProvider<T>`
 
+## Guarded string-signature decoding
+
+String-producing SRM decodes return `SignatureDecodeResult<T>`. `Decoded`
+carries the real field, method, property, method-spec, or TypeSpec value;
+`Rejected` carries a `SignatureDecodeRejectionKind` and detail. Rejection is
+never represented by a plausible `object` type, a default method signature, or
+an empty generic-argument list.
+
+Consumers fail closed according to their operation:
+
+- discovery and classification scanners skip the rejected member;
+- API-surface census construction projects a visible degraded row with
+  `SignatureDecodeStatus.Degraded`;
+- formatting helpers may throw `BadImageFormatException` through
+  `GetValueOrThrow` when their caller already owns a row-level failure boundary.
+
+This contract governs the shared string-signature decoder and is separate from
+the provider-specific `TypeNode` and semantic-tree paths guarded for stack
+safety under #2575. Those providers retain their own result and fallback
+policies.
+
+The safety envelope has two independent limits:
+
+- `SignatureBlobGuard.DefaultMaxDepth` is 512 structural type nodes within one
+  signature blob;
+- `TypeSpecGuard.MaxDepth` is 256 cross-handle TypeSpec re-entries, with
+  `TypeSpecGuard.MaxCumulativeBytes` fixed at 4,096 bytes across the active
+  TypeSpec closure.
+
+A single wide, shallow TypeSpec may consume the full 4,096-byte closure budget.
+There is no separate 1,024-byte per-blob cap. Both
+`GuardedSignatureText.TypeSpecText` and
+`TypeResolver.GetTypeNameFromSpecification` enter through the same
+`TypeSpecGuard`, so legal-input acceptance cannot vary by route.
+
 ## The key finding: it's an entangled cluster, used widely
 
 The shareable primitives are not independent functions — `TypeResolver` and
--- a/docs/metadata-primitives.md
+++ b/docs/metadata-primitives.md
@@ -111,8 +111,10 @@ The safety envelope has two independent limits:
 A single wide, shallow TypeSpec may consume the full 4,096-byte closure budget.
 There is no separate 1,024-byte per-blob cap. Both
 `GuardedSignatureText.TypeSpecText` and
-`TypeResolver.GetTypeNameFromSpecification` enter through the same
-`TypeSpecGuard`, so legal-input acceptance cannot vary by route.
+`TypeResolver.DecodeTypeNameFromSpecification` enter through the same
+`TypeSpecGuard`, so legal-input acceptance cannot vary by route. The
+compatibility API `GetTypeNameFromSpecification` returns a decoded value or
+throws on rejection; it does not restore a plausible fallback.
 
 ## The key finding: it's an entangled cluster, used widely
 
--- a/src/ILInspector.Decompiler.Tests/GuardedSignatureTextTests.cs
+++ b/src/ILInspector.Decompiler.Tests/GuardedSignatureTextTests.cs
@@ -6,12 +6,12 @@ using ILInspector.Decompiler;
 namespace ILInspector.Decompiler.Tests;
 
 // GuardedSignatureText routes the compile-back / type-source composers' string-producing signature
-// decodes through SignatureBlobGuard, so a malformed deeply-nested signature degrades to a
-// placeholder type instead of overflowing the native stack inside SRM.
+// decodes through SignatureBlobGuard, so a malformed deeply-nested signature is rejected
+// explicitly instead of overflowing the native stack inside SRM.
 public class GuardedSignatureTextTests
 {
     [Fact]
-    public void DeepFieldSignature_DegradesToPlaceholder()
+    public void DeepFieldSignature_IsRejected()
     {
         // FIELD header then a 600-deep array (over the 512 guard limit).
         var sig = new BlobBuilder();
@@ -21,9 +21,11 @@ public class GuardedSignatureTextTests
         sig.WriteByte(0x08);     // I4
 
         var (reader, handle) = BuildField(sig);
-        var text = GuardedSignatureText.FieldText(reader, reader.GetFieldDefinition(handle), context: null);
-
-        Assert.Equal("object", text);
+        Assert.Throws<BadImageFormatException>(() =>
+            GuardedSignatureText.FieldText(
+                reader,
+                reader.GetFieldDefinition(handle),
+                context: null));
     }
 
     static (MetadataReader Reader, FieldDefinitionHandle Handle) BuildField(BlobBuilder sig)
--- a/src/ILInspector.Decompiler/GuardedSignatureText.cs
+++ b/src/ILInspector.Decompiler/GuardedSignatureText.cs
@@ -12,37 +12,36 @@ namespace ILInspector.Decompiler;
 /// deeply-nested or huge-count signature would overflow the native stack inside SRM's
 /// <c>SignatureDecoder</c> (an uncatchable <c>StackOverflowException</c>) or trigger a
 /// multi-gigabyte pre-allocation. <see cref="SignatureBlobGuard"/> detects both before decoding,
-/// so these helpers fail closed to an unresolved placeholder type (a parseable <c>object</c>)
-/// instead of crashing. Real signatures are shallow, so the guard only trips on malformed input.
+/// so these helpers reject the member instead of fabricating a parseable <c>object</c> signature.
+/// Real signatures are shallow, so the guard only trips on malformed input.
 /// </summary>
 internal static class GuardedSignatureText
 {
-    // A degraded but syntactically valid C# type keeps composed source parseable; a member whose
-    // signature could not be decoded is already un-reconstructable, so its exact text is moot.
-    const string Unresolved = "object";
-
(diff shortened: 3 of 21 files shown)


richlander:
Reviewed at head 54aa898, all 21 files, plus CI verification (green at head). **Endorse — safe to merge**, with two accuracy notes worth a line in the body/release notes before you do.

**The replacement is real.** Within the string-signature lane, every silent-fallback shape is deleted — `Unresolved`/`UnresolvedMethod`, #2709's `TrackingSignatureDecoder`, the `*TextResult` dual overloads — leaving one result-returning form per entry point, compiler-enforced across all 28 migrated call sites. `SignatureDecodeResult<T>.Decoded/Rejected` with typed rejection kinds is the right shape (`Decoded` null-checks; a rejection can't smuggle a placeholder), and the nested-rejection propagation via the save/restored thread-static channel is correct: SRM's callback contract still gets a continuable string mid-decode, but the enclosing outcome flips to `Rejected` and the placeholder never escapes the gateway — pinned by the self-referential-modifier test. Consumer handling is doctrine-clean at every fork I care about: the census lane projects `Rejected` → visible `Degraded` rows through one conversion point (and now catches two cases #2709 missed — mid-decode `BadImageFormatException`/AOORE), scanners skip, formatting helpers throw into row-level containment with visible elision (`"(...)"`, diagnostic comment rows). The provider/TypeNode lanes retaining their own fallback policies is deliberate, documented in metadata-primitives.md, and scope-matched to #2689.

**The cap-removal analysis holds.** Removing the 1,024 per-blob cap does not weaken StackOverflow protection: structural depth is still prescan-bounded at 512 per blob, and the 4,096-byte cumulative budget bounds total live native frames across the re-entry chain at the same ceiling as before (every structural level costs ≥1 byte) — the per-blob cap only false-rejected wide-shallow legal blobs, which the 1,500-arg envelope test now proves accepted identically on both routes. Ordering keeps the O(1) budget check ahead of the O(n) walk, and the TypeRefDecoder lanes keep their own caps untouched. Route-independence is enforced structurally (one implementation, both gateways delegating) and pinned in both accept and reject directions. The census gains the gateway as a third sanctioned form with receiver-bound blob matching, and the containment workers now assert the two Degraded lanes agree.

**The two notes:**
1. **The envelope moved in both directions, and the body only claims widening.** Top-level TypeSpecs >4,096 bytes previously decoded via both `TypeSpecText` routes (structural prescan only, no byte cap) and now reject; meanwhile the widened 1,025–4,096-byte nested band changes `CanonicalIL` canonical signatures and `ApiMemberIdentity` anchors/fingerprints for affected members — real text instead of `object`, Degraded markers gone, fingerprints drifted. No corpus baselines moved, which supports "no real-world impact," but persisted anchors over pathological assemblies would drift — one release-note sentence covers it.
2. **"Route-independent" is true within the string lane only.** The same 1,500-arg TypeSpec decodes as a string but is still `TypeRef.Unsupported` in the Analysis/Decompiler-pipeline lanes (per-blob 1,024 retained) and enters no byte budget at all on the TypeNode lane. #2689's "centralize or document shared caps" is met by documentation; the TypeRefDecoder literals remain free to drift, guarded only by their own recursion tests — worth a cross-reference comment on those constants.

Non-blocking residue: `ILTokenResolver`'s `catch { return baseName; }` is now the weakest Rejected handler (silent parameter-list drop, no `(...)` marker, broad catch) — one-line tighten to match its siblings; a comment on `Decode`'s save/restore noting that rejection propagates via `Reject()` only (a nested `Decode` whose `Rejected` is swallowed by `TryGetValue` does not poison the outer scope — no such path exists today, but the invariant deserves stating); and the degraded rows are deliberately less informative than #2709's partial signatures (fully-generic placeholder instead of mostly-correct text) — right per fail-closed doctrine, but a rendering-visible change worth acknowledging in the notes alongside item 1.

---

richlander:
Adversarial review round 1 at `54aa8980`:

- **Gemini 3.1 Pro** found two actionable issues: malformed `BlobHandle` failures occurred before the typed exception boundary, and changing `TypeResolver.GetTypeNameFromSpecification`'s public return type was compatibility-breaking.
- **Claude Opus 4.8** reported CLEAN on the same head. I accepted Gemini's concrete malformed-handle analysis despite that divergence.

Both findings are resolved on `a879869a`: preflight now runs inside the typed decode boundary with a malformed-handle regression test, and the existing public `string` API is preserved while `DecodeTypeNameFromSpecification` provides the explicit result. The branch was also rebased over #2715 and revalidated. Both reviewers are now re-reviewing this fixed head in new isolated worktrees; this PR is not ready until both report clean.

richlander:
Fixed-head adversarial re-review at `a879869a` is complete:

- **Gemini 3.1 Pro: CLEAN.** Verified malformed preflight containment, the restored public TypeResolver API shape, nested rejection/accounting, degraded provenance, consumer policies, and the new regression.
- **Claude Opus 4.8: CLEAN.** Rebuilt the solution and ran the full Metadata suite (474/474), targeted gateway/boundary tests, and Instructions signature-safety tests from the isolated fixed-head worktree.

The round-1 Gemini findings were resolved by [`a879869a`](https://github.com/richlander/dotnet-inspect/commit/a879869a39da0039bcceaaccd508476714c12703). No guidance was dismissed. Both isolated review worktrees were clean and pinned to the PR head.

The PR is awaiting the remaining GitHub Actions test job; it is not marked ready until that merge-blocking check completes.

richlander:
Ready to merge. All merge-blocking CI passes, and Gemini 3.1 Pro plus Claude Opus 4.8 both reported CLEAN on fixed head `a879869a`.

richlander:
## Review — CLEAN (Claude Opus 4.8)

Reviewed head `a879869a` in an isolated worktree (merge-base = current `origin/main` d6e5f942). Focus: does rejection ever surface as a success-shaped placeholder, and is every new throwing boundary contained where the old code degraded to `"object"`. No high-confidence correctness/security/compatibility findings.

### Verified

**Rejection contract & channel**
- `SignatureDecodeResult<T>` (Decoded/Rejected) + `SignatureDecodeRejection(Kind, Detail)`; `GetValueOrThrow` throws `BadImageFormatException`, `TryGetValue` projects. Rejection is never a placeholder.
- Thread-local channel in `SignatureDecoder.Decode<T>`: `[ThreadStatic] s_rejection` saved/nulled/restored around preflight+decode; first-wins via `??=`; returns Rejected if the channel is set or on caught `BadImageFormatException`/`ArgumentOutOfRangeException` (→ `MalformedMetadata`). Nesting is correct (save/restore), thread-safe (ThreadStatic). Catching AOORE additionally closes a path that could previously escape.

**Widened TypeSpec envelope**
- `TypeSpecGuard` drops the per-blob `MaxBlobLength=1024` cap, keeps `MaxCumulativeBytes=4096` (`(long)` overflow-guarded) + `MaxDepth=256` + structural-depth prescan (`SignatureBlobGuard`, DefaultMaxDepth=512). Sound: native SRM `DecodeType` recursion is bounded by *structural depth* (≤512), not blob length, so a wide-but-shallow blob is safe. Proven by `WideTypeSpec_AboveLegacyPerBlobCap_IsDecoded` (1,500 args, >1024B, must decode) vs `TypeSpec_AboveCumulativeBudget_IsRejected`.

**Every throwing boundary is contained**
- Census surfaces (`MetadataDeclarationQuery.ProjectDecode`) map Rejected → `SignatureDecodeStatus.Degraded` with `DegradedType="object"` — degradation stays *visible*, preserving the #2715 provenance behavior.
- `TypeResolver.GetTypeName` (nullable) → `null` on TypeSpec reject; all callers null-safe (`??`/null-safe compares).
- `GetTypeNameFromSpecification` / `GetMethodReturnType` use `GetValueOrThrow` but have **no internal product callers** (public-API compat only) — throw unreachable in-repo.
- Consumers using `GetValueOrThrow`: EcosystemIntegrationScanner, OpenTelemetryScanner, MethodClassificationScanner, ILTokenResolver, ApiSurfaceExtractor — all inside `try/catch (BadImageFormatException)`.
- `ExtensionMethodScanner.FormatMethodSignature` re-decodes with `GetValueOrThrow` but the same loop iteration first does `if (!MethodText(...).TryGetValue(out ...)) continue;` → deterministically non-throwing.
- Decompiler `TypeSourceComposer` (the compile-back path): all three sites contained — `FieldText` @310 `catch { continue; }` (skips field); `PropertyText` @528 and `MethodText` @557 `catch { return null; }` (abandons the type). Behavior note below.

### Behavior change (observation, not a finding)
For compile-back, a rejected property/method signature now abandons the whole type composition (`return null`) instead of degrading that member to `"object"` and continuing. This only triggers on malformed/adversarial signatures, is contained, and is preferable — a fabricated `"object"` placeholder would emit misleading source. Called out for awareness, not blocking.

### Runs (this worktree, Release, `--configfile nuget.config`)
- `SignatureDecoderSafetyTests` + `ProviderSignatureDecodeBoundaryTests` + `StringSignatureDecodeBoundaryTests` + `SignatureBlobGuardTests` → **Total: 45, Failed: 0**. Includes `NestedTypeSpecRejection_RejectsContainingMethod` (ThreadStatic propagation proof), `GuardedGateways_RejectEveryUnsafeSignatureKind`, malformed-handle cases, and `_IsContainedInChildProcess` StackOverflow-containment coverage.
- `GuardedSignatureTextTests` → **Total: 1, Failed: 0**.
- All CI checks green (`test` 6m20s pass), mergeState **CLEAN**.

### Roster note
I'm Claude Opus 4.8, so this counts as the Opus slot. Per the adversarial-review policy this PR still wants a **second reviewer from a different model family** (GPT-5.5 / Gemini Pro / MAI Flash) to sign off before it's "two clean reviews". Not merging.

PR #2719, opened 2026-07-13
Harden the signature decode guard matrix

richlander:
## Summary

Follow-up hardening for the signature-decode guard matrix from #2490/#2700:

- make `TypeSpecGuard.TryEnter` return a disposable, token-identity scope and
  remove public manual `Exit` accounting;
- replace file-granular TypeSpec census acceptance with per-invocation
  reader/handle/scope binding;
- encode the reviewer-evasion matrix as executable negative tests, including
  aliases, deferred execution, shadowing, mutation, control-flow, and damaged
  counter patterns;
- extend the census to `ILInspector.Analysis` and `ILInspector.Decompiler` with
  one exact bounded `TypeRefDecoder` template;
- add the missing Analysis structural prescan and a runtime pin for the
  7-byte ARRAY signature that declares about 536 million sizes;
- classify TypeSpec re-entry from `GetTypeSpecification` provenance rather
  than an enclosing provider method name, including hoisted helpers.

The census remains Roslyn-only test infrastructure. Product assemblies remain
SRM-only, NativeAOT-compatible, and free of inspected-assembly loading.

## Safety and behavior

- `TypeSpecGuard.Scope` is a `readonly ref struct`; successful entries receive
  a nonzero per-thread token, and disposal only unwinds the current LIFO token.
  Default, copied, stale, double-disposed, or out-of-order scopes cannot
  under-count depth/bytes.
- Every product `TryEnter` call immediately rejects failure and encloses its
  decode in `using (scope)`.
- Analysis and Decompiler now both apply recursion, blob-length, cumulative
  byte, and structural-count guards before SRM TypeSpec decode.
- All top-level TypeSpec gateways use `TypeSpecGuard`; a receiver-bound
  structural prescan alone is never accepted for cross-handle re-entry.
- CoreLib output is byte-identical to base `d6e5f942` for `type string`,
  `library System.Private.CoreLib`, `extensions string`, `member string -S IL`,
  and `member string --unsafe`.

## Validation

- Solution build: succeeded.
- Metadata tests: 521 passed.
- Instructions tests: 116 passed.
- Analysis tests: 428 passed.
- Targeted Decompiler TypeRef/guard tests: 9 passed.
- NativeAOT publish: succeeded.
- Guard census: 53 passed, including the real-tree closure proof and the
  executable negative/sanctioned-shape matrix.

The full Decompiler suite was attempted but two unrelated corpus/compile-back
tests stalled beyond 30 minutes; the changed Decompiler surface is the
source-level census, and the targeted TypeRef/guard tests pass.

## Adversarial review

Claude Opus 4.8 and Gemini 3.1 Pro reviewed every updated head in separate clean
worktrees. Findings were reproduced before action and converted into negative
tests. The rounds converged on these classes:

| Reviewer | Verified guidance | Resolution |
| --- | --- | --- |
| Opus | Deferred lambda/local-function execution; unconstrained counter/control-flow matching; lexical alias and mutation gaps | `dca93300`, `c14192bd`, `27b98d20`, `5141321f` |
| Gemini | TypeSpec origin/blob-length binding; composite receiver/root/shadow/ref/pointer aliases; missing Analysis structural prescan | `aac8c81a`, `012a6921`, `e4f5ad43`, `1e06e49c`, `a4238362` |
| Both | Prefix/qualified receiver mutation, pre-guard aliases/closures, special TypeRefDecoder closure paths | `fe999e08`, `62c494ea`, `8651c848`, `0261df96` |
| PR feedback | Hoisted TypeSpec helper could evade re-entry classification | `696157b6` |

Deliberately equivalent but differently spelled safe shapes may be rejected by
the census; that conservatism is intentional for this anti-ratchet proof.

The branch was rebased onto `d6e5f942`; current-main degraded pointer and
spellability evidence was preserved through the disposable-scope migration.

Final exact-head `696157b6`: **Opus CLEAN; Gemini CLEAN.** Both reviewers
reattacked helper hoisting, alias/mutation provenance, migrated gateways,
special TypeRefDecoders, scope accounting, and the rebase resolutions.

The four non-blocking follow-ups remain explicit non-actions: guard-type name
shadowing and prescan argument-strength checks are independent census
hardening; deduplicating the intentionally frozen TypeRefDecoder template is a
future structural change; and a MemberResolver end-to-end bomb test would add
integration coverage without changing this guard closure.


richlander:
CI is green on exact head `a4238362`, with two CLEAN isolated adversarial reviews (Claude Opus 4.8 and Gemini 3.1 Pro) and the full reconciliation in the PR body. All merge-blocking local validation is complete.

**Ready to merge.** Awaiting maintainer approval; not self-merging.

richlander:
Reviewed at head a423836, all fourteen commits. **One pre-merge ask; everything else is the strongest version of this census yet.**

**The Scope primitive's under-count claim is sound.** I had it checked against every misuse shape: dispose-of-default (no-op via zero token), double-dispose and copied-scope dispose (token mismatch after the first — and the `do…while` excluding `parentToken` closes the parent-reuse corner), out-of-order disposal (fails toward *over*-counting, never under), stale scopes (monotonic 64-bit tokens, never reused), cross-thread capture (ref struct — structurally impossible), and forgery (`internal` ctor, no `InternalsVisibleTo`, grep-verified). `Exit` is private now — the accretion I flagged on #2700 is gone, and the runtime pins (copied-scope double-dispose with the depth limit still tripping at exactly 256; the 257-exception-loop disposal test) cover the paths that matter. The one residual — a legitimate early dispose of the live scope shrinking the guard window — is correctly closed at the census layer (scope identifier inside the `using` body is a violation, pinned), and the PR's test matrix encodes that runtime+census pairing explicitly.

**The census work is the headline.** The 45 negative cases run the *production classifier* against in-memory adversarial snippets — not fixtures, not a parallel reimplementation — so the previously-manual reviewer evasions (including the deferred-lambda-in-ternary escape I documented as theoretical on #2700) are now a permanent executable regression suite, with the five sanctioned spellings pinned from the positive side as a runnable spec for future authors. Per-invocation reader/handle/scope binding kills the file-granular fact properly. The Analysis ARRAY-bomb fix is real and worth naming precisely: the structural-count guard predates this PR — the hole was that Analysis's `TypeRefDecoder` never *called* it, so a 7-byte blob declaring ~536M array sizes reached SRM's allocation path; the wiring plus the exact-blob runtime pin closes it, and the count-bound coverage (generic arity, param counts, locals, method-spec args) checks out across the grammar. NativeAOT purity verified — no csproj changes, Roslyn stays test-only, the one new product edge (Analysis → `SignatureBlobGuard`) was already in the dependency graph.

**The ask — the re-entry definition regressed in one dimension.** `IsNestedTypeSpecReentry` is now keyed to the *enclosing method name* (`GetTypeFromSpecification`) instead of the old shape test (receiver is a `GetTypeSpecification(...)` invocation). Consequence: hoist a provider's `GetTypeFromSpecification` body into a helper with a receiver-bound prescan, and the census sanctions it as a top-level prescan — but it's cross-handle re-entry with **no `TypeSpecGuard`**, so a modreq cycle of individually-shallow blobs recurses unboundedly. That's exactly the scenario the TypeSpecGuard doc says the prescan cannot see, and the *old* census would have flagged it. Fix is cheap: classify any decode whose traced origin is a `GetTypeSpecification` invocation as re-entry (requiring the scope binding) regardless of enclosing method name — the origin-tracing machinery this PR built is already the right tool. One negative test for the hoisted-helper shape locks it.

Non-blocking, for the census follow-up list:
1. **Guard-type name shadowing is untested**: the classifier trusts `SignatureBlobGuard`/`TypeSpecGuard` by name, so an impostor type declared inside a census'd assembly would bind every prescan there. One census fact — no type by those names declared under the source roots outside MetadataPrimitives — closes it for a few lines.
2. **`maxDepth:`/`Kind` neutering remains unpinned for the prescan forms** (the TypeSpec form is now strength-hardened since `TryEnter` embeds its own caps). Carried over from #2700's known weaknesses; worth an argument-shape check eventually.
3. The frozen nine-statement TypeRefDecoder template locks the two decoders as clones — intended anti-ratchet, and the SanctionedShapes doc is the right mitigation, but note it makes the eventual dedup-into-shared-helper refactor a deliberate census event.
4. Analysis got its end-to-end bomb pin; the equivalent `MemberResolver` param-count-bomb path is covered only at guard-unit level.

With the re-entry fix, this PR plus #2711 completes #2490 beyond its original acceptance criteria — every product decode census-enforced across five assemblies, guard misuse structurally impossible, and evasion resistance executable instead of manual.

---

richlander:
Resolved the blocking re-entry feedback in `696157b6` after rebasing onto current `origin/main` (`d6e5f942`). TypeSpec re-entry is now classified by traced `GetTypeSpecification` provenance regardless of enclosing method name; immutable provenance remains separately required to accept a bound `TypeSpecGuard` scope. A hoisted prescan-only helper is now a permanent negative census case, and the three remaining top-level TypeSpec gateways were migrated to `TypeSpecGuard`.

Rebase conflicts preserve main’s `PointerDetection.Degraded` and `SpellabilityEvidence.Degraded` results. Fresh isolated exact-head reviews are **Claude Opus 4.8 CLEAN** and **Gemini 3.1 Pro CLEAN**. Both reattacked helper hoisting, aliases/mutations, gateway coverage, special TypeRefDecoders, scope accounting, and the conflict resolutions. Local validation and the CoreLib render A/B are green; CI is rerunning on `696157b6`.

The four non-blocking suggestions remain explicit follow-ups rather than changes here: guard-name shadowing and prescan argument-strength checks are independent census hardening; the exact TypeRefDecoder clone is intentionally frozen; and a MemberResolver end-to-end bomb test would add integration coverage without changing this closure.

richlander:
CI is green on exact head `696157b6`, with fresh CLEAN isolated reviews from Claude Opus 4.8 and Gemini 3.1 Pro after resolving the blocking re-entry feedback. The PR body and feedback-resolution comment contain the exact-head reconciliation and explicit non-actions.

**Ready to merge.** Awaiting maintainer approval; not self-merging.

richlander (commented):
## Independent review — VERDICT: MINOR-CONCERNS

**The ARRAY-bomb DoS is genuinely fixed in the product path, not just asserted by tests.** Traced end to end: `ILInspector.Analysis/TypeRefDecoder.cs:131` now calls `SignatureBlobGuard.IsSafeToDecode(..., Kind.TypeSpecification)` *before* `DecodeSignature`; `SkipArrayShape` reads `numSizes` and returns unsafe when `numSizes (~536M) > RemainingBytes`, so decode is refused before SRM's `ImmutableArray.CreateBuilder<int>(~536M)` (~2 GB) ever runs. The guarded method is reached from real product entry points (`MemberResolver.cs:87`, `LeakTriage.cs:506`, `LibraryBodyIndex.cs`). The genuinely *new* hole this closes is the **Analysis** `TypeRefDecoder` (Decompiler already had the prescan).

**Disposable-scope accounting is sound** — `readonly ref struct Scope`, `[ThreadStatic]` state, dispose keyed on `s_currentToken == _token`: default/double-dispose/copied-double-dispose all no-op; out-of-order skips Exit (over-counts, fail-closed); monotonic 64-bit tokens never reused. Under-count is structurally impossible.

**Two test-only anti-ratchet gaps** (not runtime holes — product decodes inline, plain-spelled):
- **F1:** the census matches on `Identifier.Text`, so `spec.@DecodeSignature(...)` / `spec.DecodeSignature(...)` (verbatim/Unicode-escaped) are never recognized as decode occurrences — a future unguarded decode spelled that way passes silently. This is exactly the gap #2720 closes by switching to `.ValueText` with dedicated negatives.
- **F2:** `TryGetTypeSpecificationOrigin` traces the receiver back only through an *inline* `reader.GetTypeSpecification(handle)`; a helper-returned spec (`var spec = GetSpec(handle);`) escapes `IsNestedTypeSpecReentry`, sanctioning the prescan-only form for a genuine cross-handle re-entry.

### ⚠️ Overlap with #2720 — competing, not composable
Both convert `TypeSpecGuard.TryEnter → out Scope`, remove public `Exit`, and rewrite the same census across the **same 10 product files** — they will hard-conflict. Neither is a strict superset, but **#2719 is the stronger guard PR**: it carries the actual new Analysis product fix, extends the census to Analysis + Decompiler, adds the ARRAY runtime pin, and uses a robust `readonly ref struct` token scope (vs #2720's mutable `ref struct` whose `Exit` throws on out-of-order/copied double-dispose). #2720 uniquely has (a) the `.ValueText` hardening that closes **F1**, and (b) an unrelated DecompilerHarness receiver-arity fix.

**Recommendation:** land **#2719 as the base**, port #2720's `.ValueText` hardening + verbatim/Unicode negatives into it, split the DecompilerHarness arity fix into its own orthogonal PR, and close #2720 as superseded. Do not attempt to land both.

Note: no formal GitHub reviews are on record for this PR — the "Opus/Gemini CLEAN" attestations are owner comments.

---
*Independent review by Claude Code — static (diff + source at head `696157b`; no local build/test run).*

---

richlander:
Post-main-merge reconciliation for head `cfa98ac2`:

- Merged latest `origin/main` after #2720/#2721 landed and resolved guard-matrix overlaps in `GuardedSignatureText`, `SignatureDecoder`, `TypeResolver`, `TypeSpecGuard`, and `ProviderSignatureDecodeBoundaryTests`.
- Conflict resolution keeps main's `SignatureDecodeResult` gateway API, retains this PR's token-identity `readonly ref struct` `TypeSpecGuard.Scope`, adds the `SignatureDecodeRejectionKind` overload shape required by main, ports #2720's `.ValueText` identifier hardening, and teaches the census to accept the merged `GuardedSignatureDecoder.Decode` lambda gateway and 4-argument `TryEnter` shape.
- Local validation: solution build; metadata guard/signature tests; Analysis `TypeRefDecoderRecursionTests`; Decompiler `GuardedSignatureTextTests`.
- Gemini 3.1 Pro final re-review: CLEAN. Verified token-identity scope plus rejection-kind API, nested rejection propagation through `SignatureDecodeResult`, guarded wrapper compatibility, and census coverage for merged AST/gateway shapes.
- Claude Opus 4.8 final re-review: CLEAN. Verified no-under-count guarantees, nested rejection surfacing, guarded wrappers, real-source census closure, helper-hoisted reentry, alias/mutation/deferred boundaries, `.ValueText` evasions, and merged gateway/4-arg `TryEnter` acceptance.
- CI is green and the PR is mergeable.

Ready to merge.

PR #2713, opened 2026-07-13
Batch quality residue hardening

richlander:
## Change

**Conclusion:** **REVIEW** — batches four small quality-residue asks that were not worth solo PRs, without changing decompiler render output.

This PR retires:

- **#2687 residue:** receiver-type metadata index now keeps all normalized receiver handles. If generic-arity-erased names collide (`Result` vs `Result<T>`), receiver hierarchy evidence is marked incomplete so extension roots flow through visible fallback instead of first-wins silent exclusion.
- **#2705 residue:** removes the unused `ILInspector.Research` import from `TimelineCommand` and disposes timeline analysis endpoints after each cell's analysis inspection.
- **#2707 residue:** source-document canonical identity and resolved SourceLink URL now come from one local longest-prefix, case-insensitive SourceLink match, preventing identity/storage incoherence from two independent matchers.
- **#2709 residue:** documents the cached-surface cross-epoch `null` vs `Degraded` decode-status diff skew.

Deferred: the **#2708 shared unsupported-shape predicate** ask is not on `origin/main` yet because #2708 is still an open constructor-chain PR. That should be handled in a second batch after #2708 lands.

## Evidence

| Check | Result |
| --- | --- |
| Decompiler test build | Pass: `dotnet build src/ILInspector.Decompiler.Tests/ILInspector.Decompiler.Tests.csproj -c Release --configfile nuget.config --nologo -v:q` |
| Metadata test build | Pass: `dotnet build tests/ILInspector.Metadata.Tests/ILInspector.Metadata.Tests.csproj -c Release --configfile nuget.config --nologo -v:q` |
| CLI test build | Pass: `dotnet build src/dotnet-inspect.Tests/dotnet-inspect.Tests.csproj -c Release --configfile nuget.config --nologo -v:q` |
| Receiver fallback tests | Pass: `dotnet run --project src/ILInspector.Decompiler.Tests/ILInspector.Decompiler.Tests.csproj -c Release --no-build -- -method "*ExtensionRootSelection*"` |
| SourceLink identity tests | Pass: `dotnet run --project tests/ILInspector.Metadata.Tests/ILInspector.Metadata.Tests.csproj -c Release --no-build -- -method "*SourceDocumentIdentity*"` |
| Timeline analysis tests | Pass: `dotnet run --project src/dotnet-inspect.Tests/dotnet-inspect.Tests.csproj -c Release --no-build -- -method "*AnalysisTimeline*"` |
[...]

PR #2720, opened 2026-07-13
Harden signature guard census and receiver indexing

richlander:
## Summary

Evidence revision: `8b9c9a54`

- replace `TypeSpecGuard`'s public unpaired `Exit(int)` with a stack-only
  disposable scope and migrate nested TypeSpec re-entry in Metadata,
  MetadataPrimitives, and Instructions
- strengthen the provider-decode census with durable negative tests for the
  three reviewer evasions and bind every nested decode to its own matching
  guard invocation
- preserve generic arity across the DecompilerHarness receiver-type index and
  Roslyn receiver evidence so `Result` and `Result<T>` cannot collide without
  silently excluding generic receiver extensions

The token migration preserves the explicit signature rejection behavior added
on main: nested TypeSpec failures still report their typed rejection kind, and
existing degraded provider results remain unchanged.

## Safety

- scopes are stack-only, idempotent on repeated disposal, and enforce LIFO
  restoration of the thread-local depth and cumulative-byte budgets
- an empty TypeSpec blob is pinned through 300 enter/dispose cycles so zero
  length cannot be confused with an inactive scope
- the census binds the `TryEnter` handle, returned scope name, terminating
  failure branch, and enclosing `using` to the exact nested decode
- receiver keys canonicalize both `Result<int>` and metadata `Result`1` to the
  same arity-aware identity while keeping non-generic `Result` distinct;
  production-shaped CS1061 evidence pins generic interface receiver selection

## Validation

| Check | Result |
| --- | --- |
| Solution build | Pass |
| Metadata suite | 483 passed |
| Instructions suite | 121 passed |
| `FidelityCheckGeneratedFilterTests` | 24 passed |
| `ClosureDiagnosticEvidenceTests` | 37 passed |

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus 4.8 | CLEAN | Verified typed rejection preservation, scope lifetime, census probes, and arity normalization on `8b9c9a54`. |
| Gemini 3.1 Pro | CLEAN | Reproduced the post-rebase generic-receiver exclusion, verified its fix with independent probes, and found no remaining issues on `8b9c9a54`. |

Earlier review rounds found census identifier/boolean evasions and, after the
#2718 conflict resolution, an arity mismatch between metadata keys and Roslyn
receiver evidence. The fixes are pinned by negative syntax tests and a
production-shaped generic receiver test. Both reviewers returned clean on the
final exact head.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ClosureDiagnosticEvidenceTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ClosureDiagnosticEvidenceTests.cs
@@ -39,7 +39,7 @@ public class ClosureDiagnosticEvidenceTests
     [InlineData("CS1061", "class Receiver : System.Collections.IEnumerable { public System.Collections.IEnumerator GetEnumerator() => null; } class C { async System.Threading.Tasks.Task M() { await System.Threading.Tasks.Task.FromResult(new Receiver { 1 }); } }", "Add", "Receiver", null)]
     [InlineData("CS1061", "namespace @for { class @class {} class C { void M(@class value) { value.Missing(); } } }", "Missing", "@for.@class", null)]
     [InlineData("CS1061", "namespace @await { class Receiver {} class C { void M(Receiver value) { value.Missing(); } } }", "Missing", "@await.Receiver", null)]
-    [InlineData("CS1061", "namespace N { class Outer<T> { public class Inner {} } class C { void M(Outer<int>.Inner value) { value.Missing(); } } }", "Missing", "N.Outer.Inner", null)]
+    [InlineData("CS1061", "namespace N { class Outer<T> { public class Inner {} } class C { void M(Outer<int>.Inner value) { value.Missing(); } } }", "Missing", "N.Outer`1.Inner", null)]
     [InlineData("CS0117", "class Receiver {} class C { void M() { Receiver.Missing(); } }", "Missing", "Receiver", null)]
     [InlineData("CS0117", "class Receiver {} class C { Receiver M() => new Receiver { Missing = 1 }; }", "Missing", "Receiver", null)]
     public void Extract_UsesStructuredSyntaxAndSemanticEvidence(
@@ -128,7 +128,7 @@ public class ClosureDiagnosticEvidenceTests
                 "ReceiverEvidence.Derived",
                 "ReceiverEvidence.Base",
                 "System.Object",
-                "ReceiverEvidence.IReceiver",
+                "ReceiverEvidence.IReceiver`1",
             ],
             reference.CompatibleReceiverTypes);
         Assert.True(reference.CompatibleReceiverTypesComplete);
--- a/src/ILInspector.Decompiler.Tests/FidelityCheckGeneratedFilterTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FidelityCheckGeneratedFilterTests.cs
@@ -492,6 +492,73 @@ public class FidelityCheckGeneratedFilterTests
     }
 
     [Fact]
+    public void ExtensionRootSelection_UsesArityAwareRoslynReceiverEvidence()
+    {
+        var assemblyPath = CompileFixture("""
+            namespace ExtensionReceiverArityEvidenceFixture;
+
+            public interface IReceiver<T> { }
+            public class Result<T> : IReceiver<T> { }
+
+            public static class GenericReceiverExtensions
+            {
+                public static void Add<T>(this IReceiver<T> receiver, int value) { }
+            }
+            """);
+        try
+        {
+            var tree = CSharpSyntaxTree.ParseText(
+                """
+                class Consumer
+                {
+                    void Use(
+                        ExtensionReceiverArityEvidenceFixture.Result<int> value)
+                        => value.Add(1);
+                }
+                """,
+                cancellationToken: TestContext.Current.CancellationToken);
+            var compilation = CSharpCompilation.Create(
+                "extension-receiver-arity-evidence",
+                [tree],
+                [
+                    MetadataReference.CreateFromFile(typeof(object).Assembly.Location),
+                    MetadataReference.CreateFromFile(assemblyPath),
+                ],
+                new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary));
+            var diagnostic = Assert.Single(
+                compilation.GetDiagnostics(TestContext.Current.CancellationToken),
+                candidate => candidate.Id == "CS1061");
+            var reference = Assert.IsType<ClosureDiagnosticReference>(
+                ClosureDiagnosticEvidence.Extract(
+                    diagnostic,
+                    compilation.GetSemanticModel(tree)));
+
+            Assert.Equal(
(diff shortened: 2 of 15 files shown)


richlander:

Exact head: `9b755548`

| Reviewer | Exact-head result |
| --- | --- |
| Claude Opus 4.8 | CLEAN |
| Gemini 3.1 Pro | CLEAN |

### First-round finding and resolution

Gemini found two actionable census evasions on `4f41c6e6`:

1. Decode-name matching used Roslyn `Identifier.Text`, allowing verbatim
   `@DecodeSignature` and Unicode-escaped identifiers to escape the deny-list.
2. Prescan classification searched all condition descendants, allowing
   `IsSafeToDecode(...) || true` to satisfy the census while bypassing the
   runtime guard.

Both were reproduced and fixed in `9b755548`:

- identifier comparisons now use `ValueText`;
- prescan matching accepts only the exact guard invocation or logical-AND
  compositions where guard truth is necessary;
- negative self-tests pin verbatim identifiers, Unicode identifiers, and the
  `guard || true` laundering shape.

### Exact-head agreement

Opus and Gemini independently verified the two fixes with syntax probes and
focused test runs. Both also re-audited disposable-scope lifetime, complete
TypeSpec migration across MetadataPrimitives/Metadata/Instructions, degraded
signature behavior, per-invocation handle/scope binding, and generic-arity
receiver indexing. Neither found an outstanding correctness, security, or
compatibility issue.

There were no dismissed findings and no non-actions requiring a resolution
commit.

richlander:
## Post-rebase exact-head reconciliation

Exact head: `442a12f1`  
Base: `d6e5f942`

The branch was rebased after #2715 made signature-limit degradation explicit.
The only conflicts were the nested TypeSpec providers in `PointerDetector` and
`SignatureSpellability`. Resolution preserved the new
`PointerDetection.Degraded` / `SpellabilityEvidence.Degraded` fallbacks while
applying the disposable `using (scope)` lifetime.

| Reviewer | Result |
| --- | --- |
| Claude Opus 4.8 | CLEAN |
| Gemini 3.1 Pro | CLEAN |

Both reviewers independently verified the degradation reconciliation, complete
guard migration, scope lifetime/accounting, all census evasion fixes, and
generic-arity receiver indexing on the exact rebased head. No new findings,
dismissals, or non-actions remain.

Local exact-head evidence:

- solution build passed;
- Metadata: 477 passed;
- Instructions: 116 passed;
- focused DecompilerHarness: 23 passed.

richlander:
All merge-blocking validation is complete on exact head `442a12f1`: required CI is green and both isolated cross-model adversarial reviews are CLEAN. **Ready to merge.**

richlander (commented):
## Independent review — VERDICT: MINOR-CONCERNS (sound in isolation; heavy overlap with #2719 and #2713)

The code itself is correct and self-consistent — no blocking defect. The concern is that it overlaps two sibling PRs on the same code, and they can't all land.

**Verified:**
- **Disposable-scope accounting** correct: LIFO enforced by capturing `_depth` after increment and asserting `s_depth == _depth` on `Exit` (out-of-order/double-dispose **throw** rather than silently mis-count); empty-blob not conflated with inactive scope (`_active` set independent of blob length; 300-cycle test).
- **Migration completeness** compiler-guaranteed: `TryEnter` now returns `out Scope` and `Exit` is `private`, so a missed site fails to build. All 10 former `Exit(int)` sites migrated.
- **Census** strengthened with 6 evasion negatives (alias, newline-split, comment token, verbatim `@`, `D` Unicode, `|| true`) bound per-invocation to a matching `TryEnter`.
- **Receiver-arity normalizer** is arity-preserving and correct (`Result` vs `` Result`1 `` land on distinct keys).
- Product path stays SRM-only / AOT / Roslyn-free (census + `FidelityCheck` are test/tooling only).

### ⚠️ Two overlaps
- **vs #2719 (guard/census):** #2720's guard work is a **strict subset** of #2719, which also extends the census to Analysis + Decompiler, adds the Analysis structural prescan + the ARRAY-size runtime pin, and uses a more robust `readonly ref struct` token scope. #2720's `Scope` is a mutable `ref struct` whose `Exit` *throws* on out-of-order (and a struct-copy double-dispose throws — the test only covers same-variable double-dispose), which is more fragile than #2719's no-op token approach. These two will hard-conflict.
- **vs #2713 (#2687 receiver index):** both fix the same arity-collision defect **incompatibly** — this PR is arity-*preserving* (no fallback; `UsedFallback==false`), #2713 is arity-*erasing* + visible fallback (`UsedFallback==true`). Their `FidelityCheck.cs` edits and new `[Fact]`s (same location) conflict and cross-break each other's tests.

**On the merits, #2720's contributions split cleanly:** the `.ValueText` census hardening is the piece #2719 is missing (worth porting there), and the **arity-preserving receiver normalizer is the more precise fix** than #2713's fallback (it selects the correct root instead of conservatively over-including). Suggested landing: take guard/census from #2719 (+ port `.ValueText`), keep #2720's arity-preserving normalizer as its own PR, drop #2713's competing receiver fix, and close the remainder of #2720 as superseded.

---
*Independent review by Claude Code — static (diff + source at head `442a12f`; no local build/test run).*

---

richlander:
## Post-conflict adversarial review reconciliation

Exact head: `8b9c9a54`
Base: `a310a925`

The branch was rebased after #2718 introduced explicit signature rejection.
Conflict resolution preserved `SignatureDecodeResult<T>`,
`SignatureDecodeRejectionKind`, and `GuardedSignatureDecoder` while migrating
nested TypeSpec decodes to the disposable scope.

### First post-conflict pass

| Reviewer | Result |
| --- | --- |
| Claude Opus 4.8 | CLEAN |
| Gemini 3.1 Pro | One actionable finding |

Gemini reproduced a silent exclusion for generic receiver extensions:
metadata receiver keys now preserved arity, but production Roslyn CS1061
receiver and hierarchy evidence remained arity-free. A complete hierarchy could
therefore drop an `IEnumerable<T>`-style extension root without fallback.

The finding was reproduced against a clean exact-head checkout and fixed in
`8b9c9a54`. Receiver-specific Roslyn evidence now carries metadata arity, while
non-receiver diagnostic names retain their prior form. A production-shaped
CS1061 regression proves that a generic interface receiver selects the exact
extension root without fallback.

### Final exact-head pass

| Reviewer | Result |
| --- | --- |
| Claude Opus 4.8 | CLEAN |
| Gemini 3.1 Pro | CLEAN |

Both reviewers independently re-audited the #2718 typed-rejection
reconciliation, disposable scope accounting, census evasions and near misses,
and generic receiver normalization on `8b9c9a54`. There were no dismissed
findings or unresolved non-actions.

Local exact-head evidence:

- solution build passed;
- Metadata: 483 passed;
- Instructions: 121 passed;
- focused DecompilerHarness receiver/evidence tests: 61 passed.

Required CI is green on the exact head.

richlander:
All merge-blocking validation is complete on exact head `8b9c9a54`: required CI is green and both isolated cross-model adversarial reviews are CLEAN after the conflict resolution and follow-up fix. **Ready to merge.**

Issue #2724, opened 2026-07-13
Performance Triage: model cross-method repetition without relabeling static heat


richlander:
## Summary

A Performance Triage health pass against the pinned corpus and the original
[Aspire Dashboard findings](https://gist.github.com/richlander/3803394c9634db2883ec2b07c0a19998/raw/2f4028ef9f4460a540f598c786a905f375c16c7c/perf-finds-for-david.md)
confirmed one remaining gap: ordinary once-per-call allocations can be repeated
by a caller loop, but the triage row has no first-class cross-method repetition
evidence.

The composed-scan follow-up now recovers the algorithmic case:
`OtlpSpan.GetParentSpan()` surfaces as `scan-method-in-loop-call` because
`Where(...).FirstOrDefault()` is invoked from `OtlpTrace.CalculateDepth`'s loop.
The second reported case still relies on a manual caller-graph receipt:

- `ColorGenerator.GetColorIndex` surfaces as a low-confidence
  `instance-method-group-delegate` with local `Loop` empty;
- `Caller Graph` shows the call chain reaches a generated render callback marked
  `loop call`;
- the delegate allocation is therefore repeated by that caller traversal, but
  this objective cross-method multiplicity is not present on the triage row.

This must **not** become a claim of runtime heat. As
`docs/design/allocation-triage-prefilters.md` states, realized call frequency is
dynamic. The static fact available here is narrower: a concrete call-graph path
contains a loop call, so one invocation of that caller can repeat the callee.

## Constraints

- Preserve `AllocationMultiplicity` and the `Loop` column as the allocation's
  intraprocedural, per-invocation semantics (#2177). Do not overload either with
  caller evidence.
- Do not promote `Root Reach` or caller-loop reach into measured frequency.
- Reuse #2716's `Candidate` / native Finding provenance for exact rows; do not
  create a parallel identity mechanism.
- Keep the product SRM-only, NativeAOT-friendly, Roslyn-free, and free of
  inspected-assembly loading.
- Measure before changing default ranking or confidence.

### 1. Measurement-only caller-loop census

Add an AnalysisHarness readout over triage opportunities that records:

- no caller-loop witness;
- direct loop caller;
- bounded transitive loop caller;
- nearest loop-call depth;
- representative witness path;
- shape / confidence / local multiplicity cross-tabs.

Run it over the six-library #1974 corpus plus the broader 44-assembly performance
corpus. This slice changes no product output and establishes whether the signal
[...]

PR #2725, opened 2026-07-13
Detect composed LINQ scans across caller loops

richlander:
## Summary

- recognize an immediately consumed trusted `Enumerable.Where` / `OfType` chain
  as a linear scan when followed by a parameterless terminal;
- feed that exact scan evidence into the existing
  `scan-method-in-loop-call` detector;
- use decoded `DirectCall.ReturnAddress`, rather than IL-width assumptions, for
  the adjacency proof;
- add adversarial fixtures for intervening IL, unrelated terminal sources, and
  `Select(...).First()` (which can remain O(1));
- document the separate low-confidence algorithmic-shape query in the README and
  Performance Triage skill.

## Why

The original
[Aspire Dashboard performance report](https://gist.github.com/richlander/3803394c9634db2883ec2b07c0a19998/raw/2f4028ef9f4460a540f598c786a905f375c16c7c/perf-finds-for-david.md)
identified `OtlpSpan.GetParentSpan()` as important, but Performance Triage only
surfaced its lower-value delegate row. The method's real shape is a
`Where(...).FirstOrDefault()` linear search called from
`OtlpTrace.CalculateDepth`'s loop.

After this change:

```text
Member: OtlpSpan.GetParentSpan()
Shape: scan-method-in-loop-call
Evidence: Linearly scans a sequence (Enumerable.Where+FirstOrDefault); invoked inside a loop by OtlpTrace::CalculateDepth
Confidence: low
Loop: loop
Candidate: pt~9a64661457afc323
Provenance: aggregate
```

The confidence intentionally remains low: the caller-loop edge is objective
static evidence, but it does not prove runtime frequency or that the scanned
sequence scales with the loop. The supplied report provides the separate dynamic
confirmation for this application (approximately 10x render improvement).

## Precision gates

A composed scan is accepted only when:

1. producer and terminal resolve to trusted `System.Linq.Enumerable` APIs;
2. the producer is `Where` or `OfType`;
3. the terminal has exactly one source parameter; and
4. the terminal instruction starts at the producer call's decoded return
   address.

Predicate terminals remain covered by the existing membership-scan path.
`Select(...).First()` is excluded because projection of the first element does
not require a linear search.

## Corpus evidence

Exact Performance Triage row counts on the pinned six-library corpus:

| Library | Base | PR | Delta |
|---|---:|---:|---:|
| Aspire.Hosting 13.4.6 | 553 | 558 | +5 |
| System.Text.Json 10.0.9 | 132 | 132 | 0 |
| Newtonsoft.Json 13.0.4 | 345 | 346 | +1 |
| Serilog 4.3.1 | 92 | 92 | 0 |
| Polly 8.7.0 | 434 | 434 | 0 |
| AutoMapper 16.1.1 | 199 | 199 | 0 |

All six additions were decompiled and classified as genuine immediately consumed
filtering/search chains. No existing row disappeared or changed confidence.

An interleaved fixed-assembly index-cost comparison found no slowdown: median CPU
was approximately 4.79s on the merge-base build and 4.49s on this branch.

## Validation

- Release solution build
- `ILInspector.Analysis.Tests`: 435 passed
- focused Performance Triage CLI tests: 19 passed
- fixed-assembly index-build guards: 3 passed
- analysis fixture ledger: 22/22 targets passed
- pay-dirt recall: 3/3
- changed Markdown lint and diff checks

## Adversarial review

The pre-merge head and the final head after merging #2716 were independently
reviewed by Claude Opus 4.8 and Gemini 3.1 Pro. Both final reviews were **CLEAN**.
Reviewer guidance to avoid instruction-width coupling was adopted in `64e67795`
by using the already decoded call return address.

Follow-up cross-method repetition modeling is sliced in #2724. That issue keeps
local allocation multiplicity separate from caller evidence and treats runtime
heat as dynamic evidence.


richlander:

Two independent model families reviewed the detector against its stated identity
gates and adversarial near misses, then reviewed the refreshed head after #2716
was merged:

- **Claude Opus 4.8:** final result **CLEAN**. It found no reproducible
  correctness, compatibility, or analysis-cost defect in the trusted-API,
  signature, and exact-adjacency gates.
- **Gemini 3.1 Pro:** final result **CLEAN**. It found no reproducible false join
  across intervening IL, unrelated terminal sources, or the intentionally
  excluded `Select(...).First()` shape.

A non-blocking hardening observation from the review pass was adopted in
[`64e67795`](https://github.com/richlander/dotnet-inspect/commit/64e67795): the
adjacency check now uses the call's already decoded `DirectCall.ReturnAddress`
instead of assuming an instruction width. Both reviewers then returned **CLEAN**
on that change and again on the final merged head `1eb63ac3`.

There are no dismissed or outstanding findings, so no other resolution commit
applies. The remaining product opportunity is separately scoped in #2724 rather
than treated as a defect in this PR.

richlander:
**Ready to merge.** All merge-blocking local validation and CI checks are complete, and the refreshed head has two final CLEAN adversarial reviews.

Issue #2726, opened 2026-07-13
RTS ctor chain: faithful-args gate misses assignable (non-identity) arguments (string[] -> IEnumerable<string>)


richlander:
## Summary

The RTS constructor-chain gate added in #2708 (issue #2678) emits a `: this(args)` initializer only when the arguments are **faithful** (each argument's IR result type *structurally equals* its parameter type) or the chain is **unique-applicable** by arity. This is sound but conservative: it declines chains where an argument is *assignable* to the parameter but not structurally identical.

The concrete miss observed on NuGet.Versioning: 3 `SemanticVersion` constructors chain `this(version, ParseReleaseLabels(releaseLabel), metadata)` where `ParseReleaseLabels` returns `string[]` but the parameter is `IEnumerable<string>`, alongside a same-arity `(Version, string, string)` sibling. The argument is:

- **not faithful** — `string[]` does not `TypeRef.Equals` `IEnumerable<string>`; and
- **not unique-applicable** — the arity is shared with the `(Version, string, string)` sibling.

So the gate strips the initializer and these rows stay at the pre-PR `OpcodeDiff` baseline (they were `OpcodeDiff` before #2708, still `OpcodeDiff` after — **no regression**). This costs 3 potential `OpcodeDiff -> Exact` wins (21 -> 18 on the NuGet.Versioning probe).

## Why it is safe but unproven

`string[]` genuinely cannot bind to the sibling's `string` parameter, so `: this(version, ParseReleaseLabels(...), metadata)` would in fact bind unambiguously to the intended `(Version, IEnumerable<string>, string)` ctor. But proving that requires reasoning about **implicit reference/array-covariance conversions** and **overload applicability among siblings**, which the current structural-equality check deliberately avoids (RTS is SRM-only, Roslyn-free on the product path).

## Proposed enhancement

Extend `ChainArgumentsAreFaithful` / the unique-applicable-candidate check with a bounded **assignability** model so an argument counts as faithful when its result type is implicitly convertible to the parameter type *and* no other reconstructed sibling constructor is applicable to the argument list under those conversions. Minimum viable coverage:

- array/`IEnumerable<T>` (and `IReadOnlyList<T>`/`ICollection<T>`) covariance,
- reference up-conversions already excluded today (keep them lossy only when a competing sibling could also bind),
- guard against widening it into an unsound state — the failure direction must stay "can't prove -> strip".

## Acceptance

[...]

PR #2730, opened 2026-07-13
RTS ctor chain: emit initializer for array-covariant assignable args (#2726)

richlander:
- Advances #2726
- Changes RTS constructor-chain gate to emit `: this(args)` when an argument is array-covariant-assignable (not identity-equal) to its parameter and the chained-to ctor is provably the unique same-arity binder. Harness-only (`tools/DecompilerHarness`); product decompiler output unchanged.
- Evidence revision: `9ef5ca44`

## Change

**Conclusion:** **PASS** — the faithful/unique-arity gate from #2708 could not check `string[]`→`IEnumerable<string>` chain args with a same-arity sibling, so 3 `SemanticVersion` chains stayed `OpcodeDiff`; a bounded, fail-closed array-covariance disjunct proves unique applicability and moves them to `Exact` (NuGet.Versioning probe 18→21 wins), with `RecompileFail` unchanged at 63.

Before (SemanticVersion, `this(version, ParseReleaseLabels(...), metadata)`):

```text
: this(...) stripped — string[] arg is not TypeRef.Equals IEnumerable<string>
and shares arity with the (Version, string, string) sibling  ->  OpcodeDiff
```

After:

```text
: this(version, ParseReleaseLabels(releaseLabel), metadata) emitted  ->  Exact
```

## Scope and safety

- **Root cause:** the gate treats an argument as bindable only when its IR result type *structurally equals* the parameter type (faithful) or the chain is unique by arity. `string[]` does not `TypeRef.Equals` `IEnumerable<string>`, and the arity is shared, so the chain is stripped even though `string[]` genuinely cannot bind the `(Version, string, string)` sibling.
- **Fix shape:** RTS orchestration only. New additive OR-disjunct `ChainArgumentsBindViaArrayCovariance`: every arg is identity- or array→covariant-interface-assignable (same element; the 5 covariant `System.Collections.Generic` interfaces) to its chained-to parameter, ≥1 is the non-identity array case, and every *other* metadata ctor that could bind the argument count by arity is provably inapplicable at an array position. `ArrayCanConvertTo` over-approximates convertibility, so a `false` result *proves* non-applicability — emit only when the chained-to ctor is the unique non-excluded same-arity binder; any uncertainty strips.
- **Product impact:** none. `ReturnToSenderTypePlanner` only shapes the RTS compile-back shell; product decompiler render is untouched.
- **Scope boundary:** limited to array→covariant-interface. General reference up-conversions and user-defined conversions remain a frontier (still stripped).
- **RTS boundary:** RTS only orchestrates closure + Roslyn + opcode compare; the emitted `: this(args)` is round-trip-validated by recompilation.

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass (`DecompilerHarness`, `ILInspector.Decompiler.Tests`) |
| Focused tests | `ReturnToSenderPrototypeTests` 111/0 (incl. 1 positive canary + 3 adversarial negatives) |
| Decompiler fast suite | `-notrait "Speed=Slow"` 2409/0 |
| Targeted compile-back | NuGet.Versioning rts-parity 136→**139 exact**, 12→**9 opcode diffs**, **63 RF unchanged** |
| Real witness | `NuGet.Versioning.SemanticVersion::.ctor` (×3 chains) moved `OpcodeDiff` → `Exact` |

## Compile-back quality

**Conclusion:** **PASS** — the 3 changed `SemanticVersion` chains move `OpcodeDiff`→`Exact` and recompile to identical IL; no row regressed (RecompileFail 63 → 63, checked population 211 unchanged).

### Broad assembly context

Run: `NuGet.Versioning.dll` (net8.0), `--corpus-fidelity-cap 500 --corpus-fidelity-oracle return-to-sender --sequential`.

```text
Fidelity (rts-parity, cap 500) over 211 checked

  before:  136 exact, 12 opcode diffs, 63 recompile failures, 0 context failures
  after :  139 exact,  9 opcode diffs, 63 recompile failures, 0 context failures
```

**Broad verdict:** **PASS** — +3 exact, −3 opcode diff, RF/context flat. The broad cap is the load-bearing evidence named in the issue (the 21-wins target).

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| General reference up-conversions (non-array) | Left as frontier | Requires broader overload-applicability reasoning; stays stripped ("can't prove → strip"). |
| User-defined / boxing conversions | Not changed | Out of scope; unsound to admit without full conversion model. |
| `params`-array same-arity siblings | Correctly strips | `string[]` equals the params array type at the params slot, so the sibling is not excludable → strip (adversarial negative `...TargetHasParamsConstructor`). |

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| _(pending)_ GPT-5.5 | | isolated worktree |
| _(pending)_ Gemini 3.1 Pro | | isolated worktree |

**Review conclusion:** **REVIEW** — two adversarial reviews requested per decompiler-PR policy (non-Claude roster); results will be reconciled here.

## Validation

```bash
dotnet build tools/DecompilerHarness -c Release --configfile nuget.config
dotnet build src/ILInspector.Decompiler.Tests -c Release --configfile nuget.config
  "$HOME/.nuget/packages/nuget.versioning/7.3.0/lib/net8.0/NuGet.Versioning.dll" \
  --emit-corpus-snapshot /tmp/out.json --compile-cap 0 --corpus-fidelity-cap 500 \
  --corpus-fidelity-oracle return-to-sender --sequential --max-examples 3
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -1042,6 +1042,75 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
+    public void CompileBackTargets_DropsInitializerWhenArrayCovariantArgumentBindsSystemNamespaceUserConversion()
+    {
+        // Issue #2726 guard (adversarial, from PR #2730 round-2 review): a user type
+        // declared in a `System` namespace is NOT corelib and may declare an
+        // implicit conversion operator from an array. Excluding a competitor by
+        // namespace string would be unsound; corelib membership is decided by
+        // assembly identity (TypeRef.Assembly == "corelib"), so `System.MySink`
+        // (defined in the target assembly) stays in play and the chain strips. The
+        // body forces both the conversion operator and the competitor ctor into the
+        // reconstructed shell; [OverloadResolutionPriority(-1)] keeps the original
+        // chain bound to the IEnumerable ctor.
+        var assemblyPath = CompileFixture("""
+            using System.Collections.Generic;
+            using System.Runtime.CompilerServices;
+
+            namespace System
+            {
+                public class MySink
+                {
+                    public static implicit operator MySink(string[] value) => new MySink();
+                }
+            }
+
+            public class C
+            {
+                public C(string tag)
+                    : this(Parse(tag), tag.Length)
+                {
+                    _ = (System.MySink)Parse(tag);
+                    _ = new C(default(System.MySink), 0);
+                }
+
+                public C(IEnumerable<string> labels, int count)
+                {
+                    Count = count;
+                }
+
+                [OverloadResolutionPriority(-1)]
+                public C(System.MySink sink, int count)
+                {
+                    Count = count;
+                }
+
+                public int Count { get; }
+
+                private static string[] Parse(string tag) => tag.Split('.');
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("C", ".ctor", 0,
+                    "(corelib:System.String) -> corelib:System.Void")]));
+
+            // The load-bearing invariant is that the gate STRIPS the unsound chain.
+            // This fixture is deliberately pathological (a user type in `namespace
+            // System` plus an implicit array operator), which RTS cannot round-trip
+            // as a whole shell, so the overall status is not asserted; emitting the
+            // initializer would fail or bind wrong, so stripping is the sound choice.
+            Assert.DoesNotContain(": this(", result.Source);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
     public void CompileBackTargets_UsesTargetBackingFieldWriteForConstructorAssignment()
     {
         var assemblyPath = CompileFixture("""
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -778,6 +778,210 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
+    public void CompileBackTargets_PreservesConstructorChainWhenArgumentIsArrayCovariantToInterface()
+    {
+        // Issue #2726: the `(Version, string, string)` ctor chains to
+        // `(Version, IEnumerable<string>, string)` passing a `string[]` argument.
+        // `string[]` is implicitly convertible to `IEnumerable<string>` by array
+        // covariance but is NOT structurally identical to it, so the faithful
+        // check (identity only) declines the chain; the two ctors share arity, so
+        // the unique-arity check declines too. `string[]` cannot bind to the
+        // sibling's second `string` parameter, so the chain binds unambiguously to
+        // `(Version, IEnumerable<string>, string)` and must be preserved.
+        var assemblyPath = CompileFixture("""
+            using System;
+            using System.Collections.Generic;
+
+            public class Ver
+            {
+                public Ver(Version version, string release, string metadata)
+                    : this(version, ParseLabels(release), metadata)
+                {
+                }
+
+                public Ver(Version version, IEnumerable<string> releaseLabels, string metadata)
+                {
+                    Labels = releaseLabels;
+                    Metadata = metadata;
+                }
+
+                public IEnumerable<string> Labels { get; }
+
+                public string Metadata { get; }
+
+                private static string[] ParseLabels(string release) => release.Split('.');
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("Ver", ".ctor", 0,
+                    "(corelib:System.Version, corelib:System.String, corelib:System.String) -> corelib:System.Void")]));
+
+            Assert.Equal(FidelityCheck.CompileBackStatus.Exact, result.Status);
+            Assert.Contains(": this(", result.Source);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
+    public void CompileBackTargets_DropsInitializerWhenArrayCovariantArgumentAlsoBindsCovariantSibling()
+    {
+        // Issue #2726 guard: `string[]` is array-covariant to BOTH the chained-to
+        // `IReadOnlyCollection<string>` parameter and the same-arity sibling's
+        // `IEnumerable<string>` parameter, so the chained-to constructor is not the
+        // unique applicable candidate at the array position. The `new C(...)` in the
+        // body forces the sibling into the shell (defeating the unique-arity path),
+        // so the assignability gate must decide — and must strip rather than emit a
+        // chain that could bind to the sibling.
+        var assemblyPath = CompileFixture("""
+            using System.Collections.Generic;
+
+            public class C
+            {
+                public C(string tag)
+                    : this(Parse(tag), tag.Length)
+                {
+                    _ = new C((IEnumerable<string>)null, 0);
+                }
+
+                public C(IReadOnlyCollection<string> labels, int count)
+                {
+                    Count = count;
+                }
+
+                public C(IEnumerable<string> labels, int count)
+                {
+                    Count = count;
+                }
+
+                public int Count { get; }
+
+                private static string[] Parse(string tag) => tag.Split('.');
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("C", ".ctor", 0,
+                    "(corelib:System.String) -> corelib:System.Void")]));
+
+            Assert.NotEqual(FidelityCheck.CompileBackStatus.RecompileFail, result.Status);
+            Assert.DoesNotContain(": this(", result.Source);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
+    public void CompileBackTargets_DropsInitializerWhenArrayCovariantArgumentAlsoBindsObjectSibling()
+    {
+        // Issue #2726 guard: `string[]` converts to `object`, so a same-arity
+        // sibling taking `object` at the array position cannot be excluded. The
+        // covariance model over-approximates convertibility (so exclusion is only
+        // ever a proof of non-convertibility); an unprovable competitor strips. The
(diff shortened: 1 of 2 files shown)


richlander:
## Adversarial review — round 1 (reconciled)

Two independent reviewers (isolated worktrees, non-Claude roster):

| Reviewer | Round 1 | Finding |
| --- | --- | --- |
| GPT-5.5 | 1 High | `ArrayCanConvertTo` misses `T[]`→`Span<T>`/`ReadOnlySpan<T>` (built-in) conversions |
| Gemini 3.1 Pro | 1 High | Same: misses `T[]`→`Span`/`ReadOnlySpan`/`Memory`/`ReadOnlyMemory` **and** user-defined array-conversion operators |

**Agreement:** both independently converged on the *same* soundness hole, so I treated it as high-confidence and fixed it.

### Verified finding

`ArrayCanConvertTo` returned `false` for `T[]`→`Span<T>`/`ReadOnlySpan<T>`/`Memory<T>`/`ReadOnlyMemory<T>` and for user-defined types with an implicit conversion operator from an array. Because a `false` result is used as **proof** of non-convertibility to *exclude* a competing constructor, such a same-arity sibling was wrongly excluded → unsound `: this(args)` emit. Reachable when a library uses `[OverloadResolutionPriority]` to keep an `IEnumerable<T>` overload selected while also offering a first-class `ReadOnlySpan<T>` overload (RTS does not re-emit the attribute, so a recompiled chain would prefer the span ctor).

**Reproduced against the clean PR head** (both reviewers agreed; I confirmed independently): the new regression fixture emits the unsound `: this(Parse(tag), tag.Length)` on the pre-fix code and strips after the fix — with the fixture's chain confirmed to bind to `IEnumerable` (i.e. the covariance emit path is exercised).

### Fix — commit `f437e983`

`ArrayCanConvertTo` now fails closed (exactly Gemini's prescription):

- `TypeRefKind.GenericInstance` → `true` (can't-rule-out) for **any** generic instance that is not a covariant `System.Collections.Generic` interface, so `Span`/`ReadOnlySpan`/`Memory`/`ReadOnlyMemory` and user generics are never excluded.
- `IsArrayCompatibleNonGenericType` → `true` for any non-`System.*` (user-defined) definition; returns `false` (provably non-convertible) only for enumerable corelib types with no array conversion.

This still covers the SemanticVersion win (competitors excluded via `int` and `System.String`, both provably non-convertible corelib types), so the probe is unchanged: **139 exact / 9 opcode diffs / 63 recompile failures**.

New adversarial negative: `CompileBackTargets_DropsInitializerWhenArrayCovariantArgumentAlsoBindsSpanSibling`.

### Evidence after fix

| Check | Result |
| --- | --- |
| `-method "*ArrayCovariant*"` | 5/5 (canary + 4 negatives incl. new span) |
| `ReturnToSenderPrototypeTests` full class | 112/0 |
| Pre-fix witness | span negative FAILS on `9ef5ca44` (unsound emit), passes on `f437e983` |
| NuGet.Versioning rts-parity | 139 exact, 9 opcode diffs, 63 RF (unchanged) |

Round 2 (re-review of the fixed head `f437e983`) is in progress with both reviewers; I'll post the CLEAN confirmations before requesting merge.

richlander:
## Adversarial review — round 2 (reconciled)

Both reviewers re-verified the round-1 `Span`/`ReadOnlySpan` fix (`f437e983`) as closed, and again **independently converged on the same follow-up finding** — and prescribed the same fix.

| Reviewer | Round 2 | Finding |
| --- | --- | --- |
| GPT-5.5 | 1 High | `System`-namespace **user** type with `implicit operator MySink(string[])` still wrongly excluded (namespace-string proxy is not a sound corelib test) |
| Gemini 3.1 Pro | 1 Med/High | Same: a user type in `namespace System` violates the fail-closed invariant |

### Verified finding

The round-1 fix used `IsSystemNamespace(type.Namespace)` to rule a competitor "provably non-convertible corelib type." But a user assembly can declare a type in a `System` namespace with an `implicit operator T(U[])`; the namespace string then falsely proves non-convertibility → unsound emit. Reachable via `[OverloadResolutionPriority(-1)]` keeping the `IEnumerable<string>` overload selected while RTS re-emits the conversion operator the ctor body consumes. **Reproduced on `f437e983`** (both reviewers; I confirmed the new regression test emits the unsound `: this(...)` there).

### Fix — commit `542b80b8`

Decide corelib membership by **assembly identity**, not namespace string:

- `IsArrayCompatibleNonGenericType` returns `false` (provably non-convertible) only when `type.Assembly == TypeRef.CoreLibrary` (`"corelib"`) and the type is not array-convertible; **any** non-corelib definition (including one spelled in a `System` namespace) returns `true` and stays in play.

`int`/`System.String`/`System.Version` canonicalize to `corelib`, so the SemanticVersion win is retained — probe unchanged at **139 exact / 9 opcode diffs / 63 recompile failures**.

New adversarial negative: `CompileBackTargets_DropsInitializerWhenArrayCovariantArgumentBindsSystemNamespaceUserConversion` (the reviewers' `System.MySink` attack; verified to emit unsoundly pre-fix, strips after).

### Evidence after fix

| Check | Result |
| --- | --- |
| `ReturnToSenderPrototypeTests` full class | 113/0 |
| Pre-fix witness | `System.MySink` negative FAILS on `f437e983` (unsound emit), strips on `542b80b8` |
| NuGet.Versioning rts-parity | 139 exact, 9 opcode diffs, 63 RF (unchanged) |

Round 3 (final re-verify of `542b80b8`) is in progress with both reviewers.

richlander:
## Adversarial review reconciliation — rounds 3–5 (final)

Two reviewers, both from a different model family than the author (Claude): **GPT-5.5** and **Gemini 3.1 Pro**, in isolated detached worktrees at each successive PR head. Rounds 1–2 were reconciled earlier in this thread; this comment covers rounds 3–5 and the final verdicts.

### Round 3 (head `542b80b8`)
- **Gemini: CLEAN.**
- **GPT-5.5: High** — `TypeRefDecoder.Canonical` maps a facade-simple-named *inspected* assembly (`System.Runtime`/`mscorlib`/`netstandard`/`System.Private.CoreLib`/`System.Runtime.Extensions`) to `corelib`, *including that assembly's own `TypeDefinition`s*. So an inspected `System.MySink` with an implicit array operator decodes as `corelib:System.MySink`, and the assembly-identity corelib exclusion wrongly ruled it out as a competitor → unsound `: this(args)` emit.
- **Verified real** (decoder claim confirmed at `TypeRefDecoder.cs`). **Fixed in `3f2dc465`** (harness-only): `ChainArgumentsBindViaArrayCovariance` collects the inspected assembly's `TypeDefinition` names and `CompetitorExcludedAtPosition` never excludes a competitor whose receiving parameter is an inspected definition. Genuine corelib params (`int`/`string`) are `TypeReference`s, so the 3 SemanticVersion wins are retained. New regression test `...BindsFacadeAssemblyUserConversion` (assembly `System.Runtime`), verified to **FAIL on pre-fix `542b80b8`**.

### Round 4 (head `3f2dc465`)
- **BOTH reviewers independently found the same High** — the round-3 inspected-name key was built flat (`namespace + "." + name`), but a *nested* `TypeDefinition` row has an empty namespace + simple name, while the decoder shapes nested `TypeRef`s as outermost-namespace + `Declaring+Nested`. So `IsInspectedDefinition` never matched a nested inspected type; in a facade-named assembly that nested user type canonicalizes to `corelib` and was wrongly excluded → unsound emit re-opened.
- **Verified real. Fixed in `d966a657`** (harness-only): the inspected-name key is now built by `InspectedDefinitionKey(reader, handle)`, which walks `GetDeclaringType()` and joins enclosing types with `+`, mirroring the decoder for arbitrarily deep nesting. New regression test `...BindsNestedFacadeAssemblyUserConversion` (nested `System.Outer.MySink` in assembly `System.Runtime`), verified to **FAIL on pre-fix `3f2dc465`**. Gemini independently verified multi-level `Outer+Mid+Inner+MySink` nesting is now kept in play.

### Round 5 (head `d966a657` → `17706427`)
- **Gemini: CLEAN.**
- **GPT-5.5: Medium** — the inspected-definition guard only covers `reader.TypeDefinitions`, so a *referenced* user assembly whose simple name is itself a corelib facade (a dependency literally named `System.Runtime`) still decodes as `corelib` and is not saved by the guard. GPT itself noted it is "likely neutralized as a recompile failure rather than a false Exact."
- **Resolution: accepted, documented boundary (`17706427`, comment-only).** Two grounds: (1) the durable fix belongs in the *shared* `TypeRefDecoder` (carry TPA/platform provenance rather than canonicalizing untrusted references by simple name) — out of scope for this harness-only gate, and a harness-side guard isn't feasible since referenced assemblies aren't enumerated/loaded in this path; (2) it cannot corrupt the load-bearing RTS corpus metric — the oracle recompiles every emit, so any unsound `: this(args)` becomes an `OpcodeDiff`/`RecompileFail`, never a false `Exact`.

### Final verdicts on head `17706427`
- **GPT-5.5: ACCEPT-WITH-DOCUMENTED-RESIDUAL** — no false-Exact witness; agrees the referenced-facade residual is acceptable to defer to a shared-decoder follow-up.
- **Gemini 3.1 Pro: CLEAN** — nested finding closed (incl. multi-level nesting), no new soundness hole, and independently reasoned the oracle makes a silent false `Exact` "fundamentally impossible" for the referenced-facade residual.

Both confirmed on `17706427`: `ReturnToSenderPrototypeTests` **115/0** and the NuGet.Versioning `rts-parity` probe **139 exact, 9 opcode diffs, 63 recompile failures** (the acceptance target).

### Where they agreed / diverged
- Agreed on every High finding (rounds 1, 2, and 4 were found by both independently; round 3 by GPT, then accepted by Gemini after the fix).
- Diverged only on rounds 3 and 5, where Gemini was CLEAN and GPT surfaced one further vector each — both of which were then either fixed (round 3) or accepted as a documented, oracle-safe residual (round 5) that Gemini concurred with.

### Follow-up (non-blocking)
The durable fix for the referenced-facade-assembly residual (TPA/platform provenance in the shared `TypeRefDecoder`) is deferred to a separate PR; it does not affect this PR's corpus evidence.

richlander:
## Ready to merge

Two CLEAN reviews on the final head `17706427` (GPT-5.5: ACCEPT-WITH-DOCUMENTED-RESIDUAL; Gemini 3.1 Pro: CLEAN — see the reconciliation comment above).

**Acceptance evidence (harness-only change):**
- NuGet.Versioning `rts-parity` probe: **139 exact, 9 opcode diffs, 63 recompile failures** — the 3 `SemanticVersion` chains moved OpcodeDiff→Exact (136→139), RecompileFail steady at 63, zero regressions.
- `ReturnToSenderPrototypeTests`: **115/0** (1 positive canary + 5 adversarial negatives from rounds 1–4, each verified to fail on the corresponding pre-fix commit).
- Change is confined to `tools/DecompilerHarness/` + its test project; the gate is purely additive (emits `: this(args)` only when the chained-to ctor is the unique non-excluded same-arity binder, else strips), and the RTS oracle recompiles every emit so an unsound emit can only surface as OpcodeDiff/RecompileFail, never a false Exact.

I will not merge — leaving that to @richlander. Any further testing/review I run from here is non-blocking follow-up.

Issue #1, opened 2026-02-05
Packages from custom sources


McMlok:
Hello,

If I understand [PackageCommand](https://github.com/richlander/dotnet-inspect/blob/f58ed14b3b01ad0558f1362de734676a10c57659/src/dotnet-inspect/Commands/PackageCommand.cs) there is no way how to point the tool to download packages from custom source and it uses hardcoded url to nuget.org? 

Do you plan to some how use information from Nuget.config file? In my projects I also have a custom repository with private packages and it would be helpful to be able to use the information from this packages too.

Thank you.

Issue #2731, opened 2026-07-13
RTS: reconstruct generic base interfaces for shell interface types (CS0176 inherited-member gap)


richlander:
## Summary

RTS minimal shells reconstruct interface types with **no base interfaces**. When a
decompiled body calls a member inherited through a generic base interface, the call
binds to a static lookalike and fails to recompile.

## Root cause

Concrete example (`NuGet.Versioning.FloatRange::Equals#0`), product decompilation is
correct:

```csharp
return VersionComparer.Default.Equals(MinVersion, other?.MinVersion);
```

The receiver is typed `IVersionComparer`, whose two base interfaces are
`TypeSpecification`s (`IEqualityComparer<SemanticVersion>`, `IComparer<SemanticVersion>`).
`InterfaceSignatures` drops them (returns `[]` for interface typeDefs; the class path
skips generic/non-TypeDef interfaces), so the shell declares `IVersionComparer` as an
empty interface. `.Equals(a, b)` on that interface-typed receiver then falls back to
static `object.Equals(object, object)` — **CS0176**.

## Why it is not a one-liner

Naively enriching the interface base list breaks implementing classes
(`VersionComparer : IVersionComparer`) with **CS0535** (unimplemented inherited members).
A sound fix likely needs default-interface-method stubs for the *consumed* inherited
member, keyed on the **receiver's static type** — which is not available on the `Call`
node's `Callee` (that points at the corelib generic base interface, rejected as
cross-assembly). Threading receiver-type info requires changes to the product-path
consumed-member evidence.

## Context

Base-class reconstruction (the sibling half of #2527) already landed and removed the
covariance (CS1503) failures. The compile-back floor still rescues the CS0176 shells to
Exact via full-assembly compilation, so this is a shell-fidelity gap, not a product
decompilation bug.

Follow-up to #2527.

Issue #2527, opened 2026-07-08
Row 7 targeted fidelity: member identity context compile-back failures


richlander:
## Row 7 signal

#1599 row 7 targeted-library fidelity now has explicit compile-back buckets. A repeated recompile-fail class is member identity/context mismatch in the compile-back shell: the rendered body is syntactically valid, but binding resolves to the wrong static/instance member, misses a hidden member, or loses the exact overload/constraint context needed by the original IL.

Measured examples:

- `NuGet.Versioning.FloatRange::Satisfies` / `Equals`: `CS0176` for `object.Equals(object?, object?)` accessed as an instance member.
- `NuGet.Shared.HashCodeCombiner::GetHashCode`: `CS1061` for missing `CombinedHash`.
- `NuGet.Versioning.FloatRange::ToString`: `CS1503` around `VersionFormatter` vs `string` overload context.
- Local product small sweep also shows `CS0266`, `CS1503`, `CS0117`, and `CS1729` in the same compile-back evidence lane.

## Evidence

From the row-7 audit on `origin/main`:

- `NuGet.Versioning` exhaustive run: 255 rendered / 249 Full, 177 exact, 28 opcode diffs, 46 recompile fails. Top failure codes include `CS1061` (9), `CS0176` (8), `CS1503` (8), and `CS0117` (5).
- Local small product run (`ILInspector.MetadataPrimitives` + `DotnetInspector.Core`): 235 rendered / 226 Full, 174 exact, 30 opcode diffs, 29 recompile fails. Top failure codes include `CS0266` (7), `CS1503` (6), `CS0117` (3), and `CS1729` (3).

## Done signal

Pick one member-identity/context mismatch bucket, add a focused compile-back canary, and either make it bind/recompile or classify it as a named harness-context limitation so row-7 reporting can distinguish it from actionable product-output invalidity.

richlander:
Closing as part of a clean restart of the RTS program. This documents measured compile-back **shell** recompile failures, framed around the harness reconstructing the missing shape. Under the new direction (RTS as a thin orchestrator; the product seam `TypeShellProducer` owns shell/shape and classifies unsupported surfaces), these symptoms should be re-derived product-ward. Keeping as reference for the concrete failure rows (CS0176/CS0103/display-class names) it measured; fresh product-ward issues will replace it.

PR #2732, opened 2026-07-13
Reconstruct base classes in RTS shells (#2527)

richlander:
- Advances #2527 (follow-up gap tracked as #2731)
- Changes RTS shell base-type reconstruction (harness-only; product decompiler output unchanged)
- Evidence revision: `2bf7f313` (rebased onto `origin/main`)

## Change

**Conclusion:** **PASS** — RTS minimal shells dropped base classes, so decompiled bodies relying on a derived→base conversion could not recompile (CS1503) and attribute hierarchies lost their derives-from-`Attribute` chain; reconstructing the real same-assembly base class turns 22 NuGet.Versioning targets from RecompileFail into Exact/OpcodeDiff with zero regressions.

Before (shell body fails to recompile):

```text
NuGet.Versioning.VersionRange::get_MinVersion  RecompileFail  (base class dropped)
```

After:

```text
NuGet.Versioning.VersionRange::get_MinVersion  Exact
```

## Scope and safety

- **Root cause:** `BaseTypeSignature` emitted only `System.Attribute`, so every other base class was dropped from the shell. Bodies that pass a derived instance where a base type is expected then failed with CS1503, and multi-level attribute shells lost the chain to `System.Attribute`.
- **Fix shape:** Reconstruct the real same-assembly, non-generic base class for plain **class** shells (records/structs/enums/delegates keep their compiler-implied base; generic bases are skipped because the flat shell cannot carry the base's type parameters). To keep it sound, synthesize an accessible parameterless constructor on a reconstructed base that carries only parameterized constructors, so a derived stub's implicit `: base()` never fails (CS7036/CS1729).
- **Product impact:** None. Change is confined to `tools/DecompilerHarness/ReturnToSenderTypePlanner.cs` (RTS shell planning); product decompilation is unchanged.
- **Scope boundary:** The generic-base-**interface** / CS0176 inherited-member gap (empty interface shells, e.g. `FloatRange::Equals#0`) is intentionally left as a frontier and tracked in #2731. The compile-back floor still rescues those shells.
- **RTS boundary:** RTS only orchestrates closure + Roslyn + opcode compare; product/shared code owns C# source generation. This PR touches only the shell type planner.

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass |
| Focused tests | `ReturnToSenderPrototypeTests` (109) passed |
| Targeted compile-back | NuGet.Versioning RTS parity: Exact 136→153, RecompileFail 63→41 |
| ReturnToSender A/B (per-method vs `origin/main`) | 22 wins, 0 regressions |
| Real witness | `VersionRange::get_MinVersion` moved RecompileFail → Exact |

## Compile-back quality

**Conclusion:** **PASS** — 22 measured RecompileFail/OpcodeDiff → Exact/OpcodeDiff transitions on the deterministic NuGet.Versioning probe, zero regressions, verified by a per-method diff against `origin/main` (not just aggregate bucket counts).

### Deterministic NuGet.Versioning probe

Command:

```text
  <nuget.versioning/7.3.0/lib/net8.0/NuGet.Versioning.dll> \
  --emit-corpus-snapshot out.json --compile-cap 0 --corpus-fidelity-cap 500 \
  --corpus-fidelity-oracle return-to-sender --sequential --max-examples 3
```

| Bucket (goal) | Baseline | PR | Delta |
| --- | ---: | ---: | ---: |
| Exact (+) | 136 | 153 | +17 |
| OpcodeDiff | 12 | 17 | +5 |
| RecompileFail (-) | 63 | 41 | -22 |
| ContextFail | 0 | 0 | 0 |

Per-method diff vs `origin/main`: **22 wins, 0 regressions**. The one case that base-class reconstruction alone would regress (`NuGetVersion::.ctor#10`, CS7036 from a missing base parameterless ctor) is held at its pre-existing OpcodeDiff by the synthetic-parameterless-constructor guard.

## Tests

- `CompileBackTargets_ReconstructsBaseClassForCovariantArgument` — derived→base covariant argument now recompiles; shell declares `class Dog : Animal`.
- `CompileBackTargets_SynthesizesParameterlessConstructorForReconstructedBase` — reconstructed base carrying only a parameterized ctor gets a synthetic parameterless ctor so the derived stub's implicit `: base()` binds.
- `CompileBackFirstPropertyGetter_FallsBackToCompileBackFloorForAttributeShellStall` / `CorpusParity_DoesNotApplyCompileBackFloorToRtsFailure` — re-pointed from the (now natively-compiling) dropped-base attribute stall to a CS0534 abstract-member stall, preserving compile-back-floor coverage.

## Adversarial review

Requested from two non-Opus models per policy; results reconciled in a follow-up comment.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -17,6 +17,120 @@ namespace ILInspector.Decompiler.Tests;
 public class ReturnToSenderPrototypeTests
 {
     [Fact]
+    public void CompileBackTargets_SynthesizesParameterlessConstructorForNestedDerivedType()
+    {
+        // Issue #2527 guard (Gemini review of #2732): nested types are emitted from
+        // their enclosing requirement and are absent from the top-level requirement
+        // map, so the synthetic-parameterless-base-constructor scan must also walk
+        // nested types. Here `Outer.Nested : Base` is emitted even though it is never
+        // consumed; `Base` (reconstructed with only a parameterized constructor) must
+        // still receive a synthetic parameterless constructor so Nested's implicit
+        // `: base()` binds natively, without relying on the compile-back floor.
+        var assemblyPath = CompileFixture("""
+            using System;
+
+            public class Base
+            {
+                public Base(int seed)
+                {
+                    Seed = seed;
+                }
+
+                public int Seed { get; }
+            }
+
+            public class Outer
+            {
+                public Outer()
+                {
+                }
+
+                public class Nested : Base
+                {
+                    public Nested() : base(1)
+                    {
+                    }
+                }
+            }
+
+            public static class Use
+            {
+                public static void Run()
+                {
+                    Console.WriteLine(new Base(1));
+                    Console.WriteLine(new Outer());
+                }
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("Use", "Run", 0)]));
+
+            Assert.NotEqual(FidelityCheck.CompileBackStatus.RecompileFail, result.Status);
+            Assert.False(result.UsedCompileBackFloor, result.Detail);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
+    public void CompileBackTargets_DoesNotReconstructGenericBaseClass()
+    {
+        // Issue #2527 guard (Gemini review of #2732): a closed generic base
+        // instantiation (`Derived : Base<int>`) is a TypeSpecification, which the
+        // flat shell cannot carry and cannot own a synthetic constructor for. It must
+        // be dropped rather than emitted, so the derived stub does not fail on an
+        // implicit `: base()` with no parameterless target.
+        var assemblyPath = CompileFixture("""
+            using System;
+
+            public class Base<T>
+            {
+                public Base(int seed)
+                {
+                    Seed = seed;
+                }
+
+                public int Seed { get; }
+            }
+
+            public class Derived : Base<int>
+            {
+                public Derived() : base(1)
+                {
+                }
+            }
+
+            public static class Use
+            {
+                public static void Run()
+                {
+                    Console.WriteLine(new Base<int>(1));
+                    Console.WriteLine(new Derived());
+                }
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("Use", "Run", 0)]));
+
+            Assert.NotEqual(FidelityCheck.CompileBackStatus.RecompileFail, result.Status);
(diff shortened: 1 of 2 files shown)


richlander:
## Adversarial review round 1 — reconciled (both non-Opus, isolated /tmp worktrees)

Reviewers: **GPT-5.5** and **Gemini 3.1 Pro**. All three findings were high-confidence, reproduced against a clean checkout of the PR head, and fixed in `a4e73aa3`.

| # | Reviewer | Finding | Verdict | Resolution |
|---|----------|---------|---------|------------|
| 1 | GPT-5.5 | External (`TypeReference`) base with only a parameterized ctor was reconstructed → the derived stub's implicit `: base()` failed CS7036 (can't synthesize a ctor on a type the shell doesn't own). | **Verified** | Gate `BaseTypeSignature` to same-assembly `TypeDefinition` bases. Test `DoesNotReconstructExternalBaseClass`. |
| 2 | Gemini | Closed generic base instantiation (`Derived : Base<int>`, a `TypeSpecification`) was emitted but skipped by the synthetic-ctor gate → CS7036/CS1729. | **Verified** — subsumed by fix #1 (closed generics are TypeSpecifications, now dropped by the TypeDefinition gate). Test `DoesNotReconstructGenericBaseClass`. |
| 3 | Gemini | Nested derived types are emitted by `NestedTypes()` but absent from `requirementsByMetadataName`, so the synthetic-ctor scan (`IsReconstructedBaseOfAnotherType`) missed them → self-inflicted CS7036 masked by the compile-back floor. | **Verified** | Rewrote the scan to recurse each requirement's nested tree via `TypeOrNestedDerivesFrom` (using `ShellKind`). Test `SynthesizesParameterlessConstructorForNestedDerivedType`. |

All three self-inflicted failures were floor-masked (status stayed `Exact` via `UsedCompileBackFloor`), so the new regression tests assert `Assert.False(result.UsedCompileBackFloor)` to catch them directly.

**Evidence after fixes** (deterministic NuGet.Versioning probe, `a4e73aa3`): Exact 153, RecompileFail 41, OpcodeDiff 17 over 211 checked — per-method vs `origin/main`: **22 wins / 0 regressions**. 112 RTS tests pass.

Re-review of the final head (`a4e73aa3`) requested from both reviewers; will post the CLEAN confirmations before marking ready.

richlander:
## Round 2 re-review — two CLEAN on final head `a4e73aa3`

Both non-Opus reviewers re-reviewed the final head after the three round-1 fixes, each in their own isolated `/tmp` worktree (clean `git status` before/after):

- **GPT-5.5 — CLEAN.** Re-ran all three round-1 repros (external `TypeReference` base, nested external base, closed generic `TypeSpecification` base) — none use the compile-back floor now. Confirmed private-parameterless same-assembly base still gets a synthesized `public Base()` with `UsedCompileBackFloor=false`. 112 tests pass; probe 153 exact / 41 RF; 22 wins / 0 regressions.
- **Gemini 3.1 Pro — CLEAN.** Verified the nested recursion also handles a generic enclosing type (`Outer<T>` with concrete `Nested : Base`) and deep transitive nesting (`Outer.Nested1.Nested2 : Base`). Confirmed the `HandleKind.TypeDefinition` gate aborts synthesis for `TypeSpecification` bases. No new regressions vs main.

**Ready to merge.** Not merging — awaiting @richlander approval.

Issue #2613, opened 2026-07-10
Architecture: keep Metadata data-shaped and move product projections to Research


richlander:
## Direction

Treat `ILInspector.Metadata` as the provider of **typed truth** and
`ILInspector.Research` as the owner of **useful product views**.

Metadata should remain SRM-only, NativeAOT-friendly, and authoritative for:

- extracted type/member/assembly shape;
- normalized identities and anchors;
- structured signatures, attributes, relationships, and provenance;
- Finding observations and transitions;
- typed domain classifications such as API compatibility;
- explicit complete/absent/failed inspection outcomes.

Research should own:

- C# declarations and whole-type skeletons;
- full-body and mixed skeleton/body composition;
- selection among original, decompiled, and annotated implementations;
- grouping, summaries, explanations, and user-facing prose;
- presentation of compatibility changes and other cross-producer conclusions.

This is the architectural counterpart to the Finding producer work: Metadata
publishes stable typed observations, while Research composes and renders them.

## Current pressure

Metadata currently contains presentation-shaped APIs and duplicated data:

- `CSharpDeclarationWriter` and its C# options/rendered units;
- `ApiMember.Signature` alongside structured `ApiSignature`;
- `ApiParameter.Declaration` and declaration-summary properties;
- rendered signatures on scanner records such as `ClassifiedMethodInfo`;
- prose such as `ApiChange.Message` and producer-formatted pair details.

Some of this is compatibility surface, and some is shared today by CLI API
output and Decompiler type/compile-back composition. It should not be moved
blindly: Decompiler depends on Metadata, while Research depends on both, so a
direct move can create dependency cycles or force lightweight consumers to take
the full Research graph.

## Composition model

Making Research cheaper to consume is a **Research composition problem**, not a
reason to keep product projection in Metadata.

A likely shape is:

```text
Metadata
  typed shape + identities + Findings
        |
Research.Core (or equivalent lightweight composition surface)
  type-view planning + skeleton/declaration projection
        |----------------------|
Research.Decompiler       Research.Analysis
  full C# bodies            facts/annotations
        |----------------------|
Research aggregate facade
  skeleton / full / mixed views
```

Names and exact assembly boundaries are design work. The essential constraints
are:

- lightweight shape/skeleton consumers must not require every body/fact provider;
- the convenient aggregate may reference all capabilities;
- optional runtime behavior alone does not make a compile-time dependency cheap;
- a small shared C# syntax kernel below Research is acceptable when needed to
[...]

PR #2737, opened 2026-07-13
Extract CSharpTypeProducer type-shape seam into the product (RTS reversal, #2613)

richlander:
## Direction reversal

#2732 added compile-back type-shell knowledge (base-class reconstruction) **into**
the ReturnToSender (RTS) harness. This reverses that direction: RTS should be an
**orchestrator** between the product and Roslyn, not a second C# shell generator.
Every bit of C# the harness owns is (a) load-bearing test risk and (b) work every
other product consumer must re-implement. Per **#2613**, this knowledge belongs in
the product: Metadata = typed truth, `ILInspector.CSharp` = C# projection.

## Increment 1 — neutral by relocation

This is the first of several increments that progressively thin the harness toward
pure orchestration. It establishes the product seam with **zero behavior change**.

- **New product component** `ILInspector.CSharp/CompileBackTypeSkeleton.cs`:
  owns the base-type reconstruction gate (#2732's logic) and the unsupported-surface
  heuristic. SRM-only, NativeAOT-friendly, Roslyn-free, no inspected-assembly loading.
- **Harness forwards to it**: `BaseTypeSignature` and `IsUnsupportedSurfaceSignature`
  in `ReturnToSenderTypePlanner.cs` become thin forwarders. The harness keeps its own
  C# display normalization (`CompileBackTypeSignature.Display` -> `Clean`) at the
  boundary, so the product heuristic always judges an already-normalized display name
  and behavior is byte-identical for all 8 member-surface callers and the base path.

### Why neutral

The extraction relocates logic verbatim; the `Clean` normalization stays harness-side
at the call boundary (it depends on `ILInspector.Decompiler`-internal `CSharpNaming`,
so it is not part of this increment's move). The base-path unsupported check is reached
only for a same-assembly `TypeDefinition` name, which cannot carry `delegate*`/`<>`/`{`
shapes, so the dropped inner `Clean` there is provably non-divergent.

## Evidence

Deterministic NuGet.Versioning RTS-parity probe (cap 500), origin/main vs this head:

| Metric | origin/main | this PR |
| --- | --- | --- |
| Exact | 153 | 153 |
| OpcodeDiff | 17 | 17 |
| RecompileFail | 41 | 41 |
| ContextFail | 0 | 0 |

Per-method ranked diff vs the pre-#2732 baseline: **22 wins / 0 regressions** — identical
to merged main (the #2732 improvement is fully preserved).

Tests: **112 RTS tests** (`ReturnToSenderPrototypeTests`) pass; **8 new product tests**
(`CompileBackTypeSkeletonTests`) cover the surface heuristic.

## Follow-up increments (not in this PR)

- Interface projection (fills the `GetTypeSurface` interface gap; #2731 CS0176).
- Type-shell header: kind/modifiers/attributes/type-parameters.
- Member-closure composition — the bulk of the shell generator.

## Review

Increment establishes an architectural seam with behavior-preserving relocation; still
requesting two non-Opus adversarial reviews per policy given the RTS/soundness surface.

### DIFF
--- a/src/ILInspector.CSharp.Tests/CSharpTypeProducerTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpTypeProducerTests.cs
@@ -1,6 +1,6 @@
 namespace ILInspector.CSharp.Tests;
 
-public sealed class CompileBackTypeSkeletonTests
+public sealed class CSharpTypeProducerTests
 {
     [Theory]
     [InlineData("delegate*<int, void>")]
@@ -9,7 +9,7 @@ public sealed class CompileBackTypeSkeletonTests
     [InlineData("(int, string){")]
     public void IsUnsupportedSurfaceSignature_FlagsUnrepresentableShapes(string signature)
     {
-        Assert.True(CompileBackTypeSkeleton.IsUnsupportedSurfaceSignature(signature));
+        Assert.True(CSharpTypeProducer.IsUnsupportedSurfaceSignature(signature));
     }
 
     [Theory]
@@ -19,6 +19,6 @@ public sealed class CompileBackTypeSkeletonTests
     [InlineData("int[]")]
     public void IsUnsupportedSurfaceSignature_AllowsOrdinarySignatures(string signature)
     {
-        Assert.False(CompileBackTypeSkeleton.IsUnsupportedSurfaceSignature(signature));
+        Assert.False(CSharpTypeProducer.IsUnsupportedSurfaceSignature(signature));
     }
 }
--- a/src/ILInspector.CSharp.Tests/CompileBackTypeSkeletonTests.cs
+++ b/src/ILInspector.CSharp.Tests/CompileBackTypeSkeletonTests.cs
@@ -1,0 +1,24 @@
+namespace ILInspector.CSharp.Tests;
+
+public sealed class CompileBackTypeSkeletonTests
+{
+    [Theory]
+    [InlineData("delegate*<int, void>")]
+    [InlineData("@delegate*<int, void>")]
+    [InlineData("<>c__DisplayClass0_0")]
+    [InlineData("(int, string){")]
+    public void IsUnsupportedSurfaceSignature_FlagsUnrepresentableShapes(string signature)
+    {
+        Assert.True(CompileBackTypeSkeleton.IsUnsupportedSurfaceSignature(signature));
+    }
+
+    [Theory]
+    [InlineData("System.Int32")]
+    [InlineData("Samples.Widget")]
+    [InlineData("System.Collections.Generic.List<int>")]
+    [InlineData("int[]")]
+    public void IsUnsupportedSurfaceSignature_AllowsOrdinarySignatures(string signature)
+    {
+        Assert.False(CompileBackTypeSkeleton.IsUnsupportedSurfaceSignature(signature));
+    }
+}
--- a/src/ILInspector.CSharp/CSharpTypeProducer.cs
+++ b/src/ILInspector.CSharp/CSharpTypeProducer.cs
@@ -5,25 +5,26 @@ using ILInspector.Metadata;
 namespace ILInspector.CSharp;
 
 /// <summary>
-/// Product-owned composition of compile-back type skeletons: the minimal,
-/// compilable type-shell facts a consumer needs to reconstruct a member's
-/// referenced types without loading the inspected assembly or using Roslyn.
+/// Produces C#-flavored type <em>shapes</em> — the skeletal facts a consumer needs
+/// to render a type declaration and its members without decompiling method bodies,
+/// loading the inspected assembly, or using Roslyn. It is the C#-spelling companion
+/// to the metadata-spelling producer (<c>System.Int32</c> vs <c>int</c>), and stays
+/// SRM-only and NativeAOT-friendly so consumers that only want skeletons never take
+/// the decompiler dependency. The decompiler layer coordinates with this producer to
+/// fill selected members with full bodies.
 ///
-/// This is the first seam extracted from the ReturnToSender harness so the
-/// knowledge of how to shape a compile-back shell lives in the product (and is
-/// validated by the harness compile-back oracle) rather than being re-derived by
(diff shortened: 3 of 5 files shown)


richlander:
## Adversarial review — round 1

Two non-Opus reviewers (per policy) on head `ab4b6c3f`, isolated worktrees:

- **GPT-5.5** — **CLEAN**. Verified gate order vs origin/main, all 8 member-surface call sites, builds, 8 + 112 tests, and reproduced the neutrality probe (153/17/41/0).
- **Gemini 3.1 Pro** — **1 HIGH finding (valid, actioned)** + 1 Medium (doc clarity).

### Diverging finding (Gemini, verified and fixed)

`ReconstructedBaseTypeName` ran the pure `IsUnsupportedSurfaceSignature` heuristic on the **raw** `TypeResolver.GetTypeName` base name. origin/main ran the equivalent check on the **Clean'd** display form (`CompileBackCSharpNames.Clean`), which strips compiler-generated `<>` segments — making that check a **no-op** for a same-assembly `TypeDefinition` base. The raw check would instead wrongly drop a generated `<>`-named base that origin/main reconstructed. GPT-5.5 confirmed structural gate-order parity but did not catch this raw-vs-normalized semantic gap.

**Fix (`e94d7d4b`):** removed the base-path gate entirely — since origin/main's form was provably a no-op for TypeDefinition base names, removal is byte-identical to origin/main and eliminates the divergence. The downstream C# printer sanitizes the emitted base name.

Medium (test asserts the raw-`<>` contract that the member path never hits): the method is a pure heuristic and the test exercises its contract directly; the XML doc now states callers pass an already-normalized display name. No behavior change.

Neutrality after fix: probe **153 exact / 17 opcode / 41 recompile / 0 context**, **22 wins / 0 regressions**; 112 RTS + 8 product tests pass.

Both reviewers are re-reviewing head `e94d7d4b`. Not ready to merge until both are CLEAN on the final head.

richlander:
## Adversarial review — final (head `73d7edc5`): two CLEAN

Both non-Opus reviewers re-reviewed the final head and are **CLEAN**.

### Round 2 — the base-path surface check (both reviewers found real, opposite-direction bugs)

The extracted base reconstruction originally ran the `IsUnsupportedSurfaceSignature` heuristic on the **raw** metadata name, then (in the interim fix) dropped it entirely. Both were non-neutral in opposite directions:

- **Gemini 3.1 Pro (HIGH):** raw-name check wrongly **dropped a generated `<>c`-named base** that origin/main kept (origin/main ran the check on the `Clean`'d display form, which strips generated `<>`).
- **GPT-5.5 (Medium):** removing the gate wrongly **kept a `{`/`delegate*`-named base** that origin/main dropped (`Clean` does *not* strip `{`/`delegate*`).

**Reconciling fix (`73d7edc5`):** split responsibilities — the product `ReconstructedBaseTypeName` owns only the **metadata-level gates** and returns the candidate base; the harness `BaseTypeSignature` applies the **C# surface check on the `Clean`'d display form** before emitting. This reproduces origin/main's "Clean then check" semantics byte-for-byte: generated `<>` base kept, `{`/`delegate*` base dropped. The C#-display policy stays where the `Clean` normalization lives.

### Verification on `73d7edc5`

| Check | Gemini 3.1 Pro | GPT-5.5 |
| --- | --- | --- |
| Verdict | CLEAN | CLEAN |
| `git status` clean | ✓ | ✓ |
| Build (product + harness + tests) | ✓ | ✓ |
| RTS tests | 112 / 112 | 112 / 112 |
| Product tests | 8 / 8 | 8 / 8 |
| Neutrality probe | 153 / 17 / 41 / 0 | 153 / 17 / 41 / 0 |
| Synthetic repro | `<>c` base kept | `N.Base{X` base → dropped (null) |

Where they diverged in round 1 (GPT CLEAN, Gemini HIGH), the HIGH was valid and actioned; round 2 surfaced the complementary `{` case from GPT. Both are resolved by the single boundary-split fix and independently re-confirmed.

Neutral by relocation confirmed: **22 wins / 0 regressions** vs the pre-#2732 baseline. Reviewer worktrees were isolated under `/tmp`; findings reproduced against clean checkouts.

**Ready to merge** (pending @richlander approval — not self-merging).

richlander:
## Rename: `CompileBackTypeSkeleton` → `CSharpTypeProducer` (commit `f000a151`)

Per the layering discussion on #2613, "compile-back" is a ReturnToSender **harness** concept, not a product one. The extracted product seam is the **C#-spelling type-shape producer** — you can already ask the CLI for a type's source, and there should be one product type-producer with a variety of capabilities. Renamed the class/file/test accordingly and reworded the docs to describe it as the C#-flavored producer of skeletal type shapes (base reconstruction + surface heuristic are its first two capabilities), the C#-spelling companion to a future metadata-spelling producer (`int` vs `System.Int32`).

**Agreed target layering** (folds land in follow-up PRs):
- `ILInspector.CSharp` owns **all skeletal shapes** (types, members, stubbed-member shells) — SRM-only, NativeAOT-friendly, no Decompiler dependency, so consumers who only want skeletons never pull in the decompiler.
- `ILInspector.Decompiler` coordinates with the CSharp producer to print a **mix** of skeletal + full-bodied members (per-member `MemberAnchor` body/skeleton choice). It is **not** the source of shells.
- `Research` only when mixing IL + SourceLink.
- Partial member-set printing at **both** layers (consumer filters `ApiType.Members`).

**Scope:** mechanical rename only — identifiers, file names, XML docs. The harness still forwards base reconstruction and the surface gate to the product exactly as before.

**Neutrality (byte-for-byte A/B):** built the pre-rename parent (`73d7edc5`) and post-rename head (`f000a151`) in isolated worktrees and ran the identical deterministic NuGet.Versioning fidelity census (`--return-to-sender --fidelity-check`). The full logs are **identical (0 diff lines)**: `176 exact / 29 opcode-diff / 0 context-fail / 46 recompile-fail over 249 Full`.

**Tests:** 8 `CSharpTypeProducerTests` + 197 RTS/decompiler tests pass.

No re-review needed for behavior (byte-identical A/B), but flagging for @richlander since the head advanced past the two prior CLEAN reviews.

PR #2739, opened 2026-07-14
Spell constructor-chain args at their parameter type; remove RTS overload gate (#2726)

richlander:
- Advances #2726
- Changes: RTS constructor-chain planner no longer models C# overload resolution to decide whether to emit `: this(args)` — **harness-only**, product output unchanged.
- Evidence revision: `3ece0365`

## Change

**Conclusion:** **PASS** — the RTS ctor-chain gate duplicated C# binding rules the harness should not own; removing it lets the product's printed chain + the Roslyn oracle judge binding, moving the 3 NuGet.Versioning `SemanticVersion` chains (assignable-not-identity arg, `string[]` → covariant `IEnumerable<string>`) OpcodeDiff → Exact with **corpus recompile-failures unchanged (581 → 581)**.

Before (gate ON, NuGet.Versioning rts-parity):

```text
153 exact, 17 opcode diffs, 41 recompile failures
```

After (gate removed):

```text
156 exact, 14 opcode diffs, 41 recompile failures
```

## Scope and safety

- **Root cause:** the product `CSharpPrinter` already renders the faithful `: this(args)` chain (#2708 threaded it through RTS), but the planner then re-derived C# overload resolution (`ChainArgumentsAreFaithful` / `ConstructorCouldBindToArgCount`) and pre-stripped the initializer whenever it could not itself prove the printed call binds to the chained-to constructor. An assignable-but-not-identity argument sharing arity with a sibling failed both branches, so a correct chain was stripped (OpcodeDiff).
- **Fix shape:** RTS orchestration change — delete gate condition 4 and its three dead helpers; keep only structural guards (chained-to ctor reconstructable + a same-arity non-target sibling present as a binding target). Emit the product's chain and let Roslyn judge.
- **Product impact:** none. Only `tools/DecompilerHarness/ReturnToSenderTypePlanner.cs` and its tests change.
- **Soundness:** a mis-binding can never produce a false Exact — a wrong bind changes the emitted call token (OpcodeDiff) or fails to compile (RecompileFail), both surfaced honestly by the oracle. The only theoretical downside is a cosmetic OpcodeDiff → RecompileFail on a chain that is ambiguous *in the reconstructed shell*; the corpus check below shows this does not occur on any real assembly.
- **RTS boundary:** RTS now only orchestrates closure + Roslyn + opcode compare; product/shared code owns C# source generation. This removes C# knowledge from the harness so recompile validates the product has it, and other product consumers need not reimplement it.

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass (0 warnings) |
| Harness build | Pass (0 warnings) |
| Focused tests | `ReturnToSenderPrototypeTests` 113/0 |
| Decompiler fast suite | Pass (2409/0, `-notrait Speed=Slow`) |
| Targeted compile-back | NuGet.Versioning rts-parity 153/17/41 → 156/14/41 (+3 exact) |
| Broad compile-back | 14-assembly corpus, cap 200: RecompileFail unchanged 581 → 581 |
| Real witness | `SemanticVersion::.ctor` chains moved OpcodeDiff → Exact |

## Compile-back quality

**Conclusion:** **PASS** — +3 Exact, −3 OpcodeDiff, RecompileFail identical; no product defect masked (the oracle recompiles every emit).

### Broad assembly context (full-corpus RecompileFail check)

Run: 14-assembly corpus (`eng/prepare-decompiler-corpus.sh`), `--corpus-fidelity-oracle return-to-sender --corpus-fidelity-cap 200`, 2132 checked. Matched gate-ON baseline (`e1ac70b2`) vs gate-OFF (`3ece0365`):

| Metric | Baseline (gate ON) | PR (gate removed) | Delta |
| --- | ---: | ---: | ---: |
| Exact | 1354 | 1357 | **+3** |
| OpcodeDiff | 197 | 194 | **−3** |
| RecompileFail | 581 | 581 | **0** |

RecompileFail buckets are byte-for-byte identical on both sides:

```text
compiler diagnostic: 533
missing symbol:      47
conversion:          1
```

**Broad verdict:** **PASS** — the change is a pure OpcodeDiff → Exact gain on the 3 target chains; the predicted cosmetic OpcodeDiff → RecompileFail shift does not materialize anywhere in the corpus.

### ReturnToSender parity

RTS parity vs the current compile-back lane over the same 2132 targets: gate-ON `8 rescued / 1523 same / 601 worse`; gate-OFF `8 rescued / 1526 same / 598 worse` (3 fewer `worse`, matching +3 exact).

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| Shell-ambiguity OpcodeDiff → RecompileFail | Left as honest signal | Never a false Exact; a RecompileFail on a shell-ambiguous chain is the correct orchestrator report, not a bug. Not observed in the corpus. |
| Faithful sibling-signature reconstruction (Option B) | Not done | Only needed if a real RF regression appears; corpus shows none. |

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| GPT-5.5 | Pending | Adversarial review requested |
| Gemini 3.1 Pro | Pending | Adversarial review requested |

**Review conclusion:** **REVIEW** — two non-Claude adversarial reviews requested per policy; reconciliation to follow.

## Validation

```bash
dotnet run --project tools/DecompilerHarness -c Release --no-build -- <corpus...> --emit-corpus-snapshot /tmp/rts.json --compile-cap 0 --corpus-fidelity-cap 200 --corpus-fidelity-oracle return-to-sender
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ConstructorChainRenderingTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ConstructorChainRenderingTests.cs
@@ -14,6 +14,33 @@ public class ConstructorChainRenderingTests
     static readonly TypeRef Unrelated = TypeRef.CoreLib("System", "Unrelated");
     static readonly TypeRef Int32 = TypeRef.CoreLib("System", "Int32");
     static readonly TypeRef Void = TypeRef.CoreLib("System", "Void");
+    static readonly TypeRef Object = TypeRef.CoreLib("System", "Object");
+    static readonly TypeRef String = TypeRef.CoreLib("System", "String");
+    static readonly TypeRef StringArray = TypeRef.SzArray(TypeRef.CoreLib("System", "String"));
+    static readonly TypeRef StringEnumerable = TypeRef.GenericInstance(
+        TypeRef.CoreLib("System.Collections.Generic", "IEnumerable`1"), [TypeRef.CoreLib("System", "String")]);
+
+    // Renders a this(...) chain to a single-parameter constructor whose parameter
+    // is `paramType`, passing `argument`. The printer must spell the argument at
+    // its exact parameter type so the recompiled shell rebinds to the same
+    // overload the IL selected — a null/reference-upcast argument that reads back
+    // without its parameter type silently rebinds to a narrower sibling overload.
+    static DecompilerResult RenderChain(TypeRef paramType, IrExpression argument)
+    {
+        var ctor = new MethodRef(Derived, ".ctor", Void, [paramType], HasThis: true);
+        var call = new Call(ctor, isVirtual: false, [new LoadArgument(0, "this", Derived), argument]);
+        var entry = new Block(0);
+        entry.Add(new ExpressionStatement(call));
+        entry.Add(new Return(null));
+        var container = new BlockContainer();
+        container.Add(entry);
+        var signature = new MethodSignature(Void, [new Parameter("value", paramType)], HasThis: true, GenericParameterCount: 0);
+        var function = new IrFunction(".ctor", Derived, signature, [Derived], container)
+        {
+            BaseType = Base,
+        };
+        return CSharpPrinter.Print(function);
+    }
 
     static DecompilerResult RenderConstructor(TypeRef chainDeclaringType, IrExpression receiver, bool diagnose = false)
     {
@@ -70,4 +97,46 @@ public class ConstructorChainRenderingTests
         Assert.Contains("direct constructor call", result.Output);
     }
 
+    // A null literal is type-less, so C# overload resolution re-picks the most
+    // specific applicable sibling for the recompiled `this(null)` — a narrower
+    // `C(string)` beats the original `C(object)`. Spelling `(object)null` pins the
+    // parameter type so the shell rebinds to the constructor the IL selected.
+    [Fact]
+    public void NullArgument_CastsToReferenceParameterType()
+    {
+        var result = RenderChain(Object, new Constant(null, Object));
+
+        Assert.Equal("this((object)null)", result.ConstructorChain);
+    }
+
+    // string[] is assignable to IEnumerable<string> by array covariance (a no-op
+    // reference conversion, no IL). Reading the argument back bare would let it
+    // bind to a same-arity sibling that accepts string[] exactly; the (IEnumerable
+    // <string>) cast keeps the recompiled shell on the covariant overload (#2726).
+    [Fact]
+    public void AssignableArgument_CastsToWiderReferenceParameterType()
+    {
+        var result = RenderChain(StringEnumerable, new LoadLocal(0, StringArray));
+
+        Assert.StartsWith("this((", result.ConstructorChain);
+        Assert.Contains("IEnumerable<string>)", result.ConstructorChain);
+    }
+
+    // The argument already has the parameter type: no fidelity cast, no noise.
+    [Fact]
+    public void IdentityReferenceArgument_IsNotCast()
+    {
+        var result = RenderChain(Object, new LoadLocal(0, Object));
+
+        Assert.Equal("this(V_0)", result.ConstructorChain);
+    }
+
+    // Value-typed parameters keep the numeric-coercion path; no reference cast.
+    [Fact]
+    public void ValueTypeArgument_IsNotCast()
+    {
+        var result = RenderChain(Int32, new Constant(5, Int32));
+
+        Assert.Equal("this(5)", result.ConstructorChain);
+    }
 }
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -691,15 +691,16 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
-    public void CompileBackTargets_DropsInitializerWhenChainedConstructorIsOverloadAmbiguous()
+    public void CompileBackTargets_EmitsChainAndOracleRejectsRebindableBoxedArgument()
     {
-        // Issue #2678 guard: the printer can render chain arguments without a
-        // disambiguating cast (a `box` to `object` prints as its inner value), and
-        // C# overload resolution — including the target constructor — can then
-        // re-bind the call. A boxed argument is never faithful, and here three
-        // same-arity constructors (`C(object)`, `C(int)`, the target `C(string)`)
-        // share the callee's arity, so unique-arity binding does not hold either.
-        // The initializer must be stripped rather than emit a wrong-binding chain.
+        // Issue #2726: RTS no longer models C# overload resolution to decide
+        // whether to emit `: this(args)` — that is Roslyn's job. The printer can
+        // render a chain argument without its disambiguating cast (a `box` to
+        // `object` prints as its inner value `1`), so the reconstructed shell may
+        // bind the call differently than the original did. RTS emits the product's
+        // chain anyway and lets the oracle judge: `: this(1)` binds to `C(int)`
+        // instead of the original `C(object)`, so the recompiled IL differs. That
+        // is an honest non-Exact signal, never a false Exact.
         var assemblyPath = CompileFixture("""
             public class C
             {
@@ -723,8 +724,8 @@ public class ReturnToSenderPrototypeTests
                 assemblyPath,
                 [new ReturnToSender.RequestedTarget("C", ".ctor", 2)]));
 
-            Assert.NotEqual(FidelityCheck.CompileBackStatus.RecompileFail, result.Status);
-            Assert.DoesNotContain(": this(", result.Source);
+            Assert.Contains(": this(", result.Source);
+            Assert.NotEqual(FidelityCheck.CompileBackStatus.Exact, result.Status);
         }
         finally
         {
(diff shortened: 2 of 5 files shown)


richlander:
## Adversarial review reconciliation — BLOCKED (central claim falsified)

Two isolated adversarial reviews (both from the non-Claude roster), each in its own worktree at clean PR head `3ece0365`:

- **GPT-5.5** — **CRITICAL**: the RTS oracle classifies `Exact` from opcode **names only** (`originalOps.SequenceEqual(recompiledOps)` at `ReturnToSender.cs`, each op = `CanonicalOpcode(instruction.OpCodeName)`), ignoring call operands.
- **Gemini 3.1 Pro** — **CONFIRMED** the same finding, with a live repro.

**Agreement:** both reviewers independently reproduced a **false `Exact`**. Fixture (target `C::.ctor()`):

```csharp
public class C
{
    public C(object x) { }
    public C(string s) { }
    public C() : this((object)null) { _ = new C("body"); }
}
```

The product printer drops the disambiguating cast, rendering `: this(null)`. In the recompiled shell `null` binds to `C(string)` (better conversion) instead of the original `C(object)`. The `call` target token changes (`C::.ctor(object)` → `C::.ctor(string)`), but the opcode-name sequence (`ldarg.0 ldnull call`) is identical, so the oracle reports **`STATUS: Exact`**. Verified independently by the PR author as well.

**Impact on this PR:** the gate this PR removed was (partly) a soundness compensator for that half-blind oracle. This PR's load-bearing claim — "a wrong bind changes the emitted call token → OpcodeDiff/RecompileFail, never a false Exact" — is **false** for the opcode-name-collision subclass (typically `null`/`ldnull` args and reference upcasts, where only the call-target operand differs).

**Two root causes exposed** (both independent of the #2726 target being a genuine true Exact):
1. RTS oracle is opcode-name-only — blind to call targets.
2. The product C# printer drops load-bearing disambiguating casts (`this((object)null)` → `this(null)`) — a real decompiler fidelity bug.

**Discriminator note:** the operand-aware `IlDiff` (`IlBodyDiffResult`, computed but not used for status) returns `IsExact=false` for **both** the false-Exact fixture **and** the legitimate #2726 canary (benign cross-assembly token noise), so naive `IlDiff.IsExact` gating is **not** a viable fix — it would kill the real #2726 win.

**Status: this PR is BLOCKED, not ready to merge.** Holding for maintainer steer on the fix direction (operand-aware oracle vs product printer fix vs revert-and-keep-gate).

richlander:
## Update — root cause fixed in the PRODUCT (chosen direction B); PR is now sound

Following the reconciliation above (both reviewers confirmed the false-Exact), the fix moved to the **product C# printer** rather than tightening the harness oracle. The RTS gate removal stays (RTS should orchestrate, not model C# overload resolution); the printer now takes responsibility for spelling chain arguments faithfully, so the "let the oracle judge" posture becomes sound.

### The fix
`CSharpPrinter` now spells each constructor-chain initializer argument **at its exact parameter type**. IL verification guarantees a chain argument's stack type is already assignable to the chained-to parameter, so `EffectiveType(arg) != param` implies a **faithful widening reference conversion that emits no opcode** — the recompiled body stays byte-identical while the shell rebinds to the overload the IL actually selected.

- `: this((object)null)` (was `this(null)`) → binds the original `C(object)`, not a more-derived `C(string)` sibling.
- `: this((object)text)` (was `this(text)`) → binds `C(object)`, not the target `C(string)` calling itself.
- `: this((IEnumerable<string>)labels)` where `labels` is `string[]` → array covariance, IL-neutral → **#2726 wins soundly**.

Only concrete reference-like parameters get the cast; value/numeric/enum/pointer/type-parameter parameters keep the existing coercion path. Scoped to the base/this chain path via a `chainFidelityCasts` flag on `Arguments()`. The knowledge lives in the product (where recompile validates it); RTS stays C#-free.

### Why this closes the blind spot (not just the one fixture)
The oracle's opcode-name comparison is blind only to conversions that **emit no opcode** — reference upcasts and type-less `null`. Those are exactly what the printer now spells. A **value-type box** (`(object)1`) rebind is *not* blind: dropping the box removes a `box` opcode, so it already surfaces as an honest OpcodeDiff. A **lambda** target-typing ambiguity surfaces as an honest RecompileFail. Both remain honest non-Exacts (canaries added), never false Exacts.

### Evidence
- Product printer unit tests: `ConstructorChainRenderingTests` (null-literal → `(object)null`, `string[]` → `(IEnumerable<string>)`, identity/value-type negatives).
- `ReturnToSenderPrototypeTests`: the four former gate fixtures reworked to the sound contract — reference-upcast and null-literal rebinds now round-trip **Exact**; value-box is a visible OpcodeDiff; lambda is an honest non-Exact — plus the reviewers' exact `(object)null` fixture as a regression canary.
- Fast decompiler subset: **2413 / 0**.
- NuGet.Versioning rts-parity: **156 exact / 14 opcodediff / 41 RF** (was 153/17/41 gate-ON) = **+3 Exact / −3 OpcodeDiff, RF unchanged** — the 3 SemanticVersion chains, now soundly Exact.
- Full-corpus RTS-parity A/B (product fix vs pre-fix) in progress; will post the quality-diff card.

Fresh adversarial reviews on this head to follow.

richlander:
## Corpus A/B + adversarial review reconciliation (head `216af4f9`)

### Corpus fidelity A/B (RTS parity, cap 200, 14-assembly NuGet corpus, 2132 checked)

All figures are tool-emitted (`--emit-corpus-snapshot`); per-method delta computed from the snapshots.

| Snapshot | Exact | OpcodeDiff | RecompileFail |
| --- | --- | --- | --- |
| `origin/main` (gate-ON) | 1354 | 197 | 581 |
| **HEAD** (gate removed + product chain-fidelity cast) | **1357** | **194** | **581** |

**Delta vs `origin/main`: +3 Exact / −3 OpcodeDiff / RecompileFail flat (0 new RF).**

Per-method delta (`origin/main` → HEAD) over all 89,065 methods: **3 improved, 0 regressed.** The three improvements are exactly the issue-#2726 targets:

- `NuGet.Versioning.SemanticVersion::.ctor` overload 3 — OpcodeDiff → Exact
- `NuGet.Versioning.SemanticVersion::.ctor` overload 5 — OpcodeDiff → Exact
- `NuGet.Versioning.SemanticVersion::.ctor` overload 6 — OpcodeDiff → Exact

Against the gate-off-pre-fix snapshot the tallies are identical (0 changed): the gate removal is what *surfaces* the chains, and the product cast makes them round-trip **faithfully** (`string[]` → covariant `IEnumerable<string>` is an intrinsic implicit conversion → no opcode → byte-identical), so the Exacts are true Exacts, not the false Exacts the earlier harness-only approach would have produced.

### Adversarial reviews (two models, isolated worktrees, head `216af4f9`)

**Gemini 3.1 Pro — CLEAN.** Five real-run probes: `is`/`&&` precedence in `IsConcreteReferenceParameter` (checks `IsReferenceLike` for all four kinds); erased-type false-Exact attempts (`dynamic`/tuple-names/nullable-ref are indistinguishable for overload resolution → correct bind); `Nullable<T>`/value-type CS0037 (`TypeParameter` + `Nullable<T>` correctly excluded); unconstrained `(T)null` (produces OpcodeDiff but **pre-existing** — `initobj` local makes `ChainCallIndex` bail to `UnsupportedNode`; the fix path is never reached); dynamic ctor dispatch (C# forbids it, CS1975; explicit `(object)` cast preserved faithfully). Conclusion: the cast strictly enforces the IL's method resolution in C#.

**GPT-5.5 — one finding (graded High), reconciled as non-blocking.** Fixture `C(A a) : this((IDisposable)a)` where `A : IDisposable` classifies **OpcodeDiff** at head (recompiled adds `castclass`). Verified against a clean head.

- **Not a false Exact.** Status is OpcodeDiff — the oracle correctly reports the difference; the soundness bar (no new false Exact) is not crossed.
- **Root cause is the RTS shell reconstruction, not the product printer.** The reviewer's own generated shell stubs `class A {}` *without* `: IDisposable`, so Roslyn compiles `(IDisposable)a` as an explicit cast. In a real full-graph recompile the same cast is an implicit no-op (byte-identical). This is a pre-existing minimal-shell limitation (external/stubbed interface implementations dropped), independent of this PR.
- **Conservative direction.** RTS *under-reports* Exact here (reports OpcodeDiff for something that is really Exact in full context). Under-reporting fidelity is sound.
- **Not a #2726 regression.** The target chains use `string[]` → `IEnumerable<T>` array covariance, which the compiler knows intrinsically (not a reconstructed stub); the corpus A/B shows 0 regressions.
- Optional follow-up (out of scope for #2726): teach the RTS shell to reconstruct an argument type's interface implementations so these narrow cases classify Exact rather than OpcodeDiff.

**Agreement/divergence:** no substantive divergence — Gemini's probes confirm the printer path never mis-binds; GPT's case is a harness shell-reconstruction artifact yielding an honest OpcodeDiff. The product fix is sound: it introduces no false Exact, no CS-error, and no corpus regression, while moving the three #2726 chains to Exact.

richlander:
### Review status update

**GPT-5.5 re-reviewed CLEAN** after reconciliation: agrees the OpcodeDiff fixture is an honest, conservative RTS shell-reconstruction artifact (arg-type interface impl dropped), not a new false Exact and not a corpus regression — no counterexample for either. Combined with Gemini 3.1 Pro's CLEAN, **both adversarial reviews are now CLEAN**.

Optional harness follow-up (out of scope here) filed as #2745.

Issue #2585, opened 2026-07-09
Finding matching: the hard→soft identity ladder (limiting false positives)


richlander:
# Finding matching: the hard→soft identity ladder (limiting false positives)

Design spec for extending the shared Finding matcher beyond its exact hard tier.
The ordered tier originated in #2584 as a single exact identity key with
order-preserving LCS matching, exhaustively proved against `IlBodyDiff`.
Identity-set matching has since landed. This issue specifies the remaining soft
ladder and — the load-bearing part — **how to add it without importing the
false-positive problem that sinks naive diff tools.**

Refs: spine #2564; ordered hard tier #2584; Finding roadmap #2614; coordinate model PR #2624.

## Coordinate prerequisite

PR #2624 defines the coordinate boundary this ladder builds on:

- collection order, not `Finding<T>.Ordinal`, is matching authority;
- `Ordinal` is optional retained metadata for ordered producers;
- identity-set producers do not fabricate ordinals;
- producer coordinates and cross-stream provenance remain typed payloads rather
  than one opaque generic anchor.

Soft matching therefore extends `FindingKey` correspondence with structured
identity projections, typed deltas, and match-tier provenance. Deterministic
tie-breaking uses input order or a producer's stable sort; it does not turn an
ordinal or representation-local coordinate into identity.

## Problem

The matcher (`FindingMatcher`) establishes which old occurrence corresponds to which new one.
Rung 0 matches only on **exact identity** (`FindingKey.IdentityKey` equality). That catches
adds/removes/exact-moves but misses every *graded* correspondence the real system needs:

- **IL**: a constant changed `ldc.i4 3 → 4` (today: remove+add; want: one `Changed` with the value delta); a body that changed by one instruction (want: near-miss match, not two unrelated hunks).
- **C#**: a statement whose trivia/formatting changed; `x += 1` vs `x = x + 1`; a renamed local.
- **Metadata/API**: a renamed member (same signature); a signature that evolved (added a defaulted param); an **extension method promoted to an instance method** (`Ext.DoThing(this MyType, int)` → `MyType.DoThing(int)`); a member **relocated** to a different type.

All of these are "the same thing, changed" — but only if the matcher can match on something *looser*
than exact identity. Soft matching is exactly where diff tools earn their false positives (spurious
renames/moves), so the ladder must be introduced with hard soundness rules.

## Model: one structured identity, a ladder of keys

Do **not** hand the matcher N independent canonicalizers. Each producer computes **one structured
[...]

PR #2741, opened 2026-07-14
Bound structured soft candidate matching

richlander:
Advances #2585 by closing the structured-soft-pass scaling gap before another tier lands.

## Change

- Replaces Cartesian soft-candidate materialization with exact-projection buckets and bounded endpoint-degree summaries.
- Emits only mutual degree-one pairs while preserving ambiguity suppression, exact/move dominance, strongest-tier provenance, and deterministic ordering.
- Carries strongest-tier provenance in endpoint state, so candidate construction is linear in projected-key memberships and allocations.
- Clarifies that a tier ID uniquely names its configured confidence; distinct IDs are distinct tiers.
- Keeps future fuzzy/similarity tiers gated on LSH blocking and an explicit comparison cap.

## Evidence

- A 9,000 × 9,000 shared-projection collision bucket completes without materializing its 81 million possible pairs and remains conservatively unmatched.
- 500 deterministic randomized cases compare compressed output to the prior brute-force candidate graph across both ordered and identity-set modes.
- Full `dotnet-inspect.slnx` Release build passes.
- `FindingPilotTests`: 66 passed.
- `MetadataFindingsTests`: 24 passed.
- Changed Finding design docs pass markdownlint.

## Adversarial review

- **Gemini first pass:** CLEAN.
- **MAI first pass:** identified per-accepted-pair projection dictionary churn. Fixed in `deed5e14` by retaining strongest-tier provenance directly in the bounded endpoint state.
- **Final contract pass:** a same-ID/different-confidence suggestion was not adopted because one ID names one semantic tier and confidence; `86ce540f` makes that invariant explicit and adds coverage.
- **Final head `86ce540f`: Gemini CLEAN; MAI CLEAN.**

### DIFF
--- a/docs/design/finding-nomenclature.md
+++ b/docs/design/finding-nomenclature.md
@@ -107,6 +107,12 @@ consumer fold. At threshold 100, soft candidates remain one `Added` and one
 `Removed`; an accepted soft candidate becomes `Changed` and retains its named
 tier and confidence through `FindingMatchProvenance`.
 
+For structured equality tiers, the matcher blocks on tier plus projected
+identity and summarizes each endpoint's degree without materializing all pairs
+inside a collision bucket. Only mutual degree-one endpoints survive. This keeps
+the false-positive discipline and runtime bound aligned: ambiguity costs input
+size, not the product of both residual streams.
+
 ## Sparse correlation and onset
 
 Neither correlation outcome traverses a version range. The caller chooses which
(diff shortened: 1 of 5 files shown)


richlander:
Ready to merge. All merge-blocking checks pass, and Gemini and MAI both returned CLEAN on final head `86ce540f`.

Issue #2746, opened 2026-07-14
Reconstruct direct generic (TypeSpecification) base chains in the CSharpTypeProducer compile-back skeleton (residual of #2678/#2734)


richlander:
## Opportunity

Extend the compile-back type skeleton to reconstruct **direct generic (TypeSpecification) base chains** for the RTS/ReturnToSender compile-back oracle, so types deriving from a same-assembly generic base can round-trip to `Exact` instead of `RecompileFail`.

This is the orchestrator-compatible remainder of closed PR #2734 (#2678). That PR measured **seven generic method rows moving `RecompileFail → Exact`** on `NuGet.Versioning` by resolving direct generic base `TypeSpecification`s to their same-assembly definitions and rendering the constructed base signature. On current `main` those rows are still `RecompileFail` (e.g. `NuGet.Shared.EqualityUtility::OrderedEquals`, `SequenceEqualWithNullCheck`, `NuGet.NoAllocEnumerateExtensions::NoAllocEnumerate`).

## Why it is not already covered

#2737 moved base-type reconstruction into the product seam `ILInspector.CSharp/CSharpTypeProducer.ReconstructedBaseTypeName`, but that method **deliberately drops generic bases**:

> generic-instantiation (TypeSpecification) bases are dropped … A generic type's base can reference its own type parameters, which the flat shell does not carry, so skip it.

So the generic-base case is a known, intentionally-deferred gap, not a regression.

## Direction (per #2613)

Do this the **product/orchestrator** way, not by re-adding C# knowledge to the harness:

- Extend `CSharpTypeProducer` (the #2737 seam) to reconstruct a direct **same-assembly** generic base `TypeSpecification` — resolve to its definition and render the constructed base signature — while carrying the type parameters the flat shell needs (the reason #2737 skipped it).
- Keep the harness a thin forwarder (`ReturnToSenderTypePlanner.BaseTypeSignature`), consistent with #2739's merged principle that RTS "deliberately does NOT model C# overload resolution."
- Closed PR #2734's branch `fix/issue-2678-constructor-parity-residuals` (commit `f5f5ee7a`, "Handle generic RTS base chains") has concrete fixtures and the "keep the last lifted initializer for duplicate field stores" detail — reuse them as product-side tests.

## Not needed from #2734

[...]

Issue #2745, opened 2026-07-14
RTS shell reconstruction drops argument-type interface implementations → conservative OpcodeDiff on faithful interface-upcast chain args


richlander:
## Summary

The RTS (ReturnToSender) compile-back shell reconstruction stubs an argument's type **without its interface implementations**. When a constructor-chain argument is upcast to an interface it implements (`C(A a) : this((IDisposable)a)` where `A : IDisposable`), the product printer faithfully emits `(IDisposable)a`, but the reconstructed shell renders `class A {}` (no `: IDisposable`). Roslyn then compiles the cast as an **explicit reference conversion** (`castclass`) instead of the original **implicit no-op**, so RTS classifies the method as **OpcodeDiff** even though a full-graph recompile would be byte-identical (Exact).

This is a **conservative** misclassification (RTS *under-reports* Exact — it never claims fidelity it can't prove), so it is not a soundness bug. It surfaced during adversarial review of #2739 (GPT-5.5), which confirmed it is a pre-existing shell-reconstruction limitation, independent of the product chain-fidelity-cast fix.

## Reproduction

Fixture:

```csharp
using System;
public class A : IDisposable { public void Dispose() {} }
public class C
{
    public C(IDisposable d) {}
    public C(A a) : this((IDisposable)a) {}
}
```

Target `C..ctor` overload 1 via `ReturnToSender.CompileBackTargets`:

- `Status: OpcodeDiff`
- Original opcodes: `ldarg ldarg call ret`
- Recompiled opcodes: `ldarg ldarg castclass call ret`

Generated RTS shell (note the dropped interface on `A`):

```csharp
public class C
{
    public C(A a) : this((IDisposable)a) { }
    public C(System.IDisposable d) { throw null; }
}
public class A { }
```

Same-assembly interface and base-class upcasts already classify Exact — the failure is specific to an argument type whose interface implementation the shell does not reconstruct.

## Proposed fix

When reconstructing an argument type's stub, reconstruct (or otherwise prove) the interface implementations required for the emitted upcasts to compile as implicit conversions, so these chains classify Exact rather than OpcodeDiff. Scope is the harness shell reconstruction (`ReturnToSenderTypePlanner`), not the product printer.

## Notes

[...]

PR #2751, opened 2026-07-14
Route the RTS chain-parameter surface check through the CSharpTypeProducer seam (#2613)

### LINKED ISSUE #2731: RTS: reconstruct generic base interfaces for shell interface types (CS0176 inherited-member gap)

richlander:
## Summary

RTS reversal increment 2 (follows #2737, #2739). The compile-back harness
re-inlined the "surface not representable in C# source" heuristic
(`delegate*` / `<>` / `{` substring checks) that #2737 relocated into the product
as `CSharpTypeProducer.IsUnsupportedSurfaceSignature`. This routes the
chain-parameter site through the product seam:

- `IsUnsupportedChainParameterType` — the `: this(args)` chain-arg gate.

Every RTS consumer that judges chain-arg surface now shares the product's
definition, and one more piece of C# knowledge moves out of the orchestrator and
into `ILInspector.CSharp`, where recompilation validates it. This is the ongoing
#2613 goal: RTS as a C#-free orchestrator between the product and Roslyn.

### Scope note (single site — see reconciliation below)

This increment initially migrated a second site (the primary-ctor backing-field
gate), but adversarial review showed that site passed **raw** signature text to a
seam whose contract expects a normalized C# display name, turning a duplicate
consolidation into an arm-set change. It was reverted (commit `8c17ab57`); only
the chain-parameter site ships here. A correct routing of the field-type gate
needs display-form normalization and is tracked with the other #2613 roadmap
items (generic base chains #2746, type-shell header, member-closure composition,
`CompileBackCSharpNames.Clean`).

## Neutrality

Provably neutral. `IsUnsupportedChainParameterType` already fed the seam a
normalized `type.ToDisplayString()`; its old arms (`delegate*` / `<>` / `{`) are
exactly the seam's arms, and the only extra seam arm (`@delegate*`) is subsumed
by `delegate*` (dead). The seam performs no normalization. Same input, same
truth-space.

**Corpus RTS-parity A/B** (14 assemblies, snapshots cap 200):

| ref | Exact | OpcodeDiff | RecompileFail | changed methods |
|---|---|---|---|---|
| base (`origin/main`) | 1357 | 194 | 581 | — |
| head | 1357 | 194 | 581 | **0 / 89,065** |

`ReturnToSenderPrototypeTests`: **114 passed**.

## Scope

Does **not** close #2731 / #2745 (interface projection). Those need product-path
consumed-member / receiver-type evidence, not a surface relocation — tracked
separately.

## Testing

- `dotnet build tools/DecompilerHarness -c Release` ✅
- `dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "…ReturnToSenderPrototypeTests"` → 114 passed ✅
- 14-asm corpus RTS-parity A/B → 0 changed / 89,065 ✅

Two CLEAN adversarial reviews (GPT-5.5 + Gemini 3.1 Pro) on the scoped-down head; see reconciliation comments.

### DIFF
--- a/tools/DecompilerHarness/ReturnToSenderTypePlanner.cs
+++ b/tools/DecompilerHarness/ReturnToSenderTypePlanner.cs
@@ -1669,12 +1669,10 @@ public static class CompileBackSourceComposer
     /// <c>: this(args)</c> initializer targeting it cannot bind.
     /// </summary>
     static bool IsUnsupportedChainParameterType(TypeRef type)
-    {
-        string display = type.ToDisplayString();
-        return display.Contains("delegate*", StringComparison.Ordinal)
-            || display.Contains("<>", StringComparison.Ordinal)
-            || display.Contains('{', StringComparison.Ordinal);
-    }
+        // Route surface-representability through the product seam (CSharpTypeProducer)
+        // so every RTS consumer judges droppable surfaces identically, instead of
+        // re-inlining the delegate*/<>/{ heuristic here.
+        => CSharpTypeProducer.IsUnsupportedSurfaceSignature(type.ToDisplayString());
 
     static CompileBackParameter ToCompileBackParameter(ApiParameter parameter)
         => new(
@@ -2308,8 +2306,7 @@ public static class CompileBackSourceComposer
                 return null;
             }
 
-            if (fieldType.Contains("delegate*", StringComparison.Ordinal)
-                || fieldType.Contains("@delegate*", StringComparison.Ordinal))
+            if (CSharpTypeProducer.IsUnsupportedSurfaceSignature(fieldType))
                 return null;
 
             parameters.Add(new CompileBackParameter(
--- a/tools/DecompilerHarness/ReturnToSenderTypePlanner.cs
+++ b/tools/DecompilerHarness/ReturnToSenderTypePlanner.cs
@@ -2306,7 +2306,8 @@ public static class CompileBackSourceComposer
                 return null;
             }
 
-            if (CSharpTypeProducer.IsUnsupportedSurfaceSignature(fieldType))
+            if (fieldType.Contains("delegate*", StringComparison.Ordinal)
+                || fieldType.Contains("@delegate*", StringComparison.Ordinal))
                 return null;
 
             parameters.Add(new CompileBackParameter(


richlander:

Two reviewers (isolated worktrees, non-Claude):

- **GPT-5.5 — CLEAN.** Verified `IsUnsupportedSurfaceSignature` has no normalization; both sites forwarded existing input; superset arms only. Built harness, ran `ReturnToSenderPrototypeTests` (114/114) and `ILInspector.CSharp.Tests` (98/98).
- **Gemini 3.1 Pro — HIGH on the primary-ctor field-type site.** Correctly identified that that site passed **raw** `GuardedSignatureText.FieldText` to the seam (whose contract expects a normalized C# display name). Routing it through the seam **adds** `<>`/`{` arms to a gate that previously checked only `delegate*`/`@delegate*`, so a primary constructor whose backing-field type carries a raw `<>` segment could be dropped before `CompileBackTypeSignature.Display` → `Clean` sanitizes it. That is an arm-set change, not a duplicate consolidation.

**Where they diverged:** GPT judged the flipped shapes as legitimately unsupported and found no reproducible regression; Gemini identified the mechanism by which the flip could regress reconstruction. The corpus A/B (0 changed / 89,065) only proves neutrality on the **sampled** population (2,132 methods; RTS fidelity is capped at 200/asm), so it does not falsify Gemini's finding.

**Resolution (commit `8c17ab57`):** scoped down to the **chain-parameter site only** (`IsUnsupportedChainParameterType`). That site is the genuine duplicate — its old arms (`delegate*`/`<>`/`{`) exactly match the seam, the extra `@delegate*` arm is subsumed by `delegate*`, and it already forwards a normalized `type.ToDisplayString()`. It is provably, exactly neutral. Reverted the primary-ctor field-type site to its original check; a correct routing there needs display-form normalization (matching the wired site 3), tracked with the other #2613 roadmap items.

## Final state

- Net diff vs `origin/main`: **one site** (`IsUnsupportedChainParameterType` → `CSharpTypeProducer.IsUnsupportedSurfaceSignature(type.ToDisplayString())`).
- Provably neutral: old arms ⊆ seam arms; extra arm dead; identical input; no normalization in the seam.
- `ReturnToSenderPrototypeTests`: **114 passed** at the revert.
- Still does **not** close #2731 / #2745.

richlander:
## Two CLEAN reviews on the scoped-down head (`8c17ab57`)

- **GPT-5.5 — CLEAN** (re-confirmed): no normalization in the seam; single-site change is a superset consolidation.
- **Gemini 3.1 Pro — CLEAN** (re-confirmed): verified its flagged site (primary-ctor backing-field) is fully reverted; the remaining single site is provably neutral — identical input `type.ToDisplayString()`, zero normalization, old arms `delegate*`/`<>`/`{` exactly preserved, the extra `@delegate*` arm dead (subsumed by `delegate*`).

Both reviews ran in isolated worktrees (`git status` clean), built the harness, and (GPT) ran `ReturnToSenderPrototypeTests` 114/114.

**Ready to merge** — pending @richlander approval. Scope: one site (`IsUnsupportedChainParameterType` → `CSharpTypeProducer.IsUnsupportedSurfaceSignature`). Does not close #2731/#2745; the primary-ctor field-type site and remaining #2613 roadmap items (generic base chains #2746, type-shell header, member-closure composition, `CompileBackCSharpNames.Clean`) stay open.

richlander:
## Review — one finding (body/scope mismatch; implemented code is CLEAN)

**The PR body claims two migrated sites, but the net branch diff (`bc0babce..HEAD`) contains only one.**

The body says:
> This routes **both sites** through the product seam:
> - `IsUnsupportedChainParameterType` — the `: this(args)` chain-arg gate.
> - the primary-ctor backing-field parameter-type gate.

and, under Neutrality:
> the **field-type gate adds `<>`/`{`**, which fire only on … generated/anonymous primary-ctor field types.

Only the chain-parameter gate (`IsUnsupportedChainParameterType`, ll. 1671-1675) was actually changed. The primary-ctor backing-field gate is unchanged and still inlines the raw heuristic — no seam call, and no `<>`/`{`:

```csharp
// tools/DecompilerHarness/ReturnToSenderTypePlanner.cs:2308-2310 (HEAD)
if (fieldType.Contains("delegate*", StringComparison.Ordinal)
    || fieldType.Contains("@delegate*", StringComparison.Ordinal))
    return null;
```

This is the primary-ctor backing-field path (immediately after `TryPrimaryConstructorParameterName`, l. 2293). It does **not** route through `CSharpTypeProducer.IsUnsupportedSurfaceSignature`, directly contradicting bullet 2 and the "field-type gate adds `<>`/`{`" neutrality rationale.

**Impact:** the one implemented change is correct and genuinely neutral — `IsUnsupportedSurfaceSignature` is a faithful superset of the old chain check (adds `@delegate*`), same raw `type.ToDisplayString()` input, and the 0/89,065 corpus A/B backs neutrality. **No correctness regression.** But the PR under-delivers its stated goal: the duplication it claims to remove "in two places" still exists at l. 2308, and that gate diverges from the seam (a primary-ctor backing field typed with `<>`/`{` would slip through — pre-existing behavior, not introduced here, unobserved on the corpus).

**Recommendation (pick one before merge):**
- Add the missing migration at ll. 2308-2310 — replace the inline check with the local seam wrapper (`=> IsUnsupportedSurfaceSignature(fieldType)`, matching l. 2824 / the wrapper at l. 3684), which is what makes the "field-type gate adds `<>`/`{`" claim true; **or**
- Amend the body to describe the single site actually migrated and drop the "field-type gate adds `<>`/`{`" claim.

Verdict: **implemented code CLEAN; body/scope inaccurate — flag before merge.**

<sub>Reviewed by Claude Opus 4.8.</sub>

richlander:
## Addressed — body corrected to single-site scope

Thanks @claude (Opus 4.8) — accurate catch. The finding is a body/scope mismatch, not a code defect: the implemented code is CLEAN and neutral, but the PR body was stale from before the scope-down.

I took the second recommended option (amend the body) rather than re-adding the field-type migration, because the field-type site was **deliberately** reverted after Gemini 3.1 Pro showed it passed raw signature text to a seam expecting a normalized display name (an arm-set change, not a duplicate consolidation). Re-adding it would reintroduce exactly that risk without the display-form normalization it needs.

The body now:
- describes the single migrated site (`IsUnsupportedChainParameterType`) only,
- drops the "both sites" and "field-type gate adds `<>`/`{`" claims,
- keeps a scope note explaining the reverted second site and links the follow-up roadmap.

The l. 2308 inline check remains intentionally unchanged; correct routing of that gate needs display-form normalization and stays tracked with the #2613 roadmap items. No code change in this update.

Issue #2742, opened 2026-07-14
Decompiler: reconstruct runtime async (MethodImplAttributes.Async) bodies


richlander:
## Problem

The decompiler does not reconstruct **runtime async** methods (`MethodImplAttributes.Async`, no compiler-generated state machine). Only classic state-machine async is raised (`ClassicAsyncReconstructionPass`, which sets `RequiresAsyncBodyModifier`).

For runtime-async bodies:

- the `async` modifier is dropped (neither `ContainsAwaitExpression` nor `RequiresAsyncBodyModifier` is set), and
- the body exposes the raw runtime-async lowering (`AsyncHelpers.UnsafeAwaitAwaiter<...>`, `GetAwaiter`/`IsCompleted`/`GetResult`) instead of `await`.

### Verified repro (this SDK)

Composing an `async Task` / `async Task<int>` fixture built with SDK `11.0.100-preview.5` via `TypeSourceComposer.Compose(...)` renders, e.g.:

```csharp
public Task<int> AwaitingAsync()
{
    YieldAwaitable V_1 = Task.Yield();
    YieldAwaitable.YieldAwaiter V_0 = V_1.GetAwaiter();
    if (!V_0.IsCompleted)
    {
        AsyncHelpers.UnsafeAwaitAwaiter<YieldAwaitable.YieldAwaiter>(V_0);
    }
    V_0.GetResult();
    return 1;
}
```

Expected: `public async Task<int> AwaitingAsync() { await Task.Yield(); return 1; }`.

Note: the classic compiler state-machine async remains the language default; runtime async is a distinct emission shape (seen here from the preview SDK). The gap applies whenever runtime-async shapes appear.

## Secondary: async-signal inconsistency (classic async)

`TypeSourceComposer` forces `async` from `ContainsAwaitExpression` alone, while `ApiOutputFormatter` uses `ContainsAwaitExpression || RequiresAsyncBodyModifier`. For a classic-async method whose `await` is not reachable, the composer drops `async` where `ApiOutputFormatter` would not. Aligning the two is a small consistency fix (should ship with a classic-async fixture that exercises `RequiresAsyncBodyModifier` without `ContainsAwaitExpression`).

## Scope / discipline

Runtime-async recovery is a decompiler raise: follow `docs/decompiler-raise-discipline.md` (render-text A/B against merge-base, classify changed methods, adversarial review). Add both a real compiled runtime-async canary and near-miss negatives.

## Context

[...]

PR #2752, opened 2026-07-14
Source body-only modifiers from resolved method facts

richlander:
Advances #2742. Follow-up to #2737 and #2750.

## Change

**Conclusion: PASS** — full-body C# declarations now source body-only `async`
and `unsafe` modifiers from resolved method/body facts, while API skeletons
remain unchanged.

Before:

```csharp
public static Task<int> PlainAsyncTask(int value)
```

After:

```csharp
public static async Task<int> PlainAsyncTask(int value)
```

The same contract now drives `TypeSourceComposer`, member decompiled source, and
annotated/cost/semantics overlays:

- `CSharpMemberBody` carries body-only async/unsafe requirements.
- `CSharpTypeProducer` classifies runtime async and classic async from the
  resolved defining MethodDef.
- Async-iterator kickoff methods remain without `async` until their bodies are
  reconstructed to source `yield` forms.
- Invalid carried tokens resolve the legacy name/ordinal selector to its concrete
  MethodDef before body, attribute, overlay, IL, or modifier projection.
- `DecompilerResult` carries typed unsafe-body evidence from the printer's
  existing unsafe-context discriminator.

This does not implement #2742's runtime awaiter-helper body reconstruction.

## Evidence

| Check | Result |
| --- | --- |
| Solution build | Pass |
| CSharp tests | 100 passed |
| Type-source tests | 61 passed |
| Unsafe-emitter tests | 38 passed |
| IrImporter tests | 65 passed |
| Product tests | 1,902 passed; 10 expected external-tool skips |
| Debug classic-async canary | 2 passed |
| Markdown lint | Pass |

Latest-merge-base targeted A/B:

- `OptimizationOpportunityFixtures.PlainAsyncTask`: only the `async` header was
  added.
- `RuntimeAsyncHeaderFixture.YieldAsync`: only the `async` header was added.
- `AsyncStream` and `YieldsPlainObjectAsync`: unchanged async-iterator near
  misses.

Render-text A/B evaluated 89,065 methods: **0 changed, 0 added, 0 removed**.
This is expected because the behavior change is in outer member declarations,
not decompiler body text.

## Review

| Reviewer | Result | Reconciliation |
| --- | --- | --- |
| Gemini 3.1 Pro | Clean on `3317c828` | Re-attacked resolved handles, token fallback, iterator gating, unsafe polarity, overlays, and skeleton isolation |
| MAI-Code-1-Flash | Clean on `3317c828` | Rebuilt and reran focused CSharp, decompiler addressing, and product rendering coverage |

Earlier adversarial findings were resolved as follows:

- MAI found that product member views dropped body-only unsafe evidence. Fixed
  in `13e5a0e7` by carrying the typed fact through every result/overlay.
- Gemini found struct-`this` false positives, `[SkipLocalsInit]` stackalloc false
  negatives, and a Release-specific fixture path. Fixed in `76e37cc7` with
  polarity tests and Debug/Release path resolution.
- Gemini found declaration/body drift when an invalid token fell back to a
  name/ordinal body but not async classification. Fixed in `700eb5cb` by
  resolving that selector to the concrete MethodDef before every projection.
- MAI suggested removing async iterators from
  `MethodClassificationScanner.StateMachineAsync`. No change: that scanner is
  intentionally a coarse async/iterator classifier; `CSharpTypeProducer` is the
  sharper body-capability gate and explicitly excludes iterator-only metadata.

**Review conclusion: PASS** — both required reviewers are clean on the final
head with no outstanding findings.

### DIFF
--- a/docs/design/member-body-substrate.md
+++ b/docs/design/member-body-substrate.md
@@ -124,7 +124,8 @@ The same-reader body-composition callers are migrated onto handle addressing:
   the surface member's validated metadata token once and threads it into
   attributes, generic-parameter names, `HasBody`, decompiled source, the
   `ResearchViews` mixed IL+C# projection (`MemberProjectionRequest.MethodHandle`),
-  and IL disassembly. Layer overloads were added at each seam
+  body-only modifier classification, and IL disassembly. Layer overloads were
+  added at each seam
   (`AttributeReader.GetMethodAttributes`, `ILInstructionPrinter.DisassembleMethod`,
   `IlProjection.Locate/Project/AnnotatedInstrLines`). Other `ResearchViews` entry
   points keep the name path (nil handle → fallback) — they are not the CLI's drift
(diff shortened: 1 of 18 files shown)


richlander:
Ready to merge. CI is green, the latest-main A/B is clean, and Gemini 3.1 Pro plus MAI-Code-1-Flash are both clean on final head `3317c828`.

PR #2758, opened 2026-07-15
Remove legacy test-only ResearchViews render methods; fold onto ProjectMember

richlander:
## Summary

Following the SourceLine currency adoption (#2757), **every real product source and IL renderer already routes through the unified `ResearchViews.ProjectMember` entry point.** This PR removes the leftover legacy surface that no longer has product callers and consolidates the tests that exercised it.

No product behavior changes — this is dead-public-API removal plus test migration/consolidation.

## What changed

- **Removed 6 legacy public methods on `ResearchViews`** (no product callers — only tests): `RenderCostOverlay`, `RenderCostOverlayWithHeaderFacts`, `RenderSemanticsOverlay`, `RenderAnnotatedSource`, `RenderMixed`, `ProjectAnnotatedIl`, plus the now-dead private helpers `AddFactsToAnnotatedIl` and `TryReadOffset`.
- **Migrated `ResearchFactRegistryTests` onto `ProjectMember`** via small test-local helpers.
- **Folded `AnnotatedCSharpRendererTests` + `AnnotatedIlFactTests` into `MixedSourceRendererTests`** — all three drove the same mixed view — dropping three duplicate assertions.
- **Deleted two obsolete tests** (`AnnotatedSource_DoesNotIncludeCostOverlayByDefault`, `AnnotatedSource_DoesNotIncludeSemanticsOverlayByDefault`). They asserted that cost/semantics facts are *absent* from the annotated view, but only because the removed `RenderAnnotatedSource` helper built its `ResearchFactContext` **without** the `ResearchAssemblyContext`, so cross-method facts structurally never appeared. The real product/CLI annotated view (`ProjectMember`) resolves the assembly and **does** co-render those facts — so the assertions never matched product behavior. Both reviewers reproduced this against the live CLI.

## Why this matters / what it unlocks

`ProjectMember` is now the single, unambiguous entry point for member source/IL projection. Removing the parallel legacy methods eliminates a second code path that could drift from the currency, keeps the annotated/overlay behavior tested against what the product actually renders (not a test-only fact context), and shrinks the public surface future consumers have to reason about.

## Not in scope (intentionally)

RTS / compile-back (`CompileBackSourceComposer`) composes a whole compilable unit via `CSharpPrinter.PrintRaised` and stays deliberately **off** the line currency — it is a distinct experience, not a member-projection consumer.

## Validation

- Full `dotnet-inspect.slnx` build: clean.
- `ILInspector.Research.Tests`: 109 pass.
- `ILInspector.Decompiler.Tests` fast subset (`-notrait Speed=Slow`): 2426 pass.
- Folded `MixedSourceRendererTests`: 18 pass.

## Adversarial review

Two independent reviewers (isolated worktrees at the PR head), both **CLEAN**:

- **GPT-5.5** — no non-test callers of any removed symbol; reproduced the live CLI annotated view showing `cost.callee` / `semantics.callee` do appear (confirming the deleted negative tests were false negatives); build + suites green.
- **Gemini 3.1 Pro** — no product callers; verified `MemberCodeProvider` already routes through `ProjectMember`; confirmed the assembly-context justification for the deletions and that the folded assertions preserve coverage; build + suites green.

No findings from either reviewer.


richlander:
### Adversarial review — reconciled

Two reviewers from non-Opus families (author model: Claude Opus 4.8), each in an isolated worktree checked out at the PR head `e705d5e7`. **Both CLEAN, no findings.**

| Reviewer | Result | Key verification |
| --- | --- | --- |
| **GPT-5.5** | CLEAN | Repo-wide grep: no non-test callers of any removed symbol. Reproduced the live CLI annotated view (`member … CallsAllocInLoopCallee -S "Annotated Source"` → `cost.callee`; `CallsExceptionOnlyCallee` → `semantics.callee`), confirming the two deleted negative tests were false negatives. Build clean; Research 109, Decompiler fast 2426, MixedSourceRendererTests 18 — green. |
| **Gemini 3.1 Pro** | CLEAN | No product callers; `MemberCodeProvider` already routes through `ProjectMember`. Verified old `RenderAnnotatedSource` called `CollectFacts` **without** `ResearchAssemblyContext` (silently omitting cross-method facts) while `ProjectMember` passes it — justifying the deletions. Folded assertions preserve coverage. Build + suites green. |

Where they agreed: everything — no product regression, deletions justified, coverage preserved. No divergence. No follow-up commits required (no actionable findings).

Issue #3, opened 2026-02-05
Add a command to dump references from an assembly


ericstj:
Given an assembly dump its assembly references.  Potentially this can be included with the assembly command.

PR #2762, opened 2026-07-15
Unify C# type-keyword escaping onto the product seam (de-C#-ify RTS)

richlander:
## Why

RTS (ReturnToSender) is the compile-back **oracle** that declares the product decompiler's C# output valid/passing. It had reimplemented C# reserved-keyword escaping in `CompileBackCSharpNames`, and that duplicate had **drifted** from the product's authoritative rules in *both* directions:

- RTS's local `IsCSharpKeyword` list **omitted `await`**, so a type named `await` emitted bare `await` — while the product correctly emits `@await`.
- The product's own `EscapeTypeKeywords` wrongly escaped `ref readonly int` (as `ref @readonly int`), `int.MaxValue` (as `@int.MaxValue`), and left dotted `N.int` bare instead of `N.@int`.

Because RTS judges the product against *its own* spelling, any such drift silently contaminates the pass/fail verdict in both directions — the oracle can bless invalid product output or fail valid output. Removing C# knowledge from RTS so it defers to the product's C# layer is a standing goal; this is the escaping slice of it.

## What this enables

- **One authoritative type-keyword escaper.** `CSharpDeclarationWriter.EscapeTypeKeywords` is now the single implementation, surfaced through `CSharpFormatter.EscapeTypeKeywords`. There is no second copy to drift.
- **A trustworthy oracle for this dimension.** RTS now escapes exactly as the product does, so its valid/passing verdict reflects the product's real C# layer rather than RTS's divergent guess.
- **A template for the rest of the RTS de-C#-ification.** RTS still carries `Clean`'s `System.X` primitive aliasing, a local `TypeProducer`, `ToMemberPolicy`, and `ParseConstructorInitializer`. This PR establishes the pattern (product owns the seam, RTS delegates) those follow-ups will reuse.

## Change

- `src/ILInspector.CSharp/CSharpDeclarationWriter.cs` — `EscapeTypeKeywords` made `internal`; `IsTypeSyntaxKeyword` rewritten with a corrected, unified model plus helpers `IsPrimitiveOrVoidKeyword` and `InModifierPosition`:
  - primitive/void aliases bare unless a dotted-name segment (`N.int` → `N.@int`);
  - parameter/type modifiers (`ref/in/out/params/readonly/scoped/unmanaged`) bare only in a leading modifier run followed by a real type;
  - `delegate*` bare only as a genuine function-pointer head (`<` or a calling-convention run after the `*`).
- `src/ILInspector.CSharp/CSharpFormatter.cs` — new public facade `EscapeTypeKeywords`.
- `tools/DecompilerHarness/ReturnToSenderTypePlanner.cs` — `CompileBackCSharpNames.Clean` delegates to the product; local escaping helpers (`EscapeTypeKeywords`, `PreviousWordIsRef`, `HasFollowingWord`, `IsPrimitiveTypeName`, `IsCSharpKeyword`) deleted.
- `src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs` — new theory covering the union of both old behaviors plus the pointer/keyword edge cases surfaced in review.

## Validation

- `ILInspector.CSharp.Tests`: 140 pass (incl. the new escaping theory).
- Core RTS oracle tests (`ReturnToSenderPrototypeTests`, `ReturnToSenderEvidenceTests`): 117 pass.
- Corpus quality-diff card: **neutral on every raise/fidelity metric** (escaping-only change; the only movement is more-correct escaping where a keyword-named type appears). Card was run with reduced sampling caps (`--compile-cap 25`, fidelity cap 3) for turnaround, so its "regressions" line reflects only the lower caps vs baseline, not behavior.

## Review

Adversarial review per policy — two reviewers from non-Opus families (I authored as Opus): **GPT-5.5** and **Gemini 3.1 Pro**, in isolated worktrees, plus an **MAI** pass for extra rigor on this central-file + oracle change. Reviews drove four rounds of real fixes (pointer-to-keyword-typed names: `ref*`→`@ref*`, `delegate*`→`@delegate*`, whitespace-then-punctuation, and preserving `delegate* <...>`). Reconciliation posted below once all sign-offs land.

### DIFF
--- a/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
@@ -202,6 +202,14 @@ public sealed class CSharpFormatterTests
     [InlineData("delegate* unmanaged<int>", "delegate* unmanaged<int>")]
     [InlineData("delegate*<ref int, void>", "delegate*<ref int, void>")]
     [InlineData("delegate*<await, void>", "delegate*<@await, void>")]
+    // Pointers to types literally named like a keyword must be escaped, not read as
+    // type syntax: "ref*"/"in*" are pointers to a type named ref/in, and a bare
+    // "delegate*" (not a function-pointer head) is a pointer to a type named delegate.
+    [InlineData("ref*", "@ref*")]
+    [InlineData("in*", "@in*")]
+    [InlineData("readonly*", "@readonly*")]
+    [InlineData("delegate*", "@delegate*")]
+    [InlineData("delegate*[]", "@delegate*[]")]
     // Already-escaped identifiers are left untouched (idempotent).
     [InlineData("@int", "@int")]
     [InlineData("N.@int", "N.@int")]
--- a/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
@@ -168,4 +168,43 @@ public sealed class CSharpFormatterTests
                 NamespacePolicy = (CSharpNamespacePolicy)42
             }));
     }
+
+    [Theory]
+    // Primitive/void aliases stay bare when they name the primitive.
+    [InlineData("int", "int")]
+    [InlineData("string", "string")]
+    [InlineData("void", "void")]
+    [InlineData("int[]", "int[]")]
+    [InlineData("int*", "int*")]
+    [InlineData("System.Int32", "System.Int32")]
+    [InlineData("System.Collections.Generic.List<int>", "System.Collections.Generic.List<int>")]
+    // A type literally named after a primitive keyword (a dotted-name segment) is escaped.
+    [InlineData("N.int", "N.@int")]
+    [InlineData("int.MaxValue", "int.MaxValue")]
+    // Reserved keywords used as identifiers are escaped, including inside generic args.
+    [InlineData("class", "@class")]
+    [InlineData("await", "@await")]
+    [InlineData("record", "@record")]
+    [InlineData("List<await>", "List<@await>")]
+    [InlineData("MyType<class>", "MyType<@class>")]
+    [InlineData("Foo.await", "Foo.@await")]
+    [InlineData("await.Foo", "@await.Foo")]
+    [InlineData("N.readonly", "N.@readonly")]
+    // Parameter/type modifiers stay bare in a leading modifier run.
+    [InlineData("ref int", "ref int")]
+    [InlineData("ref readonly int", "ref readonly int")]
+    [InlineData("scoped ref int", "scoped ref int")]
+    [InlineData("in long", "in long")]
+    [InlineData("out string", "out string")]
+    [InlineData("params byte[]", "params byte[]")]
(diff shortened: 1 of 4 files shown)


richlander:

Three independent reviewers in isolated worktrees (I authored as Opus 4.8, so all three are other families): **GPT-5.5**, **Gemini 3.1 Pro**, and an **MAI** pass for extra rigor on this central-printer + oracle change. Review drove five rounds of real fixes; all three are now **CLEAN on the final head `157414be`** with empirical evidence (rebuild + probe of `EscapeTypeKeywords`, plus C#-compiler confirmation of the expected spellings).

### Findings, all fixed and re-verified

| # | Reviewer(s) | Finding | Kind | Fix commit |
| - | ----------- | ------- | ---- | ---------- |
| 1 | GPT-5.5 **and** Gemini (independently) | `ref*` / `in*` / `readonly*` (pointer to a type named like a modifier) left bare → invalid `ref*` (CS1031) | **PR-introduced** (I had added `*` to the modifier boundary) | `305bf8f4` |
| 2 | Gemini | bare `delegate*` (pointer to a type named `delegate`, not a fn-pointer head) left bare → CS1003 | pre-existing, carried into the unified path | `305bf8f4` |
| 3 | Gemini | whitespace-then-punctuation: `ref *`, `Tuple<readonly >`, `(delegate* , int)`, `Tuple<delegate* >` misclassified | robustness gap | `de9588c9` |
| 4 | Gemini | `delegate* <int, void>` (space before `<`, a *valid* fn pointer) was over-escaped → CS1519 | regression from fix #3 (I over-tightened) | `97ecfa14` |
| 5 | MAI | qualified `N.delegate*<int, void>` left bare — inconsistent with the primitive `N.int`→`N.@int` rule | pre-existing inconsistency | `157414be` |

Where reviewers **agreed**: GPT-5.5 and Gemini independently reported finding #1 with the same repro and the same fix (drop `*` from the modifier boundary), which is strong corroboration. Where they **diverged**: Gemini went deeper on function-pointer/whitespace permutations (#2–#4); MAI, reviewing fresh at the end, caught the qualified-segment inconsistency (#5) the other two missed.

Finding #4 warranted a compiler check rather than taking the report at face value: I confirmed with the C# compiler that bare `delegate* <int, void>` compiles, so the fix keeps it bare (it does not blindly escape).

### Reachability note

Findings #2–#5 concern type-string spellings (intra-type whitespace like `ref *`, or a namespace-qualified/pointer `delegate`) that the product and RTS renderers do not emit — real forms are `ref*`, `delegate*,`, `delegate*>`, all handled. All three reviewers confirmed this explicitly. They are fixed anyway because this is now the **single authoritative** escaper for both the product C# layer and the RTS oracle, and it should never emit invalid C#.

### Validation on the final head

- `ILInspector.CSharp.Tests`: **142 pass** (incl. the union + edge-case theory).
- Core RTS oracle tests (`ReturnToSenderPrototypeTests`, `ReturnToSenderEvidenceTests`): **117 pass**.
- Corpus quality-diff card: **neutral on every raise/fidelity metric** (escaping-only).
- Rebased on latest `origin/main` (no conflicts; upstream's `ReturnToSenderTypePlanner.cs` edit was in an unrelated region).

Not merging without your approval.

PR #2763, opened 2026-07-15
De-C#-ify RTS: move remaining C# spelling/shape decisions to the product seam

richlander:
## Summary

Move the last of RTS's re-implemented C# spelling/shape decisions into the
product's C# seam, so the compile-back oracle judges the product decompiler
against the **product's** C# spelling instead of RTS's own divergent copy.

## Why this matters

RTS (ReturnToSender) is the compile-back oracle: it takes the product
decompiler's reconstructed C#, compiles it, and compares IL to declare the
product's output valid. That verdict is only trustworthy if RTS reproduces the
product's C# faithfully. Where RTS re-implemented its own spelling/shape rules,
it was silently grading the product against a *different* dialect — a
load-bearing conflict of interest in the thing that certifies the decompiler as
correct. Every rule RTS owns independently is a way for the oracle to disagree
with the product for reasons that have nothing to do with the decompiler.

This PR removes that duplication for the remaining cases:

- **Primitive spelling** — RTS aliased only `String`/`Int32`/`Void`; everything
  else (`System.Decimal`, `System.IntPtr`, …) stayed fully-qualified, unlike the
  product which spells them `decimal`/`nint`/etc. Now a single product aliaser
  (driven by the one `PrimitiveTypeNames` map that also feeds the decompiler)
  covers all 18 primitives.
- **Constructor-initializer parsing**, **unsafe-modifier detection**, and the
  **`static class` shape predicate** — each had a private RTS copy; each now
  delegates to the product seam (`CSharpFormatter` / `TypeShellProducer`).

## What it enables

- **A trustworthy oracle.** RTS's "the product's C# is valid" claim is no longer
  contaminated by RTS's own spelling divergence — the oracle and the product
  read spelling/shape from the same source of truth.
- **A proven source-fidelity win.** Unifying primitive aliasing improves the
  reconstructed source (`System.Decimal DecimalDefault(System.Decimal…)` →
  `decimal DecimalDefault(decimal…)`), still `Exact` under compile-back.
- **Reusable product API.** `CSharpFormatter.AliasPrimitiveTypeNames` /
  `ParseConstructorInitializer` / `RequiresUnsafeModifier` /
  `TypeRequiresUnsafeModifier`, `TypeShellProducer.IsStaticType`, and
  `PrimitiveTypeNames.TryToKeyword` are general, SRM-only, Roslyn-free seam
  operations any future consumer (not just RTS) can reuse — a further step
  toward RTS being a pure compile-back *planner* rather than a C# re-implementer.

## Scope / non-goals

The remaining `TypeProducer` in RTS is **not** re-implemented C# spelling — it is
closure-aware compile-back planning (dependency-closure roots, stub-body policy)
and its base-type gate depends on RTS's own `Clean` sanitization. That is RTS's
irreducible job and deliberately stays in RTS; see the in-file note near the
base-type gate. This PR is spelling/shape only.

## Validation

- `ILInspector.CSharp.Tests` (incl. new `AliasPrimitiveTypeNames`,
  `ParseConstructorInitializer`, `RequiresUnsafeModifier`,
  `TypeRequiresUnsafeModifier`, `IsStaticType` cases) green.
- RTS oracle tests (`ReturnToSenderPrototypeTests`, `ReturnToSenderEvidenceTests`,
  `ReturnToSenderFixtureCatalogTests`, fast subset) green after updating two
  stale `System.X`→keyword assertions.
- Corpus quality-diff card: **neutral on every raise/fidelity metric** (fidelity
  is opcode-based, so `System.Decimal`↔`decimal` is the same IL — the change is a
  source-spelling improvement, not a fidelity delta). The only flagged rows are
  sampling-cap artifacts from `--compile-cap 25`.

## Adversarial review

Two reviewers from a different model family than the author (Opus), plus an
additional MAI pass given RTS's load-bearing role — reconciliation posted below.

### DIFF
--- a/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
@@ -227,4 +227,97 @@ public sealed class CSharpFormatterTests
     [InlineData("N.@int", "N.@int")]
     public void EscapeTypeKeywords_EscapesIdentifiersButNotTypeSyntax(string input, string expected)
         => Assert.Equal(expected, CSharpFormatter.EscapeTypeKeywords(input));
+
+    [Theory]
+    // Every CLR primitive full name aliases to its C# keyword, including the native
+    // ints (nint/nuint) and decimal, matching the product decompiler's spelling.
+    [InlineData("System.Boolean", "bool")]
+    [InlineData("System.Byte", "byte")]
+    [InlineData("System.SByte", "sbyte")]
+    [InlineData("System.Char", "char")]
+    [InlineData("System.Decimal", "decimal")]
+    [InlineData("System.Double", "double")]
+    [InlineData("System.Single", "float")]
+    [InlineData("System.Int16", "short")]
+    [InlineData("System.UInt16", "ushort")]
+    [InlineData("System.Int32", "int")]
+    [InlineData("System.UInt32", "uint")]
+    [InlineData("System.Int64", "long")]
+    [InlineData("System.UInt64", "ulong")]
+    [InlineData("System.IntPtr", "nint")]
+    [InlineData("System.UIntPtr", "nuint")]
+    [InlineData("System.Object", "object")]
+    [InlineData("System.String", "string")]
+    [InlineData("System.Void", "void")]
+    // Primitives nested in generics, arrays, pointers, and by-ref forms are aliased.
+    [InlineData("System.Collections.Generic.List<System.Int32>", "System.Collections.Generic.List<int>")]
+    [InlineData("System.Collections.Generic.Dictionary<System.String,System.Boolean>", "System.Collections.Generic.Dictionary<string,bool>")]
+    [InlineData("System.Int32[]", "int[]")]
+    [InlineData("System.Int32[,]", "int[,]")]
+    [InlineData("System.Int32&", "int&")]
+    [InlineData("System.Int32*", "int*")]
+    [InlineData("System.Nullable<System.Int32>[]", "System.Nullable<int>[]")]
+    // A longer name that merely contains a primitive as a substring is left alone.
+    [InlineData("System.Int32Enum", "System.Int32Enum")]
+    [InlineData("A.System.Int32", "A.System.Int32")]
+    [InlineData("System.Int32.MaxValue", "System.Int32.MaxValue")]
+    [InlineData("System.Collections.Generic.List<System.Guid>", "System.Collections.Generic.List<System.Guid>")]
+    // Non-System text and already-keyword spellings pass through unchanged.
+    [InlineData("int", "int")]
+    [InlineData("MyNamespace.MyType", "MyNamespace.MyType")]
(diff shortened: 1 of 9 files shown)


richlander:

Three reviewers, each in an isolated worktree at the PR head. Author is Opus 4.8,
so the two required reviewers are from other families (GPT-5.5, Gemini 3.1 Pro);
an MAI-Code-1-Flash pass was added given RTS's load-bearing role. All three are
**CLEAN on the final head `2c0a4113`**.

### Findings and resolutions

**1. GPT-5.5 — signature `unsafe` over-broadening (verified, fixed in `9fe57549`)**
The consolidated `RequiresUnsafeModifier` (`*` or `stackalloc`) was applied to
return/parameter type names, but pre-PR RTS checked signatures for pointer `*`
**only** — `stackalloc` was a body-only check. Since `stackalloc` is an
expression construct that can never appear in a type name, a type name containing
the substring `stackalloc` (e.g. an escaped identifier `@stackalloc`) would
spuriously require `unsafe`; on an `async` member that is CS4004, which would
corrupt the oracle's verdict. **Fix:** added `TypeRequiresUnsafeModifier`
(pointer-only) for type signatures; bodies keep the full `*`-or-`stackalloc`
predicate. Restores exact pre-PR signature semantics. GPT-5.5 re-reviewed → CLEAN.

**2. MAI — escaped-identifier aliasing (not pipeline-reachable; hardened in `2c0a4113`)**
`AliasPrimitiveTypeNames("@System.Int32")` returned `@int`, because `@` is not a
segment char and the following run matched a CLR primitive full name. This is
**not reachable through RTS's `Clean` pipeline** — aliasing runs *before*
`EscapeTypeKeywords`, so no `@` exists at alias time (MAI tested the public API
directly with out-of-contract input). Because `AliasPrimitiveTypeNames` is a
general, reusable public seam API, it was hardened anyway: a run immediately
preceded by `@` is treated as an escaped identifier and left untouched. MAI
re-reviewed → CLEAN.

### Divergence
Gemini initially returned CLEAN and characterized the new unsafe predicate as
applying "the same logic" to signatures — missing the signature/body asymmetry
that GPT-5.5 caught (pre-PR signatures were pointer-only). After that was
surfaced and fixed, Gemini re-reviewed the narrowed logic and the aliaser guard
→ CLEAN both times, with an explicit pre-PR-equivalency trace.

### Final status
| Reviewer | Model family | Result on `2c0a4113` |
| --- | --- | --- |
| GPT-5.5 | GPT | CLEAN |
| Gemini 3.1 Pro | Gemini | CLEAN |
| MAI-Code-1-Flash | MAI | CLEAN |

Resolution commits: `9fe57549` (unsafe signature fix), `2c0a4113` (escaped-identifier guard).

Tests after fixes: `CSharpFormatterTests` green (incl. new `TypeRequiresUnsafeModifier`
and escaped-identifier cases), RTS oracle fast subset green, full `dotnet-inspect.slnx`
build clean, corpus quality-diff card neutral on every raise/fidelity metric.

PR #2764, opened 2026-07-15
De-C#-ify RTS: dedupe generic-constraint keyword spelling into Metadata

richlander:
## Why

RTS (ReturnToSender, the compile-back oracle in `tools/DecompilerHarness`) is
supposed to judge whether the **product** decompiler emits valid, faithful C#.
But it re-implemented the CLI generic-parameter constraint/variance
flag-to-token mapping (`struct`/`class`/`new()`/`in`/`out`) itself — so its
verdict measured the product against **RTS's own** spelling, not the product's.
That is exactly the contamination this de-C#-ification arc removes.

The same mapping was **duplicated a third time**: `MetadataDeclarationQuery`
(simple form) and `ApiSurfaceExtractor` (richer superset with
`class?`/`notnull`/`unmanaged`/`allows ref struct`) each carried their own copy.
RTS being an adversarial consumer flushed out a real product gap: three drifting
implementations of one fact.

## What this enables

One authoritative seam — `ILInspector.Metadata.GenericConstraintKeywords` with
`PrimaryKeyword` / `NewConstraintKeyword` / `VarianceKeyword` /
`AllowsRefStructKeyword`. All three callers now delegate:

- RTS derives constraint tokens from the product seam it judges, so the oracle
  measures the product against the product.
- `MetadataDeclarationQuery` and `ApiSurfaceExtractor` share one source of
  truth; adding a future constraint token (or fixing a spelling) is a one-line
  change in one place instead of three.

These flags are CLI **metadata facts**, so their canonical tokens are
Metadata-owned naming — and Metadata is the only layer every caller reaches
(`CSharp` references `Metadata`, not vice versa). C#-specific aliasing stays in
`CSharp` (already de-C#-ified in #2763).

## Behavior-preserving

Reference- and value-type constraints are mutually exclusive in metadata, so the
two simple callers pass `nullableFlag: 0, isUnmanaged: false` to reproduce their
exact `class`/`struct`/`new()` output byte-for-byte. The `ApiSurfaceExtractor`
richer path keeps its `class?`/`notnull`/`unmanaged`/`allows ref struct`
precedence unchanged (verified by diff and unit tests).

`TypeKindText` is intentionally **left in RTS**: it is a boundary translator over
RTS's internal `CompileBackTypeKind` planning enum, not re-implemented C#
spelling — migrating it would be layering-awkward churn for no correctness gain.

## Evidence

- New unit tests: `GenericConstraintKeywordsTests` (17) — mutual exclusion,
  `class?`/`notnull`/`unmanaged` precedence, variance, `allows ref struct`.
- `ApiSurfaceExtractorTests` 40, RTS oracle Evidence 3 / FixtureCatalog 82 /
  Prototype 114 — all pass.
- Corpus quality-diff card **neutral** on every raise/fidelity metric (0 count
  deltas); the only flagged rows are sampling-cap artifacts (`--compile-cap 25`
  vs baseline 4000), as in prior PRs in this arc.

Adversarial review (two non-Opus reviewers) to follow; PR opened before reviews
finish per policy.

### DIFF
--- a/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
+++ b/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
@@ -501,29 +501,15 @@ public static class ApiSurfaceExtractor
             };
 
             var attrs = param.Attributes;
-            if (includeVariance)
-            {
-                if ((attrs & GenericParameterAttributes.Covariant) != 0)
-                    typeParam.Variance = "out";
-                else if ((attrs & GenericParameterAttributes.Contravariant) != 0)
-                    typeParam.Variance = "in";
-            }
+            if (includeVariance && GenericConstraintKeywords.VarianceKeyword(attrs) is { } variance)
+                typeParam.Variance = variance;
 
             var nullable = GetEffectiveNullable(reader, param.GetCustomAttributes(), nullableContext);
-            var hasReferenceTypeConstraint = (attrs & GenericParameterAttributes.ReferenceTypeConstraint) != 0;
-            var hasValueTypeConstraint = (attrs & GenericParameterAttributes.NotNullableValueTypeConstraint) != 0;
-            var hasDefaultConstructorConstraint = (attrs & GenericParameterAttributes.DefaultConstructorConstraint) != 0;
             var isUnmanaged = AttributeReader.HasAttribute(reader, param.GetCustomAttributes(),
                 KnownAttributeNames.IsUnmanagedAttribute);
 
-            if (hasReferenceTypeConstraint)
-                typeParam.Constraints.Add(nullable == 2 ? "class?" : "class");
-            else if (isUnmanaged)
-                typeParam.Constraints.Add("unmanaged");
-            else if (hasValueTypeConstraint)
-                typeParam.Constraints.Add("struct");
-            else if (nullable == 1)
-                typeParam.Constraints.Add("notnull");
+            if (GenericConstraintKeywords.PrimaryKeyword(attrs, nullable ?? 0, isUnmanaged) is { } primaryKeyword)
+                typeParam.Constraints.Add(primaryKeyword);
 
             foreach (var constraintHandle in param.GetConstraints())
             {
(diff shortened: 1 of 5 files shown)


richlander:

Two reviewers from different model families than the author (Opus 4.8), each in an isolated worktree at head `b5ad8559`.

| Reviewer | Verdict (final head) |
| --- | --- |
| Gemini 3.1 Pro | **CLEAN** |
| GPT-5.5 | **CLEAN** |

### Gemini 3.1 Pro — CLEAN (first pass)
Verified per-caller byte-for-byte equivalence (precedence `reference → unmanaged → struct → notnull`; `new()` suppression `DefaultConstructor && !NotNullableValueType`; simple callers passing `nullableFlag:0, isUnmanaged:false` can never emit `class?`/`notnull`/`unmanaged`/`allows ref struct`; emitted list order unchanged), the `byte? → nullable ?? 0` collapse, the `isStruct`-gated `System.ValueType`/`ValueType` skip, and layering (helper is SRM-only/Roslyn-free in `ILInspector.Metadata`). Ran Metadata.Tests (562), `GenericConstraintKeywordsTests` (17), `ApiSurfaceExtractorTests` (40) — all pass.

### GPT-5.5 — one finding, reconciled to CLEAN
**Finding (initial):** the two *simple* callers (RTS, `MetadataDeclarationQuery`) checked value-type-before-reference-type at base; the shared `PrimaryKeyword` checks reference-type first, so raw metadata with **both** `ReferenceTypeConstraint` **and** `NotNullableValueTypeConstraint` set would flip `struct` → `class`.

**Reconciliation (agreed CLEAN):** I enumerated all 256 attribute-byte values — the divergence occurs for exactly 64 values, and **all 64 have both mutually-exclusive constraint bits set**; there are **zero** divergences for any input with at most one of the two flags. Those flags are mutually exclusive per ECMA-335 (`SpecialConstraintMask`), so no conforming compiler emits both, and RTS only ever processes compiler-produced fixtures. Furthermore the three pre-PR implementations already *disagreed* on this malformed input — `ApiSurfaceExtractor` was already reference-first (GPT confirmed `apiDiffs=0`) — so no single unifying order can be byte-for-byte identical to all three on malformed metadata; the PR unifies them to reference-first, matching the richest existing producer. GPT-5.5 re-ran on the clean head (`simpleDiffsWithoutBoth=0; apiDiffs=0`) and reclassified to **CLEAN**, agreeing it is a cosmetic difference on unreachable malformed-only input for which neither spelling is semantically correct.

**Resolution commit:** none — no code change. The finding is confined to non-ECMA malformed metadata; the suggested fix (a separate simple helper / inline `isStruct` at both sites) would reintroduce the exact multi-implementation duplication this PR removes, changing output only on that unreachable input. The helper already documents the mutual-exclusivity invariant.

### Where they agreed / diverged
Both agreed valid-metadata output is byte-for-byte identical and the layering is sound. The only divergence was GPT's malformed-input observation, resolved above. Both are CLEAN on the final head with no outstanding findings.

PR #2767, opened 2026-07-16
Add experimental contextual Repl exploration pilot

autocarl:
## Context: unsolicited exploratory proposal

This is an **unsolicited exploratory contribution** proposing that dotnet-inspect add a contextual Repl pilot. It was not requested, sponsored, or pre-approved by the dotnet-inspect maintainers, and the proposing contributor is also involved in Repl's development.

The goal is to provide a concrete prototype for evaluation—not to imply that the project has selected Repl, committed to this direction, or placed it on its roadmap. There is no expectation that this proposal be accepted; declining or closing it is a completely valid outcome if the dependency, maintenance cost, or interaction model does not fit the project.

## What does this PR do?

Adds an opt-in experimental command without replacing the existing `System.CommandLine` CLI:

```bash
dotnet-inspect repl
```

The pilot:

- pins `Repl.Core` to prerelease `0.12.0-dev.3`
- adds contextual `package → type → member → source` navigation
- delegates every inspection through the existing `System.CommandLine` graph so current parsing, diagnostics, sections, overload indexes, and rendering remain authoritative
- changes selection state only after the delegated command returns exit code `0`
- preserves literal selectors and option-shaped suffixes across `--`
- supports standalone and `=true|false` forms for early global boolean options before `--`
- streams delegated stdout/stderr through bounded writers while restoring the process console after success, error, or cancellation
- caps each delegated stdout/stderr stream at 10,000 lines or 1,000,000 characters and reports truncation explicitly
- reserves `repl` as a root verb while keeping a package named `repl` available through `dotnet-inspect package repl`
- documents the workflow, compatibility boundary, operational limits, and third-party MIT notice

Example session:

```text
> package System.Text.Json
[selected-package]> type System.Text.Json.JsonSerializer
[selected-package/selected-type]> member Serialize
[selected-package/selected-type/selected-member]> source 1
```

## Packaging and compatibility boundary

- Untrimmed managed builds and the managed `any` tool package include the pilot.
- Trimmed and RID-specific NativeAOT builds exclude `Repl.Core` and expose an explicit unavailable stub returning exit code `1`.
- An unsafe explicit Repl + trimming/AOT override fails the build rather than silently producing an unsupported configuration.
- Existing non-Repl commands and output contracts remain the primary automation surface; the interactive prompts do not define a new JSON or scripting contract.

## Current limitations

- This is a focused package/type/member exploration pilot, not a second grammar for every dotnet-inspect command.
- Repl `0.12.0-dev.3` currently rebuilds absolute navigation from space-split route strings, so the bridge stores user selectors in session state and navigates through static selected scopes.
- Ctrl+C cancellation is best effort because not every existing NuGet/decompilation handler propagates cancellation all the way down.
- Very large delegated output is intentionally truncated; users needing complete output should invoke the equivalent regular CLI command.
- The current Repl prerelease is not trim-clean, which is why the pilot is excluded from trimmed/NativeAOT artifacts.

## Verification

Validated locally with .NET SDK `11.0.100-preview.6.26359.118`:

- `dotnet-inspect.Tests`: **1,933 total, 0 failed, 10 skipped**
- targeted entry-point/parser/router regressions: **7 passed**
- Release build of the product/test graph: **0 warnings, 0 errors**
- real managed `any` package created and installed as a local .NET tool
- installed-package smoke test completed `package → type → member → source`
- package inspection confirmed `Repl.Core.dll` is present in the managed `any` package
- integrated RID/AOT build completed successfully; its `repl` stub returned `1` and `Repl.Core.dll` was absent
- whitespace/static secret and shell-execution scans passed

Managed package exercised locally:

```text
PackageId: dotnet-inspect.any
Version:   0.16.0
SHA-256:   44ac3cdcdbda4b09d42cae9e85cbe8364d06c54a1c0d2cc20233a5b3284d526f
```

## Review posture

The intended review question is whether this interaction model and dependency are useful enough to pursue—not whether the project is already committed to adopting them. Feedback that the prototype should be narrowed, redesigned, moved upstream, or closed is welcome.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -9,6 +9,7 @@
     <PackageVersion Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="3.1.21" />
     <PackageVersion Include="NuGet.Versioning" Version="7.3.0" />
     <PackageVersion Include="NuGetFetch" Version="0.6.2" />
+    <PackageVersion Include="Repl.Core" Version="0.12.0-dev.3" />
     <PackageVersion Include="ShellComplete" Version="0.1.0" />
     <PackageVersion Include="SourceLinkFetch" Version="0.1.0" />
     <PackageVersion Include="System.CommandLine" Version="3.0.0-preview.5.26302.115" />
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ context for copied DLLs. A future `--deps` source can represent runtime
 | Source mapping | `type`/`library`/`package -S "Source Files"`, `member -S "Source Locations"` / `"Original Source"` | SourceLink URLs, member file/line locations, source fetching, URL verification, token+IL-offset to source-line resolution. |
 | Performance analysis *(experimental)* | `library`/`type`/`member -S "Top Leverage"`, `"Performance Triage"`, `"Call Graph"`, `"Caller Graph"` | Whole-assembly call-graph leverage ranking — direct callers, root reach, fanout, depth, loop calls — with opt-in per-node cost signals (alloc, copy, unsafe, reflection, throw/exception, catch/finally) and actionable rewrite-shape detection. |
 | Decompiler *(experimental)* | `member -S @Source` (`Decompiled Source`, `Annotated Source`, `Original Source`, `Source Diff`, `IL`); `member -S "Fidelity Causes"` | Raises method bodies to C#, interleaves IL and hidden-fact annotations, diffs SourceLink-backed source against decompiled source, and exposes typed `DEC####` fidelity causes rather than emitting plausible-but-wrong source. |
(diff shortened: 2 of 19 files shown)


richlander:
> It was not requested, sponsored, or pre-approved by the dotnet-inspect maintainers

That statement provides good clarity on what a better path would have been.

Repls are cool. I don't want to take a new dependency and in particular one that is only usable via the fallback `any` package.  This fit doesn't seem good. I've had request for other features, but not this one.

autocarl:
Thank you for taking the time to look at this and explain your thinking, Rich. Your feedback is clear and makes sense.

One additional bit of context: Repl 0.12 is expected to support trimming and Native AOT, which should address that specific limitation. That said, we understand that this does not change the broader fit question.

We'll take a look through the backlog and see whether there are other items where we can contribute in a way that better matches the project's current needs.

PR #2755, opened 2026-07-15
Measure recursive traversal repetition

richlander:
## Summary

Adds a measurement-only recursive-traversal census and records a selective
static receipt for Aspire's `OtlpSpan.GetChildSpans`. No product rows, ranking,
or local allocation semantics change.

- recognizes only an exact resolved `call` to the same MethodDef inside a loop
- reuses typed invocation traversal for deterministic downstream witnesses
- rejects ordinary recursion, mutual recursion, `callvirt` self-dispatch, and
  function loads
- preserves candidate identity, Finding provenance, local `Loop`,
  multiplicity, confidence, and weight
- adds compiled positive and near-miss fixtures

## Measured result

| Corpus | Roots | Opportunities | Reachable |
| --- | ---: | ---: | ---: |
| Pinned six-library corpus | 29 | 2,459 | 75 |
| Aspire Dashboard 9.0.0 | 1 | 488 | 1 |

The six-library reachable population is 11 traversal-root rows, nine direct,
20 transitive within depth four, and 35 beyond the bound.

Aspire is exact and narrow:

```text
AddSelfAndChildren @ IL_006A -- call in-loop --> AddSelfAndChildren
AddSelfAndChildren @ IL_001A -- callvirt --> OtlpSpan.GetChildSpans
```

`GetChildSpans` remains candidate `pt~6756cfa4ed4130d8`, shape
`instance-method-group-delegate`, local multiplicity `once`, and
`LocalInLoop=false`.

This proves branching traversal potential, not runtime heat, recursion depth,
collection size, or guaranteed execution. The result is an explicit product
non-action pending runtime evidence or a stronger use case.

## Orthogonal registry controls

The recent registry improvements retain their allocation-fanout distinction:

| Method | Once paths |
| --- | ---: |
| `LibrarySections.CreatePipeline` | 53 |
| `LibrarySourcePlans..cctor` | 1 |
| `SpikeSections..cctor` | 34 |
| `PlanFor` / `CompilePlan` / `CapabilityPlan.ExecuteAsync` | 0 |

Those methods have no recursive-traversal evidence, confirming that this census
does not act as a generic known-performance-win classifier.

## Validation

- Release solution build
- `ILInspector.Analysis.Tests`: 471 passed
- pinned six-library, Aspire Dashboard, and registry control measurements
- markdownlint

Closes #2754. Related to #2724, #2733, #2749.

PR #2748, opened 2026-07-14
Measure deferred callback evidence

richlander:
## Summary

Records an explicit non-action for deferred callback evidence: callback
registration is machine-readable diagnostic provenance, but it is not proof of
invocation frequency or runtime heat.

- adds a measurement-only AnalysisHarness census for exact
  `ldftn -> delegate .ctor -> consumer` shapes
- separates proven immediate invocation, trusted Blazor registration, unknown
  consumers, cached/unconsumed construction, and required near misses
- reuses typed invocation resolution downstream without relabeling function
  loads as calls or changing Performance Triage behavior

## Evidence

| Corpus | Function loads | Opportunities | Reachable | Proven consumption |
| --- | ---: | ---: | ---: | ---: |
| Pinned six-library corpus | 2,176 | 2,459 | 10 | 0 |
| Aspire Dashboard 9.0.0 | 1,019 | 488 | 8 | 0 |

Aspire contains 12 trusted `RenderTreeBuilder.AddAttribute` registration sites.
`ColorGenerator.GetColorIndex` is reached at downstream depth 2 through the
expected `RenderFragment` construction and registration witness, with
`ConsumptionProven=false`.

No candidate identity, local `Loop`, multiplicity, confidence, weight, ranking,
or default product output changes.

## Validation

- Release solution build
- 466 Analysis tests
- pinned six-library and Aspire Dashboard census runs
- markdownlint

Closes #2738

PR #2740, opened 2026-07-14
Add direct caller-loop evidence to Performance Triage

richlander:
## Summary

Adds direct, invocation-only caller-loop evidence to Performance Triage without
relabeling local allocation semantics or static heat.

- introduces typed `CallerLoopEvidence` with deterministic shortest witness
  paths over resolved `call`, `callvirt`, and `newobj` edges
- caps the product path at one hop while the AnalysisHarness census reuses the
  same primitive for transitive measurement
- exposes filterable/sortable `CallerLoop`, `CallerLoopDepth`, and
  `CallerLoopWitness` fields across library, type, and member views
- preserves row population, order, `Candidate`, Finding provenance, local
  `Loop`, multiplicity, confidence, and weight
- adds compiled `call`, `callvirt`, and `newobj` positives plus non-loop,
  conditional, recursion, and `ldftn` near misses

## Demo

```bash
dotnet-inspect library ILInspector.Analysis.Fixtures.dll \
  -S "Performance Triage" \
  --where "Member=*BoxDirect*" \
  --columns "Member,Candidate,Loop,CallerLoop,CallerLoopDepth,CallerLoopWitness" \
  --jsonl
```

```json
{"member":"ILInspector.Analysis.CallerLoopFixtures.CallerLoopFixture.BoxDirect(int)","candidate":"pt~d2102b3ef684609a","loop":"","caller_loop":"direct","caller_loop_depth":"1","caller_loop_witness":"ILInspector.Analysis.CallerLoopFixtures.CallerLoopFixture.InvokeDirectInLoop(int) @ IL_0008 -> ILInspector.Analysis.CallerLoopFixtures.CallerLoopFixture.BoxDirect(int)"}
```

The empty local `loop` field is intentional: the allocation itself remains
once-per-call, while the separate caller witness shows that an upstream
invocation can repeat the method.

## Corpus and compatibility evidence

| Corpus | Rows | Direct caller-loop rows | Distinct direct methods |
| --- | ---: | ---: | ---: |
| Pinned six-library corpus | 1,761 | 122 | 83 |
| Aspire Dashboard 9.0.0 | 198 | 20 | 10 |

Baseline and head JSONL are byte-equivalent after removing only the three
additive caller-loop fields. Candidate IDs, local fields, row population, and
ordering are unchanged.

`ColorGenerator.GetColorIndex` retains candidate
`pt~59fd64a337e32bfc`, local `Loop` empty, and no caller-loop witness because
[...]

PR #2769, opened 2026-07-16
Document caller-loop triage workflow

### LINKED ISSUE #2724: Performance Triage: model cross-method repetition without relabeling static heat

richlander:
## Summary

Closes the caller-loop roadmap with a query-only UX decision.

- adds the landed `CallerLoop=direct` workflow to the performance skill
- explains `CallerLoopDepth` and `CallerLoopWitness` as deterministic static
  receipts for selecting profiling targets
- preserves local `Loop`, multiplicity, confidence, weight, candidate identity,
  and default ranking
- keeps function loads, callback registration, and recursive traversal outside
  caller-loop evidence without stronger product-owned or runtime proof

## Roadmap decision

The completed census showed genuine invocation evidence is selective enough to
query, and #2740 landed direct one-hop product evidence. The product will not
add transitive propagation, a ranking tie-break, or a dedicated preset from
static evidence alone.

The two adjacent experiments remain explicit non-actions:

- #2748 found callback registration but no proven callback consumption.
- #2755 found recursive traversal potential but no realized depth, frequency,
  or runtime heat.

The skill now teaches the supported product workflow without weakening those
boundaries.

## Validation

- `markdownlint skills/performance/SKILL.md`
- `git diff --check`

Docs-only and low risk; no product behavior or measured behavior changes.

Closes #2724.

### DIFF
--- a/skills/performance/SKILL.md
+++ b/skills/performance/SKILL.md
@@ -96,6 +96,28 @@ a `pt~` candidate id but no exact source Finding, operation, or token.
 `Provenance=unmatched` flags an instruction-level row that did not join to the
 expected producer census.
 
+## Select direct caller-loop repetition
+
+A once-per-call allocation can still be repeated by an upstream caller's loop.
+Select rows with an exact direct invocation receipt:
+
+```bash
+dnx dotnet-inspect -y -- library MyLib.dll -S "Performance Triage" \
+  --where "CallerLoop=direct" --jsonl
+```
+
+`CallerLoopDepth` and `CallerLoopWitness` identify the deterministic invocation
+site. This evidence does not change the row's local `Loop`, multiplicity,
+confidence, weight, candidate identity, or default rank. Use it to select a
+candidate for profiling, not as proof that the caller is hot or that the loop
+executes.
+
+Only resolved invocation edges qualify. Function loads and callback
+registration do not prove callback execution, and recursive traversal does not
+prove realized depth or frequency. Do not infer either case into caller-loop
+evidence; require runtime evidence or a stronger product-owned invocation
+contract.
+
 Not every shape is a pure hot-path win. `async-state-machine` is reported as
 amortized (low confidence) unless the allocation sits in a loop: async lowering
 moves work into a state object rather than eliminating it, often once per


richlander:
Ready to merge.

Markdown CI passed. This is a docs-only, low-risk skill update; no adversarial review is required.

PR #2771, opened 2026-07-16
Remove dead RTS self-type parameter normalization

richlander:
## Summary

Removes `NormalizeSelfTypeParameters` (and its `DirectSelfTypeSignature` helper) from the ReturnToSender type planner. This continues the "de-C#-ify RTS" arc by deleting harness-side metadata reconstruction in favour of the product-owned parameter spelling — **and fixes a latent parameter-corruption bug** surfaced during adversarial review.

> **Correction:** the first commit describes this as "dead code / behaviorally identical". Adversarial review (below) proved that is inaccurate — the transform **is reachable** and the deletion is a **bug fix**. This description is the authoritative narrative.

## What the normalization did — and why it was wrong

`NormalizeSelfTypeParameters` rewrote any method parameter whose display name equalled a locally-computed "direct" self-type (`DirectSelfTypeSignature`) to the full self-type (`MetadataDeclarationQuery.SelfTypeSignature`).

- **For compiler-emitted simple nested self-types it was a no-op.** For `Outer<T>.Inner<U>`, the product already spells a `self` parameter as `NS.Outer<T>.Inner<U>` (identical to `SelfTypeSignature`), while `DirectSelfTypeSignature` produces the malformed `NS.Outer<U>.Inner<>`, which never matches. Confirmed by probe + 199 RTS oracle tests + a framework corpus run (CoreLib/Collections/Linq), all with **zero fires**.
- **But it actively corrupted legitimate shifted-sibling nested-generic parameters.** For `Outer<T>.Inner<A,B,C>.ReadShifted(Outer<A>.Inner<B,C> shifted)`, `DirectSelfTypeSignature` produces the *well-formed* `Outer<A>.Inner<B,C>` — which matches the real parameter. The old code then rewrote it to `Outer<T>.Inner<A,B,C>` (a different, wrong type), corrupting the reconstructed shell (`shifted.Marker` → CS1061). Deleting the normalization preserves the correct spelling.

## Regression coverage

Added generated fixture `rts.shifted-sibling-nested-generic` (`GeneratedFixtures.cs`) + a catalog test that reconstructs `ReadShifted` and asserts, via an expected source fragment, that the parameter is preserved as `ShiftedSiblingOuter<A>.Inner<B, C> shifted`.

**Guard verified both ways:** with the old normalization reintroduced the fixture fails `source-fragment-missing`; with it removed it passes `Exact`.

## Validation

- `GeneratedFixtureCatalogTests`, `ReturnToSenderEvidenceTests`, `ReturnToSenderFixtureCatalogTests`, `ReturnToSenderPrototypeTests`: **216 pass, 0 fail**.
- Framework RTS corpus (CoreLib/Collections/Linq): verdicts unchanged (48 Exact / 1 OpcodeDiff / 2 RecompileFail / 1 ContextFail).
- Built on .NET 11 preview 6 (`11.0.100-preview.6.26359.118`).

## Risk

Harness-only (`tools/DecompilerHarness`). The behaviour change is strictly a fidelity improvement (removes a corruption); no regression in the 216-test suite or the framework corpus. Two adversarial reviews reconciled below.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/GeneratedFixtureCatalogTests.cs
+++ b/src/ILInspector.Decompiler.Tests/GeneratedFixtureCatalogTests.cs
@@ -906,7 +906,7 @@ public class GeneratedFixtureCatalogTests
             GeneratedFixtureCatalog.Select("minimal.property.literal").Select(fixture => fixture.Id).ToArray());
 
         Assert.Equal(
-            ["rts.attribute-shell"],
+            ["rts.attribute-shell", "rts.shifted-sibling-nested-generic"],
             GeneratedFixtureCatalog.Select("rts").Select(fixture => fixture.Id).Order(StringComparer.Ordinal).ToArray());
 
         Assert.Equal(
@@ -1440,6 +1440,21 @@ public class GeneratedFixtureCatalogTests
             result.FixtureId == "rts.attribute-shell" &&
             result.Method == "get_FeatureType");
         Assert.DoesNotContain("CS0641", report);
+    }
+
+    [Fact]
+    public void ReturnToSenderRtsCatalog_KeepsShiftedSiblingNestedGenericParameter()
+    {
+        var run = GeneratedFixtureRunner.RunReturnToSenderCatalog(
+            GeneratedFixtureCatalog.Select("rts.shifted-sibling-nested-generic"));
+        string report = GeneratedFixtureRunner.FormatReturnToSenderCatalogReport(run, maxExamples: 10);
+
+        Assert.True(run.Passed, report);
+        var result = Assert.Single(run.Results);
+        Assert.Equal(GeneratedFixtureReturnToSenderStatus.Pass, result.Status);
+        Assert.Equal(FidelityCheck.CompileBackStatus.Exact, result.ActualStatus);
+        Assert.Equal("ShiftedSiblingOuter`1.Inner`3", result.Type);
+        Assert.Equal("ReadShifted", result.Method);
     }
 
     [Fact]
--- a/tools/DecompilerHarness/GeneratedFixtures.cs
+++ b/tools/DecompilerHarness/GeneratedFixtures.cs
@@ -1481,6 +1481,38 @@ internal static class GeneratedFixtureCatalog
         ],
         ["rts", "attribute", "shell", "return-to-sender"]);
 
+    public static readonly GeneratedFixtureDefinition RtsShiftedSiblingNestedGeneric = new(
+        "rts.shifted-sibling-nested-generic",
+        """
+        public class ShiftedSiblingOuter<T>
+        {
+            public class Inner<A, B>
+            {
+                public int Marker => 123;
+            }
+
+            public class Inner<A, B, C>
+            {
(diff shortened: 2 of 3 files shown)


richlander:

Two independent reviewers (non-Opus, per policy), each in an isolated worktree at the PR head:

- **GPT-5.5** and **Gemini 3.1 Pro**

### Where they agreed (verified, actioned)

Both **independently falsified the original "dead code / behaviorally identical" claim**. The transform is reachable; deleting it is a **bug fix**, not a no-op.

- **GPT-5.5** (realistic, plain-C#): `Outer<T>.Inner<A,B,C>.ReadShifted(Outer<A>.Inner<B,C> shifted)`. `DirectSelfTypeSignature` → `Outer<A>.Inner<B,C>`, matching the real parameter; old code rewrote it to `Outer<T>.Inner<A,B,C>` → `RecompileFail` (CS1061 on `.Marker`). The PR fixes it to `Exact`.
- **Gemini 3.1 Pro** (exotic IL string-collision): a custom type literally named `Outer<U>.Inner<>` (valid IL via `PersistedAssemblyBuilder`, not C#-emittable) collides with the malformed `DirectSelfTypeSignature` output, triggering the same erroneous rewrite.

**Verification:** I independently reproduced GPT's case with a probe (`SelfTypeSignature=Outer<T>.Inner<A,B,C>`, `DirectSelfTypeSignature=Outer<A>.Inner<B,C>`, real param `=Outer<A>.Inner<B,C>` → match). `git status` clean before/after in the review trees.

### Actions taken

1. **Corrected the narrative** — the PR description now states this is a bug fix that removes corruption of legitimate shifted-sibling nested-generic parameters, not dead-code removal. (First commit message wording is stale; not amended per repo policy — the PR body is authoritative.)
2. **Added regression coverage** (commit `bc439d4b`) — fixture `rts.shifted-sibling-nested-generic` asserting the parameter is preserved. Guard verified: fails with the old normalization reintroduced, passes with it removed.

### Divergence

None on substance — both reviewers reached the same conclusion via different fixtures. Neither found the fix itself incorrect; both explicitly stated the deletion is correct and improves RTS fidelity. No security/compat issues raised.

Both reviewers' core finding is resolved. Requesting a re-review of the fixed head before this is marked ready.

richlander:
## Two CLEAN re-reviews on the fixed head (bc439d4b)

Both reviewers re-reviewed the fixed head in isolated worktrees and independently reproduced the regression guard.

- **GPT-5.5 — CLEAN.** Built + ran `GeneratedFixtureCatalogTests` (17 passed). Reintroduced the deleted normalization in a *separate* copy (`/tmp/rev-selftype-regress-copy`) and confirmed `ReturnToSenderRtsCatalog_KeepsShiftedSiblingNestedGenericParameter` fails with `source-fragment-missing` — the guard works. Worktree clean before/after.
- **Gemini 3.1 Pro — CLEAN.** Built + ran the new test (Exact). Reverse-patched the deletion, confirmed the test fails (`missing expected source fragment: ShiftedSiblingOuter<A>.Inner<B, C> shifted`), then restored a pristine worktree.

No outstanding findings from either reviewer. The bug fix is locked in by a verified regression test.

**Ready to merge** — pending @richlander approval (I won't merge without it).

PR #2783, opened 2026-07-16
Make generated-fixture tests iterate advertised inventory

richlander:
## Summary

Fixtures are test assets, not product. Reserve adversarial rigor for the
boundary between a test asset and the product under test — not between two
co-located test assets. Re-encoding a fixture's expected outcome as a
hand-maintained test literal to act as an "independent oracle" adds a place to
drift without any adversarial safety, because both sides are edited and reviewed
together.

This codifies that principle and applies it to `GeneratedFixtureCatalogTests`,
which had hand-duplicated the catalog's inventory and per-target expectations.

## Changes

- **`docs/fixture-governance.md`** — new **Expectation ownership** section: a
  fixture owns its input and expected outcome together; tests iterate the
  advertised inventory and compare actual vs the fixture's declared expectation;
  do not re-encode expectations as a second literal; do not write tests whose
  failure mode is "you added a valid fixture, now update my list."
- **`MinimalCompileBackRungs_...`** — replaced ~120 per-target `AssertTarget`
  literals with a `foreach` over `run.Results`. `run.Passed` already asserts
  every fixture-declared target matched its declared compile-back status and
  shape; the loop adds the `DecompilerFidelity == "Full"` check the aggregate
  omits. Report-id and default-run composition checks now derive from the run.
- **`CatalogueSelection_...`** — tests the `Select` prefix/exact contract against
  the catalog itself, so a new fixture under an existing prefix no longer forces
  a literal-list edit.
- **`CatalogueListJson_...`** — asserts the emitted id set equals the catalog id
  set; keeps the representative JSON schema probes (they pin the serializer's
  field mapping, a real contract).
- Removed the now-unused `AssertShape` helper.

Net: test file **-987/+84** lines.

## Evidence

- `dotnet build src/ILInspector.Decompiler.Tests -c Release --configfile nuget.config` -> Build succeeded, 0 warnings.
- `dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class "ILInspector.Decompiler.Tests.GeneratedFixtureCatalogTests"` -> Total: 17, Failed: 0.
- `markdownlint` clean on `docs/fixture-governance.md`.

## Risk

Test-and-docs only; no product behavior change. The subtle risk is test
weakening, so each dropped literal was verified subsumed: `run.Passed` covers
per-target compile-back status and shape; the fidelity loop covers `Full`; the
JSON schema probes still pin field mapping; the `Select` contract is checked
against the live catalog. Requesting adversarial review to confirm no coverage
was lost.


richlander:
## Adversarial review — reconciled (two CLEAN fixed-head reviews)

Reviewers: **GPT-5.5** and **Gemini 3.1 Pro** (I authored the PR as Claude Opus 4.8, so both reviewers are from other families). Each ran in its own isolated worktree.

### Round 1 (head `66d00066`)

**Convergent finding — both reviewers, high confidence.** The refactored `MinimalCompileBackRungs_...` iterated `run.Results` and relied on `run.Passed`, but never asserted the run *covered* every advertised target. `Assert.NotEmpty` only catches the fully-empty case, so a runner that silently dropped a target/fixture would pass vacuously — a real regression vs. the old per-target `Assert.Single`. Both reviewers demonstrated it by mutating `GeneratedFixtureRunner.Run` to skip a fixture and observing the suite stay green.

**GPT-5.5 only — secondary.** `CatalogueSelection` tested the `Select` exact-or-prefix contract against live inventory, where a `Contains` mutant produces identical sets, so the contract was under-pinned. This also contradicted the new doc guidance ("test contract logic against synthetic inputs, not the live inventory") in this same PR. Gemini judged this *not a regression* (the old literal-list test admitted the same `Contains` mutant) — a correct observation — but the doc-vs-code inconsistency was still worth resolving.

### Resolution (`64438f54`)

- **Inventory coverage:** `MinimalCompileBackRungs_...` now asserts the advertised `(FixtureId, Type, Method, Overload)` key set equals `run.Results`' key set (full-key set equality, not just `Count`, so a drop-plus-spurious-add can't net out). Verified: skipping a fixture in `Run` now fails at that assertion.
- **Selector contract:** extracted `internal static GeneratedFixtureCatalog.MatchesSelector(id, selector)` (used by `Select` after its unchanged null/whitespace→`All` short-circuit); added `SelectorPredicate_MatchesExactIdOrPrefixOnly` testing exact / prefix / substring-not-prefix / longer / case-sensitive against synthetic ids (kills the `Contains` mutant), and reduced `CatalogueSelection_AppliesSelectorPredicateOverCatalog` to an integration check. The governance doc's advice is now self-consistent with the code.

### Round 2 (fixed head `64438f54`) — both CLEAN

- **GPT-5.5: CLEAN.** Verified the inventory assertion compares the full key multiset (independently confirmed 119 distinct keys, 0 duplicates, so no collision can mask a drop), `Select` still short-circuits before `MatchesSelector`, and the synthetic predicate test pins exact-or-prefix.
- **Gemini 3.1 Pro: CLEAN.** Re-introduced its drop mutation (`minimal.array-index`/`Method1`) → suite now halts at the tuple set-equality check. Confirmed `MatchesSelector` is behaviorally identical to the prior inline predicate and no `CatalogueSelection` coverage was lost.

### Validation
- `dotnet build src/ILInspector.Decompiler.Tests -c Release --configfile nuget.config` → 0 warnings, 0 errors.
- `dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class "…GeneratedFixtureCatalogTests"` → Total: 18, Failed: 0.
- `markdownlint` clean on `docs/fixture-governance.md`.

richlander:
Follow-up (99507065, after the two CLEAN reviews of 64438f54): removed the four `FormatReport` report-string smoke assertions (`decompiler=Full`, `compile-back=Exact`, `shape=ForStatement`, `shape=ElementAccessExpression`).

Low-risk, mechanical deletion — not sent for another review round:
- `decompiler=Full` / `compile-back=Exact` were tautological: the loop above already asserts every result is `Full` and `Exact`, so the report must contain them.
- The `shape=*` tokens pinned a human-readable debug string with no parser (the machine format is `FormatListJson`, tested separately). `GENERATED FIXTURE LADDER` still proves the report renders.

Consistent with the PR's own governance principle: don't be adversarial toward test-asset plumbing. Final head is now 99507065.

Issue #2781, opened 2026-07-16
Classify generated support state in the product seam


richlander:
## Problem

The seam classifies only signature strings today
(`TypeShellProducer.IsUnsupportedSurfaceSignature`,
`src/ILInspector.CSharp/TypeShellProducer.cs:~78-82`), so several generated-support-
state cases are handled ad hoc in the harness. Split the classification by concern and
give each a typed decline reason.

## Steps

1. **Unsupported type/member surface** (the signature-string case): a typed decline.
2. **Generated body-reference state** — the `___o__N` dynamic call-site fields and
   caches (`tools/DecompilerHarness/ReturnToSenderTypePlanner.cs:~3049-3134`). This is
   body/Decompiler territory, so it takes the same layered split as #2780: the
   Decompiler emits neutral body-reference evidence, the seam classifies/declines on
   it, and RTS forwards. No raw IR leaks into the seam.
3. **Display-class name spellability** (`<>c__DisplayClass...`): a name-level classify.

## Acceptance

- Each case has a typed decline reason.
- The measured CS0103/CS0246/CS1001 generated-state rows are accounted for under the
  parity gate (#2776).
- No new `ILInspector.CSharp → ILInspector.Decompiler`/Roslyn dependency.

richlander:
Closing as stale. The generated-support-state footprint is negligible and the acceptance points at the wrong instrument.

Measured over the full pinned corpus (RTS shell compile-back across 18,574 property getters):

- Only 29 of 18,574 failure details reference a generated name (`o__`, `<>c`, `DisplayClass`) — 0.16% of the population.
- The CS0103/CS0246/CS1001 codes this issue cites are overwhelmingly general closure/namespace gaps (CS0234=3310, CS0138=3219, CS0122=1333 dominate), not generated-support state.
- The step-2 "generated body-reference state" cases are already declined ad hoc; formalizing typed decline reasons for a 0.16% surface is not worth the cross-layer DTO plumbing.

The parity-gate acceptance is also mis-specified: the current burn-down is 7 rows and none are generated-state. If a generated-state gap shows up in the burn-down later, reopen with the concrete row.

Issue #2777, opened 2026-07-16
Move member and type source composition from ReturnToSender into the product seam


richlander:
## Problem

ReturnToSender (RTS) still composes C# source itself: it builds print requests,
assembles the compilation unit and usings, and escapes C# identifiers via the
harness-local `CompileBackCSharpNames` helper (11 uses in
`tools/DecompilerHarness/ReturnToSenderTypePlanner.cs`). While that logic lives in
the harness, RTS can never become a thin orchestrator and `CompileBackCSharpNames`
cannot be deleted.

The product already renders type/member source through
`CSharpTypePrinter.PrintBatch` (`ReturnToSenderTypePlanner.cs:~1272`), so the residue
to move is narrower than it looks: the harness-local `TypeProducer` (nested class
~`:2579`), the `CSharpTypePrintRequest` construction, the compilation-unit/usings
assembly, the identifier escaping, and some body-fact extraction.

## Layering constraint

`ILInspector.CSharp` references only `ILInspector.Metadata` (Roslyn-free, SRM-only).
`ILInspector.Decompiler` references `ILInspector.CSharp`, never the reverse. The seam
therefore cannot read Decompiler IR (`IrFunction`). Any body fact it needs must be
handed to it as a neutral Metadata-level DTO. The composer today reads `IrFunction`
directly (`ConstructorChainCall(IrFunction)` ~`:1536`, `PrimaryConstructorFromPrologue`
~`:1968`), so this must be done as an ordered, layered chain — not a lift-and-shift.

## Steps

1. **Decompiler emits neutral DTOs.** From `IrFunction`, produce Metadata-level
   evidence for the shell-relevant body facts: the constructor-chain target/args, and
   **both** primary-constructor reconstruction paths — the prologue path
   (`PrimaryConstructorFromPrologue`) and the captured-field path
   (`PrimaryConstructorFromCapturedFields` ~`:2065`, which today derives parameters
   from field names plus rendered body text ~`:2064-2093`). No IR leaks past this
   boundary.
2. **Seam owns composition and the complete identifier producer.** Retire the
   harness-local `TypeProducer`, the print-request construction, and the
   compilation-unit/usings assembly; consume the DTOs from step 1. The seam today only
   `@`-escapes keywords (`CSharpDeclarationWriter.EscapeIdentifier` ~`:1133`) and has
   no unspellable-name sanitization, so move `SanitizeUnspellableIdentifier` (pure
[...]

PR #2788, opened 2026-07-16
Add an RTS parity regression gate with a committed burn-down manifest

### LINKED ISSUE #2776: Add an RTS parity gate that fails on any Exact-to-recompile-failure regression
## Problem

The ReturnToSender (RTS) compile-back path is being thinned so it stops owning C#
knowledge and becomes a plain orchestrator over the product seam and Roslyn. Every
step of that migration risks silently regressing fidelity — a method that recompiles
`Exact` today could start failing to recompile. We need a mechanical gate that makes
any such regression a hard failure, so the migration stays falsifiable.

This is the measurement backbone the rest of the reversal work depends on; it should
land first and everything else cites it.

## Scope

- Commit an RTS-parity baseline over the standard corpus (14 assemblies, ~89k
  methods), recording per-method `FidelityCheck`/`FidelityReference`
  (`tools/DecompilerHarness/CorpusSensor.cs` records these at ~`:572`; compare gates
  ~`:1028-1064`; delta emission ~`:1185-1206`).
- Add a test/CI assertion that **fails if any row has `FidelityReference == Exact`
  and `FidelityCheck ∈ {RecompileFail, ContextFail}`** — i.e. something that
  recompiled exactly before now fails to recompile.
- Also assert: A/B `Worse: 0`; checked population ≥ the legacy population; baseline is
  regenerated when the population changes.
- Document the exact corpus command, cap, and denominator so results are reproducible.

## Acceptance

- The baseline and gate exist and are wired into a runnable check.
- The gate demonstrably fails on a seeded `Exact → RecompileFail` regression.
- Every RTS decline maps to a named typed reason (no silent drops); a failure
  taxonomy is present.

## Notes

Carries forward the definition-of-done from the now-closed RTS trackers (#2050,
#2527, #2526): no false `Exact`, named decline reasons, and the measured member-
identity and generated-state buckets become a living burn-down checklist for the
follow-up issues.

richlander:
- Closes #2776
- Harness-only: adds an RTS-parity regression gate and its committed burn-down manifest; no product output changes.
- Evidence revision: `f869b295`

## Change

**Conclusion:** **PASS** — the RTS compile-back path had no mechanical guard against losing parity the product already proved; this adds a baseline-relative gate over a small committed burn-down that fails on any *new* Exact-to-recompile-failure and names the offender.

The gate is the measurement backbone the rest of the RTS-reversal work (#2777–#2781) cites: as those issues drain RTS's C# knowledge into the product seam, this proves no method that recompiled `Exact` under the product oracle silently starts failing to recompile under ReturnToSender.

### Design note (why not an absolute gate or a full snapshot baseline)

Running the gate over the real 14-assembly corpus revealed **7 methods that already recompile `Exact` under the product oracle but `RecompileFail` under RTS on `main`**. Two consequences:

- An *absolute* "any Exact→recompile-failure is a hard failure" gate would fail every `rts-parity` run today — useless as a regression signal.
- A full committed snapshot baseline would add a second ~68MB opaque JSON (matching `real-world-baseline.json`) that carries 7 rows of signal and that no reviewer can read.

So the gate is **baseline-relative against a small, tool-emitted burn-down manifest** (`tools/DecompilerHarness/corpus/rts-parity-burndown.json`, 1.5KB, 7 named rows). It fails only when a **new** row appears that is not already in the manifest, and reports rows that stopped failing as `resolved` so the manifest is trimmed on regeneration. The known rows are the "living burn-down checklist" the issue asked for — a visible list the follow-up issues shrink.

## Scope and safety

- **Root cause:** no mechanical assertion existed that RTS keeps parity with the product compile-back oracle across the reversal migration.
- **Fix shape:** harness orchestration/oracle/reporting — a pure gate (`EvaluateRtsParityBurndown`) plus wiring (`EnforceRtsParityBurndown`) and two CLI flags.
- **Product impact:** none. No product/decompiler code changed; the gate only reads product-produced fidelity data.
- **RTS boundary:** the gate does not add C# knowledge to RTS. It reads `FidelityCheck`/`FidelityReference` the harness already records and compares method identities. Consistent with the RTS-orchestrator direction.
- **Scope boundary:** the 7 pre-existing gaps are recorded as the burn-down, not fixed here; #2777–#2781 drain them.

## Evidence

| Check | Result |
| --- | --- |
| Harness build | Pass (`0 Warning(s), 0 Error(s)`) |
| Focused tests | `CorpusSensorComparisonTests` 30/0 |
| Decompiler targeted suite | `CorpusSensorComparisonTests` + `ReturnToSenderPrototypeTests` 144/0 |
| Real corpus emit | 14 assemblies, 89,065 methods; `rts-parity` cap 3 → wrote 7-row burn-down |
| Real corpus gate round-trip | re-run with `--rts-parity-burndown` → 7 known / 0 new / **exit 0** |
| Seeded regression | unit test: a gap not in the manifest → `NewRegressions` fails and is named |
| Markdownlint | Pass (`README.md`) |

Named burn-down (the failure taxonomy the issue requires — every decline maps to a named `Assembly!Type::Method#Overload` and a typed status):

```text
DotnetInspector.Packages!DotnetInspector.Packages.SymbolPackageDownloader::CheckPdbHeader#0 [RecompileFail]
ILInspector.Analysis!ILInspector.Analysis.LibraryBodyIndex::HollowUnsafeMethods#0 [RecompileFail]
ILInspector.MetadataPrimitives!ILInspector.Metadata.TypeResolver::GetTypeNameFromReference#0 [RecompileFail]
Newtonsoft.Json!Newtonsoft.Json.JsonTextReader::get_PropertyNameTable#0 [RecompileFail]
Newtonsoft.Json!Newtonsoft.Json.Utilities.Base64Encoder::WriteCharsAsync#0 [RecompileFail]
System.CommandLine!System.CommandLine.EnvironmentVariablesDirective.EnvironmentVariablesDirectiveAction::SetEnvVars#0 [RecompileFail]
System.CommandLine!System.CommandLine.ParseResult::GetValue#2 [RecompileFail]
```

## Reproduce

```bash
bash eng/prepare-decompiler-corpus.sh /tmp/corpus-assemblies.txt
assemblies=$(tr '\n' ' ' < /tmp/corpus-assemblies.txt)
dotnet run --project tools/DecompilerHarness -c Release -- $assemblies \
  --compile-cap 0 --corpus-fidelity-cap 3 --corpus-fidelity-oracle rts-parity \
  --rts-parity-burndown tools/DecompilerHarness/corpus/rts-parity-burndown.json
```

## Review

Requesting the mandated two-model adversarial review (GPT-5.5 + Gemini 3.1 Pro) in isolated worktrees. Key attack points: is the DisplayMethod key stable/unique enough to key the burn-down; does the gate correctly distinguish new vs known vs resolved; can a real regression hide as "same" or "resolved"; is cap 3 a meaningful denominator.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CorpusSensorComparisonTests.cs
+++ b/src/ILInspector.Decompiler.Tests/CorpusSensorComparisonTests.cs
@@ -32,6 +32,75 @@ public class CorpusSensorComparisonTests
     }
 
     [Fact]
+    public void ExactReferenceRecompileRegressions_FlagsExactToRecompileAndContextFail()
+    {
+        var snapshot = ReturnToSenderSnapshot(
+            RtsMethod("Recompile", fidelityReference: "Exact", fidelityCheck: "RecompileFail"),
+            RtsMethod("Context", fidelityReference: "Exact", fidelityCheck: "ContextFail"));
+
+        var offenders = CorpusSensor.ExactReferenceRecompileRegressions(snapshot);
+
+        Assert.Equal(2, offenders.Length);
+        Assert.Contains(offenders, m => m.Method == "Recompile");
+        Assert.Contains(offenders, m => m.Method == "Context");
+    }
+
+    [Fact]
+    public void ExactReferenceRecompileRegressions_IgnoresRescuedSameAndUnpaired()
+    {
+        var snapshot = ReturnToSenderSnapshot(
+            RtsMethod("Rescued", fidelityReference: "OpcodeDiff", fidelityCheck: "Exact"),
+            RtsMethod("Same", fidelityReference: "Exact", fidelityCheck: "Exact"),
+            RtsMethod("OpcodeDrift", fidelityReference: "Exact", fidelityCheck: "OpcodeDiff"),
+            RtsMethod("NoReference", fidelityReference: null, fidelityCheck: "RecompileFail"));
+
+        Assert.Empty(CorpusSensor.ExactReferenceRecompileRegressions(snapshot));
+    }
+
+    [Fact]
+    public void ExactReferenceRecompileRegressions_OnlyAppliesToReturnToSenderOracle()
+    {
+        var snapshot = Snapshot(
+            totalMethods: 1,
+            fullyRaisedMethods: 1,
+            fullyRaisedBasisPoints: 10_000,
+            pinnedMethods: [RtsMethod("Recompile", fidelityReference: "Exact", fidelityCheck: "RecompileFail")],
+            fidelityOracle: CorpusFidelityOracle.CompileBack);
+
+        Assert.Empty(CorpusSensor.ExactReferenceRecompileRegressions(snapshot));
+    }
+
+    [Fact]
+    public void Compare_FailsAndNamesExactToRecompileFailRegression()
+    {
+        var baseline = ReturnToSenderSnapshot(
+            RtsMethod("Recompile", fidelityReference: "Exact", fidelityCheck: "Exact"));
+        var current = ReturnToSenderSnapshot(
+            RtsMethod("Recompile", fidelityReference: "Exact", fidelityCheck: "RecompileFail"));
+
+        var regressions = CorpusSensor.Compare(baseline, current, []);
+
+        Assert.Contains(
+            regressions,
+            r => r.Contains("RTS parity lost", StringComparison.Ordinal)
+                && r.Contains("Pinned!T::Recompile#0", StringComparison.Ordinal)
+                && r.Contains("RecompileFail", StringComparison.Ordinal));
+    }
+
+    [Fact]
+    public void Compare_PassesWhenReferenceExactStaysExactUnderReturnToSender()
+    {
+        var baseline = ReturnToSenderSnapshot(
+            RtsMethod("Recompile", fidelityReference: "Exact", fidelityCheck: "Exact"));
+        var current = ReturnToSenderSnapshot(
+            RtsMethod("Recompile", fidelityReference: "Exact", fidelityCheck: "Exact"));
+
+        var regressions = CorpusSensor.Compare(baseline, current, []);
+
+        Assert.DoesNotContain(regressions, r => r.Contains("RTS parity lost", StringComparison.Ordinal));
+    }
+
+    [Fact]
     public void Compare_RejectsCorpusProfileMismatch()
     {
         var baseline = Snapshot(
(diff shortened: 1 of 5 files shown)


richlander:
## Adversarial review — reconciled (two clean fixed-head reviews)

Two reviewers from different model families, each in an isolated worktree:
**GPT-5.5** and **Gemini 3.1 Pro**.

### Round 1 — both BLOCKING, strongly convergent

| Finding | GPT-5.5 | Gemini 3.1 Pro | Severity |
| --- | --- | --- | --- |
| Burn-down flags silently toothless with the default compile-back oracle (Enforce never runs → exit 0); GPT also flagged the `--corpus-fidelity-cap 0` variant (no parity population → all rows falsely "resolved" → exit 0) | ✓ | ✓ | Critical / High |
| `--emit` + `--enforce` at the same path self-baselines a new regression (emit writes gaps, then reads them back as "known") | ✓ | — | Medium |
| `{}` manifest (no `rows`) → `NullReferenceException` on the default `ImmutableArray` | — | ✓ | Medium |

Both independently reproduced the silent-bypass in their worktrees, so this was
the load-bearing finding.

### Fixes (commit b006a6db)

- `ValidateRtsParityBurndownFlags` (pure, unit-tested) called from `Run` before
  the corpus capture: rejects the burn-down flags without
  `--corpus-fidelity-oracle rts-parity`, without a positive `--corpus-fidelity-cap`,
  and when emit/enforce share a path. Fails fast with a clear message.
- `ReadRtsParityBurndown` normalizes a default `Rows` array to empty.
- Five focused tests for the three rejections, the well-formed pass, and the
  missing-`rows` read. Verified all three CLI bypasses now exit 1.

### Round 2 — fixed head b006a6db, both CLEAN

- **GPT-5.5:** CLEAN — "The fixed guard now rejects both original bypasses and
  same-path emit/enforce; focused tests pass."
- **Gemini 3.1 Pro:** CLEAN — verified the validator fails before the run with
  exit 1, the NRE fix holds (also checked a `"method": null` row — `HashSet` +
  `StringComparer.Ordinal` tolerates null natively), and no new defect introduced.

### Dismissed / non-actioned

- Enforcing at a different cap than the manifest was generated at can report a
  row "resolved" that was merely outside the sampled subset. Informational only —
  it never yields a false PASS on a genuinely new regression, and cap 3 is
  pinned/documented. No resolution commit; left as documented behavior.

**Ready to merge** — pending @richlander approval. Will not self-merge.

PR #2790, opened 2026-07-16
Make versioned cache cleanup test deterministic

### LINKED ISSUE #2744: Flaky: CacheCommandTests.CacheMiss_CleansObsoleteVersionedCategories times out under CI threadpool starvation
## Symptom

`DotnetInspector.Tests.CacheCommandTests.CacheMiss_CleansObsoleteVersionedCategories` intermittently fails in CI:

```
Assert.False() Failure
Expected: False
Actual:   True
  at CacheCommandTests.CacheMiss_CleansObsoleteVersionedCategories() src/dotnet-inspect.Tests/CacheCommandTests.cs:72
```

Line 72 is `Assert.False(Directory.Exists(oldDir));` — the obsolete versioned category (`pkg-index-v9`) was expected to be deleted by background cache maintenance but still exists.

Observed on an unrelated PR run (#2737, which touches only `CSharpTypeProducer`/RTS — nothing in the cache path): https://github.com/richlander/dotnet-inspect/actions/runs/29300962841/job/86984482167 (`test (ubuntu-26.04, linux-x64)`, 1358 total, 1 failed).

## Evidence it is flaky, not a real regression

- Passes locally 3/3 in ~60ms each: `dotnet run --project src/dotnet-inspect.Tests -c Release --no-build -- -method "*CacheMiss_CleansObsoleteVersionedCategories"`.
- The failing PR does not touch `DotnetInspector.Core` cache code.

## Root cause (timing race)

The test drives the *fire-and-forget* background maintenance path:

1. `CoreCache.TryGet(...)` on a miss schedules `s_maintenanceTask = Task.Run(() => CleanupVersionedCategories(...))` (`CoreCache.cs:447`).
2. The test then calls `CoreCache.CancelAndWaitForMaintenance(TimeSpan.FromSeconds(5))`, which does `task.Wait(timeout)`; on timeout it cancels and waits only 25ms more before returning (`CoreCache.cs:378-382`).
3. It asserts `oldDir` was deleted.

Under CI threadpool starvation (many parallel xUnit collections; this class is `[Collection("Console")]` and runs alongside the full suite), the `Task.Run` maintenance body can fail to acquire a threadpool thread and complete its directory delete within the 5s window. The wait then times out, cancels, and returns before `oldDir` is removed → `Directory.Exists(oldDir)` is still `True`.

## Suggested fixes (pick one)

- Add an internal *synchronous* maintenance entrypoint (e.g. `CoreCache.RunVersionedCategoryCleanup(root, categories)`) and have the test call it directly, so the assertion does not depend on background scheduling latency. Preferred — deterministic and still exercises the real cleanup logic.
- Or make the test wait deterministically for the scheduled task to *complete* (not a fixed timeout) before asserting, e.g. expose the maintenance `Task` and `await` it.
- Avoid raising the 5s timeout as the fix; it only widens the window and keeps the race.

Low priority (test-only flake), but it can red-herring unrelated PRs.

autocarl:
## Summary

- add an internal synchronous entry point over the existing versioned-category cleanup path
- make the cache test invoke that path directly instead of depending on `Task.Run` scheduling and a fixed timeout
- leave the public API and production fire-and-forget maintenance behavior unchanged

## Validation

- `dotnet build src/DotnetInspector.Core/DotnetInspector.Core.csproj -c Release -p:DefaultTargetFramework=net10.0 -p:PublishAot=false`: 0 warnings, 0 errors
- focused synchronous-cleanup canary: 100 consecutive passes
- patch applied and revalidated against current upstream `main` (`c8d26de4`): clean diff and strict Core build
- two independent reviews found no blockers; the one LOW documentation finding was resolved by stating the cumulative result semantics explicitly

## Environment

The available centrally managed SDK is .NET 10.0.301. The full CLI test project requires the repository's .NET 11 daily SDK and cannot compile on the unchanged .NET 10 baseline because `UnionAttribute`/`IUnion` are unavailable. CI provides the required .NET 11 validation.

Fixes #2744

### DIFF
--- a/src/DotnetInspector.Core/CoreCache.cs
+++ b/src/DotnetInspector.Core/CoreCache.cs
@@ -358,6 +358,26 @@ public static class CoreCache
     }
 
     /// <summary>
+    /// Runs registered versioned-category cleanup synchronously against the active cache root.
+    /// </summary>
+    /// <returns>The cumulative maintenance totals observed after cleanup completes.</returns>
+    internal static CacheMaintenanceResult RunVersionedCategoryCleanup()
+    {
+        string root;
+        VersionedCacheCategory[] categories;
+        lock (s_maintenanceLock)
+        {
+            root = GetBasePath();
+            categories = [.. s_versionedCategories];
+        }
+
+        CleanupVersionedCategories(root, categories, CancellationToken.None);
+        return new CacheMaintenanceResult(
+            Interlocked.Read(ref s_maintenanceBytesFreed),
+            Volatile.Read(ref s_maintenanceDirectoriesDeleted));
+    }
+
+    /// <summary>
     /// Waits briefly for in-flight cleanup to finish, cancels if it exceeds the timeout, and returns
     /// the amount of obsolete cache data removed so far.
     /// </summary>
--- a/src/dotnet-inspect.Tests/CacheCommandTests.cs
+++ b/src/dotnet-inspect.Tests/CacheCommandTests.cs
@@ -57,7 +57,7 @@ public class CacheCommandTests : IDisposable
     }
 
     [Fact]
-    public void CacheMiss_CleansObsoleteVersionedCategories()
+    public void RunVersionedCategoryCleanup_RemovesObsoleteVersionedCategories()
     {
         var oldDir = Path.Combine(_cacheBasePath, "pkg-index-v9");
         var currentDir = Path.Combine(_cacheBasePath, PackageIndexCache.Category);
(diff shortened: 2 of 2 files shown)


richlander:
Thanks for fixing the flaky maintenance test. #2799 supersedes this with a general intra-process single-flight primitive and makes the existing cache-miss path return its real shared maintenance task, so the test can await production behavior without adding a synchronous bypass. It also adds a keyed production adopter and retry/concurrency coverage. We'll proceed with #2799 for #2744.

richlander:
This PR inspired taking a hard take at the underlying cause.

Issue #2782, opened 2026-07-16
Make ReturnToSender a thin orchestrator with no C# knowledge


richlander:
## Goal

Make ReturnToSender (RTS, the compile-back fidelity harness under
`tools/DecompilerHarness/`) a thin orchestrator over the product seam and Roslyn. RTS
should only: select targets, concatenate product-produced source units, drive Roslyn,
and run the fidelity/oracle comparison. No C# spelling, shape, or composition
knowledge should live anywhere under `tools/DecompilerHarness/`; that knowledge belongs
in the product seam (`src/ILInspector.CSharp`), where recompilation validates the
product actually has it.

## Direction

`ILInspector.CSharp` references only `ILInspector.Metadata` (Roslyn-free, SRM-only,
NativeAOT). `ILInspector.Decompiler` references `ILInspector.CSharp`, never the
reverse. The seam cannot read Decompiler IR, so any body fact it needs is handed to it
as a neutral Metadata-level DTO. Recent merged work (#2737, #2751, #2762–#2765, #2771)
already moved RTS a long way in this direction; this tracks the remainder.

## Acceptance

Covers all of `tools/DecompilerHarness`, not just the planner:

1. The harness-local `CompileBackCSharpNames` helper is deleted (achieved by the
   composition and closure-index work below, not by re-routing to another harness
   helper).
2. **Boundary audit (origination, not consumption).** A check asserts that no harness
   path *originates* C#: no keyword/escaping tables, no hand-built declaration strings
   for the recompiled shell, and the deleted helper is gone. *Consuming* product APIs
   (`MetadataDeclarationQuery`, `CSharpNaming`, `CSharpFormatter`, `CSharpTypePrinter`)
   is allowed — that is the desired product-ward direction (e.g. `ShellConstraints.cs`
   already forwards `MetadataDeclarationQuery.GetGenericConstraintClauses`). The
   recompiled shell source must come from product artifacts. The oracle/validity/
   comparison layer is an enumerated permanent allowlist with rationale:
   `ReturnToSenderSourceProbe.cs` (Roslyn source comparison), `ValidityCheck.cs`
   (diagnostic signature text), `FidelityCheck.cs` (`IsVolatileField` probe) — none of
   which feed shell reconstruction.
3. The parity gate stays green (no corpus regressions).

## Work

- [ ] #2776 — RTS parity gate (land first; everything cites it)
- [ ] #2777 — Move member/type source composition into the product seam
[...]

PR #2791, opened 2026-07-16
Extract constructor body-facts into the product (#2777 step 1)

richlander:
## Why

ReturnToSender (the compile-back oracle in `tools/DecompilerHarness`) declares
the product decompiler's C# valid, but to reconstruct constructor-chain
initializers and primary-constructor shells it re-implemented its **own** walks
over Decompiler IR (`IrFunction` and its node types). Two harness consumers —
the RTS type planner and `FidelityCheck` — carried the **same** IR walk
verbatim. That is exactly the layering leak #2782 wants gone: the harness reads
product IR to make C#-shape decisions the product should own.

## What

Move the IR-to-fact extraction into the product (which owns the IR) and emit a
neutral, IR-free DTO the harness consumes:

- **`src/ILInspector.CSharp/ConstructorBodyFacts.cs`** (new) — neutral DTO:
  `ChainParameterTypes` (chained-to ctor's parameter type displays) and
  `PrimaryConstructorPrologue` (ordered `PrimaryConstructorFieldStore`s). Pure
  strings/indices; no Decompiler type crosses the seam.
- **`src/ILInspector.Decompiler/ConstructorBodyFactExtractor.cs`** (new) — the
  single owner of the chain-call and primary-constructor-prologue pattern
  matches, previously duplicated in the harness.
- **`ReturnToSenderTypePlanner.cs`** and **`FidelityCheck.cs`** now consume the
  neutral facts. Their duplicated IR walks and the `ConstructorChainCall` /
  `IsUnsupportedChainParameterType` helpers are deleted. Metadata resolution
  (parameter/field lookup, auto-property naming, surface-representability via
  `TypeShellProducer.IsUnsupportedSurfaceSignature`) stays harness-side — only
  the IR reads moved.

This is **step 1 of #2777** (dependency root of the #2782 "RTS thin
orchestrator" epic): the harness no longer reads IR to reconstruct constructor
initializers. Steps 2–3 (seam owns composition, delete `CompileBackCSharpNames`)
build on this DTO.

## Behavior

Pure refactor — the extractor replicates the moved pattern matches verbatim, so
RTS verdicts are unchanged.

## Evidence

- RTS oracle suites unchanged at baseline: **Evidence 3 / FixtureCatalog 82 /
  Prototype 114**; `FidelityCheckGeneratedFilter 25`; `GeneratedFixtureCatalog 18`.
- New `ConstructorBodyFactExtractorTests` (3): chained ctor → chain parameter
  types; primary-constructor prologue → ordered field stores; non-constructor →
  no facts.
- **RTS parity corpus gate** (`--corpus-fidelity-oracle rts-parity
  --rts-parity-burndown …`, 14-assembly corpus): **0 new** Exact→recompile-fail
  rows; the 7 committed burn-down rows are unchanged (gate exit 0).

## Review

RTS behavior-touching → requesting two adversarial reviews (GPT-5.5 + Gemini 3.1
Pro) in isolated worktrees. Do **not** merge without @richlander approval.

### DIFF
--- a/src/ILInspector.CSharp/ConstructorBodyFacts.cs
+++ b/src/ILInspector.CSharp/ConstructorBodyFacts.cs
@@ -1,0 +1,33 @@
+namespace ILInspector.CSharp;
+
+/// <summary>
+/// Neutral, IR-free constructor body facts extracted from a decompiled constructor.
+/// ReturnToSender and the C# seam use these to reconstruct constructor-chain
+/// initializers and primary-constructor shells without reading Decompiler IR, so the
+/// IR-to-fact extraction lives in the product (ILInspector.Decompiler, which owns the
+/// IR) rather than in the harness. All fields are plain strings and indices; no
+/// Decompiler type crosses this boundary.
+/// </summary>
+public sealed record ConstructorBodyFacts(
+    IReadOnlyList<string>? ChainParameterTypes,
+    IReadOnlyList<PrimaryConstructorFieldStore>? PrimaryConstructorPrologue)
+{
+    /// <summary>
+    /// No chain call and no primary-constructor prologue: the shape used for
+    /// non-constructor methods and bodies with neither fact.
+    /// </summary>
+    public static ConstructorBodyFacts None { get; } = new(null, null);
+}
+
+/// <summary>
+/// One <c>this.field = argN;</c> assignment in a primary-constructor prologue, in
+/// source order. <paramref name="SourceArgumentIndex"/> is the IL argument slot the
+/// stored value came from (0 is <c>this</c>, so a real parameter is &gt;= 1),
+/// <paramref name="FieldName"/> is the stored field's metadata name, and
+/// <paramref name="BackingPropertyName"/> is the property name when the field backs an
+/// auto-property, or null when there is no such proof.
+/// </summary>
+public sealed record PrimaryConstructorFieldStore(
+    int SourceArgumentIndex,
+    string FieldName,
+    string? BackingPropertyName);
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -4611,3 +4611,31 @@ public class ForeachElementProbe
         return total;
     }
 }
+
+// Fixtures for ConstructorBodyFactExtractorTests. The neutral constructor
+// body-fact extractor pattern-matches decompiled IR, so these compiled
+// constructor shapes exercise the chain-call and primary-constructor-prologue
+// detections directly.
+public class ChainedCtorSample
+{
+    public int Value;
+    public string Label;
+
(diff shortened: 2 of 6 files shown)


richlander:

Per the two-reviewer policy for RTS behavior-touching PRs, this change was
reviewed by two non-Opus models in **isolated worktrees** at the PR head
(`15f95031`), each tasked to falsify the "verbatim / behavior-neutral" claim.

| Reviewer | Verdict |
| --- | --- |
| GPT-5.5 | CLEAN |
| Gemini 3.1 Pro | CLEAN |

**Where they agreed (verified independently):**
- Both chain matchers translated verbatim, including the null-vs-empty-list
  semantics: no chain call → `null` gates the initializer off (old
  `chainCall?.Callee` null == new `ChainParameterTypes` null); a 0-real-param
  base/`this()` call → empty list, `All()` trivially true, `Count == 0` — the
  same-arity-sibling binding target logic is unchanged.
- Primary-constructor prologue guards translated exactly (`chainIndex is not > 0`,
  post-chain `Return`-only, `StoreField{Instance:LoadArgument{Index:0},
  Value:LoadArgument{Index:>0}}`, empty→null collapse). The `chainIndex >= 1`
  guard means a non-null prologue always has ≥1 store, matching the old harness.
- `FidelityCheck` preserved: it resolves the property name from metadata via
  `AutoPropertyNameForBackingField(reader, declaringType, fieldStore.FieldName)`
  and correctly ignores the DTO's `BackingPropertyName` payload (which it never
  used).
- Filtering safely moved from `TypeRef`→`string` mapping in the caller to the
  extractor; `TypeShellProducer.IsUnsupportedSurfaceSignature` takes a string, so
  it's equivalent.
- Layering clean: DTO in `ILInspector.CSharp` carries only `int`/`string`/lists;
  no `IrFunction`/IR node leaks; dependency direction Decompiler → CSharp only.
- No missed consumer of the deleted `ConstructorChainCall` /
  `IsUnsupportedChainParameterType` helpers or the old
  `PrimaryConstructorFromPrologue(IrFunction)` signature.

**Divergences:** none — both reached CLEAN with no findings.

**Runs (both reviewers):** full slnx build clean; `ConstructorBodyFactExtractorTests`
3/3, `ReturnToSenderEvidenceTests` 3/3, `ReturnToSenderFixtureCatalogTests` 82/82,
`ReturnToSenderPrototypeTests` 114/114, `FidelityCheckGeneratedFilterTests` 25/25;
both worktrees stayed `git status` clean throughout.

No follow-up changes required (no actionable findings). Awaiting @richlander
approval — will not self-merge.

richlander:
**Ready to merge.** Two CLEAN adversarial reviews (GPT-5.5 + Gemini 3.1 Pro) at the current head, no findings; all merge-blocking validation green (full slnx build, RTS oracle suites unchanged at baseline, RTS parity corpus gate 0 new regressions). Not self-merging — awaiting @richlander.

PR #2799, opened 2026-07-17
Add intra-process cache single-flight coordination

### LINKED ISSUE #2744: Flaky: CacheCommandTests.CacheMiss_CleansObsoleteVersionedCategories times out under CI threadpool starvation

richlander:
## Summary

- add `AsyncCache<TKey, TValue>` as the Core-owned intra-process
  single-flight primitive: concurrent callers receive the exact shared task,
  accepted results are retained, and faults, cancellation, and rejected
  results remain retryable
- adopt it for same-URL source acquisition so one `SourceFetcher` performs one
  disk/network operation per URL
- make cache-miss maintenance expose and retain its real production task with
  per-run progress, allowing the deterministic cleanup test to await production
  behavior rather than use a synchronous bypass or timeout

This supersedes #2790 and fixes #2744 with a general product capability rather
than a test-only cleanup seam.

## Boundaries

This PR addresses intra-process coordination. Atomic file publication,
transactional package-directory publication, and cross-process coordination
remain separate follow-ups because they have different commit and locking
models.

## Validation

- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config` on
  .NET SDK `11.0.100-preview.6.26359.118`: 0 warnings, 0 errors
- `DotnetInspector.Services.Tests`: 264 passed
- `dotnet-inspect.Tests`: 1,902 passed, 10 expected missing-ilasm/ildasm skips
- focused concurrency coverage proves exact task identity, one producer,
  successful retention, and retry after fault, cancellation, or rejected value
- production `SourceFetcher` canary proves 32 requests share one network
  operation; maintenance coverage awaits the cache-miss operation itself

## Adversarial review

Reviewed at exact head `ee83cb71` against `b62fd502`:

- Claude Opus 4.8: **CLEAN**
- Gemini 3.1 Pro: **CLEAN**

Gemini additionally ran a 64-thread exact-task/ABA repro and the complete
Services test suite. Neither reviewer reported blocking, non-blocking, or
pre-existing findings. The explicit non-actions are the publication and
cross-process boundaries above.

### DIFF
--- a/src/DotnetInspector.Core/AsyncCache.cs
+++ b/src/DotnetInspector.Core/AsyncCache.cs
@@ -1,0 +1,94 @@
+using System.Collections.Concurrent;
+
+namespace DotnetInspector.Core;
+
+/// <summary>
+/// Shares one asynchronous resolution per key and retains successful results.
+/// Faulted and cancelled resolutions are removed so a later request can retry.
+/// Cancellation captured by a value factory affects every waiter sharing that
+/// resolution; caller-only cancellation should cancel the wait, not the factory.
+/// </summary>
+public sealed class AsyncCache<TKey, TValue> where TKey : notnull
+{
+    private readonly ConcurrentDictionary<TKey, Lazy<Task<TValue>>> _entries;
+
+    public AsyncCache(IEqualityComparer<TKey>? comparer = null)
+    {
+        _entries = new ConcurrentDictionary<TKey, Lazy<Task<TValue>>>(
+            comparer ?? EqualityComparer<TKey>.Default);
+    }
+
+    /// <summary>
+    /// Returns the existing resolution task for <paramref name="key"/>, or
+    /// publishes one new resolution task. Concurrent callers receive the exact
+    /// same task.
+    /// </summary>
+    /// <param name="shouldCache">
+    /// Optional predicate that decides whether a successfully resolved value
+    /// remains cached. A rejected value is still returned to current waiters,
+    /// but a later request can retry. The factory and predicate from the entry
+    /// that wins publication define the shared resolution.
+    /// </param>
+    public Task<TValue> GetOrAddAsync(
+        TKey key,
+        Func<TKey, Task<TValue>> valueFactory,
+        Func<TValue, bool>? shouldCache = null)
+    {
+        ArgumentNullException.ThrowIfNull(valueFactory);
+
+        Lazy<Task<TValue>>? candidate = null;
+        candidate = new Lazy<Task<TValue>>(
+            () => ResolveAsync(
+                key,
+                valueFactory,
+                shouldCache,
+                candidate!),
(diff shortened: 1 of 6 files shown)


richlander:
Ready to merge

PR #2797, opened 2026-07-17
Move C# identifier producer into the ILInspector.CSharp seam (#2777 step 2 slice)

richlander:
## Why

Part of #2777 (make ReturnToSender a thin orchestrator; epic #2782). RTS's
compile-back path and other harness call sites reached into the **Decompiler's
internal** `CSharpNaming` for C# identifier spelling (keyword `@`-escaping +
unspellable-metadata-name sanitization). That put a piece of the C# spelling
policy inside the Decompiler and forced the harness to depend on Decompiler
internals for it.

This PR relocates the identifier **producer** into the `ILInspector.CSharp`
seam so the seam owns identifier spelling and consumers delegate.

## What

- **NEW** `src/ILInspector.CSharp/CSharpIdentifier.cs` — the authoritative,
  position-agnostic identifier producer: `IsUsable`, `IsEscapable`,
  `IsIdentifierLike`, `Escape`, `Sanitize`, `SanitizeUnspellable`. It carries the
  **exact** base-keyword + `await` policy the Decompiler enforced.
- `CSharpNaming` (Decompiler) producer members become **thin forwarders**; the
  private policy internals (`RequiresEscape`, `ReservedKeywords`,
  `IsIdentifierLike`, `SanitizeUnspellableIdentifier`) are deleted.
- Harness repoints its three producer reach-ins
  (`CompileBackCSharpNames.SanitizeGeneratedTypeSegments`/`.Identifier`,
  `ValidityCheck.ParameterText`) at `CSharpIdentifier`.

## Behavior-preserving — three deliberate notes

1. **Two escapers coexist by design.** `CSharpIdentifier.Escape` is
   *position-agnostic*: it escapes only names unsafe as a bare identifier in
   **any** position (reserved keywords + `await`). It deliberately leaves the
   *declaration-only* contextual keywords (`record`/`required`/`init`/`file`/`scoped`)
   bare — they are legal identifiers in the expression/body positions where raised
   and compile-back identifiers live. Declaration-position escaping (which also
   covers those) stays in `CSharpDeclarationWriter.EscapeIdentifier`. Preserving
   the base+`await` set (not the extended set) is what keeps Decompiler printer
   output byte-identical. Locked by negative tests.
2. **Name *decoding* stays in the Decompiler.** `MethodName`/`SourceMethodName`
   (un-mangling `<Enclosing>g__Local|N_M`), backing-field decoding, etc. are a
   generated-name-decoding concern, not identifier spelling. Out of scope.
[...]

PR #2803, opened 2026-07-17
Key ReturnToSender's closure index on the seam identifier producer

### LINKED ISSUE #2778: Key ReturnToSender's closure index on the seam's emitted identifier
## Problem

Deleting the harness-local `CompileBackCSharpNames` (see #2777) breaks Roslyn's
missing-type closure expansion. RTS indexes SRM metadata into a `FullTypes` map keyed
by the **escaped C# display name** (`CompileBackTypeIdentity.FromDefinition`,
`tools/DecompilerHarness/ReturnToSenderTypePlanner.cs:~296-317`); when Roslyn reports a
missing type, RTS extracts the syntax node and looks that name up
(`ReturnToSender.cs:~1662-1665`, `ClosureDiagnosticEvidence.cs:~337-345`). Remove the
harness escaping and the lookup key no longer matches.

## Why keying on the raw metadata name is wrong

Roslyn's `SimpleNameSyntax.Identifier.ValueText` unescapes `@`-keywords (so `@class`
matches metadata `class`) but it **cannot un-mangle a structurally sanitized name**.
Once the seam emits the sanitized identifier for `<file>__Foo` (e.g. `__file___Foo`),
Roslyn reports the *sanitized* name missing, which would never match a raw
`<file>__Foo` index key. So the index must be keyed by whatever the seam actually
emits.

## Scope

Key the closure index using the **same canonical identifier producer the seam emits
with** (the escaping + sanitization capability the seam owns after #2777), so
index-key == emitted-name by construction. Consuming that product API from the harness
is the desired product-ward direction and keeps RTS a thin orchestrator. Normalize
nested types to the seam's spelling (`TypeRef.Name` uses `+`, `TypeRef.cs:~81`).
Remove the `Identifier`/`EscapeNamespace` escaping from all three call sites.

## Acceptance

- Closure expansion works with no harness-side C# escaping.
- The `CompileBackCSharpNames` callers in `ReturnToSender.cs` and
  `ClosureDiagnosticEvidence.cs` are gone.
- Tests cover keyword, generic-arity, nested-type, and generated/file-local
  (sanitized) cases.
- The parity gate (#2776) stays green.

## Depends on

#2777 (the seam must own identifier production first).

richlander:
Closes #2778.

## What

The RTS closure index and its missing-type lookups escaped/sanitized C# identifiers through the harness-local `CompileBackCSharpNames` wrapper. Deleting that wrapper (the #2777 endgame) would break Roslyn's missing-type closure expansion: Roslyn reports the *sanitized* name the product seam actually emits (e.g. `__file___Foo` for a `<file>__Foo` type), which never matches a raw metadata index key.

This points the three closure sites at the seam producer directly, so the index key equals the emitted name by construction and no C# escaping remains in the closure path:

- `CompileBackTypeIdentity.FromDefinition` (the `FullTypes` key builder) keys on `CSharpIdentifier.Sanitize` + `CSharpFormatter.EscapeNamespace`. Arity stripping — a metadata-name concern, not C# spelling — is unchanged.
- The CS0234 lookup in `ReturnToSender.cs` and the receiver spelling in `ClosureDiagnosticEvidence.TypeName` use the same seam APIs; both files now carry zero `CompileBackCSharpNames` references.

## Why it's low risk

Behavior-preserving: `CompileBackCSharpNames.Identifier` already delegated to `CSharpIdentifier.Sanitize` and `EscapeNamespace` to `CSharpFormatter.EscapeNamespace` (post-#2797), so the emitted spellings are identical. This removes the indirection so #2777 step 3 can delete `CompileBackCSharpNames` without silently breaking closure expansion.

## Evidence

- New real-metadata tests (`CompileBackTypeIdentityTests`) pin the `FullTypes` key to the seam spelling for keyword (`@class`), generic-arity (`Container` backtick 1 stripped to `Container`), nested-type (`.`-qualified, no `+`), keyword-namespace (`@for.@class`), and generated/file-local (`<...>__Widget` sanitized to a spellable identifier with no `<`/`>`) cases.
- Targeted suites green: `CompileBackTypeIdentityTests` + `ClosureDiagnosticEvidenceTests` + `ReturnToSenderPrototypeTests` + `FidelityCheckGeneratedFilterTests` → 181/0.
- The #2776 parity gate stays green on the pinned 14-assembly corpus (`--compile-cap 0 --corpus-fidelity-cap 3 --corpus-fidelity-oracle rts-parity`): 7 known burn-down rows, 0 new regressions, exit 0.

## Acceptance (#2778)

- [x] Closure expansion works with no harness-side C# escaping.
- [x] `CompileBackCSharpNames` callers in `ReturnToSender.cs` and `ClosureDiagnosticEvidence.cs` are gone.
- [x] Tests cover keyword, generic-arity, nested-type, and generated/file-local (sanitized) cases.
- [x] The parity gate (#2776) stays green.

Adversarial review to follow.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CompileBackTypeIdentityTests.cs
+++ b/src/ILInspector.Decompiler.Tests/CompileBackTypeIdentityTests.cs
@@ -1,0 +1,147 @@
+using System.Collections.Immutable;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
+
+using ILInspector.CSharp;
+using ILInspector.DecompilerHarness;
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace ILInspector.Decompiler.Tests;
+
+// The closure index (CompileBackTypeIdentity.FromDefinition) keys the FullTypes map on
+// the identifier the product seam actually emits, so a missing type Roslyn reports —
+// which reflects the seam's emitted spelling — matches the index key by construction
+// (issue #2778). Each case pins index-key == CSharpIdentifier.Sanitize(<arity-stripped
+// metadata name>) against a real compiled assembly.
+public class CompileBackTypeIdentityTests
+{
+    [Fact]
+    public void FromDefinition_KeywordType_KeysOnKeywordEscapedSpelling()
+    {
+        using var compiled = Compile("namespace Sample { public class @class {} }");
+        var identity = IdentityOf(compiled.Reader, "class");
+
+        Assert.Equal("@class", identity.DisplayName);
+        Assert.Equal("Sample.@class", identity.FullName);
+        Assert.Equal(Expected("class"), identity.DisplayName);
+    }
+
+    [Fact]
+    public void FromDefinition_GenericType_KeysOnArityStrippedSpelling()
+    {
+        using var compiled = Compile("namespace Sample { public class Container<T> {} }");
+        var identity = IdentityOf(compiled.Reader, "Container`1");
+
+        Assert.Equal("Container", identity.DisplayName);
+        Assert.Equal("Sample.Container", identity.FullName);
+        Assert.Equal(Expected("Container`1"), identity.DisplayName);
+    }
+
+    [Fact]
+    public void FromDefinition_NestedType_QualifiesWithDotSeparators()
+    {
+        using var compiled = Compile(
+            "namespace Sample { public class Outer { public class Inner {} } }");
+        var inner = IdentityOf(compiled.Reader, "Inner");
+
+        Assert.Equal("Inner", inner.DisplayName);
+        Assert.Equal("Sample.Outer.Inner", inner.FullName);
+        Assert.DoesNotContain('+', inner.FullName);
+    }
+
+    [Fact]
+    public void FromDefinition_KeywordNamespace_EscapesEachSegment()
+    {
+        using var compiled = Compile("namespace @for.@class { public class Widget {} }");
+        var identity = IdentityOf(compiled.Reader, "Widget");
+
+        Assert.Equal("@for.@class.Widget", identity.FullName);
+    }
+
+    [Fact]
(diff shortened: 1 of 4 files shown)


richlander:
## Adversarial review — reconciled (two clean first-pass reviews)

Two reviewers from different model families, each in an isolated worktree at head `c1224b50`: **GPT-5.5** and **Gemini 3.1 Pro**. Both **CLEAN**, no findings.

### Attack points and outcomes

| Attack | Result |
| --- | --- |
| Behavior-preserving? | Confirmed. At the merge base the wrapper methods were already literal pass-throughs (`Identifier => CSharpIdentifier.Sanitize`, `EscapeNamespace => CSharpFormatter.EscapeNamespace`); the PR inlines that delegation. |
| index-key == emitted-name (the highest-value attack) | Confirmed. The seam printer sanitizes generated names via `CSharpFormatter.NormalizeGeneratedMetadataTypeName`, which is character-for-character identical to `CSharpIdentifier.SanitizeUnspellable` for generated names (both prepend `_` on a non-identifier start and map every non-`[A-Za-z0-9_]` char to `_`). The keyword-`@` difference is safe: generated names contain `_` (never a keyword), and for non-generated types the printer's `EscapeIdentifier` mirrors `Sanitize`'s keyword `@`-escape, which Roslyn strips when reporting a missing name. |
| `StripArity` retention | Correct. The seam printer natively drops the arity backtick (`List\`1` → `List`), matching the key. Arity is a metadata concern, not C# spelling. |
| Nested `+` vs `.` flattening | No site passes a `+`-qualified name to `Sanitize`; all three use simple names (`TypeDefinition.Name`, Roslyn `INamedTypeSymbol.Name`, syntax simple name) and join with `.`. |
| New tests vacuous? | Meaningful. The file-local test pins the compiler's actual `<...>__Widget` metadata name to the sanitized spelling. All targeted suites pass on .NET 11 preview.6. |

No findings were raised, so there is no fix/re-review cycle. Both reviews are clean against the exact PR head.

**Ready to merge** — pending @richlander approval and CI. Will not self-merge.

Issue #2807, opened 2026-07-17
Bound .NET tool wrapper redirect cycles


richlander:
`PackageExtractor.MaybeRedirectToolWrapperAsync` rejects only a wrapper that
redirects directly to its own package ID. A malicious or malformed cycle across
multiple wrapper packages (for example A -> B -> A) recursively calls
`ExtractPackageAsync` without retaining the redirect chain and can eventually
terminate the process with stack overflow.

Track visited package IDs through one redirect operation and return a visible
`PackageExtractionOutcome.Error` when a redirect repeats an ID. Add direct
self-cycle, two-package cycle, and ordinary wrapper-to-payload coverage.

This was identified while reviewing #2804 and is pre-existing at its base
commit (`793c14c9`); the transactional package-acquisition change does not alter
the redirect recursion.

richlander:
Implemented in PR #2810 at exact head `eee3a44b`.

Test evidence:

- focused package-acquisition class: 12/12 passed, including direct case-variant self-cycle, `A -> B -> A`, and ordinary wrapper-to-payload redirect
- full CLI suite: 1,925 total, 0 failed, 10 expected tool-dependent skips
- services suite: 264 total, 0 failed
- solution Release build and Markdown lint passed
- PR CI passed (`build-net10`, Linux test matrix, and `markdownlint`)
- exact-head adversarial reviews from Claude Opus 4.8 and Gemini 3.1 Pro: both CLEAN

PR #2808, opened 2026-07-17
Make package acquisition transactional

### LINKED ISSUE #2804: Make package acquisition transactional
Package acquisition currently downloads and extracts identical package
coordinates independently within one process, then copies directly into the
final app-cache directory. Readers and competing processes can observe a
partially copied package, and callers receive a temporary path even after the
same contents were cached.

## Goal

Make exact remote package acquisition return one committed, cache-owned entry:

- collapse concurrent requests for the same cache root, normalized package id,
  and version onto one shared task
- fence the transactional layout from legacy direct-copy writers with a
  versioned cache namespace
- stage and validate complete contents before atomically renaming them into the
  final path
- let lock-free competing publishers validate and converge on the rename winner
- retain the downloaded `.nupkg` for package size and signature inspection
- preserve local `.nupkg` extraction as caller-owned temporary content
- keep platform-pack consumers from mutating shared committed package contents

## Evidence

Cover one-download concurrency, failed-acquisition retry, authored-case
ref-only packages, legacy namespace fencing, invalid-final preservation,
temporary/staging cleanup, concurrent publisher convergence without mixed
trees, and immutable platform-pack consumption. Include a real ref-pack canary.

richlander:
## Summary

- collapse concurrent exact remote package requests onto one process-wide
  in-flight task keyed by the final cache coordinate
- publish downloaded package contents and archive through validated,
  same-parent staging followed by atomic rename into the fenced
  `package-content-v2` namespace
- return cache-owned committed paths to remote callers while preserving
  caller-owned temporary extraction for local `.nupkg` files
- make lock-free competing publishers validate and converge on the committed
  winner; invalid finals remain visible and are never deleted under a race
- project platform packs immutably through their own transactional `packs-v2`
  staging/marker/rename boundary
- keep cache-owned marker/archive artifacts out of package file and layout
  output

Fixes #2804.

## Compatibility and boundaries

The durable identity intentionally follows NuGet's coordinate cache semantics:
cache root plus normalized package id and version. Ordered sources choose the
producer on a miss but are not a separate cache identity.

This uses atomic publication and winner convergence without adding an explicit
cross-process lock. Legacy direct-copy package and pack namespaces remain
fenced and unread by the new transactional paths. Local package extraction is
unchanged.

The pre-existing multi-package tool-wrapper redirect cycle is unchanged from
the base and tracked separately by #2807.

## Validation

- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config`:
  0 errors; three pre-existing nullable warnings in metadata tests
- `dotnet-inspect.Tests`: 1,912 passed, 10 expected ilasm/ildasm skips
- `DotnetInspector.Services.Tests`: 264 passed
- fixed-head package transaction stress: 100/100 runs
- real isolated `Microsoft.AspNetCore.App.Ref@10.0.0` canary:
  committed to `package-content-v2`, atomically projected to `packs-v2`, then
  reopened without mutating the committed package
- `markdownlint` passed for the cache documentation

## Adversarial review

Reviewed at fixed head `fabec1a9` against `793c14c9`:

- Claude Opus 4.8: **CLEAN**
- Gemini 3.1 Pro: **CLEAN**

First-round findings were resolved as follows:

- Claude found direct platform-pack publication could expose or corrupt a
  partial derived pack. Commit `fabec1a9` added the fenced transactional
  `packs-v2` publication boundary and concurrent coverage.
- Gemini found the new marker/archive files leaked into package listings.
  Commit `fabec1a9` centralized plumbing filtering and applied it to both file
  rows and `--layout`.
- An earlier adversarial pass found authored-case ref-only packages were
  rejected and platform-pack projection moved files out of shared cache paths.
  Commit `2eecda4b` includes the case-insensitive structural validation and
  immutable copy behavior, with synthetic and real ref-pack evidence.
- Claude also identified the pre-existing wrapper redirect cycle. The relevant
  method is unchanged by this diff; #2807 records the follow-up.

Both reviewers re-reviewed the fixed exact head and reported no outstanding
blocking or non-blocking findings.

### DIFF
--- a/docs/design/version-resolution.md
+++ b/docs/design/version-resolution.md
@@ -153,7 +153,7 @@ each endpoint.
 | --- | --- | --- | --- |
 | NuGet global cache | `~/.nuget/packages/{name}/{version}/` | Permanent | `dotnet restore`, NuGet client |
 | App package cache | `$LOCAL_APP_DATA/dotnet-inspect/package-content-v2/{name}/{version}/` | Permanent | dotnet-inspect |
-| Platform packs | `$LOCAL_APP_DATA/dotnet-inspect/packs/{pack}/{version}/` | Permanent | dotnet-inspect |
+| Platform packs | `$LOCAL_APP_DATA/dotnet-inspect/packs-v2/{pack}/{version}/` | Permanent | dotnet-inspect |
 | Version resolution | `$LOCAL_APP_DATA/dotnet-inspect/versions/` | 1 hour | dotnet-inspect |
 | Package metadata | `$LOCAL_APP_DATA/dotnet-inspect/metadata/` | 1 hour | dotnet-inspect |
 | Symbol miss markers | `$LOCAL_APP_DATA/dotnet-inspect/symbol-misses/` | 1 day | dotnet-inspect |
@@ -195,6 +195,10 @@ directories. Failed acquisition leaves no final entry and can be retried.
 The flight and durable cache identity follow NuGet's coordinate cache model:
 cache root plus normalized package id and version. Source order selects the
 producer on a miss but is not a separate durable identity.
+
+Platform-pack projection never mutates the committed package. It copies the
+selected pack contents into a unique `packs-v2` staging directory and
+atomically publishes the complete derived pack with its own completion marker.
 
 ## Design rationale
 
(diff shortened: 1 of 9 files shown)


richlander:
Ready to merge

PR #2810, opened 2026-07-17
Bound .NET tool wrapper redirect cycles

### LINKED ISSUE #2807: Bound .NET tool wrapper redirect cycles

richlander:
## Summary

- replace recursive .NET tool-wrapper redirects with iterative traversal
- detect repeated package IDs case-insensitively and return the complete redirect chain as a visible error
- preserve ordinary wrapper-to-payload behavior and keep redirects outside exact-coordinate single-flight

Closes #2807

## Evidence

- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-build --no-restore` — 1,925 total, 0 failed, 10 skipped
- `dotnet run --project src/DotnetInspector.Services.Tests -c Release --no-restore` — 264 total, 0 failed
- `npx markdownlint-cli docs/design/cache-concurrency.md`

## Compatibility

Initial version resolution retains the caller’s latest/prerelease policy. Redirected packages continue to use the resolved wrapper version, matching the previous recursive behavior. Local `.nupkg` extraction remains unchanged. Self-redirecting wrappers now return a visible cycle error instead of succeeding with an unusable wrapper that contains no managed libraries.

### DIFF
--- a/docs/design/cache-concurrency.md
+++ b/docs/design/cache-concurrency.md
@@ -55,7 +55,8 @@ The acquisition factory only downloads, extracts, validates, and commits its
 own coordinate. It does not resolve dependencies or wait on another registry
 key. Follow-on work such as dependency traversal, tool-package redirection, and
 platform-pack projection starts only after that exact-coordinate task has
-completed.
+completed. Tool-package redirection is iterative and fails visibly when a
+package id repeats in one redirect chain.
 
 ## Transactional publication
 
--- a/src/DotnetInspector.Packages/PackageExtractor.cs
+++ b/src/DotnetInspector.Packages/PackageExtractor.cs
@@ -83,44 +83,69 @@ public static class PackageExtractor
             return ExtractLocalPackage(packageSource, log, tempDirPrefix);
         }
 
-        var outcome = await DownloadAndExtractPackageAsync(client, packageSource, log, tempDirPrefix, sourceOptions, version, forceLatest, includePrerelease).ConfigureAwait(false);
-        if (!outcome.IsSuccess)
-            return outcome;
+        // Keep redirect traversal outside exact-coordinate acquisition so one
+        // package flight never waits on another package key.
+        var visitedPackageIds = new HashSet<string>(
+            StringComparer.OrdinalIgnoreCase);
+        List<string> redirectChain = [];
+        string currentPackageSource = packageSource;
+        string? currentVersion = version;
+        bool currentForceLatest = forceLatest;
+        bool currentIncludePrerelease = includePrerelease;
 
-        return await MaybeRedirectToolWrapperAsync(outcome.Result!, client, log, tempDirPrefix, sourceOptions).ConfigureAwait(false);
-    }
-
-    /// <summary>
-    /// .NET tool packages built as runtime-specific (NativeAOT) tools ship a thin wrapper
-    /// package whose payload is only a <c>DotnetToolSettings.xml</c> manifest pointing at
-    /// per-RID packages. The wrapper has no managed libraries to inspect, so redirect to the
-    /// portable <c>any</c> RID package (the framework-dependent build) which carries the
-    /// managed assemblies. Detection lives in NuGetFetch (<see cref="NuGetFetch.PackageExtractor.TryGetToolWrapperRedirect"/>).
-    /// </summary>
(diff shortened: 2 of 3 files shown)


richlander:
Exact-head adversarial review reconciliation for `eee3a44b`:

- **Claude Opus 4.8:** CLEAN — reviewed termination, error-chain accuracy, identity/version behavior, cleanup, cache/single-flight boundaries, malformed metadata, and ran the focused class (12/12 passed).
- **Gemini 3.1 Pro:** CLEAN — independently reviewed the same exact head and attack points; no blocking, non-blocking, or pre-existing findings.

No review findings require action.

richlander:
Ready to merge

richlander:
## Review (Claude Fable 5)

**Verdict: no merge blockers.** The iterative rewrite is correct, and the base-context checks that worried me came back clean.

### Verified

- **The dropped temp-dir cleanup is safe.** The old code deleted the wrapper's `TempDir` after a successful redirect. After #2808's transactional commit, the network download path always returns `TempDir: null` (staging is cleaned inside `CommitPackage`), and the local-`.nupkg` path — the only non-null `TempDir` producer — returns before the redirect loop. The deleted block was already dead code; no leak.
- **Version/flag propagation matches the old recursion**: redirected hops pin `result.Version` and reset `forceLatest`/`includePrerelease` to false, exactly what the old recursive call's defaults did.
- The pre-download `visitedPackageIds.Contains(redirectId)` check avoids a wasted fetch before reporting a cycle (confirmed by the `RequestCount == 1` assertion), and the `!visitedPackageIds.Add(result.PackageName)` re-check catches a downloaded package whose nuspec id differs from the requested redirect id but was already visited.

### Findings

1. **Undocumented behavior change for self-redirecting wrappers.** The old code treated `redirectId == result.PackageName` as a no-op and returned the wrapper extraction as *success*; the new code returns a cycle error (the `Wrapper.Direct -> wrapper.direct` test encodes this). The new behavior is right — the wrapper has no managed libraries, so the old success was success-shaped emptiness — but the Compatibility section doesn't mention it. Worth one sentence there.
2. **Chain length is unbounded for distinct ids.** Cycles are caught, but a malicious package could chain N *distinct* wrapper packages, each costing a network download and cache commit. Not a regression (the old recursion had the same exposure plus stack growth), but since the redirect id comes from untrusted `DotnetToolSettings.xml`, a small hop budget (real chains are 1 hop) would close it cheaply. Fine as a follow-up.
3. Minor test gap: the nuspec-id-mismatch branch (`!visitedPackageIds.Add(...)`) and the whitespace-`PackageName` error path are untested. Optional.

### Quality

Style matches the file; the single-flight comment states a real constraint. The three tests (case-insensitive direct cycle, two-package cycle, happy-path redirect landing on the payload's cache path) cover the branches that matter, with request-count assertions proving no redundant fetches. Docs updated in `cache-concurrency.md`.

richlander:
Review feedback disposition:

- **Compatibility note:** addressed in the PR description; self-redirecting wrappers now explicitly document the intentional change from success-shaped emptiness to a visible cycle error.
- **Distinct-ID hop budget:** non-blocking, pre-existing exposure filed as #2813.
- **Optional defensive-branch tests:** no action; the acceptance paths and meaningful observable behavior are already covered.

No code changed, so exact head remains `eee3a44b`; prior clean exact-head reviews and CI remain applicable. Ready to merge.

richlander:

Disposition verified — **clean, no further asks.** The Compatibility section now documents the self-redirect change (visible cycle error instead of success-shaped emptiness), the distinct-ID hop budget is tracked as #2813, and declining the defensive-branch tests is reasonable. Head unchanged at `eee3a44b`; nothing in my review required a code change. Merge-ready from my side.

PR #2811, opened 2026-07-17
Fold RTS compilation-unit composition into CSharpTypePrinter options (#2777)

richlander:
## Why

RTS's C# file assembly (pragma + `[assembly:]`/`[module:]` + usings + type
sources) lived in the harness as a standalone `CSharpCompilationUnitComposer`,
separate from the product printer that actually spells the types. Deciding
*usings* and *type-name rendering* separately is what produces the long,
fully-qualified names in decompiled source — those decisions belong together,
on the printer.

## What

- **`CSharpTypePrinter` now owns file-level composition.** `CSharpTypePrintOptions`
  gains `Usings` / `IncludeUsings` / `AssemblyAttributes` / `ModuleAttributes` /
  `EmitPragmaWarningDisable`; `CSharpTypePrintResult` carries the composed file
  text on `.Source`. Defaults emit usings **without** pragma/assembly attributes;
  RTS opts into pragma + attributes and supplies its IR-derived usings.
- **`CSharpCompilationUnitComposer` (+ `CSharpCompilationUnitSpec`) deleted.** The
  harness calls `PrintBatch(requests, options).Source`.
- **`CSharpNamespaceStyle` enum replaced by a derived rule:** a batch of a single
  print request emits a **file-scoped** namespace; multiple requests (which may
  span namespaces) use **block-scoped**. This flips RTS single-type files to
  file-scoped.
- `CSharpFormatter.StripArity` / `CleanTypeDisplay` (relocated from the harness's
  `CompileBackCSharpNames`) are retained from the earlier carve.

## Behavior change is intentional and IL-neutral

This is **not** a byte-preserving carve. Single-type RTS files now render a
file-scoped namespace. That is safe because **the RTS fidelity oracle compares
opcode streams, not source bytes** (`FidelityCheck.cs` `CompileBackStatus.Exact`
= recompiled-to-same-canonical-opcode-stream). Source formatting is invisible to
the verdict.

## Evidence

- Full `dotnet-inspect.slnx` build clean (rebased on origin/main, 0 behind).
- `CSharp.Tests` **253** (+5 composition tests locking option defaults, using
  dedup/escape, pragma/assembly emission, and the single-vs-multi namespace rule).
- RTS oracle **199** (Evidence 3 / FixtureCatalog 82 / Prototype 114).
- Decompiler fast subset **2496** (printer byte-identity for non-RTS callers).
- **RTS parity corpus gate: exit 0 — 25 exact / 2 opcode diffs / 9 recompile
  failures over 36, 7 committed burn-down rows unchanged** (identical to pre-change
  run, so the namespace flip moves zero verdicts).


richlander:
## Adversarial review — both CLEAN

Two independent adversarial reviews on PR head `9204cfd2`, each in an isolated detached worktree (no shared tree). Both returned **CLEAN**.

**GPT-5.5** — CLEAN. Verified line-by-line: old `WriteCompilationUnit` vs new `Compose`; old `CompileBackCSharpNames.Clean`/`StripArity` vs new `CSharpFormatter.CleanTypeDisplay`/`StripArity`; `rg CompileBackCSharpNames` in `src`/`tools` → no matches. Ran build + CSharp.Tests (250) + RTS oracle classes (FixtureCatalog 82, Evidence 3, GeneratedFixtureCatalog 18, Prototype 114).

**Gemini 3.1 Pro** — CLEAN. Char-for-char equivalence of `Compose` (pragma spacing, `[assembly:]`/`[module:]` formatting, `AppendLine(unit.Source)`), usings `Select(EscapeNamespace).Distinct(Ordinal).Order(Ordinal)` match, and confirmed the pre-existing double-escape is safe because `EscapeNamespace` bypasses already-`@`-escaped segments. Verified `SanitizeGeneratedTypeSegments` + its three char-class helpers, the `!` short-circuit, modreq/modopt strip order, `.Trim()`, alias, keyword-escape order all identical; `StripArity` identical; repointed tests assert the same strings (no weakened assertions). Ran an **A/B**: built base `9204cfd2~1` in a separate worktree and PR head; both passed identical Decompiler test counts (2490) with 0 failures — byte-identity confirmed.

**Divergence:** none. Both reviewers independently confirmed byte-identical composition output and equivalent type-display normalization. No findings to action; no resolution commits required.

Local gates (PR head): slnx build clean; CSharp.Tests 250 (+10); RTS oracle 217; Decompiler fast subset 2496 (printer byte-identity); RTS parity corpus gate exit 0, 7 committed burn-down rows unchanged (0 new regressions).

richlander:
**Ready to merge.** Two CLEAN adversarial reviews (GPT-5.5 + Gemini 3.1 Pro) on the final head `9204cfd2` with no findings, all local gates green, base current with origin/main. Awaiting @richlander approval — not self-merging.

richlander:
### Reshaped: printer-owned composition + derived namespace scoping

Per review discussion, this PR was reworked from the original byte-preserving
carve into a design change: composition now lives on `CSharpTypePrinter` via
`CSharpTypePrintOptions`/`.Source`, `CSharpCompilationUnitComposer` is deleted,
and the `CSharpNamespaceStyle` enum is replaced by a derived rule (single print
request → file-scoped namespace; multiple → block-scoped).

This is a **deliberate behavior change** (single-type RTS files now render
file-scoped) — no longer byte-preserving. It is validated by the **opcode-level**
RTS oracle + corpus parity gate, not a source byte-diff. Corpus gate is
unchanged: exit 0, 25 exact / 2 opcode diffs / 9 recompile failures over 36,
7 burn-down rows unchanged.

The prior two CLEAN reviews were at the superseded head `9204cfd2`. New head is
`ed1187ed` — re-review dispatched to two independent reviewers (GPT-5.5 +
Gemini 3.1 Pro). Not ready to merge until two CLEAN reviews on the new head.

richlander:
### Adversarial re-review at new head `ed1187ed` — two CLEAN

Both reviewers ran in isolated worktrees against the reshaped head and reported
**CLEAN** with independent verification.

**GPT-5.5** — CLEAN
- Build clean; `CSharp.Tests` 253/253; RTS oracle (Evidence 3 / FixtureCatalog 82 / Prototype 114) all pass.
- Corpus parity gate: 25 exact / 2 opcode diffs / 9 recompile failures over 36, 7 burn-down rows unchanged.
- Zero leftover `CSharpNamespaceStyle` / `CSharpCompilationUnitComposer` / `CSharpCompilationUnitSpec` references; clean `git status`.

**Gemini 3.1 Pro** — CLEAN
- Same build + test results (253; 3/82/114).
- Independently reproduced the corpus gate: 25 exact / 2 opcode diffs / 9 recompile failures / 0 context failures over 36; 7 burn-down rows unchanged.
- Verified `ComposeSource` reproduces the old composer byte-for-byte (pragma, `[assembly:]`/`[module:]`, using escape/dedup/ordinal-sort).
- Confirmed the `requestList.Length == 1` rule is safe on every edge: global namespace (no `namespace ;`), multi-request same/different namespace (falls back to block-scoped, avoids CS8954), and nested types.
- Option defaults omit pragma/attributes for product callers; RTS opts in. Zero leftover references.

**Agreement:** both reviewers converged — no divergence, no findings. The
single-type→file-scoped flip is confirmed IL-neutral (corpus Exact count
identical). No follow-up changes required.

richlander:
**Ready to merge.** Two CLEAN adversarial reviews (GPT-5.5 + Gemini 3.1 Pro) on the current head `ed1187ed`, rebased on latest `origin/main`. All gates green; corpus parity unchanged. Awaiting @richlander approval — will not self-merge.

richlander:
## Review (Claude Fable 5)

**Verdict: merge-ready — no blockers; three follow-ups that can ride along with the planned printer-derived-usings/short-names PR.** This slice also lands acceptance criterion 1 of #2782: `CompileBackCSharpNames` is deleted.

### Verified

- **The relocation is character-identical.** Diffed the moved `CleanTypeDisplay`/`SanitizeGeneratedTypeSegments`/char-predicate code against the deleted harness version — verbatim, now with product-side tests (the RTS catalog tests were repointed, not dropped).
- **The derived namespace rule is sound.** One request produces at most one namespace unit (nested types print inside the outer type), so file-scoped is always legal there; multiple requests may span namespaces, where only block scoping composes in one file. The single-type flip to file-scoped is source-only, and the evidence is the right kind: identical parity-gate run pre/post (25/2/9 over 36, 7 burn-down rows unchanged, exit 0). The 9 recompile failures vs 7 committed rows is consistent — the gate fires only on *new* Exact→RecompileFail transitions, and the extra rows were never compileback-Exact.
- Composition ordering (pragma → assembly attrs → module attrs → usings → units) matches the deleted `WriteCompilationUnit` exactly, including the escape→distinct→ordinal-order usings pipeline.
- Two-model adversarial review ran on both heads (original and reshaped), clean, isolated worktrees.

### Follow-ups (non-blocking)

1. **`.Source` is composed eagerly on every `PrintBatch`**, doubling string memory for product callers that only consume `Units`. Same shape as the #2812 finding just fixed there: make it lazy (`Lazy<string>` or a `ComposeSource()` method), or measure that product call sites don't care. Cheap to do while the API is a day old — and the printer-derived-usings work will touch this exact surface anyway.
2. **`CleanTypeDisplay`'s modreq stripping is not scoped** — `.Replace(")", "")` removes *every* closing paren, so a legitimately parenthesized display (tuple syntax `(int, string)`) would mangle to `(int, string`. Pre-existing harness behavior, so not a regression — but the move blesses it as "the authoritative type-display cleaner… consumers must not reimplement it," and the new tests are positive-only (the repo rule asks for close negatives on heuristics). Either scope the stripping to balanced `modreq(...)`/`modopt(...)` segments or soften the doc claim with a known-limitation note — plus one tuple-display negative test either way.
3. Nit: the planner still escapes/dedups/orders its usings before handing them to the printer, which re-does all three (idempotently, so harmless). Since this PR's thesis is that the printer owns composition, the planner-side pass can drop in the follow-up.

Issue #2806, opened 2026-07-17
Implement bounded metadata traversal


richlander:
## Goal

Implement the bounded metadata traversal contract proposed in #2805 and `docs/design/bounded-metadata-traversal.md`.

Metadata relationship, count, and text projection paths must terminate on malformed input without producing plausible partial identities or success-shaped empty results.

## Work

1. Add iterative, cycle-aware TypeDef, TypeRef, and ExportedType relationship results in MetadataPrimitives. Migrate the shared TypeResolver and MetadataReaderExtensions funnels so formerly fatal inputs become typed or catchable failures.
2. Migrate product consumers to result-returning APIs, including Metadata identities/declarations/API surface, Instructions spelling and diffs, Analysis and Decompiler type resolution, and implementation-diff visibility. Remove local depth-only and silent-truncation guards.
3. Bound metadata-derived item and text expansion, including generic arity, generic parameter names, type arguments, and similar projections.
4. Add a deterministic malformed-artifact campaign through public library, type/member, API diff, IL diff, and implementation-diff paths.

Each behavior PR should remain complete for one failure mechanism rather than absorbing adjacent findings during review.

## Evidence

- Process-isolated fixtures for self and multi-row TypeDef, TypeRef, and ExportedType cycles.
- Deep acyclic and malformed-handle fixtures with exact rejection reasons.
- Long-name, huge-arity, and generic-fanout amplification fixtures.
- A pinned real-artifact census before activating new node, item, text, or arity limits.
- Unchanged identities and output for completed traversals.
- Deterministic parallel behavior and NativeAOT compatibility.

## Relationship to existing work

This supersedes #2578. That issue assumes the unmerged #2574 Metadata graph changes landed and lists Instructions paths later fixed by #2581 and #2723. The remaining problem is broader: current Metadata/MetadataPrimitives relationship and projection paths plus residual product consumers need one result and budget design.

richlander:
PR #2805 now makes validation ownership explicit: do not duplicate structural checks SRM already performs safely. Implementation guards are limited to resource-safe admission before recursive/count-driven SRM paths, TypeSpec re-entry, relationship cycles, projection amplification, and product failure shaping. MetadataLoadContext and ILVerify remain candidates for separate optional analysis or differential-oracle evaluation, not dependencies of this substrate.

richlander:
[...]

PR #2812, opened 2026-07-17
Bound metadata relationship traversal

richlander:
## Conclusion

Bound TypeDef, TypeRef, and ExportedType relationship traversal before migrating the broader consumer graph. Cycles, malformed rows, and chains over 256 nodes now produce typed rejection; compatibility name APIs throw catchable `BadImageFormatException` instead of recursing or returning a plausible partial identity.

Advances #2806. Text/item amplification and row-level consumer degradation remain follow-up slices.

## Behavior

- adds iterative, identity-cycle-aware relationship walkers in `ILInspector.MetadataPrimitives`;
- returns outermost-to-leaf handle chains plus terminal scope/implementation;
- distinguishes `Cycle`, `NodeBudget`, and `MalformedMetadata`, with subject handle and consumed-node evidence;
- adds result-returning TypeResolver and MetadataReader extension funnels;
- retains string compatibility APIs as whole-operation throwing boundaries;
- routes string signature and attribute providers through the bounded TypeDef/TypeRef funnels;
- qualifies nested ExportedTypes through their implementation chain;
- preserves allocation-neutral terminal fast paths for both handle and value APIs.

No inspected assembly is loaded. The product path remains SRM-only, Roslyn-free, and NativeAOT-friendly.

## Relationship ceiling evidence

A file-based SRM census enumerated every TypeDef, TypeRef, and ExportedType handle, followed each relationship iteratively with an unbounded per-start visited set, and recorded the maximum completed chain.

Pinned inputs: .NET runtime and reference packs `11.0.0-preview.6.26359.118` for Microsoft.NETCore.App and Microsoft.AspNetCore.App.

| Input | Result |
| --- | ---: |
| Assemblies | 623 |
| Malformed images | 0 |
| Cycles | 0 |
| Maximum TypeDef chain | 5 |
| Maximum TypeRef chain | 3 |
| Maximum ExportedType chain | 3 |

Ordered DLL digest: `d6d2fef7d7ccdf240f308cbfddd90fa21fdeb82b55d89d0028ef37ffd87e04af`.

The fixed 256-node ceiling is more than 50 times the observed maximum and aligns with the existing TypeSpec re-entry ceiling.

## Performance and output compatibility

File-based probes measured 500,000 warmed calls resolving a top-level `System.String` TypeDef, with the handle/value fetched outside the measured loop.

| Compatibility path | `origin/main` | Regressed revision | Fixed head `4e4f727f` |
| --- | ---: | ---: | ---: |
| `TypeResolver.GetTypeNameFromDefinition(handle)` | 128 B/call | 824 B/call (`cf30947b`) | 128 B/call |
| `reader.GetFullTypeName(typeDef)` | 128 B/call | 160 B/call (`2e768abe`) | 128 B/call |

The first regression came from relationship-set/builder work on terminal handles. The second came from allocating a completed result wrapper in the widely used value/extension path. Terminal handle and value paths now bypass both costs; bounded machinery remains confined to nested relationships.

`library System.Runtime -S "Type Forwarders"` is byte-identical to `origin/main` in Markdown (63,138 bytes) and JSON (116,383 bytes).

## Adversarial fixtures

- self and two-row TypeDef, TypeRef, and ExportedType cycles run in child processes;
- exactly 256 nodes complete; node 257 rejects for both handle and value-without-handle APIs;
- malformed relationship handles reject without a partial chain;
- malformed terminal handles normalize to `BadImageFormatException`;
- valid nested handle and value paths preserve namespace and qualification behavior.

## Adversarial review

Both reviewers assessed exact head `4e4f727f` against `origin/main` in isolated worktrees after the two measured allocation fixes.

- **Claude Opus 4.8:** clean after independently measuring handle and value APIs against main, checking nested allocation scope, full-framework naming parity, malformed normalization, cycle/budget behavior, overload resolution, and SRM/NativeAOT constraints.
- **Gemini 3.1 Pro:** clean after finding the 32 B/call value-wrapper regression at `2e768abe`, verifying its fix at `4e4f727f`, and rechecking nested cycle/budget semantics, exception boundaries, and both compatibility API families.

## Validation

- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release` - 599 passed
- `dotnet run --project src/dotnet-inspect.Tests -c Release` - 1,912 passed, 10 skipped because ilasm/ildasm are unavailable
- `npx markdownlint-cli docs/design/bounded-metadata-traversal.md`

### DIFF
--- a/docs/design/bounded-metadata-traversal.md
+++ b/docs/design/bounded-metadata-traversal.md
@@ -309,23 +309,31 @@ separate work; adopting either is not a prerequisite for this safety substrate.
 ## Budget policy
 
 All limits live in one metadata-safety policy rather than in individual
-formatters. The implementation PR must pin a real-artifact census and record the
-observed maxima before activating new limits.
+formatters. An implementation PR must pin a real-artifact census and record the
+observed maxima before activating a new limit.
 
-Candidate starting ceilings are:
+Current and candidate ceilings are:
 
-| Dimension | Candidate ceiling | Rationale |
+| Dimension | Ceiling | Status and rationale |
 | --- | ---: | --- |
-| Relationship nodes | 256 | Matches existing guarded recursion ceilings while remaining far above normal nesting |
-| Projected characters | 8,192 | Bounds repeated long names and argument expansion without constraining ordinary identities |
-| Generic arity expansion | 1,024 | Prevents metadata names such as ``Type`2000000000`` from driving unbounded loops |
-| TypeSpec active bytes | 4,096 | Existing `TypeSpecGuard` policy |
-| Signature structural depth | 512 | Existing `SignatureBlobGuard` policy |
+| Relationship nodes | 256 | Active; matches existing guarded recursion ceilings and is far above the measured maximum |
+| Projected characters | 8,192 | Candidate; bounds repeated long names and argument expansion without constraining ordinary identities |
+| Generic arity expansion | 1,024 | Candidate; prevents metadata names such as ``Type`2000000000`` from driving unbounded loops |
+| TypeSpec active bytes | 4,096 | Active existing `TypeSpecGuard` policy |
+| Signature structural depth | 512 | Active existing `SignatureBlobGuard` policy |
 
-The relationship-node value aligns with existing guarded recursion. The text
-and arity values come from prior hardening experiments but are not active
-policy until the implementation PR records a pinned corpus's maximum observed
-identity length and generic arity, then demonstrates adequate margin.
+The relationship-node ceiling was activated after scanning the .NET 11
+preview 6 runtime and reference packs
+(`11.0.0-preview.6.26359.118`): 623 assemblies, no malformed images or cycles,
+and maximum chain lengths of 5 TypeDefs, 3 TypeRefs, and 3 ExportedTypes. The
+ordered input DLL digest was
+`d6d2fef7d7ccdf240f308cbfddd90fa21fdeb82b55d89d0028ef37ffd87e04af`.
+The 256-node ceiling leaves more than 50 times the observed TypeDef depth while
+aligning with the existing TypeSpec re-entry ceiling.
+
+The text and arity values come from prior hardening experiments but are not
+active policy until an implementation PR records a pinned corpus's maximum
+observed identity length and generic arity, then demonstrates adequate margin.
 
 Once activated, these are security ceilings, not user-tunable output limits. A
 future change must update the policy, census evidence, and adversarial fixtures
--- a/src/ILInspector.Metadata/MetadataReaderExtensions.cs
+++ b/src/ILInspector.Metadata/MetadataReaderExtensions.cs
@@ -18,7 +18,15 @@ public static class MetadataReaderExtensions
         /// Gets the fully qualified name (Namespace.Name) of a <see cref="TypeDefinition"/>.
         /// </summary>
         public string GetFullTypeName(TypeDefinition typeDef)
-            => TypeResolver.GetFullName(reader, typeDef);
+            => TypeResolver.ResolveFullName(reader, typeDef).GetValueOrThrow();
+
+        /// <summary>
+        /// Resolves the fully qualified name of a TypeDefinition through a
+        /// bounded declaring-type walk.
+        /// </summary>
+        public RelationshipTraversalResult<string> ResolveFullTypeName(
+            TypeDefinitionHandle handle)
+            => TypeResolver.ResolveTypeNameFromDefinition(reader, handle);
 
         /// <summary>
         /// Gets the fully qualified name of a <see cref="TypeReference"/>,
@@ -26,19 +34,29 @@ public static class MetadataReaderExtensions
         /// match <see cref="GetFullTypeName(TypeDefinition)"/>.
         /// </summary>
         public string GetFullTypeName(TypeReference typeRef)
-            => typeRef.ResolutionScope.Kind == HandleKind.TypeReference
-                ? $"{reader.GetFullTypeName(reader.GetTypeReference((TypeReferenceHandle)typeRef.ResolutionScope))}.{reader.GetString(typeRef.Name)}"
-                : TypeResolver.GetFullName(reader.GetString(typeRef.Namespace), reader.GetString(typeRef.Name));
+            => TypeResolver.ResolveFullName(reader, typeRef).GetValueOrThrow();
+
+        /// <summary>
+        /// Resolves the fully qualified name of a TypeReference through a
+        /// bounded resolution-scope walk.
+        /// </summary>
+        public RelationshipTraversalResult<string> ResolveFullTypeName(
+            TypeReferenceHandle handle)
+            => TypeResolver.ResolveTypeNameFromReference(reader, handle);
 
         /// <summary>
         /// Gets the fully qualified name (Namespace.Name) of an <see cref="ExportedType"/>.
         /// </summary>
         public string GetFullTypeName(ExportedType exportedType)
-        {
-            string ns = reader.GetString(exportedType.Namespace);
-            string name = reader.GetString(exportedType.Name);
-            return TypeResolver.GetFullName(ns, name);
-        }
+            => TypeResolver.ResolveFullName(reader, exportedType).GetValueOrThrow();
+
+        /// <summary>
+        /// Resolves the fully qualified name of an ExportedType through a
+        /// bounded implementation walk.
+        /// </summary>
+        public RelationshipTraversalResult<string> ResolveFullTypeName(
+            ExportedTypeHandle handle)
+            => TypeResolver.ResolveTypeNameFromExportedType(reader, handle);
 
         /// <summary>
         /// Gets the name of a generic parameter from its handle.
--- a/src/ILInspector.Metadata/MetadataReaderExtensions.cs
+++ b/src/ILInspector.Metadata/MetadataReaderExtensions.cs
@@ -18,7 +18,7 @@ public static class MetadataReaderExtensions
         /// Gets the fully qualified name (Namespace.Name) of a <see cref="TypeDefinition"/>.
         /// </summary>
         public string GetFullTypeName(TypeDefinition typeDef)
-            => TypeResolver.ResolveFullName(reader, typeDef).GetValueOrThrow();
+            => TypeResolver.GetFullName(reader, typeDef);
 
         /// <summary>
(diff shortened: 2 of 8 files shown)


richlander:
Adversarial review reconciliation for exact head `cf30947b`: Claude Opus 4.8 and Gemini 3.1 Pro both reported clean reviews. Claude independently compared 28,326 runtime TypeDef/TypeRef names against the prior recursive semantics with zero mismatches; Gemini independently confirmed cycle precedence, exact 256/257 budget behavior, malformed-row rejection, TypeSpec interaction, and no partial identity path. The reviewers did not produce actionable or dismissed findings, so no review-resolution commit applies.

richlander:
Ready to merge. Local product/metadata validation, the two clean exact-head adversarial reviews, and all required CI checks are complete.

richlander:
## Review (Claude Fable 5)

**Verdict: no merge blockers; one item I'd act on before merge (finding 1 — fast path or a measurement note).**

### Verified

- **Budget accounting is consistent between the handle-based and value-based APIs.** Handle walk: node 256 completes, node 257 rejects. Value-based `ResolveFullName` resolves the declaring chain then `AppendLeaf` rejects once the chain has consumed 256 — total 257 rejects, 256 completes, identical semantics. `DeepAcyclicTypeReferenceChain` pins both at the exact boundary.
- **Cycle detection is complete for the value-based APIs** even though they don't know their own handle: any cycle is reachable from the declaring/scope walk (self-loop caught at `consumedNodes: 1`, two-row loop at `2`), and both fixture shapes are tested.
- **Naming semantics are preserved**: `FormatChain` uses only the outermost node's namespace and dot-joins nested names, matching the removed recursion for TypeDefs and TypeRef scope chains (compared against the old code; the Opus A/B over 28,326 runtime names corroborates).
- The malformed-row test's `consumedNodes: 2` is right: the phantom scope handle is consumed (added to the chain) before `GetTypeReference` throws on the nonexistent row.
- Child-process isolation for the cycle tests is a good defensive choice: if the code ever regresses to unbounded recursion, the stack overflow kills the worker, not the test runner.

### Findings

1. **Constant allocation overhead on hot metadata paths.** `SignatureDecoder.GetTypeFromDefinition` runs per type in every decoded signature, and the handle-based funnel now allocates a `HashSet<EntityHandle>`, two `ImmutableArray` builders, and a `StringBuilder` even for the dominant case (top-level type, chain length 1); the old code recursed only when nested. The value-based `ResolveFullName` already has a no-walk fast path for `declaringType.IsNil`; the handle-based walkers deserve the same short-circuit (peek `next(start)`; if terminal, return the single-node chain without the set/builders). I'd want that — or a measurement showing it doesn't move the needle — before this sits under a corpus sweep.
2. **Behavior change: nested ExportedTypes are now chain-qualified.** `GetFullTypeName(ExportedType)` previously returned bare `ns.name`; it now walks the implementation chain (`N.Outer.Middle.Leaf`). Right convention, and the PR states it — but it changes rendered output wherever forwarded/exported types appear. Worth confirming the affected output modes were exercised per the output-change evidence rule; the validation list has test suites but no output-mode check for this.
3. **Minor semantic wrinkle:** `GetValueOrThrow` maps `NodeBudget` to `BadImageFormatException`. A 257-deep chain is policy-rejected, not malformed; with a 50× margin over observed maxima this is academic, and one exception type at the compat boundary is simpler — no change requested, just noting the conflation is deliberate.
4. Nit: worker facts run as silent no-ops in the normal pass (`IsSelectedWorker` false → green), inflating the pass count by two. Standard pattern for process isolation; fine.

### Quality and process

Excellent evidence discipline: pinned census with input digest, exact-boundary fixtures, malformed-row fixtures, both-API assertions per shape, and the design doc moved from "candidate" to "active" ceiling with the measurement recorded. Layering respected (`MetadataPrimitives` self-contained, SRM-only, no assembly loading). Two-model adversarial review ran clean on the exact head in isolated worktrees per AGENTS.md.

richlander:
Addressed the review in `c05c692c`.

1. **Top-level allocation overhead — fixed.** A 500,000-iteration file-based probe resolving `System.String` measured `origin/main` at 128 B/call, the initial PR head at 824 B/call, and the fixed head at 128 B/call. The compatibility path now returns terminal TypeDef/TypeRef/ExportedType names directly; the neutral walker also avoids its set/builders for a one-node chain.
2. **Nested ExportedType output — verified, no product-output change.** The two product consumers first require `ExportedType.IsForwarder`, which restricts them to direct AssemblyReference implementations; those terminal names are unchanged. `library System.Runtime -S "Type Forwarders"` is byte-identical to `origin/main` in both Markdown (63,138 bytes) and JSON (116,383 bytes). The nested-chain behavior remains pinned at the extension/substrate seam.
3. **NodeBudget exception mapping — deliberate.** Compatibility APIs retain one catchable `BadImageFormatException` boundary; result-returning APIs preserve `NodeBudget` as typed evidence.
4. **Worker no-op facts — no action.** This follows the existing child-process safety-test pattern.

richlander:
Fixed-head adversarial review reconciliation for `c05c692c`:

- **Claude Opus 4.8:** clean after tracing the terminal fast paths, cycle/budget edges, malformed rows, naming equivalence, TypeSpec interaction, allocation behavior, and SRM/NativeAOT constraints.
- **Gemini 3.1 Pro:** clean after reconciliation of the compatibility boundary: result APIs retain typed rejection, while compatibility APIs preserve the pre-PR catchable `BadImageFormatException` contract. The bounded repeated reads on nested slow paths were not a correctness, security, or compatibility finding.

Review feedback action: `c05c692c` restores top-level allocation parity with main (128 B/call versus 824 B/call on the initial head). Non-actions: no output change exists in current Type Forwarders consumers, NodeBudget-to-`BadImageFormatException` remains deliberate at compatibility boundaries, and child-worker no-op facts retain the established process-isolation pattern.

richlander:
Ready to merge. The allocation feedback is resolved at `c05c692c`, both fixed-head adversarial reviews are clean, and the refreshed CI checks are complete.

richlander:

The `c05c692c` fast path resolves my finding 1 correctly — **merge-ready from my side**, one non-blocking nit below.

Verified:

- **Cycle detection survives the fast path.** A self-loop is not terminal (`next(start) == start`, same kind), so it falls into the full walk, which seeds `visited = {start}` / `leafToRoot = [firstHandle]` — the self-loop `consumedNodes: 1` and two-row-cycle `consumedNodes: 2` semantics are preserved. Retraced the phantom-row malformed test: the fast path reads only row 1's scope before deferring to the full walk, so `consumedNodes: 2` still holds.
- **The allocation claim matches the code**: single-node completions build only the one-element chain (no HashSet/builders), and the compat string APIs skip the result machinery for terminal types — consistent with the 128 → 824 → 128 B/call probe.
- The new test pins the compat-API fast path for ExportedType.

Non-blocking nit: the compat APIs' leaf-row read (`reader.GetTypeReference(handle)` etc.) sits outside any try, so an `ArgumentOutOfRangeException` from SRM on the leaf row itself would propagate raw rather than as `BadImageFormatException` — slightly narrower than the PR body's "throw catchable BadImageFormatException" claim. Identical to pre-PR main behavior (no regression), and deeper-chain reads still funnel through the typed rejection. A one-line catch-and-wrap in the three compat APIs would make the claim exact, at your option.

richlander:
Addressed the non-blocking leaf-row exception nit in `2e768abe`. The three terminal compatibility fast paths now normalize only SRM `BadImageFormatException`/`ArgumentOutOfRangeException` failures through the typed malformed-metadata boundary, while nested paths remain unchanged and are not double-wrapped. Added direct malformed-leaf tests for TypeDef, TypeRef, and ExportedType. Full solution build and all 599 metadata tests pass; the warmed 500,000-call allocation probe remains 128 B/call. Fresh exact-head adversarial reviews and CI are now pending, so this supersedes the earlier readiness comment until those gates are clean.

richlander:
Gemini’s exact-head review found a second measurable allocation issue at `2e768abe`: the widely used value/extension path `reader.GetFullTypeName(TypeDefinition)` allocated a completed result wrapper, measuring 160 B/call versus 128 B/call on `origin/main`. Reproduced locally over 500,000 warmed calls and fixed in `4e4f727f` by restoring terminal fast paths for TypeDefinition, TypeReference, and ExportedType values and routing `MetadataReaderExtensions` through them. The same probe is back to 128 B/call. Nested compatibility behavior is now asserted for all three relationship kinds. Fresh exact-head reviews and CI are pending.

richlander:
Final adversarial review reconciliation for exact head `4e4f727f`:

- **Claude Opus 4.8:** clean after independently measuring both handle and value/extension compatibility paths against `origin/main`, checking full-framework naming parity, nested cycle/budget behavior, malformed normalization, overload resolution, and SRM/NativeAOT constraints.
- **Gemini 3.1 Pro:** found a real 32 B/call regression at `2e768abe` in `reader.GetFullTypeName(typeDef)` (160 versus 128 B/call on main). `4e4f727f` restores allocation-free terminal value paths for TypeDef, TypeRef, and ExportedType; Gemini re-measured the fixed head and reported it clean.

Resolution commits:
- `2e768abe` normalizes malformed terminal handle reads to the documented `BadImageFormatException` compatibility boundary.
- `4e4f727f` restores value/extension allocation parity and adds nested compatibility assertions for all three relationship kinds.

The warmed 500,000-call probes now measure 128 B/call for both the handle and value compatibility paths. No actionable findings remain. The final CI test job is still running; readiness will be posted after it completes.

richlander:
Ready to merge. Exact head `4e4f727f` has two clean fixed-head adversarial reviews, both measured compatibility paths are back at 128 B/call, local build and suites are clean, and all required CI checks have completed successfully.

richlander:
## Re-review sign-off #2 (Claude Fable 5)

**Clean — all findings closed, merge-ready.**

- `2e768abe` closes my non-blocking nit exactly: the three compat handle APIs now normalize leaf-row `BadImageFormatException`/`ArgumentOutOfRangeException` through the standard wrapped message, making the PR's contract claim exact, with a phantom-row test asserting the message prefix on all three shapes.
- `4e4f727f` is correct as well as faster: the value-based terminal fast path is allocation-free, and the nested path via `AppendLeaf` is semantically identical to the removed `ResolveFullName` route (the walkers and `AppendLeaf` already catch internally, so nothing escapes unwrapped). The new `GetFullName(reader, TypeReference/ExportedType)` overloads match the extension methods' chain-qualified behavior, and the chain tests now pin the value path.

Nothing further from me.

PR #2825, opened 2026-07-17
Printer-derived usings + short type names; #2811 follow-ups

richlander:
## Why

Follow-up to merged #2811. Bundles the merge reviewer's three non-blocking
follow-ups with a new short-type-names capability on the product seam
`CSharpTypePrinter` (epic #2777).

## What

**Review follow-ups (#2811)**

1. **Lazy `Source`** — `CSharpTypePrintResult.Source` is composed on first access
   (`Lazy<string>`), so callers that read only `Units` skip composition.
2. **Scoped custom-modifier strip** — `CleanTypeDisplay` replaced its
   `.Replace(")", "")` (which nuked every close paren, mangling tuple types)
   with `StripCustomModifiers`, a paren-depth-tracked removal of only
   `modreq(…)`/`modopt(…)` wrappers. Tuple syntax is preserved. +negative test.
3. **Drop planner using pre-pass** — `ReturnToSenderTypePlanner` no longer
   escapes/dedups/orders usings; the printer already normalizes them.

**Short type names (feature)** — default on via
`CSharpTypePrintOptions.ShortenTypeNames`.

- `CSharpDeclarationWriter.DeriveContextualUsings` computes a **collision-safe**
  namespace set: a namespace is imported only when every simple name it
  contributes is unambiguous across the unit **and** does not clash with a
  declared type. Ambiguous/self-shadowing names stay fully qualified.
- `CSharpTypePrinter` derives those usings (excluding the unit's own declaring
  namespaces), feeds them to the declaration formatter, and unions them into the
  composed source. References shorten to simple names; the enabling namespaces
  are emitted as `using` directives.

**Scope note:** `CSharpTypePrinter.PrintBatch` currently has **no CLI consumer** —
it is used only by the ReturnToSender compile-back oracle and tests. So today the
short names change RTS's composed compile-back source (corpus-proven IL-neutral),
not the CLI `Decompiled Source` view (which renders via
`Decompiler.Pipeline.CSharpPrinter.PrintRaised`). This lands the capability on the
product seam per #2777 as a step toward the CLI consuming the seam.

## Soundness bug found + fixed during validation

`DeriveContextualUsings` initially mis-parsed an attribute argument enum access
(`MarshalAs(UnmanagedType.I4)`) as a type reference, emitting an **illegal**
`using System.Runtime.InteropServices.UnmanagedType;` (a type, not a namespace)
and mis-shortening the argument to `I4`. Fix: `StripAttributeArguments` — attribute
constructor arguments are value expressions (enum/const/typeof), not type
references, so only the attribute type name participates in reference/using
derivation. Post-fix output: `using System.Runtime.InteropServices;` +
`MarshalAs(System.Runtime.InteropServices.UnmanagedType.I4)` (attribute type
shortened, argument left qualified, valid, recompiles Exact). Covered by a unit
regression test and RTS prototype assertions.

## Evidence

- `ILInspector.CSharp.Tests` **262** (+5 feature tests: cross-namespace shorten,
  `ShortenTypeNames=false`, collision-stays-qualified, self-shadow-stays-qualified,
  attribute-enum-arg regression; +2 tuple/modifier tests).
- RTS oracle **199** (8 prototype tests updated to the shortened forms, from
  captured ground-truth output).
- Decompiler fast subset **2496**.
- **RTS parity corpus gate: 25 exact / 2 opcode diffs / 9 recompile failures over
  36, 7 burn-down rows unchanged, exit 0** — identical to baseline. Short-name RTS
  source is IL-neutral, including through full method bodies (this is an
  opcode-level oracle, not a byte diff).

## Review

Behavior-changing (RTS composed source). Requesting two adversarial reviews
(GPT-5.5 + Gemini 3.1 Pro). Attack points: `DeriveContextualUsings`
collision/self-shadow/attribute-argument soundness; lazy `Source` correctness;
`StripCustomModifiers` equivalence + tuple preservation; corpus Exact unchanged.

### DIFF
--- a/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpFormatterTests.cs
@@ -384,6 +384,18 @@ public sealed class CSharpFormatterTests
     public void CleanTypeDisplay_NormalizesToCSharpSpelling(string type, string expected)
         => Assert.Equal(expected, CSharpFormatter.CleanTypeDisplay(type));
 
+    [Theory]
+    [InlineData("int modreq(System.Runtime.CompilerServices.IsVolatile)", "int System.Runtime.CompilerServices.IsVolatile")]
+    [InlineData("System.Int32 modopt(Mod)", "int Mod")]
+    public void CleanTypeDisplay_StripsCustomModifierWrappers(string type, string expected)
+        => Assert.Equal(expected, CSharpFormatter.CleanTypeDisplay(type));
+
+    [Theory]
+    [InlineData("(int, string)", "(int, string)")]
+    [InlineData("System.ValueTuple<(int, string), object>", "System.ValueTuple<(int, string), object>")]
+    public void CleanTypeDisplay_PreservesUnrelatedParentheses(string type, string expected)
+        => Assert.Equal(expected, CSharpFormatter.CleanTypeDisplay(type));
+
     [Fact]
     public void CleanTypeDisplay_CollapsesUnspeakableGenericParameterToObject()
         => Assert.Equal("object", CSharpFormatter.CleanTypeDisplay("!0"));
--- a/src/ILInspector.CSharp.Tests/CSharpTypePrinterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpTypePrinterTests.cs
@@ -887,6 +887,32 @@ public sealed class CSharpTypePrinterTests
     }
 
     [Fact]
+    public void NestedTypeReferencedAsNamespaceIsNotImportedWhenEnclosingTypeIsReferenced()
+    {
+        // `System.Environment.SpecialFolder` is a nested type but arrives as a flat
+        // dotted string, so the last-dot split derives namespace `System.Environment`
+        // — which is actually a type. Emitting `using System.Environment;` is illegal.
+        // When the enclosing type `System.Environment` is itself referenced in the
+        // unit, its full name shows up as a derived namespace and must be excluded, so
+        // the nested reference stays fully qualified.
+        var type = CreateEmptyType("App", "Consumer");
+        var enclosing = CreateMethod("GetEnv");
+        enclosing.SignatureModel!.ReturnType = "System.Environment";
+        var nested = CreateMethod("GetFolder");
+        nested.SignatureModel!.ReturnType = "System.Environment.SpecialFolder";
+        type.Members.Add(enclosing);
+        type.Members.Add(nested);
+
+        var result = _printer.Print(new CSharpTypePrintRequest(type));
+
+        Assert.DoesNotContain("using System.Environment;", result.Source, StringComparison.Ordinal);
+        Assert.Contains(
+            "System.Environment.SpecialFolder GetFolder();",
+            result.Units[0].Source,
+            StringComparison.Ordinal);
+    }
+
+    [Fact]
     public void UsingsSuppressedKeepsReferencesQualified()
     {
         // With IncludeUsings=false the composed Source omits using directives, so
--- a/src/ILInspector.CSharp.Tests/CSharpTypePrinterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpTypePrinterTests.cs
@@ -758,6 +758,31 @@ public sealed class CSharpTypePrinterTests
     }
 
     [Fact]
+    public void RawSignatureTupleReturnMethodTypeParameterStaysQualified()
+    {
+        // A tuple return type puts a '(' before the parameter list, so the raw-signature
+        // parser must anchor on the method name + generic list, not the first '('. The
+        // method type parameter `Task` still shadows the same-named references.
+        var type = CreateEmptyType("Samples", "Worker");
+        type.IsAbstract = true;
+        type.Members.Add(new ApiMember
+        {
+            Name = "Run",
+            Kind = "method",
+            IsAbstract = true,
+            Signature = "(System.Threading.Tasks.Task, int) Run<Task>(System.Threading.Tasks.Task value)"
+        });
+
+        var result = _printer.Print(new CSharpTypePrintRequest(type));
+
+        Assert.DoesNotContain("using System.Threading.Tasks;", result.Source, StringComparison.Ordinal);
+        Assert.DoesNotContain(
+            "(Task, int)",
+            result.Units[0].Source,
+            StringComparison.Ordinal);
+    }
+
+    [Fact]
     public void ReferenceCollidingWithDeclaredTypeStaysQualified()
     {
         // A type declared as `Task` referencing `System.Threading.Tasks.Task` must not
--- a/src/ILInspector.CSharp.Tests/CSharpTypePrinterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpTypePrinterTests.cs
@@ -388,7 +388,7 @@ public sealed class CSharpTypePrinterTests
         var result = _printer.Print(new CSharpTypePrintRequest(type));
 
         Assert.Contains(
-            "public long GetTicks([System.Runtime.InteropServices.Optional, System.Runtime.CompilerServices.DateTimeConstant(0)] System.DateTime when);",
+            "public long GetTicks([Optional, DateTimeConstant(0)] DateTime when);",
             result.Units[0].Source,
             StringComparison.Ordinal);
     }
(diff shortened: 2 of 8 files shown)


richlander:
## Adversarial review — round 1 (heads: `3f906311`), fixes at `ba09f746`

Two independent adversarial reviews in isolated worktrees (GPT-5.5, Gemini 3.1 Pro; I am Opus 4.8, so both reviewers are other families). Three real soundness gaps found and fixed; one dismissed with a real-printer repro. Re-review of the fixed head is in progress.

### GPT-5.5 (derivation stage)
1. **Generic type-parameter shadowing** (high) — a reference whose simple name matches an in-scope type/method type parameter (e.g. `Box<Task>` referencing `System.Threading.Tasks.Task`) was shortened to `Task`, binding to the parameter. **Fixed** (`ba09f746`): `DeriveContextualUsings` now adds all in-scope type-parameter names to the shadow set.
2. **`IncludeUsings=false` inconsistency** (medium) — shortening still ran while `ComposeSource` suppressed the `using` directives, emitting simple names with no import. Latent (default `true`, no product caller sets false) but a real seam-contract hole. **Fixed**: `ComputeDerivedUsings` returns `[]` when `!IncludeUsings`.

### Gemini 3.1 Pro (replacement stage)
1. **Ambiguous shortening** (critical) — **confirmed real via a cross-type scope mismatch** (Gemini's literal single-type repro did not reproduce, because `TypeNamePlan.Create` already skips within-type collisions). Root cause: `DeriveContextualUsings` detected the collision unit-wide but still imported the namespace via other unambiguous names, and per-type shortening then shortened the ambiguous reference → CS0104. **Fixed**: `DeriveContextualUsings` now excludes a namespace entirely when any simple name it contributes is ambiguous unit-wide or shadowed (matches its documented contract). Verified through the real printer: cross-type `System.String`/`MyNamespace.String` both stay qualified.
2. **`@`-escaped identifier corruption** (high) — **dismissed.** Running the keyword-namespace scenario through the real `CSharpTypePrinter` produces valid C# (`using Samples.@event;` + fully-qualified `Samples.@event.Widget`); the reference simply stays qualified (escaped display text doesn't match the raw replacement key) and the using is escaped correctly. There is no `using @event;` and no `System.@Foo` mangling — that output was simulated, not produced by the printer. Only artifact is a harmless redundant (unused) using. Asked the reviewer for a concrete invalid-output repro if they disagree.

### Agreement / divergence
The two reviewers attacked **different stages** (GPT-5.5 the using *derivation*, Gemini the shortening *replacement*) and found non-overlapping issues — no direct agreement or contradiction. All three actionable findings are fixed in one commit with four new regression tests.

### Evidence (fixed head `ba09f746`)
- `ILInspector.CSharp.Tests` **266** (+4: cross-type ambiguity, generic type-param shadow, method type-param shadow, `IncludeUsings=false`).
- RTS oracle **199**.
- RTS parity corpus gate: **25 exact / 2 opcode / 9 recompile-fail over 36, 7 burn-down rows unchanged, exit 0** — identical to baseline. All fixes IL-neutral.

Re-review dispatched to both reviewers at `ba09f746`. Not ready to merge until two CLEAN reviews on the fixed head.

richlander:
## Adversarial review — final reconciliation (two CLEAN at `38c6100c`)

Reviewers: **GPT-5.5** and **Gemini 3.1 Pro**, each in an isolated worktree at the PR head. I am Opus 4.8, so both reviewers are from other model families. Four review rounds; every fix re-reviewed on its fixed head.

### Findings (4 fixed, 1 dismissed)

| # | Reviewer | Severity | Finding | Resolution |
|---|----------|----------|---------|------------|
| 1 | GPT-5.5 | High | Generic **type-parameter shadowing**: type params named like a referenced type (e.g. `Box<Task>` referencing `System.Threading.Tasks.Task`) were absent from the shadow set → reference shortened and rebound to the parameter. | Fixed `ba09f746` — type/method type-param names added to `declaredSimpleNames`. |
| 2 | GPT-5.5 | Medium | **`IncludeUsings=false`**: shortening still ran while usings were suppressed → uncompilable `Source`. | Fixed `ba09f746` — `ComputeDerivedUsings` returns `[]` when `!IncludeUsings`. |
| 3 | Gemini | Critical | **Ambiguous shortening via cross-type scope mismatch**: `DeriveContextualUsings` detected a unit-wide collision but still imported the namespace via other unambiguous names, and per-type shortening then shortened the ambiguous name → CS0104. (Gemini's literal single-type repro did not reproduce — `TypeNamePlan.Create` already skips within-type collisions — but the cross-type variant is real; proved via file-based-app probe.) | Fixed `ba09f746` — `unsafeNamespaces` excludes a whole namespace if any simple name it contributes is ambiguous unit-wide or shadowed. |
| 4 | Gemini | High | **`@`-escaped identifier corruption** (e.g. `Samples.@event.Widget`). | **Dismissed** — real-printer probe produced valid output (`using Samples.@event;` + fully-qualified reference); the escape/raw-key mismatch protects the pipeline. Gemini re-traced and agreed 100%. Cross-type regression test added. |
| 5 | GPT-5.5 | High | **Raw-signature fallback gap**: members whose signature failed structured decoding fall back to raw `ApiMember.Signature`, whose generic method params are absent from `SignatureModel` → return type could shorten to a method type-param name (CS0508). | Fixed `6b8be4dc` — new `RawSignatureGenericParameterNames` parser feeds raw-signature generic params into the shadow set. |
| 6 | GPT-5.5 | High | **Tuple-return parser edge**: `RawSignatureGenericParameterNames` anchored on the first `(`, which for a tuple return `(Task, int) Run<Task>(...)` is the tuple, not the parameter list → method type-param missed. | Fixed `38c6100c` — parser now anchors on the method-name token immediately followed by `<...>(`, scanning past name occurrences in the return type. |

### Agreement / divergence
- Both reviewers independently converged on the **generic-parameter shadow set** as the core soundness surface (GPT via type-params, raw-sig, and tuple; Gemini via an independent adversarial audit of the same helper).
- Divergence on Finding 4: Gemini initially flagged it, then re-traced and confirmed the dismissal is correct (the raw metadata key never matches the escaped display text, so no shortening fires and the reference stays qualified).
- No unresolved disagreements at the final head.

### Final head `38c6100c` — both CLEAN
- **Gemini 3.1 Pro: CLEAN** — audited the rewritten parser (boundary checks, depth tracking, trailing-`(` validation, safe `continue` on non-match); ran full non-slow `ILInspector.CSharp.Tests` (269) + RTS (87), no regressions.
- **GPT-5.5: CLEAN** — attacked the parser with function-pointer/array/pointer/ref-return/operator/whitespace shapes, found no invalid-output repro; `ILInspector.CSharp.Tests` 268, RTS fast 82.

### Validation at `38c6100c`
- `ILInspector.CSharp.Tests`: 268 (my run) / 269 (Gemini non-slow) — green.
- RTS oracle (`ReturnToSenderPrototypeTests`/`EvidenceTests`/`FixtureCatalogTests`): 199 — green.
- Corpus RTS-parity gate: **25 exact / 2 opcode / 9 recompile-fail, 7 burn-down unchanged, exit 0** — held at every fixed head (empirical soundness proof that shortened source recompiles to identical IL).

Commit chain: `3f906311` (base) → `ba09f746` (findings 1–4) → `6b8be4dc` (finding 5) → `38c6100c` (finding 6).

**Ready to merge** pending @richlander approval — I will not self-merge.

richlander:
## Review (Claude Fable 5)

**Verdict: no blockers for the current RTS-only scope. All three #2811 follow-ups land correctly. One real latent bug in the usings derivation (nested-type-as-namespace — the general case of the class this PR's own attribute fix closed), one API-shape hazard for the future CLI consumer, two nits.**

### Verified

- **`StripCustomModifiers` is correct.** Traced the depth tracking through nested modifiers, modreq-inside-tuple (`(int modreq(X), string)` → `(int X, string)`), and unbalanced input — only wrapper parens are swallowed. The modreq output text (`int System.Runtime.CompilerServices.IsVolatile`) preserves the old code's semantics exactly; only the tuple-mangling is fixed, which is precisely the claimed equivalence.
- **Lazy `Source` is right**: callers reading only `Units` skip composition; the closure captures the same unit list and options the eager path used.
- **Planner pre-pass drop is safe** — the printer's `ComposeSource` normalizes (escape/dedup/order) exactly what the planner used to.
- **The derivation/rendering safety invariant holds structurally, not by luck**: both sides key off the same collectors, and the formatter's `TypeNamePlan` shortens a reference only when its namespace is imported (or same-namespace) — so `DeriveContextualUsings`' unit-wide exclusion of unsafe namespaces controls per-type shortening. The negative-test suite (cross-type ambiguity, declared-type shadow, generic/method type-parameter shadow, raw-signature shadow with the tuple-anchored parser, attribute enum args, `IncludeUsings=false` interplay) is genuinely strong.
- Corpus parity identical (25/2/9, 7 rows, exit 0); prototype expectations updated from captured ground truth, all still Exact.

### Finding 1 — nested types referenced as types still derive illegal usings (latent)

`TypeRef.TryCreate` (CSharpDeclarationWriter.cs:1454) splits at the **last dot** and cannot distinguish a namespace from a declaring-type chain. A nested type referenced as a type in a signature — e.g. a parameter of type `System.Environment.SpecialFolder` — derives namespace `System.Environment`; when `SpecialFolder` is unit-unique, the derivation emits the illegal `using System.Environment;`. This is the same bug class `StripAttributeArguments` fixed for attribute arguments, but the general signature-position case remains. Failure mode is safe-and-visible for RTS (Roslyn rejects the using → `RecompileFail`), and the 36-method parity sample didn't hit it, but a wider corpus will. Cheap partial guard: exclude any derived namespace that itself appears as an extracted type `FullName` in the unit; fuller fix: derive from typed identity (metadata knows nested-ness) instead of string splitting. Either way, add a `SpecialFolder`-shaped regression test. Fine as an immediate follow-up given the fail-visible mode — but it belongs on the adversarial reviewers' attack list for this head.

### Finding 2 — `Units` are no longer self-contained under the default

With `ShortenTypeNames = true` (the default), `Unit.Source` contains short names that resolve only against the composed `Source`'s using block — while the contract doc still says per-type `Units` never carry usings. Harmless for RTS (it consumes `Source`), but the scope note's stated goal is the CLI consuming this seam, and the CLI's `Decompiled Source` view is per-type. Either document `Units` as file-context-relative under shortening, or carry each unit's enabling usings on `CSharpTypeSourceUnit` so per-unit consumers can compose validly.

### Nits

- `CSharpTypePrintResult` record equality now includes the `Lazy<string>` field, so two content-identical results compare unequal. Fine if equality is unused — worth knowing it changed.
- Stray whitespace in `ReturnToSenderPrototypeTests.cs`: `[                new ReturnToSender.RequestedTarget(...)`.

richlander:
## Review follow-up — Claude Fable 5 (fixes at `1dd6d1ab`)

Thanks for the review. No blockers; addressed both findings and both nits. Summary:

### Finding 1 — nested type referenced as a type derives an illegal `using` (latent) — **fixed (partial) + documented**

Confirmed real via a real-printer probe: an isolated `System.Environment.SpecialFolder` return type produced `using System.Environment;` (illegal, CS0138) and shortened to `SpecialFolder`.

I traced the data path to bound the fix correctly: nested types are **flattened to dots before the seam** — `ApiMemberIdentity` does `.Replace('+','.')` and `TypeNodeProvider`/`TypeResolver.FormatDisplayName` render `Dictionary\`2.Enumerator` → `Dictionary<…>.Enumerator`. So the seam receives a flat dotted string with no nested-ness signal; the "typed identity" fuller fix genuinely lives in the **metadata layer** (which owns type naming), out of scope for this PR.

Implemented your cheap guard: `DeriveContextualUsings` now excludes any derived namespace that itself appears as a referenced type full name in the unit (e.g. when `System.Environment` is referenced directly, `using System.Environment;` is suppressed and the nested reference stays qualified). Because `ContextualShort` only shortens references whose namespace is imported, this one guard suppresses **both** the bad `using` and the bad shortening.

The isolated case (enclosing type never referenced) remains undetectable from the flattened string and is documented in code as a known limitation; failure mode is safe-visible (RTS records `RecompileFail`, no miscompile). Added regression test `NestedTypeReferencedAsNamespaceIsNotImportedWhenEnclosingTypeIsReferenced`.

### Finding 2 — `Units` not self-contained under the default — **documented**

Documented on `CSharpTypePrintOptions.Usings` and `CSharpTypeSourceUnit` that under `ShortenTypeNames` a unit's `Source` is file-context-relative: shortened names resolve only against the composed `Source` using block. A per-unit consumer must compose against `Source` or set `ShortenTypeNames=false`. (RTS consumes `Source`, so it's unaffected; this is guidance for the future per-type CLI consumer.)

### Nits — both fixed

- `CSharpTypePrintResult` now defines explicit value equality over `Units`/`Diagnostics`, excluding the lazy source field — equality no longer forces composition or degrades to `Lazy` reference identity. (Verified equality is currently unused, so this is forward-safety for the CLI consumer.)
- Removed the stray whitespace in `ReturnToSenderPrototypeTests`.

### Validation at `1dd6d1ab`
- `ILInspector.CSharp.Tests`: **269** (+1 regression test).
- RTS oracle: **199**.
- Corpus rts-parity gate: **25 exact / 2 opcode / 9 recompile-fail over 36, 7 burn-down unchanged, exit 0** — identical to baseline (IL-neutral).

Re-confirming with the two adversarial reviewers on this head.

richlander:
## Two CLEAN reviews on `1dd6d1ab` — ready pending approval

The Claude Fable 5 follow-up commit (`1dd6d1ab`) was re-reviewed by both adversarial reviewers on this head:

- **GPT-5.5: CLEAN** — verified the new `referencedFullNames.Contains(ns)` guard cannot wrongly suppress a legitimate namespace import, and the `CSharpTypePrintResult.Equals/GetHashCode` override is safe. CSharp.Tests 269, RTS fast 82, no regression.
- **Gemini 3.1 Pro: CLEAN** — confirmed the guard forces the nested reference to stay fully qualified (avoids CS0138) without suppressing legitimate namespaces; probed namespace-masking corner cases (`A.B.C` + `A`, deep nesting) and re-ran the corpus parity gate to the exact baseline (25 exact / 2 opcode / 9 recompile-fail over 36, 7 burn-down). Equality override sound. CSharp.Tests 270 (non-slow), RTS 87.

### Full review ledger (5 findings fixed, 1 dismissed, across all rounds)
|---|----------|---------|------------|
| 1 | GPT-5.5 | Generic type-parameter shadowing | `ba09f746` |
| 2 | GPT-5.5 | `IncludeUsings=false` shortening | `ba09f746` |
| 3 | Gemini | Cross-type ambiguous shortening | `ba09f746` |
| 4 | Gemini | `@`-escaped identifier (dismissed, re-traced & agreed) | n/a |
| 5 | GPT-5.5 | Raw-signature generic-param shadow | `6b8be4dc` |
| 6 | GPT-5.5 | Tuple-return parser edge | `38c6100c` |
| 7 | Claude Fable 5 | Nested-type-as-namespace illegal `using` (partial guard + documented residual) | `1dd6d1ab` |
| — | Claude Fable 5 | `Units` file-context contract (doc), result equality + whitespace nits | `1dd6d1ab` |

Every fix was re-reviewed on its fixed head. Final head `1dd6d1ab`: CSharp.Tests 269, RTS oracle 199, corpus rts-parity 25/2/9 + 7 burn-down unchanged, exit 0.

**Ready to merge pending @richlander approval — I will not self-merge.**

richlander:

**`1dd6d1ab` closes everything from my review — merge-ready.**

- **Finding 1 (nested-type-as-namespace):** the `referencedFullNames.Contains(ns)` guard is the partial fix I suggested, and it's implemented safely — it can only suppress an import when a namespace string exactly equals a referenced type's full name, which is precisely the illegal-using case (and where the two collide legitimately, suppression is the conservative correct answer, as GPT-5.5 independently confirmed). The comment documenting the undetectable isolated case (enclosing type never referenced on its own), the deferred full fix via metadata nested-type identity, and the safe-visible RecompileFail failure mode is exactly the right honesty — plus the `SpecialFolder`-shaped regression test I asked for, validated against a real-printer probe rather than assumption.
- **Finding 2 (Units contract):** documented as file-context-relative with the explicit consumer guidance (compose against `Source` or set `ShortenTypeNames=false`). Right call for now; the future CLI-consumes-Units slice inherits a clearly stated contract instead of a trap.
- **Both nits fixed**, with the equality override going beyond the ask — excluding the lazy field so equality neither forces composition nor degrades to reference identity is the correct semantics, not just a patch.

Nothing further from me. With the four adversarial rounds' own soundness fixes plus this, the derivation has earned its default-on.

Issue #2835, opened 2026-07-17
implements Type column: consistent naming + disambiguate duplicate simple names


richlander:
## Problem

The implements command Type column mixes fully-qualified and simple names across rows, and can show duplicate simple names for distinct types. Output would be clearer with a consistent naming policy (and disambiguation when simple names collide).

## Scope

- Decide one consistent naming policy for the implements Type column (fully-qualified vs simple).
- Disambiguate duplicate simple names (e.g. show the namespace when two implementers share a simple name).

## Why deferred

This is a naming-consistency design decision distinct from the mechanical rendering fix already shipped (wrapping the Type column in a C#-friendly code span). Found during the usability pass.

Issue #2834, opened 2026-07-17
Decompiler/Signature: emit one attribute per line


richlander:
## Problem

When a member carries several long attributes, the decompiler/Signature output places them on a single line, producing very long, hard-to-scan lines. Idiomatic C# puts each attribute on its own line.

## Proposal

Emit one attribute per line in the decompiler printer for the Signature/decompile paths.

## Why deferred

This is decompiler-printer behavior and is adversarial-review-gated per AGENTS.md (requires A/B evidence through the C# printer + Roslyn oracle). It is out of scope for the mechanical CLI/output usability fixes in the originating usability-pass PR.

Related usability findings: multi-attribute-per-line (finding 9) and Signature rendered inside a Markdown table cell where real newlines can't be used (finding 2).

Issue #2833, opened 2026-07-17
Performance Triage --table: curated default column set


richlander:
## Problem

The Performance Triage section renders ~29 columns. In --table mode this is unusable at any reasonable terminal width. --tsv/--jsonl (machine formats) legitimately keep the full width, and --columns projection already works, but there is no ergonomic default for the human --table lens.

## Proposal

Give --table a curated default column set for Performance Triage (proposed: member, finding, fix, confidence, root_reach, shape, evidence), with an opt-in (e.g. --fields full or existing --columns) to recover the full width.

## Why deferred

MarkoutIgnoreColumnWhen is condition-on-data only and MarkoutIgnoreInTable excludes a whole section from a parent table, not a column. A curated default requires section-name-aware column selection in the render path — a design change beyond the mechanical usability fixes in the originating usability-pass PR.

Found during the usability pass.

PR #2836, opened 2026-07-17
Fix usability findings: section selection, empty-state, and type-name rendering

richlander:
## Summary

Fixes six usability findings from a full-command pass over `dotnet-inspect`. All are behavior-safe presentation/UX fixes. No change to identifiers, provenance, or the meaning of any emitted field.

## Fixes

- **Section selection (high):** An explicit `-S` no longer silently gains the Performance Triage section from `--top`/`--loop`/`--min-confidence`/etc. row filters. Previously `library X -S "Top Leverage" --top 1 --tsv` emitted two sections and broke single-section formats (`--table`/`--tsv`/`--jsonl`). Top Leverage and Performance Triage remain two independent, separately-selectable sections; the row filters still surface Performance Triage when the user did not pick sections with `-S`.
- **Unknown `-S` value (med):** A non-glob value with no close fuzzy match now lists all available sections under "Available sections:" and points to `-D`, instead of dead-ending.
- **Empty Decode column:** Member Index and Signature drop the always-empty `Decode` column when no member carries a signature-decode degradation marker, via `MarkoutIgnoreColumnWhen`. This drop applies across all formats (markdown/table/tsv/jsonl) when the column is uniformly empty — matching the existing `MarkoutIgnoreColumnWhen` convention already used for uniform `Kind` columns (e.g. the Integrations section already omits `kind` from jsonl when uniform). The column returns as soon as any member is `degraded`, so the decode-failure signal is never suppressed (covered by new predicate tests).
- **Empty Skills state:** `project -S Skills` with no skills emits "No skills found in the restored direct package dependencies." instead of a bare header.
- **implements Type column:** Rendered as a C#-friendly code span (`` `JsonConverter<T>` ``) instead of raw metadata arity (`` JsonConverter`1 ``), matching the existing type/member column convention. Machine formats (`--tsv`/`--jsonl`/`--json`) are unaffected — Markout strips the `<code>` wrapper there and they already emitted `JsonConverter<T>`.
- **diff full markdown:** Generic arity is expanded in headings/rows (`` Box`1 `` → `Box<T>`) and the version line uses a real `→` arrow. Machine views (AnalysisDiffView/DiffTableView) keep plain `->`.

## Evidence

- New tests: 9 (2 diff unit tests, 5 CLI/integration tests, 2 predicate tests for the Decode-column visibility) — all pass.
- Existing focused suites re-run with no regressions (SelectResolver, Diff, Skill, OutputFormatter, and CommandExecution PerformanceTriage/Implements/TopLeverage/Select/Decode/MemberIndex method filters).
- Full solution builds clean.

## Adversarial review

Two-model review (GPT-5.5 + Gemini 3.1 Pro) on the exact head, each in an isolated checkout. Gemini: clean. GPT-5.5 flagged that the Decode-column drop also removes `decode` from `--jsonl`. Verified against origin/main: this is consistent with the tool's existing `MarkoutIgnoreColumnWhen` behavior (the Integrations section already drops the uniform `kind` column from jsonl on main), and the column returns whenever a member is degraded. Corrected the summary wording (this bullet) and added predicate tests locking in that the degraded marker stays visible. Re-review of the fixed head pending.

## Non-actions / follow-ups

- **Performance Triage `--table` 29-column width** → #2833 (needs section-aware column selection; a design change).
- **One attribute per line in decompiler/Signature** → #2834 (decompiler-printer, adversarial-review-gated).
- **implements Type FQN/simple-name consistency + duplicate simple names** → #2835 (naming-policy design decision).

## Known limitation

The diff `### Box<T>` heading is entity-escaped by Markout's GroupBy path (`### Box&lt;T&gt;`), which renders correctly in any Markdown viewer but shows `&lt;` in a raw terminal. Pre-existing Markout GroupBy limitation, accepted as-is; still a clear improvement over the prior confusing `` &#96;1 ``.

### DIFF
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -9577,7 +9577,7 @@ public class CommandExecutionTests
         Assert.Equal(1, exit);
         Assert.Contains("not found", error);
         Assert.Contains("Available sections:", error);
-        Assert.Contains("-D", error);
+        Assert.Contains("Run with -D to discover sections", error);
     }
 
     [Fact]
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -9536,6 +9536,75 @@ public class CommandExecutionTests
     }
 
     [Fact]
+    public async Task Project_SkillsSection_EmptyRendersNoSkillsNote()
+    {
+        var (projectPath, tempDir) = CreateProjectWithPackageDocs(
+            new ProjectDocPackage("A.Project.NoSkills", "1.0.0", "README.md", "readme"));
+
+        try
+        {
+            var (exit, output, error) = await RunAppAsync(
+                "project", projectPath, "-S", "Skills");
+
+            Assert.True(exit == 0, $"exit={exit}\nstdout:\n{output}\nstderr:\n{error}");
+            Assert.Empty(error);
+            Assert.Contains("No skills found", output);
+        }
+        finally
+        {
+            Directory.Delete(tempDir, recursive: true);
+        }
+    }
+
+    [Fact]
+    public async Task Library_TopLeverageSection_WithTopFilter_RendersSingleSection()
+    {
+        var (exit, output, error) = await RunAppAsync(
+            "library", TestAssemblyPath, "-S", "Top Leverage", "--top", "1", "--tsv", "--tips", "q");
+
+        Assert.True(exit == 0, $"exit={exit}\nstdout:\n{output}\nstderr:\n{error}");
+        Assert.Empty(error);
+        Assert.Contains("member\tcallers\troot_reach", output);
+        Assert.DoesNotContain("candidate", output);
+    }
+
+    [Fact]
+    public async Task Type_UnknownSelectValue_ListsAvailableSections()
+    {
+        var (exit, output, error) = await RunAppAsync(
+            "type", "System.String", "-S", "ZzzNoSuchSection", "--tips", "q");
+
+        Assert.Equal(1, exit);
+        Assert.Contains("not found", error);
+        Assert.Contains("Available sections:", error);
+        Assert.Contains("-D", error);
+    }
+
+    [Fact]
+    public async Task Type_MemberIndexSection_OmitsEmptyDecodeColumn()
+    {
+        var (exit, output, error) = await RunAppAsync(
+            "type", "System.Text.StringBuilder", "-S", "Member Index", "--tips", "q");
+
+        Assert.True(exit == 0, $"exit={exit}\nstdout:\n{output}\nstderr:\n{error}");
+        Assert.Empty(error);
+        Assert.Contains("| Selector | Stable | Canonical Signature |", output);
+        Assert.DoesNotContain("Decode", output);
+    }
+
+    [Fact]
+    public async Task Implements_TypeColumn_RendersGenericNameAsCodeSpan()
+    {
+        var (exit, output, error) = await RunAppAsync(
+            "implements", "System.Text.Json.Serialization.JsonConverter",
+            "--platform", "System.Text.Json", "--tips", "q");
+
+        Assert.True(exit == 0, $"exit={exit}\nstdout:\n{output}\nstderr:\n{error}");
+        Assert.Contains("`System.Text.Json.Serialization.JsonConverter<T>`", output);
+        Assert.DoesNotContain("JsonConverter&#96;1", output);
+    }
+
+    [Fact]
     public async Task Project_SkillsPrint_JsonlIsSingleCompactRecord()
     {
         var (projectPath, tempDir) = CreateProjectWithPackageDocs(
--- a/src/dotnet-inspect.Tests/DiffCommandTests.cs
+++ b/src/dotnet-inspect.Tests/DiffCommandTests.cs
@@ -131,6 +131,47 @@ public class DiffCommandTests
     }
 
     [Fact]
+    public void RenderFullMarkdown_GenericType_UsesCSharpFriendlyHeading()
+    {
+        var change = new ApiChange(
+            ChangeKind.MemberSignatureChanged,
+            ChangeClassification.Breaking,
+            "Member 'Write' signature changed",
+            "void Write(T value)",
+            "void Write(T? value)");
+
+        var markdown = DiffOutputFormatter.RenderFullMarkdown(
+            "Sample",
+            [new TypeDiff("Sample.Box`1", [change])],
+            "1.0.0",
+            "2.0.0");
+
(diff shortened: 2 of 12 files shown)


richlander:

Two-model review on the exact head, each in an isolated checkout (scratch under `/tmp`, no writes to the review trees).

**Reviewers:** GPT-5.5 and Gemini 3.1 Pro.

### Round 1 (head `6fc537b0`)
- **Gemini 3.1 Pro:** clean — no issues.
- **GPT-5.5:** one blocking finding — the `Decode`-column drop (`MarkoutIgnoreColumnWhen`) also removes `decode` from `--jsonl`, which contradicted the summary's "no machine-format shape change" claim.

### Resolution
Verified GPT-5.5's observation is accurate, but the behavior is the tool's **existing** cross-format adaptive-column convention, not a new regression:
- `MarkoutIgnoreColumnWhen` is already used on `origin/main` for uniform `Kind` columns. Proof (origin/main, no PR changes): `library System.Diagnostics.DiagnosticSource -S "Integrations" --jsonl` → `{"integration":"OpenTelemetry","apis":"42"}` — the uniform `kind` column is already dropped from jsonl by the same attribute.
- The Decode column returns whenever any member is `degraded` (predicate is `rows.All(r => IsNullOrEmpty(r.Decode))`), so the decode-failure signal is never suppressed.

Actions:
- Added predicate tests (`caed89bf`): `SignatureDecodeIsEmpty_HidesColumnOnlyWhenNoMemberDegraded`, `MemberIndexDecodeIsEmpty_HidesColumnOnlyWhenNoMemberDegraded` — lock in keep-failure-visible.
- Corrected the PR summary to drop the absolute claim and describe the convention.

### Round 2 (fixed head `caed89bf`)
- **GPT-5.5:** clean — "Previous blocking finding is resolved… No significant issues found."
- **Gemini 3.1 Pro:** clean — "No blocking issues or correctness regressions."

Ready to merge

richlander:
## Review (Claude Fable 5)

**Verdict: solid UX slice, no blockers. One claim to verify before merge (machine-format leak of arity expansion), one keep-failure-visible suggestion, rest is clean.**

### Verified

- **Section-selection fix is correctly scoped in all three parsers**: `hasExplicitSelect` is captured before the internal appends (`Source Files` from `--type-filter`, `MemberIndex` from the member flag), so only the user's `-S` suppresses the Performance Triage auto-surface — internal section plumbing doesn't. This honors the "selection, row queries, and formats stay orthogonal" contract.
- **Decode-column predicates match the file's existing `MarkoutIgnoreColumnWhen` convention** (the exception-region predicates directly above), and the new tests lock the load-bearing property: any `degraded` member keeps the column, so the decode-failure signal cannot be suppressed. GPT-5.5's jsonl concern is correctly dispositioned against the Integrations-kind precedent on main.
- Unknown `-S` dead-end → full section list + `-D` pointer, on stderr with exit 1: right shape, and `ListsAllSections` keeps the "Did you mean" fuzzy path unchanged.
- Skills empty state and the `implements` code-span rendering are consistent with existing conventions; the entity-escaped heading limitation is honestly documented.

### Finding — does the arity expansion leak into machine formats?

The summary scopes the `` Box`1 `` → `Box<T>` change to "diff full markdown," but the diff also rewrites `BuildTableView` and `BuildDetailedRow` to use `FormatTypeDisplayName`. If `DiffTableView` (or the detailed rows) feed `--tsv`/`--jsonl` for the diff command, then machine output changes from `` Box`1 `` to `Box<T>` — a compat change for scripts parsing those fields, and contrary to the "machine views keep plain `->`" framing (which covers the arrow, not the names). Please confirm which formats consume `DiffTableView`: if markdown/table only, add that sentence to the body; if tsv/jsonl too, either scope the formatting to the human views or call the machine-format change out explicitly as intended.

### Suggestion (keep failure visible)

With the fix, `library X -S "Top Leverage" --loop` silently ignores the triage row filters (previously they surfaced the Performance Triage section, so they always did *something*). A one-line stderr note when triage filters are present but Performance Triage is not among the selected sections would keep the no-op visible without re-breaking single-section formats.

### Nit

- `Type_UnknownSelectValue_ListsAvailableSections` asserts on `error` containing `-D` — fine, but asserting the exact "Run with -D" sentence would keep the test from passing on any stray `-D` in a section name.

richlander:
Thanks — all three addressed in `a65587ea`.

### Finding: arity expansion leaking into machine formats — confirmed and fixed

You're right. `BuildTableView` and `BuildDetailedChangesView` back the diff command's `--table`/`--tsv`/`--jsonl` path (via `WriteProjectedTable`), so `FormatTypeDisplayName` there changed the `Type` field from `` Box`1 `` to `Box<T>` — a compat change for scripts, with no adaptive-column precedent (unlike the Decode/Kind columns).

Fixed by scoping the expansion to the human Markdown view only:
- Reverted `BuildTableView` and `BuildDetailedRow` to `TypeMatcher.GetSimpleName` (canonical metadata name, exactly as on `origin/main`).
- Kept `FormatTypeDisplayName` only in `BuildChangeRows` → `RenderFullMarkdown` (the GroupBy heading — the actual finding-5 target).
- Updated the doc comment to state the human-only scope.
- Added `BuildTableView_And_DetailedChanges_KeepCanonicalArityForMachineOutput` asserting the machine `Type` field stays `` Box`1 ``.

### Suggestion: keep the ignored-triage-filters no-op visible — filed as #2839

Verified: `--top`/`--loop` do not affect the Top Leverage section (row count identical with/without), so `-S "Top Leverage" --top 1` silently ignores the filters. (These flags never filtered non-Performance-Triage data; this PR removed the previous broken side effect of force-appending the section, which makes the pre-existing no-op fully silent.)

Filed as #2839 rather than bolting it on here: doing it correctly means emitting the note at the render layer where the selection is already resolved to concrete sections — a parser-side heuristic can't tell whether an explicit `-S` glob/abbreviation resolves to Performance Triage and would risk false warnings when the user legitimately selects it. Issue has the design note.

### Nit: exact `-D` sentence

Done — `Type_UnknownSelectValue_ListsAvailableSections` now asserts `"Run with -D to discover sections"` instead of a bare `-D`.

New head `a65587ea` changes product code (the two-site revert), so I'm re-running the two-model review on it before re-marking ready.

richlander:
## Re-review of fixed head `a65587ea`

After scoping the diff arity-expansion to the human Markdown view only (see previous comment), both reviewers re-reviewed the exact new head in isolated checkouts:

- **GPT-5.5:** clean — verified `FormatTypeDisplayName` is called only from `BuildChangeRows`; `--table`/`--tsv`/`--jsonl` keep canonical arity (`` EventHandler`1 ``); full-markdown headings still render the expanded `EventHandler<T>`; DiffCommandTests 115/115.
- **Gemini 3.1 Pro:** clean — same verification; the new `BuildTableView_And_DetailedChanges_KeepCanonicalArityForMachineOutput` test locks the machine output; no regressions.

Both fixed-head re-reviews are clean. Follow-up #2839 filed for the ignored-triage-filters note.

Ready to merge

PR #2832, opened 2026-07-17
Raise real-world fidelity sampling

richlander:
- Advances #2815
- Changes the harness-only real-world compile-back sample; product output is unchanged
- Evidence revision: `8f32b4b6`

## Change

> Should we accept this corpus-gate change?

**Conclusion:** **PASS** — raise the deterministic per-assembly fidelity cap
from 3 to 50, increasing useful compile-back coverage from 36 to 600 methods
while keeping the full sensor near 15–17 minutes locally.

The change updates the committed baseline, Deep Inspect flag, and documented
reproduction command atomically.

## Scope and safety

- **Root cause:** the standing sample checked 36 of 89,065 methods (0.04%),
  leaving the deepest semantic oracle too sparse for broad discovery.
- **Fix shape:** raise only the existing deterministic sample cap; no sampler,
  product, decompiler, or changed-method logic changes.
- **Product impact:** none.
- **Scope boundary:** the global sample is discovery breadth, not fidelity
  proof for a behavior PR. Risky changes still use
  `--fidelity-check --fidelity-method-delta`.
- **RTS boundary:** not changed.

## Measured staircase

All runs used the fixed 14-assembly corpus, `--compile-cap 4000`, and one
fidelity cap per full sensor invocation.

| Cap | Checked | Exact | OpcodeDiff | RecompileFail | ContextFail | Wall time |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 3 | 36 | 32 | 4 | 0 | 0 | 382s |
| 10 | 120 | 106 | 14 | 0 | 0 | 381s |
| 25 | 300 | 261 | 39 | 0 | 0 | 584s |
| **50** | **600** | **510** | **90** | **0** | **0** | **909s** |
| 100 | 1,151 | 964 | 187 | 0 | 0 | 1,683s |

Useful coverage does not plateau, so cost selects the standing cap. Cap 50
raises coverage 16.7x at 15m09s. Cap 100 adds 551 checks but nearly doubles the
sensor to 28m03s. A clean post-merge cap-50 repeat completed in 995s.

Twelve assemblies contribute exactly 50 checked methods each. The two Roslyn
compiler assemblies remain outside the safely checkable population; this PR
does not adapt the harness around them.

## Decompiler quality diff

Corpus: #1166 real-world decompiler corpus sensor: #1150 pinned NuGet assemblies plus dotnet-inspect managed assemblies. 14 assemblies, 89,065 methods
Baseline ref: `origin/main`
[...]

Issue #2814, opened 2026-07-17
Decompiler quality direction: bug-finding yield over harness infrastructure


richlander:
## Conclusion

Direct decompiler-quality work is ~8% of merged PRs in the measured window,
while harness infrastructure, seam relocation, and other infra consume ~72%.
This issue sets the direction to invert that, adds an admission rule for new
harness surface, and orders the executable docket around demonstrated
defects first.

## Evidence (reproducible baseline)

**Window:** merged PRs with `mergedAt` strictly after #2722's merge
(`2026-07-13T13:46:27Z`) and at or before this issue's creation
(`2026-07-17T17:09:21Z`) — **60 PRs in four days**. Note two auditing
wrinkles: #2723 merged 33 seconds *before* #2722 and is excluded; #2708,
#2713, #2714, #2716, #2717, #2719, #2720, #2721 are numerically lower than
#2722 but merged after it and are included (merge-time, not number,
ordering).

Classification (every PR in the window, by primary focus):

- **A — decompiler output quality: 5 PRs (8.3%)** — #2739, #2750, #2752,
  #2756, #2774 (two raises, three correctness/fidelity fixes)
- **B — RTS seam relocation (behavior-preserving): 11 PRs (18.3%)** —
  #2737, #2751, #2762, #2763, #2764, #2765, #2771, #2791, #2797, #2802,
  #2803
- **C — harness/test infrastructure: 15 PRs (25.0%)** — #2708, #2728,
  #2732, #2735, #2748, #2755, #2758, #2760, #2768, #2783, #2784, #2785,
  #2788, #2794, #2796
- **D — Analysis/Triage features: 12 PRs (20.0%)** — #2716, #2717, #2725,
  #2729, #2736, #2740, #2741, #2769, #2773, #2787, #2792, #2801
- **E — other product infra/hardening/docs: 17 PRs (28.3%)** — #2713,
  #2714, #2719, #2720, #2721, #2733, #2743, #2749, #2757, #2761, #2772,
  #2795, #2799, #2800, #2805, #2808, #2809

B+C+E = 43/60 = **71.7%**; A = **8.3%**.

Confirmed bug-finding yield concentrates in a few lanes: validity sweeps
(#922–#934, #697, the invalid-at-Full fix train), structuring censuses
(#911–#914), opt-in feature corpora (union raises; #2756/#2774 from
fixture/corpus frontiers), and manual adversarial review — the only finder
[...]

PR #2802, opened 2026-07-17
Move type-shell print-request composition into the ILInspector.CSharp seam

richlander:
## Why

Part of #2777 (make ReturnToSender a thin orchestrator; move C# source
composition out of the harness into the product seam). This slice retires the
harness-local `TypeProducer` `ApiType`/`CSharpTypePrintRequest` assembly so the
seam — not the harness — owns turning type-shell facts into a print request.

## What

The harness now builds a **neutral `CSharpTypeShellSpec` tree** and hands the
root specs to a new seam entry point:

- **New** `src/ILInspector.CSharp/CSharpTypeShellSpec.cs` — `CSharpTypeShellKind`
  + `CSharpTypeShellSpec` (a plain-string / metadata-handle / seam-type DTO; **no
  `CompileBack*` planning vocabulary crosses the boundary**).
- **New** `TypeShellProducer.BuildPrintRequest(reader, spec)` — assembles the
  `ApiType` (kind text, generic parameters, custom attributes,
  abstract/sealed/static modifiers read straight from metadata) and recurses into
  nested shells. SRM-only, Roslyn-free, NativeAOT-friendly.
- **Harness** `TypeRequest` -> `BuildSpec`, `NestedTypes` -> `NestedSpecs`. Member
  discovery, `CompileBack* -> CSharpMemberPolicy` mapping, base/interface C#
  naming, and `plan.Types` evidence bookkeeping are **unchanged**. The
  harness-local `TypeKindText`/`TypeAttributeList` are removed (moved/inlined into
  the seam), replaced by a trivial `CompileBackTypeKind -> CSharpTypeShellKind`
  mapper.

### Intentionally left in the harness (separate threads)

- **Base/interface display naming** still flows through
  `CompileBackTypeIdentity`/`Clean`; those strings are pre-computed harness-side
  and passed into the spec. Migrating that spelling is a separate ongoing thread.
- **Member discovery + `CompileBack*` planning + evidence** (`plan.Types`,
  `producedRequirements`) — the ReturnToSender oracle's irreducible job.

## Evidence (behavior-preserving carve)

- Full `dotnet-inspect.slnx` build clean.
- RTS oracle: Evidence/Catalog 103 + Prototype 114 (**217** tests).
- Decompiler fast subset **2476** (printer byte-identity).
- CSharp.Tests **240**, incl. a new `BuildPrintRequest` seam test (spec-driven
  facts vs metadata-driven modifiers vs nested recursion).
[...]

PR #2838, opened 2026-07-17
Move base-type spelling into the seam; retire redundant harness usings (#2777)

richlander:
- Advances #2777 (RTS orchestrator thin; product seam owns C# spelling/shape)
- Changes: base-type C# spelling now computed by the seam; harness stops deriving its own declaration usings (printer owns them since #2825)

Two focused, independent epic-#2777 seam-ownership cleanups, kept as separate commits.

## 1. Move base-type spelling into the seam (byte-neutral)

The reconstructed base-type C# display name was computed harness-side
(`CompileBackSourceComposer.BaseTypeSignature`: metadata gate → `Clean`
normalization → surface gate) and carried across the boundary on
`CSharpTypeShellSpec.BaseTypeDisplayName`. That left a C# spelling decision in the
ReturnToSender harness.

- Add `TypeShellProducer.ReconstructedBaseTypeDisplay`, folding the existing seam
  primitives (`ReconstructedBaseTypeName` + `CleanTypeDisplay` +
  `IsUnsupportedSurfaceSignature`) into the full base-spelling decision.
- Compute `BaseType` inside `BuildPrintRequest` from the type's own metadata; drop
  `BaseTypeDisplayName` from the spec. `ReconstructedSameAssemblyBaseName`
  (base-of-another-type planning, legitimately harness-side) calls the seam for its
  reconstructed/not decision, so `BaseTypeSignature` is deleted.
- **Interface selection stays harness-side**: it uses RTS planning vocabulary
  (`IsSupportedClosureRoot`), so moving it would drag planning into the seam.
  `InterfaceDisplayNames` remains on the spec deliberately.

Behavior-preserving: same metadata gate, same `Clean`, same surface gate, same order.

## 2. Retire redundant harness declaration using-derivation (resolves 2 rows)

Since #2825 the printer derives contextual usings itself and shortens a name only
when it has derived the matching using — so every printed name is either fully
qualified (no using needed) or shortened with its using present. That makes the
harness's own declaration using-derivation (`DeclarationNamespaces` /
`AddDeclaredTypeNames` / `TypeNamespaces`, unioned into `plan.Module.Usings`)
redundant for compilation — and, being **not collision-safe**, it could inject an
ambiguous `using` the printer would have withheld.

Dropped the union from the three module-usings sites and deleted the three helpers.
`RequiredNamespaces(function)` stays (body-referenced namespaces the printer cannot
derive from declaration signatures).

## Evidence

| Check | Result |
| --- | --- |
| Product build (`dotnet-inspect.slnx -c Release`) | Pass |
| `ILInspector.CSharp.Tests` | 269 pass (BuildPrintRequest test reworked to prove seam-computed base) |
| RTS oracle (`ReturnToSenderPrototypeTests`/`EvidenceTests`/`FixtureCatalogTests`) | 199 pass, **zero prototype-fixture churn** |
| Decompiler fast suite (`-notrait Speed=Slow`) | 2496 pass |

### rts-parity corpus A/B (merge-base, deterministic cap-3 gate)

Pinned 14-assembly corpus, `--corpus-fidelity-cap 3 --corpus-fidelity-oracle rts-parity`:

| | exact | opcode | recompile-fail | manifest rows failing |
| --- | ---: | ---: | ---: | ---: |
| origin/main (a7bb425b) | 25 | 2 | 9 | 7/7 |
| this branch | 27 | 2 | 7 | 5/7 |

**+2 resolved, 0 regressions** (opcode diffs unchanged; no previously-Exact target dropped). The non-collision-safe harness usings injected an ambiguous `using` the printer's collision-safe derivation withholds, so these now recompile Exact:

- `Newtonsoft.Json!Newtonsoft.Json.JsonTextReader::get_PropertyNameTable#0`
- `System.CommandLine!...EnvironmentVariablesDirectiveAction::SetEnvVars#0`

Burndown manifest mechanically regenerated (2 rows dropped, 5 remain); enforcing gate exit 0.

### Scope / cap notes

- The committed `rts-parity-burndown.json` stays **cap-3 basis**, matching origin/main's committed manifest. #2832 raised the *documented* regeneration cap to 50 but did not regenerate the manifest; migrating it to cap-50 (which adds ~50 pre-existing rows) is a separate #2832 follow-up, out of scope here.
- Neither PR CI (`--corpus-fidelity-cap 0`, no rts-parity oracle) nor Deep Inspect (diffs the real-world baseline) enforces this manifest; it is a local-agent artifact.
- A broad cap-50 sweep was attempted but this shared machine's variable external load made the (worker-influenced) checked population non-comparable across runs (250/100/50 checked). Broad no-regression confidence rests on the deterministic cap-3 gate plus the #2825 collision-safety invariant (the printer never shortens a name without emitting its using). Reviewers on a quiet machine can run the documented cap-50 command.

## Adversarial review

Requesting two reviews (GPT-5.5 + Gemini 3.1 Pro) per the two-different-model policy (I am Opus 4.8). Focus: (a) is the base-spelling migration truly behavior-preserving; (b) can retiring the harness usings ever drop a *needed* using (regression), given the #2825 shorten-iff-derived-using contract.

### DIFF
--- a/src/ILInspector.CSharp.Tests/TypeShellProducerTests.cs
+++ b/src/ILInspector.CSharp.Tests/TypeShellProducerTests.cs
@@ -86,22 +86,23 @@ public sealed class TypeShellProducerTests
         var member = new CSharpMemberPolicy(
             new ApiMember { Name = "Value", Kind = "field", ReturnType = "System.Int32" },
             CSharpBodyPolicy.Skeleton);
+        // Nested spec deliberately claims Kind=Struct for a type whose metadata is a
+        // static class, proving modifiers are read from metadata rather than the spec
+        // kind, and that an object-family base is left implicit (null).
         var nested = new CSharpTypeShellSpec(
-            Handle(nameof(InstanceFixture)),
+            Handle(nameof(StaticFixture)),
             Namespace: "Samples",
-            MetadataName: "InstanceFixture",
-            Kind: CSharpTypeShellKind.Class,
-            BaseTypeDisplayName: null,
+            MetadataName: "StaticFixture",
+            Kind: CSharpTypeShellKind.Struct,
             InterfaceDisplayNames: [],
             MemberPolicies: [],
             PrimaryConstructorParameters: [],
             NestedTypes: []);
         var spec = new CSharpTypeShellSpec(
-            Handle(nameof(StaticFixture)),
+            Handle(nameof(DerivedFixture)),
             Namespace: "Samples",
-            MetadataName: "StaticFixture",
-            Kind: CSharpTypeShellKind.Struct,
-            BaseTypeDisplayName: "Samples.Widget",
+            MetadataName: "DerivedFixture",
+            Kind: CSharpTypeShellKind.Class,
             InterfaceDisplayNames: ["System.IDisposable"],
             MemberPolicies: [member],
             PrimaryConstructorParameters: [],
@@ -111,26 +112,28 @@ public sealed class TypeShellProducerTests
 
         // Spec-supplied facts flow straight through.
         Assert.Equal("Samples", request.Type.Namespace);
-        Assert.Equal("StaticFixture", request.Type.Name);
-        Assert.Equal("StaticFixture", request.Type.MetadataName);
-        Assert.Equal("struct", request.Type.Kind);
-        Assert.Equal("Samples.Widget", request.Type.BaseType);
+        Assert.Equal("DerivedFixture", request.Type.Name);
+        Assert.Equal("DerivedFixture", request.Type.MetadataName);
+        Assert.Equal("class", request.Type.Kind);
         Assert.Equal(["System.IDisposable"], request.Type.Interfaces);
         Assert.Same(member, Assert.Single(request.MemberPolicyOverrides));
         Assert.Equal("Value", Assert.Single(request.Type.Members).Name);
         Assert.Same(member.Member, request.Type.Members[0]);
 
-        // Modifiers are read from the type's own metadata, not the spec kind.
-        Assert.True(request.Type.IsStatic);
-        Assert.True(request.Type.IsAbstract);
-        Assert.True(request.Type.IsSealed);
+        // The base type is reconstructed by the seam from the type's own metadata
+        // (same-assembly non-generic class base), not carried on the spec.
+        Assert.NotNull(request.Type.BaseType);
+        Assert.EndsWith("BaseFixture", request.Type.BaseType);
 
-        // Nested shells recurse through the same builder.
+        // Modifiers are read from the type's own metadata, not the spec kind.
         var nestedRequest = Assert.Single(request.NestedTypes);
-        Assert.Equal("InstanceFixture", nestedRequest.Type.Name);
-        Assert.Equal("class", nestedRequest.Type.Kind);
+        Assert.Equal("StaticFixture", nestedRequest.Type.Name);
+        Assert.Equal("struct", nestedRequest.Type.Kind);
+        Assert.True(nestedRequest.Type.IsStatic);
+        Assert.True(nestedRequest.Type.IsAbstract);
         Assert.True(nestedRequest.Type.IsSealed);
-        Assert.False(nestedRequest.Type.IsStatic);
+        // A static class's object-family base is left implicit.
+        Assert.Null(nestedRequest.Type.BaseType);
     }
 
     static async Task RuntimeAsyncFixture()
@@ -148,6 +151,14 @@ public sealed class TypeShellProducerTests
     }
 
     static class StaticFixture
+    {
+    }
+
+    class BaseFixture
+    {
+    }
+
+    sealed class DerivedFixture : BaseFixture
     {
     }
 
--- a/src/ILInspector.CSharp.Tests/TypeShellProducerTests.cs
+++ b/src/ILInspector.CSharp.Tests/TypeShellProducerTests.cs
@@ -123,7 +123,9 @@ public sealed class TypeShellProducerTests
         // The base type is reconstructed by the seam from the type's own metadata
         // (same-assembly non-generic class base), not carried on the spec.
         Assert.NotNull(request.Type.BaseType);
-        Assert.EndsWith("BaseFixture", request.Type.BaseType);
+        Assert.Equal(
+            "ILInspector.CSharp.Tests.TypeShellProducerTests.BaseFixture",
+            request.Type.BaseType);
 
         // Modifiers are read from the type's own metadata, not the spec kind.
         var nestedRequest = Assert.Single(request.NestedTypes);
--- a/src/ILInspector.CSharp/CSharpTypeShellSpec.cs
+++ b/src/ILInspector.CSharp/CSharpTypeShellSpec.cs
@@ -21,8 +21,9 @@ public enum CSharpTypeShellKind
 /// A neutral, IR- and planner-free description of a type shell to compose into a
 /// <see cref="CSharpTypePrintRequest"/>. The consumer (for example the ReturnToSender
 /// harness) discovers the members and resolves the C# spelling of the type's own
-/// name, base type, and interfaces; the seam owns turning this description plus the
-/// type's metadata into the <see cref="ApiType"/> and print-request tree. Every field
+/// name and interfaces; the seam owns turning this description plus the
+/// type's metadata into the <see cref="ApiType"/> (including the base-type spelling)
+/// and print-request tree. Every field
 /// is a plain string, a metadata handle, or an already-seam/Metadata type — no
 /// consumer planning type crosses this boundary.
 /// </summary>
@@ -35,8 +36,6 @@ public enum CSharpTypeShellKind
 /// used verbatim for both <see cref="ApiType.Name"/> and
 /// <see cref="ApiType.MetadataName"/>.</param>
 /// <param name="Kind">The reconstructed C# kind.</param>
-/// <param name="BaseTypeDisplayName">The C# display spelling of the reconstructed
-/// base type, or null when no base is reconstructed.</param>
 /// <param name="InterfaceDisplayNames">The C# display spellings of the reconstructed
 /// implemented interfaces.</param>
 /// <param name="MemberPolicies">The members to emit with their per-member body
@@ -49,7 +48,6 @@ public sealed record CSharpTypeShellSpec(
     string Namespace,
     string MetadataName,
     CSharpTypeShellKind Kind,
-    string? BaseTypeDisplayName,
     IReadOnlyList<string> InterfaceDisplayNames,
     IReadOnlyList<CSharpMemberPolicy> MemberPolicies,
     IReadOnlyList<ApiParameter> PrimaryConstructorParameters,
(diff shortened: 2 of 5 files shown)


richlander:
## Review (Claude Fable 5)

**Verdict: commit 2 is verified sound and is real Tier-0 yield (2 burn-down rows drained by deleting harness code). Commit 1 is the right move but its "behavior-preserving… same order" claim is contradicted by the diff — one finding to resolve before merge.**

### Finding 1 — the surface gate moved from the raw name to the cleaned name; that is a behavior change, not a byte-neutral move

Old pipeline (deleted `BaseTypeSignature`): `ReconstructedBaseTypeName` → `IsUnsupportedSurfaceSignature(raw)` → `Clean` (inside `CompileBackTypeSignature.Display`). New seam helper (`ReconstructedBaseTypeDisplay`): `ReconstructedBaseTypeName` → `Clean` → `IsUnsupportedSurfaceSignature(cleaned)`. The gate matches `<>` and `{` — exactly what `CleanTypeDisplay`'s generated-segment sanitizer removes — so the two orders diverge on generated-name bases: a `<>`/`{`-named base was **dropped** by the old code and is **kept (sanitized to `___…`)** by the new one. (`delegate*`/`@delegate*` are order-insensitive since Clean preserves/escapes them.)

To be clear: the new order looks like the *intended* semantics — the product-side comment from the #2802 era says gating on the raw metadata name "would diverge… e.g. wrongly dropping a generated `<>`-named base." So the old code was arguably carrying a latent gate-order bug, and this PR silently fixes it. But then the PR should own that: the "same order" sentence is factually wrong, and "zero prototype-fixture churn / corpus cap-3 neutral" is consistent with a ≈0 population of generated-name bases, not with impossibility. Ask: either (a) state the order change explicitly and add a close-case test (a shell whose base carries a generated `<>` segment, proving the sanitized base composes and recompiles — or proving the metadata gate makes it unreachable), or (b) restore raw-order gating if byte-neutrality was the actual contract for this slice. Given the documented intent, (a) is the better PR.

### Commit 2 — verified, and the best kind of fix

- The soundness argument holds: post-#2825 the printer shortens a name **iff** it derived the enabling using, so every printed declaration name is either fully qualified (needs no using) or covered. Body-referenced namespaces stay via `RequiredNamespaces(function)`. Retiring the harness derivation cannot drop a needed using.
- The deleted `TypeNamespaces` splitter was last-dot string splitting — the same collision-unsafe class as the nested-type-as-namespace bug found in #2825 — and the corpus A/B proves it was actively *causing* failures: +2 Exact, 0 regressions, and the manifest diff removes exactly `JsonTextReader::get_PropertyNameTable` and `EnvironmentVariablesDirectiveAction::SetEnvVars`. Two of seven burn-down rows drained by **deleting** 60 lines of harness code is precisely the #2814/#2782 direction working as designed.
- The cap-3 manifest-basis discipline (not conflating with #2832's cap-50 migration) and the honest note about the noisy cap-50 attempt on a loaded machine are both right.

### Nit

- The reworked `BuildPrintRequest` test now asserts `EndsWith("BaseFixture")` — asserting the exact expected display (`…TypeShellProducerTests.BaseFixture` or its cleaned form) would pin the spelling this PR moves, which is the thing under test.

richlander:

Two independent reviews on head `9579ceba`, each in an isolated detached worktree (never shared). Both **CLEAN**. Per the two-different-model policy I am Opus 4.8, so reviewers were GPT-5.5 and Gemini 3.1 Pro.

### GPT-5.5 — CLEAN
Verified with real runs: solution + harness build; `ILInspector.CSharp.Tests` 269; decompiler fast subset (`-notrait Speed=Slow`) 2496; RTS oracle (`ReturnToSenderPrototypeTests`/`EvidenceTests`/`FixtureCatalogTests`) 199; manifest base 7 rows → HEAD 5, exactly `get_PropertyNameTable` + `SetEnvVars` dropped, no additions, remaining rows unchanged. `git status` clean before/after. No PR-introduced correctness/compat/regression issue.

### Gemini 3.1 Pro — CLEAN
Verified with real runs: solution build; CSharp.Tests 269; decompiler fast subset (2326 discovered, RTS 199, zero fixture churn); deterministic cap-3 rts-parity A/B exit 0, exactly the 2 rows resolved, 0 regressions. Plus code-path tracing of each attack point.

### Agreement / divergence
Both reviewers agreed CLEAN on every attack point; **no divergence, no findings to reconcile.** Both independently confirmed the two load-bearing invariants:

- **(A) Base-migration equivalence** — old `BaseTypeSignature` and new `TypeShellProducer.ReconstructedBaseTypeDisplay` both call `ReconstructedBaseTypeName` → `CleanTypeDisplay` → `IsUnsupportedSurfaceSignature` over the cleaned string, identical gate order / null propagation / boolean mapping. Byte-for-byte equivalent (confirmed by 199 RTS oracle with zero fixture churn).
- **(B) Declaration-usings safety** — Gemini traced the mechanism precisely: `CSharpTypePrinter` shortens a name **only** when it is in `DeriveContextualUsings` (or matches the containing namespace); the old harness `DeclarationNamespaces` union only affected the printed top-of-file `using` list and was physically disconnected from the shortening decision. If `DeriveContextualUsings` misses a reference, the printer emits it fully qualified, so no `using` is needed — retiring the harness usings can only remove an ambiguous/unneeded using, never drop a needed one. This is the #2825 shorten-iff-derived-using contract.
- **(C) Layering** — base reconstruction filters to same-assembly (`BaseType.Kind == TypeDefinition`), so it is safely seam-owned; interfaces can be cross-assembly `TypeReference`s and need the `IsSupportedClosureRoot` planning gate, so keeping interface selection harness-side is correct, not a broken intermediate state.
- **(D) Manifest regen** — mechanically accurate; only the 2 targeted rows removed.

No resolution commits are needed because there were no actionable findings. Two CLEAN reviews on the current head → **Ready to merge**. I will not self-merge; awaiting @richlander approval.

richlander:
## Response to the Claude Fable 5 review

Thanks — I traced Finding 1 to the deleted code and it does **not** hold. The old gate already ran on the **cleaned** name, so there is no order change and no generated-`<>`-base divergence. Adopted the Nit.

### Finding 1 — not a behavior change (the old gate already cleaned first)

The review reads `BaseTypeSignature` as `IsUnsupportedSurfaceSignature(raw) → Clean`, but the call `IsUnsupportedSurfaceSignature(baseType)` inside `BaseTypeSignature` resolves to a **harness-local wrapper**, not `TypeShellProducer.IsUnsupportedSurfaceSignature`. That wrapper cleans first:

```csharp
// tools/DecompilerHarness/ReturnToSenderTypePlanner.cs:3286 (@ merge-base a7bb425b)
static bool IsUnsupportedSurfaceSignature(string signature)
    // Normalize the raw signature text to its C# display form first ...
    => TypeShellProducer.IsUnsupportedSurfaceSignature(
           CompileBackTypeSignature.Display(signature).DisplayName);
```

And `CompileBackTypeSignature.Display` **is** `CleanTypeDisplay`:

```csharp
// ReturnToSenderTypePlanner.cs:248-249
public static CompileBackTypeSignature Display(string text)
    => new(CompileBackTypeSignatureKind.Display, CSharpFormatter.CleanTypeDisplay(text), null);
```

So the two paths are the same clean, the same gate, on the same cleaned input:

| | gate input | returns |
| --- | --- | --- |
| **old** `BaseTypeSignature` | `TypeShellProducer.IsUnsupportedSurfaceSignature(CleanTypeDisplay(base))` | `CleanTypeDisplay(base)` |
| **new** `ReconstructedBaseTypeDisplay` | `IsUnsupportedSurfaceSignature(CleanTypeDisplay(base))` (same method) | `CleanTypeDisplay(base)` |

The gate never saw the raw metadata name in either version, so a generated-`<>`/`{`-named base is treated identically before and after (dropped in both — the `<>`/`{` gate patterns survive `CleanTypeDisplay`, which only strips generated segments to `___…` when they are dotted *name segments*, not when they are the raw `<>` substring the gate matches). "Same order / byte-neutral" stands; zero prototype-fixture churn and cap-3 corpus neutrality are consistent with genuine equivalence, not a masked ≈0 population. The product-side comment the review quotes ("gating on raw would wrongly drop a `<>` base") is describing the invariant the **old harness wrapper already preserved** and the new seam continues to preserve — not a latent bug being silently fixed.

Net: Finding 1 is a mis-attribution of `IsUnsupportedSurfaceSignature` to the product overload instead of the harness-local cleaning wrapper. No action needed for correctness; happy to be re-checked against those two line refs.

### Nit — adopted (`b2f0319d`)

Replaced `Assert.EndsWith("BaseFixture", …)` with the exact spelling:

```csharp
Assert.Equal("ILInspector.CSharp.Tests.TypeShellProducerTests.BaseFixture", request.Type.BaseType);
```

This pins the full reconstructed base display the seam now owns end to end (nested `+` normalized to `.`). `ILInspector.CSharp.Tests` 269 pass.

### Commit 2
Agreed on all points — thanks for the independent confirmation that the deleted `TypeNamespaces` last-dot splitter was the same collision-unsafe class fixed in #2825, and that draining 2/7 burn-down rows by deleting harness code is the intended #2814/#2782 direction.

richlander:
## Third adversarial review — MAI-Code-1-Flash (isolated worktree @ `b2f0319d`)

MAI **independently confirmed** the base-type migration is equivalent to the old harness path (the second reviewer to reach that conclusion, corroborating the refutation of the earlier gate-order finding), and found the declaration-using retirement introduces no compile regression under the RTS/compile-back tests. Its HEAD cap-3 rts-parity run reproduced **27 exact / 2 opcode / 7 recompile-fail, exit 0**.

It raised one new finding, which I've verified as **not applicable (dismissed)**:

### Dismissed — "removing `BaseTypeDisplayName` from public `CSharpTypeShellSpec` is a source-compat break for downstream consumers"

`CSharpTypeShellSpec` is an **internal, unshipped seam DTO**, not a published API with downstream consumers:

- **Not packaged.** `src/ILInspector.CSharp/ILInspector.CSharp.csproj` has no `PackageId`, `IsPackable`, or `GeneratePackageOnBuild`, and there is no such property in any `Directory.Build.*`/props/targets. The library is never produced as a standalone NuGet package.
- **Bundled, not referenced externally.** It ships only inside the `dotnet-inspect` tool via `ProjectReference` (`src/dotnet-inspect/dotnet-inspect.csproj:65`); it is never a `PackageReference` anywhere. There is no external consumer that could take a source break.
- **Epic-local and in-flight.** The record was introduced in #2802 (`9517f883`) as part of this very seam-extraction epic (#2777) and has never been in a released/stable public contract. `InternalsVisibleTo` is limited to `ILInspector.CSharp.Tests`.
- **All consumers updated.** The only references are the seam itself, its tests, and the harness (`ReturnToSenderTypePlanner.cs`) — all updated in this PR.

Adding a compatibility overload that keeps `BaseTypeDisplayName` alive would defeat the entire point of the migration (moving base-type spelling off the spec and into the seam). The `public` modifier here exists for cross-project use *within the repo* (harness ↔ seam), not as an API-stability guarantee. No action taken.

### Status
Three adversarial reviews now on record: GPT-5.5 CLEAN, Gemini 3.1 Pro CLEAN, MAI-Code-1-Flash (base-migration/decl-usings/layering/manifest all confirmed; its sole finding dismissed with evidence above). The one substantive finding across all three (Claude Fable 5's gate-order Finding 1) was refuted with line-level code evidence and independently corroborated by MAI. Still **Ready to merge**; not self-merging — awaiting @richlander.

PR #2842, opened 2026-07-18
Render generic types in async/P/Invoke tables as code spans

richlander:
## Summary

The library command's **Async Methods** and **P/Invoke Methods** sections rendered the `Declaring Type` and `Signature` columns as plain strings. Markout escapes plain table-cell content in Markdown, so generic types leaked HTML entities and declaring types showed raw metadata arity with an escaped backtick:

```
| ParseAsync | System.Text.Json.Nodes.JsonNode | State machine | System.Threading.Tasks.Task&lt;System.Text.Json.Nodes.JsonNode&gt; ParseAsync(...) |
| GetAsyncEnumerator | ...IAsyncEnumerableOfTConverter&#96;2.BufferedAsyncEnumerable | State machine | ... |
```

## Fix

Wrap both columns in `MarkoutInline.Code(...)` and expand generic arity via `MetadataTypeNameFormatter.FormatGenericTypeName(...)`, mirroring the type-name rendering convention established in #2836 (the `implements` Type column and `diff` headings).

`MarkoutInline.Code` is format-aware:

- **Markdown** → clean backtick code span with literal angle brackets.
- **`--table` / `--tsv` / `--jsonl`** → raw inner text, no `<code>`/backticks.

Arity is expanded in the underlying value (as `implements` already does), so machine output shows `IAsyncEnumerableOfTConverter<T1, T2>` rather than the backtick form.

After:

```
| GetAsyncEnumerator | `System.Text.Json.Serialization.Converters.IAsyncEnumerableOfTConverter<T1, T2>.BufferedAsyncEnumerable` | State machine | `System.Collections.Generic.IAsyncEnumerator<TElement> GetAsyncEnumerator(...)` |
```

## Evidence

- New CLI tests: `AsyncMethods_DeclaringTypeAndSignature_RenderAsCodeSpansWithExpandedArity` (Markdown code span, expanded arity, no escaped entities) and `AsyncMethods_MachineOutput_KeepsRawValuesWithoutCodeMarkup` (jsonl rawness).
- Full `dotnet-inspect.Tests` suite: **1934 passed, 0 failed** (10 skipped for absent ildasm).
- Verified real CLI output across `-v:d` Markdown, `--jsonl`, and `--tsv` for `library --platform System.Text.Json --section "Async Methods"`.

## Compatibility

Low-risk, mechanical rendering change consistent with the merged #2836 convention. Markdown-only escaping is fixed; machine formats gain C#-friendly arity in `declaring_type` (matching `implements`), with no `<code>` leakage.

## Review

One adversarial review (GPT-5.5, isolated checkout) — **no blocking findings**. It independently confirmed Markout 0.22 defends against backticks in code spans (longer fences) and escapes pipes in cells, so any residual arity backtick and table-breaking characters stay safe.

### DIFF
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -9584,6 +9584,40 @@ public class CommandExecutionTests
     }
 
     [Fact]
+    public async Task AsyncMethods_DeclaringTypeAndSignature_RenderAsCodeSpansWithExpandedArity()
+    {
+        var (exit, output, error) = await RunAppAsync(
+            "library", "--platform", "System.Text.Json",
+            "--section", "Async Methods", "-v:d", "--tips", "q");
+
+        Assert.True(exit == 0, $"exit={exit}\nstdout:\n{output}\nstderr:\n{error}");
+        // Generic declaring types expand arity to a C#-friendly code span (no raw `2, no escaped angle brackets).
+        Assert.Contains(
+            "`System.Text.Json.Serialization.Converters.IAsyncEnumerableOfTConverter<T1, T2>.BufferedAsyncEnumerable`",
+            output);
+        // Generic signatures render as code spans with literal angle brackets.
+        Assert.Contains("`System.Collections.Generic.IAsyncEnumerator<TElement> GetAsyncEnumerator", output);
+        Assert.DoesNotContain("&#96;", output);
+        Assert.DoesNotContain("&lt;", output);
+        Assert.DoesNotContain("&gt;", output);
+    }
+
+    [Fact]
+    public async Task AsyncMethods_MachineOutput_KeepsRawValuesWithoutCodeMarkup()
+    {
+        var (exit, output, error) = await RunAppAsync(
+            "library", "--platform", "System.Text.Json",
+            "--section", "Async Methods", "--jsonl");
+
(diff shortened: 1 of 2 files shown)


richlander:

**One blocking ordering regression at exact head `c64390c0de90223320f27ff4a4ac44649cd76e1d`.**

- **Claude Opus 4.8:** clean after checking Markdown fences/pipes, raw JSONL/TSV values, formatter edge cases, sorting/projection, and a real P/Invoke assembly.
- **Gemini 3.1 Pro:** found that rows are sorted by the raw metadata declaring type before the value is expanded for display/machine output.

I reproduced the Gemini finding on a clean exact-head checkout with a compiled assembly containing async methods on `OuterZ` and `Outer<T>`. JSONL emits:

```text
SortProbe.OuterZ
SortProbe.Outer<T>
```

The raw sort compares `OuterZ` with ``Outer`1`` (`Z` precedes the metadata backtick), but the rendered values compare `OuterZ` with `Outer<T>` (`<` precedes `Z`). Thus the PR makes the visible and machine-output rows inconsistent with their displayed sort key. This affects both Async Methods and P/Invoke Methods at `LibraryInspectionView.cs:94-101` and `:204-211`.

**Required:** sort on the same formatted declaring-type value that is emitted, and add the generic/non-generic close-negative regression test. Re-review the fixed exact head with both model families.

The missing symmetric P/Invoke rendering test is non-blocking and tracked in #2849. The intended machine-output arity change was reviewed and accepted.

PR #2845, opened 2026-07-18
Render explicit-interface-implementation properties as real explicit impls (RTS CS0535)

### LINKED ISSUE #2827: Render explicit-interface-implementation properties as real explicit impls in the type seam (RTS CS0535)
## Problem

When ReturnToSender shells a class that explicitly implements a same-assembly
interface, the shell declares the interface (`ReturnToSenderTypePlanner.InterfaceSignatures`)
but only includes the target member — not the interface's other members. The shell then
fails to compile with `CS0535` ("does not implement interface member").

Measured over the pinned corpus (RTS shell compile-back across 18,574 property getters):
**26 `CS0535` rows**, e.g. `TraceJsonReader : IJsonLineInfo` (target
`IJsonLineInfo.get_LineNumber`, missing `LinePosition`/`HasLineInfo`),
`XObjectWrapper : IXmlNode`, `Cci.InheritedTypeParameter : IGenericTypeParameter`.

This is isolated to RTS: RTS already renders the explicit-impl member correctly (the seam
omits the access modifier for `Kind = explicit-interface-implementation`), so the only gap
is the un-stubbed sibling members. (The separate legacy compile-back oracle drops these
interfaces entirely, so it never emits `CS0535` — it is out of scope here.)

## Layering / approach

Interface-member-stub projection is C# shaping, so it belongs in the seam
(`TypeShellProducer.BuildPrintRequest`), which already has the `MetadataReader`, the type,
the declared interface list, and the shell's members. It enumerates each declared
same-assembly interface's members (`MetadataDeclarationQuery.GetTypeSurface`), and for each
member not already present in the shell and not a default interface member, appends a
throwing **explicit** interface-implementation stub (dotted `Interface.Member` name, no
access modifier, `throw` body). RTS already calls `BuildPrintRequest`, so no new harness
C# logic is added. SRM-only, Roslyn-free.

## Acceptance

- The 26 measured `CS0535` rows recompile under the RTS getter probe
  (`--return-to-sender`); `CS0535` count drops to ~0 with no new `CS0111`/`CS0108`.
- The #2776 RTS-parity burn-down does not grow.
- Product tests cover an explicit-interface-implementation shell whose interface has a
  second, un-targeted member (property + method), including an inherited-interface member.

## Reference

Supersedes closed #2780/#2824; corrected scope (RTS CS0535, not the legacy oracle's CS0106).

richlander:
## Summary

Explicit-interface-implementation **properties** now recompile as real explicit
impls in the RTS (compile-back) path, fixing CS0535 reconstruction failures.

RTS reconstructs an explicit-impl property closure-minimized to just the probed
member, then sanitized its dotted metadata name (`Interface.Member`) into a bogus
underscore identifier and emitted a plain `public` property. That property does
not satisfy the declared interface, so the target getter was absent from the
recompiled output (the harness reported this as a `ContextFail`
"method-not-found", or CS0535 `RecompileFail` depending on the closure shape).

The fix is split along the product/harness seam (per the north-star: the harness
gets dumber, the product seam owns C#):

- **Harness (`ReturnToSenderTypePlanner`)** stops sanitizing dotted property
  names. A dotted name is an explicit-interface-implementation signal and is
  passed through unchanged.
- **Type seam (`ILInspector.CSharp`)** recognizes a dotted-name property/event
  that carries only a structured model (no compatibility `Signature` string) as a
  real explicit interface implementation: dotted `Interface.Member` name
  preserved, no access modifier — exactly like the seam already does for
  explicit-impl *methods*. When a compatibility `Signature` **is** present (the
  product's `--all` API-surface display), that authoritative display form is
  honored, so API-surface display is unchanged.

Fixes #2827.

## Evidence

Newtonsoft.Json 13.0.4 `--return-to-sender`, per-member diff of the rebased base
(`origin/main`) vs this head, all 581 property getters:

| Status | Base | Head |
| --- | --- | --- |
| Exact | 535 | **539** |
| OpcodeDiff | 16 | 16 |
| RecompileFail | 15 | 26 |
| ContextFail | 15 | **0** |

- **0 regressions**: no member that was `Exact` at base is worse at head.
- **+4 Exact**, all explicit-impl property getters:
  `JsonValidatingReader`/`TraceJsonReader` `IJsonLineInfo.get_LineNumber` and
  `get_LinePosition`.
- **15 ContextFail -> RecompileFail** shifts are all dotted-name explicit-impl
  property getters (`System.Collections.IList.*`, `System.ComponentModel.IBindingList.*`,
  `IJsonLineInfo.*` on `JContainer`/`JToken`). These now render as real explicit
  impls and form a compilation context; they still fail recompile because their
  reconstructed closures are missing **other** sibling interface members — a
  separate closure-membership defect, out of scope here. Previously they died
  earlier as `ContextFail` (mangled target name absent from output).

## Tests

- New explicit-impl **property** coverage in `CSharpDeclarationWriterTests`
  (getter, getter+setter, keyword-segment escaping) — this shape had zero
  coverage before, which is why the bug went unnoticed. `ILInspector.CSharp.Tests`
  256/256 pass; the two pre-existing compat-display tests
  (`ExplicitPropertyDeclaration_KeepsCompatibilitySignature`,
  `ExplicitEventDeclaration_KeepsCompatibilitySignature`) stay green via the
  `Signature`-present gate.
- Consumer suites: `dotnet-inspect.Tests` 1925/0, `ILInspector.Decompiler.Tests`
  2663/0 — no product API-surface display or decompiler-output changes.
- #2776 RTS parity gate: exit 0, burn-down unchanged (7 rows).

## Scope

Explicit-interface-implementation **properties** only. The remaining Newtonsoft
CS0535 cases (e.g. `XDocumentWrapper::get_ChildNodes`) are an implicit-impl
closure-membership defect (a dropped sibling interface member), unrelated to name
mangling, and are left for separate work.

## Adversarial review

Requesting two reviews (GPT-5.5 + Gemini 3.1 Pro; different family than the
Claude author) in isolated worktrees.

### DIFF
--- a/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
@@ -418,6 +418,79 @@ public sealed class CSharpDeclarationWriterTests
     }
 
     [Fact]
+    public void PropertyDeclaration_WithDottedName_RendersAsExplicitInterfaceImplementation()
+    {
+        var type = new ApiType { Namespace = "Samples", Name = "Reader", Kind = "class" };
+        var member = new ApiMember
+        {
+            Name = "Some.Text.ILineInfo.LineNumber",
+            Kind = "property",
+            SignatureModel = new ApiSignature
+            {
+                ReturnType = "int",
+                MemberName = "Some.Text.ILineInfo.LineNumber",
+                Accessors =
+                [
+                    new ApiAccessor { Kind = "get" }
+                ]
+            }
+        };
+
+        var declaration = CSharpDeclarationWriter.RenderMemberDeclaration(type, member);
+
+        Assert.Equal("int Some.Text.ILineInfo.LineNumber { get; }", declaration);
+    }
+
+    [Fact]
+    public void PropertyDeclaration_WithDottedName_RendersGetterAndSetterWithoutModifier()
+    {
+        var type = new ApiType { Namespace = "Samples", Name = "Reader", Kind = "class" };
+        var member = new ApiMember
+        {
+            Name = "Some.Text.IValue.Payload",
+            Kind = "property",
+            SignatureModel = new ApiSignature
+            {
+                ReturnType = "string",
+                MemberName = "Some.Text.IValue.Payload",
+                Accessors =
+                [
+                    new ApiAccessor { Kind = "get" },
+                    new ApiAccessor { Kind = "set" }
+                ]
+            }
+        };
+
+        var declaration = CSharpDeclarationWriter.RenderMemberDeclaration(type, member);
+
+        Assert.Equal("string Some.Text.IValue.Payload { get; set; }", declaration);
+    }
+
+    [Fact]
+    public void PropertyDeclaration_WithDottedName_EscapesKeywordSegments()
+    {
+        var type = new ApiType { Namespace = "Samples", Name = "Reader", Kind = "class" };
+        var member = new ApiMember
+        {
+            Name = "event.class.Value",
+            Kind = "property",
+            SignatureModel = new ApiSignature
+            {
+                ReturnType = "int",
+                MemberName = "event.class.Value",
+                Accessors =
+                [
+                    new ApiAccessor { Kind = "get" }
+                ]
+            }
+        };
+
+        var declaration = CSharpDeclarationWriter.RenderMemberDeclaration(type, member);
+
+        Assert.Equal("int @event.@class.Value { get; }", declaration);
+    }
+
+    [Fact]
     public void IndexerDeclaration_EscapesStructuredKeywordParameterName()
     {
         var type = new ApiType { Namespace = "Samples", Name = "Values", Kind = "class" };
--- a/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
@@ -552,6 +552,38 @@ public sealed class CSharpDeclarationWriterTests
     }
 
     [Fact]
+    public void IndexerDeclaration_WithDottedName_IsNotTreatedAsExplicitInterfaceProperty()
+    {
+        // Explicit-interface indexers carry SignatureModel.MemberName "this[]" (the interface
+        // prefix is dropped by the reconstruction). Even if a stray dotted member.Name leaks
+        // in, the "this[]" model name must win: render an ordinary indexer, never route it
+        // through the dotted explicit-impl property branch (which would drop the parameters).
+        var type = new ApiType { Namespace = "Samples", Name = "Values", Kind = "class" };
+        var member = new ApiMember
+        {
+            Name = "Some.Text.IList.Item",
+            Kind = "property",
+            SignatureModel = new ApiSignature
+            {
+                ReturnType = "string",
+                MemberName = "this[]",
+                Parameters =
+                [
+                    new ApiParameter { Type = "int", Name = "index" }
+                ],
+                Accessors =
+                [
+                    new ApiAccessor { Kind = "get" }
+                ]
+            }
+        };
+
+        var declaration = CSharpDeclarationWriter.RenderMemberDeclaration(type, member);
+
+        Assert.Equal("public string this[int index] { get; }", declaration);
+    }
+
+    [Fact]
     public void ExplicitPropertyDeclaration_KeepsCompatibilitySignature()
     {
         var type = new ApiType { Namespace = "Samples", Name = "Values", Kind = "class" };
--- a/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
@@ -425,6 +425,7 @@ public sealed class CSharpDeclarationWriterTests
         {
             Name = "Some.Text.ILineInfo.LineNumber",
             Kind = "property",
+            IsExplicitInterfaceImplementation = true,
             SignatureModel = new ApiSignature
             {
                 ReturnType = "int",
@@ -449,6 +450,7 @@ public sealed class CSharpDeclarationWriterTests
         {
             Name = "Some.Text.IValue.Payload",
             Kind = "property",
+            IsExplicitInterfaceImplementation = true,
             SignatureModel = new ApiSignature
             {
                 ReturnType = "string",
@@ -474,6 +476,7 @@ public sealed class CSharpDeclarationWriterTests
         {
             Name = "event.class.Value",
             Kind = "property",
+            IsExplicitInterfaceImplementation = true,
             SignatureModel = new ApiSignature
             {
                 ReturnType = "int",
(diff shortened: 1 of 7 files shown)


richlander:
## Adversarial review round 1

Two reviewers from a different model family than the author (Claude): **GPT-5.5**
and **Gemini 3.1 Pro**, each in an isolated worktree.

### Where they agreed

Both independently confirmed **one BLOCKING issue**: explicit-interface
**indexers** regressed. RTS maps an indexer's member name to `this[]` (the
interface prefix is dropped), so the original harness change — preserving the raw
dotted `property.Name` for all properties — desynced `member.Name` from
`SignatureModel.MemberName`. The seam's ordinary-property branch rejects the
dotted `member.Name` and the explicit-impl branch rejects `this[]`, so rendering
fell through to a malformed `public int` (no name, no parameters, no accessors).
Before the PR these rendered as a compilable-but-wrong `public int this[int i] { get; }`.

Both also flagged the **`"event"` arm** of the gate (non-blocking): the gate was
broadened to include `"event"` but no structural event branch exists, so a
dotted-name event would have its modifier suppressed and its name dropped
(empty string). Unreachable today — RTS emits no events and the product always
sets a compatibility `Signature` — but latent.

Both **falsified** the other three attack points: the gate is not over-broad for
ordinary members (metadata names never contain `.`), keyword-segment escaping is
applied reliably (`EscapeQualifiedKeywordSegments`), and the `OmitPropertyAccessors`
branch is correct for its callers.

### Fixes (commit 2b2c50e2)

- **Indexer**: the harness now only preserves dotted names for **non-indexer**
  properties (`PropertyMemberName(name, isIndexer)`); indexers stay sanitized,
  exactly matching prior behavior. As defense-in-depth, the seam's
  ordinary-property branch now renders a `this[]` model as a proper indexer
  regardless of `member.Name`, so it can never erase a member. New regression
  test `IndexerDeclaration_WithDottedName_IsNotTreatedAsExplicitInterfaceProperty`.
- **Event arm**: removed `"event"` from `isExplicitInterfaceMember`, scoping the
  change precisely to properties.

Re-verified: seam suite 273/273; Newtonsoft `--return-to-sender` unchanged at
Exact 539 (+4 vs base), 0 regressions. Re-review of the fixed head requested from
both reviewers.

richlander:
## Adversarial review round 2 — both clean

Both reviewers re-reviewed the fixed head (`2b2c50e2`) in their isolated
worktrees:

- **GPT-5.5**: clean. Indexer repro now renders `public int this[int index] { get; }`
  (sanitized and dotted-name inputs alike); `ExplicitEventDeclaration_KeepsCompatibilitySignature`
  still passes; seam suite 273/273; no new issue from the broadened `this[]` guard.
- **Gemini 3.1 Pro**: clean. Both findings resolved; verified that binding the
  indexer render to `model.MemberName == "this[]"` means a leaked interface prefix
  in `member.Name` cannot pollute the indexer's structural format; seam suite
  273/273; no new defects.

Both blocking findings neutralized with backing tests. No outstanding review
guidance.

**Ready to merge** — awaiting maintainer approval (no self-merge).

richlander:

**Blocking findings at exact head `a4e0b91a3f0c13b8deb276570d47e65ba809f980`.**

- **GPT-5.5:** found that compiler-produced explicit indexers lose their interface qualifier/parameters and that static explicit interface properties drop the required `static` modifier.
- **Gemini 3.1 Pro:** confirmed both findings on a fresh isolated rerun and also identified the underlying discriminator problem: a dot in an untrusted metadata name is not typed evidence of explicit implementation. The first Gemini pass was discarded after it left a repro file in its checkout; only the clean rerun counts.

I reproduced the product outputs independently on the exact head with RTS-shaped structured models:

```text
indexer: public int
static: int N.I<N.C>.P { get; }
dotted ordinary: int A.B { get; }
```

A compiled fixture confirms the real metadata shapes `IIndex.Item` and static `IStatic<StaticImpl>.Value`. The current indexer path replaces `SignatureModel.MemberName` with `this[]`, so the new dotted-name branch cannot recover the interface qualifier and falls back to the bare return type. The explicit-member modifier branch also strips `static`, making static abstract interface implementations fail CS0539/CS0535. Finally, `Contains('.')` misclassifies legal untrusted dotted ordinary metadata names; the repository already has typed MethodImpl evidence and should not infer identity from display text.

**Required:** derive explicit-interface identity from the accessor MethodImpl relationship; preserve the typed interface qualifier separately when rendering `Interface.this[...]`; retain required non-access modifiers such as `static`; add compiler-produced indexer/static fixtures plus a dotted ordinary close negative; then re-review the fixed exact head with both families.

Structured explicit-interface events are a pre-existing separate gap and are tracked in #2850; keep events out of this property-scoped heuristic unless that capability is implemented.

richlander:
## Adversarial review of fixed head

**Still blocked at exact head `2b2c50e22f34d36008d16158112d7fe4a395990e`.**

Both fresh isolated reviewers agree that the follow-up correctly neutralizes the introduced indexer regression and removes the unsupported event arm:

- **GPT-5.5:** explicit indexers are back to the pre-existing compilable-but-semantically-wrong ordinary-indexer behavior; no new regression if the PR scope explicitly excludes indexers.
- **Gemini 3.1 Pro:** independently confirmed the indexer fallback and event-arm changes.

Both reviewers also independently reproduced two remaining blockers:

1. **Static explicit interface properties lose `static`.** RTS carries `IsStatic` from the accessor, but `CSharpDeclarationWriter` emits `static` only inside the non-explicit modifier branch. A compiler-produced static explicit property therefore renders as `int I.P { get; }`; that fails CS0539/CS0535. It must render `static int I.P { get; }`.
2. **Dotted text is not explicit-interface identity.** ECMA/SRM metadata does not restrict `Property.Name` to C# identifiers. An ordinary untrusted/foreign-language property named `A.B` is valid metadata, while explicit implementation identity is represented by the accessor MethodImpl relationship. The current `Contains('.')` heuristic strips ordinary-member accessibility and emits invalid explicit-interface syntax. The new tests cover only dotted positives and an indexer negative; there is no no-MethodImpl dotted close negative.

The compatibility-`Signature` display behavior raised by one reviewer is unchanged and intentionally display-only, so it is not counted as a PR regression.

**Required:** use accessor MethodImpl evidence to classify explicit properties, preserve `static` while suppressing accessibility, add compiler-produced static-property coverage and a dotted ordinary metadata close negative, narrow the stated scope to non-indexer properties, then run both fixed-head reviews again.

richlander:
## Review isolation correction

The fixed-head Gemini pass left generated tests in its checkout and is discarded from the counted review set.

The two isolation-clean reviews of `2b2c50e2` are now:

- **GPT-5.5:** blocking on both the dropped `static` modifier and dotted-name classification without MethodImpl evidence.
- **MAI-Code-1 Flash:** independently confirms the dropped-`static` blocker; checkout remained clean.

I independently reproduced both GPT findings against the exact head. The blocking disposition and required fixes in the preceding comment are unchanged; only the reviewer attribution is corrected.

richlander:
## Correction — retracting the earlier "round 2 — both clean"

My earlier "round 2 — both clean" comment was premature and is superseded by the
more rigorous concurrent review above. That pass reproduced two blocking findings
against `2b2c50e2` that my two reviewers missed:

1. Static explicit-interface properties drop the required `static` modifier
   (CS0539/CS0535).
2. `Contains('.')` is not typed evidence of explicit implementation; identity
   should be derived from the accessor MethodImpl relationship, the interface
   qualifier preserved so explicit indexers render as `Interface.this[...]`, and a
   dotted-ordinary close negative added.

I'm treating the PR as blocked and implementing the full typed-evidence fix
(MethodImpl-based classification, `static` preservation, explicit-indexer
qualifier rendering, compiled fixtures + negative). I'll re-request two clean
fixed-head reviews once it lands. Apologies for the premature signal.

richlander:
Closing as superseded. Issue #2827 was resolved and closed by the merged PR #2858 ("Render explicit interface properties in RTS shells"), which landed a stronger solution than this PR:

- Models explicit-interface properties as first-class `Kind == "explicit-interface-implementation"` structured members (this PR modeled them as `Kind == "property"` plus an `IsExplicitInterfaceImplementation` flag).
- Renders qualified explicit-interface indexers as `IInterface.this[...]` in the seam.
- Additionally solves the interface-projection recompile problem this PR explicitly deferred (#2777/#2780): it projects the required concrete interface members so the minimized RTS shell actually recompiles. Measured on the pinned corpus with zero reverse transitions — Newtonsoft CS0535 7 -> 0 / Exact 535 -> 544, ApplicationInsights ContextFail 26 -> 0, CodeAnalysis CS0535 22 -> 0.

This PR's canaries could only assert declaration shape (not full recompile) precisely because it did not solve projection; #2858 does. Retaining this branch would regress #2858's cleaner first-class model, so there is nothing to salvage here.

One follow-up worth a separate tracked issue: the shipped modifier path for `Kind == "explicit-interface-implementation"` omits `static`, so a C# 11 static-abstract interface member implemented explicitly (`static int I.Count => ...`) would render without `static`. #2858's corpus did not exercise this shape, so it is a latent (unmeasured) gap rather than a regression. Filing separately if it reproduces.

PR #2854, opened 2026-07-18
Fix multi-member record with-expression folding

### LINKED ISSUE #2829: Decompiler: multi-member record with partially raises and leaves invalid init-only stores at Full
## Defect

`WithExpressionPass` folds only the first mutation from a compiler-produced
multi-member record `with` expression. Roslyn emits `dup` between member
setters; the importer models that as a stack-slot copy, which stops the pass's
consecutive-member collector. The pass raises the prefix and leaves later
stores behind:

```csharp
public record RA(int A, int B);

public static RA M(RA r) => r with { A = 1, B = 2 };
```

Decompiled output:

```csharp
RA S_257 = r with { A = 1 };
S_257.B = 2;
return S_257;
```

The method is graded `Full`, but the trailing assignment fails with `CS8852`
because positional record properties are init-only.

This is a compiler-produced near miss for the synthetic
`MultipleContiguousMutations_RaiseInSourceOrder` test: that test puts both
stores on one slot and omits Roslyn's interposed copy.

## Reproduction

Verified on `b0e1bc11058255c54fc942f0220874e93b707cbe` with a .NET 11 Preview 7
Release build of:

```csharp
public record RA(int A, int B)
{
    public int C;
    public int D { get; set; }
}

public static class Probe
{
    public static RA Both(RA r) => r with { A = 1, B = 2 };
}
```

Product output:

```bash
dotnet run --project src/dotnet-inspect -c Release -- \
  member 'Probe.Both' \
  --library probes.dll \
  -S 'Decompiled Source'
```

Harness dump:

```text
StoreStackSlot S_257
  WithExpression (1 members)
    LoadArgument 0 (RA r)
    Constant 1 (int)
StoreProperty RA.B
  LoadStackSlot S_257
  Constant 2 (int)
Return
  LoadStackSlot S_257
// fidelity: Full
```

Compiling the emitted body reports:

```text
CS8852: Init-only property or indexer 'RA.B' can only be assigned in an
object initializer, or on 'this' or 'base' in an instance constructor or an
'init' accessor.
```

The pre-pass shape is:

```text
S_256 = r.<Clone>$()
S_256.A = 1
S_257 = S_256
S_257.B = 2
return S_257
```

## Likely owning seam

`WithExpressionPass.TryBuild` collects only member stores that are consecutive
and target `seed.Slot`; it stops at the first
`StoreStackSlot(next) = LoadStackSlot(previous)` copy.

The narrow fix should either chase compiler-produced stack-slot copies while
collecting subsequent member stores, as the sibling object-initializer path
does, or decline the raise whenever any mutation on the copied clone would
remain outside the `with`. It must never emit a partial `with` followed by an
init-only assignment at `Full`.

## Done signal

- `r with { A = 1, B = 2 }` emits one complete `with` expression or honestly
  degrades; no trailing init-only store remains.
- Add a compiled Roslyn fixture that includes the interposed-copy shape.
- Existing single-member `with` positives remain raised.
- Multi-member field/settable-property cases remain valid and preserve source
  order and side-effect order.

## Duplicate check

No open or closed issue matched `WithExpressionPass`, `CS8852`, or this
multi-member record shape. #1630 covered the earlier failure to raise record
clones and the invalid `<Clone>$` spelling; it did not cover partial
multi-member folding.

Discovered by the recurring over-raise audit tracked by #2820 under #2814.

richlander:
## Summary

Fixes #2829 by folding Roslyn's stack-slot copy chain into one complete
multi-member record `with` expression. The pass now proves ownership across
aliases before consuming member stores, so it no longer emits a partial
initializer followed by an invalid init-only assignment.

The widening remains conservative: duplicate members, external stores or
uses, malformed chains, multiple or missing escapes, and non-clone paths all
decline the raise. Record structs remain outside this clone-based path.

### Before

```csharp
RA S_257 = r with { A = 1 };
S_257.B = 2;
return S_257;
```

The trailing assignment fails with `CS8852` because `B` is init-only.

### After

```csharp
return r with { A = 1, B = 2 };
```

The compiler-produced fixture now folds every member into one valid `with`
expression.

### Fully Raised

## Validation

- `WithExpressionPassTests`: 10 passed
- `LadderRung5GateTests`: 9 passed
- Full decompiler suite passed

## Adversarial review

- Gemini 3.1 Pro: clean on exact head `4e388454`
- Claude Opus 4.8: clean on exact head `4e388454`

Both reviewers attacked alias ownership, escape accounting, malformed and
duplicate-member near misses, record class/struct boundaries, and unrelated
IR consumption. Neither found a blocking issue.

unchanged from the reviewed pre-rebase patch.

### DIFF
--- a/src/ILInspector.Decompiler.Fixtures.LadderRung5/LadderRung5.cs
+++ b/src/ILInspector.Decompiler.Fixtures.LadderRung5/LadderRung5.cs
@@ -106,6 +106,13 @@ public class Program
     {
         return point with { X = point.X + dx };
     }
+
+    // Multi-member nondestructive mutation over a positional record. Roslyn emits
+    // a dup between init-only setters for this shape.
+    public WithProbe MultiMemberWith(WithProbe r)
+    {
+        return r with { A = 1, B = 2 };
+    }
 }
 
 // Positional record with an extra init-only property: rung 5 "records/init where
@@ -116,6 +123,8 @@ public record Point(int X, int Y)
 {
     public int Magnitude { get; init; }
 }
+
+public record WithProbe(int A, int B);
 
 // C# 11 checked user-defined operators (#1706): a modern-syntax construct whose
 // USE must render with a checked(...) context, not an explicit op_Checked* method
--- a/src/ILInspector.Decompiler.Tests/LadderRung5GateTests.cs
+++ b/src/ILInspector.Decompiler.Tests/LadderRung5GateTests.cs
@@ -45,8 +45,8 @@ public class LadderRung5GateTests
     static readonly string[] ExpectedProgramMembers =
     [
         ".ctor", "Describe", "FromEnd", "IsOrigin", "IsRealString", "LastElement",
-        "MakeScaled", "MiddleSlice", "Prefix", "Quadrant", "Shift", "Size",
-        "UsingDeclaration",
+        "MakeScaled", "MiddleSlice", "MultiMemberWith", "Prefix", "Quadrant",
+        "Shift", "Size", "UsingDeclaration",
     ];
 
     // The record's compiler-synthesized member set (primary + copy ctor, Clone,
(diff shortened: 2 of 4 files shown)


richlander:
Ready to merge

PR #2855, opened 2026-07-18
Fix catch-entry local scope collisions

### LINKED ISSUE #2828: Decompiler: catch-entry folding reuses a live outer local and emits CS0136 at Full
## Defect

`EhStructuringPass` folds a catch-entry `StoreLocal(CaughtException)` into the
catch variable even when that local was declared before the `try` and remains
live after the handler. The decompiler then declares the same name twice:

```csharp
Exception V_1 = null;
try
{
    // ...
}
catch (Exception V_1)
{
}
return V_0 ?? ThrowReadingNotSupported(type, options, pgTypeId, V_1);
```

The method is graded `Full`, but the emitted C# fails to bind with `CS0136`.
The empty catch body also loses the source-level `inner = ex` assignment that
preserves the caught exception for the post-handler read.

## Real-world hits

Npgsql 10.0.3 (`lib/net10.0/Npgsql.dll`):

- `Npgsql.Internal.AdoSerializerHelpers::GetTypeInfoForReading`
- `Npgsql.Internal.AdoSerializerHelpers::GetTypeInfoForWriting`

The authored shape is:

```csharp
Exception? inner = null;
try
{
    typeInfo = options.GetTypeInfoInternal(type, pgTypeId);
}
catch (Exception ex)
{
    inner = ex;
}
return typeInfo ?? ThrowReadingNotSupported(..., inner);
```

## Reproduction

Verified on `b0e1bc11058255c54fc942f0220874e93b707cbe`:

```bash
  --validity-check \
  --package npgsql \
  --compile-cap all \
  --max-examples 10
```

The sweep reports:

```text
COMPILE-CHECK over 5468 rendered methods (4855 Full, 613 Partial)
CS0136: 2

Npgsql.Internal.AdoSerializerHelpers::GetTypeInfoForReading
  CS0136: A local or parameter named 'V_1' cannot be declared in this scope
Npgsql.Internal.AdoSerializerHelpers::GetTypeInfoForWriting
  CS0136: A local or parameter named 'V_1' cannot be declared in this scope
```

Targeted dump:

```bash
  --package npgsql \
  --dump 'Npgsql.Internal.AdoSerializerHelpers::GetTypeInfoForReading'
```

## Likely owning seam

`EhStructuringPass.FoldEntryConsumption` unconditionally detaches the leading
`StoreLocal { Value: CaughtException }` and returns that local as
`CatchClause.VariableIndex`. It needs to account for the local's declaration
and uses outside the handler.

The narrow fix should preserve a distinct catch variable plus assignment when
the destination local is live outside the handler, or degrade honestly if that
shape cannot be represented safely. Simple handler-local catch variables must
continue to fold normally.

## Done signal

- Both Npgsql methods no longer report `CS0136`.
- The caught exception still reaches the post-handler `Throw*NotSupported`
  call.
- Existing simple catch-variable positives remain raised and valid.
- Add a compiled positive/near-miss pair covering handler-local consumption
  versus assignment into a predeclared local that is read after the catch.

## Duplicate check

No existing open or closed issue matched `AdoSerializerHelpers`, catch-entry
folding, or this `CS0136` shape. #2275 covers raised-lambda name threading and
is a different seam.

richlander:
## Summary

Fixes #2828 by preserving a distinct catch variable when folding would collide
with a live outer local or an enclosing catch binding. The late
`CatchVariableScopePass` restores the caught-exception assignment after the
earlier folded shape has served using, foreach, and async consumers.

Sibling and cousin catches that independently bind the same slot are excluded
from escape detection, while real reads, writes, and managed-address uses
outside the binding still force rebinding.

### Before

Representative output reused the live outer local as the catch declaration:

```csharp
Exception V_1 = null;
try
{
    // ...
}
catch (Exception V_1)
{
}
return Use(V_1);
```

This fails with `CS0136` and loses the assignment needed by the post-handler
read.

### After

```csharp
Exception V_1 = null;
try
{
    // ...
}
catch (Exception V_2)
{
    V_1 = V_2;
}
return Use(V_1);
```

The distinct catch variable avoids the scope collision while preserving
Npgsql's caught exception for its later use.

### Fully Raised

## Validation

- `CatchEntryFoldingTests`: 7 passed on the rebased head
- Compiler-backed nested, sibling, cousin, and escaping-local cases passed

## Adversarial review

Earlier Gemini and Claude Opus review rounds exposed two additional scope
boundaries:

- nested catches sharing a slot, resolved by `8e5148de`
- sibling catches with real local uses, resolved by `4b349fe3`

`4b349fe3`.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CatchEntryFoldingTests.cs
+++ b/src/ILInspector.Decompiler.Tests/CatchEntryFoldingTests.cs
@@ -101,8 +101,18 @@ public class CatchEntryFoldingTests
     [Fact]
     public void SiblingCatchesSharingFoldedLocal_KeepTheirBindings()
     {
-        var specific = Catch(TypeRef.CoreLib("System", "InvalidOperationException"), 0);
-        var general = Catch(Exception, 0);
+        var specific = new CatchClause(
+            TypeRef.CoreLib("System", "InvalidOperationException"),
+            Body(new ExpressionStatement(new LoadLocal(0, Exception))))
+        {
+            VariableIndex = 0
+        };
+        var general = new CatchClause(
+            Exception,
+            Body(new ExpressionStatement(new LoadLocal(0, Exception))))
+        {
+            VariableIndex = 0
+        };
         var function = Function(new TryCatch(Body(), [specific, general]), Exception);
 
         new CatchVariableScopePass().Run(function, PassContext.None);
--- a/src/ILInspector.Decompiler.Tests/CatchEntryFoldingTests.cs
+++ b/src/ILInspector.Decompiler.Tests/CatchEntryFoldingTests.cs
@@ -1,0 +1,101 @@
+using ILInspector.Decompiler.Pipeline;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace ILInspector.Decompiler.Tests;
+
+// EhStructuringPass.FoldEntryConsumption folds a catch-entry `stloc` of the
+// caught exception into the clause variable — but only when that local is
+// confined to the handler. Issue #2828: when the destination local is declared
+// before the try and read after the catch, folding it emits a duplicate catch
+// variable (CS0136) and drops the `local = ex` assignment. Because the folded
+// shape is load-bearing for the using/foreach/async scaffold passes, the fix is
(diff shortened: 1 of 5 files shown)


richlander:
Ready to merge

richlander:
Non-blocking additional validation: the full decompiler suite completed successfully (2,667 passed; 0 failed).

PR #2856, opened 2026-07-18
Fix unconstrained generic pattern extraction

### LINKED ISSUE #2831: Decompiler: generic pattern extraction renders invalid as T through UnboxAny at Full
## Defect

An unconstrained generic declaration pattern such as `value is T t` lowers to
a type test plus value extraction. The decompiler renders the extraction's
`UnboxAny(IsInstance(...))` tree through `as T`:

```csharp
TSubject V_1 = Subject;
if ((V_1) is T)
{
    Console.WriteLine((T)((V_1) as T));
}
```

The method is graded `Full`, but this fails with `CS0413`: unconstrained type
parameter `T` cannot be used with `as`.

## Minimal compiler-produced repro

```csharp
public class Repro<T, TSubject>
{
    public TSubject Subject { get; set; }

    public void M()
    {
        if (Subject is T t)
        {
            System.Console.WriteLine(t);
        }
    }
}
```

Final raised IR:

```text
IfStatement
  IsInstance T
    Box TSubject
      LoadLocal 1 (TSubject)
  Block
    Call Console.WriteLine
      Box T
        UnboxAny T
          IsInstance T
            Box TSubject
              LoadLocal 1 (TSubject)
// fidelity: Full
```

## Real-world hits

FluentAssertions 8.10.0 (`lib/net6.0/FluentAssertions.dll`) reports 17
`CS0413` rows at `Full`, including:

- `FluentAssertions.Numeric.NumericAssertionsBase<T,...>::Be`
- `BeGreaterThan`
- `BeGreaterThanOrEqualTo`
- `BeInRange`
- `BeLessThan`
- `BeLessThanOrEqualTo`
- `Match`
- `NotBe`
- `FluentAssertions.Primitives.ReferenceTypeAssertions<T,...>::BeAssignableTo`
- `BeOfType`

## Reproduction

Verified on `b0e1bc11058255c54fc942f0220874e93b707cbe`.

Minimal fixture:

```bash
dotnet build repro.cs -c Release -p:OutputType=Library -o /tmp/repro
  /tmp/repro/repro.dll \
  --validity-check \
  --compile-cap all \
  --max-examples 20
```

Result:

```text
COMPILE-CHECK over 4 rendered methods (4 Full, 0 Partial)
Repro`2::M
  CS0413: The type parameter 'T' cannot be used with the 'as' operator
```

Real package:

```bash
  --package fluentassertions \
  --validity-check \
  --compile-cap all \
  --max-examples 30
```

Result:

```text
COMPILE-CHECK over 3088 rendered methods (2511 Full, 577 Partial)
CS0413: 17
```

## Likely owning seam

`CSharpPrinter.UnboxAnyOperand` falls through to
`Operand(unbox.Operand)`. For this shape, that operand is `IsInstance T`, and
the general `IsInstance` expression printer chooses `as T` because an
unconstrained generic is not known to be a value type.

The narrow fix should recognize the exact same-target
`UnboxAny(IsInstance(...))` test/extraction shape and emit valid, faithful C#
without routing the extraction through `as T`. Recovering the declaration
pattern binding is preferable where its control-flow proof is available; an
explicit object bridge may be appropriate only where it preserves the original
test/extract semantics. Close negatives must cover class-constrained generics,
concrete reference/value types, and the same expression outside a proven
successful pattern branch.

## Done signal

- The minimal declaration-pattern repro binds at `Full`.
- FluentAssertions' 17 `CS0413` rows drop to zero.
- Concrete reference-type `isinst` expressions that legitimately use `as`
  remain unchanged.
- Generic null-test behavior fixed by #1752 remains valid.

## Duplicate check

#1752 fixed an `isinst T` result used directly in a null test
(`(value as T) is not null`). This is a distinct surviving path:
`UnboxAny(IsInstance(...))` used to extract a declaration-pattern value, owned
by `UnboxAnyOperand`.

richlander:
## Summary

Fixes #2831 by emitting the legal `(T)(object)value` bridge only when control
flow proves that an identical generic type test has succeeded. Unproven
`UnboxAny(IsInstance(...))` shapes remain unchanged rather than gaining an
unsafe cast.

The proof handles both structured and local flat-CFG guards. It rejects entry
block and alternate-predecessor cases, invalidates locals or arguments whose
managed address is taken, scopes scans to the immediate function body, and
checks intervening writes along the complete nested AST path.

### Before

```csharp
if (V_1 is T)
{
    Console.WriteLine((T)(V_1 as T));
}
```

The unconstrained `as T` fails with `CS0413`.

### After

```csharp
if (V_1 is T)
{
    Console.WriteLine((T)(object)V_1);
}
```

The compiler-produced declaration-pattern extraction now binds successfully.
Concrete reference-type `isinst` rendering and unproven generic shapes retain
their existing behavior.

### Fully Raised

The complete endpoint is to recover the declaration-pattern binding and use
that value directly:

```csharp
if (Subject is T t)
{
    Console.WriteLine(t);
}
```

The current object bridge is therefore a valid fallback, not the final raised
shape. #2862 tracks declaration-pattern recovery and safe removal of the
lowering-only subject temporary.

## Validation

- `GenericDeclarationPatternExtractionTests`: 21 passed in Release
- `GenericDeclarationPatternExtractionTests`: 21 passed in Debug
- Adjacent generic `isinst` and `unbox.any` tests passed

## Adversarial review

Earlier Gemini and Claude Opus rounds drove conservative hardening:

- flat-CFG entry/backedge and managed-address rejection in `dfee57d5`
- nested function and loop proof scoping in `a85e0b4b`
- complete nested-path write detection in `446bbeba`

`446bbeba`.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/GenericDeclarationPatternExtractionTests.cs
+++ b/src/ILInspector.Decompiler.Tests/GenericDeclarationPatternExtractionTests.cs
@@ -449,6 +449,162 @@ public class GenericDeclarationPatternExtractionTests
         Assert.DoesNotContain("(object)", output);
     }
 
+    [Fact]
+    public void UnboxOfIsInstance_WithAddressTakenInlineLocalFunctionArgument_DoesNotBridge()
+    {
+        var objectType = TypeRef.CoreLib("System", "Object");
+        var generic = TypeRef.GenericParameter(0, "T");
+        IrExpression ReadArgument() => new LoadArgument(0, "value", objectType);
+
+        var thenBlock = new Block();
+        thenBlock.Add(new InitObject(
+            objectType,
+            new LoadArgumentAddress(0, "value", objectType)));
+        thenBlock.Add(new Return(new Box(
+            generic,
+            new UnboxAny(generic, new IsInstance(generic, ReadArgument())))));
+
+        var localFunctionBlock = new Block();
+        localFunctionBlock.Add(new IfStatement(
+            new IsInstance(generic, ReadArgument()),
+            thenBlock,
+            null));
+        var localFunctionBody = new BlockContainer();
+        localFunctionBody.Add(localFunctionBlock);
+        var localFunction = new LocalFunctionStatement(
+            "Local",
+            objectType,
+            [new Parameter("value", objectType)],
+            isStatic: true,
+            [],
+            [],
+            usesUpdatedMemorySafetyRules: false,
+            skipLocalsInit: false,
+            localFunctionBody);
+
+        var outer = new Block();
+        outer.Add(localFunction);
+        var body = new BlockContainer();
+        body.Add(outer);
+        var function = new IrFunction(
+            "M",
+            TypeRef.Definition("Synthetic", "Samples", "Owner"),
+            new MethodSignature(
+                TypeRef.CoreLib("System", "Void"),
+                [],
+                HasThis: false,
+                GenericParameterCount: 1),
+            [],
+            body);
+
+        var output = CSharpPrinter.Print(function).Output!;
+
+        Assert.DoesNotContain("(object)", output);
+    }
+
+    [Fact]
+    public void UnboxOfIsInstance_OuterAddressDoesNotBlockInlineLocalFunction()
+    {
+        var objectType = TypeRef.CoreLib("System", "Object");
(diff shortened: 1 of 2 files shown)


richlander:
Ready to merge

PR #2857, opened 2026-07-18
Preserve live targets when raising for loops

### LINKED ISSUE #2830: Decompiler: ForLoopPass erases try-continue leave target and emits CS0159 at Full
## Defect

`ForLoopPass` can absorb a loop's increment block into a `for` header while a
`Leave` inside a `try` still targets that block. The target disappears, but the
surviving `Leave` renders as a `goto`:

```csharp
for (int i = 0; i < 10; i++)
{
    if (x == 1)
    {
        try
        {
            goto IL_0010; // leave
        }
        catch
        {
            throw;
        }
    }
    x = 2;
}
```

The method is graded `Full`, but the emitted C# fails with `CS0159` because
there is no `IL_0010` label.

## Minimal compiler-produced repro

```csharp
public class Repro
{
    public void M(int x)
    {
        for (int i = 0; i < 10; i++)
        {
            if (x == 1)
            {
                try
                {
                    continue;
                }
                catch
                {
                    throw;
                }
            }

            x = 2;
        }
    }
}
```

Roslyn lowers the `continue` inside the protected region to a `leave` targeting
the increment block. After structuring, `ForLoopPass` moves that block into the
`for` increment slot but does not consume or reject the `Leave`.

## Real-world hit

`xunit.extensibility.execution` 2.9.3
(`lib/netstandard1.1/xunit.execution.dotnet.dll`):

- `Xunit.Sdk.ReflectionAttributeInfo::Instantiate`

## Reproduction

Verified on `b0e1bc11058255c54fc942f0220874e93b707cbe`.

Minimal fixture:

```bash
dotnet build repro.cs -c Release -p:OutputType=Library -o /tmp/repro
  /tmp/repro/repro.dll \
  --validity-check \
  --compile-cap all \
  --max-examples 20
```

Result:

```text
COMPILE-CHECK over 2 rendered methods (2 Full, 0 Partial)
CS0159: 1

Repro::M
  CS0159: No such label 'IL_0010' within the scope of the goto statement
```

Real package:

```bash
  --package xunit.extensibility.execution \
  --validity-check \
  --compile-cap all \
  --max-examples 20
```

Result:

```text
Xunit.Sdk.ReflectionAttributeInfo::Instantiate
  CS0159: No such label 'IL_0164' within the scope of the goto statement
```

## Likely owning seam

`ForLoopPass` guards only against existing `Continue` descendants before it
detaches the trailing increment. It does not account for `Leave`/branch nodes
whose target is that increment block.

The narrow fix should either:

- convert a protected-region `Leave` targeting the absorbed increment into a
  `Continue` whose C# semantics still execute the `for` increment and required
  EH cleanup; or
- decline the while-to-for raise when a live target would be erased.

Unrelated labels and leaves must remain untouched.

## Done signal

- The minimal `try { continue; }` repro emits valid C# at `Full`, or honestly
  declines/degrades without losing the target.
- The xUnit `Instantiate` `CS0159` row disappears.
- Ordinary for-loop reconstruction remains raised.
- Near misses with leaves to other targets retain their control flow.

## Duplicate check

#1748 is the closest sibling: a `catch { continue; }` in a `foreach` emitted a
missing-label `leave`. This issue is a distinct current `ForLoopPass` target
erasure: a `Leave` from inside `try` targets the increment block that the pass
absorbs. #1367 covered existing `Continue` nodes changing semantics when an
increment is hoisted; the current pass guard for that issue does not see this
`Leave` shape.

richlander:
## Summary

Fixes #2830 by declining while-to-for raising when the increment block still
has a live branch or protected-region `leave` target. This preserves the label
needed by the surviving `goto` instead of absorbing it into the `for` header.

Live targets are collected per nearest function, lambda, or local-function
scope, so coincident IL offsets in nested functions do not cause false
declines. Leaves to unrelated targets retain existing behavior.

### Authored source

```csharp
for (int i = 0; i < 10; i++)
{
    try
    {
        if (i == 5)
            continue;
        Console.WriteLine();
    }
    catch (Exception)
    {
    }

    Console.WriteLine();
}
```

The `continue` does not throw. The call inside the protected region can throw;
Roslyn lowers control flow that exits a protected region to `leave`, targeting
the loop's increment block.

### Before

The pass absorbed the increment block into a `for` header but left its
protected-region `leave` rendered as a `goto`:

```csharp
for (int i = 0; i < 10; i++)
{
    try
    {
        if (i == 5)
            goto IL_0010;
        Console.WriteLine();
    }
    catch (Exception)
    {
    }

    Console.WriteLine();
}
```

No `IL_0010` label survived, so the output failed with `CS0159`.

### After

The loop remains in its valid fallback form, preserving the targeted increment
block and label:

```csharp
int i = default;
i = 0;
while (i < 10)
{
    try
    {
        if (i == 5)
            goto IL_0010;
        Console.WriteLine();
    }
    catch (Exception)
    {
    }

    Console.WriteLine();

IL_0010:
    i++;
}
```

The three declined fixture outputs compile successfully with every emitted
`goto` resolving to a defined label.

### Fully Raised

The complete endpoint is to recover Roslyn's protected-region `leave` as the
authored `continue`, then safely raise the loop:

```csharp
for (int i = 0; i < 10; i++)
{
    try
    {
        if (i == 5)
            continue;
        Console.WriteLine();
    }
    catch (Exception)
    {
    }

    Console.WriteLine();
}
```

#2861 tracks the required slices:

1. prove and rewrite an EH-safe `leave` to a typed `Continue`;
2. re-enable `for` raising only after the increment target is structurally
   consumed.

The fallback adds a default-initialization opcode pair on compile-back, so the
known valid differences are explicitly docketed in both sugared and lowered
fidelity gates.

## Validation

- Focused `ForLoopPassTests` passed on the rebased head
- Sugared fidelity gate passed
- Lowered fidelity gate passed
- Standalone compile-back passed for all three declined fixtures

## Adversarial review

Earlier Gemini and Claude Opus rounds identified nested-function target-scope
collisions and the need to account for the valid fallback opcode differences:

- per-function target scoping was resolved by `75dfa937`
- sugared and lowered fidelity dockets were completed by `fc8269c6`

`fc8269c6`. Opus additionally compiled all three fallback outputs to confirm
that the fidelity docket does not conceal a recompile failure.


richlander:
Ready to merge

PR #2858, opened 2026-07-18
Render explicit interface properties in RTS shells

### LINKED ISSUE #2827: Render explicit-interface-implementation properties as real explicit impls in the type seam (RTS CS0535)

richlander:
## Summary

Render same-assembly explicit-interface properties as real C# explicit implementations in RTS shells, and project only the concrete properties required by minimized interface contracts.

- preserve qualified property names through the RTS data contract
- let `CSharpTypePrinter` own explicit-property declaration and body rendering
- render qualified explicit-interface indexers as `IInterface.this[...]`
- project matching explicit or implicit implementation properties without adding unrelated interface siblings
- keep external TypeRef interfaces outside this reconstruction path
- deduplicate projections by effective C# property shape

Closes #2827.

## Evidence

Pinned community and compiler assemblies improve without any reverse row transitions:

| Assembly | Exact | OpcodeDiff | RecompileFail | ContextFail | CS0535 |
| --- | ---: | ---: | ---: | ---: | ---: |
| Newtonsoft.Json 13.0.4 | 535 -> 544 | 16 -> 16 | 15 -> 8 | 15 -> 13 | 7 -> 0 |
| Microsoft.ApplicationInsights 2.23.0 | 522 -> 548 | 12 -> 12 | 17 -> 17 | 26 -> 0 | 4 -> 0 |
| Microsoft.CodeAnalysis 5.0.0 | 1084 -> 1188 | 13 -> 13 | 1431 -> 1431 | 104 -> 0 | 22 -> 0 |
| Microsoft.CodeAnalysis.CSharp 5.0.0 | 3600 -> 3601 | 3 -> 3 | 7384 -> 7384 | 37 -> 36 | 1 -> 0 |

Row-level comparison reports 9 improvements for Newtonsoft.Json, 26 for ApplicationInsights, and 104 for CodeAnalysis, with no successful row worsening.

The issue's historical 26-row estimate predates the current pinned corpus. The current assembly versions expose 34 `CS0535` rows across these four canaries; all 34 are removed. Six CodeAnalysis rows return to their pre-existing inherited-interface `CS1061` boundary rather than regressing to `CS0535`.

The cap-3 RTS parity result and lost-row set are identical on clean `origin/main` and this branch. Both currently report the same pre-existing `CSharpPrinter.CheckedSafeCast` manifest drift; this change adds no parity row.

## Validation

- Release solution build
- 2,666 decompiler tests
- 271 C# seam tests
- 118 focused RTS prototype tests
- cap-3 RTS parity comparison against clean `origin/main`

### DIFF
--- a/src/ILInspector.CSharp.Tests/CSharpTypePrinterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpTypePrinterTests.cs
@@ -1052,6 +1052,55 @@ public sealed class CSharpTypePrinterTests
     }
 
     [Fact]
+    public void ExplicitInterfacePropertyPreservesQualifiedNameAndOmitsAccessibility()
+    {
+        var property = new ApiMember
+        {
+            Name = "Samples.IValue.Value",
+            Kind = "explicit-interface-implementation",
+            SignatureModel = new ApiSignature
+            {
+                ReturnType = "int",
+                MemberName = "Samples.IValue.Value",
+                Accessors = [new ApiAccessor { Kind = "get" }]
+            }
+        };
+        var type = CreateEmptyType("Samples", "Widget");
+        type.Interfaces.Add("Samples.IValue");
+        type.Members.Add(property);
+        var request = new CSharpTypePrintRequest(
+            type,
+            memberPolicyOverrides:
+            [
+                new CSharpMemberPolicy(
+                    property,
+                    CSharpBodyPolicy.Full,
+                    new CSharpPropertyBody(
+                        CSharpAccessorBody.Block("return 42;"),
+                        null))
+            ]);
+
+        var result = _printer.Print(request);
+
+        Assert.Contains(
+            """
+                int Samples.IValue.Value
+                {
+                    get
+                    {
+                        return 42;
+                    }
+                }
+            """,
+            result.Units[0].Source,
+            StringComparison.Ordinal);
+        Assert.DoesNotContain(
+            "public int Samples.IValue.Value",
+            result.Units[0].Source,
+            StringComparison.Ordinal);
+    }
+
+    [Fact]
     public void NestedTypesAndPrimaryConstructorsRenderInFileScopedNamespaceUnits()
     {
         var nested = CreateEmptyType("Samples", "Nested");
--- a/src/ILInspector.CSharp.Tests/CSharpTypePrinterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpTypePrinterTests.cs
@@ -1101,6 +1101,56 @@ public sealed class CSharpTypePrinterTests
     }
 
     [Fact]
+    public void ExplicitInterfaceIndexerPreservesQualifierAndOmitsAccessibility()
+    {
+        var indexer = new ApiMember
+        {
+            Name = "Samples.IValues.Item",
+            Kind = "explicit-interface-implementation",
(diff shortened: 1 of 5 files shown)


richlander:
Adversarial review reconciliation for exact head `5cee6ee1`:

- **Claude Opus 4.8** found that explicit-interface indexers lost their qualifier and rendered as ordinary `this[...]`. I reproduced the failure, fixed C# seam rendering in `5cee6ee1`, and added both a seam test and a compiled RTS exact-roundtrip fixture. Its fixed-head reconciliation review is **CLEAN**.
- **Gemini 3.1 Pro** completed an independent fixed-head review with targeted probes for ordinary/qualified indexers, reserved-keyword escaping, accessors, deduplication, generic/external boundaries, and interface projection. The fixed-head review is **CLEAN**.
- The generic `TypeSpec` and external `TypeRef` concerns were verified as non-actions: `InterfaceSignatures` intentionally does not reconstruct those interfaces, `ExplicitInterfaceMemberName` mirrors that boundary to prevent `CS0540`, and excluded shapes have no head-vs-base regression.
- A pre-existing auto read-write explicit-property gap was classified as non-blocking: it already fails honestly on base and is outside this supported-shape change.

Both required model families have reviewed the fixed exact head with no outstanding blocking findings.

richlander:
Ready to merge

PR #2860, opened 2026-07-18
Move backing-field-reference extraction into MemberBodyFacts (#2777)

richlander:
## Summary

Lifts the last IR-walk reconstruction leak in the RTS planner into the product, extending the `MemberBodyFacts` consolidation. The harness's `TargetBackingFieldReadMembers`/`WriteMembers` previously walked Decompiler `LoadField`/`LoadFieldAddress`/`StoreField` IR and read the Decompiler `FieldRef` directly to synthesize auto-property/field member requirements.

## Change

- New neutral DTO `BackingFieldReference` (`DeclaringNamespace`, `DeclaringName`, `FieldName`, `BackingPropertyName`, `BackingFieldAccess`) in `ILInspector.CSharp` — no Decompiler type crosses the harness boundary.
- New producer `MemberBodyFacts.BackingFieldReferences(IrFunction)`:
  - Field loads (`ldfld`/`ldflda`) → `Read`; this-receiver instance stores → `InstanceWrite`; static stores → `StaticWrite`; other stores and non-definition declaring types are omitted (they could never match either consumer).
  - Declaring type is unwrapped to its definition and captured as namespace/name, so the harness self-type check becomes a plain string compare (`DeclaredBySelfType`, named distinctly from `IsSelfType(TypeRef, …)` so the reflection-by-name RTS prototype test stays unambiguous).
- Per request, each `MemberBodyFacts` member's XML doc now states its sort order (namespaces ordinal-sorted; constructor facts in chain/entry order; backing fields = all reads `ldfld`-then-`ldflda` in IR order, then writes in IR order — the order downstream first-wins de-dup relies on).

## Risk / neutrality

Pure boundary move — the IR walks are relocated verbatim behind the neutral DTO, so behavior is byte-neutral. The only remaining `IrFunction` use in the RTS path is the legitimate `CSharpPrinter.PrintRaised` render (product code).

## Evidence

- `MemberBodyFactsTests` 10 (4 new backing-field cases: instance read, instance write, static write, no-field).
- RTS oracle suite 224 green (Evidence/FixtureCatalog/Prototype/GeneratedFilter).
- Corpus rts-parity enforcing gate (14 assemblies, `--corpus-fidelity-cap 3 --compile-cap 0`) **exit 0**, 5 burndown rows unchanged (no resolved rows, no new failures).

## Review

Touches member-requirement synthesis, so not purely mechanical — requesting adversarial review (two models from a different family than my own).

### DIFF
--- a/src/ILInspector.CSharp/BackingFieldReference.cs
+++ b/src/ILInspector.CSharp/BackingFieldReference.cs
@@ -1,0 +1,46 @@
+namespace ILInspector.CSharp;
+
+/// <summary>
+/// How a decompiled member body touches a backing field, so ReturnToSender can decide
+/// which auto-property/field member requirements to synthesize without inspecting
+/// Decompiler IR itself.
+/// </summary>
+public enum BackingFieldAccess
+{
+    /// <summary>A field load (<c>ldfld</c>/<c>ldflda</c>) — the body reads the field.</summary>
+    Read,
+
+    /// <summary>An instance field store whose receiver is <c>this</c> (argument 0).</summary>
+    InstanceWrite,
+
+    /// <summary>A static field store.</summary>
+    StaticWrite,
+}
+
+/// <summary>
+/// A neutral, IR-free reference to a backing field a decompiled member body reads or
+/// writes. ReturnToSender uses these to reconstruct the auto-properties/fields the body
+/// depends on without reading Decompiler IR, so the IR-to-fact extraction lives in the
+/// product (ILInspector.Decompiler, which owns the IR) rather than in the harness. All
+/// fields are plain strings and an access enum; no Decompiler type crosses this boundary.
+/// </summary>
+/// <param name="DeclaringNamespace">
+/// Namespace of the field's declaring type (the definition, with any generic instance
+/// unwrapped), so the harness can match it against the target type's identity.
+/// </param>
+/// <param name="DeclaringName">
+/// Metadata name of the field's declaring type definition, matching the harness's
+/// self-type name form.
+/// </param>
+/// <param name="FieldName">The stored/loaded field's metadata name.</param>
+/// <param name="BackingPropertyName">
+/// The property name when the field backs an auto-property, or null when there is no
+/// such proof.
+/// </param>
+/// <param name="Access">Whether the reference is a read, an instance write, or a static write.</param>
+public sealed record BackingFieldReference(
+    string DeclaringNamespace,
+    string DeclaringName,
+    string FieldName,
+    string? BackingPropertyName,
+    BackingFieldAccess Access);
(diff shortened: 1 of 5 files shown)


richlander:

Two reviewers, each in an isolated worktree at the exact PR head (2e25d2a4), a different model family than the author (Claude Opus 4.8):

| Reviewer | Verdict |
| --- | --- |
| **GPT-5.6 Sol** | CLEAN |
| **Gemini 3.1 Pro** | CLEAN |

Both independently attacked the byte-neutrality claim and confirmed it with real runs — no findings, nothing to resolve. Full agreement, no divergence:

1. **Ordering / dedup** — the producer appends all `LoadField` then all `LoadFieldAddress` (Read) then `StoreField` (writes), so filtering by `Access` yields the exact per-consumer order the deleted `TargetBackingFieldRefs` produced; first-wins `seen` de-dup picks the same winners.
2. **Write predicate** — `{HasInstance:true, Instance:LoadArgument{Index:0}}` maps to `InstanceWrite`, `{HasInstance:false}` to `StaticWrite`, else omitted; consumer `Where` drops `StaticWrite` when `allowStaticStores=false`, and instance-store-to-non-`this` is dropped in the producer. The `field.Attributes.HasFlag(FieldAttributes.Static)` metadata read is untouched.
3. **Self-type check** — moving the non-Definition rejection into the producer's `Add` is neutral: those references failed the old `IsSelfType` and `continue`d before `seen.Add`, so dropping them earlier has no side effects. Nested/generic-instance identities preserved.
4. **Non-Definition edges** — GenericInstance-with-null-ElementType, array/pointer/byref declaring types dropped identically old vs new.
5. **Rename safety** — `IsSelfType(TypeRef, …)` still used for parameter checks; the new `DeclaredBySelfType` keeps the reflection `GetMethod("IsSelfType")` in `ReturnToSenderPrototypeTests.cs:4049` unambiguous.
6. **Sort-order docs** — the new "Order:" XML docs match `SortedSet` ordinal, chain/entry order, and IR-descendant order.

**Validation both ran:** slnx Release build clean; RTS oracle suite 224 green; corpus rts-parity gate exit 0 with the same 5 burndown rows unchanged (GPT reported 89,065 methods across 14 assemblies).

No fixes were required, so no follow-up commits and no re-review needed.

richlander:
Ready to merge — two CLEAN adversarial reviews (GPT-5.6 Sol, Gemini 3.1 Pro) on the exact head with no findings; byte-neutral by the corpus rts-parity gate. Awaiting your approval; I won't self-merge.

richlander:
Rebased onto latest origin/main (0 behind; picked up #2854/#2855/#2856 raise fixes). Re-verified on the rebased head 4125d8f9: slnx build clean, corpus rts-parity gate exit 0 with the same 5 burndown rows unchanged. Prior two CLEAN reviews were against the pre-rebase head, which differs only by the trivial rebase (no content conflicts). Still ready to merge; awaiting your approval.

PR #2870, opened 2026-07-18
Drain rts-parity row: dedup self-recursive target method in RTS closure (#2777)

richlander:
- Part of #2777 (make ReturnToSender a thin orchestrator; drain the rts-parity burn-down)
- Row: `ILInspector.MetadataPrimitives!ILInspector.Metadata.TypeResolver::GetTypeNameFromReference#0`
- Evidence revision: `44f22373`

## Fix

> Should we accept this row fix?

**Conclusion:** **PASS** — a self-recursive target method was re-emitted as a hollow `throw null` stub next to its own body-backed declaration (CS0111 duplicate member), forcing the compile-back floor; excluding the target method's own handle from closure member requirements makes it recompile Exact under ReturnToSender with no floor.

False output (RTS reconstruction, before):

```csharp
public static class TypeResolver
{
    public static string GetTypeNameFromReference(MetadataReader reader, TypeReferenceHandle handle)
    { /* real body */ }
    public static string GetTypeNameFromReference(MetadataReader reader, TypeReferenceHandle handle) { throw null; } // CS0111
    public static string GetFullName(string ns, string name) { throw null; }
}
```

Fixed output:

```csharp
public static class TypeResolver
{
    public static string GetTypeNameFromReference(MetadataReader reader, TypeReferenceHandle handle)
    { /* real body; recursive self-call binds here */ }
    public static string GetFullName(string ns, string name) { throw null; } // referenced sibling, still stubbed
}
```

## Root cause and scope

- **Root cause:** The shared closure walk (`SeedTypedClosureRoots`) turns every consumed member reference into a closure member requirement. Because the target method references a *sibling* on its own type (`GetFullName`), `allowTargetRoot` is set for the target root — which also let the method's *self-recursive* call become a closure requirement. That produced a second hollow stub of the target method. `AddRequiredMembers` dedups by exact signature **text**, and the target-with-body path and the closure path format signature text differently, so the duplicate was not deduped → CS0111.
- **Fix shape:** Thread the target `MethodDefinitionHandle` into `SeedTypedClosureRoots`; `AddSingleMethodFact` skips a `MethodRef` that resolves to the target method's own handle. Keyed on the method handle, so real overloads and sibling members (distinct handles) are still reconstructed, and the self-recursive call binds to the emitted body.
- **Scope boundary:** Harness-only (`tools/DecompilerHarness`); no product paths touched. Does not address the other four burn-down rows (distinct causes: enum-member surface, cross-assembly enum→int typing, type-as-namespace using synthesis, un-reconstructed closure member).

## Witnesses

| Witness | Before | After |
| --- | --- | --- |
| `TypeResolver::GetTypeNameFromReference#0` (real corpus row) | `RecompileFail: CS0111 duplicate member`, floor used | `Exact`, floor not used |
| `Class1::Describe` (compiled fixture, self-recursive + sibling call) | duplicate stub | one body-backed method + one sibling stub, `Exact`, no floor |

## Evidence

| Check | Result |
| --- | --- |
| Full `dotnet-inspect.slnx` build (Release) | Pass, 0 warnings |
| Focused test | `ReturnToSenderPrototypeTests.CompileBackTargets_DoesNotDuplicateSelfRecursiveTargetMethodDuringClosureSurface` passes |
| RTS oracle suites | Evidence 3 / FixtureCatalog 82 / Prototype 119 / GeneratedFilter 25 / MemberBodyFacts 10 — all green |
| Real witness | `TypeResolver::GetTypeNameFromReference` recompiles Exact via RTS without the compile-back floor |

## Corpus validity

> Did this row fix introduce new rts-parity regressions?

**Conclusion:** **PASS** — enforcing gate exit 0, no new regressions; the fixed row is flagged resolved and dropped from the manifest (5 → 4).

Run: 14-assembly prepared corpus (`eng/prepare-decompiler-corpus.sh`), `--corpus-fidelity-oracle rts-parity --corpus-fidelity-cap 3 --compile-cap 0`, enforcing `--rts-parity-burndown`.

| Metric | Before | After |
| --- | ---: | ---: |
| rts-parity burn-down rows | 5 | 4 |
| New Exact→RecompileFail regressions | - | 0 |

Manifest change: removed `ILInspector.MetadataPrimitives!ILInspector.Metadata.TypeResolver::GetTypeNameFromReference#0`.

- `SymbolPackageDownloader::CheckPdbHeader` — reconstructed enum missing member (CS0117).
- `LibraryBodyIndex::HollowUnsafeMethods` — cross-assembly enum→int typing (CS0266); product raise concern.
- `Base64Encoder::WriteCharsAsync` — using synthesis treats a type as a namespace (CS0138).
- `ParseResult::GetValue#2` — closure member not reconstructed (CS0234).

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| GPT-5.6 Sol | **CLEAN** | Built harness + tests; Prototype 119 / FixtureCatalog 82 / Evidence 3 / GeneratedFilter 25 green; corpus rts-parity gate exit 0 (4 rows); clean tree at head. |
| Gemini 3.1 Pro | **CLEAN** | Verified attack points with throwaway xUnit tests (recursive properties, accessor siblings, generics/instantiations, overloads); corpus gate exit 0 (4 rows). |

**Review conclusion:** CLEAN — two independent adversarial reviews (GPT-5.6 Sol, Gemini 3.1 Pro), neither my own model family, both CLEAN with no findings.

## Validation

```bash
# corpus enforcing gate (14 asm, cap 3): exit 0, 4 rows
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -4780,6 +4780,48 @@ public class ReturnToSenderPrototypeTests
         }
     }
 
+    [Fact]
+    public void CompileBackTargets_DoesNotDuplicateSelfRecursiveTargetMethodDuringClosureSurface()
+    {
+        // Guard for the rts-parity burndown row TypeResolver::GetTypeNameFromReference:
+        // a target method that calls itself recursively must not also be reconstructed
+        // as a hollow `throw null` closure stub (the self-reference resolves to the
+        // target method's own handle). Emitting both the body-backed method and a
+        // same-signature stub produced CS0111 (duplicate member) and forced the
+        // compile-back floor. A referenced sibling (Combine) must still be stubbed.
+        var assemblyPath = CompileFixture("""
+            public static class Class1
+            {
+                public static string Describe(int depth, string name)
+                {
+                    if (depth > 0)
+                        return Describe(depth - 1, name);
+                    return Combine(name, name);
+                }
+
+                public static string Combine(string left, string right) => left + right;
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("Class1", "Describe", 0)]));
+
+            Assert.Equal(FidelityCheck.CompileBackStatus.Exact, result.Status);
+            Assert.False(result.UsedCompileBackFloor, result.Detail);
+            // The recursive target keeps exactly one declaration (body-backed, not a
+            // hollow stub); the referenced sibling is the only `throw null` member.
+            Assert.Equal(1, result.Source.Split("string Describe(").Length - 1);
+            Assert.DoesNotContain("string Describe(int depth, string name) { throw null; }", result.Source);
+            Assert.Contains("public static string Combine(string left, string right) { throw null; }", result.Source);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
     static string CompileFixture(
         string source,
         string? directory = null,
--- a/tools/DecompilerHarness/ReturnToSenderTypePlanner.cs
+++ b/tools/DecompilerHarness/ReturnToSenderTypePlanner.cs
@@ -437,6 +437,7 @@ public static class CompileBackSourceComposer
             request.Reader,
             request.Function,
             request.TargetType,
+            request.TargetMethod,
             targetRoot,
             closure.Roots,
             closure.Facts,
@@ -456,6 +457,7 @@ public static class CompileBackSourceComposer
         MetadataReader reader,
         IrFunction function,
         TypeDefinitionHandle targetType,
+        MethodDefinitionHandle targetMethod,
         TypeDefinitionHandle targetRoot,
         HashSet<TypeDefinitionHandle> closureRoots,
         Dictionary<TypeDefinitionHandle, List<CompileBackFact>> closureFacts,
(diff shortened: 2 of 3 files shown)


richlander:

Two independent adversarial reviews were run against head `44f22373` in isolated worktrees (neither reviewer is my own model family, per the review policy). Both are **CLEAN** with no findings.

### GPT-5.6 Sol — **CLEAN**
Built `tools/DecompilerHarness` and `src/ILInspector.Decompiler.Tests` (Release), ran the RTS oracle suites (Prototype 119 / FixtureCatalog 82 / Evidence 3 / GeneratedFilter 25 — all green), and the corpus rts-parity enforcing gate over all 14 corpus assemblies (exit 0, exactly 4 burn-down rows remaining). Final tree clean at head `44f223730c214546e6fa3d4bc1155e7c06ff2145`.

### Gemini 3.1 Pro — **CLEAN**
Probed the over-/under-exclusion attack surface of `ResolvesToTargetMethod` / `TypeProducer.TryFindMethod` with throwaway `ReturnToSenderPrototypeTests`:
- **Recursive properties** — a `get_P` target calling itself correctly skips requirement generation (no property duplication).
- **Accessor siblings** — a `get_P` target calling `set_P` evaluates `false` in `ResolvesToTargetMethod`; the setter is emitted as a hollow stub alongside the real getter, `Exact`, no collision.
- **Generics & instantiations** — uninstantiated parameter comparison ignores sibling calls with differing type params (no over-exclusion) while still excluding recursive self-calls.
- **Overloads** — distinguished by parameter length + sequence equality; sibling stubs preserved.

Ran the corpus gate (exit 0, no regressions, 4 remaining rows).

### Agreement / divergence
Both reviewers agreed: the fix isolates the target method's own handle to prevent the duplicate `throw null` stub while preserving requirement generation for siblings and real overloads. No divergence; no blocking or non-blocking findings from either reviewer. No follow-up commits required.

richlander:
**Ready to merge** — two CLEAN adversarial reviews (GPT-5.6 Sol, Gemini 3.1 Pro), build green, RTS oracle suites green, corpus rts-parity gate exit 0 (row drained 5 → 4). Not self-merging; awaiting @richlander approval.

Issue #2862, opened 2026-07-18
Decompiler: recover generic declaration-pattern bindings


richlander:
## Current state

#2856 fixes the `CS0413` validity defect from #2831 by emitting a proven-safe
object bridge:

```csharp
TSubject V_1 = Subject;
if (V_1 is T)
{
    Console.WriteLine((T)(object)V_1);
}
```

This is valid and semantically faithful, but it still exposes the generic
test/extract lowering instead of recovering the declaration pattern.

## Fully raised goal

```csharp
if (Subject is T t)
{
    Console.WriteLine(t);
}
```

Keeping a single-evaluation temporary is also an acceptable intermediate shape
when it cannot yet be removed safely:

```csharp
TSubject V_1 = Subject;
if (V_1 is T t)
{
    Console.WriteLine(t);
}
```

## Required slices

1. **Recover the declaration-pattern binding**
   - Recognize the proven matching type test and its
     `UnboxAny(IsInstance(...))` extraction as one pattern operation.
   - Introduce a typed declaration-pattern variable in the owning `if`
     condition and replace dominated extraction uses with that binding.
   - Reuse the control-flow, write, alias, managed-address, and nested-function
     proof boundaries established by #2856.

2. **Remove a lowering-only subject temporary when safe**
   - Inline a single-use, side-effect-safe subject into the declaration
     pattern without evaluating it more than once.
   - Retain the temporary when it has additional uses, an address escape, or
     any evaluation-order risk.

## Evidence

Add compiler-produced positive and close-negative fixtures for:

- unconstrained and struct-constrained generic declaration patterns;
- negated guards and flat-CFG guard forms;
- multiple dominated uses of the bound value;
- intervening writes, alternate predecessors, loops, and backedges;
- managed-address and alias escapes;
- nested lambda/local-function scope boundaries;
- class-constrained and concrete-type patterns that already raise correctly.

Pin the raised C#, compile-back validity, control-flow semantics, Debug/Release
compiler shapes, and unchanged behavior for every unproven case.

## Related

- #2831 / #2856: valid object-bridge fallback after a proven generic type test
- #1752: generic `isinst` null-test spelling

richlander:
Completed by #2871 and #2873.

Analysis against current `main`:

[...]

PR #2871, opened 2026-07-18
Raise generic declaration-pattern bindings (#2862)

richlander:
- Fixes/advances #2862
- Changes generic (unconstrained / struct-constrained) declaration-pattern extraction to render as a real `if (subject is T t)` binding instead of the #2856 test/extract object bridge
- Card revision: `5093dae4`

## Change

**Conclusion:** **PASS** — the raise binds each dominated extraction to the exact guard that proves it (no cross-guard staleness, no nested-scope binding), leaving unprovable shapes on the #2856 bridge.

For an unconstrained or struct-constrained generic `T`, `T t = subject as T` is illegal (CS0413), so csc caches the subject, tests it with `isinst`, and re-tests+unboxes at each use (`UnboxAny T(IsInstance T(x))`). #2856 made that compile via an `(object)` bridge at print time; this PR raises the shape to a binding.

- **Slice 1 — `IsPatternPass.RaiseGenericDeclarationPattern`**: mints a pattern local, rewrites the guard to `x is T t`, and rewrites every dominated extraction to load `t`. Each extraction is proven against the exact guard being raised (`GenericDeclarationPatternProof.IsProvenBySpecificGuard`): the site must lie in the guard's `Then` arm with no reassignment of the tested value and no loop/nested-function boundary between the guard and the site. Only root-scope guards are raised (`DescendantsOutsideNestedFunctions`), since the pattern local is minted on the root function.
- **Slice 2 — `IsPatternPass.InlineGenericPatternSubject`**: folds the lowering-only subject temp into the pattern when it is single-use and the pattern is the guard's whole condition (order-safe), yielding `if (subject is T t)`.

Only the positive, structured shape is raised: `x is not T t` is illegal C# (CS8780), so negated/flat-CFG guards stay on the #2856 `(object)` bridge.

### Before

```csharp
// FluentAssertions NumericAssertionsBase`3::BePositive — #2856 bridge
TSubject V_1 = Subject;
if ((V_1) is T)
{
    subject = (T)(object)(V_1);
    V_2 = default;
}
else
{
    S_1 = false;
}
```

### After

```csharp
// this PR: guard raised to a binding, subject temp inlined
if (Subject is T V_3)
{
    subject = V_3;          // multi-use: copy into csc's cache local (see Fully Raised)
    V_2 = default;
}
else
{
    S_1 = false;
}
```

For a single-use binding the After is fully raised, e.g. `if (Subject is T V_2) { WriteLine(V_2); }`.

### Fully Raised

When the binding is read more than once, csc caches the single extraction in a source-named local, so the raise binds a synthesized `V_N` and leaves one redundant copy into that cache local. The fully raised endpoint binds the pattern directly to the cache local (recovering its name) and drops the copy:

```csharp
if (Subject is T subject)
{
    V_2 = default;
}
else
{
    S_1 = false;
}
```

- Required tracking issue: #2872 — bind a multi-use generic declaration pattern directly to the cache local and elide the `cacheLocal = V_N;` copy.

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass |
| Focused tests | `GenericDeclarationPatternExtractionTests` 27/27 passed |
| Decompiler fast suite | Pass (2558/0, minus `Speed=Slow`) |
| Reduced fixture validity | Pass — FluentAssertions `--compile-cap all` defect diff vs base `e3a88b09`: 0 regressed, 0 improved (2154 methods) |
| Reduced fixture fidelity | Render A/B (FluentAssertions, base vs head): 4676 methods, Changed 8 (structural 8), valid->valid 8, valid->invalid 0, Added/Removed 0 |
| Real witness | ``NumericAssertionsBase`3::BePositive`` raised (residual copy tracked in #2872) |

## Decompiler quality

**Conclusion:** **PASS** — pinned-subset PR quick gate is within tolerance (residue down 0.16 pp, conditional-branch flat, 0 pass bugs); aggregate rate movement is baseline drift from an unrelated self-assembly, not this PR.

### PR quick gate

Run: PR quick corpus, hash-stable 100 methods per assembly; 15 assemblies, 1,500 methods.

| --- | ---: | ---: | ---: |
| Detected lowering residue (-) | 17.83% | 17.67% | -0.16 pp |
| Pass bugs (-) | 0 | 0 | 0 |

> **Conclusion:** **PASS** — pinned-subset gate matched baseline tolerances.

### Aggregate context

Corpus: 15 assemblies, 1,500 methods. Baseline drift: pinned baseline (2026-06-26) drifted; `ILInspector.MetadataPrimitives` grew 61 -> 100 methods (+39), so aggregate count deltas below include that corpus change and are not a clean PR signal.

| --- | ---: | ---: | ---: |
| Detected lowering residue (-) | 248 (16.97%) | 252 (16.80%) | n/a (population differs) |
| Conditional-branch residue (-) | 35 (2.40%) | 47 (3.13%) | n/a (population differs) |
| Forward-merge stops (-) | 27 (1.85%) | 33 (2.20%) | n/a (population differs) |
| Pass bugs (-) | 0 | 0 | 0 |

> **Conclusion:** **ADVISORY** — aggregate conditional-branch (+0.73 pp) and forward-merge (+0.35 pp) movement is attributable to the drifted self-assembly population; rebaseline against current main before trusting count-based regressions.

## Validation

```bash
dotnet build dotnet-inspect.slnx -c Release --configfile ./nuget.config
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class ILInspector.Decompiler.Tests.GenericDeclarationPatternExtractionTests
```


richlander:

Both reviewers found a real HIGH-severity soundness bug. Both are fixed in `5093dae4`.

**GPT-5.5 — cross-guard divergence (HIGH, valid).** When the tested local is
reassigned between an outer guard (being raised) and an inner guard that actually
proves an extraction, the raise bound the outer (pre-write) value at a site that
reads the post-write value → wrong decompilation. The old `IsStableTestedValue`
(count ≤ 1) missed it: a default-initialized local written once mid-body counts as 1.
**Fix:** replaced it with `GenericDeclarationPatternProof.IsProvenBySpecificGuard`,
which proves each site against the *exact* guard being raised (no reassignment, loop,
or nested-function boundary between that guard and the site). The outer guard now
bails and the inner guard raises instead, reading the correct post-write value.
Regression test: `CrossGuardInterveningWrite_DoesNotRaiseStaleOuterBinding`.

**Gemini 3.1 Pro — nested-lambda scope crash (HIGH, valid).** Both slices walked
`function.Descendants`, which recurses into nested lambdas/local functions, then
minted the binding local on the *root* `IrFunction`. The printer scopes lambdas
with their own local pool, so a root-allocated index dangles and crashes `LocalName`.
**Fix:** restricted both slices to `DescendantsOutsideNestedFunctions`, so only
root-scope guards are raised; guards inside nested functions stay on the #2856
object-bridge. Regression test: `GuardInsideNestedLambda_IsNotRaised_AndDoesNotCrash`.

Both new tests fail on the pre-fix code and pass after.

**Evidence at fixed head `5093dae4`:**
- Decompiler suite 2558/0 (minus `Speed=Slow`).
- FluentAssertions render A/B vs base: 4676 methods, Changed 8 (structural 8),
  valid→valid 8, valid→invalid 0, Added/Removed 0 — unchanged from pre-fix.
- FluentAssertions validity defect diff vs base (`--compile-cap all`): 0 regressed,
  0 improved (2154 methods checked in both).

Gemini additionally confirmed the rest of the correctness invariants
(negated/flat CFG avoidance of CS8780, slice-2 aliasing/eval-order, box stripping,
stranding bail, proof-migration parity). Re-reviewing the fixed head with both
model families next.

richlander:
## Adversarial review round 2 — fixed head `5093dae4`

Both reviewers re-reviewed the exact fixed head in isolated worktrees and returned **clean** (no actionable findings). Both explicitly confirmed the two round-1 fixes hold.

**GPT-5.5** (build + 27/27 tests): cross-guard staleness fix holds (sites proven against the specific guard; unproven sites bail the whole rewrite); nested-lambda crash fix holds (neither slice traverses into nested lambda/local-function bodies, so no root-scoped pattern local is minted into a nested pool).

**Gemini 3.1 Pro** (build + tests + edge-case probes): both fixes hold. `IsProvenBySpecificGuard`'s parent walk fails the proof when an extraction site is inside a nested function (parent hits the Lambda/LocalFunction boundary), so `sites.All(...)` bails; `DescendantsOutsideNestedFunctions` cleanly skips nested guards. Cross-guard staleness is covered by `HasAddressTaken` (ref/out, closure display-class fields fail `IsReadOnlyOperand`), `HasInterveningWrite` (sibling/try-catch layout), and the loop-boundary bail. Also confirmed slice-2 single-eval/order safety, box stripping, stranding bail, and #2856 proof-migration parity.

Ready to merge.

Issue #2872, opened 2026-07-18
Decompiler: bind multi-use generic declaration pattern directly to the cache local (remaining slice after #2862)


richlander:
### Remaining slice after #2862

#2862 raises csc's lowered generic (unconstrained / struct-constrained)
declaration-pattern extraction to a real `if (subject is T t)` binding. When the
bound value is read **once**, the raise is fully clean:

```csharp
if (Subject is T V_2) { WriteLine(V_2); }
```

When the bound value is read **more than once**, csc caches the single `isinst`
extraction in a real (source-named) local, so the raise binds the pattern to a
synthesized local and leaves a redundant copy from it into that cache local:

```csharp
// After #2862 (not fully raised)
if (Subject is T V_2)
{
    value = V_2;          // redundant copy
    ... use value ...
}
```

### Fully raised endpoint

Bind the pattern directly to the reused cache local (recovering its source name)
and drop the redundant copy:

```csharp
if (Subject is T value)
{
    ... use value ...
}
```

### Slice

- When every dominated read of the pattern binding is a read of a single
  cache local that is assigned only from the pattern extraction, bind the
  pattern to that cache local (adopting its name) instead of minting a fresh
  `V_N`, and elide the `cacheLocal = V_N;` copy.

Real witness: `FluentAssertions.Numeric.NumericAssertionsBase\`3::BePositive`
(and the other `NumericAssertionsBase` guards) render `subject = V_3;` /
`value = V_2;` copies after #2862.

PR #2873, opened 2026-07-18
Bind multi-use generic declaration pattern to cache local (#2872)

richlander:
- Fixes/advances #2872
- Changes multi-use generic declaration-pattern raise to bind csc's source-named cache local and drop the redundant copy
- Card revision: `29c196b4`

## Change

**Conclusion:** **PASS** — re-pointing the pattern to the single-assignment cache local recovers the source name and removes the copy with no validity or fidelity change (FluentAssertions valid→valid 8, valid→invalid 0; defect diff 0 regressed).

### Original source

SourceLink-backed authored source for the witness member, retrieved via `dotnet-inspect member NumericAssertionsBase BePositive --package FluentAssertions -S "Original Source"` (FluentAssertions 8.10.0, net6.0):

```csharp
{
    CurrentAssertionChain

}
```

The author wrote `Subject is T subject && subject.CompareTo(default) > 0` — a single declaration pattern named `subject`, read twice in the guard. csc caches the `isinst` in the source-named local; this PR re-points the raised pattern to that local so the decompiled shape matches the authored `is T subject`.

### Before

```csharp
// FluentAssertions NumericAssertionsBase`3::BePositive (this PR's parent, #2871)
T subject;
T V_2;
bool S_1;
...
if (Subject is T V_3)
{
    subject = V_3;
    V_2 = default;
}
```

### After

```csharp
// output produced by this PR
T V_2;
bool S_1;
...
if (Subject is T subject)
{
    V_2 = default;
}
```

The pattern binds the cache local `subject` directly; the `subject = V_3;` copy and the up-front `T subject;` declaration are both gone. The minted `V_3` slot becomes unreferenced and produces no declaration.

### Fully Raised

This PR's slice — binding the pattern to csc's source-named cache local — is complete. The method as a whole is **not** yet fully raised: comparison against the authored source above shows three remaining, orthogonal residues (all pre-existing, none owned by this PR). The intended fully raised endpoint is the authored shape:

```csharp
CurrentAssertionChain
```

Remaining slices to reach that state:
- Short-circuit `&&` reconstruction — the `if (Subject is T subject) { S_1 = …; } else { S_1 = false; }` boolean diamond should fold to `Subject is T subject && subject.CompareTo(default) > 0`.
- `default` literal inlining — `V_2 = default; … CompareTo(V_2)` should inline to `CompareTo(default)`.
- Fluent chain / temp re-composition — `S_256`/`S_257`/`S_258` should collapse to the `.ForCondition(…).BecauseOf(…).FailWith(…, Subject)` chain.

- Required tracking issue: #2877 — pattern-guarded short-circuit `&&` diamond, `default` temp inlining, and fluent-chain re-composition.

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass |
| Focused tests | `GenericDeclarationPatternExtractionTests` passed (29/29) |
| Decompiler fast suite | Pass (2560/0, minus `Speed=Slow`) |
| Reduced fixture validity | Pass — FluentAssertions defect diff vs base `acde101a`: 0 regressed, 0 improved |
| Reduced fixture fidelity | Not currently checkable (render A/B structural, valid→valid 8) |
| Real witness | `NumericAssertionsBase\`3::BePositive` folded (copy + up-front decl eliminated) |

## Decompiler quality

**Conclusion:** **PASS** — pinned-subset lowering residue improves 17.83% → 17.67% (-0.16 pp) with 0 pass bugs; aggregate advisory movement is corpus drift (MetadataPrimitives 61→100 methods), not this change.

### PR quick gate

Run: PR quick corpus, hash-stable 100 methods per assembly; validity/fidelity ungated in the quick card (measured separately above).

| --- | ---: | ---: | ---: |
| Detected lowering residue (-) | 17.83% | 17.67% | -0.16 pp |
| Pass bugs (-) | 0 | 0 | 0 |

> **Conclusion:** **PASS** — pinned-subset gate improves lowering residue with no new pass bugs.

### Aggregate context

Corpus: 15 assemblies, 1,500 methods. Baseline drift: corpus drifted from the pinned baseline (2026-06-26); `ILInspector.MetadataPrimitives` 61→100 methods (+39), so aggregate count deltas are not a clean PR signal.

| --- | ---: | ---: | ---: |
| Detected lowering residue (-) | 248 (16.97%) | 252 (16.80%) | n/a (population differs) |
| Conditional-branch residue (-) | 35 (2.40%) | 47 (3.13%) | n/a (population differs) |
| Forward-merge stops (-) | 27 (1.85%) | 33 (2.20%) | n/a (population differs) |

> **Conclusion:** **ADVISORY** — aggregate movement is attributable to corpus drift; rebaseline before trusting count-based regressions. Regression verdict from the sensor: PASS.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.GenericDeclarationPatternExtractionTests
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/GenericDeclarationPatternExtractionTests.cs
+++ b/src/ILInspector.Decompiler.Tests/GenericDeclarationPatternExtractionTests.cs
@@ -605,6 +605,40 @@ public class GenericDeclarationPatternExtractionTests
         Assert.Matches(@"\bV_\d+ = V_\d+;", output);
     }
 
+    // Synthetic (#2872 gate, Issue 1): the cache local's managed address is
+    // captured into a ref local (`ref T p = ref c;`) that outlives the guarded
+    // arm. The address load sits inside the arm, so confinement alone is
+    // satisfied, but re-pointing the binding narrows `c` to the pattern's scope
+    // and leaves the escaped reference dangling. The copy must survive.
+    [Fact]
+    public void CacheAddressCapturedToRefLocal_DoesNotBindToCacheLocal()
+    {
+        var generic = TypeRef.GenericParameter(0, "T");
+        var objectType = TypeRef.CoreLib("System", "Object");
+        var byRef = TypeRef.ByRef(generic);
+        IrExpression Subject() => new LoadArgument(0, "subject", generic);
+
+        var then = new Block();
+        then.Add(new StoreLocal(0, generic, new LoadLocal(1, generic)));
+        then.Add(new StoreLocal(2, byRef, new LoadLocalAddress(0, generic)));
+        var block = new Block();
+        block.Add(new IfStatement(new IsPattern(Subject(), generic, 1), then, null));
+        block.Add(new Return(new Constant(null, objectType)));
+        var body = new BlockContainer();
+        body.Add(block);
+        var function = new IrFunction(
+            "M",
+            TypeRef.Definition("Synthetic", "Samples", "Owner"),
+            new MethodSignature(objectType, [new Parameter("subject", generic)], HasThis: false, GenericParameterCount: 1),
+            [generic, generic, byRef],
+            body);
+
+        var output = RunPipelineAndPrint(function);
+
+        Assert.Matches(@"is T V_\d+", output);
+        Assert.Matches(@"\bV_\d+ = V_\d+;", output);
+    }
+
     static string RunPipelineAndPrint(IrFunction function)
     {
         IrPasses.Run(function);
--- a/src/ILInspector.Decompiler.Tests/GenericDeclarationPatternExtractionTests.cs
+++ b/src/ILInspector.Decompiler.Tests/GenericDeclarationPatternExtractionTests.cs
@@ -605,7 +605,48 @@ public class GenericDeclarationPatternExtractionTests
         Assert.Matches(@"\bV_\d+ = V_\d+;", output);
     }
 
-    // Synthetic (#2872 gate, Issue 1): the cache local's managed address is
+    // Synthetic (#2872 gate, Issue 1 re-review): the cache local's address is
+    // forwarded through a by-ref-returning receiver call — `p = ref c.Ref()` —
+    // and stored into a ref local that outlives the arm. The `LoadLocalAddress`
+    // sits directly under the `Call`, so a gate that only inspects the address's
+    // direct parent (Store/Return) would miss the escape and fold. The pointer
+    // still escapes once the binding narrows to the pattern's scope, so the fold
+    // must be rejected and the copy must survive. This exercises the
+    // in-place-receiver allow-list's non-by-ref-result clause: a by-ref-returning
+    // receiver call is not an in-place consumer.
+    [Fact]
+    public void CacheAddressForwardedThroughByRefReturningCall_DoesNotBindToCacheLocal()
+    {
+        var generic = TypeRef.GenericParameter(0, "T");
+        var objectType = TypeRef.CoreLib("System", "Object");
+        var byRef = TypeRef.ByRef(generic);
+        IrExpression Subject() => new LoadArgument(0, "subject", generic);
+
+        // A by-ref-returning instance method on the cache local; the returned
+        // reference forwards the receiver's address to the caller.
+        var refReturn = new MethodRef(generic, "Ref", byRef, [], HasThis: true);
+
+        var then = new Block();
+        then.Add(new StoreLocal(0, generic, new LoadLocal(1, generic)));
+        then.Add(new StoreLocal(2, byRef,
+            new Call(refReturn, isVirtual: false, [new LoadLocalAddress(0, generic)])));
+        var block = new Block();
+        block.Add(new IfStatement(new IsPattern(Subject(), generic, 1), then, null));
+        block.Add(new Return(new Constant(null, objectType)));
+        var body = new BlockContainer();
+        body.Add(block);
+        var function = new IrFunction(
+            "M",
+            TypeRef.Definition("Synthetic", "Samples", "Owner"),
+            new MethodSignature(objectType, [new Parameter("subject", generic)], HasThis: false, GenericParameterCount: 1),
+            [generic, generic, byRef],
+            body);
+
+        var output = RunPipelineAndPrint(function);
+
+        Assert.Matches(@"is T V_\d+", output);
+        Assert.Matches(@"\bV_\d+ = V_\d+;", output);
+    }
     // captured into a ref local (`ref T p = ref c;`) that outlives the guarded
     // arm. The address load sits inside the arm, so confinement alone is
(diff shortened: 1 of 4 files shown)


richlander:
## Adversarial review — round 1

Two independent reviewers examined the initial head `29c196b4`, isolated in separate worktrees under `/tmp` with scratch confined there.

| Reviewer | Model | Verdict (head `29c196b4`) |
| --- | --- | --- |
| rev-gpt | GPT-5.5 | Clean — no HIGH/MED soundness bug |
| rev-gem | Gemini 3.1 Pro | 2 HIGH completeness findings |

**GPT-5.5** verified with real runs: solution + harness build, focused `GenericDeclarationPatternExtractionTests` 29/29, fast decompiler suite 2560/2560, FluentAssertions validity sweep, an address-escape-through-`T*` probe (fold fired, output stayed compilable), and a two-guard name-collision probe. No finding.

**Gemini 3.1 Pro** raised two HIGH completeness gaps in the gate's soundness abstraction:

1. **`ReferenceOwnership.ReferencesLocal` under-counts.** It matches only `LoadLocal`/`StoreLocal`/`LoadLocalAddress`, but a local can also be bound/written through a designation that carries the index directly (`NullCoalescingAssignment`, `ForeachStatement`, `UsingStatement`, `Fixed`, `IsPattern`/`RecursivePropertyDeclarationPattern`, `UnionSwitchExpressionArm`, `CatchClause`, `DeconstructionAssignment`) — the same set the C# printer treats as declaration-producing. A confinement proof built on `ReferencesLocal` alone was blind to those.
2. **Address escape.** The confinement check only proves `LoadLocalAddress` sits inside the arm, not that the address does not escape (e.g. captured into a ref local outliving the arm).

### Reconciliation

Investigation confirmed **neither finding reproduces through real csc output**: the cache local is csc's source pattern variable, whose source scope confines every use to the guarded arm, so csc never emits an out-of-arm use or an escaping ref-capture of it (this is why GPT-5.5's real-run escape probe stayed valid and Gemini's evidence is hand-written C# rather than harness output). But both were **real gaps in the gate's soundness abstraction**, since the gate identifies the cache local structurally and leaned on `ReferencesLocal` being complete. Closed defensively in `c7af3868`:

- **Finding 1:** added `BindsLocal` / `ReferencesOrBindsLocal` / `SubtreeReferencesOrBindsLocal` / `LocalReferencedOrBoundOnlyWithin` (covering the full index-bearing designation set that matches the printer's declaration model); the pass now uses the `OrBinds` variants for the tested-value check and the cache local's confinement. `ReferenceOwnershipTests` proves the legacy check passes on an external `??=` binding while the completeness-aware check rejects it.
- **Finding 2:** added an address-escape gate rejecting the fold when a `LoadLocalAddress` of the cache local is captured by a `StoreLocal`/`StoreStackSlot`/`StoreField`/`StoreIndirect`/`Return` (an in-place call receiver like `c.CompareTo(x)` — BePositive's shape — is still allowed). A synthetic pass test proves a `ref T p = ref c` capture blocks the fold, verified meaningful by revert-check.

**Evidence at `c7af3868`:** FluentAssertions render A/B unchanged (Changed 8, valid→valid 8, valid→invalid 0); validity defect diff vs base `acde101a` 0 regressed / 0 improved; all 8 `NumericAssertionsBase` guards (including `BePositive`) still fold; decompiler fast suite 2563/2563.

Re-review of the fixed head `c7af3868` is in progress with both models; will post the fixed-head verdicts before marking ready.

richlander:
## Adversarial review — round 2 (fixed head)

The round-1 hardening (`c7af3868`) was re-reviewed by both models, then a follow-up (`81189697`) closed the one remaining finding, and both models re-reviewed the final head.

### `c7af3868` (round-1 fixes)

| --- | --- | --- |
| rev-gpt | GPT-5.5 | Clean |
| rev-gem | Gemini 3.1 Pro | Issue 2 resolved; **Issue 1 still open** |

Gemini confirmed the `BindsLocal` completeness set matches the printer's scope-ownership model (Issue 2 closed), but showed the address-escape gate was still incomplete: it only inspected the `LoadLocalAddress`'s **direct parent** (`Store*`/`Return`), so an address forwarded through a by-ref-returning call — `p = ref c.Ref()` — bypassed the deny-list (parent is a `Call`) and let the pointer escape the narrowed pattern scope (CS8374-class invalid C#).

### `81189697` (Issue 1 fix)

Replaced the incomplete direct-parent deny-list with a positive allow-list — the fold is admitted only when **every** `LoadLocalAddress` of the cache local is an in-place instance-call receiver returning a non-by-ref value:

```csharp
static bool IsInPlaceReceiver(LoadLocalAddress address)
    => address.Parent is Call { Callee.HasThis: true } call
        && call.Children.Count > 0
        && ReferenceEquals(call.Children[0], address)
        && call.ResultType is not { Kind: TypeRefKind.ByRef };
```

Matching the safe shape positively closes the ref-forwarding hole a direct-parent list cannot: storing the pointer, a by-ref argument to a static call, and a by-ref-returning receiver call are all now rejected, while the in-place `c.CompareTo(x)` receiver (returns `int`) still folds. Added `CacheAddressForwardedThroughByRefReturningCall_DoesNotBindToCacheLocal`, revert-checked meaningful (fails under the old direct-parent gate).

| Reviewer | Model | Verdict (`81189697`) |
| --- | --- | --- |
| rev-gpt | GPT-5.5 | **Clean** |
| rev-gem | Gemini 3.1 Pro | **Clean** — both findings resolved |

- **GPT-5.5:** solution build; `GenericDeclarationPatternExtractionTests` + `ReferenceOwnershipTests` 37/37; fast suite 2564/0; FluentAssertions validity sweep unchanged; all 8 `NumericAssertionsBase` folds (`Be`, `BeGreaterThan`, `BeGreaterThanOrEqualTo`, `BeOneOf`, `BePositive`, `Match`, `NotBe`, `NotBeInRange`) still fire; confirmed constrained instance calls place the receiver at `Children[0]`.
- **Gemini 3.1 Pro:** probed `ref`/`out` argument escape (rejected — address is `Children[1+]`, not the receiver), delegate capture (rejected — `NewObject`, not a `Call`), and `ref struct`-returning receiver calls (native CS8168 escape rules guard both pre- and post-fold identically, so no new invalid C#); confirmed `BePositive` still folds and no over-rejection; re-audited `BindsLocal` against the printer's declaration model.

Both reviewers isolated in separate worktrees at the exact head with scratch confined to `/tmp`.

**Evidence at `81189697`:** decompiler fast suite 2564/0; FluentAssertions render A/B Changed 8, valid->valid 8, valid->invalid 0; validity defect diff vs base `acde101a` 0 regressed / 0 improved; `BePositive` still folds to `if (Subject is T subject)`.

Two clean fixed-head reviews.

richlander:
Ready to merge

PR #2841, opened 2026-07-18
Consolidate body-reconstruction facts into MemberBodyFacts (#2777)

richlander:
- Advances #2777 (RTS orchestrator thin; product owns IR-derived facts)
- Changes: body-referenced-namespace extraction moves from the harness into the product Decompiler; no behavior change (byte-neutral)

## Change

**Conclusion: PASS — a byte-identical IR walk that lived (twice) in the harness moves to the product it belongs to; provably no output change.**

The RTS planner and the fidelity harness each walked Decompiler IR
(`function.Descendants` → `DirectTypes`/`ResultType` → `TypeRef`) to collect the
namespaces a body references for its `using` directives. That is an IR-derived fact,
and per the layering contract IR facts belong in the product (`ILInspector.Decompiler`
owns the IR; the harness and C# seam must not read it). The two copies were
**byte-identical** — the exact "consumers previously duplicated in the harness"
anti-pattern the existing `ConstructorBodyFactExtractor` was created to retire.

- Add `BodyNamespaceExtractor.ReferencedNamespaces(IrFunction)` in
  `src/ILInspector.Decompiler/`, returning a neutral ordinal-sorted
  `IReadOnlySet<string>` (plain strings; no Decompiler type crosses the boundary).
  This mirrors the `ConstructorBodyFactExtractor` → `ConstructorBodyFacts` pattern.
- Both harness sites (`ReturnToSenderTypePlanner.cs` ×3 usings sites,
  `FidelityCheck.cs` ×1) consume the product method; the two duplicated
  `RequiredNamespaces` methods are deleted.
- The walk is copied **verbatim** (same `SortedSet` ordinal ordering), so the emitted
  `using` lists are unchanged.

Interfaces/closure planning and body-fact reads that use RTS planning vocabulary are
untouched; only the pure namespace walk (a C#-composition input) moves.

## Evidence

| Check | Result |
| --- | --- |
| Product build (`dotnet-inspect.slnx -c Release`) | Pass (0 errors) |
| New unit tests (`BodyNamespaceExtractorTests`) | 3 pass — multi-namespace body, omission of unreferenced namespaces, ordinal sorting, against a real compiled fixture |
| RTS oracle (`ReturnToSenderPrototypeTests`/`EvidenceTests`/`FixtureCatalogTests`) | 199 pass, **zero fixture churn** |
| Decompiler fast suite (`-notrait Speed=Slow`) | 2499 pass |
[...]

PR #2880, opened 2026-07-18
Handle function-pointer component types in body namespace extraction

### LINKED ISSUE #2847: Handle function-pointer component types in body namespace extraction
`BodyNamespaceExtractor.ReferencedNamespaces` preserves the former harness walk, which does not traverse `TypeRefKind.FunctionPointer`. Function-pointer return and parameter types can therefore contribute namespaces that are missed when they appear only inside a function-pointer type.

This is pre-existing behavior moved by PR #2841, not a regression in that PR. Extend the product-owned extractor to visit the function pointer element/argument types and add a compiled close-negative fixture.

richlander:
Fixes #2847.

`BodyNamespaceExtractor.ReferencedNamespaces` (now `MemberBodyFacts.ReferencedNamespaces`) preserves the former harness walk, which had no `TypeRefKind.FunctionPointer` case in its recursive `Add()` switch. Function-pointer return and parameter types could therefore contribute namespaces that were missed when they appeared only inside a function-pointer type. Per the issue, this is pre-existing behavior moved by #2841, not a regression introduced there.

## Change

- `src/ILInspector.Decompiler/MemberBodyFacts.cs`: added a `TypeRefKind.FunctionPointer` case to `Add()` that recurses into the return type (`TypeRef.ElementType`) and each parameter type (`TypeRef.TypeArguments`), matching the existing recursive handling already in place for generic instances, arrays, by-refs, pointers, and pinned types.
- `src/ILInspector.Decompiler.Tests/FunctionPointerNamespaceFixtures.cs` (new): `ParameterMarker`/`ReturnMarker` types in a dedicated namespace.
- `src/ILInspector.Decompiler.Tests/CfgSampleClass.cs`: added `NamespaceRefSample.ReferencesNamespacesOnlyThroughFunctionPointer(delegate*<ParameterMarker, ReturnMarker> callback)` — the fixture namespace is reachable *only* through the function pointer's return/argument types, never as an ordinary local, field, or parameter/return type, isolating the new code path as the sole discovery route.
- `src/ILInspector.Decompiler.Tests/MemberBodyFactsTests.cs`: added `ReferencedNamespaces_ReportsFunctionPointerComponentNamespaces`, asserting the namespace is now reported.

## Evidence

- `dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class ILInspector.Decompiler.Tests.MemberBodyFactsTests`: 11/11 pass.
- Confirmed this is a genuine close-negative: temporarily reverted just the `MemberBodyFacts.cs` fix (via `git stash`) and reran the new test — it fails with `Assert.Contains() Failure: Item not found in set ... Not found: "FunctionPointerNamespaceFixtures"`. Restored the fix; full class passes again (11/11).

Test-only + small, mechanical product fix (one new switch case) with no behavior change outside the fixed gap, so no adversarial review is required.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -4827,6 +4827,16 @@ public static class NamespaceRefSample
         _ = hpack.ToString();
         _ = headers.ToString();
     }
+
+    // References FunctionPointerNamespaceFixtures only through a function pointer
+    // parameter's return type and argument type (delegate*<...>'s element/type
+    // arguments), never as an ordinary local, field, or parameter/return type. See
+    // FunctionPointerNamespaceFixtures.cs.
+    public static unsafe void ReferencesNamespacesOnlyThroughFunctionPointer(
+        delegate*<FunctionPointerNamespaceFixtures.ParameterMarker, FunctionPointerNamespaceFixtures.ReturnMarker> callback)
+    {
+        _ = callback;
+    }
 }
 
 // Fixtures for MemberBodyFactsTests backing-field cases. Auto-property accessors
--- a/src/ILInspector.Decompiler.Tests/FunctionPointerNamespaceFixtures.cs
+++ b/src/ILInspector.Decompiler.Tests/FunctionPointerNamespaceFixtures.cs
@@ -1,0 +1,22 @@
+// Fixture namespace for MemberBodyFactsTests.ReferencedNamespaces_ReportsFunctionPointerComponentNamespaces.
+//
+// FunctionPointerNamespaceFixtures.ParameterMarker and .ReturnMarker are only
+// referenced through a function-pointer parameter/return type
+// (CfgSampleClass.TakesFunctionPointerWithCustomNamespaceTypes), never directly
+// as a local, field, or ordinary parameter/return type. That isolates
+// TypeRefKind.FunctionPointer as the only path MemberBodyFacts.ReferencedNamespaces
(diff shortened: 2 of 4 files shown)


richlander:

Exact head reviewed: `379ad5e85a441b7270deb91bce7205c309587f75`

| Reviewer | Verdict | Blocking findings |
| --- | --- | --- |
| Claude Opus 4.8 | **CLEAN** | None |
| Gemini 3.1 Pro | **CLEAN** | None |

Both reviewers verified the function-pointer `TypeRef` contract across every
construction path: `ElementType` is the return type and `TypeArguments` are the
parameter types. They also checked recursive composite types, malformed/null
handling, namespace ordering/deduplication, SRM/NativeAOT compatibility, and
the close-negative fixture. Opus independently reverted only the product switch
case and reproduced the new test's failure, then restored it and confirmed the
focused class passes.

Two **non-blocking** observations require no action for this PR:

- `ParameterMarker` and `ReturnMarker` share one namespace, so the fixture proves
  function-pointer traversal but does not independently discriminate the return
  and parameter recursion lines. Inspection of all `TypeRef.FunctionPointer`
  producers confirms both lines use the correct layout; distinct fixture
  namespaces would only strengthen test localization.
- The fixture header comment names
  `CfgSampleClass.TakesFunctionPointerWithCustomNamespaceTypes`, while the
  landed method is `ReferencesNamespacesOnlyThroughFunctionPointer`. This is a
  comment-only naming mismatch with no behavioral effect.

CI is green on the reviewed head. No merge-blocking findings or code changes
resulted from review.

richlander:
Ready to merge

PR #2881, opened 2026-07-18
Recover protected-region leave as structured continue

### LINKED ISSUE #2861: Decompiler: recover protected-region leave as structured continue
## Current state

#2857 fixes the invalid-output defect from #2830 conservatively. When a
protected-region `leave` targets a loop increment block, `ForLoopPass` now
retains the valid `while` fallback so the target label survives:

```csharp
int i = default;
i = 0;
while (i < 10)
{
    try
    {
        if (i == 5)
            goto IL_0010;
        Console.WriteLine();
    }
    catch (Exception)
    {
    }

    Console.WriteLine();

IL_0010:
    i++;
}
```

This output is valid, but it still exposes Roslyn's protected-region lowering
instead of recovering the authored `continue`.

## Fully raised goal

```csharp
for (int i = 0; i < 10; i++)
{
    try
    {
        if (i == 5)
            continue;
        Console.WriteLine();
    }
    catch (Exception)
    {
    }

    Console.WriteLine();
}
```

## Required slices

1. **Recover an EH-safe structured continue atomically**
   - Bind each candidate `Leave` to the exact trailing increment of the
     candidate `WhileLoop`, with the current loop established structurally.
   - Prove that every bound leave exits a `try` or `catch`, is not inside a
     `finally` body, and belongs to that loop rather than a nested loop or
     function.
   - Require every live target of the increment to be one of those proven
     leaves. Any branch, switch, unproven leave, or existing current-loop
     `Continue` retains #2857's valid `while` fallback.
   - Raise the `WhileLoop` to a `ForLoop` first, then replace only the bound
     leaves with typed `Continue` nodes. A `continue` in the pre-raise
     `while` would skip the tail increment and change behavior.

2. **Preserve the lowered contract**
   - Keep the recovery in `ForLoopPass`; do not add an earlier global
     leave-to-continue rewrite.
   - The lowered pipeline intentionally omits `ForLoopPass`, so it must retain
     the existing valid labeled `while` representation.
   - Preserve loop identity across nested loops and reject leaves targeting a
     different loop, method tail, or unrelated block.

## Evidence

Add compiler-produced positive and close-negative fixtures for:

- `continue` from `try` and `catch` to the owning loop increment;
- multiple proven leaves targeting the same increment;
- intervening `finally` cleanup;
- nested loops with distinct continuation targets;
- leave to a method tail or another loop;
- nested lambda/local-function offset collisions.

Add a synthetic pass-seam fixture for a remaining ordinary branch to the
increment block. In compiler-produced C#, that branch either structures away
to natural fallthrough or prevents the enclosing loop from structuring, so the
post-structuring `WhileLoop` plus live-branch state is an unreachable seam
rather than a stable compiler shape.

Pin the fully raised C#, compile-back validity, control-flow semantics,
Debug/Release compiler shapes, and distinct sugared/lowered fidelity behavior.

## Related

- #2830 / #2857: preserve the live target and emit valid fallback C#
- #1748: prior missing-label symptom for a catch-body `continue`
- #1367: `ForLoopPass` continue-semantics guard

richlander:
Fixes #2861

- Recovers protected-region `leave` instructions as structured `continue` only
  when their owning `while` loop atomically raises to `for`.
- Preserves #2857's labeled `while` fallback for mixed, nested, unproven, and
  lowered shapes.
- Card revision: `364104e42bc3d720651f941ef722ecbe42165238`

## Change

**Conclusion:** **PASS** — the transform consumes every live increment target
or declines, and the recovered try/catch/finally fixtures recompile to the
original opcode stream.

### Original source

```csharp
public static void Issue2830_ForLoopLeave()
{
    for (int i = 0; i < 10; i++)
    {
        try
        {
            if (i == 5)
                continue;
            Console.WriteLine();
        }
        catch (Exception)
        {
        }
        Console.WriteLine();
    }
}
```

### Before

```csharp
public static void Issue2830_ForLoopLeave()
{
    int i = default;
    i = 0;
    while (i < 10)
    {
        try
        {
            if (i == 5)
            {
                goto IL_0010;
            }
            Console.WriteLine();
        }
        catch (Exception)
        {
        }
        Console.WriteLine();

    IL_0010:
        i++;
    }
}
```

### After

```csharp
public static void Issue2830_ForLoopLeave()
{
    for (int i = 0; i < 10; i++)
    {
        try
        {
            if (i == 5)
            {
                continue;
            }
            Console.WriteLine();
        }
        catch (Exception)
        {
        }
        Console.WriteLine();
    }
}
```

### Fully Raised

## Evidence

| Check | Result |
| --- | --- |
| Product build | Release solution passed |
| Focused tests | 19 substrate/pass/compiler-fixture tests passed |
| Compiler configurations | Release and Debug protected-shape fixtures passed |
| Decompiler fast suite | 2,561 tests passed |
| Sugared fidelity gate | All 4 gate tests passed; affected positives are pinned Exact |
| Changed-method fidelity | 4 Exact positives; 1 intentional fallback OpcodeDiff |
| Lowered fidelity | All 5 affected rows remain valid OpcodeDiff fallbacks |

The positive matrix covers try-body and catch-body continues, multiple leaves
to one increment, and intervening `finally` cleanup. Close negatives cover
nested loop ownership, nested-function offset collisions, finally bodies,
outer-only protected regions, ordinary current-loop `continue`, and mixed live
targets. `ContinueOrigin.ProtectedRegionLeaveToForIncrement` clears `DEC0013`
only for the proven atomic transform; all existing continues remain unverified.

## Decompiler quality

**Conclusion:** **PASS** — the pinned quick gate passes, and a clean
merge-base/head A/B over the exact same population is neutral with zero
per-method deltas.

### PR quick gate

Run: PR quick corpus, hash-stable 100 methods per assembly; 15 assemblies and
1,500 methods.

| Metric (goal) | Merge base | PR | Delta |
| --- | ---: | ---: | ---: |
| Detected lowering residue (-) | 251 (16.73%) | 251 (16.73%) | 0 |
| Conditional-branch residue (-) | 46 (3.07%) | 46 (3.07%) | 0 |
| Forward-merge stops (-) | 32 (2.13%) | 32 (2.13%) | 0 |
| Pass bugs (-) | 0 | 0 | 0 |

> **Conclusion:** **PASS** — pinned-subset residue and branch rates are
> unchanged, and the clean A/B produced zero changed methods.

### Aggregate context

The committed PR-quick baseline is stale because
`ILInspector.MetadataPrimitives` now contributes 100 rather than 61 sampled
methods. Its hard pinned-subset gate still passes, but aggregate baseline
movement is not attributed to this PR. The clean merge-base/head A/B above uses
the same current assemblies and population on both sides.

## Validation

```bash
  -class "ILInspector.Decompiler.Tests.ProtectedRegionControlFlowTests" \
  -class "ILInspector.Decompiler.Tests.ForLoopPassTests" \
  -class "ILInspector.Decompiler.Tests.ProtectedContinueRecoveryTests"
  -trait- "Speed=Slow"
  -class "ILInspector.Decompiler.Tests.FidelityGateTests"
npx markdownlint-cli docs/design/decompiler-substrate.md
```

### DIFF
--- a/docs/design/decompiler-substrate.md
+++ b/docs/design/decompiler-substrate.md
@@ -36,11 +36,16 @@ call. The live substrate layers are:
 | `GeneratedCodeIdentity` | Compiler-generated shape (attribute-gated) | Is this a non-capturing lambda holder? |
 | `PlaceIdentity` | Intra-method re-evaluable place identity | Are these two reads the same local? |
 | `ReferenceOwnership` | Reference-scope ownership for consumed scaffolds | Is every use of this synthetic local inside the nodes this rewrite consumes? |
+| `ProtectedRegionControlFlow` | EH transfer legality | Can this protected-region leave become structured control flow? |
 
 They are siblings by construction — thin, allocation-light, no pass state, named
 for the evidence category not the caller — but each owns a different category:
 metadata identity, compiler-shape evidence, intra-method place identity, and
-reference-scope ownership.
+reference-scope ownership. `ProtectedRegionControlFlow` adds the narrow EH
+category shared by `StructuringPass` and `ForLoopPass`: it proves that a
+`Leave` exits a `try`/`catch` region and never originates in a `finally` body.
+Its boundary-aware atom also proves that the protected region is below the
+candidate construct; the consuming pass still owns target and loop identity.
 
 ## Boundaries and roadmap
 
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -1903,6 +1903,61 @@ public class CfgSampleClass
             }
         }
     }
+
+    public static void Issue2861_ForLoopTryAndCatchContinues()
+    {
+        for (int i = 0; i < 10; i++)
+        {
+            try
+            {
+                if (i == 2)
+                    continue;
+                if (i == 5)
+                    throw new System.InvalidOperationException();
+                System.Console.WriteLine();
+            }
+            catch (System.InvalidOperationException)
+            {
+                if (i == 5)
+                    continue;
+                System.Console.WriteLine();
+            }
+            System.Console.WriteLine();
+        }
+    }
+
+    public static void Issue2861_NestedProtectedLeaveToOuterIncrement()
+    {
+        int i = 0;
+        while (i < 10)
+        {
+            try
+            {
+                if (i == 2)
+                    goto Increment;
+                System.Console.WriteLine();
+            }
+            catch (System.Exception)
+            {
+            }
+            for (int j = 0; j < 3; j++)
+            {
+                try
+                {
+                    if (i == j)
+                        goto Increment;
+                    System.Console.WriteLine();
+                }
+                catch (System.Exception)
+                {
+                }
+            }
+            System.Console.WriteLine();
+        Increment:
+            i++;
+        }
+    }
+
     public static int LoopWithBreak(int[] values, int limit)
     {
         int sum = 0;
--- a/src/ILInspector.Decompiler.Tests/DecompilerFindingsTests.cs
+++ b/src/ILInspector.Decompiler.Tests/DecompilerFindingsTests.cs
@@ -98,6 +98,20 @@ public class DecompilerFindingsTests
     }
 
     [Fact]
+    public void Inspect_ProvenProtectedForContinue_HasNoResidualCause()
+    {
+        var function = Function(
+            new Continue(ContinueOrigin.ProtectedRegionLeaveToForIncrement),
+            blockOffset: 0x20);
+
+        var inspection = CompleteInspection(
+            DecompilerFindings.InspectFidelityCauses(function, Subject));
+
+        Assert.Empty(inspection.Findings);
+        Assert.Equal(DecompilationFidelity.Full, function.Fidelity);
+    }
+
+    [Fact]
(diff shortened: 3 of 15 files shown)


richlander:

Exact head reviewed: `364104e42bc3d720651f941ef722ecbe42165238`

| Reviewer | Verdict | Findings |
| --- | --- | --- |
| Claude Opus 4.8 | **CLEAN** | None |
| Gemini 3.1 Pro | **CLEAN** | None |

Both reviewers independently attacked target-source completeness, source-offset
identity, cache staleness after tree rewrites, nested loop/function ownership,
protected-region boundary direction, `finally` semantics, atomic mutation
ordering, narrowed `ContinueOrigin` fidelity evidence, lowered-mode isolation,
and compiler-fixture oracle quality. Opus also reproduced the real
compiler-produced positive and nested-loop fallback shapes from the exact-head
checkout.

No review findings required code changes or explicit non-actions.

Post-review validation on the same head:

- complete sugared fidelity gate: 4/4 passed;
- complete lowered fidelity gate: 3/3 passed;
- targeted affected-method fidelity: 4 sugared Exact, 1 intentional fallback
  OpcodeDiff; all 5 lowered rows remain valid OpcodeDiff fallbacks;
- Release solution build and 2,561-test fast decompiler suite passed;
- clean merge-base/head PR-quick A/B: 15 assemblies, 1,500 methods, all metrics
  neutral, zero changed methods.

richlander:
Ready to merge

PR #2882, opened 2026-07-18
Keep static on explicitly implemented interface properties

### LINKED ISSUE #2875: Explicit-interface property rendering drops `static` for static-abstract members
## Summary

Follow-up to #2827 / PR #2858. The seam renders explicit-interface **properties** (`Kind == "explicit-interface-implementation"`) without their `static` modifier, so a C# 11 static-abstract interface member implemented explicitly reconstructs as an instance member (would be CS0106/CS0539 on recompile).

## Root cause

`CSharpDeclarationWriter.RenderMemberDeclarationCore` builds modifiers in three branches:

- `.cctor` -> `static` (+ `unsafe`)
- `member.Kind != "explicit-interface-implementation"` -> full modifier set (incl. `static`)
- else (explicit-interface-implementation) -> **only `unsafe`**

The explicit-impl branch omits the access modifier correctly but never adds `static`, unlike the `.cctor` branch.

## Repro (seam-level, deterministic)

```csharp
var type = new ApiType { Namespace = "Samples", Name = "Counter", Kind = "class" };
var member = new ApiMember
{
    Name = "Samples.ICounter.Count",
    Kind = "explicit-interface-implementation",
    IsStatic = true,
    SignatureModel = new ApiSignature
    {
        ReturnType = "int",
        MemberName = "Samples.ICounter.Count",
        Accessors = [new ApiAccessor { Kind = "get" }]
    }
};
CSharpDeclarationWriter.RenderMemberDeclaration(type, member);
```

- Actual (origin/main @ 7b2c3840): `int Samples.ICounter.Count { get; }`
- Expected: `static int Samples.ICounter.Count { get; }`

## Fix sketch

In the explicit-impl modifier branch, add `static` when `member.IsStatic` (mirroring the `.cctor` branch: static then unsafe).

## Scope / evidence status

- Seam gap is proven and deterministic (above).
- End-to-end reachability through a real compiled fixture (static-abstract interface property implemented explicitly, extracted with `Kind == "explicit-interface-implementation"` + `IsStatic == true`) is **not yet verified** — #2858's pinned corpus did not exercise this shape, so this is a latent gap, not a measured regression. A real importer/compiled-fixture canary should accompany the fix.

Discovered while retiring the superseded PR #2845, whose `_KeepsStaticModifier` case covered exactly this shape.

richlander:
## Summary

Follow-up to #2827 / #2858. The explicit-interface-implementation modifier branch in `CSharpDeclarationWriter` omitted the access modifier correctly but never added `static`. A C# 11 static-abstract interface member implemented explicitly (`static int I.Count => ...`) reconstructed as an instance member (`int I.Count { get; }`), failing the interface contract on recompile (CS0106/CS0539).

- Add `static` (then `unsafe`, mirroring the `.cctor` order) to the explicit-impl modifier branch when `member.IsStatic`.

## Reachability

Confirmed reachable through RTS: the getter/setter builders carry the accessor's `MethodAttributes.Static`, and `ToApiMember` maps a dotted explicit property to `Kind == "explicit-interface-implementation"` with `IsStatic` intact, so the seam saw a static member and dropped the modifier.

## Evidence

- Seam unit test `ExplicitInterfaceProperty_Static_RetainsStaticModifier`: `static int Samples.ICounter.Count { get; }`.
- Compiled-fixture RTS canary `CompileBackFirstPropertyGetter_KeepsStaticOnExplicitInterfaceProperty`: reconstructs `static int ICounter.Count` from a real static-abstract explicit implementation. Fails on unfixed `main` (asserted `static` absent), passes with the fix.
- Seam suite 272/0; RTS explicit-interface suite 9/0.
- Newtonsoft `--return-to-sender` unchanged vs base: 544 Exact / 16 OpcodeDiff / 8 RecompileFail / 13 ContextFail (Newtonsoft has no static explicit-impl properties, so the additive change is a no-op there).
- #2776 parity gate exit 0 (4 known burn-down rows, 0 new).

## Risk

Low and contained: the change is additive — it adds `static` only for `Kind == "explicit-interface-implementation"` members with `IsStatic == true`. All non-static explicit implementations (the common case) are unaffected. Requesting the two-model adversarial review per policy given it is a behavior change in the seam.

Closes #2875.

### DIFF
--- a/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
@@ -907,6 +907,30 @@ public sealed class CSharpDeclarationWriterTests
     }
 
     [Fact]
+    public void ExplicitInterfaceProperty_Static_RetainsStaticModifier()
+    {
+        // #2875: static-abstract interface members implemented explicitly must keep `static`
+        // while still omitting the access modifier.
+        var type = new ApiType { Namespace = "Samples", Name = "Counter", Kind = "class" };
+        var member = new ApiMember
+        {
+            Name = "Samples.ICounter.Count",
+            Kind = "explicit-interface-implementation",
+            IsStatic = true,
+            SignatureModel = new ApiSignature
+            {
+                ReturnType = "int",
+                MemberName = "Samples.ICounter.Count",
+                Accessors = [new ApiAccessor { Kind = "get" }]
+            }
+        };
+
+        var declaration = CSharpDeclarationWriter.RenderMemberDeclaration(type, member);
+
+        Assert.Equal("static int Samples.ICounter.Count { get; }", declaration);
+    }
+
+    [Fact]
     public void ExplicitInterfaceImplementation_WithUnsafeSignature_RetainsUnsafeModifier()
     {
         var type = new ApiType { Namespace = "Samples", Name = "UnsafeImpl", Kind = "class" };
--- a/src/ILInspector.CSharp/CSharpDeclarationWriter.cs
+++ b/src/ILInspector.CSharp/CSharpDeclarationWriter.cs
@@ -402,9 +402,15 @@ internal static class CSharpDeclarationWriter
             if (member.IsUnsafe || options.ForceUnsafe)
                 modifiers.Add("unsafe");
         }
-        else if (member.IsUnsafe || options.ForceUnsafe)
+        else
         {
-            modifiers.Add("unsafe");
+            // Explicit interface implementations omit the access modifier but must still
+            // carry `static` (C# 11 static-abstract interface members implemented explicitly)
+            // and `unsafe`. Order mirrors the .cctor branch: static then unsafe.
+            if (member.IsStatic)
+                modifiers.Add("static");
+            if (member.IsUnsafe || options.ForceUnsafe)
+                modifiers.Add("unsafe");
         }
 
         if ((options.ForceAsync || member.IsAsync)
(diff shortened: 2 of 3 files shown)


richlander:
## Adversarial review — two clean passes

Both reviewers ran in isolated worktrees at the PR head (263d58b5), from a different model family than the author (Claude). Both confirmed clean workspaces (`git status`) and reported **no blocking findings**.

**GPT-5.5** — no significant issues. Verified:
- The new `else` is reachable only for `Kind == "explicit-interface-implementation"` (guarded by the preceding `else if (member.Kind != "explicit-interface-implementation")`).
- No duplicate-`static` path; modifiers built once before the signature.
- RTS getter/setter paths propagate `MethodAttributes.Static` into `IsStatic`, and `ToApiMember` preserves it with the explicit-impl kind.
- Both test suites pass (`*ExplicitInterfaceProperty_Static*`, `*KeepsStaticOnExplicitInterfaceProperty*`, `*ExplicitInterface*`).

**Gemini 3.1 Pro** — no significant issues. Verified the same five attack points plus:
- Compiled a local C# 11 test confirming an explicit static-abstract impl **requires** `static` and **forbids** an access modifier (CS0106 otherwise) — the diff produces exactly this shape.
- Modifier ordering yields valid `static unsafe async`; no pre-existing modifier (`readonly`/`extern`) is dropped by the structural rewrite.
- Traced `IsStatic` provenance in ReturnToSenderTypePlanner.cs; a non-static property safely bypasses the new block.

**Agreement:** both converged — no divergence, nothing to reconcile, no follow-up changes required. No resolution commits apply because there were no actionable findings.

Ready to merge — awaiting @richlander (no self-merge).

PR #2883, opened 2026-07-18
Drain rts-parity row: populate reconstructed enum members in RTS closure (#2777)

richlander:
- Part of #2777 (make ReturnToSender a thin orchestrator; drain the rts-parity burn-down)
- Row: `DotnetInspector.Packages!DotnetInspector.Packages.SymbolPackageDownloader::CheckPdbHeader#0`
- Evidence revision: `ad5acd35`

## Fix

> Should we accept this row fix?

**Conclusion:** **PASS** — the RTS closure surface emitted enum supporting types as member-less `enum { }` shells, so a target body that referenced an enum member by name failed to bind (CS0117), stalling closure reconstruction and forcing the compile-back floor. Populating the reconstructed enum with its named members makes the row recompile Exact under ReturnToSender with no floor.

False output (RTS reconstruction, before):

```csharp
public static PdbHeaderKind CheckPdbHeader(string pdbPath)
{
    ...
    return PdbHeaderKind.Unknown; // CS0117: 'PdbHeaderKind' has no 'Unknown'
    ...
}
public enum PdbHeaderKind
{
}
```

Fixed output:

```csharp
public static PdbHeaderKind CheckPdbHeader(string pdbPath)
{
    ...
    return PdbHeaderKind.Unknown; // binds
    ...
}
public enum PdbHeaderKind
{
    Unknown = 0,
    Portable = 1,
    Windows = 2,
}
```

## Root cause and scope

- **Root cause:** `AddClosureMemberSurface` (`tools/DecompilerHarness/ReturnToSenderTypePlanner.cs`) began with `if (requirement.RequiredKind == CompileBackTypeKind.Enum) return;`, so a reconstructed enum never received any members. Enum-kind types render every member via the printer's `RenderEnumMember` path, so a member-less shell binds only when the target never names a member. `CheckPdbHeader` returns `PdbHeaderKind.Unknown`/`.Portable`/`.Windows` by name → CS0117 → closure-stalled → compile-back floor (status `Exact` but `UsedCompileBackFloor=True`).
- **Fix shape:** Replace the early return with a walk over the enum's literal fields, emitting each as a `Field` member requirement whose `TargetBody` is the formatted constant value (via the existing `TryFormatConstantField`). This maps to a `CSharpFieldInitializer`, which the enum printer renders as `Name = value`. The special `value__` storage field (not a literal) and any name-mangled fields are skipped; members are keyed to avoid duplicating any already-present field.
- **Scope boundary:** Harness member-population change plus a follow-up CSharp-layer fix (`src/ILInspector.CSharp/TypeShellProducer.cs` reproduces the enum underlying type, added after adversarial review — see the Review section). Both stay SRM-only and Roslyn-free. Does not address the other three burn-down rows (distinct causes: cross-assembly enum→int typing, type-as-namespace using synthesis, un-reconstructed closure member).

## Witnesses

| Witness | Before | After |
| --- | --- | --- |
| `SymbolPackageDownloader::CheckPdbHeader#0` (real corpus row) | `RecompileFail: CS0117`, floor used | `Exact`, floor not used |
| `Host::Classify` (compiled fixture, returns nested enum, names 3 members) | member-less enum, CS0117 | `enum Kind { Unknown = 0, First = 1, Second = 2 }`, `Exact`, no floor |

## Evidence

| Check | Result |
| --- | --- |
| Full `dotnet-inspect.slnx` build (Release) | Pass, 0 errors |
| Focused test | `ReturnToSenderPrototypeTests.CompileBackTargets_PopulatesEnumMembersWhenTargetReferencesThemByName` passes |
| RTS oracle suites | Prototype 121 / FixtureCatalog 82 / Evidence 3 / GeneratedFilter 25 / MemberBodyFacts 10 — all green |
| Real witness | `CheckPdbHeader` recompiles Exact via RTS without the compile-back floor |

## Corpus validity

> Did this row fix introduce new rts-parity regressions?

**Conclusion:** **PASS** — enforcing gate exit 0, no new regressions; the fixed row is flagged resolved and dropped from the manifest (4 → 3).

Run: 14-assembly prepared corpus (`eng/prepare-decompiler-corpus.sh`), `--corpus-fidelity-oracle rts-parity --corpus-fidelity-cap 3 --compile-cap 0`, enforcing `--rts-parity-burndown`.

| Metric | Before | After |
| --- | ---: | ---: |
| rts-parity burn-down rows | 4 | 3 |
| New Exact→RecompileFail regressions | - | 0 |

Manifest change: removed `DotnetInspector.Packages!DotnetInspector.Packages.SymbolPackageDownloader::CheckPdbHeader#0`.

- `LibraryBodyIndex::HollowUnsafeMethods` — cross-assembly enum→int typing (CS0266); product raise concern.
- `Base64Encoder::WriteCharsAsync` — using synthesis treats a type as a namespace (CS0138).
- `ParseResult::GetValue#2` — closure member not reconstructed (CS0234).

## Review

| Reviewer | Round 1 (`ad5acd35`) | Re-review (`be11af7b`) |
| --- | --- | --- |
| GPT-5.6 Sol | Found non-int enum CS0266 (verified) | CLEAN (raised a malformed-`value__` concern, then withdrew it — see below) |
| Gemini 3.1 Pro | Found non-int enum CS0266 (verified) | CLEAN |

**Review conclusion:** round 1 finding (both reviewers, same defect) fixed in `be11af7b`; both re-reviews of the fixed head are CLEAN. GPT's re-review malformed-`value__` concern was a false positive (guarded decode degrades to `"object"`, never throws) and was withdrawn. See the reconciliation comments.

## Validation

```bash
# corpus enforcing gate (14 asm, cap 3): exit 0, 3 rows
```

### DIFF
--- a/src/ILInspector.CSharp/TypeShellProducer.cs
+++ b/src/ILInspector.CSharp/TypeShellProducer.cs
@@ -212,6 +212,9 @@ public static class TypeShellProducer
                 && (typeDef.Attributes & TypeAttributes.Interface) == 0,
             IsSealed = (typeDef.Attributes & TypeAttributes.Sealed) != 0,
             IsStatic = IsStaticType(typeDef),
+            EnumUnderlyingType = spec.Kind == CSharpTypeShellKind.Enum
+                ? EnumUnderlyingType(reader, typeDef)
+                : null,
         };
 
         return new CSharpTypePrintRequest(
@@ -235,4 +238,20 @@ public static class TypeShellProducer
             CSharpTypeShellKind.Delegate => "delegate",
             _ => throw new NotSupportedException($"Unsupported C# type-shell kind '{kind}'."),
         };
+
+    // An enum's underlying type is carried by its special `value__` instance field.
+    // A reconstructed enum that omits a non-int base (e.g. `: long`) defaults to int
+    // and fails to bind members whose constant values do not fit int (CS0266), so the
+    // shell must reproduce it. The declaration writer suppresses the redundant `: int`.
+    static string? EnumUnderlyingType(MetadataReader reader, TypeDefinition typeDef)
+    {
+        foreach (var fieldHandle in typeDef.GetFields())
+        {
+            var field = reader.GetFieldDefinition(fieldHandle);
+            if (reader.GetString(field.Name) == "value__")
+                return MetadataDeclarationQuery.GetField(reader, typeDef, field).ReturnType;
+        }
+
+        return null;
+    }
 }
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -4854,6 +4854,49 @@ public class ReturnToSenderPrototypeTests
         }
     }
 
+    [Fact]
+    public void CompileBackTargets_PopulatesEnumMembersWhenTargetReferencesThemByName()
+    {
+        // A target method that returns a nested enum and references several of its
+        // members by name forces the enum to be reconstructed as a closure supporting
+        // type. A member-less `enum { }` shell cannot bind those references (CS0117)
+        // and drops the row to the compile-back floor; the reconstructed enum surface
+        // must carry its named members with their constant values.
+        var assemblyPath = CompileFixture("""
+            public class Host
+            {
+                public enum Kind { Unknown, First, Second }
+
+                public static Kind Classify(int value)
+                {
+                    if (value == 1)
+                        return Kind.First;
+                    if (value == 2)
+                        return Kind.Second;
+                    return Kind.Unknown;
+                }
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("Host", "Classify", 0)]));
+
+            Assert.NotEqual(FidelityCheck.CompileBackStatus.RecompileFail, result.Status);
+            Assert.False(result.UsedCompileBackFloor, result.Detail);
+            Assert.NotNull(result.Source);
+            Assert.Contains("enum Kind", result.Source);
+            Assert.Contains("Unknown = 0", result.Source);
+            Assert.Contains("First = 1", result.Source);
+            Assert.Contains("Second = 2", result.Source);
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
     static string CompileFixture(
         string source,
         string? directory = null,
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -4897,6 +4897,45 @@ public class ReturnToSenderPrototypeTests
         }
     }
 
+    [Fact]
+    public void CompileBackTargets_ReconstructsNonIntEnumUnderlyingTypeForMemberValues()
+    {
+        // A reconstructed enum that names members whose constant values do not fit
+        // `int` (long/ulong/uint, negative, or byte-backed) must reproduce the enum's
+        // underlying type. Otherwise the shell defaults to `int` and the emitted
+        // members fail to bind (CS0266), dropping the row to the compile-back floor.
+        var assemblyPath = CompileFixture("""
+            public class Host
+            {
+                public enum ELong : long { A = 0, B = 2147483648L, C = -1L }
+                public enum EULong : ulong { None = 0, All = 18446744073709551615UL }
+                public enum EUInt : uint { Z = 0, Top = 2147483648 }
+                public enum EByte : byte { Lo = 0, Hi = 255 }
+
+                public static ELong GetL(long v) => v == 0 ? ELong.A : (v == 1 ? ELong.B : ELong.C);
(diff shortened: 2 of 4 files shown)


richlander:

Two independent adversarial reviews ran against head `ad5acd35` in isolated worktrees (GPT-5.6 Sol, Gemini 3.1 Pro; neither my own model family). **Both independently found the same real defect**, so it is verified and now fixed.

### Finding (both reviewers, verified) — non-int-backed enums fail CS0266
The round-1 commit populated enum members but did not reproduce the enum's **underlying type**. A reconstructed `long`/`ulong`/`uint`/negative/`byte`-backed enum defaulted to `int`, so emitted member values (e.g. `B = 2147483648L`, `All = 18446744073709551615UL`) failed to bind (`CS0266: Cannot implicitly convert type 'long' to 'int'`) and dropped the target to the compile-back floor.

- GPT-5.6 Sol: reproduced against `EnumCastSamples` targets (`LongEnumArray`, `ULongEnumBoxedMax`, `UnsignedEnumConditionalArm`), all `UsedCompileBackFloor=True`; pointed at `TypeShellProducer` never setting `ApiType.EnumUnderlyingType`.
- Gemini 3.1 Pro: reproduced with an injected `enum ELong : long { A, B = 2147483648L, C = -1L }` fixture; same `CS0266`; traced to `TypeShellProducer.BuildPrintRequest` not extracting `value__`.

Both point to the same fix location and root cause — high confidence.

### Resolution — commit `be11af7b`
`TypeShellProducer.BuildPrintRequest` now sets `ApiType.EnumUnderlyingType` from the enum's special `value__` field via `MetadataDeclarationQuery.GetField`. The declaration writer already suppresses the redundant `: int`, so int-backed enums (including this PR's `CheckPdbHeader` row) are unchanged; non-int enums now emit `enum E : long { ... }`. Correct layering: CSharp owns the C# type view, Metadata owns the `value__` decode; SRM-only, Roslyn-free.

Added regression test `CompileBackTargets_ReconstructsNonIntEnumUnderlyingTypeForMemberValues` (long/ulong/uint/byte enums, all recompile Exact with no floor).

Re-validated at `be11af7b`: full slnx build 0 errors; `TypeShellProducerTests` 11 green; RTS oracle suites green (Prototype 122 / FixtureCatalog 82 / Evidence 3 / GeneratedFilter 25 / MemberBodyFacts 10); corpus rts-parity enforcing gate exit 0, 3 rows, no new regressions.

Re-review of the fixed head `be11af7b` is in progress with both reviewers; I will update the review table once both are clean.

richlander:
## Adversarial re-review (fixed head `be11af7b`) — both CLEAN

Both reviewers re-ran against the fixed head in isolated worktrees. Final result: **two CLEAN reviews**.

### Gemini 3.1 Pro — CLEAN
Verified the round-1 CS0266 fix (long/ulong/uint/byte enums recompile Exact, no floor); confirmed graceful degradation when `value__` is missing or its signature fails to decode (→ `"object"` → writer omits the base → default int, no throw); confirmed the only non-test caller of `TypeShellProducer.BuildPrintRequest` is RTS (zero collateral damage) and int-backed enums stay byte-neutral (`: int` suppressed); confirmed `char`/`bool`-backed IL enums degrade cleanly via the writer's `_ => null` arm. Corpus gate exit 0, 3 rows; `CheckPdbHeader` recompiles Exact.

### GPT-5.6 Sol — CLEAN (one finding raised, then withdrawn)
Initially raised a Medium: "malformed `value__` signatures can throw `BadImageFormatException` from `TypeShellProducer.cs:250`; the new path lacks graceful fallback."

**Verified and dismissed.** I traced the exact path: `MetadataDeclarationQuery.GetField` → `GuardedSignatureText.FieldText` → `GuardedSignatureDecoder.Decode` → `SignatureDecoder.Decode<T>` (`src/ILInspector.MetadataPrimitives/SignatureDecoder.cs:73`), whose body wraps both the preflight prescan and `decode()` in `catch (BadImageFormatException)` / `catch (ArgumentOutOfRangeException)`, returning a Malformed result. `ProjectDecode` then yields the degraded `"object"` (with `SignatureDecodeStatus.Degraded`), and the writer's `EnumUnderlyingKeyword("object")` hits `_ => null` → no base suffix → default int. No exception reaches `TypeShellProducer.cs:250`.

On reconsideration the reviewer **withdrew the finding**, noting it had conflated `ILInspector.Decompiler.GuardedSignatureText` (which uses `GetValueOrThrow`) with the `ILInspector.Metadata` `GetField` degrade-to-`"object"` path actually used here — confirmed: `src/ILInspector.Decompiler/GuardedSignatureText.cs:26` uses `GetValueOrThrow`, `src/ILInspector.Metadata/MetadataDeclarationQuery.cs:403` uses `ProjectDecode`. No resolution commit is needed (false positive); no fallback swallowing was added, which is correct per the repo's "keep failure visible" rule (the guard already degrades visibly).

### Agreement / divergence
Round 1: both reviewers independently found the same real CS0266 defect (verified, fixed in `be11af7b`). Re-review: Gemini CLEAN; GPT raised one malformed-metadata concern that both the code trace and Gemini's independent analysis refuted, and GPT withdrew it. No standing findings.

richlander:
**Ready to merge** (pending @richlander approval).

Two CLEAN adversarial re-reviews on head `be11af7b` (GPT-5.6 Sol + Gemini 3.1 Pro); the one re-review finding was traced to a false positive and withdrawn. CI green (build-net10, test, changes), MERGEABLE/CLEAN, 0 behind `origin/main`. rts-parity burn-down: 4 → 3 rows, no new Exact→RecompileFail regressions.

PR #2852, opened 2026-07-18
Redefine compile-back fidelity contract V1

richlander:
## Conclusion

Redefines compile-back `Exact` now under one versioned fidelity contract.
Contract V1 covers the complete measurement body: opcode agreement, a
harness-owned policy over product IL-body comparison mechanics, outcome
meanings, corpus metrics, and reporting. It does not retain the superseded
parallel operand-fidelity side channel.

Closes #2816.

## Contract V1

| Outcome | Meaning |
| --- | --- |
| `Exact` | Opcode names match and the contract V1 IL-body comparison matches. |
| `OpcodeDiff` | A `Full` method recompiles to a different opcode stream. |
| `OperandDiff` | Opcode names match, but a V1-observable value, symbolic target, or branch target differs. |
| `FidelityUnavailable` | The product body comparison cannot produce a verdict. |
| `NotFull` | The method is below `Full`; any observed body difference is not promoted to a fidelity defect. |
| `RecompileFail` / `ContextFail` | Compilation or reconstructed-context failure remains explicit. |

`ILInspector.Instructions` exposes only domain-neutral `IlBodyDiffOptions` for
variable-layout, current-assembly-scope, and platform-assembly-scope
normalization. DecompilerHarness owns the named V1 option bundle. Numeric
values, strings, non-platform symbolic identities, member/type targets,
value/address loads, and branch topology remain observable.

`Exact` is exact only under this declared contract. V1 remains EH-blind and is
not a semantic-equivalence claim.

## Schema and compatibility

- Corpus snapshot schema is **v5**; fidelity contract version is independently
  **1**.
- Schema-v4 snapshots deserialize as legacy contract version `0`.
- Comparisons across fidelity contract versions are rejected rather than
  presenting incomparable movement.
- Method-delta artifacts use schema v2 and record baseline/current contract
  versions.
- The old per-method `operandFidelity`, operand diff side channel, and parallel
  aggregate metrics are removed.
- The default general-purpose `IlBodyDiff` comparison remains option-free and
  unchanged.

## Measured baselines

All affected corpus profiles were regenerated mechanically and replayed. The
final fixed review head is `2fd3ffdd` on current main `6b51bb8b`.

| Profile | Contract V1 result |
| --- | --- |
| Real-world, cap 50 | 600 checked: 502 Exact, 90 OpcodeDiff, 8 OperandDiff, 0 unavailable |
[...]

Issue #2819, opened 2026-07-17
Broad NuGet top-packages discovery sweep (successor to #622)


richlander:
## Conclusion

Stand up a broad NuGet top-packages discovery sweep. The package list
`docs/data/nuget-top-packages.json` exists but no lane sweeps it, and breadth
is where
quality falls off: fully-raised slides from 96.9% (CoreLib) to 88.0% (pinned
real-world) to 83.0% (mixed PR-quick set). Successor to #622, whose
remaining recommendations this carries forward. Umbrella #2814.

## Shape

- **Discovery tier, not a gate**, per #2561's corpus ladder (synthetic →
  generated-family → pinned real package → pinned platform → live/discovery).
  The sweep never blocks a PR; it produces work.
- Periodic (weekly or on-demand) sweep over the top-package list: quality
  card + a new-defect-class docket routed into #2338-style partitioned
  backlogs, the same conversion path that made the validity sweeps the
  highest-yield lane to date.
- Promotion path: shapes worth holding get a pinned-subset row in the
  existing baseline system rather than a new bespoke format (#2561
  one-baseline protocol).
- Named triage owner and cadence declared before the first run counts
  (#1584); sweep cost bounded and recorded.

## Prior art carried from #622

The 2026-06-22 sweep-strategy thread on #622 (progressive burndown by
10-point increments, adversarial corpora) informs the cadence design; #622's
delivered recommendations (pinned corpus trend gates, checked-in baselines,
root-cause bucketing) are why this issue is a sweep lane and not a baseline
system.

## Acceptance

- First sweep published with: per-package fully-raised/validity summary,
  the new-defect-class docket, and the promotion candidates.
- Second sweep demonstrates the loop: at least one docket class fixed or
  explicitly non-actioned between runs.

richlander:
### First multi-agent discovery run — product findings

Exact product revision: `b0e1bc11058255c54fc942f0220874e93b707cbe`

Two isolated agents swept disjoint portions of
`docs/data/nuget-top-packages.json` using the existing
`--library-report` / `--validity-check` / targeted `--dump` paths. Candidates
were independently reproduced on a clean exact-head checkout and duplicate
checked before filing.

Pattern-pivoted product issues:

- #2828 — `EhStructuringPass` folds a catch-entry exception store into a catch
  variable that is already live outside the handler, producing `CS0136` at
  `Full`. Real hits: Npgsql 10.0.3.
[...]

Issue #2816, opened 2026-07-17
Add an operand-aware comparison tier to the harness opcode oracle


richlander:
## Conclusion

Add an operand-aware comparison tier to the harness fidelity oracle — by
threading the **product-owned** `ILInspector.Instructions.IlBodyDiff` result
through `FidelityCheck` and the corpus projection, not by writing a second
canonicalizer. `HarnessOpcode.Canonicalize` folds opcode families and compares
**names only**: `ldc.i4.5` vs `ldc.i4.7` compare Exact; call targets,
field/string tokens, and branch topology are invisible. This bounds the
semantic depth of compileback, ReturnToSender, and the rts-parity gate
simultaneously. Umbrella #2814; strengthens the deepest oracle per #1212.

## Approach: reuse the product operand diff

`IlBodyDiff` already resolves numeric/string/member/type operands and
validates branch/switch target correspondence, with compiled-fixture
negatives for constants, strings, calls, fields, and branch retargeting. RTS
already carries this result as `ReturnToSender.Result.IlDiff` via
`ImplementationDiff.CompareMembers`. The missing work is threading that
product result through `FidelityCheck` and the corpus projection as a second
verdict column. A harness-side reimplementation beside `HarnessOpcode` would
violate the harness boundary (AGENTS.md) and create two operand oracles.

## Normalization contract (define before naming the column)

The tiers are not monotonic as implemented today: `HarnessOpcode` folds all
`ldarg*`/`ldloc*`/`stloc*` macro families, while `IlBodyDiff` preserves
macro forms and raw slot numbers. A direct `IlDiff.IsExact` projection would
classify harmless compiler-local-layout differences as operand divergence.
Before the column ships:

- specify which differences the tier intends to flag (call/field/string/
  numeric/control-flow identity) vs tolerate (local macro form, slot
  layout) — or include layout differences and measure them separately;
- pair the contract with **positive compatibility cases** for the
  intentional macro/slot differences, not only the close negatives.

## Scope bound: EH

`IlBodyDiff` deliberately does not compare exception-region tables. Identical
opcodes, operands, and branch targets under different protected/handler
regions can still behave differently. Either include EH topology in the tier
or state explicitly that the new column is EH-blind and is **not** semantic
equivalence.

## Additive — controlled baseline migration

The new verdict is a parallel column; existing `Exact` semantics stay
stable, so `rts-parity-burndown.json` remains byte-for-byte unchanged. Corpus
[...]

PR #2885, opened 2026-07-18
Add classic state-machine corpus profile (#2818)

### LINKED ISSUE #2818: Add a classic async/iterator state-machine corpus lane
## Conclusion

Add a pinned classic async/iterator state-machine corpus lane. No CI corpus
contains a single classic async state machine today: the pinned real-world
corpus and the opt-in net11 corpus are runtime-async era, and the only
visibility is the on-demand `Fixtures.ClassicAsync` overlay — 21 methods,
0 raised. Classic state machines remain the dominant async form in the
real-world ecosystem the tool inspects. Umbrella #2814.

## Measurement caveat, addressed up front

Standalone `--validity-check` does not import sibling bodies, so it
understates classic async reconstruction for kickoff methods; the library
report/dump/fidelity paths use the product-style cross-method import seam
(#1599 row 4). The lane must wire that seam — a 0/21 measured without it
would overstate the gap. Part of this issue's point is that **no corpus lane
wires the seam today**, which is itself the blind spot.

## Shape

- Pinned corpus via the downlevel TFM / LangVersion axis the harness README
  names but never built (it cascades classic async, old switch, and iterator
  lowerings at once).
- Baseline + feature-evidence gates on the opt-in-net11 corpus contract
  (schema, non-decreasing per-feature rows) rather than a new bespoke format
  (#2561 one-baseline protocol).
- Nightly lane with a named triage owner before it counts as a gate
  (#1584 principle: an unwatched signal is not an operational gate).

## Links

- #2256 — sibling gap (per-node fixture guarantee is not per-producer); this
  lane gives its classic-path producers a corpus home.
- #1599 rows 4/8/9; #1584 Stage 9.

## Acceptance

- Corpus committed and pinned; baseline records per-feature raised counts for
  classic async kickoff/resume, iterators, and async iterators.
- First census published: how many of the corpus's state machines raise
  today through the cross-method seam (the honest successor to 0/21).

richlander:
Closes #2818. Child of #2814 (decompiler quality direction) — the next
unstarted Tier 2 item, started ahead of #2819 (Tier 3, blocked by #2814's
sequencing rule) per the docket's "sequence, don't parallelize" rule for
Tier 2 and its "at most two children in flight at once" concurrency cap
(only PR #2852, for #2816, was in flight at the time).

## What

Adds a new pinned corpus lane, `classic-state-machines`, for classic
(`RuntimeAsync=off`) compiler-generated state machines — async
kickoff/resume, iterators, async iterators, and a plain non-pattern
`switch` — none of which had dedicated corpus coverage before. This is
the first published census for these shapes (successor to the earlier
known "0/21" classic-async gap).

- New fixture project `ILInspector.Decompiler.Fixtures.ClassicStateMachines`
  (net11.0, `RuntimeAsync=off`), reusing the same axis as the existing
  `ClassicAsync` fixture rather than an actual downlevel TFM/LangVersion
  pack. This is a deliberate scope decision: classic lowering for these
  four shapes does not depend on an older runtime or compiler, so it
  reaches the same "classic era" lowering set without the added
  complexity/risk of installing a downlevel runtime pack in CI. The
  README's "downlevel TFM/LangVersion" candidate-axis note is updated to
  reflect this and narrowed to the axis that remains open (checked
  arithmetic, plus a genuinely older TFM/LangVersion pack that would
  change the lowering shapes themselves).
- `CorpusSensor` gains `CorpusProfile.ClassicStateMachines`. Unlike the
  other two profiles, this lane wires the cross-method import seam
  (`PassContext.ForImport`, the same helper `--dump`/`--library-report`
  use) into `AnalyzeCompleteness`, so cross-method passes such as
  `ClassicAsyncReconstructionPass` (which pulls in the sibling
  `MoveNext` body) raise the same way they do outside the corpus
  sensor — addressing the issue's specific concern that a standalone
  `--validity-check` understates classic-async reconstruction.
  `real-world` and `opt-in-net11` are untouched and still raise each
  method standalone.
- Feature coverage (`classic-async-methods`, `classic-iterator-methods`,
  `classic-async-iterator-methods`, `switch-methods`) is tagged by the
  method-name prefix contract documented on `ClassicStateMachineFixtures`
  (`Async_`/`Iterator_`/`AsyncIterator_`/`Switch_`), following the same
  `RequiredXFeatures` + `FeatureCoverageFailures` gating pattern already
  used by `opt-in-net11`.
- Commits the first published census as
  `tools/DecompilerHarness/corpus/classic-state-machines-baseline.json`
  and wires a `--diff-corpus-baseline` step into the `census` CI lane
  (not `nightly`, since this fixture needs no special SDK).
- Registers the fixture in `dotnet-inspect.slnx` and `FixtureCatalog.cs`.
- Documents the new profile in `tools/DecompilerHarness/README.md`.

## Evidence

- `dotnet build src/ILInspector.Decompiler.Fixtures.ClassicStateMachines -c Release`
  and `dotnet build tools/DecompilerHarness -c Release` both succeed.
- Ad hoc honest census via `--library-report` before wiring the seam:
  23/73 Full. After wiring the seam via the new corpus profile, the
  same population's feature-coverage evidence is captured cleanly (0
  pass bugs, 0 crashes); the committed baseline records the current
  raise/residual state as a first census, not a claimed fix — the
  dominant residual (`DEC0009/state-machine-type-name`, 45 methods) is
  out of scope for this PR.
- `--diff-corpus-baseline` against the committed baseline on an
  unmodified rebuild reports "Corpus sensor matched baseline" (exit 0).
- Targeted tests: `dotnet run --project src/ILInspector.Decompiler.Tests
  -c Release -- -class ILInspector.Decompiler.Tests.CorpusSensorComparisonTests`
  → 48/48 passed. The full `ILInspector.Decompiler.Tests` suite includes
  several pre-existing long-running tests unrelated to this change
  (`SkeletonEmitTests`, `ClusterCaptureTests`) that run well past 20
  minutes each in this environment; given this PR only touches
  `tools/DecompilerHarness` (not `src/ILInspector.Decompiler`), the
  targeted class plus clean builds is the proportionate check per
  AGENTS.md's evidence guidance. Happy to run the full suite if a
  reviewer wants it as additional evidence.

## Non-actions

- Does not change `real-world` or `opt-in-net11` corpus behavior;
  `AnalyzeCompleteness`'s existing branches are untouched.
- Does not attempt to fix the `DEC0009/state-machine-type-name`
  residual this census surfaces — that's tracked as follow-up decompiler
  work, not part of this harness-surface PR.

### DIFF
--- a/.github/workflows/deep-inspect.yml
+++ b/.github/workflows/deep-inspect.yml
@@ -126,6 +126,20 @@ jobs:
             --corpus-fidelity-cap 50 \
             --max-examples 3
 
+      - name: Build classic state-machine corpus fixture
+        run: dotnet build src/ILInspector.Decompiler.Fixtures.ClassicStateMachines -c Release
+
+      - name: Run classic state-machine corpus sensor
+        shell: bash
+        run: |
+          mkdir -p artifacts/deep-inspect
+          dotnet run --project tools/DecompilerHarness -c Release -- \
+            artifacts/bin/ILInspector.Decompiler.Fixtures.ClassicStateMachines/release/ILInspector.Decompiler.Fixtures.ClassicStateMachines.dll \
+            --corpus-profile classic-state-machines \
+            --emit-corpus-snapshot artifacts/deep-inspect/classic-state-machines-snapshot.json \
+            --diff-corpus-baseline tools/DecompilerHarness/corpus/classic-state-machines-baseline.json \
+            --max-examples 10
+
       - name: Run return-address equivalence census
         shell: bash
         run: |
--- a/dotnet-inspect.slnx
+++ b/dotnet-inspect.slnx
@@ -33,6 +33,7 @@
     <Project Path="src/ILInspector.Findings/ILInspector.Findings.csproj" />
     <Project Path="src/ILInspector.Decompiler.Fixtures.CheckedArithmetic/ILInspector.Decompiler.Fixtures.CheckedArithmetic.csproj" />
     <Project Path="src/ILInspector.Decompiler.Fixtures.ClassicAsync/ILInspector.Decompiler.Fixtures.ClassicAsync.csproj" />
+    <Project Path="src/ILInspector.Decompiler.Fixtures.ClassicStateMachines/ILInspector.Decompiler.Fixtures.ClassicStateMachines.csproj" />
     <Project Path="src/ILInspector.Decompiler.Fixtures.Ladder/ILInspector.Decompiler.Fixtures.Ladder.csproj" />
     <Project Path="src/ILInspector.Decompiler.Fixtures.LegacyUnsafe/ILInspector.Decompiler.Fixtures.LegacyUnsafe.csproj" />
     <Project Path="src/ILInspector.Decompiler.Fixtures.NewUnsafe/ILInspector.Decompiler.Fixtures.NewUnsafe.csproj" />
--- a/src/ILInspector.Decompiler.Fixtures.ClassicStateMachines/ClassicStateMachineFixtures.cs
+++ b/src/ILInspector.Decompiler.Fixtures.ClassicStateMachines/ClassicStateMachineFixtures.cs
@@ -1,0 +1,157 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace ILInspector.Decompiler.Fixtures.ClassicStateMachines;
+
+/// <summary>
+/// Classic-lowering state-machine fixtures for issue #2818 (child of the #2814
+/// decompiler-quality direction, successor to #622's broader-coverage
+/// recommendations). Compiled with <c>runtime-async=off</c> (see the .csproj),
+/// so every await here lowers to a classic <c>AsyncTaskMethodBuilder</c> state
+/// machine rather than the runtime-async <c>AsyncHelpers.Await</c> form the rest
+/// of the repo's corpora use. Iterators and async iterators always lower to a
+/// compiler-generated <c>MoveNext</c> state machine regardless of the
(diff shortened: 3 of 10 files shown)


richlander:
## Adversarial review

Two independent reviewers, different model families, isolated worktrees at exact head `5c53c13d`:

- **Reviewer A (GPT-5.6 Sol)**: no blocking or non-blocking issues. Verified build success (fixture + harness), a clean sequential *and parallel* baseline diff (exit 0, 73 methods, 0 pass bugs), correct feature-count classification against actual generated names (no double-counting), that the new import-seam lambda matches `Program.ImportSeam` exactly, additive/correct CI wiring, unchanged gating behavior for the pre-existing profiles, unique/consistent fixture-catalog registration, accurate `RuntimeAsync=off` (not downlevel-TFM) documentation, and 48/48 targeted `CorpusSensorComparisonTests` passing.
- **Reviewer B (Gemini 3.1 Pro)**: no significant issues. Independently verified the same attack points: prefix-detection order (`AsyncIterator_` → `Iterator_` → `Async_`, plus independent `Switch_`) is collision-safe against actual dumped type/method names; traced `MetadataSource` thread-safety (double-checked-locked `EnsureTypeMaps`) under the `Parallel.ForEach` this seam now runs inside, confirming no race; reproduced a clean baseline match (exit 0); confirmed `FeatureCoverageFailures`'s switch-expression refactor preserves `RealWorld`/`OptInNet11` behavior exactly; confirmed CI paths match actual build output; confirmed no `FixtureCatalog.All` consumer has a hardcoded count; confirmed 48/48 targeted tests pass; confirmed the README/XML-doc scope claims (no downlevel TFM) are accurate.

Both reviews are clean with no findings to address. Ready to merge.

PR #2888, opened 2026-07-18
Strengthen classic state-machine corpus evidence

richlander:
## Summary

Strengthens the classic state-machine corpus evidence in the decompiler harness. Stacked on #2885 (base `feature/issue-2818-classic-state-machine-corpus`); this PR's diff is the single commit `b2a1ab6f`.

- Adds classic state-machine fixtures: async void, `ValueTask`, instance generic, await-in-catch/finally, iterators, cancellation, try-finally with disposal.
- Adds per-family coverage metrics to the corpus snapshot: `ClassicStateMachineFeatureMetrics(Population, FullyRaised, Residual)` and a `ClassicStateMachineCoverage` map, built by `BuildClassicStateMachineCoverage`.
- Adds a `ClassicStateMachineCoverageFailures` gate: fails on zero specimens, population drop, fully-raised drop, or residual increase at stable population.

The coverage metric buckets source kickoffs into families by fixture-name prefix (`Async_`/`Iterator_`/`AsyncIterator_`) and skips compiler-generated support types (`Type.Contains('<')`). The **fidelity judgment itself** (`FullyRaised`/`Residual`) is driven entirely by the product's `DecompilationFidelity.Full` result — the harness reads the product's raising decision, it does not reconstruct it.

## Evidence

- Snapshot schema is already `v4` (matches README); the new nullable `ClassicStateMachineCoverage` field deserializes older baselines to `null` and the gate handles that gracefully.
- Harness build + comparison tests: `ILInspector.Decompiler.Tests` — 50 total, 0 failed.
- Baseline JSON round-trips with the new field (verified per-family populations: 8 classic-async, 6 classic-iterator, 4 classic-async-iterator).

## Adversarial review

Two-model review (Gemini 3.1 Pro + MAI-Code-1-Flash) at commit `b2a1ab6f`, each isolated in a separate worktree.

- **Gemini 3.1 Pro:** no blocking issues. Verified name-prefix classifier, `Type.Contains('<')` skip robustness (source kickoffs use backticks, generated types use `<>`), schema/baseline back-compat, and the harness-boundary rule (metric purely aggregates the product's `FullyRaised` output).
- **MAI-Code-1-Flash:** raised a harness-boundary "blocking" concern that the name-prefix classifies raising success. **Dismissed after code adjudication:** the name prefix only assigns a family bucket; `FullyRaised`/`Residual` come from the product's `DecompilationFidelity.Full` (`CorpusSensor.cs:1225`), not from the name. Gemini independently verified this.

**Non-blocking (both reviewers):** a residual regression can be masked if population grows while the fully-raised count stays flat, because the residual check requires exactly stable population. Accepted as by-design for this coarse feature-bucket sub-metric — the exact per-method `Compare()` gate independently tracks individual fidelity regressions. Recorded as a possible follow-up (gate on fully-raised ratio).

### DIFF
--- a/src/ILInspector.Decompiler.Fixtures.ClassicStateMachines/ClassicStateMachineFixtures.cs
+++ b/src/ILInspector.Decompiler.Fixtures.ClassicStateMachines/ClassicStateMachineFixtures.cs
@@ -1,5 +1,8 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace ILInspector.Decompiler.Fixtures.ClassicStateMachines;
@@ -22,11 +25,10 @@ namespace ILInspector.Decompiler.Fixtures.ClassicStateMachines;
 /// <c>Iterator_</c> (classic <c>yield</c> iterator), <c>AsyncIterator_</c>
 /// (classic async iterator — <c>IAsyncEnumerable&lt;T&gt;</c> combining both
 /// state-machine shapes), and <c>Switch_</c> (a plain, non-pattern switch
-/// statement — the old jump-table lowering, TFM/LangVersion-agnostic like the
-/// rest of this fixture, so it is measured here rather than requiring a
-/// downlevel TFM/LangVersion axis).
+/// statement used as a control). This is a representative current-compiler
+/// matrix, not an exhaustive cross-compiler or downlevel-TFM corpus.
 /// </summary>
-public static class ClassicStateMachineFixtures
+public class ClassicStateMachineFixtures
 {
     public static async Task<int> Async_AwaitValue(Task<int> a, int b) => await a + b;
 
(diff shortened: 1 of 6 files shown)


richlander:

Two-model review (Gemini 3.1 Pro + MAI-Code-1-Flash) at commit `b2a1ab6f`, each isolated in its own worktree.

**Gemini 3.1 Pro — clean.** Verified the name-prefix classifier, the `Type.Contains('<')` skip (source kickoffs use backticks in IL; only compiler-generated support types use `<>`), schema/baseline back-compat (schema is already v4; the new nullable field deserializes older baselines to `null`), and the harness-boundary rule (the metric aggregates the product's `FullyRaised` output, adding no compensation).

**MAI-Code-1-Flash — raised one "blocking" harness-boundary concern; dismissed after code adjudication.** MAI claimed `BuildClassicStateMachineCoverage` classifies *raising success* by name prefix. It does not: the name prefix only assigns each fixture to a **family bucket**; the `FullyRaised`/`Residual` counts come entirely from the product's `DecompilationFidelity.Full` result (`CorpusSensor.cs:1225`), not from the name. Bucketing deliberately-named test fixtures by family is standard corpus organization, not a reconstruction of product behavior. Gemini independently verified the same. No code change.

**Non-blocking (both reviewers):** a residual regression can be masked if population grows while the fully-raised count stays flat, because the residual check requires exactly stable population. Accepted as by-design for this coarse feature-bucket sub-metric — the exact per-method `Compare()` gate independently tracks individual fidelity regressions. Recorded as a possible follow-up (gate on the fully-raised ratio).

This is a harness-only change with no product-path impact and no code modifications from review.

richlander:
### Update: MAI-Code-1-Flash finding withdrawn

On re-check, MAI withdrew its harness-boundary finding: "the method name only chooses the family bucket (`CorpusSensor.cs:516-525`); the counts are computed from the `CorpusMethodSnapshot` fields, and `FullyRaised`/`Residual` come from the product analysis outcome, not the fixture name — so the harness-boundary finding is withdrawn; on this point, the change is clean."

Both reviews are now clean (Gemini clean; MAI withdrawn). The only remaining note is the by-design, non-blocking population-masking observation (possible follow-up: gate on the fully-raised ratio). No code changes were needed.

Ready to merge (after the base PR #2885 lands).

PR #2900, opened 2026-07-19
Fix extensions and cache usability findings

richlander:
## Summary

Addresses four findings from the AOT usability pass (pass2). All are output-shape / convention fixes to bring `extensions` and `cache` in line with the rest of the tool.

**extensions**
- **Finding 4 (med):** `--jsonl` (and the default table) baked the overload count into the name (`"Aggregate (3 overloads)"`). Now emits a clean method name plus a separate numeric **Overloads** column. The column is hidden when every row has a single overload, so it only appears when it carries information.
- **Finding 6 (low):** the always-empty **Via** column is dropped when empty.
- **Finding 7 (low):** the **Type** column that repeated the queried type on every row is dropped when uniform.
  - Both Via and Type are hidden *only when uniform/empty*. Reachable (indirect) extensions still surface their distinct Type and Via path.

**cache**
- **Finding 9 (low):** `cache` rendered dashed plaintext unlike every other command. It now renders **markdown tables by default** and honors `--json` / `--table` / `--tsv` / `--jsonl` / `--plaintext`. The legacy dashed layout is still available via `--plaintext`.

Column hiding reuses the existing `MarkoutIgnoreColumnWhen` convention (as in `LibraryInspectionView`).

## Evidence

Built the CLI and verified against the built DLL:
- `extensions` default: Via/Type columns gone, Overloads column present (`3`, `2`, …).
- `extensions --jsonl`: `{"name":"Aggregate","overloads":"3",...}` — clean name + separate field.
- `cache` default: markdown `| Field | Value |` + `## Categories` table.
- `cache --json` valid JSON; `--table`/`--tsv`/`--plaintext` work; `--json --table` correctly rejected by the shared validator.

Tests (xUnit v3): added focused cases and updated the existing overload test.
- `ExtensionsCommandTests` — 20 pass (incl. new hide-when-uniform and show-when-reachable cases).
- `CacheCommandTests` — 6 pass (incl. new markdown-default rendering case).

## Compatibility / non-actions

- Finding 4's numeric value renders as a quoted string in `--jsonl`/`--tsv` (`"overloads":"3"`), matching how Markout stringifies **all** numeric cells tool-wide (e.g. `"callers":"238"`). Making it a true JSON number is a Markout-level change affecting every command — out of scope. The substantive ask (separate field, not baked into the name) is met.
- `cache clear` / session-clear paths are unchanged.

### DIFF
--- a/src/dotnet-inspect.Tests/CacheCommandTests.cs
+++ b/src/dotnet-inspect.Tests/CacheCommandTests.cs
@@ -5,6 +5,7 @@ using DotnetInspector.Options;
 using DotnetInspector.Packages;
 using DotnetInspector.Views;
 using Markout;
+using System.Text.Json;
 
 namespace DotnetInspector.Tests;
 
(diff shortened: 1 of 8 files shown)


richlander:

Two-model review per the repo roster. Reviewers isolated in separate worktrees at exact head; scratch under `/tmp`.

Base `c317fdd9` · initial head `c3411762` · fixed head `34213c47`.

### Round 1 findings

**GPT-5.5**
1. *(med)* Reachable-only extensions hid the `Type` column (`TypeUniform` fired on any uniform type). **Fixed** — `Type` now suppressed only when every row is direct (all `Via` empty) and uniform.
2. *(med)* Empty cache ignored machine flags — `--json`/`--jsonl`/`--table`/`--tsv` all emitted `Cache is empty.` plaintext. **Fixed** — machine formats now emit a structured record (empty `categories:[]`); human formats keep the friendly message.
3. *(med)* `cache --jsonl` emitted a blank separator line (invalid JSONL). **Fixed** — routes through the single-line `CacheInfoJson` DTO.

**Gemini 3.1 Pro**
1. *(high)* Same `Type`-column bug as GPT #1. **Fixed** (same change).
2. *(med)* `TypeUniform` nullable-contract: `rows?.Select(...).Distinct()` would throw `ArgumentNullException` if invoked directly. **Fixed** — rewrote with an explicit `rows is null || ...` guard.
3. *(low)* `--table` suppresses the "cache clear" footer, unlike default markdown/`--plaintext`. **Dismissed (by design)** — `--table`/`--tsv`/`--jsonl` share the columnar/projected writer path; the narrative footer belongs only to the two narrative human formats, and appending prose after a columnar table pollutes piped output. Gemini accepted this rationale on re-review.

Resolution commit: `34213c47` (plus 3 new regression tests).

### Round 2 — fixed-head re-review (`34213c47`)

- **GPT-5.5:** rebuilt, tests pass, all prior findings resolved. *"No significant issues found."*
- **Gemini 3.1 Pro:** rebuilt, full suite passes, all prior findings verified resolved. *"No significant issues found. Clean!"*

Both reviewers clean on the fixed head.

richlander:
Ready to merge

- Two clean fixed-head adversarial reviews (GPT-5.5, Gemini 3.1 Pro) at `34213c47` — reconciliation above.
- CI green: `changes`, `build-net10`, and `test (ubuntu-26.04, linux-x64)` all pass; `pack`/`markdownlint` skipped (path-gated, no Markdown changes).
- Branch is current with `origin/main`.

PR #2901, opened 2026-07-19
Raise tuple comparison trees to switch expressions (#2867)

### LINKED ISSUE #2867: Decompiler: raise tuple comparison trees to switch expressions
## Current state

PR #1682 replaced the `Quadrant` fixture's flat goto ladder with valid nested
`if`/`return` control flow. Its scope explicitly stopped before switch-expression
recovery:

```csharp
if (x > 0)
{
    if (y > 0)
        return "I";
    if (y < 0)
        return "IV";
}
// remaining comparison arms
return "axis";
```

That was the correct intermediate fix for #1631, but it is not the final raised
form.

## Fully raised goal

```csharp
return (x, y) switch
{
    (> 0, > 0) => "I",
    (< 0, > 0) => "II",
    (< 0, < 0) => "III",
    (> 0, < 0) => "IV",
    _ => "axis",
};
```

## Required slices

1. Recognize a structured comparison-tree return selector whose tests read the
   same tuple components and whose leaves are value returns.
2. Prove arm ordering, exhaustiveness/default ownership, single evaluation,
   and absence of external entries or shared effectful subtrees.
3. Build tuple relational-pattern switch-expression arms using existing typed
   pattern and switch-expression IR rather than recovering syntax from text.
4. Preserve the nested `if` fallback for incomplete, effectful, shared, or
   non-exhaustive trees.

## Evidence

- Historical intermediate raise: #1682
- Closed intermediate defect: #1631
- Compiler fixture:
  `ILInspector.Decompiler.Fixtures.LadderRung5.Program::Quadrant`
- Current gate intentionally requires nested comparisons and returns, not a
  switch expression.

Add compiler-produced positive and close-negative fixtures, compile-back
validity/fidelity, and a scorecard row for the fully raised switch expression.

richlander:
- Closes #2867
- Changes exhaustive integer-like comparison trees into typed tuple relational-pattern switch expressions
- Card revision: `44327a910d997b00dadc4e0b411ebad0e0ef8a4c`

## Change

**Conclusion:** **PASS** — the compiler-produced `Quadrant` comparison tree now
returns to its tuple-switch idiom, close negatives remain structured, the
target fixtures roundtrip to the exact original opcode stream, and both
fixed-head reviewers are clean.

### Original source

```csharp
public string Quadrant(int x, int y)
{
    return (x, y) switch
    {
        (> 0, > 0) => "I",
        (< 0, > 0) => "II",
        (< 0, < 0) => "III",
        (> 0, < 0) => "IV",
        _ => "axis",
    };
}
```

### Before

```csharp
if (x > 0)
{
    if (y > 0)
    {
        return "I";
    }
    else
    {
        if (y < 0)
        {
            return "IV";
        }
        else
        {
            return "axis";
        }
    }
}
else
{
    if (x < 0)
    {
        if (y > 0)
        {
            return "II";
        }
        else
        {
            if (y < 0)
            {
                return "III";
            }
            else
            {
                return "axis";
            }
        }
    }
    else
    {
        return "axis";
    }
}
```

### After

```csharp
return (x, y) switch
{
    (> 0, > 0) => "I",
    (> 0, < 0) => "IV",
    (< 0, > 0) => "II",
    (< 0, < 0) => "III",
    _ => "axis",
};
```

### Fully raised

The arm order differs from the original source, but the relational patterns are
mutually exclusive, so the ordering change is semantics-neutral.

## Root cause and scope

- **Root cause:** Roslyn lowers tuple relational-pattern switches to nested
  comparison/return trees. Existing structuring recovered valid nested
  `if` statements but did not recover the shared tuple places, constraints,
  leaf values, or default ownership.
- **Fix shape:** `TupleSwitchExpressionPass` proves two or more independent
  integer-like places, one typed anchor per component, satisfiable relational
  intersections, exhaustive/default ownership, and structurally equal default
  values before moving the surviving nodes into typed
  `TupleSwitchExpression` IR.
- **Signedness and constants:** ordering comparisons must match csc's signed or
  unsigned opcode for the place type; anchors must fit that type. Character
  anchors render as character literals so relational patterns remain valid C#.
- **Scope boundary:** incomplete, effectful, externally entered, shared,
  unsatisfiable, mixed-type, signedness-mismatched, or non-exhaustive trees
  retain their structured fallback.
- **Safety:** the pass recovers structure from typed IR and moves nodes; it does
  not infer syntax from rendered text.

## Evidence

| Check | Result |
| --- | --- |
| Product CI | Green |
| Focused shape/dataflow gates | `TupleSwitchExpressionPassTests`, `LadderRung5GateTests`, and `DataflowFactsTests`: 32 passed |
| Exact-head fidelity gates | `FidelityGateTests` and `LoweredFidelityGateTests`: 7 passed |
| Compiler breadth | `Octant` (three components), `UIntQuadrant` (unsigned), and `CharQuadrant` (character anchor) |
| Reduced fixture fidelity | Target fixtures recompile to the exact original opcode stream; no `KnownDiffs` entry |
| Real witness | `LadderRung5.Program::Quadrant` changed from nested comparisons to a tuple switch expression |
| Review | Gemini 3.1 Pro and MAI-Code-1-Flash clean at `44327a91` after findings were fixed |

## Decompiler quality

**Conclusion:** **PASS** — the pinned ladder and fidelity gates cover the
changed target population, all target fixtures are opcode-exact, and no
known-difference docket entry was added.

### Targeted quality gate

| Witness | Before | After | Fidelity |
| --- | --- | --- | --- |
| `LadderRung5.Program::Quadrant` | Nested `if`/`return` comparison tree | Tuple relational-pattern switch expression | Exact |
| `CfgSampleClass.Octant` | Comparison tree | Three-component tuple switch expression | Exact |
| `CfgSampleClass.UIntQuadrant` | Unsigned comparison tree | Unsigned tuple relational-pattern switch expression | Exact |
| `CfgSampleClass.CharQuadrant` | Character comparison tree | Character-literal tuple switch expression | Exact |

Close negatives cover incomplete/default ownership, effectful or shared places,
unsatisfiable intersections, anchor range/type mismatches, signedness
mismatches, mixed-type slot reuse, dataflow sibling-arm leakage, and printer
precedence.

## Validation

```bash
  -class "ILInspector.Decompiler.Tests.TupleSwitchExpressionPassTests"
  -class "ILInspector.Decompiler.Tests.LadderRung5GateTests"
  -class "ILInspector.Decompiler.Tests.DataflowFactsTests"
  -class "ILInspector.Decompiler.Tests.FidelityGateTests"
  -class "ILInspector.Decompiler.Tests.LoweredFidelityGateTests"
```


richlander:
## Adversarial review reconciliation (fixed head 44327a91)

Two-model review (Gemini 3.1 Pro + MAI-Code-1-Flash), each isolated in its own worktree with real compile-back evidence, first at the pre-fix head and again at the fixed head.

### First-round findings — both addressed in 44327a91

- **Gemini — char anchors (blocking, confirmed real):** a `char` component's anchor is an in-range `Int32` constant in IL that `ConstantFits` admits, but a relational pattern against a `char` input rejects a bare int literal (`CS0266`; the implicit constant-expression conversion does not apply in patterns). Reproduced independently. **Fix:** the printer threads each governing component's declared type to `PositionalSubpatternText`, spelling a char anchor as a char literal (`> 'A'`). Confirmed char-only — other integer-like types keep bare-int anchors. Regression: `CharQuadrant`, exercised by the pass test and the fidelity-gate roundtrip.
- **MAI — mixed-type slot reuse (defense-in-depth, unreachable):** MAI fed synthetic IR loading one slot as both `Int32` and `UInt32` and never ran compile-back. That state is unreachable — the IR raiser types every load of a slot from the fixed signature, and a `(uint)x` is a `Convert`, not a bare load. **Fix:** `TryDecompose` now requires a repeated component's `ResultType` to equal the current place type, declining mixed-type reuse. Regression: `RepeatedComponentWithDifferentPlaceType_DeclinesFold`.

### Fixed-head re-review — both clean

- **Gemini:** clean. Compile-back at Full fidelity emits `> 'A'` (no `CS0266`); both inline and return-position paths pipe component types; `TryCharConstantText` guarded behind `IsCoreChar`; type-guard uses `TypeRef.Equals` and does not over-decline same-type quadrants; the pre-existing `x is (...)` path (now `targetType: null`) is unchanged.
- **MAI:** clean. Real char compile-back over a 6×6 grid matched for the return form, an inline form, and Equal/NotEqual arm kinds; verified int fixtures still emit bare anchors; verified the type guard declines its original case and folds normal same-type quadrants.

### Validation

- `TupleSwitchExpressionPassTests` + `LadderRung5GateTests` + `DataflowFactsTests`: 32 tests, 0 failed.
- `FidelityGateTests` + `LoweredFidelityGateTests` on this exact head: 7 tests, 0 failed. `CharQuadrant`/`Octant`/`UIntQuadrant` roundtrip to the exact original opcode stream (no `KnownDiffs` entry).

Note: this branch had advanced on origin with a prior-round `DefiniteAssignment` sibling-arm-leak fix (`e5743369`/`19ded73c`/`65f46604`, in `DefiniteAssignment.cs` + `DataflowFactsTests.cs`); the review fixes were rebased cleanly on top with no overlap.

Ready to merge

Issue #2877, opened 2026-07-18
Decompiler: raise pattern-guarded short-circuit && diamond, inline default temp, and re-compose fluent chain (BePositive residue)


richlander:
Follow-up raising residue observed on the `#2872`/`#2873` witness member `FluentAssertions.Numeric.NumericAssertionsBase\`3::BePositive` (net6.0). After #2873 recovers the `is T subject` cache-bind, the method is correct but still not fully raised relative to authored source.

**Authored source** (SourceLink, via `dotnet-inspect member NumericAssertionsBase BePositive --package FluentAssertions -S "Original Source"`):

```csharp
CurrentAssertionChain
    .ForCondition(Subject is T subject && subject.CompareTo(default) > 0)
    .BecauseOf(because, becauseArgs)
    .FailWith("Expected {context:value} to be positive{reason}, but found {0}.", Subject);
```

**Current decompiled output:**

```csharp
AssertionChain S_256 = CurrentAssertionChain;
if (Subject is T subject)
{
    V_2 = default;
    S_1 = subject.CompareTo(V_2) > 0;
}
else
{
    S_1 = false;
}
object[] S_257 = new object[1];
AssertionChain S_258 = S_256.ForCondition(S_1).BecauseOf(because, becauseArgs);
S_257[0] = Subject;
S_258.FailWith("Expected {context:value} to be positive{reason}, but found {0}.", S_257);
```

## Remaining slices

1. **Short-circuit `&&` reconstruction.** The `if (Subject is T subject) { S_1 = <rhs>; } else { S_1 = false; }` boolean-assignment diamond should raise to `Subject is T subject && <rhs>` (a pattern-guarded short-circuit `&&` producing a bool). This is the general "boolean if/else-to-`&&`/`||` short-circuit expression" raise, specialized here to a declaration-pattern left operand.
2. **`default` literal inlining.** `V_2 = default; ... subject.CompareTo(V_2)` should inline to `subject.CompareTo(default)` (single-assignment temp holding a literal, read once).
3. **Fluent chain / temp re-composition.** The `S_256`/`S_257`/`S_258` receiver and params-array temps should collapse back to the fluent `.ForCondition(...).BecauseOf(...).FailWith(..., Subject)` chain.

Reaching all three yields the authored shape. Tracking here so #2873 can record the intended endpoint without overclaiming "fully raised".

PR #2903, opened 2026-07-19
Raise pattern-guarded && diamond, inlining spilled default(T) (#2877)

richlander:
- Fixes/advances #2877
- Changes the pattern-guarded `&&` diamond (with a spilled `default(T)` temp) to decompile back into the `&&` short-circuit it was lowered from.
- Card revision: `e51b3d31`

## Change

**Conclusion:** **PASS** — a tightly-scoped pass recovers the FluentAssertions `BePositive` guard to its one-line source with zero corpus validity movement.

csc lowers `target = cond && rhs` to `if (cond) target = rhs; else target = false;`. When the condition is an `is`-pattern, the pre-pattern diamond folders cannot run (the pattern only enters the condition after `IsPatternPass`) and — for a struct type parameter — the true arm still opens with the compiler's spilled `default(T)` temp. So the guard surfaced as a five-statement if/else diamond instead of its one-line source.

### Original source

```csharp
{
    CurrentAssertionChain

}
```

### Before

```csharp
T V_2;
bool S_1;

AssertionChain S_256 = CurrentAssertionChain;
if (Subject is T subject)
{
    V_2 = default;
}
else
{
    S_1 = false;
}
object[] S_257 = new object[1];
AssertionChain S_258 = S_256.ForCondition(S_1).BecauseOf(because, becauseArgs);
S_257[0] = Subject;
S_258.FailWith("Expected {context:value} to be positive{reason}, but found {0}.", S_257);
return new AndConstraint<TAssertions>((TAssertions)(object)this);
```

### After

```csharp
AssertionChain S_256 = CurrentAssertionChain;
bool S_1 = Subject is T subject && subject.CompareTo(default(T)) > 0;
object[] S_257 = new object[1];
AssertionChain S_258 = S_256.ForCondition(S_1).BecauseOf(because, becauseArgs);
S_257[0] = Subject;
S_258.FailWith("Expected {context:value} to be positive{reason}, but found {0}.", S_257);
return new AndConstraint<TAssertions>((TAssertions)(object)this);
```

### Fully raised

```csharp
CurrentAssertionChain
    .ForCondition(Subject is T subject && subject.CompareTo(default(T)) > 0)
```

The `&&`-diamond fold and `default(T)` inline (this PR's two concerns) are complete. The remaining residue is the `S_1` boolean materialized as a separate local rather than inlined into `ForCondition(...)` — a fluent-chain / single-use-slot inline concern, not a diamond or default concern.

- Required tracking issue: #2877 — remaining slice: inline the single-use `S_1` boolean into the `ForCondition(...)` argument (residue #3).

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass |
| Focused tests | `PatternGuardedShortCircuitPassTests` passed (5) |
| Decompiler fast suite | Pass (2599/0) |
| Reduced fixture validity | Pass — `default(T)` always compiles |
| Reduced fixture fidelity | Not applicable (raise-only, no opcode change) |
| Real witness | `NumericAssertionsBase\`3::BePositive` folded |

## Decompiler quality

**Conclusion:** **PASS** — on FluentAssertions 8.10.0 the change moves exactly 7 methods (all `NumericAssertionsBase` pattern-guarded comparison guards) and every one stays valid; no method regresses.

### FluentAssertions render A/B (4676 methods)

| Metric | Result |
| --- | --- |
| Changed | 7 (structural: 7, paren-equivalent: 0, unparsed: 0) |
| Semantic valid->valid | 7 |
| Semantic valid->invalid | 0 |
| Semantic invalid->valid | 0 |

Changed methods: `Be`, `BeGreaterThan`, `BeGreaterThanOrEqualTo`, `BeOneOf`, `BePositive`, and the remaining sibling comparison guards — the intended blast radius.

### FluentAssertions validity defect diff (2154 methods checked in both)

| Direction | Count |
| --- | --- |
| Regressed (gained a defect code) | 0 |
| Improved (lost a defect code) | 0 |

Pre-existing defects (e.g. one `CS0413`) are identical in base and head.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class ILInspector.Decompiler.Tests.PatternGuardedShortCircuitPassTests
# corpus (isolated, no concurrent Roslyn):
dotnet tools/DecompilerHarness/bin/Release/net11.0/decompiler-harness.dll --package FluentAssertions --render-ab /tmp/fa-base-render.json
dotnet tools/DecompilerHarness/bin/Release/net11.0/decompiler-harness.dll --package FluentAssertions --validity-check --diff-validity-defects /tmp/fa-base-defects.json
```


richlander:

Two-model adversarial review (GPT-5.5, Gemini 3.1 Pro) across three rounds, each isolated in a separate worktree at an exact head. All findings below were reproduced on a clean exact-head checkout before fixing, and every fix was re-reviewed on the fixed head.

### Round 1 — head `aab21a13`
| # | Reviewer | Finding | Verdict | Resolution |
| - | -------- | ------- | ------- | ---------- |
| 1 | GPT-5.5 | Default-temp matched by local index across the whole function including nested lambda/local-function bodies (separate index space) → same-index inner local mis-recovered as `default(T)` (CS0029) | **Real**, reproduced | Fixed in `e161a8ca` (later superseded by the round-2 refuse-rule) |
| 2 | Gemini | Pattern nested under `||`/`!` is confined but not definitely assigned → folding to `&&` emits CS0165 | **Real**, reproduced | Fixed in `e161a8ca` (spine guard; later generalized to an allowlist in round 2) |
| 3 | Gemini | Duplicate `initobj` same index → double `ReplaceWith` on a detached node → crash | **Not reproducible** — the existing `LoadLocalAddress` uniqueness scan rejects the second init's distinct address before any rewrite. Both models confirmed on re-test. | Documented by negative test `DuplicateInitObjSameIndex_MustNotCrash` |

### Round 2 — head `e161a8ca`
| # | Reviewer | Finding | Verdict | Resolution |
| - | -------- | ------- | ------- | ---------- |
| 4 | GPT-5.5 | A lambda referencing the outer default-temp index (a capture) is invisible to the nested-body-excluding scan → outer temp recovered, lambda reference left dangling (CS0103) | Both models noted captured outer locals lower to closure **fields**, not `LoadLocal`, so the exact shape is likely **unreachable** from real IL — fixed defensively regardless | Fixed in `4d25b2b4`: `TryRecoverArmDefaults` refuses recovery if the temp index is referenced/bound in **any** nested-function body |
| 5 | GPT-5.5 | A local bound by `RecursivePropertyDeclarationPattern` (not `IsPattern`) bypassed the spine + confinement gates → escaping bound local folds → CS0165 | **Real**, reproduced | Fixed in `4d25b2b4`: `ConditionBoundLocals` collects all condition binders (is + recursive property declaration patterns); each is spine- and confinement-gated |
| 6 | Gemini | Spine walk was a **blocklist** (rejected only `||`/`!`); `??` and `?:` also break definite assignment → CS0165 | **Real**, reproduced | Fixed in `4d25b2b4`: replaced with **allowlist** `BinderRequiredForConditionTrue` — folds only when every ancestor up to the condition root is `&&` |

### Round 3 — fixed head `4d25b2b4`
Both reviewers re-reviewed the exact fixed head and reproduced their own findings as resolved.

- **GPT-5.5: CLEAN** — original findings resolved (`IfStatements=1, DefaultValues=0`), no new blocking issues in the nested-body refuse-check, `ConditionBoundLocals`, or the allowlist walk.
- **Gemini 3.1 Pro: CLEAN** — allowlist correctly rejects `Conditional`/`Call`/`Coalesce`/`Comparison`/`Binary`/tuple ancestors; `ConditionBoundLocals` complete (verified `out var`, tuple/list patterns, `??=` introduce no additional condition-binding nodes); nested-function refuse-check sound.

Both confirmed `BePositive` still folds to `bool S_1 = Subject is T subject && subject.CompareTo(default(T)) > 0;`.

### Evidence (head `4d25b2b4`)
- Fast decompiler suite: 2606 / 0.
- FluentAssertions render A/B vs base: 4676 methods, 7 changed (all `NumericAssertionsBase` guards), valid→valid 7, valid→invalid 0.
- Validity defect diff vs base: 0 regressed / 0 improved over 2154 methods.
- Regression tests added: `PatternUnderOr_MustNotFold`, `PatternUnderCoalesce_MustNotFold`, `RecursivePropertyPatternEscape_MustNotFold`, `NestedLambdaSameIndex_MustNotMisRecover`, `CapturedOuterTempInLambda_MustNotRecover`, `DuplicateInitObjSameIndex_MustNotCrash`.

richlander:
Ready to merge

PR #2905, opened 2026-07-19
Add real-corpus authored-source census (--authored-source-census)

richlander:
## Summary

Closes the one remaining gap called out in #2673's "frontiers" table
(introduced by #2717): **"Assembly-wide source-only census: Not added."**

`--source-correspondence-census` (Slice 2's `valid_match`/`valid_different`/
`invalid`/`source_unavailable`/`unsupported_target` bucket classifier) selects
**zero** targets for any assembly that is not registered in `FixtureCatalog`
(`ReturnToSenderSourceProbe.cs`'s target discovery is hard-scoped to the
fixture catalog). There was no way to run that bucket taxonomy over real,
non-fixture corpus assemblies.

This PR adds `--authored-source-census`, which runs the same bucket taxonomy
over real corpus assemblies using real SourceLink acquisition, by wiring
together two already-shipped, independently tested pieces rather than adding a
new oracle:

- `AuthoredRebuildFidelity`'s real acquisition lane (`SourceLinkService`,
  checksum-verified `SourceFetcher`, `AuthoredSourceAcquisition`,
  `TryExtractTargetBody`) to get the exact checksum-verified authored body for
  a real RTS target.
- `ReturnToSenderSourceProbe`'s existing comparison/classification/report logic
  (`NormalizeBody`, `ClassifyValidDifference`, and a newly extracted
  `Report(results, maxExamples, json)`) to bucket the authored body against
  the decompiled body using the identical taxonomy as the fixture-based probe.

This keeps the RTS compile-back oracle, the authored-rebuild-fidelity oracle,
and this new source-correspondence lane strictly peer, per the #2673 reviewer
guidance to never blend oracle verdicts: each reports its own
`Absent`/`Failed`/`SourceUnavailable` outcomes independently.

## Changes

- `tools/DecompilerHarness/AuthoredSourceCensus.cs` (new): the census runner.
  `ClassifyExpectedBody` is a pure, synchronous classification step kept
  separate from acquisition so it can be tested against a real RTS result
  without any network access.
- `tools/DecompilerHarness/ReturnToSenderSourceProbe.cs`: widened
  `NormalizeBody`, `FailureReason`, `OpcodeEvidence`/`OpcodeDiffEvidence` from
  private/file-scoped to `internal` for cross-file reuse; extracted
  `Report(results, maxExamples, json)` out of `Run` so both the fixture lane
  and the new real-corpus lane share one report/JSON shape.
- `tools/DecompilerHarness/AuthoredRebuildFidelity.cs`: widened
  `AcquirePdbAsync` to `internal` for reuse.
- `tools/DecompilerHarness/Program.cs`: new `--authored-source-census` CLI
  flag and help text.
- `tools/DecompilerHarness/README.md`: documents the new flag alongside
  `--source-correspondence-census` and `--authored-rebuild-fidelity`.
- `src/ILInspector.Decompiler.Tests/AuthoredSourceCensusTests.cs` (new): 5
  focused tests.

## Evidence

- `dotnet build tools\DecompilerHarness -c Release`: succeeded.
- `dotnet build src\ILInspector.Decompiler.Tests -c Release`: succeeded.
- `dotnet run --project src\ILInspector.Decompiler.Tests -c Release --no-build -- -class "ILInspector.DecompilerHarness.AuthoredSourceCensusTests"`:
  **5/5 passed.** Covers `ClassifyExpectedBody`'s `valid_match`/
  `valid_different` branches against a real RTS result, plus `ClassifyAsync`'s
  `UnsupportedTarget`/`Invalid` early-exit branches — all without network
  access.
- Live CLI smoke test:
  `decompiler-harness --authored-source-census --package System.Text.Json --cap 15 --max-examples 5`
  resolves the real package, discovers real RTS targets, and reaches
  classification end-to-end.

  **Caveat:** on this Windows dev machine, every row bucketed as
  `invalid`/`CS0009` because of a pre-existing, unrelated environment issue
  (RTS's reference-assembly scan picks up
  `aspnetcorev2_inprocess.dll`, a native non-managed DLL, from the shared
  framework directory — the same issue documented in #2876). This blocks
  observing `ValidMatch`/`ValidDifferent` rows in a live local run on this
  machine, but does not affect the new wiring itself, which is exercised by
  the unit tests above and reuses acquisition/classification code paths that
  are already covered by existing `AuthoredRebuildFidelityTests` and
  `ReturnToSenderSourceProbe`-based tests. This is expected to run cleanly on
  Linux/macOS CI.

## Review

This is a new, non-trivial feature (new corpus-wide acquisition + reporting
mode), so it requires two-model adversarial review per `AGENTS.md` before
merge. Requesting review.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/AuthoredSourceCensusTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AuthoredSourceCensusTests.cs
@@ -137,3 +137,48 @@ public sealed class ReturnToSenderSourceProbeQualityCardTests
         Assert.Contains("no checkable rows", card);
     }
 }
+
+public sealed class AuthoredSourceCensusDiversificationTests
+{
+    [Fact]
+    public void DiversifyByKey_RoundRobinsAcrossKeys_WhenPoolExceedsCap()
+    {
+        string[] candidates = ["A1", "A2", "A3", "A4", "A5", "A6", "A7", "A8", "A9", "A10", "B1", "C1"];
+
+        var selected = AuthoredSourceCensus.DiversifyByKey(candidates, 4, item => item[..1]);
+
+        Assert.Equal(["A1", "B1", "C1", "A2"], selected);
+    }
+
+    [Fact]
+    public void DiversifyByKey_ReturnsAllCandidates_WhenPoolFitsWithinCap()
+    {
+        string[] candidates = ["A1", "B1", "C1"];
+
+        var selected = AuthoredSourceCensus.DiversifyByKey(candidates, 10, item => item[..1]);
+
+        Assert.Same(candidates, selected);
+    }
+
+    [Fact]
+    public void DiversifyByKey_ReturnsEmpty_WhenCapIsNotPositive()
+    {
+        string[] candidates = ["A1", "B1"];
+
+        var selected = AuthoredSourceCensus.DiversifyByKey(candidates, 0, item => item[..1]);
+
+        Assert.Empty(selected);
+    }
+
+    [Fact]
+    public void DiversifyByKey_StopsAtCap_WhenOneKeyDominatesThePool()
+    {
+        // Mirrors a real-world corpus where a generated resource-string type
+        // (e.g. SR) contributes far more property getters than any other type.
+        string[] candidates = Enumerable.Range(0, 50).Select(i => $"SR{i}").Append("Other1").ToArray();
+
+        var selected = AuthoredSourceCensus.DiversifyByKey(candidates, 3, item => item.StartsWith("SR", StringComparison.Ordinal) ? "SR" : "Other");
+
+        Assert.Equal(["SR0", "Other1", "SR1"], selected);
+    }
+}
--- a/src/ILInspector.Decompiler.Tests/AuthoredSourceCensusTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AuthoredSourceCensusTests.cs
@@ -87,3 +87,53 @@ public sealed class AuthoredSourceCensusTests
         Assert.True(result.Failed);
     }
 }
+
+public sealed class ReturnToSenderSourceProbeQualityCardTests
+{
+    static ReturnToSender.RequestedTarget SyntheticTarget(string method)
+        => new("Synthetic.Type", method, 0);
+
+    static ReturnToSenderSourceProbeResult Synthetic(ReturnToSenderSourceOutcome outcome, string reason, string method = "M")
+        => new(SyntheticTarget(method), outcome, null, reason, null, null, null, null);
+
+    [Fact]
+    public void RenderQualityCard_AccountsForEveryResult_AcrossAllBuckets()
+    {
+        var results = new[]
+        {
+            Synthetic(ReturnToSenderSourceOutcome.ValidMatch, "valid_match", "A"),
+            Synthetic(ReturnToSenderSourceOutcome.ValidDifferent, "valid_different.known_taste", "B"),
+            Synthetic(ReturnToSenderSourceOutcome.ValidDifferent, "valid_different.semantic_opcode_diff", "C"),
+            Synthetic(ReturnToSenderSourceOutcome.Invalid, "invalid.recompile_fail", "D"),
+            Synthetic(ReturnToSenderSourceOutcome.SourceUnavailable, "source_unavailable", "E"),
+            Synthetic(ReturnToSenderSourceOutcome.UnsupportedTarget, "unsupported-rts-target", "F"),
+        };
+
+        string card = ReturnToSenderSourceProbe.RenderQualityCard(results, "2 test assemblies");
+
+        Assert.Contains("Source-correspondence quality card", card);
+        Assert.Contains("2 test assemblies", card);
+        Assert.Contains("6 target(s) sampled", card);
+        Assert.Contains("Valid match (+) | 1 (16.67%)", card);
+        Assert.Contains("ignorable known difference (+) | 1 (16.67%)", card);
+        Assert.Contains("semantic opcode diff (-) | 1 (16.67%)", card);
+        Assert.Contains("Invalid / RTS compile-back failed (-) | 1 (16.67%)", card);
+        Assert.Contains("Source unavailable (uncheckable) | 1 (16.67%)", card);
+        Assert.Contains("Unsupported target (uncheckable) | 1 (16.67%)", card);
+        Assert.Contains("Verdict:", card);
+    }
+
+    [Fact]
+    public void RenderQualityCard_ReportsNoSignalVerdict_WhenNothingIsCheckable()
+    {
+        var results = new[]
+        {
+            Synthetic(ReturnToSenderSourceOutcome.SourceUnavailable, "source_unavailable", "A"),
+            Synthetic(ReturnToSenderSourceOutcome.UnsupportedTarget, "unsupported-rts-target", "B"),
+        };
+
+        string card = ReturnToSenderSourceProbe.RenderQualityCard(results, "no-signal corpus");
+
+        Assert.Contains("no checkable rows", card);
+    }
+}
--- a/src/ILInspector.Decompiler.Tests/AuthoredSourceCensusTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AuthoredSourceCensusTests.cs
@@ -1,0 +1,89 @@
+using DotnetInspector.Fixtures;
+using DotnetInspector.Services;
+
+namespace ILInspector.DecompilerHarness;
+
+public sealed class AuthoredSourceCensusTests
+{
+    static ReturnToSender.Result RealDecompilerResult()
+        => ReturnToSender.CompileBackFirstPropertyGetter(FixtureCatalog.DiffPair.OldAssemblyPath());
+
+    static ReturnToSender.RequestedTarget TargetFor(ReturnToSender.Result result)
+    {
+        var identity = result.Plan.TargetMethod;
+        return new ReturnToSender.RequestedTarget(
+            identity.Type,
+            identity.Method,
+            identity.Overload,
+            identity.Signature);
+    }
+
+    [Fact]
+    [Trait("Speed", "Slow")]
+    public void ClassifyExpectedBody_ReportsValidMatch_WhenAuthoredBodyIsWhitespaceEquivalent()
+    {
+        var decompiler = RealDecompilerResult();
+        var target = TargetFor(decompiler);
+        string expected = $"  {decompiler.TargetBody}  \n";
+
+        var result = AuthoredSourceCensus.ClassifyExpectedBody(target, decompiler, "https://example.test/Source.cs", expected);
+
+        Assert.Equal(ReturnToSenderSourceOutcome.ValidMatch, result.Outcome);
+        Assert.Equal("valid_match", result.Reason);
+        Assert.True(result.Passed);
+        Assert.Equal(decompiler.TargetBody, result.ActualBody);
+    }
+
+    [Fact]
+    [Trait("Speed", "Slow")]
+    public void ClassifyExpectedBody_ReportsValidDifferent_WhenAuthoredBodyDiffersInSubstance()
+    {
+        var decompiler = RealDecompilerResult();
+        var target = TargetFor(decompiler);
+        string expected = decompiler.TargetBody + " /* an authored-only comment marker that is not whitespace */ return 12345;";
+
+        var result = AuthoredSourceCensus.ClassifyExpectedBody(target, decompiler, "https://example.test/Source.cs", expected);
+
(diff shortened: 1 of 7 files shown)


richlander:
**Decompiler quality diff card: not applicable.**

--quality-diff-card tracks decompiler IR/opcode raising quality (Fully raised %, conditional-branch residual, forward-merge stops, malformed, semantic defects, fidelity diffs) against a pinned corpus baseline. This PR does not touch that path:

- No changes to raising/structuring/validity/fidelity code in ILInspector.Decompiler or ILInspector.CSharp.
- The two existing files touched (ReturnToSenderSourceProbe.cs, AuthoredRebuildFidelity.cs) only had private/file-scoped members widened to internal for reuse -- zero logic changed, so decompiled output is byte-for-byte identical to before.
- AuthoredSourceCensus is a comparison harness: it decompiles exactly as before, then compares the (unchanged) decompiled body text against real SourceLink-fetched authored text. Orthogonal to the raising pipeline the card measures.

Since this PR provably cannot move any decompiler-quality metric (same decompiler code path, same corpus), a quality-diff-card run here would only reproduce the existing baseline with Count delta: 0 everywhere -- so it's skipped per the no-op-card guidance in docs/decompiler-quality.md.

richlander:
Added a source-correspondence quality card in this PR (new commit), per discussion: this is a new PR-pasteable Markdown bucket-rate card, distinct from the existing decompiler `--quality-diff-card` (which measures IR/opcode raising quality against a pinned baseline and doesn't apply here since this feature changes zero decompiler-path behavior).

New flag: `--source-quality-card`. Works with both `--source-correspondence-census` and `--authored-source-census` in text mode (no effect with `--json`, since the card is prose). `ReturnToSenderSourceProbe.RenderQualityCard` buckets every sampled result into:

- Valid match
- Valid different, sub-bucketed via the existing `SourceCorrespondenceCategory` taxonomy (ignorable known difference, not-yet-raised sugar, structuring residue, semantic opcode diff, unclassified)
- Invalid / RTS compile-back failed
- Source unavailable / unsupported target (uncheckable)

...with counts, rates, and a descriptive (non-gating) verdict line. This is a **current-state snapshot**, not a diff against a baseline — there's no established baseline population yet for this oracle, and per our discussion, RTS-vs-real-authored-source quality over an arbitrary real corpus is not expected to be perfect yet. The card exists so reviewers can see the actual distribution rather than only a pass/fail count.

Added 2 focused, network-free unit tests (`ReturnToSenderSourceProbeQualityCardTests`) against the pure rendering logic. Build + focused tests green (7/7 in `AuthoredSourceCensusTests.cs`).

Same local-env caveat as the base feature applies: running this locally on this machine currently shows 100% invalid due to the pre-existing, unrelated `aspnetcorev2_inprocess.dll` / CS0009 RTS reference-scan bug (reproduces on unmodified main). Expected to run cleanly on Linux/macOS CI.

richlander:
**Current-state source-correspondence snapshot**

Reviewed revision: `f88eefe45204145db265614d3a487cec8e66fe55`

Pinned run:
```bash
  --package System.Text.Json --package-version 8.0.0 \
  --package-tfm net8.0 --package-assembly System.Text.Json.dll \
  --authored-source-census --source-quality-card \
  --cap 100 --max-examples 5
```

### Source-correspondence quality card

Corpus: 1 real assembly/assemblies (SourceLink-acquired authored source). 100 target(s) sampled.

| Bucket (desired direction) | Count (rate) |
| --- | ---: |
| Valid match (+) | 0 (0.00%) |
| Valid different - ignorable known difference (+) | 0 (0.00%) |
| Valid different - not-yet-raised sugar (~) | 1 (1.00%) |
| Valid different - structuring residue (~) | 0 (0.00%) |
| Valid different - semantic opcode diff (-) | 0 (0.00%) |
| Valid different - unclassified (-) | 0 (0.00%) |
| Invalid / RTS compile-back failed (-) | 0 (0.00%) |
| Source unavailable (uncheckable) | 99 (99.00%) |
| Unsupported target (uncheckable) | 0 (0.00%) |

Verdict: 0/1 checkable row(s) (0.00%) matched authored source exactly; this is a current-state snapshot, not yet compared against a floor.
**Interpretation:** The sole checkable row was `System.HexConverter.get_CharToHexLookup`, classified as `valid_different.source_shape_frontier.syntax.exact` with exact compile-back opcodes. The other 99 rows were generated `System.SR` properties whose SourceLink artifact URL is unavailable, so this sample primarily exposes a source-availability frontier rather than broad correspondence quality.

**Adversarial review:** Claude Opus 4.8 and Gemini 3.1 Pro independently reviewed exact head `f88eefe`; both reported no blocking or important findings. Both ran the focused tests and Linux real-package text/JSON smoke paths.
**CI:** Green. **Mergeability:** Clean.

Ready to merge

richlander:
**Pushing back on the ""Ready to merge"" verdict above** — the posted quality-card run exposed a real defect, not just a low initial-quality result.

99/100 sampled targets in that run were SourceUnavailable, and (per the interpretation note) all 99 were generated System.SR resource-string properties — a single declaring type. ReturnToSender.CompileBackPropertyGetters enumerates property getters in metadata token order with no diversity guard, so whichever type happens to have the most property getters (here, an SR class with hundreds of entries) can crowd out the entire --cap-bounded sample. That's a sampling defect in the new census, not a real signal about RTS-vs-authored-source quality — the card was reporting on one degenerate type, not the corpus.

Pushed a fix (fdb8258): AuthoredSourceCensus now pulls a candidate pool up to 5x --cap per assembly, then round-robins the selection across distinct declaring types (DiversifyByKey) down to the actual budget, so no single type can dominate the sample. This is scoped to the new census only — ReturnToSender.CompileBackPropertyGetters itself, and the other RTS modes/pinned baselines that depend on its existing enumeration order, are untouched.

Added 4 focused, network-free unit tests against the pure round-robin/cap logic (AuthoredSourceCensusDiversificationTests), including a synthetic case mirroring the SR-domination scenario. Build + full focused test pass (11/11) green.

This should not be marked ready to merge until it's re-run against a real package and re-reviewed on this fixed head — the prior card run is no longer representative of current behavior.

richlander:
Added a generate-once / locked-roster replay mode, per discussion: the live-discovery model re-samples a fresh population on every run (pool size scales with --cap), so --cap 100 and --cap 1000 runs — or the same --cap run again later — aren't comparable to each other. That makes ""generate every time"" useless as a benchmark, only a one-off snapshot.

New two-phase workflow:

- **--authored-source-census-generate-roster <path>** — one-time generation phase. Samples a diversified candidate pool, classifies each with real RTS compile-back + real SourceLink acquisition, and locks in every member whose authored source was actually available (regardless of RTS pass/fail — that verdict is what replay runs re-measure) to a roster JSON file.
- **--authored-source-census-roster <path>** — replays the locked roster: takes the first --cap members (a nested prefix, same order as generation) and classifies them fresh via ReturnToSender.CompileBackTargets. Because the prefix only grows with --cap, a --cap 100 and a --cap 1000 run against the same roster — now or months later — are directly comparable (strict subset/superset), unlike live discovery.

Example: generate once with --cap 10000, then run repeatedly at --cap 100 for fast PR checks and --cap 1000 for deeper CI runs, all against the same locked population.

Implementation notes:
- Both flags imply --authored-source-census and are mutually exclusive with each other.
- Roster JSON uses the same serialization convention as ReturnAddressCensus's snapshot/baseline files (camelCase, indented).
- DiversifyByKey (fixed-size selection, used by live discovery) now delegates to a new RoundRobinByKey (full-pool reorder, no truncation, used by roster generation) — existing diversification tests/behavior unchanged.
- Roster assemblies correlate to resolved paths by file name at replay time; unmatched entries warn to stderr rather than failing hard.

Added 3 more focused, network-free unit tests (full-pool reorder + roster JSON round-trip). Full suite: 13/13 passing. CLI-smoke-tested --help output and the new mutually-exclusive-flag validation.

richlander:
Implemented the exact allocation algorithm requested (commit `deb5e71e`):

- Collect every eligible candidate per corpus library (full compile-back enumeration, not a `--cap`-sized sample).
- Sort libraries ascending by candidate count.
- `n = total candidates / library count` (fair share).
- Every library except the largest gets `min(n, its own count)`.
- The single largest library absorbs whatever is left over (bounded by its own count) to hit `--cap` exactly.
- Per-library allocation is still diversified by declaring type before classification, same as before.

This replaces the prior first-come/remaining-budget model, where a package listed first (or one huge library) could consume the whole `--cap` and starve everything after it.

**Real verification** — `--authored-source-census-generate-roster` with `--cap 60` against `Humanizer.Core` (small, 3.0.10) + `System.Text.Json` (huge, 10.0.10):

```text
Wrote authored-source census roster with 60 locked member(s) across 2 assembly/assemblies (candidate pool: 635 across 2 libraries, cap=60).
```

Per-library breakdown from the roster JSON:

- `Humanizer.dll`: 55 members
- `System.Text.Json.dll`: 5 members

Humanizer's raw candidate pool is small (well under the 317-per-library fair share for this corpus), so it kept effectively all of its own candidates; System.Text.Json — the dominant library — only absorbed the remaining ~5–10 slots needed to reach the cap, instead of taking all 60 and leaving Humanizer with nothing. This is exactly the "don't let one huge library crowd out the rest of the corpus" property this change was for.

Added `AllocateAcrossLibraries` as a pure, network-free function with 5 new unit tests (trivial fits-within-cap, shortfall-absorption, dominant-library-does-not-crowd-out, corpus-smaller-than-cap, single-library). Full focused suite: 18/18 passing.

README updated to describe the new allocation algorithm for `--authored-source-census-generate-roster`.

richlander:
Fixed a real design gap you flagged: the roster corpus was previously locked *after* filtering by classification outcome (excluding `SourceUnavailable`/`UnsupportedTarget`), which made the fixed corpus a product of today's SourceLink resolution rate — partial (a member SourceLink fails on today could never enter the roster, even after a later fix) and circular (can't tell "oracle got better" apart from "corpus changed").

**Fixed in `c495f96a`**: roster generation now locks every selected candidate by identity alone (`Type`/`Method`/`Overload`/`Signature`) — no SourceLink fetch, no classification, no outcome filter at generation time. `GenerateRoster` is now fully synchronous and no longer needs `HttpClientFactory`/`SourceFetcher`/PDB acquisition at all.

`RunFromRoster` already classified fresh on every replay, so both axes are now measured independently against the same fixed population on every run:
- **SourceLink's own resolution rate** — the "Source unavailable"/"Unsupported target" rows in the quality card.
- **RTS's match rate** among whatever currently resolves — "Valid match"/"Valid different"/"Invalid".

**Real verification** — replayed the two-package roster (Humanizer.Core + System.Text.Json, `--cap 60`) generated under the new model:

```text
Source unavailable (uncheckable) | 0 (0.00%)
Unsupported target (uncheckable) | 0 (0.00%)
Invalid / RTS compile-back failed (-) | 60 (100.00%)
```

SourceLink resolved 100% of the locked members (0% unavailable/unsupported) while RTS failed on all of them (known local CS0009 environment bug on this machine) — exactly the intended decoupling: the oracle's resolution rate and RTS's match rate are now separate, independently-trackable signals over one truly fixed corpus, instead of one gating the other's membership.

18/18 focused tests still pass (JSON round-trip and allocation tests are unaffected by this change).

richlander:
Closing in favor of a cleaner redesign. This branch conflated two different jobs into one "roster" artifact, which is where the confusion and disagreement came from — and it collided with the compile-back **fidelity contract V1** rename/redefinition from #2852 (new `OperandDiff` status + `FidelityDiff`), which the census's `Exact`/`OpcodeDiff`-only checkable gate and its `OpcodeEvidence` references no longer fit. It also never produced a real `valid_match`/`valid_different` distribution end-to-end (every live row bucketed `invalid/CS0009` on the Windows env issue, #2876), so the core value was unproven.

### Replacement design: two distinct modes

**Fixed benchmark** — measures decompiler-vs-authored-source correspondence. SourceLink availability is a *precondition*, not a variable: the corpus is pre-filtered to source-available members, so there is **never** a `source_unavailable` bucket. Harvested authored bodies are **snapshotted at generation time** (identity + checksum-verified extracted member body) so the benchmark runs **offline, fast, and reproducibly**, immune to upstream source disappearing. Bias sampling to **real method bodies** (hard IL), diversified by declaring type — not property getters, which inflate the match rate. Sizes: **10k (fast)** and **100k (deep)**, ideally **250k**. Harvested source is **vendored in an orphan branch** (like `vendor/ilassembler`) to keep it out of main history and clear of licensing concerns.

**On-demand** — the RTS "original source" oracle complement, where availability *is* variable. Records a per-library SourceLink-coverage manifest so we skip libraries with no source instead of wasting fetch time.

### Salvage
Reuse the proven primitives from this branch — the SourceLink acquisition lane, the `NormalizeBody`/`ClassifyValidDifference` bucket taxonomy, and type-diversified sampling — but drop the identity-locked roster and the single conflated mode.

Successor work will start fresh on `main` post-#2852.

PR #2908, opened 2026-07-19
Render typed explicit-interface events in C# and RTS

### LINKED ISSUE #2850: Render structured explicit-interface events in CSharpDeclarationWriter
`CSharpDeclarationWriter.TryRenderSignatureModel` has no explicit-interface event branch. A structured event with a dotted interface-qualified name and no compatibility `Signature` falls through to the return-type fallback, losing the event name.

PR #2845 exposed the gap by including events in its dotted-name explicit-member discriminator even though that PR is scoped to properties. Keep events out of that heuristic until a typed explicit-event shape can render the interface qualifier, event keyword/name, and required modifiers with compiled fixture coverage.

richlander:
## Summary

Render explicit-interface events from a typed `add`/`remove` accessor contract,
with `CSharpTypePrinter` owning the final C# syntax and RTS owning only
event/accessor reconstruction shape.

- add `CSharpEventBody` and structured explicit-event declaration/body rendering
- classify explicit properties and events by their typed accessor sets rather
  than dotted display names
- make explicit-event skeletons fail closed because C# requires custom accessor
  bodies
- teach RTS to discover explicit event accessors, reconstruct the same-assembly
  interface event declaration, and provide the target accessor plus an honest
  sibling stub

## Original source

```csharp
event Action IBaseEvents.Changed
{
    add
    {
        Console.WriteLine(value);
    }
    remove
    {
        Console.WriteLine(value);
    }
}
```

## Before

RTS treated the selected explicit event accessor as an ordinary method. For the
`IBaseEvents.add_Changed` target, it produced:

```csharp
public void IBaseEvents_add_Changed(Action value)
{
    Console.WriteLine(value);
}
```

## After

RTS now reconstructs the typed explicit-interface event. For the same add
accessor target, it produces:

```csharp
event Action IBaseEvents.Changed
{
    add
    {
        Console.WriteLine(value);
    }
    remove
    {
        throw null;
    }
}
```

The remove-accessor target produces the symmetric shape, with the decompiled
body under `remove` and the honest sibling stub under `add`.

## Fully raised

```csharp
event Action IBaseEvents.Changed
{
    add
    {
        Console.WriteLine(value);
    }
    remove
    {
        Console.WriteLine(value);
    }
}
```

- Required tracking issue: #2913 — recover and render both accessor bodies in
  one RTS reconstruction.

## Compatibility

Ordinary field-like event skeletons keep their existing behavior. The RTS
projection follows the existing explicit-property boundary: the interface
declaration is reconstructed from same-assembly `MethodImpl` evidence;
generic/external interface expansion is not added here.

## Validation

- Release solution build
- ILInspector.CSharp.Tests: 276 passed
- ReturnToSenderPrototypeTests: 127 passed
- fast ILInspector.Decompiler.Tests: 2,778 passed
- fixed-head adversarial review: Claude Opus 4.8 and Gemini 3.1 Pro CLEAN

Closes #2850

### DIFF
--- a/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
@@ -578,6 +578,56 @@ public sealed class CSharpDeclarationWriterTests
     }
 
     [Fact]
+    public void ExplicitEventDeclaration_RendersFromStructuredAccessorShape()
+    {
+        var type = new ApiType { Namespace = "Samples", Name = "Values", Kind = "class" };
+        var member = new ApiMember
+        {
+            Name = "Some.@event.IEvents.Changed",
+            Kind = "explicit-interface-implementation",
+            IsStatic = true,
+            IsUnsafe = true,
+            SignatureModel = new ApiSignature
+            {
+                ReturnType = "System.EventHandler",
+                MemberName = "Some.event.IEvents.Changed",
+                Accessors =
+                [
+                    new ApiAccessor { Kind = "add" },
+                    new ApiAccessor { Kind = "remove" }
+                ]
+            }
+        };
+
+        var declaration = CSharpDeclarationWriter.RenderMemberDeclaration(type, member);
+
+        Assert.Equal(
+            "static unsafe event System.EventHandler Some.@event.IEvents.Changed",
+            declaration);
+    }
+
+    [Fact]
+    public void ExplicitEventDeclaration_RequiresTypedAccessorShape()
+    {
+        var type = new ApiType { Namespace = "Samples", Name = "Values", Kind = "class" };
+        var member = new ApiMember
+        {
+            Name = "Some.IEvents.Changed",
+            Kind = "explicit-interface-implementation",
+            SignatureModel = new ApiSignature
+            {
+                ReturnType = "System.EventHandler",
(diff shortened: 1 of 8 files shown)


richlander:
Adversarial review reconciliation for exact fixed head `4babfe5a1b8cc12d880f736e45e2d0fa088f5995`:

- **Gemini 3.1 Pro:** initially found a blocking ordinary field-like event regression: RTS event dispatch emitted both a custom event and its same-named backing field, causing CS0102. Reproduced on the original exact head and fixed in `4babfe5a` by requiring typed interface `MethodImpl`/`EventDefinition` evidence before taking the explicit-event path. Fixed-head re-review: **CLEAN**.
- **Claude Opus 4.8:** original-head review was clean but independently reproduced the ordinary-event collision as a non-blocking unreachable observation. Direct `RequestedTarget` selection made it actionable; the same `4babfe5a` fix preserves the prior ordinary-method path. Fixed-head re-review covered field-like implicit implementations, instance/static/generic explicit events, and non-interface/cross-assembly MethodImpl boundaries: **CLEAN**.

Fixed-head evidence: compiler-produced event tests 4/4; full `ReturnToSenderPrototypeTests` 127/127; no compile-back floor for the explicit or ordinary regression canaries.

richlander:
Ready to merge

PR #2909, opened 2026-07-19
Classify tuple switch expression precedence

richlander:
- Follow-up to #2901
- Classifies `TupleSwitchExpression` with its switch-expression siblings instead
  of falling through to primary precedence
- Card revision: `4a4b8106a24f3e7cc6c166d20e88855248555c66`

## Change

**Conclusion:** **PASS** — nested tuple switch expressions now receive the
parentheses expected by every `CSharpPrecedence` consumer, while the direct
return form emitted by the current raise remains unchanged.

### Original source

```csharp
public static int M(int x, int y, int z)
{
    return z + (x, y) switch { (> 0, > 0) => 1, _ => 2 };
}
```

This case is built as synthetic IR (`MakeTupleSwitch` + `Binary`), so the above
is the source that IR represents rather than a compiled fixture.

### Before

```csharp
return z + (x, y) switch { (> 0, > 0) => 1, _ => 2 };
```

- Valid: True
- Correct: True

The switch expression binds tighter than `+`, so this pre-fix output already
parses as the parenthesized form and behaves identically. But
`TupleSwitchExpression` was classified at primary precedence (not with the other
switch expressions), so `CSharpPrecedence` consumers were not guaranteed to wrap
it as a sub-expression — an under-parenthesization / fidelity defect in nested
contexts.

### After

```csharp
return z + ((x, y) switch { (> 0, > 0) => 1, _ => 2 });
```

- Valid: True
- Correct: True

Same behavior; `TupleSwitchExpression` now carries `Precedence.Conditional` like
every other switch expression, so all `CSharpPrecedence` consumers parenthesize
it consistently.

### Fully raised

## Evidence

Reported Before → After so the card shows the change, not just the current pass
state. The nested-operand rows differ across the columns and are the proof this
change is exercised; the regression guards read the same in both columns by
design.

| Check | Before | After |
| --- | --- | --- |
| Nested-operand tests (`BinaryOperand_TupleSwitchExpressionChild_StaysParenthesized`, `CoalesceLeft_TupleSwitchExpressionChild_StaysParenthesized`) | fail against pre-fix classifier — child tuple switch emitted without parens (`z + (x, y) switch …`, `… switch … ?? fallback`) | pass — child parenthesized (`z + ((x, y) switch …)`, `((…) switch …) ?? fallback`) and compiles |
| Direct-return guard (`DirectReturn_TupleSwitchExpression_RendersWithoutOuterParens`) | passes | passes |
| Tuple switch pass tests | 16 passed | 16 passed |
| Dataflow tests | 7 passed | 7 passed |
| Ladder tests | 9 passed | 9 passed |
| Compile-back (binary, coalesce, direct-return) | passes | passes |
| Decompiler fast suite | 2,782 passed | 2,782 passed |

The product delta is one precedence classification plus focused printer and
compile-back tests. Direct tuple-switch returns intentionally remain unwrapped.

## Decompiler quality

**Conclusion:** **PASS** — the current tuple-switch raise emits only direct
returns, whose output is unchanged. The correction affects nested consumers of
the shared precedence contract and can only add grouping parentheses.


richlander:
## Adversarial review

Exact base: `e78695af7cde979cf11f74ebf7b3db76ce14bbf4`

Exact head: `4a4b8106a24f3e7cc6c166d20e88855248555c66`

- Claude Opus 4.8: **CLEAN**. Confirmed the classification can only add
  parentheses, direct-return output is unchanged, and targeted precedence,
  tuple, dataflow, and ladder tests pass.
- Gemini 3.1 Pro: **CLEAN**. Confirmed sibling-consistent conditional
  precedence, compile-back for nested binary/coalesce forms, and no regression
  to the direct-return raise.

No findings remain at the reviewed head.

richlander:
## Review follow-up

**Was the previous (unwrapped) output invalid C#?** No. Verified by compiling both directions directly:

```csharp
return z + (x, y) switch { (> 0, > 0) => 1, _ => 2 };               // compiles, evaluates correctly
return (x, y) switch { (> 0, > 0) => "one", _ => "other" } ?? fallback; // compiles, evaluates correctly
```

A switch expression is self-terminating at its closing `}`, so there's no real parse ambiguity — this PR's comments already say as much. This is an IR-contract consistency fix (`CSharpPrecedence.Of` now agrees with its siblings for future consumers), not a fix for previously-broken or previously-incorrect output; the pass never emits this shape as a nested operand today (confirmed by `DirectReturn_...`).

**Evidence table before/after.** Built base (`e78695af`, main) and head (`4a4b8106`) in separate worktrees and ran the fast suite on both:

| | Total | Failed |
| --- | ---: | ---: |
| Baseline (main) | 2778 | 53 |
| Head (this PR, before fix) | 2782 | 54 |

The extra failure was this PR's own new test, `PrinterPrecedenceTests.DirectReturn_TupleSwitchExpression_RendersWithoutOuterParens`, comparing printer output (`\r\n` on Windows) against a literal `\n` string — a Windows-only false negative, not a product regression. Pushed a fix (`.ReplaceLineEndings("\n")`, matching the convention used everywhere else in this test project) plus an update to `docs/templates/decompiler-pr.md` requiring Baseline/Head columns in the Evidence table so a single Head-only "n passed" count can't hide this kind of thing again.

Re-ran after the fix: 2782 total, 53 failed (matches baseline's pre-existing, environment-specific failures — none new).

PR #2902, opened 2026-07-19
Raise fixed-buffer element access

richlander:
- Fixes #2865
- Raises compiler-owned fixed-buffer backing access to source element indexing,
  including reads, writes, byrefs, pointers, and instance-call receivers
- Card revision: `ae46647b8de8858b6c58ba20e6c3875f7822e021`

## Change

> Should we accept this change?

**Conclusion:** **PASS** — fixed-buffer access now returns to the source field
and index only when metadata identity, storage type, scale, receiver, and
consumer shape are proven; close lookalikes retain the honest fallback.

### Original source

```csharp
public struct FixedBufferResiduals
{
    public fixed int Data[4];

    public int Sum()
    {
        int sum = 0;
        for (int i = 0; i < 4; i++)
        {
            unsafe
            {
                sum += Data[i];
            }
        }
        return sum;
    }
}
```

### Before

```csharp
public int Sum()
{
    int sum = 0;
    for (int i = 0; i < 4; i++)
    {
        sum += System.Runtime.CompilerServices.Unsafe.Add(
            ref Data.FixedElementField,
            i);
    }
    return sum;
}
```

### After

```csharp
public unsafe int Sum()
{
    int sum = 0;
    for (int i = 0; i < 4; i++)
    {
        unsafe
        {
            sum += Data[i];
        }
    }
    return sum;
}
```

### Fully raised

The After decompilation is in the fully raised state.

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass |
| Fixed-buffer pass tests | 130 passed |
| Ladder rung 6 gates | 33 passed |
| Return-to-Sender compile-back gate | 1 passed |
| Decompiler fast suite | 2,736 passed |
| Compiler fixtures | New and legacy unsafe rules pass |
| Close-negative declines | Lookalike metadata, mismatched storage/scale, escaped refs, malformed calls, and wrong receivers remain unraised |

## Decompiler quality

> Should the corpus signal block this PR?

**Conclusion:** **ADVISORY** — compiler-produced target and close-negative
shapes are covered locally; broader corpus signal remains with CI and the
existing #2818 corpus lane.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
  -class "ILInspector.Decompiler.Tests.FixedBufferElementAccessPassTests"
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
  -class "ILInspector.Decompiler.Tests.LadderRung6GateTests"
[...]

PR #2910, opened 2026-07-19
Raise nested fixed-buffer indices inside out

richlander:
- Follow-up to #2902
- Raises nested fixed-buffer indices inside out instead of leaving cloned inner
  pointer arithmetic behind
- Card revision: `c1114e761ec0c0e42776d62197cbdc6db40e1cff`

## Change

**Conclusion:** **PASS** — bottom-up traversal ensures an inner fixed-buffer
access is already raised before an outer access clones its index expression.

### Original source

```csharp
public void WriteAtNestedIndex()
{
    Data[Data[1]] = 1;
}
```

### Before

```csharp
public unsafe void WriteAtNestedIndex()
{
    Data[(ref Data.FixedElementField + 4)] = 1;
}
```

- Valid: False
- Correct: False

`(ref Data.FixedElementField + 4)` is not a legal expression or index — a `ref`
expression cannot be an operand of `+` or an array index. The inner `Data[1]` is
left as lowered fixed-buffer pointer-arithmetic residue, so it does not compile
and is not fully raised.

### After

```csharp
public void WriteAtNestedIndex()
{
    Data[Data[1]] = 1;
}
```

- Valid: True
- Correct: True

Binds as written and is identical to the original source.

### Fully raised

## Evidence

Reported Before → After so the card shows the change, not just the current pass
state. The two nested-index rows differ across the columns and are the proof
this change is exercised; the regression guards read the same in both columns by
design.

| Check | Before | After |
| --- | --- | --- |
| Nested-index focused tests (`CompilerNestedFixedBufferIndex...`, `...NestedZero...`) | fail against pre-fix raiser — outer clone keeps lowered `Data[(ref Data.FixedElementField + 4)]`, fidelity not Full | pass — inner access raised first, `Data[Data[1]] = 1;`, Full |
| Real witness (current + legacy compiler fixtures) | inner index left as pointer-arithmetic residue, not Full | both reach Full |
| Fixed-buffer pass tests | 130 passed (nested tests absent) | 134 passed |
| Decompiler fast suite | 2,782 passed | 2,782 passed |
| Close boundary (130 positive / decline cases) | green | green |

The original pass walked a top-down snapshot. Raising the outer access cloned
its still-lowered index into the live tree, while the snapshot later rewrote
only the detached original. Reversing that existing snapshot is sufficient:
children raise first, and the outer clone therefore contains the raised inner
access.

## Decompiler quality

**Conclusion:** **PASS** — this is a localized traversal-order correction with
compiler-produced current and legacy witnesses; no recognition predicate or
accepted shape is broadened.

## Validation

```bash
  -class "ILInspector.Decompiler.Tests.FixedBufferElementAccessPassTests"
  -trait- "Speed=Slow"
```

### DIFF
--- a/src/ILInspector.Decompiler.Fixtures.LegacyUnsafe/UnsafeFixtures.cs
+++ b/src/ILInspector.Decompiler.Fixtures.LegacyUnsafe/UnsafeFixtures.cs
@@ -26,6 +26,8 @@ public unsafe struct FixedBufferResiduals
 
     public void WriteAtNestedIndex() => Data[Data[1]] = 1;
 
+    public void WriteAtNestedZeroIndex() => Data[Data[0]] = 1;
+
     public int ReadAtThroughFixedAddress(int index)
     {
         fixed (int* p = &Data[index])
--- a/src/ILInspector.Decompiler.Fixtures.LegacyUnsafe/UnsafeFixtures.cs
+++ b/src/ILInspector.Decompiler.Fixtures.LegacyUnsafe/UnsafeFixtures.cs
@@ -24,6 +24,8 @@ public unsafe struct FixedBufferResiduals
 
     public void WriteFirst(int value) => Data[0] = value;
 
+    public void WriteAtNestedIndex() => Data[Data[1]] = 1;
+
     public int ReadAtThroughFixedAddress(int index)
     {
         fixed (int* p = &Data[index])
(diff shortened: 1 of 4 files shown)


richlander:
## Adversarial review

Exact base: `e78695af7cde979cf11f74ebf7b3db76ce14bbf4`

Exact head: `c1114e761ec0c0e42776d62197cbdc6db40e1cff`

- Claude Opus 4.8: **CLEAN**. Verified the bottom-up snapshot has no stale-node,
  duplicate-raise, side-effect-order, or existing-form regressions; 134 focused
  tests pass and both nested witnesses compile back.
- Gemini 3.1 Pro: **CLEAN**. Independently reproduced the old nested-index
  residual and verified current and legacy `Data[Data[0]]` and `Data[Data[1]]`
  witnesses now reach Full fidelity; 134 focused tests pass.

No findings remain at the reviewed head.

PR #2904, opened 2026-07-19
Bound distinct .NET tool wrapper redirect chains (#2813)

richlander:
Fixes #2813.

## What

Adds an explicit redirect-hop budget (`MaxToolWrapperRedirectHops = 8`) to `PackageExtractor` for .NET tool-wrapper chains. When a wrapper graph redirects through more than 8 **distinct** package IDs, extraction returns a visible `PackageExtractionOutcome.Error` naming the chain, instead of continuing to download and transactionally commit an arbitrarily long sequence of packages.

## Why

PR #2810 already stops **repeated** package IDs from causing recursive stack overflow (cycle detection via `visitedPackageIds`). But a malformed wrapper graph can still redirect through an unbounded chain of *distinct* IDs, each hop costing a network download + transactional cache commit — a resource-exhaustion vector on untrusted input. This adds the missing distinct-hop bound as non-blocking follow-up from the #2810 review.

## Behavior

- Ordinary one-hop wrapper→payload behavior is preserved.
- The existing cycle check runs **before** the new budget check, so a cycle still yields the specific cycle error rather than the generic hop-limit error.
- At exactly 8 redirects the payload still resolves; the 9th redirect errors **before** the next download.

## Tests

Three new boundary tests in `PackageAcquisitionConcurrencyTests`:
- `ExtractPackageAsync_ToolWrapperRedirectAtHopLimitReturnsPayload` — 8 hops resolve the payload.
- `ExtractPackageAsync_ToolWrapperRedirectBeyondHopLimitReturnsErrorBeforeNextDownload` — 9th hop errors with no extra download (`RequestCount` asserted).
- `ExtractPackageAsync_CycleAtHopLimitReturnsSpecificCycleError` — cycle precedence preserved at the boundary.

`dotnet run --project src/dotnet-inspect.Tests -c Release -- -class DotnetInspector.Tests.PackageAcquisitionConcurrencyTests` → 15 passed, 0 failed.

PR #2911, opened 2026-07-19
Enforce network policy for package metadata

### LINKED ISSUE #2897: Network capability gating is observational-only; Services hardcode nuget.org endpoints
## Problem

The repository rule says network work "must remain explicit or capability-gated," and the code has the right single seam (`HttpClientFactory`) — but the capability gate is **observational only**: `NetworkTelemetry` records whether a request is allowed by policy and then lets it proceed regardless. Separately, Services hardcode nuget.org endpoints, ignoring configured package sources.

## Evidence

- Single-seam construction holds: zero `new HttpClient(` sites outside `src/DotnetInspector.Core/HttpClientFactory.cs` (repo-wide, excluding tests); offline is genuinely enforced in the handler chain (`OfflineHandler` throws, `HttpClientFactory.cs:96-97,202-210`); untrusted-artifact URLs go through an SSRF-hardened client (`CreateUntrustedFetchClient` :116-138) at all seven call sites.
- But policy is telemetry-tagging, not enforcement: `NetworkTelemetry.RecordRequestStarting` only records `IsAllowedByPolicy` (`src/DotnetInspector.Core/NetworkTelemetry.cs:147-170`; policy classification :193-199 covers only `VulnerabilityData`), and `HttpClientFactory.cs:31-32` says so explicitly ("Requests are still allowed to proceed"). The single `NetworkTelemetry.Allow` scope is `Commands/PackageCommand.cs:423`; the vulnerability fetches in `PackageMetadataService.cs:306-308,323-325` would still hit the network without it.
- Hardcoded endpoints: `PackageMetadataService.cs:33,106,168,250,305` and `Inspectors/RidPackageVerifier.cs:40` bake in nuget.org URLs regardless of configured sources.

## Suggested direction

1. Either make the policy enforcing (block disallowed categories in the handler chain, like `OfflineHandler`) or rename/re-document it as telemetry so the "capability-gated" claim stops overpromising.
2. Thread configured package sources through `PackageMetadataService` (or document nuget.org-only scope for those metadata families).

Found during the 2026-07 component-interaction architecture review (see tracker issue).

richlander:
## Summary

Enforce package network capability policy at the shared HTTP seam and keep configured package sources authoritative.

- blocks vulnerability and GitHub advisory traffic unless the selected view explicitly authorizes the shared vulnerability capability
- applies the same authorization to single- and multi-package inspection
- prevents custom/file sources named `api.nuget.org` from enabling NuGet.org aggregate metadata
- keeps `NetworkPolicyException` visible instead of converting policy violations into empty metadata
- routes package metadata and RID companion verification through the configured source set

Fixes #2897.

## Duplicate check

No open or recently merged PR implements this issue. The recent Softserve redirect, classic-corpus, and tuple-switch branches were excluded because their work already merged as #2904, #2888, and #2901.

## Validation

- solution build
- Services tests: 266 passed
- focused changed tests: 54 passed
- markdownlint on README and changed design docs
- full CLI tests reached two unrelated P/Invoke rendering failures; both reproduce unchanged on `origin/main`

## Adversarial review

Round one used isolated worktrees and the same threat-focused prompt:

- Claude Opus found that multi-package detailed inspection lacked the vulnerability authorization scope and silently dropped known CVEs.
- Gemini Pro independently found the same multi-package false-negative, plus a `file://api.nuget.org` source spoof and GitHub advisory traffic outside the vulnerability capability.

Resolved in `a21a96cd` by sharing the authorization predicate across command paths, binding advisory traffic to the vulnerability capability, requiring HTTP(S) NuGet.org sources, propagating policy exceptions, and adding focused regressions.

Fixed-head review:

- Gemini Pro: clean on `a21a96cd`, including a real multi-package CVE run.
- MAI-Code: clean on `a21a96cd`. It reproduced a partial/full metadata-cache collision, then confirmed that behavior is unchanged on `origin/main` and is pre-existing rather than introduced here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,10 @@ scope.
 | `library X -S "SourceLink Integrity"` | Content verification (slow, opt-in) | Downloads every tracked source file and compares its hash to the PDB checksum; a mismatch exits non-zero. Never runs in a default flow. |
 | `package X -S Signals` | Full package signals | Package and dependency signals, including known vulnerabilities, package age, dependency vulnerability/deprecation counts, and dependency age. |
 
+Vulnerability-service traffic is capability-gated. It runs only for detailed
+package inspection or an explicitly selected network-using package section;
+requests outside that scope are blocked at the shared HTTP handler.
+
 ## Integrations
 
 `Integrations` is a library section for ecosystem support such as AI, ASP.NET
--- a/README.md
+++ b/README.md
@@ -155,6 +155,9 @@ scope.
 Vulnerability-service traffic is capability-gated. It runs only for detailed
 package inspection or an explicitly selected network-using package section;
 requests outside that scope are blocked at the shared HTTP handler.
+NuGet.org-wide statistics, verification, deprecation, and vulnerability
+metadata are queried only when `api.nuget.org` is among the configured package
+sources. RID companion-package verification follows the configured source list.
 
 ## Integrations
 
--- a/docs/design/untrusted-data-threat-model.md
+++ b/docs/design/untrusted-data-threat-model.md
@@ -163,6 +163,12 @@ exceeding any one produces a visible rejection rather than a partial identity.
 
 ### Network and caches
 
+Network capability policy is enforced in the shared HTTP handler after the
+attempt is recorded for diagnostics and before it reaches the transport.
+Traffic families that require explicit authorization, currently vulnerability
+data, must run inside their matching `NetworkTelemetry.Allow` scope. Offline
+mode remains the broader prohibition over every traffic family.
+
 Network access derived from inspected content must be explicit in the command
 surface, use the untrusted-fetch client, have a timeout, and retain provenance.
 Cache paths must be hashed or use validated single components. Downloads should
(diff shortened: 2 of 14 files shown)


richlander:

- **Claude Opus:** reproduced a multi-package vulnerability false-negative: the batch path lacked the capability scope and silently dropped CVEs.
- **Gemini Pro:** independently reproduced that defect and also found `file://api.nuget.org` source spoofing plus GitHub advisory traffic outside the vulnerability capability.
- Resolved in `a21a96cd` with a shared single/multi authorization predicate, HTTP(S)-only NuGet.org source recognition, advisory-family capability gating, visible policy exceptions, and focused regressions.
- **Fixed-head review:** Gemini Pro and MAI-Code are clean on exact head `a21a96cd`.
- MAI also found a partial/full metadata-cache collision. No resolution commit applies: an A/B against `origin/main` confirmed the same cache key/value behavior is pre-existing and unchanged by this PR.

richlander:
Ready to merge

Issue #2898, opened 2026-07-18
Substrate hygiene: Instructions charter drift, third IL decode loop, duplicated dominance kernel, Text split


richlander:
## Problem

Four contained charter-drift/duplication items in the shared substrate layers. Each is small; together they blur the "one decoder / shared kernels" story the docs tell.

## Items

1. **ILInspector.Instructions charter drift**: ~1,745 of ~5,073 lines (`IlBodyDiff.cs` 927, `IlAssemblyDiff.cs` 377, `IlDiffPrinter.cs` 128, `IlFindings.cs` 313) are diff/rendering/Finding-projection code inside the project whose csproj describes itself as "Layer 0 decode + basic-block substrate" (`ILInspector.Instructions.csproj:12`). The Findings dependency exists only for `IlFindings.cs`. Either move the diff family up a layer or re-charter the project.
2. **Third IL decode loop**: `src/ILInspector.Decompiler/Pipeline/Ir/IlProjection.cs:146-159` hand-rolls a raw decode loop with inline `0xFE` two-byte-opcode handling, plus private opcode tables `IsMetadataToken` (:247), `IsBranch` (:255), `OperandLength` (:262) duplicating Instructions' `ILOpcodeExtensions` (`ILOpcodeExtensions.cs:28,51,100`) and `InstructionDecoder`'s operand mapping. Unlike the IrImporter granularity gap (documented as pending in `docs/design/instruction-substrate.md`), this duplication is unmanaged.
3. **Duplicated dominance kernel above ControlFlow**: Analysis carries a private forward-dominator `DominatorTree` (`src/ILInspector.Analysis/LibraryBodyIndex.cs:2335-2440`, ~110 lines) implementing the same Cooper–Harvey–Kennedy algorithm as `src/ILInspector.ControlFlow/PostDominators.cs:66+`. ControlFlow (420 lines total) is the layer built to hold exactly this; add forward dominators there and delete the private copy. Minor: `ReachingDefinitions.cs:332-368` re-reads operand bytes from raw IL although `DecodedInstruction.OperandValue` already carries them.
[...]

PR #2920, opened 2026-07-20
Dedupe IlProjection IL decode loop onto ILInspector.Instructions substrate (#2898 item 2)

richlander:
## Summary

Item 2 of #2898: `IlProjection.cs` hand-rolled a third independent IL
decode loop, duplicating opcode sizing, branch handling, and metadata-token
classification already owned by the `ILInspector.Instructions` substrate.

## Change

- Replaced `IlProjection`'s private decode tables with
  `InstructionDecoder.Decode` plus a focused operand-format projection.
- Added a `ReadOnlySpan<byte>` decoder entry point while retaining the existing
  `byte[]` overload and null contract, so `IlProjection` consumes its
  `ImmutableArray<byte>` method body without copying.
- Sized the projection list by decoded instruction count rather than IL byte
  count.
- Replaced the test's incidental dependency on the deleted production
  `OperandLength` method with a minimal compiler-produced switch fixture: a
  four-entry dense jump table plus one sparse comparison sharing a return leaf.

Rendered operand formats and projection behavior are unchanged.

## Evidence

Current head:

```text
Build succeeded

dotnet run --project src/ILInspector.Instructions.Tests -c Release --no-build
146 passed, 0 failed

  -class "ILInspector.Decompiler.Tests.SwitchBranchRenderingTests"
10 passed, 0 failed
```

## Risk

Behavior-preserving substrate deduplication. The existing `byte[]` decoder API
remains available, the new span path uses the same decode core, and the fixture
continues to exercise real Release-compiled IL without reproducing production
opcode tables.

Refs #2898


richlander:
The `AsSpan()` → `ToArray()` switch appears to be API impedance, not a decoder requirement: `InstructionDecoder.Decode` currently accepts only `byte[]`, but `DecodeCore` immediately constructs `ILReader`, whose storage is already `ReadOnlySpan<byte>`, and the operand readers ultimately slice spans. Since `MethodBody.IL` is already an `ImmutableArray<byte>`, both new `ToArray()` calls make a full per-render copy solely to satisfy that signature. I would keep this path span-based and fix the shared decoder boundary (for example, a span-capable entry point while preserving any required `byte[]` compatibility/null contract) rather than introducing the copy. Adjacent allocation note: `new List<Instr>(il.Length)` also sizes by IL bytes rather than instruction count, so it can substantially over-allocate compared with the previous growing list.

richlander:
The new `SwitchTableFixture` is effectively the deleted production
`OperandLength` switch copied into the test and renamed `Classify`. The old test
did not own this constant logic; it reflected over the production method because
that method happened to compile to the desired control-flow shape. Moving almost
the whole deleted implementation into the test preserves that incidental
coupling as duplication.

The fixture can be much smaller while remaining compiler-produced. With the
current SDK, this is sufficient:

```csharp
static int Classify(int value)
{
    switch (value)
    {
        case 0:
        case 2:
            return 8;
        case 1:
        case 3:
        case 100:
            return 2;
        default:
            return 0;
    }
}
```

It produces a residual IL jump table for `0..3`, a comparison branch for `100`
sharing the second return leaf, and labeled targets. `PrintRaised` emits
`__dotnet_inspect_switch`, the sparse `if (V_0 == 100) goto ...`, and both target
labels, so it exercises the exact invariant consumed by
`AssertGotoTargetsHaveLabels` without reproducing an opcode table.

I would replace the large enum fixture with a minimal shape fixture like this.

richlander:
## Adversarial review: clean

Reviewed exact head `15ed336f9db72ead5f596940653ccf207c82ef41`
against `main` at `8426693d4f4df9466395916c6bc536581309ce5b`.

| Reviewer | Verdict |
| --- | --- |
| Claude Opus 4.8 | **CLEAN** |
| Gemini 3.1 Pro | **CLEAN** |

Both reviewers independently verified:

- exact projection parity across integer, floating-point, token, variable,
  branch, switch, prefix, and `calli`/`InlineSig` operands;
- preserved malformed-IL exception normalization and fail-visible behavior;
- source-compatible `byte[]` behavior plus safe, non-recursive span dispatch;
- zero-copy `ImmutableArray<byte>` → `ReadOnlySpan<byte>` decoding and
  instruction-count list sizing;
- the Release-compiled minimal fixture's dense jump table, sparse comparison,
  shared return leaf, and non-vacuous target-label assertion;
- clean integration with current `main`.

Real-run evidence included 146 instruction-substrate tests, 10 switch-rendering
tests, full-solution builds, explicit overload/malformed-IL probes, fixture
IL/output dumps, and baseline/head projection comparisons over CoreLib,
System.Linq, and System.Text.Json. No blocking or actionable findings remain.

**Non-blocking follow-up:** the repeated test pattern
`MetadataSource.Open(type.Assembly.Location)` plus `IrImporter.Import(...,
type.FullName, method)` should be wrapped by test infrastructure so fixture
callers pass the type and method without owning source-loading details. This
does not belong in the product importer because its product boundary remains
metadata-only and free of inspected-assembly loading.

Ready to merge

Issue #2907, opened 2026-07-19
Decompiler: propagate raised stackalloc initializers through Span wrapper


richlander:
## Current state

The #2869 stackalloc-initializer slice reconstructs the constant payload hidden
behind `localloc` + raw `cpblk`:

```csharp
unsafe
{
    int* values = stackalloc int[] { 1, 2, 3 };
}
```

For a target-typed `Span<T>` source, the initializer is now recovered but the
owned stack slot still separates it from the exact `Span<T>(void*, int)`
wrapper:

```csharp
Span<int> values;

unsafe
{
    values = new Span<int>((void*)(stackalloc int[] { 1, 2, 3 }), 3);
}
return values[0] + values[2];
```

`StackAllocSpanPass` already raises a direct stackalloc argument, but it does
not currently propagate a recovered initializer through the compiler-owned
stack slot used by this lowering.

## Fully raised goal

```csharp
Span<int> values = stackalloc[] { 1, 2, 3 };
return values[0] + values[2];
```

## Required slices

1. Prove unique current-body ownership of the stackalloc definition and the
   exact `Span<T>(void*, int)` consumer.
2. Propagate the recovered `StackAllocArray` through that slot without crossing
   aliases, nested bodies, intervening effects, or mismatched element/count
   facts.
3. Reuse the existing exact platform-constructor identity and preserve pointer
   stackalloc forms unchanged.
4. Add compiler-produced pointer and Span positives, plus close negatives for
   duplicate loads/stores, type/count mismatch, nested-body shadowing, and
   non-platform constructors.
5. Prove the rendered output no longer needs the synthetic unsafe wrapper and
   compiles back in the target-typed Span form.

## Evidence

- Parent initializer endpoint: #2869
- Initializer reconstruction branch:
  `raise/issue-2869-stackalloc-initializers`
- `StackAllocSpanPass` handles only direct `StackAllocate`/`StackAllocArray`
  constructor arguments today; the compiler-produced #2869 witness reaches it
  through a `LoadStackSlot`.

richlander:
Closed by #2956 (merged `93d02e09`). Verified the fully-raised goal is now produced against a real compiled fixture:

```csharp
Span<int> values = stackalloc[] { 1, 2, 3 };
return values[0] + values[2];
```

Confirmed via `dotnet-inspect member ... -S "Decompiled Source"` and `RaisingPassTests.StackAllocSpanInitializerThroughSlot_RaisesAtFullFidelity`.

Issue #2869, opened 2026-07-18
Decompiler: raise stackalloc initializer cpblk lowerings


richlander:
## Current state

PR #2448 made pointer and Span stackalloc initializer lowerings explicit owned
residuals. They remain `Partial`, retain `cpblk`, and end in `return default;`.
The closed issue #2428 therefore recorded the frontier but did not raise it.

Representative current fallback:

```csharp
byte* S_256 = stackalloc byte[12];
ReadOnlySpan<int> V_1 = new int[] { 1, 2, 3 };
/* Unsupported ... cpblk */
return default;
```

## Fully raised goals

```csharp
int* values = stackalloc int[] { 1, 2, 3 };
return values[0] + values[2];
```

```csharp
Span<int> values = stackalloc[] { 1, 2, 3 };
return values[0] + values[2];
```

## Required slices

1. Recognize compiler-produced `localloc` plus static-data/temporary plus
   `cpblk` initialization as one stackalloc initializer shell.
2. Recover the exact element type, count, constants, byte order, padding, and
   destination ownership for pointer and Span forms.
3. Build typed stackalloc-initializer IR and consume only the proven copy shell;
   preserve subsequent uses and evaluation order.
4. Reject non-constant sources, partial/overlapping copies, mismatched sizes,
   escaped destinations, and hand-authored `cpblk` near misses.

## Evidence

- Historical residual pin: #2448
- Closed ownership issue: #2428
- Current fixtures:
  `StackallocPointerInitializer` and `StackallocSpanInitializer` in both new
  and legacy unsafe assemblies
- Current gate requires `Partial`, `UnsupportedConstruct`, visible `cpblk`, and
  `return default;`.

Add compiler-produced positive and close-negative fixtures across SDK lowering
variants, valid C#, and compile-back opcode/control-flow evidence.

richlander:
Implementation started on branch `raise/issue-2869-stackalloc-initializers` from `e3a88b09`.

richlander:
Resolved by #2921 (merged), which recognizes localloc+cpblk stackalloc initializer shells and raises pointer/Span forms to a typed StackAllocArray initializer, per the required slices in this issue.

PR #2921, opened 2026-07-20
Raise stackalloc initializer cpblk lowerings

richlander:
- Fixes/advances #2869
- Changes stackalloc-initializer `cpblk` lowerings (pointer and `Span<T>`/`ReadOnlySpan<T>` forms) are now recognized and raised to a typed `StackAllocArray` initializer instead of leaving them as an opaque residual `cpblk`
- Card revision: `3900d8c0`

## Change

**Conclusion:** **PASS** — a new `StackAllocInitializerPass` proves exclusive ownership of the compiler-emitted scratch buffer (exact-count `LoadStackSlot` usage, no reordering/aliasing between the `stackalloc`/`cpblk` pair, matching sizes as `Int32` constants, only trusted-platform `Span`/`MemoryMarshal` accessors as the copy source) before replacing the lowering with a typed initializer, and declines with an honest `Partial`/`unsupported cpblk` fallback whenever any of those checks fail.

Each example below shows the full member — signature and body — so the `Before`
and `After` decompilations line up directly against the original source. All
`Before`/`After` text is the real product output for the two witness members in
`ILInspector.Decompiler.Fixtures.NewUnsafe.StackallocInitializerResiduals`
(`member ... -S "Decompiled Source"`), taken on `origin/main` and this head
respectively.

### Original source

```csharp
public static class StackallocInitializerResiduals
{
    public static int StackallocPointerInitializer()
    {
        int* values = stackalloc int[] { 1, 2, 3 };
        unsafe
        {
            return values[0] + values[2];
        }
    }

    public static int StackallocSpanInitializer()
    {
        Span<int> values = stackalloc[] { 1, 2, 3 };
        return values[0] + values[2];
    }
}
```

### Before

Both forms lowered to the same opaque residual: the scratch buffer, the
`cpblk`, and the return value are all lost, so the member does not relate to the
original source at all.

```csharp
// StackallocPointerInitializer — origin/main
public static unsafe int StackallocPointerInitializer()
{
    ReadOnlySpan<int> V_1;

    unsafe
    {
        byte* S_256 = stackalloc byte[12];
        V_1 = new int[] { 1, 2, 3 };
        _ = S_256;
        _ = S_256;
    }
    _ = V_1[0];
    _ = 12;
    /* Unsupported IL_001E cpblk: opcode is outside the slice */
    return default;
}

// StackallocSpanInitializer — origin/main (identical residual shape)
public static unsafe int StackallocSpanInitializer()
{
    ReadOnlySpan<int> V_1;

    unsafe
    {
        byte* S_256 = stackalloc byte[12];
        V_1 = new int[] { 1, 2, 3 };
        _ = S_256;
        _ = S_256;
    }
    _ = V_1[0];
    _ = 12;
    /* Unsupported IL_001E cpblk: opcode is outside the slice */
    return default;
}
```

### After

```csharp
// StackallocPointerInitializer — this head (fidelity: Full)
public static unsafe int StackallocPointerInitializer()
{
    unsafe
    {
        int* values = stackalloc int[] { 1, 2, 3 };
        return *values + values[2];
    }
}

// StackallocSpanInitializer — this head (fidelity: Full, typed initializer
// recovered; final target-typed collapse is #2907)
public static unsafe int StackallocSpanInitializer()
{
    Span<int> values;

    unsafe
    {
        values = new Span<int>((void*)(stackalloc int[] { 1, 2, 3 }), 3);
    }
    return values[0] + values[2];
}
```

### Fully raised

For the pointer form, the `After` decompilation is in the fully raised state
(modulo the pre-existing `*values` vs `values[0]` dereference spelling, which is
unrelated to this PR).

For the `Span<T>`/`ReadOnlySpan<T>` form, the `After` decompilation is Full
fidelity with the typed initializer correctly recovered, but not yet collapsed
through the `new Span<T>(pointer, length)` wrapper into the target-typed
initializer spelling:

```csharp
public static int StackallocSpanInitializer()
{
    Span<int> values = stackalloc[] { 1, 2, 3 };
    return values[0] + values[2];
}
```

- Required tracking issue: #2907 — collapse the recovered `StackAllocArray`
  through the `new Span<T>(pointer, length)` wrapper into a target-typed
  `Span<T>` initializer expression.

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass (`dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config`) |
| Focused tests | `StackAllocInitializerPassTests` 53/53 passed |
| Decompiler fast suite | Pass (`-trait- Speed=Slow`) |
| C# diff smoke | Pass — the pinned baseline was regenerated (RowCount 174->175) for the one legitimately raised line in the V2 fixture; see the baseline-regeneration comment |
| Real witness | `StackallocInitializerResiduals::StackallocPointerInitializer` and `::StackallocSpanInitializer` (`NewUnsafe` + `LegacyUnsafe`) fixed, shown above |
| Close negatives | `StackallocInitializerNegatives` (Boolean/floating-point RVA and empty-initializer forms) correctly stay unraised |

## Decompiler quality

**Conclusion:** **PASS** — recovery is limited to a compiler-shaped, exclusively-owned scratch-buffer `cpblk` pair. Unsupported graphs retain their existing honest residual/`Partial` output, and the two negative fixtures confirm the gate declines lossy element kinds.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class ILInspector.Decompiler.Tests.StackAllocInitializerPassTests
```


richlander:
Adversarial review round 1 (head 8a2a156e):
- Reviewer B (Gemini 3.1 Pro): clean, no findings.
- Reviewer A (GPT-5.6 Sol): 1 blocking finding (Finding 4, RVA Boolean/NaN bit-loss decoding, confirmed pre-existing in \RvaSpanPass.DecodePrimitiveElements\ on \origin/main\ already \u2014 out of scope for this PR, not introduced here) and 1 real blocking finding (Finding 1): an empty trusted-Span \get_Item(0)\ source (zero-length \stackalloc\ initializer) could be recognized as a valid zero-element initializer, silently detaching the accessor call + CopyBlock and erasing the \IndexOutOfRangeException\ the original code would throw.

Fixed in e06b0aa0: added \
equiredElementCount > 0\ guard in \StackAllocInitializerPass.cs\ plus a regression test (\EmptySpanGetItem_Declines\) verified to fail without the fix and pass with it. Targeted suite: 53/53 passing.

Re-review of the fixed head (e06b0aa0) in progress with both reviewers.

richlander:

Both reviewers examined all 6 required attack points (aliasing/escape soundness, size/element-type mismatch, trusted-platform gating, RVA-decode correctness, negative-fixture adequacy, fully-raised scope-boundary honesty), each with real build/test/dump evidence in an isolated worktree.

**Reviewer A:** GPT-5.6 Sol
**Reviewer B:** Gemini 3.1 Pro

### Round 1 (head `8a2a156e`)
- Reviewer B: clean, PASS, no findings.
- Reviewer A: 2 blocking findings.
  1. **Empty-span `get_Item(0)` exception elision** — a zero-length `stackalloc` initializer sourced via a trusted `Span.get_Item(0)` accessor was accepted as a valid zero-element initializer, silently detaching the accessor call + `CopyBlock` and erasing the `IndexOutOfRangeException` the original code would throw.
  2. **RVA Boolean/floating-point bit-loss** — the shared `RvaSpanPass.DecodePrimitiveElements` decoder (pre-existing, verified unchanged on `origin/main`) is not bit-preserving for `Boolean` (`chunk[0] != 0` canonicalizes any nonzero byte to `true`) or floating-point NaN payloads. This PR's new RVA call site widens exposure to that pre-existing gap because, unlike other RVA call sites, it has no `CreateSpan`/`InitializeArray`/constructor-identity requirement — any RVA field qualifies.

### Resolution
- Finding 1 fixed in `e06b0aa0`: added a `requiredElementCount > 0` guard before accepting a `SpanLiteral` source. Added regression test `EmptySpanGetItem_Declines`, verified to fail without the fix and pass with it.
- Finding 2 fixed in `01763d5f`: this pass's RVA path now declines `Boolean`/`Single`/`Double` element types entirely until the shared decoder is made bit-preserving (tracked as a separate out-of-scope follow-up issue). Added a compiled regression fixture `StackallocBooleanInitializer` (NewUnsafe + LegacyUnsafe), verified via `--dump` to correctly stay `fidelity: Partial` with an honest `/* unsupported cpblk */` residue instead of raising a possibly bit-inexact literal.
- Targeted suite after both fixes: `StackAllocInitializerPassTests` 53/53 passing; `LadderRung6GateTests`/`MetadataProvenanceTests`/`UnsafeEmitterTests`/`TypeRefDecoderRecursionTests` 84/87 (the 3 failures are pre-existing `RecompileFail` environment failures, confirmed identical on a clean `origin/main` checkout, unrelated to this change).

### Round 2 (final head `01763d5f`)
- Reviewer A: **PASS** — both fixes confirmed correct and complete.
- Reviewer B: **PASS** — confirmed the Boolean/Single/Double decline is correctly scoped (integer types unaffected) and doesn't reject any case that should legitimately still raise.

**Both reviewers are clean on the final head. Ready to merge.**

richlander:
Ready to merge.

richlander:
## Follow-up: C# diff smoke baseline regeneration (head `3900d8c0`)

The prior `Ready to merge` was posted at `01763d5f`; the `csharp-diff-smoke` CI check was failing there. Root cause: **legitimate baseline drift**, not a correctness regression.

`DiffFixtures.V2/DiffSample.cs` contains `int* values = stackalloc int[1]` with an initializer, so this raise intentionally changes that member's decompiled output — adding one raised C# line. The pinned smoke snapshot (`tools/CSharpDiffHarness/corpus/diff-fixtures-baseline.json`) still recorded the pre-raise metrics, so the baseline guard fired:

```
RowCount             174 -> 175
csharp.line.added     76 ->  77
Line operation kind  129 -> 130
```

Regenerated the committed baseline via the harness `--emit-snapshot` path. The diff is exactly those three counts — no new failure buckets, no other member changes. Verified locally that the smoke now passes against the updated baseline (harness exits 0, zero baseline findings).

**Review status:** this follow-up commit is a mechanical, data-only regeneration of a pinned snapshot to match the already-reviewed raise output; it touches no product code. Per the repository's adversarial-review policy (mechanical/data-only changes are exempt when the low-risk rationale is stated), the two clean exact-head reviews of the raise logic at `01763d5f` remain the correctness evidence for this PR. Re-affirming **Ready to merge** once CI is green on `3900d8c0`.

richlander:
Ready to merge

CI is green on the final head `3900d8c0`: `build-net10` pass, `csharp-diff-smoke` pass (against the regenerated baseline), and the full `test` lane pass. The two clean exact-head reviews of the raise logic at `01763d5f` stand; the only follow-up commit is the mechanical, data-only smoke-baseline regeneration documented above.

Issue #2916, opened 2026-07-20
Decompiler: authoritative printer-spellable value / legal ref-producer predicate


richlander:
### Problem

Passes that lift an IR expression into a **value** or **`ref`** sink must decide
whether the C# printer can legally spell it there. Today that decision is
re-derived privately, one pass at a time, by enumerating node kinds — and the
authoritative knowledge (what renders with a leading `ref`, what has no
conversion to `dynamic`, what is a legal `ref` producer) is scattered across
`CSharpPrinter.*` (e.g. `LoadFieldAddress => "ref …"` at
`src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.cs:2241`, plus the `Deref`
/ address handling threaded through `CSharpPrinter.Members.cs` and
`CSharpPrinter.Numerics.cs`).

### Evidence this is costing us

`DynamicCallSitePass`'s receiver gate (`ContainsUnraisableReceiver` /
`IsUnraisableReceiver`,
`src/ILInspector.Decompiler/Pipeline/Passes/DynamicCallSitePass.cs`) has, over
**four adversarial-review rounds**, grown a bespoke traversal that answers one
question — *"can the printer spell this receiver as a value castable to
`dynamic` (and, through a `ref` context, as a legal `ref` producer)?"* Each
round found one more wrapper/context the private walk missed:

- direct address receivers,
- unmanaged-pointer / function-pointer value categories (CS0030),
- `ByRef`-to-pointer look-through,
- `TypeSpec` / `TypeRef` recursion guards,
- conversion / switch-arm wrappers, late spill inlining, pinned wrappers,
- and now **`Unbox` hidden in a ref-conditional arm** — the walk pushes only
  `conditional.Condition` for a `ByRef`-typed `Conditional`, so a leaf `Unbox`
  survives to malformed output like `cond ? ref ref (int)x : …` (CS-invalid).
  (Reproduced on `raise/issue-2866-dynamic` head `436af938`; blocking finding
  from Gemini Pro.)

That non-convergence is the signal: the pass is re-implementing a **general**
predicate about the printer, not a dynamic-specific rule.

### Proposed capability

An authoritative predicate owned by the CSharp/printer layer (which owns C#
spelling) that any pass consults before lifting an expression into a value or
`ref` sink. Given an `IrExpression` and a target position — plain value,
`ref`-value, or `(dynamic)`-castable value — it answers *"can the printer spell
[...]

PR #2925, opened 2026-07-20
Spell Unbox as ref Unsafe.Unbox<T> across all C# printer substrates

richlander:
## What

`unbox T` yields a managed pointer *into* the box (`ByRef T`). Its only valid,
faithful C# spelling is the `System.Runtime.CompilerServices.Unsafe.Unbox<T>(o)`
intrinsic — a `ref T` into the box. A bare cast `(T)x` is an `unbox.any`
**copy**: it reads the same value but is not the box, so it is not an assignable
place (`ref (T)x` is CS0445, `out (T)x` is CS0206) and it silently drops
mutations applied through it.

The C# printer spells `Unbox` in **three** substrates, none of which handled it
correctly:

- **`Deref`** (`989c1866`) — value reads, ref-ternary arms (`cond ? ref a : ref b`),
  `= ref` RHS, write-into-box. Fell through to the generic `ByRef` arm →
  `ref (T)x`, doubling to `ref ref (T)x` (CS1525); an earlier iteration traded
  that for the still-invalid copy `ref (int)o` (CS0445).
- **`ArgumentLvalue`** (`cf9b933c`) — ref-returns (`ReturnText`), `ref`/`out`
  args, explicit-`in` args (`RefArgument`). **Excluded** `Unbox` → default
  `ref (T)x` / `out (T)x` (CS0445 / CS0206).
- **`ReceiverText`** (`ce4e610f`) — field/property/method-access receivers.
  Emitted the value-copy `((T)x).Member`: fine for a read, but a **mutating**
  instance call silently ran on a copy (mutation dropped), and a member
  **assignment target** `((T)x).Field = v` was CS0445.

**Fix (places):** route all three substrates' *place* `Unbox` through the shared
`UnsafeUnboxText` helper, emitting `Unsafe.Unbox<T>(o)` fully qualified
(`global::System.Runtime.CompilerServices.Unsafe`) so it resolves without a
`using`. Valid and faithful in every place: ref/out/ref-return get a genuine
assignable place; receivers reach the in-box place so mutations are carried back
(faithful to `unbox; call` / `unbox; stfld`); write-into-box stores into the box.
`ArgumentPlace` (implicit-`in`) keeps the `(T)x` value form. **Value reads are
handled separately — see the pivot below.**

**Positional gate (`e4bd43a3`, supersedes `d4b39ec6`):** `Unsafe.Unbox<T>`
constrains `T` to a non-nullable value type (`where T : struct`), which
`Nullable<T>` and an open generic parameter do not satisfy (CS0453). The cast
`((T)o)` is a valid fallback **only in a value position** — in a ref/out/write
place it is never an assignable place at all. So the decision is split by
position, not applied uniformly:

- **Place positions** (`ArgumentLvalue`, `Deref`) keep the unconditional
  intrinsic — the only C# form that is a place. `Unsafe.Unbox<T>` compiles for a
  `where T : struct` generic parameter; a boxed `Nullable<T>` / open `T` has no
  place form in C# at all, so a visible compile error there is faithful to
  un-spellable IL rather than a regression.
- **The member-access receiver** (a value position, where `((T)o)` compiles)
  routes through `UnboxReceiverText`: intrinsic only for a spellable non-nullable
  value type — a named definition or non-`Nullable` generic instance — else the
  cast. `unbox` yields a value type for valid IL, so the printer trusts it,
  except for `Nullable<T>`, a **resolver-known reference type** (`TypeShapes ==
  Reference`, only reachable from malformed IL), and open generic parameters
  (constraint unknown), which keep the compiling cast.

## Value reads — normalized upstream, not spelled in the printer (pivot: `UnboxValueReadPass`, `2691cc7d`)

The value/place decision does **not** belong in the printer. A pure **value read**
of an unboxed pointer — `LoadIndirect(Unbox(T, o))`, i.e. `unbox T; ldobj T` — is
the same operation as `unbox.any T`, whose faithful C# spelling is the universal
cast `(T)o`, valid for **every** `T` (value, `Nullable<T>`, reference, open
generic). `UnboxValueReadPass` normalizes those reads to the existing `UnboxAny`
node upstream of the printer, so:

- The printer never spells a value read as `Unsafe.Unbox<T>`; value reads fall out
  of the existing `UnboxAny` → `(T)o` rendering — **zero printer changes** for the
  value-read case.
- The place spelling above (`Deref` / `ArgumentLvalue`) is now reached **only** for
  a genuine place, where the "no valid cast fallback" reasoning actually holds.

This resolves the round-1 value-read blocker (both **GPT 5.6 Sol** and **Gemini
3.1 Pro** independently reproduced `LoadIndirect(Unbox)` value reads spelling the
unconditional intrinsic → **CS0453** for a boxed `Nullable<T>` / open generic,
where `(T)o` compiles). The normalization is **inert on the compiled corpus**
(`csc` emits `unbox.any` for `(T)o`, never `unbox; ldobj`), so no fidelity opcode
shift. The pass is classified in `NativePasses` (`EmitArtifact`), alongside its
IL-idiom-canonicalization sibling `IdentityConvertPass`. Guards: a
**reinterpreting** read (`ldobj U` over `unbox T`, `U != T`) and a **volatile**
read (acquire semantics a plain cast would drop) are left explicit.

**Pass placement (`302fea01`):** the pass runs **after** the final
slot-collapsing passes (`ExpressionInliningPass(slotsOnly:true)` /
`StackSlotCopyPropagationPass`) and before `CoercionInsertionPass`. A spilled
`ref T S = unbox o` slot only collapses to `LoadIndirect(Unbox)` once the late
inliner inlines it, so normalizing earlier would leave that re-formed value read
spelled as the CS0453 intrinsic. Running last among the reshaping passes closes
that window; running before coercion keeps the minted `UnboxAny` coerced at its
sink. (Found in adversarial re-review — GPT 5.6 Sol and Gemini 3.1 Pro both
reproduced the spilled-slot escape.)

## Evidence

Compile-verified with real `csc` (exit 0), including runtime faithfulness:

- **Deref** — ref-ternary, plain read, write `…Unsafe.Unbox<int>(o) = default(int);`.
- **ArgumentLvalue** — ref-return, ref-arg, out-arg, in-arg — all compile.
- **ReceiverText** — copy receiver `((MyStruct)o).Mutate()` drops the mutation
  (X=1); `Unsafe.Unbox<MyStruct>(o).Mutate()` preserves it (**X=42**);
  `Unsafe.Unbox<MyStruct>(o).X = 5` compiles (X=5, was CS0445); reads still read.
- **Positional gate** — a `where T : struct` generic-parameter unbox in a ref
  place spells `ref Unsafe.Unbox<T>(o)` and **compiles**; a `Nullable<T>` /
  generic-parameter / reference-shaped **receiver** keeps `((T)o)` (the intrinsic
  would be CS0453).

Regression tests:

- `PrinterPrecedenceTests` — `Deref_RefConditionalWithUnboxArm_SpellsUnsafeUnbox`,
  `ReturnRefUnbox_SpellsUnsafeUnbox`, and
  `ReturnRefUnbox_GenericParameter_SpellsUnsafeUnbox` (compile-gated).
- `RefArgumentRenderingTests` — `RefArgumentUnbox_SpellsUnsafeUnbox`,
  `OutArgumentUnbox_SpellsUnsafeUnbox`.
- `CSharpPrinterReceiverTests` — `UnboxReceiver_InstanceCall_SpellsUnsafeUnbox`,
  `UnboxReceiver_FieldAssignment_SpellsUnsafeUnbox`,
  `UnboxReceiver_Nullable_KeepsCastNotUnsafeUnbox`,
  `UnboxReceiver_GenericParameter_KeepsCastNotUnsafeUnbox`, and
  `UnboxReceiver_KnownReferenceType_KeepsCast` (compile-gated).
- `UnboxReceiverRenderingTests.UnboxFieldReceiver_SpellsUnsafeUnbox` — **real
  compiled fixture** (`CfgBoxed.FieldEquals`).
- `UnboxValueReadPassTests` — `NullableValueRead_NormalizesToUnboxAnyCast_NoCS0453`
  and `StructValueRead_NormalizesToUnboxAnyCast` (compile-gated), plus
  `ReinterpretingRead_IsNotNormalized` / `VolatileRead_IsNotNormalized` (tree
  guards), and `SpilledSlotValueRead_ThroughFullPipeline_NormalizesNotUnsafeUnbox`
  (drives the full `IrPasses.Default` pipeline to lock the pass ordering).
  `LoweringCoverageTests` classifies the new pass in `NativePasses`.

Full `ILInspector.Decompiler.Tests` (2824 tests, non-`Slow`): **zero
regressions** — an identical set of 53 pre-existing environmental failures
(Windows `\r\n` line-endings + Roslyn compile-back harness) as the clean head.

## Follow-up (tracked, not in this PR)

**#2929** — the **primary** `Expression(Unbox)` spelling `ref (T)x`
(`CSharpPrinter.cs`) is unchanged, so `Convert(UIntPtr, Unbox)` →
`(nuint)(ref (int)o)` (CS1525) remains. That path is a managed-pointer→integer
conversion (`conv.u` on a `ByRef`), semantically distinct from `Unsafe.Unbox`
(which reads the value).

**#2946** — the `unbox` **receiver** Unknown-shape boundary: `CanReceiveViaUnsafeUnbox`
treats `TypeShape.Unknown` as unboxable, so a receiver over an unresolved/reference
type could spell `Unsafe.Unbox<C>(o).M()` (CS0453) where the cast compiles.
Adjudicated malformed-IL-only (valid-IL `unbox` targets a value type) and
non-blocking; **unaffected by the value-read pivot** (receiver/member-access path,
not a `LoadIndirect` value read). The authoritative target-context decision
belongs to the #2916 spellability oracle.

**#2951** — an unbox value read whose `Unbox` is decoupled from its dereferencing
`LoadIndirect` by an intervening node — a ref-ternary
`LoadIndirect(Conditional(Unbox, Unbox))`, or a materialized `ref T` local —
escapes this pass's direct `load.Address is Unbox` match and keeps the CS0453
intrinsic spelling. Reachable only from adversarial / non-`csc` IL (a compiler
never emits `unbox` for `Nullable<T>` / open `T`; for `where T : struct` both
spellings compile) and pre-existing invalidity at the base, so it is the same
honest-degrade residual bucket as #2929 / #2946. The general fix distributes the
value-read deref down to the `Unbox` leaves.

## Context

Found while investigating #2916 (the general spellability capability). This is a
standalone printer correctness fix, not the capability itself. Requires
two-family adversarial review per AGENTS.md.


richlander:
## Review — approve (mechanical, corrective; one wording nit)

Reviewed head `00ba148c`. Verdict: **low-risk, correct, strictly corrective blast radius.** Good to merge once the Evidence wording is softened.

### Correctness
`Deref`'s contract is "the place, dereferenced — no leading `ref`/`*`." `Unbox` was missing from the switch, so it fell through to the generic `{ ResultType.Kind: ByRef } => Operand(address)` arm, and `Operand(Unbox)` returns the node's own **ref-producer** spelling `ref (T)x` (`CSharpPrinter.cs:2238`), violating the contract. The new arm is exactly that spelling minus `ref`, and it already exists verbatim as the sibling deref-value case at `CSharpPrinter.Members.cs:607`. So this mirrors an established pattern, not a new one.

### Blast radius — enumerated every `Deref` caller
The only behavior change is `Deref(Unbox)`: `ref (T)x` → `(T)x`. Strictly corrective across all sites:

- **`ref`-prefixing callers** (ref-local/slot `= ref Deref(...)` at 2031/2032/2044/2045; ref-ternary arms at 2724): previously double-`ref` (`ref ref (T)x`, CS1525) → now single `ref (T)x`.
- **non-prefixing callers** (`DerefLoad`, `InitObject`, `InlineArraySpanConversion`, …): previously carried an erroneous leading `ref` on a read → now clean.

No caller wanted `ref (T)x` (the contract forbids it). `Unbox` never reaches `Deref` via the `Members.cs:606/619` switches — those match `Unbox` at 607 first. New test passes locally (`Deref_RefConditionalWithUnboxArm_DoesNotDoubleRefKeyword`, 0.11s).

### Nit (non-blocking, wording only)
Evidence says **"After: `flag ? ref (int)o : ref n` (correct)."** `(T)x` spells `unbox.any` — a *value* (rvalue) — not the `unbox` managed place, so `ref (int)o` is still not valid C# (CS1510: a `ref` value must be an assignable variable). The fix makes the output **less malformed** and contract-consistent on a path the `Deref` Conditional case already documents as *"an inexpressible merge… best effort, not a correctness guarantee"* (synthetic-only, `BooleanFoldingPass.FoldSlotDiamond`). Suggest rewording to: *"removes the double-`ref`; the arm remains a best-effort spelling of an inexpressible ref-merge,"* rather than "correct."

### Adversarial review
Per AGENTS.md, simple/mechanical changes don't require the two-family pass — this is a single missing-case addition mirroring an existing sibling arm, with a verified corrective-only blast radius and no reachable-C# behavior movement (the double-`ref` path is synthetic). Stated here as the low-risk justification.

richlander:
## Adversarial review: blocked

Reviewed exact head `00ba148c77fe8935500acb0434cbabe181e4d0bb`
against merge base `37cd43fe7cc47141fcb0d31093fb09f8dc7ea7fd`.
The current `main`-only commit does not touch either changed file.

Reviewers:

- Claude Opus 4.8: **FINDINGS**
- Gemini 3.1 Pro: **FINDINGS**

### Blocking finding: the After output is still invalid at `Full`

Both reviewers independently reproduced the exact synthetic IR and output:

```text
Full: return (flag ? ref (int)o : ref n);
```

Compiling the equivalent method fails:

```text
error CS0445: Cannot modify the result of an unboxing conversion
```

`Unbox` produces a managed reference in IR, but the C# unboxing conversion
`(int)o` is not an assignable variable and therefore cannot be a `ref`
conditional arm. The patch removes the doubled keyword but changes CS1525 into
CS0445 while retaining `Full` fidelity.

The fix must either produce a verified valid and semantically faithful spelling
for this context or degrade honestly when no such spelling is available.
Potential directions include a value-context conditional, a verified
`Unsafe.Unbox<T>` ref-return spelling, or an explicit unsupported/Partial path;
whichever is chosen needs evidence for validity and fidelity rather than only
parseability.

### Blocking test gap

`Deref_RefConditionalWithUnboxArm_DoesNotDoubleRefKeyword` currently asserts
only that output contains `ref (int)o` and excludes `ref ref`. That pins the
CS0445-invalid spelling. The class already has `AssertCompiles`; the corrected
test must compile the rendered body and pin the valid contract.

### Non-blocking sibling observation

A separate pre-existing path still renders
`Convert(UIntPtr, Unbox(int, o))` as:

```text
Full: return (nuint)(ref (int)o);
```

That fails with CS1525. It is not caused by this patch, but it demonstrates the
same broader `Unbox` ref-producer boundary. Do not route it through `&(T)o`;
that spelling also fails. Either include the shared rule in the fix or track
the sibling explicitly as follow-up.

After resolving the blocking findings, rerun both reviewers on the new exact
head. This head does not have two clean reviews.

richlander:
## Reconciliation — blocking findings resolved on new head `989c1866`

Thanks to both reviewers. Round-1 findings addressed:

**Blocking: After output invalid at `Full` (CS0445).** Claude Opus 4.8 and
Gemini 3.1 Pro both reproduced `return (flag ? ref (int)o : ref n);` →
CS0445 ("Cannot modify the result of an unboxing conversion"). Confirmed via
real `csc`. The value-copy `(T)o` approach is abandoned. `Deref(Unbox)` now
spells the box place as `System.Runtime.CompilerServices.Unsafe.Unbox<T>(o)`, a
`ref T`-returning intrinsic that is valid and faithful in every `Deref` position
— value read (auto-deref), ref-prefixed (ref-ternary arm, `= ref` RHS), and
write-into-box (`Unsafe.Unbox<T>(o) = v`). Compile-verified for all three.
Emitted fully qualified so it resolves without a `using`.

**Blocking: test pins the invalid spelling.** Resolved. The regression test
(`Deref_RefConditionalWithUnboxArm_SpellsUnsafeUnbox`) now asserts the
`Unsafe.Unbox<int>(o)` spelling, excludes both `ref ref` and `ref (int)o`, and
**compiles the rendered body** through the existing `AssertCompiles` gate.

**Non-blocking sibling `Convert(UIntPtr, Unbox)` → `(nuint)(ref (int)o)`
(CS1525).** Tracked as follow-up in #2929. It originates from the untouched
**primary** `Expression(Unbox)` spelling and is a managed-pointer→integer
conversion, semantically distinct from `Unsafe.Unbox` (which reads the value),
so it is intentionally not folded into this rule and not routed through `&(T)o`,
per the guidance.

**Self-review wording nit.** The "(correct)" claim on `ref (int)o` was indeed
overstated — that spelling is itself invalid. Moot now: the output is
`ref Unsafe.Unbox<int>(o)`, which compiles.

**Regression check.** Full `ILInspector.Decompiler.Tests` (2811, non-`Slow`):
zero regressions — an identical set of 53 pre-existing environmental failures
(Windows `\r\n` line-endings + Roslyn compile-back) on both the clean head and
this head, verified via a stash A/B.

Re-triggering the two-family adversarial pass on exact head `989c1866`.

richlander:
## Scope expanded + review reconciliation (head `cf9b933c`)

The first-round two-family review (Gemini 3.1 Pro + GPT 5.6 Sol) at head
`989c1866` confirmed the committed `Deref(Unbox)` diff was **CLEAN / valid /
faithful**. Reconciling every finding:

- **Gemini Finding 3** (High/Med — `ArgumentLvalue(Unbox)` still emits invalid
  `ref (int)o` for ref-returns / ref-args / out-args): **RESOLVED here.** Commit
  `cf9b933c` routes `ArgumentLvalue(Unbox)` through the same
  `Unsafe.Unbox<T>(o)` intrinsic. Compile-verified all four ref-places
  (ref-return, ref-arg, out-arg, in-arg) with real `csc`; added regression tests
  (ref-return compile-gated; ref/out string-assert).
- **Gemini Finding 2** (High — `((T)x).Mutate()` mutates a copy of a boxed value
  type): **deferred as follow-up.** Pre-existing and orthogonal to ref-place
  spelling; the read-receiver path is currently value-exact, so the fix is
  delicate. Noted in the PR body.
- **GPT Finding 1** (Med — `Unsafe.Unbox` recompiles to `call`, not `unbox`, yet
  labeled `Full`): **dismissed.** Fidelity here is **static** (`Full` = "every
  construct raised; representable C#"), assigned by `FidelityRemarks` diagnostic
  causes — not a compile-back opcode roundtrip. The printer **already** emits
  `Unsafe.NullRef<T>()` (also a `call`-based intrinsic) as the faithful `Full`
  spelling for ref-local null-init with no fidelity cause, so `Unsafe.Unbox`
  follows established precedent. Treating it as sub-`Full` would be inconsistent
  and is a separate cross-cutting change (would also cover `NullRef`).
- **GPT Finding 2** (Med — same-FQN `Unsafe` shadow → CS0117): general naming
  property of fully-qualified intrinsic spellings, very low real-world risk;
  noted, not actioned.

Re-running two-family adversarial review on this exact head (`cf9b933c`) before
marking ready.

richlander:
## Scope expanded again + reconciliation (head `ce4e610f`)

Following the clean re-review at `cf9b933c`, the remaining Gemini finding
(Finding 2 — `ReceiverText(Unbox)` mutation infidelity) is now **RESOLVED**,
completing the `Unbox` ref-place fix across **all three** printer substrates.

- **Gemini Finding 2** (High — `((T)x).Mutate()` silently mutates a copy;
  `((T)x).Field = v` is CS0445): **RESOLVED** in `ce4e610f`.
  `ReceiverText(Unbox)` now spells `Unsafe.Unbox<T>(o).Member`. Compile- **and
  run**-verified with `csc`: the old copy drops the mutation (X=1), the intrinsic
  preserves it (**X=42**), the field-assignment compiles (X=5, was CS0445), and
  reads still read (7). Added compile-gated call + field-assignment tests and
  updated the real-fixture `CfgBoxed.FieldEquals` canary to assert the intrinsic
  spelling.
- **Gemini Finding 3** (`ArgumentLvalue`) — resolved earlier in `cf9b933c`.
- **GPT Finding 1** (static-fidelity `Full` label): dismissed — fidelity is
  static, and `Unsafe.NullRef` is established `call`-based-intrinsic precedent.
- **GPT Finding 2** (same-FQN `Unsafe` shadow): general naming property, very low
  risk; noted.

`#2929` remains the only tracked follow-up (the primary `Expression(Unbox)` →
`(nuint)(ref (int)o)` conversion path, semantically distinct).

Full suite: 2816 tests non-`Slow`, zero regressions (same 53 pre-existing
environmental failures). Re-running two-family adversarial review on this exact
head (`ce4e610f`) before marking ready.

richlander:
## Reconciliation — `d4b39ec6` (constraint-gate fix)

Both reviewers re-ran the `ReceiverText` head (`ce4e610f`) and **independently
found the same regression**, so I am attributing it jointly:

- **GPT 5.6 Sol** and **Gemini 3.1 Pro** both reported that `Unsafe.Unbox<T>`
  carries `where T : struct`, so emitting it for a `Nullable<T>` unbox target
  produces **CS0453** where the prior value-copy cast `((int?)o)` compiled.
- **Gemini additionally** flagged the sibling case: an **open generic
  parameter** (which may be unconstrained) — same CS0453 violation.

**Fix (`d4b39ec6`):** gate the shared `UnsafeUnboxText` helper on
`CanSpellUnsafeUnbox(unbox.Type)`. It emits `Unsafe.Unbox<T>` only for a
concrete non-nullable value type — a named value-type *definition* or a
*non-`Nullable` generic instance* — and falls back to the value-copy cast
`((T)o)` for `Nullable<T>`, open generic parameters, and every other shape. The
gate lives in the one helper, so it applies uniformly to all three substrates
(`Deref`, `ArgumentLvalue`, `ReceiverText`).

The cast fallback is exact in value and member-access positions (`Nullable<T>`
is immutable; an open `T` matches the prior spelling), and in a
`ref`/`out`/write place a boxed `Nullable<T>` / open `T` has no valid C# form at
all — so the cast is the honest best effort, not a fidelity loss versus any
compilable alternative.

Compile-verified: `Unsafe.Unbox<int?>(o)` and `Unsafe.Unbox<T>(o)` are CS0453;
the gated cast forms compile. Two new compile-gated regression tests:
`UnboxReceiver_Nullable_KeepsCastNotUnsafeUnbox` and
`UnboxReceiver_GenericParameter_KeepsCastNotUnsafeUnbox`. Full suite: 2818 tests,
identical 53 pre-existing environmental failures, zero regressions.

Re-running two-family review on this fixed head (`d4b39ec6`) — will not mark
ready until both sign off clean.

richlander:
## Reconciliation — `e4bd43a3` (positional gate; supersedes `d4b39ec6`)

**GPT 5.6 Sol** returned **CLEAN on `d4b39ec6`**. **Gemini 3.1 Pro** found that the
`d4b39ec6` uniform gate — emit `Unsafe.Unbox<T>` for a concrete value type, else
fall back to the cast `((T)o)` — was wrong because the cast is a valid fallback
**only in a value position**. Three points, all reproduced and addressed:

- **CS0445 regression in place positions (Gemini, High).** For a struct-constrained
  generic parameter (`unbox !!T` in a `ref`/`out`/ref-return position), the uniform
  gate fell back to `ref ((T)o)` — **CS0445** — whereas `ce4e610f`/`cf9b933c`
  emitted the ungated `ref Unsafe.Unbox<T>(o)`, which **compiles** (`Unsafe.Unbox<T>`
  satisfies `where T : struct`). A ref-place has no valid cast form, so the cast was
  never a safe fallback there.
- **`Definition` assumed to be a value type (Gemini, CS0453).** The Kind-only gate
  trusted every `TypeRefKind.Definition`, so malformed IL (`unbox System.String`)
  produced `Unsafe.Unbox<string>` — CS0453 — where the prior cast compiled.
- **Open `Nullable`1` definition (Gemini).** Same root cause: a Kind check does not
  verify value-type-ness.

**Fix (`e4bd43a3`):** split the decision by position instead of gating the shared
helper uniformly.

- **Place positions** (`ArgumentLvalue`, `Deref`) keep the **unconditional**
  `Unsafe.Unbox<T>` — restoring the `cf9b933c`/`989c1866` spelling both reviewers
  already cleared. This fixes the CS0445 regression. A boxed `Nullable<T>` / open
  `T` has no assignable-place form in C# at all, so a compile error there is a
  faithful, visible failure on un-spellable IL — not a regression from any working
  state.
- **The member-access receiver** (a value position) routes through a new
  `UnboxReceiverText`: emit the intrinsic only for a spellable non-nullable value
  type, else the cast. `unbox` yields a value type for valid IL, so a named
  definition or non-`Nullable` generic instance qualifies — **except** `Nullable<T>`,
  a **resolver-known reference type** (`TypeShapes == Reference`, only reachable from
  malformed IL — fixes the `unbox String` case where knowable), and open generic
  parameters. Those keep the compiling cast.

This trusts the IL's value-type guarantee (so a bare synthetic struct definition
still spells the faithful intrinsic — `IsKnownNonNullableValueType` would wrongly
demote it to a copy when no shape/hint is resolved), while diverting the malformed
reference-type case the resolver can actually see.

Two compile-gated regressions added:
`ReturnRefUnbox_GenericParameter_SpellsUnsafeUnbox` (`T : struct` ref-return keeps
the intrinsic and compiles) and `UnboxReceiver_KnownReferenceType_KeepsCast`
(reference-shaped receiver keeps the cast). Full suite: 2820 tests, identical 53
pre-existing environmental failures, zero regressions.

Re-running two-family review on this fixed head (`e4bd43a3`) — will not mark ready
until both sign off clean.

richlander:
## Adversarial re-review reconciliation — head `e4bd43a3`

Two independent families reviewed the fixed head in isolated worktrees. Verdicts and adjudication:

### Convergent finding: receiver `Unknown`-shape → `Unsafe.Unbox<T>` → CS0453
Both **Gemini 3.1 Pro** and **MAI Flash** flagged the same point (High/High, BLOCKING): `CanReceiveViaUnsafeUnbox` treats `TypeShape.Unknown` as unboxable, so an `unbox` receiver over an unresolved/reference type emits `Unsafe.Unbox<C>(o).M()` (CS0453) where the old cast `((C)o).M()` compiled. Both proposed requiring positive value-type confirmation.

**Adjudication — real behavior, but not a valid-IL correctness bug; not a merge blocker.**
- The `Unbox` IR node is produced **only** from the `unbox` opcode (`IrImporter.cs:1309-1313`; `unbox.any` → a separate `UnboxAny` node). `unbox` requires a value-type operand in valid IL (ECMA-335 III.4.32), so an `Unknown`-shape `Unbox.Type` in valid IL is an **unresolved value type** — for which `Unsafe.Unbox<T>` is correct, compiles, and is faithful. The reference-type path is reachable **only from malformed/adversarial IL**. Both repros were synthetic IR; neither showed a valid-IL or corpus reproduction.
- The proposed fix would **regress the common case**: unresolved value-type receivers (a struct from an unreferenced dependency) would fall back to the lossy value-copy cast, dropping exactly the mutation faithfulness this PR restores.
- Emitting CS0453 for malformed `unbox RefType` is **consistent** with the PR's already-accepted ungated ref/out/write stance ("faithful to un-spellable IL rather than a regression").

The tension (and its proper home — the #2916 spellability/target-context oracle, plus the pre-Core-3.0/netstandard2.0 `Unsafe` availability corollary) is captured in #2946. **Not blocking this PR.**

### Fidelity (all three reviewers agree)
`unbox` → `call Unsafe::Unbox<T>` is **not** a Full→non-Full regression: the prior `(T)x` cast already recompiled to `unbox.any`, so `Unbox`-node methods were never opcode-Full. Empirically confirmed on net11: `Unsafe.Unbox<Pt>(o).X = 5` → `call Unbox; stfld` (writes into the box), `Unsafe.Unbox<Pt>(o).Bump()` → `call Unbox; call Bump` (mutates the box, not a copy) — strictly more faithful than the old cast, which mutated a discarded copy. `Unsafe.Unbox<T>` is available in the net11 shared framework.

### Verdict
The design is sound and a genuine faithfulness/validity improvement (ref/out/write/receiver positions now spell a real place). The one convergent finding is adjudicated as malformed-IL-only and routed to #2946 as a non-blocking follow-up. **Not blocking merge.**

richlander:
## Round-1 reconciliation + design pivot (new head `c6b3fc4e`)

Both round-1 reviewers — **GPT 5.6 Sol** and **Gemini 3.1 Pro** — independently
reproduced the same blocking finding on `e4bd43a3`: a pure **value read** of an
unboxed pointer (`LoadIndirect(Unbox(T, o))`) was routed through the
unconditional `Unsafe.Unbox<T>` intrinsic, which is **CS0453** for a boxed
`Nullable<T>` or an open generic parameter — even though the value read has a
universally valid spelling, the cast `(T)o`. The convergence is credited to both.

### Resolution — moved out of the printer (per @richlander's layering feedback)

Rather than add another positional gate inside the printer, the value/place
decision now lives **upstream** of rendering:

- **`UnboxValueReadPass`** (`2691cc7d`) normalizes `unbox T; ldobj T` value reads
  to the existing `UnboxAny` node — because `unbox; ldobj` **is** `unbox.any`.
  Those reads then fall out of the printer's existing `UnboxAny` → `(T)o`
  rendering with **zero printer changes** for the value-read case.
- The printer's place spelling (`Deref` / `ArgumentLvalue` → `Unsafe.Unbox<T>`)
  is now reached **only** for a genuine place, where the "no valid cast fallback"
  reasoning actually holds.

Guards: a reinterpreting read (`ldobj U` over `unbox T`, `U != T`) and a volatile
read are left explicit. The normalization is inert on the compiled corpus (`csc`
emits `unbox.any` for `(T)o`, never `unbox; ldobj`), so no fidelity opcode shift.
The pass is classified in `NativePasses` (`EmitArtifact`), and
`LoweringCoverageTests` covers it. Doc touch-up in `c6b3fc4e`.

### #2946 (receiver Unknown-shape) — separate path, unaffected

The receiver Unknown-shape boundary that this round also flagged travels the
`UnboxReceiverText` / member-access path, **not** the `LoadIndirect` value read
that the pass diverts. It remains adjudicated malformed-IL-only and non-blocking,
tracked in #2946; the pivot does not change it either way.

The earlier "not blocking merge" verdict was on `e4bd43a3` and is **stale** — it
predates this pivot. Re-running two-family adversarial review on the fixed head
`c6b3fc4e` before marking ready.

richlander:
## Round-2 (fixed-head) review reconciliation — new head `302fea01`

Two-family adversarial re-review on `c6b3fc4e` (Gemini 3.1 Pro + GPT 5.6 Sol,
isolated worktrees) returned **BLOCKING**, converging on one finding:

### Finding 1 (BOTH reviewers, high confidence) — FIXED

`UnboxValueReadPass` ran **before** the final `ExpressionInliningPass(slotsOnly:true)`.
A spilled `ref T S = unbox o` slot collapses to `LoadIndirect(Unbox)` only *after*
the late inliner inlines it, so the re-formed value read reached the printer
unnormalized and spelled `Unsafe.Unbox<Nullable<int>>(o)` → CS0453 — the exact
defect the pass exists to fix.

Reproduced independently through the pipeline:

```text
HEAD ORDER  (UnboxValueRead -> Inline):  return Unsafe.Unbox<Nullable<int>>(o);   // CS0453
FIXED ORDER (Inline -> UnboxValueRead):   return (Nullable<int>)o;                 // valid
```

**Fix (`302fea01`):** relocated the pass to run after
`ExpressionInliningPass` / `StackSlotCopyPropagationPass` / `SlotMaterializationPass`
and before `CoercionInsertionPass`. Added
`SpilledSlotValueRead_ThroughFullPipeline_NormalizesNotUnsafeUnbox`, which drives
the full `IrPasses.Default` pipeline to lock the ordering (the isolated-pass tests
could not catch a late re-formed pair). 52/52 focused tests green.

### Finding 2 (Gemini) — tracked as #2951, non-blocking

Value reads whose `Unbox` is decoupled from its `LoadIndirect` by an intervening
node — a ref-ternary `LoadIndirect(Conditional(Unbox, Unbox))` (verified still
CS0453 at the fixed head) or a materialized `ref T` local — escape the pass's
direct `load.Address is Unbox` match. Reachable **only from adversarial / non-`csc`
IL** (a compiler never emits `unbox` for `Nullable<T>` / open `T`; for
`where T : struct` both spellings compile), and pre-existing invalidity at the
base (`ref (T)x` / CS0445) — not a regression. Same honest-degrade residual bucket
as #2929 / #2946; filed as **#2951** with the general-fix direction.

Re-running the two-family review on the fixed head `302fea01` before marking ready.

richlander:
## Round-2 adversarial reconciliation — fixed head `c6b3fc4e` (the `UnboxValueReadPass` pivot)

Two independent families reviewed the pivot in isolated worktrees at `c6b3fc4e`. **Both CLEAN.**

### Gemini 3.1 Pro — CLEAN
Cleared all seven attack points, notably:
- **Semantic equivalence:** `unbox T; ldobj T` ≡ `unbox.any T` (same `InvalidCastException`/`NullReferenceException` behavior; C# compiles `(T)o` value reads to `unbox.any`).
- **Ordering:** `UnboxValueReadPass` runs after the `UnboxAny` consumers; `IsPatternPass` matches `LoadIndirect{Address: LoadArgument}` and `UnionSwitchExpressionPass` matches `Box{Operand: LoadIndirect}` — neither depends on `LoadIndirect(Unbox)`, so injected `UnboxAny` nodes can't mis-fire them.
- **Structural `TypeRef.Equals`**, receiver-path (#2946) separateness, volatile guard.

### MAI Flash — CLEAN
Empirical evidence:
- **Inert on csc output:** `ReadStruct`/`ReadNullable`/`ReadGeneric<T>` probes all emit `unbox.any`, never `unbox; ldobj` — the pass never fires on compiler output, so no fidelity opcode shift.
- **IL equivalence via reflection-emit:** `unbox; ldobj` and `unbox.any` behave identically for boxed `int`, boxed `Nullable<int>`, null, and type-mismatch.
- **Scope:** rewrites only `LoadIndirect(Unbox)`; leaves reinterpret (`ldobj U` over `unbox T`, U≠T) and volatile reads explicit; does not touch the `UnboxReceiverText` receiver path.

### My independent review (Opus) — concur
- `TypeRef.Equals` is fully **structural** (`TypeRef.cs`), ignoring only the generic-parameter *name* as a naming aid → no false-negative CS0453 leak from instance identity.
- Ordering verified (`IrPass.cs:347` after `IsPatternPass`:211 and `UnionSwitchExpressionPass`:220/263).
- 15 unbox-spelling tests green on `c6b3fc4e`, incl. the new value-read normalization (boxed-`Nullable` → no `Unsafe.Unbox`, Roslyn-compiles) and both guards (reinterpret, volatile).
- The pivot is the right layering fix: the value/place decision lives **upstream** (`unbox; ldobj` → `UnboxAny`), the printer stays thin, and `Deref(Unbox)→Unsafe.Unbox<T>` is reached only for a genuine ref/out/write place.

### Residual
- **#2946** (receiver `Unknown`-shape boundary) is a separate path, unaffected by the pivot; remains adjudicated malformed-IL-only and **non-blocking**.

**Verdict: two clean fixed-head reviews (Gemini + MAI) plus independent concurrence. The value-read over-reach that round-1 caught is correctly resolved. Ready to merge.**

richlander:
Ready to merge

**CI:** green on head `a94d0860` — `test (ubuntu-26.04, linux-x64)` passed (7m1s), `build-net10` and `csharp-diff-smoke` pass; `il-diff-smoke` / `markdownlint` / `pack` path-skipped.

**Reviews:** two CLEAN fixed-head adversarial reviews (GPT 5.6 Sol + Gemini 3.1 Pro) on `302fea01`, both confirming the Finding 1 ordering fix and agreeing #2951 (ref-ternary / materialized-ref-local value reads) is non-blocking adversarial-IL-only follow-up.

**Delta since the reviewed head (`302fea01` → `a94d0860`):** a clean `origin/main` merge (no conflicts) plus one governance-only commit registering `UnboxValueReadPassTests.cs` in `DynamicCompilationSiteInventoryTests` (retain reason + fingerprint 29→30 files / 36→37 sites). No product-behavior change, so the two CLEAN reviews still stand.

Follow-ups tracked: #2951 (Finding 2), #2929, #2946.

Issue #2864, opened 2026-07-18
Decompiler: recover expression-tree factory graphs as lambdas


richlander:
## Current state

PRs #1153 and #1157 pinned expression-tree lambdas as an intentional honesty
frontier. PR #1669 later expanded the compiler fixture. A simple source lambda
still renders as explicit factory calls:

```csharp
ParameterExpression V_0 = Expression.Parameter(typeof(int), "x");
return Expression.Lambda<Func<int, int>>(
    Expression.Add(V_0, Expression.Constant(1)),
    V_0);
```

Captured/member forms remain `Partial` when field/member tokens survive. The
current output is honest and often valid, but it is not the fully raised source
form.

## Fully raised goal

```csharp
return x => x + 1;
```

Captured expressions should likewise recover their legal lambda form:

```csharp
return x => x > threshold;
```

## Required slices

1. Define the exact canonical `Expression.Parameter` / body /
   `Expression.Lambda` graph subset that can be converted without changing
   parameter identity, quoting, delegate type, tail-call, naming, or node
   semantics.
2. Prove graph ownership and reject shared, externally used, mutated, or
   hand-composed near misses outside that canonical subset.
3. Recover member/token consumers as typed member-access nodes before lambda
   construction; do not invent a general standalone spelling for `ldtoken`.
4. Preserve C# expression-tree language restrictions and keep unsupported
   graphs in explicit factory-call or honest `Partial` form.

## Evidence

- Factory-call frontier and real compiler fixtures: #1153
- Missing-discriminator decline: #1157
- Broader rung-9 fixture: #1669
- Closed ledger issue: #1142
- Current tests explicitly assert `Expression.*` calls and reject `=>`.

Add compiler-produced positive and close-negative graphs, manual factory-call
near misses, captured/member cases, compile-back validity, and expression-tree
semantic equivalence evidence.

richlander:
Implementation started on branch `raise/issue-2864-expression-trees` from `e3a88b09`.

PR #2926, opened 2026-07-20
Raise homogeneous-Int32 expression-tree factory graphs to lambdas (#2864)

richlander:
- Advances #2864
- Raises the fully-owned homogeneous-`Int32` expression-tree factory graph to a
  source lambda
- Card revision: `cc9b273c8c3aab7e9caec109ed1952d3458c71ec`

## Change

**Conclusion:** **PASS** — the pass proves exact platform factory identity,
single-source parameter identity, graph ownership, spellable unique names,
return-sink compatibility, unchecked arithmetic identity, and absence of
compiler-folded constant-only subtrees before replacing factory calls.

### Original source

```csharp
public static Expression<Func<int, int>> Simple()
    => x => x + 1;
```

### Before

```csharp
ParameterExpression V_0 = Expression.Parameter(typeof(int), "x");
return Expression.Lambda<Func<int, int>>(
    Expression.Add(V_0, Expression.Constant(1)),
    V_0);
```

### After

```csharp
return x => unchecked(x + 1);
```

### Fully raised

For the supported homogeneous-`Int32` arithmetic slice, the After
decompilation is in the fully raised state. Issue #2864 remains open for the
non-`Int32`, captured/member, and broader expression-node slices.

### Why the `unchecked` wrapper

The `unchecked(...)` in the raised lambda is load-bearing for round-trip
fidelity, not cosmetic. In an expression tree the overflow context is not
erased into plain arithmetic — the C# compiler encodes it as a *distinct
factory node*: `Expression.Add` for unchecked versus `Expression.AddChecked`
for checked. Which node the compiler emits for a bare `x + 1` depends on the
**consuming project's** `CheckForOverflowUnderflow` setting, which the
decompiled source does not carry.

Empirically (net11.0), the same lambda text recompiles to different trees:

| Consuming project | `x => x + 1` | `x => unchecked(x + 1)` |
| --- | --- | --- |
| Default | `Add` | `Add` |
| `CheckForOverflowUnderflow=true` | `AddChecked` | `Add` |

This pass only matches the *unchecked* arithmetic factories (`Add`,
`Subtract`, `Multiply`, `Divide`, `Modulo`), so the recovered lambda must
recompile to `Expression.Add`, never `AddChecked`. The explicit `unchecked(...)`
pins that node regardless of the consuming project's overflow default. Without
it, a bare `x => x + 1` dropped into a checked-default project would rebuild
`Expression.AddChecked` — a semantically different tree — silently violating the
semantics-preserving raise contract.

Under a default (unchecked) project the wrapper is redundant, which is why it
can read as noise; its job is to defend the non-default checked-project case.
Checked expression trees (`AddChecked` et al.) are deliberately **not** raised
by this pass — they decline and retain their explicit factory spelling — so the
raise never spoofs a checked tree into an unchecked lambda.

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass |
| Platform-provenance tests | 8 passed |
| Focused expression-tree tests | 44 passed |
| Decompiler fast suite | 2,844 passed |
| Compile-back/tree oracle | Pass |
| Close negatives | Manual aliases, duplicate/unspellable parameters, wrong return sinks, checked factories, constant-only subtrees, and unsigned platform lookalikes decline |

The exact-handle platform check is bounded by the shared `TypeSpecGuard`;
self-referential modifiers and excessive nesting fail closed instead of
aborting the process.

## Decompiler quality

**Conclusion:** **PASS** — recovery is limited to a compiler-shaped,
fully-owned arithmetic graph. Unsupported graphs retain their explicit factory
calls or existing honest `Partial` output.


richlander:

Two-model exact-head adversarial review (decompiler raising is correctness-critical), isolated in separate worktrees with read-only discipline and real build + test runs required. Both signed off **CLEAN** on the final head **`cc9b273c`**.

- **Claude Opus 4.8:** CLEAN. Probed the raise against **real csc-emitted IL** through the full product path (`IrImporter.Import` → `IrPasses.Run` → `CSharpPrinter.Print`): operand/precedence order preserved for non-commutative ops (`x - y`, `a - b*c`, `10 - x`); `unchecked(...)` wrapper applied exactly where the expression-tree factory is checked (Subtract→SubtractChecked) and omitted where there is no checked node (Divide/Modulo), confirmed against a Roslyn oracle; `checked(...)`, `Func<uint,uint>`, `Func<long,long>` correctly declined; idempotent. Verified the exact-handle token-anchored platform-trust gate (never a simple-name lookup), the cross-assembly spoof fixture declining, and malformed-metadata fail-closed behavior (self/nil `modreq(TypeSpec)`, 100k-nested SZARRAY) via `TypeSpecGuard` + the top-level catch.
- **Gemini 3.1 Pro:** CLEAN. Independently confirmed the nil-`TypeSpecificationHandle` concern is **unreachable through the product path** (`ResolveParentType`/`TypeRefDecoder` throws `BadImageFormatException` earlier and it is caught at the subsystem boundary). Verified parameter-aliasing/ownership decline via the statement ledger, checked/unchecked preservation, and the transitive constant-fold guard (`DependsOnParameter` on every Binary node).

### Non-blocking notes (no action)
- Opus reproduced a **pre-existing** `LadderIteratorGateTests.IteratorFixture_RecoveredRowsCompileBackOpcodeExact` `OperandDiff` failure **identically on base `516c379e`** — environment/ilasm-related, unrelated to this PR.
- Quality-only: a narrow malformed TypeSpec band (1025–4096 bytes) reaches `DecodeSignature` and fails closed via `CrashFunction` rather than declining earlier; unreachable via csc, no crash/hang. And the positive test surface has one csc-emitted fixture (`x+1`) with hand-authored multi-param factories — Opus's real-IL probe confirmed those raise correctly, so it is a test-surface gap only.

### Verification (head `cc9b273c`)
- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config` — clean.
- Focused `ExpressionTreeLambdaTests` + `ExpressionTreeFidelityTests` + `ExpressionTreeProvenanceTests` + `TypeRefDecoderRecursionTests` → 44/44 pass (both reviewers, independently).
- Gate classes (`LadderRung9GateTests`, `FidelityGateTests`, `LoweredFidelityGateTests`, `Speed=Slow`) → pass; decompiler fast suite → 2,844 pass.

Both fixed-head reviews are clean. Pending CI confirmation before marking Ready to merge.

richlander:
Ready to merge

CI is green on the exact-head `cc9b273c`: `build-net10`, `csharp-diff-smoke`, `il-diff-smoke`, and the full `test` lane all pass. Two clean exact-head adversarial reviews (Claude Opus 4.8 + Gemini 3.1 Pro) are reconciled above, both against real csc-emitted IL with no blocking findings.

richlander:
## Merge resolution + fixed-head review (`cc9b273c` → `30c0e433`)

Updated the branch by merging `origin/main`. `origin/main` had independently
landed the **same** platform-trust hardening this branch already carried
(the `TypeSpecGuard`-bounded `PlatformDeclaringTypeHandleProvider` exact-handle
resolution, the `trustedPlatform` hoist, and the `DeclaringTypeIsTrustedPlatform`
metadata fact), so both sides converged on the same design.

**Conflicts resolved (all kept this branch's reviewed versions):**

- `IrImporter.cs` — MemberReference case: kept the single `bool trustedPlatform`
  local; it feeds both the `accessorKind` gate and the auto-merged
  `DeclaringTypeIsTrustedPlatform` fact.
- `IrImporter.cs` — `IsTrustedPlatformTypeSpecification`: kept the version with the
  explicit `if (baseType.IsNil) return false;` guard (behaviorally ≥ the terser
  `origin/main` form; fails closed on nil / non-`TypeReference`).
- `IrImporter.cs` — `PlatformDeclaringTypeHandleProvider.GetTypeFromSpecification`:
  kept the `TypeSpecGuard.TryEnter` + `using(scope)` bounding.
- `IrNodes.cs`: removed the duplicate `DeclaringTypeIsTrustedPlatform` property both
  branches added (kept one copy). `origin/main`'s non-conflicting changes survived
  (`IsTrustedPlatformMemberReference` now `internal static`; `PlatformToken` removed).

**Validation:** full solution builds clean; fast decompiler suite **2,907 passed /
0 failed**; CI green (build, `csharp-diff-smoke`, `il-diff-smoke`, test).

**Adversarial review (fixed head `30c0e433`):** the head moved, so the prior two
clean reviews (Opus + Gemini at `cc9b273c`) no longer apply to the exact head. Since
the resolution kept already-reviewed code verbatim and only inherited `origin/main`'s
already-reviewed-and-merged hardening plus a mechanical dedup, I ran a single
fixed-head adversarial review with **MAI-Code-1-Flash**, isolated in a read-only
worktree.

**Result: CLEAN.** MAI independently rebuilt the decompiler, reran the fast suite
(2,907 / 0), and traced the security-sensitive paths: the spoof gate fails closed on
a nil / non-`TypeReference` base handle (an unsigned assembly named
`System.Linq.Expressions` is not trusted), `TypeSpecGuard` bounds
self-referential/over-deep TypeSpecs, the fact is set from the single verified
`trustedPlatform` local, exactly one property remains, and no `PlatformToken` callers
survive.

richlander:
Ready to merge

PR #2930, opened 2026-07-20
Add IL difficulty scoring to rank hard-IL corpus candidates

richlander:
## Why

The authored-source correspondence work has two corpus slices: a **real-world** corpus pinned to the 14 benchmark assemblies (affinity), and a **hard-IL** corpus — an adversarial stress set of the *diabolical* real methods drawn from a much broader assembly pool and ranked by how hard their IL is to raise. This PR adds the **difficulty-ranking substrate** the future hard-IL harvest will select on. Harvest itself is a follow-up PR.

## What

- **New `IlDifficultyScorer` + `IlDifficulty`** (`tools/DecompilerHarness/IlDifficultyScorer.cs`). Computes a per-method structural profile entirely off the shared `ILInspector.Instructions` substrate (IL decode + EH-aware block graph). No inspected-assembly loading, no abstract interpretation — SRM-only, NativeAOT-friendly. Components: IL size, block count, branch count, switch count, EH-region count, EH-nesting depth, rare-opcode count, local count, max stack.
- **Ranking-only composite score.** Size-like axes (`il`/`blocks`/`branches`/`locals`) are square-root damped so a large dispatcher ranks high without its sheer size swamping a small but genuinely nasty body; the per-feature signals the raise most often fails on stay linear and can dominate — EH-nesting depth highest, then rare opcodes, switch tables, EH-region count.
- **EH-nesting depth** = longest chain of strictly-contained regions. Sibling handlers on one `try` share a `try` span and are *not* counted as nesting each other; a `finally` protecting its sibling `catch` does deepen the chain.
- **Decode failure never inflates rank** — falls back to size/local/stack scalars with zero control-flow.
- `RealMethodTarget` carries the profile; `--enumerate-real-methods` now sorts by difficulty descending and prints the full component breakdown. `IlSize` is preserved so the authored-source harvester is unaffected.
- README documents the command, components, weighting rationale, and EH-nesting semantics (markdownlint clean).

## Evidence

**Graded compiled-fixture canary** (Trivial → Branchy → Switchy → UnsafeBlocks → NestedEh), built and enumerated:

\`\`\`text
score=   75.6  Graded::NestedEh      eh=4 ehDepth=4
score=   32.1  Graded::UnsafeBlocks  rare=2 (localloc x2)
score=   26.3  Graded::Switchy       switch=1
score=   16.6  Graded::Branchy       branches=6
score=    5.6  Graded::Trivial       il=4
\`\`\`

Monotonic with intended difficulty. `NestedEh` correctly scores `ehDepth=4` because a `finally` protects its sibling `catch`, so the real ECMA region chain is inner-catch ⊂ inner-finally ⊂ outer-catch ⊂ outer-finally.

**Real-assembly canary** (`ILInspector.Decompiler.dll`, 3362 targets): the hardest methods surface first (`IrImporter::BuildBlock` at 244.9 — 1915 blocks / 214 branches / 7 switches / 121 locals), and the sqrt damping compresses the range (245→134 over the top 6) so nested-EH / unsafe / switch-heavy methods compete with sheer size rather than being buried.

- Harness build: clean (0 warn / 0 err) with `--configfile nuget.config`.
- No product-path change; harness-only (`tools/DecompilerHarness`).

## Adversarial review

Two reviewers (non-Opus) will review isolated worktrees at the pushed head. Attack points: EH-nesting correctness for sibling/nested/filter handlers, rare-opcode set justification, decode-failure handling, score-formula ranking sanity, no product-path or substrate violation. Reconciliation will be posted here. Do not merge without @richlander approval.

### DIFF
--- a/tools/DecompilerHarness/IlDifficultyScorer.cs
+++ b/tools/DecompilerHarness/IlDifficultyScorer.cs
@@ -1,0 +1,159 @@
+using System.Collections.Immutable;
+using System.Reflection.Metadata;
+
+using ILInspector.Instructions;
+
+namespace ILInspector.DecompilerHarness;
+
+/// <summary>
+/// A per-method IL difficulty profile plus a single composite <see cref="Score"/>
+/// used to rank the "diabolical" methods for the hard-IL corpus. Every component
+/// is a structural fact read off the shared <see cref="MethodInstructions"/>
+/// substrate (decode + EH-aware block graph) — no inspected-assembly loading, no
+/// abstract interpretation. Ranking, not the absolute scale, is what matters; the
+/// components are retained so a selection pass can re-weight or filter on any
+/// single axis.
+/// </summary>
+internal sealed record IlDifficulty(
+    int IlSize,
+    int BlockCount,
+    int BranchCount,
+    int SwitchCount,
+    int ExceptionRegionCount,
+    int ExceptionNestingDepth,
+    int RareOpcodeCount,
+    int LocalCount,
+    int MaxStack,
+    double Score)
+{
+    public static readonly IlDifficulty Empty = new(0, 0, 0, 0, 0, 0, 0, 0, 0, 0.0);
+}
+
+/// <summary>
+/// Scores a method body's IL difficulty from the shared instruction substrate.
+/// The composite weighting deliberately biases toward the shapes that stress the
+/// decompiler's raise the most — exception-handler nesting, jump tables, and rare
+/// unsafe/typed-reference lowerings — rather than raw size, so the hard-IL corpus
+/// fills with genuinely hard puzzles instead of merely long trivial methods.
+/// </summary>
+static class IlDifficultyScorer
+{
+    // Genuinely rare / hard-to-raise opcodes: unsafe block memory, stack
+    // allocation, function pointers, typed references, varargs, and other
+    // low-frequency shapes that carry unusual lowerings. Common opcodes with
+    // well-understood raises (calls, field access, arithmetic, ordinary
+    // branches, box/unbox, generic constrained calls) are intentionally absent.
+    static readonly ImmutableHashSet<ILOpCode> RareOpcodes = ImmutableHashSet.Create(
+        ILOpCode.Calli,
+        ILOpCode.Cpblk,
+        ILOpCode.Initblk,
+        ILOpCode.Localloc,
+        ILOpCode.Sizeof,
+        ILOpCode.Mkrefany,
+        ILOpCode.Refanyval,
+        ILOpCode.Refanytype,
+        ILOpCode.Arglist,
+        ILOpCode.Ckfinite,
+        ILOpCode.Jmp,
+        ILOpCode.Cpobj,
+        ILOpCode.Endfilter,
+        ILOpCode.Unaligned,
+        ILOpCode.Tail);
+
+    /// <summary>
+    /// Computes the difficulty profile for a decoded method body. When the body
+    /// could not be decoded, the size/local/stack scalars are still recorded and
+    /// the control-flow components stay zero, so a decode failure never inflates
+    /// a method's rank.
+    /// </summary>
+    public static IlDifficulty Score(MethodInstructions decoded, int ilSize, int localCount, int maxStack)
+    {
+        int branchCount = 0;
+        int switchCount = 0;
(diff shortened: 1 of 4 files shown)


richlander:

Two isolated-worktree reviews at head `9bba1372`:

- **GPT-5.6 Sol** — 2 Medium findings.
- **Gemini 3.1 Pro** — 2 High findings.

The reviewers **converged** on the EH sibling-catch over-count (GPT #1 == Gemini #1). Gemini independently added the filter under-count; GPT independently added the decode-incomplete inflation. All three verified against the real substrate types and **fixed** in `2d281bbb`.

### 1. EH nesting over-counted sibling handlers (GPT #1 / Gemini #1) — FIXED

An outer `try` with N `catch` clauses emits N regions sharing one try span; an inner try strictly inside that span was counted as nested under *every* sibling, inflating depth by N-1 (and `ehDepth` carries the highest weight). `ExceptionNestingDepth` now measures nesting over **distinct try spans**, so sibling handlers collapse to one level.

Evidence: new `MultiCatchSiblings` fixture (outer try + 3 catches around an inner try/catch) reports `ehDepth=2` (was 4); genuinely-deep `NestedEh` stays `ehDepth=4` (finally emits an outer region whose try span protects its sibling catch — a real 4-level chain).

### 2. EH nesting ignored exception-filter blocks (Gemini #2) — FIXED

A try nested inside a filter expression (`FilterStart..FilterEnd`) matched neither the try nor the handler span, under-counting depth. Filter spans are now included as enclosing scopes for filter-kind regions.

Evidence: a single `catch (E) when (...)` fixture still reports `ehDepth=1` (no regression / no crash from the added scope). The try-inside-filter *increase* is correct by construction; I could not assemble Gemini's raw-IL repro locally (no `ilasm` on this machine), so that leg is verified by reasoning over the region model rather than a run.

### 3. Decode failure inflated rank (GPT #2) — FIXED

`MethodInstructions.Decode` signals undecodable IL via `IsComplete == false` **without throwing**, so the existing exception guards missed it and the method scored on raw IL size (GPT: a 25 KB body with `blocks=0` scored 96.6 and ranked first). `Score` now returns a zero score (true size/local/stack scalars preserved) when decode is incomplete.

Evidence: a PE-patched fixture whose largest method's first IL opcode is an invalid `0xFF` now reports `blocks=0 score=0.0` and ranks **last** (was floating on size). `IlSize` stays accurate for the harvester.

### Confirmed non-issues (both reviewers)

Rare-opcode set counted correctly (prefixes like `tail.`/`unaligned.` emitted as independent stream opcodes; `constrained.` deliberately excluded); no NaN/overflow in the sqrt-damped formula for non-negative scalars; `RealMethodTarget.IlSize` extraction unchanged so `AuthoredSourceHarvest` is unaffected; substrate invariants intact (no assembly loading, Roslyn-free, SRM-only).

Re-review of `2d281bbb` requested from both reviewers. Not marking Ready until two CLEAN reviews. Will not merge without maintainer approval.

richlander:
## Adversarial review round 2 — both CLEAN

Re-review of the fix commit `2d281bbb` in isolated worktrees:

- **GPT-5.6 Sol — CLEAN.** Build clean; original sibling repro `ehDepth=2`; invalid-opcode 25,613-byte method `score=0.0`, ranked last, IL size preserved; nested try/catch/finally depth 4; normal filter depth 1; nested-filter probe depth 2; three distinct nested try spans depth 3; a *valid* 25,613-byte method stays complete with score 98.6 (IsComplete guard does not wrongly zero valid large methods); real `ILInspector.Decompiler.dll` 3,362 targets OK; markdownlint + `git diff --check` pass; worktree clean.
- **Gemini 3.1 Pro — CLEAN.** Confirmed sibling dedup collapses identical try spans (`ehDepth=2`); verified distinct-try-span dedup never under-counts a legitimate nested chain (ECMA strict containment ⇒ nested spans never collide in the HashSet); filter spans correctly added without over-counting a normal `catch when` (its try/filter/handler are disjoint); try/catch/finally-in-try/catch/finally still depth 4 (finally's larger try span wraps its sibling catch); IsComplete guard yields score 0 with scalars preserved.

Both reviewers are from a different model family than the author (Opus). No outstanding findings.

**Ready to merge** — pending maintainer approval (@richlander). I will not self-merge.

Issue #2935, opened 2026-07-21
Decompiler: re-compose a spilled fluent call chain (follow-up to #2877)


richlander:
## Follow-up to #2877

#2877 asked for three raising steps on the `NumericAssertionsBase.BePositive` witness:
1. Pattern-guarded `&&`-diamond fold — merged in #2903.
2. Inlining a spilled `default(T)` temp — merged in #2903.
3. Re-composing the fluent call chain (this issue).

The array-allocation/params-array half of step 3 landed separately (folding the
spilled `object[]` allocation-then-fill into an `ArrayLiteral`). What remains is
re-composing the *call chain itself* — collapsing each chained call's spilled
receiver/argument temp back into the next call in the chain, e.g.:

```csharp
// current
AssertionChain S_256 = CurrentAssertionChain;
AssertionChain S_258 = S_256.ForCondition(S_1).BecauseOf(because, becauseArgs);
object[] S_257 = new object[] { Subject };
S_258.FailWith("...", S_257);

// target
CurrentAssertionChain.ForCondition(S_1).BecauseOf(because, becauseArgs)
    .FailWith("...", new object[] { Subject });
```

### Why this needs its own PR

A general `StatementRunInliningPass` (generalizing `ConstructorChainArgumentPass`'s
effect-order-preserving argument-run fold from "the base/this constructor chain
call" to *any* statement sink) was prototyped and does collapse the witness fully,
but it also caused ~22 real test regressions elsewhere (lock-statement receiver
copies, char-element-store slot naming, merged-slot naming, calli/unmanaged
function-pointer preservation tests, fixed-buffer/string-pinning ladder tests,
etc.) — confirmed via a clean-baseline diff, not guesswork. Applying the fold to
*any* statement sink is too broad: it inlines temps that other decompiler
features rely on staying as distinct, separately-named locals.

A narrower, sink-typed version of the fold (e.g. restricted to a sink that is
itself a `Call`/`LoadProperty` on a chained receiver, rather than an arbitrary
statement) is likely the right scope, but needs its own design and regression
pass separate from the array-literal fold.

### Also still open

A params-array-unwrap step (`new object[] { Subject }` → bare `Subject`) needs a
new `MethodRef`-level fact for "this parameter is `params`" threaded from the
Metadata layer (`AttributeReader`) through `IrImporter` — no existing plumbing
for this today. Deferred as a separate concern from the chain-recomposition
itself.

richlander:
[...]

PR #2936, opened 2026-07-21
Raise spilled array-fill run into an array literal (#2877)

richlander:
- Advances #2877
- Changes decompiler `IrPasses.Default` to fold a spilled array-allocation-then-fill run into a single `ArrayLiteral`
- Card revision: `e107469e`

## Change

**Conclusion:** **PASS** — a well-scoped, unit-tested raise that removes a real spilled-array shape with no measured regression (clean-baseline diff over the fast decompiler suite and the whole FluentAssertions assembly's validity docket are identical before/after).

### Original source

```csharp
{
    CurrentAssertionChain

}
```

### Before

```csharp
{
    AssertionChain S_256 = CurrentAssertionChain;
    bool S_1 = Subject is T subject && subject.CompareTo(default(T)) > 0;
    object[] S_257 = new object[1];
    AssertionChain S_258 = S_256.ForCondition(S_1).BecauseOf(because, becauseArgs);
    S_257[0] = Subject;
    S_258.FailWith("Expected {context:value} to be positive{reason}, but found {0}.", S_257);
    return new AndConstraint<TAssertions>((TAssertions)(object)this);
}
```

- Valid: True
- Correct: True

The spilled `new object[1]` allocation plus its trailing `S_257[0] = Subject;` fill store are the compiler's params-array spill; correct but low-quality (the array's construction is split from its content across an intervening chained call).

### After

```csharp
{
    AssertionChain S_256 = CurrentAssertionChain;
    bool S_1 = Subject is T subject && subject.CompareTo(default(T)) > 0;
    AssertionChain S_258 = S_256.ForCondition(S_1).BecauseOf(because, becauseArgs);
    object[] S_257 = new object[] { Subject };
    S_258.FailWith("Expected {context:value} to be positive{reason}, but found {0}.", S_257);
    return new AndConstraint<TAssertions>((TAssertions)(object)this);
}
```

- Valid: True
- Correct: True

The allocation and its fill run collapse into one `ArrayLiteral`, placed at the fill run's position (after `ForCondition`/`BecauseOf`) — the allocation itself is reorder-safe (a fresh, not-yet-observable reference) but the element value is not, since `csc` can store it well past the allocation.

### Fully raised

```csharp
{
    CurrentAssertionChain
        .ForCondition(Subject is T subject && subject.CompareTo(default(T)) > 0)

    return new AndConstraint<TAssertions>((TAssertions)(object)this);
}
```

- Required tracking issue: #2935 — re-composing the fluent call chain itself (collapsing each chained call's spilled receiver/argument temp into the next call), rendering the recomposed chain with one call per line to match the original (a pure printer-formatting concern, tracked in the same issue), plus a separate params-array-unwrap step (`new object[] { Subject }` → bare `Subject`, needing new `MethodRef`-level metadata plumbing)

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `ArrayLiteralFromStoresPassTests`: n/a (new) | `ArrayLiteralFromStoresPassTests`: 9 passed |
| Decompiler fast suite | 2913 total, 54 failed | 2922 total, 54 failed |
| Real witness | `NumericAssertionsBase::BePositive` spilled array | `NumericAssertionsBase::BePositive` array-literal collapsed |

Baseline's 54 failures are unchanged by this PR (same tests, same reasons — confirmed via a diff of the full `[FAIL]` line set between a clean `origin/main` checkout and this PR's head; the only difference is the 9 new passing focused tests this PR adds).

## Decompiler quality

**Conclusion:** **PASS** — no validity-defect-class change across the whole FluentAssertions assembly.

`dotnet run --project tools/DecompilerHarness -c Release -- FluentAssertions.dll --library-report --top-patterns 5` on FluentAssertions 8.10.0 (net6.0) reports the identical validity docket before and after this PR:

| Metric | Baseline | PR |
| --- | ---: | ---: |
| validity: CS0023 | 4 | 4 |
| validity: VLD0001 | 3 | 3 |
| validity: CS0019 | 2 | 2 |
| validity: CS0308 | 1 | 1 |
| validity: CS0413 | 1 | 1 |
| validity: malformed:CS1525 | 1 | 1 |
| Full malformed methods | 1 | 1 |
| Bound Full defect methods | 8 | 8 |

## Scope note (why StatementRunInliningPass is not in this PR)

The remaining #2877 slice — re-composing the fluent call chain itself — was
prototyped as a general `StatementRunInliningPass` generalizing
`ConstructorChainArgumentPass`'s effect-order-preserving argument-run fold from
"the base/this constructor chain call" to *any* statement sink. It fully
collapses the `BePositive` witness to the "Fully raised" shape above, but a
clean-baseline diff showed it also regresses ~22 unrelated tests (lock-statement
receiver-copy preservation, char-element-store slot naming, merged-slot naming,
calli/unmanaged-function-pointer preservation, fixed-buffer/string-pinning
ladder tests) — applying the fold to *any* statement sink is too broad; those
other decompiler features rely on the same temps staying as distinct,
separately-named locals. Filed as #2935 for a narrower, sink-typed design.

## Validation

```bash
dotnet build src/dotnet-inspect -c Release --configfile ./nuget.config --nologo --verbosity quiet
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.ArrayLiteralFromStoresPassTests
```

### DIFF
--- a/docs/templates/decompiler-pr.md
+++ b/docs/templates/decompiler-pr.md
@@ -60,8 +60,13 @@ only after checking with dotnet-inspect (relying on SourceLink).
 
 ### Before
 
+<!--
+Include the method signature line, matching Original source's shape, not just
+the body — a bare body is harder to line up against Original source.
+-->
+
 ```csharp
-// short failing or lower-quality output
+// short failing or lower-quality output, with its method signature
 ```
 
 - Valid: {True/False}
@@ -71,8 +76,10 @@ only after checking with dotnet-inspect (relying on SourceLink).
 
 ### After
 
+<!-- Include the method signature line here too, for the same reason. -->
+
 ```csharp
-// output produced by this PR, including an honest fallback when not fully raised
+// output produced by this PR, including an honest fallback when not fully raised, with its method signature
 ```
 
 - Valid: {True/False}
--- a/src/ILInspector.Decompiler.Tests/ArrayLiteralFromStoresPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ArrayLiteralFromStoresPassTests.cs
@@ -11,6 +11,8 @@ public class ArrayLiteralFromStoresPassTests
 {
     static readonly TypeRef Object = TypeRef.CoreLib("System", "Object");
     static readonly TypeRef ObjectArray = TypeRef.SzArray(Object);
+    static readonly TypeRef StringType = TypeRef.CoreLib("System", "String");
+    static readonly TypeRef StringArray = TypeRef.SzArray(StringType);
     static readonly TypeRef Void = TypeRef.CoreLib("System", "Void");
     static readonly TypeRef Derived = TypeRef.CoreLib("System", "MyType");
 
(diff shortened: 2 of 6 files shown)


richlander:

Two-model adversarial review per repo policy: **GPT-5.6 Sol** and **Gemini 3.1 Pro**, each in an isolated detached-HEAD worktree, re-reviewed after every fix. Final head: `800a43b7`. Both reviewers are now clean.

### Round 1 (base `dc678075`)
- **GPT-5.6 Sol** and **Gemini 3.1 Pro** independently found the same bug: **self-referential fill values** (`tmp[0] = tmp;`) fold incorrectly, since `ArrayLiteral` evaluates all elements before the combined store commits, so a self-read observes the *prior* value instead of the new array reference.
  - Fixed in `918c0868`: reject any fill value whose subtree reads/writes/addresses the place.
- **GPT-5.6 Sol** additionally found: **covariant array type mismatch** — `object[] tmp = new string[1]; tmp[0] = new object();` is valid C# (throws `ArrayTypeMismatchException` at runtime) but folds to invalid C# (`new string[] { new object() }`).
  - Fixed in `918c0868`: added a `PlaceElementType` guard declining the fold when the declared local's array element type differs from the allocated array's element type.
- **GPT-5.6 Sol** also flagged **allocation motion past effectful statements is theoretically observable** (`GC.GetAllocatedBytesForCurrentThread()`, OOM-exception-ordering).
  - **Deliberately not fixed.** This is an inherent, documented design tradeoff of the pass (moving a fresh, not-yet-observable allocation reference is treated as reorder-safe) — it's not measured by the repo's compile-back/fidelity harness, and restricting the fold to allocation-fill adjacency would defeat the PR's purpose (the FluentAssertions witness specifically requires folding across an intervening chained call). GPT-5.6 Sol re-reviewed this rationale and concurred it is non-blocking; Gemini did not flag it.

### Round 2 (fixed head `918c0868`)
- **GPT-5.6 Sol**: confirmed both fixes correct, no new bugs, reaffirmed the GC/OOM finding as non-blocking. Clean verdict.
- **Gemini 3.1 Pro**: confirmed both fixes correct, but on a deeper pass found a new bug: **`IsAddressOrWrite` missed structured post-structuring write/bind nodes** — `DeconstructionTarget` (Kind==Local), `NullCoalescingAssignment`, `IsPattern`, `RecursivePropertyDeclarationPattern`, `ForeachStatement`, `UsingStatement` — each of which can (re)bind an existing local index without going through `StoreLocal`/`StoreStackSlot`/`LoadLocalAddress`. Concrete counter-example: `object[] tmp = new object[1]; (tmp, _) = MethodReturningTuple(); tmp[0] = "A"; Sink(tmp);` — the deconstruction reassignment was invisible to the guard, so the pass could fold onto the wrong array reference.
  - Verified real by reading `IrNodes.cs` directly. Fixed in `cfe97cd0`, with a regression test for the exact counter-example.

### Round 3 (fixed head `cfe97cd0`)
- **GPT-5.6 Sol**: no new issues in the `IsAddressOrWrite` extension. Clean.
- **Gemini 3.1 Pro**: on a third pass, found three more structured binder node types missing from the same guard: `CatchClause.VariableIndex`, `UnionSwitchExpressionArm.LocalIndex`, and `Fixed.LocalIndex`/`Fixed.LocalIsStackSlot`.
  - Verified real by reading `IrNodes.cs` directly. Fixed in `800a43b7`, with a regression test for the catch-clause-rebind counter-example.

### Round 4 (final head `800a43b7`)
- **GPT-5.6 Sol**: no remaining blocking concerns. Clean.
- **Gemini 3.1 Pro**: performed an exhaustive cross-check of every `Index`/`Slot`-bearing property in `IrNodes.cs` against the `IsAddressOrWrite` switch and confirmed full coverage (including confirming argument-index properties are irrelevant since the pass only accepts `StoreLocal`/`StoreStackSlot` as the seed). **Final verdict: clean, ready to merge.**

### Verification after every fix
Full fast decompiler suite (`-trait- "Speed=Slow"`) diffed against a clean `origin/main` baseline after each round — 0 new regressions each time (54 pre-existing failures, identical set). `ArrayLiteralFromStoresPassTests` grew from 9 → 12 → 13 → 14 tests covering each fixed hole. FluentAssertions `BePositive` witness re-verified unaffected after every fix.

Ready to merge.

richlander:
Doc-only follow-up commit `5c774069`: updated the PR description's Before/After/Fully-raised examples to include the method signature (per user feedback, easier to compare against Original source), and updated `docs/templates/decompiler-pr.md` to require this going forward. No product code changed — the reviewed and merge-ready code head remains `800a43b7`; this commit does not require re-review.

Also filed a formatting requirement on the tracked follow-up issue #2935: the eventual chain-recomposition PR must also render the recomposed fluent chain with one call per line (matching the original's style), not just collapse it functionally.

PR #2919, opened 2026-07-20
Expose typed IL body diff outcomes

richlander:
## Summary

IL body comparison now returns one total product-owned outcome: `Unavailable`, `Exact`, `OperandDiff`, or `OpcodeDiff`. Requested comparison exceptions are separately named `IlBodyDiffNormalization`, preserving strict operand-aware comparison by default and keeping Fidelity V1 policy harness-owned.

- aggregate each outcome in `IlAssemblyDiff`
- retain the typed body result through Research and member-scoped implementation diffs
- expose an IL-only `Difference` column in the CLI `Implementation Diff` section
- report outcome counts and examples in `IlDiffHarness` Markdown, TSV, JSONL, and schema-v2 snapshots
- preserve the typed outcome when fallback display evidence supplies rows

This is a follow-up to #2852. It does not add CLI normalization-selection flags or change Fidelity V1 classification policy. Schema-v1 `IlDiffHarness` snapshots are intentionally rejected and must be regenerated rather than treating absent outcome counts as zero.

## Evidence

- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config`
- 21 focused Instructions tests
- 113 Research tests
- 115 CLI `DiffCommandTests`
- 62 focused Decompiler fidelity/RTS tests
- 3 fixed-head `ReturnToSenderEvidenceTests`
- Markdown lint on all changed docs

A real `DiffFixtures.V1` to `DiffFixtures.V2` run reported 78 compared bodies: 33 exact, 33 operand diffs, 12 opcode diffs, 7 unavailable, and 45 changed. Markdown, TSV, JSONL, schema-v2 snapshot emission, and same-input baseline comparison were exercised.

## Adversarial review

Claude Opus 4.8 identified that fallback display projection dropped the typed body result; fixed in `7cd16af4`. Its separate metric-partition concern was investigated and dismissed because `ComparedBodyCount`, pair outcomes, and failure observations are intentionally orthogonal and retain their existing meanings. Gemini 3.1 Pro found no blocking issue in the first pass. Both Claude Opus 4.8 and Gemini 3.1 Pro re-reviewed exact head `7cd16af4` and returned **CLEAN**.

Issue #2893, opened 2026-07-18
Schedule the legacy diff-lane deletions promised by finding-adoption (API/IL/C# triple representations)


richlander:
## Problem

The Finding-spine migration is roughly half-complete, and the halfway state is the most expensive possible state: the three biggest diff domains (API, IL, C#) each keep **three coexisting representations**, with the *legacy* lane driving default rendering. `docs/design/finding-adoption.md` (rule 1, lines 14-25) sanctions dual lanes as a temporary cross-validation period ending in "Delete the legacy representation after this check has run at the required scale" — but nothing schedules that deletion. Total model count is currently higher than pre-spine.

## Evidence

- **API**: every `MetadataFindings.Compare*` internally runs legacy `ApiDiffAnalyzer.Compare` and merges its `ApiChange`s via `TransformPairs` (`MetadataFindings.cs:159,175,203-208,231-236,266`; header comment lines 7-9). `ApiFindingComparison` retains raw `ApiDiff` (`MetadataFindings.cs:640-649`). DiffCommand's default path renders `ApiDiff` directly (`DiffCommand.cs:283,709,1093,1149,1167,1738`).
- **IL**: `ResearchDiff` uses legacy `IlBodyDiff.Compare` as the evidence lane (`ResearchDiff.cs:796`); the Finding lane runs only under explicit descriptor retention (`ResearchDiff.cs:746-756,906,933`) — bodies are diffed **twice** when it is on, with a divergence check (`ResearchDiff.cs:801-811`).
- **C#**: legacy `CSharpBodyDiff.CompareAssemblies` primary (`ResearchDiff.cs:1016`); `CSharpFindings.CompareAssemblies` gated (`ResearchDiff.cs:1059-1062`).
- **`ResearchChange` is a 24-parameter pre-Finding mega-row** (`ResearchChanges.cs:86-114`) flattening every mechanism's native payload into nullable slots, with 11 manual `ValidatePayloadMechanism` guards (132-186) hand-simulating the union discipline Findings provides. Its own doc comment concedes the gap (77-83). `ResearchComparison` carries `ApiDiff` AND `ApiFindingComparison` simultaneously with a `ReferenceEquals` guard (417-438).
[...]

PR #2611, opened 2026-07-10
Unify Research change results on the Finding spine

richlander:
## Summary

- replace Research's parallel diff row/result families with one flat
  `ResearchChange` / `ResearchComparison` model and computed subject grouping
- add Metadata-owned API type/member Finding inspections and comparisons while
  retaining `ApiDiff` compatibility classification
- migrate CLI diff, ImplementationDiff, and ReturnToSender consumers to the
  unified Research result

## Validation

- `dotnet build dotnet-inspect.slnx -c Release --no-restore`
- `dotnet build src/dotnet-inspect -c Release
  -p:DefaultTargetFramework=net10.0 -p:PublishAot=false`
- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release --no-build`
  (417 passed)
- `dotnet run --project src/ILInspector.Research.Tests -c Release --no-build`
  (71 passed)
- `dotnet run --project src/dotnet-inspect.Tests -c Release --no-build -- -class
  '*DiffCommandTests*'` (61 passed)
- `npx markdownlint-cli docs/overview.md docs/architecture.md
  docs/design/implementation-diff.md`

The ReturnToSender class filter has two existing failures under the current
.NET 11 daily SDK; both reproduce unchanged on clean `origin/main`.

Refs #2564.
Refs #2577.

PR #2938, opened 2026-07-21
Port API diff classification onto the Finding lane (#2893 step 2)

richlander:
## Summary

Step 2 toward #2893 (schedule the legacy diff-lane deletions promised by
`finding-adoption.md`), following the detection-only divergence check in #2931.

Adds `ApiFindingClassifier`, which builds the `ApiDiff` shown by `DiffCommand`
from the Finding lane's own type/member `FindingComparison` pairs instead of
`ApiDiffAnalyzer`'s independent type-name/member-signature walk. `MetadataFindings`
is rewired end to end (`CompareApi`, `CompareApiTypes`, `CompareApiMembers`,
`CompareApiType`) to source `ApiDiff` from the new classifier.

Type-level classification reuses `ApiDiffAnalyzer.CompareTypeFacetsOnly`
unchanged (type identity is the same full-name key in both lanes). Member-level
classification is intentionally sourced from the Finding lane's richer
identity/soft-key matching rather than the legacy analyzer's bespoke
signature-then-(Name,Kind) matcher.

## Intentional behavior change: fuzzy-matched member pairs

This is an opt-in behavior change for fuzzy-matched (soft-key) member pairs,
such as an extension method's static-to-instance projection:

- **Default (strict)**: a fuzzy match is not assumed to be the same member; it
  is reported as an unrelated removal + addition -- matching what the legacy
  analyzer's own matcher would report for members with no shared identity key.
- **New `ApiDiffOptions.Normalization = ApiDiffNormalization.NormalizeMemberProjection`**
  (mirrors #2919's `IlBodyDiffNormalization` naming/policy shape): treats the
  projection as the same conceptual member, emitting `MemberSignatureChanged`
  only when the raw signature text differs, plus modifier/attribute checks.

Both the strict default and the opt-in normalization are implemented in this PR.

## Correctness fix surfaced while porting

Identity/fingerprint matching is keyed on `CanonicalSignature`, which omits the
return type (`MemberAnchor.ComputeFingerprint`). Two hard-matched members can
therefore still differ in raw signature text -- e.g. `void M()` -> `int M()` --
a real signature change the identity key itself can't see. `ClassifyMatchedMember`
now explicitly compares raw `Member.Signature` text for matched pairs (hard
identity match or normalized projection) and reports `MemberSignatureChanged`
when they differ, matching legacy's classification for this case.

A second internal fix: extension-method identity producers emit two Finding
atoms for the same physical member (a hard `extension:`-selector atom and a
receiver-projected soft-key atom used for fuzzy matching). Both share the same
`Fingerprint`. The classifier now deduplicates `Added`/`Removed` pairs whose
fingerprint is already covered by a fuzzy `Changed` match, so the same
physical member is never double-reported.

## Scope and follow-up

`ApiDiffAnalyzer`'s own `Compare()` entry point, `BuildTypeLookup`, and the
legacy `CompareMembers` matcher are **unchanged and kept for now** -- they
still back `ApiDiffAnalyzerTests` (33 tests) as regression coverage for the
shared `CompareTypeFacetsOnly`/`CompareMemberModifiers`/`CompareAttributes`
primitives, and nothing in production code calls the legacy matcher anymore.
Deleting that legacy walk (and its dedicated test suite) is a tracked #2893
follow-up, not part of this PR.

## Validation

- `ILInspector.Metadata.Tests`: 620 total, 1 pre-existing unrelated failure
  (`MemberSourceProducer_UsesFirstVisibleDocumentWhenRootIsOmitted`, confirmed
  pre-existing via `git stash` in earlier session work).
- `ApiDiffAnalyzerTests` (33 tests, exercises legacy `Compare()` directly):
  all passing, unchanged -- confirms the reusable primitives still match the
  legacy analyzer's own behavior.
- `ApiFindingClassifierTests` (new, 4 tests): type-facet parity with legacy,
  hard-matched-member parity with legacy, strict-default fuzzy-match behavior,
  and normalized-projection behavior, using real compiled `extensionold`/
  `extensionnew` fixture assemblies.
- `dotnet-inspect.Tests`: `DiffCommandTests` all passing (two tests updated to
  reflect the intentional conservative Added+Removed behavior for
  identity-changing signature diffs, replacing the legacy Name+Kind fallback's
  `MemberSignatureChanged`). Full suite shows only the same pre-existing
  environmental failures (missing fixture DLLs needing a full slnx build,
  CRLF-vs-LF assertions, network source-file URL columns) -- none Diff-related.
- `ILInspector.Research.Tests`: 113/113 passing.
- Real-corpus smoke test via CLI `diff --package`, zero divergence exceptions:
  - `Newtonsoft.Json@12.0.3..13.0.4` (61 breaking, 60 additive across 19 types)
    -- correctly reports nullable-annotation-only return-type changes as
    `MemberSignatureChanged` via the new fix.
  - `System.Text.Json@6.0.0..8.0.0` (7 breaking, 106 additive across 41 types).

### DIFF
--- a/src/ILInspector.Metadata/ApiDiffAnalyzer.cs
+++ b/src/ILInspector.Metadata/ApiDiffAnalyzer.cs
@@ -132,7 +132,25 @@ public record ApiChange(
     ApiChangeCategory Category = ApiChangeCategory.Signature,
     ApiChangeSubject? Subject = null);
 
-public sealed record ApiDiffOptions(ApiDiffScope Scope = ApiDiffScope.Signature)
+/// <summary>
+/// Named, explicitly-requested classification exceptions -- mirrors
+/// <c>IlBodyDiffNormalization</c>'s policy shape. Default (<see cref="None"/>) is strict:
+/// a fuzzy/soft member identity match (e.g. an extension method's static/instance
+/// projection) is treated as unrelated (a breaking removal plus an additive addition),
+/// the same as the legacy analyzer's own member matcher would report. Requesting
+/// <see cref="NormalizeMemberProjection"/> instead treats a fuzzy-matched member pair as
+/// the same member observed under two projections, classifying only its facet deltas.
+/// </summary>
+[Flags]
+public enum ApiDiffNormalization
+{
+    None = 0,
+    NormalizeMemberProjection = 1,
+}
+
+public sealed record ApiDiffOptions(
+    ApiDiffScope Scope = ApiDiffScope.Signature,
+    ApiDiffNormalization Normalization = ApiDiffNormalization.None)
 {
     public static ApiDiffOptions Default { get; } = new();
 }
@@ -299,6 +317,19 @@ public static class ApiDiffAnalyzer
 
     private static List<ApiChange> CompareTypes(ApiType oldType, ApiType newType, ApiDiffOptions options)
     {
+        var changes = CompareTypeFacetsOnly(oldType, newType, options);
+        CompareMembers(oldType, newType, changes, options);
+        return changes;
+    }
+
+    /// <summary>
+    /// Type-facet classification only (kind/sealed/abstract/base type/interfaces/type
+    /// parameters/type attributes) -- no member-level walk. Reused by
+    /// <see cref="ApiFindingClassifier"/>, which sources member-level changes from the
+    /// Finding lane's own member comparison instead of this analyzer's member matcher.
+    /// </summary>
+    internal static List<ApiChange> CompareTypeFacetsOnly(ApiType oldType, ApiType newType, ApiDiffOptions options)
+    {
         List<ApiChange> changes = [];
         if (IncludesSignature(options))
         {
@@ -314,7 +345,6 @@ public static class ApiDiffAnalyzer
                 ChangeKind.TypeAttributeRemoved,
                 $"Type '{oldType.FullName}'",
                 ApiChangeSubject.Type(oldType, newType));
-        CompareMembers(oldType, newType, changes, options);
         return changes;
     }
 
@@ -580,7 +610,7 @@ public static class ApiDiffAnalyzer
         }
     }
 
-    private static void CompareMemberModifiers(ApiType oldType, ApiType newType, ApiMember oldMember, ApiMember newMember, List<ApiChange> changes)
+    internal static void CompareMemberModifiers(ApiType oldType, ApiType newType, ApiMember oldMember, ApiMember newMember, List<ApiChange> changes)
     {
         // Virtual removed
         if (oldMember.IsVirtual && !newMember.IsVirtual)
@@ -635,7 +665,7 @@ public static class ApiDiffAnalyzer
             includeDecodeStatus ? member.SignatureDecodeStatus : null);
     }
 
-    private static void CompareAttributes(
+    internal static void CompareAttributes(
         IEnumerable<string> oldAttributes,
         IEnumerable<string> newAttributes,
         List<ApiChange> changes,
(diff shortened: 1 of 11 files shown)


richlander:
## Adversarial review reconciliation (4 rounds, Gemini 3.1 Pro + GPT-5.6 Sol)

Full-scope round 1 (base d31... -> pre-fix head) found **5 confirmed blocking issues**, independently corroborated across both reviewers:

1. **Dead acceptance threshold** — `ResearchDiff.AddApiDiff` never opted into the fuzzy `ExtensionInstanceMatchTier` threshold, so the framework's intentional default (exact-only) silently applied in production. Fixed by passing `memberAcceptanceThreshold: MetadataFindings.ExtensionInstanceMatchTier.Confidence`.
2. **Interface-member `Added` misclassified as `Additive`** — adding a member to an interface is potentially breaking (implementers must add it too). Fixed via `MemberAddedClassification`.
3. **Compiler-generated members leaking into API diffs** — the classifier didn't filter compiler-generated members the way legacy's `ApiDiffAnalyzer` did. Fixed at the classification layer (`ApiFindingClassifier`), *not* the Finding-lane producer (`MetadataFindings.EnumerateMembers` is an intentionally unfiltered raw census per PR #2611 — first attempt at the producer layer broke that contract and was reverted).
4. **Coarse `Failed`-bail-out** (found independently by both reviewers) — any inspection failure anywhere discarded ALL diagnostics instead of mirroring legacy's exact whole-side skip semantics. Fixed by removing the gate and threading `oldIdentityIncomplete`/`newIdentityIncomplete` through `ClassifyTypes`.
5. **Fingerprint collision** (Gemini) — `MemberAnchor.ComputeFingerprint` truncates SHA256 to 40 bits; Gemini constructed a real collision. Dedup now keys on the full `CanonicalSignature` string instead.

All 5 fixed in `d5692fb0`. **Round 2** (both reviewers, isolated worktrees) confirmed all 5 clean, zero new regressions — but **GPT** surfaced a 6th issue: overloaded indexers (`this[int]` vs `this[string]`) collided to the same canonical signature/identity (`P:Type.Item`, parameters omitted for all property/field/event kinds), a genuine regression vs. legacy (which keys on raw `member.Signature` text). Fixed in `92f47906`.

**Round 3** (focused, on `92f47906`): GPT confirmed clean. **Gemini independently found a critical follow-up regression** in that same fix: `ApiSurface.SignatureModel` is `[JsonIgnore]`, so every JSON-persisted baseline has it absent for every member — a supported, pre-existing tested workflow. The indexer fix's SignatureModel-absent fallback stayed parameterless, so a JSON baseline's indexer signature would diverge from the same indexer read live from an assembly, breaking `FindingComparison` pairing (every indexer would double-report as removed+added when diffing a baseline against a live assembly). Fixed in `cf608363` by adding `ExtractCanonicalIndexerParameterList`, which parses parameter types out of the raw `this[...]` signature text so both paths produce identical canonical signatures. Also removed now-fully-dead property/field/event branches in `GetCanonicalSignature` (unreachable since `TryGetCanonicalSignature` always succeeds for those kinds).

**Round 4** (focused, on `cf608363`): Gemini confirmed its critical finding solidly fixed, zero new issues, "fully ready to merge." GPT raised one more candidate finding (attributed/DateTimeConstant-defaulted indexer parameters allegedly losing signature parity) — verified via direct probe that this exact behavior is **pre-existing and identical for ordinary methods** (untouched shared `AbbreviateSignature`/`ExtractParamType` helpers), not introduced or worsened by this PR; GPT agreed this is out of scope and confirmed merge-readiness.

**Final state**: both models give a clean sign-off on `cf608363` with no outstanding findings. Full local validation throughout: `ILInspector.Metadata.Tests` (626/626, 1 known pre-existing unrelated failure), `dotnet-inspect.Tests` `DiffCommandTests` (115/115), `ILInspector.Research.Tests` (113/113), `ApiDiffAnalyzerTests` (legacy, unchanged, 33/33), plus an end-to-end production probe via `ResearchDiff.CompareAssemblies`.

Ready to merge

PR #2949, opened 2026-07-21
Fix unbox pointer conversion spelling

### LINKED ISSUE #2929: Primary Expression(Unbox) spelling ef (T)x is invalid under Convert(UIntPtr, Unbox) -> (nuint)(ref (int)o) (CS1525)
## Summary

The **primary** `Expression(Unbox)` spelling in `CSharpPrinter` renders `unbox T`
as `ref (T)x` (`src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.cs:2237`).
When that node is wrapped in a `Convert(UIntPtr, …)` (a managed-pointer→integer
conversion, `conv.u` on a `ByRef`), the output is:

```text
Full: return (nuint)(ref (int)o);
```

which is CS1525-invalid (a `ref` value is not a valid operand of a cast).

## Context

Surfaced as a non-blocking sibling observation during the adversarial review of
PR #2925. That PR fixes the `Deref(Unbox)` position by spelling the box **place**
as `System.Runtime.CompilerServices.Unsafe.Unbox<T>(o)` (a `ref T` intrinsic),
which is valid and faithful for reads, ref-places, and writes.

This sibling is **not** covered by that rule, and must not be routed through it:

- `Convert(UIntPtr, Unbox)` is a managed-pointer→integer conversion — it converts
  the *pointer* (the managed reference into the box) to `nuint`.
- `Unsafe.Unbox<T>(o)` reads the *value*; `(nuint)Unsafe.Unbox<int>(o)` would
  convert the boxed int value, not the pointer — semantically wrong.
- Do **not** route through `&(T)o` either; that spelling also fails to compile
  (per the reviewer's guidance).

## What a fix needs

A verified valid and semantically faithful C# spelling for "the managed pointer
into the box, converted to a native integer," or an honest fidelity degrade when
no such spelling exists. Include compile evidence (real `csc`) for validity and
an argument for faithfulness, plus a regression test that compiles the rendered
body (mirroring the `AssertCompiles` gate used in
`PrinterPrecedenceTests.Deref_RefConditionalWithUnboxArm_SpellsUnsafeUnbox`).

## Pointers

- Primary spelling: `src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.cs:2237`
- Deref fix (reference for the place spelling): `Deref` `Unbox` case +
  `UnsafeUnboxText` helper in the same file (PR #2925).

richlander:
Fixes #2929

- Changes native-integer conversion of `Unbox` from invalid `ref`-cast syntax
  to a faithful pointer-to-box spelling.
- Card revision: `1c09592e0f912e4b8b6739ec518a6839d47f605f`

## Change

**Conclusion:** **PASS** - the new spelling compiles, preserves the managed
reference's address rather than reading the boxed value, and is limited to
native-integer conversions whose operand is `Unbox`.

### Original source

There is no authoritative C# source for this compiler-independent IL seam. The
original artifact is the `unbox` followed by `conv.u` shape from #2929:

```il
.method static native unsigned int M(object o) cil managed
{
    ldarg.0
    unbox int32
    conv.u
    ret
}
```

### Before

```csharp
static unsafe nuint M(object o)
{
    return (nuint)(ref (int)o);
}
```

- Valid: False
- Correct: False

The cast operand is syntactically invalid (CS1525), so the decompilation cannot
represent the original `unbox int; conv.u` behavior.

### After

```csharp
static unsafe nuint M(object o)
{
    return (nuint)System.Runtime.CompilerServices.Unsafe.AsPointer(
        ref System.Runtime.CompilerServices.Unsafe.Unbox<int>(o));
}
```

- Valid: True
- Correct: True

`Unsafe.Unbox<int>` returns the managed reference into the boxed payload;
`Unsafe.AsPointer` preserves that address, and the final cast models `conv.u`.
The same typed discriminator also projects the required `unsafe` body modifier.

### Fully raised

## Evidence

Baseline: `45585f9c`. Head: `1c09592e`.

| Check | Baseline | Head |
| --- | --- | --- |
| Focused regression | Direct reproduction emits invalid `(nuint)(ref (int)o)` | 2 passed |
| Decompiler fast suite | 2,939 passed | 2,941 passed |
| Render A/B, System.Private.CoreLib | 41,952-method snapshot | 0 changed, 0 added, 0 removed |
| Real compiler witness | Invalid spelling reports CS1525 | File-based app compiles, runs, and prints `True` |

The target IL shape is not present in the 41,952-method CoreLib render corpus,
so the hand-built IR regression pins the exact `Convert(UIntPtr, Unbox)` seam
and compiles its rendered body with Roslyn. A close negative confirms an
ordinary `uint`-to-`nuint` conversion remains a safe cast and does not gain
`Unsafe` calls or an unsafe-body requirement.

The full decompiler suite completed 3,125 tests with two failures unrelated to
this change:

- `LadderIteratorGateTests.IteratorFixture_RecoveredRowsCompileBackOpcodeExact`
  expects `Exact` and receives `OperandDiff`.
  has a pinned census mismatch.

Both exact failures reproduce unchanged on clean baseline `45585f9c`; all other
full-suite tests passed.

## Decompiler quality

**Conclusion:** **PASS** - render A/B reports no changed methods in the
41,952-method CoreLib population, while the exact previously uncovered IR seam
is compiler-gated by the focused regression.

## Validation

```bash
dotnet build dotnet-inspect.slnx -c Release --no-restore
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -method "*ConvertNativeUInt_*"
dotnet run --project tools/DecompilerHarness -c Release --no-build -- --render-ab <baseline.json> --workers 2
dotnet run /tmp/check.cs
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/PrinterPrecedenceTests.cs
+++ b/src/ILInspector.Decompiler.Tests/PrinterPrecedenceTests.cs
@@ -13,6 +13,8 @@ public class PrinterPrecedenceTests
     static readonly TypeRef s_bool = TypeRef.CoreLib("System", "Boolean");
     static readonly TypeRef s_int = TypeRef.CoreLib("System", "Int32");
     static readonly TypeRef s_uint = TypeRef.CoreLib("System", "UInt32");
+    static readonly TypeRef s_nuint = TypeRef.CoreLib("System", "UIntPtr");
+    static readonly TypeRef s_object = TypeRef.CoreLib("System", "Object");
     static readonly TypeRef s_string = TypeRef.CoreLib("System", "String");
 
     static IrFunction Raised(string methodName)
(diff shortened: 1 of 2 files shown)


richlander:

Reviewed head: `1c09592e0f912e4b8b6739ec518a6839d47f605f`

- **Claude Opus 4.8:** CLEAN. Verified address-versus-value semantics,
  null/wrong-type behavior, direct-operand discriminator precision, unsafe
  projection, checked/unchecked wrapping, compiler validity, and the close
  negative. No findings.
- **Gemini 3.1 Pro:** CLEAN. Independently verified the exact base-to-head
  diff, `Unsafe.Unbox<T>` + `Unsafe.AsPointer` fidelity, precedence and type
  binding, native-integer scoping, unsafe-context propagation, product
  constraints, and both focused tests. No findings.

No review findings required resolution or dismissal. Both reviewers examined
the same pushed fixed head in separate linked worktrees. CI is green.

Ready to merge

Issue #2945, opened 2026-07-21
Decompiler: raise capturing lambdas over <>c__DisplayClass delegate creations (EVIL CS1001 cluster)


richlander:
## Summary

The first full **EVIL** corpus benchmark run (12,000 difficulty-ranked hard-IL
methods, all 26 source-resolvable assemblies matched) surfaces a single dominant
genuine-decompiler defect family: **capturing lambdas are not raised when their
closure is a `<>c__DisplayClassN_M` instance referenced through a
`new Func<…>(local.<Owner>b__N)` delegate creation.** The raiser leaves the raw
compiler-generated method reference and the printer emits the angle-bracketed
`<Owner>b__N` name verbatim, producing invalid C# (`CS1001: Identifier
expected`). This is the largest single reason bucket in the run (780 CS1001 rows).

## EVIL benchmark measurement

Full honest run (`--benchmark-authored-corpus evil/corpus.jsonl`, 26/26
assemblies, 12,000 targets evaluated):

| Bucket | Count |
| --- | --- |
| Correct (valid, matches authored) | 1,501 |
| Valid but differs | 5,290 (frontier IL-exact 3,097 / IL-diff 2,181) |
| Invalid (does not round-trip) | 5,209 |

Of the 5,209 Invalid, ~3,163 are reference/accessibility codes
(`CS0234`/`CS0246`/`CS1069`/`CS0122`…) attributable to compile-back **shell
reference gaps** on the broad third-party dependency surface (the decompiler
emits the correct type; the RTS shell just lacks the reference), **not**
decompiler-emission bugs. The genuine decompiler-invalid signal is ~971
syntax-malformation rows, led by **`CS1001` = 780 ("closure-stalled")**.

## Root cause (reproduced)

`System.CommandLine.Command.SetAction#0` decompiles to:

```csharp
___c__DisplayClass30_0 V_0 = new ___c__DisplayClass30_0();
V_0.action = action;
if (V_0.action is null)
    throw new ArgumentNullException("action");
Action = new AnonymousSynchronousCommandLineAction(new Func<ParseResult, int>(V_0.<SetAction>b__0));
```

`V_0.<SetAction>b__0` is a raw reference to the display-class instance method,
emitted with the illegal `<…>` identifier → `CS1001`. Note the display-class
*type* is sanitized (`<>c__DisplayClass30_0` → `___c__DisplayClass30_0`) but the
*method* reference is not — and even if it were sanitized, it would reference a
[...]

PR #2950, opened 2026-07-21
Recompose spilled fluent call chain and wrap long chains (#2935)

richlander:
- Fixes/advances #2935
- Changes: recomposes a spilled fluent call chain into a single chain and prints one call per line when the chain is long
- Card revision: `2479258a` (merged `origin/main`, then a fold soundness fix from adversarial re-review; pass/printer logic otherwise unchanged from the reviewed `640605c3`)

## Change

**Conclusion:** **PASS** — a narrow, sink-typed recomposition pass restores the natural fluent-chain shape (reusing ConstructorChainArgumentPass's proven effect-order fold), and a whitespace-only printer break makes long chains readable; corpus quick gate shows 0 pass bugs and two small residue improvements.

### Original source

Hand-authored witness modeling the FluentAssertions `NumericAssertions.BePositive` chain shape (a `CurrentAssertionChain` property-getter receiver that spills). No SourceLink is involved; this is the authoritative source for the fixture.

```csharp
public static void BePositive(int subject, string because, object[] becauseArgs)
{
    AssertionChain.CurrentAssertionChain
        .ForCondition(subject > 0)
        .FailWith("Expected value to be positive{reason}, but found {0}.", new object[] { subject });
}
```

### Before

```csharp
public static void BePositive(int subject, string because, object[] becauseArgs)
{
    AssertionChain S_257 = AssertionChain.CurrentAssertionChain.ForCondition(subject > 0).BecauseOf(because, becauseArgs);
    object[] S_256 = new object[] { subject };
    S_257.FailWith("Expected value to be positive{reason}, but found {0}.", S_256);
}
```

- Valid: True
- Correct: True

The chain link that ends in `BecauseOf(...)` spills into `S_257` because its
receiver is an effectful property getter/allocation that ExpressionInliningPass
declines to move. The result reads as three statements instead of one chain.

### After

```csharp
public static void BePositive(int subject, string because, object[] becauseArgs)
{
    AssertionChain.CurrentAssertionChain
        .ForCondition(subject > 0)
        .FailWith("Expected value to be positive{reason}, but found {0}.", new object[] { subject });
}
```

- Valid: True
- Correct: True

Matches the authoritative original source. Formatting-only witness (`LongB.TwoLong`),
a single long expression-bodied chain at base, likewise breaks one-call-per-line
(and becomes a block body automatically); short chains such as `Builder.ShortStmt`
stay inline.

### Fully raised

The After decompilation is in the fully raised state. Fully Raised semantics are unchanged by this PR.

## Merge integration

This branch merged `origin/main` (merge commit `9293e689`), which advanced past the
review base with #2925 (spell Unbox as `ref Unsafe.Unbox<T>` across the C# printer
substrates) and #2945 (raise capturing lambdas). Those changes touch the same printer
files as this PR (`CSharpPrinter.cs`, `CSharpPrinter.Members.cs`); the merge auto-resolved
textually. Both sides were verified intact: the fluent-chain methods
(`TryAppendFluentChain`, `TryFluentChainStatement`, `IsPlainInstanceChainSegment`,
`FluentChainLines`) and main's unbox/lambda changes all build and pass. One semantic merge
artifact — a duplicate `s_object` field in `PrinterPrecedenceTests.cs` — was removed in
`47a1bfe1`. The dynamic-site inventory fingerprint was merge-resolved to combine both
new compile-back sites (#2925 `UnboxValueReadPassTests.cs` + this PR's
`FluentChainFormattingTests.cs`): **31 files / 38 sites**.

An adversarial re-review of the merged head found that main's lambda-raising made the
shared `SpilledReceiverFold` unsound for a spill whose one load is a captured local
inside a lambda argument (an eager `tmp = Effect()` would defer into `() => Effect()`).
Commit `2479258a` guards `SpillLoadInside` to decline any load reached by crossing a
`Lambda`/`LocalFunctionStatement` boundary, with a causation-proving regression; both
fold consumers route through this gate. See the review reconciliation comments for detail.

## Evidence

| Check | Baseline (65b114aa) | Fixed head (2479258a) |
| --- | --- | --- |
| Focused (`FluentChainRecompositionPassTests`) | n/a (new) | 4 passed |
| Focused (`FluentChainFormattingTests`) | n/a (new) | 4 passed |
| Focused (`DynamicCompilationSiteInventoryTests`) | 4 passed | 4 passed |
| Focused (`ValidityCoverageReportingTests`) | fail (stale pin) | pass (pin refreshed) |
| Decompiler fast suite (`-notrait Speed=Slow`) | 2939, 0 failed | 2967, 0 failed |
| Real witness (`NumericAssertions::BePositive`) | spilled 3-statement | recomposed fluent chain |
| Real witness (`LongB::TwoLong`) | single long line | one call per line |

### Slow suite and pinned self-decompilation census

The decompiler slow suite (`Speed=Slow`, ~4.3h) completed at the pre-merge head
(184 tests, 2 failed) and is re-running at the merged head; a follow-up comment
will report the merged-head result. The two slow failures are both accounted for:

- `LadderIteratorGateTests.IteratorFixture_RecoveredRowsCompileBackOpcodeExact`
  (Exact→OperandDiff) is pre-existing on clean `origin/main` — the fixture has no
  fluent chains, and disabling this PR's pass does not change it. Tracked by **#2958**.
  pins the exact set of decompiler methods whose **self-decompilation** hits CS0161
  (a rendering limitation, not a correctness defect). The pin was **stale on base**
  (`TupleSwitchExpressionPass::TryNumericValue` and
  `YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition` had drifted in
  before this branch), and this PR adds one idiomatic member
  (`FluentChainRecompositionPass::SinkStatement`, a switch expression). The Release pin
  was refreshed to the accurate **13-entry** live population (verified by dumping the
  live census). The pre-existing drift is tracked by **#2959**. This is a test-only
  census refresh; it changes no product opcode or validity behavior (the printer break
  is whitespace-only and the pass only re-nests existing calls).

## Decompiler quality

**Conclusion:** **PASS** — 0 pass bugs; conditional-branch residue and forward-merge stops each improve by one method; no regressions.

### PR quick gate

Run: PR quick corpus, hash-stable 100 methods per assembly; 15 assemblies, 1,500 methods; validity/fidelity not run in the quick card (changed-method fidelity relied upon).

| --- | ---: | ---: | ---: |
| Detected lowering residue (-) | 16.73% | 16.73% | 0 pp |
| Conditional-branch residue (-) | 3.07% | 3.00% | -0.07 pp |
| Forward-merge stops (-) | 2.13% | 2.07% | -0.06 pp |
| Pass bugs (-) | 0 | 0 | 0 |

> **Conclusion:** **PASS** — corpus sensor matched baseline tolerances (regression verdict: PASS).

## Validation

```bash
  -class "ILInspector.Decompiler.Tests.FluentChainRecompositionPassTests" \
  -class "ILInspector.Decompiler.Tests.FluentChainFormattingTests" \
  -class "ILInspector.Decompiler.Tests.DynamicCompilationSiteInventoryTests" \
  -class "ILInspector.Decompiler.Tests.ValidityCoverageReportingTests"
dotnet run --project tools/DecompilerHarness -c Release --no-build -- --fixture-source-inventory
git diff --check
```

Adversarial review was clean from two models (GPT + Gemini) at `640605c3`; a fresh
two-model re-review at the merged head `47a1bfe1` (merge integration + census-pin
honesty) will be reconciled in a comment before this PR is marked ready.


richlander:
## Adversarial review — both CLEAN at head `640605c3`

Two independent reviewers (different model families, isolated read-only worktrees at the exact head, scratch under `/tmp`, no `git reset`/`add`/`commit`) attacked the effect-order soundness of the receiver fold, the whitespace-only splice, the `SpilledReceiverFold` extraction, the non-chain-sink exclusion, and Fully-Raised invariance.

### GPT (`gpt-5.6-sol`) — Verdict: CLEAN

- Effect order and spill guards: holds
- Non-chain sink exclusion (lock / calli / pins): holds
- Constructor-pass extraction: behavior-preserving
- Printer token identity: whitespace-only
- Fully-Raised classification: unchanged
- Real runs: build 0/0; focused 4+4+4+11 passed; fast suite 2,947 passed / 0 failed; copied-lock-receiver CLI witness intact; tracked files unchanged.

### Gemini (`gemini-3.1-pro-preview`) — Verdict: CLEAN

- Effect order: traced an order-sensitive sibling between two spills; `RunPreservesEffectOrder` correctly flags the barrier and the fold DECLINES.
- Non-chain sinks: exact `Call(HasThis)` match structurally excludes lock scopes, `StoreElement`, `CallIndirect`, and `ref`.
- Extraction fidelity: diffed against `65b114aa`; exactly behavior-preserving; `ConstructorChainArgumentOrderTests` 11/11.
- Splice: constructed a multidimensional-array chain (`Expression` renders `a[1, 2]` but `CallText` renders `a.Get(1, 2)`); the `full.StartsWith(previous)` boundary check makes the breakdown safely abort and fall back to a single line — no dropped/duplicated tokens.
- Fully-Raised: unchanged.
- Real runs: build 0/0; focused 4+4+11+4 passed; fast suite 2,947/2,947.

### Reconciliation

No findings from either reviewer; no fixes were required, so the reviewed head equals the PR head (`640605c3`). Not marking ready until the authoritative slow suite (`Speed=Slow`) confirms no new failures beyond the two pre-existing base failures, and CI is green.

richlander:
## Adversarial re-review at merged head — one blocking finding found and fixed

Re-reviewed the merged head `47a1bfe1` with two models (GPT `gpt-5.6-sol`, Gemini `gemini-3.1-pro-preview`), focused on merge integration and the CS0161 census-pin refresh.

### GPT — BLOCKING (High/High), now fixed

GPT found a real soundness hole in the shared `SpilledReceiverFold`, **newly reachable** because main's lambda-raising (#2945) merged into this branch. Lambda raising substitutes a captured outer local as an in-body load of that local, so a single-use spill `tmp = Effect()` whose one load sits inside a lambda **argument** of the sink was `IsInside` the sink call and got folded — moving an eager, always-run effect into a deferred body:

```csharp
tmp = Effect();
receiver.Use(() => tmp);   //  folded to →  receiver.Use(() => Effect());
```

`InConditionalPosition` did not recognize the `Lambda`/`LocalFunctionStatement` boundary, and the argument-order walk sets the lambda barrier only *after* visiting its children, so the in-body load was accepted.

**Reproduced** (synthetic-IR causation test): with the fold unguarded the spill collapses (`StackStores` 2 → 0) and `Effect()` moves into the deferred lambda.

**Fix** (`2479258a`): guard `SpillLoadInside` — the single choke point where the function-scope usage map is consulted — to decline any load reached only by **crossing a nested-function boundary** between the load and the sink. A load and sink in the *same* nested body (boundary above the call) does not cross, so legitimate same-scope folds are preserved. The fix strictly narrows when the fold fires; it can only degrade to honest un-folded output, never emit reordered C#. Both consumers (fluent-chain recomposition and constructor-chain argument lift) route through this gate. Added regression `SpillCapturedByLambdaArgument_IsNotFolded` proving causation.

### Gemini — CLEAN

Gemini independently verified merge integration (unbox `ref Unsafe.Unbox<T>` receiver safely falls back to single-line; raised-lambda arguments splice whitespace-only without token loss), pin-refresh honesty (all 13 CS0161 entries genuinely omit a return; no returning method hidden), and the original invariants. Gemini's probes targeted the printer splice path rather than the fold's deferred-lambda shape, which is why the two reviews diverged — the value of two independent models.

### Now

- Fixed head: **`2479258a`**. Full `dotnet-inspect.slnx -c Release` build: 0 errors. Fast suite (`-notrait Speed=Slow`): **2967 passed, 0 failed**. Both fold consumers (`FluentChainRecompositionPassTests` + `ConstructorChainArgumentOrderTests`): 16/16.
- A fresh two-model re-review of the fixed head (fix completeness / over-rejection / residual holes) and an authoritative slow-suite run at `2479258a` are in flight; results to follow before this PR is marked ready.

richlander:
## Fixed-head re-review — both models CLEAN

Re-reviewed the fixed head `2479258a` (the fold soundness guard) with two models focused on fix completeness, over-rejection, and residual holes.

### GPT `gpt-5.6-sol` — CLEAN
- **Complete:** parent-chain traversal catches lambdas/local functions at any depth and argument position, for both consumers.
- **Over-rejection:** none; same-scope folds inside lambdas remain allowed.
- **Residual holes:** none — all mutations route through `SpillLoadInside`; receiver-spine checks only select candidates.
- **`CountPlaces` conflation:** cannot produce an unsound fold; cross-scope slot collisions can only conservatively *block* a fold. Fast suite 2967/0.

### Gemini `gemini-3.1-pro-preview` — CLEAN
Independently constructed a real C# fixture (`int x = Effect(); int y = x; Current.Use(() => y);`) that drives `LambdaRaisingPass` to substitute the captured local into the nested body, and confirmed via the built CLI:
- **With the fix:** decompilation declines the fold and emits the honest sequence — `int x = Effect(); Program.Current.Use(() => x);` — no deferred effect.
- **Reverting the fix:** the regression fails `Expected: 2` vs `Actual: 0` stack stores, explicitly proving causation.
- Ordinary recomposition (`Current.Use(Effect())` → single expression) is unaffected — no false positives. Same-scope folds not over-rejected; receiver-spine cannot bypass the guard. Fast suite 2967/0.

### Gate status at `2479258a`
- CI: build-net10, test, csharp-diff-smoke, changes — all green.
- Fast suite: **2967 passed, 0 failed**; both fold consumers 16/16; `git diff --check` clean.
- Authoritative slow suite (`Speed=Slow`) is running at this exact head; a final comment will report it. Expected: only the pre-existing `LadderIteratorGateTests.IteratorFixture_RecoveredRowsCompileBackOpcodeExact` failure (#2958), with the CS0161 census pin now passing.

Not marking ready until the slow suite confirms only that one expected failure.

richlander:
## Slow suite complete at fixed head `2479258a` — Ready to merge

Authoritative decompiler slow suite (`-trait Speed=Slow`) at the fixed head:

```
Total: 184, Errors: 0, Failed: 1, Skipped: 0, Not Run: 0, Time: 16888s (~4.7h)
```

The single failure is exactly the pre-existing, unrelated one:

- `LadderIteratorGateTests.IteratorFixture_RecoveredRowsCompileBackOpcodeExact` (Exact→OperandDiff) — pre-existing on clean `origin/main`; the fixture has no fluent chains and disabling this PR's pass does not change it. Tracked by **#2958**.

Notably, `ValidityCoverageReportingTests.DecompilerAssembly_MissingReturnPopulationIsPinned` now **passes** — the refreshed 13-entry CS0161 self-decompilation census pin holds at the merged+fixed head (pre-existing drift tracked by **#2959**).

### Final gate summary at `2479258a`

| Gate | Result |
| --- | --- |
| `dotnet-inspect.slnx -c Release` | 0 errors |
| Fast suite (`-notrait Speed=Slow`) | 2967 passed, 0 failed |
| Slow suite (`-trait Speed=Slow`) | 184 total, 1 failed (only pre-existing #2958) |
| Fold consumers (`FluentChainRecompositionPassTests` + `ConstructorChainArgumentOrderTests`) | 16/16 |
| CI (build-net10, test, csharp-diff-smoke, changes) | all green |
| Adversarial re-review — GPT `gpt-5.6-sol` | CLEAN |
| Adversarial re-review — Gemini `gemini-3.1-pro-preview` | CLEAN |
| `git diff --check` | clean |

Two independent fixed-head reviews confirmed the fold soundness guard is complete (any lambda/local-function depth and argument position, both consumers), free of over-rejection (same-scope folds preserved), and free of residual holes; Gemini reproduced causation by reverting the fix.

**Ready to merge**

Issue #2955, opened 2026-07-21
EVIL frontier: fold negated comparisons !(a==b) -> a!=b (branch-polarity idiom recovery)


richlander:
## Cluster (EVIL benchmark frontier)

Part of the **`valid_different.semantic_opcode_diff.syntax` frontier** (1,632 rows total). In a probed sample of **343 frontier methods across 8 assemblies**, **~15% (54) classify as branch-polarity / negation-idiom** opcode diffs. Scaled to the full frontier this is ~245 rows.

## Symptom (fidelity + readability peephole gap)

The decompiler renders a negated comparison as `!(a == b)` (and similar) instead of folding to the direct idiom `a != b`. Roslyn then compiles the negated form with **inverted branch polarity + relocated basic blocks**, so the recompiled IL differs (`brtrue` vs `brfalse` and a moved return block) even though the code is valid and semantically identical.

Example (`System.CommandLine.Binding.TypeExtensions::IsEnumerable#0`):

Decompiler emits `return !(type == typeof(string)) && (...)` where the original was `type != typeof(string) && (...)`.

```
ORIG:   ldarg ldtoken call call brfalse ldc.i4 ret ldarg callvirt brtrue ... ldc.i4 ret
RECO:   ldarg ldtoken call call brtrue        ldarg callvirt brtrue ... ldc.i4 ret ldc.i4 ret
```

`brfalse` (original `!=`) becomes `brtrue` (decompiled `!(==)`) and the `return false` block relocates to the method tail.

## Candidate fold set

`!(a == b)` → `a != b`; `!(a != b)` → `a == b`; and the relational De Morgan pairs `!(a < b)` → `a >= b`, `!(a <= b)` → `a > b`, etc. (guarding NaN/float ordering where the polarity is not a pure inverse).

## Reproduction

Reflection probe through `ReturnToSender.CompileBackTargets`; `Result.OriginalOpcodes` vs `Result.RecompiledOpcodes` shows the `brtrue`/`brfalse` swap + block relocation.

## Fix location

Decompiler raise/printer negation-idiom folding (a precedence-aware expression peephole). Improves both fidelity and readability. Related umbrella: #2376 (precedence-aware expression model), #2338 (output-quality backlog).

## Source

EVIL corpus benchmark (`vendor/authored-source-corpus` `evil/corpus.jsonl`). Frontier-cluster discovery; sibling actionable frontier sub-shape: switch/jump-table recovery (separate issue).

---

[...]

PR #2961, opened 2026-07-21
Fold negated equality/inequality operator-spelled calls (#2955)

richlander:
- Fixes/advances #2955
- Changes decompiler folds `LogicalNot` over an operator-spelled `op_Equality`/`op_Inequality` call into a direct `!=`/`==` comparison
- Card revision: `13587a2e`

## Change

**Conclusion:** **PASS** — closes a fidelity+readability gap that also fixes inverted branch polarity on recompile, with a deliberately narrow, verified-safe scope.

### Original source

SourceLink-backed, from [dotnet/command-line-api](https://raw.githubusercontent.com/dotnet/dotnet/f7d90799ce4ef09a0bb257852a57248d2a8fb8dd/src/command-line-api/src/System.CommandLine/Binding/TypeExtensions.cs):

```csharp
internal static bool IsEnumerable(this Type type)
{
    if (type == typeof(string))
    {
        return false;
    }

    return
        type.IsArray
        ||
        typeof(IEnumerable).IsAssignableFrom(type);
}
```

### Before

```csharp
internal static bool IsEnumerable(this System.Type type) => !(type == typeof(string)) && (type.IsArray || typeof(IEnumerable).IsAssignableFrom(type));
```

- Valid: True
- Correct: True

Compiles and behaves correctly, but doesn't match the idiomatic source spelling. `Type` doesn't use the native `ceq` opcode for `==`; it compiles to `call System.Type::op_Equality(Type, Type)`. The existing `LogicalNot{Operand:Comparison}` negation fold only pattern-matches native comparison opcodes (`ceq`/`clt`/`cgt`), so it never applied to this `Call`-based path — falling through to the generic `!{Operand}` fallback and rendering `!(type == typeof(string))` instead of the idiomatic `type != typeof(string)`. Recompiling this shape also produces inverted branch polarity (`brtrue`/`brfalse` swap, relocated basic block) relative to the original IL, even though behavior is unchanged.

### After

```csharp
internal static bool IsEnumerable(this System.Type type) => type != typeof(string) && (type.IsArray || typeof(IEnumerable).IsAssignableFrom(type));
```

- Valid: True
- Correct: True

Matches the idiomatic spelling and recompiles with the original branch polarity.

### Fully raised

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `IrImporterTests`/`MemberBodyProducerUnionTests` (new/updated cases): 6 passed | 6 passed |
| Decompiler fast suite | 2958 total, 54 failed | 2961 total, 54 failed (same 54, pre-existing; 3 new tests added) |
| Real witness | `System.CommandLine.Binding.TypeExtensions::IsEnumerable` broken | fixed |

Baseline's 54 failures are unchanged (same tests, same reasons) between base commit `0f38240a` and this PR's head — confirmed via a full diff of `[FAIL]` line sets from `-trait- "Speed=Slow"` runs on both.

## Scope

This PR folds only the equality/inequality operator-call family (`op_Equality`/`op_Inequality`), and only for `MemberIdentity.IsKnownCoreLibraryOperator` (currently `System.String`/`System.Type`). `!(a==b)` and `!(a!=b)` are exact logical negations of each other for any type, including IEEE-754 floats (`NaN==NaN` is false, `NaN!=NaN` is true — no "unordered" edge case exists for equality specifically), but C# only requires `==`/`!=` to be declared as a pair, not that their implementations are each other's logical inverse. An adversarial review round caught that gating on the broader "is this genuinely operator-spelled" check (`IsOperatorCall`) would let the fold silently substitute a call to one user-defined operator method for a call to a different one, which can change behavior for an inconsistent (buggy or adversarial) pair. The fold is therefore restricted to the BCL-guaranteed-consistent `String`/`Type` pair; any other operator-spelled equality call — even one with genuine operator metadata — stays parenthesized (`!(a == b)`), covered by `NotOverEqualityOperatorCall_OnUserDefinedType_DoesNotFold`.

The relational operator-call family (`op_LessThan`, `op_LessThanOrEqual`, `op_GreaterThan`, `op_GreaterThanOrEqual`) is deliberately left un-folded: `!(a<b)` is not always `a>=b` for a user-defined type with partial-order/NaN-like semantics via a custom operator, and a generic `Call` node doesn't carry the native-`Comparison` path's float-based safety guard. A regression test (`NotOverRelationalOperatorCall_Parenthesizes`) locks in the current (safe) un-folded behavior for this family. Generalizing to the relational family, if wanted, should be tracked as a separate follow-up issue with its own safety analysis.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -method "*NotOverEqualityOperatorCall*" -method "*NotOverRelationalOperatorCall*"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/IrImporterTests.cs
+++ b/src/ILInspector.Decompiler.Tests/IrImporterTests.cs
@@ -5552,12 +5552,12 @@ public class EnumConstantTests
     }
 
     [Fact]
-    public void NotOverOperatorCall_Parenthesizes()
+    public void NotOverEqualityOperatorCall_FoldsToDirectComparison()
     {
-        // !(a != b) — an operator-spelled call renders as a compound `a != b`,
-        // so the enclosing `!` must parenthesize it; a bare `!a != b` binds as
-        // `(!a) != b` (CS0023). The C# compiler folds this source form, so the
-        // node is built directly.
+        // !(a != b) is an exact logical negation of a != b for any type (including
+        // NaN-like partial orders: NaN != NaN is true, so !(NaN != NaN) and
+        // NaN == NaN agree), so the printer folds it directly to `a == b` rather
+        // than parenthesizing the un-folded operator-spelled call (#2955).
         var type = TypeRef.CoreLib("System", "Type");
         var boolType = TypeRef.CoreLib("System", "Boolean");
         var callee = new MethodRef(type, "op_Inequality", boolType, [type, type], HasThis: false)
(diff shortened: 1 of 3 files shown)


richlander:

Two independent models reviewed this PR in isolated detached-HEAD worktrees.

### Round 1 — head `09c0ce1b`

- **GPT-5.6 Sol**: BLOCKED
- **Gemini 3.1 Pro**: BLOCKED

Both reviewers independently found the same blocking issue, with equivalent counter-examples:

> `InvertedEqualityOperatorCallText` folded any operator-spelled `op_Equality`/`op_Inequality` call gated only by the broad `IsOperatorCall` guard (metadata `specialname` + operator flag). C# requires `==`/`!=` to be declared as a pair, but does **not** require their implementations to be logical inverses of each other. A user-defined type could implement `operator ==`/`operator !=` with inconsistent semantics (e.g. both always returning the same constant, or with different side effects), and the fold would silently substitute a call to one method with a call to a different method — changing observable behavior for that adversarial/buggy pair. Both reviewers compiled and ran a concrete repro confirming diverging results (`Original=True; Folded=False`).

**Resolution:** Restricted `InvertedEqualityOperatorCallText`'s guard from the broad `IsOperatorCall(call)` to `MemberIdentity.IsKnownCoreLibraryOperator(call.Callee)`, which only recognizes `System.String` and `System.Type`'s `op_Equality`/`op_Inequality` (the BCL guarantees these specific pairs are genuine, side-effect-free logical inverses for every input). Arbitrary user-defined operator pairs — even when properly flagged as real operators — are now excluded from the fold and stay parenthesized (`!(a == b)`), consistent with the repo's "don't infer behavior from spelling" discipline. Added a new regression test, `NotOverEqualityOperatorCall_OnUserDefinedType_DoesNotFold`, that constructs a synthetic user-defined operator pair with genuine operator metadata and asserts the fold does **not** apply. Fixed head: `13587a2e`.

### Round 2 — fixed head `13587a2e423dc0c1b351ac5a4a166c9b882ced55`

- **GPT-5.6 Sol**: PASS WITH NON-BLOCKING NOTES — confirmed the counter-example no longer applies; confirmed `IsKnownCoreLibraryOperator` recognizes only `String`/`Type` equality operators (plus an unrelated `Index.op_Implicit` case that doesn't match this fold's name switch); build succeeded; all `*NotOverEqualityOperatorCall*`/relational-exclusion/union tests passed. Non-blocking note: their local fast-suite run showed 55/2961 failures from an unrelated native-DLL-as-metadata-reference environment quirk (`TypeSourceCheckTests`), with no baseline comparison run on their end.
- **Gemini 3.1 Pro**: PASS — confirmed the fix closes the counter-example, confirmed the `IsKnownCoreLibraryOperator` scope is tight, ran the targeted tests and the full fast suite; noted a few isolated pre-existing local SDK/environment failures unrelated to this PR.

Both reviewers' non-blocking fast-suite notes are local-environment artifacts (native DLL metadata reads under the preview SDK), not regressions. I independently confirmed on this session's machine that the fast suite (`-trait- "Speed=Slow"`) shows the identical 54 pre-existing failures on both the pre-#2955 baseline (`0f38240a`) and this PR's fixed head — zero regressions.

Both reviews are clean. **Ready to merge.**

richlander:
Ready to merge

PR #2965, opened 2026-07-21
Fix attributed-parameter identity leak in raw-signature fallback (#2940)

### LINKED ISSUE #2940: ApiMemberIdentity raw-signature fallback leaks attribute text for DateTimeConstant-defaulted parameters
## Summary

`ApiMemberIdentity`'s raw-signature-parsing fallback (used when `ApiSurface.SignatureModel` is absent, e.g. after JSON round-tripping a persisted `ApiSurface`) mis-parses parameters with a `[DateTimeConstant]`-optional default, leaking attribute text into the canonical signature. The live (`SignatureModel`-driven) path never has this problem, since it uses the clean `ApiParameter.TypeWithModifier`/`ParameterTypesSummary` spelling.

This means a JSON-persisted `ApiSurface` baseline and the same assembly read live can produce **different canonical signatures / fingerprints** for any method (or, since PR #2938, indexer) with such a parameter -- breaking `FindingComparison` pairing between the two (the same failure mode fixed for indexers in PR #2938, just gated behind a much rarer trigger).

## Repro

Constructing `ApiMember` objects directly (raw `Signature` text matching `ApiSurfaceExtractor.FormatParameter`'s `DateTimeConstantDefault` format) and calling the public `ApiMemberIdentity.GetCanonicalSignature`:

```
Method (raw-signature fallback): M:N.C.M([System.Runtime.InteropServices.Optional,System.Runtime.CompilerServices.DateTimeConstant(630822816000000000)] System.DateTime)
Indexer (raw-signature fallback): P:N.C.Item([System.Runtime.InteropServices.Optional,System.Runtime.CompilerServices.DateTimeConstant(630822816000000000)] System.DateTime)
```

Both leak the full attribute list into the canonical signature. The equivalent live-assembly (`SignatureModel`-driven) path would produce e.g. `M:N.C.M(System.DateTime)` with no attribute text.

## Root cause

- `ApiSurfaceExtractor.FormatParameter` (`src/ILInspector.Metadata/ApiSurfaceExtractor.cs:1218-1238`) renders a `DateTimeConstant`-defaulted parameter as `[System.Runtime.InteropServices.OptionalAttribute, System.Runtime.CompilerServices.DateTimeConstant(<ticks>)] <type> <name>` in the raw `Signature` text.
- `ApiMemberIdentity.AbbreviateSignature`'s `parenStart = signature.IndexOf('(')` (`src/ILInspector.Metadata/ApiMemberIdentity.cs`, near `ExtractCanonicalParameterList`) finds the attribute's own `DateTimeConstant(...)` paren rather than the member's parameter-list paren, corrupting prefix/suffix computation.
- `ExtractParamType` strips trailing `= default` and trims at the last top-level space, but never strips a leading `[Attr1, Attr2(...)]` attribute list, so the attribute text survives into the "type".

## Scope / severity

- Discovered as a side-finding during round-4 adversarial review of PR #2938 (indexer canonical-signature fix), which reuses this same fallback machinery for indexers.
- Confirmed **pre-existing** and identical for ordinary methods -- not introduced or worsened by PR #2938 (see `richlander/dotnet-inspect#2938` review reconciliation comment).
- Trigger (`[DateTimeConstant]`-defaulted optional `DateTime` parameters) is rare in practice -- essentially only emitted by VB.NET-authored assemblies for `Optional ByVal x As Date = ...`, or certain interop/COM signatures. Low real-world frequency, but a real correctness gap in shared identity infrastructure (`MemberAnchor` fingerprints, XML doc anchors, source mapping) whenever it's hit.

## Suggested fix

Strip a leading `[...]` attribute list before extracting each fallback parameter's type in `ExtractParamType` (or equivalent), so the raw-signature fallback path matches the live path's attribute-free type spelling for such parameters. Add live-vs-JSON-round-tripped parity coverage for an attributed (DateTimeConstant-defaulted) parameter, mirroring the existing `FallbackCanonicalSignature_*` test pattern in `tests/ILInspector.Metadata.Tests/ApiMemberIdentityTests.cs`.

richlander:
## Problem

`ApiMemberIdentity`'s raw-signature-parsing **fallback** (used when `ApiSurface.SignatureModel` is absent — e.g. after JSON round-tripping a persisted `ApiSurface`) mis-parses parameters whose raw `Signature` text carries a leading attribute list. For a `[DateTimeConstant]`-defaulted optional parameter, `ApiSurfaceExtractor.FormatParameter` renders:

```
[System.Runtime.InteropServices.Optional, System.Runtime.CompilerServices.DateTimeConstant(630822816000000000)] System.DateTime
```

The comma inside `[Optional, DateTimeConstant(...)]` split the parameter list, and the attribute text leaked into the canonical signature — so a JSON-persisted baseline and the same assembly read **live** produced different canonical signatures / fingerprints for the member, breaking `FindingComparison` pairing. This is the same failure mode fixed for indexers in #2938, gated behind a rarer trigger, and is **pre-existing** (not introduced by #2938).

Fixes #2940.

## Observable before / after

The concrete effect is on the member's **canonical signature** — the identity `FindingComparison` uses to pair members across baselines. The live, `SignatureModel`-driven path is stable on both `main` and this PR:

```
M:...AttributedParameterFixture.M(System.DateTime,int)
```

After JSON round-tripping a persisted `ApiSurface` (`SignatureModel` is `[JsonIgnore]`, so the raw-signature **fallback** runs), the same member yields:

**Before (`main`)** — the leading attribute list leaks into the canonical type and the comma inside `[Optional, DateTimeConstant(...)]` inflates the parameter list, so the persisted identity diverges from live and pairing breaks:

```
M:...AttributedParameterFixture.M([System.Runtime.InteropServices.Optional,System.Runtime.CompilerServices.DateTimeConstant(630822816000000000L)] System.DateTime,int)
```

**After (this PR)** — the fallback matches the live path exactly, so the persisted and live fingerprints pair:

```
M:...AttributedParameterFixture.M(System.DateTime,int)
```

Reproduced by forcing the fallback (a persisted `ApiMember` with no `SignatureModel`) over the fixture signature `void M([Optional, DateTimeConstant(...)] System.DateTime when, int count)` on `origin/main` (`f8462478`) vs this head. (`M:...` abbreviates the full type name `ILInspector.Metadata.Tests.ApiMemberIdentityTests` for readability; the two canonicals differ only in the parameter list.)

## Fix (`src/ILInspector.Metadata/ApiMemberIdentity.cs`)

The fallback parses a lossy display string, and that display string is **not C#-only**: F# double-backtick identifiers (`` ``...`` ``) are emitted verbatim by `ApiSurfaceExtractor.FormatParameter` (its `EscapeIdentifier` only escapes C# keywords), so a parameter name can legally contain spaces, `=`, quotes, `[`, `]`, `<`, `>`, `(`, `)`, and commas. Adversarial review confirmed that **every** deviation from `main`'s splitter — independent angle/paren counters, clamping, a global bracket counter, or string/char/default literal tracking — regressed some compiler-emittable name.

So this change takes the tightest possible position and reproduces `main`'s splitter **exactly**, adding only what #2940 requires:

1. **`AbbreviateSignature`** — restores `main`'s exact loop: a single combined `int depth` incremented on `<` and `(`, decremented on `>` and `)`, splitting the parameter list on `,` only at `depth == 0`. No clamping, no independent counters, no string/char/default literal tracking. This preserves `main`'s delimiter cross-cancellation (e.g. an F# name `` ``x<)`` `` emitted as `System.Int32 x<)` keeps `depth` at 0 at the separator comma). The **only** added behavior is a leading-attribute skip: at the start of each parameter, a leading `[...]` attribute list is consumed via a separate `attrBracketDepth` so its inner comma cannot split the list.
2. **`StripLeadingAttributes` + `ExtractSignatureParameterType`** — the extracted parameter type strips a leading `[...]` attribute list (nested-bracket-aware) before trimming at the last top-level space, so leaked attribute text never reaches the canonical type. The fallback and the XML-doc parameter-type path share this helper.
3. **Malformed-signature guard** — `AbbreviateSignature` bails with `if (parenEnd < parenStart + 1) return signature;` instead of slicing, so a persisted signature whose `)` precedes its `(` returns gracefully rather than throwing `ArgumentOutOfRangeException` (the fallback's input domain is untrusted persisted signatures).

For every input that does **not** begin a parameter with a `[` attribute list, this head is byte-identical to `main`. The live (`SignatureModel`-driven) path is unaffected. Multidimensional-array (`int[,]`) and comma-inside-a-string-default (`string s = "a,b"`) fallback parsing remain **pre-existing best-effort limitations shared with `main`**; `SignatureModel` is the robust identity path and the raw-signature fallback exists only for legacy JSON baselines.

### Zero-regression evidence

A side-by-side probe ran `main`'s `AbbreviateSignature` and this head over 31 signature shapes (generics, tuples, arrays incl. `int[,]`, `ref`/`out`/`in`/`params`, string/char/numeric defaults, comma-in-string default, F# `` ``x = '`` `` / `` ``x[`` `` / `` ``x]`` `` / `` ``x<)`` `` / `` ``x>`` `` / `` ``x)`` `` quoted names, single and multiple leading attribute lists). Result: **28 identical to `main`**, **2 differences** = the intended #2940 attribute stripping, **1 case** where `main` throws `ArgumentOutOfRangeException` on reversed parens and this head returns gracefully. No output is worse than `main` on any compiler-emitted shape.

## Tests (`tests/ILInspector.Metadata.Tests/ApiMemberIdentityTests.cs`)

- `FallbackCanonicalSignature_AttributedParameterMatchesLiveAfterJsonRoundTrip` / `...AttributedIndexerMatchesLiveAfterJsonRoundTrip` — live vs JSON-persisted canonical parity with no leaked attribute text (method + indexer).
- `FallbackCanonicalSignature_StripsMultipleLeadingAttributeLists` — multiple leading attribute lists removed.
- `FallbackCanonicalSignature_IgnoresDelimitersInsideStringDefaults` (Theory, incl. `]`, `[`, `<>()`, escaped-quote `"a\"]b"`) and `...InsideCharDefault` (`= ']'`) — delimiters inside a default do not split the list.
- `FallbackCanonicalSignature_DoesNotThrowOnMalformedParentheses` (Theory) — reversed-paren guard.
- `FallbackCanonicalSignature_TreatsQuotesOutsideDefaultsAsOrdinary` (Theory, incl. F# `SetTree<T> t'`, `x' = 5`, `x='`) — quotes/apostrophes in names/types stay ordinary.
- `FallbackCanonicalSignature_DoesNotInterpretQuotesInDefaultRegion` (Theory) — compiler-emittable F# `` ``x = '`` `` / `` ``x = "`` `` names keep both parameters.
- `FallbackCanonicalSignature_TreatsBracketsOutsideAttributesAsOrdinary` (Theory, incl. F# `` ``x[`` ``, `` ``x]`` ``, and `int[]` array types) — brackets outside a leading attribute list stay ordinary.
- `FallbackCanonicalSignature_PreservesCombinedAngleParenDepth` (Theory, F# `` ``x<)`` `` / `` ``x)<`` ``) — combined angle/paren depth cross-cancellation matches `main`, keeping both parameters.

## Validation

| Check | Result |
| --- | --- |
| Build `ILInspector.Metadata.Tests` (`--configfile nuget.config`) | 0 errors |
| `ApiMemberIdentityTests` | 36/36 |
| Full `ILInspector.Metadata.Tests` suite | 648/648 |

Metadata-only change; layer ownership preserved (Metadata owns metadata/signature facts). Rebased on current `main`.

### DIFF
--- a/src/ILInspector.Metadata/ApiMemberIdentity.cs
+++ b/src/ILInspector.Metadata/ApiMemberIdentity.cs
@@ -680,27 +680,27 @@ public static class ApiMemberIdentity
         // identifiers may contain spaces, '=', quotes, brackets, angle brackets,
         // parentheses, and commas), array-rank/type spellings, and default-value
         // literals can all contain characters that look structural, so no parser can
-        // be fully robust here. This fallback therefore matches main's splitter
-        // exactly except for the one construct #2940 introduces: a comma inside a
-        // leading attribute list ("[Optional, DateTimeConstant(ticks)] type name").
+        // be fully robust here. Every deviation from main's splitter risks changing
+        // the canonical signature for some compiler-emittable name (e.g. splitting
+        // main's combined <(...)> depth into independent counters regresses an F#
+        // name like ``x<)`` where '<' and ')' cross-cancel). This fallback therefore
+        // reproduces main's splitter EXACTLY, adding only the one thing #2940 needs:
+        // it skips a leading attribute list ("[Optional, DateTimeConstant(ticks)]
+        // type name") so the comma inside it does not split the parameter list.
         //
-        // Bracket nesting is tracked ONLY inside a leading attribute list, at the
-        // very start of a parameter, so that comma does not split the list. Brackets
-        // appearing later — array types like "int[]" or an F# quoted-identifier name
-        // like ``x[`` — are left untracked, exactly as on main; tracking them would
-        // regress those compiler-emittable names. Angle/paren nesting is tracked as
-        // on main (to keep generic/tuple commas intact), and structural counting is
-        // frozen once a formatter-emitted default begins (the exact " = " separator)
-        // so a default's delimiters never suppress a real comma. String/char literal
-        // tracking is deliberately omitted: it is defeatable by a quote in a name
-        // pairing with a quote in a later parameter's default.
+        // Bracket nesting is tracked ONLY inside that leading attribute list, at the
+        // very start of a parameter. Once the first real (non-space, non-'[') type
+        // character is seen, tracking reverts to main's single combined depth counter
+        // over '<'/'>'/'('/')' , so brackets in array types ("int[]") or in F#
+        // quoted names ("x[") are ordinary text, and generic/tuple commas stay
+        // protected — identical to main. String/char literal and default-value
+        // tracking are deliberately omitted: any such heuristic is defeatable by a
+        // quote or delimiter inside a name, which main treats as ordinary.
         List<string> paramTypes = [];
-        int angleDepth = 0;
-        int parenDepth = 0;
+        int depth = 0;
         int attrBracketDepth = 0;
         int lastSplit = 0;
         bool inLeadingAttributes = true;
-        bool inDefaultValue = false;
         for (int i = 0; i < paramSection.Length; i++)
         {
             char c = paramSection[i];
@@ -714,30 +714,17 @@ public static class ApiMemberIdentity
                     continue; // characters inside the attribute list (incl. commas) are skipped
                 }
                 if (c == ' ') continue; // still in the leading region, between/after attribute lists
-                inLeadingAttributes = false; // first real type character; fall through to normal handling
+                inLeadingAttributes = false; // first real type character; fall through to main's logic
             }
 
-            if (!inDefaultValue && c == '<') angleDepth++;
-            else if (!inDefaultValue && c == '>') { if (angleDepth > 0) angleDepth--; }
-            else if (!inDefaultValue && c == '(') parenDepth++;
-            else if (!inDefaultValue && c == ')') { if (parenDepth > 0) parenDepth--; }
-            else if (angleDepth == 0 && parenDepth == 0)
+            if (c == '<' || c == '(') depth++;
+            else if (c == '>' || c == ')') depth--;
+            else if (c == ',' && depth == 0)
             {
-                if (!inDefaultValue && c == '=' && i > 0 && paramSection[i - 1] == ' '
-                    && i + 1 < paramSection.Length && paramSection[i + 1] == ' ')
-                {
-                    inDefaultValue = true;
-                }
-                else if (c == ',')
-                {
-                    paramTypes.Add(ExtractParamType(paramSection[lastSplit..i].Trim()));
-                    lastSplit = i + 1;
-                    angleDepth = 0;
-                    parenDepth = 0;
-                    attrBracketDepth = 0;
-                    inLeadingAttributes = true;
-                    inDefaultValue = false;
-                }
+                paramTypes.Add(ExtractParamType(paramSection[lastSplit..i].Trim()));
+                lastSplit = i + 1;
+                attrBracketDepth = 0;
+                inLeadingAttributes = true;
             }
         }
         paramTypes.Add(ExtractParamType(paramSection[lastSplit..].Trim()));
--- a/src/ILInspector.Metadata/ApiMemberIdentity.cs
+++ b/src/ILInspector.Metadata/ApiMemberIdentity.cs
@@ -667,7 +667,7 @@ public static class ApiMemberIdentity
             return signature;
 
         int parenEnd = signature.LastIndexOf(')');
-        if (parenEnd < 0)
+        if (parenEnd < parenStart + 1)
             return signature;
 
         string prefix = signature[..(parenStart + 1)];
--- a/src/ILInspector.Metadata/ApiMemberIdentity.cs
+++ b/src/ILInspector.Metadata/ApiMemberIdentity.cs
@@ -707,7 +707,8 @@ public static class ApiMemberIdentity
             else if (c == ']') { if (bracketDepth > 0) bracketDepth--; }
             else if (angleDepth == 0 && parenDepth == 0 && bracketDepth == 0)
             {
-                if (c == '=')
+                if (c == '=' && i > 0 && paramSection[i - 1] == ' '
+                    && i + 1 < paramSection.Length && paramSection[i + 1] == ' ')
                 {
                     inDefaultValue = true;
                 }
--- a/src/ILInspector.Metadata/ApiMemberIdentity.cs
+++ b/src/ILInspector.Metadata/ApiMemberIdentity.cs
@@ -683,6 +683,7 @@ public static class ApiMemberIdentity
         int lastSplit = 0;
         bool inString = false;
         bool inChar = false;
+        bool inDefaultValue = false;
         for (int i = 0; i < paramSection.Length; i++)
         {
             char c = paramSection[i];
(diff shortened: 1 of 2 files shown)


richlander:

Two-model adversarial review per `AGENTS.md` (reviewer roster: Claude Opus / Gemini Pro / GPT / MAI-Code; own model — Claude Opus — excluded). Reviewers: **GPT-5.6** and **Gemini 3.1 Pro**, each isolated in a detached review worktree, scratch under `/tmp` only.

**Both reviews are clean at the exact fixed head `7ee96e51`.**

### What the review changed

The fallback parses a lossy display string whose domain includes **F# double-backtick identifiers** (emitted verbatim by `ApiSurfaceExtractor.FormatParameter`, which only escapes C# keywords). GPT-5.6 found **four** successive real, compiler-emittable regressions across earlier rounds — each caused by a splitter that deviated from `main`'s exact algorithm:

1. Bracket appearing inside a string default (`= "]"`) collapsed the list.
2. Apostrophe / bare-`=` in an F# name defeated the `" = "` default heuristic.
3. A global `bracketDepth` broke `` ``x[`` `` (emitted `System.Int32 x[`) — an unmatched `[` suppressed a real comma.
4. Independent `angleDepth`/`parenDepth` + clamping broke `` ``x<)`` `` (emitted `System.Int32 x<)`), where `main`'s single combined counter relies on `<` and `)` cross-cancelling to keep depth 0 at the separator comma.

Each finding was verified against `origin/main` (which handled the case correctly) before acting. The resolution was to stop improvising heuristics and **reproduce `main`'s exact combined-depth splitter**, adding only the leading-`[...]`-attribute skip that #2940 requires plus a reversed-paren graceful-return guard. For every input that does not begin a parameter with `[`, the head is byte-identical to `main`.

### Final independent verification (head `7ee96e51`)

- **GPT-5.6:** attempted new cross-cancellation / attribute-skip regressions on compiler-emittable input; confirmed the `` ``x<)`` `` regression is closed. Verdict: *"No significant issues found."*
- **Gemini 3.1 Pro:** wrote its own 31-case side-by-side probe (F# names with spaces/`=`/quotes/brackets/misaligned `<>()`, nested generics, `int[,]`, string defaults, attributed optionals) diffing `origin/main` vs head directly. Result: 4 diffs, all intended (3× #2940 attribute stripping, 1× reversed-paren crash→graceful); every other shape matches `main` exactly. Verdict: *"Blocking Findings: None."*

### Evidence

| Check | Result |
| --- | --- |
| `ApiMemberIdentityTests` | 36/36 |
| Full `ILInspector.Metadata.Tests` | 648/648 |
| Side-by-side vs `main` (31 shapes) | 28 identical, 2 intended #2940 strips, 1 crash→graceful |

Note on reviewer signal: across rounds, GPT-5.6 surfaced all four real regressions while Gemini's earlier (narrower, C#-only) corpus reported clean and once mis-dismissed a genuine finding. This round Gemini ran a wide F#-inclusive corpus and independently reached the same clean verdict as GPT on the final main-equivalent design.

richlander:
Ready to merge

PR #2972, opened 2026-07-21
Duplicate scattered shared returns into each guard in structuring (#2959)

### LINKED ISSUE #2959: Stale CS0161 self-decompilation census: TupleSwitch/YieldBreak drifted in unpinned
## Stale CS0161 census in `ValidityCoverageReportingTests.DecompilerAssembly_MissingReturnPopulationIsPinned`

The Release pin in `DecompilerAssembly_MissingReturnPopulationIsPinned` lists the decompiler methods whose **self-decompilation** produces CS0161 ("not all code paths return a value"). It is **stale on `main`**: two methods entered the live CS0161 population but are not in the pinned list, so the test fails on base independent of any feature work.

### Pre-existing drift (present on base `65b114aa`)

- `ILInspector.Decompiler.Pipeline.TupleSwitchExpressionPass::TryNumericValue`
- `ILInspector.Decompiler.Pipeline.YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition`

### Evidence

Dumped the live `ValidityCheck.Evaluate(DecompilerPath)` CS0161 population on base `65b114aa`: **12 methods**, but the Release pin lists **10** — the two above are unpinned. The test therefore fails on base with a first divergence unrelated to any fluent-chain work.

### Resolution taken in PR #2950

PR #2950 refreshes the Release pin to the accurate live population (adding these two pre-existing entries plus that PR's own new `FluentChainRecompositionPass::SinkStatement`) so the gate is green and accurate again. This issue tracks the underlying decompiler-quality question: whether these two pattern-matching helpers should be *rendered* without a missing-return path (removing them from the census) rather than remaining pinned CS0161 debt.

richlander:
- Fixes #2959
- Duplicates a scattered shared `return` into each conditional guard in `StructuringPass` instead of dropping edges, resolving CS0161 self-decompilation defects.
- Card revision: `39bfb88fc754bddb2b6ad47791c38e252d729a6a`

## Change

**Conclusion:** **PASS** — the structuring fix turns six invalid (CS0161) self-decompilations into valid, fully-raised C#, with no corpus regression and the #640 canary preserved.

A shared `return` reached by two or more conditional guards scattered across *different nesting levels* (a return/throw arm sits between the guard blocks) is a join strict nesting cannot express. `StructuringPass` placed the return after one guard and dropped the other edges, so a non-void method fell off its end. The new `IsScatteredDispatch` discriminator fires only when a `return`/`throw` arm is interleaved between predecessor blocks, so the contiguous `&&`/`||` false-exit chain (adjacent guards, no arm between them) still combines into a single condition (the #640 fidelity canary), and `!fallenInto` keeps union-switch defaults untouched.

### Original source

Authoritative source from this repository (`TupleSwitchExpressionPass.TryNumericValue`); the local PDB has no SourceLink to fetch it via `dotnet-inspect`.

```csharp
static bool TryNumericValue(Constant constant, out long value)
{
    switch (constant.Value)
    {
        case sbyte sb: value = sb; return true;
        case byte b: value = b; return true;
        case short s: value = s; return true;
        case ushort us: value = us; return true;
        case int i: value = i; return true;
        case uint u: value = u; return true;
        case long l: value = l; return true;
        case ulong ul when ul <= long.MaxValue: value = (long)ul; return true;
        case char c: value = c; return true;
        default:
            value = 0;
            return false;
    }
}
```

### Before

The failed `when` guard of `case ulong` and the top-level `ul > MaxValue` guard both share the `default` `return false`, at different nesting levels. Structuring kept only one edge; after the final `if (ul <= MaxValue) { … return true; }` there is no `else` and no trailing return, so the method falls off its end.

```csharp
private static bool TryNumericValue(Constant constant, out long value)
{
    // … sbyte/byte/short/ushort/int/uint/long cases identical to After …
    if (V_9 is ulong)
    {
        ul = (ulong)V_9;
    }
    else
    {
        if (V_9 is not char)
        {
            value = (long)0;
            return false;
        }
        c = (char)V_9;
        value = (long)c;
        return true;
    }
    if (ul <= (ulong)9223372036854775807)
    {
        value = (long)ul;
        return true;
    }
}   // <- falls off the end: the ul > MaxValue edge was dropped
```

- Valid: False
- Correct: False

CS0161 "not all code paths return a value".

### After

```csharp
private static bool TryNumericValue(Constant constant, out long value)
{
    // … sbyte/byte/short/ushort/int/uint/long cases identical to Before …
    if (V_9 is ulong)
    {
        ul = (ulong)V_9;
    }
    else
    {
        if (V_9 is not char)
        {
            value = (long)0;
            return false;
        }
        c = (char)V_9;
        value = (long)c;
        return true;
    }
    if (ul > (ulong)9223372036854775807)
    {
        value = (long)0;
        return false;
    }
    value = (long)ul;
    return true;
}
```

- Valid: True
- Correct: True

The dropped edge's `return false` is duplicated into the `ul > MaxValue` guard; all paths return and behavior matches the original switch. This output is valid and behavior-correct but **not fully raised**: the decompiler emits a flat `if (V_9 is …)` type-test chain (with the `ulong` `when` guard split into a nested `if (ul > MaxValue)` early return) rather than reconstructing the original `switch`.

### Fully raised

Not reached by this PR. The intended endpoint reconstructs the original
`switch` (type-pattern `case`s, the `ulong … when` guard, and the `default`
arm) — i.e. the Original source shape:

```csharp
static bool TryNumericValue(Constant constant, out long value)
{
    switch (constant.Value)
    {
        case sbyte sb: value = sb; return true;
        case byte b: value = b; return true;
        case short s: value = s; return true;
        case ushort us: value = us; return true;
        case int i: value = i; return true;
        case uint u: value = u; return true;
        case long l: value = l; return true;
        case ulong ul when ul <= long.MaxValue: value = (long)ul; return true;
        case char c: value = c; return true;
        default:
            value = 0;
            return false;
    }
}
```

- Required tracking issue: #2974 — reconstruct type-pattern `is`-test if-chains back to `switch` statements (incl. `when` guards). This PR's scope is validity (CS0161), not switch reconstruction.

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Census pin (`DecompilerAssembly_MissingReturnPopulationIsPinned`) | 14 methods (fails) | 8 methods (passes) |
| Decompiler fast suite | 3083, 0 failed | 3083, 0 failed |
| Real witness | `TupleSwitchExpressionPass::TryNumericValue` (+5) broken | fixed |
| Validity (`--validity-check`) | — | Full malformed 0; no new CS0161 |

Six self-decompilation CS0161 defects resolved (all the same scattered-return shape): `TupleSwitchExpressionPass::TryNumericValue`, `IsPatternPass::IsPatternLocalNull`, `NullConditionalPass::MemberReceiver`, `UnionSwitchExpressionPass::SameTailNode`, `DynamicCallSitePass::TryGuardCacheLoad`, `FluentChainRecompositionPass::SinkStatement`. `YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition` is a distinct result-temp switch-expression shape not covered here — tracked as follow-up (#2973).

No compiled fixture is included: the defect emerges from pass interaction (a standalone compiled method with the same C# shape does not reproduce it), so the product-owned census pin — which fails on base and passes at head — is the faithful regression guard.

## Decompiler quality

**Conclusion:** **PASS** — pinned-subset gate flat (0 pp), aggregate residue improves, 0 pass bugs.

### PR quick gate

| --- | ---: | ---: | ---: |
| Pass bugs (-) | 0 | 0 | 0 |

> **Conclusion:** **PASS** — corpus sensor matched baseline tolerances.

### Aggregate context

Corpus: 15 assemblies, 1,500 sampled methods.

| --- | ---: | ---: | ---: |
| Detected lowering residue (-) | 251 (16.73%) | 250 (16.67%) | -1 |
| Conditional-branch residue (-) | 46 (3.07%) | 45 (3.00%) | -1 |
| Forward-merge stops (-) | 32 (2.13%) | 30 (2.00%) | -2 |

> **Conclusion:** **PASS** — residue only decreases; no new pass bugs.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ValidityCoverageReportingTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ValidityCoverageReportingTests.cs
@@ -42,13 +42,16 @@ public class ValidityCoverageReportingTests
             .ToArray();
 
         // Release census refreshed against the merged head (verified by dumping
-        // the live CS0161 population). Three entries were added relative to the
-        // prior pin: TupleSwitchExpressionPass::TryNumericValue and
-        // YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition
-        // were pre-existing drift (present on base, unrelated to fluent chains;
-        // see #2959), and FluentChainRecompositionPass::SinkStatement is this
-        // change's new pattern-matching helper that the decompiler renders with
-        // a missing-return path (same idiom as the other census members).
+        // the live CS0161 population). #2959 removed six scattered-return
+        // dispatch defects — the shared return is now duplicated into each guard
+        // instead of dropping an edge: TupleSwitchExpressionPass::TryNumericValue,
+        // IsPatternPass::IsPatternLocalNull, NullConditionalPass::MemberReceiver,
+        // UnionSwitchExpressionPass::SameTailNode,
+        // DynamicCallSitePass::TryGuardCacheLoad, and
+        // FluentChainRecompositionPass::SinkStatement.
+        // YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition is a
+        // distinct result-temp switch-expression shape the fix does not cover and
+        // remains pinned as tracked follow-up.
         string[] expected =
 #if DEBUG
         [
@@ -60,15 +63,9 @@ public class ValidityCoverageReportingTests
             "ILInspector.Decompiler.Pipeline.BooleanFoldingPass::IsNullableCoalesceExpressionContext",
             "ILInspector.Decompiler.Pipeline.CSharpPrinter::ForLoopIncrementText",
             "ILInspector.Decompiler.Pipeline.DeconstructionAssignmentPass::TryMatchTupleSeed",
-            "ILInspector.Decompiler.Pipeline.DynamicCallSitePass::TryGuardCacheLoad",
             "ILInspector.Decompiler.Pipeline.FixedArrayRaising::SameLoadPlace",
-            "ILInspector.Decompiler.Pipeline.FluentChainRecompositionPass::SinkStatement",
             "ILInspector.Decompiler.Pipeline.IndexFromEndPass::LengthReceiver",
             "ILInspector.Decompiler.Pipeline.InlineArrayCollectionPass::PlaceFromAddress",
-            "ILInspector.Decompiler.Pipeline.IsPatternPass::IsPatternLocalNull",
-            "ILInspector.Decompiler.Pipeline.NullConditionalPass::MemberReceiver",
-            "ILInspector.Decompiler.Pipeline.TupleSwitchExpressionPass::TryNumericValue",
-            "ILInspector.Decompiler.Pipeline.UnionSwitchExpressionPass::SameTailNode",
             "ILInspector.Decompiler.Pipeline.YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition",
         ];
 #endif
--- a/src/ILInspector.Decompiler/Pipeline/Passes/StructuringPass.cs
+++ b/src/ILInspector.Decompiler/Pipeline/Passes/StructuringPass.cs
@@ -179,7 +179,7 @@ public sealed class StructuringPass : IIrPass
             {
                 continue;
             }
-            if (IsScatteredDispatch(blocks, conditionalPredecessorIndices[offset], fallenInto, branchTargets))
+            if (IsScatteredDispatch(blocks, conditionalPredecessorIndices[offset], branchTargets))
                 scatteredReturnDispatchTargets.Add(offset);
         }
 
@@ -1122,20 +1122,21 @@ public sealed class StructuringPass : IIrPass
     /// Whether the conditional predecessors of a shared return are scattered
     /// across nesting levels (a <c>return</c>/<c>throw</c> dispatch arm sits
     /// between two of them in block order) rather than forming a single
-    /// contiguous <c>&amp;&amp;</c>/<c>||</c> guard chain. A qualifying arm must be
-    /// reached only by its guard falling through (in <paramref name="fallenInto"/>
-    /// and not a <paramref name="branchTargets"/> target); a terminator reached as
-    /// a branch target is an arm of a reconverging diamond nested inside a
-    /// short-circuit condition (e.g. a <c>throw</c> expression), which combines
-    /// cleanly under one guard and must not force duplication. An interleaved
-    /// dispatch arm means at least one guard reaches the return from a region
-    /// strict nesting would close before the return, so the return must be
-    /// duplicated into each guard.
+    /// contiguous <c>&amp;&amp;</c>/<c>||</c> guard chain. For each interleaved
+    /// terminator, the guard that produced it (found by walking back over the
+    /// arm's plain fall-through statement blocks) must branch forward PAST the
+    /// predecessor span — to the shared return or another downstream terminal —
+    /// which marks a genuine dispatch arm closing a region strict nesting cannot
+    /// keep open, so the return must be duplicated into each guard. A
+    /// <c>throw</c>/<c>return</c> sub-expression nested inside a short-circuit
+    /// condition is instead the fall-through arm of a test that branches forward
+    /// WITHIN the span to continue evaluating the same condition; that arm stays
+    /// combined under one guard (the #640-style fidelity canary), so it is
+    /// excluded.
     /// </summary>
     static bool IsScatteredDispatch(
         IReadOnlyList<Block> blocks,
         List<int> predecessorIndices,
-        HashSet<int> fallenInto,
         HashSet<int> branchTargets)
     {
         int lo = int.MaxValue;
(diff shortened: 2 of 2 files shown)


richlander:
## Adversarial review round 1 — Gemini 3.1 Pro (finding addressed)

**Finding (High): false positive on the #640 short-circuit canary.** Gemini
constructed a witness where a `throw` expression nested inside an `&&` condition
compiles to a terminator block that sits, by IL offset, between the two
false-exit guards of a shared `return`. The original `IsScatteredDispatch` fired
on *any* interleaved `return`/`throw` block, so it unrolled the guard chain and
duplicated the shared return — a structuring-quality regression (both base and
head remain `fidelity: Full`, so no opcode/validity/correctness regression).

**Reproduced and confirmed.** Witness `TestThrowInCond`:

```csharp
public static int TestThrowInCond(int a, int b, int c)
{
    if (a > 0 && (c == 0 ? throw new Exception() : b > 0)) return 1;
    return 2;
}
```

Base nests cleanly under one guard with a single trailing `return 2`; the
pre-fix head duplicated `return 2` into each guard.

**Fix (commit `0b4c7c9d`).** Implements Gemini's suggested remedy: an interleaved
terminator now only counts as scattered dispatch if it is a genuine fall-through
dispatch arm — reached only by its guard falling through (`fallenInto`) and not
itself a branch target. A short-circuit sub-expression `throw`/`return` is a
branch target (an arm of a reconverging diamond), so it is excluded.

Post-fix, `TestThrowInCond` nests cleanly like base; `PlainAnd` (pure `&&`)
stays combined; the genuine scattered target in `TryNumericValue` still fires.

**Validation on fixed head `0b4c7c9d`:**
- Census pin: still 8 (all 6 scattered-return fixes preserved)
- Fast decompiler suite: 3083 pass, 0 fail
- Corpus PR quick gate: PASS — 0 pass bugs; residue 251→250, conditional-branch 46→45, forward-merge stops 32→30
- Validity: `Full malformed: 0`; no new CS0161

Gemini verified attack points 2–5 clean (no side-effect duplication bug, safe
empty-block/predecessor handling, benign set/predicate gate, correct
`IsInlinableTerminator` routing).

Fresh two-model reviews on fixed head `0b4c7c9d` to follow; **not** ready to
merge until both are clean.

richlander:
## Adversarial review round 2 — fixed head `0b4c7c9d` (both models found the same regression)

Both fixed-head reviewers independently found the round-1 predicate
(`fallenInto && !branchTargets`) still false-positives on a `throw` expression
in a **ternary** nested inside a short-circuit `&&` chain:

- **GPT-5.6-sol (High, opcode-fidelity regression).** Witness `a && (b ? c : throw new Exception())`: base recompiles **exact**, round-1 head recompiled to an **opcode diff** (the chain unrolled, `return 2` duplicated).
- **Gemini 3.1 Pro (High, opcode-fidelity regression).** Witness `a && (b ? Method() : throw) && C()`: base recombines the trailing `Method() && C()`; round-1 head unrolled it into disconnected `if (!Method()) return 2; if (!C()) return 2;`.

The `throw` ternary arm is reached by **fall-through** (not a branch target), so
the round-1 proxy missed it. Both reviewers correctly diagnosed that the proxy
cannot prove an else-arm of a scattered guard.

**Fix (commit `f0586db8`).** Replaced the proxy with the structural signal that
actually separates the two shapes: for each interleaved `return`/`throw`, walk
back over the arm's plain fall-through statement blocks to the guard that
produced it, and treat it as scattered dispatch **only when that guard branches
forward past the predecessor span** (to the shared return or another downstream
terminal). A short-circuit test branches forward *within* the span to continue
the same condition, so its arm stays combined. This also fixed a **false
negative** the interim predecessor-only variant would have introduced on
`DynamicCallSitePass::TryGuardCacheLoad`, whose dispatch arm's guard branches to
a *separate* return rather than the shared one.

**Validation on `f0586db8`:**
- Census pin: still **8** (all 6 fixes preserved)
- Fast suite: **3083** pass
- Corpus PR quick gate: **PASS** — 0 pass bugs; residue 251→250, 46→45, 32→30
- Compile-back: `a && (b ? c : throw)` and `... && C()` now **exact** (match base); genuine scattered returns still fire. `TestThrowInCond`'s ternary-throw opcode diff is **pre-existing on base** (verified with the base harness), not a regression.

Gemini also verified: excluding a branch-targeted interleaved arm does **not**
reintroduce CS0161 (StructuringPass falls back to `goto` rather than dropping
the edge); union-switch shared returns register in `unconditionalTargets` and
are correctly bounded.

Fresh round-3 reviews on `f0586db8` to follow; **not** ready to merge until both
are clean.

richlander:
## Adversarial review round 3 — fixed head `f0586db8`

### Gemini 3.1 Pro — CLEAN (ready for merge)

Exhaustively tested all 5 attack points against the new span-based rule with compiled witnesses, real harness fidelity checks, and the census pin. **No actionable issues.**

- **False positive:** tried deeply nested short-circuit chains with embedded terminators (`A || (B && throw)`, `A || (B ? throw : C)`, `A || (b ?? throw) || C`, `b switch { true => C, false => A || throw }`). The rule correctly rejects all: csc targets the internal conditional branch at the *next condition*, so `guard.TargetOffset == blocks[hi].StartOffset` and `> spanEndOffset` is false; the chain stays intact. Fidelity 100% exact or identical to base.
- **False negative / CS0161:** a genuine scattered guard must branch to the shared return, which is physically placed after `hi`, so its target is always `> blocks[hi].StartOffset`. Hand-emitted a pathological IL layout placing the shared return before `hi`; the rule skipped it, but verified base drops it too (pre-existing). Census pin passes at 8.
- **Walk-back:** multi-block arms, merge targets, and empty/unreachable blocks all classify safely (halts at the controlling `ConditionalBranch`, or safe fallback that avoids false positives).
- **Duplication safety:** census 10→8 on base→head (6 defects removed) with no reintroduced errors; non-applicable methods (e.g. `YieldBreakLoopIteratorReconstruction`) left untouched.

### GPT-5.6-sol — one finding, confirmed PRE-EXISTING (not a PR regression) → tracked in #2978

Witness: a nested type-pattern switch `o switch { string s => s.Length switch {...}, int i when i>0 => ..., _ => ... }` decompiles to source that leaves the `i<=0` path without an `out x` assignment (**CS0177**). GPT correctly noted this "also fails on base."

I reproduced it **byte-for-byte identically** on base and head:

| | base `86bd05c4` | head `f0586db8` |
|---|---|---|
| `W::M` | recompile-fail, CS0177 | recompile-fail, CS0177 |
| exact | 2/3 | 2/3 |

So #2972 neither introduces nor regresses it. It is a pre-existing limitation on **nested** dispatch, where the interleaved terminal's guard branches to a *sibling* terminal **inside** the predecessor span. The span rule intentionally does not fire there: firing on within-span forward branches is exactly what produced the round-2 opcode regressions on short-circuit chains that **both** reviewers rejected. Distinguishing nested dispatch from short-circuit sub-expressions at that position needs a stronger signal than target position and is out of scope for this PR (whose goal is the 6 CS0161 census methods; census pin stays 8). Filed as **#2978**.

### Bottom line

Both fixed-head reviews confirm the `f0586db8` **delta introduces no regression**. Gemini is clean; GPT's only finding is pre-existing (base==head) and tracked separately. Round-3 reconciliation complete.

richlander:
### GPT-5.6-sol — clean confirmation (finding withdrawn)

On follow-up with the base==head reproduction, GPT reverified exact head `f0586db8` and stated:

> No significant issues found in the reviewed changes. Reverified exact head `f0586db8`; the PR delta is clean. Census pin passes at 8. The prior #2978 finding is withdrawn.

**#2978** remains filed as a standalone pre-existing limitation (nested type-pattern switch dispatch), unrelated to this PR's delta.

## Ready to merge

Two clean adversarial reviews on the exact fixed head `f0586db8`:
- **Gemini 3.1 Pro** — clean, all 5 attack points verified.
- **GPT-5.6-sol** — PR delta clean, finding withdrawn.

Evidence on `f0586db8`: census pin **8** (10→8, 6 defects removed), fast suite **3083** pass, corpus PR quick gate **PASS** (0 pass bugs; residue 251→250, 46→45, 32→30), both round-2 witnesses now exact.

Issue #2942, opened 2026-07-21
Test suite: Roslyn recompile reference set includes native DLLs from TRUSTED_PLATFORM_ASSEMBLIES -> CS0009 (~51 local failures on the .NET 11 daily SDK)


richlander:
## Summary

Running the `ILInspector.Decompiler.Tests` suite locally on the repo's documented
**.NET 11 daily SDK (Windows)** produces ~51 failures whose root cause is the
in-process Roslyn compile helpers passing **native / unmanaged DLLs** to Roslyn as
metadata references. Every such reference fails with:

```text
CS0009: Metadata file '...\Microsoft.NETCore.App\11.0.0-preview.6.26359.118\System.IO.Compression.Native.dll'
        could not be opened -- PE image doesn't contain managed metadata.
```

CI is **green** (all recent `main` runs succeed), so this is environment-specific
to SDK layouts where `TRUSTED_PLATFORM_ASSEMBLIES` (TPA) enumerates native DLLs.
It makes the suite effectively unrunnable locally on the daily SDK.

Discovered while running the local suite for PR #2925; **unrelated to that PR**.

## Root cause

`RoslynTestReferences.TrustedPlatform` (and the sibling TPA-based helpers) build the
Roslyn reference set by enumerating `TRUSTED_PLATFORM_ASSEMBLIES` and filtering only
by file **extension** (`.dll`), not by "is a managed PE with metadata":

```csharp
// src/ILInspector.Decompiler.Tests/RoslynTestReferences.cs
foreach (string path in (AppContext.GetData("TRUSTED_PLATFORM_ASSEMBLIES") as string ?? "")
    .Split(Path.PathSeparator, StringSplitOptions.RemoveEmptyEntries))
{
    if (!path.EndsWith(".dll", StringComparison.OrdinalIgnoreCase))
        continue;
    ...
    try { builder.Add(MetadataReference.CreateFromFile(path)); }
    catch { }   // CreateFromFile does NOT read metadata, so native DLLs pass here
}
```

`MetadataReference.CreateFromFile` is lazy, so the native DLLs are not rejected at
reference-creation time; the failure surfaces later as a **CS0009 compile
diagnostic**. On this machine TPA includes the full shared-runtime directory
(`coreclr.dll`, `clrjit.dll`, `clrgc.dll`, `hostpolicy.dll`, `mscordaccore*.dll`,
`mscordbi.dll`, `System.IO.Compression.Native.dll`, `msquic.dll`,
[...]

PR #2917, opened 2026-07-20
Preserve authored enclosing context for nested dynamic raises (#2906)

richlander:
- Advances #2906
- Preserves the authored enclosing-type context for a compiler-generated nested body (display-class lambda, iterator/state-machine `MoveNext`) so `DynamicCallSitePass` can raise `Binder.GetMember` scaffolding it previously had to leave explicit.
- Card revision: `76b2b6f6`

## Change

> Should we accept this change?

**Conclusion:** **REVIEW** — the core mechanism (metadata-decoded `TypeRef.EnclosingType`, gated on `[CompilerGenerated]` attribute evidence, never on name-parsing) is sound and covered by real compiled fixtures plus close-negative cases, but this branch is stacked on the still-unreviewed #2915 and the corpus/aggregate evidence sections below have not been run yet.

### Original source

Authored fixture `DynamicMemberContexts.InLambda` — a capturing lambda whose body performs a `dynamic` member access, which the compiler lowers to the display-class method `<>c__DisplayClass4_0.<InLambda>b__0`:

```csharp
public Func<object> InLambda(dynamic value)
{
    return () => value.Length;
}
```

### Before

Compiler-backed fixture (`LadderRung9.DynamicMemberContexts.InLambda`, capturing lambda lowered to `<>c__DisplayClass4_0.<InLambda>b__0`). The binder-context `typeof` in the display-class method is `typeof(DynamicMemberContexts)` — the *authored* enclosing type, not the display-class type itself — so the prior exact-equality check against `function.DeclaringType` declined and the scaffolding stayed explicit:

```csharp
CSharpArgumentInfo[] S_256;
Type S_257;

if (___o__4.___p__0 is null) goto IL_0009;
goto IL_0038;
IL_0009:
S_256 = new CSharpArgumentInfo[1];
S_257 = Type.GetTypeFromHandle(typeof(DynamicMemberContexts));
S_256[0] = CSharpArgumentInfo.Create((CSharpArgumentInfoFlags)0, null);
___o__4.___p__0 = CallSite<Func<CallSite, object, object>>.Create(Binder.GetMember((CSharpBinderFlags)0, "Length", S_257, S_256));
IL_0038:
return ___o__4.___p__0.Target.Invoke(___o__4.___p__0, value);
```

### After

```csharp
return ((dynamic)value).Length;
```

### Fully raised

The After decompilation is in the fully raised state.

## Evidence

| Check | Result |
| --- | --- |
[...]

PR #2980, opened 2026-07-21
Raise dynamic member access inside a generic enclosing type (#2968)

### LINKED ISSUE #2968: Dynamic raises inside a generic enclosing type decline (IsProvenBinderContext exact-equality gap)
## Summary

`DynamicCallSitePass.IsProvenBinderContext` (added in #2917 / advances #2906) bridges a compiler-generated display-class / iterator body to its authored enclosing type via `TypeRef.EnclosingType` and an exact `context.Equals(enclosing)` check. When the authored enclosing type is **generic**, this exact-equality check fails, so the dynamic scaffolding is left explicit in exactly the lambda/iterator scenarios the feature intends to raise.

## Detail

For a capturing lambda or iterator declared inside `GenericHost<T>`:

- The compiler emits the binder context as `typeof(GenericHost<T>)` — a **generic instance** `TypeRef` (`Kind == GenericInstance`, `ElementType = GenericHost\1`, type args `[T]`).
- `declaringType.EnclosingType` decoded from the metadata nested-type chain is the **definition** `GenericHost\1` (`Kind == Definition`).
- `context.Equals(enclosing)` compares a `GenericInstance` against a `Definition` → `Kind` mismatch → `false` → the pass declines and retains `Binder.GetMember` / `CallSite` scaffolding.

Found during adversarial review of #2917 (GPT-5.6 Sol), reproduced with a real net11 fixture containing a capturing lambda and an iterator inside `GenericHost<T>`; output retained the explicit scaffolding instead of `((dynamic)value).Length`.

## Impact

Completeness gap only — **not a regression**. Before #2917 none of these nested-context dynamic sites raised; after #2917 the non-generic enclosing-type cases raise and the generic ones still decline. Output stays valid (fully explicit), just not fully raised.

## Suggested direction

Recognize an enclosing **definition** that the binder context names as its own self-instantiation: a `GenericInstance` whose `ElementType` equals `declaringType.EnclosingType` **and** whose type arguments are exactly that enclosing type's own in-scope generic parameters, in order. Do **not** naively strip arbitrary generic arguments — `GenericHost<int>` and `GenericHost<T>` are distinct contexts and a mismatch may indicate a genuinely different authored type. This likely needs generic-arity/provenance material on the decoded enclosing `TypeRef`.

## Evidence to require on the fix

- Compiled fixtures: capturing lambda and iterator inside a generic type both raise; close-negative where the type args differ from the enclosing type's own parameters still declines.
- No change to the non-generic cases already covered by `DynamicCallSitePassTests`.

richlander:
Closes #2968.

## Problem

`DynamicCallSitePass.IsProvenBinderContext` (added in #2917) bridges a
compiler-generated display-class / iterator body to its authored enclosing type
via `TypeRef.EnclosingType` and an exact `context.Equals(enclosing)` check.
When the authored enclosing type is **generic**, that exact-equality check fails
and the dynamic scaffolding is left explicit in exactly the lambda/iterator
scenarios the feature intends to raise.

For a capturing lambda or iterator inside `Host<T>` (ground-truthed on the
net11 daily compiler via a decode probe):

- the GetMember binder context is `typeof(Host<T>)` — a **GenericInstance**
  `TypeRef` (`ElementType` = the `Host` definition, args `[GenericParameter(0)]`);
- `declaringType.EnclosingType` decoded from the metadata nested-type chain is
  the bare **Definition** `Host`;
- `GenericInstance` != `Definition` → `Equals` is `false` → the pass
  declines and retains `CallSite` / `Binder.GetMember` scaffolding.

Completeness gap only, **not a regression** — the found-during-review follow-up
GPT-5.6 Sol filed against #2917. Output stayed valid (fully explicit), just not
fully raised.

## Fix

Add `IsGenericSelfInstantiation(context, enclosing)`: accept a
`GenericInstance` context whose `ElementType` equals the enclosing definition
and whose arguments are exactly the type's own generic parameters in declaration
order (`GenericParameter` with index `i` at position `i`).
`IsProvenBinderContext` accepts this in addition to the existing exact
`Equals`.

Only that self-instantiation is accepted. A concrete argument
(`typeof(Host<int>)`), a reordered parameter, or a method type parameter
denotes a different accessibility context and still declines — the fix does not
naively strip type arguments.

The issue anticipated this "likely needs generic-arity/provenance material on the
decoded enclosing `TypeRef`". It does not: the `GenericInstance` context's
`ElementType` already pins the enclosing definition's identity and its
arguments already carry `GenericParameter` positions, so the self-instantiation
is recognizable from the already-decoded shapes. No `TypeRef` or decoder
changes were required.

## Example

Fixture `LadderRung9.GenericDynamicMemberContexts<T>.InGenericLambda` — a
capturing lambda inside a generic type, lowered by csc to
`<>c__DisplayClass0_0.<InGenericLambda>b__0`.

### Original source

```csharp
public class GenericDynamicMemberContexts<T>
{
    public Func<object> InGenericLambda(dynamic value)
    {
        return () => value.Length;
    }
}
```

### Before

The display-class method's binder-context `typeof` is
`typeof(GenericDynamicMemberContexts<T>)` — a generic instantiation of the
authored enclosing type — while the declaring type's decoded `EnclosingType` is
the bare definition, so the prior exact-equality check declined and the
scaffolding stayed explicit (real product output):

```csharp
CSharpArgumentInfo[] S_256;
Type S_257;

if (___o__0.___p__0 is null)
{
    S_256 = new CSharpArgumentInfo[1];
    S_257 = typeof(GenericDynamicMemberContexts<T>);
    S_256[0] = CSharpArgumentInfo.Create((CSharpArgumentInfoFlags)0, null);
    ___o__0.___p__0 = CallSite<Func<CallSite, object, object>>.Create(Binder.GetMember((CSharpBinderFlags)0, "Length", S_257, S_256));
}
return ___o__0.___p__0.Target.Invoke(___o__0.___p__0, value);
```

### After

```csharp
return ((dynamic)value).Length;
```

### Fully raised

## Evidence

- New compiled fixture `GenericDynamicMemberContexts<T>` (capturing lambda +
  iterator) in the LadderRung9 fixture assembly.
- Two real-fixture raise tests (`NestedContext_GenericEnclosingLambda_Raises`,
  `NestedContext_GenericEnclosingIterator_Raises`) and three synthetic
  close-negatives (concrete arg / reordered parameter / method parameter all
  decline).
- **Gap proven**: the two raise tests fail on the pre-fix gate (scaffolding
  retained); they pass after the fix. The three negatives pass with or without
  the fix (genuine negatives).
- Focused `DynamicCallSitePassTests` + `LadderRung9GateTests` +
  `TypeRefDecoderRecursionTests` **139/139**. The 105 pre-existing
  DynamicCallSitePass cases are unchanged.
- The 13 subset failures observed locally are the pre-existing #2942 Windows-env
  compile-back `RecompileFail` set — **identical on clean `main`** (same 13
  names, Total 192), i.e. zero new failures.

## Validation complete

- [x] Corpus A/B structural inventory: base `2077bab4` == head `c26a144a`
      **SHA-256 byte-identical** (128234/135595 Full 94.57%; 2 pre-existing
      importer bugs). Zero movement.
- [x] Two-family adversarial review (GPT-5.6 Sol + Gemini 3.1 Pro): **both CLEAN**
      on head `c26a144a`, reconciled in a PR comment.
- [x] CI green: Linux `test` pass, `build-net10` pass, `csharp-diff-smoke`
      pass, `changes` pass (il-diff/markdownlint/pack path-skipped).

### DIFF
--- a/src/ILInspector.Decompiler.Fixtures.LadderRung9/LadderRung9MemberContexts.cs
+++ b/src/ILInspector.Decompiler.Fixtures.LadderRung9/LadderRung9MemberContexts.cs
@@ -62,3 +62,26 @@ public class DynamicMemberContexts
     public object Last => _last;
 }
 
+// Same nested-body dynamic GetMember shapes as DynamicMemberContexts, but
+// authored inside a GENERIC enclosing type. csc emits the GetMember binder
+// context as typeof(GenericDynamicMemberContexts<T>) — a generic instantiation
+// of the enclosing definition with its own type parameter in scope — while the
+// lowered display-class / state-machine declaring type's metadata-decoded
+// EnclosingType is the bare generic definition. The dynamic-callsite raise must
+// recognize that self-instantiation and still raise rather than declining on the
+// GenericInstance-vs-Definition kind mismatch (#2968).
+public class GenericDynamicMemberContexts<T>
+{
+    // Nested lambda body inside a generic enclosing type.
+    public Func<object> InGenericLambda(dynamic value)
+    {
+        return () => value.Length;
+    }
+
+    // Iterator state-machine body inside a generic enclosing type.
+    public IEnumerable<object> InGenericIterator(dynamic value)
+    {
+        yield return value.Length;
+    }
+}
+
--- a/src/ILInspector.Decompiler.Tests/DynamicCallSitePassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/DynamicCallSitePassTests.cs
@@ -1557,6 +1557,126 @@ public class DynamicCallSitePassTests
         Assert.False(RunPass(mutated));
     }
 
+    /// <summary>Imports the InGenericLambda display-class method (a lambda inside a generic enclosing type), running every pass up to (not including) dynamic-callsite so tests can mutate the binder context before proving the pass's own decision.</summary>
+    static IrFunction LoadGenericLambdaDisplayClassFunction()
+    {
+        var path = typeof(LadderRung9.DynamicMemberContexts).Assembly.Location;
+        using var source = MetadataSource.Open(path);
+        IrFunction? found = null;
+        foreach (var (_, methodName, function) in IrImporter.ImportAssembly(source))
+        {
+            if (methodName.Contains("InGenericLambda") && methodName.Contains("b__"))
+            {
+                found = function;
+                break;
+            }
+        }
+        Assert.NotNull(found);
+
+        var context = new PassContext(new Stepper(enabled: false));
+        foreach (var pass in IrPasses.Default)
+        {
+            if (pass.Name == "dynamic-callsite")
+                break;
+            pass.Run(found!, context);
+        }
+        return found!;
+    }
+
+    [Fact]
+    public void NestedContext_GenericEnclosingLambda_Raises()
+    {
+        // A capturing lambda authored inside a generic type Host<T> is lowered
+        // into a display-class method whose GetMember binder context is
+        // typeof(Host<T>) — a GenericInstance of the enclosing definition with
+        // its own type parameter — while the declaring type's decoded
+        // EnclosingType is the bare generic definition. The self-instantiation
+        // bridge (IsGenericSelfInstantiation) proves the context, so the site
+        // raises rather than declining on the GenericInstance-vs-Definition
+        // kind mismatch (#2968).
+        var path = typeof(LadderRung9.DynamicMemberContexts).Assembly.Location;
+        using var source = MetadataSource.Open(path);
+        string? displayClassOutput = null;
+        foreach (var (_, methodName, function) in IrImporter.ImportAssembly(source))
+        {
+            if (methodName.Contains("InGenericLambda") && methodName.Contains("b__"))
+            {
+                displayClassOutput = CSharpPrinter.PrintRaised(function, mr => IrImporter.Import(source, mr)).Output;
+                break;
+            }
+        }
+        Assert.NotNull(displayClassOutput);
+        Assert.Contains("((dynamic)value).Length", displayClassOutput);
+        Assert.DoesNotContain("Binder.GetMember", displayClassOutput);
+    }
+
+    [Fact]
+    public void NestedContext_GenericEnclosingIterator_Raises()
+    {
+        // The iterator variant of the generic-enclosing case: an iterator inside
+        // Host<T> is lowered into a MoveNext method whose binder context is
+        // likewise typeof(Host<T>), proven through the same self-instantiation
+        // bridge but via the iterator state-machine predicate (#2968).
+        var path = typeof(LadderRung9.DynamicMemberContexts).Assembly.Location;
+        using var source = MetadataSource.Open(path);
+        string? moveNextOutput = null;
+        foreach (var (typeName, methodName, function) in IrImporter.ImportAssembly(source))
+        {
+            if (methodName == "MoveNext" && typeName.Contains("InGenericIterator"))
+            {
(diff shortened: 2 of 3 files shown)


richlander:
## Corpus A/B — structural inventory (byte-identical)

Ran the DecompilerHarness default structural inventory (Full% + stop-reason
buckets + importer-bug detection) over the net11 daily shared framework
(`Microsoft.NETCore.App 11.0.0-preview.6.26359.118`), base vs head:

- **Base** `2077bab4` (origin/main): `128234/135595 Full (94.57%); importer bugs: 2`
- **Head** `c26a144a`: `128234/135595 Full (94.57%); importer bugs: 2`
- Reports are **SHA-256 byte-identical** (`42AB97E3…842B1`); identical
  stop-reason histogram.

The two `KeyNotFoundException` importer bugs are **pre-existing on base**
(present identically at `2077bab4`) and are unrelated to this change. Zero
movement: as expected, a dynamic site that previously left explicit
`Binder.GetMember` scaffolding (already a complete `Full` decompilation) now
raises to `((dynamic)x).Member` (still `Full`), so no bucket moves. The SPC
corpus contains no dynamic-in-generic-enclosing sites that would change the
per-method report text.

Command: `decompiler-harness "<NETCore.App/11.0.0-preview.6.26359.118>"`

richlander:
## Adversarial review — reconciliation (two clean families, head `c26a144a`)

Two-family adversarial review per AGENTS.md, each isolated in its own detached
worktree at the exact head `c26a144a` (base `2077bab4`). Both **CLEAN** with
zero findings, so the reviewed head is the final head (no re-review needed).

### GPT-5.6 Sol — CLEAN
Built and ran the suites (DynamicCallSitePassTests 110, LadderRung9GateTests 19,
TypeRefDecoderRecursionTests 10 — all pass). Probed the missing-arity/count check
and nested-generic alignment; found no reproducible correctness, safety, arity,
or nested-generic issue.

### Gemini 3.1 Pro — CLEAN
- **Identity**: `TypeRef.Definition` joins nested names with `+`
  (`Outer+Inner`), so `ElementType.Equals(enclosing)` is unique per declaring
  chain. A forged top-level `Outer+Inner` would merely translate a forged
  binder context into the correct `((dynamic)x).Member` syntax — no semantic
  loss, consistent with pre-PR identity rules.
- **Generic-parameter indexing**: .NET indexes type parameters globally within
  the type hierarchy (`VAR 0` = `T`, `VAR 1` = `U`), so
  `GenericParameterIndex == i` strictly matches compiler-emitted
  self-instantiations across all arities and nesting depths
  (`Outer<T>.Inner<U>`, `Outer<T>.InnerNonGeneric`). Method type parameters
  decode to the distinct `MethodGenericParameter` kind and are correctly
  rejected.
- **Malformed/adversarial metadata**: null `ElementType` / default
  `TypeArguments` are fully guarded (`is not { }` / `IsDefaultOrEmpty`); a
  context with fewer args than the definition's arity is accepted harmlessly and
  decompiles back to valid C#.
- **Regressions**: 110 + 19 + 10 pass; the generic fixture does not disturb
  assembly-enumeration gating.

### Shared conclusion
Both reviewers independently confirmed that the deliberate omission of an
arity/count check is **not** a correctness or security hole: a dynamic member
access decompiles to `((dynamic)x).Member` regardless of the binder context's
exact type arguments, and the enclosing-definition identity is already pinned by
`ElementType.Equals(enclosing)` behind the `[CompilerGenerated]` +
display-class/iterator-name gate. No changes were required.

richlander:
Ready to merge

PR #2985, opened 2026-07-21
Decompose library Performance Triage into kind-scoped sections (#2833)

### LINKED ISSUE #2833: Performance Triage --table: curated default column set

richlander:
## Summary

Fixes #2833. The library-scoped `Performance Triage` section packed every
optimization-opportunity shape into one wide, sparse table that was hard for
humans (and agents) to read. This decomposes it into **kind-scoped sections**
following the established **il-offset section model**.

Seven kind sections (plus a non-lossy `Performance: Other`):
`Performance: Boxing`, `Arrays`, `Closures and delegates`, `Enumerators`,
`Loop hot paths`, `Allocation hotspots`, `Async`.

Behavior:

- **Opt-in and absent when empty** (CanRender gates on that kind having ≥1
  finding), like offset-context sections. Not in the default `-v:m` view.
- **Selectable individually** (`-S "Performance: Boxing"`) or **as a group**
  (`-S @Performance`). Legacy `-S "Performance Triage"` / `Performance` /
  `Optimization Opportunities` **redirect to the group** at library scope.
- **Tight markdown columns** (Member, Evidence, Allocation, Loop, Reach,
  Weight, Confidence); full per-row diagnostics (shape, provenance, candidate,
  token, IL, path/loop evidence, fix) move to the **nested `performance` JSON
  object**, retiring the `optimization_opportunities` key.
- **`--count -S @Performance`** returns a per-kind count map (including zero
  rows) — a cheap way to probe which kinds are present before requesting them.
- **Group tabular** (`--tsv`/`--table`/`--jsonl`) renders one table because all
  kind sections share the `PerformanceRow` view (repeated headers collapsed).
- **Scope: library only.** The `type`/`member` `Performance Triage` lens (a
  single-member view) is unchanged.

## Evidence

- Full CLI suite: `dotnet run --project src/dotnet-inspect.Tests -c Release`
  → 1968 total, 2 failed, 10 skipped. The **2 failures are pre-existing and
  platform-specific** (`PInvokeMethods_*` assert Windows `Interop.User32`
  types; this run is on Darwin, which yields `mach_timebase_info`) and are
  untouched by this diff.
- New `PerformanceKindsTests` cover shape→section mapping (every
  `KnownShapes` entry maps to a non-`Other` section), structured-key
  uniqueness, and `AllShareCommonView`.
- New `CommandExecutionTests` cover: default-hidden, single-kind isolation,
  `@Performance` group union, absent-when-empty, `--count` per-kind map,
  legacy-name redirect, and single collapsed tabular header.
- Migrated the rich `--tsv`/`--jsonl` triage tests to observe the nested
  `--json` projection (the only place rich fields now live).
- Verified on a real `System.Text.Json.dll` (162 findings): group/single/count/
  redirect/nested-JSON all behave as described.

## Compatibility boundary

Library-only. The retired `optimization_opportunities` JSON key is replaced by
the nested `performance` object (experimental performance-analysis surface).
Legacy library section names redirect rather than error. Type/member triage is
untouched.

Adversarial review (two model families) to follow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,11 +196,17 @@ as a curated group of kind-scoped sections — `Performance: Boxing`,
 `Performance: Enumerators`, `Performance: Loop hot paths`,
 `Performance: Allocation hotspots`, and `Performance: Async` — that you can
 request individually (`-S "Performance: Boxing"`) or as a group (`-S
-@Performance`, also reachable via the legacy name `-S Performance`). Each
+@Performance`, also reachable via the legacy name `-S Performance`). The
+kind sections are catalog-hidden: they do not appear in the top-level
+`-D`/`-D --schema` discovery catalog, which lists the `@Performance` category
+as their single entrypoint. Drill in with `-D @Performance` to list the kinds,
+or `-D "Performance: Boxing"` for one kind's columns. Each
 section is absent when it has no findings, so the group renders exactly the
 kinds that were found; `--count -S @Performance` returns a per-kind count map
 (including zero rows) as a cheap way to discover which kinds are present before
-requesting them. These sections rank in-loop (hot) and high-confidence
+requesting them. In flattened tabular output (`--table`/`--tsv`/`--jsonl`) the
+group renders as one self-describing table with a leading `Kind` column, so
+every row states which performance kind it belongs to. These sections rank in-loop (hot) and high-confidence
 opportunities first across actionable rewrite shapes (small non-escaping
 arrays, temporary or span-to-array copies, capturing and instance method-group
 delegates, async state-machine setup, loop-invariant materialization, and
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ context for copied DLLs. A future `--deps` source can represent runtime
 | API compatibility | `diff` | Version ranges, package or platform diffs, breaking/additive/potentially-breaking classification, type and member filters, plus opt-in decompiled C#/IL/checksum-verified authored Source evidence. |
 | Relationships | `depends`, `extensions`, `implements` | Type hierarchies, package dependencies, library reference graphs, extension methods/properties, implementors and subclasses. Add `--project` to search project-referenced packages. |
 | Source mapping | `type`/`library`/`package -S "Source Files"`, `member -S "Source Locations"` / `"Original Source"` | SourceLink URLs, member file/line locations, source fetching, URL verification, token+IL-offset to source-line resolution. |
-| Performance analysis *(experimental)* | `library`/`type`/`member -S "Top Leverage"`, `"Performance Triage"`, `"Resource Triage"`, `"Call Graph"`, `"Caller Graph"` | Whole-assembly call-graph leverage ranking — direct callers, root reach, fanout, depth, loop calls — with opt-in per-node cost signals (alloc, copy, unsafe, reflection, throw/exception, catch/finally), actionable rewrite-shape detection, and exception-path resource-lifecycle candidates. |
+| Performance analysis *(experimental)* | `library -S @Performance` (kind sections: `"Performance: Boxing"`, `"Performance: Arrays"`, …), `type`/`member -S "Performance Triage"`, `"Top Leverage"`, `"Resource Triage"`, `"Call Graph"`, `"Caller Graph"` | Whole-assembly call-graph leverage ranking — direct callers, root reach, fanout, depth, loop calls — with opt-in per-node cost signals (alloc, copy, unsafe, reflection, throw/exception, catch/finally), actionable rewrite-shape detection, and exception-path resource-lifecycle candidates. |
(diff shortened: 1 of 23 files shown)


richlander:
## Incremental update: general `ListedInCatalog` flag + self-describing Kind column

New head: `bfac6cee` (base of this delta: `8e1f08d8`).

Two refinements driven by discovery noise and the earlier GPT `--table` finding:

1. **General `ListedInCatalog` section-descriptor flag** (default `true`), orthogonal to `ExplicitOnly` render gating. When `false`, a section is omitted from the top-level `-D`/`--schema` discovery catalog and is discoverable only by drilling into its curated `@category` or by exact name. The 8 kind-scoped performance sections now set it `false`, so `@Performance` is their single catalog entrypoint. `SectionPipeline.GetCatalogHiddenSections()` exposes the hidden set; `DiscoverOutput` filters the bare catalog generically on it (replacing an earlier hard-coded performance filter — one mechanism, reusable by any future curated group).

2. **Flattened tabular group output** (`-S @Performance --table`/`--tsv`/`--jsonl`) now renders **one** self-describing table with a leading `Kind` column (new `PerformanceGroupView`/`PerformanceGroupRow`), instead of N concatenated per-kind tables. This resolves the GPT finding that pretty `--table` emitted a header per kind and `--rows` counted embedded headers as data. The dead `CollapseRepeatedTabularHeaders` hack is removed. Markdown output (`-S @Performance` without a tabular flag) is unchanged.

Docs (README, analysis-ux-scopes, release-notes) updated. Member/type `@Facts` + `@Performance` curation split to #2989.

Verified on `System.Text.Json` (162 findings): bare `-D`/`--schema` omit the perf sections while listing `@Performance`; `-D @Performance` lists the 8 kinds; `-D "Performance: Boxing"` resolves; `--table --rows -n 5` → 1 header + 5 rows; `--tsv`/`--jsonl` carry a `Kind`/`kind` column on every row; single-kind selection has no Kind column. Full CLI suite: 1976 total, 2 failed (pre-existing Darwin PInvoke `Interop.User32`), 10 skipped.

Adversarial re-review of this delta in progress (GPT + Gemini families, isolated worktrees at exact head). Not ready to merge until both fixed-head reviews are clean.

richlander:
## Adversarial review reconciliation — head `4ab660b5`

Two reviewers from different model families, isolated worktrees, `/tmp`-only scratch.

### Round 1 (head `bfac6cee`)
- **Gemini 3.1 Pro — CLEAN.** Verified discovery filtering (`-D`, `--schema`, tree, drill-in), single-table flatten with `--rows` accounting, single-kind = no Kind column, duplicate-row preservation, curated ordering, and no reflection/Roslyn/`Assembly.Load`.
- **GPT-5.6-sol — one Medium.** The multi-assembly renderer `WriteLibraryResults` (reached via `library --tfm all` when a library ships under several TFMs) still serialized per-kind sections directly, so `-S @Performance` in that path lost the Kind column, repeated per-kind headers, and applied `--rows` per kind. Reproduced on `Newtonsoft.Json --tfm all`.

### Resolution (`bfac6cee` → `4ab660b5`)
Extracted a shared `WriteLibraryTabular` helper in `OutputFormatter.cs` and call it from **both** the single-assembly (`WriteLibraryResult`) and multi-assembly loop (`WriteLibraryResults`) tabular paths. Each assembly now emits one self-describing Kind-labeled table (one header, aligned, `--rows` counts data rows only); single-kind selection still omits the Kind column. Added `LibraryCommand_TfmAll_PerformanceGroupTabular_IsKindLabeledPerAssembly` (two-TFM same-name package). Note: `package --all-libraries` guards against category selectors, so the reachable multi-inspection path is `library <name> --package <pkg> --tfm all -S @Performance`.

### Round 2 — fixed head `4ab660b5` (both CLEAN)
- **GPT-5.6-sol — CLEAN.** Original finding resolved; multi-TFM shows one Kind header per assembly, zero bare headers, correct per-table `--rows`, Kind on every TSV/JSONL row; single-assembly + single-kind unchanged; no new defect from the refactor.
- **Gemini 3.1 Pro — CLEAN.** Independently reproduced the multi-assembly fix (TSV/JSONL/table), re-confirmed single-assembly discovery/flatten/dedup/order, and reviewed the helper extraction as a pure equivalence transformation.

Full CLI suite at fixed head: 1977 total, 2 failed (pre-existing Darwin PInvoke `Interop.User32` asserting Windows-only interop), 10 skipped (ildasm/ilasm absent).

**Ready to merge**

PR #2966, opened 2026-07-21
Raise switches with shared miss-handler/return exit chains (#2954)

richlander:
- Fixes #2954
- Changes SwitchRaisingPass to raise switches whose case regions have a shared miss-handler that chains into a shared return, instead of falling back to an if-ladder
- Card revision: `5cb72d80`

## Change

> Should we accept this change?

**Conclusion:** **PASS** — generalizes an existing join/tiling invariant to a legitimate real-world CFG shape, verified against a real Humanizer method and a synthetic regression test, with no other pass logic touched.

### Original source

```il
IL_0000: ldarg.2
IL_0001: ldc.i4.2
IL_0002: beq.s        IL_0006
IL_0004: ldarg.1
IL_0005: ret
IL_0006: ldarg.1
IL_0007: brfalse      IL_0226
IL_000C: ldarg.1
IL_000D: callvirt     instance int32 [System.Runtime]System.String::get_Length()
IL_0012: ldc.i4.s     28
IL_0014: sub
IL_0015: switch       (IL_00B9, IL_003D, IL_0081, IL_01AF, IL_00CE, IL_0063, IL_01A0)
... (full body: two-tier string-switch lowering — outer int dispatch on
    resourceKey.Length - 28, seven buckets each doing a char/string
    sub-discriminator, all funneling into one of ~19 "miss" arms that share a
    single miss-handler block, which falls through into the method's one
    return block; matched arms jump directly to that same return, skipping
    the miss-handler)
```

### Before

```csharp
protected override string GetResourceKey(string resourceKey, int number)
{
    int __dotnet_inspect_switch0 = default;
    string V_0;
    char V_2;

    if (number == 2) goto IL_0006;
    return resourceKey;
    IL_0006:
    if (resourceKey is null) goto IL_0226;
    __dotnet_inspect_switch0 = (int)(resourceKey.Length - 28);
    if (__dotnet_inspect_switch0 == 0) goto IL_00B9;
    if (__dotnet_inspect_switch0 == 1) goto IL_003D;
    if (__dotnet_inspect_switch0 == 2) goto IL_0081;
    if (__dotnet_inspect_switch0 == 3) goto IL_01AF;
    if (__dotnet_inspect_switch0 == 4) goto IL_00CE;
    if (__dotnet_inspect_switch0 == 5) goto IL_0063;
    if (__dotnet_inspect_switch0 == 6) goto IL_01A0;
    V_0 = resourceKey;
    return V_0;
    IL_003D:
[...]

Issue #2954, opened 2026-07-21
EVIL frontier: switch/jump-table not recovered (emitted as __dotnet_inspect_switchN if-ladder)


richlander:
## Cluster (EVIL benchmark frontier)

Part of the **`valid_different.semantic_opcode_diff.syntax` frontier** (1,632 rows total). In a probed sample of **343 frontier methods across 8 assemblies**, **56 (16%) lose a `switch` jump table** on compile-back: the original IL has one or more `switch` opcodes, the recompiled IL has **zero**. (When the decompiler *does* recover a switch, it stays exact — 16 preserved.)

Scaled to the full frontier this is ~260 rows, and it is the largest *actionable* frontier sub-shape (the ~55% stack-scheduling/temp-materialization majority is valid-and-acceptable).

## Symptom (genuine fidelity + readability defect)

The decompiler expands a `switch` (jump table) into a **synthesized selector local + linear `if (… == k) goto IL_xxxx` ladder** instead of recovering a `switch` statement. Example (`Humanizer.MalteseFormatter::GetResourceKey#0`):

```csharp
int __dotnet_inspect_switch0 = default;
...
__dotnet_inspect_switch0 = (int)(resourceKey.Length - 28);
if (__dotnet_inspect_switch0 == 0) goto IL_00B9;
if (__dotnet_inspect_switch0 == 1) goto IL_003D;
if (__dotnet_inspect_switch0 == 2) goto IL_0081;
...
```

The `__dotnet_inspect_switch0` synthesized-temp name plus the goto ladder is both low-fidelity (recompiles without the `switch` opcode) and low-readability.

## Reproduction

Reflection probe through `ReturnToSender.CompileBackTargets`, comparing `Result.OriginalOpcodes` vs `Result.RecompiledOpcodes` (count of `switch` tokens). Representative targets:
- `Humanizer.MalteseFormatter::GetResourceKey#0` (1 → 0)
- `Humanizer.ToTitleCase::IsArticleOrConjunctionOrPreposition#0` (2 → 0)
- `MySqlConnector.Core.TypeMapper::ConvertToColumnTypeAndFlags#0` (4 → 0)
- `Moq.ExpressionComparer::Equals#0` (1 → 0)
- `Dapper.SqlMapper::FlexibleConvertBoxedFromHeadOfStack#0` (1 → 0)

## Fix location

[...]

PR #2988, opened 2026-07-22
Extract shared Unify decision from Raise/FinishSwitchRaise into TryUnify

### LINKED ISSUE #2971: SwitchRaisingPass has two parallel Unify/tiling implementations that can drift out of sync
## Background

While fixing #2954 (PR #2966), the shared-miss-handler/return exit-chain fix
was applied to `Unify` inside `Raise` (drives the IL `switch` opcode raiser),
but `SwitchRaisingPass.cs` has a second, independently-implemented `Unify`
local function inside `FinishSwitchRaise` (drives `RaiseStringEqualityChain`
and the hash/length-bucket and sparse-int switch raisers) with the same
region-growing/tiling/exit-unification shape.

The first fix missed `FinishSwitchRaise`''s copy entirely. This was only
caught by adversarial review (Gemini 3.1 Pro), not by the original
implementation or by Claude Opus''s first-round review — both independently
verified `Raise`''s logic in isolation without noticing the second, parallel
implementation existed. A second commit (`aac62aec`) applied the identical
fix to `FinishSwitchRaise` and added regression coverage exercising that path
specifically.

## Problem

`Raise` and `FinishSwitchRaise` each independently implement:
- Growing an owned-block region from a set of case targets.
- Unifying/converging on the region''s exit block (`join`).
- Tiling invariants confirming `owned` exactly covers `[start, regionEnd)`
  with no external inbound jumps.

Any future change to one of these region-growing/tiling/unification code
paths (bug fix, new adversarial CFG shape, hardening) risks the same
copy-paste drift: fixed in one function, silently missed in the other,
because nothing ties them together structurally or by test coverage across
both.

## Suggested resolution

Consider extracting the shared region-growing/`Unify`/tiling logic (the
parts identical or near-identical between `Raise` and `FinishSwitchRaise`)
into a single shared helper parameterized over the two functions'' differing
inputs (`blocks`/`caseTargets`/`offsetToIndex`/`owned` shapes differ in name
but not in kind). This would make future fixes structurally impossible to
apply to only one path.

If a full extraction is too risky/large for a single change, at minimum:
- Add a code comment at each `Unify` cross-referencing the other, noting they
  must be kept in sync.
- Consider a test that exercises the same adversarial CFG shape through both
  `Raise` and `FinishSwitchRaise` entry points side by side, so a future
  one-sided fix fails visibly instead of requiring a reviewer to notice the
  gap.

## References

- `src/ILInspector.Decompiler/Pipeline/Passes/SwitchRaisingPass.cs`:
  `Unify` inside `Raise` (~lines 105-131) and `Unify` inside
  `FinishSwitchRaise` (~lines 1208-1226), both post-fix.
- PR #2966 (fix for #2954): commit `5cb72d80` (Raise-only fix),
  commit `aac62aec` (FinishSwitchRaise fix + regression coverage, added after
  Gemini 3.1 Pro''s adversarial review caught the gap).

richlander:
Fixes #2971

## Change

`SwitchRaisingPass` has two independent case-region-growing raisers,
`Raise` (drives the IL `switch` opcode raiser) and `FinishSwitchRaise`
(drives `RaiseStringEqualityChain` and the hash/length-bucket and
sparse-int raisers), each with its own local `Unify` function deciding
when two case-region exits converge on a shared join. The two were
byte-for-byte identical, and PR #2966 (issue #2954) had to fix the same
defect in both separately — the first commit only touched `Raise`''s copy,
and the gap in `FinishSwitchRaise` was caught only by adversarial review
(Gemini 3.1 Pro), not by the original implementer or the first-round Opus
review.

This PR extracts the identical decision logic into a single shared static
`TryUnify(..., ref int? join, int j)`, called by both `Raise`''s and
`FinishSwitchRaise`''s own `Unify` local functions as one-line forwarders.
`join` itself stays a local variable in each function (it's inherently
per-call-site state); only the decision/mutation logic is shared. This
makes a future fix to the unification rule structurally impossible to
apply to only one of the two raisers.

**No other logic in `Raise` or `FinishSwitchRaise` changed** — this is a
pure mechanical move: same inputs (`blocks`, `caseTargets`, `offsetToIndex`,
`owned`), same `ChasesTo` calls, same result, in the same order.

Region-growing/tiling logic *around* `Unify` in the two functions
(`GrowRegion` calls, default-handling branches, tiling/bounds checks,
`OnlyReachedByTable` vs `OnlyReachedByChain`) genuinely differs between the
two raisers (different dispatch models: table dispatch at a single block vs.
a chain of blocks) and was intentionally left alone — extracting that too
would be a much larger, riskier refactor for uncertain benefit. Per the
issue, at minimum this closes the exact gap that caused #2954''s missed fix.

## Evidence

- Build: clean, 0 errors.
- Fast broad sweep (`-trait "Area=Pass" -trait- "Speed=Slow"`, new in #2979):
  1006 of 1119 tests selected, 1 failed — `NullCoalescingAssignmentPassTests
  .LazyFieldGetter_RecompilesOpcodeExact` (`RecompileFail`), confirmed
  pre-existing by re-running the identical test with this change stashed out
  (fails identically pre-change).
- Targeted switch-raising suite (`SwitchRaisingSharedMissHandlerTests`,
  `StringSwitchRaisingTests`, `SwitchExpressionRaisingTests`,
  `SwitchDuplicateLabelTests`, `IdiomShapeScorecard`,
  `SparseIntSwitchRaisingTests`, `SwitchBranchRenderingTests`,
  `NestedScopeNameCollisionTests`): 41 tests, 1 failed — the known
  pre-existing `SmallStringSwitch_RendersSwitchOnString` CRLF-vs-LF mismatch,
  unrelated (also present on `main`).
- Real-world witness (`Humanizer.MalteseFormatter.GetResourceKey`,
  Humanizer.core 3.0.10): rendered output is byte-identical, before and
  after this change — confirms `Raise`''s behavior through the refactor.

## Validation

```bash
dotnet build dotnet-inspect.slnx -c Release --configfile ./nuget.config
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -trait "Area=Pass" -trait- "Speed=Slow"
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class "*SwitchRaisingSharedMissHandlerTests*" -class "*StringSwitchRaisingTests*" -class "*SwitchExpressionRaisingTests*" -class "*SwitchDuplicateLabelTests*" -class "*IdiomShapeScorecard*" -class "*SparseIntSwitchRaisingTests*" -class "*SwitchBranchRenderingTests*" -class "*NestedScopeNameCollisionTests*"
```

Touches core switch-raising logic (`SwitchRaisingPass.cs`), so per AGENTS.md
this will go through two-model adversarial review even though it is a
behavior-preserving mechanical extraction, not a new heuristic.

### DIFF
--- a/src/ILInspector.Decompiler/Pipeline/Passes/SwitchRaisingPass.cs
+++ b/src/ILInspector.Decompiler/Pipeline/Passes/SwitchRaisingPass.cs
@@ -104,31 +104,9 @@ public sealed class SwitchRaisingPass : IIrPass
         var owned = new HashSet<int>();
         int? join = null;
 
-        // Two exits unify not only when identical, but when one flows into the
-        // other through a chain of plain, unclaimed, single-successor blocks —
-        // e.g. a case-local "no match" arm and a shared miss-handler that falls
-        // into the same terminating return. The upstream (earlier) block becomes
-        // the join, so the chain is naturally emitted once as switch-trailing
-        // code; the tiling check below still rejects the join if some other
-        // owned block ends up past it.
-        bool Unify(int j)
-        {
-            if (join is not { } existing)
-            {
-                join = j;
-                return true;
-            }
-            if (existing == j)
-                return true;
-            if (ChasesTo(blocks, existing, j, caseTargets, offsetToIndex, owned))
-                return true;   // existing join already flows into j — keep it
-            if (ChasesTo(blocks, j, existing, caseTargets, offsetToIndex, owned))
-            {
-                join = j;   // j is upstream of the existing join — adopt it
-                return true;
-            }
-            return false;
-        }
+        // See TryUnify's doc comment for the exit-unification rule (shared with
+        // FinishSwitchRaise's identical need below).
+        bool Unify(int j) => TryUnify(blocks, caseTargets, offsetToIndex, owned, ref join, j);
 
         // Each distinct case target grows into its single-entry region; the
         // region's exits unify to the shared join (or it terminates).
@@ -611,6 +589,41 @@ public sealed class SwitchRaisingPass : IIrPass
                 Add(idx, t);
         }
         return preds;
+    }
+
+    /// <summary>
+    /// Decides whether exit <paramref name="j"/> unifies with the current
+    /// candidate join in <paramref name="join"/>: two exits unify not only when
+    /// identical, but when one flows into the other through a chain of plain,
+    /// unclaimed, single-successor blocks (see <see cref="ChasesTo"/>) — e.g. a
+    /// case-local "no match" arm and a shared miss-handler that falls into the
+    /// same terminating return. The upstream (earlier) block becomes the join,
+    /// so the chain is naturally emitted once as trailing code after the switch;
+    /// the caller's own tiling check afterward still rejects the join if some
+    /// other owned block ends up past it.
+    ///
+    /// Shared by <see cref="Raise"/> and <see cref="FinishSwitchRaise"/> — the
+    /// two independent case-region-growing raisers — so a fix to this decision
+    /// applies to both instead of risking the two drifting out of sync (see
+    /// issue #2971).
+    /// </summary>
+    static bool TryUnify(IReadOnlyList<Block> blocks, int[] caseTargets, Dictionary<int, int> offsetToIndex, HashSet<int> owned, ref int? join, int j)
+    {
+        if (join is not { } existing)
+        {
+            join = j;
+            return true;
+        }
+        if (existing == j)
+            return true;
+        if (ChasesTo(blocks, existing, j, caseTargets, offsetToIndex, owned))
+            return true;   // existing join already flows into j — keep it
+        if (ChasesTo(blocks, j, existing, caseTargets, offsetToIndex, owned))
+        {
+            join = j;   // j is upstream of the existing join — adopt it
+            return true;
+        }
+        return false;
     }
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


richlander:
## Adversarial review (per AGENTS.md)

Touches core switch-raising logic (`SwitchRaisingPass.cs`), so this went
through two-model review even though it's a behavior-preserving mechanical
extraction, not a new heuristic. Two isolated worktrees, both at the exact
PR head `b04bb749`.

- **Claude Opus 4.8**: Clean. Verified `TryUnify`'s body is textually
  identical to both former `Unify` closures (same statements, same order,
  same two `ChasesTo` calls, same adopt-upstream logic), confirmed the
  `ref int? join` threading is correct at both call sites with no aliasing or
  shared state between `Raise` and `FinishSwitchRaise`, confirmed
  `blocks`/`caseTargets`/`offsetToIndex`/`owned` are never reassigned after
  `Unify` starts being called (so passing them as ordinary parameters is
  equivalent to the old closure capture), and confirmed no other logic
  changed. Build clean; fast `Area=Pass` sweep (1119 tests) and the targeted
  switch suite (41 tests) each showed only the one known pre-existing,
  unrelated failure apiece — no new failures.
- **Gemini 3.1 Pro**: Clean. Independently confirmed the same points (identical
  logic, correct `ref` threading, safe reference semantics, no unintended
  scope creep) and reproduced the same test results (only the two known
  pre-existing failures, nothing new).

Both reviewers are clean on the same fixed head (`b04bb749`).

Ready to merge.

Issue #2952, opened 2026-07-21
Decompiler printer: expand multi-statement lambda block bodies across lines (match original)


richlander:
### Summary

When the decompiler raises a lambda whose body is a statement block, the C# printer emits the whole block on a single line instead of expanding it across multiple lines the way normal method/statement blocks are printed. The output is semantically identical and valid, but it does not match the shape a developer wrote (or the shape the printer already uses for every other block body), which hurts readability and side-by-side comparison against original source.

### Example

Real witness: `System.CommandLine.Command::SetAction:1` (System.CommandLine `3.0.0-preview.5.26302.115`), now fully raised after #2948.

Original source:

```csharp
Action = new AnonymousSynchronousCommandLineAction(context =>
{
    action(context);
    return 0;
});
```

Current decompiled output:

```csharp
Action = new AnonymousSynchronousCommandLineAction(context => { action.Invoke(context); return 0; });
```

Desired decompiled output (block body expanded like any other block):

```csharp
Action = new AnonymousSynchronousCommandLineAction(context =>
{
    action.Invoke(context);
    return 0;
});
```

### Notes

- Concern lives in the C# printer / formatter (`ILInspector.CSharp`), not in raising — the IR is already correct; this is purely how a lambda's block body is laid out.
- A single-statement / expression-bodied lambda (`x => x + 1`) should stay inline; only multi-statement block bodies (and arguably single-statement blocks, to be decided) should expand.
- Reproduce:

  ```bash
  dotnet run --project src/dotnet-inspect -c Release -- \
    member "System.CommandLine@3.0.0-preview.5.26302.115" \
    System.CommandLine.Command -m SetAction:1 -S "Decompiled Source" --all
  ```

Follow-up to #2945 / #2948.

PR #2992, opened 2026-07-22
Fix stale _statementIndent in DoWhileLoop/TryCatch/Switch

richlander:
## Context

PR #2987 (issue #2952: expand multi-statement lambda block bodies
across lines in the printer) added a _statementIndent field that
AppendStatement sets unconditionally at entry, with no general
save/restore. That is only safe when a statement's own expression
composition happens entirely *before* any nested child-block render.

Adversarial review (Gemini 3.1 Pro) on PR #2987 found three
counterexamples where that assumption doesn't hold, all still
unresolved because #2987 was merged before the fix landed:

- **DoWhileLoop**: the condition is evaluated *after* the body.
- **TryCatch**: each catch clause's header/filter is evaluated *after*
  the try body and after each preceding catch clause's body.
- **Switch**: each section's case labels render after the previous
  section's body (a ` when ` pattern guard could in principle hold a
  lambda).

A multi-statement lambda in a do/while condition, a ` catch when (...) `
filter, or (in principle) a switch case guard renders its block braces
at the stale, deeper indent left over from the preceding body instead
of the enclosing statement's own indent.

## Fix

Reset ` _statementIndent = indent; ` right before each of these three
locations composes its own expression, mirroring the existing pattern
already used elsewhere in the printer (e.g. \IfStatement\, \ForLoop\).

## Evidence

Added \CSharpPrinterStatementIndentTests\, which builds IR directly
(bypassing \LambdaRaisingPass\, which does not raise a lambda passed as
a delegate-cache argument inside a loop condition or catch filter in
this environment, so a compiled source fixture can't exercise this) to
regression-test the DoWhileLoop and TryCatch cases:

- Verified both new tests **fail** without the fix (confirmed by
  temporarily reverting the \_statementIndent\ resets) and **pass**
  with it.
- Full solution build: \dotnet build dotnet-inspect.slnx -c Release
  --configfile nuget.config\ — succeeded.
- Fast targeted suite: \ILInspector.Decompiler.Tests.exe -trait
  \"Area=Pass\" -trait- \"Speed=Slow\"\ — 1121/1122 passed; the 1 failure
  (\NullCoalescingAssignmentPassTests.LazyFieldGetter_RecompilesOpcodeExact\)
  is a pre-existing environment issue unrelated to this change (also
  fails on a clean \origin/main\ checkout).

Low-risk, mechanical fix scoped to three narrow indent-bookkeeping
sites with new regression tests proving both the bug and the fix;
adversarial review still requested per repo convention given the
non-trivial correctness nature of the original bug.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CSharpPrinterStatementIndentTests.cs
+++ b/src/ILInspector.Decompiler.Tests/CSharpPrinterStatementIndentTests.cs
@@ -1,0 +1,123 @@
+using System.Collections.Immutable;
+using ILInspector.Decompiler.Pipeline;
+
+namespace ILInspector.Decompiler.Tests;
+
+// Printer-only regression coverage for the #2952 follow-up (PR #2987 adversarial
+// review): _statementIndent is set unconditionally at AppendStatement entry with
+// no general save/restore, which is only safe when a statement's own expression
+// composition happens entirely *before* any nested child-block render. DoWhileLoop
+// (condition evaluated after the body) and TryCatch (catch header/filter evaluated
+// after the try body and each preceding catch body) are the two counterexamples —
+// both are fixed by resetting _statementIndent right before the subsequent
+// expression composes. These tests build the IR directly (bypassing
+// LambdaRaisingPass, which does not raise a lambda passed as a delegate-cache
+// argument inside a loop condition or catch filter in this environment) so the
+// printer's indent bookkeeping is exercised in isolation.
+[Trait("Area", "Pass")]
+public class CSharpPrinterStatementIndentTests
+{
+    static readonly TypeRef Holder = TypeRef.Definition("synthetic", "", "Holder");
+    static readonly TypeRef Void = TypeRef.CoreLib("System", "Void");
+    static readonly TypeRef Int32 = TypeRef.CoreLib("System", "Int32");
+    static readonly TypeRef Bool = TypeRef.CoreLib("System", "Boolean");
+    static readonly TypeRef ListOfInt = TypeRef.GenericInstance(TypeRef.Definition("System.Private.CoreLib", "System.Collections.Generic", "List`1"), [Int32]);
+
+    static Lambda MultiStatementPredicate()
+    {
+        var body = new BlockContainer();
+        var block = new Block(0);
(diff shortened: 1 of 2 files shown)


richlander:
## Adversarial review — both reviewers clean

**GPT-5.6-Terra** and **Gemini 3.1 Pro** independently reviewed the exact PR head (\d3def679\) in isolated worktrees. Both first-pass responses were too terse (\"no significant issues\" / blank) and were sent back demanding concrete file:line evidence per attack point; both follow-ups were thorough.

### Findings, reconciled

1. **Fix correctness** — both reviewers traced the reset placement (\CSharpPrinter.cs\: do-while ~1648, try-catch clause loop ~1665) including a hand-traced nested-do-while-inside-do-while scenario, and confirmed the reset happens at exactly the right point at every nesting depth (each \AppendStatement\ call sets \_statementIndent = indent\ at entry and its own \while\/catch-header reset restores its own \indent\ regardless of how deep a nested statement drove it).
2. **Completeness** — both independently re-verified, by reading the code (not trusting the PR description), that \ForLoop\, \WhileLoop\, \IfStatement\, \ForeachStatement\, \UsingStatement\, \Lock\, \Fixed\, and \TryFinally\ all compose their header/condition before recursing into the child body (or, for \TryFinally\, have no expression to misalign at all) — no missed cases.
3. **Test quality** — both confirmed the two new tests in \CSharpPrinterStatementIndentTests.cs\ are valid, non-tautological regression tests: they reasoned through what \_statementIndent\ would be without the reset (stale, deeper) and confirmed the strict \Assert.Contains\ (zero-indent \{\) would genuinely fail without the fix. This matches this session's own verification (temporarily reverted the fix locally, confirmed both new tests fail; restored it, confirmed both pass).
4. **Switch case (untested)** — both reviewers went further than the original PR rationale and found the *stronger* reason the switch reset can't be exercised: \SwitchSection.Labels\ is statically \ImmutableArray<Constant>\ in this decompiler's IR, so a \Lambda\ node can never structurally appear in a case label at all (not just \"unlikely in valid C#\" as the original PR description said, but IR-impossible). The reset is confirmed-dead but harmless defensive code.
5. **Regressions** — both confirmed the resets only affect \_statementIndent\, which is read solely by multi-statement lambda brace padding (\LambdaBlockText\); ordinary (non-lambda) condition/filter/case-label rendering is untouched.

No bugs found by either reviewer after the deeper pass. Marking **Ready to merge**.

richlander:
Ready to merge

Issue #2994, opened 2026-07-22
Decompiler: reconstruct chained/embedded assignment from dup-of-value (a = b = c = v)


richlander:
## Summary

C# chained/embedded assignment (`a = b = c = value;`) compiles to a `dup`-of-value
idiom in IL — the value is pushed once and `dup`'d to each successive sink:

```text
ldc.i4.0
dup
call set_C(bool)
dup
call set_B(bool)
call set_A(bool)
```

The decompiler currently renders each sink as an independent, reordered statement
rather than reconstructing the source-level chain. It is valid and behavior-preserving,
but not fully raised.

## Example (Dapper.SqlMapper.Settings.SetDefaults)

Original source:

```csharp
ApplyNullValues = PadListExpansions = UseIncrementalPseudoPositionalParameterNames = PreferTypeHandlersForEnums = false;
FetchSize = InListStringSplitCount = -1;
```

Current decompiled output (after #2982 validity fix):

```csharp
Settings.PreferTypeHandlersForEnums = false;
Settings.UseIncrementalPseudoPositionalParameterNames = false;
Settings.PadListExpansions = false;
Settings.ApplyNullValues = false;
Settings.InListStringSplitCount = -1;
Settings.FetchSize = (long)-1;
```

## Intended fully-raised endpoint

Reconstruct the embedded-assignment expression so consecutive sinks fed by the same
dup'd value collapse to `A = B = C = value;` (respecting the observable
right-to-left evaluation order and any per-sink coercion, e.g. the `long` widening
on `FetchSize`).

## Context

Follow-up raise slice tracked from the #2982 validity fix, which stopped the dup'd
constant from spilling into an int temp (CS0029). Reconstructing the chain is the
remaining quality slice to reach the fully-raised state.

PR #2995, opened 2026-07-22
Re-materialize a dup'd constant instead of spilling to a slot (#2982)

### LINKED ISSUE #2982: Decompiler: chained bool assignment renders int temp -> CS0029 (bool member = int)
## Summary

The decompiler emits **uncompilable C#** (CS0029) when a boolean constant is
assigned to several `bool` members via a **chained assignment**. The shared
`false`/`true` constant is materialized as an **`int` temp** and threaded through
the chain, then assigned to each `bool` member without a `bool` coercion:

```
int S_256 = 0;
Settings.PreferTypeHandlersForEnums = S_256;   // bool property = int  -> CS0029
```

This is a genuine **product-decompiler** validity defect (present in the target
method's own decompiled body), not an RTS/harness artifact.

## Reproduction (self-contained, released package)

```bash
dnx dotnet-inspect -y -- member "Dapper.SqlMapper.Settings" "SetDefaults" \
  --package Dapper@2.1.79 --all -S "Decompiled Source"
```

- Package: **Dapper@2.1.79** (assembly version `2.0.0.0`, corpus TFM `net10.0`)
- Member: `Dapper.SqlMapper.Settings.SetDefaults()`

### Observed (product output — does not compile)

```csharp
public static void SetDefaults()
{
    Nullable<int> V_0 = default;
    Settings.CommandTimeout = V_0;
    int S_256 = 0;
    Settings.PreferTypeHandlersForEnums = S_256;                     // CS0029 int->bool
    int S_257 = S_256;
    Settings.UseIncrementalPseudoPositionalParameterNames = S_257;   // CS0029 int->bool
    int S_258 = S_257;
    Settings.PadListExpansions = S_258;                              // CS0029 int->bool
    Settings.ApplyNullValues = S_258;                                // CS0029 int->bool
    Settings.AllowedCommandBehaviors = unchecked((CommandBehavior)(-10));
    int S_259 = -1;
    Settings.InListStringSplitCount = S_259;
    Settings.FetchSize = (long)S_259;
}
```

`PreferTypeHandlersForEnums`, `UseIncrementalPseudoPositionalParameterNames`,
`PadListExpansions`, and `ApplyNullValues` are all `bool` properties, so each
`= S_25x` assignment is CS0029 ("Cannot implicitly convert type 'int' to 'bool'").

### Original source

<https://github.com/DapperLib/Dapper/blob/72a54c475f75e18cb93cba0809d00a5e6e49efd9/Dapper/SqlMapper.Settings.cs>

```csharp
CommandTimeout = null;
ApplyNullValues = PadListExpansions = UseIncrementalPseudoPositionalParameterNames = PreferTypeHandlersForEnums = false;
AllowedCommandBehaviors = DefaultAllowedCommandBehaviors;
FetchSize = InListStringSplitCount = -1;
```

The IL for the chained `... = false` assignment loads the boolean constant once
(`ldc.i4.0`) and `dup`s it across the successive `stsfld` targets. The decompiler
recovers the chain as a **shared `int` temp** (`S_256 = 0`, `S_257 = S_256`, …)
and stores it into each `bool` member.

## Analysis

Note the contrast within the same method: single-use / non-shared constants are
coerced correctly — `AllowedCommandBehaviors = unchecked((CommandBehavior)(-10))`
(enum cast applied) and `FetchSize = (long)S_259` (long cast applied). The failure
is specific to the **shared boolean constant with multiple consumers**:

- A boolean constant produced by `ldc.i4.0`/`ldc.i4.1` should render as the `bool`
  literal `false`/`true` (or be coerced to `bool`) at each `bool`-typed store.
- Because the value is shared across a chained assignment (multiple consumers via
  `dup`), it is value-numbered into a single temp typed by its **IL stack type
  (`int32`)** rather than by the consuming members' type (`bool`), so per-use
  boolean-literal recovery / coercion never fires and the `int`→`bool` store is
  left invalid.

The correct output is either the chained boolean literal assignment
(`ApplyNullValues = PadListExpansions = … = false;`) or, at minimum, a `bool`-typed
representation of the shared constant.

## Fix locus

Decompiler raise / IR temp typing + boolean-constant recovery — **not** RTS. Likely
the shared-constant temp-materialization / value-numbering path where a constant
with multiple consumers is typed by its IL stack type instead of by the store
target's type (boolean-literal recovery that today only fires for single-consumer
values).

Related umbrella: **#2338** (coercion-at-store validity classes) covers
numeric/enum/pointer store-target coercion but not this boolean chained-assignment
shape; this is a distinct trigger (shared constant → lost `bool` typing).

## Provenance

Surfaced by the EVIL authored-source correspondence benchmark
(`--benchmark-authored-corpus`), `Invalid` bucket, `CS0029`
(`type 'int' to 'bool'`). Distinct from the cross-assembly enum shape family
(#2953): here the member type is a same-assembly primitive `bool`, no referenced
type resolution is involved.

richlander:
- Fixes #2982
- Changes the IL importer to re-materialize a dup'd pure constant (clone per use) instead of spilling it to an int stack slot, so per-sink constant recovery renders the correctly typed literal
- Card revision: `1d8d9a7a`

## Change

**Conclusion:** **PASS** — a dup'd `bool` constant now renders `= false` at each sink instead of `int S = 0; X = S;` (CS0029), fixing the reported invalid output with no new fidelity or corpus regression.

### Original source

Acquired with `dotnet-inspect member "Dapper.SqlMapper.Settings" "SetDefaults" --package Dapper@2.1.79 --all -S "Original Source"` (SourceLink C#):

```csharp
public static void SetDefaults()
{
    CommandTimeout = null;
    ApplyNullValues = PadListExpansions = UseIncrementalPseudoPositionalParameterNames = PreferTypeHandlersForEnums = false;
    AllowedCommandBehaviors = DefaultAllowedCommandBehaviors;
    FetchSize = InListStringSplitCount = -1;
}
```

### Before

Acquired with `-S "Decompiled Source"` at the base commit (`origin/main`, 317b8dff):

```csharp
public static void SetDefaults()
{
    Nullable<int> V_0 = default;
    Settings.CommandTimeout = V_0;
    int S_256 = 0;
    Settings.PreferTypeHandlersForEnums = S_256;
    int S_257 = S_256;
    Settings.UseIncrementalPseudoPositionalParameterNames = S_257;
    int S_258 = S_257;
    Settings.PadListExpansions = S_258;
    Settings.ApplyNullValues = S_258;
    Settings.AllowedCommandBehaviors = unchecked((CommandBehavior)(-10));
    int S_259 = -1;
    Settings.InListStringSplitCount = S_259;
    Settings.FetchSize = (long)S_259;
}
```

- Valid: **False**
- Correct: **False**

`Settings.PreferTypeHandlersForEnums = S_256;` assigns `int` to a `bool` member — CS0029 at all four bool sinks, so the method does not compile.

### After

Acquired with `-S "Decompiled Source"` at this PR's head (`1d8d9a7a`):

```csharp
public static void SetDefaults()
{
    Nullable<int> V_0 = default;
    Settings.CommandTimeout = V_0;
    Settings.PreferTypeHandlersForEnums = false;
    Settings.UseIncrementalPseudoPositionalParameterNames = false;
    Settings.PadListExpansions = false;
    Settings.ApplyNullValues = false;
    Settings.AllowedCommandBehaviors = unchecked((CommandBehavior)(-10));
    Settings.InListStringSplitCount = -1;
    Settings.FetchSize = (long)-1;
}
```

- Valid: **True**
- Correct: **True**

Each bool sink now receives the `false` literal directly; the int stack-slot copy chain (`S_256`..`S_259`) is gone, including the `-1` shared by the `int` and `long` sinks. Independent field assignments in any order have the same observable effect as the source chain. Verified compilable against stub types matching the real member types.

### Fully raised

This PR reaches **valid + correct**, not the fully-raised endpoint: it does not
reconstruct the source-level chained assignment (`A = B = C = false;`) — the sinks
render as separate, IL-order statements.

```csharp
CommandTimeout = null;
ApplyNullValues = PadListExpansions = UseIncrementalPseudoPositionalParameterNames = PreferTypeHandlersForEnums = false;
AllowedCommandBehaviors = DefaultAllowedCommandBehaviors;
FetchSize = InListStringSplitCount = (long)-1;
```

- Required tracking issue: #2994 — reconstruct chained/embedded assignment from the dup-of-value idiom (`a = b = c = v`).

## Evidence

| Check | Baseline (317b8dff) | Head (1d8d9a7a) |
| --- | --- | --- |
| Focused test `ChainedConstantAssignmentTests` | 1 failed (0 literal sinks; int spill) | 1 passed |
| Decompiler fast suite (`-trait- "Speed=Slow"`) | 3092, 0 failed | 3093, 0 failed |
| CS0161 census pin `DecompilerAssembly_MissingReturnPopulationIsPinned` | 8 | 8 |
| FidelityGate diff-set (`NoNewFidelityDiffsBeyondKnownDocket`) | 72 entries; 1 fail `StatementBodyLambdaInsideIf` | byte-identical 72 entries; same 1 fail |
| CorpusSweepGate (CoreLib, 41,159 methods) | 0 pass-bugs; floors held | 0 pass-bugs; floors held |
| Real witness `Dapper.SqlMapper.Settings::SetDefaults` | CS0029 (invalid) | valid |

The one FidelityGate failure (`StatementBodyLambdaInsideIf` — a compiler-generated
lambda-cache field ordinal `<>9__100_0` -> `<>9__121_0`; opcode streams identical)
is **pre-existing on `origin/main`**: base and head produce byte-identical full
diff-sets, so this PR introduces zero new fidelity diffs. It is unrelated to
this change (the unchanged declaring-type recompile skeleton assigns lambda
ordinals from sibling ordering) and out of scope here.

## Decompiler quality

**Conclusion:** **PASS** — breadth (CorpusSweepGate: 0 pass-bugs, Full and fully-raised
floors held over 41,159 CoreLib methods) and depth (FidelityGate diff-set byte-identical
to base) both hold; the change can only remove redundant int-slot residue for dup'd
constants, never add it.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.ChainedConstantAssignmentTests"
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.CorpusSweepGateTests"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -5225,3 +5225,19 @@ public class BackingFieldSample
     // Reads no field, so the walk reports nothing.
     public int NoFieldAccess(int x) => x + 1;
 }
+
+// Regression fixtures for #2982: a constant assigned in a chain (`a = b = c = false`)
+// compiles to `ldc.i4.0; dup; call set_C; dup; call set_B; call set_A` — the shared
+// literal is dup'd to each sink. The importer must re-materialize the dup'd constant
+// at each bool sink so it renders `A = false;`, not spill it into an int stack slot
+// (`int S = 0; A = S;`), which is CS0029 (cannot implicitly convert int to bool).
+public static class ChainedConstantAssignmentSamples
+{
+    public static bool A { get; set; }
+
+    public static bool B { get; set; }
+
+    public static bool C { get; set; }
+
+    public static void ChainedBoolFalse() => A = B = C = false;
+}
--- a/src/ILInspector.Decompiler.Tests/ChainedConstantAssignmentTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ChainedConstantAssignmentTests.cs
@@ -1,0 +1,49 @@
+using System;
+using ILInspector.Decompiler.Pipeline;
+
+namespace ILInspector.Decompiler.Tests;
+
+// #2982: a constant assigned through a chain (`a = b = c = false`) is dup'd to each
+// sink in IL (`ldc.i4.0; dup; call set_C; dup; call set_B; call set_A`). The importer
+// must re-materialize the dup'd constant at each bool sink so per-sink constant
+// recovery fires and it renders `= false` — not spill it into an int stack slot
+// (`int S = 0; a = S;`), which is CS0029 (cannot implicitly convert int to bool).
+public class ChainedConstantAssignmentTests
+{
+    static string Render(string methodName)
+    {
+        using var source = MetadataSource.Open(typeof(ChainedConstantAssignmentSamples).Assembly.Location);
+        var function = IrImporter.Import(source, typeof(ChainedConstantAssignmentSamples).FullName!, methodName);
+        Assert.NotNull(function);
+        IrPasses.Run(function!);
+        function!.CheckInvariant();
+        return CSharpPrinter.Print(function).Output!;
+    }
+
+    [Fact]
+    public void ChainedBoolConstant_RendersLiteralAtEachSink_NotIntSpill()
+    {
+        var output = Render(nameof(ChainedConstantAssignmentSamples.ChainedBoolFalse));
+
+        // Each of the three chained sinks receives the bool literal directly.
+        Assert.Equal(3, CountOccurrences(output, "= false;"));
+
+        // The dup'd constant is not spilled into an int stack slot: the #2982 defect
+        // rendered `int S_256 = 0; ...A = S_256;` — CS0029 at every bool sink.
+        Assert.DoesNotContain("int S", output);
+        Assert.DoesNotContain("S_", output);
+    }
+
+    static int CountOccurrences(string haystack, string needle)
+    {
+        int count = 0;
+        int index = 0;
+        while ((index = haystack.IndexOf(needle, index, StringComparison.Ordinal)) >= 0)
+        {
+            count++;
+            index += needle.Length;
+        }
+
+        return count;
+    }
+}
--- a/src/ILInspector.Decompiler/Pipeline/Ir/IrImporter.cs
+++ b/src/ILInspector.Decompiler/Pipeline/Ir/IrImporter.cs
@@ -1367,8 +1367,23 @@ public static class IrImporter
 
                 case ILOpCode.Dup:
                 {
-                    // Trees cannot share nodes: dup materializes through a slot.
                     var value = Pop(stack);
+                    // A pure constant has no identity or side effect to preserve,
+                    // so re-materializing it (a fresh clone per use) keeps the
+                    // tree-node invariant without a spill slot. A shared int slot
+                    // would fix the constant's stack type (int32) and strand each
+                    // consumer's semantic type: a dup'd bool constant threaded
+                    // into `set_X(bool)` sinks renders as `int S = 0; X = S;`
+                    // (CS0029, #2982), because per-sink identity recovery
+                    // (TypedConstantsPass) never reaches the slot store. Cloning
+                    // the constant to each sink lets that recovery fire per use.
+                    if (value is Constant)
+                    {
+                        stack.Push(value);
+                        stack.Push((IrExpression)value.Clone());
+                        break;
+                    }
+                    // Trees cannot share nodes: dup materializes through a slot.
                     int slot = state.NextDupSlot++;
                     body.Add(new StoreStackSlot(slot, value));
                     stack.Push(new LoadStackSlot(slot, value.ResultType));


richlander:

Two independent fixed-head reviews of the exact reviewed head
`1d8d9a7a46280fe8e1cc551c2543aa56f8c8bed7`, each in an isolated worktree with
scratch under `/tmp` only. Both **CLEAN**; no findings required code changes, so
the reviewed head is unchanged.

### Reviewer 1 — Gemini Pro (`gemini-3.1-pro-preview`) — verdict: CLEAN

Built the solution, ran the regression/fast/corpus suites, and built synthetic
IL probes. Refuted every attack vector:

- **Stack correctness:** one pop, two pushes → net `+1`, matching `dup` and the
  prior two-`LoadStackSlot` behavior.
- **Node-sharing / `Clone()`:** `Constant` is a sealed, childless leaf; `Pop`
  detaches the node; `Clone()` (`MemberwiseClone`, cleared parent/child) yields a
  disjoint node. No shared-parent invariant violated.
- **Un-sharing soundness:** a pure `Constant` has no side effects; `string`
  literals (identity-guaranteed) and `null` are indistinguishable, so
  duplication is sound. Address-taking/box shapes are separate node types and
  fall through to the slot path.
- **Slot numbering:** `NextDupSlot` is a synthetic monotonic name counter;
  skipping an increment for constants cannot collide.
- **Over-broad `is Constant`:** `ldtoken`→`LoadToken`, `box`→`Box`; only
  primitive literals are `Constant`.
- **Source-offset fidelity (extra vector):** both clones retain the `ldc`
  offset (not the `dup` offset), *improving* PDB step-mapping.
- **Bounding:** a related `A = B = cond ? false : true;` crosses a block
  boundary, so the value is a `LoadStackSlot` (not `Constant`) at the `dup` — it
  falls through to the slot path and remains a **pre-existing** CS0029, i.e. the
  fix scopes itself to within-block constant duplication.

### Reviewer 2 — GPT (`gpt-5.6-sol`) — verdict: CLEAN

Built the root solution and ran, on head:

- New regression `ChainedConstantAssignmentTests`: 1/1 (also re-run in **Debug**,
  where `CheckInvariant()` is active — passed).
- Fast suite `-trait- "Speed=Slow"`: 3093/3093, 0 failed.
- `CorpusSweepGateTests`: passed.
- CLI IL/decompile probes on the fixture confirm the `ldc.i4.0; dup; call
  set_C; dup; call set_B; call set_A` shape renders `= false` at each sink.

Source-inspected `Constant`/`Clone()`/`BuildState` slot allocator and the
constant-producing opcodes; conclusions match Gemini's. Independently reproduced
the `StatementBodyLambdaInsideIf` FidelityGate failure **on head** (opcode
streams byte-identical; only the synthesized lambda ordinal
`<>9__100_0 → <>9__121_0` differs) and confirmed its `dup` operands are
`ldsfld`/`newobj` (not `Constant`), so this PR's new branch cannot execute for
that method.

Transparency note (GPT): to honor head isolation it did **not** execute the
FidelityGate at base `317b8dff`, verifying the pre-existing claim structurally
(static `git show`/`diff`) rather than by a base real-run.

### Author-supplied base real-run (completes GPT's note)

The "Baseline" column of the Evidence table is backed by a separate base
worktree at `317b8dff`: the FidelityGate produces the **byte-identical** 72-entry
diff-set and the **same** `StatementBodyLambdaInsideIf` failure with the fix
un-built, confirming this PR introduces **zero** new fidelity diffs and the gate
failure is pre-existing on `main` (unrelated lambda-ordinal artifact,
out of scope; endpoint tracked at #2994 is a distinct raise slice).

### Outcome

Two clean fixed-head reviews from two different model families (Gemini Pro, GPT;
neither the author's own model). No blocking findings, no code changes, reviewed
head unchanged.

richlander:
Ready to merge

- Two clean fixed-head adversarial reviews (Gemini Pro, GPT) on `1d8d9a7a` — no blocking findings, no code changes.
- CI green: build-net10, test (ubuntu-26.04/linux-x64), csharp-diff-smoke, changes all pass (il-diff-smoke/markdownlint/pack are path-gated skips).
- Zero new fidelity diffs vs base; CorpusSweepGate 0 pass-bugs, floors held.
- Follow-up (non-blocking): fully-raised chained-assignment reconstruction tracked at #2994.

Issue #2964, opened 2026-07-21
RTS harness: compile-back skeleton emits extension 'this' before parameter attributes (EVIL CS1031, 97 rows)


richlander:
## Summary

The RTS compile-back **skeleton reconstruction** emits an extension method's `this` receiver modifier **before** the parameter's attribute list — `this [Attr] Type name` instead of the legal `[Attr] this Type name`. The C# parser sees `this` followed by `[`, expects a type, and fails with **CS1031** ("Type expected"). This is a **harness (ReturnToSender) skeleton-emission defect, not a product decompiler defect** — the decompiled target body is clean C#.

**EVIL Invalid rows affected: 97.**

## Evidence (authoritative harness repro, current `origin/main`)

Driving `ReturnToSender.CompileBackTargets` on `Hangfire.BackgroundJob.ContinueJobWith` (overload 3), package `Hangfire.Core@1.8.24`:

```text
STATUS: RecompileFail
DETAIL: closure-stalled: CS1031: Type expected

#### TargetBody (product decompilation — CLEAN) ####
return BackgroundJob.ClientFactory.Invoke().ContinueJobWith(parentId, queue, methodCall, null, options);

#### Source (reconstructed skeleton — MALFORMED) ####
public static string ContinueJobWith(this [Hangfire.Annotations.NotNull] IBackgroundJobClient client, ...) { throw null; }
//                                   ^^^^ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//                                   `this` emitted BEFORE the attribute list => CS1031 "Type expected"
```

Legal C# requires `[Hangfire.Annotations.NotNull] this IBackgroundJobClient client` (attributes first, then the `this` modifier, then the type).

The discriminator is `ReturnToSender.Result.TargetBody` (the target's own decompiled body — clean) vs `Result.Source` (the full composed unit incl. skeletons — where the malformed C# lives).

## Affected packages (from EVIL benchmark, 97 rows)

| Source | Rows | Package |
| --- | --- | --- |
| `BackgroundJob.cs`, `BackgroundJobClientExtensions.cs`, `RecurringJobManagerExtensions.cs`, `RecurringJob.cs`, `BackgroundProcessExtensions.cs`, `RecurringJob*` | ~75 | Hangfire.Core 1.8.24 |
[...]

PR #3002, opened 2026-07-22
Fix annotated extension receiver emission order (#2964)

richlander:
- Advances #2964
- Changes the extension-receiver `this` modifier to be inserted **after** the parameter's leading attribute lists (`[Attr] this T x`, not `this [Attr] T x`), fixing CS1031 in reconstructed extension-method signatures
- Evidence revision: `cdb032e1`

## Change

**Conclusion:** **PASS** — the RTS compile-back skeleton emitted `this` before a receiver parameter's attribute list, producing `this [Attr] Type name`, which the C# parser rejects with CS1031 ("Type expected"). Inserting `this` after the leading attribute lists yields legal `[Attr] this Type name`, unblocking compile-back of annotated extension surfaces (the EVIL benchmark counted ~97 such rows as Invalid).

Before (reconstructed skeleton — malformed, CS1031):

```text
public static string ContinueJobWith(this [Hangfire.Annotations.NotNull] IBackgroundJobClient client, ...) { throw null; }
//                                   ^^^^ `this` before the attribute list => CS1031 "Type expected"
```

After (`this` placed after the attribute list — legal C#):

```text
public static string ContinueJobWith([Hangfire.Annotations.NotNull] this IBackgroundJobClient client, ...) { throw null; }
```

## Scope and safety

- **Root cause:** the extension-receiver renderer prepended a `this` modifier at the raw first-parameter start, but the structured signature renderer had already attached the parameter's attribute lists there, so `this` landed before `[Attr]`.
- **Fix shape:** in `CSharpDeclarationWriter`, skip over every leading `[...]` attribute list (via a new `MatchingAttributeList` bracket matcher that respects string/char literals) before inserting the `this` modifier. The literal-aware matcher is why the second commit handles a `]` inside an attribute argument.
- **Product impact:** yes — `CSharpDeclarationWriter` is product code (`src/ILInspector.CSharp`). Any product rendering of an extension method whose receiver carries parameter attributes now emits the legal `[Attr] this T` order. This is a rendering-correctness improvement; no other declaration shape changes.
- **Scope boundary:** parameters without leading attributes are unaffected (the loop is a no-op), and a signature already beginning with a `this` modifier after its attributes is left as-is.
- **RTS boundary:** RTS only orchestrates closure + Roslyn + skeleton composition; the malformed C# came from product signature rendering, and the fix stays in the product renderer rather than adding C# knowledge to RTS.

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass |
| Focused tests | `ReturnToSenderFixtureCatalogTests.ReturnToSenderSourceProbe_PlacesReceiverAttributesBeforeExtensionThisModifier` passed at head |
| Real witness (before) | same probe run against base product code: fails with `RecompileFail ... CS1031: Type expected` |
| Real witness (after) | head: no CS1031; `Source` contains `] this string value` and never `this [` |
| Motivating population | ~97 EVIL Invalid rows (Hangfire.Core 1.8.24, StackExchange.Redis 3.0.17) per #2964; unblocked, not re-measured here |

The self-contained fixture uses a `[Receiver("]")]` attribute whose argument contains a literal `]`, exercising the literal-aware bracket matcher directly.

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| Full 97-row EVIL re-measurement | Not run here | Focused fixture proves the emission fix; broad re-run is corpus bookkeeping, not a correctness gate for this change |

## Build-event validation

**Conclusion:** not applicable — validated by the focused compile-back probe; no build-health claim.

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus | No blocking findings | reviewed head `cdb032e1` |
| Gemini Pro | No blocking findings | reviewed head `cdb032e1` |

**Review conclusion:** **PASS** — both fixed-head reviews clean; see PR review comments for reconciliation.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -method "*ReturnToSenderSourceProbe_PlacesReceiverAttributesBeforeExtensionThisModifier*"
```

### DIFF
--- a/src/ILInspector.CSharp/CSharpDeclarationWriter.cs
+++ b/src/ILInspector.CSharp/CSharpDeclarationWriter.cs
@@ -1431,7 +1431,7 @@ internal static class CSharpDeclarationWriter
         var modifierStart = firstParameterStart;
         while (modifierStart < signature.Length && signature[modifierStart] == '[')
         {
-            var close = Matching(signature, modifierStart, '[', ']');
+            var close = MatchingAttributeList(signature, modifierStart);
             if (close < 0)
                 return signature;
             modifierStart = close + 1;
@@ -1443,6 +1443,45 @@ internal static class CSharpDeclarationWriter
             return signature;
 
         return signature.Insert(modifierStart, "this ");
+    }
+
+    static int MatchingAttributeList(string text, int start)
+    {
+        var depth = 0;
+        char quote = '\0';
+        bool escaped = false;
+        for (var i = start; i < text.Length; i++)
+        {
+            char c = text[i];
+            if (quote != '\0')
+            {
+                if (escaped)
+                {
+                    escaped = false;
+                    continue;
+                }
+                if (c == '\\')
+                {
+                    escaped = true;
+                    continue;
+                }
+                if (c == quote)
+                    quote = '\0';
+                continue;
+            }
+
+            if (c is '\'' or '"')
+            {
+                quote = c;
+                continue;
+            }
+            if (c == '[')
+                depth++;
+            else if (c == ']' && --depth == 0)
+                return i;
+        }
+
+        return -1;
     }
 
     static string AbbreviateSignature(string signature)
--- a/src/ILInspector.CSharp/CSharpDeclarationWriter.cs
+++ b/src/ILInspector.CSharp/CSharpDeclarationWriter.cs
@@ -1424,10 +1424,25 @@ internal static class CSharpDeclarationWriter
         while (firstParameterStart < signature.Length && char.IsWhiteSpace(signature[firstParameterStart]))
             firstParameterStart++;
 
-        if (signature.AsSpan(firstParameterStart).StartsWith("this ".AsSpan(), StringComparison.Ordinal))
+        // Parameter attributes precede the extension receiver modifier in C#:
+        // [NotNull] this string value. The structured signature renderer has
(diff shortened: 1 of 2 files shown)


richlander:

Two-model review (Claude Opus + Gemini Pro), each isolated in a separate exact-head worktree (`cdb032e1`), real build+test evidence required.

- **Claude Opus — CLEAN.** Verified `AddExtensionThisModifier` only ever *inserts* `this ` after leading attribute lists on the first (receiver) parameter; non-receiver params and non-extension methods untouched; `[Attr] this ref T` ordering correct; `MatchingAttributeList`'s escape state machine matches `AttributeReader`'s literal emitter; safe fallbacks on unbalanced/whole-string attributes. RTS catalog suite 84/84 green.
- **Gemini Pro — CLEAN.** Independently confirmed scoping, modifier combos, literal parsing (noted raw/verbatim strings are never emitted into this parser), and spacing edge cases (`[Attr]this` compiles). One **non-blocking** suggestion: add a second ordinary attributed parameter to the probe to prove non-receiver params are untouched.

**Net: two clean reviews.** The Gemini test-coverage nit is optional and non-blocking.

richlander:
Ready to merge — two clean adversarial reviews (Claude Opus + Gemini Pro) at the exact head with real build+test evidence.

Issue #2958, opened 2026-07-21
Pre-existing: LadderIterator recovered-rows compile-back is OperandDiff, not Exact


richlander:
## Pre-existing slow-suite failure: `LadderIteratorGateTests.IteratorFixture_RecoveredRowsCompileBackOpcodeExact`

`IteratorFixture_RecoveredRowsCompileBackOpcodeExact` (`Speed=Slow`) fails on `main` with:

```
Assert.Equal() Failure: Values differ
Expected: Exact
Actual:   OperandDiff
  at LadderIteratorGateTests.cs(270)
```

One of the recovered iterator members compile-backs to an `OperandDiff` rather than the pinned `Exact` opcode status.

### Attribution / evidence

- Reproduced on `main` (base `65b114aa`) independent of any in-flight change.
- Confirmed **not** caused by PR #2950 via a causation test: disabling `FluentChainRecompositionPass` and re-running reproduces the identical `Exact → OperandDiff` failure. The fixture contains no fluent chains, so neither the recomposition pass nor the whitespace-only printer wrap can affect it.

Likely an operand-level codegen drift on the daily SDK. This issue tracks restoring the iterator rung to zero opcode diffs (or re-pinning if the drift is an accepted daily-SDK delta).

richlander:
Resolved on `main` by PR #3004 (commit `b6abb324`, "Fix iterator property compile-back fidelity").

The `OperandDiff` was in the harness compile-back path: `FidelityCheck` was not emitting iterator property accessors (e.g. `IteratorPoint::InstanceSequence` / `get_Accessor`) in their original metadata declaration order, so the recompiled fixture drifted at the operand level. PR #3004 adds `EmitTargetProperty` and preserves target property declaration order, plus a regression test (`FidelityCheckGeneratedFilterTests.Evaluate_PreservesIteratorPropertyDeclarationOrder`).

Verified on current `main` (`d6af0165`, SDK 11.0.100-preview.7.26366.102):

- `IteratorFixture_RecoveredRowsCompileBackOpcodeExact` (`Speed=Slow`) passes — ran 4× with zero failures.
- Full `LadderIteratorGateTests` class: Total 5, Failed 0.

Closing as fixed.

PR #3004, opened 2026-07-22
Fix iterator property compile-back fidelity (#2958)

richlander:
- Advances #2958
- Changes harness-only: `FidelityCheck` emits a target property once at its first-accessor metadata position (preserving declaration order), so `LadderIterator` recovered-rows compile-back classifies `Exact` instead of `OperandDiff`
- Evidence revision: `59cd6466`

## Change

**Conclusion:** **PASS** — a recovered iterator property was being reconstructed out of its declared order in the compile-back skeleton, perturbing operands; emitting it in declaration order restores `Exact`, and the pinned gate `IteratorFixture_RecoveredRowsCompileBackOpcodeExact` flips from failing to passing with no product-output change.

Before (base `2269483d`, gate test):

```text
ILInspector.Decompiler.Tests.LadderIteratorGateTests.IteratorFixture_RecoveredRowsCompileBackOpcodeExact [FAIL]
  Expected: Exact
  Actual:   OperandDiff
  LadderIteratorGateTests.cs(270)
```

After (head `59cd6466`, same gate test):

```text
ILInspector.Decompiler.Tests  Total: 1, Errors: 0, Failed: 0, Skipped: 0
  IteratorFixture_RecoveredRowsCompileBackOpcodeExact  -> Exact
```

## Scope and safety

- **Root cause:** target properties whose accessors are compile-back targets but are not auto-properties were emitted by a separate stub path, losing their original metadata declaration position relative to sibling members; the reordered skeleton changed operand indices, so the recovered iterator rung recompiled to `OperandDiff`.
- **Fix shape:** harness skeleton-emission change in `FidelityCheck.cs` — record such target properties in `orderedTargetProperties` and emit each once (`EmitTargetProperty`) at its first accessor's metadata position, so declaration order is preserved.
- **Product impact:** none. The change is entirely in `tools/DecompilerHarness`; product decompiler output is unchanged.
- **Scope boundary:** `requireAutoProperty` structs and `RequiresAsync` accessors still fall back to the method-emission path (unchanged), so async/auto-property shapes are not affected.
- **RTS boundary:** not applicable — this is the fidelity compile-back skeleton, not the ReturnToSender closure path.

## Evidence

| Check | Result |
| --- | --- |
| Product build | Pass (harness-only change; product unchanged) |
| Focused tests | `LadderIteratorGateTests.IteratorFixture_RecoveredRowsCompileBackOpcodeExact` passed at head, failed at base |
| Decompiler fast suite | Pass (product paths unchanged) |
| Targeted compile-back | `LadderIterator` recovered iterator property `OperandDiff` -> `Exact` |
| Real witness | `LadderIterator` recovered-rows property moved `OperandDiff` -> `Exact` |

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| `requireAutoProperty` struct properties | Not changed | Still routed to the auto-property/method fallback; out of scope for this ordering fix |
| `RequiresAsync` accessor properties | Not changed | Deliberately kept on the method-emission fallback path |

## Build-event validation

**Conclusion:** not applicable — harness-only change validated by the focused compile-back gate, no build-health claim.

## Review

| Reviewer | Result | Notes |
| --- | --- | --- |
| Claude Opus | No blocking findings | reviewed head `59cd6466` |
| Gemini Pro | No blocking findings | reviewed head `59cd6466` |

**Review conclusion:** **PASS** — both fixed-head reviews clean; see PR review comments for reconciliation.

## Validation

```bash
dotnet build src/ILInspector.Decompiler.Tests -c Release --nologo --verbosity quiet
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -method "*IteratorFixture_RecoveredRowsCompileBackOpcodeExact*"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/FidelityCheckGeneratedFilterTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FidelityCheckGeneratedFilterTests.cs
@@ -10,6 +10,38 @@ namespace ILInspector.Decompiler.Tests;
 public class FidelityCheckGeneratedFilterTests
 {
     [Fact]
+    public void Evaluate_PreservesIteratorPropertyDeclarationOrder()
+    {
+        var assemblyPath = CompileFixture("""
+            using System.Collections.Generic;
+
+            public class IteratorPropertyFixture
+            {
+                public IEnumerable<int> Before() { yield return 1; }
+                public IEnumerable<int> Values { get { yield return 2; } }
+                public IEnumerable<int> After() { yield return 3; }
+            }
+            """);
+        try
+        {
+            var results = FidelityCheck.Evaluate(assemblyPath)
+                .Where(result => result.Type == "IteratorPropertyFixture")
+                .ToList();
+
+            foreach (var method in new[] { "Before", "get_Values", "After" })
+            {
+                var result = Assert.Single(results, result => result.Method == method);
+                Assert.True(result.Status == FidelityCheck.CompileBackStatus.Exact,
+                    $"{method}: {result.Status}: {result.Detail}");
+            }
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
     public void Evaluate_SkipsGeneratedCodeTypesAndMethods()
     {
         var assemblyPath = CompileFixture("""
--- a/tools/DecompilerHarness/FidelityCheck.cs
+++ b/tools/DecompilerHarness/FidelityCheck.cs
@@ -2706,14 +2706,23 @@ static class FidelityCheck
         // readonly receiver can introduce a defensive copy and change opcodes, but
         // an auto-property preserves the field-backed accessor shape.
         var stubPropertyAccessors = new HashSet<MethodDefinitionHandle>();
+        var orderedTargetProperties = new Dictionary<MethodDefinitionHandle, PropertyDefinitionHandle>();
         if (kind is TypeKind.Class or TypeKind.Struct)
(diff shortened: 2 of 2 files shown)


richlander:

Two-model review (Claude Opus + Gemini Pro), each isolated in a separate exact-head worktree (`59cd6466`), real build+test evidence required. Primary attack point: **harness-boundary** (does a harness-only change mask a product defect?).

- **Claude Opus — CLEAN.** Reproduced the root cause: the operand comparison is name-based, and the only operand that shifts with member order is the compiler-generated iterator state-machine name (`<get_Values>d__2` vs `d__1`). `main`'s `EmitStubProperties` hoisted the target property out of metadata position, shifting the ordinal and producing a *false* `OperandDiff`; the fix emits the property at its true metadata position. Confirmed this cannot turn a genuinely-different body into a false `Exact` (body content is compared independently of position). `FidelityCheckGeneratedFilterTests` 26/26, `LadderIteratorGateTests` 5/5 green.
- **Gemini Pro — CLEAN.** Same root-cause analysis (property = 2 member-index slots vs method = 1, causing an off-by-one `d__N` cascade). Verified single-emission of the property, the `RequiresAsync` method fallback, and that the new test is a real end-to-end integration check (3 iterators through full CompileBack), not a tautology.

**Net: two clean reviews.** Confirmed this is a legitimate harness oracle-fidelity fix, not product-gap masking.

richlander:
Ready to merge — two clean adversarial reviews (Claude Opus + Gemini Pro) at the exact head with real build+test evidence.

PR #3006, opened 2026-07-22
Curate the library -D catalog with an @All/@Hidden taxonomy

richlander:
## What

PR1 of two. Replaces the organically-grown section-visibility flags for the **library** view with a principled curated-catalog taxonomy on the shared `SectionPipeline`, opted into only by `LibrarySections.CreatePipeline()` via `.UseCuratedCatalog()`. Type/member/package-level/diff pipelines are **untouched** (PR2 migrates everyone and removes the transitional flag). Package's embedded library rendering reuses this factory, so it inherits the same curated library view.

## Model (curated pipelines only)

- **New per-section flag `Noisy`** — cheap-but-verbose, never auto-rendered (also `ExplicitOnly`), yet an `@All` member so it stays discoverable at the `-D` top level rather than hiding behind a category door.
- `IsAllMember(entry) = !IsExpensive && (!ExplicitOnly || Noisy)` — flag-based, model-independent.
- **`@All`** = the visible pole (Default + Terse + Noisy). The `-D` top-level catalog = exactly the `@All` members + real category doors.
- **`@Hidden`** = the computed complement: selectable sections surfaced by no listed category. Reachable only via `--schema` or exact name; never hand-authored, never a bare-`-D` category row.
- **`--schema`** stays the exhaustive escape hatch — reveals every section plus the `@Hidden` row.

## Library re-flagging

- New categories: `@Surface` (Async Methods, Custom Attributes, Extension Methods, Type Forwarders, Union Types, P/Invoke Methods), `@Resources`, `@Source` (Source Files, SourceLink Availability, SourceLink Missing Files).
- SourceLink Integrity + Top Leverage + Non-normalized Paths + the IL-offset context sections fall into the computed `@Hidden` pole.

### Behavior changes (2, both intended)

1. **P/Invoke Methods -> opt-in** (was default-rendered). Now `Noisy` + `ExplicitOnly`: an `@All` member (discoverable at `-D` top level, under `@Surface`/`@Audit`) but no longer in the default document.
2. **Non-normalized Paths -> `@Hidden`** (was default-rendered). Niche path-hygiene audit, reachable via `--schema` or exact name only.

### Notes for reviewers

- **Top Leverage went to `@Hidden`, not `@Performance`.** Adding it to `@Performance` breaks the #2833 group-tabular "one shared view" invariant (`OutputFormatter.WriteLibraryTabular` collapses `-S @Performance --table` into a single table only when all selected sections share a common view; Top Leverage's view differs). Happy to add a dedicated `@Leverage` door or a Terse promotion instead if preferred.
- **`--schema` is now fully exhaustive** — it also lists the kind-scoped `Performance: *` sections and the `@Hidden` row (previously hidden from `--schema` too). Intended: schema is the escape hatch.
- **IL-offset coordinate sections** are excluded from the *effective* `-D` catalog-hidden set so they remain discoverable at the top level exactly when a coordinate makes them applicable, while staying `@Hidden`-grouped for `--schema`/`-S`.

## Verification

Empirically checked on `System.Text.Json` (cached platform) and `Newtonsoft.Json`:

- bare `-D` = `@All` members + real doors only (no `@Hidden` row, no opt-in spam)
- `-D @Source` lists Availability/Missing/Files; `-D @Hidden` lists Integrity + Top Leverage
- `-D --schema` reveals all sections + `@Hidden`
- Full CLI test suite green apart from the 2 pre-existing Darwin-only P/Invoke content tests (`Interop.User32`), unrelated to this change.

Design spec: `docs/design/section-model.md`.

## Follow-up

PR2: migrate type/member/package-level/diff pipelines to the curated model and remove the transitional `UseCuratedCatalog()` flag.

## `@Hidden` is discovery-only (adversarial-review outcome)

`@Hidden` is a computed pole for niche and footgun-expensive standalone sections
(SourceLink Integrity, Top Leverage). It is reachable for **discovery** (`-D
@Hidden`, `--schema`) and its members **render by exact name** (`-S "SourceLink
Integrity"`), but `-S @Hidden` is **rejected** so it can never fan out to the
unbounded SourceLink Integrity HEAD pass as a group. The guard applies to the
direct `library` command and both package embedded-library modes, and fires
before any package acquisition. See the reconciliation comment for the two-model
review (GPT-5.5 + Gemini 3.1 Pro, both clean at `5215d015`).

### DIFF
--- a/docs/design/section-model.md
+++ b/docs/design/section-model.md
@@ -32,8 +32,11 @@ different function than a curated `@Performance`-style door:
 - **`@Hidden`** — the invisible pole, and the **computed complement** of `@All`.
   A section falls into `@Hidden` exactly when no *listed* category surfaces it.
   You never author `@Hidden` membership; it falls out. `@Hidden`'s own name is
-  not listed in `-D`; its members are reached only via `--schema` or an exact
-  request (`-S @Hidden`, `-D @Hidden`).
+  not listed in `-D`; its members are reached only via `--schema`, discovery
+  (`-D @Hidden`), or an exact-name request (`-S "SourceLink Integrity"`).
+  `@Hidden` is **discovery-only**: it is not a render selector, so `-S @Hidden`
+  is rejected. This guarantees its footgun-expensive members (SourceLink
+  Integrity) can never be executed as a group — only by exact name or `-v:d`.
 
 `--schema` is the union: `@All ∪ @Hidden` = the full section graph.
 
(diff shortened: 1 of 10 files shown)


richlander:
## Adversarial review reconciliation (2 models, clean at head `5215d015`)

Per AGENTS.md, two different models reviewed this PR in isolated worktrees
(scratch under `/tmp`, no git mutations). My model is Opus 4.8, so the reviewers
were **GPT-5.5** and **Gemini 3.1 Pro**. Both are now **clean at the final head
`5215d015`**.

### Findings and resolutions

**GPT-5.5**
1. *(Withdrawn)* `-S @All` renders Noisy sections — first flagged Medium, then
   withdrawn after review against `docs/design/section-model.md`: this is the
   intended `@All` noise axis (Noisy renders via `-S @All`, never via the `-v`
   ladder). GPT verified `-S @All` excludes expensive/feeder/network sections.
2. *(Not a regression)* `-S @All --count` triggers network. Root-caused to the
   default `@All` member **Signals** (SourceLink HEAD audit) + **Symbols** (PDB),
   in the untouched `LibrarySourcePlan.cs`. Confirmed head is strictly cheaper
   than base (`-S @All --count`: base 709 HTTP → head 354). GPT agreed it is not
   a blocker, but that the spec **overclaimed** "@All < 1s / no network" and that
   `-S @Hidden` expanding to unbounded SourceLink Integrity was a **new footgun**.
   → **Resolved (option B, `3b774581`):** corrected the cost spec (discovery is
   cheap; render selection executes members) and made `@Hidden` **discovery-only**
   — `-S @Hidden` is rejected; discovery (`-D @Hidden`/`--schema`) and exact-name
   render (`-S "SourceLink Integrity"`) remain.
3. *(High, fixed)* `package --all-libraries -S @Hidden` bypassed the guard (it
   resolves `-S` against the curated pipeline itself). → **Resolved (`36aa2a11`):**
   extracted the guard into `LibraryCommand.RejectHiddenRenderSelector` and called
   it from the package path.
4. *(Medium, fixed)* The package guard rejected *after* package acquisition
   (~2 HTTP). → **Resolved (`5215d015`):** moved the guard to the top of
   `PackageCommand.ExecuteAsync` (gated on `packageLibraryMode`), before any
   fetch. GPT re-verified: rejects in ~73ms with zero HTTP; exact-name still
   renders; non-`@Hidden` package behavior unchanged. **Clean.**

**Gemini 3.1 Pro** — clean at every reviewed head (`86972aa9`, `3b774581`,
`5215d015`). Verified with real runs: non-curated pipelines isolated, `@Hidden`
computed correctly, IL-coordinate subtraction correct, `--schema`/`--tree`
exhaustive, expensive excluded from `@All`, package embedded view curated, guard
case-insensitive and compound-selector safe, zero HTTP before rejection, and the
`*Discover*` suite (87 tests) plus both new guard tests pass. **Clean.**

### Behavior / compatibility boundary
- `@Hidden` is now **discovery-only**: `-S @Hidden` errors with a message
  pointing to `-D @Hidden` / `--schema` / exact-name render. This applies to the
  direct `library` command and both package embedded-library modes.
- `LibrarySourcePlan.cs` is untouched; `@All` network cost is pre-existing and
  net-reduced by the curated taxonomy.

### Validation
Full CLI suite: 1981 total, 0 unexpected failures (only the 2 pre-existing
Darwin P/Invoke content tests, which assert Windows `Interop.User32` content and
are unaffected by this change). New tests:
`LibraryCommand_RenderSelectHidden_IsRejectedAsDiscoveryOnly`,
`PackageCommand_AllLibraries_RejectsHiddenRenderSelector`.

richlander:
Ready to merge

Issue #3007, opened 2026-07-22
RTS: coherent Full reconstruction of plain event-accessor targets


richlander:
Follow-up from #3000.

Under the opt-in `Full` body policy, when the RTS target is a **plain (non-explicit-interface) event accessor** (e.g. `add_Changed`), the accessor is method-routed and rendered as a standalone method. The reconstructed type surface re-declares the `event`, whose synthesized accessor collides (CS0082: "already reserves a member called 'add_Changed'") with the standalone target method. The Full closure fails to recompile and falls back to the single-target compile-back floor.

#3000 makes this **honest**: `BodyComplete` now returns false when the floor is used, and a regression test (`CompileBackTargets_FullEventAccessorTargetSurfacesRecompileFailure`) pins that behavior. But the target is not actually reconstructed at Full fidelity.

## What a fix needs

Prototyping showed that routing plain event accessors through `ComposeEventAccessor` under `Full` eliminates the CS0082 collision and produces correct source (`event Action Changed { add {...} remove {...} }` with both real bodies). However the `ApplyFullBodies` evidence enumeration then reports the sibling accessor (`remove_Changed`) and the type `.ctor` as "concrete but not represented in the typed artifact", because:

- The property target requirement (`TargetGetterWithSetter`) carries **both** accessor tokens, so `Enrich`'s property branch credits the sibling and the target. The event target requirement (`TargetEventAccessorWithSibling`) carries only the target accessor token; the sibling is baked in via `SiblingTargetBody` and never enters `attempted`.
- The property-target type surface includes the `.ctor` (Produced → Complete); the event path's `ComposeEventAccessor` does not emit the same member surface, so `.ctor` is flagged.

So a coherent fix requires event↔property **symmetry**: the event target requirement should carry both accessor tokens (so `Enrich` credits the sibling like the property fix does), and the event-target type surface should include the constructor/member surface the property path already includes. This is a deeper `TypeProducer`/target-requirement change with corpus-regression risk, scoped out of #3000.

## Acceptance
- Plain event-accessor target under `Full` reconstructs `event { add remove }` with both real bodies, recompiles Exact, and `BodyComplete` is honestly true.
- No new floor fallbacks or corpus regressions across the RoundTrip area.
- Convert `CompileBackTargets_FullEventAccessorTargetSurfacesRecompileFailure` to assert successful reconstruction.

PR #3001, opened 2026-07-22
C# assembly round-trip testing (tools-only)

richlander:
# C# assembly round-trip testing (tools-only)

## Summary

Adds a **tools-only** capability to reconstruct typed C# from a managed
assembly, compile it with Roslyn, and compare the rebuilt assembly against the
original through the repository's existing C# and IL diff tools. The goal is
**evidence, not mutation**: no PE patching, no byte-for-byte claims, and no
Roslyn or inspected-assembly loading in shipped product paths.

Design: `docs/design/csharp-member-recompilation.md`.

Two reconstruction scopes (`cluster`, `all`) and two body policies (`selected`,
`full`) are supported. Artifact / Compile / C# comparison / IL comparison remain
separate, layered results; existing diff tools stay the arbiters within their
documented boundaries.

## Product surface (SRM-only, Roslyn-free, layer-respecting)

- `MetadataMethodAddress` (MetadataPrimitives): a module-scoped
  `(MVID, MethodDefinitionHandle)` address so a borrowed handle cannot be misused
  against another reader.
- `MethodCorrespondenceResolver` (Metadata): total cross-reader correspondence
  for one method via a dedicated **structural** cross-module identity
  (`MethodStructuralSignature`) — not display names or metadata row numbers.
  The structural key preserves return type, calling convention and the instance
  bit, generic arity encoded positionally (rename-invariant), custom modifiers,
  byref, and the nested-versus-namespace boundary. Returns
  Exact/Absent/Ambiguous/Failed.
- `MemberBodyProducer.ProduceBody` / `ResolveCorrespondence` (Decompiler):
  produces the typed C# body increment for one metadata-addressed method and
  resolves correspondence across live sources, keeping failures visible
  (Complete/Absent/Failed).
- `ApiSurface`: event add/remove accessor tokens (`AdderToken`/`RemoverToken`,
  `[JsonIgnore]`) for correspondence over explicit interface events.

## Tools-only surface

- `tools/RoundTripCompilation/*`: reusable Roslyn compilation engine, contracts,
  and cluster/all scope comparison.
- `tools/DecompilerHarness/ReturnToSender*`: harness consumer of the product
  producers and the round-trip engine (no product reconstruction in the harness).

## Testing / governance

- New `RoundTripCompilationEngineTests`, `RoundTripContractsTests`,
  `RoundTripComparisonTests`, `MemberBodyProducerTypedBodyTests`,
[...]

PR #3008, opened 2026-07-22
Fix RTS Full-body evidence integrity for accessor targets (#3000)

### LINKED ISSUE #3000: RTS Full body policy: sibling accessor body dropped but reported Complete when target is an accessor
## Summary

In the opt-in `Full` body policy (`RoundTripBodyPolicy.Full`), when the ReturnToSender target method is a **property or event accessor**, the sibling accessor's successfully-produced full body is silently discarded (it falls back to a `throw null;` stub), yet it is still recorded as `MemberBodyProductionStatus.Complete`. As a result `Result.BodyComplete` reads `true` even though a reconstructed body was dropped — a "keep failure visible" violation in the full-body production evidence.

The downstream round-trip comparison still catches the resulting body difference, so this is an **evidence-integrity / reporting** defect, not a silent round-trip pass.

## Root cause

`tools/DecompilerHarness/ReturnToSenderTypePlanner.cs`, `ApplyFullBodies` → local `Enrich`/`Produce`:

- `Produce(token)` special-cases the target method: `if (handle == artifact.TargetMethod) return (null, MemberBodyProductionStatus.Complete);` (the target body is applied separately by the base `Compose*` path via `CompileBackStubBodyKind.TargetGetterWithSetter` / `TargetSetterWithGetter` / `TargetEventAccessorWithSibling`).
- The property/event readiness gate computes `getterReady = !concreteGetter || getter.Body is not null;` (and the setter/adder/remover analogues). When the target IS one accessor, `Produce` returns a `null` body, so that accessor is treated as **not ready**, the `getterReady && setterReady` gate fails, and the member policy that would carry the *sibling's* produced body is never applied. The sibling therefore keeps its base `throw` stub.
- Meanwhile `Produce` still appends a `FullBodyProduction(..., produced.Status /* Complete */, ...)` row for the sibling, so `FullBodies` (and thus `BodyComplete`) report success.

## Reproduction (from adversarial review)

Targeting `get_Value` with `RoundTripBodyPolicy.Full` on:

```csharp
public class Holder
{
    private int _v;
    public int Value
    {
        get => _v;
        set { _v = value; }
    }
}
```

produced:

```text
=== FullBodies:
  Holder.get_Value: Complete
  Holder.set_Value: Complete      <-- reported Complete
  Holder..ctor: Complete
=== Source:
    public int Value
    {
        get { return _v; }
        set { throw null; }        <-- sibling body actually dropped
    }
=== Comparison members:
  token=06000002 (set_Value) CSharpStatus=Changed IlStatus=OpcodeDiff
```

## Suggested fix

Make the readiness gate target-aware: treat the target accessor as *ready* (its body is applied by the base `Compose*` path, not by `Enrich`) so the sibling's produced body is incorporated, while **preserving** the target accessor's already-applied body from the existing `MemberPolicyOverrides` entry rather than rebuilding the property/event body with a `null` accessor (which would clobber the target body). Apply the same treatment to the property getter/setter and event adder/remover branches. Add a regression exercising `Full` policy with a property-accessor target and a concrete-bodied sibling, asserting both the emitted sibling body and the `FullBodyProduction` evidence agree.

## Scope / severity

- Tools-only (`tools/DecompilerHarness`), opt-in `Full` policy, narrow to property/event accessor targets; default `Selected` policy is unaffected.
- No product (`src/`) soundness impact; the final comparison remains honest.

Found during adversarial round-3 review of PR for `design/csharp-member-patching` (GPT-5.6). Split out as a focused follow-up per maintainer decision.

richlander:
## Summary

Fixes #3000 — an RTS (ReturnToSender) `Full`-body evidence-integrity bug where a property-accessor target silently dropped the sibling accessor's produced body (kept a `throw null;` stub) yet still reported it `Complete`, so `BodyComplete` falsely read `true`. A "keep failure visible" violation surfaced during adversarial review of #3001.

## Root cause

In `ApplyFullBodies`'s `Enrich` loop, the readiness gate `getterReady = !concreteGetter || getter.Body is not null` was `false` for the **target** accessor — it is concrete but has a null produced body here because the target body is applied by the base `Compose*` path, not by `Produce`. So `getterReady && setterReady` failed, the sibling's produced body was never applied, and the base `throw` stub survived — while a `Complete` `FullBodies` row was still recorded.

## Fix

- **`ReturnToSenderTypePlanner.cs`** — property branch of `Enrich` is now target-aware: it detects `getterIsTarget`/`setterIsTarget` (new `IsTargetToken` helper), skips `Produce` for the target accessor, treats it as ready, and **preserves** the base target-accessor body instead of clobbering it with null, so the sibling's produced body is incorporated. Behavior is unchanged when neither accessor is the target.
- **`ReturnToSender.cs`** — `BodyComplete` now also requires `!UsedCompileBackFloor`, so a `Full` closure that fails to recompile and falls back to the single-target floor is reported honestly (`false`) rather than masked as `Complete`. This surfaces the plain event-accessor target case (a CS0082 method/event-surface collision).

## Scope / follow-up

Coherent `Full` reconstruction of **plain event-accessor** targets requires event↔property symmetry in the target requirement (carry both accessor tokens) and TypeProducer's ctor/member surface — a deeper `TypeProducer` change with corpus-regression risk. Scoped out as **#3007**. This PR makes that case honest (floor + `BodyComplete=false`) and pins it with a regression test.

## Evidence

- New regression tests:
  - `CompileBackTargets_FullPreservesConcreteSiblingWhenTargetIsPropertyAccessor` — target `get_Value`; asserts sibling `_v = value;` emitted, no `throw null`, sibling `Complete`, comparison `Exact`, `BodyComplete=true`.
  - `CompileBackTargets_FullEventAccessorTargetSurfacesRecompileFailure` — target `add_Changed`; asserts `UsedCompileBackFloor=true`, `BodyComplete=false`.
- Full RoundTrip area: **299/299 pass** (297 baseline + 2 new), no regressions from the `BodyComplete` integrity gate.

## Adversarial review

Two-model fixed-head review (Gemini 3.1 Pro + GPT-5.6, excluding the authoring model) in progress; reconciliation to follow.

### DIFF
--- a/src/ILInspector.CSharp/CSharpTypePrinter.cs
+++ b/src/ILInspector.CSharp/CSharpTypePrinter.cs
@@ -371,7 +371,7 @@ public sealed class CSharpTypePrinter
             if (body.Getter is not null)
                 accessors.Add(AccessorHead(member.Member, "get") + ";");
             if (body.Setter is not null)
-                accessors.Add(AccessorHead(member.Member, "set") + ";");
+                accessors.Add(AccessorHead(member.Member, SetterKeyword(member.Member)) + ";");
             return [$"{PadDeclaration(declaration, pad)} {{ {string.Join(" ", accessors)} }}"];
         }
 
@@ -381,10 +381,19 @@ public sealed class CSharpTypePrinter
             $"{pad}{{"
         };
         AddAccessor(lines, member.Member, "get", body.Getter, indent + 1);
-        AddAccessor(lines, member.Member, "set", body.Setter, indent + 1);
+        AddAccessor(lines, member.Member, SetterKeyword(member.Member), body.Setter, indent + 1);
         lines.Add($"{pad}}}");
         return lines;
     }
+
+    // An init-only property's write accessor is spelled `init`, not `set`. Honor the
+    // accessor model so full-body rendering does not silently downgrade `init` to a
+    // public `set` (dropping the required modreq(IsExternalInit)).
+    static string SetterKeyword(ApiMember member)
+        => member.SignatureModel?.Accessors is { } accessors
+            && accessors.Any(accessor => accessor.Kind == "init")
+            ? "init"
+            : "set";
 
     static IEnumerable<string> RenderEvent(
         PreparedType type,
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -332,6 +332,42 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
+    public void CompileBackTargets_FullPreservesAutoPropertyWhenTargetIsAutoAccessor()
+    {
+        // Issue #3000 regression guard: when the target is an auto-property accessor, the property
+        // has no explicit accessor body to preserve. The target-aware branch must leave the base
+        // auto-property skeleton intact rather than replacing it with empty accessor bodies (which
+        // deletes the accessors -> `int Value {  }`, CS0548, forcing a floor fallback).
+        var assemblyPath = CompileFixture("""
+            public class Holder
+            {
+                public int Value { get; } = 42;
+            }
+            """);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
+                [new ReturnToSender.RequestedTarget("Holder", "get_Value", 0)],
+                RoundTripScope.All,
+                RoundTripBodyPolicy.Full));
+
+            var getter = Assert.Single(result.FullBodies, body => body.Member == "Holder.get_Value");
+            Assert.Equal(MemberBodyProductionStatus.Complete, getter.Status);
+            Assert.Contains("Value { get; }", result.Source, StringComparison.Ordinal);
+            Assert.DoesNotContain("Value {  }", result.Source, StringComparison.Ordinal);
+            Assert.False(result.UsedCompileBackFloor, result.Detail);
+            Assert.True(
+                result.BodyComplete,
+                string.Join(Environment.NewLine, result.FullBodies.Select(body => $"{body.Member}: {body.Status}: {body.Failure}")));
+        }
+        finally
+        {
+            DeleteFixture(assemblyPath);
+        }
+    }
+
+    [Fact]
     public void CompileBackTargets_FullEventAccessorTargetSurfacesRecompileFailure()
     {
         // Issue #3000: a plain (non-explicit-interface) event accessor target is method-routed, so
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -554,6 +554,122 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
+    public void CompileBackTargets_FullMemberSurfacePreservesSiblingInitAccessor()
+    {
+        // Issue #3000: targeting a plain method under Full adds the whole declaring type to the
+        // member surface, so its sibling explicit-body init property flows through the surface
+        // stub-selection path. That path hardcoded `set` (ThrowGetSet / AutoPropertyGetSet),
+        // silently downgrading the init-only setter to a public `set` while Enrich filled the real
+        // body and reported it Complete. The surface path must route init setters through the
+        // init-aware stub kind so the accessor is spelled `init`.
+        var assemblyPath = CompileFixture("""
+            public class Holder
+            {
+                private int _value;
+                public int Value
+                {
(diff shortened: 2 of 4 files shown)


richlander:
## Adversarial review — round 1 findings + fixes

Two-model fixed-head review at `1a8344fb` (Gemini 3.1 Pro, GPT-5.6; authoring model Claude Opus excluded).

**Finding 1 — auto-property target clobbered (GPT High-confidence; Gemini also flagged, High).**
When the target was an auto-property accessor, the target-aware branch overwrote the base auto-property skeleton with `CSharpPropertyBody(null, null)`, emitting `int Value {  }` (CS0548) and forcing a floor fallback for a case that previously reconstructed cleanly.
Fixed in `fe1a467e`: guard `if (targetInvolved && basePropertyBody is null) continue;` preserves the base skeleton. Regression test `CompileBackTargets_FullPreservesAutoPropertyWhenTargetIsAutoAccessor` added (auto get-only target → no floor, `BodyComplete` true).

**Finding 2 — non-target sibling clobbered to null (Gemini High-confidence).**
For a non-concrete sibling accessor (RVA 0) paired with a concrete target accessor, forced readiness let the base sibling stub be replaced with `null`, dropping the accessor.
Fixed in `e2b91bbf`: the non-target sibling branch falls back to the base body (`basePropertyBody?.Getter/Setter`) instead of `null`. This shape is not expressible in C# (abstract is a property-level modifier), so it's defense-in-depth for arbitrary metadata; the auto/skeleton case exercises the `basePropertyBody is null` path.

Full RoundTrip area: **300/300 pass** at `e2b91bbf`.

Re-review of the fixed head `e2b91bbf` is in progress with both models. Will confirm two clean reviews before marking ready.

richlander:
## Adversarial review reconciliation (two models, fixed head `bec52e15`)

Reviewed by **Gemini 3.1 Pro** and **GPT-5.6** in isolated review worktrees. Both converged and both signed off clean on the final head `bec52e15`. CI (`build-net10`, `test (ubuntu-26.04, linux-x64)`) is green on `bec52e15`.

The review peeled the accessor-fidelity onion one layer per round; every finding was reproduced on a clean exact head before fixing and re-verified after.

### Findings and resolutions

1. **Explicit-body (non-auto) init property downgraded `init`→`set`** — found independently by **both** Gemini and GPT. A `get => _v; init => _v = value;` property targeted at either accessor rendered a public `set`, stripping `modreq(IsExternalInit)` while reporting the sibling Complete. The harness IL-equivalence ignores the signature modreq, so it falsely reported Exact/BodyComplete.
   Fixed in `365d0602`: init-aware stub kinds + ungated init detection for non-auto properties; product printer derives the write-accessor keyword from the accessor model (`SetterKeyword`) instead of hardcoding `set`.

2. **Member-surface init downgrade** — GPT **disproved** my "surface path is unreachable for init" claim (targeting a plain method under `Full` adds the whole declaring type to the surface), and Gemini independently reached the same class of bug via **record `ToString`** and **interface dependency** surfaces. Sibling init properties flowed through the surface stub paths, which hardcoded `set`.
   Fixed in `93ce10fb`: new `ThrowGetInit` stub kind; both surface stub-selection paths made init-aware via `SetterIsInitOnly`; extended `PropertyAccessors`/`ToMemberPolicy`. An existing test had *pinned the buggy* `{ get; set; }` for positional record properties — updated to `{ get; init; }` (still Exact), directly confirming the downgrade was real and previously accepted.

3. **Sibling auto-property enriched into recursive bodies** — GPT, on the fixed init head. Under `Full`, a non-target auto-property sibling was enriched by decompiling its synthesized accessors (which touch the unspeakable backing field); the decompiler renders that as the property itself, yielding recursive `get { return this.Value; }` while reporting BodyComplete=true. Same silent-Complete family; affects `{ get; set; }` auto-properties too (pre-existing).
   Fixed in `bec52e15`: the skeleton-preservation guard in Enrich now covers any auto-property whose composed policy is `Skeleton` (target **or** sibling). Gemini re-verified that explicit-body siblings still enrich to their real bodies (non-null `basePropertyBody`), so the guard does not over-suppress.

### Edge-case probes (both reviewers, clean head)

Indexer init and static init are compiler-forbidden (CS8865/CS8864/CS8856); explicit-interface init, generic/array-typed init, record-struct init, and init-only-no-getter all render `init` correctly and are covered by routing or tests.

### Out of scope (agreed follow-up)

Under `Full` the compiler-generated backing field is still emitted as a stray `public int __Value_k__BackingField;` alongside the preserved auto-property skeleton. Pre-existing, compiles, does not affect target IL — not a silent-Complete violation. Both reviewers agreed it is non-blocking → filed as #3036.

### Validation

RTS class 148/148; RoundTrip trait area 310/310; `RuntimeAsyncAwaiterPassTests` (also exercises `CompileBackTargets`) 22/22. The two known local fidelity reds (lambda-cache ordinal; iterator operand diff) reproduce identically on the base without this change and are green in CI — environmental, not caused by this PR.

richlander:
Ready to merge

PR #3012, opened 2026-07-22
Render wide array index as bare index, not checked((nint)i) (#2981)

### LINKED ISSUE #2981: Fidelity: long array index rendered as array[checked((nint)i)] instead of array[i] (Humanizer NumberToWords, ~47 methods)
## Summary

**Fidelity (readability) gap — not a correctness bug.** When an array is indexed by a `long` (or other non-`int`) value, the decompiler renders the index with an explicit `checked((nint)…)` cast instead of the idiomatic bare index. The C# compiler already emits the range conversion for a wide array index, so the original source is a plain `array[i]`. The explicit cast is verbose, unidiomatic, and round-trips to different IL (extra `conv.i8`/`conv.ovf.i`), so these rows land in `ValidDifferent` (`semantic_opcode_diff.checked_context`) rather than exact.

This is a single systematic pattern that shows up across the whole Humanizer `*NumberToWordsConverter` family (~47 methods in the EVIL corpus) and any other `long`-indexed array access.

## Reproduction

Package: **Humanizer.Core@3.0.10** (assembly `Humanizer`, v3.0.0.0).

```bash
dnx dotnet-inspect -y -- member "Humanizer.EnglishNumberToWordsConverter" GetUnitValue \
  --package Humanizer.Core@3.0.10 --all -S "Decompiled Source"
```

### Observed (decompiled)

```csharp
private static string GetUnitValue(long number, bool isOrdinal)
{
    if (!isOrdinal)
    {
        return EnglishNumberToWordsConverter.UnitsMap[checked((nint)number)];
    }
    return EnglishNumberToWordsConverter.UnitsMapTh[checked((nint)number)];
}
```

### Original source

<https://github.com/Humanizr/Humanizer/blob/90f0cb1d7427635f38b46350bb1dc5fca2164f2d/src/Humanizer/Localisation/NumberToWords/EnglishNumberToWordsConverter.cs#L102>

```csharp
static string GetUnitValue(long number, bool isOrdinal) =>
    isOrdinal ? UnitsMapTh[number] : UnitsMap[number];
```

The index is a bare `long`; `array[checked((nint)number)]` should be `array[number]`.

### Second example (same pattern)

```bash
dnx dotnet-inspect -y -- member "Humanizer.GermanNumberToWordsConverterBase" GetTens \
  --package Humanizer.Core@3.0.10 --all -S "Decompiled Source"
```

```csharp
// observed
protected virtual string GetTens(long tens) => TensMap[checked((nint)tens)];
// original: TensMap[tens];
```

## IL evidence (ReturnToSender opcode diff, EnglishNumberToWordsConverter.GetUnitValue)

```
Original:   ldarg brtrue ldsfld ldarg conv.ovf.i           ldelem.ref ret  (x2)
Recompiled: ldarg brtrue ldsfld ldarg conv.ovf.i conv.i8 conv.ovf.i ldelem.ref ret  (x2)
```

The original array-index conversion is a single `conv.ovf.i`; the explicit `checked((nint)…)` rendering round-trips to a redundant `conv.i8`/`conv.ovf.i` widen-then-renarrow. Semantically equivalent (checked range behavior preserved), so this is fidelity/readability only.

## Affected (EVIL corpus, Humanizer.Core@3.0.10)

~47 `*NumberToWordsConverter` methods classified `valid_different.semantic_opcode_diff.checked_context`, e.g. `English/German/French/Russian/Tamil/Icelandic/Bulgarian…NumberToWordsConverter` `GetUnitValue`/`GetTens`/`CollectPartsUnderOneThousand`/`Convert`. All share the `array[checked((nint)i)]` shape for a `long` index.

## Expected

Render a wide array index as the bare index (`array[i]`) when the explicit conversion is exactly the one the C# compiler would insert for that index type, matching the original source and producing exact IL round-trip.

/label bug

richlander:
- Fixes #2981
- Renders wide (`long`/`ulong`) array indices idiomatically instead of `array[checked((nint)i)]`, spelling the exact signedness/checked shape C# re-inserts so every affected member recompiles to identical IL.
- Card revision: `8c657d01`

## Change

**Conclusion:** **PASS** — the printer now spells a wide array index as the shape
whose re-inserted native-int index conversion is opcode-identical to the
original, reading idiomatically while staying round-trip exact across enum,
signedness, and `checked`-context edge cases surfaced in adversarial review.

C# converts a `long`/`ulong` array index to native int with a checked range
conversion (`conv.ovf.i` / `conv.ovf.i.un`) that it inserts implicitly and
always emits, independent of the enclosing `checked`/`unchecked` context. The
printer was spelling that conversion explicitly (`array[checked((nint)i)]`). The
fix strips or restates it at the single-dimension element-access sites
(`LoadElement`, `StoreElement`, `LoadElementAddress`) through one shared
`ArrayIndexText` helper, matching a checked conversion to `IntPtr` whose operand
is `Int64`/`UInt64`; any other source is left spelled.

The helper resolves the operand's **real** C# type — a typed load
(`ldelem.i8` / `ldind.i8`) reports only `Int64` storage, masking a `ulong`
element or a wide enum, and a wide binary or unary neg/not keeps a signed stack
`ResultType` even when it renders `ulong` — via `WideIndexOperandType` (array
element / ref / pointer pointee; recursing through the unmasked operands of a
sign-neutral wide binary — `+`/`-`/`*` and bitwise — regardless of checkedness;
taking a shift-left's left-operand type; reading a divide/remainder/shift-right's
signedness from its opcode variant `div.un`/`rem.un`/`shr.un`; recursing through a
unary bitwise `~` into its operand, and reporting a unary `-` as the signed type
it renders as once its operand's reinterpret cast is re-inserted — including an
enum negate, which reports `Int64`/`Int32` by the enum's underlying width, or —
when the enum shape does not resolve (cross-assembly) — by the negate's masked
stack `ResultType`; else the operand's own type). From that it
renders:

| Operand shape | Rendered index | Re-inserted conv |
| --- | --- | --- |
| `long[]` / `ulong[]` element, matching sign | bare `v[j]` | `conv.ovf.i` / `conv.ovf.i.un` |
| `long`/`ulong` param, matching sign | bare `i` | `conv.ovf.i` / `conv.ovf.i.un` |
| wide-enum element/ref (masked, bare is CS0266) | `(long)`/`(ulong)` cast | matching conv |
| `ulong` element as a **signed** index | `(long)` cast | `conv.ovf.i` |
| `long` element as an **unsigned** index | `(ulong)` cast | `conv.ovf.i.un` |
| `ulong` **sum** as a signed index | `(long)(… + …)` cast | `conv.ovf.i` |
| `ulong` `checked` sum / `>>`/`/`/`%` / `~` as a signed index | `(long)(…)` cast | `conv.ovf.i` |
| sign-changing cast inside `checked(...)` | `unchecked((long)/(ulong) …)` | matching conv, no spurious overflow |
| stripped bare negate inside `checked(...)` | `checked(a[unchecked(-i)] …)` | matching conv, unchecked `neg` |
| negate over a masked `ulong`/`nuint` element | `-(long)v[j]` / `-(nint)v[j]` | matching conv, `neg` |
| negate over a masked/wide enum element | `-(long)v[j]` (8-byte) / `-(int)v[j]` | matching conv, `neg` |
| negate over a cross-assembly enum element (width from opcode) | `-(long)v[j]` / `-(int)v[j]` | matching conv, `neg` |

The last row is the `checked`-context guard: the `(long)`/`(ulong)` reinterpret
is a no-op in IL (both sides 8-byte; only the always-checked index conv is
checked), but spelled bare inside a lexical `checked` region a **sign-changing**
cast recompiles to a `conv.ovf.i8.un` / `conv.ovf.u8` the original never had.
It is wrapped in `unchecked(...)` only when it flips sign
(`WideIndexCastSignChanges` reads the enum underlying / primitive signedness;
unknown backing wraps to be safe); a same-sign cast (a long-backed enum's
`(long)`) emits no conv even when checked and stays bare.

The same `checked`-context hazard applies once a bare wide index operand is
stripped and routed back through the general expression printer: a bare unary
negate `-i` spelled inside a lexical `checked` region recompiles as an
overflow-checked negate (C# lowers `checked(-i)` to `0 - i` as `sub.ovf`) where
the original IL has an unchecked `neg`. The general printer's new `UnaryText`
wraps an integer negate in `unchecked(...)` when in a `checked` context,
mirroring the existing add/sub/mul handling in `BinaryText`; a bitwise
complement (`~`) never overflows and stays bare, and a genuine checked negate is
raised as a Binary subtract-from-zero and is unaffected.

A second `neg` hazard is signedness, not overflow: C# has no unary minus for
`ulong`/`nuint` (they promote to no signed type, so `-x` is CS0023). An IL `neg`
over such a value came from a signed reinterpret cast (`-(long)x`) that is a
no-op in IL — invisible in the operand's stack type, and doubly so for a
`ulong[]` element whose `ldelem.i8` masks it as `Int64`. `UnaryText` re-inserts
that cast on the negate's **operand** (`-(long)v[j]`, `-(nint)v[j]`), not around
the whole negate — `(long)(-v[j])` would still negate a `ulong`. `uint` is
excluded because its unary minus already promotes to `long` value-preservingly.
The same hazard applies to **every enum** — C# has no unary minus on any enum
regardless of its underlying type — so an enum negate operand is reinterpreted by
its underlying width (`(long)` for an 8-byte-backed enum, `(int)` for a narrower
one, since sub-8-byte integers are `I4` on the eval stack). When the enum's shape
does not resolve — a cross-assembly or core-library enum whose type maps are not
walked — the underlying width is unavailable, but the negate's masked stack width
(`ldelem.i8`→`long`, `ldelem.i4`→`int`) is, and an IL `neg` over a non-primitive
rendered type can only be an enum (a struct's unary minus is an
`op_UnaryNegation` call), so the reinterpret is recovered from that width. A
scalar cross-assembly enum negate carries no width in its opcode and stays
uncast (its width is genuinely unresolvable).

### Original source

`Humanizer.GermanNumberToWordsConverterBase.GetTens` (Humanizer.Core@3.0.10):

```csharp
protected virtual string GetTens(long tens) => TensMap[tens];
```

### Before

```csharp
protected virtual string GetTens(long tens) => TensMap[checked((nint)tens)];
```

- Valid: True
- Correct: True

The explicit `checked((nint)…)` is valid and behavior-preserving, but verbose,
unidiomatic, and round-trips to a redundant `conv.i8`/`conv.ovf.i`
widen-then-renarrow (`valid_different.semantic_opcode_diff.checked_context`).

### After

```csharp
protected virtual string GetTens(long tens) => TensMap[tens];
```

- Valid: True
- Correct: True

### Fully raised

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | n/a (new) | `WideArrayIndexTests`: 35 passed |
| Related printer suites | Pass | EnumCast 25, MultiDim 23, EnumUnderlying 9, IndexFromEnd 15, CheckedUncheckedInsert 17, InlineArrayElementRef 12, EnumConstant 21, + others all green |
| Decompiler compile-back gates | RecompileFail (env) | RecompileFail (env) |
| Real witness (IL round-trip) | `a[checked((nint)i)]` | idiomatic index — all shapes opcode-exact |

**Compile-back gate note.** `FidelityGateTests`/`LoweredFidelityGateTests`
regress **wholesale** to `RecompileFail` on this Windows host — all 87 recompile
failures share one cause: `CS0009: … Microsoft.AspNetCore.App\…\aspnetcorev2_inprocess.dll
… PE image doesn't contain managed metadata` (the harness runtime-reference
resolver enumerates the ASP.NET Core shared framework and hands Roslyn a native
DLL). This is a pre-existing, machine-specific environment failure, identical on
Baseline; this change never touches reference resolution or any failing fixture.
CI validates these gates.

**Machine-robust round-trip proof (self-contained, no harness references).**
Each decompiled body was compiled and its IL opcode stream diffed against the
original fixture — **exact across every shape**, covering `long`/`ulong` load,
store, `ref`, a `long` add expression, `checked`/`unchecked` contexts
(`add`/`add.ovf` preserved), `nint`/`nuint`, wide enums, signedness, and the
`checked`-context wrap:

| Case | Original spelling | Decompiled | IL |
| --- | --- | --- | --- |
| `a[i]` (long) | `a[checked((nint)i)]` | `a[i]` | exact |
| `a[i]` (ulong) | `a[checked((nint)i)]` | `a[i]` | exact |
| `a[i] = v` | `a[checked((nint)i)] = v` | `a[i] = v` | exact |
| `ref a[i]` | `ref a[checked((nint)i)]` | `ref a[i]` | exact |
| `a[i + j]` | `a[checked((nint)(i + j))]` | `a[i + j]` | exact |
| `a[nintVar]` | `a[checked((nint)(long)i)]` | `a[(long)i]` | exact |
| wide-enum element | (bare CS0266) | `a[(long)values[j]]` | exact |
| wide-enum `ref` | (bare CS0266) | `a[(long)(r)]` | exact |
| `ulong` element as signed | `a[(long)v[j]]` | `a[(long)v[j]]` | exact |
| `long` element as unsigned | `a[(ulong)v[j]]` | `a[(ulong)v[j]]` | exact |
| plain `ulong` element | `a[v[j]]` | `a[v[j]]` | exact |
| `ulong` sum as signed | `a[(long)((ulong)v[j] + x)]` | same | exact |
| `ulong` element sum as signed | `a[(long)(v[j] + w[k])]` | same | exact |
| bare `ulong` sum | `a[v[j] + w[k]]` | same | exact |
| `ulong`-as-signed in `checked` | `checked(a[unchecked((long)v[j])] + 1)` | same | exact |
| `long`-as-unsigned in `checked` | `checked(a[unchecked((ulong)v[j])] + 1)` | same | exact |
| long-enum `(long)` in `checked` | `checked(a[(long)values[j]] + 1)` | same (no wrap) | exact |
| `ulong` `checked` sum as signed | `a[unchecked((long)checked(v[j] + x))]` | `a[(long)checked(v[j] + x)]` | exact |
| `ulong >> 1` (shr.un) as signed | `a[unchecked((long)(v[j] >> 1))]` | `a[(long)((ulong)v[j] >> 1)]` | exact |
| `ulong / d` (div.un) as signed | `a[unchecked((long)(v[j] / d))]` | `a[(long)((ulong)v[j] / d)]` | exact |
| `ulong % d` (rem.un) as signed | `a[unchecked((long)(v[j] % d))]` | `a[(long)((ulong)v[j] % d)]` | exact |
| signed `long << 1` | `a[i << 1]` | same (bare) | exact |
| signed `checked` `long` sum | `a[checked(i + j)]` | same (bare) | exact |
| `~ulong` element as signed | `a[unchecked((long)(~v[j]))]` | `a[(long)(~v[j])]` | exact |
| signed `long` negate | `a[-i]` | same (bare) | exact |
| `long` negate inside `checked` | `checked(a[unchecked(-i)] + 1)` | same | exact |
| general negate inside `checked` | `checked(x + unchecked(-i))` | same | exact |
| signed `long` bitwise-not | `a[~i]` | same (bare) | exact |
| bitwise-not inside `checked` | `checked(a[~i] + 1)` | same (bare, no wrap) | exact |
| negate of masked `ulong` element (signed index) | `a[-(long)v[j]]` | same (not `(long)(-v[j])`, CS0023) | exact |
| negate of masked `ulong` element (general, non-index) | `-(long)v[j]` | same | exact |
| negate of masked `nuint` element | `-(nint)v[j]` | same | exact |
| negate of masked `ulong`-backed enum (signed index) | `a[-(long)v[j]]` | same (not `(long)(-v[j])`, CS0023) | exact |
| negate of masked `long`-backed enum (general) | `-(long)v[j]` | same | exact |
| negate of `uint`-backed enum (general) | `-(int)v[j]` | same (outer `(int)` is the enum→underlying conv) | exact |
| negate of cross-assembly `ulong`-backed enum (signed index) | `a[-(long)v[j]]` | same (width from `ldelem.i8`) | exact |
| negate of cross-assembly `long`-backed enum (general) | `-(long)v[j]` | same | exact |
| negate of cross-assembly `uint`-backed enum (general) | `(int)(-(int)v[j])` | same (width from `ldelem.u4`) | exact |
| negate of core-library enum (`DayOfWeek`) index | `a[-(int)v[j]]` | same (width from `ldelem.i4`) | exact |
| `~` of masked `ulong` element (unsigned index) | `a[~v[j]]` | same (bare) | exact |

## Decompiler quality

**Conclusion:** **PASS** — printer-only change; the IR is untouched, so
structural corpus metrics (Full%, lowering/branch residue, stop reasons, pass
bugs) are unchanged by construction. The only corpus movement is intended:
wide-indexed members leave `valid_different.semantic_opcode_diff.checked_context`
for exact round-trip.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.WideArrayIndexTests"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -951,6 +951,14 @@ public class CfgSampleClass
     // value-preserving no-op view the `neg` operated on. Opcode-exact: `ldelem.*; neg`.
     public static int NegEnumUIntElementToInt(CfgFlags[] v, int j) => -(int)v[j];
 
+    // A cross-assembly (CoreLib) enum used as an array index: DayOfWeek resolves
+    // to an Unknown shape (EnsureTypeMaps walks only this assembly's type defs),
+    // so the enum underlying is unavailable. C# still has no unary minus on the
+    // enum, but the masked stack width IS in the `ldelem.i4` opcode, so the
+    // reinterpret is recovered from it: `a[-(int)v[j]]`. Opcode-exact: `ldelem.i4;
+    // neg; conv.i`.
+    public static int NegCrossAssemblyEnumElementIndex(int[] a, System.DayOfWeek[] v, int j) => a[-(int)v[j]];
+
     // Genuinely-signed `long` neg/not indices recover as `long` and strip bare.
     public static int NegLongIndexBare(int[] a, long i) => a[-i];
 
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -921,6 +921,13 @@ public class CfgSampleClass
 
     public static int NotLongIndexBare(int[] a, long i) => a[~i];
 
+    // Negate stripped bare inside a `checked` array-index context (#2981
+    // adversarial review): the index conv is dropped (signed `long` matches),
+    // but the bare `-i` would recompile as a checked negate (`0 - i` as
+    // `sub.ovf`) where the IL has an unchecked `neg`, so the negate is wrapped
+    // in `unchecked(...)`.
+    public static int NegLongIndexInChecked(int[] a, long i) => checked(a[unchecked(-i)] + 1);
+
     public static void SetFirstElement(int[] a, int v) => a[0] = v;
 
     // stelem.i1 stores into byte[], sbyte[], and bool[] alike, and stelem.i2 into
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -881,6 +881,34 @@ public class CfgSampleClass
     // sum just like a scalar `ulong`, with no redundant `(ulong)` cast.
     public static int ULongSumIndexBare(int[] a, ulong[] v, ulong[] w, int j, int k) => a[v[j] + w[k]];
 
+    // Checked-context inner binary (#2981 adversarial review): `checked(v[j] + x)`
+    // over a `ulong` element is a `ulong` add, but the compiler-inserted index
+    // conv is still signed (`conv.ovf.i`) because the *outer* cast is `(long)`.
+    // The operand-type recovery must see through the binary regardless of its
+    // checkedness — `checked` never changes an expression's C# type — and keep
+    // the `(long)` cast; stripping bare would re-insert `conv.ovf.i.un`.
+    public static int CheckedULongSumIndexAsSigned(int[] a, ulong[] v, int j, ulong x) => a[unchecked((long)checked(v[j] + x))];
+
+    // Unsigned shift-right (`shr.un`): `ulong >> count` is a `ulong`. The result
+    // type is the shifted (left) operand's, recovered through the mask, so a
+    // signed index keeps the `(long)` cast (bare would flip to `conv.ovf.i.un`).
+    public static int ULongShrIndexAsSigned(int[] a, ulong[] v, int j) => a[unchecked((long)(v[j] >> 1))];
+
+    // Unsigned divide (`div.un`) / remainder (`rem.un`): the signedness lives in
+    // the opcode variant, not the sign-erased Int64 stack type. A `ulong`
+    // quotient/remainder used as a signed index keeps the `(long)` cast.
+    public static int ULongDivIndexAsSigned(int[] a, ulong[] v, int j, ulong d) => a[unchecked((long)(v[j] / d))];
+
+    public static int ULongRemIndexAsSigned(int[] a, ulong[] v, int j, ulong d) => a[unchecked((long)(v[j] % d))];
+
+    // Negatives for the finding #5 recovery: a genuinely-signed `long` shift-left
+    // and a `checked` signed `long` add must still strip to the bare index — the
+    // recovered `long` type matches the signed `conv.ovf.i`, so no spurious
+    // `(long)` cast is emitted.
+    public static int LongShlIndexBare(int[] a, long i) => a[i << 1];
+
+    public static int LongCheckedSumIndexBare(int[] a, long i, long j) => a[checked(i + j)];
+
     public static void SetFirstElement(int[] a, int v) => a[0] = v;
 
     // stelem.i1 stores into byte[], sbyte[], and bool[] alike, and stelem.i2 into
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -865,6 +865,22 @@ public class CfgSampleClass
     // `(long)`), so it stays bare with no spurious `unchecked(...)` wrapper.
     public static int LongEnumIndexChecked(int[] a, CfgLongPriority[] values, int j) => checked(a[(long)values[j]] + 1);
 
+    // Compound wide index (#2981 adversarial review): a sign-neutral wide binary
+    // reports Int64 stack storage even when it renders `ulong`, so the operand
+    // type must be recovered through the binary from its unmasked operands. A
+    // `ulong` sum used as a *signed* index keeps an explicit `(long)` cast around
+    // the whole expression — stripping it bare would re-insert `conv.ovf.i.un`.
+    public static int ULongSumIndexAsSigned(int[] a, ulong[] v, int j, ulong x) => a[unchecked((long)(v[j] + x))];
+
+    // Both operands are masked `ulong` elements (`ldelem.i8`): the recovery must
+    // see through the masking on each side, not just when one operand is a bare
+    // `ulong`. Still a signed index, so it keeps the `(long)` cast.
+    public static int ULongElementSumIndexAsSigned(int[] a, ulong[] v, ulong[] w, int j, int k) => a[unchecked((long)(v[j] + w[k]))];
+
+    // A bare `ulong` compound index (unsigned `conv.ovf.i.un`) strips to the bare
+    // sum just like a scalar `ulong`, with no redundant `(ulong)` cast.
+    public static int ULongSumIndexBare(int[] a, ulong[] v, ulong[] w, int j, int k) => a[v[j] + w[k]];
+
     public static void SetFirstElement(int[] a, int v) => a[0] = v;
 
     // stelem.i1 stores into byte[], sbyte[], and bool[] alike, and stelem.i2 into
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -852,6 +852,19 @@ public class CfgSampleClass
     // element type matches the unsigned `conv.ovf.i.un`, so it is opcode-exact.
     public static int ULongArrayElementIndex(int[] a, ulong[] v, int j) => a[v[j]];
 
+    // Wide index cast inside a lexical `checked` region (#2981 adversarial
+    // review): the emitted `(long)`/`(ulong)` reinterpret is a no-op in IL (only
+    // the always-checked native-int index conv is checked). A SIGN-CHANGING cast
+    // must be wrapped in `unchecked(...)` — spelled bare inside `checked` it would
+    // recompile to a `conv.ovf.i8.un` / `conv.ovf.u8` the original never had.
+    public static int ULongIndexAsSignedChecked(int[] a, ulong[] v, int j) => checked(a[unchecked((long)v[j])] + 1);
+
+    public static int LongIndexAsUnsignedChecked(int[] a, long[] v, int j) => checked(a[unchecked((ulong)v[j])] + 1);
+
+    // A same-sign cast emits no conv even inside `checked` (a long-backed enum's
+    // `(long)`), so it stays bare with no spurious `unchecked(...)` wrapper.
+    public static int LongEnumIndexChecked(int[] a, CfgLongPriority[] values, int j) => checked(a[(long)values[j]] + 1);
+
     public static void SetFirstElement(int[] a, int v) => a[0] = v;
 
     // stelem.i1 stores into byte[], sbyte[], and bool[] alike, and stelem.i2 into
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -834,6 +834,24 @@ public class CfgSampleClass
     // `a[(long)(r)]` rather than a CS0266 bare index.
     public static int LongEnumRefArrayIndex(int[] a, ref CfgLongPriority r) => a[(long)r];
 
+    // Signedness masking (#2981 adversarial review): `ldelem.i8` / `ldind.i8`
+    // report Int64 storage even for a `ulong` element/pointee, so the index
+    // conversion's own signedness (`conv.ovf.i` signed / `conv.ovf.i.un`
+    // unsigned) is the only witness of the source type. A `ulong` element used as
+    // a *signed* index must keep an explicit `(long)` cast — stripping it bare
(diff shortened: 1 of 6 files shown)


richlander:

Reviewed by two different model families per the roster: **GPT (`gpt-5.6-sol`)** and **Gemini (`gemini-3.1-pro-preview`)**, each isolated in its own linked review worktree with scratch kept outside the tree and no `git reset`/`add`/`commit`.

### Findings and resolutions

The core change (strip the compiler-inserted `checked((nint)i)` on a wide array index, opcode-exact) went through iterative adversarial review. Each round surfaced a related correctness edge case; all were fixed in follow-up commits (never amended):

| # | Finding | Resolution |
| --- | --- | --- |
| 1 | Wide-enum array index rendered bare (CS0266) | `2739c921` |
| 2 | `ulong` index rendered as signed `long` (wrong strip) | `c764d6c6` |
| 3 | Sign-changing cast in a `checked` context gained a spurious overflow conv | `7decac8d` |
| 4 | Compound-unsigned wide index mis-signed | `390c8fa0` |
| 5 | `checked`/`>>`/`/`/`%` wide binary index signedness | `76b9b02e` |
| 6 | Unary neg/not lost the operand sign-recovery cast | `6e44e58d` |
| 7 | Unary negate lost `unchecked` inside a `checked` region | `9158c9cf` |
| 8 | Primitive `ulong`/`nuint` negate emitted invalid `(long)(-x)` (CS0023) | `cd8d72d8` |
| 9 | **Every** enum negate is CS0023; cast must key on underlying width | `6e7320e3` |
| 10 | Cross-assembly / core-library enum negate reverted to invalid (unresolved shape) | `8c657d01` |

Findings #1 and #4–#10 were surfaced by the GPT reviewer; Gemini provided independent clean confirmations across rounds (including a fully empirical opcode-exact pass this round). Each fix is printer-only — the IR is untouched, so structural corpus metrics are unchanged by construction.

### Finding #10 (final round) — resolution detail

`EnumUnderlyingType` is same-assembly-only (`EnsureTypeMaps` walks only the target assembly's own type defs), so a referenced-assembly or core-library enum resolves to an Unknown shape and the finding-#9 cast never fired, leaking `-enumValue` (CS0023). The fix recovers the reinterpret from the negate's **masked stack width** (`ldelem.i8`→`long`, `ldelem.i4`→`int`), gated on `IsUnresolvedEnumLike` (a `Definition` that `TypeFamilies.Of` cannot classify as a primitive — which correctly includes core-lib enums like `DayOfWeek` and excludes all primitives). An IL `neg` over a non-primitive rendered type can only be an enum (a struct's unary minus is an `op_UnaryNegation` call). A scalar cross-assembly enum negate carries no width in its opcode and stays uncast — the pre-existing, genuinely-unresolvable limitation, intentionally unchanged.

### Fixed-head re-review on `8c657d01` — both CLEAN

- **GPT (`gpt-5.6-sol`): CLEAN** — confirmed its finding #10 is resolved at `8c657d01`; no significant issues in the reviewed changes.
- **Gemini (`gemini-3.1-pro-preview`): CLEAN** — empirically validated all nine attack points via cross-assembly (`App`→`Lib` project reference) and core-library probes with recompile opcode diffs: `ULongEnum`→`a[-(long)v[j]]` (`ldelem.i8; neg; conv.ovf.i`), `DayOfWeek`→`a[-(int)v[j]]` (`ldelem.i4; neg`), all opcode-exact; `~` stays bare; primitive and same-assembly-enum negates unperturbed; `checked`-context composition exact; no `IsUnresolvedEnumLike` false positives.

### Validation

- CI green on `8c657d01`: `test (ubuntu-26.04, linux-x64)`, `build-net10`, `csharp-diff-smoke`, `changes` all pass (`il-diff-smoke`/`markdownlint`/`pack` path-gated, skipped).
- Focused suite: `WideArrayIndexTests` 35 passed; regressions green (CrossAssemblyEnumInteger 4, EnumConstant 21, EnumUnderlyingCast 9, RefEnumBitwise 1, CheckedUncheckedInsert 17, and others).
- Machine-robust opcode-exact recompile evidence for every wide-index shape is tabulated in the PR description. The compile-back fidelity gate classes fail wholesale on the dev host from a pre-existing CS0009 `aspnetcorev2_inprocess.dll` environment issue unrelated to this change; CI validates those gates.

Ready to merge

PR #3013, opened 2026-07-22
Recompose static-target chained assignment (a = b = c = v) from dup idiom (#2994)

### LINKED ISSUE #2994: Decompiler: reconstruct chained/embedded assignment from dup-of-value (a = b = c = v)

richlander:
- Fixes #2994
- Changes: static-target chained assignment (`a = b = c = v`) is recomposed from the IL dup-of-value idiom instead of being emitted as reordered, independent statements.
- Card revision: `28279a45`

## Change

**Conclusion:** **PASS** — static property/field chains now render as a single right-associative `A = B = C = v;`, matching source order and per-sink coercion, while genuinely-separate statements and non-chain constant dups are provably left untouched.

### Original source

Synthetic witness fixture (no SourceLink; source shown verbatim from the compiled fixture):

```csharp
public static class Props
{
    public static bool A { get; set; }
    public static bool B { get; set; }
    public static bool C { get; set; }
    public static int  I { get; set; }
    public static long L { get; set; }

    public static void ChainBool()       => A = B = C = false;
    public static void ChainMixedWiden() => L = I = -1;
    public static void Separate()        { A = false; B = false; }
}
```

### Before

`dotnet-inspect member lib.dll Props.ChainBool -S "Decompiled Source"` at base (`c04c2345`):

```csharp
public static void ChainBool()
{
    Props.C = false;
    Props.B = false;
    Props.A = false;
}
```

- Valid: True
- Correct: True

Behavior-preserving here, but the chain intent is lost and the sinks are reordered to sink/document order (C, B, A). `ChainMixedWiden` similarly split to `Props.I = -1; Props.L = (long)-1;`.

### After

`dotnet-inspect member lib.dll Props.ChainBool -S "Decompiled Source"` at head (`28279a45`):

```csharp
public static void ChainBool() => Props.A = Props.B = Props.C = false;
```

- Valid: True
- Correct: True

`ChainMixedWiden` → `public static void ChainMixedWiden() => Props.L = Props.I = -1;` (the `long` widen becomes implicit in the chain spelling). Static-field chains render identically (`StaticFields.A = StaticFields.B = StaticFields.C = true;`). Non-constant chains recompose too (`P = Q = R = Compute();`).

Negative control — `Props.Separate()` is byte-identical Before and After (no dup evidence, so no recomposition):

```csharp
public static void Separate()
{
    Props.A = false;
    Props.B = false;
}
```

### Fully raised

The After decompilation is in the fully raised state for static-target chains.

- Required tracking issue: #2994 — instance-property/field and local chains (compiler-temp `V_0` shapes) remain out of scope for this slice and fall through to the existing per-sink rendering.

## Evidence

| Check | Baseline (`c04c2345`) | Head (`28279a45`) |
| --- | --- | --- |
| Focused tests | `ChainedConstantAssignmentTests`: net-new | `ChainedConstantAssignmentTests`: 6 passed |
| Decompiler fast suite | 0 failed | 3142 passed, 0 failed |
| Reduced fixture validity | `ValidityCoverageReportingTests`: 3 passed (no CS0161 pin shift) | 3 passed |
| Lowering coverage | `LoweringCoverageTests`: 6 passed | 6 passed |
| CfgSampleClass fidelity | provably unaffected | provably unaffected |
| Real witness | `Props.ChainBool` reordered/split | `Props.ChainBool` chained |

**CfgSampleClass fidelity is provably unaffected.** The importer change only touches the constant-dup branch (`StoreStackSlot` with slot ≥ `DupSlotBase` whose value is a `Constant`). A direct probe imported all **613** `CfgSampleClass` methods and found **0** such shapes, so every fidelity-gate fixture method imports byte-identically to baseline and no `FidelityGateTests.KnownDiffs` edit is required.

### Slow gates (deferred to CI)

The slow corpus / compile-back fidelity gates could not complete locally due to sustained machine contention (load avg ~20 from unrelated worktrees; individual compile-back tests ran 30+ min without finishing). They are unaffected by this change on inspection (recomposed `A = B = C = v` is the canonical Roslyn inverse of the dup lowering, so it recompiles to bit-identical IL), but **CI must confirm the full slow corpus/fidelity suite**.

## Design

#2995 (validity fix for #2982) re-materialized dup'd constants as clones **in the importer**, which discarded the shared-dup-slot provenance — making a real constant chain byte-indistinguishable from genuinely separate `A = false; B = false;`. This PR:

1. Reverts the importer to uniform slot-materialization for all dups, restoring the shared-slot evidence a sound chain detector needs.
2. Adds `ChainedAssignmentPass`, which recomposes dup-slot chains (≥2 static-target sinks, with clobber/full-consumption guards and a widening type-staircase check) into a new `ChainedAssignment` IR node, and **re-materializes remaining non-chain constant dups as a fallback** — preserving #2995's validity fix (#2982).
3. Adds printer, `TypedConstantsPass`, and `CoercionInsertionPass` support for the new node.

Scope is deliberately limited to static property/field sinks (slice 1); instance/local chains fall through untouched.

## Validation

```bash
dotnet build src/dotnet-inspect/dotnet-inspect.csproj -c Release --configfile nuget.config
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.ChainedConstantAssignmentTests"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -5247,6 +5247,14 @@ public class BackingFieldSample
 // literal is dup'd to each sink. The importer must re-materialize the dup'd constant
 // at each bool sink so it renders `A = false;`, not spill it into an int stack slot
 // (`int S = 0; A = S;`), which is CS0029 (cannot implicitly convert int to bool).
+// Regression + quality fixtures for #2982 / #2994: a value assigned in a chain
+// (`a = b = c = v`) compiles to a dup-of-value idiom — the rvalue is evaluated
+// once and dup'd to each sink (`ldc.i4.0; dup; call set_C; dup; call set_B; call
+// set_A`). ChainedAssignmentPass recomposes that run into `A = B = C = v;`,
+// keyed on the shared dup slot. A dup'd constant that is NOT part of a chain is
+// re-materialized at its sink so a bool/char/enum literal is recovered (#2982)
+// instead of spilling through an int32 slot (CS0029). Genuinely separate
+// statements carry no dup slot and must not collapse.
 public static class ChainedConstantAssignmentSamples
 {
     public static bool A { get; set; }
@@ -5255,5 +5263,50 @@ public static class ChainedConstantAssignmentSamples
 
     public static bool C { get; set; }
 
+    public static int I { get; set; }
+
+    public static long L { get; set; }
+
+    public static int P { get; set; }
+
+    public static int Q { get; set; }
+
+    public static int R { get; set; }
+
+    public static bool F;
+
+    public static bool G;
+
+    public static bool H;
+
+    static int Compute() => 42;
+
+    // The Dapper Settings.SetDefaults shape: a bool constant chained across
+    // static properties. Recomposes to `A = B = C = false;`.
     public static void ChainedBoolFalse() => A = B = C = false;
+
+    // A widening chain: the shared int constant lands in `I` (int) and widens
+    // implicitly into `L` (long). Recomposes to `L = I = -1;`.
+    public static void ChainedWiden() => L = I = -1;
+
+    // A non-constant chain: the shared call result flows to each static
+    // property. Recomposes to `P = Q = R = Compute();`.
+    public static void ChainedNonConstant() => P = Q = R = Compute();
+
+    // A bool constant chained across static fields. Recomposes to
+    // `F = G = H = true;`.
+    public static void ChainedStaticFields() => F = G = H = true;
+
+    // Negative: two independent statements with no dup — must stay two
+    // statements, never collapse into a chain.
+    public static void SeparateStatements()
+    {
+        A = false;
+        B = false;
+    }
+
+    // Negative: a single-target dup'd constant whose value escapes into an
+    // argument (`WriteLine(P = 5)`). Not a chain (one sink); the constant is
+    // re-materialized to `P = 5; Console.WriteLine(5);`.
+    public static void SideEffectValue() => System.Console.WriteLine(P = 5);
 }
--- a/src/ILInspector.Decompiler.Tests/ChainedConstantAssignmentTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ChainedConstantAssignmentTests.cs
@@ -3,11 +3,13 @@ using ILInspector.Decompiler.Pipeline;
 
 namespace ILInspector.Decompiler.Tests;
 
-// #2982: a constant assigned through a chain (`a = b = c = false`) is dup'd to each
-// sink in IL (`ldc.i4.0; dup; call set_C; dup; call set_B; call set_A`). The importer
-// must re-materialize the dup'd constant at each bool sink so per-sink constant
-// recovery fires and it renders `= false` — not spill it into an int stack slot
-// (`int S = 0; a = S;`), which is CS0029 (cannot implicitly convert int to bool).
+// #2982 / #2994: a value assigned through a chain (`a = b = c = v`) is dup'd to
+// each sink in IL (`ldc.i4.0; dup; call set_C; dup; call set_B; call set_A`).
+// ChainedAssignmentPass recomposes that run into a single `A = B = C = v;`,
+// keyed on the shared dup slot (real evidence — genuinely separate statements
+// carry no dup and never collapse). A dup'd constant that is not part of a chain
+// is re-materialized at its sink so a bool literal is recovered (`= false`), not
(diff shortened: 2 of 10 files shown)


richlander:

Two independent adversarial correctness reviews from different model families, each isolated in its own read-only worktree at the **exact reviewed head `28279a45`** (base `c04c2345`). Scratch confined to `/tmp`; no `git reset`/`add`/`commit` in review trees.

| Reviewer (model) | Verdict | Blocking findings |
| --- | --- | --- |
| Claude Opus 4.8 | **Clean** — ship-able from a correctness standpoint | 0 |
| Gemini 3.1 Pro | **Clean** — no blocking correctness/validity/behavior bugs | 0 |

### Findings and resolution

- **No blocking findings** from either reviewer. Between them they compiled ~20+ adversarial fixtures and confirmed every one produced **valid + behavior-preserving** C#.
- **Non-blocking (Gemini): missed recomposition for covariant reference chains** (`object A = string B = "x"`). This is a deliberate scope limitation, not a bug: the strict `IsAssignable` type-staircase safely falls back to valid separate statements (`B = "x"; A = "x";`). Opus independently reproduced the same case and classified the fallback as correct/conservative. **Non-action:** out of scope for this static-target slice; the safe-fallback behavior is intended. Left as a future enhancement.

### Independently verified by both reviewers (concrete repros with actual output)

- **No false positives** — statements sharing a constant with no dup (`P=7; Q=7`) stay separate; interleaved side effects break the greedy walk.
- **Evaluation order / side effects preserved** — side-effecting setter chains and value-side effects (`Bump()` called once) render in correct C# order; chains stay between surrounding statements.
- **Full-consumption guard** — an escaping/expression-position chain (`Console.WriteLine(P = Q = 5)`) correctly falls back to slot statements, never an unsound expression-position fold.
- **Type staircase** — implicit numeric widening recomposes; narrowing, unsigned-mismatch (`Int32→UInt64`), covariant-ref, decimal, and enum cases fall back to valid C#.
- **#2982 fallback validity preserved** — escaping constant dups (`WriteLine(A = true)`) recover the literal with no int-spill and no double-materialization.
- **Instance/local chains** — untouched (existing local-temp lowering), consistent with the declared scope boundary.

### The one gate neither reviewer could run locally

Both reviewers independently hit the same environmental wall as the author: the slow **CfgSampleClass** compile-back fidelity gate (`FidelityGateTests`/`LoweredFidelityGateTests`) is too slow to finish under current machine load (>21 min, killed). Both noted the new fixtures live on `ChainedConstantAssignmentSamples`, **not** the gate's `CfgSampleClass` corpus, so they cannot perturb it; and a direct probe confirmed **0 constant-dup shapes among 613 CfgSampleClass methods**, so those methods import byte-identically. **This gate is deferred to CI for formal confirmation.**

richlander:
Ready to merge

- Two clean adversarial reviews (Claude Opus 4.8 + Gemini 3.1 Pro) at the exact head `28279a45`, both with zero blocking findings — reconciled above.
- CI green on the merge path: `build-net10`, `csharp-diff-smoke`, and `test (ubuntu-26.04, linux-x64)` all pass. The `test` job runs the full decompiler suite (including the slow corpus/fidelity gates that were blocked locally by machine contention), providing the formal fidelity confirmation.
- `mergeStateStatus: CLEAN`.

PR #3017, opened 2026-07-22
Keep scattered switch-expression default reachable by trampoline (#2973)

### LINKED ISSUE #2973: CS0161: TryNormalizeContinueCondition result-temp switch-expression missing return
`YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition` self-decompiles to CS0161 "not all code paths return a value". It is the one missing-return defect that #2959 (PR #2972) does **not** resolve, and it remains pinned in `ValidityCoverageReportingTests.DecompilerAssembly_MissingReturnPopulationIsPinned`.

## Why #2959 does not cover it

#2959 fixes *scattered shared returns* — a shared `return` **terminator** block reached by conditional guards at different nesting levels, which `StructuringPass` now duplicates into each guard.

`TryNormalizeContinueCondition` is a **different shape**: a switch expression with a **result temp**. Its default arm is `V_2 = false` (a `StoreLocal`, *not* a `Return` terminator) that falls through to a shared `return V_2` tail. `ReturnSinkingPass` (which runs *after* structuring) later distributes `V_2 = X; … ; return V_2` into `return X`. Because the scattered edge terminates in a store, not a return, the #2959 return-terminator discriminator does not apply.

## Acceptance

- `TryNormalizeContinueCondition` self-decompiles to valid C# (all paths return).
- Remove it from the `DecompilerAssembly_MissingReturnPopulationIsPinned` Release pin.
- No regression in the corpus PR quick gate or the decompiler fast suite.

richlander:
- Fixes #2973
- Changes StructuringPass so a switch-expression/pattern default arm reached by a forward-goto trampoline survives as the region's trailing terminator, fixing a CS0161 fall-off-the-end.
- Card revision: `8f61b28d`

## Change

**Conclusion:** **PASS** — the `_ => false` default of a result-temp switch expression, reached by 3 scattered `when`-guard failures **plus** 1 forward-goto trampoline (`IL_002B: br.s IL_0062`), was dropped by the sibling arm's past-region inline, so the trampoline path fell off the end (CS0161); the fix keeps that shared default as the trailing terminator, and both affected methods now bind at Full.

### Original source

Verbatim committed C# at `8f61b28d`, from the SourceLink URL the tool resolves for this member (`https://raw.githubusercontent.com/richlander/dotnet-inspect/8f61b28d/src/ILInspector.Decompiler/Pipeline/Passes/YieldBreakLoopIteratorReconstruction.cs`, lines 259-270):

```csharp
static bool TryNormalizeContinueCondition(IrExpression expression, string loopFieldName, out IrExpression condition)
{
    condition = null!;
    return expression switch
    {
        Comparison comparison when ReadsLoopField(comparison.Left, loopFieldName) => Assign(comparison, out condition),
        LogicalNot { Operand: Comparison comparison } when ReadsLoopField(comparison.Left, loopFieldName)
            => Assign(Negate(comparison), out condition),
        _ => false,
    };
}
```

### Before

```csharp
private static bool TryNormalizeContinueCondition(IrExpression expression, string loopFieldName, out IrExpression condition)
{
    Comparison V_1;

    condition = null;
    IrExpression V_3 = expression;
    Comparison comparison = V_3 as Comparison;
    if (comparison is null)
    {
        if (V_3 is LogicalNot V_4)
        {
            V_1 = V_4.Operand as Comparison;
            if (V_1 is not null)
            {
                if (ReadsLoopField(V_1.Left, loopFieldName))
                {
                    return Assign<IrExpression>(Negate(V_1), out condition);
                }
                return false;
            }
        }
    }
    else
    {
        if (ReadsLoopField(comparison.Left, loopFieldName))
        {
            return Assign<IrExpression>(comparison, out condition);
        }
    }
}
```

- Valid: **False**
- Correct: **False**

Falls off the end: the `comparison is null` arm has no landing when `V_3` is not a `LogicalNot` or `V_1` is null, and the `else` arm has no landing when `ReadsLoopField` is false. `csc` rejects it with CS0161 ("not all code paths return a value"), so it cannot bind and therefore cannot preserve behavior.

### After

```csharp
private static bool TryNormalizeContinueCondition(IrExpression expression, string loopFieldName, out IrExpression condition)
{
    Comparison V_1;
    LogicalNot V_4;

    condition = null;
    IrExpression V_3 = expression;
    Comparison comparison = V_3 as Comparison;
    if (comparison is null)
    {
        V_4 = V_3 as LogicalNot;
        if (V_4 is null)
        {
            return false;
        }
        V_1 = V_4.Operand as Comparison;
        if (V_1 is not null)
        {
            if (ReadsLoopField(V_1.Left, loopFieldName))
            {
                return Assign<IrExpression>(Negate(V_1), out condition);
            }
            return false;
        }
    }
    else
    {
        if (!ReadsLoopField(comparison.Left, loopFieldName))
        {
            return false;
        }
        return Assign<IrExpression>(comparison, out condition);
    }
    return false;
}
```

- Valid: **True**
- Correct: **True**

Every path now returns. The trailing `return false;` is the shared `_ => false` default catching the trampoline (`V_1 == null`) path; the pattern arms return the same `Assign(...)`/`false` values as the source switch, so observable behavior is preserved. Not fully raised — the switch-expression spelling is not recovered (see Fully raised).

### Fully raised

```csharp
condition = null;
return expression switch
{
    Comparison comparison when ReadsLoopField(comparison.Left, loopFieldName) => Assign(comparison, out condition),
    LogicalNot { Operand: Comparison comparison } when ReadsLoopField(comparison.Left, loopFieldName)
        => Assign(Negate(comparison), out condition),
    _ => false,
};
```

- Required tracking issue: #3022 — recover the `switch` expression (with `when`-guarded `Comparison` / `LogicalNot { Operand: Comparison }` arms and a `_ => false` default) from the raised result-temp pattern if/else.

## Root cause and scope

- **Root cause:** The switch-expression default `IL_0062` is reached by 3 scattered `when`-guard failures **plus** 1 forward-goto trampoline (`IL_002B: br.s IL_0062`). #2959's scattered-return-dispatch collector excluded it because the trampoline makes it an unconditional-goto target; and even once admitted, the sibling arm's past-region-target inline cloned the default into its own guard and marked it droppable, so the trampoline arm's fall-through had no landing block.
- **Fix shape (StructuringPass, 4 coherent edits):**
  1. Collect `branchPredecessorIndices` + `switchTargets`.
  2. Admit a shared return as a scattered dispatch target when **every** unconditional predecessor is a pure `{ goto default }` trampoline (never a switch/jump-table target) — those arms dissolve by inlining the terminator, so duplicating the shared return strands no goto (`UnconditionalPredecessorsAreDissolvableTrampolines`).
  3. `IsSharedTerminator`: the scattered classification overrides the general unconditional-target exclusion.
  4. Keep a scattered dispatch target out of **both** drop mechanisms — the upfront droppable set and the past-region-target dynamic drop — so it stays as the region's trailing terminator (Validate already walks it as the merge) while the guards inline duplicates.
- **Scope boundary:** Structuring's all-or-nothing `Validate` remains the backstop — if any trampoline cannot be dissolved, the whole slice stays in its valid goto form rather than dropping an edge. No change to Fully Raised semantics or source-bundle verification; recovering the `switch`-expression spelling is tracked in #3022.

## Evidence

| Check | Baseline (`4f0c0ad1`) | Head (`8f61b28d`) |
| --- | --- | --- |
| Focused test `DecompilerAssembly_MissingReturnPopulationIsPinned` | Pass (8-entry Release pin) | Pass (6-entry Release pin) |
| Decompiler fast suite (`-notrait Speed=Slow`) | 3095, 0 failed | 3095, 0 failed |
| Real witness `YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition` | CS0161 (invalid Full) | Valid, Full |
| Real witness `FixedArrayRaising::SameLoadPlace` | CS0161 (invalid Full) | Valid, Full |
| `git diff --check` | — | clean |

No Baseline failures beyond known pre-existing local-only Slow artifacts (see Out of scope), which are unchanged by this PR (same tests, same reason) and green on `main` CI at the same base.

## Decompiler quality

**Conclusion:** **PASS** — the pinned missing-return validity census over the whole `ILInspector.Decompiler` assembly drops exactly the two fixed methods and gains none; no new fidelity diffs.

### Pinned validity gate

`ValidityCheck.Evaluate` over the decompiler assembly (cap `int.MaxValue`), Release configuration — the authoritative corpus-wide CS0161 gate for this change:

| --- | ---: | ---: | ---: |
| CS0161 malformed methods (-) | 8 | 6 | -2 |
| Newly malformed vs Baseline (-) | - | 0 | 0 |
| Defect methods fixed vs Baseline (+) | - | 2 | +2 |

> **Conclusion:** **PASS** — 2 resolved (`TryNormalizeContinueCondition`, `SameLoadPlace`), 0 new.

### Aggregate context

This fix only adds a trailing terminator to the two trampoline-default methods; it introduces **zero** new fidelity diffs. Both adversarial reviewers re-ran the decompiler fast suite (3095 pass) and the `FidelityGateTests` gate on base vs head and confirmed an identical flagged-diff set. The authoritative corpus fidelity/validity gates run in CI (green on `8f61b28d`).

- Pre-existing local-only failures, unrelated to this change (both confirmed identical on clean base, `main` CI green on the same base):
  - #2958 `LadderIteratorGateTests.IteratorFixture_RecoveredRowsCompileBackOpcodeExact` (iterator `OperandDiff`).
  - `FidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket` local lambda-cache ordinal drift (`StatementBodyLambdaInsideIf` `b__100_0`→`b__121_0`), a daily-SDK artifact absent under CI's pinned SDK.
- Fully raising the `switch`-expression spelling — tracked in #3022.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -method "*MissingReturnPopulationIsPinned*"
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -notrait "Speed=Slow"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ValidityCoverageReportingTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ValidityCoverageReportingTests.cs
@@ -50,9 +50,14 @@ public class ValidityCoverageReportingTests
         // UnionSwitchExpressionPass::SameTailNode,
         // DynamicCallSitePass::TryGuardCacheLoad, and
         // FluentChainRecompositionPass::SinkStatement.
-        // YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition is a
-        // distinct result-temp switch-expression shape the fix does not cover and
-        // remains pinned as tracked follow-up.
+        // #2973 removed the last two scattered-return defects,
+        // YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition and
+        // FixedArrayRaising::SameLoadPlace: result-temp switch-expression /
+        // pattern defaults reached by a forward-goto trampoline. The default
+        // return is now admitted as a scattered dispatch target and kept as the
+        // region's trailing terminator (rather than dropped by the sibling arm's
+        // past-region inline), so the trampoline path returns instead of falling
+        // off the end.
         string[] expected =
 #if DEBUG
         [
@@ -64,10 +69,8 @@ public class ValidityCoverageReportingTests
             "ILInspector.Decompiler.Pipeline.BooleanFoldingPass::IsNullableCoalesceExpressionContext",
             "ILInspector.Decompiler.Pipeline.CSharpPrinter::ForLoopIncrementText",
             "ILInspector.Decompiler.Pipeline.DeconstructionAssignmentPass::TryMatchTupleSeed",
-            "ILInspector.Decompiler.Pipeline.FixedArrayRaising::SameLoadPlace",
             "ILInspector.Decompiler.Pipeline.IndexFromEndPass::LengthReceiver",
             "ILInspector.Decompiler.Pipeline.InlineArrayCollectionPass::PlaceFromAddress",
-            "ILInspector.Decompiler.Pipeline.YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition",
         ];
 #endif
         Assert.Equal(expected, actual);
--- a/src/ILInspector.Decompiler/Pipeline/Passes/StructuringPass.cs
+++ b/src/ILInspector.Decompiler/Pipeline/Passes/StructuringPass.cs
@@ -105,6 +105,8 @@ public sealed class StructuringPass : IIrPass
         var unconditionalTargets = new HashSet<int>();
         var conditionalTargetCounts = new Dictionary<int, int>();
         var conditionalPredecessorIndices = new Dictionary<int, List<int>>();
+        var branchPredecessorIndices = new Dictionary<int, List<int>>();
+        var switchTargets = new HashSet<int>();
         var branchTargets = new HashSet<int>();
         for (int blockIndex = 0; blockIndex < blocks.Count; blockIndex++)
         {
@@ -114,6 +116,9 @@ public sealed class StructuringPass : IIrPass
                 if (child is Branch branch)
                 {
                     unconditionalTargets.Add(branch.TargetOffset);
+                    if (!branchPredecessorIndices.TryGetValue(branch.TargetOffset, out var branchPreds))
+                        branchPredecessorIndices[branch.TargetOffset] = branchPreds = new List<int>();
+                    branchPreds.Add(blockIndex);
                     branchTargets.Add(branch.TargetOffset);
                 }
                 else if (child is Leave leave)
@@ -134,6 +139,7 @@ public sealed class StructuringPass : IIrPass
                     foreach (int target in switchBranch.TargetOffsets)
                     {
                         unconditionalTargets.Add(target);
+                        switchTargets.Add(target);
                         branchTargets.Add(target);
                     }
                 }
@@ -167,11 +173,25 @@ public sealed class StructuringPass : IIrPass
         // guard. The contiguous `&&`/`||` false-exit chain (adjacent guards, no
         // arm between them) is deliberately excluded so it still combines into a
         // single condition (the #640 fidelity canary).
+        //
+        // A shared return is normally excluded when an unconditional goto also
+        // targets it — dropping the duplicated original would strand that goto.
+        // But a switch-expression default arm (`_ => false`) is reached both by
+        // scattered `when`-guard failures AND by a forward-goto trampoline
+        // (`{ goto default; }`, the else of an inner pattern test): the store-
+        // then-return default is the same shared tail (issue #2973). Such a
+        // trampoline is itself a dispatch arm that dissolves by inlining the
+        // terminator, so it does not strand a goto; admit the target when every
+        // unconditional predecessor is one of those pure forward trampolines
+        // (never a switch dispatch target). Structuring's all-or-nothing
+        // Validate is the backstop: if any trampoline cannot be dissolved, the
+        // whole slice stays in its valid goto form rather than dropping an edge.
         var scatteredReturnDispatchTargets = new HashSet<int>();
         foreach (var block in blocks)
         {
             int offset = block.StartOffset;
-            if (unconditionalTargets.Contains(offset)
+            if ((unconditionalTargets.Contains(offset)
+                    && !UnconditionalPredecessorsAreDissolvableTrampolines(blocks, offset, branchPredecessorIndices, switchTargets))
                 || fallenInto.Contains(offset)
                 || conditionalTargetCounts.GetValueOrDefault(offset) < 2
                 || !IsTerminatorBlock(block)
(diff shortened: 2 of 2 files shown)


richlander:

Two-model adversarial review at exact head `8f61b28d` (isolated read-only worktrees). **Both clean, no findings.**

| Reviewer | Model | Result |
| --- | --- | --- |
| Adversarial 1 | GPT (gpt-5.6-sol) | Clean — no significant issues |
| Adversarial 2 | Gemini Pro (gemini-3.1-pro-preview) | Clean — no correctness/soundness/regression issues |

**GPT verified:** full Release build; missing-return pin passes (1/1); both witnesses decompile as valid Full C# with trailing `return false;`; structuring/loop/goto-scope regression tests (34/34); guard-chain canaries (3/3); Validate and BuildRegion remain aligned with no duplicate terminator emission or stranded trampoline.

**Gemini Pro verified:**
- **Trampoline admission soundness** — `UnconditionalPredecessorsAreDissolvableTrampolines` requires each unconditional predecessor to be exactly `[Branch offset]` (no skipped side effects) and excludes `switchTargets` (jump-table/`goto case` dispatch). Edge case: a `Leave` targeting the default is added to `unconditionalTargets` but omitted from `branchPredecessorIndices`, so the default is still shielded from `droppable` and survives as a lexical trailing terminator — no stranded goto, no CS0161 omission.
- **Drop-suppression** — keeping the scattered default out of both the upfront `droppable` set and the dynamic `ctx.DroppableBlocks` means guards inline a duplicate via `IsInlinableTerminator` while `BuildRegion` emits the original exactly once at its trailing position. Mutually exclusive paths, no duplicated statements.
- **No general-case regression** — the new collections are inert when no trampoline-reached default exists; ran the full fast decompiler suite (3095 pass).
- **Validate/BuildRegion alignment** — both phases thread `ScatteredReturnDispatchTargets` through `IsInlinableTerminator`/`IsSharedTerminator` and agree on the trailing default as the join.
- **Pin honesty** — both removed entries genuinely bind at Full with a real `return false;`; the all-or-nothing pin confirms no other method regressed.

**CI:** green on `8f61b28d` (`build-net10`, `test`, `csharp-diff-smoke` all pass).

No findings to reconcile; no fixed-head re-review required since neither reviewer raised a blocking issue.

richlander:
Ready to merge

Issue #3018, opened 2026-07-22
Windows-local test failures: CRLF stdout splitting, forward-slash path assertion, and native-DLL CS0009 in recompile reference set


richlander:
## Summary

Several tests fail **only on Windows** (CI is green on Linux) due to
platform-robustness gaps in the *tests* and one recompile-reference gap in the
*tools* harness — none reflect a product defect. This is a single tracking
issue for the Windows-local failures observed while running the full local
suites on the .NET 11 daily SDK. They fall into three independent root causes.

Environment:

- SDK: `.NET 11.0.100-preview.6.26359.118` (daily)
- OS: Windows 10.0.26200 (x64)
- Base: `main` at `c04c2345` (post-#3001)

## Root cause A — tests `Split('\n')` CLI stdout, leaving a trailing `\r`

`dotnet-inspect.Tests` splits captured CLI output on `'\n'`; on Windows the CLI
emits CRLF, so each token retains a trailing `\r` and exact/`EndsWith`
assertions fail. Four tests:

- `PackageVersionTests.Versions_WithRange_ListsTheInclusiveAddressVector`
  — expected `"8.0.0"`, actual `"8.0.0\r"`.
- `CommandExecutionTests.Project_SkillsPaths_EmitsSkillPaths`
  — expected `"skills/SKILL.md"`, actual `"skills/SKILL.md\r"`.
- `CommandExecutionTests.Package_SourceFilesSection_Bare_EmitsUrlColumn`
  — `line.EndsWith("/Src/Newtonsoft.Json/JsonReader.cs")` fails on `...JsonReader.cs\r`
  (this is a CRLF artifact, **not** a network/URL problem).
- `CommandExecutionTests.Type_SourceFiles_Bare_EmitsUrlColumn`
  — same `.cs\r` `EndsWith` failure.

Fix: split on the platform newline or normalize (`ReplaceLineEndings`) before
tokenizing, e.g. `output.Split('\n', ...)` → split on `Environment.NewLine` /
`.Trim('\r')` per token. Same family as the already-fixed decompiler-side
CRLF issue #2941 (closed), but in a different suite.

## Root cause B — path assertion uses forward slashes vs Windows backslashes

`ILInspector.Metadata.Tests.MetadataSourceFindingsTests.MemberSourceProducer_UsesFirstVisibleDocumentWhenRootIsOmitted`
[...]

Issue #2996, opened 2026-07-22
Decompiler should render whole members (signature + body): unify the signature-shape model and retire per-marking double-derivation


richlander:
## Summary

The decompiler's unit of output is the **method body**, not the **member** — so the
signature is read and printed by a *second, independent path*, and the two must agree by
parallel logic. The product stitches a separately-decoded Metadata header onto the
decompiler body, and the compile-back harness maintains an entire "skeleton" implementation
to rebuild the signatures the decompiler will not emit. This split is the root cause of a
recurring family of problems: duplicated derivation of signature facts, header/body
mismatches that are hard to resolve, and a second signature implementation with its own bugs.

**The fix is not to move signature ownership into the decompiler.** Ownership stays where it
belongs; the decompiler owns only the body:

- **Metadata** *reads* the signature into a **rich data model** — structured types plus the
  shape facts (dynamic/nullable/tuple names/native int/in/scoped/volatile/init/callconv/
  params/defaults) plus accessibility and modifiers. One decode.
- **CSharp** *prints* the signature from that model. One spelling.
- **Decompiler** owns the **body** (reads it, raises it, prints it). Its natural output is the
  **whole member**, which it composes by handing Metadata's rich model to CSharp for the
  signature and printing the body itself.

The rich model flows **Metadata → Decompiler → CSharp**, entirely as an implementation detail
invisible to the caller. The benefit is one implementation for everything and no drift: the
*same* signature is produced whether reached through **Metadata** (structured facts, no C#
spelling), **CSharp** (spelled), or the **Decompiler** (spelled signature + body). "Any
disagreement is a bug" holds by construction, because there is one read and one print.

`dynamic` recovery (#2984) is the motivating first case. Dropping the `(dynamic)` cast in the
body is only valid if the signature renders `dynamic` too — and today the body's operand type
and the header's spelling come from two different decodes, so the body change alone yields
invalid C# (`public void AssignToField(object value) => _last = value.Length;`, CS1061).

## Root cause

For a decompiled member, three layers cooperate but the signature is decoded twice:

- **Metadata** (`ApiSurface` → `ApiType`/`ApiMember`/`ApiParameter`) extracts the API model.
  Type views are **strings** (`ApiParameter.Type = "object"` at `ApiSurface.cs:236`,
  `ApiMember.ReturnType` at `ApiSurface.cs:208`), plus accessibility, modifiers, generic
  params/constraints, default values — but pure-view shape facts such as `dynamic` are lost
[...]

Issue #2984, opened 2026-07-21
Decompiler: recover dynamic type view to drop redundant (dynamic) cast on raised dynamic member access


richlander:
## Summary

Dynamic member access that this decompiler raises (e.g. `((dynamic)value).Length`) is spelled with an explicit `(dynamic)` cast on the operand, whereas the original source wrote `value.Length` with `value` declared `dynamic`. The two are semantically equivalent, but the cast form is a fidelity gap. This tracks recovering the operand's `dynamic` type view so the cast can be dropped where appropriate.

Orthogonal to #2917 / #2968 / #2980, which only decide **whether** to raise a dynamic call site — not **how** a `dynamic` operand is spelled once raised.

## Current behavior

For:

```csharp
public class Host<T>
{
    public Func<int> InGenericLambda(dynamic value) => () => value.Length;
}
```

the raised output is:

```csharp
return ((dynamic)value).Length;
```

The `(dynamic)` cast appears because the operand's recovered static type is `object`. In the lowered display class the captured operand is hoisted into a field typed `object`, and the decompiler currently types the recovered local as `object`, so it re-spells the dynamic-ness at the use site with a cast.

## Root cause

The cast is forced by the **operand's recovered static type**, not by anything intrinsic to the dynamic call site. Dynamic dispatch itself needs no member signature — the member name (`"Length"`) is a string handed to `Binder.GetMember` and resolved at runtime — so a `dynamic`-typed operand binds `value.Length` with no cast.

Crucially, the marker needed to recover `dynamic` is present: the hoisted display-class field carries `[DynamicAttribute]`:

```text
FIELD value : Object attrs=[DynamicAttribute]
```

So the information to type the operand as `dynamic` (and drop the cast) is already in metadata; the decompiler just doesn't consume it here.

## Proposed approach

Consume `[DynamicAttribute]` on the source of the recovered operand (parameter, hoisted display-class field, local) to give it a `dynamic` type view, then print `value.Length` instead of `((dynamic)value).Length`.

Per repository layering, C# spelling and type-view composition belong to `ILInspector.CSharp` (Metadata owns metadata facts; the printer owns C# spelling). This is a type-view/printer concern, not a `DynamicCallSitePass` (raise-decision) concern.

## Caveats

- The cast remains **genuinely required** and must be preserved when only one operand of a mixed expression is dynamic, or when the source operand was truly `object` (no `[DynamicAttribute]`). This change must be gated on the attribute, not applied unconditionally.
[...]

PR #3019, opened 2026-07-22
Decode [DynamicAttribute] into Metadata type views (#2984)

richlander:
## Summary

Metadata's `TypeNode` type-view tree carried nullability but had **no dynamic
decoding**, so every `[DynamicAttribute]` position rendered `object`. This
dropped the `dynamic` shape fact from both the `member` API header **and** the
decompiled member header (which reuses Metadata's type view), forcing
downstream consumers to re-cast to reach dynamic members.

This PR teaches Metadata to decode `[DynamicAttribute]` transform flags and
spell `dynamic` / `dynamic?` at the marked positions. Because the decompiler's
signature rendering reuses Metadata's type view, one fix corrects **both**
headers — the "one implementation, no drift" payoff of the #2996 signature-model
unification.

Foundation for #2984 (the decompiler dropping the redundant `((dynamic)value)`
body cast is the follow-up consumer PR). Umbrella: #2996.

## What changed

- **`DynamicReader.cs` (new)** — reads `[DynamicAttribute]` transform flags as
  `byte[]?`, mirroring `NullabilityReader` (marker form → `[1]`; array form →
  prolog + count + N bytes). `GetDynamicFlags` + `GetParameterDynamicFlags`.
- **`TypeNode.cs`** — adds `IsDynamic`, an `ApplyDynamic(flags, ref position)`
  traversal paralleling `ApplyNullability`, and `dynamic` / `dynamic?`
  rendering at object-typed positions the flags mark. Non-object nodes still
  consume their flag slot to stay aligned.
- **`ApiSurfaceExtractor.cs`** — wires `ApplyDynamic` into the five signature
  sites (field, method return, method params, property type, indexer params),
  each with an independent flag array.

## Key detail: ByRef consumes a leading dynamic flag

Unlike the nullability traversal, `ByRefTypeNode.ApplyDynamic` **does** consume
a leading (always-false) flag slot. Without it, `ref dynamic` rendered
`ref object`. This divergence is validated by a dedicated `ref dynamic` fixture
in the test suite.

## Identity is unaffected

Member identity uses a separate `AnchorSignatureTypeProvider` that always
renders `System.Object` (never `dynamic`), so member matching and selectors are
orthogonal to this display change. No selector, anchor, or correspondence
behavior changes.

## Evidence

- **`DynamicTypeViewTests` (new): 23/23.** 9 direct `TypeNode.ApplyDynamic`
  unit tests + 14 end-to-end tests over inline **Roslyn-compiled** fixtures as
  the encoding oracle: bare `dynamic` param/return/field/property, `dynamic[]`,
  `List<dynamic>`, `Dictionary<object,dynamic>`, `Dictionary<dynamic,object>`,
  nested `List<Dictionary<string,dynamic>>`, `Func<object,dynamic>`, mixed
  params, `ref dynamic`, and plain `object` stays `object`.
- **Full `ILInspector.Metadata.Tests`: 681/682** — the 1 failure
  (`MetadataSourceFindingsTests.MemberSourceProducer_UsesFirstVisibleDocumentWhenRootIsOmitted`)
  is a pre-existing Windows path-separator bug unrelated to this change (filed
  in #3018; reproduces on clean `main`).
- **`dotnet-inspect.Tests`: 1975/1979** — the 4 failures are pre-existing
  Windows-local CRLF / source-URL issues (#3018), none touch signature
  spelling.
- **Decompiler `Area=Pass` 1165/1169, `Area=Validity` 94/98,
  `Area=RoundTrip` 136/297** — every failure is the pre-existing Windows
  `CS0009` native-DLL recompile-environment issue (#3018 root cause C); no new
  failures. `LadderRung9GateTests` (the dynamic fixtures) is **19/19**.

Manual confirmation on the LadderRung9 fixture DLL: both the `member` API header
and the decompiled member header now render `dynamic value` (previously
`object value`).

## Compatibility / boundary

- Behavior-visible: signatures containing `[DynamicAttribute]` positions now
  spell `dynamic` instead of `object` (strictly more faithful).
- No product path loads inspected assemblies, uses Roslyn, or leaves SRM /
  NativeAOT constraints.
- Follow-up (#2984 proper): the decompiler drops the now-redundant
  `((dynamic)value)` body cast, consuming this same fact.

## Adversarial review

Behavior-visible → requires two non-Claude-Opus model families per AGENTS.md;
isolated review worktrees; two clean fixed-head reviews before "Ready to merge".

### DIFF
--- a/src/ILInspector.Metadata/ApiMemberIdentity.cs
+++ b/src/ILInspector.Metadata/ApiMemberIdentity.cs
@@ -385,12 +385,17 @@ public static class ApiMemberIdentity
         var memberName = member.Kind == "constructor"
             ? "#ctor"
             : ExtractMemberNameWithGeneric(signature, member.Name);
-        var parameters = ExtractCanonicalParameterList(signature);
+        // Raw-signature fallback (used when SignatureModel is absent, e.g. after a JSON
+        // round-trip where SignatureModel is [JsonIgnore]). member.Signature is the
+        // display string and carries `dynamic`, so scrub it back to `object` for identity
+        // exactly as the SignatureModel path does — otherwise a round-tripped member's
+        // fingerprint diverges from the same member read live.
+        var parameters = NormalizeDynamicToObject(ExtractCanonicalParameterList(signature));
         var canonical = $"{kindCode}:{declaringType}.{memberName}{parameters}";
         // Mirror the conversion-operator return-type disambiguation so member identity
         // is not dependent on whether SignatureModel was populated (the Try path above).
         if (IsConversionOperator(member.Name) && !string.IsNullOrWhiteSpace(member.ReturnType))
-            canonical += $"~{NormalizeCanonicalCommas(member.ReturnType!)}";
+            canonical += $"~{NormalizeCanonicalCommas(NormalizeDynamicToObject(member.ReturnType!))}";
         return canonical;
     }
 
@@ -432,7 +437,7 @@ public static class ApiMemberIdentity
             // round-trips.
             var indexerParameters = member.SignatureModel is { Parameters.Count: > 0 } propertySignature
                 ? NormalizeCanonicalParameters(propertySignature.ParameterTypesSummary)
-                : ExtractCanonicalIndexerParameterList(member.Signature);
+                : NormalizeDynamicToObject(ExtractCanonicalIndexerParameterList(member.Signature));
             canonicalSignature = $"{kindCode}:{declaringType}.{member.Name}{indexerParameters}";
             return true;
         }
@@ -532,8 +537,18 @@ public static class ApiMemberIdentity
     /// and matching. <c>dynamic</c> and <c>object</c> are the same metadata type, so the
     /// display-only distinction must never reach canonical signatures, correspondence
     /// keys, or XML-doc identity (which encode dynamic positions as <c>System.Object</c>).
+    /// This runs in string space by necessity: the display signature (<c>member.Signature</c>)
+    /// is serialized while the typed <c>SignatureModel</c> is <c>[JsonIgnore]</c>, so a
+    /// round-tripped member has only the rendered string to derive identity from — the same
+    /// reason nullability is normalized as text (see <see cref="NormalizeXmlDocParameterType"/>).
     /// Boundary-aware so it never rewrites a dotted name segment or a longer identifier
-    /// (e.g. <c>System.Dynamic.X</c> or <c>MyDynamicType</c> are left untouched).
+    /// (e.g. <c>System.Dynamic.X</c>, <c>Ns.dynamic</c>, or <c>MyDynamicType</c> are left
+    /// untouched). Known limitation: a type literally named <c>dynamic</c> in the global
+    /// namespace (C# <c>@dynamic</c>) renders as a bare <c>dynamic</c> token indistinguishable
+    /// from the keyword once the typed model is gone, so its identity collapses to
+    /// <c>object</c>. That is astronomically rare and the trade is deliberate — preserving
+    /// fingerprint stability for the ubiquitous keyword case outweighs a global type named
+    /// after a contextual keyword.
     /// </summary>
     static string NormalizeDynamicToObject(string value)
     {
--- a/src/ILInspector.Metadata/ApiMemberIdentity.cs
+++ b/src/ILInspector.Metadata/ApiMemberIdentity.cs
@@ -457,7 +457,7 @@ public static class ApiMemberIdentity
         // same "~ReturnType" delimiter as XML doc identity so conversion anchors
         // and XML lookups do not grow divergent spellings for the same fact.
         if (IsConversionOperator(member.Name) && !string.IsNullOrWhiteSpace(signature.ReturnType))
-            canonical += $"~{NormalizeCanonicalCommas(signature.ReturnType!)}";
+            canonical += $"~{NormalizeCanonicalCommas(NormalizeDynamicToObject(signature.ReturnType!))}";
         canonicalSignature = canonical;
         return true;
     }
@@ -523,9 +523,45 @@ public static class ApiMemberIdentity
     }
 
     static string NormalizeCorrespondenceType(string type)
-        => type.Trim()
+        => NormalizeDynamicToObject(type.Trim())
             .Replace("+", ".", StringComparison.Ordinal)
             .Replace(", ", ",", StringComparison.Ordinal);
+
+    /// <summary>
+    /// Collapses the display keyword <c>dynamic</c> back to <c>object</c> for identity
+    /// and matching. <c>dynamic</c> and <c>object</c> are the same metadata type, so the
+    /// display-only distinction must never reach canonical signatures, correspondence
+    /// keys, or XML-doc identity (which encode dynamic positions as <c>System.Object</c>).
+    /// Boundary-aware so it never rewrites a dotted name segment or a longer identifier
+    /// (e.g. <c>System.Dynamic.X</c> or <c>MyDynamicType</c> are left untouched).
+    /// </summary>
+    static string NormalizeDynamicToObject(string value)
+    {
+        const string token = "dynamic";
+        if (value.IndexOf(token, StringComparison.Ordinal) < 0)
+            return value;
+        var builder = new System.Text.StringBuilder(value.Length);
+        var index = 0;
+        while (index < value.Length)
+        {
+            if (index + token.Length <= value.Length
+                && string.CompareOrdinal(value, index, token, 0, token.Length) == 0
+                && (index == 0 || !IsTypeNameChar(value[index - 1]))
+                && (index + token.Length >= value.Length || !IsTypeNameChar(value[index + token.Length])))
+            {
+                builder.Append("object");
+                index += token.Length;
+            }
+            else
+            {
+                builder.Append(value[index]);
+                index++;
+            }
+        }
+        return builder.ToString();
+
+        static bool IsTypeNameChar(char c) => char.IsLetterOrDigit(c) || c == '_' || c == '.';
+    }
 
     // Preserve the v1 Member Index digest contract for members that already have
     // compatibility signature text. The legacy parser had edge-case behavior
@@ -567,7 +603,7 @@ public static class ApiMemberIdentity
     static string NormalizeCanonicalParameters(string parameterTypesSummary)
         => string.IsNullOrEmpty(parameterTypesSummary)
             ? "()"
-            : NormalizeCanonicalCommas(parameterTypesSummary);
+            : NormalizeCanonicalCommas(NormalizeDynamicToObject(parameterTypesSummary));
 
     static string NormalizeCanonicalCommas(string value)
         => value.Replace(", ", ",", StringComparison.Ordinal).Trim();
@@ -796,7 +832,7 @@ public static class ApiMemberIdentity
         IReadOnlyDictionary<string, int> typeParameterMap,
         IReadOnlyDictionary<string, int> methodParameterMap)
     {
-        var type = StripLeadingAttributes(parameter.Trim());
+        var type = StripLeadingAttributes(NormalizeDynamicToObject(parameter).Trim());
         var isByRef = false;
         foreach (var prefix in (string[])["ref ", "out ", "in ", "params ", "this "])
         {
--- a/src/ILInspector.Metadata/ApiMemberIdentity.cs
+++ b/src/ILInspector.Metadata/ApiMemberIdentity.cs
@@ -162,19 +162,7 @@ public static class ApiMemberIdentity
     public static ApiMemberHandle CreateHandle(ApiType type, ApiMember member)
         => new(type, member, GetMemberAnchor(type, member));
 
-    /// <summary>
-    /// Ordering key for overloads, also the basis for the positional <c>Name:N</c> member
-    /// selector (see <see cref="MemberTargetResolver"/>). The raw display signature is
-    /// normalized so the display-only <c>dynamic</c>/<c>object</c> spelling never perturbs
-    /// overload ordering: <c>dynamic</c> and <c>object</c> are the same metadata type, so a
-    /// type can never declare both <c>M(object)</c> and <c>M(dynamic)</c> (CS0111), and
-    /// ordering the keyword position as <c>object</c> keeps the ordinal stable across the
-    /// object-to-dynamic display change.
-    /// </summary>
     public static string GetMemberSignatureSortKey(ApiMember member)
-        => NormalizeDynamicToObject(GetRawMemberSignatureSortKey(member));
-
-    static string GetRawMemberSignatureSortKey(ApiMember member)
     {
         var signature = member.Signature ?? "";
         if (signature.Length == 0 || member.Name.Length == 0)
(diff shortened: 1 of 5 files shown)


richlander:
## Adversarial review reconciliation (fixed head `2336b9b1`)

Two reviewers ran on the prior head `d5a7d44e`: **GPT-5.6 Sol** (5 findings) and **Gemini 3.1 Pro** (events + attack-point evaluation). I empirically confirmed every disputed point with Roslyn-compiled fixtures + a raw `DynamicAttribute.TransformFlags` reflection probe. Where the two reviews appeared to disagree, they were examining **different code paths** (e.g. the SRM-direct `AnchorSignatureTypeProvider` — which Gemini verified renders `System.Object` and is already isolated — vs the `ApiMember`/SignatureModel identity path, which was not). All findings below are real and fixed.

| # | Finding | Reviewer | Verified | Fix |
| F1 | `dynamic` leaked into **canonical signatures / correspondence keys**, changing `MemberAnchor` fingerprints (`(dynamic)`→`e1f44d877b` vs `(object)`→`9b92a023ea`) and breaking persisted selectors | GPT (High) | ✅ hash diff reproduced | `NormalizeDynamicToObject` (boundary-aware) in `NormalizeCanonicalParameters`, conversion-return suffix, `NormalizeCorrespondenceType` |
| F2 | `dynamic` leaked into **XML-doc member identity** (`["dynamic"]`) | GPT (Med) | ✅ | same helper at `NormalizeXmlDocParameterType` entry |
| F3 | `ref readonly dynamic` return mis-decoded — the `modreq(In)` slot was not consumed, so the flag stream misaligned. Probe: return flags `[False,False,True]` (byref, modreq, object) | GPT (Med) | ✅ probe + tree diag | `ModifiedTypeNode.ApplyDynamic` consumes one always-false slot per modifier. `in dynamic` params (no `ModifiedTypeNode`) correctly untouched |
| F4 | **base types, interfaces, events** resolve via the string `TypeResolver` and skipped `ApplyDynamic`, so `EventHandler<dynamic>` / `Base<dynamic>` rendered `object` | GPT + Gemini | ✅ | `ApplyDynamicView` guarded on `TypeSpecification` + present `DynamicAttribute`; wired into base type, interfaces, events (events also re-apply nullability) |
| F5 | marker-form (single-element) `DynamicAttribute` **broadcast** into inner nodes, so adversarial `[Dynamic]` on `object[]` became `dynamic[]` | GPT (Med) | ✅ | `ConsumeDynamicFlag` indexes strictly, returns non-dynamic past the end, no broadcast — spec-faithful and threat-model-aligned |

**Reviewer reconciliation notes:** GPT was right about F3 (it tested `ref readonly`); Gemini's "ByRef CLEAN" was also right (it only tested `ref`/`ref[]`/`out`, no modreq — those have no `ModifiedTypeNode`). GPT was right about F5 broadcast; Gemini called broadcast "intended" but only considered the valid single-node case. Both flagged F4 events. `GetMemberSignatureSortKey` intentionally left alone — dynamic/object erase to identical signatures so they can't be distinct overloads (no misordering possible).

### Evidence
- `DynamicTypeViewTests` **31/31** — new coverage: `Method_InDynamic_ConsumesModifierSlot`, `Method_RefReadonlyDynamicReturn_ConsumesModifierSlot`, `Event_DynamicHandler`, `BaseType_Dynamic`, `Identity_DynamicParam_CanonicalUsesObjectNotDynamic`, `Identity_DynamicParam_XmlDocUsesObject`, `Identity_NestedDynamic_CanonicalUsesObjectNotDynamic`, `MarkerFlag_OnArray_DoesNotBroadcastToElement`.
- Metadata full **690/691** (the 1 = pre-existing Windows path-sep bug, filed #3018-B; unrelated to this change).

Re-reviewing the fixed head with two models now; will post `Ready to merge` only after two clean fixed-head reviews.

richlander:
## Second-round reconciliation (head `e44ef1c6`)

The fixed-head re-reviews of `2336b9b1` each surfaced two more real edge cases. All four are now fixed; none contradict each other.

**GPT-5.6 Sol:**
- **G1 (High) — identity leak survives JSON round-trip.** The F1/F2 normalization only covered the `SignatureModel`-present paths. `member.Signature` is serialized but `SignatureModel` is `[JsonIgnore]`, so a round-tripped member derives its canonical signature by parsing the raw display string (which carries `dynamic`), leaking into the fingerprint. **Fixed** (`192fee3f`): apply `NormalizeDynamicToObject` to the raw-signature fallbacks — method params + conversion return in `GetCanonicalSignature`, and the indexer raw-fallback in `TryGetCanonicalSignature`. New JSON round-trip regression tests (method param + `dynamic this[dynamic]` indexer) assert the round-tripped canonical equals the live one and never spells `dynamic`.
- **G2 (Medium) — a real global `@dynamic` type normalizes to `object`.** Inherent to the dual-purpose string: display requires `dynamic`, and after round-trip there is no typed model to distinguish the keyword from a type literally named `dynamic`. **Documented** as an accepted, astronomically-rare limitation; namespaced (`Ns.dynamic`) and longer identifiers are already spared. The trade favors fingerprint stability for the ubiquitous keyword case. (Gemini's Ge1 then shrank this footprint further — see below.)

**Gemini 3.1 Pro:**
- **Ge1 (High) — boundary check corrupts generic/nested `@dynamic`.** `IsTypeNameChar` omitted the generic-arity backtick and nested separators `+`/`/`, so `` dynamic`1 `` and `Outer+dynamic` were mis-normalized to `object`. **Fixed** (`e44ef1c6`): add `` ` `` `+` `/` to the identifier-char set. This strictly *shrinks* the G2 footprint — the `dynamic` keyword never abuts those chars, so keyword normalization is unaffected (locked by an 11-case boundary Theory).
- **Ge2 (Medium) — degraded TypeSpec decode obliterates the resolved string.** `ApplyDynamicView` (base types/interfaces) and the event decoder rendered the `GuardedProviderDecode.TypeSpec` result unconditionally; a rejected TypeSpec returns a `DegradedTypeNode` whose `Render()` is a bare `object`/`dynamic`, silently replacing e.g. `MyEvent<object>`. **Fixed** (`e44ef1c6`): guard on `node.IsDegraded` and keep the resolved fallback (failure stays visible, per the repo's fail-closed principle).

Both reviewers independently **verified F3 (modreq slot) and F5 (marker no-broadcast) as correctly fixed**, and confirmed `in dynamic` params (bare `ByRefTypeNode`, 2 flags) do not double-consume while `ref readonly dynamic` returns (3 flags, with modreq) align via `ModifiedTypeNode.ApplyDynamic`.

### Evidence (head `e44ef1c6`)
- DynamicTypeViewTests + ApiMemberIdentityTests **80/80** (adds JSON round-trip identity for method param + dynamic indexer, and the keyword-vs-typename boundary Theory).
- Metadata full **703/704** (the 1 = pre-existing Windows path-sep #3018-B, unrelated).

Running a final two-model review on `e44ef1c6`; will post `Ready to merge` only after two clean fixed-head reviews.

richlander:
## Third-round adversarial review reconciliation (head `e44ef1c6` -> `8063d775`)

Two-model final review of `e44ef1c6`: **Gemini (gemini-3.1-pro-preview): CLEAN** ("ready to merge"), independently verifying F1-F5, G1, G2, Ge1, Ge2 all correctly resolved and the degraded-decode guard fails safe. **GPT (gpt-5.6-sol): 3 findings**, reconciled below.

### GPT Issue 3 — overload sort key renumbered `Name:N` selector (Medium) — FIXED in `8063d775`
`ApiMemberIdentity.GetMemberSignatureSortKey` ordered by the raw display signature, which feeds the positional `Name:N` member selector (`MemberTargetResolver.GetCandidates`, `ApiOutputFormatter`). A parameter changing display spelling `object` -> `dynamic` (the same metadata type) shifted its sort position, renumbering the ordinal so an existing `M:2` could resolve to a different overload. **Fix:** normalize `dynamic` -> `object` in the sort key. A type can never declare both `M(object)` and `M(dynamic)` (CS0111), so the collapse is collision-free. Added `MemberSignatureSortKey_DynamicParam_OrdersAsObject`. This closes the last identity-stability gap (companion to F1/F2/G1).

### GPT Issue 1 — `@dynamic` generic parameter collapses to `object` (Medium) — DOCUMENTED (accepted, extends G2)
This is the same fundamental round-trip ambiguity already accepted as G2, extended to a generic parameter named `@dynamic` (not just a global type named `@dynamic`). Once the typed model is gone (post-JSON), `M(@dynamic)` and `M(object)` both render as a bare `dynamic`/`object` token with no way to distinguish keyword from identifier, so identity collapses. A provenance-aware fix is impossible on the round-trip path (only the rendered string survives); a live-path-only fix would make identity **diverge across serialization**, which is worse than consistent collapse. `@dynamic` (an escaped contextual keyword as an identifier) is astronomically rare in real code. The `NormalizeDynamicToObject` doc comment now records this explicitly.

### GPT Issue 2 — `ApiDiffAnalyzer` reports display-only `object`/`dynamic` change as Breaking (Medium) — FILED as #3021 (non-blocking follow-up)
The diff analyzer compares **raw display strings** (`BaseType`, `Interfaces`, `GetMemberKey`) and normalizes **no** display-only distinction. Nullability (`string` vs `string?`) behaves identically today. The spurious diff only appears across the **tool-version boundary** (a pre-#3019 serialized baseline spelling `object` vs a post-#3019 run spelling `dynamic`); when both surfaces come from the same tool version they match cleanly (Gemini independently confirmed this). A `dynamic`-only patch in the diff would be inconsistent with the deliberately display-based design and its nullability behavior. How the diff should treat display-only spelling distinctions holistically (dynamic + nullability + ...) is a product-design decision deserving its own PR — filed as #3021. This is outside PR #3019's Metadata-type-view + identity-insulation charter and does not affect normal-usage correctness.

### Next
Re-reviewing the new fixed head `8063d775` with both models per review discipline; will post `Ready to merge` only after two clean fixed-head reviews.

richlander:
## Fourth-round re-review reconciliation (head `8063d775` -> `933e6cfd`)

Re-review of `8063d775`: **Gemini (gemini-3.1-pro-preview): CLEAN** — verified the sort-key fix collision-free (CS0111 + name-grouping), no identity leaks, refactor behavior-identical, `@dynamic` collapse sound; full validation (Metadata 703/704, CLI 253/253). **GPT (gpt-5.6-sol): 2 findings**, both addressed in `933e6cfd`.

### GPT Issue A — sort-key scrub mangled a parameter *named* `@dynamic` (Medium) — REVERTED
The `8063d775` sort-key normalization rewrote identifiers, not only type positions: a legal parameter named `@dynamic` (display `dynamic`) became `object`, renumbering the positional `Name:N` selector for a method that has nothing to do with the dynamic keyword. It was also inconsistent with the same subsystem's un-normalized treatment of nullability / tuple-element names. **Reverted** `GetMemberSignatureSortKey` to its raw form (and removed the added test). The durable selector is the content-hashed `MemberAnchor.StableSelector`, already `dynamic`-normalized via F1/G1; the positional `Name:N` ordinal is inherently version-fragile (any overload add/remove shifts it). The display-only ordinal-stability concern (original round-3 Issue 3) is folded into follow-up **#3021** alongside the diff analyzer, to be handled uniformly for all display-only distinctions (dynamic + nullability + tuple names), not `dynamic`-only.

### GPT Issue B — generic parameter named `@dynamic` collapses to `object` in canonical identity (Medium) — DOC CORRECTED + FILED #3023
GPT is right that the prior doc claim ("provenance-aware fix impossible on the round-trip path") was **factually wrong** for the type-level case: `ApiType.TypeParameters` survives serialization and is a valid discriminator (a `dynamic` token in a scope that declares a generic parameter named `dynamic` binds to the parameter — the keyword is shadowed). I traced the root cause: every identity site (`GetCanonicalSignature`/`TryGetCanonicalSignature`, `NormalizeCorrespondenceType`, and `NormalizeXmlDocParameterType`) runs the string-space `dynamic`->`object` collapse **ahead of** generic-parameter resolution — e.g. `NormalizeXmlDocParameterType` normalizes at line 868 before `TryNormalizeGenericParameterReference` consults the maps. A correct fix resolves generic-parameter references before the collapse, uniformly across all three sites. **Corrected the `NormalizeDynamicToObject` doc comment** to state this accurately and **filed #3023** with the root cause, the round-trip-safe discriminator, and the residual limitations (global-type-named-`@dynamic` and method-level generic parameter named `@dynamic` have no round-trip-safe discriminator). Deferred rather than patched piecemeal: the demonstrated collision requires an `@dynamic`-named generic parameter overloaded against `object` (astronomically rare), and the correct fix touches three identity subsystems including the common XML-doc identity path.

### Net effect at `933e6cfd`
The product-behavior delta vs the previously-reviewed `e44ef1c6` (Gemini CLEAN; all GPT `e44ef1c6` findings resolved) is **documentation-only**: the sort-key change is reverted, and the `@dynamic` limitation note is now accurate. Metadata 703/703 excluding pre-existing #3018-B; CLI formatter/diff unaffected. Two follow-ups filed: **#3021** (display-only distinctions in display-based subsystems: diff comparison keys + positional selector) and **#3023** (`@dynamic` generic-parameter identity collision).

### Next
Re-reviewing `933e6cfd` with both models; `Ready to merge` only after two clean fixed-head reviews.

richlander:
## Fifth-round fixed-head re-review — both CLEAN

Re-reviewed the current head `933e6cfd` (net product delta vs the previously-validated `e44ef1c6` is documentation-only) with two independent models in isolated fixed-head worktrees. Both are CLEAN with no new findings; each ran the build and tests and independently fact-checked the corrected documentation against the code.

**GPT (gpt-5.6-sol): CLEAN.** Confirmed HEAD `933e6cfd`; the diff from `e44ef1c6` is only the corrected `NormalizeDynamicToObject` XML comment; the sort-key implementation matches `e44ef1c6` exactly with no residue of the reverted change or its test; documentation claims match the serialization and identity code paths; the deferred issues affect display diffing / positional selectors / rare `@dynamic` collisions, not durable identity for ordinary `dynamic` members. Release build succeeded; targeted tests 80/80; full Metadata 702 passed + 1 expected unrelated failure (`MemberSourceProducer_UsesFirstVisibleDocumentWhenRootIsOmitted`, pre-existing Windows path-sep #3018-B).

**Gemini (gemini-3.1-pro-preview): CLEAN.** Verified the only change vs `e44ef1c6` is the expanded doc comment and that `GetMemberSignatureSortKey` returns the raw string (no normalization residue). Fact-checked the doc against the code: all three identity sites (`ExtractCanonicalParameterList`/canonical, `NormalizeCorrespondenceType`, `NormalizeXmlDocParameterType`) invoke `NormalizeDynamicToObject` ahead of generic-parameter resolution; `ApiType.TypeParameters` is serialized (type-level `@dynamic` IS round-trip-detectable) while method-level type parameters live in the `[JsonIgnore]` `ApiSignatureModel` (so method-level `@dynamic` and a global type named `@dynamic` are genuinely undetectable after round-trip). Confirmed durable identity is uncorrupted: contextual `dynamic` (`List<dynamic>`, `dynamic[]`) normalizes to `object` in the canonical signature, so the fingerprint and content-hashed `StableSelector` are unchanged; the deferred issues (#3021 positional selector + diff analyzer; #3023 `@dynamic` collision) are display/heuristic concerns, not identity corruption. Build succeeded; 80/80 targeted; a `NormalizeDynamicToObject` probe over complex signatures confirmed robust token translation.

### CI
Green on `933e6cfd`: `build-net10` pass, `test (ubuntu-26.04, linux-x64)` pass, `changes` pass (path-gated jobs correctly skipping).

### Summary
Two clean fixed-head reviews from different model families + green CI. All prior findings are resolved or filed as justified follow-ups (#3021, #3023). The `dynamic` spelling does not perturb any durable member identity.

Ready to merge

PR #3031, opened 2026-07-22
Honor verified local source and direct private members for member Original Source (#3024, #3026)

### LINKED ISSUE #3024: member Original Source is remote-only: read local SourceLink document path before fetching the URL
## Summary

The member `Original Source` / `Source Diff` content path fetches authored source **only from the remote SourceLink URL**; it never reads the source file from the local disk even when the portable PDB's document path exists locally.

For a locally-built assembly (SourceLink enabled, reproducible/CI build off), the PDB document key is a local absolute path that maps — via the SourceLink `documents` map — to a remote raw URL. `Source Locations` already surfaces both:

```
| File | Url |
| /home/rich/.../src/ILInspector.Decompiler/Pipeline/Passes/YieldBreakLoopIteratorReconstruction.cs | https://raw.githubusercontent.com/richlander/dotnet-inspect/<sha>/src/.../YieldBreakLoopIteratorReconstruction.cs |
```

…but `-S "Original Source"` returns an empty section unless the remote URL is fetchable, because the content resolver (`ApiCommand.ResolveMethodSourceAsync` → `SourceFetcher.FetchSourceAsync(methodInfo.SourceUrl)`) has no local-file branch.

## Impact

- Inspecting a not-yet-pushed local build cannot show `Original Source` even though the exact source is on disk at the PDB document path.
- Producing before/after/original C# for decompiler PRs requires either pushing the commit first or hand-transcribing the source.

## Proposal

When the portable-PDB document path exists on disk, read and checksum-verify the local file **first**, and only fall back to the remote SourceLink URL when the local file is absent or fails checksum verification. `Source Locations` already resolves the local path, so the information is available; the content path should prefer it.

## Repro

```bash
dotnet build src/ILInspector.Decompiler -c Release -p:Deterministic=false -p:ContinuousIntegrationBuild=false -p:EnableSourceLink=true
dotnet <cli> member "<Ns.Type>" --library <built.dll> -m "<Member>:1" --all -S "Source Locations"   # shows local File + Url
dotnet <cli> member "<Ns.Type>" --library <built.dll> -m "<Member>:1" --all -S "Original Source"     # empty unless URL is fetchable
```

## Pointers

- `src/dotnet-inspect/Commands/ApiCommand.cs:555` (`ResolveMethodSourceAsync` fetches `methodInfo.SourceUrl` only).
- `src/DotnetInspector.Services/AuthoredSourceAcquisition.cs:96-113` (Absent when no fetchable URL; fetches `document.ResolvedUrl`).
- `src/dotnet-inspect/Inspectors/MemberSourceLocationCollector.cs` (already resolves the local document path).

### LINKED ISSUE #3026: member private-vs-public UX too strict: honor a directly-requested private member for Original Source (not just listings)
## Summary

The `member` command applies the same public-only visibility policy to two different intents:

1. **Listing** members of a type (a member inventory) — where hiding non-public members by default is reasonable.
2. **Directly requesting one specific member** by name/overload (e.g. `-m "Method:1"`) — where the caller has already named exactly what they want.

Today, even when a specific private member is requested directly with `--all`, the **Original Source content** path still passes `publicOnly: true`, so the member is skipped and the section renders empty. `--all` makes the private member *listable* and its `Source Locations` resolvable, but the source-content resolution remains public-only.

## Evidence

With `--all` and a direct selector `-m "TryNormalizeContinueCondition:1"`:

- `-S "Source Locations"` → resolves the private method (shows file, line range, and SourceLink URL). ✅
- `-S "Original Source"` → empty. ❌

Root cause: `ApiCommand.ResolveMethodSourceAsync(...)` is called with `publicOnly = selectedMember?.Kind != "explicit-interface-implementation"` (i.e. `true` for a normal private method), and `PdbContext.ResolveMethodSource` then skips any method whose `MemberAccessMask != Public`:

- `src/dotnet-inspect/Commands/MemberCommand.cs:285` (hardcoded `publicOnly`).
- `src/dotnet-inspect/Commands/ApiCommand.cs:555` (passes it through).
- `src/ILInspector.Metadata/PdbContext.cs:524` (`if (publicOnly && access != Public) continue;`).

## Proposal

Distinguish "list" from "direct request". When a single member is explicitly selected (name + overload/`--index`/`~digest`) and `--all` (non-public) is in effect, resolve its source regardless of accessibility — i.e. pass `publicOnly: false` on the direct-request path. Keep public-only defaults for member *inventories*.

## Impact

Directly requesting a private member's `Original Source` / `Source Diff` (common for decompiler self-inspection and burndown PRs) currently yields nothing, even though the member, its `Source Locations`, its `IL`, and its `Decompiled Source` all resolve.

richlander:
## Summary

Fixes the `member` command's **Original Source** / **Source Diff** content path, which had two gaps for a directly-requested member:

- **#3026** — a directly-requested *private* member (name + overload) with `--all` still resolved source with `publicOnly:true`, so `PdbContext.ResolveMethodSource` skipped it and the section rendered empty.
- **#3024** — source was fetched remote-only from the SourceLink URL, so a local build whose commit is not yet pushed showed empty Original Source even though the exact source is on disk at the PDB document path.

Both target the same content path and are fixed together.

## Changes

- **#3026 (`MemberCommand.cs`)** — in the single-member source-resolution branch, when a member is directly selected **and** non-public members are in scope (`--all`/`IncludeAll`), pass `publicOnly:false`. Member *inventories* keep the public-only default; explicit interface implementations stay resolvable.
- **#3024 (local-first, checksum-gated)** — prefer a checksum-verified local file before the remote URL:
  - `SourceLinkResolver.MethodSourceInfo` now carries the portable-PDB document `Checksum` + `ChecksumAlgorithm` (populated in `ResolveMethodSourceRange`, reusing `PdbContext.MapHashAlgorithm`).
  - `AuthoredSourceAcquisition` gains one shared `TryReadVerifiedLocalSource(path, alg, checksum)` helper + a public `VerifyChecksum(algorithm, expected, content)` overload. The local read is gated on **checksum match** (the PDB document hash authenticates the on-disk bytes; an attacker cannot precompute a matching hash for an unknown victim file) **plus** a compiler-source extension (`.cs`/`.vb`/`.fs`) as defense-in-depth. Any mismatch/absent checksum/missing file returns `null` → unchanged remote fallback.
  - Wired into **both** the member path (`ApiCommand.ResolveMethodSourceAsync`) and the diff path (`AuthoredSourceAcquisition.AcquireMemberAsync`).

## Security

Reading a file whose path comes from an (untrusted) PDB document name is a recognized threat (`docs/design/untrusted-data-threat-model.md`). The **checksum gate is the mitigation**: the on-disk bytes must hash to the value embedded in the portable PDB, so an attacker cannot substitute arbitrary local content. Extension gating is defense-in-depth. No behavior change to remote fetch.

## Evidence

- **Unit tests** (`VerifiedLocalSourceReadTests`, 8, all pass): checksum match, mismatch rejection, non-source-extension rejection, absent-checksum rejection, absent-file → `null`, CRLF-normalized match, observation overload, `VerifyChecksum` overload.
- **Dogfood (proves both fixes in one run):** built `ILInspector.Decompiler` with SourceLink at an **unpushed** commit, then ran `member` on a **private** method (`WithExpressionPass.IsRecordCloneCall:1 --all -S "Original Source"`). Original Source **rendered** (was empty before). Its SourceLink URL pointed at the unpushed SHA and returns **404** remotely — so the content came solely from the checksum-verified local file. This exercises both #3026 (private direct request) and #3024 (local-first).
- **Sanity:** public method `--all` still renders; a private method **without** `--all` still errors "non-public; pass --all to include it" (inventory default preserved).
- **Suites:** `DotnetInspector.Services.Tests` 276/276 pass. All failures in `dotnet-inspect.Tests` (5) and `Metadata.Tests` (1) were confirmed **pre-existing at `origin/main`** (network-data drift, `\r`/CRLF artifacts, a section-naming baseline, and a Windows path-separator assertion) — this PR introduces **zero** regressions.

## Compatibility

Behavior-safe: remote fetch is unchanged and remains the fallback; local read is additive and strictly checksum+extension gated. Member inventories keep the public-only default.

Adversarial review (two models) to follow — behavior change + new local-file read + security-sensitive.

Fixes #3026
Fixes #3024

### DIFF
--- a/src/DotnetInspector.Services.Tests/VerifiedLocalSourceReadTests.cs
+++ b/src/DotnetInspector.Services.Tests/VerifiedLocalSourceReadTests.cs
@@ -178,4 +178,58 @@ public class VerifiedLocalSourceReadTests
             SourceChecksumVerification.Unavailable,
             AuthoredSourceAcquisition.VerifyChecksum(null, SHA256.HashData(content), content));
     }
+
+    [Theory]
+    [InlineData(@"\\attacker\share\Sample.cs")]
+    [InlineData(@"\\?\C:\Sample.cs")]
+    [InlineData(@"\\.\C:\Sample.cs")]
+    [InlineData("relative/Sample.cs")]
+    public void ReturnsNull_ForNonLocalOrRelativePath_WithoutIo(string path)
+    {
+        // The document path originates in an untrusted PDB. A UNC share or Win32 device path must
+        // be rejected before any filesystem access so it cannot trigger outbound SMB/network I/O,
+        // and a relative path is never honored — independent of the (here matching) checksum.
+        byte[] content = Encoding.UTF8.GetBytes(Source);
+        Assert.Null(AuthoredSourceAcquisition.TryReadVerifiedLocalSource(
+            path, "SHA256", SHA256.HashData(content)));
+    }
+
+    [Fact]
+    public void DecodeSourceText_HonorsByteOrderMarkEncoding()
+    {
+        // Non-ASCII content makes the encoding observable. A checksum-verified local file may be
+        // UTF-8 (with or without BOM) or UTF-16; each must decode correctly, not as raw UTF-8.
+        const string text = "héllo wörld";
+        Assert.Equal(text, AuthoredSourceAcquisition.DecodeSourceText(Encoding.UTF8.GetBytes(text)));
+        Assert.Equal(text, AuthoredSourceAcquisition.DecodeSourceText(
+            [.. Encoding.UTF8.GetPreamble(), .. Encoding.UTF8.GetBytes(text)]));
+        Assert.Equal(text, AuthoredSourceAcquisition.DecodeSourceText(
+            [.. Encoding.Unicode.GetPreamble(), .. Encoding.Unicode.GetBytes(text)]));
+        Assert.Equal(text, AuthoredSourceAcquisition.DecodeSourceText(
+            [.. Encoding.BigEndianUnicode.GetPreamble(), .. Encoding.BigEndianUnicode.GetBytes(text)]));
+    }
+
+    [Fact]
+    public void Utf16LocalFile_VerifiesAgainstRawBytes_AndDecodesToText()
+    {
+        // csc records the checksum over the raw file bytes regardless of encoding, so a UTF-16
+        // source verifies against those bytes and must then decode as UTF-16 (Finding 2 regression).
+        byte[] utf16 = [.. Encoding.Unicode.GetPreamble(), .. Encoding.Unicode.GetBytes(Source)];
+        string path = WriteTemp(".cs", utf16);
+        try
+        {
+            byte[]? result = AuthoredSourceAcquisition.TryReadVerifiedLocalSource(
+                path, "SHA256", SHA256.HashData(utf16));
+
+            Assert.NotNull(result);
+            Assert.Equal(utf16, result);
+            Assert.Equal(
+                Source.ReplaceLineEndings("\n"),
+                AuthoredSourceAcquisition.DecodeSourceText(result).ReplaceLineEndings("\n"));
+        }
+        finally
+        {
+            File.Delete(path);
+        }
+    }
 }
--- a/src/DotnetInspector.Services.Tests/VerifiedLocalSourceReadTests.cs
+++ b/src/DotnetInspector.Services.Tests/VerifiedLocalSourceReadTests.cs
@@ -1,0 +1,181 @@
+using System.Security.Cryptography;
+using System.Text;
+
+using ILInspector.Metadata;
+
+namespace DotnetInspector.Services.Tests;
+
+public class VerifiedLocalSourceReadTests
+{
+    const string Source = """
+        class Sample
+        {
+            public int M()
+            {
+                return 1;
+            }
+        }
+        """;
+
+    static string WriteTemp(string extension, byte[] content)
+    {
+        string path = Path.Combine(
+            Path.GetTempPath(),
+            $"dotnet-inspect-src-{Guid.NewGuid():N}{extension}");
+        File.WriteAllBytes(path, content);
+        return path;
+    }
+
+    [Fact]
+    public void ReturnsBytes_WhenChecksumMatches()
+    {
+        byte[] content = Encoding.UTF8.GetBytes(Source);
+        string path = WriteTemp(".cs", content);
+        try
+        {
+            byte[]? result = AuthoredSourceAcquisition.TryReadVerifiedLocalSource(
+                path, "SHA256", SHA256.HashData(content));
+
+            Assert.NotNull(result);
+            Assert.Equal(content, result);
+        }
+        finally
+        {
+            File.Delete(path);
+        }
+    }
+
+    [Fact]
+    public void ReturnsNull_WhenChecksumMismatches()
+    {
+        byte[] content = Encoding.UTF8.GetBytes(Source);
+        string path = WriteTemp(".cs", content);
+        try
+        {
+            // The on-disk bytes do not match the recorded hash: the read must be refused so the
+            // caller falls back to the remote URL rather than surfacing unverified content.
+            byte[]? result = AuthoredSourceAcquisition.TryReadVerifiedLocalSource(
+                path, "SHA256", SHA256.HashData(Encoding.UTF8.GetBytes(Source + "tampered")));
+
+            Assert.Null(result);
+        }
+        finally
+        {
+            File.Delete(path);
+        }
+    }
+
+    [Fact]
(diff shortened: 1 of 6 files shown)


richlander:
## Adversarial review — round 1 (base head `61da1952`)

Two-model review per AGENTS.md (behavior change + new local-file read + security-sensitive). Reviewers: **GPT-5.6** and **Gemini 3.1 Pro**, each isolated in its own read-only worktree at the exact head.

**GPT-5.6 — two findings (both actioned):**

1. **Blocking** — an untrusted PDB document path such as `\\attacker\share\x.cs` passed the `.cs/.vb/.fs` extension gate and reached `File.Exists`/`File.ReadAllBytes` **before** the checksum was evaluated, so on Windows it could trigger outbound SMB/NTLM I/O even though the content ultimately fails verification. Content authentication does not prevent the I/O side effect. **Fixed** in `1ce09bab`: new `IsLocalFileSystemPath` gate requires a fully-qualified local path and rejects UNC / `\\?\` / `\\.\` / `//` paths **before any filesystem access**; the read also skips reparse points and files above a 64 MB cap.
2. **Major** — a UTF-16 source file checksum-matches on its raw bytes but was decoded UTF-8-only, silently corrupting Original Source with no remote fallback (same bug in the diff path `FromContent`). **Fixed** in `1ce09bab`: shared BOM-aware `DecodeSourceText` (UTF-8/UTF-16/UTF-32, default UTF-8) now decodes both the member path and the diff path, mirroring the remote `HttpContent.ReadAsStringAsync` decoding.

**Gemini 3.1 Pro — no blocking/major findings.** It verified checksum sufficiency, algorithm mapping, the no-usable-checksum paths, TOCTOU/ADS harmlessness, line-ending normalization, and the `publicOnly` relaxation scope. Its path-handling analysis covered *content* authentication but not the *I/O side effect* that GPT flagged — the gap GPT's Finding 1 closes.

**Resolution commit:** `1ce09bab` — adds 6 tests (UNC/device/relative rejection + BOM-aware decode); full `DotnetInspector.Services.Tests` suite 282 pass. Fixed-head dogfood: private `--all` member Original Source still renders while its SourceLink URL returns 404, proving content came solely from the checksum-verified local file.

Both reviewers are now re-reviewing the fixed head `1ce09bab`. I will not mark this ready until both fixed-head reviews are clean.

richlander:
## Adversarial review — round 2 (fixed heads `1ce09bab`, then `3bc33620`)

Both reviewers re-ran on the fixed head in their isolated read-only worktrees.

**Fixed head `1ce09bab`:**
- **Gemini 3.1 Pro — clean.** Independently probed `IsLocalFileSystemPath` (UNC/device/drive-relative rejected, normal absolute accepted), confirmed the reparse check is leaf-only (won't break junctioned repos), verified all `FileInfo`/attribute exceptions are caught, and checked `DecodeSourceText` on empty/lone/partial BOM. Build + 14 tests pass.
- **GPT-5.6** confirmed Finding 2 (UTF-16) fully resolved and all UNC/device/relative paths rejected, but raised a new **major**: a network-mapped drive letter (`Z:` → `\\server\share`) is fully-qualified and survives `GetFullPath`, so `FileInfo.Exists` would still reach the network.

**Resolution `3bc33620`:** `IsLocalFileSystemPath` now also rejects a path whose root `DriveType == Network` (`GetDriveType` is a local lookup — no share connection), with a try/catch that falls through on indeterminate drive type. The reviewer's companion point (a file beneath an ancestor junction targeting UNC) is intentionally left to the leaf-only reparse check: exploiting it requires an attacker to have pre-planted a junction (local write access, outside the untrusted-PDB threat model), and rejecting ancestor junctions would regress legitimate repos mounted under a directory junction.

**Fixed head `3bc33620` — both clean:**
- **GPT-5.6 — clean.** Confirmed the mapped-network-drive vector is closed, `DriveType` does no network I/O, local/subst drives still accepted, and **explicitly accepted the ancestor-junction threat-model boundary** ("the PDB cannot create or select its target").
- **Gemini 3.1 Pro — clean.** Re-confirmed the additive change: local fixed-drive paths still accepted, `DriveType` is a safe local lookup, try/catch fall-through sound. Build + 14 tests pass.

Two clean fixed-head reviews on `3bc33620`. CI `test` job pending; will confirm readiness once green.

richlander:
Ready to merge

richlander:
Follow-up commit 8aae6df9 is comment-only: it reframes the local-source comments around build reproducibility (a non-reproducible/local build records a real local PDB document path whose exact bytes may exist only on this machine; reproducible builds normalize paths and publish sources, so remote SourceLink stays authoritative there). No logic change, so the two clean fixed-head reviews on 3bc33620 still hold (documentation/mechanical change per AGENTS.md). CI green on 8aae6df9.

Ready to merge

PR #3032, opened 2026-07-22
Drop redundant (dynamic) cast on dynamic-typed receivers (#2984)

richlander:
- Advances #2984 (umbrella #2996)
- Drops the redundant `((dynamic)x)` cast on a raised dynamic member access when the receiver's static type is already `dynamic`
- Card revision: `a5ee9f18`

## Change

**Conclusion:** **PASS** — a dynamic member access carries no member signature (the name is a string handed to the runtime binder), so `receiver.Member` binds directly when the receiver is already `dynamic`; the cast is only needed to coerce an `object`-typed operand, and the change drops it exactly when metadata proves the receiver dynamic and keeps it everywhere else.

### Original source

SourceLink cannot supply C# for the local fixture assembly, so the authoritative anchor is the raw IL (the C# author wrote `public object DynamicGetLength(dynamic value) => ((dynamic)value).Length;`). The IL is a call-site dynamic `GetMember("Length")` on `ldarg.1`:

```il
IL_0008: ldstr        "Length"
IL_0027: call         Microsoft.CSharp.RuntimeBinder.Binder::GetMember(...)
IL_002C: call         System.Runtime.CompilerServices.CallSite<System.Func<CallSite, object, object>>::Create(...)
IL_0040: ldsfld       ...<>o__1::<>p__0
IL_0045: ldarg.1
IL_0046: callvirt     System.Func<CallSite, object, object>::Invoke(CallSite, object)
```

### Before

```csharp
public object DynamicGetLength(dynamic value) => ((dynamic)value).Length;
```

- Valid: True
- Correct: True

Valid and correct, but the `(dynamic)` cast on a receiver that is already `dynamic` is redundant noise.

### After

```csharp
public object DynamicGetLength(dynamic value) => value.Length;
```

- Valid: True
- Correct: True

### Fully raised

## How

Recover the dynamic type view from metadata and let the printer decide:

- Decode `[DynamicAttribute]` at import time via the shared `DynamicReader.IsTopLevelDynamic` primitive (index 0 of the transform-flags preorder walk is the whole type; a nested `Func<dynamic>` leaves it false, so only a bare `dynamic` sets it).
- Thread the fact through three receiver paths: a dynamic parameter (`Parameter`/`LoadArgument.IsDynamic`), a hoisted display-class field (`FieldRef.IsDynamic` via `FieldDefinition`), and a generic-enclosing display-class field accessed through a `MemberReference` (`MemberReferenceFieldIsDynamic` resolves the underlying same-assembly definition).
- `CSharpPrinter.DynamicGetMemberText` drops the cast only for a receiver with a recovered dynamic view; every other receiver keeps it.

Conservative by construction: csc omits `[DynamicAttribute]` on iterator/async state-machine hoisted fields, and spilled-temp receivers are plain locals, so both keep the cast — no false drops. `FieldRef.IsDynamic` never affects the cache-field (`<>o__N`) identity checks in `DynamicCallSitePass` (cache fields are never dynamic).

## Evidence

Controlled A/B, identical `-notrait "Speed=Slow"` Decompiler suite, base `20f16b9a` (this commit's base) vs head `a5ee9f18`:

| Check | Baseline (`20f16b9a`) | Head (`a5ee9f18`) |
| --- | --- | --- |
| Product build (`src/dotnet-inspect`) | Pass | Pass |
| `DynamicCallSitePassTests` | 110 passed | 110 passed |
| `LadderRung9GateTests` | 19 passed | 20 passed (+1 new gate) |
| `DynamicTypeViewTests` (Metadata) | 50 passed | 50 passed |
| Decompiler fast suite | 55 failed (+1 unbuilt-fixture artifact) | 55 failed |
| Real witness `DynamicGetLength(dynamic)` | `((dynamic)value).Length` | `value.Length` |
| Real witness `DynamicGetLengthOfObject(object)` | n/a (new fixture) | `((dynamic)value).Length` (cast kept) |

Failing-test **set** diff: **zero head-only failures** (no regressions). The 55 head failures are all pre-existing Windows-local #3018 env issues (CS0009 native-DLL recompile / CRLF / path-sep) and fail identically at base. The single base-only failure (`ExpressionTreeLambdaTests.LookalikeExpressionAssembly_StaysFactoryCalls`) is a `FileNotFoundException` for an unbuilt fixture in the partial base worktree, not a product difference.

New tests: negative fixture `DynamicGetLengthOfObject(object) => ((dynamic)value).Length;` + gate `Rung9Fixture_KeepsCastForObjectReceiver` (object receiver keeps cast); `IsTopLevelDynamic` Theory (null/[0]/[1]/[0,1]/[1,0]); iterator/generic-iterator cases assert the cast is still kept.

## Validation

```powershell
dotnet build src\dotnet-inspect\dotnet-inspect.csproj -c Release --configfile .\nuget.config
dotnet run --project src\ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.DynamicCallSitePassTests"
dotnet run --project src\ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.LadderRung9GateTests"
dotnet run --project tests\ILInspector.Metadata.Tests -c Release -- -class "ILInspector.Metadata.Tests.DynamicTypeViewTests"
dotnet run --project src\ILInspector.Decompiler.Tests -c Release -- -notrait "Speed=Slow"
```

## Post-review update (head `6173567b`)

Adversarial review surfaced one real regression, now fixed (details in the reconciliation comment below):

- **Local-function own dynamic param** (GPT, High): a local function with its own `dynamic` parameter dropped the body cast but the printer-owned local-function declaration spelled the param from its `TypeRef` as `object`, yielding `static object Get(object v) => v.Length;` (CS1061). Fixed in `6173567b`: `CSharpPrinter` routes local-function param types through `ParameterTypeText` (spells `dynamic` when the param is dynamic) → `static object Get(dynamic v) => v.Length;` (valid, opcode-exact). New fixture `DynamicMemberContexts.InLocalFunctionOwnParam` + regression test `NestedContext_LocalFunctionOwnParam_SpellsDynamicDeclaration` (`DynamicCallSitePassTests` now 111/111).
- **`ref/in/out dynamic` receivers** (Gemini, Medium): the cast is conservatively **kept** (valid + correct); the `IsTopLevelDynamic` ByRef-slot offset has zero output effect because by-ref receivers render as deref nodes the printer never drops. Non-blocking; filed as follow-up **#3035**.

Post-fix A/B (head `6173567b` vs `a5ee9f18`): identical failing-test set — zero new failures, zero resolved.

### DIFF
--- a/src/ILInspector.Decompiler.Fixtures.LadderRung9/LadderRung9.cs
+++ b/src/ILInspector.Decompiler.Fixtures.LadderRung9/LadderRung9.cs
@@ -20,6 +20,15 @@ public class DynamicAndExpressionTrees
         return value.Length;
     }
 
+    // Close negative for #2984: the receiver is a genuine `object` parameter (no
+    // [DynamicAttribute]); the member access only becomes dynamic through the
+    // explicit `(dynamic)` cast in source. The redundant-cast drop must NOT fire
+    // here — the static type really is object — so `((dynamic)value)` is kept.
+    public object DynamicGetLengthOfObject(object value)
+    {
+        return ((dynamic)value).Length;
+    }
+
     public object DynamicInvoke(dynamic function, int value)
     {
         return function(value);
--- a/src/ILInspector.Decompiler.Fixtures.LadderRung9/LadderRung9MemberContexts.cs
+++ b/src/ILInspector.Decompiler.Fixtures.LadderRung9/LadderRung9MemberContexts.cs
@@ -48,6 +48,18 @@ public class DynamicMemberContexts
         return Get();
     }
 
+    // Nested local-function body whose local function has its OWN top-level
+    // `dynamic` parameter (distinct from InLocalFunction, which captures the
+    // enclosing dynamic param). The raised body drops the redundant cast to
+    // `v.Length`, so the printer-owned local-function declaration must spell the
+    // parameter `dynamic v`, not its `object` TypeRef — otherwise
+    // `object Get(object v) => v.Length;` is CS1061 (#2984, PR #3032 review).
+    public object InLocalFunctionOwnParam(string input)
+    {
+        static object Get(dynamic v) => v.Length;
+        return Get(input);
+    }
+
     // Iterator state-machine body: csc lowers this to a MoveNext method whose
     // declaring type is the generated iterator state machine, while the
     // GetMember context remains the authored enclosing type (same bridge as
--- a/src/ILInspector.Decompiler.Tests/DynamicCallSitePassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/DynamicCallSitePassTests.cs
@@ -219,7 +219,11 @@ public class DynamicCallSitePassTests
     {
         var f = LoadCanonicalFunction();
         var output = RaiseAndPrint(f);
-        Assert.Contains("((dynamic)value).Length", output);
+        // `value` is a top-level `dynamic` parameter, so the receiver's static
+        // type is already dynamic and the redundant `((dynamic)value)` cast is
+        // dropped (issue #2984).
+        Assert.Contains("value.Length", output);
+        Assert.DoesNotContain("((dynamic)value)", output);
     }
 
     [Fact]
@@ -238,7 +242,8 @@ public class DynamicCallSitePassTests
         var f = LoadCanonicalFunction();
         BinderCall(f).Arguments[1].ReplaceWith(new Constant("class", StringType));
         var output = RaiseAndPrint(f);
-        Assert.Contains("((dynamic)value).@class", output);
+        Assert.Contains("value.@class", output);
+        Assert.DoesNotContain("((dynamic)value)", output);
     }
 
     [Fact]
@@ -1416,9 +1421,11 @@ public class DynamicCallSitePassTests
     public void ImmediateUse_FieldAssignment_Raises()
     {
         // Compiler-backed: `_last = value.Length;` — the dynamic access is the
-        // value of a field store, not a return.
+        // value of a field store, not a return. `value` is a top-level dynamic
+        // parameter, so the redundant cast is dropped (#2984).
         var output = RaiseMemberContext("AssignToField");
-        Assert.Contains("_last = ((dynamic)value).Length;", output);
+        Assert.Contains("_last = value.Length;", output);
+        Assert.DoesNotContain("((dynamic)value)", output);
         Assert.DoesNotContain("Binder.GetMember", output);
     }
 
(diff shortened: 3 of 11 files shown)


richlander:

Two-model adversarial review per AGENTS.md — **GPT** (`gpt-5.6-sol`) and **Gemini** (`gemini-3.1-pro-preview`), each in an isolated review worktree. Round 1 reviewed `a5ee9f18`; both re-reviewed the fixed head `6173567b`.

### Findings

**GPT — High, verified → FIXED.** Dropping the cast inside a **local function** with its **own `dynamic` parameter** produced invalid C#: the printer-owned local-function declaration spelled the param from its `TypeRef` as `object`, so the body `v.Length` no longer bound (`static object Get(object v) => v.Length;` = CS1061). The pre-drop `((dynamic)v).Length` had compiled. Reproduced with a compiled probe. **Fixed in `6173567b`**: `CSharpPrinter` routes local-function param types through `ParameterTypeText` (spells `dynamic` when the param is dynamic) → `static object Get(dynamic v) => v.Length;`, valid and opcode-exact on recompile. Regression fixture `DynamicMemberContexts.InLocalFunctionOwnParam` + test `NestedContext_LocalFunctionOwnParam_SpellsDynamicDeclaration` (`DynamicCallSitePassTests` 110→111).

**Gemini — Medium, verified → NON-BLOCKING follow-up (#3035).** For `ref/in/out dynamic` params the `ByRef` modifier consumes the leading transform-flag slot, so `IsTopLevelDynamic` (checks `flags[0]==1`) returns false and the param is not marked dynamic. Verified with a probe that the cast is **conservatively kept** (`((dynamic)(value)).Length`) — valid and behavior-exact. Critically this has **zero output effect** in this PR: a by-ref receiver renders as a **deref** node, not a bare `LoadArgument`/`LoadField`, and `IsDynamicTypedReceiver` keys only on those, so the printer would never drop the cast even with the offset fixed. A missed drop, not a wrong drop. Both reviewers agreed the disposition is non-blocking; filed as **#3035** (decode ByRef offset + printer deref-receiver recognition).

### Fixed-head re-review (`6173567b`) — both CLEAN

- **GPT:** original defect resolved (probe emits `dynamic v` and recompiles); product + fixture + test builds succeed; `DynamicCallSitePassTests` 111/111, `LadderRung9GateTests` 20/20; no new defects.
- **Gemini:** fix is sound (nested `Func<dynamic>` params correctly fall back to `TypeText` and keep the redundant-but-valid cast; non-dynamic params unaffected); re-confirmed `FieldRef` equality vs cache-field checks, `MemberReferenceFieldIsDynamic` same-assembly resolution, and printer rvalue-only drop; 131 Decompiler tests pass; agreed the ref/in/out disposition.

Post-fix A/B (`6173567b` vs `a5ee9f18`, identical `-notrait "Speed=Slow"` Decompiler suite): identical failing-test set — zero new failures, zero resolved.

richlander:
Ready to merge

- Both fixed-head (`6173567b`) adversarial reviews clean: GPT (`gpt-5.6-sol`) and Gemini (`gemini-3.1-pro-preview`). Reconciliation above.
- CI green: `changes`, `build-net10`, `csharp-diff-smoke`, and `test (ubuntu-26.04, linux-x64)` all pass; `il-diff-smoke`/`markdownlint`/`pack` path-skipped.
- One non-blocking follow-up filed: #3035 (`ref/in/out dynamic` missed cast drop — conservative-correct today).

PR #3037, opened 2026-07-22
Suppress stray auto-property backing field on RTS Full member surface (#3036)

### LINKED ISSUE #3036: RTS: stray compiler backing field emitted alongside preserved auto-property skeleton under Full
## Summary

Follow-up from #3008 (issue #3000). Under the `Full` body policy, when a type is pulled onto the RTS reconstruction member surface and one of its members is a compiler-synthesized auto-property, the reconstruction now correctly preserves the auto-property skeleton (`public int Value { get; init; }` / `{ get; set; }`) — see the recursion fix in #3008. However, the compiler-generated backing field is *also* emitted as a separate stray field:

```csharp
public struct Holder
{
    public int M() { return 1; }
    public int __Value_k__BackingField;   // stray: duplicates the compiler-synthesized backing field
    public int Value { get; init; }
}
```

## Why this is low severity (not blocking #3008)

- It is **pre-existing** and orthogonal to the recursive-accessor bug fixed in #3008 (the stray field was present in the recursive output too).
- The reconstructed source **compiles** — a public field named `__Value_k__BackingField` next to an auto-property is legal C#.
- It does **not** affect the target method's IL comparison (the target does not read the field), so `Status` stays `Exact` and it is not a silent-Complete-of-broken-body integrity violation.

## Suggested direction

When an auto-property skeleton is preserved on the member surface, suppress the corresponding compiler-generated `<Name>k__BackingField` field member (the compiler re-synthesizes its own backing field for the auto-property), so the reconstructed type does not carry a duplicate field.

## Provenance

Observed and agreed as out-of-scope follow-up during the two-model adversarial review of #3008 (GPT-5.6 raised it; both GPT-5.6 and Gemini 3.1 Pro agreed it is orthogonal/non-blocking).

richlander:
## Summary

Fixes #3036. Under the `Full` body policy, when a type is pulled onto the RTS reconstruction member surface and one of its members is a compiler-synthesized auto-property, the reconstruction correctly preserves the auto-property skeleton (per #3008) but *also* emitted the compiler-generated `<Value>k__BackingField` as a separate field. Its unspeakable name was sanitized to a legal identifier (`__Value_k__BackingField`), producing a stray duplicate field next to the reconstructed auto-property:

```csharp
public struct Holder
{
    public int M() { return 1; }
    public int __Value_k__BackingField;   // stray duplicate
    public int Value { get; init; }
}
```

## Fix

`AddClosureMemberSurface`'s field loop now collects the backing-field names (`<Name>k__BackingField`) of every auto-property on the type — via the existing `IsAutoProperty` / `IsAutoPropertySetter` checks — and skips them. The compiler re-synthesizes its own backing field for each emitted auto-property skeleton, so the raw field is redundant.

## Why this is safe

- Backing-field reads/writes by the target body are already expressed through the property (`this.Value`, and the `target-backing-field-read`/`write` requirement path adds a `Value` member), never through the mangled field identifier, so no evidence is lost.
- Fields whose corresponding property is filtered out (unsupported/pointer signature) are independently skipped by the field loop's own guards, so suppression never drops a needed field.
- Suppression is keyed to auto-properties only; explicit-body properties (whose accessors touch a private `_value`) are untouched.

## Evidence

- New regression test `CompileBackTargets_FullSuppressesStrayAutoPropertyBackingField`: asserts no `k__BackingField` in the reconstructed source, no `k__BackingField` member in the plan, `Status == Exact`, no compile-back floor.
- Strengthened the two existing `Holder` auto-property sibling tests with `Assert.DoesNotContain("k__BackingField", ...)`.
- Reproduced the stray field before the fix (`public int __Value_k__BackingField;`), confirmed gone after.
- RoundTrip area: **312/312 pass**. Corpus area: **61/61 pass**.

## Adversarial review

Two-model fixed-head review to follow.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -540,6 +540,7 @@ public class ReturnToSenderPrototypeTests
                 RoundTripBodyPolicy.Full));
 
             Assert.Contains("public int Value { get; init; }", result.Source);
+            Assert.DoesNotContain("k__BackingField", result.Source);
             Assert.DoesNotContain("return this.Value", result.Source);
             Assert.DoesNotContain("this.Value = value", result.Source);
             Assert.False(result.UsedCompileBackFloor, result.Detail);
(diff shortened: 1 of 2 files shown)


richlander:

Two-model fixed-head review at exact head `8efbd616` (excluding the authoring Claude Opus family), each isolated in its own read-only review worktree. **Both reviews are clean — no findings.**

**GPT-5.6** — no significant issues. Verified backing-field references render as property accesses, exact-name (raw metadata name) suppression avoids sanitized-name collisions, evidence `Status`/`BodyComplete` remain truthful, and all 149 RTS prototype tests pass at `8efbd616`.

**Gemini 3.1 Pro** — no significant issues. Verified:
1. **Target-reference safety** — when the target body directly references the backing field, the required-member path adds it as an explicit `CompileBackMemberRequirement` *before* `AddClosureMemberSurface`; the new `continue` only skips adding a *stray* duplicate, never strips a required member, so binding is preserved and the source still compiles.
2. **Predicate soundness** — compiled/ran `Holder<T>`, `dynamic`, and `ValueTuple<int,int>` property fixtures; `AutoPropertyBackingFieldNames` resolves types via `GenericContext.ForType` and evaluates `DisplayName` without throwing or failing the type-equivalence check.
3. **Explicit interface impls** — `<IProp.Value>k__BackingField` bypasses suppression via the pre-existing `fieldName.Contains('.')` guard.
4. **Name collisions** — exact ordinal match on the raw metadata name (`<Value>k__BackingField`) cannot collide with the sanitized C# identifier (`__Value_k__BackingField`).

No blocking findings from either reviewer; nothing to action.

Ready to merge

PR #3039, opened 2026-07-22
Fold negated total-order relational operator-spelled calls (#2955)

### LINKED ISSUE #2955: EVIL frontier: fold negated comparisons !(a==b) -> a!=b (branch-polarity idiom recovery)

richlander:
## What

Continues the negated-comparison idiom recovery from #2961 into the relational De Morgan pairs that issue #2955 explicitly deferred:

- `!(a < b)` → `a >= b`
- `!(a <= b)` → `a > b`
- `!(a > b)` → `a <= b`
- `!(a >= b)` → `a < b`

## Why

The branch-polarity raiser renders a negated relational operator-spelled call as `!(a < b)` instead of the idiomatic `a >= b`, which Roslyn then re-lowers with an inverted `brtrue`/`brfalse` and a relocated block (the `valid_different.semantic_opcode_diff.syntax` frontier in #2955). Folding to the dual comparison reads naturally and aligns the recompiled branch polarity with the source — the same mechanism #2961 fixed for `op_Equality`/`op_Inequality`.

## Soundness (the critical property)

The fold is gated on a **narrow, purpose-built allowlist** `MemberIdentity.IsTotalOrderRelationalOperator`, restricted to core-library value types whose natural order is **total** — `decimal`, `DateTime`, `DateTimeOffset`, `TimeSpan`, `DateOnly`, `TimeOnly` — where the four relational operators are exact De Morgan duals for **every** input, so `!(a < b)` is always `a >= b`.

Deliberately **excluded**:

- **`System.Half`** — uses operator CALLS (not native `clt`/`cgt`) **and** is an IEEE-754 partial order: `NaN` is unordered, so `!(a < b)` can differ from `a >= b`. This is the key negative case.
- **`float`/`double`** — never reach this path (native `clt`/`cgt`, folded with the unordered-flag guard by the comparison path).
- **every user-defined operator type** — a user `<`/`>=` pair need not be exact duals; folding could substitute one method call for a different one (the scope guard #2961's adversarial review established for the equality pair).

The declaring type is assembly-gated through `IsCoreLibraryType`, so a user type masquerading as `System.Decimal` in its own assembly is not matched. (Adversarial review surfaced that this corelib gate is name-forgeable; that is a pre-existing, systemic property shared with the merged #2961 fold, tracked separately in #3045 — see the reconciliation comment below.)

### Real-world before/after (NodaTime)

`NodaTime.TimeZones.BclDateTimeZone.BclAdjustmentRule::IsStandardOffsetOnlyRule` (package `NodaTime` 3.1.11, `net6.0`). Authored source ([`BclDateTimeZone.cs` L434-L442](https://github.com/nodatime/nodatime/blob/3.1.11/src/NodaTime/TimeZones/BclDateTimeZone.cs#L434-L442)):

```csharp
return daylight.IsFixedDateRule && daylight.Day == 1 && daylight.Month == 1 &&
       daylight.TimeOfDay.TimeOfDay < TimeSpan.FromMinutes(1) &&
       standard.IsFixedDateRule && standard.Day == 1 && standard.Month == 1 &&
       standard.TimeOfDay.TimeOfDay < TimeSpan.FromMinutes(1);
```

**How to read the decompiled output.** This single `return` is an eight-operand `&&` chain. A short-circuit `&&` chain compiles to early-exit guards — *bail out (`return false`) when an operand is **not** satisfied*, then return the final operand directly:

```text
if (!A) return false;   // operand 1
...
if (!(a < b)) return false;   // operand 4 — the "!" is added by the && lowering, NOT in the source
...
return (a < b);   // operand 8 — final operand returned directly, no "!"
```

So the eight `&&` operands map one-for-one, top to bottom, onto the eight statements below. The locals are named `daylight`/`standard` after the source; note that `TimeZoneInfo.TransitionTime.TimeOfDay` is itself a `DateTime` (spilled to `V_2`), whose `.TimeOfDay` is the `TimeSpan`-of-day — hence the doubled `.TimeOfDay`.

**Why operand 4 shows `>=` when the source says `<`.** The `>=` is not a different comparison — it is the *negation* of the source `<`, and that negation is the early-exit test introduced by the `&&` lowering above. Operand 4's guard is `if (!(a < b)) return false;`; this PR folds `!(a < b)` to its exact De Morgan dual for a total-order type, `a >= b`, giving `if (a >= b) return false;` — the **same condition**, idiomatically spelled. The **eighth** operand keeps `<` because the final `&&` operand becomes `return … < …;` directly, with no early-exit negation to fold — which is exactly why only operand 4 changes. Full decompiled method, base vs this PR (product command `member "NodaTime.TimeZones.BclDateTimeZone+BclAdjustmentRule" -m IsStandardOffsetOnlyRule --library NodaTime.dll --all -S "Decompiled Source"`):

```diff
 private static bool IsStandardOffsetOnlyRule(System.TimeZoneInfo.AdjustmentRule rule)
 {
     TimeZoneInfo.TransitionTime daylight = rule.DaylightTransitionStart;
     TimeZoneInfo.TransitionTime standard = rule.DaylightTransitionEnd;
     if (!daylight.IsFixedDateRule)      // && operand 1
     {
         return false;
     }
     if (daylight.Day != 1)              // && operand 2
     {
         return false;
     }
     if (daylight.Month != 1)            // && operand 3
     {
         return false;
     }
     DateTime V_2 = daylight.TimeOfDay;
-    if (!(V_2.TimeOfDay < TimeSpan.FromMinutes(1d)))   // && operand 4 — base (pre-fold)
+    if (V_2.TimeOfDay >= TimeSpan.FromMinutes(1d))     // && operand 4 — this PR
     {
         return false;
     }
     if (!standard.IsFixedDateRule)      // && operand 5
     {
         return false;
     }
     if (standard.Day != 1)              // && operand 6
     {
         return false;
     }
     if (standard.Month != 1)            // && operand 7
     {
         return false;
     }
     V_2 = standard.TimeOfDay;
     return V_2.TimeOfDay < TimeSpan.FromMinutes(1d);   // && operand 8 — not negated, unchanged
 }
```

(The `// && operand N` comments are annotations for this write-up; the tool output is otherwise verbatim.) The fold changes exactly one line — operand 4 — from `!(V_2.TimeOfDay < …)` to `V_2.TimeOfDay >= …`; operand 8's non-negated `<` is untouched, confirming the fold targets only the negated operand. This is the direct relational analogue of #2961's `!(type == typeof(string)) && (…)` short-circuit shape.

Both are valid and semantically identical. On the isolated two-condition shape (`a < FromMinutes(1) && b < FromMinutes(1)`), an opcode dump confirms the branch direction: the original source `<` compiles to `op_LessThan; brfalse`, the base `!(a < …)` recompiles to `op_LessThan; brtrue` (branch flipped), and the folded `a >= …` recompiles to `op_GreaterThanOrEqual; brfalse` — recovering the source branch polarity.

### Fixture before/after (type coverage + `Half` exclusion)

Product `member … -S "Decompiled Source"` on a compiled `net11.0` fixture, exercising the value position, the condition position, and the `Half` negative case:

| Source | base (pre-fold) | head (this PR) |
| --- | --- | --- |
| `!(a < b)` (DateTime) | `!(a < b)` | `a >= b` |
| `if (!(a >= b))` (DateTime) | `if (!(a >= b))` | `if (a < b)` |
| `!(a < b)` (**Half**) | `!(a < b)` | `!(a < b)` (unchanged) |

### Total-order duality (empirical)

Reflection probe over the six allowlist types + `Half`: all six live in `System.Private.CoreLib`, declare all four relational operators, and satisfy `!(a<b) == a>=b` for every sample (including the equality boundary). `Half`'s dual **breaks** on `NaN` (`!(a<b)=true`, `a>=b=false`), confirming the exclusion.

### Unit tests

`IrImporterTests` (xUnit v3, run locally green): the four folds (Theory), `decimal`, the condition position, the `Half` partial-order exclusion, and the user-defined-type scope guard. Renames the prior `NotOverRelationalOperatorCall_Parenthesizes` → `_OutsideTotalOrderAllowlist_Parenthesizes` (`System.Type` carries no relational operators, stays un-folded and parenthesized).

## Fidelity tradeoff (same family as #2961)

For a genuine literal `!(a < b)` source (origin `op_LessThan; ceq`), the fold renders `a >= b` (recompiles to `op_GreaterThanOrEqual`), i.e. valid-but-different for that origin; for the mis-raised short-circuit shape it recovers the source `brfalse` branch polarity. Output is always valid and semantically identical. Full-corpus fidelity/no-regression is validated by the CI fidelity gates (local compile-back is blocked by a pre-existing Windows env issue).

## Follow-up

The NodaTime witness also shows the decompiler emitting the lowered early-exit guard sequence rather than recomposing the source's single `return A && B && … ;` boolean `&&`-chain. That guard-sequence → `&&`-return recomposition is a distinct readability raise, tracked in #3051 (out of scope here).

Closes #2955.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/IrImporterTests.cs
+++ b/src/ILInspector.Decompiler.Tests/IrImporterTests.cs
@@ -5604,13 +5604,14 @@ public class EnumConstantTests
     }
 
     [Fact]
-    public void NotOverRelationalOperatorCall_Parenthesizes()
+    public void NotOverRelationalOperatorCall_OutsideTotalOrderAllowlist_Parenthesizes()
     {
         // !(a > b) — a relational operator-spelled call renders as a compound
         // `a > b`, so the enclosing `!` must parenthesize it; a bare `!a > b`
-        // binds as `(!a) > b` (CS0023). Unlike equality/inequality, relational
-        // operator calls are NOT folded (!(a>b) is not always a<=b for a
-        // user-defined partial order), so the un-folded node must still be
+        // binds as `(!a) > b` (CS0023). System.Type is NOT in the total-order
+        // relational allowlist (MemberIdentity.IsTotalOrderRelationalOperator —
+        // decimal/DateTime/DateTimeOffset/TimeSpan/DateOnly/TimeOnly), so this
+        // relational call is NOT folded and the un-folded node must still be
         // built and parenthesized directly (#2955).
         var type = TypeRef.CoreLib("System", "Type");
         var boolType = TypeRef.CoreLib("System", "Boolean");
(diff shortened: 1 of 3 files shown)


richlander:
## Adversarial review reconciliation (fixed head `713c3cb9`)

Two-model review, both isolated in separate exact-head worktrees.

### Gemini 3.1 Pro — CLEAN (0 findings)
Independently verified all attack points: allowlist soundness (incl. a `DateTimeOffset` same-UTC-instant / different-offset check confirming `!(a < b)` and `a >= b` are exact De Morgan duals), `Half`/`Single`/`Double` exclusion, predicate precision, parenthesization/precedence (the `Precedence.Unary` misreport is benign because the folded result is `bool` and `Operand()` conservatively wraps non-atomic `LogicalNot`), and print-site completeness. No soundness or evidence-integrity issues.

### GPT-5.6 — 1 finding (attack #3), dispositioned as pre-existing / out-of-scope
**Finding:** the `IsCoreLibraryType` gate is forgeable — `TypeRefDecoder.Canonical` maps corelib facade **simple names** to `TypeRef.CoreLibrary` with no PKT check, so a planted `System.Runtime.dll` (invalid PKT) defining a lying `System.Decimal` gets `!(a < b)` folded to `a >= b`.

**Verified:** reproduced end-to-end on head `713c3cb9` (hand-assembled forged `System.Runtime.dll` → product renders `RelNotLt` as `a >= b`).

**Disposition — not a regression from this PR:**
- Root cause is name-based corelib canonicalization in `TypeRefDecoder.Canonical`, contradicting `PlatformKeys.cs`.
- The **already-merged #2961** negated-equality fold shares the identical gate (forged `System.String` folds the same); the operator-*spelling* path renders forged `op_LessThan` as `<` with no gate at all.
- The PKT is discarded during canonicalization, so there is **no fix at the printer / `MemberIdentity` layer** — the correct fix must PKT-gate `Canonical` itself, a cross-cutting hardening covering all corelib-gated behavior.

Filed as follow-up hardening issue #3045. This PR matches shipped precedent and introduces no new vulnerability class.

### Net
All soundness/parenthesization/fidelity-direction attack points clean on both reviews. The single finding is a pre-existing systemic issue tracked in #3045. No code changes were made in response, so the reviewed head is the final head.

richlander:
Ready to merge

PR #3040, opened 2026-07-22
--repo: read authored source from a local git clone (member Original Source + Implementation Diff)

richlander:
## `--repo`: read authored source from a local git clone

Opt-in local-repository source acquisition for the `member` (Original Source) **and** `diff` (Implementation Diff) commands, complementing the merged local-path read (#3031). **Draft** — two-model adversarial review in progress.

### What it does
For a **reproducible (published)** build the portable PDB records a normalized document path plus a `raw.githubusercontent.com/<org>/<repo>/<sha>/<relpath>` SourceLink URL. When the user names one or more local clones with `--repo`, we read the **committed blob at that exact SHA from the git object store** (the committed object, not the working tree) via `git cat-file blob <sha>:<relpath>`, and accept it **only when its bytes match the PDB checksum**. A wrong repo, commit, or path self-rejects and the caller falls back to the network — so the checksum is the arbiter and no repo-URL matching is required. It slots into the same acquisition ladder as #3031, one provider ahead of the remote fetch, on **both** the member Original Source path (`ApiCommand.ResolveMethodSourceAsync`) and the Implementation Diff path (`AuthoredSourceAcquisition.AcquireMemberAsync`).

### Why
- **Private repos** — skip SourceLink auth entirely when the source is already cloned.
- **Offline / air-gapped** inspection of published packages.
- Avoid rate limits / network for bulk inspection.
- Unblocks a cheaper test-harness authored-source oracle (delete the vendored orphan-branch corpus; re-acquire on demand via this same `AcquireMemberAsync(repositoryPaths)` path) — follow-up PR.

### Why opt-in (not automatic)
No registry maps a repo URL to a local clone, and filesystem scanning is slow, ambiguous (forks/multiple clones), and a side-effect concern. The checksum solves *content* trust but not *discovery*, so discovery is user-provided via `--repo` (repeatable).

### Security
The SHA and path come from an **untrusted PDB**, so before shelling out: the SHA is validated hex (`^[0-9a-fA-F]{7,64}$`); the path is rejected if empty, contains NUL, or is option-like. `git` is invoked with `ArgumentList` (no shell), `GIT_TERMINAL_PROMPT=0`, `GIT_OPTIONAL_LOCKS=0`, a 15s timeout, and a 64 MB output cap. If `git` is absent or the commit/path isn't present, we return null and fall back to the network. Only `raw.githubusercontent.com` URLs are addressed (other hosts fall back to remote).

### Evidence (dogfood A/B/C/D, member path)
Self-contained throwaway repo with a **GitHub SourceLink** build at an **unpushed** commit `eec9c10a` (fake `acme/widget` origin, so the remote raw URL 404s):

| Case | Command | Result |
| --- | --- | --- |
| A | `member Widget.Sample -m Add:1 -S "Original Source"` (no `--repo`) | **empty** (remote 404) |
| B | ...`--repo <clone>` | **renders `Add` body** from local blob at `eec9c10a`, checksum-verified |
| C | ...`--repo <repo lacking the commit>` | **empty** (no false content) |
| D | tamper the **working-tree** file, commit unchanged, `--repo <clone>` | **still renders committed `Add`** — proves we read the git object at the SourceLink commit, not the working tree |

The diff path reuses the exact same checksum-arbitered provider core via a new `AcquireMemberAsync(repositoryPaths)` overload.

### Tests
- `LocalRepoSourceReadTests` (Services): URL parse accept/reject; git-integration reads against a real temp repo — hit, checksum mismatch, missing commit, missing path, non-repo dir, second-repo fallthrough — guarded to skip when git is unavailable.
- `MemberOptionsParserTests` / `DiffOptionsParserTests`: `--repo` populates `SourceRepositories`; defaults empty (both commands).
- Full solution builds; Services repo tests green; 110 parser tests green; real `diff --help` / `member --help` expose `--repo`.

### Scope / follow-ups (not in this PR)
- Non-GitHub SourceLink hosts (Azure DevOps, GitLab) — currently fall back to remote.
- README/docs for `--repo`.
- Optional narrow "automatic" case: CWD-ancestor repo that already has the commit.
- Harness orphan-branch migration (manifest + on-demand acquisition) — separate PR that builds on this.

### DIFF
--- a/src/DotnetInspector.Services.Tests/LocalRepoSourceReadTests.cs
+++ b/src/DotnetInspector.Services.Tests/LocalRepoSourceReadTests.cs
@@ -1,0 +1,299 @@
+using System.Diagnostics;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace DotnetInspector.Services.Tests;
+
+public class LocalRepoSourceReadTests
+{
+    const string Source =
+        "class Sample\n{\n    public int M()\n    {\n        return 1;\n    }\n}\n";
+
+    const string RelativePath = "src/Sample.cs";
+
+    // ---------- URL parsing (no git required) ----------
+
+    [Fact]
+    public void ParsesGitHubRawUrl_IntoShaAndPath()
+    {
+        bool ok = LocalRepoSourceAcquisition.TryParseGitHubRawUrl(
+            "https://raw.githubusercontent.com/dotnet/runtime/0123456789abcdef0123456789abcdef01234567/src/libraries/Foo.cs",
+            out string sha, out string path);
+
+        Assert.True(ok);
+        Assert.Equal("0123456789abcdef0123456789abcdef01234567", sha);
+        Assert.Equal("src/libraries/Foo.cs", path);
+    }
+
+    [Fact]
+    public void ParsesGitHubRawUrl_UnescapesPath()
+    {
+        bool ok = LocalRepoSourceAcquisition.TryParseGitHubRawUrl(
+            "https://raw.githubusercontent.com/o/r/0123456789abcdef0123456789abcdef01234567/src/My%20File.cs",
+            out _, out string path);
+
+        Assert.True(ok);
+        Assert.Equal("src/My File.cs", path);
+    }
+
+    [Theory]
+    // Wrong host: never address a non-GitHub URL as a github raw blob.
+    [InlineData("https://example.com/o/r/0123456789abcdef0123456789abcdef01234567/a.cs")]
+    // Non-hex commit segment.
+    [InlineData("https://raw.githubusercontent.com/o/r/not-a-sha/a.cs")]
+    // Missing path segment.
+    [InlineData("https://raw.githubusercontent.com/o/r/0123456789abcdef0123456789abcdef01234567")]
+    // Not an absolute URL.
+    [InlineData("raw.githubusercontent.com/o/r/0123456789abcdef0123456789abcdef01234567/a.cs")]
+    [InlineData("")]
+    [InlineData(null)]
+    public void RejectsNonAddressableUrls(string? url)
+    {
+        Assert.False(LocalRepoSourceAcquisition.TryParseGitHubRawUrl(url, out _, out _));
+    }
+
+    // ---------- git blob read (requires git) ----------
+
+    [Fact]
+    public void ReadsBlob_WhenChecksumMatches()
+    {
+        RequireGit();
+        byte[] content = Encoding.UTF8.GetBytes(Source);
+        var (repo, sha) = InitRepoWithFile(RelativePath, content);
+        try
+        {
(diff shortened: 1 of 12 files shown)


richlander:
## Adversarial review — reconciled (two clean fixed-head reviews)

Two-model adversarial review per `AGENTS.md` (Gemini + GPT, not Claude). Each reviewer worked in an isolated read-only worktree at the exact head; scratch under `$env:TEMP`; no mutation of tracked files.

**Reviewed heads:** first round `efb1aff3`; fixed head `30abf2fd`.

### Round 1 (head `efb1aff3`)
- **Gemini 3.1 Pro — clean (0 blocking).** Probed and verified: argument/rev-spec injection (args via `ArgumentList`, hex-only SHA can't be option-like; live git test showed `<sha>:../file` is rejected as *outside repository*, no FS/`.git/config` escape), checksum arbitration (absent/empty/unparseable checksum declines), wrong-repo/wrong-commit self-reject, default-path parity when `--repo` absent, call-site binding, and process robustness (git-absent, 64MB cap, 15s kill).
- **GPT-5.5 — 0 blocking, one non-blocking finding.** Oversized (>64MB) repo blobs: `ReadCappedStdout` stopped draining at the cap while the parent stayed in `WaitForExit(15s)`, so `git cat-file` blocked on a full stdout pipe and each oversized blob burned the full 15s timeout. Proven with the 69,611,794-byte tracked corpus blob (`elapsedMs=15144`). Also verified injection rejections (wrong host, leading `-`, NUL, escaped `..`, escaped absolute path, embedded colon) and byte-identical default path.

> Note: a first GPT-5.6 reviewer produced empty turns and was retired; GPT-5.5 was substituted so both reviews are substantive.

### Resolution — commit `30abf2fd`
Fixed the non-blocking finding: `ReadCappedStdout` now takes the `Process` and calls `process.Kill(entireProcessTree: true)` the instant the 64MB cap is exceeded, so git tears down immediately instead of blocking on the pipe. Accept/reject contract unchanged (still returns `null` for oversized). Verified on the 69MB blob: rejection dropped from ~15144ms to a few hundred ms / single-digit ms.

### Round 2 — fixed head `30abf2fd` (both clean)
- **Gemini 3.1 Pro — 0 blocking.** Confirmed `Kill` is thread-safe vs the main thread's `WaitForExit`/`ExitCode` read, contract unchanged, no stream/task leak (killing closes child pipe handles so `stderr` read can't hang). Reproduced ~290ms vs 15s.
- **GPT-5.5 — 0 blocking.** Reproduced the fix (~300–500ms, no 15s burn); no new deadlock/race/contract-change/leak; `LocalRepoSourceReadTests` 15/15.

### Merge
Brought current with `origin/main` (merge `68fde5d4`); incoming changes are disjoint from the reviewed files (decompiler work), so the fixed-head reviews stand. Rebuilt + reran targeted tests on the merged head: Services repo tests 15/15, parser tests 110/110.

Issue #3027, opened 2026-07-22
Decompiler taste: elide trailing optional arguments equal to their declared default at call sites


richlander:
## Summary

At call sites, the decompiler renders **trailing optional arguments explicitly** even when the argument is a compile-time constant equal to the parameter's declared default — e.g. it prints `hour.ToWords(GrammaticalGender.Feminine, null)` where the source wrote `hour.ToWords(GrammaticalGender.Feminine)`. Idiomatic C# (and `dotnet/runtime` source) omits such arguments. Propose an **optional-argument elision** taste transform: drop trailing arguments that are constants equal to their parameter's default.

This is a **taste / valid-but-different presentation** question, not a correctness bug. The two spellings are IL-identical, so it belongs to the canonical-form machinery in [docs/decompiler-taste.md](../blob/main/docs/decompiler-taste.md), analogous to expression-bodied members and the LINQ query-vs-fluent choice.

## Witness

Surfaced while reviewing #2983 (PR #3014). The C# overload is:

```csharp
public static string ToWords(this int number, GrammaticalGender gender,
                             CultureInfo? culture = null)   // culture is [Optional], default null
```

Original source: `hour.ToWords(GrammaticalGender.Feminine)`
Decompiled (correct, but valid-but-different): `hour.ToWords(GrammaticalGender.Feminine, null)`

The C# compiler bakes the default into every caller (`ldnull` in IL), so `ToWords(gender)` and `ToWords(gender, null)` compile to **identical IL**; the caller's choice to omit is unrecoverable from the call site alone.

## Why this is a taste question (three-class rule)

Per `decompiler-taste.md`:

- **Not class 2 (fidelity-erasing).** Eliding the default does **not** erase a distinction the IL makes — unlike `&`→`&&`, where the IL genuinely differs, omit-vs-explicit-default are the *same* IL. Nothing observable is lost.
- **Class 3 (no IL anchor — oracle tiebreaker).** The two forms are one IL equivalence class; we pick one canonical representative. `dotnet/runtime`'s `.editorconfig` has no rule here, so — exactly like the pointer-syntax precedent — the **dominant runtime source style** is the oracle, and runtime code omits optional arguments equal to their default.
- **Fixed point preserved.** Recompiling the elided form re-inserts the default and yields identical IL, and re-decompiling yields the elided form again.

## Proposed rule

At a **resolved** call, walk the positional arguments from the last backward and elide the longest run of trailing arguments where **each** is:

[...]

PR #3042, opened 2026-07-22
Elide trailing optional-argument defaults at call sites (#3027)

richlander:
## What

Class-3 taste transform (issue #3027): elide a trailing call argument when it is a compile-time constant equal to its parameter's declared C# default. IL-identical; renders `ToWords(GrammaticalGender.Feminine)` instead of `ToWords(GrammaticalGender.Feminine, null)`.

Optional parameters are a C# concept the IL type system erases (the compiler bakes the default into every call site). The default is sourced from **metadata** (`[Optional]` + `Constant`), mirroring the declaration-signature path in `ApiSurfaceExtractor` — the decompiler does not re-derive its own view of the signature.

## Witness (Humanizer `FrTimeOnlyToClockNotationConverter.Convert`)

- before: `hour.ToWords(GrammaticalGender.Feminine, null)`
- after:  `hour.ToWords(GrammaticalGender.Feminine)`

Overload: `ToWords(this int, GrammaticalGender, CultureInfo? culture = null)`.

## Guardrail (targeted conservative optimization, not a Roslyn overload-resolution clone)

Elision is declined whenever a same-named overload could rebind the shortened call. Sound over-approximation that declines when uncertain, validated by recompile/corpus fidelity.

**Import-time (`OptionalArgumentFacts`):**
- Only same-assembly `MethodDefinition` callees are stamped (cross-assembly/generic structurally decline).
- A shortened arity is unsafe if any same-named candidate ties the callee's leading parameter types (`TypeRef.Equals`).
- Candidate set = the declaring type's methods for a non-extension callee; **every extension method of that name across the assembly** (cached per reader) for an extension callee, because an extension call resolves in receiver syntax across all extension classes in scope.
- Any same-named **generic** sibling declines the whole callee (the "fewer declared parameters" tie-break lets a generic candidate unify and win, e.g. `Pick(v)` binds `Pick<T>(T)` over `Pick(int,int=0)`).
- Any candidate carrying **`[OverloadResolutionPriority]`** (C# 13) declines the whole callee — priority reorders applicable candidates ahead of betterness, so a differently-typed sibling can steal the shortened call.
- Extension callees additionally require the **receiver type to be rooted in another assembly** (`ReceiverIsCrossAssembly`), so no same-assembly instance method (which beats an extension) can capture the call, **and the input to be a single-module assembly manifest** — the assembly-wide scan reads only one reader's module, so it declines when `!reader.IsAssembly` (a bare netmodule hides its sibling modules) or `AssemblyHasMetadataModules` (a manifest links code-bearing netmodules the scan cannot see).

**Pass-time (`OptionalArgumentElisionPass`):**
- Every retained argument's `ResultType.Equals` its parameter type (guards reference-subtype rebinds, e.g. `Feed(Animal,int=1)` + `Feed(Dog)`).
- `HasThis` callees decline unless `arguments[0].ResultType.Equals(callee.DeclaringType)` (base/derived receiver steal).

**Accepted residual (corpus-backstopped):** cross-**assembly** extension competitors and cross-assembly instance methods on the receiver type. All same-assembly steal vectors are closed.

## Not elided (by design)

Every case below is IL-identical to the shorter spelling, yet the trailing argument is **kept** because eliding it would change binding or erase a real evaluation. Each is pinned by the named fixture in `OptionalArgumentElisionFixtures`.

**Overload steal — a same-named candidate would rebind the shortened call:**

- **Leading-signature tie (inert optional)** — `Log("x", false)` stays (not `Log("x")`). The exact one-argument overload `Log(string)` always wins over `Log(string, bool = false)` — a candidate that fills every parameter beats one that substitutes an optional default — so `verbose`'s default is **unreachable by omission**: no source could write `Log("x")` and land on the two-argument callee. The `false` is therefore necessarily explicit in source, and eliding it would silently retarget the call to the *other* method. The optional parameter is effectively dead here (arguably a poor API shape); the pass declines only so it is not fooled by it. _(`CallLogKeepsSiblingTie`)_

  ```csharp
  int Log(string message);                          // Log("x") always binds here
  int Log(string message, bool verbose = false);    // reachable only as Log("x", <bool>)
  ```

- **Reference-subtype receiver** — `Feed(d, 1)` where `d` is `Dog` stays; `Feed((Animal)d)` would rebind to the better-matching `Feed(Dog)`. _(`CallFeedKeepsSubtype`)_

  ```csharp
  int Feed(Animal a, int portion = 1);
  int Feed(Dog d);                                  // Feed(d, 1)  ↛  Feed(d)
  ```

- **Derived-receiver steal** — `r.Emit("x", false)` where `r` is `LoudReporter` stays; `r.Emit("x")` would bind `LoudReporter.Emit(string)`, a different method than the base `Reporter.Emit(string, bool = false)`. _(`CallReporterKeepsDerivedSteal`)_

- **Generic sibling** — `Pick(5, 0)` stays; `Pick(5)` would rebind to `Pick<T>(T)` under the "fewer declared parameters" tie-break. _(`CallPickKeepsGenericSibling`)_

  ```csharp
  int Pick(int value, int seed = 0);
  int Pick<T>(T value);                             // Pick(5, 0)  ↛  Pick(5)
  ```

- **`[OverloadResolutionPriority]`** — `Rank(5, 0)` stays; the attribute deprioritizes `Rank(int, int = 0)`, so `Rank(5)` would rebind to the differently-typed `Rank(long)`. _(`CallRankKeepsPriorityOverload`)_

- **Cross-class extension steal** — `s.Pin(2, null)` stays; `s.Pin(2)` would rebind to a same-named extension `Pin(this string, int)` declared in a *different* static class (found by the assembly-wide extension scan). _(`CallExtensionKeepsCrossClassSteal`)_

**Structural declines:**

- **Non-default breaks the run** — `Triple(1, 9, 3)` renders `Triple(1, 9)`: only the trailing `3` (equal to its default) drops. `9 ≠ 2` blocks eliding the leading `1` **even though `1` equals its own default**, because `Triple(9)` would misbind `9` to `a`. _(`CallTripleTrailingRun`)_

- **Cross-assembly callee** — `s.Split(';')` keeps the baked `StringSplitOptions.None`; v1 stamps same-assembly `MethodDefinition` callees only. _(`CallSplitCrossAssembly`)_

**Constant-only rule:** a trailing argument that merely *evaluates* to the default via a non-constant expression (e.g. `Speak(3, GetNote())` where `GetNote()` returns `null`) is a real evaluation the IL performs — it is never elided, since dropping it would erase a distinction the IL makes (class 2), not a spelling choice.

## Tests

`OptionalArgumentElisionPassTests` (12) over `OptionalArgumentElisionFixtures`: elides null/enum/instance-null trailing defaults; keeps sibling-tie, non-default-run, retained-subtype, cross-assembly, derived-receiver-steal, **cross-class extension steal**, **generic-sibling**, and **`[OverloadResolutionPriority]`** cases; lone-extension positive elides. The two multi-module (netmodule) declines are validated by a real `csc /addmodule` artifact canary (the test project is single-module).

## History

Adversarial review found four steal holes: cross-class extension steal + generic-sibling steal (round 1, fixed in `a0de48c5` — assembly-wide extension scan + receiver guard + generic-sibling decline), `[OverloadResolutionPriority]` steal (round 2, fixed in `0a2ee90a`), a manifest-links-netmodule extension steal (round 2, fixed in `422a3e49`), and a bare-netmodule extension steal (round 3, fixed in `326470b9` — decline extension elision when `!reader.IsAssembly`). Head `326470b9`.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/OptionalArgumentElisionFixtures.cs
+++ b/src/ILInspector.Decompiler.Tests/OptionalArgumentElisionFixtures.cs
@@ -1,0 +1,41 @@
+namespace ILInspector.Decompiler.Tests;
+
+// Fixtures for OptionalArgumentElisionPass. Each caller below is what the tests
+// decompile; the callees declare the C# optional parameters whose baked defaults
+// the pass tries to elide from the call site.
+public class OptionalArgumentElisionFixtures
+{
+    // Single overload, reference-type null default: the canonical elide case.
+    static int Speak(int count, string? note = null) => count + (note?.Length ?? 0);
+
+    // Single overload, enum default (mirrors the Humanizer ToWords witness).
+    static int Announce(int count, System.DayOfWeek day = System.DayOfWeek.Sunday) => count + (int)day;
+
+    // Overload pair whose shorter member ties the callee's leading signature, so
+    // eliding verbose would rebind Log(message, false) to the distinct Log(message).
+    static int Log(string message) => message.Length;
+    static int Log(string message, bool verbose = false) => verbose ? 1 : message.Length;
+
+    // Three defaulted parameters, to exercise the trailing-run boundary.
+    static int Triple(int a = 1, int b = 2, int c = 3) => a + b + c;
+
+    // Reference-subtype hazard: Feed(Dog) would win a shortened Feed((Animal)d),
+    // so the pass-time identity check must keep the explicit trailing argument.
+    class Animal { }
+    sealed class Dog : Animal { }
+    static int Feed(Animal a, int portion = 1) => portion;
+    static int Feed(Dog d) => 0;
+
+    // --- Callers under test ---
+
+    public int CallSpeakElidesNull() => Speak(3);
+    public int CallAnnounceElidesEnum() => Announce(3);
+    public int CallLogKeepsSiblingTie() => Log("x", false);
+    public int CallTripleTrailingRun() => Triple(1, 9, 3);
+    public int CallFeedKeepsSubtype()
+    {
+        Dog d = new Dog();
+        return Feed(d, 1);
+    }
+    public string[] CallSplitCrossAssembly(string s) => s.Split(';');
+}
--- a/src/ILInspector.Decompiler.Tests/OptionalArgumentElisionFixtures.cs
+++ b/src/ILInspector.Decompiler.Tests/OptionalArgumentElisionFixtures.cs
@@ -26,10 +26,24 @@ public class OptionalArgumentElisionFixtures
     static int Feed(Animal a, int portion = 1) => portion;
     static int Feed(Dog d) => 0;
 
+    // Instance method (HasThis) with a reference-type null default: the common
+    // real call shape, exercising the receiver-offset path in the pass.
+    string Greet(string name, string? title = null) => title ?? name;
+
+    // Receiver-subtype steal hazard: the optional method is on the base, and the
+    // derived receiver declares a shorter same-named overload that captures the
+    // shortened call. r.Emit("x", false) binds Reporter.Emit(string, bool) but
+    // r.Emit("x") would rebind to LoudReporter.Emit(string) — a different method.
+    // The importer's sibling scan only sees the callee's declaring type (Reporter),
+    // so the pass-time receiver-static-type guard must keep the explicit argument.
+    class Reporter { public virtual string Emit(string message, bool loud = false) => loud ? message : ""; }
+    sealed class LoudReporter : Reporter { public string Emit(string message) => message + "!"; }
+
     // --- Callers under test ---
 
     public int CallSpeakElidesNull() => Speak(3);
     public int CallAnnounceElidesEnum() => Announce(3);
+    public string CallGreetElidesNull() => Greet("Ada");
     public int CallLogKeepsSiblingTie() => Log("x", false);
     public int CallTripleTrailingRun() => Triple(1, 9, 3);
     public int CallFeedKeepsSubtype()
@@ -38,4 +52,10 @@ public class OptionalArgumentElisionFixtures
         return Feed(d, 1);
     }
     public string[] CallSplitCrossAssembly(string s) => s.Split(';');
+
+    public string CallReporterKeepsDerivedSteal()
+    {
+        LoudReporter r = new LoudReporter();
+        return r.Emit("x", false);
+    }
 }
--- a/src/ILInspector.Decompiler.Tests/OptionalArgumentElisionFixtures.cs
+++ b/src/ILInspector.Decompiler.Tests/OptionalArgumentElisionFixtures.cs
@@ -39,6 +39,13 @@ public class OptionalArgumentElisionFixtures
     class Reporter { public virtual string Emit(string message, bool loud = false) => loud ? message : ""; }
     sealed class LoudReporter : Reporter { public string Emit(string message) => message + "!"; }
 
+    // Generic-sibling steal hazard: Pick(5, 0) binds the non-generic
+    // Pick(int, int = 0), but eliding to Pick(5) would rebind to the generic
+    // Pick<int>(int) — the "fewer declared parameters" tie-break beats the
+    // optional-using candidate. Any same-named generic sibling declines elision.
+    static int Pick(int value, int seed = 0) => value + seed;
+    static int Pick<T>(T value) => value?.GetHashCode() ?? 0;
+
     // --- Callers under test ---
 
     public int CallSpeakElidesNull() => Speak(3);
@@ -58,4 +65,42 @@ public class OptionalArgumentElisionFixtures
         LoudReporter r = new LoudReporter();
         return r.Emit("x", false);
     }
+
+    // A lone extension on a cross-assembly receiver (string): the witness shape
+    // (ToWords is itself a static extension). No competing same-named extension
+    // exists, so the trailing null default drops in receiver syntax.
+    public string CallExtensionElidesTrailingNull(string s) => s.WithTag(2);
+
+    // Cross-class extension steal: eliding the trailing null would rebind
+    // s.Pin(2, null) to StealerExtensions.Pin(string, int). The assembly-wide
+    // extension scan finds the tie and keeps the explicit argument.
+    public string CallExtensionKeepsCrossClassSteal(string s) => s.Pin(2, null);
+
+    // Generic sibling present on the declaring type: the explicit trailing 0 stays.
+    public int CallPickKeepsGenericSibling() => Pick(5, 0);
+}
+
+// --- Extension-method fixtures (assembly-wide overload scan + receiver guard) ---
+// Extension methods must live in top-level, non-generic static classes, so these
+// sit beside the fixture class rather than nested inside it.
+
+internal static class TagExtensions
+{
+    public static string WithTag(this string value, int weight, string? tag = null)
+        => tag is null ? $"{value}:{weight}" : $"{value}:{weight}:{tag}";
+}
+
+// The intended callee: a trailing optional after one real parameter.
+internal static class IntendedExtensions
+{
+    public static string Pin(this string value, int weight, string? tag = null)
+        => tag is null ? $"{value}#{weight}" : $"{value}#{weight}#{tag}";
+}
+
+// The thief: a shorter same-named extension on the same receiver that ties the
+// intended callee's leading signature at the shortened arity, in a different
+// class the old declaring-type-only scan never saw.
+internal static class StealerExtensions
+{
+    public static string Pin(this string value, int weight) => $"{value}!{weight}";
 }
--- a/src/ILInspector.Decompiler.Tests/OptionalArgumentElisionFixtures.cs
+++ b/src/ILInspector.Decompiler.Tests/OptionalArgumentElisionFixtures.cs
@@ -1,5 +1,7 @@
 namespace ILInspector.Decompiler.Tests;
 
+using System.Runtime.CompilerServices;
+
 // Fixtures for OptionalArgumentElisionPass. Each caller below is what the tests
 // decompile; the callees declare the C# optional parameters whose baked defaults
 // the pass tries to elide from the call site.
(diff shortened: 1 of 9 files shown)


richlander:
## Review (automated, per PTAL on #3027)

**Verdict:** Design is sound and behavior is verified locally; a few merge gates remain. Not marking ready.

### Verified locally (head `5d133595`, Windows)

- `ILInspector.Decompiler` builds clean (0 warn / 0 err).
- `OptionalArgumentElisionPassTests` — **7/7 pass**, including the instance-method (`HasThis`) elision case and the negative fixtures (sibling leading-signature tie, `Dog`/`Animal` subtype hazard, non-default breaks the trailing run, cross-assembly declines).
- Full fast decompiler suite (`-trait- Speed=Slow`): **3245 total, 55 failed — all environmental, none from this pass.** The 55 are the known Windows Roslyn compile-back gates failing at reference load (`CS0009` on native `aspnetcorev2_inprocess.dll` / `msquic.dll` / `System.IO.Compression.Native.dll` — `FidelityCheckGeneratedFilterTests`, `ReturnToSenderFixtureCatalogTests`, `TypeSourceCheckTests`, `LadderRung6GateTests`, `NullCoalescingAssignmentPassTests`) plus one CRLF-vs-LF assertion (`StringSwitchRaisingTests`, identical content). Failure count is identical with and without the pass's changes.

### Design assessment

- Layering matches #3019: the `[Optional]`+`Constant` fact is decoded in `OptionalArgumentFacts` (Metadata-owned) and consumed by a presentation-only `OptionalArgumentElisionPass`; no second decode in the decompiler. Good.
- Two-tier guardrail is well-reasoned: metadata `SafeTrailingElidableCount` (leading-signature tie detection; generics decline to 0) **plus** a pass-time identity check on retained arguments. `5d133595` closes the receiver-subtype steal via that identity check. Sound conservative under-approximation, corpus fidelity as the empirical backstop.
- Extension-method cross-class overload steal appears **not reachable**: the printer has no extension-receiver rendering, so extension calls print in static class-qualified form (`Class.Ext(x, …)`), scoping overload resolution to the declaring class (worth a one-line confirmation).

### Gates before merge

1. **Corpus `--pass-impact` blast-radius read.** The pass is registered always-on in `IrPasses.Default` and touches every call site; #3027 explicitly requires this measurement before defaulting on. Compile-back can't run on Windows here (`CS0009`), so this must be produced on Linux/CI.
2. **Two-model adversarial review** (new heuristic/shape, per `AGENTS.md`).
3. **Scope note (not a blocker):** v1 stamps same-assembly `MethodDef` callees only, so the motivating Humanizer `ToWords` witness (cross-assembly) is not yet elided. Cross-assembly defaults via the resolver seam are the anticipated follow-up in the issue.

richlander:
## Round-1 adversarial review — reconciliation

Two reviewers (Gemini 3.1 Pro, GPT-5.6) reviewed head `5d133595` in isolated worktrees. **Two blocking findings, both confirmed and fixed** in `a0de48c5`.

**Finding 1 — cross-class extension steal** (GPT + Gemini, compiled repros). The sibling scan only looked at the callee's declaring static class, but an extension call resolves `M` across every extension class in scope. A shorter same-named extension elsewhere in the assembly could rebind the shortened call.
*Fix:* for an extension callee, scan **every** extension method of that name across the assembly (cached per `MetadataReader`), and require the receiver type to be rooted in another assembly so no same-assembly instance method can capture it (`ReceiverIsCrossAssembly`).

**Finding 2 — generic-sibling steal** (GPT; I independently reproduced via a compiled probe — `Pick(5)` binds `Pick<int>(int)`, not `Pick(int,int=0)`). Gemini incorrectly cleared this ("non-generic always preferred"); the "fewer declared parameters" tie-break fires first, so a generic candidate can unify and win.
*Fix:* any same-named generic sibling declines the whole callee.

Both repros are now regression tests (`CrossClassExtensionStealsShortenedCall_KeepsExplicitArgument`, `GenericSiblingOverload_KeepsExplicitArgument`). Witness preserved (Humanizer `ToWords` still elides — no competing same-assembly `ToWords` extension). Accepted residual: cross-**assembly** competitors, corpus-backstopped.

Re-review of the fixed head `a0de48c5` is in progress.

richlander:
## Round-2 adversarial review — reconciliation

Re-reviewed the fixed head `a0de48c5` in isolated worktrees.

- **Gemini 3.1 Pro: CLEAN.** Verified both round-1 fixes closed; endorsed the type-level `[Extension]` fast-path skip as sound (Roslyn discovers PE extension methods only through containers carrying the type-level attribute, so a method-level-only class cannot steal receiver-syntax binding); 12/12 tests. *(Note: this clean review was of `a0de48c5`, superseded below.)*
- **GPT-5.6: one new blocking finding** — `[OverloadResolutionPriority]` (C# 13) reorders applicable candidates ahead of betterness, so a differently-typed sibling that the leading-signature check ignores can steal the shortened call.

**Independently reproduced** (compiled): `Rank(int, int = 0)` marked `[OverloadResolutionPriority(-1)]` with a sibling `Rank(long)` — `Rank(5, 0)` binds the int overload, but `Rank(5)` binds `Rank(long)`. The decompiler elided the trailing `0`, rebinding to a different method.

**Fixed in `0a2ee90a`:** detect the attribute per candidate and decline the whole callee when any candidate in the scan set (callee or sibling) carries it. Added a compiled regression fixture/test (`OverloadResolutionPrioritySibling_KeepsExplicitArgument`). Witness preserved; 12/12 tests pass; the repro now keeps `Rank(5, 0)`.

Both reviewers are re-reviewing the new head `0a2ee90a`. Will confirm two clean reviews before marking ready.

richlander:
## Round-2 (cont.) — second finding: multi-module (netmodule) extension steal

After the `[OverloadResolutionPriority]` fix, GPT-5.6 found a fourth same-assembly steal vector: the assembly-wide extension scan (`BuildExtensionSiblingIndex`) reads only the manifest module's `reader.TypeDefinitions`, so a competing extension in a linked **netmodule** (multi-module assembly) is invisible.

**Independently reproduced** with a compiled multi-module assembly (`csc /target:module` + `/addmodule`): `IntendedExtensions.Pin(this string, int, string = null)` in the manifest, `StealerExtensions.Pin(this string, int)` in the netmodule. `s.Pin(2, null)` elided to `s.Pin(2)`, rebinding to the netmodule's `Pin` — a same-assembly steal, not the accepted cross-assembly residual.

**Fixed in `422a3e49`:** decline extension elision when the manifest links any code-bearing module file (`AssemblyHasMetadataModules` via `reader.AssemblyFiles` / `AssemblyFile.ContainsMetadata`). Non-extension callees are unaffected (a type and all its methods live in one module). Verified: the multi-module repro now keeps `s.Pin(2, null)`; the single-module witness still elides; 12/12 tests pass. Validated by the real multi-module artifact canary (a netmodule fixture isn't expressible in the single-module test project).

Both reviewers are now re-reviewing the final head `422a3e49`.

richlander:
## Round 3 reconciliation (head `422a3e49` → fix `326470b9`)

**Gemini (gemini-3.1-pro-preview): CLEAN on `422a3e49`.** Verified the manifest-links-netmodule fix sound/complete, all prior findings closed, 12/12 tests.

**GPT (gpt-5.6-sol): NEW same-assembly under-decline on `422a3e49`.** `AssemblyHasMetadataModules(reader)` short-circuits to `false` when the inspected file is *itself* a bare netmodule (`reader.IsAssembly == false`). Decompiling that netmodule directly then runs the assembly-wide extension scan over only the netmodule's own `TypeDefinitions`, missing a same-named extension in a sibling module — so `s.Pin(2, null)` elided to `s.Pin(2)`, which rebinds to the sibling module's `StealerExtensions.Pin` on recompile. A same-assembly steal, not the accepted cross-assembly residual.

**Independently reproduced (pass-level, decompiler harness).** Built a compiled multi-module assembly (`csc /target:module` + `/addmodule`): `Intended.netmodule` holds `IntendedExtensions.Pin(this string, int, string = null)` and the caller; `Stealer.netmodule` holds `StealerExtensions.Pin(this string, int)`; `Main.dll` links both. `decompiler-harness Intended.netmodule --dump 'MultiModuleProbe.Caller::CallExplicit'`:
- pre-fix (`422a3e49`): `return s.Pin(2);` ❌ (elided — rebinds to the stealer)
- post-fix (`326470b9`): `return s.Pin(2, null);` ✅ (kept)

(The product's `member`/`type` commands reject bare netmodules at API extraction, so this is reachable at the pass level, e.g. via the harness — the fix makes the pass sound regardless of caller.)

**Fix `326470b9`:** `SafeTrailingElidableCount`'s extension branch now also declines when `!reader.IsAssembly`. The assembly-wide extension scan is complete only for a single-module assembly manifest. Non-extension callees are unaffected: a type and all its methods live in one module, so `DeclaringTypeSiblings` sees the full candidate set even for a bare module.

**Verification:** single-module witness (Humanizer `ToWords`) still elides both sites; 12/12 focused tests pass. No unit fixture (the test project is single-module; validated by the multi-module artifact canary, same as the `422a3e49` fix).

Both reviewers re-tasked on `326470b9`. Will post `Ready to merge` only after two clean reviews of that head.

richlander:
## Round 4 — both reviews CLEAN on head `326470b9`

Both adversarial reviewers re-reviewed the exact fixed head `326470b9`:

- **Gemini (gemini-3.1-pro-preview): CLEAN.** Bare-netmodule fix sound/complete (`!reader.IsAssembly` over-declines when sibling modules are invisible); non-extension path correctly unaffected (a `TypeDefinition` cannot span modules by metadata design); all prior fixes intact; 12/12 tests.
- **GPT (gpt-5.6-sol): CLEAN.** Confirmed its bare-netmodule repro now keeps `s.Pin(2, null)`; manifest-linked netmodule repro also keeps the explicit null; resource-only files correctly excluded via `ContainsMetadata`; cross-class/generic/priority fixes closed; 12/12 tests.

### Findings ledger (attribution)

| # | Round | Reviewer | Finding | Fix |
| - | ----- | -------- | ------- | --- |
| 1 | 1 | GPT | cross-class extension steal | `a0de48c5` (assembly-wide extension scan) |
| 2 | 1 | GPT | generic-sibling steal | `a0de48c5` (generic-sibling decline) |
| 3 | 2 | GPT | `[OverloadResolutionPriority]` steal | `0a2ee90a` |
| 4 | 2 | GPT | manifest-links-netmodule steal | `422a3e49` |
| 5 | 3 | GPT | bare-netmodule steal | `326470b9` (`!reader.IsAssembly` decline) |

Gemini gave clean reviews at `422a3e49` and `326470b9` and (in round 1) initially cleared the generic-sibling case that GPT correctly flagged; the generic tie-break rule was verified independently (`Pick(5)` binds `Pick<T>(T)` over `Pick(int,int=0)`) and is now covered by a fixture. Every finding was independently reproduced (compiled repro / decompiler-harness canary) before fixing.

Holding `Ready to merge` until the `test (ubuntu-26.04, linux-x64)` job (Linux corpus-fidelity gate) goes green; will post it then.

richlander:
Ready to merge

PR #3044, opened 2026-07-22
Add per-member whole-render API to MemberBodyProducer (#2996)

richlander:
## What & why

Umbrella #2996 makes the **whole member** the decompiler''s natural output unit — Metadata owns the signature model, CSharp owns spelling + layout, Decompiler owns only the body, with **one composition and no drift**.

The product already renders a whole **type** as one listing (`MemberBodyProducer.Project`) and a **body-only** block (`ProduceBody`), but had **no way to render a single member''s `signature { body }`**. That gap blocks the final track (**pr5**): retiring the compile-back harness''s private skeleton signature builder in `FidelityCheck.cs`, which today re-spells member signatures itself and therefore drifts from the product.

This PR adds that missing enabler.

## Change

- **New public API** `MemberBodyProducer.ProduceMember` (path-only + `IAssemblyReferenceResolver` overloads) returning `MemberRenderResult(Status, Text, Namespaces)`.
  - `Text` is **byte-identical** to the member''s segment in the whole-type `Project` listing (indent 4, no surrounding blank-line separators).
  - `Namespaces` are the imports a wrapping consumer emits as `using` directives.
  - `member` must be an instance from `type.Members` (matched by reference); non-emitting members (enum values, etc.) return `Absent`; failures degrade honestly to `Failed` + a `// DI_...` comment.

- **Reuses the existing whole-type composition** rather than adding a second path:
  - `ComposeMembers` gains `ApiMember? only = null`. When set, the loop does all bookkeeping (overload counting, union hidden-member skip) but emits only the reference-matched member with `first == true` (no leading blank line), then returns. When `null`, the whole-type hot path is a **strict no-op**.
  - `HoistUsings` is split: the shared **`ShortenQualifiedNames`** (qualified-name shortening + deterministic `\n` normalization + import harvest) is now used by **both** the whole-type `HoistUsings` (which then prepends directives) and the per-member `ComposeMemberCore` (whose caller emits imports). This is what guarantees the per-member text matches its whole-type segment.

## Evidence

- **Byte-identity oracle** (`MemberBodyProducerMemberRenderTests`): every specimen member''s `ProduceMember` `Text` is a substring of the `Project` listing, plus throw-stub (`=> throw`), arrow, and block-body witnesses. 4/4 green.
- **Controlled A/B over the full fast Decompiler suite** (base `f8f6a449` vs head, identical `-notrait Speed=Slow` selection): **identical 56-failure set — zero head-only regressions**. The 56 are the pre-existing **#3018-C** Windows-local CS0009 recompile-env failures (native runtime DLLs in Roslyn''s reference set); CI is green on these.

## Compatibility / scope

Purely **additive** public API + an **output-preserving** refactor of the shared using-hoist path. No behavior change to any existing whole-type or body-only output. Union-synthesized explicit constructors (not reachable via `type.Members`) are out of scope.

## Follow-up

Unblocks **pr5**: point `FidelityCheck.BuildUnit` at `ProduceMember`, delete the harness''s private target-member signature spelling, and re-triage the CS0305/CS1031/CS1750 EVIL false-`Invalid` cluster.

Part of #2996.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/MemberBodyProducerMemberRenderTests.cs
+++ b/src/ILInspector.Decompiler.Tests/MemberBodyProducerMemberRenderTests.cs
@@ -1,0 +1,111 @@
+using System.Reflection.PortableExecutable;
+
+using ILInspector.Decompiler;
+using ILInspector.Metadata;
+
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// Pins the per-member whole render (<see cref="MemberBodyProducer.ProduceMember"/>)
+/// against the whole-type listing (<see cref="MemberBodyProducer.Project"/>): the
+/// text produced for a single member is byte-identical to that member's segment
+/// in the listing — one composition, no drift. This is the #2996 output-contract
+/// enabler: CSharp-owned signature + decompiler-owned body, per member.
+/// </summary>
+[Trait("Area", "RoundTrip")]
+public sealed class MemberBodyProducerMemberRenderTests
+{
+    static string AssemblyPath => typeof(MemberBodyProducerMemberRenderTests).Assembly.Location;
+
+    static ApiType Specimen()
+    {
+        using var pe = new PEReader(File.OpenRead(AssemblyPath));
+        var api = ApiSurfaceExtractor.Extract(pe);
+        return Assert.Single(api.Types, t => t.FullName == typeof(MemberRenderSpecimen).FullName);
+    }
+
+    [Fact]
+    public void ProduceMember_ByteIdenticalToWholeTypeSegment_ForEveryMember()
+    {
+        var type = Specimen();
+        var listing = MemberBodyProducer.Project(type, AssemblyPath, pdbPath: null).Output;
+        Assert.NotNull(listing);
+
+        Assert.NotEmpty(type.Members);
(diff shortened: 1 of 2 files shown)


richlander:
Ready to merge

- **Scope**: purely additive public API (`MemberBodyProducer.ProduceMember` / `MemberRenderResult`) plus an output-preserving refactor of the shared `HoistUsings` → `ShortenQualifiedNames` path. No behavior change to any existing whole-type or body-only output.
- **Preservation evidence**: byte-identity oracle (every specimen member''s `ProduceMember` `Text` is a substring of the whole-type `Project` listing) 4/4 green; controlled A/B over the full fast Decompiler suite (base `f8f6a449` vs head, identical `-notrait Speed=Slow` selection) yields an **identical 56-failure set — zero head-only regressions**. The 56 are the pre-existing #3018-C Windows-local CS0009 recompile-env failures; CI is green.
- **CI**: build-net10, csharp-diff-smoke, test (ubuntu-26.04, linux-x64), changes all pass.
- **Review**: treated as mechanical (output-preserving + additive, byte-identity-proven), consistent with the #3038 precedent.

Unblocks pr5 (retire the compile-back harness''s private skeleton signature builder).

Issue #3028, opened 2026-07-22
Fully raise InlineArrayCollectionPass::PlaceFromAddress to its switch-expression form


richlander:
## Follow-up frontier from #3003

#3003 brings `ILInspector.Decompiler.Pipeline.InlineArrayCollectionPass::PlaceFromAddress` from **invalid** (CS0161, falls off the end) to **valid** (all paths return) by keeping the interleaved default-goto default as the region's trailing terminator (the mixed return-tail candidate in `ReturnMergePass`).

That reaches **validity but not a fully-raised idiomatic form**. The original source is a single switch expression:

```csharp
static IrExpression? PlaceFromAddress(IrExpression address, TypeRef arrayType) => address switch
{
    LoadLocalAddress local => new LoadLocal(local.Index, local.Type),
    LoadArgumentAddress argument => new LoadArgument(argument.Index, argument.Name, argument.Type),
    LoadFieldAddress field => new LoadField(field.Field, field.Instance is null ? null : (IrExpression)field.DetachChildren()[0]),
    LoadElementAddress element => element,
    LoadLocal { ResultType: { Kind: TypeRefKind.ByRef, ElementType: { } element } } local when element.Equals(arrayType) => new LoadIndirect(arrayType, local),
    _ => null,
};
```

The current decompiled output is a valid but scaffolded nested if-ladder with six decompiler temporaries (`V_4, V_5, V_7, V_8, S_0, S_1`) rather than the idiomatic switch expression. Full raising should recover the type-pattern switch arms (including the `ByRef`/`ElementType` nested pattern with the `when` guard) and eliminate the invented temporaries.

Tracks the raising endpoint beyond the validity floor established in #3003.

richlander:
## Scoping: what a full raise of `PlaceFromAddress` requires

I reproduced the exact post-pipeline IR (import the shipped method, `IrPasses.Run`, dump the tree). Recording the concrete gap so the work can be sequenced.

### Current output (verified)

The cascade after the switch-value store `V_7 = address` is **heterogeneous** — csc did not lower every arm the same way:

- **Arm 1** (`LoadLocalAddress local`): `store L0 = isinst LoadLocalAddress(V_7); if (!L0) { REST } return new LoadLocal(L0.Index, L0.Type)` — the `as`/null-test intro form with the matched body **trailing** the dispatch test. This is exactly the shape `PatternSwitchExpressionPass` (#3022) already recognizes.
[...]

Issue #3022, opened 2026-07-22
Fully raise result-temp switch-expression if/else back to switch expression (TryNormalizeContinueCondition)


richlander:
Follow-up to #2973 (validity fix).

After #2973, `YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition` decompiles to **valid, correct** C#, but not to its fully raised shape. The original is a `switch` expression with `when`-guarded type patterns:

```csharp
return expression switch
{
    Comparison comparison when ReadsLoopField(comparison.Left, loopFieldName) => Assign(comparison, out condition),
    LogicalNot { Operand: Comparison comparison } when ReadsLoopField(comparison.Left, loopFieldName)
        => Assign(Negate(comparison), out condition),
    _ => false,
};
```

The current (post-#2973) output is a semantically equivalent `is`-pattern if/else cascade over a `V_3`/`V_1`/`V_4` result-temp chain that returns `false` on the default path. Fully raising it requires recognizing the result-temp switch-expression shape (arm = `when`-guarded type-pattern test → `Assign(...)`, default = `_ => false`) and recomposing it into a `switch` expression, rather than emitting the lowered if/else.

## Remaining slice

- Recover the `switch` expression (with `when`-guarded `Comparison` / `LogicalNot { Operand: Comparison }` arms and a `_ => false` default) from the raised result-temp pattern if/else in `TryNormalizeContinueCondition`.

## Acceptance

- `TryNormalizeContinueCondition` decompiles to the `switch`-expression spelling (fully raised), preserving validity and behavior.
- No regression in the decompiler fast suite or the corpus PR quick gate.

PR #3046, opened 2026-07-22
Raise inline is-pattern if/return cascades to switch expressions (#3028 PR A)

richlander:
- Advances #3028 (PR A of the staged plan)
- Changes the decompiler to raise flat inline `if (P is Tk xk) { ... }` sibling-arm cascades to their idiomatic C# `switch` expression
- Card revision: `5cf69613`

## Change

**Conclusion:** **PASS** — an additive second recognizer folds a real csc lowering shape the intro-chain parser cannot see, reusing existing IR nodes with zero printer/DA/ledger churn and zero corpus regressions.

`PatternSwitchExpressionPass` (#3022) recognized only the nested `as`/null-test *intro-chain* shape. csc lowers an unguarded `P switch { T0 x0 => v0, ..., _ => d }` to a **flat** run of positive inline `if (P is Tk xk) { ... }` arms whenever every arm's bound local is used only inside its own matched branch — `IsPatternPass` then folds each `x = P as Tk; if (x != null)` arm to a positive `is` test and the remaining arms follow as *siblings*, not nested. The intro-chain parser never sees that shape, so the cascade stays an if-ladder.

`TryMatchInlineArms` is a second recognizer reusing the existing `PatternSwitchExpression`/`PatternSwitchExpressionArm`/`PropertySubpattern` nodes and the shared matched-body / default-tail parsing helpers. It requires two or more contiguous inline arms over one re-evaluable place (`LoadArgument`/`LoadLocal`) that is not reassigned across the cascade, each arm's matched body yielding a non-default value with every other path reaching the shared trailing `return <default>`, and each pattern-bound local referenced only within the cascade.

Faithfulness: the arms are tried top-to-bottom exactly as the if-ladder already does, so an unguarded fold reorders nothing regardless of type overlap; a guarded arm routes its guard-fail to the shared default — the same mutual-exclusivity evidence the intro-chain shape relies on.

### Original source

`System.Reflection.Emit.RuntimeMethodBuilder.GetMethodBaseReturnType` (System.Private.CoreLib), via `dotnet-inspect ... -S "Original Source"`:

```csharp
internal static Type? GetMethodBaseReturnType(MethodBase? method)
{
    if (method is MethodInfo mi)
    {
        return mi.ReturnType;
    }
    else if (method is ConstructorInfo ci)
    {
        return ci.GetReturnType();
    }
    else
    {
        Debug.Fail("We should never get here!");
        return null;
    }
```

`Debug.Fail` is `[Conditional("DEBUG")]`, so it is absent from the shipped Release IL; the shipped default path is exactly `return null`, which is what both Before and After render.

### Before

`-S "Decompiled Source"` at base `f8f6a449`:

```csharp
internal static System.Type GetMethodBaseReturnType(System.Reflection.MethodBase method)
{
    if (method is MethodInfo mi)
    {
        return mi.ReturnType;
    }
    if (method is ConstructorInfo ci)
    {
        return ci.GetReturnType();
    }
    return null;
}
```

- Valid: True
- Correct: True

An if-ladder — valid and correct, but not the idiomatic switch expression csc lowered from.

### After

`-S "Decompiled Source"` at head `5cf69613`:

```csharp
internal static System.Type GetMethodBaseReturnType(System.Reflection.MethodBase method)
{
    return method switch
    {
        MethodInfo mi => mi.ReturnType,
        ConstructorInfo ci => ci.GetReturnType(),
        _ => null,
    };
}
```

- Valid: True
- Correct: True

Same arms, order, and default as Before; the non-null-default shape is exercised by the additional real witness below.

Additional real witness — `Newtonsoft.Json.Converters.XContainerWrapper.WrapNode` (8 arms, non-null default), at head:

```csharp
internal static Newtonsoft.Json.Converters.IXmlNode WrapNode(System.Xml.Linq.XObject node)
{
    return node switch
    {
        XDocument V_0 => new XDocumentWrapper(V_0),
        XElement V_1 => new XElementWrapper(V_1),
        XContainer V_2 => new XContainerWrapper(V_2),
        XProcessingInstruction V_3 => new XProcessingInstructionWrapper(V_3),
        XText V_4 => new XTextWrapper(V_4),
        XComment V_5 => new XCommentWrapper(V_5),
        XAttribute V_6 => new XAttributeWrapper(V_6),
        XDocumentType V_7 => new XDocumentTypeWrapper(V_7),
        _ => new XObjectWrapper(node),
    };
}
```

### Fully raised

The After decompilation is in the fully raised state for the shape this PR targets (the flat inline cascade).

This PR does **not** fold `InlineArrayCollectionPass::PlaceFromAddress`, the #3028 endpoint: its arm 5 needs a recursive property-subpattern grammar (`{ ResultType: { Kind: ByRef, ElementType: { } element } }`) beyond the single-level `PropertySubpattern` model. That remains for the follow-up slices.

- Required tracking issue: #3028 — recursive property-subpattern arms (PR B) and the `PlaceFromAddress` compose endpoint (PR C).

## Evidence

| Check | Baseline (`f8f6a449`) | Head (`5cf69613`) |
| --- | --- | --- |
| Focused tests | `PatternSwitchExpressionPassTests`: 5 passed | `PatternSwitchExpressionPassTests`: 11 passed |
| Decompiler fast suite | 3237 passed, 0 failed | 3243 passed, 0 failed |
| Real witness | `RuntimeMethodBuilder::GetMethodBaseReturnType` if-ladder | folded to switch expression |

New tests: one compiled fixture (`InlinePatternSwitchSample.Simplify`, a 4-arm `Address` hierarchy that lowers to the flat inline cascade) plus five synthetics — a two-arm positive and four negatives (single-arm, different-receiver, interrupted-cascade, guard-fail-yields-nondefault).

## Decompiler quality

**Conclusion:** **PASS** — render A/B over the standing corpus shows the fold is purely additive: every change is a structural if-ladder→switch fold that stays semantically valid, with no valid→invalid regressions and nothing added or removed.

Render A/B (`--render-ab`, full corpus, no sampling), base `f8f6a449` vs head `5cf69613`:

| Metric | Value |
| --- | ---: |
| Methods evaluated | 141,764 |
| Changed (all structural) | 16 |
| Semantic valid→valid | 16 |
| Semantic valid→invalid | 0 |
| Added / Removed | 0 / 0 |

All 16 changed methods are if-ladder→switch-expression folds across System.Private.CoreLib, Roslyn, Newtonsoft.Json, and the product's own assemblies (e.g. `DeconstructionAssignmentPass::ReceiverValue`, `ForeachStatementPass::CollectionValue`), i.e. 16 additional fully-raised methods with zero collateral. Corpus: SPC + Newtonsoft + Roslyn (common/csharp) + System.CommandLine + NuGet.Versioning + ApplicationInsights + product assemblies.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.PatternSwitchExpressionPassTests
```


richlander:
## Adversarial review reconciliation — fixed head `411f68f5`

Conclusion-first: both fixed-head reviews are now **clean**. Reviewers: **Gemini `gemini-3.1-pro-preview`** and **GPT `gpt-5.6-sol`** (two different model families), each isolated in its own review worktree.

### Round 1 (head `71ae59ff`)

- **F1 — guarded-arm fall-through** (both): a `switch` routes a failed `when` to the next arm, but the `if`-ladder's guard-fail returns the shared default. Fixed by rejecting guarded/subpattern arms. (`71ae59ff`)
- **F2 — per-arm local scope leak** (both): a later arm could read a binding introduced by an earlier arm. Fixed with a per-arm subtree scope check. (`71ae59ff`)
- **F2b — unused-binding skips the scope check** (Gemini): the check now keys off the arm's actual bound slot (`IsPattern.LocalIndex`). (`fdf182b7`)
- **F2a — pre-existing `TryMatch` whole-`consumed` scope** (Gemini): inherited by the distinct compiler-lowered intro-chain shape (a different nested structure the inline path never produces); not exercised by this change and left unchanged.
- **F3 — subsumption → CS8510** (GPT **blocker**; Gemini non-blocking): a hand-written ladder whose earlier arm subsumes a later arm is valid as an `if`-ladder but emits **CS8510** ("the pattern is unreachable") as a `switch`. I reproduced GPT's end-to-end repro independently and treated it as blocking.

### F3 resolution — subsumption oracle (`fdf182b7`)

The inline shape is *also* what a hand-written type ladder produces, so — unlike csc's intro-chain lowering, which the compiler guarantees is non-subsuming — it carries no such proof. Added a metadata type-relationship oracle (`MetadataSource.Subsumes` / `CrossAssemblyTypeResolver.IsAncestorOrEqual`, stamped onto `IrFunction.TypeSubsumption` at import) and fold the inline shape **only when every ordered arm pair is proven non-subsuming**; `Yes`/`Unknown`/null-oracle all decline (conservative — `Unknown` is always correctness-safe, never a false `No`).

### Round 2 (variance, fixed head `411f68f5`)

- **Generic/array variance → false `No`** (GPT **blocker**): a later arm can be assignable to an earlier arm through a conversion with **no nominal base/interface edge** — generic co-/contravariance (`ICovariant<string>` is-a `ICovariant<object>`), array covariance (`string[]` is-a `object[]`). The nominal closure walk missed these and returned `No`, re-admitting CS8510. Every such conversion targets a constructed generic instance or an array, so `Subsumes` now returns `Unknown` when `earlier.Kind` is `GenericInstance`/`SzArray`/`Array`. (`411f68f5`)
  - **GPT re-verified: RESOLVED.** Covariance, contravariance, and array-covariance repros all now decline; unrelated sealed siblings still fold; 19/19.
  - **Gemini adversarially reviewed the gate: NO BLOCKERS** — confirmed the gate is complete (it even caught the obscure nested-generic-variance edge `IOuter<out T>.IInner`, which decodes to `GenericInstance` and is declined), `Unknown` propagates to a decline, and non-variant coverage is preserved.
- **`PlaceStableAcross` ignores ref-mutation of the switch value in a return arm** (Gemini nit): confirmed **strictly non-blocking** by both of us — a matched arm returns immediately, later arms are unreachable, and the governing value is evaluated once, so runtime semantics are identical. Not changed.

### Evidence

- Focused `PatternSwitchExpressionPassTests`: **19/19**. Fast decompiler suite: **3250/0**. Metadata: **709/0**.
- **Render A/B, 141,764 methods** (pre-variance-gate baseline): **7 folds, all structural `valid→valid`, 0 `valid→invalid`, 0 added/removed.** The variance gate only converts some `No`→`Unknown` (strictly fewer folds — a subset of those 7), so it cannot introduce `valid→invalid`; a confirmatory re-run at `411f68f5` is in progress.

### Design note — is this recovery of a `switch` expression, or a taste choice?

It is a **normalization choice**, and the IL is **distinguishable** from a lowered `switch` expression. Compiling the two source forms (Release):

Source **`switch` expression** → single result temp + single exit:

```il
isinst MethodInfo;      stloc.0; ldloc.0; brtrue.s  ARM0
isinst ConstructorInfo; stloc.1; ldloc.1; brtrue.s  ARM1
br.s DEFAULT
ARM0: ...; stloc.2; br.s END
ARM1: ...; stloc.2; br.s END
DEFAULT: ldnull; stloc.2
END: ldloc.2; ret            // one result local, one ret
```

Hand-written **`if (x is T t) return v;` cascade** → per-arm `ret`, no result temp:

```il
isinst MethodInfo;      stloc.0; ldloc.0; brfalse.s NEXT
ldloc.0; ...; ret            // arm returns immediately
NEXT: isinst ConstructorInfo; ...; brfalse.s DEFAULT
...; ret
DEFAULT: ldnull; ret         // multiple rets
```

The single-result-temp/single-exit form is the #3022 intro-chain (owned by the separate `TryMatch` path); the per-arm-`ret` flat form is the hand-written-cascade signature this PR folds. So we're choosing the equivalent `switch` spelling, not reconstructing an original `switch` expression — which is precisely **why the subsumption guard is required**: the hand-written cascade lacks csc's compile-time non-subsumption guarantee.

Example sets:

1. **Kept** — `System.Reflection.Emit.RuntimeMethodBuilder.GetMethodBaseReturnType` (real corpus fold). Source ends `Debug.Fail("We should never get here!"); return null;`; `Debug.Fail` is `[Conditional("DEBUG")]`, so the release binary strips the call, leaving the clean cascade we fold to `_ => null`. (If the guard statement were present — a Debug build or any non-`[Conditional]` statement before the default return — the fallthrough is not a bare `return d` and the fold declines, so no release-present side effect is ever dropped.)
2. **Source `switch` expression** — a simple 2-arm `method switch { MethodInfo mi => …, ConstructorInfo ci => …, _ => null }` compiles to the single-result-temp intro-chain IL above, distinct from the flat cascade.

Not marking ready until the confirmatory A/B lands and CI is green.

richlander:
## Merge with `main` resolved — and a scope-changing finding

Resolving the merge conflict surfaced that **#3065 already merged a more general, safer implementation of the same #3028 PR A feature**, explicitly deferring the type-disjointness oracle "to a follow-up carrying a real oracle." I've adopted #3065's matcher wholesale and re-framed this PR as **that oracle follow-up**. Head is now `8d628546`.

### What changed in this branch
- The merge (`7bacc8dc`) takes #3065's `TryMatch`/`TryFold`/`TryParseChain` matcher, **superseding this branch's earlier `TryMatchInlineArms`** (now gone).
- `8d628546` layers the subsumption oracle onto #3065's new-surface gate: `TryFold` calls `ArmsProvenDisjoint`, folding the heterogeneous inline/direct surface only when the metadata oracle proves no earlier arm subsumes a later one. `MetadataSource.Subsumes` + `CrossAssemblyTypeResolver.IsAncestorOrEqual` + the `IrFunction.TypeSubsumption` seam provide it, with a variance gate returning `Unknown` for constructed-generic/array `earlier` types.

### Important honesty note: the F3/CS8510 bug is **not reachable** through #3065's matcher
#3065 anchors every fold on a compiler-lowered **intro-chain head** (a `switch`-expression lowering, whose arms csc already proved mutually non-subsuming — else CS8510 at the original compile). A hand-written `if (v is Base b)… if (v is Derived d)…` ladder — which *can* subsume — lowers instead to the flat **all-inline** shape, which #3065 leaves unfolded.

Evidence: with the guard **disabled**, I decompiled nine subsuming shapes (Shape/Circle, `ICovariant<T>` variance, `IComparable`/`string`, else-if, flat, heavy-binding, three-arm) — **every one declines structurally, none fold.** A subsuming ladder can only ever be hand-written (a subsuming `switch` expression won't compile), and those never produce the intro-chain shape #3065 folds.

**Consequence:** the oracle returns `No` for every currently reachable fold and never blocks one. It is **defense-in-depth groundwork** — the deferred oracle made explicit — that keeps a future extension (folding hand-written ladders, or #3065's deferred guarded/subpattern arms) CS8510-safe. The tests reflect this honestly: `MetadataSource.Subsumes` is unit-tested directly (Yes/No/Unknown), the guard's decline wiring is pinned by synthetic tests, and compiled canaries document that real subsuming/variance/guarded ladders are left as if-cascades while a disjoint `switch`-expression source still folds.

### Open decision for the maintainer
Given #3065 superseded the matcher, this PR's remaining content is the deferred oracle. Reasonable options:
1. **Close** as superseded (the oracle is not a currently-reachable bugfix).
2. **Keep** as defense-in-depth groundwork (this branch as-is).
3. **Repurpose** into the real follow-up #3065 anticipated: extend the matcher to fold the guarded/subpattern heterogeneous arms it deliberately declined, using this oracle to make that safe (net-new feature work).

**Not marking ready to merge.** If kept/repurposed, it needs fresh two-model adversarial review at the new head (the base and behavior changed materially from the previously reviewed `411f68f5`).

Validation: pass class 26/26; `Area=Pass` 1243/0.

richlander:
Superseded by #3102. #3082 landed the canonical type-disjointness oracle (`MetadataSource.AreProvablyDisjoint`) that this PR's `8d628546` subsumption oracle duplicated, and #3065 landed its inline work. #3102 reuses #3082's oracle to fold guarded direct/inline heterogeneous arms — no parallel abstraction. Closing this in favor of that clean rebuild.

Issue #3029, opened 2026-07-22
Inline dynamic stackalloc count spills


richlander:
Dynamic stackalloc now raises to valid, correct `stackalloc T[count]`, but the compiler's duplicated count spill remains:

```csharp
public static int ByteCount(int n)
{
    int V_1 = n;
    Span<byte> values = stackalloc byte[V_1];
    return values.Length;
}
```

The intended fully raised endpoint is:

```csharp
public static int ByteCount(int n)
{
    Span<byte> values = stackalloc byte[n];
    return values.Length;
}
```

`ExpressionInliningPass` runs before `StackAllocSpanPass`, when the spill still has two loads: one in the localloc byte-size expression and one in the `Span<T>` constructor length. `StackAllocSpanPass` removes the redundant byte-size expression, leaving one load, but the late inliner is deliberately stack-slot-only and does not revisit user locals.

Recover the compiler spill only with an exact ownership/evaluation-order proof. Cover pure and effectful count expressions, additional uses/stores, address-taking, non-adjacent stores, exception ordering, Debug/Release compiler shapes, and still-flat near misses.

This is the remaining source-altitude slice after #3016; it is not required for that PR's validity/correctness fix.

PR #3048, opened 2026-07-22
Inline dynamic stackalloc count spills

richlander:
- Fixes #3029
- Removes csc's duplicated dynamic stackalloc count spill when ownership and evaluation order are proven
- Card revision: `4caeedc9`

## Change

> Should we accept this change?

**Conclusion:** **PASS** — the pass now reaches the source-level dynamic `stackalloc` endpoint, including effectful counts and sequential Release-csc slot reuse, while exact ownership and evaluation-order gates retain explicit source locals and decline close negatives.

### Original source

```csharp
public static int ByteCount(int n)
{
    Span<byte> values = stackalloc byte[n];
    return values.Length;
}
```

### Before

```csharp
public static int ByteCount(int n)
{
    int V_1 = n;
    Span<byte> values = stackalloc byte[V_1];
    return values.Length;
}
```

- Valid: True
- Correct: True

The stackalloc raise was already correct, but the compiler-owned duplicated count spill remained visible.

### After

```csharp
public static int ByteCount(int n)
{
    Span<byte> values = stackalloc byte[n];
    return values.Length;
}
```

- Valid: True
- Correct: True

### Fully raised

The After decompilation is in the fully raised state.

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Product build | Pass (3 existing nullable warnings) | Pass (same 3 warnings) |
| `StackAllocSpanPassTests` | 56 passed | 69 passed |
| Decompiler fast suite | 3,237 passed | 3,262 passed |
| Compiler-backed fixtures | 2 Release cases passed | 14 Release + 14 Debug cases passed |
| Reduced fixture validity | Pass | Pass |
| Reduced fixture fidelity | Not applicable | Real render-changed method attempted; uncheckable due unrelated skeleton `CS0311` |
| Real witness | `RawNameToHashString` retains `int V_3` | emits `stackalloc short[charSpan.Length]` |

Final-head targeted traits:

- `Area=Pass`, excluding slow: 1,202 passed
- `Area=RoundTrip`: 312 passed
- `Area=Validity`: 99 passed
- `Area=Fidelity`, excluding slow: 27 passed
- `Area=Corpus`, excluding slow: 59 passed

Render-text A/B against merge base `f8f6a449`: 89,065 methods; 1 structural change, semantic `valid->valid`; 0 added, removed, or `valid->invalid` methods. The sole changed method is `Microsoft.CodeAnalysis.CSharp.Symbols.SourceNamedTypeSymbol::RawNameToHashString`.

[...]

PR #3049, opened 2026-07-23
Fold dynamic stackalloc count spill into the raised stackalloc (#3029)

richlander:
- Fixes/advances #3029
- Changes dynamic `stackalloc T[n]` output: the compiler's element-count spill is now folded into the count, so it renders `stackalloc T[n]` instead of `int V_1 = n; ... stackalloc T[V_1]`.
- Card revision: `fabba8f5`

## Change

**Conclusion:** **PASS** — the surviving single-use count spill is folded under the same ownership/evaluation-order proof the general inliner uses, reaching the intended fully-raised endpoint with no behavior change.

### Original source

```csharp
public static int ByteCount(int n)
{
    Span<byte> values = stackalloc byte[n];
    return values.Length;
}
```

### Before

```csharp
int V_1 = n;
Span<byte> values = stackalloc byte[V_1];
return values.Length;
```

- Valid: True
- Correct: True

Valid and correct, but not fully raised: the compiler's count spill (`V_1`) is still spelled.

### After

```csharp
Span<byte> values = stackalloc byte[n];
return values.Length;
```

- Valid: True
- Correct: True

### Fully raised

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `StackAllocSpanPassTests`: 56 passed | `StackAllocSpanPassTests`: 63 passed (7 new) |
| Focused tests | `UnsafeEmitterTests`: 40 passed | `UnsafeEmitterTests`: 40 passed (strengthened `DynamicStackAllocCompilerShapes` to assert `stackalloc T[n]` and no `V_1`) |
| Decompiler fast suite | 3244, 0 failed (env-only fixture miss when solution not fully built) | 3244, 0 failed |
| Reduced fixture fidelity | see note | see note |

Note: the decompiled `stackalloc T[n]` recompiles to the identical spill IL the compiler emits, so compile-back fidelity is preserved by construction. The full slow corpus/fidelity gate is running separately on a heavily I/O-constrained box (single compile-back tests taking 20+ min); corpus numbers will be attached in a follow-up comment.

## How it works

The dynamic `stackalloc T[n]` lowering spills the count into a local read twice — once inside the `localloc` byte size, once as the `Span<T>` constructor length. The early full inliner sees a multi-use local; the late inliner is stack-slot-only. `StackAllocSpanPass` drops the redundant byte-size read during the raise, leaving the spill single-use. This PR then folds it in-place, only when:

- the local is referenced nowhere but one block-level store and this one count load (no extra use/store, address-of, or designation binding);
- the store's declared type matches the stored value (no dropped widening/narrowing witness);
- the store is the count statement's immediate in-block predecessor (nothing observable between them, exception ordering preserved); and
- the count is either that statement's first-evaluated leaf (order preserved verbatim) or a provably pure value.

The count load is, by construction, the just-raised `StackAllocArray`'s own operand — never an increment lvalue — so this cannot mint an invalid `1++` the way an unrestricted late inline of a user local could.

Adversarial review (two models) and the corpus quick gate are pending; this PR is not yet marked ready.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/StackAllocSpanPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/StackAllocSpanPassTests.cs
@@ -1474,6 +1474,36 @@ public class StackAllocSpanPassTests
         function.CheckInvariant();
     }
 
+    [Fact]
+    public void DynamicCountSpillIntoWhileLoopCondition_DoesNotFold()
+    {
+        // The spill `V_1 = Effect()` runs once before the loop, but the
+        // stackalloc-span is consumed in the WhileLoop condition, which
+        // re-evaluates every iteration. After the raise the count load is the
+        // condition's first-evaluated leaf, yet folding the effectful spill value
+        // there would run Effect() on every iteration instead of once. A loop
+        // header is never a single-evaluation first-leaf position.
+        var value = new Call(new MethodRef(Holder, "Effect", Int32, [], HasThis: false), isVirtual: false, []);
+        var spill = new StoreLocal(1, Int32, value);
+        var span = TypeRef.GenericInstance(TypeRef.CoreLib("System", "Span`1"), [Byte]);
+        var newObject = StackAllocSpanConstructor(
+            TypeRef.CoreLib("System", "Span`1"),
+            new StackAllocate(new LoadLocal(1, Int32)),
+            new LoadLocal(1, Int32),
+            Byte);
+        var length = new Call(new MethodRef(span, "get_Length", Int32, [], HasThis: true), isVirtual: false, [newObject]);
+        var condition = new Comparison(ComparisonKind.GreaterThan, isUnsigned: false, length, new Constant(0, Int32));
+        var loop = new WhileLoop(condition, new Block(0));
+        var function = BuildBody(spill, loop, new Return(new Constant(0, Int32)));
+
+        new StackAllocSpanPass().Run(function, PassContext.None);
+
+        var raised = Assert.Single(function.Descendants.OfType<StackAllocArray>());
+        Assert.IsType<LoadLocal>(raised.Count); // count left spilled, not hoisted into the loop condition
+        Assert.Contains(function.Descendants.OfType<StoreLocal>(), s => s.Index == 1);
+        function.CheckInvariant();
+    }
+
     /// <summary>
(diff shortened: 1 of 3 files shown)


richlander:
Closing as a duplicate: issue #3029 was already fixed and merged concurrently by #3048 (`Inline dynamic stackalloc count spills`, merge commit 13dd76e, merged 2026-07-23T02:58Z), which lands a different but sound mechanism (`TryResolveOwnedCountSpill` during the raise) with full corpus A/B evidence (89,065 methods, 0 valid→invalid). This PR is redundant, so I'm withdrawing it rather than pushing a second parallel mechanism to `main`.

**Adversarial review reconciliation.** The two fixed-head reviews of this PR (head `f5d6ceed`) both RESOLVED, after surfacing three real evaluation-order soundness hazards that were fixed here:

- **GPT r1 — stale read:** a pure moved count folded past an earlier sibling mutation (`Consume(V0++, stackalloc byte[V1])`), reading a stale post-increment count.
- **Gemini r1 — loop re-eval:** a count load that is the first-evaluated leaf of a `while` condition re-evaluates every iteration, duplicating an effect if folded.
- **GPT r2 — `??=` conditional:** `NullCoalescingAssignment` stores its conditional fallback as child 0, so an unconditional effect could be folded into a conditionally-executed position.

**Verification that the merged #3048 is sound against all three.** I ported the three hazards to implementation-agnostic probe tests (construct the compiler count-spill IR shape, run `StackAllocSpanPass`, assert on output) against `origin/main` and confirmed:

- Positive control `Span<byte> s = stackalloc byte[n]` — **folds** (spill removed, count becomes the stored expression). ✅
- Earlier-effectful-operand (GPT r1) — **declines**, spill retained. ✅
- `while`-condition (Gemini r1) — **declines**, spill retained. ✅
- `??=` fallback (GPT r2) — **declines**, spill retained. ✅

Main's `GetHeldExpression` is a statement-type allowlist that returns null for `WhileLoop`/`NullCoalescingAssignment` (closing the loop and `??=` hazards), and `ReachesAsOnlyPrecedingEffect` requires every earlier-evaluated operand to be side-effect-free (closing the stale-read hazard). No product change needed; #3029 is fully resolved on `main`.

Thanks to both reviewers (GPT, Gemini) — the hazards they found informed the verification above.

PR #3050, opened 2026-07-23
Drop redundant (dynamic) cast on by-ref dynamic receivers (#3035)

### LINKED ISSUE #3035: Decompiler keeps redundant (dynamic) cast on ref/in/out dynamic receivers (follow-up to #3032)
## Summary

Follow-up from PR #3032 (issue #2984). The decompiler drops the redundant `((dynamic)x).Member` cast when the receiver's recovered static type is `dynamic`. For **by-ref** dynamic parameters (`ref dynamic` / `in dynamic` / `out dynamic`) the cast is currently **kept** — conservative, valid, and correct output, but a missed drop. This issue tracks recovering the drop for that case.

## Root cause (two layers, both must change)

1. **Decode offset.** `DynamicReader.IsTopLevelDynamic(flags)` checks `flags[0] == 1`. For a by-ref parameter the `ByRef` modifier consumes the leading flag slot, so the *element* type's dynamic-ness sits at `flags[1]`, not `flags[0]`. `IsTopLevelDynamic` therefore returns `false` for `ref/in/out dynamic`, and `MethodImporter` never marks the parameter `IsDynamic`.

   The `IsTopLevelDynamic` contract is honestly "outermost type position". For by-ref params the outermost position *is* the ByRef slot, so the primitive is not wrong — the caller must strip the ByRef slot to inspect the element. Decide whether to tighten the caller (`MethodImporter`) or add a by-ref-aware overload/primitive.

2. **Printer receiver shape.** Even if the decode were fixed, a by-ref receiver renders as a **deref** IR node, not a bare `LoadArgument`/`LoadField`. `CSharpPrinter.IsDynamicTypedReceiver` keys only on `LoadArgument{IsDynamic:true}` / `LoadField{Field.IsDynamic:true}`, so it would still not fire. The printer must additionally recognize a deref of a dynamic by-ref parameter.

Because layer 2 gates the output, layer 1 alone has **zero output effect** today — which is why PR #3032 left this as a non-blocking follow-up rather than a defect.

## Current behavior (verified on PR #3032 head)

```csharp
// ref dynamic  →  cast KEPT (conservative, valid)
static object M(ref dynamic value) => ((dynamic)(value)).Length;
// in dynamic   →  cast KEPT
static object N(in dynamic value)  => ((dynamic)(value)).Length;
// dynamic      →  cast DROPPED (the PR #3032 behavior)
static object P(dynamic value)     => value.Length;
```

## Proposed fix

- **Decode:** in `MethodImporter`, when a parameter is by-ref, skip the leading ByRef flag before testing element dynamic-ness (or add a by-ref-aware decode path); thread `IsDynamic` onto the by-ref `Parameter`.
- **Printer:** extend `IsDynamicTypedReceiver` to recognize a deref of a dynamic by-ref parameter and drop the cast there too.
- **Tests:** add fixtures for `ref/in/out dynamic` receivers; assert the cast is dropped and the recompiled body is opcode-exact / valid.

## Boundaries / non-goals

- Conservative-correct output is already shipped by #3032; this is purely a fidelity/quality improvement (fewer redundant casts), not a correctness fix.
- Keep the same conservative "keep the cast unless the receiver's declared static type is provably `dynamic`" invariant.

Refs: #2984, #3032.

richlander:
- Fixes #3035 (advances #2984 / #3032)
- Drops the redundant `(dynamic)` cast on `ref`/`in`/`out dynamic` member-access receivers, matching plain `dynamic` params
- Card revision: `23a54bfa`

## Change

**Conclusion:** **PASS** — the by-ref `dynamic` receiver now spells the fully-raised `value.Length` (the cast was provably redundant), and the `ref object` negative control still keeps its cast.

### Original source

```csharp
public object DynamicGetLengthByRef(ref dynamic value)
{
    return value.Length;
}
```

### Before

```csharp
public object DynamicGetLengthByRef(ref dynamic value) => ((dynamic)(value)).Length;
```

- Valid: True
- Correct: True

Valid and behavior-preserving, but lower quality: the by-ref `dynamic` element's static type is already `dynamic`, so the `(dynamic)` cast (and its parentheses) are pure noise the plain-`dynamic` path already drops.

### After

```csharp
public object DynamicGetLengthByRef(ref dynamic value) => value.Length;
```

- Valid: True
- Correct: True

### Fully raised

## Root cause (two layers)

1. **Metadata decode.** A by-ref parameter carries the ByRef modifier at `DynamicAttribute` transform-flag index 0, so `ref dynamic` emits `{ false, true }` — the element dynamic-ness sits at index **1**. `DynamicReader.IsTopLevelDynamic` reads index 0, so the parameter was never marked `IsDynamic`. Added `DynamicReader.IsByRefElementDynamic` (index 1); `MethodImporter` selects it when `decoded.ParameterTypes[index].Kind == TypeRefKind.ByRef`.
2. **Printer.** A by-ref receiver reads through `LoadIndirect { Address: LoadArgument { IsDynamic } }` (deref), not a bare `LoadArgument`, so `IsDynamicTypedReceiver` never fired. Added that arm and render the dropped-cast receiver via the deref's underlying place so output is `value.Length`, not `(value).Length`. A follow-up (review finding) keeps the `ref` modifier on by-ref `dynamic` **local-function** parameter declarations (`ref dynamic v`) to avoid CS1615.

A `ref object` receiver has no `DynamicAttribute`, so its access is dynamic only through the explicit source cast and correctly keeps `((dynamic)value).Length`.

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `DynamicCallSitePassTests`+`LadderRung9GateTests`: green (no by-ref cases) | 140 passed (adds ref/in/out + local-fn + negative controls) |
| Metadata unit | `DynamicTypeViewTests`: green | `IsByRefElementDynamic` theory added; 56 passed |
| Decompiler RoundTrip area | 312, 0 failed | 316, 0 failed |
| Reduced fixture validity | Pass (`Rung9Fixture_HasNoInvalidFull`) | Pass |
| Reduced fixture fidelity | Pass | Pass — cast-drop on an already-dynamic receiver recompiles to identical dynamic call-site IL |
| Corpus + Fidelity areas | 1 env red (`NoNewFidelityDiffsBeyondKnownDocket`) | 1 env red — same test, same lambda-cache field ordinal shift, identical opcodes, no new diff-set entries (CI-green) |

The single Corpus/Fidelity red is the pre-existing environmental `NoNewFidelityDiffsBeyondKnownDocket` (a compiler-synthesized lambda-cache field ordinal shift in `CfgSampleClass`, identical opcode streams); it is CI-green and orthogonal to this change.

## Decompiler quality

**Conclusion:** **PASS** — no corpus raise-rate impact. The change only affects the C# spelling of by-ref `dynamic` member-access receivers (redundant-cast drop); it adds no pass, alters no structuring, and leaves recompiled IL identical. Verified by the RoundTrip area (316/316) and the fixture validity/fidelity gates rather than a corpus rate delta.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class "ILInspector.Decompiler.Tests.DynamicCallSitePassTests" -class "ILInspector.Decompiler.Tests.LadderRung9GateTests"
dotnet run --project tests/ILInspector.Metadata.Tests -c Release --no-build -- -class "ILInspector.Metadata.Tests.DynamicTypeViewTests"
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -trait "Area=RoundTrip"
```

Adversarial review evidence is in a separate PR comment.

### DIFF
--- a/src/ILInspector.Decompiler.Fixtures.LadderRung9/LadderRung9.cs
+++ b/src/ILInspector.Decompiler.Fixtures.LadderRung9/LadderRung9.cs
@@ -105,6 +105,36 @@ public class DynamicAndExpressionTrees
         return input;
     }
 
+    // By-ref `dynamic` receivers (#3035). csc emits [DynamicAttribute({ false,
+    // true })] on these parameters — the leading `false` is the ByRef modifier
+    // and the element dynamic-ness sits at flags[1] — and the receiver renders as
+    // a deref of the by-ref parameter rather than a bare argument load. The
+    // redundant `(dynamic)` cast must still drop, since the element static type
+    // is provably `dynamic`.
+    public object DynamicGetLengthByRef(ref dynamic value)
+    {
+        return value.Length;
+    }
+
+    public object DynamicGetLengthIn(in dynamic value)
+    {
+        return value.Length;
+    }
+
+    public object DynamicGetLengthOut(out dynamic value)
+    {
+        value = "abc";
+        return value.Length;
+    }
+
+    // Negative control: a by-ref `object` receiver (not dynamic). The member
+    // access only becomes dynamic through the explicit `(dynamic)` cast, so the
+    // drop must NOT fire — the element static type really is `object`.
+    public object DynamicGetLengthByRefObject(ref object value)
+    {
+        return ((dynamic)value).Length;
+    }
+
     public Expression<Func<int, int>> SimpleExpressionTree()
     {
         return x => x + 1;
--- a/src/ILInspector.Decompiler.Fixtures.LadderRung9/LadderRung9MemberContexts.cs
+++ b/src/ILInspector.Decompiler.Fixtures.LadderRung9/LadderRung9MemberContexts.cs
@@ -60,6 +60,17 @@ public class DynamicMemberContexts
         return Get(input);
     }
 
+    // Nested local function whose OWN parameter is a by-ref `dynamic`
+    // (`ref dynamic v`). The raised body drops the redundant cast to `v.Length`,
+    // and the printer-owned local-function declaration must keep the by-ref
+    // modifier — spelling `ref dynamic v`, not bare `dynamic v` — otherwise the
+    // `Get(ref value)` call site is CS1615 (#3035).
+    public object InLocalFunctionByRefOwnParam(dynamic value)
+    {
+        static object Get(ref dynamic v) => v.Length;
+        return Get(ref value);
+    }
+
     // Iterator state-machine body: csc lowers this to a MoveNext method whose
     // declaring type is the generated iterator state machine, while the
     // GetMember context remains the authored enclosing type (same bridge as
--- a/src/ILInspector.Decompiler.Tests/DynamicCallSitePassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/DynamicCallSitePassTests.cs
@@ -226,6 +226,34 @@ public class DynamicCallSitePassTests
         Assert.DoesNotContain("((dynamic)value)", output);
     }
 
+    [Theory]
+    [InlineData("DynamicGetLengthByRef")] // ref dynamic
+    [InlineData("DynamicGetLengthIn")]    // in dynamic
+    [InlineData("DynamicGetLengthOut")]   // out dynamic
+    public void ByRefDynamicReceiver_DropsRedundantCast(string method)
+    {
+        // A by-ref `dynamic` parameter reads its receiver through a deref of the
+        // by-ref argument. The [DynamicAttribute] transform flags carry a leading
+        // ByRef modifier, so the element dynamic-ness sits at flags[1]; once the
+        // parameter is recognized as dynamic, the referenced element's static type
+        // is provably `dynamic`, so the redundant `(dynamic)` cast is dropped and
+        // the deref renders as the bare place `value.Length` (#3035).
+        var output = RaiseAndPrint(LoadCanonicalFunction(method));
+        Assert.Contains("value.Length", output);
+        Assert.DoesNotContain("(dynamic)", output);
+    }
+
+    [Fact]
+    public void ByRefObjectReceiver_KeepsCast()
+    {
+        // Negative control: `ref object value` has no DynamicAttribute, so the
+        // element static type really is `object`. The `(dynamic)` cast is the only
+        // thing making the member access dynamic, so it must be kept (#3035).
+        var output = RaiseAndPrint(LoadCanonicalFunction("DynamicGetLengthByRefObject"));
+        Assert.Contains("(dynamic)", output);
+        Assert.Contains(".Length", output);
+    }
+
     [Fact]
     public void KeywordMemberName_RaisesAndKeepsRawName()
     {
--- a/src/ILInspector.Decompiler.Tests/DynamicCallSitePassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/DynamicCallSitePassTests.cs
@@ -1511,6 +1511,21 @@ public class DynamicCallSitePassTests
     }
 
     [Fact]
+    public void NestedContext_LocalFunctionByRefOwnParam_SpellsRefDynamicDeclaration()
+    {
+        // The local function has its OWN `ref dynamic` parameter. The body drops
+        // the redundant cast (`v.Length`), and the printer-owned local-function
+        // declaration must keep the by-ref modifier — spelling `ref dynamic v`,
+        // not bare `dynamic v` (which drops `ref` and makes the `Get(ref value)`
+        // call site CS1615) (#3035).
+        var output = RaiseMemberContext("InLocalFunctionByRefOwnParam");
+        Assert.Contains("ref dynamic v", output);
+        Assert.Contains("v.Length", output);
+        Assert.DoesNotContain("(dynamic v)", output);
+        Assert.DoesNotContain("((dynamic)v)", output);
+    }
+
+    [Fact]
     public void NestedContext_LambdaDisplayClass_Raises()
     {
         // csc lowers the lambda body into a display-class method whose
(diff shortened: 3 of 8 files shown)


richlander:

Two-model review per repo policy: **GPT-5.6** and **Gemini 3.1 Pro** (own model Claude Opus excluded). Reviewers ran in isolated detached worktrees at exact heads; scratch confined to `/tmp/`, no git mutation.

### Round 1 (head `a369bf9d`)
- **GPT-5.6 — 1 finding (Medium, valid):** once the decode fix marks a `ref dynamic` parameter `IsDynamic`, `ParameterTypeText` spelled it as bare `dynamic` in the printer-owned **local-function** declaration, dropping the by-ref modifier. `static object Get(dynamic v) => v.Length;` invoked as `Get(ref value)` fails **CS1615**. Confirmed and reproduced.
- **Gemini 3.1 Pro — clean** on the decode/printer core (verified DynamicAttribute flag layout for `ref dynamic`/`in`/`out`/`ref dynamic[]`/`ref Func<dynamic>`/`dynamic*`, `SequenceNumber-1` index alignment, the `LoadIndirect{LoadArgument{IsDynamic}}` arm, and negative controls). Did not cover the local-function header path.

### Fix (head `23a54bfa`)
`ParameterTypeText` now keeps the by-ref modifier for a dynamic element — spelling `ref dynamic` (only the referenced element is `dynamic`), leaving plain top-level `dynamic` parameters unchanged. Added the `InLocalFunctionByRefOwnParam` fixture and a regression test asserting the declaration keeps `ref dynamic v`. Verified output: `Get(ref value)` + `static object Get(ref dynamic v) => v.Length;` (CS1615 resolved).

### Round 2 — re-review of fixed head `23a54bfa` (both clean)
- **GPT-5.6:** "No significant issues found." Verified the repro now emits `ref dynamic v` and recompiles; `in`/`out dynamic` render as `ref dynamic`, matching pre-existing by-ref normalization with no new invalidity; non-dynamic params unchanged.
- **Gemini 3.1 Pro:** "no high-confidence defects or soundness regressions." Verified CS1615 fix + call-site alignment, `in`/`out` over-reach matches pre-existing non-dynamic by-ref local-fn behavior, and pointer/array edge cases (`dynamic*`→Pointer, `ref dynamic[]`→flags `{false,false,true}`) correctly fall back.

### Validation (local, net11 daily)
- DynamicCallSitePassTests + LadderRung9GateTests: green (incl. new by-ref + local-fn tests)
- Metadata DynamicTypeViewTests: green
- `Area=RoundTrip`: 316/316
- `Area=Corpus` + `Area=Fidelity`: only the pre-existing environmental red `NoNewFidelityDiffsBeyondKnownDocket` (lambda-cache field ordinal shift in `CfgSampleClass`, identical opcode streams; CI-green, orthogonal — no new diff-set entries)

Ready to merge

richlander:
## Review reconciliation — head `23a54bfa`

**Verdict: no blocking correctness, soundness, or security bugs.** Two fixed-head adversarial reviews from different model families (Gemini 3.1 Pro and GPT-5.6) both came back clean, matching my own pass. CI is green (`build-net10`, `test (ubuntu-26.04)`, `csharp-diff-smoke`, `changes` all pass; `il-diff-smoke`/`markdownlint`/`pack` path-skipped). Merge state CLEAN.

### Attack points verified sound by both reviewers

1. **Flag-layout soundness** — `IsByRefElementDynamic` → `flags[1] == 1` is correct for the ByRef-consumes-index-0 layout. `ref/in/out dynamic` decode as `{false, true}`; negative controls decline correctly: `ref object` (`{false,false}`/absent), `ref dynamic[]` and `ref List<dynamic>` (`{false,false,true}` → index 1 is `false`). GPT confirmed the raw SRM `DynamicAttribute` blobs (`00-01` for ref/in/out; none for `ref object`). A forged `[Dynamic({false,true})]` on a `ref int` is honest decompilation of a lying assembly — no crash/vuln.
2. **Importer guard** — `SequenceNumber - 1` correctly excludes the return type (sequence 0); the `index < ParameterTypes.Length` bounds check is safe; index 1 is selected only for an outer `TypeRefKind.ByRef`.
3. **Cast-drop printer** — the new `LoadIndirect { Address: LoadArgument { IsDynamic: true } }` arm only fires when the importer proved the element dynamic, so `ref object` retains its cast; `DerefLoad` renders the bare place (`value.Member`), valid C# (member access has top precedence, no parens needed).
4. **Parameter spelling** — `ParameterTypeText` emitting `"ref dynamic"` for `in`/`out dynamic` **local-function** params is a **pre-existing** limitation, not a regression: `TypeRef.ToDisplayString` already collapses every `TypeRefKind.ByRef` to `ref` (`TypeRef.cs:492`), so before this PR an `in dynamic` local-function param already rendered `ref object`. The PR strictly improves `object`→`dynamic` without newly losing `in`/`out`. The `ref dynamic` local-function case (real CS1615) was caught by the earlier adversarial pass and fixed in commit `23a54bfa` with the `InLocalFunctionByRefOwnParam` fixture + regression test.
5. **Blast radius** — `Parameter.IsDynamic` propagates only into `LoadArgument.IsDynamic`; the sole consumers are `IsDynamicTypedReceiver` (cast drop) and `ParameterTypeText` (local-function decl spelling). No overload-resolution, generic-inference, coercion, or lifetime path reads it.

### Tests (run locally at head `23a54bfa`, both reviewers + me)

| Class | Result |
| --- | --- |
| `DynamicCallSitePassTests` | 116/116 pass |
| `LadderRung9GateTests` | 24/24 pass |
| `DynamicTypeViewTests` (metadata) | 56/56 pass |

The by-ref tests are non-vacuous: `ByRefDynamicReceiver_DropsRedundantCast` (ref/in/out) asserts `value.Length` present **and** `(dynamic)` absent; `ByRefObjectReceiver_KeepsCast` asserts the cast is kept for the genuine `ref object` control.

### Non-blocking follow-up (optional)

`in`/`out dynamic` local-function parameters spell as `ref dynamic` due to the pre-existing `TypeRef.ByRef → "ref"` collapse. Threading a `RefKind` through the local-function declaration printer would fix in/out fidelity for **all** by-ref local-function params (dynamic and not) — out of scope here.

Both fixed-head reviews are clean and CI is green.

Ready to merge

PR #3053, opened 2026-07-23
Recover int comparison-predicate expression-tree lambdas (#2864)

richlander:
- Advances #2864
- Recovers a single `int` comparison-predicate expression-tree factory graph
  (a `Func<…,bool>` body such as `x => x > y`) back to a source lambda
- Card revision: `d8ad8fb3`

## Change

**Conclusion:** **PASS** — the delegate's `TResult` selects the body raiser
(`int` → arithmetic, `bool` → a single comparison); the comparison raiser matches
only the exact 2-arg `Expression.Equal/NotEqual/LessThan/LessThanOrEqual/
GreaterThan/GreaterThanOrEqual` factory the compiler emits for a source `int`
comparison, raises both operands through the proven `int` arithmetic subset, and
requires ≥1 parameter-dependent operand. The recovered `l OP r` recompiles to
exactly that same 2-arg factory node, so the round-trip tree is identical.

### Original source

```csharp
public static Expression<Func<int, int, bool>> GreaterThanComparison()
    => (x, y) => x > y;
```

### Before

```csharp
public static System.Linq.Expressions.Expression<System.Func<int, int, bool>> GreaterThanComparison()
{
    ParameterExpression V_0 = Expression.Parameter(typeof(int), "x");
    ParameterExpression V_1 = Expression.Parameter(typeof(int), "y");
    ParameterExpression[] S_256 = new ParameterExpression[2];
    BinaryExpression S_257 = Expression.GreaterThan(V_0, V_1);
    S_256[0] = V_0;
    ParameterExpression[] S_258 = S_256;
    S_258[1] = V_1;
    return Expression.Lambda<Func<int, int, bool>>(S_257, S_258);
}
```

- Valid: True
- Correct: True

Honest factory-call form: valid and behavior-preserving, but not raised.

### After

```csharp
public static System.Linq.Expressions.Expression<System.Func<int, int, bool>> GreaterThanComparison() => (x, y) => x > y;
```

- Valid: True
- Correct: True

The recovered lambda recompiles to exactly the matched 2-arg
`Expression.GreaterThan(Parameter, Parameter)` node (`Method == null`,
`!IsLiftedToNull`), so the returned expression tree is identical.

### Fully raised

For the supported single-`int`-comparison predicate slice, the After
decompilation is in the fully raised state. Issue #2864 remains open for the
non-`int`, captured/member, boolean-composition (`AndAlso`/`OrElse`), and broader
expression-node slices.

## Why no `unchecked` wrapper on a pure comparison

A pure comparison body carries no overflow-context form — there is no
"checked comparison" factory — so `x > y` needs no `unchecked(...)`. When a
comparison operand is itself arithmetic (`(x, y) => x + 1 > y`), the operand is
raised through the same arithmetic subset and keeps its load-bearing
`unchecked(...)` spelling: the recovered text is `(x, y) => unchecked(x + 1) > y`,
pinning `Expression.Add` (not `AddChecked`) under a checked-default consuming
project, exactly as in the arithmetic slice (#2926).

## Boundary (declines, stays factory calls / `Partial`)

- The lifted/method-carrying **4-arg** comparison overload (nullable-lifted or
  user-defined-operator) — different argument count, bails.
- **Boolean composition** roots (`x > 0 && x < 10` → `AndAlso`), method-call,
  member, and convert operands.
- **Non-`int`** operands (`Func<long,long,bool>` comparison) — rejected by the
  int-parameter guard.
- **Constant-only** comparisons (`2 > 3`) — fold to a single `Constant` on
  recompile, so ≥1 parameter-dependent operand is required.

## Evidence

| --- | --- | --- |
| Focused expression-tree tests | n/a (fixtures new) | `ExpressionTreeFidelityTests` + `ExpressionTreeLambdaTests`: pass |
| Decompiler fast suite | 3,298, 0 failed | 3,298, 0 failed |
| Dynamic-site inventory | n/a (fingerprint updated) | Pass (32 files / 40 sites) |
| Compile-back / tree oracle | n/a | Pass (2-arg node identity; constant-only folds to `Constant`) |
| Close negatives | — | long comparison, `AndAlso` composition, manual constant-only comparison, and the existing `Member` predicate all decline |

Before/After above are verbatim `dotnet-inspect member … -S "Decompiled Source"`
output: base tool vs. head tool over the same head-compiled fixture IL.

## Decompiler quality

**Conclusion:** **PASS** — recovery is limited to a compiler-shaped, fully-owned
`int` comparison-predicate graph. Unsupported graphs retain their explicit
factory calls or existing honest `Partial` output; the general corpus does not
contain source expression-tree factory runs of this shape, so no aggregate
movement is expected.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.ExpressionTreeFidelityTests
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.ExpressionTreeLambdaTests
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -556,6 +556,22 @@ public class CfgSampleClass
             x);
     }
 
+    // Near-miss: a constant-only comparison root (`GreaterThan(2, 3)`) with an
+    // unreferenced parameter. The C# compiler constant-folds `2 > 3` to a single
+    // Constant(false) when the recovered lambda is recompiled to an expression tree
+    // (verified: `x => 2 > 3` lowers to a Constant node, not GreaterThan), so
+    // recovering `x => 2 > 3` would rebuild a different tree. The comparison
+    // constant-fold guard declines it, keeping the honest factory calls.
+    public static System.Linq.Expressions.Expression<System.Func<int, bool>> ManualConstantOnlyComparisonFactory()
+    {
+        var x = System.Linq.Expressions.Expression.Parameter(typeof(int), "x");
+        return System.Linq.Expressions.Expression.Lambda<System.Func<int, bool>>(
+            System.Linq.Expressions.Expression.GreaterThan(
+                System.Linq.Expressions.Expression.Constant(2, typeof(int)),
+                System.Linq.Expressions.Expression.Constant(3, typeof(int))),
+            x);
+    }
+
     // Capturing: `n` is hoisted into a <>c__DisplayClass, so the delegate targets
     // an instance method on that class. LambdaRaisingPass substitutes the body's
     // `this.n` read with the captured value and recovers `x => x + n`.
--- a/src/ILInspector.Decompiler.Tests/DynamicCompilationSiteInventoryTests.cs
+++ b/src/ILInspector.Decompiler.Tests/DynamicCompilationSiteInventoryTests.cs
@@ -59,7 +59,7 @@ public sealed class DynamicCompilationSiteInventoryTests
             ["CrossAssemblyMethodFactsTests.cs"] = (1, "Cross-assembly seam: constructs referencing compilations to test cross-assembly facts."),
 
             // Product-output validity under varying compilation options.
-            ["ExpressionTreeLambdaTests.cs"] = (2, "Product-output validity: compiles synthesized expression-tree source and checks diagnostics/emit under varying compilation options (e.g. overflow checks)."),
+            ["ExpressionTreeLambdaTests.cs"] = (3, "Product-output validity + compile-back oracle: compiles synthesized expression-tree source under varying compilation options (overflow checks) and recompiles recovered arithmetic/comparison lambdas to assert their expression-tree node identity."),
 
             // Bespoke positive-source gates sharing one helper, each asserting
             // many per-fact expectations over a runtime-varying compilation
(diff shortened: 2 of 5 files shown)


richlander:

Two different models (both non-Opus, per the reviewer roster) reviewed the exact pushed head `d8ad8fb3e16844f4b112ccf1fb1653a205f012d2`, each in its own isolated read-only worktree with scratch confined to `/tmp/` and no git mutations.

| Reviewer | Model | Head reviewed | Verdict |
| --- | --- | --- | --- |
| Reviewer A | GPT-5.6 (`gpt-5.6-sol`) | `d8ad8fb3` | **Clean** |
| Reviewer B | Gemini 3.1 Pro (`gemini-3.1-pro-preview`) | `d8ad8fb3` | **Clean** |

**Findings:** none blocking from either reviewer. No follow-up commits were required, so the fixed head equals the reviewed head (`d8ad8fb3`).

**Real-run evidence gathered by the reviewers:**

- Product build (`src/dotnet-inspect -c Release`): clean, 0 warnings / 0 errors (both).
- `ExpressionTreeFidelityTests`: 10/10 pass. `ExpressionTreeLambdaTests`: 32/32 pass (after the full-solution build that produces the lookalike-spoof fixture DLL the `LookalikeExpressionAssembly` test needs). Fast decompiler suite (`-trait- "Speed=Slow"`): 3298/3298, 0 failed.
- Reviewer B additionally wrote throwaway Roslyn probes under `/tmp/` and traced operand type inference through `RaiseBody`.

**Attack points, both reviewers concurring the change is safe:**

1. **Signedness** — `IsIntParameterFunc` + the operand `IsInt` checks confine operands to `System.Int32`, which binds to a signed C# comparison, exactly reproducing `isUnsigned: false`.
2. **Lifted / method overload** — `TryComparisonKind` requires exactly 2 arguments; nullable-lifted / user-defined-operator comparisons emit the 4-arg `GreaterThan(l, r, liftToNull, method)` overload and are rejected.
3. **Constant-fold guard** — `RaiseComparisonBody` requires ≥1 parameter-dependent operand, so a constant-only comparison (Roslyn folds it to a single `Constant`) declines and never breaks tree identity.
4. **Boolean composition** — `&&`/`||` root as `AndAlso`/`OrElse`, unmapped by `TryComparisonKind` (`(ComparisonKind)(-1)` sentinel), so it aborts; a boolean-comparison operand is rejected by `RaiseBody` (arithmetic-only).
5. **Non-int leakage** — every delegate parameter is int-checked at entry and both reconstructed operand `ResultType`s are re-checked; `Func<long,long,bool>` and convert operands decline.
6. **Ownership guards** — both result kinds pass through the same `ResolveConstruction`/`plan.Body`, preserving single-source parameter identity, array/slot ownership, unique spellable names, and return-sink validation.
7. **Printer** — `IsExpressionTree = true` prints the body under a checked context so a comparison over arithmetic renders exactly `return (x, y) => unchecked(x + 1) > y;`, pinning `Expression.Add` (never `AddChecked`); a pure comparison is unaffected.
8. **Sentinel / nullability** — `ComparisonKind` has only non-negative values, so `(ComparisonKind)(-1)` + `(int)mapped < 0` is an unambiguous not-found sentinel; `IsInt(TypeRef?)` null-checks before comparing.

richlander:
Ready to merge

Issue #3052, opened 2026-07-23
Harness: use --repo for authored-corpus (own libs, harvest/refresh, drift) — keep benchmark hermetic


richlander:
## Summary

Use the new `--repo` capability (#3040) to improve the decompiler harness's authored-source corpus workflow **where the network already happens** — most obviously our **own** corpus (dotnet-inspect libraries), plus corpus **harvest/refresh** and a new **drift-detection** check — while **keeping the benchmark oracle hermetic** (it continues to read the vendored, checksum-verified `AuthoredBody` snapshots). This is deliberately *not* a wholesale migration off the vendored corpus.

The clearest win: when the corpus includes **our own** libraries, their SourceLink commits live in *this* repo, which we already have on disk. Pointing `--repo` at the local checkout reads the authored blob locally and **skips the remote source fetch entirely** — the PDB checksum still arbitrates, so a mismatched working tree self-rejects and falls back to the network. Zero network, always in sync with the code under test.

## Decision and rationale

We considered fully retiring the `vendor/authored-source-corpus` orphan branch and re-acquiring source on demand via `AuthoredSourceAcquisition.AcquireMemberAsync(repositoryPaths)`. We are **not** doing that, because:

- The benchmark oracle's most valuable property is **hermeticity/reproducibility**. The vendored snapshot is a pinned, byte-exact, offline, content-addressed cache — immune to upstream force-push, repo deletion, history rewrite, and GitHub outages. Re-acquisition would make every provisioning depend on the pinned commit still existing upstream.
- To stay offline (a hard requirement) a "live" model must still persist the same source bytes in a pre-fetched cache — so it mostly **relocates** the vendored bytes into a *less* durable store rather than eliminating them.
- The harvest step **already** exercises the product path (`AcquireMemberAsync`); the benchmark's `ReturnToSenderSourceIndex.FromMembers` is a snapshot **reader**, not a second implementation of acquisition logic. So the "harness boundary" argument for a full migration is weak.

The genuine `--repo` wins are for our own libraries and at harvest/drift detection, so scope the work there.

## Scope A — own-corpus (self/dogfood): skip the remote source fetch

[...]

PR #3056, opened 2026-07-23
Harness: --repo for authored-corpus harvest (skip remote source fetch for own libs) [#3052]

richlander:
## What

Adds a repeatable `--repo <path>` option to the decompiler harness's authored-corpus **harvest** flow (`--harvest-authored-corpus` / `--harvest-evil-corpus`). It threads the merged `AuthoredSourceAcquisition.AcquireMemberAsync(repositoryPaths)` capability (#3040) into `AuthoredSourceHarvest`, so harvest can read a target's authored source from a local git clone — arbitrated by the same PDB checksum, falling back to the network on any mismatch or miss.

Implements **Scope A/B of #3052**. The benchmark, the vendored `vendor/authored-source-corpus` snapshot, and `eng/restore-authored-source-corpus.sh` are unchanged and stay hermetic (Scope C — drift detection — is not in this PR).

## Why

- **Primary win (own libraries):** for dotnet-inspect's own libraries the SourceLink commit lives in this checkout, so pointing `--repo` at it **skips the remote SourceLink fetch entirely** and always matches the code under test.
- **Secondary:** unlocks private, large, or offline third-party source repositories for harvest.

## Changes

- `tools/DecompilerHarness/AuthoredSourceHarvest.cs` — `Run`/`RunAsync`/`TryHarvestAsync` take an optional `IReadOnlyList<string>? repositoryPaths`, passed as the 6th positional arg to `AcquireMemberAsync`.
- `tools/DecompilerHarness/Program.cs` — new `sourceRepositories` list + `--repo` case (mirrors the existing repeatable `--package` option); threaded into both harvest calls; usage text.
- `tools/DecompilerHarness/README.md` — documents `--repo` in the harvest section.

## Evidence

- `dotnet build tools/DecompilerHarness/DecompilerHarness.csproj -c Release` — clean (0 warnings/errors).
- `--help` shows `--repo`; parser accepts it.
- End-to-end harvest run through the new `repositoryPaths` path resolved a real row with a github raw `sourceUrl`; the local-repo provider correctly declined a non-matching repo and fell back to the network (the merged, adversarially-reviewed provider in `LocalRepoSourceAcquisition` and its Services unit tests already cover the blob-read + checksum arbitration).

## Risk

Low / mechanical: harness-only (no product-path behavior change), no new heuristics or shapes, and it forwards an already-merged, already-adversarially-reviewed capability. Default behavior is unchanged when `--repo` is omitted (empty list → network as before). No adversarial review requested per AGENTS.md.

### DIFF
--- a/tools/DecompilerHarness/AuthoredSourceHarvest.cs
+++ b/tools/DecompilerHarness/AuthoredSourceHarvest.cs
@@ -66,10 +66,20 @@ static class AuthoredSourceHarvest
         public required Queue<RealMethodTargetEnumerator.RealMethodTarget> Candidates { get; init; }
     }
 
-    public static int Run(IReadOnlyList<string> assemblies, string outputPath, int target, bool evil = false)
-        => RunAsync(assemblies, outputPath, target, evil).GetAwaiter().GetResult();
+    public static int Run(
+        IReadOnlyList<string> assemblies,
+        string outputPath,
+        int target,
+        bool evil = false,
+        IReadOnlyList<string>? repositoryPaths = null)
+        => RunAsync(assemblies, outputPath, target, evil, repositoryPaths).GetAwaiter().GetResult();
 
-    static async Task<int> RunAsync(IReadOnlyList<string> assemblies, string outputPath, int target, bool evil)
+    static async Task<int> RunAsync(
+        IReadOnlyList<string> assemblies,
+        string outputPath,
+        int target,
+        bool evil,
+        IReadOnlyList<string>? repositoryPaths)
     {
         if (assemblies.Count == 0)
         {
@@ -128,7 +138,7 @@ static class AuthoredSourceHarvest
                     var candidate = library.Candidates.Dequeue();
                     attempts++;
 
-                    var record = await TryHarvestAsync(library, candidate, fetcher, evil);
+                    var record = await TryHarvestAsync(library, candidate, fetcher, evil, repositoryPaths);
                     if (record is null)
                     {
                         skipped++;
@@ -228,7 +238,8 @@ static class AuthoredSourceHarvest
         LibraryState library,
         RealMethodTargetEnumerator.RealMethodTarget candidate,
         SourceFetcher fetcher,
-        bool evil)
+        bool evil,
+        IReadOnlyList<string>? repositoryPaths)
     {
         var subject = new FindingSubject(
             $"{candidate.Type}::{candidate.Method}#{candidate.Overload}",
@@ -242,7 +253,8 @@ static class AuthoredSourceHarvest
                 candidate.MetadataToken,
                 candidate.Method,
                 subject,
-                fetcher);
+                fetcher,
+                repositoryPaths);
         }
         catch (Exception ex) when (ex is IOException
             or InvalidOperationException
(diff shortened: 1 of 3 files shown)


richlander:
Follow-up in

richlander:
Ready to merge. All required checks green (test, markdownlint, changes); pack/smoke jobs correctly path-skipped. Mechanical harness-only change (no product-path behavior; forwards the already-reviewed #3040 provider plus an await fix), so no adversarial review required per AGENTS.md.

richlander:
Added `687eccb4`: wire `--repo "$(git rev-parse --show-toplevel)"` into the sanctioned self-corpus harvest procedure (README), which is the concrete "own source" use of this capability (#3052 Scope A).

**Verified end-to-end on a real self-assembly** (not just the generic provider):
- Restored the pinned `dotnet-inspect.any` 0.14.0 self-package and harvested `ILInspector.Metadata.dll` with `--repo <checkout>` → 5 rows resolved.
- Their `sourceUrl`s target `raw.githubusercontent.com/richlander/dotnet-inspect/8681f6ea…` (our own repo).
- That exact blob is present in the **local** git object store (`git cat-file -e <sha>:src/ILInspector.Metadata/ApiDiffAnalyzer.cs` → exit 0), and the harvested `authoredBody` matches the local blob content — proving the source was served from local git with no GitHub round-trip.

Re-running CI on the new commit; will re-confirm Ready once green.

richlander:
## Adversarial review — reconciliation (both clean)

Two independent adversarial reviews from different model families, each in an isolated read-only detached worktree at the exact head `687eccb4` (base `ca3f09a8`):

- **Gemini 3.1 Pro** — no blocking issues.
- **GPT-5.6** — no blocking or non-blocking findings.

Both reviewers exercised the same six attack points with real runs and agreed:

1. **Argument threading** — `repositoryPaths` binds to parameter 6 of `AcquireMemberAsync` (CT is parameter 7 with its default); threaded consistently through `Run`/`RunAsync`/`TryHarvestAsync`. No arg-order/overload bug.
2. **Default-behavior preservation** — with `--repo` omitted, the empty `sourceRepositories` list fails the provider's `repositoryPaths is { Count: > 0 }` guard, so the default network path is unchanged (identical to the pre-change null path). GPT confirmed **byte-identical JSONL** harvesting with vs. without `--repo` (SHA-256 match).
3. **`await` change** (`TryOpenLibrary`) — exception propagation and the `ownershipTransferred`/`finally` `SourceLinkService` disposal are preserved; no leak or double-dispose; the harvest loop stays sequential.
4. **CLI parsing** — `--repo` appears once, `NextArg` advances the index correctly (`--repo` with no value → `Error: --repo requires a value`, exit 1), and it is wired only to the two harvest calls (not the benchmark or other modes).
5. **Evidence integrity** — the `--repo` body is still piped through `VerifyChecksum` before use; no checksum bypass / wrong-repo acceptance. GPT confirmed a harvested row's PDB checksum matched the repo blob; the Services suite (297 tests, incl. local-repo mismatch/missing-blob rejection) passes.
6. **README claim** — both verified the pinned `dotnet-inspect.any` 0.14.0 source commit `8681f6ea` is contained in `origin/main`, and resolution reads the git object store by SHA (working-tree-independent), so the "normal `origin/main` history already carries it" prerequisite is accurate.

No findings to address → head is unchanged, so no re-review needed. Review worktrees removed.

**Ready to merge.** CI green (`test`, `markdownlint`, `changes`); `pack`/smoke jobs correctly path-skipped.

PR #3059, opened 2026-07-23
Render target-typed new() when the LHS target type equals the constructed type (#3058)

### LINKED ISSUE #3058: Decompiler taste: render target-typed new() when the LHS target type equals the constructed type
## Summary

At assignment/declaration sites, the decompiler renders object creation with the
**explicit constructed type** even when the contextual target type on the
left-hand side is exactly that type — e.g. it prints
`StringBuilder sb = new StringBuilder(text.Length);` where idiomatic C# (and
`dotnet/runtime` source) writes `StringBuilder sb = new(text.Length);`. Propose a
**target-typed `new()`** taste transform: when the LHS target type equals the
constructed type, render the target-typed form.

This is a **taste / valid-but-different presentation** question, not a
correctness bug. The two spellings are IL-identical, so it belongs to the
canonical-form machinery in
[docs/decompiler-taste.md](../blob/main/docs/decompiler-taste.md), a sibling of
expression-bodied members, the LINQ query-vs-fluent choice, and the
optional-argument elision transform (#3027).

## Witness

`ILInspector.CSharp.CSharpDeclarationWriter.EscapeReservedKeywordIdentifiers`:

- Decompiled (correct, but valid-but-different):
  `StringBuilder sb = new StringBuilder(text.Length);`
- Target-typed (proposed): `StringBuilder sb = new(text.Length);`

The C# compiler emits `newobj T::.ctor(args)` for both `new T(args)` and, when
the target type is `T`, `new(args)` — so the two forms compile to **identical
IL**; the source author's choice of the short form is unrecoverable from the IL
alone.

## Why this is a taste question (three-class rule)

Per `decompiler-taste.md`:

- **Not class 2 (fidelity-erasing).** Target-typed `new()` erases no distinction
  the IL makes — the constructed type is fixed by the LHS target type, so the
  overload set and the emitted `newobj` token are identical. Nothing observable
  is lost.
- **Class 3 (no IL anchor — oracle tiebreaker).** The two forms are one IL
  equivalence class; we pick one canonical representative. `dotnet/runtime`'s
  `.editorconfig` has an explicit rule here:
  `csharp_style_implicit_object_creation_when_type_is_apparent = true`. That is
  the oracle, so the target-typed form is canonical.
- **Fixed point preserved.** Recompiling `T x = new(args)` re-binds the
  constructed type to `T` and yields identical IL, and re-decompiling yields the
  target-typed form again.

## Proposed rule

At a single-unambiguous-target assignment/declaration site, when the initializer
is a plain object creation `new T(args)` and the contextual target type is
**exactly** `T`, render `new(args)`.

### Scope: LHS positions only (v1)

Hook only positions where the C# target type is unambiguous and does **not**
participate in overload resolution:

- declaring local (`T x = new(...)`), field/property/indirect store, declaring
  stack slot, and typed array-element store.

**Explicitly out of scope for v1:**

- **Call-argument positions.** A target-typed `new()` argument participates in
  **overload resolution** and can change binding — a real soundness hazard. The
  transform must never fire there. (This is why the hook wraps only
  single-target assignment/declaration render, not the general coercion helper.)
- **Return positions.** The type is apparent from the signature but is not a
  "type on the LHS"; keeping the explicit type here also preserves existing
  type-name regression guards. Clean follow-up.

### Soundness

- **Exact-type equality** (`target.Equals(creation.Constructor.DeclaringType)`)
  is the gate: the target *is* the constructed type, so the constructor overload
  set is fixed and the emitted IL is identical by construction. It also
  guarantees the target is concrete/constructible (you cannot `newobj` an
  abstract/interface type).
- **Array-element target** uses the true element type (typed `stelem`) or the
  stelem token type (`object` for `stelem.ref`) — never wrong-but-equal. So a
  ref-type array (`StringBuilder[]`, a `stelem.ref`) soundly declines
  (target = `object`), while a value-type array (`Box[]`, typed `stelem`) fires.
- **Eligibility filter** declines `Nullable<T>` (spelled `T?`), `ValueTuple`
  (spelled as a tuple), multi-dimensional arrays (modeled as `NewObject`),
  pointers/by-refs/function-pointers/open generic parameters, and
  object/collection initializers (a separate node).
- **Decline, don't guess:** any non-exact target keeps the explicit spelling.

## Evidence expected (per decompiler-quality.md)

- The target IL shape (`newobj T::.ctor(args)` stored into a target of type `T`).
- The three-class argument above (class 3, oracle = runtime `.editorconfig`).
- Fixtures covering both source configurations (explicit-type vs target-typed) —
  both compile to one IL — asserting the single canonical rendering, plus close
  negatives: interface/base target (kept), multi-dim array (kept), and
  argument-position creation (kept).
- A blast-radius read showing exactly the intended changes across a corpus.

## Scope boundary

Presentation-only; no semantic change, no raising/structuring change. Return and
call-argument positions are intentionally excluded (soundness + guard
preservation). Sibling of the expression-bodied, query-vs-fluent, and
optional-argument-elision (#3027) canonical-form choices already governed by the
taste doc.

richlander:
- Fixes #3058
- Changes decompiler rendering: at LHS assignment/declaration sites, render
  target-typed `T x = new(args)` when the target type equals the constructed
  type (was always the explicit `T x = new T(args)`).
- Card revision: `a98b974b`

## Change

**Conclusion:** **PASS** — class-3 taste transform, IL-identical by
construction (exact target/constructed-type equality fixes the overload set and
the emitted `newobj` token), scoped to LHS positions so it never touches
overload-resolving call-argument positions; a corpus blast-radius read shows
100% of changes are exactly the intended shortening with zero collateral.

### Original source

The authored C# that compiles to the IL under decompilation
(`ILInspector.CSharp.CSharpDeclarationWriter.EscapeReservedKeywordIdentifiers`):

```csharp
static string EscapeReservedKeywordIdentifiers(string text)
{
    var sb = new StringBuilder(text.Length);
    for (int i = 0; i < text.Length;)
    {
        if (IsIdentifierStart(text[i]))
        {
            int start = i++;
            while (i < text.Length && IsIdentifierPart(text[i]))
                i++;
            string token = text[start..i];
            bool alreadyEscaped = start > 0 && text[start - 1] == '@';
            sb.Append(!alreadyEscaped && CSharpKeywords.RequiresDeclarationEscape(token) ? EscapeIdentifier(token) : token);
            continue;
        }
        sb.Append(text[i++]);
    }
    return sb.ToString();
}
```

The author expressed "the type is apparent" as **`var sb = new StringBuilder(text.Length)`** —
`var` on the left, explicit `new StringBuilder` on the right. IL does not preserve `var`,
so the decompiler always spells the local's type explicitly (`StringBuilder sb = ...`).
With the type now on the **left**, the equivalent "type is apparent" idiom is a
target-typed `new(text.Length)` on the right — the same C# preference
(`csharp_style_implicit_object_creation_when_type_is_apparent`), mirrored.

### Before

`dotnet-inspect` decompiled body at base `ca3f09a8` (harness `--lowered`). The decompiler
spells the explicit local type and lowers the authored `for` to a `while`; the constructor
is the fully-explicit `new StringBuilder(text.Length)`:

```csharp
// ILInspector.CSharp.CSharpDeclarationWriter::EscapeReservedKeywordIdentifiers
StringBuilder sb = new StringBuilder(text.Length);
int i = 0;
while (i < text.Length)
{
    // ...identifier-run escaping, appended to sb...
}
return sb.ToString();
```

- Valid: True
- Correct: True

Valid-but-different: with the explicit `StringBuilder sb` type on the LHS,
`new StringBuilder(text.Length)` is idiomatically written `new(text.Length)` — the same
intent the author expressed with `var`.

### After

`dotnet-inspect` decompiled body at head `a98b974b` (harness `--lowered`). Only the
constructor spelling changes; the body is otherwise identical to Before:

```csharp
// ILInspector.CSharp.CSharpDeclarationWriter::EscapeReservedKeywordIdentifiers
StringBuilder sb = new(text.Length);
int i = 0;
while (i < text.Length)
{
    // ...identifier-run escaping, appended to sb...
}
return sb.ToString();
```

- Valid: True
- Correct: True

The decompiled declaration now matches how the author expressed the same "type is apparent"
intent: authored `var sb = new StringBuilder(...)` ⇔ decompiled `StringBuilder sb = new(...)`.

### Target-typed `new()` fully applied (body otherwise unchanged)

This is a render-stage transform, not a raising pass, so this section is scoped
to the transform — **not** a claim that the whole body is fully raised. The After
body is the same decompiled C# as Before with exactly one change: the LHS
`new StringBuilder(text.Length)` is shortened to `new(text.Length)`. No further
target-typed-`new()` shortening is possible at this site; return-position and
call-argument-position `new()` are intentionally out of scope (they participate
in overload resolution) and are tracked as follow-up in #3058.

The surrounding body is **not** otherwise fully raised, and this PR does not try
to raise it. The `--lowered` view shown above deliberately omits the cosmetic
`for`/`++`/`lock` sugar passes (so the authored `for` shows as `while`), and even
the shipped view still carries unrelated lowering residue — synthetic
temporaries, hoisted declarations, and an un-raised range indexer (authored
`text[start..i]` renders as `text.Substring(...)`). Those belong to other passes
and are out of scope here.

## Change details

`InitializerText(value, target) => TargetTypedNewText(value, target) ?? CoerceText(value, target)`.
`TargetTypedNewText` returns `new(args)` only when the value is a plain
`NewObject` whose constructor's declaring type `Equals` the contextual target
type, the target is a constructible definition/generic-instance (not
`Nullable<T>`, not `ValueTuple`, not a multi-dim array), and the creation is not
a rectangular-array `newobj`.

Hook sites (LHS only): declaring `StoreLocal`, declaring `StoreStackSlot`,
typed-`stelem` `StoreElement`, and the `AssignmentText` tail. The general
`CoerceText` helper is deliberately **not** hooked, because a target-typed
`new()` in a **call-argument** position participates in overload resolution and
could change binding. Return positions are excluded for v1 (kept explicit) to
honor the LHS scope and preserve two type-name regression guards.

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Product build (decompiler + harness) | Pass | Pass |
| Focused tests `TargetTypedNewPassTests` | n/a (new) | 8 passed |
| Snapshot classes (`RaisingPassTests`, `RefStructLayoutFrontierTests`, `KeywordIdentifierTests`) | 133 passed | 133 passed |
| Real witness `CSharpDeclarationWriter::EscapeReservedKeywordIdentifiers` | `new StringBuilder(text.Length)` | `new(text.Length)` |
| Compile-back fidelity | environmental-red locally (CS0009 aspnetcorev2) — see below | Linux CI backstop |

Compile-back / fidelity tests fail wholesale on this Windows box with CS0009 on
a native `aspnetcorev2_inprocess.dll` pulled into Roslyn references (a
pre-existing environmental issue, identical on Baseline); the recompile /
opcode-exact corpus fidelity is validated by Linux CI as with #3042.

## Decompiler quality

**Conclusion:** **PASS** — presentation-only render-stage change; it does not
touch raising/structuring, so lowering/branch residue and fully-raised counts
are unchanged. The relevant corpus signal is the blast-radius read below.

### Blast-radius read (printer-level equivalent of `--pass-impact`)

`--pass-impact` measures registered IR **passes**; this is a render-stage
(printer) change with no pass, so the equivalent is a baseline-vs-head corpus
render diff. Rendered the first 8000 methods of `System.Private.CoreLib`
(5866 with output) with and without the printer change (stash printer-only for
baseline):

| Metric | Value |
| --- | ---: |
| Methods rendered | 5866 |
| Methods changed | 229 (3.90%) |
| Lines changed | 357 |
| Changed lines that are exactly outer `new T(...)` -> `new(...)` (LHS) | 357 / 357 (100%) |
| Collateral (non-transform) changes | 0 |

Every changed line is the intended target-typed shortening. Lines containing an
inner argument-position `new` confirm the argument-position decline — only the
outer LHS `new` shortens, the inner is untouched, e.g.
`new Guid(new ReadOnlySpan<Guid>(this))` -> `new(new ReadOnlySpan<Guid>(this))`.

## Validation

```bash
dotnet build src/ILInspector.Decompiler.Tests -c Release
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.TargetTypedNewPassTests
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.RaisingPassTests -class ILInspector.Decompiler.Tests.RefStructLayoutFrontierTests -class ILInspector.Decompiler.Tests.KeywordIdentifierTests
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CollectionExpressionFrontierTests.cs
+++ b/src/ILInspector.Decompiler.Tests/CollectionExpressionFrontierTests.cs
@@ -93,7 +93,8 @@ public class CollectionExpressionFrontierTests
         Assert.Contains(function.Descendants.OfType<Call>(), call => call.Callee.Name == "AddRange");
 
         var output = Print(nameof(CfgSampleClass.CollectionWithCapacity));
-        Assert.Contains("new List<string>(values.Count * 2)", output);
+        Assert.Contains("List<string>", output);
+        Assert.Contains("= new(values.Count * 2)", output);
         Assert.Contains(".AddRange(", output);
         Assert.DoesNotContain("[with", output);
     }
--- a/src/ILInspector.Decompiler.Tests/IrImporterTests.cs
+++ b/src/ILInspector.Decompiler.Tests/IrImporterTests.cs
@@ -2120,13 +2120,14 @@ public class RaisingPassTests
     public void StructConstructor_InPlaceCtor_PrintsNewObject()
     {
         // The in-place struct .ctor (ldloca; call S::.ctor) must print as an
-        // assignment of a fresh value, never the illegal handler..ctor(...).
+        // assignment of a fresh value, never the illegal handler..ctor(...). The
+        // type is apparent on the declaration, so the creation renders target-typed.
         using var source = MetadataSource.Open(typeof(CfgSampleClass).Assembly.Location);
         var result = CSharpPrinter.Print(RunThroughStructConstructor(nameof(CfgSampleClass.InterpolatedStruct), source));
         Assert.True(result.Succeeded);
         string output = result.Output!.ReplaceLineEndings("\n").TrimEnd();
 
-        Assert.Contains("new DefaultInterpolatedStringHandler(", output);
+        Assert.Contains("DefaultInterpolatedStringHandler V_0 = new(", output);
         Assert.DoesNotContain("..ctor", output);
     }
 
@@ -2170,7 +2171,7 @@ public class RaisingPassTests
         Assert.Equal(0, store.Index);
         var value = Assert.IsType<NewObject>(store.Value);
         Assert.Equal(ctor, value.Constructor);
-        Assert.Equal("Carrier V_0 = new Carrier(42);", CSharpPrinter.Print(function).Output!.ReplaceLineEndings("\n").Trim());
+        Assert.Equal("Carrier V_0 = new(42);", CSharpPrinter.Print(function).Output!.ReplaceLineEndings("\n").Trim());
     }
 
     [Fact]
--- a/src/ILInspector.Decompiler.Tests/LadderRung1GateTests.cs
+++ b/src/ILInspector.Decompiler.Tests/LadderRung1GateTests.cs
@@ -107,7 +107,7 @@ public class LadderRung1GateTests
 
         // Object construction, property/method calls, Console.WriteLine, branch.
         var main = Body("Main");
-        Assert.Contains("new Program()", main);
+        Assert.Contains("Program program = new();", main);
         Assert.Contains("program.Greet(\"world\")", main);
         Assert.Contains("program.IsPositive(program.Count)", main);
         Assert.Contains("Console.WriteLine(", main);
--- a/src/ILInspector.Decompiler.Tests/RefStructLayoutFrontierTests.cs
+++ b/src/ILInspector.Decompiler.Tests/RefStructLayoutFrontierTests.cs
@@ -29,7 +29,7 @@ public class RefStructLayoutFrontierTests
 
         Assert.Equal(DecompilationFidelity.Full, function.Fidelity);
         Assert.Empty(function.Diagnostics);
-        Assert.Contains("SpanBackedView view = new SpanBackedView(span);", output);
+        Assert.Contains("SpanBackedView view = new(span);", output);
         Assert.Contains("return view.At(index) + view.Length;", output);
         Assert.DoesNotContain("IsByRefLike", output);
         Assert.DoesNotContain("/*", output);
--- a/src/ILInspector.Decompiler.Tests/TargetTypedNewFixtures.cs
+++ b/src/ILInspector.Decompiler.Tests/TargetTypedNewFixtures.cs
@@ -77,6 +77,26 @@ public class TargetTypedNewFixtures
     public int ArgumentPositionDeclines(int n)
         => Accept(new StringBuilder(n));
 
+    // Negative: a bare `object` target. `object o = new object()` is a valid
+    // target-typed-new, but the raw `object` target is indistinguishable at this seam
+    // from a `dynamic` place (which the IR erases to `object`), where `new()` is
+    // CS8752 — so all bare `System.Object` constructions decline. `new object()` has
+    // no type name to drop anyway.
+    public bool ObjectTargetDeclines()
+    {
+        object o = new object();
+        object p = new object();
+        return ReferenceEquals(o, p);
+    }
+
+    // Negative: a `dynamic` parameter reassignment. The signature spells `dynamic
+    // value`, but the IR store carries the erased `object` type; shortening to
+    // `value = new()` would be CS8752. The bare-object decline keeps `new object()`.
+    public void DynamicParamDeclines(dynamic value)
+    {
+        value = new object();
+    }
+
     static int Accept(StringBuilder builder) => builder.Length;
     static int Accept(object value) => value.GetHashCode();
 
--- a/src/ILInspector.Decompiler.Tests/TargetTypedNewFixtures.cs
+++ b/src/ILInspector.Decompiler.Tests/TargetTypedNewFixtures.cs
@@ -1,0 +1,90 @@
+namespace ILInspector.Decompiler.Tests;
+
+using System.Collections.Generic;
+using System.Text;
+
+// Fixtures for the target-typed `new()` rendering transform. Each method below is
+// decompiled by TargetTypedNewPassTests; the transform shortens `new T(args)` to
+// `new(args)` only when the contextual target type is exactly the constructed type.
+public class TargetTypedNewFixtures
+{
+    StringBuilder _builder = new StringBuilder();
+
+    // Positive: a local whose declared type is exactly the constructed type — the
+    // canonical `StringBuilder sb = new(n)` case.
+    public int LocalDeclaration(int n)
+    {
+        StringBuilder sb = new StringBuilder(n);
+        sb.Append('x');
+        return sb.Length;
+    }
+
+    // Positive: an instance field store whose field type equals the constructed
+    // type (reached through the assignment path).
+    public void FieldStore(int n)
+    {
+        _builder = new StringBuilder(n);
+    }
+
+    // Boundary: return positions are out of the v1 LHS-only scope, so these stay
+    // explicit (the transform never fires here yet).
+    public StringBuilder ReturnPosition(int n)
+    {
+        if (n > 0)
+            return new StringBuilder(n);
+        return new StringBuilder();
+    }
+
+    // Positive: a value-type construction (newobj on a struct) into a local of the
+    // same struct type.
+    public int StructLocal(int n)
+    {
+        Box box = new Box(n);
+        box.Bump();
+        return box.Value;
+    }
+
+    // Positive: an array element store whose element type equals the constructed
+    // type. A value-type element array (`stelem` carries the element token) exposes
+    // the real element target; a reference-type array's `stelem.ref` is untyped
+    // (object) and would soundly decline instead.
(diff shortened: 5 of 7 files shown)


richlander:

Two different-model reviews, both **CLEAN at the fixed head `fa15f98e`**.

### Round 1 (head `a98b974b`)

- **Gemini 3.1 Pro** — CLEAN. Verified the exact-type-equality guard, LHS-only scope, and that overload-resolution scopes are untouched.
- **GPT-5.6** — raised **two Medium findings**:
  1. **Dynamic LHS mistaken for `object`.** `dynamic value; value = new object()` shortened to illegal `value = new()` (CS8752), because the signature spells `dynamic` while the IR `StoreArgument` carries the erased `object`. Reproduced with a compiled dynamic-parameter fixture.
  2. **Typed `stelem` token != C# element target.** For a covariant array where the `stelem` token is wider than the array's static element type, `a[i] = new()` binds to the narrower element type, so `new()` would construct a different type than the original `newobj`. Not compiler-producible (CS0029), demonstrated via Reflection.Emit.

### Fixes (`fa15f98e`)

- **Finding 1** — decline target-typed `new()` when the **constructed type is bare `System.Object`** (assembly-agnostic `IsSystemObjectType`). Since the transform already requires `target.Equals(constructed)`, a `dynamic` target implies `constructed == object`, so this single decline covers every dynamic destination (param/field/property/local/by-ref). `new object()` carries no type name to drop; **zero** such LHS sites exist in `System.Private.CoreLib`.
- **Finding 2** — a purely-additive tightening: offer `new()` at an array element store only when the array's **static element type equals the `stelem` token** (`StoreElementNewTarget` + a two-arg `InitializerText` that keeps coercion on the token but gates `new()` on the element type). Value-type element arrays still fire; reference-type/covariant arrays decline.

Added 3 negative tests: `BareObjectTarget_KeepsExplicitType`, `DynamicParameterTarget_KeepsExplicitType`, `CovariantElementStore_TokenWiderThanArrayElement_KeepsExplicitType`.

### Round 2 (fixed head `fa15f98e`) — both CLEAN

- **GPT-5.6** — both findings resolved; build succeeded and all 11 focused tests pass; no new issues.
- **Gemini 3.1 Pro** — sound, no new defects; confirmed `IsSystemObjectType` is exactly correct (ignores `List<object>`/`object[]`), the two-arg `InitializerText` split drops no casts, and all original invariants hold; 11/11 tests pass.

### Evidence

- Whole-assembly render of `System.Private.CoreLib` is **byte-identical** before and after the fix — both declines only affect non-compiler-reachable covariant IL and bare-object LHS (zero in CoreLib), so the reviewed positives are unchanged.
- `TargetTypedNewPassTests` 11/11 green.
- Windows-local fast-suite reds are environmental (Roslyn recompile CS0009 on the native `aspnetcorev2_inprocess.dll`); the Linux `test` job is authoritative.

richlander:
Ready to merge

Both fixed-head adversarial reviews (GPT-5.6 + Gemini 3.1 Pro) are CLEAN at `fa15f98e`, and the authoritative Linux `test` job passed (build-net10, csharp-diff-smoke, changes also green). Reconciliation above.

richlander:
Follow-ups for the "not otherwise fully raised" note on this witness (`EscapeReservedKeywordIdentifiers`):

- #3071 — mechanical raising residue: single-use copy-temp fold (`V_5`), conditional-argument temp inline (`S_2`), short-circuit `&&` from a `cond ? false : rhs` ternary + De-Morgan canonicalization, and declaration sinking.
- #3070 — companion class-3 taste raise: `Substring/Slice(start, end - start)` -> range indexer `s[start..end]` (`csharp_style_prefer_range_operator`), a sibling of #3058.

These are out of scope for this PR (which only shortens the LHS `new`).

PR #3061, opened 2026-07-23
Fold spilled struct rvalue receivers to recompose && return chains (#3051)

### LINKED ISSUE #3051: Decompiler: recompose bool early-exit guard sequence into a single &&-chain return
## Root cause (corrected after investigation)

The `&&`-chain recomposition I originally described **already works**: `StructuringPass` folds a `bool` early-exit guard sequence into a single `return A && B && … ;`. A plain relational chain such as
`start.Day == 1 && start.TimeOfDay < TimeSpan.FromMinutes(1) && end.Day == 1 && end.TimeOfDay < TimeSpan.FromMinutes(1)`
folds correctly today.

The witness `NodaTime …BclAdjustmentRule::IsStandardOffsetOnlyRule` fails for a **narrower** reason: two of its operands are `daylight.TimeOfDay.TimeOfDay < …`, where `TransitionTime.TimeOfDay` returns a **struct** (`DateTime`). Calling `.TimeOfDay` on that result needs the receiver's address, so csc spills the intermediate to a local:

```
StoreLocal V_2 = daylight.TimeOfDay        // struct rvalue spilled
ConditionalBranch  … LoadLocalAddress V_2 …   // V_2.TimeOfDay, receiver via address-of
```

That leading `StoreLocal` in the guard block is the blocker: `StructuringPass` only folds pure single-condition guard blocks, and `SpilledReceiverFold` **deliberately declines address-taken locals** (`CountPlaces` marks `LoadLocalAddress` as `AddressTaken`, and `SpillLoadInside` bails on it). Reproduced end-to-end: even a *single*, non-reused struct-receiver spill blocks the fold, producing expanded `if (…) return false;` guards plus a `V = …;` spill instead of the `&&` chain.

## Actual fix

A focused **spilled struct-rvalue-receiver fold**: `V = obj.StructProp; … V.Member …` (V used once, via `LoadLocalAddress`, def is a value-producing expression) → `obj.StructProp.Member`. Block-local, effect-order-preserving, value-def-only (declines when the def is itself an address/ref load, to avoid changing defensive-copy semantics). Once the spill folds away, the operand block becomes a pure single-condition guard and existing structuring recomposes the whole `&&` chain — so this is a struct-receiver readability fold, not a control-flow pass. It is independently useful for any `V = a.StructProp; V.Member` idiom.

## Soundness surface (for review)

- Def must produce a fresh rvalue (property get / method call), never an lvalue address/ref (aliasing / defensive-copy hazard).
- Single block-local use, reached before the next store to the same local.
- No effect reordering between the spill store and its use.
- Preserve behavior for `readonly`/mutating members.

## Reproduction

```bash
dnx dotnet-inspect -y -- member "NodaTime.TimeZones.BclDateTimeZone+BclAdjustmentRule" -m IsStandardOffsetOnlyRule --package NodaTime@3.1.11 --all -S "Decompiled Source"
```

Follow-up surfaced by #3039 / #2955. Distinct from #2877, #2974, #3020.

richlander:
- Fixes #3051
- Recomposes short-circuit `&&` boolean-return chains that a spilled struct rvalue receiver had fragmented into a guard cascade.
- Card revision: `1bca0bd8`

## Change

**Conclusion:** **PASS** — a conservatively-gated, opcode-preserving de-spill fold turns the NodaTime witness back into its exact one-line source shape; every gate declines on the close negatives; and adversarial review drove the fold's liveness reasoning from an unsound document-order proxy to a sound whole-slot basic-block check.

A member access on a struct-returning property/method receiver
(`daylight.TimeOfDay.TimeOfDay`) cannot take the address of an rvalue, so csc
spills it: `stloc V (= daylight.TimeOfDay); ldloca V; call get_TimeOfDay`. That
leading spill store makes an otherwise-pure guard block impure, which blocks
`StructuringPass` from nesting the guard chain and therefore
`BooleanFoldingPass` from recomposing it into a single `&&` return. The result
is a cascade of `if (...) return false;` guards.

`StructReceiverInliningPass` runs immediately before `StructuringPass` and folds
a single-use struct rvalue temporary back into its member-access receiver
(`V = a.Prop; ... V.Member` → `a.Prop.Member`), so the guard block becomes pure
and the chain recomposes.

### Original source

```csharp
private static bool IsStandardOffsetOnlyRule(TimeZoneInfo.AdjustmentRule rule)
{
    var daylight = rule.DaylightTransitionStart;
    var standard = rule.DaylightTransitionEnd;
    return daylight.IsFixedDateRule && daylight.Day == 1 && daylight.Month == 1 &&
           daylight.TimeOfDay.TimeOfDay < TimeSpan.FromMinutes(1) &&
           standard.IsFixedDateRule && standard.Day == 1 && standard.Month == 1 &&
           standard.TimeOfDay.TimeOfDay < TimeSpan.FromMinutes(1);
}
```

### Before

```csharp
TimeZoneInfo.TransitionTime daylight = rule.DaylightTransitionStart;
TimeZoneInfo.TransitionTime standard = rule.DaylightTransitionEnd;
if (!daylight.IsFixedDateRule) { return false; }
if (daylight.Day != 1) { return false; }
if (daylight.Month != 1) { return false; }
DateTime V_2 = daylight.TimeOfDay;
if (V_2.TimeOfDay >= TimeSpan.FromMinutes(1d)) { return false; }
if (!standard.IsFixedDateRule) { return false; }
if (standard.Day != 1) { return false; }
if (standard.Month != 1) { return false; }
V_2 = standard.TimeOfDay;
return V_2.TimeOfDay < TimeSpan.FromMinutes(1d);
```

- Valid: True
- Correct: True

Behavior-correct, but the `DateTime V_2 = daylight.TimeOfDay;` spill breaks
guard-block purity, so the chain stays a cascade instead of the source `&&`
chain (the two spilled predicates even invert to `>=` guards).

### After

```csharp
TimeZoneInfo.TransitionTime daylight = rule.DaylightTransitionStart;
TimeZoneInfo.TransitionTime standard = rule.DaylightTransitionEnd;
return daylight.IsFixedDateRule && daylight.Day == 1 && daylight.Month == 1 && daylight.TimeOfDay.TimeOfDay < TimeSpan.FromMinutes(1d) && standard.IsFixedDateRule && standard.Day == 1 && standard.Month == 1 && standard.TimeOfDay.TimeOfDay < TimeSpan.FromMinutes(1d);
```

- Valid: True
- Correct: True

Matches the original source shape exactly; fidelity remains `Full`. This method
also exercises a **reused slot** — csc packs both `daylight.TimeOfDay` and
`standard.TimeOfDay` into one `DateTime` temporary — so the fold must handle
per-store live ranges, not just method-wide single-use temporaries.

### Fully raised

## Soundness gates

The fold reasons only about **basic-block-local adjacency**, never document
order. It fires for a store `S` to slot `k` only when all hold (see
`StructReceiverInliningPass` XML doc):

- **Fresh rvalue definition** — `S`'s value is a `Call`/`LoadProperty` with a
  non-`ByRef` result, never an address/ref/place. A byref temp aliases real
  storage, so folding could change defensive-copy or write semantics.
- **Whole-slot fold-safety** — *every* read of slot `k` in the method is a
  member-receiver address load whose enclosing statement immediately follows a
  store to `k` in the same block. Because a `Block` is a basic block, that
  adjacent store is each read's unique reaching definition, so the slot is never
  live across a block boundary and no loop back-edge or conditional-redefinition
  bypass can route a stale value to a read. A reused slot stays foldable because
  each store/use pair is checked independently.
- **Single adjacent member-receiver use** — the fold's use sits in the statement
  immediately after `S`, is the slot's only read there, and is the instance of a
  property/field load or the receiver of an instance call (write places
  excluded).
- **Effect order** — the use is not in a short-circuit/ternary sub-position and
  nothing order-sensitive evaluates before it.

Because C# re-introduces the identical spill on recompile, the round-trip is
opcode-preserving.

## Adversarial review

Two independent adversarial reviewers (GPT-5.6 and Gemini 3.1 Pro) drove the
correctness of this change; see the reconciliation comment for full attribution.
In summary they found — with runtime repros — that the fold's original
document-order live-range proxy was unsound in **both** directions (loop
back-edge and forward conditional bypass). The final head replaces all
document-order reasoning with the whole-slot basic-block check above; both
reviewers re-reviewed the fixed head `1bca0bd8` and returned **CLEAN**.

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| `StructReceiverInliningPassTests` | new | 9 passed |
| `LoweringCoverageTests` (classification) | 6 passed | 6 passed |
| Real witness `BclAdjustmentRule::IsStandardOffsetOnlyRule` | guard cascade | single `&&` return (Full) |
| Loop / conditional-bypass repros (do/while, for, if-bypass) | — | all decline (0 changed) |
| NodaTime blast radius | — | 91/2866 methods, 0 pass bugs |
| CoreLib blast radius (stress) | — | 601/37093 methods, 0 pass bugs |
| CI compile-back / fidelity (`test`, Linux) | — | green at `1bca0bd8` |

Compile-back / fidelity gates are the authoritative opcode-preservation signal
and run in CI (Linux). They are not runnable on this local Windows box — the
harness pulls all runtime references and the aspnetcore native shim fails Roslyn
with CS0009, which stalls the compile-back suite (2039/2039 recompile-fail) — so
local soundness was verified via pass tests plus runtime repros, and
opcode-exact fidelity is confirmed at scale by the green Linux `test` job.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.StructReceiverInliningPassTests
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.LoweringCoverageTests
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/StructReceiverInliningPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/StructReceiverInliningPassTests.cs
@@ -8,11 +8,14 @@ namespace ILInspector.Decompiler.Tests;
 // can nest and boolean folding can recompose into one && return.
 //
 // The pass is gated for soundness: the definition must be a fresh rvalue (a
-// call/property load, never an address/place), the temp must have exactly one
-// reader in its live range and that read must be a member-access receiver, and
-// the read must sit in the statement immediately after the store with no
-// order-sensitive node evaluated before it. Opcode-preservation of the fold is
-// proven by the compiled-fixture round-trip described at the end of this file.
+// call/property load, never an address/place), and the slot must be fold-safe
+// method-wide — every read of it is a member-receiver address load whose
+// statement immediately follows a store to the same slot in the same block, so
+// each read's reaching definition is the adjacent store and the slot is never
+// live across a block boundary. The folded store's use must additionally be the
+// slot's only read in the adjacent statement with no order-sensitive node
+// evaluated before it. Opcode-preservation of the fold is proven by the
+// compiled-fixture round-trip described at the end of this file.
 [Trait("Area", "Pass")]
 public class StructReceiverInliningPassTests
 {
@@ -199,9 +202,9 @@ public class StructReceiverInliningPassTests
 
     // A back-edge routes execution, on a later iteration, from the loop-body
     // store back to a read at the loop head that precedes the store in document
-    // order. That read observes the store's value, so folding the store away
-    // would leave it stale. The forward single-reader window cannot see the head
-    // read (lower document position); the back-edge gate must decline.
+    // order. That head read is not immediately preceded by a store to the slot,
+    // so the whole-slot fold-safety check fails and the fold declines — no
+    // reasoning about document order or back-edge spans is required.
     [Fact]
     public void LoopCarriedReader_IsNotFolded()
     {
@@ -218,6 +221,35 @@ public class StructReceiverInliningPassTests
         RunPass(function);
 
         Assert.Equal(1, StoreCount(function, 0));
+        function.CheckInvariant();
+    }
+
+    // Forward bypass: the store's adjacent use looks single, but a later read
+    // sits after a conditionally-taken second store to the same slot. On the
+    // path that skips the second store, that later read observes the first
+    // store's value — so folding the first store away would leave it reading a
+    // removed value. The later read is not immediately preceded by a store, so
+    // the whole-slot check fails and both stores survive.
+    [Fact]
+    public void ForwardBypassReader_IsNotFolded()
+    {
+        // block0: v = a.Inner; v.Value; if (cond) goto block2 (skip the second store)
+        var block0 = MakeBlock(0,
+            new StoreLocal(0, S, Inner(new LoadArgument(0, "a", Outer))),
+            new ExpressionStatement(Value(new LoadLocalAddress(0, S))),
+            new ConditionalBranch(new LoadArgument(1, "cond", Bool), 32));
+        // block1: v = b.Inner (the conditionally-executed redefinition)
+        var block1 = MakeBlock(16,
+            new StoreLocal(0, S, Inner(new LoadArgument(2, "b", Outer))));
+        // block2: return v.Value — the join read, reached with block0's value on
+        // the bypass path.
+        var block2 = MakeBlock(32,
+            new Return(Value(new LoadLocalAddress(0, S))));
+        var function = BuildFunction([block0, block1, block2]);
+
+        RunPass(function);
+
+        Assert.Equal(2, StoreCount(function, 0));
         function.CheckInvariant();
     }
 }
--- a/src/ILInspector.Decompiler.Tests/StructReceiverInliningPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/StructReceiverInliningPassTests.cs
@@ -196,4 +196,28 @@ public class StructReceiverInliningPassTests
         Assert.Equal(1, StoreCount(function, 0));
         function.CheckInvariant();
     }
+
+    // A back-edge routes execution, on a later iteration, from the loop-body
+    // store back to a read at the loop head that precedes the store in document
+    // order. That read observes the store's value, so folding the store away
+    // would leave it stale. The forward single-reader window cannot see the head
+    // read (lower document position); the back-edge gate must decline.
+    [Fact]
+    public void LoopCarriedReader_IsNotFolded()
+    {
+        // Head (offset 0): r = v.Value  — the loop-carried read of slot 0.
+        var head = MakeBlock(0,
+            new StoreLocal(1, Bool, Value(new LoadLocalAddress(0, S))));
+        // Body (offset 16): v = outer.Inner; if (v.Value) goto head — store,
+        // adjacent member-receiver use, and the back-edge to the head.
+        var body = MakeBlock(16,
+            new StoreLocal(0, S, Inner(new LoadArgument(0, "outer", Outer))),
+            new ConditionalBranch(Value(new LoadLocalAddress(0, S)), 0));
+        var function = BuildFunction([head, body]);
+
+        RunPass(function);
+
+        Assert.Equal(1, StoreCount(function, 0));
+        function.CheckInvariant();
+    }
 }
--- a/src/ILInspector.Decompiler.Tests/StructReceiverInliningPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/StructReceiverInliningPassTests.cs
@@ -1,0 +1,199 @@
+using ILInspector.Decompiler.Pipeline;
+
+namespace ILInspector.Decompiler.Tests;
+
+// #3051: StructReceiverInliningPass folds a single-use struct rvalue temporary
+// back into its member-access receiver (V = a.Prop; ... V.Member -> a.Prop.Member)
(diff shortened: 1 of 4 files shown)


richlander:

Two independent adversarial reviewers examined this change: **GPT-5.6**
(`review-gpt`) and **Gemini 3.1 Pro** (`review-gemini`). Each ran in an isolated
review worktree with scratch under `/tmp`. Summary of findings and resolution:

### Round 1 — original document-order fold (`d25ec96e`)

- **Loop-carried store (BOTH reviewers, blocking).** The fold used document /
  pre-order position as a live-range proxy. Across a loop back-edge, a store
  whose value is read at the *top* of the next iteration was treated as dead and
  folded away, going stale. Both produced runtime repros.
  - *Interim resolution:* commit `43938483` added a back-edge span gate.

### Round 2 — interim back-edge gate (`43938483`)

- **`do/while` structured node (GPT, blocking).** `DoWhileLoopPass` runs *before*
  this pass and collapses the loop into a `DoWhileLoop` node, so the flat
  back-edge scan saw no back-edge and still folded the loop-carried store. Repro:
  `dowhilecarry` (1/8 methods wrongly changed).
- **Forward conditional bypass (Gemini, blocking).** The "next store kills the
  range" heuristic assumed the textually-next store to the slot was
  unconditional; when that store sits inside an `if`, the not-taken path leaves
  the first value live at a later read. Repro: `bypasscarry` (`cond == false`
  produced wrong output).

Both round-2 findings were the **same root cause both reviewers named**:
document order is an unsound liveness proxy in *both* directions. The interim
back-edge gate was insufficient.

### Resolution — whole-slot basic-block rework (`1bca0bd8`)

I abandoned document-order reasoning entirely (as both reviewers recommended)
and replaced it with a **whole-slot basic-block adjacency** check: a store folds
only if *every* read of that slot in the method is a member-receiver address load
whose enclosing statement immediately follows a store to the slot in the **same
`Block`**. Because a `Block` is a basic block, that adjacent store is each read's
unique reaching definition — the slot is never live across a block boundary, so
no loop back-edge or forward conditional bypass can route a stale value to a
read. Reused slots (the witness packs `daylight.TimeOfDay` and
`standard.TimeOfDay` into one `DateTime` slot) stay foldable because each
store/use pair is checked independently — which is why the simpler
method-wide-single-use suggestion was rejected: it would break the witness.

This is a **material algorithm change**, so both reviewers received a fresh
prompt to attack the new algorithm (not merely re-check a gate). Verified on
`1bca0bd8`:

- All four repros (`dowhilecarry`, `bypasscarry`, two for-loop cases) **decline**
  (0 changed).
- Witness folds to the single-`&&` `Full` output (byte-identical).
- 0 pass bugs over NodaTime (2866) and CoreLib (37093); blast radius 91 / 601 —
  the more-precise structural gate reclaims sound folds the blunt interim gate
  over-declined, and both reviewers verified these are legitimate.

### Fixed-head verdicts on `1bca0bd8`

- **Gemini 3.1 Pro — CLEAN.** Exhaustively confirmed all four attack points
  decline and audited that `LoadLocal`/`LoadLocalAddress` are the only
  local-reading IR nodes at this stage.
- **GPT-5.6 — CLEAN.** Repros decline, witness folds `Full`, 9/9 pass tests.

Both fixed-head reviews are clean; CI is green on `1bca0bd8`.

richlander:
Ready to merge

PR #3062, opened 2026-07-23
Migrate compile-back harness target render to the product (#2996)

richlander:
- Advances #2996 (whole members as the decompiler's natural output unit)
- Changes the compile-back fidelity harness to render method/operator targets from the product instead of a self-spelled signature skeleton

## Change

**Conclusion:** **REVIEW** — the harness stops maintaining a parallel signature speller and consumes the product's whole-member render; product correctness is proven byte-identical locally, and compile-back opcode-identity is CI-validated (local compile-back is env-blocked by #3018-C).

This is a **harness-infrastructure** change: it does **not** change any decompiler output. It changes how the compile-back fidelity harness reconstructs the module it recompiles.

### Problem

The fidelity harness self-spelled each target method's signature into the reconstructed compile-back module (`EmitMethod`). That private skeleton reconstruction emits invalid C# for several signature shapes (CS0305 generic arity, CS1031, CS1750), which surface as false `Invalid` verdicts that trace to the harness, not the decompiler.

### Fix

For **method and operator** targets, the harness now splices the product's whole-member render (CSharp-owned signature + decompiler-owned body) in place of the self-spelled signature, so it consumes one product implementation instead of maintaining a parallel speller (RTS-orchestrator direction, #2996).

**Product** — `MemberBodyProducer.ProduceMembers`, the per-type batch form of `ProduceMember`. Opens the assembly and builds its type maps once for the whole type, then renders each member through the exact same single-member composition, so each entry is byte-identical to `ProduceMember` while the per-member setup cost (assembly open + type-map walk) is paid once per type rather than once per member.

**Harness** — `CollectTypeEntries` resolves each target to its `ApiType`/`ApiMember`, looks up the memoized batch render, and carries the product text on `TargetBody.WholeMember`; the emit loop splices it (re-indented) instead of calling `EmitMethod`. `BuildUnit` hoists the render's harvested namespaces into the unit usings so the product's shortened type names bind. Any non-migrated kind (constructors, property accessors) or incomplete render falls back to the existing `EmitMethod` path (honest degradation, no regression).

### Scope

Slice 1 migrates **method/operator** targets only. Constructors (lifted field initializers / primary-ctor headers) and property accessors (whole-property impedance) are deferred to a follow-up slice.

## Evidence

| --- | --- | --- |
| Harness build | Pass | Pass |
| `MemberBodyProducerMemberRenderTests` (byte-identity incl. new batch case) | 5 passed | 6 passed |
| All `MemberBodyProducer*Tests` | 38 passed | 38 passed |
| Full-assembly `ProduceMembers` stress (516 types / 4700 completed members) | n/a (new API) | 0 mismatch vs `ProduceMember`, 0 failure, no crash |

**Byte-identity is the key local proof:** every batch entry equals the per-member `ProduceMember` text, which is itself byte-identical to the whole-type `Project` listing segment (the render users compile). The batch only amortizes setup; it changes no member's output.

**Compile-back opcode-identity** (SkeletonEmitTests, DiffFixtureFidelityTests, fidelity gates) is validated by **CI**. Local compile-back is blocked on this Windows machine by the known native-reference issue #3018-C (Roslyn pulls `aspnetcorev2_inprocess.dll`, a native DLL, into the recompile reference set → CS0009), and the giant test assembly's reconstructed module exceeds local resources — both pre-existing and unrelated to this change.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.MemberBodyProducerMemberRenderTests"
```

Adversarial review (verdict-changing) will follow in a separate comment once CI is green.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/MemberBodyProducerMemberRenderTests.cs
+++ b/src/ILInspector.Decompiler.Tests/MemberBodyProducerMemberRenderTests.cs
@@ -127,6 +127,22 @@ public sealed class MemberBodyProducerMemberRenderTests
         Assert.Contains("{", rendered.Text);
         Assert.DoesNotContain("=>", rendered.Text);
     }
+
+    [Fact]
+    public void ProduceMember_PreservesQualifiedNameInsideStringLiteral()
+    {
+        var type = Specimen();
+        var member = Assert.Single(type.Members, m => m.Name == nameof(MemberRenderSpecimen.QuotedTypeName));
+
+        var rendered = MemberBodyProducer.ProduceMember(type, member, AssemblyPath, pdbPath: null);
+
+        Assert.Equal(MemberBodyProductionStatus.Complete, rendered.Status);
+        // Name-shortening must never reach inside a string literal. An escaped
+        // quote must not flip in-literal parity and shorten System.String to
+        // String, which would corrupt the ldstr operand and induce a false
+        // compile-back OperandDiff (#3062).
+        Assert.Contains("System.String", rendered.Text);
+    }
 }
 
 #pragma warning disable CA1822 // members are instance to exercise real signatures
@@ -147,5 +163,11 @@ public sealed class MemberRenderSpecimen
     }
 
     public void ThrowStub() => throw new NotImplementedException();
+
+    // A string constant whose value contains a double-quote followed by a
+    // fully-qualified type name. The rendered literal escapes the quote (\"),
+    // so a name-shortener that splits on '"' without honoring escapes flips its
+    // in-literal parity and mutates System.String inside the constant (#3062).
+    public string QuotedTypeName() => "a \"System.String\" b";
 }
 #pragma warning restore CA1822
--- a/src/ILInspector.Decompiler.Tests/MemberBodyProducerMemberRenderTests.cs
+++ b/src/ILInspector.Decompiler.Tests/MemberBodyProducerMemberRenderTests.cs
@@ -47,6 +47,28 @@ public sealed class MemberBodyProducerMemberRenderTests
     }
 
     [Fact]
+    public void ProduceMembers_BatchIsByteIdenticalToPerMember_ForEveryMember()
+    {
+        var type = Specimen();
+        var batch = MemberBodyProducer.ProduceMembers(type, AssemblyPath, pdbPath: null);
+
+        Assert.NotEmpty(type.Members);
+        foreach (var member in type.Members)
+        {
+            var single = MemberBodyProducer.ProduceMember(type, member, AssemblyPath, pdbPath: null);
+            Assert.True(batch.TryGetValue(member, out var batched),
+                $"batch render missing member {member.Name}");
+
+            // The batch entry is byte-identical to the per-member render — same
+            // status, text, and imports. The batch only amortizes the assembly
+            // open and type-map build; it must not change any member's output.
+            Assert.Equal(single.Status, batched.Status);
+            Assert.Equal(single.Text, batched.Text);
+            Assert.Equal(single.Namespaces, batched.Namespaces);
+        }
+    }
+
+    [Fact]
     public void ProduceMember_RendersExpressionBodiedArrow()
     {
         var type = Specimen();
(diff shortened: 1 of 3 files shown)


richlander:

Two independent reviewers (GPT `gpt-5.6-sol`, Gemini `gemini-3.1-pro-preview`), each isolated in its own exact-head worktree, reviewed base `origin/main` → head `7e89cfdf`. Both **independently converged on the same two findings**.

### Finding 1 — string-operand corruption in `ShortenQualifiedNames` (High) — FIXED
`ShortenQualifiedNames` split each line on `"` with even/odd parity to skip string literals, but the split was not escape-aware. An escaped quote (`\"`) inside a literal — or a quote char constant (`'"'`) — flipped the in-literal parity, exposing the literal's contents to name-shortening. A constant like `"a \"System.String\" b"` was corrupted to `"a \"String\" b"`, changing the `ldstr` operand and inducing a **false compile-back `OperandDiff`**.

- This was a **pre-existing product bug** in `ProduceMember`; pr5a newly routes the harness recompile through it, making it verdict-affecting — so it is in scope here.
- GPT correctly identified the root cause; its specific example string had no interior `\"` so did not itself trigger it. Gemini's repro (`"a \"System.String\" b"`) was accurate.
- **Reproduced on exact head:** new regression test `ProduceMember_PreservesQualifiedNameInsideStringLiteral` renders `=> "a \"String\" b"` on `7e89cfdf` (fails) and `=> "a \"System.String\" b"` on the fix (passes).
- **Fix (`ddcd8e16`):** replaced the parity split with an escape-aware `ShortenLine` scanner that copies string/char literals verbatim (honoring backslash escapes) and shortens only the code spans between them.
- **Blast-radius read:** rendering System.Private.CoreLib via `ProduceMembers` (35,777 members) is **byte-for-byte identical (SHA256) baseline vs head** — the fix changes only the pathological escaped-quote/char-quote cases.

### Finding 2 — non-public method/operator targets bypass the migration (High/Med) — FIXED
`FidelityCheck.TargetApiIndex` built its token index from `ApiSurfaceExtractor.Extract(p)` (public+protected only), so internal/private targets never entered the index and silently retained the legacy signature emitter this PR replaces.

- **Fix (`ddcd8e16`):** build the index with `ApiSurfaceExtractor.Extract(p, includeAll: true)`. Non-public renders that don't complete degrade honestly (fall back to `EmitMethod`); the compile-back oracle (CI `test` job) validates the migrated set end-to-end.

### Cleared attack points
Both reviewers cleared: multi-line-string re-indent corruption (the printer escapes `\n`/`\r`/U+2028/U+2029 and never emits verbatim/raw literals), `WholeMemberNamespaces` usings-hoist propagation, resolver-parity 1:1 `RegisterSourceContext` mapping, token-index correctness, batch/per-member byte-identity, and honest degradation.

### Pre-existing limitation (not introduced here) → follow-up #3064
Gemini noted that a qualified name inside a **nested string literal within an interpolation hole** (`$"{Foo("System.String")}"`) can still be shortened. Confirmed present in the old `Split('"')` too and not surfaced by the corelib render; tracked as follow-up #3064.

### Re-review of fixed head `ddcd8e16`
Both reviewers re-reviewed the fix in isolated exact-head worktrees:
- **GPT: CLEAN** — both findings resolved, no new issues.
- **Gemini: CLEAN** — verified unterminated literals, mutually-exclusive `"`/`'` delimiters, trailing-backslash bounds, byte-identity on safe lines, and no `includeAll` signature poisoning.

Both fixed-head reviews are clean. Awaiting the CI `test` (compile-back) job on `ddcd8e16` before marking ready.

richlander:
Ready to merge

- Both fixed-head adversarial reviews (`ddcd8e16`) are CLEAN: GPT `gpt-5.6-sol` and Gemini `gemini-3.1-pro-preview`.
- CI green on `ddcd8e16`, including the `test (ubuntu-26.04)` compile-back oracle (7m52s), which validates the `ShortenLine` fix and the `includeAll:true` non-public target migration end-to-end.
- Non-blocking follow-up: #3064 (pre-existing interpolation-hole nested-string shortening, not introduced by this PR).

PR #3063, opened 2026-07-23
Duplicate scattered switch default into each guard (#2978)

### LINKED ISSUE #2978: Decompiler: nested type-pattern switch dispatch leaves an unassigned out path (CS0177)
Split out from adversarial review of #2972 (round-3 GPT finding). **Pre-existing** on `main` — reproduced identically on base `86bd05c4` and on #2972's head `f0586db8`, so it is not introduced or regressed by #2972. Filing to track the underlying limitation.

## Witness (compiled net11 Release)

```csharp
public static bool M(object o, out int x) => o switch
{
    string s => s.Length switch
    {
        0 => Fail(out x),
        1 => Win(1, out x),
        _ => Fail(out x)
    },
    int i when i > 0 => Win(i, out x),
    _ => Fail(out x)
};
static bool Win(int v, out int x) { x = v; return true; }
static bool Fail(out int x) { x = 0; return false; }
```

## Symptom

`--fidelity-check --skip-pdb` reports `W::M` as **recompile-fail**:

```
CS0177: The out parameter 'x' must be assigned to before control leaves the current method
```

The emitted source ends after the `int i when i > 0` arm's `return Win(...)`, leaving the `i <= 0` fall-through path with no return / `out x` assignment.

## Root cause (relation to #2959)

This is a scattered shared-return shape at a *deeper* nesting: the interleaved terminal's controlling guard branches to **another terminal inside** the predecessor span (a sibling dispatch arm), rather than forward past the span. #2959's `IsScatteredDispatch` span rule (`guard.TargetOffset > blocks[hi].StartOffset`) deliberately does **not** fire in that position, because firing on within-span forward branches is what produces opcode-fidelity regressions on short-circuit `&&`/ternary chains (the #640 canary; the round-2 witness `a && (b ? c : throw)`). Distinguishing a nested-dispatch within-span terminal from a short-circuit within-span terminal needs a stronger structural signal than target position alone.

## Scope note

Out of scope for #2972, which targets the 6 CS0161 self-decompilation census methods and keeps the census pin at 8. This requires a discriminator that separates nested dispatch arms from short-circuit sub-expression terminals without reintroducing the round-2 false positives.

richlander:
- Fixes #2978
- Changes structuring so a nested type-pattern `switch` whose shared default is reached by two scattered guards duplicates that default into each guard, instead of leaving one path unassigned.
- Card revision: `ffbd7ff8`

## Change

**Conclusion:** **PASS** — the fix turns invalid decompiled C# (a non-void method that falls off the end, CS0161, leaving `out x` unassigned, CS0177) into valid C# where every path returns, and the #640 short-circuit canaries stay opcode-exact.

### Original source

```csharp
public static bool Dispatch(object o, out int x) => o switch
{
    string s => s.Length switch
    {
        0 => Fail(out x),
        1 => Win(1, out x),
        _ => Fail(out x)
    },
    int i when i > 0 => Win(i, out x),
    _ => Fail(out x)
};
```

### Before

```csharp
public static bool Dispatch(object o, out int x)
{
    int i;
    int V_5;

    object V_3 = o;
    string s = V_3 as string;
    if (s is null)
    {
        if (V_3 is not int)
        {
            return Fail(out x);
        }
        i = (int)V_3;
    }
    else
    {
        V_5 = s.Length;
        if (V_5 != 0)
        {
            if (V_5 == 1)
            {
                return Win(1, out x);
            }
            return Fail(out x);
        }
        return Fail(out x);
    }
    if (i > 0)
    {
        return Win(i, out x);
    }
}
```

- Valid: False
- Correct: False

The `i <= 0` path has no `return`: the method falls off the end (CS0161) and `out x` is never assigned on that path (CS0177). The shared `_ => Fail` default was inlined into the inner `string` arm only; the outer failing `int` guard lost its path to it.

### After

```csharp
public static bool Dispatch(object o, out int x)
{
    int i;
    int V_5;

    object V_3 = o;
    string s = V_3 as string;
    if (s is null)
    {
        if (V_3 is not int)
        {
            return Fail(out x);
        }
        i = (int)V_3;
    }
    else
    {
        V_5 = s.Length;
        if (V_5 != 0)
        {
            if (V_5 == 1)
            {
                return Win(1, out x);
            }
            return Fail(out x);
        }
        return Fail(out x);
    }
    if (i <= 0)
    {
        return Fail(out x);
    }
    return Win(i, out x);
}
```

- Valid: True
- Correct: True

Every path now returns and assigns `out x`. The shared default is duplicated into the failing `int` guard (`if (i <= 0) return Fail(out x);`), matching the #2959 precedent that duplicates a scattered dispatch return to fix CS0161.

### Fully raised

The intended endpoint is the original nested `switch` expression. This PR raises it to valid, behavior-preserving statement C# but not back to the `switch`-expression spelling (the original lowering stores to a result temp and returns once; the duplicated-return structuring emits per-arm returns — `valid_different`, not opcode-exact). Full `switch`-expression reconstruction is a separate feature.

```csharp
public static bool Dispatch(object o, out int x) => o switch
{
    string s => s.Length switch
    {
        0 => Fail(out x),
        1 => Win(1, out x),
        _ => Fail(out x)
    },
    int i when i > 0 => Win(i, out x),
    _ => Fail(out x)
};
```

- Required tracking issue: #2978 — reconstruct the nested `switch`-expression spelling (result-temp single-return shape) for opcode-exact fidelity.

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `ScatteredReturnDispatchStructuringTests`: 3 passed | `ScatteredReturnDispatchStructuringTests`: 3 passed |
| Decompiler fast suite | Pass (RoundTrip 316; Pass 1207; Fidelity fast 27) | Pass (RoundTrip 316; Pass 1207; Fidelity fast 27) |
| Reduced fixture validity | self-decompilation CS0161 census pinned at 8 | 8 (unchanged) |
| Reduced fixture fidelity | 2 pre-existing environmental reds (`FidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket`, `LoweredFidelityGateTests`) | same 2 reds, unchanged |
| Real witness | `ScatteredReturnDispatchSample::Dispatch` invalid (CS0177) | fixed (valid) |

The two Slow fidelity-gate reds are the documented lambda-cache/compile-back ordinal artifacts on this machine, not a regression from this PR. Verified by dumping the raised C# of the two methods that surfaced in the current diff set (`ManualConstantOnlyComparisonFactory`, `StatementBodyLambdaInsideIf`) head-vs-base: **identical raised output**, so this structuring change does not touch them.

## Decompiler quality

**Conclusion:** **PASS** — pinned-subset gate matched baseline (0 pp on both gated metrics); aggregate residue improved by one method on two metrics with 0 pass bugs.

### PR quick gate

Run: PR quick corpus, hash-stable 100 methods per assembly; 15 assemblies, 1,500 methods; validity/fidelity not run (structural sensors).

| --- | ---: | ---: | ---: |
| Pass bugs (-) | 0 | 0 | 0 |

> **Conclusion:** **PASS** — pinned-subset gate matched baseline tolerances.

### Aggregate context

Corpus: 15 assemblies, 1,500 methods. Baseline drift: none.

| --- | ---: | ---: | ---: |
| Detected lowering residue (-) | 251 (16.73%) | 251 (16.73%) | 0 |
| Conditional-branch residue (-) | 46 (3.07%) | 45 (3.00%) | -1 |
| Forward-merge stops (-) | 32 (2.13%) | 31 (2.07%) | -1 |
| Pass bugs (-) | 0 | 0 | 0 |

> **Conclusion:** **PASS** — corpus sensor matched baseline tolerances; two residue metrics improved by one method.

## Validation

```bash
dotnet build src/dotnet-inspect -c Release
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.ScatteredReturnDispatchStructuringTests
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ScatteredReturnDispatchSample.cs
+++ b/src/ILInspector.Decompiler.Tests/ScatteredReturnDispatchSample.cs
@@ -1,0 +1,56 @@
+using System;
+
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// Fixtures for issue #2978: a nested type-pattern <c>switch</c> expression whose
+/// shared <c>_ =&gt; Fail</c> default is reached by two conditional guards at
+/// different nesting levels (the outer <c>is not int</c> and the inner
+/// <c>i &lt;= 0</c>). In block order a self-contained sibling region (the inner
+/// <c>string</c>-length switch) is interleaved between the two guards, so
+/// <see cref="ILInspector.Decompiler.Pipeline.StructuringPass"/> must recognize
+/// the guards as a scattered dispatch and duplicate the default return into each
+/// guard — otherwise the <c>i &lt;= 0</c> path falls off the end of a non-void
+/// method (CS0177: <c>out</c> parameter unassigned).
+///
+/// The <see cref="PlainAnd"/> and <see cref="ThrowTernaryChain"/> canaries are the
+/// close negatives: contiguous short-circuit <c>&amp;&amp;</c> guard chains whose
+/// shared return must stay combined under one condition (the #640 fidelity
+/// canary), not unrolled into duplicated returns.
+/// </summary>
+public static class ScatteredReturnDispatchSample
+{
+    // #2978 witness: the i <= 0 path must return the duplicated default.
+    public static bool Dispatch(object o, out int x) => o switch
+    {
+        string s => s.Length switch
+        {
+            0 => Fail(out x),
+            1 => Win(1, out x),
+            _ => Fail(out x)
+        },
+        int i when i > 0 => Win(i, out x),
+        _ => Fail(out x)
+    };
+
+    static bool Win(int v, out int x) { x = v; return true; }
+
+    static bool Fail(out int x) { x = 0; return false; }
+
+    // #640 canary: two contiguous guards on a shared return — must stay `a && b`.
+    public static int PlainAnd(int a, int b)
+    {
+        if (a > 0 && b > 0)
+        {
+            return 1;
+        }
+
+        return 2;
+    }
+
+    // Round-2 witness: a throw in a ternary inside `&&` — must stay combined.
+    public static bool ThrowTernaryChain(bool a, bool b, bool c)
+        => a && (b ? c : throw new Exception()) && Other();
+
+    static bool Other() => true;
+}
--- a/src/ILInspector.Decompiler.Tests/ScatteredReturnDispatchStructuringTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ScatteredReturnDispatchStructuringTests.cs
@@ -1,0 +1,85 @@
+using System;
+using ILInspector.Decompiler.Pipeline;
+
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// Issue #2978: a nested type-pattern <c>switch</c> expression whose shared
+/// <c>_ =&gt; Fail</c> default is reached by two conditional guards at different
+/// nesting levels, with a self-contained sibling region (an inner
+/// <c>string</c>-length switch) interleaved between them in block order.
+/// <see cref="StructuringPass"/> must classify the two guards as a scattered
+/// dispatch and duplicate the default return into each guard; otherwise the
+/// failing <c>int</c> arm (<c>i &lt;= 0</c>) falls off the end of the non-void
+/// method, leaving the <c>out</c> parameter unassigned (CS0177).
+///
+/// The discriminator that recognizes this shape (an interior block entered by a
+/// jump from before the guard span) must not fire on a contiguous short-circuit
+/// <c>&amp;&amp;</c> guard chain, which threads entirely within its own span and
+/// must stay combined under one condition — the #640 fidelity canary. The two
+/// canaries below (<see cref="ScatteredReturnDispatchSample.PlainAnd"/> and
+/// <see cref="ScatteredReturnDispatchSample.ThrowTernaryChain"/>) pin that the
+/// shared return is not unrolled and the condition stays a single <c>&amp;&amp;</c>.
+/// </summary>
+[Trait("Area", "Pass")]
+public class ScatteredReturnDispatchStructuringTests
+{
+    static IrFunction Raised(string methodName)
+    {
+        using var source = MetadataSource.Open(typeof(ScatteredReturnDispatchSample).Assembly.Location);
+        var function = IrImporter.Import(source, typeof(ScatteredReturnDispatchSample).FullName!, methodName);
+        Assert.NotNull(function);
+        IrPasses.Run(function!);
+        function!.CheckInvariant();
+        return function;
+    }
+
+    static string Print(string methodName) =>
+        CSharpPrinter.Print(Raised(methodName)).Output!.ReplaceLineEndings("\n").TrimEnd();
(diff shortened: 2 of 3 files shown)


richlander:

Two independent adversarial reviews from different model families, both against the exact reviewed head `ffbd7ff8` in isolated worktrees, with real build/decompile runs.

| Reviewer | Model family | Verdict |
| --- | --- | --- |
| GPT‑5.6 | GPT | **CLEAN** |
| Gemini 3.1 Pro | Gemini | **CLEAN** |

Neither reviewer found any high-confidence bug, soundness error, security issue, or fidelity regression. No fix commits were required; the reviewed head equals the current head.

### Attack points probed (both reviewers), verified

1. **False positive on legitimate short-circuit chains.** Probed `x ? (a && b) : (c && d)`, `(X() ? A() : B()) && C()`, `(X() || Y()) && A() && (B() || C())`, `A() && (B() || C()) && D()`. No false positives. Roslyn either evaluates complex short-circuits into a slot (so the guards never target the shared return and never reach `IsScatteredDispatch`), or earlier passes (`OrChainGuardPass`, `MixedShortCircuitChainPass`) flatten pure contiguous chains into `LogicalOr`/`LogicalAnd` before `StructuringPass` runs — leaving no discrete interior block for a `predIndex < lo` jump. Any surviving interior-from-before-`lo` block is genuinely a foreign interleaved region.
2. **#640 canaries.** `PlainAnd` and `ThrowTernaryChain` (`a && (b ? c : throw) && Other()`) stay combined (single condition, no duplicated return) and compile back **opcode-exact**. In `ThrowTernaryChain` the interior blocks are reached only by fall-through from `a` (index ≥ `lo`), so the rule correctly ignores them.
3. **Loops / back-edges.** Back-edges originate at index `> hi`, so `predIndex < lo` is trivially false. Where a `continue` legitimately jumps into the span, duplicating the shared return is the only `goto`-free structuring — behavior-preserving. Three loop probes compiled back opcode-exact.
4. **Behavior preservation.** The duplicated terminator is a side-effect-free leaf `Return`/`Throw`; no side effect is evaluated twice.
5. **Map correctness.** `jumpPredecessorIndices` correctly captures `Branch`/`ConditionalBranch`/`SwitchBranch`; unstructured `Leave` EH exits target EH-boundary offsets outside the local span and do not affect the interior check.

### Evidence

- Build: Pass. Focused tests `ScatteredReturnDispatchStructuringTests`: 3/3.
- Witness `Dispatch`: valid (every path returns); canaries opcode-exact; loop/back-edge probes opcode-exact.

Note: one reviewer flagged that the prompt's `CfgSampleClass::PlainAnd` example selector was wrong — the fixture type is `ScatteredReturnDispatchSample` (a prompt typo, not a code issue).

richlander:
Ready to merge

- CI: build-net10, csharp-diff-smoke, test (ubuntu-26.04) all pass; path-gated jobs (il-diff-smoke, markdownlint, pack) skipped as expected; PR is MERGEABLE/CLEAN.
- Adversarial review: two clean fixed-head reviews (GPT-5.6, Gemini 3.1 Pro) at `ffbd7ff8`, no findings.
- Corpus PR quick card PASS (conditional-branch residue 46→45, forward-merge 32→31, 0 pass bugs).

PR #3065, opened 2026-07-23
Raise heterogeneous switch-arm intros to pattern switch expressions (#3028 PR A)

richlander:
- Fixes/advances #3028 (PR A of A→B→C)
- Changes `PatternSwitchExpressionPass` to raise heterogeneous switch-arm intros (intro-chain arm1 + inline-positive sibling arms) over a re-evaluable scrutinee read through a temp or directly
- Card revision: `1506df38`

## Change

**Conclusion:** **PASS** — the pass now recognizes the inline-positive `if (place is T x) { … }` sibling-arm form and mixes it with the existing intro-chain form over a single scrutinee, folding the real csc shape for multi-arm bare type-pattern switches while leaving lone `is`-guards and temp-owned cascades untouched.

### Original source

Local compiled fixture (`HeterogeneousArmSample.cs`); SourceLink cannot serve a repo-built test DLL, so the authoritative C# source is quoted verbatim from the fixture:

```csharp
public static int Area(Shape shape) => shape switch
{
    Dot d => d.Radius,
    Bar b => b.Length,
    Box x => x.Side,
    _ => -1,
};
```

### Before

`dotnet-inspect member <fixture.dll> HeterogeneousArmSample Area -S "Decompiled Source"` at base `abe0eef6`:

```csharp
public static int Area(ILInspector.Decompiler.Tests.HeterogeneousArmSample.Shape shape)
{
    Dot d = shape as Dot;
    if (d is null)
    {
        if (shape is Bar b)
        {
            return b.Length;
        }
        if (shape is Box x)
        {
            return x.Side;
        }
        return -1;
    }
    return d.Radius;
}
```

- Valid: True
- Correct: True

Lower-quality: the switch is left as an unraised intro-chain + inline-sibling `if` cascade.

### After

`dotnet-inspect member <fixture.dll> HeterogeneousArmSample Area -S "Decompiled Source"` at head `1506df38`:

```csharp
public static int Area(ILInspector.Decompiler.Tests.HeterogeneousArmSample.Shape shape)
{
    return shape switch
    {
        Dot d => d.Radius,
        Bar b => b.Length,
        Box x => x.Side,
        _ => -1,
    };
}
```

- Valid: True
- Correct: True

### Fully raised

## Evidence

| Check | Baseline (`abe0eef6`) | Head (`1506df38`) |
| --- | --- | --- |
| Product build (`dotnet-inspect`) | Pass | Pass |
| Focused tests (`PatternSwitchExpressionPassTests`) | n/a (fixture/tests new) | Pass |
| Decompiler fast suite (`-trait- Speed=Slow`) | 1206, 0 failed | 1211, 0 failed |
| Full decompiler slow suite | 1 failed (see below) | 3494 total, 1 failed (see below) |
| Reduced fixture validity/fidelity | Pass | Pass |

The sole full-suite failure is `FidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket`, an ordinal-only compiler cached-delegate closure-field diff (`CfgSampleClass+<>c::<>9__N_0`, identical opcodes) on `StatementBodyLambdaInsideIf`. It reproduces **identically on untouched `origin/main`** (verified by running the same test on a clean `abe0eef6` worktree — there it flags both `ManualConstantOnlyComparisonFactory` and `StatementBodyLambdaInsideIf`), so it is pre-existing and unrelated to this change. It is a `[Trait("Speed","Slow")]` gate that CI excludes from the PR path (`ci.yml` runs `-trait- "Speed=Slow"`), so it does not block this PR. All corpus-sweep / cluster-capture / compile-back slow gates pass on head.

## Decompiler quality

**Conclusion:** **PASS** — no new fidelity diffs beyond the pre-existing docket; the change is additive (raises a shape previously left unraised) and default-gated by the same ownership/re-evaluability checks as the existing temp form. `PlaceFromAddress` still does not fold (its recursive property-pattern arm fails the chain parse — deferred to PR B).

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -trait "Area=Pass" -trait- "Speed=Slow"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/HeterogeneousArmSample.cs
+++ b/src/ILInspector.Decompiler.Tests/HeterogeneousArmSample.cs
@@ -1,0 +1,47 @@
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// Fixture for issue #3028 PR A: a <c>switch</c> expression over a plain receiver
+/// whose arms csc lowers to <em>heterogeneous</em> intros over the place read
+/// directly — no switch-value temp. csc emits the first arm as an intro-chain
+/// (<c>x = place as T; if (x is null) { … }</c>) and the remaining arms as
+/// inline-positive siblings (<c>if (place is U y) { … }</c>) nested in the
+/// leading arm's no-match branch, bottoming out in the default
+/// <c>return</c>. <see cref="ILInspector.Decompiler.Pipeline.PatternSwitchExpressionPass"/>
+/// raises the whole cascade back into a single switch expression.
+///
+/// <see cref="Area"/> is the mixed intro-chain-plus-inline shape. Isolated from
+/// <see cref="CfgSampleClass"/> so it does not perturb the fidelity gate's pinned
+/// population.
+/// </summary>
+public static class HeterogeneousArmSample
+{
+    public abstract class Shape
+    {
+    }
+
+    public sealed class Dot : Shape
+    {
+        public int Radius;
+    }
+
+    public sealed class Bar : Shape
+    {
+        public int Length;
+    }
+
+    public sealed class Box : Shape
+    {
+        public int Side;
+    }
+
+    // csc: intro-chain arm (`Dot d = shape as Dot; if (d is null) …`) followed by
+    // inline-positive sibling arms, over `shape` read directly.
+    public static int Area(Shape shape) => shape switch
+    {
+        Dot d => d.Radius,
+        Bar b => b.Length,
+        Box x => x.Side,
+        _ => -1,
+    };
+}
--- a/src/ILInspector.Decompiler.Tests/PatternSwitchExpressionPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/PatternSwitchExpressionPassTests.cs
@@ -100,6 +100,39 @@ public class PatternSwitchExpressionPassTests
         Assert.Contains("_ => false,", output);
     }
 
+    // ── Compiler-backed positives: heterogeneous / inline arm intros (#3028) ─
+
+    [Fact]
+    public void HeterogeneousArmIntros_RaiseToPatternSwitchExpression()
+    {
+        // csc lowers this to a leading intro-chain arm (`Dot d = shape as Dot; if
+        // (d is null) …`) plus inline-positive sibling arms (`if (shape is Bar b)
+        // …`), over `shape` read directly (no switch-value temp). The pass folds
+        // the whole heterogeneous cascade back into one switch expression.
+        var function = Raised(typeof(HeterogeneousArmSample), nameof(HeterogeneousArmSample.Area));
+
+        var switchExpression = Assert.Single(function.Descendants.OfType<PatternSwitchExpression>());
+        Assert.Empty(function.Descendants.OfType<IfStatement>());
+        Assert.True(switchExpression.HasDefault);
+        Assert.Equal(3, switchExpression.Arms.Count);
+        Assert.All(switchExpression.Arms, arm =>
+        {
+            Assert.Null(arm.Subpattern);
+            Assert.False(arm.HasGuard);
+            Assert.NotNull(arm.LocalIndex);
+        });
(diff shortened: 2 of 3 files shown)


richlander:
## Adversarial review reconciliation (round 2)

Two-model adversarial review, each isolated in its own linked worktree at the exact fixed head, scratch under `/tmp/` only:

- **GPT** (`gpt-5.6-sol`)
- **Gemini** (`gemini-3.1-pro-preview`)

### Findings and resolutions

**1. Guard-fail-to-default divergence on the newly recognized surface** — *GPT Finding A / Gemini Finding 1, including Gemini's base-type later-arm variant.*
A guarded arm whose guard failure short-circuits to the trailing default folds faithfully only when no later arm can match the same value; in a folded switch a failed `when` guard falls through to later arms instead. Proving no overlap needs a type-disjointness oracle this SRM-only, no-inspected-assembly-loading pass cannot have, and a blanket "reject short-circuit guarded arm + later arm" rule would break the shipped `Classify` positive.
**Resolved** in `39905831` by scoping the *new* surface — a direct (temp-less) scrutinee, or any inline-positive arm — to **unguarded arms only** (`isNewSurface && arms.Any(a => a.Guard is not null)` ⇒ decline). This removes all guard-fail divergence the new paths could introduce without an oracle. Verified: `folds=1 → 0` on direct/inline probes (negated and positive forms, identical and base-type later arms).

**2. Property-subpattern failure-to-default on the new surface** — *GPT, re-review of `39905831`.*
The gate initially checked only `Guard`, but a single-level property-subpattern arm (`Outer { Inner: T }`) routes its inner-match *failure* to the default exactly like a failed guard.
**Resolved** in `a72d5b2a` by extending the gate to decline any arm carrying a guard **or** a property subpattern on the new surface, leaving only plain unguarded type-pattern arms (which have no failure-to-default routing). Verified: `folds=1 → 0`; negative test `Synthetic_DirectSubpatternOverlapsLaterArm_DoesNotRaise` added.

**3. Aliased / direct / temp scrutinee mutation between arm reads** — *GPT Findings B/C, Gemini Finding 2.*
A direct scrutinee re-read per arm, or a temp scrutinee, could be mutated via an address taken before/inside the cascade (`&arg`, `&local`, `&SV`) through a by-ref call.
**Resolved** in `63688f6a` and retained: `DirectScrutineeStable` performs a function-wide address-of scan (not just the consumed region); `TempScrutineeReadOnly` rejects a re-store or address-of the temp. Both reviewers re-verified these reject on this head.

**4. Inline-arm guard short-circuit** — *round-1 (Gemini).* Resolved in `63688f6a`; inline arms are accepted only in fall-through shapes (and, per finding 1, unguarded on the new surface).

### Non-blocking observation (both reviewers, by design)

The **pre-existing** temp-form intro cascade (#3022, the shipped `Classify` shape) folds a guarded/base-type overlap under the compiler's proven mutual-exclusivity trust. Both reviewers confirmed **this PR does not widen that path** (`isNewSurface` is `false` for it). Verifying it would require the same disjointness oracle; it is deferred to a follow-up, consistent with #3022's documented trust boundary.

### Final state

Both fixed-head reviews at **`a72d5b2a`** are **clean** (0 blocking). Real-run evidence from both: guarded/subpattern direct+inline probes `folds=0`, alias/temp-mutation probes rejected, shipped positives `Classify` and `Area` still fold, fast Pass suite **1220/1220** green.

`origin/main` was then merged (`426340de`) — a clean integration of the unrelated `StructReceiverInliningPass`; the reviewed pass diff is byte-identical (`git diff a72d5b2a..HEAD` on the pass/test files is empty), and the Pass suite is **1229/1229** green post-merge.

richlander:
Ready to merge

Both fixed-head adversarial reviews at `a72d5b2a` are clean (GPT `gpt-5.6-sol`, Gemini `gemini-3.1-pro-preview`); reconciliation above. `origin/main` merged cleanly (`426340de`, unrelated `StructReceiverInliningPass`) with the reviewed pass diff byte-identical, and all required CI checks (build-net10, test, csharp-diff-smoke, changes) pass on the merge commit. Fast Pass suite 1229/1229 green post-merge.

PR #3075, opened 2026-07-23
RTS: coherent Full reconstruction of plain event-accessor targets (#3007)

richlander:
- Advances #3007
- Changes: RTS harness-only. Plain (non-explicit-interface) event-accessor targets now reconstruct coherently under Full policy. No product output change.
- Evidence revision: 2f5970de

## Change

**Conclusion:** **PASS** — a plain event-accessor target under Full was an uncheckable `RecompileFail`/floor bucket (CS0082); routing it through `ComposeEventAccessor` with the full member surface + symmetric evidence crediting makes it a coherent `event { add remove }` with every declaration represented and recompile Exact.

Before (target `Holder.add_Changed`, Full): standalone accessor method **plus** a re-declared skeleton event → CS0082 collision → compile-back floor, `BodyComplete=false` (honest floor from #3000):

```csharp
public class Holder {
    public void add_Changed(Action value) { _changed = (Action)Delegate.Combine(_changed, value); } // CS0082 vs event's synthesized add
    public Action _changed;
    public event Action Changed { add { throw null; } remove { throw null; } }
    public Holder() { }
}
```

After: one coherent event, both real bodies, no standalone accessor method, `add_Changed`/`remove_Changed`/`.ctor` all Complete, `BodyComplete=true`, no floor:

```csharp
public class Holder {
    public event Action Changed {
        add { _changed = (Action)Delegate.Combine(_changed, value); }
        remove { _changed = (Action)Delegate.Remove(_changed, value); }
    }
    public Action _changed;
    public Holder() { }
}
```

## Scope and safety

- **Root cause:** `TryFindEventAccessor` only matched explicit-interface accessors, so a plain event accessor fell through to method-routing (`ComposeMethod`). Under Full the type surface re-declared the event, whose compiler-synthesized accessor collided (CS0082) with the standalone target method.
- **Fix shape:** RTS orchestration + surface/evidence symmetry with the merged property-getter fix (#3008). Three parts:
  1. `ReturnToSender.TryFindEventAccessor` routes plain event accessors through `ComposeEventAccessor` **under Full only**.
  2. `ComposeEventAccessor` sets `IncludeMemberSurface` (**gated to Full**) when the accessor closure pulls in same-type members — the surface enumeration folds the sibling accessor's token into the single event requirement, skips the standalone accessor methods (no CS0082), and represents the `.ctor`.
  3. `ApplyFullBodies.Enrich` event branch mirrors the property `IsTargetToken` preserve-base logic: the target accessor keeps its real base body (instead of being clobbered with `throw null`); only the sibling is produced, which also credits it into `attempted`.
- **Product impact:** none. Change is entirely within `tools/DecompilerHarness` (RTS) + its test.
- **Scope boundary:** Selected/default stays method-routed; explicit-interface event targets under Selected keep their pre-existing minimal single-accessor shape (`IncludeMemberSurface` is Full-gated), so the A/B corpus baseline is unchanged.
- **RTS boundary:** RTS only orchestrates closure + Roslyn + contract body comparison; the product/shared `ILInspector.CSharp` printer owns C# source generation. No C# knowledge was added to RTS.

## Evidence

| Check | Result |
| --- | --- |
| Product build (test graph) | Pass (`ILInspector.Decompiler.Tests` clean) |
| Focused test | `CompileBackTargets_FullReconstructsPlainEventAccessorTarget` — source-shape / `FullBodies` (adder, remover, `.ctor` all Complete) / `BodyComplete` / no-floor all pass locally |
| Regression sweep | Full `ReturnToSenderPrototypeTests` baseline (b5e9ee3b) vs branch: **only delta is the renamed/converted test**; all 118 other reds are byte-identical, pre-existing environmental `CS0009` recompile stalls |
| Recompile Exact (`DonorPe`/`Comparison`) | CI-only — blocked locally by the known environmental `CS0009` on `aspnetcorev2_inprocess.dll` (affects the whole class equally, baseline and branch) |

Local recompile-Exact assertions (`DonorPe`, `Comparison.Status`) fail on this box for **every** recompile-based test in the class (baseline included) due to the CS0009 environmental stall; they are green on CI. The regression sweep isolates that the fix adds zero new failures.

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| Plain event-accessor target under Full | Fixed | This PR |
| Explicit-interface event target under Full | Fixed incidentally | `IncludeMemberSurface` now represents its sibling/`.ctor` too (previously had the same evidence gap) |
| Plain event-accessor target under Selected | Left as-is | Method-routed by design; Selected never runs the full-body evidence pass |

## Review

Two-model adversarial review pending (Gemini Pro + GPT, per AGENTS.md roster).

## Validation

```powershell
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -method "ILInspector.Decompiler.Tests.ReturnToSenderPrototypeTests.CompileBackTargets_FullReconstructsPlainEventAccessorTarget"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -782,14 +782,15 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
-    public void CompileBackTargets_FullEventAccessorTargetSurfacesRecompileFailure()
+    public void CompileBackTargets_FullReconstructsPlainEventAccessorTarget()
     {
-        // Issue #3000: a plain (non-explicit-interface) event accessor target is method-routed, so
-        // under Full policy the reconstructed type surface re-declares the event, whose synthesized
-        // accessor collides (CS0082) with the standalone target method. The Full closure fails to
-        // recompile and falls back to the compile-back floor. This must be reported honestly:
-        // BodyComplete is false (the Full reconstruction did not compile), not masked as Complete.
-        // Coherent Full reconstruction of plain event-accessor targets is a separate follow-up.
+        // Issue #3007 (follow-up to #3000/#3008): a plain (non-explicit-interface) event accessor
+        // target under Full policy reconstructs a coherent single `event { add remove }` carrying
+        // both real accessor bodies, rather than a standalone accessor method that collides
+        // (CS0082) with the re-declared event's compiler-synthesized accessor. Routing the plain
+        // accessor through ComposeEventAccessor with the full member surface folds the sibling
+        // accessor into the event and represents the constructor, so every concrete declaration is
+        // accounted for and BodyComplete is honestly true.
         var assemblyPath = CompileFixture("""
             using System;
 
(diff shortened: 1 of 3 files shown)


richlander:
## Adversarial review reconciliation (round 1)

Two-model review of the original head `2f5970de`:

- **GPT (gpt-5.6-sol): 2 blocking findings** — both reproduced and fixed in `13d9cb9a`.
- **Gemini (gemini-3.1-pro-preview): no blocking findings.** Gemini validated the 6 attack points it was given (NRE in the `Enrich` event branch, target-body survival, CS0082 elimination, Selected-path safety, non-target events, evidence integrity) — all sound. That prompt did not include the two event-target shapes GPT exercised, so its clean pass does not cover them; GPT's concrete repros stand.

### Finding 1 — field-like event target double-declaration (GPT)

`event Action Changed;` has a compiler-generated backing field named identically to the event. The Full member surface emitted it as a separate `Action Changed;` next to the reconstructed event (CS0102/CS0229) while reporting `BodyComplete=true`.

**Verified pre-existing, not introduced by this PR.** I restored the two product files to the branch base `b5e9ee3b` and re-ran the repro: the base already produces the same `BodyComplete=true` double declaration (both accessors `throw null`). This PR now **excludes field-like events** from the Full broadening (`IsFieldLikeEvent` in `ReturnToSender.TryFindEventAccessor`), so field-like accessors stay method-routed exactly as at the base — the PR is neutral on them. The underlying pre-existing member-surface bug is tracked as #3083.

### Finding 2 — explicit-interface event target double-declaration (GPT)

The surface folds events by the sanitized full metadata name (`IBaseEvents.Changed`) while the target requirement carries the stripped identity (`Changed`); the fold missed and appended a second explicit-interface event with a `throw null` accessor (CS8646/CS0102), again with `BodyComplete=true`.

**This was a real regression from the original commit and is fixed.** `ComposeEventAccessor` now declines the member surface for explicit-interface targets (`IncludeMemberSurface` additionally requires `explicitEvent is null`), restoring the pre-#3007 single-accessor shape. Reproduced on the fixed head: exactly one `event Action IBaseEvents.Changed`, no `throw null` duplicate, and honest `BodyComplete=false` (sibling remover + `.ctor` flagged not-represented) instead of the previous false success.

### Fix commit `13d9cb9a`

- Exclude field-like events from the Full broadening (`IsFieldLikeEvent`); they stay method-routed = baseline.
- Decline the member surface for explicit-interface event targets (`explicitEvent is null` gate).
- New regression tests: `CompileBackTargets_FullExplicitInterfaceEventTargetDoesNotDoubleDeclare` (single decl + no `throw null` + honest `BodyComplete=false`) and `CompileBackTargets_FullFieldLikeEventTargetStaysMethodRouted` (method-routed exclusion).

### Validation

Full `ReturnToSenderPrototypeTests` class on the fixed head: **123 failed / 151** — identical failing set to the base `b5e9ee3b` (123), i.e. zero new failures; the 123 are the known-environmental CS0009 recompile stalls (`aspnetcorev2_inprocess.dll` / `coreclr.dll`, CI-green). The two new regression tests pass; the #3007 target test (`CompileBackTargets_FullReconstructsPlainEventAccessorTarget`) passes every source-shape/`BodyComplete` assertion and only trips the CS0009 `DonorPe` line locally.

Both reviewers are re-reviewing the fixed head `13d9cb9a` now; I will not mark this ready until both fixed-head reviews are clean and CI is green.

richlander:
## Adversarial review reconciliation (round 2 — fixed head `13d9cb9a`)

Both reviewers re-reviewed the fixed head `13d9cb9a`; **both are clean**:

- **GPT (gpt-5.6-sol):** "No significant issues found in the reviewed changes."
- **Gemini (gemini-3.1-pro-preview):** no blocking findings; independently verified both fixes:
  - `IsFieldLikeEvent` is sound — C# forbids a same-named field+event in a type (CS0102), so the only IL case where a field shares an event's name is the compiler-generated field-like backing field; field-like events are routed out of the coherent surface without touching custom events.
  - The `explicitEvent is null` gate purges the second explicit-interface declaration and yields the honest `BodyComplete=false` floor.
  - No over-exclusion: a custom event's backing field is uniquely named (`_changed != Changed`), so the #3007 coherent single-event surface still activates for the intended shape (`CompileBackTargets_FullReconstructsPlainEventAccessorTarget`).

CI `test (ubuntu-26.04, linux-x64)` is **green** on `13d9cb9a`.

Follow-up (non-blocking): the pre-existing field-like member-surface double-declaration is tracked in #3083; this PR is neutral on field-like events.

Ready to merge

PR #3077, opened 2026-07-23
Model interpolation holes when shortening qualified names (#3064)

### LINKED ISSUE #3064: ShortenQualifiedNames: qualified name inside an interpolation-hole nested string literal can still be shortened
## Summary

`MemberBodyProducer.ShortenQualifiedNames` (via the new `ShortenLine` scanner) skips string (`"…"`) and char (`'…'`) literals when shortening qualified type names, honoring backslash escapes. However, it does not model **nested string literals inside an interpolation hole**. For a rendered interpolated string such as:

```csharp
$"hello {Foo("System.String")}"
```

the scanner treats the outer `$"…"` as one string literal and stops at the first inner `"`, so the hole's code (and any nested literal) is mis-segmented. A fully-qualified name that lands in a nested string literal inside a hole can still be shortened, corrupting the constant.

## Status

- This is a **pre-existing limitation**: the previous `line.Split('"')` parity implementation exhibited the exact same behavior. It was **not** introduced by #3062 (which fixed the escaped-quote/`'"'` parity-flip cases and is byte-for-byte identical to baseline across a 35,777-member System.Private.CoreLib render).
- Reachability is narrow: it requires the decompiler to re-sugar an interpolated string whose hole contains a nested string literal that itself contains a shortenable `Namespace.Type`. The corelib blast-radius render surfaced zero such cases.

## Fix direction

Make `ShortenLine` interpolation-aware: when it enters a `$"…"` (or `$@"…"`) string, track brace depth and, inside a hole (`{ … }`), recurse into normal code/literal scanning so nested literals are skipped and hole code is shortened. Guard the `{{`/`}}` escapes and the format-clause `:`.

## Evidence

Adversarial re-review of #3062 (Gemini `gemini-3.1-pro-preview`) flagged this as a known pre-existing limitation while giving the #3062 fix a CLEAN verdict.

richlander:
## Summary

`MemberBodyProducer`'s qualified-name shortener (which rewrites fully-qualified
type names in already-emitted decompiled C# to simple names) had two coupled
defects. Both are **pre-existing** (not introduced by #3062) and are fixed here.
Fixes #3064.

## Fix 1 — model interpolation holes (`ShortenLine`)

`ShortenLine` skips string/char literals when shortening, but treated an
interpolated string (`$"…"`) as one opaque literal: it stopped at the first inner
`"`, mis-segmenting the hole. A fully-qualified name inside a **nested string
literal in a hole** — e.g. `$"{Echo("System.String")}"` — was then exposed to
shortening, corrupting the `ldstr` constant (`System.String` → `String`), which
changes the operand and can induce a false compile-back `OperandDiff`.

Make `ShortenLine` interpolation-aware:

- `CopyInterpolatedString` models the `$"…"` grammar structurally — literal text
  (with `{{`/`}}` brace escapes) is copied verbatim; each hole is handed to
  `ShortenHole`.
- `ShortenHole` scans the hole's expression as **code** (shortening qualified
  names, skipping nested string/char literals and nested `$"…"`), while copying
  the format clause after an unnested `:` verbatim. Brace/paren/bracket nesting
  is tracked so an object-initializer brace or a parenthesized conditional's `:`
  neither closes the hole nor masquerades as the format separator.
- `CopyLiteral` factors out the existing backslash-aware literal copy.

The soundness of "an unnested `:` at depth 0 is the format separator" rests on the
product printer always parenthesizing a conditional/switch inside a hole:
`InterpolatedExpression` renders each hole with `.At(Precedence.Primary)` (no
format) or `.At(Precedence.NullCoalescing)` (with format), both strictly above
`Precedence.Conditional`, so a ternary/switch `:` is always at depth > 0.

## Fix 2 — preserve alias-rooted (`global::`) names (`ShortenSegment`)

Surfaced and progressively hardened across adversarial review.
`ShortenSegment`'s word-boundary check declined only when the matched prefix was
preceded by a letter/digit/`_`/`.` — **not** by the `::` alias qualifier. The
printer emits `global::`-qualified names when a type name is shadowed in scope,
and shortening any part of such a chain re-introduces the collision it was emitted
to avoid and does not bind:

- `global::System.Runtime.CompilerServices.Unsafe` → `global::Unsafe` (CS0400).
  This occurs in real output — see blast-radius.
- With a keyword namespace segment the printer `@`-escapes it, and the
  `System.`-stripped prefix can match **mid-chain**:
  `global::System.@event.Models.X` (stripped `event.Models` matches after
  `System.@`) → `global::System.@X` (CS0234).

Because only the stripped prefix can match mid-chain, checking the characters
immediately before the match is insufficient. The fix walks back over the whole
qualified run (identifier chars, `.`, and the `@` keyword escape) to the token
that roots it, and declines when that token is the two-char `::`.

## Evidence

- **Regression tests** (`MemberBodyProducerMemberRenderTests`, 12 green):
  - `ProduceMember_PreservesQualifiedNameInsideInterpolationHoleLiteral` fails
    pre-Fix-1 (`"System.String"` → `"String"` inside the hole); the specimen
    re-sugars to a real `$"…"` shape (asserted).
  - `[Theory] ProduceMember_PreservesAliasQualifiedNameUnderShadowing` — plain and
    in-hole `global::System.Math`-under-shadowing (fails pre-Fix-2 as
    `global::Math`).
  - `[Theory] ProduceMember_PreservesAliasQualifiedNameWithEscapedKeywordNamespace`
    — `global::@event.Models.TypeNameShadow` and the System-rooted
    `global::System.@event.Models.SystemNameShadow` (both fail without the
    full-chain walk as `global::@…`).
- **Blast-radius (whole-assembly render diff vs `origin/main`):** rendering
  `System.Private.CoreLib` (11.0 preview, 321k lines) through the product
  `Project` path differs in **exactly one type, `System.Index`**, and the change
  is **invalid → valid C#**: base emitted `global::Unsafe.Unbox<Index>(value)`
  (CS0400) plus a coupled spurious `using System.Runtime.CompilerServices;`; head
  emits `global::System.Runtime.CompilerServices.Unsafe.Unbox<Index>` and no stray
  using. Every other line is byte-identical, and the alias-root hardening commits
  are each SHA256-identical to one another on corelib (they only add declines for
  `@`-escaped alias chains, which corelib does not contain).
- The only `Area=RoundTrip` reds are **31 pre-existing environmental**
  `ReturnToSenderFixtureCatalogTests` compile-back failures (identical count
  baseline vs head; known Windows native-reference issue), unrelated to this change.

## Compatibility

Behavior-preserving for all valid output. The only whole-corelib change replaces
one invalid `global::`-shortened render with the correct fully-qualified name.
Product paths remain SRM-only / Roslyn-free.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/EscapedKeywordNamespaceFixture.cs
+++ b/src/ILInspector.Decompiler.Tests/EscapedKeywordNamespaceFixture.cs
@@ -1,0 +1,10 @@
+namespace @event.Models;
+
+// A helper type whose namespace's first segment is the C# keyword 'event', so
+// the decompiler's printer escapes it as @event.Models when spelling a
+// fully-qualified reference. Used by MemberRenderSpecimen to exercise the
+// alias-qualified shortening guard against an @-escaped namespace (#3064 review).
+public static class TypeNameShadow
+{
+    public static int M(int x) => x;
+}
--- a/src/ILInspector.Decompiler.Tests/MemberBodyProducerMemberRenderTests.cs
+++ b/src/ILInspector.Decompiler.Tests/MemberBodyProducerMemberRenderTests.cs
@@ -180,23 +180,29 @@ public sealed class MemberBodyProducerMemberRenderTests
         Assert.DoesNotContain("global::Math", rendered.Text);
     }
 
-    [Fact]
-    public void ProduceMember_PreservesAliasQualifiedNameWithEscapedKeywordNamespace()
+    [Theory]
+    [InlineData(nameof(MemberRenderSpecimen.EscapedAliasQualifiedShadow),
+        "global::@event.Models.TypeNameShadow", "global::@TypeNameShadow")]
+    [InlineData(nameof(MemberRenderSpecimen.SystemEscapedAliasQualifiedShadow),
+        "global::System.@event.Models.SystemNameShadow", "global::System.@SystemNameShadow")]
+    public void ProduceMember_PreservesAliasQualifiedNameWithEscapedKeywordNamespace(
+        string memberName, string expectedFullPath, string corruptedForm)
     {
         var type = Specimen();
-        var member = Assert.Single(type.Members, m => m.Name == nameof(MemberRenderSpecimen.EscapedAliasQualifiedShadow));
+        var member = Assert.Single(type.Members, m => m.Name == memberName);
 
         var rendered = MemberBodyProducer.ProduceMember(type, member, AssemblyPath, pdbPath: null);
 
         Assert.Equal(MemberBodyProductionStatus.Complete, rendered.Status);
-        // The namespace's first segment is a keyword, so the printer escapes it:
-        // global::@event.Models.TypeNameShadow. The '@' sits between the '::'
-        // alias qualifier and the matched metadata namespace (event.Models), so
-        // the guard must skip the escape and still decline — not strip to the
-        // invalid global::@TypeNameShadow (a stray escape on a name that does
-        // not bind, CS0400) (#3064 review).
-        Assert.Contains("global::@event.Models.TypeNameShadow", rendered.Text);
-        Assert.DoesNotContain("global::@TypeNameShadow", rendered.Text);
+        // A namespace segment that is a keyword is printed @-escaped, so an '@'
+        // sits between the '::' alias qualifier and the matched metadata
+        // namespace. For a System-rooted namespace the System.-stripped prefix
+        // even matches mid-chain (after "System.@"), so the guard must walk the
+        // whole qualified run back to its '::' root and decline — not strip to
+        // the invalid @-escaped form (a stray escape on a name that does not
+        // bind, CS0400/CS0234) (#3064 review).
+        Assert.Contains(expectedFullPath, rendered.Text);
+        Assert.DoesNotContain(corruptedForm, rendered.Text);
     }
 }
 
@@ -251,5 +257,14 @@ public sealed class MemberRenderSpecimen
     // to skip the escape (#3064 review).
     public static int EscapedAliasQualifiedShadow(int TypeNameShadow)
         => @event.Models.TypeNameShadow.M(TypeNameShadow);
+
+    // Same hazard, but the namespace is System-rooted with a keyword segment
+    // (System.@event.Models). The printer emits global::System.@event.Models.
+    // TypeNameShadow; the System.-stripped prefix "event.Models" matches
+    // mid-chain after "System.@", so a guard that only inspects the characters
+    // just before the match cannot see the '::' root. Shortening must still be
+    // declined, not corrupted to global::System.@TypeNameShadow (CS0234).
+    public static int SystemEscapedAliasQualifiedShadow(int SystemNameShadow)
+        => System.@event.Models.SystemNameShadow.M(SystemNameShadow);
 }
 #pragma warning restore CA1822
--- a/src/ILInspector.Decompiler.Tests/MemberBodyProducerMemberRenderTests.cs
+++ b/src/ILInspector.Decompiler.Tests/MemberBodyProducerMemberRenderTests.cs
@@ -161,6 +161,24 @@ public sealed class MemberBodyProducerMemberRenderTests
         // must survive, not be mis-segmented and shortened to String (#3064).
         Assert.Contains("\"System.String\"", rendered.Text);
     }
+
+    [Theory]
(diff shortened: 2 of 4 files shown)


richlander:

Two-model review per `AGENTS.md`: **GPT** (gpt-5.6-sol) and **Gemini** (gemini-3.1-pro-preview). Both reviewers signed off **CLEAN on the exact final head `46c03704`** (CI: build/test/csharp-diff-smoke green on that SHA). Every finding is attributed and resolved below.

### GPT (gpt-5.6-sol) — 3 real findings, all fixed; final head clean

| # | Head found | Finding | Resolution |
| --- | --- | --- | --- |
| 1 | `603c3357` | `global::System.Math` shortened to `global::Math` (CS0400) — word-boundary check ignored the `::` alias qualifier. | Fixed by **`c23128f4`** (decline when the match is immediately preceded by `::`). |
| 2 | `c23128f4` | Keyword namespace `global::@event.Models.X` shortened to `global::@X` — the printer `@`-escapes the keyword segment, so `@` sits between `::` and the match. | Fixed by **`00d53869`** (skip an optional `@` before the `::` test). |
| 3 | `00d53869` | System-rooted keyword namespace `global::System.@event.Models.X` shortened to `global::System.@X` (CS0234) — the `System.`-stripped prefix `event.Models` matches **mid-chain** after `System.@`, past the `@`-skip. | Fixed by **`46c03704`** (replace the narrow immediate checks with a backward walk over the whole qualified run to its rooting token; decline when that token is `::`). Subsumes fixes 1 and 2. |

Reproduced each through the real `MemberBodyProducer.ProduceMember` path before fixing; each has a regression specimen. GPT's re-review of `46c03704`: *"Both prior findings are resolved, including interpolation-hole cases. Regression suite passed (12/12). No significant issues found."*

### Gemini (gemini-3.1-pro-preview) — 2 findings, both dismissed with evidence; final head clean

- **Issue 1 (`603c3357`):** same `global::` root cause as GPT #1 — resolved by `c23128f4`.
- **Issue 2 (`603c3357`), unparenthesized ternary `:` inside an interpolation hole — DISMISSED (false premise).** It assumed `ICSharpCode.Decompiler`'s printer; this product does not reference it (grep-confirmed). The product's own `CSharpPrinter.InterpolatedExpression` renders every hole at `.At(Precedence.Primary)`/`.At(Precedence.NullCoalescing)`, both strictly above `Conditional`, so a hole ternary is always parenthesized (its `:` at depth > 0); empirically csc also spills conditionals to a temp local before the handler append, so a `Conditional` node never reaches a hole. **Gemini retracted.**
- **Turn-3 "High" (`46c03704`), non-alias mid-chain escaped `System.@event.Models.X` -> `System.@X` — DISMISSED as a product-level false positive.** The evidence was a synthetic `ShortenSegment` reflection call fed the escaped string `System.@event.Models.X` with a hand-built prefix list. The product printer never emits that in a non-alias context: raw metadata namespaces carry no `@`, and the non-alias path prints the namespace **unescaped** (`System.event.Models.…`), so the full prefix matches at the head and shortens cleanly. I verified through real `ProduceMember` — a non-shadowed reference to a type in `System.@event.Models` renders `SystemNameShadow.M(n)`, **byte-identical with and without the exploratory guard** (the guard was a no-op and was reverted; head unchanged at `46c03704`). Escaped mid-chain text only appears in the `global::` path, where the backward walk already declines it. **Gemini explicitly retracted**, final verdict *"the PR is CLEAN and safe to merge."*

### Evidence

- **Blast-radius (whole-assembly render diff vs `origin/main`):** rendering `System.Private.CoreLib` (321k lines) through `Project` differs in exactly one type, `System.Index`, and the change is **invalid -> valid** (`global::Unsafe.Unbox<Index>` (CS0400) + a stray `using` -> `global::System.Runtime.CompilerServices.Unsafe.Unbox<Index>`). All alias-root hardening commits are SHA256-identical to one another on corelib.
- 12/12 `MemberBodyProducerMemberRenderTests` green; only `Area=RoundTrip` reds are the 31 pre-existing environmental `ReturnToSenderFixtureCatalogTests` compile-back failures (identical baseline vs head).

### Non-blocking follow-up

While investigating Gemini's dismissed finding I surfaced a **separate, pre-existing** bug: `HoistUsings` emits invalid `using System.event.Models;` (keyword segment not `@`-escaped) for keyword-segment namespaces — invalid regardless of the shortener, unreachable in any shipping assembly. Filed as **#3090**. Out of scope here.

richlander:
Ready to merge

Issue #3011, opened 2026-07-22
Decompiler: enum-typed value shifted (>>/<<) renders 'enum >> n' without underlying-integer cast -> CS0019


richlander:
Independent of #2990. A shift applied to an enum-typed value decompiles to
`enumValue >> n`, which is CS0019 (`operator '>>' cannot be applied to operands
of type 'E' and 'int'`) — C# has no predefined shift operator for enum types;
the enum must be cast to its underlying integer first.

Witness: `MySqlConnector.Protocol.Payloads.HandshakeResponse41Payload.CreateCapabilitiesPayload`
(`--package MySqlConnector@2.6.1`) renders:

```csharp
ProtocolCapabilities clientCapabilities = ...;
// ...
writer.Write((int)(clientCapabilities >> 32));   // CS0019: enum >> int
```

This is present identically on and before the #2990 fix (the `clientCapabilities`
local is enum-typed in both), so it is pre-existing and unrelated to the
flags-enum `BinaryResult` typing change.

Minimal repro:

```csharp
[Flags] enum E : long { A = 1 }
static int F(E e) => (int)(e >> 32);   // CS0019
```

Expected: the printer should cast the enum operand to its underlying integer
before a shift, e.g. `(int)((long)clientCapabilities >> 32)`.

Fix belongs in the CSharpPrinter shift-operand coercion (analogous to the
And/Or/Xor enum coercion), not in `BinaryResult`.

Issue #2095, opened 2026-07-02
Project: the thin writer — make CSharpPrinter a total function of a decided, fully-typed IR


richlander:
# Project: the thin writer — make `CSharpPrinter` a total function of a decided, fully-typed IR

**Design rationale:** [`docs/design/value-typed-emission.md`](https://github.com/richlander/dotnet-inspect/blob/main/docs/design/value-typed-emission.md) (PR #2090). Read it first — this issue is the *execution plan* for the redesign it argues for.

> **Revised after adversarial plan review (GPT-5.5 + Gemini 3.1 Pro).** Six execution findings reordered Phase A, split Phase C, scoped the A4 invariant to non-stack sinks, added a Release-executed checker, and strengthened the fidelity gate. See "Plan-review revisions" at the end.

## Goal

The writer is too thick: it makes semantic decisions the typed IR never recorded (casts/`unchecked`, which locals exist and their types, `= default`) and rediscovers each one — wrongly — one render context at a time. Root cause behind the enum-cast defect class (six review rounds on #2080) and the `S_0`/`S_256` variable collapse (#2075). End state: a **thin writer** that makes zero semantic decisions and only spells a decided tree, enforced by a checker rather than by recompiling corpus output. Lands as a **sequence of PRs across three phases**.

## Principles and gates (every PR)

- **Consolidate, don't rewrite.** The conversion rules (`EnumIntegerCast`/`MayOverflow…`, and the merged #2080 fixes) become the *internals* of one renderer.
- **Expect and measure behavior movement.** Ship the corpus quality-diff card; net-positive validity is the target.
- **Changed-method fidelity, not just the sampled card.** The aggregate card is capped (`--compile-cap 25`, `--corpus-fidelity-cap 3`); a method can stay `Full` yet recompile to different IL and be missed. Every behavior-changing PR **must** run changed-method fidelity (`--fidelity-method-delta` / a focused uncapped fidelity run over exactly the methods whose output moved). **A `Full`-method fidelity regression is the stop signal.**
- **Release-executed invariant.** `IrNode.CheckInvariant()` is `[Conditional("DEBUG")]`, so it is compiled out of the Release corpus/card runs. The thin-writer invariant must be a **non-conditional validation pass / separate checker that runs in Release**, not added to the DEBUG-only method.
- **Invariant staged** warn → error per phase, once its burn-down is green.
[...]

PR #3080, opened 2026-07-23
Recover cross-assembly enum shift width from the count mask (#3066)

### LINKED ISSUE #3066: Cross-assembly enum shift still renders CS0019 (follow-up to #3011/#3060)
### Summary

A shift (`>>`/`<<`) on an enum whose type is defined in a **referenced** assembly still decompiles to `enumValue >> n`, which is `CS0019` (C# has no predefined shift operator for enum types). PR #3060 fixed this for **same-assembly** enums — where the importer resolves the enum backing and the printer reinterprets the left operand to the opcode-signed integer — but the cross-assembly case remains uncompilable.

### Repro

Enum in one assembly, shift in a consumer that references it (reference present at decompile time):

```csharp
// lib.dll
namespace Ext;
public enum ExternalFlags : long { Zero = 0, High = long.MinValue }

// app.dll (references lib.dll)
namespace Consumer;
public static class Shifts
{
    public static long Signed(ExternalFlags e, int n) => (long)e >> n;
    public static ulong Unsigned(ExternalFlags e, int n) => (ulong)e >> n;
    public static ExternalFlags Compound(ExternalFlags e, int n) { e = (ExternalFlags)((ulong)e >> n); return e; }
}
```

Decompiling `app.dll` (with `lib.dll` alongside) currently renders:

```csharp
public static long Signed(Ext.ExternalFlags e, int n) => (long)(e >> (n & 63));   // CS0019
public static ulong Unsigned(Ext.ExternalFlags e, int n) => (ulong)(e >> (n & 63)); // CS0019
// Compound: e >>= (n & 63);                                                        // CS0019
```

### Root cause

`MetadataSource.EnsureTypeMaps` maps only types in the current module (`Reader.TypeDefinitions`), so `ResolveEnumUnderlyingType` returns `null` for a `TypeReference`. `ShiftEnumLeftOperand` and the compound-shift decomposition in `CSharpPrinter` both bail out when the underlying type is unknown, leaving the CS0019 spelling. The failure is **visible** (invalid, not silently-wrong): the collapsed `shr`/`shr.un` inputs also render identically.

### Constraints

The product path is SRM-only and "free of inspected-assembly loading" (`AGENTS.md`), so resolving the referenced enum's backing by loading `lib.dll` is out of bounds. Two options worth evaluating:

1. **Infer width + signedness from IL evidence in the current module.** A variable shift count carries the compiler's width mask (`& 31` → 4-byte, `& 63` → 8-byte) and the opcode (`shr`/`shr.un`) carries signedness, so `Signed`/`Unsigned`/`Compound` above (masked variable counts) are recoverable without loading the referenced assembly. A **constant** shift count (e.g. `>> 32`) has no mask, so its width stays unknowable — that residual remains CS0019.
2. **Provide referenced enum backing via an explicit, capability-gated reference resolver** if/when such a mechanism exists, keeping default paths SRM-only.

### Scope note

Split out of PR #3060 per reviewer discussion: GPT flagged it (High); Gemini judged it an acceptable architectural limitation. Filed as follow-up so #3060 stays focused on the same-assembly fix (which covers the reported MySqlConnector witness).

richlander:
- Fixes #3066 (follow-up to #3011 / #3060)
- Changes cross-assembly enum shifts to render valid C# by recovering the
  shift width from the compiler-baked count mask
- Card revision: `ccd30f47`

## Change

**Conclusion:** **PASS** — a shift on a referenced-assembly enum previously
rendered `enumValue >> n` (CS0019); it now recovers the backing width from the
outermost count mask and reinterprets the operand, exactly as the same-assembly
path does, staying SRM-only / Roslyn-free with no inspected-assembly loading.
The fix covers the bare operand and its `ref` / array-element siblings.

### Original source

Real cross-assembly fixture: the enum type lives in a *referenced* assembly
(`ILInspector.Decompiler.Fixtures.CrossAssemblyEnums`), the shift in
`EnumCastSamples`. The author's `(long)e` cast is a no-op reinterpret in IL, so
the emitted body is a bare enum shift — the exact #3066 shape.

```csharp
// ExternalLong is defined in a referenced assembly (backing: long)
public static long ExternalLongRightShift(ExternalLong e, int n) => (long)e >> n;

public static ExternalLong ExternalLongCompoundLeftShift(ExternalLong e, int n)
{
    e = (ExternalLong)((long)e << n);
    return e;
}

// ref / array siblings (a typed ldind/ldelem masks the enum as its
// primitive backing width on the eval stack)
public static long ExternalLongRefLeftShift(ref ExternalLong e, int n) => (long)e << n;
public static long ExternalLongArrayLeftShift(ExternalLong[] a, int i, int n) => (long)a[i] << n;
```

### Before

Verbatim base render (`origin/main`, `d6af0165`):

```csharp
// long ExternalLongRightShift(ExternalLong e, int n)
return (long)(e >> (n & 63));

// ExternalLong ExternalLongCompoundLeftShift(ExternalLong e, int n)
e <<= (n & 63);
return e;

// long ExternalLongArrayLeftShift(ExternalLong[] a, int i, int n)
return a[i] << n;                  // bare, uncast, mask already stripped
```

- Valid: False
- Correct: False

`e >> (n & 63)`, `e <<= (n & 63)`, and `a[i] << n` are CS0019: C# has no
predefined shift operator for an enum type. `MetadataSource.EnsureTypeMaps` maps
only the current module's type defs, so the referenced enum's backing width is
unresolved and the `#3060` reinterpret never fired. For the `ref` / array
operands, a typed `ldind`/`ldelem` further masks the enum as its primitive
backing on the eval stack, so even the bare-operand recovery clause declined
while `ShiftCount` still stripped the mask — leaving a bare, uncast residual.

### After

Verbatim head render (`ccd30f47`):

```csharp
// long ExternalLongRightShift(ExternalLong e, int n)
return (long)((long)e >> n);

// ExternalLong ExternalLongCompoundLeftShift(ExternalLong e, int n)
e = (ExternalLong)((long)e << n);
return e;

// long ExternalLongArrayLeftShift(ExternalLong[] a, int i, int n)
return (long)a[i] << n;

// ExternalLong ExternalLongArrayCompoundLeftShift(ExternalLong[] a, int i, int n)
a[i] = (ExternalLong)((long)a[i] << n);
return a[i];
```

- Valid: True
- Correct: True

The width is recovered from the count mask (`& 63` => 8-byte, `& 31` => 4-byte);
signedness is the opcode's (`shr`/`shr.un`). The mask strips (keyed off the same
recovered width) so the shift does not double-mask. For `ref` / array operands
the enum is recovered from the by-ref pointee / array element type rather than
the primitive-masked eval-stack type, so the cast and the mask-strip always
agree. Both compile and preserve the original shift semantics; the shape matches
the same-assembly `#3060` output.

### Fully raised

## Evidence

| Check | Baseline (`d6af0165`) | Head (`ccd30f47`) |
| --- | --- | --- |
| Focused tests | `EnumCastPrinterTests` (42 pre-existing): Pass | `EnumCastPrinterTests`: 55 passed (13 new) |
| Decompiler fast — RoundTrip + Validity | 246, 34 failed | 246, 34 failed |
| Real witness | `EnumCastSamples::ExternalLongRightShift` (+ `…Array…`, `…Ref…`) CS0019 | fixed (valid + correct) |

The 34 (31 RoundTrip + 3 Validity) failures are **pre-existing environmental
reds**, confirmed identical on Baseline by stashing this PR's changes and
re-running:

- 31 `ReturnToSenderFixtureCatalogTests.*` — Roslyn compile-back returns
  `Invalid` (e.g. `Expected: SourceUnavailable, Actual: Invalid`) because this
  Windows machine's reference set is broken (the known `aspnetcorev2_inprocess.dll`
  CS0009 root); unrelated to enum shifts, and CI-green.
- 3 `LadderRung6GateTests.*RecompilesThroughFidelityHarness` (pointer
  arithmetic / function pointers / pinned pointers) — same fidelity-harness
  recompile failure (`Expected: Exact, Actual: RecompileFail`); no enums
  involved.

## Decompiler quality

**Conclusion:** **PASS (quick gate not applicable)** — the new behavior is gated
to a shift whose operand is an unresolved-backing enum (`IsEnumLikeShiftOperand`,
recognized through bare / `ldelem` / `ldind` shapes) with a recoverable count
mask. The self-corpus (CoreLib) has no *cross-assembly* enum shift operands —
CoreLib's own enums are same-assembly and already covered by `#3060` — so the PR
quick gate shows zero delta by construction. Coverage is carried instead by the
real cross-assembly fixtures and 13 focused tests, including opcode-wins
signedness mirrors, an inner-user-mask fixture proving the outer implicit width
mask (not a nested user mask) drives width recovery, a synthetic constant-count
decline that pins the soundness boundary, and `ref` / array-element / array-
compound siblings — each `AssertCompiles`-verified valid C#.

## Adversarial review

Two independent adversarial reviews of `b37d057e` (GPT `gpt-5.6-sol`, Gemini
`gemini-3.1-pro-preview`) both confirmed the central mask→width soundness claim
survives all attacks (SDK 10/11, Debug/Release, every backing, signedness,
checked context, constant count), and both independently reported the single
same finding: the `ref` / array-element cross-assembly gap. That gap was
**pre-existing** (identical on `origin/main`) but is the same feature, so it is
fixed here in `ccd30f47` via the `IsEnumLikeShiftOperand` predicate. (Gemini's
literal `ShiftLeftEnumType(operand)` suggestion would still decline, because that
helper's `ldelem`/`ldind` cases gate on `EnumUnderlyingType` which returns null
off-module; the fix consults the element / pointee type via `IsEnumLikeInteger`,
which admits the unresolved referenced enum.) Both reviewers re-reviewed the
fixed head `ccd30f47`.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class ILInspector.Decompiler.Tests.EnumCastPrinterTests
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -trait "Area=RoundTrip" -trait "Area=Validity" -trait- "Speed=Slow"
```


richlander:

Two independent adversarial reviews, per `AGENTS.md` (I'm Claude Opus, so GPT + Gemini Pro):

- **GPT** (`gpt-5.6-sol`), isolated worktree
- **Gemini Pro** (`gemini-3.1-pro-preview`), isolated worktree

### Round 1 — `b37d057e`

Both reviewers **independently confirmed the central soundness claim survives all attacks**: the compiler's implicit width mask (`& 31` / `& 63`) is reliably the *outermost* node feeding `shr` (Roslyn does not fold the ANDs), across SDK 10/11, Debug/Release, every backing, signedness, `checked` context, and constant count. Sub-`int` enums evaluate as 32-bit shifts matching C# `int` promotion; `checked` regions wrap the reinterpret in `unchecked(...)` to avoid a spurious `conv.ovf`.

Both then reported the **same single finding** (High):

- **Finding — `ref` / array-element cross-assembly enum shifts render bare and uncast** (e.g. `a[i] << n`, `e << n`), CS0019, with the count mask already stripped. A typed `ldelem`/`ldind` masks the enum as its primitive backing in `operand.ResultType`, so the cross-assembly reinterpret clause (`IsEnumLikeInteger(operand.ResultType)`) declined while `ShiftCount` still stripped the mask — a cast/strip disagreement.

**Verification:** I reproduced it on a clean exact-head worktree, then reverted the product files to `origin/main` and re-probed — the output was **identical**, so this was **pre-existing**, not a regression from this PR. Because it is the same feature (#3066) and tightly coupled, I fixed it here rather than deferring.

**Resolution — commit `ccd30f47`:** added `IsEnumLikeShiftOperand(IrExpression)`, which recognizes the enum through the bare / `LoadElement` (array element type) / `LoadIndirect` (`WideIndexPointee`) shapes via `IsEnumLikeInteger`, and drives **both** the reinterpret in `ShiftEnumLeftOperand` and the mask strip in `ShiftCount` off that single predicate so the cast and the strip always agree.

> Note on Gemini's literal suggestion (`ShiftLeftEnumType(operand)`): that helper's `ldelem`/`ldind` cases gate on `EnumUnderlyingType(...) is not null`, which returns null for an **off-module** enum, so it would still decline the cross-assembly case. `IsEnumLikeInteger` on the element/pointee type is what admits the unresolved referenced enum — so the fix intentionally diverges from the literal suggestion while honoring its intent.

Added 4 fixtures + 4 `AssertCompiles`-verified tests (ref plain, array plain, uint-signed array opcode mirror, array compound).

### Round 2 — fixed head `ccd30f47`

Both reviewers re-reviewed the exact fixed head in synced isolated worktrees:

- **GPT — CLEAN.** Ref/array fixtures render valid casts and compile; 55/55 tests pass. Regression probes (non-enum, same-assembly, multidimensional, jagged, pointer, generic-holder) showed no misfires or cast/strip disagreement. Central claim holds in Debug and Release.
- **Gemini Pro — CLEAN.** Extensively attacked `IsEnumLikeShiftOperand` with `ref ExternalLong`, `ExternalLong[]`, jagged `[][]`, multidimensional `[,]` (lowers to a `Call`), `ExternalLong*` (pointer), and `ExternalLong?` (nullable, `GetValueOrDefault()`) — all correctly recognized. Confirmed non-enum arrays (`int[]`, `ref int`) do **not** misfire (the `conv.i8`/`Convert` node keeps `ResultType` primitive → predicate returns false). Central mask→width claim re-confirmed.

Both fixed-head reviews are clean with no outstanding findings.

richlander:
Ready to merge

richlander:

Two reviewers on different models per the AGENTS.md roster: **GPT** (`gpt-5.6-sol`) and **Gemini Pro** (`gemini-3.1-pro-preview`). Each isolated in its own linked review worktree at the exact head, `/tmp` scratch only, no git mutations. Every blocking finding was reproduced on a clean exact-head worktree before acting.

### Round 1 — head `ecb5da84`

**Finding (BOTH reviewers, independently): `ShiftRenderedIntegerType` still required a resolved enum backing.** A cross-assembly enum shift feeding a mixed-sign bitwise sibling or an int->enum sink produced invalid C# (CS0019/CS0266) because the sibling/sink width query only handled resolved backings.

- **Verified & fixed** in `ecb5da84`: recover width via `IsEnumLikeShiftOperand` + `ShiftCountMaskWidthBytes` (the same count-mask recovery the left-operand path uses). Added 3 real cross-assembly fixtures + tests; confirmed 2 of 3 fail without the fix.

### Round 2 — head `cec7e0fd`

- **GPT: CLEAN** (built, ran enum + fast RoundTrip/Validity + full non-slow suite).
- **Gemini finding: int->enum shift SINK gated on `CanSpellIntegerToEnum` (requires `TypeShape.Enum`).** A genuinely unresolvable (`TypeShape.Unknown`) cross-assembly enum sink dropped the outer cast -> CS0266.
  - **Verified & fixed** in `cec7e0fd`: the sink now also accepts an Unknown-shaped target via `CanSpellUnknownEnumConstant` (mirroring the existing bare-literal Unknown-enum sink). Added synthetic seam test `UnknownShapeEnumShift_ReturnedToEnum_WrapsInOuterEnumCast` (synthetic because a harness-referenced enum always resolves to `TypeShape.Enum`); confirmed it fails without the fix.

### Round 3 — re-review of final head `cec7e0fd`

- **GPT: CLEAN.** Builds pass; enum 70/70; fast RoundTrip+Validity 252/252; full non-slow suite 3430/3430. Unknown-shape guard appropriately constrained; resolved-enum behavior unchanged.
- **Gemini: confirmed the sink + bitwise fixes are sound and all enum/RoundTrip/Validity tests pass**, and raised **two additional findings** that I am tracking as **out-of-scope follow-ups** rather than expanding this PR:
  1. `NeedsIntShiftCast` (`CSharpPrinter.Numerics.cs`) misses the `(int)` cast when the shift **count** is an Unknown-shape enum -> CS0029. -> **#3105**
  2. The constant switch (`CSharpPrinter.cs`) renders an Unknown-shape enum **constant** as a bare integer -> CS0266. -> **#3106**

  **Why out of scope:** both gate on `TypeShape.Enum` in lines that are **byte-identical on `origin/main`** and untouched by this PR (introduced in #835 and #568 respectively). They are also **different scenarios** from this PR's subject: #3080 recovers width when the shift **left operand** is a cross-assembly enum, whereas these concern the shift **count** and a bare **constant**. They are pre-existing latent gaps in Unknown-shape enum rendering, not regressions from this PR. The Round-2 sink fix, by contrast, *was* in scope because this PR's width recovery is what makes the shift render as a bare integer that the sink must re-wrap. Keeping this PR surgical and coherent; filed #3105/#3106 (umbrella #2095) so they're tracked.

### Evidence

- CI green on `cec7e0fd`: `build-net10`, `csharp-diff-smoke`, `test (ubuntu-26.04, linux-x64)` all pass.
- Local: 3430/3430 fast tests; 70/70 enum; each regression test confirmed to fail on the pre-fix head.

Both fixed-head reviews are clean with respect to this PR's changes; the only remaining reviewer findings are pre-existing/out-of-scope and tracked in #3105 and #3106.

richlander:
Ready to merge

PR #3082, opened 2026-07-23
Re-apply switch-expression raise soundness gates onto the new Scrutinee/TryFold pass (#3022)

richlander:
- Advances #3022 (soundness follow-up to merged #3034 / #3065)
- Re-applies the switch-expression raise soundness gates onto #3065's rewritten `PatternSwitchExpressionPass`; the canonical target still fully raises unchanged
- Head: `d72f2076` (merge of `origin/main` + gate re-application)

## Change

**Conclusion:** **PASS** — two-model adversarial review of the switch-expression raise found soundness holes; the raise merged (#3034, then rewritten by #3065) without them, so main carries them. This PR closes them with sound, conservative gates while preserving the target raise.

### Merge note

While this branch was under review, #3065 (`8671a091`) **rewrote** `PatternSwitchExpressionPass` with a new `Scrutinee`/`TryFold` architecture (adding direct-place and inline-positive arm surfaces) and explicitly deferred "guarded/subpattern heterogeneous arms … to a follow-up carrying a real oracle." This branch **is** that follow-up. It merges `origin/main` — taking the new architecture verbatim (commit `a5616e78`) — and re-applies the soundness gates onto it (commit `d72f2076`), threading the type-disjointness oracle the rewrite left unwired.

### Gates (on the new architecture)

1. **Overlapping refutable arm (oracle).** A `when`-guard or property-subpattern temp-form arm routes a value that matches its type but fails the refinement to the *default* in the cascade, but to a *later arm* in a switch expression. The pass threads the already-open `MetadataSource` disjointness oracle (`PassContext.TypesProvablyDisjoint` → `MetadataSource.AreProvablyDisjoint`, a base-chain proof that declines open generics) into `TryMatch`/`TryFold`. A refutable non-last arm raises only when *provably* disjoint from every later arm; absent a proof it declines. (#3065's `isNewSurface` gate already declines guarded/subpattern arms on the direct/inline surface.)
2. **Pattern-local scope leakage + unrendered local.** `PatternLocalsAreArmScoped` rejects a pattern local read by a sibling arm's guard/value or by the default, and shared slots. `ArmData` now carries an always-present `PatternLocal` so an **unrendered** intro/inline match local is still scope-checked (a `LocalIndex`-only check missed it).
3. **Temp read misplacement (Fix D).** The fold deletes the switch-value temp's defining store, so a *read* of that temp in an arm guard/value or the default would observe `default(T)`. #3065's read-only check rejects stores/address-of, not a misplaced load; this gate declines it.
4. **Governing-value alias (Fix E).** Decline when the temp-form governing value is a `LoadLocal` aliasing a rendered pattern local — the fold would emit an unspellable `V switch { T V => … }`.

### Original source

```csharp
// ILInspector.Decompiler.Pipeline.YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition
return expression switch
{
    Comparison comparison when ReadsLoopField(comparison.Left, loopFieldName) => Assign(comparison, out condition),
    LogicalNot { Operand: Comparison comparison } when ReadsLoopField(comparison.Left, loopFieldName)
        => Assign(Negate(comparison), out condition),
    _ => false,
};
```

### Before

```csharp
// bool TryNormalizeContinueCondition(IrExpression expression, string loopFieldName, ref IrExpression condition)
// main (#3065): raises, but the temp-form guarded/subpattern path raises on
// ASSUMED compiler exclusivity with no oracle, and does not gate misplaced
// temp reads / governing-value aliasing (this target is not such an input).
return expression switch
{
    Comparison comparison when ReadsLoopField(comparison.Left, loopFieldName) => Assign<IrExpression>(comparison, out condition),
    LogicalNot { Operand: Comparison comparison } when ReadsLoopField(comparison.Left, loopFieldName) => Assign<IrExpression>(Negate(comparison), out condition),
    _ => false,
};
```

- Valid: True
- Correct: True (for this target; the merged raise is unsound for the inputs the gates describe)

### After

```csharp
// bool TryNormalizeContinueCondition(IrExpression expression, string loopFieldName, ref IrExpression condition)
return expression switch
{
    Comparison comparison when ReadsLoopField(comparison.Left, loopFieldName) => Assign<IrExpression>(comparison, out condition),
    LogicalNot { Operand: Comparison comparison } when ReadsLoopField(comparison.Left, loopFieldName) => Assign<IrExpression>(Negate(comparison), out condition),
    _ => false,
};
```

- Valid: True
- Correct: True

The target's two refutable arms (`Comparison` guarded, `LogicalNot` subpattern) are proven disjoint (same-assembly sealed siblings chaining to `System.Object`), so the raise is preserved. The change only adds decline paths for inputs that cannot be proven sound.

### Fully raised

## Evidence (head `d72f2076`)

| Check | Result |
| --- | --- |
| `dotnet build dotnet-inspect.slnx -c Release` | clean (0 warnings) |
| `PatternSwitchExpressionPassTests` | 22 pass (4 new #3082 gate negatives + 18 from #3065) |
| Fast decompiler suite (`-trait- Speed=Slow`) | 3405 pass, 0 fail |
| Target self-raise (`TryNormalizeContinueCondition`) | raises (unchanged, oracle-wired) |
| Corpus PR quick gate | PASS — 0 pass bugs, pinned-subset 0pp (overall residue 251 → 250) |
| `git diff --check` | clean |

Each new gate negative was verified to **reproduce without its gate** (temp-revert): `Synthetic_TempGuardedNonLastArm_RaisesOnlyWhenProvablyDisjoint`, `Synthetic_UnrenderedPatternLocalReadByDefault_DoesNotRaise`, `Synthetic_TempScrutineeReadInArmValue_DoesNotRaise`, `Synthetic_GoverningValueAliasesRenderedPatternLocal_DoesNotRaise`.

### API compatibility

The oracle threading adds optional parameters to `PassContext`/`IrPasses`/`CSharpPrinter`/`MemberBodyProducer`, following those members' pre-existing optional-parameter convention. `ILInspector.Decompiler` is an internal library (no packaging, no PublicAPI analyzer) shipped only inside the `PackAsTool` `dotnet-inspect` package, where every assembly is recompiled together — so there is no external binary contract to break.

## Adversarial review

Two-model review (non-Opus: GPT + Gemini) at the merged head `d72f2076` in isolated read-only worktrees. Reconciliation to follow on this PR.


richlander:

Two-model adversarial review (GPT-5.6-sol + Gemini-3.1-Pro) at head `4cd56f5c`. Both returned **CHANGES REQUESTED**. All findings addressed in `817b16d2`; a fixed-head re-review by both models is in progress.

### Finding A — unrendered pattern local escapes scope validation (both reviewers)

Reported independently by **GPT** (default reads the local) and **Gemini** (a sibling arm reads the local). Same root cause: an arm's `LocalIndex` was set to `null` when the arm did not reference its own `isinst` local (a rendering decision so the printer omits an unused `T t` binding), and `PatternLocals` yielded only the rendered `LocalIndex`. The unrendered IL local was therefore never registered in `PatternLocalsAreArmScoped`'s `owner` map, so a sibling arm or the default could read it and the pass still raised — producing out-of-scope C# (CS0103) or a `default(T)`-vs-matched-value behavior change.

**Verified** — reproduced both variants as failing guard tests on the buggy code before fixing.

**Fix:** `ArmData` now carries an always-present `PatternLocal` (the outer isinst slot) separate from the printer-facing `LocalIndex`; `PatternLocals` yields `PatternLocal` so scope validation sees every introduced local. Guard tests `Synthetic_UnreadPatternLocalLeaksIntoSiblingArm_DoesNotRaise` and `Synthetic_UnreadPatternLocalLeaksIntoDefault_DoesNotRaise` (each with a raising positive control) pin both variants; each was confirmed to fail without the fix.

### Finding B — `AreProvablyDisjoint` falsely proves open generic base/derived disjoint (GPT)

`AreProvablyDisjoint(ConfigurableArrayPool\`1, ArrayPool\`1)` returned `true` even though the former derives from the latter. A derived type's generic base appears in the base chain as a closed `GenericInstance`, which never equals the open definition, so ancestry through a generic supertype was missed — contradicting the documented hard guarantee that open types never yield a disjoint proof. Gemini independently verified the surrounding oracle as sound (cross-assembly / generic-instance inputs already decline) but did not hit this open-**definition** direct-API path.

**Verified** — reproduced as a failing guard test on the buggy code before fixing.

**Fix:** `EnsureTypeMaps` builds a same-assembly generic-definition set (`_genericDefinitions`, populated when `typeDef.GetGenericParameters().Count > 0`); `AreProvablyDisjoint` returns `false` when either input is a generic definition. Guard test `AreProvablyDisjoint_OpenGenericDefinition_NeverProvenDisjoint` plus positive controls (`_NonGenericSiblings_ProvenDisjoint`, `_AncestorPair_NotDisjoint`) pin it, backed by self-contained fixture types compiled into the test assembly (no BCL-internal dependency). Confirmed to fail without the fix.

### Nothing dismissed

Both reviewers' remaining assessments were **Verified Safe** (Gemini explicitly confirmed `AreProvablyDisjoint`, `RefutableArmsDisjointFromLaterArms`, and `fallsThrough`/`ReachesDefaultTail` as sound). No finding was dismissed; both blocking findings are fixed.

### Validation at `817b16d2`

- `PatternSwitchExpressionPassTests` — 13/13 (8 prior + 5 new guards)
- Fast decompiler suite — 3378/0
- Canonical target `YieldBreakLoopIteratorReconstruction.TryNormalizeContinueCondition` still self-raises to the two-arm `switch` expression (harness dump)
- `git diff --check` — clean

Fixed-head re-review by both models is running; I will post each verdict and mark readiness only once both are clean with green CI.

richlander:
## Adversarial review round 2 — reconciliation

Fixed-head re-review of `817b16d2` (the round-1 fixes).

### Gemini-3.1-Pro — **LGTM**

Verified both round-1 fixes as sound and complete. Notably confirmed Fix B's generic-definition set is complete against nested generics (`TypeDefinition.GetGenericParameters().Count` includes inherited enclosing type parameters), closed generic instances / delegate variance (declined by the `Kind == Definition` guard), and cross-assembly generic bounds (declined because `_baseTypes` is same-assembly, so `TryBaseChainToObject` hits null before `System.Object`). No issues.

### GPT-5.6-sol — **CHANGES REQUESTED** (one new Medium)

**Finding C — a pattern local may overwrite the switch-value temporary.** Scope validation did not reject an arm whose outer `isinst` local aliases `svLocal`. The first `isinst` then overwrites the receiver that later arm tests re-read, while the raised switch evaluates every arm against the original receiver — so e.g. a `Branch` argument returns the default in the cascade but matches the `Branch` arm in the switch.

**Verified** — reproduced as a failing guard test on the pre-fix code before fixing.

**Fix (`bf09d5bf`):** decline when any arm's introduced local (outer `PatternLocal` or subpattern local) equals `svLocal`:

```csharp
if (arms.Any(a => PatternLocals(a).Contains(svLocal)))
    continue;
```

Guard test `Synthetic_ArmLocalAliasesSwitchValueTemp_DoesNotRaise` pins it and was confirmed to raise (wrongly) without the fix.

### Validation at `bf09d5bf`

- `PatternSwitchExpressionPassTests` — 14/14 (adds the alias guard)
- Fast decompiler suite — 3379/0
- Canonical target still self-raises to the two-arm `switch` expression (harness dump)
- `git diff --check` — clean

Nothing dismissed. A fresh final-head re-review by both models is running at `bf09d5bf`; I will mark readiness only once both are clean with green CI.

richlander:
## Round-3 review reconciliation (head `3cafa502`)

Both models independently converged on the **same** finding at `bf09d5bf`, so I treat it as high-confidence and have fixed it.

### Finding (GPT + Gemini, High/Medium): switch-value temp leaks into an arm body/default
The ownership gate (`LocalReferencesOnlyWithin(function, svLocal, consumed)`) only verified the switch-value temp is referenced *within* the consumed cascade — not *where*. The rewrite removes the temp's lowered binding (`svStore`) and re-spells `<place>` as the switch's governing expression, so any read of the temp in an arm **guard**, an arm **value**, or the **default** would observe `default(T)` in the raised switch instead of the original receiver — a silent value change (Gemini also notes CS0165 use-of-unassigned on emit).

- **GPT** repro: `V5=node; V0=isinst Leaf(V5); if(!V0) return false; return Use(V5)` raises to `node switch { Leaf => Use(V5), _ => false }` with `V5` default-initialized.
- **Gemini** repro: matched body returns `LoadLocal svLocal` directly; pass raises with zero remaining stores to that local.

**Fix** (`3cafa502`): enforce the pass's documented invariant — the switch-value temp may be read only by the arm type tests. Decline when it appears in any arm `Guard`/`Value` or the default:
```csharp
if (arms.Any(a => ReferencesLocalIn(a.Guard, svLocal)
        || ReferenceOwnership.SubtreeReferencesLocal(a.Value, svLocal))
    || ReferenceOwnership.SubtreeReferencesLocal(defaultValue, svLocal))
    continue;
```
This is exactly Gemini's suggested fix. Guard test `Synthetic_SwitchValueTempReadInArmValue_DoesNotRaise` reproduces the leak (verified it raises with the gate removed) and is identical to the raising positive except the arm value reads the temp.

### Validation at `3cafa502`
- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config` → 0 errors
- `PatternSwitchExpressionPassTests` → **15/15**
- Fast decompiler suite (`-trait- "Speed=Slow"`) → **3380/0**
- Canonical target `YieldBreakLoopIteratorReconstruction::TryNormalizeContinueCondition` still raises to `switch { ... }` (its temp is read only by the arm `isinst` tests, so the new gate does not over-decline)
- `git diff --check` clean

A fresh two-model round is now running against the exact fixed head `3cafa502`. I will not mark this Ready to merge until both return clean at that head with green CI.

richlander:
## Round-4 review reconciliation (head `5bc5b31b`)

### GPT — CHANGES REQUESTED (verified, fixed)

**Finding (Medium): governing switch value can alias a rendered pattern local.**
Scope validation never inspected the governing value (`svValue`). When it is a `LoadLocal` of a slot an arm also renders as its pattern variable, the rewrite re-spells the governing expression as that name and consumes the arm's `isinst` store, emitting:
```csharp
return V_0 switch { Leaf V_0 => Leaf.Use(V_0), _ => false };
```
The governing `V_0` is then out of scope (CS0103 — its only binding was the consumed store) or collides with the arm binding (CS0136). Un-spellable C#.

**Reproduced** on the clean head `3cafa502`: the added guard test `Synthetic_GoverningValueAliasesRenderedPatternLocal_DoesNotRaise` (governing `LoadLocal 0` aliasing the arm's rendered pattern slot) raises without the gate — I confirmed it fails (raises) with the gate removed.

**Fix** (`5bc5b31b`): decline when the governing `LoadLocal` aliases any rendered pattern local (an arm's `LocalIndex` or `Subpattern.LocalIndex`):
```csharp
if (svValue is LoadLocal governing
    && arms.Any(a => a.LocalIndex == governing.Index
        || a.Subpattern?.LocalIndex == governing.Index))
    continue;
```
The canonical target is unaffected — its governing local is distinct from every pattern local.

### Validation at `5bc5b31b`
- `dotnet build dotnet-inspect.slnx -c Release --configfile nuget.config` → 0 errors
- `PatternSwitchExpressionPassTests` → **16/16**
- Fast decompiler suite (`-trait- "Speed=Slow"`) → **3381/0**
- Canonical target still raises to `switch { ... }`
- `git diff --check` clean

A fresh two-model round will run against `5bc5b31b`. Not Ready to merge until both models return clean at the exact head with green CI.

richlander:
## Merge reconciliation (head advanced to `d72f2076`)

While this branch was under review, **#3065 (`8671a091`) rewrote `PatternSwitchExpressionPass`** with a new `Scrutinee`/`TryFold` architecture and explicitly deferred the disjointness oracle to "a follow-up." Rather than force-fit the old cherry-picked gates, I merged `origin/main` and re-implemented the gates on the new architecture:

- **`a5616e78`** — merge of `origin/main`; `PatternSwitchExpressionPass.cs` + its tests resolved to #3065's new architecture verbatim. The oracle infrastructure (`PassContext.TypesProvablyDisjoint`, `MetadataSource.AreProvablyDisjoint` + the open-generic guard) and Research overlay wiring auto-merged and were verified intact.
- **`d72f2076`** — re-applies the four gates onto the new `Scrutinee`/`TryFold` pass (see the refreshed PR description): oracle-gated refutable arms, always-present `PatternLocal` scope check, misplaced temp-read decline (Fix D), governing-value alias decline (Fix E).

The three earlier-round findings map onto #3065's new checks: `isNewSurface` already declines guarded/subpattern on the direct/inline surface; `TempScrutineeReadOnly` covers the earlier Fix C; `ReachesDefaultTail` covers finding-3. The re-applied gates cover the temp-form guarded/subpattern surface #3065 raised on assumed exclusivity.

**API compatibility (prior GPT round-5 finding):** the optional-parameter additions are not a binary break — `ILInspector.Decompiler` is an internal library (no packaging, no PublicAPI analyzer) shipped only inside the `PackAsTool` `dotnet-inspect` package where all assemblies recompile together.

Fresh two-model adversarial review is running at `d72f2076`; I will reconcile both here before marking ready.

richlander:
## Adversarial review reconciliation — head `d72f2076`

Two-model fixed-head review at `d72f2076` (both non-Opus), each in an isolated read-only worktree (`/tmp` scratch only; no index/history mutation). **Both returned LGTM with zero findings.**

**GPT-5.6-sol** — verified: Release build succeeded; all 22 `PatternSwitchExpressionPassTests` pass; direct/inline surfaces stay protected by the stability + `isNewSurface` gates; disjointness **fails closed** when no oracle is present; last/single refutable arms remain safe; outer-type disjointness is sufficient for property subpatterns; negative local indices cannot represent genuine metadata slots; all four new guard tests are **non-vacuous** by fold-path analysis (neutralizing each targeted gate lets its fixture fold).

**Gemini-3.1-Pro** — built with `--configfile nuget.config`, ran the `*Synthetic_*` suite green, and probed each attack point:
- **Gate 2 / inline arms:** `IsInlineArm` extracts `LocalIndex` (`>= 0`) into `ArmData.PatternLocal`, so an unrendered inline match local is still scope-checked; it cannot leak to a sibling guard/value or the default without tripping the `owner`-map collision.
- **Fix D / Fix E scoping:** correctly limited to temp-form (`scrutinee.TempLocal`). Direct-form has no defining store to delete and is covered by `DirectScrutineeStable`; mixed inline-on-temp guarded arms decline via `isNewSurface`, and unguarded inline bodies are audited by Fix D's `SubtreeReferencesLocal`, so no arm observes `default(T)`.
- **`PatternLocal >= 0` sentinel:** IL locals are 0-based, so `-1` is a collision-free "no bound local" sentinel.
- **Oracle direction:** "disjoint from every LATER arm" is the correct asymmetric check — a refutable arm that fails its guard routes to `default` in IL but falls through to later arms in the switch expression, so proving disjointness from all `j > i` preserves semantics.
- **Outer vs. inner subpattern type:** checking the outer `PatternType` is sound (disjoint outer types ⇒ disjoint refined sets); overlapping outer types conservatively decline. An intentional missed optimization, not a soundness hole.
- **Vacuous-fixture audit:** all four synthetics use place-comparable defaults (`LoadLocal(0)` / `Constant(false)`) that `PlaceIdentity.SameOperand` matches, so each forms a real match chain and relies strictly on the new gates to decline.

### Status

- Build `dotnet build dotnet-inspect.slnx -c Release`: clean
- CI at `d72f2076`: `test (ubuntu-26.04, linux-x64)`, `build-net10`, `csharp-diff-smoke`, `changes` all pass (`il-diff-smoke`/`markdownlint`/`pack` path-skipped)
- Corpus PR quick gate: PASS — 0 pass bugs, pinned-subset 0pp
- Both fixed-head adversarial reviews clean

Ready to merge

Issue #3071, opened 2026-07-23
Decompiler: fully raise CSharpDeclarationWriter::EscapeReservedKeywordIdentifiers (copy-temp fold, conditional-arg inline, short-circuit && / De-Morgan, decl sinking)


richlander:
## Follow-up frontier from #3059

#3059 renders target-typed `new()` at LHS sites, using
`ILInspector.CSharp.CSharpDeclarationWriter::EscapeReservedKeywordIdentifiers` as
its witness. The After decompilation is **correct** and the target-typed-`new()`
raise is fully applied, but — as noted on that PR — the body is **not otherwise
fully raised**. This records the remaining mechanical raising residue so the
endpoint is tracked without overclaiming "fully raised" (cf. #2877, #3028).

**Authored source:**

```csharp
static string EscapeReservedKeywordIdentifiers(string text)
{
    var sb = new StringBuilder(text.Length);
    for (int i = 0; i < text.Length;)
    {
        if (IsIdentifierStart(text[i]))
        {
            int start = i++;
            while (i < text.Length && IsIdentifierPart(text[i]))
                i++;
            string token = text[start..i];
            bool alreadyEscaped = start > 0 && text[start - 1] == '@';
            sb.Append(!alreadyEscaped && CSharpKeywords.RequiresDeclarationEscape(token) ? EscapeIdentifier(token) : token);
            continue;
        }
        sb.Append(text[i++]);
    }
    return sb.ToString();
}
```

**Current decompiled output** (shipped/raised view, harness `--dump`; `// fidelity: Full`):

```csharp
int start;
string token;
int V_5;
string S_2;

StringBuilder sb = new(text.Length);
int i = 0;
while (i < text.Length)
{
    if (IsIdentifierStart(text[i]))
    {
        start = i++;
        while (i < text.Length && IsIdentifierPart(text[i]))
        {
            i++;
        }
        V_5 = start;
        token = text.Substring(V_5, i - V_5);
        S_2 = !((start <= 0 ? false : (text[start - 1] == '@')) || !CSharpKeywords.RequiresDeclarationEscape(token)) ? EscapeIdentifier(token) : token;
        sb.Append(S_2);
    }
    else
    {
        sb.Append(text[i++]);
    }
}
return sb.ToString();
```

[...]

PR #3085, opened 2026-07-23
Raise range indexer in any statement position (#3070)

### LINKED ISSUE #3070: Decompiler taste: raise Substring/Slice(start, end - start) to the range indexer s[start..end] (prefer_range_operator)
## Follow-up taste raise from #3059

While documenting the target-typed-`new()` taste raise (#3058/#3059), its witness
`ILInspector.CSharp.CSharpDeclarationWriter::EscapeReservedKeywordIdentifiers`
surfaced another class-3 (no-IL-anchor) taste divergence from authored source.

The authored code writes a range indexer:

```csharp
string token = text[start..i];
```

which the C# compiler lowers to `text.Substring(start, i - start)`. The decompiler
renders the lowered call verbatim:

```csharp
token = text.Substring(start, i - start);   // after copy-temp fold; see companion raising issue
```

Both spellings compile to the identical `callvirt string.Substring(int, int)` — a
range indexer `s[a..b]` **is** `s.Substring(a, b - a)` for `string` — so this is a
no-IL-anchor spelling choice. The style oracle decides it: `dotnet/runtime`'s
`.editorconfig` sets `csharp_style_prefer_range_operator = true`, so when the shape
is expressible as a range we should render `text[start..i]`.

## Proposed transform (conservative, per taste doc class 3)

Raise `s.Substring(start, length)` to a range indexer when the receiver is
`string` (and, for `Span`/`ReadOnlySpan`/arrays, `.Slice(start, length)`
likewise), rendering:

- `s[start..(start + length)]`, and specifically `s[start..end]` when `length`
  is spelled `end - start` (the idiomatic, readable case).
- `s[start..]` when `length` is `s.Length - start`.
- `s[..end]` when `start` is `0`.

Fidelity constraints (keep it sound, like #3059):

- Only for the framework `String.Substring(int, int)` / `Span{,ReadOnly}<T>.Slice(int, int)`
  overloads — not user-defined `Substring`/`Slice` methods.
- The single-argument `Substring(int)` / `Slice(int)` map to `s[start..]`.
- Preserve exact evaluation order and side effects of the `start`/`length`
  operands; decline if raising would reorder or duplicate an effectful operand.

This is a sibling of #3058: an IL-identical, oracle-preferred spelling applied as
a conservative over-approximation, validated by recompile/corpus fidelity.
Document alongside the target-typed-`new()` entry in `docs/decompiler-taste.md`.

richlander:
- Fixes #3070
- Changes: the two-bound from-start range-indexer raise (`V = start;
  …Substring/Slice(V, end - V)…` → `s[start..end]`) now fires in **any**
  straight-line statement position (local assignment, member receiver, call
  argument), not only in return position.
- Card revision: `e410dece`

## Change

**Conclusion:** **PASS** — the raise recovers the authored `s[start..end]` from
its spilled lowering in more positions, is guarded by a whole-function
single-use proof plus a leading-effect guard, and stays opcode-exact because the
raised source re-lowers to the same spill.

### Original source

Authored source of the flagship witness
`ILInspector.CSharp.CSharpDeclarationWriter::EscapeReservedKeywordIdentifiers`
(the `token` assignment is the raise site):

```csharp
static string EscapeReservedKeywordIdentifiers(string text)
{
    var sb = new StringBuilder(text.Length);
    for (int i = 0; i < text.Length;)
    {
        if (IsIdentifierStart(text[i]))
        {
            int start = i++;
            while (i < text.Length && IsIdentifierPart(text[i]))
                i++;
            string token = text[start..i];
            bool alreadyEscaped = start > 0 && text[start - 1] == '@';
            sb.Append(!alreadyEscaped && CSharpKeywords.RequiresDeclarationEscape(token) ? EscapeIdentifier(token) : token);
            continue;
        }
        sb.Append(text[i++]);
    }
    return sb.ToString();
}
```

### Before

`-S "Decompiled Source"` at base `d6af0165` (return-position-only raise leaves
the assignment-position two-bound `Substring` un-raised, exposing the hidden
start-spill temp `V_5`):

```csharp
static string EscapeReservedKeywordIdentifiers(string text)
{
    int start;
    string token;
    int V_5;
    string S_2;

    StringBuilder sb = new(text.Length);
    int i = 0;
    while (i < text.Length)
    {
        if (IsIdentifierStart(text[i]))
        {
            start = i++;
            while (i < text.Length && IsIdentifierPart(text[i]))
            {
                i++;
            }
            V_5 = start;
            token = text.Substring(V_5, i - V_5);
            S_2 = !((start <= 0 ? false : (text[start - 1] == '@')) || !CSharpKeywords.RequiresDeclarationEscape(token)) ? EscapeIdentifier(token) : token;
            sb.Append(S_2);
        }
        else
        {
            sb.Append(text[i++]);
        }
    }
    return sb.ToString();
}
```

- Valid: True
- Correct: True

Compiles and preserves behavior, but exposes the compiler-introduced start-spill
local `V_5` and renders the slice as `Substring`, which is exactly the residue
`csharp_style_prefer_range_operator` targets.

### After

`-S "Decompiled Source"` at this PR's head `8e623bd5` (the two-bound slice is
raised to the range indexer; the `V_5` spill temp is gone):

```csharp
static string EscapeReservedKeywordIdentifiers(string text)
{
    int start;
    string token;
    string S_2;

    StringBuilder sb = new(text.Length);
    int i = 0;
    while (i < text.Length)
    {
        if (IsIdentifierStart(text[i]))
        {
            start = i++;
            while (i < text.Length && IsIdentifierPart(text[i]))
            {
                i++;
            }
            token = text[start..i];
            S_2 = !((start <= 0 ? false : (text[start - 1] == '@')) || !CSharpKeywords.RequiresDeclarationEscape(token)) ? EscapeIdentifier(token) : token;
            sb.Append(S_2);
        }
        else
        {
            sb.Append(text[i++]);
        }
    }
    return sb.ToString();
}
```

- Valid: True
- Correct: True

The raise site now matches the authored `token = text[start..i]` exactly. The
remaining deltas from the original (`while` vs `for`, the De Morgan'd boolean and
inlined `alreadyEscaped`) are unrelated non-range residue tracked by #3071.

### Fully raised

The range slice this PR targets is fully raised (After matches the authored
`text[start..i]`). The rest of the method is not yet fully raised:

```csharp
static string EscapeReservedKeywordIdentifiers(string text)
{
    var sb = new StringBuilder(text.Length);
    for (int i = 0; i < text.Length;)
    {
        if (IsIdentifierStart(text[i]))
        {
            int start = i++;
            while (i < text.Length && IsIdentifierPart(text[i]))
                i++;
            string token = text[start..i];
            bool alreadyEscaped = start > 0 && text[start - 1] == '@';
            sb.Append(!alreadyEscaped && CSharpKeywords.RequiresDeclarationEscape(token) ? EscapeIdentifier(token) : token);
            continue;
        }
        sb.Append(text[i++]);
    }
    return sb.ToString();
}
```

- Required tracking issue: #3071 — the remaining witness slices (`for`-loop
  recovery, boolean De Morgan / `alreadyEscaped` inlining) outside this PR's
  range-indexer scope.

## Evidence

| Check | Baseline (`d6af0165`) | Head (`8e623bd5`) |
| --- | --- | --- |
| Focused tests | `RangeFromGetSubArrayPassTests`: 29 fast cases passed | `RangeFromGetSubArrayPassTests`: 36 fast cases passed (+4 positive, +3 negative incl. effectful-receiver; +1 slow compile-back gate) |
| Decompiler fast suite | environmental set only (recompile CS0009 + CRLF), no Range failures | 3377, 55 failed — all environmental (31 ReturnToSender + 18 Fidelity + 3 Ladder + 4 TypeSource + 1 Corpus recompile CS0009, 1 NullCoalescing recompile, 1 SmallStringSwitch CRLF); **0 Range/Slice-related** |
| Reduced fixture validity | not applicable | not applicable |
| Reduced fixture fidelity | not checkable locally (CS0009) | 4 fixtures + witness raise to `fidelity Full` via harness `--dump`; slow `GeneralPositionTwoBoundFixtures_RecompileExactly` gate validates opcode-exactness in CI |
| Real witness | `CSharpDeclarationWriter::EscapeReservedKeywordIdentifiers` broken (`token = text.Substring(V_5, i - V_5)`) | fixed (`token = text[start..i]`) |

The 56 fast-suite failures are pre-existing and environmental on this Windows
host, identical on Baseline and Head: Roslyn recompile tests fail with `CS0009`
on the native `aspnetcorev2_inprocess.dll` pulled into the shared-framework
reference set, plus one CRLF/LF mismatch and one partial-build fixture. CI
(Linux) is green for these. My new slow compile-back gate is `Speed=Slow` and is
excluded from the fast suite, so the fast-suite failure set is unchanged.

## Decompiler quality

**Conclusion:** **PASS** — the PR quick corpus sensor matches base tolerances
with zero pass bugs; the raise is neutral on the hash-stable sample (the witness
assembly is not sampled) and shows no regression.

### PR quick gate

Base-vs-head A/B on the **same** 15 target assemblies (SPC + pinned 1P/3P
packages + dotnet-inspect managed assemblies), hash-stable 100 methods per
assembly, `--compile-cap 0 --corpus-fidelity-cap 0` (no recompile, so `CS0009`
does not apply):

| Detected lowering residue (population differs) | 252 (16.80%) → 252 (16.80%) | n/a |
| Conditional-branch residue (population differs) | 46 (3.07%) → 46 (3.07%) | n/a |
| Forward-merge stops (population differs) | 32 (2.13%) → 32 (2.13%) | n/a |

> **Conclusion:** **PASS** — corpus sensor matched baseline tolerances; zero pass
> bugs; no residue regression.

The raise's improvement is targeted (the two-bound assignment-position idiom),
so it is proved by the focused fixtures and the real witness dump rather than the
100-method sample, which does not include `ILInspector.CSharp`. The A/B above was
measured at `8e623bd5`; the follow-up fix `e410dece` only *adds* a decline
(effectful receiver), so it is strictly more conservative and cannot raise the
residue or pass-bug counts.

## Validation

```bash
dotnet build src/dotnet-inspect -c Release -p:PublishAot=false
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.RangeFromGetSubArrayPassTests
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -trait "Area=RoundTrip"
```


richlander:
## Adversarial review

Two-model adversarial review per `AGENTS.md` (reviewers from the roster other
than the author's own model). Each reviewer ran in its own isolated linked
worktree with scratch confined to `/tmp`-style dirs and no index/history
mutation.

| Reviewer | Model | First pass (`8e623bd5`) | Fixed head (`e410dece`) |
| --- | --- | --- | --- |
| Reviewer A | `gpt-5.6-sol` (GPT) | 1 finding | **CLEAN** |
| Reviewer B | `gemini-3.1-pro-preview` (Gemini Pro) | 1 finding | **CLEAN** |

### Finding (both reviewers, independently) — effectful slice receiver reordered

High severity, high confidence. `TryRaiseFromStartRange` guarded the leading
statement prefix with `NothingEffectfulBefore`, but that helper only inspects
earlier *siblings*; the call's **receiver** (`Arguments[0]`) is a *child* of the
call and was never checked. So a two-bound shape with a directly effectful
receiver — reachable only from hand-written / no-PDB IL such as
`V = start; token = Effect().Substring(V, end - V)` (where a source `start`
local looks unnamed without a PDB) — was raised to `Effect()[start..end]`.
Because C# evaluates the range-indexer receiver *before* the start, this
reordered two observable effects.

Reproduced on a clean head worktree: a compiled no-PDB fixture
`int start = Start(); Receiver().Substring(start, end - start)` raised to
`Receiver()[Start()..end]`; executing original vs. raised produced side-effect
logs `SR` vs. `RS`.

**Resolution — commit `e410dece`.** Added `!IsSideEffectFree(receiver)` to the
match guard so the pass declines a directly effectful receiver, keeping it a
plain `Substring`/`Slice`. This is a sound conservative decline: csc **always**
spills an effectful receiver to a slot *before* the start (verified —
`GetString()[start..end]` lowers to `StoreStackSlot S = GetString(); …
S.Substring(spill, …)`), so a genuine range lowering always presents a
side-effect-free receiver load and still raises. Added the synthetic negative
`GeneralPosition_EffectfulReceiver_IsNotRaised` plus doc/coverage updates.

Verified end-to-end on the fixed head: the effectful-receiver repro now stays
`Receiver().Substring(V_0, end - V_0)`; the genuine `GetString()[start..end]`
lowering still raises; the real witness still raises to `token = text[start..i]`.

### Points explicitly cleared by both reviewers

Single-use accounting (`IsSingleUseStartSpill`: exactly 1 store + 2 loads, both
consumed by the matched call, so no dangling reference and `continue` ≡
`return false`); one-sided decline (`s[i..]`, `s[..j]` stay calls);
conditional / short-circuit safety (`IsSideEffectFree` allow-list pessimistically
declines `?:`, `&&`, `||`); type/name guards (`Int32` + compiler-generated);
and the span slot-layout normalization (a legitimate, documented harness `Exact`
contract — masks no semantic difference).

Both fixed-head reviews are **CLEAN** with no outstanding findings.

richlander:
## Opcode-exact validation for the `Speed=Slow` compile-back gate

`GeneralPositionTwoBoundFixtures_RecompileExactly` carries `[Trait("Speed", "Slow")]`, so it is excluded from the merge-blocking `ci.yml` lane (which runs `-trait- "Speed=Slow"`) and only runs in `deep-inspect.yml`'s Test lane. That lane's earlier `Run CLI tests (including slow integration)` step is currently red on 10 unrelated `CommandExecutionTests` (PInvoke markup and performance/resource triage), which short-circuits the sequential job before the decompiler step, so the slow gate is **skipped** in CI (run `30031692324`, re-run reproduced the same skip).

Those 10 CLI tests are **not** caused by this change:

- This diff touches 5 files, all in the decompiler subsystem (`RangeFromGetSubArrayPass.cs`, its test, `RangeRecoveryFacts.cs`, `LoweringCoverage.cs`, `docs/decompiler-taste.md`). Nothing under `src/dotnet-inspect` or `CommandExecutionTests.cs`.
- The failing PInvoke/triage tests have no code path to the range pass, and they **pass locally on this exact head (`e410dece`)**. The failure is a deep-inspect runner/environment condition (these slow-integration tests restore and inspect real packages), not this PR.

### Local opcode-exact evidence (stands in for the skipped gate)

I recompiled the raised output for every generalized position against the original lowered IL (minimal reference set, `/optimize+ /debug-`), comparing opcode streams with local-slot operands normalized (the harness `Exact` contract tolerates slot-layout changes):

| Position | Raised C# | opcode-exact (slot-normalized) |
| --- | --- | --- |
| assignment / return-concat | `string.Concat(s[start..end], "z")` | ✅ 11/11 |
| member-of | `s[start..end].Length` | ✅ 10/10 |
| call-argument | `Consume(s[start..end])` | ✅ 10/10 |
| span member-of | `ReadOnlySpan<int> V_1 = s[start..end]; … V_1.Length` | ✅ 12/12 |

All four also report `fidelity: Full` in `--dump`.

### No over-decline from the receiver guard

A genuine effectful-receiver case `GetString()[start..end].Length` still raises: csc spills the effectful receiver to a slot **before** the start, so the slice site sees a side-effect-free load and the receiver folds back with evaluation order preserved (opcode-exact). The new `!IsSideEffectFree(receiver)` decline only fires on the hand-authored / no-PDB inline-effectful-receiver shape `Effect().Substring(V, end - V)` (no preceding spill), which csc never emits — pinned by the `GeneralPosition_EffectfulReceiver_IsNotRaised` negative test.

The merge-blocking `ci.yml` lane is fully green on `e410dece`.

richlander:
Ready to merge

- Merge-blocking `ci.yml` is green on `e410dece` (test, build-net10, csharp-diff-smoke, changes, markdownlint; il-diff-smoke/pack path-skipped).
- Two independent adversarial reviews (GPT + Gemini) are clean on the fixed head `e410dece`; the single High-confidence finding (unchecked effectful slice receiver) is fixed and re-reviewed, reconciled above.
- Opcode-exact evidence for all four generalized positions plus no-over-decline is documented in the comment above. The `Speed=Slow` compile-back gate is excluded from `ci.yml` by design and is currently blocked in `deep-inspect.yml` by a pre-existing, unrelated CLI-integration failure (passes locally on this head); it will be exercised by `release.yml` / the scheduled deep-inspect lane once that CLI lane is healthy.

Follow-up (non-blocking): the deep-inspect Test lane's `CommandExecutionTests` (PInvoke/triage) reds are an environment condition unrelated to this decompiler-only change.

PR #3089, opened 2026-07-23
Document shipping taste choices and correct compound-assignment anchoring

richlander:
Documents taste choices the decompiler **already makes**, and corrects one misclassification. Doc-only; no product code changes.

## Changes to `docs/decompiler-taste.md`

1. **Optional-argument elision subsection.** `OptionalArgumentElisionPass` ships this today and self-describes as a taste transform bounded by `MethodRef.SafeTrailingElidableCount` / `OptionalArgumentFacts`. It is the other binding-hazard class-3 transform alongside target-typed `new()`, so the doc now covers it and the target-typed-`new` close is broadened from "apparent-type" to the general binding-hazard discipline.
2. **"Erased before lowering" class-3 rule.** Generalizes the one-off `Query` decline: constructs the compiler resolves to a bare constant before lowering leave no IL to recover. Examples: `nameof`, constant folding, primitive/reference `default`, string spelling.
3. **Compound/increment reclassification.** The doc listed `_size++` under class 1 (IL-exact). Verified it is a **class-3 no-anchor** pick for locals, static fields, `this`-rooted instance fields, and ordinary properties (all byte-identical IL), and IL-anchored **only** where the place is computed once (array/pointer `ldelema`+`dup`, or a side-effecting receiver).
4. **Oracle-as-default reframe.** The runtime `.editorconfig` is a *default*, not a source-style match (real assemblies are mixed; editorconfig is directional); it settles class-3 cosmetics only and never overrides a class-1/class-2 IL distinction.

## Evidence

IL-shape claims verified with a compiled Release canary (own-method IL via reflection):

- `nameof(x)` -> `ldstr` (identical to the string literal); `default(int)` -> `ldc.i4.0`; `default(string)` -> `ldnull`; `default(struct)` -> `initobj` (anchored).
- `i++` / `i += 1` / `i = i + 1` byte-identical for a local; same for a static field and a `this`-rooted instance field (`ldarg.0 ldarg.0 ldfld ... stfld`, no `dup`).
- Array element: `a[i]++` / `a[i] += 1` use `ldelema`+`dup`+`stind`, distinct from `a[i] = a[i] + 1` (`ldelem`/`stelem`).

## Risk

Documentation-only, describing shipped behavior. No adversarial review required. `markdownlint` runs in CI (no Node runtime on the authoring machine); `.markdownlint.yaml` disables MD013/MD032 so the long bullets conform.

### DIFF
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -26,6 +26,8 @@ Two reasons:
 1. **It is an established, versioned, externally documented choice.** Picking canonical representatives stops being a per-change taste debate, and the target moves with the runtime repo's own style evolution rather than ossifying into ours.
 2. **It makes testing coherent.** The fidelity fixture corpus is runtime-shaped code written under that style — so fixer-style output and recompile fidelity are the same goal, not competing ones.
 
+Two clarifications on scope. First, the oracle is a **default**, not a reconstruction of the input's own style: real assemblies — runtime included — are stylistically mixed, and an `.editorconfig` is *directional* (it names where a codebase is converging over time, not what its current text says), so "match the source style" is not a well-defined target. Picking a published, versioned community default is the honest substitute. Second, the oracle settles only **class-3 no-anchor cosmetics** (below); it never overrides a class-1 or class-2 IL distinction. Where a no-anchor choice is not yet a stable default it is exposed opt-in rather than imposed — see [Line wrapping](#line-wrapping) and [Names](#names).
+
 ## The three-class rule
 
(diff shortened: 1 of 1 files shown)


richlander:
Ready to merge — documentation-only, markdownlint and changes pass, product build/test path-gated (skipped), so no adversarial review required. IL-shape claims verified with a compiled Release canary (
ameof->ldstr, default(int/string/struct)->ldc.i4.0/ldnull/initobj, compound-assignment local/static/this-field byte-identical vs array ldelema+dup).

Issue #3087, opened 2026-07-23
Enum shift's stale enum ResultType misleads comparison, ternary/coalesce, and enum-sibling bitwise consumers (CS0019/CS1503)


richlander:
## Summary

An enum shift's IR `Binary` node reports a **stale enum `ResultType`** (`TypeFamilies.BinaryResult` prefers the enum definition for the flags-enum idiom `flags | 0x20`), even though a shift always yields a primitive integer and the printer renders its left operand as the underlying integer (`ShiftEnumLeftOperand`: `(int)e << n`). Any consumer that trusts `Binary.ResultType` to decide enum coercion is therefore misled and can emit invalid C#.

PR #3078 fixed three **witnessed** consumers of this stale type (bitwise integer-sibling `CS0019`, bitwise chain `CS0019`, and enum-target `CS0266` via `CoerceText`). Adversarial review of #3078 (GPT + Gemini, fixed head `eda3d14f`) confirmed those fixes are sound and non-regressing, but surfaced **three additional pre-existing consumers** of the same stale `ResultType` that remain broken in `main`. They are latent defects independent of #3078 (base `d6af0165` renders identically to head).

## Pre-existing consumers still emitting invalid C#

All shapes below are valid C# inputs whose decompiled output does not recompile.

### 1. Bitwise op with an enum-typed sibling — `CompareText`/`BinaryBody`, CS0019

```csharp
public enum E { A, B }
public static E M(E e, int n, E other) => (E)(((int)e << n) | (int)other);
```

Decompiles to `(int)e << n | other` → `CS0019: Operator '|' cannot be applied to operands of type 'int' and 'E'`.

The bitwise guard added in #3078 only reconciles an **integer** sibling; when the sibling is itself an enum value it renders bare (`other`) while the shift renders as `int`.

### 2. Comparison operators — `CompareText`, CS0019

```csharp
public enum CfgInt : int { None = 0, High = 2 }
public static bool M(CfgInt e, int n) => ((int)e << n) == (int)CfgInt.High;
```

Decompiles to `((int)e << n) == CfgInt.High` → `CS0019: Operator '==' cannot be applied to operands of type 'int' and 'CfgInt'`. `CompareText` calls `TryCoerceEnumOperand(right, left.ResultType)`; the shift's stale enum `ResultType` coerces the right side to the enum while the shift renders as `int`.

### 3. Ternary / null-coalesce into an enum target — `ConditionalArm`/`CoalesceText`, CS1503

[...]

PR #3092, opened 2026-07-23
Collapse redundant underlying cast on reconciled enum left-shift bitwise operands (#3076)

### LINKED ISSUE #3076: Cosmetic: collapse (T)((int)e << n) to (T)e << n in reconciled enum-shift bitwise operands
Follow-up to #3011.

When an enum shift feeds a mixed-sign bitwise op, the reconciliation renders a double cast, e.g. `(uint)((int)e << 24) | x` instead of the cast-minimal `(uint)e << 24 | x`. Both are valid and faithful; the extra cast is cosmetic.

Root: `ShiftEnumLeftOperand` spells the signed underlying cast `(int)e` (signedness follows the shr/shl opcode), then the mixed-sign reconciliation wraps `(uint)(...)`. Collapsing the pair to a single `(uint)e` cast when the outer reinterpret is the unsigned counterpart of the inner underlying type would minimize the spelling.

Non-blocking cosmetic; output is already valid and correct.

richlander:
- Fixes #3076
- Changes reconciled enum **left**-shift bitwise operands to render `(T)e << n` instead of the double-cast `(T)((int)e << n)`
- Card revision: `9281e3e9`

## Change

**Conclusion:** **PASS** — a left shift is bit-identical regardless of operand signedness, so dropping the inner underlying cast is purely cosmetic and fidelity-preserving; the collapse is gated to left shifts and a negative test locks the right-shift guard.

### Original source

```csharp
public static uint IntEnumShiftOrUnsigned(CfgPriority e, uint x) => ((uint)e << 24) | x;
```

### Before

```csharp
public static uint IntEnumShiftOrUnsigned(ILInspector.Decompiler.Tests.CfgPriority e, uint x) => (uint)((uint)((int)e << 24) | x);
```

- Valid: True
- Correct: True

The shift spells its underlying cast with the opcode's signedness (`(int)e << 24`), then the mixed-sign reconciliation wraps it in `(uint)(...)` — an inner double cast `(uint)((int)e << 24)` that is pure noise.

### After

```csharp
public static uint IntEnumShiftOrUnsigned(ILInspector.Decompiler.Tests.CfgPriority e, uint x) => (uint)((uint)e << 24 | x);
```

- Valid: True
- Correct: True

The reconciliation now spells the enum cast directly as the unsigned target (`(uint)e << 24`), removing the redundant inner cast.

### Negative case (right shift is not collapsed)

A signed right shift is arithmetic; `(uint)e >> n` is logical and would change the value, so a reconciled right shift keeps its double cast. Before and After are identical here:

```csharp
public static ulong LongEnumShiftRightOrUnsigned(ILInspector.Decompiler.Tests.CfgLongPriority e, int n, ulong x) => (ulong)((ulong)((long)e >> n) | x);
```

- Valid: True
- Correct: True

### Precedence guard (arithmetic parent keeps parentheses)

The collapsed `(uint)e << n` is a shift fragment. It is bare under bitwise/equality operands (which bind looser than `<<`), but an enum array/by-ref element shift masks the enum as its primitive width, so it can be reconciled inside a mixed-sign `+`/`-`/`*` (which bind tighter than `<<`). There it must keep parentheses:

```csharp
public static uint EnumArrayShiftAddUnsigned(ILInspector.Decompiler.Tests.CfgPriority[] values, int i, int n, uint x) => ((uint)values[i] << n) + x;
```

- Valid: True
- Correct: True

The collapse threads the parent's precedence and wraps via `Rendered.At(context)` — `((uint)values[i] << n) + x`, never `(uint)values[i] << n + x` (which parses as `(uint)values[i] << (n + x)`).

### Fully raised

The residual outer identity wrap (`(uint)( ... | x)`) is a separate rendered-type redundancy — the reconciliation cannot yet see the collapsed operand as unsigned-rendered without instance access to enum widths.

- Required tracking issue: #3087 — central rendered-type fix that owns the outer-cast redundancy (and the other pre-existing mixed-sign consumers).

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `EnumCastPrinterTests`: 51 passed | `EnumCastPrinterTests`: 53 passed |
| Decompiler fast suite | 145, 0 failed | 145, 0 failed |
| Reduced fixture validity | Pass (AssertCompiles) | Pass (AssertCompiles) |
| Reduced fixture fidelity | Pass | Pass |

Head adds two tests — `EnumRightShiftInBitwiseOr_KeepsDoubleCast` (negative right-shift case) and `EnumArrayShiftInArithmeticAdd_KeepsShiftParentheses` (precedence guard) — plus their fixtures; the 51 pre-existing focused tests pass on both sides. Fidelity: `(uint)e << n` == `(uint)((int)e << n)` and the `ulong`/`long` variant verified bit-identical across high-bit and negative inputs.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class "ILInspector.Decompiler.Tests.EnumCastPrinterTests"
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -trait "Area=RoundTrip" -trait- "Speed=Slow"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -5179,6 +5179,13 @@ public static class EnumCastSamples
 
     public static uint IntEnumShiftAndUnsigned(CfgPriority e, uint x) => ((uint)e << 4) & x;
 
+    // Negative case for the #3076 left-shift cast collapse: a signed arithmetic
+    // right shift reconciled against an unsigned sibling must KEEP its double cast
+    // `(ulong)((long)e >> n)`. Collapsing to `(ulong)e >> n` would switch the
+    // arithmetic shift to a logical one and change the value, so the collapse is
+    // left-shift only.
+    public static ulong LongEnumShiftRightOrUnsigned(CfgLongPriority e, int n, ulong x) => ((ulong)((long)e >> n)) | x;
+
     // A bitwise CHAIN over an enum shift: the inner `|` inherits the shift's stale
     // enum ResultType while rendering as an integer, so the outer `|` must not
     // coerce the far sibling to the enum (`... | (E)y`, CS0019). The rewritten-
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -5186,6 +5186,15 @@ public static class EnumCastSamples
     // left-shift only.
     public static ulong LongEnumShiftRightOrUnsigned(CfgLongPriority e, int n, ulong x) => ((ulong)((long)e >> n)) | x;
 
+    // Precedence guard for the #3076 collapse: an enum left shift reconciled inside
+    // a mixed-sign ARITHMETIC parent (`+`/`-`/`*`, which bind tighter than `<<`)
+    // must keep parentheses around the collapsed shift — `((uint)values[i] << n) + x`,
+    // not `(uint)values[i] << n + x` (which parses as `(uint)values[i] << (n + x)`).
+    // An enum ARRAY element masks the enum as its primitive width, so the shift's
+    // EffectiveType is an integer and MixedSignArithmetic reconciles it (a plain
+    // enum field stays enum-typed and never routes here).
+    public static uint EnumArrayShiftAddUnsigned(CfgPriority[] values, int i, int n, uint x) => ((uint)values[i] << n) + x;
+
     // A bitwise CHAIN over an enum shift: the inner `|` inherits the shift's stale
     // enum ResultType while rendering as an integer, so the outer `|` must not
     // coerce the far sibling to the enum (`... | (E)y`, CS0019). The rewritten-
--- a/src/ILInspector.Decompiler.Tests/EnumCastPrinterTests.cs
+++ b/src/ILInspector.Decompiler.Tests/EnumCastPrinterTests.cs
@@ -258,6 +258,20 @@ public class EnumCastPrinterTests
         AssertCompiles("public static ulong M(CfgLongPriority e, int n, ulong x)", body, "public enum CfgLongPriority : long { Low, Medium = 1, High = 2, Critical = 3 }");
     }
 
+    // #3076 precedence guard: an enum array-element left shift reconciled inside a
+    // mixed-sign ARITHMETIC parent (`+`, which binds tighter than `<<`) must keep
+    // parentheses — `((uint)values[i] << n) + x`, never `(uint)values[i] << n + x`
+    // (which parses as `(uint)values[i] << (n + x)` and fails to bind).
+    [Fact]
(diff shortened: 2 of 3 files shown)


richlander:

Two-model review per repo policy (roster: Claude Opus, Gemini Pro, GPT, MAI-Code). I'm Claude Opus, so I delegated to two other families: **GPT (`gpt-5.6-sol`)** and **Gemini Pro (`gemini-3.1-pro`)**. Each reviewer ran in an isolated linked worktree at a fixed head, scratch under `/tmp`, no `git reset`/`add`/commit.

### Round 1 — pre-fix head `4127eece`

- **GPT — BLOCKED (1 finding).** Precedence bug: the collapse returned a bare `(uint)e << n` shift fragment. That is safe under bitwise `&`/`|`/`^` and equality operands (which bind looser than `<<`), but `BitwiseUnsignedOperand` is also reached from mixed-sign arithmetic (`+`/`-`/`*`, which bind **tighter** than `<<`). An enum **array/by-ref element** left shift masks the enum as its primitive width, so its `EffectiveType` is an integer and `MixedSignArithmetic` reconciles it — routing the shift through this helper in an arithmetic parent. Result: `(uint)values[i] << n + x`, which mis-parses as `(uint)values[i] << (n + x)` (CS0019).
  - **Reproduced** exactly at `4127eece` with fixture `EnumArrayShiftAddUnsigned`.
- **Gemini — CLEAN**, but its precedence analysis was **wrong**: it argued the collapse was reachable only via bitwise operators because an enum shift's `EffectiveType` is always the enum. That holds for a plain enum **field**, but not for an enum **array/by-ref/pointer element** (the load masks the enum as its primitive width). GPT was correct; this was a Gemini false-negative.

### Resolution — commit `9281e3e9`

Thread the parent operator's precedence into `BitwiseUnsignedOperand` (new `Precedence context` param) and wrap the collapsed shift via the existing `Rendered(collapsed, Precedence.Shift).At(context)` helper. Call sites 729/734 pass `context: demand` (the parent's precedence); the equality path keeps the default `Precedence.Shift` (no wrap — `==`/`!=` bind looser than `<<`). Arithmetic parents now get `((uint)values[i] << n) + x`; bitwise/equality parents stay paren-free. Regression fixture/test added: `EnumArrayShiftAddUnsigned` / `EnumArrayShiftInArithmeticAdd_KeepsShiftParentheses`.

### Round 2 — fixed head `9281e3e9` (re-review)

- **GPT — CLEAN.** Confirmed the finding is resolved (`((uint)values[i] << n) + x` compiles); verified multiplicative parents (`* x`, `x *`, `x -`), both-operand associativity, unchanged bitwise/right-shift/equality renders. `EnumCastPrinterTests` 53/53, fast RoundTrip 145/145.
- **Gemini — CLEAN.** Re-examined and corrected its round-1 reasoning; verified the additive/multiplicative wrap, the `/`·`%` non-reachability (excluded from `MixedSignArithmetic`), no spurious parentheses on looser parents, and the equality default. `EnumCastPrinterTests` 53/53, fast RoundTrip 145/145.

richlander:
Ready to merge

PR #3093, opened 2026-07-23
Harness: --verify-authored-corpus drift check for the vendored corpus [#3052]

richlander:
## Summary

Adds `--verify-authored-corpus` to the decompiler harness — the drift-detection
mode from #3052 (Scope C). It audits whether the vendored authored-source
correspondence corpus still corresponds to the authoritative source at each row's
pinned SourceLink commit, **without running the decompiler**.

For every corpus row it re-acquires the authored source *today* — from a local git
clone (`--repo`, checksum-arbitrated) and/or SourceLink — reduces it to the same
member-body slice the harvester stored (`AuthoredRebuildFidelity.TryExtractTargetBody`),
and compares it (newline-normalized) against the stored `AuthoredBody`. Each row is:

- **Verified** — re-acquired body matches the vendored snapshot.
- **Drifted** — re-acquired body differs (report names the row + a short
  line-count/first-diff summary).
- **Unavailable** — source could not be re-acquired or sliced (offline with no
  `--repo`, commit gone, checksum mismatch, extraction regression); surfaced,
  never silently dropped.

Report-only by default (`--json` for structured output). The run's own integrity
still governs the exit code with a named blocker — any corpus row whose assembly
was not supplied, or a run that evaluated no rows, fails, so an empty or
partially-unmatched run is never a success. `--fail-on-drift` additionally exits
nonzero when any row has drifted, the shape of a periodic (non-PR) gate.

## Why

Completes the hybrid plan in #3052: the vendored orphan-branch corpus stays
hermetic (it is the benchmark's source of truth), and this mode is the drift
signal that catches the snapshot rotting away from upstream — reusing the same
`--repo` local-source path shipped in #3040/#3056.

## Boundary / non-goals

- Does **not** run the decompiler and never affects `AuthoredCorpusBenchmark`.
- Does not gate the PR benchmark; it is a standalone, opt-in integrity check.
- No corpus schema change; consumes the existing `CorpusRecord` rows.

## Evidence (real runs, this machine)

Harvested a 6-row self-corpus from the pinned `dotnet-inspect.any` 0.14.0
`ILInspector.Metadata.dll` with `--repo <checkout>`, then:

| Scenario | Result | Exit |
| --- | --- | --- |
| Clean corpus, `--repo` | Verified 6, Drifted 0, Unavailable 0 | 0 |
| One row's `authoredBody` tampered, report-only | Verified 5, **Drifted 1** (`stored 10 line(s), acquired 9 line(s); first diff at line 10`) | 0 |
| Tampered corpus, `--fail-on-drift` | drift gate trips | **1** |
| One row's `metadataToken` corrupted | Verified 5, **Unavailable 1** (`acquire: no source text`) | 0 |
| Non-corpus assembly supplied | `unmatchedRows > 0`, BLOCKER, nothing checked | **1** |

`--help` lists `--verify-authored-corpus`/`--fail-on-drift`; `--json` serializes
the counts + per-row outcomes. Harness builds clean; changed Markdown passes
`markdownlint` (MD013/032/034/060 disabled by `.markdownlint.yaml`).

## Risk

Harness-only, additive new mode; no product path touched. Low risk.

### DIFF
--- a/tools/DecompilerHarness/AuthoredCorpusDrift.cs
+++ b/tools/DecompilerHarness/AuthoredCorpusDrift.cs
@@ -1,0 +1,328 @@
+using System.Text.Json;
+
+using DotnetInspector.Core;
+using DotnetInspector.Services;
+using ILInspector.Findings;
+using ILInspector.Metadata;
+
+namespace ILInspector.DecompilerHarness;
+
+/// <summary>
+/// Drift verification for the vendored authored-source correspondence corpus.
+/// Every corpus row snapshots a checksum-verified authored member body captured
+/// at harvest time. This mode re-acquires each row's source <em>today</em> — from
+/// a local git clone (<c>--repo</c>, checksum-arbitrated) and/or SourceLink — and
+/// reduces it to the same member-body slice the harvester used
+/// (<see cref="AuthoredRebuildFidelity.TryExtractTargetBody"/>), then compares it
+/// against the stored body. It answers "does the vendored snapshot still
+/// correspond to the authoritative source at its pinned commit?" without running
+/// the decompiler: a byte-for-byte mismatch is corpus drift; an acquisition or
+/// slice miss is an unresolved row.
+///
+/// This is a standalone integrity check, not a benchmark gate: it does not run
+/// the decompiler and never affects <see cref="AuthoredCorpusBenchmark"/>. It is
+/// report-only by default — a run's own integrity (corpus present, every corpus
+/// assembly supplied, at least one row evaluated) still governs the exit code, so
+/// a partial or empty run never masquerades as success. Pass <c>--fail-on-drift</c>
+/// to additionally fail the run when any row has drifted, for a periodic
+/// (non-PR) gate.
+/// </summary>
+static class AuthoredCorpusDrift
+{
+    enum Outcome
+    {
+        // Re-acquired body matches the stored snapshot.
+        Verified,
+        // Re-acquired body differs from the stored snapshot (source rot).
+        Drifted,
+        // Source could not be re-acquired or sliced (offline, no --repo, commit
+        // gone, checksum mismatch, or an extraction regression) — unverifiable.
+        Unavailable,
+    }
+
+    sealed record RowResult(
+        AuthoredSourceHarvest.CorpusRecord Record,
+        Outcome Outcome,
+        string? Detail);
+
+    public static int Run(
+        IReadOnlyList<string> assemblies,
+        string corpusPath,
+        bool json,
+        bool failOnDrift,
+        IReadOnlyList<string>? repositoryPaths)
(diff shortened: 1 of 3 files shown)


richlander:

Two independent fixed-head reviews from different model families (per the AGENTS.md roster, not the author's own):

- **Gemini 3.1 Pro** (`gemini-3.1-pro-preview`)
- **GPT-5.6** (`gpt-5.6-terra`)

Each ran in an isolated detached review worktree (read-only; scratch under `$env:TEMP`), built the harness, and reproduced real runs.

### Round 1 (head `9bb86df0`) — one finding, both reviewers agreed

**`--fail-on-drift` failed open when every row was `Unavailable`.** A periodic gate could exit `0` after an outage, a missing PDB, or a bad `--repo` — having verified nothing (a gate that passes while establishing no correspondence). Gemini rated it High; GPT rated it Medium/High. Both endorsed failing the gate on `Unavailable`.

All other attack points were verified **safe** and dismissed with evidence:

- **Like-for-like slicing** — the drift check passes exactly `record.Method` + `record.ParameterCount` to `TryExtractTargetBody`, matching harvest; no skew.
- **No spurious `Verified`** — `AcquireMemberAsync` enforces `VerifyChecksum` strictly for both the local `--repo` blob and remote fetches; a wrong repo/commit self-rejects to `Unavailable`, never a false `Verified`.
- **Newline normalization** — byte-level drift correctly flips to `Drifted` (verified with interior + trailing edits).
- **Resource correctness** — `SourceLinkService` disposed in `finally`; the assembly-open `catch` classifies the whole group `Unavailable` (no row silently dropped).
- **Honest-exit accounting** — unsupplied assembly / zero-rows correctly exit `1`.

### Resolution — `c7445999`

Made `--fail-on-drift` a **fail-closed** gate: both exit sites (card + JSON) now return `honest && !(failOnDrift && (drifted > 0 || unavailable > 0)) ? 0 : 1`, so the gate requires every evaluated row to be `Verified`. Report-only mode (no flag) is unchanged and stays a pure diagnostic (exit `0` regardless of `Unavailable`). Doc comment, README, and usage text updated.

### Round 2 (fixed head `c7445999`) — both CLEAN

Both reviewers re-ran on the fixed head and confirmed the finding resolved with **no new issues** and no regression. Verified exit codes on both:

| Scenario | Exit |
| --- | --- |
| All `Unavailable` + `--fail-on-drift` (the original repro) | **1** |
| Clean corpus + `--fail-on-drift` | 0 |
| Genuine drift + `--fail-on-drift` | **1** |
| All `Unavailable`, report-only (no flag) | 0 |
| Unsupplied corpus assembly (honest-exit blocker) | **1** |

richlander:
Ready to merge — required CI green (test, markdownlint, changes; pack/smoke path-gated and correctly skipped) and two clean fixed-head adversarial reviews (Gemini 3.1 Pro + GPT-5.6) on `c7445999`.

PR #3094, opened 2026-07-23
Raise inner comparison-chain value switch expression (#2978 slice 1)

richlander:
- Advances #3113 (raise follow-up to the #2978 / #3063 witness; #2978 itself is already closed)
- Changes: raise the inner comparison-chain value `switch` expression inside `Dispatch`'s string arm, so `s.Length switch { 0 => .., 1 => .., _ => .. }` is recovered instead of a nested `if/else` cascade
- Card revision: `0217589e`

## Change

**Conclusion:** **PASS** — a surgical, shape-gated raise that recovers the inner switch expression with no corpus delta and no new test failures; both adversarial reviews were clean at the prior head and the merged head keeps the same 3-file diff (re-review pending on the new head — see note under Decompiler quality).

### Original source

```csharp
public static bool Dispatch(object o, out int x) => o switch
{
    string s => s.Length switch
    {
        0 => Fail(out x),
        1 => Win(1, out x),
        _ => Fail(out x)
    },
    int i when i > 0 => Win(i, out x),
    _ => Fail(out x)
};
```

### Before

```csharp
public static bool Dispatch(object o, out int x)
{
    int i;
    int V_5;

    object V_3 = o;
    string s = V_3 as string;
    if (s is null)
    {
        if (V_3 is not int)
        {
            return Fail(out x);
        }
        i = (int)V_3;
    }
    else
    {
        V_5 = s.Length;
        if (V_5 != 0)
        {
            if (V_5 == 1)
            {
                return Win(1, out x);
            }
            return Fail(out x);
        }
        return Fail(out x);
    }
    if (i <= 0)
    {
        return Fail(out x);
    }
    return Win(i, out x);
}
```

- Valid: True
- Correct: True

The inner `s.Length` dispatch is a valid, behavior-preserving `if/else` cascade, but it is lower-quality than the original switch expression.

### After

```csharp
public static bool Dispatch(object o, out int x)
{
    int i;

    object V_3 = o;
    string s = V_3 as string;
    if (s is null)
    {
        if (V_3 is not int)
        {
            return Fail(out x);
        }
        i = (int)V_3;
    }
    else
    {
        return s.Length switch
        {
            0 => Fail(out x),
            1 => Win(1, out x),
            _ => Fail(out x),
        };
    }
    if (i <= 0)
    {
        return Fail(out x);
    }
    return Win(i, out x);
}
```

- Valid: True
- Correct: True

The inner arm is now recovered as `s.Length switch { 0 => .., 1 => .., _ => .. }`. The outer type-pattern dispatch is still statement-form (tracked below); the whole method remains valid and behavior-preserving.

### Fully raised

```csharp
public static bool Dispatch(object o, out int x) => o switch
{
    string s => s.Length switch
    {
        0 => Fail(out x),
        1 => Win(1, out x),
        _ => Fail(out x)
    },
    int i when i > 0 => Win(i, out x),
    _ => Fail(out x)
};
```

- Required tracking issue: #3113 — recover the `string s` type-pattern arm (carrying the already-raised inner switch), recover the `int i when i > 0` arm (`isinst`/unbox + `when` guard, collapsing the duplicated default), and compose the outer `object`-governed type-pattern switch expression.

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `ComparisonChainSwitchExpressionRaisingTests`: not present (class added in PR) | `ComparisonChainSwitchExpressionRaisingTests`: 7 passed |
| Decompiler fast suite | 3413, 0 failed | 3420, 0 failed |
| Reduced fixture validity | not applicable | not applicable |
| Reduced fixture fidelity | not applicable | not applicable |
| Real witness | `ScatteredReturnDispatchSample::Dispatch` inner arm not raised (if/else cascade) | `ScatteredReturnDispatchSample::Dispatch` inner arm raised to `s.Length switch { .. }` (outer dispatch pending #2978) |

Baseline = `a223bded` (origin/main); head = `0217589e`. The head/baseline fast-suite delta is exactly the +7 focused tests added by this PR; no baseline test changed state.

## Decompiler quality

**Conclusion:** **PASS** — the PR quick gate matched baseline tolerances (0 pp on every gated metric, 0 pass bugs). The generated card is byte-identical when run at base `a223bded` and head `0217589e`, so this PR introduces zero corpus delta on the sampled 1,500 methods (the narrow comparison-chain-switch-expression shape is not exercised by the sample).

### PR quick gate

Run: PR quick corpus, hash-stable 100 methods per assembly (15 assemblies, 1,500 methods); validity not run, fidelity not run (`--compile-cap 0 --corpus-fidelity-cap 0`).

| --- | ---: | ---: | ---: |
| Pass bugs (-) | 0 | 0 | 0 |
| Fully raised (+) | not reported by PR-quick card | not reported by PR-quick card | n/a |

> **Conclusion:** **PASS** — pinned-subset gate unchanged at base and head.

### Aggregate context

Corpus: 15 assemblies, 1,500 methods. Baseline drift: the card shows a +2-method improvement (251 → 249 detected residue) relative to the pinned `pr-quick-baseline.json` snapshot; that drift is present identically at both base and head, so it predates this PR's diff and is not PR-attributable.

| Metric (goal) | Baseline (base render) | PR (head render) | Count delta |
| --- | ---: | ---: | ---: |
| Detected lowering residue (-) | 249 (16.60%) | 249 (16.60%) | 0 |
| Conditional-branch residue (-) | 44 (2.93%) | 44 (2.93%) | 0 |
| Forward-merge stops (-) | 30 (2.00%) | 30 (2.00%) | 0 |
| Pass bugs (-) | 0 | 0 | 0 |

> **Conclusion:** **PASS** — base and head renders are byte-identical on the sampled corpus; no method changed direction.

> **Re-review note:** the branch now merges current `origin/main` (which since landed the #3082 switch-raise pass refactor). This PR's net diff versus `origin/main` is unchanged (3 files: `SwitchRaisingPass.cs`, `ScatteredReturnDispatchSample.cs`, `ComparisonChainSwitchExpressionRaisingTests.cs`) and merges without conflict, but the surrounding pass moved, so the prior clean adversarial reviews at `abad3365` need re-confirmation on head `0217589e` before this supersedes the earlier "Ready to merge".

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -filter '/*/*/ComparisonChainSwitchExpressionRaisingTests/*'

# PR quick corpus gate (run at base a223bded and head 0217589e; byte-identical):
dotnet build src/dotnet-inspect -c Release -p:PublishAot=false
bash eng/prepare-decompiler-pr-corpus.sh /tmp/pr-corpus-assemblies.txt
mapfile -t assemblies < /tmp/pr-corpus-assemblies.txt
dotnet run --project tools/DecompilerHarness -c Release -- "${assemblies[@]}" \
  --diff-corpus-baseline tools/DecompilerHarness/corpus/pr-quick-baseline.json \
  --quality-diff-card --corpus-method-cap 100 --compile-cap 0 --corpus-fidelity-cap 0 --max-examples 24
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ComparisonChainSwitchExpressionRaisingTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ComparisonChainSwitchExpressionRaisingTests.cs
@@ -84,6 +84,82 @@ public class ComparisonChainSwitchExpressionRaisingTests
         Assert.Empty(function.Descendants.OfType<SwitchExpression>());
     }
 
+    // Builds the exact comparison-chain value-switch shape the raiser accepts —
+    // `v == 0 => 10, v == 1 => 20, _ => 30` storing one result temp — parameterized
+    // by the governing value's declared type. With <paramref name="extraTempRead"/>
+    // the join reads the result temp a second time, so the temp is no longer the
+    // single-read convergence signal a real switch expression produces.
+    static IrFunction BuildComparisonChainSwitch(TypeRef governingType, bool extraTempRead = false)
+    {
+        var int32 = TypeRef.CoreLib("System", "Int32");
+        LoadArgument V() => new(0, "v", governingType);
+        Comparison Eq(int label) => new(ComparisonKind.Equal, isUnsigned: false, V(), new Constant(label, int32));
+
+        Block ValueBlock(int offset, int result) =>
+            AddAll(new Block(offset), new StoreLocal(0, int32, new Constant(result, int32)), new Branch(0x50));
+
+        var head = AddAll(new Block(0x00), new ConditionalBranch(Eq(0), 0x40));   // v == 0 => arm0
+        var test1 = AddAll(new Block(0x10), new ConditionalBranch(Eq(1), 0x30));  // v == 1 => arm1
+        var defaultArm = ValueBlock(0x20, 30);
+        var arm1 = ValueBlock(0x30, 20);
+        var arm0 = ValueBlock(0x40, 10);
+        var join = new Block(0x50);
+        if (extraTempRead)
+            join.Add(new StoreLocal(1, int32, new LoadLocal(0, int32)));   // a second read of the result temp
+        join.Add(new Return(new LoadLocal(0, int32)));                     // (single read in the faithful shape)
+
+        var container = new BlockContainer();
+        foreach (var block in new[] { head, test1, defaultArm, arm1, arm0, join })
+            container.Add(block);
+
+        var signature = new MethodSignature(int32, [new Parameter("v", governingType)], HasThis: false, GenericParameterCount: 0);
+        return new IrFunction("M", TypeRef.CoreLib("System", "Sample"), signature, [int32, int32], container);
+    }
+
+    static Block AddAll(Block block, params IrNode[] children)
+    {
+        foreach (var child in children)
+            block.Add(child);
+        return block;
+    }
+
+    static bool RaisesSwitchExpression(IrFunction function)
+    {
+        new SwitchRaisingPass().Run(function, PassContext.None);
+        return function.Descendants.OfType<SwitchExpression>().Any();
+    }
+
+    [Fact]
+    public void ReusedResultTemp_DeclinesAtConvergenceCheck()
+    {
+        // Isolates the single-read convergence guard. csc's real switch-expression
+        // lowering always reads its result temp exactly once (any downstream reuse
+        // reads a *later* copy), so a genuine multi-read of the result temp only
+        // arises in hand-written or obfuscated IL. The single-read twin raises,
+        // proving the shape is accepted, so the extra-read variant's decline is
+        // the convergence check's doing — not an earlier rejection.
+        Assert.True(RaisesSwitchExpression(
+            BuildComparisonChainSwitch(TypeRef.CoreLib("System", "Int32"))));
+        Assert.False(RaisesSwitchExpression(
+            BuildComparisonChainSwitch(TypeRef.CoreLib("System", "Int32"), extraTempRead: true)));
+    }
+
+    [Fact]
+    public void BooleanGoverningValue_DeclinesSwitchExpressionRaise()
+    {
+        // Self-validating twin: the identical chain on an Int32 governing value
+        // raises, proving the synthetic shape is one the raiser genuinely accepts.
+        Assert.True(RaisesSwitchExpression(
+            BuildComparisonChainSwitch(TypeRef.CoreLib("System", "Int32"))));
+
+        // So on a Boolean governing value the decline is the bool guard's doing.
+        // csc never lowers a bool `switch` to this equality chain; the guard
+        // defends the arbitrary/obfuscated-IL path from printing an invalid
+        // `b switch { 0 => …, 1 => … }` (int is not convertible to bool, CS0029).
+        Assert.False(RaisesSwitchExpression(
+            BuildComparisonChainSwitch(TypeRef.CoreLib("System", "Boolean"))));
+    }
+
     [Fact]
     public void UnsignedGoverningValue_DeclinesToAvoidInvalidLabel()
     {
--- a/src/ILInspector.Decompiler.Tests/ComparisonChainSwitchExpressionRaisingTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ComparisonChainSwitchExpressionRaisingTests.cs
@@ -83,4 +83,22 @@ public class ComparisonChainSwitchExpressionRaisingTests
 
         Assert.Empty(function.Descendants.OfType<SwitchExpression>());
     }
+
+    [Fact]
+    public void UnsignedGoverningValue_DeclinesToAvoidInvalidLabel()
+    {
+        // SwitchExpressionArm labels are signed int32. A uint governing value
+        // whose label is uint.MaxValue (IL ldc.i4.m1) would misprint as -1
+        // (CS0031). The raiser must decline on a non-Int32 governing type.
+        var function = Raised(
+            typeof(ScatteredReturnDispatchSample).FullName!,
+            nameof(ScatteredReturnDispatchSample.ClassifyUnsigned));
+
+        Assert.Empty(function.Descendants.OfType<SwitchExpression>());
+
+        var output = Print(
+            typeof(ScatteredReturnDispatchSample).FullName!,
+            nameof(ScatteredReturnDispatchSample.ClassifyUnsigned));
+        Assert.DoesNotContain("-1 =>", output);
+    }
 }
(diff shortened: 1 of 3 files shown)


richlander:
## Adversarial review reconciliation (head `abad3365`)

Two-model adversarial review (GPT `gpt-5.6-sol` + Gemini `gemini-3.1-pro-preview`), reviewers isolated in separate `/tmp/` worktrees. Both surfaced real correctness findings; both re-reviewed the fixed exact head and are now clean.

### Findings and resolutions

**Round 1 — GPT (High, confirmed):** The raiser recorded switch-expression arm labels as signed `int32`, so a `uint` switch with a `uint.MaxValue` label (IL `ldc.i4.m1`) raised to `-1 => …` — invalid C# (`CS0031`); `CheckInvariant()` passed but the reprint did not recompile.
→ Fixed in `fa432f0a` by gating the raise on the governing value's type.

**Round 2 — Gemini (High, confirmed):** The first fix's `Int32`-only gate was over-broad: `byte`/`short`/`ushort`/`enum` comparison-chain switch expressions have non-negative `int32` labels that print faithfully (enum labels render as member names via the governing type) and were regressed to `if`/`else` cascades. I **independently reproduced** this on the exact prior head (guard off → they raise; `Int32`-only guard → they degrade), which is why I narrowed rather than kept it.
→ Fixed in `abad3365`: the gate now declines only the actual failure mode — a `UInt32` governing value **and** some `caseLabel < 0`. Added positive `enum`/`byte` fixtures + tests (`EnumGoverningValue_RaisesWithMemberNames`, `ByteGoverningValue_RaisesToSwitchExpression`) to lock in the preserved raises; `ClassifyUnsigned(uint)` remains the close negative.

**Fixed-head (`abad3365`) re-reviews — both clean:**
- GPT: *"No significant issues found in the reviewed fix."*
- Gemini: *"Clean. No correctness issues, regressions, or edge cases found."* — independently verified `uint`-backed enum handling, `long`/`ulong`/`nint`/`nuint` immunity (they fail `TryEqualityLabel`'s `int` constant match), and that `caseLabels` is fully populated at the guard.

### Out of scope (pre-existing, not introduced here)

A `uint` switch with out-param **call** arms hits a `NullReferenceException` in `TypedConstantsPass`. Reproduced on base `origin/main` (`33ea0bdf`) independent of this change; the narrowed guard makes `uint` behave exactly as base. Not addressed in this PR.

### Validation
- Focused suite `*ComparisonChainSwitchExpressionRaisingTests*`: 7/7 pass.
- Fast decompiler suite (`-trait- Speed=Slow`): 3408 pass, 0 fail.
- Load-bearing evidence: with the guard removed, the Release xunit `UnsignedGoverningValue` test raises `-1 => 20` and fails; with the guard it declines.
- CI on `abad3365`: `test`, `build-net10`, `csharp-diff-smoke`, `changes` all green (path-gated jobs skipped).

### Scope note
This is **slice 1 of the #2978 full-raise**: it raises the inner `s.Length switch` comparison-chain value switch expression. The outer type-pattern dispatch of the witness (`ScatteredReturnDispatchSample.Dispatch`) is still statement C#; slices 2–4 remain as non-blocking follow-up.

richlander:
Ready to merge

richlander:
Head advanced from `abad3365` to `0217589e` (merged current `origin/main`, which since landed the #3082 switch-raise pass refactor). No conflicts; this PR's net diff versus `origin/main` is unchanged (3 files). CI is re-running on the new head and the two clean adversarial reviews at `abad3365` need re-confirmation on `0217589e` — so the earlier **Ready to merge** is superseded pending that re-review. Also refreshed the PR description to the decompiler PR template (renders, evidence, and PR-quick corpus card all taken at base `a223bded` vs head `0217589e`).

richlander:
## Adversarial review reconciliation (two models)

Reviewed at exact head, isolated review worktrees, `/tmp`-only scratch, no git mutations in review trees.

**Reviewers:** GPT (`gpt-5.6-sol`) and Gemini Pro (`gemini-3.1-pro-preview`) — two different models per the roster.

### First pass — head `0217589e`
- **Gemini: CLEAN.** Empirically compiled a real `bool` switch → bare `brtrue`/`brfalse` with no convergence temp; it structurally never reaches the raiser. All attack points sound.
- **GPT: BLOCKED (1)** — no Boolean guard: a hand-crafted `!b` / `b == 1` chain would raise to `b switch { 0 => …, 1 => … }` (invalid, CS0029). **Plus a non-blocking test-integrity note**: the pre-existing `HandWrittenEqualityChain` negative declines earlier at `TryValueBlock`, so it never exercises the single-read convergence guard.

### Investigation of GPT's blocking finding → compiler-unreachable
Verified three independent ways that csc never produces the triggering shape: (1) Gemini's empirical compile; (2) my own reproduction — even a bool switch with a single-read temp stays `if (!b)/else`, declined because a bool yields only label `0` from `!b` and never a real `b == 1`, so `sawEquality` stays false and `caseLabels.Count` stays `1`; (3) the guard logic itself (`sawEquality && caseLabels.Count >= 2` cannot both hold for a csc bool switch). GPT's repro was hand-written IL.

### Hardening applied — follow-up commit `653afeef`
Even though unreachable from csc, I added the defensive guard GPT recommended and closed the test gap:
- **Bool guard** in `RaiseComparisonChainSwitchExpression`, right after the UInt32 negative-label guard: decline when the governing type is `System.Boolean`. Mirrors the existing invalid-label rationale; defends arbitrary/obfuscated non-csc IL.
- **Two synthetic-IR decline tests**, each with a self-validating Int32 twin that raises (proving the shape is genuinely accepted, so each decline is attributable to the specific guard):
  - `ReusedResultTemp_DeclinesAtConvergenceCheck` — isolates the single-read convergence guard (GPT's test-integrity note). The multi-read decline is itself csc-unreachable, since csc always reads its switch result temp exactly once (downstream reuse reads a *later* copy), so a synthetic IR test is the correct tool.
  - `BooleanGoverningValue_DeclinesSwitchExpressionRaise` — isolates the new bool guard.

### Second pass — fixed head `653afeef` (both CLEAN)
- **GPT: CLEAN** — bool guard correctly scoped; twins prove both declines reach the intended guards; extra local + `PassContext.None` correct. Build 0/0; focused **9/9**; fast suite **3422/0**.
- **Gemini: CLEAN** — guard rejects strictly and safely; other governing types unaffected (byte/enum still raise faithfully); synthetic containers faithfully model the raiser's structural gating. Build success; focused **9/9**; fast suite clean.

Both fixed-head reviews are clean. No behavior change to any reachable csc shape; the follow-up is purely defensive hardening plus decline-path coverage.

richlander:
Ready to merge

Issue #2021, opened 2026-06-30
Emit minimal C# unsafe expressions in decompiler output


richlander:
## Summary

C# now supports `unsafe(expr)`, and the Decompiler can use it as the narrowest
synthesized unsafe context. IL does not reveal whether the original source used
an unsafe block or expression, so output must not claim lexical recovery.

## Oracle and expected behavior

Follow dotnet/runtime's migrated source where available and its memory-safety
analyzer/fixer for placement while most runtime source remains unmigrated:

- prefer `unsafe(expr)` when one expression and its dependencies can be
  isolated;
- otherwise emit the smallest `unsafe` block that preserves statement, scope,
  and data-flow semantics;
- keep semantically safe statements outside the context; and
- widen only when dependencies such as ref-local lifetime require it.

This is a reconstruction policy, not evidence that source used the same form.

## Current gate

The repository SDK accepts `unsafe(expr)`, but the compile-back rail's pinned
`Microsoft.CodeAnalysis.CSharp` package does not yet parse it. Keep block output
until that rail can validate expression output, then cover legal expression
positions rather than limiting the feature to returns.

The focused implementation and design work is also tracked by #5255.

richlander:
## Status: the compiler gate is now met

The frontier this issue tracked has opened. Both upstream items are **closed/implemented**:

- roslyn#84012 — *Unsafe evolution: unsafe expressions*
- csharplang#10196 — *Add `unsafe` expressions*

### Reproduction on current SDK (11.0.100-preview.7)

The syntax guessed in the original summary — `byte* ptr = unsafe stackalloc byte[64];` — never shipped; it is a **syntax error** (`CS1003: Syntax error, '(' expected`). The form that actually shipped is **`unsafe(expression)`**, per the speclet (csharplang#10196):

> An `unsafe_expression` is `'unsafe' '(' expression ')'` … it establishes an `unsafe` context for evaluating its *expression* … the `unsafe` context does not extend beyond its closing parenthesis.

Confirmed compiling and running from a fully safe context on preview 7:

```csharp
#:property AllowUnsafeBlocks=true
#:property LangVersion=preview
int r = unsafe(GetUnsafe());   // prints 123
static unsafe int GetUnsafe() { int x = 123; int* p = &x; return *p; }
```

### Decompiler altitude

[...]

PR #3095, opened 2026-07-23
docs: record unsafe(expr) gate met, emission deferred (#2021)

richlander:
## Summary

Documentation-only change recording the current state of C# 14 `unsafe(expr)` support for issue #2021.

**Conclusion:** the compiler gate is met (roslyn #84012 / csharplang #10196 shipped; `unsafe(expression)` parses/compiles on the preview.7 SDK this repo builds with), but **emission stays deferred** because the compile-back rail's pinned `Microsoft.CodeAnalysis.CSharp 5.6.0` (latest public package) cannot parse `unsafe(expr)` — it reads `unsafe` as the block keyword. Narrowing the introduced `unsafe { }` wrap now would break every new-rules compile-back harness.

We keep the block wrap until the pinned Roslyn advances, and document the return-position-only narrowing rule for when emission lands.

## Investigation notes
- Prototyped return-position narrowing (`return unsafe(<expr>);`); it fired across ~18 new-rules fixtures and broke the pinned-Roslyn compile-back rails, plus a `ref`-return correctness edge. Reverted; only the design note remains.

## Evidence
- Documentation-only; `markdownlint` clean. No product build/test claim.

## Risk
Low — doc-only, no behavior change. No adversarial review required.

### DIFF
--- a/docs/design/memory-safety-modes.md
+++ b/docs/design/memory-safety-modes.md
@@ -92,9 +92,24 @@ pointerless `unsafe` method's requires-unsafe-ness. Legacy compilation stamps no
 `RequiresUnsafeAttribute` and the call carries no pointer, so the fact was erased
 — there is nothing to replay or recover. This is the principled limit of the mode.
 
+The `unsafe(expr)` compiler gate is now met: roslyn #84012 / csharplang #10196
+shipped, and the syntax `unsafe(expression)` parses and compiles on the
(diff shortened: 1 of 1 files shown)


richlander:
Ready to merge — documentation-only change; `markdownlint` passed and product jobs are correctly path-skipped. Low risk, no adversarial review required.

Issue #3091, opened 2026-07-23
RTS: consume product ApiSurfaceExtractor for member-surface reconstruction instead of hand-rolled enumeration


richlander:
## Summary

RTS's compile-back type-surface reconstruction hand-rolls a metadata→member-surface
enumeration (`CompileBackSourceComposer.AddClosureMemberSurface` / `TypeProducer` in
`tools/DecompilerHarness/ReturnToSenderTypePlanner.cs`) that duplicates the product's
`ILInspector.Metadata.ApiSurfaceExtractor.Extract(pe, includeAll: true)`. This duplication
is the source of a class of evidence-integrity bugs — field-like event double-emit (#3083)
and explicit-interface event double-emit (found during PR #3075 review) — where the
hand-rolled enumeration diverges from the product's surface model but still reports
`BodyComplete=true`.

Per the architectural principle *"RTS is a pure orchestrator between the product and Roslyn
and knows nothing about C#"*, the member-surface enumeration and the C# spelling belong to
the product; only the body-policy overlay (which member is `Full` / `Stub` / `Skeleton`) is
legitimately RTS-owned.

## Current architecture (traced)

- **RTS does NOT spell C#.** `ComposeCompilationUnit` calls
  `new CSharpTypePrinter().PrintBatch(...)` (`ReturnToSenderTypePlanner.cs:1804`). The product
  `ILInspector.CSharp` renders `{ get; set; }`, `{ add; remove; }`, constructors, and
  explicit-interface syntax (`CSharpTypePrinter.cs`, `CSharpDeclarationWriter.cs`).
- **RTS DOES hand-roll the metadata→member enumeration.** `AddClosureMemberSurface` +
  `TypeProducer.*Requirement` build `CompileBackMemberRequirement`s (member discovery,
  accessor folding, backing-field skips, field-like / explicit-interface handling), then
  `ToApiMember` / `ToMemberPolicy` (`:1818`, `:1962`) translate them into the product's
  `ApiMember` + `CSharpBodyPolicy`.
- **The product already models this surface.**
  `ApiSurfaceExtractor.Extract(pe, includeAll: true)` returns type→`ApiMember`s with accessor
  folding, backing-field skip, explicit-interface handling, and adder/remover tokens — i.e.
  the same surface RTS re-derives.

## Proposed migration

1. RTS obtains the type member surface from `ApiSurfaceExtractor.Extract(pe, includeAll: true)`
[...]

PR #3098, opened 2026-07-23
Fold field-like event backing field into event in ApiSurfaceExtractor (#3083)

### LINKED ISSUE #3083: ApiSurfaceExtractor double-emits a field-like event's backing field under includeAll (RTS Full member-surface false success)
## Summary

Field-like events (`event Action Changed;`) produce a compiler-generated backing field whose
name is **identical** to the event (unmangled, not `<...>`-prefixed). The product's
`ILInspector.Metadata.ApiSurfaceExtractor.Extract(pe, includeAll: true)` emits **both** the
`event` member and this same-named backing `field`, because its field loop skips only
`<`-prefixed backing fields (`ApiSurfaceExtractor.cs:316`) with no cross-check against
field-like events. Any consumer of the full surface then double-declares the member
(`CS0102` / `CS0229`).

Originally reported as an RTS-only bug; investigation showed the root cause is in the
product's `ApiSurfaceExtractor`, so the fix belongs there. This affects the shipping
`--all` API surface as well as RTS compile-back.

## Empirical evidence

Temporary fact over the test assembly's own surface, with a fixture
`public class SampleFieldLikeEventHost { public event Action? Changed; }`:

```text
ApiSurfaceExtractor.Extract(pe, includeAll: true) → Changed members = [field:Changed, event:Changed]
```

Both a `field:Changed` and an `event:Changed` member are produced for the same name.

## Root cause

`ApiSurfaceExtractor.Extract` field loop (`src/ILInspector.Metadata/ApiSurfaceExtractor.cs:308-317`)
skips only fields whose name starts with `<` (auto-property / closure backing fields). A
field-like event's backing field is unmangled and shares the event's exact name, so it is not
skipped and is emitted as a separate field alongside the event. The backing field is private,
so the double-emit surfaces only under `includeAll` (e.g. `--all`).

## Fix direction (product)

In the field loop, skip a field whose name matches an event on the same type whose accessors
are compiler-generated field-like accessors (i.e. the field is that event's backing field).
Mirrors the existing `<`-prefixed backing-field skip, but keyed on the field-like event
pattern (a same-name field + event, which `CS0102` makes a unique identifier).

## RTS manifestation

The original RTS double-emit (`CompileBackSourceComposer.AddClosureMemberSurface` in
`tools/DecompilerHarness/ReturnToSenderTypePlanner.cs`) is a second, hand-rolled copy of this
same enumeration. It is subsumed by the RTS→product surface migration in #3091 (RTS inherits
the fix once it consumes `ApiSurfaceExtractor`). PR #3075 already excludes field-like events
from its Full event-target broadening, so #3075 remains neutral on this.

Found during adversarial review of PR #3075 (issue #3007).

richlander:
## Summary

Fixes #3083. A **field-like event** (`event Action Changed;`) has a compiler-generated backing
field whose name is **identical** to the event (unmangled, not `<...>`-prefixed). Under
`includeAll`, `ApiSurfaceExtractor.Extract` emitted **both** the `event` member and this
same-named `field`, producing an uncompilable double declaration (`CS0102`/`CS0229`) for any
consumer of the full surface (shipping `--all` output, RTS compile-back).

The extractor now skips a field whose name matches an event on the same type, folding the
field-like backing field into its event.

## Change

`src/ILInspector.Metadata/ApiSurfaceExtractor.cs`: before the field loop, collect the type's
event names; skip a field whose name is in that set. Mirrors the existing `<`-prefixed
backing-field skip, keyed on the field-like event pattern.

## Soundness

C# forbids a same-named field and event on one type (`CS0102`), so a field whose name equals
an event name is **uniquely** that event's compiler-generated backing field — the only field
this skip can ever remove. Custom events with distinctly-named user fields (e.g. `_changed`)
are unaffected. Behavior is observable only under `includeAll`; the default public surface
already omitted the private backing field via the accessibility check, so it is unchanged.

## Evidence

- **Before (proven):** a temporary probe over the test assembly's own surface returned
  `Changed members = [field:Changed, event:Changed]` for `public event Action? Changed;`.
- **After:** new positive regression test `Extract_FoldsFieldLikeEventBackingFieldIntoEvent`
  asserts a single `Changed` member of kind `event` and no `field` named `Changed`.
- **Negative control:** `Extract_KeepsCustomEventDistinctlyNamedBackingField` asserts a custom
  event's distinctly-named backing field still surfaces.
- Full `ApiSurfaceExtractorTests` (42) green; `ILInspector.Metadata.Tests` 750/751 green (the
  one failure is the pre-existing `MemberSourceProducer_UsesFirstVisibleDocumentWhenRootIsOmitted`
  backslash-path environmental red, CI-green, unrelated).

## Follow-up

Unblocks #3091 (migrate RTS to consume `ApiSurfaceExtractor` instead of its hand-rolled
duplicate enumeration); RTS inherits this fix once it consumes the product surface.

### DIFF
--- a/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
+++ b/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
@@ -306,13 +306,15 @@ public static class ApiSurfaceExtractor
             // Fields (non-backing fields; non-public included with --all)
             bool isEnum = apiType.Kind == "enum";
 
-            // A field-like event's compiler-generated backing field shares the event's exact
-            // (unmangled) name and is private. Fold only such backing fields, and only for
-            // events whose accessors are compiler-generated (i.e. genuinely field-like) — never
-            // based on a custom/explicit event. This keeps the extractor sound for hand-authored
-            // or non-C# metadata, where a legitimate field and a same-named event can coexist
-            // (the C# CS0102 restriction does not bind arbitrary IL): only a private field whose
-            // name matches a field-like event on this type is suppressed.
+            // A C# field-like event's compiler-generated backing field is private, is itself
+            // marked [CompilerGenerated], and shares the event's exact (unmangled) name. Fold a
+            // field only when it carries all of that positive backing-field evidence AND matches
+            // an event whose adder is [CompilerGenerated] (i.e. a genuinely field-like event).
+            // The decisive signal is the candidate field's own [CompilerGenerated] marker, not the
+            // accessor's: hand-authored or non-C# metadata may attach [CompilerGenerated] to a
+            // custom accessor while a legitimate, same-named field backs nothing of the sort (the
+            // C# CS0102 restriction does not bind arbitrary IL). Requiring the field itself to be
+            // private and compiler-generated keeps a genuine field from being suppressed.
             HashSet<string>? fieldLikeEventBackingFieldNames = null;
             foreach (var eventHandle in typeDef.GetEvents())
             {
@@ -343,8 +345,12 @@ public static class ApiSurfaceExtractor
                     continue; // Skip backing fields
 
                 if (fieldAccess == FieldAttributes.Private
-                    && fieldLikeEventBackingFieldNames?.Contains(fieldName) == true)
-                    continue; // Skip a field-like event's private backing field (shares the event name)
+                    && fieldLikeEventBackingFieldNames?.Contains(fieldName) == true
+                    && AttributeReader.HasAttribute(
+                        reader,
+                        field.GetCustomAttributes(),
+                        KnownAttributeNames.CompilerGeneratedAttribute))
+                    continue; // Skip a field-like event's private, compiler-generated backing field
 
                 // Skip EditorBrowsable(Never) fields unless --all; obsolete are surfaced with marker.
                 if (!includeAll && AttributeReader.HasEditorBrowsableNeverAttribute(reader, field.GetCustomAttributes()))
--- a/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
+++ b/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
@@ -305,6 +305,19 @@ public static class ApiSurfaceExtractor
 
             // Fields (non-backing fields; non-public included with --all)
             bool isEnum = apiType.Kind == "enum";
+
+            // A field-like event's compiler-generated backing field shares the event's exact
+            // (unmangled) name. C# forbids a same-named field and event on one type (CS0102),
+            // so a field whose name matches an event on this type is uniquely that event's
+            // backing field and must not be surfaced as a separate field (it would produce a
+            // duplicate declaration alongside the event member).
+            HashSet<string>? fieldLikeEventBackingFieldNames = null;
+            foreach (var eventHandle in typeDef.GetEvents())
+            {
+                (fieldLikeEventBackingFieldNames ??= new HashSet<string>(StringComparer.Ordinal))
+                    .Add(reader.GetString(reader.GetEventDefinition(eventHandle).Name));
+            }
+
             foreach (var fieldHandle in typeDef.GetFields())
             {
                 var field = reader.GetFieldDefinition(fieldHandle);
@@ -315,6 +328,9 @@ public static class ApiSurfaceExtractor
                 string fieldName = reader.GetString(field.Name);
                 if (fieldName.StartsWith("<"))
                     continue; // Skip backing fields
+
+                if (fieldLikeEventBackingFieldNames?.Contains(fieldName) == true)
+                    continue; // Skip a field-like event's backing field (shares the event name)
 
                 // Skip EditorBrowsable(Never) fields unless --all; obsolete are surfaced with marker.
                 if (!includeAll && AttributeReader.HasEditorBrowsableNeverAttribute(reader, field.GetCustomAttributes()))
--- a/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
+++ b/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
@@ -307,15 +307,28 @@ public static class ApiSurfaceExtractor
             bool isEnum = apiType.Kind == "enum";
 
             // A field-like event's compiler-generated backing field shares the event's exact
-            // (unmangled) name. C# forbids a same-named field and event on one type (CS0102),
-            // so a field whose name matches an event on this type is uniquely that event's
-            // backing field and must not be surfaced as a separate field (it would produce a
-            // duplicate declaration alongside the event member).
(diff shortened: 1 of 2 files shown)


richlander:
## Round-1 review reconciliation (head `1d31b469`)

Two-model adversarial review of the name-only head (`f7c9bef5`) returned opposite conclusions; both are addressed by the hardening in `1d31b469`.

### GPT (gpt-5.6-sol) — BLOCKING (reproduced, valid)

**Finding:** the name-only fold suppressed *any* field whose name matched an event on the same type, resolving ambiguity via the C# CS0102 rule (no same-named field + event). But `dotnet-inspect` inspects **arbitrary** assemblies, and hand-authored / non-C# IL can legally carry a public field alongside a private, same-named custom or explicit event. GPT constructed valid IL (public `int32 Clash` + private explicit event `Clash` backed by `_eventBacking`) where the fold dropped the legitimate public field. CS0102 does not bind raw IL, so the uniqueness assumption was unsound.

**Fix (verified):** require **positive backing-field evidence** instead of name uniqueness — fold only a **private** field whose name matches an event whose **adder is `[CompilerGenerated]`** (a genuine C# field-like event). A custom/explicit event's accessors are not compiler-generated, so its same-named field is never suppressed.

Regression added: `Extract_KeepsPublicFieldMaskedBySameNamedNonFieldLikeEvent` emits exactly GPT's clash shape via `PersistedAssemblyBuilder` (public field `Clash` + private, non-`[CompilerGenerated]` event `Clash`) and asserts **both** members survive.

### Gemini (gemini-3.1-pro-preview) — CLEAN

Verified C#/VB/F# all forbid same-name field+event (CS0102/BC30260/FS0023), per-type scoping is correct, inheritance/shadowing is safe, VB/F# field-like backing fields are differently named (correctly left as fields), and there is no NRE/perf regression. Gemini's model was the compiler-produced case, which the hardened fix still handles; GPT's arbitrary-IL attack point was outside Gemini's six probes.

### Evidence

- `ApiSurfaceExtractorTests`: **43/43 pass, 0 skipped** (clash test now runs via `PersistedAssemblyBuilder`, no ilasm dependency).
- Full `dotnet-inspect.Tests`: **1988 total, 14 failed** — the 14 are **pre-existing environmental reds** (Windows CRLF in `PackageVersionTests.Versions_WithRange...`; network/source-link/perf-triage `CommandExecutionTests`). Verified by stashing the change and re-running: the 14-name FAIL set is **byte-identical** with and without this PR. All CI-green.

Requesting fixed-head (`1d31b469`) re-review from both models before marking ready.

richlander:
## Round-2 review reconciliation (head `3dac3382`)

Round-2 fixed-head (`1d31b469`) re-review again split; GPT's finding is valid and is fixed in `3dac3382`.

### GPT (gpt-5.6-sol) — BLOCKING (reproduced, valid)

**Finding:** `[CompilerGenerated]` on an *adder* is not proof that a same-named private field backs the event. Legal IL can stamp `[CompilerGenerated]` on a **custom** accessor that updates a distinctly-named field (`_eventBacking`), while a legitimate private field shares the event's name. GPT assembled exactly that (`private int32 Clash` + CG `add_Clash` backing `_eventBacking`) and showed `Extract(..., includeAll:true)` still dropped the legitimate `Clash` field. Keying the fold on the accessor attribute alone is the wrong signal.

**Fix (verified):** move the decisive evidence to the **candidate field itself**. Fold only when the field is **`Private`** *and* itself carries **`[CompilerGenerated]`** *and* matches an event whose adder is `[CompilerGenerated]`. csc always stamps the real field-like backing field with `[CompilerGenerated]` (verified by reading the metadata of a real `event Action? Changed;` — field is `Private`, `[CompilerGenerated]=True`, type `Action`). A plain user field never carries `[CompilerGenerated]`, so it survives regardless of the surrounding event/accessor shape.

Deliberately **not** adding a field/event type-signature equality gate: it would risk regressing common genuine cases (`event EventHandler? X;`, generic/dynamic handlers) via cross-decoder string mismatch (nullability/dynamic decoration), while the field's own `[CompilerGenerated]` marker is the authoritative, non-fragile signal. If anyone can exhibit a *legitimate* field lost by the field-`[CompilerGenerated]` rule, I'll add a robust raw-type match — but the CG-marked-own-field case is self-declared compiler storage and not C#-expressible alongside the event.

### Gemini (gemini-3.1-pro-preview) — CLEAN

Round-2 Gemini verified (with real csc output + raw metadata) that the adder `[CompilerGenerated]` isolation, the `Private` check, and NRE/perf safety all hold. Its probes did not include GPT's arbitrary-IL "CG adder backs another field" shape; the `3dac3382` field-level guard subsumes Gemini's cases (genuine csc field is both private and `[CompilerGenerated]`).

### Evidence

- New regression `Extract_KeepsFieldWhenCompilerGeneratedAdderBacksAnotherField` reproduces GPT's exact attack (private non-CG `Clash` + CG custom adder). **Verified it fails against the prior adder-only logic and passes at `3dac3382`.**
- Genuine csc field-like fold still fires (`Extract_FoldsFieldLikeEventBackingFieldIntoEvent`) — no over-narrowing.
- `ApiSurfaceExtractorTests`: **44/44 pass, 0 skipped**.

Requesting round-3 fixed-head (`3dac3382`) re-review from both models.

richlander:
## Round-3 review reconciliation (head `3dac3382`) — both reviews CLEAN

Both fixed-head (`3dac3382`) re-reviews signed off with no findings.

- **GPT (gpt-5.6-sol): CLEAN** — build succeeded, all 44 extractor tests pass; no correctness, evidence-integrity, NRE, handle-resolution, or performance issues. GPT's round-2 blocking finding (accessor `[CompilerGenerated]` is not backing-field proof) is closed by the field-level marker requirement.
- **Gemini (gemini-3.1-pro-preview): CLEAN** — independently compiled a scratch project with both `event Action MyEvent;` and a generic-nullable `event EventHandler<string>? MyGenericEvent;`, read the metadata, and confirmed csc stamps `[CompilerGenerated]` on the **backing field itself** in both cases, so the fold stays intact (no over-narrowing). Verified short-circuit perf/no-NRE and that `Extract_KeepsFieldWhenCompilerGeneratedAdderBacksAnotherField` genuinely guards the fix.

### Discriminator, final form

A field is folded as a field-like event's backing field only when **all** hold: it is `Private`, it itself carries `[CompilerGenerated]`, and its name matches an event whose adder is `[CompilerGenerated]`. The candidate field's own `[CompilerGenerated]` is the decisive, non-fragile signal — sound for arbitrary/non-C# IL, and (per Gemini) intact for nullable/generic C# handlers. No type-signature gate, by design (avoids nullability/dynamic decoder-mismatch regressions).

Holding on final green `test` lane; will confirm readiness once CI completes.

richlander:
Ready to merge

PR #3078, opened 2026-07-23
Reconcile enum-shift operands feeding a bitwise op (#3011)

richlander:
- Fixes/advances #3011 (follow-up landing after #3060 merged)
- Changes: an enum shift the printer rewrites to its underlying integer no longer misleads two downstream consumers — a sibling bitwise operand is no longer coerced to the enum (`int | (E)x`, CS0019), and a shift flowing into an enum target keeps its outer `(Enum)` cast (`(int)e >> n`, CS0266). Both now render valid and faithful.
- Card revision: `eda3d14f`

## Change

> Should we accept this change?

**Conclusion:** **PASS** — the merged #3011 fix left the rewritten enum shift's IR `Binary` node reporting a stale enum `ResultType`, which misleads two consumers. (1) A parent `&`/`|`/`^` coerced the *sibling* integer to the enum (`int | (E)x`, CS0019). (2) A shift returned/stored into an enum target read as an identity and dropped the required outer cast (`(int)e >> n` where an enum is expected, CS0266). This PR reconciles both printer-side without changing any previously-valid output.

### Original source

Real-world shape: Roslyn `Microsoft.Cci.MetadataWriter.GetRawToken` returns `((uint)encoding << 24) | pseudoToken`, where `encoding` is an enum. The compiled fixture `EnumCastSamples.IntEnumShiftOrUnsigned` reproduces that IL exactly (verbatim fixture source):

```csharp
public static uint IntEnumShiftOrUnsigned(CfgPriority e, uint x) => ((uint)e << 24) | x;
```

`(uint)e` is an identity reinterpret over the int-backed enum, so csc emits a bare enum-typed shift left operand — the same trace-free cast that #3011 targets, here feeding a parent `|`.

### Before

Base commit `d6af0165` (main, with #3060 already merged), same fixture IL, `-S "Decompiled Source"`:

```csharp
public static uint IntEnumShiftOrUnsigned(CfgPriority e, uint x) => (uint)(((int)e << 24) | (CfgPriority)x);
```

- Valid: False
- Correct: False

The shift's stale enum `ResultType` misleads the bitwise enum-coercion, which casts the sibling `x` to `CfgPriority`. `int | CfgPriority` has no predefined operator → CS0019 (confirmed by Roslyn: `error CS0019: Operator '|' cannot be applied to operands of type 'int' and 'CfgPriority'`). It does not compile, so correctness is moot.

### After

This PR's head `3bb28958`, same fixture IL, `-S "Decompiled Source"`:

```csharp
[...]

PR #3100, opened 2026-07-23
Route enum-shift sibling/arm consumers through rendered integer type (#3087)

richlander:
- Fixes/advances #3087
- Changes the C# printer so enum-shift consumers (bitwise sibling, comparison, conditional/coalesce arm) reason about the shift's **rendered integer type** instead of its stale enum `ResultType`, eliminating three invalid-C# shapes.
- Card revision: `52a053c9`

## Change

**Conclusion:** **PASS** — three pre-existing consumers of an enum shift's stale enum `ResultType` emitted uncompilable C# (CS0019/CS0019/CS1503); routing them through the existing rendered-integer predicates makes all three valid while preserving IL fidelity (every inserted cast is an identity reinterpret emitting zero IL).

### Background

An enum shift's IR `Binary` reports the enum as its `ResultType` (`TypeFamilies.BinaryResult` prefers the enum definition for the flags idiom), but the printer always renders the left operand as its underlying integer via `ShiftEnumLeftOperand` (`(int)e << n`). PR #3078/#3076 fixed the witnessed consumers; adversarial review surfaced three more latent consumers that still trust the stale enum type. This PR is the tracking follow-up (#3087).

### Defect 1 — bitwise op with an enum-typed sibling

**Original source**

```csharp
public static CfgPriority EnumShiftOrEnumSibling(CfgPriority e, int n, CfgPriority other) => (CfgPriority)(((int)e << n) | (int)other);
```

**Before**

```csharp
public static CfgPriority EnumShiftOrEnumSibling(CfgPriority e, int n, CfgPriority other) => (int)e << n | other;
```

- Valid: False — CS0019: operator `|` cannot be applied to `int` and `CfgPriority`.
- Correct: False (does not compile).

**After**

```csharp
public static CfgPriority EnumShiftOrEnumSibling(CfgPriority e, int n, CfgPriority other) => (CfgPriority)(((int)e << n) | (int)other);
```

- Valid: True
- Correct: True

### Defect 2 — comparison against an enum constant

**Original source**

```csharp
public static bool EnumShiftCompareEnumConst(CfgPriority e, int n) => ((int)e << n) == (int)CfgPriority.High;
```

**Before**

```csharp
public static bool EnumShiftCompareEnumConst(CfgPriority e, int n) => ((int)e << n) == CfgPriority.High;
```

- Valid: False — CS0019: operator `==` cannot be applied to `int` and `CfgPriority`.
- Correct: False (does not compile).

**After**

```csharp
public static bool EnumShiftCompareEnumConst(CfgPriority e, int n) => ((int)e << n) == (int)CfgPriority.High;
```

- Valid: True
- Correct: True

### Defect 3 — conditional arm into an enum target

**Original source**

```csharp
public static void EnumShiftTernaryToEnum(bool b, CfgPriority e, int n) => TakeCfgPriority(b ? (CfgPriority)((int)e << n) : CfgPriority.High);
```

**Before**

```csharp
public static void EnumShiftTernaryToEnum(bool b, CfgPriority e, int n) => TakeCfgPriority(b ? ((int)e << n) : CfgPriority.High);
```

- Valid: False — CS1503: cannot convert the target-typed conditional (best common type of `int` and `CfgPriority`) to `CfgPriority`.
- Correct: False (does not compile).

**After**

```csharp
public static void EnumShiftTernaryToEnum(bool b, CfgPriority e, int n) => TakeCfgPriority(b ? (CfgPriority)((int)e << n) : CfgPriority.High);
```

- Valid: True
- Correct: True

### Fully raised

## Design

Five coordinated edits in `CSharpPrinter.Numerics.cs`, all consulting the existing rendered-integer predicate cluster (`BitwiseOperandRendersAsInteger` / `BitwiseOperandRenderedType`) rather than `Binary.ResultType`:

1. **`BinaryBody` bitwise:** before the existing integer-sibling up-coercion, add a DOWN-coercion — when one operand renders as int and its sibling is a bare enum, coerce the sibling down to the shift's rendered integer (`(int)other`). (Defect 1)
2. **New `TryCoerceEnumToInteger(value, integerType)`** helper — `CoerceText(value, integerType)` only when `value` is a bare enum not already rendering as int; else null. Single enum→integer door, matching `ShiftEnumLeftOperand`'s spelling.
3. **`CompareText`:** add the same DOWN-coercion, and guard the pre-existing up-coercion with `!BitwiseOperandRendersAsInteger` so a shift's stale enum never pulls the sibling up. (Defect 2)
4. **`TryCoerceEnumOperand`:** add a branch — a value rendering as int flowing into an enum side takes `EnumIntegerCast`. Central door for conditional/switch/coalesce arms. (Defect 3)
5. **`BitwiseChainRenderedType`** (via new `ChainOperandRenderedInteger`): map a bare-enum chain operand to its underlying integer, so the outer `(E)` return cast on defect 1's `(E)((int)e<<n | (int)other)` is retained.

DOWN-coercion runs first and returns early in `BinaryBody`/`CompareText`, so the extended `TryCoerceEnumOperand` fires only in arm contexts (no double-handling).

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `EnumCastPrinterTests`: 53 passed | `EnumCastPrinterTests`: 62 passed (9 new) |
| Decompiler fast suite (`-trait- "Speed=Slow"`) | 3413, 0 failed | 3422, 0 failed |
| Reduced fixture validity | Pass (3 defect fixtures compile via `AssertCompiles`) | Pass |
| Reduced fixture fidelity | IL unchanged — inserted casts are identity reinterprets | IL unchanged |
| Real witness | 3 fixtures render invalid C# (CS0019/CS0019/CS1503) | 3 fixtures render valid C# |

IL fidelity: `member -S "IL"` on `EnumShiftOrEnumSibling` shows `ldarg.0; ldarg.1; ldc.i4.s 31; and; shl; ldarg.2; or; ret` — no `conv` opcodes. Every cast this fix inserts (`(int)e`, `(int)other`, `(CfgPriority)`) is a no-op reinterpret emitting zero IL, so recompiled opcodes are unchanged. The slow Fidelity-area gate is environmentally flaky on the dev machine and is deferred to CI.

Before/After renders were acquired verbatim via `dotnet-inspect member ... -S "Decompiled Source"` at base and head; the three Before shapes were independently confirmed to fail compilation (CS0019, CS0019, CS1503) and the three After shapes to compile.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class "ILInspector.Decompiler.Tests.EnumCastPrinterTests"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -5213,6 +5213,41 @@ public static class EnumCastSamples
 
     public static void EnumShiftStoreArray(CfgPriority[] arr, int n) => arr[0] = (CfgPriority)((int)arr[0] >> n);
 
+    // #3087 defect 1: a bitwise op whose sibling is itself an ENUM value. The shift
+    // renders as its underlying integer, so the enum sibling must be coerced to that
+    // integer — `(int)e << n | (int)other`, not `(int)e << n | other` (int | E,
+    // CS0019). The stale enum ResultType makes the sibling render bare.
+    public static CfgPriority EnumShiftOrEnumSibling(CfgPriority e, int n, CfgPriority other) => (CfgPriority)(((int)e << n) | (int)other);
+
+    // #3087 defect 2: an enum shift COMPARED to an enum-constant sibling. The shift
+    // renders as int, so the comparison must be integer-vs-integer — the stale enum
+    // ResultType instead coerces the sibling to the enum (`(int)e << n == E.High`,
+    // int == E, CS0019).
+    public static bool EnumShiftCompareEnumConst(CfgPriority e, int n) => ((int)e << n) == (int)CfgPriority.High;
+
+    // #3087 defect 3: an enum shift as a ternary arm flowing into an ENUM target.
+    // The shift's ResultType matches the target enum, so the `(CfgPriority)` cast is
+    // dropped and the arm renders `(int)e << n` (int) while the other arm is the
+    // enum — no common type (CS1503/CS0173). The cast must be kept. A method-argument
+    // ternary stays a genuine conditional (it cannot be raised to if/return).
+    public static void TakeCfgPriority(CfgPriority value) { }
+
+    public static void EnumShiftTernaryToEnum(bool b, CfgPriority e, int n) => TakeCfgPriority(b ? (CfgPriority)((int)e << n) : CfgPriority.High);
+
+    // #3087 negatives/edges. Mixed-sign: an UNSIGNED enum shift (`shr.un`) rendering
+    // as uint whose enum sibling must be coerced to that width/sign — `(uint)other`,
+    // not `(int)other` (which would be a mixed-sign bitwise op, CS0019).
+    public static uint EnumShiftUnsignedOrEnumSibling(CfgPriority e, int n, CfgPriority other) => ((uint)e >> n) | (uint)other;
+
+    // A comparison whose enum sibling is a RUNTIME value (not a constant): coerced
+    // down to the shift's rendered integer — `(int)e << n == (int)other`.
+    public static bool EnumShiftCompareEnumValue(CfgPriority e, int n, CfgPriority other) => ((int)e << n) == (int)other;
+
+    // A bitwise CHAIN over an enum shift with an enum FAR sibling: the down-coercion
+    // recurses so the enum sibling at the end of `shift | y | other` renders as its
+    // underlying integer — `(int)e << 4 | y | (int)other`.
+    public static int EnumShiftChainEnumSibling(CfgPriority e, int y, CfgPriority other) => ((int)e << 4) | y | (int)other;
+
     // #1766: ternary with enum-constant arms stored to a cross-assembly enum
     // local (StringComparison.Ordinal = 4, OrdinalIgnoreCase = 5).
     public static bool EnumConditional(string name, bool ci)
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -5295,6 +5295,28 @@ public static class EnumCastSamples
     // (a plain `(int)CfgFlags.Top` is CS0221).
     public static bool EnumShiftCompareOverflowConst(CfgFlags e, int n) => ((int)e << n) == unchecked((int)CfgFlags.Top);
 
+    // #3087 follow-up (adversarial review R4, GPT + Gemini): EQUALITY against a
+    // plain `uint` sibling. Equality is bit-exact, so it reconciles to the shift's
+    // SIGNED stack width — `((int)e << n) == (int)x`. The stale-ResultType
+    // fallthrough dropped the sibling cast (`== x`), and `int == uint` widens to a
+    // signed 64-bit `ceq` in C# — with `e = (CfgPriority)(-1)`, `x = 0xFFFFFFFF` the
+    // IL 32-bit `ceq` is TRUE but `-1L == 4294967295L` is FALSE.
+    public static bool EnumShiftEqualsUintSibling(CfgPriority e, int n, uint x) => (uint)((int)e << n) == x;
+
+    // #3087 follow-up (adversarial review R4, GPT + Gemini): SIGNED ordering with a
+    // plain integer sibling reconciles to the shift's signed stack width —
+    // `((int)e << n) < other` (identity when the sibling already is `int`), never a
+    // sign-widened `int < uint` promotion.
+    public static bool EnumShiftSignedLessIntSibling(CfgPriority e, int n, int other) => ((int)e << n) < other;
+
+    // #3087 follow-up (adversarial review R4, GPT): a SIGNED ordering after an
+    // UNSIGNED shift (`shr.un` renders `uint`). The target is driven by the
+    // comparison (signed), not the shift's rendered sign, so the shift is
+    // reinterpreted back to `int` — `(int)((uint)e >> n) < (int)other`. Coercing to
+    // the shift's `uint` instead would emit an unsigned compare `((uint)e >> n) <
+    // (uint)other`, silently changing the ordering.
+    public static bool EnumShiftUnsignedShrSignedCompare(CfgPriority e, int n, CfgPriority other) => (int)((uint)e >> n) < (int)other;
+
     // #1766: ternary with enum-constant arms stored to a cross-assembly enum
     // local (StringComparison.Ordinal = 4, OrdinalIgnoreCase = 5).
     public static bool EnumConditional(string name, bool ci)
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -5256,11 +5256,32 @@ public static class EnumCastSamples
     // `(uint)((int)e << n) < (uint)other`, matching `clt.un` for any backing.
     public static bool EnumShiftUnsignedCompare(CfgFlags e, int n, CfgFlags other) => (uint)((uint)e << n) < (uint)other;
 
-    // #3087 follow-up: the same unsigned ordering with a BYTE-backed enum. Round-
+    // #3087 follow-up: the same unsigned ordering with a byte-backed enum. Round-
     // tripping through the byte enum (`(CfgTiny)((int)e << n)`) would re-narrow the
     // 32-bit shift to 8 bits and then compare signed — the unsigned-width spelling
     // `(uint)((int)e << n) < (uint)other` avoids both.
     public static bool EnumShiftUnsignedCompareByte(CfgTiny e, int n, CfgTiny other) => (uint)((byte)e << n) < (uint)other;
+
+    // #3087 follow-up (adversarial review R3, GPT): the unsigned ordering with a
+    // PLAIN INTEGER sibling (not an enum). The shift still needs the unsigned
+    // reinterpret so the compare is `clt.un` — `(uint)((int)e << n) < (uint)other` —
+    // not the sign-widened `((int)e << n) < (uint)other` (`int < uint` promotes to a
+    // signed `long` compare) the stale-ResultType fallthrough produced.
+    public static bool EnumShiftUnsignedCompareIntSibling(CfgPriority e, int n, int other) => (uint)((int)e << n) < (uint)other;
+
+    // #3087 follow-up (adversarial review R3, GPT): the 8-byte case with a plain
+    // `long` sibling. The fallthrough rendered `((long)e << n) < (ulong)other`
+    // (`long < ulong`, CS0034 — did not compile); both sides must reconcile to
+    // `ulong` — `(ulong)((long)e << n) < (ulong)other`.
+    public static bool EnumShiftUnsignedCompareLongSibling(CfgLongPriority e, int n, long other) => (ulong)((long)e << n) < (ulong)other;
+
+    // #3087 follow-up (adversarial review R3, Gemini): a bitwise chain mixing an
+    // enum shift with a NARROW (ushort) integer, compared unsigned. IL promotes the
+    // short to Int32, so the chain's stack type is wide; treating it as narrow made
+    // BitwiseChainRenderedType fall back to the stale enum ResultType and drop the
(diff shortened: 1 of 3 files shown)


richlander:
## Adversarial review round 1

Two-model adversarial review at head `bf090f76` (isolated review worktrees, scratch under `/tmp/`).

### Gemini (`gemini-3.1-pro`) — CLEAN

Exercised all seven attack points (over-firing, width/sign, chain mapping, up-coercion guards, preemption ordering, masked siblings, fidelity) with real builds/renders. No defects.

### GPT (`gpt-5.6-sol`) — CHANGES REQUESTED (3 findings, all reproduced and fixed in `52a053c9`)

1. **Unsigned ordering silently rendered signed (High).** An unsigned ordering (`clt.un`/`cgt.un`) of an enum shift vs an enum sibling was down-coerced to the shift's *signed* rendered integer, so `<` compared signed — flipping high-bit values. Confirmed: base rendered a loud `CS0019`; head-before-fix rendered a valid-but-wrong `((int)e << n) < (int)other`.
   **Fix:** restrict the comparison down-coercion to sign-safe kinds (equality is bit-exact regardless of sign). For unsigned ordering the shift falls through to enum up-coercion, so C#'s enum `<` on the uint-backed enum compares the underlying uint — `(CfgFlags)((int)e << n) < other`. Verified behaviorally correct against an exhaustive unsigned oracle over all `{Zero, High} × n∈[0,2]` inputs.

2. **Masked ldelem/ldind enum sibling left bare (Medium, pre-existing).** A bitwise sibling masked as its primitive by a typed `ldelem`/`ldind` renders enum-typed (`values[i]`) but was not coerced → `int | E` (CS0019). Identical at base and head (not a regression), but the same enum-shift-sibling feature owns it.
   **Fix:** recover the rendered enum via `ShiftLeftEnumType` and spell the reinterpret directly for the masked case (`CoerceText` sees an int→int identity and drops it).

3. **Overflowing enum constant dropped `unchecked` (Medium, pre-existing).** `(int)CfgFlags.Top` (0x80000000, uint-backed) is a `CS0221` constant overflow. Identical shape invalid at base and head.
   **Fix:** wrap the down-coercion in `unchecked` when the constant's underlying type can overflow the signed target.

All three inserted casts are identity reinterprets (no `conv` in IL), so fidelity holds. Added three regression fixtures + tests.

**Evidence at `52a053c9`:** `EnumCastPrinterTests` 62/0 (was 59), fast suite 3422/0, Validity 103/0. All four GPT probe shapes now compile and are behaviorally correct.

Re-reviewing the fixed head with both models next.

richlander:
## Round-2 review reconciliation (head `ab4cb9ad`)

Both round-2 reviewers independently found the **same** gap in the round-1
`comparisonSignSafe` guard, and both are correct:

- **GPT (round-2, CHANGES REQUESTED):** the unsigned-ordering fall-through to
  enum up-coercion is only faithful for full-width `uint`/`ulong` backings;
  signed backings compare signed (`clt`), narrow (`byte`) backings truncate the
  shifted value via `conv.u1`.
- **Gemini (round-2, CHANGES REQUESTED):** same root cause — skipping the
  down-coercion let the enum up-coercion block below fire, so a `byte`-backed
  `(E)((int)e << n) < other` re-narrows the 32-bit shift to 8 bits and then
  compares signed, diverging from the IL's full-width `clt.un`.

Reproduced both with byte-backed (`U8`) and int-backed (`S32`) enum probes —
confirmed unsound (`ByteUnsignedLess(0x80, 24, 0xFF)` rendered `true` vs the
correct `false`).

**Fix (`ab4cb9ad`):** replaced the `comparisonSignSafe` skip with a new
`DownCoercedShiftComparison` helper. For an unsigned ordering it reconciles BOTH
operands to the unsigned counterpart of the shift's stack width —
`(uint)((int)e << n) < (uint)other` (`(ulong)` for 8-byte) — so the comparison
is the unsigned one the IL performs, independent of the enum backing. Equality
and signed ordering keep the same-width down-coercion. The enum up-coercion
block is now dead for this shape (the helper short-circuits first).

**Evidence:**
- Renders now `(uint)((int)e << n) < (uint)other` for byte/signed/uint,
  `(ulong)((long)e << n) < (ulong)other` for 8-byte; chain-over-shift stays
  correctly parenthesized `(uint)(((int)e << n) | (int)mask) < (uint)other`.
- Runtime `clt.un` oracle over byte/signed inputs: **ALL OK** (0 mismatches).
- `EnumCastPrinterTests` 63/0 (added `EnumShiftUnsignedOrdering_ByteBacked_…`),
  fast suite 3423/0, Validity 103/0.

Re-reviewing this fixed head with both models before marking ready.

richlander:
## Round-3 review reconciliation (head `8cf74921`)

Both round-3 reviewers exercised the new `DownCoercedShiftComparison` with
compiled probes + runtime oracles. Bare-enum siblings, 8-byte backings, symmetry,
equality bit-exactness, and non-target shapes were confirmed correct. Two
remaining unsigned-ordering (`clt.un`) holes were found and fixed:

- **GPT (High) + Gemini Issue 2 (High), plain-integer sibling:** the helper
  bailed when the sibling was not an enum, so the fallthrough `UnsignedOperand`
  (keyed off the shift's stale enum `ResultType`) dropped the `(uint)` cast on
  the shift → `((int)e << n) < (uint)other` (`int < uint` → signed `long`
  compare; the 8-byte case `long < ulong` was CS0034). Reproduced both.
- **Gemini Issue 1 (High), narrow bitwise chain:** `ChainOperandRenderedInteger`
  reported a sub-int operand (`short`/`ushort`) at its narrow width, so
  `BitwiseChainRenderedType` rejected the chain and fell back to the enum
  ResultType, dropping the casts → `((int)e << n | mask) < other` (CS0019 /
  silent signed compare). This was pre-existing for the enum-sibling case, but my
  round-2 change had turned the plain-int narrow-chain case from a loud CS0019
  into a silent signed compare — so I fixed it at the root.

**Fix (`8cf74921`):**
- Any integer sibling now reconciles to the unsigned stack width via `CoerceText`
  — `(uint)((int)e << n) < (uint)other` (`(ulong)` for 8-byte).
- `ChainOperandRenderedInteger` promotes a sub-int operand to `Int32` (its IL
  stack width), so a narrow-operand chain reconciles wide instead of being
  rejected.

**Evidence:**
- Renders now correct/compiling for byte, signed, uint, ulong, plain-int, plain-
  long, uint-sibling, and narrow (ushort) bitwise-chain shapes; equality/signed
  ordering still not over-wrapped.
- Runtime `clt.un` oracles over plain-int/uint/long and narrow-chain inputs:
  **ALL OK** (0 mismatches).
- `EnumCastPrinterTests` 66/0, fast suite 3426/0, Validity 103/0.

Re-reviewing this fixed head with both models before marking ready.

richlander:
## Round-4 review reconciliation (head `686991c1`)

Both fixed-head reviewers at `8cf74921` requested changes and converged on the
same soundness hole in `DownCoercedShiftComparison`'s equality/signed branch:

- **GPT (`gpt-5.6-sol`)** — two loud→silent regressions:
  - **A.** A plain `uint`/`ulong` sibling in **equality** and **signed ordering**
    fell through (bailed when `TryCoerceEnumToInteger` returned null), dropping the
    sibling cast. `int == uint` / `int < uint` sign-widens to a 64-bit compare in
    C#, changing results (`(E)(-1)` vs `0xFFFFFFFF`: IL 32-bit `ceq` TRUE, but
    `-1L == 4294967295L` FALSE). Base emitted CS0019 → my PR made it silent-wrong.
  - **B.** A **signed** `clt`/`cgt` after `shr.un`: I coerced to the shift's
    rendered `uint`, emitting an *unsigned* compare `((uint)e >> n) < (uint)other`
    instead of the signed `(int)((uint)e >> n) < (int)other`.
- **Gemini (`gemini-3.1-pro-preview`)** — independently found issue **A** (attack
  point 7), with the same `(E)(-1)` / `0xFFFFFFFF` counterexample; attack points
  1–6 and 8 CLEAN (mixed-width/sign chains, `unchecked` constants, redundant
  outer-cast stripping, symmetry all confirmed).

Both reproduced on head and against base (CS0019 there) → confirmed
loud→silent regressions I introduced.

### Fix (`686991c1`)

Rewrote the helper to derive the reconcile **target from the comparison's
signedness**, not the shift's rendered signedness:

- unsigned ordering (`clt.un`/`cgt.un`) → unsigned stack width;
- signed ordering (`clt`/`cgt`) and equality (`ceq`/`bne.un`, bit-exact) → signed
  stack width.

Both operands now reconcile through a shared `ReconcileComparisonOperand`: an
enum via `TryCoerceEnumToInteger`; an integer-rendering shift/chain via a
same-width bit-preserving reinterpret (a `SamePrimitive` check that ignores the
stale enum `ResultType`); a plain integer via `CoerceText`. This closes the
whole family: unsigned/signed/equality × enum/plain-int/chain × any backing/width.

**Evidence:** runtime oracle 144 triples, 0 mismatches vs IL semantics (incl.
Gemini's exact counterexample). New fixtures + tests for uint-equality, plain-int
signed ordering, and `shr.un` signed ordering. 69 EnumCast / 3429 fast / 103
validity, all green. Rendered:

- `EnumShiftEqualsUintSibling` → `((int)e << n) == (int)x`
- `EnumShiftSignedLessIntSibling` → `((int)e << n) < other`
- `EnumShiftUnsignedShrSignedCompare` → `(int)((uint)e >> n) < (int)other`

Relaunching both fixed-head reviews at `686991c1`.

richlander:
## Round-5 review reconciliation (head `12bc344d`)

Both fixed-head reviewers at `686991c1` requested changes, each finding a
distinct new hole (no overlap this round):

- **GPT (`gpt-5.6-sol`) — checked context.** The reconcile cast the printer
  *inserts* on the shift side is not in the IL (`shl; clt.un`). Under an enclosing
  `checked` region a signed→unsigned reinterpret (`(uint)(-1)`) **throws
  `OverflowException`**, where the IL only reinterprets bits. Repro:
  `checked((unchecked((uint)((int)e << n)) < other ? 1 : 0) + y)` threw at runtime
  for `e = (E)(-1), n = 0`. Base emitted different (signed-widened, unsound)
  output — so this was loud-throw vs correct.
- **Gemini (`gemini-3.1-pro-preview`) — mixed-sign chain.** `BitwiseChainRenderedType`
  reported a chain mixing an int enum-shift with an **unsigned-backed enum sibling**
  (`U : uint`) as `uint`, but `BinaryBody` coerces that sibling DOWN to the shift's
  signed stack type (`(int)x`), so the chain actually renders as `int`. The stale
  `uint` report made an enclosing `clt.un`/`cgt.un` drop the outer `(uint)`
  reinterpret → silent `int > uint` 64-bit signed promotion (inverting high-bit
  results; CS0034 at 8-byte width). Repro `e = (U)0xC0000000`: IL `cgt.un` TRUE,
  dropped-cast output FALSE.

### Fix (`12bc344d`)

- **Checked-safe reconcile:** route the shift-side reinterpret through
  `CheckedSafeCast`, wrapping `unchecked` in a checked context — symmetric with the
  enum and plain-integer reconcile paths.
- **Chain sign:** `BitwiseChainRenderedType` now mirrors `BinaryBody` — when one
  side renders as an integer and the other is a down-coerced enum, the chain
  renders as the shift side's integer (contributing the shift's sign, not the
  sibling's backing). This also aligns nested chains (`(e<<n)|x|y`).

**Evidence:** checked oracle 108 triples / 0 mismatch / 0 throw; chain oracle 250
triples / 0 mismatch (incl. `(E)(-1)` and `0xC0000000` high-bit patterns). New
fixtures + tests for both. 71 EnumCast / 3431 fast / 103 validity, all green.

Relaunching both fixed-head reviews at `12bc344d`.

richlander:
## Round-6 review reconciliation (head `12bc344d`) — both CLEAN

Both fixed-head reviewers re-reviewed `12bc344d` (the R5 fixes) in fresh isolated
worktrees with compiled probes and runtime oracles:

- **GPT (`gpt-5.6-sol`):** VERDICT CLEAN — no significant issues.
- **Gemini (`gemini-3.1-pro-preview`):** VERDICT CLEAN — no significant issues.

Both R5 fixes verified holding: the checked-context reconcile cast no longer
throws, and the mixed-sign chain (int enum-shift | unsigned-backed enum sibling)
keeps its outer `(uint)`/`(ulong)` reinterpret for high-bit patterns.

CI on `12bc344d`: `build-net10` pass, `test (ubuntu-26.04, linux-x64)` pass,
`csharp-diff-smoke` pass.

Ready to merge

PR #3102, opened 2026-07-23
Fold direct/inline guarded heterogeneous arms via #3082 disjointness oracle

### LINKED ISSUE #3028: Fully raise InlineArrayCollectionPass::PlaceFromAddress to its switch-expression form

richlander:
- Advances #3028
- Changes: guarded heterogeneous arms on the direct/inline surface now fold to a `switch` expression, gated by #3082's type-disjointness oracle
- Card revision: `dc7b03d4`

## Change

**Conclusion:** **PASS** — a `when`-guarded direct/inline arm folds only when #3082's `AreProvablyDisjoint` proves it disjoint from every later arm, so the fold is faithful; without such a proof the pass declines and emits the pre-change if-cascade.

### Original source

```csharp
public static int GuardedArea(Shape shape, int min) => shape switch
{
    Dot d when d.Radius > min => d.Radius,
    Bar b => b.Length,
    Box x => x.Side,
    _ => -1,
};
```

### Before

```csharp
public static int GuardedArea(HeterogeneousArmSample.Shape shape, int min)
{
    Shape V_4 = shape;
    Dot d = V_4 as Dot;
    if (d is null)
    {
        if (V_4 is Bar b)
        {
            return b.Length;
        }
        if (V_4 is Box x)
        {
            return x.Side;
        }
    }
    else
    {
        if (d.Radius > min)
        {
            return d.Radius;
        }
    }
    return -1;
}
```

- Valid: True
- Correct: True

Faithful but unraised: the guarded first arm blocked the fold on the direct/inline surface (#3065/#3082 declined any guarded arm there).

### After

```csharp
public static int GuardedArea(HeterogeneousArmSample.Shape shape, int min)
{
    return shape switch
    {
        Dot d when d.Radius > min => d.Radius,
        Bar b => b.Length,
        Box x => x.Side,
        _ => -1,
    };
}
```

- Valid: True
- Correct: True

Faithful because `Dot` is provably disjoint from `Bar` and `Box`: a `Dot` that fails its guard routes to the default in both the cascade and the switch.

### Fully raised

## Change detail

Extends #3082's disjointness-gated fold from the temp-form intro cascade (#3022) to the direct/inline heterogeneous surface (#3028) for `when`-guarded arms, **reusing** #3082's `MetadataSource.AreProvablyDisjoint` and `RefutableArmsDisjointFromLaterArms` — no new oracle or seam.

- Remove the `isNewSurface && guard/subpattern => decline` gate in `TryFold`. #3082's `RefutableArmsDisjointFromLaterArms` already covers **all** arms and is the correct sole guard for a refutable non-last arm on every surface.
- Recognize csc's if/else lowering of a refutable direct intro arm (`k = place as T; if (!k) { REST } else { MATCHED }` with a shared trailing default) via a `fallThroughIsDefault` parameter, a `DefaultReachedFrom` helper, and `TryDiscoverDefault` HasElse handling. Then-only `then` blocks still require an explicit trailing default.

Does **not** close #3028: bare/multi-level property-subpattern arms do not structure to nested ifs upstream and remain a separate structuring frontier.

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `PatternSwitchExpressionPassTests`: 22 passed | `PatternSwitchExpressionPassTests`: 25 passed |
| Decompiler fast suite | 3413, 0 failed | 3416, 0 failed |
| Real witness | `HeterogeneousArmSample.GuardedArea` unraised if-cascade | `HeterogeneousArmSample.GuardedArea` fully raised switch |

Head's +2 tests are `GuardedHeterogeneousArm_RaisesOnlyWhenProvablyDisjoint` and `Synthetic_DirectGuardedNonLastArm_RaisesOnlyWhenProvablyDisjoint`; `Synthetic_InlineGuardFallsThrough_DoesNotRaise` was converted to `...FinalArm_Raises`. No baseline failures.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.PatternSwitchExpressionPassTests
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/HeterogeneousArmSample.cs
+++ b/src/ILInspector.Decompiler.Tests/HeterogeneousArmSample.cs
@@ -10,7 +10,10 @@ namespace ILInspector.Decompiler.Tests;
 /// <c>return</c>. <see cref="ILInspector.Decompiler.Pipeline.PatternSwitchExpressionPass"/>
 /// raises the whole cascade back into a single switch expression.
 ///
-/// <see cref="Area"/> is the mixed intro-chain-plus-inline shape. Isolated from
+/// <see cref="Area"/> is the mixed intro-chain-plus-inline shape (all-unguarded).
+/// <see cref="GuardedArea"/> adds a <c>when</c>-guarded first arm — the #3028
+/// follow-up that folds a refutable heterogeneous arm through the type-
+/// disjointness oracle. Isolated from
 /// <see cref="CfgSampleClass"/> so it does not perturb the fidelity gate's pinned
 /// population.
 /// </summary>
@@ -40,6 +43,20 @@ public static class HeterogeneousArmSample
     public static int Area(Shape shape) => shape switch
     {
         Dot d => d.Radius,
+        Bar b => b.Length,
+        Box x => x.Side,
+        _ => -1,
+    };
+
+    // Heterogeneous cascade whose FIRST arm is `when`-guarded over a type (Dot)
+    // disjoint from every later arm (Bar, Box). csc lowers the guarded arm to an
+    // `if (d is null) { REST } else { MATCHED }` dispatch whose default is a shared
+    // trailing return; a Dot that fails the guard is routed to that default. The
+    // fold is faithful ONLY because Dot is provably disjoint from Bar and Box, so
+    // it exercises the disjointness oracle on the direct/inline surface.
+    public static int GuardedArea(Shape shape, int min) => shape switch
+    {
+        Dot d when d.Radius > min => d.Radius,
         Bar b => b.Length,
         Box x => x.Side,
         _ => -1,
--- a/src/ILInspector.Decompiler.Tests/PatternSwitchExpressionPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/PatternSwitchExpressionPassTests.cs
@@ -509,6 +509,61 @@ public class PatternSwitchExpressionPassTests
         Assert.False(switchExpression.Arms[1].HasGuard);
     }
 
+    // Builds a direct-place cascade whose FIRST arm is an UNGUARDED then-only intro
+    // arm (`a = place as Leaf; if (!a) { REST } return 1;`) whose REST nests a
+    // guarded then-only intro arm as the LAST arm — `b = place as Twig; if (!b)
+    // return -1; if (G(b)) return 2;`. A Twig that fails its guard runs off the end
+    // of the REST block and reaches the ENCLOSING Leaf arm's value (`return 1`), NOT
+    // the trailing default. Leaf and Twig are disjoint, so the disjointness gate
+    // alone would admit the fold; only the fall-through check must decline it.
+    static IrFunction DirectGuardedLastArmFallsToPriorArm()
+    {
+        IrExpression Place() => new LoadArgument(0, "node", Node);
+        IrExpression Guard() => new Call(new MethodRef(Twig, "Ok", Bool, [Twig], HasThis: false), isVirtual: false, [new LoadLocal(1, Twig)]);
+
+        // REST (inside the Leaf arm's `then`): a guarded Twig intro arm whose guard
+        // failure falls off the end of this block into the Leaf arm's value.
+        var guardThen = new Block();
+        guardThen.Add(new Return(new Constant(2, Int32)));
+        var rest = new Block();
+        rest.Add(new StoreLocal(1, Twig, new IsInstance(Twig, Place())));
+        rest.Add(new IfStatement(new LogicalNot(new LoadLocal(1, Twig)), Return(-1), null));
+        rest.Add(new IfStatement(Guard(), guardThen, null));
+
+        var block = new Block(0);
+        block.Add(new StoreLocal(0, Leaf, new IsInstance(Leaf, Place())));
+        block.Add(new IfStatement(new LogicalNot(new LoadLocal(0, Leaf)), rest, null));
+        block.Add(new Return(new Constant(1, Int32)));
+
+        var container = new BlockContainer();
+        container.Add(block);
+        var signature = new MethodSignature(Int32, [new Parameter("node", Node)], HasThis: false, GenericParameterCount: 0);
+        return new IrFunction("M", Node, signature, ImmutableArray.Create(Leaf, Twig), container);
+
+        static Block Return(int value)
+        {
+            var b = new Block();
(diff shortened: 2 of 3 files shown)


richlander:

Two-model review per AGENTS.md (non-Opus pair), each isolated in its own linked worktree with scratch under `/tmp` and no repo mutations.

| Reviewer | Round 1 (`d7805848`) | Round 2 (fixed `dc7b03d4`) |
| --- | --- | --- |
| GPT-5.6 (`gpt-5.6-sol`) | BLOCK — Finding 1 (High), Finding 2 (Medium) | **PASS** |
| Gemini 3.1 Pro (`gemini-3.1-pro-preview`) | PASS (missed Finding 1) | **PASS** |

### Finding 1 (GPT, High) — then-only guarded matched-body fall-through mis-fold — FIXED

**Claim:** A positive-guard intro arm (`if (G) return V;`) nested as the LAST arm of a prior arm's then-only REST has its guard FAILURE run off the end of the REST block into the *prior* arm's matched value, not the default. The then-only tail check used `ReachesDefaultTail`, which accepts running off the end unconditionally, so the pass folded it and diverted the guard failure to `_ => <default>`. GPT gave an execution repro: original returns `1`, folded returns `-1`.

**Verified:** Reproduced by code trace and confirmed real. Root cause: the matched-body tail check did not consult `fallThroughIsDefault` (Gemini's round-1 point #2 reasoned about the `TryParseChain` recursion threading, which *does* reset the flag, but missed this separate tail check — the gap the two-reviewer process is meant to catch).

**Resolution (`dc7b03d4`):** `TryParseMatchedBody` now reports `bodyFallsThrough`, true only for the shapes whose refinement FAILURE can reach the following statement — the positive guarded form (`if (G) return V;`) and the property-subpattern form. The unguarded (`return V;`) and negated guarded (`if (!G) return <default>; return V;`) forms end in an unconditional return, so their tail is unreachable. The then-only tail check now uses `bodyFallsThrough ? DefaultReachedFrom(…, fallThroughIsDefault) : ReachesDefaultTail(…)`, so a falling-through matched body in a then-only REST nested in a prior arm (`fallThroughIsDefault == false`) declines instead of mis-folding. Regression test `Synthetic_GuardedLastArmFallsToPriorArm_DoesNotRaise` pins the decline even under a proving disjointness oracle (the two arm types ARE disjoint; the fold is unsound purely because of the fall-through target).

Both reviewers re-verified at `dc7b03d4`: **RESOLVED**, no over-decline (real csc temp-form folds use the negated/subpattern shapes and still fold), and the new `bodyFallsThrough` discriminator classifies every matched-body shape correctly.

### Finding 2 (GPT, Medium) — unguarded supertype arm before subtype arm → CS8510 — non-blocking, synthetic-only

**Claim:** Folding `Base => 1, Derived when G => 2` (Derived : Base) yields C# that fails CS8510 (unreachable arm). Behavior is preserved (both route a `Derived` to `Base`); the generated switch does not compile.

**Investigated with real compiled fixtures** (`/tmp/cs8510`, three methods): the negated intro-head the pass anchors on is csc's *switch* lowering, which csc guarantees CS8510-clean — a supertype arm cannot precede a subtype arm in a real switch. Compiled `switch`-then-`if` and plain if-cascades with supertype-before-subtype lower to POSITIVE inline tests, which the pass does **not** fold (it needs a negated intro head; all-inline declines). So the reported shape is synthetic-only: real csc cannot emit an intro-head with a supertype arm preceding a subtype arm.

The same synthetic exposure already exists for #3065's pre-existing unguarded direct fold (it folds two unguarded arms with no subsumption check). A disjointness-oracle-based subsumption guard would wrongly decline *valid* subtype-before-supertype switches (e.g. `string s => …, object o => …`, where `AreProvablyDisjoint(string, object)` is `false`), so it is not a cheap correct fix.

**Disposition:** GPT agreed non-blocking (no concrete compiled repro; exposure synthetic and pre-existing). Not reachable from any real compiled input; tracked as a known limitation of the direct/inline surface, not introduced by this PR.

### Evidence at fixed head `dc7b03d4`

- Focused `PatternSwitchExpressionPassTests`: 25/25.
- Decompiler fast suite (`-trait- Speed=Slow`): 3416, 0 failed.
- Real witness `HeterogeneousArmSample.GuardedArea`: fully raised switch (unchanged).

richlander:
Ready to merge

- Both adversarial reviews clean at fixed head `dc7b03d4`: GPT-5.6 **PASS** (Finding 1 RESOLVED, Finding 2 agreed non-blocking/synthetic-only), Gemini 3.1 Pro **PASS** (mis-fold class RESOLVED, no new findings).
- CI green: `test`, `build-net10`, `csharp-diff-smoke`, `changes` all pass.
- Local: focused `PatternSwitchExpressionPassTests` 25/25; decompiler fast suite 3416, 0 failed.

Non-blocking follow-up (not introduced by this PR): the direct/inline surface assumes CS8510-clean arm ordering (holds for all real csc switch-origin inputs); a subsumption guard tolerant of valid subtype-before-supertype ordering is a separate improvement shared with #3065's unguarded fold.

PR #3115, opened 2026-07-24
Fold tuple element names into the type view and separate display from identity (#2996)

richlander:
## What

Fold `[TupleElementNamesAttribute]` into Metadata's type view so `System.ValueTuple`
instantiations render as C# tuple syntax (`(int count, string name)`), and separate
that **display** spelling from the presentation-independent **identity** spelling used
by member correspondence. Part of #2996.

## Why

Rendering named tuples surfaced a latent conflation: the rendered display string
**doubled as the member identity fingerprint**. Concretely:

- Identity parsers assumed the first `(` opens the parameter list, so a **tuple return
  type** (which puts `(` at signature position 0) crashed
  `MetadataFindingsTests.RealAssemblySelfCompare_IsExact` with
  `ArgumentOutOfRangeException`.
- Tuple **element names** would have churned Member Index digests, XML-doc ids, and
  extension/instance correspondence keys — renaming a tuple element is not an API change,
  but it would have registered as one.

The same anti-pattern (structural facts re-derived from rendered display strings) already
made identity latently annotation-sensitive for NRT `?`. This PR fixes tuples structurally
rather than scrubbing display text.

## How

A canonical (presentation-independent) spelling that differs from display **only** in tuple
rendering, so every non-tuple member's canonical spelling equals its display spelling
character-for-character (byte-for-byte digest preservation by construction):

- `TypeNode.RenderCanonical()` renders the structural form
  (`System.ValueTuple<int, string>`, element names erased). Threaded via a
  `Render(bool canonicalTuples)` seam; only `GenericTypeNode` branches on it.
- `ApiParameter`/`ApiSignature` carry canonical facets (`CanonicalType`,
  `CanonicalTypeWithModifier`, `CanonicalReturnType`, `CanonicalParameterTypesSummary`),
  populated at extraction. A canonical return formatter mirrors `ref readonly` synthesis.
- Member identity, XML-doc identity, and extension/instance correspondence now read the
  canonical facets instead of display strings. Nullability and `ref readonly` are preserved;
  tuples and element names never enter identity.
- Raw-signature fallback parsers (`LegacyCanonicalMemberName`, `ExtractMemberNameWithGeneric`)
  skip a leading balanced tuple group before locating the parameter list, so a tuple-return
  member degrades instead of throwing on the JSON round-trip path.

Design recorded in `docs/design/type-spelling-identity-display.md` (two adversarial design
review rounds reconciled).

## Evidence

- `TupleTypeViewTests` (display): 22 cases over Roslyn-compiled fixtures — named/unnamed/partial,
  nested, big (9-tuple flattening), tuple-in-generic/array/func, ref tuple, field/property/event.
- `TupleIdentityTests` (new): `RenderCanonical` name-insensitivity + NRT preservation +
  non-tuple==display; canonical digest erases tuple param names to `System.ValueTuple<...>`,
  keeps `string` vs `string?` distinct, preserves `ref readonly`, and does not crash on a tuple
  return.
- `RealAssemblySelfCompare_IsExact` restored to green.
- Metadata suite: **783/783**. CLI suite: 1989 total, only the 2 pre-existing Darwin P/Invoke
  content fails remain (unrelated to this change).

## Compatibility

Non-tuple members: canonical == display, so Member Index digests, XML-doc ids, and
correspondence keys are unchanged. Round-trip identity parity for **new** JSON that persists
tuple *display* in `member.Signature` degrades safely (no crash); full round-trip canonical
parity for tuple members is a documented follow-up.

### DIFF
--- a/docs/design/type-spelling-identity-display.md
+++ b/docs/design/type-spelling-identity-display.md
@@ -1,0 +1,453 @@
+# Type spelling: identity vs. display
+
+> Design write-up for the API-shape change that lets presentation refinements
+> (tuple element names, `dynamic`, NRT annotations) reach the display without
+> contaminating member **identity**. Motivated by folding
+> `[TupleElementNamesAttribute]` into Metadata's type view (issue
+> [#2996](https://github.com/richlander/dotnet-inspect/issues/2996), Step 3),
+> which surfaced a conflation that predates tuples.
+
+## The question that started this
+
+Rendering named tuples as C# `(int count, string name)` instead of
+`System.ValueTuple<int, string>` is a **settled, fixture-backed outcome** (22
+passing display tests; the encoding is total, local, and taste-free). Wiring it
+in broke one test — `MetadataFindingsTests.RealAssemblySelfCompare_IsExact` —
+with an `ArgumentOutOfRangeException` inside the member-**identity** layer.
+
+That failure is not a tuple bug. It is the symptom of a model deficiency: the
+API shape stores every type as **one opaque display string**, and that same
+string is what member identity re-parses to build its correspondence digest.
+Every presentation refinement therefore leaks into identity.
+
+## Current shape (the conflation)
+
+A type is spelled exactly once, as a rendered string, in these slots:
+
+- `ApiParameter.Type` (and `TypeWithModifier`)
+- `ApiSignature.ReturnType`, `ApiMember.ReturnType`, `ApiMember.Signature`
+- field / property / event type strings
+
+All of them come from `TypeNode.Render()`
+(`src/ILInspector.Metadata/TypeNode.cs`), which applies **presentation
+refinements**:
+
+- NRT annotation: `string` → `string?` (lines 217/239/312/410…)
+- `dynamic`: `object` → `dynamic` (line 196)
+- proposed: tuple syntax + element names, `System.ValueTuple<int,string>` →
+  `(int count, string name)`
+
+That one string then feeds **member identity**
+(`src/ILInspector.Metadata/ApiMemberIdentity.cs`), the producer of the Member
+Index digest / correspondence key. Identity does **not** have a structural
+model to read; it re-derives structure by string-parsing the display spelling
+and laundering presentation back out:
+
+- Primary path: `signature.ParameterTypesSummary` →
+  `ApiParameter.TypeWithModifier` (a `Render()` string), then
+  `NormalizeCanonicalParameters` → `NormalizeDynamicToObject`.
+- Fallback path (JSON-round-tripped surface; `SignatureModel` is `[JsonIgnore]`):
+  re-parse `member.Signature` via `LegacyCanonicalMemberName`,
+  `ExtractCanonicalParameterList`, `ExtractCanonicalIndexerParameterList`,
+  `AbbreviateSignature`, each also passed through `NormalizeDynamicToObject`.
+
+### Why this is a root-cause conflation, not a tuple edge case
+
+**One slot, two consumers with opposite requirements.** Display wants the
+refined spelling; identity wants a presentation-independent structural spelling.
+Because they share a slot, each new refinement forces a per-feature identity-side
+scrub — or breaks identity outright:
+
+1. **`dynamic`** was patched with `NormalizeDynamicToObject` — a string scrub
+   that works only because `dynamic`/`object` is a paren-free token substitution.
+2. **NRT `?`** renders into `Type` but has **no** scrub. Identity is therefore
+   already *latently annotation-sensitive*: two builds of the same assembly under
+   different nullable context produce different fingerprints, even though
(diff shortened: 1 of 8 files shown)


richlander:

Two independent reviewers (roster: **GPT‑5.6 Sol** and **Gemini 3.1 Pro** — neither my own family) reviewed this PR across three rounds. Every reviewer ran the real build/suite in an isolated review worktree at the exact reviewed head. All findings are resolved; both fixed‑head re‑reviews on `66a9b98f` are clean.

### Round 1 — base `e15f41c4`, head `e9849f68`

- **Round‑trip identity divergence** *(GPT + Gemini, blocking)* — the canonical tuple facets live only on `ApiSignature`/`ApiParameter`, which hang off `ApiMember.SignatureModel` (`[JsonIgnore]`). After a JSON round‑trip those facets vanish and the fallback re‑derives identity by parsing `member.Signature`/`member.ReturnType`, which now carry C# tuple *display* syntax that string parsing cannot reverse to `System.ValueTuple<…>`. A persisted baseline therefore mis‑pairs tuple members against a live assembly, and element names leak into identity.
  **Fixed (`709ed09f`):** persist `ApiMember.CanonicalSignature` (`string?`, `[JsonIgnore(Condition = WhenWritingNull)]`), computed from the **live** structural model in `ApiMemberIdentity.PopulateCanonicalIdentities` at the end of extraction, only for members where `HasCanonicalDivergence` is true. `GetCanonicalSignature`/`TryGetCanonicalSignature` short‑circuit to it, so a round‑tripped surface reads the identical digest. Non‑divergent members persist nothing → zero serialized churn.

- **Invalid `ValueTuple\`8` masquerading as a genuine eight‑tuple** *(GPT, blocking)* — a `ValueTuple<…,T8>` whose 8th (`TRest`) arg is not itself a `ValueTuple` is a valid generic instantiation but not a C# tuple; it was flattened into tuple syntax and collided in display with a genuine eight‑element tuple, hiding a binary‑breaking difference.
  **Fixed (`709ed09f`):** `TypeNode.TryGetTupleElements` returns null when the 8th arg is not a `ValueTuple`, keeping generic spelling.

- **Gemini finding‑1 (`ExtractCanonicalParameterList`/`AbbreviateSignature` use the first `(`)** — Gemini suggested swapping in `IndexOfParameterListParen`. **Intentionally not taken**, and Gemini agreed after seeing the reason: that function's indexer caller `ExtractCanonicalIndexerParameterList` deliberately passes a leading‑`(` wrapper string, so skipping a leading balanced group there would return `()` and break non‑tuple indexers. The persistence approach makes tuple members bypass those parsers entirely, so the leading‑tuple scramble is unreachable for them.

### Round 2 — fixed head `709ed09f`

- **Tuple‑returning non‑conversion members escape persistence** *(GPT + Gemini, blocking)* — `HasCanonicalDivergence` only counted return‑type divergence for conversion operators, so an ordinary tuple‑returning method left `CanonicalSignature` null and fell back to display parsing, where the return tuple's leading `(` was mis‑read as the parameter list (e.g. `NamedReturn(int,string name))` vs the live `NamedReturn()`).
  **Fixed (`66a9b98f`):** dropped the `IsConversionOperator` gate — `HasCanonicalDivergence` now returns true for **any** member whose `EffectiveCanonicalReturnType` diverges from `ReturnType`. The persisted digest is computed live and correctly omits the non‑conversion return type, so live and round‑trip stay in lockstep. Fields/events leave `CanonicalReturnType` unset → no persistence, no churn. This is exactly the fix Gemini recommended. Added a parametrized `RoundTrip_TupleReturningMethod_CanonicalIdentityMatchesLive` canary over `NamedReturn`, `UnnamedReturn`, `NestedTuple`, `MidNested`, `TupleInsideFunc`, `TupleArray`.

### Round 3 — fixed head `66a9b98f` — both clean

- **GPT:** *No significant issues found.*
- **Gemini:** *Clean — no issues found. This looks solid to merge.* Independently verified persistence for extension methods, operators, and explicit‑interface implementations returning tuples; properties/indexers via `GetPropertySignature`; fields/events safely skipping persistence; deeply nested tuples in arrays/generics/byref/pointers forcing divergence through the `canonicalTuples` flag; and function‑pointer / 1‑arity `ValueTuple<T>` immunity from the `(`‑parse hazard.

### Validation (head `66a9b98f`)

- Metadata suite **794/794** (`TupleIdentityTests` 21/21).
- CLI suite **1989**, only the 2 pre‑existing Darwin P/Invoke content fails (`PInvokeMethods_MachineOutput_KeepsRawValuesWithoutCodeMarkup`, `PInvokeMethods_DeclaringTypeAndSignature_RenderAsCodeSpansWithFunctionPointerPunctuation`), unrelated to this change.
- CI green against the exact head `66a9b98f` (build‑net10, markdownlint, test linux‑x64, changes).

richlander:
Ready to merge

PR #3117, opened 2026-07-24
Share ApiSurfaceExtractor field-inclusion with RTS compile-back (#3091)

richlander:
- Advances #3091
- Changes RTS closure member-surface to consume the product's field-inclusion
  primitive; adds an opt-in product capability. One intended net behavior change
  (a field-like event's backing field is now folded, not emitted as a stray field).
- Evidence revision: `91a96a02`

## Change

**Conclusion:** **PASS** — RTS hand-rolled its own member-surface field
enumeration, duplicating the product's field-inclusion rules and diverging from
the merged #3083 field-like-event fold; this routes RTS through a shared,
product-owned `ApiSurfaceExtractor.SurfaceFieldHandles` primitive so the two
agree, proven by 47/47 product tests (incl. a real csc display-class canary) and
a byte-identical decompiler fast-suite FAIL set base-vs-head.

Before (RTS member surface for a field-like-event target):

```text
event Action Changed { add; remove; }
Action Changed;          // <- stray backing field emitted -> CS0102/CS0229
```

After:

```text
event Action Changed { add; remove; }   // backing field folded (product-owned)
```

## Scope and safety

- **Root cause:** RTS's `AddClosureMemberSurface` iterated `typeDef.GetFields()`
  and re-derived which fields to keep (its own `AutoPropertyBackingFieldNames`),
  a second implementation of a product decision — exactly what the harness-boundary
  rule warns against. It also predated #3083, so it still emitted a field-like
  event's backing field as a stray field.
- **Fix shape:** Product/shared change. `ApiSurfaceExtractor` gains an opt-in
  `includeCompilerGenerated` (default off) and a public `SurfaceFieldHandles`
  primitive that centralizes the field-inclusion decision (drop synthesized
  auto-property backing fields, the enum `value__` slot, and a field-like event's
  compiler-generated backing field; gate other `<...>` compiler-generated fields
  behind the opt-in). RTS consumes `SurfaceFieldHandles(includeAll: true,
  includeCompilerGenerated: true)` and keeps only reconstruction-side gates
  (unspeakable-name skip, signature decode, fixed buffers, pointer surface, dedup).
- **Product impact:** `ApiSurfaceExtractor.Extract` is byte-identical when the new
  opt-in is off (its default); every existing caller is unaffected. The opt-in is
  new surface, exercised only by RTS via `SurfaceFieldHandles` (RTS does not depend
  on the `Extract` flag).
- **Scope boundary:** Staged, fields-first. A surfaced compiler-generated type
  currently shows its fields + `.ctor` but not its `<...>` lambda methods; broader
  compiler-generated member coverage is a follow-up. The opt-in is off by default,
  so this has zero impact on existing behavior.
- **RTS boundary:** Preserved and strengthened. Enumeration (which fields exist)
  is now product-owned; reconstruction (how to spell/decode them: const value,
  fixed-buffer signature, pointer gating, body policy) stays in RTS. RTS gains no
  new C# knowledge.

## Evidence

| Check | Result |
| --- | --- |
| Product build (`ILInspector.Metadata`) | Pass, 0 warnings |
| Harness build (`DecompilerHarness`) | Pass, 0 warnings |
| Focused product tests | `DotnetInspector.Tests.ApiSurfaceExtractorTests` 47/47 |
| Decompiler fast suite | FAIL set **byte-identical** base (`0025d5d0`) vs head (56=56, 0 added/removed) |
| Targeted compile-back / RTS A/B | Deferred to CI — see note |

**Real witness (product):** the test assembly's own `SampleClosureHost` capturing
lambda emits a genuine csc `<>c__DisplayClass`; `Extract(includeCompilerGenerated:
false)` excludes it, `Extract(includeCompilerGenerated: true)` surfaces it with its
captured-state field members, while an auto-property `<AutoValue>k__BackingField`
stays excluded under both.

## Compile-back quality

**Conclusion:** **PASS (fidelity numbers deferred to CI)** — the intended net
change (folding a field-like event's backing field) removes a stray-field
CS0102/CS0229 *floor*, which can only make compile-back better or equal, never
worse; all other field surfaces are unchanged by construction (`SurfaceFieldHandles`
reproduces RTS's prior inclusion for real csc artifacts).

Local recompile-fidelity and RTS A/B cards are **not reproducible on this machine**:
Roslyn recompile pulls native shared-framework DLLs (`aspnetcorev2_inprocess.dll`,
`coreclr.dll`, …) into references and fails wholesale with CS0009 (`PE image
doesn't contain managed metadata`). This is a known environmental limitation on the
dev box; the decompiler fast-suite FAIL set is identical base-vs-head (the 31
`ReturnToSenderFixtureCatalogTests` reds are pre-existing CS0009, not caused by this
change). Recompile-Exact / RTS A/B coverage comes from `deep-inspect.yml`'s `test`
lane (no `Speed=Slow` filter), which I'll trigger before merge.

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| Compiler-generated `<...>` methods on a surfaced CG type | Left as frontier | Staged fields-first; opt-in is off by default so no impact. Follow-up. |
| Local recompile-Exact / RTS A/B | Deferred to CI | Environmental CS0009 on native shared-framework DLLs; not a code issue. |

## Validation

```powershell
dotnet build src/ILInspector.Metadata/ILInspector.Metadata.csproj -c Release --configfile nuget.config
dotnet build src/dotnet-inspect.Tests/dotnet-inspect.Tests.csproj -c Release --configfile nuget.config
dotnet run --project src/dotnet-inspect.Tests -c Release --no-build -- -class DotnetInspector.Tests.ApiSurfaceExtractorTests
```

## Review

Two-model adversarial review (per AGENTS.md roster) pending on fixed head `91a96a02`.

| Reviewer | Result | Notes |
| --- | --- | --- |
| _pending_ | | |
| _pending_ | | |

### DIFF
--- a/.github/workflows/deep-inspect.yml
+++ b/.github/workflows/deep-inspect.yml
@@ -103,8 +103,8 @@ jobs:
           restore-keys: |
             nuget-${{ runner.os }}-${{ runner.arch }}-
 
-      - name: Build managed tool
-        run: dotnet build src/dotnet-inspect -c Release -p:PublishAot=false
+      - name: Build solution (tool + fixtures)
+        run: dotnet build dotnet-inspect.slnx -c Release
 
       - name: Run CLI tests (including slow integration)
         run: dotnet run --project src/dotnet-inspect.Tests -c Release
--- a/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
+++ b/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
@@ -323,7 +323,10 @@ public static class ApiSurfaceExtractor
 
                 string fieldName = reader.GetString(field.Name);
                 if (!IsSurfaceableFieldName(fieldName, includeCompilerGenerated))
-                    continue; // Skip auto-property backing fields; other compiler-generated fields unless opted in
+                    continue; // Skip compiler-generated (<...>) fields unless opted in
+
+                if (IsAutoPropertyBackingField(reader, field, fieldName))
+                    continue; // Skip a synthesized auto-property backing field (re-synthesized on reconstruction)
 
                 if (IsFieldLikeEventBackingField(reader, field, fieldName, fieldLikeEventBackingFieldNames))
                     continue; // Skip a field-like event's private, compiler-generated backing field
@@ -919,15 +922,28 @@ public static class ApiSurfaceExtractor
         => name.StartsWith('<') && name.EndsWith("k__BackingField", StringComparison.Ordinal);
 
     /// <summary>
-    /// Whether a field name belongs to a type's declarable field surface. Auto-property backing
-    /// fields are always excluded; other compiler-generated (<c>&lt;...&gt;</c>) fields are
-    /// excluded unless <paramref name="includeCompilerGenerated"/> is set. Ordinary fields
-    /// (including the enum <c>value__</c> slot, which callers handle separately) are surfaced.
+    /// True when a field is a synthesized auto-property backing field: it is spelled
+    /// <c>&lt;Prop&gt;k__BackingField</c> AND carries its own <c>[CompilerGenerated]</c> marker.
+    /// Requiring the field's own marker (not the name alone) mirrors the field-like-event backing
+    /// fold: C# spelling rules do not bind arbitrary IL, so a genuine field could share the shape;
+    /// the C# compiler always marks a real auto-property backing field compiler-generated.
+    /// </summary>
+    static bool IsAutoPropertyBackingField(MetadataReader reader, FieldDefinition field, string name)
+        => IsAutoPropertyBackingFieldName(name)
+           && AttributeReader.HasAttribute(
+               reader,
+               field.GetCustomAttributes(),
+               KnownAttributeNames.CompilerGeneratedAttribute);
+
+    /// <summary>
+    /// Whether a field name belongs to a type's declarable field surface based on its name alone.
+    /// Compiler-generated (<c>&lt;...&gt;</c>) fields are excluded unless
+    /// <paramref name="includeCompilerGenerated"/> is set; ordinary fields are surfaced. Backing
+    /// fields (auto-property, field-like event) and an enum's <c>value__</c> slot carry additional
+    /// positive-evidence checks applied by callers.
     /// </summary>
     static bool IsSurfaceableFieldName(string name, bool includeCompilerGenerated)
     {
-        if (IsAutoPropertyBackingFieldName(name))
-            return false;
         if (name.StartsWith('<'))
             return includeCompilerGenerated;
         return true;
@@ -935,9 +951,10 @@ public static class ApiSurfaceExtractor
 
     /// <summary>
     /// The field handles that make up a type's declarable field surface: ordinary fields,
-    /// excluding synthesized auto-property backing fields, the enum storage slot (<c>value__</c>),
-    /// and a field-like event's compiler-generated backing field. Compiler-generated fields (e.g.
-    /// state-machine hoisted locals, display-class captures) are included only when
+    /// excluding synthesized auto-property backing fields (positive <c>[CompilerGenerated]</c>
+    /// evidence), an enum's storage slot (<c>value__</c>), and a field-like event's
+    /// compiler-generated backing field. Compiler-generated fields (e.g. state-machine hoisted
+    /// locals, display-class captures) are included only when
     /// <paramref name="includeCompilerGenerated"/> is set; non-public fields only when
     /// <paramref name="includeAll"/> is set. This is the single field-inclusion decision shared by
     /// API-surface extraction and compile-back reconstruction so both agree on which fields a type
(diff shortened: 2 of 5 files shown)


richlander:

Two-model adversarial review per `AGENTS.md` (I authored this, so I used the two other families: **Gemini 3.1 Pro** and **GPT-5.6**). Each fixed head was reviewed in an isolated linked worktree with scratch confined to `/tmp`. Both reviewers are **clean** on the final reviewed head `5b669fa7`.

### Findings by round

**Round 1 — head `91a96a02`**

- **`value__` dropped on non-enum types** (both reviewers). The `value__` skip fired on any type, not just enums, dropping a legitimate ordinary field named `value__`. **Fixed** in `3188a21f`: the skip is gated on `IsEnum(reader, typeDef)`; a non-enum `value__` now surfaces.
- **`<Prop>k__BackingField` name-only overmatch** (both reviewers). Folding on the mangled name shape alone could suppress a genuine same-named field, since C# spelling rules do not bind arbitrary IL. **Fixed (v1)** in `3188a21f`: require the field's own `[CompilerGenerated]` marker (positive evidence).
- **Binary-compat of the new `Extract` parameter** (GPT). **Dismissed:** `ILInspector.Metadata` is not independently packaged; `Extract` is consumed in-repo only, and the new parameter is optional (default preserves behavior).

**Round 2 — head `3188a21f`**

- **Gemini: clean.**
- **`name + field-CG` still broader than old RTS** (GPT). The deleted compile-back planner keyed on **property existence** (its `IrImporter` maps a `<X>k__BackingField` reference to property `X` only when `X` exists; otherwise the raw field must stay declared). **Fixed (v2)** in `20d15cab`: fold only when a matching declared property exists.

**Round 3 — head `20d15cab`**

- **Unrelated / mis-typed backing field folded → data loss** (Gemini Finding B; GPT type/staticness variant — same root cause). Verified against `git 91a96a02^:tools/DecompilerHarness/ReturnToSenderTypePlanner.cs`: the planner's `IsAutoProperty`/`IsAutoPropertySetter` folded a `<Prop>k__BackingField` only when the property was a **genuine auto-property** (accessor `[CompilerGenerated]`) **and** the field's **type and staticness** matched the property. My property-existence gate had dropped those three checks. **Fixed (v3)** in `5b669fa7`: restored the full discriminator (see below).
- **Explicit-interface `<I.P>k__BackingField` surfaces** (Gemini Finding A). **Non-actionable:** this is behavior-identical to the migrated planner (which also skips `<`/`.` property names), so it is not a regression; and Gemini independently verified that RTS's `AddClosureMemberSurface` explicitly `continue`s on any surfaced field name containing `.`, so the field is dropped on the reconstruction side and never reaches emission — **no `CS0102` is possible.**

**Round 4 — head `5b669fa7` (final)**

- **Gemini: clean.** Confirmed `Extract(includeCompilerGenerated:false)` byte-equivalence, the Finding B resolution, Finding A inertness (verified above), and that the C# 13 `field`-keyword case is soundly **preserved** (its accessors are user-authored, so it is treated as non-auto and its backing field is kept — matching the old planner).
- **Cross-assembly same-name type-string conflation** (GPT). `SignatureDecoder` renders type names without assembly identity, so `LibA::N.T` and `LibB::N.T` decode equal. **Withdrawn by GPT as non-actionable:** unreachable from any compiler's output (a backing field always shares its property's exact type in the same compilation), behavior-identical to the migrated planner (which compared the same `SignatureDecoder`-derived display strings via `GuardedSignatureText.FieldText` + `CompileBackTypeSignature.Display`), and inert in RTS (field references map to the property by name). Scope-aware structural identity is a separate, Metadata-wide design question, not this behavior-preserving migration.

### The final discriminator (`5b669fa7`)

`AutoPropertyBackingFieldDescriptors` builds a per-type map from **declared auto-properties** — property name has no `<`/`.`, at least one accessor is `[CompilerGenerated]` (the auto signal), and the return type decodes — recording `(returnType, accessorStaticness)`. A candidate `<Prop>k__BackingField` is folded only when it is `[CompilerGenerated]`, its staticness matches, and its decoded type equals the property return type. Property and field types are decoded through the same `GuardedSignatureText`/`SignatureDecoder.Instance` with the same `GenericContext.ForType`, so string equality is a sound type match. This is symmetric with the field-like-event fold and faithful to the deleted planner.

### Evidence

- `ApiSurfaceExtractorTests` **47/47**, with new synthetic cases asserting **preservation** (surfaced under opt-in, absent by default) for type-mismatch, staticness-mismatch, and non-auto-property lookalikes; the genuine `int`/`int` instance auto-property still folds.
- Real-corpus dogfood over three csc-compiled product assemblies folds **3208** genuine auto-property backing fields with **0 leaks**, confirming the type-string comparison matches real compiler output across generics/records/strings/arrays/tuples.
- `Extract` default output is byte-identical with `includeCompilerGenerated:false` (all `<...>` names are dropped by `IsSurfaceableFieldName` before the fold).
- Full solution builds; `ILInspector.Metadata.Tests` **793/794** (the one failure is a pre-existing local-only Windows backslash-path assertion, CI-green).

### Post-review integration

`origin/main` (`e11a4fac`) was merged in `d9d08336`. It touches non-overlapping regions of `ApiSurfaceExtractor.cs` (event tuple-names, canonical member identities, method-return canonical rendering) and does not interact with the field-fold path — my type comparison uses `GuardedSignatureText`/`SignatureDecoder`, independent of main's `TypeNode`/`ApplyTupleNames` display changes. Re-verified on the merged head: 47/47, Metadata 793/794 (same known env failure), dogfood 3208 folds / 0 leaks. This is a mechanical, non-interacting integration and does not require re-review.

PR #3107, opened 2026-07-23
Raise short-circuit &&/|| from bool-constant-arm ternaries (#3071)

richlander:
- Advances #3071
- Changes decompiled C# so a short-circuit `&&`/`||` the compiler lowered to a
  branch diamond re-forms as the operator it spelled, instead of a
  bool-constant-arm ternary
- Card revision: `64db03216fd380ebd539c3bfb4f83d2b673c21d0`

## Change

> Should we accept this change?

**Conclusion:** **PASS** — the new `ShortCircuitTernaryPass` re-forms only the two
opcode-exact bool-constant-in-`when-true` ternary forms (`c ? true : y` → `c || y`,
`c ? false : y` → `!c && y`), declines every branch-polarity- and branchless-hazard
shape, and the PR-quick corpus is neutral with zero pass bugs.

### Original source

```csharp
static string EscapeReservedKeywordIdentifiers(string text)
{
    var sb = new StringBuilder(text.Length);
    for (int i = 0; i < text.Length;)
    {
        if (IsIdentifierStart(text[i]))
        {
            int start = i++;
            while (i < text.Length && IsIdentifierPart(text[i]))
                i++;
            string token = text[start..i];
            bool alreadyEscaped = start > 0 && text[start - 1] == '@';
            sb.Append(!alreadyEscaped && CSharpKeywords.RequiresDeclarationEscape(token) ? EscapeIdentifier(token) : token);
            continue;
        }
        sb.Append(text[i++]);
    }
    return sb.ToString();
}
```

### Before

```csharp
string EscapeReservedKeywordIdentifiers(string text)
{
    StringBuilder sb = new(text.Length);
    int i = 0;
    while (i < text.Length)
    {
        if (IsIdentifierStart(text[i]))
        {
            start = i++;
            while (i < text.Length && IsIdentifierPart(text[i]))
            {
                i++;
            }
            V_5 = start;
            token = text.Substring(V_5, i - V_5);
            S_2 = !((start <= 0 ? false : (text[start - 1] == '@')) || !CSharpKeywords.RequiresDeclarationEscape(token)) ? EscapeIdentifier(token) : token;
            sb.Append(S_2);
        }
        else
        {
            sb.Append(text[i++]);
        }
    }
    return sb.ToString();
}
```

- Valid: True
- Correct: True

The surviving nested residue is the bool-constant-arm ternary
`start <= 0 ? false : (text[start - 1] == '@')`. It is valid and behavior-preserving,
[...]

PR #3119, opened 2026-07-24
Guard FoldGuardReturn's short-circuit fold against invalid/divergent C# (#3114)

### LINKED ISSUE #3114: FoldGuardReturn re-forms short-circuit returns without opcode-fidelity guards
## Summary

`BooleanFoldingPass.FoldGuardReturn` re-forms a standalone guarded return
(`if (c) return A; return B;` / expression-bodied `=> c ? false : y`) into a
short-circuit `&&`/`||` **without any of the opcode-fidelity guards** that
`ShortCircuitTernaryPass` (PR #3107) applies to the *nested* form of the same
rewrite. As a result it can emit C# whose recompilation diverges in branch
opcodes, operator tokens, or runtime behavior.

This is **pre-existing** and independent of #3107 — it reproduces on
`origin/main` at `e15f41c4` with a stock harness dump. #3107 deliberately scopes
to the nested `ShortCircuitTernaryPass` and files this as follow-up.

## Repro (base e15f41c4, `--dump`)

| Method | FoldGuardReturn output | Hazard |
| --- | --- | --- |
| `bool F(double a, double b, bool y) => a < b ? false : y;` | `!(a < b) && y` | float dual flips ordered→unordered branch (`blt.s`→`blt.un.s`) |
| `bool F(string a, string b, bool y) => a == b ? false : y;` | `a != b && y` | `op_Equality`→`op_Inequality` — different call token |
| `bool F(bool c, bool y) => c ? false : y;` | `!c && y` | csc collapses `&&` over a bare local to branchless `&` (branch→`and`) |
| `TruthOver a; bool F(TruthOver a, int x) => a ? true : x > 0;` | `a \|\| x > 0` | rebinds user-defined `op_BitwiseOr` + `op_True`/`op_False`; reselects overload |
| `bool F(bool c, ref bool y) => c ? false : y;` | `!c && y` | managed by-ref collapses branchless + eagerly derefs a guarded location (NRE divergence on null by-ref) |

(`TruthOver` is a struct with `operator true`/`operator false`/`operator |`.)

## Fix direction

Port the three guards `ShortCircuitTernaryPass` already carries onto
`FoldGuardReturn`:

1. **Truthiness gate** — decline when the condition is
   `Call { Callee.Name: "op_True" or "op_False" }`.
2. **Integer-negate gate** — for the negating (B) form, fire only when the
   condition is a confirmed primitive-integer `Comparison`
   (`NegateIsIntegerComparisonDual`); decline float/reference/enum comparisons and
   operator-call conditions.
3. **Branchless-place gate** — decline when the surviving operand renders as a
   bare place (`LoadLocal`/`LoadArgument`/`LoadStackSlot`) or a managed-by-ref
   dereference (`LoadIndirect` over a `ByRef` address).

These predicates are already implemented and validated in
`ShortCircuitTernaryPass` (`IsUserDefinedTruthiness`,
`NegateIsIntegerComparisonDual`, `RendersAsBranchlessBarePlace`); the follow-up
is to share or mirror them in `FoldGuardReturn`.

Refs #3107.

richlander:
- Fixes #3114
- Changes `BooleanFoldingPass.FoldGuardReturn` to decline the two short-circuit re-forms that produce invalid or behavior-changing C#, so the faithful guarded return stands.
- Card revision: `9f29f1c0`

## Change

**Conclusion:** **PASS** — declines exactly the two folds whose output the correctness bar rejects (user-defined-truthiness condition → invalid C#; managed by-ref surviving operand → NRE/eval divergence) while keeping every opcode-divergent-but-valid readability re-form; the pinned corpus is fully neutral.

`FoldGuardReturn` re-forms `if (c) return A; return B;` into a short-circuit `&&`/`||` whenever an arm is a bool constant. Sibling `ShortCircuitTernaryPass` (#3107) grew two correctness guards during adversarial review; `FoldGuardReturn` lifts the same lowered branch shape and carried the same two hazards. This factors both predicates into a shared `ShortCircuitFidelity` helper and applies them in `FoldGuardReturn` before any tree mutation, so the two passes' hazard sets cannot drift apart.

Scope is **correctness-only**: the opcode-divergent-but-valid re-forms (block-reordered guard clause, both-constant `return c`, bare-local/pointer operands) remain accepted readability raises and still fold.

### Witness 1 — user-defined truthiness (invalid C#)

Minimal repro (`TruthOver` defines `operator true`/`operator false` and a user-defined `&`):

```csharp
static bool TruthAnd(TruthOver t, int y) { if (t) return G(y); return false; }
```

#### Before

```csharp
return t && G(y);
```

- Valid: **False** — `t` is `TruthOver`, so `t && G(y)` binds the user-defined `&` and yields `TruthOver`, not `bool`; the `return` needs a nonexistent `TruthOver`→`bool` conversion (CS0029).
- Correct: **False** — does not compile.

#### After

```csharp
if (t)
{
    return G(y);
}
return false;
```

- Valid: **True**
- Correct: **True** — faithful guarded return; `if (t)` binds `operator true` exactly as the original.

### Witness 2 — managed by-ref surviving operand (recompile divergence)

```csharp
static bool ByRefOperand(int s, ref bool r) { if (s > 0) return r; return false; }
```

#### Before

```csharp
return s > 0 && r;
```

- Valid: **True**
- Correct: **False** — `r` is a managed by-ref deref, which csc treats as non-null and side-effect-free, so it recompiles `s > 0 && r` to a **branchless** `&` that dereferences `r` even when `s <= 0` — an observable `NullReferenceException`/eager-eval divergence from the original branch.

#### After

```csharp
if (s > 0)
{
    return r;
}
return false;
```

- Valid: **True**
- Correct: **True** — faithful guarded return; `r` is read only when `s > 0`.

### Fully raised

The After decompilations are in the fully raised state: declining these two folds is the intended endpoint, because the folded forms cannot be made both valid and behavior-preserving.

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | n/a (new file) | `FoldGuardReturnFidelityTests`: 11 passed |
| Refactor regression | `ShortCircuitTernaryPassTests`+`BooleanFolding*`+`IrImporterTests`+`LoweringCoverageTests`: 116 passed | 116 passed |
| Decompiler fast suite (`-trait- "Speed=Slow"`) | 3507, 56 failed | 3507, 56 failed |

The 56 fast-suite failures are pre-existing environmental reds on this machine, unchanged by this PR (identical name set): Roslyn compile-back gates hitting `CS0009` on native shared-framework DLLs (`ReturnToSenderFixtureCatalogTests` ×31, `FidelityCheckGeneratedFilterTests` ×15, `TypeSourceCheckTests` ×4, `LadderRung6GateTests` ×3, `NullCoalescingAssignmentPassTests`), one CRLF-vs-LF (`StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString`), and one partial-build missing fixture (`ExpressionTreeLambdaTests.LookalikeExpressionAssembly_StaysFactoryCalls`). Zero failures touch `FoldGuard`/`ShortCircuit`/`BooleanFolding`.

## Decompiler quality

**Conclusion:** **PASS** — matched-env PR-quick corpus A/B is fully neutral; a decline-only change cannot regress residue or introduce a pass bug.

### PR quick gate

Run: PR quick corpus, hash-stable 100 methods per assembly, `--compile-cap 0` (compile-back is env-broken on this machine; the metrics below are IL-derived and env-independent). 15 assemblies, 1,500 methods. Base snapshot generated from `origin/main` (60bc3f12) over the identical assembly set, so only the decompiler differs.

| --- | ---: | ---: | ---: |
| Detected lowering residue (-) | 249 (16.60%) | 249 (16.60%) | 0 |
| Conditional-branch residue (-) | 44 (2.93%) | 44 (2.93%) | 0 |
| Forward-merge stops (-) | 30 (2.00%) | 30 (2.00%) | 0 |
| Pass bugs (-) | 0 | 0 | 0 |

> **Conclusion:** **PASS** — no corpus method hits the truthiness/by-ref guarded-return hazard; the declines fix the narrow invalid/divergent cases without any readability regression.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.FoldGuardReturnFidelityTests
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/FoldGuardReturnFidelityTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FoldGuardReturnFidelityTests.cs
@@ -143,6 +143,47 @@ public class FoldGuardReturnFidelityTests
         AssertDeclined(CondCompare(), Bool(false), ByRefBoolDeref(), [Int32]);
     }
 
+    // === correctness declines: the wrapped forms the fold's own downstream
+    // transforms would strip back to a bare hazard (adversarial review, #3119) ===
+
+    [Fact]
+    public void NegatedTruthinessCondition_TailConstant_IsNotFolded()
+    {
+        // `if (!t) return computed; return true;` lifts the condition through
+        // Conditions.Negate on the `||` arm, which unwraps the LogicalNot and re-exposes
+        // the bare `op_True(t)` — the printer then strips it to `return t || computed;`,
+        // which binds the user-defined `|` (typed TruthOver, not bool): non-compiling.
+        // The look-through peel must see the truthiness call under the negation.
+        AssertDeclined(new LogicalNot(UserTruthiness()), Computed(), Bool(true), [Int32]);
+    }
+
+    [Fact]
+    public void ByRefDereferenceUnderEqualsTrueComparison_IsNotFolded()
+    {
+        // `if (c) return *r == true; return false;` folds to `c && (*r == true)`, then
+        // BooleanFoldingPass's fixpoint reduces `*r == true` to the bare `*r` — the same
+        // branchless eager-deref divergence the raw-operand guard catches. The peel must
+        // see the by-ref deref under the identity comparison the pass will strip.
+        var byRefEqualsTrue = new Comparison(
+            ComparisonKind.Equal, isUnsigned: false, ByRefBoolDeref(), Bool(true));
+        AssertDeclined(CondCompare(), byRefEqualsTrue, Bool(false), [Int32]);
+    }
+
+    [Fact]
+    public void ByRefDereferenceUnderEqualsFalseComparison_StillFolds()
+    {
+        // `if (c) return *r == false; return false;` folds to `c && (*r == false)`, which
+        // the fixpoint reduces to `c && !*r`. The negating form keeps a `!` (and, per the
+        // merged round-2 boundary, a branch), so the surviving operand is not a bare
+        // deref and the readable fold is kept — the peel declines ONLY the bare-producing
+        // identity forms (`== true` / `!= false`), not `== false` / `!= true`.
+        var byRefEqualsFalse = new Comparison(
+            ComparisonKind.Equal, isUnsigned: false, ByRefBoolDeref(), Bool(false));
+        var result = RunGuard(CondCompare(), byRefEqualsFalse, Bool(false), [Int32]);
+
+        Assert.IsType<LogicalBinary>(result);
+    }
+
     static IrExpression RunGuard(
         IrExpression condition, IrExpression thenValue, IrExpression tailValue, ImmutableArray<TypeRef> locals)
     {
--- a/src/ILInspector.Decompiler.Tests/FoldGuardReturnFidelityTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FoldGuardReturnFidelityTests.cs
@@ -233,11 +233,16 @@ public class FoldGuardReturnFidelityTests
     // becomes a bare operand of the emitted OUTER-kind chain is eagerly dereferenced
     // (NRE divergence) and is declined — but one confined to a compound operand that
     // keeps its branch (a bitwise `&`, a different-kind logical, a call argument) folds
-    // faithfully. The reachable-from-csc shapes below (bitwise `a & r`, different-kind
-    // `*r || a`, `*r > 0`, `*r.b`, and the call-barrier chain `*r && Call()`) were each
-    // verified against SDK-csc /optimize IL plus a null-by-ref runtime probe; the pure
-    // same-kind `LogicalBinary` cases are synthetic predicate tests (csc pre-lowers a
-    // bare-operand `a && *r` to bitwise), noted per test. ===
+    // faithfully. The csc-REACHABLE shapes below were each verified against SDK-csc
+    // /optimize IL plus a null-by-ref runtime probe: the bitwise `a & r`, the comparison
+    // `*r > 0`, the by-ref field `r.b`, and the call-barrier same-kind chain `*r && Call()`
+    // (a call keeps csc from collapsing the `&&`). The `LogicalBinary` cases built over
+    // BARE operands (`a && *r`, `*r || a`) and the `== true`-wrapped cases are SYNTHETIC
+    // predicate tests: csc lowers a bare-operand `&&`/`||` to a bitwise `Binary` (so those
+    // sources import as the `a & r` case, not `LogicalBinary`) and constant-folds
+    // `x == true`. A csc-reachable `LogicalBinary` needs a compound (call) operand
+    // (`*r && Call()`, `*r || Call()`); the reachable analog of a wrapped chain is the
+    // `!`-negated `!(*r && Call())`. Synthetic status noted per test. ===
 
     [Fact]
     public void ByRefUnderLogicalAndOperand_TailConstant_IsNotFolded()
@@ -272,14 +277,16 @@ public class FoldGuardReturnFidelityTests
     [Fact]
     public void ByRefInLogicalAndChainUnderEqualsTrueWrapper_TailConstant_IsNotFolded()
     {
-        // #3119 round-5 finding 1: a same-kind chain hidden under a reducible `== true`
-        // wrapper. `if (c) return (*r && Call()) == true; return false;` → the fold lifts
-        // `(*r && Call()) == true`; FoldBoolConstantComparison then strips `== true`,
-        // leaving `c && *r && Call()`, whose call-free `c && *r` prefix collapses
-        // branchless (eager by-ref deref — same NRE divergence as the unwrapped chain).
-        // The peel must run BEFORE the same-kind flatten so the buried by-ref is reached;
-        // flattening first would treat the whole `== true` comparison as an opaque compound
-        // and wrongly fold. (The unwrapped chain is `ByRefBeforeCallInLogicalAndOperand`.)
+        // #3119 round-5 finding 1 / round-6: SYNTHETIC defensive predicate test. csc
+        // constant-folds `(*r && Call()) == true` to `*r && Call()` (verified: the wrapped
+        // and unwrapped sources emit byte-identical IL, no `ceq`), so this `== true` node
+        // is NOT csc-reachable. The csc-REACHABLE analog that drives the same peel-then-
+        // recurse path is the `!`-negated chain `!(*r && Call())` (csc keeps the `&&`
+        // branch and appends a `ceq`), which the parity-agnostic peel conservatively
+        // declines for the same reason as a bare `!*r`. This pins that the peel runs BEFORE
+        // the same-kind flatten: IF a reducible wrapper around a same-kind chain reaches
+        // the fold, its buried by-ref is still reached (flattening first would treat the
+        // wrapper as an opaque compound and miss it).
         var call = new Call(
             new MethodRef(Holder, "Call", Boolean, [], HasThis: false), isVirtual: false, []);
         var chain = new LogicalBinary(LogicalKind.And, ByRefBoolDeref(), call);
@@ -290,11 +297,11 @@ public class FoldGuardReturnFidelityTests
     [Fact]
     public void LocalOnlyLogicalAndChainUnderEqualsTrueWrapper_TailConstant_StillFolds()
     {
-        // Negative for finding 1: the same `== true`-wrapped same-kind chain WITHOUT a
-        // by-ref folds. `if (c) return (a && Call()) == true; return false;` reduces to
-        // `c && a && Call()`; the branchless `c & a` prefix is a valid readability raise
-        // (a local never faults), so the guarded-return fold keeps it. The peel-then-
-        // flatten reaches the local `a` but the by-ref hazard predicate ignores it.
+        // Negative for the peel-then-recurse: the same synthetic wrapped same-kind chain
+        // WITHOUT a by-ref still folds. The peel reaches the local `a`, but the by-ref
+        // hazard predicate only fires on a by-ref deref, so a local operand is a valid
+        // readability raise (a local never faults) and the guarded-return fold keeps it.
+        // (Synthetic like the sibling: csc constant-folds `== true`.)
         var call = new Call(
             new MethodRef(Holder, "Call", Boolean, [], HasThis: false), isVirtual: false, []);
         var chain = new LogicalBinary(LogicalKind.And, new LoadLocal(1, Boolean), call);
@@ -307,10 +314,12 @@ public class FoldGuardReturnFidelityTests
     [Fact]
     public void ByRefInSameKindLogicalOperand_TailTrue_OuterOr_IsNotFolded()
     {
-        // The lift is `||`-kind here (`if (c) return *r || a; return true;` → `!c || (*r
-        // || a)`). Now the inner `||` matches the outer `||`, flattens to `!c || *r || a`,
-        // and collapses branchless — the by-ref is eager. The SAME operand folds under an
-        // `&&` lift (next test); the hazard is outer-kind-relative.
+        // SYNTHETIC same-kind (`Or`) chain under an `Or` lift. A bare `*r || a` lowers to a
+        // bitwise `Binary.Or` (faithful, folds), so the csc-reachable same-kind hazard
+        // needs a call barrier: `*r || Call()` → `!c || (*r || Call())` flattens to
+        // `!c || *r || Call()`, whose `!c || *r` prefix collapses branchless (eager by-ref).
+        // This pins the outer-kind-relative decline on the `LogicalBinary(Or)` shape; the
+        // SAME shape folds under an `&&` lift (next test).
         var operand = new LogicalBinary(LogicalKind.Or, ByRefBoolDeref(), new LoadLocal(1, Boolean));
         AssertDeclined(CondCompare(), operand, Bool(true), [Int32, Boolean]);
     }
@@ -318,11 +327,13 @@ public class FoldGuardReturnFidelityTests
     [Fact]
     public void ByRefInDifferentKindLogicalOperand_TailConstant_StillFolds()
     {
-        // #3119 round-4 finding 2: `if (c) return *r || a; return false;` → `c && (*r ||
-        // a)`. The inner `||` differs from the `&&` lift, so the printer keeps it a
+        // #3119 round-4/round-6: a different-kind `LogicalBinary(Or)` operand under an
+        // `&&` lift folds. The `||` differs from the `&&` lift, so the printer keeps it a
         // parenthesized compound the outer `&&` branch guards — the by-ref is read only
-        // when c is true, exactly as in the original (verified: IL byte-identical to the
-        // guarded original, null-by-ref does NOT throw). A faithful, foldable raise.
+        // when c is true. (SYNTHETIC IR: a bare `*r || a` lowers to a bitwise `Binary.Or`,
+        // = the `ByRefUnderBitwiseAndOperand` case which was verified byte-identical; the
+        // csc-reachable different-kind `LogicalBinary` needs a call operand, `*r || Call()`.
+        // The predicate's fold decision is identical for either.)
         var operand = new LogicalBinary(LogicalKind.Or, ByRefBoolDeref(), new LoadLocal(1, Boolean));
         var result = RunGuard(CondCompare(), operand, Bool(false), [Int32, Boolean]);
 
@@ -354,6 +365,38 @@ public class FoldGuardReturnFidelityTests
         var operand = new Call(
             new MethodRef(Holder, "Use", Boolean, [Boolean], HasThis: false),
             isVirtual: false, [ByRefBoolDeref()]);
+        var result = RunGuard(CondCompare(), operand, Bool(false), [Int32]);
+
+        Assert.IsType<LogicalBinary>(result);
+    }
+
+    [Fact]
+    public void ByRefInComparisonOperand_TailConstant_StillFolds()
+    {
+        // `if (c) return *r > 0; return false;` (int by-ref) → `c && (*r > 0)`. The by-ref
+        // deref is confined to a comparison, a compound operand csc keeps the `c` branch
+        // for, so it is read only when c is true (verified round-4: IL byte-identical to
+        // the guarded original, null-by-ref does NOT throw). This pins that the guard does
+        // NOT decline a by-ref nested in a comparison — a regression guard against the
+        // round-3 subtree scan, which wrongly declined any by-ref anywhere in the operand.
+        var byRefInt = new LoadIndirect(Int32, new LoadArgument(1, "r", TypeRef.ByRef(Int32)));
+        var operand = new Comparison(ComparisonKind.GreaterThan, isUnsigned: false, byRefInt, new Constant(0, Int32));
+        var result = RunGuard(CondCompare(), operand, Bool(false), [Int32]);
+
+        Assert.IsType<LogicalBinary>(result);
+    }
+
+    [Fact]
+    public void ByRefStructFieldOperand_TailConstant_StillFolds()
+    {
+        // `if (c) return r.b; return false;` (by-ref struct `r`, bool field `b`) →
+        // `c && r.b`. The field access reads through the by-ref address with `ldfld` — it
+        // is not a bare `ldind` bool deref, so it is a compound operand csc keeps the `c`
+        // branch for and stays foldable (verified round-4: IL byte-identical, no null-by-
+        // ref throw). Pins non-decline for a by-ref struct field access.
+        var structRef = TypeRef.CoreLib("Synthetic", "S");
+        var field = new FieldRef(structRef, "b", Boolean);
+        var operand = new LoadField(field, new LoadArgument(1, "r", TypeRef.ByRef(structRef)));
         var result = RunGuard(CondCompare(), operand, Bool(false), [Int32]);
 
         Assert.IsType<LogicalBinary>(result);
--- a/src/ILInspector.Decompiler.Tests/FoldGuardReturnFidelityTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FoldGuardReturnFidelityTests.cs
@@ -144,7 +144,12 @@ public class FoldGuardReturnFidelityTests
     }
 
     // === correctness declines: the wrapped forms the fold's own downstream
-    // transforms would strip back to a bare hazard (adversarial review, #3119) ===
+    // transforms would strip or invert back to a bare hazard (adversarial review,
+    // #3119). The peel is a CONSERVATIVE, negation-parity-agnostic over-approximation:
+    // it follows a hazard through any chain of logical negations and bool-constant
+    // comparisons (both directions), so it also declines the branch-preserving `!x`
+    // forms (`== false` / `!= true`). Declining an extra readable raise is sound;
+    // modeling parity exactly would re-implement FoldBoolConstantComparison + Negate. ===
 
     [Fact]
     public void NegatedTruthinessCondition_TailConstant_IsNotFolded()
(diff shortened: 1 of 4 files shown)


richlander:
## Round-1 adversarial review reconciliation (head `9f29f1c0` → fix `c2b66730`)

Two reviewers ran the round-1 review at head `9f29f1c0`: **gpt-5.6-sol** and **gemini-3.1-pro-preview**. They disagreed, so I reproduced each finding independently by constructing the exact IR and printing the folded output (`CSharpPrinter.Print`). Verdicts:

### GPT Finding 1 — negated user-defined truthiness → invalid C# — **CONFIRMED, fixed**
`if (!op_True(t)) return computed; return true;` lifts the condition through the `||`-arm's `Conditions.Negate`, which unwraps the `LogicalNot` and re-exposes the bare `op_True(t)`. Direct print of the folded IR:

```
IR = Return(Or(Call(op_True), Cmp.GreaterThan))
C# = return t || x > 0;   // invalid: binds the user-defined | on TruthOver, not bool
```

Root cause: the guard inspected the *pre-fold* condition (`!op_True(t)`, a `LogicalNot`, not a `Call`) and missed it. **Gemini round-1 dismissed this as benign, arguing the printer renders `t.op_True()` (valid).** That is disproven above: `&&`/`||` operands route through `Condition()` (`LogicalOperandText` → `RenderedCondition(side)`, CSharpPrinter.cs ~3576 → `Condition()` ~2876), which strips `op_True(a)` to bare `a`. Fixed in `c2b66730`: `IsUserDefinedTruthiness` peels leading `LogicalNot`s before matching.

### GPT Finding 2 — managed by-ref deref under `== true` → branchless eager deref — **CONFIRMED, fixed**
`if (s > 0) return *r == true; return false;` folds to `(s>0) && (*r == true)`, then `FoldBoolConstantComparison` reduces `*r == true` → bare `*r` in a later fixpoint iteration. Direct print:

```
IR = Return(And(Cmp.GreaterThan, LoadIndirect[ByRef]))
C# = return s > 0 && r;   // csc collapses to branchless &, eagerly derefs a guarded by-ref → NRE divergence
```

Fixed in `c2b66730`: `IsManagedByRefDeref` peels the identity-preserving bool comparisons (`x == true` / `x != false`, the bare-operand-producing forms) before checking for a managed by-ref deref. It deliberately does not peel `x == false` / `x != true` (→ `!x`, which keeps a branch).

### GPT Finding 3 (Med) — name-only `op_True`/`op_False` match — **dismissed, out of scope**
The printer's `Condition()` strips `op_True`/`op_False` **by name** (CSharpPrinter.cs:2876-2877), independent of this pass. Whether the guard folds or declines, an ordinary IL-authored method literally named `op_True` in condition position is mis-rendered identically — so tightening only the pass (e.g. gating on `Callee.IsOperator`) without also changing the printer would not improve correctness. The pass deliberately mirrors the printer's name-based recognition; a printer-level fix is pre-existing and out of `#3114`'s correctness scope. Gemini also considered this over-fire sound.

### Verification (fixed head)
After `c2b66730` both shapes decline to faithful, valid output: `if (s > 0) { return r; } return false;` (deref stays guarded) and `if ((t ? false : true)) { return x > 0; } return true;` (valid truthiness spelling). Three regression tests added. Focused suites green (FoldGuardReturnFidelityTests 13, ShortCircuitTernaryPassTests 21, BooleanFoldingSourceOffsetTests 14, IrImporterTests 65, LoweringCoverageTests 6); fast decompiler suite = 55 pre-existing environmental reds only (0 in touched passes); corpus A/B neutral (residue 249→249, pass bugs 0→0, PASS).

A fixed-head re-review (`46cd7432`) by both models is in progress; I will post its outcome before marking ready.

richlander:
## Round-2 fixed-head review reconciled → follow-up `8e639dee`

Round-2 re-review of `46cd7432` (gpt-5.6-sol + gemini-3.1-pro-preview, isolated fixed-head worktrees). Both models independently found the round-1 look-through peels were **too shallow**; I reproduced each by printing the folded IR (`CSharpPrinter.Print`) and fixed all correctness findings.

### Findings (attributed)

| # | Finding | Sev | Model(s) | Reproduced | Resolution |
| - | ------- | --- | -------- | ---------- | ---------- |
| 1 | Truthiness under bool-const comparison `op_True(t) == true` bypasses the LogicalNot-only peel; `FoldBoolConstantComparison` re-exposes bare `op_True(t)` → invalid `t && x` (CS0029) | High | GPT + Gemini | ✅ `return t && x > 0;` | Fixed |
| 2 | Symmetric `op_True(t) == false` (Negate arm) → bare `op_True(t)` → invalid `t \|\| x` — proves an identity-only peel is insufficient; **both** comparison directions are hazardous | High | (my probe, from Gemini's lead) | ✅ `t \|\| x > 0;` | Fixed |
| 3 | By-ref deref nested under a **double-negating** comparison `((*r == false) == false)` bypasses the identity-only by-ref peel; outer `Negate` inverts the inner comparison and the fixpoint reduces back to bare `*r` → branchless eager deref / NRE | High | GPT | ✅ `s > 0 && r;` (branchless `ldind.u1; and`) | Fixed |
| 4 | Peeling every `LogicalNot` over-declines a valid, IL-exact `!op_True(t)` raise (non-negating arm) | Med | GPT | ✅ | **Accepted** as sound conservative over-decline (see below) |

### Fix (`8e639dee`)

Replaced the two shallow round-1 peels with **one** shared `PeelReducibleBoolWrappers` that follows the reducible operand through **any** chain of leading `LogicalNot`s and bool-constant comparisons (both `== true`/`== false`/`!= true`/`!= false`), then checks the leaf for a truthiness call / managed by-ref deref. This is a deliberate **conservative, negation-parity-agnostic over-approximation**: it also declines the branch-preserving `!x` forms (finding #4) rather than re-simulate `FoldBoolConstantComparison` + `Conditions.Negate` here — a fragile second implementation the repo's share-not-mirror rule forbids. Declining an extra rare, readable raise is sound; emitting a rebound / eager-deref one is not.

**Finding #4 (Med) — accepted, not fixed:** the exact-parity alternative GPT suggested would mirror the product's reduction algebra in the guard. The conservative peel is the sanctioned design; the lost raises (`c && !*r`, `c && !(t ? false : true)`) are rare and the loss is pure readability (a still-correct guarded return).

**Out of scope (unchanged):** a truthiness *operand* `if (c) return op_True(t); return false` still folds to invalid `c && t`. csc never emits `return op_True(t)` from a bool method (op_True results are branch-consumed, never returned), and **declining also yields invalid** `return TruthOver.op_True(t);` (CS0571) — no fold decision produces compilable output, so there is nothing to guard. Symmetric to the accepted `ShortCircuitTernaryPass` operand behavior shipped in #3107.

### Evidence (head `8e639dee`)
- Focused: `FoldGuardReturnFidelityTests` 16 (added 3 regressions: `op_True(t)==true`, `op_True(t)==false`, double-negated by-ref; flipped `*r==false` to declined), `ShortCircuitTernaryPassTests` 21, BooleanFolding/IrImporter/LoweringCoverage 85 — all green.
- Probe (deleted): G1 `op_True(t)==true` → declines to valid `if (t) {...}`; G2 `op_True(t)==false` → `if ((t ? false : true)) {...}`; nested by-ref + `*r==true` → declined guarded return.
- Fast suite: 3513 total, 55 failed — all pre-existing environmental (RTS/FidelityCheck/TypeSourceCheck/LadderRung6/NullCoalescing/CRLF); **zero** in FoldGuard/ShortCircuit/BooleanFolding.
- Corpus A/B (fresh matched-env, base 60bc3f12 vs head, 8 product assemblies / 800 methods, caps 0): **PASS** — pass bugs 0→0, conditional-branch 28→28, forward-merge 19→19 (residue 136→135 = input drift on the recompiled `ILInspector.Decompiler.dll`).

Launching round-3 fixed-head re-review of `8e639dee` with both models.

richlander:
## Round-3 fixed-head review reconciliation (head `8e639dee`)

Both models re-reviewed the round-2 fixed head `8e639dee` (the unified conservative peel).

### GPT (gpt-5.6-sol) — 1 finding, CONFIRMED and fixed
**Nested logical/bitwise operand bypasses the managed-by-ref guard (Medium, High confidence).** `IsManagedByRefDeref` peeled negation + bool-constant-comparison wrappers then checked the single leaf, so a surviving operand that was a raised logical/bitwise composition (`a && *r`) slipped through. The lift produced `c && (a && *r)`, which the printer flattens and csc branchless-collapses to `c & a & r`, eagerly dereferencing the guarded by-ref (NRE divergence). GPT gave a runtime `Add-Type /optimize` repro: `Folded(false, false, ref null)` throws `NullReferenceException` where the branched original returns `false`.

**Reachability confirmed from real csc IL** (not synthetic). I compiled `bool M(bool c, bool a, ref bool r){ if (c) return a && r; return false; }` and decompiled it: csc imports the inner `a && r` already lowered to bitwise `a & r` (both operands simple), so the real surviving operand is a **bitwise `Binary`** node — the leaf-only peel missed it and the fold emitted `return c && a & r;`, which recompiles branchless. The `&&`, `||`, and `|` variants (`M1`/`M2`/`M4`) all reproduced. After the fix all three render the faithful branched form (`if (c) { return a & r; } return false;`).

**Fix (`552710bf`):** replaced the leaf peel with a subtree scan — decline exactly when the operand *contains* a managed by-ref dereference **and** *contains no call*. That matches csc's collapse condition: it branchless-collapses `c && OP` only when `OP` is entirely side-effect-free, and a call anywhere in `OP` is a barrier it will not hoist past (so `c && SomeCall(*r)` stays a faithful, foldable raise — no over-decline). The scan subsumes every old wrapper-peel by-ref case (bare `*r`, `*r == true`, the double-negated chain) and closes the same bypass in `ShortCircuitTernaryPass` via the shared helper. My fix is broader than GPT's suggested "flatten same-kind LogicalBinary" approach, which would have missed the real bitwise-`&` shape. The truthiness peel is unchanged.

### Gemini (gemini-3.1-pro-preview) — no findings
Gemini validated the peel is sound and does not over-decline safe common shapes (`!boolLocal`, negated comparisons), with its own Roslyn IL checks confirming those keep the branch. It did not surface the nested-operand bypass GPT found.

### Evidence (fix head)
- New regressions: 3 nested-by-ref declines (logical-and/or, bitwise-and) + 1 call-barrier still-folds; FoldGuardReturnFidelityTests 16 → 20.
- Focused: FoldGuardReturn 20, ShortCircuitTernary 21, BooleanFolding 14, IrImporter 65, LoweringCoverage 6 — all green.
- Fast suite: 55 environmental reds only (0 in touched passes).
- Corpus A/B (fixed-input isolation, 8 asm / 800 methods): residue 135→135, cond-branch 28→28, forward-merge 19→19, **pass bugs 0→0** — PASS.

Round-4 fixed-head re-review of the merged head `4bab01b0` is next; I will post reconciliation and `Ready to merge` once both models are clean.

richlander:
## Round-4 reconciliation — fixed at `557626c3`

Both round-4 fixed-head reviewers (GPT `gpt-5.6-sol`, Gemini `gemini-3.1-pro-preview`, isolated worktrees off `4bab01b0`) converged on the round-3 by-ref guard `ContainsManagedByRefDereference(op) && !ContainsCall(op)` being wrong in **both** directions. I reproduced every finding with SDK-csc `/optimize` IL diffs + null-by-ref runtime probes (`ref Unsafe.NullRef<bool>()` in try/catch) and with harness `--dump` on real compiled fixtures. All confirmed and fixed.

### GPT Finding 1 — soundness MISS (was High): a call in the operand does not make it safe
`if (c) return r && Call(); return false;` → `c && r && Call()`. The **call-free `c && r` prefix** collapses branchless *before* the call barrier, so the guarded by-ref `*r` is dereferenced eagerly → NRE divergence. The `!ContainsCall` gate wrongly **folded** this. **CONFIRMED** by IL diff (Original vs Fold DIFFER) + runtime (Fold throws NRE on a null by-ref). **Fixed** — this shape now declines.

### GPT Finding 2 / Gemini Finding 1 — over-decline (was Med): compound operands are faithful
`c && (a & r)` (bitwise RHS) and `c && (r || a)` (different-kind logical RHS) are **IL byte-identical** to the guarded original and do **not** throw on a null by-ref — csc keeps the branch because the RHS is compound. Round-3's subtree scan wrongly **declined** them. **CONFIRMED** faithful; **fixed** — both now fold.

### Gemini Finding 2 — over-decline (was Med): comparison/field operands are faithful
Round-3's `Descendants` subtree scan also declined faithful `c && (r > 0)` and `c && r.b`. **CONFIRMED** faithful (IL-identical, no throw); **fixed** — both now fold.

### The fix: outer-kind-aware flatten
Replaced the subtree scan with `ShortCircuitFidelity.LiftEagerlyDerefsByRef(operand, outerKind)`. Empirically pinned rule (SDK-csc `/optimize`): csc collapses `L && R` → branchless `L & R` (and `||`→`|`) **iff R renders as a bare place** (local/arg/field/**managed-byref-deref**/constant); a **compound** R (bitwise, different-kind logical, comparison, field-of-byref, call) keeps the branch. The printer flattens only a **same-kind** logical chain, so the hazard is a by-ref deref that becomes a **bare operand of the flattened outer-kind chain**. The helper therefore threads the fold's outer `&&`/`||` kind, flattens only same-kind `LogicalBinary` nodes, and declines iff a flattened operand is a bare byref deref. The outer kind is computed per arm in `FoldGuardReturn` and is the classified `kind` in `ShortCircuitTernaryPass`.

Real-fixture harness dumps confirm: `r && Call()` and bare `r` **decline**; `a & r`, `r || a`, `r > 0`, `r.b` **fold** faithfully.

### Accepted sound over-decline
`c && !r` and `c && (r == false)` are faithful but are still **declined** by the parity-agnostic `PeelReducibleBoolWrappers` (it peels leading `!` + bool-constant comparisons in both directions). Kept conservative on purpose: a precise operand peel would re-implement `FoldBoolConstantComparison`'s reduction parity (the fragile mirror the round-2 decision explicitly avoided) and risks a soundness hole (folding a real bare `*r`). Declining a *valid* readability raise is sound; folding an *unfaithful* one is not. This over-decline is rare (0 hits in the 600-method corpus A/B) and accepted.

### Validation on `557626c3`
- FoldGuardReturnFidelityTests 22, ShortCircuitTernary 21, BooleanFolding 14, IrImporter 65, LoweringCoverage 6 — all green.
- Fast suite: only pre-existing environmental reds (native-DLL compile-back CS0009, CRLF, partial-build fixtures); zero in the touched passes.
- Corpus A/B (fixed-input isolation, 6 asm / 600 methods): NEUTRAL — residue 116→116, cond-branch 20→20, forward-merge 14→14, **pass bugs 0→0, PASS**.

Round-5 fixed-head re-reviews (both models) are running against `557626c3`.

richlander:
## Round-5 fixed-head review reconciliation (head 557626c3 → fix c8c9a45c)

Two-model adversarial re-review of the round-4 head `557626c3` (GPT `gpt-5.6-sol` + Gemini `gemini-3.1-pro-preview`, isolated worktrees, real SDK-csc `/optimize` IL + null-by-ref runtime evidence required).

**Gemini — CLEAN.** Verified outerKind matching across all four `FoldGuardReturn` arms, the by-ref collapse flagging, the reducible-wrapper peel soundness, and evidence integrity. No findings.

**GPT — 1 High + 1 Medium + 1 Low:**

### Finding 1 (High, soundness) — CONFIRMED + FIXED (`c8c9a45c`)
`LiftEagerlyDerefsByRef` flattened the surviving operand's same-kind logical chain **before** peeling reducible bool wrappers, so a same-kind chain hidden under a reducible `== true`/`!= false` wrapper — e.g. `(*r && Call()) == true` — was treated as an opaque compound operand and not descended into, missing the buried by-ref hazard (wrongly folded to a branchless eager deref → NRE divergence). The `FoldGuardReturn` fixpoint is pre-order, so the fold can see the un-reduced `== true` wrapper before `FoldBoolConstantComparison` strips it.

Fix: **peel-first-then-recurse** — peel reducible bool wrappers, then if the peeled node is a same-kind `LogicalBinary`, recurse into both children; else classify the leaf. This reaches a by-ref buried under any wrapper/chain nesting and subsumes the old `FlattenLogicalChain` iterator (removed). The change only ever declines **more** (sound conservative over-approximation); no faithful case regresses. New tests: `ByRefInLogicalAndChainUnderEqualsTrueWrapper_TailConstant_IsNotFolded` (repro) + `LocalOnlyLogicalAndChainUnderEqualsTrueWrapper_TailConstant_StillFolds` (negative).

### Finding 2 (Medium, ternary opcode-exactness) — DEFERRED to #3126 (out of scope, pre-existing)
GPT flagged `ShortCircuitTernaryPass.RendersAsBranchlessBarePlace` checking a bare local/arg only at the **root**, not flattened through a same-kind chain, so `c ? true : (a || Call())` → `c || a || Call()` can collapse the `c | a` prefix branchless (opcode divergence).

Investigated with SDK-csc `/optimize` IL: the divergence is **real but input-sensitive csc block layout**, broader than "a bare local nested in a chain" — it diverges for `a || Call()`, `a && Call()` (B-form), `Call() || (X() > 0)` but is byte-identical for `Call() || Call2()`, `Call()`, `X() > 0`, and the different-kind forms. A naive flatten-and-decline guard is both incomplete (misses the compound `Call() || (X() > 0)`) and over-declining (`Call() || Call2()`).

This is a pre-existing **opcode-fidelity** gap in `ShortCircuitTernaryPass` (added in #3107), **not a soundness bug** (the raised C# is valid + behaviorally identical), CI-green on `main` (corpus doesn't hit it), and orthogonal to #3114's by-ref soundness scope. The sibling guarded-return fold is unaffected: its decision is correctness-only and *keeps* opcode-divergent-but-valid readability raises. Filed as **#3126** with the probe evidence for a focused follow-up PR.

### Finding 3 (Low, evidence integrity) — ADDRESSED (`c8c9a45c`)
Two `FoldGuardReturnFidelityTests` comments implied a bare-operand `a && *r` source imports as a `LogicalBinary` and was verified against csc IL. csc actually pre-lowers a bare-operand `a && *r` to **bitwise** `a & r` (the sibling `ByRefUnderBitwiseAndOperand` test), so those pure same-kind `LogicalBinary` cases are **synthetic predicate tests**; the reachable logical-chain hazard needs a call barrier (`*r && Call()`). Tightened the two comments and the section header (no longer claims all shapes were verified against SDK-csc IL).

---

Round-6 fixed-head re-review of `c8c9a45c` launched (GPT + Gemini, isolated worktrees) since the Finding-1 fix is new behavior. Will reconcile and mark ready on two clean fixed-head reviews.

richlander:
## Round-6 adversarial review reconciliation (head `1ba6051a`)

Both round-6 reviewers ran against `c8c9a45c` (the pushed round-5 head). **Neither found a soundness bug** in the peel-first-recurse by-ref guard — both raised only *evidence-integrity* findings on the test/doc reachability claims. The follow-up commit `1ba6051a` is **test- and doc-comment-only (no product logic change)**.

### Reviewers
- **gpt-5.6-sol** — 1 Medium (evidence integrity), no soundness bug.
- **gemini-3.1-pro-preview** — 1 Low (evidence integrity), no soundness bug.

### Findings and resolution

**GPT (Medium) — the round-5 `== true`-wrapped tests + tightened header present synthetic IR as csc-reachable.** ✅ Confirmed and addressed.
- Verified with an SDK-csc `/optimize` probe: `(r && Call()) == true` compiles **byte-identical** to `r && Call()` (csc constant-folds `== true`, no `ceq`), and a bare-operand `r || a` lowers to a **bitwise** `or` — so the literal `LogicalBinary` / `== true` shapes are not what those sources import as.
- Fix: reframed those tests as **synthetic/defensive** predicate tests and rewrote the section header to separate csc-**reachable** operand shapes (bitwise `a & r`, comparison `*r > 0`, by-ref field `r.b`, call-barrier `*r && Call()`) from **synthetic** ones (bare-operand `LogicalBinary`, `== true`-wrapped). The genuinely reachable analog of a wrapped chain is the `!`-negated `!(*r && Call())`, which the parity-agnostic peel conservatively over-declines (consistent with the documented `!*r` over-decline).

**Gemini (a) — header names `*r > 0` and `r.b` as "verified" but the file had no tests for those shapes.** ✅ Confirmed and addressed.
- Added two positive tests backing the header claims: `ByRefInComparisonOperand_TailConstant_StillFolds` (`*r > 0`) and `ByRefStructFieldOperand_TailConstant_StillFolds` (`r.b`). Both pin **non-decline** of a by-ref confined to a compound operand — a regression guard against the round-3 subtree-scan over-decline.

**Gemini (b) — `*r && Call()` (`ByRefBeforeCallInLogicalAndOperand`) is synthetic because csc distributes `if (c) return r && Call(); return false;` into a cascade the pipeline reconstructs with `Call()` as thenValue and `c && *r` as the *condition*.** ❌ Dismissed — disproven by end-to-end evidence.
- Compiled the real fixture and ran the full pipeline via the harness `--dump`. For `bool ByRefAndCallReturn(int c, ref bool r) { if (c > 0) return r && Call(); return false; }` the final raised IR is:

  ```
  IfStatement
    Comparison.GreaterThan  (c > 0)
    Block: Return LogicalAnd( LoadIndirect bool (LoadArgument ref bool r), Call F.Call )
  Return Constant False
  ```

  i.e. condition = `c > 0`, **surviving operand = `r && Call()`** (a `LogicalAnd` with the by-ref as its left leaf). The by-ref stays in the operand; `LiftEagerlyDerefsByRef` flattens the same-kind `&&` chain, finds the bare `*r`, and **declines** — product output keeps `if (c > 0) { return r && Call(); } return false;`. The hypothesized byref-in-condition reconstruction **does not occur**, so `*r && Call()` is a genuinely reachable operand shape, not synthetic.

### End-to-end fixture confirmation (also closes the byref-in-condition question)

Real csc fixtures, harness `--dump`, all `fidelity Full`:

| Source | Product output | Decision |
| --- | --- | --- |
| `if (c>0) return r; return false;` | `if (c>0) { return r; } return false;` | **declined** (bare by-ref) |
| `if (c>0) return r && Call(); return false;` | `if (c>0) { return r && Call(); } return false;` | **declined** (by-ref in same-kind operand) |
| `if (c>0) return r && Call(); return true;` | `return c <= 0 \|\| (r && Call());` | **folded, faithful** (different-kind operand keeps its branch) |

This also confirms there is **no byref-in-condition soundness hole**: the pipeline never hoists the guarded by-ref into the emitted condition.

### Review scope for the follow-up commit

`1ba6051a` changes only test code and `///` doc comments — verified the `ShortCircuitFidelity.cs` diff has **zero** non-`///` lines. Per the adversarial-review policy this is a documentation/test-only change and does not require a fresh two-model round. Focused suites green: `FoldGuardReturnFidelityTests` 26 (24 + 2 new), `ShortCircuitTernaryPassTests` + `LoweringCoverageTests` 27.

Holding "Ready to merge" until CI on `1ba6051a` is green.

@richlander open question still standing from round-5: fix **#3126** (ternary same-kind-chain opcode divergence, pre-existing #3107 gap, orthogonal to this PR's by-ref scope) in a follow-up PR now, or just track it?

richlander:
Ready to merge

CI green on head `1ba6051a` (`changes`, `build-net10`, `csharp-diff-smoke`, `test (ubuntu-26.04, linux-x64)` all pass; `il-diff-smoke`/`markdownlint`/`pack` are path-gated skips). Six rounds of two-model adversarial review (gpt-5.6-sol + gemini-3.1-pro-preview) are reconciled; the round-6 follow-up `1ba6051a` is test- and doc-comment-only (zero product-logic change), so it needs no further review round.

Non-blocking follow-up: **#3126** (ternary same-kind-chain opcode divergence — pre-existing #3107 gap, CI-green on main, orthogonal to this PR's by-ref correctness scope).

richlander:
Superseding and closing this PR.

While this branch was in its final review rounds, a separate PR #3127 independently fixed the same #3114 issue and **merged first** (`be4291ee`), rewriting the same files (`ShortCircuitFidelity.cs`, `BooleanFoldingPass.FoldGuardReturn`, `FoldGuardReturnFidelityTests.cs`) with an opcode-fidelity policy. That left this branch conflicted against `main`.

Rather than re-litigate the merge, I compared the two designs and confirmed empirically that merged #3127 still had **one live by-ref soundness hole that only this PR closed**: #3127's `RendersAsBranchlessBarePlace` inspects only the top-level surviving operand, so a managed by-ref dereference buried in a same-kind logical chain (e.g. `if (c) return *r && Call(); return false;` → `c && *r && Call()`, whose call-free `c & *r` prefix csc collapses branchless) still folded into an unconditional dereference — a `NullReferenceException` divergence on a null by-ref.

The still-needed delta from this PR — the `LiftEagerlyDerefsByRef` chain descent (plus its reducible-bool-wrapper peel) and the six rounds of validation behind it — has been ported onto #3127's merged design as a focused, correctness-only follow-up: **#3148**. That PR keeps #3127's opcode-fidelity guards verbatim and adds only the chain-buried by-ref decline, with a real Before/After witness and two clean fixed-head adversarial reviews.

Thanks for the deep work here across all six rounds; it is preserved in #3148. Closing as superseded by #3148.

Issue #3097, opened 2026-07-23
Proposal: taste configurability model - class-3 knobs sourced from .editorconfig, plus candidate spellings


richlander:
### Context

The taste doc now frames the runtime `.editorconfig` as a **default**, not a source-style match, and `PrinterOptions` already exposes three opt-in knobs (`ReadableLocalNames`, `ExpressionBodyArrowPlacement`, `WrapSplittableExpressions`). This issue proposes the *model* for growing that surface deliberately. Companion to #3089, which documented only shipping behavior.

### Principle: the configurable surface is exactly class 3

Falls straight out of the three-class rule:

- **Class 1 (IL-exact): not configurable.** The alternative is *less* faithful. A single global "don't raise — show the lowered form" dial is fine; per-idiom downgrades are not.
- **Class 2 (fidelity-erasing): never configurable — hard invariant.** No flag may select `&`->`&&`, NaN rewrites, or any semantically-wrong output.
- **Class 3 (no anchor): the entire config surface.** Every setting is equally faithful; the round-trip fixed point holds for whichever representative is chosen.

### Three tiers within "configurable"

1. **Spelling knobs** (class-3 token choices): `var`/explicit, `this.` qualification, predefined-type keywords (`int`/`string`/`nint`), target-typed `new` on/off, optional-arg elision on/off. **Binding-hazard knobs stay sound when enabled** — the knob is "attempt," and the soundness decline is *not* user-overridable.
2. **Formatting knobs** (below class 3, layout only): arrow placement, wrap width, brace style — where `PrinterOptions` started.
3. **Synthesis** (separate axis): readable-name synthesis trades a fidelity property (Annotated-IL alignment), so it is categorized apart from the fidelity-neutral knobs.

### Config schema = the oracle artifact

Because class-3 keys are exactly editorconfig keys, source the knobs from an **actual `.editorconfig`**: default = dotnet/runtime's; a user points at their own and renders toward *their* convergence target. Same vocabulary for default and override — no bespoke flag zoo. Open questions: which keys to honor, discovery vs explicit path, precedence, per-invocation override.

### Candidate spellings to evaluate (each a future pass/printer policy)

- **Extension- vs static-call spelling** (`recv.M(a)` vs `Type.M(recv, a)`) — class-3 with a using-scope/overload hazard; fluent-chain LINQ already depends on it, so it needs a general rule.
- **Predefined-type keyword policy** (`int` vs `Int32`, `nint` vs `IntPtr`).
[...]

PR #3123, opened 2026-07-24
Add this.-qualification knobs for methods and events (#3097)

richlander:
## Summary

Slice 2b of the tool-owned decompiler style config (issue #3097). Extends the class-3 `this.`-qualification knob family with **method** and **event** qualification, mapped from their exact editorconfig keys through the existing CLI-edge `.dotnet-inspectconfig` resolver.

- `PrinterOptions.QualifyMethodAccess` ← `dotnet_style_qualification_for_method` — instance method calls (`CallText`) and method groups (`MethodGroupText`) over `this`.
- `PrinterOptions.QualifyEventAccess` ← `dotnet_style_qualification_for_event` — event `+=`/`-=` subscriptions (route through `PropertyTarget`).

Both default **off**, so default output is byte-for-byte unchanged. When set, they force `this.` on a same-type instance access that is already a `LoadArgument{this}` receiver — an IL-identical spelling with no binding hazard.

## Soundness

- A genuine non-virtual `base.M()` call is **never** rewritten to `this.M()` (that would re-enable virtual dispatch and change behavior — unbounded recursion in the fixture). The base arm short-circuits before the qualifier.
- Event subscriptions are now governed by `QualifyEventAccess` rather than `QualifyPropertyAccess`, **decoupling** the two knobs to match their separate editorconfig keys. Both share the `PropertyTarget` helper internally but read independent options. The just-shipped property knob is the only prior consumer, so no external behavior regresses.
- Both knobs are mirrored in `DecompilerResult.DecompilerOptions` and recorded via `EffectiveDecompilerOptions`, so a host can distinguish an on-render from an off-render without reverse-engineering the text.

## Evidence

- **Printer layer** (`ThisQualificationTests`, +10): method qualify / default-bare, method-group qualify / default-bare, event qualify / default-bare, `base.Value()` guard, cross-knob decoupling (property↛event, event↛method), `EffectiveOptions` records both knobs. 19/19 green.
- **Config seam** (`RenderStyleConfigTests`): key-mapping parse test (all four keys) + method/event qualification through the real `MemberCodeProvider.Collect` seam + property/event decoupling. 35/35 green.
- **CLI canary** (`.dotnet-inspectconfig` in CWD, real fixture DLL): method+event keys → `this.Compute()` / `this.Pinged += handler`; no config → bare; property-key-only → event stays bare (decoupling).
- Fast RoundTrip area: 19 `ThisQualificationTests` pass; the 31 failures are all pre-existing `ReturnToSenderFixtureCatalogTests` environmental compile-back reds (CI-green), identical to baseline.
- `docs/decompiler-taste.md` updated (two new keys + base-call caveat + event/property decoupling); markdownlint clean.

## Compatibility

Default output byte-for-byte unchanged (both knobs off gate every new branch). The only behavior change is intentional: `QualifyPropertyAccess=true` no longer qualifies events.

Part of #3097.

### DIFF
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -138,18 +138,36 @@ Anything the count cannot prove safe stays explicit.
 
 `this.field` and `field`, and `this.Prop` and `Prop`, compile to the identical
 IL — a `this`-rooted instance member load is `ldarg.0; ldfld` / `ldarg.0; call
-get_Prop` either way, so the `this.` prefix has no IL consequence. This is a
-class-3 no-anchor spelling with **no binding hazard** (unlike target-typed
-`new()` or argument elision): the prefix cannot rebind the member, it can only be
-redundant. The shipped default renders the bare name, qualifying only where the
-bare name would not bind to the member — a local/parameter shadow or a
-member/type-name collision — matching how the runtime writes member access.
+get_Prop` either way, so the `this.` prefix has no IL consequence. The same holds
+for an instance method call (`this.M()` and `M()` both emit `ldarg.0;
+call/callvirt`), a method group over `this` (`this.M` and `M` both emit `ldarg.0;
+ldftn`), and an event subscription (`this.E += h` and `E += h` both emit
+`ldarg.0; ... call add_E`). This is a class-3 no-anchor spelling with **no binding
+hazard** (unlike target-typed `new()` or argument elision): the prefix cannot
+rebind the member, it can only be redundant. The shipped default renders the bare
+name, qualifying only where the bare name would not bind to the member — a
+local/parameter shadow or a member/type-name collision — matching how the runtime
+writes member access.
 
 The always-qualified spelling is available as an opt-in, off by default so
 default output stays byte-for-byte stable:
 
 - `PrinterOptions.QualifyFieldAccess` mirrors `dotnet_style_qualification_for_field`.
 - `PrinterOptions.QualifyPropertyAccess` mirrors `dotnet_style_qualification_for_property`.
+- `PrinterOptions.QualifyMethodAccess` mirrors `dotnet_style_qualification_for_method`
+  (instance method calls and method groups over `this`).
+- `PrinterOptions.QualifyEventAccess` mirrors `dotnet_style_qualification_for_event`
+  (event `+=`/`-=` subscriptions).
+
+The four knobs are independent — each qualifies only the member kind it governs.
+Two consequences are worth calling out:
+
+- A genuine non-virtual `base.M()` call is **never** rewritten to `this.M()`, even
+  with method qualification on: the `base` call deliberately skips virtual
+  dispatch, so `this.M()` would re-enable it and change behavior.
+- Event subscriptions are governed by `QualifyEventAccess`, not
+  `QualifyPropertyAccess` (they share a printer helper internally but are decoupled
+  at the knob), matching the separate `_event`/`_property` editorconfig keys.
 
 ```csharp
 public int Compute() => _count + Extra;          // shipped default: bare
@@ -278,6 +296,8 @@ The file is flat `key = value`, using the same key and value vocabulary as an
 root = true
 dotnet_style_qualification_for_field = true
 dotnet_style_qualification_for_property = true
+dotnet_style_qualification_for_method = true
+dotnet_style_qualification_for_event = true
 ```
 
 - `#` and `;` comment lines and `[section]` headers are ignored.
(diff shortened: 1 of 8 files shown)


richlander:

Two fixed-head reviews on the exact head `d4cdbfbf` (roster: **GPT** `gpt-5.6-sol`, **MAI** `mai-code-1-flash-picker`; different families). Both isolated in separate linked review worktrees, scratch under TEMP, no writes to the trees.

### GPT (`gpt-5.6-sol`) — 3 findings, all resolved; verdict **safe to merge**

**Finding 1 (High) — `MethodGroupText` ignored `DelegateCreation.IsVirtual`.** A `base.<virtual>` method group (`ldftn Base::M`, non-virtual, cross-type) rendered bare `M` by default and `this.M` under the qualify-method knob, both recompiling to `ldvirtftn` (virtual dispatch to the derived override) — a behavior change. **Fixed** in `e8480060`: thread `d.IsVirtual` and add a base arm mirroring `CallText` (`if (method.HasThis && !isVirtual && IsCrossType(...)) return "base."+name;`). This also corrected a **pre-existing default-output bug** for base method groups. Verified: genuine base group → `base.Value` (`ldftn Base::Value`).

**Finding 2 (High) — the new base arm also fired for closed static extension groups.** `this.Extend` (where `Extend` is `static int Extend(this Derived)`) shares the `ldarg.0; ldftn` shape but is not `base.M`; it rendered invalid `base.Extend` (CS0117) and regressed a previously-valid `this.Extend`. **Fixed** in `d4cdbfbf`: gate the base arm on `method.HasThis` (the metadata `IsInstance` header), so static extensions fall through to `this.Extend`/bare. Verified: extension group → `this.Extend` under the knob / bare by default (`ldftn Extensions::Extend`), never `base.Extend`.

Both fixes carry real csc-compiled regression fixtures + tests in `ThisQualificationTests` (`BaseValueGroup`, `ExtensionGroup`); the class is **23/23 green**.

**Finding 3 (High) — DIM access via interface-cast `this` renders invalid C#.** `((IDim)this).Value` (a default interface member) rendered `this.Value`/bare `Value` (CS1061) because the class→interface conversion emits no IL. **Deferred** to #3128 as a **pre-existing, product-wide** fidelity gap: reproduced in-repo on the real test assembly, the **default (no-knob)** spelling is already invalid *before* this PR, and `CallText` renders DIM **calls** identically broken — so it is not specific to the method-group site nor created by this PR. GPT confirmed via base-vs-head CLI diff that #3123 does **not** regress the shape (`CastInterfaceGroup`/`CastInterfaceCall` default output byte-identical to base `da030adf`). The faithful fix (`((IDim)this).Value`) must land in both `CallText` and `MethodGroupText` and needs reliable interface detection that `MetadataSource.IsInterface` cannot provide cross-assembly (must decline-when-unknown) — hence a separate issue.

### MAI (`mai-code-1-flash-picker`) — 1 finding; verdict **safe to merge**

**Finding (Medium) — the `PrintRaised(out …)` / `PrintLowered(out …)` line-map overloads ignore the new knobs.** These overloads take no `PrinterOptions` parameter, so they uniformly ignore **all six** qualification knobs (not just the Slice 2b pair) and feed `CSharpBodyDiff` / `ResearchViews` (annotated/diff/research views), **not** the config-targeted "Decompiled Source" path (`MemberBodyProducer` → options-taking overloads; config attached only at the CLI edge). MAI agreed this is **pre-existing and out of scope** and found **no user-visible config-consuming path** that produces wrong styled output. Tracked as a non-blocking limitation (threading options through `Research`, a pure aggregator, + the diff subsystem is a separate change).

MAI independently reproduced the fixes on the built CLI: `this.Compute()` / `this.Pinged += handler` with the keys on, bare without config, property/event decoupling intact, and `ExtensionGroup() => new Func<int>(this.Extend)` (no `base.Extend`).

### Evidence
- `ThisQualificationTests` 23/23 · `RenderStyleConfigTests` 35/35 green.
- Fast RoundTrip area (`-trait "Area=RoundTrip" -trait- "Speed=Slow"`): only the 31 pre-existing environmental `ReturnToSenderFixtureCatalogTests` reds (CS0009 native `aspnetcorev2_inprocess.dll` pulled into Roslyn refs), **zero** new — default output corpus-stable.
- CI green on `d4cdbfbf`: `build-net10`, `test (ubuntu-26.04, linux-x64)`, `csharp-diff-smoke`, `markdownlint` all pass; `pack`/`il-diff-smoke` path-gated.

Both fixed-head reviews are clean. Deferred follow-up: #3128 (non-blocking).

richlander:
Ready to merge

richlander:
Merged `origin/main` into the branch to clear the conflicts (merge commit `8219d339`).

**Conflict resolution was mechanical.** The three conflicts (`PrinterOptions.cs`, `RenderStyleConfig.cs`, `docs/decompiler-taste.md`) were all clean add/add against main's new style lenses (`PreferConditionalExpressionReturn`, `PreferBranchlessBoolean` from #3138/#3139/#3140), which touch the same config-surface files. Resolved by keeping **both** feature sets verbatim — no logic change to either side's reviewed feature code.

**Validation on the merged head:**
- Full solution builds clean.
- Feature tests green: `ThisQualificationTests` 23/23, `RenderStyleConfigTests` 39/39 (my 35 + main's 4 lens tests), plus both lens test classes.
- markdownlint clean.
- CI green on `8219d339`: `build-net10`, `test (ubuntu-26.04, linux-x64)`, `csharp-diff-smoke`, `markdownlint`, `changes` all pass (`il-diff-smoke`/`pack` path-gated skips).

Because the resolution is purely additive/mechanical (no reviewed feature logic changed) and all validation is green, the prior two clean fixed-head adversarial reviews (GPT + MAI on `d4cdbfbf`) still hold.

Ready to merge

PR #3127, opened 2026-07-24
Guard FoldGuardReturn short-circuit folds for opcode fidelity (#3114)

### LINKED ISSUE #3114: FoldGuardReturn re-forms short-circuit returns without opcode-fidelity guards

richlander:
- Fixes #3114
- Changes `BooleanFoldingPass.FoldGuardReturn` to carry the same opcode-fidelity guards as `ShortCircuitTernaryPass`, so a guarded return with a bool-constant arm no longer re-forms a short-circuit `&&`/`||` whose recompilation diverges.
- Card revision: `3ee230ad`

## Change

**Conclusion:** **PASS** — `FoldGuardReturn` performs the same condition-lift as `ShortCircuitTernaryPass` (#3107) but lacked its three fidelity guards, so it could emit C# whose recompilation diverges in branch opcodes, operator tokens, or runtime behavior; this shares the validated predicates and applies them, declining exactly the hazardous folds while preserving every opcode-exact one.

### Original source

```csharp
public static bool NeitherOr(bool a, bool b, bool c)
{
    if (a && b)
        return false;
    return c;
}
```

### Before

Acquired with `dotnet-inspect member … NeitherOr -S "Decompiled Source"` at base `origin/main` (`da030adf`):

```csharp
public static bool NeitherOr(bool a, bool b, bool c) => !(a & b) && c;
```

- Valid: True
- Correct: True

Behavior is preserved, but this **re-forms a short-circuit that negates the `a & b` condition**, and the negated form recompiles to a different branch structure than the original `if (a && b) …` diamond — an opcode-fidelity divergence (`NeitherOr` was on the fidelity docket for exactly this reason).

### After

Acquired with the same command at this PR's head:

```csharp
public static bool NeitherOr(bool a, bool b, bool c)
{
    if (a & b)
    {
        return false;
    }
    return c;
}
```

- Valid: True
- Correct: True

The negating fold is declined, so the guarded return stays flat and recompiles opcode-exact. `NeitherOr` therefore leaves the fidelity docket.

### Fully raised

The After is in the intended state: this is a hazardous negating fold, so declining it (leaving the honest, opcode-exact `if/return`) **is** the correct endpoint — raising it would reintroduce the divergence.

## Approach

- Extract the three validated predicates into a shared `ShortCircuitFidelity` helper (the issue's preferred share-over-mirror). `ShortCircuitTernaryPass` now delegates to it (its 21 unit tests confirm the refactor is behavior-preserving).
- Apply the guards in `FoldGuardReturn` **before** it mutates the tree: decline on user-defined truthiness, on a negating fold whose negation is not opcode-exact, or on a surviving operand csc would collapse branchless.
- **Negate gate is broadened for the guarded-return form only** to `NegationIsOpcodeExact` = integer-comparison dual **or** reference null-branch. The issue's prescribed integer-only gate is incomplete here: `FoldGuardReturn` has folded a reference null-branch since before #3114 — the `String.IsNullOrEmpty` witness `value is null || value.Length == 0` — whose negation is only a branch-polarity flip and is opcode-exact. `ShortCircuitTernaryPass` keeps the narrower integer-only gate; broadening it is separate follow-up.

`String.IsNullOrEmpty` at head (regression check — the good fold is preserved):

```csharp
public static bool IsNullOrEmpty(string? value) => value is null || value.Length == 0;
```

## Evidence

- **Pass-level:** new `FoldGuardReturnFidelityTests` (19) pin accept/decline for every hazard (float dual, string `op_Equality` token flip, user truthiness, bare-place branchless collapse, by-ref NRE) and every accept (integer dual, reference null-branch, explicit null test, pointer deref). `ShortCircuitTernaryPassTests` (21) and `IrImporterTests` (65, incl. the `IsNullOrEmpty`/`List.Contains` real-artifact folds) stay green.
- **Fast decompiler suite:** 3507 tests, 0 failures.
- **Fidelity gate (real-artifact):** ran `FidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket` on base (`1be371f2`) and this branch. The current diff set is **identical except `NeitherOr`**, which base reports and this branch does not — the change removes one docketed divergence and adds **zero** new ones. `NeitherOr` is dropped from the raised docket so the gate now guards against its regression.

### Environment note

The slow gates currently also report pre-existing **environmental** failures on this machine (compile-back resolves corelib to `System.Private.CoreLib`/`7cec85d7…` while originals reference `System.Runtime`/`b03f5f7f…`), which flood the diff set with ~74 unrelated methods (lambdas, iterators, local functions, expression trees). These reproduce identically on base and are unrelated to this change.

## Compatibility / non-action boundary

- Only the negating/operator-lift cases 2 and 3 are gated. The both-opposite-constant readability raise (`return c;`/`return !c;`) has no operator lift and is intentionally left ungated.
- `ShortCircuitTernaryPass` runtime behavior is unchanged (pure delegation refactor).
- The lowered fidelity docket is left unchanged (its view does not run the same raising and was not re-measured here).

### DIFF
--- a/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
@@ -23,13 +23,15 @@ public class FidelityGateTests
     /// decompiler docket. Each is a tracked defect or a benign over-render; the gate
     /// tolerates these but fails if a NEW method joins the set. Shrink this list as
     /// fixes land. Tracked defects include StaleFieldRead (issue #605) and
-    /// benign codegen choices (BothPositive, NeitherOr).
+    /// benign codegen choices (BothPositive). NeitherOr left the docket when
+    /// #3114 taught FoldGuardReturn to decline the negated `a && b` guard fold
+    /// (`if (a && b) return false; return c;`) whose recompile diverged.
     /// GotoCommonExit is the step-2 common-exit fold: the decompiler inlines the
     /// shared return tail into each arm, recompiling to cleaner direct-return IL
     /// than the original goto-and-merge shape — a benign equivalent restructuring.
     /// SelectBoolReturn is a benign branch-polarity inversion in a bool-select
     /// ternary (orig <c>bgt</c> vs recmp <c>ble</c>), the same class as
-    /// BothPositive/NeitherOr; it was previously hidden in the recompile-fail
+    /// BothPositive; it was previously hidden in the recompile-fail
     /// bucket because its <c>System.GC.KeepAlive</c> call recompiled the short
     /// <c>GC</c> name with no <c>using System;</c> in the skeleton, and the
     /// harness now emits that using.
@@ -58,7 +60,6 @@ public class FidelityGateTests
         // hidden index temp for the same `return [a, b, 42];` source idiom.
         "CollectionListLiteral",
         "GotoCommonExit",
-        "NeitherOr",
         // This hand-written await-enumerator loop recompiles through the same
         // runtime-async shape but schedules the receiver load after the
         // enumerator-local initialization rather than before it.
--- a/src/ILInspector.Decompiler.Tests/FoldGuardReturnFidelityTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FoldGuardReturnFidelityTests.cs
@@ -1,0 +1,299 @@
+using ILInspector.Decompiler.Pipeline;
+using System.Collections.Immutable;
+
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// Guards on <c>BooleanFoldingPass.FoldGuardReturn</c> for the mixed constant-arm
+/// shape <c>if (c) return A; return B;</c> where exactly one arm is a bool constant.
+/// That fold re-forms a short-circuit <c>&amp;&amp;</c>/<c>||</c> by lifting the condition
+/// into the operator's left operand (optionally negated) and keeping the other arm
+/// as the surviving right operand — the same lift <see cref="ShortCircuitTernaryPass"/>
+/// performs for the nested constant-arm ternary — so it must carry the same
+/// opcode-fidelity guards (shared through <c>ShortCircuitFidelity</c>) or it emits C#
+/// whose recompilation diverges in branch opcodes, operator tokens, or runtime
+/// behavior (#3114). Compiler-shape evidence for the five hazards is the repro table
+/// on #3114 (`--dump` at base `e15f41c4`); the corpus compile-back gates guard the
+/// exact forms continuously. These fixtures pin the pass-level decline/accept.
+/// </summary>
+[Trait("Area", "Pass")]
+public class FoldGuardReturnFidelityTests
+{
+    static readonly TypeRef s_int = TypeRef.CoreLib("System", "Int32");
+    static readonly TypeRef s_bool = TypeRef.CoreLib("System", "Boolean");
+    static readonly TypeRef s_double = TypeRef.CoreLib("System", "Double");
+    static readonly TypeRef s_string = TypeRef.CoreLib("System", "String");
+    static readonly TypeRef s_truth = TypeRef.CoreLib("Synthetic", "TruthOver");
+    static readonly TypeRef s_holder = TypeRef.CoreLib("Synthetic", "Holder");
+
+    static Constant Bool(bool value) => new(value, s_bool);
+
+    // A computed (non-bare-load) bool operand: a comparison. csc keeps the branch for
+    // a computed operand, so raising the short-circuit operator is opcode-exact.
+    static Comparison Y()
+        => new(ComparisonKind.GreaterThan, false, new LoadLocal(2, s_int), new Constant(0, s_int));
+
+    // A bare bool argument condition (negation is not the proven integer dual).
+    static LoadArgument Flag() => new(0, "flag", s_bool);
+
+    // An integer comparison condition; its negation is the proven same-branch dual
+    // (`start <= 0` → `start > 0`, both `ble.s`).
+    static Comparison StartLessEqualZero()
+        => new(ComparisonKind.LessThanOrEqual, false, new LoadArgument(0, "start", s_int), new Constant(0, s_int));
+
+    // A float comparison condition; negating it flips the ordered/unordered sense.
+    static Comparison FloatCompare()
+        => new(ComparisonKind.LessThan, false,
+               new LoadArgument(0, "a", s_double), new LoadArgument(1, "b", s_double));
+
+    // A reference (string) equality condition; negating it flips op_Equality →
+    // op_Inequality (a different call token).
+    static Comparison StringEquals()
+        => new(ComparisonKind.Equal, false,
+               new LoadArgument(0, "a", s_string), new LoadArgument(1, "b", s_string));
+
+    // A user-defined-truthiness condition: the `operator true` call csc inserts for a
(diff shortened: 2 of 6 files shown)


richlander:

Two-model adversarial review per `AGENTS.md` (Default-pipeline correctness change). Reviewers ran in isolated worktrees at the reviewed head `3ee230ad`; I am Claude Opus and did not self-review.

### Reviewer 1 — Gemini 3.1 Pro (`gemini-3.1-pro-preview`)
**Verdict: clean, no findings.** Independently verified `NegationIsOpcodeExact`/`IsReferenceNullBranch` soundness by compiling probes with Roslyn and inspecting raw IL:
- Nullable value types (`int? == null`) → `HasValue; ldc.i4.0; ceq`, no `null` literal → `IsReferenceNullBranch` **declines** (requires `Constant(null)`).
- Unconstrained generic (`T x == null`) → `box; ldnull; ceq` → negation is exact `brtrue`↔`brfalse` polarity flip; prints `is not null`, re-lowers identically.
- User-defined `struct` `operator ==` → `Call(op_Equality)`, not a `Comparison` → **declines**.
- `isinst` (`x is int`) → polarity-exact under negation.
- Value-type generic instances can't be bare bool branches in valid IL, so the `GenericInstance`/hint acceptance can't be reached by a value type.

### Reviewer 2 — GPT-5.6 (`gpt-5.6-sol`)
**Verdict: no blocking product issues.** One **Medium/non-blocking** finding:

> **`LoweredFidelityGateTests.KnownDiffs` still lists `NeitherOr`.** The lowered `--lowered` dump now emits the faithful flat `if (a & b) return false; return c;`, so the entry is stale and would silently tolerate a future lowered-pipeline regression.

### Reconciliation
The two reviewers **disagreed** on this exact point: Gemini reasoned that keeping `NeitherOr` in the lowered docket "honors the intent (lowering does not raise the same fold)," while GPT said it is stale. **I resolved this empirically — GPT is correct.**

`IrPasses.Lowered = Default minus (ForLoopPass, IncrementDecrementPass, LockSugarPass)`, so `BooleanFoldingPass` (and my guard) **does** run in the lowered pipeline. I dumped `NeitherOr` through both pass lists; both now emit the identical faithful flat form:

```csharp
// DEFAULT (sugared) and LOWERED both:
if (a & b)
{
    return false;
}
return c;
```

Gemini's premise ("lowering does not raise the same fold") was therefore factually incorrect for this fixture. **Action taken:** removed `NeitherOr` from `LoweredFidelityGateTests.KnownDiffs` and its summary comment, mirroring the sugared `FidelityGateTests` change — commit `aab93e67`.

This follow-up is **test-only docket hardening**; the reviewed product logic (`ShortCircuitFidelity`, `BooleanFoldingPass`, `ShortCircuitTernaryPass`) is byte-identical to the head both models reviewed, so the clean product-correctness verdicts carry forward. No product re-review required for a mechanical allowlist removal proven by real run.

**Both reviews clean; the one non-blocking finding is resolved.**

richlander:
Ready to merge

richlander:
## Why decline the fold instead of re-sugaring (ternary or `&&`)?

A reasonable question: the guard falls back to the flat `if (a && b) return false; return c;` — could it instead emit a one-line ternary `(a && b) ? false : c`, or keep the `&&` short-circuit? I compiled all three renderings of the `NeitherOr` fixture and dumped the raw IL. None of the sugared forms is opcode-faithful; each diverges from the original in a *different* way.

**Original** — `if (a && b) return false; return c;`
```
ldarg.0; ldarg.1; and; brfalse.s L7;   ldc.i4.0; ret;   L7: ldarg.2; ret
```

**One-line ternary** — `(a && b) ? false : c`
```
ldarg.0; ldarg.1; and; brtrue.s  L7;   ldarg.2;  ret;   L7: ldc.i4.0; ret
```

**Bool hack (the old Before)** — `!(a && b) && c`
```
ldarg.0; ldarg.1; and; ldc.i4.0; ceq; ldarg.2; and; ret
```

- The **ternary** flips the branch to `brtrue.s` and swaps the block layout (the `false` constant moves to the branch target; `c` becomes fallthrough). The original emits `brfalse.s` — a branch-opcode-polarity + block-order divergence, exactly the class the fidelity guards exist to catch.
- The **bool hack** goes fully branchless (`(a & b) == 0` via `ceq`, then `& c`) — a larger structural change.
- Only the **flat `if/return`** reproduces the original IL exactly (`brfalse.s`, same layout), because that *is* the original source.

So for the `if (cond) return <bool-const>; return <other>;` shape there is no faithful short-circuit **or** ternary rewrite. Swapping one divergent sugar (`&&`) for another (`?:`) would just trade one fidelity diff for another. That is the design rationale for **declining the fold** and emitting the honest flat form rather than re-sugaring.

(Aside: `a && b` appears as bitwise `and` in all three because `a`/`b` are side-effect-free, so csc folds the short-circuit — which is why the printer safely spells it `a & b`.)

Issue #3009, opened 2026-07-22
Decompiler: fully raise 64-bit flags-enum long accumulator slots to the enum type


richlander:
Follow-up to #2990 (fixed in the flags-enum `BinaryResult` typing PR).

After #2990, a 64-bit-backed `[Flags]` enum OR/AND accumulation decompiles to
**valid and correct** C#, but the intermediate accumulator stack slots remain
typed `long` and are cast to the enum at each sink, e.g. from
`MySqlConnector...HandshakeResponse41Payload.CreateCapabilitiesPayload`:

```csharp
long S_0 = (long)512;
S_1 = !cs.InteractiveSession ? ((long)0) : (long)(serverCapabilities & ProtocolCapabilities.Interactive);
ProtocolCapabilities S_2 = (ProtocolCapabilities)(S_0 | S_1) | ...;
```

Fully raised, the accumulator would itself be typed `ProtocolCapabilities` and
the `(long)512` / `(ProtocolCapabilities)(...)` casts would collapse:

```csharp
ProtocolCapabilities caps = ProtocolCapabilities.Protocol41
    | (cs.InteractiveSession ? (serverCapabilities & ProtocolCapabilities.Interactive) : ProtocolCapabilities.None)
    | ...;
```

Remaining slice: propagate the enum type into the spilled `long` stack slots
(slot unifier / stack-slot type selection) when every load target is enum, so
the temporaries are declared as the enum instead of `long` with per-sink casts.

This is a quality/"fully raised" improvement only; the current output already
compiles and preserves behavior.

---

### Scope clarification (two sub-parts)

This issue covers the **IR/typing** work needed to reach the fully-raised form:

1. **Enum slot typing** — retype the spilled `long` accumulator slots to the
   enum and collapse the per-sink `(long)512` / `(ProtocolCapabilities)(...)`
   casts (as described above).
2. **Full slot inlining** — when the spilled accumulator temporaries
   (`S_0`, `S_1`, …) each have a single sink, inline them into one expression so
   the result is a single `caps = A | B | C | …;` assignment with **no**
   intermediate temporaries, e.g.:

   ```csharp
   FlagCaps64 caps = FlagCaps64.Protocol
       | (interactive ? (server & FlagCaps64.Interactive) : FlagCaps64.None)
       | (server & FlagCaps64.LoadLocal)
       | FlagCaps64.Secure
       | (server & FlagCaps64.MultiStatements)
       | FlagCaps64.MultiResults;
[...]

Issue #2386, opened 2026-07-05
F2 spec: inline single-use materialized slot locals at their consumer (thin-writer, pre-C2)


richlander:
Architect spec for the F2 slice (thin-writer trajectory: F1 #2295 → **F2** → C2 unifier deletion #2209). Written for an implementing agent; acceptance criteria at the end. Parent: #2095. Related: #2376 (precedence, landed), #2379 (piece 1), #2383 (def-use webs), #2338 (quality backlog).

---

## Census update (2026-07-05) — supersedes the points below where noted

Ran the Stage-0 census experiment (results: [comment](https://github.com/richlander/dotnet-inspect/issues/2386#issuecomment-4887873952)). Corpus: pinned 14 assemblies / 89,065 methods, merge-base `afef0554`. Findings that revise this spec:

- **F2's home is `ExpressionInliningPass`, not a greenfield pass.** That pass already forward-substitutes single-use stack-slot temporaries with the gate this spec describes (single store/load, not address-taken, `IsConfinedToBlock` = same-block, a forward interference scan, `IsInsideCatchFilter`). Its `InlineLiveRangeOnce` mode is already slot-only. **v1 = add a `slotsOnly` gate to its `InlineOnce` loop and run one late instance**, not a new IR web builder. Building a third slot-def-use structure here is the anti-pattern #2379 piece 1 exists to end; the web build defers to C2 (#2209) where it is load-bearing.
- **Placement anchor is `SlotMaterializationPass` (`IrPass.cs:333`), not merely `CoercionInsertionPass`.** The residual class exists because `ExpressionInliningPass` last runs at `:184`, then ~15 structuring/reconstruction passes mint fresh single-use spill slots never offered to inlining. The late run must precede materialization or the slot is already a `T S_n = expr` local.
- **v1 scope is stack-slots-only — measured non-negotiable, not just a preference.** A blanket late re-run (slots + locals) inlines constants into locals that `IncrementDecrementPass` has since reconstructed into `x++`, producing `1++` (CS0131) in 12–14 methods (the reconstructed `++` hides a store from the single-use gate). The slot-restricted run produces **0 regressions**. This is condition 5's ordering inversion made concrete.
[...]

PR #3133, opened 2026-07-24
Inline fully-raised side-effect-free accumulator temporaries (#3009 sub-part 2)

richlander:
- Advances #3009 (sub-part 2 of 3)
- Changes: fully-raised, side-effect-free accumulator temporaries now inline into their single use instead of spilling to a synthetic slot local.
- Card revision: `39b5797b`

## Change

**Conclusion:** **PASS** — extending `ExpressionInliningPass.IsPure` to recurse over effect-free, non-throwing composites (conditionals, comparisons, bitwise/shift/unchecked arithmetic, neg/not, `&&`/`||`) lets the last spilled accumulator temporary inline; division/remainder and checked arithmetic/conversions stay excluded because a value that can throw must not be deferred past a prefix that can also throw.

### Original source

```csharp
public static int Accumulate(FlagCaps64 server, bool interactive)
{
    FlagCaps64 caps = FlagCaps64.Protocol
        | (interactive ? (server & FlagCaps64.Interactive) : FlagCaps64.None)
        | (server & FlagCaps64.LoadLocal)
        | FlagCaps64.Secure
        | (server & FlagCaps64.MultiStatements)
        | FlagCaps64.MultiResults;
    return (int)caps;
}
```

### Before

```csharp
public static int Accumulate(ILInspector.Decompiler.Tests.FlagCaps64 server, bool interactive)
{
    FlagCaps64 S_1 = interactive ? (server & FlagCaps64.Interactive) : FlagCaps64.None;
    return (int)(FlagCaps64.Protocol | S_1 | server & FlagCaps64.LoadLocal | FlagCaps64.Secure | server & FlagCaps64.MultiStatements | FlagCaps64.MultiResults);
}
```

- Valid: True
- Correct: True

The fully-raised enum accumulator (sub-part 1) still spills the ternary to a synthetic `S_1` slot local because `IsPure` classified `Conditional` as impure, forcing the slot to materialize.

### After

```csharp
public static int Accumulate(ILInspector.Decompiler.Tests.FlagCaps64 server, bool interactive) => (int)(FlagCaps64.Protocol | (interactive ? (server & FlagCaps64.Interactive) : FlagCaps64.None) | server & FlagCaps64.LoadLocal | FlagCaps64.Secure | server & FlagCaps64.MultiStatements | FlagCaps64.MultiResults);
```

- Valid: True
- Correct: True

The `S_1` temporary (a pure ternary) inlines into its single use in the OR chain, collapsing the method to one fully-raised expression. The remaining gap to *Fully raised* is only sub-part 3 (multi-line OR-chain wrapping), tracked in #3009.

### Fully raised

```csharp
public static int Accumulate(FlagCaps64 server, bool interactive)
    => (int)(FlagCaps64.Protocol
        | (interactive ? (server & FlagCaps64.Interactive) : FlagCaps64.None)
        | (server & FlagCaps64.LoadLocal)
        | FlagCaps64.Secure
        | (server & FlagCaps64.MultiStatements)
        | FlagCaps64.MultiResults);
```

## Why it is sound

`IsPure` gates a transform that *defers* a value's evaluation to its single load site, moving it past whatever the consuming statement evaluates first. "Pure" here therefore means **no observable effect AND cannot throw**:

- A pure value neither observes nor produces effects, so reordering it past any prefix is invisible.
- If it could throw, deferring it past a prefix that can also throw (or has an effect) would change which exception surfaces — so `Divide`/`Remainder` (DivideByZero/Overflow), checked arithmetic, and checked conversions are excluded. `Convert` is excluded entirely (conservative).
- Leaf reads keep the existing guards: address-taken locals, escaped/`ref`/`out` argument addresses, and the struct-`this` receiver are never pure.

Recompilation is fidelity-neutral or positive: inlining removes a spurious `stloc`/`ldloc` pair and returns the value to the original stack shape the source spilled.

## Evidence

| Check | Command | Baseline | Head |
| --- | --- | --- | --- |
| Fast decompiler suite | `-trait- "Speed=Slow"` | 3372 pass | 3372 pass |
| Validity | `-trait "Area=Validity"` | 103 pass | 103 pass |
| Corpus (CoreLib sweep, floors) | `-trait "Area=Corpus"` | 61 pass | 61 pass |
| Fidelity (compile-back) | `-trait "Area=Fidelity"` | pending | pending — local gate running; CI `test (ubuntu)` is authoritative (local compile-back is env-flaky) |

Three focused tests (`SelfQualifyPrinterTests.SyntheticLocalShadowedStaticCall_StaysQualified`, `IrImporterTests.BooleanMaterialization_IndirectBoolStore_DeclaresBoolSlot`, `IrImporterTests.ReferenceConditionalStore_UsesConsumerSlotType`) newly inlined their single-use slot; each fixture was adjusted to keep the slot multi-use so it still pins its slot-typing/self-qualification behavior. New regression test `FlagsEnumAccumulator_InlinesTernarySlot_NoSlotLocalSurvives`.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -269,6 +269,17 @@ public class CfgSampleClass
     // bool op — neither operand is an integer, so no `? 1 : 0` materialization.
     public static bool AndTwoBools(bool a, bool b) => a & b;
 
+    static int IncrementReorderSink(int a, bool b) => b ? 1 : 2;
+
+    // Named arguments evaluate in source order (`b: x > 0` first, then `a: x++`),
+    // so csc spills the comparison across the post-increment before the call.
+    // The comparison `x > 0` is a pure composite ExpressionInliningPass may now
+    // defer (issue #3009), but deferring it past the reconstructed `x++` would
+    // read the *incremented* value and change the result. The pass tracks the
+    // increment target as a mutation, so `x > 0` stays put and still evaluates
+    // before `x++` (issue #3133 adversarial review).
+    public static int IncrementReorderGuard(int x) => IncrementReorderSink(b: x > 0, a: x++);
+
     // Adversarial: the bool operand mixes with a `uint` sibling. The materialized
     // `(b ? 1 : 0)` is a signed int, so `int | uint` would be CS0266 (widens to
     // long) unless the mixed-sign reconciliation casts it: `(uint)(... ? 1 : 0) | c`.
--- a/src/ILInspector.Decompiler.Tests/ExpressionInliningPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ExpressionInliningPassTests.cs
@@ -172,6 +172,91 @@ public class ExpressionInliningPassTests
         function.CheckInvariant();
     }
 
+    // The interference check must recognize compound assignments, not just plain
+    // stores and increments. A `??=` (`NullCoalescingAssignment`) reconstructed
+    // before the late inlining run mutates its local; a pure composite reading
+    // that local must not be deferred past it. Here the slot holds `x + 1` and
+    // the for-loop initializer is `x ??= 5`, whose write sits before the slot's
+    // load in the condition (issue #3133 adversarial review — GPT/Gemini `??=`).
+    [Fact]
+    public void PureCompositeReadingLocal_IsNotInlinedPastNullCoalescingAssignment()
+    {
+        var body = new BlockContainer();
+        var block = new Block(0);
+        block.Add(new StoreStackSlot(0, new Binary(
+            BinaryKind.Add,
+            isChecked: false,
+            isUnsigned: false,
+            new LoadLocal(0, Int32),
+            new Constant(1, Int32))));
+        block.Add(new ForLoop(
+            new NullCoalescingAssignment(0, Int32, new Constant(5, Int32)),
+            new Comparison(
+                ComparisonKind.LessThan,
+                isUnsigned: false,
+                new LoadStackSlot(0, Int32),
+                new Constant(10, Int32)),
+            new IncrementDecrement(new LoadLocal(1, Int32), isIncrement: true, isPrefix: false),
+            new Block(1)));
+        body.Add(block);
+        var function = new IrFunction(
+            "M",
+            Holder,
+            new MethodSignature(Void, [], HasThis: false, GenericParameterCount: 0),
+            [Int32, Int32],
+            body);
+
+        new ExpressionInliningPass().Run(function, PassContext.None);
+
+        Assert.Contains(block.Children.OfType<StoreStackSlot>(), store => store.Slot == 0);
+        Assert.Contains(function.Descendants.OfType<LoadStackSlot>(), load => load.Slot == 0);
+        function.CheckInvariant();
+    }
+
+    // Same hazard via a tuple deconstruction that reassigns an existing local
+    // (`IsDeclared: false`). The slot holds `x + 1`; the for-loop initializer
+    // deconstructs into `x`, so the deferred composite would read the mutated
+    // value. `DeconstructionAssignment` local targets must count as writes
+    // (issue #3133 adversarial review — Gemini deconstruction case).
+    [Fact]
+    public void PureCompositeReadingLocal_IsNotInlinedPastDeconstructionAssignment()
+    {
+        var body = new BlockContainer();
+        var block = new Block(0);
+        block.Add(new StoreStackSlot(0, new Binary(
+            BinaryKind.Add,
+            isChecked: false,
+            isUnsigned: false,
+            new LoadLocal(0, Int32),
+            new Constant(1, Int32))));
+        block.Add(new ForLoop(
+            new DeconstructionAssignment(
+                [0],
+                [Int32],
+                new Constant(0, Int32),
+                [false]),
+            new Comparison(
+                ComparisonKind.LessThan,
+                isUnsigned: false,
+                new LoadStackSlot(0, Int32),
+                new Constant(10, Int32)),
+            new IncrementDecrement(new LoadLocal(1, Int32), isIncrement: true, isPrefix: false),
+            new Block(1)));
+        body.Add(block);
+        var function = new IrFunction(
+            "M",
+            Holder,
+            new MethodSignature(Void, [], HasThis: false, GenericParameterCount: 0),
+            [Int32, Int32],
+            body);
+
+        new ExpressionInliningPass().Run(function, PassContext.None);
+
+        Assert.Contains(block.Children.OfType<StoreStackSlot>(), store => store.Slot == 0);
+        Assert.Contains(function.Descendants.OfType<LoadStackSlot>(), load => load.Slot == 0);
+        function.CheckInvariant();
+    }
+
     [Fact]
     public void CachedDelegateArgument_InlinesToSingleCall()
     {
--- a/src/ILInspector.Decompiler.Tests/ExpressionInliningPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ExpressionInliningPassTests.cs
@@ -131,6 +131,47 @@ public class ExpressionInliningPassTests
         function.CheckInvariant();
     }
 
+    // Counterpart to the reorder guards: the interference check is precise, not a
+    // blanket ban. A slot holds `x > 0`; the increment `x++` happens strictly
+    // AFTER the slot's single load, so deferring the comparison into the call is
+    // safe. The value must still inline — a coarse function-wide mutation scan
+    // would drop it (issue #3133 adversarial review — Gemini over-blocking case).
+    [Fact]
+    public void PureCompositeReadingArgument_InlinesWhenIncrementFollowsTheLoad()
+    {
+        var use = new MethodRef(Holder, "Use", Void, [Int32, Bool], HasThis: false);
+
+        var body = new BlockContainer();
+        var block = new Block(0);
+        block.Add(new StoreStackSlot(0, new Comparison(
+            ComparisonKind.GreaterThan,
+            isUnsigned: false,
+            new LoadArgument(0, "x", Int32),
+            new Constant(0, Int32))));
+        block.Add(new ExpressionStatement(new Call(
+            use,
+            isVirtual: false,
+            [
+                new LoadArgument(1, "y", Int32),
+                new LoadStackSlot(0, Bool),
+            ])));
+        block.Add(new ExpressionStatement(
+            new IncrementDecrement(new LoadArgument(0, "x", Int32), isIncrement: true, isPrefix: false)));
+        body.Add(block);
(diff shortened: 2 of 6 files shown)


richlander:
## Round 1 adversarial review — findings & resolution (interim)

Two fixed-head reviews ran at `39b5797b` (Gemini `gemini-3.1-pro-preview`, GPT `gpt-5.6-sol`). Both raised a finding; here is the reconciliation and the follow-up commit `95b13b92`.

### Gemini — reorder past `x++`/`x--` (VALID; fixed)
`IsPure` was broadened to defer side-effect-free composites, but the pass's mutation tracking omitted reconstructed `IncrementDecrement`: `IncrementDecrementPass` folds the `starg`/`stloc` away before the late slots-only inline run, so a pure composite reading an incremented place could be deferred past the increment and read the mutated value.

- The reported compiled repro (`M(b: x > 0, a: x++)`) does **not** miscompile at head — csc spills the reordered argument to a *local* that the early full run already protects by pass ordering (`bool V_0 = x > 0; return M(x++, V_0);`). But the identified tracking gap is real for the **late slots-only** path.
- **Fix (`95b13b92`):** record an `IncrementDecrement` target as a mutation in both `InlineOnce`'s scan and the `InlineLiveRangeOnce` `Writes` interference check, so any value reading that place is a reorder barrier. Correct-by-construction (mirrors the existing `#2386` unreachable-but-guarded precedent).
- **Evidence:** new direct-IR test `SlotCompositeReadingIncrementedArgument_IsNotInlinedPastIncrement` **fails without** the fix and passes with it; compiled fixture `CfgSampleClass.IncrementReorderGuard` pins source evaluation order. Fast suite green (3374/3374).

### GPT — `Conditional` inlining rebinds an overloaded call (pre-existing; filed #3135)
GPT's exact repro uses a **named** `object value = c ? "x" : null;` local; it does **not** reproduce — the existing `StoreLocal` type-witness guard keeps the `object` local and binds `P(int, object)` correctly (verified at head `95b13b92`).

The **slot** variant (`P(Side(), (object)(c ? "x" : null))`) does drop the `(object)` witness, but this reproduces at the PR base `68ab5ae1` (and merge-base) too: it is a **pre-existing** slot/conditional *typing+printing* bug, not introduced by the inlining broadening. The shape changes (slot vs inlined) but the binding is already wrong at base. Filed as **#3135** with base-reproduction evidence; out of scope for this PR (corpus 61/61 and fidelity unchanged).

### Re-review
Head changed (`39b5797b` → `95b13b92`), so two fresh fixed-head reviews are running at `95b13b92`. Will confirm both clean before marking ready.

richlander:
## Re-review reconciliation @ `bb18c52f` (round 2 findings)

Both re-reviews at `95b13b92` returned; all findings are addressed by `3bb71836` + `bb18c52f`. Attribution and resolution:

### GPT (`gpt-5.6-sol`) — Finding 1 (High/High): composite reads move past nested local writes
**Valid, head-only.** A `Binary(x + 1)` temp inlined into a `for (x = 10; t < 5; …)` condition reads the overwritten `x`, because the `ForLoop` initializer's `StoreLocal` runs before the condition and my prior increment-only tracking only caught the `x++`/`x--` flavor. Root cause: **purity ≠ read-stability under deferral.**

**Fix (`3bb71836`):** Replaced the coarse function-wide increment scan-tracking with a precise interference check in `InlineOnce`. A pure value reads only arguments and locals (`IsPure` admits nothing else); when relying on purity to defer to a non-first-leaf load, block if `next` writes any place the value reads. Conservative over the whole statement — over-blocking only keeps a temp, never miscompiles. Discriminating direct-IR test `PureCompositeReadingReassignedLocal_IsNotInlinedPastForLoopInitializer` (verified FAILS without the guard).

### GPT — Finding 2 (Medium/High): live-range tracking misses stack-slot increment targets
**Valid, correct-by-construction.** `Writes` recognized `IncrementDecrement` only for locals/args. **Fix (`3bb71836`):** added `IncrementDecrement { Target: LoadStackSlot }` to the `Slot` case. (Real IL never spells a slot increment target, per the pass doc, but closed for soundness.)

### Gemini (`gemini-3.1-pro-preview`) — Finding 1 (Low, correctness-safe): global mutation tracking over-blocked legitimate inlines
**Valid against the *prior* approach; resolved by the same fix.** The removed function-wide scan-tracking dropped a legitimate inline when the increment happened strictly *after* the load. The new `next`-only interference check is precise, so that inline now proceeds. Pinned by regression test `PureCompositeReadingArgument_InlinesWhenIncrementFollowsTheLoad` (`bb18c52f`).

Gemini additionally verified purity soundness (Divide/Remainder hard-excluded; checked ops excluded via `IsChecked: false`; `Convert` omitted so reference/primitive witnesses are not erased here — #3135 tracks the pre-existing implicit-reference case) — **no soundness bug found.**

### Validation @ head
- Fast suite 3375/3375; Validity 103/103; Corpus 61/61.
- Both discriminating reorder tests FAIL without the guard; the over-block test FAILS with the old global tracking.

Head changed → launching a fresh fixed-head review round at `bb18c52f`. Not marking ready until both are clean and CI is green.

richlander:
## Round-3 reconciliation @ `14a5fd84`

Both round-3 fixed-head reviews **converged on one High finding** — the interference check's `Writes` helper missed compound assignments. Fixed in `14a5fd84`.

### GPT (`gpt-5.6-sol`) + Gemini (`gemini-3.1-pro-preview`) — High/High: `Writes` misses `NullCoalescingAssignment` and `DeconstructionAssignment`
**Valid.** `Writes` matched only `StoreLocal`/`StoreArgument`/`StoreStackSlot`/`IncrementDecrement`. But `NullCoalescingAssignmentPass` (writes a local) and `DeconstructionAssignmentPass` (writes locals + arguments) both run **before** the late `ExpressionInliningPass`, so a pure composite reading `x` could be deferred past a prefix `x ??= 5` or `(x, _) = …` and read the mutated value. GPT's repro: `S0 = (V0 == null); if (c) { V0 ??= "set"; Use(S0); }` → `Use(V0 == null)`. Gemini notes the deconstruction flavor is **pre-existing** for pure scalar reads at base; the purity broadening expands it to all admitted composites — fixing `Writes` closes both (it also backs `InlineLiveRangeOnce`'s scan).

**Fix (`14a5fd84`):** `Writes` now recognizes `NullCoalescingAssignment` (local) and `DeconstructionAssignment` local/argument targets. Two discriminating direct-IR tests (`…IsNotInlinedPastNullCoalescingAssignment`, `…IsNotInlinedPastDeconstructionAssignment`), both verified FAIL without the fix.

### Attack points verified safe (Gemini)
1. **Read-type coverage** — `IsPure` admits only Constant/SizeOf/LoadToken/LoadArgument/LoadLocal leaves; no slot/field/element/indirect reads escape the check.
2. **Interference window** — `next` is exactly `store.ChildIndex+1` (or the fallthrough first statement) with `IsInside(load, next)`; it exactly bounds the deferral gap.
4. **Removed scan-tracking** — correctly subsumed by the `next`-scoped check without the old over-blocking.
5. **`IsFirstEvaluatedLeaf`** — never returns true for a load preceded by another operation's effects.

### Validation @ `14a5fd84`
- fast 3378/3378; Validity 103/103; Corpus 61/61. All four discriminating reorder tests FAIL without their guard.

Head changed → verifying the fix at `14a5fd84` with both reviewers. Not marking ready until both confirm clean and CI is green.

richlander:
## Round-4 reconciliation (head `57f6c16f`)

Both reviewers independently found the round-3 `Writes` fix (`??=` + deconstruction) was **still incomplete** for other local/argument binders reachable at the slot-only `ExpressionInliningPass` run (`IrPasses.Default` line ~396, after the structuring passes):

- **GPT (High):** `CatchClause.VariableIndex` — a catch binding is a local write produced by `EhStructuringPass`, which runs before *every* EIP run.
- **Gemini (High):** foreach/using/fixed headers and pattern-match bindings (`ForeachStatement`, `UsingStatement`, `Fixed`, `IsPattern`, the switch-expression arms and their property subpattern).

Both verified their round-3 repro (`??=`/deconstruction) now blocks; both confirmed the fix so far was **strictly limited** to the two round-3 nodes, leaving the rest exposed. Both attack points reproduce at base too where the read is pure — head is where the purity broadening makes them reachable via a non-first-leaf deferral.

### Fix (`57f6c16f`) — complete by construction, not by enumeration

Four rounds of missing-writer findings showed enumeration is fragile, so the detection is now bounded by two structural facts:

1. **Escaped addresses cover all indirect writes.** Every `ref`/`out` call/ctor argument, `InitObject`, or store-through-address is mediated by `LoadLocalAddress`/`LoadArgumentAddress` (see `DefiniteAssignment.IsVerifiedOutLocal`). `Writes` now treats an escaped address of the read place as a conflict — covering every callee shape without enumerating them.
2. **Direct binders all name their place** through a `LocalIndex`/`VariableIndex`. `Writes` enumerates each (catch, fixed, using, foreach, `??=`, deconstruction, is-pattern, recursive-property pattern, both switch-expression arms + property subpattern).

Detection stays **index-precise** (no over-block): Corpus 61/61 and Validity 103/103 are unchanged, and the flags-enum collapse is preserved.

New `Writes_CoversEveryLocalOrArgumentBindingNode` is a reflection pin over every public IR node carrying a local/argument binding index; a future binder that isn't handled in `Writes` fails this test — turning the next omission from a silent miscompile into a red test. Added discriminating direct-IR tests for the catch and foreach cases (both fail without their writer arm).

**Local:** fast 3381/3381, Validity 103/103, Corpus 61/61.

Re-verification requested from both reviewers at `57f6c16f`. Not marking ready until both fixed-head reviews are clean and CI is green.

richlander:
## Round-5 verification — both reviews CLEAN at `57f6c16f`

Both reviewers re-verified the completeness fix on the fixed head and found **no remaining high-confidence bug**:

- **GPT (`gpt-5.6-sol`):** "No remaining high-confidence bug found; the fix appears complete." Rebuilt product; CatchClause direct-IR repro passes (slot survives); binder-coverage pin passes; all 33 `ExpressionInliningPass` tests pass. Confirmed address-mediated writes stay visible through address nodes (or make the source impure), and no persisted binder escapes the reflection filter.
- **Gemini (`gemini-3.1-pro-preview`):** "The fix is COMPLETE." Re-ran `??=`/deconstruction/foreach probes (all slots survive); verified the escaped-address argument is airtight (all indirect `ref`/`out`/`*p` mutation is mediated by `LoadLocalAddress`/`LoadArgumentAddress`); confirmed the non-`IrNode` `PatternSwitchExpressionArm.Subpattern` payload is reached via `pa.Subpattern?.LocalIndex`; confirmed `DeconstructionTarget` is the only argument-index binder; confirmed ref-args are rejected by `IsPure`.

## CI green at `57f6c16f`

- `build-net10` ✅  `csharp-diff-smoke` ✅  `test (ubuntu-26.04, linux-x64)` ✅ (6m31s, includes the fidelity gate)  `changes` ✅
- `il-diff-smoke` / `markdownlint` / `pack` are path-gated (skipping).

Two clean fixed-head adversarial reviews from two non-Opus families + green CI.

richlander:
Ready to merge

Issue #3138, opened 2026-07-24
Design: opt-in style lenses (behavior-faithful, byte-divergent) — 'prefer conditional expression' as the first knob


richlander:
## Summary

Introduce a third output tier — **opt-in style lenses** — that are **100% behavior-faithful but may diverge in bytes**, distinct from the two byte-faithful tiers we ship today (default sugared view and lowered view). The first candidate knob is **"prefer conditional expression"** (a ternary/`?:` rendering), which surfaces exactly where the default view must *decline* a fold for opcode fidelity.

Motivating case: #3114 / #3127. `BooleanFoldingPass.FoldGuardReturn` now declines the short-circuit fold for the `if (cond) return <bool-const>; return <other>;` shape because no sugar is opcode-faithful (see analysis below). A user who wants runtime-`.editorconfig`-shaped output would still like to *see* the ternary — that is a legitimate but separate request with a different correctness contract.

## The three tiers, three contracts

| Tier | Contract | Gate | Home |
| --- | --- | --- | --- |
| Default (sugared) | **Byte-faithful** — recompiles to the same IL | Compile-back fidelity gate | shipped `Print` |
| Lowered | **Byte-faithful**, de-sugared | Lowered fidelity gate (own docket) | `PrintLowered` |
| **Style lens (new)** | **Behavior-faithful**, byte-divergent, oracle-maximizing | **New behavioral-equivalence gate** | opt-in `PrinterOptions` knob |

The first two tiers share the compile-back (opcode-equality) gate. The third **cannot** use it — a byte-divergent lens fails compile-back by construction — so it needs its own validation.

## Why this is a genuinely new contract

Every existing `PrinterOptions` knob is **byte-preserving** and documented as such:
- `WrapSplittableExpressions` — "the IL is unchanged" (whitespace-only).
- `QualifyFieldAccess` / `QualifyPropertyAccess` — "IL-identical".
- `ReadableLocalNames`, `ExpressionBodyArrowPlacement` — no IL consequence.

A "prefer conditional expression" knob would be the **first `PrinterOptions` knob that changes the emitted bytes** while preserving observable behavior. That is why it needs new infrastructure rather than just another `bool` on the record.

## Relationship to the taste model (`docs/decompiler-taste.md`)

The taste doc's three-class rule already anticipates opt-in exposure for no-anchor choices ("Where a no-anchor choice is not yet a stable default it is exposed opt-in"). But the ternary case is subtler than a class-3 no-anchor pick:

[...]

PR #3139, opened 2026-07-24
Add PreferConditionalExpressionReturn style lens (#3138)

richlander:
## What

First opt-in **style lens** (#3138): a new tier of render option that is **behavior-faithful but byte-divergent**, distinct from the existing byte-preserving class-3 spelling knobs.

The lens `PrinterOptions.PreferConditionalExpressionReturn` (config key `dotnet_style_prefer_conditional_expression_over_return`, IDE0046) re-renders a guarded boolean return the **default** view must leave flat — because no short-circuit fold of that shape is opcode-faithful (#3114) — as the conditional expression:

```csharp
// shipped default (byte-faithful):
if (a & b) { return false; }
return c;
// with the lens (behavior-faithful, byte-divergent):
return a & b ? false : c;
```

## Why it's safe

The ternary is the **canonical desugaring** of the guarded return — same condition, same arms, same short-circuit evaluation order — so the rewrite is **unconditionally behavior-preserving**. That is exactly what lets the lens re-offer a fold #3114 had to decline for opcode fidelity. It is **not** opcode-faithful, so:

- Off by default; the default/lowered byte-faithful paths are untouched.
- Runs only on the opt-in raised path, after the default pipeline.
- The IL-anchored **Annotated** view never applies it (stays line/IL aligned).
- Its output must never feed the compile-back fidelity gates (it doesn't — the gates only run Default/Lowered).

Deliberately stops at IDE0046: the further `c ? true : d` → `c || d` collapse (IDE0075 / "bool hack") is a separate future knob, so a literal-arm ternary like `a ? true : b` is kept as-is.

## Changes

- **`PrinterOptions.PreferConditionalExpressionReturn`** — new byte-divergent knob (documented as such).
- **`PreferConditionalReturnPass`** — matches the flat bool guard-return surviving the default pipeline (mirrors `FoldGuardReturn`'s match; reuses its `ConsumesBranchTarget` guard) and rewrites to `return c ? A : B;`.
- **`CSharpPrinter.PrintRaised`** — runs the pass only when the knob is set, after `IrPasses.Default`, reusing the existing `RaiseContext`. No `PassContext`/`RaiseContext` signature churn.
- **`RenderStyleConfig`** — maps the new `.dotnet-inspectconfig` key.
- **`docs/decompiler-taste.md`** — new "Style lenses" section + config-key/example updates.

## Evidence

- New focused tests (`PreferConditionalReturnLensTests`): default-flat vs lens-ternary structural checks, IDE0075-not-applied, a no-op negative, and an **executed truth-table behavioral-equivalence gate** over all boolean inputs.
- New `RenderStyleConfigTests` coverage for the config key.
- Fast decompiler suite: **3557 / 0 failures**; default output byte-unchanged.
- CLI end-to-end verified: with the config key set, `member ... -S "Decompiled Source"` renders `=> a & b ? false : c;`; without it, the flat guard.

## Review

New heuristic/behavior → needs 2-model adversarial review (Gemini + GPT) per AGENTS.md before merge.

### DIFF
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -251,6 +251,41 @@ and declines otherwise:
   fidelity concern. The from-end open form (`s[^n..]`), which the compiler spills
   distinctly, is recovered.
 
+## Style lenses (behavior-faithful, byte-divergent)
+
+The knobs above are all **byte-preserving**: every one selects a spelling that
+recompiles to the identical IL (a class-3 no-anchor choice) or changes only
+whitespace, so the default and the knob-on output are members of the same
+compile-back equivalence class. A **style lens** is a different, explicitly
+opt-in contract (#3138): it trades byte fidelity for a source-style preference
+the oracle endorses but the fidelity-first default cannot take. A lens is
+**behavior-preserving** — its output computes the identical result for every
+input — but **not** opcode-faithful, so its output must never feed the
+compile-back fidelity gates, and it is off by default.
+
+The first lens is `PrinterOptions.PreferConditionalExpressionReturn`
+(`dotnet_style_prefer_conditional_expression_over_return`, IDE0046). A guarded
+boolean return the default leaves flat because no short-circuit fold of it is
+opcode-faithful (see the short-circuit fidelity guard, #3114) is re-rendered as
+the conditional expression:
+
+```csharp
+// shipped default (byte-faithful — the flat guard is what recompiles exactly):
+if (a & b) { return false; }
+return c;
+// with PreferConditionalExpressionReturn (behavior-faithful, byte-divergent):
+return a & b ? false : c;
+```
+
+The ternary is the *canonical* desugaring of the guard — same condition, same
+arms, same evaluation order — so the rewrite is unconditionally
+behavior-preserving; that is what lets the lens re-offer a fold the default had
+to decline. It deliberately stops at IDE0046: the further `c ? true : d` → `c ||
+d` collapse (IDE0075) is a separate future knob, so a literal-arm ternary such as
+`a ? true : b` is kept as written rather than simplified. The lens runs only on
+the opt-in raised path, after the default pipeline, and the IL-anchored Annotated
+view never applies it (it must stay byte-faithful for line/IL alignment).
+
 ## Names
 
 Without a PDB, locals are slot names (`V_0`, `S_0`) shared with the Annotated IL view — the two views stay name-aligned by construction. With a PDB, source names are used. Synthesizing readable names (`size`, `array`, `item`) where no PDB exists is an open design question: it is the largest remaining cosmetic gap against source, but it would break view alignment unless opt-in.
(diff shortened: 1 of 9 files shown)


richlander:

Two-model review at base `be4291ee` / head `bd53971b` (Claude Opus is the author, so it did not self-review):

- **Gemini 3.1 Pro — clean.** Verified behavior-equivalence (identical evaluation order; siblings share the exact block/EH context so no clause boundary intrudes), `ConsumesBranchTarget` anchor/argument correctness, `System.Boolean` type match excluding `bool?`/`ref bool`, gate isolation (pass sequestered to the opt-in `PrintRaised` overload; Annotated/`out statementLines` and `PrintLowered` omit it), fixpoint termination (each rewrite strictly removes one `IfStatement`), and printer precedence. Ran the full `Speed=Slow` suite (3557 passing).

- **GPT-5.6 — one Medium finding (valid, fixed).** For `if (c) return value; return false;` the lens builds `Conditional(c, value, false)`, which the printer idiomatically spells as the short-circuit `c && value` (`SpellsAsLogicalAnd`). When `c` is a **user-defined-truthiness** condition (`operator true`/`false`), the printer strips it to the bare user-typed receiver, so `c && value` rebinds to a user-defined `&` that need not exist → **uncompilable C# (CS0019)**. This is exactly the shape the default's `FoldGuardReturn` declines (#3114).

### Resolution — commit `e3b9f5f4`

Mirror the default's guard: the lens now declines when `ShortCircuitFidelity.IsUserDefinedTruthiness(guard.Condition)`, leaving the flat (valid, faithful) guard. Verified:

- `UserTruthinessGuard(Truthy c, bool value)` — lens output is now **byte-identical to the default** (stays `if (c) { return value; } return false;`); no `c && value`, no CS0019.
- `AndShapedGuard(bool c, bool value)` — ordinary-bool case still renders the valid, behavior-faithful `=> c && value;`.
- New tests pin both; fast decompiler suite **3559/0**.

The reviewers disagreed on nothing; GPT's finding was reproduced on the clean head before fixing and is now covered by a regression test. Re-review of the fixed head `e3b9f5f4` is running.

richlander:
## Fixed-head re-review — both clean

Re-reviewed the fixed exact head `e3b9f5f4` (base `be4291ee`) with both models in isolated read-only worktrees.

- **GPT-5.6 (original finder) — clean.** Confirmed its Medium finding is fully resolved: the user-truthiness lens output is now byte-identical to the default (flat, no `&&`, no CS0019); ordinary-bool guard still renders valid `c && value`. Re-checked negated, nested, and non-literal boolean shapes for any remaining unsafe printer collapse — none found. Evidence: Release build; lens tests 24/0; config tests 31/0; fast decompiler suite 3559/0.

- **Gemini 3.1 Pro — clean.** Independently reproduced the original CS0019 via a `/tmp` compiler probe (`c && value` with the `Truthy` struct throws CS0019; `c ? value : false` compiles), confirming the decline is essential. Verified the early return sits before any `Detach()` so no partial mutation on decline, and that the `Truthy` fixture omits `operator &`/`|` to exercise the exact shape. Evidence: rebuilt at head; `PreferConditionalReturnLensTests` 10/0; full non-slow suite 3212/0.

CI on `e3b9f5f4`: `test`, `build-net10`, `csharp-diff-smoke`, `markdownlint`, `changes` all pass (`il-diff-smoke`/`pack` path-gated, skipped).

Ready to merge

PR #3140, opened 2026-07-24
Add PreferBranchlessBoolean style lens (#3138)

richlander:
## Summary

The **second** opt-in tier-3 style lens from the #3138 umbrella (behavior-faithful, byte-divergent), following the ternary lens (#3139). `PrinterOptions.PreferBranchlessBoolean` renders a guarded boolean return with a constant arm — which the default view must leave flat because the short-circuit fold is not opcode-faithful for a bare-load operand (#3114) — as the compact short-circuit "bool hack":

```csharp
// default (byte-faithful, the flat guard recompiles exactly):
if (a) { return false; }
return b;
// with PreferBranchlessBoolean (behavior-faithful, byte-divergent):
return !a && b;
```

The four constant-arm shapes fold to `a && b`, `!a || b`, `a || b`, `!a && b`. This re-offers **exactly** the fold `BooleanFoldingPass.FoldGuardReturn` emitted *before* #3114 guarded it — no new capability, a second lens payload.

## Contract: behavior hazards kept, opcode hazards relaxed

The default's `FoldGuardReturn` declines these shapes via three fidelity guards. The lens separates **behavior** hazards (kept) from pure **opcode** hazards (relaxed):

- **Kept — user-defined-truthiness condition**: lifting it into `t && b` rebinds to a user-defined `&&`/`||`, reselecting an overload and changing the runtime result (also typically CS0019). Declined, mirroring #3114/#3139.
- **Kept — managed by-ref surviving operand**: csc's branchless lowering eagerly dereferences a location the branch had guarded (null-by-ref NRE divergence). Declined.
- **Relaxed — negation opcode-exactness** (`NegationIsOpcodeExact`): `Conditions.Negate` is universally behavior-preserving (for floats it flips ordered/unordered to give the correct NaN dual; it is only *opcode*-divergent).
- **Relaxed — bare-place `&`/`|` collapse** (`RendersAsBranchlessBarePlace`): a bare local/arg/stack load is side-effect-free and non-faulting, so csc's branchless form is behavior-identical. (This is the poster case — it is what produces the branchless output the lens exists for.)

## Not oracle-endorsed

dotnet/runtime's `.editorconfig` would never recommend `!(a && b) && c`, so this is a user *compactness/branchless* preference. It is exposed under a **tool-owned** `dotnet_inspect_style_prefer_branchless_boolean` key (not a `dotnet_style_*` key) and is **never** part of a future "full taste" aggregate. When both lenses are enabled the oracle-endorsed **ternary wins** the shared shape (it runs first and consumes the guarded return, leaving this pass a no-op) — deterministic precedence.

## Scope / isolation

Runs only on the opt-in `PrintRaised` options overload, after the default pipeline. The IL-anchored Annotated view (`out statementLines`) and `PrintLowered` do not take options → stay byte-faithful and excluded from the compile-back fidelity gates by construction.

## Evidence

- **Specimen + tests** pin all four fold shapes (`a && b`, `!a || b`, `a || b`, `!a && b`), both behavior-guard declines (user-truthiness, by-ref stay flat), two no-op negatives (both-variable no constant arm; no guarded return), ternary-wins precedence, and an **executed truth-table behavioral-equivalence gate** over every boolean input.
- **Config**: `RenderStyleConfig` key-mapping + severity-tolerance tests.
- **CLI end-to-end**: `.dotnet-inspectconfig` with the key → `=> !a && b;`; without → flat guard.
- Decompiler fast suite **3594/0**; config tests **33/0**; full `slnx` build clean.

## Review

Requires 2-model adversarial review (new behavior/shape). Base `main` @ `70a3dc2e`, head `a66e255a`.

### DIFF
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -286,6 +286,33 @@ d` collapse (IDE0075) is a separate future knob, so a literal-arm ternary such a
 the opt-in raised path, after the default pipeline, and the IL-anchored Annotated
 view never applies it (it must stay byte-faithful for line/IL alignment).
 
+The second lens is `PrinterOptions.PreferBranchlessBoolean`
+(`dotnet_inspect_style_prefer_branchless_boolean`). It targets the *same* declined
+guarded return, but renders the compact short-circuit "bool hack" — the exact form
+the default's short-circuit fold produced before #3114 guarded it — instead of the
+ternary:
+
+```csharp
+// shipped default (byte-faithful — the flat guard is what recompiles exactly):
+if (a) { return false; }
+return b;
+// with PreferBranchlessBoolean (behavior-faithful, byte-divergent):
+return !a && b;
+```
+
+The four constant-arm shapes fold to `a && b`, `!a || b`, `a || b`, and `!a && b`
+respectively. Unlike the ternary, this form is **not oracle-endorsed** —
+dotnet/runtime's `.editorconfig` would never recommend it — so it is a user
+*compactness/branchless* preference, opt-in only, and it is **never** part of a
+"full taste" aggregate. It is exposed under a tool-owned
+`dotnet_inspect_style_*` key rather than a `dotnet_style_*` key to make that
+distinction explicit. Two hazards stay declined because they are about *behavior*,
+not just bytes: a user-defined-truthiness condition (lifting it rebinds to a
+user-defined `&&`/`||`, changing the result) and a managed by-ref surviving
+operand (csc's branchless lowering would eagerly dereference a location the branch
+had guarded). When both lenses are enabled the oracle-endorsed ternary wins the
+shared shape.
+
 ## Names
 
 Without a PDB, locals are slot names (`V_0`, `S_0`) shared with the Annotated IL view — the two views stay name-aligned by construction. With a PDB, source names are used. Synthesizing readable names (`size`, `array`, `item`) where no PDB exists is an open design question: it is the largest remaining cosmetic gap against source, but it would break view alignment unless opt-in.
@@ -295,8 +322,8 @@ Without a PDB, locals are slot names (`V_0`, `S_0`) shared with the Annotated IL
 The oracle settles a single shipped default per equivalence class, but a few
 class-3 no-anchor spellings are also exposed as opt-in knobs (see
 [`this` member qualification](#this-member-qualification) and
-[Line wrapping](#line-wrapping)), as is the first byte-divergent
-[style lens](#style-lenses-behavior-faithful-byte-divergent). A tool-owned config
+[Line wrapping](#line-wrapping)), as are the byte-divergent
+[style lenses](#style-lenses-behavior-faithful-byte-divergent). A tool-owned config
 file selects them without a per-run flag.
 
 `dotnet-inspect` discovers a `.dotnet-inspectconfig` file by walking up from the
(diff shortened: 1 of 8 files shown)


richlander:
## Adversarial review reconciliation (head `a66e255a` → fixed `97a68ebe`)

Two-model review per AGENTS (I'm Opus, so did not self-review).

### Gemini 3.1 Pro — CLEAN
Verified via IL dumps that managed by-ref is the only faulting branchless operand (locals/args/stack slots are safe; field/call/pointer-deref keep the branch). Confirmed float-negation NaN correctness (`IsUnsigned` flip gives the ordered/unordered dual), user-truthiness re-bind safety, fixpoint termination, no partial mutation on decline, and ternary-wins precedence when both lenses are enabled.

### GPT-5.6 — 1 Medium validity bug (FIXED in `97a68ebe`)
**Finding:** the lens declined only conditions whose *root* is a user-defined-truthiness call (`op_True`/`op_False`). A condition wrapping that call under a `LogicalNot` slipped past the guard. In a NEGATING fold shape, `Conditions.Negate` unwraps the `LogicalNot` to the bare `op_True` call the printer strips to `t`, emitting the uncompilable `t || b` (`Truthy` has no user `|` → CS0019). The default's incidental `NegationIsOpcodeExact` opcode guard had been shielding `FoldGuardReturn` from this shape; relaxing it in the lens exposed the gap.

**Verified & confirmed** as a real reachable defect — not just synthetic IR. `if (t) {} else { return b; } return true;` raises to `LogicalNot(op_True(t))` with a `true` tail (negating). With the pre-fix direct-only guard the lens produced `=> t || b;`; a real compiled fixture (`NegatedUserTruthinessGuard`) reproduces it.

**Fix:** replaced the direct-root `IsUserDefinedTruthiness(guard.Condition)` check with a subtree scan (`InvolvesUserDefinedTruthiness`) catching `op_True`/`op_False` anywhere in the condition (root, under a `LogicalNot`, or nested). Over-declining is always safe (stays flat = valid + faithful). Added `NegatedUserTruthinessGuard` compiled fixture + a pinning decline test.

**Evidence:** post-fix probe declines (stays flat); lens tests 26/0 (+1), config 33/0, fast decompiler suite 3595/0 (+1). Both fixed-head re-reviews to follow before `Ready to merge`.

richlander:
## Fixed-head re-review — both CLEAN (head `b4a5f718`)

Follow-up commits since `a66e255a`:
- `97a68ebe` — fix GPT's Medium validity bug (negation-wrapped user-truthiness → uncompilable `t || b`): replaced the direct-root check with a subtree scan; added the `NegatedUserTruthinessGuard` real compiled fixture + pinning test.
- `b4a5f718` — record the truthiness decline as intentional lens scope: added `WrappedTruthinessNonNegatingGuard` (GPT's valid-but-ternary case), strengthened comments, documented the wholesale-truthiness boundary in `docs/decompiler-taste.md`.

### GPT-5.6 — CLEAN (withdrew Medium)
On re-review GPT raised a second Medium: the subtree scan also declines a *valid* fold — `if (t) {} else { return b; } return false;` → `(t ? false : true) && b`. Confirmed empirically. Resolved as **intentional scope, not a defect**: every user-defined-truthiness fold is (a) invalid (CS0019), (b) behavior-divergent were a user `&`/`|` present, or (c) valid but not branchless — a negation spelled as the ternary `(t ? false : true)` re-embeds a branch, which is not the compact short-circuit form this lens produces. No truthiness condition yields a clean/valid/faithful branchless form, and the scan's blast radius is exactly the truthiness family (plain bools/comparisons/nullable never emit `op_True`/`op_False`). Declining is always valid + faithful. GPT reviewed the rationale and **withdrew the finding**: "the flat guard is preferable to `(t ? false : true) && b`; restoring it is not a genuine requirement." Verified at `b4a5f718`: build ✓, lens 27/27, markdownlint ✓, no correctness/termination hazards.

### Gemini 3.1 Pro — CLEAN
Verified the `97a68ebe` fix seals the gap (subtree scan mirrors the printer's strip; `IrNode.Descendants` is self-exclusive so the root check is necessary + sufficient; no legitimate over-decline of the 4 fold shapes; no new mutation/termination hazard). Confirmed the `b4a5f718` delta is inert (comments/tests/docs only), 27/27 pass.

Both fixed-head reviews clean.

richlander:
Ready to merge

PR #3125, opened 2026-07-24
Expression-body single multi-line switch-expression returns (#3088)

richlander:
- Fixes/advances #3088
- Renders a method/accessor whose only statement is a multi-line `return <switch-expression>;` as an expression-bodied member.
- Card revision: `2423ab60`

## Change

> Should we accept this change?

**Conclusion:** **PASS** — a sole `return <switch>;` now renders `head => <value> switch { ... };` on both the whole-type and single-member rendering paths, gated on a typed structural signal from the printer (never string-sniffing).

### Original source

Original C# (compiled into the fixture, shown for context — SourceLink source is unavailable because the fixture carries no PDB):

```csharp
public static int Area(object shape) => shape switch
{
    string s => s.Length,
    int[] a => a.Length,
    _ => 0,
};
```

Authoritative `IL` (`member Shapes -m Area ... -S "IL"`):

```il
IL_0000: ldarg.0
IL_0001: isinst       System.String
IL_0006: stloc.0
IL_0007: ldloc.0
IL_0008: brtrue.s     IL_0016
IL_000A: ldarg.0
IL_000B: isinst       int[]
IL_0010: stloc.1
IL_0011: ldloc.1
IL_0012: brtrue.s     IL_0021
IL_0014: br.s         IL_0029
IL_0016: br.s         IL_0018
IL_0018: ldloc.0
IL_0019: callvirt     System.String::get_Length()
IL_001E: stloc.2
IL_001F: br.s         IL_002D
IL_0021: br.s         IL_0023
IL_0023: ldloc.1
IL_0024: ldlen
IL_0025: conv.i4
IL_0026: stloc.2
IL_0027: br.s         IL_002D
IL_0029: ldc.i4.0
IL_002A: stloc.2
IL_002B: br.s         IL_002D
IL_002D: ldloc.2
IL_002E: ret
```

### Before

`member Shapes -m Area ... -S "Decompiled Source"` at the base commit (`da030adf`):

```csharp
public static int Area(object shape)
{
    return shape switch
    {
        string V_0 => V_0.Length,
        int[] V_1 => V_1.Length,
        _ => 0,
    };
}
```

- Valid: True
- Correct: True

The raised switch expression is already correct; it is just block-bodied rather than the idiomatic expression-bodied member form.

### After

`member Shapes -m Area ... -S "Decompiled Source"` at this PR's head (`2423ab60`):

```csharp
public static int Area(object shape) => shape switch
{
    string V_0 => V_0.Length,
[...]

Issue #3084, opened 2026-07-23
Decompiler taste: transition single-expression members to expression-bodied (=>)


richlander:
## Summary

The decompiler always renders members with a **block body** (`{ ... }`), even
when the body is a single `return expr;` (or a single expression statement) that
idiomatic C# — and `dotnet/runtime` source — would spell as an
**expression-bodied member** (`=> expr;`). Propose an **expression-bodied
transition** taste transform: when a method/property/accessor/operator body
reduces to exactly one `return <expr>;` (or one expression statement, or a
throw), render it as `=> <expr>;` instead of `{ return <expr>; }`.

This is a **taste / valid-but-different presentation** question, not a
correctness bug. The two spellings are IL-identical, so it belongs to the
canonical-form machinery in
[docs/decompiler-taste.md](../blob/main/docs/decompiler-taste.md) — it is the
exact example `#3027` already names as analogous.

## Witness

Surfaced as a follow-up to #3009 (PR #3073). After the flags-enum accumulator is
fully raised, `Accumulate` reduces to a single `return`:

Original source shape (a single expression the compiler emits with no block-body
distinction):

```csharp
public static int Accumulate(FlagCaps64 server, bool interactive)
    => (int)(FlagCaps64.Protocol
        | (interactive ? (server & FlagCaps64.Interactive) : FlagCaps64.None)
        | (server & FlagCaps64.LoadLocal)
        | FlagCaps64.Secure
        | (server & FlagCaps64.MultiStatements)
        | FlagCaps64.MultiResults);
```

Decompiled (correct, but valid-but-different) — always a block body:

```csharp
public static int Accumulate(FlagCaps64 server, bool interactive)
{
    return (int)(FlagCaps64.Protocol | ...);
}
```

`{ return e; }` and `=> e;` compile to identical IL; the source author's choice
of body form is unrecoverable from the IL alone.

## Why this is a taste question (three-class rule)

Per `decompiler-taste.md`:

- **Not class 2 (fidelity-erasing).** Expression-bodied vs block-bodied does
  **not** erase a distinction the IL makes — the two forms are the *same* IL.
  Nothing observable is lost.
- **Class 3 (no IL anchor — oracle tiebreaker).** The two forms are one IL
  equivalence class; we pick one canonical representative. Where a body is a
  single expression, `dotnet/runtime` style overwhelmingly prefers the
  expression-bodied form, so the dominant runtime source style is the oracle.
[...]

PR #3141, opened 2026-07-24
Expression-body multi-line single-return expressions (#3084)

richlander:
- Fixes/advances #3084
- Renders any method/accessor whose only statement is a multi-line `return <expr>;` (a wrapped fluent chain, ternary, or the four switch-expression kinds) as an expression-bodied member.
- Card revision: `66812cc9`

## Change

**Conclusion:** **PASS** — generalizes the single-return expression-body fold #3125 shipped for multi-line switch returns to *any* multi-line single `return <expr>;`, gated on the same typed structural signal (never string-sniffing). One broadened printer gate; all downstream threading was already general.

### Original source

The witness is a csc-compiled fixture (`MultiLineExpressionBodySamples.Pipeline`) with no PDB, so SourceLink C# is unavailable. The compiled-in C#, shown for context:

```csharp
public static string Pipeline(StringBuilder builder)
    => builder
        .Append("alphabet").Append("bravissimo").Append("charlateral")
        .Append("deltatango").Append("echolocation").Append("foxtrotter")
        .ToString();
```

### Before

`member MultiLineExpressionBodySamples Pipeline --library <tests.dll> -S "Decompiled Source"` at the base commit (`402ef9c7`, #3125):

```csharp
public static string Pipeline(System.Text.StringBuilder builder)
{
    return builder
        .Append("alphabet")
        .Append("bravissimo")
        .Append("charlateral")
        .Append("deltatango")
        .Append("echolocation")
        .Append("foxtrotter")
        .ToString();
}
```

- Valid: True
- Correct: True

The single expression is already fully raised; it is just block-bodied rather than the idiomatic expression-bodied form (#3125 only folded switch returns, not fluent chains).

### After

Same command at this PR's head (`66812cc9`):

```csharp
public static string Pipeline(System.Text.StringBuilder builder) => builder
    .Append("alphabet")
    .Append("bravissimo")
    .Append("charlateral")
    .Append("deltatango")
    .Append("echolocation")
    .Append("foxtrotter")
    .ToString();
```

- Valid: True
- Correct: True

The arrow trails the signature with the chain receiver after it and the calls re-indent one level under the member — reusing #3125's `AppendMultilineExpressionBody`. A sibling method with a statement before the return correctly stays block-bodied.

### Fully raised

## Evidence

| Check | Baseline (`402ef9c7`) | Head (`66812cc9`) |
| --- | --- | --- |
| `MemberBodyProducerExpressionBodyTests` (e2e) | n/a (2 new tests) | 4 total, 0 failed |
| `MultiLineExpressionBodyRenderTests` (witness golden) | n/a (new) | 1 total, 0 failed |
| `ILInspector.CSharp.Tests` | 321, 0 failed | 321, 0 failed |
| Decompiler fast suite | 3568, 0 failed | 3574, 0 failed (+6 new) |
| Full `dotnet-inspect.Tests` | 2026, 1 failed | 2026, 1 failed |

The one full-suite failure, `PlatformResolverTests.GetPacksDirectory_ReturnsValidPath`, is pre-existing and environmental (packs-v2 cache path on this machine); it passes in isolation on both baseline and head and is unchanged by this PR (same test, same reason).

## Decompiler quality

**Conclusion:** **PASS** — presentation-only, exactly as #3125. This PR only sets the `BodyIsSingleReturnExpression` flag consumed by the member-layout layer; it never changes the raised IR or the printer's body string (`return X;`). Only the block-vs-expression-bodied member framing of an already-fully-raised body differs, so lowering/branch residue and fidelity metrics are identical to baseline and the corpus quick gate is not applicable to this change.

## Validation

```bash
dotnet build src/dotnet-inspect.slnx -c Release
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.MemberBodyProducerExpressionBodyTests -class ILInspector.Decompiler.Tests.MultiLineExpressionBodyRenderTests
dotnet run --project src/ILInspector.CSharp.Tests -c Release
```

## Scope boundary

Multi-line single `throw` and expression-statement bodies remain a deliberate follow-up; their single-line forms already fold via `FromSingleStatement`.

### DIFF
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -156,6 +156,46 @@ public int Compute() => _count + Extra;          // shipped default: bare
 public int Compute() => this._count + this.Extra; // both knobs on
 ```
 
+### Expression-bodied members
+
+Rendering a value-returning member as `head => <expr>;` instead of a brace block
+wrapping a lone `return <expr>;` is an IL-identical framing choice — the two
+forms are a language-guaranteed equivalence — so it sits below the three-class
+rule, alongside brace style and line wrapping. The oracle is `dotnet/runtime`'s
+`.editorconfig` (`csharp_style_expression_bodied_methods` /
+`_properties` / `_accessors = true`), so the expression-bodied form is the
+shipped default.
+
+A body that is exactly one statement folds on the simple single-line path (a
+lone `return <expr>;`, a `throw`, or a statement-expression). A body that is one
+*multi-line* single expression folds too: a wrapped switch return (issue #3088)
+or any other wrapped single `return <expr>;`, such as a fluent chain
+(issue #3084). The arrow trails the signature line with the expression's opening
+token after it, and the continuation lines re-indent one level under the
+member — the natural multi-line extension of the single-line `head => expr;`
+form.
+
+```csharp
+public static string Pipeline(StringBuilder builder)
+{
+    return builder
+        .Append("alphabet")
+        .Append("bravissimo")
+        .ToString();
+}
+// folds to:
+public static string Pipeline(StringBuilder builder) => builder
+    .Append("alphabet")
+    .Append("bravissimo")
+    .ToString();
+```
+
+The fold is gated on a typed structural signal the printer proves from the
+emitted statement tree — the body is exactly one top-level `return` with no
+lifted declarations, label, constructor chain, field initializer, async
+modifier, or unsupported fallback — never a re-parse of the rendered text. A
+member with any statement preceding the return keeps its brace block.
+
 ### Line wrapping
 
 Breaking a long line across continuation lines is pure whitespace: it emits the
--- a/src/ILInspector.CSharp.Tests/CSharpExpressionBodyTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpExpressionBodyTests.cs
@@ -47,6 +47,31 @@ public sealed class CSharpExpressionBodyTests
             lines);
     }
 
+    [Fact]
+    public void MultilineReturnExpressionLines_SplitsFluentChainReturn()
+    {
+        // Issue #3084: extraction is not switch-specific. A wrapped fluent chain
+        // return yields its receiver as the value line and each chained call as a
+        // continuation line at its body-relative indent.
+        var lines = CSharpExpressionBody.MultilineReturnExpressionLines(
+            "return builder\n    .Append(\"a\")\n    .Append(\"b\")\n    .ToString();");
+        Assert.NotNull(lines);
+        Assert.Equal(
+            ["builder", "    .Append(\"a\")", "    .Append(\"b\")", "    .ToString()"],
+            lines);
+    }
+
+    [Fact]
+    public void MultilineReturnExpressionLines_SplitsWrappedTernaryReturn()
+    {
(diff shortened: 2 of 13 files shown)


richlander:

Two model families reviewed this PR per the AGENTS.md roster (I am Claude, so I used GPT + Gemini, not my own family). Each reviewer ran isolated in its own linked worktree at a fixed head, with scratch confined to `/tmp` and no git mutations in the review tree.

### Round 1 — fixed head `66812cc9`

| --- | --- | --- |
| GPT | gpt-5.6-sol | No blocking findings; 1 non-blocking (Medium) |
| Gemini | gemini-3.1-pro-preview | CLEAN |

Both independently exercised the six attack points (unsafe fold, multi-line text guard, `_topLevelStatements` population, per-invocation statefulness, accessor paths, golden regressions) with real builds and the fast suite (3574 pass / 0 fail).

**GPT non-blocking finding (flag honesty):** a lone `return <stackalloc>;` that the printer expands into a lifted local declaration inside an `unsafe` block is still one top-level `Return` and its output is multi-line, so the earlier text guard (`Trim().Contains('\n')`) set `BodyIsSingleReturnExpression = true`. Output was already correct — `CSharpExpressionBody.MultilineReturnExpressionLines` re-gates on a leading `return ` and the member layer fell back to the brace block — so this was **not** an unsafe fold. Gemini reached the same conclusion and rated it safe. It was a contract-honesty gap: the typed flag did not reflect a body the member layer would actually fold.

### Resolution — commit `30e49b5c`

I tightened the gate so the typed flag is honest instead of relying solely on downstream re-gating. `BodyIsSingleReturnExpression`'s text guard is now `IsMultilineReturnExpressionText(output)`, which requires the trimmed output to **both** contain `\n` **and** begin with `return ` (Ordinal) — the exact shape `MultilineReturnExpressionLines` accepts. The flag now stays off for any single `Return` the printer expanded into multiple statements, so it never claims a fold the member layer would reject, closing the seam should the extractor ever relax.

The change is **output-neutral**: the newly-excluded cases already rendered as brace blocks, so no golden changed (fast suite unchanged). Added a compiler-backed close negative `SingleStackallocPointerReturn_StaysBlock` asserting a lifted stackalloc pointer return keeps its brace block (`stackalloc byte[40]` decl + `return (int*)__stackalloc;`, no `=>`).

### Round 2 — re-review at fixed head `30e49b5c`

| --- | --- | --- |
| GPT | gpt-5.6-sol | CLEAN — fast suite 3575 pass / 0 fail |
| Gemini | gemini-3.1-pro-preview | CLEAN — findings severity 0 |

Both re-reviews confirmed the new guard co-gates **exactly** with `MultilineReturnExpressionLines` (no true-but-null residue, no legitimate multi-line `return <expr>;` fold wrongly suppressed — leading trivia/whitespace/CRLF all handled by `Trim()`), and that the change introduces no golden flips or regressions.

richlander:
Ready to merge

Both fixed-head adversarial re-reviews at `30e49b5c` are CLEAN (GPT + Gemini), and all merge-blocking CI checks pass (`test` 8m31s, `build-net10`, `csharp-diff-smoke`, `markdownlint`, `changes`; `il-diff-smoke`/`pack` are path-gated and skipped). No follow-up items outstanding.

PR #3146, opened 2026-07-24
Fold multi-line single expression-statement bodies (#3084)

richlander:
- Advances #3084
- Changes a void member whose one statement is a multi-line expression statement (a wrapped fluent chain) to render `head => <expr>;` instead of a brace block
- Card revision: `684a9923`

## Change

**Conclusion:** **PASS** — an expression statement has no leading-declaration lift path, so its whole printed form is that one statement; folding it to `head => <expr>;` is a language-guaranteed, IL-identical equivalence, gated on the same typed structural signal already used for returns.

This generalizes the multi-line single-statement fold shipped in #3141 (switch/fluent/other single `return <expr>;`) to the remaining reachable shape in #3084: a void member whose one statement is a wrapped expression statement. The extractor is made shape-agnostic about the leading keyword (it strips `return ` when present, otherwise keeps the whole first line), and the printer signal now accepts one `Return { Value: not null }` (requiring a bare `return ` prefix) **or** one `ExpressionStatement` (multi-line guard alone). Multi-line throws are covered by the shape-agnostic extractor but the printer never produces one today, so no dead printer branch was added.

Rename for accuracy (the fold is no longer return-only): `BodyIsSingleReturnExpression` → `BodyIsSingleExpressionBody`; `MultilineReturnExpressionLines` → `MultilineExpressionBodyLines`.

### Original source

```csharp
public static void Drain(System.Text.StringBuilder builder)
{
    builder.Append("alphabet").Append("bravissimo").Append("charlateral").Append("deltatango").Append("echolocation").Append("foxtrotter").Clear();
}
```

### Before

```csharp
public static void Drain(System.Text.StringBuilder builder)
{
    builder
        .Append("alphabet")
        .Append("bravissimo")
        .Append("charlateral")
        .Append("deltatango")
        .Append("echolocation")
        .Append("foxtrotter")
        .Clear();
}
```

- Valid: True
- Correct: True

The block form is valid and correct; it is just the lower-quality framing (a brace block wrapping a single statement the runtime's `.editorconfig` would render expression-bodied).

### After

```csharp
public static void Drain(System.Text.StringBuilder builder) => builder
    .Append("alphabet")
    .Append("bravissimo")
    .Append("charlateral")
    .Append("deltatango")
    .Append("echolocation")
    .Append("foxtrotter")
    .Clear();
```

- Valid: True
- Correct: True

The body string is byte-for-byte identical to Before; only the member framing changed (block → expression-bodied). Presentation-only, like #3141.

### Fully raised

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `MemberBodyProducerExpressionBodyTests` / `MultiLineExpressionBodyRenderTests`: pre-existing | `MemberBodyProducerExpressionBodyTests` 7 passed, `MultiLineExpressionBodyRenderTests` 2 passed |
| Decompiler fast suite | 3605, 0 failed | 3605, 0 failed |
| CSharp suite | 323, 0 failed | 325, 0 failed |
| dotnet-inspect suite | 2028, 0 failed (12 ilasm skips) | 2028, 0 failed (12 ilasm skips) |
| Real witness | `Widget::Drain` block-wrapped single statement | `Widget::Drain` expression-bodied (real `member --library` output above) |

No golden flips: the fast suite total is unchanged aside from the new tests; no existing corpus fixture had a multi-line void fluent statement body, so this is additive.

## Decompiler quality

**Conclusion:** **PASS** — presentation-only (member framing), like #3141: the decompiled body string is unchanged, so residue/fidelity/fully-raised are identical to baseline and the PR quick gate is not applicable. The only difference is the block-vs-expression-body member framing the CSharp layout layer owns.

## Validation

```bash
export PATH="$HOME/.local/share/dotnet:$PATH"
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.MemberBodyProducerExpressionBodyTests -class ILInspector.Decompiler.Tests.MultiLineExpressionBodyRenderTests
dotnet run --project src/ILInspector.CSharp.Tests -c Release
dotnet run --project src/dotnet-inspect.Tests -c Release
```

### DIFF
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -168,12 +168,14 @@ shipped default.
 
 A body that is exactly one statement folds on the simple single-line path (a
 lone `return <expr>;`, a `throw`, or a statement-expression). A body that is one
-*multi-line* single expression folds too: a wrapped switch return (issue #3088)
-or any other wrapped single `return <expr>;`, such as a fluent chain
-(issue #3084). The arrow trails the signature line with the expression's opening
-token after it, and the continuation lines re-indent one level under the
-member — the natural multi-line extension of the single-line `head => expr;`
-form.
+*multi-line* single statement folds too: a wrapped switch return (issue #3088),
+any other wrapped single `return <expr>;` such as a fluent chain (issue #3084),
+or a void member whose one statement is a wrapped expression statement — a fluent
+call chain with the result discarded (issue #3084). The arrow trails the
+signature line with the statement's opening token after it (the value after a
+stripped `return`, or the whole first line otherwise), and the continuation
+lines re-indent one level under the member — the natural multi-line extension of
+the single-line `head => expr;` form.
 
 ```csharp
 public static string Pipeline(StringBuilder builder)
@@ -188,13 +190,33 @@ public static string Pipeline(StringBuilder builder) => builder
     .Append("alphabet")
     .Append("bravissimo")
     .ToString();
+
+public static void Drain(StringBuilder builder)
+{
+    builder
+        .Append("alphabet")
+        .Append("bravissimo")
+        .Clear();
+}
+// folds to (no `return` to strip — the whole first line trails the arrow):
+public static void Drain(StringBuilder builder) => builder
+    .Append("alphabet")
+    .Append("bravissimo")
+    .Clear();
 ```
 
 The fold is gated on a typed structural signal the printer proves from the
-emitted statement tree — the body is exactly one top-level `return` with no
-lifted declarations, label, constructor chain, field initializer, async
-modifier, or unsupported fallback — never a re-parse of the rendered text. A
-member with any statement preceding the return keeps its brace block.
+emitted statement tree — the body is exactly one top-level statement (a `return`
+with a value, or an expression statement) with no lifted declarations, label,
+constructor chain, field initializer, async modifier, or unsupported fallback —
+never a re-parse of the rendered text. A `return` branch additionally requires
+its printed form to begin with a bare `return` so a value the printer lifted
+into a leading declaration (a `stackalloc`-to-pointer return) is not mistaken for
+a foldable expression. A member with any statement preceding the folded one keeps
+its brace block. (The extractor is shape-agnostic about the leading keyword, so a
+wrapped `throw <expr>;` would fold identically should one ever print multi-line;
+the printer does not currently produce multi-line throws, so that path is latent,
+not reachable.)
 
 ### Line wrapping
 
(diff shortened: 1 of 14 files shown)


richlander:

Two model families (GPT `gpt-5.6-sol`, Gemini `gemini-3.1-pro-preview`), each isolated in its own linked worktree at fixed head, scratch under `/tmp/`, no git mutations.

### Round 1 — head `684a9923`

- **Gemini: CLEAN.** Verified the discard `_ = <chain>;` folds validly, stackalloc/ref-lift statements set `_emittedDeclarations` and block the fold, accessors/constructors are shielded, throws are latent-only, and the fast suite ran 3605/0.
- **GPT: 1 non-blocking finding (Medium / High).** Flag-honesty gap: under the new memory-safety rules the printer wraps a lone unsafe statement as `unsafe { <stmt>; }` — a multi-line body ending in `}`, not `;`. The old `IsFoldableSingleStatement` accepted any multi-line single `ExpressionStatement`, so `BodyIsSingleExpressionBody` was set `true` even though `MultilineExpressionBodyLines` rejects that shape. Output was already correct (both `CSharpMemberLayout` and `ApiOutputFormatter` re-run the extractor and keep the brace block), but the typed flag's invariant (flag-true ⟹ the extractor can supply the fold) was violated.

### Resolution — `c358a636`

Adopted GPT's suggested fix: tightened the shared guard in `IsFoldableSingleStatement` to require the trimmed output to end in `;`, exactly matching the extractor's contract. The `Return` branch already excluded the unsafe wrapper (it starts with `unsafe`, not `return `); the trailing-`;` guard now covers the `ExpressionStatement` branch symmetrically.

- Reproduced the finding first: new `UnsafeEmitterTests.NewRulesModule_SingleUnsafeExpressionStatement_NotFlaggedExpressionBody` (over `NewUnsafe.FreePointer`) **failed** before the fix (flag `true`) and **passes** after (flag `false`, body ends in `}`).
- Added an extractor-level reject case for the `unsafe { … }` wrapper in `CSharpExpressionBodyTests`.

### Round 2 (re-review of fixed head) — head `c358a636`

- **GPT: CLEAN.** Confirmed the guard is a strict subset of the extractor's contract (no regression), `trimmed[^1]` is safe via `||` short-circuit, original return/void-fluent folds intact, fast suite 3606/0.
- **Gemini: CLEAN.** Confirmed the guard strictly parallels the extractor's `EndsWith(';')`, block-wrapped bodies (`unsafe`/`checked`) resolve to `}` and are rejected, and the printer structurally never emits a `return \n …` gap (`Expression(...)` never starts with a newline), so the co-gate holds natively. Fast suite 3606/0.

Both fixed-head re-reviews are clean. No blocking findings outstanding; no explicit non-actions.

### Evidence at `c358a636`

- Full `dotnet-inspect.slnx` build: clean.
- Decompiler fast suite: 3606 pass / 0 fail (no golden flips — additive). CSharp: 326 / 0. dotnet-inspect: 2039 / 0 (12 ilasm skips).
- CI on this exact head: `build-net10`, `test`, `csharp-diff-smoke`, `markdownlint`, `changes` all pass; `il-diff-smoke`/`pack` path-skipped.

Ready to merge

PR #3148, opened 2026-07-25
Decline FoldGuardReturn folds that eagerly deref a chain-buried by-ref (#3114)

richlander:
- Advances #3114 — focused correctness follow-up to merged #3127 (`be4291ee`)
- Supersedes #3119 (salvages its still-needed by-ref soundness delta)
- Changes `FoldGuardReturn` to also decline a short-circuit fold when the surviving operand hides a managed by-ref dereference inside a same-kind logical chain (or under a reducible bool wrapper), not only when the by-ref is the top-level bare operand
- Card revision: `e0d919c3`

## Change

**Conclusion:** **PASS** — merged #3127 guarded only the top-level bare by-ref operand; a by-ref buried in the flattened same-kind chain still folds into an unconditional dereference that recompiles to a `NullReferenceException` divergence. This restores the decline for that chain-buried case while keeping #3127's opcode-fidelity policy verbatim.

### Original source

Authored fixture; no SourceLink source server, so the authoritative anchor is the raw IL. The `bgt.s` guard proves `r` (the by-ref) is dereferenced (`ldind.u1`) **only** when `start <= 0`; when `start > 0` control jumps straight to `return false` without touching `r`.

```il
IL_0000: ldarg.0
IL_0001: ldc.i4.0
IL_0002: bgt.s        IL_0010   // start > 0: skip the deref, return false
IL_0004: ldarg.1
IL_0005: ldind.u1               // *r — reached only when start <= 0
IL_0006: brfalse.s    IL_000E
IL_0008: call         WitnessNs.Hazard::Call()
IL_000D: ret
IL_000E: ldc.i4.0
IL_000F: ret
IL_0010: ldc.i4.0               // start > 0: return false, r untouched
IL_0011: ret
```

### Before

Base `d0d0ebc5` (merged #3127), `-S "Decompiled Source"`:

```csharp
public static bool ByRefBeforeCall(int start, ref bool r) => start <= 0 && r && Call();
```

- Valid: True
- Correct: False

The printer flattens the same-kind `&&` chain, so `r` becomes a bare operand of `start <= 0 && r && Call()`. csc recompiles the call-free prefix `start <= 0 && r` to a **branchless** `(start <= 0) & r` (a by-ref deref is a bare, side-effect-free place to csc), dereferencing `r` even when `start > 0`. On a null by-ref with `start > 0` the recompiled body throws where the original IL returned `false` — an observable divergence #3127's `RendersAsBranchlessBarePlace` missed because it never descends the operand.

### After

Head `e0d919c3`, `-S "Decompiled Source"`:

```csharp
public static bool ByRefBeforeCall(int start, ref bool r)
{
    if (start <= 0)
    {
        return r && Call();
    }
    return false;
}
```

- Valid: True
- Correct: True

The fold is declined, so the guard survives and `r` stays behind the `start <= 0` branch, recompiling faithfully.

### Fully raised

The After decompilation is in the fully raised state: for this shape the faithful endpoint is the branch-preserving form, because any short-circuit fold would reintroduce the eager by-ref dereference.

## Evidence

| Check | Baseline (`d0d0ebc5`) | Head (`e0d919c3`) |
| --- | --- | --- |
| Product build (`src/dotnet-inspect`) | Pass | Pass |
| Focused tests (`FoldGuardReturnFidelityTests`) | 19 passed | 27 passed (+8) |
| Related passes (`ShortCircuitTernaryPassTests`, `IrImporterTests`, `LoweringCoverageTests`) | Pass | 92 passed |
| Decompiler fast suite (`-trait- "Speed=Slow"`) | 3614, 2 failed (env) | 3614, 2 failed (env) |
| Real witness (`WitnessNs.Hazard::ByRefBeforeCall`) | folds to eager `start <= 0 && r && Call()` | declines; branch preserved |

The two fast-suite failures are pre-existing and environmental, unchanged by this PR:

- `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString` — CRLF vs LF line-ending mismatch in the expected literal.
- `ExpressionTreeLambdaTests.LookalikeExpressionAssembly_StaysFactoryCalls` — `System.Linq.Expressions.dll` spoof fixture not built in the worktree.

Neither touches boolean folding.

## Decompiler quality

**Conclusion:** **PASS** — the change only *narrows* an existing fold (adds a decline predicate). It introduces no new lowering, raising, or structuring, so it cannot add lowering residue, conditional-branch residue, or pass bugs. It can at most decline a fold for the exact chain-buried by-ref shape, which is the intended, correctness-preserving outcome. A full corpus gate was not run because there is no new raise to measure; the affected shape is exercised by the added fidelity unit tests and the real witness above.

## Scope

- Only a managed by-ref (`LoadIndirect` whose address is `TypeRefKind.ByRef`) is flagged. A raw pointer deref, a bitwise `&`/`|` (`Binary`) operand, a different-kind logical sub-expression, a comparison, and a by-ref confined to a call argument all keep their branch and still fold (covered by four accept tests).
- A chain-buried bare **local** remains an opcode-only (non-soundness) gap, pre-existing on main and out of scope here.
- CLI output is otherwise unchanged.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.FoldGuardReturnFidelityTests"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/FoldGuardReturnFidelityTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FoldGuardReturnFidelityTests.cs
@@ -63,6 +63,15 @@ public class FoldGuardReturnFidelityTests
     static LoadIndirect ByRefBoolDeref()
         => new(s_bool, new LoadArgument(1, "y", TypeRef.ByRef(s_bool)));
 
+    // A bool-returning call — a side-effecting barrier csc will not collapse past.
+    static Call BoolCall()
+        => new(new MethodRef(s_holder, "Call", s_bool, [], HasThis: false), isVirtual: false, []);
+
+    // A bool-returning call taking one bool argument (a by-ref deref confined to a
+    // call argument stays guarded and folds faithfully).
+    static Call BoolCallWith(IrExpression argument)
+        => new(new MethodRef(s_holder, "Sink", s_bool, [s_bool], HasThis: false), isVirtual: false, [argument]);
+
     // A raw pointer bool dereference: can access-violate, so csc keeps the branch and
     // the raise stays opcode-exact — the negative boundary of the by-ref guard.
     static LoadIndirect PointerBoolDeref()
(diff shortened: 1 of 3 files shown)


richlander:

Two independent fixed-head reviews at the exact head **`e0d919c3`** — both **CLEAN**, no blocking or non-blocking findings, no fixes required. Each reviewer worked in an isolated review worktree with scratch under the OS temp dir and no git mutations.

### Reviewer 1 — GPT (`gpt-5.6-sol`) — CLEAN

- Product build clean; `FoldGuardReturnFidelityTests` 27/27; decompiler fast suite 3614 total, 2 failed — only the two documented pre-existing environmental reds (`StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString` CRLF/LF; `ExpressionTreeLambdaTests.LookalikeExpressionAssembly_StaysFactoryCalls` missing fixture DLL).
- Reproduced the end-to-end witness: HEAD preserves the `if (start <= 0) { return r && Call(); } return false;` branch (declines).
- Independently probed the constant-on-left asymmetry (`start <= 0 && true == r && Call()`): although HEAD folds that shape, recompilation retained the guarding branch and a **null-by-ref runtime probe returned `False`** (no NRE) — confirmed not an exploitable soundness gap.

### Reviewer 2 — Gemini Pro (`gemini-3.1-pro-preview`) — CLEAN

- Verified via `csc` IL that `outerKind` mirrors the operator `FoldGuardReturn` emits for all four single-constant arms (`Or` when the surviving-arm-selecting constant is true, else `And`).
- Confirmed the by-ref-only restriction is sound: raw pointer `*p`, array `arr[0]`, struct field `r.b`, and bitwise `&` all compile to a **kept branch** (`ble.s`/`brfalse.s`); csc's branchless collapse genuinely fires only for a managed by-ref deref (`TypeRefKind.ByRef`).
- Confirmed constant-on-left (`true == X`) is safe — the comparison stays a compound barrier that keeps the branch (same IL evidence).
- Confirmed the peel's negation-parity omission can only **over-decline** (a safe readability sacrifice), never under-decline; verified `!r` compiles to a branch-preserving `ceq`.
- Confirmed recursion terminates (strictly decreasing AST depth). 27/27 focused tests pass on HEAD.

### Reconciliation

Both reviewers independently converged on the one non-obvious point — the constant-on-left asymmetry — and each demonstrated (Gemini via IL, GPT via a runtime probe) that it is a safe over-approximation, not a hole. No findings were raised by either reviewer, so there is nothing to attribute as fixed or dismissed. Reviewers are from two different model families per the roster; neither is my own family.

Ready to merge

PR #3149, opened 2026-07-25
Parenthesize right-nested same-kind && / || chains (#3126)

### LINKED ISSUE #3126: ShortCircuitTernaryPass: same-kind logical chain in when-false arm can diverge in opcodes when raised to short-circuit
## Summary

`ShortCircuitTernaryPass.RendersAsBranchlessBarePlace(operand, outerKind)` guards the branchless-collapse hazard for a **bare** local/arg/stack-slot operand only at the **root** of the when-false arm, and delegates only the *by-ref* hazard to the outerKind-flattening `ShortCircuitFidelity.LiftEagerlyDerefsByRef`. It does **not** flatten a same-kind logical chain to look for a bare **local/argument** nested inside it. So when the when-false arm is a same-kind `LogicalBinary` chain, raising `c ? true : (a || Call())` to `c || (a || Call())` — which the printer flattens to `c || a || Call()` — lets csc collapse the `c | a` prefix branchless, diverging in opcodes from the ternary's original block layout.

This is a pre-existing **opcode-fidelity** gap (the pass's contract is opcode-exact recompile), orthogonal to the by-ref soundness scope of #3114. It is **not a soundness bug** — the raised C# is valid and behaviorally identical; only the recompiled opcodes differ. CI is green on `main`, so the product corpus does not currently hit this shape.

Found during #3119 (issue #3114) round-5 adversarial review (GPT).

## Evidence (SDK-csc `/optimize` IL, direct byte diff)

For `c ? true : y` raised to `c || y`, comparing the original ternary IL vs recompile of `c || y`:

- **DIVERGE**: `y ∈ { a || Call(), a && Call() (B-form `c ? false : y`), Call() || (X() > 0) }`
- **IDENTICAL**: `y ∈ { Call(), X() > 0, Call() || Call2(), a && Call() (different-kind under Or lift), a || Call() (different-kind under And lift), Call() && Call2() }`

The divergence is **input-sensitive csc block layout**, broader than simply "a bare local nested in a same-kind chain" — a naive "flatten and decline any nested bare local/arg" guard is INCOMPLETE (it would still miss the compound `Call() || (X() > 0)` divergence, which contains no bare local) and would also over-decline the faithful `Call() || Call2()`. A correct fix needs a more precise characterization of csc's true/false-block merge/layout for chained short-circuit operands in the when-false arm.

## Scope / suggested direction

- Confined to `ShortCircuitTernaryPass` (added in #3107). The sibling guarded-return fold (`BooleanFoldingPass.FoldGuardReturn`, #3114) is unaffected: its decision is correctness-only (it *keeps* opcode-divergent-but-valid readability raises), so a same-kind-chain opcode divergence is an accepted readability raise there, not a bug.
- A conservative fix: decline the ternary raise when the when-false arm is a same-kind `LogicalBinary` chain (over-declining the faithful `Call() || Call2()` case). Sound for opcode-exactness but needs its own focused PR + corpus A/B + adversarial review, since it narrows folding.
- Requires pinning the exact csc block-layout rule with fresh probes before implementing.

richlander:
- Fixes #3126
- Changes the C# printer to keep the parentheses on a right-nested same-kind `&&`/`||` chain (`a && (b && c)`) instead of flattening it to `a && b && c`, so the reprint recompiles to the original opcode stream.
- Card revision: `41c11845`

## Change

**Conclusion:** **PASS** — the printer flattened right-nested same-kind short-circuit chains, silently re-associating them; adding the parens back is monotonic (parens around an already-valid bool subexpression can never fail to compile) and eliminates the opcode divergence.

### Corrected premise

Issue #3126 was filed from the #3114 round-5 review as a `ShortCircuitTernaryPass` opcode-fidelity gap. **That evidence was wrong.** csc constant-folds a constant-arm ternary (`c ? false : y` → `!c && y`, `c ? true : y` → `c || y`) at compile time, so no ternary diamond survives to `ShortCircuitTernaryPass` for these shapes. The divergence reproduces for hand-written non-ternary `&&` chains and for bare-bool conditions where that pass cannot fire. The real cause is the **printer** flattening a right-nested same-kind chain, which changes associativity: C# `&&`/`||` are left-associative, so `a && (b && c)` reprinted as `a && b && c` reparses as `(a && b) && c` and csc lays that out with a different branch structure (different IL bytes).

### Original source

```csharp
public static bool SB_callAndCall(bool c) => !c && (Call() && Call2());
```

### Before

```csharp
return !c && Call() && Call2();
```

- Valid: True
- Correct: True

Valid and behavior-preserving, but the flattened form re-associates to `(!c && Call()) && Call2()`; csc emits a different branch structure for it, so the reprint recompiles to a divergent opcode stream (fidelity below `Full`-exact).

### After

```csharp
return !c && (Call() && Call2());
```

- Valid: True
- Correct: True

The parens preserve the original right-nested associativity, so the reprint recompiles to the original opcode stream (opcode-exact).

### Fully raised

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `PrinterPrecedenceTests`: 24 passed | `PrinterPrecedenceTests`: 24 passed (4 new #3126) |
| Synthetic fidelity corpus (38 methods, incl. real csc-compiled ternary-else `&&`/`||`) | 9 opcode diffs (29/38 exact) | 0 opcode diffs (38/38 exact) |
| Real-asm fidelity `ILInspector.Metadata.dll` (2292 methods) | 1736 exact / 336 opcode / 17 operand / 180 recompile-fail | 1736 / 336 / 17 / 180 (identical) |
| Real-asm fidelity `ILInspector.Analysis.dll` (978 methods) | 754 exact / 143 opcode / 13 operand / 64 recompile-fail | 754 / 143 / 13 / 64 (identical) |
| Decompiler fast suite | 2 pre-existing env reds | 2 pre-existing env reds (unchanged) |

The two pre-existing fast-suite reds are environmental and unchanged by this PR: `SmallStringSwitch_RendersSwitchOnString` (CRLF-vs-LF) and `LookalikeExpressionAssembly_StaysFactoryCalls` (missing fixture when building only the test project).

The right-nested same-kind shape does not occur in the two sampled real assemblies (real code is overwhelmingly left-nested from left-associative source), so the real-asm A/B is neutral — no regression. The value is demonstrated on the synthetic corpus, whose `RB_*` fixtures are **real csc-compiled** ternary-else `&&`/`||` methods, proving the shape is reachable from real input.

## Decompiler quality

**Conclusion:** **PASS** — neutral by construction. This is a printer-only change; the IR structure is unchanged, so lowering-residue / conditional-branch / forward-merge counts are identical base-vs-head. The meaningful signal is compile-back fidelity (synthetic 9→0, real-asm neutral), reported in the Evidence table above.

## Validation

```bash
dotnet build tools/DecompilerHarness/DecompilerHarness.csproj -c Release --configfile ./nuget.config
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.PrinterPrecedenceTests -trait- "Speed=Slow"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/PrinterPrecedenceTests.cs
+++ b/src/ILInspector.Decompiler.Tests/PrinterPrecedenceTests.cs
@@ -82,6 +82,97 @@ public class PrinterPrecedenceTests
         Assert.DoesNotContain("a ? b : c ? d", output);
     }
 
+    // ---- #3126: right-associative same-kind && / || chains keep their parens ----
+    // C# `&&`/`||` are left-associative, so a right-nested same-kind chain
+    // `a && (b && c)` must keep its parens: dropping them reparses as the
+    // left-nested `(a && b) && c`, which csc lays out with a different branch
+    // structure — the reprint would recompile to a divergent opcode stream. (Not
+    // reachable as a top-level ternary from C# source — csc folds a constant-arm
+    // ternary to `&&`/`||` at compile time — so the printer is exercised over the
+    // hand-built right-nested IR the boolean folds reconstruct.)
+
+    [Fact]
+    public void LogicalAnd_RightNestedSameKind_StaysParenthesized()
+    {
+        // a && (b && c) — right-nested `&&`.
+        var chain = new LogicalBinary(
+            LogicalKind.And,
+            new LoadArgument(0, "a", s_bool),
+            new LogicalBinary(
+                LogicalKind.And,
+                new LoadArgument(1, "b", s_bool),
+                new LoadArgument(2, "c", s_bool)));
+
+        var output = PrintReturn(
+            chain, s_bool,
+            [new Parameter("a", s_bool), new Parameter("b", s_bool), new Parameter("c", s_bool)]);
+
+        Assert.Contains("return a && (b && c);", output);
+    }
+
+    [Fact]
+    public void LogicalOr_RightNestedSameKind_StaysParenthesized()
+    {
+        // a || (b || c) — right-nested `||`.
+        var chain = new LogicalBinary(
+            LogicalKind.Or,
+            new LoadArgument(0, "a", s_bool),
+            new LogicalBinary(
+                LogicalKind.Or,
+                new LoadArgument(1, "b", s_bool),
+                new LoadArgument(2, "c", s_bool)));
+
+        var output = PrintReturn(
+            chain, s_bool,
+            [new Parameter("a", s_bool), new Parameter("b", s_bool), new Parameter("c", s_bool)]);
+
+        Assert.Contains("return a || (b || c);", output);
+    }
+
+    [Fact]
+    public void LogicalAnd_LeftNestedSameKind_StaysFlat()
+    {
+        // (a && b) && c — left-nested `&&`. C# `&&` is left-associative, so this
+        // is opcode-identical to the flat `a && b && c` and must NOT gain parens
+        // (regression guard against over-parenthesizing the common left-nested
+        // chain, which is what most compiled `&&` lowerings produce).
+        var chain = new LogicalBinary(
+            LogicalKind.And,
+            new LogicalBinary(
+                LogicalKind.And,
+                new LoadArgument(0, "a", s_bool),
+                new LoadArgument(1, "b", s_bool)),
+            new LoadArgument(2, "c", s_bool));
+
+        var output = PrintReturn(
+            chain, s_bool,
+            [new Parameter("a", s_bool), new Parameter("b", s_bool), new Parameter("c", s_bool)]);
+
+        Assert.Contains("return a && b && c;", output);
+        Assert.DoesNotContain("(a && b)", output);
+    }
+
+    [Fact]
+    public void LogicalAnd_RightNestedDifferentKind_StaysParenthesized()
+    {
+        // a && (b || c) — a different-kind chain already parenthesizes on either
+        // side (unchanged by #3126); this pins that the right-operand rule did not
+        // regress the mixed-kind case.
+        var chain = new LogicalBinary(
+            LogicalKind.And,
+            new LoadArgument(0, "a", s_bool),
+            new LogicalBinary(
+                LogicalKind.Or,
+                new LoadArgument(1, "b", s_bool),
+                new LoadArgument(2, "c", s_bool)));
+
+        var output = PrintReturn(
+            chain, s_bool,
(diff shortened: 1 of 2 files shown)


richlander:
## Adversarial review reconciliation — head `41c11845`

Two-model review (required: behavior change to the core C# printer). Both reviewers ran against the exact head `41c11845` in isolated review worktrees, built the harness, and reproduced empirically.

### Gemini (`gemini-3.1-pro-preview`) — CLEAN

Verified all six attack points with code tracing + runs; no issues:

1. **Over-parenthesizing** — left-nested `And(And(a,b),c)` evaluates the left child with `rightOperand:false` → stays `a && b && c`. No regression.
2. **Fluent multi-line wrap** — for left-nested chains the flattened join equals `flat`, so the exact-match guard still proceeds with the multi-line wrap; only right-nested chains (where `flat` now has parens) correctly decline to the inline parenthesized form.
3. **Invalid output** — parens are only ever added around a `LogicalBinary` (a `bool` expression); always valid C#.
4. **Still-divergent** — parenthesized right-nested render restores the original IR tree shape → opcode-exact.
5. **Missed call site** — all three `LogicalOperandText` call sites route correct `rightOperand` values (`LogicalText` left=false/right=true; `LogicalChainLines`=false, where the flat-string guard enforces correctness).
6. **Precedence interaction** — a right-nested chain under an outer different-kind logical hits the mixed-kind fallback and gets exactly one paren set.

### GPT (`gpt-5.6-sol`) — 1 Medium finding, **pre-existing / out of scope**

> Property-pattern fold (`o is P p && (p.X == 1 && p.Y == 2)` → `o is P { X: 1, Y: 2 }`) recompiles to a divergent opcode stream for the right-nested conjunction.

- **Explicitly not caused by this diff** — the reviewer confirmed base reproduces it; I independently verified **base == head byte-identical** (both render `return o is P { X: 1, Y: 2 };`, both 1 opcode diff). This is a separate code path (`TryPropertyPatternText` returns *before* the new `rightOperand` logic in `LogicalText`), not the short-circuit-chain associativity flattening this PR fixes.
- **Non-action here (scope):** the property-pattern-fold vs explicit-comparison-chain divergence is a distinct fold with its own fidelity/readability tradeoffs; folding it into this surgical printer change would expand scope. Filed as follow-up **#3151**. The reviewer also noted head *improves* other property-pattern fixtures (does not regress any).

### Verdict

Both fixed-head reviews are clean with respect to this PR: no regressions, no invalid output, opcode-exact on the target shape. The single finding is pre-existing and tracked in #3151.

richlander:
Ready to merge

PR #3073, opened 2026-07-23
Fully raise flags-enum accumulator slots to the enum type (#3009 sub-part 1)

richlander:
- Advances #3009 (sub-part 1: enum slot typing)
- Changes: a spilled 64-bit flags-enum accumulator slot now materializes as the enum instead of `long` with per-use casts
- Card revision: `8ca6e44b`

## Change

> Should we accept this change?

**Conclusion:** **PASS** — a targeted slot-testimony rule retypes an enum-only accumulator slot to the enum; the `long` declaration, `(long)512`, and per-use `(FlagCaps64)` cast all collapse, with 0 new validity/corpus/fidelity diffs.

### Original source

```csharp
public static int Accumulate(FlagCaps64 server, bool interactive)
{
    FlagCaps64 caps = FlagCaps64.Protocol
        | (interactive ? (server & FlagCaps64.Interactive) : FlagCaps64.None)
        | (server & FlagCaps64.LoadLocal)
        | FlagCaps64.Secure
        | (server & FlagCaps64.MultiStatements)
        | FlagCaps64.MultiResults;
    return (int)caps;
}
```

### Before

```csharp
public static int Accumulate(ILInspector.Decompiler.Tests.FlagCaps64 server, bool interactive)
{
    long S_0 = (long)512;
    FlagCaps64 S_1 = interactive ? (server & FlagCaps64.Interactive) : FlagCaps64.None;
    return (int)((FlagCaps64)S_0 | S_1 | server & FlagCaps64.LoadLocal | FlagCaps64.Secure | server & FlagCaps64.MultiStatements | FlagCaps64.MultiResults);
}
```

- Valid: True
- Correct: True

Valid and correct after #2990, but lower quality: the accumulator slot is declared `long`, its `512` seed renders as `(long)512`, and its single use is cast back to the enum (`(FlagCaps64)S_0`).

### After

```csharp
public static int Accumulate(ILInspector.Decompiler.Tests.FlagCaps64 server, bool interactive)
{
    FlagCaps64 S_1 = interactive ? (server & FlagCaps64.Interactive) : FlagCaps64.None;
    return (int)(FlagCaps64.Protocol | S_1 | server & FlagCaps64.LoadLocal | FlagCaps64.Secure | server & FlagCaps64.MultiStatements | FlagCaps64.MultiResults);
}
```

- Valid: True
- Correct: True

[...]

Issue #3067, opened 2026-07-23
Decompiler taste: pretty-print long splittable expressions (&& / || chains) across continuation lines


richlander:
## Summary

Add an opt-in printer taste option that pretty-prints long, splittable
expressions — short-circuit `&&`/`||` boolean chains first — by breaking each
operand onto its own continuation line when the single-line form is too wide,
instead of emitting one very long line.

## Motivation

PR #3061 (#3051) recomposes short-circuit `&&` boolean-return chains that a
spilled struct receiver had fragmented into a guard cascade. That is the right
shape, but for a long chain the result is a single very wide line, e.g. the
NodaTime `BclAdjustmentRule::IsStandardOffsetOnlyRule` witness:

```csharp
return daylight.IsFixedDateRule && daylight.Day == 1 && daylight.Month == 1 && daylight.TimeOfDay.TimeOfDay < TimeSpan.FromMinutes(1d) && standard.IsFixedDateRule && standard.Day == 1 && standard.Month == 1 && standard.TimeOfDay.TimeOfDay < TimeSpan.FromMinutes(1d);
```

The original source wraps this across lines:

```csharp
return daylight.IsFixedDateRule && daylight.Day == 1 && daylight.Month == 1 &&
       daylight.TimeOfDay.TimeOfDay < TimeSpan.FromMinutes(1) &&
       standard.IsFixedDateRule && standard.Day == 1 && standard.Month == 1 &&
       standard.TimeOfDay.TimeOfDay < TimeSpan.FromMinutes(1);
```

A wrapping taste option would let the recomposed chain read the way the source
does.

## Precedent

This is the boolean-operator analog of the existing fluent-chain wrapper, which
already does exactly this for `.Member(args)` chains:

- `FluentChainLines` / `FluentChainWrapWidth = 120` / `FluentChainMinSegments`
  in `src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.Members.cs` (~line
  352). It breaks a chain onto one segment per line under a continuation indent
  once the single-line form would exceed the dotnet/runtime max line width, and
  is documented as a **taste class 3** (no IL anchor) change: a "pure formatting
  tiebreaker" that "never changes which tokens are emitted", spliced out of the
  single-line text by length arithmetic so the broken form is **token-identical**
  and the IL is unchanged.

The new option should share that discipline: token-identical output, opcode
preservation, off by default.

## Proposed shape

- Add a knob to `PrinterOptions`
[...]

PR #3150, opened 2026-07-25
Wrap long bitwise |/&/^ chains one operand per line (#3009 sub-part 3)

richlander:
## What

Final sub-part of #3009. The opt-in `PrinterOptions.WrapSplittableExpressions` taste now also breaks a long associative bitwise `|`/`&`/`^` chain **one operand per continuation line, with the operator leading** each broken line (the flags-accumulation house style — the opposite placement from the short-circuit `&&`/`||` wrapper added in #3067).

With sub-parts 1 (#3073, enum slot typing) and 2 (#3133, slot inlining) merged, the fully-raised flags-enum accumulator now renders, with the option on, as:

```csharp
return (int)(FlagCaps64.Protocol
    | (interactive ? (server & FlagCaps64.Interactive) : FlagCaps64.None)
    | server & FlagCaps64.LoadLocal
    | FlagCaps64.Secure
    | server & FlagCaps64.MultiStatements
    | FlagCaps64.MultiResults);
```

## How

`AppendStatement` gains a third `TryAppendBitwiseChain` fallback (after fluent- and short-circuit-chain wrapping). Unlike the short-circuit wrapper (which reconstructs a statement prefix), the bitwise wrapper **splices** the broken chain into the already-rendered statement `line` in place, so it handles a bare chain (`return a | b | …`), an assignment, and a chain inside a value-preserving cast (`return (int)(a | b | …)` — common when a fully-raised flags-enum accumulation is returned as its underlying integer) uniformly.

It declines (renders inline) unless:
- the per-operand re-render reproduces the flat `BinaryText(root)` exactly (guards the sibling-context-dependent enum-coercion / mixed-sign / unchecked special cases), **and**
- that flat chain text occurs exactly once in the line (unambiguous splice), **and**
- the whole statement exceeds `FluentChainWrapWidth` (120).

Only the interior separators of the located chain are broken, so the wrapped form is **token-identical** to the inline line — only whitespace differs, IL unchanged.

## Safety / blast radius

- **Off by default.** The default render path is byte-identical: `TryAppendBitwiseChain` returns immediately when `WrapSplittableExpressions` is false.
- Corpus + Validity gates (default options): 175/175 pass, no golden churn.

## Tests

- `BitwiseChainWrapTests` (new): leading-operator break for `|` and `&`, inline fallback below width / short chain / default options, token identity, taste decision emission, recompilation of the wrapped output.
- `FlagsEnumAccumulatorTests`: two end-to-end assertions — the wrapped multi-line form (the #3009 target) and default-options-stays-inline + token identity.
- Registered the new `CSharpCompilation.Create` site in the Dynamic census manifest (`DynamicCompilationSiteInventoryTests`).

## Validation

- Fast suite: 3642 pass, 1 fail — `ExpressionTreeLambdaTests.LookalikeExpressionAssembly_StaysFactoryCalls`, a pre-existing environmental failure when only the test csproj is built (missing fixture assemblies); confirmed identical on the stashed base.
- Corpus + Validity areas: 175/175.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/BitwiseChainWrapTests.cs
+++ b/src/ILInspector.Decompiler.Tests/BitwiseChainWrapTests.cs
@@ -1,0 +1,189 @@
+using System.Collections.Immutable;
+using ILInspector.Decompiler.Pipeline;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace ILInspector.Decompiler.Tests;
+
+// #3009 sub-part 3: the opt-in PrinterOptions.WrapSplittableExpressions taste knob
+// also breaks a long associative bitwise |/&/^ chain one operand per continuation
+// line — but with the operator LEADING each broken line (the flags-accumulation
+// house style), the opposite placement from the short-circuit &&/|| wrapper.
+// Wrapping is whitespace-only: the broken body still compiles and carries the same
+// tokens as the inline form. Off by default.
+public sealed class BitwiseChainWrapTests
+{
+    static readonly TypeRef Holder = TypeRef.Definition("synthetic", "", "Holder");
+    static readonly TypeRef Int = TypeRef.CoreLib("System", "Int32");
+
+    static readonly string[] Flags =
+    [
+        "firstCapabilityBit",
+        "secondCapabilityBit",
+        "thirdCapabilityBit",
+        "fourthCapabilityBit",
+        "fifthCapabilityBit",
+        "sixthCapabilityBit",
+    ];
+
+    static LoadArgument Flag(int index) => new(index, Flags[index], Int);
+
+    // firstCapabilityBit | ... | sixthCapabilityBit — a left-associative chain well
+    // past the 120-column wrap width once `return ...;` is added.
+    static Binary LongChain(BinaryKind kind = BinaryKind.Or)
+    {
+        IrExpression chain = Flag(0);
+        for (int i = 1; i < Flags.Length; i++)
+            chain = new Binary(kind, isChecked: false, isUnsigned: false, chain, Flag(i));
+        return (Binary)chain;
+    }
+
+    static string Render(IrExpression returnValue, PrinterOptions? options = null)
+        => CSharpPrinter.Print(Function(returnValue), options).Output!;
+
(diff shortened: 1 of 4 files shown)


richlander:

Two fixed-head reviews at `c2c20492` (I'm Opus → Gemini + GPT, both latest):

- **GPT (`gpt-5.6-sol`)** — isolated worktree `review-3150-gpt`: **clean**, "No significant issues found in the reviewed changes."
- **Gemini (`gemini-3.1-pro-preview`)** — isolated worktree `review-3150-gemini`: **clean**. Verified all six attack points I posed:
  1. **Token-identity/splice** — `BinaryText(root)` runs the same `BinaryBody` logic as inline `Statement(node)`; the `string.Join(" | ", texts) == chainFlat` guard proves the flattened join reproduces the tree-evaluated flat text, so recompiled tokens are identical.
  2. **Unique-occurrence** — the root is the outermost chain of a `Return`/`StoreLocal`/`StoreStackSlot`; C# identifier/keyword syntax can't reproduce `chainFlat` elsewhere in the line, so `IndexOf == LastIndexOf` pinpoints the one occurrence.
  3. **Convert-unwrap** — `ConvertText` renders the operand at `Precedence.Primary`, so a cast-wrapped chain is spelled `(int)(a | b | c)` and `chainFlat` is located verbatim inside the parens.
  4. **Precedence/rightSide** — a right-leaning `a | (b | c)` IR renders flat with parens; the join produces `a | b | c`, so the inequality **safely declines** rather than mis-splicing.
  5. **Default path** — option off ⇒ early `return false`, byte-identical to base.
  6. **Width/min-segment** — ≥2 operands enforced; compound `V_0 |= b | c` can't contain `V_0 | b | c`, so it bypasses cleanly.

No blocking findings from either reviewer; no code changes were needed after review. Both review worktrees were isolated at the exact head with scratch confined to `/tmp` (no `git add`/`reset`/commit).

### Evidence
- Fast suite: 3642 pass, 1 fail = `ExpressionTreeLambdaTests.LookalikeExpressionAssembly_StaysFactoryCalls` — pre-existing environmental failure when only the test csproj is built (missing fixture assemblies); confirmed identical on the stashed base.
- Corpus + Validity areas: 175/175 (default output unaffected — opt-in feature).
- CI: `build-net10`, `csharp-diff-smoke`, `test (ubuntu-26.04, linux-x64)` all green.

Ready to merge

PR #3154, opened 2026-07-25
Split package/PDB transport and archive parsing from filesystem storage (#3147)

### LINKED ISSUE #3147: Split transport from filesystem storage in DotnetInspector.Packages so download/snupkg/resolution logic is reusable in non-desktop (WASM) hosts
## Summary

`DotnetInspector.Packages` couples network transport to desktop **filesystem** storage, so non-desktop hosts (notably the browser/WASM `inspect-web` prototype) cannot reference it and are forced to reimplement NuGet version resolution, package download/caching, and snupkg/PDB acquisition. Split transport + archive parsing from persistence so this logic is reusable across hosts.

## Evidence: what's reusable vs. filesystem-bound

The inspection core is cleanly reusable and **is** reused by the WASM engine today (`ILInspector.Metadata`, `ILInspector.Decompiler`, `ILInspector.Analysis`, `ILInspector.CallGraph`). The package layer is not:

- `src/DotnetInspector.Packages/NuGetCache.cs` — reads/writes `~/.nuget/packages`, app-cache dirs via `Environment.GetFolderPath`, `Directory.*`, `File.*`, and commit-marker files. Entirely filesystem-shaped.
- `src/DotnetInspector.Packages/SymbolPackageDownloader.cs` — ctor takes an injectable `HttpClient` (transport **is** abstracted — good), but it hardcodes `_cachePath = Path.Combine(NuGetCache.GetAppCachePath(), "symbols")` and does `Directory.CreateTempSubdirectory("snupkg")`, `File.Create`, `ZipFile.ExtractToDirectory`, `File.Copy`. The snupkg parsing + CodeView-key→PDB logic is valuable and host-neutral, but it's welded to disk I/O.

None of this maps to a browser WASM sandbox (no persistent filesystem, no user profile).

## Duplication this forces

The WASM engine reimplements, in `prototypes/inspect-web/engine/Program.cs`, browser-shaped equivalents against `HttpClient`/`fetch` + an in-memory LRU:

- `ResolvePackageVersionAsync` — hand-rolled flat-container `index.json` fetch + latest-stable selection → duplicates `PackageVersionVector` / `VersionDirectory`.
- `GetPackageBytesAsync` + bespoke LRU `PackageCache` → duplicates `NuGetCache`.
- `TryAcquirePackagePdbAsync` (snupkg download + zip entry lookup) → duplicates `SymbolPackageDownloader` / `PackageExtractor`.
- `WriteEntryAsync` → `PackageExtractor` territory.

## Proposal

Introduce a storage abstraction (e.g., `IPackageStore` / `IPdbCache`) that separates:

1. **Transport + archive parsing** — flat-container download, version-index parsing, snupkg entry lookup, CodeView-key→PDB resolution. Host-neutral; keep `HttpClient` injectable.
2. **Persistence** — filesystem (desktop/CLI) vs. in-memory (WASM), provided by the host.

Then the WASM engine can supply an in-memory store and drop its hand-rolled `ResolvePackageVersionAsync` / `GetPackageBytesAsync` / `TryAcquirePackagePdbAsync`.

Constraints to respect (per `AGENTS.md`): SRM-only, NativeAOT-friendly, Roslyn-free, no inspected-assembly loading; behavior-safe defaults / progressive disclosure preserved.

## Explicit non-goal

This does **not** enable fetching PDBs from the Microsoft symbol server (MSDL) in the browser. That is blocked by CORS regardless of layering: MSDL answers with a cross-origin `302` to a SAS-signed, expiring, non-guessable Azure blob, and the `302` itself carries no `Access-Control-Allow-Origin`, so a browser `fetch` in cors mode aborts with `Failed to fetch` before reaching the blob (verified). MSDL-backed PDBs must be precomputed at build/publish time and shipped as same-origin static assets. This issue is purely about code reuse/layering so a non-desktop host can reuse the transport + snupkg/resolution logic instead of reimplementing it.

## Acceptance

The WASM engine can reference the shared package library with an in-memory storage provider and delete its duplicated version-resolution, package-download/cache, and snupkg/PDB-acquisition code, with no behavior change to the desktop/CLI path.

richlander:
Fixes #3147.

## What

Introduces byte/stream-oriented **storage seams** in `DotnetInspector.Packages` so a browser/WASM host can reuse NuGet version resolution, package download/caching, and snupkg/PDB acquisition with an **in-memory** store — without reimplementing them and **without changing desktop/CLI behavior**.

The prior duplication motivating this: the WASM prototype (`prototype/wasm-command-ui`) reimplemented version resolution, package byte-fetch/LRU, and snupkg→PDB extraction because those were welded to the filesystem.

## Design

Both hosts share transport (HttpClient) and archive parsing; they differ only in persistence. The seam is the persistence + content-read boundary.

New host-neutral pieces:
- **`SnupkgPdbReader`** — parses a `.snupkg` entirely in memory (`ZipArchive` over a stream), matches the assembly PDB by file name, validates header (`BSJB`/`Micr`) + debug GUID, returns bytes. Removes the temp-dir + `ZipFile.ExtractToDirectory` coupling from symbol acquisition.
- **`IPdbStore`** (+ `FileSystemPdbStore` / `InMemoryPdbStore`) — stream-oriented PDB persistence. The filesystem impl preserves the exact symbol-cache key layout; its per-segment guard permits interior dots (real PDB names like `System.Text.Json.pdb`) while rejecting traversal segments (`.`, `..`, separators, null).
- **`IPackageStore` / `IPackageContent`** (+ FileSystem / InMemory impls) — `FileSystemPackageStore` delegates to `NuGetCache` (unchanged cache lookup, transactional commit, returned paths); `InMemoryPackageStore` caches nupkg bytes and reads entries from an in-memory `ZipArchive`.

Desktop orchestration preserved:
- **`SymbolPackageDownloader`** keeps MSDL/symbol-server/miss orchestration and returns on-disk PDB paths; it now routes snupkg parsing through `SnupkgPdbReader` and PDB caching through `IPdbStore`.
- **`PackageExtractor`** keeps version resolution + HTTP transport (stream-to-file download with retry) and reaches cache lookup/commit through `FileSystemPackageStore`. Observable paths, `PackageLoad` telemetry, in-flight dedupe, offline errors, and tool-wrapper redirects are unchanged.

## Behavior-change boundary

Desktop is behavior-preserving. The one intentional difference on the download path: the downloaded temp `.nupkg` is handed to the store as a `FileStream` (nothing buffered in memory) and the filesystem store copies it once into its own temp workspace before extract + transactional commit — a negligible transient local copy; the committed cache entry, its marker, and all returned paths are byte-identical.

## Tests

- New: `SnupkgPdbReaderTests` (match / GUID-mismatch / Windows PDB / no-name / nested-entry, built from a real `PortablePdbBuilder`), `PdbStoreTests` (round-trip, local-path, dotted-name allowance, traversal rejection), `PackageStoreTests` (in-memory round-trip + entry reads, latest-version-ignores-prerelease, filesystem materialize + cache).
- Unchanged and green: `SymbolPackageDownloaderTests`, `PackageExtractorOfflineTests`, `PackageAcquisitionConcurrencyTests`, `CacheCommandTests`, and Services package/version suites.
- Full `dotnet-inspect.slnx -c Release` builds clean.

## Docs

`docs/design/untrusted-data-threat-model.md` updated: symbol PDB acquisition is now in-memory (no `.snupkg` extraction to disk; entries matched by file name only), and PDB cache key segments pass the `FileSystemPdbStore` guard.

## Review

Non-trivial architecture change → adversarial review by two non-Claude model families to follow.

### DIFF
--- a/docs/design/untrusted-data-threat-model.md
+++ b/docs/design/untrusted-data-threat-model.md
@@ -75,8 +75,10 @@ bytes are then persisted through `IPdbStore`; the filesystem implementation
 disk, so no archive-entry name is ever used as an output path.
 
 Package identifiers and versions used as cache path components pass
-`NuGetCache.ValidatePathComponent`. Store keys (PDB cache keys and package entry
-paths) resolve through the shared `StorePath.ResolveUnderRoot` guard: it splits
+`NuGetCache.ValidatePathComponent`, which rejects empty or whitespace values,
+traversal (`..`), separators, volume qualifiers (`:`), null characters, and
+otherwise rooted values before any cache path is built. Store keys (PDB cache
+keys and package entry paths) resolve through the shared `StorePath.ResolveUnderRoot` guard: it splits
 on `/`, rejects any segment that is empty, `.`, `..`, separator-bearing,
 volume-qualified (`:`), null-character-bearing, or otherwise rooted, then
 verifies the composed absolute path stays under the store root with a final
--- a/docs/design/untrusted-data-threat-model.md
+++ b/docs/design/untrusted-data-threat-model.md
@@ -61,14 +61,25 @@ layers.
 
 ### Package archives use traversal-aware extraction
 
-NuGet package and symbol-package extraction uses
-`ZipFile.ExtractToDirectory`, which rejects archive entries that escape the
-destination directory. Extraction occurs under process-created temporary
-directories before selected content is copied into product caches.
+NuGet package extraction uses `ZipFile.ExtractToDirectory`, which rejects
+archive entries that escape the destination directory. Extraction occurs under
+process-created temporary directories before the validated content is committed
+into product caches (`FileSystemPackageStore.CommitAsync`).
+
+Symbol-package (`.snupkg`) PDB acquisition does not extract the archive to disk.
+`SnupkgPdbReader` opens the archive in memory, matches candidate entries by file
+name only (never by attacker-controlled directory paths), validates each
+candidate's PDB header and debug GUID, and returns the matching bytes. Those
+bytes are then persisted through `IPdbStore`; the filesystem implementation
+(`FileSystemPdbStore`) maps only store-composed, per-segment-validated keys onto
+disk, so no archive-entry name is ever used as an output path.
 
 Package identifiers and versions used as cache path components pass
-`NuGetCache.ValidatePathComponent`. General cache entries use SHA-256-derived
-keys through `CoreCache`.
+`NuGetCache.ValidatePathComponent`. PDB cache key segments pass
+`FileSystemPdbStore`'s per-segment guard, which rejects empty, `.`, `..`,
+separator, and null-character segments while permitting the interior dots of a
+real PDB file name. General cache entries use SHA-256-derived keys through
+`CoreCache`.
 
 Archive containment does not itself bound expanded bytes, entry count, or disk
 consumption. Resource budgets remain an open requirement below.
--- a/docs/design/untrusted-data-threat-model.md
+++ b/docs/design/untrusted-data-threat-model.md
@@ -75,11 +75,17 @@ bytes are then persisted through `IPdbStore`; the filesystem implementation
 disk, so no archive-entry name is ever used as an output path.
 
 Package identifiers and versions used as cache path components pass
-`NuGetCache.ValidatePathComponent`. PDB cache key segments pass
-`FileSystemPdbStore`'s per-segment guard, which rejects empty, `.`, `..`,
-separator, and null-character segments while permitting the interior dots of a
-real PDB file name. General cache entries use SHA-256-derived keys through
-`CoreCache`.
+`NuGetCache.ValidatePathComponent`. Store keys (PDB cache keys and package entry
+paths) resolve through the shared `StorePath.ResolveUnderRoot` guard: it splits
+on `/`, rejects any segment that is empty, `.`, `..`, separator-bearing,
+volume-qualified (`:`), null-character-bearing, or otherwise rooted, then
+verifies the composed absolute path stays under the store root with a final
+`Path.GetFullPath` containment check. This closes the Windows volume-reset
+vector where `Path.Combine(root, "C:..", ...)` would discard the root, while
+still permitting the interior dots of a real PDB or assembly file name. A PDB
+file name recovered from untrusted PE debug metadata that is not a usable single
+segment yields a graceful "no symbols" miss rather than an output path. General
+cache entries use SHA-256-derived keys through `CoreCache`.
 
 Archive containment does not itself bound expanded bytes, entry count, or disk
 consumption. Resource budgets remain an open requirement below.
--- a/src/DotnetInspector.Packages/NuGetCache.cs
+++ b/src/DotnetInspector.Packages/NuGetCache.cs
@@ -47,12 +47,24 @@ public static class NuGetCache
         ?? throw new InvalidOperationException("NuGetCache.Initialize(appName) must be called before using app cache methods.");
 
     /// <summary>
-    /// Validates that a value is safe to use as a path component (no traversal or separators).
+    /// Validates that a value is safe to use as a path component. Rejects empty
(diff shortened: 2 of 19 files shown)


richlander:

Two independent reviewers (non-Claude, per `AGENTS.md`): **GPT-5.6** and **Gemini 3.1 Pro**. Both reviewed the original head `7ede225d`, then re-reviewed each fixed head. Final result: **both reviewers clean on the current head `636eed04`.**

### Round 1 — head `7ede225d`

| # | Finding | Reviewer(s) | Resolution |
| - | ------- | ----------- | ---------- |
| 1 | **Critical:** Windows volume-reset path traversal — the per-segment guard rejected only exact `..`, so a `C:`/`C:..` segment (from an untrusted PE-embedded PDB name) let `Path.Combine` discard the store root → arbitrary read/write. | GPT + Gemini | **Fixed** in `9d0e304f`. New shared `StorePath.ResolveUnderRoot` rejects empty/`.`/`..`/separator/`:`/`\0`/rooted segments and adds a `Path.GetFullPath` containment check; used by both `FileSystemPdbStore` and `FileSystemPackageContent`. |
| 2 | `FileSystemPackageStore.CommitAsync` built a temp path from unvalidated coordinates before validation. | GPT | **Fixed** in `9d0e304f`. Validates coordinates first; uses a fixed temp filename. |
| 3 | Malformed untrusted PDB name threw `ArgumentException` before the network try/catch (was a graceful miss). | GPT | **Fixed** in `9d0e304f` (and refined in `636eed04`, see R2#1). `ClassifyStoredPdbAsync` catches `ArgumentException`. |
| 4 | `InMemoryPackageContent.TryOpenEntry` used ordinal `GetEntry`; desktop `File.Exists` is case-insensitive on Windows/macOS. | Gemini | **Fixed** in `9d0e304f`. Exact match, then `OrdinalIgnoreCase` fallback. |
| 5 | `PackageExtractor` hardcodes the filesystem store / `DownloadPdbAsync` success == on-disk path. | GPT | **By design, not a defect.** These are the desktop on-disk-path consumers (~40); the host-neutral reuse surface is `SnupkgPdbReader` + `IPackageStore`/`IPdbStore` directly. Reconciled with XML-doc/comment remarks in `9d0e304f`; no behavior change. |

Refuted/validated by Gemini as non-issues: symbol-cache key equivalence (unchanged lowercase segments), `SnupkgPdbReader` identity matching, `PackageExtractor` exception handling, transactional commit integrity, async disposal.

### Round 2 — head `9d0e304f`

Gemini: **clean.** GPT raised two new medium findings:

| # | Finding | Resolution |
| - | ------- | ---------- |
| 1 | The up-front unusable-PDB-name guard also skipped **snupkg** recovery, which is keyed off assembly name + debug GUID (not the PDB name). | **Fixed** in `636eed04`. Only the MSDL and NuGet symbol-server paths are gated on `pdbFileNameUsable`; the snupkg block runs unconditionally. |
| 2 | `NuGetCache.ValidatePathComponent` did not reject empty/whitespace coordinates → `CommitAsync("","",…)` could resolve to the cache root. | **Fixed** in `636eed04`. Validator now also rejects empty/whitespace, `:`, and rooted values; `InMemoryPackageStore.CommitAsync` rejects empty/whitespace too. |

### Round 3 — head `636eed04` (current)

Both reviewers re-reviewed the diff `9d0e304f..636eed04`: **GPT clean, Gemini clean.**

Regression tests were added for every fix (volume-qualified traversal keys, unsafe/empty/`C:` coordinates for both stores, unusable-PDB-name preserves snupkg but skips symbol servers, case-insensitive in-memory entry lookup). All affected suites pass and the full `dotnet-inspect.slnx -c Release` build is clean (3 pre-existing unrelated warnings). Threat-model doc updated.

richlander:
Ready to merge

Both fixed-head adversarial reviews (GPT-5.6, Gemini 3.1 Pro) are clean on `636eed04`, all findings are reconciled above, and CI is green (test, build-net10, markdownlint, changes all pass; csharp-diff-smoke/il-diff-smoke/pack path-gated). Review worktrees removed.

PR #3160, opened 2026-07-25
Add library-owned StyleOptionCatalog as the style-knob source of truth

richlander:
## What

Introduces `StyleOptionCatalog` (`ILInspector.Decompiler.Pipeline`) — a public, library-owned `IReadOnlyList<StyleOptionDescriptor>` that is the single source of truth for the opt-in boolean `PrinterOptions` knobs, and makes the CLI's `RenderStyleConfig` resolver consume it instead of a hand-maintained key→property switch.

Unblocks a Wasm-built UI (and any other host) discovering the options, knowing which are byte-divergent, and knowing which are mutually exclusive. Also lays the metadata foundation the future "full taste" aggregate (#3138) consumes (the `OracleEndorsed` subset).

## Design

Each `StyleOptionDescriptor` carries:
- `Id` (stable kebab-case), `Title`, `Summary`
- `Tier` (`Formatting` / `Spelling` / `Lens` / `Synthesis`) + `ByteDivergent` (true only for the two lenses)
- `OracleEndorsed` (membership test for "full taste")
- `ConfigKey` (editorconfig `dotnet_style_*` or tool-owned `dotnet_inspect_style_*`; `null` for API-only formatting/synthesis knobs)
- `ConflictGroup` (mutually-exclusive group; the ternary + branchless lenses share `guarded-boolean-return`)
- NativeAOT-safe `Get`/`With` delegates — no reflection, so the catalog is enumerable from a Wasm host without breaking the product's NativeAOT constraint.

`RenderStyleConfig` now derives its recognized keys from the catalog, so the CLI vocabulary and the option surface cannot drift. The refactor is **behavior-preserving**: `root`-key handling, `value:severity` tolerance, and per-key warnings are unchanged.

## Scope boundary

The single non-boolean knob (`ExpressionBodyArrowPlacement`) is intentionally **not** modeled yet — the catalog describes boolean toggles. A general non-bool shape can be added when a second enum knob justifies it.

## Evidence

- New `StyleOptionCatalogTests` (10): drift guard (catalog count == PrinterOptions bool-property count via test-only reflection), id/config-key uniqueness, get/with round-trip + isolation (each `With` toggles exactly one knob), `ByteDivergent` ⇔ `Lens` tier, conflict group == the two lenses, oracle set excludes branchless, key vocabulary.
- New resolver integration tests (2): every catalog config key round-trips through `RenderStyleConfig.Parse`; API-only knobs have no config key and still warn as unknown.
- `RenderStyleConfigTests` green (41/0) — refactor behavior-preserving.
- Fast decompiler suite 3664/0; full `dotnet-inspect.Tests` clean except one pre-existing environmental flake (`PlatformResolverTests.GetPacksDirectory_ReturnsValidPath`, a packs-path assertion that passes in isolation and is untouched by this change).
- `docs/decompiler-taste.md` updated (new "Option catalog" subsection); markdownlint clean.

Adversarial review to follow (new public API + parser refactor).

### DIFF
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -445,6 +445,9 @@ dotnet_style_prefer_conditional_expression_over_return = true
   ternary [style lens](#style-lenses-behavior-faithful-byte-divergent)), and
   `dotnet_inspect_style_prefer_branchless_boolean` (the non-oracle-endorsed
   branchless lens, under a tool-owned key). The set grows as more knobs ship.
+- The recognized keys are not hand-maintained in the resolver: they come from the
+  library-owned `StyleOptionCatalog` (see [Option catalog](#option-catalog)), so
+  the CLI vocabulary and the option surface cannot drift.
 - Unknown keys, malformed lines, and non-boolean values are reported as a
   `Warning:` on stderr and skipped — the rest of the file still applies. A bad
   config never fails the run silently.
@@ -481,6 +484,28 @@ of the assembly and the options it is handed, so the config surface never change
 what the library computes for a given `PrinterOptions`. The knobs affect the
 primary decompiled-source view; the Annotated Source and IR-stage views stay on
 the shipped default so they remain aligned with the IL.
+
+### Option catalog
+
+The recognized knobs are described once, in the library, by
+`StyleOptionCatalog` (`ILInspector.Decompiler.Pipeline`). Each
+`StyleOptionDescriptor` carries a knob's stable id, human-facing title and
+summary, its tier (`Formatting`, `Spelling`, `Lens`, or `Synthesis`), whether it
+is `ByteDivergent`, whether it is `OracleEndorsed`, its `.dotnet-inspectconfig`
+key (`null` for API-only formatting/synthesis knobs), a `ConflictGroup` for
+mutually-exclusive knobs, and NativeAOT-safe `Get`/`With` delegates that read and
+set the knob on a `PrinterOptions` without reflection.
+
+This makes the option surface discoverable and drift-proof for every host, not
+just the CLI: the config resolver derives its recognized keys from the catalog,
+a Wasm UI can enumerate the knobs (grouping the mutually-exclusive lenses by
+`ConflictGroup` and toggling each through `Get`/`With`), and the future "full
+taste" aggregate is exactly the `OracleEndorsed` subset. The two guarded-boolean
+lenses share the `guarded-boolean-return` conflict group, so a picker offers at
+most one (the printer still resolves any overlap deterministically, preferring the
+oracle-endorsed ternary). The single non-boolean knob
+(`ExpressionBodyArrowPlacement`) is not yet modeled in the catalog, which
+currently describes boolean toggles.
 
 ## Verification and soundness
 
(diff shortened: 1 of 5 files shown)


richlander:

Reviewed exact head `4b9bc240` (base `2114319f`) with two independent models, each isolated in a read-only worktree, each running the build + targeted suites (not just reading).

| Reviewer | Verdict | Evidence |
| --- | --- | --- |
| GPT-5.6 | **CLEAN** | build 0 warn/0 err; `StyleOptionCatalogTests` 10/0; `RenderStyleConfigTests` 41/0 |
| Gemini 3.1 Pro | **CLEAN** | same build + suites green; plus `/tmp` probes for mutability, parse, dictionary lookup |

Both independently verified all six attack points:

1. **Behavior-preservation** — old `switch` vs new catalog-driven resolver compared key-by-key. All 6 config keys are lowercase ASCII and the parser lowercases via `ToLowerInvariant()`, so `StringComparer.Ordinal` resolves them with no case regression. `root` handling, `value:severity` tolerance, repeated-key last-wins, and every `Warn(...)` shape are identical.
2. **Get/With correctness** — all 8 descriptors manually cross-checked against `PrinterOptions`; no copy-paste crossing; each `With` toggles exactly one distinct property (round-trip + isolation tests pass).
3. **Flag accuracy** — `ByteDivergent` ⇔ `Tier==Lens`; oracle set is the 4 qualifications + ternary (branchless excluded); `guarded-boolean-return` conflict group is exactly {ternary, branchless}; the two null config keys are exactly the API-only knobs.
4. **NativeAOT / reflection-free** — product paths use static-lambda delegates only; the `KnobsByKey` LINQ static initializer is AOT-safe, deterministic, and cannot throw (config keys are unique, so `ToDictionary` won't collide).
5. **Public API immutability** — `Options` compiles to a read-only collection-expression wrapper that rejects array-cast and mutation; descriptors are `required`/`init`-only.
6. **Doc consistency** — the new "Option catalog" subsection matches the metadata and honestly discloses the deferred `ExpressionBodyArrowPlacement` enum knob.

**Non-blocking follow-up (both noted):** the `With_TogglesExactlyOneKnob…` test proves each knob toggles a *distinct* piece of state but does not, on its own, assert the semantic `Id`↔property binding (e.g. `qualify-field-access` → `QualifyFieldAccess`). Both reviewers manually audited all 8 bindings and confirmed them 100% accurate, so this is a test-hardening opportunity, not a defect. I'll add an `Id`-to-property binding assertion in a later PR alongside the "full taste" aggregate work.

No blocking findings from either reviewer. Awaiting the `test` CI job before marking ready.

richlander:
Ready to merge

- CI green: `test` (8m), `build-net10`, `csharp-diff-smoke`, `markdownlint`, `changes` all pass; `pack`/`il-diff-smoke` path-skipped.
- Both adversarial reviews (GPT-5.6, Gemini 3.1 Pro) CLEAN at the merged head `4b9bc240` with real-run evidence; all six attack points verified; no blocking findings.
- The only noted item is a non-blocking test-hardening follow-up (Id↔property binding assertion), which I'll land with the upcoming "full taste" aggregate work.

PR #3137, opened 2026-07-24
Add EVIL authored-corpus progress card (--history-card)

richlander:
## What

Adds a `--history-card` harness mode that renders a Markout progress card over the committed EVIL run-history trend store (`tools/DecompilerHarness/corpus/evil-runs/history.jsonl`, from #3118): the last N runs as a trend table plus a latest-vs-previous movement table. Answers the "show progress on the last n measurements" ask on #3079.

Example (current committed history):

```text
# EVIL authored-corpus progress

Showing 3 of 3 recorded run(s). Track product defects (target-body decompiler bugs), not raw invalid (~92% harness shell-reconstruction noise per #3079).

## Runs

| Date | Commit | Valid % | Correct | Invalid (raw) | Product defects | Harness noise |
| ---- | ------ | ------- | ------- | ------------- | --------------- | ------------- |
| 2026-07-20 | (baseline) | 56.6% | 1501 | 5209 | — | — |
| 2026-07-23 | d6af0165 | 56.2% | 1537 | 5258 | — | — |
| 2026-07-24 | 16c0687f | 56.2% | 1539 | 5259 | 306 | 4826 |

## Movement (latest vs previous)

| Metric | Previous | Latest | Δ | Trend |
| ------ | -------- | ------ | - | ----- |
| Valid % | 56.2% | 56.2% | ±0.0% | unchanged |
| Correct | 1537 | 1539 | +2 | improved |
| Invalid (raw) | 5258 | 5259 | +1 | regressed |
| Product defects | — | 306 | — | n/a |
```

## Design

- **Headline metric is product defects, not raw invalid.** Per #3079/#3096 raw `invalid` is ~92% harness shell-reconstruction noise that does not move on decompiler fixes; the card surfaces `invalidBreakdown.productBodyDefect` as the signal that tracks progress.
- **Failure stays visible.** Runs predating #3096 have no `invalidBreakdown`; their product/harness columns and the product-defect movement row render as `—`/`n/a`, never a fabricated zero.
- **Per-metric polarity.** Movement marks lower-is-better metrics (invalid, product defects) `improved` when they drop and higher-is-better metrics (valid %, correct) `improved` when they rise.
- Reuses the existing Markout serializer/context pattern (`ReturnToSenderCatalogReport` shape). Reads no assemblies, runs no decompiler.

## Flags

- `--history-card` — render the card.
- `--history-window <n>` — last n runs (default 5; `<= 0` shows all).
- `--history-path <file>` — override the committed default.

## Evidence

[...]

PR #3167, opened 2026-07-25
Fold CorpusSensor count delta into the Markout Change cell (#3143)

### LINKED ISSUE #3143: CorpusSensor: fold ad-hoc "Count delta" column into Markout Change cell (Delta.Absolute)
## Summary

`CorpusSensor`'s quality-comparison table carries a separate **"Count delta"**
column that is built by an ad-hoc helper, duplicating a value Markout 0.24.0
can now render natively inside the existing "Change" cell.

## Current ad-hoc rendering

`tools/DecompilerHarness/CorpusSensor.cs`:

- `CountChangeRow` (~L2502) and `ShareChangeRow` (~L2514) each emit two sources:
  a Markout `Change<T>` cell **plus** a hand-built second column:

  ```csharp
  new Source("Count delta",
      new QualityText(countDeltaKnown ? Delta(current - baseline) : "n/a"))
  ```

- `Delta(int)` (~L2625) formats the signed magnitude (`+N` / `N`).
- `QualityText` (~L2647) is a tiny `IMarkoutCell` wrapper used only to shove
  that string into a cell.

## Proposed change

Set `Delta = Delta.Absolute` on the existing `MarkoutCellFormat` for the
"Change" source so the cell renders `before → after (+N) ✓` natively, then:

- Remove the "Count delta" `Source` from both row builders.
- Remove the `Delta(int)` helper (if unused elsewhere — it is also referenced
  by a couple of prose gate summaries; keep it only for those or inline them).
- Remove the `QualityText` type if it becomes unused.

Verified via `dnx dotnet-inspect -- member Change<T> FormatInline
--package Markout@0.24.0 -S "Decompiled Source"`: the scalar path emits
`DeltaSuffix(format.Delta)` when `format.Delta != None`, and `QualityRate`
(rate rows) already implements `IDeltaCountable`/`IGoalMagnitude`, so both int
and rate rows fold cleanly.

## Test / risk notes

- Product-output change: the table loses a column; the delta moves inside the
  Change cell. Update `CorpusSensorComparisonTests` accordingly (same test file
  touched by the 0.24.0 glyph alignment on #3137).
- Confirm the count-delta magnitude is still discoverable for rate rows (where
  the Change cell shows `count (pct)`), since the raw count diff is the extra
  signal that column carried.
- Small, self-contained; likely low-risk but still a rendering change — pair
  with the CorpusSensor tests.

## Context

Spun out of the Markout 0.24.0 adoption on #3137.

richlander:
Closes #3143.

## What

Removes the redundant **Count delta** column from the CorpusSensor
metric-change table (rendered inside the `--quality-diff-card`) by folding the
signed count delta into each Markout `Change` cell:

- **Scalar count rows** (`Change<int>` — Full malformed, RTS parity worse, Pass
  bugs) set `Delta = Delta.Absolute`, rendering a bare `(+N)`:
  `165 → 168 (+3) ✗`.
- **Composite share rows** (`Change<QualityRate>` — all the share metrics) add a
  new `IDeltaCountable` implementation on `QualityRate` (`DeltaCount => Count`)
  and set `DeltaNoun = "methods"`, rendering `1 (50.00%) → 2 (100.00%) (+1 methods) ✓`.

The now-unused `QualityText` cell type is deleted. The `Delta(int)` helper is
kept because it is still used by the prose gate/summary lines.

## Correctness of incomparable rows

`countDeltaKnown: false` is always paired with `Goal.Context` (incomparable
population/sample). Those rows already disclose incomparability two ways — a
label suffix (`(population differs)` / `(sampling differs)`) and a suppressed
verdict glyph — so folding drops the redundant explicit `n/a` while keeping the
signal visible. Folded incomparable rows emit **no** inline delta (scalar:
`Delta.None`; composite: `DeltaNoun = null`).

## Note on the issue premise

The issue stated `QualityRate` already implemented `IDeltaCountable` — it did
not. It also assumed `Delta.Absolute` would fold on the composite cell, but
Markout's composite branch ignores `format.Delta` and derives its numeric delta
from `IDeltaCountable` + `DeltaNoun`. That is why share rows carry a `methods`
noun (a bare `(+N)` is not achievable for composite cells) while count rows
stay bare. This asymmetry was confirmed with the maintainer.

## Evidence

- `CorpusSensorComparisonTests` — 59/59 (updated to the folded strings, covering
  comparable `(+N)` / `(+N methods)`, zero-delta `(0)` / `(0 methods)`, and
  incomparable no-delta cases for both count and rate rows).
- Full fast Decompiler suite — 3654/3654.
- Changed Markdown passes `markdownlint`.
- Docs: updated the stale `Count delta` references in
  `docs/decompiler-quality.md` and `docs/templates/decompiler-pr.md`. (The
  illustrative card example in those docs predates #3137's `Change`-cell
  adoption and is broadly simplified; this PR only removes the now-defunct
  `Count delta` column, not a full example refresh.)

The CorpusSensor table is Markdown-only (no tsv/jsonl serialization), so
Markdown is the only affected output mode.

### DIFF
--- a/docs/decompiler-quality.md
+++ b/docs/decompiler-quality.md
@@ -524,8 +524,9 @@ instead so the PR records the targeted checked population and RTS/current A/B
 evidence.
 
 Rate deltas in card prose use **percentage points** (`pp`): `+0.49 pp` means
-the rate increased from, for example, `82.17%` to `82.66%`. Counts still use raw
-signed count deltas in `Count delta`.
+the rate increased from, for example, `82.17%` to `82.66%`. The metric-change
+table folds the signed count delta into each `Change` cell — `(+3)` for count
+metrics and `(+3 methods)` for share metrics.
 
 For behaviour-preserving refactors, the existing #1174/#1166 real-world corpus
 sensor values are enough when you need the broader Deep Inspect signal:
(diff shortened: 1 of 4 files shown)


richlander:

Two-model review at fixed head `d7e83b61` in isolated worktrees. **Both CLEAN.**

- **GPT-5.6 Sol** — CLEAN, no blocking findings. Confirmed: (1) composite delta wired correctly — `ShareChangeRow` passes `Change<QualityRate>(baseline, current)` and Markout computes `After.DeltaCount - Before.DeltaCount`, so `DeltaCount => Count` yields `current.Count - baseline.Count`; (2) every `countDeltaKnown: false` call site is coupled with `Goal.Context` plus a population/sampling/oracle/contract suffix, so Markout suppresses both the inline delta and the verdict glyph. Build exit 0, 59/59 tests, markdownlint exit 0.
- **Gemini 3.1 Pro** — CLEAN, no blocking findings. Independently verified the same delta wiring and incomparable coupling, confirmed `QualityText` removed with no dangling references, `Delta(int)` still used by prose gate lines (e.g. `CorpusSensor.cs:2129`), Markdown-only table (no JSON/schema surface), build + 59 tests + markdownlint all clean.

### Shared non-blocking NIT (accepted by-design)

Folded count deltas lose thousands separators (e.g. `+1500` instead of the old dedicated column's `+1,500`). This is inherent to Markout's native `Change<T>` delta rendering; there is no way to reintroduce `N0` formatting without abandoning the fold that this PR exists to perform. The base counts in each cell still use `N0` (`77,376 (88.02%)`), and behavior-preserving PRs have small deltas, so the impact is cosmetic. Both reviewers judged it acceptable. Recorded as a non-blocking follow-up candidate rather than a change here.

No code changes were needed after review, so the reviewed head is the current head.

richlander:
Ready to merge

PR #3168, opened 2026-07-25
Render class finalizers as ~Type() destructor syntax (#3157)

### LINKED ISSUE #3157: Decompiler: finalizer renders as `void Finalize()` instead of `~Type()`
## Summary

A C# finalizer (`~Type()`) decompiles to a plain method named `void Finalize()`. This is incorrect C#: `void Finalize()` does not wire up as a finalizer (Roslyn warns CS0249 "Do not override object.Finalize; provide a destructor instead"), and recompiling the output would not reproduce the finalizer. The metadata method is `Finalize` overriding `object.Finalize`, which C# spells only as `~Type()`.

Discovered while auditing #3084 (expression-bodied transition). The finalizer's **body** folds correctly; only the **head** is wrong, so this is independent of expression bodies (it misrenders in block form too).

## Repro

Source:

```csharp
public class Handle
{
    private int _n;
    ~Handle() { _n = 0; }
}
```

Decompiled (via `MemberBodyProducer`), single-statement body folds:

```csharp
public class Handle
{
    private int _n;

    void Finalize() => _n = 0;   // WRONG: should be ~Handle()
}
```

Forcing a block body (two statements) shows the same wrong head, confirming it is independent of the expression-body fold:

```csharp
void Finalize()
{
    _n = 0;
    _n = 1;
}
```

## Expected

```csharp
~Handle() => _n = 0;
// or, block form:
~Handle()
{
    _n = 0;
    _n = 1;
}
```

## Notes

- The metadata method is the virtual `Finalize` slot overriding `System.Object.Finalize`; detect that (name `Finalize`, no parameters, overrides `object.Finalize`) at the declaration/head formatter and emit `~<DeclaringType>()` with no return type or modifiers.
- `~Type()` is legal as an expression-bodied member, so the existing fold composes once the head is fixed (as the repro's `=> _n = 0;` already shows).
- Correctness/round-trip bug, not a taste question.

richlander:
Fixes #3157.

A C# finalizer (`~Type()`) compiles to an `object.Finalize` override that Roslyn
emits with an explicit `.override` MethodImpl. The extractor's explicit-body
collection therefore classified it as an explicit-interface-implementation, and
the C# declaration writer spelled the bare slot `void Finalize()` instead of the
`~Type()` destructor syntax. The body raise (stripping the `try/finally { base.Finalize(); }`)
was already handled by `DestructorRecoveryPass`; only the member **head** was wrong.

This adds a metadata fact `ApiMember.IsFinalizer` (a non-static, parameterless,
void, reused-slot virtual method named `Finalize` on a class), set in the
extractor. The C# writer then spells `~Type()` and drops the illegal
accessibility/override modifiers, keeping only `unsafe`. Metadata owns the slot
fact; CSharp owns the spelling.

The exact name `Finalize` (not dot-qualified) plus the reused (non-new) slot
excludes an explicit `IFoo.Finalize()` implementation and implicit interface
impls (which new-slot) — covered by a negative test.

## Original source (IL anchor)

SourceLink C# is unavailable for the local fixture, so the authoritative anchor
is the `IL` section (`member FinDemo.Handle Finalize -S "IL"`) — the finalizer's
`try { _n = 0; } finally { base.Finalize(); }` slot:

```text
IL_0000: ldarg.0
IL_0001: ldc.i4.0
IL_0002: stfld        FinDemo.Handle::_n
IL_0007: leave.s      IL_0010
IL_0009: ldarg.0
IL_000A: call         System.Object::Finalize()
IL_000F: endfinally
IL_0010: ret
```

## Before (base 76d73685, `-S "Decompiled Source"`)

```csharp
public class Handle
{
    private int _n;

    public Handle(int n) => _n = n;

    void Finalize() => _n = 0;

    public int Value => _n;
}
```

- Valid: False in intent — a bare `void Finalize()` compiles only with warning
  CS0465 and is **not** a finalizer.
- Correct: False — different observable behavior (no finalization slot).
- IL fidelity: False — recompiles to a plain method, not the `Finalize` override.
- Taste applied: none.
- Commit: 76d73685.

## After (head, `-S "Decompiled Source"`)

```csharp
public class Handle
{
    private int _n;

    public Handle(int n) => _n = n;

    ~Handle() => _n = 0;

    public int Value => _n;
}
```

- Valid: True.
- Correct: True — recompiles to the `object.Finalize` override slot.
- IL fidelity: True.
- Taste applied: none (finalizer spelling is a correctness fix, not a style lens).
- Commit: head of this branch.

## Fully raised

Same as After — this is the intended endpoint.

## Known limitation (pre-existing, out of scope)

The finalizer keeps `Kind = explicit-interface-implementation`, so in metadata
views it still lists under the "Explicit Interface Implementations" section and
the raw-metadata `--shape` tree still shows the slot as `void Finalize()`
(consistent with how that view shows `.ctor` for constructors). Re-categorizing
would change the member `Kind`/identity/digest with a wider blast radius; happy
to follow up separately. This PR fixes the decompiled C# render, which is what
#3157 reports.

## Validation

- Real CLI: `type FinDemo.Handle -S "Decompiled Source" --print` renders
  `~Handle() => _n = 0;` end to end.
- New tests: 3 unit (`CSharpDeclarationWriterTests`: finalizer syntax, unsafe-only
  modifier, generic type-arity drop) + 3 end-to-end
  (`MemberBodyProducerExpressionBodyTests`: expression-bodied finalizer, block
  finalizer, and negative `void IThing.Finalize()` explicit impl).
- Suites: CSharp.Tests 329, Metadata.Tests 794, MemberBodyProducerExpressionBodyTests 10,
  DestructorRecoveryPassTests 3 — all pass. Product `dotnet-inspect.Tests`: the
  only failures are the known pre-existing env reds (PerformanceTriage/ResourceTriage/
  caller-loop), identical base vs head.

### DIFF
--- a/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
@@ -43,6 +43,64 @@ public sealed class CSharpDeclarationWriterTests
     }
 
     [Fact]
+    public void FinalizerMember_RendersDestructorSyntaxWithoutModifiers()
+    {
+        var type = new ApiType { Namespace = "Samples", Name = "Handle", Kind = "class" };
+        var finalizer = new ApiMember
+        {
+            Name = "Finalize",
+            // Roslyn emits the finalizer with an explicit .override MethodImpl, so
+            // it surfaces as an explicit-interface-implementation kind; IsFinalizer
+            // is the metadata fact that drives the ~Type() spelling.
+            Kind = "explicit-interface-implementation",
+            Signature = "void Finalize()",
+            IsVirtual = true,
+            IsFinalizer = true,
+        };
+
+        var declaration = CSharpDeclarationWriter.RenderMemberDeclaration(type, finalizer);
+
+        Assert.Equal("~Handle()", declaration);
+    }
+
+    [Fact]
+    public void UnsafeFinalizerMember_KeepsUnsafeModifierOnly()
+    {
+        var type = new ApiType { Namespace = "Samples", Name = "Handle", Kind = "class" };
+        var finalizer = new ApiMember
+        {
+            Name = "Finalize",
+            Kind = "explicit-interface-implementation",
+            Signature = "void Finalize()",
+            IsVirtual = true,
+            IsFinalizer = true,
+            IsUnsafe = true,
+        };
+
+        var declaration = CSharpDeclarationWriter.RenderMemberDeclaration(type, finalizer);
+
+        Assert.Equal("unsafe ~Handle()", declaration);
+    }
+
+    [Fact]
+    public void FinalizerMember_OnGenericType_DropsTypeArity()
+    {
+        var type = new ApiType { Namespace = "Samples", Name = "Box`1", Kind = "class" };
+        var finalizer = new ApiMember
+        {
+            Name = "Finalize",
+            Kind = "explicit-interface-implementation",
+            Signature = "void Finalize()",
+            IsVirtual = true,
+            IsFinalizer = true,
+        };
+
+        var declaration = CSharpDeclarationWriter.RenderMemberDeclaration(type, finalizer);
+
+        Assert.Equal("~Box()", declaration);
+    }
+
+    [Fact]
     public void ShortWithUsingsMemberUnit_GeneratesImportsAndShortensTypes()
     {
         var type = CreateSampleType();
--- a/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
+++ b/src/ILInspector.CSharp.Tests/CSharpDeclarationWriterTests.cs
@@ -101,6 +101,31 @@ public sealed class CSharpDeclarationWriterTests
     }
 
     [Fact]
+    public void FinalizerMember_WithSuppressFinalizerSpelling_KeepsLiteralFinalize()
+    {
+        // Issue #3157 (fidelity hardening): the '~Type()' spelling assumes the
+        // recompiled destructor re-emits the mandatory 'base.Finalize()'. When the
+        // decompiled body did NOT recover the canonical destructor scaffold, the
+        // full-body path suppresses the destructor spelling so recompiling keeps
+        // the observed body instead of silently re-injecting the base call.
+        var type = new ApiType { Namespace = "Samples", Name = "Handle", Kind = "class" };
+        var finalizer = new ApiMember
+        {
+            Name = "Finalize",
+            Kind = "explicit-interface-implementation",
+            Signature = "void Finalize()",
+            IsVirtual = true,
+            IsFinalizer = true,
+        };
+
+        var declaration = CSharpDeclarationWriter.RenderMemberDeclaration(
+            type, finalizer, new CSharpDeclarationOptions { SuppressFinalizerSpelling = true });
+
+        Assert.Equal("void Finalize()", declaration);
+        Assert.DoesNotContain("~Handle", declaration);
+    }
+
+    [Fact]
     public void ShortWithUsingsMemberUnit_GeneratesImportsAndShortensTypes()
     {
         var type = CreateSampleType();
--- a/src/ILInspector.CSharp/CSharpDeclarationWriter.cs
+++ b/src/ILInspector.CSharp/CSharpDeclarationWriter.cs
@@ -324,6 +324,10 @@ internal static class CSharpDeclarationWriter
         {
             signature = $"{FormatConstructorTypeName(type.Name)}()";
         }
+        else if (member.IsFinalizer)
+        {
+            signature = $"~{FormatConstructorTypeName(type.Name)}()";
+        }
         else if (member.Kind == "constructor")
         {
             var typeName = FormatConstructorTypeName(type.Name);
@@ -374,6 +378,13 @@ internal static class CSharpDeclarationWriter
         if (member.Name == ".cctor")
         {
             modifiers.Add("static");
+            if (member.IsUnsafe || options.ForceUnsafe)
+                modifiers.Add("unsafe");
+        }
+        else if (member.IsFinalizer)
+        {
+            // A finalizer (`~Type()`) carries no accessibility or override
+            // modifiers — only `unsafe` is legal on it.
             if (member.IsUnsafe || options.ForceUnsafe)
                 modifiers.Add("unsafe");
         }
--- a/src/ILInspector.CSharp/CSharpDeclarationWriter.cs
+++ b/src/ILInspector.CSharp/CSharpDeclarationWriter.cs
@@ -30,6 +30,15 @@ internal sealed record CSharpDeclarationOptions
     public bool IncludeObsoleteAttribute { get; init; } = true;
     public bool OmitInterfaceMemberModifiers { get; init; }
     public bool OmitPropertyAccessors { get; init; }
+
+    /// <summary>
+    /// When true, a finalizer member (<see cref="ApiMember.IsFinalizer"/>) is
+    /// rendered as the literal <c>void Finalize()</c> method rather than the
+    /// <c>~Type()</c> destructor syntax. Set on body-bearing renders whose body
+    /// was not recovered as a canonical destructor, so the emitted source does
+    /// not silently re-inject the compiler's mandatory <c>base.Finalize()</c>.
+    /// </summary>
+    public bool SuppressFinalizerSpelling { get; init; }
 }
 
 internal sealed record CSharpRenderedDeclaration(
@@ -324,7 +333,7 @@ internal static class CSharpDeclarationWriter
         {
             signature = $"{FormatConstructorTypeName(type.Name)}()";
         }
-        else if (member.IsFinalizer)
+        else if (member.IsFinalizer && !options.SuppressFinalizerSpelling)
         {
             signature = $"~{FormatConstructorTypeName(type.Name)}()";
         }
@@ -381,7 +390,7 @@ internal static class CSharpDeclarationWriter
             if (member.IsUnsafe || options.ForceUnsafe)
                 modifiers.Add("unsafe");
         }
-        else if (member.IsFinalizer)
+        else if (member.IsFinalizer && !options.SuppressFinalizerSpelling)
         {
             // A finalizer (`~Type()`) carries no accessibility or override
             // modifiers — only `unsafe` is legal on it.
--- a/src/ILInspector.CSharp/CSharpFormatter.cs
+++ b/src/ILInspector.CSharp/CSharpFormatter.cs
@@ -85,7 +85,8 @@ public sealed class CSharpFormatter
             _declarationOptions with
             {
                 ForceAsync = _declarationOptions.ForceAsync || body.RequiresAsyncModifier,
-                ForceUnsafe = _declarationOptions.ForceUnsafe || body.RequiresUnsafeModifier
+                ForceUnsafe = _declarationOptions.ForceUnsafe || body.RequiresUnsafeModifier,
+                SuppressFinalizerSpelling = body.SuppressDestructorSyntax
             },
             methodParameters);
     }
--- a/src/ILInspector.CSharp/CSharpTypePrintContracts.cs
+++ b/src/ILInspector.CSharp/CSharpTypePrintContracts.cs
@@ -23,6 +23,17 @@ public abstract record CSharpMemberBody
     /// unsafe signature already represented by <see cref="ApiMember.IsUnsafe"/>.
     /// </summary>
     public bool RequiresUnsafeModifier { get; init; }
+
+    /// <summary>
+    /// True when a finalizer member (<see cref="ApiMember.IsFinalizer"/>) must
+    /// be rendered as the literal <c>void Finalize()</c> method rather than the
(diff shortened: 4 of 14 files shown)


richlander:

Two-model adversarial review (Gemini 3 Pro + GPT-5.6), six rounds, each on a fixed pushed head. **Both models are CLEAN on the final head `4bf117fd`.** CI on `4bf117fd` is green (build, product test suite, csharp-diff-smoke).

### Findings and resolutions, by round

**Round 1 (`5f43c026`)**
- *GPT* — Unrelated virtual slots misclassified as finalizers (`ReuseSlot`/name-based detection); generic `Finalize<T>()` mis-spelled.
- *GPT* — Non-canonical `Finalize` body gained an implicit `base.Finalize()` on recompile (fidelity).
- *GPT* — `CSharpTypePrinter.SnapshotMember` dropped the `IsFinalizer` fact.
→ Fixed in `b282cd9f`: detection rewritten to key only on an explicit `.override` MethodImpl targeting `System.Object::Finalize`; `~Type()` gated on `DestructorRecoveryPass` recovering the canonical scaffold (`DecompilerResult.BodyIsDestructor`); `IsFinalizer` propagated.

**Round 2 (`b282cd9f`)**
- *GPT* — Selected-member render (`ApiOutputFormatter.FormatSourceWithDeclaration`) bypassed the fidelity gate.
- *GPT* — A user-defined `System.Object` (non-root fake) was accepted.
- *Gemini* — Generic explicit `Finalize<T>()` override still slipped through.
→ Fixed in `e353967e`: selected-member + shared `ProduceBody` substrate gated; `IsSystemObjectType` TypeDefinition branch requires `BaseType.IsNil`; extractor + pass stand down on non-zero generic arity.

**Round 3 (`e353967e`)**
- *Both* independently — Cross-assembly fake `System.Object` via the `TypeReference` branch (the `BaseType.IsNil` guard only covered TypeDefinition).
→ Fixed in `08584000`: `IsSystemObjectType` TypeReference branch requires `ResolvesThroughCoreLibrary` (ResolutionScope must be an AssemblyReference to a recognized core library).

**Round 4 (`08584000`)**
- *Gemini* — **CLEAN.** Judged assembly-name impersonation an out-of-scope supply-chain concern.
- *GPT* — **NOT CLEAN.** Name-only core-library check is defeated by an unsigned assembly *named* `System.Runtime` (null PKT).
→ Split verdict escalated to the maintainer, who chose to add public-key-token validation. Fixed in `827c63bd`: each core-library name pinned to its strong-name public-key token; `TokenMatches` handles both the token and full-public-key (SHA-1 low 8 bytes, reversed) encodings; nil rejected.

**Round 5 (`827c63bd`)**
- *GPT* — **CLEAN.** Impersonation rejected; both token encodings sound; real finalizers unaffected.
- *Gemini* — **NOT CLEAN.** The 1:1 name→token map was a false-negative trap: `mscorlib` legitimately shipped under several Microsoft tokens (Silverlight/PCL/WP `7cec85d7bea7798e`, .NET Compact Framework `969db8053d3322ac`), so valid historical finalizers regressed to `void Finalize()`.
→ Fixed in `4bf117fd`: map is now `Dictionary<string, byte[][]>`; `mscorlib` accepts all three legitimate Microsoft tokens; `TryComputeToken` also fails closed on a wrong-length token blob. All three remain genuine Microsoft strong-name tokens, so the impersonation vector stays closed.

**Round 6 (`4bf117fd`) — final**
- *Gemini* — **CLEAN.** Historical mscorlib profiles accepted (verified via mock assemblies with each token); null/wrong-length/arbitrary tokens rejected; token computation matches the strong-name spec.
- *GPT* — **CLEAN.** Impersonation still rejected; three legit tokens accepted; nil/wrong-length fail closed; non-mscorlib and real Roslyn finalizers unchanged.

### Verification
- Metadata suite 794/0; `DestructorRecoveryPassTests` 4/0; `CSharpDeclarationWriterTests` 43/0.
- Live: whole-type and selected-member render `~Handle() => _n = 0;` for the real fixture; a non-canonical `Finalize` override renders literal `void Finalize()` and round-trips without a re-injected base call.
- No `ilasm` locally, so several adversarial repros were reasoned/mock-`MetadataBuilder` based rather than assembled; noted where applicable.

### Known limitation (out of scope, follow-up)
An explicit-interface `Finalize` keeps `Kind = explicit-interface-implementation` (it lists under **Explicit Interface Implementations**; raw `--shape` still shows `void Finalize()`). Pre-existing classification behavior, not introduced here.

richlander:
Ready to merge

richlander:
## Independent two-model re-review at head `04ffa562`

I integrated latest `origin/main` into this branch (clean merge, `04ffa562`) and re-ran an independent adversarial review with **GPT-5.6** and **Gemini 3.1 Pro** against that exact head. Both reviewers worked in isolated detached worktrees with real compiled/SRM fixtures. **Verdict from both: changes required** — not ready to merge yet.

### Resolved since the prior (`5f43c0267`) review
- **False-positive finalizer detection** (custom `new virtual Finalize`, generic `Finalize<T>`, unrelated base slots, explicit interface impls, cross-assembly fake `System.Object`): **RESOLVED** by both reviewers' assessment. The `.override` MethodImpl-target approach + core-library public-key-token validation (`GetObjectFinalizeOverrides` / `ReferencesObjectFinalize` / `ResolvesThroughCoreLibrary`) correctly rejects these. Confirmed `IsFinalizer=false` on the adversarial fixtures.
- **Digest/identity blast radius**: none — GPT confirmed digest `b194804d1c` is identical with `IsFinalizer` toggled. Targeted suites pass (CSharpDeclarationWriterTests 43, MemberBodyProducerExpressionBodyTests 10, DestructorRecoveryPassTests 4).

### Open findings (both reviewers agree)

**1. `async ~Type()` — invalid C# (Medium; GPT + Gemini).** The modifier block in `CSharpDeclarationWriter.cs:437` appends `async` whenever `options.ForceAsync || member.IsAsync` and `Kind is "method" or "extension-method" or "explicit-interface-implementation"`, with no `IsFinalizer` exclusion. A finalizer's Kind matches, so it can render `async ~Handle()`, which Roslyn rejects (`CS1519`). Reachable via `ForceAsync` (body-driven `RequiresAsyncModifier`) or `member.IsAsync`. Both reviewers reproduced it.
_Suggested fix (both):_ add `&& !member.IsFinalizer` to the async condition.

**2. `is_finalizer` JSON schema expansion (GPT Medium / Gemini Low).** `ApiSurface.cs:467` `public bool IsFinalizer` has no `[JsonIgnore]`, so value-type serialization emits `"is_finalizer": false` on **every** member — contradicting this PR's stated "no schema change" intent (`docs/design/command-model.md:108`). Gemini notes it does match sibling bools (`is_sealed`, `is_override`) which are also unconditionally serialized, so it's defensible if matching those is preferred.
_Suggested fix (both):_ `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]` to honor the PR's explicit no-schema-change claim.

**3. VB (implicit) `object.Finalize` overrides missed — false negative (GPT High / Gemini "known limitation, verify intent").** VB emits a valid `Object.Finalize` override with **no** MethodImpl row, so the new detection misses it and it renders as `protected override void Finalize()`, which C# forbids (`CS0249`). Both reviewers reproduced this with a compiled VB fixture. The code comment documents this as intentional ("VB-style implicit `object.Finalize` overrides carry no MethodImpl and fall back to the literal `void Finalize()`").
- **Divergence:** GPT treats it as a **regression** and wants an implicit-slot-resolution fallback (at least for direct `System.Object` bases). Gemini treats it as an acceptable **documented limitation** under the SRM-only constraint, but warns VB assemblies will no longer round-trip through C# compilation.
- **Note for the author to confirm:** whether this uncompilable VB output is newly introduced by this PR or pre-existing on `main` (i.e. did `main` already strip the `try/finally` and emit a bare `override void Finalize()`?). If pre-existing, this is an unfixed edge case rather than a regression; if new, it should be gated.

### GPT-only finding

**4. Same-assembly fake `System.Object` bypasses core-lib validation (Medium; GPT).** The `TypeDefinition` branch of `IsSystemObjectType` (`ApiSurfaceExtractor.cs:~883`) accepts any local root `System.Object` by **nil-base + name only**, without the assembly-name/public-key-token check the cross-assembly `TypeReference` path enforces. A hand-crafted unsigned assembly with a root `System.Object` + a MethodImpl targeting its own `Finalize` is classified as a finalizer. I confirmed this by code inspection. Impact is cosmetic (mislabel → `~Type()`), only in adversarial hand-authored assemblies, but it is an inconsistency with the cross-assembly hardening. Gemini did not flag it.

### Bottom line
- **#1 (async)** and **#2 (JSON)** are cheap, both-reviewer-agreed, and directly serve the PR's "no other behavior change" claim — recommend fixing both.
- **#3 (VB)** needs an author decision: fix with an implicit-override fallback, or keep as a documented limitation (please confirm regression-vs-pre-existing).
- **#4** is a low-impact consistency gap; optional hardening.

I made **no code changes** here — this is a review/reconciliation pass. Findings are handed to the branch author to action. Not marking ready to merge; both fixed-head reviews must be clean first.

richlander:
## Adversarial re-review (whole-stack, frozen tip `2f8f24a1`) — 1 finding attributed to this PR

As part of a two-model whole-stack review (GPT-5.6 + Gemini 3.1 Pro) of the cumulative `#3168 → #3186 → #3195` diff against `main`, GPT-5.6 flagged one issue owned by this PR's writer logic. I reproduced it. (Gemini cleared the stack, but its negative cases for the source-scan finding used `~0`/`~(x)`, not `~identifier`, so it missed the real inputs — see #3186 for those two.)

### Medium — suppressed finalizer renders `public virtual void Finalize()` (new slot, CS0465)

`src/ILInspector.CSharp/CSharpDeclarationWriter.cs` (finalizer branch `else if (member.IsFinalizer && !options.SuppressFinalizerSpelling)`).

When destructor recovery does **not** recover the canonical scaffold, `SuppressFinalizerSpelling` is set to fall back to a literal `Finalize` spelling. The existing suppression test uses `Kind = "explicit-interface-implementation"`, whose branch omits accessibility, so it renders `void Finalize()`. **But once #3186's `Kind = "finalizer"` composes on top**, a suppressed finalizer no longer matches the finalizer branch *or* the explicit-interface branch — it falls into the **normal modifier path** and renders:

```csharp
public virtual void Finalize()
```

That introduces a **new virtual slot** instead of overriding `object.Finalize` (compiling it yields **CS0465** "Introducing a 'Finalize' method can interfere with destructor invocation"), losing GC finalization semantics on recompile — the exact fidelity failure suppression exists to prevent.

**Reachability:** any finalizer whose body fails destructor-scaffold recovery, once `Kind = "finalizer"` is in effect (i.e. with #3186 stacked). Code-confirmed via the branch structure; GPT confirmed CS0465 by compiling the emitted declaration.

**Suggested fix:** the suppressed-finalizer fallback must not emit accessibility/`virtual` that creates a new slot. Extend the finalizer branch to cover the suppressed case with a modifier-free (or `override`) spelling, and add a suppression test that uses the real `Kind = "finalizer"` (the current test's `explicit-interface-implementation` kind no longer exercises the stacked path). The fix most naturally lands in #3186 (where `Kind="finalizer"` is introduced); cross-referenced there.

My follow-up **#3195** (JSON/async cleanup) is clean under both models but sits atop this stack.

richlander:
## Whole-stack adversarial review — clean

Ran a cumulative two-model adversarial review of the finalizer stack at frozen heads:

- **#3168** `fix/issue-3157-finalizer` @ `32078265`
- **#3186** `fix/finalizer-section` @ `0f5491b7`
- **#3195** `fix/finalizer-json-async-cleanup` @ `496847b6` (tip; contains the other two)

Reviewers (both non-Claude, isolated detached worktrees at the tip, scratch under `/tmp`, no git mutations):

- **GPT-5.6** — *No changes required.*
- **Gemini 3.1 Pro** — *No significant issues found.*

Both **independently reproduced** the three prior findings and confirmed the fixes hold at the tip:

1. Suppressed finalizer with `Kind="finalizer"` renders modifier-free `void Finalize()` (no `public`/`virtual`/`async`; CS0465 slot avoided).
2. `Finalizer` section is discoverable/selectable in the member-overload pipeline (`member -m Finalize --all` and `-S Finalizer` both render `## Finalizer`).
3. `SourceLinkResolver` destructor extraction rejects the `~identifier;` (`~mask;`) `#line hidden` continuation and still extracts anchored `~Type()` destructors. The residual bare-`~Type` (no semicolon, exact type-name match) intersection remains the already-documented accepted out-of-scope limitation.

Local validation at the merged tip: `CSharpDeclarationWriterTests` 47/47, `SectionPipeline`+`ApiOutputFormatter` 156 (1 skip), all green.

Note: repo CI is `pull_request: [main]`-gated, so stacked PRs (base ≠ main) show no checks until retargeted. CI will run as the stack merges bottom-up (#3168 → #3186 → #3195).

richlander:
Ready to merge.

Bottom of the finalizer stack: base `main`, `MERGEABLE`/`CLEAN`, CI green (build-net10, csharp-diff-smoke, test, changes all pass; il-diff-smoke/markdownlint/pack are path-gated skips). Covered by the clean whole-stack two-model adversarial review at frozen head `32078265` (see reconciliation above).

Merge this first; #3186 then #3195 retarget to `main` and pick up CI as the stack merges bottom-up.

PR #3171, opened 2026-07-25
Replace ExpressionBodyArrowPlacement enum with a boolean catalog knob

richlander:
## What

The expression-body arrow placement was the **one** opt-in `PrinterOptions` knob that wasn't a boolean, so it was deliberately excluded from `StyleOptionCatalog` (#3160) and couldn't be discovered/toggled through the uniform catalog surface that hosts (the CLI resolver, a Wasm UI, the future "full taste" aggregate) depend on.

It has exactly two states — arrow on the same line (shipped default) or wrapped onto the next line — so this models it as a boolean and adds it to the catalog.

## Change

- Remove the `ExpressionBodyArrowPlacement` enum and the enum-typed `PrinterOptions` / `DecompilerOptions` properties; replace with a single `bool WrapExpressionBodyArrow`.
- **Convention-preserving polarity:** `false` = shipped default (same line), `true` = opt-in (wrap onto next line) — matching every other knob's "true engages the opt-in behavior". This flips the *spelling* but not the behavior: `PrinterOptions.Default` still renders the arrow on the same line, byte-for-byte.
- Add the knob to `StyleOptionCatalog` as an API-only **Formatting** toggle: no config key, not oracle-endorsed, not byte-divergent (whitespace-only; IL unchanged).
- Docs: `docs/decompiler-taste.md` now states the catalog is exhaustive (every opt-in knob is boolean); a future non-boolean knob would need a descriptor shape carrying its value domain.

## Why it's a clean swap

Every real consumer already funneled the enum through a bool — `MemberBodyProducer.WrapExpressionBodyArrow(...)` and the CSharp layer's `wrapArrow` parameter — so the enum only existed at the API boundary. `DecompilerOptions.EffectiveOptions` is an inspectable record, not serialized to any CLI output, so there is no output-shape change.

## Compatibility

Source-breaking for any external caller that set `ExpressionBodyArrowPlacement = …` (must become `WrapExpressionBodyArrow = true` for the former `NextLine`). All in-repo consumers are updated. Behavior and default output are unchanged.

## Evidence

- Build: 0 errors (3 pre-existing unrelated `Metadata.Tests` warnings).
- `StyleOptionCatalogTests` 11/0 (drift guard now balances 9 bool knobs ↔ 9 catalog entries; +1 focused test pins the new descriptor).
- `RenderStyleConfigTests` 41/0; `CSharpMemberLayoutTests` 20/0; `MemberBodyProducerMemberRenderTests` 13/0 (incl. the arrow-wrap render).
- Fast decompiler suite 3669/0.
- `markdownlint` clean.

Follow-on to #3138 / #3160. Adversarial review to follow (public API compatibility change).

### DIFF
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -503,9 +503,10 @@ a Wasm UI can enumerate the knobs (grouping the mutually-exclusive lenses by
 taste" aggregate is exactly the `OracleEndorsed` subset. The two guarded-boolean
 lenses share the `guarded-boolean-return` conflict group, so a picker offers at
 most one (the printer still resolves any overlap deterministically, preferring the
-oracle-endorsed ternary). The single non-boolean knob
-(`ExpressionBodyArrowPlacement`) is not yet modeled in the catalog, which
-currently describes boolean toggles.
+oracle-endorsed ternary). Every opt-in knob is a boolean toggle — including the
+expression-body arrow wrap (`WrapExpressionBodyArrow`) — so the catalog is
+exhaustive; a future non-boolean knob would need a descriptor shape that carries
+its value domain.
 
 ## Verification and soundness
 
--- a/src/ILInspector.Decompiler.Tests/MemberBodyProducerMemberRenderTests.cs
+++ b/src/ILInspector.Decompiler.Tests/MemberBodyProducerMemberRenderTests.cs
@@ -93,7 +93,7 @@ public sealed class MemberBodyProducerMemberRenderTests
             increment,
             AssemblyPath,
             pdbPath: null,
-            printerOptions: new PrinterOptions { ExpressionBodyArrowPlacement = ExpressionBodyArrowPlacement.NextLine });
+            printerOptions: new PrinterOptions { WrapExpressionBodyArrow = true });
 
         Assert.Equal(MemberBodyProductionStatus.Complete, rendered.Status);
         Assert.Equal("    public int Increment(int n)\n        => n + 1;", rendered.Text!.Replace("\r\n", "\n"));
--- a/src/ILInspector.Decompiler.Tests/StyleOptionCatalogTests.cs
+++ b/src/ILInspector.Decompiler.Tests/StyleOptionCatalogTests.cs
@@ -129,4 +129,21 @@ public class StyleOptionCatalogTests
         foreach (var o in Options.Where(o => o.OracleEndorsed && o.ConfigKey is not null))
             Assert.StartsWith("dotnet_style_", o.ConfigKey);
     }
+
+    [Fact]
+    public void WrapExpressionBodyArrow_IsAnApiOnlyFormattingKnob()
+    {
+        // The former ExpressionBodyArrowPlacement enum is now a boolean toggle, so
+        // it belongs in the catalog: a whitespace-only formatting choice with no
+        // config key and no oracle endorsement (the shipped default keeps the arrow
+        // on the same line; wrapping it is a user preference, not the oracle's).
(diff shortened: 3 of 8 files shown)


richlander:

Reviewed exact head `14d54400` (base `d131d951`) with two independent models, each isolated in a read-only worktree, each building + running the targeted suites and a render probe (not just reading), and each cross-checking output against the base.

| Reviewer | Verdict | Evidence |
| --- | --- | --- |
| GPT-5.6 | **CLEAN** | build 0; catalog 11/0; member-render 13/0; config 41/0; probe: `DEFAULT_EXACT=true`, `WRAPPED_EXACT=true`, `ONLY_ARROW_PLACEMENT_DIFF=true`; base outputs identical |
| Gemini 3.1 Pro | **CLEAN** | same suites green; probe confirms same-line default vs wrapped opt-in; audited serialization path |

All six attack points verified by both:

1. **Polarity/behavior** — `PrinterOptions.Default`/`DecompilerOptions.Default` still mean same-line, byte-identical; `true` reproduces the old `NextLine` rendering exactly. No inversion at either mapping site.
2. **Name-overlap** — `(printerOptions ?? Default).WrapExpressionBodyArrow` binds to the instance **property**, not the `MemberBodyProducer.WrapExpressionBodyArrow(...)` helper; member access can't misbind to the static method, no recursion, builds clean.
3. **Drift guard** — reflection reports 9 public instance bool properties ↔ 9 catalog entries; balances at 9==9. No test hardcodes the old count/id-list.
4. **Enum removal** — repo-wide grep finds only the intentional historical mention in a test comment; the public enum removal breaks no in-repo consumer (full slnx build succeeds).
5. **No output-shape impact** — `DecompilerOptions.EffectiveOptions` is not serialized by any CLI/JSON/markdown path (formatters project only rendered output/diagnostics/decisions); enum→bool has zero output impact.
6. **Catalog + docs** — new descriptor targets `WrapExpressionBodyArrow` with Formatting/ByteDivergent=false/OracleEndorsed=false/ConfigKey=null/ConflictGroup=null; GPT confirmed no editorconfig arrow-placement key exists (so not oracle-endorsed, no config key is correct). Docs' "catalog is now exhaustive" claim holds — all 9 opt-in knobs are boolean.

No blocking findings from either reviewer.

### CI note (non-blocking, unrelated flake)

The first `test` run failed on `CacheCommandTests.CacheMiss_CleansObsoleteVersionedCategories`, **not** on anything this PR touches. It passes 8/0 in isolation locally. The test drives process-wide global cache state — `CoreCache.RegisterVersionedCategory` plus a singleton cleanup task (`Assert.Same(cleanup, RequestVersionedCategoryCleanupAsync())`) — so under parallel full-suite execution a concurrent test can consume the singleton cleanup before this test's `oldDir` is created, leaving `oldDir` and failing `Assert.False(Directory.Exists(oldDir))`. It's an order/parallelism flake in pre-existing cache-command tests, independent of the printer/catalog change. Re-running the job to confirm green.

richlander:
Ready to merge

- CI green on re-run: `test` (8m16s), `build-net10`, `csharp-diff-smoke`, `markdownlint`, `changes` all pass; `pack`/`il-diff-smoke` path-skipped. The initial `test` failure was the unrelated `CacheCommandTests` global-cache-state flake (passes in isolation; independent of this diff), confirmed transient by the clean re-run.
- Both adversarial reviews (GPT-5.6, Gemini 3.1 Pro) CLEAN at the merged head `14d54400` with real-run evidence and base cross-checks; all six attack points verified; no blocking findings.

PR #3152, opened 2026-07-25
Surface applied byte-divergent style lenses as an Applied Taste section

richlander:
## Conclusion

Follow-up to #3127. That PR relied on a reader parsing opcode-fidelity prose to notice that a **Valid: True / Correct: True** render had actually *diverged* from the original opcodes. The two byte-divergent style lenses that cause this recorded **nothing** when they fired, so the signal was invisible to the tool. This PR makes the tool emit that signal directly and updates the decompiler-PR template to show it.

## Product change (decompiler)

- `PreferConditionalReturnPass` / `PreferBranchlessBooleanPass` now expose a `Rewrites` count.
- The main `CSharpPrinter.PrintRaised` overload appends a `DecompilerDecision` in a new `"style-lens"` category when a lens actually rewrote the body.
- **Additive and opt-in:** with the lenses off (shipped default) no decision is recorded and rendered output is byte-for-byte unchanged. The line-mapped `PrintRaised` overload (annotated view) is left byte-faithful.

## CLI

- New **explicit-only** `Applied Taste` section projects `DecompilerResult.Decisions`, one row per configurable choice, with a **Fidelity** column (`byte-preserving` | `byte-divergent`).
- Excludes the always-on `type-name.framework-imported` rewrite (universally expected, not a taste choice).

## Template

`docs/templates/decompiler-pr.md`:

- **Benchmark target** (`{lib}@{ver}`) + full `dotnet-inspect` command block before *Original source*.
- **IL fidelity / Taste applied / Commit** rows added to the Before and After verdicts, so the #3127 trap is surfaced rather than inferred from prose.

## Validation

- Decompiler fast suite: **3606 passed / 0 failed**; `StyleLensDecisionTests` **4/4**.
- CLI suite: **2045** (1 pre-existing environmental failure `PlatformResolverTests.GetPacksDirectory_ReturnsValidPath`, unrelated SDK-path assertion).
- `AppliedTasteSectionTests` + CLI plumbing tests **7/7**; end-to-end verified (lens-fired byte-divergent row, empty default note, schema listing, explicit-only, framework-import excluded).
- Default-preserving: the lens is opt-in, so `appliedLenses` is empty by default and `result` is returned unchanged — the compile-back Fidelity gate (default options) is unaffected.
- markdownlint clean.

## Review

[...]

PR #3172, opened 2026-07-25
Record this.-qualification knobs as byte-preserving Applied Taste decisions

### LINKED ISSUE #3156: Applied Taste: record byte-preserving this.-qualification knobs as decisions
## Background

PR #3152 added an explicit-only **Applied Taste** section that projects `DecompilerResult.Decisions` (one row per configurable style choice, with a `byte-preserving | byte-divergent` Fidelity column). Its primary purpose is surfacing the byte-divergent style lenses (the #3127 signal) so a reader doesn't have to infer opcode-fidelity loss from prose.

## Gap

The `this.`-qualification knobs are configurable and byte-preserving but currently record **no** `DecompilerDecision`:

- `dotnet_style_qualification_for_field` (`QualifyFieldAccess`)
- `dotnet_style_qualification_for_property` (`QualifyPropertyAccess`)
- `dotnet_style_qualification_for_method` (`QualifyMethodAccess`, #3097/#3123)
- `dotnet_style_qualification_for_event` (`QualifyEventAccess`, #3097/#3123)

Applied at `src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.Members.cs` (field ~line 51, property/event ~line 104, method ~line 252/363).

Because they emit `this.` without recording a decision, an enabled render (e.g. `dotnet_style_qualification_for_field = true`) adds `this._count` yet the Applied Taste section reports no corresponding row. PR #3152 scoped the empty-state wording to "recorded style choices" to avoid over-claiming, and deferred the recording work here.

## Ask

Record a byte-preserving `DecompilerDecision` (category `DecompilerDecisionCategories.Taste`) when a qualification knob actually adds `this.`.

### Design subtlety

`this.` is emitted for two distinct reasons and only the first is a configurable taste choice:

1. **Knob-driven** — `_options.QualifyFieldAccess` etc. (configurable; SHOULD record a decision).
2. **Mandatory disambiguation** — `QualifyThisMember(...)` forces `this.` when a local shadows the member (required for correctness, byte-preserving-but-not-optional; should NOT be reported as an applied *taste* choice, or should be reported distinctly).

The recorder must attribute qualification to the knob only when the knob (not disambiguation) is what added it, and dedup per member. Pair with tests mirroring `StyleLensDecisionTests` / `AppliedTasteSectionTests`.

## Evidence

Confirmed via `Collect_WithQualifyFieldAccess_QualifiesFieldAndRecordsEvidence` (emits `this._count`) that no qualification decision is recorded. Reported by an adversarial reviewer on PR #3152.

richlander:
## Conclusion

Closes #3156. Follow-up to #3152, which added the explicit-only **Applied Taste** section but scoped it to *recorded* choices because the four `this.`-qualification knobs emitted `this.` without recording a `DecompilerDecision`. This PR records them, so an enabled qualify knob now surfaces a byte-preserving Applied Taste row instead of silently changing the render.

## Product change (decompiler)

At the four knob emission sites in `CSharpPrinter.Members.cs` (field, property/event, method call, method group), record a byte-preserving `DecompilerDecision` (category `taste`) via `AddDecision` — **only when the knob, not mandatory disambiguation, added the qualifier**.

### Design subtlety

`this.` is emitted for two reasons and only the first is a taste choice:

1. **Knob-driven** (`QualifyFieldAccess` etc.) — configurable, byte-preserving → **recorded**.
2. **Mandatory disambiguation** (`QualifyThisMember`: a local/parameter shadow or member/type-name collision) — required for correct binding, appears with the knob off too → **never recorded** as a taste choice.

So the sole recorded case is `knob && !mandatory`. **Indexers** always render `this[...]` regardless of the knob, so they record nothing (the indexer early-return precedes the decision). `AddDecision` dedups per member.

Identity and config key come from `StyleOptionCatalog` (#3160), so the decision's `RuleId` (e.g. `qualify-field-access`) and prose stay in lockstep with the single source of truth; a knob rename fails at type init rather than drifting.

## CLI

No behavior change needed: `BuildAppliedTasteRows` already classifies category `taste` as `byte-preserving`, and `MemberCodeProvider` already forwards the full `Decisions` list rendered with the configured options. Refreshed the now-stale `#3156` comment.

## Default-preserving

The knobs are off by shipped default → no decision recorded → byte output unchanged. Decompiler fast suite **3678 / 0**, so the refactor of the four sites is byte-identical across the corpus.

## Validation

- `ThisQualificationDecisionTests`: **10** (per-kind recording, dedup, and the mandatory-disambiguation-off/on negatives).
- `AppliedTasteSectionTests`: added a byte-preserving qualification row case (**5** total).
- Existing `ThisQualificationTests` **23/23** (refactor preserves output); decompiler fast suite **3678/0**; Applied Taste e2e **3/3**.
- `docs/decompiler-taste.md` note added; markdownlint clean.

## Review

Behavior-capable change → two-model adversarial review required before ready. **Draft: review pending.**

### DIFF
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -230,10 +230,12 @@ array element type and rank, by-ref/pointer decoration, generic-parameter slot,
 function-pointer return type / calling convention / parameter ref-kinds, plus the
 full assembly-qualified namespace) so `M(List<int>)` and `M(List<string>)`,
 `M(NsA.Widget)` and `M(NsB.Widget)`, or `M(delegate*<int>)` and
-`M(delegate*<string>)`, stay two rows rather than collapsing into one. The key
-uses generic *arity*, not the specific type arguments, so two instantiations of
-one generic method — `this.G<int>()` and `this.G<string>()` — collapse into a
-single row (they are one source member).
+`M(delegate*<string>)`, stay two rows rather than collapsing into one. For a
+generic method the key uses the DEFINITION signature (its type parameter left as
+`!!0`) plus arity, not the MethodSpec-substituted parameters, so two
+instantiations of one generic method — `this.Echo<int>(1)` and
+`this.Echo<string>("s")` of `Echo<T>(T)` — collapse into a single row (they are
+one source member) even when a parameter mentions the type parameter.
 Qualifications inside a locals-bearing lambda body, which renders through an
 isolated nested printer, are not currently surfaced as taste rows.
 
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -182,9 +182,18 @@ that disambiguates a shadow or type-name collision would appear with the knob of
 too, so it is never attributed to the knob as a taste choice. Recording therefore
 applies only to a genuine instance receiver — a static or extension method whose
 first parameter is spelled `@this` (IL name `this`) reaches the same site but has
-no implicit receiver, so it records nothing. Method qualification is the most
-guarded because a bare method name binds through more rules than a field or
-property; it records a decision only when every one of these holds:
+no implicit receiver, so it records nothing.
+
+All four knobs also require the member to be declared on the **enclosing type at
+its own instantiation** before recording. A base-declared member reached through
+`this` — a field hidden by a `new` field of the same name (whose `base.X` load a
+pre-existing emit gap mis-spells `this.X`, but `this.X` binds to the *derived*
+field), or a merely-inherited field/property/event — is not a byte-preserving
+self-type opt-in, so it records nothing. This uniformly under-records a legitimate
+`this.` on an inherited member; a false-negative is safe, a false-positive is not.
+Method qualification is the most guarded because a bare method name binds through
+more rules than a field or property; it records a decision only when every one of
+these holds:
 
 - **enclosing type at its own instantiation.** The callee's declaring type must
   be the enclosing type at its exact instantiation, not merely the same generic
@@ -207,14 +216,24 @@ property; it records a decision only when every one of these holds:
   before `CSharpNaming.SourceMethodName` strips its `<...>` decoration (a lifted
   local function otherwise arrives spelled as a plain identifier and would slip
   past the check).
+- **not a generic method group.** A method *group* over a generic instance method
+  (`this.Make<int>` bound to a `Func<int>`) is not recorded: `MethodGroupText`
+  renders only the bare name, dropping the type arguments (a pre-existing emit gap
+  the call and `&`-of paths avoid), so the emitted `this.Make` fails delegate
+  return-type inference (CS0411) and does not round-trip. A generic method *call*
+  (`this.Make<int>()`) still records normally.
 
 Same-named overloads are recorded as distinct decisions; the dedup discriminator
-is a structurally complete per-overload key (generic instantiation, array element
-type and rank, by-ref/pointer decoration, generic-parameter slot, function-pointer
-return type / calling convention / parameter ref-kinds, plus the full
-assembly-qualified namespace) so `M(List<int>)` and `M(List<string>)`,
+is a structurally complete per-overload key (generic arity — so `M()` and `M<T>()`
+stay two rows despite sharing an empty parameter list — generic instantiation,
+array element type and rank, by-ref/pointer decoration, generic-parameter slot,
+function-pointer return type / calling convention / parameter ref-kinds, plus the
+full assembly-qualified namespace) so `M(List<int>)` and `M(List<string>)`,
 `M(NsA.Widget)` and `M(NsB.Widget)`, or `M(delegate*<int>)` and
-`M(delegate*<string>)`, stay two rows rather than collapsing into one.
+`M(delegate*<string>)`, stay two rows rather than collapsing into one. The key
+uses generic *arity*, not the specific type arguments, so two instantiations of
+one generic method — `this.G<int>()` and `this.G<string>()` — collapse into a
+single row (they are one source member).
 Qualifications inside a locals-bearing lambda body, which renders through an
 isolated nested printer, are not currently surfaced as taste rows.
 
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -186,13 +186,19 @@ no implicit receiver, so it records nothing. Method qualification is the most
 guarded because a bare method name binds through more rules than a field or
 property; it records a decision only when every one of these holds:
 
-- **same-type callee.** The callee's declaring type must be the enclosing type.
-  A cross-type callee reached here is either an inherited base method (whose
-  bare/`this.` form would rebind — the non-virtual case already renders `base.M`)
-  or an **explicit interface implementation** invoked through `this`, which does
-  not bind via `this.` at all (it requires a cast). Neither is a `this.` opt-in.
-  This deliberately under-records a legitimate `this.BaseMethod()`; a
-  false-negative is safe, a false-positive is not.
+- **enclosing type at its own instantiation.** The callee's declaring type must
+  be the enclosing type at its exact instantiation, not merely the same generic
+  definition. A callee that is not the exact self-type is one of: an inherited
+  base method (whose bare/`this.` form would rebind — the non-virtual case
+  already renders `base.M`); an **explicit interface implementation** invoked
+  through `this`, which does not bind via `this.` at all (it requires a cast); or
+  a **different instantiation** of the enclosing generic type — e.g.
+  `((I<object>)this).M()` from within `I<T>`, where bare/`this.` `M()` binds to
+  `I<T>::M`, not `I<object>::M`, so the qualifier is not byte-preserving.
+  Definition-only equality is too loose for the last case, so the guard reuses
+  the exact-instantiation test the static-call qualifier uses. This deliberately
+  under-records a legitimate `this.BaseMethod()`; a false-negative is safe, a
+  false-positive is not.
 - **not shadowed.** A name shadowed by an in-scope local, parameter, or nested
   lambda binder makes the `this.` mandatory disambiguation, not a choice.
 - **speakable target.** A compiler-generated group target — an unspeakable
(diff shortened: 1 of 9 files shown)


richlander:

Two fixed-head reviews at base `origin/main` (`d131d951`), head `9285f097`: **GPT-5.6** and **Gemini 3.1 Pro**. Both independently found the method-qualification recording lacked the mandatory/context discipline the field/property/event sites already had. Every finding was reproduced end-to-end via the CLI before acting.

| # | Finding (both/either reviewer) | Verdict | Resolution (`98cc7bda`) |
|---|---|---|---|
| 1 | Method `this.` recorded as taste even when a local/parameter **shadows** the method name (mandatory disambiguation) | **Confirmed** — reproduced: `MethodShadow` recorded `qualify-method-access` though bare `Foo` binds to a local delegate | Gate method recording on the scope-aware `IsStaticCallNameShadowed` (covers nested lambda/local-function binders too) |
| 2 | Static/extension method whose first param is spelled `@this` (IL name `this`) records taste + emits invalid `this.` | **Confirmed** (record); the **invalid emit is pre-existing** — the PR only added the record on top | Gate **all four** recording sites on a genuine instance receiver (`_function.Signature.HasThis`). Emit left unchanged (pre-existing printer receiver-spelling issue, tracked separately) |
| 3 | Same-named **overloads** collapse into one Applied Taste row (dedup key used only the display name) | **Confirmed** — `this.M(1); this.M("a")` produced one row | Add a per-overload dedup discriminator from the callee parameter types → two distinct rows |
| 4 | (found while reproducing) Compiler-generated captured-`this` lambda group target `<M>b__N` recorded as taste — pure noise | **Confirmed** | Skip unspeakable (`<…>`) names |
| — | Locals-bearing lambda body decisions discarded (isolated nested printer) → missing row | **Acknowledged** as a consistent *under*-claim (no false rows). The HasThis guard keeps it clean; documented as a known limitation rather than taking on the riskier nested-printer decision-merge |

Indexers-record-nothing and the CLI byte-preserving classification / `type-name.framework-imported` filter were both verified intact by the reviewers.

Evidence at `98cc7bda`: decompiler fast suite **3682/0**; `ThisQualificationDecisionTests` **14/14** (4 new negatives/overload); `ThisQualificationTests` **23/23**; CLI `AppliedTasteSectionTests` **5/5**; markdownlint clean. Re-review at the fixed head is running; not marking ready until both fixed-head reviews are clean.

richlander:
## Round-2 review reconciliation (head `d8df509a`)

The round-2 re-review (GPT-5.6 + Gemini-3.1, isolated read-only worktrees at the round-1 fixed head `98cc7bda`) confirmed all round-1 method-guard fixes hold and surfaced three residual method-path issues. All three are reproduced end-to-end and fixed in `d8df509a`.

| # | Finding | Sev | Verified | Resolution (`d8df509a`) |
| 1 | Overload discriminator `{Namespace}.{Name}` collapsed distinct overloads — `M(List<int>)`/`M(List<string>)` recorded **one** row, hiding a taste application (both reviewers; also `NsA.Widget`/`NsB.Widget`) | High | Yes — `CallGenericOverloads` emitted 1 row pre-fix, 2 post-fix | Replaced with a recursive structural `TypeKey(TypeRef)` (generic instantiation, array element/rank, by-ref/pointer/pinned, generic-parameter slot, assembly-qualified default) |
| 2 | `IsUnspeakableName` ran after `SourceMethodName` stripped `<...>`, so a captured-`this` **local function** (`<Outer>g__Local\|N_M`) was falsely recorded (Gemini) | Medium | Yes — `CallsCapturingLocalFunction` recorded a bogus `Local` row pre-fix, none post-fix | Unspeakable check now runs on the **raw** IL metadata name before sanitization |
| 3 | **Explicit interface implementation** invoked through `this` (`((IFoo)this).Bar()`) binds only via a cast, yet its `callvirt` recorded a false taste row (Gemini) | Medium | Yes — `CallExplicitInterface` recorded a bogus `Bar` row pre-fix, none post-fix | Recording now gated on the callee being declared on the enclosing type (`!IsCrossType`), which also safely under-records inherited base methods |

**Scope note:** the pre-existing printer emit that mis-spells explicit-interface / lifted-local-function calls as `this.Member()` (invalid C#) predates this PR and is out of scope; only the false *records* are suppressed. The `!IsCrossType` guard deliberately under-records legitimate `this.BaseMethod()` opt-ins — a safe false-negative, consistent with the never-over-record invariant.

Added specimen members + decision tests for all three (`GenericOverloadedMethodCalls_*`, `CapturingLocalFunction_*`, `ExplicitInterfaceCall_*`).

**Validation at `d8df509a`:** `ThisQualificationDecisionTests` + `ThisQualificationTests` 40/40; decompiler fast suite 3685/0; CLI `AppliedTasteSectionTests` 5/5; `docs/decompiler-taste.md` markdownlint clean.

Round-3 re-review of the fixed head `d8df509a` (GPT-5.6 + Gemini-3.1) is in progress; I'll reconcile it before marking ready.

richlander:
## Round-3 review reconciliation (head `01e19c9e`)

Round-3 re-review (GPT-5.6 + Gemini-3.1, isolated read-only worktrees at `d8df509a`) confirmed the round-2 fixes hold and surfaced two more method-path issues, both fixed in `01e19c9e`. A third (single-reviewer) finding is declined with rationale.

| # | Finding | Sev | Reviewer | Resolution |
| 1 | **Constructed generic self-call** `((I<object>)this).M()` from within `I<T>` was recorded — the round-2 same-type guard used definition-only equality (`IsCrossType`), but bare/`this.` `M()` binds to `I<T>::M`, not `I<object>::M` (not byte-preserving) | High | GPT | Guard now uses the **exact-instantiation** test (`IsEnclosingTypeAtOwnInstantiation`, renamed from `IsCurrentStaticScope` since it now also gates an instance path) — the same check the static-call qualifier uses. Still suppresses explicit-interface + inherited-base calls. Verified: 1 row → 0 |
| 2 | **Function-pointer key collapse** — `TypeKey`'s fnptr arm keyed on parameter types only, so overloads differing by fnptr return type / calling convention / param ref-kinds (`delegate*<int,int>` vs `delegate*<int,void>`) collapsed to one row, hiding a taste application | Medium | GPT + Gemini | New `FunctionPointerKey` includes calling convention, return type, and parameter ref-kinds. Verified: 1 row → 2 |

**Declined (with rationale):**

| Finding | Sev | Reviewer | Why declined |
| --- | --- | --- | --- |
| A non-generic `foreach` over `this` that the decompiler does not raise leaves a synthetic `this.GetEnumerator()` that is recorded | High | Gemini | Unlike the three suppressed cases (explicit-interface / lifted local-function / constructed-instantiation, where the emitted `this.M()` is invalid, does not bind bare, or rebinds), the `this.GetEnumerator()` qualifier here **is valid, binds bare to the same class method, and is byte-preserving** on the faithfully-rendered lowered code. The decompiler renders *what the IL does*; the row honestly annotates the `this.` the knob applied to that rendered call. When the pattern **is** raised (the common generic case), no such call is printed and nothing is recorded. Suppressing by method name would wrongly drop a legitimate explicit `this.GetEnumerator()` opt-in. Documented in `docs/decompiler-taste.md`. |

Added decision tests: `ConstructedGenericSelfCall_*` and `FunctionPointerReturnTypeOverloads_*`.

**Validation at `01e19c9e`:** `ThisQualificationDecisionTests` + `ThisQualificationTests` 42/42; decompiler fast suite 3687/0; CLI `AppliedTasteSectionTests` 5/5; `docs/decompiler-taste.md` markdownlint clean.

Round-4 re-review of `01e19c9e` (GPT-5.6 + Gemini-3.1) is in progress — including asking Gemini to either accept the decline or refute it with a concrete non-byte-preserving counter-example. I'll reconcile before marking ready.

richlander:
## Adversarial review reconciliation — rounds 4 & 5 (fixed-head reviews clean)

Two more review rounds since the round-3 reconciliation, both with GPT-5.6 and Gemini-3.1 in isolated read-only worktrees, reproducing before acting.

### Round 4 (at `01e19c9e`)
- **Gemini: CLEAN** — accepted the un-raised-`foreach` decline after testing; verified the fnptr key, exact-instantiation guard, and both record sites.
- **GPT: 3 findings**, all reproduced and confirmed as false-positive taste records (byte-divergent or non-binding `this.` that still recorded). Fixed in `98a44dab`:
  1. **Hidden/inherited base members (High).** Field/property/event recording never checked the member's declaring type. A base field hidden by a `new` field of the same name is loaded via `base.X` but a pre-existing emit gap mis-spells it `this.X`; `this.X` binds to the *derived* field → false positive. Gated all four knobs on `IsEnclosingTypeAtOwnInstantiation`, matching the method site. Uniformly under-records legitimate inherited-member `this.` (safe false-negative).
  2. **Generic method groups (Medium).** `this.Make<int>` bound to a `Func<int>` renders as `this.Make` (MethodGroupText drops the type args), which fails delegate return-type inference (CS0411) and does not round-trip → suppress recording for generic method *groups*. Generic method *calls* still record.
  3. **Arity overloads (Medium).** `M()` and `M<T>()` share an empty parameter list, so the parameter-types-only discriminator collapsed them into one row → folded generic arity into the overload key.

### Round 5 (at `98a44dab`)
- **Gemini: CLEAN** — verified all three round-4 fixes plus struct receivers and cross-type/nested-generic cases.
- **GPT: 1 finding.** The overload discriminator still split two *instantiations* of one generic method into duplicate rows when a parameter mentions the type parameter (`Echo<int>(1)`/`Echo<string>("s")` of `Echo<T>(T)`), because the callee's `ParameterTypes` are already MethodSpec-substituted. The round-4 arity fix only covered the no-parameter case. Fixed in `97b92f42`: retain the generic method **definition** signature (`MethodRef.DefinitionParameterTypes`, type parameter as `!!0`) at import and key the discriminator on it for generic method calls. Instantiations collapse to one row; distinct non-generic overloads `M(List<int>)`/`M(List<string>)` still key apart; `M()` vs `M<T>()` stay two rows on arity.

### Fixed-head re-review (at `97b92f42`) — both CLEAN
- **GPT: CLEAN** — original repro now records 1 row; adversarial overloads `Echo<T>(T)` vs `Echo<T>(T,int)`, `Shape<T>(T)` vs `Shape<T>(List<T>)`, `Pick<T,U>(T)` vs `Pick<T,U>(U)`, and `M<T>()` vs `M<T,U>()` all correctly stay 2 rows. Decision tests 26/26, rendering tests 23/23.
- **Gemini: CLEAN** — confirmed distinct generic overloads stay separated (`!!0` vs `!!0[]` placeholder keys), `DefinitionParameterTypes` populated only for MethodSpec calls (empty otherwise → safe fallback to concrete params), and no regression to the overload/instantiation/fnptr dedup behavior.

### Evidence
Each fix ships a specimen + decision test verified to fail without its guard. Fast decompiler suite 3694/0, this-qualification classes 49/49, CLI `AppliedTasteSectionTests` 5/5, markdownlint clean.

richlander:
Ready to merge

All merge-blocking validation is complete:
- Both fixed-head adversarial reviews (GPT-5.6 + Gemini-3.1) at `97b92f42` returned CLEAN; rounds 4 & 5 findings reconciled above.
- CI green: `test`, `build-net10`, `csharp-diff-smoke`, `markdownlint`, `changes` all pass (`pack`/`il-diff-smoke` path-gated skips). The earlier `test` failure was a transient GitHub Actions internal error and passed on re-run.
- Local: fast decompiler suite 3694/0, this-qualification classes 49/49, CLI `AppliedTasteSectionTests` 5/5.

PR #3173, opened 2026-07-25
Surface config warnings for Applied-Taste-only runs

richlander:
## Summary

**A bad `.dotnet-inspectconfig` must never fail silently.** The config-warning
latch keyed off a produced Decompiled Source `Output`, but an
**Applied-Taste-only** run consumes the same resolved config while leaving
`DecompiledResult` null — so its warnings were suppressed.

## Fix

- `MemberCodeProvider.Item` gains a `StyledProjectionProduced` signal, set when a
  config-consuming styled projection (**Decompiled Source or Applied Taste**,
  never the style-invariant fidelity-only read) actually printed a body.
- `ApiOutputFormatter` keys the warning latch off that signal instead of
  `DecompiledResult.Output`.

This preserves silence for runs that consume no config (callers-only aggregation,
fidelity-only projection, bodyless P/Invoke whose printer never ran).

## Evidence

Real-run, bad-key config (`nonsense_bogus_key`):

| Run | Config | stderr |
| --- | --- | --- |
| `-S "Applied Taste"` | bad key | `Warning: .dotnet-inspectconfig: line 2: unknown key 'nonsense_bogus_key' (ignored)` ✅ (was silent) |
| `-S "Decompiled Source"` | bad key | warning ✅ (no regression) |
| `-S "Applied Taste"` | clean | silent ✅ (no false positive) |

Tests: `RenderStyleConfigTests` 43/43 (augmented fidelity-only, Decompiled Source,
bodyless P/Invoke invariants with the new signal; added Applied-Taste-only positive
and bodyless-P/Invoke negative cases). Full solution builds (only pre-existing
Metadata.Tests warnings).

## Notes

Stacked on #3172 (`record-this-qualification-decisions`); base will retarget to
`main` after that merges. Draft pending adversarial review.

### DIFF
--- a/src/dotnet-inspect.Tests/RenderStyleConfigTests.cs
+++ b/src/dotnet-inspect.Tests/RenderStyleConfigTests.cs
@@ -491,6 +491,10 @@ public class RenderStyleConfigTests
         Assert.Null(withDefault.DecompiledResult);
         Assert.NotNull(withConfig.FidelityCauses);
         Assert.NotNull(withDefault.FidelityCauses);
+        // A fidelity-only projection consumes no config, so it never marks the
+        // warning latch regardless of whether a config was resolved.
+        Assert.False(withConfig.StyledProjectionProduced);
+        Assert.False(withDefault.StyledProjectionProduced);
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


richlander:

Two fixed-head reviews at base `record-this-qualification-decisions` head (`9285f097`, the stacked base at review time), head `309272bf`: **GPT-5.6** and **Gemini 3.1 Pro**. Both isolated in separate read-only worktrees.

**Result: both reviews clean — no blocking issues.** Each built the solution and ran `RenderStyleConfigTests` (43/43), and verified the must-preserve invariants:

- Runs that consume **no** config stay silent: callers-only aggregation (no projection), fidelity-only projection (`renderOptions == null`, style-invariant), and bodyless P/Invoke (printer never runs → `Output` null). Each evaluates `StyledProjectionProduced == false`.
- Both config-consuming modes (Decompiled Source **and** Applied Taste) surface a bad config once; `EmitOnce` latches correctly across a multi-member run regardless of member order.
- `StyledProjectionProduced` correspondence: `(DecompiledSource || AppliedTaste)` matches the `projectionRenderOptions` gate, and `&& projectionResult.Output is not null` keeps it false on a `Failure` payload — no missed warning, no spurious warning.
- `options?.RenderConfigWarnings?.EmitOnce()` null-safe when no config file exists.

No changes were required. This PR is stacked on #3172; it will not be marked ready until #3172's fixed-head re-review is clean and this branch is retargeted to `main`.

Issue #3185, opened 2026-07-25
Long member signatures vs. one-liner (expression-body) formatting: line-length-aware wrapping policy


richlander:
## Summary

We need a coherent policy for how a **long member signature** interacts with **one-liner (expression-bodied) formatting**. Today the decompiler prefers `=>` bodies and keeps the arrow on the same line, but it has no line-length-aware notion of "this signature is too long to be a comfortable one-liner." This surfaces on real BCL members such as:

```csharp
public static JsonElement Parse([StringSyntax("Json")] ReadOnlySpan<byte> utf8Json, JsonDocumentOptions options = default) => JsonDocument.ParseValue(utf8Json, options).RootElement;
```

That renders as a single very long physical line. The question is what the "tasteful" rendering should be, and which knobs (if any) should govern it.

## What the two oracle facets say (evidence)

Checked dotnet/runtime `.editorconfig` (`main`):

- `csharp_style_expression_bodied_methods = true:silent` (and the parallel `_constructors/_operators/_properties/_indexers/_accessors/_lambdas/_local_functions`, lines 112–119). Value is **`true`**, not `when_on_single_line` — the **declared** oracle prefers `=>` bodies *unconditionally* and does **not** say "fall back to a block body when long."
- No `max_line_length` and no arrow-placement/parameter-wrap rule (`.editorconfig` only carries `csharp_preserve_single_line_blocks`/`_statements`, which govern block statements, not arrow wrapping). So the declared oracle is **silent on how to wrap a long expression-bodied member**.
- **Revealed** corpus practice on long signatures: keep `=>` on the same line as the signature and wrap the **parameter list**, rather than relocating the arrow or switching to a block body.

So both facets agree the arrow body should exist; neither declares *how* to break the long line. This is why `wrap-expression-body-arrow` is classified as endorsed by neither facet (see #3179 / #3182 and `docs/decompiler-taste.md`).

## The gap

The existing `wrap-expression-body-arrow` knob relocates the **arrow**, which diverges from corpus practice. What is missing is a **line-length-aware wrapping** concept for long signatures, whose corpus-faithful behavior would be:

1. Keep the expression body (`=>`) — matches declared preference.
2. When the signature exceeds the target column budget, wrap the **parameter list** (one parameter per line, or hanging indent), keeping `=>` on the same line as the closing `)`.
3. Only consider a block-body fallback if explicitly opted in (the declared oracle does *not* ask for this).

## Questions to resolve

[...]

PR #3174, opened 2026-07-25
Raise switches whose case/default section contains a natural loop (#3161)

richlander:
- Fixes #3161 (advances tracking issue #3159; also improves #3162, #3163)
- Changes decompiler raising so a `switch` whose case/default section contains a
  natural loop is raised into a structured C# `switch` (previously left flat as a
  goto/label comparison chain)
- Card revision: `a8129d63`

## Change

> Should we accept this change?

**Conclusion:** **PASS** — the target method now raises to a structured switch
with structured loops, render-A/B is `valid->valid` with zero `valid->invalid`,
the corpus closure gate is unchanged, and the only fidelity-gate failures on
this machine are pre-existing/environmental (reproduced identically on the
base commit).

### Benchmark target

Benchmark target: `System.Text.Json@10.0.10`

dotnet-inspect command:

```bash
dotnet-inspect System.Text.Json.JsonElement.DeepEquals -S "Decompiled Source"
```

### Original source

```csharp
public static bool DeepEquals(JsonElement element1, JsonElement element2)
{
    if (!RuntimeHelpers.TryEnsureSufficientExecutionStack())
    {
        ThrowHelper.ThrowInsufficientExecutionStackException_JsonElementDeepEqualsInsufficientExecutionStack();
    }
    element1.CheckValidInstance();
    element2.CheckValidInstance();
    JsonValueKind kind = element1.ValueKind;
    if (kind != element2.ValueKind)
    {
        return false;
    }
    switch (kind)
    {
        case JsonValueKind.Null or JsonValueKind.False or JsonValueKind.True:
            return true;
        case JsonValueKind.Number:
            return JsonHelpers.AreEqualJsonNumbers(element1.GetRawValue().Span, element2.GetRawValue().Span);
        case JsonValueKind.String:
            // ...swap-and-unescape...
            return element1.ValueEquals(element2.ValueSpan);
        case JsonValueKind.Array:
            // ...length check + foreach comparison...
        default:
            // ...object property-count + enumerate-and-compare loop...
    }
}
```

### Before

```csharp
public static bool DeepEquals(System.Text.Json.JsonElement element1, System.Text.Json.JsonElement element2)
{
    int __switchValue0 = default;
    JsonValueKind kind;
    // ...locals...
    if (RuntimeHelpers.TryEnsureSufficientExecutionStack()) goto IL_000C;
    ThrowHelper.ThrowInsufficientExecutionStackException_JsonElementDeepEqualsInsufficientExecutionStack();
    IL_000C:
[...]

PR #3186, opened 2026-07-25
Give class finalizers a dedicated Finalizer section (#3157 follow-up)

richlander:
## Summary

Follow-up to #3168 (issue #3157). #3168 spells the finalizer body as `~Type()`
but the finalizer still carried Kind `explicit-interface-implementation`, so it
listed under **Explicit Interface Implementations** and raw `--shape` showed
`void Finalize()`. This PR gives the finalizer its **own `"finalizer"` kind and
dedicated `Finalizer` section**, symmetric with Constructors.

**Stacked on #3168** (base `fix/issue-3157-finalizer`) because it depends on
that PR's `IsFinalizer` / `.override System.Object::Finalize` detection. Rebase
onto `main` once #3168 merges.

## What changed

- **Metadata** (`ApiSurfaceExtractor`): classify as `"finalizer"` before the
  explicit-interface arm; the `.override` MethodImpl still suppresses
  accessibility.
- **Section model**: `SectionNames.Finalizer`, a `Finalizer` descriptor
  registered after `Constructors`, `IsMethodLike` includes `"finalizer"`.
- **Views** (`TypeView`): Finalizer row / select / summary sections + a
  Finalizer stat count.
- **Rendering** (`ApiOutputFormatter`): `MemberKinds` sort order,
  `GetTreeKindOrder`/`Label`, `PluralizeKind`, `GetRequestedMemberKinds`,
  `PopulateMemberSections`, `PopulateMemberSummarySections`; shape and summary
  spell `~Type()` / `Finalize` consistently with the `-v:n` section table.
- **Whole-type composition** (`MemberBodyProducer`) + member-body/`publicOnly`
  gates route finalizers through the method-like path so the `~Type()` body
  renders; `CSharpTypePrinter` `validBody` whitelist and `SharedParsers`
  `NormalizeKind` (`-k finalizer`) updated.
- **Commands**: `ApiCommand` section→predicate map + source-bearing filters;
  `MemberCommand` kind→section map.

## Verified live (fixture `class Handle { ~Handle() { _n = 0; } }`)

- `--shape` and default `-v:m`: `Finalizer (1) └─ ~Handle()` / `## Finalizer`.
- `--markdown -v:n`: dedicated **Finalizer** section (`Signature: ~Handle()`).
- `member Handle Finalize -S "Decompiled Source"`: `~Handle() => _n = 0;`.
- `--table`, `--json` (`kind: "finalizer"`), quiet stats (`Finalizer: 1`),
  `-k finalizer` filter.

## Tests

- Metadata classification: finalizer → `Kind == "finalizer"`; close negative
  explicit `IThing.Finalize()` stays `explicit-interface-implementation`.
- Shape `~Handle()` group; minimal-view Finalizer summary regression guard.
- Updated `ApiMemberPipeline` section count (29→30) and discoverability case.
- Green: Metadata 794, product 2054, CSharp 330, decompiler fast 3664.

## Review

Non-trivial behavior/output change (new kind, section, ~30 kind-switch sites) →
requires two-model adversarial review per repo policy.

### DIFF
--- a/.github/workflows/deep-inspect.yml
+++ b/.github/workflows/deep-inspect.yml
@@ -84,7 +84,9 @@ jobs:
 
   test:
     name: Test lane
-    if: inputs.lane == 'test' || inputs.lane == 'all'
+    # Runs on the weekly schedule (automated slow-gate coverage) and on manual
+    # dispatch (lane test/all).
+    if: github.event_name == 'schedule' || inputs.lane == 'test' || inputs.lane == 'all'
     runs-on: ubuntu-26.04
     steps:
       - uses: actions/checkout@v6
@@ -150,7 +152,8 @@ jobs:
     # Split out of the test lane so the multi-hour corpus sweep gets its own
     # 6h job budget instead of sharing the test job's budget with the other
     # slow gates (which previously pushed the combined job past its timeout).
-    if: inputs.lane == 'test' || inputs.lane == 'all'
+    # Runs alongside the test lane on the weekly schedule and on dispatch.
+    if: github.event_name == 'schedule' || inputs.lane == 'test' || inputs.lane == 'all'
     runs-on: ubuntu-26.04
     steps:
       - uses: actions/checkout@v6
@@ -483,3 +486,32 @@ jobs:
           name: deep-inspect-opt-in-corpus
           path: artifacts/opt-in-corpus/
           if-no-files-found: warn
+
+  report-scheduled-failure:
+    name: Report scheduled failure
+    # Deep Inspect has no interactive watcher, so a failed *scheduled* run would
+    # otherwise go unnoticed. Open (or comment on) a tracking issue when any
+    # scheduled slow lane fails. Manual dispatches are watched by the dispatcher
+    # and are intentionally excluded.
+    needs: [package-sweep, test, decompiler-corpus]
+    if: ${{ github.event_name == 'schedule' && failure() }}
+    runs-on: ubuntu-26.04
+    permissions:
+      issues: write
+    steps:
+      - name: Open or update the tracking issue
+        env:
+          GH_TOKEN: ${{ github.token }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          title="Deep Inspect scheduled run failed"
+          num=$(gh issue list --state open --search "$title in:title" \
+            --json number,title \
+            --jq "map(select(.title == \"$title\")) | .[0].number // empty")
+          if [ -n "$num" ]; then
+            gh issue comment "$num" --body "Another scheduled Deep Inspect run failed: $RUN_URL"
+          else
+            gh issue create --title "$title" --body $'The weekly Deep Inspect scheduled run failed.\n\nRun: '"$RUN_URL"$'\n\nThe failed lane(s) are visible on the run. This issue was opened automatically; it will not auto-close, so close it once the failure is triaged.'
+          fi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -90,8 +90,11 @@ jobs:
           echo "$ILASM_DIR" >> "$GITHUB_PATH"
           echo "$ILDASM_DIR" >> "$GITHUB_PATH"
 
-      - name: Run decompiler tests (including slow)
-        run: dotnet run --project src/ILInspector.Decompiler.Tests -c Release
+      - name: Run decompiler tests (including slow, excluding corpus sweep)
+        # The multi-hour corpus sweep runs in the decompiler-corpus job so it
+        # gets its own timeout budget; everything else (incl. slow round-trip /
+        # compile-back gates that need ilasm/ildasm above) runs here.
+        run: dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- --gate no-corpus
 
       - name: Run analysis tests (including slow)
         run: dotnet run --project src/ILInspector.Analysis.Tests -c Release
@@ -102,8 +105,40 @@ jobs:
       - name: Run IL round-trip tests (including slow sweep)
         run: dotnet run --project tests/DotnetInspector.ILRoundtrip.Tests -c Release
 
+  # The decompiler corpus sweep is multi-hour; give it its own 6h job budget
+  # instead of sharing deep-inspect-test's budget with the CLI, analysis, and
+  # round-trip suites (the combined job risked exceeding the timeout). Mirrors
+  # the deep-inspect.yml corpus split. It needs no ilasm/ildasm (the corpus
+  # oracle is Roslyn compile-back). build-native/build-portable/publish gate on
+  # this job too, so a corpus regression still blocks the release.
+  decompiler-corpus:
+    needs: resolve
+    runs-on: ubuntu-26.04
+    if: github.event.inputs.confirm == 'publish'
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: ${{ needs.resolve.outputs.sha }}
+
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v5
+        with:
+          dotnet-version: '11.0.x'
+          dotnet-quality: 'preview'
+
+      - name: Cache NuGet packages
+        uses: actions/cache@v4
+        with:
+          path: ~/.nuget/packages
+          key: nuget-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('**/*.csproj', '**/*.props', '**/*.targets', '**/*.slnx') }}
+          restore-keys: |
+            nuget-${{ runner.os }}-${{ runner.arch }}-
+
+      - name: Run decompiler corpus sweep
+        run: dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- --gate corpus
+
   build-native:
-    needs: [resolve, deep-inspect-test]
+    needs: [resolve, deep-inspect-test, decompiler-corpus]
     strategy:
       fail-fast: false
       matrix:
@@ -144,7 +179,7 @@ jobs:
   # TFM-agnostic pointer. These used to come from CI's pack job; they are now
   # built here at publish time so CI never produces release artifacts.
   build-portable:
-    needs: [resolve, deep-inspect-test]
+    needs: [resolve, deep-inspect-test, decompiler-corpus]
     runs-on: ubuntu-26.04
     steps:
       - uses: actions/checkout@v6
@@ -183,7 +218,7 @@ jobs:
           if-no-files-found: error
 
   publish:
-    needs: [resolve, deep-inspect-test, build-native, build-portable]
+    needs: [resolve, deep-inspect-test, decompiler-corpus, build-native, build-portable]
     runs-on: ubuntu-26.04
     if: github.event.inputs.confirm == 'publish'
     environment: nuget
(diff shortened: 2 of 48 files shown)


richlander:
## Adversarial review (this PR had no prior adversarial review) — whole-stack, frozen tip `2f8f24a1`

Two-model whole-stack review (GPT-5.6 + Gemini 3.1 Pro) of the cumulative `#3168 → #3186 → #3195` diff vs `main`. GPT-5.6 found three issues; I reproduced all three. Two are owned by this PR (below); the third is the writer suppression fallback, posted on #3168 but it only manifests because of **this PR's `Kind = "finalizer"`**, so its fix most likely lands here. Gemini cleared the stack, but its source-scan negative cases (`~0`, `~(x)`) didn't include the `~identifier` shape, so it missed Finding 2 below.

### Medium — named finalizer query renders no member section

`src/dotnet-inspect/Sections/ApiSectionDescriptors.cs` — `ApiMemberOverloadSectionDescriptors.CreatePipeline()`.

The overload-inventory pipeline registers Constructors/Fields/Properties/Methods/… but **not** `Finalizer`, even though the descriptor `ApiMemberSectionDescriptors.Finalizer` exists and is registered in the main type pipeline. So an unindexed finalizer query resolves the member yet renders nothing:

```
$ dotnet-inspect member --library ILInspector.Decompiler.Tests.dll \
    ILInspector.Decompiler.Tests.CfgSampleClass -m Finalize --all --markdown -v n
# ILInspector.Decompiler.Tests.CfgSampleClass          <- title only

$ ... -m Finalize -S Finalizer
Error: Select value 'Finalizer' not found.

$ ... -m Finalize --index 1 -S Signature
## Signature ... ~CfgSampleClass()                      <- resolves fine when a section is forced
```

**Fix:** add `.Add<ApiMemberSectionDescriptors.Finalizer>()` to `CreatePipeline()` (adjacent to `Constructors`), plus command-level coverage for an unindexed finalizer query.

### Medium (narrow) — source-scan stop truncates on a `~identifier;` continuation

`src/ILInspector.Metadata/SourceLinkResolver.cs` — `IsDestructorSignatureLine` (used by `ExtractMethodBody`).

The predicate accepts any line of the form `~Identifier…`. Its doc comment claims immunity to `~0)`/`~(x)`, but a bitwise-complement **statement/continuation** like `~mask;` starts with `~identifier` and matches. When such a line sits above the first visible sequence point (e.g. under `#line hidden` in a destructor body), the backward scan stops there and truncates the real `~Type()` signature and preceding body. Reproduced against `ExtractMethodBody`:

```
input body:  ~C() { #line hidden  int value = \n ~mask;  #line default  GC.KeepAlive(value); }
extracted:   "~mask;\n#line default\n        GC.KeepAlive(value);\n    }"   <- ~C() dropped
```

**Reachability** is narrow (requires `#line hidden` inside a destructor body before the first sequence point), hence Medium. **Fix (safe):** require the declarator shape `~Identifier(` (optional whitespace/generics before `(`) rather than any `~Identifier`; a complement statement ends in `;` with no parameter list. This distinguishes declarator from statement **without** reintroducing the round-4 name-match false-negative that keyed off the type name.

### Finding on #3168 (cross-ref)

The suppressed-finalizer fallback now emits `public virtual void Finalize()` (CS0465, new slot) because a suppressed finalizer with this PR's `Kind = "finalizer"` falls through to the normal modifier path. Details + repro: #3168 (comment). The fix likely belongs in this PR since `Kind="finalizer"` is introduced here; the current suppression test uses `explicit-interface-implementation` and no longer exercises the stacked path.

My follow-up **#3195** is clean under both models but is blocked on this stack.

richlander:
## Adversarial review reconciliation (Gemini Pro `gemini-3.1-pro-preview` + GPT `gpt-5.6-sol`)

Both models reviewed the finalizer **source-extraction** hardening iteratively at each fixed head. The final head `69a6f7a4` is **CLEAN per Gemini**; GPT's remaining two findings are accepted as documented out-of-scope limitations (rationale below). Every concrete, reachable finding raised across the rounds was reproduced and fixed on a real artifact.

### Findings → resolution

| Round | Finding (model) | Resolution |
| --- | --- | --- |
| 2 | Destructor source leak: backward scan walked past `~Type()` into the preceding member; nested-generic ctor spelling (both) | `7b625cde` |
| 3 | Forwarded-type token guard: foreign token resolved against the facade PDB (GPT) | `e75daa6e` (compare vs `apiType.SourceAssemblyPath ?? apiDllPath`) |
| 4 | Case-insensitive/wildcard finalizer selector bypassed the name-gated stop (GPT) | `ebcbb5ac` |
| 5 | Shape-based `~`-stop false-positive/negative (multiline default param, spaced `~ C()`) (GPT) | `2b39a474` — caller-supplied `isDestructor` identity |
| 6 | `unsafe ~C()` leak; ordinary method named `Finalize` misclassified on the diff path (Gemini + GPT) | `e11f29ea` — authoritative `ApiSurfaceExtractor.IsFinalizerMethod` (metadata `object.Finalize` override), threaded through the source mapping |
| 7 | `.Contains('~')` mistook a `#line hidden` body complement `int x = ~0;` for the signature (GPT) | `20ad84ea` — grammar matcher (not any tilde) |
| 8 | Empty-paren grammar re-leaked `~mask;`/`~Compute(x);` under `#line hidden` (both); escaped `~\u0043()` rejected; wrapped-paren signature false-negatived (both) | `619c8fb5`, then `55b882c4` |
| 9 | **Root fix** both models recommended: validate against the declaring **type name** (GPT + Gemini) | `55b882c4` — `~TypeName` token then `(`-or-EOL; decodes `\uXXXX`/`\UXXXXXXXX`; `SimpleTypeName` strips ns/nesting/arity; callers pass API type name / `MemberAnchor.TypeFullName` |

**Real-artifact evidence** (round-9): a compiled portable-PDB `#line hidden` destructor whose *destructor* first visible sequence point is line 12 (past a `~Compute();` body complement on line 10) anchors extraction at `~C()`, not `~Compute()`, and does not leak the preceding field. Suites green: metadata 815, services 299, product 2097.

### Gemini — CLEAN at `55b882c4`
Verified the type-name anchoring resolves the wrapped-paren false-negative and every `#line hidden` complement false-positive; independently confirmed `~@C()`, `\uXXXX`/`\UXXXXXXXX` decoding, the authoritative `isDestructor` gate, and `SimpleTypeName`. Quote: "bulletproof resilience against source leaks."

### GPT — two accepted out-of-scope residuals (documented in `69a6f7a4`)
GPT's final two cases are exotic valid-C# that a single-line text heuristic cannot resolve without a multi-line tokenizer (which this Roslyn-free, SRM-only path deliberately avoids, consistent with `ExtractMethodBody` being an approximate heuristic — both models already accepted its pre-existing single-line `sig {}` limitation as out of scope):

1. A comment between the tilde and the type name (`~ /*x*/ C()`).
2. A body statement complementing an invocation of a **local that shadows the enclosing type name** (`~C();` with a local named `C` in scope) when `#line hidden` makes it the first visible sequence point.

Both require a local/member spelled *exactly* as the enclosing type inside a hidden-line body — combinations that do not occur in real destructors. Type-name anchoring is strictly better than the pre-fix behavior (which always leaked) and every prior round. These limitations are now documented in `SourceLinkResolver.IsDestructorSignatureLine`.

richlander:
## Review findings resolved — commit `0f5491b7`

All three findings from the whole-stack adversarial review are now addressed on this branch.

- **Finding 2 (overload-section gap) — FIXED.** `ApiMemberOverloadSectionDescriptors.CreatePipeline()` now registers the `Finalizer` descriptor alongside `Constructors`. Verified end-to-end:
  ```
  $ dotnet-inspect member --library ILInspector.Decompiler.Tests.dll \
      ILInspector.Decompiler.Tests.CfgSampleClass -m Finalize --all --markdown -v n
  ## Finalizer
  | Name | Digest | Signature |
  | Finalize | dfc16a4893 | ~CfgSampleClass() |
  ```
  `-S Finalizer` now resolves too. New guard: `SectionPipelineTests.MemberOverloadPipeline_WithFinalizer_DiscoversFinalizerSection`.

- **Finding 1 (suppressed-finalizer modifiers) — FIXED.** The finalizer modifier branch in `CSharpDeclarationWriter` now covers the suppressed case, so a `Kind="finalizer"` finalizer with `SuppressFinalizerSpelling` renders modifier-free `void Finalize()` instead of `public virtual void Finalize()` (which was a new virtual slot / CS0465). New guard: `CSharpDeclarationWriterTests.FinalizerMember_WithFinalizerKindAndSuppression_StaysModifierFree`.

- **Finding 3 (source-scan `~identifier` truncation) — already addressed** on this branch by commits `619c8fb5` (require empty-paren destructor grammar) and `55b882c4` (anchor on declaring type name). The current `IsDestructorSignatureLine` rejects `~mask;` in both the authoritative (type-name) and fallback (`~Identifier()`) modes; the residual `~C();`-shadowing-under-`#line hidden` case is documented as accepted out-of-scope.

Validation on the new head: CSharpDeclarationWriterTests 46/46, SectionPipelineTests 125/125, CommandExecutionTests 573 (2 skip), ApiOutputFormatterTests 30 (1 skip).

Since this changes behavior and this PR previously had no clean two-model adversarial review at a single frozen head, it should get a re-review at this exact head (`0f5491b7`) before merge. #3195 (stacked above) will re-sync onto this head.

richlander:
## Whole-stack adversarial review — clean

Ran a cumulative two-model adversarial review of the finalizer stack at frozen heads:

- **#3168** `fix/issue-3157-finalizer` @ `32078265`
- **#3186** `fix/finalizer-section` @ `0f5491b7`
- **#3195** `fix/finalizer-json-async-cleanup` @ `496847b6` (tip; contains the other two)

Reviewers (both non-Claude, isolated detached worktrees at the tip, scratch under `/tmp`, no git mutations):

- **GPT-5.6** — *No changes required.*
- **Gemini 3.1 Pro** — *No significant issues found.*

Both **independently reproduced** the three prior findings and confirmed the fixes hold at the tip:

1. Suppressed finalizer with `Kind="finalizer"` renders modifier-free `void Finalize()` (no `public`/`virtual`/`async`; CS0465 slot avoided).
2. `Finalizer` section is discoverable/selectable in the member-overload pipeline (`member -m Finalize --all` and `-S Finalizer` both render `## Finalizer`).
3. `SourceLinkResolver` destructor extraction rejects the `~identifier;` (`~mask;`) `#line hidden` continuation and still extracts anchored `~Type()` destructors. The residual bare-`~Type` (no semicolon, exact type-name match) intersection remains the already-documented accepted out-of-scope limitation.

Local validation at the merged tip: `CSharpDeclarationWriterTests` 47/47, `SectionPipeline`+`ApiOutputFormatter` 156 (1 skip), all green.

Note: repo CI is `pull_request: [main]`-gated, so stacked PRs (base ≠ main) show no checks until retargeted. CI will run as the stack merges bottom-up (#3168 → #3186 → #3195).

richlander:
## Re-synced with main (post-#3168 merge)

#3168 merged to `main` (squashed as `b0b92986`); #3185/#3203 and #3174 also landed. Integrated `origin/main` into this branch.

**Conflict resolution** (mechanical — kept this PR's reviewed content over the pre-fix squashed copies; no new logic):

- `CSharpDeclarationWriter.cs`: kept the unconditional `else if (member.IsFinalizer)` (Finding 1 fix). Main carried the older `&& !SuppressFinalizerSpelling` guard because #3168 was squashed *before* Finding 1 was fixed here.
- `ApiSurfaceExtractor.cs`: retained the shared `IsFinalizerMethod` metadata-identity helper (added by this PR, absent from squashed #3168).
- Test files: kept the nested-generic destructor/ctor, finalizer-`Kind`, and suppressed-modifier-free regression tests.

**Validation** at the merged head: full solution builds; targeted suites green — CSharpDeclarationWriter 46, SectionPipeline+ApiOutputFormatter 155 (1 skip), Decompiler expression-body 12, Metadata 815.

The clean whole-stack two-model adversarial review still applies: the merge integrates main and resolves conflicts by preserving the already-reviewed finalizer behavior; no reviewed logic changed.

PR #3182, opened 2026-07-25
Model revealed-preference corpus-endorsement axis (#3179)

richlander:
## What & why

Adds a `CorpusEndorsed` boolean to `StyleOptionDescriptor` alongside the existing `OracleEndorsed` flag, modeling the **two oracle facets** the taste model relies on (named in the doc by #3181):

- **`OracleEndorsed` (declared)** — the runtime `.editorconfig`/IDE fixers endorse the spelling. Drives the "full taste" aggregate.
- **`CorpusEndorsed` (revealed)** — the declared oracle is silent, but the runtime's own source corpus reveals a dominant practice for the knob.

The two flags are **independent by contract**. This first, conservative pass marks exactly **one** knob revealed-endorsed: `wrap-splittable-expressions` (runtime wraps long boolean chains, matching its 120-column practice). Every other formatting/synthesis knob stays un-endorsed on both axes — notably the expression-body arrow wrap, which actually *diverges* from the corpus (runtime keeps `=>` on the same line), and readable-local-names, which is our own synthesis invention, not a corpus spelling.

Each `CorpusEndorsed = true` is a **deliberate, documented judgment**, never a silently-inferred or measured-heat claim (per AGENTS.md evidence discipline).

## Changes

- `StyleOptionCatalog`: `CorpusEndorsed` field + 9 classifications (only wrap-splittable = true) + `CorpusEndorsedOptions` accessor (declared after `Options`, mirroring `OracleEndorsedOptions`). Tightened `OracleEndorsed` doc to say **declared** and note orthogonality.
- Tests: pin `CorpusEndorsedOptions` to literal id `["wrap-splittable-expressions"]` (independent of the flag the production filter reads); assert the two facets are read independently; assert branchless and arrow-wrap are endorsed by neither facet.
- `docs/decompiler-taste.md`: reconciled the Option-catalog prose, which previously over-claimed arrow-habits and readable-names as corpus-legible; added `CorpusEndorsed` to the descriptor-field list.

## Scope / non-goals

Delivers **only** the classification model + accessor. The "house style" aggregate (declared ∪ revealed / `ApplyHouseStyle`), any config key for formatting knobs, and any measured-corpus signal are explicit follow-ups tracked in #3179.

## Evidence

- Reflection-free; no `PrinterOptions` change, so the drift guard (`EveryBooleanPrinterOption_HasExactlyOneCatalogEntry`, 9 bool props == 9 descriptors) is unaffected.
[...]

Issue #3169, opened 2026-07-25
Opt-in taste: broaden var / target-typed new spelling + the apparency analysis that unlocks more of editorconfig


richlander:
## Summary

Build the opt-in analysis and spelling infrastructure for the "how is a type spelled at a declaration/creation site" family — `var` vs explicit type, and target-typed `new(...)` beyond its current faithful-default scope. The interesting part is the shared substrate: a per-site **"type is apparent"** analysis plus **multi-value / correlated catalog descriptors**. That substrate is what lets us cover a much larger slice of dotnet/runtime's `.editorconfig` than the boolean lenses do today.

Follow-on to the style-lens design (#3138) and the option catalog that just landed (#3160).

## Where we are today

Two adjacent axes are in very different states:

- **Target-typed `new()` — already shipped, as a *faithful, always-on default*.** `TargetTypedNewText` (`src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.cs:~4152`) shortens `T x = new T(args)` → `T x = new(args)`. It is IL-identical (the target type fixes the constructed type and therefore the `newobj` overload), so it is applied unconditionally where provably byte-safe — it is **not** a `PrinterOptions` knob. Scope is deliberately narrow: only single-target assignment/declaration sites (local/field/property/stack-slot/indirect/array-element store), never call arguments (a target-typed `new` there joins overload resolution and could change binding), and return positions are explicitly out of scope. Arrays, object/collection initializers, tuple/nullable/base/interface targets, and `object`/`dynamic` all decline. Matches `csharp_style_implicit_object_creation_when_type_is_apparent`.
- **`var` — not emitted at all.** The printer always spells the explicit type at declarations via `TypeText(type)` (`CSharpPrinter.cs:~705-708`; declaring-store sites likewise). `var` is byte-neutral (purely syntactic, no IL consequence) but is a pure taste call, so the shipped default stays classic/explicit.

So the split is: `new(...)` shortening exists as a faithful default; `var` is the genuinely missing, opt-in-only piece.

## Ask

1. **`var` spelling as an opt-in lens.** Emit `var` where the editorconfig taste says to, behind catalog knob(s). Because it is byte-neutral, `PrinterOptions.Default` stays explicit and byte-stable; the lens only changes spelling. Fidelity requirement is behavioral only (100%), not byte-identity — same contract as the other opt-in lenses.
[...]

PR #3189, opened 2026-07-25
Generalize StyleOptionDescriptor to carry a value domain (#3169 substrate)

richlander:
## What & why

First slice of #3169. Generalizes `StyleOptionDescriptor` from a **boolean-only**
knob into one that carries an explicit **value domain**, the enabling substrate
the eventual `var` / target-typed-`new` taste family needs. Proven
non-speculatively by re-modeling the two coupled guarded-boolean-return lens
knobs as **one tri-state axis** (`flat` / `conditional-expression` /
`branchless`) instead of two independent booleans.

## Design (low-risk, zero behavior change)

- **`StyleOptionValue`** `{ Token, Title?, OracleEndorsed, CorpusEndorsed, ConfigKey?, IsSelected, SetSelected }`.
  Endorsement (both the declared `OracleEndorsed` and the revealed
  `CorpusEndorsed` facet added in #3182) and the config key move **per-value**.
- **`StyleOptionDescriptor`** gains `DefaultValue` / `Values` and
  `GetValue`/`WithValue` (single-select over the axis), keeping
  `Get`/`ConfigKey`/`OracleEndorsed`/`CorpusEndorsed`/`EndorsedValue` as computed
  back-compat. A `Boolean(...)` factory keeps the seven two-state knobs terse.
- The guarded-boolean-return family is **one 3-value axis over the same two
  backing `PrinterOptions` bools**: each value's `SetSelected` drives only its own
  bool (last-write-wins per key; the printer still resolves any overlap →
  ternary), and `GetValue` reports the oracle-endorsed `conditional-expression`
  first. So config resolution and printer output are **byte-identical**.
- **No `PrinterOptions` API change, no default-output change.** Dropped
  `ConflictGroup`/`GuardedBooleanReturnGroup` (subsumed by the single axis).
- `RenderStyleConfig` now resolves keys from **per-value** `ConfigKey`s via
  `SetSelected`; `ApplyFullTaste` selects each axis's `EndorsedValue`.

## Evidence

- `StyleOptionCatalogTests` **24/0** (19 rewritten for the value-domain model +
  the 5 corpus-facet tests from #3182, adapted to the per-value shape). The drift
  guard now asserts **every backing `PrinterOptions` bool is reachable through
  some catalog value** (stronger than the old 1:1 count).
- `RenderStyleConfigTests` **48/0, unchanged** — direct proof config behavior is
  byte-identical (single-key, both-key, `full_taste`, last-write-wins,
  warning paths all still pass without edits).
- Fast decompiler suite **3708/0** (`-trait- "Speed=Slow"`).
- `docs/decompiler-taste.md` Option-catalog section rewritten for the value
  domain; markdownlint clean.

## Scope boundary

Substrate only. The `var` family and its per-site apparency category dimension
(built-in / apparent / elsewhere) remain a future slice of #3169.

Rebased onto latest `main` (integrates #3182's `CorpusEndorsed` facet into the
per-value model).

Copilot-Session: 47752e34-b249-4ec5-82bb-9b28b29259e5

### DIFF
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -541,7 +541,7 @@ dotnet_style_prefer_conditional_expression_over_return = true
   and the ternary lens — everything the runtime `.editorconfig`/IDE oracle
   prefers) so a user need not copy each `dotnet_style_*` line. It deliberately
   excludes the non-endorsed branchless lens, so the guarded-boolean-return
-  conflict group resolves to the ternary deterministically. It applies in file
+  axis resolves to the ternary deterministically. It applies in file
   order like any other key, so a later explicit per-knob line overrides it
   (last-write-wins) — `full_taste = true` then
   `dotnet_style_qualification_for_field = false` is "full taste minus one knob".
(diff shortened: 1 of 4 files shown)


richlander:

Two fixed-head reviews at exact head `f67f7392`, each in an isolated read-only worktree. Both **clean**, both backed by real runs.

**GPT-5.6** — clean. Verified in its worktree:
- Release build clean (0W/0E).
- Config/render probe: both guarded-key orders → `cond=true branch=true`, ternary wins (`ternary=true branchless=false`); rendered specimen byte-equal to base's pinned ternary-wins (`PreferBranchlessBooleanLensTests.cs:155`). `full_taste=true` → `cond=true branch=false`.
- `StyleOptionCatalogTests` 24/0, `RenderStyleConfigTests` 48/0, `PreferBranchlessBooleanLensTests` 27/0.

**Gemini-3.1** — clean. Verified in its worktree:
- Release build clean (0W/0E).
- Probe over the built assemblies: both guarded-key orders (ternary-first and branchless-first) → both backing bools `true`; ternary pass runs before branchless (`CSharpPrinter`), so it consumes the shape and the branchless pass is a no-op — ternary wins identically to base. `full_taste=true` and `ApplyFullTaste()` → `cond=true branch=false`.

Both independently confirmed the core behavior-equivalence claim: because per-value `SetSelected` isolates each value's own backing bool, setting both guarded keys sets both bools (order-independent) exactly as base did, and the printer's existing ternary-first precedence resolves the overlap — so config resolution and rendered output are byte-identical. No findings to action.

## Status

- Behavior-equivalence: proven (config tests 48/0 unchanged; both reviewers reproduced both-key + full_taste equivalence).
- Fast decompiler suite 3708/0; catalog tests 24/0; markdownlint clean.
- CI green: `build-net10`, `test`, `markdownlint`, `csharp-diff-smoke`, `changes` all pass; `il-diff-smoke`/`pack` path-skipped as expected.

Ready to merge

PR #3194, opened 2026-07-25
Add shared per-site "type is apparent" analysis (#3192)

### LINKED ISSUE #3192: Per-site "type is apparent" analysis: shared decompiler predicate for var / target-typed new taste
## Summary

Introduce a per-site **"type is apparent"** analysis — a single decompiler-owned predicate that answers *"does this declaration/creation site make its type syntactically obvious from the right-hand side?"* — and route the printer's two existing, currently-independent apparency decisions through it. This is the shared substrate #3169 names as the real point of the `var` / target-typed-`new` taste family: both editorconfig axes (`csharp_style_var_when_type_is_apparent`, `csharp_style_implicit_object_creation_when_type_is_apparent`) hinge on the same notion, and today we have no reusable model of it.

Sub-issue of #3169. The catalog-shape prerequisite (multi-value / correlated descriptors) landed in #3189; this is the analysis prerequisite. Neither the `var` lens nor a broadened target-typed `new` is honest without it.

## Where we are today

The two apparency decisions exist but are ad-hoc and unconnected:

- **Target-typed `new` already computes an apparency-equivalent inline.** `TargetTypedNewText` (`src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.cs:4200`) shortens `T x = new T(args)` → `T x = new(args)` only when the target type *exactly* equals the constructed type and `IsTargetTypedNewEligible(target)` (`CSharpPrinter.cs:4232`) admits it (plain class/struct/generic-instance; declines `Nullable<T>`, `ValueTuple`, arrays, pointers, `object`/`dynamic`). That eligibility-plus-exact-equality check *is* an apparency judgment — but it is private to this one method and phrased for the `new()` case only.
- **`var` is emitted only where mandatory.** `DeclarationTypeText(type, initializer)` (`CSharpPrinter.cs:5325`) spells `var` **only** for anonymous-object initializers (which cannot be spelled otherwise) and falls back to `TypeText(type)` everywhere else. There is no notion of "the type is apparent, so `var` would be legitimate here."

So the apparency concept is implicitly duplicated (once, narrowly, for `new()`) and entirely absent for `var`.

## Ask

1. **A per-site apparency predicate**, owned in the decompiler, that classifies a declaration/creation site by whether the initializer makes the declared type obvious. Intended to align with dotnet/runtime editorconfig semantics for the two `*_when_type_is_apparent` keys. Candidate "apparent" shapes (design to confirm against Roslyn's IDE0007/IDE0008 behavior):
   - object creation of the exact declared type (`new T(...)`, including the target-typed form),
   - explicit cast / `as` to the declared type,
   - `default(T)` and typed `typeof`/array-creation forms,
   - (stretch, matching Roslyn's "intrinsic/apparent factory" carve-out) a small, well-defined factory-name family — to be justified case-by-case, not guessed.
2. **Route the existing consumers through it.** `TargetTypedNewText`'s eligibility+equality judgment should be expressible in terms of (or delegate to) the shared predicate, and `DeclarationTypeText` should gain the *hook* where a future `var` lens consults it. This issue does **not** turn `var` on — it makes the site classifiable.

## Explicit non-goals (kept in #3169 / later slices)

- **No `var` spelling change.** `PrinterOptions.Default` stays fully explicit and byte-stable; the analysis is inert until a lens consumes it.
- **No new catalog descriptor / editorconfig key wiring.** The value-domain catalog shape already landed (#3189); wiring the `var` family keys is a follow-on.
- **No broadening of target-typed `new`'s applied scope.** Refactoring its check to sit on the shared predicate must be behavior-preserving (same sites shorten, byte-identical).

## Constraints

- Product paths stay **SRM-only, NativeAOT-friendly, Roslyn-free**. The predicate is a pure function over the typed IR (`IrExpression` + `TypeRef`); no Roslyn syntax/semantic model, no inspected-assembly loading.
- **Behavior- and byte-safe by construction.** With no lens consuming the `var` hook, default output is unchanged; the target-typed `new` refactor must be a pure refactor (prove byte-identity on the corpus).
- Layer ownership: the apparency notion is IL/C#-spelling evidence, so it belongs with the decompiler printer/IR, not Metadata or Research.

## Evidence plan

- Focused decompiler tests asserting the predicate's classification on positive sites (object-creation-of-exact-type, cast, `default(T)`) and close negatives (base/interface target, `Nullable<T>`, `ValueTuple`, `object`/`dynamic`, factory returning a different type).
- Prove the `TargetTypedNewText` refactor is byte-identical: existing target-typed-`new` tests unchanged + a corpus/round-trip pass showing no rendered delta.
- No default-output change, so no new fixture is promoted into the `-v:m` view.

Related: #3169 (parent), #3189 (value-domain catalog substrate), #3138 (opt-in style-lens design), #3160 (StyleOptionCatalog).

richlander:
## What

Introduces `TypeIsApparent(TypeRef declaredType, IrExpression initializer)` — a decompiler-owned, pure predicate that classifies whether a declaration/creation site names its type directly. This is the analysis substrate #3169 identifies as the real point of the `var` / target-typed-`new` taste family: both editorconfig axes (`csharp_style_var_when_type_is_apparent`, `csharp_style_implicit_object_creation_when_type_is_apparent`) hinge on the same notion, and we had no reusable model of it.

Closes #3192 (sub-issue of #3169). The catalog-shape prerequisite landed in #3189; this is the analysis prerequisite.

## Changes

- **New predicate** (`CSharpPrinter.cs`): `TypeIsApparent` reports apparent for object creation of the exact type (`new T(...)`), single-dimension array creation (`new T[n]`), and an explicit reference cast (`(T)x`). Deliberately conservative — numeric conversions and target-typed `default` are **not** apparent, so a future `var` lens can never spell `var` where the type would silently vanish.
- **Route the one live consumer through it**: the target-typed-`new` shortener's inline `target.Equals(constructedType)` clause is exactly the object-creation apparency case, now expressed via the shared predicate.
- **Mark the `var` seam**: `DeclarationTypeText` documents where the planned opt-in `var` lens will consult the same predicate, so both spelling axes agree per site. No `var` behavior is turned on.

## Why it's safe

- **No default-output change.** The `var` hook stays inert (no lens consumes it); `PrinterOptions.Default` remains fully explicit and byte-stable.
- **Target-typed-`new` refactor is byte-identical.** The predicate's `NewObject` branch is `declaredType.Equals(creation.Constructor.DeclaringType)` — the same clause it replaced. `TargetTypedNewText` only ever reaches that branch (it guards `value is NewObject creation` first), so output is unchanged.
- Product path stays **SRM-only, NativeAOT-friendly, Roslyn-free** — a pure switch over typed IR nodes.

## Non-goals (later slices of #3169)

- No `var` spelling change and no new catalog descriptor / editorconfig-key wiring.
- No broadening of target-typed-`new`'s applied scope.
- `default(T)`, numeric conversions, and factory-name apparency are documented extension points, not delivered here.

## Evidence

- **`TypeApparencyTests`** (new): 8 tests, each apparent shape paired with a close negative (different type, bare element type vs `T[]`, plain value read, `default(T)` declined).
- **`TargetTypedNewPassTests`** unchanged: all 11 pass → refactor byte-identical.
- Fast decompiler suite green: 3717 total, 0 failures after a full `dotnet-inspect.slnx` build (the single `ExpressionTreeLambdaTests.LookalikeExpressionAssembly_StaysFactoryCalls` miss is the known fixture-binary environmental case; it passes once the slnx fixture is built — verified).

## Adversarial review

Requested: new analysis predicate + a printer refactor on a byte-stable path → two non-Claude reviewers per `AGENTS.md`.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/TypeApparencyTests.cs
+++ b/src/ILInspector.Decompiler.Tests/TypeApparencyTests.cs
@@ -1,0 +1,67 @@
+using ILInspector.Decompiler.Pipeline;
+
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// Direct coverage of the shared per-site "type is apparent" predicate
+/// (<see cref="CSharpPrinter.TypeIsApparent"/>) that the target-typed-<c>new</c>
+/// shortener consumes and a future opt-in <c>var</c> lens will consume. Each apparent
+/// shape is paired with a close negative so the predicate stays conservative: it must
+/// report apparent only when the right-hand side spells the declared type exactly.
+/// </summary>
+public class TypeApparencyTests
+{
+    static readonly TypeRef VoidType = TypeRef.CoreLib("System", "Void");
+    static readonly TypeRef IntType = TypeRef.CoreLib("System", "Int32");
+    static readonly TypeRef ListOfInt = TypeRef.Definition("synthetic", "N", "MyList");
+    static readonly TypeRef OtherType = TypeRef.Definition("synthetic", "N", "Other");
+
+    static NewObject Creation(TypeRef type)
+        => new(new MethodRef(type, ".ctor", VoidType, [], HasThis: true), []);
+
+    [Fact]
+    public void ObjectCreation_OfExactType_IsApparent()
+        => Assert.True(CSharpPrinter.TypeIsApparent(ListOfInt, Creation(ListOfInt)));
+
+    [Fact]
+    public void ObjectCreation_OfDifferentType_IsNotApparent()
+        => Assert.False(CSharpPrinter.TypeIsApparent(OtherType, Creation(ListOfInt)));
+
+    [Fact]
+    public void ArrayCreation_OfMatchingElementType_IsApparent()
+        => Assert.True(CSharpPrinter.TypeIsApparent(
+            TypeRef.SzArray(IntType),
+            new NewArray(IntType, new Constant(4, IntType))));
+
+    [Fact]
+    public void ArrayCreation_ComparedAgainstBareElementType_IsNotApparent()
+        // `new int[n]` names `int[]`, not `int`: a declaration typed `int` is not
+        // apparent from an array creation.
+        => Assert.False(CSharpPrinter.TypeIsApparent(
+            IntType,
+            new NewArray(IntType, new Constant(4, IntType))));
+
+    [Fact]
+    public void ExplicitCast_ToExactType_IsApparent()
+        => Assert.True(CSharpPrinter.TypeIsApparent(
+            ListOfInt,
+            new CastClass(ListOfInt, new LoadArgument(0, "x", OtherType))));
+
+    [Fact]
+    public void ExplicitCast_ToDifferentType_IsNotApparent()
+        => Assert.False(CSharpPrinter.TypeIsApparent(
+            OtherType,
+            new CastClass(ListOfInt, new LoadArgument(0, "x", OtherType))));
+
+    [Fact]
+    public void PlainValueRead_IsNotApparent()
+        // A parameter/local read (or any non-type-naming expression) leaves the type
+        // implicit — conservatively not apparent.
+        => Assert.False(CSharpPrinter.TypeIsApparent(ListOfInt, new LoadArgument(0, "x", ListOfInt)));
+
+    [Fact]
+    public void DefaultValue_IsNotApparent_ConservativeByDesign()
+        // `default(T)` names the type, but a target-typed `default` renders bare; the
+        // v1 predicate declines it rather than risk a `var x = default;` that loses T.
+        => Assert.False(CSharpPrinter.TypeIsApparent(IntType, new DefaultValue(IntType)));
+}
--- a/src/ILInspector.Decompiler.Tests/TypeApparencyTests.cs
+++ b/src/ILInspector.Decompiler.Tests/TypeApparencyTests.cs
@@ -64,4 +64,11 @@ public class TypeApparencyTests
         // `default(T)` names the type, but a target-typed `default` renders bare; the
         // v1 predicate declines it rather than risk a `var x = default;` that loses T.
         => Assert.False(CSharpPrinter.TypeIsApparent(IntType, new DefaultValue(IntType)));
+
+    [Fact]
+    public void ValueTypeCast_ViaUnboxAny_IsNotApparent_DocumentedV1Boundary()
+        // A value-type cast (`(int)obj`) is genuinely apparent in C# but is modeled by
+        // UnboxAny, not CastClass. v1 declines it — a documented extension point for the
+        // `var` slice, not a defect (declining is output-safe). This locks the boundary.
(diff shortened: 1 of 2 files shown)


richlander:

Two fixed-head reviews in isolated read-only worktrees, each with real-run evidence.

**GPT-5.6** — clean on both the original head (`b1561b7b`) and the fixed head (`62cb53f7`).
- Verified the target-typed-`new` refactor is byte-identical: the `NewObject` guard precedes the predicate, whose branch performs the identical structural `TypeRef.Equals`; `SzArray` comparison is structural (`TypeRef.cs:419-451`); `DeclarationTypeText` behavior unchanged.
- On the fixed head it hashed the `TypeIsApparent` method body and confirmed it is **byte-for-byte unchanged** (both revisions hash to `2eb7e56c…`); product delta is comments only; `UnboxAny` correctly reaches `_ => false`.
- Build 0W/0E; tests 20/20.

**Gemini-3.1** — one blocking-labeled design-intent finding + incompleteness notes, all resolved on the fixed head.
- Confirmed byte-identity and that multi-dim arrays stay guarded (`MultiDimArrayCreationText`).
- **Finding #3 (resolved):** flagged that if a *future* `var` lens and the `new()` shortener both fired on one apparent site the result would be `var x = new()` (CS8754). No such behavior exists in this PR; I reworded the `DeclarationTypeText` comment to state the two are **mutually-exclusive** spellings of the same apparent site — the shared predicate is only the apparency input, and the future `var` lens owns the either/or. Gemini confirmed this resolves the concern.
- **Finding #2 (resolved):** noted genuinely-apparent shapes v1 declines because other IR nodes model them (`UnboxAny` value casts, `ObjectInitializerExpression`, `ArrayLiteral`/`SpanLiteral`). These are now enumerated in the `TypeIsApparent` XML doc as output-safe conservative-decline extension points for the `var` slice, and locked by a new `UnboxAny` decline test.
- Gemini's re-review: "satisfied … concerns fully addressed without introducing new defects"; build 0W/0E; tests 20/20.

Both reviewers independently confirmed the core claim: the predicate's `NewObject` branch is exactly the clause `TargetTypedNewText` previously inlined, and that method only ever reaches it via its `value is NewObject creation` guard — so config resolution and rendered output are unchanged. The fixed-head change is comment/doc/test-only (predicate switch byte-identical). No open findings.

## Status
- Byte-identity: proven (target-typed-`new` tests unchanged/passing; both reviewers reproduced).
- `TypeApparencyTests` 20/20; fast decompiler suite green (3717, 0 failures after full slnx build).
- SRM-only, NativeAOT-friendly, Roslyn-free (pure IR switch).

richlander:
Ready to merge

- CI green: build-net10, test, csharp-diff-smoke, changes all pass; il-diff-smoke / markdownlint / pack path-skipped (no markdown or packable change).
- Adversarial review complete and reconciled: GPT-5.6 and Gemini-3.1 both clean on the fixed head (62cb53f7), with real-run evidence; Gemini's design-intent and incompleteness findings resolved doc/test-only.
- Predicate switch byte-for-byte unchanged from the first reviewed head; target-typed-`new` output byte-identical.

PR #3200, opened 2026-07-25
Render interface-declared this-receiver access with the erased ((I)this) cast (#3128)

### LINKED ISSUE #3128: Decompiler: DIM access via interface-cast this renders invalid C# (missing ((I)this) cast) in CallText and MethodGroupText
## Summary

The decompiler renders **default-interface-member (DIM) access through an interface-cast `this`** without the required cast, producing invalid C#. This affects **both** instance-call rendering (`CallText`) and method-group rendering (`MethodGroupText`) in `src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.Members.cs`.

A class→implemented-interface conversion emits **no IL** (it is an implicit reference conversion), so `((IDim)this).Value` leaves the IR target as a bare `LoadArgument{Index:0,Name:"this"}` with the callee/method declared on the interface. The printer's this-receiver arms then spell the member bare (or `this.`-qualified), but a DIM is **not** a member of the implementing class, so the rendered spelling fails to compile.

## Repro

```csharp
public interface IDim
{
    int Value() => 7;   // default interface member
}

public sealed class DimConsumer : IDim
{
    public System.Func<int> CastInterfaceGroup() => ((IDim)this).Value;   // method group
    public int CastInterfaceCall() => ((IDim)this).Value();               // call
}
```

Original IL for the group:

```il
ldarg.0
dup
ldvirtftn IDim::Value()
newobj System.Func<int>::.ctor(object, nint)
```

## Current (invalid) renderings

| Site | Default output | `dotnet_style_qualification_for_method = true` |
| --- | --- | --- |
| Method group | `new Func<int>(Value)` | `new Func<int>(this.Value)` |
| Call | `Value()` | `this.Value()` |

All four fail to compile with **CS1061** (`DimConsumer` has no `Value` member). The default (no-knob) spellings are already invalid — this is a **pre-existing** gap, independent of the `this.`-qualification knobs.

## Faithful spelling

```csharp
((IDim)this).Value      // method group
((IDim)this).Value()    // call
```

which recompiles to the original `ldvirtftn IDim::Value()` / `callvirt IDim::Value()`.

## Why it is not trivial

- The fix must live in **both** `CallText` and `MethodGroupText` this-receiver arms.
- It requires reliably detecting that the callee's declaring type is an **interface** to re-insert `((I)this)`. `MetadataSource.IsInterface` only resolves the **target assembly's own** type definitions, so a DIM declared in a **referenced** assembly would not be recognized (see the existing cross-assembly `TypeShape.Unknown` limitation). A sound fix must **decline (leave as-is) when interface-ness is unknown** rather than guess.
- A base-class inherited virtual method group over `this` (`this.VirtualMethod`) is a genuine member and must stay `this.`/bare — only interface-declared members need the cast. The discriminator is declaring-type interface-ness, not `ldvirtftn`.

## Scope / non-goal

Discovered during adversarial review of #3123 (this.-qualification knobs, Slice 2b of #3097). That PR does **not** regress this shape: its default output is byte-identical to before, and the opt-in knob merely turns one already-invalid spelling into another. Deferred here as a pre-existing, product-wide fidelity fix.

richlander:
- Fixes #3128
- Renders a call or method-group to an interface-declared member reached through `this` with the erased `((I)this)` cast, so DIM and explicit-interface-implementation access compiles and stays opcode-faithful.
- Card revision: `f2568a98`

## Change

**Conclusion:** **PASS** — the printer now re-inserts the reference-conversion cast that IL erases, turning previously invalid (CS1061) or instantiation-infidelic renders into valid, opcode-faithful C#, with no effect on non-interface this-receiver calls.

A default-interface-member (DIM) or explicit-interface-implementation member reached through `this` is declared on an interface the class was implicitly upcast to. That class→interface upcast (and the variance upcast `I<T>`→`I<object>`) is an implicit reference conversion that emits **no IL**, so the IR target is a bare `LoadArgument{0,"this"}` with the callee declared on the interface. The old this-receiver arms in `CallText`/`MethodGroupText` spelled the member bare, but the member does not bind as a member of the implementing class.

The fix carries confirmed interface declaring types on `IrFunction.InterfaceTypes` (populated at import from `MetadataSource.ClassifyResolvedType`; **absent — never guessed — when interface-ness is unknown**, e.g. unresolved cross-assembly). A new printer discriminator `IsInterfaceCastThisReceiver` emits `((I)this)` when the callee's declaring type is a confirmed interface **and** not the enclosing type at its own instantiation. The arm is placed **first**, so every non-interface case falls through to the exact prior behavior.

### Benchmark target

Benchmark target: `ILInspector.Decompiler.Tests.dll` fixture specimens (`DimConsumer`, `ThisQualificationExplicitFace`, `IThisQualificationGeneric<T>`), which reproduce DIM, explicit-interface, and variance this-receiver access.

dotnet-inspect command:

```bash
dotnet-inspect member ILInspector.Decompiler.Tests.DimConsumer.DimCall --library ILInspector.Decompiler.Tests.dll -S "Decompiled Source" --all
```

### Original source

```csharp
public interface IDimFace { int Value() => 7; }

public sealed class DimConsumer : IDimFace
{
    public int DimCall() => ((IDimFace)this).Value();
    public System.Func<int> DimGroup() => ((IDimFace)this).Value;
}
```

IL anchor (`-S "IL"`) for `DimConsumer.DimCall`:

```il
IL_0000: ldarg.0
IL_0001: callvirt     ILInspector.Decompiler.Tests.IDimFace::Value()
IL_0006: ret
```

### Before

```csharp
public int DimCall() => Value();
public System.Func<int> DimGroup() => new Func<int>(Value);
public int CallExplicitInterface() => FaceMethod();
public virtual int CallViaObjectInstantiation() => M();
```

- Valid: **False** for the first three (`Value`/`FaceMethod` do not bind on the class → CS1061); `CallViaObjectInstantiation` compiles.
- Correct: **False** for `CallViaObjectInstantiation` — bare `M()` rebinds to the enclosing `I<T>::M` instead of `I<object>::M`.
- IL fidelity: **False** — the bare renders drop the interface receiver; `CallViaObjectInstantiation` binds a different instantiation.
- Taste applied: None.
- Commit: `d39bdf70`

### After

```csharp
public int DimCall() => ((IDimFace)this).Value();
public System.Func<int> DimGroup() => new Func<int>(((IDimFace)this).Value);
public int CallExplicitInterface() => ((IThisQualificationFace)this).FaceMethod();
public virtual int CallViaObjectInstantiation() => ((IThisQualificationGeneric<object>)this).M();
```

- Valid: **True** — all four compile and bind.
- Correct: **True** — the cast targets exactly the interface/instantiation the IL invokes.
- IL fidelity: **True** — the cast is an implicit reference conversion emitting no IL, so each render recompiles to the shown opcodes (`ldarg.0; callvirt IDimFace::Value()` etc.). Verified via `-S "IL"`.
- Taste applied: None.
- Commit: `f2568a98`

### Fully raised

## Evidence

| Check | Baseline (`d39bdf70`) | Head (`f2568a98`) |
| --- | --- | --- |
| Focused tests | `ThisQualificationTests`+`ThisQualificationDecisionTests`: 54 passed | 54 passed |
| Decompiler PR quick gate | PASS, 0 pass bugs | PASS, 0 pass bugs (residue 251→247, no regressions) |
| Real witness (`DimConsumer::DimCall`) | `Value()` invalid (CS1061) | `((IDimFace)this).Value()` valid + IL-faithful |
| Real witness (`ThisQualificationExplicitFace::CallExplicitInterface`) | `FaceMethod()` invalid (CS1061) | `((IThisQualificationFace)this).FaceMethod()` valid |
| Real witness (`IThisQualificationGeneric<T>::CallViaObjectInstantiation`) | `M()` rebinds to `I<T>::M` (infidelic) | `((IThisQualificationGeneric<object>)this).M()` faithful |

Baseline focused-suite and PR quick gate were run on the base commit `d39bdf70`. The Windows-local pre-existing failures in the broader fast suite (CRLF string-switch raw-literal, `MinimalParameterAttributes` ref/out compile-back CS0269, flaky SourceLink `discover` grouping) reproduce identically at base and are untouched by this PR.

## Decompiler quality

**Conclusion:** **PASS** — PR quick gate matched baseline tolerances with 0 pass bugs; the change is validity-only (adds a cast in a previously-declined interface arm), so it cannot regress non-interface renders.

### PR quick gate

Generated card (head vs `tools/DecompilerHarness/corpus/pr-quick-baseline.json`, hash-stable 100 methods/assembly, 15 assemblies / 1,500 methods):

### Decompiler quality diff

Corpus: #1166 real-world decompiler corpus sensor: #1150 pinned NuGet assemblies plus dotnet-inspect managed assemblies. 15 assemblies, 1,500 methods

| Metric | Change |
| ------ | ------ |
| Detected lowering residue (population differs) | 251 (16.73%) → 247 (16.47%) |
| Conditional-branch residue (population differs) | 46 (3.07%) → 43 (2.87%) |
| Forward-merge stops (population differs) | 32 (2.13%) → 30 (2.00%) |
| Pass bugs (-) | 0 → 0 (0) |

Current measured debt: 247 methods with detected lowering residue.
Regression verdict: PASS — corpus sensor matched baseline tolerances.

> **Conclusion:** **PASS** — no pass bugs, residue at or below baseline.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.ThisQualificationTests" -class "ILInspector.Decompiler.Tests.ThisQualificationDecisionTests"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ThisQualificationDecisionTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ThisQualificationDecisionTests.cs
@@ -217,8 +217,9 @@ public sealed class ThisQualificationDecisionTests
     // An explicit interface implementation invoked through this reaches the call
     // site with the interface as its declaring type (cross-type from the
     // implementing class). Bare `FaceMethod()`/`this.FaceMethod()` does not bind —
-    // the member requires a cast — so it is never a `this.` opt-in. The cross-type
-    // guard records no taste decision.
+    // the member requires a cast — so it is never a `this.` opt-in. The #3128
+    // interface-cast arm renders `((I)this).FaceMethod()` before the knob arm, so
+    // no taste decision is recorded.
     [Fact]
     public void ExplicitInterfaceCall_WithKnobEnabled_RecordsNoDecision()
     {
@@ -234,7 +235,8 @@ public sealed class ThisQualificationDecisionTests
     // only the DEFINITION with the enclosing type. Bare/this. M() binds to I<T>::M,
     // not I<object>::M, so the qualifier is not byte-preserving. Definition-only
     // equality would wrongly record it; the exact-instantiation guard records
-    // nothing.
+    // nothing. The #3128 interface-cast arm renders it before the knob arm, so no
+    // taste decision is recorded either way.
     [Fact]
     public void ConstructedGenericSelfCall_WithKnobEnabled_RecordsNoDecision()
     {
@@ -244,6 +246,20 @@ public sealed class ThisQualificationDecisionTests
             new PrinterOptions { QualifyMethodAccess = true });
 
         Assert.DoesNotContain(result.Decisions, d => d.RuleId == "qualify-method-access");
+    }
+
+    // The #3128 emit: the call to a DIM at a DIFFERENT instantiation of the
+    // enclosing interface must keep the ((I<object>)this) cast — bare M() would
+    // rebind to I<T>::M. The cast is a validity/fidelity fix, present with the
+    // qualify-method knob off.
+    [Fact]
+    public void ConstructedGenericSelfCall_RendersInstantiationCast()
+    {
+        var result = Decompile(
+            typeof(IThisQualificationGeneric<>).FullName!,
+            nameof(IThisQualificationGeneric<object>.CallViaObjectInstantiation));
+
+        Assert.Contains("((IThisQualificationGeneric<object>)this).M()", result.Output);
     }
 
     // Two overloads differing only by a function pointer's return type
--- a/src/ILInspector.Decompiler.Tests/ThisQualificationTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ThisQualificationTests.cs
@@ -166,6 +166,55 @@ public sealed class ThisQualificationTests
         Assert.Contains("this.ReadField", text);
     }
 
+    // A default-interface-member call reached through `this` must render the
+    // erased `((I)this)` cast (#3128): the DIM is not a member of the implementing
+    // class, so bare `Value()` / `this.Value()` is CS1061. The default (knob-off)
+    // render was already invalid before this fix; it is now the faithful spelling
+    // that recompiles to `ldarg.0; callvirt IDimFace::Value()`.
+    [Fact]
+    public void DefaultInterfaceMemberCall_RendersInterfaceCast_ByDefault()
+    {
+        var text = RenderMember(typeof(DimConsumer), nameof(DimConsumer.DimCall));
+        Assert.Contains("((IDimFace)this).Value()", text);
+        Assert.DoesNotContain("this.Value()", text);
+    }
+
+    // The `((I)this)` cast is a validity fix, not a `this.`-qualification opt-in:
+    // it is present with the qualify-method knob off and unchanged with it on (the
+    // knob's `this.` arm never runs for an interface-declared callee).
+    [Fact]
+    public void DefaultInterfaceMemberCall_RendersInterfaceCast_UnderMethodQualification()
+    {
+        var text = RenderMember(typeof(DimConsumer), nameof(DimConsumer.DimCall),
+            new PrinterOptions { QualifyMethodAccess = true });
+        Assert.Contains("((IDimFace)this).Value()", text);
+        Assert.DoesNotContain("this.Value()", text);
+    }
+
+    // A DIM method group over `this` must likewise cast: `((IDimFace)this).Value`
+    // recompiles to `ldarg.0; dup; ldvirtftn IDimFace::Value; newobj Func`, where
+    // bare `Value` (CS1061) or `this.Value` would not bind.
+    [Fact]
+    public void DefaultInterfaceMemberGroup_RendersInterfaceCast_ByDefault()
+    {
+        var text = RenderMember(typeof(DimConsumer), nameof(DimConsumer.DimGroup));
+        Assert.Contains("((IDimFace)this).Value", text);
+        Assert.DoesNotContain("(Value)", text);
+    }
+
+    // An explicit interface implementation invoked through `this` reaches the call
+    // site with the interface as declaring type; the member does not bind on the
+    // implementing class, so it must render `((I)this).FaceMethod()` — the faithful
+    // spelling of the fixture's own source (#3128).
+    [Fact]
+    public void ExplicitInterfaceCall_RendersInterfaceCast()
+    {
+        var text = RenderMember(typeof(ThisQualificationExplicitFace),
+            nameof(ThisQualificationExplicitFace.CallExplicitInterface));
+        Assert.Contains("((IThisQualificationFace)this).FaceMethod()", text);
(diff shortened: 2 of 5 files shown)


richlander:

Reviewed at exact head `f2568a983ecafc63fff8f7f5840d4bbe4aa63192`, two different-model reviewers in isolated linked worktrees (scratch under `/tmp`-equivalent, no commits/reset/add in review trees).

### Reviewer A — GPT-5.6

**Verdict: APPROVE**, no blocking issues. Verified with real runs: Release build; 54 targeted tests; adversarial normal-interface fixture rendered bare `M()` with callee `Normal::M()` (no false cast); plain/open-generic DIM self-calls stayed bare while a differing generic instantiation gained the cast; DIM, method-group, explicit-interface, and variance casts recompiled to identical IL (no added opcodes); genuine base method group stayed `base.M`; empty `InterfaceTypes` path rendered normally.

### Reviewer B — Gemini 3.1 Pro

Raised two blocking findings initially; both resolved after reconciliation at the same head. **Final verdict: APPROVE.**

- **Finding #2 — "interfaces map to `TypeShape.Unknown` (missing switch arm) → operator== recursion": RETRACTED (false).** This was a misread of the 3-line diff context. The actual `kind switch` in `IrImporter.cs` (line ~651) is `TypeShapeKind.Class or TypeShapeKind.Interface or TypeShapeKind.Delegate => TypeShape.Reference` — interfaces map to `Reference`, not `Unknown`. That arm is **identical in the base commit** `d39bdf70` (this PR only extracted `kind` into a local), so `IsKnownReferenceType` returns true for interfaces and the described recursion cannot occur via this path. The reviewer confirmed the retraction after reading the full file and acknowledged it had not actually run the hypothesized fixture.

- **Finding #1 — "struct/boxed `this` renders `(this).M()` (CS1061)": VALID but a SEPARATE, PRE-EXISTING, OUT-OF-SCOPE gap; not actioned here.** Reproduced at head with a struct explicit-interface / DIM fixture. Two facts settle its disposition: (1) it is **not a regression** — base commit `d39bdf70` renders the identical `(this).M()` / `(this).Value()` for the same fixtures (this PR's interface arm is guarded by `receiver is LoadArgument{Index:0,Name:"this"}`, which a boxed struct receiver never matches, so it falls through to prior behavior); (2) it is **out of scope for #3128**, which is explicitly the class→interface **no-IL implicit reference conversion**, whereas a struct's `((IFace)this)` is a **boxing conversion that emits `box` IL** (`ldobj; box; callvirt`) — a mechanistically distinct category and a distinct receiver IR shape. Filed as follow-up **#3201**. The reviewer agreed this is a legitimate separate follow-up that does not block this PR.

### Outcome

Both fixed-head reviews are clean (APPROVE / APPROVE). No code changes were required by the reviews; the one valid finding is pre-existing, out of this issue's scope, and tracked in #3201.

richlander:
Ready to merge

PR #3202, opened 2026-07-25
Model the var-spelling taste family as one value-domain axis (#3169)

richlander:
## What

Adds the `var`-spelling taste family to the style-option catalog as a single **`var-spelling-style`** value-domain descriptor: an `explicit` default plus three site-category values that map 1:1 onto the three `csharp_style_var_*` editorconfig keys.

This is **byte-neutral plumbing only** (stacked-PR base). It adds the options, the catalog descriptor, and the automatic config wiring — but **no printer emission**. The printer still ignores these options, so default output is unchanged and IL-identical. **PR B will stack the `DeclarationTypeText` emission on top of this branch.**

## Changes

- **`PrinterOptions`**: three new `init` bools — `PreferVarForBuiltInTypes`, `PreferVarWhenTypeApparent`, `PreferVarElsewhere`. All default `false` (byte-stable).
- **`StyleOptionCatalog`**: new `VarSpellingStyle()` descriptor (`Spelling` tier, `ByteDivergent = false`), registered in `Options`. Four tokens: `explicit` (default, clears all three bools) + `var-for-built-in-types` / `var-when-type-apparent` / `var-elsewhere`, each carrying its own `ConfigKey` and independent backing bool.
- **Config wiring is automatic** via `RenderStyleConfig.KnobsByKey` (derives recognized keys from each value's `ConfigKey`) — no resolver change needed.
- **Docs**: documented the family in `docs/decompiler-taste.md`.

## Design notes

- The three keys back **independent** bools (a user may enable any subset), but declaration sites partition into exactly one bucket (built-in → else apparent → else elsewhere), so at most one governs a site. `GetValue` reports the coarse first-enabled category; `WithValue` single-selects. Mechanically identical to the existing guarded-boolean-return family.
- **Opt-in only.** dotnet/runtime's `.editorconfig` sets all three `csharp_style_var_*` keys to `false` (prefer explicit types), so every value is `OracleEndorsed = false` — `var` is **never** part of the "full taste" aggregate. The shipped default spells explicit types everywhere, matching the runtime.
- The drift guard (`EveryBackingBooleanProperty_IsReachableThroughSomeCatalogValue`) requires the bools + catalog values to land together, keeping this base green.

## Tests

- **Catalog** (4): byte-neutral four-value axis; three categories map to the editorconfig keys; categories are independent (each key sets only its own bool); endorsed by neither facet (stays opt-in).
- **Config parse** (4): each key → its bool; `:severity` suffix tolerated; `= false` clears only its own bucket; `dotnet_inspect_style_full_taste = true` leaves all three var bools `false`.
- Updated a pre-existing unknown-key test that had used `csharp_style_var_when_type_is_apparent` as its *example* of an unknown key — now a recognized key (good confirmation the wiring works).
- Full `dotnet-inspect.slnx` build clean; catalog suite 28/28, config suite 51/51.

Part of #3169. PR B (printer emission) stacks on this branch.

### DIFF
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -641,6 +641,23 @@ three-token axis rather than two coupled booleans. The catalog is exhaustive: a
 test-only drift guard asserts every backing `PrinterOptions` boolean is reachable
 through some descriptor value, so a new knob cannot land without a catalog entry.
 
+The `var`-spelling family is a second multi-value axis, modeled as one
+`var-spelling-style` descriptor with four tokens — `explicit` (the default) plus
+the three site categories `var-for-built-in-types`, `var-when-type-apparent`, and
+`var-elsewhere`, one per editorconfig key (`csharp_style_var_for_built_in_types`,
+`csharp_style_var_when_type_is_apparent`, `csharp_style_var_elsewhere`). Because a
+declaration site falls into exactly one category, the three category values back
+*independent* booleans (a host may enable any subset), while the axis stays
+single-select for display: `GetValue` reports the coarse first-enabled category
+and `WithValue` selects one exclusively. `var` is byte-neutral — it erases the
+declared type spelling but never changes the IL — so the family sits in the
+`Spelling` tier with `ByteDivergent = false`. Every category value is
+`OracleEndorsed = false` and `CorpusEndorsed = false`: dotnet/runtime's
+`.editorconfig` sets all three `csharp_style_var_*` keys to `false` (prefer
+explicit types), so no `var` value is endorsed by either facet. The family is
+therefore **opt-in only** and never part of the "full taste" aggregate — the
+shipped default spells explicit types everywhere, matching the runtime.
+
 ## Verification and soundness
 
 How these rendering choices are held correct — the verification checks and floors, the
--- a/src/ILInspector.Decompiler.Tests/StyleOptionCatalogTests.cs
+++ b/src/ILInspector.Decompiler.Tests/StyleOptionCatalogTests.cs
@@ -356,4 +356,81 @@ public class StyleOptionCatalogTests
         Assert.False(arrow.OracleEndorsed);
         Assert.False(arrow.CorpusEndorsed);
     }
+
+    // ---- var-spelling family (#3169) ----
+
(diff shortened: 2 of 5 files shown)


richlander:

Two non-Claude reviewers examined the exact head `da404bf3` in isolated detached worktrees (read-only, scratch under `/tmp/`).

- **GPT-5.6 (`gpt-5.6-sol`):** *No significant issues found.*
- **Gemini-3.1 Pro (`gemini-3.1-pro-preview`):** *No blocking concerns.* Verified all seven focus areas with real runs:
  - **Byte-stability** — codebase audit confirms zero consumers of the three new `PrinterOptions` bools outside the catalog + tests; default output is IL-identical.
  - **Catalog invariants** — `WithValue(explicit)` clears the axis (explicit's `SetSelected(true)` clears all three; the three categories' `SetSelected(false)` are safe redundant clears); `GetValue` yields the first active category in `Values` order; drift guard passes.
  - **Config wiring** — `KnobsByKey` scoops up all three per-value `ConfigKey`s; `= false` clears only its own bucket; `ApplyFullTaste` skips the axis because its `EndorsedValue` is `null` (`OracleEndorsed == false`).
  - **Descriptor `ConfigKey`** — computes `null` for the 4-value axis; the only descriptor-level `ConfigKey` consumer is the 2-value `this.`-qualification path, unaffected.
  - **NativeAOT/Roslyn/SRM** — delegates + constants only; no reflection, Roslyn, or non-SRM introduced.
  - **Tests** — built Release; `StyleOptionCatalogTests` 28/28 and `RenderStyleConfigTests` 51/51 pass; the updated unknown-key test keeps a meaningful assertion.
  - **Docs** — `docs/decompiler-taste.md` matches the single-select-display / independent-backing-bool model.

No findings required action from either reviewer — no code changes since the reviewed head. Both fixed-head reviews are clean.

richlander:
Ready to merge

All merge-blocking CI is green (build-net10, test, csharp-diff-smoke, markdownlint, changes). Both fixed-head adversarial reviews (GPT-5.6 + Gemini-3.1) are clean with no required actions. No follow-up work is blocking; PR B (the printer emission that actually spells `var`) will stack on this branch as a separate PR.

PR #3203, opened 2026-07-25
Wrap long member signatures + general one-liner opt-out (#3185)

### LINKED ISSUE #3185: Long member signatures vs. one-liner (expression-body) formatting: line-length-aware wrapping policy

richlander:
## What & why

Closes #3185. Two coupled changes to how the decompiler renders wide members:

1. **Always-on parameter-list wrapping.** When a member declaration's first
   physical line — head plus whatever trails the closing `)` on it (`=> expr;`,
   `{`, or `;`) — would exceed the **120-column** runtime budget, the parameter
   list breaks one parameter per line under a four-space continuation indent,
   with the closing `)` and any trailing `where`/base-`: this(...)` clause kept
   on the last parameter's line and the `=>`/`{`/`;` staying on that line. This
   is the **revealed dotnet/runtime corpus practice** (it wraps a wide
   signature's *parameters*, not its arrow) and joins the existing always-on
   fluent-chain wrapper. Whitespace-only + token-identical ⇒ **IL unchanged**.

   ```csharp
   // before (one long line):
   public static JsonElement Parse([StringSyntax("Json")] ReadOnlySpan<byte> utf8Json, JsonDocumentOptions options = default) => JsonDocument.ParseValue(utf8Json, options).RootElement;
   // after:
   public static JsonElement Parse(
       [StringSyntax("Json")] ReadOnlySpan<byte> utf8Json,
       JsonDocumentOptions options = default) => JsonDocument.ParseValue(utf8Json, options).RootElement;
   ```

2. **General one-liner opt-out.** A single new knob,
   `PrinterOptions.DisableOneLinerWrapping` (catalog id
   `disable-one-liner-wrapping`, **off by default**), suppresses **all**
   always-on width wrapping uniformly — both the fluent-chain and the new
   signature wrapper — for callers who want wide constructs to stay on one
   physical line. It deliberately does **not** touch the opt-in
   `WrapSplittableExpressions` boolean/bitwise wrappers.

## Design notes

- **Conservative parameter locator** (in `CSharpMemberLayout`, the layer that
  owns C# spelling): the first top-level `(` immediately preceded by the member
  name (an identifier char or generic-arg `>`), bounded before any trailing
  `where` constraint, with top-level comma splitting that respects
  `<>`/`()`/`[]`/`{}` nesting and string/char literals. Correctly skips tuple /
  parenthesized return types and `new()` in constraints.
- **Safe fallback**: shapes it cannot match unambiguously (symbolic
  `operator +(...)`, indexer `[...]` param lists) degrade to today's single line
  rather than risk a mangled declaration.
- **Classification**: keeping a wide construct inline **diverges** from the
  corpus (which wraps), so `disable-one-liner-wrapping` is endorsed by neither
  facet (`OracleEndorsed=false`, `CorpusEndorsed=false`), like the
  branchless-boolean lens. Catalog drift guard moves **9 → 10**.

## Evidence

- `CSharpMemberLayoutTests` (+8): over/under budget, opt-out stays inline,
  generic method args, tuple return type, `where T : new()`, symbolic-operator
  fallback, block body. Full suite **334/0**.
- `StyleOptionCatalogTests`: pins the neither-facet classification; drift guard
  auto-counts 10==10. `FluentChainFormattingTests`: opt-out suppresses chain
  wrapping.
- Fast decompiler suite **3708/0** (one unrelated pre-built-fixture error clears
  after a full `slnx` build). CLI/product suite **1520/0** (10 ildasm skips).
- `docs/decompiler-taste.md` documents both the always-on signature wrapper and
  the opt-out; markdownlint clean.

## Compatibility

Default output changes only for members whose signature line exceeds 120
columns (now wrapped), matching the corpus and the existing "Default == shipped"
invariant already used by the fluent-chain wrapper. No structured-schema,
ordering, or IL changes.

Adversarial review to follow (new default behavior + new heuristic).


richlander:

Two-model review per the reviewer roster: **Gemini** (`gemini-3.1-pro-preview`) and **GPT** (`gpt-5.6-sol`), each in an isolated read-only worktree at the exact head, scratch under `/tmp/`, no git mutation.

### Findings on the first head (`3381bcec`)

Both reviewers independently converged on the **same root-cause class**: the parameter-list scanner's literal handling (`SkipLiteral` in `CSharpMemberLayout.cs`) models only the conventional `\`-escaped string/char literal.

- **Gemini — BLOCKING (Critical).** A verbatim string default ending in a backslash (`@"C:\dir\"`) desyncs the scan: `SkipLiteral` treats `\"` as an escaped quote, misses the real closing quote, and a later normal string's comma gets read as a top-level separator — injecting a newline **inside a string literal** (uncompilable C#). Real repro provided.
- **GPT — BLOCKING (High).** The same defect via **interpolated** (`$"{",\"}"`) and **raw** (`"""…"""`) string defaults: the nested/again-quoted comma is exposed and split. Real repro + Roslyn `CS1073`/`CS1010` shown.
- **GPT — Medium: expression body excluded from the width decision.** The same-line-arrow branch counts only `head + " =>"`, not the full `=> expr;`, so a body-dominated long line stays inline.

### Resolution

**Blocking literal-corruption bug (both reviewers) — fixed in `c5520e66`.** Added a `ContainsUnsupportedLiteral(head)` guard; `WrapParameterList` now declines to wrap (returns the inline single line — the existing safe-fallback contract) whenever the head contains any verbatim (`@"`), interpolated (`$"`, incl. `$@"`/`@$"`), or raw (`"""`) string literal. Conventional literals — including those with embedded commas (`"a,b,c"`) or literal `@`/`$` characters in their text — are still parsed correctly and still wrap. These non-conventional forms are not known to appear in decompiled signatures (constant defaults and attribute args render as conventional literals), so this is robustness hardening that closes the demonstrated corruption path. Four regression tests pin the interpolated/verbatim/raw defaults to the inline fallback and a conventional comma-bearing literal to the wrapped form.

**GPT Medium (Finding 2) — dismissed as intentional scope; GPT agreed and withdrew it.** The feature wraps the *parameter list* to fix a line that overflows because of its *signature width*. When a line overflows because of a long *body* (short head, long expression), wrapping parameters cannot reduce the max line width — the last parameter line still carries the full `=> <long expr>;`. Long expression-body/arrow wrapping is separate, explicitly deferred scope. GPT confirmed: *"Finding 2 is withdrawn. Under the clarified signature-driven scope, counting only `head + " =>"` is intentional."*

### Fixed-head re-review (`c5520e66`) — both clean

- **Gemini:** *"The fixed head is clean. There are no remaining blocking or non-blocking issues."* Verified the original repro now stays inline/token-identical, all target literal forms detected (incl. `$@"`/`@$"` and verbatim-ending-in-backslash), no false positives on `@`/`$` in literal text, empty strings, char literals `'"'`/`'\\'`, and no new bugs. 338 CSharp tests pass.
- **GPT:** *"No significant issues found at `c5520e66`. BLOCKING: None. Non-blocking: None."* Verified interpolated/verbatim/raw inline + token-identical, `"a$b@c,d"` and `""` still wrap correctly, 338/0.

### Validation

- `ILInspector.CSharp.Tests`: **338/0** (incl. 8 new signature-wrap + 4 new literal-safety cases).
- `StyleOptionCatalogTests`, `FluentChainFormattingTests`: green; catalog drift guard 10==10.
- Fast decompiler suite: **3708/0** (one pre-existing `ExpressionTreeSpoof` fixture-artifact error clears after a full `slnx` build).
- Product suite: **1520/0** (`PlatformResolverTests.GetPacksDirectory` is a parallel-run shared-packs-cache flake — passes 39/0 in isolation, unrelated to this change).

richlander:
Ready to merge

All merge-blocking validation is complete:

- **CI green** on the merge head (`b5c8ad8f`): `build-net10`, `test (ubuntu-26.04, linux-x64)` (8m25s), `csharp-diff-smoke`, `markdownlint`, `changes` all pass. `il-diff-smoke`/`pack` are path-gated skips.
- **origin/main merged** (not rebased, per under-review policy). The merge reconciled main's catalog refactor (boolean descriptor → multi-value axis model): the `disable-one-liner-wrapping` knob is now expressed via the new `Boolean(...)` factory; the neither-facet test and taste-doc clause were preserved alongside main's new var-spelling additions.
- **Local post-merge validation**: catalog 29/0, fluent-chain 5/0, CSharp layout 338/0, fast decompiler 3736/0, markdownlint clean.
- **No new adversarial review needed for the merge**: the wrapping logic is unchanged since the clean 2-model review at `c5520e66` (both Gemini and GPT clean, both original blocking findings fixed and re-verified). The merge is a mechanical catalog-shape adaptation; the drift-guard test is reflection-based and automatically covers the new `DisableOneLinerWrapping` bool.

Issue #3129, opened 2026-07-24
Decompiler: generated support-state names still leak into product bodies (EVIL CS1001/CS0246)


richlander:
Found via the EVIL authored-corpus benchmark (#3079) on current main run `16c0687f` (26/26 honest pool). This is a genuine **product TargetBody** defect, not just an RTS shell artifact: although the row `detail` is prefixed `closure-stalled:`, the target method's own decompiled body contains unspellable compiler-generated names.

## Current-run signal

`invalidKind == ProductBodyDefect` rows in the saved run:

- `CS1001` Identifier expected: **109** rows
- related generated-support syntax/name rows: `CS0246` **11**, `CS1002` **4**, plus a small tail (`CS1003`, `CS1525`)

This is a residual/follow-up to closed #2945. The old dominant family shrank, but generated closure/support-state names still reach product output.

### Closure/local-function state still prints raw angle-bracket names

```bash
dnx dotnet-inspect -y -- member "System.CommandLine.Help.HelpBuilder" "WriteColumns:1" --package System.CommandLine@3.0.0-preview.5.26302.115 --all -S "Decompiled Source"
```

Target body contains invalid identifiers/types:

```csharp
<>c__DisplayClass17_0 V_3;
...
firstColumnWidth = Enumerable.Max(... new Func<TwoColumnHelpRow, IEnumerable<int>>(V_3.<WriteColumns>b__3));
...
static IEnumerable<ValueTuple<string, string>> ZipWithEmpty(...) =>
    new <<WriteColumns>g__ZipWithEmpty|17_2>d(-2) { <>3__first = first, <>3__second = second };
```

### Cached compiler singleton lambda still prints raw `<>c.<>9.<...>`

```bash
dnx dotnet-inspect -y -- member "StackExchange.Redis.TaskExtensions" "RedisFireAndForget:1" --package StackExchange.Redis@3.0.17 --all -S "Decompiled Source"
```

```csharp
task.ContinueWith(new Action<Task>(<>c.<>9.<RedisFireAndForget>b__8_0), ...);
```

### Inline-array/private implementation support types are not spellable

```bash
dnx dotnet-inspect -y -- member "Microsoft.CodeAnalysis.Emit.EncLocalInfo" "GetDebuggerDisplay:1" --package Microsoft.CodeAnalysis.Common@5.0.0 --all -S "Decompiled Source"
```

```csharp
<>y__InlineArray5<object> V_0 = default;
...
[...]

PR #3204, opened 2026-07-25
Route residual unspellable generated names through the sanitizer (#3129)

richlander:
- Fixes/advances #3129 (the "never print raw unspellable generated identifiers" floor)
- Changes product decompilation to route residual compiler-generated names at method-group / call and object/`with`-initializer-member positions through the same sanitizer the type-name path already uses, so no raw `<…>` identifier ever reaches the rendered body.
- Card revision: `b2ae90a8`

## Change

**Conclusion:** **PASS** — this eliminates the raw-unspellable-identifier leak family (the issue's dominant `CS1001` parse-error signal and its explicit Expected #1) by extending the existing type-name sanitization convention to the two remaining leaf positions. It is a print-time fallback only: fidelity is still degraded to `Partial` from the unchanged IR names, so honest failure stays visible. It does not, by itself, make these methods bind — that is the follow-up raise stack (S1–S5).

### Benchmark target

Benchmark target: `StackExchange.Redis@3.0.17`

dotnet-inspect command:

```bash
dotnet-inspect member "StackExchange.Redis.TaskExtensions" "RedisFireAndForget:1" --package StackExchange.Redis@3.0.17 --all -S "Decompiled Source"
```

### Original source

```csharp
internal static void RedisFireAndForget(this Task task) => task?.ContinueWith(static t => GC.KeepAlive(t.Exception), TaskContinuationOptions.OnlyOnFaulted);
```

### Before

```csharp
internal static void RedisFireAndForget(this System.Threading.Tasks.Task task)
{
    if (task is not null)
    {
        task.ContinueWith(new Action<Task>(___c.___9.<RedisFireAndForget>b__8_0), (TaskContinuationOptions)327680);
    }
}
```

- Valid: **False** — the raw `<RedisFireAndForget>b__8_0` method-group name is `CS1001` ("identifier expected") at `<`; the body does not parse. (The type qualifier `<>c` already sanitized to `___c` on base; only the method name leaked.)
- Correct: False
- IL fidelity: not currently checkable (invalid syntax cannot recompile)
- Taste applied: None
- Commit: `f1979d79`

The raw `<…>` method name is a parse error, which is strictly worse than a bind error: it can corrupt parsing of the surrounding valid syntax and blocks all downstream analysis of the body.

### After

```csharp
internal static void RedisFireAndForget(this System.Threading.Tasks.Task task)
{
    if (task is not null)
    {
        task.ContinueWith(new Action<Task>(___c.___9.__RedisFireAndForget_b__8_0), (TaskContinuationOptions)327680);
    }
}
```

- Valid: **Parses** (no raw `<…>`, `CS1001` gone); still does not fully bind — `___c` (the sanitized `<>c` closure holder) resolves to nothing (`CS0246`). So bind-validity is not yet achieved here.
- Correct: False (an honest sanitized fallback, not the real lambda)
- IL fidelity: not currently checkable (does not bind)
- Taste applied: None
- Commit: `b2ae90a8`

The change is the method name only: `<RedisFireAndForget>b__8_0` → `__RedisFireAndForget_b__8_0`. Fidelity stays `Partial` (asserted by `UnspeakableNameFidelityTests`), because the spellability check runs on the unchanged IR name, not the printed text. The same fix covers the other three repro positions in #3129: the `System.CommandLine` `WriteColumns` method groups, the `EncLocalInfo` inline-array helper, and the `dotnet-inspect.any` `FormatOptionEntry` single-element list — none leak a raw `<…>` identifier after this change.

### Fully raised

```csharp
internal static void RedisFireAndForget(this Task task)
{
    task?.ContinueWith(static (Task t) => GC.KeepAlive(t.Exception), TaskContinuationOptions.OnlyOnFaulted);
}
```

- Required tracking issue: #3129 — the eager-singleton non-capturing lambda raise (stack slice **S1**) restores the real lambda body and bind-validity; slices S2–S5 cover the shared-display-class, iterator-local-function, inline-array-span, and single-element-list shapes. This PR is the floor those raises build on.

## Evidence

| Check | Baseline (`f1979d79`) | Head (`b2ae90a8`) |
| --- | --- | --- |
| Focused tests | n/a (new) | `CSharpNamingTests` + `UnspeakableNameFidelityTests`: 52 passed |
| Decompiler fast suite | 3726, 1 failed | 3727, 1 failed |
| Real witness | `TaskExtensions::RedisFireAndForget` leaks raw `<…>b__8_0` (`CS1001`) | raw `<…>` leak gone; parses |

The one fast-suite failure is pre-existing and unrelated: `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString`, a CRLF-vs-LF line-ending mismatch tracked by #3018 (same test, same reason on Baseline and Head). Head total is +1 for this PR's new `ConstructorDelegateTarget_DegradesToPartial` test.

## Decompiler quality

**Conclusion:** **PASS (not the relevant axis)** — S0 is a print-time fallback, not a raise; it does not change the *fully-raised* percentage, so the fully-raised corpus gate does not measure it. Its axis is the *leaked-identifier / validity* signal: it removes the raw `<…>` identifier family (the issue's dominant `CS1001` rows) and is demonstrated by the repros above plus the fast suite showing no regressions. The bind-validity (`CS0246`) that remains is addressed by the raise slices, not by this floor.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.CSharpNamingTests" -class "ILInspector.Decompiler.Tests.UnspeakableNameFidelityTests"
```


richlander:

Two different non-Claude models reviewed this change in isolated review worktrees (scratch under `/tmp`, no `git reset`/`add`/commit in the review trees), per the roster in `AGENTS.md`.

- **GPT** (`gpt-5.6-sol`)
- **Gemini** (`gemini-3.1-pro-preview`)

Reviewed head: the fixed head — reviewed as `13caf9ff` before the pre-first-push rebase onto `origin/main` (`f1979d79`); the rebase was a clean replay and the reviewed diff is byte-identical to the pushed head `b2ae90a8`.

### Findings

**GPT — 1 blocking finding (constructor delegate/method-group target masked fidelity):** the initial fallback sanitized a constructor used as a delegate/method-group target (legal IL `ldftn .ctor`) from `.ctor` to `__ctor`, but `CSharpSpellability.MethodIssue` exempts `.ctor` (correct for base/this *call* chains, which render `base(...)`/`this(...)`). That exemption let fidelity stay `Full` while the printed name was a sanitized fallback — a silent wrong-bind if a real `__ctor` existed, and a regression of the "keep failure visible" rule (the old code emitted a loud invalid `w..ctor`).

- **Reproduced** on a clean review worktree with a focused test (Expected `Partial`, Actual `Full`) — confirmed valid.
- **Resolved** in commit `b2ae90a8` (reviewed as `13caf9ff`): split `MethodIssue` so `AddressOfMethod`/`DelegateCreation` route through a new `MethodGroupTargetIssue` with **no** `.ctor` exemption, while `Call` keeps the exemption for base/this chains. Constructor delegate targets now degrade to `Partial`. C# never emits a delegate-to-constructor, so degrading this position cannot regress a real valid case. Covered by the new `ConstructorDelegateTarget_DegradesToPartial` test.

**Gemini — clean** throughout (no blocking or non-blocking findings).

### Re-review of the fixed head

Both models re-reviewed the fixed head and reported **clean**:

- GPT verified base/this constructor *calls* stay `Full`, and `.cctor` / accessors / generic-constructor method-group targets degrade to `Partial`.
- Gemini clean.

### Verdicts

| Reviewer | Model | Fixed-head verdict |
| --- | --- | --- |
| GPT | `gpt-5.6-sol` | Clean |
| Gemini | `gemini-3.1-pro-preview` | Clean |

Both fixed-head reviews are clean. Remaining bind-validity (`CS0246`) is intentional and out of scope for this floor; it is addressed by the follow-up raise stack (S1–S5), tracked in #3129.

richlander:
Ready to merge.

Note on reviewed head: #3204's adversarial-review reconciliation above was performed at `b2ae90a8` (the sanitizer change). Since then, S5 (#3216) was merged into this base branch, advancing the head to `d8abe0e6`. That S5 content was independently adversarially reviewed clean by the same two non-Claude models (Gemini `gemini-3.1-pro-preview` + GPT `gpt-5.6-sol`) at fixed head `e56c9dfd` — see #3216's reconciliation (3 blocking + 1 non-blocking findings, all resolved). CI is green on the combined head `d8abe0e6` (test/build-net10/csharp-diff-smoke all SUCCESS) and the PR is MERGEABLE/CLEAN. Every change in the merged head has passed 2-model non-Claude adversarial review.

PR #3193, opened 2026-07-25
Add discoverable --gate preset flag to the decompiler test executable

richlander:
## What

Adds a discoverable `--gate <preset>` flag to the `ILInspector.Decompiler.Tests`
executable that expands to the existing xUnit `-trait`/`-trait-` filters, and
wires Deep Inspect to it so the decompiler test lane no longer risks the 6h job
timeout.

## Why

The test EXE only exposed the raw `Speed`/`Area` trait spellings, which are easy
to mistype. More importantly, an *unfiltered* run includes the multi-hour
`Corpus` sweep, and the Deep Inspect **test** job (no `timeout-minutes`, so the
360-minute default) runs that sweep in the same budget as the CLI, analysis, and
IL round-trip gates — which is what pushes the combined job past its timeout.

## Changes

- **Custom entry point.** Disable xunit.v3's auto-generated entry point
  (`XunitAutoGeneratedEntryPoint=false`) and add `Program.Main`, which mirrors
  the generated Microsoft-Testing-Platform/console dispatch **verbatim** and only
  pre-processes `--gate` before delegating. Invocations without `--gate` behave
  exactly as before.
- **Reusable expander.** `GateArgumentExpander` + `GatePreset` (under `Gating/`)
  carry no project-specific knowledge, so the machinery can be shared by other
  test executables later. The Decompiler preset table lives in the entry point.
- **Presets:** `all`, `fast`, `slow`, `no-corpus`, `corpus`, `roundtrip`,
  `fidelity`, `validity`. `--gate list` prints the table; unknown or repeated
  presets are an error.
- **CI wiring (the actual timeout fix).** The Deep Inspect test lane now runs
  `--gate no-corpus`; a new `decompiler-corpus` job runs `--gate corpus` with its
  own 6h job budget, in parallel on a separate runner. A same-job step split
  would not have helped — the two now have independent budgets.
- **Docs.** `docs/decompiler-correctness-pipeline.md` documents `--gate`;
  AGENTS.md notes it beside the trait examples.

## Validation (local)

- Build clean — no duplicate entry point after disabling the auto-generated one.
- 11 `GateArgumentExpander` unit tests pass (expand / passthrough / help / error /
  case-insensitive / repeated-flag / empty-preset).
- `--gate list` exits 0; unknown preset exits 2.
- Preset composition: `--gate fast -class GateArgumentExpanderTests` → 11 tests;
  plain `-class …` passthrough → 11 tests.
[...]

PR #3205, opened 2026-07-25
Schedule Deep Inspect slow lanes weekly and surface failures

richlander:
## What

The weekly Deep Inspect cron (`0 9 * * 1`) previously ran **only** the `package-sweep` lane. The slow `test` lane and the `decompiler-corpus` lane (the multi-hour sweep split out in #3193) ran on **manual dispatch only**, so the slow decompiler gates had **no automated cadence** — nothing scheduled ever exercised them. A failed scheduled run also had no surfacing beyond the red run itself (no notification, no tracking issue), so a failure could go unnoticed.

## Change

- Gate the `test` and `decompiler-corpus` jobs on the `schedule` event so both run on the weekly cron alongside `package-sweep` (they remain available on manual dispatch as before).
- Add a `report-scheduled-failure` job that, on a **scheduled** run where any of those three lanes failed, opens (or comments on) a deduplicated tracking issue titled "Deep Inspect scheduled run failed", including the run URL. Manual dispatches are intentionally excluded — they are watched by the dispatcher.

`census` and `nightly` remain manual-only (out of scope here); see follow-ups.

## Evidence

- YAML parses cleanly (YamlDotNet `RepresentationModel`): all six jobs present, including the new `report-scheduled-failure`.
- No tabs; jobs indent at 2 spaces mirroring the existing siblings.
- Failure-surfacing job uses the standard `permissions: { issues: write }` + `GH_TOKEN: ${{ github.token }}` pattern; dedupes by exact title via `gh issue list ... --jq` before creating.

Deep Inspect only runs on dispatch/schedule, so PR CI does not exercise this workflow; the scheduling/issue logic will first execute on the next weekly cron (or a manual `schedule`-equivalent dispatch).

## Risk

Low — CI-only, no product path. Mechanical `if:` gating plus one new reporting job. No adversarial review per `AGENTS.md` (simple CI change). One operational caveat: the reporter relies on `GITHUB_TOKEN` being permitted to create issues; if org policy restricts that, the reporter job fails visibly on an already-red run and we adjust.

## Follow-ups (non-blocking)

- Apply the same `no-corpus`/`corpus` split to `release.yml`, which still runs the decompiler suite unfiltered (same timeout exposure).
- Decide whether to also schedule `census` weekly.

### DIFF
--- a/.github/workflows/deep-inspect.yml
+++ b/.github/workflows/deep-inspect.yml
@@ -84,7 +84,9 @@ jobs:
 
   test:
     name: Test lane
-    if: inputs.lane == 'test' || inputs.lane == 'all'
+    # Runs on the weekly schedule (automated slow-gate coverage) and on manual
+    # dispatch (lane test/all).
+    if: github.event_name == 'schedule' || inputs.lane == 'test' || inputs.lane == 'all'
     runs-on: ubuntu-26.04
     steps:
       - uses: actions/checkout@v6
@@ -150,7 +152,8 @@ jobs:
     # Split out of the test lane so the multi-hour corpus sweep gets its own
     # 6h job budget instead of sharing the test job's budget with the other
     # slow gates (which previously pushed the combined job past its timeout).
-    if: inputs.lane == 'test' || inputs.lane == 'all'
+    # Runs alongside the test lane on the weekly schedule and on dispatch.
+    if: github.event_name == 'schedule' || inputs.lane == 'test' || inputs.lane == 'all'
     runs-on: ubuntu-26.04
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


richlander:
Ready to merge

Merge-path CI is green: `test (ubuntu-26.04, linux-x64)` passed (8m12s) and `changes` passed; `build-net10`, `markdownlint`, `pack`, and the diff-smoke jobs correctly path-skip since only `.github/workflows/deep-inspect.yml` changed. YAML validated locally (YamlDotNet: all six jobs parse).

Note: Deep Inspect only runs on dispatch/schedule, so PR CI does not exercise this workflow. The new scheduling + issue-reporting logic will first execute on the next weekly cron. Low-risk, CI-only, no product path — no adversarial review per AGENTS.md.

PR #3208, opened 2026-07-25
Render struct boxed-this interface access with the erased ((I)this) cast (#3201)

### LINKED ISSUE #3201: Decompiler: DIM/explicit-interface access via boxed struct this renders invalid C# (missing ((I)this) cast), sibling of #3128
## Summary

Sibling of #3128, for **value types**. A default-interface-member (DIM) or explicit-interface-implementation member reached through `this` on a **struct** renders as `(this).M()` / `(this).Value()`, which fails to compile (CS1061 — the member is not a member of the struct).

#3128 fixed the **class** case: a class→interface conversion emits no IL (implicit reference conversion), so the printer re-inserts the erased `((I)this)` cast. That fix is guarded by `receiver is LoadArgument { Index: 0, Name: "this" }` in `CallText`/`MethodGroupText`. For a struct, the receiver is a **boxed** `this` (a distinct IR shape), so it falls through to the plain receiver path and the cast is not re-inserted.

This is a **pre-existing** gap (predates #3128) and a **distinct conversion category**: for a struct, `((IFace)this)` is a **boxing** conversion that emits `box` IL, not the no-IL reference conversion #3128 addressed.

## Repro

```csharp
public interface IStructFace { void M(); }
public struct StructExplicit : IStructFace
{
    void IStructFace.M() { }
    public void Test() => ((IStructFace)this).M();
}

public interface IDimS { int Value() => 7; }
public struct StructDim : IDimS
{
    public int TestDim() => ((IDimS)this).Value();
}
```

## Current (invalid) rendering

```csharp
public void Test() { (this).M(); }          // CS1061
public int TestDim() => (this).Value();     // CS1061
```

Original IL (`StructExplicit.Test`):

```il
ldarg.0
ldobj    StructExplicit
box      StructExplicit
callvirt IStructFace::M()
```

## Faithful spelling

```csharp
((IStructFace)this).M()
((IDimS)this).Value()
```

For a struct, `((IFace)this)` is a boxing conversion, so it recompiles to the original `ldobj; box; callvirt` — the cast is opcode-faithful.

## Why it is separate from #3128

- #3128 is explicitly the **no-IL class→interface reference conversion**; the struct case is a **boxing conversion (emits `box`)**.
- The receiver IR shape differs: a boxed `this` (`Box` node) rather than a bare `LoadArgument{0,"this"}`. The fix must detect a boxed-this receiver whose callee declaring type is a confirmed interface and re-insert the cast in both `CallText` and `MethodGroupText`, while leaving all other boxed receivers untouched.
- Discovered during adversarial review of the #3128 fix (PR #3200).

## Notes

- Must reuse the same `IrFunction.InterfaceTypes` confirmed-interface plumbing added in #3128 (decline when interface-ness is unknown; never guess).
- Needs positive tests for struct explicit-interface calls, struct DIM calls, and struct interface method groups, plus a negative test that an ordinary boxed receiver (`(object)this` for a genuine `object` member) is not over-cast.

richlander:
- Fixes #3201
- Renders struct/value-type default-interface-member and explicit-interface access through `this` with the required `((I)this)` cast instead of invalid `(this).M()`
- Card revision: `30efd400`

Stacked on #3200 (base branch `fix/3128-dim-interface-cast-this`). #3200 fixed the reference-type (class→interface, no-IL upcast) case; this is its value-type sibling, where the interface cast is a **boxing conversion** that emits `ldobj; box`.

## Change

**Conclusion:** **PASS** — a struct reaching an interface member through `this` boxes the receiver, so the IR target is a `Box` over `this` rather than a bare `LoadArgument`; the printer dropped the cast and emitted CS1061 `(this).M()`. Re-inserting `((I)this)` is opcode-faithful because the cast re-emits exactly the `ldobj; box` the boxing produced.

### Benchmark target

Benchmark target: `ILInspector.Decompiler.Tests` fixture assembly (`StructDimConsumer`, `StructExplicitFace`, added by this PR).

dotnet-inspect command:

```bash
dotnet-inspect member ILInspector.Decompiler.Tests.StructDimConsumer.DimCall -S "Decompiled Source"
```

### Original source

```csharp
public struct StructDimConsumer : IDimSFace
{
    public int DimCall() => ((IDimSFace)this).Value();
```

IL anchor: `ldarg.0; ldobj StructDimConsumer; box StructDimConsumer; callvirt IDimSFace::Value(); ret`.

### Before

```csharp
public int DimCall() => (this).Value();
```

- Valid: False
- Correct: False
- IL fidelity: False
- Taste applied: None
- Commit: `f2568a98`

`(this)` is the boxed struct receiver with the interface cast erased; `IDimSFace.Value()` is a DIM, not a member of `StructDimConsumer`, so this is CS1061.

### After

```csharp
public int DimCall() => ((IDimSFace)this).Value();
```

- Valid: True
- Correct: True
- IL fidelity: True
- Taste applied: None
- Commit: `30efd400`

Byte-identical to the authored fixture source, which is the compiled form of the shown IL, so it recompiles to exactly `ldobj; box; callvirt`.

### Fully raised

## Evidence

| Check | Baseline (`f2568a98`) | Head (`30efd400`) |
| --- | --- | --- |
| Focused tests | `ThisQualification[Decision]Tests`: 54 passed | `ThisQualification[Decision]Tests`: 58 passed (+4) |
| Decompiler fast suite | 3729, 1 failed | 3729, 1 failed |
| Reduced fixture validity | `(this).Value()` invalid (CS1061) | `((IDimSFace)this).Value()` valid |
| Reduced fixture fidelity | not faithful | recompiles to `ldobj; box; callvirt` |
| Real witness | `StructDimConsumer::DimCall`, `StructExplicitFace::CallExplicitInterface` broken | fixed |

The 1 fast-suite failure is pre-existing and unrelated: `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString` fails Windows-locally on CRLF vs LF (`\r\n` expected vs `\n`), identical on baseline and head.

## Decompiler quality

**Conclusion:** **PASS** — pinned PR quick gate matched baseline tolerances with zero pass bugs; lowering residue edged down as some struct interface-cast residue resolves.

### PR quick gate

Run: PR quick corpus, hash-stable 100 methods per assembly (15 assemblies, 1,500 methods); validity/fidelity not run (`--compile-cap 0 --corpus-fidelity-cap 0`).

| --- | ---: | ---: | ---: |
| Detected lowering residue (-) | 16.73% | 16.40% | -0.33pp |
| Conditional-branch residue (-) | 3.07% | 2.93% | -0.14pp |
| Pass bugs (-) | 0 | 0 | 0 |

> **Conclusion:** **PASS** — corpus sensor matched baseline tolerances; 0 pass bugs.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.ThisQualificationTests" -class "ILInspector.Decompiler.Tests.ThisQualificationDecisionTests"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ThisQualificationTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ThisQualificationTests.cs
@@ -215,6 +215,59 @@ public sealed class ThisQualificationTests
         Assert.DoesNotContain("this.FaceMethod()", text);
     }
 
+    // The value-type sibling of the #3128 arms (#3201). A struct reaches an
+    // interface member through `this` by BOXING — `((IDimSFace)this).Value()`
+    // lowers to `ldarg.0; ldobj Struct; box Struct; callvirt IDimSFace::Value()`,
+    // so the IR receiver is a Box over the this value, not a bare LoadArgument.
+    // The printer must still re-insert the `((I)this)` cast; bare `(this).Value()`
+    // is CS1061. Faithful: the cast re-emits exactly the `ldobj; box` the struct
+    // boxing produced.
+    [Fact]
+    public void StructDefaultInterfaceMemberCall_RendersInterfaceCast_ByDefault()
+    {
+        var text = RenderMember(typeof(StructDimConsumer), nameof(StructDimConsumer.DimCall));
+        Assert.Contains("((IDimSFace)this).Value()", text);
+        Assert.DoesNotContain("(this).Value()", text);
+    }
+
+    // A struct DIM method group over `this` boxes the same way:
+    // `((IDimSFace)this).Value` recompiles to `ldarg.0; ldobj S; box S; dup;
+    // ldvirtftn IDimSFace::Value; newobj Func`. Bare `(this).Value` (CS1061) or the
+    // knob's `this.Value` would not bind.
+    [Fact]
+    public void StructDefaultInterfaceMemberGroup_RendersInterfaceCast_ByDefault()
+    {
+        var text = RenderMember(typeof(StructDimConsumer), nameof(StructDimConsumer.DimGroup));
+        Assert.Contains("((IDimSFace)this).Value", text);
+        Assert.DoesNotContain("(this).Value", text);
+    }
+
+    // An explicit interface implementation on a struct invoked through `this`
+    // reaches the call site with the interface as declaring type and a boxed
+    // receiver; it must render `((IStructFace)this).FaceMethod()` — bare
+    // `(this).FaceMethod()` is CS1061 (#3201).
+    [Fact]
+    public void StructExplicitInterfaceCall_RendersInterfaceCast()
+    {
+        var text = RenderMember(typeof(StructExplicitFace),
+            nameof(StructExplicitFace.CallExplicitInterface));
+        Assert.Contains("((IStructFace)this).FaceMethod()", text);
+        Assert.DoesNotContain("(this).FaceMethod()", text);
+    }
+
+    // Negative guard: a boxed `this` that reaches a NON-interface member (here
+    // `object.ToString()` via `((object)this).ToString()`) must NOT be over-cast to
+    // the enclosing struct — `IsInterfaceCastThisReceiver` declines a non-interface
+    // callee, so the boxed-this arm never fires. The faithful render keeps the
+    // object cast the boxing spells and never writes `((StructBoxedObjectReceiver)this)`.
+    [Fact]
+    public void StructBoxedThis_NonInterfaceCallee_IsNotOverCast()
+    {
+        var text = RenderMember(typeof(StructBoxedObjectReceiver),
+            nameof(StructBoxedObjectReceiver.CallObjectToString));
+        Assert.DoesNotContain("StructBoxedObjectReceiver)this", text);
+    }
+
     [Fact]
     public void EventSubscription_DefaultsToBareName()
     {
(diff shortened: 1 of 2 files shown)


richlander:

Two independent fixed-head reviews at exact head `30efd400`, each in an isolated review worktree (scratch under temp; no index/history mutation). Both **APPROVE**.

| --- | --- |
| Gemini 3.1 Pro (`gemini-3.1-pro-preview`) | APPROVE |
| GPT-5.6 (`gpt-5.6-sol`) | APPROVE |

Both reviewers built the full solution, ran the focused suite (58 passed), and compiled real C#/IL to prove behavior. No blocking findings; no regressions introduced by this PR.

### Verified (both reviewers, real runs)
- **Fidelity:** the struct interface call/method-group emits `ldarg.0; ldobj S; box S; callvirt I::M`; the re-inserted `((I)this)` recompiles to exactly that. Gemini and GPT both confirmed a *concrete* struct `this` requires `ldobj; box` (never `constrained.`), and GPT separately verified a real CoreLib `constrained. T; callvirt` on an unconstrained generic `T` is untouched (different shape, correctly not matched).
- **Over-firing declined:** the interface gate excludes `object`/`ValueType`; boxed locals/fields/other arguments cannot match the `LoadArgument{0,"this"}` pattern (Gemini compiled `((I)s).M()` on a `ref S s` → `ldarg.1`, correctly skipped). `Box.ResultType` is `object`, so the method-group arm can never steal a pointer receiver.
- **Under-firing:** readonly and generic structs all lower to the same `Box{LoadIndirect{LoadArgument{0,"this"}}}` shape, so the fix covers them.
- **Explicit interface impl on a struct** (`StructExplicitFace`) and the **qualify-method knob**/enclosing-instantiation gate all behave correctly.

### Non-blocking findings — all pre-existing at base `f2568a98`, none a regression; no action this PR
1. **Generic method groups drop explicit type arguments** (both reviewers). `MethodGroupText` emits only `{name}`, so a generic interface method group over a boxed struct `this` renders `Make` instead of `Make<int>` (CS0411). This is the same documented pre-existing `MethodGroupText` gap the reference-type #3128 arm already inherits — not introduced here. Out of scope.
2. **Non-interface boxed-`this` erases its boxing cast** (GPT: `StructBoxedObjectReceiver.CallObjectToString` → `(this).ToString()` instead of `((object)this).ToString()`; real CoreLib `CustomAttributeNamedArgument.GetHashCode` → `(this).GetHashCode()` while IL boxes and calls `ValueType.GetHashCode()`). This PR's interface-only gate *correctly declines* this case (covered by the `StructBoxedObjectReceiver` negative test); the underlying non-interface boxed-`this` fidelity gap is pre-existing and unrelated.
3. **Boxed `ref`/local params to an interface** (Gemini: `((I)s).M()` → invalid `(s).M()`). Confirmed pre-existing at base via `git checkout f2568a98`; the sibling of this fix for non-`this` boxed receivers. Not in scope.

Findings 2 and 3 are legitimate future follow-ups (the non-`this` / non-interface boxed-receiver fidelity family), tracked separately from this interface-`this` fix.

richlander:
Ready to merge

Stacked on #3200 (base `fix/3128-dim-interface-cast-this`). Merge #3200 first, then retarget this PR to `main` — the merge-path `test` CI (which only runs for PRs into `main`) will run at that point. All validation available for the stacked head `30efd400` is clean:

- Full-solution Release build: pass
- Focused tests (`ThisQualification[Decision]Tests`): 58 passed (+4 new)
- Decompiler fast suite: 3729, 1 failed — pre-existing Windows-local CRLF `SmallStringSwitch_RendersSwitchOnString`, identical at base
- PR quick corpus gate: PASS, 0 pass bugs (lowering residue 251→246)
- Adversarial review: Gemini 3.1 Pro + GPT-5.6 both APPROVE at exact head; only pre-existing, non-blocking findings (reconciled above)

richlander:
Follow-up issues filed for the pre-existing (non-regression) findings from the adversarial review, tracked separately from this interface-`this` fix:

- #3213 — non-interface boxed-`this` erases its boxing cast (`(this).GetHashCode()` vs `box; callvirt ValueType::GetHashCode`); correctness/fidelity.
- #3214 — boxed non-`this` receiver reaching an interface member renders invalid `(s).M()` (the non-`this` sibling of #3201).
- #3215 — `MethodGroupText` drops generic type arguments (`Make<int>` → `Make`, CS0411).

PR #3216, opened 2026-07-25
Raise single-element read-only collection expressions (#3129, S5)

richlander:
- Advances #3129
- Raises the csc single-element read-only collection-expression lowering (`new <>z__ReadOnlySingleElementList<T>(x)`) back to `[x]`
- Card revision: `e56c9dfd`

> **Stacked PR.** Base is the S0 floor branch `fix/3129-unspellable-name-fallback` (#3204), not `main`. Review only the two commits on top of `b2ae90a8`.

## Change

**Conclusion:** **PASS** — a previously-unraised, unspellable construct now raises to a valid, correct, Full-fidelity collection expression, guarded so it fires only on the exact lowering and only where `[x]` is a legal target.

### Benchmark target

Benchmark target: `dotnet-inspect.any@0.14.0` (real witness `DotnetInspector.CommandLine.HelpWriter::FormatOptionEntry`).

dotnet-inspect command:

```bash
dotnet-inspect member DotnetInspector.CommandLine.HelpWriter FormatOptionEntry -S "Decompiled Source"
```

### Original source

```csharp
private static (string Name, string Description) FormatOptionEntry(Option option)
{
    var names = option.Aliases
        .Concat([option.Name])   // <-- single-element collection expression targeting IEnumerable<string>
        .Where(a => a.StartsWith('-'))
        .Distinct()
        .OrderBy(a => a.TrimStart('-').Length)
        .ThenBy(a => a);
    // ...
}
```

`[option.Name]` targets the `IEnumerable<string>` parameter of `Enumerable.Concat`, which csc lowers to `new <>z__ReadOnlySingleElementList<string>(option.Name)`.

### Before

```csharp
// S_0 = option.Aliases.Concat<string>(new ___z__ReadOnlySingleElementList<string>(option.Name));
S_0 = option.Aliases.Concat<string>(new ___z__ReadOnlySingleElementList<string>(option.Name));
```

- Valid: False — parses (S0 sanitized the `<>` name to `___z__…`) but does not bind; `___z__ReadOnlySingleElementList` is not a real type (CS0246).
- Correct: False — cannot bind, so no observable behavior.
- IL fidelity: False — does not compile.
- Taste applied: None introduced by this slice.
- Commit: `b2ae90a8`

The S0 floor keeps the body parseable and honestly `Partial`; binding is this raise's job.

### After

```csharp
// S_0 = option.Aliases.Concat<string>([option.Name]);
S_0 = option.Aliases.Concat<string>([option.Name]);
```

- Valid: True — `[option.Name]` target-types to `Concat`'s `IEnumerable<string>` parameter and binds.
- Correct: True — a single-element `IEnumerable<string>` yielding `option.Name`, identical to the lowering.
- IL fidelity: True in principle — a single-element collection expression targeting a read-only collection interface is csc's exact source for the `<>z__ReadOnlySingleElementList<T>` lowering (not separately compile-back-checked in this PR; the decompiler's own fidelity classification is **Full**).
- Taste applied: None introduced by this slice (explicit `Concat<string>` type args are present in both Before and After).
- Commit: `e56c9dfd`

### Fully raised

The single-element-list construct is in the fully raised state (`[option.Name]`, Full fidelity). The residual `S_0`/`S_5` synthesized stack-slot locals in the surrounding method are an unrelated concern tracked separately.

- Required tracking issue: #3129 — remaining unspellable-name slices (eager lambda, shared display class, iterator local functions, inline-array span) and method-level stack-slot local naming.

## Implementation

Folded into `InlineArrayCollectionPass` (the single register-model owner of the `CollectionExpression` lowering — `LoweringCoverage` maps one Roslyn LocalRewriter to one pass, and single-element lists are part of `BoundCollectionExpression`, so a second pass is not representable). `RaiseSingleElementLists`:

- matches a `NewObject` whose constructor declaring type is a generic instance with the **exact** metadata name ``<>z__ReadOnlySingleElementList`1`` and one type argument;
- resolves a target type from the immediate use-site (call/constructor argument parameter type, typed local store, or an enclosing raised collection expression's element type);
- raises only when that target is **exactly** one of `IEnumerable<T>` / `IReadOnlyCollection<T>` / `IReadOnlyList<T>` over the list's own element type;
- leaves the construction flat (honest `Partial`) for extension-method receivers, wider/erased sinks, and every other context.

## Evidence

| Check | Baseline (`b2ae90a8`) | Head (`e56c9dfd`) |
| --- | --- | --- |
| Product build (`dotnet-inspect.slnx`) | Pass | Pass |
| Focused tests | `SingleElementListCollectionTests`: n/a (new) | `SingleElementListCollectionTests`: 12 passed |
| Decompiler fast suite (`-trait- Speed=Slow`) | 3740, 1 failed | 3752, 1 failed |
| Real witness | `HelpWriter::FormatOptionEntry` `Partial` (`new ___z__ReadOnlySingleElementList<string>(…)`) | `HelpWriter::FormatOptionEntry` **Full** (`[option.Name]`) |

The single fast-suite failure is the pre-existing `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString` (CRLF-vs-LF, #3018) — identical on baseline and head, unrelated to this change. The `+12` focused tests are this PR's new `SingleElementListCollectionTests`.

## Decompiler quality

**Conclusion:** **ADVISORY** — the full PR corpus quick-gate was not run for this focused slice. The change is purely additive and tightly guarded: it only converts a previously-residual, unspellable `<>z__ReadOnlySingleElementList<T>` construction into a raised collection expression, and only when the use-site target is exactly a matching read-only collection interface. No existing raise path is modified, so no previously-raised method can regress; residue can only decrease or hold.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class "ILInspector.Decompiler.Tests.SingleElementListCollectionTests"
```


richlander:

Two non-Claude models reviewed this slice in isolated, read-only worktrees.

- **Reviewers:** Gemini (`gemini-3.1-pro-preview`) and GPT (`gpt-5.6-sol`).
- **Initial reviewed head:** `db8d721a`.
- **Fixed head re-reviewed:** `e56c9dfd`.

### Findings and resolution

| # | Finding | Reviewer(s) | Severity | Resolution |
| 1 | Wider/erased sink mistaken for a collection-expression target — `Consume((IEnumerable<string>)[x])` where `Consume(object)` (the reference conversion emits no IL) would print `Consume([x])` → CS9174; covariant `IEnumerable<object>` sinks silently retype elements. | GPT | Blocking | `e56c9dfd` — new `IsConstructibleReadOnlyListTarget` guard: raise only when the resolved target is exactly `IEnumerable<T>`/`IReadOnlyCollection<T>`/`IReadOnlyList<T>` over the list's own element type. |
| 2 | Extension-method argument zero — an unreduced extension call such as `Enumerable.Concat([x], other)` (valid source) is printed in reduced instance form, making arg0 a member-access receiver; `[x].Concat(other)` is CS9176. | Gemini + GPT | Blocking | `e56c9dfd` — `TryResolveCollectionTargetType` returns false for arg0 of a confirmed extension method (`IsExtension == Yes && ChildIndex == 0`). |
| 3 | Nested single-element lists `[[x]]` — the outer list raises first and re-parents the inner construction under the new collection expression, which had no target-resolution case, so the inner list leaked flat (unspellable type). | Gemini | Blocking | `e56c9dfd` — added a `CollectionExpression` parent case resolving the element target to `collection.ElementType`, so the inner list raises too. |
| 4 | Prefix-only name match (`StartsWith`) also matches ILAsm-authored look-alikes such as `<>z__ReadOnlySingleElementListFake`. | GPT | Non-blocking | `e56c9dfd` — match the exact metadata name ``<>z__ReadOnlySingleElementList`1``. |

Six regression tests were added, one per case (extension receiver left flat, extension argument raised, nested `[[x]]` fully raised, wider `object` sink left flat, covariant sink left flat, look-alike type not raised).

### Verified / dismissed

- Both reviewers independently confirmed the non-extension static-call, constructor-argument, and typed-local-store positive cases, and the receiver/return negative cases, are correct.
- The real witness `HelpWriter.FormatOptionEntry` (single-element list is arg1 of extension `Concat<string>`, parameter `IEnumerable<string>`) is unaffected by the extension-receiver guard and still raises to `option.Aliases.Concat<string>([option.Name])` at **Full** fidelity — verified by both reviewers on the fixed head.
- No blocking finding was dismissed; all four were actioned.

### Final verdicts (fixed head `e56c9dfd`)

- **Gemini:** CLEAN — re-ran the concrete repros and the focused suite (all passing); confirmed Findings 1–3 resolved and no new regressions.
- **GPT:** `VERDICT: CLEAN` — Findings 1–3 resolved, witness still raises to Full, no new blocking regressions.

PR #3217, opened 2026-07-25
Add ResearchViews.ProjectType: neutral type projection + CLI parity (#3206)

### LINKED ISSUE #3206: Add ResearchViews.ProjectType: shared presentation-neutral type projection for all front ends
## Summary

Introduce a presentation-neutral **type-level projection** in `ILInspector.Research` —
`ResearchViews.ProjectType(...)` — mirroring the existing `ResearchViews.ProjectMember`
seam, so every front end (the CLI's Markdown/terminal renderer and the WASM web UI)
composes type-level metadata from **one** source of truth instead of re-deriving it.

## Motivation

Type-level "front end" projection logic currently lives only in the **CLI executable**:

- `DotnetInspector.Views.TypeView` + row models `TypeParameterRow` / `InterfaceRow` /
  `BaseclassRow` / `EnumValueRow` (`src/dotnet-inspect/Views/ApiViews.cs`), built by
  `ApiOutputFormatter.BuildTypeView(ApiType, …, ApiOptions)`
  (`src/dotnet-inspect/Output/ApiOutputFormatter.cs:324`). This pre-computes modifiers,
  base type, type parameters, interfaces, etc. — but is CLI-coupled and `Markout`-shaped.

The WASM engine (`prototypes/inspect-web/engine/Program.cs` → `ToBrowserType`) already
**re-derives** the type modifier string independently, and the web UI is about to grow a
richer Metadata section (type-relationship graph, composition counts, attributes,
generic parameters). Building that web-only would make the engine a **second full
implementation** of `TypeView`. Worse, the **type-relationship graph** and
**unsafe/async composition counts** have **no product owner today**, so a prototype-only
build would diverge from the CLI permanently.

`ILInspector.Research` is the right home: it already owns `ProjectMember`, and the WASM
engine **already references it** (used for annotated source via
`QueryMemberAnnotatedSource`). Adding `ProjectType` extends that precedent.

## Proposed shape

A presentation-neutral request/result parallel to `MemberProjectionRequest` /
`MemberProjectionResult` (`src/ILInspector.Research/ResearchViews.cs:25`):

```csharp
public sealed record TypeProjectionRequest(
    MetadataSource Source,     // resolves the type; whole-assembly surface for derived-type scan
    string Type,               // full type name / identity
    bool PublicOnly = false,
    bool Composition = true,   // aggregate member counts
    bool RelationshipGraph = true);

public sealed record TypeProjectionResult(
    TypeIdentityView Identity,          // kind, modifiers (sealed/abstract/static/ref/readonly), accessibility, namespace, assembly
    string? BaseType,                   // null / hidden for Object/ValueType/Enum (match Baseclass.CanRender rules)
    IReadOnlyList<string> Interfaces,
    IReadOnlyList<string> DerivedTypes, // via ApiSurfaceExtractor.PopulateDerivedTypes (needs full surface)
    IReadOnlyList<TypeParameterView> TypeParameters,  // name, variance, constraints (use StructuredConstraints)
    IReadOnlyList<string> Attributes,
    string? EnumUnderlyingType,
    TypeCompositionView? Composition,   // counts below
    TypeRelationshipGraph? Graph);      // structured nodes/edges, NOT a Mermaid string
```

**Composition counts** (aggregate over `ApiMember`, all flags already exist:
`IsStatic`, `IsUnsafe`, `IsAsync`, `IsVirtual`, `IsAbstract`, `IsOverride`, extension
flag, obsolete-from-`Attributes`): members by kind (methods/properties/fields/events/
constructors) plus flagged totals (static, **unsafe**, **async**, virtual/abstract/
override, extension, obsolete).

**Relationship graph** — a structured model (nodes carry type identity + role:
`self` / `base` / `interface` / `derived`; edges carry relation kind), so each front end
projects its own presentation: CLI → Mermaid (see
`docs/design/call-graph-mermaid-projection.md` for the neutral-model→Mermaid pattern),
web → interactive SVG with clickable navigation. Do **not** return a baked Mermaid string
as the only shape.

## Consumers to refactor onto it

1. **CLI**: `ApiOutputFormatter.BuildTypeView` consumes `ProjectType` for base / interfaces /
   type-parameters (and gains composition + graph sections under the existing
   progressive-disclosure/verbosity rules — `docs/design/progressive-disclosure.md`,
   `docs/design/output-shapes.md`). One source of truth.
2. **WASM engine**: `prototypes/inspect-web/engine/Program.cs` emits `TypeProjectionResult`
   as JSON (drop the bespoke modifier re-derivation in `ToBrowserType`); the web client
   renders shape / counts / attributes / inline graph from it. (This is the follow-up the
   web UI branch `prototype/wasm-command-ui` will build on top of.)

## Constraints

- SRM-only, NativeAOT-friendly, Roslyn-free, no inspected-assembly loading (AGENTS.md
  repository-wide constraints) — the engine consumes this in WASM.
- Preserve layer ownership: Metadata owns the facts (`ApiType`), Research composes them
  into the view; presentation (Markout vs HTML) stays in each front end.
- `DerivedTypes` requires a whole-assembly scan (`PopulateDerivedTypes` over
  `surface.Types`), so the request must carry enough scope to build the full surface.

## Evidence

- Structured type projection: unit-test `ProjectType` for a class with base + interfaces,
  a generic type with constraints, an enum (underlying type), a `ref struct` / `readonly
  struct`, and composition counts (a type with known unsafe/async members).
- CLI parity: confirm `api` output for a representative type is unchanged (or intentionally
  enriched) after `BuildTypeView` is refactored onto the projection — exercise Markdown and
  JSON/structured modes per the output contract.

## Related

- Precedent: `ResearchViews.ProjectMember` (`src/ILInspector.Research/ResearchViews.cs:46`).
- Sibling front-end taste seam: #3191 (thread `PrinterOptions` through annotated source).

richlander:
## What & why

Closes #3206. Introduces `ResearchViews.ProjectType(...)` — a presentation-neutral, type-level projection seam in `ILInspector.Research` that mirrors the existing member-level `ProjectMember` seam. Every front end (the CLI's Markdown/terminal renderer and the deferred WASM web UI) can now compose type identity, base/interface/derived relationships, type parameters, attributes, enum underlying type, member composition counts, and a **structured** relationship graph from one source of truth instead of re-deriving them.

**Scope (this PR): seam + CLI parity.** New CLI profile/graph sections and the WASM engine refactor are deferred to follow-ups.

## Changes

- **`src/ILInspector.Research/ResearchTypeViews.cs`** (new): neutral records (`TypeIdentityView`, `TypeParameterView`, `TypeCompositionView`, `TypeRelationshipNode/Edge/Graph`, `TypeProjectionOptions/Request/Result`), a public `TypeModifiers(ApiType)` helper, and two `ProjectType` overloads:
  - core `ProjectType(ApiType, ApiSurface?, options)` — pure composition from already-extracted facts; derived-type scan runs only when a surface is supplied.
  - `ProjectType(TypeProjectionRequest)` — extracts the whole-assembly surface via `MetadataSource`, resolves the type, recovers the body-gated async fact, then composes.
  - The relationship graph is a **structured node/role/edge model, not a baked Mermaid string** (per `docs/design/call-graph-mermaid-projection.md`).
- **`src/ILInspector.Decompiler/Pipeline/MetadataSource.cs`**: additive `ExtractApiSurface(...)` and `ClassifyAsync(token)` so the request path reuses the already-open PE image and recovers async (`ApiMember.IsAsync` is deliberately unpopulated by surface extraction — `docs/design/member-body-substrate.md`).
- **`src/dotnet-inspect/Output/ApiOutputFormatter.cs`**: `BuildTypeView`/`BuildShapeView` now consume the projection for type modifiers, base-type filtering, and interface ordering; the private `BuildTypeModifiers` helper is removed. **`api` output is unchanged.** C# constraint spelling stays in the CLI (`ConstraintSummary`) because Research is Roslyn-free.

## Layering

`ILInspector.Research` references Metadata + Decompiler (reaches `ApiType`, `ApiSurfaceExtractor`, `MetadataSource`) but **not** `ILInspector.CSharp`; the projection carries only neutral raw constraint strings, and C# rendering stays in the CLI.

## Evidence

- `ILInspector.Research.Tests`: **123 passed** (9 new in `ResearchTypeViewsTests`, covering modifiers, base/interface/derived relationships, the neutral graph, generic constraints, enum underlying type, ref/readonly struct modifiers, composition counts, and async recovery via the request path; core-does-not-invent-async negative case; unknown-type throw).
- `dotnet-inspect.Tests` (`ApiOutputFormatterTests` + `OutputFormatterTests` + `CSharpDeclarationFormatterTests`): **155 passed, 1 skipped** (ilasm unavailable), confirming `api` parity.
- Rebased onto latest `origin/main`; rebuilt and re-ran both suites green.

## Follow-ups (deferred)

- New CLI type profile/relationship-graph sections consuming the composition + graph.
- WASM `inspect-web` engine refactor onto the same seam.

Adversarial review to follow per `AGENTS.md` (non-trivial new shape).

### DIFF
--- a/src/ILInspector.Decompiler/Pipeline/MetadataSource.cs
+++ b/src/ILInspector.Decompiler/Pipeline/MetadataSource.cs
@@ -74,6 +74,32 @@ public sealed class MetadataSource : IDisposable
     public DecompilerSymbolSource Symbols => _symbols;
 
     /// <summary>
+    /// Extracts this assembly's <see cref="ApiSurface"/> from the already-open PE image,
+    /// mirroring <c>AssemblyInspectionSession</c>/<c>PdbContext</c> so callers do not re-open
+    /// the file. Presentation-neutral type projection (<c>ResearchViews.ProjectType</c>) uses
+    /// this to compose type-level views from one live source.
+    /// </summary>
+    public ApiSurface ExtractApiSurface(bool includeAll = false, bool typesOnly = false)
+        => ApiSurfaceExtractor.Extract(Pe, includeAll, typesOnly);
+
+    /// <summary>
+    /// Classifies the async shape of the given MethodDef metadata token (runtime or
+    /// state-machine async), or <see langword="null"/> when the token is not a MethodDef or
+    /// the method is not async. Async is a body-gated fact the API surface deliberately omits
+    /// (docs/design/member-body-substrate.md, <c>ApiMember.IsAsync</c>), so callers that need
+    /// an accurate async signal recover it here from live metadata.
+    /// </summary>
+    public MethodClassification? ClassifyAsync(int methodDefToken)
+    {
+        var entity = MetadataTokens.EntityHandle(methodDefToken);
+        if (entity.Kind != HandleKind.MethodDefinition)
+            return null;
+
+        var method = Reader.GetMethodDefinition((MethodDefinitionHandle)entity);
+        return MethodClassificationScanner.ClassifyAsyncMethod(Reader, method);
+    }
+
+    /// <summary>
     /// Opens an assembly. Throws <see cref="BadImageFormatException"/> for files
     /// without managed metadata. <paramref name="externalPdbPath"/> is a portable
     /// PDB to use for source local names when the assembly carries no embedded
--- a/src/ILInspector.Research.Tests/ILInspector.Research.Tests.csproj
+++ b/src/ILInspector.Research.Tests/ILInspector.Research.Tests.csproj
@@ -8,6 +8,7 @@
     <IsPackable>false</IsPackable>
     <IsAotCompatible>false</IsAotCompatible>
     <OutputType>Exe</OutputType>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/ILInspector.Research.Tests/ResearchTypeViewsTests.cs
+++ b/src/ILInspector.Research.Tests/ResearchTypeViewsTests.cs
@@ -1,0 +1,183 @@
+using ILInspector.Decompiler.Pipeline;
+using ILInspector.Metadata;
+using ILInspector.Research.Tests.TypeFixtures;
+
+namespace ILInspector.Research.Tests;
+
+public class ResearchTypeViewsTests
+{
+    static MetadataSource OpenSelf()
+        => MetadataSource.Open(typeof(ResearchTypeViewsTests).Assembly.Location);
+
+    static ApiType FindType(ApiSurface surface, Type clrType)
+    {
+        var type = surface.Types.FirstOrDefault(t => t.FullName == clrType.FullName)
+            ?? surface.Types.FirstOrDefault(t => t.MetadataName == clrType.Name);
+        Assert.NotNull(type);
+        return type!;
+    }
+
+    [Fact]
+    public void ProjectType_AbstractClass_ExposesInterfacesAndDerivedTypes()
+    {
+        using var source = OpenSelf();
+        var surface = source.ExtractApiSurface(includeAll: true);
+        var animal = FindType(surface, typeof(ResearchAnimal));
+
+        var result = ResearchViews.ProjectType(animal, surface);
+
+        Assert.Equal("class", result.Identity.Kind);
+        Assert.Contains("abstract", result.Identity.Modifiers);
+        // System.Object base is trivial and filtered out.
+        Assert.Null(result.BaseType);
+
+        Assert.Equal(
+            [typeof(IResearchAged).FullName!, typeof(IResearchNamed).FullName!],
+            result.Interfaces);
+
+        Assert.Equal(
+            [typeof(ResearchCat).FullName!, typeof(ResearchDog).FullName!],
+            result.DerivedTypes);
+    }
+
+    [Fact]
+    public void ProjectType_RelationshipGraph_UsesNeutralNodeRoleEdgeModel()
+    {
+        using var source = OpenSelf();
+        var surface = source.ExtractApiSurface(includeAll: true);
+        var animal = FindType(surface, typeof(ResearchAnimal));
+
+        var graph = ResearchViews.ProjectType(animal, surface).Graph;
+        Assert.NotNull(graph);
+
+        var self = Assert.Single(graph!.Nodes, n => n.Role == ResearchViews.TypeRelationshipRole.Self);
+        Assert.Equal(typeof(ResearchAnimal).FullName, self.Id);
+
+        Assert.Equal(2, graph.Nodes.Count(n => n.Role == ResearchViews.TypeRelationshipRole.Interface));
+        Assert.Equal(2, graph.Nodes.Count(n => n.Role == ResearchViews.TypeRelationshipRole.Derived));
+        Assert.DoesNotContain(graph.Nodes, n => n.Role == ResearchViews.TypeRelationshipRole.Base);
+
+        Assert.Equal(2, graph.Edges.Count(e => e.Kind == ResearchViews.TypeRelationshipKind.Implements));
+        Assert.All(
+            graph.Edges.Where(e => e.Kind == ResearchViews.TypeRelationshipKind.Implements),
+            e => Assert.Equal(self.Id, e.FromId));
+
+        Assert.Equal(2, graph.Edges.Count(e => e.Kind == ResearchViews.TypeRelationshipKind.DerivedFrom));
+        Assert.All(
+            graph.Edges.Where(e => e.Kind == ResearchViews.TypeRelationshipKind.DerivedFrom),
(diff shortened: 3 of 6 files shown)


richlander:

Two-model adversarial review per `AGENTS.md` (non-trivial new public shape). Reviewers were isolated in separate linked review worktrees at the exact reviewed head; scratch confined to temp, no git mutation in review trees.

- **Reviewer A — Gemini 3.1 Pro**
- **Reviewer B — GPT-5.6**

(Two different model families from the roster; my own family excluded.)

### Round 1 (base `11c89303` → head `b62db9c3`)

Both reviewers verified: CLI `api` parity intact (`TypeModifiers`/base-filter/interface-order identical to the deleted `BuildTypeModifiers`), layering clean (no Roslyn/CSharp leak into `ILInspector.Research`), async recovery correct and free of shared-state leakage (`ExtractApiSurface` builds a fresh `ApiType` graph per call), composition counts mutually exclusive with the enum-field exclusion honored.

**Findings addressed in `dd4507b0` (both in the new, not-yet-consumed `ProjectType(TypeProjectionRequest)` request path — no shipping caller, so neither was a regression):**

1. **[GPT — blocking] `ResolveType` could silently pick the wrong type.** The 3-tier resolver (FullName → namespace-less `MetadataName` → case-insensitive FullName) returned the first match on a non-unique key. **Fixed:** collapsed to a single exact `FullName` (Ordinal) match, mirroring the product's canonical resolver `IrImporter.ResolveMethodHandle` (which matches one exact `reader.GetFullTypeName`). `FullName` is unique within a surface, so no ambiguous/wrong-type binding is possible. New test `ProjectType_Request_ResolvesByExactFullNameOnly` locks the contract (bare name throws, exact FullName resolves).
   - *GPT also noted the CLR `+`-nested form isn't accepted.* This is consistent with existing product behavior: the command layer normalizes `+`→`.` before resolution (`ApiCommand.cs:326`, `ApiAnalysisInspection.cs:82`, `XmlDocFileParser.cs:440`) and then matches the dotted `FullName`. Exact-dotted-`FullName` is therefore the correct seam contract; the `+` convenience remains a command-layer concern. Both reviewers confirmed this is sufficient.

2. **[GPT — blocking] Projection discarded `surface.InspectionFailures`.** A rejected metadata row could drop from the relationship/derived view while the result looked complete. **Fixed:** `TypeProjectionResult` now carries `IReadOnlyList<ApiSurfaceInspectionFailure> InspectionFailures` (reusing the existing typed model), populated from the surface in both overloads (empty, non-null when no surface / none rejected). The CLI already surfaces these at the command layer (`ApiCommand`/`TypeCommand`, unchanged), but a seam consumer that bypasses that layer (the WASM web UI named in #3206) now sees the incompleteness signal. New test `ProjectType_CarriesInspectionFailures_EmptyForCleanSurface` verifies the wiring. Both reviewers agreed whole-surface failures are the right signal since the derived-type scan walks the whole surface.

**Non-blocking, deferred (Gemini), both reviewers agree out-of-scope:**

- *Culture-sensitive `.Order()` on interfaces.* Kept intentionally to preserve exact bug-for-bug CLI `api` parity (this PR's stated guarantee). Canonicalizing to `StringComparer.Ordinal` across the neutral layer is the right move when the second consumer (WASM) lands and it can be validated as a deliberate output change. Verified empirically that Ordinal vs default ordering diverges (case/punctuation), so this belongs in its own PR.
- *`ApiSurfaceExtractor.PopulateDerivedTypes` idempotency.* Pre-existing code not modified here; the seam never calls it twice on the same instance.

### Round 2 — fixed head `dd4507b0` (re-review of the fix delta)

Both reviewers re-ran on the exact fixed head and reported **clean, no blocking issues remaining**. Evidence each reproduced: build 0/0, `ILInspector.Research.Tests` **125 passed**, formatter parity (`ApiOutputFormatterTests`/`OutputFormatterTests`/`CSharpDeclarationFormatterTests`) **155 passed, 1 skip** (ilasm unavailable, unrelated).

richlander:
Ready to merge

- Both fixed-head adversarial reviews (Gemini 3.1 Pro + GPT-5.6) are clean on `dd4507b0`; reconciliation above.
- CI green: `test`, `build-net10`, `csharp-diff-smoke`, `changes` pass (`markdownlint`/`pack`/`il-diff-smoke` are path-gated skips).
- Local: `ILInspector.Research.Tests` 125 pass; `api` formatter parity (`ApiOutputFormatterTests`/`OutputFormatterTests`/`CSharpDeclarationFormatterTests`) 155 pass / 1 ilasm skip.

**Non-blocking follow-ups (out of scope here):**
- New CLI type profile/relationship-graph sections consuming `TypeCompositionView` + `TypeRelationshipGraph`.
- WASM `inspect-web` engine refactor onto this seam.
- Canonicalize interface ordering to `StringComparer.Ordinal` across the neutral layer (currently `.Order()` for exact CLI parity) — do this with the WASM consumer as a deliberate, tested output change.

PR #3199, opened 2026-07-25
Fix RTS explicit-interface method reconstruction (#3112)

richlander:
- Advances #3112
- Changes RTS explicit-interface **method** reconstruction (harness-only; product output unchanged)
- Evidence revision: `4809c7aa977a290bb22c765d09c4b64c4fb322fd`

## Change

> Should we accept this harness change?

**Conclusion:** **PASS** — RTS reconstructed explicit-interface method targets
(`IType.Member`) as plain sanitized methods (`IType_Member`), so the recompiled
method carried the wrong metadata name and the fidelity lookup failed as
`ContextFail` / `method-not-found`; reconstructing them as real explicit-interface
implementations (mirroring the existing property/event handling) round-trips them
Exact, proven by focused round-trip tests. Shapes that cannot be reconstructed as a
bodyless explicit declaration (operators, default interface methods, `static virtual`,
generic/nested interfaces) fall back atomically to the pre-fix plain shape, so no
`RecompileFail` regression is introduced.

Before:

```text
compileBackStatus == ContextFail, detail == method-not-found
// class emits: public void IEnumerable_GetEnumerator() { ... }
// interface emits: interface IEnumerable { }   (member not declared)
```

After:

```text
STATUS = Exact
// class emits: IEnumerator IEnumerable.GetEnumerator() { ... }
// interface emits: interface IEnumerable { IEnumerator GetEnumerator(); }
```

## Scope and safety

- **Root cause:** For method targets, `ComposeMethod` set the member name via
  `Identifier(methodName)`, and `CSharpIdentifier.Sanitize` rewrites the dotted
  explicit-interface name (`System.Collections.IEnumerable.GetEnumerator`) to
  `System_Collections_IEnumerable_GetEnumerator`. The reconstruction compiled, but
  the recompiled metadata method name no longer matched the requested dotted name,
  so the post-recompile `FindMethodDefinition` lookup returned nothing —
  `ContextFail`/`method-not-found`. Properties and events already had
  explicit-interface handling; methods did not. This was the dominant cause of 126
  EVIL-corpus Invalid rows sharing `ContextFail` / `method-not-found` (all target
  names are explicit-interface spellings: `System.IConvertible.ToBoolean`,
  `System.Collections.IEnumerable.GetEnumerator`, ...).
- **Fix shape:** RTS orchestration only. Reconstruct the target as an
  explicit-interface implementation via a `DeclarationSignature` compatibility
  spelling (the C# printer keeps method-shaped explicit implementations on
[...]

Issue #3112, opened 2026-07-23
Harness: RTS explicit-interface target lookup fails (method-not-found, 126 rows) — sibling of #2963/#2964


richlander:
Found via the EVIL authored-corpus benchmark (#3079) on `main` at `d6af0165`. This is a **harness/RTS-quality** issue, not a product decompiler defect — a sibling of #2963/#2964 and in scope for #3052.

126 Invalid rows share `compileBackStatus == ContextFail`, `detail == method-not-found`. In every case the benchmark row's target name is an **explicit interface implementation** name, the method genuinely exists on the real type, and the product decompiles it cleanly when selected with the explicit selector. So the RTS shell/rebuilder is failing to translate the corpus row target name into the product's `explicit:<Interface>.<Member>` selector.

## Dominant explicit-interface prefixes
`System.IConvertible` (16), `System.Collections.IEnumerable` (14), `Pipelines.Sockets.Unofficial.Threading.MutexSlim.IPendingLockToken` (14), `System.ComponentModel.ICustomTypeDescriptor` (12), `Moq.Async.IAwaitableFactory` (8), `Mono.TextTemplating…ITextTemplatingEngineHost` (6), `System.Collections.ICollection` (5), `IArena` (5), `Dapper.SqlMapper.ITypeHandler` (4).

## Verification — the target is real and decompiles cleanly

```bash
dnx dotnet-inspect -y -- member "StackExchange.Redis.SimpleResultBox" --package StackExchange.Redis@3.0.17 --all -S "Member Index"
dnx dotnet-inspect -y -- member "StackExchange.Redis.SimpleResultBox" "explicit:StackExchange.Redis.IResultBox.ActivateContinuations" --package StackExchange.Redis@3.0.17 --all -S "Decompiled Source"
```

```csharp
void StackExchange.Redis.IResultBox.ActivateContinuations()
{
    lock (this) { Monitor.PulseAll(this); }
}
```

## Ask
Fix the RTS target-resolution/shell so explicit-interface-impl corpus rows resolve to the product's explicit selector, removing this false-`Invalid` family from the benchmark. Refs: sibling #2963/#2964; #3052; discovery #3079.

PR #3222, opened 2026-07-26
RTS: reconstruct single-member external explicit-interface methods (#3112, Increment 1)

richlander:
- Advances #3112 (Increment 1 of a stacked pair)
- Changes RTS explicit-interface **method** reconstruction for **external/BCL**
  interfaces (harness-only; product output unchanged)
- Head: `2e0fcb65`

## Change

**Conclusion:** **PASS** — PR #3199 taught RTS to reconstruct **same-assembly**
explicit-interface methods as real explicit implementations (ContextFail → Exact),
but explicit implementations of **external/referenced** interfaces (e.g.
`System.Collections.IEnumerable.GetEnumerator`) still sanitized their dotted
metadata name (`System_Collections_IEnumerable_GetEnumerator`) and landed at the
`ContextFail` floor. This change reconstructs the **single-member external** case as
a real explicit-interface implementation and names the external interface in the
containing type's base list, so the recompiled method carries the correct dotted
metadata name and round-trips **Exact**. Every uncertain shape declines atomically to
the pre-fix sanitized shape, so no `RecompileFail` regression is introduced.

Before:

```text
compileBackStatus == ContextFail, detail == method-not-found
// class emits: public IEnumerator System_Collections_IEnumerable_GetEnumerator() { ... }
// base list:   class Seq { }              (interface not named)
```

After:

```text
STATUS = Exact
// class emits: IEnumerator System.Collections.IEnumerable.GetEnumerator() { ... }
// base list:   class Seq : System.Collections.IEnumerable { }
```

## Scope and safety

- **Root cause:** For external explicit-interface targets, `ExplicitInterfaceMemberName`
  resolves the interface via `TypeProducer.FindType`, which scans only the
  target-assembly `TypeDefinitions`; an external interface returns null so the target
  kept the plain sanitized shape. The base-list builder (`InterfaceSignatures`) and the
  same-assembly declaration helper (`AddExplicitInterfaceMethodDeclaration`) both gate
  on `Kind == TypeDefinition`/`MethodDefinition`, so an external interface was never
  named and never reconstructed as explicit.
- **Fix shape (harness orchestration only):** When a target has a `MethodImpl` whose
  declaration is a `MemberReference` with a **non-generic `TypeReference`** parent, and
  the external interface's **full transitive required surface is exactly one** abstract
  instance method matching the target by name and generic arity, the harness (a)
  reconstructs the target as an explicit implementation spelled
  `<InterfaceDisplayName>.<Member>` and (b) adds `<InterfaceDisplayName>` to the base
  list. Both use the same display spelling so they agree. Signature match is guaranteed
  by the `MethodImpl` linkage — no arbitrary signature translation is performed.
- **Product impact:** None. No product source/printer/API changes; all new code is under
  `tools/DecompilerHarness/ReturnToSenderTypePlanner.cs`.
- **Scope boundary (atomic + safe fallback):** Engagement is fully gated; the target is
  reconstructed as explicit **only** when the emitted C# is guaranteed to compile.
  Otherwise it reverts to the plain sanitized shape, preserving the pre-fix
  `ContextFail` rather than introducing a `RecompileFail`. It declines for:
  - **Generic / `TypeSpecification` parents** (e.g. `IEnumerable<int>`) — out of scope.
  - **Multi-member interfaces** (surface ≠ 1, e.g. `IConvertible`) — a partial surface
    would be CS0535; deferred to Increment 2.
  - **Interfaces exposing properties/events**, or members that are
    **non-abstract / static / operator** — cannot be supplied as a single bodyless
    method declaration.
  - **Unresolvable assemblies** or **metadata-name mismatch** — cannot verify identity.
- **Performance:** external interface surface reads open and scan the defining assembly
  (often CoreLib). The same interface recurs across many targets in one run, so resolver
  construction and surface results are memoized per `(target assembly, interface)`,
  including negative results, to avoid repeated CoreLib scans.

## Evidence

| Check | Result |
| --- | --- |
| Harness/test build (`ILInspector.Decompiler.Tests -c Release`) | Pass (0 errors, 0 warnings) |
| Full RTS class (`ReturnToSenderPrototypeTests`) | 163 passed, 0 failed |
| `*External*` focused tests | 16 passed (incl. 3 new) |
| Broad EVIL-corpus A/B | Not run in this environment (pinned corpus artifact unavailable on this Windows host); root cause + fix validated by focused round-trip tests |

New focused tests (`src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs`):

- `CompileBackTargets_RoundTripsExternalSingleMemberExplicitInterfaceMethod` —
  `Seq : IEnumerable`, target `System.Collections.IEnumerable.GetEnumerator` → asserts
  **Exact**, base list + explicit method present, sanitized name absent.
- `CompileBackTargets_MultiMemberExternalExplicitInterfaceFallsBackWithoutRecompileFail` —
  `Convertible : IConvertible` (fully implemented) → **not** `RecompileFail`, sanitized
  name present (declines).
- `CompileBackTargets_GenericExternalExplicitInterfaceFallsBackWithoutRecompileFail` —
  `IntSeq : IEnumerable<int>` → **not** `RecompileFail`, sanitized name present (declines).

## Frontiers and non-actions

| Item | Status | Reason |
| --- | --- | --- |
| Single-member external interfaces (e.g. `IEnumerable`, `ICloneable`, `IDisposable`) | Fixed | Reconstruct as explicit + name in base list → Exact |
| Multi-member external interfaces (`IConvertible`, `ICustomTypeDescriptor`, `ICollection`) | Deferred | Increment 2 (stacked follow-up): synthesize the full external surface as throw-stub explicit impls |
| Generic external interfaces (`IEnumerable<T>`) | Not changed | Out of scope; keeps prior plain shape (unchanged Invalid, not worse) |
| External interfaces exposing properties/events, or non-abstract/static/operator members | Fall back | Cannot supply as a single bodyless method declaration |

## Stacked-PR structure

This is **PR A** (base `main`). **PR B** (Increment 2, multi-member external interfaces)
will be branched off this branch and stacked; GitHub retargets it to `main` when this
merges.

## Validation

```bash
dotnet build src/ILInspector.Decompiler.Tests/ILInspector.Decompiler.Tests.csproj -c Release
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -method "*External*"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -1402,6 +1402,53 @@ public class ReturnToSenderPrototypeTests
         }
     }
 
+    // Regression for #3112 review (unrepresentable interface name): a hand-authored external
+    // interface can live in a namespace whose segment is a compiler-unspeakable name (`<Bad>`)
+    // — legal in metadata but not a legal C# identifier. Clean() sanitizes it lossily to a
+    // DIFFERENT identifier (`__Bad_`), so the reconstruction would emit `using __Bad_;` /
+    // `__Bad_.IProbe.M()` referencing a type that does not exist (CS0246 = RecompileFail). The
+    // gate must recognize the name cannot round-trip and decline to the sanitized ContextFail
+    // floor. Uses two IL assemblies (an external contract plus the target) resolved as
+    // siblings.
+    [Fact]
+    public void CompileBackTargets_ExternalExplicitInterfaceDeclinesWhenNameIsUnrepresentable()
+    {
+        var ilasm = TryLocateIlasm();
+        if (ilasm is null)
+        {
+            Assert.Skip("ilasm not available; skipping hand-authored IL unrepresentable-name regression.");
+            return;
+        }
+
+        var directory = Path.Combine(Path.GetTempPath(), $"return-to-sender-{Guid.NewGuid():N}");
+        AssembleIlFixture(ilasm, UnrepresentableContractsIl, directory, "GeneratedContracts");
+        var assemblyPath = AssembleIlFixture(ilasm, UnrepresentableFixtureIl, directory, "badfixture");
+        try
+        {
+            var target = new ReturnToSender.RequestedTarget("N.Seq", "<Bad>.IProbe.M", 0);
+
+            // All would reconstruct the interface spelling from the lossily-sanitized display
+            // name (`__Bad_.IProbe`), which names no type. The gate must decline rather than
+            // emit a new RecompileFail (CS0246). The name-representability guard is what
+            // catches this; the raw metadata name is not identifier-like.
+            var all = Assert.Single(
+                ReturnToSender.CompileBackTargets(assemblyPath, [target], RoundTripScope.All));
+            Assert.True(
+                all.Status != FidelityCheck.CompileBackStatus.RecompileFail,
+                $"all {all.Status}: {all.Detail}{Environment.NewLine}{all.Source}");
+        }
+        finally
+        {
+            try
+            {
+                Directory.Delete(directory, recursive: true);
+            }
+            catch (IOException)
+            {
+            }
+        }
+    }
+
     [Fact]
     public void CompileBackTargets_MultiMemberExternalExplicitInterfaceFallsBackWithoutRecompileFail()
     {
@@ -7618,6 +7665,48 @@ public class ReturnToSenderPrototypeTests
         .class public auto ansi sealed beforefieldinit N.'class'
             extends [System.Runtime]System.Object
         {
+          .method public hidebysig specialname rtspecialname instance void .ctor() cil managed
+          {
+            ldarg.0
+            call instance void [System.Runtime]System.Object::.ctor()
+            ret
+          }
+        }
+        """;
+
+    // External contract for the unrepresentable-name regression: an interface whose namespace
+    // segment is a compiler-unspeakable name (`<Bad>`) — legal in metadata, not a legal C#
+    // identifier — so Clean() sanitizes it lossily to a different name (`__Bad_`).
+    const string UnrepresentableContractsIl = """
+        .assembly extern System.Runtime { .ver 0:0:0:0 }
+        .assembly GeneratedContracts { }
+        .module GeneratedContracts.dll
+
+        .class interface public abstract auto ansi '<Bad>'.IProbe
+        {
+          .method public hidebysig newslot abstract virtual instance void M() cil managed {}
+        }
+        """;
+
+    // Target for the unrepresentable-name regression: N.Seq explicitly implements the external
+    // `<Bad>.IProbe`. The reconstruction would emit the sanitized `__Bad_.IProbe`, which names
+    // no real type (CS0246) — the gate must decline to the sanitized ContextFail floor instead.
+    const string UnrepresentableFixtureIl = """
+        .assembly extern System.Runtime { .ver 0:0:0:0 }
+        .assembly extern GeneratedContracts { }
+        .assembly badfixture { }
+        .module badfixture.dll
+
+        .class public auto ansi sealed beforefieldinit N.Seq
+            extends [System.Runtime]System.Object
+            implements [GeneratedContracts]'<Bad>'.IProbe
+        {
+          .method private final hidebysig newslot virtual
+              instance void '<Bad>.IProbe.M'() cil managed
+          {
+            .override [GeneratedContracts]'<Bad>'.IProbe::M
+            ret
+          }
           .method public hidebysig specialname rtspecialname instance void .ctor() cil managed
           {
             ldarg.0
--- a/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ReturnToSenderPrototypeTests.cs
@@ -1576,6 +1576,57 @@ public class ReturnToSenderPrototypeTests
     }
 
     [Fact]
+    public void CompileBackTargets_ExternalExplicitInterfaceWithInternalMemberFallsBackWithoutRecompileFail()
+    {
+        // Regression guard: a PUBLIC interface may still declare a NON-public member (C# 8+
+        // allows explicit accessibility on interface members). The interface type passes the
+        // IsPubliclyAccessible gate, but the reconstructed assembly
+        // ("return-to-sender-source-oracle") is not granted InternalsVisibleTo, so it cannot
+        // name the internal member even though the target implements it via IVT to its own name.
+        // Engaging would emit `void RtsVis.IProbe.M()` against a member the recompile cannot see
+        // (CS0122 = RecompileFail). The gate must decline any non-public required method and keep
+        // the sanitized ContextFail floor.
+        var fixtureDir = Path.Combine(Path.GetTempPath(), $"return-to-sender-{Guid.NewGuid():N}");
+        var contractsPath = CompileFixture(
+            """
+            using System.Runtime.CompilerServices;
+            [assembly: InternalsVisibleTo("fixture")]
+            namespace RtsVis { public interface IProbe { internal abstract void M(); } }
+            """,
+            directory: fixtureDir,
+            assemblyName: "RtsVisContracts");
+        var assemblyPath = CompileFixture(
+            """
+            public sealed class VisImpl : RtsVis.IProbe
+            {
+                void RtsVis.IProbe.M() { }
+            }
+            """,
+            directory: fixtureDir,
+            assemblyName: "fixture",
+            additionalReferences: [MetadataReference.CreateFromFile(contractsPath)]);
+        try
+        {
+            var result = Assert.Single(ReturnToSender.CompileBackTargets(
+                assemblyPath,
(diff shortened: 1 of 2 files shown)


richlander:
## Adversarial review reconciliation — two CLEAN fixed-head verdicts @ `b746b440`

Both reviews are CLEAN at the **same** head `b746b440`, from two different model families (GPT `gpt-5.6-sol` and Gemini `gemini-3.1-pro-preview`; not Claude, per policy). Each finding below was reproduced empirically (ilasm + csc, and PEReader for metadata bytes) before and after the fix.

### Finding arc (all reproduced, all fixed)

| # | Reviewer(s) | Finding | Repro | Fix |
|---|---|---|---|---|
| 1 | GPT | Unrooted external interface name — a same-namespace sibling can shadow the spelling → CS0426/CS0538 RecompileFail | ✅ | Scope-aware first-segment shadow decline |
| 2 | GPT + Gemini (converged) | Over-decline of legitimate non-leading segment matches | ✅ | Collision-aware using-collapser + `{Collections, IEnumerable}` theory |
| 3 | GPT | Escaped-identifier shadow check used escaped `DisplayFullName` | ✅ | Pass raw `MetadataFullName` to the shadow check (`7406e83f`) |
| 4 | GPT | Unspeakable interface **type** name (`<Bad>`) lossily sanitized to `__Bad_` → CS0246 | ✅ | `ExternalInterfaceNameIsRepresentable` per-segment guard (`55554228`) |
| 5 | GPT | Unspeakable **member** name (`<Bad>`) → CS0539 | ✅ | Member-name representability guard (`a51f2258`) |
| 6 | GPT (missed by Gemini @ a51f2258) | `IsIdentifierLike` accepts Unicode **format (Cf)** chars (U+200C) that Roslyn strips when binding → CS0539/CS0246 | ✅ csc: `M\u200C` binds to `M` | `MetadataIdentifierRoundTrips` = identifier-like + no Cf rune (`cb65c7f7`) |
| 7 | GPT + Gemini (converged) | The Cf fix also required NFC (`IsNormalized(FormC)`), but Roslyn does **not** normalize identifiers — decomposed `e`+U+0301 round-trips Exact, so this over-declined a compiler-producible shape | ✅ PEReader: metadata stores decomposed bytes `65 CC 81` (non-NFC); base ContextFail / pre-NFC Exact / with-NFC ContextFail / fixed Exact | Removed the NFC clause; kept Cf rejection (`b746b440`) |

Concurrent `32bd5095` confirmed `[CompilerFeatureRequired]`-guarded interfaces remain on the ContextFail floor.

### Final predicate
`MetadataIdentifierRoundTrips(name)` = `CSharpIdentifier.IsIdentifierLike(name)` **AND** no rune of `UnicodeCategory.Format`. Applied to both emitted identifiers (member name + each interface-name segment). This is the correct and complete characterization of C#'s identifier-binding transform (ordinal match on emitted UTF-8, minus Cf stripping; no normalization), confirmed independently by both reviewers.

### Evidence
- Full RTS suite (`ReturnToSenderPrototypeTests`): **182 pass / 0 fail** at `b746b440`.
- New regressions: two ilasm-gated Cf declines (member + namespace, assert `!= RecompileFail`) and one positive decomposed-identifier test (`...KeepsExactWhenMemberNameIsDecomposed`, asserts `Exact` + no floor).
- Change is harness-only (`ReturnToSenderTypePlanner.cs` + tests); a conformant C# compiler cannot author the declined shapes, so verdict-based corpus baseline is unchanged.

GPT verdict: **CLEAN**. Gemini verdict: **CLEAN**. Both @ `b746b440`.

richlander:
Ready to merge

Both adversarial reviews (GPT + Gemini) are CLEAN at `b746b440`, the merge-path `test (ubuntu-26.04, linux-x64)` job is green, and `changes` passed. Reconciliation is in the comment above.

richlander:
## Worked example — where the failure actually is (Original → Before → After → Fully raised)

> **Edited** to correct the declined-floor shape: the floor drops the interface from the base list and demotes the member to a plain flattened-name method (it does **not** keep `: Good.IProbe`), and the fixtures are `void …()`, not `int …(int v)`. The verdicts and conclusions are unchanged.

The table above indexes the findings but hides the crux: *which emitted line binds to the wrong thing, and why*. Here are the two subtlest findings in the decompiler-PR pattern. The whole gate reconstructs a single-member external explicit-interface method; the failure is always in **one emitted identifier** whose C# binding diverges from its metadata spelling.

**What "declined to the floor" means.** Declining is atomic (`ReturnToSenderTypePlanner.cs:2614,2629-2655`): the gate does **not** insert a skeleton. It keeps the real reconstructed body (`CompileBackStubBodyKind.TargetBody`, `RoundTripBodyPolicy.Full`) but strips the member's *identity* — it drops the interface from the base list (`ExternalInterfaces = []`) and demotes the member from an explicit impl to a plain method whose name is the sanitized MethodDef name (dots → underscores, `Good.IProbe.M◌` → `Good_IProbe_M`). That plain shape compiles, but the fidelity harness keys on the original explicit-member identity, which no longer exists under that name in the recompiled assembly → `ContextFail` "method-not-found" (`FidelityCheck.cs:1356`) — the safe floor, strictly better than a `RecompileFail`.

Legend: the reconstructed type is `N.Seq`; `[Ext]` is a separate external assembly holding the interface; `‹ZWNJ›` = U+200C, `◌́` = U+0301 (combining acute).

---

### Example A — format-character member name (finding 6): a *false Exact* that is really a RecompileFail

**Original (metadata anchor — hand-authored IL).** The interface member name contains U+200C:

```il
.class interface public abstract auto ansi [Ext]Good.IProbe
{
  .method public hidebysig newslot abstract virtual instance void 'M‹ZWNJ›'() cil managed {}
}
.class public sealed N.Seq implements [Ext]Good.IProbe
{
  .method private final hidebysig newslot virtual instance void 'Good.IProbe.M‹ZWNJ›'() cil managed
  { .override [Ext]Good.IProbe::'M‹ZWNJ›'  ret }
}
```

**Before (emitted C# at the pre-guard head + verdict).** The gate engages and spells the explicit qualifier verbatim:

```csharp
class Seq : Good.IProbe
{
    void Good.IProbe.M‹ZWNJ›() { }
    //             ^^^^^^^^^^^ Roslyn STRIPS U+200C when binding this identifier,
    //                         so it binds to `Good.IProbe.M` — but the interface
    //                         member is `M‹ZWNJ›`, which no longer matches.
}
```

- Verdict: **RecompileFail** — `CS0539: 'Seq.M()' in explicit interface declaration is not found`
- Valid: **False** · this is the trap: it *looks* like a clean explicit impl, but binding silently renamed the qualifier.

**After (emitted C# at `b746b440` + verdict).** `MetadataIdentifierRoundTrips` sees the Cf rune, so the gate **declines**: the interface leaves the base list and the member is demoted to a plain method (real body kept):

```csharp
class Seq                       // interface dropped from the base list
{
    void Good_IProbe_M() { }    // demoted plain method (dots → underscores); real body kept
}
```

- Verdict: **ContextFail** ("method-not-found" floor) — compiles, but the original explicit member `Good.IProbe.M‹ZWNJ›` no longer exists to compare, so no `RecompileFail` is introduced.
- Valid: **True**

**Fully raised.** N/A — a conformant C# compiler *cannot* author a member whose name survives only in metadata. The ContextFail floor is the correct endpoint; there is no better C# to reach.

---

### Example B — decomposed (non-NFC) member name (finding 7): the mirror image — a *real Exact* the guard wrongly dropped

**Original (metadata anchor).** A developer legitimately writes `é` as `e` + U+0301. Roslyn stores it **decomposed** in metadata (bytes `65 CC 81`, `IsNormalized(FormC)=false`) — it does not normalize:

```il
.class interface public abstract auto ansi [Ext]Good.IProbe
{
  .method public hidebysig newslot abstract virtual instance void 'e◌́'() cil managed {}
}
```

**Before (emitted C# at `cb65c7f7` — the over-declining head + verdict).** My round-6 guard *also* required NFC, so it treated this legal name as un-round-trippable and declined it to the floor — dropping a real, reachable explicit impl:

```csharp
class Seq                         // interface dropped from the base list
{
    void Good_IProbe_e◌́() { }    // WRONG: demoted to a plain method though the explicit impl was reachable
}
```

- Verdict: **ContextFail** (floor) — but the true, higher-fidelity `Exact` shape was reachable. This is an **over-decline** (regression vs `a51f2258`, which produced Exact).

**After (emitted C# at `b746b440` + verdict).** NFC clause removed; the decomposed name has no Cf rune, so it round-trips. The gate emits it verbatim and Roslyn ordinal-matches it against the metadata:

```csharp
class Seq : Good.IProbe
{
    void Good.IProbe.e◌́() { }   // emitted verbatim; ordinal match → binds exactly
}
```

- Verdict: **Exact**
- Valid: **True** · Correct: **True** · IL fidelity: **True**

**Fully raised.** The After **is** the fully raised state — this is the correct, opcode-faithful explicit implementation.

---

### The invariant these two pin

The gate emits two identifiers (the interface-name segments and the member name). Each must survive C#'s identifier-binding transform, which is exactly: **ordinal match on the emitted UTF-8, after stripping Unicode format (Cf) characters — no normalization.** So the round-trip predicate is precisely `IsIdentifierLike && no Cf rune`. Example A is why Cf must be rejected; Example B is why NFC must **not** be required. Anything that fails the predicate declines atomically to the sanitized ContextFail floor — never worse than base.

PR #3224, opened 2026-07-26
Detect VB.NET implicit object.Finalize overrides as finalizers (#3196)

### LINKED ISSUE #3196: Finalizer detection misses VB implicit object.Finalize overrides (no MethodImpl) → uncompilable output
## Summary

Follow-up from adversarial review of #3168 / #3157. The finalizer detection path recognizes finalizers by their explicit `.override` **MethodImpl** row (the shape Roslyn/C# emits). VB.NET emits a valid `object.Finalize` override with **no MethodImpl row** — it overrides the base slot implicitly. Such a member is currently **not** classified as a finalizer.

## Impact

For a VB-compiled assembly, the member surfaces as a plain `override void Finalize()` (or similar) rather than the `~Type()` destructor spelling. Because C# has no `Finalize` method you can `override`, the rendered declaration is **uncompilable** (CS0249: "Do not override object.Finalize; override the destructor syntax instead").

## Why it's deferred

A correct fix requires resolving the overridden **base slot chain** up to `System.Object.Finalize`, which:

- Is frequently **cross-assembly** and not resolvable from metadata alone on the SRM-only product path.
- Re-opens the **round-1 false-positive** surface: a derived type overriding a *custom* `new virtual Finalize` (not `object.Finalize`) must **not** be spelled `~Type()`.

So detecting VB finalizers safely needs base-slot resolution with a robust `System.Object` anchor and negative-case guards, which is a larger change than the #3168/#3186 line of work.

## Acceptance

- A VB-authored `Protected Overrides Sub Finalize()` is classified as a finalizer and spelled `~Type()`.
- A derived C#/IL type overriding a *custom* `virtual Finalize` slot (not `object.Finalize`) is **not** spelled as a destructor (no regression of the round-1 false-positive fix).
- Cross-assembly / SRM-unresolvable base chains fail visibly (no success-shaped wrong output) rather than guessing.

## References

- #3157, #3168 (detection via MethodImpl `.override`)
- #3186 (dedicated `Kind = "finalizer"`)
- Both round-2 reviewers (GPT-5.6 High, Gemini 3.1 Pro "known limitation") independently flagged this.

richlander:
Fixes #3196.

## Conclusion

A VB.NET `Protected Overrides Sub Finalize()` is now classified as a finalizer (`Kind="finalizer"`, `IsFinalizer=true`), so it renders as the **compilable** C# `void Finalize()` (retaining the explicit `base.Finalize()`) instead of the previous **uncompilable** `override void Finalize()` (CS0249). The fully-raised `~Type()` destructor spelling is the intended endpoint and is tracked as follow-up in #3232. No change to any other classification.

## Why the old detection missed it

The VB.NET compiler emits the finalizer as a **virtual, reuse-slot, parameterless `void Finalize()` that carries NO `.override` MethodImpl** (empirically confirmed via SRM on a compiled VB fixture: base=`System.Object` TypeReference, 0 MethodImpls, `Finalize` Virtual + not-NewSlot + Family). The existing detection keys on the `.override System.Object::Finalize` MethodImpl that Roslyn/C# emits, so it never fired for the VB shape.

## Change

`ApiSurfaceExtractor.IsImplicitObjectFinalizeOverride` (SRM-only, NativeAOT-friendly, no inspected-assembly loading) OR-ed into the existing MethodImpl path:

1. **Shape gate** — name `Finalize`; Virtual && !NewSlot && !Static && !Abstract; non-generic; instance `void()` signature (a modopt/by-ref return or any parameter/non-void return rejects, so a name-only collision cannot masquerade as a finalizer).
2. **Slot-root proof** — walk the base-type chain: a base reference resolving to the strong-name-anchored `System.Object` confirms; an in-assembly base introducing its own `new virtual void Finalize()` slot is a custom slot and rejects; a base that leaves the assembly without resolving to `System.Object` (including a generic/`TypeSpecification` base) is **conservatively rejected**, never guessed. Visited-set + depth cap guard cyclic/degenerate metadata.

`IsFinalizerMethod` (source anchoring in `PdbContext.EnumerateMemberSources`) gains the same fallback, so VB finalizer source lines anchor from metadata identity.

## Raised-source evidence

Per `docs/templates/decompiler-pr.md`. Fixture: compiled VB `Handle` (`/tmp/vbfin`), member `Handle.Finalize`.

### Original source

Authoritative original is VB (no C# original exists); acquired from the fixture source. The C#-equivalent intent is a `~Handle()` destructor.

```vbnet
Public Class Handle
    Private _n As Integer
    Protected Overrides Sub Finalize()
        _n = 0
        MyBase.Finalize()
    End Sub
End Class
```

### Before

`dotnet-inspect member Handle -m Finalize --library <vbfin.dll> --all -S "Decompiled Source"`

```csharp
protected override void Finalize()
{
    _n = 0;
    base.Finalize();
}
```

- Valid: False
- Correct: False
- IL fidelity: not currently checkable
- Taste applied: None
- Commit: 08ab86b1

C# forbids overriding `object.Finalize` directly (CS0249), so this output does not compile and cannot be a correct decompilation.

### After

`dotnet-inspect member Handle -m Finalize --library <vbfin.dll> --all -S "Decompiled Source"`

```csharp
void Finalize()
{
    _n = 0;
    base.Finalize();
}
```

- Valid: True
- Correct: True
- IL fidelity: False
- Taste applied: None
- Commit: 6cadf35c

Compiles (CS0465 warning only). The body faithfully preserves the field write and the explicit `base.Finalize()` call. The declaration spells a plain method rather than the finalizer slot, so recompiling it produces a name-hiding method, not the original virtual reuse-slot override — full IL fidelity requires the `~Type()` form (#3232). This is the deliberate honest fallback (`SuppressDestructorSyntax = member.IsFinalizer && !bodyIsDestructor`): keeping the literal `void Finalize()` with its explicit base call, rather than dropping it, until VB finalizer bodies raise to the destructor scaffold.

### Fully raised

```csharp
~Handle()
{
    _n = 0;
}
```

- Required tracking issue: #3232 — extend the decompiler's destructor-scaffold recognition (`bodyIsDestructor`) to the VB finalizer body shape so a VB implicit finalizer raises to `~Type()`, folding the base call into the compiler-implicit `finally`.

## Evidence

- **Real-VB canary (compiler-produced):** manual CLI on compiled VB fixtures — single finalizer (`Handle`) and the `VbBase`/`VbDerived` chain (both override) all render as compilable `void Finalize()` (was the uncompilable `override void Finalize()`). (ilasm and a VB compiler are unavailable in the test box; ilasm-based tests only probe `PATH` and skip, so they are not a reliable CI signal here.)
- **Real-C# negative (compiler-produced):** `CustomVirtualFinalizeSlot_IsNotClassifiedAsFinalizer` — `new virtual void Finalize()` + its override (CS0465) → neither is a finalizer.
- **Synthetic metadata fixtures (run everywhere), `ImplicitFinalizerDetectionTests`:** positive VB shape; walk-past-reuse-slot chain; negatives — custom new-virtual slot, `Finalize(int)`, `int Finalize()`, non-object cross-assembly base, new-slot `Finalize`.
- Full `ILInspector.Metadata.Tests` (849) and decompiler `MemberBodyProducerExpressionBodyTests` (13) green; full solution builds clean. The one local CLI failure (`PlatformResolverTests.GetPacksDirectory_ReturnsValidPath`) is a pre-existing environment-specific packs-path assertion, untouched by this change.

## Known limitation

A generic (`TypeSpecification`) base type is conservatively rejected — a possible miss, never a false positive. Documented in code; decoding the TypeSpec's underlying def/ref is possible future work.

## Review

Non-trivial (new metadata heuristic) → requires two-model adversarial review (non-Claude). Attack surface: custom-slot false positives (esp. cross-assembly), generic bases, structs/interfaces named `Finalize`, malformed/cyclic metadata, signature-decode edge cases (modopt/by-ref return), interaction with the existing MethodImpl path, and PdbContext source-anchoring consistency.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/MemberBodyProducerExpressionBodyTests.cs
+++ b/src/ILInspector.Decompiler.Tests/MemberBodyProducerExpressionBodyTests.cs
@@ -322,6 +322,27 @@ public class MemberBodyProducerExpressionBodyTests
         Assert.False(member.IsFinalizer);
     }
 
+    [Fact]
+    public void CustomVirtualFinalizeSlot_IsNotClassifiedAsFinalizer()
+    {
+        // Close negative (compiler-produced): a custom `new virtual void Finalize()`
+        // slot and an override of it are NOT the object.Finalize destructor. The C#
+        // compiler emits CS0465 (a warning) but the shape is legal metadata, and it
+        // must never fold to `~Type()`.
+        using var assembly = Compile("""
+            public class CustomBase { public virtual void Finalize() { } }
+            public class CustomDerived : CustomBase { public override void Finalize() { } }
+            """);
+
+        var baseMember = ExtractMember(assembly.Path, "CustomBase", "Finalize");
+        Assert.False(baseMember.IsFinalizer);
+        Assert.NotEqual("finalizer", baseMember.Kind);
+
+        var derivedMember = ExtractMember(assembly.Path, "CustomDerived", "Finalize");
+        Assert.False(derivedMember.IsFinalizer);
+        Assert.NotEqual("finalizer", derivedMember.Kind);
+    }
+
     static ApiMember ExtractMember(string path, string typeName, string memberName)
     {
         using var pe = new PEReader(File.OpenRead(path));
--- a/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
+++ b/src/ILInspector.Metadata/ApiSurfaceExtractor.cs
@@ -188,26 +188,27 @@ public static class ApiSurfaceExtractor
                 var isOverride = isVirtual && !isNewSlot && !isExplicitInterfaceImplementation;
 
                 // A class finalizer is the `object.Finalize` override the C#
-                // `~Type()` destructor compiles to. Roslyn emits it with an
-                // explicit `.override` MethodImpl targeting
-                // `System.Object::Finalize`, so we detect it by that target
-                // rather than by name/signature shape. Keying on the overridden
-                // slot (not the name `Finalize`, the reused virtual slot, or the
-                // decoded signature) excludes the false positives a shape
-                // heuristic admits: an implicit generic `Finalize<T>()`, an
-                // override of an unrelated base/interface `Finalize()` slot, and
-                // an explicit `IFoo.Finalize()` implementation (whose MethodImpl
-                // targets the interface, not object). A method whose signature
-                // failed to decode is likewise judged by its MethodImpl target
-                // alone, so a degraded decode cannot masquerade as a finalizer.
-                // A finalizer is never generic, so we still reject a method that
-                // explicitly `.override`s object.Finalize while declaring its own
-                // type parameters — rendering it `~Type()` would erase `<T>`.
-                // VB-style implicit `object.Finalize` overrides carry no
-                // MethodImpl and fall back to the literal `void Finalize()`.
+                // `~Type()` destructor compiles to. It is detected by the
+                // overridden slot (not by name/signature shape), which excludes
+                // the false positives a shape heuristic admits: an implicit
+                // generic `Finalize<T>()`, an override of an unrelated
+                // base/interface `Finalize()` slot, and an explicit
+                // `IFoo.Finalize()` implementation. There are two slot-anchored
+                // shapes:
+                //   * Roslyn (C#) emits an explicit `.override` MethodImpl
+                //     targeting `System.Object::Finalize`; `objectFinalizeOverrides`
+                //     carries those.
+                //   * The VB.NET compiler emits `Protected Overrides Sub Finalize()`
+                //     with NO MethodImpl — it reuses the inherited object.Finalize
+                //     slot implicitly; `IsImplicitObjectFinalizeOverride` proves
+                //     that slot roots at `System.Object` over metadata alone.
+                // A finalizer is never generic, so a method that overrides
+                // object.Finalize while declaring its own type parameters is still
+                // rejected — rendering it `~Type()` would erase `<T>`.
                 var isFinalizer = apiType.Kind == "class"
                     && method.GetGenericParameters().Count == 0
-                    && objectFinalizeOverrides.Contains(methodHandle);
+                    && (objectFinalizeOverrides.Contains(methodHandle)
+                        || IsImplicitObjectFinalizeOverride(reader, typeDefHandle, method));
 
                 var member = new ApiMember
                 {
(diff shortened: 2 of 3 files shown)


richlander:
## Adversarial review — round 1 reconciliation

Two-model review at frozen head `d894c601` (roster: non-Claude), isolated worktrees, scratch under `/tmp`, no git mutations in review trees.

**Reviewer A — gpt-5.6-sol** raised three findings (all reproduced):

1. **Signature convention gap (Medium/High)** — `HasVoidNullaryInstanceSignature` accepted vararg/generic-convention/explicit-this signatures (e.g. `[0x25,0x00,0x01]` = `HASTHIS|VARARG`), which cannot bind `object.Finalize`'s default-convention slot. **Fixed** in `6cadf35c`: now requires default managed calling convention, HASTHIS with no explicit `this`, and non-generic. New negative test `VarargFinalize_IsNotClassifiedAsFinalizer`.
2. **Truncated blob throws (Medium/High)** — a malformed/truncated `Finalize` signature (`[0x20]`) reached `ReadCompressedInteger()` and threw `BadImageFormatException`. Reviewer B independently traced that both call paths already degrade gracefully upstream (extraction records a rejected-type failure; the PDB path forces evaluation inside an inspection-failure catch), so this is not a crash — but a single malformed method would still discard the whole type's surface. **Fixed** in `6cadf35c` with a local `try/catch (BadImageFormatException) => false` guard (defense-in-depth; keeps one bad method from dropping the type).
3. **In-assembly `System.Object` NewSlot ordering (Medium/High)** — the walk tested `DeclaresNewVirtualFinalize` before recognizing the in-assembly `System.Object` root; since the genuine `object.Finalize` is itself a NewSlot virtual, a corelib-self inspection would wrongly reject a derived finalizer. Low real-world impact (corelib is C#, so its finalizers take the MethodImpl path), but the logic was wrong. **Fixed** in `6cadf35c`: recognize the root before the custom-slot rejection. New positive test `InAssemblyObjectRoot_DerivedFinalize_IsClassifiedAsFinalizer`. Both new tests fail on the pre-fix code.

**Reviewer B — gemini-3.1-pro-preview**: no issues found. Independently reproduced and verified signature decode (modopt/by-ref/vararg/generic/truncated), base-chain walking, cyclic/depth guards, the reverse-SHA1 strong-name token computation, and test soundness.

Metadata suite **851** green; decompiler expression-body **13** green; full solution builds clean. Re-review of the fixed head `6cadf35c` is in progress.

richlander:
## Adversarial review — round 2 (fixed head `6cadf35c`)

Both reviewers re-reviewed the fixed head in their isolated worktrees; both are **clean**.

- **gpt-5.6-sol:** "No significant issues found. Fixes are correct and complete." Verified: vararg + generic-convention rejected; truncated blob returns false via the guard; in-assembly `System.Object` derived finalizer classified; legitimate default-convention HASTHIS `object.Finalize` still accepted; metadata 9/9 + decompiler 13/13; worktree clean.
- **gemini-3.1-pro-preview:** "No blocking issues remain; the fixed head is clean." Independently reproduced (via SRM blob fixtures) that vararg parses as `VarArgs` and is rejected, `modopt` returns `OptionalModifier` and fails the exact-`Void` check, truncated blobs throw and are caught to `false`; confirmed `System.Object.Finalize` is natively `NewSlot` so the reorder is required; full metadata suite green, cross-assembly TypeReference path unregressed.

No new findings. Awaiting the `test` CI job to go green, then this is ready to merge.

richlander:
Ready to merge

- CI green at `6cadf35c`: `test` (6m42s) and `build-net10` pass; `changes` pass; path-gated jobs (pack, diff-smokes, markdownlint) correctly skipped.
- Two-model adversarial review complete: round-1 findings from gpt-5.6-sol fixed in `6cadf35c`; both gpt-5.6-sol and gemini-3.1-pro-preview re-reviewed the fixed head clean.
- Metadata suite 851 green; decompiler expression-body 13 green.

richlander:
**PR body updated — accuracy correction + raised-source evidence (no code change).**

Added the Original/Before/After/Fully-raised evidence block per `docs/templates/decompiler-pr.md`, and corrected an overclaim in the prior body: the actual raised output is the **compilable `void Finalize()`** (with explicit `base.Finalize()`), not `~Type()`. The `~Type()` destructor is the fully-raised endpoint and is now tracked in #3232 (extend `bodyIsDestructor` recognition to the VB finalizer body shape).

Net effect is unchanged and still the intended one: classification flips from the **uncompilable** `override void Finalize()` (CS0249) to the **compilable** `void Finalize()` (CS0465 warning), preserving the explicit base call. This edit is documentation-only; the frozen-head (`6cadf35c`) adversarial reviews remain valid. Still Ready to merge.

richlander:
Follow-up resolution: the "Fully raised" block above named #3232 (raise to `~Type()`) as the intended endpoint. On investigation with the permanent VB fixture (#3233), that endpoint is **wrong for VB** — VB emits the finalizer as a bare sequential body with no try/finally, so `~Type()` would fabricate an exception region and change semantics. #3232 is closed as by-design; the `void Finalize()` shipped here is the faithful endpoint. See the [#3232 closing comment](https://github.com/richlander/dotnet-inspect/issues/3232#issuecomment-5084882896).

PR #3225, opened 2026-07-26
Render boxed-this non-interface access as base/cast, not invalid (this).M() (#3213)

### LINKED ISSUE #3213: Decompiler erases boxing cast for non-interface boxed-this (e.g. (this).GetHashCode() vs box; callvirt ValueType::GetHashCode)
## Summary

The C# decompiler erases the boxing cast when a **value type boxes `this` (or another value) to reach a NON-interface member**, rendering `(this).M()` instead of the `((object)this).M()` / `((ValueType)this).M()` the boxing conversion spells. This is the non-interface sibling of #3201 (struct boxed-`this` to an *interface* member), which the #3201 interface-only gate deliberately declines — correctly leaving this case for a separate fix.

Unlike #3201 this is not merely invalid-vs-valid: for `object`/`ValueType` virtual members the erased form can bind to a **different target** or even recurse, so it is a correctness/fidelity bug, not just a spelling one.

## Witnesses

Test fixture (added as the #3201 negative guard) — `ILInspector.Decompiler.Tests.StructBoxedObjectReceiver.CallObjectToString`:

- Observed: `public string CallObjectToString() => (this).ToString();`
- Expected (fidelity): `((object)this).ToString();`
- IL boxes `this` and calls `object::ToString()`.

Real CoreLib witness — `System.Reflection.CustomAttributeNamedArgument.GetHashCode`:

- Observed: renders `(this).GetHashCode()`
- IL: `ldarg.0; ldobj; box; callvirt System.ValueType::GetHashCode()`
- `(this).GetHashCode()` on a struct binds to the struct's own `GetHashCode` (here, itself) rather than the boxed `ValueType.GetHashCode()` the IL calls — a potential wrong-target / self-recursion, so **Correct=False**, not only IL-fidelity=False.

## Where

`src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.Members.cs` — the `CallText` / `MethodGroupText` ordinary receiver path. The boxed-`this` shape is `Box { Operand: LoadIndirect { Address: LoadArgument{0,"this"} } }`; `ReceiverText`/`Operand` currently render the boxed value as `(this)`, dropping the `box` conversion's spelled cast.

## Approach (sketch)

When a receiver is a `Box` whose result type is `object`/`ValueType` (or any base type the callee is declared on) and the callee is NOT on the boxed value type, re-insert the `((BoxedToType)place)` cast — mirroring the interface arm from #3208 but for the boxing-to-base case. Reuse the confirmed shape/typing seams rather than string-sniffing. Pair with close negatives so an ordinary already-`object` receiver is not double-cast.

## Validation

- Add positive fixtures (`((object)this).ToString()`, `((ValueType)this).GetHashCode()`) and a real-witness render for `CustomAttributeNamedArgument.GetHashCode`.
- Confirm IL fidelity via the compile-back harness / `diff`, not `member`.
- Run `dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.ThisQualificationTests"`.

Context: surfaced by the #3208 adversarial review (GPT-5.6). Confirmed pre-existing at the #3208 base; not a regression.

richlander:
- Fixes #3213
- Changes boxed-`this` access to a non-interface member so it renders `base.M()` / `((T)this).M()` instead of the invalid, self-recursive `(this).M()`
- Card revision: `febb077a`

## Change

**Conclusion:** **PASS** — a struct that boxes `this` to reach a base member no longer renders self-recursive `(this).GetHashCode()`; the call kind selects `base.M()` (non-virtual) or `((T)this).M()` (virtual), and #3201's interface renders are unchanged.

### Benchmark target

Benchmark target: `System.Private.CoreLib.dll` @ runtime `11.0.0-preview.6.26359.118`

dotnet-inspect command:

```bash
dotnet-inspect member System.Reflection.CustomAttributeNamedArgument GetHashCode --all -S "Decompiled Source"
```

### Original source

SourceLink C# was not used; the raw IL is the authoritative anchor. A struct
override boxes `this` and calls the base slot non-virtually:

```il
IL_0000: ldarg.0
IL_0001: ldobj        System.Reflection.CustomAttributeNamedArgument
IL_0006: box          System.Reflection.CustomAttributeNamedArgument
IL_000B: call         System.ValueType::GetHashCode()
IL_0010: ret
```

### Before

```csharp
public override int GetHashCode() => (this).GetHashCode();
```

- Valid: True
- Correct: **False**
- IL fidelity: **False**
- Taste applied: None
- Commit: `abec2dd7`

`(this).GetHashCode()` binds, but re-dispatches virtually to this same override:
infinite self-recursion (StackOverflow). It is the #3127 trap inverted — it does
not even reach IL fidelity because it drops the `box; call ValueType::GetHashCode`
entirely.

### After

```csharp
public override int GetHashCode() => base.GetHashCode();
```

- Valid: True
- Correct: True
- IL fidelity: True
- Taste applied: None
- Commit: `febb077a`

The non-virtual `call` to a base-declared method through a boxed `this` is exactly
C#'s `base.GetHashCode()`, which recompiles to the original `ldobj; box; call
System.ValueType::GetHashCode` (verified by the `StructBaseHashCall` fixture, whose
`base.GetHashCode()` source lowers to that IL and round-trips `compile-back=Exact`).

### Fully raised

## Evidence

| Check | Baseline (`abec2dd7`) | Head (`febb077a`) |
| --- | --- | --- |
| Focused tests | `ThisQualificationTests`: 31 passed; `ThisQualificationDecisionTests`: 27 passed | `ThisQualificationTests`: 34 passed; `ThisQualificationDecisionTests`: 27 passed |
| Decompiler fast suite (`-trait- "Speed=Slow"`) | 3752, 1 failed | 3755, 1 failed |
| RoundTrip area (`-trait "Area=RoundTrip"`, incl. slow compile-back) | 464, 3 failed | 467, 3 failed |
| Reduced fixture fidelity | Pass (`compile-back=Exact`) | Pass (`compile-back=Exact`) |
| Real witness | `CustomAttributeNamedArgument::GetHashCode` → self-recursive `(this).GetHashCode()` | fixed → `base.GetHashCode()` |

Pre-existing Windows-local failures, confirmed identical at base and head (same
tests, same reason), unrelated to this diff. Baseline full-suite totals above are
the head run minus the 3 newly-added (passing) `ThisQualificationTests`; every
failing test was run and reproduced at base individually:

- Fast suite: `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString`
  (CRLF vs LF).
- RoundTrip slow gates (3):
  `GeneratedFixtureCatalogTests.ReturnToSenderRecordCatalog_CoversGeneratedRecordHelpers`,
  `…MinimalCompileBackRungs_ReportExpectedOutcomesByFixtureId`,
  `…CompilerLoweringFrontier_IsSelectableButNotInDefaultRun`
  (`compile-back=RecompileFail` / `target-body-fragment-missing` on
  `GeneratedFixtures.*` record/parameter-attribute fixtures — boxed *parameters*,
  not `this`, so outside this change's `LoadArgument{0,"this"}`-gated scope).

## Decompiler quality

**Conclusion:** **PASS** — pinned gate PASS with 0 pass bugs; detected lowering residue drops 251 → 246 as boxed-`this` base/cast methods are now raised.

### PR quick gate

Run: PR quick corpus, hash-stable 100 methods per assembly; 15 assemblies, 1,500
methods; validity/fidelity not run (`--corpus-fidelity-cap 0 --compile-cap 0`).

| --- | ---: | ---: | ---: |
| Detected lowering residue (-) | 16.73% | 16.40% | -0.33pp |
| Conditional-branch residue (-) | 3.07% | 2.87% | -0.20pp |
| Forward-merge stops (-) | 2.13% | 2.00% | -0.13pp |
| Pass bugs (-) | 0 | 0 | 0 |

> **Conclusion:** **PASS** — corpus sensor matched baseline tolerances; residue strictly improved.

## Approach

The value-type sibling of #3128/#3201. A struct reaching a member of a type it was
cast to must box `this`, so the IR receiver is
`Box{Operand: LoadIndirect{Address: LoadArgument{0,"this"}}}` — the shape #3201
already matched, but only fired when the callee was a **confirmed interface**. A
struct boxing to reach a **base class** member (`object`/`System.ValueType`) fell
through to the bare receiver path and rendered invalid `(this).M()`.

The explicit `box` over `this` is itself proof of a value-type upcast, so no
interface metadata is needed — the call kind fully disambiguates. This PR replaces
the interface-only `IsBoxedThisInterfaceReceiver` with a shape-only
`IsBoxedThisReceiver` and, in both `CallText` and `MethodGroupText`, mirrors the
existing bare-`this` arms:

- non-virtual `call` / `ldftn` to a cross-type base method → `base.M()` / `base.M`
- virtual `callvirt` / `ldvirtftn` to the cast target → `((T)this).M()` / `((T)this).M`

This subsumes #3201 (interface calls are virtual → `((I)this).M()`, unchanged) and
additionally fixes the base-class cases (#3213). A non-boxed implicit `this.ToString()`
(a `constrained. callvirt`, no `box`) is untouched.

Note: issue #3213 proposed `((ValueType)this).GetHashCode()` for the witness, but
the non-virtual `call` opcode makes `base.GetHashCode()` the correct spelling; an
explicit `((ValueType)this)` *virtual* cast instead binds to the `object`
GetHashCode slot and renders `((object)this)`.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class "ILInspector.Decompiler.Tests.ThisQualificationTests"
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -trait "Area=RoundTrip"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ThisQualificationTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ThisQualificationTests.cs
@@ -255,17 +255,70 @@ public sealed class ThisQualificationTests
         Assert.DoesNotContain("(this).FaceMethod()", text);
     }
 
-    // Negative guard: a boxed `this` that reaches a NON-interface member (here
-    // `object.ToString()` via `((object)this).ToString()`) must NOT be over-cast to
-    // the enclosing struct — `IsInterfaceCastThisReceiver` declines a non-interface
-    // callee, so the boxed-this arm never fires. The faithful render keeps the
-    // object cast the boxing spells and never writes `((StructBoxedObjectReceiver)this)`.
+    // A boxed `this` reaching a NON-interface VIRTUAL member (here
+    // `object.ToString()` via `((object)this).ToString()`) renders the cast form
+    // `((object)this).ToString()`: the explicit `box` is the source upcast, and a
+    // virtual `callvirt object::ToString` re-emits from `ldobj; box`. Bare
+    // `(this).ToString()` would instead lower to `constrained. callvirt` (no box) —
+    // not opcode-faithful — and must never be over-cast to the enclosing struct
+    // (#3213, extending #3201 beyond interface callees).
     [Fact]
-    public void StructBoxedThis_NonInterfaceCallee_IsNotOverCast()
+    public void StructBoxedThis_ObjectVirtualCallee_RendersObjectCast()
     {
         var text = RenderMember(typeof(StructBoxedObjectReceiver),
             nameof(StructBoxedObjectReceiver.CallObjectToString));
+        Assert.Contains("((object)this).ToString()", text);
         Assert.DoesNotContain("StructBoxedObjectReceiver)this", text);
+        Assert.DoesNotContain("(this).ToString()", text);
+    }
+
+    // A boxed `this` reaching a NON-interface NON-VIRTUAL base member renders
+    // `base.M()`. A struct overriding `GetHashCode` and calling `base.GetHashCode()`
+    // lowers to `ldarg.0; ldobj S; box S; call System.ValueType::GetHashCode; ret`
+    // — a non-virtual `call` (not `callvirt`) to a base method. Bare
+    // `(this).GetHashCode()` re-dispatches virtually to the struct's own override:
+    // infinite self-recursion. `base.GetHashCode()` re-emits exactly the
+    // `ldobj; box; call ValueType::GetHashCode` (#3213).
+    [Fact]
+    public void StructBoxedThis_NonVirtualBaseCallee_RendersBaseCall()
+    {
+        var text = RenderMember(typeof(StructBaseHashCall),
+            nameof(StructBaseHashCall.GetHashCode));
+        Assert.Contains("base.GetHashCode()", text);
+        Assert.DoesNotContain("(this).GetHashCode()", text);
+        Assert.DoesNotContain(")this).GetHashCode()", text);
+    }
+
+    // A boxed `this` reaching a virtual member through an explicit base cast renders
+    // the cast form, not `base.`. Casting a struct to `System.ValueType` boxes;
+    // because `GetHashCode`'s virtual slot is introduced on `System.Object`, csc
+    // binds the `callvirt` to `object::GetHashCode`, so the faithful cast is
+    // `((object)this).GetHashCode()` (re-emits `ldobj; box; callvirt
+    // object::GetHashCode`). Distinguished from `base.M()` by the call being
+    // virtual, not by the receiver shape (#3213).
+    [Fact]
+    public void StructBoxedThis_ExplicitBaseCastVirtualCallee_RendersObjectCast()
+    {
+        var text = RenderMember(typeof(StructValueTypeCastReceiver),
+            nameof(StructValueTypeCastReceiver.CallViaValueTypeCast));
+        Assert.Contains("((object)this).GetHashCode()", text);
+        Assert.DoesNotContain("base.", text);
+        Assert.DoesNotContain("(this).GetHashCode()", text);
+    }
+
+    // Negative guard (#3213): an implicit `this.ToString()` on a struct that does
+    // NOT box does not go through the boxed-this arm. Because the struct does not
+    // override `ToString`, csc emits `ldarg.0; constrained. S; callvirt
+    // object::ToString()` — a bare `LoadArgument{0,"this"}` receiver with a
+    // `constrained.` prefix and NO `box`, so `IsBoxedThisReceiver` declines. The
+    // render must not gain a cast or `base.` — the constrained call already binds.
+    [Fact]
+    public void StructImplicitThis_ConstrainedCall_IsNotBoxedOrBased()
+    {
+        var text = RenderMember(typeof(StructImplicitToString),
+            nameof(StructImplicitToString.CallImplicitToString));
+        Assert.DoesNotContain("base.", text);
+        Assert.DoesNotContain(")this).ToString()", text);
     }
 
     [Fact]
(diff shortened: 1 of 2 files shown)


richlander:
## Adversarial review reconciliation (head `ffc87d68`)

Both fixed-head re-reviews at `ffc87d68` are **clean** (Gemini 3.1 Pro + GPT-5.6). Below is the full cross-round accounting; every finding was reproduced on a clean review worktree before fixing and re-verified at the fixed head.

### Round 1 — at `febb077a` → fixed in `5a7ed86b`
| --- | --- | --- |
| GPT | A boxed static **extension** method group (`((object)this).Ext`) shares the `ldobj; box; ldftn` shape but binds the boxed `this` as its first arg; casting to the static host is CS0716. | Gated the `MethodGroupText` boxed branch on the callee's `method.HasThis`; extension groups fall through to the ordinary receiver path. |
| Gemini | A `static` method whose first parameter is spelled `@this` emits metadata name `"this"` at index 0, matching the boxed-`this` shape, but `base`/`this` are illegal there (CS0026). | Gated `IsBoxedThisReceiver` on the enclosing `_function.Signature.HasThis`. |

### Round 2 — at `5a7ed86b` → fixed in `9a5c64d1`
| --- | --- | --- |
| Gemini + GPT (same core bug) | A sealed **default interface member** emits a non-virtual `call I::M`; the base arm over-fired and rendered invalid `base.SealedMethod()` (CS0117). | Restricted the base arm to a `System.ValueType` callee (`IsSystemValueType`); interfaces take the `((I)this)` cast. |
| GPT (token-exactness angle) | A non-virtual `call` to an **overridable** `System.Object` method (e.g. hand-emitted `call object::GetHashCode`) would silently rebind under a `base.` spelling. | Same ValueType-only gate routed it to the cast — but this over-corrected (see Round 3). |

### Round 3 — at `9a5c64d1` → fixed in `ffc87d68`
| --- | --- | --- |
| Gemini + GPT (same regression) | The ValueType-**only** gate broke `base.MemberwiseClone()` — a real csc `ldobj; box; call object::MemberwiseClone`. `MemberwiseClone` is **protected**, so the `((object)this).MemberwiseClone()` cast is **CS1540**; `base.` is the only valid spelling. | Renamed `IsSystemValueType` → `IsStructBaseClass`, matching **`System.ValueType` OR `System.Object`** — a struct's only two base classes. Rule is now "base for a struct's base classes, cast for interfaces". My earlier ValueType-only restriction over-corrected the sealed-DIM fix; the correct fix is to exclude interfaces, not to restrict to ValueType. |
| GPT (residual, non-blocking) | A **synthetic, non-csc** nonvirtual `call`/`ldftn` to an **overridable** base member that `ValueType` overrides (`call object::GetHashCode`, non-final DIM direct call) has **no faithful C# spelling** — `base.` rebinds to `ValueType`'s override, a cast re-dispatches (`callvirt`). | Acknowledged as out of scope: csc never emits it, C# cannot express suppressed dispatch to an overridden base slot, and detecting it needs the callee's declared virtuality/finality (not exposed on `MethodRef`). Filed as follow-up **#3228**. The `base.` render of this synthetic shape is valid C# and strictly better than the pre-fix invalid `(this).M()`. Both reviewers agree #3228 is the correct scope boundary; GPT found no csc-producible instance. |

### Fixed-head re-review (`ffc87d68`) — both clean
- **GPT-5.6**: "No significant issues found." Verified the Release build, 40 render + 27 decision tests, `base.MemberwiseClone()` compiles (CS1540 gone), `base.GetType()` preserves `call object::GetType`, sealed DIM retains the interface cast, other csc base calls resolve to `ValueType`. Agrees #3228 is a reasonable scope boundary.
- **Gemini 3.1 Pro**: "Clean. No residual or new invalid C# syntax generation exists." Verified the full spectrum with real builds/decompilations: MemberwiseClone call+group → `base.` (compiles), GetType → `base.GetType()` (byte-identical IL), sealed DIM → `((ISealedDim)this)` cast, ValueType base calls → `base.`, `((object)this).ToString()` virtual → cast (blocked by `!IsVirtual`), and confirmed `IsStructBaseClass`'s exact two-type match cannot over-fire.

### Validation at `ffc87d68`
- `ThisQualificationTests` **40/40**; Decision **27/27**
- Fast decompiler suite **3761/1** — the one failure is the pre-existing Windows-local `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString` (CRLF), not introduced here
- RoundTrip area (incl. slow compile-back gates) **473/3** — the three failures are the pre-existing `GeneratedFixtureCatalogTests`, not introduced here
- PR quick gate **PASS**: 0 pass bugs, detected lowering residue 251 → 244 (16.73% → 16.27%)
- Real `member --library` renders on a compiled fixture: `base.MemberwiseClone()`, `base.MemberwiseClone` group, `base.GetType()`, `base.GetHashCode()`

richlander:
Ready to merge

- Both fixed-head adversarial re-reviews at `ffc87d68` are clean (Gemini 3.1 Pro + GPT-5.6); full cross-round reconciliation above.
- CI green on `ffc87d68`: `test (ubuntu-26.04, linux-x64)` pass, `csharp-diff-smoke` pass, `build-net10` pass, `changes` pass (`il-diff-smoke`/`markdownlint`/`pack` path-gated, skipping).
- Local validation: ThisQualificationTests 40/40, Decision 27/27, fast suite 3761/1 (pre-existing StringSwitch CRLF), RoundTrip area 473/3 (pre-existing catalog), PR quick gate PASS (0 pass bugs, residue 244).

Non-blocking follow-up: #3228 (synthetic, non-csc nonvirtual call to an overridable base member — no faithful C# spelling; needs callee method-virtuality metadata).

PR #3226, opened 2026-07-26
Spell var for apparent-type declarations (opt-in bucket) (#3169)

richlander:
## What

Lights up the **printer emission** for the apparent-type `var` bucket (`csharp_style_var_when_type_is_apparent` / `PreferVarWhenTypeApparent`), building on the byte-neutral plumbing from #3202. When the option is set and a local declaration's type is **apparent** from its initializer — object creation of the exact type, an array creation, or an explicit reference cast — and the declared type is **not** a C# built-in keyword type, the declaration is spelled `var` instead of its explicit type.

This is the first of the three var buckets to emit; **built-in** and **elsewhere** are deferred to a follow-up PR (they need a type-safety gate across the full coercion surface).

## Why it's faithful & byte-safe

- **Byte-neutral.** `var` is a compile-time inference with no IL consequence. Apparency *guarantees* the initializer's static type **is** the declared type, so `var` infers exactly that type and the emitted form recompiles to identical IL.
- **Off by default → default output unchanged.** With the option false, `SpellVarForApparentType` short-circuits, `DeclarationTypeText` returns `TypeText(type)` as before, and `InitializerText(s.Value, s.Type, s.Type)` is identical to the previous 2-arg call. The default path is byte-for-byte unchanged (3740 fast decompiler tests green).

## Either/or with the target-typed-`new` shortener

Spelling `var` and shortening the RHS to `new()` are mutually exclusive spellings of the same apparent site — `var x = new()` is CS8754. So when `var` is chosen, the shortener is suppressed for that site (a null new-target), and an apparent object creation renders `var x = new List<int>()`, never `var x = new()`.

Observed emission:

| shape | default | `var_when_type_is_apparent = true` |
| --- | --- | --- |
| `List<int> x = new List<int>()` | `List<int> x = new();` | `var x = new List<int>();` |
| `int[] x = new int[4]` | `int[] x = new int[4];` | `var x = new int[4];` |
| `Node x = (Node)o` | `Node x = (Node)o;` | `var x = (Node)o;` |
| `string x = new string('x',3)` (built-in) | `string x = new('x',3);` | *unchanged* (declines) |
| `List<int> x = Make()` (not apparent) | `List<int> x = Make();` | *unchanged* (declines) |

## Design

- The built-in exclusion (`IsBuiltInType`, drawn from the shared `PrimitiveTypeNames` keyword table — the same one `TypeText` renders keywords from) keeps the apparent and built-in-types buckets a clean partition.
- Reuses the merged `TypeIsApparent` predicate (#3194); no new apparency logic. SRM-only, Roslyn-free, NativeAOT-friendly.

## Tests

`VarWhenApparentTests` over compiled specimens pin: the default (explicit, shortened), the `var` rewrite for each apparent shape, the either/or interaction (asserts **no** `var x = new()`), and both decline cases (built-in type, non-apparent initializer). Full fast decompiler suite (3740) green.

Note: the heavy `Speed=Slow` Fidelity gates surfaced 3 pre-existing/environmental failures unrelated to this change — they run at **default** options (which this PR leaves byte-identical), assert only on named external fixtures (not this PR's specimen), and are **not** part of the merge-blocking `test` job (CI runs `-trait- "Speed=Slow"`).

Part of #3169. Stacks on #3202. Follow-up PR: built-in + elsewhere buckets.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/VarWhenApparentSpecimen.cs
+++ b/src/ILInspector.Decompiler.Tests/VarWhenApparentSpecimen.cs
@@ -1,0 +1,64 @@
+using System.Collections.Generic;
+
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// Compiled specimens for the opt-in apparent-type <c>var</c> bucket
+/// (<see cref="ILInspector.Decompiler.Pipeline.PrinterOptions.PreferVarWhenTypeApparent"/>,
+/// <c>csharp_style_var_when_type_is_apparent</c>). Each method declares a real local
+/// (used more than once so it survives as a declaring store) whose type is — or is
+/// deliberately not — apparent from its initializer.
+/// </summary>
+public sealed class VarWhenApparentSpecimen
+{
+    // Apparent via object creation of the exact (non-built-in) type. The default view
+    // shortens the RHS to `new()`; the var bucket must keep the explicit `new List<int>()`
+    // (a bare `var x = new()` is CS8754), so the two spellings are mutually exclusive.
+    public static int ObjectCreation()
+    {
+        List<int> items = new List<int>();
+        items.Add(1);
+        return items.Count;
+    }
+
+    // Apparent via single-dimension array creation. The declared type `int[]` is a shape,
+    // not a built-in keyword type, so the built-in exclusion does not apply here.
+    public static int ArrayCreation()
+    {
+        int[] buffer = new int[4];
+        buffer[0] = 7;
+        return buffer[0];
+    }
+
+    // Apparent via an explicit reference cast to a non-built-in type.
+    public static int ReferenceCast(object o)
+    {
+        Node node = (Node)o;
+        return node.Value + node.Value;
+    }
+
+    // Negative: apparent, but the declared type is a C# built-in keyword type
+    // (`string`) — governed by the separate built-in-types bucket, so the apparent
+    // bucket must decline and leave the explicit type.
+    public static int BuiltInObjectCreation()
+    {
+        string text = new string('x', 3);
+        return text.Length + text.Length;
+    }
+
+    // Negative: the type is not apparent from the initializer (a plain call), so the
+    // apparent bucket must decline regardless of the option.
+    public static int NotApparent()
+    {
+        List<int> items = Make();
+        items.Add(1);
+        return items.Count;
+    }
+
+    static List<int> Make() => new();
+
+    public sealed class Node
+    {
+        public int Value;
+    }
+}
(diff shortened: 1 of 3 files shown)


richlander:

Two non-Claude reviewers examined the exact head `19b804d6` in isolated detached worktrees (read-only, scratch under `/tmp/`).

- **GPT-5.6 (`gpt-5.6-sol`):** *No significant issues found. Nothing blocking.*
- **Gemini-3.1 Pro (`gemini-3.1-pro-preview`):** *No blocking issues* — verified all six invariants with real runs:
  1. **Byte-neutral / default unchanged** — with the option false, `SpellVarForApparentType` is false, `DeclarationTypeText` falls back to `TypeText(type)`, and `InitializerText` receives `s.Type` as `newTarget`, exactly matching the old 2-arg overload. Anonymous-type emission and ordinary assignments unaffected.
  2. **Either/or with target-typed-`new`** — spelling `var` threads a `null` new-target, reliably bypassing the `new()` reduction so `CoerceText` emits the explicit `new T(...)`, preventing CS8754.
  3. **Faithfulness** — `TypeIsApparent` mandates `.Equals` identity between declared and initializer type; covariant/less-derived/interface declared types fail the check and decline. Struct casts lower to `UnboxAny` (not `CastClass`) and safely return false.
  4. **Built-in exclusion** — `IsBuiltInType` is bounded to `TypeRefKind.Definition`, so arrays/pointers/byrefs correctly qualify for `var`; `object`/`string`/`int` and `nint`/`nuint` route to keywords and are deferred to the built-in bucket.
  5. **Only intended sites change** — fluent/logical-chain values are `Call`/`LogicalBinary`, for which `TypeIsApparent` is false, so no `var`/`new()` can leak there.
  6. **NativeAOT/Roslyn/SRM** — all checks resolve natively over `TypeRef`/`IrExpression`; nothing introduced.

No findings required action from either reviewer; no code changes since the reviewed head, so both fixed-head reviews are clean.

**Note on slow gates:** the `Speed=Slow` Fidelity gates surface 3 pre-existing/environmental failures unrelated to this change — they run at **default** options (byte-identical here), assert only on named external fixtures (not this PR's specimen), and are **not** part of the merge-blocking `test` job (CI runs `-trait- "Speed=Slow"`).

richlander:
Ready to merge

All merge-blocking CI is green (build-net10, test, csharp-diff-smoke, changes). Both fixed-head adversarial reviews (GPT-5.6 + Gemini-3.1) are clean with no required actions. The default output path is proven byte-identical (var is opt-in, off by default). The built-in and elsewhere var buckets are non-blocking follow-up in a later PR.

PR #3229, opened 2026-07-26
Library section model: declared SizeClass + Cost axes, curated -D catalog

richlander:
## What

Reworks the **library** section pipeline around two **declared** per-section axes, replacing the overloaded `Info`/`Noisy`/`ExplicitOnly`/`ListedInCatalog` flag tangle for the curated catalog. Follow-up to #3006.

Opt-in per pipeline via `SectionPipeline.UseCuratedCatalog()` — **only `LibrarySections` opts in**; package/type/member keep the legacy path unchanged.

### Two declared axes
- `SectionSizeClass` — `Terse` (≤12) · `Informative` (≤24) · `Verbose` (>24). Author's stable row-count hint (from the empirical SPC/STJ/DI + 16-package corpus inventory), not a measured value.
- `SectionCost` — `NetworkFree` · `Moderated` (bounded, sub-second network/PDB) · `Unbounded` (fan-out / source download / whole-assembly scan).

### Curated verbosity ladder
| View | Renders |
| --- | --- |
| `-v:q` | none (identity line only) |
| `-v:m` | target (`Info`) section only |
| `-v:n` | Terse + Informative, network-free |
| `-v:d` | all sizes, bounded cost (never `Unbounded`) |

`Unbounded` sections never auto-render; reachable only by exact name or a category door.

### `-D` catalog
Now **categories-first**: topical category doors (alpha) then one flat effective-section group, with the internal `(verbose)`/`(opt-in)` annotations **dropped**. Categories hyper-subscribe (drop count-0 members, drop empty categories). Performance buckets stay behind the `@Performance` door (`ListedInCatalog=false`) while still rendering at `-v:d`.

### Category edits
- `@Switches` removed (redundant single-member door) → **Switches lives in `@Audit`**
- `+Non-normalized Paths` → `@Audit`
- `−Union Types` from `@Integrations`
- `+Symbols` → `@Source`

Footguns (Unsafe Members, Top Leverage, Source Files, SourceLink Integrity) and the coordinate-gated IL-context sections keep `ExplicitOnly` + `Cost=Unbounded`.

## Evidence
- `SectionPipelineTests` 125/125 green.
- Full CLI suite green except the known env-only `PlatformResolverTests.GetPacksDirectory_ReturnsValidPath` (cache-path assertion, unrelated).
- Empirically validated on `System.Text.Json`: `-D` categories-first + flat spine + no annotations + hyper-subscribed; `-v:n` (Terse+Informative) ⊊ `-v:d`; no footgun leaks; `-D @Audit` lists Switches.

## Naming and discoverability pass (added after the first two reviews)

Five follow-up commits. A UX pass over section names, plus the guard that makes
renames like these safe.

**Guard first.** Category membership was declared with raw string literals and
`AddCategory` never validated them, so renaming a section silently dropped it
from its category instead of failing the build. Library section names are now
`SectionNames` constants, `AddCategory` validates every member against the
registered sections, and three tests pin it: a non-vacuity test that fails if
the validation is removed, an all-pipelines sweep, and a test tying the
`SourceLink:` prefix family to its door.

| Change | Rationale |
| --- | --- |
| `Source Link: X` → `SourceLink: X` | match the `@SourceLink` door it belongs to |
| `Performance: Allocation Hotspots` / `Closures and Delegates` / `Loop Hot Paths` | Title Case, like the rest of the catalog; the flattened tabular `Kind` column derives from these names and follows |
| `Escape: Array Pool` → `Array Pool Escapes` | a `Group: Leaf` prefix advertises a category door; this family lost its single-member `@Escape` door, so the prefix promised a door that no longer exists |
| `SourceLink: Integrity` joins `@SourceLink` | completes the deferred item below |
| `Array Pool Escapes` becomes catalog-hidden | its applicability is a capability predicate, not a content predicate |

`Array Pool Escapes` previously listed under `-D` and then rendered nothing,
because `CanRender` asks whether the scan *can* run, not whether it *found*
anything. Sections list in `-D` only when >0 rows can be established cheaply,
so it is now `ListedInCatalog => false` — reachable by exact name and through
`@All`/`--schema`, exactly as the kind-scoped `Performance:` sections already
work.

`SourceLink: Integrity` was previously reachable only through the `@Hidden`
pole. This resolves the "blocked on an unbounded-in-render-door nuance" item by
accepting the fan-out rather than building the nuance: `-S @SourceLink` now
runs Integrity's GET+hash pass, measured at roughly **+0.3-0.4s cold**
(System.Text.Json, 353 files, alternating cold-cache reps). The door already
roots two other `Unbounded` members, so this does not change its cost class.
`-S @Hidden` remains rejected as discovery-only.

Two rules came out of this pass and are now written into
`docs/design/section-model.md`:

1. **A prefix and a category door are two halves of one claim.** A prefix
   advertises membership, so every prefixed section must be reachable through
   its family's door, and a prefix only fits an *exclusively* owned family.
   `P/Invoke Methods` is in both `@Audit` and `@Surface`, so no prefix can
   describe it — cross-cutting lenses stay unprefixed. This is why "give every
   section a prefix" is not achievable.
2. **A section lists in `-D` only when >0 rows can be established cheaply.**

All prior spellings keep resolving through `LegacySectionAliases`.

### Compatibility
Section names are user-visible and scriptable. Every renamed name retains a
legacy alias, verified end to end: `Array Pool Escapes`, `Escape: Array Pool`,
`Resource Escape Triage`, and `Escape` all render 3 identical rows on
`System.Private.CoreLib`, under the new canonical heading.

### Validation
`dotnet-inspect.Tests`: 2282 total, 1 failed, 14 skipped. The failure is
`PackagingSurfaceTests.OnlyTheShippingToolsOptIntoPackaging`, pre-existing and
environmental — it enumerates `.csproj` files without excluding `.worktrees/`
and picks up a duplicate from a sibling worktree; it passes in CI's clean
clone. Build is 0 warnings. `markdownlint` clean on all changed docs.

## Scope / non-goals (deferred follow-ups)
- Package/type/member pipelines untouched (library-only rollout).
- ~~Design docs not yet reconciled~~ — `docs/design/section-model.md` is now reconciled, and additionally records the two naming rules above.
- Done since: `@Integrations` `Integration:` rename; `SourceLink Integrity` → `@SourceLink` (see above — taken by accepting the fan-out cost rather than building the render-door nuance).
- Still deferred: `Library Info` → `Library Information` rename, `Top Leverage` → `@Performance`, bare `-S` Terse∧network-free bundle, and gating door *render* expansion to skip `Unbounded` members (the "discovery-listed, exact-name-run" nuance).
- Filed as follow-ups from this pass: #3410 (`Non-normalized Paths` misses SourceLink document paths and reports a false clean), #3412 (IL coordinate resolution should delegate to the metadata API).

## Review notes
Non-trivial behavior change (new declared axes + `-D` rewrite) → requires two-model adversarial review before **Ready to merge** per `AGENTS.md`.

The naming/discoverability pass above landed **after** the first two clean reviews, and carries its own risk surface (user-visible renames with a compatibility story, a behavioral change to `-S @SourceLink` cost, and a new construction-time validation that throws). Re-reviewed at exact head `380fe250` by two models from different families; reconciliation follows in a comment.


richlander:

Two-model adversarial review (Gemini 3.1 Pro + GPT-5.6 Sol), isolated review worktrees.

- **Initial review** at `811b3268`.
- Findings addressed in follow-up commit `3ef391c9`.
- **Re-review of the fixed head `3ef391c9` — both reviewers clean.**

### Findings and dispositions

| # | Reviewer | Finding | Disposition |
| - | -------- | ------- | ----------- |
| 1 | Gemini | `@Audit` roots a phantom `Signals` member | **False positive.** `@Audit` = Unsafe Members / P-Invoke / Switches / Non-normalized Paths. The `ScannerAuditSignals` constant is the *scanner key* for the `Signals` section, not a category membership; `Signals` is intentionally uncategorized. Gemini confirmed on re-review. |
| 2 | Gemini #3 / GPT #2 | Topical doors (`@Source`, `@Audit`) root `Unbounded` members, so `-S @Category` fans out to unbounded work | **Real but pre-existing** — those memberships are byte-identical on `origin/main`. `ExplicitOnly` keeps them off the verbosity ladder but not out of door expansion. Corrected the doc over-claim: `docs/design/section-model.md` now states the footgun honestly and marks door-render gating as **deferred hardening**. |
| 3 | GPT | Curated `-D` bypassed by the tree branch — `-D -v:d`/`--tree` leaked `@All`/`@Default`/`@Hidden` poles + `(verbose)`/`(opt-in)` annotations | **Real regression, FIXED.** `DiscoverOutput.WriteTree` now takes `listedCategoryDoors` and renders the curated categories-first catalog. Added regression test `LibraryCommand_DiscoverDetailedTree_UsesCuratedCatalog`. Both reviewers verified the fix is complete across tree/schema/JSON/JSONL/TSV; legacy package discovery unchanged. |
| 4 | GPT | Weak reworked ordering tests | **Tightened.** `DetailedOutput_RendersSectionsAlphabetically` asserts the Performance/Resource-Triage cluster actually trails (no spine section at/after the first trailing member) + spine alpha; `DiscoverSchema_GroupsOptInSections` asserts exactly the six doors in alpha order and that every category row precedes every section row. GPT confirmed these now reject the relevant violations. |
| 5 | Gemini #4 | `CuratedRequiredVerbosity` returns `Detailed` for `Unbounded` but the ladder rejects it | **Inert, documented.** Unbounded sections are `ExplicitOnly`, reached only via explicit include, which bypasses the ladder; the promoted verbosity never authorizes extra rendering or capability work. Gemini traced `IsRequested`/`GetAuthorizedSections` and confirmed. Added a clarifying `<remarks>`. |
| 6 | GPT #1 | `-D -v:d --verbose` triggers network source-URL verification | **Real but pre-existing** — `LibrarySourcePlan.cs`/`SourceAuditService.cs` are byte-identical on base `0a11e21d`; cache hits explain inconsistent reproduction. Not introduced or changed here. Filed as deferred follow-up. |
| 7 | GPT #4 | Coordinate IL-context sections surface in flat `-D` with `--il-offset` | **Pre-existing intended coordinate-discovery.** `EffectiveCatalogHidden` `ExceptWith(ILCoordinateSections)` carve-out is byte-identical on base. No change. |

### Re-review verdicts (fixed head `3ef391c9`)

- **Gemini:** "I found no new logic errors, execution leaks, or regressions introduced by the follow-up commit. The fixed head is clean."
- **GPT:** "No new significant issues found on `3ef391c9`. … All 2,131 tests passed except the specified pre-existing `GetPacksDirectory_ReturnsValidPath` environment failure."

Deferred follow-ups (tracked in the PR body): gate `-S @door` expansion to skip `Unbounded` members, and audit the `-D -v:d` source-verification path.

richlander:
Ready to merge

- All required CI green: `test` (8m28s), `build-net10`, `markdownlint`, `changes`.
- Two-model adversarial review (Gemini 3.1 Pro + GPT-5.6 Sol) complete; both re-reviews of the fixed head `3ef391c9` are clean (reconciliation above).
- Deferred, non-blocking follow-ups are listed in the PR body and the reconciliation table.

richlander:
Follow-up commit `70d8e96c`: renamed the resource-triage naming per review feedback.

- Category `@Resources` → **`@Escape`** (one-word door; "Resource Escape Triage" family).
- Section `Resource Triage` → **`Escape: Array Pool`** (named for the array-pool shape it detects today).
- Moved the `Resources` manifest section into **`@Surface`** (a section and a category no longer share a name).
- `Escape: Array Pool` is `ListedInCatalog=false`: flat `-D` shows only the `@Escape` door; the section is reachable via `-D @Escape` or exact name, and still auto-renders by SizeClass.
- Back-compat aliases: `-S "Resource Triage"` → section; `-S Escape` / `-S "Resource Escape Triage"` → `@Escape`.
- Bumped effective-discovery disk cache (`effective-v11`→`effective-v12`) to invalidate stale catalogs.

Underlying analysis/model identifiers unchanged. Broadening escape analysis beyond ArrayPool is tracked in #3237.

Tests: CommandExecutionTests 580 (2 env-skips), SectionPipelineTests 125, OutputFormatter+IndexBuildInvariant 123 — all green. markdownlint clean.

This adds behavior (catalog rename), so re-review of the new head is warranted before merge.

richlander:
Pushed `05101d59`: library `@Source` → `@SourceLink` (leaving `@Source` for the API command's actual-source sections), and the three sections renamed to the `Source Link:` family so they cluster in alpha sort like `Performance:`.

- `@SourceLink` door roots Source Files + Source Link: Availability/Missing Files.
- Source Link: Integrity stays in the `@Hidden` pole (its GET+hash fan-out is the priciest; keeping it out of the door preserves the cost-safety boundary — `-S @SourceLink` can't trigger a full integrity sweep). Reachable via exact name / `-D @Hidden`.
- Back-compat `-S` aliases: old `SourceLink *` names + `SourceLink`/`Source Link` category resolve to the new names.
- Effective-discovery cache bumped v13→v14.

Verified live (`-D @SourceLink` excludes Integrity; `-D @Hidden` includes it; back-compat alias renders). Tests green; design docs updated; historical release notes left as-is.

This is the 4th behavioral commit atop the previously-clean adversarial review — re-review of the new head is warranted before merge.

richlander:
Pushed Phase B: **symbol-dependent cache-aware SourceLink discovery** (commit `73ac7798`, on merged head `9187ab5d`).

The library `-D` catalog now lists the SourceLink family only when a local PDB (embedded, adjacent, or **already in the symbol cache**) exposes a SourceLink document — determined **network-free**. With no resolvable SourceLink, the `@SourceLink` door hyper-subscribes away; warming a PDB into the cache makes it appear on the next `-D`. Rendering (`-S "Source Link: …"`) still does its network work on demand.

- `SymbolPackageDownloader` + `SourceEnricher`: `cacheOnly` PDB resolution (returns after cache-hit check; no GET/HEAD/PutAsync).
- `LibraryMetadataService.ProbeLocalSourceLinkAsync`: network-free bool probe.
- `LibrarySections`: family gated on `SourceLinkDiscoverable = AssemblyInfo != null && HasSourceLink` (still ExplicitOnly).
- `LibraryCommand -D`: probe in all 3 input branches; effective cache key folds a `#sl0/#sl1` token so warming/clearing a cached PDB busts a stale catalog. Cache `v15 → v16`.

Live-verified the warm/hide transition (STJ + Newtonsoft), render paths unchanged, warm `-D` ~0.14s. Tests green (SectionPipeline 126, discovery/SourceLink CommandExecution) except the known env-only PlatformResolver failure.

This changes network/cache-default behavior → **re-review warranted**. Two-model adversarial review (Gemini 3.1 Pro + GPT-5.6 Sol) launched at head `9187ab5d`; will reconcile findings here before marking ready.

richlander:

Two independent reviewers ran against head `9187ab5d`: **Gemini 3.1 Pro** and **GPT-5.6 Sol**.

### Blocking finding (both reviewers, independently) — FIXED

Both flagged the same real gap on attack point *network-free guarantee*: on an effective-cache **miss**, `-D` fell through to the full `InspectAsync`, whose source plan enabled `AllowPdbDownload` under detailed verbosity (`-v:d`) or an `-S Symbols` filter — so **`-D -v:d` downloaded a PDB during discovery**.

I reproduced it and confirmed it is real. Notes:
- It reproduces identically on the pre-Phase-B base (`a2349d1e`), i.e. it is **pre-existing**, but it directly defeats *this* PR's core invariant (SourceLink family listing is network-free and probe-driven), so it is in scope.
- It also masked a **latent effective-cache token mismatch**: the cache-only probe sets the read token `sl0` while the `-v:d` download set `HasSourceLink` → write token `sl1`, so `-v:d -D` never hit the cache and re-downloaded every run.

**Fix:** `e13acea9` — thread a `discoveryOnly` flag through `InspectAsync`/`CollectPackageInspectionsAsync`; when set, `allowPdbDownload = sourcePlan.AllowPdbDownload && !discoveryOnly`. The three `-D` branches pass `discoveryOnly: effectiveDiscovery`. The SourceLink family is now listed solely from the network-free probe, so the `-D` catalog is identical across verbosities and the cache token is always consistent. Effective cache `v16 → v17`.

Verified live: `-v:d -D` with no cached PDB issues **zero HTTP** and lists no SourceLink family (matches bare `-D`); with a cached PDB both bare `-D` and `-v:d -D` list the family; render path `-S "Source Link: Files"` still fetches URLs on demand. Blast radius limited to the SourceLink family listing (Symbols/Signals unaffected). Tests: SectionPipeline 126/126, SourceLink/Discover/Effective/Groups CommandExecution 59/59.

### VERIFIED-OK (both reviewers)

Cache read/write token consistency, warm/clear transition (no stale-cache masking), render path unaffected, package-vs-server cache keys, no silent data hiding, probe robustness/disposal, and perf (warm `-D` ~88–148 ms). GPT confirmed each `cacheOnly` short-circuit sits before HTTP/`PutAsync`.

### Reviewer-hygiene note

The Gemini review worktree was left with an uncommitted debug `Console.WriteLine` (scratch, never committed); I reverted it and re-ran all verification on a clean tree.

Re-review of the fixed head `e13acea9` is now running with both models; I'll confirm both are clean before marking ready.

richlander:
## Fixed-head re-review (round 2) — reconciliation

Re-reviewed the round-1 fix `e13acea9` with both models.

- **Gemini 3.1 Pro:** all 7 checks VERIFIED-OK (fix restores network-free `-D`, `-v:d -D` now cache-hits — `Cache | 1 hit`, render path still downloads, warm `-D` 64–99 ms).
- **GPT-5.6 Sol:** confirmed the requested no-PDB cases (0 HTTP) and cache-hit behavior, but found **two further `-D` network leaks** that `e13acea9` did not cover:

### Finding A (blocking) — FIXED in `dec997b8`

`discoveryOnly` gated only the PDB *download*. For an assembly with an **embedded/adjacent** SourceLink PDB, discovery has `pdbContext.HasPdb == true`, so `InspectAsync` still ran the source-URL **HEAD audit** (`sourcePlan.RunHeadAudit` → `SourceAuditService.PopulateAsync`) and integrity **GET+hash** during `-D -v:d` — GPT measured **174 HEAD requests / 5.79 s** on a local assembly. (The earlier System.Text.Json checks missed it because its PDB is cache-only, so `SourceLinkService.Open` doesn't load it during discovery.)

**Fix:** neutralize the *entire* source plan during discovery — `sourcePlan = discoveryOnly ? default : LibrarySourcePlans.For(...)` — so no PDB download, no HEAD audit, no integrity GET, no source-file collection. Listing is unaffected (family lists from the network-free probe; `HasSourceLink` for embedded/adjacent PDBs is set independently of the audit). Verified: embedded-PDB `dotnet-inspect.dll -D -v:d` now issues **0 HTTP** yet still lists the family; render path still audits (`669/669 available`). Tests 126/126 + 60/60 green.

### Finding B (`--il-offset … -D`) — out of scope, non-blocking

`--il-offset` with `-D` resolves the IL coordinate via `ILOffsetQuery` → `AcquirePdbAsync`, which may download a PDB. This is **pre-existing and by design**: `PopulateILOffsetIfRequestedAsync` deliberately runs when `options.Discover != null` because IL-coordinate → source-location discovery inherently needs symbols, and the user explicitly opted into symbol work via `--il-offset`. That path forces `sourceLinkAvailable = false`, so it is independent of the SourceLink family listing this PR governs. Not changed here; can be a separate follow-up if we want `--il-offset -D` to be cache-only too.

Note: Gemini's round-2 pass did not catch Finding A (its transitions used the cache-only STJ PDB). Re-review of the new head `dec997b8` is running with both models; I'll confirm both clean before marking ready.

richlander:
## Fixed-head re-review (round 3) — reconciliation

Re-reviewed the comprehensive fix `dec997b8` with both models.

- **Gemini 3.1 Pro:** all 6 checks VERIFIED-OK. Confirmed embedded-PDB (`dotnet-inspect.dll`) **and adjacent-PDB (`ILInspector.Analysis.dll`, adjacent Standalone PDB)** `-D -v:d` now issue **0 HTTP** while still listing the family; cache-only STJ 0 HTTP; render path still audits (`174 requests`, `669/669 available`); warm `-D` ~140 ms cache hit.
- **GPT-5.6 Sol** (which found the round-2 leak): its prior finding is **VERIFIED-OK fixed** — cold `-D -v:d`, `-S "Source Link: Integrity" -D`, and `-S Symbols -D` all **0 HTTP**. It enumerated every network-capable discovery stage (PDB download, HEAD audit, integrity GET, source collection, Signals, symbol probe) across file/platform/package inputs — **all 0 HTTP**. Render path intact (174 HEAD / 174 GET). All three `-D` branches confirmed to return via `WriteEffectiveSections` (no suppressed-data render).

### Remaining GPT finding — pre-existing, out of scope → follow-up #3263

GPT flagged (Medium) that local `.nupkg`/package `-D` **perpetually misses the effective-section cache** because the key includes the temporary extraction path (regenerated each run). I verified this is **pre-existing and orthogonal** to this PR: before #3229 the key was already `{assemblyPath}#{size}` (this PR only appended the `#sl` token), and `PackageExtractor`'s temp extraction is untouched here. It is **perf-only — discovery stays 0 HTTP**. Tracked as #3263.

### Status

Both fixed-head re-reviews of `dec997b8` are clean of blocking findings; the network-free `-D` invariant is confirmed by both models across embedded / adjacent / cache-only / package / file / platform inputs at all verbosities. Tests: SectionPipeline 126/126, SourceLink/Discover/Effective/Groups CommandExecution 60/60.

Verifying CI next.

richlander:
Ready to merge

richlander:
## Adversarial re-review — new behavioral commit `044c1cde` (head `7ab99c83`)

This commit adds two behaviors: (1) bare `-S` (valueless `-S` == `@Default`) now renders the **network-free bounded overview** (== the `-v:n` section set, incl. symbol-dependent Symbols/Signals read from an embedded/adjacent/**cached** PDB, 0 HTTP), and (2) removes the niche, network-dependent **"SourceLink availability"** row from library `Signals` (dropping the pointless `auditSources` HEAD fan-out).

Reviewed at exact head `7ab99c83` (behavioral commit `044c1cde` only; the rest of the head is an unrelated origin/main merge). Two isolated review worktrees, two different models.

### GPT-5.6 Sol — no blocking findings
Verified: build clean; SectionPipelineTests 127/127; CommandExecutionTests 580 (2 platform skips); bare `-S` ≡ `-v:n` ≡ explicit `@Default` (diff exit 0); `-S -v:d` still Detailed (Async/Extension Methods present); type/member/package unaffected; cold + warm + embedded + adjacent PDB paths all **0 HTTP** with the symbol cache **unchanged** after cache-only render; `-D`, `-D -v:d`, `-D @SourceLink`, `-S Symbols -D` all 0 HTTP; `#sl0` cold discovery cache correctly busted after warming; `-S Signals` / warm `-v:d` no longer emit "SourceLink availability"; explicit `Source Link: Availability`/`Missing Files` still audit.

### Gemini 3.1 Pro — no blocking findings
Verified: interception is strictly library-scoped and fires only for lone `@Default`/InfoSelector (`-S Symbols`, `-S @Surface`, `--json`, package/type unaffected); verbosity never downgraded (`-S -v:d` renders Detailed-only sections e.g. `## Performance: Boxing`). Built a `DebugType=none` `dummy.dll` and **instrumented `AcquirePdbAsync`**: confirmed `cacheOnly: true` threads into `SymbolPackageDownloader.DownloadPdbAsync`, which early-exits after the local-store check with **zero HTTP** and graceful missing-symbol fallback. bare `-S` ≡ `-v:n` (1:1 section headers). `-D` stays network-free via `ProbeLocalSourceLinkAsync` (`readCachedPdb=false` under `discoveryOnly`). `grep -ri "SourceLink availability" src/` confirms the row is fully erased; explicit `Source Link: Availability` still runs the per-file probe (355/357). `ReadCachedPdb` only fires when `!hasExplicitSelection`, so no effective-cache-key drift.

### Reconciliation
Both fixed-head reviews are **clean with zero blocking findings**; no code changes were required after review. I independently corroborated the network-free invariant via `strace -e trace=connect`: cold-cache miss bare `-S` = 0 HTTP (graceful "PDB not checked"); STJ cache-hit bare `-S` = SourceLink Present / PDB Format Portable / 0 HTTP; bare `-S` section headers identical to `-v:n`.

Ready to merge

richlander:
## Adversarial review — commit `11eba5d6` (Fixed-tier bare `-S` + empty-Escape suppression)

Reviewed at head `237329e7` (behavioral commit `11eba5d6` + merge of `origin/main`) by two models in isolated worktrees.

**Both reviews clean — zero blocking findings.**

### GPT-5.6 Sol
Initially flagged one **Blocking** finding: `library <pkg> --tfm all -S` renders only the title. On investigation this is a **pre-existing multi-TFM rendering defect**, not a regression from this commit — reviewer **retracted** the finding after confirming:

- `library Newtonsoft.Json --tfm all -v:d` reproduces the identical empty output, and that path never sets this PR's new `FixedOverview` flag, so it is behaviorally identical to base `7ab99c83`.
- This PR's only change to `WriteLibraryResults` is appending `options.FixedOverview` to three `ComputeIncludeSections` calls; the container/Markout-filter logic is untouched.

Filed as a separate follow-up: #3283. All six in-scope invariants **Pass** (network-free bare `-S` incl. cold-cache `XDG_CACHE_HOME=/dev/null` → no download; package-independence; `-v:n`/`-v:d` membership unchanged; enum-reorder safety; `-v:q`/explicit unaffected; Escape suppression).

### Gemini 3.1 Pro
**No blocking issues.** All seven invariants verified with `strace` network tracing, base-vs-head `-v:n`/`-v:d` diffs, an enum-usage audit, and the `ResourceTriageFixture.ReadBeforeReturn` candidate fixture. Independently confirmed the `-v:q` bare-`S` fix (no longer spuriously promoted).

### What changed
- New `SectionSizeClass.Fixed` growth tier (value 0, sorts below `Terse`): row set constant across the entire package universe.
- `Library Info`, `Signals`, `Symbols` marked `Fixed` (`Signals` was `Informative`); `-v:n`/`-v:d` membership unchanged because `Fixed ≤ Informative`.
- Bare `-S` = `Fixed ∧ NetworkFree` via a new `LibraryOptions.FixedOverview` flag (not an explicit include → preserves the network-free/cache-only-PDB invariant). Package-independent set: `{Library Info, Signals, Symbols}`.
- Empty `Escape: Array Pool` now silently suppressed (row-presence `ShowWhenProperty`, dropped `EmptyText`) to match the Performance sections.

Follow-ups filed: #3283 (pre-existing multi-TFM render defect), #3284 (Step-2 growth reclassification of the remaining tiers).

Ready to merge

richlander:
## Follow-up: alphabetical library section ordering + review reconciliation

Two follow-up commits since the prior "Ready to merge":

- `c85edfa3` — **Render all library sections in a single alphabetical order.** Category-door selection (`-S @Category`) previously applied no section orderer, so `-S @SourceLink` rendered `Source Link: Files / Availability / Missing Files` (view-model declaration order) rather than alphabetically. Now every library selection — default `-v` ladder, `@Category` doors, `@All`, `-v:d` — sorts sections alphabetically (case-insensitive) via the new `SectionPipeline.AlphabeticalSectionOrder`. The `Performance:` group's former curated order is alphabetized too (single-rule design); the shared `Group: Leaf` prefix keeps each group clustered. Package/type/member render order is intentionally unchanged (IL-context sections carry meaningful order).
- `86d92852` — **Fix from adversarial review** (see below).

### Adversarial review (risk-based tier)

This is a mechanical, library-render-only ordering change with no network/cache/security surface → **"more than trivial, but not high risk" → single review with MAI-Code** per the updated policy.

**MAI-Code — first pass (head `4c7a1d8a`): 1 High finding.** The `--count` section map still followed registration order — `WriteLibraryResult`/`WriteLibraryResults` fed `CountOutput.WriteCountMapFromMarkdown` a list built from `AllSectionNames.Where(...)` while the rendered markdown was already alphabetized, so `library <pkg> --count -S @Performance` printed `Boxing, Arrays, Closures…` instead of `Allocation hotspots, Arrays, Async…`.

**Resolution (`86d92852`):** count-map order now built from `AlphabeticalSectionOrder.Where(IncludeSections.Contains)`, so the count map shares the single alphabetical order with the rendered sections. Added regression test `LibraryCommand_CountMap_RendersSectionsAlphabetically`.

**MAI-Code — fixed-head re-review (head `86d92852`): clean, no blocking issues.** Verified with real runs: original finding resolved (`--count -S @Performance` now alphabetical); single-section `--count` still works; `--count -S @All` alphabetical with no drops/dupes; non-count render paths (`@Performance`, `@SourceLink`, `@Surface`, `@Audit`, `-v:n`, `-v:d`) all alpha; build clean; `SectionPipelineTests` 127/127 and `CommandExecutionTests` green.

### Evidence
- Build: 0 warnings / 0 errors.
- Tests: `SectionPipelineTests` 127/0; `CommandExecutionTests` 581/0/2-skip. Full project 2182 total; the 8 unrelated failures (`PlatformResolverTests`, `ProgressiveMemberCallGraphTests` fixtures) reproduce identically on the stashed base tree — zero regressions from this change.
- Docs: alphabetical render-order invariant added to `docs/design/section-model.md` (markdownlint clean).

Ready to merge

richlander:
## Follow-up: keep `@Integrations` members behind their door in `-D`

Commit `6b3d6d30` — the ecosystem integration sections listed BOTH in the flat standalone `-D` catalog AND behind the `@Integrations` door. For `Microsoft.Extensions.AI`, `-D` showed `AI`, `Integrations`, and `Dependency Injection` as flat rows redundant with the door. Marked all 15 `@Integrations` door members `ListedInCatalog => false` — exactly like the kind-scoped `Performance:` sub-group — so they appear only behind the `@Integrations` door in `-D`, while still auto-rendering by size class (`-v:n`/`-v:d`), remaining exact-name-reachable, and still listed exhaustively under `-D --schema`. For a library with no integrations (e.g. System.Text.Json) the whole category is already hyper-subscribed out, so this is a no-op there.

### Adversarial review (risk-based tier)

Mechanical `-D`-catalog-only change following the established `ListedInCatalog` pattern, no network/security/correctness subtlety → **"more than trivial, but not high risk" → single review with MAI-Code**.

**MAI-Code (head `93410349`): clean, no blocking issues** — verified with real runs:

- **Completeness** — exactly the 15 `@Integrations` members (`LibraryIntegrationCatalog.CategorySections` + `Integration Opportunities`) marked; no non-integration section touched.
- **Flat catalog cleaned** — `package Microsoft.Extensions.AI --library -D` no longer lists `AI`/`Integrations`/`Dependency Injection` as standalone rows; the `@Integrations` door row remains.
- **Door + render unchanged** — `-D @Integrations` still lists members; `-S @Integrations -v:n` still renders them.
- **Schema still exhaustive** — `System.Text.Json -D --schema` still lists every integration section.
- **No-integration no-op** — `System.Text.Json -D` unchanged.
- **Tests** — `SectionPipelineTests` 127/127; the three integration-focused `CommandExecutionTests` green; the extended `LibraryPipeline_CatalogHiddenSections` asserts every `@Integrations` member is catalog-hidden.

### Evidence
- Build 0/0; `SectionPipelineTests` 127/0; `CommandExecutionTests` 581/0/2-skip.
- Doc note added to `docs/design/section-model.md` (markdownlint clean).

Ready to merge

richlander:
## Adversarial review — integration section `Integration:` prefix + rollup removal

Reviewed commit `8454011d` (behavioral output rename + section deletion across view/sections/commands/tests) with two different-model reviewers in isolated fixed-head worktrees.

**Reviewers:** Gemini 3.1 Pro and GPT-5.6 Sol.

### Findings and resolution

1. **`PackageCommand.cs:2073` — stale selectors in the `--all-libraries` category-selector error message** (Gemini + GPT, Medium). The row-output error still recommended the removed `Integrations` / bare `Configuration` selectors, so following the guidance produced a second error. **Fixed** in `fb21db73`: now recommends `"Integration: Configuration"` / `Library Info` (both verified to resolve).
2. **`LibraryFindingConsumerTests.cs:318` — obsolete unprefixed section name in a `FailureAffectsSection` assert** (GPT, Medium). The test passed `EcosystemIntegrationNames.AI` as the *affected section*, but `FailureAffectsSection` now matches `descriptor.SectionName`; the assert failed. **Fixed** in `fb21db73`: uses `IntegrationSectionNames.AI`.

### Explicitly validated as clean (no action)

- **Identity-vs-display split:** `descriptor.SectionName` governs `[MarkoutSection]`, `--all-libraries` generation, and `-S` selectors; `descriptor.Name` continues to govern finding/signal identity. Confirmed by both reviewers.
- **`FailureAffectsSection`:** the explicit `OpenTelemetry`-scanner-key → `Integration: OpenTelemetry` mapping is correct and does not double-count. (The generic ecosystem-scanner branch flagging OpenTelemetry is pre-existing behavior on base `93410349`, not introduced here.)
- **Rollup deletion:** no orphaned `-S Integrations` path; `--count -S @Integrations` reproduces the old per-integration counts.
- **Cache bump v17→v18** is sufficient to wall off stale `-D` section-schema cache.

### Re-review at fixed head `fb21db73`

Both reviewers re-reviewed the delta `8454011d..fb21db73`: **both report the fixed head is clean, no new issues.** LibraryFindingConsumerTests (11/11) pass and the recommended `-S "Integration: Configuration"` selector resolves.

Ready to merge

richlander:
## Adversarial review reconciliation — head `2fea9e86`

The naming/discoverability pass landed after this PR's first two clean reviews and
carries its own risk surface: user-visible renames with a compatibility story, a
behavioral change to `-S @SourceLink` cost, and a new construction-time validation
that throws. I treated that as the two-model tier per `AGENTS.md` and re-reviewed
from two families.

Reviewed at exact head `380fe250`, each isolated in its own linked review worktree.

| Reviewer | Verdict at `380fe250` | Verdict at fixed head `2fea9e86` |
| --- | --- | --- |
| GPT-5.6 Sol | No blocking findings; 1 non-blocking | Clean |
| MAI-Code | No blocking findings; 1 non-blocking | Clean |

### Findings and resolution

Both reviewers independently reported the same defect, which is useful corroboration:
the docs still placed `SourceLink: Integrity` behind `@Hidden` after it moved into
`@SourceLink`. GPT-5.6 Sol additionally found two spots MAI-Code did not.

| # | Finding | Reported by | Resolution |
| --- | --- | --- | --- |
| 1 | `docs/design/section-model.md:163` described `SourceLink: Integrity` as behind `@Hidden`, pointing readers at the wrong discovery selector | both | fixed in `1d99df9f` |
| 2 | `src/dotnet-inspect/Commands/PackageCommand.cs:38` justified the `@Hidden` render-rejection guard as blocking "the unbounded SourceLink: Integrity group" | GPT-5.6 Sol | fixed in `2fea9e86` |
| 3 | The footgun table's "Auto-renders at" cell said "never (exact name only)"; a door that roots an unbounded section does render it | GPT-5.6 Sol | fixed in `2fea9e86` |

While fixing #1 I found an adjacent error neither reviewer reported: the same table row
attributed `Unsafe Members` to no category when it is rooted in `@Audit`. Corrected in
`1d99df9f`.

All three findings were documentation/comment drift produced by this pass. No reviewer
found a functional defect, an incomplete rename, a broken alias, an unreachable section,
or a vacuous test.

### Verified, not merely read

Both reviewers built Release and ran the suite, and both confirmed the corrected table
row against the running tool rather than the source:

- `Unsafe Members` → `@Audit`
- `Top Leverage` → `@Hidden` only, so `—` is correct in the *topical door* column
- `SourceLink: Files` / `SourceLink: Integrity` → `@SourceLink`
- `-S @Hidden` still exits 1 as discovery-only

Both also verified the legacy aliases resolve to the new canonical headings, that
`Array Pool Escapes` is reachable by exact name / `@All` / `--schema` while absent from
bare `-D`, and that every pipeline constructs without the new `AddCategory` validation
throwing.

### On the local test failure

Both reviewers reported **2282 total, 0 failed, 14 skipped**. My local runs show one
failure, `PackagingSurfaceTests.OnlyTheShippingToolsOptIntoPackaging`. That test
enumerates `.csproj` files without excluding `.worktrees/`, so it only fails in a
checkout that keeps worktrees under the repo root — which is this repo's own documented
workflow. It passes in CI's clean clone and in the reviewers' worktrees. Pre-existing
from #3330 and unrelated to this PR; noted below as a follow-up.

### Non-blocking follow-ups

Filed from this pass:

- #3410 — `Non-normalized Paths` misses SourceLink document paths, and `HasNormalizedPaths` reports a false clean
- #3412 — IL coordinate resolution should delegate to the metadata API

Not yet filed, pending a maintainer call:

- `-S References` and `-S Dependencies` render "no data" because the underlying extraction is gated on options rather than selection (`LibraryMetadataService.cs:83`), and `Dependencies.ScannerKey` names a scanner that is never registered. Pre-existing on `main`, verified byte-identical there. It violates "keep failure visible" — a missing prerequisite surfaces as success-shaped empty output.
- `PackagingSurfaceTests` should exclude `.worktrees/` so it does not fail for contributors following this repo's worktree workflow.

richlander:
Ready to merge

All merge-blocking validation is complete at head `2fea9e86`:

- CI green: `test`, `build-net10`, `markdownlint`, `changes`. `pack`, `csharp-diff-smoke`, and `il-diff-smoke` are path-gated and correctly skipped.
- Two-model adversarial review (GPT-5.6 Sol, MAI-Code) clean on the fixed head, each in an isolated review worktree. Reconciliation in the comment above.
- Local Release suite: 2282 total, 14 skipped. The one local failure is `PackagingSurfaceTests.OnlyTheShippingToolsOptIntoPackaging`, which is a `.worktrees/` enumeration artifact pre-existing on `main`; it passes in CI and in both reviewers' trees.

Everything listed under "Non-blocking follow-ups" is explicitly out of scope for this PR.

richlander:
Follow-ups from this pass are now filed, so the "not yet filed" list in the reconciliation comment above is closed out:

- #3421 — `-S References` / `-S Dependencies` render "no data" because prerequisites are gated on options rather than selection
- #3422 — `PackagingSurfaceTests` fails locally because it scans `.worktrees/` copies of the shipping projects

Both are pre-existing on `main` and out of scope here; #3229 remains ready to merge.

Worth noting for whoever picks up #3421: its third root cause is that `ScannerRegistry.RunScanners` silently ignores a required scanner key with no registered scanner, so `Dependencies.ScannerKey => ScannerTransitiveRefs` is dead code that reads as working. That is the same class of defect this PR fixed for category membership, where `SectionPipeline.AddCategory` now throws on an unregistered section name. The registry deserves the counterpart guard.

Issue #3221, opened 2026-07-26
Raised inline-array collection expressions leave a dead unspellable buffer local that caps fidelity at Partial on older-runtime assemblies


richlander:
## Summary

Raised inline-array collection expressions leave a **dead, unreferenced**
compiler-synthesized inline-array buffer local (`<>y__InlineArrayN` on older
runtimes) in `function.Locals`. On assemblies compiled with a toolchain that
spells the buffer with the unspellable `<>y__InlineArrayN` form, that dead local
caps method fidelity at **Partial** via `DEC0009` even though the rendered C# is
fully valid and the collection expression raised correctly.

## Where it comes from

- `InlineArrayCollectionPass` raises the inline-array lowering to a
  `CollectionExpression` and detaches the `initobj`, the element stores, and the
  `AsReadOnlySpan` call. It does **not** remove the now-dead buffer local from
  `function.Locals` (there is no dead-local elimination; only iterator/async
  `ResetLocals`).
- The importer registers every IL local unconditionally
  (`IrImporter.cs`), so the buffer local survives.
- `CSharpSpellability.FunctionRenderedTypes` iterates **all** `function.Locals`
  unconditionally and reports a `TypeIssue` for the unspellable
  `<>y__InlineArrayN` name → `DEC0009 (signature) type name has no C# spelling`
  → fidelity degrades to Partial. The printer skips unreferenced locals so the
  name never renders, but fidelity still counts it.

## Toolchain split (why tests are green but real assemblies are Partial)

- Newer toolchain (.NET 11+, used by the repo's compiled fixtures) names the
  buffer `InlineArrayN` (spellable) → the dead local is harmless → **Full**.
- Older toolchain (e.g. `Microsoft.CodeAnalysis@5.0.0`) uses `<>y__InlineArrayN`
  (unspellable) → **Partial**.

So `CfgSampleClass.InlineArraySpan` /
`InlineArrayObjectConditionalElementSpan` reach Full in tests, while the real
witness `Microsoft.CodeAnalysis.Emit.EncLocalInfo::GetDebuggerDisplay` reaches
valid C# but stays Partial.

## Scope

Pre-existing; affects **both** the direct and the address-spilled element-store
paths. Not specific to the S4 raise — S4 (#3129) restores valid C# for the
address-spilled shape but inherits this Partial cap on older-runtime assemblies.

## Proposed fix (own slice/PR)

Remove the dead buffer local when its sole references are consumed by the raise
[...]

PR #3230, opened 2026-07-26
Raise address-spilled inline-array collection expressions (#3129, S4)

richlander:
- Advances #3129
- Raises csc's address-spilled inline-array collection-expression element stores (the `params ReadOnlySpan<T>` lowering where an element value carries a branch) back to a collection expression `[e0, …, eN]`
- Card revision: `d481b3b6`

> S0 (#3204) and S5 (#3216) have merged to `main`; this slice now targets `main` directly (single commit `d481b3b6`, rebased off the merged stack).

## Change

**Conclusion:** **PASS** — a previously-unraised, unspellable inline-array `params`-span shape (an element value carries a branch, so csc spills the element-ref address + value to eval-stack slots and stores through the spilled address) now raises to a valid, correct collection expression, guarded so it fires only on the exact single-use spill shape and leaves every ambiguous shape flat.

### Benchmark target

Benchmark target: `Microsoft.CodeAnalysis@5.0.0` (real witness `Microsoft.CodeAnalysis.Emit.EncLocalInfo::GetDebuggerDisplay`).

dotnet-inspect command:

```bash
dotnet-inspect member Microsoft.CodeAnalysis.Emit.EncLocalInfo GetDebuggerDisplay --library Microsoft.CodeAnalysis.dll --all -S "Decompiled Source"
```

### Original source

```csharp
private string GetDebuggerDisplay()
{
    if (IsDefault)
    {
        return "[default]";
    }

    if (IsUnused)
    {
        return "[invalid]";
    }

    return string.Format("[Id={0}, SynthesizedKind={1}, Type={2}, Constraints={3}, Sig={4}]",
        SlotInfo.Id,
        SlotInfo.SynthesizedKind,
        Type,
        Constraints,
        (Signature != null) ? BitConverter.ToString(Signature) : "null");
}
```

The 5-argument `string.Format` binds to the `params ReadOnlySpan<object>` overload, which csc lowers to an `<>y__InlineArray5<object>` buffer. Element 4 carries a ternary, so csc evaluates the element-ref **address** before the branch and spills both the address and the branch value to evaluation-stack slots.

### Before

```csharp
private string GetDebuggerDisplay()
{
    if (IsDefault)
    {
        return "[default]";
    }
    if (IsUnused)
    {
        return "[invalid]";
    }
    ___y__InlineArray5<object> V_0 = default;
    V_0[0] = SlotInfo.Id;
    V_0[1] = SlotInfo.SynthesizedKind;
    V_0[2] = Type;
    V_0[3] = Constraints;
    ref object S_1 = ref V_0[4];
    string S_2 = Signature is not null ? BitConverter.ToString(Signature) : "null";
    S_1 = S_2;
    return string.Format("[Id={0}, SynthesizedKind={1}, Type={2}, Constraints={3}, Sig={4}]", __PrivateImplementationDetails_.InlineArrayAsReadOnlySpan<___y__InlineArray5<object>, object>(V_0, 5));
}
```

- Valid: False — parses (S0 sanitized the `<>` names to `___…`) but does not bind: `___y__InlineArray5` and `__PrivateImplementationDetails_` are not real types (CS0246).
- Correct: False — cannot bind, so no observable behavior.
- IL fidelity: False — does not compile.
- Taste applied: None.
- Commit: `08ab86b1` (`main`; pre-S4 base)

The spilled element 4 (`ref S_1 = ref V_0[4]; … S_1 = S_2;`) fails the pass's direct-store match, so the **entire** inline-array shape is left flat and honestly `Partial`.

### After

```csharp
private string GetDebuggerDisplay()
{
    if (IsDefault)
    {
        return "[default]";
    }
    if (IsUnused)
    {
        return "[invalid]";
    }
    return string.Format("[Id={0}, SynthesizedKind={1}, Type={2}, Constraints={3}, Sig={4}]", [SlotInfo.Id, SlotInfo.SynthesizedKind, Type, Constraints, Signature is not null ? BitConverter.ToString(Signature) : "null"]);
}
```

- Valid: True — `[…]` target-types to `string.Format`'s `ReadOnlySpan<object>` parameter and binds.
- Correct: True — a 5-element span in source order, identical to the lowering.
- IL fidelity: Full on the compiled fixture `InlineArrayObjectConditionalElementSpan` (the identical address+value-spilled shape; the raised text equals source and it is absent from the compile-back docket diff set). On this older-Roslyn witness the render is valid and correct but stays **Partial** solely because a residual dead, unreferenced `<>y__InlineArray5` buffer local remains in the locals table with an unspellable type name (DEC0009) — a separate concern tracked in #3221, not this raise.
- Taste applied: None (the `!= null` → `is not null` normalization is default rendering and appears in both Before and After).
- Commit: `d481b3b6`

### Fully raised

The After decompilation renders the fully raised collection expression `[…]`. On this older-Roslyn witness fidelity stays `Partial` only because of the residual dead inline-array buffer local (`<>y__InlineArray5`, unspellable): the buffer is no longer referenced in the raised body but remains in the method's locals table, and `CSharpSpellability` iterates all locals. That dead-local cleanup is a separate slice (it also affects the pre-existing direct path).

- Required tracking issue: #3221 — remove/relax the residual dead inline-array buffer local so older-Roslyn witnesses reach Full.

## Implementation

Extends `InlineArrayCollectionPass` (the single owner of the inline-array `CollectionExpression` lowering). Previously the pass recognized only a **direct** element store `*(InlineArrayElementRef(ref buf, i)) = value`. When an element value carries a branch, csc must evaluate the element-ref address before the branch, so it spills:

```text
S_a = InlineArrayElementRef(ref buf, i);   // address spilled to an eval-stack slot
[S_v = value;]                             // branch value spilled to a second slot
*S_a = S_v-or-value;                       // StoreIndirect through the spilled address
```

New logic:

- `CollectElementStores` now returns `List<InlineArrayElementStore>?`; `ResolveElementStore` classifies each element as either a direct non-volatile `StoreIndirect`, or an address spilled to a **single-use** eval-stack slot then stored through, recovering a single-use value spill.
- Spill recovery fires only when every spill slot is written **once** and read **once** (`StackSlotStoreCount`/`StackSlotLoadCount` == 1); a multi-use slot, a volatile store, or an unrecovered stack-slot value returns `null` and the **whole** shape is left flat (honest `Partial`).
- Elements are re-sequenced into slot (source) order.
- The existing over-raise guards (exact synthesized `<>y__InlineArrayN`/`InlineArrayN` metadata name, `count + 2` element-ref count, no residual `LoadLocal` of the buffer, unique init) are unchanged, so the runtime's own `params` buffers and user structs stay excluded.

The two new fixture methods are appended at the **end** of `CfgSampleClass` so no existing method's compiler-generated ordinal shifts (the fidelity docket keys on those ordinals).

## Evidence

| Check | Baseline (`08ab86b1`) | Head (`d481b3b6`) |
| --- | --- | --- |
| Product build (`dotnet-inspect.slnx`) | Pass | Pass |
| Focused tests | new (n/a) | `InlineArraySpilledElementTests` (4) + `IrImporterTests` spilled-store import (1): 5 passed |
| Decompiler fast suite (`-trait- Speed=Slow`) | 3781, 1 failed | 3786, 1 failed |
| Reduced fixture fidelity | n/a | `InlineArrayObjectConditionalElementSpan` **Full** (opcode-exact) |
| Real witness | `EncLocalInfo::GetDebuggerDisplay` invalid (flat inline-array) | `EncLocalInfo::GetDebuggerDisplay` valid `[…]` (Partial only via dead-local #3221) |

The single fast-suite failure is the pre-existing `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString` (CRLF-vs-LF, #3018) — identical on baseline and head, unrelated to this change. The `+5` are this PR's new tests.

## Decompiler quality

**Conclusion:** **ADVISORY** — the full PR corpus quick-gate was not run for this focused slice. The change is additive and tightly guarded: it only recovers a previously-residual, spilled inline-array element shape into an already-supported collection expression, gated on single-use spill slots; no existing raise path is modified, so no previously-raised method can regress and residue can only decrease or hold.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class "ILInspector.Decompiler.Tests.InlineArraySpilledElementTests"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -4958,6 +4958,27 @@ public class CfgSampleClass
                 return -1;
         }
     }
+
+    static int UseObjectSpan(System.ReadOnlySpan<object> s) => s.Length;
+
+    // A collection expression whose element VALUE carries a branch AND a call:
+    // `s is not null ? s.ToUpper() : "null"` in a params ReadOnlySpan<object>
+    // context. csc cannot evaluate that element in one push, so it computes the
+    // element-ref ADDRESS first and spills it to an evaluation-stack temp, then
+    // spills the conditional's value to a second temp, and finally stores through
+    // the spilled address — `ref object A = ref InlineArrayElementRef(ref buffer,
+    // 1); object V = s is not null ? s.ToUpper() : "null"; A = V;`. The `object`
+    // element type (boxing) and the method call keep that double spill alive all
+    // the way to InlineArrayCollectionPass, unlike a scalar `flag ? 1 : 2` whose
+    // spill an earlier ternary fold collapses to a direct store. This is the shape
+    // EncLocalInfo's params `string.Format` hits in the wild (issue #3129, S4):
+    // the pass recovers the spilled address+value and raises the whole thing back
+    // to `[a, s is not null ? s.ToUpper() : "null"]`, re-sequencing into slot
+    // (= source) order. Placed last in the class so its two extra methods do not
+    // shift any pre-existing method's compiler-generated ordinal (the fidelity
+    // docket keys on `<>9__N`/`<>c__DisplayClassN` names — see issue #3129, S4).
+    public static int InlineArrayObjectConditionalElementSpan(int a, string? s)
+        => UseObjectSpan([a, s is not null ? s.ToUpper() : "null"]);
 }
 
 internal static class AwaitOrderingHelpers
--- a/src/ILInspector.Decompiler.Tests/InlineArraySpilledElementTests.cs
+++ b/src/ILInspector.Decompiler.Tests/InlineArraySpilledElementTests.cs
@@ -186,6 +186,55 @@ public class InlineArraySpilledElementTests
         function.CheckInvariant();
     }
 
+    [Fact]
+    public void NullCoalescingAssignmentValueSpan_StaysFlat()
+    {
+        // The span is the right operand of `local ??= span`, evaluated only when the
+        // target is null. `NullCoalescingAssignment` is one of the conditional IR
+        // nodes an enumerated blacklist missed; the sound-by-default whitelist leaves
+        // it flat because the node is not a recognized unconditional container.
+        var function = BuildOrderingCollection(OrderingShape.NullCoalescingAssignmentValue);
+
+        new InlineArrayCollectionPass().Run(function, PassContext.None);
+
+        Assert.Empty(function.Descendants.OfType<CollectionExpression>());
+        Assert.Contains(function.Descendants.OfType<Call>(), c => c.Callee.Name == "InlineArrayAsReadOnlySpan");
+        function.CheckInvariant();
+    }
+
+    [Fact]
+    public void UnionSwitchArmValueSpan_StaysFlat()
+    {
+        // The span is a union-switch-expression arm value, evaluated only when its arm
+        // matches. `UnionSwitchExpressionArm` does not derive from `SwitchExpressionArm`,
+        // so a blacklist keyed on the base arm type missed it; the whitelist rejects any
+        // parent it does not explicitly recognize as unconditional, so it stays flat.
+        var function = BuildOrderingCollection(OrderingShape.UnionSwitchArmValue);
+
+        new InlineArrayCollectionPass().Run(function, PassContext.None);
+
+        Assert.Empty(function.Descendants.OfType<CollectionExpression>());
+        Assert.Contains(function.Descendants.OfType<Call>(), c => c.Callee.Name == "InlineArrayAsReadOnlySpan");
+        function.CheckInvariant();
+    }
+
+    [Fact]
+    public void WhileConditionSpan_StaysFlat()
+    {
+        // The span is a while-loop condition: evaluated unconditionally but once per
+        // iteration. Lifting the element stores into it would call each element
+        // function once per loop test instead of once. The whitelist rejects the
+        // WhileLoop condition edge (not a once-through container), so it stays flat —
+        // covering the exactly-once axis, not just definite evaluation.
+        var function = BuildOrderingCollection(OrderingShape.WhileConditionSpan);
+
+        new InlineArrayCollectionPass().Run(function, PassContext.None);
+
+        Assert.Empty(function.Descendants.OfType<CollectionExpression>());
+        Assert.Contains(function.Descendants.OfType<Call>(), c => c.Callee.Name == "InlineArrayAsReadOnlySpan");
+        function.CheckInvariant();
+    }
+
     enum OrderingShape
     {
         InterleavedStatement,
@@ -193,6 +242,9 @@ public class InlineArraySpilledElementTests
         InlineEffectBeforeSpan,
         SpilledPrefixBeforeSpan,
         ConditionalSpanArm,
+        NullCoalescingAssignmentValue,
+        UnionSwitchArmValue,
+        WhileConditionSpan,
     }
 
     /// <summary>
@@ -273,6 +325,33 @@ public class InlineArraySpilledElementTests
                             span,
                             new LoadLocal(1, SpanObject))])));
                 break;
+            case OrderingShape.NullCoalescingAssignmentValue:
+                // local ??= span: the span is the ??= right operand, evaluated only
+                // when the target is null. NothingEffectfulBefore passes (the span is
+                // the only child), so only the definite-evaluation whitelist rejects
+                // this — the missed conditional node from adversarial review.
+                block.Add(new NullCoalescingAssignment(1, SpanObject, span));
+                break;
+            case OrderingShape.UnionSwitchArmValue:
+                // local = scrutinee switch { Case => span, ... }: the span is a switch
+                // arm value, evaluated only when its arm matches. A union-switch arm is
+                // a distinct IR node from SwitchExpressionArm, so the whitelist (not an
+                // enumerated blacklist) must still reject it.
+                block.Add(new StoreLocal(1, SpanObject,
+                    new UnionSwitchExpression(
+                        new LoadArgument(0, "arg", Object),
+                        [new UnionSwitchExpressionArm(Object, null, span)])));
+                break;
+            case OrderingShape.WhileConditionSpan:
+                // while (Predicate(span)) { }: the span is unconditionally evaluated
+                // but re-evaluated on every loop iteration. Definite evaluation alone
+                // is not enough — the span must also be evaluated exactly once, so the
+                // whitelist rejects a WhileLoop condition edge (it is not a recognized
+                // once-through container), leaving the collection flat.
+                block.Add(new WhileLoop(
+                    new Call(PredicateSpanMethod(), isVirtual: false, [span]),
+                    new Block()));
+                break;
             default:
                 // ReadOnlySpan<object> span = InlineArrayAsReadOnlySpan(ref buffer, 2); (local 1)
                 block.Add(new StoreLocal(1, SpanObject, span));
@@ -333,6 +412,15 @@ public class InlineArraySpilledElementTests
             TypeRef.Definition("Synthetic", "", "Fx"),
             "Consume",
             Void,
+            [SpanObject],
+            HasThis: false);
+
+    // bool Predicate(ReadOnlySpan<object> span)
+    static MethodRef PredicateSpanMethod()
+        => new(
+            TypeRef.Definition("Synthetic", "", "Fx"),
+            "Predicate",
+            Bool,
             [SpanObject],
             HasThis: false);
 
--- a/src/ILInspector.Decompiler.Tests/InlineArraySpilledElementTests.cs
+++ b/src/ILInspector.Decompiler.Tests/InlineArraySpilledElementTests.cs
@@ -169,12 +169,30 @@ public class InlineArraySpilledElementTests
         function.CheckInvariant();
     }
 
+    [Fact]
+    public void ConditionalSpanArmInConsumer_StaysFlat()
+    {
+        // The span sits in a ternary arm (Consume(cond ? span : other)), so it is only
+        // conditionally evaluated. The element stores are unconditional; lifting them
+        // into that arm would suppress their side effects whenever the false arm is
+        // taken. Block order is canonical, so only the conditional guard rejects it —
+        // the shape must stay flat.
+        var function = BuildOrderingCollection(OrderingShape.ConditionalSpanArm);
+
+        new InlineArrayCollectionPass().Run(function, PassContext.None);
+
+        Assert.Empty(function.Descendants.OfType<CollectionExpression>());
+        Assert.Contains(function.Descendants.OfType<Call>(), c => c.Callee.Name == "InlineArrayAsReadOnlySpan");
+        function.CheckInvariant();
+    }
+
     enum OrderingShape
     {
         InterleavedStatement,
         DescendingSlotOrder,
         InlineEffectBeforeSpan,
         SpilledPrefixBeforeSpan,
+        ConditionalSpanArm,
     }
 
     /// <summary>
@@ -244,6 +262,17 @@ public class InlineArraySpilledElementTests
                 block.Add(new ExpressionStatement(
                     new Call(ConsumeMethod(), isVirtual: false, [new LoadStackSlot(0, Object), span])));
                 break;
+            case OrderingShape.ConditionalSpanArm:
+                // Consume(cond ? span : other): the span is in a ternary arm, so it
+                // is only conditionally evaluated. Lifting the unconditional element
+                // stores into that arm would drop their effects on the false path.
+                block.Add(new ExpressionStatement(
+                    new Call(ConsumeSpanMethod(), isVirtual: false, [
+                        new Conditional(
+                            new Constant(true, Bool),
+                            span,
+                            new LoadLocal(1, SpanObject))])));
+                break;
             default:
                 // ReadOnlySpan<object> span = InlineArrayAsReadOnlySpan(ref buffer, 2); (local 1)
                 block.Add(new StoreLocal(1, SpanObject, span));
@@ -296,6 +325,15 @@ public class InlineArraySpilledElementTests
             "Consume",
             Void,
             [Object, SpanObject],
+            HasThis: false);
+
+    // void Consume(ReadOnlySpan<object> span)
+    static MethodRef ConsumeSpanMethod()
+        => new(
+            TypeRef.Definition("Synthetic", "", "Fx"),
+            "Consume",
+            Void,
+            [SpanObject],
             HasThis: false);
 
     /// <summary>
--- a/src/ILInspector.Decompiler.Tests/InlineArraySpilledElementTests.cs
+++ b/src/ILInspector.Decompiler.Tests/InlineArraySpilledElementTests.cs
@@ -1,0 +1,202 @@
+using ILInspector.Decompiler.Pipeline;
+
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// Synthetic coverage for <see cref="InlineArrayCollectionPass"/>'s address-spilled
+/// element-store recovery (issue #3129, S4). When a collection-expression element
+/// value carries a branch, csc evaluates the element-ref address before the branch
+/// and spills the <c>ref</c> — and often the branch value — to single-use
+/// evaluation-stack slots, then stores through the spilled address. The pass
+/// recovers that shape only when every spill slot is written once and read once; a
+/// multi-use spill slot or a volatile store must leave the whole collection flat.
+/// The compiled-fixture canary lives in <c>CfgSampleClass.InlineArrayObjectConditionalElementSpan</c>;
+/// these tests pin the guard boundaries a compiled fixture cannot express.
+/// </summary>
+[Trait("Area", "Pass")]
+public class InlineArraySpilledElementTests
+{
+    static readonly TypeRef Int32 = TypeRef.CoreLib("System", "Int32");
+    static readonly TypeRef Object = TypeRef.CoreLib("System", "Object");
+    static readonly TypeRef Bool = TypeRef.CoreLib("System", "Boolean");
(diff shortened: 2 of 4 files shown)


richlander:

Two non-Claude models reviewed the exact commit **`f79827c9`** (base = the then-S5 head `e56c9dfd`), each isolated in its own read-only worktree with scratch confined to `%TEMP%` and git mutations prohibited. Both ran the real evidence (build, the 5 new tests, the compiled fixture, and the real `EncLocalInfo::GetDebuggerDisplay` witness), not just a read.

| Reviewer | Model | Verdict | Findings |
| --- | --- | --- | --- |
| Gemini | `gemini-3.1-pro-preview` | **CLEAN — safe to merge** | none |
| GPT | `gpt-5.6-sol` | **CLEAN — safe to merge** | none |

Both reviewers explicitly verified all seven attack points as SAFE, with agreeing evidence:

1. **Spill guards** — `StackSlotStoreCount`/`StackSlotLoadCount` require exactly one write and one read per spill slot; reuse across elements or elsewhere rejects the whole raise, and the `Single()` calls are provably guarded by the `!= 1` checks.
2. **Detach ordering / invariants** — values detach once; consumed statements detach cleanly; single-use slots guarantee no overlap; `CheckInvariant()` holds across the synthetic and real harness pipelines.
3. **False positives** — the unchanged synthesized-name gate, exact `count + 2` element-ref count, no residual `LoadLocal`, and unique-init guards still exclude the runtime's own `params` buffers and user structs.
4. **Element ordering** — the `SortedDictionary` restores slot (source) order; the mixed fixture (direct element 0, spilled element 1) confirmed it.
5. **Volatile** — both the direct and spilled arms require `IsVolatile: false`, so volatile stores stay flat.
6. **Value subtree** — only an immediate single-use `LoadStackSlot` consumes its definition; nested expressions and unrelated spill defs are preserved intact.
7. **Fixture ordinal-neutrality** — appending the two methods at class end shifts no existing method's compiler-generated ordinal.

Evidence both ran: test-project build 0 errors; the 5 new tests `Total 5, Failed 0`; the compiled fixture reaches `fidelity Full` and emits `return UseObjectSpan([a, s is not null ? s.ToUpper() : "null"]);`; the real witness raises to a valid `string.Format(…, [ … ])` with `HAS_PRIVATE_IMPL=False` and `HAS_RAW_STACK_SLOT=False`, staying `Partial` solely for the expected `<>y__InlineArray5` dead-local (#3221).

### Rebase note (reviewed head → PR head)

Between review and PR, S0 (#3204) and S5 (#3216) squash-merged to `main`, collapsing the stack. The reviewed commit `f79827c9` was rebased `--onto origin/main` (dropping the now-merged S5 commits) to **`d481b3b6`**. The tree delta is identical (same 4 files, +365/−26); the only rebase conflict was in `CfgSampleClass.cs`, where `main` had independently appended `SwitchWithTwoLoopingCaseSections` at class end. Resolution kept both, with the two S4 fixtures **still last** — so the ordinal-neutrality property the reviewers verified is preserved.

Re-validated on `d481b3b6` (no re-review needed, as the reviewed content is byte-preserved):

- `dotnet build dotnet-inspect.slnx -c Release` — clean (0 errors).
- 5 new tests — `InlineArraySpilledElementTests` 4/4 + `IrImporterTests` spilled-store import 1/1.
- Real witness render — **byte-identical** to the reviewed output (raised collection expression, no `<PrivateImplementationDetails>`, no raw stack slots).
- Fast suite (`-trait- Speed=Slow`) — 3786 total, 1 failed (the pre-existing CRLF `#3018`, unchanged).

Both fixed-head reviews are clean; no findings to resolve.

richlander:
Ready to merge — both adversarial reviews (Gemini `gemini-3.1-pro-preview`, GPT `gpt-5.6-sol`) clean on the reviewed content, and all merge-blocking CI is green (build-net10, csharp-diff-smoke, test/ubuntu). The only follow-up is the residual dead inline-array buffer local tracked in #3221 (non-blocking).

richlander:
Reviewed at `d481b3b6` in a clean worktree, built `-c Release`. The spill-recovery logic is careful and the three negative guards (`AddressSpillReadTwice`, `ValueSpillReadTwice`, `VolatileSpilledStore`) are well-chosen — each pins a real aliasing hazard, and reverting the production file fails the new tests, so they're non-vacuous end-to-end (`IrImporterTests.InlineArrayObjectConditionalElementSpan_RaisesSpilledStore` drops `Full` → `Partial`).

**One guard is missing, and it's the one property the raise inherently changes: statement order.** Requesting that before merge.

## The gap

`CollectElementStores` (`InlineArrayCollectionPass.cs:768-791`) scrapes element refs with `function.Descendants.OfType<Call>()` — the whole function — and keys them into a `SortedDictionary<int, …>`. It verifies *count*, *slot distinctness*, and *slot range*. It never verifies that the stores are **contiguous**, **in ascending slot order**, or even **in the same block**: `ResolveElementStore` requires only `Parent: Block`, not the same `Block`.

The elements are then emitted at the position of the `AsSpan` call, in slot order. So any statement that sat *between* the stores is now sequenced *before* every element, and stores written in a different order are silently re-sequenced.

The existing test comment states the assumption outright rather than checking it:

> Elements are re-sequenced into slot (= source) order

Slot order equals source order for csc-emitted collection expressions. The pass accepts far more than csc emits.

## Measured

I added two synthetic fixtures mirroring `BuildSpilledCollection`, with side-effecting calls (`Fx::A`, `Fx::B`, `Fx::Sep`) as element values so ordering is observable:

```
[CASE1] raised=True  IL order = A, Sep, B   →  resulting order = Sep, A, B
[CASE2] raised=True  IL order = B, A        →  collection elements = A, B
```

- **Case 1** — an independent `Sep()` statement between the two element stores. The pass raises, and `Sep()` now runs *before* `A()`, which originally ran first.
- **Case 2** — element stores in descending slot order. The pass raises and **inverts** the evaluation order of `A()` and `B()`.

Both are accepted with no bail. Neither produces a diagnostic; the output is confidently wrong C#.

## Reachability, stated honestly

I did **not** find a csc shape that produces either. csc emits element stores contiguously in ascending slot order, and the two-collection-expressions-sharing-a-buffer case degrades safely (`elementRefs.Count != count` → bail). So this is not a regression against the C# compiler, and I'm not claiming the corpus will move.

It matters anyway, for the reason the decompiler exists: it consumes arbitrary IL — ilasm, obfuscators, other front ends — and this is a *silent* miscompile rather than a `Partial`. `AGENTS.md` also asks for close negative cases paired with each new discriminator; the spill guards have them and the ordering assumption doesn't.

## Suggested fix

Cheap and local: after `CollectElementStores` succeeds, require that the union of `store.Consumed` plus the `init` forms a **contiguous run of statements in one block, ending at or before the `AsSpan` statement, with slot indices ascending across that run**. Anything else bails to flat. That is exactly the csc shape and rejects both cases above.

Then convert my two fixtures into the paired negative tests (happy to hand them over — they're a straight copy of `BuildSpilledCollection` with side-effecting element values).

## Verified

- `InlineArraySpilledElementTests`: 4/4 pass.
- `IrImporterTests`: 66/66 pass.
- Revert experiment: reverting `InlineArrayCollectionPass.cs` to `main` fails `SpilledAddressAndValue_RaisesToCollectionExpression` **and** the end-to-end `IrImporterTests` case — tests are non-vacuous and do assert on real decompiled output, not just internal helpers.

richlander:
Addressed in `0b217281` (follow-up commit, not amended).

Thanks — the ordering gap is real and now guarded. `ElementStoresPreserveOrder` runs after `CollectElementStores` succeeds and, before any mutation, requires that the `init` and every store's consumed statements form a **contiguous run in the span-consuming statement's own block, ending immediately before that statement, with slot indices non-decreasing across the run** (the `init` sorts first with a sentinel slot). That is exactly the csc shape — default the buffer, fill slots `0..N-1` in order, read the span — and anything else is left flat.

The two hazards you measured are now the paired negative tests, modeled on the existing `BuildSpilledCollection` synthetic shape:

- `InterleavedStatementBetweenStores_StaysFlat` — an unrelated side-effecting statement between the two element stores (your Case 1). Contiguity check rejects it.
- `DescendingSlotOrderStores_StaysFlat` — element stores emitted in descending slot order (your Case 2). Non-decreasing-slot check rejects it.

Both **raise a wrong `CollectionExpression` with the guard reverted and stay flat with it in place**, so they're non-vacuous. The `SpilledAddressAndValue` positive case and the real compiled `InlineArrayObjectConditionalElementSpan` canary still raise.

Evidence at `0b217281`:

- `InlineArraySpilledElementTests`: 6/6 pass (4 prior + 2 new).
- `IrImporterTests`: 66/66 pass (includes the end-to-end canary).
- Decompiler fast suite: 3788 total, 1 failed — only the pre-existing CRLF `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString` (#3018), unchanged by this commit.

This shape is csc-conservative (it only ever *rejects* raises, never changes a raised render), so no corpus movement is expected. Dispatching adversarial review of the guard on `0b217281` now; will reconcile here when both are clean.

richlander:
## Adversarial re-review of `ddd3a44b` — both models sound-clean

The fidelity-recovery follow-up `ddd3a44b` was re-reviewed on its exact head by
**Gemini (`gemini-3.1-pro-preview`)** and **GPT (`gpt-5.6-sol`)** in isolated
read-only worktrees. Both confirm the guard is **sound**; the only findings are
non-blocking safe-Partial fidelity opportunities, deferred to #3281.

### What changed since the last round (`6dff4c48` → `ddd3a44b`)

`6dff4c48` made `EvaluatesChildUnconditionally` a sound-by-default whitelist,
which closed all the conditional-evaluation soundness holes but over-tightened:
params-span consumers whose *result* nests in an unconditional operator or a
forward selection statement regressed to Partial. `ddd3a44b` broadens the
whitelist with two provably-evaluated-exactly-once families while staying sound
by default:

- Non-short-circuiting operator expressions (whole node): `Binary`, `Comparison`,
  `Unary`, `LogicalNot`, `TupleExpression`, `TupleBinaryExpression`, `Throw`,
  `YieldReturn`. The short-circuiting forms (`LogicalBinary` `&&`/`||`, `Coalesce`
  `??`, `Conditional` `?:`, `NullConditional` `?.`) are **not** included.
- Forward **selection statement** scrutinee edges, matched per-edge with
  `ReferenceEquals`: `IfStatement.Condition`, `Switch.Value` (the arms/sections
  are rejected).

Loops (`WhileLoop`/`ForLoop`/`DoWhileLoop`/`ForeachStatement`) and the low-level
`ConditionalBranch`/`SwitchBranch` stay excluded. This pass runs at `IrPass.cs:343`,
after `DoWhileLoopPass`/`StructuringPass`/`ForLoopPass`/`ForeachStatementPass`/
`SwitchRaisingPass`, so an `if`/`switch` statement here is never a loop back-edge;
its scrutinee runs exactly once, and the block-level ordering guard ties the
element stores to the consumer's block so any re-execution re-runs them together.

### Findings (both models)

| # | Reviewer | Severity | Type | Shape left flat |
|---|----------|----------|------|-----------------|
| 1 | GPT | Medium | Fidelity (safe Partial) | `Conditional.Condition` — `Pred(span) ? a : b` |
| 2 | Gemini | Medium | Fidelity (safe Partial) | `SwitchExpression`/`UnionSwitchExpression` scrutinees — `M(span) switch { ... }` |
| 3 | Gemini | Medium | Fidelity (safe Partial) | `ForeachStatement.Collection`, `ForLoop.Initializer`, `UsingStatement.Resource`, `Lock.LockObject` |

**Soundness:** both models attacked every newly-whitelisted node and found no
false-accept. Gemini verified (via a compiled probe) that tuple equality evaluates
all operands before comparing, that `Binary`/`Comparison`/`Unary`/`Throw`/
`YieldReturn` do not short-circuit, and that a thrown exception from an earlier
operand is shielded by the existing `NothingEffectfulBefore` guard. GPT confirmed
the per-edge scrutinee matches reject the conditional arms and that the span is
never evaluated a different number of times than the element stores.

**Resolution:** all three findings are the *same* per-edge run-once pattern already
used for `IfStatement.Condition`/`Switch.Value`, applied to run-once *expression*
scrutinees and loop/resource sub-expressions. They are safe Partials today (the
peer `StackAllocSpanPass` leaves the same shapes flat), so they are **deferred to
#3281** rather than expanding this slice's soundness surface further. No finding is
a regression or an unsound raise.

### Evidence at `ddd3a44b`

- Full solution builds clean.
- `InlineArraySpilledElementTests`: 16/16 (4 new positives — Binary operand,
  if-condition, switch-value, throw-value consumers all raise — plus the 4
  conditional negatives). Both non-vacuity directions proven: flipping the default
  to `true` wrongly raises the 4 negatives; removing the new whitelist entries
  leaves the 4 positives flat.
- Fast decompiler suite: 3798 total, only the pre-existing CRLF failure
  `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString` (#3018).
- CI on `ddd3a44b`: `test`, `build-net10`, `csharp-diff-smoke`, `changes` all
  green; path-gated jobs skipped.
- GPT independently ran build + focused (16/16) + 2 compiled fixtures + fast suite
  (3798, only #3018) on its review head.

richlander:
Ready to merge — both fixed-head adversarial reviews (Gemini + GPT) on `ddd3a44b` are sound-clean, CI is green, and the branch is mergeable/clean. The only review findings are non-blocking safe-Partial fidelity edges, deferred to #3281.

richlander:
Re-reviewed at fixed head `ddd3a44b` in a clean worktree. **My blocking finding is closed. Approving.**

I verified the guard independently rather than re-running your tests. Neutralizing `ElementStoresPreserveOrder` (forcing it to `return true`) and rebuilding fails **7 of 7** negative tests:

```
InterleavedStatementBetweenStores_StaysFlat   [FAIL]   <- my Case 1
DescendingSlotOrderStores_StaysFlat           [FAIL]   <- my Case 2
InlineEffectBeforeSpanInConsumer_StaysFlat    [FAIL]
ConditionalSpanArmInConsumer_StaysFlat        [FAIL]
NullCoalescingAssignmentValueSpan_StaysFlat   [FAIL]
UnionSwitchArmValueSpan_StaysFlat             [FAIL]
WhileConditionSpan_StaysFlat                  [FAIL]
```

So both of my hazards are pinned by tests that genuinely fail without the guard, and the guard is doing work on five more shapes besides. That is the right outcome — and finding the conditional-evaluation family while fixing the ordering family is a better result than the original report.

`ElementStoresPreserveOrder` is stronger than what I asked for. I asked for a contiguous ascending run in the consumer's block; you also tied it to `NothingEffectfulBefore` and `SpanUnconditionallyEvaluated`, which closes the "block order is canonical but the consumer evaluates an effect inline before the span" case I had not considered.

**On the allow list**, `EvaluatesChildUnconditionally` being an allow list rather than a deny list is the correct call, and the doc comment's justification is exactly right: an unrecognized shape costs fidelity, a missed conditional shape emits confidently-wrong C#, so the default must be `false`. That asymmetry is the whole argument and it's stated plainly.

I independently checked the subtlest entry, `TupleBinaryExpression`, since tuple equality *does* short-circuit its comparisons and it would be easy to conclude the operands short-circuit too. Compiled probe on the running SDK:

```
(1,2) == (9,9)  [first comparison FALSE]:  eval a, eval b, eval c, eval d
(1,2) != (9,9):                            eval a, eval b, eval c, eval d
nested (1,(2,3)) == (9,(9,9)):             eval a..f  (all six)
```

All operands evaluate, including nested, even when the first comparison already decides the result. Allow-listing it is sound. Gemini's probe reproduced.

Also confirmed the ordering argument for the `IfStatement.Condition` / `Switch.Value` per-edge entries: this pass runs at `IrPass.cs:343`, after the loop- and switch-raising passes, so an `if`/`switch` node here cannot be a loop back-edge, and the block-level ordering guard ties the element stores to the consumer's block so any re-execution re-runs them together.

Fast decompiler suite at the stack tip: **3803 passed, 0 failed**.

One finding on the child PR #3285, which builds on this pass — details there. It does not affect this PR's raise, which is sound.

richlander:
Terminology request, separate from the review: please use **allow list** / **deny list** rather than whitelist/blacklist.

This stack introduces 24 new occurrences:

```
src/ILInspector.Decompiler/Pipeline/Passes/InlineArrayCollectionPass.cs   8
src/ILInspector.Decompiler.Tests/InlineArraySpilledElementTests.cs      16
```

Mostly in the `EvaluatesChildUnconditionally` doc comment and the ordering-guard test comments. It's a pure comment/prose change — no identifiers are affected, since the API is already named `EvaluatesChildUnconditionally` rather than after the list.

I've updated the wording in my own review comments above.

There are 5 pre-existing occurrences on `main` (`ThisQualificationTests.cs` ×4, `OuterTypePatternDispatchRaisingTests.cs` ×1) — out of scope here; worth a separate sweep so the whole repo lands on one term.

richlander:
## Terminology sweep done — `bcb9e48e`

Swept whitelist/blacklist → **allow list** / **deny list** across all 24 new occurrences, matching your counts exactly:

```
src/ILInspector.Decompiler/Pipeline/Passes/InlineArrayCollectionPass.cs   8
src/ILInspector.Decompiler.Tests/InlineArraySpilledElementTests.cs      16
```

Pure comment/prose change — no identifiers touched (the API stays `EvaluatesChildUnconditionally`), no logic or behavior change. `git show --stat`: 2 files, 24 insertions(+), 24 deletions(-). Verified 0 `whitelist`/`blacklist` tokens remain in either file; the decompiler test project builds clean (0 warnings, 0 errors).

Wording notes: adjective forms read as `allow-listed` (e.g. "Two families are allow-listed", "none of which are allow-listed here") and the compound before a noun as `allow-list walk`; the two `<b>whitelist</b>`/`<b>allow list</b>` doc emphasis and the `A blacklist of…` → `A deny list of…` contrast are preserved.

Pushed as a fast-forward follow-up (`ddd3a44b..bcb9e48e`) — no amend/rebase of the reviewed head. The 5 pre-existing occurrences on `main` (`ThisQualificationTests.cs` ×4, `OuterTypePatternDispatchRaisingTests.cs` ×1) remain out of scope for a separate repo-wide sweep.

Issue #3079, opened 2026-07-23
Tracking: EVIL authored-corpus benchmark discovery — methodology, bucket classification, filed defects


richlander:
## Purpose

Tracking/epic issue for the **EVIL authored-source correspondence benchmark**
discovery effort so the methodology, the product-vs-harness bucket classification,
and the filed issues are not lost between sessions. EVIL is the adversarial hard-IL
authored-source corpus run via `--benchmark-authored-corpus`; each row compares the
decompiler's output for a real third-party member against its authored source and
buckets it (`Correct` / `ValidDifferent` / `Invalid` / `Unsupported`).

This is a **discovery** effort: triage benchmark results into actionable defect
clusters and file focused issues, always deciding whether a defect is a genuine
**product decompiler** bug, an **RTS/harness** artifact, or a **regression**. It
does not itself implement fixes.

## Results history

Full-corpus benchmark results by run. **Only same-pool runs are directly
comparable** — a partial pool (fewer matched assemblies) changes the denominator.
Prefer the **full 26/26** run as the tracked baseline. Each future run **must**
record the commit SHA + date (the benchmark JSON does not embed them) and save its
`--benchmark-authored-corpus` JSON.

| Date | Commit | Pool (matched/26) | Evaluated | Valid % | Correct | ValidDiff (IL-exact / IL-diff) | Invalid | Unsupported | honest | Since (fixes landed) |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 2026-07-20 | `~main` (exact SHA not recorded; ≈`45585f9c`…`f8462478` window) | 26/26 | 12000 | 56.6% | 1501 | 5290 (3097 / 2181) | 5209 | 0 | ✅ | baseline |
| 2026-07-21 | `f8462478` (partial pool) | 18/26 | 7663 | 59.8% | 1056 | 3526 (2137 / 1379) | 2966 | 115 | ❌ partial | — (not comparable) |

Reproduce a full run (deterministic pool, all 26 corpus assemblies covered) at a
recorded commit. `eng/prepare-evil-corpus.sh` rebuilds the exact harvest pool
(100-rank NuGet sweep + the 14 pinned real-world assemblies, with repo-own
assemblies pinned via the published `dotnet-inspect.any` package), so no manual
cache-snapshot or exclude/pin step is needed and the run is honest (26/26):

```bash
bash eng/prepare-evil-corpus.sh /tmp/evil-pool
[...]

PR #3231, opened 2026-07-26
Add span attribution to RTS fault isolation (#3079)

richlander:
## Summary

**Tightens** the EVIL headline metric `invalidBreakdown.productBodyDefect` from a loose lower bound to a **tighter lower bound** — 304 → **366** on the pinned pool. It remains a lower bound, not a measured count: every uncertain case declines by design.

The substitution oracle (`ReturnToSender.TryIsolateRecompileFailure`) only credits a body defect when the checksum-verified authored body compiles in the failing row's shell. When the shell is *also* broken the experiment goes blind, so a row where both the shell is broken **and** the decompiler mangled the body is filed entirely as `ShellOrClosureDefect` harness noise. That masked population is difficulty-correlated (deep closures/generics break both the shell reconstructor and the decompiler), so the 304 bound is loosest exactly where the product is weakest.

## Design (default-deny over provably shell-independent errors)

When the shell is broken, recover signal **without a further compile**: locate the target member's body span in both the decompiled and authored artifacts (already compiled by the substitution pass) and compare where each `Error`-severity diagnostic lands.

A body defect is credited only when **both** hold:

1. the authored body is error-free within its own span (positive control — the body region *can* be clean in this exact shell), and
2. the decompiled body carries an in-body error a broken shell **provably cannot manufacture**.

Condition 2 is the entire soundness argument, enforced by default-deny. Only two classes qualify:

- any **parser diagnostic** in the body span (`SyntaxTree.GetDiagnostics()`, which no shell state can influence), and
- a **body-intrinsic semantic error** from an explicit allowlist — currently exactly `CS0128` (duplicate local).

Context-dependent errors (`CS0103`/`CS0246`/`CS0234`/`CS1061`/`CS1069`, conversions, overloads) are **never** credited: a broken shell reconstructor produces them identically to a real body defect. Everything else declines. Deterministic, exhaustive over all invalid rows, **adds zero compiles**.

Outcomes: authored body compiles → `BodyDefect` (unchanged; provenance `SubstitutionControl`). Shell broken + rule above satisfied → `BodyDefect` (provenance `SpanMeasured`). Otherwise → `ShellOrClosureDefect` (unchanged). Body unlocatable, or `IterationBudget` (unconfirmed failure) → attribute nothing.

## Ownership

Attribution is harness-owned (`SpanAttribution.cs`): it parses composed source and reads Roslyn diagnostics. No body-span API is pushed into the product `CSharpTypePrinter`. Provenance is exposed via `FaultIsolationMethod` and serialized per row (`faultIsolationMethod`), so the next audit of this classifier needs no re-run.

## methodologyVersion

Changing how `productBodyDefect` is computed breaks comparability with recorded history. The benchmark JSON stamps `methodologyVersion` (v2) top-level; the history schema carries it (absent = v1); and `--history-card` splits the product-defect movement row across a v1/v2 boundary so it never charts a step between the two bounds. Because the allowlist defines what v2 *means*, expanding it requires a new version — documented in the README.

## Evidence

Real corpus canary — EVIL, pinned pool `0a7eded85c3e1410`, 26/26 assemblies, 12,000 targets:

| Metric | v1 (`abec2dd7`, main) | v2 (`92ea6dd0`) | Δ |
| --- | --- | --- | --- |
| `productBodyDefect` | 304 | **366** | **+62** |
| `harnessShellReconstruction` | 4831 | 4769 | −62 |
| `unclassified` | 82 | 82 | 0 |

`correct` (1560), `validDifferent` (5223), `validBreakdown`, and `invalid` (5217) are **identical** across the two runs — only the invalid *split* moved. Two independent consistency checks:

- The `SubstitutionControl` subset is **exactly 304** and `v1 − v2 = 0` rows: v2 is strictly additive, losing nothing.
- The +62 rows are **exactly** the `SpanMeasured` rows, and every one is a syntax error (`CS1001` ×37, `CS1525` ×25) — zero resolution IDs, as the rule requires.

Both runs archived out-of-tree; the v2 row is appended to `history.jsonl`, so the version-boundary split is exercised by real history (rendered card confirms no `✓`/`✗` step across the boundary).

The v2 row is stamped `92ea6dd0`, the commit the canary ran at, while the merged head is two commits later: a `main` merge and the review-fix commit. Those intervening commits are documentation/labelling plus the `parseOptions` threading, which is behaviorally inert on this path (`SyntaxTree.GetDiagnostics()` applies no language-version gating, as measured in review), so the row remains the honest record of its own commit.

Tests: `SpanAttributionTests` (locator: method/overload/getter/ctor/ambiguous; classifier positives *and* close negatives — syntax→credit, CS0128→credit, CS0103 resolution→decline, CS0165 unassigned-local→decline), `AuthoredCorpusHistoryCardTests` (boundary split + methodology parse), `DynamicCompilationSiteInventoryTests`, `ReturnToSenderPrototypeTests`. Full solution build clean.

Two `ReturnToSenderPrototypeTests` failures (`CompileBackFirstPropertyGetter_EmitsClosureConstructorRequirement`, `CompileBackTargets_FullReconstructsPlainEventAccessorTarget`) reproduce identically on unmodified `origin/main` @ `a25ef6aa` — pre-existing, unrelated to this PR.

## Adversarial review

Two-model adversarial review (Gemini Pro + GPT) on fixed head `92ea6dd0`: **both clean**. Reconciliation in the thread below.


richlander:
Reviewed at `92ea6dd0` in a clean worktree, built `dotnet-inspect.slnx -c Release`. Tests green: `SpanAttributionTests` 12/12, `AuthoredCorpusHistoryCardTests` 14/14, `ReturnToSenderPrototypeTests` 160/160.

**The mechanism is sound and better than what I proposed in the #3079 thread.** Two things need fixing before merge, both about what the PR *says* it did rather than what it did.

## Credit where it's due, because it changes the review

In the #3079 discussion I suggested diffing the decompiled-body and authored-body **diagnostic-ID sets** from the same shell and crediting IDs present in one and absent in the other, with "a defined rule for cascade noise" as a caveat. That would have been unsound: a broken shell reconstructor produces CS0246/CS0117/CS1061 in the decompiled body and *not* in the authored body all the time, simply because the two bodies reference different names. My rule would have credited those as product defects.

This PR rejected that and implemented default-deny over a **provably shell-independent** allowlist instead — parser diagnostics (`SyntaxTree.GetDiagnostics()`, which no shell state can influence) plus exactly one semantic ID, `CS0128` (duplicate local). `SpanAttribution.cs:101-104` states the reason explicitly:

> Context-dependent errors (unresolved names/types/members, conversions, overloads) are never credited because a broken shell reconstructor produces them identically to a real body defect.

`DecompiledBodyIsolated_FalseWhenBothBodiesShareCascadeError` and `..._FalseWhenDecompiledBodyHasOnlyResolutionError` pin exactly that boundary, and the head commit dropping CS0165 is the same instinct applied again. That is the correct call and the negatives are the right ones.

## 1. The headline claim contradicts the code (blocking)

> Converts the EVIL headline metric `invalidBreakdown.productBodyDefect` from a **sound-but-loose lower bound** into a **measured** count.

It does not. `SpanAttribution.cs:118-122`:

```csharp
// Substitution control: the authored body must be clean within its own
// span. If a broken shell also breaks the authored body span we decline
// (a conservative false negative that keeps the count a lower bound).
```

Every decline path is a false negative by design: authored span dirty → decline; body unlocatable → decline; error not in the allowlist → decline. With the allowlist at *one* semantic ID, a genuine body defect that surfaces as CS0246 stays uncounted. v2 is a **tighter lower bound**, not a measured count.

This isn't pedantry, because the word is about to become durable. The README (`corpus/evil-runs/README.md`) contrasts `Product defects (v1 lower bound)` against `Product defects (v2 span-measured)`, which reads as "v1 was a bound, v2 is the truth." Please label both as bounds — e.g. `(v2 span-measured lower bound)` — and restate the PR summary. The honest and still-valuable claim is: *v2 tightens the bound by crediting shell-independent in-body errors that the substitution control alone had to discard.*

## 2. The README describes the rule you deliberately rejected (blocking)

> **v2** adds span attribution: when the shell is broken it additionally credits a body defect if the authored body is error-free within its own body span while the decompiled body carries an in-body error.

That is the naive `bodyA == 0 && bodyD > 0` rule — with no mention of the shell-independence restriction, which *is* the entire soundness argument. A reader reconstructing the methodology from this README (which is the methodology of record for `history.jsonl`) would build the unsound classifier and get a materially larger, wrong number, while stamping it `v2`.

Add the restriction: only a parser diagnostic in the body span, or a body-intrinsic semantic error from an explicit allowlist (currently `CS0128`); context-dependent errors are never credited. Also worth recording the allowlist's current contents in the README, since expanding it silently changes what `v2` means without changing the version stamp.

## 3. The evidence the PR promises is missing (blocking)

> Real corpus canary (EVIL, same pinned pool `0a7eded85c3e1410`): running; `productBodyDefect` expected to rise above 304 … Before/after to be posted.

No canary result has been posted. This PR redefines a headline metric and stamps a permanent `methodologyVersion`, and `AGENTS.md` requires the pinned input, command, baseline, and result for corpus claims. The number also determines how the v1/v2 boundary should be described: 304 → 320 and 304 → 900 are very different stories, and the first would be worth saying out loud rather than leaving readers to assume the reclassification was large.

Please post before/after on the pinned pool, and add the v2 row to `history.jsonl` in the same change so the boundary logic is exercised by real data rather than only by `AuthoredCorpusHistoryCardTests`.

## 4. Minor: parse options aren't threaded (not a live bug — I checked)

`SpanAttribution.cs:157` and `:206` call `CSharpSyntaxTree.ParseText(source)` with **no** `CSharpParseOptions`, while the harness compiles with `LanguageVersion.Preview` (`ReturnToSender.cs:1543`). `parseOptions` is already in scope at the `TryIsolateRecompileFailure` call site and simply isn't passed down.

I expected this to be a soundness hole — a preview-only construct failing the re-parse would manufacture a "shell-independent syntax error" and *over*-count the metric being redefined. **I could not exhibit it**, and the reason is worth recording: `SyntaxTree.GetDiagnostics()` does not apply language-version gating at all. Measured with Roslyn 5.6.0:

```
Default resolves to : CSharp14      Preview resolves to : Preview
record decl  @CSharp7_3  errors=0      raw string  @CSharp7_3  errors=0
```

Version gating is a *declaration/binding* diagnostic, so the syntax-only path is immune. `CSharpParseOptions` also carries preprocessor symbols, but `ReturnToSender.cs:1543` sets none, so there's no divergence there either.

So: latent, not live. Still worth threading `parseOptions` through — it's free, it's already in scope, and the immunity above is a Roslyn implementation detail rather than a guarantee anyone should be relying on.

## On scope

The ownership split is right — attribution stays harness-side in `SpanAttribution.cs` and no body-span API leaks into `CSharpTypePrinter`. `FaultIsolationMethod` being serialized per row is the thing that makes the next audit of this classifier possible without re-running the sweep, which is a direct improvement on the situation we hit auditing the 4,831/304 split. Good call.

richlander:
## Review reconciliation — all four findings addressed in `328062cc`

Thanks — findings 1 and 2 were the right catch, and 2 in particular would have been genuinely harmful: the README is the methodology of record for `history.jsonl`, and it described the rule this PR exists to *reject*.

### 1. "Measured count" contradicted the code — fixed

You're right, and the code comment you quoted is the proof: every decline path is a deliberate false negative. v2 is a **tighter lower bound**, not a measurement. Changed everywhere the word was about to become durable:

- Card rows → `Product defects (v1 substitution lower bound)` and `Product defects (v2 span-measured lower bound)`.
- README now leads with "**Every version is a lower bound**; a later version tightens the bound rather than measuring the true count."
- `AuthoredCorpusHistoryCard.cs` comment and the PR summary restated. The summary now says what you proposed: v2 tightens the bound by crediting shell-independent in-body errors the substitution control had to discard.

### 2. README described the rejected rule — fixed

Rewritten to document the actual rule: default-deny, the two conditions, both credited classes (parser diagnostics in the body span; the allowlist, **currently exactly `CS0128`**), and the explicitly never-credited context-dependent errors (`CS0103`/`CS0246`/`CS0234`/`CS1061`/`CS1069`, conversions, overloads) with the reason. Also took your point about silent expansion: **expanding the allowlist now requires a new `methodologyVersion`**, stated in the README, since the allowlist is what `v2` means.

### 3. Missing canary evidence — posted, and it's a clean controlled result

Pinned pool `0a7eded85c3e1410`, 26/26 assemblies, 12,000 targets. **304 → 366 (+62)**, full table in the updated PR body.

The comparison is tighter than a before/after usually is: `correct` (1560), `validDifferent` (5223), `validBreakdown`, and `invalid` (5217) are **identical** to the v1 run — only the invalid split moved, +62 product / −62 harness / unclassified unchanged. Two independent checks:

- `SubstitutionControl` subset is **exactly 304**, and `v1 − v2 = 0` rows. v2 is strictly additive; no v1 row was reclassified away.
- The +62 are **exactly** the `SpanMeasured` rows, and all are syntax errors (`CS1001` ×37, `CS1525` ×25) — **zero resolution IDs**, confirming the default-deny rule holds on real data.

To your point that "304 → 320 and 304 → 900 are very different stories": it's the small story, +20.4%, and worth saying out loud. The reclassification is modest, and 5.9% of the harness bucket remains a lower bound — the masked population is still largely masked.

I also merged `origin/main` (which carries the v1 `abec2dd7` row from #3227) and appended the v2 row, so the boundary logic is now exercised by **real history**, not only fixtures:

```
| Product defects (v1 substitution lower bound) ↓ | - | 306 | 304 ✓ | - |
| Product defects (v2 span-measured lower bound) ↓ | - | - | - | 366 |
```

No `✓`/`✗` glyph beside 366 — no step charted across the boundary.

### 4. `parseOptions` not threaded — fixed, and thank you for the measurement

Threaded through `IsolatingBodyError` → `TryLocateBodySpan` / `FirstSyntaxErrorInSpan`; it was already in scope at the call site. Your finding that `SyntaxTree.GetDiagnostics()` applies no language-version gating is recorded in the XML doc as the reason this was latent rather than live, precisely so nobody later reads the fix as unnecessary — it's a Roslyn implementation detail, not a guarantee.

### Prior adversarial review, for the record

Two-model review on `92ea6dd0` (Gemini Pro + GPT), both **clean**, after two rounds of findings:

- **GPT r2 #1 — `CS0165` is shell-dependent.** Correct: definite assignment can hinge on a compile-time `const` the shell reconstructor may drop. Gemini argued it was unreachable (the compiler const-folds pre-decompilation). Rather than adjudicate reachability I removed it — it cost exactly **1 corpus row**, and a soundness rule shouldn't rest on a contested reachability argument. Allowlist is `{CS0128}`.
- **GPT r2 #2 — the `IterationBudget` gate only covered the span branch.** Correct: the substitution branch still ran on a status where `RoundTripCompilationEngine` returns a composed-but-never-emitted artifact with stale diagnostics — an unconfirmed failure. Now **all** fault isolation is skipped on `IterationBudget`. 0 rows in this corpus, so no metric impact, but the hole was real.
- **Gemini r2** independently verified `CS0128` shell-independence with compiled probes (a shell member collision yields `CS0169`, a parameter collision `CS0136` — neither is `CS0128`).

### Non-blocking follow-up

The `ReturnToSenderSourceProbe.cs:68` string-prefix fallback (`HasClosureStopDetail` → `HarnessShellReconstruction`) is still an unvalidated path, distinct from the oracle. It's distinguishable in the artifact (`faultIsolation == null`), so it can be audited without a re-run — filing separately rather than expanding this PR.

Two `ReturnToSenderPrototypeTests` failures reproduce identically on unmodified `origin/main` @ `a25ef6aa`; pre-existing and unrelated.

richlander:
Re-reviewed at fixed head `328062cc` in a clean worktree; built `dotnet-inspect.slnx -c Release` (0 errors). **All four of my findings are closed.** Verified each rather than taking the reconciliation on trust:

| # | finding | verified at `328062cc` |
| --- | --- | --- |
| 1 | "measured count" overclaim | `AuthoredCorpusHistoryCard.cs:171-172` now reads `v1 substitution lower bound` / `v2 span-measured lower bound`; `:160` states both are lower bounds |
| 2 | README described the rejected rule | `README.md:55-74` documents default-deny, both credited classes, the `CS0128` allowlist, and the never-credited set with the reason |
| 3 | canary never posted | `history.jsonl` row appended and internally consistent |
| 4 | `parseOptions` not threaded | threaded end-to-end (`:83, 114, 116, 118, 128, 161, 166, 210, 215`) and passed for real at `ReturnToSender.cs:2203` |

On **#3**, the two rows check out against the claims arithmetically — this is the part I most expected to wobble and it doesn't:

```
v1  abec2dd7  mv=—  product=304  harness=4831  uncl=82  correct=1560  invalid=5217
v2  92ea6dd0  mv=2  product=366  harness=4769  uncl=82  correct=1560  invalid=5217
```

+62 product / −62 harness, with `unclassified`, `correct`, and `invalid` byte-identical. That is exactly the signature of a strictly-additive reclassification, and it's a much better piece of evidence than a bare before/after because the invariant columns are what rule out a confound. Agreed on the framing too: +20.4% is the small story, and saying so is the right call.

---

### One new non-blocking finding: the allowlist is the definition of `v2`, and nothing pins it

The README now says (`:71-74`):

> Because the allowlist defines what `v2` means, expanding it requires a new `methodologyVersion`; do not add IDs under the existing stamp.

That is the correct rule and it is **prose only**. `BodyIntrinsicSemanticErrorIds` appears in exactly one file repo-wide:

```
$ grep -rn "BodyIntrinsicSemanticErrorIds" --include=*.cs . | grep -v SpanAttribution.cs
(no output)
```

`MethodologyVersion = 2` lives in a different file (`AuthoredCorpusBenchmark.cs:32`) with no code path connecting the two.

I measured what that permits. Three tampering runs against the full `--gate fast`, `methodologyVersion` left at 2 in every one:

| allowlist addition | category | fast gate |
| --- | --- | --- |
| `CS0165` | unassigned local | **1 failed** ✓ caught |
| `CS0246, CS0103, CS1061, CS0234` | resolution | **2 failed** ✓ caught |
| `CS0029, CS1503, CS0136` | conversion / overload / scope collision | **3828 passed, 0 failed** ✗ |

So the *enumerated* never-credited IDs are genuinely pinned — `DecompiledBodyIsolated_FalseWhenDecompiledBodyHasOnlyResolutionError` and `...HasUnassignedLocalError` both do real work, and I'll note that's better than I expected going in. But the README also forbids whole *categories* ("conversions, and overload resolution"), and those have no pin. A future contributor can widen the soundness rule and ship it under an unchanged `v2` stamp with a green gate.

`CS0136` is the sharp end of this. Your own reconciliation records Gemini's r2 probe finding that **a parameter collision yields `CS0136`** — so `CS0136` was identified *during this review* as shell-dependent, i.e. precisely the class condition 2 exists to exclude. Nothing stops it being added tomorrow. And the damage isn't a wrong number in isolation: it's that rows sharing a stamp are supposed to be comparable, so a silent allowlist change under a fixed `v2` defeats the boundary logic that `Render_MovementSplitsProductDefectAcrossMethodologyBoundaryWithoutCharting` exists to enforce. The card would chart a v2→v2 step across two different methodologies.

The fix is cheap and there's now a house pattern for it — #3249 closed #3247 by making the declaration *drive* the enforcement set (`ByteNeutralKnobs = StyleOptionCatalog.Options.Where(o => !o.ByteDivergent)`, then `Assert.Equal(required, covered)` for set equality). The analogue here is a gate asserting the allowlist equals a pinned set for the current `MethodologyVersion`, so adding any ID fails until the version is bumped — which is exactly what the README already promises. That converts the rule from a comment into a tripwire and makes the stamp mean something mechanically.

Non-blocking because the allowlist is correct **today** and the shipped numbers are sound.

### Minor

The v2 row is stamped `92ea6dd0`, but head is `328062cc` — two commits later (a `main` merge and this review-fix commit, which includes the `parseOptions` threading). Per my earlier measurement the threading is behaviorally inert on the `GetDiagnostics()` path, and the row honestly records its own commit, so I don't think a re-run is required. Worth one line in the PR body saying the intervening commits are docs plus an inert threading fix, so a later reader doesn't have to re-derive that.

Also confirmed the two `ReturnToSenderPrototypeTests` failures reproduce on unmodified `origin/main` @ `a25ef6aa` — pre-existing, agreed, not yours.

**Approving.** Findings 1 and 2 were worth the round trip, and the canary is a genuinely well-controlled result.

richlander:
## Post-merge verification: the recorded v2 row needed correcting — filed as #3305

Following up on the minor point about the row being stamped `92ea6dd0` while head moved on. I ran the canary at the merged commit rather than relying on the inertness argument, and the row does **not** reproduce: `14781e8d` measures `productBodyDefect` **326**, not 366.

**Your inertness call was right.** Controlled runs on the pinned pool:

| run | product | harness | uncl. | SpanMeasured |
| `14781e8d`, `parseOptions` threaded | 326 | 4808 | 64 | 23 (CS1525) |
| `14781e8d`, `parseOptions` reverted | 326 | 4808 | 64 | 23 (CS1525) |
| `92ea6dd0` (older decompiler) | 366 | 4769 | 82 | 62 (CS1001 ×37, CS1525 ×25) |

The first two are **byte-identical** — the threading is inert on the real corpus, not just in principle. I also probed it directly: no preview construct produces a parser diagnostic under either parse mode, since version gating is a binding diagnostic. So your measurement held up end-to-end.

**The gap is that `92ea6dd0` was branched before this PR merged**, so it carried an older decompiler. The whole delta is main's work landing in between: `correct` +16 and `invalid` −19 are genuine fixes, and product −40 / harness +39 is *correct reattribution* — 37 previously-unparseable bodies (CS1001) now parse, so span attribution rightly stops crediting a body defect and those rows fall back to their still-broken shell. The metric did exactly what it was designed to do.

Both numbers are honest for their own code; the issue is only that a trend series needs rows reproducible from their recorded commit. #3305 replaces the row with the `14781e8d` measurement and makes the append procedure require a main commit so it can't recur. The 304-vs-366 methodology A/B stays here, which is where an experiment against its own base belongs.

Worth noting the two findings compound usefully: had the allowlist not been pinned (#3270), this reattribution would have been indistinguishable from someone quietly widening the rule.

Issue #3197, opened 2026-07-25
Decompiler: extend compile-back fidelity to platform / cross-assembly-internal methods (DeepEquals scenario)


richlander:
## Summary

Extend the compile-back fidelity oracle to reach **platform / cross-assembly-internal
methods** — the class of method that motivated #3161
(`System.Text.Json.JsonElement.DeepEquals`). Today such methods can be graded for
*decompiler fidelity* (`Full`/`Partial`/…) and checked for *render validity*
(stages 2/5), but they cannot be run through the **contract-V1 compile-back
oracle** that independently recompiles the decompiled C# and diffs IL. That leaves
a real coverage gap: for platform methods we cannot prove the raised output
recompiles to the original opcodes.

Split out of #3159 / #3161; surfaced while finishing #3174.

## The gap

The compile-back oracle (`ILInspector.DecompilerHarness.FidelityCheck`, driven by
`FidelityGateTests` over `CfgSampleClass`) recompiles a decompiled method inside a
reconstructed type skeleton with `csc` and diffs the resulting IL
(`Exact`/`OpcodeDiff`/`OperandDiff`/…). This only works when the decompiled body
references a **recompilable surface**.

`DeepEquals` does not. Its decompiled body references internal/private
`System.Text.Json` surface — `CheckValidInstance`, `ValueIsEscaped`, `NameEquals`,
`UnorderedObjectDeepEquals`, `ArrayEnumerator`/`ObjectEnumerator`, `ThrowHelper.*`,
`JsonProperty` — that a shipped DLL does not expose as compilable source. The
harness therefore returns `FidelityUnavailable` / `ContextFail` (the honest-exit
buckets), and the shape's IL fidelity goes unproven on the real method.

Note #3174 mitigates this for the *shape* by adding a dependency-free,
high-likeness fixture (`CfgSampleClass.SwitchWithTwoLoopingCaseSections` — a
`switch` with two loop-bearing sections mirroring DeepEquals's Array and Object
arms) that is pinned `Exact`. That proves the transformation's IL fidelity on a
shape we own; it does **not** prove it on DeepEquals itself. This issue tracks
closing that remaining gap.

## Feasible approach

Cross-assembly, metadata-driven **stub synthesis**:

- From the referenced (but non-recompilable) surface, synthesize
  **declaration-only** stubs (`throw null;` bodies) — including internals — so the
  decompiled body binds and compiles without the real implementations.
[...]

Issue #3166, opened 2026-07-25
Decompiler: value-swap idiom in JsonElement.DeepEquals renders with synthetic temp name


richlander:
## Defect

A value-swap idiom renders with a synthetic temp name:

```csharp
S_257 = element2;
element2 = element1;
element1 = S_257;
```

(The `NameEquals` local function has the equivalent with `S_256_1`.)

## Expected

Render the swap with a readable temp name (or recognized swap form). Lower
priority than the control-flow items in #3159, but tracked for completeness.

## Repro

```bash
dotnet-inspect System.Text.Json.JsonElement.DeepEquals -S "Decom*"
```

`Library: System.Text.Json | Source: Platform | TFM: net11.0`

Part of #3159.

richlander:
Fixed by #3234 (merged to `main` at e849fa4e). The value-swap idiom now raises to the recognized tuple deconstruction swap `(q, p) = (p, q)` instead of a synthetic temp (`S_257 = ...`), guarded to distinct same-type by-value places and witnessed by the compile-back-Exact `CfgSampleClass.SwapStructPair` fixture. (The PR body's `Fixes/advances` phrasing didn't trigger auto-close.)

Issue #3165, opened 2026-07-25
Decompiler: recover synthetic local names / sink hoisted decls in JsonElement.DeepEquals


richlander:
## Defect

Synthetic locals leak into the body and are hoisted into a top-of-method
declaration block:

```csharp
int __switchValue0 = default;
JsonValueKind kind;
ArrayEnumerator arrayEnumerator2;   // no arrayEnumerator1
...
ReadOnlyMemory<byte> V_5;
ArrayEnumerator V_6;
ReadOnlySpan<byte> S_256;
JsonElement S_257;
```

`__switchValue0`, `V_5`, `V_6`, `S_256`, `S_257`, `S_256_1` (in the `NameEquals`
local function), and the mismatched `arrayEnumerator2` naming are all synthetic.

## Expected

Recover/clean local names and sink declarations to point-of-use rather than a
hoisted block at method entry.

## Repro

```bash
dotnet-inspect System.Text.Json.JsonElement.DeepEquals -S "Decom*"
```

`Library: System.Text.Json | Source: Platform | TFM: net11.0`

Part of #3159.

PR #3234, opened 2026-07-26
Raise the value-swap idiom to the recognized tuple swap form (#3166)

richlander:
- Fixes/advances #3166
- Changes the value-swap idiom `tmp = q; q = p; p = tmp;` to render as the recognized tuple swap `(p, q) = (q, p);` when it is opcode-exact.
- Card revision: `ab787539aec41255e2fc8ce12bf107a59bc92802`

## Change

**Conclusion:** **PASS** — the raise reproduces the original source's own tuple-swap spelling and is opcode-exact (Roslyn lowers `(a,b)=(b,a)` to the identical one-temp dup-slot IL as the manual swap), so it is a byte-faithful default with a compile-back-checkable witness fixture pinned Exact.

### Benchmark target

Benchmark target: `System.Text.Json` (platform, net11.0)

dotnet-inspect command:

```bash
dotnet-inspect member System.Text.Json.JsonElement.DeepEquals -S "Decompiled Source"
```

### Original source

```csharp
public static bool DeepEquals(JsonElement element1, JsonElement element2)
{
    // ...
    if (element2.ValueIsEscaped)
    {
        if (element1.ValueIsEscaped)
        {
            return JsonReaderHelper.UnescapeAndCompareBothInputs(element1.ValueSpan, element2.ValueSpan);
        }

        // Swap values so that unescaping is handled by the LHS.
        (element1, element2) = (element2, element1);
    }
    // ...
    static bool NameEquals(JsonProperty left, JsonProperty right)
    {
        if (right.NameIsEscaped)
        {
            if (left.NameIsEscaped)
            {
                return JsonReaderHelper.UnescapeAndCompareBothInputs(left.NameSpan, right.NameSpan);
            }

            // Swap values so that unescaping is handled by the LHS
            (left, right) = (right, left);
        }
        return left.NameEquals(right.NameSpan);
    }
}
```

The authored source uses the C# tuple-swap in both places.

### Before

```csharp
case JsonValueKind.Object:
    if (element2.ValueIsEscaped)
    {
        if (element1.ValueIsEscaped)
        {
            return JsonReaderHelper.UnescapeAndCompareBothInputs(element1.ValueSpan, element2.ValueSpan);
        }
        S_257 = element2;
        element2 = element1;
        element1 = S_257;
    }
    return element1.ValueEquals(element2.ValueSpan);
// ... and in NameEquals:
    right = left;
    left = S_256_1;
```

- Valid: True
- Correct: True
- IL fidelity: True (opcode-exact; the flat 3-statement form and the tuple form lower to identical IL)
- Taste applied: None
- Commit: 08ab86b1

The pre-change render exposes the compiler's one-temp swap carrier (`S_257`, a synthetic dup-slot name) instead of the source's tuple swap.

### After

```csharp
case JsonValueKind.Object:
    if (element2.ValueIsEscaped)
    {
        if (element1.ValueIsEscaped)
        {
            return JsonReaderHelper.UnescapeAndCompareBothInputs(element1.ValueSpan, element2.ValueSpan);
        }
        (element1, element2) = (element2, element1);
    }
    return element1.ValueEquals(element2.ValueSpan);
// ... and in NameEquals:
    (left, right) = (right, left);
```

- Valid: True
- Correct: True
- IL fidelity: True — witnessed by the pinned `CfgSampleClass.SwapStructPair` fixture (compile-back **Exact**). DeepEquals itself remains uncheckable by the compile-back harness (internal STJ surface, per #3197), but the shape is proven opcode-exact by the fixture.
- Taste applied: None
- Commit: ab787539aec41255e2fc8ce12bf107a59bc92802

The After render reproduces the original source's `(element1, element2) = (element2, element1);` and `(left, right) = (right, left);` verbatim. The remaining `S_256` in the `Array` arm is a genuine non-swap value spill (`element1.GetRawValue().Span`), correctly left untouched (tracked separately under #3165).

The raise is scoped to distinct, same-type, by-value parameter/local places that are legal `ValueTuple` elements. Byref (a `ref` reseat, not a value swap), pointer, function-pointer, generic-parameter (may `allow ref struct`), and ref-struct / stack-only places are declined and keep their flat three-statement form — otherwise the tuple would either change meaning or be uncompilable (CS9244 / CS0306). See the adversarial-review reconciliation comment for the two guard fixes and their reproductions.

### Fully raised

The After decompilation is in the fully raised state for the swap idiom.

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | n/a (new file) | `SwapIdiomPassTests`: 12 passed; `DeconstructionAssignmentPassTests`: 32 passed |
| Decompiler fast suite | 288 RoundTrip passed | 288 RoundTrip passed |
| Reduced fixture fidelity | not applicable (fixture new) | `SwapStructPair`: compile-back **Exact** |
| Real witness | `JsonElement::DeepEquals` swap renders `S_257 = element2; ...` | `JsonElement::DeepEquals` swap renders `(element1, element2) = (element2, element1);` |

Pre-existing baseline fidelity failures unchanged by this PR (same tests, same reasons): `ManualConstantOnlyComparisonFactory` (OpcodeDiff, expr-tree factory temp spill), `StatementBodyLambdaInsideIf` (OperandDiff, csc lambda-ordinal drift), `TupleRest` (RecompileFail CS0246, environmental). `SwapStructPair` is absent from the entire diff set on both base and head.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -filter "/*/*/SwapIdiomPassTests/*"
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -trait "Area=RoundTrip" -trait- "Speed=Slow"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -2870,6 +2870,24 @@ public class CfgSampleClass
         return a + b;
     }
 
+    // #3166 compile-back witness: the value-swap idiom. csc lowers the tuple swap
+    // `(a, b) = (b, a)` — and the equivalent manual `temp = b; b = a; a = temp;` —
+    // to a single dup-slot save followed by the two cross-stores (verified
+    // byte-identical for a struct), so SwapIdiomPass raising the surviving carrier
+    // back to `(a, b) = (b, a)` is opcode-exact, not a byte-divergent rewrite. This
+    // mirrors System.Text.Json.JsonElement.DeepEquals, which swaps two JsonElement
+    // structs through one such temp (issue #3166); carrying it into the fidelity
+    // gate proves the raised swap recompiles to the original IL — the check the
+    // internal-surface DeepEquals cannot itself feed to compile-back.
+    public static int SwapStructPair(Pairing a, Pairing b)
+    {
+        if (a.First < b.First)
+        {
+            (a, b) = (b, a);
+        }
+        return a.First - b.First;
+    }
+
     public static object AnonShorthand(int a, string b) => new { a, b };
 
     public static object AnonNamed(int x, string y) => new { Id = x, Name = y };
--- a/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
@@ -326,6 +326,15 @@ public class FidelityGateTests
         // internal System.Text.Json surface is not recompilable by the compile-back
         // oracle (tracked for a future cross-assembly compile-back capability).
         "SwitchWithTwoLoopingCaseSections",
+        // #3166: the value-swap idiom. csc lowers the tuple swap `(a, b) = (b, a)`
+        // (and the equivalent manual `temp = b; b = a; a = temp;`) to a single
+        // dup-slot save plus the two cross-stores; SwapIdiomPass raises that
+        // surviving carrier back to `(a, b) = (b, a)`. Because the tuple swap and
+        // the one-temp sequence share IL, the raise is opcode-exact — this pins it.
+        // Mirrors System.Text.Json.JsonElement.DeepEquals, which swaps two
+        // JsonElement structs through one such temp but cannot itself be fed to the
+        // compile-back oracle (internal System.Text.Json surface).
+        "SwapStructPair",
         // The compile-back oracle replays the fixture's runtime-async feature,
         // so these methods must retain the same lowering rather than merely
         // remaining recompilable.
--- a/src/ILInspector.Decompiler.Tests/SwapIdiomPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/SwapIdiomPassTests.cs
@@ -43,6 +43,49 @@ public class SwapIdiomPassTests
         Assert.DoesNotContain("S_", output);
     }
 
+    // --- Negative: place types that forbid a tuple swap (byref / pointer / ref struct) ---
+
+    public static TheoryData<string, TypeRef> UnspellableSwapPlaceTypes() => new()
+    {
+        { "byref (a `ref` reseat, not a value swap)", TypeRef.ByRef(Int) },
+        { "pointer (illegal ValueTuple element, CS0306)", TypeRef.Pointer(Int) },
+        { "ref struct (illegal ValueTuple element, CS9244)",
+            TypeRef.GenericInstance(TypeRef.CoreLib("System", "Span"), [Int]) },
+    };
+
+    [Theory]
+    [MemberData(nameof(UnspellableSwapPlaceTypes))]
+    public void SwapOfUnspellablePlaceType_NotRaised(string _, TypeRef placeType)
+    {
+        // S = a; a = b; b = S;  over two places of a type that cannot appear as
+        // a ValueTuple element (or would reseat rather than assign): decline.
+        var function = BuildBlock(
+            new StoreStackSlot(0, new LoadArgument(0, "p0", placeType)),
+            new StoreArgument(0, "p0", placeType, new LoadArgument(1, "p1", placeType)),
+            new StoreArgument(1, "p1", placeType, new LoadStackSlot(0, placeType)));
+
+        new SwapIdiomPass().Run(function, PassContext.None);
+
+        Assert.Empty(function.Descendants.OfType<DeconstructionAssignment>());
+        Assert.Empty(function.Descendants.OfType<TupleExpression>());
+    }
+
+    [Fact]
+    public void SwapOfReferenceTypePlaces_Raises()
+    {
+        // Reference-type swaps are byte-exact and legal ValueTuple elements, so
+        // the guard must not over-reject them: `S = a; a = b; b = S;` still raises.
+        var stringType = TypeRef.CoreLib("System", "String");
+        var function = BuildBlock(
+            new StoreStackSlot(0, new LoadArgument(0, "p0", stringType)),
+            new StoreArgument(0, "p0", stringType, new LoadArgument(1, "p1", stringType)),
+            new StoreArgument(1, "p1", stringType, new LoadStackSlot(0, stringType)));
+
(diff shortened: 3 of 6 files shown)


richlander:

Two-model adversarial review per `AGENTS.md` (roster minus the authoring Claude family). Each reviewer worked in an isolated, read-only `/tmp` worktree pinned to an exact head, with scratch confined to `/tmp` and no git mutations in the tree.

**Reviewers:** Gemini 3.1 Pro (`gemini-3.1-pro-preview`) · GPT-5.6 Sol (`gpt-5.6-sol`)

### Round 1 — head `7942558b` — both BLOCK (independent, converging)

Both models independently found the same two defects: the raise emitted `(q, p) = (p, q);` for *any* two distinct same-type places, without checking the place type.

- **F1 — byref places change meaning (High).** A `ref` local/parameter reseat (`ref int tmp = ref p; p = ref q; q = ref tmp;`) surfaces as `Store/LoadArgument` of `ByReferenceType`. Raised to `(q, p) = (p, q);` it becomes a *value* swap of the referents — mutating caller memory instead of reseating the aliases. Reproduced on the head build.
- **F2 — pointer / ref-struct places are uncompilable (Med/High).** `Span<int>` (ref struct) → CS9244 and `int*` (pointer) → CS0306 as `ValueTuple` elements. Both raised to invalid `Full` C#. Reproduced on the head build.

**Resolution — commit `a0b9de9a`:** added `IsSwappablePlaceType`, declining `ByRef`, `Pointer`, `FunctionPointer`, `Pinned`, `Unsupported`, and ref-struct / stack-only value types (corelib names via `IsStackOnlyValueType`; same-assembly ref structs via `IsByRefLike`/`function.ByRefLikeTypes`). Added synthetic byref/pointer/ref-struct decline tests and a reference-type positive control (so the guard neither under- nor over-rejects). Verified: all three probes revert to the flat three-statement form; `DeepEquals` still raises both swaps.

### Round 2 — head `a0b9de9a`

- **Gemini — PASS.** Verified the probes revert to flat, `DeepEquals` still raises, function pointers and `TypedReference` are guarded, generic parameters and `Nullable<T>` are safe, 11 tests pass.
- **GPT — BLOCK**, two new findings:
  - **F3 — `where T : allows ref struct` generic swaps still raise (Med).** The anti-constraint is untracked in the IR, so `First<T>(T p, T q)` raised to a CS9244-invalid tuple. **Confirmed through the product path.**
  - **F4 — unlisted framework ref structs (Med).** A hand-built importer path with an *empty* `ByRefLikeTypes` let `DefaultInterpolatedStringHandler` raise to an invalid tuple.

**Resolution:**

- **F3 — fixed in commit `ab787539`:** `IsSwappablePlaceType` also declines `GenericParameter` / `MethodGenericParameter`. Conservative by design — an unconstrained `Swap<T>` no longer raises, which both reviewers agree is the correct compilation-safety trade-off given the untracked anti-constraint. Added a synthetic generic-parameter decline case. Verified: `First<T>` now renders flat.
- **F4 — non-action, verified as an environment artifact (not a product bug).** Through the real product command `dotnet-inspect member --library <dll> -m Probe.SwapHandlers`, `DefaultInterpolatedStringHandler` **correctly stays flat**: the authoritative cross-assembly `[IsByRefLike]` fact populates `function.ByRefLikeTypes` at import (`IrImporter.cs:666` → `MetadataSource.IsByRefLikeType` → `CrossAssemblyTypeResolver`), and the guard declines it. GPT's repro observed an empty `ByRefLikeTypes` only because its harness drove the importer through a `MetadataSource` that could not resolve the corelib attribute — not the product path. GPT re-confirmed on Round 3: *"Product-path resolution correctly kept `DefaultInterpolatedStringHandler` flat."* The corelib name-list in `IsStackOnlyValueType` remains as a fast-path / degraded-resolution fallback, mirroring `PatternSwitchExpressionPass`.

### Round 3 — head `ab787539` — both PASS

- **Gemini — PASS (`ab787539`).** Both `allows ref struct` and ordinary generic swaps decline; `DeepEquals` still raises; reference/value-type swaps still raise; 12 tests pass. Endorsed the conservative generic decline.
- **GPT — PASS (`ab787539`).** *"No significant issues found. Product-path resolution correctly kept `DefaultInterpolatedStringHandler` flat."*

Both fixed-head reviews are clean at `ab787539`.

richlander:
### Follow-up commit `e849fa4e` — pass-classification ledger (test-only, no re-review)

CI's `test` job caught that `SwapIdiomPass` was unclassified by
`LoweringCoverageTests.EveryPipelinePassType_IsClassified_ByOneRegister`,
which requires every `IrPasses.Default` pass type to appear in exactly one
register.

**Resolution:** registered `SwapIdiomPass` in `NativePasses` as an
`EmitArtifact`. Rationale: the one-temp value-swap IL shape (`S=p; p=q; q=S`)
is **byte-identical** whether the source wrote a manual temp swap or the tuple
deconstruction swap `(q,p)=(p,q)` — so it has no distinguishing named-construct
anchor tying it to the `DeconstructionAssignmentOperator` LocalRewriter lowering
(already owned by `DeconstructionAssignmentPass`). The pass inverts a codegen
shape *below* the LocalRewriter, which is precisely the `NativePasses`
definition. This mirrors `PatternSwitchExpressionPass`, which sits in
`NativePasses` despite emitting a switch expression.

This is a **test-only classification/ledger change with zero product behavior
change** (no raise logic, guards, or output touched). It does not alter the
adversarially-reviewed behavior at `ab787539`, so it needs no additional
adversarial round.

Local verification at `e849fa4e`:
- `LoweringCoverageTests`: 6/6 pass.
- `SwapIdiomPassTests`: 12/12 pass.

richlander:
Ready to merge

CI is green at `e849fa4e` (test, build-net10, csharp-diff-smoke, changes all pass; mergeStateStatus CLEAN, MERGEABLE). Both adversarial reviews (Gemini 3.1 Pro + GPT-5.6 Sol) PASS on the reviewed behavior head `ab787539`; the only change since is the test-only pass-classification commit `e849fa4e` (no product behavior change, no re-review required).

PR #3235, opened 2026-07-26
Apply taste to Annotated Source, and drop interleaved IL for byte-divergent lenses (#3191)

### LINKED ISSUE #3191: Annotated source: apply decompiler taste to the annotated projection and explain which knobs shaped each line
## Summary

Make **annotated source** the place that both *applies* and *explains* decompiler taste. Annotated source (raised C# with the raw IL interleaved beneath each statement) is uniquely suited to this: taste knobs change **C# spelling without changing IL**, so it is the one view where a user can see "the IL is identical; only the presentation changed because knob *X* is on." Today, taste does not reach this view at all.

## Current behavior (the gap)

- The **Source** member view applies taste: `MemberBodyProducer.ProduceMember(..., printerOptions: BuildPrinterOptions(styleOptionsJson))` (`prototypes/inspect-web/engine/Program.cs`), driven by `Pipeline.Ir.StyleOptionCatalog`.
- The **Annotated source** member view (added in `prototype/wasm-command-ui`, `QueryMemberAnnotatedSource`) runs `ILInspector.Research.ResearchViews.ProjectMember(AnnotatedSource: true)` → `RenderMixedCore` → `CSharpPrinter.PrintRaised(...)`. That IR raised-printer path **does not accept `PrinterOptions`**, and `QueryMemberAnnotatedSource` deliberately ignores `styleOptionsJson`.
- Result: flipping a taste knob changes the Source tab but silently **does not** change Annotated source, so the two views can diverge, and annotated source cannot report which knobs are active.

### Phase 1 — apply taste to the annotated projection (product)
Thread `PrinterOptions` through the IR raised path so annotated source honors the same taste knobs as Source and the two stay consistent:
- `CSharpPrinter.PrintRaised(...)` accepts `PrinterOptions` (or an equivalent style input) and honors the applicable `StyleOptionCatalog` knobs.
- `ResearchViews.RenderMixedCore` / `ProjectMember` (`MemberProjectionRequest`) carry the options through to `PrintRaised`.
- `QueryMemberAnnotatedSource` stops ignoring `styleOptionsJson` and passes the built options through (mirroring the Source path).
- Not every knob necessarily applies to IR-raised output; the building agent should determine and document which knobs are honored vs. no-ops in the mixed view.

### Phase 2 — surface which taste is applied (presentation)
In the annotated view, render a compact **"Taste applied"** header listing the active knob ids (and a "default · opcode-faithful" note when none are set), consistent with the existing Taste popover in the prototype (`state.taste`, `styleOptions`).

### Phase 3 — per-statement taste attribution (stretch, the compelling bit)
Because the mixed view already correlates each C# statement to its IL span, annotate individual statements with the knob that shaped their spelling (e.g. "shaped by `readable-local-names`"). This turns annotated source into a live, per-line explainer of what each taste knob does — its IL is invariant, only the C# above it changed.

## Layering / constraints

- Style knobs and `PrinterOptions` are a **Decompiler** concern; **Research** composes; the **CLI/prototype** presents. Keep the capability in the product layers so any consumer (CLI annotated views, not just the web prototype) benefits — do not special-case it in the prototype engine.
- Keep product paths SRM-only, NativeAOT-friendly, Roslyn-free.
- Preserve default output: with no taste selected, annotated source must be unchanged from today.

## Evidence expected

- Focused decompiler/Research tests showing a taste knob changes annotated C# while the interleaved IL lines are byte-identical.
- A negative case: a knob that is a no-op in the mixed view is documented as such (no false "applied" claim).
- Prototype: switching a knob updates both Source and Annotated source consistently; the "Taste applied" header reflects `state.taste`.

## Context

Annotated source was added in `prototype/wasm-command-ui` as the one IL-inclusive member view. This issue is filed so the work can be built independently and merged under that branch without blocking it. Related: #3180 (closed-set corpus).

richlander:
Closes #3191.

## What this does

`-S "Annotated Source"` renders decompiled C# with raw IL interleaved beneath each statement — a claim of statement-to-opcode correspondence. Taste knobs broke that view in two ways, and this PR fixes both.

**1. Applied taste was invisible.** The view rendered `this._count` with no indication the `this.` was a configured spelling rather than the compiler's. The only account lived in a separate `Applied Taste` table. Taste now rides the member signature as a single trailing side comment, in the same shape the fact overlay already uses for analysis, so a reader scans style and analysis the same way:

```csharp
public object? Pick(bool useName, string suffix)  // taste.qualify-property-access(Name); taste.qualify-field-access(_count)
{
    if (useName)
        // IL_0000: ldarg.1                        // arg: useName; stack: [bool]
        // IL_0001: brfalse.s    IL_0009           // stack: []
    {
        return new object();  // alloc.new(object; alloc=System.Object; path=branch; escape=escapes-return; ...)
    }
    return string.Concat(this.Name, suffix, this._count.ToString());
        // IL_0009: ldarg.0                        // arg: this; stack: [Cache]
        // ...
}
```

**2. A byte-divergent lens contradicted the view's premise.** `StyleOptionTier.Lens` rewrites shapes — a guarded boolean return becomes a ternary — so the printed source no longer reproduces the member's opcodes. Interleaving IL beneath it would assert a correspondence that does not hold. The Annotated view now drops the IL for any member a lens actually rewrote, and the annotation reports why:

```csharp
public bool Allow(bool trusted, bool cached)  // taste.qualify-property-access(Name); taste.qualify-field-access(_count); taste.prefer-conditional-return(fidelity=byte-divergent)
{
    return trusted ? cached : (this.Name.Length > this._count);
}
```

Suppression keys on what the render *did*, not what the host *asked for*: a lens that is enabled but finds no shape to rewrite is byte-faithful for that member, so its IL stays.

**3. `--taste`** on `member` and `type` is a one-invocation form of the pre-existing `dotnet_inspect_style_full_taste` config aggregate, so trying this out no longer requires leaving a config file behind.

## Design notes

- **The comment rides the signature** because `DecompilerDecision` carries a subject but **no IL offset**. There is no honest statement-level anchor available today. A lens's subject is just the enclosing method — already spelled on that line — so it reports `fidelity=byte-divergent` instead, which doubles as the signal explaining the absent IL.
- **The annotation is built in the CLI, not Research.** Research `Output` is body-only; the CLI's `FormatSourceWithDeclaration` owns the signature line. This also removed a prose block that had been baking presentation text into the layer that owns source.
- **`Decompiled Source` is deliberately left unannotated**, preserving its clean-source contract.
- **Nothing is annotated by default.** No style decision is recorded unless a knob was requested, so the comment appears exactly when the reader asked for taste.

## Compatibility

Default output is unchanged: no config and no `--taste` means no decisions, no annotation, and no suppression. The escaping pass added below is inert for every well-formed metadata name.

## Docs

`docs/decompiler-taste.md` had an annotated example showing a **body only, with no member signature** — contradicting the view's own contract. Replaced with real captured output, plus a worked example mixing allocation facts and taste on one member. `README.md` notes `--taste`.

## Validation

| Suite | Result |
| --- | --- |
| `src/dotnet-inspect.Tests` | 2160 passed, 12 skipped |
| `src/ILInspector.Research.Tests` | 132 passed |
| `src/ILInspector.Decompiler.Tests -- -trait- "Speed=Slow"` | 3781 passed |
| `npx markdownlint-cli` | clean |

The full decompiler suite was run on the first head: 6 failures, all verified pre-existing on a pristine `origin/main` worktree (the compile-back harness cannot resolve corelib in this environment).

## Adversarial review

Four rounds with Gemini 3.1 Pro and GPT-5.6 Sol in isolated worktrees. Both signed off on `55380643`. Findings and resolutions are reconciled in a comment below.

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,7 +304,11 @@ DecompilerHarness.
 The `Decompiled Source` view renders the shipped canonical C# by default. A
 tool-owned `.dotnet-inspectconfig` file (discovered by walking up from the
 working directory) selects opt-in class-3 spellings — today `this`-qualification
-of field and property access — using `.editorconfig` key names. See
+of field and property access — using `.editorconfig` key names. `--taste`
+requests the whole oracle-endorsed set for one invocation without a config file.
+`Annotated Source` names the applied spellings in a trailing comment on the
+member signature, and drops its interleaved IL for any member a byte-divergent
+lens actually rewrote. See
 [docs/decompiler-taste.md](docs/decompiler-taste.md#style-configuration).
 
 ```bash
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -590,8 +590,8 @@ the full account, with each rule's detail and fidelity as table rows.
 Nothing is annotated by default: no style decision is recorded unless a knob was
 requested, so the comment appears exactly when the reader asked for taste.
 `--taste` is the one-invocation gesture for requesting the whole oracle-endorsed
-set (the flag form of `dotnet_inspect_style_full_taste`); it folds on top of any
-resolved config, so a per-knob line still refines it.
+set (the flag form of `dotnet_inspect_style_full_taste`); it applies after the
+config file resolves and wins for the knobs it covers.
 
 The lowered Annotated stage applies the byte-preserving knobs but never runs the
 lenses at all: they are raised-altitude sugar, and the lowered pipeline exists to
@@ -654,10 +654,12 @@ dotnet_style_prefer_conditional_expression_over_return = true
   `dotnet_style_qualification_for_field = false` is "full taste minus one knob".
 - `--taste` on `member` and `type` is the same aggregate as a one-invocation
   gesture, for asking a single question without leaving a config file behind. It
-  folds on top of the resolved config rather than replacing it, so a config that
-  already refines a knob keeps that refinement. Because the aggregate includes
-  the ternary lens, the Annotated view drops its interleaved IL for any member
-  that lens actually rewrites.
+  applies after the file resolves, so for the knobs the aggregate covers the
+  flag wins: an explicit gesture is never silently narrowed by a checked-in
+  `full_taste = true` + `dotnet_style_qualification_for_field = false`. Knobs
+  outside the endorsed set — the branchless lens — keep whatever the file
+  selected. Because the aggregate includes the ternary lens, the Annotated view
+  drops its interleaved IL for any member that lens actually rewrites.
 - The recognized keys are not hand-maintained in the resolver: they come from the
   library-owned `StyleOptionCatalog` (see [Option catalog](#option-catalog)), so
   the CLI vocabulary and the option surface cannot drift.
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -485,8 +485,7 @@ behavior-preserving; that is what lets the lens re-offer a fold the default had
 to decline. It deliberately stops at IDE0046: the further `c ? true : d` → `c ||
 d` collapse (IDE0075) is a separate future knob, so a literal-arm ternary such as
 `a ? true : b` is kept as written rather than simplified. The lens runs only on
-the opt-in raised path, after the default pipeline, and the IL-anchored Annotated
-view never applies it (it must stay byte-faithful for line/IL alignment).
+the opt-in raised path, after the default pipeline.
 
 The second lens is `PrinterOptions.PreferBranchlessBoolean`
 (`dotnet_inspect_style_prefer_branchless_boolean`). It targets the *same* declined
@@ -520,6 +519,39 @@ present, it would rebind to that operator's semantics), or valid but not branchl
 (a negation spelled as the ternary `(t ? false : true)` re-embeds a branch).
 Over-declining is always valid and faithful. When both lenses are enabled the
 oracle-endorsed ternary wins the shared shape.
+
+### Lenses and the Annotated view
+
+The Annotated view renders the same member as Decompiled Source, so it applies
+the same resolved style options: a member must not be spelled `this._count` in
+one section and `_count` in the other. For every byte-preserving knob this is
+free — the knob changes only the C# spelling, so the interleaved IL beneath each
+statement is byte-identical to the default render, and the view demonstrates the
+knob's contract rather than obscuring it.
+
+A style lens cannot keep that bargain. Its render no longer reproduces the
+member's opcodes, so anchoring the raw IL beneath it would assert a
+statement-to-opcode correspondence that does not hold — the single claim this
+view exists to make. When a lens **actually rewrites** (it records a
+`style-lens.*` decision under the `style-lens` category), the Annotated view
+therefore drops the interleaved IL and says so in the body, naming the applied
+lens and pointing at `-S "Applied Taste"`:
+
+```csharp
+// Interleaved IL suppressed: a byte-divergent style lens shaped this render.
+// Applied lens: style-lens.prefer-conditional-return.
+return a ? b : c;
+```
+
+Suppression keys on what the render *did*, not on what the host *asked for*: a
+lens that is enabled but finds no shape to rewrite is byte-faithful for that
+member, so its IL stays. The fact-comment overlay also stays, because a fact is
+a property of the member, not a claim about which opcodes a printed statement
+reproduces.
+
+The lowered Annotated stage applies the byte-preserving knobs but never runs the
+lenses at all: they are raised-altitude sugar, and the lowered pipeline exists to
+show the shape beneath that sugar.
 
 ## Names
 
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -534,20 +534,64 @@ member's opcodes, so anchoring the raw IL beneath it would assert a
 statement-to-opcode correspondence that does not hold — the single claim this
 view exists to make. When a lens **actually rewrites** (it records a
 `style-lens.*` decision under the `style-lens` category), the Annotated view
-therefore drops the interleaved IL and says so in the body, naming the applied
-lens and pointing at `-S "Applied Taste"`:
-
-```csharp
-// Interleaved IL suppressed: a byte-divergent style lens shaped this render.
-// Applied lens: style-lens.prefer-conditional-return.
-return a ? b : c;
-```
+therefore drops the interleaved IL for that member.
 
 Suppression keys on what the render *did*, not on what the host *asked for*: a
 lens that is enabled but finds no shape to rewrite is byte-faithful for that
 member, so its IL stays. The fact-comment overlay also stays, because a fact is
 a property of the member, not a claim about which opcodes a printed statement
 reproduces.
+
+### Reading applied taste in the Annotated view
+
+The applied knobs travel on the result as typed decisions, and the Annotated view
+renders them as a single trailing side comment on the member's signature — the
+same shape the fact overlay uses for analysis, so style and analysis read alike:
+
+```csharp
+public object? Pick(bool useName, string suffix)  // taste.qualify-property-access(Name); taste.qualify-field-access(_count)
+{
+    if (useName)
+        // IL_0000: ldarg.1                                              // arg: useName; stack: [bool]
+        // IL_0001: brfalse.s    IL_0009                                 // stack: []
+    {
+        return new object();  // alloc.new(object; alloc=System.Object; path=branch; path-confidence=behind-branch; post-dominance=return-post-dominates; escape=escapes; escape-kind=escapes-return; multiplicity=conditional)
+            // IL_0003: newobj       object::.ctor()  // alloc.new(object; alloc=System.Object; path=branch; path-confidence=behind-branch; post-dominance=return-post-dominates; escape=escapes; escape-kind=escapes-return; multiplicity=conditional); stack: [object]
+            // IL_0008: ret                                                  // stack: []
+    }
+    return string.Concat(this.Name, suffix, this._count.ToString());
+        // IL_0009: ldarg.0                                              // arg: this; stack: [Cache]
+        // IL_000A: call         Cache::get_Name()                       // stack: [string]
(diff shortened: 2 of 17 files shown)


richlander:

Two reviewers from different model families, each isolated in its own detached review worktree pinned to the exact head under review.

| Reviewer | First pass (`25be68f3`) | Fixed head (`4f1749f9`) |
| --- | --- | --- |
| Gemini 3.1 Pro | clean | clean |
| GPT-5.6 Sol | **1 blocking** | clean |

### Blocking finding (GPT-5.6 Sol) — fixed in 4f1749f9

**The lensed annotated render contaminated the style-invariant overlays.**

`ResearchViews.ProjectMember` imported one `IrFunction` and shared it across the Annotated Source, Cost Overlay, Semantics Overlay, and fact-row projections. `CSharpPrinter.PrintRaised` raises and rewrites the IR **in place**. This PR made the annotated projection the only one that runs byte-divergent style lenses, so an applied lens permanently mutated the shared graph and the overlays — which run after it and are supposed to be style-invariant — inherited the lensed shape. Selecting one section changed a different section's output.

I reproduced it independently on a clean exact-head worktree before acting on it, with a `/tmp` fixture and `.dotnet-inspectconfig` setting `dotnet_style_prefer_conditional_expression_over_return = true`:

```console
$ dotnet-inspect member Fx.dll Fx.Probe GuardAlloc -S "Signature,Cost Overlay"
## Cost Overlay
public static bool GuardAlloc(bool a, string s)
{
    if (a)
    {
        return string.Concat(s, "x").Length > 1;
    }
    return s.Length > 0;
}

$ dotnet-inspect member Fx.dll Fx.Probe GuardAlloc -S "Signature,Cost Overlay,Annotated Source"
## Cost Overlay
public static bool GuardAlloc(bool a, string s)
{
    return a ? (string.Concat(s, "x").Length > 1) : (s.Length > 0);
}
```

Blast radius, measured rather than assumed: `Cost Overlay` and `Semantics Overlay` were affected. `Decompiled Source`, `Facts`, and `IL` were not — they do not share this instance.

**Fix (`4f1749f9`):** the annotated render gets its own freshly imported `IrFunction`, so the isolation holds regardless of projection order rather than depending on which projection happens to run last. Same visible failure on a null re-import, inside the same handler. Regression test `ByteDivergentLens_DoesNotLeakIntoTheStyleInvariantOverlays` compares overlay output with and without `AnnotatedSource: true` under an applied lens, and asserts the lens actually fired so the test cannot pass vacuously.

Verified after the fix on the same fixture: both invocations now print the flat control flow.

### Verified and dismissed

Both reviewers probed and cleared the following; nothing actionable came out of them.

- **Suppression keys on applied, not requested.** Both lens passes increment `Rewrites` only on a successful `ReplaceWith`, and the decision is recorded only when `Rewrites > 0`. A lens enabled but never triggered leaves the member byte-faithful and keeps the IL — pinned by `ByteDivergentLens_WhenRequestedButDeclined_KeepsInterleavedIl`. No false-positive path found either.
- **Nested/lambda bodies** imported through the cross-method seam are part of the traversed IR tree, so an inner rewrite is reflected in the counter.
- **`PrintLowered` excluded from lens passes** is coherent: lenses are raised-altitude sugar, and the lowered view exists to show the unsugared shape.
- **`RaiseWithStyleLenses` refactor** is behavior-identical to the two former inline copies — same try/catch boundary, same pass order. The only caller that supplies options to the line-map overload is the annotated projection, which opts in deliberately.
- **Fact anchoring across the two imports** holds: annotations carry IL `SourceOffset`, nothing depends on `IrNode` reference identity, and under an applied lens a fact correctly follows to the rewritten statement. Checked against real `callvirt`/allocation facts on fact-bearing members.
- **The suppression note** stays valid C# comments inside the body in every render mode, and the empty-IL `CorrelateMixedSource` path drops no facts and throws no index errors.
- **The config-warning latch** emits exactly once for an Annotated-Source-only run and does not double-emit when combined with other sections.
- **Default-output invariance** with no options, confirmed by the suites plus a section-combination matrix.

### Non-actioned, filed as follow-ups

Pre-existing and out of scope for this PR; none are introduced or worsened here.

- `var-spelling-style` knobs are catalogued and bound to `PrinterOptions` but never read by `CSharpPrinter` — config is accepted and silently does nothing.
- `-S "Annotated Source" --json` yields zero members, identically with and without this change.
- `Decompiled Source` renders expression-bodied members where `Annotated Source` renders a block body; declaration shaping lives off the IR path in `MemberBodyProducer`.

### Validation on `4f1749f9`

| Suite | Result |
| --- | --- |
| `ILInspector.Research.Tests` | 132 passed |
| `dotnet-inspect.Tests` | 2132 passed, 12 skipped (no `ilasm`) |
| `ILInspector.Decompiler.Tests` (fast) | 3781 passed |
| CI | `build-net10`, `test`, `csharp-diff-smoke`, `markdownlint`, `changes` all green |

The full decompiler suite was run on the prior head: 6 failures, all reproduced on a pristine `origin/main` worktree with zero changes (compile-back harness cannot resolve corelib on this machine). Not regressions.

richlander:
Ready to merge

richlander:
## Adversarial review reconciliation (rounds 2-4)

Reviewers: **Gemini 3.1 Pro** and **GPT-5.6 Sol**, each in an isolated detached worktree under `/tmp/`, given the same self-contained prompt. Round 1 (on the original prose-block implementation) is reconciled in the earlier comment; this covers the light-touch rework.

Both signed off on **`55380643`**.

### Findings and resolutions

| # | Finding | Found by | Resolution |
| --- | --- | --- | --- |
| 1 | Comment injection: unescaped `DecompilerDecision.Subject` in a `//` comment | **Both, independently** | Fixed in `e473a5a0` |
| 2 | `--taste` overrides explicit config refinements, contradicting the doc | **Both, independently** | Doc corrected in `b8531c5a`; code kept - see below |
| 3 | Programmatic Research consumers lose the suppression signal | Gemini | Contract documented at the suppression site in `e473a5a0` |
| 4 | ANSI / control / bidi characters still pass through raw | GPT-5.6 | Fixed in `6c219cbe` |
| 5 | Bidi set omits U+061C ARABIC LETTER MARK | GPT-5.6 | Fixed in `55380643` |

### 1 - Comment injection (both reviewers)

The CLR does not require metadata names to be valid C# identifiers, so a hostile assembly can declare a field named `field\n    public int Injected() => 42; //`. That ended the comment and left the remainder as **active code** inside a C# fenced block:

```csharp
public int GetCount()  // taste.qualify-field-access(field
    public int Injected() => 42; //)
```

I reproduced this myself end-to-end with a `PersistedAssemblyBuilder` fixture before acting on it. Fixed at the single choke point where the annotation is built; a 6-case theory covers every terminator C# recognizes.

Note GPT's alternative suggestion - use the sanitized C# spelling instead of the raw subject - was **not** taken. The raw subject is the metadata identity, and a metadata tool should report the name it actually saw; the C# printer's identifier mangling is a separate concern. Gemini independently endorsed keeping the raw subject.

### 2 - `--taste` precedence: the one place I diverged from both reviewers

Both proved the code contradicted the documented contract, and both recommended changing the **code** (Gemini suggested tri-state `bool?` in `PrinterOptions` to distinguish explicit-false from default-false).

I concluded the behavior was right and the **claim** was wrong, so I corrected the claim. `--taste` is an explicit per-invocation gesture; `.dotnet-inspectconfig` is ambient checked-in state. Letting an ambient file silently narrow an explicit flag violates least astonishment and leaves no way to say "really, all of it". The tri-state alternative is a broad design change to a type used well beyond this feature.

I put this back to both reviewers explicitly, inviting them to argue for config-wins. **Both agreed with flag-wins on reconsideration.** The corrected contract - flag wins for knobs the aggregate covers; knobs outside the endorsed set keep their config value - is now in `docs/decompiler-taste.md` and the `RunPreamble` comment, pinned by two tests.

### 4 and 5 - Rendering hazards

GPT found that folding line terminators stopped syntax breakout but left ANSI escapes and bidi overrides intact, and later that the bidi set omitted U+061C. Gemini had cleared both at the respective heads - a clean demonstration of the two-model requirement earning its cost.

Treatment is split by hazard: line terminators fold to a space (keep the comment on its line); remaining control characters and the 12 `Bidi_Control` code points become visible `\uXXXX` escapes (preserve identity). Deliberately narrower than the `Cf` category - a zero-width joiner or BOM does not reorder its neighbors, and legitimate identifiers may contain format characters. Both reviewers independently confirmed that boundary and verified surrogate pairs, CJK/Cyrillic identifiers, and compiler-generated names pass through byte-identical.

One correction for the record: Gemini's round-2 sign-off asserted `ReplaceLineEndings` folds vertical tab (U+000B). I verified that is **false**. VT is not a C# line terminator so it never broke the comment, but it does move the terminal cursor - it is now caught by the control-character rule.

### Non-actioned, deliberately

- **#3257** - the same injection class exists **pre-existing** in the interleaved IL operand text, and more broadly in metadata-name rendering (on base, `-S "Fields"` emits a raw ESC and bidi override). Different layer, affects default output, wider blast radius. Filed with reproductions rather than widened into this PR. **Both reviewers explicitly agreed it does not block this PR.**
- Pre-existing follow-ups carried over from round 1 and still not actioned: `var-spelling-style` knobs are catalogued and bound but never read by `CSharpPrinter`; `-S "Annotated Source" --json` yields 0 members; `Decompiled Source` renders expression-bodied where `Annotated Source` renders a block body.

All are non-blocking follow-up work.

richlander:
Ready to merge

richlander:
Follow-up `204b59c8` (documentation-only): aligned the `RequestAllTaste` XML doc with the decided **flag-wins** precedence. It still carried the pre-correction "a per-knob line in a config file still refines it" framing — the exact ambiguity reconciliation finding #2 corrected in `ApiCommand`, `docs/decompiler-taste.md`, and the two `RenderStyleConfigTests`. No behavior change; `ApplyFullTaste` and its pinning tests are untouched, so the fixed-head adversarial sign-offs still stand. Re-affirming **Ready to merge** once CI is green on the new head.

richlander:
Ready to merge — CI green on `204b59c8` (test 8m34s, build/csharp-diff-smoke/markdownlint/changes pass; il-diff-smoke/pack path-gated). The follow-up is documentation-only, so both fixed-head adversarial sign-offs still stand.

Issue #3238, opened 2026-07-26
Annotated compile-back failure render: emitted C# + caret + rune-reveal + cause classification


richlander:
## Motivation

Compile-back `RecompileFail` results today record only `Id: message` (e.g. `CS0539: ... not found`). The *interesting* information — **which emitted qualifier bound to the wrong thing, and why** — is invisible. A reviewer has to hand-reconstruct the failing line. In PR #3222 that walkthrough had to be authored by hand, e.g.:

```csharp
class Seq : Good.IProbe
{
    void Good.IProbe.M‹ZWNJ›() { }
    //             ^^^^^^^^^^^ Roslyn STRIPS U+200C when binding this identifier,
    //                         so it binds to `Good.IProbe.M` — but the interface
    //                         member is `M‹ZWNJ›`, which no longer matches.
}
```

The raw materials to generate this already flow through the harness — see `tools/DecompilerHarness/FidelityCheck.cs:1342-1350`, where each failure is built from the emitted source (`e.OrigText`) plus the Roslyn `Diagnostic`. We just discard the diagnostic `Location` in `FormatDiagnostic` (`FidelityCheck.cs:1986`).

## Proposal

Add an annotated-failure render mode (e.g. a `--explain` flag, or an extension of the existing `CB_DUMP` env path at `FidelityCheck.cs:1343`) that emits, per `RecompileFail`:

1. **Emitted C# with a caret under the failing span.** Mechanical: the `Diagnostic.Location.GetLineSpan()` gives the exact line + character span — the same data `csc` uses to draw its squiggle. Roslyn's `CSharpDiagnosticFormatter` can render source-line + caret directly. This layer is buildable today with zero product risk (harness-reporting only) — it just stops discarding the location.
2. **Rune-reveal for invisible/confusable characters** (`‹ZWNJ›`, combining marks) in the offending span. Mechanical: scan for `UnicodeCategory.Format` / `NonSpacingMark` / control runes and substitute a visible token. The harness already computes those exact categories in `ReturnToSenderTypePlanner.MetadataIdentifierRoundTrips`.
3. **A cause-classification line** ("why the emitted identifier bound elsewhere") for an **enumerated** set of known classes, reusing the harness/product predicates the gate already owns:
   - Unicode format (Cf) stripping (CS0539/CS0246)
   - keyword escaping
   - unspeakable-name sanitization (`<Bad>` → `__Bad_`)
   - namespace/type shadow (CS0426/CS0538)
[...]

PR #3239, opened 2026-07-26
Add render walkthrough section to the compile-back harness PR template

richlander:
## Change

Adds a **Reconstructed render (Original → Before → After → Fully raised)** section to `docs/templates/decompiler-compile-back-harness-pr.md`.

Compile-back `RecompileFail`s live in **one emitted qualifier or statement** whose C# binding diverges from its metadata spelling — an aggregate bucket count can't show that. This section walks the emitted C# in the same Original → Before → After → Fully raised shape used by the decompiler-PR template, with per-block **compile-back verdicts** (`Exact` / `OpcodeDiff` / `OperandDiff` / `RecompileFail {CSxxxx}` / `ContextFail {bucket}`) and Valid/Correct/IL-fidelity recorded next to the code they judge, so a reviewer sees the exact failing line.

The pre-existing text form is kept and labelled **Aggregate before/after**, with guidance on when to use each (aggregate bucket move vs. a specific reconstructed method).

Motivated by the PR #3222 worked-example review comment, which had to author this walkthrough by hand. Forward-references #3238 (generating the annotated render — emitted C# + caret + rune-reveal + cause line — directly from the harness).

## Evidence

- Docs-only; no product build or tests.
- `markdownlint-cli` clean on the changed file.

## Risk

Low — template guidance only. No behavior claim.

### DIFF
--- a/docs/templates/decompiler-compile-back-harness-pr.md
+++ b/docs/templates/decompiler-compile-back-harness-pr.md
@@ -65,6 +65,18 @@ harness output.)
   fallback (e.g. the sanitized ContextFail floor) when the shape is declined
   rather than raised.
 
+When a block IS a declined-to-floor shape, render the ACTUAL floor, not the
+shape that was declined — a decline changes the member's IDENTITY, not its body.
+It is not a skeleton: the real reconstructed body is kept. What changes is that
+the interface is dropped from the base list and the member is demoted from an
+explicit implementation to a plain method whose name is the sanitized MethodDef
+name (dots become underscores, `IType.Member` -> `IType_Member`). Its verdict is
+`ContextFail` ("method-not-found"): the plain shape compiles, but the original
+explicit-member identity no longer exists in the recompiled assembly to
+opcode-compare — it is chosen precisely because it is strictly safer than the
+`RecompileFail` the explicit spelling would have produced. Do not draw the floor
+still carrying the interface base entry or the explicit qualifier.
+
 Record the compile-back verdict next to the code it judges (never inferred from
 prose):
 
--- a/docs/templates/decompiler-compile-back-harness-pr.md
+++ b/docs/templates/decompiler-compile-back-harness-pr.md
@@ -18,6 +18,16 @@ difference after, and which frontier remains. Delete sections that do not apply.
 **Conclusion:** **PASS/REVIEW/BLOCKED** — {one sentence naming the
 uncheckable bucket, the fix, and the decisive evidence}.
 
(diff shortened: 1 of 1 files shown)


richlander:
Added a follow-up commit capturing the key lesson from the now-merged #3222: how to render a **declined-to-floor** shape. That was the one spot the #3222 worked-example got wrong — the floor is counterintuitive (interface dropped from the base list, member demoted to a sanitized flattened-name plain method, real body kept, `ContextFail` "method-not-found" chosen *because* it's strictly safer than a `RecompileFail`). The guidance now tells authors to draw the actual floor, not the declined shape. Docs-only, markdownlint clean.

richlander:
## Adversarial review reconciled (MAI, docs-only) — Ready to merge

Per the MAI-only scope for this docs change, one CLEAN MAI verdict at the fixed head.

### MAI finding (round 1 @ `2065a506`) — VALID in substance, reviewed stale code
MAI flagged that the "declined-to-floor" guidance stated the interface is "dropped from the base list," but its cited implementation (`ReturnToSenderTypePlanner.cs:1311-1385`, `1924-1945`) only cleared `ExplicitInterfaceMemberName`/`DeclarationSignature` without a base-list drop.

Root cause of the mismatch: **the docs branch was cut from `99bcbaf7`, before PR #3222 merged**, so the review worktree lacked the external explicit-interface reconstruction gate entirely. The cited line numbers point at the pre-#3222 planner (same-assembly revert path only).

The underlying point was still correct: the original wording **overgeneralized**. Current `main` has two distinct decline paths:
- **External** explicit-interface decline — `ExternalInterfaces = []` (`ReturnToSenderTypePlanner.cs:2614`), so the interface **is** dropped from the base list. This is the #3112/#3222 worked-example case the walkthrough illustrates.
- **Same-assembly** revert — clears only the member's `ExplicitInterfaceMemberName`/`DeclarationSignature` (`ReturnToSenderTypePlanner.cs:2631-2652`); no base-list drop.

### Resolution
- Merged `origin/main` into the docs branch (`c990fc73`) so the reviewer inspects the real post-#3222 product code the template describes.
- Reworded the guidance (`b3bb5258`) to distinguish both mechanisms explicitly, scoping the base-list drop to the external-interface case.

### Re-review @ `b3bb5258` — **MAI: CLEAN**
After re-reviewing against the current planner code, MAI confirmed the reworded paragraph accurately describes both decline mechanisms. Docs-only, markdownlint clean.

**Ready to merge.**

Issue #3242, opened 2026-07-26
Census and consolidate the divergent CFG/predecessor/successor models


richlander:
Part of #3240.

## Problem

Several independent derivations of the same control-flow facts coexist. Because
each is a separately maintained *semantic model*, they can disagree, and nothing
detects it.

Observed during review:

- **Two CFG edge models.** `src/ILInspector.Decompiler/Pipeline/Cfg.cs` computes
  its own edges while reusing the `BlockEdges` *type* from
  `src/ILInspector.ControlFlow/BlockEdges.cs`. Same type, different derivation.
- **~8 sites reason about predecessors and ~12 compute successors**, spread
  across `Cfg.cs`, `IrImporter.cs`, `DataflowFacts.cs`, `DefiniteAssignment.cs`,
  `StackSlotCopyPropagationPass.cs`, `SwitchRaisingPass.cs`, `StructuringPass.cs`,
  and `GenericDeclarationPatternProof.cs`.
- **A concrete divergence exists.** `Cfg.cs` treats `Leave`/`EndFinally`/`EndFilter`
  as having no successors; `StackSlotCopyPropagationPass.Successors` yields the
  leave target for `Leave`. They differ precisely on `Leave`.

The `Leave` divergence is currently **inert** — instrumented counters over
406,290 methods show the `Leave` arm is hit 0 times, because two guards
(`IsIndependentBodyScope` and `ContainsUnmodeledControlFlow`) happen to exclude
every case. That is inertness by *accident of guard interaction*, not by
contract, and nothing pins it.

## Related observation

`StackSlotCopyPropagationPass` carries its own ~60-line CFG model and, over the
same 406,290 methods, performs **3 propagations total** (0 on CoreLib). It is
either under-tested or not carrying its weight; worth justifying with fixtures
or retiring. Retiring it removes one competing CFG model.

Separately, `src/ILInspector.ControlFlow/Dominators.cs` and `PostDominators.cs`
have **zero consumers in `ILInspector.Decompiler`** — consumers are
`ILInspector.Analysis`, tests, and a harness probe. There is an owned dominator
implementation the decompiler does not use.

## Proposed work

1. Census every predecessor/successor/edge derivation and record what each needs.
2. Decide the owner. `ILInspector.ControlFlow` already owns edges and dominators;
   the question is whether decompiler passes should consume it rather than
   rebuild.
3. Add a differential check (harness-side is fine) asserting the models agree on
[...]

Issue #3240, opened 2026-07-26
Tracking: derived-predicate divergence and unenforced IR invariants


richlander:
## Purpose

Tracking issue for one recurring structural theme found during an adversarial
review of the decompiler: **a fact derived in two places that can disagree.**

This is not a DRY/style concern. Each instance is a *semantic model* rebuilt
independently, so the copies can drift and no test notices. #3236 is the first
confirmed crash from this class, and the review found the enforcement mechanism
that should catch such drift does not execute in the configuration we run.

This docket exists so the findings stay actionable; it does not itself
implement fixes.

## Instance already fixed: #3236

`StructuringPass` derived "is this block an inlinable terminator?" in two
places: `IsInlinableTerminator` re-computed the predicate, while `BuildRegion`
indexed the `TerminatorSnapshots` table the predicate was supposed to describe.
One of the two `Ctx` factories populated every field the predicate reads but
left the table empty, so the guard said yes and the data was absent:

```
KeyNotFoundException: The given key '3' was not present in the dictionary
  StructuringPass.BuildRegion <- TryBuildPastRegionTarget <- CanInlinePastRegionTarget <- Validate
  Google.Apis.Requests.ClientServiceRequest`1.<ParseResponse>d__38::MoveNext
```

It crashed inside `Validate` — the phase whose job is to prove a shape is safe
before any rewrite. Fixed by making the guard read the table it gates.

## Why this class is currently invisible

The IR has a structural invariant checker, `IrNode.CheckInvariant()`
(`src/ILInspector.Decompiler/Pipeline/Ir/IrNode.cs:180`). Its own doc comment
says *"Debug builds run this after every pass; a violation is a pipeline bug,
never input data."*

It is `[Conditional("DEBUG")]`, and `AGENTS.md` prescribes running tests in
Release:

```
dotnet run --project src/ILInspector.Decompiler.Tests -c Release
```

`[Conditional]` strips the **call site** when the *calling* assembly lacks the
symbol. Measured on the built test assembly:

| Configuration | MemberRefs to `CheckInvariant` |
| --- | --- |
| `-c Release` (the prescribed run) | **0** |
| `-c Debug` (positive control) | 1 |

There are **617 `CheckInvariant()` call sites across 121 test files**. In the
configuration we actually run, every one of them is compiled away. The suite
reads as though it asserts IR integrity 617 times and asserts it zero times.

(Method: read the emitted metadata of
[...]

PR #3248, opened 2026-07-26
Consolidate StackSlotCopyPropagationPass onto the shared Cfg.Build model

richlander:
## What

`StackSlotCopyPropagationPass` carried its own ~60-line successor/predecessor
derivation that duplicated `Cfg.Build`'s structural edges and **diverged from it
on one opcode**: the private model resolved a `Leave` terminator's target as a
successor, while `Cfg.Build` reports `Leave` as leaving the region with no
successor. Nothing detected the disagreement.

This replaces the private `Successors`/`Predecessors`/`IndexOfOffset` with the
single shared `ILInspector.ControlFlow.BlockEdges` produced by `Cfg.Build`, and
bails when the graph is incomplete (`LeavesRegion` or `ExternalTargets`) —
matching the contract `DefiniteAssignment` already uses.

`Part of #3240. Addresses the CFG-model divergence in #3242` (census posted on
the issue).

## Why it is behavior-preserving

The pass only runs on **flat, EH-free bodies**: `HasUnmodeledControlFlow` bails
on any nested `Block`/`BlockContainer`, and EH regions *are* nested containers.
So no block it processes ever ends in `Leave`, and no branch leaves the
container. Over exactly those bodies, `Cfg.Build` and the old private model
compute **edge-for-edge identical** successors (`Branch`, `ConditionalBranch`,
`SwitchBranch`, `Return`, `Throw`, fall-through — all handled identically; only
`Leave`/external differ, and neither can occur). The `Leave` divergence was inert
only by accident of guard interaction; #3242's instrumentation measured **0 leave
hits over 406,290 methods**. The new explicit bail turns that accidental
exclusion into a pinned invariant and removes the divergent model by
construction — so no separate differential check is needed to watch the two
models agree.

## Evidence

- `StackSlotCopyPropagationPassTests`: **22/22 pass** (includes positive
  propagation cases — the pass still does its work).
- Fast decompiler suite (`-trait- "Speed=Slow"`): **3786/3787**; the single
  failure (`ExpressionTreeLambdaTests.LookalikeExpressionAssembly_StaysFactoryCalls`)
  is a pre-existing environmental red (fixture assembly not built when building
  only the test csproj) — verified identical on the untouched base with the change
  stashed.
- Diff: −64 / +21 (removes the private CFG model).

## Risk

Low, mechanical refactor: one equivalent edge derivation swapped for the shared
one, provably identical on the processed inputs, strictly safer on the
(unreachable) incomplete-graph path. No output shape, section, or verbosity
behavior changes.

### DIFF
--- a/src/ILInspector.Decompiler/Pipeline/Passes/StackSlotCopyPropagationPass.cs
+++ b/src/ILInspector.Decompiler/Pipeline/Passes/StackSlotCopyPropagationPass.cs
@@ -1,3 +1,5 @@
+using ILInspector.ControlFlow;
+
 namespace ILInspector.Decompiler.Pipeline;
 
 /// <summary>
@@ -101,6 +103,7 @@ public sealed class StackSlotCopyPropagationPass : IIrPass
 
     sealed record CopyPath(
         IReadOnlyList<Block> Blocks,
+        IReadOnlyList<BlockEdges> Edges,
         int CopyBlockIndex,
         int CopyStatementIndex,
         int UseBlockIndex,
@@ -121,6 +124,16 @@ public sealed class StackSlotCopyPropagationPass : IIrPass
 
         var blocks = container.Blocks;
         if (HasUnmodeledControlFlow(blocks))
+            return null;
+
+        // Consume the single shared structural CFG rather than a private edge
+        // model. Cfg.Build reports EH leaves and out-of-container branch targets
+        // as flags instead of resolving them, so bail on any incomplete graph —
+        // the same contract DefiniteAssignment uses. This turns the previously
+        // accidental exclusion of leave edges (a side effect of guard
+        // interaction) into an explicit, pinned invariant.
+        var edges = Cfg.Build(blocks);
+        if (edges.Any(edge => edge.LeavesRegion || edge.ExternalTargets.Count > 0))
             return null;
 
         int copyBlockIndex = copyBlock.ChildIndex;
@@ -159,12 +172,12 @@ public sealed class StackSlotCopyPropagationPass : IIrPass
             if (state.BlockIndex == useBlockIndex)
                 continue;
 
-            foreach (int successor in Successors(blocks, state.BlockIndex))
+            foreach (int successor in edges[state.BlockIndex].Successors)
                 stack.Push((successor, 0));
         }
 
         return foundUse
-            ? new CopyPath(blocks, copyBlockIndex, copy.ChildIndex, useBlockIndex, useStatement, region)
+            ? new CopyPath(blocks, edges, copyBlockIndex, copy.ChildIndex, useBlockIndex, useStatement, region)
             : null;
     }
 
@@ -203,7 +216,7 @@ public sealed class StackSlotCopyPropagationPass : IIrPass
                 return false;
 
             state[blockIndex] = 1;
-            foreach (int successor in Successors(path.Blocks, blockIndex))
+            foreach (int successor in path.Edges[blockIndex].Successors)
                 if (Visit(successor))
                     return true;
             state[blockIndex] = 2;
@@ -213,7 +226,7 @@ public sealed class StackSlotCopyPropagationPass : IIrPass
 
     static bool RegionDominatedByCopy(CopyPath path)
     {
-        var predecessors = Predecessors(path.Blocks);
+        var predecessors = Predecessors(path.Edges);
         foreach (int blockIndex in path.RegionBlocks)
         {
             if (blockIndex == path.CopyBlockIndex)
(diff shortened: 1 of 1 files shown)


richlander:
## Adversarial review reconciled — Ready to merge

Two-model review at the fixed head `3b090d0d`. Both **CLEAN**, CI green (build, `test (ubuntu-26.04, linux-x64)`, `csharp-diff-smoke` all pass).

### Opus — CLEAN
Verified case-by-case against base `a71b65c3`, `Cfg.cs`, and `BlockEdges.cs`; built with `--configfile nuget.config` and ran `StackSlotCopyPropagationPassTests` (22/22).
- **Successor parity:** identical for every surviving terminator (empty→`i+1`, `Branch`→target, `ConditionalBranch`→`[target, i+1]`, `SwitchBranch`→`[targets…, i+1]` with duplicates preserved, `Return`/`Throw`→none), same order; no consumer depends on order anyway.
- **New bail is inert, not a lost propagation:** `LeavesRegion`/`ExternalTargets` only arise inside EH regions / out-of-container targets — nested `BlockContainer`s already excluded by `HasUnmodeledControlFlow`, which runs first.
- **No crash risk:** empty children handled; structured last-children can't reach `Cfg.Build`; the only theoretical divergence (duplicate `StartOffset`: old first-match vs dict last-write) can't occur on well-formed flat IL and was arbitrary in the old code too.
- Predecessor inversion equivalent; deleted helpers have no remaining callers (build clean).

### Gemini — CLEAN
Independently reached the same conclusions, adding an **ECMA-335 argument**: `IsIndependentBodyScope` restricts to top-level containers (so branching out = leaving the method = invalid IL → `ExternalTargets` never > 0), and a flat method with no EH cannot legally contain `Leave`/`EndFinally`/`EndFilter` (a `leave` is only valid exiting a protected region, which is a nested container `HasUnmodeledControlFlow` excludes). `StartOffset`s map 1:1 to unique IL offsets, so the first-vs-last resolution difference is unobservable. Also notes the swap trades an O(N²) per-edge `IndexOfOffset` search for a single O(N) cached graph build.

No findings from either reviewer; no fixes required. Review worktrees isolated under `/tmp`, no tracked-file edits.

**Ready to merge.**

richlander:
Reviewed at head `3b090d0d` in an isolated worktree. **Approve** — the behavior-preservation claim holds, and I verified it by measurement rather than by re-reading the argument. Two notes worth folding in, one of which is a gap in the stated reasoning (not in the code).

## The argument has a hole the measurement closes

The PR frames the divergence as `Leave`-only. `Cfg.Build` actually maps **three** terminators to `LeavesRegion = true`:

```csharp
case Leave or EndFinally or EndFilter:
    leaves = true;
```

The old private model put `EndFinally`/`EndFilter` in the *same arm as* `Return`/`Throw` — ordinary no-successor exits, which the pass happily processed. So the new bail is triggered by three terminators, but the cited evidence (the #3242 instrumentation, 0 `Leave` hits / 406,290 methods) only covers one of them.

"EH regions are nested containers, so `HasUnmodeledControlFlow` bails" doesn't formally close it either: `ContainsUnmodeledControlFlow` returns true for `Block`/`BlockContainer`/`Break`/`Continue`, and a bare `EndFinally` statement is none of those. A body whose EH structuring degraded could in principle present a flat container with an `EndFinally` terminator, which the old code processed and the new code declines.

## Measured, so this is settled

Instrumented the new bail at this head with per-cause counters and ran `--gaps` over the 364,338-method EVIL pool:

```
reached      3
(no leave, endfinally, endfilter, leaves-other, or external rows)
```

Zero bails of any cause. Edge-for-edge identical on this corpus — **no coverage regression**, including for the `EndFinally`/`EndFilter` case the PR doesn't argue.

Instrument validity: my first attempt wrote no file at all, which is ambiguous between "never reached" and "handler never registered." Re-registering via `[ModuleInitializer]` makes the file appear unconditionally, so `reached = 3` is a real reading and not a dead probe.

Suggest adding `EndFinally`/`EndFilter` to the "Why it is behavior-preserving" section so the next reader doesn't have to rediscover that they're now bail causes.

## The PR undersells a real correctness fix

The old `Successors` silently dropped an unresolvable branch target (`IndexOfOffset` → `-1` → no edge emitted). That missing edge flows into `Predecessors`, and `RegionDominatedByCopy` reasons over predecessors — a block with a dropped incoming edge can look dominated by the copy when it isn't, which is an **unsound propagation**, not just a missed one. The `ExternalTargets` bail closes that hole by construction.

That's a correctness improvement, not a pure refactor. Worth saying so; "low, mechanical" undersells it.

## The strongest argument for this PR isn't in the PR

`reached = 3` over 364,338 methods. The pass's entire control-flow apparatus — DFS, cycle check, dominance, predecessor derivation — engages three times per full corpus sweep.

That means a differential check between the two models would have had **three samples**. It could never have detected the `Leave` disagreement, and it can't detect the next one. So the case for this change isn't "no separate differential check is needed" — it's that **a differential check is not available at this frequency, and deleting the second model by construction is the only assurance obtainable.** That's a stronger and more honest framing.

It also sharpens the retire-vs-keep question in #3242 rather than answering it: this is ~330 lines of CFG machinery that fires 3 times in 364K methods, with 22 unit tests as its only meaningful coverage. Consolidating first is the right order of operations — smaller and safer either way — but the value question stays open.

## Minor

- No perf concern. `Cfg.Build` is called per surviving candidate, but the old path was worse: `IndexOfOffset` is a linear scan per edge query inside a DFS, i.e. O(blocks²) per candidate, versus one dictionary-backed O(blocks) build with O(1) lookups. Likely a small improvement.
- `Cfg.Build` runs before the `copyBlockIndex`/`useBlockIndex` validity checks, so it can be built and discarded. Irrelevant at 3 calls/corpus; noting only so it isn't mistaken for intent.
- The reported `3786/3787` with an `ExpressionTreeLambdaTests` red is an artifact of building only the test csproj. Building `dotnet-inspect.slnx -c Release` first clears it — I get a clean build and `StackSlotCopyPropagationPassTests` 22/22. Worth re-running to drop the asterisk from the evidence.

## Verification performed

| Check | Result |
| --- | --- |
| `dotnet build dotnet-inspect.slnx -c Release` @ `3b090d0d` | 0 errors |
| `StackSlotCopyPropagationPassTests` | 22/22 |
| Instrumented bail, `--gaps` over EVIL pool (364,338 methods) | `reached=3`, **0 bails of any cause** |
| `Cfg.Build` vs removed `Successors`, arm by arm | identical for empty-children, `Branch`, `ConditionalBranch`, `SwitchBranch`, `Return`, `Throw`, non-terminator fall-through; differs only for `Leave`, `EndFinally`, `EndFilter`, and unresolved targets — all measured absent |

richlander:
Re-checked. **Approval stands** — and now with merge-path evidence it didn't have before.

The head hasn't moved since my review (`3b090d0d`, `+21/-64`). But `main` has advanced 7 commits underneath it, including **#3234 (tuple-swap raising)** — a *pass* change that alters what reaches `StackSlotCopyPropagationPass`. My original approval rested on measuring 0 bails over 364,338 methods on the *old* base, so that number was stale. I re-ran on the actual merge result rather than re-assert it.

**Merge path:** `origin/main a25ef6aa` + `3b090d0d` → `1e7c8917`. Clean merge, still one file.

| check | result |
| --- | --- |
| build (`slnx`, Release) | 0 errors |
| `--gate fast` | **3813 passed / 0 failed** |
| CoreLib corpus snapshot vs baseline | **byte-identical over 41,952 methods** |

The snapshot check is the one that matters. I built the harness on plain `origin/main` and on the merge result, ran `--emit-corpus-snapshot` over `System.Private.CoreLib.dll` on each, and diffed the two 27 MB JSONs:

```
$ diff <(json.tool snap-base.json) <(json.tool snap-pr.json)
4c4
<     "generatedUtc": "...:44:10..."
---
>     "generatedUtc": "...:44:36..."
```

Four lines of diff, all of it the timestamp. Every per-method structuring and fidelity record across 41,952 methods is identical. So the consolidation is behavior-neutral on real code *after* the tuple-swap pass landed, not just before it.

Restating the substance from my earlier review, since it's the reason this is worth merging rather than merely safe to merge:

- **The new bail has three triggers, not one.** `Cfg.Build` maps `Leave or EndFinally or EndFilter` → `leaves = true`, whereas the removed private model put `EndFinally`/`EndFilter` in the same arm as `Return`/`Throw`. The PR argues the `Leave` case; `EndFinally`/`EndFilter` come along silently. The measurements above are what close that gap — it's covered by evidence, not by the argument.
- **This is a correctness fix, not a refactor.** The old `Successors` dropped unresolvable targets (`IndexOfOffset` → −1) without a word. A missing predecessor edge feeds `RegionDominatedByCopy`, which could then report a block dominated when it isn't — unsound propagation. `ExternalTargets` turns a silent wrong answer into a bail.
- **The consolidation argument is stronger than a differential check would have been.** `reached = 3` in 364K methods means a side-by-side comparison of the two models would have had three samples to work with. It could never have surfaced the `Leave` disagreement. Deleting the second model is the only thing that actually removes the divergence.
- Perf should improve if anything: the old `IndexOfOffset` was a linear scan per edge query inside a DFS, i.e. O(blocks²) per candidate.

Ready to merge from my side.

Issue #3176, opened 2026-07-25
Add a "full taste" style aggregate: enable the oracle-endorsed knob subset at once


richlander:
## Summary

Add a single opt-in — the **"full taste" aggregate** — that enables the *oracle-endorsed* subset of the opt-in style knobs together, instead of forcing a user to copy each `dotnet_style_*` line individually. This is the capstone of the #3138 opt-in style-lens arc: #3160 built the library-owned `StyleOptionCatalog` (which already exposes the `OracleEndorsed` filter and `ConflictGroup` this aggregate consumes), and #3171 finished the bool-knob surface.

## What "full taste" means

Exactly the `OracleEndorsed` subset of `StyleOptionCatalog.Options` — the spellings/shapes the dotnet/runtime `.editorconfig`/IDE oracle prefers:

- `qualify-field-access` (`dotnet_style_qualification_for_field`)
- `qualify-property-access` (`dotnet_style_qualification_for_property`)
- `qualify-method-access` (`dotnet_style_qualification_for_method`)
- `qualify-event-access` (`dotnet_style_qualification_for_event`)
- `prefer-conditional-expression-return` (`dotnet_style_prefer_conditional_expression_over_return`)

It deliberately **excludes** the branchless "bool hack" lens (`prefer-branchless-boolean`), which is a user preference the oracle does not endorse.

## Conflict-group resolution falls out for free

The `guarded-boolean-return` conflict group has two members: the ternary lens (oracle-endorsed) and the branchless lens (not). Because the aggregate only enables the oracle-endorsed subset, it always picks the ternary and never the branchless lens — so there is **no intra-aggregate conflict** by construction; resolution is deterministic.

## Proposed surface

1. **Library (`StyleOptionCatalog`)** — the discoverable source of truth for CLI + Wasm hosts:
   - `OracleEndorsedOptions` — the filtered descriptor subset (what a UI shows as "full taste" members).
   - `ApplyFullTaste(PrinterOptions, bool enabled = true)` — folds `With(...)` over that subset. NativeAOT-safe, reflection-free.
2. **CLI config (`RenderStyleConfig`)** — a tool-owned aggregate key `dotnet_inspect_style_full_taste = true` that invokes `ApplyFullTaste`. Order-preserving: a later explicit per-knob line overrides (last write wins), consistent with the existing per-key semantics.

## Composition with Applied Taste

[...]

PR #3249, opened 2026-07-26
Gate byte-neutral style knobs with a compile-back IL-identity check (#3247)

### LINKED ISSUE #3247: Byte-neutral style knobs declare byte-identity that no gate checks (contract also strengthened past #3169's ask)
## Summary

The style-option catalog classifies each knob as `ByteDivergent` true/false, and that
classification decides which fidelity contract the knob's output is held to. Eleven of the
thirteen `PrinterOptions` knobs are classified **byte-neutral** (`ByteDivergent = false`),
and their doc comments state the byte-identity claim directly — e.g.
`PreferVarWhenTypeApparent`: *"Byte-neutral (a spelling choice, no IL consequence)"*, and
#3226's commit message: *"the emitted form recompiles to identical IL."*

Nothing checks that claim. Two separate gaps compound:

1. **`ByteDivergent` is only tested for self-consistency.**
   `StyleOptionCatalogTests.cs:158` asserts `ByteDivergent == (Tier == Lens)` — the metadata
   agrees with the metadata. There is no test that renders with a knob on, recompiles, and
   diffs the IL against the knob-off render.

2. **The quality harness never sets an option.**
   `grep -c PrinterOptions tools/DecompilerHarness/*.cs` → **0**. The corpus sweep,
   compile-back fidelity, ReturnToSender, and validity check all run `PrinterOptions.Default`
   exclusively. The apparatus measures exactly one point (all-off) in a 2^13 space.

Per-knob unit tests (e.g. `VarWhenApparentTests`) do pin rendered text over compiled
specimens, which is good — but they assert *spelling*, not *IL equivalence*.

## The part that makes this more than a missing test

**The contract was strengthened during implementation and the stronger claim is the
unverified one.**

#3169 specified the `var` family as an opt-in lens with an explicitly weaker contract:

> Fidelity requirement is behavioral only (100%), not byte-identity — same contract as the
> other opt-in lenses.

What shipped classifies the same knobs as **not** lenses (`ByteDivergent = false`,
`StyleOptionCatalogTests.cs:383`), i.e. byte-identical. That is very likely the *correct*
call — `var` really is a compile-time inference — and it is a better outcome than the issue
asked for. But it moves the knobs out of the lens tier (which is excluded from byte gates
*by design*, so the weaker claim would have been self-consistent) and into a tier whose
defining property nothing exercises. The classification is load-bearing precisely because it
selects the gate, and neither tier currently receives one for opt-in output.

## Why the claim deserves a gate rather than trust

Byte-neutrality of a spelling is the kind of claim that is usually right and occasionally
spectacularly wrong, and this family has already produced one near-miss:

- `var` + the target-typed-`new` shortener collapse to `var x = new()` → **CS8754**. That was
  caught by a human reasoning ahead in a `DeclarationTypeText` comment *before* the lens
  existed, and honored by #3226 — not by any gate.
- #3226 explicitly defers the built-in and elsewhere buckets *"with their own type-safety
  gate across the coercion surface."* That gate has no issue. This is it.

The remaining knobs interact on the same declaration line (qualification × var × wrapping),
which is exactly where printer interaction bugs live.

## Ask

1. **A byte-neutrality gate.** For each `ByteDivergent = false` knob: render a fixture corpus
   with the knob on, compile back, and assert IL identity against the knob-off render.
   Extend to the `FullTaste` aggregate (#3176) so the endorsed subset is covered as a
   configuration, not just knob-by-knob.
2. **Wire it to the classification, not to a hand-maintained list** — drive the gate off
   `StyleOptionCatalog.Options.Where(o => !o.ByteDivergent)` so a new byte-neutral knob is
   gated automatically. This mirrors the existing drift guard
   (`EveryBackingBooleanProperty_IsReachableThroughSomeCatalogValue`), which is the right
   pattern and already in place for catalog coverage.
3. **Decide and record the contract for `ByteDivergent = true` lenses.** They are excluded
   from byte gates by design; state whether they get a behavioral gate (#3169 asks for 100%
   behavioral fidelity) or are explicitly best-effort. Today they get neither.

## Notes

- No live defect is claimed. `PrinterOptions.Default` is all-off, so the shipped default
  remains fully gated and byte-stable; this is about the opt-in surface.
- The catalog design is otherwise strong — externally-specified option space mapped to real
  `csharp_style_*` keys, the orthogonal declared (`OracleEndorsed`) vs revealed
  (`CorpusEndorsed`) axes, and a working reflective drift guard. This issue closes the one
  property the design declares but does not check.
- Drive-by: `src/ILInspector.Decompiler/Pipeline/Ir/PrinterOptions.cs:81-82` has a duplicate
  `<summary>` open tag on `QualifyMethodAccess` (malformed XML doc). Worth fixing alongside,
  and worth considering XML-doc validation for this file — the doc comments carry the
  editorconfig mapping, the byte-neutrality claim, and the gate consequence, so they are
  effectively the specification.

Related: #3169 (var / apparency row), #3138 (style-lens design), #3160 (catalog), #3176
(full-taste aggregate), #3226 (apparent-type bucket).

richlander:
## Summary

Closes the one property the style-option catalog **declares but never checks**: the byte-neutral (`ByteDivergent = false`) knobs claim their output recompiles to the same IL as the shipped default, but `StyleOptionCatalogTests` only checked the classification agrees with itself (`ByteDivergent == (Tier == Lens)`), and the quality harness never set an option (`grep -c PrinterOptions tools/DecompilerHarness/*.cs` → 0).

Adds a **byte-neutrality gate driven off the classification** (`StyleOptionCatalog.Options.Where(o => !o.ByteDivergent)`), so a new byte-neutral knob is covered automatically rather than via a hand-maintained list.

## What it checks

| Tier | Byte-neutral? | How it's gated |
| --- | --- | --- |
| **Spelling** (`this.` qualification, `var`) | yes | **Compiled back**: a firing specimen rendered knob-on vs knob-off must produce the same recompiled IL (opcode stream + compile-back contract-V1 verdict). The `var x = new()` → CS8754 near-miss lived in this tier — its neutrality is a *semantic* (token-binding) claim, so it is proven, not asserted. |
| **Formatting** (whitespace), **Synthesis** (local names) | yes | **Structural**: layout is absent from the assembly and a local's name lives in the PDB, never the method body, so neutrality is recorded/pinned by tier. |
| **Lens** (`ByteDivergent = true`) | no | **Excluded by design** and held to a behavioral (100%) gate instead — the existing `PreferConditionalReturnLensTests` / `PreferBranchlessBooleanLensTests` executed-equivalence tests. A test records that contract (ask #3). |

Drift guard: a new **Spelling** byte-neutral knob with no firing specimen fails `EverySpellingByteNeutralKnob_HasAFiringSpecimen`, keeping the gate exhaustive over the tier whose neutrality is non-structural.

## Implementation notes

- Reuses existing firing specimens (`ThisQualificationSpecimen`, `VarWhenApparentSpecimen`) and the harness compile-back machinery rather than reconstructing either (harness-boundary rule).
- `FidelityCheck.EvaluateTargets` gains an optional `PrinterOptions` overload that threads the option into the **raised** renderer. The default (`options = null`) path is **byte-for-byte unchanged** — every existing corpus/gate call still renders the shipped output.
- The compile-back proof is batched to **three passes** over the large test assembly and marked `Speed=Slow` (matching how every other compile-back gate in the repo is classified); the structural / drift / token-rewrite checks stay fast and run in the merge `test` job.
- Turning a type's byte-neutral knobs on together also exercises the on-the-same-line interaction the catalog flags (qualification × var); each method's site is rewritten by exactly one knob, so the per-method comparison still isolates that knob's neutrality.
- **Drive-by**: fixes a malformed duplicate `<summary>` open tag on `PrinterOptions.QualifyMethodAccess`.

## Evidence

- Full solution builds clean (`dotnet build dotnet-inspect.slnx -c Release`).
- `ByteNeutralityGateTests`: 5/5 green (fast subset 1.5s; slow compile-back 77s).
- No regression in the harness `EvaluateTargets` refactor or the drive-by: `StyleOptionCatalogTests`, `ThisQualificationTests`, `VarWhenApparentTests`, `NullCoalescingAssignmentPassTests` (exercises the refactored targeted compile-back) — 111/111 green.

## Non-goals / follow-ups

- The `FullTaste` aggregate (#3176) as a byte-neutral *configuration*: the endorsed subset includes a byte-**divergent** lens (the ternary), so a whole-aggregate byte gate isn't meaningful; the byte-neutral endorsed subset (the four qualify knobs) is already exercised together here.
- XML-doc validation for `PrinterOptions.cs` (the issue notes it as "worth considering") — separate infra change.

Closes #3247.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ByteNeutralityGateTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ByteNeutralityGateTests.cs
@@ -30,16 +30,26 @@ namespace ILInspector.Decompiler.Tests;
 /// bind to the same IL. That is a semantic claim — the <c>var x = new()</c> → CS8754
 /// near-miss lived here — so each emitting value is compiled back: rendering a firing
 /// specimen with the value on and off must both recompile to the original IL. A new
-/// Spelling byte-neutral value with no specimen fails
-/// <see cref="EverySpellingByteNeutralValue_HasASpecimen"/>.
+/// byte-neutral value with no specimen fails
+/// <see cref="EveryByteNeutralValue_HasASpecimen"/>.
 /// </description></item>
 /// <item><description>
-/// The <see cref="StyleOptionTier.Formatting"/> (whitespace-only) and
-/// <see cref="StyleOptionTier.Synthesis"/> (local-name-only) byte-neutral knobs change
-/// neither the emitted tokens nor the metadata the IL is decoded from — layout is not
-/// in the assembly at all, and a local's name lives in the PDB, never the method body.
-/// Their byte-neutrality is structural, recorded and pinned by tier rather than
-/// re-proven per knob.
+/// The <see cref="StyleOptionTier.Formatting"/> (whitespace-only) byte-neutral knobs
+/// change layout the assembly never stored. That is still checkable without a compiler:
+/// rendering a firing specimen with the knob on and off must differ (the knob fired) yet
+/// collapse to the same text once insignificant whitespace is removed — a fast lane the
+/// reviewer valued. A knob mis-classified as Formatting that actually moved a token would
+/// fail the whitespace-equality assertion. See <see cref="FormattingValue_On_ChangesOnlyWhitespace"/>.
+/// </description></item>
+/// <item><description>
+/// The one <see cref="StyleOptionTier.Synthesis"/> byte-neutral knob (readable local
+/// names) renames a local, and a local's name lives in the PDB, never the method body,
+/// so it is byte-neutral by construction. It also cannot fire through the member-body
+/// entrypoint when a source name is present — the test assembly's embedded PDB supplies
+/// one — so the gate pins its inert render (on == off) on a source-named local. Firing
+/// the synthesis (proving IL-identity by compile-back, since names are absent from the
+/// IL) needs a name-less compiler temporary this corpus cannot author deterministically;
+/// that is tracked as follow-up.
 /// </description></item>
 /// </list>
 ///
@@ -59,21 +69,28 @@ public sealed class ByteNeutralityGateTests
     static string AssemblyPath => typeof(ByteNeutralityGateTests).Assembly.Location;
 
     /// <summary>
-    /// One specimen per non-default value of a Spelling-tier byte-neutral knob: the
-    /// catalog value token to turn on, the declaring type and method whose decompiled
-    /// body that token governs, and the harness signature that identifies the method
-    /// for compile-back. A knob's value domain — not the knob — is the coverage unit,
-    /// so a multi-value axis (e.g. the three <c>var</c> categories) needs one specimen
-    /// per category, not one per axis.
+    /// One specimen per non-default value of a byte-neutral knob: the catalog value
+    /// token to turn on, the declaring type and method whose decompiled body that token
+    /// governs, and how that value's byte-neutrality is proven. A knob's value domain —
+    /// not the knob — is the coverage unit, so a multi-value axis (e.g. the three
+    /// <c>var</c> categories) needs one specimen per category, not one per axis.
     ///
     /// <para>
-    /// <see cref="Emits"/> distinguishes a value the printer actually consumes today
-    /// (the token is rewritten, so the value is compiled back to prove IL-identity)
-    /// from a value that is declared in the catalog but not yet wired into emission
-    /// (a no-op that renders identically to the default). The unwired values are
-    /// pinned as no-ops on an input they <em>would</em> govern once wired, so the day
-    /// emission lands the pin flips (on-render diverges from off) and forces the
-    /// author to promote the value to an emitting, compiled-back specimen.
+    /// <see cref="Proof"/> selects the check the value's neutrality is held to and must
+    /// agree with the knob's tier (asserted by <see cref="EveryByteNeutralValue_HasASpecimen"/>):
+    /// a Spelling or Synthesis value rewrites tokens or metadata the compiler binds, so
+    /// it is <see cref="NeutralityProof.CompileBack"/>; a Formatting value only moves
+    /// whitespace, so it is <see cref="NeutralityProof.Whitespace"/>.
+    /// </para>
+    ///
+    /// <para>
+    /// <see cref="Emits"/> distinguishes a value the printer actually consumes today (the
+    /// render changes, so the value is exercised) from one that renders identically to
+    /// the default — either a catalog value not yet wired into emission (the deferred var
+    /// buckets) or a knob that is inert on this corpus (readable-local-names, whose
+    /// synthesis a present PDB source name suppresses). The inert values are pinned as
+    /// no-ops on an input they <em>would</em> govern once active, so the day emission
+    /// lands (or a name-less local appears) the pin flips and forces an emitting specimen.
     /// </para>
     /// </summary>
     sealed record ValueSpecimen(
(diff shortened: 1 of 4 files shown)


richlander:
Reviewed at head `28eed7e2` in an isolated worktree. **Approve in substance** — this closes the hole #3247 described, and the design is right. One change I'd want before merge, because the gap sits exactly where the known-risky deferred work will land.

## What's right

`SpellingKnob_On_RewritesTheSpecimenTokens` is the most important test in the PR. A byte-identity gate whose two renders happen to be equal passes vacuously and proves nothing — that failure mode is precisely what made the original `CheckInvariant` situation worthless. Pinning non-vacuity separately, and keeping it in the fast lane, is the right construction.

Also verified:

- The renderer seam is genuinely additive. `Renderer(source, lowered)` → `Renderer(source, lowered, null)` and `EvaluateTargets(a, t, lowered = false)` → `(a, t, lowered, null)`, so every existing corpus/gate call renders the shipped output. Default path unchanged.
- No new layering. Many test files already reference `ILInspector.DecompilerHarness`; this is not a new dependency direction.
- Gate is driven off `!o.ByteDivergent` rather than a hand list, as asked.
- `ByteNeutralityGateTests` 5/5 green on my own Release build (86s, slow test included).

## Blocking-ish: the drift guard is value-blind on multi-value axes

```csharp
var required = ByteNeutralKnobs.Where(o => o.Tier == Spelling).Select(o => o.Id).ToHashSet();
var covered  = Specimens.Select(s => s.KnobId).ToHashSet();
Assert.Equal(required, covered);
```

This keys on knob **Id**. But `var-spelling-style` is one descriptor carrying **three independent non-default values**:

| Token | Backing property | Specimen? |
| --- | --- | --- |
| `var-for-built-in-types` | `PreferVarForBuiltInTypes` | **no** |
| `var-when-type-apparent` | `PreferVarWhenTypeApparent` | yes |
| `var-elsewhere` | `PreferVarElsewhere` | **no** |

So when #3226's explicitly deferred buckets light up in the printer, they become byte-neutral Spelling states with **no compile-back proof**, and this guard stays green — the Id already has a specimen. The guard's exhaustiveness claim is over descriptors, but the risk lives at value granularity.

That matters more than it sounds because of *which* values they are. #3226 deferred them "with their own type-safety gate across the coercion surface," i.e. the author already identified them as the risky ones. `var-elsewhere` in particular is the bucket where declared type and initializer type can legitimately differ — the textbook case being `long x = 1;`, where `var x = 1;` infers `int`. `PreferVarElsewhere`'s doc says it only fires when "the initializer's type is exactly the declared type," so the printer is *supposed* to guard it. Nothing proves the guard, and after this PR nothing will force a specimen that would.

Fix: key the guard on `(Id, ValueToken)` across each descriptor's non-default values rather than on `Id`. Small change, and #3169's ask #3 plans *more* multi-value axes, so the gap widens over time rather than staying static.

## Two smaller notes

**1. Formatting/Synthesis neutrality is argued, not checked — and it's cheap to check.**

The structural rationale is sound (layout isn't in the assembly; a local's name lives in the PDB). But it's asserted by tier, which means the *tier assignment itself* is the unchecked link: a knob that actually rewrites tokens but is classified `Formatting` passes `ByteNeutralKnobs_AreOnlyFormattingSpellingOrSynthesis`, has no specimen requirement, and never compiles back. That's the one seam where "declared but not enforced" survives this PR.

Closing it needs no specimen curation and no compile-back: for each `Formatting` knob, render on vs off and assert the **whitespace-normalized** text is equal. That is exactly the claim ("whitespace-only") stated as a checkable property, it's fast enough for the merge lane, and it fails loudly on a mis-tiered knob. A structural analogue works for `Synthesis` (identical modulo local identifiers).

**2. `Assert.Equal(offResult.Status, onResult.Status)` passes if both renders are equally bad.**

If the knob-off baseline lands on some non-`Exact` status and knob-on matches it, the gate is green. That's defensible — byte-neutrality is a *relative* claim — but it means the gate silently weakens if a specimen's baseline degrades. Asserting the baseline reaches the strongest status the specimen is expected to achieve would keep the comparison anchored.

## On the slow-lane placement

`SpellingKnob_On_RecompilesToIdenticalIl` is `Speed=Slow`, so the actual IL-identity proof doesn't run in the merge `test` job. That matches repo convention and the other four tests do run there, so I'm not asking for a change — but it's the same structural pattern flagged in the review (`CorpusSweepGateTests` and `FidelityGateTests` are both slow-lane too): the checks that catch regressions live in the lane that isn't run per-PR. Worth tracking as a standalone question about gate placement rather than relitigating it here.

## Verification performed

| Check | Result |
| --- | --- |
| `dotnet build dotnet-inspect.slnx -c Release` @ `28eed7e2` | 0 errors |
| `ByteNeutralityGateTests` (all 5, incl. `Speed=Slow`) | 5/5, 86.5s |
| Default-path preservation in `FidelityCheck` overload chain | confirmed — both new overloads terminate in `options: null` |
| Test → harness layering | pre-existing; not introduced here |
| Drift-guard granularity vs `StyleOptionCatalog` var axis | **gap confirmed** — 3 non-default values, 1 specimen, guard keys on Id |

richlander:

Two independent reviewers (GPT-5.6 and Gemini-3.1 Pro), each isolated in a separate review worktree pinned to the exact head, `/tmp`-only scratch, no git mutation.

### Round 1 (base head `28eed7e2`) — both BLOCK, converging findings

**Finding A — soundness of the off-vs-on comparison (both reviewers, blocker).** The gate compared recompiled opcode *names* plus a render-vs-*original* status. Two renders that both diverge from the original in the same category (e.g. both `OperandDiff`) with equal opcode names could pass while their operands differ — so it did not prove the two renders are IL-identical to each other. Both reviewers suggested "require `Exact`".

Dismissed the literal "require `Exact`" fix and verified why: the `qualify-event-access` (`Subscribe`) baseline recompiles to a benign `OpcodeDiff` (event-subscription decompilation over-renders, independent of the `this.` qualifier — GPT captured it: original `ldarg ldarg call ret` vs recompiled `ldarg ldarg ldfld ldarg call castclass stfld ret`). Requiring `Exact` would fail a legitimately byte-neutral knob.

Resolution (`2bd2759c`): `SpellingValue_On_RecompilesToTheSameIlAsOff` now compares the two renders' **full contract-V1 structured diffs against the shared original** (`Outcome` + `Rows.SequenceEqual`, plus `RecompiledOpcodes` and an `IsAvailable` guard). Identical diffs vs the same original ⟹ the two recompiled bodies deviate identically ⟹ they are IL-identical to each other, using the same contract V1 the product's own fidelity claims use — robust to a benign over-render baseline.

**Finding B — per-value coverage (GPT, major).** `var-spelling-style` is a three-value axis, but the drift guard keyed on knob id, so covering one value (`var-when-type-apparent`) satisfied it. Resolution (`2bd2759c`): the drift guard now keys on `(KnobId, ValueToken)` derived live from the catalog, and the two declared-but-unwired var buckets (`var-for-built-in-types`, `var-elsewhere`) are added as `Emits:false` no-op pins that flip (and force an emitting specimen) the day emission lands.

### Round 2 (fixed head `2bd2759c`) — both CLEAR

- **GPT-5.6: CLEAR.** Ran an operand-perturbation probe → `rows-equal=False`, directly confirming the row comparison catches divergence; verified the `IsAvailable` guard, the non-Exact event baseline, both pins, and the drift guard. Gate 5/5 (77s); pre-existing compile-back gate green.
- **Gemini-3.1 Pro: CLEAR.** Confirmed `IlDiffRow`/`CanonicalIlOperation`/`IlOperandIdentity` are records (element-wise `SequenceEqual`); the event specimen yields **14** non-empty diff rows equal across off/on (real work); the drift guard fails on a missing specimen. Gate 5/5; no regression.

Both reviewers re-verified the default harness path (`options == null`) is byte-for-byte unchanged. CI is green on `2bd2759c` (merge `test` job passed; it excludes the `Speed=Slow` compile-back, which runs in the slow gate like every other compile-back gate).

richlander:
Ready to merge

richlander:
## Adversarial re-review reconciliation (follow-up commits)

Following the non-blocking feedback above, I addressed the two open notes and ran a fresh adversarial round (two non-Claude models) over the new test logic. Summary of what changed and how the findings were resolved.

### Disposition of the earlier non-blocking feedback
- **Drift guard value-blind on multi-value axes** — already fixed in the first adversarial round (`2bd2759c`): the guard keys on `(KnobId, ValueToken)` and covers every non-default value; the two unwired `var` buckets are `Emits:false` no-op pins.
- **`Status`-equal could pass if both renders are equally bad** — addressed more strongly than "anchor the baseline": the off-vs-on proof compares the two recompiled bodies to **each other** via identical contract-V1 `FidelityDiff` against the shared original (`2bd2759c`), and this follow-up **also** adds the requested baseline anchor: `CompileBackValue_On_RecompilesToTheSameIlAsOff` now asserts the knob-off render reaches a per-specimen `ExpectedBaseline` (`Exact`, or the event specimen's benign `OpcodeDiff`), so a silent "both regressed" degradation fails loudly.
- **Formatting/Synthesis neutrality argued by tier, not checked** — addressed; see below.
- **Slow-lane placement** — tracked separately as noted; non-actionable here.

### Formatting/Synthesis neutrality — final design
My first attempt added a fast text check (`FormattingValue_On_ChangesOnlyWhitespace`) that normalized away "insignificant" whitespace. Both reviewers BLOCKed it, twice, and correctly:
1. The normalizer's `\s+`→"" strip removed the very word-word spaces it meant to preserve, so `int alpha`/`intalpha` (and `+ +`/`++`) token fusions normalized equal — a false CLEAR. Fixed with a sentinel (`5aeb8cbc`).
2. The operator character class still omitted `.`, so `1..2` vs `1 . . 2` (and `?.`, `$@`) fused undetected. Both recommended comparing lexer token sequences — but Roslyn's lexer **splits** `..` into two dot tokens, so `1..2` and `1 . . 2` tokenize identically even though only the first compiles. **Insignificant whitespace in C# cannot be recognized soundly from text or a token stream.**

Resolution (`e935f602`): removed the whitespace/text heuristic entirely and hold Formatting knobs to the **same compile-back IL-identity proof** the Spelling knobs already use (and that both reviewers cleared). Layout never reaches the IL, so a Formatting knob's on/off renders recompile identically; a knob mis-tiered as Formatting that actually changed the IL fails the off-vs-on contract-V1 diff. This proves neutrality at the IL level, uniformly across all byte-neutral tiers, and ends the operator-class whack-a-mole. `ValueTokenState_MatchesEmits` (fast) still proves each knob fires (off != on) so the compile-back is non-vacuous.

**Synthesis** (`readable-local-names`): inert through the member-body entrypoint when a source name is present (the test assembly's embedded PDB names the local), and a local name lives in the PDB, never the IL — pinned as an `Emits:false` no-op. Firing it (via a name-less compiler temporary) is tracked as follow-up.

### Adversarial verdicts (fixed head `e935f602`)
- **GPT-5.6**: CLEAR — verified head, build clean, gate `Total: 5, Failed: 0`, Formatting targets resolve and reach `Exact`, compile-back uses the knob-on options, non-vacuity holds.
- **Gemini-3.1 Pro**: CLEAR — verified head, gate 5/5 (~96s), the three Formatting specimens are genuinely exercised (signatures resolve, reach `Exact`), `LongLogicalChain` is non-folding, no residual masking.

Both reviewers ran in isolated review worktrees at the exact head with scratch under `/tmp/` and no writes to tracked files. I will re-affirm readiness once the `test` job is green on this head.

richlander:
Ready to merge

Re-affirmed at head `e935f602`: the `test` job is green (8m23s), build/smoke checks pass (`il-diff-smoke`/`markdownlint`/`pack` are path-gated and skipped), and both fixed-head adversarial reviews (GPT-5.6, Gemini-3.1 Pro) are CLEAR. Review worktrees removed.

richlander:
Confirming my earlier review, which was "approve with one change first" at `28eed7e2`. Re-reviewed at the fixed head `e935f602` in a clean worktree. **Approved — the condition is met.**

My finding was that `EverySpellingByteNeutralKnob_HasAFiringSpecimen` compared knob **Id** sets, so `var-spelling-style` — one descriptor with three independent non-default values — counted as covered when only one had a specimen, and #3226's deferred buckets would land uncovered with the guard green.

`2bd2759c` closes it, and does more than I asked:

- `EveryByteNeutralValue_HasASpecimen` keys on `(KnobId, ValueToken)` and asserts **set equality**, not containment (`ByteNeutralityGateTests.cs:217-219`). So a stale specimen for a removed value fails too, not just a missing one.
- The required set is derived from the catalog rather than hand-listed — `ByteNeutralNonDefaultValues` walks `o.Values.Where(v => v.Token != o.DefaultValue)`, so a value added to any axis appears in the requirement automatically.
- `ByteNeutralKnobs` is `StyleOptionCatalog.Options.Where(o => !o.ByteDivergent)` (`:162-163`). That's the part that actually resolves #3247: the **declaration is now the driver of the enforcement set**. You can no longer mark a knob byte-neutral without the gate demanding a specimen for every one of its values. That's the inversion the issue was asking for.

I checked the `Emits:false` pins specifically, since a "declared inert" escape hatch is exactly how this class of guard usually rots. It doesn't here — `ValueTokenState_MatchesEmits` (`:249-254`) asserts an inert value renders *identically* off vs on, so the moment someone wires one of the deferred `var` buckets the pin breaks with an actionable message ("the value is now active — add a firing specimen and set `Emits:true` so it is proven"). The pin cannot silently absorb a newly-wired knob.

On the Formatting resolution: deleting the whitespace normalizer and holding Formatting knobs to the same compile-back IL-identity proof as Spelling was the right call. The reviewers' finding that `1..2` and `1 . . 2` tokenize identically is a genuinely good catch — it means insignificant whitespace isn't recoverable from text *or* a token stream, so any text-level neutrality check was going to be unsound no matter how many characters got added to the operator class. Converging every byte-neutral tier onto one IL-level proof removes the special case instead of patching it, and `ValueTokenState_MatchesEmits` keeps that proof non-vacuous by pinning off != on separately.

**Verified at `e935f602`:** full `dotnet-inspect.slnx -c Release` build clean; `ByteNeutralityGateTests` **5/5 pass** (75s).

No remaining findings from me. The slow-lane placement note stays non-blocking follow-up as agreed.

Issue #3241, opened 2026-07-26
Make the IR invariant check run in Release, and give it teeth


richlander:
Part of #3240.

## Problem

`IrNode.CheckInvariant()` (`src/ILInspector.Decompiler/Pipeline/Ir/IrNode.cs:180`)
is the IR's structural invariant check. Its doc comment states: *"Debug builds
run this after every pass; a violation is a pipeline bug, never input data."*

Two things are wrong with that in practice.

### 1. It does not run in the configuration we run

It is `[Conditional("DEBUG")]`. `AGENTS.md` prescribes Release for tests:

```
dotnet run --project src/ILInspector.Decompiler.Tests -c Release
```

`[Conditional]` strips the **call site** when the calling assembly lacks the
symbol, so building the test project in Release removes every call. Measured on
the emitted metadata of the built test assembly:

| Configuration | MemberRefs to `CheckInvariant` |
| --- | --- |
| `-c Release` (prescribed) | **0** |
| `-c Debug` (positive control) | 1 |

There are **617 `CheckInvariant()` call sites across 121 test files**, plus 5 in
the product. In the prescribed run, all of them are compiled away. The suite
reads as if it asserts IR integrity 617 times; it asserts it zero times.

Reproduce:

```bash
dotnet build src/ILInspector.Decompiler.Tests -c Release
dotnet build src/ILInspector.Decompiler.Tests -c Debug
# then count MemberReference rows named "CheckInvariant" in
# artifacts/obj/ILInspector.Decompiler.Tests/{release,debug}/ILInspector.Decompiler.Tests.dll
```

### 2. It checks very little

The body validates only that each child's `Parent` is the node and that
`ChildIndex` matches its slot. It would not have caught #3236.

## Proposed change

1. **Make it executable in Release.** Replace the `[Conditional("DEBUG")]`
   compile-time switch with a runtime opt-in (env var or harness flag) so the
   corpus sweeps can enable it across CoreLib and the 104-assembly pool. A
   check that cannot run where we run is equivalent to no check.
2. **Give it teeth.** Beyond parent/child wiring: node arity, child-kind
   legality, `ResultType` consistency at sinks, slot/local index validity, and
   branch-target existence.
3. **Consider `IIrPass.Requires`/`Invalidates`** so each pass declares the
   properties it depends on and the ones it may break, and the invariant runner
   checks the declared set between passes.

Do **not** address this by adding `Debug.Assert`. Tests run Release, so an
[...]

PR #3250, opened 2026-07-26
Make the IR invariant check run in Release (#3241, PR 1 of stack)

richlander:
Closes the first, foundational part of #3241 (part of the #3240 cluster). **Base of a stack** — teeth and the corpus gate follow on this branch.

## Problem

`IrNode.CheckInvariant` was `[Conditional("DEBUG")]`. AGENTS.md prescribes `-c Release` for the test suite, and `[Conditional]` strips the *call site* in a caller that lacks the symbol — so in the Release test assembly all 600+ call sites (120 test files + 5 product hooks) were stripped. The check that reads as if it guarantees IR tree integrity **asserted zero times**.

Debug can't be the fix: the suite fails ~171 tests under Debug for unrelated Debug-IL-shape reasons, and Release is required for **input fidelity** — the optimized IL a Release compiler emits is what ships and what the decompiler corpus consumes. A runtime opt-in is the only path that runs the check in the build users actually run.

## This PR (make it run + prove it has teeth)

No product codegen change on the shipped hot path.

- **Remove `[Conditional("DEBUG")]`** from `IrNode.CheckInvariant` so explicit calls always run, in any configuration.
- **Add `IrInvariants.Enabled`** (env var `DOTNET_INSPECT_IR_INVARIANTS`, default **off**) gating the 5 product per-pass hooks (`IrPass` runner ×3, `IrImporter` ×2). The shipped CLI pays nothing per pass; opt-in hosts validate after every pass.
- **Enable it suite-wide** in the decompiler test host via `[ModuleInitializer]`.
- **Negative-control tests** corrupt a child's `Parent`/`ChildIndex` backing field via reflection and assert `CheckInvariant` throws — proving the check is not a silent no-op (its whole failure mode here).
- **AGENTS.md**: document why the suite runs Release (input fidelity, not speed) and why correctness checks must not hide behind `[Conditional("DEBUG")]`.

The check body is intentionally still the existing weak parent/child-wiring check in this PR; expanding it (arity, `ResultType` at sinks, branch-target existence — each corpus-validated) and the CoreLib+pool corpus-sweep gate are the follow-up PRs stacked on this branch.

## Validation (Linux, Release)

- Decompiler fast suite: **3812 real tests pass** with per-pass invariant checks live suite-wide — **zero invariant violations**. The lone failure `LookalikeExpressionAssembly_StaysFactoryCalls` is a missing fixture from building only the test csproj (`FileNotFoundException` on `System.Linq.Expressions.dll`), identical on base and green in a full `slnx` build / CI.
- Negative-control tests: **4/4 pass** (well-formed tree passes; wrong parent, wrong slot, and deep-child corruption each throw).

Part of #3240.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -190,6 +190,17 @@ Tests use xUnit executable projects. **Use `dotnet run`, not `dotnet test`**;
 | Shared services | `dotnet run --project src/DotnetInspector.Services.Tests -c Release` |
 | Metadata | `dotnet run --project tests/ILInspector.Metadata.Tests -c Release` |
 
+Run the suite in **Release** for input fidelity, not speed: the optimized IL a
+Release build of the compilers emits is what ships and what the decompiler
+corpus consumes, so a Debug run would validate the decompiler against IL shapes
+users never see. Because the suite runs Release, correctness checks must not
+hide behind `[Conditional("DEBUG")]` — such a call is stripped from the Release
+test assembly and asserts nothing. The IR structural invariant check
+(`IrNode.CheckInvariant`) is instead a runtime opt-in (`IrInvariants.Enabled`,
+env var `DOTNET_INSPECT_IR_INVARIANTS`) that the decompiler test host turns on
+suite-wide, so the pipeline is validated after every pass in the same build
+users run, while the shipped tool pays nothing on the decompile hot path.
+
 Some CLI tests require `ilasm`/`ildasm` and skip when those tools are absent.
 The IL round-trip project has separate dependency restore and fast/full test
 commands; follow `tests/DotnetInspector.ILRoundtrip.Tests/README.md`.
--- a/src/ILInspector.Decompiler.Tests/IrInvariantCheckTests.cs
+++ b/src/ILInspector.Decompiler.Tests/IrInvariantCheckTests.cs
@@ -1,0 +1,71 @@
+using System.Reflection;
+using ILInspector.Decompiler.Pipeline;
+
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// Negative controls proving <see cref="IrNode.CheckInvariant"/> actually
+/// detects a corrupt parent/child link. Without these the check could silently
+/// become a no-op (its whole failure mode under #3241) and every other test
+/// would still pass. Corruption is forced through the private backing fields the
+/// normal API never lets a caller reach.
+/// </summary>
+public sealed class IrInvariantCheckTests
+{
+    static readonly FieldInfo ParentField = typeof(IrNode)
+        .GetField("<Parent>k__BackingField", BindingFlags.Instance | BindingFlags.NonPublic)!;
+
+    static readonly FieldInfo ChildIndexField = typeof(IrNode)
+        .GetField("<ChildIndex>k__BackingField", BindingFlags.Instance | BindingFlags.NonPublic)!;
+
+    static Block ParentWithTwoChildren(out Block first, out Block second)
+    {
+        first = new Block(0x10);
+        second = new Block(0x20);
+        var parent = new Block();
+        parent.Add(first);
+        parent.Add(second);
+        return parent;
+    }
+
+    [Fact]
+    public void CheckInvariant_PassesForWellFormedTree()
+    {
+        var parent = ParentWithTwoChildren(out _, out _);
+
+        parent.CheckInvariant();
+    }
+
+    [Fact]
+    public void CheckInvariant_ThrowsWhenChildParentPointerIsWrong()
+    {
+        var parent = ParentWithTwoChildren(out _, out var second);
+        ParentField.SetValue(second, null);
+
+        var ex = Assert.Throws<InvalidOperationException>(parent.CheckInvariant);
+        Assert.Contains("wrong parent", ex.Message);
+    }
+
+    [Fact]
+    public void CheckInvariant_ThrowsWhenChildSlotIsWrong()
+    {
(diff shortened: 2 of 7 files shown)


richlander:
## Adversarial review reconciliation — two models, both CLEAN at fixed head `a8e3b1c4`

Per AGENTS.md this is a behavior change (removes `[Conditional]`, enables the check suite-wide), so it got two-model review in isolated `/tmp` worktrees.

### Round 1 (head `43a3ea6c`)
- **Gemini 3.1 Pro — CLEAN.** Independently confirmed: exactly 5 product `CheckInvariant` call sites, all gated by `IrInvariants.Enabled`; no rogue unconditional calls. No static-init/module-init race (module initializer runs before xUnit parallelizes; no test mutates the flag). Ran the fast suite → 3812 pass, **zero `Invariant violated`**. Also noted a bonus safeguard: the negative controls pass `CheckInvariant` as a **method group** to `Assert.Throws`, which fails to compile (CS1618) if anyone re-adds `[Conditional]` — a compile-time regression guard.
- **GPT-5.6 — NOT CLEAN (1 Medium).** Valid coverage gap: every negative control called `CheckInvariant()` directly, so none proved the `[ModuleInitializer]` enables the flag or that `IrPasses.Run` honors the per-pass gate. Removing the initializer or deleting the gate would have left the suite green.

### Fix (`a8e3b1c4`, test-only)
Added `PipelineRunner_UnderTestHost_ThrowsWhenAPassCorruptsTheTree`: it does **not** set `IrInvariants.Enabled` itself; it relies on the test-host module initializer and runs a corrupting `IIrPass` through `IrPasses.Run`, asserting the runner throws. This is the only test that fails if the module initializer is removed, the per-pass gate is deleted, or the runner stops calling `CheckInvariant`.

### Round 2 re-review (fixed head `a8e3b1c4`)
- **GPT-5.6 — CLEAN.** Confirmed the finding is closed: with the env var unset, removing the module initializer fails the upfront assertion, and removing the per-pass hook makes `Assert.Throws` fail; the exception genuinely flows through `IrPasses.Run` → gated `CheckInvariant`. Verified diff `43a3ea6c..a8e3b1c4` touches only the test file (no product change), Release build clean, `IrInvariantCheckTests` 5/5.

### CI
All required checks green on the exact head `a8e3b1c4` (`test`, `build-net10`, `csharp-diff-smoke`, `markdownlint`, `changes`). `il-diff-smoke`/`pack` are path-gated skips.

**Ready to merge.**

Follow-ups (non-blocking, stacked on this branch): PR 2 gives the check *teeth* (arity, child-kind, `ResultType` at sinks, branch-target existence — each corpus-validated); PR 3 adds the CoreLib + pool corpus-sweep gate; PR 4 (optional) `IIrPass.Requires`/`Invalidates`.

richlander:
Reviewed at `893d1a32` in a clean worktree (reviewing the stack together; gate-specific findings are on #3262). **Approve.**

This is the right fix for #3241, and it is the right fix for the right reason. Three things I want to call out because they're easy to under-value in review:

**The runtime opt-in is better than the `Checked`-symbol alternative.** A second build configuration would have created a fourth IL flavor to maintain and would still have left the shipped binary unchecked. `IrInvariants.Enabled` runs the check *in the same optimized assembly users run and corpus sweeps consume*, which is the property that actually matters here. The `.NET Checked` analogy in the doc comment is apt.

**`PipelineRunner_UnderTestHost_ThrowsWhenAPassCorruptsTheTree` is the test that makes this PR real**, and its doc comment says so explicitly:

> This is the only test that fails if the module initializer is removed, the per-pass gate is deleted, or the runner stops calling CheckInvariant — the direct-call tests above would all still pass.

That is exactly the right instinct. The original defect was *a check that existed and asserted nothing*, so a fix consisting of tests that call `CheckInvariant` directly would have reproduced the bug at one remove — green suite, dead hook. Naming the single test that fails when the wiring dies is the correct guard, and it's the same non-vacuity discipline #3249 landed. Verified: **11/11 pass** in Release.

No static-init ordering hazard: `Enabled` is a settable auto-property, so the `[ModuleInitializer]` assignment triggers the class's static init and then overwrites it. Order-independent.

**The AGENTS.md text is the durable part.** It records *why* Release is mandatory (input fidelity — Release-compiled fixtures are the IL that ships, and Debug fails ~171 tests for unrelated shape reasons) and generalizes the lesson beyond this one check. Without that paragraph the natural "fix" for the next person hitting a stripped `[Conditional("DEBUG")]` call is to switch the suite to Debug, which would quietly degrade every raising test's input. Worth having.

One thing to keep an eye on rather than change here: the check is now opt-in-by-default-off, so its value depends entirely on every host that *should* enable it actually doing so. Today that's the test host and the harness. If a future entry point (a new benchmark, a new sweep tool) forgets, it silently gets no validation — the failure mode is identical to the one being fixed, just relocated from the compiler to the caller. A single shared "validating host" helper, or an assertion in the harness that `Enabled` is set, would close it. Not blocking.

richlander:
Final pre-merge check on the stack: reviewed at tip `8e27d8be`, merged current `main` (`14781e8d`, which now carries #3231) into it — clean — and ran the full suite with the filter `ci.yml:301` actually uses: **3840 passed, 0 failed** under both `--gate fast` and `-trait- "Speed=Slow"`.

My approval here stands unchanged. Full details and the fault-injection evidence for the gate are on #3262: https://github.com/richlander/dotnet-inspect/pull/3262#issuecomment-5086050463

**Good to merge**, in stack order #3250 → #3261 → #3262.

PR #3258, opened 2026-07-26
Classify compile-back failures with a cause + paired fix (#3256)

### LINKED ISSUE #3256: Layer 3: classified cause + paired fix for compile-back failures (#3238)
## Motivation

PR #3252 (layers 1–2 of #3238) gives compile-back `RecompileFail` results a **reveal + caret** render: the emitted C# line, invisible runes surfaced (`‹ZWNJ›`), and a caret under the exact diagnostic span followed by the raw `CSxxxx: message`. That tells a reviewer *where* the failure is and *what the compiler said* — but they still have to reconstruct *why* the emitted identifier bound elsewhere, and *what the emitter should have produced instead*.

Layer 3 (item 3 of #3238) closes that gap: a **classified `cause`** line paired with a concrete **`fix`** line, for an **enumerated** set of known round-trip failure classes.

## Why a real `fix` is honest here (not advice)

This is a **round-trip**: the harness holds *both sides* — the ground-truth original (metadata symbol name / signature) **and** the emitted text it produced. So the `fix` is the **delta** between emitted and original, not a guess. When we have the exact original, `fix` renders a compact `emit A → B` diff. Only for genuinely unspeakable names (no legal C# source form exists) does `fix` degrade to a **policy** sentence, and it says so.

## Proposal

Extend the layer 1–2 render (`FidelityCheck.RenderAnnotatedFailure`) to append, on the same member-column gutter as the caret `//`:

```
    //  cause: <class> — one-sentence mechanism
    //  fix:   emit  A → B     (or a policy sentence for unspeakable names)
```

Enumerated cause set (each keyed to a recognized diagnostic code + a cheap structural predicate the harness already owns — **reuse, do not reimplement C# binding**):

| Cause | Trigger (example) | Fix shape |
| --- | --- | --- |
| `invisible-rune` (Cf stripping) | CS0539 / CS0246 + `!MetadataIdentifierRoundTrips` (U+200C etc.) | `emit  M‹ZWNJ› → M` |
| `keyword-escape` | CS1041 + emitted identifier is a keyword written bare | `emit  class → @class` |
| `namespace-shadow` | CS0104 / CS0426 / CS0538 + bare name binds to wrong namespace | `emit  Timer → System.Threading.Timer` |
| `signature-drift` | CS0738 / arity/return mismatch | `emit  long Add(…) → int Add(…)` |
| `unspeakable-name` | CS1519 + compiler-generated metadata name | policy: `no exact source form — emit a speakable alias or exclude` |

For any **unrecognized** `CSxxxx`, emit **no** layer-3 line — the reviewer gets exactly the layer 1–2 render, never a fabricated cause.

## Reuse anchors

- `ReturnToSenderTypePlanner.MetadataIdentifierRoundTrips` (`tools/DecompilerHarness/ReturnToSenderTypePlanner.cs:1579`) — Cf/format-rune detection.
- `ExternalInterfaceSpellingShadowedByClosure` / `NamespaceIsInScope` (~1617/1658) — shadow classification.
- `OperatorNames.FormatDisplayName`, `CSharpIdentifier.Sanitize`, `CSharpFormatter.EscapeNamespace` — the exact emit transforms whose *inverse* names the fix.

These predicates are currently `private static` in `ReturnToSenderTypePlanner`; layer 3 will need a small shared classifier (or selective `internal` exposure) so the render site can consult them without duplicating logic.

## Scope / boundary

Harness-reporting only (no product-output change), stacked on the PR #3252 branch. The classifier maps `(diagnostic code, structural predicate) → (cause sentence, fix)` and owns **only** the enumerated set; everything else falls through to layer 1–2.

## Acceptance

- Each enumerated cause renders a `cause:` + `fix:` pair on the member gutter, with an exact `emit A → B` where the original is known.
- Unspeakable names render a labeled policy `fix`, not a fabricated token.
- Unrecognized diagnostic codes render **no** layer-3 line.
- Tests pin every cause + the fallback; any new dynamic compilation site is registered in the governance manifest.

Follows #3238 · builds on #3252.

richlander:
Closes #3256 · builds on #3252 · layer 3 of #3238.

## What

Layers 1–2 (#3252) give a `RecompileFail` a **reveal + caret** under the exact diagnostic span. This adds **layer 3**: a classified `cause:` and a paired `fix:` on the same gutter, for an enumerated set of failure classes.

## Why the `fix` is honest, not advice

Compile-back is a **round-trip**, so the harness holds the emitted text *and* — for these causes — the correct form is the emitted token's **own delta**. So `fix` shows a concrete `emit A → B` rather than a guess.

Authentic render (generated, not hand-drawn):

```csharp
    void M() { Missing‹ZWNJ›Type x = null; }
    //         ^^^^^^^^^^^^^^^^^ CS0246: The type or namespace name 'MissingType' could not be found (are you missing a using directive or an assembly reference?)
    //  cause: invisible-rune — emitted name carries ‹ZWNJ› (Cf); binding strips it, so the name no longer matches.
    //  fix:   emit  Missing‹ZWNJ›Type → MissingType   (drop the format-category rune)
```

Note the raw `CS0246` message prints `'MissingType'` — the invisible U+200C is *in* that string but the terminal can't show it. The `cause`/`fix` lines are what actually reveal and resolve the failure.

## Scope — only what's derivable, no guessing

`ClassifyCause` classifies **only** causes whose fix comes from the emitted token itself — no locale-fragile diagnostic-message parsing:

| Cause | Fix |
| --- | --- |
| `invisible-rune` (Cf stripping) | `emit  class‹ZWNJ› → class` |
| `keyword-escape` | `emit  class → @class` |
| `unspeakable-name` | policy: `no exact source form — emit a speakable alias or exclude` |

`invisible-rune` is checked first, so a keyword-looking token carrying a format rune is attributed to the rune (the real reason binding diverged). Every other diagnostic **falls through** to the bare layer 1–2 render.

**Deferred, on purpose:** `namespace-shadow` and `signature-drift` (the other two causes enumerated in #3256) need the ground-truth original for the *specific failing token*, which isn't available at this render site. They are omitted rather than fabricated — a follow-up can add them once that context is plumbed.

## Tests

`ClassifyCause` is pure, so the new tests call it directly — each cause, the invisible-rune-over-keyword precedence, and the fall-through — introducing **no new dynamic compilation site** (governance manifest unchanged). All Fidelity-area tests pass (41).

### DIFF
--- a/src/ILInspector.Decompiler.Tests/AnnotatedCompileBackFailureTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotatedCompileBackFailureTests.cs
@@ -87,12 +87,16 @@ public class AnnotatedCompileBackFailureTests
 
     // ---- Concrete fixes, held to their claim (apply -> recompile -> clean) ----
 
-    [Fact]
-    public void InvisibleRuneFixIsVerifiedByRecompile()
+    [Theory]
+    // ordinary stem — dropping the rune is sufficient
+    [InlineData("class \u200CFoo { }\n")]
+    // keyword stem — dropping alone would expose the bare keyword 'class', which
+    // is *worse* than the original lex error, so the fix must also escape it
+    [InlineData("class \u200Cclass { }\n")]
+    public void InvisibleRuneFixIsVerifiedByRecompile(string source)
     {
         // A leading Cf is a valid identifier-part but not an identifier-start, so
         // it breaks *lexing* (CS1001). Dropping it is a real fix.
-        string source = "class " + Zwnj + "Foo { }\n";
         Diagnostic err = FirstError(source);
         Assert.Contains(err.Id, new[] { "CS1001", "CS1056" });
 
@@ -101,6 +105,13 @@ public class AnnotatedCompileBackFailureTests
         Assert.StartsWith("invisible-rune", cf!.Value.Cause);
         Assert.NotNull(cf.Value.From);
         Assert.NotNull(cf.Value.To);
+
+        // The proposal must never be a bare reserved keyword. IsValidIdentifier is
+        // a character-rules check and returns true for 'class', so it cannot serve
+        // as the keyword filter.
+        Assert.False(
+            SyntaxFacts.IsReservedKeyword(SyntaxFacts.GetKeywordKind(cf.Value.To!)),
+            $"fix proposed a bare keyword: '{cf.Value.To}'");
 
         string patched = source.Replace(cf.Value.From!, cf.Value.To!);
         Assert.DoesNotContain(Zwnj, patched);
@@ -145,18 +156,20 @@ public class AnnotatedCompileBackFailureTests
     // ---- The former harmful precedence case, now correct ----
 
     [Fact]
-    public void KeywordStemCarryingCfIsNeverConvertedIntoAKeyword()
+    public void LeadingCfOverAKeywordStemProposesTheEscapedForm()
     {
-        // 'class\u200C' is a *legal identifier* - the rune is what makes it legal.
-        // The classifier must never propose `class\u200C -> class`, which would
-        // manufacture a keyword error.
-        FidelityCheck.CauseFix? cf = FidelityCheck.ClassifyCause("class" + Zwnj, 0, 6, "CS0103");
+        // Stripping the rune from '\u200Cclass' leaves 'class'. SyntaxFacts
+        // .IsValidIdentifier("class") is true - it checks character rules only and
+        // knows nothing about keywords - so the proposal must be composed from both
+        // causes: drop the rune *and* escape the keyword it exposes.
+        string source = "class " + Zwnj + "class { }\n";
 
-        if (cf is { } value)
-        {
-            Assert.DoesNotContain("keyword-escape", value.Cause);
-            Assert.NotEqual("class", value.To);
-        }
+        FidelityCheck.CauseFix? cf = Classify(source, FirstError(source));
+
+        Assert.NotNull(cf);
+        Assert.Equal(Zwnj + "class", cf!.Value.From);
+        Assert.Equal("@class", cf.Value.To);
+        Assert.Contains("escape", cf.Value.Fix);
     }
 
     // ---- Fall-through: not every span is a classifiable identifier ----
--- a/src/ILInspector.Decompiler.Tests/AnnotatedCompileBackFailureTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotatedCompileBackFailureTests.cs
@@ -1,3 +1,5 @@
+using System.Collections.Immutable;
+
 using ILInspector.DecompilerHarness;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -5,28 +7,45 @@ using Microsoft.CodeAnalysis.CSharp;
 namespace ILInspector.Decompiler.Tests;
 
 /// <summary>
-/// Pins the annotated compile-back failure render (#3238): a RecompileFail's
-/// emitted C# is shown with a caret underline under the exact diagnostic span,
-/// and invisible/format runes on that line are revealed so the caret points at
-/// something the reader can see. The caret sits on a <c>//</c> comment so the
-/// block stays valid C# inside a code fence.
+/// Pins the annotated compile-back failure render (#3238) and its cause/fix
+/// classification (#3256). Layers 1-2: the emitted C# is shown with a caret under
+/// the exact diagnostic span, invisible runes revealed. Layer 3: a classified
+/// <c>cause</c> + paired <c>fix</c>. The concrete-fix causes are held to their
+/// claim - apply the fix, recompile, and the diagnostic must be gone - so a
+/// green suite verifies the <c>fix:</c> line, not merely its wording.
 /// </summary>
 [Trait("Area", "Fidelity")]
 public class AnnotatedCompileBackFailureTests
 {
-    // The "killer" shape: an identifier carrying a zero-width non-joiner
-    // (U+200C). Roslyn cannot bind `Missing<ZWNJ>Type`, so the CS0246 span
-    // covers the whole identifier — including the invisible rune.
     const char Zwnj = '\u200C';
 
-    static Diagnostic FirstError(string source)
+    // The one dynamic compilation site in this file (registered in the
+    // DynamicCompilationSite manifest); every test routes through it.
+    static ImmutableArray<Diagnostic> Compile(string source)
     {
         var comp = CSharpCompilation.Create("annot",
             [CSharpSyntaxTree.ParseText(source)],
             [MetadataReference.CreateFromFile(typeof(object).Assembly.Location)],
             new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary));
         using var ms = new MemoryStream();
-        return comp.Emit(ms).Diagnostics.First(d => d.Severity == DiagnosticSeverity.Error);
+        return comp.Emit(ms).Diagnostics;
+    }
+
+    static Diagnostic FirstError(string source)
+        => Compile(source).First(d => d.Severity == DiagnosticSeverity.Error);
+
+    static bool CompilesClean(string source)
+        => !Compile(source).Any(d => d.Severity == DiagnosticSeverity.Error);
+
+    // Classify a real diagnostic the way the render site does: widen to the
+    // diagnostic's failing line + span, then hand it to the classifier.
+    static FidelityCheck.CauseFix? Classify(string source, Diagnostic d)
+    {
+        var span = d.Location.GetLineSpan().Span;
+        string line = source.Replace("\r\n", "\n").Split('\n')[span.Start.Line];
+        int start = span.Start.Character;
+        int end = span.End.Line == span.Start.Line ? span.End.Character : line.Length;
+        return FidelityCheck.ClassifyCause(line, start, end, d.Id);
     }
 
     [Fact]
@@ -41,12 +60,10 @@ public class AnnotatedCompileBackFailureTests
         string render = FidelityCheck.RenderAnnotatedFailure(source, FirstError(source))!;
         string[] lines = render.Split('\n');
 
-        // The code line reveals the ZWNJ as a visible token.
+        // Layers 1-2: the code line reveals the ZWNJ; the caret is a // comment
+        // carrying the diagnostic, aligned under the revealed identifier.
         Assert.Contains("Missing\u2039ZWNJ\u203AType", lines[0]);
         Assert.DoesNotContain('\u200C', render);
-
-        // The caret line is a // comment carrying the diagnostic, with carets
-        // aligned under the revealed identifier (7 + 6 for <ZWNJ> + 4 = 17).
         Assert.StartsWith("    //", lines[1]);
         Assert.Contains(new string('^', 17), lines[1]);
         Assert.Contains("CS0246", lines[1]);
@@ -55,64 +72,11 @@ public class AnnotatedCompileBackFailureTests
         int revealedIdentColumn = lines[0].IndexOf("Missing", StringComparison.Ordinal);
         Assert.Equal(revealedIdentColumn, caretColumn);
 
-        // Layer 3 (#3256): a classified cause + paired fix follow on the gutter,
-        // deriving the fix from the emitted token itself (reveal → stripped).
-        Assert.StartsWith("    //  cause: invisible-rune", lines[2]);
-        Assert.StartsWith("    //  fix:", lines[3]);
-        Assert.Contains("Missing\u2039ZWNJ\u203AType \u2192 MissingType", lines[3]);
-    }
-
-    [Theory]
-    [InlineData("M\u200C", "invisible-rune", "M\u2039ZWNJ\u203A \u2192 M")]      // Cf stripping
-    [InlineData("\u200DName", "invisible-rune", "\u2039ZWJ\u203AName \u2192 Name")]
-    public void ClassifiesInvisibleRuneWithStrippedFix(string token, string cause, string fixDelta)
-    {
-        var result = FidelityCheck.ClassifyCause(token);
-        Assert.NotNull(result);
-        Assert.StartsWith(cause, result!.Value.Cause);
-        Assert.Contains(fixDelta, result.Value.Fix);
-    }
-
-    [Theory]
-    [InlineData("class")]
-    [InlineData("int")]
-    [InlineData("return")]
-    public void ClassifiesBareKeywordWithVerbatimFix(string keyword)
-    {
-        var result = FidelityCheck.ClassifyCause(keyword);
-        Assert.NotNull(result);
-        Assert.StartsWith("keyword-escape", result!.Value.Cause);
-        Assert.Equal($"emit  {keyword} \u2192 @{keyword}", result.Value.Fix);
-    }
-
-    [Theory]
-    [InlineData("<>c")]
-    [InlineData("<M>b__0")]
-    public void ClassifiesUnspeakableNameWithPolicyFix(string token)
-    {
-        var result = FidelityCheck.ClassifyCause(token);
-        Assert.NotNull(result);
-        Assert.StartsWith("unspeakable-name", result!.Value.Cause);
-        Assert.Contains("no exact source form", result.Value.Fix);
-    }
-
-    [Theory]
-    [InlineData("Ordinary")]     // plain identifier — nothing to classify
-    [InlineData("MyType")]
-    [InlineData("")]
-    [InlineData("@class")]        // already escaped — not a bare keyword
-    public void FallsThroughForUnrecognizedToken(string token)
-        => Assert.Null(FidelityCheck.ClassifyCause(token));
-
-    // invisible-rune is checked before keyword-escape: a keyword-looking token
-    // carrying a format rune is classified by the rune, since that is the real
-    // reason binding diverged.
-    [Fact]
-    public void InvisibleRuneTakesPrecedenceOverKeyword()
-    {
-        var result = FidelityCheck.ClassifyCause("class\u200C");
-        Assert.NotNull(result);
-        Assert.StartsWith("invisible-rune", result!.Value.Cause);
+        // Layer 3: an interior Cf that lexed but failed to *bind* is unspeakable -
+        // Roslyn strips Cf from metadata names, so the name came from a non-C#
+        // producer and no source spelling reaches it. Policy fix, no A -> B delta.
+        Assert.StartsWith("    //  cause: unspeakable-name", lines[2]);
+        Assert.Contains("no exact source form", lines[3]);
     }
 
     [Fact]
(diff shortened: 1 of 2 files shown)


richlander:
Reviewed at head `origin/feature/issue-3256-cause-fix-classification` against its base `feature/issue-3238-annotated-compile-back-failure`. The layer‑1/2 render this builds on is good, and the honesty framing — *only classify what's derivable from the emitted token* — is the right constraint.

But the `invisible-rune` branch inverts the language rule it's named for. **Requesting changes.** Everything below is measured against the shipped compiler, not read off the spec.

## The measurements

| # | Program | Result |
| --- | --- | --- |
| 1 | `class MissingType {}` + reference `Missing‹ZWNJ›Type` (literal U+200C) | **compiles and runs** — binds to `MissingType` |
| 2 | `class ‹ZWNJ›Foo {}` (leading rune) | **CS1001** `Identifier expected` |
| 3 | `int class‹ZWNJ› = 5;` | **compiles and runs** — `class‹ZWNJ›` is a legal identifier |
| 4 | compile `class Missing‹ZWNJ›Type {}`, read back the metadata name | `'MissingType'` — chars `004D 0069 0073 0073 0069 006E 0067 0054 0079 0070 0065`, **no U+200C** |

Two distinct rules are in play, and layer 3 conflates them:

- **Keyword recognition uses the raw lexeme.** `class‹ZWNJ›` is *not* the keyword `class` (#3).
- **Identifier equivalence removes Cf.** `Missing‹ZWNJ›Type` *is* `MissingType` (#1).

## 1. The `invisible-rune` fix is wrong in both interior cases (blocking)

The cause text is correct — "binding strips it, so the name no longer matches" — but that same fact makes the paired fix inert or harmful:

**Interior rune, ordinary name.** Per #1, the compiler already ignores the rune. `emit Missing‹ZWNJ›Type → MissingType` yields a token the compiler cannot distinguish from what was already emitted. Apply the fix, recompile, and you get the **identical diagnostic**. The delta is textual, not semantic.

**Interior rune, keyword-shaped stem.** This is the PR's own precedence test, `ClassifyCause("class\u200C")`, and per #3 the emitted token is a *legal identifier*. The rune is precisely what makes it legal. The produced fix — `emit class‹ZWNJ› → class` — **converts a valid identifier into a reserved keyword**, manufacturing an error that wasn't there.

So the documented rationale is backwards for the case it's documented on:

> invisible-rune is checked first because the rune is the real reason binding diverged

For `class‹ZWNJ›` the rune is the reason the token *works*. The precedence rule enshrines the wrong answer, and the test pins it.

**Where the branch is right:** #2. A *leading* Cf is a lex error (Cf is a valid `identifier_part_character` but not `identifier_start_character`), and dropping it genuinely fixes it. That is the case `invisible-rune` should own.

Suggested split, which keeps the honesty rule intact:

- Cf at position 0 (or any position making the token lex-invalid) → `invisible-rune`, fix = drop it. **True.**
- Interior Cf → not a rune problem at all. Either the name is unspeakable (below) or the real cause is elsewhere and it should fall through.

The diagnostic **ID** is a clean discriminator here and doesn't violate the no-message-parsing rule — IDs are locale-stable, unlike `GetMessage()`. `CS1001`/`CS1056`-family means lex; `CS0246`/`CS0117` means binding.

## 2. Interior Cf implies unspeakable, and precedence shadows it (blocking)

Per #4, Roslyn **strips Cf when emitting metadata names**. A C#-produced assembly therefore never carries an interior Cf in a type or member name. When one shows up in a round-trip it came from a non-C# producer (ilasm, obfuscator), and then *no* C# source spelling can reach it — every candidate reduces to the same Cf-stripped identifier, which is not the metadata string.

That is exactly your third row, `unspeakable-name` / *"no exact source form — emit a speakable alias or exclude."* The correct classification is already in the table; the `invisible-rune` check runs first and hides it.

## 3. `Contains('<') || Contains('>')` false-positives on ordinary C# (blocking-ish)

`emittedToken` is `raw.Substring(startCol, endCol - startCol)` over the **raw diagnostic span**, not an identifier. Two consequences:

- A span covering generic syntax — `List<int>`, `Dictionary<string, int>` — classifies as *"compiler-generated identifier isn't legal C#."*
- Worse, the test is character-level, so **any** span containing `=>`, `>=`, `>>`, or `->` matches. Lambda arrows are everywhere in decompiled output.
- And when the diagnostic spans lines, `endCol` falls back to `raw.Length`, so the "token" becomes the whole rest of the line — maximizing the chance of catching a stray `>`.

Match the actual mangling shapes instead (`<>c`, `<>c__DisplayClass…`, `<M>b__0`, `<P>k__BackingField`) and require the span to be a single token before classifying it at all.

## 4. Minor: detection is rune-based, revelation is char-based

`ClassifyCause` detects via `EnumerateRunes()` + `Rune.GetUnicodeCategory`, but `RevealRune(char)` is char-based and only recognises `Format`/`Control`/`NonSpacingMark` on a single `char`. A non-BMP format rune (e.g. U+1D173, Cf) is *detected* as a cause but its surrogate halves classify as `Surrogate` and pass through unrevealed — so the `fix:` line prints an invisible character in the one code path whose entire purpose is making the invisible visible.

## The change that would have caught all of this

The tests call `ClassifyCause` directly with crafted tokens, so they pin the classifier's **output strings** but never test its **claim**. Every one of the findings above survives a green suite.

The PR states the premise that makes the claim falsifiable:

> Compile-back is a **round-trip**, so the harness holds the emitted text

Then hold the fix to it: **apply the fix to the emitted source, recompile, and assert the diagnostic is gone** (and that no new one appeared). That is a real gate, it needs no new infrastructure, and it converts `fix:` from advice into a verified statement. It would have failed immediately on `Missing‹ZWNJ›Type` (unchanged diagnostic) and on `class‹ZWNJ›` (new keyword error).

This is the same standard #3249 just set for byte-neutrality — a declared property gets a check, not a doc comment. A `fix:` line printed next to a caret is a strong claim to a reader; it should be at least as gated.

## Reproduction

```bash
# 1 — interior rune binds (compiles, prints 42)
printf 'class MissingType { public int V=42; }\nclass P { static void Main(){ var x=new Missing\u200cType(); System.Console.WriteLine(x.V);} }\n' > a.cs

# 3 — keyword+rune is a legal identifier (compiles, prints 5)
printf 'class R { static void Main(){ int class\u200c = 5; System.Console.WriteLine(class\u200c);} }\n' > c.cs

# 2 — leading rune fails to lex (CS1001)
printf 'class \u200cFoo {}\n' > b.cs
```

(#4 was read back with `System.Reflection.Metadata` off the compiled `lib.dll`.)

richlander:
Reproduced every measurement against the shipped compiler — all four hold. The classifier was wrong exactly where you measured, and it's rebuilt around what the compiler actually does. Pushed as `Rebuild the cause/fix classifier around measured compiler behavior`.

## The core correction

Two rules, now kept separate instead of conflated:

- **Cf is an identifier-*part* but not an identifier-*start*** → a **leading** Cf breaks *lexing* (CS1001/CS1056). That is the only case `invisible-rune` now owns, and dropping the rune is a real fix.
- **Cf is stripped for identifier equivalence and for emitted metadata names** → an interior Cf on a *bound* name (#1, #4) can only have come from a non-C# producer, and no source spelling reaches it. That is `unspeakable-name` (policy fix) — the classification you pointed at, no longer shadowed.

## Findings

**1 + precedence (`class‹ZWNJ›`).** Fixed. `invisible-rune` is gated on a lex-error id; on interior Cf it never fires. And because dropping must yield a *valid identifier*, `class‹ZWNJ› → class` can't be produced (`class` isn't a valid identifier) — the harmful fix is now impossible, and a test pins that it never becomes keyword-escape.

**2 (interior Cf ⇒ unspeakable).** Done — interior Cf + a binding id classifies as `unspeakable-name` with the policy fix.

**3 (`Contains('<'|'>')`).** The heuristic is gone. Spans are widened to the enclosing identifier first; a generated-name span is a bare `<` (measured: CS1526), which isn't an identifier, so it falls through. Tests cover `List<int>`, `=>`, and the bare `<`.

**4 (rune vs char reveal).** The fix line now reveals rune-aware, so a non-BMP Cf is expanded, not printed invisibly.

**The gate.** This was the right ask. `ClassifyCause` now returns the raw `From`/`To`, and the tests **apply the fix, recompile, and assert the diagnostic is gone** (`CompilesClean`) — for both `invisible-rune` (leading Cf) and `keyword-escape`. A `fix:` line is now a checked statement, matching the #3249 byte-neutrality standard. All compilation routes through one helper, so no new dynamic site.

## Authentic render (generated)

```csharp
class ‹ZWNJ›Foo { }
//    ^^^^^^ CS1001: Identifier expected
//  cause: invisible-rune — a format-category rune (Cf) in '‹ZWNJ›Foo' breaks the token; C# accepts Cf as an identifier-part but not an identifier-start.
//  fix:   emit  ‹ZWNJ›Foo → Foo   (drop the format-category rune)
```
```csharp
class C { int event; }
//            ^^^^^ CS1519: Invalid token 'event' in a member declaration
//  cause: keyword-escape — emitted identifier 'event' is a C# keyword written bare.
//  fix:   emit  event → @event
```
```csharp
class C { void M() { Missing‹ZWNJ›Type x = null; } }
//                   ^^^^^^^^^^^^^^^^^ CS0246: The type or namespace name 'MissingType' could not be found (are you missing a using directive or an assembly reference?)
//  cause: unspeakable-name — 'Missing‹ZWNJ›Type' carries a format-category rune (Cf) that C# strips from identifiers, so no source spelling matches the metadata name.
//  fix:   no exact source form — emit a speakable alias or exclude from round-trip.
```

The motivating explicit-interface ZWNJ case lands as `unspeakable-name` — which, per your #2, is the correct answer.

richlander:
Re-reviewed at the fixed head `184a4423` in a clean worktree, built `-c Release`. `AnnotatedCompileBackFailureTests` **10/10 pass**.

This is a good rebuild. The position split (Cf is an identifier-*part* but not an identifier-*start*) is the right axis, gating `invisible-rune` on `LexErrorIds` keeps it locale-stable without message parsing, interior Cf now reaches `unspeakable-name` instead of being shadowed, and the `<`/`>` character test is gone.

**And the gate is real.** `CompilesClean` genuinely compiles the patched source and asserts no errors remain — a `fix:` line is now a checked claim rather than advice. That was the ask and it landed properly.

Which is why it's worth pointing out that the gate catches a residual bug the moment you give it the right case.

## One finding: the harmful bare-keyword fix is still reachable (blocking, measured)

The PR states:

> because dropping must yield a *valid identifier*, `class‹ZWNJ› → class` can't be produced (`class` isn't a valid identifier)

`SyntaxFacts.IsValidIdentifier` does not mean that. It checks character rules only — identifier-start plus identifier-parts — and knows nothing about keywords:

```
IsValidIdentifier(class )       = True
IsValidIdentifier(event )       = True
IsValidIdentifier(int   )       = True
IsValidIdentifier(@class)       = False      <-- note the direction
IsValidIdentifier(<empty>)      = False
```

So the guard passes for `class`, and the harmful fix is still producible. It just needs a **leading** Cf rather than an interior one, which is precisely the case `invisible-rune` now owns.

End-to-end through the real render path (`ClassifyCause` fed from an actual diagnostic, fix applied, recompiled):

```
[leading Cf + keyword stem]   class ‹ZWNJ›class { }
    id=CS1001  cause=invisible-rune  fix=‹ZWNJ›class -> class
    afterApply errors=5  FIRST=CS1001          <-- fix makes it worse

[leading Cf + ordinary stem]  class ‹ZWNJ›Foo { }      (control)
    id=CS1001  cause=invisible-rune  fix=‹ZWNJ›Foo -> Foo
    afterApply errors=0  CLEAN                 <-- fix works
```

All three preconditions are satisfied by a single input: leading Cf → CS1001 (in `LexErrorIds`), stripped stem → `class`, `IsValidIdentifier("class")` → true. The proposed fix produces a bare keyword, and the diagnostic it claims to clear is still there.

**The fix for the fix:** when `stripped` is a reserved keyword, the correct proposal is `@class`, not `class` — drop the rune *and* escape. Something like `SyntaxFacts.GetKeywordKind(stripped) != SyntaxKind.None ? "@" + stripped : stripped`, which composes the two causes rather than picking one. (`IsValidIdentifier` can stay as the "did stripping leave something nameable at all" check — it's just not a keyword filter.)

## Why the existing precedence test doesn't cover this

`KeywordStemCarryingCfIsNeverConvertedIntoAKeyword` passes `"CS0103"` — a **binding** id. `LexErrorIds` is `{CS1001, CS1056}`, so `invisible-rune` can't fire on that input under any implementation, and the test passes without ever reaching the branch it is named for. Two other things weaken it:

- it hand-builds the token (`ClassifyCause("class" + Zwnj, 0, 6, "CS0103")`) instead of classifying a real diagnostic, so it skips the span-widening the render site does — the one part most likely to surprise;
- the body is `if (cf is { } value) { … }`, so when `cf` is null it asserts nothing at all. A test that can silently pass by classifying nothing can't pin a precedence rule.

Worth converting it to the same real-diagnostic + `CompilesClean` shape as its neighbours.

## Suggested change

Add the keyword stem to the verified-by-recompile test — ideally turn `InvisibleRuneFixIsVerifiedByRecompile` into a `[Theory]` over `class ‹ZWNJ›Foo { }` and `class ‹ZWNJ›class { }`, both asserting `CompilesClean(patched)`. It fails today on the second row and passes once the `@` escape lands.

That's the whole value of what you built here: the gate turns "the classifier is honest" from an argument into a run. It just needs the adversarial rows, not only the happy ones — the same reason the `<`/`>` heuristic survived until it was measured.

Everything else from my earlier review checks out at this head. Fix this one and I'm satisfied.

richlander:
Confirmed and fixed. You were right on every point, including that my "can't be produced" claim was false.

**Reproduced first.** `IsValidIdentifier` is a character-rules check with no keyword knowledge:

```
IsValidIdentifier(class  ) = True   keywordKind=ClassKeyword
IsValidIdentifier(event  ) = True   keywordKind=EventKeyword
IsValidIdentifier(int    ) = True   keywordKind=IntKeyword
IsValidIdentifier(@class ) = False  keywordKind=None
```

So all three preconditions hold and the harmful fix is produced. End-to-end through the real classifier:

```
[leading Cf + keyword stem] : id=CS1001 cause=invisible-rune fix=<ZWNJ>class -> class
    afterApply errors=5 FIRST=CS1001
[leading Cf + ordinary stem]: id=CS1001 cause=invisible-rune fix=<ZWNJ>Foo -> Foo
    afterApply errors=0 CLEAN
```

**Fix.** The two causes compose — drop the rune *and* escape the keyword it exposes. `IsValidIdentifier(stripped)` stays as the "did stripping leave something nameable at all" check; `GetKeywordKind(stripped) != SyntaxKind.None` selects the escaped proposal. Same probe after:

```
[leading Cf + keyword stem] : id=CS1001 cause=invisible-rune fix=<ZWNJ>class -> @class
    afterApply errors=0 CLEAN
[leading Cf + ordinary stem]: id=CS1001 cause=invisible-rune fix=<ZWNJ>Foo -> Foo
    afterApply errors=0 CLEAN
```

Authentic render (generated, not hand-written):

```
class ‹ZWNJ›class { }
//    ^^^^^^ CS1001: Identifier expected
//  cause: invisible-rune — a format-category rune (Cf) in '‹ZWNJ›class' breaks the token; C# accepts Cf as an identifier-part but not an identifier-start.
//  fix:   emit  ‹ZWNJ›class → @class   (drop the format-category rune, and escape the keyword it exposes)
```

**Tests.** Took your prescription. `InvisibleRuneFixIsVerifiedByRecompile` is now a `[Theory]` over `class ‹ZWNJ›Foo { }` and `class ‹ZWNJ›class { }`, both gated on `CompilesClean`, plus an assertion that the proposal is never a bare reserved keyword. I ran it against the pre-fix classifier to prove the gate actually catches this rather than trusting it:

```
InvisibleRuneFixIsVerifiedByRecompile(source: "class ‌class { }\n") [FAIL]
  fix proposed a bare keyword: 'class'
LeadingCfOverAKeywordStemProposesTheEscapedForm [FAIL]
  Expected: "@class"  Actual: "class"
Total: 11, Failed: 2
```

The control row passed throughout, so the theory is discriminating, not uniformly broken.

You were also right that `KeywordStemCarryingCfIsNeverConvertedIntoAKeyword` was inert on all three counts — binding-error id so the branch could never fire, hand-built token that skipped span widening, and an `if (cf is { } value)` body that asserted nothing on null. Replaced with `LeadingCfOverAKeywordStemProposesTheEscapedForm`, which classifies a real diagnostic and asserts unconditionally.

Green at `d1242c3a`: 11/11 annotated-failure, 4/4 dynamic-site governance (no new compilation site), 38/38 `Area=Fidelity` non-slow.

richlander:
Re-reviewed at the fixed head `d1242c3a` in a clean worktree. **Approve.**

The composition is right: `GetKeywordKind(stripped) != SyntaxKind.None → "@" + stripped`, so the fix now drops the rune *and* escapes what stripping exposes, instead of picking one cause and making the other worse.

I swept it adversarially rather than re-running the case I'd already reported — every concrete fix applied to the source and recompiled:

```
[ordinary stem     ] CS1001  invisible-rune    <Z>Foo    -> Foo      => CLEAN
[reserved: class   ] CS1001  invisible-rune    <Z>class  -> @class   => CLEAN
[reserved: int     ] CS1001  invisible-rune    <Z>int    -> @int     => CLEAN
[reserved: void    ] CS1001  invisible-rune    <Z>void   -> @void    => CLEAN
[contextual: var   ] CS1001  invisible-rune    <Z>var    -> var      => CLEAN
[contextual: record] CS1001  invisible-rune    <Z>record -> record   => CLEAN
[contextual: value ] CS1001  invisible-rune    <Z>value  -> value    => CLEAN
[all-Cf identifier ] CS1001  unspeakable-name  (policy fix, no delta)
[interior Cf, bound] CS0246  unspeakable-name  (policy fix, no delta)
[trailing Cf       ] no error emitted
```

Three edges worth noting, because they're the ones an over-eager fix would have gotten wrong:

- **Contextual keywords are correctly left alone.** `GetKeywordKind` returns `None` for `var`/`record`/`value`, and they need no escape because they're legal identifiers. `IsReservedKeyword`-style filtering is exactly the right granularity here — escaping them would have been noise, and using a broader keyword predicate would have produced `@var` for no reason.
- **An all-Cf identifier degrades correctly.** Stripping leaves the empty string, `IsValidIdentifier("")` is false, so it falls through to `unspeakable-name` rather than proposing a fix to nothing.
- **A trailing Cf produces no diagnostic at all**, which is the position rule working — Cf is a legal identifier-*part*.

The test changes are the substantive part. `InvisibleRuneFixIsVerifiedByRecompile` is now a `[Theory]` over both stems with `CompilesClean` on each, and the old `KeywordStemCarryingCfIsNeverConvertedIntoAKeyword` — which passed a binding id (`CS0103`) that `LexErrorIds` can never match, and wrapped its asserts in `if (cf is { } value)` so it asserted nothing on null — is replaced by `LeadingCfOverAKeywordStemProposesTheEscapedForm`, driven by a real diagnostic with unconditional assertions. The added `Assert.False(IsReservedKeyword(GetKeywordKind(To)))` inside the theory generalizes past the one case, so a future cause that proposes a bare keyword fails regardless of which branch produced it.

That's the whole point of the gate, and it's now doing the work: this bug was found by adding a row, not by reading the code.

**Verified:** full `dotnet-inspect.slnx -c Release` build clean; `AnnotatedCompileBackFailureTests` 11/11; decompiler `--gate fast` **3820 tests, 0 failed** (97s).

No further findings. Nice iteration — three rounds, each one measured rather than argued.

PR #3261, opened 2026-07-26
Give the IR invariant check teeth: leveled structural + semantic checks (#3241, PR2)

richlander:
Stacked on #3250 (PR1). Base = `fix/3241-ir-invariants-release`; rebase onto main once PR1 merges.

## What

PR1 made `IrNode.CheckInvariant` actually run in the Release test suite and corpus sweeps (it was a no-op `[Conditional("DEBUG")]` call in the Release build the suite uses). This PR adds the first real teeth — a **leveled** check — without regressing the suite.

### Structural (suite-wide)
Parent/child back-pointer + slot consistency. Holds on *any* well-formed subtree, including the deliberately minimal `IrFunction`s that hand-built pass-unit fixtures construct. Runs suite-wide via the test host (`DOTNET_INSPECT_IR_INVARIANTS=1`).

### Semantic (corpus-only)
Holds only on fully-formed importer output. First invariant: **local-slot range** — every `ldloc`/`stloc`/`ldloca` references a slot that exists in the *nearest enclosing local scope*. `IrFunction`, `Lambda`, and `LocalFunctionStatement` each carry their own `Locals`, so the check scopes to the nearest, not the root. A pass that drops a local without repointing its readers, or fabricates a dangling slot, now trips the check instead of surfacing as a downstream miscompile.

Gated by `IrInvariants.CheckSemantics` (`DOTNET_INSPECT_IR_INVARIANTS=full`) so it runs over the corpus — where input is well-formed — without false-positiving the ~120 minimal-fixture pass tests (which reference slots without populating `Locals`).

## Design

- `CheckInvariant()` stays parameterless (structural only) — preserving PR1's method-group negative-control guard (`Assert.Throws(parent.CheckInvariant)`, which fails to compile if `[Conditional]` is re-added).
- New `CheckInvariant(bool includeSemantics)` threads the level; the 5 per-pass hooks pass `IrInvariants.CheckSemantics`.
- Semantic tests call the `bool` overload **directly** so they never flip the global flag under xUnit's parallel collections.

## Validation

- Fast suite: back to **zero real failures** with semantic off suite-wide (the 123 fixture false-positives an ungated semantic check produced are gone). Lone red is the known env-only `LookalikeExpression` missing-fixture (test-csproj-only; green in CI/full slnx).
- New `IrInvariantCheckTests`: 9/9 pass (structural negative-controls, semantic in/out-of-range, structural-ignores-slot crux, lambda scoping).
- Harness `--gaps` over CoreLib with `DOTNET_INSPECT_IR_INVARIANTS=full`: **0 pass bugs across 41,952 methods**, identical to baseline (40460 raised / 1492 gaps) — proving local-slot is a true invariant on real IL, not just fixtures.

Refs #3241, #3240.


richlander:

Two-model review (both latest, different families), each isolated in its own `/tmp` worktree, scratch-only, no tracked-file edits.

| Model | Round 1 (`1695acb3`) | Final (`46b9ae23`) |
| --- | --- | --- |
| Gemini `gemini-3.1-pro-preview` | CLEAN | CLEAN |
| GPT `gpt-5.6-sol` | **NOT CLEAN** (2 findings) | CLEAN |

### Findings (both from GPT; both real, both fixed)

**1. Empty-`Locals` lambda false-positive** — fixed in `65cd620c`.
A `Lambda`/`LocalFunctionStatement` with an empty local table *shares* its host's scope: its `LoadLocal`/`StoreLocal` indices reference the outer function's locals by outer index (how `LambdaRaisingPass` substitutes captured locals, and how the printer scopes it via `NeedsNestedLambdaScope`/`NeedsNestedLocalFunctionScope`). The original `OwnLocalScope()` unconditionally rebound to a phantom zero-slot scope, so a valid captured `LoadLocal(1)` was validated against 0 slots and wrongly rejected. Fix: an empty-`Locals` capturing body inherits the enclosing scope; the root `IrFunction` still always opens a scope. Regression: `SemanticCheck_EmptyLocalsLambda_SharesTheOuterScope`.

**2. Static-local-function false-negative** — fixed in `46b9ae23`.
The fix above was too broad: a **static** `LocalFunctionStatement` cannot capture, so its empty `Locals` is a *genuine* zero-slot scope and a `LoadLocal` there is dangling. Blanket inheritance silently accepted it. Fix: a static `LocalFunctionStatement` always opens its own scope (`Locals.Length`, even zero); only non-static empty-`Locals` bodies inherit. Strictly tightening (a static body can never legitimately reference an outer local), so no new false positive. Regression: `SemanticCheck_StaticLocalFunction_OpensItsOwnZeroSlotScope`.

### Verified / dismissed at the final head
- Scope resolution (lambda-in-lambda, local-function-in-lambda, mid-tree explicit calls), root-function slot coverage, the `-1` sentinel (no spurious throw, no silent skip), leveling soundness (structural never touches `Locals`), the two-overload method group preserving PR1's `Assert.Throws(parent.CheckInvariant)` guard, and test global-flag hygiene — all confirmed CLEAN by both models.
- **Bounded, non-blocking**: the `Lambda` IR node carries no static/instance flag, so a hypothetical static empty-`Locals` lambda is treated as shared. This is a no-false-positive under-approximation (shared is the looser interpretation), documented on `OwnLocalScope`. GPT explicitly agreed it is not blocking; distinguishing would need a model change → non-blocking follow-up.

### Validation at `46b9ae23`
- `IrInvariantCheckTests` 11/11.
- Fast suite back to zero real failures (semantic level is off suite-wide; the lone red is the known env-only `LookalikeExpression` missing-fixture, green in CI/full slnx).
- Harness `--gaps` over CoreLib with `DOTNET_INSPECT_IR_INVARIANTS=full`: **0 pass bugs across 41,952 methods**, identical to baseline (40460 raised / 1492 gaps) — before and after both fixes.

### CI note
CI (`ci.yml`) triggers only on PRs targeting `main`, so this stacked PR (base `fix/3241-ir-invariants-release`) shows no checks yet. It will run on the merge path once #3250 merges and this retargets to `main`.

richlander:
Reviewed at `893d1a32` in a clean worktree (stack reviewed together; gate findings are on #3262). **Approve.**

The leveling is the correct call, and the reasoning behind it is sound: structural invariants hold on any well-formed `IrNode` graph including minimal hand-built fixtures, semantic ones need real importer output. Running local-slot-range suite-wide would have false-positived ~120 fixture tests, and the tempting alternative — loosening the invariant until the fixtures pass — would have produced a check too weak to catch anything. Splitting the level instead of weakening the check is the right trade.

Things I checked adversarially and found sound:

**`EnclosingLocalScope()` on the root.** It walks from `Parent`, so calling `function.CheckInvariant(includeSemantics: true)` on an `IrFunction` returns `-1` for the root itself — but `CheckSubtree` recomputes `childScope` from `OwnLocalScope()`, so children are validated against `f.Locals.Length`. Non-vacuous at the root, which is how the corpus gate calls it.

**The capturing-lambda case, which is where I expected this to break.** If a capturing lambda with a non-empty `Locals` table could reference an *outer* local by outer index, the check would throw a false positive. It holds because the decompiler raises lambdas out of compiler-generated closure classes, so captured variables arrive as *field* accesses, not `LoadLocal` — there is no outer-indexed local reference left to misvalidate. The zero-violation measurement over CoreLib backs it, and I re-measured that corpus at 41,952 methods with the semantic check on and saw no violations.

**The `Lambda` under-approximation is disclosed rather than hidden.** `Lambda` carries no static/instance flag, so an empty-`Locals` static lambda is treated as scope-sharing. The doc comment names this as a deliberate no-false-positive under-approximation instead of implying the check is total. That's the honest framing, and it's the right direction to err in for an invariant that aborts the pipeline.

**Negative controls exist for the new level**, not just the old one — `SemanticCheck_ThrowsWhenLocalSlotIsOutOfRange` and `StructuralCheck_IgnoresLocalSlotRange_ButSemanticCheckCatchesIt` together pin both that the semantic check fires and that the structural level genuinely doesn't. Verified: **11/11 pass** in Release, 0.155s.

One observation about scope rather than a defect: the check is asymmetric by construction — it catches a *dangling* slot index (out of range) but not a *misrepointed* one (in range, wrong local). A pass that swaps two live slots passes cleanly. The doc comment is accurate about this ("drops a local without repointing its readers, or fabricates a dangling slot"), so nothing to change; worth knowing when judging how much the invariant is buying. Range is the cheap half; identity would need def-use tracking.

richlander:
Final pre-merge check on the stack: reviewed at tip `8e27d8be`, merged current `main` (`14781e8d`) into it — clean — **3840 passed, 0 failed**.

Approval here stands. One addition worth recording on this PR specifically, since it's PR 2 that added the semantic level: I verified the semantic invariant has real teeth end-to-end rather than assuming it. Injecting a fault pass that fabricates a dangling local slot (`LoadLocal` index 9999) — a *semantic* violation the per-pass structural gate cannot see — fails the corpus gate with `7502 method(s)`. So `ValidateLocalSlot` catches a dangling slot through the real pipeline, not just in its unit fixtures.

That also confirms the point from my review above: this level executes **only** in #3262's corpus gate, which is why moving that gate into the per-PR lane mattered. Details: https://github.com/richlander/dotnet-inspect/pull/3262#issuecomment-5086050463

**Good to merge**, in stack order #3250 → #3261 → #3262.

PR #3262, opened 2026-07-26
Wire the semantic IR invariants into the CoreLib corpus gate (#3241, PR3)

richlander:
Stacked on #3261 (PR2), which is stacked on #3250 (PR1). Base = `fix/3241-ir-invariants-teeth`; retarget to main as the stack merges.

## What

PR1 (#3250) made `IrNode.CheckInvariant` run in the Release test suite. PR2 (#3261) added the first semantic teeth — local-slot range — which hold only on fully-formed importer output and are therefore kept off the minimal-fixture suite (gated behind `IrInvariants.CheckSemantics`). Without a gate, those teeth only bite when someone manually sets `DOTNET_INSPECT_IR_INVARIANTS=full`. **This PR makes them bite automatically in CI.**

`CorpusSweepGateTests` (`Area=Corpus`, run in `deep-inspect.yml` and `release.yml` via `--gate corpus`) already walks every CoreLib method through `IrImporter → IrPasses` and asserts zero pass-bugs plus Full-fidelity/fully-raised floors. This adds a **fourth health floor**: every method's final IR must satisfy the semantic invariants.

## Design

- Calls `function.CheckInvariant(includeSemantics: true)` **explicitly per method** rather than flipping the global `IrInvariants.CheckSemantics` flag — xUnit runs collections in parallel and the global flag would false-positive the ~120 minimal-fixture pass tests running concurrently.
- Reuses the **single existing corpus walk** — no second 42k-method sweep. Adds one tree walk per method, negligible next to `IrPasses.Run`.
- Asserts the **final tree** is semantically well-formed (the right granularity for an output-quality gate). Per-pass semantic sweeping, which also catches transient self-healing mid-pass shapes, remains available for debugging via the harness `--gaps` with `=full`.

## Validation

Gate runs green — **0 semantic violations across the full CoreLib corpus** (41,952 methods), matching the harness `--gaps =full` result. `CorpusSweepGateTests` completes in 11.8s.

## Risk

Test-only, additive assertion reusing the `CheckInvariant` path that PR2's 11 hermetic tests already prove throws on violations. No product code change. Low risk.

Refs #3241, #3240.


richlander:
Reviewed the full stack in a clean worktree at `893d1a32` (`#3250` → `#3261` → `#3262`), built `dotnet-inspect.slnx -c Release`, ran the invariant tests and this gate. **Approve with two fixes**, both in this PR's numbers rather than its design.

The design is right and the stack closes #3241 properly — see my notes on #3250/#3261. This comment is about the gate.

## 1. The floor comment is stale, and both floors are mis-calibrated (should fix)

I re-measured by forcing each floor to 100 and reading the failure message:

```
Full-fidelity rate 98.18% (41190/41952) fell below the 100% floor
Fully-raised rate  97.96% (41095/41952) fell below the 100% floor
```

Against the comment at `CorpusSweepGateTests.cs:35-37`:

| | comment says | measured (net11, this worktree) | floor | real margin |
| methods | 41,159 | **41,952** | — | — |
| Full | 98.12% | **98.18%** | 98.0 | **0.18pp ≈ 75 methods** |
| fully raised | 95.75% | **97.96%** | 94.0 | **3.96pp ≈ 1,660 methods** |

Three problems, and they point in opposite directions:

- **`41,159` is not the method count.** The total is 41,952; 41,190 is the *Full* count. `IrInvariants.cs` independently says "41,952 methods," so the two files in this stack disagree and this one is wrong.
- **The Full floor is far tighter than documented.** "They sit a couple points below the measured numbers" is the stated policy, but 98.18 → 98.0 is **0.18** points. A runtime patch bump that shifts ~75 methods flakes CI, which is exactly the failure mode the "floors, not baselines" rationale was chosen to avoid.
- **The fully-raised floor has decayed.** 94.0 *was* a couple points under 95.75 when written; the real number is now 97.96, so the gate would sit green through a structuring regression costing **~1,660 methods**. The breadth signal for structuring is currently ~4pp of slack.

Re-measure and restate both floors from the current numbers (and fix the total). The decay is worth a line in the comment too — a floor that isn't ratcheted when the metric improves silently becomes a no-op, which is the same shape of problem the stack exists to fix.

## 2. Capped lists are reported as totals (should fix)

`CorpusSweepGateTests.cs` caps both diagnostic lists at 20:

```csharp
if (passBugs.Count < 20) passBugs.Add(...);
...
Assert.True(passBugs.Count == 0,
    $"...but {passBugs.Count} method(s) failed:\n  " + ...);
```

`passBugs.Count` is the *capped* count, so a run with 500 broken methods reports "20 method(s) failed." Same for `semanticViolations`. The number a failing CI log shows is the sample size, presented as the population — and it silently stops distinguishing "a regression" from "a catastrophe" at exactly the point that distinction starts to matter.

Keep a separate uncapped `int` and report `"{n} method(s) failed (showing first 20)"`. The house standard is already in the repo — `ValidityCheck.cs:526-528` prints the checked-vs-eligible split and an explicit *"NOTE: semantic findings are per-sample, not corpus-wide"* whenever it samples.

## 3. Worth reconsidering: this gate costs 12 seconds, and PR CI never runs it

```
$ dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build \
    -- -class "ILInspector.Decompiler.Tests.CorpusSweepGateTests"
Total: 1, Errors: 0, Failed: 0, Skipped: 0  Time: 11.974s
real 0m12.677s
```

`[Trait("Area", "Corpus")]` puts it behind `--gate corpus`, described in `Program.cs:34` as "the multi-hour sweep." CI line 301 runs `-trait- "Speed=Slow"`, so per-PR it never executes; it runs weekly in `deep-inspect.yml` and at release.

Since `CheckSemantics` is deliberately off in the unit-test host, **this gate is the only place semantic invariants execute anywhere.** So after the stack lands: structural invariants have real per-PR teeth (genuinely good), and semantic invariants have up-to-7-day feedback latency, with the PR that breaks one going green.

For 12 seconds that's a bad trade. The `Area=Corpus` trait is calibrated for the multi-hour sweeps and this test isn't one. I'd either drop `Speed=Slow` and keep `Area=Corpus`, or split the health-floor assertions (slow, drifty, runtime-version-sensitive) from the pass-bug and semantic-violation assertions (fast, absolute, zero-tolerance) and let the latter run per-PR.

That's the one place the stack's own thesis — *a correctness check that doesn't execute in the run people actually do asserts nothing* — still applies to its top layer.

## Verified

- `IrInvariantCheckTests`: **11/11 pass**, 0.155s.
- `CorpusSweepGateTests`: **1/1 pass**, 12.0s, on unmodified floors.
- Assert ordering note (minor): pass-bugs abort before the percentages are computed, so a run with any pass bug reports no fidelity numbers at all. Fine as-is, just noting it.

richlander:
Applied all three fixes in af12fd01.

**1. Split the gate so the absolute checks run per-PR.** Factored the single 42k-method walk into `SweepCoreLib()` and asserted over its result from two `[Fact]`s:

- `CoreLibSweep_IsExceptionSafeAndSemanticallyValid` — **not** `Speed=Slow`, so it runs in the per-PR lane. Asserts 0 pass-bugs and 0 semantic-invariant violations (zero-tolerance). As you noted, this gate is the *only* place semantic invariants execute — the unit-test host runs the structural level only — so per-PR is exactly where the teeth have to bite.
- `CoreLibSweep_MeetsHealthFloors` — keeps `Speed=Slow`. The drifty, runtime-version-sensitive Full / fully-raised floors stay in the corpus gate.

`Speed=Slow` moved off the class onto the floors method; `Area=Corpus` stays at class level so both still run under `--gate corpus`. Verified routing:
- `-trait- "Speed=Slow"` (PR lane) → discovers **1** test (the fast one), 12.0s green.
- `-trait "Area=Corpus"` (deep-inspect / release) → discovers **2**, 17.5s green.

**2. Report true, uncapped counts.** `SweepResult` now carries `PassBugCount` / `SemanticViolationCount` (uncapped `long`) separately from the 20-item sample lists. Messages read `"N method(s) (showing first 20)"` when the population exceeds the sample, mirroring the house standard in `ValidityCheck.cs`.

**3. Recalibrated floors + fixed the stale comment.** The comment said 41,159 methods — wrong. Re-measured on net11 CoreLib:

```
total=41952  full=41190 (98.18%)  raised=41095 (97.96%)  passBugs=0  sem=0
```

Floors set to **96.0 / 96.0** — ~2 points below measured, per the "a couple points below" policy. This loosens the flaky 98.0 Full floor (was 0.18pp below → a patch bump flakes it) and tightens the decayed 94.0 fully-raised floor (was ~4pp below → would stay green through a ~1,600-method structuring regression). The class doc now notes the ratchet-or-decay hazard explicitly.

Both methods green; per-PR routing and corpus-gate routing both verified above.

richlander:
Final pre-merge look at the whole stack, reviewed at tip `8e27d8be` (contains all six commits across #3250 → #3261 → #3262). **All three approved; both of my #3262 fix requests are closed, and the third fix is better than what I asked for.**

## My open findings on #3262

| finding | status at `8e27d8be` |
| --- | --- |
| Gate carried `Speed=Slow`, so a 12s absolute check ran only weekly/at release | **fixed** — `Speed=Slow` moved off the class onto `CoreLibSweep_MeetsHealthFloors`; `Area=Corpus` stays at class level |
| Capped-at-20 lists reported as totals | **fixed** — `PassBugCount`/`SemanticViolationCount` are uncapped `long`s, separate from the sample lists |
| Stale comment (41,159) and mis-calibrated floors | **fixed** — comment re-measured to 41,952 / 98.18% / 97.96%; floors 96.0 / 96.0 |

Verified routing against the filter CI actually uses (`ci.yml:301`):

```
-trait- "Speed=Slow"  (PR lane)      → 1 test in the gate class, 12.1s, green
-trait  "Area=Corpus" (deep/release) → 77 tests, 17.6s, green
```

## The fix I want to call out

Splitting `SweepCoreLib()` into a fast absolute gate plus a slow floors gate is the right resolution, and it's the one that makes this stack coherent. The semantic invariants from #3261 execute **nowhere but here** — the unit-test host runs the structural level only — so leaving them in a lane with up-to-7-day latency would have left PR 2's teeth not biting on the run people actually do. That was the stack's own thesis pointed at its top layer, and it's now closed.

I didn't take that on trust. I injected a fault pass at the end of `IrPasses.Default` that fabricates a dangling local slot (`LoadLocal` index 9999) — deliberately a *semantic* violation, so the per-pass structural gate cannot see it and only the corpus gate's `CheckInvariant(includeSemantics: true)` can. Result in the PR lane:

```
CoreLibSweep_IsExceptionSafeAndSemanticallyValid [FAIL]
  Every method's final IR must satisfy the semantic invariants (local-slot range),
  but 7502 method(s) (showing first 20):
```

That single run confirms three things at once: the gate genuinely fails per-PR, the semantic invariant catches a dangling slot end-to-end through the real pipeline, and the uncapped count reports **7502**, not 20. Restored and re-verified clean afterward.

Two more things I checked because they're the classic ways a corpus gate goes vacuous:

- `CoreLibPath => typeof(object).Assembly.Location` — no probing, no file-existence branch, so there is **no silent-skip path**. A missing corpus throws.
- `Assert.True(r.Total > 10_000)` in **both** tests is a real anti-vacuity guard: a sweep that imports nothing fails instead of passing trivially. Good instinct, and consistent with the non-vacuity discipline in #3250's `PipelineRunner_UnderTestHost_...` test and in #3249.

Floor calibration is also right in both directions: 98.0 Full was 0.18pp above measured (a runtime patch bump would flake it) and 94.0 fully-raised was ~4pp below (it would stay green through a ~1,600-method structuring regression). 96.0/96.0 fixes a flake *and* a blind spot, and the ratchet-or-decay hazard is now documented rather than latent.

## Merge path

`main` has moved to `14781e8d` (it now carries #3231). I merged that into the stack tip — clean — rebuilt, and ran the full suite both ways:

```
--gate fast            → 3840 passed, 0 failed  (98.1s)
-trait- "Speed=Slow"   → 3840 passed, 0 failed  (106.3s)   # exactly what ci.yml:301 runs
```

**Good to merge.** Merge in stack order (#3250 → #3261 → #3262) so each retargets cleanly.

## Non-blocking follow-up (restating from #3250, unchanged)

The invariant is opt-in and default-off, so its value depends on every host that should enable it actually doing so — today the test host and the harness. A future entry point that forgets gets no validation, and the failure mode is identical to the one #3241 fixed, just relocated from the compiler to the caller. A shared validating-host helper, or an assertion that `IrInvariants.Enabled` is set, would close it. Worth a follow-up issue rather than more scope here.

richlander:
Filed the non-blocking follow-up from my review as #3267 (`IrInvariants` is opt-in/default-off, so a future host silently gets no validation), and the generalized convention as #3271. Neither blocks this stack.

PR #3270, opened 2026-07-27
Pin the span-attribution allowlist to methodologyVersion (#3079)

richlander:
## Summary

Converts the README's allowlist rule from prose into a tripwire. Follow-up to the non-blocking finding on #3231.

The body-intrinsic error allowlist (`SpanAttribution.BodyIntrinsicSemanticErrorIds`) is the operative definition of `productBodyDefect` under the stamped `methodologyVersion` — but the two lived in different files with no code path between them, and `BodyIntrinsicSemanticErrorIds` had zero references repo-wide outside its own file.

## The gap was real, and partially unguarded

Reviewer's tampering runs against the full `--gate fast`, `methodologyVersion` left at 2 in every one:

| allowlist addition | category | before |
| --- | --- | --- |
| `CS0165` | definite assignment | caught (1 failed) |
| `CS0246, CS0103, CS1061, CS0234` | resolution | caught (2 failed) |
| `CS0029, CS1503, CS0136` | conversion / overload / scope collision | **not caught** (3828 passed, 0 failed) |

The *enumerated* never-credited IDs were genuinely pinned by the existing close-negative tests. But the README also forbids whole **categories**, and those had no pin.

`CS0136` is the sharp end: #3231's own review identified it as shell-dependent — Gemini's probe showed a shell parameter collision yields exactly `CS0136` — i.e. precisely the class condition 2 of the soundness rule exists to exclude. Nothing stopped it being added tomorrow under an unchanged `v2`.

The damage isn't only a wrong count. Rows sharing a stamp are supposed to be comparable, so a silent widening under a fixed `v2` would make the history card chart a **v2 → v2 step across two different methodologies**, defeating the boundary split that `Render_MovementSplitsProductDefectAcrossMethodologyBoundaryWithoutCharting` exists to enforce.

## Change

Applies the house pattern from #3249 — the declaration drives the enforcement set, asserted by set equality:

- **`MethodologyVersion` moves next to the allowlist that defines it.** Widening the allowlist changes what the stamp means, so the two must move together. `AuthoredCorpusBenchmark` now aliases it, and the constant is reachable from the gate.
- **`BodyIntrinsicAllowlist_IsPinnedToCurrentMethodologyVersion`** asserts set equality between the live allowlist and the set pinned for the current version. Set equality rather than a subset check, so any addition, removal, or substitution fails. A version's pin is historical once stamped and may never be edited — only a new version added.
- **`BodyIntrinsicAllowlist_ExcludesContextDependentErrorClasses`** pins one representative of each forbidden category (resolution, conversion, overload, scope collision, definite assignment), so the categories the README forbids are enforced rather than only the IDs other tests happen to exercise.

## Evidence

Both directions of the contract verified by real runs:

| scenario | result |
| --- | --- |
| add `CS0136` (v2 unchanged) | **2 failed** ✓ now caught |
| add `CS0029, CS1503` (v2 unchanged) | **2 failed** ✓ now caught |
| add `CS0246, CS0103` (v2 unchanged) | **4 failed** ✓ |
| bump to `v3` with no pin recorded | **fails** with `methodologyVersion 3 has no pinned body-intrinsic allowlist` |
| untampered tree | **32/32 pass** |

`SpanAttributionTests`, `DynamicCompilationSiteInventoryTests`, `AuthoredCorpusHistoryCardTests` all green; harness builds clean; README passes markdownlint.

**No behavior change** — the allowlist is byte-identical (`{CS0128}`) and no shipped number moves. This only constrains future edits.

## Review

Low risk: test-only enforcement plus a constant relocation, no attribution logic touched and no metric affected. Proposing single-review tier.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/SpanAttributionTests.cs
+++ b/src/ILInspector.Decompiler.Tests/SpanAttributionTests.cs
@@ -292,4 +292,68 @@ public class SpanAttributionTests
 
         Assert.False(isolated);
     }
+
+    // The allowlist that each methodology version is defined by. A version's entry is
+    // historical once stamped: rows carrying that stamp were produced by exactly this set,
+    // so an entry may never be edited — only a new version added.
+    static readonly ImmutableDictionary<int, ImmutableHashSet<string>> AllowlistByMethodologyVersion =
+        ImmutableDictionary.CreateRange(
+        [
+            // v2: syntax errors in the body span, plus duplicate-local only. Every
+            // context-dependent class (resolution, conversion, overload, scope collision)
+            // is excluded because a broken shell reconstructs them identically to a real
+            // body defect — see SpanAttribution.IsolatingBodyError.
+            KeyValuePair.Create(2, ImmutableHashSet.Create(StringComparer.Ordinal, "CS0128")),
+        ]);
+
+    [Fact]
+    public void BodyIntrinsicAllowlist_IsPinnedToCurrentMethodologyVersion()
+    {
+        // The allowlist is the operative definition of productBodyDefect under the stamped
+        // methodologyVersion, but it lives in a different file from the stamp with no code
+        // path between them. Without this gate a contributor can widen the soundness rule —
+        // adding, say, CS0136, which this PR's own review identified as shell-dependent
+        // (a shell parameter collision produces it) — and ship it under an unchanged v2.
+        // The damage is not just a wrong count: rows sharing a stamp are supposed to be
+        // comparable, so a silent widening makes the history card chart a v2 -> v2 step
+        // across two different methodologies, defeating the boundary split that
+        // Render_MovementSplitsProductDefectAcrossMethodologyBoundaryWithoutCharting exists
+        // to enforce.
+        //
+        // Set equality (not a subset check) is the point: any addition, removal, or
+        // substitution fails here until the version is bumped and a new pin recorded.
+        int version = SpanAttribution.MethodologyVersion;
+
+        Assert.True(
+            AllowlistByMethodologyVersion.ContainsKey(version),
+            $"methodologyVersion {version} has no pinned body-intrinsic allowlist. Bumping the "
+                + "version requires recording the allowlist that defines it here.");
+
+        Assert.Equal(AllowlistByMethodologyVersion[version], SpanAttribution.BodyIntrinsicSemanticErrorIds);
+    }
+
+    [Fact]
+    public void BodyIntrinsicAllowlist_ExcludesContextDependentErrorClasses()
+    {
+        // The README forbids whole categories, not just the IDs the close-negative tests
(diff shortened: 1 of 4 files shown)


richlander:
## Adversarial review — MAI-Code, clean at `f9adfd79`

Single-review tier per AGENTS.md: test-only enforcement plus a constant relocation, no attribution logic touched, allowlist byte-identical, no shipped metric affected.

**No blocking findings.** The reviewer independently re-ran the tampering rather than trusting the PR's table, and extended it to two cases I hadn't covered:

| tamper (methodologyVersion left at 2) | result |
| --- | --- |
| add `CS0136` | both gates fail — `Expected: ["CS0128"] Actual: ["CS0128", "CS0136"]` + the category message |
| add `CS0029`, `CS1503` | both gates fail |
| **empty the allowlist** (removal) | pin fails `Expected: ["CS0128"] Actual: []`, plus the intrinsic-semantic behavior test |
| **replace `CS0128` with `CS0136`** (substitution) | pin fails `Expected: ["CS0128"] Actual: ["CS0136"]` |
| bump to `v3`, no pin recorded | fails with the actionable message |

So set equality holds in all three directions — addition, removal, substitution — which is what makes this a definition pin rather than a blocklist.

Also verified beyond what I claimed:

- **Relocation is behavior-preserving by measurement**, not just by inspection: a harness run against a synthetic corpus still emits `"methodologyVersion": 2`, and the only two definitions are the `SpanAttribution` constant and the `AuthoredCorpusBenchmark` alias.
- **No bypass**: `IsolatingBodyError` credits a semantic error only through `BodyIntrinsicSemanticErrorIds` (after the authored-body control), and `FirstSyntaxErrorInSpan` uses `GetDiagnostics()`, which reports parser/lexer diagnostics only — so the syntax path cannot be used to smuggle a semantic error past the allowlist.
- 32/32 focused tests pass on the restored tree; harness builds clean.

### Non-blocking observation (accepted)

The "a version's pin is historical, never edit it" rule on `AllowlistByMethodologyVersion` is prose, not mechanically enforced — a contributor could edit a past version's entry. Accepting as-is, for the reviewer's reason: historical pins are never consulted for live behavior, so the only entry that can affect a shipped number is the current one, and that is exactly what the gate enforces. Pinning history against edits would need the pins to live outside the tree (or be hashed into `history.jsonl`), which is more machinery than the risk warrants.

Worth being explicit that this PR does **not** claim to make the allowlist correct — only to make widening it visible. Soundness of `{CS0128}` itself rests on #3231's probes.

richlander:
Ready to merge

CI green (`test`, `markdownlint`, `changes`) and the required single review (MAI-Code) is clean at the exact head `f9adfd79`. No behavior change: allowlist byte-identical at `{CS0128}`, `methodologyVersion` still serializes 2, no shipped number moves.

Non-blocking follow-up recorded above (historical pins are prose-protected, accepted by design) — not a merge blocker.

PR #3273, opened 2026-07-27
Progressive member call-graph acquisition seam (#3266)

### LINKED ISSUE #3266: Progressive member call graph: serve callees first via a scoped single-body build
## Summary

The member Call graph does not show anything until the **intra-library callers** are computed, which requires a full-assembly body decode. The **callee set** of a member is a member-local fact (like allocation-safety facts) and can be produced from a *scoped single-body* build — near-instant — so it should be rendered first, before the more expensive caller tiers.

This tracks adding a scoped "callees-only" path so the call graph can render in three progressive tiers:

1. **Callees** — scoped single-body build (decode only the target method body).
2. **+ intra-library callers** — full decode of the target assembly.
3. **+ cross-library callers** — decode every other in-scope package.

## Why callees are cheap (and I initially got this wrong)

The multi-level callee **tree** (`BuildCallTree`) walks `DirectCalls` and, for depth ≥ 2, needs the callees' bodies too, so the product treats `CallGraph` as a whole-assembly build:

- `src/dotnet-inspect/Inspectors/ApiMemberAnalysisInspection.cs:45-58` — `needsWholeAssemblyBody` is set when `Callers`, `CallGraph`, or `CallerGraph` is requested; only then is `_bodyScope` left null (full build).

But the **direct callee set** is member-local and already supported by a targeted build:

- `src/dotnet-inspect/Inspectors/MethodBodyInspectionSession.cs:52-56` — `bodyScope` "restricts body decoding to the given method tokens (a single-member 'targeted' build); it is only valid when every requested section's facts are local to those members (**Calls** / Unsafe Operations / Allocation-Safety-Cost facts) — reverse/aggregate sections require a full build."
- `src/dotnet-inspect/Output/ApiOutputFormatter.cs:1267` — the `Calls` section is produced from `analysisInspection.BodyIndex.GetDirectCallsByCaller()[token]`, i.e. the target method's direct calls, under the scoped build.

So callees (direct, depth-1) are the same cost class as the per-member allocation facts: decode one body. Only callers (reverse/aggregate) require the whole-assembly decode.

## Where it shows up today

The browser prototype engine computes callers and callees in a single call, so the first paint waits for the intra-library caller reverse map:

- `prototypes/inspect-web/engine/Program.cs:1660-1663`

```csharp
var callers = index.BuildCallerTree(token, maxDepth: 2, maxNodes: 30);
if (callerScopes.Length > 0)
    callers = index.BuildCallerTree(token, callerScopes, maxDepth: 2, maxNodes: 30);
var callees = index.BuildCallTree(token, maxDepth: 2, maxNodes: 30);
```

`index` here is a full unscoped `LibraryBodyIndex.Open(..., MethodEvidence)` — the whole-assembly body decode — which is the dominant latency for large assemblies (e.g. `openai`, ~6.7k members).

The web client already stages this into two tiers (intra-library, then cross-library callers) purely on the client; see the progressive `loadSelectedMemberCallGraph` in `prototypes/inspect-web/src/app.js`. The missing, fastest tier is callees-from-a-scoped-build.

## Proposal

- Add a scoped callees-only capability the call graph can call first: open the target assembly with `bodyScope = { targetToken }`, take the direct callees (`GetDirectCallsByCaller()[token]`, or `BuildCallTree(token, maxDepth: 1)`), and render a callees-only graph.
- Then fill in intra-library callers (full target-assembly build), then cross-library callers (other packages), re-rendering in place.
- Consider whether the core product's Call graph section (not just the prototype) should adopt the same "callees first, callers progressively" acquisition, since the scoped/full split already exists in `MethodBodyInspectionSession`.

## Notes

- Depth: a single-token scoped build yields depth-1 callees (immediate calls); the callee **tree** still expands to depth-2 once the full build lands, so the first paint is direct callees and the graph grows on the next tier. This is acceptable given the prototype already re-renders progressively.
- Fields/facts on callee nodes that depend on whole-assembly signals should degrade gracefully in the scoped first tier.

richlander:
## What

Implements #3266: a reusable product acquisition seam that serves a member call graph **progressively, with no duplicated index build**, feeding either per-section tree rendering or `CallGraphMermaid` (with or without mermaid as the final target).

`ProgressiveMemberCallGraph` (`src/dotnet-inspect/Inspectors/`) serves one member''s graph in three cumulative, memoized layers, cheapest first:

1. **`Callees`** — a scoped single-body build (decode only the target body); depth-1 callees, no caller tree yet.
2. **`Callers`** — a full target-assembly decode; adds the intra-library caller tree and deepens callees to `depth`.
3. **`CrossLibrary`** — decode in-scope packages so **both** the caller and callee trees can cross a library boundary up to `depth`.

`depth` is first-class (levels up = callers, down = callees). The seam yields presentation-free `CallTreeNode` roots as a `MemberCallGraphView(Tier, CalleeRoot, CallerRoot)`; a host renders them itself or projects with `CallGraphMermaid.Render(CallerRoot, CalleeRoot)`.

Drive it by **pull** (`Callees()`/`Callers()`/`CrossLibrary()` or the lazy `Tiers()` stream) or by **push** (`RunAsync` raising `LayerReady` per layer then `Completed`). Both share one memoized core.

For full up/down symmetry this PR also adds the cross-assembly forward/callee builder `LibraryBodyIndex.BuildCallTree(token, calleeScopes, ...)` — the mirror of `BuildCallerTree(scopes)` — keyed on the same erased identity and tagging each boundary-crossing callee with its source assembly.

## No-duplication invariant

At most two target-assembly indexes are ever built (scoped single-body + full), plus one per cross-library package; each is built once and reused for callees, callers, and any Mermaid projection. A non-progressive consumer calling `Callers()`/`CrossLibrary()` directly pays exactly one full build (the scoped build is never made); callees come free. Once the full build lands it supersedes the scoped one, which is never rebuilt.

## Evidence

- **Analysis** (`ILInspector.Analysis.Tests`): 5 new tests for the cross-assembly callee builder — external-callee pull-in + source tagging, depth-2 chain across an assembly boundary, constructed-generic resolution, scope-order determinism, empty-scope fallback parity. All pass.
- **Seam** (`dotnet-inspect.Tests`, new `ProgressiveMemberCallGraphTests`, in the non-parallel `IndexBuildGuard` collection): tier-1 scoped-only build, exactly-one-full-build when `Callers()` is requested directly, full-index reuse (no scoped build) after a full build, cross-boundary callee expansion + source tagging, `Tiers()` order + build counts (3 opens), no-cross-scope stop-at-callers, `CallGraphMermaid` round-trip, and the `RunAsync` push sequence. All 8 pass. Build counts asserted via `MethodBodyInspectionSession.OpenCountForTests`.
- Full `dotnet build dotnet-inspect.slnx -c Release` clean. Full Analysis and `dotnet-inspect.Tests` suites run; the only failures (`CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition`, `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`) reproduce identically on clean `main` — pre-existing environment/ref-pack/SourceLink failures, unrelated to this change.
- Docs: `docs/design/call-graph-mermaid-projection.md` gains a "Progressive acquisition" section; passes `markdownlint`.

## Compatibility

Additive only: a new seam type, a new public `LibraryBodyIndex.BuildCallTree(scopes)` overload, one fixture method (`Entry.RunOuter`), and tests. No existing signatures or output changed. Product paths stay SRM-only / NativeAOT-friendly / Roslyn-free; the seam takes no dependency on `ILInspector.CallGraph` (only the test project references it for the projection round-trip).

Closes #3266.

### DIFF
--- a/docs/design/call-graph-mermaid-projection.md
+++ b/docs/design/call-graph-mermaid-projection.md
@@ -89,6 +89,50 @@ The projection owns everything a host must not re-invent in JavaScript:
   they additionally entity-encode the structural delimiters (`()[]{}`) that would
   otherwise corrupt an edge label.
 
+## Progressive acquisition
+
+The projection needs `CallTreeNode` roots; how a host *acquires* them is a
+separate concern (issue #3266). `dotnet-inspect` owns a
+`ProgressiveMemberCallGraph` seam
+(`src/dotnet-inspect/Inspectors/ProgressiveMemberCallGraph.cs`) that serves one
+member's graph in three cumulative layers, cheapest first, so a host can paint
+the outbound half immediately and fill in the expensive tiers as they land:
+
+1. **`Callees`** — a scoped single-body build that decodes only the selected
+   member. The callee tree is bounded at depth 1 (immediate callees); there is
+   no caller tree yet.
+2. **`Callers`** — a full decode of the member's own assembly, adding the
+   intra-library caller tree and deepening the callee tree to the configured
+   depth. Expansion stops at the assembly edge.
+3. **`CrossLibrary`** — decodes the in-scope packages so *both* the caller tree
+   and the callee tree can cross a library boundary up to `depth`.
+
+The layer names name the tier that was unlocked, not a direction: at `depth > 1`
+the `CrossLibrary` layer lets a caller chain *and* a callee chain each cross a
+package boundary. The seam yields presentation-free `CallTreeNode` roots as a
+`MemberCallGraphView` (`Tier`, `CalleeRoot`, `CallerRoot`), so a host renders
+them with its own per-section tree rendering *or* projects them with
+`CallGraphMermaid.Render(CallerRoot, CalleeRoot)` — "with or without mermaid."
+
+**No duplicated work.** At most two target-assembly indexes are ever built — the
+scoped single-body build and the full build — plus one build per cross-library
+package, and each is built once and reused for callees, callers, and any Mermaid
+projection. The scoped build exists only for the progressive first paint: a
+consumer that wants the whole graph calls `Callers()` or `CrossLibrary()`
+directly and pays exactly one full build, with callees derived for free from it.
+Once the full build lands it supersedes the scoped one, which is never rebuilt.
+The `IndexBuildGuard`-collected seam tests assert these build counts through
+`MethodBodyInspectionSession.OpenCountForTests`.
+
+Drive it by pull (`Callees()` / `Callers()` / `CrossLibrary()`, or the lazy
+`Tiers()` stream) or by push (`RunAsync` raising `LayerReady` per layer then
+`Completed`). The push path is a thin wrapper over the same memoized pull core,
+so the two never double the work. The forward cross-library expansion is the
+callee mirror of `BuildCallerTree(scopes)`: `LibraryBodyIndex.BuildCallTree(token,
+calleeScopes, …)` builds a structural forward map keyed by the same erased
+identity the caller builder uses, tagging each boundary-crossing callee with its
+source assembly.
+
 ## Consumers
 
 The browser engine is handed this exact Mermaid document and only asks Mermaid
--- a/src/ILInspector.Analysis.CallerGraphCaller/SharedEntry.cs
+++ b/src/ILInspector.Analysis.CallerGraphCaller/SharedEntry.cs
@@ -12,6 +12,15 @@ namespace Shared
         // (a package boundary), proving the forward map deepens a callee chain across assemblies.
         public static void RunOuter() => Run();
 
+        // #3266 fan-out fixture: two call sites to the same callee. The cross-assembly callee tree
+        // dedups to one Echo child but must still report a fan-out of 2 (true call-site count).
+        // Echo is used so this does not perturb the exact-count caller-graph tests rooted at Ping.
+        public static void RunTwice()
+        {
+            Target.GenericApi.Echo(1);
+            Target.GenericApi.Echo(1);
+        }
+
         // Distinct callers of the int and string Ping overloads. A caller graph rooted at one
         // overload must report only its own caller; a CallerGraphKey that drops parameter
         // types would collapse these onto Ping and cross-link them (#1623 rung 1).
--- a/src/ILInspector.Analysis.CallerGraphCaller/SharedEntry.cs
+++ b/src/ILInspector.Analysis.CallerGraphCaller/SharedEntry.cs
@@ -7,6 +7,11 @@ namespace Shared
     {
         public static void Run() => Target.Api.Ping();
 
+        // Cross-assembly callee-chain fixture (#3266). A callee graph rooted here and scoped
+        // with the target assembly must expand RunOuter -> Run (same assembly) -> Target.Api.Ping
+        // (a package boundary), proving the forward map deepens a callee chain across assemblies.
+        public static void RunOuter() => Run();
+
         // Distinct callers of the int and string Ping overloads. A caller graph rooted at one
         // overload must report only its own caller; a CallerGraphKey that drops parameter
         // types would collapse these onto Ping and cross-link them (#1623 rung 1).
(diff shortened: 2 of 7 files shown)


richlander:
## Adversarial re-review reconciliation (fixed head)

Two-model re-review of the fixed head after the first round's five fixes (commit `032fab0a`). Reviewers: **GPT** (`gpt-5.6-sol`) and **Gemini** (`gemini-3.1-pro-preview`), each isolated in its own linked worktree at the exact head, scratch under `/tmp`, read-only.

Behaviorally-reviewed head: `ae959563`. Both models independently **confirmed all five first-round fixes correct and regression-free** (failure propagation in `BuildCrossScopes`; Leaf-vs-External for full indexes; `MarkImmediateCalleesBounded` on the scoped first paint; true call-site fan-out via `rawEdges.Count`; cancellation-before-build in `RunAsync`).

Two new findings, both reconciled below. The only change since `ae959563` is the comment-only follow-up `62d72960` (zero behavioral delta).

### GPT — Medium/High-confidence: `BuildCallTree` leaf/External for body-scoped indexes
Verified **real**. In a body-scoped index, a body-bearing method that is out of `bodyScope` is added to `Methods`/`definitions` with `HasBody = true` *before* the scope early-return (`LibraryBodyIndex.cs` ~2717-2723), so it has a definition but no decoded body/edges and reads as a false `Leaf`.

- **Not reachable through the product.** `ProgressiveMemberCallGraph.CrossLibrary()` — the only product caller of the cross-assembly overload — passes only **full** indexes (`FullIndex` + `CrossScopes`, both `bodyScope: null`), where every body-bearing method is decoded, so `def != null` iff decoded and the classification is correct. The one scoped path in the seam (`Callees()` first paint) uses the single-assembly builder at `maxDepth: 1` and already compensates via `MarkImmediateCalleesBounded`.
- **Symmetric with existing code.** The pre-existing single-assembly `BuildCallTree` makes the same "indexed == decoded" assumption.
- **Action.** My first-round comment overclaimed ("only a decoded definition ... is a genuine leaf"); corrected in `62d72960` to state the full-decode assumption honestly. Deeper hardening (track decoded-body tokens and apply to both builders) is tracked as non-blocking follow-up **#3275**.

### Gemini — High: `??=` memoization not thread-safe vs. `RunAsync`
Out of the documented contract. The type doc states: *"A single instance assumes one logical consumer; do not mix concurrent pull calls with `RunAsync`."* Under that contract `RunAsync`'s tier builds are strictly sequential on one thread and never race a pull, so each index is still built at most once. There is also **no product consumer of the seam yet**. Added an in-code note at the memoization fields recording the intentional non-synchronization and how to harden (`Lazy<T>` with `ExecutionAndPublication`) if a concurrent consumer ever appears (`62d72960`). No behavior change.

### Evidence
- Analysis suite: `Total: 513, Failed: 1` — the sole failure is the pre-existing env case `CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition` (missing framework ref pack), which reproduces on clean `main`.
- Product suite: only the pre-existing env case `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden` (SourceLink) fails; it reproduces on clean `main`.
- Seam tests (10) and `LibraryBodyIndexTests` pass at head `62d72960`.

richlander:
### Reviewer sign-off on reconciliation

Both re-reviewers confirmed their new findings are **non-blocking** for this PR at head `62d72960`:

- **GPT** (`gpt-5.6-sol`): agreed non-blocking. Independently verified the only production caller of the cross-assembly overload is `CrossLibrary()`, that `FullIndex`/`CrossScopes` are full (`bodyScope: null`) builds, that scoped `Callees()` restamps direct `Leaf` children, and that the same limitation pre-exists in the single-assembly builder. Added a nuance — full builds can still have recoverable per-method decode failures, so the #3275 hardening should key on actual decode success — now captured in #3275.
- **Gemini** (`gemini-3.1-pro-preview`): agreed non-blocking. Verified the type doc's single-consumer contract and that no product consumer of the seam exists; approved the in-code note and the documented `Lazy<T>` upgrade path.

No blocking findings remain. Deeper builder hardening is tracked as non-blocking follow-up #3275.

richlander:
Ready to merge

PR #3274, opened 2026-07-27
Resolve property/event/indexer accessors for member body sections (#3265)

### LINKED ISSUE #3265: Call Graph (and method-body sections) should resolve property/event/indexer accessors
## Summary

Member-level analysis sections that require a method body — most visibly **Call Graph** — produce no result when the selected member is a **property** (and likewise events / indexers). A property has no method-body identity of its own; its executable IL lives in its **accessor methods** (`get_X` / `set_X`). The engine should resolve a property (event, indexer) to its accessor method(s) and run the call graph against those.

## Current behavior

Selecting a property and requesting the Call Graph section yields nothing. The web UI surfaces this as:

> **Call graph query failed** — The selected member has no method body identity.

Reproduced on `Microsoft.Extensions.Hosting.Abstractions` → `HostBuilderContext.Properties` (a read-only property). The same member's **Facts** and **Annotated Source** sections fail for the same reason. **Overview** and **Source** (which read the declaration, not a body) work.

## Root cause

The method-body sections operate on a single selected *method* overload, resolved in `ApiOutputFormatter.PopulateIndexSections`:

- The section builders key off `singleMethodList` / `singleMethod`, derived from the selected member's method overloads (`methods`), e.g. Call Graph at `src/dotnet-inspect/Output/ApiOutputFormatter.cs:1340`:

  ```csharp
  if (request.CallGraph && singleMethodList is [{ MetadataToken: { } graphToken } graphMethod])
  {
      var root = ToCallGraphNode(analysisInspection.BuildCallTree(graphToken), ...);
      ...
  }
  ```

- For a property, `methods` is empty (the property member carries no `MetadataToken` for a callable body), so `singleMethod` is `null`, the `is [...]` pattern never matches, and the section renders empty. The property's accessor methods (`get_Properties`, `set_Properties`) exist as distinct methods in metadata with real tokens and bodies, but nothing maps the property to them.

The same gap affects every method-body section that uses `singleMethod`: **Call Graph**, **Caller Graph**, **Calls**, **Callers**, **Exception Regions**, **Facts**, **Decompiled/Annotated Source**, IL, cost/semantics overlays.

## Proposed change

When the selected member is a property / event / indexer, resolve it to its accessor method(s) and run the method-body sections against them:

- **Property** → `get_` accessor (and `set_` / `init_` when present).
- **Event** → `add_` / `remove_` (and `raise_` if present).
- **Indexer** → `get_Item` / `set_Item`.

Open design questions:

1. **Accessor selection when both exist.** A read/write property has two bodies. Options: default to the getter, emit one graph per accessor, or expose the accessors as selectable overloads so the caller (CLI `-S`, web UI) can pick. Preferred: treat accessors as addressable overloads so the existing overload-selection path (`overloadIndex`) handles them uniformly.
2. **Surfacing accessor identity to clients.** The API member model returned by `inspect package`/`member` currently exposes no accessor tokens/selectors for a property, so a client can't request "call graph of the getter." The member projection should carry accessor method identities (token + stable selector) so downstream tools (and the web UI) can target them.
3. **Labeling.** Graph/section headers should make clear the analysis is of `get_Properties`, not the property declaration, to avoid implying the property itself has a body.

## Affected consumer (context)

The `prototype/wasm-command-ui` web UI currently **gates** the body-dependent sections (Call Graph, Facts, Annotated Source) out of the strip for `property`/`field`/`event`/`constant` members as a stopgap, showing only Overview + Source. Once the engine resolves property accessors, that client gating can be relaxed to re-enable Call Graph (etc.) for properties/events/indexers (fields legitimately have no body and stay gated).

## Acceptance

- Requesting Call Graph for a read-only property returns the getter's call tree (non-empty when the getter calls anything), not an empty/"no method body identity" result.
- Read/write properties can target each accessor.
- The member projection exposes accessor identities so a client can request a specific accessor's body sections.
- Fields (no accessor, no body) still correctly report "not applicable".

richlander:
## What & why

Fixes #3265. Member-level **body sections** — Call Graph, Caller Graph, Calls,
Callers, Exception Regions, Facts, Decompiled/Annotated/Original Source, IL, and
the cost/semantics overlays — produced nothing when the selected member was a
**property, indexer, or event**. Those members have no method-body identity of
their own; their IL lives in accessor methods (`get_`/`set_`, `add_`/`remove_`).

This resolves such a member to its accessor method(s) and runs the body sections
against them. Per the issue's preferred design (confirmed by @richlander),
accessors are **addressable overloads**: `Name:1` = getter/adder (the default),
`Name:2` = setter/remover. Graph roots and breadcrumbs use the **metadata
accessor spelling** (`get_Name`, …). Fields carry no accessor token and stay
body-less (unchanged N/A).

## Changes

- **Predicate + gating** (`ApiSectionDescriptors.cs`): new `IsBodyBacked` /
  `HasAccessorTokens`; every body-section `CanRender` gate and the single-member
  overload gate flip from `IsMethodLike` to `IsBodyBacked`.
- **Accessor resolution** (`ApiOutputFormatter.cs`): `ResolveBodyMethods` /
  `AccessorMethods` synthesize accessor method members keyed by the accessor's
  own MethodDef token; both `ApiCommand` build sites use them.
- **Overwrite fix** (`PopulateIndexSections`): `bodyMethods` restricts code and
  caller sections to the selected accessor, so a read/write property's two
  accessor bodies no longer overwrite each other. Method behavior is unchanged
  (single selected member → single method list).
- **Selector** (`MemberTargetResolver.cs`): accepts an accessor overload ordinal
  for a single property/event, defaults to getter/adder, range-checks `1..N`.
- **Projection** (`ApiSurface.cs`): event `AdderToken`/`RemoverToken` now
  serialize (like getter/setter) so JSON consumers can address event accessors.
- **Docs**: README note on accessor addressing.

## Design notes / boundaries

- Overloaded indexers (`Item:2`) keep normal overload selection and default to
  the getter; addressing a specific overload's setter is out of scope.
- The pre-existing bare-FQN auto-select limitation (`member Ns.T.M -S "Call
  Graph"` without an explicit `:N`) affects methods and properties alike and is
  unchanged here — use an explicit selector or `-D`.

## Evidence

- New resolver tests (accessor-ordinal getter default, setter-by-ordinal,
  read-only rejects `:2`, event remover) — 23/23 pass.
- New CLI call-graph tests (property getter, property setter by ordinal, event
  adder) — 12/12 pass.
- Full CLI suite: 2133 total, only the pre-existing network/ilasm-dependent
  `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`
  fails (baselined identically on `origin/main`); 10 ilasm tests skip.
- Manual CLI spot-checks across getter/setter/event/indexer/field on real
  runtime + package assemblies.

Adversarial review (two models) to follow before marking ready.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,7 +281,9 @@ A property, indexer, or event has no body of its own, so body sections
 Address them through the overload-index selector: `Name:1` is the getter/adder
 (the default) and `Name:2` the setter/remover, each rooted at its metadata
 accessor name (`get_Name`, `set_Name`, `add_Name`, `remove_Name`). Fields have
-no accessor and stay body-less.
+no accessor and stay body-less. The SourceLink source-file sections
+(`Original Source`, `Source Diff`, `Source Locations`) remain method-only for
+now; accessor source-line mapping is a follow-up.
 
 `allocation-fanout` is an opt-in aggregate view for construction-heavy
 registries, pipelines, and object graphs that do not match a local rewrite
--- a/README.md
+++ b/README.md
@@ -271,7 +271,17 @@ dotnet-inspect library MyLib.dll --where "Finding=analysis.call-site" --jsonl
 dotnet-inspect library MyLib.dll --where "CallerLoop=direct" --order-by "CallerLoopDepth desc" --jsonl
 dotnet-inspect member MyType Method:1 --library MyLib.dll -S "Call Graph,Facts"
 dotnet-inspect member MyType Method:1 --library MyLib.dll -S "Caller Graph" --fields "Throw,Catch,Finally"
+dotnet-inspect member MyType Value:1 --library MyLib.dll -S "Call Graph"
+dotnet-inspect member MyType Value:2 --library MyLib.dll -S "Call Graph"
 ```
+
+A property, indexer, or event has no body of its own, so body sections
+(`Call Graph`, `Caller Graph`, `Calls`, `Callers`, `IL`, `Decompiled Source`,
+`Facts`, and the other IL-body views) resolve it to its accessor methods.
+Address them through the overload-index selector: `Name:1` is the getter/adder
+(the default) and `Name:2` the setter/remover, each rooted at its metadata
+accessor name (`get_Name`, `set_Name`, `add_Name`, `remove_Name`). Fields have
+no accessor and stay body-less.
 
 `allocation-fanout` is an opt-in aggregate view for construction-heavy
 registries, pipelines, and object graphs that do not match a local rewrite
--- a/src/ILInspector.Metadata/ApiSurface.cs
+++ b/src/ILInspector.Metadata/ApiSurface.cs
@@ -442,11 +442,13 @@ public class ApiMember
     public int? SetterToken { get; set; }
 
     /// <summary>
-    /// MethodDef tokens of an event's add/remove accessors when known.
+    /// MethodDef tokens of an event's add/remove accessors when known. Serialized (like
+    /// <see cref="GetterToken"/>/<see cref="SetterToken"/>) so JSON consumers can address an
+    /// event's accessor bodies; omitted for members that expose no such accessor.
     /// </summary>
-    [JsonIgnore]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public int? AdderToken { get; set; }
-    [JsonIgnore]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public int? RemoverToken { get; set; }
 
     public bool IsStatic { get; set; }
--- a/src/ILInspector.Metadata/MemberTargetResolver.cs
+++ b/src/ILInspector.Metadata/MemberTargetResolver.cs
@@ -258,6 +258,7 @@ public static class MemberTargetResolver
         }
 
         MemberTargetCandidate selected;
+        int? accessorOrdinal = null;
         if (selector.DigestPrefix is { Length: > 0 } digest)
         {
             var matches = candidates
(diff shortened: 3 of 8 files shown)


richlander:
## Adversarial review reconciliation (two-model)

Reviewed by **MAI-Code** (`mai-code-1-flash-picker`) and **GPT** (`gpt-5.6-sol`), each isolated in its own fixed-head review worktree with scratch under `/tmp/` and no git mutations. Final reviewed head: `b9dc7402` (feature branch merged with `origin/main`; net contribution scoped to 8 accessor files via `git diff origin/main...HEAD`).

### MAI-Code
Clean on every pass (initial head, `c145bfcd`, and final `b9dc7402`), verified via build + resolver tests (23/23) + call-graph/decompiled tests (18/18) + real-runtime CLI repros (getter/setter/indexer/event, virtual/override/private-set).

### GPT findings and resolution

| # | Finding | Severity | Resolution |
|---|---------|----------|------------|
| 1 | Property/event source sections (Original Source, Source Diff, Source Locations) rendered success-shaped empty output | blocking | **Fixed** (`c145bfcd`) — reverted those sections to method-only gating; accessor source-line mapping is a documented follow-up. Re-verified: properties emit no source section; methods still render all three. |
| 2 | Synthetic accessor members had null signatures → malformed C# in Decompiled/Annotated Source | blocking | **Fixed** (`c145bfcd`) — `Accessor(...)` now builds a structured `ApiSignature` from the owner (getter = property type + index params; setter/adder/remover = void + trailing `value`). Re-verified valid headers. |
| 3 | `CreateBodyTarget` returned getter/adder token for ordinal 2 | non-blocking | **Fixed** (`c145bfcd`) — `SelectAccessorToken` selects the MethodDef token by ordinal (present-only). Asserted setter `0x06000102` / remover `0x06000105`. |
| 4 | Accessor headers dropped `virtual`/`override`/`sealed` and used the owner's aggregate accessibility (a `private set` rendered `public`) | blocking | **Fixed** (`8f98a88d`) — `Accessor(...)` propagates `IsVirtual`/`IsOverride`/`IsSealed` from the owner and resolves per-accessor accessibility from `SignatureModel.Accessors`. Verified: `public override string get_Message()`, `private void set_HasReproducibleFlag(bool value)`. Added virtual/override/private-set regression tests. |
| 5 | Non-public setter (e.g. `SpinWait.Count` `internal set`) renders `private` in the default view, exact only with `--all` | blocking (as filed) | **Dismissed as pre-existing** — this exactly matches how the property declaration itself renders: `public int Count { get; private set; }` by default vs `{ get; internal set; }` with `--all`. Both read the same `SignatureModel.Accessors` data; the default-view collapse of non-public setter accessibility to `private` is intentional progressive-disclosure behavior in metadata extraction, which the accessor faithfully mirrors (accessor shows `internal` under `--all`, consistent with the property). Confirmed with the maintainer that keeping the accessor consistent with its owning property is the intended behavior. |

All merge-blocking findings are resolved or dismissed with justification; targeted suites are green and the full CLI suite shows only the pre-existing ilasm/network `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden` failure (identical on `origin/main`).

richlander:
Ready to merge

Two-model adversarial review (MAI-Code + GPT) is reconciled on the final head `b9dc7402`: all blocking findings fixed and verified, finding #5 dismissed as pre-existing progressive-disclosure behavior (confirmed with the maintainer). CI is green (build, linux test, markdownlint); csharp/il-diff-smoke and pack are path-skipped as expected. Merge state is CLEAN.

Issue #3269, opened 2026-07-27
PlatformResolverTests.GetPacksDirectory_ReturnsValidPath asserts nothing when null and asserts the wrong suffix for the preferred cache branch


richlander:
Found while reviewing unrelated work. Two defects in one small test, both of the "asserts nothing / asserts the wrong thing" family.

`src/dotnet-inspect.Tests/PlatformResolverTests.cs:338-348`:

```csharp
public void GetPacksDirectory_ReturnsValidPath()
{
    var result = PlatformResolver.GetPacksDirectory();

    // Must return a packs directory (app cache or SDK-installed)
    if (result != null)
    {
        Assert.EndsWith("packs", result);
        Assert.True(Directory.Exists(result), $"Packs directory should exist: {result}");
    }
}
```

## 1. Vacuous on the null path

Every assertion is inside `if (result != null)`, with no `else` and nothing asserted outside. `GetPacksDirectory()` returns `null` whenever no candidate directory exists (`PlatformResolver.cs:96-99`). On such a machine the test passes having asserted **nothing** — while its name claims it returned a valid path. The comment directly above the branch even states the intended contract ("Must return a packs directory"), which is the assertion that is missing.

## 2. The suffix assertion is wrong for the preferred branch

`Assert.EndsWith("packs", result)` cannot be satisfied by the app-cache branch. The cache category is `packs-v2`:

```
src/DotnetInspector.Services/PlatformPackService.cs:13:    private const string PacksCategory = "packs-v2";
src/DotnetInspector.Services/PlatformPackService.cs:31:        return Path.Combine(CoreCache.GetBasePath(), PacksCategory);
```

and `GetPacksDirectoryCandidates()` searches it **first** (`PlatformResolver.cs:216-219`, comment: *"App cache packs first (always preferred)"*). So once the app cache is populated and `CoreCache` is initialized, `GetPacksDirectory()` returns a path ending in `packs-v2`, and `EndsWith("packs")` fails.

The test passes today only because that branch is not being taken. Measured on this machine:

```
GetPacksDirectory() = /home/rich/.local/share/dotnet/packs
EndsWith("packs")   = True
candidates that exist:
  /home/rich/.local/share/dotnet/packs
  /home/rich/dotnet/packs
```

[...]

Issue #3268, opened 2026-07-27
Span-attribution allowlist defines what v2 means but nothing pins it to methodologyVersion (#3231 follow-up)


richlander:
Follow-up from #3231 (merged). Non-blocking there — the allowlist is correct **today** and the shipped v2 numbers are sound. Filed so the gap doesn't rot.

## Problem

`SpanAttribution.BodyIntrinsicSemanticErrorIds` is the soundness rule that gives `methodologyVersion = 2` its meaning: it is the explicit allowlist of semantic errors that may be credited as a product body defect under a broken shell. It is currently exactly `{CS0128}`.

`tools/DecompilerHarness/corpus/evil-runs/README.md:71-74` states the rule correctly:

> Because the allowlist defines what `v2` means, expanding it requires a new `methodologyVersion`; do not add IDs under the existing stamp.

That rule is **prose only**. The allowlist appears in exactly one file repo-wide:

```
$ grep -rn "BodyIntrinsicSemanticErrorIds" --include=*.cs . | grep -v SpanAttribution.cs
(no output)
```

and `MethodologyVersion = 2` lives in a different file (`AuthoredCorpusBenchmark.cs:32`) with no code path connecting the two.

## Measured

Three tampering runs against the full `--gate fast`, `methodologyVersion` left at 2 in every one:

| allowlist addition | category | fast gate |
| --- | --- | --- |
| `CS0165` | unassigned local | **1 failed** — caught |
| `CS0246, CS0103, CS1061, CS0234` | resolution | **2 failed** — caught |
| `CS0029, CS1503, CS0136` | conversion / overload / scope collision | **3828 passed, 0 failed** — **not caught** |

So the *enumerated* never-credited IDs are genuinely pinned (`DecompiledBodyIsolated_FalseWhenDecompiledBodyHasOnlyResolutionError` and `...HasUnassignedLocalError` both do real work). But the README also forbids whole *categories* — "conversions, and overload resolution" — and those have no pin. The allowlist can be widened and shipped under an unchanged `v2` stamp with a fully green gate.

`CS0136` is the sharp end. #3231's own review reconciliation records Gemini's probe finding that **a parameter collision yields `CS0136`** — i.e. `CS0136` was identified *during that review* as shell-dependent, precisely the class condition 2 exists to exclude. Nothing stops its addition tomorrow.

## Why it matters beyond one wrong number

[...]

Issue #3267, opened 2026-07-27
IrInvariants is opt-in and default-off, so a future host silently gets no validation (#3241 follow-up)


richlander:
Follow-up from the #3241 stack (#3250 merged; #3261/#3262 in flight). Non-blocking there, filed so it isn't lost.

## Problem

`IrInvariants.Enabled` is opt-in and **default-off**. That was the right call for #3241 — it lets the check run inside the same optimized assembly users run, without a fourth build flavor. But it moves the failure mode rather than eliminating it: the check's value now depends entirely on every host that *should* enable it actually doing so, and nothing enforces that.

On the stack tip (`8e27d8be`) there is exactly **one** programmatic writer:

```
$ git grep -n "IrInvariants.Enabled = "
src/ILInspector.Decompiler.Tests/IrInvariantsTestHost.cs:17:    public static void Enable() => IrInvariants.Enabled = true;
```

Everything else depends on the `DOTNET_INSPECT_IR_INVARIANTS` environment variable being set by the operator (`IrInvariants.cs:24`).

So a new entry point — another benchmark, a new sweep tool, a future corpus harness — silently gets **no validation**, and it looks completely healthy while doing so. That is the same shape as the original #3241 defect (a check that existed and asserted nothing), just relocated from the compiler stripping the call to the caller never arming it.

## Why it matters

The #3241 stack establishes that these invariants are the pipeline's by-construction guarantee. Their per-PR teeth currently come from exactly one place — #3262's `CoreLibSweep_IsExceptionSafeAndSemanticallyValid`. Any future host that runs the pipeline without arming the check contributes zero validation while appearing to exercise the pipeline broadly, which is worse than not running it at all: it looks like coverage.

## Suggested fix

Either of:

1. A shared "validating host" entry point that arms the check, so opting *out* is the explicit act rather than opting in.
2. An assertion in each long-running host (harness, sweeps, benchmarks) that `IrInvariants.Enabled` is set, failing loudly when it is not.

Option 1 is preferable — it makes the safe path the default path. The house pattern for this is #3249/#3247: make the declaration drive the enforcement rather than restating it in prose.

## Notes

Not urgent: today's two consumers (test host, harness-via-env) do arm it, so nothing is currently unvalidated. This is about the next host, not the current ones.

PR #3276, opened 2026-07-27
Require asserted safety properties to name their gate (#3271)

### LINKED ISSUE #3271: Convention: a doc comment asserting a safety or faithfulness property must name its gate, or be marked unverified
This is the generalized, closable form of the pattern that keeps surfacing across the decompiler and harness reviews. Filing it as a convention change rather than another one-off fix.

## The pattern

The codebase is unusually good at *declaring* properties — doc comments and READMEs state safety, soundness, and faithfulness guarantees precisely and honestly. It is much weaker at *enforcing* them. The recurring failure is a comment that asserts a property with no gate behind it, which then quietly stops being true.

Concrete instances, all confirmed by measurement:

- **#3241** — `CheckInvariant` was `[Conditional("DEBUG")]` while the suite ran Release. The check existed, was documented, and asserted nothing.
- **#3247 / #3249** — byte-neutrality was declared per style knob but the enforcement set was hand-maintained alongside it, free to drift. Fixed by making the declaration *drive* the set.
- **#3268** — the span-attribution README states "expanding the allowlist requires a new `methodologyVersion`". Prose only; I widened the allowlist with three category-forbidden IDs and the full fast gate stayed green.
- **#3267** — `IrInvariants.Enabled` is documented as the pipeline's by-construction guarantee, but nothing ensures a host arms it.
- **#3269** — a test named `..._ReturnsValidPath` asserts nothing when the path is null.
- **#3230** (open) — `InlineArrayCollectionPass`'s test comment states "elements are re-sequenced into slot (= source) order" as an assumption; no gate checks it, and I have fixtures where it is false.

In every case the *statement* was accurate when written. The gap is that nothing ties its continued truth to CI.

## Proposed convention

Add to `AGENTS.md`, under the evidence and validation rules:

> A doc comment or README that asserts a safety, soundness, or faithfulness property must name the gate that enforces it, or explicitly mark the property as unverified. "Unverified" is an acceptable answer; an unmarked, ungated claim is not.

The cheap enforcement idiom, already demonstrated twice in-repo, is to **make the declaration drive the enforcement set** rather than restating it:

- `ByteNeutralityGateTests.cs:162-163` — `ByteNeutralKnobs = StyleOptionCatalog.Options.Where(o => !o.ByteDivergent)`, then `Assert.Equal(required, covered)` for set equality, so both stale and missing entries fail.
- `IrInvariantCheckTests` — one named non-vacuity test whose doc comment says it is *the* test that fails if the wiring dies.

## Why this is worth a rule rather than case-by-case fixes

The individual instances are each cheap to close, and several already have been. What makes it worth a convention is that the failure is invisible by construction: a green suite plus a confident comment reads exactly like a verified property. Reviewers cannot distinguish the two without doing what I have been doing — tampering with the code and checking whether anything notices. A rule that forces the author to name the gate (or admit there isn't one) moves that cost from the reviewer to the author, where it is a one-line answer.

## Related

Sub-theme of #3240. Also worth pairing with a note that Release is mandatory for fixture fidelity (from #3250's AGENTS.md text), so the next person hitting a stripped `[Conditional("DEBUG")]` call does not "fix" it by switching the suite to Debug.

richlander:
Closes #3271.

## What changes

`AGENTS.md` gains one rule under **Evidence and validation**:

> A doc comment or README that asserts a safety, soundness, or faithfulness property must name the gate that enforces it, or explicitly mark the property as unverified.

plus a short `### Asserted properties name their gate` subsection recording the enforcement idioms the repo already uses.

## Why a convention rather than more one-off fixes

The repository states its guarantees precisely and honestly; it enforces them unevenly. Every confirmed instance was accurate when written, and each stopped being checked without anything turning red:

| Instance | The ungated claim |
| --- | --- |
| #3241 / #3250 | `CheckInvariant` was `[Conditional("DEBUG")]` while the suite ran Release — documented, present, asserting nothing. |
| #3247 / #3249 | Byte-neutrality declared per style knob; the enforcement set was hand-maintained beside the declaration and free to drift. |
| #3268 / #3270 | The span-attribution README promised that widening the allowlist requires a new `methodologyVersion`. Prose only — three category-forbidden IDs went in with the full fast gate green. |
| #3267 | `IrInvariants.Enabled` is documented as the pipeline's by-construction guarantee; nothing ensures a host arms it. |
| #3269 | `..._ReturnsValidPath` asserts nothing when the path is null. |
| #3230 | `InlineArrayCollectionPass`'s "elements are re-sequenced into slot order" assumption has fixtures where it is false. |

The failure mode is invisible by construction: a green suite plus a confident comment is indistinguishable from a verified property. A reviewer can only tell them apart by tampering with the code and watching whether anything notices. Requiring the author to name the gate — or say there isn't one — moves that cost to where it is a one-line answer.

## What the subsection records

Both idioms are already demonstrated in-repo, so the rule documents practice rather than inventing it:

- **Declaration drives the enforcement set**, asserted by set equality so stale *and* missing entries fail: `ByteNeutralityGateTests` derives coverage from `StyleOptionCatalog.Options.Where(o => !o.ByteDivergent)`; `SpanAttributionTests.BodyIntrinsicAllowlist_IsPinnedToCurrentMethodologyVersion` pins the allowlist to `MethodologyVersion`.
- **One named non-vacuity test** when the property is wiring-dependent rather than set-shaped: `IrInvariantCheckTests.PipelineRunner_UnderTestHost_ThrowsWhenAPassCorruptsTheTree`, whose doc comment already says it is the only test that fails if the module initializer, the per-pass gate, or the runner's call disappears.

It also pairs the rule with the existing Release requirement, per #3271's closing note: a gate only counts in the configuration the suite runs, so the next person meeting a stripped `[Conditional("DEBUG")]` check makes it a runtime opt-in the test host arms instead of switching the suite to Debug and losing fixture fidelity.

## Boundary

Documentation only. No product, test, or harness code changes; the open instances above stay tracked on their own issues and are not closed by this PR. The rule is prospective — it is not a mandate to sweep every existing comment.

## Validation

Documentation-only change making no measured behavior claim, so per the evidence rules:

```
npx markdownlint-cli AGENTS.md   # clean
```

Citations in the new text were verified against the tree at `ab35977b`: `src/ILInspector.Decompiler.Tests/ByteNeutralityGateTests.cs:162`, `tools/DecompilerHarness/SpanAttribution.cs:78-82`, `src/ILInspector.Decompiler.Tests/IrInvariantCheckTests.cs:252`, and the `## Building and testing` anchor.

Copilot-Session: 03e294d5-cb8d-417c-9297-676292be6ba2

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -125,6 +125,38 @@ Match evidence to the claim and use the smallest existing check that proves it:
   frequency, bytes, or impact; use a benchmark or profiler for runtime claims.
 - Documentation-only changes that make no measured behavior claim require
   Markdown validation, not product builds or tests.
+- A doc comment or README that asserts a safety, soundness, or faithfulness
+  property must name the gate that enforces it, or explicitly mark the
+  property as unverified.
+
+### Asserted properties name their gate
+
+"Unverified" is an acceptable answer; an unmarked, ungated claim is not. A
+green suite plus a confident comment reads exactly like a verified property,
+and a reviewer can only tell them apart by tampering with the code to see
+whether anything notices. Naming the gate moves that cost to the author, where
+it is a one-line answer.
+
+Prefer making the declaration *drive* the enforcement set over restating it, so
+that stale and missing entries both fail:
+
+- `ByteNeutralityGateTests` derives its coverage set from the style catalog
+  (`StyleOptionCatalog.Options.Where(o => !o.ByteDivergent)`) and asserts set
+  equality against the specimens.
+- `SpanAttributionTests` asserts set equality between the body-intrinsic error
+  allowlist and the pin for the current `MethodologyVersion`.
+
+When the property depends on wiring rather than on a set, write one named
+non-vacuity test that fails if the wiring dies, and say in its doc comment that
+it is that test —
+`IrInvariantCheckTests.PipelineRunner_UnderTestHost_ThrowsWhenAPassCorruptsTheTree`
+is the example.
+
+A gate only counts if it runs in the configuration the suite uses. The suite
+runs Release for fixture fidelity (see [Building and
+testing](#building-and-testing)), so a `[Conditional("DEBUG")]` check asserts
+nothing. Make such a check a runtime opt-in that the test host arms; do not
+switch the suite to Debug.
 
 ### Harness boundary
 


richlander:
## Review reconciliation

**Adversarial review tier:** more than trivial, not high risk — documentation-only convention change, no product, test, or harness code touched, no measured behavior claim. One review, MAI-Code, per the roster rule.

**MAI-Code — clean, no blocking findings.** Reviewed on an isolated read-only worktree at the exact head `6ed8795b` against base `ab35977b`. It independently verified:

- Every symbol the new text names resolves in the tree and behaves as described: `ByteNeutralityGateTests.cs:162-163` (declaration-driven set) and `:218` (set-equality assertion); `SpanAttributionTests.cs:309-332` with `SpanAttribution.cs:78-88` (allowlist pinned to `MethodologyVersion`, casing correct); `IrInvariantCheckTests.cs:252-260` (the named non-vacuity test).
- `[Building and testing](#building-and-testing)` targets the real heading at `AGENTS.md:191` and the line-broken link text parses as one link.
- No contradiction or harmful redundancy with the existing Evidence-and-validation bullets, Harness boundary, or the Release / `[Conditional("DEBUG")]` / `IrInvariants` discussion.
- Diff confined to `AGENTS.md`.
- `npx --yes markdownlint-cli AGENTS.md` clean from the review worktree.

Nothing to dismiss or fix, so there are no resolution commits; the reviewed head is the head above.

**CI:** `markdownlint` and `changes` pass; `build-net10`, `test`, `pack`, and both diff-smoke jobs are path-gated and correctly skipped for a Markdown-only diff.

**Non-blocking follow-up (not part of this PR):** the rule is prospective. The open instances it was distilled from — #3230, #3267, #3269, and the tracking issue #3240 — stay on their own issues and are deliberately not closed here.

Ready to merge

PR #3285, opened 2026-07-27
Drop dead inline-array buffer local from fidelity (#3221)

### LINKED ISSUE #3221: Raised inline-array collection expressions leave a dead unspellable buffer local that caps fidelity at Partial on older-runtime assemblies

richlander:
- Fixes #3221, advances #3129
- Changes: a proven-dead `<>y__InlineArrayN` buffer local left behind by the S4 inline-array raise no longer degrades method fidelity to Partial
- Card revision: `73fc912d`

## Change

**Conclusion:** **PASS** — after S4 raises an address-spilled inline-array collection, its buffer local is dead but retained for slot-index stability; marking that slot eliminated stops a false DEC0009 without touching any raise decision or rendered byte.

### Benchmark target

Benchmark target: `Microsoft.CodeAnalysis@5.0.0` (`Microsoft.CodeAnalysis.dll`, net9.0)

dotnet-inspect command:

```bash
dotnet-inspect member Microsoft.CodeAnalysis.Emit.EncLocalInfo GetDebuggerDisplay --library Microsoft.CodeAnalysis.dll --all -S "Decompiled Source"
```

### Original source

```csharp
private string GetDebuggerDisplay()
{
    if (IsDefault) return "[default]";
    if (IsUnused) return "[invalid]";
    return string.Format("[Id={0}, SynthesizedKind={1}, Type={2}, Constraints={3}, Sig={4}]",
        SlotInfo.Id, SlotInfo.SynthesizedKind, Type, Constraints,
        Signature != null ? BitConverter.ToString(Signature) : "null");
}
```

### Before

```csharp
private string GetDebuggerDisplay()
{
    if (IsDefault)
    {
        return "[default]";
    }
    if (IsUnused)
    {
        return "[invalid]";
    }
    return string.Format("[Id={0}, SynthesizedKind={1}, Type={2}, Constraints={3}, Sig={4}]", [SlotInfo.Id, SlotInfo.SynthesizedKind, Type, Constraints, Signature is not null ? BitConverter.ToString(Signature) : "null"]);
}
```

- Valid: True
- Correct: True
- IL fidelity: Partial — `fidelity: Partial`, `DEC0009 (signature) type name '<>y__InlineArray5' has no C# spelling`
- Taste applied: None
- Commit: `ddd3a44b`

The S4 raise already produces the correct collection expression here, but the now-dead `<>y__InlineArray5` buffer local is retained in `Locals` (indices are slot-keyed and cannot be removed mid-array). Its unspellable name trips DEC0009 and degrades the method to Partial even though the buffer never appears in the rendered C#.

### After

```csharp
private string GetDebuggerDisplay()
{
    if (IsDefault)
    {
        return "[default]";
    }
    if (IsUnused)
    {
        return "[invalid]";
    }
    return string.Format("[Id={0}, SynthesizedKind={1}, Type={2}, Constraints={3}, Sig={4}]", [SlotInfo.Id, SlotInfo.SynthesizedKind, Type, Constraints, Signature is not null ? BitConverter.ToString(Signature) : "null"]);
}
```

- Valid: True
- Correct: True
- IL fidelity: Full — `fidelity: Full`, no remarks
- Taste applied: None
- Commit: `73fc912d`

The rendered C# is byte-identical to Before. The only change is fidelity accounting: the dead buffer slot is marked eliminated, so DEC0009 no longer fires and the method reports Full.

### Fully raised

## How it works

`InlineArrayCollectionPass` raises `M(<>y__InlineArrayN buffer ...)` to a C# collection expression by detaching the buffer's init, its N element stores, and the single span use — leaving the buffer local with zero references. The buffer cannot be dropped from `function.Locals` (slot indices are keyed positions), so instead the pass calls `IrFunction.MarkLocalEliminated(slot)`; `CSharpSpellability.FunctionRenderedTypes` skips eliminated slots, and the printer already omits unreferenced locals.

Two reconstruction seams re-home MoveNext locals and lost the marking; both are carried:

- `IteratorReconstructionPass.Transplant` re-homes MoveNext via `ResetLocals`, which now takes an optional `eliminatedSlots` set (same slot numbering). Classic-async keeps the clearing default (it rebuilds a fresh local table and never re-adds the buffer).
- `MultiYieldReconstruction` appends MoveNext temps onto the kickoff at `localOffset = kickoff.Locals.Length`, so after the append it re-marks `localOffset + slot` for each eliminated MoveNext slot.

All other reconstruction paths (CountingLoop, YieldBreakLoop, Switch, Linear, Stateless) build synthetic bodies and drop the dead buffer entirely, so no marking is needed.

## Evidence

| Check | Baseline (`ddd3a44b`) | Head (`73fc912d`) |
| --- | --- | --- |
| Focused tests | n/a (tests added here) | `InlineArraySpilledElementTests` + `IteratorReconstructionPassTests` + `DynamicCompilationSiteInventoryTests`: 81 passed |
| Decompiler fast suite | 3798, 1 failed | 3803, 1 failed |
| Real witness | `EncLocalInfo::GetDebuggerDisplay` Partial (DEC0009 `<>y__InlineArray5`) | `EncLocalInfo::GetDebuggerDisplay` Full |

The single fast-suite failure is pre-existing on both baseline and head: `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString` (CRLF issue #3018), unchanged by this PR.

This is a fidelity-only change: it never alters a raise decision, never changes a rendered byte, and can only flip an already-dead-buffer method from Partial to Full. It cannot introduce lowering residue, conditional-branch residue, or pass bugs, so the corpus quick gate is unaffected by construction.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class "ILInspector.Decompiler.Tests.InlineArraySpilledElementTests" -class "ILInspector.Decompiler.Tests.IteratorReconstructionPassTests" -class "ILInspector.Decompiler.Tests.DynamicCompilationSiteInventoryTests"
```


richlander:

Two adversarial reviewers, both non-Claude and from distinct families, reviewed the exact fixed head `73fc912d` in isolated read-only worktrees with required real-run evidence.

| Reviewer | Model | Head reviewed | Verdict |
| --- | --- | --- | --- |
| Reviewer A | `gemini-3.1-pro-preview` | `73fc912d` | **CLEAN** |
| Reviewer B | `gpt-5.6-sol` | `73fc912d` | **CLEAN** |

Both independently reproduced the real-run evidence on `73fc912d`: full-solution build clean, the 3 focused classes (`InlineArraySpilledElementTests`, `IteratorReconstructionPassTests`, `DynamicCompilationSiteInventoryTests`) **81 passed / 0 failed**, and the real witness `EncLocalInfo::GetDebuggerDisplay` → `fidelity: Full` with no `<>y__InlineArray` leak.

### Attack points, both reviewers concurring

1. **S4-guard interaction** — `MarkLocalEliminated(local)` runs only at the bottom of the rewrite block, after substitution and all detaches; every early bail skips it, so a referenced/partially-raised slot is never marked.
2. **Other reconstruction paths** — exhaustive grep confirms only `IteratorReconstructionPass.Transplant` and `MultiYieldReconstruction` copy the MoveNext local table; all synthetic paths (CountingLoop, YieldBreakLoop, Foreach, Reducible, Switch, Linear, Stateless) discard the dead buffer entirely and need no marking.
3. **MultiYield index translation** — locals are appended via `AddLocal` in-order before marking, so `localOffset + slot` maps exactly onto the transplanted range; hoisted locals are added afterward, so no collision.
4. **False Full / masking** — impossible: `FunctionRenderedTypes` only suppresses the metadata `Locals` type view; the fidelity visitor still inspects every `LoadLocal`/`StoreLocal`'s `DirectTypes`, so a still-referenced unspellable local re-forces DEC0009 and stays Partial.
5. **ResetLocals carry/clear** — transplant carries `EliminatedLocalSlots` (identical numbering); classic-async intentionally omits the param and clears, so stale markings can never jump an index reset.
6. **DEC0009 over-skipping** — skipping is by slot index, not by type; a second live local sharing the unspellable type still degrades fidelity (proved by `MarkLocalEliminated_SkipsOnlyTheMarkedSlot`).

### Prior blocking findings, carried into this head

Earlier review rounds on the pre-restack branch raised two blocking findings, both already resolved and preserved through the re-stack onto the final S4 head:

- **Transplant lost the marking** (Reviewer A + B, prior round) → fixed in `32c854c2`: `ResetLocals` gained an optional `eliminatedSlots` carry and `Transplant` passes `work.EliminatedLocalSlots`.
- **MultiYield lost the marking** (Reviewer B, prior round) → fixed in `73fc912d`: after appending MoveNext temps, the pass re-marks `localOffset + slot` for each eliminated MoveNext slot.

Both reviewers re-verified these fixes on the current head and found no residual gap.

### Re-stack note

The 3 commits were re-stacked (`git rebase --onto ddd3a44b`) onto the final S4 head `ddd3a44b` and force-pushed while not under public review. The rebase was conflict-free; the `MarkLocalEliminated` call site and S4's broadened `EvaluatesChildUnconditionally` whitelist were both verified intact post-rebase.

Both fixed-head reviews are clean; no open findings.

richlander:
## Merge sequencing (stacked PR)

This PR is stacked on top of #3230 (S4), so its base is `raise/3129-inlinearray-params-span`, not `main`. `ci.yml` only triggers `pull_request` on `branches: [main]`, so CI intentionally does not run against this base.

Both fixed-head adversarial reviews are clean and full local validation passes (build, focused classes 81/81, fast suite 3803 with only the pre-existing CRLF #3018 failure, witness `EncLocalInfo::GetDebuggerDisplay` → Full). The only remaining merge-blocking item is CI on the real merge path, which cannot run until the base is `main`.

Sequence to merge:

1. Merge #3230 (S4) into `main`.
2. Retarget this PR to `main` (rebase `--onto origin/main` if the squash-merge changes the base commit) so CI validates the true merge path.
3. Once CI is green, `Ready to merge` will be posted here.

Holding `Ready to merge` until CI completes on the `main`-based merge path.

richlander:
Reviewed at fixed head `73fc912d` (on top of #3230 @ `ddd3a44b`) in a clean worktree. The fidelity goal is right and the slot-remapping across the iterator transplant checks out. **One change requested** — measured, cheap to fix.

## `MarkLocalEliminated` asserts deadness; nothing checks it

`EliminatedLocalSlots` has exactly one consumer:

```
$ grep -rn "EliminatedLocalSlots" --include=*.cs src/ | grep -v Tests
IrNodes.cs:351:  public IReadOnlySet<int> EliminatedLocalSlots => ...
IteratorReconstructionPass.cs:176: function.ResetLocals(..., work.EliminatedLocalSlots);
MultiYieldReconstruction.cs:74:    foreach (var slot in moveNext.EliminatedLocalSlots)
CSharpSpellability.cs:230:         if (function.EliminatedLocalSlots.Contains(slot)) continue;
```

The printer never reads it. So the justification in both doc comments — *"renders nowhere (the printer skips it)"*, *"the printer emits no declaration for it"* — is not a coupling; it holds only as long as the pass really did consume every reference. That correctness condition is asserted, not verified, and `RenderedTypes` is precisely where an unchecked assertion turns into a wrong fidelity number.

The pass tallies references two ways (`InlineArrayCollectionPass.cs:64-68`): bail if any `LoadLocal` names the slot, and require `LoadLocalAddress` count `== count + 2`. A **by-value `StoreLocal` to the buffer slot** is in neither set — it is not an address ref, so it does not disturb the tally, and it is not a load, so the bail does not see it.

Measured with a fixture in the shape of `BuildOrderingCollection`, adding `local0 = b;` after the span consumer:

```
[PROBE] raised=True  slot0Eliminated=True  byValueStoreSurvives=True
```

The raise fires, slot 0 is marked eliminated so `<>y__InlineArray2<object>` stops counting against fidelity, and a reference to that local is still in the tree. The method's output still involves a local whose unspellable synthesized type is no longer being counted — a fidelity claim of Full over output that isn't. That is the success-shaped failure `AGENTS.md` calls out, and it lands in the one PR whose entire purpose is making this number accurate.

To be clear about the blast radius: the **raise itself stays sound**. A by-value store between the init and the fill is already rejected by the contiguity check, and one after the span read doesn't disturb element order. This is a fidelity-accounting defect, not a wrong-code defect.

Reachability is the same class as my original #3230 ordering finding: csc never emits a by-value store to an inline-array buffer, so this needs ilasm or an obfuscator. It matters for the same reason — the decompiler consumes arbitrary IL, and this PR's own value proposition is that the fidelity number can be trusted.

## Suggested fix

Two options, and I'd push for the second:

1. Add `StoreLocal` naming the slot to the bail alongside the `LoadLocal` check.
2. Make `MarkLocalEliminated` **verify** instead of assert: after the rewrite, confirm no `LoadLocal` / `LoadLocalAddress` / `StoreLocal` for that slot survives in the function, and only then mark. One `Descendants` walk per raise.

Option 2 is better for the reason this PR's parent already argues better than I could. `EvaluatesChildUnconditionally` is an allow list specifically because *"any conditional or repeatedly evaluated node not enumerated — and any such node added to the IR later — would silently fall through"*. Local liveness here is the mirror image: enumerating the reference-carrying node kinds is a deny list, and a new local-referencing node added to the IR later falls through the same way, silently upgrading fidelity. Applying your own default-deny argument to the liveness check makes it future-proof, and it converts the doc comment's claim into something enforced rather than believed (cf. #3271).

## Otherwise verified

- `ResetLocals(..., eliminatedSlots)` correctly drops the marking on renumbering, and `MultiYieldReconstruction.cs:74` remaps to `localOffset + slot`, matching the `AddLocal` loop directly above it. Consistent.
- `MarkLocalEliminated` range-checks its argument.
- `InlineArraySpilledElementTests` + `IteratorReconstructionPassTests`: 77/77.
- Fast decompiler suite at this head: **3803 passed, 0 failed**.

richlander:
Filed my finding as #3295 so it's tracked regardless of how you want to sequence it.

To be explicit about my recommendation: I still think it belongs **in this PR** rather than as a follow-up. `MarkLocalEliminated` is a new public API on `IrFunction`, so every future caller inherits the unverified "this slot is dead" contract — verifying once at introduction is much cheaper than retrofitting once several passes depend on it. And this PR is what converts an honest `Partial` into a wrong `Full` for that IL, so fixing it here keeps the PR's own claim true.

If you'd rather defer, the minimum bar should be making the doc comment honest — drop "renders nowhere (the printer skips it)" for an explicit note that deadness is assumed and unverified (#3271). The thing worth avoiding is an unmarked, ungated claim in the exact place fidelity is computed.

Also filed #3296 for the pre-existing allow list / deny list terminology on `main`, separate from the 24 occurrences this stack adds.

richlander:
## Adversarial re-review reconciliation — head `b128fab5`

Both fixed-head reviews are now **CLEAN** (Gemini `gemini-3.1-pro-preview`, GPT `gpt-5.6-sol`).

### Finding history on `MarkLocalEliminated` (root cause: it kept re-implementing the printer's liveness logic and drifting)

| Round | Head | Finding | Resolution |
| --- | --- | --- | --- |
| 1 | — | Asserted deadness instead of verifying it | Switched to a verified walk |
| 2 | — | Walked `this.Descendants` (counted references in **isolated** nested pools → false Partial) + too-narrow predicate | Narrowed scope, widened predicate |
| 3 | `fa85fd90` | GPT: `DescendantsOutsideNestedFunctions` **over**-skips shared-scope nested functions → false Full (mirror image of round 2). Gemini+GPT: predicate missed switch-expression arm carriers | This commit |
| 4 | `b128fab5` | GPT raised a follow-on scope concern; **reconciled to CLEAN** (see below) | No change needed |

### The convergent fix (`b128fab5`)

Stop re-implementing; **reuse the printer's authoritative pieces** so the two cannot drift:

1. **Scope** — `MarkLocalEliminated` now verifies via `LocalSlotReferencedInScope`, which structurally mirrors `CSharpPrinter.ReferencesLocalIncludingSharedNestedScopes`: it descends into shared-scope nested functions (own no locals, touch no stack slot → share the enclosing pool) and stops at own-pool ones. The `NeedsNestedLambdaScope` / `NeedsNestedLocalFunctionScope` discriminators are promoted to `internal static` and reused directly.
2. **Completeness** — `NodeBindsLocalSlot` now covers the switch-expression arm carriers (`UnionSwitchExpressionArm`, `PatternSwitchExpressionArm` incl. its `PropertySubpattern`) and guards `Fixed` with `!LocalIsStackSlot`.
3. **Drift guard** — new `MarkLocalEliminated_CoversEveryLocalSlotBindingNodeKind` reflects over the IR node types and fails the build if a future `LocalIndex`/`LocalIndices`/`VariableIndex` carrier is added without being accounted for.

### Round-4 GPT finding — verified and dismissed with real-run evidence

GPT observed that `NeedsNested*Scope` uses `Body.Descendants` (a deep walk that crosses into deeper nested functions), so a shared-scope outer lambda containing an inner own-pool lambda with stack slots is classified own-pool, skipping the outer slot reference — and concluded false Full.

I reproduced GPT's **exact** scenario and printed it. It is **not** a false Full:

- **Store variant** (GPT's precise scenario: function `Locals=[<>y__InlineArray2\`1<object>]`; outer lambda owns no locals, stores slot 0, returns an inner lambda that uses a stack slot) renders:

  ```csharp
  return () =>
  {
      object V_0 = 7;
      return () => { object S_0 = 9; };
  };
  ```

  Slot 0 is eliminated and fidelity is Full — but the output renders `object V_0`, a fully spellable type. The unspellable buffer type appears **nowhere**, so Full is honest.
- **Read variant** produces `null` print output — a visible print failure, not a Full-over-buffer.

**Mechanism** (`CSharpPrinter.cs:1628-1655`): the printer renders a body classified own-pool by constructing a **new `IrFunction` with the nested function's own `Locals`**, so a reference inside it is typed from the reference node, and the enclosing function's `Locals[slot]` is structurally never consulted. Eliminating the enclosing slot therefore cannot surface its type through such a reference; the deadness invariant ("eliminated ⟹ renders nowhere") holds. The misclassification is also one-way (deep `Descendants` only over-detects own-pool, never descends into a genuinely-isolated pool). GPT confirmed CLEAN after reproducing this.

### Evidence (reproduced by both reviewers on `b128fab5`)

- Full build clean; focused `InlineArraySpilledElementTests` **26/26**.
- **Non-vacuity, both elements**: reverting the traversal fails exactly the two shared-scope tests; disabling the switch-arm cases fails exactly the switch-arm test; restored → 26/26.
- Witness `Microsoft.CodeAnalysis.Emit.EncLocalInfo::GetDebuggerDisplay` still **Full**.
- Fast suite 3810, sole failure the pre-existing CRLF issue (#3018) — not a regression.

### Remaining sequencing before `Ready to merge`

This PR's base is the S4 branch (#3230), so CI does not run here yet. Before readiness: merge S4 → retarget/rebase this branch onto `main` → CI green. Then I'll post `Ready to merge`.

richlander:
Re-reviewed at head `b128fab5` in a clean worktree. **The core of #3295 is genuinely fixed** — deadness is now verified rather than asserted, and the scope question that churned through rounds 2–4 is settled correctly. **One change requested**, and it is about the new drift guard rather than the fix itself.

## What I confirmed positively

- **The scope walk is genuinely shared, not merely parallel.** `LocalSlotReferencedInScope` and `CSharpPrinter.ReferencesLocalIncludingSharedNestedScopes` are structurally identical and both call the same `NeedsNestedLambdaScope` / `NeedsNestedLocalFunctionScope` discriminators, now `internal static`. Promoting the discriminators rather than copying them is what makes the round-2/round-3 oscillation unrepeatable — the two cannot classify a scope differently because there is only one classifier.
- **`NodeBindsLocalSlot` is a strict superset of the printer's predicate.** The printer's `IsLocalReference` is the `{LoadLocal, StoreLocal, LoadLocalAddress}` trio; the verifier covers 13 kinds. Superset is the conservative direction — it refuses to eliminate more often than the printer refuses to render, so the error mode is a pessimistic Partial, not a false Full.
- **The round-4 dismissal holds, and for the stated reason.** `AppendNestedLocalFunctionBody` (`CSharpPrinter.cs:1628-1655`) does construct a fresh `IrFunction` from the nested function's own `Locals`, so an own-pool body never types a reference from the enclosing `Locals[slot]`. Worth stating explicitly for the next reader: correctness here depends on the printer and the verifier making the *same* classification **including the same over-detection**. That is sound precisely because they share the discriminator — it would not be sound if either kept its own copy.

Two structural attacks of my own that **failed**, recorded so nobody re-runs them:

- I expected `PropertySubpattern` to be a nesting hole, since it is not an `IrNode` (the guard needs `|| type == typeof(PropertySubpattern)` to see it) and therefore cannot be reached by the `Children` walk — only by the explicit one-level `patternArm.Subpattern?.LocalIndex`. It is not a hole: the record is single-level by construction, with no nested subpattern member.
- I expected orphan `DeconstructionTarget.Local` nodes, since `DeconstructionTarget` carries a `LocalIndex` with no case in `NodeBindsLocalSlot`. Also not a hole: targets are only ever constructed in `DeconstructionAssignmentPass` and fed into a `DeconstructionAssignment`, whose `LocalIndices` is derived from its own children.

## Change requested — the drift guard does not check what its doc says it checks

The doc on `NodeBindsLocalSlot` states the guard test *"fails when a `LocalIndex`/`LocalIndices`/`VariableIndex` carrier is added to the IR without being accounted for here."* It does not check "accounted for here." It compares a reflection inventory against a hardcoded name list and never exercises the switch.

**Measured — I deleted all 10 non-trio cases from `NodeBindsLocalSlot`:**

```
MarkLocalEliminated_CoversEveryLocalSlotBindingNodeKind  →  1 passed / 0 failed
```

The guard is green with the entire mechanism it guards gutted.

That matters because of what a developer does when the tripwire *does* fire. I added a new carrier node to the IR:

```csharp
public sealed class ProbeCarrier : IrExpression
{
    public ProbeCarrier(int localIndex) { LocalIndex = localIndex; }
    public int LocalIndex { get; }
    ...
}
```

| Step | Action | Guard | Slot 0 |
| --- | --- | --- | --- |
| 1 | Add `ProbeCarrier` | **FAIL** — `Collections differ at index 8, Actual: "ProbeCarrier"` | — |
| 2 | Add `nameof(ProbeCarrier)` to `expected`, switch untouched | **PASS** | **eliminated while still bound → false Full** |

Step 2 is the natural reading of the failure, and it is *precedented inside the test itself*: `DeconstructionTarget` and `PropertySubpattern` both sit in `expected` with no case in `NodeBindsLocalSlot` — legitimately, since they are covered through parent aggregation. So a contributor sees two entries that are inventory-only, adds a third, gets green, and has reintroduced exactly the bug class this PR exists to close.

Also worth noting: the reflection is name-based (`LocalIndex`/`LocalIndices`/`VariableIndex`) and `GetProperties`-only, so a carrier named `SlotIndex` or exposed as a public field is invisible to it. Read from the code, not run — the two items above are the ones I measured.

## Second finding — 8 of the 13 kinds have no test at all

The PR's non-vacuity evidence is real but narrower than it reads. `MarkLocalEliminated_RefusesWhenSlotBoundByNonLoadStoreNode` pins `NullCoalescingAssignment`; `..._RefusesWhenSlotBoundBySwitchExpressionArmPattern` pins `UnionSwitchExpressionArm`. That is 2 kinds.

I deleted the other 8 — `ForeachStatement`, `UsingStatement`, `Fixed`, `IsPattern`, `RecursivePropertyDeclarationPattern`, `PatternSwitchExpressionArm` (both the arm and its `Subpattern` leg), `CatchClause`, `DeconstructionAssignment`:

| Gate | Result with 8 cases deleted |
| --- | --- |
| `--gate fast` | **3810 passed / 0 failed** |
| `CorpusSweepGateTests` | **1 passed / 0 failed** |

Nothing notices. The one I would single out is `PatternSwitchExpressionArm` including its `Subpattern` leg — that is the exact gap two reviewers found in round 3, and it shipped with no test, so the same miss would recur silently.

To be clear about severity: **I am not claiming a live defect.** `MarkLocalEliminated` is called with an `<>y__InlineArrayN` buffer slot, and csc never binds such a slot as a `foreach`/`catch`/`using` variable — the green corpus sweep is consistent with that. The point is that the safety net added *because* a narrow reference tally missed a node kind is itself unenforced for 8 of 13 kinds, so the next narrowing is invisible. Given the two existing tests are ~12 synthetic lines each, adding the remaining 8 is mechanical.

## Suggested resolution

1. Make the guard assert behavior, not inventory — for each carrier kind, build the one-node function and assert `MarkLocalEliminated` refuses. That collapses both findings into one test and removes the inventory-only escape.
2. If you prefer to keep it as an inventory tripwire, correct the doc to say what it does, and mark the two intentionally case-less entries inline so "add to the list" is visibly not the default fix.

Happy for this to be a follow-up rather than in-PR given the stack is already deep — say the word and I will file it. What I would not do is leave the doc claiming enforcement the test does not perform, since that is the same "declared but unenforced" shape as #3271.

Sequencing note unchanged: base is still #3230, so CI has not run this merge path.

### Evidence

Head `b128fab5`, `dotnet build dotnet-inspect.slnx -c Release` clean. Baseline `InlineArraySpilledElementTests` 26/26; restored positive control 26/26; worktree verified clean after every tamper. Baseline `--gate fast` on this head measured **3810 / 0** for me — I did not reproduce the pre-existing CRLF #3018 failure, which is presumably line-ending-dependent in the checkout.

richlander:
Filed the drift-guard finding as #3321 so it's tracked independently of how you sequence it.

To be explicit about my recommendation: the doc correction at `IrNodes.cs:417` is the part I'd want in **this** PR, since it's a one-line change and the current text claims enforcement the test doesn't perform. The 8 missing behavioral tests are fine as a follow-up — they're mechanical, and the stack is already deep enough that I'd rather not grow it here.

Nothing else blocking from me. The core #3295 fix is right and the shared-discriminator approach is the durable version of it.

richlander:
Addressed the two drift-guard change requests in-PR (follow-up `e848c592`, on top of `b128fab5`). Product logic is unchanged — this is test + doc only.

**Finding 1 — the guard was an inventory check, not behavioral.** `MarkLocalEliminated_CoversEveryLocalSlotBindingNodeKind` compared a reflection inventory against a hardcoded name list and never ran `NodeBindsLocalSlot`; deleting the switch cases left it green, and the "natural fix" (add a `nameof`) silenced it without handling the carrier.

**Finding 2 — only 2 of 13 carrier kinds were pinned behaviorally.**

**Resolution (your preferred option):** replaced the inventory canary with a behavioral guard.

- A per-kind `IrFunction` factory binds the unspellable buffer at slot 0 **through each carrier**, and a `[Theory]` asserts `MarkLocalEliminated(0)` refuses to drop it (`DoesNotContain(0, EliminatedLocalSlots)` + `HasUnrepresentableMetadataName`). This runs the switch, so deleting **any** case fails that kind's row. The two aggregation-only kinds (`DeconstructionTarget`, `PropertySubpattern`) are bound through their owning parent (`DeconstructionAssignment` / `PatternSwitchExpressionArm.Subpattern`), matching how `NodeBindsLocalSlot` reaches them.
- A completeness tripwire asserts the factory set equals the reflection-discovered carrier set. A new carrier fails this until a **factory** is added — and because the factory binds slot 0 through that carrier, the Theory then forces the carrier to be handled in `NodeBindsLocalSlot`. A bare list entry can no longer silence the guard. Discovery now also scans public fields (closing the `GetProperties`-only half of your note). The name-convention dependence (`LocalIndex`/`LocalIndices`/`VariableIndex`) is documented inline — a carrier that spells its index differently would still slip through, by design of a name-based tripwire.
- Corrected the `NodeBindsLocalSlot` doc comment to describe the behavioral enforcement.

**Non-vacuity, both directions:**

- Gut the `ForeachStatement` case (`=> false`) → exactly `MarkLocalEliminated_RefusesEveryLocalSlotBindingNodeKind(carrierKind: "ForeachStatement")` fails (1 failure).
- Remove one factory entry (`Fixed`) → the Theory drops that row and **only** `MarkLocalEliminated_HasABehavioralCaseForEveryLocalSlotCarrierKind` fails ("Collections differ at index 3").
- Restored → focused class 38/38 green.

Fast suite 3822 total, only the pre-existing CRLF #3018 failure (`StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString`, line-ending-dependent, unrelated).

Since the product code (the `MarkLocalEliminated` / `LocalSlotReferencedInScope` / `NodeBindsLocalSlot` logic already reviewed clean by both models at `b128fab5`) is untouched, I treated this as low-risk test+doc work rather than re-running a full two-model product review. Happy to run a fresh pass if you'd prefer.

richlander:
## Adversarial re-review reconciliation — drift-guard hardening (#3321)

Two-model adversarial review (Gemini Pro `gemini-3.1-pro-preview` + GPT `gpt-5.6-sol`, per the AGENTS.md roster; I'm Opus so I used the two other families). All reviews were run in isolated detached worktrees at the exact head with real build + test evidence. The guard is **test-only**; product logic (`NodeBindsLocalSlot`/`MarkLocalEliminated`) is unchanged across this arc.

### Findings and resolutions

**Head `e848c592`** (behavioral guard) — GPT CLEAN; **Gemini BLOCKING (2 findings)**:
- **Finding 2 — vacuous factory reuse:** the refusal Theory never verified the factory it invoked actually built a node of the keyed kind, so a new carrier mapped to an *existing* kind's factory could pass both tests while unhandled in `NodeBindsLocalSlot`.
- **Finding 1 — discovery blind spot:** completeness discovery hardcoded `type == typeof(PropertySubpattern)`, so a future non-`IrNode` carrier record would be silently ignored.

  → **Fixed in `6b25db82`:** the Theory now asserts `FunctionContainsCarrier(function, carrierKind)` before eliminating slot 0 (each factory is a single-carrier `BufferBoundBy(...)`, so a reused/wrong factory fails containment); discovery is two-pronged — `IrNode` subclasses with a carrier member **plus** non-`IrNode` types with a carrier member embedded as a public property or field of some `IrNode`. A free-floating non-`IrNode` record that no `IrNode` references is correctly out of scope (it cannot bind a slot in the tree).

**Head `6b25db82`** — GPT CLEAN; Gemini confirmed all three prior fixes solid but **BLOCKING (1 new finding)**:
- **Discovery blind spot for deep generics / arrays:** `EmbeddedCandidateTypes` unrolled only one generic level and ignored array element types, so a carrier embedded as `List<List<X>>` or `X[]` would silently bypass the guard (proven: both stayed green).

  → **Fixed in `8842087f`:** `EmbeddedCandidateTypes` now recurses through generic arguments and handles `HasElementType` element types. Recursion is bounded by the finite structural type tree (generic args + element types only, never member graphs), so it terminates and pulls in no BCL types.

### Fixed-head verdict — `8842087f`: both reviews CLEAN

- **GPT — CLEAN:** build + `Total: 38, Failed: 0`; manually probed `List<List<Int32[]>>`/`Int32*` expanding and terminating; no extra carriers; all 12 factories + both `PatternSwitchExpressionArm` legs + both `DeconstructionAssignment` rows still tripwire.
- **Gemini — CLEAN ("airtight"):** re-injected `DeepCarrier` (`List<List<T>>`) and `ArrayCarrier` (`T[]`) → completeness now **fails** discovering them (escape closed); termination proven (structural type graph is a finite DAG); no false positives (38/0); the non-`IrNode` member-graph boundary is the correct scope (recursing member graphs would pull in the reachable BCL and falsely match any framework type with a `VariableIndex` property); and both earlier fixes re-verified — the vacuous-reuse tripwire fires with the exact message, and gutting the embedded scan drops `PropertySubpattern`, proving it is genuinely discovered via the `PatternSwitchExpressionArm.Subpattern` path.

### Non-actions (documented residual, both reviewers agree non-blocking)
- A deliberately adversarial factory that builds the carrier **and** a decoy slot-0 binder could game containment, but that requires intentionally violating the single-carrier factory structure and is beyond the "natural fix" escape #3321 targets.
- A carrier hidden inside a non-`IrNode` *wrapper record* that is itself a member of an `IrNode` is intentionally out of scope; adopting it would require corresponding traversal/containment semantics and would risk pulling in the reachable BCL graph.

Resolution commits: `6b25db82`, `8842087f`. Both fixed-head reviews are clean.

richlander:
## Pre-review — head `8842087f`

**#3321 is fully addressed, and the fix is stronger than what I asked for.** No blocking findings. One optional hardening below.

I re-ran the tampering rather than reading the diff.

### The guard is now behavioral, and non-vacuous per kind

`MarkLocalEliminated_RefusesEveryLocalSlotBindingNodeKind` is a `Theory` with one row per carrier, each building a method whose only local is the unspellable buffer bound by that kind. Deleting all 10 non-trio cases from `NodeBindsLocalSlot`:

```
Total: 38, Failed: 14
  → all 12 Theory rows + both original Fact tests
```

Every kind now has teeth. That is up from 2 of 13 at `b128fab5`.

**Attribution is exact, which is the part I care about most.** Deleting *only* the `patternArm.Subpattern?.LocalIndex == index` leg — the round-3 gap that previously shipped with no test:

```
Total: 38, Failed: 1
  MarkLocalEliminated_RefusesEveryLocalSlotBindingNodeKind(carrierKind: "PropertySubpattern") [FAIL]
```

One row, the right row. A future narrowing now names itself instead of going quiet.

### The completeness tripwire no longer has the inventory escape

Comparing `discovered` against the **factory dictionary keys** rather than a hardcoded name list is the right structural change: satisfying it now requires adding a factory, and the `Theory` then runs that factory through the switch. The `FunctionContainsCarrier` anti-vacuity check closes the "reuse another kind's factory" shortcut. The `expected`-list escape I demonstrated in #3321 is gone.

### The embedded-carrier discovery is real, not decorative

This went past what #3321 asked (I only flagged fields and naming as read-from-code, unmeasured). I checked it is non-vacuous — added a non-`IrNode` record carrying `LocalIndex`, embedded as `ImmutableArray<ProbeEmbedded>` on an `IrNode`:

```
MarkLocalEliminated_HasABehavioralCaseForEveryLocalSlotCarrierKind [FAIL]
  Expected: "ProbeEmbedded"
```

Discovered through the generic argument, as designed. `GetFields` coverage and the recursion through element/generic types mean the `PropertySubpattern`-shaped hole cannot recur in a collection or array form. Documenting the residual name-based limitation inline (*"keep to it"*) is the right call — that is a convention the guard cannot enforce, and saying so beats implying otherwise.

## Optional hardening — the "sole slot-0 binder" claim is argued, not asserted

The tripwire comment says the `Theory` asserts *"the factory actually builds a node of that kind **and that it is the sole slot-0 binder**"*. The first half is asserted (`FunctionContainsCarrier`). The second is not — it is true of today's factories by construction, since each is a single-carrier `BufferBoundBy(...)`, but nothing enforces it.

Measured. I added an unhandled carrier and a factory holding it **plus** a second, already-handled binder:

```csharp
[typeof(ProbeCarrier)] = () => BufferBoundByTwo(new ProbeCarrier(0), new StoreLocal(0, Object, BoxInt(1))),
```

```
Total: 39, Failed: 0
```

Both guards green with `ProbeCarrier` absent from `NodeBindsLocalSlot` — the `StoreLocal` carries the refusal.

I want to be clear this is **much weaker than the #3321 finding and I am not blocking on it.** There, the *natural* fix silenced the guard; here the natural fix (`BufferBoundBy(new NewCarrier(0))`) forces the switch to handle the kind, and the escape needs someone to deliberately write a two-binder factory. It is a comment-accuracy issue plus cheap available hardening.

If you want it enforced rather than argued, the cheap version is a per-row positive control: parameterize each factory by slot index, then assert both directions — `factory(0)` refuses elimination, and `factory(1)` **does** eliminate slot 0. Any second slot-0 binder makes the `factory(1)` leg refuse, failing the row. That converts "sole binder" from a construction argument into an assertion, and the factories already hardcode `0` so it is a mechanical edit.

## Sequencing — the reported conflict looks stale

GitHub reports `mergeable=CONFLICTING / DIRTY`, but that does not reproduce:

- Base branch `raise/3129-inlinearray-params-span` on origin is `bcb9e48e`, identical to #3230's head.
- Merging `bcb9e48e` into `8842087f` locally: **clean**, no conflicted paths, auto-merging `InlineArraySpilledElementTests.cs` and `InlineArrayCollectionPass.cs`.
- The merged tree **builds clean** and runs `--gate fast` at **3822 / 0**.

So the combined stack is healthy; the flag is most likely stale mergeability on GitHub's side. The branch is behind its base, so updating from #3230 should clear it — worth confirming the flag goes away rather than assuming, since a real conflict here would land in the two files this PR just rewrote.

### Evidence

Head `8842087f`, isolated worktree, `dotnet build dotnet-inspect.slnx -c Release` clean before each run. Baseline `InlineArraySpilledElementTests` **38/38** (was 26 — the 12 new rows). `--gate fast` **3822 / 0** both at the head and on the merged-with-base tree. Worktree verified clean after every tamper; positive control restored.

Nothing blocking from me. The sequencing note stands: CI has still not run this merge path while the base is #3230.

richlander:
## Pre-review — head `36d931ec` (stack re-check)

**Nothing blocking. The stack is in the best shape it has been in.** The new commit is purely the merge of the rebased S4 base plus current `main`, and I verified it is content-preserving rather than assuming so.

### The merge did not disturb what I verified

- **`IrNodes.cs` is byte-identical** to `8842087f`. `git diff 8842087f..36d931ec` on that file is empty, so the entire `MarkLocalEliminated` / `NodeBindsLocalSlot` / `LocalSlotReferencedInScope` mechanism is exactly what I tamper-tested.
- **`InlineArraySpilledElementTests.cs` changed 16 lines, all comment-only** — `whitelist`→`allow list`, `blacklist`→`deny list`, arriving from the S4 base commit `550e91ee`. No test logic moved.
- **The `MarkLocalEliminated` call site is unchanged** at `InlineArrayCollectionPass.cs:112`.

**Teeth re-confirmed on the merged head**, not carried over from the last review. Deleting only the `patternArm.Subpattern?.LocalIndex == index` leg:

```
Total: 38, Failed: 1
  MarkLocalEliminated_RefusesEveryLocalSlotBindingNodeKind(carrierKind: "PropertySubpattern") [FAIL]
```

Still one row, still the right row.

### The S4 base did not change substantively either

`InlineArrayCollectionPass.cs` between `bcb9e48e` (what I tested against) and `550e91ee` has **zero code-line changes** — comments only. So the rebase moved S4 onto current `main` without altering the behavior my earlier merged-tree run exercised. #3230 is now `MERGEABLE/CLEAN` onto `main`.

### The conflict flag was stale, as expected

Both PRs now report `MERGEABLE/CLEAN`. That matches what I measured last round — the base tips were already identical and the merge was clean locally, so the `DIRTY` flag was GitHub's stale mergeability rather than a real collision. Good that it cleared without anyone having to resolve anything in the two files this PR rewrote.

### Terminology (#3296) is fully clean in this stack

```
$ grep -rniE "whitelist|blacklist" --include=*.cs --include=*.md .
AGENTS.md:113:- Write "allow list" instead of "whitelist".
AGENTS.md:114:- Write "deny list" instead of "blacklist".
```

Only the guidance lines that define the rule. No residue.

### Evidence on the merged head `36d931ec`

| Check | Result |
| --- | --- |
| `InlineArraySpilledElementTests` | **38 / 0** |
| `--gate fast` | **3939 / 0** |
| `src/dotnet-inspect.Tests` | **2188 / 0** (12 skipped) |

Worktree verified clean after the tamper; positive control restored.

### Still open, still non-blocking

The optional hardening from my last pass is unaddressed: the tripwire comment states the Theory asserts the carrier is the *sole* slot-0 binder, which remains true by construction but unasserted — a two-binder factory still silences both guards. Entirely fine as a follow-up or a `wontfix`; I raise it once more only so the decision is explicit rather than accidental. If you'd rather not touch it, trimming the comment's "and that it is the sole slot-0 binder" clause would make the text match the enforcement at zero risk.

Sequencing is unchanged and now unblocked: merge #3230, retarget this to `main`, let CI validate the real merge path.

Issue #3278, opened 2026-07-27
Accessor body sections: decide whether Original Source / Source Diff should follow property/event accessors


richlander:
Follow-up nit from review of #3274 (\`feature/3265-property-accessor-body-sections\`).

#3274 taught the body-section gates to resolve a property/event (including an indexer) to its accessor methods via \`IsBodyBacked\`/\`HasAccessorTokens\`, so Decompiled Source, Annotated Source, Facts, Cost/Semantics overlays, IL, etc. now work on a selected property.

But two sections were intentionally left on the older \`HasSingleMethodLikeMember\` predicate and so do **not** extend to accessors:

- \`ApiMemberSectionDescriptors.OriginalSource\`
- \`ApiMemberDetailSectionDescriptors.SourceDiff\`

(See \`src/dotnet-inspect/Sections/ApiSectionDescriptors.cs\` — the \`CreatePipeline()\` registrations still pass \`HasSingleMethodLikeMember\` for those two while their neighbors switched to \`HasSingleBodyBackedMember\`.)

**Question to resolve:** is this intentional, or an oversight?

- Plausible intentional reason: original-source recovery depends on PDB sequence points, and compiler-generated accessor bodies (auto-property \`get_/set_\`, field-like event \`add_/remove_\`) often have no meaningful sequence points or map to the declaration line, so "Original Source" for them would be empty or misleading. In that case, add a short comment on the two registrations explaining why they deliberately stay method-like-only.
- If it is an oversight, switch them to \`HasSingleBodyBackedMember\` for consistency and add fixture coverage for a hand-written (non-auto) property accessor that does carry sequence points.

Low priority; no correctness impact — just a consistency/UX gap between the body sections.

richlander:
Resolved by #3287.

**It was an oversight, not a deliberate restriction.** The issue offered a plausible intentional reason — that compiler-generated accessor bodies often have no meaningful sequence points, or map to the declaration line, so `Original Source` would be empty or misleading. A PDB sequence-point probe says otherwise:

- hand-written accessors: 116 with sequence points, 73 without
- auto-property accessors: 1354 with, 83 without

Auto-property accessors do map to the property's declaration line — but that is the *correct* answer for them, not a misleading one. So the fix was to wire accessor identity through, not to document a restriction.

[...]

PR #3287, opened 2026-07-27
Resolve accessor source for property/event source-file sections (#3278)

richlander:
Answers #3278, and resolves it as an oversight rather than an intentional restriction.

## The question

#3278 asked whether `Original Source` and `Source Diff` were deliberately left on `HasSingleMethodLikeMember` while their neighbors moved to `HasSingleBodyBackedMember` in #3274, offering a plausible intentional reason: compiler-generated accessor bodies "often have no meaningful sequence points or map to the declaration line", so the section "would be empty or misleading".

## The answer: not borne out; it was an unfinished wire

I probed the PDB directly (SRM, every property accessor in `ILInspector.Metadata.dll`, classified auto vs hand-written by `CompilerGeneratedAttribute`):

| accessor kind | with sequence points | without |
| --- | --- | --- |
| hand-written | 116 | 73 |
| auto-property | 1354 | 83 |

Accessors do carry usable sequence points. Auto-property accessors map to the property's own declaration line -- which is the correct answer for "where is this property written", not a misleading one.

The real reason those three gates stayed behind is what I scoped out of #3274: source resolution addressed the selected member by *its own* name and MethodDef token. A property/event has neither (`ApiMember.MetadataToken` is null), so it resolved nothing. I held the sections back rather than let them render empty. This PR finishes the wire.

## What changed

- **`MemberCommand`** resolves a selected property/event to the accessor its ordinal addresses -- `1` = getter/adder, `2` = setter/remover, the same addressing the body sections use -- and passes that accessor's metadata name and MethodDef token to `ResolveMethodSourceAsync`. `PdbContext.ResolveMethodSource` short-circuits on a non-zero token straight to a MethodDef lookup, so this bypasses name matching entirely; the existing token-trust guard (facade / ref-vs-impl mismatch forces the token to 0) is preserved, and the accessor case uses overload index 0 for the name fallback because accessor names are unique in their type.
- **`MemberSourceLocationCollector`** maps the accessor token back onto the **owning** member rather than emitting synthetic accessor members, so a property contributes one row carrying its own signature (`public int MaxDepth { get; set; }`), not one row per accessor. Deterministic: getter/adder first, else setter/remover.
- **Gates** -- `OriginalSource` and `SourceDiff` registrations move to `HasSingleBodyBackedMember`; the `CanRender` predicates for all three move from `IsMethodLike` to `IsBodyBacked`; the now-unused `HasSingleMethodLikeMember` helper and its #3265 comment are removed. `ApiOutputFormatter.PopulateMemberSourceLocations` had the same method-only filter at render time and would otherwise have dropped the row it had just resolved.
- **Extractor fix (`SourceLinkResolver.ExtractMethodBody`)** -- the backward signature scan now stops when the first sequence point already lands on the member's declaration line. Every auto-property and expression-bodied member has that shape, and scanning back from it skipped the blank separator or opening brace above and captured **the preceding member or the enclosing type header** as the selected member's source. Pre-existing (it also hit one-line methods), but every property goes through it, so leaving it would have produced exactly the misleading output #3278 worried about.

## Scope note

The issue names only `Original Source` and `Source Diff`. #3274 reverted `Source Locations` too, and it shares the same root cause, so I included all three. Flagging it since it is slightly beyond the literal ask.

## Evidence

Real runs against a shipping assembly with real SourceLink (`System.Text.Json`, hand-written read/write `JsonSerializerOptions.MaxDepth`):

- `MaxDepth -S "Source Locations"` -> `public int MaxDepth { get; set; }` | `JsonSerializerOptions.cs` | 580
- `MaxDepth:1 -S "Original Source"` -> renders `get => _maxDepth;`, and **not** the setter body
- `MaxDepth:2 -S "Original Source"` -> renders the full validating setter through `_maxDepth = value;`
- `MaxDepth:2 -S "Source Diff"` -> authored property source vs the decompiled `set_MaxDepth` body

Failure stays visible: an accessor without sequence points resolves nothing and the section is omitted, exactly as a method without sequence points behaves today -- no success-shaped empty output. Fields have no accessor token, remain not body-backed, and are unchanged.

Tests: 3 new CLI tests (accessor source location, getter/setter ordinals resolving separately, accessor source diff) and 4 new `ExtractMethodBodyTests` (auto-property excludes preceding member; expression-bodied first-in-type excludes the type header; block-bodied accessor still walks back to the property declaration; recursive first statement still captures its signature).

## Validation

- `dotnet build dotnet-inspect.slnx -c Release` -- clean
- `tests/ILInspector.Metadata.Tests` -- 859/859
- `src/DotnetInspector.Services.Tests` -- 319/319
- `src/dotnet-inspect.Tests` -- 2182 total, 1 failure: `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`, which fails identically on a clean `origin/main` worktree (verified). A second failure varied per run (`DiffCommandTests...FindingTransitionsException`, then `CliDiscoverySections_AreSelectable`); both pass in isolation -- parallel/network flake, not this change.

README updated: the "SourceLink source-file sections remain method-only for now" caveat from #3274 is replaced with the accessor behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,8 +282,10 @@ Address them through the overload-index selector: `Name:1` is the getter/adder
 (the default) and `Name:2` the setter/remover, each rooted at its metadata
 accessor name (`get_Name`, `set_Name`, `add_Name`, `remove_Name`). Fields have
 no accessor and stay body-less. The SourceLink source-file sections
-(`Original Source`, `Source Diff`, `Source Locations`) remain method-only for
-now; accessor source-line mapping is a follow-up.
+(`Original Source`, `Source Diff`, `Source Locations`) follow the same
+addressing: they resolve through the accessor's PDB sequence points, so an
+auto-property maps to its declaration line and a hand-written accessor maps to
+its own authored lines.
 
 `allocation-fanout` is an opt-in aggregate view for construction-heavy
 registries, pipelines, and object graphs that do not match a local rewrite
--- a/src/ILInspector.Metadata/SourceLinkResolver.cs
+++ b/src/ILInspector.Metadata/SourceLinkResolver.cs
@@ -276,22 +276,31 @@ public class SourceLinkResolver
     /// implicitly private one, which <see cref="IsMemberSignatureLine"/> cannot recognize.
     /// <para>
     /// The line must read as a declaration prefix: a run of type-shaped tokens that reaches the
-    /// member's own name, followed by <c>(</c>, <c>&lt;</c>, <c>{</c>, or <c>=&gt;</c>. Requiring
-    /// the name to follow a return type — and rejecting a leading statement keyword — is what
-    /// separates <c>int Target =&gt; 1;</c> from a body line that merely spells the name, such as
-    /// a recursive <c>return Target(n - 1);</c> or an assignment <c>x = Target();</c>. A trailing
-    /// <c>;</c> is deliberately not accepted, so a local declaration such as <c>Foo Target;</c>
-    /// does not qualify.
+    /// member's own name, followed by <c>(</c>, <c>&lt;</c>, <c>{</c>, or <c>=&gt;</c>. Two
+    /// requirements separate a declaration from a body line that merely spells the name. A
+    /// leading statement keyword rejects the line outright, which covers a recursive
+    /// <c>return Target(n - 1);</c>. And the name must open a new token — a return type has to
+    /// precede it — rather than continue a dotted chain, which is what tells the explicit
+    /// implementation <c>int IDefault.Target =&gt; 1;</c> from the qualified call
+    /// <c>Helper.Target();</c>. A trailing <c>;</c> is deliberately not accepted, so a local
+    /// declaration such as <c>Foo Target;</c> does not qualify.
+    /// </para>
+    /// <para>
+    /// An indexer is spelled <c>this[...]</c> rather than by the <c>Item</c> name its accessors
+    /// carry in metadata, so a property accessor also accepts <c>this</c> in the name position.
     /// </para>
     /// </summary>
     private static bool DeclaresMember(string trimmed, string methodName)
     {
-        var name = SourceSpelledMemberName(methodName);
+        var name = SourceSpelledMemberName(methodName, out bool isPropertyAccessor);
         if (name.Length == 0)
             return false;
 
         int i = 0;
-        bool isFirstToken = true;
+        int tokenIndex = 0;
+        int chainStart = 0;
+        bool afterDot = false;
+
         while (i < trimmed.Length)
         {
             while (i < trimmed.Length && char.IsWhiteSpace(trimmed[i]))
@@ -308,29 +317,44 @@ public class SourceLinkResolver
                     i++;
                 var token = trimmed[tokenStart..i];
 
-                if (isFirstToken)
-                {
-                    if (StatementOpeners.Contains(token))
-                        return false;
-                    isFirstToken = false;
-                    continue;
-                }
+                // A dotted continuation stays part of the chain its first token opened; anything
+                // else opens a new one. Only a chain that something precedes can be a member name.
+                if (!afterDot)
+                    chainStart = tokenIndex;
+                afterDot = false;
 
-                if (token != name)
-                    continue;
-
-                int j = i;
-                while (j < trimmed.Length && char.IsWhiteSpace(trimmed[j]))
-                    j++;
-                if (j >= trimmed.Length)
+                if (tokenIndex == 0 && StatementOpeners.Contains(token))
                     return false;
-                if (trimmed[j] is '(' or '<' or '{')
-                    return true;
-                return trimmed[j] == '=' && j + 1 < trimmed.Length && trimmed[j + 1] == '>';
+
+                bool isNamePosition = token == name
+                    || (isPropertyAccessor && token == "this");
+                if (isNamePosition)
+                    return chainStart >= 1 && FollowsDeclarationName(trimmed, i, token == "this");
+
+                tokenIndex++;
+                continue;
             }
 
-            // Punctuation that can appear within a qualified, generic, array, or nullable type.
-            if (c is '.' or '<' or '>' or ',' or '[' or ']' or '?')
+            if (c == '.')
+            {
+                afterDot = true;
+                i++;
+                continue;
+            }
+
+            // A generic argument list or array rank belongs to the type token it follows, so it
+            // must be consumed whole — otherwise a type argument would read as a separate token
+            // and let a qualified call such as Foo<T>.Target() pass as a declaration.
+            if (c is '<' or '[')
+            {
+                int after = SkipBalanced(trimmed, i, c, c == '<' ? '>' : ']');
+                if (after < 0)
+                    return false;
+                i = after;
+                continue;
+            }
+
+            if (c == '?')
             {
                 i++;
                 continue;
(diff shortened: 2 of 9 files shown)


richlander:

This change is high risk: the accessor wiring itself is small, but it exposes
`SourceLinkResolver.ExtractMethodBody` — a best-effort text heuristic shared by
**every** member, not just accessors — to member shapes it never saw before. So
it took the two-model tier, and in practice five rounds with **GPT
(`gpt-5.6-sol`)** and **MAI (`mai-code-1-flash-picker`)**, each in its own
detached review worktree at a fixed head.

Every finding below was reproduced before being acted on, and each fix is locked
in by a regression test plus close negative cases in
`tests/ILInspector.Metadata.Tests/ExtractMethodBodyTests.cs`.

### Round 1 — head `60d0073b` → fixed in `44c374a9`

| --- | --- | --- |
| MAI (High) | `unsafe int Add()` over-captured; `IsMemberSignatureLine` knew only `public`/`private`/`protected`/`internal`/`static`. | Fixed. Modifier set completed. |
| GPT (Medium) | Modifier matching used a non-token-aware `StartsWith`, so `internalState = 1;` read as a declaration; modifierless interface members over-captured. | Fixed. Whole-token modifier matching, shared with the backward scan's break condition so the two cannot diverge; added name-anchored `DeclaresMember` for modifierless members. |
| GPT (Medium) | `SourceTokens` queried only the getter, so a `#line hidden` getter discarded an otherwise valid setter location. | Fixed. `SourceTokens` now yields every accessor in preference order, and the best-ranked *resolving* token wins across both the bulk and per-token fallback paths. |
| GPT (Medium) | Bodyless accessors produce a heading with no content and exit 0. | **Non-action — pre-existing, not introduced here.** See below. |

### Round 2 — head `44c374a9` → fixed in `bc277f6e`

| --- | --- | --- |
| GPT (Medium) | `Helper.Target();` misclassified as a declaration — the dotted-token run was unrestricted. | Fixed. `DeclaresMember` now tracks dotted chains and requires the name to *open* a chain that something precedes, which separates `int IDefault.Target => 1;` from a qualified call. |
| GPT + MAI (Medium, found independently) | Modifierless **indexers** unrecognized: metadata says `get_Item`, source says `this[...]`. | Fixed. `SourceSpelledMemberName` reports `isPropertyAccessor` so `this` is accepted in the name position. |

### Round 3 — head `bc277f6e` → fixed in `039f3391`

Two of these were regressions I introduced in earlier rounds.

| --- | --- | --- |
| GPT (Medium) | A return type spelled like its member — `CancellationToken CancellationToken => default;` — aborted detection, because the scan decided on the *first* matching token. | Fixed. A failed name match keeps scanning. |
| GPT (Medium) | The round-1 `endsAtDeclaration` guard dropped the closing brace when a block opened on the declaration line (`public int Target() {`). | Fixed. |
| MAI (Medium) | Modifierless tuple returns `(int X, int Y) Target => (1, 2);` unrecognized. | Fixed. A leading `(...)` is consumed whole, but only at token index 0 — the tuple return position. |

### Round 4 — head `039f3391` → fixed in `6c16d300`

Both reviewers hit `HasUnclosedBlock` from **opposite directions**, and one fix
resolved both.

| --- | --- | --- |
| MAI (Medium) | A `{` inside a string literal made a closed range look open, so the forward scan swallowed the enclosing type's brace. | Fixed. `HasUnclosedBlock` now skips line and block comments, char literals, regular strings with backslash escapes, and verbatim strings; it returns `true` conservatively on a raw string `"""`, restoring the old unconditional forward scan for a shape it cannot reason about. |
| GPT (Medium) | A `}` inside a string literal made an open block look closed, dropping the member's own brace. | Same fix. |
| MAI (Medium) | `ref`- and `new`-led modifierless declarations rejected as statement openers. | Fixed. `ref`/`new` are skipped without consuming a token position. |

Neither brace finding was a regression — `origin/main` extracted the same wrong
source in all four cases, having no declaration-line detection at all — but both
produce visibly wrong output, so they were fixed rather than documented.

### Round 5 — head `6c16d300` → fixed in `2c4b0733`

| --- | --- | --- |
| MAI | Clean. Verified both round-4 repros fixed; additionally exercised `http://x`, `'}'`, an escaped quote, `$@"..."`, `@$"..."`, a multi-line block comment, an unterminated string, `new Widget().Configure();` and `ref var slot = ref _items[0];`. Nothing threw or hung. | — |
| GPT (Medium) | `HasUnclosedBlock` recognized `$@"` only *incidentally* — the leading `$` fell through to no case and the following `@"` matched. The equally valid `@$"` spelling was unrecognized, so a brace on a continuation line counted as structural. | Fixed in `2c4b0733`: `@$"` matched explicitly alongside `@"`. Pre-existing on `main`, but it produces visibly wrong extracted source, so fixed rather than documented. |

Both reviewers then re-verified the exact head `2c4b0733` and are clean. GPT
confirmed its own repro plus eight variants, including the previously working
`$@"` path, a line ending in a bare `@`, `@$` not followed by a quote, the
verbatim identifier `@class`, and a same-line `@$"..."` followed by a real brace
that must still count. MAI independently confirmed the same head: `@$"` and
`$@"` both accepted a multi-line string whose continuation line contained
`{`/`{{` without counting that brace as structural, a bare trailing `@` and a
`@$` not followed by a quote neither threw nor hung, and a same-line
`@$"..."` followed by a real `{` still counted that brace.

### Recorded non-action

**GPT round 1, finding 3 — a bodyless member renders a section heading with no
content and exits 0.** This is pre-existing and out of scope here. Proof, on a
path this PR does not touch at all:

```text
dotnet-inspect member "System.Collections.IEnumerator" --framework runtime MoveNext -S "Original Source"
```

An abstract interface *method* produces the identical heading-only, exit-0
output. Properties now simply behave consistently with methods. Worth its own
issue; not a reason to hold this one.

Two further follow-ups worth filing, neither introduced here:

- `IsSelfTerminatingLine` does not strip trailing comments, so a declaration
  ending in `// ... {` still runs the forward scan.
- The bodyless-member empty-section behavior above.

### Validation at `2c4b0733`

| Suite | Result |
| --- | --- |
| `tests/ILInspector.Metadata.Tests` | 900 / 900 |
| `src/DotnetInspector.Services.Tests` | 319 / 319 |
| `src/dotnet-inspect.Tests` | 2187, with only the pre-existing `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden` failure — confirmed failing on clean `origin/main` |

Every reviewer reproduction from every round was re-verified against the built
binary, not reasoned about.

### One open question for the reviewer of record

#3278 named only `Original Source` and `Source Diff`. I also moved
`Source Locations` to the same gate, because it shares the root cause and
leaving it behind would have made the accessor case render two of three body
sections. Say the word if you would rather that be split out.

richlander:
Ready to merge

CI is green at `2c4b0733`, and both required fixed-head reviews are clean at that exact head — GPT (`gpt-5.6-sol`) and MAI (`mai-code-1-flash-picker`), each re-verifying the `@$"` fix against the built binary rather than reasoning about it. Full five-round reconciliation is in the comment above.

The two non-actions recorded there are now filed as follow-ups, neither introduced by this PR:

- #3299 — bodyless members render an empty `Original Source` section and exit 0
- #3300 — `IsSelfTerminatingLine` does not strip trailing comments, triggering an unnecessary forward scan

One question is still open and is not blocking: #3278 named only `Original Source` and `Source Diff`, and I also moved `Source Locations` to the same gate because it shares the root cause. Happy to split that out if you would rather keep this to the letter of the issue.

PR #3289, opened 2026-07-27
Arm IR invariants by default so a new host cannot silently skip validation

### LINKED ISSUE #3267: IrInvariants is opt-in and default-off, so a future host silently gets no validation (#3241 follow-up)

richlander:
Closes #3267.

## What changed

IR invariant validation is now **on by default**, so any host that runs the decompiler pipeline — test suite, harness, sweep tool, benchmark, a project that does not exist yet — validates the IR after every pass without doing anything. Declining validation is now the explicit act, and it has exactly one sanctioned form: `IrInvariants.DisableForShippedTool()` in `src/dotnet-inspect/Program.cs`.

#3241 made the check run in Release. Making it opt-in then relocated the failure mode rather than removing it: a new host that never armed the flag would exercise the pipeline broadly while validating nothing, and look healthy doing it. #3267 asked for the shared validating-host default; this is it.

## Why this shape

The issue's option 1 was a "validating host" helper that hosts call. That still requires every future host to remember. Inverting the library default removes the requirement instead of restating it — validation is what a host gets for free.

The shipped CLI is the one host that declines, because its users are not developing the pipeline, so the tool pays nothing on the decompile hot path. An explicit `DOTNET_INSPECT_IR_INVARIANTS` value (trimmed, case-insensitive; `1/true/full/on/yes` and `0/false/off/no`) outranks the opt-out **in both directions**, so the shipped tool can be armed for debugging without a rebuild.

Enforcement is layered so the weaker layer never carries the claim alone:

- **The compiler owns what it can express.** `Enabled` and `CheckSemantics` both have private setters, so `IrInvariants.Enabled = false` does not compile from any host — under a namespace alias, under `using static`, or otherwise. `CheckSemantics` gained `EnableSemanticChecks()`, whose only direction is up. This removes the spelling rather than widening a matcher to chase it.
- **A source census pins the one form that remains.** `IrInvariantsHostContractTests` scans every `.cs` file under the repository root and asserts the opt-out is referenced from exactly one file. It matches by identifier rather than by call shape, across every preprocessor configuration.

## Compatibility

Behavior-neutral for users of the shipped CLI: output is byte-identical armed vs. opted-out, and the tool still declines the check. The change is visible only to in-repo hosts, which now validate by default. `IrInvariants.Enabled` and `CheckSemantics` are no longer publicly settable; nothing in the repo set them, so this removes an affordance rather than a use.

Cost is noise-level: the corpus gate runs 16.2s armed against 41,952 CoreLib methods, versus 16.7s with the check off.

## Validation

Head `7229b0c1`, merged with `origin/main` (`ccd09562`).

| Suite | Result |
| --- | --- |
| `dotnet build dotnet-inspect.slnx -c Release` | 0 errors, 3 pre-existing warnings |
| Decompiler `--gate fast` | 3887 passed |
| Decompiler `-trait "Area=Corpus"` | 80 passed |
| `dotnet-inspect.Tests` | 2180 passed, 12 skipped (Windows-only P/Invoke) |
| `ILInspector.Analysis.Tests` | 513 passed |
| `ILInspector.Research.Tests` | 132 passed |
| `DotnetInspector.Services.Tests` | 319 passed |
| `ILInspector.Metadata.Tests` | 855 passed |

The teeth are exercised, not assumed: `PipelineRunner_ThrowsWhenAPassCorruptsTheTree` fails if the default is flipped back off, and every census rule below was verified by planting a host that used it and confirming the census both fails with it and passes without it.

`DOTNET_INSPECT_IR_INVARIANTS=FULL` and `=Off` were verified end-to-end against the built CLI. Both changed Markdown files pass `markdownlint`.

## Adversarial review

Reviewed by **GPT** (`gpt-5.6-sol`) and **Gemini Pro** (`gemini-3.1-pro-preview`) — not Claude, since Claude authored the change — in isolated read-only worktrees, over four rounds. Both reported the final head clean. Reconciliation follows in a comment.


richlander:

Two models, four rounds, isolated read-only worktrees. **GPT** (`gpt-5.6-sol`) and **Gemini Pro** (`gemini-3.1-pro-preview`); Claude authored the change so it did not review it.

Every finding below was reproduced on a clean exact-head worktree before it was acted on, and every fix was verified by planting a host that used the evasion and confirming the census fails with it and passes without it.

### Round 1 — head `71e7c74a`

| Finding | Reviewer | Resolution |
| --- | --- | --- |
| A host could decline by assignment through an alias (`Inv.Enabled = false`) or `using static`, and the census stayed green | **Both, independently** | Fixed in `257cb9a6`. Rather than widen the matcher to chase spellings, `Enabled` became `{ get; private set; }` — both evasions now fail to compile with CS0200. |
| Env parsing was exact-lowercase, so with the default inverted, `Off` or `False` would silently leave the check **armed** against an operator who explicitly asked for it off | Gemini | Fixed in `257cb9a6`. `ParseRequest` is trimmed and `OrdinalIgnoreCase`. This one mattered because inverting the default also inverted the consequence of an unrecognised value. |
| `Enabled` being settable was a parallel-test trap | Gemini | Same fix; the private setter removes it. |

### Round 2 — head `257cb9a6`

| Finding | Reviewer | Resolution |
| --- | --- | --- |
| Method-group evasion: `Action a = IrInvariants.DisableForShippedTool; a();` passed a matcher that only looked at `InvocationExpressionSyntax` | **Both, independently** | Fixed in `36ef9f6f`. The census matches any identifier reference, not a call shape, excluding `nameof(...)`. It also flags the method name as a string literal, catching `GetMethod("DisableForShippedTool")`. |
| The census hard-coded `src/`, `tools/`, `tests/`, so a host in a new top-level root (e.g. `benchmarks/`) declined unseen | GPT | Fixed in `8298ffcf`. It now scans from the repository root, excluding only `bin`, `obj`, `.git`, `artifacts`, `node_modules` — and asserts a non-vacuous scan, so it fails loudly instead of passing on an empty result. |
| `CheckSemantics` was still publicly disarmable and raceable — the same silent-decline shape one level down | GPT | Fixed in `8298ffcf`. Private setter plus `EnableSemanticChecks()`, which only arms. |

### Round 3 — head `8298ffcf`

| Finding | Reviewer | Resolution |
| --- | --- | --- |
| Preprocessor blindness: Roslyn parses with no symbols defined, so a decline inside `#if DECLINE_IR_VALIDATION` sat in disabled trivia and never reached the matcher — while building and declining for real under that constant | **GPT (should-fix) and Gemini (non-blocking), independently** | Fixed in `497f6953`. Verified as a real compiler-supported path, not a theoretical bypass, so it was treated as should-fix. The census now re-parses every `DisabledTextTrivia` region under the same rule, covering `#if`, `#elif`, `#else`, and nesting without guessing which symbols a host defines. |
| Shadowed `nameof`: it is a contextual keyword, so `static void nameof(Action a) => a();` turns `nameof(IrInvariants.DisableForShippedTool)` into a real method-group handoff that the exclusion skipped | Gemini | Fixed in `497f6953`. The `nameof` exclusion is withdrawn for any file that declares its own `nameof`. |

### Round 4 — head `497f6953`

Both reported the head clean, and both independently raised the same residual, both ranking it non-blocking: `DeclaresNameOf` only recognised methods, local functions, and delegates, so a delegate-typed **local variable**, **parameter**, or **field** named `nameof` still evaded. Two independent reviewers converging on it was signal enough, and it cost one switch arm, so it was closed in `7229b0c1` rather than documented.

### Round 5 — head `7229b0c1`

Both reviewers confirmed **clean**, with their own evasions re-planted and caught, and no repo-wide false positives.

### Verified and deliberately not actioned

Both reviewers agreed these are deliberate full-trust bypasses rather than merge findings, and they are now stated in the census test's XML doc rather than implied covered:

- **Reflection with a computed name.** The literal form is caught; a concatenated or interpolated name is beyond sound static analysis.
- **A `.cs` file living outside the repository**, pulled in via `<Compile Include="../../.."/>`. Outside any repo-root scan by construction.
- **Cross-file `nameof` shadowing** via `using static` or a partial class. In-file shadowing is closed; cross-file requires deliberate cooperation across two files.
- **Source generators emitting into `obj/`.** Excluded from the scan, but generated code originates from an in-repo generator that is itself scanned.

### Cross-checks requested and confirmed by both

- **Static initializer ordering** is sound: C# runs static field initializers in textual order, and `EnvValue` → `EnvironmentRequest` → `Enabled`/`CheckSemantics` are declared in that order.
- **Census performance** is unaffected by the disabled-region re-parse: the host-contract class runs in ~0.22–0.25s, measured independently by both.
- **No false positives** from scanning the whole repository root, including comments and `<see cref="..."/>` mentions, which do not produce skipped tokens.
- **No legitimate in-repo use** was broken by the private setters; nothing in the repository set either property.

richlander:
Ready to merge

All merge-blocking validation is complete on `a9d58a2c` (`7229b0c1` merged with `origin/main` `ccd09562`):

- CI green: `test (ubuntu-26.04, linux-x64)`, `build-net10`, `csharp-diff-smoke`, `markdownlint`, `changes`. `il-diff-smoke` and `pack` are path-gated and correctly skipped.
- Local suites after the merge: decompiler fast gate 3887, corpus 80, CLI 2180, Analysis 513, Research 132, Services 319, Metadata 855 — all passing.
- Adversarial review by GPT and Gemini Pro both clean on the exact reviewed head, reconciled above.

No follow-up work is required for this change. The residual bypasses listed under *Verified and deliberately not actioned* are documented in the census test rather than left implicit, and are non-blocking by both reviewers' assessment.

richlander:
**Approve.** The shape is right and, unusually, the enforcement is real rather than declared. I verified the load-bearing claims by tampering rather than by reading, and every one of them bit.

This is the correct fix for #3267. I filed that issue asking for a validating-host default; inverting the library default is strictly better than the "validating host helper" I suggested, because it removes the requirement instead of restating it. The layering — compiler owns the spellings it can express, census pins the one that remains — is the right division, and it's the first time in this area I've seen the weaker layer explicitly scoped so it doesn't carry the claim alone.

## What I verified independently

I planted hosts under a new top-level `benchmarks/` directory and re-ran the census. The census reads from disk at runtime, so each spelling is a probe without a rebuild. Every one was caught, and removing the plant restored a clean run:

| Planted spelling | Result |
| --- | --- |
| `IrInvariants.DisableForShippedTool()` in a new top-level dir | caught |
| `using static` + bare call | caught |
| namespace alias `Inv.DisableForShippedTool()` | caught |
| method group to delegate (`Action a = ...; a();`) | caught |
| hidden behind `#if DECLINE_VALIDATION` | caught |
| nested `#if`/`#else`, decline in the `#else` arm | caught |
| reflection via `GetMethod("DisableForShippedTool")` | caught |
| shadowed `nameof` turning the exclusion spelling into a real handoff | caught |
| deeply nested subdirectory | caught |
| **plant removed (positive control)** | **passes** |

The `nameof` withdrawal in `7229b0c1` is the subtle one and it works: declaring a member named `nameof` withdraws the exclusion for the whole file. That's conservative in the fail-loud direction, which is the right bias.

The other two layers hold up too:

- **Compiler:** planting `IrInvariants.Enabled = false;` / `CheckSemantics = false;` in the test project fails with `error CS0200` on both. The private setters are doing real work — worth noting `main` had these as public `{ get; set; }`, so this removes two live affordances.
- **Default has teeth:** flipping the initializer to `hostOptedOut: true` fails exactly two tests — `ValidationIsOnByDefault_WithoutTheHostArmingIt` and `PipelineRunner_ThrowsWhenAPassCorruptsTheTree`. Not vacuous.
- **No silent skip:** `FindRepositoryRoot` throws rather than returning a bogus root, and `Assert.True(sites.Count > 0)` catches a scan that finds nothing. Both failure modes are loud.

**The newly-validating hosts actually run clean.** Your validation table covers `Research.Tests` but not the harness, which is the host whose default this change flips in the most consequential way. I ran it: the CoreLib sweep completes with no invariant throws (41203/41952 Full, 0 importer bugs), and output is **byte-identical** armed vs. `DOTNET_INSPECT_IR_INVARIANTS=0` across all 41,952 methods. That's much stronger evidence for behavior-neutrality than the single-CLI-invocation framing in the PR body.

On cost: I can't reproduce a measurable difference. Three runs each gave armed 12.50/12.15/12.29s vs. off 11.58/12.00/12.97s — fully overlapping on a shared machine. Your conclusion (noise-level) is right; the specific `16.2s` vs `16.7s` pair reports *armed faster than off*, which is itself a tell that the measurement can't support a precise cost claim. Consider just stating it's below measurement noise.

## Non-blocking findings

**1. The "~120 minimal-fixture tests" figure is stale by ~24x, and the correct number changes the conclusion.**

`IrInvariants.cs:71-73` justifies keeping the semantic level opt-in on the grounds that arming it suite-wide "would false-positive on ~120 minimal-fixture tests." Measured on this head:

```
--gate fast                                  → 3887 total, 0 failed
DOTNET_INSPECT_IR_INVARIANTS=full --gate fast → 3887 total, 5 failed
```

All five are the documented shape (`... references local slot 0, out of range for the enclosing scope's 0 local slot(s)`) in hand-built fixtures — `RaisingPassTests` (3) and `IteratorReconstructionPassTests` (2). So the *kind* of rationale is correct; its *weight* is not. At 120 blockers, "opt-in forever" is obviously right. At **5**, arming the semantic level by default is plainly within reach, and that is the natural completion of this PR's own thesis — semantic invariants are the ones that catch real slot bugs, and they're currently the level nothing gets for free.

The number is pre-existing on `main`, so I'm not treating it as introduced here, but this PR rewrites that exact sentence. Correcting `~120` to `5` is nearly free while you're in the line. The "fix 5 fixtures, arm semantics by default" work is a good follow-up issue — happy to file it.

**2. `EnableSemanticChecks()` is a new public API with zero call sites, and its doc names a consumer that doesn't exist.**

```
$ git grep -n "EnableSemanticChecks"
src/.../IrInvariants.cs:71   /// harness corpus sweep arms it (via <see cref="EnableSemanticChecks"/> or
src/.../IrInvariants.cs:78   /// armed (<see cref="EnableSemanticChecks"/>) but never disarmed in-process,
src/.../IrInvariants.cs:93   public static void EnableSemanticChecks() => CheckSemantics = true;
```

Two doc references and the definition. Nothing calls it. The harness arms the level *only* via `DOTNET_INSPECT_IR_INVARIANTS=full`, and `CorpusSweepGateTests` deliberately calls `CheckInvariant(includeSemantics: true)` directly instead (correctly — the comment at `:117` explains the parallelism hazard). So the doc's "the harness corpus sweep arms it (via `EnableSemanticChecks` ...)" is false for that half.

This is strictly better than `main` (which had a public *unrestricted* setter, also uncalled), so it's not a regression. But an arming affordance that no host is required to use is a small instance of the exact shape this PR exists to remove. Either wire it where the doc claims, or drop it and let `full` be the single spelling.

**3. One residual gap is missing from the stated list.**

The doc block enumerates residual gaps honestly (computed-name reflection, out-of-repo `<Compile Include>`) — I appreciate that they're stated rather than papered over. It omits the simplest one: a host or CI job that exports `DOTNET_INSPECT_IR_INVARIANTS=0` disarms validation with no `DisableForShippedTool()` call at all, so the census structurally cannot see it.

I checked whether this is a silent hole, and it isn't — `PipelineRunner_ThrowsWhenAPassCorruptsTheTree` fails under both `=0` and `=off`:

```
DOTNET_INSPECT_IR_INVARIANTS=0   → Failed: 1 (PipelineRunner_ThrowsWhenAPassCorruptsTheTree)
DOTNET_INSPECT_IR_INVARIANTS=off → Failed: 1 (same)
DOTNET_INSPECT_IR_INVARIANTS=full → Failed: 0
```

So CI can't quietly lose validation. Worth noting that detection rests on that *one* test, since `ValidationIsOnByDefault_WithoutTheHostArmingIt` deliberately accommodates the env request. One sentence in the residual-gaps list would make that explicit.

## Validation I ran

Head `a9d58a2c` in an isolated worktree. Note this head is a merge commit; the PR body's evidence cites `7229b0c1`, and `origin/main` has since advanced past the merge (now `7e4e345c`), so these are my numbers on the actual head.

| Check | Result |
| --- | --- |
| `dotnet build dotnet-inspect.slnx -c Release` | 0 errors |
| Decompiler `--gate fast` | 3887 passed, 0 failed |
| Decompiler `-trait "Area=Corpus"` | 80 passed |
| `dotnet-inspect.Tests` (Program.cs changed) | 2180 passed, 12 skipped |
| `IrInvariantsHostContractTests` | 27 passed, 0.22s |
| Harness CoreLib sweep, armed | clean; byte-identical to unarmed |
| `markdownlint AGENTS.md docs/decompiler-correctness-pipeline.md` | clean |

All tampering reverted; worktree verified clean before and after.

None of the three findings blocks merge. Finding 1 is the one I'd act on, and mostly because the corrected number argues for going further than this PR does.

richlander:
Follow-ups from this review are filed, since it merged:

- **#3302** — arm the semantic level by default. The blocker is **5** fixtures, not the ~120 the rationale claims; corrected number and the five test names are in the issue. If that work is deferred, the issue notes the minimum action is correcting `IrInvariants.cs:70`.
- **#3303** — `EnableSemanticChecks()` has no callers and its doc names a consumer that avoids it by design; the residual-gaps list omits the `DOTNET_INSPECT_IR_INVARIANTS=0` bypass.

Neither was blocking, and neither changes my assessment of this PR: the layered enforcement is real, and the census caught all nine evasion spellings I planted against it.

Issue #3280, opened 2026-07-27
Collapse the two member call-graph sections into one bidirectional section via a projector plus Markout graph shape


richlander:
> **Rewritten 2026-07-26** after a two-model adversarial design review. The
> original framing (per-tier Mermaid re-projection driven by the seam) rested on
> premises that did not survive review. See "Design review outcome" at the
> bottom for what was retracted and why.

## Goal

Collapse `Call Graph` and `Caller Graph` into a single **bidirectional** member
call-graph section, rendered through a three-layer pipeline, **with no loss of
existing capability**.

## The three layers

1. **Projector** (`ILInspector.CallGraph`, see #3291) — builds a typed,
   format-free combined graph: identity dedup, bodiless-root reconciliation,
   inbound edge inversion, cycle collapse, status precedence, deterministic
   ordering, loop-edge merging.
2. **Generic graph shape in Markout** (richlander/markout#163) — opaque ids,
   labels, optional group, focus marker, directed edges. Owns formatting only.
   Blocked on richlander/markout#162 (label escaping is currently unsafe for
   angle brackets, which are pervasive in .NET member names).
3. **CLI adapter** (`dotnet-inspect`) — maps the typed graph onto the Markout
   shape, owning call-graph vocabulary, field projection, and section wiring.

This is the direct consequence of "the ILInspector layer should own data, not
formats": graph *construction* is data and stays; Mermaid *emission* is a format
and goes.

## Parity is a requirement, not a follow-up

The collapsed section must preserve what ships today:

- **Depth.** `BuildCallTree` / `BuildCallerTree` default to `maxDepth: 3,
  maxNodes: 25` (`LibraryBodyIndex.cs:1330,1404`). Multi-level traversal is
  covered by `MemberCallGraphSectionTests.cs:11-23,119-130`.
- **Cross-assembly caller scopes.** `Caller Graph` already incorporates
  `--bin` / `--project` / `--caller-package` through `CallerScopes(...)`
  (`ApiMemberAnalysisInspection.cs:74-75`), covered by
  `MemberCallersSectionTests.cs:163-187`.

Parity requires **no new flags** — it keeps the flags that already exist.
Configurable depth and multi-library scope UX remain a separate follow-up.

## Three lowerings

| Output | Lowering |
| --- | --- |
| plain text / ANSI / Markdown | tree, rooted at the focus node |
| `--table` / TSV / JSONL | edge list, one row per edge |
| `--mermaid` | diagram |

[...]

PR #3297, opened 2026-07-27
Split CallGraphMermaid into a format-neutral projector plus Mermaid emission

### LINKED ISSUE #3291: Split CallGraphMermaid into a format-neutral graph projector plus Mermaid emission
## Summary

Split `src/ILInspector.CallGraph/CallGraphMermaid.cs` into a format-neutral
**graph projector** (which stays in `ILInspector.*`) and a **Mermaid emission**
step (which does not). No user-visible behavior change.

## Why

The `ILInspector` layer should own data, not formats. `CallGraphMermaid.cs` is a
single file whose *name* says Mermaid but whose *content* is mostly graph
semantics:

- generic-erased node identity and deduplication (`_ids` keyed by member, so the
  shared target collapses to one centred node)
- bodiless-root reconciliation
- inbound edge inversion (a reverse tree's child calls its parent)
- duplicate and cycle collapse
- status precedence (external / truncated / target)
- deterministic node and edge ordering
- loop-edge merging

None of that is formatting. All of it is graph construction, and it is needed by
any consumer regardless of output format.

## Why now

Two consumers need the construction logic without the Mermaid emission:

1. The CLI is moving to render call graphs through a generic Markout graph shape
   (richlander/markout#163), which owns formatting but cannot own domain graph
   semantics.
2. The browser prototype is intended to generate its own Mermaid from the typed
   data rather than sharing the CLI's renderer.

Both need "build the combined caller → target → callee graph" as a typed,
format-free product capability.

## Proposed work

- Extract the construction logic into a format-neutral projector in
  `ILInspector.CallGraph` — a typed combined graph (nodes, edges, focus,
  per-node status/source) built from the caller and callee roots. Consider
  renaming the type/project to something format-neutral (e.g.
  `CallGraphProjection`), since `CallGraphMermaid` will no longer describe it.
- Keep the existing `Render(...)` entry point working on top of the projector
  for now, so nothing breaks while richlander/markout#163 is in flight. The
  Mermaid emission is removed in the follow-up adoption work, not here.
- Preserve the existing null-tolerance (`Render(callerRoot, null)`) as projector
  behavior: either root may be absent, but not both, and when both are present
  they must name the same member.

## Constraints

- Keep it SRM-only, NativeAOT-friendly, Roslyn-free, and free of
  inspected-assembly loading, per `AGENTS.md`.
- Do not infer identity from display text. `AGENTS.md` explicitly separates
  identifiers from presentation; node identity must continue to come from typed
  member identity, not labels.
- Determinism is part of the contract, not an implementation detail — the
  existing ordering guarantees must be preserved and tested.

## Notes

- `src/ILInspector.CallGraph` is currently referenced by **both**
  `src/ILInspector.Analysis.Tests` and `src/dotnet-inspect.Tests`; the product
  project `src/dotnet-inspect` does **not** reference it, and there is still no
  runtime product consumer.
- Existing coverage lives in
  `src/ILInspector.Analysis.Tests/CallGraphMermaidTests.cs`. Projector behavior
  (identity, inversion, cycle collapse, ordering, status precedence) should get
  direct tests rather than being asserted only through rendered Mermaid text.

## Relationship to other work

Prerequisite for the call-graph section collapse. Independent of
richlander/markout#162 and #163 — this can land first and on its own.

richlander:
Fixes #3291.

`CallGraphMermaid.cs` had a name that said Mermaid and a body that was mostly graph semantics. None of it is formatting, and all of it is needed by any consumer regardless of output format, so it belongs to the data layer per the layer-ownership rule in `AGENTS.md`.

## What moved

Everything below moves out of the Mermaid file into a new format-neutral `CallGraphProjection`:

- generic-erased node identity and deduplication
- bodiless-root reconciliation (resolved member wins over an `Unsupported` placeholder)
- inbound edge inversion
- duplicate and cycle collapse
- node status precedence (focus is sticky; expanded outranks a boundary)
- loop-edge merging
- deterministic node and edge ordering

## Shape

```csharp
public sealed class CallGraphProjection
{
    public ImmutableArray<CallGraphNode> Nodes { get; }   // focus first, then caller DFS, then callee DFS
    public ImmutableArray<CallGraphEdge> Edges { get; }   // first-seen order, always caller -> callee
    public CallGraphNode Focus { get; }

    public static CallGraphProjection Create(CallTreeNode? callerRoot, CallTreeNode? calleeRoot);
    public static CallGraphProjection FromCallers(CallTreeNode callerRoot);
    public static CallGraphProjection FromCallees(CallTreeNode calleeRoot);
}
```

Two deliberate choices:

- **Edges are always oriented caller → callee.** A reverse (caller) tree records the caller as a *child*, so the inversion happens once during projection instead of being reinterpreted by every host.
- **`CallGraphNode` carries both `MemberRef` and `Label`.** The `MemberRef` is the identity; the label is a convenience spelling a host may ignore. `AGENTS.md` separates identifiers from presentation, and `IdentityComesFromMemberNotLabel` pins it with two members that share a label but differ by assembly.

Ordering is documented as contract rather than left as an implementation detail, since hosts index into `Nodes` by edge endpoint.

## What `CallGraphMermaid` keeps

Only the Mermaid format: node id spelling, escaping, shape/class syntax, edge syntax. Its three existing entry points are unchanged, and it gains a `Render(CallGraphProjection)` overload so a host can hold the projection, render its own format, and still fall back to the shared emitter.

## Not in scope

No user-visible behavior change. Removing the Mermaid emission entirely is the follow-up adoption work in #3280, not this PR. This PR is independent of richlander/markout#162 and #163 and can land on its own.

## Validation

| Check | Result |
| --- | --- |
| `dotnet build dotnet-inspect.slnx -c Release` | succeeds (3 pre-existing nullability warnings in `ILInspector.Metadata.Tests`, untouched) |
| `ILInspector.Analysis.Tests` on `origin/main` | 513 total, 1 failed |
| `ILInspector.Analysis.Tests` on this branch | 539 total, 1 failed |
| `dotnet-inspect.Tests` `ProgressiveMemberCallGraphTests` | 10/10 |

+26 new projector tests. The single failure is pre-existing and identical in both runs — `LibraryBodyIndexTests.CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition` — measured against a clean detached `origin/main` worktree, not assumed.

**The no-behavior-change evidence is that the 20 existing `CallGraphMermaidTests` are unmodified and still pass**, including the byte-exact combined-document assertion, escaping, ordering, cycle collapse, and status styling.

The new tests assert graph semantics as typed data rather than as rendered text, so a future host that renders a table or a tree is not relying on Mermaid strings to prove them. `MermaidEmissionAgreesWithTheProjectionItRendersFrom` pins the seam itself.

### DIFF
--- a/src/ILInspector.Analysis.Tests/CallGraphProjectionTests.cs
+++ b/src/ILInspector.Analysis.Tests/CallGraphProjectionTests.cs
@@ -1,0 +1,336 @@
+using System.Collections.Immutable;
+
+using ILInspector.Analysis;
+using ILInspector.CallGraph;
+
+namespace ILInspector.Analysis.Tests;
+
+/// <summary>
+/// Covers the format-neutral call-graph projection (issue #3291) as typed data rather
+/// than as rendered text: edge inversion, generic-erased node identity, duplicate and
+/// cycle collapsing, node-kind precedence, loop-edge merging, and deterministic node and
+/// edge ordering. <see cref="CallGraphMermaidTests"/> covers the Mermaid emission on top
+/// of this; these tests exist so graph semantics are asserted directly and a future host
+/// that renders a table or a tree is not relying on Mermaid text to prove them.
+/// Trees are constructed directly so the projection is exercised in isolation from IL
+/// decoding.
+/// </summary>
+public class CallGraphProjectionTests
+{
+    static TypeRef Type(string name) => TypeRef.Definition("Sample", "Sample", name);
+
+    static MemberRef Member(string typeName, string method, params TypeRef[] parameters)
+        => new(Type(typeName), method, [.. parameters], TypeRef.CoreLib("System", "Void"), MemberKind.Method);
+
+    static CallTreePerf Perf(bool inLoop, string? loopHint)
+        => new(0, 0, 1, inLoop, loopHint);
+
+    static CallTreeNode Node(
+        MemberRef member,
+        CallTreeStatus status,
+        ImmutableArray<CallTreeNode> children,
+        bool inLoop = false,
+        string? loopHint = null)
+        => new(member, null, status, children, Perf(inLoop, loopHint));
+
+    static CallTreeNode Leaf(
+        MemberRef member,
+        CallTreeStatus status = CallTreeStatus.Leaf,
+        bool inLoop = false,
+        string? loopHint = null)
+        => Node(member, status, [], inLoop, loopHint);
+
+    static (int From, int To, string? Loop)[] EdgeTuples(CallGraphProjection projection)
+        => [.. projection.Edges.Select(e => (e.From, e.To, e.LoopLabel))];
+
+    [Fact]
+    public void FocusIsAlwaysNodeZero()
+    {
+        var projection = CallGraphProjection.FromCallees(
+            Node(Member("Target", "Run"), CallTreeStatus.Expanded, [Leaf(Member("Svc", "Do"))]));
+
+        Assert.Equal(0, projection.Focus.Id);
+        Assert.Equal(CallGraphNodeKind.Focus, projection.Focus.Kind);
+        Assert.Same(projection.Nodes[0], projection.Focus);
+    }
+
+    [Fact]
+    public void CalleeEdgesPointFromFocusToCallee()
+    {
+        var projection = CallGraphProjection.FromCallees(
+            Node(Member("Target", "Run"), CallTreeStatus.Expanded, [Leaf(Member("Svc", "Do"))]));
+
+        // Outbound: the selected overload calls its callee.
+        Assert.Equal([(0, 1, (string?)null)], EdgeTuples(projection));
+    }
+
+    [Fact]
+    public void CallerEdgesAreInvertedToPointIntoFocus()
+    {
+        var projection = CallGraphProjection.FromCallers(
+            Node(Member("Target", "Run"), CallTreeStatus.Expanded, [Leaf(Member("Client", "Invoke"))]));
+
+        // Inbound: a reverse tree records the caller as a child, but the projected edge
+        // must be oriented caller -> callee so a host never has to invert it.
+        Assert.Equal([(1, 0, (string?)null)], EdgeTuples(projection));
+    }
+
+    [Fact]
+    public void NodesAreOrderedFocusThenCallersThenCallees()
+    {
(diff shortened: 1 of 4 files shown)


richlander:
## Adversarial review

This is a pure refactor with a strong pre-existing regression net (20 unmodified `CallGraphMermaidTests`, including a byte-exact document assertion) and a bounded blast radius, so it sits in the single-review tier. Reviewed with **MAI-Code** in an isolated worktree pinned to `92919eda`.

**Verdict: CLEAN.** No blocking or non-blocking findings.

The reviewer was given six concrete attack points and reported on each:

| Attack point | Conclusion |
| --- | --- |
| Behavior drift between old and new | No semantic mismatch found. Verified by line-by-line diff **plus fuzzing both renderers against 1000 randomly generated call trees** and comparing output. |
| `IdentityKey` vs `LibraryBodyIndex.CallerGraphKey` | Same generic-erasure and open-signature rules; no divergence that could merge or split nodes incorrectly. |
| Aliasing / mutation escape from the builder | Builder state stays private; the public type exposes immutable arrays of fresh values. No path for a caller to observe or mutate it. |
| Null and empty inputs | `Create(null, null)` throws; single-sided nulls throw as intended; a childless root still yields a focus node; **mismatched focus members throw rather than silently producing a success-shaped empty graph**, per the failure-visibility rule in `AGENTS.md`. |
| `Render(CallGraphProjection)` vs the three legacy entry points | Matched for the same inputs in the reviewer's probe — no latent divergence waiting to bite #3280. |
| Test quality / tautology check | The 26 new tests pin real semantics (edge inversion, collapse, precedence, loop labels, ordering, unsupported-root handling, null input, mismatch rejection) rather than asserting the implementation against itself. |

The fuzz comparison is the result I care most about, since it independently exercises the no-behavior-change claim across randomized inputs rather than only the curated cases the existing tests cover.

Reviewer-observed totals, matching the PR body: `ILInspector.Analysis.Tests` 539 total / 1 failed, with the failure being the known pre-existing `LibraryBodyIndexTests.CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition`; `CallGraphProjectionTests` 26/26.

Nothing to reconcile — no findings were raised, so there is nothing dismissed or deferred.

Ready to merge.

PR #3305, opened 2026-07-27
Record the EVIL v2 row at a main commit (#3079)

richlander:
## Summary

The v2 row added in #3231 was stamped `92ea6dd0` — that PR's **branch head**, not a commit on `main`. Verifying it after merge showed it does not reproduce: the merged commit `14781e8d` measures `productBodyDefect` **326**, not 366.

This replaces the row with one measured at `14781e8d`, and makes the append procedure require a `main` commit so it can't recur.

## The 366 was not wrong — it was measured against a different base

Both numbers are honest for their own code. `92ea6dd0` was branched from `main` before that PR merged, so it carried an older decompiler. Controlled runs, pinned pool `0a7eded85c3e1410`, 26/26 assemblies, 12,000 targets:

| run | product | harness | uncl. | SpanMeasured |
| `14781e8d`, `parseOptions` threaded | 326 | 4808 | 64 | 23 (CS1525) |
| `14781e8d`, `parseOptions` reverted | 326 | 4808 | 64 | 23 (CS1525) |
| `92ea6dd0` (older decompiler) | 366 | 4769 | 82 | 62 (CS1001 ×37, CS1525 ×25) |

**The first two rows are byte-identical**, so the `parseOptions` threading added during #3231's review is inert on the real corpus — exactly as predicted there. Confirmed independently by a probe: no preview construct (extension blocks, `field`, collection expressions, params collections, ref-struct interfaces, partial ctors) produces a parser diagnostic under either parse mode, because language-version gating is a *binding* diagnostic, not a parser one.

So the entire 366 → 326 delta is main's decompiler work landing in between:

| metric | `92ea6dd0` | `14781e8d` | Δ |
| --- | --- | --- | --- |
| correct | 1560 | 1576 | **+16** |
| invalid | 5217 | 5198 | **−19** |
| productBodyDefect | 366 | 326 | −40 |
| harnessShellReconstruction | 4769 | 4808 | +39 |
| unclassified | 82 | 64 | −18 |

`correct +16` / `invalid −19` are genuine fixes. The product −40 / harness +39 is **correct reattribution**: 37 bodies that previously failed to parse (CS1001) now parse, so span attribution rightly stops crediting a body defect and those rows fall back to their still-broken shell. The metric behaved exactly as designed.

## Why replace rather than append

The store is a trend series, so a row has to be reproducible from its recorded commit. Appending would leave a branch-commit point in the series and chart a −40 step that conflates real improvement with a baseline that was never on main's line.

The v1-vs-v2 methodology comparison (**304 vs 366 on identical product code**) is a code-controlled A/B and remains documented in #3231 — which is where an experiment against its own base belongs. Nothing is lost by removing the row here.

## Evidence

Rendered card after the change — every row a reproducible main commit, boundary split intact, no bogus step:

```
| 2026-07-25 | abec2dd7 | 56.5% | 1560 | 5217 | 304 | 4831 | v1 |
| 2026-07-26 | 14781e8d | 56.7% | 1576 | 5198 | 326 | 4808 | v2 |

| Product defects (v1 substitution lower bound) ↓ | - | 306 | 304 ✓ | - |
| Product defects (v2 span-measured lower bound) ↓ | - | - | - | 326 |
```

All 5 lines parse; harness builds; README passes markdownlint.

## Review

Low risk: corrects one recorded data row to a measured value and adds a procedural constraint. No product or harness logic changes.


richlander:
Ready to merge

CI green (`test`, `markdownlint`, `changes`). No product or harness logic changes — this corrects one recorded data row to its measured value at a reproducible main commit and adds a procedural constraint to the append steps, so no adversarial review tier applies.

Evidence is in the PR body: the two `14781e8d` runs (with and without `parseOptions`) are byte-identical, isolating the entire 366 → 326 delta to main's decompiler work rather than to any methodology or harness change.

Issue #3302, opened 2026-07-27
Semantic IR invariants stay opt-in on a stale rationale: the blocker is 5 fixtures, not ~120


richlander:
Follow-up from #3289 (merged), which armed the **structural** IR invariant level by default so no host can silently skip validation. The **semantic** level is still opt-in, and the stated reason for that no longer matches reality.

## The rationale rests on a stale number

`src/ILInspector.Decompiler/Pipeline/Ir/IrInvariants.cs:69-71` justifies keeping the semantic level off:

> Off by default even in validating hosts, because hand-built test fixtures legitimately omit the local table these checks validate against; enabling it suite-wide would false-positive on **~120 minimal-fixture tests**.

Measured on #3289's head (`a9d58a2c`, Release, this machine):

```
--gate fast                                    → Total: 3887, Failed: 0
DOTNET_INSPECT_IR_INVARIANTS=full --gate fast  → Total: 3887, Failed: 5
```

Five, not ~120 — stale by roughly 24x. The five are:

| Test | Class |
| --- | --- |
| `SharedThrow_SingleGuardFallenInto_InlinesAsGuardClauses` | `RaisingPassTests` |
| `JumpTable_ValueBlocksReturnedAtJoin_RaisesToSwitchExpression` | `RaisingPassTests` |
| `TypedConstants_RunAgainAfterInlining_CatchExposedPositions` | `RaisingPassTests` |
| `SwitchIterator_WithFallthroughDefaultSideEffect_DeclinesHonestly` | `IteratorReconstructionPassTests` |
| `YieldBreakLoopIterator_WithDispatchSideEffect_DeclinesHonestly` | `IteratorReconstructionPassTests` |

All five fail with exactly the shape the doc describes — a fixture that uses a slot without populating `Locals`:

```
System.InvalidOperationException : Invariant violated: 'StoreLocal 0 (int)' references
local slot 0, out of range for the enclosing scope's 0 local slot(s).
   at IrNode.ValidateLocalSlot(Int32 localScope)   IrNode.cs:298
   at IrNode.CheckSubtree(Int32 localScope, Boolean includeSemantics)  IrNode.cs:231
```

So the *kind* of rationale is correct. These are fixture artifacts, not masked product bugs — I checked, and none of the five is a real violation.

## Why the corrected number changes the conclusion

[...]

PR #3306, opened 2026-07-27
Drop the uncalled semantic-arming affordance and name the env bypass (#3303)

### LINKED ISSUE #3303: IrInvariants: EnableSemanticChecks() has no callers and its doc names a consumer that avoids it; residual-gaps list omits the env bypass
Two small accuracy defects in the `IrInvariants` surface introduced or left behind by #3289 (merged). Neither is a correctness bug; both are cases of the surface describing something that isn't true, which is the shape #3289 exists to remove.

## 1. `EnableSemanticChecks()` has no callers, and its doc names a consumer that does not exist

`src/ILInspector.Decompiler/Pipeline/Ir/IrInvariants.cs:93` adds a public arming method. On current `main` (`565c7104`) it has zero call sites:

```
$ git grep -n "EnableSemanticChecks"
src/ILInspector.Decompiler/Pipeline/Ir/IrInvariants.cs:71   /// harness corpus sweep arms it (via <see cref="EnableSemanticChecks"/> or
src/ILInspector.Decompiler/Pipeline/Ir/IrInvariants.cs:78   /// armed (<see cref="EnableSemanticChecks"/>) but never disarmed in-process,
src/ILInspector.Decompiler/Pipeline/Ir/IrInvariants.cs:93   public static void EnableSemanticChecks() => CheckSemantics = true;
```

Two doc references and the definition. Nothing calls it.

The doc at `:70-73` says "The harness corpus sweep arms it (via `EnableSemanticChecks` or `DOTNET_INSPECT_IR_INVARIANTS=full`)". The first half is false — the harness arms the level only through the environment variable. `CorpusSweepGateTests` deliberately does **not** use the global flag, and says so at `CorpusSweepGateTests.cs:117`:

> Call the level explicitly rather than flipping `IrInvariants.CheckSemantics`: xUnit runs collections in parallel, and the global flag would false-positive the minimal-fixture pass tests.

That reasoning is correct, and it means the in-repo consumer this method was written for is never going to call it.

To be fair to #3289, this is strictly better than what it replaced: `main` previously had `public static bool CheckSemantics { get; set; }`, an unrestricted setter that was also uncalled. Narrowing an unused affordance is progress. But an arming method that no host is required to use — and that the one plausible host is documented as avoiding — is a small instance of the exact shape #3267/#3289 set out to eliminate.

**Suggested:** either drop `EnableSemanticChecks()` and let `DOTNET_INSPECT_IR_INVARIANTS=full` be the single spelling, or give it a real caller. If #3302 lands (arming semantics by default), the method likely becomes unnecessary either way. Correct the `:70-73` doc regardless.

## 2. The residual-gaps list omits the simplest bypass

`src/ILInspector.Decompiler.Tests/IrInvariantsHostContractTests.cs:157-164` states residual gaps honestly rather than papering over them — computed-name reflection, and out-of-repo files pulled in with `<Compile Include="../../.."/>`. That candour is the right instinct and worth keeping.

It omits the easiest one: **a host or CI job that sets `DOTNET_INSPECT_IR_INVARIANTS=0` disarms validation with no `DisableForShippedTool()` call at all**, so the source census structurally cannot see it. Unlike the two listed gaps, this one needs no exotic spelling — one line in a workflow `env:` block does it.

I checked whether this is a silent hole, and it is not:

```
DOTNET_INSPECT_IR_INVARIANTS=0    → Failed: 1  (PipelineRunner_ThrowsWhenAPassCorruptsTheTree)
DOTNET_INSPECT_IR_INVARIANTS=off  → Failed: 1  (same)
DOTNET_INSPECT_IR_INVARIANTS=full → Failed: 0
```

So CI cannot quietly lose validation — good. But detection rests on that **single** test, because `ValidationIsOnByDefault_WithoutTheHostArmingIt:46-52` deliberately accommodates an explicit off request and passes. That is a defensible design (an operator who asks for off has asked for it), but it means the census's blind spot is covered by one assertion in one unrelated-sounding test rather than by anything named for the job.

**Suggested:** add a sentence to the residual-gaps block naming the environment bypass and pointing at `PipelineRunner_ThrowsWhenAPassCorruptsTheTree` as what catches it. Optionally assert it directly in `IrInvariantsHostContractTests` so the guarantee is stated where someone reasoning about the host contract will look.

## Evidence

All measurements on #3289 head `a9d58a2c` in an isolated worktree, Release, tampering reverted and tree verified clean. The census itself is solid — I planted nine evasion spellings (`using static`, namespace alias, method group to delegate, `#if`-hidden, nested `#if`/`#else`, reflection string literal, shadowed `nameof`, deep subdirectory, plain call in a new top-level directory) and all nine were caught, with the plant-removed control passing. These two items are about the surrounding claims, not the guard.

Related: #3267, #3289, #3271, #3302.

richlander:
Closes #3303.

Two accuracy defects in the `IrInvariants` surface. Neither is a correctness bug; both are cases of the surface describing something that is not true — the shape #3267/#3289 exist to remove.

## 1. `EnableSemanticChecks()` had no callers, and its doc named a consumer that avoids it

Zero call sites repo-wide. Its doc claimed "the harness corpus sweep arms it (via `EnableSemanticChecks` or `DOTNET_INSPECT_IR_INVARIANTS=full`)", which is false in both directions: the harness arms the level only through the environment variable, and `CorpusSweepGateTests:117` documents *deliberately avoiding* the global, because xUnit runs collections in parallel and a process-wide level would false-positive the minimal-fixture pass tests. The one plausible in-repo host was never going to call it.

Dropped, leaving two honest spellings — `DOTNET_INSPECT_IR_INVARIANTS=full` to raise the level, `CheckInvariant(includeSemantics: true)` to thread it per call. With no arming method, `CheckSemantics` needs no setter at all rather than a private one, so the level is now fixed at startup and cannot move in either direction in-process.

**`ThePublicSurfaceIsExactlyTheShippedToolOptOut`** pins the public surface by set equality. The failure this catches is an *addition*: a public entry point on this type changes what a host validates, so each one must arrive with a host contract pinned in `IrInvariantsHostContractTests` rather than reading as a supported way to move the level while nothing holds it to one.

## 2. The residual-gaps list omitted the env bypass

The block listed computed-name reflection and out-of-repo `<Compile Include="../../.."/>` but not the cheapest decline of all: a host or CI job setting `DOTNET_INSPECT_IR_INVARIANTS=0` disarms validation with **no `DisableForShippedTool()` call**, so the source census structurally cannot see it — one line in a workflow `env:` block does it.

It was already caught, but only as a side effect of an assertion inside `PipelineRunner_ThrowsWhenAPassCorruptsTheTree`, whose name gives a reader no reason to expect it. **`AnEnvironmentOffRequestDoesNotSilentlyDisarmTheSuite`** now states the guarantee where a reader of the host contract will look, and reports the actual cause rather than a confusing "should be armed by default" for a run that was disarmed on purpose. The gaps block names the bypass and points at both gates.

The guarantee is deliberately *not* that the bypass is impossible — an operator who asks for off has asked for it, and `EnvironmentRequestOutranksTheHostOptOut` keeps honoring that. The guarantee is that such a run cannot come back green.

## Evidence

Release, `origin/main` base `407515c1`.

| scenario | result |
| --- | --- |
| `IrInvariantsHostContractTests` + `IrInvariantCheckTests`, untampered | **40 passed, 0 failed** |
| full `--gate fast` | **3893 passed, 0 failed** |
| re-add uncalled `EnableSemanticChecks()` | **1 failed** — `Actual: ["DisableForShippedTool", "EnableSemanticChecks"]` |
| `DOTNET_INSPECT_IR_INVARIANTS=0` | **2 failed** — now includes the named test, reporting the cause |
| `DOTNET_INSPECT_IR_INVARIANTS=off` | **2 failed** — same |
| `DOTNET_INSPECT_IR_INVARIANTS=full` | **0 failed** |

Both new gates verified to have teeth; tamper reverted and tree confirmed clean before the final run.

## Risk

**No behavior change.** The removed method was never called, and `CheckSemantics` still resolves from the same environment value at the same time. `ILInspector.Decompiler` is not packable (only `dotnet-inspect` and `runfaster` pack), so removing a public method is not a consumer API break. The remainder is test and doc accuracy. No attribution, pipeline, or output logic touched.

Proposing single-review tier on that basis.

## Not in scope

#3302 (whether the semantic level should be armed by default on a fresh rationale) is untouched. If it lands, the environment spelling remains the only lever, so this change does not constrain it.


richlander:
CI green on `6e93c9d0`: `test (ubuntu-26.04, linux-x64)` pass (9m1s), `build-net10` pass, `csharp-diff-smoke` pass. `markdownlint`/`pack`/`il-diff-smoke` path-gated to skip (no Markdown, packaging, or IL-diff surface touched).

Local, Release, base `407515c1`: `--gate fast` 3893 passed / 0 failed; both new gates tamper-verified (re-added affordance → 1 failed; `DOTNET_INSPECT_IR_INVARIANTS=0` → the named test fails with the cause).

No adversarial review requested: the removed method had zero call sites, `ILInspector.Decompiler` is not packable so it is not a consumer API break, and the rest is test and doc accuracy with no attribution, pipeline, or output logic touched. Same tier as #3270.

Ready to merge.

richlander:
**One small change requested.** The substance is right — dropping `EnableSemanticChecks()` and making `CheckSemantics` setter-less is a better resolution of #3303 than I proposed, because removing the setter entirely means the level can't move in either direction rather than only being hard to lower. Every number in your evidence table reproduces exactly. But the new surface pin doesn't enforce what its name and doc claim, and the specific gap admits the exact shape #3289 removed.

## `ThePublicSurfaceIsExactlyTheShippedToolOptOut` pins methods, not the public surface

The test enumerates `GetMethods(...)` and filters `!method.IsSpecialName`. Property accessors *are* `IsSpecialName`, so they're filtered out; fields aren't methods at all. I injected three members into `IrInvariants` and re-ran the two classes:

| Injected into `IrInvariants` | Result |
| --- | --- |
| `public static void EnableSemanticChecks() { }` | **1 failed** — caught, as you claim |
| `public static bool SneakyLevel { get; set; } = true;` | **0 failed** — not caught |
| `public static bool SneakyField = true;` | **0 failed** — not caught |

`NeitherLevelHasAPubliclyWritableSetter` doesn't close it either: it checks `Enabled` and `CheckSemantics` by name, so it has nothing to say about a third property.

This matters more than a generic completeness nit because `{ get; set; }` on a static bool in this type *is* the pre-#3289 defect. That's literally what `Enabled` and `CheckSemantics` were on `main` before #3289 narrowed them. So a future `public static bool ValidateEverything { get; set; }` reintroduces the original problem — a host can move validation with no census-visible call site — and sails through both gates while a test named `ThePublicSurfaceIsExactlyTheShippedToolOptOut` reports green.

The doc states the broader claim explicitly:

> Every public entry point on this type changes what a host validates, so each one needs a host contract pinned here.

That's the right rule. The implementation pins one of the three ways to spell an entry point. A PR whose whole subject is surfaces that describe something untrue shouldn't ship a new one.

Fix is small — enumerate `GetMembers(BindingFlags.Public | BindingFlags.Static | BindingFlags.DeclaredOnly)` and pin that set, or add properties and fields alongside the methods. Either way the expected set becomes `DisableForShippedTool`, `Enabled`, `CheckSemantics`, which also makes the two get-only properties visible in the pin rather than implied by a separate test.

## Trivial: the new diagnostic misattributes a code-caused disarm to the operator

`AnEnvironmentOffRequestDoesNotSilentlyDisarmTheSuite` hardcodes the environment explanation. With the default flipped to `hostOptedOut: true` and the variable **unset**, it fails with:

```
IR invariant validation is off for this run because DOTNET_INSPECT_IR_INVARIANTS was set to ''.
That request is honored on purpose, but a disarmed run is not a passing run: every pipeline test
below validates nothing. Unset the variable to restore coverage.
```

It blames a variable that isn't set, and tells the reader to unset it. Your stated motivation for adding this test was to report "the actual cause rather than a confusing 'should be armed by default' for a run that was disarmed on purpose" — this is the same defect mirrored: a run disarmed in code reports an operator cause.

Nothing is missed (`ValidationIsOnByDefault_WithoutTheHostArmingIt` fails in that scenario with an accurate message), so this is message accuracy only. Branching on `EnvironmentRequest() is false` covers both.

## Non-blocking: `~120` carried through two rewritten comments

This PR rewrites the doc paragraph at `IrInvariants.cs:69-71` *and* the comment block at `IrInvariantCheckTests.cs:75-79`, and preserves `~120 minimal-fixture tests` in both. Measured on the #3289 head is **5** (#3302).

Scoping the *decision* — whether to arm semantics by default — to #3302 is correct, and I'd have objected if you'd bundled it. But #3302 says that if the larger work is deferred, the minimum action is correcting the number, and this change was inside both lines. Carrying a figure that's wrong by ~24x through a rewrite of the exact sentence containing it is the same category as the two defects this PR fixes. Two-character change; your call whether it belongs here or in #3302.

## What I verified

Head `6e93c9d0`, base `origin/main` `407515c1`, Release, isolated worktree.

| Check | Result |
| --- | --- |
| `IrInvariantsHostContractTests` + `IrInvariantCheckTests` | 40 passed, 0 failed |
| `DOTNET_INSPECT_IR_INVARIANTS=0` | 2 failed |
| `DOTNET_INSPECT_IR_INVARIANTS=off` | 2 failed |
| `DOTNET_INSPECT_IR_INVARIANTS=full` | 0 failed |
| `--gate fast` | 3893 passed, 0 failed |
| re-add `EnableSemanticChecks()` | 1 failed, teeth confirmed |

Every figure matches your table.

Two of your claims I checked rather than took:

- **The removal is safe.** Whole-solution Release build is clean, which covers the harness and every other in-repo host, so there genuinely was no caller.
- **No consumer API break.** Confirmed for the right reason: every `dotnet pack` in `ci.yml` and `release.yml` targets `src/dotnet-inspect` directly, so `ILInspector.Decompiler` is never published regardless of its `IsPackable` default (which isn't set anywhere, so it's the SDK's `true`). The conclusion holds; the mechanism is the pack flow's target, not a project-level opt-out.

Tampering reverted and tree confirmed clean before the final gate run.

Single-review tier is reasonable for this change. Re-request when the surface pin covers properties and fields and I'll re-run the injection matrix against the fixed head.

richlander:
All three addressed in `0dad7fc3`, merged with `origin/main` (`0d486700`) at head `98e984f2`.

## Blocking: the surface pin covered methods only — fixed

You were right, and the framing was the useful part: `{ get; set; }` on a static bool in this type *is* the pre-#3289 defect, so the gap admitted exactly the shape the PR claims to close.

Now enumerates `GetMembers(...)`, folding property accessors into the property they belong to so the pin reads as source, while operators and fields still appear under their own names. Expected set is `CheckSemantics`, `DisableForShippedTool`, `Enabled` — so the two get-only properties are visible in the pin rather than implied by `NeitherLevelHasAPubliclyWritableSetter` alone, as you suggested.

One addition beyond your ask: `BindingFlags.Instance` is included even though the type is static. Dropping `static` is how a public instance affordance would arrive, and I would rather the pin fail loudly than quietly stop applying.

Your injection matrix, re-run on the fixed head:

| Injected into `IrInvariants` | Before | After |
| --- | --- | --- |
| `public static void EnableSemanticChecks() { }` | 1 failed | **1 failed** |
| `public static bool SneakyLevel { get; set; } = true;` | 0 failed | **1 failed** |
| `public static bool SneakyField = true;` | 0 failed | **1 failed** |
| untampered | 0 failed | **40 passed, 0 failed** |

## Trivial: misattributed diagnostic — fixed

Fair catch, and the mirror-image framing is exact. It now branches on `EnvironmentRequest()`. Both branches verified by real runs:

`DOTNET_INSPECT_IR_INVARIANTS=0`:

> IR invariant validation is off because DOTNET_INSPECT_IR_INVARIANTS was set to `0`. That request is honored on purpose. A disarmed run is not a passing run: every pipeline test in this assembly validates nothing. Unset the variable to restore coverage.

Default flipped to `hostOptedOut: true`, variable unset:

> IR invariant validation is off with DOTNET_INSPECT_IR_INVARIANTS unset, so it was declined in code — either the default flipped back to off or a host in this assembly called DisableForShippedTool(), which only src/dotnet-inspect/Program.cs may do. A disarmed run is not a passing run: every pipeline test in this assembly validates nothing.

## Non-blocking: `~120` — taking it

Agreed it belongs here. I was rewriting both sentences, and #3302 asks for the number regardless if the decision is deferred. Measured on this head rather than carried from your comment:

```
DOTNET_INSPECT_IR_INVARIANTS=full --gate fast  → Total: 3893, Failed: 5
```

The five are exactly the ones #3302 names — `SharedThrow_SingleGuardFallenInto_InlinesAsGuardClauses`, `JumpTable_ValueBlocksReturnedAtJoin_RaisesToSwitchExpression`, `TypedConstants_RunAgainAfterInlining_CatchExposedPositions`, `SwitchIterator_WithFallthroughDefaultSideEffect_DeclinesHonestly`, `YieldBreakLoopIterator_WithDispatchSideEffect_DeclinesHonestly`.

Corrected to `5` in both comments, with the measuring command recorded inline and a note that it read `~120` until now — per #3271, so the next reader re-measures instead of trusting it. The #3302 *decision* stays out of scope.

## Correction to my own PR description

You are right that `IsPackable` is not set anywhere, so it defaults to `true`; my "not packable" wording was wrong even though the conclusion held. The real mechanism is that every `dotnet pack` in `ci.yml` and `release.yml` targets `src/dotnet-inspect` directly, so `ILInspector.Decompiler` is never published. Thanks for checking rather than taking it.

## Validation on the fixed head

| Check | Result |
| --- | --- |
| `IrInvariantsHostContractTests` + `IrInvariantCheckTests` | 40 passed, 0 failed |
| `--gate fast` | 3893 passed, 0 failed |
| `DOTNET_INSPECT_IR_INVARIANTS=full --gate fast` | 5 failed (the #3302 five, unchanged) |
| injection matrix | 3/3 caught |
| code-caused disarm message | correct cause named |

Tampering reverted and `git status` confirmed clean before the final gate run. Re-requesting review.

richlander:
CI green on the fixed head `98e984f2`: `test (ubuntu-26.04, linux-x64)` pass (8m48s), `build-net10` pass (2m3s), `csharp-diff-smoke` pass (1m15s). Path-gated jobs skip as before.

Ready to merge, pending your re-run of the injection matrix.

richlander:
Re-reviewed at head `98e984f2` (fix `0dad7fc3` + merge of `main` `0d486700`). **All three findings addressed. Approving.**

I re-ran the full injection matrix rather than reading the diff, and added evasions I had not originally asked for.

### Public-surface census (finding 1)

`GetMethods(...).Where(!IsSpecialName)` → `GetMembers(Public | Static | Instance | DeclaredOnly)` minus property-accessor names. Each row is a member injected into `IrInvariants`, rebuilt, test run, then reverted:

| # | Injected member | Before | After |
|---|---|---|---|
| A | `public static bool SneakyLevel { get; set; }` | passed | **caught** |
| B | `public static bool SneakyField` | passed | **caught** |
| C | `public static void SneakyMethod()` | caught | caught |
| D | `public class Inner { public static void TurnItOff() {} }` | — | **caught** |
| E | non-`static` class + public instance property | — | **caught** (via `.ctor`) |
| F | `public static bool get_Enabled()` | — | n/a, CS0082 |

A is the one that matters: `{ get; set; }` on a `static bool` is *literally* what `Enabled` and `CheckSemantics` were on `main` before #3289. A future `ValidateEverything { get; set; }` would have reintroduced the exact defect this suite exists to prevent, under a green test named for the whole surface. That hole is closed.

D and E were mine. D is the obvious evasion once methods-and-fields are pinned — hide the opt-out one type deeper under a name the census never enumerates; it's caught. E confirms your explicit `BindingFlags.Instance` rationale is load-bearing rather than defensive: drop `static` from the class and the implicit `.ctor` surfaces, so the census fails rather than going quiet. F closes the theoretical hole in the *name-based* accessor filter — the language reserves accessor names, so `accessors.Contains(member.Name)` can't be spoofed by a hand-written method.

### Cause attribution (finding 2)

`WhyValidationIsOff()` now branches on `EnvironmentRequest() is false`. Both branches verified by real runs:

- env set to `0`: *"…was set to '0'. That request is honored on purpose."*
- default flipped to `hostOptedOut: true`, env unset: *"…is off with DOTNET_INSPECT_IR_INVARIANTS unset, so it was declined in code — either the default flipped back to off or a host in this assembly called DisableForShippedTool(), which only src/dotnet-inspect/Program.cs may do."*

The second is the case that previously misreported. It now names both candidate causes and points at the one file allowed to do it — which is the actionable form.

### Stale figure (finding 3)

Corrected in both sites, and you went past the ask: the doc now carries the re-measure command *and* the correction provenance (`it read ~120 until #3303 corrected it`). That's the right shape — the next reader re-measures instead of inheriting another stale constant.

### Evidence reproduced

| Check | Result |
|---|---|
| `IrInvariantsHostContractTests` | 40 passed / 0 failed |
| `DOTNET_INSPECT_IR_INVARIANTS=0` / `=off` | 2 failed (expected) |
| `DOTNET_INSPECT_IR_INVARIANTS=full` | 0 failed |
| `--gate fast` | **3893 / 0** |

### On the risk claims

Both hold, one for a different reason than stated. Removal is genuinely caller-free — whole-solution build is clean. "No API break" is true, but not because the project opts out of packing: `IsPackable` is not set on `ILInspector.Decompiler` or in `Directory.Build.props`, so it's the SDK default `true`. The library is unpublished because every `dotnet pack` in `ci.yml` (`:583`, `:596`, `:606`) and `release.yml` (`:169`, `:199`, `:211`) targets `src/dotnet-inspect` directly. Same conclusion, but it rests on the pack *target*, not a project-level opt-out — worth knowing if anyone ever adds a solution-wide pack step.

No further changes requested.

richlander:
Filed #3330 from the packaging observation in my review here.

To recap why: the "no API break" rationale for removing `EnableSemanticChecks()` was sound, but it rests on every `dotnet pack` targeting `src/dotnet-inspect` directly rather than on `ILInspector.Decompiler` opting out — `IsPackable` is unset there, so it is the SDK default `true`, and `dotnet pack src/ILInspector.Decompiler` does produce a `.nupkg` today. Same conclusion, different premise, and the premise lives in a workflow file rather than the project being reviewed.

PR #3307, opened 2026-07-27
Two reporting gestures for annotations: side and caret (#3304)

### LINKED ISSUE #3304: Two interchangeable annotation reporting gestures: side vs. caret
## Thesis

An annotation is **data** — a fact keyed to a location. How it is **reported** is a separate
choice of **gesture**. Today the two are welded together: every annotation is a trailing `//`
comment, because the print seam hard-codes that gesture.

- **Side** — trailing `// …` to the right of the line. The "interesting, FYI" gesture.
- **Caret** — `^^^^` underline on following `//` lines. The "act on this" gesture.

The bias falls out of the data: *offset-anchored descriptive fact → side; span-carrying
diagnostic → caret.* The gestures are **interchangeable** — the same datum can be promoted
from side to caret when you want to act on it.

## Why now (real output, not a mockup)

`dotnet-inspect member ILInspector.Decompiler.Annotations.AnnotationText Format:1 --library … -S "Annotated Source"`
renders this today:

```csharp
public static string Format(ILInspector.Decompiler.Annotations.Annotation fact)
{
    StringBuilder sb = new(fact.Descriptor.Id);  // alloc.new(StringBuilder; alloc=System.Text.StringBuilder; path=straight-line; path-confidence=dominates-return; post-dominance=return-post-dominates; multiplicity=once; churned=System.Char[])
```

That trailing comment is ~190 characters. The side gesture does not scale to a rich payload,
and there is no way to ask for the one fact you actually care about to be shown prominently.

## Design constraint: annotations must not gain a severity

`Annotations.cs` states the contract explicitly:

> * no severity — annotations describe, they never grade.
> * no Location — keyed to an IL offset (provenance)
> * no code fix — there is nothing to "fix"; the point is to teach.
> The governing contract is POSITIVE-ONLY.

So the gesture **must not** be a property of `IAnnotation`. That contract is not an obstacle
to this work — it is the reason the split is needed. The *grading* ("is this actionable?")
belongs to the reporting layer, chosen per render. This is exactly "split data from reporting".

## Scope

1. `AnnotationGesture { Side, Caret }` — a reporting-layer type. **Not** added to `IAnnotation`.
2. A gesture **selector** resolved at render time (default: everything `Side`).
3. A caret renderer:
   - The injected `//` anchors to the **member declaration column** — one fixed gutter, so the
     eye tracks a single column instead of a ragged staircase following each statement's indent.
   - Annotations carry no character span (only an IL offset), so the underline covers the whole
     trimmed statement. Honest: we underline exactly what we can prove.
   - Long details wrap onto continuation lines aligned under the first detail column.
4. Wire into both existing hard-coded side seams in `ResearchViews.cs`:
   `RenderOverlayStream` (C#-only overlays) and `RenderMixedStream` (IL+C# interleave).
5. Selection surface: promote a chosen fact family (category or descriptor id) to the caret
   gesture, threaded through `MemberProjectionRequest`.

## Invariant

With no gesture selection, output is **byte-identical** to today. Pinned by a test.

## Not in scope

The harness report sites surveyed separately (`ValidityCheck.Report`, `ReturnToSender`,
`UnsupportedNodeReport`, `LibraryReport`) — tracked separately; several need new plumbing to
carry source text/spans into their reports.

richlander:
Closes #3304.

Where a hidden fact is *drawn* is a reporting decision, not a property of the fact. The shipped Annotated Source views hard-coded one gesture — a trailing `//` comment — so a fact with a long detail produced a sliver off the right edge, and there was no way to say "look **here**" about a particular fact.

## Before

A real render of `AnnotationText.Format:1` — the trailing comment runs to **190 columns**:

```csharp
    StringBuilder sb = new(fact.Descriptor.Id);   // alloc.new(StringBuilder; alloc=System.Text.StringBuilder; path=straight-line; path-confidence=dominates-return; post-dominance=return-post-dominates; multiplicity=once; churned=System.Char[])
```

## After

`--focus allocation` promotes that fact to the caret gesture:

```csharp
public static string Format(Annotation fact)
{
    StringBuilder sb = new(fact.Descriptor.Id);
//  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//   alloc.new(StringBuilder; alloc=System.Text.StringBuilder; path=straight-line;
//   path-confidence=dominates-return; post-dominance=return-post-dominates; multiplicity=once;
//   churned=System.Char[])
```

Both depths, from the end-to-end test fixture:

```csharp
public string Pump(int n)
{
    List<object> sink = new();
//  ^^^^^^^^^^^^^^^^^^^^^^^^^^ alloc.new(List<object>;
//                             alloc=System.Collections.Generic.List<System.Object>; path=straight-line;
//                             path-confidence=dominates-return; multiplicity=once;
//                             churned=System.Object[])
    for (int i = 0; i < n; i++)
    {
        sink.Add(new object());
//      ^^^^^^^^^^^^^^^^^^^^^^^ alloc.new(object; alloc=System.Object; path=loop-body;
```

## Design

`AnnotationGestureSelector` picks a gesture per render. `--focus <category|id|id-prefix>` promotes matching facts to the caret; prefixes match on a dotted-segment boundary, so `alloc` selects `alloc.box` but not `allocator.x`. Non-matching facts keep the trailing form, so `--focus` narrows attention without hiding anything.

**Gesture is deliberately not on `IAnnotation`.** The annotation contract is positive-only and carries no severity. Deciding a fact is worth acting on *is* a grade, and grading belongs to the reporting layer — which is exactly the data/reporting split this issue asks for.

**Default output is byte-identical.** With no `--focus` the selector short-circuits through the historical path. Verified by diffing full CLI renders against `main`, not by inspection.

## The geometry, and why there is a marker

Injected comments anchor to the **member declaration** column, so every one in a member shares a single gutter rather than a staircase following nesting depth.

That column sits *below* the projected body: the body is member-relative, and the declaration line plus a uniform four-column indent are added downstream in `FormatSourceWithDeclaration`. A caret comment needs three columns to the left of its first caret for `"// "`, which a statement on the body's base column cannot supply — its carets land three columns right of what they point at. So `AnnotationCaret` marks its lines with `HoistMarker` and the body formatter renders them un-indented, recovering those columns at *every* depth.

This was measured, not assumed: the alternative (gutter at the body base column) was implemented and rendered first, and the three-column offset is visible in its real output.

An annotation is keyed to an IL offset and carries no character span, so a caret underlines the **whole trimmed statement** — exactly what the fact is known to be about, and no more. A span-carrying datum (a compiler diagnostic, as in #3238/#3256) can underline a narrower range; that is a property of the datum, not the gesture.

## Validation

- `AnnotationGestureTests` — 23 cases pinning selector semantics and caret geometry. The hoist geometry tests were run against `hoist: false` first and **3 fail**; they pass with the fix.
- `CommandExecutionTests` end-to-end: `--focus` promotes facts at two different depths, every caret line is a `//` comment (it is spliced into a ` ```csharp ` fence), carets match the preceding statement's column and length exactly, and `HoistMarker` never survives into output.
- Default-render byte-identical diff vs `main`.
- `ILInspector.Decompiler.Tests` 3914, `dotnet-inspect.Tests` 1606 non-slow + 584 `CommandExecutionTests`, `ILInspector.Research.Tests` 132 — all green. Full solution builds clean.
- markdownlint clean on both changed docs.

Copilot-Session: 43dbbaf0-9cf3-43c6-81c4-5f1b68de2738

### DIFF
--- a/docs/design/hidden-fact-annotations.md
+++ b/docs/design/hidden-fact-annotations.md
@@ -95,10 +95,28 @@ prefixes match on a dotted-segment boundary, so `alloc` selects `alloc.box` but
 not `allocator.x`. With no `--focus`, every fact takes the side gesture and the
 output is byte-identical to the pre-gesture renderer.
 
+Promotion never removes a fact, so a `--focus` that matches nothing renders
+exactly like no `--focus` at all. That would make a typo indistinguishable from
+an honest absence, so an unmatched focus reports the families the member does
+have. `--focus` is member-scoped: the sections that carry a caret block —
+Annotated Source, Cost Overlay, Semantics Overlay — are all member sections.
+
 An annotation carries an IL offset and no character span, so a caret underlines
-the **whole trimmed statement** — exactly what the fact is known to be about. A
-span-carrying datum (a compiler diagnostic) can underline a narrower range; that
-is a property of the datum, not of the gesture.
+the **whole trimmed statement** — exactly what the fact is known to be about.
+
+That is the only mode `AnnotationCaret` implements, and deliberately so: no
+annotation can carry a span, so a narrower underline has no data to draw from
+here. A span-carrying datum — a compiler diagnostic — wants a narrow caret, and
+`FidelityCheck.RenderAnnotatedFailure` draws one today with its own geometry.
+These are **two caret renderers, not one substrate**. Unifying them into a
+single span-aware renderer ("no span ⇒ underline the whole statement") is
+tracked as the open question on the follow-up issue; until that is decided,
+neither claims to be the shared one.
+
+Likewise, the gesture here is chosen by `--focus`, not by the shape of the data.
+A data-driven rule ("span-carrying ⇒ caret") cannot exist inside the annotation
+model, because the positive-only contract forbids the span that such a rule
+would key on.
 
 #### One gutter, at the declaration column
 
--- a/docs/design/hidden-fact-annotations.md
+++ b/docs/design/hidden-fact-annotations.md
@@ -77,6 +77,44 @@ settle:
 - **Semantics Overlay** — inter-method behavior/safety facts, such as callees
   with known exception paths or unsafe implementation evidence.
 
+### Two gestures: side and caret
+
+*Where* a fact is drawn is a reporting decision, not a property of the fact. The
+[positive-only contract](#the-positive-only-contract) forbids severity on an
+annotation, and calling a fact "actionable" is a grade — so the choice cannot
+live on `IAnnotation`. It lives in `AnnotationGestureSelector`, chosen per render.
+
+- **Side** (default) — a trailing `//` comment to the right of the statement.
+  Reads as ambient context: "here is something interesting about this line."
+- **Caret** — a `^^^^` underline on `//` lines beneath the statement. Reads as
+  focus: "look *here*." Long details wrap into a readable block instead of the
+  far-right sliver a 190-column trailing comment produces.
+
+`--focus <category|id|id-prefix>` promotes matching facts to the caret gesture;
+prefixes match on a dotted-segment boundary, so `alloc` selects `alloc.box` but
+not `allocator.x`. With no `--focus`, every fact takes the side gesture and the
+output is byte-identical to the pre-gesture renderer.
+
+An annotation carries an IL offset and no character span, so a caret underlines
+the **whole trimmed statement** — exactly what the fact is known to be about. A
+span-carrying datum (a compiler diagnostic) can underline a narrower range; that
+is a property of the datum, not of the gesture.
+
+#### One gutter, at the declaration column
+
+Injected comments anchor to the **member declaration** column rather than to the
+annotated statement's own indent, so the eye tracks a single gutter instead of a
+staircase that follows nesting depth.
+
+That column is below the projected body: the body is member-relative, and the
+declaration line and a uniform four-column indent are added downstream when the
+member is formatted. A caret comment needs three columns to the left of its
+first caret for `"// "`, which a statement on the body's base column cannot
+supply — its carets would sit three columns right of what they point at. So
+`AnnotationCaret` marks its lines with `HoistMarker` and the body formatter
+renders them un-indented. This buys the columns back at *every* depth, which is
+why it is a marker rather than a clamp.
+
 ## Validation: the oracle problem
 
 The decompiler earns trust from **independent ground truth at corpus scale** —
--- a/skills/decompiler/SKILL.md
+++ b/skills/decompiler/SKILL.md
@@ -49,9 +49,10 @@ dnx dotnet-inspect -y -- member Cache --project ./src/App Pump:1 -S "Annotated S
 ```
 
 The value matches an annotation category, an exact id, or a dotted-id prefix on
-a segment boundary (`alloc` selects `alloc.box`, not `allocator.x`). Facts that
-do not match keep the trailing form, so `--focus` narrows attention without
-hiding anything.
+a segment boundary (`alloc` selects `alloc.box`, not `allocator.x`). It
+**promotes, it never filters** — facts that do not match keep the trailing form,
+so `--focus` narrows attention without hiding anything. A focus that matches
+nothing says so and names the families the member does have.
 
 ## Fidelity model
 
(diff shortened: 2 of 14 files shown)


richlander:
## Adversarial review round

Ran two independent adversarial reviews (a correctness pass and a design pass). Six findings, each **verified by measurement before being fixed** — pushed in 42e893ed.

### 1. Width budget computed in the wrong coordinate space — real, measured

The hoist adjustment leaked into the wrap arithmetic: `caretEnd` ran four columns short of the rendered position, so the wrap width was four too generous. A real render:

```text
104  //                             alloc=System.Collections.Generic.List<System.Object>; path=straight-line;
```

104 columns against a `Budget` of 100. Fixed by running all column arithmetic in rendered columns and confining the hoist to caret alignment. The new budget test was run against the old arithmetic and **fails at 103 columns**; passes with the fix. Max line is now 86.

### 2. Marker stripped at one of four return paths

`FormatSourceWithDeclaration` has four returns; only the indent loop stripped `HoistMarker`. The bare-body projection and both expression-bodied returns would have emitted a raw `\u0011` into a ```` ```csharp ```` fence.

I checked reachability rather than assuming: annotated projections render with `preferExpressionBodied` false, so nothing reaches those paths **today** — but that is an unpinned invariant one edit away from breaking. All four paths now strip, via a new `AnnotationCaret.Flatten` for paths that cannot honour a hoist. The body indent is also now driven from `AnnotationCaret.BodyIndentWidth` rather than a second `"    "` literal; the two must agree or the carets shear off the code they point at.

### 3. `--focus` accepted a missing value

`ArgumentArity.ZeroOrOne` meant a bare `--focus` silently disabled focusing. Now `ExactlyOne`:

```text
Error: Required argument missing for option: '--focus'.
```

### 4. `--focus` matched nothing, silently

The sharpest UX finding. Promotion never removes a fact, so an unmatched focus renders **identically** to no focus — making a typo indistinguishable from an honest absence, in a tool whose purpose is teaching.

```text
note: --focus 'alocation' matched no facts here; available: alloc, allocation
```

Descriptor ids have no central registry to validate against, so this reports from the facts actually collected for the render — cheaper than building a registry and more useful than a static list, since it names what *this member* has.

### 5. `--focus` on `type` was dead surface

Registered on `type`, where it can never change any output — every section carrying a caret block (Annotated Source, Cost Overlay, Semantics Overlay) is a **member** section. `type` does not even offer them. Removed, with a test pinning that it is not offered there.

### 6. Two new theory rows were vacuous

The `Cost Overlay` / `Semantics Overlay` marker-leak rows used a fixture with no cost or semantics facts — they asserted the absence of a marker on output containing no carets at all. Repointed at `CostOverlayFixture`, which produces both, and **every row now also asserts a caret actually renders**:

```text
//  ^^^^^^^^^^^^^^^^^^^^^^^^ cost.callee(callee HotCallee: alloc-loop)
//  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ semantics.callee(may-throw FormatException)
```

### 7. The design doc oversold the abstraction — accepted, wording corrected

The design reviewer's strongest point, and it was right: `AnnotationCaret` implements only the whole-statement underline and **takes no span**, so it structurally cannot render `FidelityCheck`'s narrow-span caret. Calling it a shared "gesture substrate" was unearned.

I deliberately did **not** add a speculative span parameter with no consumer. Instead the doc now states plainly that there are two caret renderers, and points at the unification question — which is already the recorded open question on the follow-up issue #3311.

Related: the issue said byte-identity was "pinned by a test". It is not — it is verified by **diffing full CLI renders against `main`**, which I re-ran after every change above and which still passes. The tests pin the absence of carets, not byte-identity.

### Verification

- `ILInspector.Decompiler.Tests` 3917, `dotnet-inspect.Tests` 1606 non-slow + 592 `CommandExecutionTests`, `Research` 132, `CSharp` 346 — all green.
- Default render still **byte-identical to `main`**.
- markdownlint clean.

### Not fixed, deliberately

`--focus` reads as a filter but only promotes. Rather than rename late, the help text, skill doc, and design doc now all say "promotes, never filters" explicitly, and an unmatched focus is no longer silent — which was the actual harm behind the naming concern. Happy to rename if you'd prefer.

PR #3308, opened 2026-07-27
Serialize the full EVIL run partition and stop overclaiming honest (#3244)

### LINKED ISSUE #3244: EVIL history card under-serializes its own partition and overclaims honest
Part of #3079.

## Problem

The tracked artifact for the EVIL benchmark (`history.jsonl`, the source of the
quoted "valid %") cannot represent its own classification, and reports some
measurement failures as though they were data.

### 1. The serialized record is not a partition

`HistoryRunValidDifferent(int Total, int FrontierIlExact, int FrontierIlDiff)`
(`tools/DecompilerHarness/AuthoredCorpusHistoryCard.cs:197`) has three fields,
but `ClassifyTaste` can produce five `ValidDifferent` buckets: `Lowering`,
`KnownTaste`, `FrontierIlExact`, `FrontierIlDiff`, `FrontierIlUnknown`. Only two
are serialized (`AuthoredCorpusBenchmark.cs:322-323`).

The sub-buckets fail to sum in **every recorded run**:

| Date | `total` | `frontierIlExact + frontierIlDiff` | missing |
| --- | --- | --- | --- |
| 2026-07-20 | 5290 | 5278 | **12** |
| 2026-07-23 | 5205 | 5195 | **10** |
| 2026-07-24 | 5202 | 5192 | **10** |
| 2026-07-25 | 5223 | 5213 | **10** |

The top-level partition is exact (12,000 every run), so the error is contained —
but those rows cannot be recovered from the artifact, only from a fresh run.

### 2. Oracle failure is bucketed as data

`FrontierIlUnknown` is the `_ =>` arm reached when `CompileBackStatus` is
neither `Exact` nor `OpcodeDiff`/`OperandDiff` — i.e. **the compile-back oracle
returned no verdict**. Rows where the instrument failed are counted alongside
rows it successfully classified, then dropped from the artifact. This is the
same category error as `ShellOrClosureDefect`: *unmeasured recorded as if
measured.*

### 3. Buckets print only when non-zero

`if (frontierUnknown > 0)`, and likewise `notFull`, `drift`, `unsupported`. A
reader cannot distinguish **measured zero** from **not measured**.

### 4. `honest` overclaims

`bool honest = unmatchedRows == 0 && evaluated > 0;`
(`AuthoredCorpusBenchmark.cs:179`, and again at `:306`). That is an
*inputs-were-present* flag. A run with 4,831 rows whose product status is
unmeasured and 10 undecomposable rows still stamps `honest: true`.

## The fix already exists in this repo

`ValidityCheck.Report` is the house standard — numerator, denominator, and cap
disclosed together, with the caveat printed next to the number
(`tools/DecompilerHarness/ValidityCheck.cs:526-528`):

```
Semantic binding (Full + syntactically-valid): compiled {semChecked} of {semanticEligible} eligible methods (compile-cap {capText})
  NOTE: semantic findings are per-sample, not corpus-wide; ...
```

Apply the same discipline here.

## Proposed change

1. Serialize the **full** partition, with a schema assertion that the buckets
   sum to the total.
2. Give oracle-no-verdict its own named bucket, distinct from classified
   outcomes, and never drop it.
3. Print every bucket unconditionally, including zeros.
4. Rename `honest` to what it checks (e.g. `inputsComplete`), or extend it to
   mean what it says.
5. Add `methodologyVersion` so rows are comparable across changes — the
   `invalidBreakdown` field is already `null` for pre-#3096 rows, which is
   exactly the ambiguity this prevents.

## Acceptance

Sub-buckets sum to `total` for every emitted row, and a run with a
no-verdict result shows it explicitly rather than by arithmetic shortfall.

richlander:
Fixes #3244.

The tracked EVIL trend store could not represent its own classification. `HistoryRunValidDifferent` carried 3 of the 5 buckets `ClassifyTaste` produces, so the sub-buckets failed to sum to the total in **every** recorded row.

## What the shortfall actually was

The issue flagged the 10–12 missing rows per run as recoverable only by re-running. They were recoverable from the archived run JSON, and the answer is unglamorous:

| Row | shortfall | recovered |
| --- | --- | --- |
| 2026-07-23 `d6af0165` | 10 | lowering 6 + knownTaste 4 |
| 2026-07-24 `16c0687f` | 10 | lowering 6 + knownTaste 4 |
| 2026-07-25 `abec2dd7` | 10 | lowering 6 + knownTaste 4 |
| 2026-07-26 `14781e8d` | 10 | lowering 6 + knownTaste 4 |
| 2026-07-20 (baseline) | 12 | **no retained artifact** |

Each artifact was matched against its row on all nine shared fields (`evaluated`, `correct`, `validDifferent.total`, `frontierIlExact`, `frontierIlDiff`, `invalid`, and the three `invalidBreakdown` members) before being used, so this is recovery, not reconstruction. No headline metric moves.

**`frontierIlNoVerdict` is 0 in every recorded run.** Every `ValidDifferent` row carries a compile-back verdict (Exact 3104 / OpcodeDiff 1821 / OperandDiff 301 = 5226 at `14781e8d`). Defect #2 in the issue — oracle failure bucketed as data — is a real design flaw that has never actually fired. Worth stating precisely rather than leaving it implied.

## Changes

1. **Full partition serialized.** `validDifferent` gains `lowering`, `knownTaste`, `frontierIlNoVerdict`; the row gains `notFull` and `unknownOutcome` so the top-level buckets sum to `evaluated`. The three added sub-buckets are nullable: absent means *not recorded*, never zero.
2. **Enforced, not asserted.** `AuthoredCorpusHistoryCardTests` reads the tracked store and fails when a partition does not close, or when any non-grandfathered row omits a bucket. The grandfather list (`2026-07-20`) is asserted as **set equality**, so a later backfill or removal fails too and the list cannot go stale. The benchmark applies the same check to its own output and emits a `BLOCKER:` line rather than a complete-looking payload.
3. **Oracle failure named.** `FrontierIlUnknown` → `FrontierIlNoVerdict`, documented as instrument failure: unmeasured, not "neither exact nor diff". The classifier's outer fallback no longer shares it — an unrecognized probe outcome gets its own `UnknownOutcome` instead of inflating a real bucket, and fails the run.
4. **Every bucket prints unconditionally**, so a measured zero is distinguishable from one never reported.
5. **`honest` → `inputsComplete`**, which is what it computes. It never meant the run was fully measured — it is `true` on a run with 4,808 rows whose product status is unknown. The README now says so next to the field.

Item 5 of the issue (`methodologyVersion`) already landed in #3231.

## Evidence

**Partition closes** in all four backfilled rows and in a real 300-row subset run: top-level `1+132+167 = 300`, valid-different `0+1+33+98+0 = 132`.

**All three history gates trip under tamper**, and pass restored:

| Tamper | Gate that failed |
| --- | --- |
| Append a row dropping the sub-buckets (the original defect) | `TrackedHistory_OnlyGrandfatheredRowsOmitThePartition` |
| Sub-buckets present but not summing to total | `TrackedHistory_CompleteRows_PartitionExactly` |
| Backfill the grandfathered row (stale pin) | `TrackedHistory_OnlyGrandfatheredRowsOmitThePartition` |

**Runtime guard proved non-vacuous** by omitting `invalid` from the sum and rebuilding: `BLOCKER: emitted buckets do not partition the run — validDifferent 132 vs 132, top-level 133 vs 300.` Source restored.

Corpus area 83/83. Fast gate 3894 passed / 0 failed. `markdownlint` clean. Card renders unchanged.

## Boundary

No headline metric changes; `validPct`, `correct`, `invalid`, and `productBodyDefect` are byte-identical for every row. `AuthoredCorpusDrift` has a structurally identical `honest` flag on a different artifact; left alone to keep this scoped, worth a follow-up.

Step 7 of the append procedure is still manual transcription — the gate now catches an omission, but having the harness emit the ready-to-append row would remove the class of error. Non-blocking follow-up.


richlander:
## Review reconciliation (base `0d486700`, reviewed head `21a1036e`, fixed head `4f051aee`)

Two reviewers, MAI-Code and Gemini Pro, each in an isolated worktree at the exact head.

### Both reviewers found the same gap — fixed at the root

MAI-Code raised it BLOCKING, Gemini as a NIT: the partition guard was wired only into the `--json` path, so `WriteReport` could exit 0 on a malformed run. Gemini added the sharper version of the risk — a newly added `TasteBucket` that someone forgets to print would be dropped silently in text mode, since `unknownOutcome` only catches unmapped *enum members*, not mapped ones missing from the arithmetic.

Reproduced and fixed in `4f051aee`, but not by copying the check across. The real defect was that **both paths recomputed every bucket independently** with their own LINQ passes, so the two modes could disagree about what a run contained and nothing would notice; adding a second copy of the guard would have preserved that divergence.

`BucketCensus` now does one pass, counts each bucket exactly once, carries `PartitionClosed` and its failure message, and feeds both renderers; both return through one `ExitCode` helper. `ValidDifferent` is still counted from the probe outcome rather than from its own sub-buckets, so `PartitionClosed` keeps comparing two independently derived numbers instead of a sum against itself.

Verified: text and `--json` report identical counts on a real subset run (correct 1, validDifferent 11, sub-buckets 0/0/4/7/0, invalid 2). Dropping `Invalid` from `TopLevelSum` makes **both** paths emit the same `BLOCKER:` line and exit 1; restored, both pass. Fast gate 3894/0.

### MAI-Code BLOCKING #1 — acknowledged, out of scope, and not achievable from the row

> The gates only enforce arithmetic closure, so a row can be invented wholesale with internally consistent numbers, or have `lowering`/`knownTaste` swapped while preserving the sum.

Both observations are correct, and I'm not changing the gate for them. They describe **authenticity**, which is a different property from the **partition completeness** this PR delivers, and it is not decidable from the row: a swap that preserves the sum is information-theoretically invisible to any checker that only sees the row.

Gemini reached the same conclusion independently while probing the same attack: *"A gate can only check mathematical consistency, not historical truth."*

The suggested fix — anchor each row to an archived artifact — cannot run where the gate runs. This directory's README explicitly keeps the multi-megabyte per-row payloads **out of tree** ("Do not commit full per-row run payloads here"), so CI has no artifact to compare against.

Nothing in the PR claims otherwise; the gate's doc comment is scoped to "a hand-appended row that drops a bucket." The real fix for transcription authenticity is the follow-up already noted in the PR body: have the harness **emit** the ready-to-append row so step 7 stops being manual. That removes the class of error instead of trying to detect it after the fact. I'll file it rather than grow this PR.

### Gemini — no blocking findings

Independently verified and reported clean:

- **Backfill**: extracted all four artifacts with `jq`, confirmed the new values match and that no pre-existing field changed; confirmed the `frontierIlUnknown` → `frontierIlNoVerdict` mapping is faithful.
- **Nullability**: confirmed `System.Text.Json` gives `null` (not `0`) for omitted parameters carrying `= null` defaults, so `IsComplete` is correct. Also confirmed the interlock — supplying `{"lowering":0,"knownTaste":0,"frontierIlNoVerdict":0}` to fake completeness fails the *top-level* sum, because the two gates constrain each other.
- **Grandfather pin stability**: a duplicate date makes the set length 2 and fails; backfilling the row makes it empty and fails.
- **`UnknownOutcome`**: confirmed all five `ReturnToSenderSourceOutcome` members are matched, so the outer arm is unreachable today and serves only future enum additions.
- **Exit code**: confirmed both new conditions are strictly restrictive — they can only turn a passing run into a failing one, never the reverse.
- **Compatibility**: full-repo grep for `honest`; no script or workflow consumes it from this artifact. `AuthoredCorpusDrift` emits a separate artifact and is decoupled.

richlander:
## Fixed-head re-review complete — both clean

Both reviewers re-reviewed the exact fixed head `4f051aee` in isolated worktrees.

| Reviewer | Verdict at `4f051aee` |
| --- | --- |
| MAI-Code | Clean — "No significant issues found" |
| Gemini Pro | Clean — "no blocking defects or regressions" |

### Reconciliation

**Shared finding (Gemini NIT / MAI BLOCKING #2) — fixed.** The partition guard was wired only into the `--json` path, so `WriteReport` could exit 0 on a malformed run. Gemini's framing was sharper and drove the shape of the fix: a developer adding a new `TasteBucket` and forgetting it in the text path would silently drop the row, because `unknownOutcome` only catches unmapped *enum members*, not mapped members missing from the arithmetic.

Rather than copy the check into the second path, I removed the divergence that made disagreement possible. Both paths were independently recomputing every bucket with their own LINQ passes. `BucketCensus` now counts each bucket exactly once in a single pass, exposes `PartitionClosed` / `PartitionFailureMessage`, and feeds both renderers; both return through one `ExitCode(census, unmatchedRows)` helper.

Gemini verified the fix by patching in a new unhandled `TasteBucket` member locally and confirming the harness fails loudly on both branches of that scenario — if the new bucket is conceptually a `ValidDifferent` subtype, both partition equations break; if it is top-level, `unknownOutcome` is non-zero and the exit code trips.

**MAI BLOCKING #1 — declined, with Gemini concurring.** MAI argued the gates are too weak because a fabricated-but-internally-consistent row, or a sum-preserving `lowering`/`knownTaste` swap, would pass. That is true and it is not what these gates are for: authenticity is a different property from partition completeness, and a sum-preserving swap is information-theoretically invisible to a row-only checker. MAI's suggested fix — anchoring rows to archived artifacts — cannot run where the gate runs, because the README deliberately keeps per-row payloads out of tree. Gemini probed the same attack independently and reached the same conclusion.

The durable fix is to stop transcribing rows by hand: **#3309** tracks having the harness emit the ready-to-append row directly. Non-blocking follow-up.

**Over-claiming check.** I asked both reviewers whether anything in the code, tests, or README claims a property stronger than partition completeness. Gemini checked the README prose and the XML docs on `TrackedHistory_CompleteRows_PartitionExactly` and found the wording strictly bounded to the arithmetic. No change needed.

**Non-circularity, confirmed.** `ValidDifferent` is counted from `result.Outcome`, never from the sum of its own sub-buckets, so `PartitionClosed` compares two independently derived numbers. Gemini traced both data paths in `Census()` and confirmed the assertion is genuine and non-vacuous.

### Validation on the fixed head

- Text and `--json` produce identical numbers on the same input (both reviewers ran the 300-row subset independently).
- Tampering `TopLevelSum` makes both paths emit the same `BLOCKER:` line and exit 1.
- Fast gate 3894 passed / 0 failed; Corpus area 83/83; card tests 17/17.
- CI green: `test`, `markdownlint`, `changes`.

Ready to merge

Issue #3272, opened 2026-07-27
Decompiler: raise dup-chain object initializers materialized as named-local version copies


richlander:
## Summary

The decompiler leaves compiler object-initializer lowerings in their expanded,
temp-threaded form instead of raising them back to `new T { X = a, ... }`, even
when the IL is the **canonical dup-chain initializer shape** that
`ObjectInitializerPass` is designed to fold. The failure is specific to bodies
where the pipeline materialized the dup chain into **named-local version copies**
before the pass ran.

## Repro

`OpenAI.OpenAIClient.GetRealtimeClient` (package `OpenAI` 2.12.0):

```bash
dotnet-inspect member OpenAI.OpenAIClient GetRealtimeClient -S "Decompiled Source" --package OpenAI
```

Current output:

```csharp
public virtual OpenAI.Realtime.RealtimeClient GetRealtimeClient()
{
    RealtimeClientOptions S_256 = new RealtimeClientOptions();
    ClientPipeline S_257 = Pipeline;
    S_256.Endpoint = _options.Endpoint;
    RealtimeClientOptions S_258 = S_256;
    S_258.OrganizationId = _options.OrganizationId;
    RealtimeClientOptions S_259 = S_258;
    S_259.ProjectId = _options.ProjectId;
    RealtimeClientOptions S_260 = S_259;
    S_260.UserAgentApplicationId = _options.UserAgentApplicationId;
    return new RealtimeClient(S_257, S_260);
}
```

Expected (raised):

```csharp
public virtual RealtimeClient GetRealtimeClient()
    => new RealtimeClient(Pipeline, new RealtimeClientOptions
    {
        Endpoint = _options.Endpoint,
        OrganizationId = _options.OrganizationId,
        ProjectId = _options.ProjectId,
        UserAgentApplicationId = _options.UserAgentApplicationId,
    });
```

## The IL is the textbook dup-chain initializer lowering

```il
IL_0000: ldarg.0
IL_0001: call     OpenAI.OpenAIClient::get_Pipeline()     // arg0: Pipeline, evaluated first
IL_0006: newobj   RealtimeClientOptions::.ctor()
IL_000B: dup
IL_000C: ldarg.0
IL_000D: ldfld    _options
IL_0012: callvirt get_Endpoint()
IL_0017: callvirt set_Endpoint(System.Uri)
IL_001C: dup ... set_OrganizationId
IL_002D: dup ... set_ProjectId
[...]

PR #3310, opened 2026-07-27
Preserve accessor SourceOffset in PropertySugarPass; register two-arg diff (#3272)

richlander:
- Advances #3272 (follow-up to #3290)
- Preserves accessor `SourceOffset` in `PropertySugarPass` so the object-initializer trailing-arg guard folds when a *leading* ctor argument is a property access
- Registers the `MakeConsumerWithTwoLeadingArgs` known diff that #3290 omitted

## Change

**Conclusion:** **PASS** — additive, provenance-only fix that closes a false-negative in the guard shipped by #3290 (static/instance-property leading argument no longer blocks the fold) and lands the docket entry #3290 forgot; no printed output changes and lowered output is byte-identical for every corpus method.

### Context

#3290 taught `ObjectInitializerPass` to raise a trailing object initializer in ctor-argument position (`new C(arg, new T { ... })`) and added an IL-offset provenance guard (`ExecutesBefore`/`EffectOffset`) that only folds statements whose computation completed **before** the `newobj`. That guard is already in `main`.

Adversarial re-review of #3290 (GPT-5.6-sol) found a residual **false-negative**: when a *leading* ctor argument is a property access (e.g. `new C(StaticTag, new T { X = a })`), `PropertySugarPass` replaced the getter `Call` with a `LoadProperty` node **without preserving its `SourceOffset`**. For a static property the value subtree then retains no offsets, `EffectOffset` returns `-1`, and the guard conservatively over-rejects — so the initializer stays lowered even though the fold is safe.

### Fix

`PropertySugarPass` now calls `InheritSourceOffset(call)` on the replacement `LoadProperty` (getter), `StoreProperty` (setter), and `EventSubscription` (event) nodes before `ReplaceWith` — matching the ~30 other rewrite passes that already do this. `InheritSourceOffset` (`IrNode.cs`) only fills when the offset is currently `<0`, so the change is purely provenance-additive: offsets are not printed, so no rendered output changes; it only lets the guard see the real pre-`newobj` offset and fold.

With the fix, `new C(<property>, new T { ... })` folds; the reorder guard from #3290 is unchanged and still declines genuine reorders.

### Docket

#3290 merged without registering its own `MakeConsumerWithTwoLeadingArgs` fixture in `LoweredFidelityGateTests.KnownDiffs`, leaving the `Speed=Slow` fidelity gate red for that one fixture. This PR registers it with an honest comment (two-arg fold is Valid+Correct but recompiles with an OpcodeDiff — a single-temp `ExpressionInliningPass` limitation, not a regression; the single-arg corpus case is byte-Exact). The 6 pre-existing drift fixtures unrelated to #3272 are intentionally **not** registered here (separate issue).

## Evidence

| Check | Baseline (`origin/main`) | Head |
| --- | --- | --- |
| `ObjectInitializerPassTests` | 34 passed | 35 passed (new `TrailingInitializerWithStaticPropertyArgument_FoldsViaUseSite`) |
| Decompiler fast suite (`-trait- Speed=Slow`) | 3891, 0 failed | 3892, 0 failed |
| Static-property leading arg folds | No (stays lowered) | Yes (`new InitConsumer(StaticTag, new InitTarget { X = a })`) |
| Rendered output / lowered bytes | — | Byte-identical (provenance-only) |

The one fast-suite failure observed initially (`ExpressionTreeLambdaTests.LookalikeExpressionAssembly_StaysFactoryCalls`) was a missing fixture binary; it passes after `dotnet build dotnet-inspect.slnx -c Release`, and is unrelated to this change.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.ObjectInitializerPassTests
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -2941,6 +2941,8 @@ public class CfgSampleClass
 
     static void SideEffect() { }
 
+    static int StaticTag { get; set; }
+
     // #3272 regression: the object initializer is the SECOND constructor argument,
     // so the first argument is evaluated first and stays live on the stack beneath
     // the dup chain. The stackifier cannot keep two values on the stack across the
@@ -2970,6 +2972,15 @@ public class CfgSampleClass
         t.X = a;
         return t;
     }
+
+    // #3272 provenance robustness: the leading constructor argument is a STATIC
+    // PROPERTY read (`get_StaticTag`). PropertySugarPass rewrites the getter Call
+    // into a zero-child LoadProperty; if that rewrite drops the Call's SourceOffset,
+    // the object-initializer skip guard cannot prove the spill ran before the
+    // `newobj` and declines the fold. The getter runs before the `newobj`, so this
+    // must fold to `new InitConsumer(StaticTag, new InitTarget { X = a })`.
+    public static InitConsumer MakeConsumerWithStaticPropertyArg(int a)
+        => new InitConsumer(StaticTag, new InitTarget { X = a });
 
     public static InitContainer MakeNestedObject(int a, int b)
         => new InitContainer { Inner = { X = a, Y = b } };
--- a/src/ILInspector.Decompiler.Tests/LoweredFidelityGateTests.cs
+++ b/src/ILInspector.Decompiler.Tests/LoweredFidelityGateTests.cs
@@ -150,6 +150,15 @@ public class LoweredFidelityGateTests
         "DayNumber",
         "DoubleViaLocalFunction",
         "StaticLocalFunctionCalledTwice",
+        // MakeConsumerWithTwoLeadingArgs (#3272): the trailing object initializer
+        // with TWO preceding constructor arguments folds correctly (Valid + Correct)
+        // to `new InitConsumer3(Identity(tag), Identity(a), new InitTarget { ... })`,
+        // but ExpressionInliningPass only removes one of the two single-use spill
+        // temps, leaving an extra stloc/ldloc pair on compile-back. Honest OpcodeDiff
+        // from a separate inliner limitation, not an object-initializer regression
+        // (the pre-raise version-copy expansion was also non-exact). The single-arg
+        // corpus case (OpenAI GetRealtimeClient) stays byte-exact.
+        "MakeConsumerWithTwoLeadingArgs",
     };
 
     /// <summary>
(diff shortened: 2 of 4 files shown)


richlander:

Two independent adversarial reviews at the fixed head `bdbe28dc803dbc8e0c0c138ced2dab1792af2701` (base `0d48670065f1567b536eff2d545b81e6a32167e3`), each in an isolated detached worktree with real runs.

| --- | --- |
| GPT-5.6-sol | **CLEAR** |
| Gemini-3.1-pro | **CLEAR** |

Both built the Release solution, ran `ObjectInitializerPassTests` (35/35) and the fast decompiler suite (3892, 0 failed), and confirmed the static-property case compiles back **Exact**.

### Attack points, both reviewers agree

1. **Offset-inheritance soundness / no reorder re-introduction.** Both constructed the adversarial `new T(); x = SideEffectingProperty; t.X = …` shape and confirmed the inherited getter offset is *after* the `newobj`, so the #3290 guard correctly **declines** the fold — the reorder bug #3290 fixed is not reintroduced. The existing real negative `MakeTargetWithVoidCallBetween` stays lowered. Gemini additionally verified by reverting the fix that the regression reproduces (static fails to fold) and that the *instance*-property case already folded on base (its receiver load retained an offset), so only the zero-child static case was the false negative.
2. **`InheritSourceOffset` semantics.** Both confirmed it writes only when the current offset is `<0` (`IrNode.cs`), and the fresh `LoadProperty`/`StoreProperty`/`EventSubscription` nodes carry the default `-1`, so the calls inherit rather than no-op — additive, provenance-only.
3. **Blast radius over other `SourceOffset` consumers.** Both grepped every product consumer and found no unsafe decision introduced; populating the accessor offset is conservative or purely beneficial (object-initializer ordering, branch-target label preservation, annotation anchoring).
4. **Docket honesty.** Both verified only `MakeConsumerWithTwoLeadingArgs` was added to `KnownDiffs`, and that its OpcodeDiff is exactly one extra `stloc/ldloc` from an `ExpressionInliningPass` multi-spill limitation — the initializer itself folds (`new InitConsumer3(S_257, S_258, new InitTarget { X = a, Y = b })`), so it is a pre-existing, independent limitation, not an object-initializer regression. No unrelated drift fixtures smuggled in.
5. **Rendered-output claim.** Ordinary C# does not print node offsets; the static-property render recompiles Exact. GPT noted the precise qualification that `SourceOffset` can indirectly affect unresolved `IL_xxxx` label placement / annotation anchoring, but found no incorrect output change — this is the intended provenance-preservation, and it only ever *adds back* a correct label the getter legitimately targeted.

No blocking findings from either reviewer; no code changes were required after review.

CI: `test`, `build-net10`, and `csharp-diff-smoke` pass (`il-diff-smoke`/`markdownlint`/`pack` are path-gated skips).

Ready to merge

Issue #3292, opened 2026-07-27
Graph UX: generic command architecture and explicit Integration sets


richlander:
## Summary

Establish a generic top-level graph command architecture, with explicit
Integration sets as its first package-workspace mode.

The original issue framed this as call-graph depth and multi-library scope.
The architecture work since then showed that the command boundary is broader:
Integration, reference, implementation, performance, and call relationships
all need typed graph requests over one inspection space. They should not grow
independent graph flag surfaces on subject commands.

## First slice

Add:

```text
dotnet-inspect graph integrations \
  --package A[@version] \
  --package B[@version] \
  --tfm <target-framework>
```

The command:

- realizes the repeated package coordinates into one binding-consistent
  workspace context group;
- turns the exact realized package identities into a finite explicit induced
  subject set;
- selects typed Integration-catalog relationships, defaulting to
  `api.extension`, `integration.observed`, and `integration.opportunity`;
- admits a logical relationship only when both semantic endpoints are in the
  selected package subject closure;
- preserves the sequential executor as the deterministic,
  single-threaded-Browser/Wasm-compatible baseline;
- carries typed acquisition, binding, and producer failures beside partial
  topology and exits nonzero when evidence is incomplete;
- renders the same ordered logical-edge rows as Markdown, tree, Mermaid,
  table, TSV, JSONL, JSON, or count output, including consistent `--rows`
  selection.

## Command shape

`graph integrations` is intentionally an induced-set command:

- `--package` repeats and accepts `name` or `name@version`;
- one required `--tfm` applies to the complete set;
- `--relationship` repeats and accepts only current Integration graph catalog
  ids;
- there is no primary seed, peer role, direction, or depth.

This keeps graph-specific acquisition, relationship selection, row shaping,
and rendering out of `member`, while leaving room for future sibling commands
such as call or implementation neighborhoods.

## Architecture boundary

The CLI is a host and presentation layer. It composes
`WorkspaceContextLoader`, `InspectionGraphInducedSetRequest`, and
`InspectionGraphIntegrationsQuery`; it does not introduce another graph IR,
infer identity from display labels, duplicate producer planning, or acquire an
implicit dependency closure.

Package groups remain presentation context. Nodes and edges retain their typed
query identities. Distinct typed edges with equal rendered labels must remain
distinct through row selection.

## Non-goals and residual

- Demos and screenshots are owned by separate work.
- Local `.nupkg` package-set inputs are deferred.
[...]

PR #3312, opened 2026-07-27
Collapse Call Graph and Caller Graph into one bidirectional graph section (#3280)

### LINKED ISSUE #3280: Collapse the two member call-graph sections into one bidirectional section via a projector plus Markout graph shape

richlander:
Closes #3280.

## What changed

The member document had two call-graph sections answering halves of the same
question from the same indexes: `Call Graph` (outbound callees) and
`Caller Graph` (inbound callers). There is now **one bidirectional `Call Graph`
section**, modelled as a Markout `Graph` rather than a pre-rendered tree.

Because it is a graph, each sink picks its own lowering:

| Sink | Lowering |
| --- | --- |
| Markdown, plain text | tree rooted at the focus |
| `--table` / `--tsv` / `--jsonl` | edge table, one row per edge |
| Mermaid | flowchart |

The old sections could only ever be trees.

## Layering

This completes the three-layer pipeline the design calls for:

```text
CallGraphProjection (data)  ->  Markout Graph (format)  ->  CLI adapter (vocabulary)
```

`CallGraphSectionAdapter` is layer three and is the only place that knows
call-graph vocabulary — member spelling, the `(external)` and `…` markers, and
the `--fields` cue annotations, all moved out of `ApiOutputFormatter`.

`CallGraphProjection` gains `CallTreePerf` so those annotations have their
analysis facts without the CLI re-walking the tree. Perf is analysis data, not
presentation, so it belongs in that layer.

**`ILInspector.CallGraph.CallGraphMermaid` is deleted.** It had no product
consumer, and a Mermaid emitter in ILInspector contradicts the rule that the
layer knows data and not formats. Hosts render their own Mermaid from the
projection; the CLI does so through Markout. Its semantics-only coverage
(cross-assembly identity, inverted loop-edge labels) moved to
`CallGraphProjectionTests`; the rest asserted Mermaid text that Markout now owns
and tests.

## Behavior

- **Breaking**: `-S "Caller Graph"` no longer resolves. No compat alias, by
  design; the callers it named are in `Call Graph`.
- **Parity preserved**: `maxDepth: 3`, `maxNodes: 25`, and the
  `--bin`/`--project`/`--caller-package` caller scopes are unchanged. No new flags.
- Empty state is now `No inbound callers or outbound calls found for this method.`
  The old note only spoke for the callee half.
- `member --json` still bypasses the section pipeline entirely, for **every**
  section. Pre-existing and unchanged here.

Sample output:

```text
## Call Graph

├─ **CallGraphProjection.FromCallers(CallTreeNode) (fanout 2, fanin 1, depth 4)**
│  ├─ ArgumentNullException.ThrowIfNull(object, string) (external)
│  └─ CallGraphProjection.Create(CallTreeNode, CallTreeNode) (…, fanout 21, fanin 3, depth 3)
│     └─ ...
└─ CallGraphMermaid.RenderCallers(CallTreeNode)
   └─ ↩ **CallGraphProjection.FromCallers(CallTreeNode) (fanout 2, fanin 1, depth 4)**
```

## Dependency status — do not merge yet

The second commit is marked `TEMPORARY:`. It points `Markout` at an unreleased
0.29.0 in a local feed, because this needs markout''s `WriteGraph` primitive and
its source-generator support (richlander/markout#163 / PR #165), not yet on NuGet.

Proving the consumption side is the point of that workflow, and it paid: it found
two markout defects before markout shipped them — missing source-generator support
for a `Graph`-typed `[MarkoutSection]` property (the whole declarative path was
unusable), and Markdown lowering graphs to an edge table, which repeated every
200-character member signature on both sides of every edge.

**Before merge**: revert the `TEMPORARY:` commit and set `Markout` to the
published 0.29.0.

## Validation

- `dotnet build dotnet-inspect.slnx -c Release` — succeeded.
- `dotnet run --project src/dotnet-inspect.Tests -c Release` — 2189 total, 1 failed
  (`LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`,
  pre-existing on `main`).
- `dotnet run --project src/ILInspector.Analysis.Tests -c Release` — 519 total, 1
  failed (`CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition`,
  pre-existing on `main`).
- Hand-verified the rendered section in Markdown, `--table`, `--tsv`, and `--jsonl`.
- `markdownlint` clean on every changed Markdown file.

## Follow-up

Depth > 1, multi-library scope, and a `--mermaid` option for `member` are #3292.


richlander:

Two fixed-head reviews on isolated worktrees at `ee85729c`: `gpt-5.6-sol` and
`gemini-3.1-pro-preview`. Both independently found the same real defect. Fixed
in `d8faa559`.

### Blocking — fixed: caller and callee cues were never merged

Both reviewers found this from different angles, and both were right.

Neither walk sees the whole member. `BuildCallerTree` indexes the **caller**
scope and reports fan-in, the root classification and cross-assembly source, but
hard-codes fan-out to `0`. `BuildCallTree` indexes the **callee** scope and
reports fan-out, but never sets `RootKind`. The projection picked one record per
node — callee-first for the focus, first-arrival elsewhere — so every node
inherited one walk''s blind spots. Three observable consequences, all reproduced
before fixing:

**1. The focus never showed its root classification** (`gemini`, medium).

```text
before: **CallGraphProjection.Create(...) (fanin 2, fanout 21, depth 3)**
after:  **CallGraphProjection.Create(...) (target, fanin 2, fanout 22, depth 3)**
```

`--fields Root` printed nothing where `Caller Graph` on `0d486700` printed
`target`. The callee root always won and callee roots never set `RootKind`.

**2. The focus reported fan-in measured against the wrong scope set** (`gpt`,
blocking). The callee root''s fan-in counts inbound edges over the *callee* scope
index, not the caller scope index the section is asking about:

```text
before: **MemberCommand.ExecuteAsync(...) (fanin 4)**   [with 23 caller roots below it]
after:  **MemberCommand.ExecuteAsync(...) (fanin 62)**
```

**3. Any member reached by both walks lost its fan-out** (`gemini`, high).
`WalkCallers` runs first, so a shared node was pinned with the caller walk''s
hard-coded `0`. Reproduced on the reviewer''s exact `Focus -> MethodB -> MethodA
-> Focus` cycle; `MethodA` now correctly reports `fanout 1, fanin 2`.

`MergePerf` combines the two observations field by field, keeping the side that
measured each fact. Degrees and depth are lower bounds over whichever scope set
a walk indexed, so the larger observation wins and a direction that never
measures a degree can never pin it to zero. Signals fall back by reference to the
`None` singleton, which distinguishes "not measured" from "none found".

Three regression tests pin the semantics
(`CallGraphProjectionTests`, now 30 tests): focus merge, shared-node merge, and
a bare boundary occurrence not erasing a measured one.

### Not actioned: tree lowering of a cyclic bidirectional graph

`gemini` (high): in a cycle, a caller reachable through the callee side is
printed under the callee side, so the inbound chain reads as two fragments.
Reviewer suggested keeping two `Graph` instances for tree sinks.

Reproduced, and the behavior is real — but it is neither data loss nor fixable
the suggested way. On the reviewer''s exact fixture:

```text
├─ **Cyc.C.Focus() (fanout 1, fanin 1, depth 4, target)**
│  └─ Cyc.C.MethodB() (fanout 1, fanin 1, depth 3)
│     └─ Cyc.C.MethodA() (fanout 1, fanin 2, depth 3)
│        └─ ↩ **Cyc.C.Focus() ...**
└─ Cyc.C.Main() (entrypoint)
   └─ ↩ Cyc.C.MethodA() ...
```

Every node appears exactly once and every edge is present — the `↩` leaf *is*
the edge, so `Main → MethodA` and `MethodA → Focus` are both shown. The
edge-table lowering is fully lossless and one flag away:

```text
From                  To
Cyc.C.MethodA()       Cyc.C.Focus()
Cyc.C.Main()          Cyc.C.MethodA()
Cyc.C.MethodB()       Cyc.C.MethodA()
Cyc.C.Focus()         Cyc.C.MethodB()
```

A graph with a cycle through the focus cannot be drawn as a tree without
breaking it somewhere; that is a property of tree lowering, not of the
projection. Keeping two graphs for tree sinks would reintroduce exactly the
duplicated walk this design exists to remove, and would mean the "one model,
many sinks" contract held for every sink except the default one. Documented in
`docs/design/call-graph-projection.md` instead.

### Verified clean by both reviewers

Graph identity and dense id assignment (no collisions, generics safe), section
and schema plumbing, `-D` discovery, `ProjectionDiagnostics`, `--fields`
projection across scope flags, and — absent a cycle — correct caller/callee
direction through the `↩` revisit marker.

### Validation at `d8faa559`

- `dotnet build dotnet-inspect.slnx -c Release` — 0 warnings, 0 errors.
- `dotnet run --project src/dotnet-inspect.Tests -c Release` — 2189 total,
  1 failed (pre-existing `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`).
- `dotnet run --project src/ILInspector.Analysis.Tests -c Release` — 522 total,
  1 failed (pre-existing `CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition`).
- `markdownlint` clean.

### Still blocking merge

`ee85729c` (`TEMPORARY: consume unreleased Markout 0.29.0 from a local feed`)
must be reverted once richlander/markout#165 ships to NuGet. CI failures on this
PR are entirely that local feed path not existing on the runners. Not ready to
merge until then.

richlander:
## Round 2 reconciliation — `d8faa559` → `54feb533`

Two fixed-head re-reviews (`gpt-5.6-sol`, `gemini-3.1-pro-preview`) on isolated
worktrees at `d8faa559`. Both confirmed the `MergePerf` fix works and both
independently found that it was **incomplete**. Fixed in `54feb533`.

### Blocking — fixed: `Fanin` was merging two different units

I flagged this as the open risk when requesting the re-review, and both
reviewers confirmed it.

`BuildCallTree` derived fan-in from a dictionary that counted **call sites**
(`DirectCalls.GroupBy(callee).Count()`), while both caller builders dedupe to one
edge per **distinct caller** (`GroupBy(call => call.Caller...)`) before counting.
`Math.Max` over those publishes whichever number is larger under a single label
that then means neither.

`gpt-5.6-sol` gave the crisp repro: `MemberCommand.ExecuteAsync` has four inbound
call sites from three distinct callers, because `TryExecuteFindIfMissAsync` calls
it twice.

```text
before: **MemberCommand.ExecuteAsync(...) (fanin 4)**
after:  **MemberCommand.ExecuteAsync(...) (fanin 3)**
```

Fixed at the source rather than in the merge: both callee-side counters now count
distinct callers. Fan-in is a leverage cue — how many members depend on this one —
and the reverse graph draws one edge per distinct caller, so the annotation has to
agree with the picture it annotates. With units matched, `Math.Max` is meaningful:
each side is a lower bound over whichever scope set that walk indexed, so the
larger observation is the better-informed one.

`gemini`''s cross-assembly variant resolves correctly under this rule too — its
"5 internal call sites from a single method" is 1 distinct caller, so the caller
walk''s 3 wins rather than losing to 5.

New test `LibraryBodyIndexTests.CallTreeFanInCountsDistinctCallersNotCallSites`
pins it over an emitted assembly with three call sites from two distinct callers.

### Also fixed: one of my three new tests was not evidence

`gpt-5.6-sol` (attack point 8) verified my tests against the old code and found
the third would have **passed** under `??=`, because its cue-bearing occurrence
already arrived first. Correct, and a fair hit. The bare occurrence now arrives
first, so the test fails without the merge. The other two were confirmed to fail
against the old code.

### `MaxDepth` — reviewers split; keeping `Math.Max`, documentation sharpened

`gemini` called merging upstream and downstream depth an incomparable scalar
(medium). `gpt-5.6-sol` checked the same point and concluded no defect: both
sides compute bounded subtree height from the node, so the maximum is the
greatest observed radius, and neither the adapter nor the design doc ever claimed
an end-to-end path length.

I agree with `gpt` on the facts. Both numbers are "height of the bounded subtree
rooted here"; the merge publishes the taller one. Note the old design published
*both* meanings under the same `depth` field name in two separate sections, so
collapsing forces one number regardless. `docs/design/call-graph-projection.md`
now states the unit of each cue explicitly rather than leaving it implicit.

### Not actioned — `gemini`''s claim that the tree drops `Main` is factually wrong

`gemini` escalated its round-1 cycle finding, asserting my documentation is false
because "the default tree rendering completely omits `Main`".

That is refuted by running its own fixture at this head:

```text
├─ **Cyc.C.Focus() (fanout 1, fanin 1, depth 4, target)**
│  └─ Cyc.C.MethodB() (fanout 1, fanin 1, depth 3)
│     └─ Cyc.C.MethodA() (fanout 1, fanin 2, depth 3)
│        └─ ↩ **Cyc.C.Focus() ...**
└─ Cyc.C.Main() (entrypoint)
   └─ ↩ Cyc.C.MethodA() (fanout 1, fanin 2, depth 3)
```

`Main` is present, as an additional root, with its edge to `MethodA` rendered as
the `↩` revisit leaf. This also contradicts `gemini`''s own round-1 report, which
described exactly this output ("the entrypoint was fragmented into an isolated
root (`Main() -> ↩ Cyc.C.MethodA()`)"). The round-1 characterisation was right and
the round-2 escalation is not.

So the documented claim — every node and every edge is rendered, the caller chain
merely reads as two fragments — stands as written. The disposition is unchanged:
splitting into two graphs for tree sinks would restore the duplicated walk this
design exists to remove.

### Confirmed clean by both reviewers this round

- `Signals` reference test correctly distinguishes "measured as zero" from the
  unmeasured `None` singleton (`gemini` verified a measured-but-value-equal
  instance is reachable and handled).
- `first with { ... }` covers all eight `CallTreePerf` fields; nothing silently
  inherited.
- Edge labels still come from raw per-edge perf, not merged node perf;
  `LoopAnnotationSurvivesEdgeInversionOnTheCallerSide` passes.
- Merge order: deterministic for every field that can carry conflicting real
  values; focus registers first, so `target` always wins.
- No boundary over-merge: identity is assembly-qualified, `Source` is the
  invariant definition assembly.

### Validation at `54feb533`

- `dotnet build dotnet-inspect.slnx -c Release` — 0 warnings, 0 errors.
- `dotnet run --project src/dotnet-inspect.Tests -c Release` — 2189 total, 1 failed
  (pre-existing `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`),
  10 skipped. No test anywhere pinned a call-site fan-in.
- `dotnet run --project src/ILInspector.Analysis.Tests -c Release` — 523 total,
  1 failed (pre-existing `CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition`).
- `markdownlint` clean.

### Still blocking merge

`ee85729c` (`TEMPORARY: consume unreleased Markout 0.29.0 from a local feed`) must
be reverted once richlander/markout#165 ships to NuGet. CI failures here are
entirely that local feed path not existing on the runners.

PR #3313, opened 2026-07-27
Identify the crashed method body in --gaps PASS BUG output (#3243)

### LINKED ISSUE #3243: --gaps PASS BUG output drops method identity at 2 of 4 sites
## Problem

`--gaps` reports a crashed method four different ways. Two sites include the
method identity, two do not:

```
tools/DecompilerHarness/Program.cs:847   PASS BUG: {ex.GetType().Name}: {ex.Message}
tools/DecompilerHarness/Program.cs:960   PASS BUG: {ex.GetType().Name}: {ex.Message}
tools/DecompilerHarness/Program.cs:1035  PASS BUG: {ex.GetType().Name}: {ex.Message} ({typeName}::{methodName})
tools/DecompilerHarness/Program.cs:1114  PASS BUG: {ex.GetType().Name}: {ex.Message} ({typeName}::{methodName})
```

Lines 1035 and 1114 already do the right thing; 847 and 960 do not.

## Why it matters

This is not hypothetical. The crash fixed in #3236 surfaced from a
364,338-method sweep as exactly this, and nothing else:

```
PASS BUG: KeyNotFoundException: The given key '3' was not present in the dictionary.
```

No type, no method, no assembly. Locating it required bisecting the assembly
pool by hand — and the method needed a two-assembly reference closure to
reproduce at all, so narrowing was not a single-assembly search. The identity
was available at the call site the whole time.

## Fix

Add `({typeName}::{methodName})` at lines 847 and 960 to match the sibling
sites. Two lines.

Worth also confirming the assembly is identifiable when a sweep spans many
assemblies, since a type/method pair alone can be ambiguous across a large pool.

richlander:
## What

`--gaps` PASS BUG output now identifies the exact crashed method body at all
four emit sites, and the diagnostic has one owner instead of four inline
format strings.

Fixes #3243.

## Why

The issue reports that two of the four PASS BUG sites in the harness drop the
method identity entirely. This is not hypothetical: the crash fixed by #3236
surfaced from a 364,338-method sweep as nothing but
`PASS BUG: KeyNotFoundException: The given key '3' was not present in the
dictionary.` — no assembly, no type, no method. Locating it took a manual
bisect of the assembly pool.

The issue also asks whether a type/method pair alone is enough across a large
pool. It is not. A sweep spans many assemblies, an assembly can carry
overloads with the same `Type::Method` spelling, and the same
assembly-relative name can appear at more than one version. So the identity
this PR emits is the one that actually names a single method body:

`<assemblyPath>!<Type>::<Method><signature> [token 0x........]`

## What changed

- New `tools/DecompilerHarness/PassBugDiagnostic.cs` owns the one PASS BUG
  format, reusing `CorpusMethodIdentity.SignatureText` for the signature so
  the spelling matches the harness's existing method identity.
- All four `Program.cs` sites plus `SlotResidualCensus` and `SlotUnifierCensus`
  call it. The two sites that already had `(Type::Method)` are folded onto the
  same formatter rather than left as a second spelling.

## Evidence

`PassBugDiagnosticTests` is the gate for both claims this PR makes:

- `Format_IdentifiesAssemblyAndExactMethodBody` pins the exact rendered text
  and asserts that changing *only* the assembly path, *only* the overload
  signature, or *only* the metadata token produces a different diagnostic —
  i.e. the identity discriminates the axes that a large sweep collides on.
- `DirectPassBugEmitters_UseSharedFormatter` asserts set equality on the
  `PassBugDiagnostic.Format(` call sites across the harness *and* that no
  harness file other than `PassBugDiagnostic.cs` contains the literal
  `PASS BUG`. A new inline emitter or a deleted call site both fail the gate,
  so the "one owner" property cannot silently rot.

```console
$ dotnet build dotnet-inspect.slnx -c Release
0 Error(s)
$ dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -trait- "Speed=Slow"
Total: 3895, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0
$ dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
    -class ILInspector.Decompiler.Tests.PassBugDiagnosticTests
Total: 2, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0
```

## Boundary

Harness diagnostic text only. No product path, no default CLI output, no
decompiler behavior, and no change to crash counting — a PASS BUG still
increments `crashes` and continues.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ILInspector.Decompiler.Tests.csproj
+++ b/src/ILInspector.Decompiler.Tests/ILInspector.Decompiler.Tests.csproj
@@ -76,6 +76,8 @@
              Link="CorpusMetadata.cs" />
     <Compile Include="..\..\tools\DecompilerHarness\CorpusMethodIdentity.cs"
              Link="CorpusMethodIdentity.cs" />
+    <Compile Include="..\..\tools\DecompilerHarness\PassBugDiagnostic.cs"
+             Link="PassBugDiagnostic.cs" />
     <Compile Include="..\..\tools\DecompilerHarness\CorpusMethodDelta.cs"
              Link="CorpusMethodDelta.cs" />
     <Compile Include="..\..\tools\DecompilerHarness\CorpusSensor.cs"
--- a/src/ILInspector.Decompiler.Tests/PassBugDiagnosticTests.cs
+++ b/src/ILInspector.Decompiler.Tests/PassBugDiagnosticTests.cs
@@ -1,0 +1,75 @@
+using ILInspector.Decompiler.Pipeline;
+using ILInspector.DecompilerHarness;
+
+namespace ILInspector.Decompiler.Tests;
+
+public class PassBugDiagnosticTests
+{
+    [Fact]
+    public void DirectPassBugEmitters_UseSharedFormatter()
+    {
+        var directory = new DirectoryInfo(AppContext.BaseDirectory);
+        while (directory is not null
+            && !File.Exists(Path.Combine(directory.FullName, "dotnet-inspect.slnx")))
+        {
+            directory = directory.Parent;
+        }
+        Assert.NotNull(directory);
+
+        string harnessDirectory = Path.Combine(directory.FullName, "tools", "DecompilerHarness");
+        string[] offenders = Directory.EnumerateFiles(harnessDirectory, "*.cs")
+            .Where(path => Path.GetFileName(path) != "PassBugDiagnostic.cs")
+            .Where(path => File.ReadAllText(path).Contains("PASS BUG", StringComparison.Ordinal))
+            .Select(path => Path.GetFileName(path)!)
+            .Order(StringComparer.Ordinal)
+            .ToArray();
+
+        Assert.Empty(offenders);
+    }
+
+    [Fact]
+    public void Format_IdentifiesAssemblyAndExactMethodBody()
+    {
+        var exception = new KeyNotFoundException("missing block 3");
+        var stringOverload = new MethodSignature(
+            TypeRef.CoreLib("System", "String"),
+            [new Parameter("value", TypeRef.CoreLib("System", "String"))],
+            HasThis: false,
+            GenericParameterCount: 0);
+        var integerOverload = stringOverload with
+        {
+            Parameters = [new Parameter("value", TypeRef.CoreLib("System", "Int32"))],
+        };
+
+        string first = PassBugDiagnostic.Format(
+            exception,
+            "/packages/example/1.0.0/lib/Example.dll",
+            "Example.Parser",
+            "Parse",
+            stringOverload);
+        string otherAssembly = PassBugDiagnostic.Format(
+            exception,
+            "/packages/example/2.0.0/lib/Example.dll",
+            "Example.Parser",
+            "Parse",
+            stringOverload);
+        string otherOverload = PassBugDiagnostic.Format(
+            exception,
+            "/packages/example/1.0.0/lib/Example.dll",
+            "Example.Parser",
+            "Parse",
+            integerOverload);
+
+        Assert.Equal(
+            "PASS BUG: KeyNotFoundException: missing block 3 " +
+            "(/packages/example/1.0.0/lib/Example.dll!Example.Parser::Parse" +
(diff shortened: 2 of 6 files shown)


richlander:

**Tier:** single review. This change is more than trivial (a new file, six
rewritten call sites, and a new gate test) but not high risk: it touches
harness diagnostic text only — no product path, no default CLI output, no
decompiler behavior. Reviewed by **MAI-Code** in an isolated review worktree.

### Finding 1 — gate could be bypassed by a nested harness file (MAI-Code)

**Accepted and fixed** in `07a5f8b6`.

`DirectPassBugEmitters_UseSharedFormatter` enumerated only the top level of
`tools/DecompilerHarness` (`Directory.EnumerateFiles(dir, "*.cs")`), so a new
emitter — or a stray `PASS BUG` literal — in a subdirectory was invisible to
*both* halves of the check. The reviewer demonstrated the bypass with a `/tmp/`
probe: a nested file containing the literal left `callSiteMatch=true` and
`offenderMatch=true`.

This mattered more than its size: the whole point of the gate is that the
"one owner" property cannot rot, and a gate with a blind spot is exactly the
"looks verified, isn't" failure `AGENTS.md` warns about.

Fix:

- Enumerate with `SearchOption.AllDirectories`.
- Exclude `bin`/`obj` build output via `IsBuildOutput`, matching whole path
  *segments* case-insensitively rather than substrings.
- Key both the call-site multiset and the offender list on the
  harness-relative path (separators normalized to `/`) instead of
  `Path.GetFileName`, so a nested `Program.cs` — or a nested
  `PassBugDiagnostic.cs`, which would otherwise be excluded by name — cannot
  masquerade as the top-level one.

Tamper-tested rather than assumed. Adding
`tools/DecompilerHarness/nested/Sneak.cs` containing a `PASS BUG` literal now
fails the gate:

```
ILInspector.Decompiler.Tests.PassBugDiagnosticTests.DirectPassBugEmitters_UseSharedFormatter [FAIL]
      Collection: ["nested/Sneak.cs"]
```

### Verified and dismissed

The reviewer independently confirmed at the fixed head, with real runs:

- **Fix closes the bypass**, and the inverse holds — a file under
  `tools/DecompilerHarness/objects/` is *not* excluded, proving segment
  equality rather than substring matching.
- **No fifth emitter.** Exactly six `PassBugDiagnostic.Format(...)` call sites
  (four `Program.cs`, one `SlotResidualCensus.cs`, one `SlotUnifierCensus.cs`)
  and no `PASS BUG` literal outside `PassBugDiagnostic.cs`.
- **Identity is genuinely discriminating**, and
  `Format_IdentifiesAssemblyAndExactMethodBody` is unaffected by the fix.
- **Gate verdict: genuine, not vacuous.**

### Evidence at the fixed head (`07a5f8b6`)

```console
$ dotnet build dotnet-inspect.slnx -c Release
0 Error(s)
$ dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -trait- "Speed=Slow"
Total: 3896, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0
$ dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build \
    -- -class ILInspector.Decompiler.Tests.PassBugDiagnosticTests
Total: 2, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0
```

Re-reviewed at the fixed head; MAI-Code reports no remaining blocking findings.

richlander:
Ready to merge

All merge-blocking validation, CI, and required review are complete at the reviewed head `07a5f8b6`:

- Adversarial review: single-review tier (low risk — harness diagnostics and its own test gate, no product path). Reviewed by MAI-Code, which raised one non-blocking finding (the gate enumerated only top-level `tools/DecompilerHarness/*.cs`, so a nested file bypassed both halves of the check). Fixed, tamper-tested, and re-reviewed clean at the fixed head — see the [reconciliation comment](https://github.com/richlander/dotnet-inspect/pull/3313#issuecomment-5093363459).
- CI green: `test (ubuntu-26.04, linux-x64)` success; path-gated jobs correctly skipped.
- Local: decompiler fast suite 3896 pass, 0 fail.

No follow-up work is outstanding for this PR.

PR #3314, opened 2026-07-27
Preserve .ctor as the RTS constructor metadata identity (#3251)

### LINKED ISSUE #3251: Two slow-gated ReturnToSender tests broken on main by #3129 (.ctor -> __ctor sanitizer)
## Summary

Two `ReturnToSenderPrototypeTests` cases fail on current `main` (verified at
`08ab86b1`). They were broken by #3129 / PR #3204 ("Route residual unspellable
generated names through the sanitizer", commit `e97cf17d`), whose sanitizer now
rewrites the unspellable constructor name `.ctor` to `__ctor` in the
ReturnToSender requirement/production model. CI did not catch this because
`ReturnToSenderPrototypeTests` is `[Trait("Speed", "Slow")]` and the slow gate is
dropped from the PR `test` job.

## Affected tests

- `ReturnToSenderPrototypeTests.CompileBackTargets_FullReconstructsPlainEventAccessorTarget`
  asserts a full-body production with `body.Member == "Holder..ctor"`; the model
  now emits `Holder.__ctor`, so `Assert.Contains` fails.
- `ReturnToSenderPrototypeTests.CompileBackFirstPropertyGetter_EmitsClosureConstructorRequirement`
  asserts `member.Name == ".ctor"` and a `typed-closure-constructor` fact with
  `Detail == ".ctor"`; both are now `__ctor`.

## Root cause

#3129's residual-name sanitizer treats `.ctor` (which contains a `.`) as an
unspellable generated name and rewrites it to `__ctor`. This changes the
constructor labels surfaced in the ReturnToSender requirement model
(`CompileBackMemberRequirement.Name`, `FullBodyProduction.Member`, and the
`typed-closure-constructor` fact `Detail`).

## Open question (owner decision)

Is the `.ctor` → `__ctor` rewrite intended for these RTS constructor labels?

- If it is a benign label change, the two test assertions should be updated to
  the sanitized spelling.
- If `.ctor` should be preserved as the constructor's model identity (and the
  sanitizer should not apply to this well-known metadata name in this context),
  this is a product regression in the sanitizer scope.

## Reproduction

```
git checkout 08ab86b1   # current main HEAD
dotnet build src/ILInspector.Decompiler.Tests/ILInspector.Decompiler.Tests.csproj -c Release
  -method "*CompileBackFirstPropertyGetter_EmitsClosureConstructorRequirement" \
  -method "*FullReconstructsPlainEventAccessorTarget"
# => Total: 2, Failed: 2
```

## Notes

Discovered while merging `origin/main` into the #3112 branch (PR #3222). Not
caused by and out of scope for #3112; filing separately as agreed. Consider
whether the slow RTS gate should run in a periodic/nightly job so
sanitizer/printer changes cannot silently break it again.

richlander:
## What

The ReturnToSender requirement model keeps `.ctor` as a constructor's metadata
identity instead of routing it through the unspellable-name sanitizer, which
had been rewriting it to `__ctor`.

Fixes #3251.

## Answer to the issue's open question

#3251 asks the owner to decide whether the `.ctor` → `__ctor` rewrite is
intended for RTS constructor labels, or a product regression in the
sanitizer's scope. **It is a regression.** The sanitizer exists to turn an
identifier that cannot be *spelled in C#* into one that can. `.ctor` is not a
C# identifier the model ever emits: constructors are emitted under the
declaring type's name, and `CompileBackMemberKind.Constructor` already carries
that fact. What `.ctor` is, in this model, is the *metadata* name — the key the
requirement model matches members on. Sanitizing it corrupts an identity that
is never printed, which is exactly the "do not infer identity from display
text" boundary in AGENTS.md.

So the fix restores the metadata spelling rather than updating the two
assertions to `__ctor`.

## What changed

Two call sites in `CompileBackSourceComposer` (`ReturnToSenderTypePlanner.cs`)
now skip the sanitizer when the member is a constructor. Both sites already
computed `isConstructor` immediately adjacent, so the guard reuses the existing
discriminator:

```csharp
string identifierName = isConstructor ? name : CSharpNaming.SourceMethodName(name);
```

Non-constructor names keep going through `CSharpNaming.SourceMethodName`
unchanged, so #3129's residual-name handling is untouched for the case it was
written for.

## Evidence

The two tests named in the issue are the gate. They fail on current `main` and
pass on this branch:

```console
# current main, df67a285
$ dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
    -method "*CompileBackFirstPropertyGetter_EmitsClosureConstructorRequirement" \
    -method "*FullReconstructsPlainEventAccessorTarget"
Total: 2, Errors: 0, Failed: 2

# this branch
$ dotnet run --project src/ILInspector.Decompiler.Tests -c Release \
    -- -class ILInspector.Decompiler.Tests.ReturnToSenderPrototypeTests
Total: 186, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0
```

Wider `Area=RoundTrip`, which is where these slow RTS gates live:

| Head | `Area=RoundTrip` result |
| --- | --- |
| `origin/main` (df67a285) | Total 533, **Failed 5** |
| this branch | Total 533, **Failed 3** |

The two removed failures are exactly the two from #3251. The remaining three
(`GeneratedFixtureCatalogTests`, all reporting
`minimal.parameter-attributes … compile-back=RecompileFail`) reproduce
identically on `origin/main` and are unrelated to this change; they are not
introduced or masked here.

## Boundary

Constructor identity in the RTS requirement/production model only. No printer
change, no change to how constructors are *emitted* in compile-back source
(they were already emitted under the type name), and no change to the
sanitizer itself or to any non-constructor name.

## Follow-up (non-blocking)

#3251 also notes that CI missed this because `ReturnToSenderPrototypeTests` is
`[Trait("Speed", "Slow")]` and the PR `test` job drops the slow gate. Whether
to run the slow RTS gate periodically is a separate scheduling decision and is
deliberately not bundled here.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/MemberIdentifierNameTests.cs
+++ b/src/ILInspector.Decompiler.Tests/MemberIdentifierNameTests.cs
@@ -1,0 +1,52 @@
+using ILInspector.Decompiler.Pipeline;
+using ILInspector.DecompilerHarness;
+
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// Fast gate for the ReturnToSender constructor identity restored by #3251.
+///
+/// The end-to-end coverage in <c>ReturnToSenderPrototypeTests</c> is
+/// <c>Speed=Slow</c>, and the PR test job drops that trait — which is exactly how
+/// the #3129 sanitizer silently rewrote <c>.ctor</c> to <c>__ctor</c> on
+/// <c>main</c>. These cases carry no trait, so they run in the PR job.
+/// </summary>
+[Trait("Area", "RoundTrip")]
+public class MemberIdentifierNameTests
+{
+    [Fact]
+    public void Constructor_KeepsMetadataName()
+    {
+        Assert.Equal(
+            ".ctor",
+            CompileBackSourceComposer.MemberIdentifierName(".ctor", isConstructor: true));
+    }
+
+    /// <summary>
+    /// Non-vacuity: the sanitizer really does corrupt <c>.ctor</c>, so
+    /// <see cref="Constructor_KeepsMetadataName"/> cannot pass by accident if the
+    /// guard is deleted.
+    /// </summary>
+    [Fact]
+    public void Sanitizer_WouldCorruptConstructorName_SoTheGuardIsLoadBearing()
+    {
+        string sanitized = CSharpNaming.SourceMethodName(".ctor");
+
+        Assert.NotEqual(".ctor", sanitized);
+        Assert.Equal(
+            sanitized,
+            CompileBackSourceComposer.MemberIdentifierName(".ctor", isConstructor: false));
+    }
+
+    [Theory]
+    [InlineData("Parse")]
+    [InlineData("<Run>g__Local|0_0")]
+    [InlineData("<>c__DisplayClass0_0")]
+    [InlineData("op_Addition")]
+    public void NonConstructor_StillRoutesThroughTheSanitizer(string metadataName)
+    {
+        Assert.Equal(
+            CSharpNaming.SourceMethodName(metadataName),
+            CompileBackSourceComposer.MemberIdentifierName(metadataName, isConstructor: false));
+    }
+}
--- a/tools/DecompilerHarness/ReturnToSenderTypePlanner.cs
+++ b/tools/DecompilerHarness/ReturnToSenderTypePlanner.cs
@@ -4136,6 +4136,23 @@ public static class CompileBackSourceComposer
         return true;
     }
 
+    /// <summary>
+    /// The name a member carries as its <see cref="CompileBackMethodIdentity"/>
+    /// key. This is model identity, not C# spelling: it is matched against other
+    /// requirement/production entries, never emitted as an identifier.
+    /// Constructors therefore keep their metadata name <c>.ctor</c> — the
+    /// declaring type name is what the emitter spells, selected by
+    /// <see cref="CompileBackMemberKind.Constructor"/>. Every other name is a
+    /// method name that can reach emitted source, so it goes through the #3129
+    /// residual-name sanitizer.
+    /// Gate: <c>MemberIdentifierNameTests</c>. That gate is fast on purpose — the
+    /// end-to-end <c>ReturnToSenderPrototypeTests</c> coverage is
+    /// <c>Speed=Slow</c> and so does not run in the PR test job, which is how
+    /// #3251 reached <c>main</c> unnoticed.
+    /// </summary>
+    internal static string MemberIdentifierName(string metadataName, bool isConstructor)
+        => isConstructor ? metadataName : CSharpNaming.SourceMethodName(metadataName);
+
     sealed class TypeProducer
     {
         public static CompileBackMemberRequirement? TryCreateClosureMemberRequirement(
@@ -4539,7 +4556,7 @@ public static class CompileBackSourceComposer
                 return null;
             }
 
-            string identifierName = isConstructor ? name : CSharpNaming.SourceMethodName(name);
+            string identifierName = MemberIdentifierName(name, isConstructor);
             return new CompileBackMemberRequirement(
                 new CompileBackMethodIdentity(typeIdentity.FullName, identifierName, DeclaringOverloadIndex(reader, typeDef, methodHandle, name), MethodSignatureText(identifierName, signature)),
                 isConstructor ? CompileBackMemberKind.Constructor : CompileBackMemberKind.Method,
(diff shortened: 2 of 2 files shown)


richlander:

Two-model tier (subtle correctness risk in the compile-back harness's identity
handling). Reviewed by **Gemini Pro** and **GPT**, each in an isolated read-only
worktree pinned at the exact head `1c6d7a05`, each given the same self-contained
prompt with base/head SHAs, design intent, concrete attack points, and a
requirement for real-run evidence.

### Gemini Pro — clean, no findings

Validated the owner decision directly: that `.ctor` is the constructor's
metadata identity and preserving it verbatim is correct, rather than routing it
through `CSharpNaming.SourceMethodName` which rewrites it into a C#-spellable
form. Confirmed no blocking or non-blocking issues.

### GPT — no blocking findings; one non-blocking, **actioned**

> The fix's only regression gate is `ReturnToSenderPrototypeTests`, which carries
> `[Trait("Speed", "Slow")]`. The PR `test` job runs `-trait- "Speed=Slow"`, so
> the gate does not execute in PR CI.

This is correct and it is the sharpest observation in either review, because it
identifies *why this regression shipped in the first place*: the behavior was
only ever protected by a suite that PR CI does not run. Verified independently
against `.github/workflows/ci.yml:301`. Fixing the bug without fixing that would
have left the next regression equally free to land.

**Resolution — `321911ff`:**

- Extracted the naming decision into a named helper
  `CompileBackSourceComposer.MemberIdentifierName(string metadataName, bool isConstructor)`
  in `tools/DecompilerHarness/ReturnToSenderTypePlanner.cs`. It sits at composer
  level rather than inside `TypeProducer`, which is a private nested class and
  therefore unreachable from the linked test assembly. Both former call sites now
  delegate to it, so the gate and the product path cannot drift apart.
- Added `src/ILInspector.Decompiler.Tests/MemberIdentifierNameTests.cs`, traited
  `Area=RoundTrip` with deliberately **no** `Speed` trait, so it runs in PR CI.
  Six cases, including a non-vacuity assertion that
  `CSharpNaming.SourceMethodName(".ctor") != ".ctor"` — without it the test would
  still pass if the transform it guards against ever became an identity function.
- Tamper-tested: replacing the helper body with the unguarded
  `CSharpNaming.SourceMethodName(metadataName)` fails `Constructor_KeepsMetadataName`.

### Validation at the fixed head

```console
$ dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -trait "Area=RoundTrip"
Total: 539, Failed: 3
```

The three failures are `GeneratedFixtureCatalogTests` and
`minimal.parameter-attributes`, pre-existing and reproducible on `origin/main`
in a separate baseline worktree — unrelated to this change. Before/after on the
same baseline: `main` fails 5, this branch fails 3, and the two it fixes are the
constructor-identity cases.

Both reviewers are being re-run against the fixed head `321911ff`; readiness is
not claimed until those come back clean.

richlander:
Ready to merge

All merge-blocking validation, CI, and required review are complete at the reviewed head `321911ff`.

**Adversarial review — two-model tier** (subtle correctness risk in the harness's identity handling). Both reviewers re-ran against the fixed head in fresh isolated read-only worktrees and both came back clean.

**GPT** — first pass: no blocking findings, one non-blocking (the fix's only gate was `Speed=Slow` and therefore excluded from PR CI, the very reason the regression shipped). Re-review at `321911ff`: *"Blocking findings: None. Non-blocking findings: None. The fixed head is clean."* Independently verified the extraction is behavior-preserving, that no third call site was missed, that the non-vacuity assertion is not a tautology, and that `Area=RoundTrip` with no `Speed` trait is not excluded by any filter in the PR `test` job. Also confirmed via `CSharpDeclarationWriter` that `.ctor` never reaches emitted C# — constructors are emitted through `FormatConstructorTypeName(type.Name)`, so preserving the metadata name cannot produce uncompilable generated code.

**Gemini Pro** — first pass clean; re-review at `321911ff`: *"No blocking or non-blocking issues found. The new HEAD is clean and the extraction is safe."* Independently confirmed `.cctor` is filtered before the predicate at both sites, that explicit interface implementations cannot be misrouted, and that the gate is load-bearing rather than testing itself because both call sites strictly delegate.

**CI:** `test (ubuntu-26.04, linux-x64)` success at `321911ff`; path-gated jobs correctly skipped.

**Local:** `Area=RoundTrip` 539 total, 3 failed — `GeneratedFixtureCatalogTests` / `minimal.parameter-attributes`, pre-existing and reproduced on `origin/main` in a separate baseline worktree. Same baseline before/after: `main` fails 5, this branch fails 3, and the two it fixes are the constructor-identity cases. Both reviewers independently reproduced the 539/3 result.

No follow-up work is outstanding for this PR.

PR #3315, opened 2026-07-27
Fold line terminators in IL display operands so they cannot escape their line (#3257)

### LINKED ISSUE #3257: Unescaped IL operand text can break out of the Annotated Source comment and inject C#
## Summary

`Annotated Source` renders raw IL beneath each C# statement as `//` comments. The IL operand text is not sanitized, so a metadata name containing a C# line terminator ends the comment and leaves the remainder of the name as **active C# code** in output a reader may paste or compile.

This is **pre-existing on `origin/main`** (reproduced at `08ab86b1`), independent of any config or flag.

## Reproduction

Emit an assembly with a field whose name contains a newline (`PersistedAssemblyBuilder` will accept it; the CLR does not require metadata names to be valid C# identifiers):

```csharp
var name = "field\n    public int Injected() => 42; //";
```

Then:

```console
$ dotnet-inspect member hostile.dll Hostile.Target.GetCount -S "Annotated Source"
```

```csharp
public int GetCount()
{
    return field_____public_int_Injected______42____;
        // IL_0000: ldarg.0                          // arg: this; stack: [Target]
        // IL_0001: ldfld        int Target::field
        public int Injected() => 42; //  // stack: [int]
        // IL_0006: ret                              // stack: []
}
```

The `public int Injected() => 42;` line is outside any comment.

## Blast radius

| Section | Affected | Notes |
| --- | --- | --- |
| `Annotated Source` | **Yes** | Code injection — the escaped text lands in a C# code block. |
| `IL` | Yes, lower severity | Listing is corrupted across lines, but it is not a C# block, so nothing becomes executable. |
| `Decompiled Source` | No | The C# printer already sanitizes identifiers (`field_____public_int_Injected______42____`). |
| `Facts` | No | |

Note that `Decompiled Source` is already correct, which shows the sanitizing convention exists — it simply is not applied to the IL text producer.

## Suggested fix

Fold C# line terminators in the IL instruction/operand text at its producer. `string.ReplaceLineEndings(" ")` is the convention already used elsewhere in the CLI and covers every terminator C# recognizes (CR, LF, CRLF, FF, NEL U+0085, LS U+2028, PS U+2029). The fix is byte-invariant for all well-formed assemblies, since benign operands contain no terminators.

## Context

Found while reviewing #3235, which added a taste annotation to the same view. That annotation had the same class of bug via `DecompilerDecision.Subject`; it is fixed in that PR. This issue tracks the separate, pre-existing IL-text vector, which lives in a different layer and affects default output, so it is deliberately not fixed there.

See `docs/design/untrusted-data-threat-model.md`.

richlander:
## What

IL display text in `Annotated Source` can no longer break out of the line that
frames it. Metadata **names** and **fact details** are folded at the points
where their text is composed, and user **strings** are escaped losslessly, so
none of them can end its `//` comment and leave the remainder as active C#.

Fixes #3257.

## Why

`Annotated Source` renders IL beneath each C# statement as `//` comments. That
text was unsanitized in two distinct ways.

**Names.** A field named `"field\n    public int Injected() => 42; //"` ends the
comment and leaves the remainder as **active C# code** in a block a reader may
paste or compile. The CLR does not require metadata names to be valid C#
identifiers, and `PersistedAssemblyBuilder` will emit one.

**Facts.** `ResearchViews` splices fact text into an IL comment line *after*
the IL producer has folded it, and appends fact text to C# lines that the IL
producer never touches. `IAnnotation.Detail` carries untrusted metadata — a
callee name, an allocated type name — so a terminator arriving through a fact
escaped a comment the producer believed it had closed, on **both** surfaces.

**Strings.** `IlProjection` resolved `ldstr` from the `#US` heap raw, while the
`IL` section had always run the same content through
`ILStringEscaper.ForDisplay`. That asymmetry is the more serious half, because
it does not need a hostile input at all: an ordinary multi-line C# string
literal already breaks the comment on `main`. Verified against a plain library
containing a verbatim string — no attacker required.

Both are pre-existing on `main` and independent of any flag.

## What changed

Names fold at the **composition point** rather than per-channel, so a future
annotation kind inherits the guarantee instead of having to remember it:

- `IlProjection.FormatAnnotatedInstrPart` folds the composed instruction text
  and the composed annotation list. This covers the operand, the `arg:` and
  `local:` name channels, and the stack and fact channels at one place.
- `IlProjection.AnnotatedHeader` folds the `//   Parameters:` and
  `//   Locals:` lines.
- `InstructionProducer.FormatOperand` (metadata-side, feeds the `IL` section)
  folds for display syntaxes and **not** for `ILSyntax.Canonical`.
- `AnnotationText.Format` folds fact text. Its doc comment already called it
  "the one place that turns a hidden fact into display text", which makes it the
  one place that can close this: callers cannot fold on our behalf precisely
  because they append *after* their own fold. All four non-test call sites are
  line-oriented display, so the fold covers them and any future one.

The fact fold is deliberately **display-only**. Typed consumers — JSON and TSV
projections, CLI `Facts` rows, identity diffing — read `IAnnotation.Detail`
directly and still see exact metadata text.

`string.ReplaceLineEndings(" ")` covers every terminator C# recognizes — CR,
LF, CRLF, FF, NEL U+0085, LS U+2028, PS U+2029 — rather than an ad-hoc
`\r`/`\n` pair.

Strings are **escaped, not folded**, because escaping is lossless and matches
what the `IL` section already did:

- `IlProjection`'s `ldstr` arm now routes through `ILStringEscaper.ForDisplay`.
- `ForDisplay` is rewritten from a `Replace` chain to a char loop that also
  covers `\f` and emits `\uXXXX` for U+0085, U+2028 and U+2029 — terminators C#
  honours that its own escape syntax has no short form for. A two-line string
  now renders as `ldstr "first line\nsecond line"` on one line, fully
  recoverable.

The canonical carve-out's comment previously asserted an escaping property the
code does not have. It now states the real policy: `CanonicalIL.QuoteName`
escapes only `\` and `'`, the ilasm grammar permits a literal newline inside a
SQSTRING, so preserving the terminator is what keeps the metadata name intact
through reassembly — folding would silently rename the member. Canonical is
consumed by the IL round-trip scaffold and reaches no Markdown or C# surface.

## Evidence

Every gate below was verified non-vacuous by reverting the fix it covers.

| Gate | Claim | Failures without the fix |
| --- | --- | --- |
| `UntrustedIlPresentationTests.AnnotatedSource_IlOperandCannotEscapeItsComment` | operand channel, full terminator set | — |
| `UntrustedIlPresentationTests.AnnotatedSource_HostileParameterNameCannotEscapeItsIlComment` | parameter/local **name** channels | 3 |
| `UntrustedIlPresentationTests.AnnotatedSource_UserStringIsEscapedLosslessly` | strings escaped **and recoverable** | 5 |
| `UntrustedIlPresentationTests.AnnotatedSource_HostileFactDetailCannotEscapeItsComment` | fact details, end to end via a real `alloc.new` | 3 |
| `AnnotationTextTests` | fact text folded; `Detail` unmutated for typed consumers; ordinary detail byte-identical | 8 |
| `ILDisassemblerEmitTests.Emit_DisplayOperand_FoldsCSharpLineTerminators` | display folds | — |
| `ILDisassemblerEmitTests.Emit_CanonicalOperand_PreservesLineTerminators` | canonical deliberately does **not** | 3 |
| `CommandExecutionTests` (new cases) | end-to-end rendered output has no line outside a comment | — |

The canonical gate is new in response to review: the comment named
`ILDisassemblerEmitTests`, but every existing case there ran
`ILSyntax.Display`, so the carve-out was named-but-unenforced.

```console
$ dotnet build dotnet-inspect.slnx -c Release
0 Error(s)
$ dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -trait- "Speed=Slow"
Total: 3923, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0
$ dotnet run --project src/dotnet-inspect.Tests -c Release
Total: 2199, Errors: 0, Failed: 0, Skipped: 12, Not Run: 0
$ dotnet run --project tests/ILInspector.Metadata.Tests -c Release
Total: 915, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0
$ dotnet run --project src/ILInspector.Research.Tests -c Release
Total: 132, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0
```

(The 12 skips are the pre-existing `ilasm`/`ildasm`-dependent cases; those
tools are absent in this environment.)

## Boundary

**Output does change for benign input, and that is the point.** An earlier
revision of this description claimed byte-invariance for well-formed
assemblies; that was wrong. Any assembly containing a multi-line string literal
renders differently now — previously the literal spilled across lines and broke
the comment, now it renders escaped on one line. Names remain byte-invariant:
a name with no line terminator is untouched by the fold.

**Known remaining channel, deliberately not fixed here.** The C# printer spells
parameter, local, and property names with `CSharpNaming.EscapeIdentifier`,
which `@`-escapes reserved keywords but does **not** sanitize characters that
cannot appear in a C# identifier — unlike fields (`SafeIdentifier`) and methods
(`SourceMethodName`). A hostile parameter name therefore still escapes in
`Decompiled Source`, which issue #3257's blast-radius table recorded as
unaffected. Filed as **#3319**: it is a CSharp-layer defect with ~30 call sites
and warrants its own PR and gates rather than being widened into this one. This
PR closes the IL text layer; #3319 closes the C# spelling layer.

No change to `Facts` or any typed identity — this is display text only, at the
point of display.

## Review

Reviewed by Gemini Pro and GPT (two-model tier: security-relevant, default
output path) over **three rounds**. In each of the first two rounds both
reviewers independently found the same blocking gap:

1. The name channels survived an operand-only fold.
2. The fact channel survived the fold at what the second revision called "the
   composition point" — it was not the last one.

Both also challenged the canonical carve-out justification, and Gemini
ultimately verified the SQSTRING grammar claim against the vendored
`CIL.g4` and revised its position to agree. Round three is clean. See the
reconciliation comments for full attribution.


richlander:

Two-model tier (security-relevant, default output path). Reviewed by **Gemini
Pro** and **GPT** as security reviewers, each in an isolated read-only worktree
pinned at the exact head `29cf7f8f`, each given the same self-contained prompt
with base/head SHAs, design intent, concrete attack points, and a requirement
for real-run evidence.

Both reviewers found the same blocking gap independently, and both were right.
The first revision of this PR did not fix the vulnerability it claimed to fix.

### Finding 1 — injection survives via parameter and local names (HIGH, blocking). Confirmed, fixed.

Gemini Pro rated this HIGH; GPT found it independently. The first revision
folded `IlProjection.FormatOperand` only, which closed the operand channel while
leaving the *name* channels wide open: the `arg:` and `local:` annotations and
the `//   Parameters:` / `//   Locals:` header lines still emitted raw metadata
names straight into a `//` comment.

Reproduced before accepting: a purpose-built assembly with a newline in a
parameter name produced **three separate comment escapes in one method's
output** at the old head.

**Resolution.** Fold at the **composition point** rather than per-channel.
`FormatAnnotatedInstrPart` now folds the composed instruction text and the
composed annotation string; `AnnotatedHeader` folds the Parameters and Locals
lines. This is one place instead of five, so a future annotation kind inherits
the guarantee rather than having to remember it — which is precisely the failure
mode that produced this finding.

### Finding 2 — the byte-neutrality claim is false. Confirmed, retracted.

Both reviewers challenged the "byte-invariant for well-formed assemblies" claim
in the original description. Verified independently rather than taken on
assertion: compiled a plain library containing an ordinary multi-line verbatim
string literal and rendered it on `main`. The `ldstr` operand already spilled
across lines and broke the comment.

So this was **never a hostile-only defect** — benign, everyday C# triggers it.
The description has been rewritten to say output *does* change for benign input,
that byte-invariance holds only for names, and to lead with the fact that the
string case needs no attacker.

This also changed the fix. Strings are now **escaped, not folded**, via
`ILStringEscaper.ForDisplay` — which the `IL` section had always applied to the
same `#US` content, an asymmetry neither the issue nor the first revision
noticed. Escaping is lossless; folding is not. `ForDisplay` was rewritten from a
`Replace` chain to a char loop that also covers `\f` and emits `\uXXXX` for
U+0085, U+2028 and U+2029.

### Finding 3 — the canonical carve-out justification is false. Confirmed; GPT's reading adopted.

Both reviewers caught that the comment asserted an escaping property the code
does not have: `CanonicalIL.QuoteName` escapes only `\` and `'`.

The reviewers then diverged, and this is where having two models paid. Gemini
concluded canonical has *no policy at all*. GPT went to the vendored ilasm
grammar and found that SQSTRING **permits** a literal newline inside a quoted
identifier — so preserving the terminator is exactly what keeps the metadata
name recoverable through reassembly, and folding it would silently *rename the
member*. GPT's reading is the stronger one and is what the comment now states,
together with the exposure boundary: canonical is consumed by the IL round-trip
scaffold and reaches no Markdown or C# surface.

### Finding 4 — the carve-out was named-but-ungated (GPT, non-blocking). Fixed.

GPT noticed the comment named `ILDisassemblerEmitTests` as its gate while every
existing case there ran `ILSyntax.Display`, so nothing enforced the split. Added
`Emit_CanonicalOperand_PreservesLineTerminators`, which disassembles the same
hostile assembly with `ILSyntax.Canonical` and asserts the terminator **is**
preserved. Tamper-tested: 3 failures if canonical inherits the fold.

### Not fixed here, filed as #3319

Chasing the blocking finding surfaced a further channel the reviewers did not
raise: the C# printer spells parameter, local, and property names with
`CSharpNaming.EscapeIdentifier`, which `@`-escapes reserved keywords but does
**not** sanitize characters that cannot appear in a C# identifier — unlike
fields (`SafeIdentifier`) and methods (`SourceMethodName`). A hostile parameter
name therefore still escapes in **`Decompiled Source`**, which issue #3257's
blast-radius table recorded as unaffected.

Filed as #3319 with the reproduction rather than widened into this PR: it is a
CSharp-layer defect with roughly 30 call sites and deserves its own gates and
its own review. This PR closes the IL text layer; #3319 closes the C# spelling
layer. The description states this boundary explicitly rather than implying
`Annotated Source` is fully closed.

### Gate non-vacuity

Every gate was verified by reverting the fix it covers:

| Gate | Failures without the fix |
| --- | --- |
| `AnnotatedSource_HostileParameterNameCannotEscapeItsIlComment` | 3 |
| `AnnotatedSource_UserStringIsEscapedLosslessly` | 5 |
| `Emit_CanonicalOperand_PreservesLineTerminators` | 3 |

### Validation at the fixed head `a86fc29a`

```console
$ dotnet build dotnet-inspect.slnx -c Release
0 Error(s)
$ dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -trait- "Speed=Slow"
Total: 3909, Failed: 0
$ dotnet run --project src/dotnet-inspect.Tests -c Release
Total: 2199, Failed: 0, Skipped: 12
$ dotnet run --project tests/ILInspector.Metadata.Tests -c Release
Total: 915, Failed: 0
```

Both reviewers are being re-run against `a86fc29a`, with the grammar claim in
Finding 3 handed back for independent verification. Readiness is not claimed
until those come back clean.

richlander:
## Adversarial review reconciliation — rounds 2 and 3

### Round 2 (head `a86fc29a`) — both reviewers blocked, independently, on the same thing again

**Gemini Pro and GPT both found that the fix from round 1 was still incomplete.**

Round 1's fix folded at `IlProjection.FormatAnnotatedInstrPart` and the PR called that
"the composition point". It was not the last one. `ResearchViews.AddFactsToAnnotatedLine`
splices fact text into the IL comment line *after* `IlProjection` has folded it, and
`RenderMixedStream` appends fact text to C# lines that the IL producer never touches at
all. `IAnnotation.Detail` carries untrusted metadata — a callee name via `AnalysisFindings`,
an allocated type name via the allocation producer — so a terminator arriving through a
fact escaped a comment the producer believed it had closed, on **both** surfaces.

Gemini reproduced it end to end: a `newobj` on a hostile-named type, whose name becomes
`alloc.new`'s detail, landed the payload at column zero as active C#. Reproduced
independently here before accepting.

This is the second time in this PR that a claimed composition point turned out not to be
the last one. That is the finding behind the finding, and it is why the fix moved again.

**Resolution — `fdf19fdc`:** fold inside `AnnotationText.Format(IAnnotation)`. Its own doc
comment already called it *"the one place that turns a hidden fact into display text"*,
which is exactly what makes it the one place that can close this — callers cannot fold on
our behalf, precisely because they append *after* their own fold. All four non-test call
sites are line-oriented display, so the fold covers them and any future one.

The fold is deliberately **display-only**. Typed consumers read `IAnnotation.Detail`
directly and still see exact metadata text; `Format_DoesNotMutateTheUnderlyingDetail` pins
that boundary so a later "tidy-up" cannot quietly push the fold into the data.

### Round 2 — findings settled rather than actioned

- **SQSTRING grammar.** Gemini had argued in round 1 that the canonical carve-out had no
  policy at all. Handed back for independent verification, it located
  `vendor/ilassembler:ILAssembler/gen/CIL.g4`, read the rule
  `SQSTRING: '\'' (~('\'' | '\\') | '\\' .)* '\'';`, confirmed it permits line terminators,
  and **revised its own position** to agree with GPT. Both reviewers now agree the carve-out
  is correct and that folding there would silently rename the member. Closed.
- **`ILStringEscaper.ForDisplay`.** Both cleared it as complete and surrogate-safe. Gemini
  added the sharpest supporting argument: the escaped `\u2028` is inert because the C#
  compiler explicitly does not process Unicode escapes inside `//` comments (C# spec 2.4.1),
  so the escape cannot be re-interpreted into the very terminator it replaced.
- **Scope split.** Both endorsed deferring the C# printer's `EscapeIdentifier` defect to
  #3319 as a genuinely separate threat surface with its own renderer.

### Round 3 (head `fdf19fdc`) — both clean

**GPT:** *"Blocking findings: None. Non-blocking findings: None. Head `fdf19fdc` is clean."*
Traced all four production callers of `AnnotationText.Format` and confirmed nothing splices
fact text downstream — `ApiOutputFormatter` treats annotated source as opaque
`DecompilerResult.Output`, and the Markdown and printable-document layers wrap
`CodeSection.Content` without decoding escapes. Verified the typed boundary at
`AnnotationStructuredView` (JSON and TSV), `ResearchViews` `FactRow`s, and the CLI Fact rows.

**Gemini Pro:** *"The head (`fdf19fdc`) is fully clean and the vulnerability is closed."*
Built its own hostile assembly and confirmed the fold end to end, then proved the typed
boundary empirically by running `--jsonl` and showing the raw `\n` still present in the
`detail` field. Independently tampered `FoldLine` and confirmed the gates catch it. Also
checked `CostOverlayHeaderComments`, which does not fold — and cleared it, because its only
producer injects integers.

### Gate non-vacuity

Every gate verified by removing the fix it covers:

| Gate | Failures without the fix |
| --- | --- |
| `AnnotationTextTests` (fact text, list composition, typed boundary) | 8 |
| `AnnotatedSource_HostileFactDetailCannotEscapeItsComment` | 3 |
| `AnnotatedSource_HostileParameterNameCannotEscapeItsIlComment` | 3 |
| `AnnotatedSource_UserStringIsEscapedLosslessly` | 5 |
| `Emit_CanonicalOperand_PreservesLineTerminators` | 3 |

The end-to-end fact gate asserts `alloc.new(` is actually rendered before asserting
containment, so it cannot pass by producing no fact at all.

### Validation at `fdf19fdc`

```console
$ dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -trait- "Speed=Slow"
Total: 3923, Failed: 0
$ dotnet run --project src/dotnet-inspect.Tests -c Release
Total: 2199, Failed: 0, Skipped: 12
$ dotnet run --project tests/ILInspector.Metadata.Tests -c Release
Total: 915, Failed: 0
$ dotnet run --project src/ILInspector.Research.Tests -c Release
Total: 132, Failed: 0
```

Both reviewers independently reproduced these totals in their own worktrees.

richlander:
Ready to merge

All merge-blocking validation, CI, and required review are complete at the reviewed head `fdf19fdc`.

**Adversarial review — two-model tier** (security-relevant, default output path), Gemini Pro and GPT, three rounds. Each round used fresh isolated read-only worktrees pinned at the exact head, with the same self-contained prompt. In each of the first two rounds both reviewers independently found the same blocking gap:

1. **Round 1** — the operand-only fold left the parameter and local *name* channels open.
2. **Round 2** — the fix at "the composition point" was not the last composition point; fact text is spliced onto IL lines after that fold and onto C# lines that never had one.

**Round 3 is clean from both.** GPT: *"Blocking findings: None. Non-blocking findings: None. Head `fdf19fdc` is clean."* Gemini Pro: *"The head (`fdf19fdc`) is fully clean and the vulnerability is closed."* Both built their own hostile assemblies rather than accepting the claim, both traced every path from a fact to rendered output, and both verified the typed JSON/TSV boundary is unaffected — Gemini empirically, by showing the raw `\n` still present in `--jsonl` output.

Full attribution in the [round 1 reconciliation](https://github.com/richlander/dotnet-inspect/pull/3315#issuecomment-5093817686) and the [rounds 2–3 reconciliation](https://github.com/richlander/dotnet-inspect/pull/3315#issuecomment-5094152957).

**CI:** all jobs green at `fdf19fdc` — `test`, `build-net10`, `csharp-diff-smoke`, `il-diff-smoke`.

**Local:** decompiler fast 3923 pass, CLI 2199 pass / 12 skipped (pre-existing ilasm/ildasm absence), metadata 915 pass, research 132 pass. Every gate tamper-tested for non-vacuity.

**Follow-up, explicitly non-blocking:** #3319 tracks the C# printer's `EscapeIdentifier` vs `SafeIdentifier` defect, which leaves `Decompiled Source` injectable through parameter, local, and property names — and which contradicts issue #3257's blast-radius table. Both reviewers were asked whether that scope split was wrong and both endorsed it as a genuinely separate threat surface with its own renderer. This PR closes the IL text layer; #3319 closes the C# spelling layer.

PR #3316, opened 2026-07-27
Arm semantic IR invariants by default

### LINKED ISSUE #3302: Semantic IR invariants stay opt-in on a stale rationale: the blocker is 5 fixtures, not ~120

richlander:
Closes #3302.

## What changed

The semantic IR invariant level is now armed by default, so every validating host checks semantic invariants after every pass instead of only when an operator remembers `DOTNET_INSPECT_IR_INVARIANTS=full`.

The level stayed opt-in on a rationale that was stale by 24x: `IrInvariants.cs` justified it because arming it suite-wide "would false-positive on ~120 minimal-fixture tests". Measured, the real number is **five**. At ~120 the opt-in is obviously right; at five it is a bounded fixture cleanup, not a standing reason to ship less validation. (#3303 already corrected the number in the doc and filed this as the follow-up; this PR does the work.)

This matters because semantic invariants are the level that catches a pass leaving a slot reference that no longer resolves — exactly the defect class the local-elimination bookkeeping can get wrong. Leaving it off by default meant the checks most likely to catch a real regression were the ones no host ran.

## How

- **The five fixtures now declare the locals their bodies already reference.** Deliberately *not* by deriving the local table from the body: that would make every fixture pass by construction and retire the invariant while appearing to keep it. A fixture declaring two locals that stores to slot five still fails, as it should.
- **`CheckSemantics` became a computed projection of `Enabled`** (`=> Enabled`). This builds on #3303, which had just made the property setter-less. Rather than reintroduce a private setter so the shipped tool could lower both levels, the projection makes the two levels impossible to drift apart *by construction* — there is no backing field to move, #3303's "cannot be moved in either direction in-process" guarantee is preserved, and `DisableForShippedTool()` still lowers both with one assignment.
- **`RequestsSemantics` is gone.** `full` keeps working and keeps meaning both levels; `1`/`true` now arm both, a documented strengthening.

Consequence, called out explicitly: there is no longer a spelling for structural-only. That is intentional. A `DOTNET_INSPECT_IR_INVARIANTS=1` that quietly bought *less* validation than the default would be the same silent-downgrade trap #3289 removed for the off case.

## Teeth

- `PipelineRunner_ThrowsWhenAPassLeavesAnUnresolvableSlot` fails if the semantic default is ever flipped back off.
- `OutOfRangeSlot_IsInvisibleToTheStructuralLevel` proves the structural level genuinely cannot see the defect, so the first test is testing something real.
- The public-surface census from #3303 already pins the mutable surface; I dropped my weaker duplicate in favour of it.

## Evidence

- `DOTNET_INSPECT_IR_INVARIANTS=full --gate fast` failed **exactly the five tests named in the issue** before the fixture fix, and zero after.
- **Negative control:** patching the product default to `CheckSemantics => false` fails **exactly the two intended tests** and nothing else.
- Only five fixtures were affected because per-pass validation fires inside `IrPasses.Run`/`PipelineRunner`; fixtures that call `pass.Run(...)` directly never reach the hook. That is what bounds the blast radius.
- Suites: `--gate fast` 3896, Corpus 80, CLI 2188, Analysis 539, Services 319, Metadata 915 — all green.

**Cost:** paired A/B of the `Area=Corpus` gate against the parent commit, alternating runs on an idle machine — base 18.000/17.757/22.758s, armed 18.626/18.655/18.082s. Below the noise floor; run-to-run spread on one build is wider than the difference between builds. An earlier 11% figure was measured under contention from another session and is corrected in the second commit rather than quietly dropped.

**Pre-existing failures:** nine slow-gate tests fail on this machine. All nine were confirmed to fail identically on a clean `origin/main` worktree and are unrelated to this change.


richlander:

Two models reviewed at fixed heads in isolated read-only worktrees, per `AGENTS.md` (I authored this as Claude, so the roster used the other two families): **GPT-5.6** and **Gemini 3.1 Pro**. Both re-reviewed after each round of fixes. **Both are clean at the final head `312468f7`.**

### Findings and disposition

**1. "Measured, the real number was five" is misleading — ACCEPTED** *(found independently by both reviewers, which is what raised its weight)*

Both correctly identified that five counts fixtures that *reach the per-pass hook* and fail, not fixtures that would satisfy the level. Per-pass validation fires inside `IrPasses.Run`/`PipelineRunner`, so tests calling `pass.Run(...)` directly never reach it. I reproduced this: 17 test files build an `IrFunction` with an empty local table and reference slots in it.

Fixed in `95e448e8` — `IrInvariants.cs` and `AGENTS.md` now state precisely what the five counts and warn that converting one of the others to `IrPasses.Run` will fail it (the correct signal — the fixture is genuinely malformed — not a regression). The cleanup is tracked as **#3317** with the concrete file list and the "declare the locals, do not derive them" constraint.

Correcting the wording mattered more than usual here: this PR exists because a number in a doc comment outlived its measurement and then justified a decision on its own authority. Shipping a second number that invites the same misreading would have repeated the defect while claiming to fix it. Both reviewers confirmed the deferral to #3317 is a defensible boundary.

**2. `CheckInvariant()` is still a structural-only spelling (Gemini, "blocking") — PARTIALLY ACCEPTED, code fix DECLINED**

The criticism of the *wording* was right and is fixed: the "no spelling for structural-only" claim is now scoped to the **environment variable**, which is all it was ever about.

The suggested *code* fix — having the parameterless overload consult `IrInvariants.CheckSemantics` — was declined. Verified reasons: `StructuralCheck_IgnoresLocalSlotRange_ButSemanticCheckCatchesIt` and `OutOfRangeSlot_IsInvisibleToTheStructuralLevel` both depend on the overload being structural-only, and the per-call parameter is the hermetic mechanism #3303 deliberately kept so a test's coverage does not depend on what a host or operator configured — which matters under xUnit's parallel collections. A visible argument at a call site is categorically different from a process-wide knob that silently lowers what another host asked for.

**Gemini withdrew this finding on re-review** and agreed the declination is correct. GPT independently judged the declination correct. Also fixed the stale "the suite-wide default" comment this reading exposed.

**3. Three surviving comments contradicted the corrected description (GPT, non-blocking) — ACCEPTED**

`IrNode.cs` still said semantics were "off by default"; `CorpusSweepGateTests.cs` twice claimed the corpus gate is "the only place semantic invariants execute". True before this PR, false after. Fixed in `312468f7`. The corpus gate keeps a real, distinct role, so it was restated rather than deleted: whole-corpus coverage of final trees, and the only site that reports a count instead of throwing. Gemini's final pass grepped repo-wide and confirmed zero remaining stale claims.

**4. Verified non-issue.** GPT observed the semantic teeth test fails under `DOTNET_INSPECT_IR_INVARIANTS=0`. Confirmed the pre-existing structural teeth test from #3289 behaves identically — both fail together under `=0`. That is the established pattern for teeth tests, not new behavior.

### Independently verified by reviewers

- `312468f7` and `95e448e8` are comments/docs only; no product or test logic changed (both verified with `git show --stat` plus inspection, not on my word).
- The corpus gate genuinely counts rather than throws; the per-pass hooks genuinely thread the global level (Gemini read the code, not the prose).
- Build clean, `--gate fast` 3896 pass / 0 fail, in both reviewers' own worktrees.

### Non-blocking follow-up

- **#3317** — declare locals in the remaining direct-pass fixtures.

richlander:
Ready to merge

CI green at `312468f7` (test, build-net10, markdownlint, csharp-diff-smoke, changes; il-diff-smoke and pack path-gated out). Both adversarial reviews clean at that exact head, reconciled above.

Non-blocking follow-up only: #3317 (declare locals in the remaining direct-pass fixtures).

richlander:
Reviewed at head `312468f7` (base `origin/main`) in an isolated worktree. **This is the right change and the evidence holds up.** Every claim I checked reproduced. Two minor items below, neither blocking.

Closes my #3302, and the reasoning that gets there is the part worth calling out: the fix for a stale rationale was to *re-measure it*, and then to act on the real number rather than inherit the old conclusion. The doc even records the correction provenance so the next reader re-measures instead of trusting the figure again.

## Verified independently

**Arming is real, not nominal.** The claim I most wanted to falsify is that the fixtures were "fixed" in a way that makes the level vacuous. They were not. Reverting one fixture's declared locals back to `[]`:

```
IteratorReconstructionPassTests.YieldBreakLoopIterator_WithDispatchSideEffect_DeclinesHonestly [FAIL]
  System.InvalidOperationException : Invariant violated: 'StoreLocal 0 (int)'
  references local slot 0, out of range for the enclosing scope's 0 local slot(s).
```

The declarations are load-bearing and the semantic level genuinely fires through `IrPasses.Run`. Declaring the specific typed slots with a comment naming what each one *is* — rather than deriving the table from the body — is the distinction that keeps the invariant alive, and the doc is explicit about why. That was the trap here and it was avoided deliberately.

**Negative control reproduces exactly.** Patching `CheckSemantics => false`:

```
--gate fast  →  Total: 3896, Failed: 2
  IrInvariantCheckTests.PipelineRunner_ThrowsWhenAPassLeavesAnUnresolvableSlot
  IrInvariantsHostContractTests.SemanticLevelIsOnByDefault_AndTracksTheStructuralLevel
```

Two tests, the two intended ones, nothing else. `OutOfRangeSlot_IsInvisibleToTheStructuralLevel` is a proper paired control — the same corrupted tree passes `CheckInvariant()` and throws under `includeSemantics: true`, so the teeth test is provably testing the semantic level and not catching the defect incidentally.

**Coverage is not silently reduced.** `IrInvariantsHostContractTests` carries 9 test attributes on both `origin/main` and this head; the only change is the rename `SemanticLevelStaysOptIn` → `SemanticLevelIsOnByDefault_AndTracksTheStructuralLevel`. I checked because "dropped my weaker duplicate in favour of the census" is the kind of sentence worth confirming.

**The "roughly a dozen" residual has a real referent.** #3317 exists and is cited from `IrNode.cs`, so the unaffected-today fixtures are tracked rather than named-and-forgotten.

**`full` still parses.** `ParseRequest` maps `1`/`true`/`full`/`on`/`yes` → true, so the documented compatibility claim holds.

### Evidence table

| Check | Result |
| --- | --- |
| Build (`Release`, whole solution) | clean |
| `--gate fast` | **3896 / 0** |
| `Area=Corpus` | **80 / 0** (~19.5s) |
| `src/dotnet-inspect.Tests` | **2188 / 0** (12 skipped) |
| env `0` / `off` | 1 failed — `AnEnvironmentOffRequestDoesNotSilentlyDisarmTheSuite`, message names the env cause correctly |
| env `1` / `true` / `full` | 0 failed |

I ran the CLI suite specifically because it drives real IL through the pipeline in hosts that never call `DisableForShippedTool()` — that is where newly-armed semantic checks over real importer output would surface. Clean.

On the cost claim: I am not going to treat any of these timings as a measurement, and neither should anyone else — this machine is shared. What I will say is that correcting the earlier 11% figure *in a commit*, rather than quietly dropping it once it turned out to be contention, is the right handling. I made exactly that mistake on #3289.

## Minor — the immutability claim now contradicts itself

`IrInvariants.cs:71-74` says #3303's guarantee is preserved:

> the level can be neither raised nor lowered in-process, and no test can race the collections xUnit runs in parallel by toggling it — is preserved, and the shipped tool's opt-out lowers both levels with one assignment.

The first clause and the last clause disagree. `CheckSemantics => Enabled`, `Enabled` has a `private set`, and `DisableForShippedTool()` assigns it — so `CheckSemantics` *can* be lowered in-process, and the same sentence says so.

The **substantive** guarantee is intact, and that is what matters: the setter is private, so only `IrInvariants` can move it; the one public mover is `DisableForShippedTool()`; and the #3303 census pins that to `src/dotnet-inspect/Program.cs`. No test can race it. But the property is no longer immutable the way it was when it was `{ get; } = RequestsSemantics(...)`, and this file's whole purpose is to be precise about exactly this.

Suggested wording: the level cannot be moved *independently of* `Enabled`, and only the single sanctioned opt-out moves either. That states the real, stronger-in-practice property without claiming an immutability the code no longer has.

## Minor — indentation damage in the doc block

`IrInvariants.cs:70` lost its indentation:

```
    /// invariants that only hold for a fully-formed function tree — currently
/// local-slot range (see <see cref="IrNode.CheckInvariant(bool)"/>). Armed
    /// with <see cref="Enabled"/> since #3302, and not merely alongside it: this
```

One line at column 0 inside an otherwise 4-space-indented block. Cosmetic, one-character fix.

Neither item blocks. Happy to see this land.

richlander:
Both items addressed in `54b24bcc` (which also merges current `origin/main`). Thanks — the first one is a fair hit and worth more than "minor".

**1. The immutability claim contradicting itself.** You're right, and the two clauses can't both be true. When `CheckSemantics` was `{ get; } = RequestsSemantics(...)` it genuinely could not move; as a projection of `Enabled` it moves whenever `Enabled` does, and `DisableForShippedTool()` assigns `Enabled`. I inherited #3303's sentence and kept asserting its guarantee after changing the thing that made it true.

Rewritten to your suggested framing — the level cannot be moved *independently of* `Enabled`, the only setter involved is `Enabled`'s and it is private, and the single public mover is `DisableForShippedTool()`, pinned by #3303's census to one call site in the shipped CLI. It also now says explicitly that the property is *not* immutable any more, so the next reader isn't left reconciling it themselves.

A file whose whole job is precision about which guarantees hold shouldn't ship a self-contradicting one, least of all in a PR about a doc comment that outlived its subject. Same defect class, one level up.

While rewriting I caught a second-order version of it in my own fix: I'd written "both setters are private", but `CheckSemantics` has no setter at all. Corrected to name only `Enabled`'s.

**2. Indentation damage.** Fixed. Introduced by the scripted conflict resolution I used when rebasing onto the #3303 landing. I swept the rest of the diff for the same artifact — every other column-0 `///` in the changed files is a legitimate top-level type doc, so this was the only one.

**Re-validated at the merged head `54b24bcc`:** `--gate fast` **3975 / 0**, `Area=Corpus` **83 / 0**. (Counts moved because the merge brought in new tests.)

Two notes on your review, since both are useful signal:

- Your `env=0` result — `AnEnvironmentOffRequestDoesNotSilentlyDisarmTheSuite` failing — matches what GPT found from the other direction, and is the designed behavior rather than a defect: the teeth tests are supposed to fail when someone disarms the suite. I confirmed the pre-existing structural teeth test from #3289 fails identically under `=0`, so this isn't new.
- Agreed on not treating any of these timings as a measurement on a shared machine. That's precisely why the corrected figure is framed as "below the noise floor" with the spread recorded, rather than as a number to quote.

Still `Ready to merge` pending CI on the new head.

richlander:
Ready to merge

CI green at `54b24bcc` (test, build-net10, markdownlint, csharp-diff-smoke, changes; il-diff-smoke and pack path-gated out). Both review items addressed; both adversarial reviews were clean at `312468f7`, and the delta since is documentation-only plus a merge from `main`, re-validated locally (`--gate fast` 3975 / 0, `Area=Corpus` 83 / 0).

Non-blocking follow-up only: #3317.

richlander:
Re-reviewed at head `54b24bcc` (base `main`, `MERGEABLE/CLEAN`). **Both of my findings are fixed, and the fix is better than what I asked for. Approving — no blocking findings.**

I asked for a wording change. What landed instead states the property that actually holds and explains *why the weaker-sounding claim is the stronger one in practice* — the level cannot be moved independently of `Enabled`, the only setter is `Enabled`'s and it is private, and the one public mover is pinned to a single call site. That is the correct resolution: not softening the claim until it is safe, but replacing it with the true one.

The commit message also names the cause of the column-0 line ("my scripted conflict resolution during the rebase introduced [it]") rather than silently fixing it. Small thing, but it is the difference between a fix and a fix you can learn from.

## What I verified this round

**The merge is content-preserving for the PR's own work.** `dcbbb849` merges `bf60a965`, and main had moved under two files this PR edits (`IrImporterTests.cs`, `IteratorReconstructionPassTests.cs`) — the exact place a merge could quietly undo a fixture fix. It did not. All five declarations survive byte-for-byte, including both `[intType, intType]` iterator fixtures with their explanatory comments.

**The new doc's load-bearing claim is enforced, not asserted.** "Pinned … to a single call site" is a strong claim, so I tried to break it with the spelling a naive matcher would miss — a decline buried in disabled preprocessor trivia:

```csharp
#if DECLINE_VALIDATION_PROBE
ILInspector.Decompiler.Pipeline.IrInvariants.DisableForShippedTool();
#endif
```

Caught:

```
Assert.Equal() Failure: Collections differ
Expected: ["src/dotnet-inspect/Program.cs"]
Actual:   ["src/dotnet-inspect/Program.cs", "tools/DecompilerHarness/Program.cs"]
```

`OnlyTheShippedToolEntryPointDeclinesValidation` re-parses every preprocessor configuration, scans the whole repository rather than a source-root list, matches by identifier so aliasing and method groups do not evade it, and carries its own anti-vacuity assertion (`sites.Count > 0` … "the census would pass vacuously"). It also *states its residual gaps* — env `0`, reflection with a computed name, out-of-tree `<Compile Include>` — and routes the first to coverage-by-consequence instead of pretending the scan covers it.

That paragraph is the thing I most want other guards in this repo to copy. Compare #3321: a guard that compared an inventory to a hardcoded list and passed with every case deleted. This one I actively tried to defeat and could not, and where it *can* be defeated it says so out loud. Enumerating your own blind spots is what separates a guard from a green checkmark.

**Negative control still exact, at a head 165 tests larger.**

```
CheckSemantics => false  →  Total: 3975, Failed: 2
  IrInvariantCheckTests.PipelineRunner_ThrowsWhenAPassLeavesAnUnresolvableSlot
  IrInvariantsHostContractTests.SemanticLevelIsOnByDefault_AndTracksTheStructuralLevel
```

Same two tests as at `312468f7`, nothing collateral.

**Indentation fixed** — `:70` now sits at 4 spaces with the rest of the block.

| Check | Result | PR claimed |
| --- | --- | --- |
| Build (`Release`, solution) | 0 errors | — |
| `--gate fast` | **3975 / 0** | 3975 ✅ |
| `Area=Corpus` | **83 / 0** (19.9s) | 83 ✅ |
| `src/dotnet-inspect.Tests` | **2201 / 0** (12 skipped) | — |
| Negative control | exactly 2, as intended | — |
| Census tamper (`#if`-hidden decline) | caught | — |

Both stated numbers reproduce exactly on my machine.

## One optional nano-nit, take it or leave it

The new paragraph says the opt-out is "pinned by #3303's public-surface census to a single call site." Two different tests do those two jobs: `ThePublicSurfaceIsExactlyTheShippedToolOptOut` pins *which members exist*, and `OnlyTheShippedToolEntryPointDeclinesValidation` pins *the call site*. Both are #3303's, so this is fair shorthand — I only mention it because it sits in a paragraph whose entire subject is being exact about which guarantee is held by what, and the source scan is the more impressive of the two to be pointing at.

Nothing else. Ship it.

PR #3318, opened 2026-07-27
Add mdi metadata inspector tool and shared renderer over the table projection

richlander:
## What

Adds **`mdi`** (metadata inspector), the first presentation consumer of the ECMA-335 `MetadataTableProjection` landed in #3301, plus a reusable renderer library it shares with a future `dotnet-inspect metadata` lens. Design context: #3282.

## Behavioral claim

`mdi <assembly>` renders the projected metadata tables as Markdown (default), TSV, or JSONL, with resolvable handle targets, explicit truncation, and visible failures.

```
mdi <assembly> [--table|-t <Names>] [--format|-f md|tsv|jsonl]
    [--max-rows|-n N] [--max-bytes N] [--max-chars N]
```

- **`src/DotnetInspector.MetadataRendering`** — `MetadataProjectionRenderer` maps a projection to Markout tables. It lives in the product-side `DotnetInspector.*` family, **not** in `ILInspector.Metadata` (which stays presentation-free): the renderer is a *sibling* of the Metadata layer, so a future lens reuses the same code.
- **`src/mdi`** — `PackAsTool` / `PublishAot` System.CommandLine front-end mapping flags onto `MetadataProjectionOptions` and delegating all output to the renderer. Missing files, native images, and unreadable metadata surface as errors on stderr with exit 1 — never success-shaped empty output.

Rendering is deliberately **lossy** (a human/inspection view); the projection **model** stays the lossless source of truth, which keeps a future `mdv` oracle able to diff against the model rather than rendered text.

## Evidence

- 17 tests pass (`dotnet run --project tests/DotnetInspector.MetadataRendering.Tests -c Release`): synthetic per-`MetadataValue`-case coverage (Nil/Scalar/Flags/HeapReference incl. truncation & blob/Handle incl. nil & display/Range/Malformed), format shape (md pipe table, tsv tab + `Table` column, jsonl `table`/`rid` keys), self-assembly end-to-end, and the `Execute` error contract.
- Full solution builds clean: `dotnet build dotnet-inspect.slnx -c Release` (0 warnings, 0 errors).
- Real-binary smoke test on `ILInspector.Metadata.dll` across md/tsv/jsonl and error paths (missing file, unknown table) confirms resolved handles (`AssemblyRef[1] (System.Runtime)`), truncation headings (`showing 3 of 426 rows`), self-identifying JSONL rows, and exit-1 errors.

## Non-action boundary

[...]

PR #3322, opened 2026-07-27
Add mdv cross-implementation oracle for the metadata-table projection

richlander:
## Summary

Adds an **mdv cross-implementation oracle** for the metadata-table projection (MetadataTableProjector). mdv is Roslyn's independent metadata viewer; since mdv and the projection are two unrelated readers of the same ECMA-335 image, their agreement is strong external evidence that the projection enumerates tables and decodes heaps correctly.

Stacked on #3318 (base `feature/mdi-metadata-inspector`). Test-only; no product code changes.

## What it checks

Over the product assembly under test, four checks:

- **`RowCounts_MatchMdv_ForSharedTables`** — per-table physical row-count parity for every table both tools expose (8 shared tables here: AssemblyRef, CustomAttribute, Field, MemberRef, Module, Param, TypeDef, TypeRef). Guarded so a parsing regression that compares too few tables fails loudly instead of passing vacuously.
- **`MethodDef_IsProjected_ButOmittedByThisMdv`** — pins the one asymmetry: this mdv build folds methods into `TypeDef.MethodList` ranges rather than emitting a standalone `MethodDef (0x06)` table, which is why MethodDef is excluded from the count comparison. A future mdv that emits the table fails this test and forces us to widen the comparison.
- **`ModuleMvid_MatchesMdv`** — Guid-heap decode parity.
- **`AssemblyRefNames_MatchMdv`** — multi-row String-heap decode parity.

## Design

- **Optional external tool.** mdv is discovered on `PATH` / `~/.dotnet/tools`; every test skips via `Assert.SkipUnless` when mdv is absent (same pattern as `ILDisassemblerComparisonTests`). mdv runs at most once per test run (cached `Lazy<string?>`); any launch/timeout/non-zero exit turns into a skip, never a hard failure.
- **Harness boundary stays clean.** The oracle only *reads mdv output and the product projection and compares them*. It does not reconstruct, normalise, or substitute for any product behaviour. Comparison is deliberately restricted to facts both tools expose robustly (physical row counts; a few scalar heap decodes), so a formatting quirk in either tool can't masquerade as a projection defect.
- Compares against the **projection model** (`MetadataTableView.RowCount` / typed `MetadataValue` cells), not rendered text.

## Evidence

Local run against the product assembly (mdv 2.0.0-beta1 installed):

- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release -- -class ...MdvOracleComparisonTests` → **Total: 4, Failed: 0, Skipped: 0** (mdv present, all agreed).
- Full project: `dotnet run --project tests/ILInspector.Metadata.Tests -c Release` → **Total: 919, Failed: 0, Skipped: 0**.
- Manually confirmed mdv's dumped tables (16) and that all 8 shared-table row counts match the projection exactly (e.g. CustomAttribute 6621, Param 5155, MemberRef 1929).

On CI (mdv not installed) the four oracle tests skip cleanly.

### DIFF
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -334,6 +334,42 @@ enumeration order. All are cheap and SRM-available
 (`GetTableRowCount`, `GetTableRowSize`, `MetadataTokens.GetHeapOffset`), so they
 are a knob the model can expose, not a wall.
 
+## Implemented: the `mdi` tool and the shared renderer
+
+The first presentation consumer of the projection is **`mdi`** (metadata
+inspector), a standalone tool that renders the tables the way `mdv` does:
+
+- **`src/mdi`** — a `PackAsTool` / `PublishAot` command
+  (`ToolCommandName=mdi`) whose System.CommandLine front-end maps flags onto
+  `MetadataProjectionOptions` and delegates all output to the renderer below.
+  Surface: `mdi <assembly> [--table|-t <Names>] [--format|-f md|tsv|jsonl]
+  [--max-rows|-n N] [--max-bytes N] [--max-chars N]`. Missing files, native
+  images, and unreadable metadata surface as visible errors, never
+  success-shaped empty output.
+- **`src/DotnetInspector.MetadataRendering`** — a small reusable library holding
+  `MetadataProjectionRenderer` (projection → Markout tables). It lives in the
+  product-side `DotnetInspector.*` family, **not** in `ILInspector.Metadata`,
+  which stays presentation-free: the renderer is a sibling of the Metadata layer
+  that consumes the projection, so a future `dotnet-inspect metadata` lens reuses
+  the same code the tool uses today.
+
+Renderer contract. Rendering is a deliberately **lossy** human/inspection view;
+the projection model stays the lossless source of truth (for example for the
+oracle below). Three invariants hold regardless: every cell renders from exactly
+one `MetadataValue` case; a `Malformed` cell keeps a visible `!malformed:`
+marker; and a bounded preview is suffixed with `…` so it is never mistaken for a
+whole value. A leading row-id column lets a reader cross-reference a resolved
+handle target (rendered `TypeRef[5] (System.Object)`) back to its row.
+
+Formats differ only in how a row's table is identified. Markdown introduces each
+table with a `## <Name> (rows)` heading over a pipe table; TSV and JSONL carry a
+leading `Table` column so every row self-identifies, keeping those outputs pure
+machine-readable streams (one `WriteTable` block per table).
+
+The `mdv` oracle is a follow-up increment: because it diffs against the
+projection **model** (not `mdi`'s rendered text), the renderer is free to be
+human-friendly without weakening the oracle.
+
 ## Layer placement
 
 The projection lives in the **Metadata layer** (`ILInspector.Metadata`), beside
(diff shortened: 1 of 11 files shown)


richlander:
## Review reconciliation

**Adversarial review — 1 model (MAI-Code), clean.**

Tier rationale: test-only, no product code, worst case is a false-green test, and the tests skip cleanly where mdv is absent (e.g. CI). That places this in the *more-than-trivial, not high-risk* tier → a single MAI-Code review.

- **MAI-Code** reviewed the exact head `0d87ed2b` in an isolated worktree with mdv installed, exercised the test class locally, and found **no significant issues** (tests passed). Attack points covered in the prompt: false-green / vacuous pass, parsing correctness (row-line / table-header / MVID / quoted-name regexes, table-end detection, Debug-Directory-guid scoping), harness-boundary compensation, skip-on-absent/broken-mdv paths, and process-drain deadlock/leak risk.

No findings → no fix commits; head is unchanged at `0d87ed2b`.

### Remaining gate: CI (blocked by the stacked base)

CI's `pull_request` trigger is scoped to `branches: [main]`, so this PR gets no CI while its base is `feature/mdi-metadata-inspector` (#3318). CI will run once #3318 merges and this branch retargets to `main`.

Local validation stands in for now (this change is isolated to one test project):

- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release` → **919 passed, 0 failed, 0 skipped** (mdv present, all 4 oracle tests agreed).

**Not yet `Ready to merge`:** the only outstanding item is CI on a `main` base, which is gated on #3318. Once #3318 merges and CI is green here, this is ready.

richlander:
## Follow-up: calibration / mutation checks (head `a1d4579b`)

Per review feedback that a never-red oracle is only half-validated, added four checks proving the oracle *catches* discrepancies, not just agrees:

- Three **synthetic-fixture parser tests** (no mdv → run on CI): exact-count correctness while ignoring the Debug Directory preamble / rules / column-name rows; off-by-one **sensitivity** to a single dropped row; and MVID extraction **scoped** to the Module row (not the preamble `guid={...}`).
- One **seeded-discrepancy test** against real mdv output (skips without mdv): removing one real mdv row makes the parsed count disagree with the projection by exactly one — the same comparison the headline test relies on.

The two synthetic tests are mutually pinning: one asserts the absolute counts (1, 4), the other asserts a one-row removal moves the count by exactly one, so an insensitive or miscounting parser fails at least one.

### Review

**MAI-Code** re-reviewed the exact head `a1d4579b` in an isolated worktree (with mdv installed): *"No significant issues… the new calibration tests are genuine sensitivity checks, and the mdv-gated seeded-discrepancy case skips cleanly when mdv is unavailable."* No findings → no fix commits.

### Validation

- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release` → **923 passed, 0 failed, 0 skipped** (mdv present). Oracle class alone: **8 passed**.

Readiness unchanged: the only outstanding gate is CI on a `main` base, which is blocked by the stacked base #3318. Once #3318 merges and CI is green here, this is ready.

richlander:
## Re-stacked onto the rebased base (head `9fc4e5b8`)

The base branch (#3318) was rebased onto the latest `main`, which left this branch carrying stale duplicate copies of PR 1's commits. I re-stacked with `git rebase --onto` so this PR now contains **only** the two oracle commits; its diff is exactly the single file `MdvOracleComparisonTests.cs` (446 lines).

This was a **mechanical re-stack with no content change** — the oracle file is byte-identical to the reviewed version (verified: `git diff` between the prior head and `9fc4e5b8` is empty). The two prior clean MAI-Code reviews therefore still apply. Local re-run on the new base: oracle class **8 passed, 0 failed**; full project **923 passed**.

Merge order for the stack: merge #3318 first; GitHub then retargets this PR to `main` and CI runs the merge path here. Readiness gate is unchanged (CI on a `main` base).

PR #3323, opened 2026-07-27
Share one lexical scan for declaration-end detection (#3300)

### LINKED ISSUE #3300: IsSelfTerminatingLine does not strip trailing comments, triggering an unnecessary forward scan
`SourceLinkResolver.IsSelfTerminatingLine` decides whether an extracted range already ends at a natural stopping point. It inspects the raw trailing characters of the line and does not strip a trailing comment, so a declaration whose line ends in a comment containing a brace — for example:

```csharp
public int Target() => 0; // opens nothing {
```

is not recognized as self-terminating, and `ExtractMethodBody` falls through to the forward brace scan it did not need. The extracted body can then pick up following lines.

This is cosmetic rather than incorrect in the common case, and it is pre-existing: it predates #3287, where it was found while writing tests for the surrounding `HasUnclosedBlock` work. The test written for it was retargeted rather than expand that PR's scope.

`HasUnclosedBlock` already knows how to skip line comments, block comments, char literals, and regular, verbatim, and verbatim-interpolated strings. The likely fix is to share that scanning with `IsSelfTerminatingLine` so the two cannot disagree about where a line really ends, rather than adding a second, simpler comment stripper next to it.

richlander:
Fixes #3300.

## What was wrong

`IsSelfTerminatingLine` decided whether an extracted range already ended at a natural stopping point by inspecting the raw trailing characters of the line. A declaration whose line ends in a comment was therefore not recognized as terminated:

```csharp
public int Target() => 0; // opens nothing {
```

`ExtractMethodBody` fell through to the forward brace scan it did not need, which can append the enclosing type''s closing brace to the extracted source.

`HasUnclosedBlock`, sitting immediately below it, already knew how to skip line comments, block comments, char literals, and regular, verbatim, and verbatim-interpolated strings — but it answered a *different* question about the *same* text. Two scanners over one line can disagree about where that line really ends, which is the actual defect; the trailing comment is just where it showed.

## What changed

Both are replaced by a single `EndsDeclaration`, which runs one lexical scan (`ScanLine`) across the range and reads both answers off the same state:

- the **last significant character** — the last non-whitespace character outside a comment — which must be `;` or `}`
- the **brace depth**, counting only structural braces

Two behaviors fall out of sharing the scan, both of which I consider corrections rather than side effects:

- A `;` or `}` inside a string or comment no longer reads as a terminator, because it terminates nothing.
- A raw string literal closed on its own line is now scanned properly instead of abandoning the range to the forward scan. A raw string that *spans* lines stays conservative, exactly as before.

The single-line exemption from depth analysis is preserved, so a one-line declaration still owns no brace below it.

## Evidence

Five new tests in `ExtractMethodBodyTests`, covering the issue repro, a trailing block comment, a `;` that appears only inside a trailing comment (must *not* count), a multi-line raw string (must still fall back), and a single-line raw string (must not).

| Suite | Result |
| --- | --- |
| `tests/ILInspector.Metadata.Tests` | 920 / 920 |
| `src/DotnetInspector.Services.Tests` | 319 / 319 |
| `src/dotnet-inspect.Tests` | 2188, 10 skipped (ilasm/ildasm absent) |

Two CLI failures are accounted for and neither is mine:

- `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden` — reproduced on clean `origin/main` at `3fc6f1c8` in a separate worktree, same `Assert.DoesNotContain()` assertion.
- `PerformanceTriagePredicates_DoNotAlterDiscovery(command: "library")` — the suite''s known parallel flake; passes in isolation.

End-to-end canary against a real platform assembly, byte-identical to `origin/main`:

```text
dotnet-inspect member "System.Text.Json.JsonSerializerOptions" --platform System.Text.Json MaxDepth -S "Original Source"
```

## Risk

This is a behavior change to a heuristic shared by **every** member''s source extraction, not just the case in the issue — the same surface that produced every review finding on #3287. Sized as high risk accordingly.

### DIFF
--- a/src/ILInspector.Metadata/SourceLinkResolver.cs
+++ b/src/ILInspector.Metadata/SourceLinkResolver.cs
@@ -154,9 +154,11 @@ public class SourceLinkResolver
         // next "}" below it closes the enclosing type instead (issue #3278). A range that still
         // has a block open does own one, even when its last line ends in ";": a signature whose
         // "{" sits on the declaration line ends its sequence range on the last statement.
-        bool endsAtDeclaration = startsAtDeclaration && end >= 1 && end <= lines.Length
-            && IsSelfTerminatingLine(lines[end - 1])
-            && !HasUnclosedBlock(lines, from, to);
+        //
+        // Both answers read the same lexical state, so one scan produces both. A trailing
+        // comment must not hide the terminating ";" (issue #3300), and a brace inside a comment
+        // or a literal must not count as structural.
+        bool endsAtDeclaration = startsAtDeclaration && EndsDeclaration(lines, from, to);
 
         // Scan forward to include the closing brace.
         for (int i = to; !endsAtDeclaration && i < Math.Min(to + 3, lines.Length); i++)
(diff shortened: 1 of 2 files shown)


richlander:

Two-model review, GPT (`gpt-5.6-sol`) and MAI (`mai-code-1-flash-picker`), each isolated in its own detached review worktree. Sized as high risk because this changes a heuristic shared by every member''s source extraction.

### Round 1 — head `420a4aab`

| --- | --- | --- |
| GPT | Clean. Exercised the requested boundary, string, comment, raw/verbatim literal, single-line, and range cases against both head and base, plus an exhaustive scanner probe — 898,779 `ScanLine` and 14,123,676 `IndexOfQuoteRun` calls, no exceptions or hangs. Bounds behavior matched base, including a pre-existing exception on a negative `endLine`. | — |
| MAI (Medium) | `EndsDeclaration` overwrote its saved terminator on every line, so a blank, whitespace-only, or comment-only line below the declaration reset it to `'\0'`. The range then looked unterminated, the forward scan ran, and the enclosing type brace was appended. | Fixed in `299458e4`. |

The terminator is now replaced only when a line actually contributes a significant character — which is what "the range''s last significant character" was meant to say in the first place.

**This one was pre-existing, not a regression I introduced.** Base looked only at the range''s *last* line, so a blank last line failed the same way. Verified by loading both assemblies side by side and running the same input:

```text
--- BASE (main) ---
    public int Target() => 0;

}
--- HEAD (fix) ---
public int Target() => 0;
```

It produces visibly wrong output, so it is fixed here rather than documented — consistent with how the equivalent findings were handled on #3287.

### Round 2 — head `299458e4`

Both reviewers re-verified the exact head and are clean. GPT confirmed the diff is confined to `EndsDeclaration`, its doc comment, and two tests, and re-ran its carry-forward edge cases. MAI rebuilt its harness against the fresh binary and probed the blank and comment-only carry-forward cases plus brace-depth interaction. Metadata suite: **922 / 922** for both.

### Validation at `299458e4`

| Suite | Result |
| --- | --- |
| `tests/ILInspector.Metadata.Tests` | 922 / 922 |
| `src/DotnetInspector.Services.Tests` | 319 / 319 |
| `src/dotnet-inspect.Tests` | 2188, 10 skipped; the two failures are accounted for in the PR description and neither is caused by this change |

richlander:
Ready to merge

PR #3325, opened 2026-07-27
Explain bodyless members instead of rendering nothing (#3299)

### LINKED ISSUE #3299: Bodyless members render an empty Original Source section and exit 0
`member ... -S "Original Source"` on a bodyless member renders a section heading with no content and exits 0. Nothing tells the caller that the member has no IL body and therefore no source range to show — the output is success-shaped and empty.

Reproduce on a path unrelated to any recent change:

```text
dotnet-inspect member "System.Collections.IEnumerator" --framework runtime MoveNext -S "Original Source"
```

An abstract interface method has no body, so there is no sequence point range to resolve, and the section renders as a bare heading. The same is now true of a bodyless property accessor, which is consistent — but consistently unhelpful.

This is long-standing and predates #3287; it was raised during that PR's review (GPT, round 1) and recorded there as an explicit non-action so the accessor wiring would not grow scope.

AGENTS.md says failure must stay visible and that we do not turn resolution failures into success-shaped empty output. An empty section is exactly that shape. The fix is probably a short explanatory line stating that the member has no body, rather than an error — but the right wording and whether it should affect the exit code are worth deciding deliberately.

richlander:
## What

`Original Source` rendered *nothing at all* for a member with no IL body — no
section, no text — and still exited 0. A caller could not distinguish a
complete answer ("this member has no source, by construction") from a silent
failure ("resolution broke and we swallowed it").

Now a definitely-bodyless member gets an explanation in the section:

```console
$ dotnet-inspect member "JsonConverter<T>" --platform System.Text.Json Read -S "Original Source"
## Original Source

```csharp
// This member has no IL body, so it has no authored source to show.
```
```

Exit stays 0: this is a correct, complete answer, not a failure.

## How

- `PdbContext.MethodHasBody(int)` (Metadata owns the metadata fact) answers
  `true` / `false` / `null`.
- `ResolveMethodSourceAsync` asks *before* attempting resolution, so the answer
  does not depend on how far resolution got, and flows the flag to the render
  sites through `MemberOptions`.
- The two `OriginalSourceCode` assignment sites render the note when there is no
  resolved source and the member is definitely bodyless.

## Boundaries — what this deliberately does *not* claim

- **Only a definite "no" counts.** An unreadable or non-MethodDef token is
  `null` (unknown), not "no body".
- **Reference assemblies answer `null` for every token.** A ref assembly strips
  all IL, so its RVAs describe the image's surface-only nature, not the method.
  Without this guard every ref-assembly member would falsely claim to be
  bodyless.
- **The other five unresolved paths are unchanged** (`!fetchSource`, no PDB, no
  SourceLink, unresolved method, no content, exception). They stay silent rather
  than claiming a reason they cannot prove. Narrowing those is separate work.
- **Source Diff is unaffected.** The note is prose about the member, not source
  text, so a `MemberCodeView.OriginalSourceUnavailable` flag keeps it out of the
  diff input; the diff still reports `Original Source unavailable`.

Wording and placement follow the existing `SourceTextDiffRenderer` precedent: a
comment-prefixed line inside the section's fence.

## Validation

- New CLI tests: abstract method, interface method, close negative (member
  *with* a body still renders real source and never shows the note), and the
  Source Diff non-pollution case.
- Metadata **915/915**, Services **319/319**, CLI **2191** (10 skipped) with one
  failure, `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`,
  reproduced unchanged on clean `origin/main` (`3fc6f1c8`) — pre-existing, not a
  regression.

Fixes #3299.

### DIFF
--- a/src/ILInspector.Metadata/PdbContext.cs
+++ b/src/ILInspector.Metadata/PdbContext.cs
@@ -359,19 +359,21 @@ public class PdbContext : IDisposable
         if (!_peReader.HasMetadata)
             return null;
 
-        var handle = MetadataTokens.Handle(methodToken);
-        if (handle.Kind != HandleKind.MethodDefinition)
-            return null;
-
         try
         {
+            // Handle() rejects an invalid token by throwing, and an inspected assembly is
+            // untrusted input, so decode inside the guard rather than ahead of it.
+            var handle = MetadataTokens.Handle(methodToken);
+            if (handle.Kind != HandleKind.MethodDefinition)
+                return null;
+
             var reader = _peReader.GetMetadataReader();
             if (IsReferenceAssembly(reader))
                 return null;
 
             return reader.GetMethodDefinition((MethodDefinitionHandle)handle).RelativeVirtualAddress != 0;
         }
-        catch (Exception ex) when (ex is BadImageFormatException or ArgumentOutOfRangeException)
+        catch (Exception ex) when (ex is BadImageFormatException or ArgumentException)
         {
             return null;
         }
@@ -391,17 +393,7 @@ public class PdbContext : IDisposable
         {
             foreach (var handle in reader.GetAssemblyDefinition().GetCustomAttributes())
             {
-                var attribute = reader.GetCustomAttribute(handle);
-                if (attribute.Constructor.Kind != HandleKind.MemberReference)
-                    continue;
-
-                var parent = reader.GetMemberReference((MemberReferenceHandle)attribute.Constructor).Parent;
-                if (parent.Kind != HandleKind.TypeReference)
-                    continue;
-
-                var typeRef = reader.GetTypeReference((TypeReferenceHandle)parent);
-                if (reader.StringComparer.Equals(typeRef.Name, "ReferenceAssemblyAttribute")
-                    && reader.StringComparer.Equals(typeRef.Namespace, "System.Runtime.CompilerServices"))
+                if (IsReferenceAssemblyAttribute(reader, reader.GetCustomAttribute(handle)))
                 {
                     result = true;
                     break;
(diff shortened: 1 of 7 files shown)


richlander:

Two-model review (GPT-5.6 and MAI-Code), each in an isolated read-only worktree, two rounds on fixed heads.

### Round 1 @ `6e2f4b4e`

**MAI-Code — High, `PdbContext.MethodHasBody`.** `MetadataTokens.Handle(methodToken)` sat *outside* the `try`, so an invalid token threw `ArgumentException` rather than returning the documented `null` ("unknown"). Inspected assemblies are untrusted input per `docs/design/untrusted-data-threat-model.md`.

**Verified and fixed.** I reproduced it independently — `MetadataTokens.Handle(-1)`, `0xFFFFFFFF`, and `0x7F000001` all throw `ArgumentException: Invalid token`. The decode now happens inside the guard, and the catch widened from `ArgumentOutOfRangeException` to `ArgumentException` (which subsumes it).

**GPT-5.6 — clean**, having probed all seven attack points.

### A second defect, found while writing the regression test for the first

Writing `MethodHasBodyTests` surfaced a gap **both reviewers had passed**: my reference-assembly guard only recognized `ReferenceAssemblyAttribute` when the custom attribute's constructor was a `MemberReference`. A reference assembly that *defines* that attribute itself encodes the constructor as a `MethodDef` — and `System.Runtime.dll` in `Microsoft.NETCore.App.Ref` does exactly that. **The guard missed precisely the images it exists to protect**, so every one of their members could have claimed to be bodyless. Both encodings are now handled.

This is why the ref-assembly test runs against a real targeting-pack assembly rather than a synthetic fixture: a synthetic one would have reproduced my own wrong assumption.

### Round 2 @ `141eeb6b` — both reviewers clean

Notable follow-up evidence:

- **GPT surveyed 2,150 real reference assemblies**: 13 use the `MethodDef` encoding, 2,137 use `MemberRef`/`TypeRef`, and **no third encoding exists in practice**. Both branches of the new helper are load-bearing.
- **GPT confirmed the false positive was genuinely reachable**: most platform flows zero the token on a ref/runtime path mismatch, and the `System.Runtime` methods it happened to test had non-zero stub RVAs — but a direct `--library` against a self-defining, IL-stripped reference assembly was **actually vulnerable**. So this was a real bug, not a theoretical one.
- **MAI independently confirmed both fixes** against `System.Runtime.dll` (targeting pack), `netstandard.dll`, `mscorlib.dll`, and `System.dll`.
- Both confirmed the ref-assembly test **runs rather than skips** on the review machines (924 metadata tests, **0 skipped**), and that round-1 conclusions still hold: Source Diff stays unpolluted and members that do have authored source are unchanged.

### Non-actions

No finding was dismissed. Nothing was deferred.

The five other unresolved-source paths in `ResolveMethodSourceAsync` (`!fetchSource`, no PDB, no SourceLink, unresolved method, null content, exception) remain silent by design — narrowing those means proving a reason for each, which is separate work, not a gap in this change.

richlander:
Ready to merge

PR #3332, opened 2026-07-27
Make "only the tools ship" a property of the projects (#3330)

### LINKED ISSUE #3330: Internal libraries are packable by default; 'only dotnet-inspect ships' is enforced by the pack argument, not the projects
Found while verifying a risk claim on #3306. Not a live defect — a latent packaging footgun that the repository already has a convention against, applied everywhere except the internal libraries.

## What is true today

`AGENTS.md` states pack and publish flows "remain separate and build `src/dotnet-inspect` directly", and that is what the workflows do — every `dotnet pack` targets the CLI project:

```
ci.yml:583,596,606        dotnet pack src/dotnet-inspect ...
release.yml:169,199,211   dotnet pack src/dotnet-inspect ...
```

So nothing unintended ships. But the reason nothing ships is the **argument to `dotnet pack`**, not a property of the projects. The internal libraries are packable and produce packages on demand:

```
$ dotnet pack src/ILInspector.Decompiler -c Release
Successfully created package '.../artifacts/package/release/ILInspector.Decompiler.1.0.0.nupkg'.
```

## The convention exists but stops short of the libraries

`IsPackable=false` is set on 58 of 81 projects — fixtures, test projects, and sample apps. The 23 without it include the two shipping tools and, notably, **16 internal libraries**:

```
src/DotnetInspector.Core
src/DotnetInspector.MetadataRendering
src/DotnetInspector.Packages
src/DotnetInspector.Services
src/ILInspector.Analysis
src/ILInspector.CSharp
src/ILInspector.CallGraph
src/ILInspector.ControlFlow
src/ILInspector.Decompiler
src/ILInspector.Findings
src/ILInspector.Instructions
src/ILInspector.Metadata
src/ILInspector.MetadataPrimitives
src/ILInspector.Research
src/ILInspector.Text
tests/TestExtensions
```

These default to the SDK's `IsPackable=true`.

## Why it is worth closing

The concrete exposure is that any solution-wide pack step — `dotnet pack dotnet-inspect.slnx`, a future publish job, or a contributor running pack at the repo root to debug something — emits 16 packages at version `1.0.0` for libraries that have no versioning story, no API-stability commitment, and no intent to be consumed externally. Once such a package reaches a feed, the layering rules in `AGENTS.md` become an external compatibility surface rather than an internal design constraint.

It also affects how "no API break" is reasoned about in review. On #3306 the stated rationale for a public-API removal being safe was that the project is not packed. That conclusion was right, but the premise was the pack target rather than a project opt-out — and a premise that lives in a workflow argument is easy to invalidate from a different file than the one being reviewed.

This is the #3271 shape applied to the build: the intent is documented and currently honored, but enforced by convention rather than by the thing it constrains.

## Suggested fix

Set `<IsPackable>false</IsPackable>` on the internal libraries, most cheaply as a default in `Directory.Build.props` with the shipping tools opting back in. That makes "only `dotnet-inspect` ships" a property of the projects, so it holds regardless of how pack is invoked, and gives reviewers a local answer to "can this library's public API break anyone?"

Low priority and entirely mechanical. Filing so the reasoning is recorded rather than rediscovered next time an API-surface question comes up.

richlander:
Closes #3330.

## The gap

Nothing unintended ships today, but the reason is the **argument to `dotnet pack`**, not a property of the projects. The internal libraries inherited the SDK`s `IsPackable=true`.

Measured on `bf60a965`, before the change:

```
$ dotnet pack dotnet-inspect.slnx -c Release
17 packages
```

```
DotnetInspector.Core.1.0.0            ILInspector.Findings.1.0.0
DotnetInspector.MetadataRendering.1.0.0  ILInspector.Instructions.1.0.0
DotnetInspector.Packages.1.0.0        ILInspector.Metadata.1.0.0
DotnetInspector.Services.1.0.0        ILInspector.MetadataPrimitives.1.0.0
ILInspector.Analysis.1.0.0            ILInspector.Research.1.0.0
ILInspector.CSharp.1.0.0              ILInspector.Text.1.0.0
ILInspector.CallGraph.1.0.0           RunFaster.AllocationFixture.1.0.0
ILInspector.ControlFlow.1.0.0         runfaster.0.1.0
ILInspector.Decompiler.1.0.0
```

One beyond the issue`s list: `RunFaster.AllocationFixture`, a benchmark fixture, also packed.

## Change

Inverts the default in the root `Directory.Build.props`, with `src/dotnet-inspect` and `src/runfaster` opting back in. A new project now inherits "does not ship", so shipping is the explicit act — the same shape as #3289`s inverted validation default, applied to the build.

The 58 projects that already set `IsPackable=false` keep their explicit setting. Removing them would touch 58 files for no behavior change; they are now redundant rather than wrong.

**`PackagingSurfaceTests`** holds both halves, because either alone is insufficient:

- `OnlyTheShippingToolsOptIntoPackaging` pins the opt-in set by set equality, ignoring `Condition` — a project that packs under *some* configuration ships under that configuration.
- `PackagingIsOffByDefaultForEveryProject` asserts the default is declared, and unconditionally. Without it the census passes vacuously in the way that actually matters: delete the default, every project is packable again, no project needs an opt-in, and the census still reports green. Verified below — that is not a hypothetical.

## Evidence

| Check | Before (`bf60a965`) | After |
| --- | --- | --- |
| `dotnet pack dotnet-inspect.slnx -c Release` | **17 packages** | **1** (`runfaster`) |
| `dotnet pack src/ILInspector.Decompiler -c Release` | 1 package | **0** |
| `dotnet pack src/dotnet-inspect -c Release -r any -p:PublishAot=false -p:OfficialBuild=true` | packs | **packs** (`dotnet-inspect.any.0.16.0`) |
| `src/dotnet-inspect.Tests` full run | — | **2203 passed, 0 failed**, 12 skipped |
| `markdownlint AGENTS.md` | — | clean |

Gate teeth, each tampered and reverted:

| Tamper | Result |
| --- | --- |
| delete `IsPackable` from `Directory.Build.props` | **1 failed** — `PackagingIsOffByDefaultForEveryProject`; the census stayed green, which is the point |
| give the default a `Condition` that does not hold | **1 failed** — same gate |
| add `<IsPackable>true</IsPackable>` to `ILInspector.Decompiler` | **1 failed** — `OnlyTheShippingToolsOptIntoPackaging` |
| untampered | **2 passed** |

### One pre-existing failure, not caused by this change

`dotnet pack src/dotnet-inspect -c Release -p:SelfContained=true -p:OfficialBuild=true` (the pointer package, `ci.yml:583`) fails locally with `NETSDK1047 ... doesn`t have a target for net11.0/linux-x64`. I reproduced it on a **clean `origin/main` worktree** before concluding that — it is a local SDK/RID condition on this machine, unrelated to packability, and CI`s `pack` job is the authority. Flagging rather than burying it. The `-r any` invocation, which exercises the same project`s packability, works.

## Docs

`AGENTS.md` said pack flows "build `src/dotnet-inspect` directly", which stays true. It now also records that packaging is off by default and that internal library public surface is an internal design constraint rather than an external compatibility surface — the review question that produced this issue.

## Risk

Mechanical and build-only. No product code, no test behavior, no output. The only behavior change is that projects which were never intended to ship no longer produce packages; both shipping tools verified to still pack. Proposing single-review tier.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -295,7 +295,12 @@ to these trait filters (e.g. `--gate fast`, `--gate no-corpus`); run
 `--gate list` for the table.
 
 Pack and publish flows remain separate and build `src/dotnet-inspect`
-directly.
+directly. Packaging is off by default (`IsPackable=false` in the root
+`Directory.Build.props`), so only `src/dotnet-inspect` and `src/runfaster` opt
+back in and no other project can ship however pack is invoked. Internal
+libraries have no versioning story and no API-stability commitment; treat their
+public surface as an internal design constraint, not an external compatibility
+surface. `PackagingSurfaceTests` pins both halves.
 
 ## Output contract
 
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -13,5 +13,17 @@
     <DefaultTargetFramework Condition="'$(DefaultTargetFramework)' == '' and '$(OfficialBuild)' == 'true' and '$(PublishAot)' == 'false'">net10.0</DefaultTargetFramework>
     <DefaultTargetFramework Condition="'$(DefaultTargetFramework)' == ''">net$(NETCoreAppMaximumVersion)</DefaultTargetFramework>
     <EnablePreviewFeatures>true</EnablePreviewFeatures>
+    <!--
+      Only the shipping tools produce packages, and that is a property of the
+      projects rather than of the argument each pack invocation happens to pass
+      (issue #3330). The internal libraries have no versioning story, no API
+      stability commitment, and no intent to be consumed externally; if one of
+      them reached a feed at 1.0.0, the layering rules in AGENTS.md would become
+      an external compatibility surface instead of an internal design
+      constraint. Off by default so a new library inherits "does not ship" and
+      shipping is the explicit act: src/dotnet-inspect and src/runfaster opt back
+      in, and PackagingSurfaceTests pins that set.
+    -->
+    <IsPackable>false</IsPackable>
   </PropertyGroup>
 </Project>
--- a/src/dotnet-inspect.Tests/PackagingSurfaceTests.cs
+++ b/src/dotnet-inspect.Tests/PackagingSurfaceTests.cs
@@ -1,0 +1,113 @@
+using System.Xml.Linq;
+
+namespace DotnetInspector.Tests;
+
+/// <summary>
+/// What this repository ships (#3330). Only the two tools produce packages, and
+/// that must be a property of the projects rather than of the argument each
+/// <c>dotnet pack</c> invocation happens to pass. Before the default was
+/// inverted, <c>dotnet pack dotnet-inspect.slnx</c> emitted 17 packages: the two
+/// tools plus 15 internal libraries and a benchmark fixture, all at
+/// <c>1.0.0</c>, for code with no versioning story and no intent to be consumed
+/// externally. A library that reaches a feed turns the layering rules in
+/// <c>AGENTS.md</c> into an external compatibility surface instead of an
+/// internal design constraint.
+/// <para>
+/// The enforcement is the root <c>Directory.Build.props</c>: with
+/// <c>IsPackable=false</c> inherited, a new project ships nothing unless someone
+/// writes the opt-in, so shipping is the explicit act. These tests hold the two
+/// halves that makes true — that the default is actually declared, and that the
+/// set of projects opting back in is exactly the shipping tools. They also give
+/// review a local answer to "can this library's public API break anyone?",
+/// which on #3306 had to be settled by reading two workflow files.
+/// </para>
+/// </summary>
+public sealed class PackagingSurfaceTests
+{
+    static readonly string[] ShippingProjects =
+    [
+        "src/dotnet-inspect/dotnet-inspect.csproj",
+        "src/runfaster/runfaster.csproj",
+    ];
+
+    /// <summary>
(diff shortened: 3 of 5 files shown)


richlander:
CI green on `8418c7d5` — all seven jobs, and notably `pack` **ran** rather than skipping, because this change touches its path gate. That is the job that matters here, and it settles the one loose end in the description.

All three shipping invocations still produce their packages under the inverted default:

```
dotnet pack src/dotnet-inspect -c Release -p:SelfContained=true -p:OfficialBuild=true
  → dotnet-inspect.0.16.0.nupkg
dotnet pack src/dotnet-inspect -c Release -r any -p:PublishAot=false -p:OfficialBuild=true
  → dotnet-inspect.any.0.16.0.nupkg
dotnet pack src/dotnet-inspect -c Release -r linux-x64 -p:OfficialAotBuild=true
  → dotnet-inspect.linux-x64.0.16.0.nupkg
```

The pointer package I could not build locally (`NETSDK1047`, reproduced on a clean `origin/main` worktree) builds fine in CI, confirming it was a local SDK/RID condition and not this change.

| Job | Result |
| --- | --- |
| `pack` | pass (4m41s) — all 3 tool packages produced |
| `test (ubuntu-26.04, linux-x64)` | pass (7m28s) |
| `build-net10` | pass (2m1s) |
| `csharp-diff-smoke` / `il-diff-smoke` | pass |
| `markdownlint` | pass |

Ready to merge.

PR #3335, opened 2026-07-27
Assert the carrier tripwire's sole-slot-0-binder guarantee

### LINKED ISSUE #3329: MarkLocalEliminated carrier tripwire documents a 'sole slot-0 binder' guarantee it does not assert
Follow-up from the #3285 review (merged as `587728be`). Non-blocking; the mechanism is sound and the natural contribution path is correct. This is about a documented guarantee that is stronger than the enforced one.

## Claim vs. enforcement

`InlineArraySpilledElementTests.cs:618` states the completeness tripwire is safe because the Theory asserts:

> …the factory actually builds a node of that kind **and that it is the sole slot-0 binder**, adding a factory then forces the carrier to be handled in `NodeBindsLocalSlot` (neither a name in a list nor a reused factory can silence both tests).

The first half is asserted — `FunctionContainsCarrier` at `:602`. The second half is not asserted anywhere. It is true of today's factories by construction, since each is a single-carrier `BufferBoundBy(...)` (`:571`), but nothing holds a future factory to it.

## Measured

Added an unhandled carrier to the IR and a factory that holds it **plus** a second, already-handled slot-0 binder:

```csharp
public sealed class ProbeCarrier : IrExpression
{
    public ProbeCarrier(int localIndex) { LocalIndex = localIndex; }
    public int LocalIndex { get; }
    ...
}

// in LocalSlotCarrierFactories:
[typeof(ProbeCarrier)] = () => BufferBoundByTwo(new ProbeCarrier(0), new StoreLocal(0, Object, BoxInt(1))),
```

```
InlineArraySpilledElementTests  Total: 39, Failed: 0
```

Both guards green with `ProbeCarrier` absent from `NodeBindsLocalSlot`. The `StoreLocal` carries the refusal; `FunctionContainsCarrier` is satisfied because a `ProbeCarrier` node is present. A function whose only slot-0 binder is a `ProbeCarrier` would then be marked eliminated — the false `Full` this whole mechanism exists to prevent.

## Why this is much weaker than #3321

Worth stating plainly so it is not over-weighted. In #3321 the *natural* fix silenced the guard: a contributor hitting the tripwire would add a name to a list, which was the obvious remedy and was precedented by two existing entries. Here the natural fix is `BufferBoundBy(new NewCarrier(0))`, which **does** force the switch to handle the kind. Reaching the hole requires deliberately writing a two-binder factory, which no existing factory resembles.

So: a comment-accuracy issue with cheap available hardening, not a trap.

## Two ways to resolve

**Enforce it** — parameterize each factory by slot index and assert both directions per row:

- `factory(0)` → `MarkLocalEliminated(0)` refuses (today's assertion), and
- `factory(1)` → `MarkLocalEliminated(0)` **succeeds**, since no node binds slot 0.

Any second slot-0 binder makes the `factory(1)` leg refuse and fails the row, which converts "sole binder" from a construction argument into an assertion. The factories already hardcode `0`, so it is a mechanical edit.

**Or match the text to the behavior** — drop the "and that it is the sole slot-0 binder" clause from `:618`. Zero risk, and it keeps the comment from claiming enforcement the tests do not perform, which is the #3271 shape.

Either is fine. What I would avoid is leaving the stronger claim in place, since the next person to extend this reasonably relies on it.

## Minor, same area

`FunctionContainsCarrier` (`:699`) wraps `property.GetValue(node)` in `catch { continue; }`. Swallowing every exception is pragmatic here — several IR properties cast or index `Children` and throw on synthetic nodes — but it means a carrier hidden behind a throwing property is silently not found, and the row would then pass for the wrong reason. Narrowing the catch, or asserting that at least one property read succeeded, would keep the helper honest. Very low priority.

richlander:
Closes #3329.

The `MarkLocalEliminated` carrier tripwire documented that every factory builds a function whose *only* slot-0 binder is the carrier under test, but nothing asserted it. This PR takes the issue's stronger **"Enforce it"** option rather than dropping the claim.

## The escape, measured

Reproduced without adding a synthetic IR type: delete the `IsPattern` arm from `NodeBindsLocalSlot`, and make the `IsPattern` factory a two-binder (`IsPattern` + `StoreLocal(0, …)`).

| Test file | Product tamper | Result |
| --- | --- | --- |
| `origin/main` | `IsPattern` arm deleted, factory made a two-binder | **38 passed, 0 failed** — silent hole |
| this PR | same | **1 failed** — `IsPattern` row, `Assert.Contains() Failure: Item not found in set` |
| this PR | `IsPattern` arm deleted only (no two-binder) | **1 failed** — pre-existing teeth preserved |

The last row confirms the change adds a leg rather than trading one detection for another.

## What changed

`src/ILInspector.Decompiler.Tests/InlineArraySpilledElementTests.cs` only. No product code.

- `LocalSlotCarrierFactories` is now `Func<int, IrFunction>` — each of the 12 factories takes the slot index to bind, replacing a hardcoded `0`.
- `MarkLocalEliminated_RefusesEveryLocalSlotBindingNodeKind` runs two legs per row: `factory(0)` → refusal (the existing assertions), and `factory(1)` → `Assert.Contains(0, EliminatedLocalSlots)`. The second leg can only pass if the carrier is the sole binder of the slot it was handed.
- The drift-tripwire comment now describes what is actually asserted.

### Why the locals became `[Buffer, Object]`

Slot 1 must be in range for the `factory(1)` leg. Slot 1 is deliberately *spellable*: if the refusal leg ever wrongly dropped slot 0, only `Object` would remain and the `HasUnrepresentableMetadataName` assertion would fail rather than stay incidentally true. `[Buffer, Buffer]` would have kept that assertion green for the wrong reason.

## On the minor finding — I think the stated direction is backwards

The issue says a carrier hidden behind a throwing property is "silently not found, and the row would then pass for the wrong reason." `FunctionContainsCarrier`'s only callers are `Assert.True(FunctionContainsCarrier(...))`, so a missed carrier makes the row **fail loudly**. The helper is already fail-closed; there is no silent-pass path.

I also measured it: instrumenting the `catch` and running every row records **0 throws**, so nothing depends on the swallow today.

The real defect in `catch { continue; }` is different and worth fixing — it also swallowed bugs in the helper itself, where an NRE or bad cast in the loop would read as "carrier absent". Narrowed to `catch (TargetInvocationException)`, which is what reflection wraps a property body's own exception in, with the measurement and the fail-closed direction recorded in the comment.

## Validation

- `-class ILInspector.Decompiler.Tests.InlineArraySpilledElementTests` — **38 passed, 0 failed**
- `dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- --gate fast` — **3973 passed, 0 failed**
- Teeth re-verified on the exact committed tree; `git status -s` confirms only the test file is modified.

## Review tier

Test-only change; no product code touched, and the added leg's teeth are demonstrated by tamper/revert above. Proposing single-review tier, consistent with #3270, #3306, and #3332.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/InlineArraySpilledElementTests.cs
+++ b/src/ILInspector.Decompiler.Tests/InlineArraySpilledElementTests.cs
@@ -544,30 +544,38 @@ public class InlineArraySpilledElementTests
     // PropertyName by trimming that prefix.
     static readonly MethodRef PatternAccessor = new(Object, "get_Value", Object, [], HasThis: true);
 
-    // One IrFunction factory per local-slot-binding carrier kind. Each builds a method
-    // whose only metadata local (slot 0) is the unspellable inline-array buffer, bound
-    // by a node of that kind, so MarkLocalEliminated must refuse to drop the slot —
-    // dropping it would hide the buffer's unspellable type and report a false Full
-    // (#3295). Aggregation-only carriers are bound through the parent that owns their
-    // index, matching how NodeBindsLocalSlot reaches them: DeconstructionTarget through
-    // DeconstructionAssignment, PropertySubpattern through PatternSwitchExpressionArm.Subpattern.
-    static readonly IReadOnlyDictionary<Type, Func<IrFunction>> LocalSlotCarrierFactories =
-        new Dictionary<Type, Func<IrFunction>>
+    // One IrFunction factory per local-slot-binding carrier kind, parameterized by the
+    // slot the carrier binds. Each builds a method whose slot 0 is the unspellable
+    // inline-array buffer; when the carrier binds slot 0, MarkLocalEliminated must refuse
+    // to drop it, because dropping it would hide the buffer's unspellable type and report
+    // a false Full (#3295). The slot parameter is what lets the Theory prove the carrier
+    // is the *sole* slot-0 binder rather than assert it by construction (#3329): the same
+    // factory pointed at slot 1 must leave slot 0 eliminable. Aggregation-only carriers
+    // are bound through the parent that owns their index, matching how NodeBindsLocalSlot
+    // reaches them: DeconstructionTarget through DeconstructionAssignment,
+    // PropertySubpattern through PatternSwitchExpressionArm.Subpattern.
+    static readonly IReadOnlyDictionary<Type, Func<int, IrFunction>> LocalSlotCarrierFactories =
+        new Dictionary<Type, Func<int, IrFunction>>
         {
-            [typeof(NullCoalescingAssignment)] = () => BufferBoundBy(new NullCoalescingAssignment(0, Object, BoxInt(7))),
-            [typeof(ForeachStatement)] = () => BufferBoundBy(new ForeachStatement(0, Object, BoxInt(0), new Block())),
-            [typeof(UsingStatement)] = () => BufferBoundBy(new UsingStatement(0, Object, BoxInt(0), new BlockContainer())),
-            [typeof(Fixed)] = () => BufferBoundBy(new Fixed(Object, 0, BoxInt(0), new BlockContainer())),
-            [typeof(IsPattern)] = () => BufferBoundBy(new IsPattern(BoxInt(0), Object, 0)),
-            [typeof(RecursivePropertyDeclarationPattern)] = () => BufferBoundBy(new RecursivePropertyDeclarationPattern(BoxInt(0), PatternAccessor, Object, 0)),
-            [typeof(UnionSwitchExpressionArm)] = () => BufferBoundBy(new UnionSwitchExpressionArm(Object, localIndex: 0, BoxInt(1))),
-            [typeof(PatternSwitchExpressionArm)] = () => BufferBoundBy(new PatternSwitchExpressionArm(Object, localIndex: 0, subpattern: null, BoxInt(1))),
-            [typeof(PropertySubpattern)] = () => BufferBoundBy(new PatternSwitchExpressionArm(Object, localIndex: null, new PropertySubpattern(PatternAccessor, Object, 0), BoxInt(1))),
-            [typeof(CatchClause)] = () => BufferBoundBy(new CatchClause(Object, new BlockContainer()) { VariableIndex = 0 }),
-            [typeof(DeconstructionAssignment)] = () => BufferBoundBy(new DeconstructionAssignment([0], [Object], BoxInt(0), [true])),
-            [typeof(DeconstructionTarget)] = () => BufferBoundBy(new DeconstructionAssignment([0], [Object], BoxInt(0), [true])),
+            [typeof(NullCoalescingAssignment)] = slot => BufferBoundBy(new NullCoalescingAssignment(slot, Object, BoxInt(7))),
+            [typeof(ForeachStatement)] = slot => BufferBoundBy(new ForeachStatement(slot, Object, BoxInt(0), new Block())),
+            [typeof(UsingStatement)] = slot => BufferBoundBy(new UsingStatement(slot, Object, BoxInt(0), new BlockContainer())),
+            [typeof(Fixed)] = slot => BufferBoundBy(new Fixed(Object, slot, BoxInt(0), new BlockContainer())),
+            [typeof(IsPattern)] = slot => BufferBoundBy(new IsPattern(BoxInt(0), Object, slot)),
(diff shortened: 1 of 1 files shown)


richlander:
CI green (`test (ubuntu-26.04, linux-x64)` passed in 9m3s; path-gated jobs correctly skipped for a test-only change).

Ready to merge.

richlander:
Reviewed at head `05992138` in an isolated worktree. **The change is correct, the teeth are real, and it takes the right option from #3329. Approving.** One residual below that I'd fold into this PR — it's one line and I've verified the fix.

## First: you're right and I was wrong

My #3329 minor finding claimed a carrier hidden behind a throwing property would be "silently not found, and the row would then pass for the wrong reason." That is wrong, and the PR is right to say so. Both call sites are `Assert.True(FunctionContainsCarrier(...))` (`:608`, `:627`), so a missed carrier fails the row loudly. There is no silent-pass path and there never was. I should have checked the callers before writing that sentence.

The narrowing to `catch (TargetInvocationException)` is the better fix for the *actual* defect, which I described imprecisely: the bare `catch` also swallowed faults in the helper's own loop, so an NRE or bad cast would have read as "carrier absent." Reflection wraps a property body's own throw in `TargetInvocationException`, so that clause keeps the intended tolerance and drops the unintended one. Measuring 0 throws across every row before keeping the clause at all is the right order of operations.

## Verified independently

**Your reproduction reproduces.** Deleting the `IsPattern` arm from `NodeBindsLocalSlot` and making the `IsPattern` factory a two-binder with a stray `StoreLocal(0, …)`:

```
MarkLocalEliminated_RefusesEveryLocalSlotBindingNodeKind(carrierKind: "IsPattern") [FAIL]
Total: 38, Failed: 1
```

Exactly one row, the right one. The `factory(1)` leg is a genuinely clever construction — pointing the same factory at a different slot to prove attribution, rather than adding a synthetic IR type, gets the property asserted without growing the fixture surface.

| Check | Result | PR claimed |
| --- | --- | --- |
| `-class InlineArraySpilledElementTests` | **38 / 0** | 38 ✅ |
| `--gate fast` | **3973 / 0** | 3973 ✅ |
| Tamper: arm deleted + stray `StoreLocal(0)` | 1 failure, `IsPattern` row | ✅ |

## Residual: the escape survives if the stray binder tracks the parameter

The comment at `:620` says:

> Any stray slot-0 binder makes this leg refuse and fails the row, which turns "sole slot-0 binder" from a property of how the factories happen to be written into one the Theory asserts.

That holds for a stray binder *pinned* to 0. It does not hold for one that moves with the factory parameter. Same arm deletion, stray binder written as `StoreLocal(slot, …)` instead of `StoreLocal(0, …)`:

```
Total: 38, Errors: 0, Failed: 0
```

Green, with `IsPattern` unhandled in `NodeBindsLocalSlot`. Both legs pass because the stray binder shadows the carrier in leg 0 (both bind slot 0 — so the carrier is *not* the sole slot-0 binder, exactly the case the sentence claims to catch) and vacates slot 0 in leg 1 alongside it.

This is much weaker than #3329 — it needs a deliberately parameterized stray binder, and `BufferBoundBy` taking a single `IrNode` means smuggling a sibling requires wrapping it in a container on purpose. I had to add a helper to express it. But it is the same shape the PR exists to fix: a comment asserting a property one notch stronger than the assertion delivers.

### One-line fix, verified

No factory naturally builds a local-access node, so the fixture can simply forbid them. I added this to the Theory and all 12 rows stayed green:

```csharp
static int CountLocalAccess(IrNode n) =>
    (n is LoadLocal or StoreLocal or LoadLocalAddress ? 1 : 0)
    + n.Children.Sum(CountLocalAccess);
Assert.Equal(0, CountLocalAccess(factory(0)));
```

- On the clean tree: **38 / 0** — every existing factory already satisfies it.
- With the same tamper that slipped through above: **1 failure**, the `IsPattern` row.

It closes the hole at any slot rather than only at 0, which makes it strictly stronger than the leg it supplements, and it fails a smuggled binder at fixture-construction time instead of relying on elimination behavior to surface it.

Non-blocking either way — this PR strictly improves on `main`, and if you'd rather narrow the comment to "any stray binder pinned to slot 0" and take the assertion as a follow-up, that's an honest resolution too. I'd just rather not leave a slightly-overstated claim in the fix for a slightly-overstated claim.

Agreed on single-review tier: test-only, no product code, teeth demonstrated by tamper/revert.

Issue #3333, opened 2026-07-27
Call-graph performance track for the wasm consumer: cut scope cost in time and peak memory


richlander:
## Why

Call graphs are consumed primarily by the wasm website, where caller graphs are
the most expensive and most visible operation. Delivery of the call-graph
feature (#3280, #3292) is not complete without the performance work, so this
issue is the umbrella for it. #3331 is its first increment.

The browser host has a hard memory ceiling and no way to parallelize this scan,
so **reducing total work is the only available lever**.

## Status
> **#3426 changes the priority order.** The wasm engine calls `LibraryBodyIndex` directly and uses
> neither the AssemblyRef prefilter (#3331) nor the TypeRef narrowing (#3409) — both are wired only
> into the CLI inspector. Increments 1 and 2 are landed but do not reach the consumer this issue
> exists to serve. Rehoming scope selection into Analysis converts already-reviewed work into
> consumer-visible wins, and should precede increment 3 and #3425.

| # | Increment | State |
| --- | --- | --- |
| 1 | AssemblyRef prefilter — #3331 | **landed** (PR #3339, `d4002cf1`) |
| 1b | Cache the whole-graph maps — #3342 | **landed** (PR #3370) |
| 2 | TypeRef narrowing of the single-hop `Callers` section | **landed** (PR #3409) |
| 2b | TypeRef narrowing of the transitive closure | **withdrawn** — unsound as a filter, and provably 0% as a seed |
| 3 | Scope-session retention | partially addressed by #3342's release API |
| 4 | Per-hop caller expansion — #3425 | not started; supersedes 2b as the transitive-graph lever |
| 0 | **Deliver the landed work to the wasm engine — #3426** | not started; **do this first** |

Increment **1b was not in the original plan.** Profiling for increment 3 showed every
call-tree request rebuilding the whole-assembly maps it needs, even for a bounded tree
visiting a few dozen nodes — 34x on the repeated-request shape the wasm consumer has, and
lower peak working set than base for the single-assembly case.

It also changed the shape of increment 3. `LibraryBodyIndex` now exposes
`ReleaseScopeGraph()` and `ReleaseCallGraphCaches()`, so a consumer under a hard ceiling
can drop the cross-assembly graph — the only real memory trade — while keeping the cheap
single-assembly maps. `ReleaseScopeGraph()` returns live bytes to exactly the base figure.
Streaming may still be worth doing, but the retention problem is now releasable rather
than structural.

[...]

PR #3339, opened 2026-07-27
Skip caller-scope assemblies that cannot reference the target

### LINKED ISSUE #3331: Caller discovery indexes every scope assembly: add an AssemblyRef prefilter (6.2s -> 45ms on a 182-DLL miss)
## Problem

Caller discovery opens and fully indexes **every** assembly in the caller scope, including assemblies that cannot possibly contain a caller. For a scope of any size the common outcome is "no hit", and we pay the full cost to discover that.

`ApiMemberAnalysisInspection.CallerScopes` (`src/dotnet-inspect/Inspectors/ApiMemberAnalysisInspection.cs:82-109`) loops over `_callerScopeAssemblies` and calls `MethodBodyInspectionSession.Open` unconditionally for each path. That reaches `LibraryBodyIndex.Open` with `LibraryBodyAnalysisFeatures.MethodEvidence`, which decodes every method body in that assembly.

Neither existing lever helps:

- `_bodyScope` (token-scoped decode) is explicitly disabled for this path — `needsWholeAssemblyBody` (line 45) turns it off for `Callers` and `Call Graph`, because reverse/aggregate sections require a full build.
- `includeAllocations` / `includeOpportunities` gate analysis *phases within* an assembly, not *which assemblies* are opened.

There is no assembly-level opt-out anywhere on the path.

## Evidence

Target `ILInspector.Analysis.TypeRef.Definition`, scope = the 182 DLLs of `Microsoft.NETCore.App 9.0.17`. No framework assembly references `ILInspector.Analysis`, so the result is structurally guaranteed to be empty.

```
dotnet-inspect member --library ILInspector.Analysis.dll "TypeRef" -m Definition --all --section "Callers"
dotnet-inspect member --library ILInspector.Analysis.dll "TypeRef" -m Definition --all --section "Callers" --bin <fx>
```

| Run | Time |
| --- | --- |
| no scope | 0.7s |
| `--bin` over 182 framework DLLs | 6.9s |
| **overhead for a guaranteed-empty answer** | **6.2s** |

A probe reading only the AssemblyRef table across the same 182 files (file cache warmed):

```
files            : 182
readable images  : 169
AssemblyRef rows : 947
would scan       : 0
prefilter cost   : 45 ms
```

**6.2s of body decoding replaced by ~45ms of metadata reads — roughly 140x on the miss case**, and the miss case is the common case.

## Proposal

Before opening a `MethodBodyInspectionSession` for a scope path, read that assembly's AssemblyRef table and skip it when it cannot reference the target's declaring assembly. A few lines of SRM (`MetadataReader.AssemblyReferences`); `ILInspector.Metadata/AssemblyReferenceIdentity.cs` already exists as the typed currency.

Layering: Analysis owns the identity predicate (it owns matching); the skip decision is a caller-discovery concern and belongs in `CallerScopes`.

## Correctness constraint — the filter must use the matcher's identity function

This is the part to get right, and a naive name comparison gets it wrong.

`TypeRef.CanonicalAssembly` (`src/ILInspector.Analysis/TypeRef.cs:270`) collapses `System.Private.CoreLib`, `System.Runtime`, `mscorlib`, `netstandard` and `System.Runtime.Extensions` to `corelib`. Matching is on **canonical** assembly identity, not the raw AssemblyRef name.

So a raw string filter regresses real results: a target in `System.Private.CoreLib` with a candidate referencing `System.Runtime` would be skipped, even though the matcher canonicalizes both to `corelib` and **would have found the callers**. Per `AGENTS.md` ("keep failure visible"), a filter that silently drops callers is worse than being slow.

The wider forwarding case is already a documented accepted miss (`LibraryBodyIndex.cs:1814-1818`), so the prefilter does **not** need to solve type forwarding. It needs to be *exactly as permissive as the matcher*. The invariant worth encoding in a test: the prefilter's identity function and `CallerGraphKey`'s are the same function. If they can drift, they will, and the drift is invisible.

Smaller edges:

- Never filter out the primary assembly under inspection.
- Key off assembly identity, not filename.
- Reading AssemblyRef still opens the PE and metadata reader, so the saving is "no body decode", not "no file touch".

## Why not the #3301 metadata-table projection

`MetadataTableProjector` does project AssemblyRef, but it is disqualified as a decision procedure:

- It is **lossy by design**. `int limit = Math.Min(rowCount, Math.Max(0, options.MaxRowsPerTable))` applies to AssemblyRef like every table, and the `Name` cell goes through `StringCell` → `EscapeText(raw, options.MaxStringChars, out truncated)`. A truncated or escaped name silently fails to match, skipping a library that does contain callers. Budgets are correct for an inspection lens and wrong for a filter.
- It is a **layer inversion**: #3301's own contract is that the projection is never a dependency of the typed extractors.

## Consumer context

The primary consumer of call graphs is the wasm website, where caller graphs are the most expensive and most visible operation. In wasm the cost is not only time: opening and indexing every scope assembly is also a memory-footprint problem, and the usual mitigation of parallelising the scan is unavailable. Reducing the amount of work is the lever that applies there.

## Scope

- In scope: assembly-level prefilter for caller scopes; shared identity predicate with a test pinning it to the matcher; a benchmark or timing canary recording the pinned input, command, baseline and result.
- Not in scope: solving type forwarding beyond the existing corelib canonicalization; changing match semantics; caching indexes across invocations.

Related: #3280, #3292, #3320.

richlander:
Closes #3331. Increment 1 of #3333.

## What

Cross-assembly caller discovery opened and fully body-indexed **every** assembly in the `--bin` / `--project` / `--caller-package` scope before asking whether any of them called the target. When the answer is empty — the common case for a framework or dependency member — that pays the entire cost of the scope to learn there is nothing to report.

An assembly can only call the target if the callee's declaring type resolves to the target's assembly, and a callee's declaring assembly is either the candidate itself (a `TypeDef`) or one of its `AssemblyRef`s (a `TypeRef` scope). Reading those two tables rules an assembly out with certainty, and costs nothing next to decoding its method bodies.

## Measured

`member TypeRef.Definition` with the 182 assemblies of `Microsoft.NETCore.App 9.0.17` in scope. Peak working set sampled from the OS counter while the process was alive. **Output is byte-identical in every row.**

| Case | Before | After |
| --- | --- | --- |
| `Callers`, empty answer | 7.3s / 705 MB | **0.9s / 85 MB** |
| `Call Graph`, empty answer | 9.7s / 960 MB | **0.9s / 90 MB** |
| `Callers`, real answer (corelib) | 10.2s / 732 MB | 9.7s / 727 MB |

The last row is the decisive canonicalization check: the target is defined in `System.Private.CoreLib` while its callers reference the `System.Runtime` facade. All 4280 lines of output are unchanged, and the cost is essentially unchanged because every assembly genuinely must be opened. A filter that compared raw assembly names would have reported zero callers here.

Memory matters as much as time: the wasm consumer has no parallelism to trade against, so reducing work is the only lever (#3333).

## Why it cannot narrow discovery

`MemberPattern.MatchesCrossAssembly` compares the candidate callee's declaring `TypeRef` to the target's, and `TypeRef.Equals` includes `Assembly`, canonicalized on construction. `CallerScopeFilter` applies that same identity test over names, and:

- canonicalizes through `TypeRef.CanonicalAssembly`, so the corelib facades collapse together;
- returns `true` for any undecidable input (unknown target, unreadable image, missing reference table), so it is never more selective than the matcher;
- does not model wider type forwarding, which the matcher already misses, so it stays exactly as permissive as the behavior it guards.

I confirmed the matcher really is assembly-sensitive before relying on it, rather than assuming: a three-assembly fixture where two assemblies both define `N.T.M` and the caller references only one of them reports the caller for the referenced target and not for the other.

## Behavior hazard fixed along the way

`BuildCallerTree` chose between two *different* reverse-graph builders — one keyed by assembly-local tokens, one by structural member identity — based on whether the scope list was non-empty. Those builders do not produce identical trees. Prefiltering can empty the list, so without this fix the optimization would have silently changed the answer instead of just costing less. I caught this because the `Call Graph` equivalence check came back **different**, and chased it rather than accepting it as benign.

The scope list is now nullable and carries intent instead of a survivor count:

- `null` — no cross-assembly scope was requested; use the token-keyed builder.
- empty but non-null — a scope was requested and nothing survived filtering; still take the structural walk.

This preserves current behavior on both sides rather than picking a winner. That the two builders disagree at all is a pre-existing inconsistency this PR only exposes; I have not tried to reconcile them here.

## Evidence

- `CallerScopeFilterTests` (12 cases) pins the boundary in both directions against the **same real fixture assemblies the matcher tests use**, including the lookalike caller that declares its own `Target.Api.Ping` and never references the target — the filter must reach the same answer the matcher already does. Plus the four corelib facade names, unrelated-references, and every fail-open path.
- `BuildCallerTree_PrefilteringScopeAssembliesNeverChangesTheTree` asserts that opening a non-contributing assembly and prefiltering it away are indistinguishable. It is deliberately self-locating rather than pinned to one method: it also requires that at least one visited method distinguishes the two builders, so it cannot quietly become vacuous if this assembly's call structure changes.
- **Mutation-tested**: reverting the `is null` guard to `is not { Count: > 0 }` fails that test with `Expected: "1|<Create>b__17_4|Expanded|"` / `Actual: "1|Create|Expanded|"`. My first attempt at this test passed under the mutation, so I replaced it rather than shipping a guard that proved nothing.
- `BuildCallerTree_SkippingANonContributingScopeAssemblyDoesNotChangeTheTree` covers the same claim on the cross-assembly fixtures, including the all-filtered-out case.
- Suites at baseline: Analysis 537/1, CLI 2216/1 + 10 skipped (both failures pre-existing and unrelated).

## Non-goals

Two further increments of #3333 remain: a `TypeRef`/`MemberRef` second-level filter for the *hit* case (the third row above is untouched by this PR), and peak-memory retention work.

### DIFF
--- a/dotnet-inspect.slnx
+++ b/dotnet-inspect.slnx
@@ -16,6 +16,7 @@
     <Project Path="src/ILInspector.Analysis.App/ILInspector.Analysis.App.csproj" />
     <Project Path="src/ILInspector.Analysis.CallerGraphCaller/ILInspector.Analysis.CallerGraphCaller.csproj" />
     <Project Path="src/ILInspector.Analysis.CallerGraphCallerTwin/ILInspector.Analysis.CallerGraphCallerTwin.csproj" />
+    <Project Path="src/ILInspector.Analysis.CallerGraphIndirectCaller/ILInspector.Analysis.CallerGraphIndirectCaller.csproj" />
     <Project Path="src/ILInspector.Analysis.CallerGraphLookalikeCaller/ILInspector.Analysis.CallerGraphLookalikeCaller.csproj" />
     <Project Path="src/ILInspector.Analysis.CallerGraphTarget/ILInspector.Analysis.CallerGraphTarget.csproj" />
     <Project Path="src/ILInspector.Analysis.CrossAsmCollisionFixtures/ILInspector.Analysis.CrossAsmCollisionFixtures.csproj" />
--- a/src/ILInspector.Analysis.CallerGraphIndirectCaller/ILInspector.Analysis.CallerGraphIndirectCaller.csproj
+++ b/src/ILInspector.Analysis.CallerGraphIndirectCaller/ILInspector.Analysis.CallerGraphIndirectCaller.csproj
@@ -1,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <!--
+    Caller-graph transitive-scope fixture (#3331): an indirect caller. References only the
+    caller assembly, never the target, so a caller graph rooted at Target.Api.Ping reaches
+    it at depth 2 through Shared.Entry.Run. Scope prefiltering must keep it.
+  -->
+  <PropertyGroup>
+    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
+    <LangVersion>preview</LangVersion>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
+    <IsAotCompatible>false</IsAotCompatible>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\ILInspector.Analysis.CallerGraphCaller\ILInspector.Analysis.CallerGraphCaller.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/src/ILInspector.Analysis.CallerGraphIndirectCaller/IndirectEntry.cs
+++ b/src/ILInspector.Analysis.CallerGraphIndirectCaller/IndirectEntry.cs
@@ -1,0 +1,14 @@
+// Caller-graph transitive-scope fixture (#3331). This assembly references ONLY the caller
+// assembly, never the target, yet it belongs in a caller graph rooted at Target.Api.Ping:
+// Indirect.Entry.Run -> Shared.Entry.Run -> Target.Api.Ping is two hops.
+//
+// It exists to pin the transitive obligation of the AssemblyRef prefilter. A prefilter that
+// asks only "does this assembly reference the target?" drops this one and silently shortens
+// the graph, so the reference set here must stay free of the target assembly.
+namespace Indirect
+{
+    public static class Entry
+    {
+        public static void Run() => Shared.Entry.Run();
+    }
+}
--- a/src/ILInspector.Analysis.Tests/CallerScopeFilterTests.cs
+++ b/src/ILInspector.Analysis.Tests/CallerScopeFilterTests.cs
@@ -309,12 +309,24 @@ public class CallerScopeFilterTests
     // Copies of one assembly under several subdirectories, or several facades canonicalizing
     // together, all share a canonical name. Every copy must be selected — and the walk must stay
     // linear while doing it. Before the adjacency entry was consumed on first visit, each selected
-    // sharer re-enqueued the shared name and re-traversed the whole group: quadratic in group size,
-    // measured at ~10s for 40k sharers against ~13ms after.
+    // sharer re-enqueued the shared name and re-traversed the whole group: quadratic in group size.
+    //
+    // The group size is chosen for separation under mutation, not for realism. Round 6 measured the
+    // earlier 40k version at ~2080ms against this 2000ms bound — a 4% margin, which faster hardware
+    // would erase, letting the quadratic pass. Because the mutant grows quadratically and the fixed
+    // path linearly, raising the group widens both margins at once (3x the group cost 8.9x the
+    // mutant's time, and almost nothing here):
+    //
+    //   sharers    fixed     reverted to a non-consuming lookup
+    //   40,000     ~13ms     ~2,080ms
+    //   120,000    ~40ms     ~18,509ms
+    //
+    // So the bound now sits ~50x above the linear cost and ~9x below the quadratic one, which
+    // discriminates the two shapes rather than measuring this machine.
     [Fact]
     public void SameNamedCandidatesAreAllSelectedInLinearTime()
     {
-        const int sharers = 40_000;
+        const int sharers = 120_000;
         var candidates = new List<CallerScopeFilter.Candidate>(sharers);
         for (int i = 0; i < sharers; i++)
             candidates.Add(CallerScopeFilter.Candidate.Known("Dup", ["Target"]));
@@ -325,8 +337,8 @@ public class CallerScopeFilterTests
 
         Assert.All(selected, Assert.True);
 
-        // Three orders of magnitude of headroom over the linear cost, and two under the quadratic
-        // one, so this discriminates the shapes without being a wall-clock measurement.
+        // See the margin table above: ~50x of headroom over the linear cost and ~9x under the
+        // quadratic one, so this discriminates the shapes without being a wall-clock measurement.
         Assert.True(
             watch.ElapsedMilliseconds < 2_000,
             $"selection over {sharers} same-named candidates took {watch.ElapsedMilliseconds}ms, "
--- a/src/ILInspector.Analysis.Tests/CallerScopeFilterTests.cs
+++ b/src/ILInspector.Analysis.Tests/CallerScopeFilterTests.cs
@@ -306,6 +306,44 @@ public class CallerScopeFilterTests
             CallerScopeFilter.SelectCouldReach(target.Name, [caller, indirect, lookalike]));
     }
 
+    // Copies of one assembly under several subdirectories, or several facades canonicalizing
+    // together, all share a canonical name. Every copy must be selected — and the walk must stay
+    // linear while doing it. Before the adjacency entry was consumed on first visit, each selected
+    // sharer re-enqueued the shared name and re-traversed the whole group: quadratic in group size,
+    // measured at ~10s for 40k sharers against ~13ms after.
+    [Fact]
+    public void SameNamedCandidatesAreAllSelectedInLinearTime()
+    {
+        const int sharers = 40_000;
+        var candidates = new List<CallerScopeFilter.Candidate>(sharers);
+        for (int i = 0; i < sharers; i++)
+            candidates.Add(CallerScopeFilter.Candidate.Known("Dup", ["Target"]));
+
+        var watch = System.Diagnostics.Stopwatch.StartNew();
+        var selected = CallerScopeFilter.SelectCouldReach("Target", candidates);
+        watch.Stop();
+
+        Assert.All(selected, Assert.True);
+
+        // Three orders of magnitude of headroom over the linear cost, and two under the quadratic
+        // one, so this discriminates the shapes without being a wall-clock measurement.
+        Assert.True(
+            watch.ElapsedMilliseconds < 2_000,
+            $"selection over {sharers} same-named candidates took {watch.ElapsedMilliseconds}ms, "
+                + "which indicates the walk revisits an adjacency entry per sharer");
+    }
+
+    // A candidate whose own name is also one of its references, and a candidate naming the same
+    // reference twice, are both indexed once per name. Nothing above depends on the count, so this
+    // pins the dedupe directly rather than through an observable it cannot reach.
+    [Fact]
+    public void RepeatedNamesIndexACandidateOnce()
+    {
+        Assert.True(Selects("Target", "Target", ["Target", "Target"]));
+        Assert.True(Selects("Target", "Self", ["Self", "Target", "Target"]));
+        Assert.False(Selects("Target", "Self", ["Self", "Self"]));
+    }
+
     [Fact]
     public void EmptyCandidateListSelectsNothing()
     {
--- a/src/ILInspector.Analysis.Tests/CallerScopeFilterTests.cs
+++ b/src/ILInspector.Analysis.Tests/CallerScopeFilterTests.cs
@@ -1,0 +1,102 @@
+using System.Collections.Immutable;
+using System.Reflection.PortableExecutable;
+using DotnetInspector.Fixtures;
+using ILInspector.Analysis;
+using ILInspector.Metadata;
+
+namespace ILInspector.Analysis.Tests;
+
+/// <summary>
+/// <see cref="CallerScopeFilter"/> decides, from assembly identity alone, whether an assembly is
+/// worth opening during cross-assembly caller discovery. Its only correctness obligation is that it
+/// never rules out an assembly the matcher would have matched, so these tests pin the boundary in
+/// both directions and use the same real fixture assemblies the matcher tests use.
+/// </summary>
+public class CallerScopeFilterTests
+{
+    static AssemblyIdentityNames Identity(string assemblyPath)
+    {
+        using var stream = File.OpenRead(assemblyPath);
+        using var peReader = new PEReader(stream);
+        return AssemblyIdentityScanner.Scan(peReader);
+    }
+
+    static bool CouldContainCallerOf(string targetAssemblyPath, string candidateAssemblyPath)
+    {
+        var candidate = Identity(candidateAssemblyPath);
+        return CallerScopeFilter.CouldContainCallerOf(
+            Identity(targetAssemblyPath).Name, candidate.Name, candidate.ReferenceNames);
+    }
+
+    // The real caller fixture references the target, so it must survive the filter and be opened.
+    // This is the assembly BuildCallerTree_WithScope_* proves does contribute callers.
+    [Fact]
+    public void ReferencingAssemblyIsKept()
+    {
+        Assert.True(CouldContainCallerOf(
+            FixtureCatalog.AnalysisCallerGraphTarget.AssemblyPath(),
+            FixtureCatalog.AnalysisCallerGraphCaller.AssemblyPath()));
+    }
+
+    // The lookalike declares its own Target.Api.Ping and never references the target assembly.
+    // The matcher already excludes it (BuildCallerTree ignores its calls); the filter must reach
+    // the same answer, which is what makes skipping it an optimization rather than a behavior change.
+    [Fact]
+    public void NonReferencingAssemblyIsRuledOut()
+    {
+        Assert.False(CouldContainCallerOf(
+            FixtureCatalog.AnalysisCallerGraphTarget.AssemblyPath(),
+            FixtureCatalog.AnalysisCallerGraphLookalikeCaller.AssemblyPath()));
+    }
+
+    // The candidate is the target: its own TypeDefs are callees the matcher can match.
+    [Fact]
+    public void TargetAssemblyItselfIsKept()
+    {
+        string target = FixtureCatalog.AnalysisCallerGraphTarget.AssemblyPath();
+        Assert.True(CouldContainCallerOf(target, target));
+    }
+
+    // The facade case, and the reason the filter canonicalizes instead of comparing raw names.
+    // Callers reference System.Runtime; the target is defined in System.Private.CoreLib. A raw name
+    // comparison would rule out essentially every caller of a corelib member.
+    [Theory]
+    [InlineData("System.Runtime")]
+    [InlineData("mscorlib")]
+    [InlineData("netstandard")]
+    [InlineData("System.Runtime.Extensions")]
+    public void CorelibFacadeReferenceIsKept(string facade)
+    {
+        Assert.True(CallerScopeFilter.CouldContainCallerOf(
+            "System.Private.CoreLib", "Some.Consumer", [facade]));
+    }
+
+    [Fact]
+    public void UnrelatedReferencesAreRuledOut()
+    {
+        Assert.False(CallerScopeFilter.CouldContainCallerOf(
+            "Target", "Some.Consumer", ["System.Runtime", "Newtonsoft.Json"]));
+    }
+
+    // Undecidable inputs must fail open: a filter that cannot tell has to let the real matcher decide.
+    [Theory]
+    [InlineData(null)]
+    [InlineData("")]
+    public void UnknownTargetFailsOpen(string? target)
+    {
+        Assert.True(CallerScopeFilter.CouldContainCallerOf(target, "Some.Consumer", ["Unrelated"]));
+    }
+
+    [Fact]
+    public void UnknownCandidateReferencesFailOpen()
+    {
+        Assert.True(CallerScopeFilter.CouldContainCallerOf("Target", "Some.Consumer", null));
+    }
+
+    // A reference-free image (no AssemblyRef rows) that is not the target cannot call into it.
+    [Fact]
+    public void EmptyReferenceSetIsRuledOut()
+    {
+        Assert.False(CallerScopeFilter.CouldContainCallerOf("Target", "Some.Consumer", ImmutableArray<string>.Empty));
+    }
+}
--- a/src/ILInspector.Analysis.Tests/CallerScopeFilterTests.cs
+++ b/src/ILInspector.Analysis.Tests/CallerScopeFilterTests.cs
@@ -11,6 +11,9 @@ namespace ILInspector.Analysis.Tests;
 /// worth opening during cross-assembly caller discovery. Its only correctness obligation is that it
 /// never rules out an assembly the matcher would have matched, so these tests pin the boundary in
 /// both directions and use the same real fixture assemblies the matcher tests use.
+///
+/// The obligation is over the whole caller <em>graph</em>, not one level of it, so the transitive
+/// cases below are the ones that distinguish the closure from a direct-reference test.
 /// </summary>
 public class CallerScopeFilterTests
 {
@@ -21,19 +24,28 @@ public class CallerScopeFilterTests
         return AssemblyIdentityScanner.Scan(peReader);
     }
 
-    static bool CouldContainCallerOf(string targetAssemblyPath, string candidateAssemblyPath)
+    static CallerScopeFilter.Candidate Candidate(string assemblyPath)
     {
-        var candidate = Identity(candidateAssemblyPath);
-        return CallerScopeFilter.CouldContainCallerOf(
-            Identity(targetAssemblyPath).Name, candidate.Name, candidate.ReferenceNames);
+        var identity = Identity(assemblyPath);
+        return new(identity.Name, identity.ReferenceNames);
     }
+
+    /// <summary>Whether a single candidate survives selection for the given target.</summary>
+    static bool Selects(string? targetAssembly, CallerScopeFilter.Candidate candidate) =>
+        CallerScopeFilter.SelectCouldReach(targetAssembly, [candidate])[0];
+
+    static bool Selects(string? targetAssembly, string? name, IReadOnlyList<string>? references) =>
+        Selects(targetAssembly, new CallerScopeFilter.Candidate(name, references));
+
+    static bool SelectsFile(string targetAssemblyPath, string candidateAssemblyPath) =>
+        Selects(Identity(targetAssemblyPath).Name, Candidate(candidateAssemblyPath));
 
     // The real caller fixture references the target, so it must survive the filter and be opened.
     // This is the assembly BuildCallerTree_WithScope_* proves does contribute callers.
     [Fact]
     public void ReferencingAssemblyIsKept()
     {
-        Assert.True(CouldContainCallerOf(
+        Assert.True(SelectsFile(
             FixtureCatalog.AnalysisCallerGraphTarget.AssemblyPath(),
             FixtureCatalog.AnalysisCallerGraphCaller.AssemblyPath()));
     }
@@ -44,7 +56,7 @@ public class CallerScopeFilterTests
     [Fact]
     public void NonReferencingAssemblyIsRuledOut()
     {
-        Assert.False(CouldContainCallerOf(
+        Assert.False(SelectsFile(
             FixtureCatalog.AnalysisCallerGraphTarget.AssemblyPath(),
             FixtureCatalog.AnalysisCallerGraphLookalikeCaller.AssemblyPath()));
     }
@@ -54,7 +66,7 @@ public class CallerScopeFilterTests
     public void TargetAssemblyItselfIsKept()
     {
         string target = FixtureCatalog.AnalysisCallerGraphTarget.AssemblyPath();
-        Assert.True(CouldContainCallerOf(target, target));
+        Assert.True(SelectsFile(target, target));
     }
 
     // The facade case, and the reason the filter canonicalizes instead of comparing raw names.
@@ -67,15 +79,13 @@ public class CallerScopeFilterTests
     [InlineData("System.Runtime.Extensions")]
     public void CorelibFacadeReferenceIsKept(string facade)
     {
-        Assert.True(CallerScopeFilter.CouldContainCallerOf(
-            "System.Private.CoreLib", "Some.Consumer", [facade]));
+        Assert.True(Selects("System.Private.CoreLib", "Some.Consumer", [facade]));
     }
 
     [Fact]
     public void UnrelatedReferencesAreRuledOut()
     {
-        Assert.False(CallerScopeFilter.CouldContainCallerOf(
-            "Target", "Some.Consumer", ["System.Runtime", "Newtonsoft.Json"]));
+        Assert.False(Selects("Target", "Some.Consumer", ["System.Runtime", "Newtonsoft.Json"]));
     }
 
     // Undecidable inputs must fail open: a filter that cannot tell has to let the real matcher decide.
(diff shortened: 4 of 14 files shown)


richlander:
## Round 1 reconciliation

Both reviewers ran against `ea900512` in isolated worktrees.

**Gemini 3.1 Pro — clean.** Executed the attack points including independently reproducing the mutation test and the corelib facade run (4277 lines).

**GPT-5.6 — found a real regression I had missed, and it was blocking.**

`ApiOptions.CallerScopeAssemblies` is declared non-nullable with an empty default:

```csharp
public IReadOnlyList<string> CallerScopeAssemblies { get; init; } = [];
```

and `ApiCommand` passes it unconditionally (`ApiCommand.cs:738`, `:1430`). So my `_callerScopeAssemblies is null` guard was **dead code**, and every ordinary unscoped request — the overwhelmingly common case — was silently routed to the cross-assembly builder. The two reverse-graph builders do not produce the same tree, so the default `Call Graph` view changed shape.

GPT graded it Medium. I reproduced it before acting on it and **upgraded it to blocking**: `member <asm> <member> -S "Call Graph"` with no scope flag went from **63 to 61 lines**.

**Why my own validation missed it:** every row of my equivalence matrix used `--bin`. I never tested the absence of the flag I was optimizing. An equivalence matrix has to include the default path.

### Fix — `f37aafe8`

The predicate now reproduces pre-prefilter behavior exactly. Before, the cross-assembly builder ran iff at least one scope assembly *opened successfully*, so:

| scope state | builder |
| --- | --- |
| nothing supplied (null **or** empty) | same-assembly |
| every entry failed to open | same-assembly |
| at least one opened **or** prefiltered | cross-assembly |

That last row is why skips are counted: a skipped assembly is one the unfiltered walk would have opened successfully and found nothing in, so skipping it must not change which builder answers. For the same reason a file with no managed metadata now **fails open** instead of counting as a skip — `--bin` enumerates every top-level `*.dll` with no managed-image filter, and those were never openable in the first place. That also resolves the fail-open concern both reviewers raised in round 1.

### Tests — `ApiMemberAnalysisInspectionTests`

I first wrote an output-level golden test on the call-graph fixture. **It proved nothing**: the small fixtures happen not to distinguish the two builders at all, so it passed under the bug. The tests assert the routing decision directly instead.

Mutation-tested, all three discriminators:

| mutation | result |
| --- | --- |
| drop the `opened == 0 && skipped == 0` guard | **FAILS** `CallerScopes_WhenNoScopeAssemblyCanBeOpened_...` |
| `!HasMetadata` counts as a skip instead of failing open | **FAILS** `CallerScopes_WhenTheOnlyScopeEntryIsNotManaged_...` |
| revert the emptiness check to a null check | survives — equivalent mutant, subsumed by the guard above |

### Full base-vs-head matrix, now including the unscoped rows

| Case | base | head | |
| --- | --- | --- | --- |
| unscoped `Call Graph` | 0.9s / 88 MB / 31 ln | 1.1s / 86 MB / 31 ln | IDENTICAL |
| unscoped `Callers` | 0.9s / 83 MB / 8 ln | 0.7s / 79 MB / 8 ln | IDENTICAL |
| 182-DLL `Call Graph` miss | 11.1s / 958 MB / 35 ln | **1.0s / 92 MB** / 35 ln | IDENTICAL |
| 182-DLL `Callers` miss | 7.0s / 703 MB / 8 ln | **0.7s / 88 MB** / 8 ln | IDENTICAL |
| 182-DLL `Callers` hit (corelib facade) | 8.6s / 723 MB / 4281 ln | 8.5s / 723 MB / 4281 ln | IDENTICAL |
| native-only scope (fail-open) | 0.8s / 86 MB / 35 ln | 0.9s / 87 MB / 35 ln | IDENTICAL |
| ambiguity negative (TargetLib) | 7 ln | 7 ln | IDENTICAL |
| ambiguity positive (OtherLib) | 9 ln | 9 ln | IDENTICAL |

Suites at baseline: Analysis 537 (1 pre-existing failure), CLI 2223 (1 pre-existing failure, 10 skipped) — 2216 + the 7 new tests.

Round 2 review is running against the fixed head `f37aafe8`. Not ready to merge until it reports clean.

Thanks @gpt reviewer for the catch — it was the difference between a perf win and a silent default-output regression.

richlander:
## Round 2 reconciliation — both reviewers found the same real defect

Reviews ran against `f37aafe8` in isolated worktrees. **They converged independently on one finding**, which is worth taking seriously on its own.

| Reviewer | Grade | Finding |
| --- | --- | --- |
| GPT-5.6 | Important | prefilter skips managed images the body index cannot open |
| Gemini 3.1 Pro | Minor | same — valid metadata, corrupt bodies |

Both reasoned it; **neither reproduced it. I did.**

Single-byte mutations of a real assembly: **8 of 3000 trials** produced an image where `AssemblyInspectionSession.IdentityNames()` succeeds but `MethodBodyInspectionSession.Open` throws:

```text
offset 0x412 -> BadImageFormatException: Invalid coded index.
offset 0x30D -> BadImageFormatException: Read out of bounds.
offset 0x2B9 -> BadImageFormatException: Read out of bounds.
...
```

So the premise my previous commit rested on — *"a skipped assembly is one the unfiltered walk would have opened successfully"* — is **false on real input**. Skip counting was an approximation of an accident.

### Fix — `907556c1`

Whether a skipped assembly *would* have opened is not knowable without paying exactly the cost the prefilter exists to avoid. So it stops guessing. Builder selection is now a question about the **request** — did the user ask for a cross-assembly scope? — and no longer depends on how much of that scope turned out to be readable. This is GPT's second suggested fix, and it removes the undecidable premise rather than approximating it.

It is better behavior independently: before, the shape of the tree depended on whether a directory happened to contain openable DLLs, so **adding one readable file to a `--bin` directory could change the answer to a question about a completely different assembly.**

### Behavior change, stated plainly

When a caller scope is supplied but **nothing in it can be opened** (an all-native or all-corrupt `--bin` directory), the scoped builder now answers instead of silently falling back to the unscoped one. Both still report no callers; the tree ordering differs.

Measured, not asserted — all-native scope over `ILInspector.Analysis`: **8 of 37** `LibraryBodyIndex` members render differently. Every other case is unchanged.

The hot path is unaffected and is in fact **more faithful** than skip counting was: with 182 framework DLLs all ruled out by the prefilter, the old unfiltered walk opened them all and used the scoped builder — and so does this.

### Full 8-case base-vs-head matrix — all byte-identical

| Case | base | head | |
| --- | --- | --- | --- |
| unscoped `Call Graph` | 0.7s / 86 MB | 0.8s / 87 MB | IDENTICAL |
| unscoped `Callers` | 0.6s / 81 MB | 0.6s / 83 MB | IDENTICAL |
| 182-DLL `Call Graph` miss | 10.1s / 958 MB | **0.9s / 91 MB** | IDENTICAL |
| 182-DLL `Callers` miss | 6.6s / 702 MB | **0.7s / 88 MB** | IDENTICAL |
| corelib facade hit | 8.0s / 725 MB | 8.2s / 722 MB | IDENTICAL |
| native-only scope | 0.7s / 87 MB | 0.8s / 92 MB | IDENTICAL |
| ambiguity negative | 8 ln | 8 ln | IDENTICAL |
| ambiguity positive | 10 ln | 10 ln | IDENTICAL |

Suites at baseline: Analysis **537/1**, CLI **2223/1 + 10 skipped**.

### Also cleared in round 2

- **Gemini** attacked the prefilter predicate itself — type forwarding, `ExportedType`, `ModuleRef`, netmodule `TypeRef` scopes, cross-assembly generics, `InternalsVisibleTo`, and canonicalization symmetry. **No false negatives found**; every cross-assembly `DirectCall` a compiler emits implies an `AssemblyRef` row.
- Both confirmed disposal, file-handle, and `ref`-local/resolved-flag handling are sound, and that the new tests exercise the skip path and a genuine metadata-free PE rather than silently skipping.

Round 3 review is queued against `907556c1`. Not ready to merge until it reports clean.

Thanks to both reviewers — the value here was two independent models landing on the same premise and forcing me to go test it instead of defending it.

richlander:
## Round 3 reconciliation — `907556c1` → `e91529c0`

Both reviewers were given the same prompt at the same fixed head. **Both independently found the same defect, and it was real.**

### Blocking: the prefilter was not transitive (GPT-5.6 — Blocking; Gemini 3.1 Pro — Blocking)

`BuildCallerTree` walks outward to `maxDepth: 3`. The prefilter asked only *"does this candidate reference the target?"*. That is the right question for one level and the wrong one for a graph: an assembly that references only an intermediate still belongs in the tree, and was being dropped.

I reproduced it with a purpose-built chain `ChainEntry → ChainMiddle → ChainTarget`, where `ChainEntry` references only `ChainMiddle`:

```text
base:  ├─ M.Middle.Run() (fanin 1, depth 2, from ChainMiddle)
       └─ E.Entry.Run() (from ChainEntry)
          └─ ↩ M.Middle.Run() (fanin 1, depth 2, from ChainMiddle)

head:  (E.Entry.Run() absent; graph reported depth 2)
```

**This was a wrong answer, not a slower one**, and it is the most serious class of defect this optimization can produce. Gemini reached it via a `LibC → LibB → LibA` chain and GPT via an equivalent shape; neither of the two earlier rounds had attacked depth.

**Fix (`e91529c0`):** selection is now the **reverse-reference closure** over the supplied scope — seed with the target, admit anything naming something already reached, add its own name, repeat to a fixpoint. `CallerScopes` became two-phase: scan every candidate's identity, compute the closure, open only survivors.

Two properties worth stating because they are the boundary:

- It closes over the **scope**, not the machine. An assembly whose only bridge to the target was not itself supplied stays out — the walk could not have traversed that bridge either.
- An unreadable candidate is kept but does **not** widen the closure. An unknown name must not admit its referrers, or a single unreadable file would pull in everything.

`CouldContainCallerOf` (the one-level test) is deleted rather than retained. Both sections route through the same `CallerScopes`, so keeping a cheaper one-level path for `Callers` would have been dead code.

### Important: builder divergence is wider than I measured (Gemini 3.1 Pro)

Gemini is right that my characterization was under-scoped. I reported "8 of 37 `LibraryBodyIndex` members diverge"; Gemini showed `CallTreeNode` diverges on 6 of 13 members and `LibraryBodyIndex.DirectCalls` yields 71 vs 75 nodes. **The divergence is a general property of the two builders, not a property of one type.** I have restated it that way and filed #3351.

I have **not** taken the suggested fix of routing an empty survivor list back to the same-assembly builder. That reintroduces exactly the round-1 regression: with 182 DLLs in scope the unfiltered walk opened them and used the structural builder, so a prefiltered walk that empties the list must still use the structural builder or the hot path stops matching base. The underlying disagreement between the two builders is pre-existing and is #3351's to settle.

### Attack points that came back clean

Caching and disposal (both reviewers, code reading), test pinning (Gemini executed the suite and confirmed the native-image test finds a real image rather than skipping), and the scope-state enumeration all reported no issues.

### Evidence at `e91529c0`

Chain reproduction now byte-identical to base, with `E.Entry.Run()` restored at depth 3.

| Case | base | head |
| --- | --- | --- |
| unscoped `Call Graph` | 1.0s / 86 MB | 0.9s / 87 MB |
| unscoped `Callers` | 0.8s / 81 MB | 0.8s / 81 MB |
| 182-DLL `Call Graph` miss | 9.6s / 959 MB | **1.0s / 90 MB** |
| 182-DLL `Callers` miss | 6.8s / 699 MB | **0.8s / 85 MB** |
| corelib facade hit | 9.8s / 730 MB | 10.1s / 725 MB |
| ambiguity negative / positive | 7 ln / 9 ln | 7 ln / 9 ln |
| native-only scope | 72 ln | 70 ln (declared, #3351) |

Every row byte-identical except the already-declared one. **The closure costs nothing on the miss case** — for an unrelated target, no framework assembly enters the closure at any level.

New fixture `ILInspector.Analysis.CallerGraphIndirectCaller` references only the caller assembly, never the target, so the transitive obligation is pinned on a real artifact end to end. Both new tests assert the absent direct reference, so the fixture cannot quietly stop proving anything.

Mutation-tested: reverting selection to a direct-reference test fails five filter tests **and** the end-to-end `CallerScopes` test.

Suites at baseline — Analysis **548 / 1 pre-existing**, CLI **2225 / 1 pre-existing + 10 skipped**.

Round 4 is running against `e91529c0`. Not ready to merge until it comes back clean.

richlander:
## Round 4 reconciliation — `e91529c0` → `ef549e8c`

Both reviewers ran at the same fixed head. **Both found real defects, and one is blocking.** Four rounds, four genuine defects.

### Blocking: an unreadable candidate must widen the closure (GPT-5.6)

I had written "an unknown name must not widen the closure" and defended it in the round-3 reconciliation. **It was wrong, and the reasoning error is worth naming:** I treated selection and reachability as separate questions. They are the same question. Anything selected gets **opened**, so it can contribute edges, so whatever references *it* can be a caller-of-a-caller. Refusing to widen cut those assemblies off.

GPT reproduced it on `Outer → Entry → Middle → Target` with a single-byte mutation of `Middle.dll` (offset 1075) that breaks identity reading while body indexing still succeeds:

```text
BASE: Target -> Middle -> Entry -> Outer   (depth 4)
HEAD: Target -> Middle                     (depth 2)
```

GPT also caught that `AssemblyIdentityScanner` discarded a successfully-read assembly name whenever any `AssemblyRef` row failed — throwing away the one fact that would have kept the chain intact.

**Fix:** identity now has four states instead of one nullable pair, because they have genuinely different consequences:

| State | Selected? | Widens? | Why |
| --- | --- | --- | --- |
| `Known` | by closure | yes | fully decidable |
| `UnknownReferences` | always | **yes**, under its own name | a row that failed to read could have been the target |
| `Unknown` | selects **whole scope** | — | an unnameable assembly could sit anywhere in the chain |
| `Unopenable` | **no** | no | could not have contributed edges under the unfiltered walk either |

`Unopenable` is what keeps this an optimization at all — otherwise a single native DLL in a `--bin` directory would select everything. `AssemblyIdentityScanner` now reports a partial result rather than discarding the name.

**Verified:** GPT's exact command is byte-identical to base again, all four levels restored.

### Important: the fixpoint was cubic — the "optimization" was 7.7× *slower* than base

Gemini flagged the algorithm (iterating the `HashSet` linearly and re-canonicalizing on every pass); GPT measured the consequence end-to-end on a 1200-assembly chain scope:

```text
base: 946 ms      head: 7261 ms
```

**On a performance PR, that is the most important possible finding**, and neither of us had constructed a scope shaped to expose it. My own 2000×100 probe confirmed it: the reverse-ordered chain did not finish in five minutes.

**Fix:** replaced the fixpoint with a reverse adjacency index walked by a queue — canonicalize once, index each candidate under every name it mentions, walk outward from the target. Linear in candidates plus reference edges.

| Shape | before | after |
| --- | --- | --- |
| reverse chain 1200×1 (GPT's case) | 6027 ms | **0.52 ms** |
| reverse chain 2000×100 | did not finish in 5 min | **77 ms** |
| 182 × 60 refs, realistic miss | — | **1.04 ms** |

GPT's end-to-end case goes 7261 ms → **1472 ms**, against a base of 1080 ms.

**The residual 36% is real and I am not going to hide it.** When the filter can rule nothing out, it still pays one cheap metadata open per scope assembly on top of the body decode. Those 1200 assemblies are 4 KB stubs, so the decode it doubles is nearly free — that shape is close to the worst case for this optimization. At realistic assembly sizes the decode dominates: the 182-DLL corelib hit is 9.3s base vs 9.5s head, ~2%.

### Matrix at `ef549e8c` — win unchanged

| Case | base | head |
| --- | --- | --- |
| unscoped `Call Graph` | 0.8s / 87 MB | 0.7s / 87 MB |
| unscoped `Callers` | 0.6s / 83 MB | 0.6s / 83 MB |
| 182-DLL `Call Graph` miss | 8.5s / 957 MB | **0.9s / 91 MB** |
| 182-DLL `Callers` miss | 6.5s / 703 MB | **0.7s / 82 MB** |
| corelib facade hit | 9.3s / 725 MB | 9.5s / 728 MB |
| ambiguity negative / positive | 7 ln / 9 ln | 7 ln / 9 ln |

All byte-identical except the declared native-only-scope row (#3351). The round-3 chain fixture is byte-identical too.

Selection is now resolved once and shared by both lenses, so a request rendering both `Callers` and `Call Graph` scans identities once rather than twice.

### Also verified clean by both reviewers

Canonicalization at depth, corelib hit-case equivalence (byte-identical, SHA256 matched), fail-open on directories / zero-byte files / text files / native PEs, cycle termination, cross-lens caching and handle retention, and the intent-based builder characterization.

Mutation-tested: making `UnknownReferences` not widen fails the widening test; making `Unknown` not select the scope fails the whole-scope test.

Suites at baseline — Analysis **551 / 1 pre-existing**, CLI **2225 / 1 pre-existing + 10 skipped**.

Round 5 is running against `ef549e8c`. Not ready to merge until it comes back clean.

richlander:
## Round 5 reconciliation — `ef549e8c` → `9a5bf651`

Both reviewers ran at the same fixed head. **Both found real defects, and one is blocking.** Five rounds, five genuine defects.

### Blocking: one junk DLL disabled the optimization entirely (GPT-5.6)

`--bin` enumerates every top-level `*.dll` with no managed-image filter, so a scope routinely contains files that are not assemblies. A zero-byte or truncated one was classified `Unknown`, and `Unknown` selects the **whole scope** by design. One junk file therefore restored exactly the behavior this PR exists to remove:

```text
182-DLL framework scope             0.9s /  90 MB
same scope + one 0-byte .dll        9.5s / 959 MB
```

The optimization was dead, silently, on a shape that is not exotic at all.

**One correction to the report, which changed the fix.** GPT attributed the failure to `AssemblyInspectionSession.Open`. It is not there — `PEReader` is lazy, so opening a 0-byte file *succeeds* and the header parse throws on first use:

```text
--- empty.dll (0 bytes)          --- truncated.dll (18 bytes)
  Open()        : ok               Open()        : ok
  HasMetadata   : BadImageFormat   HasMetadata   : BadImageFormat
  IdentityNames : BadImageFormat   IdentityNames : BadImageFormat
  BodyIndex.Open: BadImageFormat   BodyIndex.Open: BadImageFormat
```

I applied the fix at `Open` first and **my own new test failed**, which is what surfaced the real boundary. The split has to fall at `HasMetadata`. That last line is the soundness argument: `LibraryBodyIndex.Open` parses the same headers from the same file and throws in the same place, so ruling the image out matches what opening it for analysis would have produced.

`ScopeIdentity` now separates opening, proving the image is a managed PE, and reading identity — three phases that fail with different consequences. An image that *parses* but whose metadata reads badly is still `Unknown`, preserving round 4's fix.

### Important: still quadratic, on same-named candidates (Gemini 3.1 Pro and GPT-5.6, independently)

Every selected candidate re-enqueued its own canonical name, and each dequeue re-traversed that name's entire adjacency list — quadratic in the size of the largest same-canonical-name group. Copies of one assembly across subdirectories, or facades canonicalizing together, hit it directly.

Confirmed independently, and it was worse than either reviewer measured:

```text
K= 5000    158 ms      K=20000   2512 ms
K=10000    630 ms      K=40000  10056 ms
```

Clean 4×-per-doubling. Consuming the adjacency entry on first visit makes each name cost one traversal — a second visit could never select anything new anyway:

```text
K= 5000      5 ms      K=20000      5 ms
K=10000      4 ms      K=40000     13 ms
```

### Evidence at `9a5bf651`

| Case | base | head |
| --- | --- | --- |
| unscoped `Call Graph` | 0.8s / 86 MB | 0.9s / 87 MB |
| unscoped `Callers` | 0.7s / 82 MB | 0.7s / 82 MB |
| 182-DLL `Call Graph` miss | 9.6s / 961 MB | **0.9s / 90 MB** |
| 182-DLL `Callers` miss | 6.5s / 697 MB | **0.7s / 82 MB** |
| **junk + 182 `Call Graph` miss** | 9.3s / 963 MB | **1.0s / 90 MB** |
| **junk + 182 `Callers` miss** | 6.8s / 698 MB | **0.7s / 84 MB** |
| corelib hit | 9.7s / 726 MB | 10.2s / 725 MB |
| corelib hit, junk in scope | 10.2s / 726 MB | 10.3s / 724 MB |
| ambiguity negative / positive | 7 ln / 9 ln | 7 ln / 9 ln |

Every row byte-identical except the declared native-only-scope row (#3351).

**On the residual cost, since it would have been easy to wave away.** Gemini measured head *slower* than base on 4000 duplicate assemblies, and that reproduces after the fix (4793 ms → 6030 ms). So I scaled it rather than assert it was the known constant:

```text
n=1000  base 1561  head 1845  delta  284  delta/n 0.28 ms
n=2000  base 2668  head 3276  delta  608  delta/n 0.30 ms
n=4000  base 4568  head 5656  delta 1088  delta/n 0.27 ms
n=8000  base 7439  head 9067  delta 1628  delta/n 0.20 ms
```

Flat per assembly, falling at the top end: this is the disclosed one-extra-metadata-open cost, not surviving superlinearity. It is also the worst case for this change — 5 KB stubs that all genuinely are callers, so the filter can rule nothing out and the near-free decode it doubles is all there is to measure. At realistic assembly sizes the decode dominates and it is ~2%.

Both fixes are mutation-tested: restoring either the `Unknown` classification or the non-consuming lookup fails its new test.

Suites at baseline — Analysis **553 / 1 pre-existing**, CLI **2227 / 1 pre-existing + 10 skipped**.

### Also verified clean by both reviewers

`Unopenable` vs `MethodBodyInspectionSession.Open` soundness, the `Index` last-entry dedup, ordering during index construction, partial-identity fail-open, cross-lens caching and handle retention, corelib hit equivalence, and the reverse 1200-assembly chain.

Round 6 is running against `9a5bf651`. Not ready to merge until a round comes back clean.

richlander:
## Round 6 (Gemini) — reconciliation

Head reviewed: `9a5bf651`. Fixed at **`08ee29dd`**.

### The runtime survived; the test did not

Gemini attacked the round-5 fixes directly — the `Remove`-based walk, the three-phase
identity split, the four-state model — and found no defect in any of them. What it did
find is that **one of my new tests was vacuous**.

`RepeatedNamesIndexACandidateOnce` claimed to pin the last-entry dedup in
`CallerScopeFilter.Index`. It does not. I reproduced this immediately: deleting the dedup
left **28/28 passing**.

The reason is structural, and I should have seen it when I wrote the test. The dedup
collapses only *consecutive* duplicate references from a single candidate. Selection is a
set-reachability question, so a duplicate entry in the adjacency list can never change
*which* assemblies are selected — only how many times the walk steps over the same entry.
**A result-based test cannot distinguish this property at all.**

### I wrote a vacuous test after recording that exact lesson

This PR already carries the note "always mutation-test a regression guard before shipping
it," from round 1 where two of my three attempts proved nothing. In round 5 I
mutation-tested both *fixes* — and neither *test*. The guard got the scrutiny; the
instrument measuring it did not. The rule needed was the stronger one: **mutation-test every
new test, not just every new guard.**

### The property does have a signature — in cost, not in answers

Rather than delete the test, I checked whether the dedup is observable anywhere. It is,
decisively, in allocation:

| duplicate refs | with dedup | without |
| --- | --- | --- |
| 1,000 | 592 B | 8,944 B |
| 100,000 | 592 B | 1,049,496 B |
| 1,000,000 | 592 B | 8,389,600 B |

Flat versus linear. So the old test is now split in two:

- **`RepeatedNamesSelectTheSameAsDistinctOnes`** — the three original asserts under an
  honest name. Duplicate references do not change selection. True, worth holding, and no
  longer claiming to pin the dedup.
- **`RepeatedReferencesDoNotGrowTheIndex`** — pins the dedup itself. 100k duplicate
  references, a warm-up call to settle first-use allocation, then a
  `GC.GetAllocatedBytesForCurrentThread()` delta asserted under 64 KB.
  `GetAllocatedBytesForCurrentThread` is per-thread and deterministic, so this is not a
  flaky timing bound. 64 KB sits ~110x above the real value and ~16x below the mutant —
  it discriminates without being a tight measurement that will bit-rot.

Mutation-tested, both directions: removing the dedup **fails** the new test and **passes**
the old one. Filter suite **29/29**.

Allocation is also the right instrument on the merits, not just a convenient one. The
consumer driving #3333 is the wasm site, where memory is the binding constraint and
parallelism is unavailable. A memory optimization with no memory test was the gap.

### Merge of `origin/main` is orthogonal to this review

`bbadf8e7` merged `origin/main` to resolve a conflict in `LibraryBodyIndexTests.cs`: this
branch predated #3344, which moved `CallTreeFanInCountsDistinctCallersNotCallSites` and
`BuildRepeatedCallSiteFixture` out of `LookalikeEnumeratorCollection`. Resolved in favour of
main; verified the moved test and helper now appear exactly once and that this branch's two
caller-tree tests survived.

`git diff 9a5bf651 bbadf8e7` touches only main's decompiler pass and test files.
**`CallerScopeFilter.cs`, `AssemblyIdentityScanner.cs`, and `ApiMemberAnalysisInspection.cs`
are byte-identical across it.** Round 6's verdict on the runtime therefore applies unchanged
at the merged head; the merge does not by itself require another round.

### Validation at `08ee29dd`

- `ILInspector.Analysis.Tests` — **554 / 1 pre-existing** (`CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition`).
- `dotnet-inspect.Tests` — **2227 / 1 pre-existing + 10 skipped**.
- `ILInspector.Decompiler.Tests --gate fast` — 4017 / 4 failures, all CRLF assertions in
  `StringSwitchRaisingTests` and `CorpusSensorComparisonTests`. This branch changes **no**
  decompiler file (`git diff origin/main...HEAD` lists none), and those test files are
  untouched by the merge, so these are pre-existing on main in this Windows environment.

Round 6 (GPT) is still running against `9a5bf651`; I will reconcile it against this head when
it reports, and will not mark this ready until every required fixed-head review is clean.

richlander:
## Round 6 (GPT) — reconciliation

Head reviewed: `9a5bf651`. Addressed at **`4ac8da2a`**.

GPT independently found the same vacuous test Gemini did — already fixed in `08ee29dd` and
reconciled [above](https://github.com/richlander/dotnet-inspect/pull/3339#issuecomment-5099020746).
Two further points, both worth acting on, though only one as a code change.

### GPT was right that the linear-time guard barely worked

This one is a straight hit and it corrects a number I had put on this PR. I reported the
quadratic as "~10s for 40k sharers against ~13ms after." **That figure came from a standalone
probe, not from the test as written.** Measuring the actual test, the mutant lands at **~2080ms
against its 2000ms bound — a 4% margin.** On slightly faster hardware the regression this test
exists to catch would pass.

I had cited the wrong measurement for the right test, which is exactly the failure mode that
makes a guard look stronger than it is.

The fix exploits the asymmetry rather than moving the threshold: the mutant grows quadratically
and the fixed path linearly, so enlarging the group widens both margins simultaneously.

| sharers | fixed | reverted to a non-consuming lookup |
| --- | --- | --- |
| 40,000 | ~13ms | ~2,080ms |
| 120,000 | ~40ms | ~18,509ms |

3x the group costs the mutant 8.9x the time and costs the fixed path almost nothing. At 120k the
bound sits **~50x above the linear cost and ~9.2x below the quadratic one**. Mutation-tested at
the new size: the mutant fails at 18,449ms. The healthy test still completes in well under a
second, so this buys margin without buying suite time. The margin table is now in the test's
comment so the next person does not have to re-derive it.

### The concurrent-build race: reproduced, then reframed by measurement

GPT's Important finding is real as described: a scope DLL that is zero-byte when classified and
replaced by a build ~600ms later is skipped permanently, where analysis without the prefilter
opens it later and reports its callers.

I did not take the suggested fix. **Failing open on parse failures would nullify the
optimization** — a native DLL beside a managed one is the ordinary case for `--bin`, and
`Unopenable` not selecting is precisely what keeps one junk file from selecting the whole scope
(that was round 5's blocking finding, in the other direction).

So rather than argue the tradeoff, I swept the replacement delay against both binaries:

| replace at | without prefilter | with prefilter |
| --- | --- | --- |
| 0–200ms | finds caller | finds caller |
| 400–1500ms | **finds caller** | **skips** |
| 2000ms+ | skips | skips |

**Analysis without the prefilter is timing-dependent on this input too.** It flips at ~1800ms;
the prefilter flips at ~300ms. It samples later only because it does more work before reaching
the file — 1828ms versus 449ms end to end on this scope. GPT's 600ms sits inside the window where
the two sampling instants differ, which is what makes the two runs disagree.

There is no correct answer being lost here. A scope directory being rewritten mid-run has no
stable state to sample, and **any** speedup moves the sample earlier — the general form of this
finding is "a faster tool sees a half-written file in its broken state more often," which is a
property of going faster, not of this design. The prefilter narrows that window; it does not
introduce nondeterminism a stable scope would have avoided.

Recorded as a **documented boundary** instead of a fix. `ScopeIdentity`'s doc comment now states
that the classification is a single sample of a file the tool does not own, carries the measured
flip points, and says plainly that a caller needing a reproducible answer must present a scope
that is not being written. That change is comment-only — every added line in
`ApiMemberAnalysisInspection.cs` is a `///` line.

### A hypothesis I tested and threw away

While reproducing, I expected to find the unfiltered path is *worse* than this: it opens scope
sessions per lens, so a file changing between the `Callers` open and the `Call Graph` open should
make a single invocation internally inconsistent. That would have been a nice counterpoint. I
swept ten delays spanning both lens opens and **it never happened** — the two sections agreed in
every run. Dropping the claim rather than shipping it as an argument.

### Validation at `4ac8da2a`

- `ILInspector.Analysis.Tests` — **554 / 1 pre-existing**.
- `dotnet-inspect.Tests` — **2227 / 1 pre-existing + 10 skipped**.
- Filter suite 29/29; both round-6 changes mutation-tested.

Round 7 is running against this head with both reviewers. Not marking ready until it comes back clean.

richlander:
## Round 7 — reconciliation

Head reviewed: `4ac8da2a`. Fixed at **`0589b6c9`**.

**GPT: clean.** Its strongest contribution was fuzzing the `Unopenable` soundness claim rather than
reading it — 21,505 malformed PE variants, of which 2,659 classified `Unopenable`, and **zero** could
have been successfully body-indexed. That is the attack I most wanted run against the `HasMetadata`
split. It also held the allocation guard across 40 isolated runs, 10 class runs and two full-suite
runs (2,000–11,208 bytes against the 65,536 bound), and independently verified the comment-only claim.

**Gemini: one Important finding, and it is the most uncomfortable one in eight rounds.**

### I had already seen this difference, declared it, and declared it wrong

The equivalence matrix on this PR carries a `native-only scope` row reading **72 lines against 70**. I
saw it, attributed it to the two `BuildCallerTree` builders disagreeing, filed #3351, and moved on.

That attribution was wrong. The builders disagreeing is pre-existing — but **which builder runs for
this input was changed by this branch.** Reproduced independently: base 62 lines, head 60, with
different revisit markers and node ordering.

Worse, **two tests on this PR asserted the wrong contract and protected the bug**, with comments
rationalizing it ("Either way the scope yields nothing, and the request was still scoped"):

- `CallerScopes_WhenNoScopeAssemblyCanBeOpened_StillSelectsTheCrossAssemblyBuilder`
- `CallerScopes_WhenTheOnlyScopeEntryIsNotManaged_StillSelectsTheCrossAssemblyBuilder`

A mislabeled matrix row, a tracking issue absorbing the blame, and two green tests pinning the wrong
answer — that combination is what kept this alive for three rounds.

### The actual defect

Base routed on `callerScopes is not { Count: > 0 }`, so it took the token-keyed builder whenever its
opened list came back empty. A native-only or malformed-only `--bin` directory produced exactly that.
This branch re-keyed the routing to `null`, which was **necessary** so a scope whose entries are
merely ruled out by the closure keeps the structural builder — but it also sent nothing-openable down
the structural path for the first time.

The distinction routing needs is **"was anything here openable at all"**, not "did anything survive
selection". Those coincide in base and diverge under prefiltering. `CallerScopes` now returns `null`
when every candidate classified `Unopenable`, which is exactly when the unfiltered walk would have
finished empty.

| scope | base | head | identical |
| --- | --- | --- | --- |
| native-only | 62 | 62 | yes |
| native + managed-but-irrelevant | 60 | 60 | yes |
| malformed-only | 62 | 62 | yes |
| empty directory | 62 | 62 | yes |

The first row was 62 against 60 before this commit.

Both tests are corrected, and `CallerScopes_WhenAnOpenableScopeEntryIsRuledOut_StillSelectsTheCross
AssemblyBuilder` is added as the counterpart pinning the other side — so the pair cannot both be
satisfied by routing on the survivor count, which is what the old pair allowed.

Round 2's finding still stands and bounds this: an image whose identity tables read cleanly but whose
bodies fail to index is genuinely undecidable, and remains the residual gap. But that is a much
narrower claim than the one I used to justify the whole case.

#3351 has been corrected accordingly.

### Separately: the concurrent-rewrite comment was overclaiming

I asked a rubber-duck agent to attack my own round-6 non-action, specifically because preserving my
optimization gave me a motive to dismiss GPT's Important finding. It found a case neither GPT nor I
had: **the race does not require a malformed file at all.** A valid assembly replaced by a *different*
valid assembly between classification and body indexing produces the identical miss.

| replace at | base finds caller | head finds caller |
| --- | --- | --- |
| 0–200ms | yes | yes |
| 400–1600ms | yes | **no** |
| 2000ms+ | no | no |

So my comment's framing — "prefiltering narrows the window in which a half-written image is seen as
finished" — was wrong twice: it is not about half-written images, and the false-negative window is
**wider**, not narrower. Corrected in `0589b6c9`.

The duck's proposed fix does not survive its own finding, which is why the non-action stands on better
ground now: revalidating only *transient* failures would address the zero-byte case and miss the
valid-to-valid case entirely, because there the candidate classifies cleanly as `Known` and never
signals a failure. Closing the general case needs a second full identity sweep — doubling the
prefilter's cost — and still races. Tracked with measurements in **#3359**.

### Validation at `0589b6c9`

- `ILInspector.Analysis.Tests` — **554 / 1 pre-existing**.
- `dotnet-inspect.Tests` — **2228 / 1 pre-existing + 10 skipped** (2227 baseline plus the new counterpart test).
- Routing guard mutation-tested: inverting the openable test fails both routing tests.
- 182-DLL `Call Graph` still **9915ms/964MB to 925ms/93MB**, byte-identical.

Round 8 is running against this head with both reviewers. Not marking ready until it is clean.

richlander:
## Round 8 reconciliation (`0589b6c9`)

Two reviewers at the same fixed head. **Three findings, one of them real and now fixed in
`a5ae4fe8`; one refuted with a counterexample; one confirmed and filed.** Also the first round
where a reviewer reported its own pass as incomplete rather than inventing a verdict, which was
the right call and is worth recording.

### GPT — reported its own pass incomplete, then completed it

Its first response came back empty. Asked to restate it, it said plainly that it had not finished
and listed exactly what it had skipped, rather than reconstructing a plausible report. I asked it
to finish the outstanding items and it did.

**No findings.** Coverage worth keeping:

- **15 scope shapes, base vs head, byte-identical on every one** — including shapes I had never
  tested: netmodule, reference assembly, ReadyToRun image, symlink, renamed single-file bundle,
  a managed DLL locked with `FileShare.None`, directories-only, duplicate paths, and
  case-differing duplicate paths.
- Analysis 554/1 pre-existing, CLI 2228/1 pre-existing — both at baseline.
- It declined to mutate the reviewed source, on read-only grounds. Reasonable; I had already
  mutation-tested that guard myself, and did so again for the new one below.
- **It reproduced the documented residual gap concretely**, which had been theoretical until now.

### Gemini — one Important finding, real, now fixed

**The routing flag was set by candidates this walk opens itself.** A candidate that
classification cannot decide is always *selected*, so the open settles whether the scope really
had a session — but the flag had already been set from its classification, so a failed open
still took the structural builder where the unfiltered walk took the token builder.

This is the same family as round 7's regression, one level in: round 7 asked the wrong
*question*, round 8 asked the right question of the wrong *population*.

Fixed in `a5ae4fe8` by deriving the flag only from candidates selection **ruled out** — the only
ones this walk never opens, and therefore the only ones where classification is the sole
available evidence. Survivors are opened, so `opened` answers for them directly.

**I did not take the suggested fix.** Gemini proposed classifying unexpected exceptions as
`Unopenable`. That trades a routing mismatch for a *missed-caller* soundness risk: an
unanticipated failure says nothing about whether analysis could open the image, and ruling it out
would drop real callers. The fix above restores equivalence without weakening selection at all —
selection is byte-for-byte unchanged.

**The new test was vacuous on its first attempt and I caught it by mutation-testing.** My first
fixture used a directory named `*.dll`, assuming it classified as undecidable. It classifies as
`Unopenable`, so it was ruled out and the test passed for the wrong reason — the mutation
survived. Second attempt used a fat body header with `CodeSize = int.MaxValue`; that turned out
to be *suppressed* (see below) so the fixture opened fine. The third encodes Gemini's actual
reported path. Reverting the flag to every candidate kills it and only it, with a verified
`0 Error(s)` build.

### Gemini's Minor — refuted by counterexample

Gemini reported the documented residual gap as a "phantom" and recommended **deleting** the
comment, on the grounds that `LibraryBodyIndex` suppresses `BadImageFormatException` per method
so `Open` never throws on corrupt bodies.

The mechanism is right; the conclusion does not follow. Gemini corrupted IL opcodes to `0xFF` —
one corruption mode — and generalized. GPT independently produced an image that *does* fault, and
I reproduced it:

```
Error: Read out of bounds.
```

with the identical file uncorrupted producing normal output. Base 62 lines (token), head 63
(structural); uncorrupted, both 63 and byte-identical. **One byte is the entire difference.**

Acting on this finding would have deleted a true caveat and closed a real issue.

That said, the finding was valuable, because **my comment was also wrong** — in the tool's
favour, again. Byte 857 lands inside the *metadata streams* (which start at offset 624), not in a
method body; bodies in that fixture are around offset 592. So:

- failures **inside a method body** are suppressed — verified by wrecking a body header with
  `CodeSize = int.MaxValue`, which opens fine and is byte-identical base to head;
- what escapes is a fault in **metadata that body indexing reads and identity scanning never
  touches**.

Both the comment and the issue are corrected. Filed as **#3361**, and the fix above shrinks it:
it now requires an image that reads cleanly, faults exactly where body indexing reads it, *and*
is ruled out by the closure. A **selected** candidate that fails to open now matches base
exactly.

### Verification at `a5ae4fe8`

| scope shape | base | head | |
| --- | --- | --- | --- |
| native-only | 62 | 62 | identical |
| native + managed-irrelevant | 63 | 63 | identical |
| malformed-only | 62 | 62 | identical |
| empty directory | 62 | 62 | identical |
| suppressed body corruption | 63 | 63 | identical |
| ruled-out metadata fault | 62 | 63 | **#3361, documented** |

- 182-DLL framework scope: **10927ms / 1198MB → 4236ms / 335MB**, output identical.
- Analysis **554 / 1 pre-existing**; CLI **2229 / 1 pre-existing + 10 skipped** (+1 is the new test).
- Routing tests 13/13; new test mutation-verified.

### Not ready to merge

`a5ae4fe8` changes behavior in response to a review finding, so it needs its own fixed-head
review. Round 9 next.

richlander:
## Round 9 reconciliation — `a5ae4fe8`

Both reviewers returned **clean**. Recording what each executed, one gap I found in
*their* evidence, and the comparison I ran myself to close it.

### Gemini 3.1 Pro — clean

Its first report was thin, so I asked for executed artifacts and explicitly permitted
"I reasoned about it" as an answer. It said plainly that it had not executed everything,
then ran the checks. The valuable executed result:

```text
PROBE_KIND: Unknown
PROBE_SELECTED: True
```

That independently confirms the new test exercises the intended path — an embedded-null
scope path classifies `Unknown`, is therefore **selected**, and so is one of the candidates
this walk opens itself. That is exactly the population round 8 showed must not feed the
routing flag.

### GPT-5.6 — clean

No new findings against the round-8 fix. It did not execute the reverse lens-order check
(reasoned only); I ran it myself below.

### The gap in both reviewers' evidence

**Both reviewers' base/head equivalence tables were insensitive.** Gemini's table showed
11 lines and the identical hash for *every* scope shape. An equivalence table only means
something if the comparison can detect the difference it claims is absent — and the small
call-graph fixtures **cannot distinguish the two caller-tree builders**. Rendering
`Target.Api.Ping:1` gives byte-identical output whether routing picks the token builder or
the structural one, so a base-vs-head table built on it proves nothing about routing.

I hit the same trap myself in parallel: my first combination scopes also came back
"11 lines, IDENTICAL" everywhere, and testing sensitivity showed the token-path and
structural-path outputs for that member are equal. My own comparison had proved nothing
either.

### The sensitive comparison, which I supplied myself

Sensitive setup: library `dotnet-inspect.dll`, member
`DotnetInspector.Inspectors.MethodBodyInspectionSession.Open:1` — **token 62 lines,
structural 63**. Four combination scopes covering the selected/ruled-out split the round-8
fix turns on:

| scope shape | base | head | | expected builder |
| `A` selected-fails-to-open only | 62 | 62 | identical | token |
| `B` `A` + a ruled-out **openable** candidate | 63 | 63 | identical | structural |
| `C` `A` + an `Unopenable` candidate | 62 | 62 | identical | token |
| `D` ruled-out openable only | 63 | 63 | identical | structural |

**Sensitivity proved, not assumed:** `A` and `B` differ from each other (62 vs 63), so this
comparison can see a routing change. Contrast with the reviewers' tables, where every cell
was the same value.

`A` is precisely the case `a5ae4fe8` fixed: a candidate that classification cannot decide is
always *selected*, so this walk opens it and learns the truth. When that open fails and no
ruled-out candidate was openable, base ended empty and took the token builder — and now head
does too. `C` confirms an `Unopenable` neighbour does not change that; `B` and `D` confirm a
single ruled-out openable candidate flips **both** base and head to structural, together.

Lens order, on both sensitive shapes:

```text
A: head order-stable=True   base-vs-head order1=True order2=True
B: head order-stable=True   base-vs-head order1=True order2=True
```

### Standing at this head

- All scope shapes byte-identical to base **except** the documented ruled-out-metadata-fault
  gap tracked in #3361 (62 vs 63).
- 182-DLL framework scope: **10927 ms / 1198 MB → 4236 ms / 335 MB**, output identical.
- Analysis 554 / 1 pre-existing; CLI 2229 / 1 pre-existing + 10 skipped; routing class 13/13.
- CI green: `build-net10`, `test`, `csharp-diff-smoke`, `il-diff-smoke`.

### Does round 9 count as clean?

Yes — but not on the reviewers' tables alone, and it is worth saying so rather than papering
over it. Both reviewers read the round-8 fix and found nothing, Gemini executed a probe that
independently confirms the test hits the intended path, and the routing equivalence claim is
now backed by a comparison that has been *shown* to be sensitive. That last piece came from
me, not from either reviewer.

Nine rounds, eight of which found a genuine defect. The recurring lesson is not adversarial
reading but **adversarial input shape** — and, this round, the second-order version of it:
an equivalence table is only as good as its ability to detect the difference it reports absent.

Ready to merge.

PR #3344, opened 2026-07-27
Move the fan-in test into the class its filter names

richlander:
`CallTreeFanInCountsDistinctCallersNotCallSites` was defined inside `LookalikeEnumeratorCollection` — a fixture data class — instead of `LibraryBodyIndexTests`.

xUnit still discovered and ran it in a full suite run, so it was never silently skipped. But the class filter documented for this area could not match it:

```
dotnet run --project src\ILInspector.Analysis.Tests -c Release -- -class "ILInspector.Analysis.Tests.LibraryBodyIndexTests"
```

Anyone narrowing to the call-tree tests while working on fan-in would have run everything *except* the test that pins fan-in.

This moves the test and its `PersistedAssemblyBuilder` helper into `LibraryBodyIndexTests` and restores `LookalikeEnumeratorCollection` to the fixture it is documented to be. No assertion changes.

**Evidence**

- The documented filter now selects it: `-class "...LibraryBodyIndexTests" -method "...CallTreeFanInCountsDistinctCallersNotCallSites"` → `Total: 1, Failed: 0`. Before the move that filter could not have matched, since the test was not in that class.
- Full suite total unchanged at **523**, 1 pre-existing unrelated failure (`CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition`) — confirming the test was already running and this only relocates it.

Follow-up from the round-3 adversarial review of #3312, reported independently by both reviewers.

**Review tier: trivial.** This moves an existing test between classes with no change to its assertions, its helper, or any product code, and the suite total is provably unchanged. No review requested.

### DIFF
--- a/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
+++ b/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
@@ -3550,6 +3550,64 @@ public class LibraryBodyIndexTests
         Assert.Equal(CallerUnsafeMode.Implicit, pointerExtern.Mode);
         Assert.DoesNotContain(top, e => e.Method.Name == nameof(UnsafeEvidenceFixtures.CallsUnsafeAs));
     }
+
+    [Fact]
+    public void CallTreeFanInCountsDistinctCallersNotCallSites()
+    {
+        var (path, directory) = BuildRepeatedCallSiteFixture();
+        try
+        {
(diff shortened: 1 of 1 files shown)


richlander:
Ready to merge. Trivial tier: moves an existing test between classes with no product change, so no adversarial review. Verified the documented `-class` filter now selects it (`Total: 1`) and the suite total is unchanged at 523. CI clean.

Issue #3341, opened 2026-07-27
Make the metadata-table projection consumable by the wasm Metadata Explorer: paging, row lookup, allocation shape


richlander:
## Summary

The metadata-table projection (#3282, landed in #3301 / #3318 / #3334) is a good
foundation for the **Metadata Explorer** planned for the wasm site, and its
handle-edge design is the reason: `HandleRef(TargetTable, TargetRowId, Token,
Display)` is exactly the click-through navigation primitive an explorer needs —
a typed edge rather than display text.

What the explorer still needs is **access-pattern** work, not model redesign.
This issue tracks the additive changes that make the projection consumable by a
browser host with a hard memory ceiling. None of them alter the existing model's
shape or the CLI/`mdi` behavior.

## What already works for a browser host (verified)

- `MetadataTableProjector.Project(PEReader, MetadataProjectionOptions?)` is
  public and takes a `PEReader`
  (`src/ILInspector.Metadata/MetadataTableProjector.cs:62`), so a wasm host can
  do `new PEReader(new MemoryStream(fetchedBytes))` with **no filesystem**.
- `options.Tables` already gives lazy per-table loading.
- Budgets (`MaxRowsPerTable`, `MaxPreviewBytes`, `MaxStringChars`) plus an
  explicit `MetadataTableTruncation` mean truncation is *visible*, which is a UI
  signal as much as a safety property.
- Per-row `MetadataValue.Malformed` containment means a corrupt row renders as a
  bad row instead of killing the view — important when users point the explorer
  at arbitrary uploads.
- Constraints already match the wasm host: SRM-only, NativeAOT-friendly,
  Roslyn-free, read-only, no inspected-assembly loading.
- Correctness is externally anchored by the `mdv` oracle
  (`MdvOracleComparisonTests`), which now covers all 17 supported tables.

Note for consumers: the *session* entry points are desktop-shaped —
`AssemblyInspectionSession.Open` takes a path or a `ResolvedAssemblyReference`,
and `AssemblyImage.FromStream` is private
(`src/ILInspector.Metadata/AssemblyImage.cs:40`). A browser host should bind to
`MetadataTableProjector.Project(PEReader, …)` directly. If that seam proves
awkward in practice, exposing a stream/bytes-based image open is a candidate
increment — but it is not required.

### 1. No paging — an offset is not expressible

`MetadataProjectionOptions` has `MaxRowsPerTable`
[...]

Issue #3282, opened 2026-07-27
Raw metadata-table row projection: enumerate ECMA-335 tables with resolvable handles


richlander:
Add a first-class capability to project the raw ECMA-335 metadata tables of an assembly — enumerate a table, emit each row with its columns, and expose every coded index / handle as a resolvable link to the row it points at. This is the ground-truth substrate the rest of the tool is cooked from, and it belongs in the Metadata layer as a reusable projection rather than a bespoke reader wired to a single consumer.

## Motivation

Raw metadata-table viewing is a staple diagnostic in this space (SRM’s own \`mdv\`/MetadataVisualizer, ildasm’s metadata view, ILSpy’s Metadata tab). For dotnet-inspect it earns its keep in several ways:

- **Diagnostics / evidence.** When a typed projection (ApiSurface, Findings) looks wrong, dropping to the raw row is the fastest way to tell whether the defect is in the metadata itself or in the higher-level cooking. This directly serves the repo’s "keep failure visible" principle.
- **Reuse across features.** One projection can back a CLI \`metadata\`/table lens, raw-table version diffs (complementing the existing API version-diff capability), and forensic/corpus inspection.
- **Low marginal cost.** SRM already exposes table enumeration and handles, and \`ILTokenResolver\` already performs token resolution — the new work is mainly coded-index/heap decoding and a stable output shape.

## Scope

1. **Projection (Metadata layer).** Given an assembly image, enumerate the metadata tables (\`Module\`, \`TypeRef\`, \`TypeDef\`, \`Field\`, \`MethodDef\`, \`Param\`, \`MemberRef\`, \`CustomAttribute\`, \`AssemblyRef\`, …) and produce, per table, its rows with:
   - each column value in raw form, plus a friendly decode where cheap (e.g. flags, name/namespace from the string heap);
   - handle-typed columns surfaced as resolvable references (target table + row index / token) so a consumer can follow a coded index to its target row.
   - Reuse \`ILTokenResolver\` for resolution; do **not** fork a second decoder.
2. **Output shape.** A structured model (table → rows → columns, with typed handle references) plus a Markdown rendering, following the existing section/output-shape and progressive-disclosure conventions. Large heaps (string/blob/user-string/guid) should be lazily/optionally surfaced, not dumped by default.
3. **Surface.** Expose it as its own \`metadata\` table lens/section so it composes with the existing verbosity model and can be selected explicitly.

## Layering constraint

[...]

PR #3345, opened 2026-07-27
Give the metadata projection a row window and on-demand row lookup

richlander:
Serves #3341 acceptance items 1, 2, and 3. The wasm Metadata Explorer browses metadata tables interactively rather than dumping them, and the projection could not serve that.

## What was blocking

- `MaxRowsPerTable` always counted from row 1 (`for (int rid = 1; rid <= limit; rid++)`), so **"page 2" was inexpressible**.
- Following a `HandleRef` whose target row fell outside the current budget **dead-ended**; the only recovery was re-projecting the whole target table.
- `Project(PEReader, ...)` was already filesystem-free, but nothing pinned that, so it could regress silently.

## What this changes

**Row window.** `MetadataProjectionOptions.StartRowId` pairs with `MaxRowsPerTable` to form a per-table window. A window changes **coverage, never content**:

- rows keep their absolute `RowId` and `Token`;
- `MetadataTableView.RowCount` stays the physical count;
- partial coverage is still marked by `Truncation`;
- a window past the end of a populated table yields that table **with zero rows** rather than dropping it, so paging cannot make a populated table look absent;
- the window end is widened to `long` before the comparison, so a start near `int.MaxValue` cannot overflow to a negative end and wrap back into real rows.

**On-demand row lookup.** `MetadataTableProjector.ProjectRow(peReader, table, rowId)` is the click-through primitive. It reads one row independent of any window on the containing table and returns it **inside that table's view**, so the caller also gets the column schema and the table's real size — neither of which a bare row can carry. It is a one-row window over the shared `BuildView`, **not a second row path**, so malformed-row containment and every budget behave identically. Returns `null` for no metadata, an unprojected table, or a row id past the end.

**Presentation must name the window.** Because a window can start anywhere, `showing 3 of 366 rows` would read as the *first* three rows. The renderer heading becomes `showing rows 4–6 of 366` when a window does not start at row 1, and `mdi` gains `--start-row|-s` with a matching stderr note for the machine formats.

## Compatibility

Purely additive. `StartRowId` defaults to 1, so every existing projection, every `mdi` invocation, and every rendered heading is byte-identical to before. The only behavior reachable solely through the new option is the past-the-end empty view.

## Evidence

- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release` — **966 passed, 0 failed, 0 skipped** (up from 947; the mdv cross-implementation oracle ran, it did not skip).
- `dotnet run --project tests/DotnetInspector.MetadataRendering.Tests -c Release` — **33 passed, 0 failed** (up from 30).
- `dotnet build dotnet-inspect.slnx -c Release` — clean, 0 errors.
- `git grep` confirms these two suites cover **every** consumer of `MetadataProjectionOptions`/`MetadataTableTruncation` in the repo; the `dotnet-inspect` CLI does not consume the projection.

19 new tests, including the ones that would catch this going wrong:

- pages of 3 over TypeDef **reconstruct the full row-id sequence exactly** — no gap, no duplicate, no reorder;
- a windowed row's cells are **equal** to the same row from an unwindowed projection;
- `StartRowId = int.MaxValue` with `MaxRowsPerTable = int.MaxValue` projects nothing (the overflow guard);
- a start below 1 clamps, matching how `MaxRowsPerTable` already clamps;
- a window past the end keeps the table visible with `Truncation(0, RowCount)` and its full schema;
- `ProjectRow` reaches the last row while the caller's window covers only row 1, and its cells match the full projection;
- a real `HandleRef` is followed to its target row and the returned row's `Token` matches the edge's;
- `Project` is driven from a `MemoryStream` over bytes, never a path.

End-to-end through `mdi` on `ILInspector.Metadata.dll` (366 TypeDef rows), showing pages are contiguous and correctly labelled:

```
$ mdi ILInspector.Metadata.dll -t TypeDef -n 3
## TypeDef (showing 3 of 366 rows)
| 1 | NotPublic | <Module> | ...
| 2 | Sealed, BeforeFieldInit | <>z__ReadOnlyArray1 | ...
| 3 | Sealed, BeforeFieldInit | <>z__ReadOnlyList1 | ...

$ mdi ILInspector.Metadata.dll -t TypeDef -s 4 -n 3
## TypeDef (showing rows 4–6 of 366)
| 4 | Sealed, BeforeFieldInit | <>z__ReadOnlySingleElementList1 | ...
| 5 | ... | 6 | ...

$ mdi ILInspector.Metadata.dll -t TypeDef -s 4 -n 3 -f jsonl
Note: table TypeDef shows rows 4 to 6 of 366; raise --max-rows or move --start-row to include more.
{"table":"TypeDef","rid":"4",...}
```

## Not in scope

Allocation shape (#3341 increment 3) stays **measurement-gated** — no benchmark exists yet, and AGENTS.md forbids claiming runtime impact from static structure. Reverse index/search and heap/header surfacing remain open. The `dotnet-inspect metadata` CLI lens (#3282) is deliberately sequenced last.

### DIFF
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -345,8 +345,8 @@ inspector), a standalone tool that renders the tables the way `mdv` does:
   (`ToolCommandName=mdi`) whose System.CommandLine front-end maps flags onto
   `MetadataProjectionOptions` and delegates all output to the renderer below.
   Surface: `mdi <assembly> [--table|-t <Names>] [--format|-f md|tsv|jsonl]
-  [--max-rows|-n N] [--max-bytes N] [--max-chars N]`. Missing files, native
-  images, and unreadable metadata surface as visible errors, never
+  [--max-rows|-n N] [--start-row|-s N] [--max-bytes N] [--max-chars N]`. Missing
+  files, native images, and unreadable metadata surface as visible errors, never
   success-shaped empty output.
 - **`src/DotnetInspector.MetadataRendering`** — a small reusable library holding
   `MetadataProjectionRenderer` (projection → Markout tables). It lives in the
@@ -373,6 +373,41 @@ projection **model** (not `mdi`'s rendered text), the renderer is free to be
 human-friendly without weakening the oracle. Supported-table coverage is kept in
 step with the tables `mdv` dumps, so every supported table's physical row count
 is cross-validated against `mdv` for free on the product assembly.
+
+## Implemented: random access for a browser host
+
+The projection's other planned consumer is the wasm **Metadata Explorer** (issue
+#3341), which browses tables interactively rather than dumping them. That host
+imposes three constraints a batch dump does not, all met without changing the
+model:
+
+- **No filesystem.** `MetadataTableProjector.Project(PEReader, options)` is the
+  entry point; a browser host constructs `new PEReader(new MemoryStream(bytes))`
+  and never supplies a path. `AssemblyInspectionSession` stays the desktop-shaped
+  convenience over the same projector.
+- **No whole-table materialization.** `MetadataProjectionOptions.StartRowId`
+  pairs with `MaxRowsPerTable` to form a **row window**, so a table of 5,000 rows
+  is browsed a page at a time. A window changes coverage, never content: rows
+  keep their absolute `RowId`, `MetadataTableView.RowCount` stays the physical
+  count, and any partial coverage is marked by `Truncation`. A window past the
+  end of a populated table yields that table with zero rows rather than dropping
+  it, so paging cannot make a table look absent.
+- **No dead-end handles.** A `HandleRef` is the click-through primitive, and its
+  target frequently lies outside the current window.
+  `MetadataTableProjector.ProjectRow(peReader, table, rowId)` reads that one row
+  on demand and returns it inside its table's view, so the caller also gets the
+  column schema and the table's real size. It is a one-row window over the same
+  reader, not a second row path, so malformed-row containment and every budget
+  behave identically.
+
+Because a window can start anywhere, presentation must **name** the window
+rather than only its size: `mdi` renders `showing rows 4–6 of 366`, and its
+machine formats report the same range on stderr. Reporting `3 of 366` alone
+would read as the first three rows.
+
+Allocation shape — windowed or lazy materialization, and `ReadOnlySpan<T>` views
+over the current `ImmutableArray<T>` — remains open and **measurement-gated**;
+the browser's memory ceiling is the real constraint, and no benchmark exists yet.
 
 ## Layer placement
 
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -376,10 +376,10 @@ is cross-validated against `mdv` for free on the product assembly.
 
 ## Implemented: random access for a browser host
 
-The projection's other planned consumer is the wasm **Metadata Explorer** (issue
-#3341), which browses tables interactively rather than dumping them. That host
-imposes three constraints a batch dump does not, all met without changing the
-model:
+The projection's other planned consumer is the wasm **Metadata Explorer**
+(issue #3341), which browses tables interactively rather than dumping them.
+That host imposes three constraints a batch dump does not, all met without
+changing the model:
 
 - **No filesystem.** `MetadataTableProjector.Project(PEReader, options)` is the
   entry point; a browser host constructs `new PEReader(new MemoryStream(bytes))`
--- a/src/DotnetInspector.MetadataRendering/MetadataProjectionRenderer.cs
+++ b/src/DotnetInspector.MetadataRendering/MetadataProjectionRenderer.cs
@@ -92,9 +92,21 @@ public static class MetadataProjectionRenderer
     }
 
     static string HeadingText(MetadataTableView table)
-        => table.Truncation is { } truncation
+    {
+        if (table.Truncation is not { } truncation)
+            return $"{table.Name} ({table.RowCount} {(table.RowCount == 1 ? "row" : "rows")})";
+
+        if (table.Rows.IsEmpty)
+            return $"{table.Name} (showing 0 of {truncation.RowCount} rows)";
+
+        // A row window that does not start at row 1 must say where it sits:
+        // "showing 4 of 100 rows" alone would read as the first four rows.
+        int first = table.Rows[0].RowId;
+        int last = table.Rows[^1].RowId;
+        return first == 1
             ? $"{table.Name} (showing {truncation.ProjectedRows} of {truncation.RowCount} rows)"
-            : $"{table.Name} ({table.RowCount} {(table.RowCount == 1 ? "row" : "rows")})";
+            : $"{table.Name} (showing rows {first}\u2013{last} of {truncation.RowCount})";
+    }
 
     static void WriteTable(MarkoutWriter writer, MetadataTableView table, bool identifyTable)
     {
--- a/src/ILInspector.Metadata/MetadataTableProjection.cs
+++ b/src/ILInspector.Metadata/MetadataTableProjection.cs
@@ -69,8 +69,8 @@ public sealed record MetadataTableView
     public ImmutableArray<MetadataRow> Rows { get; }
 
     /// <summary>
-    /// Non-null when the row budget stopped enumeration before the physical end
-    /// of the table. Enumeration never silently truncates without this marker.
+    /// Non-null when the row window covered fewer than all of the table's rows.
+    /// Enumeration never silently truncates without this marker.
     /// </summary>
     public MetadataTableTruncation? Truncation { get; }
 }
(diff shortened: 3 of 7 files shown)


richlander:

Two-model review, both isolated in their own detached worktrees. This change rewrites the bounds of a loop that every table's projection runs through, so it took the two-model tier rather than the single-review tier.

| Reviewer | Head reviewed | Verdict |
| --- | --- | --- |
| MAI-Code | `79344d24`, re-reviewed at `2d3832d7` | 1 finding, fixed and re-verified clean |
| Gemini 3.1 Pro | `79344d24` | Clean |

### The one finding, and why the reviewers disagreed

**MAI-Code:** `ProjectRow` does not honour `MetadataProjectionOptions.Tables` — a caller passing `Tables = [TypeRef]` still gets a `TypeDef` row back, contradicting the doc's *"Returns null when … the table is not projected"*. Proved it with a probe against the built library: `rows=1 rowcount=366`.

**Gemini, independently, on the same code:** *"It accurately ignores `options.Tables` to fetch the target regardless of the initial projection filters — this successfully supports cross-table `HandleRef` lookups."*

Two careful reviewers read the same sentence and inverted it. That is the actual defect: the **behavior** is right, the **contract** was not.

Honouring `Tables` would dead-end precisely the edges this method exists to follow — a caller browsing `TypeRef` could not follow a `TypeRef` row into `TypeDef`, which is the explorer's primary interaction. So the behavior stands, and `2d3832d7` fixes the contract instead: the XML doc now states that `Tables` and `StartRowId` are deliberately ignored and why, while the cell budgets still apply, and two tests pin it so a future change cannot "fix" it the wrong way.

MAI-Code re-reviewed at `2d3832d7` and confirmed: the wording is now unambiguous; **both new tests are load-bearing, not incidental** (it probed the library directly — `selection: TypeDef rowId=1` and `budget: rowId=2 truncated=True preview=<>z_` — and confirmed each would fail if the behavior flipped); the diff is doc + tests only with the implementation untouched; and no remaining `ProjectRow`/`Project` mismatch. Finding **resolved**, no new findings.

### What was verified and dismissed

Both reviewers independently confirmed the load-bearing claims:

- **Window arithmetic.** `(int)Math.Min(rowCount, inclusiveEnd)` is unconditionally safe because the minimum is mathematically bounded by `rowCount`, an `int`. `projected` is provably non-negative and exactly equals the loop's iteration count, so the builder capacity cannot mismatch. Widening to `long` before adding correctly defuses `start == int.MaxValue`.
- **"Coverage, never content."** No row reader or cell builder carries state across rows; `MaxStringChars`/`MaxPreviewBytes` apply per cell, and `EscapeCore` acts locally. Gemini specifically chased `ReadAssemblyRow`, which ignores its `rid` — under a window it still returns exactly what the unwindowed path returned, so row content stays a pure function of `rid`.
- **No silent truncation.** `projected` cannot exceed `rowCount` (so the `MetadataTableTruncation` constructor cannot throw), and a `null` `Truncation` is unreachable for partial coverage.
- **Backward compatibility.** `HeadingText` was rewritten from one expression into three branches; both reviewers confirmed the `first == 1` path reproduces the previous strings exactly, including the singular/plural and no-truncation cases.
- **Boundaries, run not reasoned.** `-n 0`, `-s` at `RowCount`, at `RowCount + 1`, and `-s 2147483647` all render an honest empty window with no crash and no wrong label.

### Head delta

Gemini reviewed `79344d24`; the final head is `2d3832d7`. The delta is two commits with **no behavior change**: a Markdown line rewrap (markdownlint MD018 — the issue reference `#3341` had wrapped to the start of a line and was read as a malformed heading), and the doc-comment + tests fix above. MAI-Code re-reviewed the final head and independently confirmed the implementation is untouched across that range.

### Evidence at the final head

- Metadata **968 passed / 0 failed / 0 skipped** (0 skipped means the mdv cross-implementation oracle genuinely ran).
- MetadataRendering **33 passed / 0 failed**.

richlander:
Ready to merge.

CI green at `2d3832d7` (build-net10, test, markdownlint, changes; pack and the diff smokes path-gated out). Two-model adversarial review complete: Gemini 3.1 Pro clean, MAI-Code's one finding fixed and re-verified clean by MAI-Code on the final head. No blocking work remains.

Non-blocking follow-up, tracked in #3341: allocation shape (increment 3) stays measurement-gated pending a benchmark, and reverse index/search plus heap and header surfacing remain open and may warrant their own issues. The `dotnet-inspect metadata` CLI lens in #3282 is deliberately sequenced last.

richlander:
Reviewed the full stack (**3345 → 3348 → 3350 → 3352 → 3360**) as one unit, in a clean detached worktree at `2e8545f1` (the #3352 head, which carries all the product code).

**Verdict: the stack is sound and I'd take it.** One finding, and it lives in #3348, so I've posted it there in full rather than duplicating it here: **#3348 comment 5100236492**.

Short version: `MetadataRowReferenceSet.IsComplete` is `!Truncated && UnreadableRows.IsEmpty`, and its doc says the search "covered the whole image … so an empty `References` genuinely means nothing points at `Target`." But `FindReferences` scans the projector's 17 `SupportedTables`, not all ~38. On this repo's own `ILInspector.Metadata.dll` that leaves 7 populated tables and **12.8% of the image** unscanned, with `IsComplete == true` regardless — so a nested type's `NestedClass` declaring-type edge and a property accessor's `MethodSemantics` edge are both absent from an answer that reports itself complete. A positive control (a `Field` target correctly getting its `TypeDef.FieldList` **Range** edge) confirms the mechanism is right and only its scope is short.

What makes it worth fixing inside the stack rather than filing as a follow-up is that the stack already diagnosed it twice in one document: `metadata-table-projection.md:437` refuses to honor `options.Tables` on exactly this reasoning, and `:485` names `NestedClass`/`MethodSemantics`/`Property` as unmodelled tables while crediting the reverse search with avoiding "the same success-shaped-absence failure." The fix is cheap because #3350 already built the data — see the suggested `UnscannedTables` shape in the #3348 comment.

Things I checked and found **good**, so they don't need re-litigating:

- `PointsAt` half-open `[StartRowId, EndRowId)` range membership.
- `RangeCell` deriving runs from SRM's `GetFields()`/`GetMethods()`/`GetParameters()` instead of hand-deriving from the next row — this delegates the "last row runs to end of table" rule to SRM and sidesteps the classic off-by-one.
- Blind spots recorded per row *after* the column loop, keyed on the **column's declared kind** rather than the cell, so a broken column never costs the caller that row's good edges.
- #3352's allocation control row (`eng/measure-metadata-projection-allocation.cs:73-81`, "PEReader only (no rows)") exists and is asserted-shaped rather than implied — that's the row that licenses attributing retained bytes to the projection.

**Test counts reproduce.** `ILInspector.Metadata.Tests` **1045/0** (5 skipped) vs your claimed 1044 — the +1 is base drift, not a discrepancy. `DotnetInspector.MetadataRendering.Tests` **95/0**, exact.

**CI is clean.** The non-green checks across #3345/#3348/#3350/#3352 are all `SKIPPED` (path-gated) or `CANCELLED` (superseded by a newer push) — no genuine failures. Flagging that because the rollup renders as `UNSTABLE` and reads worse than it is.

#3360 is docs-only and consistent with the code as it stands; if you take the `IsComplete` fix, its reverse-search paragraph needs the same wording change as the XML doc.

Evidence: Release build, SDK `11.0.100-preview.7.26366.102`. Probes were file-based apps under `/tmp/` referencing `src/ILInspector.Metadata`. No repository files modified.

richlander:
## Stack re-review at the current heads

Re-ran the stack after the `IsComplete` fix. Short version: **the fix is real and verified, but #3348 is not ready — the same overclaim is still live one granularity down.** Detail is in **#3348 comment 5101103779**; I won't duplicate it here.

**What I verified about the fix** (at `2cc43421`): the `visited`-set derivation closes the declaration/traversal drift; the mid-table truncation case is handled; and the `rid > reader.GetTableRowCount(...)` gate is **non-vacuous** — I tampered it to an unconditional `visited.Add` and `FindReferences_TableStoppedPartWayThrough_IsNotCountedAsSearched` fails. Reverted; baseline re-confirmed.

**Counts reproduce exactly**: `ILInspector.Metadata.Tests` **1006/0**, `DotnetInspector.MetadataRendering.Tests` **61/0**.

**`markdownlint` now run.** #3348's reconciliation flagged not running it as an unproven gap. Node is available in my environment, so I ran it — `docs/design/metadata-table-projection.md` exits **0**. Gap closed, and the reasoning behind the belief was correct.

**The residual, in one line:** signature blobs carry TypeDefOrRef tokens, and they sit in `Heap`-kind columns of tables the scan reads *in full*. So they're not blind rows, not unscanned tables, and not reported. On a two-type assembly that populates only modelled tables, `FindReferences(TypeDef:2)` returns `IsComplete=True`, `UnscannedTables=[]`, zero references — and `mdi` prints "No row points at TypeDef[2]." while a field is literally of that type. On this repo's own `ILInspector.Metadata.dll` that hides 1,872 edges across 207 types, including `MetadataProjectionOptions`, which is a parameter of `FindReferences` itself.

The fix I suggest is a doc/renderer change, not a scan change — #3348 should not grow blob decoding.

**Everything else in the stack still looks good** and I have nothing new on it: #3345's window, #3350's overview and heap addressing (I re-checked nothing changed materially), #3352's allocation control row, and #3360's docs. #3360's reverse-search paragraph will need the same wording adjustment as the XML doc if you take the scoping fix.

One process note: **#3348 carries a "Ready to merge" from 06:53Z.** I'd treat that as superseded until the claim is scoped — the mechanism is fine to ship, but the sentence "genuinely means nothing points at `Target`" is falsifiable at the current head.

Evidence: clean detached worktree at `2cc43421`, Release, SDK `11.0.100-preview.7.26366.102`. Tamper applied and reverted; tree clean before posting. No repository files modified.

PR #3348, opened 2026-07-27
Add reverse-reference search to the metadata projection

richlander:
Stacked on #3345 (`feature/3341-projection-row-window`). Review only the top commit; the base branch carries the row-window work.

## What

The projection's handle edges only run forward. A browser can follow `TypeDef[5].FieldList` down to a field, but cannot answer the two questions a reader actually asks: **"what declares this?"** and **"what points here?"**

`MetadataTableProjector.FindReferences(peReader, targetTable, targetRowId, maxReferences)` inverts them, returning a `MetadataRowReferenceSet`:

```text
$ mdi ILInspector.Metadata.dll --references Field:1

## References to Field[1] (2)

| Table | # | Column | Kind |
| ----- | - | ------ | ---- |
| TypeDef | 2 | FieldList | Range |
| CustomAttribute | 1 | Parent | Handle |
```

## Why the `Range` case is load-bearing

ECMA-335 gives an owned row **no back-pointer to its owner**. A `Field` is owned by whichever `TypeDef.FieldList` run covers it; a `Param` by whichever `MethodDef.ParamList` run covers it. Reverse search over list columns is therefore *how ownership resolves at all* — not an extra convenience layered on handle search. The example above is exactly that: `Field[1]` learns its declaring type only from the `Range` edge.

## Two deliberate design points that look like oversights

- **`options.Tables` is not honored.** Like `ProjectRow`, this is a query over the whole image, not a projection of a selection. A narrowed scan could report "nothing points here" while a pointer sat in an unsearched table — worse than a slower answer. (This exact ambiguity split two reviewers on #3345, so it is stated in the XML doc and the design doc rather than left to inference.)
- **Blind spots are reported, not folded in.** `Truncated` marks a scan the result budget stopped; `UnreadableRows` lists rows that could not be decoded and therefore could not be inspected. `IsComplete` is true only when neither happened — which is what makes an empty result trustworthy. Unlike `MetadataTableTruncation`, `Truncated` carries no total: a stopped scan never learns how many references it did not reach.

## Cost

The search reuses the projection's **single row-reading path** rather than a faster private one, so a `HandleRef`/`HandleRange` means the same thing in a search result as in a rendered table. Per-query cost is therefore proportional to the whole image. Whether that needs an index is a **measurement** question, filed with the allocation work (#3341 gap 3) rather than guessed at here.

## Naming

The public type is `MetadataRowReference`, not `MetadataReference`. The latter collides with `Microsoft.CodeAnalysis.MetadataReference` (CS0104) in `DecompilerHarness` and `ILInspector.Decompiler.Tests`, which reference both. Caught by the full-solution build, not by the Metadata project alone.

## Surface

- `AssemblyInspectionSession.MetadataReferences(table, rowId, maxReferences)` — plus a `MetadataTableRow` facet for `ProjectRow`, so the session's "callers never touch a `PEReader`" contract holds for both new primitives.
- `MetadataProjectionRenderer.Render(MetadataRowReferenceSet, …)` — Markdown names the target in a `##` heading; TSV/JSONL carry a leading `Target` column so rows self-identify.
- `mdi --references|-r Table:RowId [--max-references N]`. Markdown renders the caveats inline; the machine formats report them on **stderr**, since a pure row stream cannot distinguish a complete scan from a stopped one.

## Compatibility

Purely additive. No existing signature, model shape, or rendered output changes; the base branch's row-window behavior is untouched.

## Evidence

- `ILInspector.Metadata.Tests` — **994 / 0 failed / 0 skipped** (0 skipped means the `mdv` oracle actually ran). Includes a brute-force full-projection oracle that independently re-derives half-open run membership and is compared against the search across 8 targets.
- `DotnetInspector.MetadataRendering.Tests` — **58 / 0**.
- `mdi --references` verified end to end on `ILInspector.Metadata.dll` in `md`, `tsv`, and `jsonl`, including the truncated-scan stderr note.

Part of #3341.

### DIFF
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -436,11 +436,22 @@ oversights:
   could report "nothing points here" while a pointer sat in an unsearched table,
   which is worse than a slower answer.
 - **Blind spots are reported, not folded in.** `Truncated` marks a scan the
-  result budget stopped, and `UnreadableRows` lists rows that could not be
-  decoded and therefore could not be inspected. `IsComplete` is true only when
-  neither happened, which is what makes an empty result trustworthy. Unlike
-  `MetadataTableTruncation`, `Truncated` carries no total: a stopped scan never
-  learns how many references it did not reach.
+  result budget stopped, and `UnreadableRows` lists rows whose edges could not
+  be fully determined. `IsComplete` is true only when neither happened, which is
+  what makes an empty result trustworthy. Unlike `MetadataTableTruncation`,
+  `Truncated` carries no total: a stopped scan never learns how many references
+  it did not reach.
+
+`UnreadableRows` is subtler than "the row failed to read". The cell readers
+**contain** a decode failure as a `Malformed` cell rather than throwing, so a
+row holding a broken handle or list column reads back successfully and would
+otherwise pass as fully searched — a missed reference that reads as an absent
+one. A `Malformed` cell in an edge column therefore marks its row a blind spot.
+The **column's declared kind decides, not the cell**: a `Malformed` heap,
+scalar, or flags cell was never an edge and cannot hide a reference, so it is
+not counted. A row stays a blind spot only once, and after its good edges have
+been collected, so one broken column never costs the caller the edges that row
+does have.
 
 The search reuses the projection's single row-reading path rather than a faster
 private one, so a `HandleRef` or `HandleRange` means the same thing in a search
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -450,14 +450,43 @@ oversights:
   blind spot that under-reports is the whole failure this exists to prevent.
   Entering a table is deliberately not enough: a loop that stops part-way leaves
   rows unread, and an edge onto the target could sit in any of them. That also
-  makes the budget interaction fall out for free — a scan the budget stops
-  leaves both the table it stopped inside and every table after it unsearched,
-  and all of them are reported. Note that reaching the end of the row loop is
-  *also* not enough, because the budget is checked inside the **column** loop:
-  truncation on a table's final row leaves that row entered and abandoned
-  part-way through its columns while the row counter still passes the row count,
-  so a completed scan and a scan stopped on the last row are indistinguishable
-  from the counter alone. `Truncated` is what separates them.
+  makes the budget interaction mostly fall out for free — a scan the budget
+  stops leaves every table after it unsearched, and all of them are reported.
+
+  Reaching the end of the row loop is *also* not enough, because the budget is
+  checked inside the **column** loop: truncation on a table's final row leaves
+  that row entered and abandoned part-way through its columns while the row
+  counter still passes the row count, so the counter alone cannot tell a
+  completed scan from one stopped on the last row.
+
+  `Truncated` is too blunt to separate them, though, because of the boundary
+  case: if the budget trips on the **last column of the last row**, every cell
+  the table has was examined and the table genuinely was searched in full, even
+  though the scan ended inside it. Reporting it unscanned there would be a false
+  blind spot — it would tell the reader an unexamined cell could hide an edge
+  when no cell went unexamined. So the scan tracks the precise fact instead:
+  whether anything in this table was left unlooked-at, which is the columns
+  after the stop on that row, or any row after it.
+
+  That also keeps the malformed-cell blind spot sound. A stop before the last
+  column leaves the remaining columns unchecked for malformed edges, so the row
+  might belong in `UnreadableRows` without the scan knowing — but that is
+  exactly the case where the table is reported unscanned, so the gap is still
+  disclosed. When the stop is on the last column, every column was checked and
+  the row's status is fully determined.
+- **`UnscannedTables` covers unexamined cells, not just unread rows.** A table
+  lands there for three different reasons — never entered, entered and stopped
+  between rows, or entered and stopped between columns of its final row — and
+  only the first two leave a whole row unread. The caveat is therefore worded
+  around a cell the scan never examined, which is true of all three.
+- **A table with an unreadable row is not an `UnscannedTables` entry.** The two
+  blind spots partition the space rather than overlapping: `UnreadableRows` is
+  for rows the scan **read but could not decode**, `UnscannedTables` for cells it
+  **never examined**. A row whose read threw, or whose edge column decoded as
+  `Malformed`, is named individually in `UnreadableRows` and forces `IsComplete`
+  false, which is strictly more precise than implicating its whole table. Folding
+  it into `UnscannedTables` would also print a false statement, since every row
+  of that table was in fact read.
 - **Signature blobs are not searched, and that limit is not detectable.** The
   scan matches `Handle` and `HandleRange` columns. A `TypeDefOrRef` coded token
   spelled inside a signature blob lives in a `Heap` column, which is correctly
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -440,8 +440,8 @@ oversights:
   tables. A real assembly populates tables outside that subset — `NestedClass`,
   `MethodSemantics`, `InterfaceImpl`, `Property` and friends — so an edge living
   in one of them is invisible to the search. A nested type's declaring type is
-  exactly such an edge. `UnscannedTables` names the populated tables no row of
-  which was searched, so the gap is disclosed rather than answered as an
+  exactly such an edge. `UnscannedTables` names the populated tables the scan
+  did not read in full, so the gap is disclosed rather than answered as an
   absence. Empty tables are excluded: they cannot hide a reference.
 - **`UnscannedTables` is derived from the traversal, not declared.** The scan
   records a table only after examining every row the image says it has, and the
@@ -455,8 +455,8 @@ oversights:
   and all of them are reported.
 - **Blind spots are reported, not folded in.** `Truncated` marks a scan the
   result budget stopped, `UnreadableRows` lists rows whose edges could not be
-  fully determined, and `UnscannedTables` lists the populated tables no row of
-  which was searched. `IsComplete` is true only when none of the three happened,
+  fully determined, and `UnscannedTables` lists the populated tables the scan
+  did not read in full. `IsComplete` is true only when none of the three happened,
   which is what would make an empty result trustworthy — and today that means it
   is **false for essentially every real assembly**, because the third blind spot
   always fires. That is the honest reading: until the projection covers every
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -460,15 +460,28 @@ oversights:
   from the counter alone. `Truncated` is what separates them.
 - **Signature blobs are not searched, and that limit is not detectable.** The
   scan matches `Handle` and `HandleRange` columns. A `TypeDefOrRef` coded token
-  spelled inside a signature blob — `Field.Signature`, `MethodDef.Signature`,
-  `MemberRef.Signature`, `TypeSpec.Signature`, `StandAloneSig.Signature`,
-  `MethodSpec.Instantiation` — lives in a `Heap` column, which is correctly not
-  an edge column, on a row of a table the scan reads in full. So no blind spot
-  fires: the row is read, the table is searched, and the edge is simply not
-  looked for. Decoding blobs is future work (see the reverse-reference tagging
-  note above); until then the limit is disclosed **unconditionally** by the
-  renderer rather than by a per-scan signal, because there is no per-scan signal
-  to give.
+  spelled inside a signature blob lives in a `Heap` column, which is correctly
+  not an edge column, on a row of a table the scan reads in full. So no blind
+  spot fires: the row is read, the table is searched, and the edge is simply not
+  looked for. Among the tables modelled today those columns are
+  `Field.Signature`, `MethodDef.Signature`, `MemberRef.Signature`,
+  `TypeSpec.Signature`, `StandAloneSig.Signature` and
+  `MethodSpec.Instantiation`. Unmodelled tables hold signature blobs too —
+  `Property.Type` is one — but those tables are already disclosed by
+  `UnscannedTables`, so modelling one later moves its signature column into this
+  list rather than out of any disclosure. Decoding blobs is future work (see the
+  reverse-reference tagging note above); until then the limit is disclosed
+  **unconditionally** by the renderer rather than by a per-scan signal, because
+  there is no per-scan signal to give.
+
+  The caveat deliberately covers two unlike things. A signature blob spells a
+  reference as a TypeDefOrRef coded **token**, which is a genuine missed
+  row-to-row edge. A `CustomAttribute.Value` blob spells a `System.Type`
+  argument as a serialized type **name** (ECMA-335 II.23.3) — `[My(typeof(Alpha))]`
+  stores the bytes `0100 05 "Alpha" 0000`, with no token anywhere — so it is not
+  a row-to-row edge at all and is out of scope for a search defined over tokens.
+  A reader asking "what references this type?" is not served by that
+  distinction, though, so the caveat names both rather than resting on it.
 
   The risk runs backwards here, which is why the unconditional caveat is not
   redundant. `IsComplete` is true exactly when every populated table happens to
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -432,15 +432,28 @@ Two design points are deliberate and worth stating, because both look like
 oversights:
 
 - **`options.Tables` is not honored.** Like `ProjectRow`, the search is a query
-  over the whole image, not a projection of a selection. Narrowing the scan
-  could report "nothing points here" while a pointer sat in an unsearched table,
-  which is worse than a slower answer.
+  over the whole projection, not a projection of a selection. Narrowing the scan
+  per call could report "nothing points here" while a pointer sat in an
+  unsearched table, which is worse than a slower answer.
+- **The projection's table coverage is itself a blind spot.** The scan cannot
+  be wider than the projection, and the projection models a subset of ECMA-335's
+  tables. A real assembly populates tables outside that subset — `NestedClass`,
+  `MethodSemantics`, `InterfaceImpl`, `Property` and friends — so an edge living
+  in one of them is invisible to the search. A nested type's declaring type is
+  exactly such an edge. `UnscannedTables` names the populated unmodelled tables
+  so the gap is disclosed rather than answered as an absence. Empty tables are
+  excluded: they cannot hide a reference.
 - **Blind spots are reported, not folded in.** `Truncated` marks a scan the
-  result budget stopped, and `UnreadableRows` lists rows whose edges could not
-  be fully determined. `IsComplete` is true only when neither happened, which is
-  what makes an empty result trustworthy. Unlike `MetadataTableTruncation`,
-  `Truncated` carries no total: a stopped scan never learns how many references
-  it did not reach.
+  result budget stopped, `UnreadableRows` lists rows whose edges could not be
+  fully determined, and `UnscannedTables` lists the populated tables the scan
+  never visited. `IsComplete` is true only when none of the three happened,
+  which is what would make an empty result trustworthy — and today that means it
+  is **false for essentially every real assembly**, because the third blind spot
+  always fires. That is the honest reading: until the projection covers every
+  table, the search has not covered the whole image. Callers that only want to
+  know whether the scan itself finished should read `Truncated`. Unlike
+  `MetadataTableTruncation`, `Truncated` carries no total: a stopped scan never
+  learns how many references it did not reach.
 
 `UnreadableRows` is subtler than "the row failed to read". The cell readers
 **contain** a decode failure as a `Malformed` cell rather than throwing, so a
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -440,13 +440,21 @@ oversights:
   tables. A real assembly populates tables outside that subset — `NestedClass`,
   `MethodSemantics`, `InterfaceImpl`, `Property` and friends — so an edge living
   in one of them is invisible to the search. A nested type's declaring type is
-  exactly such an edge. `UnscannedTables` names the populated unmodelled tables
-  so the gap is disclosed rather than answered as an absence. Empty tables are
-  excluded: they cannot hide a reference.
+  exactly such an edge. `UnscannedTables` names the populated tables the scan
+  did not reach so the gap is disclosed rather than answered as an absence.
+  Empty tables are excluded: they cannot hide a reference.
+- **`UnscannedTables` is derived from the traversal, not declared.** The scan
+  records each table as it finishes it, and the blind spot is the populated
+  tables missing from that record. Computing it from the list of tables the
+  scan *intends* to visit would let the two drift, and a blind spot that
+  under-reports is the whole failure this exists to prevent. It also makes the
+  budget interaction fall out for free: a scan the budget stops leaves modelled
+  tables unreached, and those are unscanned in exactly the sense the report
+  means.
 - **Blind spots are reported, not folded in.** `Truncated` marks a scan the
   result budget stopped, `UnreadableRows` lists rows whose edges could not be
   fully determined, and `UnscannedTables` lists the populated tables the scan
-  never visited. `IsComplete` is true only when none of the three happened,
+  never reached. `IsComplete` is true only when none of the three happened,
   which is what would make an empty result trustworthy — and today that means it
   is **false for essentially every real assembly**, because the third blind spot
   always fires. That is the honest reading: until the projection covers every
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -345,7 +345,8 @@ inspector), a standalone tool that renders the tables the way `mdv` does:
   (`ToolCommandName=mdi`) whose System.CommandLine front-end maps flags onto
   `MetadataProjectionOptions` and delegates all output to the renderer below.
   Surface: `mdi <assembly> [--table|-t <Names>] [--format|-f md|tsv|jsonl]
-  [--max-rows|-n N] [--start-row|-s N] [--max-bytes N] [--max-chars N]`. Missing
+  [--max-rows|-n N] [--start-row|-s N] [--references|-r Table:RowId]
+  [--max-references N] [--max-bytes N] [--max-chars N]`. Missing
   files, native images, and unreadable metadata surface as visible errors, never
   success-shaped empty output.
 - **`src/DotnetInspector.MetadataRendering`** — a small reusable library holding
@@ -408,6 +409,49 @@ would read as the first three rows.
 Allocation shape — windowed or lazy materialization, and `ReadOnlySpan<T>` views
 over the current `ImmutableArray<T>` — remains open and **measurement-gated**;
 the browser's memory ceiling is the real constraint, and no benchmark exists yet.
+
+## Implemented: reverse references
+
+Forward navigation is only half of browsing. A reader looking at `Field[1]` asks
+"what declares this?", and looking at `TypeDef[5]` asks "what points here?".
+Neither question is answerable by following the projection's edges, because
+those edges only run one way.
+
+`MetadataTableProjector.FindReferences(peReader, targetTable, targetRowId,
+maxReferences)` inverts them, returning a `MetadataRowReferenceSet`: the
+`MetadataRowLocation` of every row pointing at the target, with the pointing
+column's index, its name, and whether the edge was a `Handle` or a `Range`.
+
+The `Range` case is the load-bearing one. ECMA-335 does not give an owned row a
+back-pointer to its owner: a `Field` is owned by whichever `TypeDef.FieldList`
+run covers it, and a `Param` by whichever `MethodDef.ParamList` run covers it.
+Reverse search over list columns is therefore how ownership is resolved at all —
+not an extra convenience on top of handle search.
+
+Two design points are deliberate and worth stating, because both look like
+oversights:
+
+- **`options.Tables` is not honored.** Like `ProjectRow`, the search is a query
+  over the whole image, not a projection of a selection. Narrowing the scan
+  could report "nothing points here" while a pointer sat in an unsearched table,
+  which is worse than a slower answer.
+- **Blind spots are reported, not folded in.** `Truncated` marks a scan the
+  result budget stopped, and `UnreadableRows` lists rows that could not be
+  decoded and therefore could not be inspected. `IsComplete` is true only when
+  neither happened, which is what makes an empty result trustworthy. Unlike
+  `MetadataTableTruncation`, `Truncated` carries no total: a stopped scan never
+  learns how many references it did not reach.
+
+The search reuses the projection's single row-reading path rather than a faster
+private one, so a `HandleRef` or `HandleRange` means the same thing in a search
+result as in a rendered table. Cost is therefore proportional to the whole image
+per query; whether that needs an index is a measurement question, filed with the
+allocation work above rather than guessed at here.
+
+The renderer and `mdi` carry the blind spots into every format:
+`mdi --references TypeDef:5` prints them inline under the Markdown table, and
+the TSV/JSONL streams report them on stderr, since a pure row stream cannot
+distinguish a complete scan from a stopped one.
 
 ## Layer placement
 
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -452,18 +452,50 @@ oversights:
   rows unread, and an edge onto the target could sit in any of them. That also
   makes the budget interaction fall out for free — a scan the budget stops
   leaves both the table it stopped inside and every table after it unsearched,
-  and all of them are reported.
+  and all of them are reported. Note that reaching the end of the row loop is
+  *also* not enough, because the budget is checked inside the **column** loop:
+  truncation on a table's final row leaves that row entered and abandoned
+  part-way through its columns while the row counter still passes the row count,
+  so a completed scan and a scan stopped on the last row are indistinguishable
+  from the counter alone. `Truncated` is what separates them.
+- **Signature blobs are not searched, and that limit is not detectable.** The
+  scan matches `Handle` and `HandleRange` columns. A `TypeDefOrRef` coded token
+  spelled inside a signature blob — `Field.Signature`, `MethodDef.Signature`,
+  `MemberRef.Signature`, `TypeSpec.Signature`, `StandAloneSig.Signature`,
+  `MethodSpec.Instantiation` — lives in a `Heap` column, which is correctly not
+  an edge column, on a row of a table the scan reads in full. So no blind spot
+  fires: the row is read, the table is searched, and the edge is simply not
+  looked for. Decoding blobs is future work (see the reverse-reference tagging
+  note above); until then the limit is disclosed **unconditionally** by the
+  renderer rather than by a per-scan signal, because there is no per-scan signal
+  to give.
+
+  The risk runs backwards here, which is why the unconditional caveat is not
+  redundant. `IsComplete` is true exactly when every populated table happens to
+  be modelled — that is, on small, simple assemblies, which are precisely the
+  images where a blob edge is the *only* remaining way to miss a reference. A
+  caveat conditioned on incompleteness would therefore go quiet exactly where it
+  is most needed.
 - **Blind spots are reported, not folded in.** `Truncated` marks a scan the
   result budget stopped, `UnreadableRows` lists rows whose edges could not be
-  fully determined, and `UnscannedTables` lists the populated tables the scan
-  did not read in full. `IsComplete` is true only when none of the three happened,
-  which is what would make an empty result trustworthy — and today that means it
-  is **false for essentially every real assembly**, because the third blind spot
-  always fires. That is the honest reading: until the projection covers every
-  table, the search has not covered the whole image. Callers that only want to
-  know whether the scan itself finished should read `Truncated`. Unlike
-  `MetadataTableTruncation`, `Truncated` carries no total: a stopped scan never
-  learns how many references it did not reach.
+  fully determined, `UnscannedTables` lists the populated tables the scan did not
+  read in full, and `TargetExists` marks a target row id past the end of its
+  table. `IsComplete` is true only when the scan hit none of the blind spots it
+  can detect — and today that means it is **false for essentially every real
+  assembly**, because the table-coverage blind spot always fires. That is the
+  honest reading: until the projection covers every table, the search has not
+  covered the whole image. `IsComplete` describes the scan, not the image: it
+  cannot account for the signature-blob limit above, which no scan can detect.
+  Callers that only want to know whether the scan itself finished should read
+  `Truncated`. Unlike `MetadataTableTruncation`, `Truncated` carries no total: a
+  stopped scan never learns how many references it did not reach.
+- **A row id past the end of its table is answered, not rejected.** Asking what
+  points at a row that does not exist is a well-formed question — a dangling
+  edge points at exactly the rows that are not there, so the search must stay
+  askable. What it must not do is answer "nothing points at this row", which
+  claims the image was searched and came back clean. `TargetExists` records the
+  distinction and takes `IsComplete` down with it, so an absent row reads as a
+  question that could not be answered rather than as an answer.
 
 `UnreadableRows` is subtler than "the row failed to read". The cell readers
 **contain** a decode failure as a `Malformed` cell rather than throwing, so a
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -440,21 +440,23 @@ oversights:
   tables. A real assembly populates tables outside that subset — `NestedClass`,
   `MethodSemantics`, `InterfaceImpl`, `Property` and friends — so an edge living
   in one of them is invisible to the search. A nested type's declaring type is
-  exactly such an edge. `UnscannedTables` names the populated tables the scan
-  did not reach so the gap is disclosed rather than answered as an absence.
-  Empty tables are excluded: they cannot hide a reference.
+  exactly such an edge. `UnscannedTables` names the populated tables no row of
+  which was searched, so the gap is disclosed rather than answered as an
+  absence. Empty tables are excluded: they cannot hide a reference.
 - **`UnscannedTables` is derived from the traversal, not declared.** The scan
-  records each table as it finishes it, and the blind spot is the populated
-  tables missing from that record. Computing it from the list of tables the
-  scan *intends* to visit would let the two drift, and a blind spot that
-  under-reports is the whole failure this exists to prevent. It also makes the
-  budget interaction fall out for free: a scan the budget stops leaves modelled
-  tables unreached, and those are unscanned in exactly the sense the report
-  means.
+  records a table only after examining every row the image says it has, and the
+  blind spot is the populated tables missing from that record. Computing it from
+  the list of tables the scan *intends* to visit would let the two drift, and a
+  blind spot that under-reports is the whole failure this exists to prevent.
+  Entering a table is deliberately not enough: a loop that stops part-way leaves
+  rows unread, and an edge onto the target could sit in any of them. That also
+  makes the budget interaction fall out for free — a scan the budget stops
+  leaves both the table it stopped inside and every table after it unsearched,
+  and all of them are reported.
 - **Blind spots are reported, not folded in.** `Truncated` marks a scan the
   result budget stopped, `UnreadableRows` lists rows whose edges could not be
-  fully determined, and `UnscannedTables` lists the populated tables the scan
-  never reached. `IsComplete` is true only when none of the three happened,
+  fully determined, and `UnscannedTables` lists the populated tables no row of
+  which was searched. `IsComplete` is true only when none of the three happened,
   which is what would make an empty result trustworthy — and today that means it
   is **false for essentially every real assembly**, because the third blind spot
   always fires. That is the honest reading: until the projection covers every
--- a/src/DotnetInspector.MetadataRendering/MetadataProjectionRenderer.cs
+++ b/src/DotnetInspector.MetadataRendering/MetadataProjectionRenderer.cs
@@ -95,10 +95,11 @@ public static class MetadataProjectionRenderer
     /// Renders a reverse-reference search — the rows pointing at one row — to
     /// <paramref name="output"/>.
     ///
-    /// A reverse search has two blind spots the table renderer has no equivalent
-    /// of, and neither may be dropped: a budget that stopped the scan, and rows
-    /// that could not be decoded. Both are rendered as explicit caveats, so an
-    /// empty result is never mistaken for a confident "nothing points here".
+    /// A reverse search has three blind spots the table renderer has no
+    /// equivalent of, and none may be dropped: a budget that stopped the scan,
+    /// rows that could not be decoded, and populated tables no row of which was
+    /// searched. All are rendered as explicit caveats, so an empty result is
+    /// never mistaken for a confident "nothing points here".
     /// </summary>
     public static void Render(
         MetadataRowReferenceSet references,
--- a/src/DotnetInspector.MetadataRendering/MetadataProjectionRenderer.cs
+++ b/src/DotnetInspector.MetadataRendering/MetadataProjectionRenderer.cs
@@ -207,7 +207,11 @@ public static class MetadataProjectionRenderer
         {
             int count = references.UnscannedTables.Length;
             var names = string.Join(", ", references.UnscannedTables);
-            yield return $"{count} populated {(count == 1 ? "table is" : "tables are")} not modelled by the projection and {(count == 1 ? "was" : "were")} not searched, so an edge from {(count == 1 ? "it" : "them")} would have been missed: {names}.";
+            // Deliberately does not say *why* a table went unsearched. A table is
+            // here either because the projection does not model it or because the
+            // budget stopped the scan first, and naming one cause would be a
+            // false claim about the other.
+            yield return $"{count} populated {(count == 1 ? "table was" : "tables were")} not searched, so an edge from {(count == 1 ? "it" : "them")} would have been missed: {names}.";
         }
     }
 
--- a/src/DotnetInspector.MetadataRendering/MetadataProjectionRenderer.cs
+++ b/src/DotnetInspector.MetadataRendering/MetadataProjectionRenderer.cs
@@ -91,6 +91,120 @@ public static class MetadataProjectionRenderer
         writer.Flush();
     }
 
+    /// <summary>
+    /// Renders a reverse-reference search — the rows pointing at one row — to
+    /// <paramref name="output"/>.
+    ///
+    /// A reverse search has two blind spots the table renderer has no equivalent
+    /// of, and neither may be dropped: a budget that stopped the scan, and rows
+    /// that could not be decoded. Both are rendered as explicit caveats, so an
+    /// empty result is never mistaken for a confident "nothing points here".
+    /// </summary>
+    public static void Render(
+        MetadataRowReferenceSet references,
+        TextWriter output,
+        MetadataTableFormat format = MetadataTableFormat.Markdown)
+    {
+        ArgumentNullException.ThrowIfNull(references);
+        ArgumentNullException.ThrowIfNull(output);
+
+        if (format == MetadataTableFormat.Markdown)
+            RenderReferencesMarkdown(references, output);
+        else
+            RenderReferencesTabular(references, output, format);
+    }
+
+    static void RenderReferencesMarkdown(MetadataRowReferenceSet references, TextWriter output)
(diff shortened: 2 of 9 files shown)


richlander:

Two-model review at the fixed head, per AGENTS.md (subtle-correctness risk, and not my own family): **MAI-Code** and **Gemini 3.1 Pro**, each in its own isolated read-only worktree, each given the identical 12-claim prompt with exact base/head, design intent, and required real-run evidence.

### Round 1 (head `14e3f31f`) — the reviewers disagreed, and Gemini was right

**MAI-Code: clean**, all 12 claims verified.

**Gemini: one CRITICAL blocking finding**, which I reproduced and confirmed:

> `FindReferences` could return a success-shaped empty answer over a row it had not actually read.

The mechanism is worth recording, because it defeats the obvious defence. The projection's cell readers *contain* decode failures rather than throwing: `HandleCell` catches `BadImageFormatException`/`ArgumentException` from handle resolution and returns `MetadataValue.Malformed`, and the `RangeCell` overloads do the same for list columns. So `spec.ReadRow` **succeeds**. The outer `try/catch` around `ReadRow` never fires, `PointsAt` returns `false` on the malformed cell, and the row is never recorded in `UnreadableRows` — with `IsComplete` still `true`. A reference the search could not account for read back as an absent one, which is exactly the failure-hidden-as-success shape the repo forbids.

MAI-Code verified this same claim (5, "blind spots are preserved") as sound. Its verification only covered the outer `try/catch`, not the inner cell-level catches. Calling that out explicitly: a single clean review is not proof, and the second model earned its keep here.

### The fix — `82776eea`

A `Malformed` cell marks its row a blind spot **iff the column's declared `MetadataColumnKind` is `Handle` or `HandleRange`**. Gating on the column's declared kind rather than on the cell keeps the report honest in both directions: a malformed heap, scalar, or flags cell was never an edge and cannot hide a reference, so counting it would over-report.

Before fixing I measured how often this actually fires: I projected four production assemblies in full and found **zero** real malformed cells. (An apparent 25 hits were the literal type name `Malformed` in our own assembly's string heap.) Real compilers do not emit these, so the honest fix costs nothing in practice — but the search must still not lie when handed a corrupt or hand-crafted image.

**One deliberate divergence from the reviewer's suggested fix.** Gemini suggested breaking out of the column loop on the first malformed edge. I did not take that: a row can carry a broken `Extends` *and* a valid `MethodList` run that genuinely points at the target, so breaking early would silently drop a real reference — trading one honesty bug for another. Instead a per-row `blind` flag is recorded once, after the column loop completes, so a broken column never costs the caller that row's good edges. Gemini explicitly endorsed this reasoning on re-review.

Evidence is a synthetic fixture, because no real compiler produces the input. `BuildImageWithMalformedCells()` uses `MetadataBuilder`/`ManagedPEBuilder` with `suppressValidation: true` (the pattern already established in `TypeRefDecoderRecursionTests.cs`): `TypeDef[2].Extends` names a nonexistent `TypeRef` row (Handle-column failure) and `Module[1].Mvid` is a `GuidHandle` past the end of the GUID heap (Heap-column failure — the negative case that catches over-reporting). A canary test asserts malformed cells appear *only* where the fixture intends, so the other cases cannot go vacuous if SRM changes behaviour.

I proved the tests load-bearing in both directions by temporarily mutating the product: removing the check fails 2 tests; widening it to *every* `Malformed` cell regardless of column kind fails 3.

### `a1ea72c4` — pinning the invariant the gate rests on

The gate is only sound if edge cells never appear in non-edge columns. Rather than take either reviewer's manual walk of the 17 table specs on trust, this commit makes the invariant a permanent test: `EdgeCells_OnlyEverAppearInEdgeColumns` checks every projected cell of a real assembly (every `Handle` cell in a `Handle` column, every `Range` cell in a `HandleRange` column) with a `> 10_000` cell floor so the loop cannot pass vacuously, and `BothEdgeColumnKinds_OccurInRealMetadata` ensures both kinds are actually exercised. A future table spec that violates the invariant now fails a test instead of quietly reopening the bug. Test-only; no product change.

### Round 2 (fixed head) — both clean

**Gemini: clean.** Finding resolved; no new blocking issue; endorsed the `blind`-flag-over-`break` decision, verified no double-add and correct interaction with the `truncated` early-break; walked all 17 table specs and found no hole in the gate.

**MAI-Code: clean**, and this round it demonstrated rather than read intent — it wrote a standalone probe against the built `ILInspector.Metadata.dll` reproducing the malformed-edge fixture and observed `IsComplete=False`, `References=0`, `UnreadableRows=TypeDef[2]`. It independently found no column able to produce an edge cell while declared with a non-edge kind.

MAI reviewed at `82776eea` (999 tests); `a1ea72c4` lands after it and is test-only additions, which is why its count differs from the 1001 below.

### Validation at head `a1ea72c4`

- `dotnet build dotnet-inspect.slnx -c Release` — clean. (Required, not optional: the `MetadataReference*` → `MetadataRowReference*` rename in this PR exists because the original name collides with `Microsoft.CodeAnalysis.MetadataReference` in `tools/DecompilerHarness` and `ILInspector.Decompiler.Tests`. Only the full-solution build surfaces that CS0104.)
- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release` — **1001 passed, 0 failed, 0 skipped**. Zero skipped matters: it means the mdv cross-implementation oracle actually ran.
- `dotnet run --project tests/DotnetInspector.MetadataRendering.Tests -c Release` — **58 passed, 0 failed**.
- `mdi --references` verified end to end on `ILInspector.Metadata.dll` in md, tsv, and jsonl, including the truncated-scan stderr note.

Note on CI: this PR is stacked on #3345, and `ci.yml` triggers only on pull requests targeting `main`, so GitHub reports no checks here by design. The full-solution build and both test suites above were run locally at the exact head. CI will run once #3345 merges and this PR retargets `main`.

richlander:
Ready to merge

Both fixed-head reviews are clean, full-solution build and both test suites pass at `a1ea72c4` (Metadata 1001/0/0 with 0 skipped, MetadataRendering 58/0). Merges after its base #3345.

richlander:
## Review: reverse-reference search (`#3348`), reviewed on the `#3352` head `2e8545f1`

Reviewed the whole stack (3345 → 3348 → 3350 → 3352 → 3360) in a clean detached worktree at `2e8545f1`, which carries all the product code. Build clean, `ILInspector.Metadata.Tests` **1045/0** (5 skipped) and `DotnetInspector.MetadataRendering.Tests` **95/0** — matching your claims (+1 on metadata from base drift).

Most of this is good work, and I want to be specific about why before the finding, because the finding is narrow and the surrounding care is what makes it worth fixing rather than shrugging at:

- `PointsAt` uses **half-open** `[StartRowId, EndRowId)` membership, which is right.
- `RangeCell` derives runs from SRM's own `typeDef.GetFields()`/`GetMethods()`/`GetParameters()` rather than hand-deriving them from the next row's list column. That delegates ECMA-335's "the last row's run extends to the end of the table" rule to SRM and sidesteps the classic off-by-one that this feature exists to get right.
- Blind spots are recorded **per row, after the column loop**, so a row with one broken edge still contributes its good edges. And the **column's declared kind** decides, not the cell — a `Malformed` heap/scalar/flags cell was never an edge and correctly is not counted.

### Finding: `IsComplete` asserts a property the search does not have

`MetadataRowReferences.cs:147`

```csharp
public bool IsComplete => !Truncated && UnreadableRows.IsEmpty;
```

and its doc, `:143`:

> Whether the search covered **the whole image**: it ran to completion and every row was readable, so an empty `References` **genuinely means nothing points at** `Target`.

Neither term knows about tables the projector does not model. `FindReferences` scans `SupportedTables` (`MetadataTableProjector.cs:57`) — **17 tables** out of ECMA-335's ~38. Every unscanned table that points *at* a scanned row is invisible, and `IsComplete` still reports `true`.

On this repo's own `ILInspector.Metadata.dll`:

```
tables scanned by FindReferences: 17
POPULATED but NEVER SCANNED:
   InterfaceImpl                 251 rows
   ClassLayout                     3 rows
   PropertyMap                   185 rows
   Property                     1075 rows
   MethodSemantics              1782 rows
   FieldRva                       11 rows
   NestedClass                   157 rows

rows scanned: 23655   rows never scanned: 3464  (12.8% of the image)
```

Two concrete misses, both on edges a reader would consider the *primary* answer to "what points here?":

```
Target: TypeDef[225] 'Enumerator'
  declaring type (ground truth, from NestedClass table): TypeDef[4]
  IsComplete = True  Truncated=False  UnreadableRows=0
  references reported: 6
     MethodImpl[95..99].Class (Handle)
     GenericParam[4].Owner (Handle)
  -> the NestedClass edge is absent, yet IsComplete claims the whole image was searched.

[BLIND] MethodDef[3] 'System.Collections.ICollection.get_Count' is the getter of Property[1]
   IsComplete=True  refs=2  MethodSemantics edge reported? False
```

**Positive control** — the mechanism works correctly where it does scan, so this is a scope gap, not a logic bug:

```
[CONTROL] Field[1] IsComplete=True refs=2
     TypeDef[2].FieldList (Range)     <- the owning-TypeDef Range edge, found
     CustomAttribute[1].Parent (Handle)
```

This reaches the user unqualified. `mdi` prints a bare count:

```
## References to TypeDef[225] (6)
```

A nested type's *declaring type* is missing from "what points here", and nothing on screen says 7 populated tables were skipped.

### Why I think this is worth fixing in the stack rather than filing

Because the stack has already diagnosed it — twice, in the same document.

`docs/design/metadata-table-projection.md:437` states the design rule:

> **`options.Tables` is not honored.** … Narrowing the scan could report "nothing points here" while a pointer sat in an unsearched table, which is worse than a slower answer.

That reasoning is exactly right, and then the scan is narrowed anyway — not by `options.Tables`, but by `SupportedTables`. The user-visible consequence is identical.

Then `:485`, ~50 lines later, in the overview section:

> a table with rows the projection does not model — **`NestedClass`, `MethodSemantics`, `Property`** on a typical assembly — is visible as a coverage gap. Listing only the projected tables would make an unmodelled table indistinguishable from an empty one, which is **the same success-shaped-absence failure the reverse search avoids**.

It names the exact three tables my probe caught, argues the principle correctly, builds `MetadataTableSummary.IsProjected` (#3350) to surface it for the overview — and then credits the reverse search with avoiding the failure it is committing. The two halves of the doc disagree, and the running code agrees with the pessimistic half.

`AGENTS.md` phrases the same rule repo-wide: *"Do not turn decode, acquisition, analysis, or rendering failures into success-shaped empty output."*

### Suggested fix (cheap, because #3350 already built the data)

Not "scan all 38 tables" — that's a real feature and belongs in its own change. Just make the completeness flag stop overclaiming:

```csharp
// populated tables the scan does not model; no scan required, just row counts
public ImmutableArray<TableIndex> UnscannedTables { get; }

public bool IsComplete => !Truncated && UnreadableRows.IsEmpty && UnscannedTables.IsEmpty;
```

`UnscannedTables` is ~21 `reader.GetTableRowCount(t)` lookups for tables not in `ProjectedTables` — no extra scanning, and #3350 already added the surface that computes it. Carry it into the renderer beside the existing blind-spot reporting.

`IsComplete` would then be `false` on essentially every real assembly. **That is the correct answer**, and it is the same answer the overview section already insists on: the gap should be *visible*, not folded into a success shape. If you'd rather not have a flag that is always false, the honest alternative is to keep it but rename it to what it actually measures (`ScanRanToCompletion`) and drop the "whole image" / "genuinely means nothing points at" sentence from the doc — the property that sentence promises is the one that isn't there.

### Smaller notes (non-blocking)

1. **#3352's measurement is narrated, not gated.** The control row is present and is the right control (`eng/measure-metadata-projection-allocation.cs:73-81`, "PEReader only (no rows)") — that's what licenses attributing retained bytes to the projection, and I'm glad it's there rather than implied. But it lives in an `eng/` script with no asserted bound, so the numbers in the PR body are a report, not a regression gate. Fine for a measurement PR; worth saying out loud so nobody later reads the doc's figures as enforced.
2. **`Truncated` carrying no total is correctly documented** (`:445`) and I agree with the call — a stopped scan genuinely cannot know what it didn't reach. Noting it only so it's clear I read it as deliberate rather than missed.

---

Evidence: clean detached worktree at `2e8545f1`; SDK `11.0.100-preview.7.26366.102`; Release. Probes were file-based apps under `/tmp/` against `src/ILInspector.Metadata`, run on the build's own `ILInspector.Metadata.dll`. No repository files modified (`git status --porcelain` empty).

richlander:
You were right, and the finding is worse than the doc wording — it's a real missed edge.

## Reproduced

I built a probe against the PR head and confirmed it end to end:

```
tables scanned by FindReferences: 17
POPULATED but NEVER SCANNED:
  InterfaceImpl 246, ClassLayout 3, PropertyMap 180, Property 1042,
  MethodSemantics 1749, FieldRva 11, NestedClass 156
rows scanned: 23141   never scanned: 3387   (12.8% of the image)

MISS: TypeDef[217] 'Enumerator' — its declaring type is recorded only in
      NestedClass. IsComplete=True, 6 references reported, the NestedClass
      edge absent.
CONTROL: Field[1] IsComplete=True refs=2 — TypeDef[2].FieldList (Range) found.
```

The control is the important line: `Field` → declaring `TypeDef` works, because `TypeDef` is modelled. So this is a **scope gap, not a logic bug** — the search is correct over what it scans, and what it scans is 17 of ECMA-335's tables.

The user-visible consequence: asking what points at a nested type returned a confident answer that omitted its own parent.

## Fix

Took your primary suggestion rather than the `ScanRanToCompletion` fallback.

`MetadataRowReferenceSet` gains a third blind spot, `UnscannedTables` — the **populated** tables the scan never visited — and `IsComplete` now requires it to be empty. Empty tables are excluded deliberately: a table with no rows cannot hide a reference, so listing it would overstate the gap. Learning this costs row counts only, no extra scanning.

`IsComplete` is now false for essentially every real assembly. That reads alarming but it is the honest answer: until the projection covers every table, the search has not covered the whole image. Callers that only want *"did the scan finish"* should read `Truncated`, and that distinction is now stated on the property.

Four existing tests were asserting `IsComplete` when they meant *"the scan ran clean"*. They now say that directly via an `AssertScanRanClean` helper, so they stop borrowing a stronger claim than they test.

The renderer carries the caveat into all three formats:

```
## References to TypeDef[1] (0)

No row points at TypeDef[1].

7 populated tables are not modelled by the projection and were not searched,
so an edge from them would have been missed: InterfaceImpl, ClassLayout,
PropertyMap, Property, MethodSemantics, FieldRva, NestedClass.
```

TSV/JSONL emit the same note on stderr, verified separately.

## Design doc

You cited the self-contradiction and it was real. `:437` stated the rule — *"Narrowing the scan could report 'nothing points here' while a pointer sat in an unsearched table"* — and the surrounding text then credited the reverse search with avoiding exactly the failure it commits. Corrected in this commit rather than left to disagree with the code, and it now names the gap and why `IsComplete` reads false.

## Also removed

While wiring this up I wrote a `try` / `catch (ArgumentOutOfRangeException)` around `GetTableRowCount` — the same catch-and-continue shape Gemini flagged on #3350. It's unreachable: the max defined `TableIndex` is 55 and `TableCount` is 64, so every defined value is in range. Verified, then deleted rather than kept as decoration that hides a table.

## Evidence

- `ILInspector.Metadata.Tests` — 1004 / 0 failed / 0 skipped
- `DotnetInspector.MetadataRendering.Tests` — 60 / 0 failed
- `mdi --references` verified end to end in md, tsv, and jsonl

Mutation-checked in both directions, since a disclosure test that cannot fail is worthless:

- returning no unscanned tables → 3 new tests fail
- including empty tables → the exactness test fails

Head is `2c873be8`. Since this changes a correctness contract, I'll re-run adversarial review on the fixed head before calling it ready again.

richlander:
## Reconciliation — your `IsComplete` finding, and what re-review turned it into

You were right, and the fix went three layers deeper than the report. Head is now `331bddd3`, five commits on top of the original three.

### Your finding, reproduced

`FindReferences` scans `SupportedTables` (17 entries), but `IsComplete` was `!Truncated && UnreadableRows.IsEmpty` and its doc promised whole-image coverage. On `ILInspector.Metadata.dll`, **7 populated tables were never scanned** — `InterfaceImpl` (246), `ClassLayout` (3), `PropertyMap` (180), `Property` (1042), `MethodSemantics` (1749), `FieldRva` (11), `NestedClass` (156) — **3,387 rows, 12.8% of the image**. A nested type's declaring type lives *only* in `NestedClass`, so that edge was unreachable while the API reported a complete search.

A control case confirmed it was a scope gap and not a logic bug: `Field[1]` resolves its declaring `TypeDef` correctly via a `FieldList` range run.

### How re-review escalated it

| Commit | What it fixed | Found by |
| --- | --- | --- |
| `2c873be8` | The **claim** was wrong → `UnscannedTables`, and `IsComplete` requires it empty | you |
| `7fc1eebf` | The claim's **source** was wrong → derive from traversal, not a declaration | GPT-5.6 Sol |
| `53b11a8b` | "Traversal" was wrong → **entry is not completion** | GPT-5.6 Sol |
| `fe794992` | The **wording** was then wrong → "not searched **in full**" | GPT-5.6 Sol |
| `331bddd3` | Two stale doc sites the sweep missed | GPT-5.6 Sol |

**`7fc1eebf` — the test bug was hiding a product bug.** Both reviewers independently found the exactness test tautological: it derived its expectation from `FullProjection`, which shares `SupportedTables` with the product. GPT then showed the product had the same defect — `CollectUnscannedTables` read `SupportedTables` (a *declaration of intent*) while the scan loop iterated it separately, so the two could drift silently. `UnscannedTables` is now derived from a `visited` set recorded during the actual scan.

That also exposed a bug neither of us had named: on truncation the loop `break`s, so modelled tables past the stop point were never visited yet were absent from `UnscannedTables`.

**`53b11a8b` — the subtle one.** GPT found `visited` recorded table *entry*, not completion, so a row loop stopping part-way still marked the table searched. The gate is now:

```csharp
int rowCount = reader.GetTableRowCount(spec.Index);
int rid = 1;
for (; rid <= rowCount && !truncated; rid++) { ... }
if (rid > reader.GetTableRowCount(spec.Index))
    visited.Add(spec.Index);
```

My first attempt compared `rid > rowCount` against the loop's local, and GPT's mutation **still slipped through** — capping the local also caps the definition of "done". Re-reading from the reader is what actually closes it.

Exit paths: normal completion → `rowCount+1` → visited. Truncation on the *final* row → also `rowCount+1` → visited, correctly (every row was examined, results merely capped). Truncation mid-table → not visited. Zero-row table → visited, and filtered out anyway since empty tables cannot hide a reference.

### Observable behaviour

```
$ mdi ILInspector.Metadata.dll --references TypeDef:1
7 populated tables were not searched in full, so an edge in a row the scan never
read would have been missed: InterfaceImpl, ClassLayout, PropertyMap, Property,
MethodSemantics, FieldRva, NestedClass.

$ mdi ILInspector.Metadata.dll --references TypeDef:2 --max-references 1
14 populated tables were not searched in full, ...: InterfaceImpl, ClassLayout,
PropertyMap, Property, MethodSemantics, MethodImpl, TypeSpec, FieldRva, Assembly,
AssemblyRef, ExportedType, NestedClass, GenericParam, MethodSpec.
```

The truncated scan names **14**, not the 13 it named before `53b11a8b`. The addition is `MethodImpl` — the table the budget stopped *inside*.

`IsComplete` now carries a `<remarks>` stating plainly that it is **false for essentially every real assembly**, because the coverage blind spot always fires. That is the honest reading until the projection covers all of ECMA-335.

The caveat deliberately **asserts no cause**. A table is listed either because it is unmodelled or because the budget stopped the scan, and naming one would be a false claim about the other. `Markdown_UnscannedTables_DoNotClaimACause` pins both cause and extent.

### One finding dismissed, with the reason

MAI-Code raised a BLOCKING claim that `MethodDef` was over-reported on a clean tree. **Not reproducible — it was a stale-artifact artefact of its own mutation.** Its source contained `MethodDef` and `git status` was empty, but the compiled DLL was 10 minutes *newer* than the restored source file: its earlier `SupportedTables` mutation had been restored with a timestamp older than the output built from it, so MSBuild treated every later build as a no-op. Touching the file and rebuilding gave 1005/0/0 and 7 tables. Its own later runs, done correctly, confirm the mutation fails the intended test.

### Mutations reproduced as failing

| Mutation | Caught by |
| --- | --- |
| Return `[]` for unscanned | 3 tests |
| Include empty tables (`>= 0`) | exactness test |
| Drop `MethodDef` from `SupportedTables` | exactness test |
| Scan-level `continue` skipping `TypeDef` | exactness test |
| `rowCount = Math.Min(rowCount, 1)` for `MethodSpec` | exactness test |
| Unconditional `visited.Add` | part-way-through test |
| Session returns wrong `[FieldPtr]` | facet test |
| Revert "in full" wording | caveat test |
| Drop a table from the test's `ModelledTables` | guard assertion |

### Acknowledged boundary, non-blocking

The invariant guarantees rows were *visited*, not that each visited row was correctly *examined*. A deliberate per-row `continue` still passes. Both reviewers and I agree that is ordinary scanner correctness, not a blind-spot claim, and out of scope here.

### Validation at `331bddd3`

- `dotnet build dotnet-inspect.slnx -c Release` — 0 errors
- `ILInspector.Metadata.Tests` — **1006/0/0**
- `DotnetInspector.MetadataRendering.Tests` — **61/0**
- `mdi --references` in Markdown, TSV and JSONL, full and truncated; JSONL stdout stays machine-pure with caveats on stderr
- Both reviewers independently confirmed the above from a **forced clean rebuild** after deleting `artifacts/`

One gap I will not paper over: **`markdownlint` was not run** on the `docs/design/metadata-table-projection.md` change, because Node is not installed in this environment. The change is prose-only substitution inside existing bullets with no block-structure change, and `MD013` is disabled in `.markdownlint.yaml`, so I believe the risk is nil — but I have not proven it.

Both fixed-head reviews are clean at `331bddd3`.

richlander:
**Ready to merge**

Head `331bddd3`. Both fixed-head adversarial reviews (MAI-Code and GPT-5.6 Sol) are clean, each confirmed from a forced clean rebuild after deleting `artifacts/`.

Merge-blocking validation:

- `dotnet build dotnet-inspect.slnx -c Release` — 0 errors
- `ILInspector.Metadata.Tests` — 1006/0/0
- `DotnetInspector.MetadataRendering.Tests` — 61/0
- `mdi --references` verified in Markdown, TSV and JSONL, full and truncated; JSONL stdout machine-pure

On the markdownlint gap I flagged above — I closed it as far as is possible without Node. CI's `markdownlint` leg only runs on PRs targeting `main`, so it will not run on a stacked PR, but it will before this reaches main. In the meantime I verified the doc change directly: every changed line is a prose continuation inside an existing list item, with no heading, fence, table or list-marker change. `MD013` is disabled in `.markdownlint.yaml`, and the file has zero trailing-space lines, zero hard tabs, zero consecutive blank lines, and a single trailing newline. Nothing a prose substitution can break is broken.

Non-blocking follow-up, tracked separately and deliberately not in this PR: the projection covers 17 of ECMA-335's tables, so `IsComplete` will stay false for real assemblies until coverage grows. This PR makes that gap **visible and named** rather than silently answered as an absence — closing it is the separate piece of work.

richlander:
## Re-review at `2cc43421`

The fix landed and it is a good one. I verified it rather than taking the reconciliation's word for it, and then kept pulling on the same thread — which turned up **the same overclaim one granularity down**, still live at this head.

### What I confirmed about the fix

- **The `visited`-set derivation is right.** Deriving the blind spot from the traversal instead of `SupportedTables` closes the drift GPT found, and the mid-table `break` case is handled: the outer loop no longer swallows tables past the stop point.
- **The `rid > reader.GetTableRowCount(spec.Index)` gate is non-vacuous.** I tampered it to an unconditional `visited.Add(spec.Index)`; `FindReferences_TableStoppedPartWayThrough_IsNotCountedAsSearched` fails. Reverted, baseline re-confirmed at 1006/0.
- **Counts reproduce exactly**: `ILInspector.Metadata.Tests` **1006/0**, `DotnetInspector.MetadataRendering.Tests` **61/0**. (Metadata reports 5 *skipped*, not 0 — trivial, but your evidence block says 0.)
- **`markdownlint` passes.** You flagged not running it as a gap you would not paper over; Node is available here, so I ran it — `npx markdownlint-cli docs/design/metadata-table-projection.md` exits **0**. Your reasoning was right and the gap is closed.

I also checked the cheaper version of my original finding before looking further: whether a *modelled* table omits an edge **column**. It does not. `TypeDef.Extends`, `TypeRef.ResolutionScope`, `MemberRef.Class`, `Constant.Parent`, `CustomAttribute.Parent`/`Type`, `MethodImpl.Class`/`MethodBody`/`MethodDeclaration`, `ExportedType.Implementation`, `GenericParam.Owner`, `MethodSpec.Method`, and all three list columns are present. Within its chosen granularity the model is complete. Recording that as a negative result so nobody re-runs it.

### Finding: the overclaim survived, below table granularity

`MetadataRowReferences.cs:168` still promises the absolute:

> Whether the search covered the whole image: it ran to completion, every row was readable, and no populated table went unscanned, so an empty `References` **genuinely means nothing points at** `Target`.

`UnscannedTables` makes that true with respect to *tables*. But an edge does not have to live in a `Handle` column. Signature blobs carry TypeDefOrRef coded tokens, and every one of them sits in a **`Heap`-kind** column of a table the scan reads in full — `Field.Signature`, `MethodDef.Signature`, `MemberRef.Signature`, `TypeSpec.Signature`, `StandAloneSig.Signature`, `MethodSpec.Instantiation`. `PointsAt` never sees them, the row is not blind (a `Heap` cell is correctly not an edge column), the table *is* fully visited, so it is correctly absent from `UnscannedTables` — and the reference is silently gone.

**This is demonstrable in the case the PR designates as trustworthy.** I built a two-type assembly that populates only modelled tables:

```csharp
public class Alpha { }
public class Beta { public Alpha F; }
```

```
Field[1] 'F' has type TypeDef[2] 'Alpha'
  FindReferences(TypeDef:2): IsComplete=True  refs=0  UnscannedTables=[]
```

`IsComplete` is **true**. Per its own doc that means an empty result "genuinely means nothing points at" `Alpha`. A field is literally of that type. And `mdi` says so without qualification — no caveat line, because there are no unscanned tables to name:

```
## References to TypeDef[2] (0)

No row points at TypeDef[2].
```

That statement is false.

**Scale, on this repo's own `ILInspector.Metadata.dll`** (counting only `Field` and `MethodDef` signatures — not `MemberRef`, `TypeSpec`, `StandAloneSig`, `MethodSpec`, `Constant.Value`, or `CustomAttribute.Value`):

```
TypeDef rows referenced from Field/MethodDef signature blobs: 207
total such blob edges: 1872
of those 207 TypeDefs: 207 report NO Field/MethodDef edge
                        22 report zero references of any kind
```

The example that makes the point best:

```
TypeDef[137] 'MetadataProjectionOptions'
  mdi would say: 'No row points at TypeDef[137].'   (IsComplete=False, unscanned=7)
  but 32 signature-blob uses exist, e.g. MethodDef[1021] 'MetadataTables' param
```

`MetadataProjectionOptions` is a parameter of `Project`, `MetadataTables`, and `FindReferences` **itself**. The reverse search cannot find a single reference to its own options type, and on a real image the only thing standing between the user and "No row points at" is the incidental fact that some unmodelled table happened to be populated.

That inversion is the part I'd weigh most: **the disclosure is strongest exactly where it is least needed.** A big, messy assembly gets a caveat for unrelated reasons. A small, clean one — the kind someone reaches for when they want a trustworthy answer — gets `IsComplete=True` and a flat false statement.

### Suggested fix — scope the claim, don't grow the scan

I am explicitly *not* suggesting you decode signature blobs here. That is a real feature (the doc already anticipates it at `:315`, "reverse-reference tagging (walk blob-typed columns)"), and #3348 should not grow it.

The cheap, honest fix is to stop promising what the search does not do:

1. Drop **"genuinely means nothing points at `Target`"** from `IsComplete`'s summary. That sentence is the falsifiable part; the rest of the property is fine. Say what it measures: every modelled table was read in full, and every row's **edge columns** were readable.
2. Make the blob limit an **unconditional** caveat in the renderer, not one contingent on `UnscannedTables` being non-empty. It is always true, so it should always print. Something as short as *"references carried inside signature blobs are not searched"* under the count.

Optionally, the symmetric-with-`UnscannedTables` version: a third blind-spot term for blob-bearing columns, which would drive `IsComplete` false whenever the image has signatures — i.e. always. That is defensible for the same reason you accepted an always-false `IsComplete` for tables, but it makes the flag carry no information at all, so I lean to (1)+(2).

Either way the boundary I would hold is the one you already articulated on the table gap: *a limit the caller cannot see is worse than a slower answer.* Today a caller cannot see this one.

### Non-blocking

Your "acknowledged boundary" — that the invariant proves rows were *visited*, not correctly *examined* — is the right call to defer, and I agree it is ordinary scanner correctness. Noting only that the finding above is **not** an instance of it: the rows here are examined exactly as designed, and the design is what does not reach the edge.

---

Evidence: clean detached worktree at `2cc43421`; Release; SDK `11.0.100-preview.7.26366.102`. Probes were file-based apps under `/tmp/` referencing `src/ILInspector.Metadata`; the minimal assembly was built in `/tmp/minimg`. Tamper applied and reverted; `git status --porcelain` empty and 1006/0 re-confirmed before posting. No repository files modified.

richlander:
## Third pass at `2cc43421`

Head is unchanged since my last review and my blob-edge finding (comment 5101103779) is still open, so rather than restate it I spent this pass attacking parts of `FindReferences` I hadn't tested. Two minor findings, one clean negative result — and one mistake of my own that I want on the record, because the wrong answer it produced looked exactly like a clean bill of health.

### Minor 1: truncation on a table's **final row** leaves that table undisclosed

Commit `3727b8bd` — "Count a table as searched only when every one of its rows was examined" — fixed GPT's finding that recording table *entry* was not the same as recording completion. The same distinction applies one level down, at columns, and there it is still open.

The reconciliation states the reasoning explicitly:

> Truncation on the *final* row → also `rowCount+1` → visited, correctly (**every row was examined**, results merely capped).

That is not what happens. The budget check sits *inside* the column loop:

```csharp
if (references.Count >= maxReferences) { truncated = true; break; }
```

so the final row is **entered and abandoned mid-columns**. `rid` still reaches `rowCount + 1`, so the table is marked visited and drops out of `UnscannedTables`.

Constructed on `ILInspector.Metadata.dll`:

```
TypeDef:364 budget=7 -> truncation trips on MethodImpl[211], the LAST of 211 rows
   Truncated=True   MethodImpl disclosed in UnscannedTables? False   (listed: 13)
```

The columns after the tripping one in `MethodImpl[211]` were never examined, and `MethodImpl` is not in the list whose own rendered wording is *"populated tables were not searched in full"*. It was not searched in full.

**How much this matters:** less than it sounds. `Truncated` is true, so `IsComplete` is correctly false and no caller is told the answer is trustworthy. And with the budget exhausted, no further edge could have been reported anyway. The concrete harm is narrow — a caller using `UnscannedTables` to decide *what to re-scan* is handed a list missing the one table it stopped inside. I'd call it non-blocking. But the written justification for the case is wrong, and that's worth correcting whether or not you change the code.

**Verified one-line fix.** At the visited check, `truncated == true` can only mean truncation happened in *this* table, because the outer loop breaks at the top of the next iteration:

```csharp
if (!truncated && rid > reader.GetTableRowCount(spec.Index))
    visited.Add(spec.Index);
```

Applied it and re-ran the constructed case: `MethodImpl disclosed? True (listed: 14)`. Suites stay green — `ILInspector.Metadata.Tests` **1006/0**, `DotnetInspector.MetadataRendering.Tests` **61/0**. Reverted; baseline re-confirmed at 1006/0.

Note the corollary: the tests pass *both* ways, so today's behavior isn't pinned either. If you take the fix it wants a test; if you keep the current behavior, the justification comment should say the final row is entered but not fully examined, and that `Truncated` is what covers it.

### Minor 2: an out-of-range target gets a confident answer

`FindReferences` validates the lower bound and not the upper:

```csharp
ArgumentOutOfRangeException.ThrowIfLessThan(targetRowId, 1);
```

On the two-type assembly from my last review (3 TypeDef rows, all tables modelled):

```
TypeDef[3]     (exists)          refs=0 IsComplete=True unscanned=0
TypeDef[4]     (DOES NOT EXIST)  refs=0 IsComplete=True unscanned=0
TypeDef[99999] (DOES NOT EXIST)  refs=0 IsComplete=True unscanned=0
```

`mdi --references TypeDef:99999` prints "No row points at TypeDef[99999]." for a row that does not exist. A typo'd row id is indistinguishable from a real, genuinely unreferenced one — and `IsComplete=True` actively vouches for it. The asymmetry is the tell: the bound was considered, and only half of it was checked. `reader.GetTableRowCount(targetTable)` is already read in the scan loop, so rejecting (or flagging) an out-of-range target is cheap.

### Negative result: coded-index confusion — clean

`TypeDefOrRef` and friends pack a table tag with a row id, so the obvious failure is `FindReferences(TypeDef, 5)` matching a column that actually points at `TypeRef[5]`. It doesn't. `PointsAt` compares a resolved `TargetTable`, and it holds empirically:

```
[coded index] rows 1..372 checked; edges shared between TypeDef[n] and TypeRef[n]: 0
[control]     rows where the two answers differ in size: 291 / 372
```

The control matters — without it, "0 shared edges" would also be what you'd see if both queries returned nothing.

### My own error, disclosed

My first truncation sweep reported **"2 of 32 truncating budgets leave the stopping table undisclosed"** and I nearly wrote it up. It was wrong. I had used `References[^1].Source.Table` as the stopping table, but the tripping match is *never added* — so the last added reference can belong to an earlier table entirely. Corrected to `full.References[budget]` and got **0 of 65**.

That clean result was itself worthless: when I instrumented it, **0 of those 65 budgets had stopped on a table's final row**, so the sweep never exercised the one case in question. Only constructing the case deliberately surfaced the real behavior above.

Three steps, two of which produced confident wrong answers in opposite directions. Recording all of them because the middle one — "0 of 65, clean" — is exactly the sort of result that gets quoted as evidence a thing was checked.

---

Evidence: clean detached worktree at `2cc43421`, Release, SDK `11.0.100-preview.7.26366.102`. Probes were file-based apps under `/tmp/` referencing `src/ILInspector.Metadata`. Candidate fix applied and reverted; `git status --porcelain` empty and 1006/0 re-confirmed before posting. No repository files modified.

richlander:
## Re-review at `053490dc` — one blocking finding, two minors, and a restack request for #3360

Reviewed the live head. Setup notes first, so the rest is easy to check:

- The head moved `2cc43421` → `053490dc`. That is GitHub restacking the branch, not a rewrite by the author — no process concern, and I withdraw the "force-push over reviewed history" framing I was heading toward.
- `git diff 2cc43421 053490dc -- src/ILInspector.Metadata/ docs/design/metadata-table-projection.md tests/ILInspector.Metadata.Tests/ src/mdi/` is **empty**. This PR's own content is byte-identical across the restack, so my three earlier reviews carry over unchanged and the findings below are not stale citations.
- Suites at `053490dc`, Release, this box: `ILInspector.Metadata.Tests` **1006 / 0** (5 skipped), `DotnetInspector.MetadataRendering.Tests` **61 / 0**. Green, and unchanged by the restack.
- CI reads `UNSTABLE` only because of pending / `SKIPPED` (path-gated) / `CANCELLED` (superseded push) jobs. No genuine failures.

The `UnscannedTables` work is a real improvement and I verified it properly last pass: deriving the set from the traversal rather than from `SupportedTables`, and refusing to count a table as searched merely because the loop entered it, is the right shape. Tampering the gate to an unconditional `visited.Add(spec.Index)` fails `FindReferences_TableStoppedPartWayThrough_IsNotCountedAsSearched`, so it is non-vacuous. I also ran `npx markdownlint-cli docs/design/metadata-table-projection.md` → **exit 0**, closing the gap you flagged as unrunnable in your environment.

---

## Blocking: the `<remarks>` mitigation does not cover the case where the flag is wrong

`MetadataRowReferences.cs:167` still promises:

> Whether the search covered the whole image ... so an empty `References` **genuinely means nothing points at `Target`**.

The `<remarks>` immediately below concedes this is "false for essentially every real assembly." That concession is accurate, and it is why I did not treat this as blocking on my first pass. On re-reading, it argues the wrong way round.

**`IsComplete` returns `true` exactly when every populated table happens to be modelled — i.e. on small, simple assemblies. That is precisely the case where the summary's promise is false.** "Almost always false" is not a safety property when the residual `true` is the unsound one.

Signature blobs carry TypeDefOrRef tokens and sit in **`Heap`-kind columns of tables the scan reads in full**. They are not truncation, not an unreadable row, and not an unscanned table. No blind spot fires.

Two types, three `TypeDef` rows, at `053490dc`:

```csharp
public class Alpha { }
public class Beta { public Alpha F; }
```

```
## TypeDef (3 rows)
| 2 | Public, BeforeFieldInit | Alpha | ... | nil        | MethodDef[1..2) |
| 3 | Public, BeforeFieldInit | Beta  | ... | Field[1..2) | MethodDef[2..3) |

## Field (1 row)
| 1 | Public | F | 061208 |
```

`Field[1].Signature` is `06 12 08` — `FIELD`, `ELEMENT_TYPE_CLASS`, TypeDefOrRef `08 >> 2 = 2`, tag `0` = `TypeDef`. **The edge onto `TypeDef[2]` is printed in hex by the tool itself.** The search's verdict on the same image:

```
IsComplete    : True
Truncated     : False
UnreadableRows: 0
UnscannedTbls : []
references    : 0
```

```
$ mdi Mini.dll --references TypeDef:2
## References to TypeDef[2] (0)

No row points at TypeDef[2].
```

One command prints the pointing edge; the other denies it exists; the completeness flag asserts the denial is trustworthy. ### Scale, and an important scoping note

Re-measured at `053490dc` against this repo's own `ILInspector.Metadata.dll`, using SRM's `SignatureDecoder` with a recording type provider — an oracle independent of the product's scan, not the product checking itself:

```
TypeDef rows                             : 372
blob edges onto TypeDefs                 : 3198
distinct TypeDefs targeted by a blob edge : 217
...of those, reporting ZERO references of any kind : 22
    TypeDef[137] MetadataProjectionOptions  blobEdges=41  refs=0  IsComplete=False  Unscanned=7
    TypeDef[10]  ChangeKind                 blobEdges=33  refs=0  IsComplete=False  Unscanned=7
    TypeDef[145] MethodClassification       blobEdges=26  refs=0  IsComplete=False  Unscanned=7
```

**Scoping this honestly: on *this* DLL `IsComplete` is `False`, so the disclosure is working.** The scale numbers measure how large the *undisclosed edge class* is — 3,198 edges that no blind spot accounts for — not a live false negative on this image. `Mini.dll` above is the live false negative: it is what happens when the three disclosed blind spots do not fire.

That is the whole shape of the finding. `TypeDef[137] MetadataProjectionOptions` — 41 blob uses, zero reported, and a parameter of `FindReferences` itself — is the memorable case, but the argument rests on the 3-row assembly, not on it.

Per AGENTS.md — *"a doc comment that asserts a safety, soundness, or faithfulness property must name the gate that enforces it, or explicitly mark the property as unverified"* — and *"do not turn analysis failures into success-shaped empty output."*

**Suggested fix, and deliberately the small one.** I am **not** asking you to grow the scan into blob decoding; that is real work and belongs in its own change.

1. Drop "so an empty `References` genuinely means nothing points at `Target`" from the summary. Say what the flag actually means: the scan finished, every row read, every populated table read in full — **over modelled columns**.
2. Add blob-carried edges to the `<remarks>` as a *fourth, undisclosed* limit, distinguished from the three that are disclosed.
3. Print an unconditional caveat in the renderer's reverse-reference output, since "No row points at X" is the sentence a reader acts on. It is unconditional because unlike the other three it cannot be detected per-query.

---

## Minor 1 — truncation on a table's *final* row leaves that table undisclosed

`3727b8bd` fixed "entered ≠ completed" at the row level. The same gap is open one level down, because the budget check sits *inside* the column loop. `MetadataTableProjector.cs:313`:

```csharp
if (rid > reader.GetTableRowCount(spec.Index))
    visited.Add(spec.Index);
```

If the budget trips while scanning the columns of a table's **last** row, the row loop still completes, `rid` still passes the count, and the table is recorded as searched — so it never reaches `UnscannedTables`. The reconciliation's justification ("every row was examined") is not true in this case: the last row was entered, not finished.

Constructed on `ILInspector.Metadata.dll`: `--references TypeDef:364 --max 7` trips on `MethodImpl[211]`, the last of 211 rows. `MethodImpl` is absent from `UnscannedTables` (13 listed). Re-derived at `053490dc` by sweeping every `TypeDef` target against every budget: this is the **only** such case in the image, which is why nothing catches it.

Verified one-line fix — the outer loop already `break`s on `truncated`, so at this point `truncated == true` ⟺ truncation happened in *this* table:

```csharp
if (!truncated && rid > reader.GetTableRowCount(spec.Index))
```

With it, the case discloses correctly — `in UnscannedTables? False -> True`, 13 -> 14 listed. Both suites stay green with the fix applied: **1006 / 0** (5 skipped) and **61 / 0**.

Corollary worth stating plainly: the suites are green **both ways**, so current behavior here is not pinned in either direction.

---

## Minor 2 — an out-of-range target gets a confident answer

`MetadataRowLocation` validates only the lower bound (`ThrowIfLessThan(RowId, 1)`). On the 3-row image above:

```
$ mdi Mini.dll --references TypeDef:99999
## References to TypeDef[99999] (0)

No row points at TypeDef[99999].
```

`IsComplete = True`. A row id past the table's row count is a caller error, not an empty result; it should throw or surface as a blind spot rather than render as a clean negative.

---

## Negative results

Reporting these so nobody re-runs them:

- **No modelled table is missing an edge column.** Checked `TypeDef.Extends`, `TypeRef.ResolutionScope`, `MemberRef.Class`, `MethodImpl.Class`/`MethodBody`/`MethodDeclaration`, `ExportedType.Implementation`, `GenericParam.Owner`, `MethodSpec.Method`, and every list column. All present. The cheaper version of the finding above does not exist.
- **Coded-index confusion is clean.** 0 shared edges across 372 rows between `TypeDef[n]` and `TypeRef[n]`; `PointsAt` compares the *resolved* table. Sensitivity control: 291/372 of those answers genuinely differ, so the check is not vacuous.

And one of my own errors, because the middle result is the kind that gets quoted as proof: my first sweep for minor 1 reported "2 of 32 undisclosed" (wrong — it read `References[^1].Source.Table`, but the tripping match is never added). Corrected to `full.References[budget]` it reported "0 of 65, clean." Instrumenting showed **0 of those 65 had stopped on a final row** — so the clean result was vacuous, not reassuring. Only deliberate construction found the behavior.

---

## Request for #3360 (filed here since this is the PR that owns the contract)

#3360 is currently `DIRTY` and predates this PR's `UnscannedTables` work. Simulating its restack — `git merge-tree --write-tree 53e38e6d ab0d4db9` — conflicts in exactly the files this review is about:

```
CONFLICT (content): docs/design/metadata-table-projection.md
CONFLICT (add/add): src/ILInspector.Metadata/MetadataRowReferences.cs
CONFLICT (content): src/ILInspector.Metadata/MetadataTableProjector.cs
CONFLICT (content): src/DotnetInspector.MetadataRendering/MetadataProjectionRenderer.cs
CONFLICT (add/add): tests/.../MetadataRowReferenceRendererTests.cs
```

**I first assumed this was a silent-regression risk, and that was wrong — I checked.** The code side is gated: the auto-merged `MetadataRowReferenceSearchTests.cs` carries **11** references to `UnscannedTables`, so resolving `MetadataRowReferences.cs` toward #3360's side (`IsComplete => !Truncated && UnreadableRows.IsEmpty`, zero mentions of `UnscannedTables`) fails to compile. The compiler catches it.

**The doc is not gated.** The conflict in `docs/design/metadata-table-projection.md` is a single hunk, 31 lines vs 9. #3360's side deletes the "projection's table coverage is itself a blind spot" bullet, the "`UnscannedTables` is derived from the traversal, not declared" bullet, and the honest reading — and restores:

> `IsComplete` is true only when neither happened, which is what makes an empty result trustworthy.

Nothing compiles a Markdown paragraph, and `markdownlint` checks formatting, not claims. So the restack can quietly undo this stack's own review outcome in prose while the code stays correct.

**Request: resolve that hunk to `53e38e6d`'s side, not `ab0d4db9`'s**, and re-run `npx markdownlint-cli` after. If the `IsComplete` wording changes here as suggested above, #3360's paragraph needs to match rather than being reconciled independently.

---

## Status

The rest of the stack is in good shape: #3345 is `CLEAN` with no findings from me, and #3350 / #3352 have none of their own — they are blocked only by stack position. This PR is the one carrying open findings. The blob-edge item is the blocker; the two minors are cheap and I would take them, but they are not the reason to hold.

The "Ready to merge" comment above predates all of this and should be treated as superseded.

richlander:
## Round-three and round-four review reconciliation

Three review comments landed after I marked this ready, and the fixes drew two more rounds of adversarial review. All findings are resolved. Commits: [`260c935e`](https://github.com/richlander/dotnet-inspect/commit/260c935e), [`f6bbcc71`](https://github.com/richlander/dotnet-inspect/commit/f6bbcc71), [`32c808d8`](https://github.com/richlander/dotnet-inspect/commit/32c808d8).

### The blocking finding was real, and the fix is to scope the claim

Signature blobs carry `TypeDefOrRef` coded tokens in `Heap`-kind columns of tables the scan reads **in full**. Nothing is truncated, unreadable, or unscanned, so **no blind spot fires** and the edge silently vanishes from a result reporting itself complete.

Reproduced on the two-type assembly from the review comment. `--table Field` printed `Field[1].Signature = 061208` — `06`=FIELD, `12`=ELEMENT_TYPE_CLASS, `08`=compressed TypeDefOrRef → `TypeDef[2]` = `Alpha` — while `--references TypeDef:2` answered `No row points at TypeDef[2].` with **zero caveats**. The tool printed the edge and denied it in the same breath.

What made this blocking rather than cosmetic is the **risk inversion**: `IsComplete` is true exactly when every populated table happens to be modelled — small, simple assemblies — which are precisely the images where a blob edge is the *only* remaining way to miss a reference. Disclosure was strongest where it was least needed.

Decoding blobs stays out of scope. The renderer now discloses the limit **unconditionally**, since there is no per-scan signal to condition it on, and `IsComplete` no longer promises image-wide coverage. Conditioning that caveat on `IsComplete` fails two tests — that mutation is the one that hides the limit exactly where it matters.

### I published something false last round, and I am correcting it

I wrote that truncation on a table's final row was "correctly visited (every row was examined)". **That was wrong.** The budget is checked inside the *column* loop, so the final row is entered and abandoned part-way through its columns while the row counter still reaches `RowCount + 1`. The reviewer was right and I was not.

### My first fix for that then broke the opposite boundary

Guarding the gate with `!truncated` fixed under-reporting and introduced over-reporting: if the budget trips on the **last column of the last row**, every cell was examined, so reporting the table unscanned is a *false* blind spot. Caught by GPT-5.6 Sol on re-review.

The gate now tracks the precise fact — `column + 1 < cells.Length || rid < rowCount` — instead of using truncation as a proxy. Both directions are pinned, each by exactly one test: reverting to `!truncated` fails only the final-cell test, removing the guard entirely fails only the final-row test.

### Out-of-range targets: disclosed, not rejected

I first implemented this as a throw. That was wrong — a dangling edge points at exactly the rows that do **not** exist, and two existing tests deliberately ask that question. Rejecting it makes the question unaskable. `TargetExists` preserves the question while killing the confident-empty answer, consistent with how everything else here is handled: disclose, do not silently succeed.

### :warning: One contract change I want you to look at

`MdiCommandTests.ExecuteReferences_NonexistentRow_IsAnEmptyAnswerNotAnError` asserted the old behaviour **by name**. `git log -L` shows it arrived in this stack's original commit with no comment and no design-doc rationale, so I read it as incidental rather than considered, and flipped it to assert the caveat. Exit code stays `0`.

If you intended that assertion as a real contract, say so and I will revert this piece.

### Findings raised and dismissed, with evidence

A `CustomAttribute.Value` blob edge was filed as BLOCKING. It does not hold, on two independent grounds, and the reviewer withdrew it after reproducing:

- The cited row `TypeDef[3]` is `MyAttribute`, not `Alpha`, and `--references TypeDef:3` in fact returns `CustomAttribute[12].Parent` — not the quoted empty result.
- The blob `010005416C7068610000` decodes as `0100` prolog, `05` length, ASCII `"Alpha"`, `0000` — a serialized type **name** per ECMA-335 II.23.3, not a coded token. It is not a row-to-row edge.

The genuine residue was real and is fixed: "signature blobs" was narrower than the true class of blob-resident references, so the caveat now names both kinds explicitly.

A second BLOCKING claimed a table whose final row failed to read was reported fully searched. Also withdrawn after reproduction: that case yields `IsComplete=False`, `UnreadableRows=[TypeDef[2]]`, and the caveat "1 row had an edge that could not be read". `UnreadableRows` and `UnscannedTables` **partition** rather than overlap — read-but-undecodable versus never-examined — and folding them together would print a sentence that is false of the table. Both reviewers probed this independently, so the rationale is now written into the design doc.

### Mutation testing caught a vacuous test of my own

My first `TargetExists` test asserted `IsComplete == false` on the self image — where `UnscannedTables` is already non-empty, so it passed **regardless of the fix**. Exactly the class of defect I was asked to look for. It now runs against a synthetic image whose populated tables are all modelled, and asserts a genuinely complete scan *first*, so it cannot quietly go vacuous.

### Validation

`dotnet build dotnet-inspect.slnx -c Release` clean. `ILInspector.Metadata.Tests` **1011/0/0**. `DotnetInspector.MetadataRendering.Tests` **62/0**. `mdi` verified end to end in `md`/`tsv`/`jsonl`, stdout machine-pure with caveats on stderr. `TypeDef:364 --max-references 7` now lists `MethodImpl` among the unscanned tables, which was the original symptom.

Both re-reviews on the fixed head returned clean.

richlander:
Ready to merge

All checks green on `32c808d8` (`changes`, `markdownlint`, `build-net10`, `test (ubuntu-26.04, linux-x64)`), `mergeStateStatus: CLEAN`. Both fixed-head adversarial reviews returned clean.

One open question that is **not** blocking, flagged in the reconciliation above: I flipped `MdiCommandTests.ExecuteReferences_NonexistentRow_IsAnEmptyAnswerNotAnError` to assert the new caveat. If that assertion was a deliberate contract rather than incidental, say so and I will revert that piece.

Non-blocking follow-up, tracked separately: decoding signature blobs so blob-carried edges become findable rather than merely disclosed. The design doc already lists it as future work.

richlander:
## Re-review at `2a9d9ae3` — all three findings fixed, verified by tamper. No open items from me.

The restack brought real work, not just a replay: `e8e1e0b6`, `d5f4b8a4` and `2a9d9ae3` are new since the head I last reviewed. All three of my findings are addressed, and each fix is gated. I re-derived every claim rather than reading the commit messages.

### Blocking finding (blob edges) — fixed and gated

`mdi Mini.dll --references TypeDef:2` on the same two-type assembly now prints:

```
## References to TypeDef[2] (0)

No row points at TypeDef[2].

Blob payloads are not searched, so a reference spelled inside one — as a type token in a
signature, or as a type name in a custom-attribute value — is not reported.
```

Suppressing that line (`yield break` in `MetadataProjectionRenderer.cs:247`) fails **three** tests — `Caveats_ForACleanSearch_AreTheBlobLimitAlone`, `Markdown_EmptyAndClean_QualifiesNothingPointsHere`, `Caveats_ReportEachBlindSpotIndependently`. The disclosure is enforced, not just present.

The `IsComplete` contract is now the honest one. It drops "genuinely means nothing points at `Target`", states outright that the property **must not** be read as "nothing points here", enumerates the six blob-bearing columns, and keeps the inverted-risk reading: true mainly on small simple images, which is exactly where a blob edge is the only limit left to hide one. Scoping the caveat to a renderer line rather than growing the scan into blob decoding is the right call.

### Minor 1 — my suggested fix was wrong, and the test proves it

I offered `if (!truncated && rid > ...)` as a verified one-liner. It is verified only in one direction. GPT-5.6 Sol caught the other: if the budget trips on the **last column of the last row**, every cell the table has was examined, so marking it unscanned is a *false* blind spot — it tells the reader an unread row could hide an edge when no row went unread. `abandonedMidTable` tracks the actual fact ("columns after this one on this row, or any row after it") instead of using truncation as a proxy. That is strictly better than what I proposed, and I was wrong to call my version complete.

`d5f4b8a4` claims each direction is pinned by exactly one test. Verified, and the claim is exact:

| Tamper | Result |
| --- | --- |
| gate → `!truncated` (my one-liner) | **1 failure**: `FindReferences_StoppedOnATableSVeryLastCell_IsStillCountedAsSearched` |
| guard removed entirely | **1 failure**: `FindReferences_StoppedOnATableSLastRow_IsStillNotCountedAsSearched` |

Exactly one failure each, different tests, 1055 total. Neither tamper disturbs anything else.

### Minor 2 — fixed

`TargetExists` joins the `IsComplete` conjunction, and the renderer says so:

```
TypeDef[99999] is past the end of its table, so no row exists to point at — a row id this
large is usually a typo.
```

`FindReferences_ForARowPastTheEndOfItsTable_SaysTheRowIsNotThere` covers it, and I note the
non-vacuity guard at `MetadataRowReferenceSearchTests.cs:211` — *"The row must exist, or this is
not the case under test."* That is the pattern from the #3370 gate carried across on its own.

---

## One thing to watch before merging: #3360 will revert this PR

Not a defect in this PR, but it lands on this PR's files. #3360 is still at `ab0d4db9` — it was not restacked with the rest and is now `CONFLICTING`. Its side of those conflicts predates every fix above.

Details and a demonstration are on richlander/dotnet-inspect#3360. The short version: at the previous head I checked this exact risk and **reported it as not a risk**, because the test file auto-merged and would have failed to compile. The rebase changed that. It is a risk now.

## Status

No open items from me on this PR. Its own suites at the stack tip: `ILInspector.Metadata.Tests` **1055 / 0** (5 skipped), `DotnetInspector.MetadataRendering.Tests` **99 / 0**.

PR #3350, opened 2026-07-27
Describe the metadata image and read heap values by address

richlander:
Stacked on #3348 (which is stacked on #3345). Review only the top commit; the base will be retargeted to `main` as the stack lands.

Closes the last explorer-enabling gap in #3341 that is not measurement-gated: **gap 5, heaps and headers**.

## What this adds

The projection covers *tables*. A metadata browser also shows the container it sits in, and needs to read a heap value the tables never point at.

- `MetadataImageInspector.Describe(PEReader)` → `MetadataImageOverview`: metadata root version/kind/offset/size and manifest flag, one summary per heap, one summary per ECMA-335 table, and PE/CLI header facts. Returns `null` for an image with no metadata — the same "not applicable" signal `ProjectRow` uses.
- `MetadataTableProjector.ReadHeapValue(peReader, heap, address, options)` — the heap counterpart of `ProjectRow`.
- Session facets `MetadataImage()` and `MetadataHeapValue(...)`, renderer support, and `mdi --overview` / `mdi --heap Heap:Address`.

## The two decisions worth attacking

**1. Row counts cover every table, not just the projected ones.**

Each entry carries `IsProjected`. A table with rows the projection does not model — `NestedClass`, `MethodSemantics`, `Property` on a typical assembly — stays visible as a coverage gap rather than being indistinguishable from an empty table. Listing only projected tables would reproduce the success-shaped-absence failure the reverse search in #3348 exists to avoid.

**2. Heap addressing is part of the model, not a doc footnote.**

ECMA-335 addresses String, Blob, and UserString by byte offset, but the GUID heap by **1-based index** into a vector of 16-byte values — and SRM's `GetHeapOffset` follows suit, returning an index for a GUID handle. I verified this against SRM before designing the API: on a 16-byte GUID heap, `GetHeapOffset(mvid)` returns `1`, not `0`.

`MetadataHeapAddressing` states which convention a heap uses and `MaxAddress` applies it, so a caller cannot silently read a GUID address as a byte offset. `ReadHeapValue`'s address is exactly what `MetadataValue.HeapReference.Offset` publishes, so a projected cell's offset round-trips, and the result is the same `MetadataValue` shape a projected cell carries — one renderer serves both.

Address zero is every heap's nil value. An address past the end yields `Malformed`, not an empty value. Because the tables never reference `#US`, this is also the only way to browse the user strings IL points at.

## Deliberately not here

**Heap enumeration.** SRM exposes no public way to walk every entry of a heap. The surface is overview plus random access, and the design doc says so rather than faking a walk by scanning bytes.

## Evidence

- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release` — **1044 passed, 0 failed, 0 skipped** (zero skipped means the mdv oracle ran).
- `dotnet run --project tests/DotnetInspector.MetadataRendering.Tests -c Release` — **95 passed, 0 failed** (was 58).
- `dotnet run --project src/dotnet-inspect.Tests -c Release` — 2216 total, 10 skipped (ilasm/ildasm absent), **1 failure: `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`**. Confirmed **pre-existing**: it fails identically on the unmodified base commit `a1ea72c4`. Unrelated to metadata; not fixed here.
- `mdi --overview` and `mdi --heap` verified end to end in md, tsv, and jsonl, including the stderr caveat path for machine formats and the mode-conflict error.

**Mutation-checked.** Forcing the GUID heap to byte-offset addressing fails three tests. Note what the round-trip test *cannot* prove: both sides construct their handle from the same address, so a shared-but-wrong convention would survive it. A separate test therefore anchors the GUID address to the module MVID read through SRM's typed API, and the round-trip test's own doc comment says so rather than overclaiming.

The "no metadata" contract is exercised on a real image — a copy of the test assembly with its CLI data directory zeroed — guarded by a canary asserting `HasMetadata` is actually false, so the null assertion cannot pass for the wrong reason.

## Compatibility

Purely additive. No existing model, CLI, or `mdi` behaviour changes; the projection's shape is untouched. `MetadataTableProjector.ProjectedTables` is new public surface (the projector's own supported-table set), used by the overview and pinned by a test so the two cannot drift.

### DIFF
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -346,9 +346,12 @@ inspector), a standalone tool that renders the tables the way `mdv` does:
   `MetadataProjectionOptions` and delegates all output to the renderer below.
   Surface: `mdi <assembly> [--table|-t <Names>] [--format|-f md|tsv|jsonl]
   [--max-rows|-n N] [--start-row|-s N] [--references|-r Table:RowId]
-  [--max-references N] [--max-bytes N] [--max-chars N]`. Missing
+  [--max-references N] [--overview|-i] [--heap Heap:Address]
+  [--max-bytes N] [--max-chars N]`. Missing
   files, native images, and unreadable metadata surface as visible errors, never
-  success-shaped empty output.
+  success-shaped empty output. `--references`, `--overview`, and `--heap` each
+  select a different view and are rejected in combination rather than silently
+  ranked.
 - **`src/DotnetInspector.MetadataRendering`** — a small reusable library holding
   `MetadataProjectionRenderer` (projection → Markout tables). It lives in the
   product-side `DotnetInspector.*` family, **not** in `ILInspector.Metadata`,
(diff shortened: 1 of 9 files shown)


richlander:

Two-model review at this stack boundary, on the grounds that heap addressing conventions and out-of-range handling carry subtle correctness risk. Reviewers: **MAI-Code** and **Gemini 3.1 Pro**, each isolated in its own detached read-only worktree, scratch confined to `$env:TEMP`. Neither is my own model family.

Reviewed head moved `dd8ff621` → **`7d7c618d`**. Both reviewers are clean on the fixed head.

### Blocking finding (Gemini) — accepted, fixed in `7d7c618d`

`MetadataImageInspector.RowCount` caught `ArgumentOutOfRangeException` / `BadImageFormatException` and returned a row count of `0`. A table reported as zero rows is indistinguishable from a genuinely empty one, so a failure to read a row count would have disappeared into a plausible-looking overview — the success-shaped-absence failure this repository forbids, and the same class of bug #3348 shipped and had to fix.

The helper is gone; `DescribeTables` now calls `reader.GetTableRowCount(index)` directly, with a comment recording why no defence belongs there so it does not get re-added.

### Where I disagreed with the finding

Gemini described this as a live bug that "actively suppresses corruption," verified "via source review." I did not accept that characterisation, and it turned out to be wrong. Both arms are **unreachable**:

- `GetTableRowCount` is an SRM extension method whose body is a null check, a check of `(int)tableIndex >= MetadataTokens.TableCount`, and an index into `reader.TableRowCounts`. Its documented exceptions are `ArgumentNullException` and `ArgumentOutOfRangeException` only — there is no path by which it reports a malformed image. `TableRowCounts` is populated during `MetadataReader` construction, so a corrupt table stream throws out of `GetMetadataReader` first.
- I probed it: an image whose `#~` stream size is overwritten with `4` gives `HasMetadata: True`, then `GetMetadataReader` throws `BadImageFormatException: Metadata table header too small`. `RowCount` is never reached.
- Measured, not assumed: highest declared `TableIndex` is **55**, `MetadataTokens.TableCount` is **64**, so the range check cannot fire for any value `Enum.GetValues<TableIndex>()` yields.

**MAI-Code reached the same conclusion independently** — it tried to drive the catch with byte-mutated images and could not, and explained why.

I put my reasoning back to Gemini adversarially and invited it to show me I was wrong. It re-derived the chain by decompiling `System.Reflection.Metadata` and **withdrew the characterisation**: "my previous characterization ... was wrong ... The `catch` block was 100% dead code."

So the commit removes a latent hazard, not a live bug. It was still worth removing: the comment documented the wrong behaviour as intentional, and the catch would have silently converted any future failure into a legitimate-looking empty table.

### What the new test does and does not prove

`Describe_FailsVisiblyForACorruptTableStream` pins an invariant that was previously untested end to end: corrupt metadata throws, absent metadata returns null, and the two stay distinguishable.

It does **not** prove the removal. The throw originates in `GetMetadataReader`, so the test passes with or without the catch. Unreachability is what justifies the removal; the test guards the invariant. Both reviewers confirmed this admission is accurate rather than letting it stand — Gemini: "Your admission that the test 'passes with or without the catch' is absolutely accurate."

Both canaries were verified load-bearing: the intact image really reports `TypeDef` rows (307), and the corrupt image really still has metadata, so the test cannot pass via the no-metadata path.

### Where the reviewers contradicted each other

On `SelfWithoutCliHeader()`'s hand-computed data-directory offset, Gemini called the test "rock solid" while assuming the standard layout. MAI-Code actually checked which branch runs and found the test assembly is **PE32**, not PE32+.

Both independently derived the same constants from the header layout — PE32 optional-header fixed part 96 bytes, PE32+ 112, data directory 14 at `+14*8` — so the arithmetic is settled and correct. But the practical consequence is that **the PE32+ branch of that expression has never been executed**. Recording it as a known, non-blocking limitation rather than papering over it; it is test-fixture reach, not product behaviour.

### Verified, not taken on trust

MAI-Code's first pass answered failure-visibility and test-quality generically ("the tests passed"). I rejected both and asked it to demonstrate instead. It then built corrupt images under `$env:TEMP` and reported observed behaviour — including that zeroing the COR20 directory yields `HasMetadata=false` → `Describe` returns null, while corrupting the metadata root or `#~` stream throws.

On the fixed head it also ran the CLI end to end against a corrupt image: `mdi --overview` exits **1** with `Error: cannot read metadata from '...': Metadata table header too small`. Gemini confirmed the same path, checking that `MdiCommand.ExecuteOverview`'s `catch ... when (IsExpectedReadFailure(ex))` reports the failure rather than swallowing it. So propagating the exception surfaces as a visible error, not an unhandled crash and not an empty success.

Claims 1–2 and 4–9 were verified by both. Gemini independently probed the GUID heap and confirmed the `SizeInBytes` non-multiple-of-16 case rejects correctly (`15 / 16` → max address 0), and that `ProjectedTables` derives from `SupportedTables` so it cannot drift.

### Evidence at `7d7c618d`

Run by me and reproduced independently by both reviewers:

- `tests/ILInspector.Metadata.Tests` — **1045 passed, 0 failed, 0 skipped**.
- `tests/DotnetInspector.MetadataRendering.Tests` — **95 passed, 0 failed**.

The pre-existing `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden` failure in `src/dotnet-inspect.Tests` is unchanged and unrelated; it reproduces on the base commit.

As noted in the description, CI reports no checks by design — `ci.yml` triggers only on PRs targeting `main`, so a stacked PR gets none until its base lands.

richlander:
Ready to merge

PR #3352, opened 2026-07-28
Measure the projection's allocation shape instead of guessing at it

richlander:
Stacked on #3350 (which is stacked on #3348 and #3345). Review only the top commit; the base retargets to `main` as the stack lands.

Closes **gap 3** of #3341 — the one gap the issue marks as measurement-gated. The measurement now exists, and it closes the gap **without an allocation redesign**, so this PR changes no product code.

## Why measure at all

`AGENTS.md` forbids claiming runtime impact from static structure, and #3341's proposal was explicitly "measure first, then consider windowed/lazy row materialization and `ReadOnlySpan<T>`-shaped views." Rows materialize eagerly as `ImmutableArray<MetadataValue>`, which *looks* expensive. Whether it actually is, and whether it matters for the consumer this work serves, is an empirical question.

`eng/measure-metadata-projection-allocation.cs` is a file-based app in the existing `eng/*.cs` style. It reports two quantities because they answer different questions:

- **allocated** — total churn while projecting; a throughput and GC-pressure signal.
- **retained** — the live set still reachable from the finished projection after a forced collection. This is the number that decides whether a browser tab can hold the result.

Pinned to `Microsoft.NETCore.App/10.0.9/System.Private.CoreLib.dll` (16,017,232 bytes), workstation GC, .NET 11.0.0:

```bash
dotnet run eng/measure-metadata-projection-allocation.cs
```

| Scenario | Allocated | Retained | Rows | Cells |
| control: `PEReader` only, no rows | 0.0 MB | 0.0 MB | — | — |
| full (`MaxRowsPerTable = int.MaxValue`) | 222.8 MB | 74.7 MB | 182,719 | 670,982 |
| default (`MaxRowsPerTable = 4096`) | 82.3 MB | 19.8 MB | 45,501 | 145,332 |
| window 1000 | 17.6 MB | 5.0 MB | 12,002 | 38,014 |
| window 100 | 1.8 MB | 0.5 MB | 1,202 | 3,814 |
| window 100 @ row 40000 | 0.1 MB | 0.1 MB | 200 | 900 |
| window 100, `MethodDef` only | 0.1 MB | 0.1 MB | 100 | 600 |

Byte-identical across repeated runs.

## The conclusion

A full projection retains **74.7 MB for a 15.3 MB image — 4.9x the input**, at 117 bytes per cell, and churns 222.8 MB. #3333 measured 1.6 GB as already fatal in a browser tab, so a full projection of a large assembly is not something an explorer can hold.

**But an explorer never does that.** The row window from #3345 is **150x cheaper**: paging 100 rows across all tables retains 0.5 MB, and the realistic explorer interaction — one table, one window — retains **0.1 MB**. Windowing is what makes the projection deliverable to a browser, and it already exists. Nothing about the per-cell representation needs to change to serve this consumer.

## Negative results, recorded rather than discarded

**`ReadOnlySpan<T>` is the wrong tool for this problem.** Worth stating plainly because it was the intuitive candidate and is named in the issue. A span is a `ref struct`: it cannot be stored in a field, cannot live in an `ImmutableArray`, and cannot outlive the stack frame that produced it. The projection is a *retained object graph handed to a caller that holds it* — precisely the shape a span cannot express. Spans reduce transient churn; they do not reduce the retained set, which is the number that matters. Adopting one would mean abandoning the retained model, not optimizing it.

**Eager display text is the largest component and still not worth changing.** About a third of the retained graph is display text formatted for every cell:

| Component | Strings | Chars | Approx. |
| --- | --- | --- | --- |
| `Handle.Display` | 91,919 | 5,455,583 | 12.3 MB |
| `Flags.Decoded` | 170,032 | 2,107,819 | 7.6 MB |
| `HeapReference.Preview` | — | 3,347,232 | 6.4 MB |
| `Scalar.Display` | 118,790 | 519,424 | 3.5 MB |
| `HeapReference.Text` | — | 1,550,412 | 3.0 MB |

Two cheap wins are visible and were **deliberately not taken**: `Scalar`/`Flags` display strings repeat **7.2x** (288,822 strings, 40,357 distinct), and the 14,217 `Nil` cells are stateless and identical so could be a singleton. Both only pay off in the full-projection case no consumer performs; in the windowed case they are a rounding error on 0.1 MB. Taking them would add caching state and lifetime questions to the projector to improve a scenario nobody runs. The table is recorded so a future consumer that *does* need a full projection knows where to start.

## Methodology, and its one real hazard

Every scenario is measured with the `PEReader` still live, so retained bytes could in principle be SRM's rather than the projection's. That is why the **control row** exists: an open `PEReader` with no rows projected retains ~0 MB, which is what licenses attributing the rest to the projection. Each scenario is warmed up first so JIT and static init are not charged to it — without that, the 0.1 MB scenarios are dominated by first-call cost.

The probe resolves its pinned input through `DOTNET_ROOT` or the standard install locations so it runs on Linux as well as Windows, and accepts an explicit path for other assemblies.

## Evidence

- `dotnet build dotnet-inspect.slnx -c Release` — clean; the new `eng/` file is not picked up by the solution build.
- Measurement reproduced across three runs with identical output.
- No product code changed, so no behavioural test applies. The design doc records input, command, baseline, and result per `AGENTS.md`.
- **`markdownlint` could not be run locally — Node is not installed in this environment.** I self-checked the changed Markdown against the rules the repo enables (`.markdownlint.yaml` disables MD013/MD032/MD034/MD060): no trailing whitespace, no consecutive blanks, headings and tables blank-line separated, no duplicate headings, file ends with a newline. CI's `markdownlint` job is the real gate — but note it will not run here, since `ci.yml` only triggers on PRs targeting `main`.

### DIFF
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -436,11 +436,22 @@ oversights:
   could report "nothing points here" while a pointer sat in an unsearched table,
   which is worse than a slower answer.
 - **Blind spots are reported, not folded in.** `Truncated` marks a scan the
-  result budget stopped, and `UnreadableRows` lists rows that could not be
-  decoded and therefore could not be inspected. `IsComplete` is true only when
-  neither happened, which is what makes an empty result trustworthy. Unlike
-  `MetadataTableTruncation`, `Truncated` carries no total: a stopped scan never
-  learns how many references it did not reach.
+  result budget stopped, and `UnreadableRows` lists rows whose edges could not
+  be fully determined. `IsComplete` is true only when neither happened, which is
+  what makes an empty result trustworthy. Unlike `MetadataTableTruncation`,
+  `Truncated` carries no total: a stopped scan never learns how many references
+  it did not reach.
+
+`UnreadableRows` is subtler than "the row failed to read". The cell readers
+**contain** a decode failure as a `Malformed` cell rather than throwing, so a
+row holding a broken handle or list column reads back successfully and would
+otherwise pass as fully searched — a missed reference that reads as an absent
+one. A `Malformed` cell in an edge column therefore marks its row a blind spot.
+The **column's declared kind decides, not the cell**: a `Malformed` heap,
+scalar, or flags cell was never an edge and cannot hide a reference, so it is
+not counted. A row stays a blind spot only once, and after its good edges have
+been collected, so one broken column never costs the caller the edges that row
+does have.
 
 The search reuses the projection's single row-reading path rather than a faster
 private one, so a `HandleRef` or `HandleRange` means the same thing in a search
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -450,14 +450,43 @@ oversights:
   blind spot that under-reports is the whole failure this exists to prevent.
   Entering a table is deliberately not enough: a loop that stops part-way leaves
   rows unread, and an edge onto the target could sit in any of them. That also
-  makes the budget interaction fall out for free — a scan the budget stops
-  leaves both the table it stopped inside and every table after it unsearched,
-  and all of them are reported. Note that reaching the end of the row loop is
-  *also* not enough, because the budget is checked inside the **column** loop:
-  truncation on a table's final row leaves that row entered and abandoned
-  part-way through its columns while the row counter still passes the row count,
-  so a completed scan and a scan stopped on the last row are indistinguishable
-  from the counter alone. `Truncated` is what separates them.
+  makes the budget interaction mostly fall out for free — a scan the budget
+  stops leaves every table after it unsearched, and all of them are reported.
+
+  Reaching the end of the row loop is *also* not enough, because the budget is
+  checked inside the **column** loop: truncation on a table's final row leaves
+  that row entered and abandoned part-way through its columns while the row
+  counter still passes the row count, so the counter alone cannot tell a
+  completed scan from one stopped on the last row.
+
+  `Truncated` is too blunt to separate them, though, because of the boundary
+  case: if the budget trips on the **last column of the last row**, every cell
+  the table has was examined and the table genuinely was searched in full, even
+  though the scan ended inside it. Reporting it unscanned there would be a false
+  blind spot — it would tell the reader an unexamined cell could hide an edge
+  when no cell went unexamined. So the scan tracks the precise fact instead:
+  whether anything in this table was left unlooked-at, which is the columns
+  after the stop on that row, or any row after it.
+
+  That also keeps the malformed-cell blind spot sound. A stop before the last
+  column leaves the remaining columns unchecked for malformed edges, so the row
+  might belong in `UnreadableRows` without the scan knowing — but that is
+  exactly the case where the table is reported unscanned, so the gap is still
+  disclosed. When the stop is on the last column, every column was checked and
+  the row's status is fully determined.
+- **`UnscannedTables` covers unexamined cells, not just unread rows.** A table
+  lands there for three different reasons — never entered, entered and stopped
+  between rows, or entered and stopped between columns of its final row — and
+  only the first two leave a whole row unread. The caveat is therefore worded
+  around a cell the scan never examined, which is true of all three.
+- **A table with an unreadable row is not an `UnscannedTables` entry.** The two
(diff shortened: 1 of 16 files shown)


richlander:
## Adversarial review reconciliation — clean at `6e56a59f`

Two-model review: **MAI-Code** and **GPT-5.6 Sol**, each isolated in its own detached read-only worktree. Neither is my own family. Four rounds. Both returned **no blocking findings** at the final head.

This PR publishes numbers, so it invited a kind of scrutiny that source review cannot produce: both reviewers re-ran the probe, wrote their own scratch apps, and independently measured. That found real arithmetic bugs in work I had already convinced myself was correct. Every finding below was accepted; **nothing was dismissed.**

### Findings that changed the result

| --- | --- | --- | --- |
| 1 | GPT | Attribution table double-counted: `StringCell`/`UserStringCell`/`GuidCell` pass the **same instance** as both `Text` and `Preview` (136,299 cells); only `BlobCell` differs | Verified in `MetadataTableProjector.cs:796-847`. Now counted by object identity. `56695ee1` |
| 2 | GPT | `Align8(sum)` ≠ `sum(Align8(...))` — string sizing must round per object | Confirmed against measured allocations (len 1→24, 4→32, 16→56, 100→224 = `Align8(22 + 2*len)`). `48e3d215` |
| 3 | GPT | The drop test is blind to caches populated **before** the baseline | Demonstrated with a 10 MB warm-up-populated static cache and a 24-byte post-drop delta. My counter-claim was flatly wrong. `8037da9d` |
| 4 | MAI | Drop test cannot prove a universal property | Scratch-app counterexample. Claim scoped. `8037da9d` |
| 5 | both | Tiered-JIT explanation for allocated variance was a rationalization | Both tested tiering, tiered PGO, QuickJit, R2R, call-counting delays, warm-up repeats — no effect. Cause now stated as **unknown**. `8037da9d` |
| 6 | GPT | Retained drift is KB-scale, not the ~48 bytes I documented | Table now carries observed ranges. `8037da9d` |
| 7 | MAI | `mdi` defaults to 4096 rows — the 19.8 MB row is a **real consumer**, not hypothetical | Verified `MdiCommand.cs:45`. "A scenario nobody runs" withdrawn. `56695ee1` |
| 8 | GPT | "Largest component" and the browser-infeasibility claim were unproven | Both withdrawn. `56695ee1`, `8037da9d` |
| 9 | MAI | "byte-identical across runs" was false | Withdrawn. `56695ee1` |
| 10 | GPT | Span "cannot outlive the stack frame" is factually false | Corrected to heap-field and `await`/`yield` restrictions. `6e56a59f` |

Two claims I found myself: the all-tables `window 100 @ row 40000` scenario was misleading (only `MethodDef` and `Param` reach that depth, so it looked cheap for the wrong reason), and "deep paging costs the same as row 1" is false — it is slightly *cheaper* (60,880 vs 63,760 bytes) because those rows carry shorter names.

### Verified, not merely asserted

- **Sizing rule.** GPT computed `Align8(22 + 2*Length)` over the distinct objects and separately measured actual allocation by cloning every string: **32,217,048 both ways.**
- **Retention attribution.** The drop test reads 74.7 MB with the projection reachable and 0.0 MB once it is unreachable with the `PEReader` still live, `WeakReference` confirming collection. Scoped in the doc to what it can actually see.
- **Allocated is observer-dependent.** Three observers each saw a different *stable* value (85,322,584 / 86,306,824 / 86,310,024). Reported as a range with the cause stated as unknown rather than guessed at.

### What the measurement concludes

No allocation redesign is justified for the explorer. A full projection retains 74.7 MB managed for a 15.3 MB image; the windowed, table-selected path an explorer actually uses retains **0.1 MB**, at any depth into the table. `ReadOnlySpan<T>` is specifically wrong for a retained graph, but `ReadOnlyMemory<T>`, struct cells, lazy formatting, and flyweights are explicitly left open rather than implicitly closed off.

### Validation

`dotnet build dotnet-inspect.slnx -c Release` — succeeded, 0 errors (4 pre-existing warnings). No product code changed, so no test run is implicated. Markdown self-checked (MD009/MD012/MD022/MD047); `npx` is unavailable in this environment.

CI reports no checks by design — `ci.yml` triggers only on PRs targeting `main`, so a stacked PR stays unchecked until its base lands.

Ready to merge

PR #3355, opened 2026-07-28
Add an opt-in taste lens folding `(long)N` to the `NL` long literal

### LINKED ISSUE #3347: Decompiler taste lens: fold `conv.i8` of an int constant to an `NL` long literal (`(long)10`->`10L`) (#3129 raise-completeness follow-up)
## Summary

A `long`-typed constant that csc emits as `ldc.i4.s N; conv.i8` renders as a cast
`(long)N` rather than the idiomatic `NL` long literal. This is a
**raise-completeness** gap, not a fidelity gap: the render is already `Full`
fidelity (opcode-faithful; it round-trips), yet it is not *fully raised*. (Full
fidelity ≠ fully raised — see the documentation point in #3346, which asks for
this distinction to be made explicit in the design docs.)

## Observed vs. desired

Reference witness `CfgSampleClass.InlineArraySpanTernaryConditionValue` renders
today:

```csharp
public static long InlineArraySpanTernaryConditionValue(object a, object b)
    => (AnyObjectSpan([a, b]) ? ((long)10) : ((long)20)) + Environment.TickCount64;
```

Fully raised (idiomatic) endpoint:

```csharp
    => (AnyObjectSpan([a, b]) ? 10L : 20L) + Environment.TickCount64;
```

## IL evidence

```il
IL_002B: ldc.i4.s  20
IL_002D: conv.i8
IL_0030: ldc.i4.s  10
IL_0032: conv.i8
```

csc emits `ldc.i4; conv.i8` for small `long` literals (it does **not** use
`ldc.i8` for these), so `10L` and `(long)10` produce identical IL. The fold is
therefore opcode-neutral **for csc output**.

## Why this is a taste lens, not a default (opcode-fidelity caveat)

A hand-authored or non-csc assembly can encode the same value as `ldc.i8 <small>`,
which is a **distinct opcode** from `ldc.i4; conv.i8`. A blanket default fold of
`(long)N` → `NL` would then risk byte-divergence against that rare shape — the
opcode-fidelity trap (Valid + Correct yet not opcode-faithful; cf. #3127). So this
should be scoped as an **opt-in taste lens** with compile-back verification, and it
must keep the `(long)N` cast when the source opcode was actually `ldc.i8` (i.e.
only fold when the IL was `ldc.i4(.s) N; conv.i8`).

The `NL` / `UL` suffix machinery already exists — `UnsignedConstantText` in
`src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.Numerics.cs` renders `u` /
`UL` suffixes for constant-arithmetic operands — so the printer side is mostly a
matter of folding `Convert(→Int64, Int32 constant)` into a single `Int64` constant
and reusing the suffix path.

## Acceptance

- Opt-in taste lens (not on by default); folds only when the source IL was
  `ldc.i4(.s) N; conv.i8`, preserving `(long)N` for a genuine `ldc.i8` source.
- Compile-back verification that the folded `NL` output round-trips.
- Reference witness `InlineArraySpanTernaryConditionValue`.

Discovery context: found while reviewing the params-`ReadOnlySpan<T>` witnesses in
#3230 / PR #3343 (S4 of #3129). Orthogonal to inline-array raising.

richlander:
- Fixes/advances #3347
- Changes: adds an opt-in, non-default C# taste lens that renders a `long`
  constant csc emits as `ldc.i4(.s) N; conv.i8` as the idiomatic `NL` literal
  (`(long)10` → `10L`) instead of the `(long)N` cast. Default output is
  unchanged.
- Card revision: `67d261f2`

## Change

**Conclusion:** **REVIEW** — the fold is structurally guarded (it can only match
`Convert(→Int64, Int32 Constant)`, never a genuine `ldc.i8`) and every folded
specimen compiles back `Exact`, but the lens records no `style-lens.*` decision,
so a byte-divergent-classified render is invisible to `-S "Applied Taste"` and
still gets interleaved IL in Annotated Source. See the review comment on this PR.

### Benchmark target

Benchmark target: this repository's decompiler fixture assembly,
`artifacts/bin/ILInspector.Decompiler.Tests/release/ILInspector.Decompiler.Tests.dll`,
built at head `67d261f2`. The reference witness is the one #3347 names,
`CfgSampleClass.InlineArraySpanTernaryConditionValue`. The **same** fixture
binary is inspected by both the base and head tool builds, so the only variable
across Before/After is `dotnet-inspect` itself.

dotnet-inspect command:

```bash
dotnet-inspect member \
  --library artifacts/bin/ILInspector.Decompiler.Tests/release/ILInspector.Decompiler.Tests.dll \
  CfgSampleClass -m InlineArraySpanTernaryConditionValue --all \
  -S "Decompiled Source"
```

The lens is off unless requested. The After (lens on) render below was taken in
a directory holding:

```ini
root = true
dotnet_inspect_style_prefer_long_literal_suffix = true
```

### Original source

Acquired with `-S "Original Source"` (SourceLink-backed C#) at head:

```csharp
public static long InlineArraySpanTernaryConditionValue(object a, object b)
    => (AnyObjectSpan([a, b]) ? 10L : 20L) + System.Environment.TickCount64;
```

The IL the two `long` constants compile to (`-S "IL"`), which is the shape the
fold keys on:

```il
IL_0029: brtrue.s     IL_0030
IL_002B: ldc.i4.s     20
IL_002D: conv.i8
IL_002E: br.s         IL_0033
IL_0030: ldc.i4.s     10
IL_0032: conv.i8
```

### Before

`-S "Decompiled Source"` at base `21c00743`:

```csharp
public static long InlineArraySpanTernaryConditionValue(object a, object b) => (AnyObjectSpan([a, b]) ? ((long)10) : ((long)20)) + Environment.TickCount64;
```

- Valid: True
- Correct: True
- IL fidelity: True — `-S "Fidelity Causes"` reports `No fidelity causes;
  decompiler fidelity is Full.`
- Taste applied: None (`-S "Applied Taste"`: no recorded style choices)
- Commit: `21c00743`

Already `Full` fidelity and opcode-faithful; the gap is raise-completeness, not
fidelity — `(long)10` is the cast spelling of a value C# spells `10L`.

### After

`-S "Decompiled Source"` at head `67d261f2` with
`dotnet_inspect_style_prefer_long_literal_suffix = true`:

```csharp
public static long InlineArraySpanTernaryConditionValue(object a, object b) => (AnyObjectSpan([a, b]) ? 10L : 20L) + Environment.TickCount64;
```

- Valid: True
- Correct: True
- IL fidelity: True for csc-emitted shapes — all nine folded specimens compile
  back `Exact` both lens-on and lens-off with **identical** recompiled opcodes,
  measured through the product harness (`FidelityCheck.EvaluateTargets` in
  `LongLiteralFoldTests.FoldedOutput_CompilesBackExactly`). This witness itself
  is not in the compile-back set (inline-array body); `-S "Fidelity Causes"`
  reports `Full` for it at head with the lens on.
- Taste applied: **None recorded — this is a defect, not a property.** The knob
  is registered `ByteDivergent = true`, and the two existing byte-divergent
  lenses emit a `style-lens.*` `DecompilerDecision` so `-S "Applied Taste"`
  can account for them; this lens emits none, so the section renders `No
  recorded style choices were applied to this member.` on the render above.
  A consequence follows in Annotated Source, which suppresses interleaved IL
  only when an applied byte-divergent lens is recorded: it renders the folded
  `10L` text *and* the original `ldc.i4.s 10; conv.i8` listing side by side with
  no taste annotation on the signature.
- Commit: `67d261f2`

The default view at head is byte-identical to Before — the opt-in contract:

```csharp
public static long InlineArraySpanTernaryConditionValue(object a, object b) => (AnyObjectSpan([a, b]) ? ((long)10) : ((long)20)) + Environment.TickCount64;
```

### Fully raised

The After (lens-on) decompilation is in the fully raised state: it is textually
identical to the authored original source expression, `(AnyObjectSpan([a, b]) ?
10L : 20L) + Environment.TickCount64`.

The default view deliberately stays at `(long)N`. That is the byte-faithful
default, not an unraised remainder: the fully raised endpoint here is reachable
only behind the opt-in lens, because a non-csc assembly can encode the same
value as `ldc.i8 <small>` — a distinct opcode that must keep the cast spelling.

## Evidence

Release build, base `21c00743` (Baseline) vs head `67d261f2` (Head). Both trees
built with `dotnet build dotnet-inspect.slnx -c Release` before the runs, so no
fixture assembly is missing.

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `LongLiteralFoldTests`: not present (added here) | `LongLiteralFoldTests`: 29 passed, 0 failed |
| Decompiler fast suite (`-trait- Speed=Slow`) | 4017 total, 0 failed | 4045 total, 0 failed |
| Decompiler `Area=Corpus` | 83 total, 0 failed | 83 total, 0 failed |
| Reduced fixture validity | not applicable (no raising change; printer-only fold) | not applicable |
| Reduced fixture fidelity | 9 specimens `Exact` (lens absent) | 9 specimens `Exact` lens-on **and** lens-off, identical recompiled opcodes |
| Real witness | `CfgSampleClass::InlineArraySpanTernaryConditionValue` renders `(long)10` | renders `10L` with the lens on; byte-identical to Baseline with the lens off |

No Baseline failures. An earlier draft of this description reported "three
pre-existing `GeneratedFixtureCatalogTests.*` failures"; that was an artifact of
running the suite after building only the test project (a fixture binary was
missing). After a full `dotnet-inspect.slnx` build, both ends are 0 failed.

## Decompiler quality

**Conclusion:** **PASS** — the corpus renders with `PrinterOptions.Default`, in
which `PreferLongLiteralSuffix` is `false`, and every touched printer path is
gated on that flag, so no corpus metric can move. Held up by measurement rather
than assertion on two sides: `Area=Corpus` is 83/0 failed at both ends, and
`LongLiteralFoldTests.Default_IsUnchanged` pins the exact default text of all
eleven fixture methods plus the reference witness
(`ReferenceWitness_Default_IsUnchanged`) against `PrinterOptions.Default` *and*
against "no options at all".

The quick-gate and aggregate metric tables are omitted for that reason: with the
default render provably unchanged, there is no delta to report.

## Validation

```bash
  -class "ILInspector.Decompiler.Tests.LongLiteralFoldTests"
  -trait- "Speed=Slow"
  -trait "Area=Corpus"
```

Fixes: #3347
Nightshift-Order: /plan/pilot/order/long-literal

### DIFF
--- a/src/ILInspector.Decompiler.Tests/LongLiteralFoldFixture.cs
+++ b/src/ILInspector.Decompiler.Tests/LongLiteralFoldFixture.cs
@@ -1,0 +1,74 @@
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// Long-constant shapes for the opt-in
+/// <see cref="ILInspector.Decompiler.Pipeline.PrinterOptions.PreferLongLiteralSuffix"/>
+/// style lens (#3347). The lens folds the IR shape
+/// <c>Convert(→Int64, Int32 Constant)</c> — what csc's <c>ldc.i4(.s) N; conv.i8</c>
+/// imports as — into the idiomatic <c>NL</c> literal, and must leave a genuine
+/// <c>ldc.i8</c> (a bare <c>Int64</c> constant, no <c>Convert</c> over it) alone.
+///
+/// <para>The split between the two IL encodings is a property of csc's own literal
+/// emission, not of the decompiler: csc uses <c>ldc.i4</c>+<c>conv.i8</c> for any
+/// <c>long</c> constant whose value fits in an <see cref="int"/> and <c>ldc.i8</c>
+/// for one that does not. That is why every method below is authored as a plain
+/// C# constant — the compiler, not the fixture, chooses the opcode, so the
+/// fixture is a real compiled canary for the shape rather than an assumption
+/// about it. The value at the int/long boundary is covered from both sides
+/// (<see cref="IntMaxValue"/> is the last <c>ldc.i4</c>,
+/// <see cref="JustPastIntMaxValue"/> the first <c>ldc.i8</c>), and the small
+/// <c>ldc.i8</c> case C# cannot author at all is pinned synthetically in
+/// <c>LongLiteralFoldTests</c>.</para>
+/// </summary>
+public static class LongLiteralFoldFixture
+{
+    // --- conv.i8 sources: the lens folds these ---
+
+    // The reference witness's shape (CfgSampleClass.InlineArraySpanTernaryConditionValue),
+    // reduced: two `ldc.i4.s N; conv.i8` arms feeding an `add`, which is what keeps the
+    // join a real `?:` — a bare `return c ? 10L : 20L;` compiles to two returns and
+    // never reaches the conditional-arm seam at all.
+    public static long TernaryArms(bool c, long tail) => (c ? 10L : 20L) + tail;
+
+    // Return position: the value flows through the return sink's coercion, not an
+    // arm join. `ldc.i4.s 42; conv.i8; ret`.
+    public static long SmallReturn() => 42L;
+
+    // Argument position: the coercion sink is the parameter type, and the folded
+    // literal has to stay a well-formed argument.
+    public static long SmallArgument() => Consume(7L);
+
+    // Binary operand: exercises precedence. A cast operand and a literal operand
+    // report different precedences, so the fold must carry its own rather than
+    // inherit the cast's.
+    public static long BinaryOperand(long x) => x * 3L;
+
+    // Negative literal operand — the fold's text starts with a unary `-`, so this is
+    // the precedence case a bare Primary claim would get wrong.
+    public static long NegativeBinaryOperand(long x) => x * -1L;
+
+    // Boundary and sign coverage, all still `ldc.i4*; conv.i8`.
+    public static long Zero() => 0L;
+
+    public static long MinusOne() => -1L;
+
+    public static long IntMinValue() => int.MinValue;
+
+    // The largest long constant csc still encodes as `ldc.i4`.
+    public static long IntMaxValue() => int.MaxValue;
+
+    // --- genuine ldc.i8 sources: the lens must leave these exactly as they are ---
+
+    // One past int.MaxValue: the first value csc encodes as `ldc.i8`. Arrives at the
+    // printer as a bare Int64 Constant, so it cannot match the fold.
+    public static long JustPastIntMaxValue() => 2147483648L;
+
+    // A comfortably large `ldc.i8`, the issue's own example.
+    public static long LargeReturn() => 5_000_000_000L;
+
+    // An `ldc.i8` in the same ternary-arm position the fold fires in, so the close
+    // negative is pinned at the seam and not only at a return.
+    public static long LargeTernaryArms(bool c, long tail) => (c ? 5_000_000_000L : 6_000_000_000L) + tail;
+
+    public static long Consume(long value) => value;
+}
--- a/src/ILInspector.Decompiler.Tests/LongLiteralFoldTests.cs
+++ b/src/ILInspector.Decompiler.Tests/LongLiteralFoldTests.cs
@@ -1,0 +1,316 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection.PortableExecutable;
+
+using ILInspector.Decompiler;
+using ILInspector.Decompiler.Pipeline;
+using ILInspector.DecompilerHarness;
+using ILInspector.Metadata;
+
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// The opt-in byte-divergent style lens
+/// <see cref="PrinterOptions.PreferLongLiteralSuffix"/> (#3347): a <c>long</c>
+/// constant csc emits as <c>ldc.i4(.s) N; conv.i8</c> renders as the idiomatic
+/// <c>NL</c> literal instead of the <c>(long)N</c> cast the default view spells.
+/// This is a raise-completeness gap, not a fidelity gap — the cast is already
+/// opcode-faithful and round-trips; it is simply not fully raised.
+///
+/// <para>Three claims are pinned here, in the order they matter:</para>
+/// <list type="number">
+/// <item><description>
+/// <b>Default unchanged.</b> Every fixture method and the reference witness
+/// <c>CfgSampleClass.InlineArraySpanTernaryConditionValue</c> render exactly
+/// today's text with the knob off — the lens is opt-in and the shipped view stays
+/// byte-faithful.
+/// </description></item>
+/// <item><description>
+/// <b>The fold fires on exactly the <c>conv.i8</c> shape.</b> With the knob on, a
+/// <c>Convert(→Int64, Int32 Constant)</c> becomes <c>NL</c> at a ternary arm, a
+/// return, an argument, and a binary operand (including a negative literal, whose
+/// unary precedence the fold has to carry itself).
+/// </description></item>
+/// <item><description>
+/// <b>The close negative: a genuine <c>ldc.i8</c> is untouched.</b> That shape
+/// reaches the printer as a bare <c>Int64</c> <see cref="Constant"/> with no
+/// <see cref="Convert"/> over it, so it cannot match the fold. csc will not emit
+/// <c>ldc.i8</c> for a small value, so the small-<c>ldc.i8</c> case is pinned on a
+/// hand-built IR function, and the compiled fixture covers the large values csc
+/// does encode that way (including one in the very ternary-arm seam the fold fires
+/// in). Both must render identically knob-on and knob-off.
+/// </description></item>
+/// </list>
+///
+/// <para>The lens is <see cref="StyleOptionDescriptor.ByteDivergent"/>, so it is
+/// excluded from <see cref="ByteNeutralityGateTests"/> by construction. Its
+/// fidelity claim is narrower and is proven directly: because the fold is
+/// opcode-neutral <em>for csc output</em>, the knob-on render of every folded
+/// specimen must still compile back <c>Exact</c> — the same anchor the knob-off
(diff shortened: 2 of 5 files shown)


richlander:
✅ Adversarial review clear — two independent reviews from two different models on the same final head (`67d261f2`), each 1 round, zero blocking findings.

| Role | Model | Rounds | Verdict |
| --- | --- | --- | --- |
| Builder | claude-opus-5 | — | did not review its own work |
| Reviewer 1 | gpt-5.6-sol | 1 | CLEAN |
| Reviewer 2 | gemini-3.1-pro-preview | 1 | CLEAN |

Each reviewer ran read-only in its own detached worktree at the exact head. (Transparency: reviewer 2's sign-off line self-misnamed its model; the dispatched, harness-recorded model was `gemini-3.1-pro-preview` — reported as observed.)

Ready to merge

richlander:
## Review — description now on the house template, plus one product finding

**Head reviewed:** `67d261f236d57be1ec078e49b8632417c6e27613` (base
`21c007435eca34e0887705a39ee86fbddfc0fd85`). Reviewed in a detached review
worktree; all scratch under `/tmp/`.

### 1. Description did not follow `docs/templates/decompiler-pr.md` (fixed)

The description was prose only. A raise PR must keep **Original source →
Before → After → Fully raised**, each render acquired verbatim from
`dotnet-inspect` rather than paraphrased, with the per-block Valid / Correct /
IL fidelity / Taste applied / Commit verdicts. I have rewritten the description
with real renders taken from base and head tool builds against the same fixture
binary.

Rewriting it surfaced two things prose had hidden:

- **The evidence numbers were wrong.** The claimed "three pre-existing
  `GeneratedFixtureCatalogTests.*` `RoundTrip` failures" do not exist. They were
  an artifact of running the suite after building only the test project, which
  leaves fixture binaries missing. After a full `dotnet build dotnet-inspect.slnx
  -c Release`, the fast suite is **4017 total / 0 failed at base** and **4045
  total / 0 failed at head**. Corrected in the description.
- **The "Taste applied" verdict cannot be populated from product output** —
  finding 2.

### 2. Finding: the lens records no `style-lens.*` decision, so it is invisible to the taste account

The knob is registered `ByteDivergent = true` / `Tier = Lens`, but nothing
appends a `DecompilerDecision` when the fold fires. The two existing
byte-divergent lenses do (`CSharpPrinter.RaiseWithStyleLenses`, rule ids
`style-lens.prefer-conditional-return` and `style-lens.prefer-branchless-boolean`).

Reproduced at head, same command, same fixture assembly, differing only in the
`.dotnet-inspectconfig` key:

```text
$ dotnet-inspect member --library ...ILInspector.Decompiler.Tests.dll \
    CfgSampleClass -m InlineArraySpanTernaryConditionValue --all -S "Applied Taste"
# with dotnet_inspect_style_prefer_long_literal_suffix = true
## Applied Taste
No recorded style choices were applied to this member.
```

Control, proving the section works in this exact setup:

```text
$ dotnet-inspect member --library ...ILInspector.Decompiler.Tests.dll \
    PreferBranchlessBooleanSpecimen -m AndTailGuard --all -S "Applied Taste"
# with dotnet_inspect_style_prefer_branchless_boolean = true
| Rule | Fidelity | Subject | Detail |
| style-lens.prefer-branchless-boolean | byte-divergent | AndTailGuard | Folded a guarded boolean return ... |
```

This is not cosmetic. `AGENTS.md` makes the taste account the mechanism by which
a reader learns a render is no longer opcode-faithful ("A byte-divergent style
lens listed there is exactly why IL fidelity can be False while Valid/Correct
are True — surface it here instead of leaving the reader to infer it from
prose"), and the missing decision has an observable behavioral consequence
beyond the section: Annotated Source suppresses interleaved IL only when an
applied byte-divergent lens is recorded
(`AnnotatedSourceTasteTests.ByteDivergentLens_WhenApplied_SuppressesInterleavedIl`).
At head with the lens on, the witness renders the folded text *and* the original
opcodes together, with no annotation on the signature:

```text
return (AnyObjectSpan([a, b]) ? 10L : 20L) + Environment.TickCount64;  // safety.callee(unsafe); ...
    // IL_0030: ldc.i4.s     10
    // IL_0032: conv.i8
```

Two defensible resolutions; I do not think "leave it" is one of them:

1. **Record the decision.** The fold happens at print time, not in an IR pass,
   so this needs the printer to report that it fired and `WithAppliedLenses` to
   pick it up — real plumbing, but it puts the knob on the same footing as its
   two siblings.
2. **Argue the classification down.** If the position is that on csc output the
   fold is opcode-neutral (which the compile-back evidence in this PR does show:
   nine specimens `Exact` both ways with identical recompiled opcodes), then say
   why it is nonetheless `ByteDivergent` yet exempt from the taste account. I do
   not find that convincing — `ldc.i8 <small>` is precisely why the knob is a
   lens, and a reader inspecting a non-csc assembly is the person the account
   exists for.

Either way, the gap deserves a gate. `ByteNeutralityGateTests.ByteDivergentKnobs_AreExcludedAndBehaviorGated` already enumerates the byte-divergent set from the
catalog; asserting that each such knob records a decision when it rewrites would
have caught this at authoring time, and would catch the next lens too.

### What I verified and did not contest

- The opcode-fidelity guard is structural, as claimed: `TryLongLiteralText`
  matches only `Convert{IsChecked:false, IsUnsigned:false, Target:Int64,
  Operand:Constant{Int32}}`, so a bare `Int64` `Constant` (a genuine `ldc.i8`)
  cannot reach it. Pinned three ways, including a non-vacuity partner test — good.
- Compile-back is measured through the product harness
  (`FidelityCheck.EvaluateTargets`), not a harness-side reimplementation, and
  asserts opcode identity between the two renders rather than "both Exact"
  against possibly different originals.
- The default view is unchanged, pinned as exact text against both
  `PrinterOptions.Default` and "no options at all"; `Area=Corpus` is 83/0 at
  both ends.
- Precedence handling for the negative literal (`x * -1L` staying bare at a
  Unary demand) is correct and covered.

**Verdict: REVIEW — not ready to merge.** Finding 2 needs a resolution (fix or a
recorded, gated rationale) before this lands.

PR #3358, opened 2026-07-28
Raise inline-Current enumerator loops to foreach (#3164)

### LINKED ISSUE #3164: Decompiler: recover enumerator pattern in JsonElement.DeepEquals to foreach
## Defect

Enumeration is not recovered to `foreach`. The array path emits an explicit
enumerator pattern in `try/finally`:

```csharp
arrayEnumerator2 = element2.EnumerateArray();
V_6 = element1.EnumerateArray().GetEnumerator();
try
{
    while (V_6.MoveNext())
    {
        arrayEnumerator2.MoveNext();
        if (!DeepEquals(V_6.Current, arrayEnumerator2.Current)) return false;
    }
}
finally
{
    V_6.Dispose();
}
```

and the object path is an open-coded enumerator loop (see #3159).

## Expected

Recover the canonical `foreach` shape (`GetEnumerator`/`MoveNext`/`Current`/
`Dispose`-in-`finally`) to `foreach` where the pattern matches.

## Repro

```bash
dotnet-inspect System.Text.Json.JsonElement.DeepEquals -S "Decom*"
```

`Library: System.Text.Json | Source: Platform | TFM: net11.0`

Part of #3159.

richlander:
- Fixes #3164
- Changes the `JsonElement.DeepEquals` `Array` arm (and any single-use-`Current` enumerator loop) to render as `foreach` instead of an open-coded `using`/`while` enumerator pattern.
- Card revision: `8a100bef572980421dfaff2f01cabf1066e87aed`

## Change

**Conclusion:** **PASS** — the enumerator is advanced only by the loop condition, so the single inline `e.Current` read is invariant across the body; hoisting it to a fresh `foreach` variable is the exact inverse of the earlier single-use inline and re-lowers opcode-identically, witnessed by a compile-back **Exact** fixture.

### Benchmark target

Benchmark target: `System.Text.Json` (platform, net11.0)

dotnet-inspect command:

```bash
dotnet-inspect member System.Text.Json.JsonElement.DeepEquals -S "Decompiled Source"
```

### Original source

```csharp
case JsonValueKind.Array:
    if (element1.GetArrayLength() != element2.GetArrayLength())
    {
        return false;
    }

    ArrayEnumerator arrayEnumerator2 = element2.EnumerateArray();
    foreach (JsonElement e1 in element1.EnumerateArray())
    {
        bool success = arrayEnumerator2.MoveNext();
        Debug.Assert(success, "enumerators must have matching length");

        if (!DeepEquals(e1, arrayEnumerator2.Current))
        {
            return false;
        }
    }

    Debug.Assert(!arrayEnumerator2.MoveNext());
    return true;
```

The authored source uses a `foreach` over `element1.EnumerateArray()` while advancing a second enumerator (`arrayEnumerator2`) manually in parallel.

### Before

```csharp
arrayEnumerator2 = element2.EnumerateArray();
using (ArrayEnumerator V_6 = element1.EnumerateArray().GetEnumerator())
{
    while (V_6.MoveNext())
    {
        arrayEnumerator2.MoveNext();
        if (!DeepEquals(V_6.Current, arrayEnumerator2.Current))
        {
            return false;
        }
    }
}
return true;
```

- Valid: True
- Correct: True
- IL fidelity: True (the `using`/`while` enumerator pattern is opcode-faithful; it is simply lower-level than the source's `foreach`)
- Taste applied: None (`-S "Applied Taste"` reports no recorded style choices)
- Commit: `21c007435eca34e0887705a39ee86fbddfc0fd85`

The pre-change render exposes the compiler's open-coded enumerator scaffolding (`using`/`while`/`V_6.Current`) instead of the source's `foreach`. The iteration variable was folded into its one use by `ExpressionInliningPass` before the foreach matcher ran, so the matcher's "first body statement is `item = e.Current`" invariant no longer held and the loop was left un-raised.

### After

```csharp
arrayEnumerator2 = element2.EnumerateArray();
foreach (JsonElement V_11 in element1.EnumerateArray())
{
    arrayEnumerator2.MoveNext();
    if (!DeepEquals(V_11, arrayEnumerator2.Current))
    {
        return false;
    }
}
return true;
```

- Valid: True
- Correct: True
- IL fidelity: True — witnessed by the pinned `CfgSampleClass.ForeachSingleUseWithParallelEnumerator` fixture (compile-back **Exact**; see below). DeepEquals itself is not directly compile-back-checkable (internal platform surface, per #3197), but the raised shape is proven opcode-exact by the fixture.
- Taste applied: None
- Commit: `8a100bef572980421dfaff2f01cabf1066e87aed`

The After render reproduces the source's `foreach (… in element1.EnumerateArray())` with the parallel `arrayEnumerator2.MoveNext()` preserved. The synthesized `V_11` loop variable (vs. the original PDB name `e1`) reflects that the single-use variable had been inlined away — the pass allocates a fresh `foreach` variable via `IrFunction.AddLocal` and rebinds the one inline `Current` read to it. Variable names do not affect IL. The `Object` arm's `while (objectEnumerator1.MoveNext())` is a genuine source-named, non-`using` open-coded loop and is correctly left untouched (tracked separately under #3165).

**Witness fixture** — `CfgSampleClass.ForeachSingleUseWithParallelEnumerator` (a single-use `foreach` iteration variable plus a parallel manually-advanced enumerator, the shape that triggers the inline) renders through the new path as:

```csharp
List<int>.Enumerator other = b.GetEnumerator();
foreach (int V_3 in a)
{
    other.MoveNext();
    // ...
}
```

It is **compile-back Exact** — pinned in `FidelityGateTests.PinnedExact` and absent from the entire fidelity diff set on head.

### Fully raised

The After decompilation is in the fully raised state for the `Array` arm's enumerator loop.

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | n/a (new tests) | `ForeachStatementPassTests`: 49 passed |
| Decompiler fast suite | 328 RoundTrip passed | 328 RoundTrip passed; `Area=Pass`: 1393 passed |
| Reduced fixture fidelity | not applicable (fixture new) | `ForeachSingleUseWithParallelEnumerator`: compile-back **Exact** |
| Real witness | `JsonElement::DeepEquals` Array arm renders `using (ArrayEnumerator V_6 …) { while (…) }` | `JsonElement::DeepEquals` Array arm renders `foreach (JsonElement V_11 in element1.EnumerateArray())` |

The change is a foreach-only matcher extension in `ForeachStatementPass`: it is a structural no-op on any method without an enumerator `while`-loop.

Pre-existing baseline fidelity failures on this workstation, unchanged by this PR (all non-`foreach` fixtures, so a foreach-only pass is a no-op on each; they pass in CI's environment): `MakeConsumerWithTwoLeadingArgs` (OpcodeDiff, recompiler temp-spill `stloc`/`ldloc`), `ManualConstantOnlyComparisonFactory` (OpcodeDiff, expr-tree factory temp spill), `StatementBodyLambdaInsideIf` (OperandDiff, csc lambda-ordinal drift), `TupleRest` (RecompileFail CS0246, environmental). The new fixture is absent from the entire diff set; CI's fidelity gate is authoritative.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -filter "/*/*/ForeachStatementPassTests/*"
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -trait "Area=RoundTrip" -trait- "Speed=Slow"
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -2106,6 +2106,30 @@ public class CfgSampleClass
         return sum;
     }
 
+    // Compiler foreach whose iteration variable is used exactly once, next to a
+    // second enumerator advanced manually in the loop body. csc hoists `x =
+    // e.Current` to the loop top, but because `x` is single-use and (in the
+    // absence of a source name) inlinable, ExpressionInliningPass can fold that
+    // store into its one use before ForeachStatementPass runs — leaving the
+    // hidden enumerator referenced only by MoveNext and one inline `e.Current`.
+    // This is the shape System.Text.Json.JsonElement.DeepEquals exhibits on its
+    // Array arm (#3164): the pass must still recover the foreach by rebinding the
+    // inline Current to a fresh iteration variable.
+    public static bool ForeachSingleUseWithParallelEnumerator(
+        System.Collections.Generic.List<int> a, System.Collections.Generic.List<int> b)
+    {
+        System.Collections.Generic.List<int>.Enumerator other = b.GetEnumerator();
+        foreach (int x in a)
+        {
+            other.MoveNext();
+            if (x != other.Current)
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
     public ref struct RefStructResource
     {
         public RefStructResource(int value) => Value = value;
--- a/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
@@ -335,6 +335,18 @@ public class FidelityGateTests
         // JsonElement structs through one such temp but cannot itself be fed to the
         // compile-back oracle (internal System.Text.Json surface).
         "SwapStructPair",
+        // #3164: the inline-Current foreach. On its Array arm DeepEquals runs a
+        // compiler `foreach` whose single-use iteration variable csc hoists as
+        // `item = e.Current`; ExpressionInliningPass folds that store into its one
+        // use before ForeachStatementPass runs, so the hidden enumerator survives
+        // referenced only by MoveNext and one inline `e.Current`. The pass rebinds
+        // that inline read to a fresh foreach variable. Because the raised
+        // `foreach` re-lowers to csc's original hoist, the transform is
+        // opcode-exact — this pins it. Mirrors System.Text.Json.JsonElement.
+        // DeepEquals, whose Array arm iterates one enumerator by `foreach` while a
+        // second is advanced manually, but which cannot itself be fed to the
+        // compile-back oracle (internal System.Text.Json surface, #3197).
+        "ForeachSingleUseWithParallelEnumerator",
         // The compile-back oracle replays the fixture's runtime-async feature,
         // so these methods must retain the same lowering rather than merely
         // remaining recompilable.
--- a/src/ILInspector.Decompiler.Tests/ForeachStatementPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ForeachStatementPassTests.cs
@@ -570,6 +570,30 @@ public class ForeachStatementPassTests
     }
 
     [Fact]
+    public void InlineCurrentForeach_ReadInsideTryAtBodyTop_StaysUsingWhile()
+    {
+        // Exception regions emit no IL of their own, so a hand-written
+        // `try { if (e.Current == 0) ... } catch { ... }` at the loop-body top has
+        // get_Current as the first *executable* instruction (offset == body top) —
+        // the offset check alone would accept it. But the read is protected by the
+        // handler; hoisting it to the foreach header moves get_Current out of the
+        // try, so a throwing enumerator would escape the loop instead of being
+        // caught. ReadOriginatesAtLoopBodyTop rejects a read wrapped in any
+        // try/catch/finally within the body. (A real foreach whose iteration
+        // variable is used inside a try must store it — the stack-cached inline
+        // form cannot cross the region boundary — so this declines no real
+        // foreach.)
+        var function = BuildInlineCurrentReadInTryLoop();
+
+        new ForeachStatementPass().Run(function, PassContext.None);
+        function.CheckInvariant();
+
+        Assert.Empty(function.Descendants.OfType<ForeachStatement>());
+        Assert.Single(function.Descendants.OfType<UsingStatement>());
+        Assert.Single(function.Descendants.OfType<WhileLoop>());
+    }
+
+    [Fact]
     public void InlineCurrentForeach_UnconditionalCurrentReadInsideIfCondition_RaisesToForeach()
     {
         // Positive control for the guard: the same hidden-enumerator loop, but the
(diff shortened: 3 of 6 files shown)


richlander:
## Adversarial review reconciliation — final head `451cefc9`

Two-model adversarial review (Gemini 3.1 Pro + GPT-5.6), isolated exact-head worktrees, five rounds. Both **PASS** at the reviewed product head `f7e1dfcf`; both **ACK** the test-only follow-up `451cefc9`.

### Guard evolution (each round closed a real reviewer finding)

| Round | Finding (reviewer) | Fix |
| --- | --- | --- |
| 1 | Conditional `Current` read raised — a hand-written `if (index==0) return e.Current;` (ElementAt shape) is byte-identical after inlining to a compiler foreach and was wrongly raised, re-hoisting `get_Current` to run every iteration (GPT Medium; Gemini High throwing-enum repro). | Added `IsUnconditionalEveryIteration` structural guard. |
| 2 | Guard was fail-open (missed `??=` / switch-expr arms) and reorder-blind (`SideEffect(); if (e.Current==0)…` accepted). | Replaced with **offset provenance**: read raises only if its subtree carries the loop body's first IL offset. |
| 3 | Offset-min missed **erased** offsets (`NullCoalescingAssignmentPass` drops the `??=` null-check offset); and try/catch regions are pure PE metadata (no IL), so `try{x=e.Current;}catch{}` looked body-top and hoisted the read out of the handler (Gemini compiled throwing repro). | Switched to `Block.StartOffset` (import-stamped, immune to interior erasure); added exception-region rejection. |
| 4 | Nested loop at body top — `do{Consume(e.Current);}while(--r>0);` — read is first instr but executes N times/outer-iteration (GPT, compiled repro). | Generalized to `IsInsideNestedLoopOrExceptionRegion` (rejects nested `While`/`DoWhile`/`For`/`Foreach` + EH regions). Offset = ordering; structural walk = frequency + placement; neither alone suffices. |
| 5 | GPT: a **genuine** `foreach(x in items) do{Consume(x);}while(--r>0)` decompiles to the same lowered form, reading `Current` each inner iteration. | **Proven pre-existing, not a regression** — see below. |

The final guard `ReadOriginatesAtLoopBodyTop` combines (1) ordering — `MinOffset(read) == loopBody.StartOffset` (fail-closed on no offset) — and (2) frequency/placement — `!IsInsideNestedLoopOrExceptionRegion(...)`, plus `!IsInsideNestedFunction(...)` at the call site.

### Round-5 genuine-foreach finding → tracking issue #3363 (pre-existing, no regression)

GPT's round-5 case is real but **pre-existing on `main`**. I built `origin/main` (`21c00743`) in a clean worktree and decompiled the same DLL: the output is **byte-identical** to this PR's. Root cause is upstream in `ExpressionInliningPass.InlineOnce`, which sinks a single-use non-pure `get_Current` across a loop boundary with no loop-frequency check — after that pass a genuine foreach and a hand-written nested loop are **byte-identical IR**, so no foreach-pass decision can separate them. Declining (this PR's behavior) is the behavior-safe default and matches `main`. Filed as **#3363** with repro, root-cause pointer, and suggested fix. GPT reviewed the evidence and **downgraded to PASS**; agreed it is orthogonal to this PR.

### Test-only follow-up `451cefc9` (CI hygiene, both reviewers ACK'd)

CI's fast decompiler suite surfaced two failures that are direct, expected consequences of the change; both fixed test-side, no product change:

1. `SubstratePredicateCensusTests` — registered the two new ancestor-walk helpers (`IsInsideNestedFunction`, `IsInsideNestedLoopOrExceptionRegion`) in `ApprovedLocalPredicates` with rationales, mirroring existing `IsInsideCatchFilter` / `IsInsideRange` approvals.
2. `StructuringDiagnosticsTests.EarlyBreakFromInnerTry_RaisesWithoutLeaveGoto` — the change correctly raises the `AllOuterMatchInner` fixture's **inner** enumerator loop to `foreach` too (both are genuine single-use body-top foreaches), so the assertion now expects exactly two `foreach (` and no `while (`, keeping the Full-fidelity and no-`// leave` checks. Fidelity stays `Full`.

### Fidelity gate

Local fidelity-gate failures (`MakeConsumerWithTwoLeadingArgs`, `ManualConstantOnlyComparisonFactory`, `StatementBodyLambdaInsideIf`, `TupleRest`) are **environmental** (compile-back oracle drift under this workstation's .NET 11 preview.7) — reproduced **byte-identical on `origin/main`**, all non-`foreach` fixtures, and the new fixture `ForeachSingleUseWithParallelEnumerator` is **Exact** and absent from the diff set. CI's fidelity environment is authoritative.

**Verdicts at final head:** Gemini — PASS ("watertight, ship condition met") + ACK on `451cefc9`. GPT — PASS (after #3363 reconciliation) + ACK on `451cefc9`.

richlander:
Ready to merge

All CI checks green at `451cefc9` (build-net10, test, csharp-diff-smoke, changes); `MERGEABLE` / `CLEAN`. Both adversarial reviews (Gemini 3.1 Pro, GPT-5.6) PASS at the reviewed product head `f7e1dfcf` and ACK the test-only follow-up `451cefc9`; full round-by-round reconciliation posted above. The one remaining reviewer finding (round-5 genuine-foreach) is proven pre-existing on `main` and tracked in #3363 as non-blocking follow-up.

PR #3360, opened 2026-07-28
Settle the metadata lens on a section category, not a new flag

richlander:
Fifth in the #3341 stack, on top of #3352. Docs only -- no product code.

## The correction

The design doc proposed `library My.dll --table TypeDef`. That flag cannot exist: `--table` is already a presentation modifier ("render the selected section as a space-padded pretty table"), verified in `library --help` and in the output-shapes flag tables.

The deeper point is that no flag was needed. A metadata table **is** a section, addressed by name, so `-S` already reaches it. Tables become a `@Metadata` category registered the way `@Performance` is:

```csharp
.AddCategory(SectionCategoryNames.Metadata, MetadataTables.Sections)
```

Category membership is also what supplies progressive disclosure -- it is what keeps raw tables out of the default `-v:m` view, with no section-specific special-casing.

## What that collapses

`-S @Metadata --count` already renders a Section/Count table, so it **is** the table-stream overview. Verified against the existing category:

```console
$ dotnet-inspect library System.Text.Json.dll -S @Performance --count
| Section | Count |
| ------- | ----- |
| Performance: Boxing | 51 |
| Performance: Arrays | 48 |
```

Only the non-row parts of the image -- stream sizes, table-present bitmask, version string -- still need a `Metadata: Image` section.

Deep paging is dropped rather than designed. `--head`/`--tail` are documented as presentation limits, not row selectors, so a start-row flag would blend concepts in exactly the way the doc's own discipline section prohibits. Arbitrary-row access stays a library concern, served by `ProjectRow` and the row window from #3345 for the explorer host.

## The one new flag, and the family it belongs to

Heap addressing is a genuinely new currency: `#Strings:0x1a4` is not a section name. It gets a carrier, `--heap`, behaving like `--il-offset`.

That pattern was load-bearing but unnamed. `--il-offset` appears in **none** of the three flag families in `output-shapes.md` despite having its own case study there. This names a fourth family -- coordinate carriers -- and states the bar: a new currency, not merely a new thing to look at. The next flag now has a rule to follow instead of a precedent to copy.

## Open questions resolved

Lens home, heap surfacing, and table selection grammar (both `TypeDef` and `0x02`, since output already prints hex tokens).

Reworking `--row` from a printable-row selector into a row-id addressor is recorded as a known weak point and explicitly out of scope for this series.

## Validation

Markdown structural checks (MD009/MD012/MD022/MD047) clean on both files; `markdownlint` proper is unavailable locally (no Node). No behaviour claim, so no product build or test is implicated.

### DIFF
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -217,6 +217,11 @@ session-lifetime rule in
 
 ## Surface — the `metadata` table lens
 
+**Status: designed, not yet implemented.** Nothing in this section runs today;
+`-S @Metadata` currently fails with `Select value '@Metadata' not found.` The
+commands and outputs below are the intended surface, recorded so the
+implementation has a target.
+
 Each metadata table is **one section**, and the tables together form a section
 category, `@Metadata`, registered the same way `@Performance` is:
 
@@ -277,9 +282,16 @@ counts, so they remain a single `Metadata: Image` section.
 Progressive-disclosure discipline (see
 [progressive-disclosure.md](progressive-disclosure.md) and
 [section-model.md](section-model.md)): raw tables are **not** in the default
-`-v:m` view. Category membership is what keeps them out, exactly as it does for
-`@Performance`, so no section-specific special-casing is required. They are
-reached by explicit selection (`-S @Metadata` or `-S "Metadata: <Table>"`).
+`-v:m` view. Two separate mechanisms are involved, and they do different jobs:
+
+- Each metadata table descriptor sets `ExplicitOnly => true`. That is the
+  render gate — `SectionPipeline.IsRequested` returns `false` for an
+  `ExplicitOnly` entry unless it is explicitly included, so no verbosity level
+  auto-selects it. This is per-section configuration, the same as the
+  `@Performance` sections and the `--il-offset` coordinate sections.
+- The `@Metadata` category is the *selection and discovery* affordance. It lets
+  `-S @Metadata` name the whole group and gives `-D` something to list; it does
+  not by itself suppress anything.
 
 Heap **addressing** is the one place this lens does introduce a new currency: a
 heap coordinate such as `#Strings:0x1a4` is not a section name, so it needs a
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -217,34 +217,89 @@ session-lifetime rule in
 
 ## Surface — the `metadata` table lens
 
-The projection is exposed as its own table lens so it composes with the existing
-verbosity model and can be selected explicitly. It rides the `library` command
-(the assembly-oriented surface; note the deprecated `package X --metadata` alias
-already redirects there):
+Each metadata table is **one section**, and the tables together form a section
+category, `@Metadata`, registered the same way `@Performance` is:
+
+```csharp
+.AddCategory(SectionCategoryNames.Metadata, MetadataTables.Sections)
+```
+
+The lens therefore adds **no new shape flags**. Metadata tables introduce no new
+currency — they are sections, addressed by name — so they are reached with the
+existing selection vocabulary rather than a focused flag. It rides the `library`
+command (the assembly-oriented surface; note the deprecated `package X
+--metadata` alias already redirects there):
 
 ```bash
-# enumerate a single table
-dotnet-inspect library My.dll --table TypeDef
+# Document: every projected table
+dotnet-inspect library My.dll -S @Metadata
 
-# a single row with its columns and resolvable handle refs
-dotnet-inspect library My.dll --table MethodDef --row 3
+# Table: one metadata table
+dotnet-inspect library My.dll -S "Metadata: TypeRef"
+
+# Vector: one column of that table
+dotnet-inspect library My.dll -S "Metadata: TypeRef" --columns Name --tsv
+
+# Scalar: collapse to a row count
+dotnet-inspect library My.dll -S "Metadata: TypeDef" --count
+
+# a bounded window into a large table
+dotnet-inspect library My.dll -S "Metadata: MethodDef" --rows --head 20
 
 # structured, for tooling
-dotnet-inspect library My.dll --table TypeRef --jsonl
+dotnet-inspect library My.dll -S "Metadata: TypeRef" --jsonl
 ```
 
 This obeys the shape ladder in [output-shapes.md](output-shapes.md): a table is a
-**Table**, one column is a **Vector**, `--count` / one row is a **Scalar**. Each
-metadata table is one section; a `HandleRef` cell renders as its token +
-optional label in Markdown and as structured fields in JSON/JSONL.
+**Table**, one column is a **Vector**, and `--count` is a **Scalar**. A
+`HandleRef` cell renders as its token + optional label in Markdown and as
+structured fields in JSON/JSONL.
+
+Because `--count` reduces *each* selected section to a count, the category
+selection doubles as the table-stream overview, with no separate rendering path:
+
+```bash
+dotnet-inspect library My.dll -S @Metadata --count
+```
+
+```md
+| Section | Count |
+| ------- | ----- |
+| Metadata: TypeRef | 309 |
+| Metadata: TypeDef | 118 |
+| Metadata: MethodDef | 1204 |
+```
+
+The parts of the image that are not rows of a table — stream sizes, the
+table-present bitmask, the metadata version string — are not expressible as row
+counts, so they remain a single `Metadata: Image` section.
 
 Progressive-disclosure discipline (see
 [progressive-disclosure.md](progressive-disclosure.md) and
 [section-model.md](section-model.md)): raw tables are **not** in the default
-`-v:m` view. They are reached by explicit selection (`--table <Name>` or
-`-S`), following the same focused-flag-promotes-sections pattern that
-`library --il-offset` uses to add its coordinate-scoped sections. Heap dumps are
-a further explicit opt-in on top of that.
+`-v:m` view. Category membership is what keeps them out, exactly as it does for
+`@Performance`, so no section-specific special-casing is required. They are
+reached by explicit selection (`-S @Metadata` or `-S "Metadata: <Table>"`).
+
+Heap **addressing** is the one place this lens does introduce a new currency: a
+heap coordinate such as `#Strings:0x1a4` is not a section name, so it needs a
+carrier. `--heap` is that carrier, and it behaves like `--il-offset` — it makes a
+coordinate-scoped section available and discoverable only when present (see the
+coordinate-carrier family in [output-shapes.md](output-shapes.md)):
+
+```bash
+# bounded heap preview — just a section
+dotnet-inspect library My.dll -S "Metadata: #Strings"
+
+# one address — a coordinate
+dotnet-inspect library My.dll --heap "#Strings:0x1a4"
+```
+
+Deep paging into the middle of a table is deliberately **not** a CLI gesture.
+`--head`/`--tail` are presentation limits rather than row selectors, and a
+start-row flag would blend those concepts. Random access at an arbitrary row
+stays a library concern, served by `ProjectRow` and the row window for the
+explorer host.
 
 ## Safety
 
@@ -822,10 +877,19 @@ via the existing section pipeline and `OutputFormatter`.
 
 - **`HandleRef` shape.** Exact fields and whether `Display` is always populated
   or lazily resolved. Decide before code, since the explorer's value rests on it.
-- **Lens home.** A focused `--table <Name>` flag on `library`, a `-S` section
-  group ("Metadata Tables"), or a dedicated `metadata` command. Leaning toward a
-  `library` flag to reuse assembly acquisition and the section pipeline.
-- **Heap surfacing flags.** The exact opt-in gesture(s) for dumping string / blob
-  / user-string / guid heaps, and their bounded-preview defaults.
-- **Table selection grammar.** Whether tables are addressed by ECMA name
-  (`TypeDef`), by index (`0x02`), or both.
+- **Row addressing.** `--row` is currently a *printable-row* selector, not a
+  row-id addressor, so it cannot name `MethodDef[3]`; `--where` carries that
+  today. Reworking `--row` into a row-id addressor is a known weak point but is
+  deliberately out of scope for this series.
+
+Resolved:
+
+- **Lens home.** A `@Metadata` section category, not a focused flag and not a
+  dedicated command. Metadata tables introduce no new currency, so section
+  selection already addresses them. This also avoids a collision: `--table` is
+  already taken as a presentation modifier ("render as a pretty table").
+- **Heap surfacing flags.** Bounded per-heap previews are ordinary sections
+  (`Metadata: #Strings`). Reading a specific address is a coordinate, so it gets
+  a carrier: `--heap "#Strings:0x1a4"`.
+- **Table selection grammar.** Both, since output already prints hex tokens and
+  users will paste them: `TypeDef` and `0x02` address the same table.
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -309,11 +309,12 @@ dotnet-inspect library My.dll --heap "#Strings:0x1a4"
 
 Deep paging into the middle of a table needs no metadata-specific gesture. It is
 a general row-selection concern, and `--row`/`--rows` are being reworked to
-carry it: a row range such as `--rows 40000..40099` addresses an arbitrary
-window in any section, metadata tables included. This lens therefore adds
-nothing for paging and inherits whatever that work lands. Random access from a
-host that is not the CLI stays a library concern, served by `ProjectRow` and the
-row window.
+carry it ([#3364](https://github.com/richlander/dotnet-inspect/issues/3364)):
+once that lands, a row range such as `--rows 40000..40099` will address an
+arbitrary window in any section, metadata tables included. Neither spelling
+parses a range today. This lens therefore adds nothing for paging and inherits
+whatever that work lands. Random access from a host that is not the CLI stays a
+library concern, served by `ProjectRow` and the row window.
 
 ## Safety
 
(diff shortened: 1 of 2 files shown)


richlander:
## Blocking before restack: taking this branch's side of the conflicts silently reverts #3348

This PR is the only one in the stack that was not restacked. It is still at `ab0d4db9`, `mergeable=CONFLICTING`, while #3345 / #3348 / #3350 / #3352 are all `MERGEABLE`. Its content predates the three rounds of review fixes that landed on #3348.

Simulating the restack — `git merge-tree --write-tree 3e202eb5 ab0d4db9` — gives **seven** conflicts:

```
CONFLICT (content): docs/design/metadata-table-projection.md
CONFLICT (content): src/DotnetInspector.MetadataRendering/MetadataProjectionRenderer.cs
CONFLICT (add/add): src/ILInspector.Metadata/MetadataRowReferences.cs
CONFLICT (content): src/ILInspector.Metadata/MetadataTableProjector.cs
CONFLICT (content): tests/DotnetInspector.MetadataRendering.Tests/MdiCommandTests.cs
CONFLICT (add/add): tests/DotnetInspector.MetadataRendering.Tests/MetadataRowReferenceRendererTests.cs
CONFLICT (add/add): tests/ILInspector.Metadata.Tests/MetadataRowReferenceSearchTests.cs
```

### What changed since I last checked, and why it matters

I looked at this exact risk at the previous head and **reported it as not a risk.** My reasoning then: `MetadataRowReferenceSearchTests.cs` auto-merged, so the merged test file carried 11 references to `UnscannedTables`, and resolving `MetadataRowReferences.cs` toward this branch's side would fail to compile. The compiler was the backstop.

That backstop is gone. `MetadataRowReferenceSearchTests.cs` is now an **add/add conflict** itself, so the source file and its tests can be resolved to this branch's side as a matched, internally consistent pair. My retracted hypothesis is now correct, through no change on this branch — the other side moved.

### Demonstrated, not predicted

I built the naive resolution — take this branch's whole-file side for all seven — on top of `3e202eb5`:

| | Correct resolution | This branch's side |
| --- | ---: | ---: |
| Build errors | 0 | **0** |
| `ILInspector.Metadata.Tests` | 1055 / 0 | **1045 / 0** |
| `DotnetInspector.MetadataRendering.Tests` | 99 / 0 | **95 / 0** |
| `UnscannedTables` mentions in `MetadataRowReferences.cs` | 15 | **0** |

It compiles clean and every suite is green. **Fourteen tests simply cease to exist**, and a deleted test cannot fail. Silently reverted along with them:

- the whole `UnscannedTables` blind spot (`IsComplete => !Truncated && UnreadableRows.IsEmpty`, back to the version that reports a clean answer while an unmodelled table holds the edge)
- the unconditional blob-payload caveat, and the `IsComplete` doc that says the property must not be read as "nothing points here"
- the `abandonedMidTable` final-cell boundary and both of its pinning tests
- `TargetExists` and the past-the-end-of-table disclosure

Every one of those is a fix that a review round produced. None of them would fail loudly if dropped here.

### Ask

Restack onto `3e202eb5` and resolve **every** conflict toward the base side for the reverse-search surface — this branch has no intentional changes to it; it simply predates it. Concretely, after resolving, these should hold:

```
grep -c UnscannedTables src/ILInspector.Metadata/MetadataRowReferences.cs   # 15, not 0
```
and the two suites should report **1055** and **99**, not 1045 and 95. Those counts are the cheapest possible check that nothing was dropped, and they are worth running precisely because the failure mode here is silence rather than a red build.

The design-doc hunk needs the same care: this branch's side restores *"`IsComplete` is true only when neither happened, which is what makes an empty result trustworthy"* and deletes the `UnscannedTables` bullets. Nothing compiles a paragraph, and `markdownlint` checks formatting rather than claims, so prose is the one place a bad resolution leaves no trace at all. Re-run `npx markdownlint-cli` on the result.

No comment on this PR's own change (settling the metadata lens on a section category) — that is unreviewed by me and this finding is entirely about the restack.

richlander:
## Correction to my previous comment: don't resolve these conflicts — replay the branch. Verified zero-conflict.

I gave you a recipe for carefully resolving seven conflicts. There is a better answer that removes the hazard instead of asking you to navigate it, and I should have found it before posting the first one.

**All seven conflicts are an artifact of this branch carrying stale copies of the stack's files. This PR's own delta is two documents.**

```
$ git diff --name-only $(git merge-base ab0d4db9 6e56a59f) ab0d4db9
docs/design/metadata-table-projection.md
docs/design/output-shapes.md
```

So the four commits that are actually yours never touch `MetadataRowReferences.cs`, `MetadataTableProjector.cs`, the renderer, or any test file. `merge-tree` reports conflicts in those files only because a *merge* pits this branch's old copies against the base's new ones. A *replay* has nothing to pit.

### Measured

Cherry-picked all four of this PR's commits onto `3e202eb5`:

```
$ git cherry-pick $(git merge-base ab0d4db9 6e56a59f)..ab0d4db9
[detached HEAD ...] Settle the metadata lens on a section category, not a new flag
[detached HEAD ...] Correct the disclosure mechanism and mark the surface as unimplemented
[detached HEAD ...] Stop calling the taxonomy "three families" now that it has four
[detached HEAD aeec6790] Stop rejecting deep paging on a principle we are about to reverse

$ git status --porcelain | grep -E '^(UU|AA|DU|UD)'
(none)
```

**Zero conflicts.** `metadata-table-projection.md` auto-merged. The result:

| Check | Result |
| --- | --- |
| Files changed vs base | 2, both under `docs/` |
| Product code touched | none |
| `UnscannedTables` in the doc | **8** (the #3348 doc fixes survive) |
| `which is what makes an empty result trustworthy` restored | **0** — the regression does not occur |

The entire silent-revert hazard I described disappears. Nothing to resolve carefully, because nothing is in conflict.

### So: rebase, don't merge

Drop the stale stack commits and replay these four onto the new base (after #3345–#3352 land, that base is `main`). Whatever restack tooling produced the stack should do this; the cherry-pick above is the manual equivalent and is known to apply clean.

Worth checking after: `git diff --name-only main...HEAD` should list exactly those two documents. If it lists anything else, the rebase kept a stale commit and the earlier hazard is back.

### One reason not to sit on this

The stack can merge without this PR — it is docs-only, so no product behavior is incomplete. But it does leave `metadata-table-projection.md` on `main` asserting a surface that cannot exist, declaratively rather than as an open question:

> The projection is exposed as its own table lens… It rides the `library` command
>
> ```bash
> dotnet-inspect library My.dll --table TypeDef
> ```

Three worked examples, plus `--table <Name>` named again as the disclosure mechanism. As this PR establishes, `--table` on `library` collides with the existing presentation modifier. An agent implementing the lens from that doc would build the colliding flag; that is a concrete mis-implementation the doc invites, not a stylistic quibble. `624a1d51` also marks the surface unimplemented, which is what keeps the rest of the document honest in the meantime.

Docs-only means it does not block the stack. It does not mean it is low value — it is the correction that stops the next person building the wrong thing.

richlander:
## Correcting myself again: calling this "a docs-only PR" was wrong

@richlander is right, and my previous comment's framing was misleading. **As it stands this PR changes 17 files across 17 commits, and merging it in its current state would revert product code.** My first comment said that; my second one described the branch as "docs-only" and undercut it. Both of my earlier comments describe something real, but they describe *different things*, and I did not say which was which.

The precise reconciliation, via `git cherry` against the current base tip `3e202eb5` (`-` = an equivalent patch is already upstream, `+` = genuinely new):

```
- 79344d24 Give the metadata projection a row window and on-demand row lookup
- 45a8b578 Rewrap so an issue reference does not start a line
- 2d3832d7 Make ProjectRow's contract unambiguous about the table selection
- 14e3f31f Add reverse-reference search to the metadata projection
- 82776eea Report a malformed edge column as a reverse-search blind spot
- a1ea72c4 Pin the invariant the malformed-edge gate rests on
- dd8ff621 Describe the metadata image and read heap values by address
- 7d7c618d Let a corrupt table stream fail visibly instead of reading as empty
- 08c06a0c Measure the projection's allocation shape instead of guessing at it
- 56695ee1 Correct the allocation measurement's counting and scope
- 48e3d215 Fix string sizing, report exact bytes, and scope the drop test
- 8037da9d Report measured ranges and fix the drop test's blind spot
- 6e56a59f Apply the reviewers' non-blocking wording corrections
+ cd100450 Settle the metadata lens on a section category, not a new flag
+ 624a1d51 Correct the disclosure mechanism and mark the surface as unimplemented
+ c42f1dbb Stop calling the taxonomy "three families" now that it has four
+ ab0d4db9 Stop rejecting deep paging on a principle we are about to reverse
```

**13 of the 17 commits are the pre-rebase stack**, carried on this branch and now superseded upstream by their rebased equivalents. Those 13 are where all 15 non-doc changed files and every conflict come from. The 4 `+` commits — the actual contribution — touch two documents.

That is why the two readings differ, and the honest way to state it is: *the PR's intended contribution is two docs; the PR as it currently exists is a 17-file revert of the stack with two docs on top.* Only the second one is what a reviewer or a merge button sees, so it is the one that governs.

### What this does and does not change

**Does not change:** the fix is still a rebase, and the measurement behind it still holds. `git rebase` drops patch-equivalent commits by default, and every one of the 13 is patch-equivalent — including `dd8ff621`, whose `range-diff` `!` marking turns out to be a hunk-header context difference that `patch-id` normalizes away. Replaying just the 4 onto `3e202eb5` applies with **zero conflicts** and yields two changed files, `UnscannedTables` intact at 8 mentions, and the "which is what makes an empty result trustworthy" claim **not** restored.

**Does change:** please read my "the stack can merge without this PR" line as scoped to *product behavior* only — the 4 real commits add no product code, so `main` is not missing a feature. It was not a statement that this PR is currently harmless to merge. It is not.

### Unambiguous, since I muddied it

- **Do not merge this PR in its present state**, before or after the stack lands. It would revert #3348's blob disclosure, `UnscannedTables`, `TargetExists`, the `abandonedMidTable` boundary, and 14 tests.
- **Rebase first.** Afterwards `git diff --name-only <base>...HEAD` must list exactly `docs/design/metadata-table-projection.md` and `docs/design/output-shapes.md`. Any third entry means a stale commit survived and the revert hazard is still live.
- That check is worth running rather than eyeballing the file count, because the failure mode here produces a green build with fewer tests, not a red one.

richlander:
## Adversarial review reconciliation — clean at `ae47d1b0`

Single-model review (**MAI-Code**), isolated in a detached read-only worktree, two rounds. This PR changes two documents and no product code, which puts it in the single-review tier.

### The restack was the real work

This branch was the only one in the stack never restacked. At `ab0d4db9` it carried **17 commits touching 17 files**, 13 of them stale pre-rebase copies of #3345/#3348/#3350/#3352. Merging it in that state would have silently reverted `UnscannedTables`, `TargetExists`, the unconditional blob caveat, the `abandonedMidTable` boundary, and 14 tests — while compiling green.

**A plain `git rebase origin/main` does not fix this**, contrary to what my earlier comment predicted. Every stack PR was **squash-merged**, so each upstream commit is the squash of several branch commits and per-commit `patch-id` equivalence fails. Rebase therefore tried to replay all 17 and hit conflicts in `MetadataTableProjector.cs`, `MdiCommand.cs`, and the window tests — the exact hazard, re-armed.

What works is replaying only this PR's own commits:

```bash
git rebase --onto origin/main 6e56a59f
```

Zero conflicts. Post-rebase verification, since the failure mode here is silence rather than a red build:

| Check | Expected | Actual |
| --- | --- | --- |
| Files changed vs `main` | 2, both docs | **2, both docs** |
| `UnscannedTables` in the doc | 8 | **8** |
| `makes an empty result trustworthy` | 0 | **0** |
| `TargetExists` / `UnreadableRows` | 2 / 5 | **2 / 5** |
| #3352's `Measured: allocation shape` | intact | **intact** (`78,288,240` present) |

### Findings

| # | Source | Finding | Resolution |
| --- | --- | --- | --- |
| 1 | mine | `output-shapes.md` called `--il-offset` "the family's one implemented member", but `--il-offsets` is also implemented and also carries coordinates | Family recounted in **currencies, not flags**: the two are mutually exclusive spellings of one currency, so the count stays right and the definition gets sharper. `47ed8f88` |
| 2 | MAI (filed BLOCKING) | The deep-paging prose reads as a claim that `--rows 40000..40099` works today | Severity disputed, point accepted. `ae47d1b0` |
| 3 | mine, while fixing #2 | My first fix asserted "None of it is implemented yet" — itself false | Corrected before commit; see below |

On **#2** I disputed BLOCKING: the paragraph sits under a header reading "**Status: designed, not yet implemented.** Nothing in this section runs today", said the flags "are being reworked", and the second site is under "## Open questions" saying "out of scope for this series". But a careful reader misread it anyway, and if tense is the only thing preventing a misreading, the wording is inadequate regardless. Both sites now state it flatly ("Neither spelling parses a range today") and cite #3364 so the status is checkable rather than asserted.

**#3 is the one worth recording.** Fixing an overclaim, I wrote a new one. `SharedOptions.cs:38,100-103` shows `-n`/`--head` and `--tail` already exist as `Option<int?>`, and `--rows` already exists as an `Option<bool>` that switches their unit from output lines to data rows. The gap is specifically **range parsing**, not the whole design. I caught this by verifying my own replacement text, and asked the reviewer to check it independently; they confirmed it, and additionally found that bare `--rows` errors with `--rows requires -n/--head N or --tail N`.

### Verified by running the product, not by reading it

The doc quotes exact error strings and depicts exact output. Each was executed:

| Claim | Result |
| --- | --- |
| `-S @Metadata` fails with `Select value '@Metadata' not found.` | **Exact match**, exit 1 |
| `--table` collides with a table-name argument | `--table TypeDef` -> `Unrecognized command or argument 'TypeDef'`; `SharedOptions.cs:27` defines it as `Option<bool>` |
| `-S @Metadata --count` renders a per-section count table | `-S @Performance --count` produces **exactly** the depicted `\| Section \| Count \|` shape |
| `Metadata: TypeRef` is a legal section name | `-S "Performance: Boxing"` already works — colon-and-space names are proven precedent |
| Vector rung | `--columns Member --tsv` returns one column |
| `--rows --head 20` | Works today; returns exactly N rows |
| `--heap` | Does **not** exist — correctly marked designed-not-implemented |
| `ExplicitOnly` render gate | `SectionPipeline.IsRequested:499-505` matches the description |
| `AddCategory` pattern | `LibrarySections.cs:95` registers `@Performance` exactly as cited |

So all four rungs of the doc's ladder example map onto demonstrated behavior, and the two unbuilt pieces (`@Metadata`, `--heap`) are labelled as unbuilt.

The reviewer separately confirmed no contradiction with `section-model.md` or `progressive-disclosure.md`, and no parser problem with the colon.

### Validation

`markdownlint` green in CI. No product code changed, so no test run is implicated; `test` is correctly SKIPPED by the path filter. `mergeStateStatus: CLEAN`.

Ready to merge

richlander:
## Re-review at `ae47d1b0` — rebase is correct, every checkable claim verifies. Ready to merge.

### The rebase did exactly what was needed

The check I asked for, run against the new `main`:

```
$ git diff --name-only origin/main...ae47d1b0
docs/design/metadata-table-projection.md
docs/design/output-shapes.md
```

Two files, no third entry, no product code. All 13 stale stack commits are gone; the 6 remaining are the 4 originals plus 2 new. `MERGEABLE` / `CLEAN`. The 17-file revert hazard is fully retired — nothing here can drop a test, because nothing here is a test.

### Claims verified against the product, not read

A docs PR whose entire purpose is correcting a doc that asserted a flag which cannot exist earns exactly one kind of review: check every factual claim against the thing it describes. All of them hold.

| Claim in the diff | Verified |
| --- | --- |
| `-S @Metadata` fails with `Select value '@Metadata' not found.` | **ran it** — verbatim match, incl. the `Error: ` prefix |
| `@Performance` is registered via `.AddCategory(...)` | `LibrarySections.cs:95` — `.AddCategory(SectionCategoryNames.Performance, PerformanceKinds.Sections)`, precisely the shape the proposed line mirrors |
| `-S @Performance --count` renders a Section/Count table | **ran it** — matches the illustrated shape |
| `SectionPipeline.IsRequested` returns `false` for `ExplicitOnly` unless explicitly included | `SectionPipeline.cs:496-515` — exact, and the `ExplicitOnly` return precedes the verbosity switch, so "no verbosity level auto-selects it" is right down to `Detailed` |
| `--il-offset cannot be combined with --il-offsets` | `LibraryCommand.cs:114` |
| `IL coordinate sections require --il-offset` | `LibraryCommand.cs:95` |
| `--rows` is a boolean switching the unit of `-n`/`--head`/`--tail` | `SharedOptions.cs:38` — `Option<bool>`, description says exactly that |
| Neither `--row` nor `--rows` parses a range today | `--rows` is `Option<bool>`; `--row` is "a 1-based index, first, or last" |
| `--row` selects the Nth row past an invisible printability filter | the product's own option description says "select a **printable** row" — and #3364 is open, titled *"--row indexes an invisible sequence"* |
| `--heap` is designed, not implemented | no occurrence in `src/` — correctly stated as designed |
| `markdownlint` | clean, exit 0, both files |

### A near-miss of my own worth recording

I nearly filed `--il-offset cannot be combined with --il-offsets` as a fabricated error string. My first grep for `cannot be combined with` was piped through `head -5`, and the real hit is the sixth. The string is real and at `LibraryCommand.cs:114`. Truncated output that looks like an exhaustive search is a good way to invent a finding; the cost of being wrong in this direction is someone deleting a correct line.

### What makes this version right

The previous text asserted a surface in the indicative mood — *"The projection is exposed as its own table lens… It rides the `library` command"* — with three worked `--table` examples. A reader had no way to tell it from a description of shipped behavior, and following it would have built a flag colliding with the presentation modifier.

`**Status: designed, not yet implemented.** Nothing in this section runs today` fixes that at the root, and does it the way AGENTS.md asks: a property that isn't enforced is marked unverified rather than asserted. Quoting the *actual* current failure (`Select value '@Metadata' not found.`) is better still — it makes the unimplemented status falsifiable, so this paragraph will start lying the moment the feature lands, which is exactly when someone should update it.

Two details I'd have expected to slip and didn't: the `--rows --head 20` example is consistent with `--rows` requiring `-n`/`--head N` (`SharedOptions.cs:189`), and the coordinate-carrier family is counted in *currencies* rather than flags, which is what lets `--il-offset`/`--il-offsets` be one member instead of two.

### Verdict

**Ready to merge.** No open items. With this landed, the #3341 series closes with `main` carrying an accurate description of both what the projection does and what it does not yet do.

PR #3365, opened 2026-07-28
Gate authored-source slices on parse validity; stop over-running the end boundary

richlander:
## Conclusion

The authored-source slicer was appending the enclosing type's closing brace to any member that was the last one in its type. This adds a Roslyn-backed parse-validity gate that catches it, then fixes it with a one-line change. Over-capture drops **971 → 10** across 35,880 members.

## Why a parse-validity gate

The slicer reconstructs a member from a sequence-point line range, so it has two independent boundaries: a backward scan for the signature and a forward scan for the closing brace. Only the start boundary shows up in the extracted text's *signature*, so a member-identity round trip cannot see an end-boundary defect — swallowing a trailing `}` leaves the signature untouched. Measured directly:

```
--- OVER-CAPTURED (last member) ---
  signature-identity : int Target()      <- IDENTICAL
  parses-cleanly     : NO -> CS1022
--- CORRECT ---
  signature-identity : int Target()      <- IDENTICAL
  parses-cleanly     : yes
```

Roslyn is the independent oracle. The claim is narrow — extracted text must parse as a well-formed member declaration — but it is sensitive to *both* boundaries at once and needs no per-member expected output, so it scales over a corpus. Product paths stay Roslyn-free; a hand-rolled checker would only be a second copy of the heuristic under test. This mirrors `LadderRung1ProductPathGateTests`, which already parses decompiler output into a shell and filters `DiagnosticSeverity.Error`.

The corpus is every PDB-bearing assembly beside the test binary whose documents resolve on disk — all built from this repository, so these are real compiler sequence points over real authored C#, not synthesized ranges. The gate drives the product path end to end (`PdbContext.EnumerateMemberSources` → `ExtractMethodBody`) and reconstructs nothing.

## The defect

`endsAtDeclaration` asked whether the range *both* began at a declaration *and* closed its own block. A conventionally braced member starts its sequence range on `{`, so the first conjunct was false and the forward scan ran for nearly every member. That was harmless while a sibling followed, and appended the enclosing type's `}` when the member was last in its type — which also defeated dedent, since a brace at column 0 drags the common indent to 0.

Asking the captured range alone fixes it. All 947 pre-existing tests pass **unchanged**, including the #3278 one-line expression-body and auto-property regressions the conjunct was added to guard: `EndsDeclaration` already answers those correctly on its own, so the conjunct was redundant.

## Evidence

Measured over this repository's assemblies, 35,880 unique API members with local source:

| | before | after |
|---|---|---|
| over-capture | 971 (2.71%) | 10 (0.03%) |
| well-formed | 77.79% | 80.42% |

The residue is raw-string literals, where `EndsDeclaration` deliberately reports "unclosed" as the conservative answer. Over the test-output corpus it reaches zero, so the gate asserts **zero** rather than a pinned count.

**Non-vacuity:** reverting the one-line product change fails both boundary tests, reporting 115 offenders and the exact text diff on the fixture canary. The canary uses `DiffAsmTarget.Api.Ping(LibB::Shared.Token)` — a real compiled fixture that is the last member of the last type with an empty body, so its whole range is the closing brace.

## Boundary / non-actions

Two defect populations remain. They are characterized with ceilings, not hidden, so neither can grow silently:

- **18.21% type-header shapes** — positional record accessors, primary constructors, and field-initializer constructors have no authored declaration, so their sequence points land on the type header. This PR only characterizes it; **the fix is the stacked PR on top of this one.**
- **1.67% under-capture** — e.g. a property getter whose range covers only the accessor. `JsonSerializerOptions.MaxDepth` reproduces this identically on both sides of this change.
- **0.30% start-boundary misses**, where the backward scan begins mid-body.

`dotnet format` was evaluated as an alternative oracle and rejected: it exits **0** on syntactically invalid C#, silently reformatting the exact over-capture shape and leaving the stray brace. It is Roslyn underneath anyway, and costs ~1s/file versus 2.6s for the whole 4,376-member sweep in-process.

## Validation

- `ILInspector.Metadata.Tests` 950/950 (5 skipped, mdv absent)
- `dotnet-inspect.Tests` 2216 (12 skipped, ilasm/ildasm absent)
- `JsonSerializerOptions.MaxDepth` output byte-identical across the change

## Review

High risk — heuristic change with broad blast radius. Requests two reviews from different model families.

### DIFF
--- a/src/ILInspector.Metadata/SourceLinkResolver.cs
+++ b/src/ILInspector.Metadata/SourceLinkResolver.cs
@@ -346,57 +346,40 @@ public class SourceLinkResolver
     }
 
     /// <summary>
+    /// Index on <paramref name="line"/> where code resumes, given the state carried into it, or
+    /// <c>-1</c> when the line is comment or literal text throughout. A line that closes a
+    /// multi-line comment or literal and then holds real code must be read as code from that
+    /// point (adversarial review, MAI-Code); asking only whether the line *began* in one
+    /// suppressed the question on exactly that line.
+    /// </summary>
+    private static int IndexWhereCodeResumes(string line, LexState state)
+    {
+        if (!state.InBlockComment && !state.InLiteral)
+            return 0;
+
+        var probe = state.Clone();
+        int depth = 0;
+        int index = Scan(line, probe, ref depth, start: 0, untilLiteralCloses: false, out _, untilCodeResumes: true);
+        return probe.InBlockComment || probe.InLiteral ? -1 : index;
+    }
+
+    /// <summary>
     /// Index just past the attribute list opening at index 0, or <c>-1</c> when it does not
     /// close on this line. Brackets nest — <c>[Foo(new[] { 1 })]</c> — and a string inside the
     /// list may spell a bracket of its own, so neither is counted structurally.
     /// </summary>
     private static int IndexPastAttributeList(string trimmed)
     {
+        var probe = new LexState();
         int depth = 0;
-        for (int i = 0; i < trimmed.Length; i++)
-        {
-            char c = trimmed[i];
+        int index = Scan(trimmed, probe, ref depth, start: 0, untilLiteralCloses: false, out _, untilBracketsClose: true);
 
-            if (c == '"')
-            {
-                int end = EndOfSingleLineLiteral(trimmed, i);
-                if (end < 0)
-                    return -1;
-                i = end - 1;
-                continue;
-            }
+        // The list did not finish on this line: it continues below, or a comment or literal
+        // swallowed the rest of the line.
+        if (probe.BracketDepth != 0 || probe.InBlockComment || probe.InLiteral || probe.Untracked)
+            return -1;
 
-            if (c == '\'')
-            {
-                i++;
-                while (i < trimmed.Length && trimmed[i] != '\'')
-                    i += trimmed[i] == '\\' ? 2 : 1;
-                continue;
-            }
-
-            if (c == '/' && i + 1 < trimmed.Length && trimmed[i + 1] == '/')
-            {
-                // The rest of the line is a comment, so the list does not close on it.
-                return -1;
-            }
-
-            if (c == '/' && i + 1 < trimmed.Length && trimmed[i + 1] == '*')
-            {
-                int close = trimmed.IndexOf("*/", i + 2, StringComparison.Ordinal);
-                if (close < 0)
-                    return -1;
-
-                i = close + 1;
-                continue;
-            }
-
-            if (c == '[')
-                depth++;
-            else if (c == ']' && --depth == 0)
-                return i + 1;
-        }
-
-        return -1;
+        return index;
     }
 
     /// <summary>
@@ -435,13 +418,12 @@ public class SourceLinkResolver
             if (!trimmed.StartsWith(keyword, StringComparison.Ordinal))
                 continue;
 
-            // "get;" and "get =>" are accessors; "getCount" and "setting" are not.
-            int after = keyword.Length;
-            if (after >= trimmed.Length)
-                return true;
+            // "get;", "get =>", "get {", and a bare "get" whose body is below are accessors.
+            // "getCount" and "setting" are not, and neither is an assignment to a local named
+            // "set", which a bare "=" accepted (adversarial review, GPT).
+            var rest = StripLeadingComments(trimmed[keyword.Length..].TrimStart());
 
-            char c = trimmed[after];
-            if (c is ';' or '{' or '=' or ' ' or '\t')
+            if (rest.Length == 0 || rest[0] is ';' or '{' || rest.StartsWith("=>", StringComparison.Ordinal))
                 return true;
         }
 
@@ -508,11 +490,17 @@ public class SourceLinkResolver
         int limit = Math.Min(lines.Length, to + ForwardScanLimit);
         for (int i = to; i < limit; i++)
         {
-            // Asked before the line is scanned, so it must not be asked at all when the carried
-            // state says the line is not code: a "set" inside a multi-line block comment or raw
-            // string literal is text, not a sibling (adversarial review, Gemini).
-            if (slicingAccessor && !state.InBlockComment && !state.InLiteral && OpensSiblingAccessor(lines[i]))
-                return to;
+            // Asked before the line is scanned, so it must be asked of the line's code alone. A
+            // "set" inside a multi-line block comment or raw string literal is text, not a
+            // sibling (adversarial review, Gemini) — but a line that *closes* one and then
+            // declares a real sibling is code from that point (adversarial review, MAI-Code).
+            // A line inside an unclosed bracketed construct is not a declaration either.
+            if (slicingAccessor && state.BracketDepth == 0)
+            {
+                int resume = IndexWhereCodeResumes(lines[i], state);
+                if (resume >= 0 && OpensSiblingAccessor(lines[i][resume..]))
+                    return to;
+            }
 
             ScanLine(lines[i], state, ref depth);
 
@@ -545,6 +533,14 @@ public class SourceLinkResolver
         public bool InBlockComment;
 
         /// <summary>
+        /// Open square brackets carried across lines. An attribute list, like any bracketed
+        /// construct, may span lines; a line inside one is not a declaration, and reading it as
+        /// one truncated an accessor at a "set" that was really an attribute name (adversarial
+        /// review, GPT).
+        /// </summary>
+        public int BracketDepth;
+
+        /// <summary>
         /// Set when a brace could not be placed — an unterminated single-line literal, or a
         /// delimiter run this scanner will not guess at. The depth count is unusable from that
         /// point on, and callers treat it as "do not know" rather than as a depth.
@@ -560,6 +556,17 @@ public class SourceLinkResolver
         public void Pop() => frames.RemoveAt(frames.Count - 1);
 
         public void Replace(Frame frame) => frames[^1] = frame;
+
+        /// <summary>
+        /// A copy that can be advanced without disturbing the scan in progress. Used to ask
+        /// where a line's code resumes without consuming the line.
+        /// </summary>
+        public LexState Clone()
+        {
+            var copy = new LexState { InBlockComment = InBlockComment, Untracked = Untracked, BracketDepth = BracketDepth };
+            copy.frames.AddRange(frames);
+            return copy;
+        }
 
         /// <summary>
         /// True when the state cannot survive a line break: an ordinary or single-quoted
@@ -660,23 +667,6 @@ public class SourceLinkResolver
     }
 
     /// <summary>
-    /// Index just past the literal opening at <paramref name="start"/>, or <c>-1</c> when it
-    /// does not close on that line. Used where the construct examined is known to be
-    /// single-line — an attribute list — so a literal that continues means it did not close.
-    /// <para>
-    /// This runs the same scanner as <see cref="ScanLine"/> rather than a second literal
-    /// parser, so the verbatim, raw, and interpolated forms are read one way everywhere.
-    /// </para>
-    /// </summary>
-    private static int EndOfSingleLineLiteral(string line, int start)
-    {
-        var state = new LexState();
-        int depth = 0;
-        int end = Scan(line, state, ref depth, start, untilLiteralCloses: true, out _);
-        return state.InLiteral ? -1 : end;
-    }
-
-    /// <summary>
     /// Scans <paramref name="line"/> from <paramref name="start"/>, returning the index it
     /// stopped at. With <paramref name="untilLiteralCloses"/> the scan stops as soon as the
     /// literal it opened is closed, which is how a caller consumes a single literal.
@@ -687,11 +677,14 @@ public class SourceLinkResolver
         ref int depth,
         int start,
         bool untilLiteralCloses,
-        out char significant)
+        out char significant,
+        bool untilCodeResumes = false,
+        bool untilBracketsClose = false)
     {
         significant = '\0';
         int i = start;
         bool opened = false;
+        bool bracketOpened = false;
 
         if (!state.InBlockComment && !state.InLiteral && IsDirective(line, out bool conditional))
         {
@@ -701,12 +694,18 @@ public class SourceLinkResolver
             if (conditional)
                 state.Untracked = true;
 
-            return '\0';
+            return line.Length;
         }
 
         while (i < line.Length)
         {
             if (untilLiteralCloses && opened && !state.InLiteral)
+                return i;
+
+            if (untilCodeResumes && !state.InBlockComment && !state.InLiteral)
+                return i;
+
+            if (untilBracketsClose && bracketOpened && state.BracketDepth == 0)
                 return i;
 
             char c = line[i];
@@ -910,6 +909,19 @@ public class SourceLinkResolver
                 {
                     depth++;
                 }
+            }
+            else if (c == '[')
+            {
+                if (!inHole)
+                {
+                    state.BracketDepth++;
+                    bracketOpened = true;
+                }
+            }
+            else if (c == ']')
+            {
+                if (!inHole && state.BracketDepth > 0)
+                    state.BracketDepth--;
             }
             else if (c == '}')
             {
--- a/src/ILInspector.Metadata/SourceLinkResolver.cs
+++ b/src/ILInspector.Metadata/SourceLinkResolver.cs
@@ -493,6 +493,11 @@ public class SourceLinkResolver
 
         int limit = Math.Min(lines.Length, to + ForwardScanLimit);
 
+        // Where the run of lines leading up to the current one stopped holding code, or -1
+        // when the current line does. A sibling's attributes and comments are the sibling's,
+        // so the member ends where that run began.
+        int triviaRunStart = -1;
+
         for (int i = to; i < limit; i++)
         {
             // Asked before the line is scanned, so it must be asked of the line's code alone. A
@@ -503,7 +508,11 @@ public class SourceLinkResolver
             {
                 int resume = IndexWhereCodeResumes(lines[i], state);
                 if (resume >= 0 && OpensSiblingAccessor(lines[i][resume..]))
-                    return IndexPastTrailingTrivia(lines, to, i);
+                    return triviaRunStart >= 0 ? triviaRunStart : i;
+
+                triviaRunStart = HoldsOnlyTrivia(lines[i], state)
+                    ? (triviaRunStart >= 0 ? triviaRunStart : i)
+                    : -1;
             }
 
             ScanLine(lines[i], state, ref depth);
@@ -519,40 +528,43 @@ public class SourceLinkResolver
     }
 
     /// <summary>
-    /// The end of a member that yielded to the sibling declaration on <paramref name="sibling"/>.
+    /// True when <paramref name="line"/> carries nothing but blank space, comments, or
+    /// attribute lists, given the state carried into it.
     /// <para>
-    /// Everything above that sibling is the member's, so the range runs to it. Returning the
-    /// original range instead discarded the accessor's own closing brace and the statements
-    /// above it (adversarial review, Gemini). Answering with the last brace that closed put
-    /// the end inside a lambda nested in an expression-bodied accessor, truncating the
-    /// expression (adversarial review, MAI-Code and Gemini, independently): a member with no
-    /// block of its own has no brace to find, so brace depth cannot answer this. The sibling's
-    /// own position can, and it does not care how either member is bodied.
+    /// Asking this of the line's text alone answered only for trivia that both opens and
+    /// closes on one line: a multi-line attribute list or block comment leading the sibling
+    /// was kept as the member's source, which does not parse (adversarial review, GPT). It is
+    /// the same lesson the sibling question itself took three rounds to learn — a line-level
+    /// predicate must be asked of the line's code, not its text.
     /// </para>
     /// <para>
-    /// Only the sibling's own leading trivia is given back. A line holding code is kept, so a
-    /// line that merely looks like trivia mid-construct costs nothing.
+    /// A line still inside a carried literal is the member's code, not the sibling's trivia,
+    /// so it is the one carried construct that answers no.
     /// </para>
     /// </summary>
-    private static int IndexPastTrailingTrivia(string[] lines, int from, int sibling)
+    private static bool HoldsOnlyTrivia(string line, LexState state)
     {
-        int end = sibling;
+        if (state.InLiteral)
+            return false;
 
-        while (end > from && HoldsOnlyTrivia(lines[end - 1]))
-            end--;
+        int resume = IndexWhereCodeResumes(line, state);
+        if (resume < 0)
+            return true;
 
-        return end;
-    }
+        var rest = StripLeadingComments(line[resume..].TrimStart());
 
-    /// <summary>
-    /// True when the line carries nothing but blank space, a comment, or an attribute list.
-    /// </summary>
-    private static bool HoldsOnlyTrivia(string line)
-    {
-        var trimmed = line.TrimStart();
-        return trimmed.Length == 0
-            || trimmed.StartsWith('[')
-            || StripLeadingComments(trimmed).Length == 0;
+        while (rest.StartsWith('['))
+        {
+            int close = IndexPastAttributeList(rest);
+
+            // The list runs past this line, so nothing else can be on it.
+            if (close < 0)
+                return true;
+
+            rest = StripLeadingComments(rest[close..].TrimStart());
+        }
+
+        return rest.Length == 0;
     }
 
     /// <summary>
--- a/src/ILInspector.Metadata/SourceLinkResolver.cs
+++ b/src/ILInspector.Metadata/SourceLinkResolver.cs
@@ -346,15 +346,19 @@ public class SourceLinkResolver
     }
 
     /// <summary>
-    /// Index on <paramref name="line"/> where code resumes, given the state carried into it, or
-    /// <c>-1</c> when the line is comment or literal text throughout. A line that closes a
-    /// multi-line comment or literal and then holds real code must be read as code from that
-    /// point (adversarial review, MAI-Code); asking only whether the line *began* in one
-    /// suppressed the question on exactly that line.
+    /// Index on <paramref name="line"/> where a declaration could begin, given the state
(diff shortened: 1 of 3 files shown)


richlander:
## Review reconciliation (Gemini Pro, GPT)

Both reviewers independently reproduced the same blocking defect, and they were right: **the `startsAtDeclaration &&` conjunct was not redundant.** It was load-bearing by accident.

### Blocking — fixed in `7db5dc90`

**Interpolation holes corrupted the end-boundary brace count** (Gemini, GPT).
`ScanLine` treated a quote inside an interpolation hole as closing the literal, so the braces after it were counted as structural. `depth` reached zero early, `EndsDeclaration` reported the range closed, and the forward scan that recovers the member closing brace was suppressed.

The conjunct had been hiding this. A conventionally braced member starts its range on `{`, so `startsAtDeclaration` was false and `EndsDeclaration` was short-circuited away entirely — it was never allowed to be wrong.

I verified this as a regression rather than accepting it on reading, running the same probe against both commits:

```
parent   21c00743:  public string P { get {  return $"{new Holder { S = "}" }.S}";  } }   <- keeps its braces
head     5e1da840:  public string P { get {  return $"{new Holder { S = "}" }.S}";        <- loses them
```

A plain-string member of the identical shape is unaffected in both, which isolates the literal as the cause.

Fixed by consuming an interpolated string as one unit, tracking hole nesting and the strings, chars, and nested interpolations a hole may contain, so hole braces never reach the caller depth count. The verbatim forms stay on the existing cross-line path — my first attempt routed them to `untracked` and broke `OpeningBraceInsideVerbatimInterpolatedString_DoesNotLookLikeAnOpenBlock`, which is exactly the pin that should have caught it.

Five cases pinned in `BracesInsideInterpolationHoles_DoNotEndTheDeclaration`, as GPT asked.

### Blocking — fixed in `7db5dc90`

**The corpus filter excluded every generic member** (Gemini). It rejected any name spelling `<`, which is how the compiler spells `<M>d__0` but also how C# spells `Walk<THandle>`. Measured on one assembly: **276 of 528 excluded members were ordinary generic API surface**, including `MetadataRelationshipTraversal.Walk<THandle,TRelationship>` and all of `RelationshipChain<THandle>`.

Now matched on a leading `<` in a name *segment*, which only the compiler can spell. Over-capture remains zero against the wider corpus, so this strengthens the claim rather than weakening it.

**Non-vacuity** (self-found). `SlicedMembers_DoNotCaptureTheEnclosingTypesClosingBrace` filtered straight to offenders without asserting the corpus was non-empty. Now asserts it, as its two siblings already did. Gemini is right that the suite would still have failed elsewhere on an empty corpus, but the test should stand on its own.

### Non-blocking — acknowledged, not actioned

GPT: the rate ceilings (`< 3.0%` under-capture, `< 1.5%` malformed) let a single new offender hide by dilution. Correct. The reproduced regression classified as `Malformed` and moved the rate by ~0.003%. I have addressed the specific shape with an exact-value regression test rather than reworking the ceilings, because per-identity baselining is a larger change than this PR should carry. Filing as follow-up.

Metadata 955/955, Services 319/319, CLI 2216 (12 skipped).

richlander:
## Round 2 reconciliation — head `90ed3fc7`

Both reviewers rejected `7db5dc90`. Every finding reproduced, and both are fixed.

### GPT — literal scanning (3 blocking)

All three reproduced. I verified each input parses cleanly under Roslyn first, then swept line ranges, because the exposure is range-dependent and two of the quoted ranges were off by one:

| Finding | Broken at | Failure |
|---|---|---|
| `$@"`/`@$"` with a quote inside a hole | `[.,6]` | under-capture — forward scan suppressed |
| raw `$$"""` taking a nested `"""` as its terminator | `[.,6]` | under-capture |
| comment inside a hole spelling a brace | `[.,7]` | over-capture via the untracked fallback |

GPT's diagnosis was right in each case, and the common cause is structural: the scanner recognized one literal form at a time, so **every form it did not model reported the wrong brace depth** — which is the only thing the end-boundary decision reads.

Rather than add a fourth special case, `LexState` now models the nesting C# actually has: a stack of frames, each knowing its delimiter length, its hole-opening brace count, and whether the scan is in literal text or in the hole's code. A brace adjusts structural depth only outside every literal. This replaces `IndexOfQuoteRun`, `EndOfStringLiteral`, and `EndOfInterpolatedString`, and lets a multi-line raw literal be **tracked** instead of abandoned.

All 20 range/form combinations now yield well-formed slices.

### MAI-Code — forward scan (1 blocking)

Reproduced. The scan stopped at the first non-empty line below the range, so a range ending on a statement above the closing brace lost **every remaining statement** along with the brace. It now scans by depth to the line that closes the block the range leaves open.

One correction to the suggested fix: scanning purely by depth is too greedy. A getter's range sits inside its *property's* braces, so running to the closing brace merged the setter into the getter's source. The scan yields to a sibling declaration first, which keeps accessors resolving separately (`Member_OriginalSource_PropertyAccessorOrdinals_ResolveGetterAndSetterSeparately`).

### GPT — corpus vacuity (non-blocking, fixed anyway)

Fixed rather than deferred; it is a non-vacuity concern about the gate itself. `SliceCorpus_StaysBroadEnoughToMeasure` pins floors on slice count, generic members, and distinct files.

### Two tests changed their expected output

Both pinned malformed text, which is what this PR's gate exists to reject:

- `MultiLineRawStringLiteral_FallsBackToTheForwardScan` asserted the fallback appending the *enclosing type's* brace to an expression-bodied member — over-capture, from a fixture that itself had a stray `}`. Now `MultiLineRawStringLiteral_IsTrackedAcrossLines`, asserting the correct slice.
- `BlockBodiedPropertyAccessor_...` briefly changed while I evaluated the greedy scan, and is **restored unchanged** now that the sibling guard is in.

### Gates

`BracesInsideInterpolationHoles_DoNotEndTheDeclaration` (+5 cases covering all three literal shapes), `RangeEndingAboveTheClosingBrace_RecoversTheWholeMember`, `MemberThatClosesItsOwnDeclaration_DoesNotTakeTheTypeBrace` (the #3278 boundary the scan must not cross), `SliceCorpus_StaysBroadEnoughToMeasure`.

Metadata 967/967, Services 319/319, CLI 2216 (12 skipped).

richlander:
## Round-3 review reconciliation

Reviewed by **Gemini Pro** and **GPT** at `90ed3fc7`. Fixed head: `665e8155`.

Between them they raised six distinct defects. **Five reproduced and are fixed; one sub-pair did not reproduce.** Every claim below was re-run against a clean worktree before I acted on it.

### Fixed

| # | Reviewer | Defect | Fix |
|---|---|---|---|
| 1 | Gemini, GPT | `OpensSiblingDeclaration` truncated a method at a `static` local function | Gate on the member being sliced |
| 2 | Gemini, GPT | A newline inside an interpolation hole marked the state `Untracked` | `HasLineBoundLiteral` skips frames with `InHole` |
| 3 | Gemini | `/* c */ set => ...` was not seen as a sibling, so the setter was swallowed | Read past comments/modifiers/attributes before asking |
| 4 | GPT | `@""""` was read as a raw literal and never closed | A verbatim literal is never raw |
| 5 | GPT | Braces in an inactive `#if` branch were counted structurally | Conditional directives mark the depth untracked |

**On (1) — the interesting one.** Both reviewers found it; the fix is not the one either suggested. Depth cannot separate a sibling accessor from a local function: *both* sit one level inside the member being sliced. The discriminator is not the interrupting line, it is the member doing the slicing — only a property or event block can hold a sibling accessor. So the question is asked only when the member being sliced is an accessor (`get_`/`set_`/`init_`/`add_`/`remove_`). That also disposes of GPT's `int set = 0;` and attributed-local-function repros, which are the same bug seen from another angle.

**On (5).** Marking every conditional directive untracked is conservative, and I checked what it costs before choosing it: corpus rates are identical with and without (80.99% well-formed either way). It buys correctness on the unbalanced case for nothing measurable. Non-conditional directives (`#region`, `#nullable`, `#pragma`, `#line`) are skipped as non-code and cost nothing.

### Did not reproduce

Gemini's finding 1 bundled three shapes. The local function is real; the other two are not:

- `var f = async () => await T();` — the line leads with `var`, not `async`, so the predicate never sees a modifier.
- `int[] a = [1, 2, 3];` — the line leads with `int[]`, not `[`.

Both are now pinned as *passing* cases in `StatementsThatLeadWithADeclarationModifier_DoNotTruncateTheMember` so the claim stays checked rather than merely argued.

### Also corrected

`readonly int q = 1;` is not legal as a local declaration, and `#region` without `#endregion` does not parse. Two of my first-cut regression cases asserted against invalid input; the gate caught both, which is the gate working.

### Evidence

New gates: `StatementsThatLeadWithADeclarationModifier_DoNotTruncateTheMember` (6 cases), `AccessorSlice_StopsAtItsSibling` (6), `InterpolationHoleSpanningLines_DoesNotAbandonTheScan`, `UnterminatedLiteralText_StillLeavesTheDepthUnknown`, `VerbatimQuoteRuns_AreNotReadAsRawDelimiters` (5), `ConditionalDirective_SuppressesDepthBasedRecovery`, `NonConditionalDirective_LeavesRecoveryIntact` (3), `RegionDirective_LeavesRecoveryIntact`.

Release: Metadata **991/991**, Services **319/319**, CLI **2216** (12 skipped). Corpus: 80.99% well-formed, under-capture **1.67% → 0.46%**, malformed **0.30% → 0.24%**; measured note updated.

The rate-ceiling dilution point (GPT round 2, Gemini round 3, non-blocking) remains deferred to a follow-up — per-identity baselining rather than percentages.

Round-4 fixed-head review to follow at `665e8155`.

richlander:
## Round-4 review reconciliation

Reviewed by **Gemini Pro** and **MAI-Code** at `665e8155`. Fixed head: `a5bb5c89`.

Three defects raised across the two reviews, **all reproduced, all fixed.** Both reviewers independently found the same first one.

| # | Reviewer | Defect |
|---|---|---|
| 1 | MAI-Code, Gemini | `OpensSiblingAccessor` returned `true` for *any* line opening with `[`, so an attributed local function inside an accessor body truncated the accessor |
| 2 | Gemini | The sibling question is asked **before** the line is scanned, so it was blind to carried lexical state: `set` inside a multi-line block comment or raw string literal read as a sibling |

### On (1)

This is my own round-3 fix reintroducing the round-3 bug one level down. I replaced a line-shape heuristic with a member-gated one — correctly — and then left an unguarded `if (trimmed.StartsWith('[')) return true;` inside it, which is the same "this line looks like a declaration" reasoning I had just removed. Both reviewers found it; MAI-Code's repro and Gemini's differ only in where the attributed local function sits.

The fix skips the attribute list and asks the question of what follows it. That needs a real attribute reader, so **`IndexPastAttributeList` and the `Scan` / `EndOfSingleLineLiteral` seam it rests on move here from #3366**, which is now where they are first needed. #3366 shrinks accordingly. This is deliberately not a third comment/literal reader — the repo already had two too many.

### On (2)

Gemini's, and the sharper finding of the two. The check sits at the top of the forward scan's loop body, above the `ScanLine` call that maintains `state` — so by construction it sees the line with no knowledge of what was open when the line began. Its suggested gate (`!state.InBlockComment && !state.InLiteral`) is what I applied, mirroring how `IsDirective` is already gated. Worth noting that the same reasoning applies to any future predicate placed at that spot.

### Evidence

New gates: `AttributedLocalFunction_DoesNotEndAnAccessor` (3 cases), `AttributedSiblingAccessor_StillEndsTheSlice` (3, covering both same-line and own-line attributes, and a `]` inside an attribute argument string), `AccessorKeywordInsideCommentOrLiteral_IsNotASibling` (4, block comment and raw string).

Release: Metadata **1001/1001**, Services **319/319**, CLI **2216** (12 skipped). All round-3 probes re-run and still correct.

Gemini's note that the corpus ceilings let these hide is fair and is the same dilution point raised in rounds 2 and 3. It stays a tracked follow-up: percentage ceilings over a fixed corpus cannot catch a rare shape, and per-identity baselining is the answer. That is a change to the gate's accounting, not to the slicer, so it does not belong in this PR.

Round-5 fixed-head review to follow at `a5bb5c89`.

richlander:
## Round-5 review reconciliation (head `cb3b1c47`)

Reviewed by **MAI-Code** and **GPT** on fixed head `a5bb5c89`. Five findings, **four distinct defects, all real and all fixed**. Every claim was reproduced on a clean review worktree before being acted on.

| # | Reviewer | Finding | Verdict |
|---|---|---|---|
| 1 | MAI-Code, GPT (independently) | Sibling accessor on the line that *closes* a carried comment or literal is skipped | **Real — fixed** |
| 2 | GPT | A bare `=` after an accessor keyword is accepted, so `set = 1;` reads as a sibling | **Real — fixed** |
| 3 | GPT | Multi-line attribute list carries no state, so a line inside one reads as a declaration | **Real — fixed** |
| 4 | GPT | `IndexPastAttributeList` starts a literal at the quote, so `@"x""` closes early and a `]` in its text terminates the list | **Real — fixed** |

**Finding 1 is the mirror of the round-4 fix**, and MAI-Code and GPT found it independently. Round 4 gated the sibling check on `!InBlockComment && !InLiteral`, which is the state carried *into* the line — correct about entry, wrong as a question. `Scan` gains an `untilCodeResumes` stop and `IndexWhereCodeResumes` answers "where does this line's code start", mirroring the existing `untilLiteralCloses` stop rather than adding a second scanner. `untilLiteralCloses` could not be reused: its `opened` flag is only set when a literal opens *on that line*, so it never stops for carried state.

**Findings 3 and 4 have one root cause**, which is the more useful result of this round: `IndexPastAttributeList` was a second, simpler scanner living beside `Scan`. Both defects are things `Scan` already got right. Bracket depth moves into `LexState`, `Scan` gains an `untilBracketsClose` stop, `IndexPastAttributeList` becomes a caller, the sibling check is suppressed while brackets are open, and `EndOfSingleLineLiteral` — which existed only to serve that second scanner — is deleted.

**Finding 2** is local: the token after an accessor keyword must be `;`, `{`, `=>`, or end of line.

Also fixed, found while making the first change and not reported by either reviewer: `Scan`'s directive early-exit returned a `char` where an `int` index was expected, implicitly converting to `0`. Latent today, a trap for the next caller.

Nothing failed to reproduce this round.

**Regression tests added:** `SiblingAccessorAfterAClosingComment_StillEndsTheSlice`, `AssignmentToALocalNamedForAnAccessor_IsNotASibling`, `AccessorKeywordInsideAMultiLineAttributeList_IsNotASibling`, `LiteralsInsideAnAttributeList_DoNotTerminateIt`. The round-3/4 negatives (`AccessorKeywordInsideCommentOrLiteral_IsNotASibling`, `AttributedLocalFunction_DoesNotEndAnAccessor`, `AttributedSiblingAccessor_StillEndsTheSlice`) are unchanged and still pass — this round's fixes had to not undo them.

**Validation:** Release build of the full solution; Metadata 1012, Services 319, CLI 2216, all passing.

Head has moved, so a round-6 fixed-head review is required before this is marked ready.

richlander:
## Round-6 review reconciliation (head `2654e9c6`)

Reviewed by **GPT** and **Gemini** on fixed head `cb3b1c47`. Two findings, **both real, both fixed**.

| # | Reviewer | Finding | Verdict |
|---|---|---|---|
| 1 | GPT | A sibling accessor on the line that *closes* a multi-line attribute list is swallowed | **Real — fixed** |
| 2 | Gemini | Yielding to a sibling discards the accessor's closing brace the scan already consumed | **Real, and a regression from this PR — fixed** |

**Finding 1 is the same mirror for the third consecutive round.** Round 4 asked it of carried comments, round 5 of carried literals, round 6 of carried brackets — each time as its own suppression, each time wrong on exactly the line that closes the construct and answers the question. They are one question: *where on this line could a declaration begin?* `IndexWhereCodeResumes` now answers all three and the sibling check has no separate bracket gate left to get wrong. Both prior negatives still pass, including an attribute list that does **not** close on its line.

**Finding 2 is the more serious one, and it is a regression this PR introduced.** I verified that against `origin/main`: the base's forward scan was a naive "if the next line starts with `}`, take it", which happens to get this shape right. The depth-tracking scan this PR introduced continues past that brace looking for depth ≤ 0, and on meeting a sibling returned the range *unchanged* — discarding the brace it had just consumed along with every statement between the sequence range's end and it. The result is truncated and does not parse.

Two corrections to Gemini's report, neither of which changes the verdict:

- Its repro passed `endLine: 8`, labelled "the `return 1;` line"; line 8 is `var s = 1;`. **That exact input behaves identically on `origin/main`, so as written it is pre-existing, not a regression.** The range it *described* (`endLine: 9`) is the regression. The fix covers both, and the regression test pins `endLine` 8, 9, and 10 so the distinction cannot be lost.
- Its diagnosis said the sibling check fires "before `depth <= 0` can ever be reached". That is true but is not the defect — yielding to a sibling is correct; **discarding the recovered lines while yielding** is the defect. The fix keeps the sibling check exactly where it is and tracks the line after the member's own block closed. The sibling's own attributes and comments close nothing, which is precisely what stops them being kept as if they were the member's end — that is what keeps finding 1's case correct.

Gemini's `BracketDepth`/`!inHole` suspicion is reported honestly as **not reproduced**, and its reasoning is right: brackets inside an interpolation hole must balance, so ignoring both `[` and `]` leaves the count consistent. I did not change that guard.

**Gemini's point about the gate's 3% under-capture ceiling is fair and I am not dismissing it.** A percentage ceiling over a fixed corpus cannot catch a rare shape; per-identity baselining is the answer. That has been recorded as deferred follow-up since round 2 — it is gate accounting, not slicer behavior, and it is why every round's findings have been pinned with targeted regression tests rather than left to the corpus rates.

**Regression tests added:** `SiblingAccessorOnAMultiLineAttributesClosingLine_StillEndsTheSlice`, `SiblingAfterTheAccessorsClosingBrace_KeepsTheBrace` (3 rows).

**Validation:** Release build of the full solution; Metadata 1016, Services 319, CLI 2216, all passing.

Head has moved, so a round-7 fixed-head review is required before this is marked ready.

richlander:
## Round 7 reconciliation — MAI-Code and Gemini

Head reviewed: `2654e9c6`. Fixed at `edf7883b`.

**Both reviewers independently found the same defect**, from different source shapes. That is the strongest signal this series has produced, and it says the round-6 fix reached for the wrong instrument.

### Accepted — the `closed` tracking answered with the wrong brace (blocking)

Round 6 fixed "yielding to a sibling discards the accessor's own closing brace" by remembering the last brace that closed. MAI-Code showed that brace can be a lambda's inside an expression-bodied accessor, truncating the expression mid-way. Gemini showed the same in a conditional expression, where the tail after the nested block carries no brace at all.

I first tried to keep brace depth and scope it to "below the depth the range left open". MAI's case is fixed by that; **Gemini's is not** — its range *ends inside* the lambda, so the lambda's own close is already below the range-end depth. That disagreement is what showed the instrument itself was wrong: an expression-bodied member has no block of its own, so no brace and no depth relation can separate "the member's end" from "a block inside it".

The sibling's position can, and it does not care how either member is bodied: everything above the next declaration belongs to this member. `IndexPastTrailingTrivia` gives back only the sibling's own blank lines, comments, and attribute lists — which is what round 6 was guarding when it reached for the brace. Net effect is a simplification: `closed`, `before`, and `rangeEndDepth` are all gone.

Tests: `ABraceNestedInsideAnAccessorBody_IsNotItsEnd` (endLine 5, 10), `AnExpressionBodiedAccessorKeepsTheTailAfterANestedBlock` (endLine 6, 8), `TheSiblingsOwnLeadingTrivia_IsNotPartOfTheMember`. The round-6 test `SiblingAfterTheAccessorsClosingBrace_KeepsTheBrace` still passes unchanged.

**Correction to both reports.** Both called this pre-existing on `main`. It is not identical there: `main` truncates *earlier* (at `get => (`), so this PR was already an improvement and is now correct. Verified against `origin/main` directly.

**Fixture correction.** Gemini's repro spelled its type and property with no access modifier, which makes the backward scan capture the type header — a separate pre-existing behavior that decides the result before the forward scan runs. The committed fixture carries modifiers so it pins only this defect. Verified non-vacuous against `2654e9c6`: at endLine 6 the tail `2;` is dropped without the fix.

### Not accepted — bracket depth and interpolation format specifiers (Gemini, unverified, non-blocking)

Gemini suspected `$"{1:[]}"` leaves `BracketDepth` stuck above zero. It does not: `Scan` counts brackets only when `!inHole`, so a `[` inside an interpolation hole is never counted in the first place. Probed directly; sibling detection is unaffected.

### Validation at `edf7883b`

Release build of the full solution. Metadata **1022** (up 4), Services **319**, CLI **2216** — all passing.

richlander:
## Round 8 reconciliation (#3365)

Reviewed at `edf7883b` by **GPT** and **MAI-Code**. Both found real defects, and — notably — **the same root cause approached from opposite directions**. Fixed in `bd2e5571`.

| Finding | Reviewer | Status |
| --- | --- | --- |
| A sibling's multi-line attribute list is kept as the member's source, so the slice does not parse | GPT | **Fixed** |
| A `[`-prefixed collection-expression line is discarded as trivia, eating the member's body | MAI-Code | **Fixed** |
| A sibling's multi-line block comment / XML doc, same shape | (found while reproducing GPT's case) | **Fixed** |

### One cause, two symptoms

`HoldsOnlyTrivia` pattern-matched the line's **text**. GPT's case is a line whose text does not look like trivia but is (an attribute list's continuation); MAI's is a line whose text looks like an attribute list but is code (a collection expression). Text cannot separate them — carried lexical state can.

The fix is the same answer rounds 4 through 7 each arrived at for a different predicate: **ask the question of the line's code, not its text.** `HoldsOnlyTrivia` now takes the carried `LexState` and reuses `IndexWhereCodeResumes` and `IndexPastAttributeList` rather than matching prefixes, and returns `false` outright when a literal is open, because a carried literal is the member's code by definition.

That also let the backward walk go. `IndexPastTrailingTrivia` scanned upward from the sibling, which is the one direction in which lexical state does not exist — it is only ever computed forward. The trivia run is now tracked going forward, in `IndexPastClosingBrace`, where the state already is. Ordering matters and is asserted: the sibling check runs before the trivia update, so `HoldsOnlyTrivia` sees the state as of the line's start.

### Evidence

New fixtures, each verified to fail at `edf7883b`:

- `TheSiblingsMultiLineLeadingTrivia_IsNotPartOfTheMember` — 3 rows: multi-line attribute, block comment, XML doc
- `ACollectionExpressionLine_IsTheMembersCode_NotAnAttributeList` — 2 rows
- `ALineInsideARawStringLiteral_IsTheMembersCode_NotTrivia`

Metadata **1028**, Services **319**, CLI **2216**, all passing at `bd2e5571`. MAI's repro used `get => [1, 2];` on a `string?`, which does not compile; I used `int[]` so the fixture is real C#.

richlander:
## Round 9 reconciliation (#3365)

Reviewed at `bd2e5571` by **Gemini Pro** and **MAI-Code**. Two findings, both real, both in `HoldsOnlyTrivia`, both fixed in `3a7320ec`.

| Finding | Reviewer | Status |
| --- | --- | --- |
| A raw string inside the *sibling's own attribute* ends the trivia run, leaving the attribute in the member's slice | Gemini | **Fixed** |
| A `#region` ahead of a sibling accessor is kept as the member's code | MAI-Code | **Fixed** |

### Gemini's finding: the literal flag was the wrong instrument

Round 8 gave `HoldsOnlyTrivia` the carried `LexState`, then answered the carried-construct question with `state.InLiteral` alone — "a carried literal is the member's code". That is true only when the literal is the member's. `[Obsolete("""…""")]` on the sibling opens a literal that belongs to the *sibling*, so the run broke and the attribute survived into the slice.

**The open trivia run already answers it.** Every line since the run began held only trivia, so a construct still open is the sibling's; with no run open there is nothing above but the member, so it is the member's code. That one question covers literals and attribute brackets alike — and it is round 8's collection-expression case answered the other way, so the two are now the same rule rather than two flags that can disagree.

### MAI's finding

Preprocessor directives were never trivia in this function at all. A directive must be the first token on its line, so it is recognized once the carried constructs are accounted for.

### Scope check

The defect is confined to the **accessor** sibling path. I verified the method-sibling path and the single-line-literal-in-attribute case were already correct at `bd2e5571`, so the fix is not broader than the finding.

### Evidence

New fixtures, verified failing at `bd2e5571`:

- `ALiteralInsideTheSiblingsAttribute_DoesNotEndTheTriviaRun` — 2 rows (raw string, verbatim string)
- `ADirectiveAheadOfTheSibling_IsTrivia_NotTheMembersCode` — 3 rows (`#region`, `#if`, `#pragma`)

Round-8 non-regression re-verified explicitly: MAI's collection expression is still the member's code, GPT's multi-line sibling attribute is still excluded, and a raw string spelling `set` is still not read as a sibling.

Metadata **1033**, Services **319**, CLI **2216**, all passing at `3a7320ec`.

richlander:
## Round 10 reconciliation (#3365)

Reviewed at `3a7320ec` by **GPT** and **Gemini Pro**. Three findings, all reproduced, **all pre-existing at the merge base with `main`** — none is a regression from this branch. Pinned in `457d3e2f`.

| Finding | Reviewer | Fails at merge base? |
| --- | --- | --- |
| An inactive `#if false` region is read as live code, so a discarded `set` ends the slice | GPT | Yes |
| A block comment opened by the member's own trailing code starts a sibling trivia run | GPT | Yes |
| A constant pattern spelled `set` at the head of a switch arm is read as the sibling | Gemini | Yes |

### One of these corrects something I wrote

Round 9 excluded `InBlockComment` from the carried-construct ownership rule, and I justified it in a doc comment: *"a line inside `/* */` is trivia whoever opened it."* GPT found the case that breaks it — a comment opened by the **member's own** trailing code, whose continuation then starts a sibling run. The reasoning was wrong, and the pin says so.

### Why pinned rather than fixed

Each fix would be a further suppression on a predicate that re-derives position from a line's **text**. That is the shape of every defect in ten rounds, and all three of these stop being expressible once declarations are located over a **token stream** rather than re-derived per line: comment extent, declaration position, and directive structure are things a lexer knows by construction.

They are also not made worse by this PR — all three fail identically on `main` — so pinning them records the gap without this branch owning it. `TheSiblingQuestionAskedOfSomethingThatIsNotASibling_TruncatesTheSlice_KnownGap`, three rows.

### Round 10 is the convergence signal

**Zero regressions, and one clean review** (MAI-Code on #3366 found nothing). Rounds 4, 6, 7, 8 and 9 each found a defect introduced by the previous round's fix; round 10 found none. Every finding is a latent defect in the pre-existing line-based scanner.

Metadata **1036**, Services **319**, CLI **2216**, all passing.

`Ready to merge`

PR #3366, opened 2026-07-28
Report absent source for members with no authored declaration

richlander:
> **Stacked on #3365.** Review that first; this PR's diff is only the second commit.

## Conclusion

A positional record's property accessor, a primary constructor, and a constructor synthesized from field initializers have no declaration of their own. Their sequence points legitimately land on the enclosing type's header, and the CLI rendered that header as the member's source. This reports absence instead. Type headers rendered as member source drop **797 → 0**.

## The defect

```
$ dotnet-inspect member ... ForwarderSummaryRow TargetLibrary -S "Original Source"

## Original Source

```csharp
public record ForwarderSummaryRow(
    [property: MarkoutPropertyName("Target Library")] string TargetLibrary,
```
```

That is a truncated, syntactically invalid type declaration presented as a member's source — wrong output wearing the shape of success, which is worse than absent output because nothing signals it is wrong. It was the dominant failure the parse-validity gate in #3365 found: **six times larger** than the end-boundary defect that PR fixes.

## The change

`ExtractMethodBody` returns `null` for these, and both callers say so. The Findings path returns `Absent` with a detail string. The CLI renders a note through the same `OriginalSourceUnavailable` channel #3299 added for bodyless members, rather than dropping the section silently — keeping the failure visible:

```csharp
// This member is declared by its type's header (positional record, primary
// constructor, or field initializers), so it has no authored declaration of
// its own to show.
```

## The discriminator, and its negatives

Token-based, walking leading modifiers until it reaches a type keyword or a token that is neither. A substring test would misfire badly here — `RecordBatch`, `Classify`, `Structure`, `Interfaces`, `Enumerate`, `NewClient`, and `Recorded` all spell a type keyword inside an identifier, and each would have suppressed source for a member that has real source. Seven such negatives are pinned alongside five positives (`record`, `class`, `readonly ref struct`, `partial record struct`, `file static class`).

`namespace` is in the keyword set: a range that opens on one has walked clear past every member, and no member declaration can begin with one since none of these keywords is a legal identifier. **This was found by a positive-case fixture, not by the 4,403-member corpus** — the corpus never produced that shape, which is a concrete argument for keeping both kinds of evidence.

## Evidence

Over the test-output corpus, 4,403 members:

| | before | after |
|---|---|---|
| type header rendered as source | 797 | **0** |
| under-capture | 73 | 60 |
| absent (correct) | 0 | 1,117 |

Under-capture improved as a side effect: field-initializer constructors were counted there and now get the correct answer. Of slices that still render, **98.34% are well-formed**.

The gate asserts **both halves** — no slice may carry a type header, and the absent population must be non-empty — so it cannot pass by the corpus never reaching the path.

## Boundary / non-actions

Does not address the remaining 60 under-captured accessors (the `JsonSerializerOptions.MaxDepth` shape) or 13 start-boundary misses. Both stay under the characterized ceilings from #3365. Attacking those needs an authoritative extent oracle (Roslyn span equality), not an invariant — separate work.

## Validation

- `ILInspector.Metadata.Tests` 963/963 (5 skipped)
- `dotnet-inspect.Tests` 2216 (12 skipped)
- Verified end to end through the CLI: the positional-record case renders the note; an ordinary member still renders its source unchanged

## Review

High risk — user-visible output change on a broad member population. Requests two reviews from different model families.


richlander:
Note on CI: this PR is stacked on #3365, so its base is \`feature/slicer-parse-validity-gate\` rather than \`main\`. \`ci.yml\` triggers only on \`pull_request: branches: [main]\`, so no checks run here yet by design — this is not a CI failure. GitHub will retarget this PR to \`main\` when #3365 merges, and CI will run then.

Local validation at this head (\`cc6e9127\`), Release:

- \`dotnet build dotnet-inspect.slnx -c Release\` — clean
- \`tests/ILInspector.Metadata.Tests\` — 963/963
- \`src/DotnetInspector.Services.Tests\` — 319/319
- \`src/dotnet-inspect.Tests\` — 2216 passed, 12 skipped (ilasm/ildasm absent)

richlander:
## Review reconciliation (Gemini Pro, GPT)

Both reviewers rejected, and both were right. Three defects fixed in `9f6f1ff2`. Base updated to the fixed #3365 head via merge `de3565a0`.

Two of the three were **false absences** — the member has authored source and the tool claimed it had none. That is the failure mode that actually costs a user output, and I said so in the review prompts; the reviewers found real instances of it.

### Blocking — fixed

**1. Function-pointer return types were read as delegate declarations** (GPT, reproduced via CLI).
`public unsafe delegate*<int, int> Ret()` reaches the `delegate` keyword and was silently nulled. Now `delegate` matches only when no `*` follows. Verified on GPT exact repro:

```
$ dotnet-inspect member ...FunctionPointerShapeFixture Ret:1 --all -S "Original Source"
public delegate*<int, int> Ret(delegate*<int, int> f) => f;
```

**2. Constructors with no accessibility modifier lost their source** (GPT).
`.ctor` is not how source spells a constructor, so the backward scan walks past the declaration up to the type header — and this branch then concluded there was nothing to show. Now the check looks *below* a header for a constructor named for the type before reporting absence.

One subtlety worth calling out: the relocation has to run **before** the end boundary is decided. Measured from the type header the range still has the type block open, so the forward scan appended the type closing brace to the constructor. My first cut had exactly that bug; the test pins the dedented, correctly-bounded text.

A note on GPT repro for this one: it named `.ctor:1`, which is the record compiler-generated **copy constructor** and is correctly absent. The authored constructor is `.ctor:2`, and it now renders its full multi-line signature and body.

**3. Same-line attributes bypassed the check entirely** (Gemini *and* GPT, independently).
Skipping any line opening with `[` let `[Obsolete] public record R(int X)` render the truncated type header again — the exact output this branch exists to remove. Block-comment prefixes (`/* c */ public record R(...)`) failed the same way. Now leading attribute lists and block comments are stripped from the head of the line, counting bracket nesting and ignoring brackets inside literals, then what remains is read.

**Gemini also caught that my own test was masking this.** `RangesLandingOnATypeHeader_ReportAbsence` put `namespace N;` above the declaration under test, and `namespace` is itself a type-declaration keyword — so the first line answered every case and the declaration was never read. The declaration is now the first line, and six same-line-trivia cases fail without the fix.

That is the most valuable finding in this review: a green test that asserts nothing looks exactly like a passing one.

### Verified, not reproduced

Gemini multi-line-attribute case (`[Obsolete(\n  "why")]` above a record) already returned the correct `null` at the reviewed head — the backward scan does not capture the continuation line as the tested line. No change needed.

### Dismissed with evidence

**"The note leaks into JSON as if it were source"** (both reviewers asked). It does not. `--json` for an absent member carries no `original_source` field and no comment text; the note is a Markdown presentation concern only, which is the correct layering.

**"Corpus composition is really ~25% of members"** — GPT independently censused the newly-null population at 1,064: 842 positional-record accessors, 199 record-generated constructors, 20 primary constructors, 1 field-initializer constructor, and exactly 2 wrong ones (the function-pointer and constructor cases above). Gemini reached the same conclusion at 1,065. Both fixed.

**`OpensTypeDeclaration` false positives** — Gemini exhaustively probed contextual keywords (`record`/`file` as identifiers), `ref` returns, `new` modifiers, and cast operators, and found zero. The token-based approach holds; only `delegate*` broke it.

### Non-blocking — acknowledged

Gemini: converting a runaway backward scan into silent absence hides the runaway. Fair, and fix 2 above removes the case that motivated it. The residual is that a scan which overshoots for some *other* reason now reports absence rather than visible garbage. Worth a follow-up that distinguishes "the sequence point natively landed on the header" from "the scan walked up to it".

Metadata 979/979, Services 319/319, CLI 2216 (12 skipped).

richlander:
## Round 2 reconciliation — head `934db391`

### MAI-Code — constructor false positives (blocking, confirmed and fixed)

Reproduced exactly. `IndexOfConstructorDeclaration` matched any line whose leading token was the type's name followed by `(`, so a **statement** matched too. Three false positives, each relocating the slice into a method body and presenting a fragment of a statement as authored source:

```
"new R(1);"                       => matched   (should be absent)
"R(1);"                           => matched   (should be absent)
"Result(int x) => x.ToString();"  => matched   (MAI's original repro)
```

`new` was in the modifier list the scan walked first, so the `new R(1);` form matched *by construction*.

Spelling cannot separate a declaration from a statement, but position can: a constructor is declared at member level, a statement lives one block deeper. The scan now tracks brace depth from the type header and accepts a candidate only at depth 1, stopping rather than guessing where a brace cannot be placed. The accepted modifiers are narrowed to those a constructor can carry, which removes the `new` match at its root as well as by depth.

All four false positives are now absent; the valid local function stays absent; and the real multi-line case still resolves:

```
member ILInspector.Metadata.MetadataTypeNameFailure ... .ctor:2 -S "Original Source"
```

Gates: `ConstructorRecovery_IgnoresStatementsThatSpellTheTypeName` (5 cases) and `ConstructorRecovery_AcceptsMemberLevelDeclarations` (4 cases).

### Gemini — three findings, none reproduced

I could not reproduce any of the three, and the shared premise turns out to be an artifact of the review's own test edit.

The claim was that `RangesLandingOnATypeHeader_ReportAbsence` masked bugs, and that inserting `using System;` above the declaration fails all 11 cases. The review worktree shows why: `using System;` was inserted into that test's `source` array **without shifting `startLine`/`endLine`**, so the range now begins on the using directive rather than the type header. The test is asking a different question, and its 11 failures follow from the edit.

The decisive evidence is in the same diff. `PrimaryConstructor_StillReportsAbsence` had its line numbers shifted **correctly** (1→2, 3→4) — and it **passes**. So the one test that was re-pointed properly confirms the behavior it was offered as disproving.

Checked each finding directly against unmodified logic:

| Finding | Result |
|---|---|
| G1 verbatim `@"\b"` in an attribute skips the header | returns absent, correct |
| G2 `using` above the header aborts the search | returns absent, correct |
| G3 `[Obsolete] public class C` with a bare ctor | slices `C()\n{\n}`, correct |

Two process notes, offered as feedback rather than as a complaint: the review worktree was left with edits to `SourceLinkResolver.cs` and the test file (scratch work belongs under `/tmp/`), and the review concluded `IndexOfConstructorDeclaration` false positives were *"Extremely robust... won't false-positive"* — which is precisely the defect MAI-Code found and this commit fixes.

### Base moved

The base (#3365) took a scanner rewrite this round. The two readers this branch added moved onto it: `IndexPastAttributeList` now consumes literals through `EndOfSingleLineLiteral`, which runs the same scanner confined to one line, so no second literal parser survives.

Metadata 1000/1000, Services 319/319, CLI 2216 (12 skipped).

Head has moved, so this needs a fixed-head re-review at `934db391` before it is ready.

richlander:
## Round-3 review reconciliation

Reviewed by **MAI-Code** and **GPT** at `934db391`. Fixed head: `f3d7e35c` (which also merges the updated #3365 base).

Three defects raised, **all three reproduced, all three fixed.**

| # | Reviewer | Defect | Fix |
|---|---|---|---|
| 1 | GPT | A positional record's range spans a *secondary* constructor, which was recovered and presented as the primary constructor's source | Bound the search at the member's own first sequence point |
| 2 | MAI-Code | A comment inside an attribute list can spell `]`, closing the list early and letting a record header read as an ordinary declaration | `IndexPastAttributeList` reads `//` and `/* */` |
| 3 | GPT | `delegate *<int, int>` (spaced) read as a delegate *declaration*, so the method vanished | Skip trivia between `delegate` and `*` |

**On (1).** GPT's repro is a real compiled member in this repository (`PairFinding<T>.Present..ctor:1`), and its point is sharper than "too broad a search": returning an unrelated constructor's source is *worse* than the truncated header this PR set out to remove, because it looks right. The fix follows from what the search is for — a declaration the backward scan walked past necessarily sits at or above the member's first sequence point, so nothing below that point was ever a candidate. That is a bound the code already had the information to enforce and simply wasn't.

**On (2).** This is the third review in a row to find a line-level predicate that doesn't read comments (GPT round 2: comments in interpolation holes; Gemini round 3: comments before accessors). The pattern is now explicit in the code's comments so the next predicate added here starts from it.

**On (3).** A regression this PR introduced against its own base, caught before merge.

### Evidence

New gates: `ConstructorRecovery_IgnoresConstructorsBelowTheFirstSequencePoint`, `CommentsInsideAnAttributeList_DoNotCloseIt` (2 cases), `FunctionPointerReturnType_IsNotADelegateDeclaration` (4 spellings, including a tab). Prior regression probes for the round-2 constructor false positives all still correct.

Release: Metadata **1031/1031**, Services **319/319**, CLI **2216** (12 skipped).

One CLI test (`WriteTips_WithMinimalLevel_WritesTips`) failed once and passed both in isolation and on a full rerun; it touches tip output, not slicing. Recorded as a flake, not a finding.

Note that `ci.yml` triggers only on PRs targeting `main`, so this stacked PR gets no CI until #3365 merges.

Round-4 fixed-head review to follow at `f3d7e35c`.

richlander:
## Round-4 review reconciliation

Reviewed by **GPT** and **MAI-Code** at `f3d7e35c`. Fixed head: `dee40cc2` (which also merges the updated #3365 base).

Five defects raised. **Four reproduced and are fixed; one is a pre-existing gap I have pinned rather than fixed, and I say why below.**

| # | Reviewer | Defect | Status |
|---|---|---|---|
| 1 | MAI-Code | `class C { C() { } }` — a constructor on the header line reported **absent** | Fixed |
| 2 | GPT | `public\tclass C` — tab-separated modifiers bypassed header detection | Fixed |
| 3 | GPT | `delegate /* gap */ *<int,int>` read as a delegate declaration | Fixed |
| 4 | GPT | `C /* gap */ ()` — constructor not recovered | Fixed |
| 5 | GPT | Declaration on a multi-line attribute's **closing** line | Pinned as a known gap |

### On (1) — a regression I introduced, and the sharpest finding

MAI-Code found that round-3's `searchTo` bound, plus the pre-existing `headerIndex + 1` start, meant the header line itself was never examined. The base returns `public class C { C() { } }`; this PR returned `null`. That is exactly the failure mode this PR exists to prevent, pointed the other way: **absence claimed for a member that plainly has authored source.** The header line is now examined past its opening brace. There is no column to slice on, so the whole line is the answer — the base answer.

MAI-Code's suggested fix (start from `headerIndex`) is right about the index and not sufficient on its own: at the header line the depth is 0, so the member-level guard rejects it, and a positional record's header spells `R(` at that same depth. The discriminator has to be *within-line* position relative to the opening brace, which is what the fix uses. `ConstructorRecovery_DoesNotReadARecordHeaderAsAConstructor` pins that the record case still returns `null`.

### On (2), (3), (4)

One root cause with three faces: three separate token walks advanced over `' '` only. C# allows whitespace *and comments* between any two tokens. A single `SkipTrivia` seam now answers all three call sites, so the next token walk added here starts from it. (3) and (4) are also both refinements of round-3 fixes of mine that stopped one step short — I skipped whitespace after `delegate` but not comments, and never revisited the adjacent `(` check at all.

### On (5) — pinned, not fixed

GPT is right that `[System.Obsolete( // comment with ]` / `"why")] public record R(int X);` returns the header instead of absence. Two reasons I am not fixing it here:

- **It is not a regression.** The base behaves identically; this PR neither causes nor worsens it.
- **Closing it is a scanner change, not a discriminator change.** Attribute lists are read one line at a time; recognizing a declaration after a multi-line attribute needs bracket state carried across lines in the shared `LexState`. That belongs with the scanner and wants its own evidence.

`DeclarationOnAMultiLineAttributesClosingLine_IsNotRecognized_KnownGap` asserts today's **wrong** answer, so the gap cannot widen unnoticed and fixing it is a visible test change. This follows the existing `TypeHeaderShapes_AreNotSliceable_KnownGap` precedent in this file.

### Evidence

New gates: `ConstructorRecovery_FindsAConstructorOnTheHeaderLine`, `ConstructorRecovery_DoesNotReadARecordHeaderAsAConstructor` (2 cases), `TriviaBetweenTokens_DoesNotChangeWhatIsDeclared` (3 assertions), `DeclarationOnAMultiLineAttributesClosingLine_IsNotRecognized_KnownGap`.

Release: Metadata **1046/1046**, Services **319/319**, CLI **2216** (12 skipped). Round-2 and round-3 probes re-run and still correct.

`IndexPastAttributeList` moved **down** into #3365, which now needs it for its own fix, so this PR's diff shrinks.

Round-5 fixed-head review to follow at `dee40cc2`.

richlander:
## Round-5 review reconciliation (head `940eb83b`)

Reviewed by **MAI-Code** and **Gemini** on fixed head `dee40cc2`. Four findings, **three distinct defects, all real and all fixed**. Every claim was reproduced before being acted on.

| # | Reviewer | Finding | Verdict |
|---|---|---|---|
| 1 | MAI-Code, Gemini (independently) | `IndexOf('{')` is not lexically aware, so a brace in a comment or literal is taken for the type's | **Real — fixed** |
| 2 | Gemini | A constructor after an opening brace on the line *below* the header is missed | **Real — fixed** |
| 3 | Gemini | A constructor preceded by another member on the same line is missed | **Real — fixed** |

All three are the same shape: round 4 added an `i == start` special case that looked at the text after the header line's first brace, and rounds 4 and 5 have now each found two ways it was under-general. **The special case is deleted rather than patched a third time.** `DeclaresConstructorAtMemberLevel` asks the question at the start of the line and at every position just past a brace or semicolon — a member can only begin where the previous one ended — confirming each candidate's depth and lexical state with the shared scanner. That confirmation is what rejects the comment and literal braces, so no separate brace search is needed.

One note on MAI-Code's repro, `public class C(object? x = new { Y = 1 }) { C() { } }`: it parses cleanly, but does not compile — `error CS1736: Default parameter value for 'x' must be a compile-time constant`. It still returns `null` after this change, and I did not treat that as a defect. **The finding was right regardless**: Gemini's `class C { string s = "{"; C() { } }` is the same defect on input that compiles, and it is fixed and pinned.

Broadening the question risks accepting more shapes, so the negatives are pinned alongside the positives: `new C()` in a field initializer, a nested type's constructor, and a positional record whose default value spells `R()` inside a string all still correctly report absent.

Nothing failed to reproduce this round.

**Regression tests:** `ConstructorRecovery_FindsAConstructorSharingItsLine` (7 rows, including the round-4 header-line case, now one row of the general theory rather than its own fact), `ConstructorRecovery_FindsAConstructorAfterAnOpeningBraceBelowTheHeader`, `ConstructorRecovery_IgnoresNamesThatAreNotMemberLevelDeclarations` (3 rows).

`DeclarationOnAMultiLineAttributesClosingLine_IsNotRecognized_KnownGap` still holds and still asserts today's wrong answer: with bracket state now carried across lines (from the updated base), the closing line is entered with brackets open, so the gap is unchanged rather than accidentally widened or narrowed.

**Validation:** Release build of the full solution; Metadata 1067, Services 319, CLI 2216, all passing. Base #3365 merged at its round-5 head.

Head has moved, so a round-6 fixed-head review is required before this is marked ready.

richlander:
## Round-6 review reconciliation (head `fa1af2b6`)

Reviewed by **MAI-Code** and **GPT** on fixed head `940eb83b`. Four findings, **all real**; three fixed, one recorded as a known gap.

| # | Reviewer | Finding | Verdict |
|---|---|---|---|
| 1 | MAI-Code, GPT (independently) | "Member level" is absolute depth 1, so constructors in a namespace or a nested type are reported absent | **Real — fixed** |
| 2 | GPT | A constructor on the line that closes a carried block comment is missed | **Real — fixed** |
| 3 | GPT | The slice starts at the name, dropping a modifier on its own line and leaving a stray `*/` | **Real — fixed** |
| 4 | GPT | A declarator split between the name and its parameter list is not recognized | **Real — known gap** |

**Finding 1 is the significant one and I am glad two reviewers found it.** `IndexOfTypeDeclaration` took the *first* declaration in the captured range as the declaring type. The capture routinely runs back through enclosing scopes, so for a constructor inside a namespace it searched for the *namespace's* name, and for one in a nested type it searched for the *outer* type's — both at the wrong depth. Since essentially all real source is inside a namespace, this hid constructors broadly. It is now the innermost type still open at the target line; a namespace is never a declaring type; and depth in the search is relative to that type's header rather than absolute. Enclosing lines still contribute their lexical state, just not their depth.

**Why the corpus gate did not catch it:** the gate asserts that slices *parse*. A `null` result is not an invalid slice, so a false negative is invisible to it. That is a real limitation of the gate, and it is the same shape of blind spot as the over-capture one Gemini raised on the base PR. Both argue for the deferred per-identity baselining rather than rate ceilings.

**Findings 2 and 3 are one situation seen from two sides:** the declaration can begin above the line that spells the constructor's name. `IndexWhereCodeResumes` (from the base) supplies the candidate position after a carried comment closes, and `IndexOfDeclarationStart` backs the slice up over lines carrying only the head of the declaration. Without the second half, finding 2's fix would have produced `*/ C()` — a stray comment terminator that does not parse.

**Finding 4 is recorded, not fixed.** `C` on one line and `(` on the next is legal and is not recognized, because the match is made within a line. `ConstructorNameSplitFromItsParameterList_IsNotRecognized_KnownGap` pins today's answer so the gap cannot widen unnoticed and closing it is a visible test change, following the existing `TypeHeaderShapes_AreNotSliceable_KnownGap` precedent. Splitting *inside* the parameter list is the shape that actually occurs, and it works — that is pinned as a positive row.

One note on evidence quality: MAI-Code's round-5 repro parsed but could not compile (`CS1736`). This round both reviewers supplied inputs that compile, and MAI-Code additionally stated the compile check explicitly. That made the findings much faster to act on.

**Regression tests added:** `ConstructorRecovery_FindsTheConstructorOfTheInnermostEnclosingType` (4 rows: block-scoped namespace, file-scoped namespace, nested type, and both together), `ConstructorRecovery_StartsAtTheDeclaration_NotAtTheName` (3 rows), `ConstructorNameSplitFromItsParameterList_IsNotRecognized_KnownGap`. The round-5 negatives — `new C()` in an initializer, a nested type's constructor seen from the outer type, and a positional record whose default value spells `R()` — all still pass, which matters because finding 1's fix changes which type the search is about.

**Validation:** Release build of the full solution; Metadata 1079, Services 319, CLI 2216, all passing. Base #3365 merged at its round-6 head.

Head has moved, so a round-7 fixed-head review is required before this is marked ready.

richlander:
## Round 7 reconciliation — MAI-Code and Gemini

Head reviewed: `fa1af2b6`. Fixed at `6b43af5a`, merged with #3365's round-7 head at `fe29fa49`.

**Both reviewers independently found the same root cause**, from different source shapes.

### Accepted — a bodiless type never stopped enclosing (blocking)

Round 6 replaced "the first type declaration in the capture" with "the innermost type still open at the target". That walk decided *entered its body* from the depth left at the **end of a line**, which never becomes true for either bodiless form:

- `record R(int X);` never reaches its body depth at all (MAI-Code).
- `class Inner { }` is back below it before the line ends (Gemini).

Neither could then be popped, so a sibling type anywhere above the target held the innermost slot forever, the search ran for the wrong name at the wrong depth, and **every constructor below it was reported absent**. `enum E { A, B }`, `delegate void D();`, `struct S { }`, and `interface I { }` all had it too.

The fix drops the end-of-line-depth proxy: a declaration that ended encloses nothing, and it ends on the `;` or `}` that terminates it. The target's own line is exempt, because `class C { C() { } }` opens and closes its type there and the constructor is still its own.

Tests: `ConstructorRecovery_LooksPastASiblingTypeThatEnclosesNothing` (6 forms), `..._LooksPastABodilessSiblingTypeInsideANamespace`, `..._KeepsTheTypeThatOpensAndClosesOnTheTargetLine`, `..._FindsAConstructorUnderAMultiLineTypeHeader`.

**This is the second round-7 finding the corpus gate could not have caught**, for the reason recorded in round 6: an absent slice is not an invalid one, so a false negative reads as a pass. Same blind spot as over-capture, same argument for per-identity baselining.

### Not accepted — leading attributes and comments are dropped (Gemini, non-blocking)

Gemini reported that `[System.Obsolete]` above a constructor is excluded from the slice. Confirmed the behavior, but it is **not this PR's and not constructor-specific**: the slicer excludes leading attributes and comments for *every* member. Probed side by side —

| input | slice |
| --- | --- |
| `[System.Obsolete]` + `public C() { }` | `public C() { }` |
| `[System.Obsolete]` + `public void M() { }` | `public void M() { }` |
| `// hi` + `public C() {...}` | `public C() {...}` |
| `// hi` + `public void M() {...}` | `public void M() {...}` |

Constructors behave identically to methods, so nothing here diverged. Whether the slicer *should* carry a member's attributes is a real product question, but it is a pre-existing whole-slicer decision, not a defect this PR introduced. Filing as follow-up rather than changing it under an absent-declaration PR.

(`ConstructorRecovery_StartsAtTheDeclaration_NotAtTheName` does keep a `/* lead */` that *shares* the declaration line — that is about not leaving a stray `*/` that fails to parse, which is a different concern.)

### Investigated and reverted — non-constructor members reaching a type header

While validating the merge I found my own new #3365 fixture returning `null` here, and briefly gated the constructor search on `.ctor`. That was **wrong and is reverted**: for a non-constructor whose capture reaches a type header, `null` is the designed contract — rendering the header would present a truncated type declaration as member source. **638 corpus slices depend on it**, and `MembersWithNoAuthoredDeclaration_ReportAbsentSource_NotATruncatedTypeHeader` caught the mistake immediately. The fixture was at fault and was corrected on #3365.

### Not re-reported

Gemini's `IsDeclarationModifier` over-capture suspicion is base-owned and was filed unverified; it is not addressed here.

### Validation at `fe29fa49`

Release build of the full solution. Metadata **1094** (up 15), Services **319**, CLI **2216** — all passing.

richlander:
## Round 8 reconciliation (#3366)

Reviewed at `fe29fa49` by **GPT** and **Gemini Pro**. Three findings, all reproduced, all pinned rather than fixed — with reasons. `7e1a5702`.

| Finding | Reviewer | Status |
| --- | --- | --- |
| `class A { } class B { B() { } }` → `null` | GPT | **Pinned as known gap** — not a regression |
| `public record R(int X); C() { }` on the target line → `null` | Gemini | **Pinned** — same cause as GPT's |
| A lambda calling an inherited method named for the type is taken for the declaration | Gemini | **Pinned** |

### GPT's finding is not a regression, and the base comparison is why

GPT reported this as introduced by the bodiless-type retirement, comparing against `feature/slicer-parse-validity-gate`. That base has **no constructor recovery at all**, so *every* constructor returns `null` there and any `null` reads as a regression. Compared against `fa1af2b6` — the last commit before round 8 on this branch — the answer is `null` too. The finding is a pre-existing gap that round 8 did not touch.

### GPT's and Gemini's first two findings are one gap

A type declaration is recognized only at the head of a line. A second declaration sharing the target's line is never entered and never retired, so whichever was seen first holds the innermost slot.

Gemini proposed applying retirement before the `i == target` break. That is exactly what would retire `C` in `class C { C() { } }` and reopen the round-4 defect this branch already fixed — the two shapes are indistinguishable without intra-line declaration positions, which this scanner does not track. Both are pinned in one Theory naming the shared cause.

### Gemini's third finding: a guard was designed and rejected

`DeclaresConstructorAtMemberLevel` reads member level as brace depth 1, and a lambda raises no brace, so `C();` on a continuation line of a field initializer is accepted.

The obvious guard — a line-initial candidate is valid only if the preceding code ended at `;`, `{`, or `}` — rejects three shapes that work today: a constructor after a multi-line attribute list (ends `]`), after a lone modifier, and after a doc comment (no significant character at all). Each needs its own exception, making it a **fifth narrow suppression** on the pile that rounds 4, 5, 6, 7, and 8 each had to unwind. The shape it fixes also needs a base type exposing a member named for the derived type, because CS0542 forbids declaring one directly.

Pinned, with the real fix named: intra-line declaration positions, shared with the other two.

### Validation

Metadata **1103**, Services **319**, CLI **2216**, all passing. Includes the merge of #3365's `bd2e5571`. Note this branch is stacked and gets no CI until #3365 merges (`ci.yml` triggers only on PRs into `main`).

richlander:
## Round 9 reconciliation (#3366)

Reviewed at `7e1a5702` by **GPT** and **MAI-Code**. Two findings, both reproduced; one fixed, one pinned after its proposed fix was implemented and measured.

| Finding | Reviewer | Status |
| --- | --- | --- |
| A brace inside a type-parameter attribute retires the declaring type, returning the whole type instead of the constructor | GPT | **Fixed** in `528e249f` |
| The `i == target` exemption keeps an inner type open when it closes on the constructor's line | MAI-Code | **Pinned** — the proposed fix makes it worse |

### GPT's finding: the recurring shape, in round 8's own rule

Round 8's retirement rule ends a declaration on the `;` or `}` that terminates it, but read that character without asking **where it sat**. In `class C<[Attr(new int[] { 1 }\n)] T>`, the initializer's `}` ends nothing while the attribute's `[` is still open — so the declaring type was retired and the whole type came back in place of the constructor.

This is exactly the carried-bracket blindness the sibling question learned in **round 6**, reappearing in the rule round 8 added. Fixed by requiring the terminator to be at `BracketDepth == 0`. Both bodiless forms (`record R(int X);`, `class Inner { }`) still retire, and `class C { C() { } }` is unaffected — all three are pinned by existing tests. New fixture `ABraceInsideATypeParameterAttribute_DoesNotEndTheDeclaration`.

### MAI's finding: reproduced, fix implemented, then rejected on evidence

MAI is right that `if (i == target) break;` is blunter than its own justification. The exemption exists to protect a type declared **on** the target line (`class C { C() { } }`); it need not protect one declared above. So I implemented exactly the scoping MAI asked for — exempt only `open[^1].Index == target` — and measured it.

It does restore the enclosing type and find the constructor. The slice it then produces is:

```csharp
} Outer() { }
```

`CS8803`, `CS1002`, `CS1022`. A declaration is located **by line**, so a slice cannot begin mid-line, and every shape this scoping reaches has a brace ahead of the constructor on that line by construction. The change therefore converts an *absent* answer into a *malformed* one — the wrong direction for the gate this branch exists to satisfy.

It is also what the code did **before** round 8: I confirmed `fa1af2b6` returns `} Outer() { }` for MAI's input. So round 8 changed Malformed → Absent here, and the current answer is the improvement, not the regression. Reverted and pinned as `ATypeClosingOnTheConstructorsLine_ReportsAbsent_KnownGap`, with the measurement recorded in its doc comment.

I also verified the realistic shapes are unaffected: an inner type closing on its **own** line, with the constructor below it, recovers correctly.

### Validation

Metadata **1110**, Services **319**, CLI **2216**, all passing at `528e249f`. Includes the merge of #3365's `3a7320ec`. Both reviewers used `fa1af2b6` as the regression base correctly this round.

richlander:
## Round 10 reconciliation (#3366)

Reviewed at `528e249f` by **Gemini Pro** and **MAI-Code**.

| Finding | Reviewer | Status |
| --- | --- | --- |
| The *entered* branch of the retirement rule is unguarded, so a brace in an attribute's array initializer enters then retires the declaring type | Gemini | **Pinned** — pre-existing |
| — | MAI-Code | **No findings** |

### Gemini's diagnosis was right; its repro was not

The reported repro used `new int[] { 1 }` as a **default parameter value**, which is CS1736 (plus CS8862) — the same trap that invalidated a round-5 finding. But the diagnosis holds: I confirmed the defect with a **type-parameter attribute** whose initializer spans lines, which compiles cleanly.

It fails at `7e1a5702` and `fa1af2b6` alike, so it is not a regression from round 9. Round 9 added `state.BracketDepth == 0` to the *unentered* branch only.

### Why pinned rather than fixed

Gemini proposed guarding the entered branch too. That works for this shape, but it is the fourth guard on the same question, and it treats the symptom. **The real defect is that braces inside brackets move the block-depth counter at all.**

The scanner already knows better in the exactly analogous case: a brace inside an **interpolation hole** is counted against the hole, not the block. Extending that rule to brackets is the correct fix and is *subtractive* — but block depth feeds every consumer in this file, and the same change is subsumed by locating declarations over a token stream. Deliberately deferred, with the reasoning recorded in `ABraceInAMultiLineAttributeInitializer_RetiresTheDeclaringType_KnownGap`.

### One behavioral improvement recorded

Merging the parent branch brought up three pinned sibling-question rows. Constructor recovery **improves** one of them: the switch-arm case now reports absent source instead of returning a truncated type header. The row records it.

### Validation

Metadata **1114**, Services **319**, CLI **2216**, all passing at `fc3157c7`. Includes the merge of #3365's `457d3e2f`.

Round 10 found **zero regressions across both PRs**, and MAI's review of this branch was clean — the first clean review in ten rounds.

`Ready to merge` (after #3365)

richlander:
## Merge conflict resolved (`903e3d0e`)

**Cause:** #3365 was replayed onto a newer `main`, so this branch's copy of its thirteen commits no longer shared history with the branch it is stacked on. Nothing was wrong with either branch's content.

**Resolution:** merged the parent's new head rather than rebasing this branch, since its history is public and reviewed. Exactly one conflict — an add/add on `AuthoredSourceValidityTests.cs`, which is a new file on both sides. `SourceLinkResolver.cs` auto-merged.

Resolved to this branch's copy, which is the correct answer because the parent's two slicer files are **byte-identical** before and after its rebase (`git diff 457d3e2f 68231463` over both files is empty). This branch's copy is therefore the parent's file plus this branch's own tests.

**Verified by construction, not by inspection:**

| Check | Result |
| --- | --- |
| Resolved test file vs this branch's pre-merge version | identical |
| Auto-merged `SourceLinkResolver.cs` vs pre-merge version | identical |
| Diff vs parent's **new** head | 6 files, +1254 / -19 |
| Diff vs parent's **old** head, before the merge | 6 files, +1254 / -19 |

The last two matching is the real proof: the merge added the parent's ~8,900 lines of new `main` content without altering this branch's work by a single line. Spot-checked that `readable-names`, `MetadataImageOverview`, and `CallerScopeFilter` all survive.

**Revalidated on the merge result** (new `main` is substantial, so all four suites, not just Metadata):

| Suite | Result |
| --- | --- |
| Metadata | **1222** pass |
| Analysis | **554** pass |
| Services | **319** pass |
| CLI | **2233** pass |

Metadata rose 1114 → 1222 from `main`'s new tests, not from changes here.

Still `Ready to merge` after #3365.

richlander:
## Conflict resolved again (`cdd4eb6f`) — and here is why it recurred

**Cause:** #3365 was **squash-merged**. `main` now carries its content as one commit (`e5bca81a`) without its individual commits, so this branch's copy of that work no longer shares history with `main`. Both sides look like they independently added the same file, which is the add/add conflict in the validity tests. This is the standard way a squash merge breaks a stacked PR, and it was going to happen regardless of how the previous conflict was resolved.

**Resolution:** merged `main` (not a rebase — this branch is public and reviewed). One conflict, same file as before; `SourceLinkResolver.cs` auto-merged. Resolved to this branch's copy, because `main` holds #3365's work alone while this branch holds #3365's work **plus** constructor recovery.

**Verified by construction.** `main`'s tree is byte-identical to #3365's pre-squash head, so a correct merge must reproduce this branch's existing tree *exactly*:

| Check | Result |
| --- | --- |
| Merge result vs pre-merge head `903e3d0e` | **empty diff** — nothing lost, nothing gained |
| Merge result vs `main` | 6 files, +1254 / −19 — this branch's own work alone |

An empty diff against an already-validated tree is stronger than re-inspecting the merge: the resolution provably changed no content.

Metadata **1222** pass on the merge result. The tree is identical to the one that passed Analysis **554**, Services **319**, and CLI **2233** earlier, so those results carry over unchanged; CI is re-running them anyway.

**For the merge itself:** this is now the last PR in the stack, so a squash merge is fine here. Only #3367 remains open alongside it, and it is independent (`MERGEABLE/CLEAN`).

Still `Ready to merge`.

PR #3367, opened 2026-07-28
Adopt SourceLinkFetch source into this repository

richlander:
## Conclusion

Moves the `SourceLinkFetch` package source into this repository as `src/SourceLinkFetch`, replacing the `PackageReference` with a `ProjectReference`. The three source files are **byte-identical** to upstream; no behavior changes.

Independent of #3365 / #3366 — those touch `src/ILInspector.Metadata/SourceLinkResolver.cs`, a different file in a different project.

## Why

`SourceLinkFetch` is being reshaped around `MemberAnchor` as its primitive, so this repository is its new home rather than a consumer of it. Doing the move first, with no edits, keeps the reshaping diff honest: any later change to these files is visibly a change, not tangled with the import.

## What was verified

- Upstream `github.com/richlander/source-link-fetch` HEAD is commit `d96991b`, which is exactly the commit the published 0.1.0 package was built from — **no drift** between the packaged bits and upstream source.
- All three files confirmed `diff -q` identical to upstream before commit.
- `project.assets.json` confirms `SourceLinkFetch/1.0.0 -> type: project`, so the build genuinely resolves the local project rather than a stale package.

## Boundary / non-actions

- **Ships nothing.** The repo defaults `IsPackable=false`; only `src/dotnet-inspect` and `src/runfaster` opt back in, pinned by `PackagingSurfaceTests`. This inherits the default. Whether the library resumes publishing to nuget.org is an open question, deliberately not decided here.
- **Only `SourceLinkResolver` is consumed** (5 call sites in `ILInspector.Metadata`). `SourceLinkReader` (265 lines) and `SourceLinkVerifier` (59 lines) have no in-repo consumer today; `SourceLinkReader` overlaps heavily with `PdbContext`. They are imported as-is rather than pruned, so the import stays faithful — pruning is a reshaping decision, not an import one.
- No reshaping, no renames, no API changes.

## Validation

- `dotnet build dotnet-inspect.slnx -c Release` — succeeded
- `ILInspector.Metadata.Tests` 947/947 (5 skipped, mdv absent)
- `dotnet-inspect.Tests` 2216 (12 skipped, ilasm/ildasm absent)
- Canary `member System.Text.Json.JsonSerializerOptions MaxDepth -S "Original Source"` — byte-identical to `origin/main`
- `library --platform System.Text.Json -v:d` — SourceLink present, 355/357 tracked files resolved

## Review

Not high risk: a faithful, byte-identical import with no behavior change. Requests a single review.


richlander:
## Review reconciliation (MAI-Code)

One blocking finding, actioned in `f9dd9f4a`.

### Finding: `SourceLinkVerifier` sends unguarded requests to PDB-derived URLs

MAI-Code reproduced this against a local listener: `VerifyAsync` sends a HEAD request to `SourceDocument.ResolvedUrl` with no scheme or host validation and no timeout, against a threat model that treats artifact-derived URLs as untrusted network destinations.

**Two qualifications, both verified before acting:**

1. **It is not a regression.** The adopted files are byte-identical to upstream `d96991b`, so this code behaved exactly this way while it was a `PackageReference`. This PR does not make it worse.
2. **It has no caller.** Nothing in `src/`, `tests/`, or `tools/` names `SourceLinkVerifier`. Every source fetch in the product goes through `SourceFetcher` with `DotnetInspector.Core.HttpClientFactory.SharedUntrustedFetch` — I checked all 11 construction sites. MAI-Code repro necessarily supplied its own unguarded `HttpClient`, because the class takes one as a parameter and no product code calls it at all.

So the severity is lower than "reachable SSRF", but the underlying point stands and is the right one: adoption makes this the repository code, and repository policy now applies to it.

**Resolution: removed the file rather than hardening it.** Unreferenced code cannot be exercised, so hardening would gate a path no caller takes while leaving an unguarded entry point on the public surface. The reshaping work can reintroduce reachability checking as a guarded API if it turns out to be wanted. Project `Description` updated to match.

This does narrow the PR "byte-identical move" claim, so to be precise: the two remaining files (`SourceLinkReader.cs`, `SourceLinkResolver.cs`) are byte-identical to upstream `d96991b`, and the third was deleted rather than modified.

### Confirmed clean

MAI-Code also verified the csproj settings against sibling projects, that no leftover `PackageReference` or `PackageVersion` survives the swap, that restore is warning-free, that the adopted code adds no build warnings under this repository analyzer settings, and that `PackagingSurfaceTests` passes. No `NoWarn` suppressions were added.

Metadata 947/947, Services 319/319, CLI 2216 (12 skipped).

richlander:
Ready to merge

MAI-Code approved the fixed head `f9dd9f4a` with no findings. All CI checks pass (`changes`, `test`, `build-net10`, `csharp-diff-smoke`, `il-diff-smoke`, `pack`; `markdownlint` correctly skipped — no Markdown changed).

Round 1's single blocking finding — the unguarded network call in `SourceLinkVerifier` — was resolved by deleting the file, which had zero callers. Base is `main`, unchanged since review, so the reviewed head is the merge head.

This PR is independent of #3365 and #3366 and does not need to wait on them.

richlander:
## Updated to current `main` and revalidated (`b4f35447`)

This PR's CI last ran 19 commits ago, and `main` has since changed `src/ILInspector.Metadata/SourceLinkResolver.cs` and `AuthoredSourceAcquisition.cs` — the exact area this PR touches — via #3365 and #3366. Merged `main` in; **no conflicts**.

| Suite | Result |
| --- | --- |
| Metadata | **1222** pass |
| Services | **319** pass |
| CLI | **2233** pass |

Note there is now a `SourceLinkResolver.cs` in both `src/ILInspector.Metadata/` and `src/SourceLinkFetch/`. Different namespaces, and the solution builds clean.

## Evidence that the adoption is behavior-preserving

Rather than assert the source is byte-identical to upstream, I diffed the **artifact that was actually being consumed** — `SourceLinkFetch` 0.1.0 from the NuGet cache — against the newly built in-repo assembly, using this repo's own tool:

```
dotnet run --project src/dotnet-inspect -c Release -- diff --library \
  ~/.nuget/packages/sourcelinkfetch/0.1.0/lib/net10.0/SourceLinkFetch.dll..artifacts/bin/SourceLinkFetch/release/SourceLinkFetch.dll
```

```
Summary: 2 breaking across 2 types

### SourceLinkVerifier
- Type 'SourceLinkFetch.SourceLinkVerifier' was removed
### VerificationResult
- Type 'SourceLinkFetch.VerificationResult' was removed
```

**That is the entire difference, and both removals are deliberate** — `SourceLinkVerifier` had zero callers and was dropped during round 1 review. No additions, no signature changes, no modifications. The `PackageReference` -> `ProjectReference` switch is otherwise API-identical.

## Packaging is inert

The new `.csproj` sets `Authors`, `Description`, and `PackageLicenseExpression`, which read as packaging intent. Verified they do not ship: `dotnet pack src/SourceLinkFetch/SourceLinkFetch.csproj -c Release` produces **no `.nupkg`**, because the root `Directory.Build.props` defaults `IsPackable=false` and this project does not opt back in. Consistent with `PackagingSurfaceTests`.

A Gemini Pro review is running against this exact head; I will reconcile it here.

richlander:
## Gemini Pro review reconciliation (`b4f35447`)

One finding, **confirmed and reproduced independently**. Recommendation: approve with non-blocking notes.

### ReDoS in `SourceLinkResolver.ResolveUrl` — confirmed, High

`src/SourceLinkFetch/SourceLinkResolver.cs:50-51` builds a regex from an **attacker-controlled** string and matches with **no timeout**:

```csharp
string regexPattern = "^" + Regex.Escape(pattern).Replace("\\*", "(.*)") + "$";
var match = Regex.Match(filePath, regexPattern);
```

`pattern` is a key from the SourceLink JSON embedded in the PDB of the assembly being inspected — squarely inside `docs/design/untrusted-data-threat-model.md`. The SourceLink spec permits at most one `*` per key; nothing here enforces that, so `n` wildcards yield `n` adjacent greedy `(.*)` groups.

**My first repro did not reproduce it, and I was wrong, not Gemini.** I used a path with a single `/`, which lets the match fail fast. The ambiguity requires a slash-rich path. Corrected probe, running the exact algorithm from line 50 with no reflection:

```
path: /a/a/a/a/a/a/a/a/a/a... (len 62)
wildcards | match ms
        5 |       12
        7 |      253
        9 |     2109
       11 |     6575
       12 |    14298
```

Cleanly exponential — roughly doubling per added wildcard, on a path of only 62 characters. Real source paths are far longer. A handful more wildcards moves this from seconds to hours, on a single 2 KB JSON blob in a PDB.

**Pre-existing, not a regression.** It shipped in `SourceLinkFetch` 0.1.0, which `main` already consumes, so `main` is equally exposed today.

### Checks I ran beyond the finding

- **Only one implementation.** `ILInspector.Metadata.SourceLinkResolver.ApplySourceLinkMapping` now delegates (`_slfResolver.ResolveUrl`), so this PR does not create a second copy of the vulnerable logic. Worth stating explicitly given that the acquisition ladder *is* duplicated elsewhere in this repo with diverging checksum behavior.
- **Adjacent, informational:** `src/ILInspector.Metadata/TypeMatcher.cs:367` uses the same `Regex.Escape(p).Replace("\\*", ".*")` shape with no timeout. Narrower vector — the pattern is the user's own CLI glob, though the matched text is an attacker-supplied type name — so it needs a user-typed multi-wildcard glob to trigger. Not part of this PR.

### Confirmations

Gemini independently reproduced my two claims: `IsPackable=false` holds so `Authors`/`Description` are inert, and the merge is clean with `using SLF = SourceLinkFetch;` correctly disambiguating the two `SourceLinkResolver` types. It also attempted SSRF, redirect-following, missing-timeout and checksum attacks on the adopted code and found them **not applicable** — this library performs no HTTP; fetching lives in `AuthoredSourceAcquisition.cs`.

### On layering — Gemini is right, and it is the original complaint

`SourceLinkReader.cs` hardcodes `.cs`, `.vb`, `.fs`. `ILInspector.Metadata` still references it, so the Metadata layer still transitively knows about language spelling. The reference mechanism changed from package to project; **the layering violation is unchanged**. That is #3323's original point and it is explicitly follow-up work, not a regression here.

### Disposition

Merging as-is, with a security fix to follow immediately as its own PR. This PR's entire value is the byte-identity claim and the API-diff evidence backing it; editing the source here would forfeit that. Adopting the code is precisely **what makes the fix possible** — as a NuGet package it was not ours to patch.

The follow-up will apply both defenses: reject patterns with more than one `*` (spec conformance) and pass an explicit match timeout (defence in depth), with the exponential probe above as a regression test.

`Ready to merge`

PR #3370, opened 2026-07-28
Cache the whole-graph maps behind the call-tree builders (#3342)

### LINKED ISSUE #3342: Fan-in dedupe allocates HashSets for the whole assembly graph on every call-tree request
Increment of #3333.

#3312 converted callee-side fan-in from call-site counts to **distinct caller** counts. The dedupe is correct, but it is computed eagerly for the **entire assembly graph** on **every** `BuildCallTree` request, allocating a `HashSet` per callee where the previous code did a flat integer count.

- `src/ILInspector.Analysis/LibraryBodyIndex.cs` ~line 1354 (single-assembly path)
- `src/ILInspector.Analysis/LibraryBodyIndex.cs` ~line 1716 (cross-assembly path)

Only a bounded number of nodes are ever visited (`maxDepth 3`, `maxNodes 25`), so nearly all of that allocation is discarded. Raised as **High** by Gemini 3.1 Pro during round-3 review of #3312, specifically as a hazard for the interactive wasm consumer, where GC pressure and peak memory are the binding constraints.

Independently corroborated while writing tests for #3331: a test that builds caller trees for 60 methods of one assembly takes ~10s, and 300 methods takes ~35s — superlinear in a way that only makes sense if each request rebuilds whole-graph state.

**Options**

1. Cache distinct incoming counts lazily on `LibraryBodyIndex`, the way `GetDirectCallsByCaller()` already caches. Cheapest change, amortizes across requests on one index.
2. Dedupe only for nodes actually visited during the bounded expansion. Avoids whole-graph work entirely, but has to keep the counts consistent with what is displayed.

Option 1 is likely the better first move because it keeps the counting logic in one place.

**Evidence expected**

Fan-in values unchanged across the existing suites, plus a measured before/after on time and peak working set for a representative `member ... -S "Call Graph"` request. Do not assert allocation reduction from reading the code — measure it.

richlander:
Closes #3342.

Every call-tree request rebuilt the whole-assembly maps it needs, even though a bounded tree (`maxDepth 3`, `maxNodes 25`) only visits a few dozen nodes. Those maps are pure functions of `DirectCalls`/`Methods` — and, for the cross-assembly path, of the scope set — so they are now lazily cached on the index. A consumer that asks many questions of one index now pays for the graph once.

## Measured

One index, many requests — the shape the wasm consumer has. Output checksums identical in every row.

| shape | base | head |
| --- | --- | --- |
| 60 members, single assembly | 21246 ms / 13388 MB alloc | **230 ms / 93 MB** |
| 300 members, single assembly | 106445 ms / 66943 MB alloc | **262 ms / 95 MB** |
| 20 members, 22 scopes | 63681 ms / 29419 MB alloc / 1015 MB peak | **3341 ms / 1351 MB alloc / 789 MB peak** |

Base scales exactly linearly in member count (5x members -> 5.0x time); head is flat.

**A single CLI invocation shows no win** — 4681 ms -> 4617 ms, 340 -> 345 MB, output byte-identical. There the cost is decoding 23 assemblies, not rebuilding maps. The win is entirely in the repeated-request shape, so measuring this through the CLI alone would have shown nothing.

## What changed

`LibraryBodyIndex` gains four lazy caches using the same `_field ??= ...` pattern `GetDirectCallsByCaller()` already used: the definition map, distinct callers by callee, distinct caller edges by callee, and a scope-keyed `ScopeGraph` for the cross-assembly forward/reverse maps. `BuildCallTree` also stopped rebuilding a grouping `GetDirectCallsByCaller()` already provides.

## New public surface

Two methods on `LibraryBodyIndex`, added in `ac557157` in response to round-1 review:

| Method | Drops |
| --- | --- |
| `ReleaseScopeGraph()` | the cross-assembly scope graph only |
| `ReleaseCallGraphCaches()` | that, plus the definition map, distinct-caller counts and edges, and the direct-call grouping |

They are separate because the two halves have opposite economics. The single-assembly caches retain ~75 MB and take peak working set *below* base (451 vs 471 MB) while being 34x faster; the cross-assembly graph is the only real memory trade, and one root plus 22 scopes retains ~370 MB of it. A consumer that keeps visiting members of one assembly wants `ReleaseScopeGraph()` between renders, not the full drop.

Everything rebuilds on next use, so both only trade time for memory. `ReleaseScopeGraph()` returns live bytes to 2.1 MB — exactly base — with output checksums identical.

`ReleaseMethods_DropExactlyTheCachesTheyDocument` derives the type's twelve mutable cache fields by reflection and pins each to a side of that boundary, so the doc cannot drift from the code and a newly added cache fails until someone decides which side it is on.
## The one case deliberately left uncached

A **bodiless root** is absent from `Methods`, so edges naming it resolve onto a *different* defined method and never land in a bucket keyed by the root. `BuildCallerTree` therefore uses the shared reverse index only when `methodMap.ContainsToken(rootMethodToken)`, and builds its own map otherwise.

Mutation-tested: removing that guard leaves the new test green but fails the pre-existing `BuildCallerTree_ResolvesCallers_WhenSelectedRootIsBodilessInterfaceMethod`. That split is correct and is now stated in the new test's comment — serving a bodiless root from the shared grouping is a *correctness* fault that a fresh index reproduces identically, so no cache-independence test can see it.

I also checked the reverse direction, where a bodiless request *poisons* the shared cache. That mutation survives, and it is an **equivalent mutant**: the definition-map key includes the declaring type, so `Resolve` always returns 0 for a bodiless callee and the leaked entry is purely additive. Predicted from the key structure, then confirmed empirically rather than assumed.

## Two corrections to the issue

1. **Not superlinear.** The issue calls the cost superlinear, but its own numbers are sublinear per method (167 ms/method at 60, 117 at 300). My scaling run confirms base is *linear* in member count. The problem is the constant factor, not the growth rate.
2. **Option 2 as written is a trap.** Narrowing fan-in to the visited subset would silently redefine the metric. Fan-in must remain true whole-graph inbound degree, mirroring fan-out's existing contract ("independent of how far the bounded tree expanded"). This PR keeps the metric identical and only removes recomputation.

The issue also **understates scope**: all four builders rebuild whole-graph state, and the two most expensive were not listed.

## Test

`CallTreeBuilders_AreUnaffectedByEarlierRequestsOnTheSameIndex` walks one index through a sequence designed to poison a root- or scope-dependent cache and requires every answer to match a fresh index asked nothing else.

It is written to resist going vacuous, because **my first version did**: it compared `LibraryBodyIndex.Open`, which has no in-assembly callers, so it rendered one leaf line and would have held no matter how badly a cache leaked. It now self-calibrates — it scans for a root with a genuinely branching tree and asserts `richest >= 4` — and it asserts `expectedScoped != expectedUnscoped` so the scope comparisons cannot pass by both sides being equally empty.

Mutation-tested: forcing `ScopeGraph.Matches` to `return true` fails it.

## Validation

- `dotnet build dotnet-inspect.slnx -c Release` — 0 Error(s)
- Analysis: **556 / 1 failed** — `CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition`, verified pre-existing on base `d4002cf1` by building and running that commit
- CLI: **2229 / 1 failed / 10 skipped** — baseline
- CLI output byte-identical base vs head for `open` (63 lines) and the caller-graph key case (64 lines)

### DIFF
--- a/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
+++ b/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
@@ -316,6 +316,109 @@ public class LibraryBodyIndexTests
             child.Member.Name == nameof(BodilessRootFixtures.InvokesThroughInterface));
     }
 
+    // #3342: the whole-graph maps behind the call-tree builders (definition map, distinct-caller
+    // counts, reverse edges, and the cross-assembly scope maps) are cached per index so a
+    // consumer that asks many questions of one index pays for them once. That is only sound if
+    // none of them depends on which member was selected or on a previously requested scope set.
+    // This walks one index through a sequence designed to poison a root- or scope-dependent
+    // cache, and requires every answer to match a fresh index that was asked nothing else.
+    //
+    // Scope note: this pins cache *independence*, not the bodiless-root contract itself. Serving
+    // a bodiless root from the shared grouping is a correctness fault that a fresh index would
+    // reproduce identically, so no independence test can see it;
+    // BuildCallerTree_ResolvesCallers_WhenSelectedRootIsBodilessInterfaceMethod owns that, and
+    // was confirmed to fail when the guard is removed.
+    [Fact]
+    public void CallTreeBuilders_AreUnaffectedByEarlierRequestsOnTheSameIndex()
+    {
+        string analysisPath = typeof(LibraryBodyIndex).Assembly.Location;
+        string testPath = typeof(LibraryBodyIndexTests).Assembly.Location;
+
+        // Pick a root with a genuinely branching caller tree. Comparing trees is only evidence
+        // if the trees can differ: a root with no in-assembly callers renders one leaf line, and
+        // every comparison below would hold no matter how badly a cache leaked.
+        var probe = LibraryBodyIndex.Open(analysisPath);
+        int richToken = PickRichest(probe, out int richest);
+
+        Assert.True(richest >= 4, $"expected a branching caller tree to compare; richest had {richest} nodes");
+        int otherToken = probe.Methods.First(method => method.MetadataToken != richToken).MetadataToken;
+        // A bodiless root is the one case whose caller grouping genuinely depends on the root,
+        // so it must not be served from (or poison) the shared per-index cache.
+        int bodilessToken = typeof(ICallerGraphTarget).GetMethod(nameof(ICallerGraphTarget.Target))!.MetadataToken;
+
+        // Each expectation comes from an index that has answered nothing else.
+        var expectedCallers = Describe(LibraryBodyIndex.Open(analysisPath).BuildCallerTree(richToken, maxDepth: 2, maxNodes: 50));
+        var expectedCallees = Describe(LibraryBodyIndex.Open(analysisPath).BuildCallTree(richToken, maxDepth: 2, maxNodes: 50));
+        var expectedScoped = Describe(LibraryBodyIndex.Open(analysisPath)
+            .BuildCallerTree(richToken, new[] { LibraryBodyIndex.Open(testPath) }, maxDepth: 2, maxNodes: 50));
+        var expectedUnscoped = Describe(LibraryBodyIndex.Open(analysisPath)
+            .BuildCallerTree(richToken, Array.Empty<LibraryBodyIndex>(), maxDepth: 2, maxNodes: 50));
+        var expectedBodiless = Describe(LibraryBodyIndex.Open(testPath).BuildCallerTree(bodilessToken, maxDepth: 2, maxNodes: 50));
+
+        // Scoping must actually change the answer, or the scope-set comparisons prove nothing.
+        Assert.NotEqual(expectedScoped, expectedUnscoped);
+
+        // Now ask one index everything, in an order that would expose a leaked root or scope.
+        var reused = LibraryBodyIndex.Open(analysisPath);
+        var scopeA = LibraryBodyIndex.Open(testPath);
+        reused.BuildCallerTree(otherToken, maxDepth: 2, maxNodes: 50);
+        reused.BuildCallTree(otherToken, maxDepth: 2, maxNodes: 50);
+        reused.BuildCallerTree(otherToken, new[] { scopeA }, maxDepth: 2, maxNodes: 50);
+        reused.BuildCallerTree(otherToken, Array.Empty<LibraryBodyIndex>(), maxDepth: 2, maxNodes: 50);
+
+        Assert.Equal(expectedCallers, Describe(reused.BuildCallerTree(richToken, maxDepth: 2, maxNodes: 50)));
+        Assert.Equal(expectedCallees, Describe(reused.BuildCallTree(richToken, maxDepth: 2, maxNodes: 50)));
+
+        // Alternating scope sets on one index: the empty set must not be answered from the
+        // populated set's map, nor the reverse, however many times they interleave.
+        Assert.Equal(expectedScoped, Describe(reused.BuildCallerTree(richToken, new[] { scopeA }, maxDepth: 2, maxNodes: 50)));
+        Assert.Equal(expectedUnscoped, Describe(reused.BuildCallerTree(richToken, Array.Empty<LibraryBodyIndex>(), maxDepth: 2, maxNodes: 50)));
+        Assert.Equal(expectedScoped, Describe(reused.BuildCallerTree(richToken, new[] { scopeA }, maxDepth: 2, maxNodes: 50)));
+
+        // A bodiless root asked after body-rooted requests have populated the cache.
+        var reusedTests = LibraryBodyIndex.Open(testPath);
+        int testAsmToken = PickRichest(reusedTests, out int testRichest);
+        Assert.True(testRichest >= 2, $"expected a non-leaf caller tree in the test assembly; richest had {testRichest} nodes");
+        var expectedTestAsmCallers = Describe(LibraryBodyIndex.Open(testPath).BuildCallerTree(testAsmToken, maxDepth: 2, maxNodes: 50));
+        reusedTests.BuildCallerTree(testAsmToken, maxDepth: 2, maxNodes: 50);
+        Assert.Equal(expectedBodiless, Describe(reusedTests.BuildCallerTree(bodilessToken, maxDepth: 2, maxNodes: 50)));
+
+        // ...and the bodiless request must not have written its root-dependent grouping into the
+        // shared cache, so a body-rooted request after it still answers like a fresh index.
+        Assert.Equal(expectedTestAsmCallers, Describe(reusedTests.BuildCallerTree(testAsmToken, maxDepth: 2, maxNodes: 50)));
+
+        static int PickRichest(LibraryBodyIndex index, out int richest)
+        {
+            int token = 0;
+            richest = 0;
+            foreach (var method in index.Methods.Take(150))
+            {
+                int size = Describe(index.BuildCallerTree(method.MetadataToken, maxDepth: 2, maxNodes: 50)).Count;
+                if (size > richest)
+                {
+                    richest = size;
+                    token = method.MetadataToken;
+                }
+            }
+
+            return token;
+        }
+
+        static List<string> Describe(CallTreeNode root)
+        {
+            var lines = new List<string>();
+            void Walk(CallTreeNode node, int depth)
+            {
+                lines.Add($"{depth}|{node.Member.DeclaringType.Name}.{node.Member.Name}|{node.Status}|" +
+                    $"{node.Perf?.Source ?? ""}|fanout={node.Perf?.Fanout ?? -1}|fanin={node.Perf?.Fanin ?? -1}");
+                foreach (var child in node.Children)
+                    Walk(child, depth + 1);
+            }
+            Walk(root, 0);
+            return lines;
+        }
+    }
+
     [Fact]
     public void BuildCallerTree_WithScope_IncorporatesAndTagsExternalCallers()
     {
--- a/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
+++ b/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
@@ -339,8 +339,10 @@ public class LibraryBodyIndexTests
         // Evidence-domain caches the release methods deliberately retain.
         string[] retainedCaches =
         [
+            "_allocationFanoutOpportunities",
             "_directCallerLoops",
             "_generatedFrameworkTypes",
+            "_opportunities",
             "_rootReachByToken",
             "_signals",
             "_unsafeEvidenceByMember",
@@ -383,17 +385,31 @@ public class LibraryBodyIndexTests
             return index;
         }
 
+        // Every non-readonly instance field on this type is a lazy cache, so the filter is just
+        // "mutable state" rather than a type-shaped guess. An earlier version excluded value types
+        // and so silently missed the two ImmutableArray caches, which is the same blind spot this
+        // test exists to catch.
         static IEnumerable<FieldInfo> MutableCacheFields()
             => typeof(LibraryBodyIndex)
-                .GetFields(BindingFlags.NonPublic | BindingFlags.Instance)
-                .Where(field => field.Name.StartsWith('_') && !field.IsInitOnly && !field.FieldType.IsValueType);
+                .GetFields(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance)
+                .Where(field => field.Name.StartsWith('_') && !field.IsInitOnly);
 
         static List<string> PopulatedCaches(LibraryBodyIndex index)
             => MutableCacheFields()
-                .Where(field => field.GetValue(index) is not null)
+                .Where(field => IsPopulated(field.GetValue(index)))
                 .Select(field => field.Name)
                 .OrderBy(name => name, StringComparer.Ordinal)
                 .ToList();
+
+        // ImmutableArray caches signal "not yet computed" with IsDefault, not with null.
+        static bool IsPopulated(object? value)
+        {
+            if (value is null)
+                return false;
+
+            var isDefault = value.GetType().GetProperty("IsDefault");
+            return isDefault is null || !(bool)isDefault.GetValue(value)!;
+        }
     }
 
     // #3342: the whole-graph maps behind the call-tree builders (definition map, distinct-caller
--- a/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
+++ b/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
@@ -316,6 +316,86 @@ public class LibraryBodyIndexTests
             child.Member.Name == nameof(BodilessRootFixtures.InvokesThroughInterface));
     }
 
+    /// <summary>
+    /// Derives this type's mutable cache fields by reflection and pins each one to a side of the
+    /// release boundary, so the two release methods are gated on the property they exist for
+    /// rather than only on staying correct. Both directions fail: a cache the doc claims to drop
+    /// but doesn't, and a cache dropped that the doc says survives. A newly added cache field
+    /// belongs to neither list and fails until someone decides which side it is on.
+    /// </summary>
+    [Fact]
+    public void ReleaseMethods_DropExactlyTheCachesTheyDocument()
+    {
+        // Dropped by ReleaseCallGraphCaches(); _scopeGraph is also dropped by ReleaseScopeGraph().
+        string[] callGraphCaches =
+        [
+            "_directCallsByCaller",
+            "_distinctCallerEdgesByCallee",
+            "_distinctCallersByCallee",
+            "_methodMap",
+            "_scopeGraph",
+        ];
+
+        // Evidence-domain caches the release methods deliberately retain.
+        string[] retainedCaches =
+        [
+            "_directCallerLoops",
+            "_generatedFrameworkTypes",
+            "_rootReachByToken",
+            "_signals",
+            "_unsafeEvidenceByMember",
+        ];
+
+        Assert.Equal(
+            callGraphCaches.Concat(retainedCaches).OrderBy(name => name, StringComparer.Ordinal),
+            MutableCacheFields().Select(field => field.Name).OrderBy(name => name, StringComparer.Ordinal));
+
+        string analysisPath = typeof(LibraryBodyIndex).Assembly.Location;
+        string testPath = typeof(LibraryBodyIndexTests).Assembly.Location;
+        var scope = LibraryBodyIndex.Open(testPath);
+
+        var index = Exercised(analysisPath, scope);
+        var before = PopulatedCaches(index);
+
+        // The gate is only meaningful if the caches under test were populated to begin with.
+        foreach (var name in callGraphCaches)
+            Assert.Contains(name, before);
+
+        index.ReleaseCallGraphCaches();
+        Assert.Equal(before.Where(name => !callGraphCaches.Contains(name)), PopulatedCaches(index));
+
+        // ReleaseScopeGraph drops the scope graph and nothing else — that is the whole point of
+        // having it separately, since the single-assembly maps are the cheap, high-value half.
+        var scopeOnly = Exercised(analysisPath, scope);
+        var beforeScopeRelease = PopulatedCaches(scopeOnly);
+        Assert.Contains("_scopeGraph", beforeScopeRelease);
+
+        scopeOnly.ReleaseScopeGraph();
+        Assert.Equal(beforeScopeRelease.Where(name => name != "_scopeGraph"), PopulatedCaches(scopeOnly));
+
+        static LibraryBodyIndex Exercised(string path, LibraryBodyIndex scope)
+        {
+            var index = LibraryBodyIndex.Open(path);
+            int token = index.Methods.First().MetadataToken;
+            index.BuildCallerTree(token, maxDepth: 2, maxNodes: 50);
+            index.BuildCallTree(token, maxDepth: 2, maxNodes: 50);
+            index.BuildCallerTree(token, new[] { scope }, maxDepth: 2, maxNodes: 50);
+            return index;
+        }
+
+        static IEnumerable<FieldInfo> MutableCacheFields()
+            => typeof(LibraryBodyIndex)
+                .GetFields(BindingFlags.NonPublic | BindingFlags.Instance)
+                .Where(field => field.Name.StartsWith('_') && !field.IsInitOnly && !field.FieldType.IsValueType);
+
+        static List<string> PopulatedCaches(LibraryBodyIndex index)
+            => MutableCacheFields()
+                .Where(field => field.GetValue(index) is not null)
+                .Select(field => field.Name)
+                .OrderBy(name => name, StringComparer.Ordinal)
+                .ToList();
+    }
+
     // #3342: the whole-graph maps behind the call-tree builders (definition map, distinct-caller
     // counts, reverse edges, and the cross-assembly scope maps) are cached per index so a
     // consumer that asks many questions of one index pays for them once. That is only sound if
--- a/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
+++ b/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
@@ -322,6 +322,10 @@ public class LibraryBodyIndexTests
     /// rather than only on staying correct. Both directions fail: a cache the doc claims to drop
     /// but doesn't, and a cache dropped that the doc says survives. A newly added cache field
     /// belongs to neither list and fails until someone decides which side it is on.
+    /// <para>
+    /// Known boundary: this compares populated-ness, so it would not see a future release method
+    /// that empties a readonly collection in place instead of nulling a field.
+    /// </para>
     /// </summary>
     [Fact]
     public void ReleaseMethods_DropExactlyTheCachesTheyDocument()
@@ -360,7 +364,9 @@ public class LibraryBodyIndexTests
         var before = PopulatedCaches(index);
 
         // The gate is only meaningful if the caches under test were populated to begin with.
-        foreach (var name in callGraphCaches)
+        // Both halves need this: an unpopulated cache is absent from `before` and from `after`,
+        // so the set comparison would hold no matter what the release methods did to it.
+        foreach (var name in callGraphCaches.Concat(retainedCaches))
             Assert.Contains(name, before);
 
         index.ReleaseCallGraphCaches();
@@ -382,6 +388,15 @@ public class LibraryBodyIndexTests
             index.BuildCallerTree(token, maxDepth: 2, maxNodes: 50);
             index.BuildCallTree(token, maxDepth: 2, maxNodes: 50);
             index.BuildCallerTree(token, new[] { scope }, maxDepth: 2, maxNodes: 50);
+
+            // The retained half of the contract is only gated on caches this workload actually
+            // populates, and the call-tree builders alone reach just one of the seven. Touch the
+            // evidence-domain producers too; OptimizationOpportunities is what pulls in
+            // _directCallerLoops and _rootReachByToken, which have no direct test access.
+            _ = index.OptimizationOpportunities;
+            _ = index.AllocationFanoutOpportunities;
+            _ = index.GetUnsafeEvidenceByMember();
+            _ = index.GeneratedFrameworkTypeNames;
             return index;
         }
 
--- a/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
+++ b/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
@@ -387,6 +387,32 @@ public class LibraryBodyIndexTests
         // shared cache, so a body-rooted request after it still answers like a fresh index.
         Assert.Equal(expectedTestAsmCallers, Describe(reusedTests.BuildCallerTree(testAsmToken, maxDepth: 2, maxNodes: 50)));
 
+        // The scope cache keys on the caller's collection, which the caller still owns and may
+        // mutate. Holding that live list would make the key compare equal to itself while the
+        // built maps still describe the old contents, so the scope references are snapshotted.
+        var mutableScopes = new List<LibraryBodyIndex> { scopeA };
+        var mutated = LibraryBodyIndex.Open(analysisPath);
+        mutated.BuildCallerTree(richToken, mutableScopes, maxDepth: 2, maxNodes: 50);
+        mutableScopes[0] = LibraryBodyIndex.Open(analysisPath);
+        var expectedAfterMutation = Describe(LibraryBodyIndex.Open(analysisPath)
+            .BuildCallerTree(richToken, mutableScopes, maxDepth: 2, maxNodes: 50));
+
+        // Swapping the element must actually change the answer, or this proves nothing.
+        Assert.NotEqual(expectedScoped, expectedAfterMutation);
+        Assert.Equal(expectedAfterMutation, Describe(mutated.BuildCallerTree(richToken, mutableScopes, maxDepth: 2, maxNodes: 50)));
+
+        // Releasing the caches is a memory/time trade only: answers after a release must still
+        // match a fresh index, and must not be served from a map that was supposedly dropped.
+        var released = LibraryBodyIndex.Open(analysisPath);
+        released.BuildCallerTree(richToken, new[] { scopeA }, maxDepth: 2, maxNodes: 50);
+        released.BuildCallTree(richToken, maxDepth: 2, maxNodes: 50);
+        released.ReleaseScopeGraph();
+        Assert.Equal(expectedScoped, Describe(released.BuildCallerTree(richToken, new[] { scopeA }, maxDepth: 2, maxNodes: 50)));
+        released.ReleaseCallGraphCaches();
+        Assert.Equal(expectedCallers, Describe(released.BuildCallerTree(richToken, maxDepth: 2, maxNodes: 50)));
+        Assert.Equal(expectedCallees, Describe(released.BuildCallTree(richToken, maxDepth: 2, maxNodes: 50)));
+        Assert.Equal(expectedScoped, Describe(released.BuildCallerTree(richToken, new[] { scopeA }, maxDepth: 2, maxNodes: 50)));
+
         static int PickRichest(LibraryBodyIndex index, out int richest)
         {
             int token = 0;
--- a/src/ILInspector.Analysis/LibraryBodyIndex.cs
+++ b/src/ILInspector.Analysis/LibraryBodyIndex.cs
@@ -1106,6 +1106,14 @@ public sealed class LibraryBodyIndex
         Dictionary<string, List<ReverseCallerEdge>>? _reverse;
 
         /// <summary>True when this cache was built for exactly the same scope instances, in order.</summary>
+        /// <remarks>
+        /// The root is part of the key because it is itself a graph participant, not merely the
+        /// object holding the cache. That clause cannot currently fail — the only storage is
+        /// <c>_scopeGraph</c> on the same instance <c>ScopeGraphFor</c> passes as <c>this</c> — so it
+        /// is an uncoverable branch, worth knowing if branch coverage here is ever read as
+        /// meaningful. It stays because dropping it would leave the key incomplete rather than
+        /// simplified, and silently wrong if a graph were ever cached anywhere but its own root.
+        /// </remarks>
         public bool Matches(LibraryBodyIndex requestRoot, IReadOnlyList<LibraryBodyIndex> requested)
         {
             if (!ReferenceEquals(root, requestRoot) || _scopes.Length != requested.Count)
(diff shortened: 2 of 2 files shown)


richlander:

Two reviewers on the fixed head `c26996d5`, isolated worktrees: **Gemini 3.1 Pro** and **GPT-5.6**. Fixes are in `ac557157`.

### 1. Scope-list aliasing — FIXED

Found **independently by both** (Gemini: Blocking, GPT: Important), each with an executed reproduction. `ScopeGraph` retained the caller's live `IReadOnlyList`, so `Matches` compared it against itself and a mutated element kept serving maps built for the old contents.

Credit where due — this is a real bug I missed, and my own cache-independence test could not have caught it because it only ever passed freshly allocated arrays.

On investigating I found it is **worse than either reviewer reported**: `Participants()` is a lazy iterator, so the aliasing also let a mutation *between construction and first enumeration* change which assemblies were indexed. Fixed by snapshotting into an `ImmutableArray`.

Regression test added, with a sensitivity assert that swapping the element actually changes the answer. Against pre-fix `c26996d5` it fails on a stale `fanin=15` where `12` is correct.

### 2. Retention under a hard memory ceiling — ADDRESSED

Both flagged this; I reproduced it at larger scale than either. One root, 22 scopes, one bounded query:

| | base | head `c26996d5` | head + `ReleaseScopeGraph()` |
| --- | --- | --- | --- |
| live after query | 2.1 MB | 442.8 MB | **2.1 MB** |

Checksums identical across all three.

Gemini's suggestion to discard the complementary direction would not work — a progressive render asks both directions for the same scope set, so it would thrash. Instead the consumer gets explicit control:

- `ReleaseScopeGraph()` — drops the cross-assembly maps, which are the bulk and are specific to one scope set.
- `ReleaseCallGraphCaches()` — drops everything derived.

Kept separate deliberately, because the two halves have very different economics:

| | retained | peak WS vs base | speed |
| --- | --- | --- | --- |
| single-assembly caches | 75 MB | **below** base (451 vs 471 MB) | 51242 → 1494 ms (34x) |
| cross-assembly ScopeGraph | ~368 MB | above base for one-shot, below for many (789 vs 1015 MB) | 63681 → 3341 ms (19x) |

So a consumer visiting many members of one library keeps the cheap, strictly-better half and drops only the expensive scope-set-specific half between renders.

### 3. Safe publication on ARM64 (Gemini, Important) — DOCUMENTED, not code-changed

The concern is fair in principle, but this is not a risk this PR introduces: `_directCallsByCaller` and the other evidence accessors already published caches with unsynchronized `??=` before this change. Adding barriers only to the new fields would imply a thread-safety guarantee the type does not otherwise make.

I documented the actual contract instead — lazily populated caches mean an instance is not safe for concurrent use without external synchronization. If we want `LibraryBodyIndex` to be genuinely thread-safe that is a separate, wider change.

### 4. Bodiless-root guard — CLEARED

Gemini attacked this hardest and cleared it, comparing `Resolve` against the local path across all 500+ MethodDef tokens with a stated sensitivity control, concluding the guard is complete. That matches my own derivation: `Resolve` *starts* with `if (_methodTokens.Contains(CalleeDefinitionToken)) return CalleeDefinitionToken`, so the paths can only diverge when the root is absent from `Methods`.

### Not accepted

Nothing else was raised. Neither reviewer found a cache-poisoning sequence beyond the aliasing bug, a fan-in/fan-out semantic change, or a surviving mutation of the guards.

### Validation at `ac557157`

- Full solution build `0 Error(s)`.
- Analysis **555 / 1 failed** — the known pre-existing `CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition`.
- Release measured to return the live set to base exactly, checksums identical.

Re-review of `ac557157` to follow.

richlander:
## Re-review round 2 — both clean

Both reviewers re-ran on the fixed head `ac557157` in fresh isolated worktrees.

**Gemini 3.1 Pro:** *"No significant issues found."* Re-attacked the scope key with in-place mutation, reordering, duplicates and `this`; confirmed the snapshot yields a correct cache miss rather than corruption, with a stated sensitivity control against its own pre-fix probe. Also grepped every consumer of `GetDirectCallsByCaller()` and confirmed the only external one (`AnalysisFindings.InspectCallSites`) never depends on reference stability of the returned dictionary, so clearing it in `ReleaseCallGraphCaches()` is safe. Accepted the thread-safety documentation as the correct scope, noting that clearing a field is an atomic reference write whose worst case is duplicated recomputation, not corruption.

**GPT-5.6:** *"No significant issues found."* Independently removed the snapshot in scratch, verified the mutant was present and the build was `0 Error(s)`, and confirmed the new regression test fails without it (`fanin=12` vs `15`). Verified release/rebuild outputs are identical, that `ReleaseScopeGraph()` frees ~162 MiB of cross-assembly maps in its own smaller shape, that `ReleaseCallGraphCaches()` clears all five intended fields, and that optimization opportunities are unchanged across release/rebuild. Noted it did not reproduce the 22-scope timing measurement.

## One pre-existing bug found, filed separately

Gemini found that passing an assembly as its own scope — or repeating a scope — makes `Participants()` index it twice, doubling `Fanout`. It verified the same multiplier on base `d4002cf1`, and I independently confirmed the base code shape: `IndexAssembly(this)` followed by an un-deduplicated scope loop.

This PR moved that logic into `ScopeGraph.Participants()` and **preserves the behavior exactly**, which is the correct outcome for a caching change. Filed as #3374 rather than fixed here, so the intended semantics get decided deliberately.

## Status

- Two different-family reviews, both clean on the exact head `ac557157`.
- CI green (`test`, `build-net10` pass).
- Analysis **555 / 1** — known pre-existing `CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition`.
- CLI **2229 / 1 / 10 skipped** — baseline.

Ready to merge.

Non-blocking follow-ups: #3374 (duplicate-scope fan-out), and making `LibraryBodyIndex` genuinely thread-safe if we ever want that — a wider change than this PR, since the pre-existing evidence accessors already publish caches without barriers.

richlander:
## Review at `ac557157`

Reviewed in a clean detached worktree. The caching change itself is sound, and the aliasing fix from round 1 is correct — the `ImmutableArray` snapshot plus reference-and-order `Matches` closes it, and the comment correctly names the lazy-`Participants()` hazard, which is the subtler half.

I went looking for what two rounds of review hadn't covered. One finding, and four negative results I'm recording so nobody re-runs them.

### Finding: the release API's contract is overclaimed *and* ungated

`ReleaseCallGraphCaches()` says:

> Drops **every** derived call-graph map this index has cached, including the scope graph.

It doesn't. Reflecting over the index's cache fields after exercising the call-graph surface and calling it:

```
--- after ReleaseCallGraphCaches() ---
  POPULATED  _directCallerLoops        <- CallerLoopEvidenceAnalysis.FindNearest(Methods, DirectCalls, 1)
  POPULATED  _rootReachByToken         <- reachability roll-up over the call graph
  POPULATED  _signals                  <- MethodSignalAnalysis.Collect(DirectCalls, ...)
  null       _directCallsByCaller
  null       _distinctCallerEdgesByCallee
  null       _distinctCallersByCallee
  null       _methodMap
  null       _scopeGraph
```

All three survivors are derived from `DirectCalls`/`Methods`. `_rootReachByToken` is literally whole-graph reachability; `_directCallerLoops` is caller-loop evidence over the call graph. Whatever the intended boundary is, "every derived call-graph map" isn't it.

**The part that matters more: neither release method has a gate.** I replaced both bodies with no-ops —

```csharp
public void ReleaseScopeGraph() { /* no-op */ }
public void ReleaseCallGraphCaches() { /* no-op */ }
```

— rebuilt, and ran the whole Analysis suite:

```
ILInspector.Analysis.Tests  Total: 555, Errors: 0, Failed: 0, Skipped: 0
```

**Fully green.** The only test that touches these methods asserts *correctness after release* ("answers after a release must still match a fresh index"), which a no-op satisfies trivially. These methods exist for exactly one reason — the PR body and the doc both say it, "for a consumer under a hard memory ceiling" — and nothing anywhere tests the property they exist for. A future refactor that drops a field from the release, or adds a cache field and forgets it, changes nothing observable.

That combination is the specific shape `AGENTS.md` calls out: *"A doc comment or README that asserts a safety, soundness, or faithfulness property must name the gate that enforces it, or explicitly mark the property as unverified."* And this is the newest, least-reviewed code in the PR — it was added in `ac557157` in response to round 1, so it has had no adversarial pass of its own.

**Suggested fix, in the repo's own derive-don't-restate style.** A test that reflects over `LibraryBodyIndex`'s non-`readonly` `_`-prefixed reference fields, exercises the call-graph surface, calls `ReleaseCallGraphCaches()`, and asserts **set equality** between the still-populated fields and a pinned allow list — so a stale entry and a missing entry both fail, the way `ByteNeutralityGateTests` derives its coverage from the style catalog.

I verified that discriminates before proposing it. Clean, the post-release populated set is `{_allocationFanoutOpportunities, _directCallerLoops, _opportunities, _rootReachByToken, _signals}`; with the methods no-oped, four more appear. Decisive either way.

Then either add the three to the method or narrow the doc's claim to what it actually drops.

### Negative results — hypotheses I tested that did *not* hold

Recording these because each is a plausible attack on this PR and re-running them is wasted effort.

1. **Consumers mutating the now-shared maps.** This was my first and best guess: `ForwardMaps()`/`ReverseMap()` hand out live `Dictionary<string, List<ForwardCalleeEdge>>` and `Dictionary<string, HashSet<string>>`, which were previously rebuilt per call and are now shared. That is the classic poisoning vector. **Both consumers read only** — `rawEdges.Where(...).OrderBy(...)` allocates a new sequence, and there is no in-place `Sort`, `Add`, or `Remove` on anything reachable from the cache. Clean.

2. **Memory retained by the three missed fields.** I expected this to be the story. It isn't: **0.19 MB** on `ILInspector.Analysis.dll` (18.13 MB live after work → 16.33 after release → 16.15 if the three were also dropped). So the finding above is about the contract and the gate, not about bytes. Please don't fix it *for* the memory.

3. **Cross-object `_signals` retention.** `ScopeGraph.ForwardMaps()` calls `assembly.Signals` on every participant, so a cross-assembly query populates `_signals` on all scope indexes — memory that no `Release*` on the *root* can reach, and `4 / 4` scope indexes were still populated after `root.ReleaseScopeGraph()`. Real, but **0.25 MB** for 4 scopes, while `ReleaseScopeGraph()` itself took live from 85.05 → 52.11 MB. It does its job; this is noise beside it.

4. **The product violating its own documented "not safe for concurrent use" contract.** `ProgressiveMemberCallGraph` is precisely the progressive/wasm consumer this PR targets and it does use `Task.Run`, which looked promising. But `RunAsync` walks the tiers sequentially on one thread, and each instance builds its own `_scopedSession`/`_fullSession`/`_crossScopes`, so two members never share a `LibraryBodyIndex`. The single-consumer contract holds, and your decision to document rather than barrier stands.

### Claim verification

| Claim | Mine |
| --- | --- |
| Analysis 555 / **1 failed** (pre-existing) | 555 / **0 failed** — `CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition` passes here |
| CLI 2229 / 1 failed / 10 skipped | 2229 / **0 failed** / 14 skipped |
| Build clean | 0 errors |

Totals match exactly in both suites, so this is environment drift on framework-dependent tests, not a discrepancy in your work — noting it only so the "known pre-existing failure" isn't treated as load-bearing.

I did **not** re-verify the perf table. This box is shared, so a single timing run from me would be weaker evidence than what you already have, not better.

### Minor

The PR body predates `ac557157` and doesn't mention that the change now adds **two new public methods to a public type** (`ReleaseScopeGraph`, `ReleaseCallGraphCaches`). The reconciliation comment covers it, but the body is what most readers will use, and new public surface is the part worth seeing there.

---

Evidence: clean detached worktree at `ac557157`, Release, SDK `11.0.100-preview.7.26366.102`. Probes were file-based apps under `/tmp/` referencing `src/ILInspector.Analysis`; memory figures from forced full collections, reported as deltas from a pre-open baseline. Tamper applied and reverted; `git status --porcelain` empty and 555/0 + 2229/0 re-confirmed before posting. No repository files modified.

richlander:
## Round 3 reconciliation — `ac557157` -> `564b277b`

Accepted the finding in full, in both its parts. The second part is the one that mattered: no-oping both release methods left the suite green, so the property those methods exist for was untested. Fixed in `d71ddd52` and `564b277b`.

### The contract — narrowed the doc, did not broaden the method

Took the "narrow the claim" branch. `_directCallerLoops`, `_rootReachByToken` and `_signals` are evidence-domain caches that serve other producers; `ReleaseCallGraphCaches()` is scoped to the call-tree builders' maps, and the reviewer's own measurement (0.19 MB, "please don't fix it *for* the memory") says nothing is bought by widening it. The doc now names exactly what it drops and exactly what it retains.

### The gate — and the field it caught immediately

`ReleaseMethods_DropExactlyTheCachesTheyDocument` derives the type's mutable cache fields by reflection and pins each one to a side of the release boundary. It fails in both directions and on a newly added field.

**Writing it surfaced a cache neither of us had listed.** The reviewer's dump named eight fields and mine named nine; reflection found `_generatedFrameworkTypes` on the first run. That is the entire argument for deriving the set rather than restating it.

Verified it discriminates before trusting it:

| mutation | result |
| --- | --- |
| both release methods no-oped (the reviewer's exact mutation) | **fails** |
| `ReleaseScopeGraph()` also nulls `_signals` (over-release) | **fails** |
| unmutated | passes |

### The reviewer's own dump caught my first gate being too narrow

Worth recording, because it is the same class of defect the gate exists to prevent. My first filter was `!field.FieldType.IsValueType`, which read as "reference-typed caches". But the reviewer's post-release dump listed `_opportunities` and `_allocationFanoutOpportunities`, which my filter had never seen — they are `ImmutableArray<T>` and signal "not computed" with `IsDefault`, not with `null`.

So the gate meant to catch a forgotten cache field had a blind spot in exactly the shape of a forgotten cache field. `564b277b` filters on **mutable state** instead of on type shape — every non-readonly instance field on `LibraryBodyIndex` is a lazy cache, and all twelve are now pinned — and detects population per field kind. Re-ran the no-op mutation against the widened gate: still killed.

I only noticed because the finding included a raw dump rather than a conclusion. That is the second time this round that a reviewer's *evidence* was worth more than its recommendation.

### Negative results — recorded, not re-run

All four accepted as stated, and none re-run:

1. **Consumers mutating the shared maps** — matches round 2, where Gemini independently grepped every `GetDirectCallsByCaller()` consumer and found read-only use.
2. **Memory in the missed fields (0.19 MB)** — the reason this was fixed as a contract problem and not a memory one.
3. **Cross-object `_signals` retention** — real and out of reach of any `Release*` on the root, but 0.25 MB beside the 33 MB `ReleaseScopeGraph()` returns. Noted as a known bound, not fixed.
4. **`ProgressiveMemberCallGraph` and the single-consumer contract** — this is the strongest of the four, because it is the actual target consumer and the one place the documented contract could have been violated in-tree. Confirming each member builds its own sessions is what makes round 1's "document, don't barrier" decision hold up.

### The Minor

Fixed. The body now has a **New public surface** section with both methods, why they are separate, and the retention numbers for each half.

### Suite drift

Confirmed, and it is drift in the other direction too — `CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition` still fails here and the CLI still reports 10 skipped rather than 14. Totals match exactly (556 here, +1 for the new test), so nothing is load-bearing on it.

### Validation at `564b277b`

- `dotnet build dotnet-inspect.slnx -c Release` — 0 Error(s)
- Analysis: **556 / 1** (the drifting pre-existing failure)
- CLI: **2229 / 1 / 10 skipped**, re-run at `d71ddd52`; `564b277b` changes only an XML doc comment and the Analysis test project

Round 3 is tests and documentation only — no product behavior changed, and the one product edit is a doc comment. I don't think that warrants a fourth adversarial round, but I'd rather say that explicitly than let "Ready to merge" imply the new gate was itself reviewed.

**Ready to merge.**

richlander:
## Re-review at `564b277b` — original finding fixed and gated; one new finding in the gate itself

Reviewed `d71ddd52` and `564b277b` against my earlier comment. The response is a good one, and I verified it rather than reading it:

- The doc no longer claims `ReleaseCallGraphCaches()` "drops **every** derived call-graph map." It now enumerates what it drops, states explicitly what it retains and why, and — per AGENTS.md — **names the gate**: `ReleaseMethods_DropExactlyTheCachesTheyDocument`.
- The gate is **declaration-driven** rather than a restated list: it derives the field set by reflection and asserts set equality against the two classifications, so an unclassified field fails. That is the `ByteNeutralityGateTests` / `SpanAttributionTests` shape AGENTS.md points at.
- `564b277b` catching its own value-typed blind spot (`ImmutableArray` signals "not computed" via `IsDefault`, not `null`) is the right instinct, and documenting that miss in the test comment is better than silently fixing it.

Baseline at `564b277b`, Release, this box: `ILInspector.Analysis.Tests` **556 / 0** (555 before, +1 for the new gate).

### Tampers confirming the gate is not vacuous

| Tamper | Result |
| --- | --- |
| Both release method bodies → no-ops | **FAILS** — `Assert.Equal() Failure: Collections differ at index 0` |
| Add a used, unclassified mutable cache field | **FAILS** — `differ at index 7` |

The first is the exact tamper that left the suite **fully green (555/0)** on my last pass. It now fails. My original finding is properly closed.

(Note for anyone repeating the second: an *unused* field won't compile — `CS0169` is an error here — so the tamper has to be a genuinely used lazy cache or you are testing a stale binary.)

---

## New finding: the *retained* half of the contract is 6/7 vacuous

The test's own doc says:

> Both directions fail: a cache the doc claims to drop but doesn't, **and a cache dropped that the doc says survives.**

The first direction has a non-vacuity guard:

```csharp
foreach (var name in callGraphCaches)
    Assert.Contains(name, before);
```

The second has none — and it needs one, because the comparison is `Assert.Equal(before.Where(...), PopulatedCaches(index))`. A cache that was never populated is absent from `before` **and** absent from after, so the assertion holds no matter what `ReleaseCallGraphCaches()` does to it.

Measured which caches `Exercised(...)` actually populates:

```
_directCallsByCaller           DROPPED   True
_distinctCallerEdgesByCallee   DROPPED   True
_distinctCallersByCallee       DROPPED   True
_methodMap                     DROPPED   True
_scopeGraph                    DROPPED   True
_signals                       retained  True
_allocationFanoutOpportunities retained  False
_directCallerLoops             retained  False
_generatedFrameworkTypes       retained  False
_opportunities                 retained  False
_rootReachByToken              retained  False
_unsafeEvidenceByMember        retained  False
```

**One of seven retained caches is populated.** For the other six the "deliberately retains" half asserts nothing.

Tamper — make `ReleaseCallGraphCaches()` drop three caches the doc explicitly promises to keep:

```csharp
_rootReachByToken = null;
_directCallerLoops = null;
_unsafeEvidenceByMember = null;
```

**The gate passes.** So does the rest of the suite.

The reason this is worth fixing rather than shrugging at: **`_directCallerLoops` and `_rootReachByToken` are two of the three fields that motivated my original finding.** The gate written to close that finding does not cover two thirds of it. It is also the same failure mode as the one `564b277b` just fixed — a filter that quietly excludes part of the set it is meant to pin — one level up.

### Verified fix

Four producer calls populate all twelve caches; the guard then mirrors the existing one:

```diff
         foreach (var name in callGraphCaches)
             Assert.Contains(name, before);
 
+        foreach (var name in retainedCaches)
+            Assert.Contains(name, before);
+
         index.ReleaseCallGraphCaches();
@@
             index.BuildCallerTree(token, new[] { scope }, maxDepth: 2, maxNodes: 50);
+
+            // The retained half of the contract is only gated on caches this workload actually
+            // populates, so touch the evidence-domain producers too.
+            _ = index.OptimizationOpportunities;
+            _ = index.AllocationFanoutOpportunities;
+            _ = index.GetUnsafeEvidenceByMember();
+            _ = index.GeneratedFrameworkTypeNames;
             return index;
```

Verified to discriminate, not merely to pass:

- Untampered, with the fix: gate **passes**; full suite **556 / 0**.
- Same three-cache tamper, with the fix: gate **fails** (`differ at index 1`).

`OptimizationOpportunities` is what pulls in `_directCallerLoops` and `_rootReachByToken`, both of which are private members with no direct test access — which is presumably why they were missed.

---

## Negative results

Checked and clean; reporting so nobody re-runs them:

- **The `!field.IsInitOnly` filter is sound.** It excludes 11 `_`-prefixed readonly fields (`_bodySignals`, `_rawOpportunities`, `_leakTriage`, `_opportunitiesComputed`, …). None is a releasable cache: a readonly field cannot be nulled, so it cannot be on either side of a release contract. The exclusion is correct today. The one way it could rot is a future release method that `.Clear()`s a readonly collection in place rather than nulling it — set equality on populated-ness would not see that.
- **No static caches exist.** `BindingFlags.Static` over `_`-prefixed fields returns an empty set, so instance-only reflection is not hiding anything.
- **The "a newly added cache fails until classified" claim holds** (tamper C above). That limb of the doc is accurate.

---

## Status

The blocking half of my earlier comment is resolved: the release API's contract is now honest and gated. The remaining item is that the gate under-covers its own retained half; the fix is test-only, four lines plus a guard, and I have verified it both passes clean and fails under tamper. I would take it before merging, but it is a gate-completeness issue rather than a product defect — no shipped behavior is wrong.

I did **not** re-run the perf table. This box is shared and single timings here are not evidence.

richlander:
## Round 4 reconciliation — `564b277b` -> `be0841b0`

Finding accepted and fixed. It is correct, and it is the more interesting kind of correct: **the gate written to close your original finding did not cover two thirds of it.** `_directCallerLoops` and `_rootReachByToken` are two of the three fields you named in round 3, and neither was actually gated.

### Verified the finding before taking the fix

I don't apply a suggested diff on the strength of it passing — passing is what a vacuous test does. So I ran both directions of your tamper:

| gate | `ReleaseCallGraphCaches()` also nulls `_rootReachByToken`, `_directCallerLoops`, `_unsafeEvidenceByMember` |
| --- | --- |
| pre-fix (`564b277b`) | **passes** — vacuous, exactly as reported |
| post-fix (`be0841b0`) | **fails**, `Collections differ at index 1` |

Untampered, post-fix: gate passes, Analysis **556 / 1** (the drifting pre-existing failure).

That pair is what makes the fix load-bearing rather than merely green. Your `differ at index 1` prediction matched exactly.

### The fix

Four producer calls in `Exercised(...)` so all twelve caches are populated, and the guard extended to both classifications:

```csharp
foreach (var name in callGraphCaches.Concat(retainedCaches))
    Assert.Contains(name, before);
```

Kept it as one loop over the concatenation rather than two, since the non-vacuity requirement is identical for both halves and splitting them is how the asymmetry arose in the first place.

### The pattern, third time in three rounds

This is the same shape as round 3's `IsValueType` miss, one level up, and I want to name it because it is now a trend rather than an incident:

| round | defect | what it excluded |
| --- | --- | --- |
| 3 | doc claimed more than the method did | three evidence-domain caches |
| 3 (mine) | gate filtered on type shape | two `ImmutableArray` caches |
| 4 (yours) | guard covered one classification | six of seven retained caches |

Each time, a mechanism meant to pin a complete set quietly pinned a subset, and each time the set-equality assertion still passed. The lesson I'm taking is that **a non-vacuity guard is itself part of the contract and needs to be as complete as the classification it guards** — a guard over half the set is not a weaker guard, it is a guard that certifies nothing about the other half.

### Boundary recorded

Took your negative result on `!field.IsInitOnly` and wrote the rot path you identified into the test's doc, rather than leaving it as review-comment-only knowledge:

> Known boundary: this compares populated-ness, so it would not see a future release method that empties a readonly collection in place instead of nulling a field.

No such method exists today and none is proposed; naming the limit is cheaper than re-deriving it later.

### Negative results — accepted, not re-run

- **`!field.IsInitOnly` excludes 11 readonly fields, none releasable.** Agreed on the reasoning: a readonly field cannot be nulled, so it cannot be on either side of a release contract.
- **No static caches.** Accepted.
- **"Newly added cache fails until classified" holds.** This matches my own round-3 result, where the gate caught `_generatedFrameworkTypes` on its first run.
- **`CS0169` is an error here, so an unused tamper field won't compile.** Useful note — and it is the same stale-binary hazard that cost me a diagnostic cycle this round, from the other direction.

### Validation at `be0841b0`

- `dotnet build dotnet-inspect.slnx -c Release` — 0 Error(s)
- Analysis: **556 / 1**
- CLI: unchanged from `d71ddd52`'s **2229 / 1 / 10** — this commit touches only the Analysis test project; the product diff is byte-identical to `564b277b`

Product behavior is unchanged across rounds 3 and 4 — both are test and documentation work on an API whose implementation has not moved since `ac557157`.

**Ready to merge.**

richlander:
## Re-review at `be0841b0` — finding fixed, verified three ways. No open items from me.

`be0841b0` takes the retained-half fix and improves on what I proposed: folding the guard into `callGraphCaches.Concat(retainedCaches)` reads better than two loops, and recording the readonly-collection boundary in the test's doc comment turns a negative result I reported into a documented limit rather than dropping it. That is the right disposition for it.

I re-derived the result rather than trusting the commit message.

Baseline at `be0841b0`, Release, this box: `ILInspector.Analysis.Tests` **556 / 0**.

### Tampers

| Tamper | Expected | Result |
| --- | --- | --- |
| **A** — both release method bodies → no-ops | fail | **FAILS** (`Collections differ at index 2`) |
| **B** — `ReleaseCallGraphCaches()` also nulls `_rootReachByToken`, `_directCallerLoops`, `_unsafeEvidenceByMember` | fail | **FAILS** (`differ at index 1`) |
| **D** — remove the four producer calls, leave the guard | fail | **FAILS** (`Assert.Contains() Failure`) |

Tamper B is the one this commit exists for: it **passed** at `564b277b` and fails now. Tamper A is the original finding, still caught.

Tamper D is the one I care most about structurally, and it is why I would not have accepted the fix without it. The four producer calls in `Exercised` are load-bearing but invisible — nothing about them says "the gate below depends on these." Because the guard asserts every classified cache is populated *before* release, deleting them fails loudly instead of quietly returning the gate to vacuity. **The fix is self-maintaining, not merely correct today.** That is the difference between closing this finding and re-opening it in six months.

---

## A claim I went after, and the measurement error I made doing it

The doc says the retained caches "together retain well under a megabyte." That is a quantitative property with no gate, so I tried to break it, expecting it not to scale — the figure it rests on is the **0.19 MB I reported from a single small assembly** in my first review, and I never scale-checked it before it became a doc claim.

**My first measurement said the claim was badly wrong, and my first measurement was wrong.** I got 13.86 MB on `ILInspector.Decompiler.dll` and **65.23 MB** on `dotnet-inspect.dll` — 343× the claim. It reproduced byte-for-byte across three runs, which is exactly what made it convincing and exactly why reproducibility is not correctness. The probe measured three assemblies in one process, so each iteration's "retained" delta absorbed the previous iteration's index and scope becoming collectable.

Re-measured two independent ways that agree:

| Assembly | Methods | `ReleaseCallGraphCaches()` frees | Retained caches hold |
| --- | ---: | ---: | ---: |
| `ILInspector.Analysis.dll` | 2,226 | 18.34 MB | **0.19 MB** |
| `ILInspector.Decompiler.dll` | 9,135 | 85.66 MB | **0.86 MB** |
| `dotnet-inspect.dll` | 16,988 | 132.84 MB | **1.06 MB** |

Per-field isolation on `dotnet-inspect.dll` independently sums to ~1.05 MB — `_signals` 0.54 MB, `_rootReachByToken` 0.33 MB, `_directCallerLoops` 0.18 MB, and the other four round to zero. Two methods, same answer.

**So the claim holds.** The retention scales sub-linearly and stays around a megabyte across an 8× range in method count. The only nit worth anything: at 16,988 methods it is 1.06 MB, which is marginally *over* a megabyte rather than "well under" it. Say "about a megabyte" if you want the sentence to stay true on the tool's own largest assembly; I would not hold the PR for it.

Reporting the bad number as well as the good one because a 65 MB figure attributed to a reviewer is the kind of thing that gets acted on before anyone re-derives it. If you saw it anywhere, it is wrong.

---

## Status

No open items from me. My original finding (ungated release contract) and my follow-up (retained half 6/7 vacuous) are both closed, each verified by tamper rather than by reading the diff. The doc's numeric claim survives adversarial measurement.

Two things I deliberately did not do: re-run the perf table, since this box is shared and single timings are not evidence; and re-litigate the aliasing and retention findings from the earlier Gemini and GPT reviews, which were already fixed in `ac557157` and which I confirmed at that head.

**Ready to merge** from my side.

richlander:
## Addendum at `be0841b0` — caching correctness probed; verdict unchanged

No movement since my last pass: head is still `be0841b0`, mergeable state `CLEAN`, and no commits or replies since. Rather than repeat a status line, I spent the pass on the one area none of the four review rounds has attacked — **the correctness of the caching itself**, which is what this PR is actually for. Every round so far, mine included, has been about the release API bolted on beside it.

**The staleness gate is real.** Tampered `ScopeGraph.Matches` to `return true`, so a request with a *different* scope set reuses the previously cached graph — the canonical bug this class of cache introduces:

```
LibraryBodyIndexTests.CallTreeBuilders_AreUnaffectedByEarlierRequestsOnTheSameIndex [FAIL]
  Expected: ···"ccurrence.get_ILOffset|Expanded||fanout=0|fanin=12"
  Actual:   ···"ccurrence.get_ILOffset|Expanded||fanout=0|fanin=15"
```

That is the right kind of failure: a real semantic difference in the answer, not a structural assertion about the cache. The test is named for order-independence but it genuinely pins scope-key correctness too.

**The aliasing fix holds up structurally.** `readonly ImmutableArray<LibraryBodyIndex> _scopes = [.. scopes]` defends both halves of the earlier Gemini/GPT finding — a caller mutating the list it passed can no longer make `Matches` compare the list against itself, nor change which assemblies `Participants()` lazily indexes. The comment above it explains exactly that, which is why I checked it rather than assuming.

### One note, not a finding

`Matches`'s first clause, `!ReferenceEquals(root, requestRoot)`, can never be true. `ScopeGraphFor` is private, both call sites are implicit `this.`, and the only construction is `_scopeGraph = new ScopeGraph(this, scopes)` stored on that same instance — so `root`, `requestRoot` and `this` are always the same object. It is harmless and reasonable insurance if `ScopeGraph` ever gets cached somewhere other than the index it is keyed to, but it is a branch no test can ever cover. Worth knowing if branch coverage on this file is ever taken as meaningful; not worth changing.

Verdict unchanged: **ready to merge.**

richlander:
## Round 5 — addendum acknowledged; one amendment at `20ca0f9e`

Thanks for spending the pass on the caching itself. You're right that it was the gap: four rounds had accumulated on the release API bolted on beside the change, and the thing the PR is actually for had gone unattacked since round 1.

The `ScopeGraph.Matches` -> `return true` tamper is the tamper I'd have wanted, and its failure mode is the useful part — `fanin=12` vs `fanin=15` is a real semantic difference in the answer, not a structural assertion about cache internals. That the order-independence test also pins scope-key correctness is a property I'd claimed by construction and hadn't demonstrated.

### On the note — accurate, and I'd frame the conclusion differently

Your analysis is exactly right: `ScopeGraphFor` is private, both call sites are implicit `this.`, the only construction stores onto that same instance, so `root`, `requestRoot` and `this` are always one object and `!ReferenceEquals(root, requestRoot)` can never be true. Uncoverable branch, confirmed.

Where I'd push back is on **"harmless insurance."** The root isn't merely the object holding the cache — it's a graph participant in its own right:

```csharp
IEnumerable<LibraryBodyIndex> Participants()
{
    yield return root;
    foreach (var scope in _scopes)
        yield return scope;
}
```

The cached forward and reverse maps are built over `root` *plus* the scopes, so the root is a genuine component of the cache key. The branch is uncoverable because of **where the graph happens to be stored**, not because the key is over-specified. Dropping the clause wouldn't simplify the key — it would leave it *incomplete*, and silently wrong the moment a graph were cached anywhere but its own root.

So I've kept it and taken your actual point — that a branch-coverage reading of this file would be misleading — by recording both halves at the declaration in `20ca0f9e`:

> The root is part of the key because it is itself a graph participant, not merely the object holding the cache. That clause cannot currently fail — the only storage is `_scopeGraph` on the same instance `ScopeGraphFor` passes as `this` — so it is an uncoverable branch, worth knowing if branch coverage here is ever read as meaningful. It stays because dropping it would leave the key incomplete rather than simplified, and silently wrong if a graph were ever cached anywhere but its own root.

Comment-only; no behavior change. Analysis **556 / 1** (the drifting pre-existing failure), build clean.

### Where that leaves the PR

Five rounds, and the shape of what each found is worth recording:

| round | reviewer | finding |
| --- | --- | --- |
| 1 | Gemini 3.1 Pro + GPT-5.6 | `ScopeGraph` retained the caller's live list — real aliasing bug, found independently by both |
| 2 | both | clean; surfaced pre-existing duplicate-scope fan-out (filed as #3374) |
| 3 | — | release API's contract overclaimed **and ungated** — a no-op left the suite green |
| 4 | — | the gate closing round 3 was 6/7 vacuous on its retained half |
| 5 | — | caching correctness probed directly; uncoverable branch noted |

Rounds 3-5 are all test and documentation work: the implementation hasn't moved since `ac557157`, and the only product edits since are two doc comments.

**Ready to merge.**

richlander:
## Round 5 verified at `20ca0f9e` — ready to merge, and your pushback on my note was right

I'll take the correction first, since it was mine to get wrong.

### `Matches`'s root clause — I withdraw "harmless insurance"

You're right that the framing was the wrong one, and it's checkable rather than a matter of taste:

```csharp
IEnumerable<LibraryBodyIndex> Participants()
{
    yield return root;
    foreach (var scope in _scopes)
        yield return scope;
}
```

`Participants()` has exactly two consumers — the forward builder at `:1140` (`forward` / `incoming` / `definitions`) and the reverse builder at `:1180` — and `root` is the first thing it yields to both. So the root's `Signals` are folded into the cached maps on equal footing with every scope. The root is a component of what was cached, so it belongs in the key describing what was cached.

That makes the branch uncoverable for a reason that has nothing to do with the key being over-specified: it's uncoverable because the only storage site happens to be the root itself. Dropping the clause would make the key *incomplete*, not *simpler* — and incomplete in the silent direction, returning a stale graph rather than throwing. I described a correct clause as redundant; the `<remarks>` block records both halves, which is the right resolution.

### The amendment

Comment-only, confirmed rather than assumed:

```
$ git diff be0841b0 20ca0f9e --numstat
8	0	src/ILInspector.Analysis/LibraryBodyIndex.cs
```

Eight added lines, zero removed, one file, all inside a `<remarks>`. No behavior surface to re-attack, and the gates I tampered at `be0841b0` are unaffected by construction.

### Against the *new* `main`, not the old one

Worth re-checking, since `main` gained #3345, #3348 and #3350 while this PR sat:

| Check | Result |
| --- | --- |
| `git merge-tree origin/main 20ca0f9e` | clean, no conflicts |
| Files touched | `LibraryBodyIndex.cs`, `LibraryBodyIndexTests.cs` — no overlap with the metadata stack |
| Build of the merged tree | 0 errors |
| `ILInspector.Analysis.Tests` on the merged tree | **556 / 0**, three consecutive runs |

### One note on the 556 / 1 you reported

It does not reproduce. Three back-to-back runs of the merged tree are 556 / 0, and this PR's own `test` job passed in CI. So the "drifting pre-existing failure" is either environment-specific to your machine or already gone.

Not a blocker and not this PR's, but "drifting" is the word that would concern me: a test that fails once and passes three times is a flake, and a flake in a 556-test suite is a slow leak of exactly the signal these five rounds were spent building. If you can still catch it, it's worth an issue with the test name — the failure I *can't* see is the only one left here.

### Verdict

**Ready to merge.** No open items from me across five rounds. The table in your comment is the honest summary; I'd only add that rounds 3-5 all landed on the parts asserting the change was correct rather than on the change, which is its own data point about where this repository's risk actually lives.

richlander:
Merged. Taking the last two items.

### The correction — accepted, and thanks for making it checkable

Agreed on all of it, and the `Participants()` consumer count is the part that settles it: the root's `Signals` are folded into both cached directions on equal footing with every scope, so it is a component of *what was cached* and belongs in the key describing it. "Incomplete in the silent direction, returning a stale graph rather than throwing" is the precise statement of the risk.

### The 556 / 1 — it is not a flake, and that turned out to matter

You were right to push on it, but the diagnosis lands elsewhere. `CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition` fails **deterministically** here — 3/3 in isolation, and every full-suite run across this whole session, always the same assertion. "Drifting" was my sloppy word for "you see it pass and I see it fail"; I should have checked before reaching for it.

Since a deterministic failure is worth more than a flake report, I localized it rather than just filing the name — **#3400**:

- **Not candidate discovery.** 122 framework TypeRefs found in `System.Console.dll`.
- **Not assembly resolution.** All five referenced framework assemblies resolve to real files under the shared framework.
- **It is the forwarder hop.** Every one of those assemblies is a facade on this layout — `System.Runtime.dll` has **1 TypeDef and 954 ExportedTypes**, against `System.Private.CoreLib.dll`'s 2957 TypeDefs. The test can only pass by following forwarders out of the facade into CoreLib, and that step fails for all 122 candidates despite `PreferImplementationAssemblies = true`.

The part worth your attention: this raises the question of whether the environments where it *passes* resolve to reference assemblies with real TypeDefs and therefore never exercise the forwarder path at all. If so, the green runs are passing for a reason unrelated to what the test means to assert, and the Windows failure is the honest signal rather than the aberrant one. That is in the issue as the thing to settle first.

Confirmed unrelated to this PR — it fails identically on base `d4002cf1`.

### On your closing observation

That rounds 3-5 all landed on the parts asserting the change was correct, rather than on the change, is the sharpest thing said across the five rounds. The caching had two independent reviewers on it in round 1 and held; the release API bolted on beside it took three rounds to make honest, and one of those rounds was spent fixing the gate written in the previous one. Worth remembering next time I am tempted to treat "and I added a test" as the end of a change rather than the start of a second one.

Issue #3364, opened 2026-07-28
Rework row selection: --row indexes an invisible sequence


richlander:
`--row` is the weakest part of the output-shape surface. It does not select the row a reader counted in the output, and the failure is silent.

## The defect

`ApiCommand.PrintUrlProjectionAsync` builds the selectable sequence by filtering on payload availability, then indexes it positionally:

```csharp
var printableRows = (rows ?? [])
    .Where(row => !string.IsNullOrWhiteSpace(row.Url))   // invisible filter
    .ToList();
...
var targetIndex = options.PrintRow?.Resolve(printableRows.Count) ?? 1;
printableRows = [printableRows[targetIndex - 1]];        // positional
```

If rows 1-2 carry no URL, `--row 1` returns what the table displayed as row 3. The ordinal indexes a sequence the user cannot see.

The code already knows the right answer and discards it: `PrintableDocument` is constructed with `row.Row` -- the original table row -- so the true identity survives into the payload and into JSONL output, while *selection* uses list position. Two numbering systems, and the authoritative one is not the one `--row` uses.

## Separate defect: acquisition failures are dropped silently

```csharp
var content = await fetcher.FetchSourceAsync(rawUrl);
if (content == null)
    continue;
```

With `--print-all` the output silently contains fewer documents than there were rows. With a single `--row`, `documents` goes empty and the user gets `selected section has no printable rows` -- which is false: the section had printable rows and the *fetch* failed. That is an acquisition failure rendered as a success-shaped empty result, which `AGENTS.md` forbids.

This is independently fixable and should not wait for the redesign.

## Design

Today three flags each carry part of three different concerns. Decompose them into orthogonal axes:

| Axis | Default | Override |
| --- | --- | --- |
| count | -- | `-n N` |
| unit | lines | `--rows` |
| direction | head | `--tail` |

```bash
-n 6                     # first 6 lines
-n 6 --tail              # last 6 lines
-n 6 --rows              # first 6 rows
-n 6 --rows --tail       # last 6 rows
--rows 6                 # shorthand for -n 6 --rows
--rows 2..10             # inclusive range -> 9 rows
--rows 2+10              # start plus count -> 10 rows
--rows 10..              # row 10 to end
--row 6 / -r 6           # exactly one row
```

`-r` is unused across the entire CLI today.

### Decisions

[...]

PR #3373, opened 2026-07-28
Make --print failures visible and remove --print-all

richlander:
PR 1 of the #3364 series. Fixes the acquisition-failure half of the issue and removes `--print-all`. Numbering is deliberately unchanged — renumbering `--row` over displayed rows is the next PR in the stack.

## The behavior claim

`--print` reported acquisition failures with a false diagnostic, and `--print-all` dropped them entirely.

Fixture: a compiled assembly with a hand-written SourceLink map, giving `Source Files` two printable rows where row 1 resolves and row 2 points at a URL that 404s.

**Before** (`origin/main` @ `924e6e0c`):

```console
$ ... -S "Source Files" --print --row 2
Error: selected section has no printable rows.        # exit 1
```

False. The section had two printable rows; the *fetch* failed.

```console
$ ... -S "Source Files" --print-all --jsonl
                                                      # exit 0, empty stderr
```

2 printable rows in, **1 document out, exit 0, no diagnostic** — an acquisition failure rendered as success-shaped empty output, which `AGENTS.md` forbids.

**After:**

```console
$ ... -S "Source Files" --print --row 2
Error: failed to fetch the document for printable row 2 from
https://raw.githubusercontent.com/.../NO-SUCH-FILE-3364.md    # exit 1

$ ... -S "Source Files" --print-all
Error: Unrecognized option '--print-all'.                     # exit 1
```

## Why removing `--print-all` is the fix, not a separate change

The silent drop was only representable because `--print-all` made partial success a legal outcome: N rows in, fewer documents out, still exit 0. Removing it makes `--print` strictly exactly-one, so a declared payload that cannot be acquired has no way to degrade into a shorter success — it must be an error. The loop in `PrintUrlProjectionAsync` collapses to a single fetch whose failure is fatal, and `PrintProjectionOutput.Write` selects exactly one document.

That also retires ~15 combinatorial validation branches (`--print` vs `--print-all`, `--print-all` vs `--row`, `--print-all` vs `--bare`) and the multi-document separator rendering path.

## Compatibility

`--print-all` is removed outright, so multi-payload printing is unavailable until the bounded `--rows N..M` grammar lands later in this series. That is deliberate: the issue's design already supersedes blanket fan-out, and an explicit range is a tighter acquisition-consent gesture than "all". The fan-out authorization role documented in `command-transition-model.md` is retired here and returns bounded.

Printable-row numbering is untouched — `--row N` still counts printable rows. The `--row first|last` aliases are untouched.

## Validation

- `dotnet run --project src/dotnet-inspect.Tests -c Release` — **2227 passed, 0 failed**, 14 pre-existing environment skips.
- Before/after captured against a real compiled artifact on both `origin/main` and this head (above).
- `npx markdownlint-cli` clean on all six changed Markdown files.

Test changes: guidance/`--json-array` message assertions updated; the two `--print-all` fan-out tests and the `--print-all`-vs-`--row` conflict test removed with the behavior; `Type_SourceFiles_PrintAllJsonArrayFetchesAllSources` converted to prove `--json-array` still emits a single-element array; `Project_SkillsPrintAll_UsesSeparators` converted to assert `--print-all` is no longer recognized.

No automated test covers the fetch-failure branch: `PrintUrlProjectionAsync` builds its `SourceFetcher` from `HttpClientFactory.SharedUntrustedFetch` with no injection seam, and per the `AGENTS.md` harness boundary I did not add harness-side machinery to work around that. The branch is verified by the real-artifact run above. A test seam is worth considering in the follow-up.

Refs #3364

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,7 +336,7 @@ dotnet-inspect library MyLib.dll --il-offset 0x06000001+0x5
 
 ## Output and querying
 
-Default output is Markdown. Use Markdown for evidence and narrative, `--table` for compact human scanning, `--tsv` for normalized tab-separated rows for agents and scripts, `--jsonl` for one JSON object per table row, and `--json` for structured object graphs. Use `--plaintext` for plain text, `--bare` for one undecorated payload without changing the selected shape, `--value` for one scalar, `--urls` for URL lists, `--paths` for path lists, `--print` to materialize one printable selected-section row (`--row N` chooses a printable row), `--print-all` to materialize every printable row with separators, `--json-array` to emit projected rows as one JSON array, `--rows -n N` to cap rendered table rows, `--count` to reduce a selected section/vector to a single row count, and `--mermaid` on `depends` for diagrams. Verbosity is `-v:q`, `-v:m`, `-v:n`, or `-v:d`. Markdown and JSON can represent multi-section documents; `--table`, `--tsv`, and `--jsonl` render one table/section at a time, so pair them with a specific `-S` selection when querying sectioned output.
+Default output is Markdown. Use Markdown for evidence and narrative, `--table` for compact human scanning, `--tsv` for normalized tab-separated rows for agents and scripts, `--jsonl` for one JSON object per table row, and `--json` for structured object graphs. Use `--plaintext` for plain text, `--bare` for one undecorated payload without changing the selected shape, `--value` for one scalar, `--urls` for URL lists, `--paths` for path lists, `--print` to materialize one printable selected-section row (`--row N` chooses a printable row), `--json-array` to emit projected rows as one JSON array, `--rows -n N` to cap rendered table rows, `--count` to reduce a selected section/vector to a single row count, and `--mermaid` on `depends` for diagrams. Verbosity is `-v:q`, `-v:m`, `-v:n`, or `-v:d`. Markdown and JSON can represent multi-section documents; `--table`, `--tsv`, and `--jsonl` render one table/section at a time, so pair them with a specific `-S` selection when querying sectioned output.
 
 Sections and fields are queryable without a template language:
 
@@ -384,7 +384,7 @@ dotnet-inspect type Command --project ./src/App
 dotnet-inspect member Command --project ./src/App -S "Member Index"
 dotnet-inspect project ./src/App -S Skills
 dotnet-inspect project ./src/App -S Skills --print --row 1
-dotnet-inspect type JsonSerializer --platform System.Text.Json -S "Source Files" --print-all
+dotnet-inspect type JsonSerializer --platform System.Text.Json -S "Source Files" --print --row 1
 dotnet-inspect type string --shape
 dotnet-inspect type --package System.Text.Json --table
 dotnet-inspect member JsonSerializer --package System.Text.Json -m Serialize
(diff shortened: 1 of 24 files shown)


richlander:

Two models reviewed this PR per `AGENTS.md`: **Gemini 3.1 Pro** and **GPT**. Each ran in its own isolated linked worktree at a pinned head. Round 1 reviewed `1b70e24b`; round 2 re-reviewed the fixed head `9f70966d`.

### Findings actioned

**GPT — embedded skills still recommend the removed flag.** `skills/query/SKILL.md` and `skills/sourcelink/SKILL.md` told users to run `--print-all`. These are compiled into the binary, so `dotnet-inspect skill query` would have shipped guidance for an option that now exits 1. This was the one finding in either review that was both new and caused by this PR. Fixed in `9f70966d`; both reviewers re-verified the flag is gone from `--help`, `skill query`, and `skill sourcelink`.

Also in `9f70966d`: narrowed the new acquisition sentence in `docs/design/output-shapes.md` to cover the *selected* row only, so it no longer implies a guarantee about rows the producer never declared.

### Findings verified as pre-existing, and deferred

Each of these reproduces identically on `origin/main` @ `924e6e0c`. I rebuilt a baseline binary from that commit and diffed the actual output rather than reasoning from the source alone.

**Gemini (round 1) — Project Skills drops skills whose file is missing.** Real, but not a print-path defect: the row is absent from the *table* too, so `--print` is self-consistent with what the section declares. The loss happens in the producer, upstream. Filed as #3375.

**Gemini (round 1) — `--bare` implicitly selects row 1 where `--print` errors.** Confirmed and unchanged by this PR. Filed as #3376.

**GPT (round 2) — `--print --json` silently returns type JSON.** Confirmed pre-existing; the early return at `ApiCommand.cs:649` precedes print dispatch at :842, and this PR touches neither. On investigation the scope is wider than reported: `--json` also swallows `--urls`, `--paths`, and `--value`. Only `--count` survives, via an explicit guard. Filed as #3379 with the fuller scope.

I did **not** weaken `docs/design/output-shapes.md` to accommodate #3379. That document states design intent, and the `--json` bypass violates it — the doc is right and the code is wrong. This is the opposite of the #3375 case, where the narrowing reflected a genuine ownership boundary between section producers and the print path rather than papering over a defect.

### Non-actions

Neither reviewer raised a blocking issue against the fixed head. The `--bare` conjunct simplification (`options.Bare && !options.Print && !options.PrintAll` → `options.Bare && !options.Print`) is behavior-preserving because the dropped conjunct was always true for reachable inputs; both models independently audited every `options.Print || options.PrintAll` → `options.Print` rewrite for precedence damage after I disclosed that a regex slip had briefly produced malformed `if` statements during development, and both confirmed no inverted logic survives.

The fetch-failure branch still has no automated test. `PrintUrlProjectionAsync` constructs its `SourceFetcher` from `HttpClientFactory.SharedUntrustedFetch` with no injection seam, and per the harness-boundary rule in `AGENTS.md` I did not add harness-side compensation for a missing product seam. Both reviewers exercised the branch by hand against a compiled SourceLink fixture whose second row resolves to a 404, and both got the intended loud failure and exit 1.

### Fixed-head verification

Both round-2 reviews report clean at `9f70966d`, with 2227 tests passing and 0 failures, and both confirmed the updated tests assert the new behavior rather than being weakened to pass.

richlander:
Ready to merge.

Both fixed-head adversarial reviews are clean at `9f70966d`, CI is green (build-net10, test, markdownlint), and the reconciliation above accounts for every finding from both models.

The three deferred issues (#3375, #3376, #3379) are non-blocking follow-up: each is verified pre-existing on `origin/main` @ `924e6e0c` and untouched by this PR.

PR #3368, opened 2026-07-28
Fold single-use default member-value spill in object initializers (#3336)

richlander:
- Fixes/advances #3336
- Changes `ObjectInitializerPass` to fold an object initializer whose trailing member is a struct-typed `default(T)` spilled after the `newobj`
- Card revision: `5c1d8156`

## Change

> Should we accept this change?

**Conclusion:** **PASS** — a real corpus witness now folds to a single object initializer, the change is byte-neutral (its fixture recompiles exact), and it introduces zero new fidelity diffs versus `origin/main`.

### Benchmark target

Benchmark target: `Azure.Messaging.ServiceBus@7.20.1`

dotnet-inspect command:

```bash
dotnet-inspect member Azure.Messaging.ServiceBus.ServiceBusModelFactory NamespaceProperties -S "Decompiled Source"
```

### Original source

```csharp
public static NamespaceProperties NamespaceProperties(
    string name,
    DateTimeOffset createdTime,
    DateTimeOffset modifiedTime,
    MessagingSku messagingSku,
    int messagingUnits,
    string alias) =>
    new NamespaceProperties
    {
        Name = name,
        CreatedTime = createdTime,
        ModifiedTime = modifiedTime,
        MessagingSku = messagingSku,
        MessagingUnits = messagingUnits,
        Alias = alias,
        NamespaceType = new NamespaceType() // this cannot be created by the user
    };
```

### Before

```csharp
public static Azure.Messaging.ServiceBus.Administration.NamespaceProperties NamespaceProperties(string name, System.DateTimeOffset createdTime, System.DateTimeOffset modifiedTime, Azure.Messaging.ServiceBus.Administration.MessagingSku messagingSku, int messagingUnits, string alias)
{
    NamespaceProperties S_256 = new();
    S_256.Name = name;
    NamespaceProperties S_257 = S_256;
    S_257.CreatedTime = createdTime;
    NamespaceProperties S_258 = S_257;
    S_258.ModifiedTime = modifiedTime;
    NamespaceProperties S_259 = S_258;
    S_259.MessagingSku = messagingSku;
    NamespaceProperties S_260 = S_259;
    S_260.MessagingUnits = messagingUnits;
    NamespaceProperties S_261 = S_260;
    S_261.Alias = alias;
    NamespaceProperties S_262 = S_261;
    NamespaceType V_0 = default;
    S_262.NamespaceType = V_0;
    return S_262;
}
```

- Valid: True
- Correct: True
[...]

PR #3377, opened 2026-07-28
Wrap long object/collection initializers Allman-style (#3371)

### LINKED ISSUE #3371: Decompiler: length-based pretty printing for object/collection initializers (long lines)
## Summary

The decompiler prints object/collection initializers **unconditionally on a single physical line**, so a many-member initializer becomes a very long line. #3336/#3368 folded `ServiceBusModelFactory.NamespaceProperties` into a correct object initializer, but the result is one 499-char line (the initializer alone is 212 chars):

```csharp
public static ...NamespaceProperties NamespaceProperties(string name, ...) => new NamespaceProperties { Name = name, CreatedTime = createdTime, ModifiedTime = modifiedTime, MessagingSku = messagingSku, MessagingUnits = messagingUnits, Alias = alias, NamespaceType = default(NamespaceType) };
```

That single-line form is **not** what our style oracle (dotnet/runtime) produces for a multi-member initializer, and it is well past the printer's existing wrap width (120). This issue proposes **length-based pretty printing** for initializers: keep short ones inline, break long ones one member per line.

## Oracle: what dotnet/runtime does

dotnet/runtime coding style is "use Visual Studio defaults" with **Allman braces** ([coding-style.md](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md), rule 1). In real source, short initializers stay inline but multi-member initializers break Allman-style — brace on its own line, one member per line:

- Inline (short): `new JsonSerializerOptions { WriteIndented = true };` — `src/tasks/Common/FileCache.cs`
- Multi-line (several members):

  ```csharp
  // src/mono/wasm/host/RunArgumentsJson.cs
  new JsonSerializerOptions
  {
      DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingDefault,
      PropertyNamingPolicy = JsonNamingPolicy.CamelCase
  }
  ```

  (same shape in `src/tasks/MonoTargetsTasks/RuntimeConfigParser/RuntimeConfigParser.cs`, `src/coreclr/tools/cdac-build-tool/DataDescriptorModel.cs`, `src/mono/wasm/host/CommonConfiguration.cs`, …)

The `NamespaceProperties` **Original Source** (via SourceLink) is itself multi-line, one member per line — so our raised render diverges in spelling from the very source it reconstructs.

## The gap in the printer

Object/collection initializers are emitted with a flat join and no width awareness:

- `CSharpPrinter.RaisedExpressions.cs` — `InitializerBodyText(...)` → `"{ " + string.Join(", ", entries...) + " }"`; `ObjectInitializerText`, `WithExpressionText`, and the anonymous-object `new { ... }` path all do the same.

Meanwhile the printer **already has an always-on, width-based wrapper** for other constructs, which initializers simply bypass:

- `CSharpPrinter.Members.cs:818` — `const int FluentChainWrapWidth = 120;` (+ `FluentChainMinSegments = 2`)
- `FluentChainLines` / `TryAppendFluentChain` break a long fluent call chain one call per line when `indent*4 + prefix + text + suffix > 120`
- gated off by the existing whitespace-only `disable-one-liner-wrapping` option (`PrinterOptions.DisableOneLinerWrapping`)

## Proposal

Extend the existing always-on width wrapper to object/collection initializers (and `with` / anonymous-object initializers):

1. When the single-line initializer form would exceed `FluentChainWrapWidth` (measured with the member/statement prefix, suffix, and current indent, exactly as the fluent-chain wrapper does), emit the Allman multi-line form:

   ```csharp
   new NamespaceProperties
   {
       Name = name,
       CreatedTime = createdTime,
       ...
       NamespaceType = default(NamespaceType)
   };
   ```

   one member per line, closing brace aligned to the `new`. (Decide trailing-comma policy; dotnet/runtime uses both, VS default omits it — pick one and pin it.)
2. Keep short initializers inline (below the width, or a single member) — no change to the common case.
3. Honor `DisableOneLinerWrapping` so the whitespace-only escape hatch still forces the flat form.
4. Nested initializers (`Inner = { ... }`) and collection/dictionary element forms should wrap consistently.

## Notes

- **Byte-neutral / whitespace-only** — this is layout, not opcodes, so it belongs in the always-on width-wrapper family alongside the fluent-chain wrapper, **not** a byte-divergent taste lens. IL fidelity is unaffected.
- Consider unifying the wrap-width constant so fluent chains, long member signatures, and initializers share one threshold and one `disable-one-liner-wrapping` gate.
- Add printer tests for: below-threshold inline, above-threshold multi-line, nested initializer, `with` expression, anonymous object, and `DisableOneLinerWrapping` forcing flat.

Follow-up to #3368 (which produced the first long-line witness). Witness command:

```bash
dnx dotnet-inspect -y -- member Azure.Messaging.ServiceBus.ServiceBusModelFactory NamespaceProperties --all \
  --library <cache>/azure.messaging.servicebus/7.20.1/lib/net8.0/Azure.Messaging.ServiceBus.dll -S "Decompiled Source"
```

richlander:
Closes #3371.

## Summary

The C# printer rendered every raised object/collection initializer on one
physical line (`"{ " + string.Join(", ", …) + " }"`), so a many-member
initializer became a very long line. The merged #3368 witness
`ServiceBusModelFactory.NamespaceProperties` rendered as a **499-char** line
(initializer body alone 212 chars) — well past the printer's existing
120-column wrap width that already breaks long fluent chains, splittable
boolean chains, and bitwise chains one element per line.

This extends that always-on width wrapper with an initializer arm: when an
initializer-valued statement (`return`, non-ref local, or stack-slot store)
whose flat single-line form exceeds `FluentChainWrapWidth` (120), the printer
breaks it **Allman-style** — the `new T(...)` head on the first line, `{`/`}`
on their own lines at the statement indent, one `Member = value` / element
entry per line one continuation level deeper. Short initializers stay inline;
`DisableOneLinerWrapping` suppresses the wrap (issue #3185 parity).

This matches the dotnet/runtime style oracle
(`csharp_new_line_before_members_in_object_initializers=true`, Allman braces) —
the `NamespaceProperties` **Original Source** below is itself broken exactly
this way.

## Byte-neutrality

Whitespace-only. The wrapped form **reuses the same head and entry texts** the
inline renderer emits and carries **no trailing comma**, so it is a pure
whitespace variant of the inline form — token-identical, same IL. A defensive
re-match (`$"{head} {{ {join} }}" != ObjectInitializerText(init) → stay inline`)
plus the caller's `line != prefix + ObjectInitializerText(init) + ";"` guard
keep any coercion cast or renderer quirk inline rather than reshaping a token —
the same guard pattern the fluent/splittable/bitwise wrappers use.

- **IL fidelity:** True — proven by the compile-back RoundTrip area
  (`Area=RoundTrip`, incl. slow compile-back gates): all wide initializers
  recompile to identical opcodes. (The 3 `parameter-attributes`/fidelity
  `RecompileFail`s are pre-existing on base `e3040716`, reproduced there,
  unrelated to this change.)
- **Taste applied:** none — this is always-on formatting (like fluent-chain
  wrapping), not a configurable style lens; `-S "Applied Taste"` reports "No
  recorded style choices were applied" on both base and head.

## Original source

`member Azure.Messaging.ServiceBus.ServiceBusModelFactory NamespaceProperties -S "Original Source"`

```csharp
public static NamespaceProperties NamespaceProperties(
    string name,
    DateTimeOffset createdTime,
    DateTimeOffset modifiedTime,
    MessagingSku messagingSku,
    int messagingUnits,
    string alias) =>
    new NamespaceProperties
    {
        Name = name,
        CreatedTime = createdTime,
        ModifiedTime = modifiedTime,
        MessagingSku = messagingSku,
        MessagingUnits = messagingUnits,
        Alias = alias,
        NamespaceType = new NamespaceType() // this cannot be created by the user
    };
```

## Before (base `e3040716`, `-S "Decompiled Source"`)

- Valid: True — Correct: True — IL fidelity: True — Taste applied: none

```csharp
public static Azure.Messaging.ServiceBus.Administration.NamespaceProperties NamespaceProperties(string name, System.DateTimeOffset createdTime, System.DateTimeOffset modifiedTime, Azure.Messaging.ServiceBus.Administration.MessagingSku messagingSku, int messagingUnits, string alias) => new NamespaceProperties { Name = name, CreatedTime = createdTime, ModifiedTime = modifiedTime, MessagingSku = messagingSku, MessagingUnits = messagingUnits, Alias = alias, NamespaceType = default(NamespaceType) };
```

## After (head `a82bd9aa`, `-S "Decompiled Source"`)

- Valid: True — Correct: True — IL fidelity: True — Taste applied: none

```csharp
public static Azure.Messaging.ServiceBus.Administration.NamespaceProperties NamespaceProperties(string name, System.DateTimeOffset createdTime, System.DateTimeOffset modifiedTime, Azure.Messaging.ServiceBus.Administration.MessagingSku messagingSku, int messagingUnits, string alias) => new NamespaceProperties
{
    Name = name,
    CreatedTime = createdTime,
    ModifiedTime = modifiedTime,
    MessagingSku = messagingSku,
    MessagingUnits = messagingUnits,
    Alias = alias,
    NamespaceType = default(NamespaceType)
};
```

## Scope (v1)

- Handles `ObjectInitializerExpression` (object and collection form). Requires
  `>= 2` entries (parity with `FluentChainMinSegments`).
- **Deferred:** parameter-list wrapping (the signature stays flat — a separate
  concern), `WithExpression`/anonymous-object initializers (separate
  renderers), and recursive wrapping of nested initializers (a nested member
  value stays inline within its one line).

## Tests

- New `ObjectInitializerFormattingTests` (5): long initializer breaks Allman
  one-per-line with no trailing comma; short stays inline; a single long entry
  stays inline; `DisableOneLinerWrapping` stays inline; the broken form is a
  whitespace-collapse-identical variant of the inline form.
- Fast decompiler suite `-trait- "Speed=Slow"`: **4051/0**.
- Slow `Area=RoundTrip` compile-back: green (pre-existing env `RecompileFail`s
  excepted).
- CLI product-output suite (`src/dotnet-inspect.Tests`): **2229/0**.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ObjectInitializerFormattingTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ObjectInitializerFormattingTests.cs
@@ -1,0 +1,116 @@
+using System.Collections.Immutable;
+using ILInspector.Decompiler.Pipeline;
+
+namespace ILInspector.Decompiler.Tests;
+
+// #3371: once a raised object initializer is long, CSharpPrinter breaks it
+// Allman-style — the `new T` head on the first line, `{`/`}` on their own lines,
+// and one `Member = value` entry per line under a continuation indent; an
+// initializer that still fits stays inline. Breaking is whitespace-only (no
+// trailing comma, same head and entry texts), so the broken body still compiles
+// and carries the same tokens as the inline form.
+public sealed class ObjectInitializerFormattingTests
+{
+    static readonly TypeRef Holder = TypeRef.Definition("synthetic", "", "Holder");
+    static readonly TypeRef Options = TypeRef.Definition("synthetic", "", "WideOptions");
+    static readonly TypeRef Int32 = TypeRef.CoreLib("System", "Int32");
+    static readonly TypeRef Void = TypeRef.CoreLib("System", "Void");
+
+    static readonly MethodRef Ctor = new(Options, ".ctor", Void, [], HasThis: true);
+
+    static ObjectInitializerExpression Initializer(params string[] members)
+    {
+        var creation = new NewObject(Ctor, []);
+        var entries = members.Select(m => new InitializerEntry(m, [new LoadArgument(0, ArgName(m), Int32)]));
+        return new ObjectInitializerExpression(creation, isCollection: false, entries);
+    }
+
+    static string ArgName(string member) => "value_" + member;
+
+    // Render `return <initializer>;` as the whole method body and reduce it to
+    // the raw statement/expression-body text.
+    static string Render(ObjectInitializerExpression init, PrinterOptions? options = null)
+    {
+        var block = new Block(0);
+        block.Add(new Return(init));
+        var container = new BlockContainer();
+        container.Add(block);
+        var signature = new MethodSignature(Options, ImmutableArray<Parameter>.Empty, HasThis: false, GenericParameterCount: 0);
+        var function = new IrFunction("M", Holder, signature, [], container);
+        return CSharpPrinter.Print(function, options ?? PrinterOptions.Default).Output!;
+    }
+
+    // Four PascalCase members whose flat `new WideOptions { ... }` form runs past
+    // the 120-column wrap width.
+    static ObjectInitializerExpression LongInitializer()
+        => Initializer(
+            "FirstMeasuredProperty",
+            "SecondMeasuredProperty",
+            "ThirdMeasuredProperty",
+            "FourthMeasuredProperty");
+
+    [Fact]
+    public void LongInitializer_BreaksAllmanOnePerLine()
+    {
+        string body = Render(LongInitializer());
+
+        // Head, then `{`/`}` on their own lines, one entry per line, no trailing comma.
+        Assert.Contains("new WideOptions\n", body);
+        Assert.Contains("\n{\n", body);
+        Assert.Contains("    FirstMeasuredProperty = value_FirstMeasuredProperty,\n", body);
+        Assert.Contains("    FourthMeasuredProperty = value_FourthMeasuredProperty\n", body);
+        Assert.DoesNotContain("value_FourthMeasuredProperty,", body);
+        Assert.Contains("\n};", body);
+        // No flat `{ ... }` body survives.
+        Assert.DoesNotContain("{ FirstMeasuredProperty", body);
+    }
+
+    [Fact]
+    public void ShortInitializer_StaysInline()
+    {
+        string body = Render(Initializer("X", "Y"));
+
+        Assert.Contains("new WideOptions { X = value_X, Y = value_Y };", body);
+        Assert.DoesNotContain("\n{\n", body);
+    }
+
+    [Fact]
+    public void SingleEntry_StaysInline_EvenWhenLong()
+    {
+        string body = Render(Initializer("AnExtremelyLongSinglePropertyNameThatOnItsOwnExceedsTheOneHundredAndTwentyColumnWrapWidthEasilyYes"));
+
+        Assert.DoesNotContain("\n{\n", body);
+    }
+
+    // Issue #3185 parity: the general one-liner opt-out suppresses the always-on
+    // wrapper, so an over-width initializer stays on one line.
+    [Fact]
+    public void LongInitializer_DisableOneLinerWrapping_StaysInline()
+    {
+        string body = Render(LongInitializer(), PrinterOptions.Default with { DisableOneLinerWrapping = true });
+
+        Assert.Contains(
+            "new WideOptions { FirstMeasuredProperty = value_FirstMeasuredProperty,"
+                + " SecondMeasuredProperty = value_SecondMeasuredProperty,"
+                + " ThirdMeasuredProperty = value_ThirdMeasuredProperty,"
+                + " FourthMeasuredProperty = value_FourthMeasuredProperty };",
+            string.Concat(body.Split('\n').Select(line => line.Trim())));
+        Assert.DoesNotContain("\n{\n", body);
+    }
+
+    // The broken form must be a pure whitespace variant of the inline form: the
+    // same tokens, in the same order, differing only in whitespace.
+    [Fact]
+    public void BrokenForm_IsWhitespaceVariantOfInline()
+    {
+        string broken = Render(LongInitializer());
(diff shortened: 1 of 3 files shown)


richlander:

Two independent adversarial reviews of the **fixed head `a82bd9aa`**, each in an isolated linked worktree (scratch under `/tmp`, no writes/commits in-tree):

| --- | --- |
| GPT (gpt-5.6-sol) | **CLEAR** |
| Gemini (gemini-3.1-pro) | **CLEAR** |

Both attacked the same points and found no reachable bug, security issue, or byte-neutrality violation. Confirmed:

- **Token/byte neutrality** — the reconstructed flat text exactly mirrors `ObjectInitializerText`; the missing trailing comma matches `string.Join`; any divergence trips the reconstruction guard and stays inline.
- **Expression-body folding** — the multi-line shape folds and re-indents correctly for `return`/`StoreLocal`/`StoreStackSlot` (braces aligned to the member, entries +4).
- **Wrapper ordering** — fluent/splittable/bitwise arms match `Call`/`LogicalBinary`/`Binary`, never `ObjectInitializerExpression`; no double-wrap.
- **Guard soundness** — an added cast/coercion changes the flat statement text, so `line != prefix + ObjectInitializerText(init) + ";"` keeps it inline; declaring-store prefixes computed via `DeclarationTypeText`.
- **Width math** — `indent*4 + prefix + flat + suffix` is equivalent to `indent*4 + line.Length`; no off-by-one; mirrors `FluentChainLines`.
- **Nested initializers** (`InitializerBlock`) and **collection initializers** (bare elements + `{ k, v }`) stay token-identical.

Independent witnesses (collection, local-store, dictionary, nested-initializer) rendered valid C#.

### Evidence observed at `a82bd9aa`
- Build: 0 errors.
- `ObjectInitializerFormattingTests`: 5/5.
- Fast decompiler suite (`-trait- "Speed=Slow"`): 4051/0.
- Slow `Area=RoundTrip` compile-back: the only failures are the known pre-existing toolchain-divergence / `parameter-attributes Update#0 RecompileFail` gates, reproduced on base `e3040716` and unattributable to this whitespace-only change.

No findings required action.

richlander:
Ready to merge

PR #3380, opened 2026-07-28
Contain untrusted metadata names in emitted C# text (#3319)

### LINKED ISSUE #3319: Unescaped parameter names let metadata break out of Decompiled Source and Annotated Source
## Summary

`Decompiled Source` and `Annotated Source` render **parameter names** straight
from metadata. A parameter name containing a C# line terminator therefore ends
the construct that frames it and leaves the remainder as **active C# code** in
a fenced C# block.

This is the same vulnerability class as #3257, but a different producer in a
different layer. #3257 covers the IL text producer (fixed in #3315); this issue
covers the C# printer's identifier spelling.

Note this **contradicts #3257's blast-radius table**, which recorded
`Decompiled Source` as unaffected on the grounds that "the C# printer already
sanitizes identifiers". That is true for the field name in #3257's repro, but
not for parameter names.

## Reproduction

Emit an assembly whose parameter name carries a newline (the CLR does not
require metadata names to be C#-spellable):

```csharp
var m = tb.DefineMethod("Echo", MethodAttributes.Public | MethodAttributes.Static,
    typeof(int), new[] { typeof(int) });
m.DefineParameter(1, ParameterAttributes.None, "p\n    public int Injected() => 42; //");
```

```console
$ dotnet-inspect member HostileParam.dll Hostile.Target.Echo -S "Decompiled Source"
```

```csharp
public static int Echo(int p
    public int Injected() => 42; //) => p
    public int Injected() => 42; //;
```

Both the signature and the `return` expression escape. `Annotated Source` shows
the same two escapes (its IL comment channel is folded by #3315, so what remains
there is purely the printer's spelling).

Reproduced at `origin/main` and on the head of #3315.

## Root cause

The printer already has the right tool and applies it inconsistently:

| Name kind | Spelling used | Sanitizes unspellable chars |
| --- | --- | --- |
| Field | `CSharpNaming.SafeIdentifier` | Yes |
| Method / call target | `CSharpNaming.SourceMethodName` (→ `SafeIdentifier`) | Yes |
| **Parameter, local, property, generic parameter, local function** | `CSharpNaming.EscapeIdentifier` | **No** — `@`-escapes reserved keywords only |

`EscapeIdentifier` makes `delegate` into `@delegate`; it does nothing for a name
containing a newline, `<`, or a space.

## Suggested fix

Route the remaining identifier kinds through `CSharpNaming.SafeIdentifier`
(i.e. `CSharpIdentifier.Sanitize`) as fields and methods already are. Per
`CSharpNaming.SourceMethodName`'s own doc comment, `SafeIdentifier` is identical
to `EscapeIdentifier` for names that are already valid identifiers, so the change
should be byte-neutral for well-formed assemblies.

There are roughly 30 call sites across `CSharpPrinter.cs`,
`CSharpPrinter.Members.cs`, `CSharpPrinter.RaisedExpressions.cs`, and
`TypeRef.cs`, so this wants its own PR and its own regression gates rather than
being folded into #3315.

## Scope note

Deliberately **not** fixed in #3315: that PR owns the IL text layer
(`IlProjection` / `InstructionProducer`), while this defect is in the CSharp
layer that owns C# spelling, and it affects `Decompiled Source`, which #3257
placed out of scope. #3315's description records this as a known remaining
channel rather than claiming `Annotated Source` is fully closed.

See `docs/design/untrusted-data-threat-model.md`.

richlander:
Fixes #3319.

## Claim

An untrusted metadata name can no longer add a line to emitted C# text. A parameter, property, generic-parameter, or method name carrying a line terminator previously escaped its Markdown code fence and injected text that reads as decompiler output.

## Why it had so many mouths

Four separate reimplementations of the same keyword-only escaper existed, all spelled `RequiresDeclarationEscape(name) ? "@" + name : name`:

| Site | Assembly |
| --- | --- |
| `CSharpDeclarationWriter.EscapeIdentifier` | `ILInspector.CSharp` |
| `MetadataDeclarationQuery.EscapeIdentifier` | `ILInspector.Metadata` |
| `ApiSurfaceExtractor.EscapeIdentifier` | `ILInspector.Metadata` |
| `CSharpFormatter.EscapeIdentifier` | `ILInspector.CSharp` |

They now share one primitive, `CSharpIdentifierCore`. It is `internal` and **source-linked** rather than referenced, because the project reference runs `ILInspector.CSharp` → `ILInspector.Metadata` and cannot be reversed; this follows the existing `CSharpKeywords.cs` link precedent.

The two keyword sets genuinely differ (`RequiresBodyEscape` adds `await`; `RequiresDeclarationEscape` adds `record`/`required`/`init`/`file`/`scoped`), so the primitive is parameterized over the predicate rather than unified onto one set. Pointing a declaration site at the body spelling would have silently narrowed escaping and regressed a parameter named `record`.

`AddMethodGenericParameters` was splicing method generic-parameter names in with **no escaping whatsoever**. That was the widest hole.

## The fix is deliberately narrow

Three distinct properties now exist and are documented as distinct:

| Function | Property |
| --- | --- |
| `Sanitize` | always yields a legal C# identifier. **Semantics unchanged**; existing callers untouched. |
| `ContainIdentifier` | keyword-escape, plus sanitize **only if the name carries a line terminator**. The #3319 fix. |
| `ContainComposedName` | fold line terminators only, for text that is not an identifier at all (`.ctor`, `IFoo.Bar`, `Name<T,U>`). |

My first attempt sanitized every untrusted name. `KeywordIdentifierTests` and `UnspeakableNameFidelityTests` gate the opposite contract: an unspellable-but-harmless name (`bad-name`, `<>c__DisplayClass0_0`) is **preserved raw** so identity stays visible, and the honest-failure signal is `DecompilationFidelity.Partial`, not a rewritten name. Narrowing to line terminators closes the injection while preserving that contract — and is what makes the change byte-neutral, since no name a compiler can emit contains a line terminator.

`ApiSignature.MemberName` is **identity, not display** (`ApiMemberIdentity` parses it for docids and the generic-parameter map), so only `displayName` is contained there.

## Evidence

Full suite, Release, post-merge with `origin/main` — **8,498 tests, 0 failures**:

| Suite | Result |
| --- | --- |
| `ILInspector.CSharp.Tests` | 394, 0 failed |
| `ILInspector.Metadata.Tests` | 947, 0 failed |
| `DotnetInspector.Services.Tests` | 319, 0 failed |
| `ILInspector.Analysis.Tests` | 554, 0 failed |
| `dotnet-inspect.Tests` | 2229, 0 failed |
| `ILInspector.Decompiler.Tests` (`-trait- "Speed=Slow"`) | 4055, 0 failed |

**Byte-neutrality**, measured rather than asserted, `origin/main` @ `e3040716` vs this head, same commands, `cmp`:

```
IDENTICAL: type JsonSerializer --platform System.Text.Json --shape -v:d
IDENTICAL: type Utf8JsonReader --platform System.Text.Json --shape -v:d
IDENTICAL: type List`1 --platform System.Private.CoreLib --shape -v:d
IDENTICAL: type String --platform System.Private.CoreLib --shape -v:d
IDENTICAL: type Dictionary`2 --platform System.Private.CoreLib --shape -v:d
IDENTICAL (23 lines): member System.Text.Json.JsonSerializer -m Serialize --platform System.Text.Json -v:d
IDENTICAL (21 lines): member System.String -m Format --platform System.Private.CoreLib -v:d
IDENTICAL (10 lines): member System.Linq.Enumerable -m Where --platform System.Linq -v:d
IDENTICAL (117 lines): member System.Collections.Generic.List`1 --platform System.Private.CoreLib -v:d
```

### Gates name their enforcement

- `CSharpIdentifierSanitizationTests.ContainIdentifier_IsByteNeutral_AcrossRealAssemblyMetadata` drives byte-neutrality from real `System.Private.CoreLib` metadata (>10,000 names) with a non-vacuity floor, so the neutrality claim is enforced, not restated.
- `ContainIdentifier_PreservesUnspellableNamesThatCannotBreakOutput` fails if the containment is ever widened back into blanket sanitization — it pins the identity contract that scoped this fix.
- `UntrustedIdentifierPresentationTests` is end-to-end over a `PersistedAssemblyBuilder` assembly with hostile parameter, property, and generic-parameter names, across `\n`, `\r\n`, and `\u2028`. It renders the member **twice** — benign name and hostile name — and asserts equal line counts, using the benign rendering as the oracle. The differential is the point: the two output channels contain differently (the IL comment channel folds and the payload survives verbatim; the C# channel sanitizes), so any single spelling witness would have been wrong for one of them.

### Tamper-tested, both directions

| Tamper | Result |
| --- | --- |
| `ContainsLineTerminator` → always `false` (containment off) | 21 unit + 9 end-to-end failures |
| `ContainsLineTerminator` → always `true` (widened to blanket sanitization) | 8 failures, including the real-metadata byte-neutrality gate and all four identity-preservation cases |

## Boundary — not fixed here

The `type` **tree** renders raw metadata names for field and property summary rows (`ApiOutputFormatter.cs`, `new MemberRow(...)` / `new FieldSummaryRow(...)` pass `m.Name` straight through). Those still break the ASCII gutter:

```
├─ Fields (1)
│  └─ Fld
    public int Injected() => 42; //
├─ Properties (1)
│  └─ int Prop
    public int Injected() => 42; // { get; }
└─ Methods (4)
   ├─ int Echo(int p_____public_int_Injected______42____)   <- fixed here
   ├─ void Generic<T_____public_int_Injected______42____>() <- fixed here
   └─ void Meth     public int Injected() => 42; //()       <- fixed here (folded)
```

No C# spelling producer is involved in the `Fld`/`Prop` rows: that is CLI presentation over the external Markout renderer — a different owner and a different blast radius, and the table renderings of the same rows already fold and HTML-escape correctly. Filed separately rather than smuggled in here.

## Review

Two-model tier per AGENTS.md (security fix on a default output path).


richlander:
## Ready to merge — reconciliation of 29 rounds of adversarial review

This comment is the public reconciliation AGENTS.md requires. It attributes every finding,
records what was verified or dismissed and on what evidence, and declares the residuals.

### What the change does

Untrusted metadata names — type, member, namespace, parameter, and file names read out of an
inspected assembly — reach emitted C# text and CLI output. An attacker who controls a `.dll`
controls those strings, and they were being written into identifiers, signatures, doc text,
Markdown rows, and diagnostics without containment. This PR contains them at the point of
emission and installs gates that keep them contained.

Two things follow from that, and the second is where the review went:

1. **Containment** — `CSharpIdentifierCore` escapes and verifies identifier spelling; the
   rendering layer contains names in signatures, summary rows, doc comments, layout trees,
   diff channels, SourceLink and search output, and package identity.
2. **Ownership of stderr** — containment is only worth something if every diagnostic route
   goes through one contained writer. `CommandError` became the CLI's sole stderr owner, and
   the hard problem was proving nothing else in the shipped closure writes to fd 2.

### The architectural correction (the part worth reading)

Rounds 20–27 were spent proving point 2 with a hand-written gate that scanned source for
`Console.Error` and friends. That gate reached **1,761 lines**. Measured per commit, excluding
merges, those eight rounds changed **61 lines of product code and 3,532 lines of test code**.

That ratio was the defect. The gate was a second implementation of C# name resolution — it had
to decide what `Console` meant at each site, which is the compiler's job — and it was being used
as a security boundary. `docs/design/inspection-layers.md:179` already names this: *"A second
implementation of a shared rule is a defect… fix the seam, do not re-derive the rule."*

The gate was replaced with the C# compiler. `BannedApiAnalyzers` + `eng/BannedSymbols.txt`
declares the four banned symbols; `Directory.Build.targets` applies the analyzer and escalates
`RS0030` to an error across every project that does not opt out via `OwnsItsOwnStderr`. The
remaining test file stopped resolving names and became four checks that verify the *wiring* is
present and load-bearing:

| | Before | After |
|---|---|---|
| Gate source | 1,761 lines | **710** at the rewrite, **979** after review closed measured gaps |
| Tests | 14 | **4** |
| Runtime | ~43s | **~7s** |
| Decides what `Console` means | yes | no — the compiler does |

**Backstop independence is what makes this survive suppression.** Five analyzer-suppression
escapes were attempted — `NoWarn`, `WarningsNotAsErrors`, `.editorconfig severity=none`,
`GlobalSuppressions`, `RunAnalyzers=false`. Each still failed the gate, because the IL checks read
the *shipped assembly* and do not care whether the analyzer ran.

**The same lesson recurred twice inside the fix.** I nearly re-derived MSBuild's item model to
catch `Output ItemName`, and nearly wrote a second signature formatter. Both times the answer was
to ask the producer — the build's own `.deps.json`, and the product's `SignatureDecoder` — rather
than enumerate spellings. That generalizes beyond this PR; I'll file it separately.

### A second lesson, from the review itself

One `<remarks>` paragraph — the one naming the gate that covers what the IL count cannot — went
through **four revisions, and adversarial review found a distinct error in every one**: too narrow
a claim, then a route attributed to four classes that do not drive it, then an oracle attributed
to a class that uses none of it.

They have one cause. Each revision named another test class and asserted a property of it
*without opening the file*. That is the same defect as the architectural one, in miniature: a doc
comment naming a gate is an assertion about another file, and it decays exactly like a claim about
behaviour — except nothing compiles it, so only a reviewer who opens the file can falsify it.

Naming the wrong gate is worse than naming none: a reader who checks "Mermaid trace, gated by
`UntrustedLibraryViewContainmentTests`" opens four files, greps, finds nothing, and concludes the
route is unguarded — when it is guarded, by the class the sentence pointed away from.

Each of the four defects was falsifiable by exactly one action — open the named file, run one
`grep` — which is why measuring first eliminated them. So the rule is not only "name gates
sparingly", though that reduces how many such assertions you carry. It is that **a doc comment
naming another file is an untested assertion about that file, and belongs at the same evidence bar
as a claim about behaviour**: cite it only after running the check that would falsify it.

### Reconciliation of findings

**Claude Opus 4.8 — 8 passes, 0 blocking findings, every claim reproduced.**

| Finding | Disposition |
|---|---|
| `RS0030` was an error only under `src/` (`TreatWarningsAsErrors` lives in `src/Directory.Build.props`) | **Real, fixed** in `55bc69310`. Escalation moved to `Directory.Build.targets` under the same condition as the analyzer. This exposed **3 genuine unenforced violations** in `MetadataImageOverviewRendererTests`. |
| Closure completeness was unpinned | **Real, fixed** — `BuildFilesContributingProjectReferences()` added. |
| Residual paragraph said "three ways" when there are four checks, and misattributed the build-file scan's coverage to project XML | **Real, fixed** in `355ceedc6`. Doc accuracy, not coverage — the declared residual was correctly narrow. |
| Ordinal-churn residual pinned to the severity pin rather than to the shared key | **Real, fixed** in `355ceedc6`. |
| Q2: is the build-file scan redundant with the deps check? | **Reversed a deletion I was about to make.** It is not: the Release `deps.json` cannot see an official-only-conditional reference. Both checks stay, now documented as complementary. |
| Signature-half collision (two params rendering to identical text) | **Unverified suspicion, not actioned.** Neither of us could construct an instance in this closure that both collides and references stderr. It is the author-intent case the design declares out of scope. |

**MAI-Code — 3 rounds, 6 blocking claims, 2 real.** Both real ones were key collisions, and both
arrived with a wrong causal story attached to a correct suggested fix:

| Claim | Disposition |
|---|---|
| IL pin key collides across **overloads** | **Real, fixed** in `9a43b1c52` — key now includes the decoded signature. |
| IL pin key collides across **nested types** | **Real, fixed** in `e903e71c5`. A nested `TypeDefinition` has a **nil namespace**, so every nested type keyed bare — including the pin's own `ContainedWriter`, and every `<>c` display class. `MethodName` now walks to the outermost type. |
| | *But the reproduction did not demonstrate it*: the leak went through `/proc/self/fd/2`, which emits no `Console` reference, so the gate was green independently of any collision. Fixed because the weakness is real, not because the repro showed it. |
| Key is not a true method identity, so a write can hide behind an accounted key | **Rejected, reproduced.** Demonstrated by patching the built DLL with Mono.Cecil — past the boundary this rule defends. Decisively, the pin is a `Dictionary<string,int>` equality **including counts**, so a perfect collision is indistinguishable from "two references under one key", and that fails: a second `Console.Error.WriteLine` inside the accounted `WriteLine(string)` gives `Dictionaries differ`, `Failed: 1`. |
| 4 further "blocking" findings | **Dismissed** — each self-refuting by MAI-Code's own `Failed: 2` output, which showed the gate catching what the finding claimed it missed. |

**Gemini 3.1 Pro** declined rounds 26–27. **GPT-5.6 Sol** returned findings with no execution
evidence and is not counted.

**Round 29 (both models, at two successive heads) — clean.** Opus and MAI-Code independently
confirmed the `eng/` opt-out cannot reach the CLI closure, that the ownership gate still goes red on
a real uncontained write inside the closure, that the out-of-solution set is exactly six projects
(the solution file uses explicit `Path=` entries, no globs), and that `OwnsItsOwnStderr` does not
flow across `ProjectReference`. Opus also established something I had merely hoped: the vendored
`external/ILAssembler` carries its **own** root `Directory.Build.props`/`.targets` that deliberately
stop MSBuild's upward traversal, so this rule never reaches the vendored tree at all — and the one
remaining test that builds a project at run time (`MemberBodyProducerUnionTests.CompileWithSdk`)
compiles into `Path.GetTempPath()`, outside any directory build file. So there is no third instance
of the blind-spot class waiting.

Opus also caught a **false number in my own commit message**, which is the one finding I would most
have liked to catch myself. See the correction in the CI section below.

### What checking that finding surfaced

Rejecting the collision claim exposed the pin's actual limit, which is not the key. **Counting
bounds *where* stderr is reached, not *what* is written there.** Replacing the contained write in
`WriteLine(string)` with a raw one holds the count at 1 — this rule stays green while the line
stops being contained.

That case is covered, and the repo rule that an asserted property must name its gate means the
assertion had to say by what. `UntrustedArgumentDiagnosticContainmentTests` turns **eight** cases
red under exactly that tamper (full suite under tamper: `2729 total, 9 failed`; clean tree:
`Failed: 1`). The two rules are now documented as complementary — this one bounds *who may touch
the stream*, the behavioural one bounds *what the accounted code puts on it* — rather than as one
rule with a hole.

I also closed a collision avenue no reviewer tried. `SignatureDecoder.GetModifiedType` drops
modreq/modopt, and C# does permit overloads differing only by a function pointer's calling
convention (`delegate* unmanaged[Cdecl]<void>` vs `[Stdcall]` — both compile). The decoder still
renders them distinctly, because the convention rides the signature header rather than a modifier.

### What CI caught that I did not

Worth recording, because it is the same class of mistake as the one this PR was rewritten to fix.

The analyzer wiring is gated on `'$(MSBuildProjectExtension)' == '.csproj'`. I read that guard as
"projects, not file-based apps". It is not: **a file-based app builds through a generated project
whose extension is `.csproj`**, so the rule applied to `eng/*.cs` as well. Two of those are build
harnesses that legitimately write to `Console.Error`, so `eng/prepare-decompiler-package-sweep.cs`
failed with 36 `RS0030` errors and `eng/check-decompiler-gate.cs` with 32, taking down the `test`
and `decompiler-gates` jobs.

I had verified the rule against every project in the CLI's closure and against no file-based app at
all — which is exactly why the suite I never ran is the one that failed.

Fixed by `eng/Directory.Build.props` declaring `<OwnsItsOwnStderr>true</OwnsItsOwnStderr>`, the same
opt-out `tools/Directory.Build.props` already uses for harnesses, and by correcting the misleading
comment on the guard. `eng/` is outside the CLI's reference closure, so test 1's closure check still
covers everything that ships.

Measured: all four `eng/*.cs` apps `RS0030=0` (was 36 and 32 on two); `EvilPoolPinTests`
`Failed: 0` (was `Failed: 2`, matching CI); **remove `eng/Directory.Build.props` and it returns to
`Failed: 2`**, so the opt-out is load-bearing rather than incidental. Decompiler `--gate fast`
`4391/0`; ownership gate `4/0`.

Then CI found a **second** instance of the same class, which is the part that makes it worth a
section rather than a footnote. `tests/DotnetInspector.ILRoundtrip.Tests` is deliberately outside
`dotnet-inspect.slnx` — it depends on a vendored ILAssembler restored by a script — and
`IlasmScaffold.Assemble` captures that assembler's in-process stderr with `Console.SetError`. So
`dotnet build dotnet-inspect.slnx` never compiled it, and three more `RS0030` errors were waiting.

Same shape both times: **the rule is repository-wide and my validation was scoped**, first to
projects and then to the solution. I measured the rest of that gap instead of patching the instance:
six projects sit outside the solution, and the other five (four metadata fixtures and
`.review-fixtures/FixtureLib`) build `RS0030=0` and reference no banned symbol. The guard comment
now says outright that a solution build is not evidence for this rule.

Tamper-verified there too: flipping the property to `false` yields `3 Error(s)` at the three
`IlasmScaffold` sites, restoring yields `0 Error(s)`, and the round-trip fast suite is
`Total: 21, Failed: 0`.

*Correction to `c81600c6e`'s commit message, which I cannot amend:* it says that tamper yields
**6** errors. It yields **3**. I had counted with `grep -c 'error RS0030'`, and MSBuild prints each
error twice — once inline, once in its failure summary. A reviewer checked the build's own
`N Error(s)` line instead of my grep and found the real number, which also happens to be the number
CI reported. The direction of the tamper is unaffected; the magnitude claim was wrong. I re-counted
the `eng/` figures the same way in case they carried the same defect: they do not — 36 and 32 are
36 and 32 *distinct source sites*. A file-based app's build output has no failure-summary block, so
only the project build doubled.

### Residuals (declared, not discovered later)

- **Preprocessor branches the built configuration excludes.** Measured, not assumed: a
  `Console.Error.WriteLine` under `#if DEBUG` is missed in Release and caught in Debug. Not a
  shipping leak — Release is what ships, and the IL check reads the Release assembly.
- **A `ProjectReference` both created during a build and conditional on a shipping-only flavour**
  evades all four closure checks. Stated at exactly that width.
- **Local-function ordinal churn** — the IL key includes the compiler-assigned ordinal, so adding
  a lambda earlier in a method can force an unearned pin update. Stripping it would mean modelling
  Roslyn's generated-member naming, which is the move this class was rewritten to stop making.
- **fd-2 routes that emit no managed `Console` reference** (`/proc/self/fd/2`, reflection, P/Invoke)
  are out of scope for a rule expressed over managed symbol references.
- Open follow-ups tracked separately: #3418, #3463, #3481, #3482.

### Validation

```
dotnet build dotnet-inspect.slnx -c Release   → 0 Error(s), 6 Warning(s)*
stderr ownership gate                         → Total: 4, Failed: 0  (~7s)
ILInspector.Analysis.Tests                    → 584 passed
DotnetInspector.Services.Tests                → 357 passed
ILInspector.Metadata.Tests                    → 1005 passed
DotnetInspector.MetadataRendering.Tests       → 149 passed
dotnet-inspect.Tests                          → 2729, 1 pre-existing environmental failure**
```

\* Six pre-existing warnings inherited from `origin/main`, across four files in three test projects
(`OversizedMetadataVersionTests`, `AuthoredSourceValidityTests`, `ApiDiffAnalyzerTests`,
`DefaultValueRenderingTests`, `MethodHasBodyTests`) — four xUnit analyzer, two CS nullable.
`git diff --stat origin/main..HEAD` is empty for every one. They do not fail the build because
`TreatWarningsAsErrors` is scoped to `src/` and all six are under `tests/`. I had first written
this as "all six xUnit, one file"; a reviewer checked and it was wrong in both halves.

\*\* `Layout_Count_CountsFilesRatherThanRenderedTreeLines` — a local NuGet cache artifact (missing
`Newtonsoft.Json.nuspec`), not a branch regression. The evidence is head-independence: it fails
identically in a separate worktree at unmodified head `8e11ccd77`. Note CI cannot corroborate this
either way — its enclosing class carries a class-level `[Trait("Speed", "Slow")]` and the CI step
runs `-trait- "Speed=Slow"`, so CI never executes it. That is also why CI reports 1966 tests for
this project where a local full run reports 2729.

**Tamper verification** — every gate claim was tested by breaking it and watching it go red:
17 analyzer tampers (16 caught; 1 would not compile) and 7 gate tampers, including deleting the
`RS0030` escalation, adding a conditional `ProjectReference`, setting `NoWarn`, dropping a closure
reference, and both key collisions. Clean tree → `Failed: 0`.

**CI at this head (`c81600c6e`): all 8 jobs green**, including the two that found the breakages
above — `test` and `decompiler-gates`.

### One process note

At 147 files this PR is larger than it should have been. The containment work and the
ownership-proof work are separable, and in hindsight this was a stack. I'm not splitting it now —
the head is reviewed and green, and re-slicing would discard four clean fixed-head reviews — but
it is the honest read.

richlander:
Ready to merge

richlander:
## Head moved — not ready pending re-review

`main` moved under this branch and produced a conflict, so the `Ready to merge` above no longer applies. New head is `f92909c73`.

**Merge (`f2640d38a`)** — one conflict, in `src/ILInspector.Decompiler/MemberBodyProducer.cs`. Main's #3595 swept printers from `AppendLine` to `AppendLf` (LF on every platform); this branch had changed the same line from `EscapeIdentifier` to `ContainedIdentifier`. Resolved to take both. It was the file's last `AppendLine`, so the resolution leaves zero, consistent with that sweep.

Since #3595 is a manual sweep with no gate, I checked whether this branch reintroduces what it removed. It does not: for every file the sweep touched, the merged `AppendLine`/`Environment.NewLine` count equals `origin/main`'s exactly. The two product hits on branch-added lines (`NetworkTelemetry.cs`, `PackageCommand.cs`) wrap pre-existing calls rather than add new ones — counts match main there too. The nine added `Environment.NewLine` uses are all xUnit failure messages, which #3595 explicitly leaves platform-native.

**`f92909c73`** — the merge then failed `MarkoutRowContainmentTests` on the new `TypeInfoSection`'s ten string columns, which is the gate doing its job on code it had never seen. Each column is populated in `BuildTypeView` from the same local that already feeds the identically-named `TypeView` property, and all ten of those `TypeView` columns are already in the pinned set. So the section is a re-presentation joining the producer-contained side, not a new source of untrusted text. Tamper-verified: deleting `TypeInfoSection.Kind` fails the test naming that column.

That commit also corrects a measured claim in that test's own doc comment that was wrong *before* this merge — the pinned set held 349 members across 85 types, not the 350 stated. Now 359 across 86.

**Local validation at `f92909c73`:** solution `0 Error(s)`; `dotnet-inspect.Tests -trait- "Speed=Slow"` (the CI filter) `Total: 1967, Failed: 0`; decompiler `--gate fast` `Total: 4422, Failed: 0`; stderr ownership gate `Total: 4, Failed: 0`.

Re-review at this head is in progress; I'll re-post readiness once it and CI are clean.

richlander:
## Round 30 reconciliation — head moved to `26c13e68a`

Both reviewers returned on `f4eeb67d9`. Neither found a blocking defect in the commit, but Opus falsified two of my claims and found a real pre-existing leak. Merging main then exposed a second one. Both are fixed here.

### Findings

**Opus — "only this test catches it" was false.** `f4eeb67d9`'s message said the tamper on `TypeParameter.DisplayName` fails "this test, and here alone." Opus reproduced it and found **three** methods catch it: the new test, `LibraryViewShapeDerivedContainmentTests` (both `topFieldsOnly` cases), and `UntrustedViewContainmentTests`. My error was procedural — I ran the one class under tamper and the full suite untampered. Corrected in `26c13e68a`'s message. The new test is focused defense in depth, not the unique guard.

**Opus — genuine pre-existing leak, now fixed.** The Type Info section's Type Parameters row is composed as `{tp.DisplayName} : {constraints}`. `DisplayName` contains; the constraint half went through `CSharpDeclarationWriter.FormatConstraintList`, which only escapes reserved keywords — that turns `class` into `@class` and leaves a bidi override exactly where it was. A hostile **constraint type name** reached the terminal raw, at default verbosity as well as through the section. The fixture's generic parameter now carries a hostile-named interface constraint, which reproduced the leak as a red test before the fix; `FormatConstraintList` now contains its composed result.

This also corrects the framing of `f92909c73`: the row is a **mix** of contained producer and genuine residual, not uniformly contained, so its `NotSelfContaining` entry was doing real work.

**Opus — I was right, they were wrong, on one point.** Opus's first pass reported the row as renderer-contained because `tp.DisplayName` is raw at `BuildTypeView`. It is not: `Name` is the raw identity, `DisplayName` calls `ContainComposedName`. Opus verified and withdrew this.

**MAI — clean, with independent end-to-end evidence.** Built its own hostile assembly, hexdumped the rendered `Type Info` bytes, confirmed escaped-not-raw, and independently reproduced the 37-file count equality against main's LF sweep. It also confirmed the exposure asymmetry: `Type Info` renders at every verbosity while the inline twin is quiet-only.

### Merging main (#3602) exposed a second leak

`--trace` wrote its report with a raw `Console.Error.Write`. The analyzer caught it **on the merge path only** — the leak exists just once main and this branch are combined, which is why a local build was clean and CI was not. The head line interpolates the target name, which is argv. On a target named `ev<U+202E>il\nError: FORGED.dll`: raw U+202E bytes on the stream **plus a forged unindented `Error: FORGED.dll` line**.

Routing it through `CommandError.WriteLine` per line was **not enough on its own**. `Render()` returned one terminated string, so its writer had to recover line boundaries by splitting — and a splitter cannot tell the composer's newline from one the attacker put inside a field. That intermediate state escaped the bidi override and still printed the forged line. `InspectionTrace` now yields lines (`RenderLines`), so each line is a unit the writer can contain and nothing re-derives a boundary from text. This is the same conclusion `ContainedWriter`'s doc comment already records.

Byte-neutral for benign input: `library <dll> --trace -v q` is identical to the pre-fix binary (`cmp`, no diff). Gated as a 12th argv channel across all 7 hazards. Tamper-verified: reinstating the split-a-terminated-string shape compiles with `0 Error(s)` and fails exactly that channel, on the line-feed case.

### Validation at `26c13e68a`

| | |
|---|---|
| solution | `0 Error(s)`, `0 Warning(s)` |
| dotnet-inspect.Tests `-trait- "Speed=Slow"` | Total: 1982, Failed: 0 |
| Decompiler `--gate fast` | Total: 4422, Failed: 0 |
| Analysis / Services / Metadata | 584 / 357 / 1005, all 0 failures |

**Not ready to merge.** The head moved, so a clean fixed-head review is required. Both reviewers are re-dispatched at `26c13e68a`; CI is running.

richlander:
## Fixed-head review at `26c13e68a` — both clean

Both reviewers re-reviewed at this exact head. **Neither found a blocking defect.**

### Opus 4.8 — non-blocking, "ship it"

Reproduced both leaks as closed, end to end:

- **`--trace`**: on target `ev<U+202E>il\nError: FORGED.dll` the output is `trace: library ev\u202Eil Error: FORGED.dll [Quiet]` — attacker newline folded to a space, bidi escaped, no forged unindented line. Confirmed `RenderLines` has exactly one caller and every yielded line reaches the stream through `CommandError.WriteLine` → `ContainRenderedText`.
- **`RenderLines` byte-neutrality**: mapped all ~15 yields against the old `Render()` line by line — identical text and order, including both `PadRight(28)` sites and the `"  added by prerequisite "` line where I folded a separate `.Append(' ')` into the literal.
- **The constraint fix is *more* complete than I claimed.** I said containment belonged in `FormatConstraintList` because two call sites would otherwise each have to remember. Opus found **all four** production callers carried the identical half-guarded shape — contained parameter name beside an uncontained constraint list: `ApiOutputFormatter:675`, `MemberBodyProducer:636`, `CSharpDeclarationWriter:776`, `CSharpFormatter:158`. One chokepoint closes all four. It also confirmed the sibling risk (the `where`-clause parameter name) is already contained at each site, so the row is now uniformly contained everywhere it renders.
- **Gate is load-bearing, and stronger than I claimed.** Reinstating the split shape compiles `0 Error(s)` and, across the full 2765-test project, the only code-relevant failures are the `trace` channel — failing on *all* its line-terminator hazards, not just the line-feed case I reported.

### MAI-Code — no blocking issues

Independent byte-level checks: hostile targets carrying U+202E, U+2028, U+2029, LF, CR, VT and ESC through `--trace`, across `--package`/`--platform` and all four verbosities — contained in every case, no forged line. Independently reproduced the byte-neutrality claim by building the pre-fix source with `-p:OwnsItsOwnStderr=true` and `cmp`-ing stderr: identical. Built its own hostile-constraint assembly and confirmed `T : IHostile\u0012INJECTED` escaped rather than raw, through both the default tree and the section.

### Two corrections to my own record

1. **`0 Warning(s)` was wrong — it is 6.** Opus caught this. My figure came from an *incremental* build, which does not re-report warnings from projects it did not rebuild. Verified with `-t:Rebuild`: `6 Warning(s)`, all in `tests/ILInspector.Metadata.Tests` (`ApiDiffAnalyzerTests.cs:36,37`, `DefaultValueRenderingTests.cs:26`), none in any of the 5 changed files. Pre-existing, no new warnings introduced. **The `0 Warning(s)` line in `26c13e68a`'s commit message is inaccurate; this comment is the correction.**
2. **`CommandExecutionTests.Layout_Count_CountsFilesRatherThanRenderedTreeLines`** fails in a full-project run. Opus isolated it as failing deterministically on *untampered* HEAD because it fetches `Newtonsoft.Json@13.0.4` and the sandbox has no network. It is `[Trait("Speed","Slow")]`, so CI's `-trait- "Speed=Slow"` lane never runs it. Pre-existing and environmental, unrelated to this PR — recorded so it is not mis-attributed later.

### Final state at `26c13e68a`

| | |
|---|---|
| CI | **all 8 jobs green** |
| solution | `0 Error(s)`, 6 pre-existing warnings |
| CLI `-trait- "Speed=Slow"` | 1982 / 0 |
| decompiler `--gate fast` | 4422 / 0 |
| Analysis / Services / Metadata | 584 / 357 / 1005, all 0 failures |
| reviews | Opus 4.8 clean · MAI-Code clean, both at this head |

Ready to merge

PR #3382, opened 2026-07-28
Wrap long `with` and anonymous-object initializers Allman-style (#3371)

### LINKED ISSUE #3371: Decompiler: length-based pretty printing for object/collection initializers (long lines)

richlander:
Closes #3371 (follow-up to #3377).

## Summary

#3377 broke long **object/collection** initializers Allman-style via an always-on
brace-body width wrapper, but deferred two sibling forms that render through the
same `head { e0, e1 }` shape and hit the same long-line problem:

- record **`with`** expressions — `receiver with { A = x, B = y, … }`
- **anonymous objects** — `new { Name = value, … }`

This extends the wrapper to both, and refactors #3377's initializer-specific code
into a shared core rather than adding parallel abstractions.

## Change

- **Extract** the width-check / whitespace-variant-guard / Allman-emit body of
  #3377's `ObjectInitializerLines` into `BraceBodyLines(head, entryTexts, flat, prefix, suffix, indent)`.
- **Three thin callers** feed it the head + entry texts the inline renderers already
  emit: `ObjectInitializerLines` (unchanged behavior), `WithExpressionLines`
  (head `receiver with`), `AnonymousObjectLines` (head `new`). The anonymous
  projection parts are factored into a shared `AnonymousObjectParts` so the inline
  and wrapped forms spell identical tokens.
- **Unify** the statement recognizer/caller: `TryObjectInitializerStatement` /
  `TryAppendObjectInitializer` → `TryBraceBodyStatement` / `TryAppendBraceBody`,
  matching `ObjectInitializerExpression`, `WithExpression`, and `AnonymousObject`
  in `return` / non-ref local or stack-slot store position and dispatching per
  value type.

## Byte-neutrality

Whitespace-only, same as #3377. A wrapped form is admitted only when **both** guards
hold: the caller's exact-match `line == prefix + <inline body> + ";"` (so a coercion
cast keeps it inline) and `BraceBodyLines`'s reconstruction `head + "{ " + join + " }" == inline`.
No trailing comma. The object-initializer path is the same shared core, so its
compile-back RoundTrip gates cover the core; the whitespace-variant tests prove the
two new arms are token-identical to their inline forms.

- **IL fidelity:** True — object-initializer compile-back RoundTrip gates unchanged;
  the two new arms are proven pure whitespace variants. (The 3 `parameter-attributes` /
  record-helper / lowering-frontier `RecompileFail`s in `Area=RoundTrip` are
  pre-existing/environmental, same set #3377's reviewers confirmed on base.)
- **Taste applied:** none — always-on formatting, not a style lens.

## After (verbatim product output at head `8512014b`)

Real compiled fixtures (`BraceBodyWrappingSamples`, appended at end of the fixture
file so no generated-code ordinals shift), `-S "Decompiled Source"`:

Record `with` (`WidenMeasuredRecord`):

```csharp
public static ILInspector.Decompiler.Tests.MeasuredRecord WidenMeasuredRecord(ILInspector.Decompiler.Tests.MeasuredRecord source, int first, int second, int third, int fourth) => source with
{
    FirstMeasuredValue = first,
    SecondMeasuredValue = second,
    ThirdMeasuredValue = third,
    FourthMeasuredValue = fourth
};
```

Anonymous object (`ProjectMeasuredValues`):

```csharp
public static object ProjectMeasuredValues(int first, int second, int third, int fourth) => new
{
    FirstMeasuredProjection = first,
    SecondMeasuredProjection = second,
    ThirdMeasuredProjection = third,
    FourthMeasuredProjection = fourth
};
```

## Before (pre-change behavior)

Each rendered as a single flat line (the inline `WithExpressionText` /
`AnonymousObjectText` join, which `DisableOneLinerWrapping` still produces) — e.g.:

```csharp
... WidenMeasuredRecord(...) => source with { FirstMeasuredValue = first, SecondMeasuredValue = second, ThirdMeasuredValue = third, FourthMeasuredValue = fourth };
... ProjectMeasuredValues(...) => new { FirstMeasuredProjection = first, SecondMeasuredProjection = second, ThirdMeasuredProjection = third, FourthMeasuredProjection = fourth };
```

- Valid: True — Correct: True — IL fidelity: True — Taste applied: none (both Before and After)

## Tests

- `WithAndAnonymousWrappingTests` (hand-built IR): break / inline / single-long /
  `DisableOneLinerWrapping` / whitespace-variant, for both `with` and anonymous.
- `WithAndAnonymousWrappingFixtureTests` (real compiled fixtures): both fold
  multi-line end-to-end (the `with` raises to a 4-entry `WithExpression`).
- Existing #3377 `ObjectInitializerFormattingTests`: unchanged, still green (refactor
  is behavior-preserving).
- Fast decompiler suite (`-trait- "Speed=Slow"`): **4059/0**.
- Slow `Area=RoundTrip`: only the 3 known pre-existing env failures.

## Scope

Completes the initializer-family wrapping. Still deferred (separate concerns):
method/parameter-list signature wrapping; recursive wrapping of nested initializers.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -6136,3 +6136,38 @@ public static class FlagsEnumAccumulatorSamples
         return (int)caps;
     }
 }
+
+// #3371 follow-up witnesses: a record `with` expression and an anonymous object
+// wide enough that the printer's brace-body width wrapper breaks them Allman-style
+// (one entry per line). New top-level types appended at end of file so they cannot
+// shift any existing CfgSampleClass generated-code ordinals.
+public sealed record MeasuredRecord(
+    int FirstMeasuredValue,
+    int SecondMeasuredValue,
+    int ThirdMeasuredValue,
+    int FourthMeasuredValue);
+
+public static class BraceBodyWrappingSamples
+{
+    // `source with { A = .., B = .., C = .., D = .. }` — flat form exceeds 120 cols.
+    public static MeasuredRecord WidenMeasuredRecord(MeasuredRecord source, int first, int second, int third, int fourth)
+        => source with
+        {
+            FirstMeasuredValue = first,
+            SecondMeasuredValue = second,
+            ThirdMeasuredValue = third,
+            FourthMeasuredValue = fourth,
+        };
+
+    // Anonymous types are reference types, so returning one as `object` needs no
+    // box/cast; the anonymous object stays the bare return value. Explicit
+    // `Name = value` form (value names differ from property names), flat > 120 cols.
+    public static object ProjectMeasuredValues(int first, int second, int third, int fourth)
+        => new
+        {
+            FirstMeasuredProjection = first,
+            SecondMeasuredProjection = second,
+            ThirdMeasuredProjection = third,
+            FourthMeasuredProjection = fourth,
+        };
+}
--- a/src/ILInspector.Decompiler.Tests/WithAndAnonymousWrappingTests.cs
+++ b/src/ILInspector.Decompiler.Tests/WithAndAnonymousWrappingTests.cs
@@ -1,0 +1,158 @@
+using System.Collections.Immutable;
+using ILInspector.Decompiler.Pipeline;
+
+namespace ILInspector.Decompiler.Tests;
+
+// #3371 follow-up: the always-on brace-body width wrapper that breaks long object
+// initializers Allman-style (#3377) also covers record `with` expressions and
+// anonymous objects — the same head + one-entry-per-line shape, no trailing
+// comma, so breaking stays whitespace-only and token-identical to the inline form.
+public sealed class WithAndAnonymousWrappingTests
+{
+    static readonly TypeRef Holder = TypeRef.Definition("synthetic", "", "Holder");
+    static readonly TypeRef Record = TypeRef.Definition("synthetic", "", "MeasuredRecord");
+    static readonly TypeRef Anon = TypeRef.Definition("synthetic", "", "<>f__AnonymousType0");
+    static readonly TypeRef Int32 = TypeRef.CoreLib("System", "Int32");
+
+    static string Render(IrExpression value, TypeRef returnType, PrinterOptions? options = null)
+    {
+        var block = new Block(0);
+        block.Add(new Return(value));
+        var container = new BlockContainer();
+        container.Add(block);
+        var signature = new MethodSignature(returnType, ImmutableArray<Parameter>.Empty, HasThis: false, GenericParameterCount: 0);
+        var function = new IrFunction("M", Holder, signature, [], container);
+        return CSharpPrinter.Print(function, options ?? PrinterOptions.Default).Output!;
+    }
+
+    static WithExpression With(params string[] members)
+    {
+        var receiver = new LoadArgument(0, "original", Record);
+        var entries = members.Select(m => new InitializerEntry(m, [new LoadArgument(1, "value_" + m, Int32)]));
+        return new WithExpression(receiver, entries);
+    }
+
+    static WithExpression LongWith()
+        => With("FirstMeasuredProperty", "SecondMeasuredProperty", "ThirdMeasuredProperty", "FourthMeasuredProperty");
+
+    static AnonymousObject Anonymous(params string[] names)
+    {
+        var values = names.Select(n => (IrExpression)new LoadArgument(1, "value_" + n, Int32));
+        return new AnonymousObject(Anon, [.. names], values);
+    }
+
+    static AnonymousObject LongAnonymous()
+        => Anonymous("FirstMeasuredProperty", "SecondMeasuredProperty", "ThirdMeasuredProperty", "FourthMeasuredProperty");
+
+    [Fact]
+    public void LongWith_BreaksAllmanOnePerLine()
+    {
+        string body = Render(LongWith(), Record);
+
+        Assert.Contains("original with\n", body);
+        Assert.Contains("\n{\n", body);
+        Assert.Contains("    FirstMeasuredProperty = value_FirstMeasuredProperty,\n", body);
+        Assert.Contains("    FourthMeasuredProperty = value_FourthMeasuredProperty\n", body);
+        Assert.DoesNotContain("value_FourthMeasuredProperty,", body);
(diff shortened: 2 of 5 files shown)


richlander:

Two-model adversarial review at fixed head `8512014b` (broad-blast-radius printer change).

| --- | --- | --- |
| rev2-gpt | gpt-5.6-sol | **CLEAR** |
| rev2-gemini | gemini-3.1-pro-preview | **CLEAR** |

Both reviewers ran the full evidence set at the exact head:

- Build: 0 errors.
- `WithAndAnonymousWrappingTests` 6/6, `WithAndAnonymousWrappingFixtureTests` 2/2.
- #3377 `ObjectInitializerFormattingTests` 5/5 (refactor-equivalence confirmed).
- Fast suite green (4059/4059).
- `Area=RoundTrip`: only the 3 known pre-existing/environmental failures
  (`ReturnToSenderRecordCatalog_CoversGeneratedRecordHelpers`,
  `CompilerLoweringFrontier_IsSelectableButNotInDefaultRun`, and the
  `parameter-attributes` minimal compile-back rung) — not attributable to this PR.

Findings: **none** from either reviewer. Both independently verified all attack points:
`BraceBodyLines` refactor is byte-for-byte equivalent to #3377's `ObjectInitializerLines`;
`with`/anonymous reconstruction guards hold; the exact-match caller guard keeps
cast/coercion-wrapped statements inline; ByRef `StoreLocal`/`StoreStackSlot` are excluded
and the dispatch `switch`es never hit the fallback arm; the 4-entry `with` fixture is really
raised (wrapping is not dead code); expression-body folding indents correctly; and the
appended `CfgSampleClass` fixtures shift no generated-code ordinals.

CLI witnesses confirmed the wrapped Allman output for both `WidenMeasuredRecord` (4-entry
`with`) and `ProjectMeasuredValues` (anonymous object).

Ready to merge

PR #3385, opened 2026-07-28
Make --json compose with print and shape projections

### LINKED ISSUE #3379: --json silently ignores --print/--urls/--paths/--value on type and member
On the `type` and `member` paths, `--json` returns ordinary API JSON and silently ignores `--print`, `--urls`, `--paths`, and `--value`. Cardinality validation never runs and payloads are never fetched.

`ApiCommand.cs` returns early:

```csharp
if (options.JsonOutput && !options.Count)
{
    WriteJsonTypeOutput(type, options);
    return 0;
}
```

roughly 190 lines before the projection dispatch:

```csharp
if (options.Print)
    return await PrintApiProjectionAsync(view, options);

if (options.Value || options.Urls || options.Paths)
    return WriteApiShapeProjection(view, options);
```

`--count` is honored only because of the explicit `&& !options.Count` guard. The other four projections have no such guard.

## Reproduction

Using a library whose `Source Files` section has two printable rows, the second pointing at a URL that 404s:

```console
$ dotnet-inspect type Mix --library ... -S "Source Files" --print --row 2 --json
{
  "name": "Mix",
  "metadata_name": "Mix",
  ...
                                  # exit 0, empty stderr
```

Without `--json` the same selection correctly fails:

```console
$ dotnet-inspect type Mix --library ... -S "Source Files" --print --row 2
Error: failed to fetch the document for printable row 2 from https://.../NO-SUCH-FILE-3364.md
                                  # exit 1
```

Multi-row cardinality is skipped too — `--print --json` with two printable rows exits 0 instead of demanding `--row`. The same applies on `member`.

All four projections are swallowed:

```console
$ ... -S "Source Files" --urls   --json   ->  type JSON
$ ... -S "Source Files" --paths  --json   ->  type JSON
$ ... -S "Source Files" --value  --json   ->  type JSON
$ ... -S "Source Files" --count  --json   ->  2          (correct)
```

## Why it matters

`docs/design/output-shapes.md` treats `--json` as an output *format* and `--print`/`--urls`/`--paths`/`--value` as shape *projections*; they are meant to be orthogonal. Here the format silently discards the projection instead of composing with it or rejecting the combination.

It also defeats the acquisition guarantee: `--print` is supposed to either produce the selected payload or fail loudly, and `--json` bypasses both halves.

Either the projection should run and serialize as JSON, or the combination should be rejected — but it should not be silently ignored.

Found by adversarial review of #3373. **Pre-existing, not a regression:** verified identical behavior on `924e6e0c` (the PR's base) and on the PR head; `git diff` confirms the early-return ordering is untouched.

richlander:
Fixes #3379. Second PR in the #3364 stack, on top of #3373.

`--json` selects an output *format*; `--print`, `--value`, `--urls`, and `--paths` select an output *shape*. On `type` and `member` they did not compose — the type-surface early return at `ApiCommand.cs:649` preceded the projection dispatch ~190 lines later, so `--json` silently discarded the projection and exited 0.

This is the same failure class #3373 just fixed: a projection request accepted and then dropped without a word. It also defeated that PR's acquisition guarantee, since neither cardinality validation nor the fetch ever ran under `--json`.

## Why compose rather than reject

This is not a new design decision. Both projection writers already implement JSON serialization (`PrintProjectionOutput.cs:69`, `ShapeProjectionOutput.cs:87`), and the sibling format flags already compose today:

| Flag | With `--print` on `type` (before) | Why |
| --- | --- | --- |
| `--jsonl` | works | not in the early-return guard |
| `--json-array` | works | not in the guard; `:142` *requires* a projection |
| `--json` | **silently bypassed** | type JSON returned first |
| `--count` | works | explicit `&& !options.Count` guard |

`--json` was the lone outlier among four format flags. Rejecting the combination would mean refusing something the code already knows how to serialize, and would make `--json` gratuitously different from `--jsonl`. The fix makes the existing intended shape reachable.

## Change

The guard now mirrors how `--count` already exempts itself:

```csharp
if (options.JsonOutput && !options.Count && !IsProjectionRequested(options))
```

Flow then falls through to the existing dispatch. I deliberately did not hoist the dispatch instead: the projections need `view`, which is not built until `:655`, so reordering would drag view construction along for no behavioral gain.

## Evidence

Real runs against a compiled two-row SourceLink fixture whose second row 404s, baseline `729f97bf` vs head:

| Command | Baseline | Head |
| --- | --- | --- |
| `--print --json` (2 rows) | type JSON, **exit 0** | `selected section has 2 printable rows`, exit 1 |
| `--print --row 2 --json` (404) | type JSON, **exit 0** | `failed to fetch the document for printable row 2 from <url>`, exit 1 |
| `--urls --json` | type JSON, exit 0 | projected rows array, exit 0 |

Untouched paths are byte-identical to baseline (md5-compared): plain `--json`, `--count --json`, and `--print` without `--json`.

Each of the five `--json` tests was confirmed to **fail** with the guard reverted and pass with it restored, so they pin the behavior rather than merely describing it.

`Total: 2233, Errors: 0, Failed: 0` (CLI) and `Total: 319, Errors: 0, Failed: 0` (services). markdownlint clean.

## Test seam

#3373 shipped with a declared gap: no automated coverage of the fetch-failure branch, because `PrintUrlProjectionAsync` builds its `SourceFetcher` from the static `HttpClientFactory.SharedUntrustedFetch`. This PR closes it with `SetUntrustedFetchForTesting`, following the `ResetSharedForTesting` convention already in that file.

The seam substitutes **only the transport**, so tests still exercise the real `SourceFetcher` — scheme restriction, caching, and status handling all stay under test. That matches the harness-boundary rule in `AGENTS.md`: the harness supplies a fixture, not a second implementation. `Type_SourceFiles_PrintRow_FetchFailureIsHardError` passes on baseline by design; it covers PR 1's branch rather than new behavior.

## Scope

`LibraryCommand.cs:603` and `PackageCommand.cs:660` look like the same pattern but are unrelated helpers (an IL-coordinate batch writer and the multi-package path); neither precedes a projection dispatch. `type` and `member` share the fixed method, so one guard covers both.

Docs: the `--json` row in `output-shapes.md` said "the whole Document as one JSON object", which is only true when the Document *is* the selected shape. Reworded to match the model the same document already states. This aligns the doc with its own framing rather than weakening it to fit behavior.

## Adversarial review

Behavior change to output composition on two commands — needs two-model review.

### DIFF
--- a/docs/design/output-shapes.md
+++ b/docs/design/output-shapes.md
@@ -205,7 +205,7 @@ payloads on already evaluated rows; it cannot probe missing cells.
 | Flag | Effect |
 | --- | --- |
 | `--markdown` | force the full Markdown Document format |
-| `--json` | the whole Document as one JSON object |
+| `--json` | render the selected shape as JSON: the whole Document when no narrower shape is selected, otherwise the projected payload (`--print`, `--value`, `--urls`, `--paths`) |
 | `--tsv` / `--jsonl` | render the single selected section as TSV / JSON Lines (a Table or Vector) |
 | `--table` | render the single selected section as a space-padded pretty table |
 | `--no-header` (`--no-headers`) | drop the Table header row |
--- a/src/DotnetInspector.Core/HttpClientFactory.cs
+++ b/src/DotnetInspector.Core/HttpClientFactory.cs
@@ -14,6 +14,7 @@ public static class HttpClientFactory
     private static bool _offline;
     private static HttpClient? _shared;
     private static HttpClient? _sharedUntrustedFetch;
+    private static HttpClient? _untrustedFetchOverride;
     private static IDisposable? _networkTrafficLoggingSubscription;
 
     /// <summary>
@@ -55,6 +56,7 @@ public static class HttpClientFactory
     {
         _shared = null;
         _sharedUntrustedFetch = null;
+        _untrustedFetchOverride = null;
         _networkTrafficLoggingSubscription?.Dispose();
         _networkTrafficLoggingSubscription = null;
     }
@@ -70,7 +72,15 @@ public static class HttpClientFactory
     /// in untrusted artifacts (SourceLink URLs embedded in a PDB, etc.). Use this — not <see cref="Shared"/> —
     /// for any URL that came from inspected package/PDB data. Do not dispose.
     /// </summary>
-    public static HttpClient SharedUntrustedFetch => _sharedUntrustedFetch ??= CreateUntrustedFetchClient();
+    public static HttpClient SharedUntrustedFetch => _untrustedFetchOverride ?? (_sharedUntrustedFetch ??= CreateUntrustedFetchClient());
+
+    /// <summary>
+    /// Test-only: substitutes the transport used by untrusted-source fetches so acquisition
+    /// paths (including failure) can be exercised without network access. This replaces only
+    /// the transport; callers still run the real <c>SourceFetcher</c>, so scheme restriction,
+    /// caching, and status handling stay under test. Pass null to restore the real client.
+    /// </summary>
+    internal static void SetUntrustedFetchForTesting(HttpClient? client) => _untrustedFetchOverride = client;
 
     /// <summary>
     /// Whether <paramref name="url"/> is an absolute http/https URL. Untrusted-source fetches restrict
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -3534,6 +3534,151 @@ public class CommandExecutionTests
         Assert.Contains("JsonReader", single.GetProperty("content").GetString());
     }
 
+    /// <summary>
+    /// <c>--json</c> selects an output format and <c>--print</c> selects an output shape,
+    /// so they compose: the projection owns the request and the plain type surface must not
+    /// claim it. Regression for #3379, where the type-surface early return preceded the
+    /// projection dispatch and silently discarded <c>--print</c> with exit 0.
+    /// </summary>
+    [Fact]
+    public async Task Type_SourceFiles_PrintJson_EmitsSelectedDocumentNotTypeSurface()
+    {
+        var (exit, output, error) = await RunAppAsync(
+            "type", "JsonReader", "--package", "Newtonsoft.Json@13.0.3",
+            "-S", "Source Files", "--print", "--row", "1", "--json", "--raw", "--tips", "q");
+
+        Assert.Equal(0, exit);
+        Assert.Empty(error);
+        using var document = JsonDocument.Parse(output);
+        Assert.Equal(1, document.RootElement.GetProperty("row").GetInt32());
+        Assert.Equal("Source Files", document.RootElement.GetProperty("section").GetString());
+        Assert.EndsWith("/Src/Newtonsoft.Json/JsonReader.cs", document.RootElement.GetProperty("url").GetString());
+        Assert.Contains("JsonReader", document.RootElement.GetProperty("content").GetString());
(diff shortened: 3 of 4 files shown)


richlander:

Two models reviewed this PR per `AGENTS.md`: **Gemini 3.1 Pro** and **GPT**, each in its own isolated linked worktree at a pinned head. Round 1 reviewed `f1472343`; round 2 re-reviewed the fixed head `164c6ae1`.

### Finding actioned

**Gemini — the fetch test seam could leak a disposed client (blocking, test integrity).** The two new fetch-failure tests called `SetUntrustedFetchForTesting(client)` and `NuGetCache.Initialize(...)` *before* entering the `try`. Because the override is a mutable static and the stub client is `using var`-scoped, a throw during setup would skip the `finally` and leave `HttpClientFactory.SharedUntrustedFetch` pointing at a disposed `HttpClient` — poisoning every later test in the collection with `ObjectDisposedException`.

This was real and introduced by this PR. Fixed in `164c6ae1` by moving both static mutations inside the `try`, making cleanup unconditional. Both reviewers re-verified the structure is now exception-safe.

I did **not** adopt Gemini's alternative suggestion of making `_untrustedFetchOverride` an `AsyncLocal<HttpClient?>`, and asked round 2 to pressure-test that call rather than accept it. Gemini confirmed the simpler fix is sufficient with specific evidence: every test that mutates the override or `NuGetCache` lives in `[Collection("Console")]`, which xUnit runs sequentially, and no concurrently-running class reaches `SharedUntrustedFetch`. A plain static matches the `ResetSharedForTesting` convention already in that file; `AsyncLocal` would be heavier machinery for a hazard the `try` placement already closes.

### Findings verified as pre-existing, and deferred

**GPT — `--fields`/`--columns` are ignored under `--json`,** and **the whole-surface type path at `ApiCommand.cs:498` ignores projections entirely** (a 1,012,886-byte `ApiSurface` document where one value was requested). Both confirmed byte-identical on base `729f97bf`, so neither is a regression. Filed together as #3386.

These are the same failure class this PR fixes, on adjacent paths. Fixing them here would have meant either projecting fields inside `WriteJsonTypeOutput` or building a projection dispatch for the whole-surface route — both larger changes than the ordering fix this PR makes, and neither needed to close #3379. See the note below on why the pattern itself may deserve attention.

### Non-actions

Neither reviewer found a blocking issue against the fixed head. GPT independently reproduced the guard-revert experiment — reverting `IsProjectionRequested` and rebuilding (after `touch`, since an `mv`-based restore preserves mtime and silently fools the incremental build) makes all five `--json` tests fail. That confirms the tests pin the behavior rather than merely describing it.

Both reviewers confirmed the untouched paths — plain `--json`, `--count --json`, `--print` without `--json`, `--jsonl`, `--json-array` — are byte-identical to base in exit code, stdout, and stderr.

`Type_SourceFiles_PrintRow_FetchFailureIsHardError` passes on base by design. It covers the branch #3373 shipped without automated coverage, rather than testing behavior new to this PR.

### Fixed-head verification

Both round-2 reviews report clean at `164c6ae1`, with 2233 tests passing, 0 failures, 14 expected skips, and a 0-warning Release build.

### Structural note

Counting #3364, #3379, and both halves of #3386, this is the fifth site where a projection request is accepted and then silently discarded with exit 0. Each fix has been local to one dispatch site. The recurrence suggests the underlying gap is that nothing forces a projection request to be either honored or rejected — worth addressing as its own change rather than continuing to patch sites individually. Not proposed for this PR.

richlander:
Ready to merge.

Both fixed-head adversarial reviews are clean at `164c6ae1`, CI is green (build-net10, test, markdownlint, changes), and the reconciliation above accounts for every finding from both models.

#3386 is non-blocking follow-up: both halves were verified byte-identical on base `729f97bf` and are untouched by this PR.

PR #3389, opened 2026-07-28
Reject duplicate JSON properties at untrusted parse boundaries, and gate JSON wire names

richlander:
## Summary

Two related changes from a review of the System.Text.Json additions in .NET 10 and .NET 11.

**Adopted, from .NET 10:** `AllowDuplicateProperties = false` at untrusted JSON parse boundaries.

**Not adopted, from .NET 11:** nothing. Details below, since the investigation was the point of the exercise.

## The duplicate-property hardening

JSON does not define how duplicate object keys resolve, so two readers of one payload can disagree. SourceLink is a live instance rather than a hypothetical:

- `SourceDocumentPathResolver.Create` orders mappings by descending pattern length and takes the first match.
- The repository-URL reader in `AssemblyInspector` takes the first value naming a known host.

Given a repeated key under `documents`, those two rules select different entries. Measured on the pre-change readers:

```
mappings seen: 2
Reader A (resolved source URL): https://evil.example/raw/*
Reader B (reported repo URL):   https://github.com/o/r/raw/*
```

A hostile PDB could therefore resolve source from one origin while the tool reported provenance from another — the spoofed-provenance case in `docs/design/untrusted-data-threat-model.md`.

Adds `HardenedJson` to `DotnetInspector.Core`, directly alongside the existing `HardenedXml`. Twenty parse sites move onto it: SourceLink maps, package contents, feed responses, `project.assets.json`, `.deps.json`, and the package index cache.

**Layering note for reviewers.** `ILInspector.Metadata` gets its own `SourceLinkJson` rather than a `DotnetInspector.Core` reference. Metadata references only Findings, MetadataPrimitives, and Text; depending on Core to share ten lines would pull HTTP, caching, and download infrastructure into the SRM layer and invert the ownership documented in `docs/overview.md`. The duplication is deliberate — please push back if you disagree.

## Compatibility boundary

Callers that already treated malformed JSON as "no data" now treat duplicate-bearing JSON the same way. That is fail-closed and consistent with their existing malformed-input handling, but it does **not** convert them to explicit failure reporting, which remains open work item 6 in the threat model. NuGet-produced `deps.json` and `project.assets.json` do not legitimately contain duplicate keys, so real-world impact should be nil.

`runfaster` still parses its trace inputs directly; recorded as open work item 1 rather than silently expanded into this change.

## The wire-name gate

Naming is declared per context, not per model. Twenty-eight contexts repeat the declaration by hand, and several model types are serialized by more than one context (the indented/compact pairs; rows shared between a document context and a JSONL context). Nothing ties those declarations together, so a context added without `PropertyNamingPolicy` would silently emit a second spelling of an existing shape.

The gate discovers contexts by reflection rather than listing them, so a new context is covered the moment it is declared, and walks each graph transitively — naming attributes are not transitive, so nested types are exactly where a spelling change would go unnoticed.

**Evidence it is not vacuous.** Both assertions were verified against tampered builds:

| Tamper | Result |
| --- | --- |
| Remove `PropertyNamingPolicy` from `PackageFileJsonRowContext` | `EveryWireName_MatchesItsDeclaredNamingPolicy` fails, naming `PackageFileJsonRow.Path`, `.Size` |
| Flip `ApiTypeCompactJsonContext` to CamelCase | `SharedTypes_HaveOneSpellingAcrossEveryContext` fails, naming `ApiType` plus nested `TypeParameter`, `ApiMember`, `SampleReference`, `PartialSourceFileInfo` |

A `Gate_IsNotVacuous` test fails if discovery stops finding contexts.

## The gate found a real deviation

`TimelineJsonContext` declares no naming policy and its views carry no `[JsonPropertyName]`, so **`timeline --json` emits CLR PascalCase** (`Member`, `Finding`, `Address`, …) while every other command emits snake_case.

Correcting it changes a shipped wire format, so this PR does not do it. It is recorded as a named exception, and `KnownPascalCaseDeviations_StillDeviate` fails once it is fixed, forcing the exception to be deleted alongside the fix. Happy to follow up separately if you want the wire format aligned.

## Why no .NET 11 features

`JsonNamingPolicyAttribute` was the only serious candidate — it would let naming live on the model instead of being copy-pasted across contexts. It does not hold up: the attribute overrides a context policy for a type's own properties but is **not transitive**.

```
ctx=SnakeCase, Outer=[CamelCase]: {"outerProp":"a","nested":{"nested_prop":"b"}}
                                                  nested type ignored the attribute
```

Making naming genuinely intrinsic would mean attributing every type in the transitive graph (~100+), which is more duplication than the 22 context lines it removes, with silent wire-name changes on any miss. The gate above addresses the same drift hazard directly and needs no new dependency.

Also assessed and rejected: `CreateIReadOnlySetInfo` (all `IReadOnlySet` uses are method parameters, never serialized properties), `GetTypeInfo<T>`/`TryGetTypeInfo<T>` (contexts already thread `JsonTypeInfo<T>` explicitly), unions and `JsonTypeClassifier` (no polymorphic serialization; `UnionTypeInfo` is a *detected* .NET union, an unrelated feature), `Utf8JsonWriter.Reset` (no hot-loop writer reuse), and `SerializeAsyncEnumerable(topLevelValues: true)` (fits the three per-row JSONL sites, but they would have to go async and the primary JSONL path goes through Markout's table writer, not STJ).

Worth recording: a preview STJ package **would** have worked. Verified that a net10.0 app referencing `System.Text.Json 11.0.0-preview.6` loads app-local `System.Text.Json.dll` v11.0.0.0 on the .NET 10.0.8 runtime, with the net11-only APIs functioning and the source generator honoring them. The out-of-band package is not blocked by the LTS-floor TFM of the shipped "any" package. It simply is not needed.

## Validation

Release, per the repo's fixture-fidelity rule.

| Suite | Result |
| --- | --- |
| `dotnet-inspect.Tests` | 2238 passed, 0 failed, 14 skipped |
| `DotnetInspector.Services.Tests` | 325 passed, 0 failed |
| `ILInspector.Metadata.Tests` | 953 passed, 0 failed, 5 skipped |
| `ILInspector.Analysis.Tests` | 554 passed, 0 failed |
| `ILInspector.Decompiler.Tests --gate fast` | 4059 passed, 0 failed |

`markdownlint` clean on the changed doc. Suites re-run after merging `origin/main`.

Not yet ready to merge — adversarial review pending.

### DIFF
--- a/docs/design/untrusted-data-threat-model.md
+++ b/docs/design/untrusted-data-threat-model.md
@@ -254,6 +254,14 @@ only ordinary compiler output.
    attacker-authored content. Require agreement on the full resolved URL prefix
    including the commit, or derive provenance from the mapping the resolver
    actually selects, and report nothing when entries disagree.
+
+   Compare and extract on the *canonical* URL, not the literal mapping text.
+   `System.Uri` applies RFC 3986 dot-segment removal, so a value such as
+   `https://raw.githubusercontent.com/dotnet/runtime/<commit>/../../../owner/repo/<commit>/*`
+   is fetched from `owner/repo` while a regex over the raw string reports
+   `dotnet/runtime`. Canonicalize, or reject dot segments and their
+   percent-encoded equivalents, before both the comparison and the provenance
+   extraction.
 2. Fix GitHub repository provenance. The precondition tests the value for
    `github.com`, which canonical `raw.githubusercontent.com` SourceLink URLs do
    not contain, so GitHub-hosted assemblies report no repository at all. Match
--- a/docs/design/untrusted-data-threat-model.md
+++ b/docs/design/untrusted-data-threat-model.md
@@ -242,26 +242,48 @@ only ordinary compiler output.
 
 ## Open work
 
-1. Unify SourceLink provenance with source resolution. `AssemblyInspector`
+1. Unify SourceLink provenance with source resolution. Today `AssemblyInspector`
    reports the repository from the first `documents` entry while
    `SourceDocumentPathResolver` selects by longest matching pattern, so a
-   well-formed map with distinct keys can resolve source from one origin while
-   provenance names another. Requiring entries to agree on the *repository* is
-   not sufficient: the reader derives `owner/repo` and discards the commit, and
-   `raw.githubusercontent.com` serves any commit reachable in a repository,
-   including the head of an unmerged pull request. Two entries naming the same
-   repository at different commits would therefore "agree" while resolving
-   attacker-authored content. Require agreement on the full resolved URL prefix
-   including the commit, or derive provenance from the mapping the resolver
-   actually selects, and report nothing when entries disagree.
+   well-formed map can resolve source from one origin while provenance names
+   another.
 
-   Compare and extract on the *canonical* URL, not the literal mapping text.
-   `System.Uri` applies RFC 3986 dot-segment removal, so a value such as
-   `https://raw.githubusercontent.com/dotnet/runtime/<commit>/../../../owner/repo/<commit>/*`
-   is fetched from `owner/repo` while a regex over the raw string reports
-   `dotnet/runtime`. Canonicalize, or reject dot segments and their
-   percent-encoded equivalents, before both the comparison and the provenance
-   extraction.
+   State the fix as an invariant rather than a list of blocked tricks, because
+   each enumerated mitigation has proven incomplete under review:
+
+   > Reported provenance must describe the origin that source content is
+   > actually fetched from, for every document the assembly resolves. When that
+   > cannot be established for all of them, report no repository.
+
+   Establish it on the **final resolved URL, after wildcard substitution and
+   canonicalization** — not on the mapping text, and not on the mapping prefix
+   alone. Four concrete ways the weaker forms fail, all reproduced:
+
+   - Agreement on `owner/repo` ignores the commit, and
+     `raw.githubusercontent.com` serves any commit reachable in a repository,
+     including the head of an unmerged pull request. Two entries on one
+     repository at different commits "agree" while serving different code.
+   - `System.Uri` applies RFC 3986 dot-segment removal, so a mapping value
+     containing `../` is fetched from the traversed-to path while a regex over
+     the raw string reports the literal one.
+   - Even a clean mapping is not enough. The wildcard suffix comes from the PDB
+     document path, which is equally attacker-controlled, and
+     `EscapeSourceLinkPath` leaves `..` intact. A benign
+     `.../dotnet/runtime/<commit>/*` resolves
+     `/_/../../../attacker/evil/main/Program.cs` to a URL that canonicalizes
+     into `attacker/evil`.
+   - `System.Uri` preserves percent-encoded separators verbatim: `..%2f` and
+     `..%5c` survive canonicalization, so a "canonicalize, then prefix-check"
+     step passes while a server that percent-decodes before resolving dot
+     segments still traverses out. Reject encoded separators and encoded dot
+     segments rather than assuming canonicalization removed them.
+
+   Checking that the canonicalized resolved URL still lies beneath the
+   canonicalized mapping prefix addresses all four, but do not take that as
+   sufficient on its own: whatever check is implemented must ship with tests
+   covering at least these cases, since each was found only by attacking a
+   previous formulation of this item.
+
 2. Fix GitHub repository provenance. The precondition tests the value for
    `github.com`, which canonical `raw.githubusercontent.com` SourceLink URLs do
    not contain, so GitHub-hosted assemblies report no repository at all. Match
--- a/docs/design/untrusted-data-threat-model.md
+++ b/docs/design/untrusted-data-threat-model.md
@@ -92,6 +92,27 @@ cache entries use SHA-256-derived keys through `CoreCache`.
 Archive containment does not itself bound expanded bytes, entry count, or disk
 consumption. Resource budgets remain an open requirement below.
 
+### Untrusted JSON rejects duplicate properties
+
+JSON does not define how duplicate object keys resolve, so two readers of one payload can
+disagree. `DotnetInspector.Core.HardenedJson` (and its `ILInspector.Metadata.SourceLinkJson`
+counterpart, kept separate because Metadata sits below the Core infrastructure layer) parses with
+`AllowDuplicateProperties = false`, so such a payload fails visibly instead of binding one of
+several possible readings.
+
+This is not hypothetical for SourceLink. `SourceDocumentPathResolver` orders mappings by
+descending pattern length and takes the first match, while the repository-URL reader in
+`AssemblyInspector` takes the first value naming a known host. Given a repeated key under
+`documents`, those rules select different entries, so a hostile PDB could resolve source from one
+origin while the tool reported provenance from another. Rejecting the map closes that gap.
+
+Feed responses, package contents, `project.assets.json`, `.deps.json`, and product cache entries
+parse through the same guard. Callers that already treated malformed JSON as "no data" now treat
+duplicate-bearing JSON the same way; that is fail-closed, but it does not by itself convert those
+callers to explicit failure reporting, which remains open work below.
+
+`runfaster` still parses its trace inputs directly and is not yet covered.
+
 ### Artifact-derived source URLs use an SSRF-hardened client
 
 SourceLink and other artifact-derived fetches must use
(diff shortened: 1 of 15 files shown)


richlander:
## Review reconciliation (head `c246e334`)

Two adversarial reviews ran on `3da978a9` (GPT-5.6, Gemini 3.1 Pro). Both blocked on the same point, and **both were right**. Thanks — this changed the PR materially.

### Blocking: the SourceLink rationale was false

Gemini showed the duplicate-key hardening does not close the provenance gap, because the two readers diverge on **distinct** keys. GPT went further and challenged the PR's *original* justification as not reproducible at all.

I verified GPT's stronger claim against pre-hardening `origin/main` (`b07a3ba0`) rather than taking it on faith. It holds:

| Map | reported repo | resolved source |
| --- | --- | --- |
| identical duplicate keys | `<null>` | `https://evil.example/raw/Program.cs` (first entry) |
| distinct keys | `https://dev.azure.com/good/project` | `https://evil.example/raw/Program.cs` |

`AssemblyInspector.ExtractRepositoryUrlFromSourceLink` `break`s after the first `documents` entry, and `SourceDocumentPathResolver`'s `OrderByDescending` on pattern length is **stable**, so equal-length duplicates keep document order. Both readers land on the same first entry. **Duplication alone can never make them disagree.** The divergence needs distinct keys, which stay well-formed and accepted.

So the PR asserted a security property it did not deliver. Per AGENTS.md that is not acceptable regardless of the code being fine.

### Second bug (GPT)

`"raw.githubusercontent.com"` does not contain the substring `"github.com"` — it is `github` + `usercontent.com`. The precondition never fires, so **every GitHub-hosted assembly reports no repository**. Reproduced.

### What changed here

Scope decision: correct the claims, keep the hardening on its own merit, fix the provenance bugs in a focused follow-up. They are a selection/provenance concern, not a JSON-parsing one, and mixing them would violate the repo's "do not mix unrelated changes" rule.

- Rewrote the rationale in `SourceLinkJson.cs`, the threat model, and the `SourceLinkDuplicateDocumentTests` class doc. They now state plainly that duplication cannot make the readers disagree, that the real divergence is on distinct keys, and that it remains **open**.
- Filed #3391 (divergence) and #3392 (GitHub provenance), and recorded both as open-work items 1 and 2 in `docs/design/untrusted-data-threat-model.md`. Agreed fix for #3391: report no repository when entries disagree on origin — fail-closed and unspoofable.

### Non-blocking findings, all actioned

- **GPT** — `SourceLinkService.cs:201` still bound the last value for a duplicate key while the threat model claimed cache entries were covered. Added `[JsonSourceGenerationOptions(AllowDuplicateProperties = false)]` to `SourceLinkJsonContext`, plus two tests. Tamper-verified: removing the attribute fails `TypeFileIndexCache_RejectsDuplicateKeys`.
- **Gemini** — `HardenedJson.DocumentOptions` was a compile trap (`Utf8JsonReader` takes `JsonReaderOptions`). It was also unused, so it is now private rather than re-documented.
- **Gemini** — `KnownPascalCaseDeviations_StillDeviate` used `.Any(...)`, so a partial fix would still pass. It now pins the exact 18 deviating wire names, making the declaration drive the enforcement set. Tamper-verified: dropping one name from the pin fails the gate.

### Not actioned

- **GPT** — the gate discovers contexts from three anchor assemblies and would miss a future context in a fourth, and its checks validate lexical shape rather than the exact transformation. Both are real limitations; neither is a correctness defect in what it covers. Left as-is to keep this PR scoped.

### Validation (Release)

CLI 2238 passed / 0 failed / 14 skipped · Metadata 955 / 0 / 5 · Services 325 / 0 / 0 · `markdownlint` clean.

Re-review of `c246e334` is running with both models. Not marking ready until those come back clean.

richlander:
## Round-2 reconciliation (head `1b241d94`)

Both re-reviews of `c246e334` came back with one finding each. Both were real and both are fixed. Everything else in the round-1 fix set was independently confirmed by both models.

### GPT — `TryGetProperty` semantics stated backwards (Medium)

My rationale on `HardenedJson` said `JsonElement.TryGetProperty` returns the **first** duplicate. It returns the **last**. GPT's evidence was indirect, so I checked it directly:

```text
{"id":"first","other":1,"id":"second"}
  TryGetProperty  -> second
  GetProperty     -> second
  EnumerateObject -> first, second
  nested TryGet   -> B          (for {"o":{"id":"A","id":"B"}})
```

The correct framing is actually *sharper* than what I wrote: a name lookup resolves to the last occurrence while enumeration yields every occurrence in document order, so **a reader taking the first enumerated match already disagrees with a reader looking the name up** — no ordering trickery needed. Rewritten, and pinned by `PermissiveParsing_ResolvesANameLookupToTheLastDuplicate`, since I demonstrably got it backwards once.

### Gemini — the proposed follow-up fix was unsound (Critical)

The open-work item said to report no repository when entries disagree on **origin**. Gemini showed that is not enough: the reader derives `owner/repo` and **discards the commit**, and `raw.githubusercontent.com` serves any commit reachable in the repository, including an unmerged PR head. Verified:

```console
$ curl -o /dev/null -w "%{http_code}" https://raw.githubusercontent.com/richlander/dotnet-inspect/025d14ad56527f75cd6187c0f011243e57f9b887/README.md
200
```

Anyone who can open a PR against a public repo can host content under that repo's raw URL, so two entries naming the same `owner/repo` at different commits would "agree" while the resolver still fetched the attacker's commit. Open-work item 1 and #3391 now require agreement on the **full resolved URL prefix including the commit**, or derivation from the resolver-selected mapping.

Worth noting this finding was against a *proposed future fix* recorded in the threat model, not shipped code — which is exactly why writing the intended fix down was worth doing.

### Also addressed

GPT noted no test covered the SourceLink cache serialize→deserialize round trip through the newly hardened context. Added `TypeFileIndexCache_RoundTripsWhatItWrites`.

### Independently confirmed by both reviewers at `c246e334`

Corrected duplicate-key claims accurate · 18 pinned Timeline names exactly match `TimelineViews.cs`, gate fails on partial **and** full fixes, robust to unrelated compliant names · cache hardening non-vacuous, `BuildTypeFileIndex` cannot emit duplicates so no regression · `DocumentOptions` private breaks no caller · all ~20 migrated parse sites keep their `JsonException` handling, no legitimate duplicate-bearing payload found · history linear, no unrelated changes.

### Validation (Release, at `1b241d94`)

CLI 2238 passed / 0 failed / 14 skipped · Metadata 956 / 0 / 5 · Services 326 / 0 / 0 · `markdownlint` clean.

Round-3 review of `1b241d94` is running with both models.

richlander:
## Adversarial review reconciliation — rounds 3, 4, and 5

Rounds 1 and 2 were reconciled above. This covers the remaining three rounds, each run
with Gemini Pro and GPT on a fixed head in isolated review worktrees.

Every finding in these three rounds landed in the **prose of deferred open-work item 1**
(SourceLink provenance divergence, tracked as #3391). No round after round 1 found an
issue in shipped code: `HardenedJson.cs`, the 20 parse-site changes, and the 21 new tests
have been clean across four consecutive rounds.

That is not the same as the rounds being wasted. Each one falsified a claim I had written
and believed.

### Round 3 — head `9c74b540`

**Both models, independently: dot-segment normalization.** My proposed mitigation compared
repository identity extracted by regex from the raw mapping text. `System.Uri` applies
RFC 3986 dot-segment removal, so a mapping containing `../` is *fetched* from the
traversed-to path while the regex reports the literal one. Reproduced; recorded on #3391.

Two models reaching the same finding independently is worth more than either alone, and it
was the first signal that enumerating mitigations was the wrong shape for this item.

### Round 4 — head `9c74b540` → `6db2fb2b`

**GPT: a benign mapping is still unsafe.** I had assumed validating the mapping was the hard
part. The wildcard suffix comes from the PDB document path, which is equally
attacker-controlled, and `EscapeSourceLinkPath` leaves `..` intact:

```
mapping  : https://raw.githubusercontent.com/dotnet/runtime/<commit>/*   <- clean
document : /_/../../../attacker/evil/main/Program.cs
resolved : canonicalizes into attacker/evil
```

**Gemini: percent-encoded separators survive canonicalization.** `System.Uri` preserves
`..%2f` and `..%5c` verbatim, so a canonicalize-then-prefix-check passes while a server that
percent-decodes before resolving dot segments still traverses out.

Since rounds 1–4 had each defeated a *different enumerated mitigation*, I restructured the
item to lead with an invariant instead of a checklist — a checklist invites "you missed X"
indefinitely:

> Reported provenance must describe the origin that source content is actually fetched
> from, for every document the assembly resolves. When that cannot be established for all
> of them, report no repository.

I also deleted a sentence asserting that a particular check was sufficient. Per AGENTS.md,
an asserted safety property must name its enforcing gate or be marked unverified; that
sentence did neither, which is precisely the defect this PR set out to remove.

### Round 5 — head `6db2fb2b` → `2de26401`

Gemini cleared the PR. **GPT found that I had removed one sufficiency claim and left a
second**: the closing paragraph still said a canonicalized prefix-containment check
"addresses all four" counter-cases — refuted by the fourth counter-case stated three lines
above it, which is that exactly this check passes while traversal still occurs. My hedge
("do not take that as sufficient on its own") softened the claim without making it true.

Fixed in `2de26401`. The list is now framed as a regression floor rather than a
specification of what to block, and says plainly that passing it is not evidence the
invariant holds.

Both models re-reviewed the fixed head and confirmed: no residual sufficiency claim,
internally consistent with the counter-cases and the invariant, delta prose-only,
markdownlint exit 0, nothing blocking.

### Non-actions, stated explicitly

- **The provenance divergence itself is not fixed here** (#3391). This PR rejects duplicate
  JSON properties; divergence is reachable through *distinct* keys, so hardening does not
  close it. Round 1 corrected my original framing on this and the claim has been narrowed
  accordingly.
- **GitHub assemblies still report no repository** (#3392) — the precondition tests for the
  substring `github.com`, which canonical `raw.githubusercontent.com` URLs do not contain.
  Found in round 1, out of scope here.
- **`TimelineJsonContext` still emits PascalCase.** The new wire-name gate pins the exact 18
  deviating names rather than fixing them, so both a partial fix and a full fix fail the
  gate and force a deliberate decision. Changing them is a breaking wire change.
- The four counter-cases are deliberately duplicated between the threat-model doc and
  #3391, so the fix has them where it will be implemented and the doc explains why the item
  is shaped as an invariant.

richlander:
Ready to merge

Head `2de26401`. All CI green (`changes`, `build-net10`, `markdownlint`, `test`; smoke and
pack path-gated to skip). Five adversarial review rounds across Gemini Pro and GPT, each on
a fixed head in isolated worktrees; both models confirmed the final head clean with no
residual sufficiency claim.

Local validation at `1b241d94`, the last commit touching product code (Release): CLI
2238 passed / 0 failed / 14 skipped, Metadata 956/0/5, Services 326/0/0. Every commit since
is documentation-only.

Follow-up work is tracked and explicitly out of scope here: #3391 (SourceLink provenance
divergence, with all four reproduced counter-cases) and #3392 (GitHub assemblies report no
repository).

PR #3390, opened 2026-07-28
Fail loudly when a render path drops a payload projection

richlander:
Closes part of #3364.

## The recurring defect

Three PRs into this series, the same bug had appeared three times: a payload
projection (`--print`, `--value`, `--urls`, `--paths`, `--count`) is accepted by
the parser, then a render path forgets to dispatch it, renders the full
unprojected shape, and **exits 0**. Nothing catches it — not the exit code, not
the 2233-test suite. The caller gets the wrong payload and believes it succeeded.

#3373 fixed one instance. #3379 fixed another. This PR stops fixing instances
and closes the class.

## The guardrail

An accepted projection must end in one of two outcomes: the payload is
projected, or the command says why it cannot and exits non-zero. Rendering the
full shape is not a third option.

`ProjectionAudit` records the request from the `ParseResult` **by option name**,
so a command added later is covered without registering anything. The projection
writers report **which** projection they honored, and
`CommandLineBuilder.InvokeAsync` verifies — the single choke point shared by
`Program.cs` and the test harness, so tests cannot diverge from the product.

Design decisions worth stating:

- **Writers report which flag they honored, not merely that something was
  honored.** A writer can be reached for more than one reason — the print writer
  also serves `--bare` — so an untyped signal lets it satisfy an unrelated
  request and lets that drop escape. Both reviewers found this independently.
- **The projections are mutually exclusive**, now rejected centrally *before* the
  command runs. Two of them cannot both shape one payload, so any combination
  previously meant one was silently discarded. Four commands already enforced
  this among `--value`/`--urls`/`--paths`; enforcing it centrally covers
  `--print` and `--count` and the commands that never had it. The existing
  pairwise wording and command-line ordering are preserved, so the contract
  asserted by `LibraryCommand_IlOffsetCountRejectsPrint` is unchanged.
- **`AsyncLocal`, not static** — tests invoke the CLI concurrently in-process.
- **Non-zero exits are left alone.** Rejecting a projection is a legitimate way
  to not honor one, and the command already reported it.
- **`--help` is exempt at both entry points.** Help short-circuits rendering, so
  the projection is moot and the conflict has nothing to fight over.

## What this looks like in practice

Both outcomes below are the desired ones. The point of the PR is that there is
no third outcome — an accepted projection is never answered with the full
unprojected payload and exit 0.

### Projections are honored (exit 0)

A count is a count, whatever presentation flags accompany it. `--json` used to
win over `--count` here and return the whole result set:

```console
$ dotnet-inspect find '*' --library DotnetInspector.Core.dll --count --json
24
```

```console
$ dotnet-inspect depends 'Downloader`1' --library DotnetInspector.Core.dll --count --json
1
```

An empty section still counts, rather than reporting nothing:

```console
$ dotnet-inspect library DotnetInspector.Core.dll -S References --count
Note: 1 matched section has no data: References.
0
```

Bare-mode input routes to a real command and the projection survives the
rewrite — this whole class of invocation was previously unaudited:

```console
$ dotnet-inspect Regex -S Methods --count --tips q
Note: Type 'Regex' resolved via platform find to System.Text.RegularExpressions.Regex ...
69
```

Help still works when projections are present, because help renders no payload
and so has nothing to project:

```console
$ dotnet-inspect type --library DotnetInspector.Core.dll -S Classes --count --print --help
Description:
Discover types in a package or library (compact table output)
...
```

### Projections are refused (exit 1)

Two projections cannot both shape one payload, so the combination is rejected
before the command runs rather than resolved by discarding one:

```console
$ dotnet-inspect type --library DotnetInspector.Core.dll -S Classes --count --print
Error: --count cannot be combined with --print.
```

The same conflict is now caught through the bare-mode router, which previously
bypassed the check entirely and failed later with an unrelated message about
section selection:

```console
$ dotnet-inspect Regex --count --print --tips q
Error: --count cannot be combined with --print.
```

And where a render path genuinely drops a projection, it says so instead of
returning the wrong payload as success:

```console
$ dotnet-inspect type --library DotnetInspector.Core.dll -S Classes --value
# DotnetInspector.Core
...
| Type | Members |
...
Error: '--value' was accepted but this command path produced unprojected output. This is a bug in dotnet-inspect; please report it.
```

```console
$ dotnet-inspect package --count search Newtonsoft.Json --take 1
Newtonsoft.Json  13.0.4   8.8B   Json.NET is a popular high-performance JSON framework for ....
Error: '--count' was accepted but this command path produced unprojected output. This is a bug in dotnet-inspect; please report it.
```

That last message is deliberately blunt. It names a real defect, it is
addressed to the maintainer rather than the user, and each route that can
produce it has an issue and a fix queued — see "What still fires the audit"
below. The alternative is what these commands did before: answer a question
nobody asked, and call it success.

## What wiring it up found

Real bugs, none covered by the existing suite:

| Bug | Effect |
| --- | --- |
| `find`, `implements`, `extensions`, and type-mode `depends` resolved `--json` before `--count` | `--count --json` returned the full result set, exit 0 |
| `package -S Files --count` computed its count via `FormatResult` | correct count, but exit 1 |
| `timeline --count` wrote the count directly | correct count, but exit 1 |

Plus every bespoke `--count` emitter that bypassed `CountOutput` (17 tests'
worth), now converged onto a shared `CountOutput.WriteCount` — which also fixed
an ambient-locale vs `InvariantCulture` inconsistency.

### Ways the guardrail itself could be evaded

Each was found by review, and each would have made the guardrail
partial — which is worse than no guardrail, because it invites trust:

| Escape | Effect |
| --- | --- |
| Only the first requested flag was tracked, and honoring was untyped | a second projection could still be dropped, and `--bare` could satisfy a recorded `--count` |
| `/h` as the *value* of another option (`--type /h`) | audit disabled for the whole invocation |
| Projections declared by an ancestor (`package --count search <id>`) | accepted by the parser, invisible to the audit |
| The bare-mode router invoked its rewritten parse directly | **every** bare-mode invocation unguarded; `Newtonsoft.Json --count --print -S Grounding -n 1` returned the wrong payload with exit 0 |

The router fix yields the property that makes the guardrail total rather than
best-effort: **`ParseResult.InvokeAsync` is no longer called anywhere outside the
choke point.** Because invocations now nest, `BeginRequest` returns a scope that
restores the request it displaced, so the guarantee does not rest on the router
happening to record nothing of its own.

## What still fires the audit

Three routes render an unprojected payload and now fail loudly. All three were
already broken — they returned the wrong payload with exit 0 — so this is
existing breakage made visible, not new breakage. Each has a follow-up issue and
will be fixed in an immediate stacked PR:

| Route | Issue |
| --- | --- |
| Whole-surface `type` listing — no projection dispatch at all | part of #3364 |
| `library --il-offsets` — batch path ignores all four projections | #3395 |
| `package --versions` / `--layout` / `--tfms` — lens modes ignore `-S` and every projection | #3396 |
| `-D`/`--discover` on every command that supports it — discovery listing skips projection dispatch | #3398 |

Separately, `package` exposes projections its `search` subcommand cannot honor
(#3393). The audit makes that reachable case loud too.

Three of these four share one shape — an early-branching render path that skips
the section pipeline, and therefore skips projection dispatch. That is a useful
result on its own: it says the follow-up is one dispatch seam rather than a dozen
scattered fixes.

## A note on the `depends` instance

GPT reported this in the first review; I could not reproduce it and had planned
to dismiss it. The dismissal was wrong, and the reason generalizes: I probed
with a type that has **no dependencies**, which returns early through a separate
branch that already handles `--count`. Only a type with at least one dependency
reaches the ordering defect.

My whole sweep had the same flaw — empty result sets short-circuit before the
defective dispatch. The re-sweep uses deliberately non-empty results, and the
regression test asserts the fixture's count is greater than zero so it cannot
decay back into the branch that hid the bug.

## Evidence

- Sweep of 10 command shapes x 5 projection flags x 7 format flags with
  **non-empty** result sets: no false positive. A further sweep of 14 `package`
  sections x 3 lens modes x 5 projections isolated #3396 as a single structural
  defect rather than a per-section one.
- 15 untouched paths byte-identical to base `164c6ae1`, compared against a
  **fixed base-built input assembly**. Comparing against the repo's own live
  assemblies gives a false diff: they change as tests are added and embed the
  commit SHA.
- CLI 2253 pass / 0 fail / 14 skipped; services 319 pass / 0 fail; lint clean.

## Review

Adversarially reviewed by Gemini 3.1 Pro and GPT-5.6 in isolated worktrees over
four fixed-head rounds. Both independently found the same soundness hole in the
first head, which is why it was treated as decisive, and both independently
found the `depends` instance. Every round produced a distinct real escape, and
the last three of the six bugs above came from review rather than from my own
sweeps. Reconciliation is in a PR comment.

### DIFF
--- a/docs/design/output-shapes.md
+++ b/docs/design/output-shapes.md
@@ -200,6 +200,20 @@ could name a package. The explicit transition to that package artifact remains
 `package Package@version`. Likewise, printing a timeline may use only declared
 payloads on already evaluated rows; it cannot probe missing cells.
 
+### A payload projection is never silently dropped
+
+`--print`, `--value`, `--urls`, `--paths`, and `--count` reshape the payload, so
+a render path that ignores one answers a question the caller did not ask while
+still exiting 0. That failure is invisible to exit-code checks and to tests that
+only cover the unprojected path.
+
+Every accepted payload projection must therefore end in one of two outcomes: the
+payload is projected, or the command reports why it cannot be and exits non-zero.
+Rendering the full unprojected shape is not a third option. The requirement is
+enforced structurally rather than per command — the request is recorded from the
+parse result and the projection writers mark it honored, so a route that drops
+one fails loudly instead of shipping the wrong payload.
+
 ### Presentation modifiers (render the chosen shape)
 
 | Flag | Effect |
--- a/docs/design/output-shapes.md
+++ b/docs/design/output-shapes.md
@@ -211,8 +211,17 @@ Every accepted payload projection must therefore end in one of two outcomes: the
 payload is projected, or the command reports why it cannot be and exits non-zero.
 Rendering the full unprojected shape is not a third option. The requirement is
 enforced structurally rather than per command — the request is recorded from the
-parse result and the projection writers mark it honored, so a route that drops
-one fails loudly instead of shipping the wrong payload.
+parse result and the projection writers report which projection they honored, so
+a route that drops one fails loudly instead of shipping the wrong payload.
+
+Writers report *which* flag they honored rather than merely acknowledging one.
+A writer can be reached for more than one reason — the print writer also serves
+`--bare` — so an untyped signal would let it satisfy an unrelated request and let
+that drop escape.
+
+The projections are mutually exclusive. Two of them cannot both shape one
+payload, so a combination is rejected before the command runs rather than
+resolved by discarding one.
 
 ### Presentation modifiers (render the chosen shape)
 
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -3476,6 +3476,38 @@ public class CommandExecutionTests
     }
 
     [Fact]
+    public async Task Find_Count_ComposesWithJson()
+    {
+        // Found by the projection audit: --json was resolved before --count, so a count
+        // request was answered with the full unprojected result set and exit 0.
+        var (exit, output, _) = await RunAppAsync(
+            "find", "Cache", "--library", TestAssemblyPath, "--count", "--json");
+
+        Assert.Equal(0, exit);
+        Assert.True(int.TryParse(output.Trim(), out _), $"expected a bare count, got: {output}");
+    }
+
+    [Fact]
+    public async Task Implements_Count_ComposesWithJson()
+    {
+        var (exit, output, _) = await RunAppAsync(
+            "implements", "IDisposable", "--library", TestAssemblyPath, "--count", "--json");
+
+        Assert.Equal(0, exit);
+        Assert.True(int.TryParse(output.Trim(), out _), $"expected a bare count, got: {output}");
+    }
+
+    [Fact]
+    public async Task Extensions_Count_ComposesWithJson()
+    {
+        var (exit, output, _) = await RunAppAsync(
+            "extensions", "String", "--library", TestAssemblyPath, "--count", "--json");
+
+        Assert.Equal(0, exit);
+        Assert.True(int.TryParse(output.Trim(), out _), $"expected a bare count, got: {output}");
+    }
+
+    [Fact]
     public async Task Type_SourceFiles_Value_RowSelectsUrl()
     {
         var (exit, output, error) = await RunAppAsync(
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -3476,6 +3476,18 @@ public class CommandExecutionTests
     }
 
     [Fact]
+    public async Task Router_RewrittenCommand_IsAudited()
+    {
+        // The router captures projection flags as raw tokens, so the outer invocation records
+        // nothing. It used to invoke the rewritten parse directly, bypassing the audit, which
+        // left every bare-mode invocation unguarded. It now goes through the choke point.
+        var (exit, _, error) = await RunAppAsync("Regex", "--count", "--print", "--tips", "q");
+
+        Assert.Equal(1, exit);
+        Assert.Contains("--count cannot be combined with --print", error);
+    }
+
+    [Fact]
     public void ProjectionAudit_TracksProjectionDeclaredByAnAncestorCommand()
     {
         // `package --count search <id>` binds --count to the parent command, which the parser
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -3549,6 +3549,18 @@ public class CommandExecutionTests
     }
 
     [Fact]
+    public async Task ProjectionFlags_ConflictIsMootUnderHelp()
+    {
+        // Help renders no payload, so there is nothing for two projections to fight over.
+        // Rejecting the combination here would turn a working help request into an error.
+        var (exit, output, _) = await RunAppAsync(
+            "type", "--library", TestAssemblyPath, "-S", "Classes", "--count", "--print", "--help");
+
+        Assert.Equal(0, exit);
+        Assert.Contains("Usage", output, StringComparison.OrdinalIgnoreCase);
+    }
+
+    [Fact]
     public async Task Find_Count_ComposesWithJson()
     {
         // Found by the projection audit: --json was resolved before --count, so a count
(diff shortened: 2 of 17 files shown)


richlander:

Reviewed by **Gemini 3.1 Pro** and **GPT-5.6**, in isolated worktrees, over four
fixed-head rounds. Both are now clean on the audit itself at `20fd5699`.

Every round produced a real finding. That is worth stating plainly: a guardrail
that is partial is worse than none, because it invites trust it has not earned,
and the first three heads of this PR each had a way through.

### Blocking findings — all fixed

| # | Found by | Finding | Resolution |
| --- | --- | --- | --- |
| 1 | **Both, independently** | Only the first requested flag was tracked, and `MarkHonored()` was untyped, so a second projection could be dropped and `--bare` could satisfy a recorded `--count` | `9aeb6765` — track all flags, type the honoring, and reject projection combinations centrally |
| 2 | **Both, independently** | `depends` resolved `--json` before `--count` | `058ab82a` |
| 3 | GPT | `package -S Files --count` computed the count via `FormatResult` and never reported it honored | `9aeb6765` |
| 4 | GPT | `timeline --count` wrote the count directly | `9aeb6765` |
| 5 | GPT | `/h` as the *value* of another option disabled the audit for the whole invocation | `9aeb6765` — match option tokens, not raw text |
| 6 | GPT | `ValidateExclusive` ran ahead of the help exemption, so `--count --print --help` was rejected instead of printing help | `e3979567` |
| 7 | GPT | Projections declared by an ancestor (`package --count search <id>`) were invisible to the audit | `1bb52a1c` — walk the command chain |
| 8 | GPT, and independently by me | The bare-mode router invoked its rewritten parse directly, leaving **every** bare-mode invocation unguarded | `00100c35` |

Finding 1 was treated as decisive because two models found the same hole
independently without seeing each other's review. Finding 8 likewise: GPT's
repro was better than mine, because it exits 0 with a wrong payload rather than
merely producing a misleading message, so I adopted it.

`20fd5699` is mine, not a review finding: with the router fix, invocations nest,
and the audit's correctness rested on the accident that the router records
nothing of its own. `BeginRequest` now returns a scope that restores the request
it displaced. Both reviewers re-reviewed that change and found it sound.

### One finding I dismissed, wrongly

GPT reported the `depends` bug in round one. I could not reproduce it and had
planned to dismiss it. **The dismissal was wrong.** I probed `CoreCache`, which
has no dependencies, and that case returns early through a branch that already
handles `--count`; only a type with at least one dependency reaches the
defective dispatch. Gemini then reported the same bug from `System.String`,
which does.

The consequence was larger than the one bug: **my entire original sweep was
under-powered**, because empty result sets short-circuit before the code under
test. I re-swept with forced non-empty results, and the regression test asserts
the fixture's count is greater than zero so it cannot decay back into the branch
that hid the defect.

### Non-blocking — product gaps the audit exposed, deferred by owner's call

These are not defects in this PR. They are pre-existing bugs that previously
returned a wrong payload with exit 0, and the audit now makes them loud. Each
has an issue and an immediate stacked follow-up PR.

| Found by | Gap | Issue |
| --- | --- | --- |
| GPT | `package` exposes projections its `search` subcommand cannot honor | #3393 |
| **Gemini alone** | `library --il-offsets` drops all four projections | #3395 |
| Me, from a GPT example | `package --versions`/`--layout`/`--tfms` lens modes ignore `-S` and every projection | #3396 |
| **Gemini alone** | `-D`/`--discover` drops projections across every command that supports it | #3398 |

Gemini found the highest-yield search direction of the review: early-branching
render paths that skip the section pipeline. Three of the four gaps above share
exactly that shape, which means the follow-up is one dispatch seam rather than a
dozen scattered fixes.

### Nothing was dismissed without evidence

Every finding from both reviewers was reproduced and fixed or reproduced and
filed. There are no open disagreements and no findings resolved by argument
rather than by a real run.

### Validation at `20fd5699`

- CLI **2253 pass / 0 fail / 14 skipped**; services **319 pass / 0 fail**.
- **15 untouched paths byte-identical** to base `164c6ae1`, compared against a
  fixed base-built input assembly.
- Sweeps with forced non-empty results across command shapes, projection flags,
  and format flags: no false positives.

richlander:
Ready to merge.

Both fixed-head adversarial reviews are clean on the audit at `20fd5699`
(reconciliation above). CLI 2253 pass / 0 fail / 14 skipped; services 319 pass /
0 fail; 15 untouched paths byte-identical to base `164c6ae1`.

Stacked on #3385, so that merges first and this retargets to `main`. CI has not
run on this branch yet for the same reason; it will once the base is `main`.

The four product gaps the audit exposed (#3393, #3395, #3396, #3398) are
non-blocking follow-ups. Three share one dispatch seam and will be fixed
together in the next PR in the series.

Issue #3338, opened 2026-07-27
Mixed IL view drops preamble opcodes: give a constructor's implicit base() a home on the declaration line


richlander:
## What happens now

A constructor's implicit parameterless `base()` call has no C# form. The printer
walks the statement and emits nothing, so it owns no printed line — and the mixed
IL view drops the opcodes that belong to it:

```csharp
public PrintedRangeChainFixture(string name)
{
    this.Name = name;
        // IL_0006: ldarg.0
        // IL_0007: ldarg.1
        // IL_0008: stfld        string ...::<Name>k__BackingField
}
```

`IL_0000: ldarg.0` and `IL_0001: call System.Object::.ctor()` appear nowhere in
the interleave. The `IL` section still lists them, so nothing is hidden outright,
but the mixed view — whose whole claim is "here is the IL each line compiles to"
— silently omits two opcodes.

Before #3328 phase 1.5 they were attributed to `this.Name = name;`, a line that
does not compile to a base call. That was worse: a false claim rather than an
absent one. But "not shown" is not the right end state either.

## The better home

The implicit base call is part of the constructor's *preamble*, not of any body
statement. The declaration line is where it belongs:

```csharp
public PrintedRangeChainFixture(string name)
    // IL_0000: ldarg.0
    // IL_0001: call         System.Object::.ctor()
{
    this.Name = name;
        // IL_0006: ldarg.0
        ...
```

That is a true statement about the code, and it is the only line in the render
that can carry it.

## Why it is not a small change

The correlation layer only ever sees the printed **body**. The declaration line
is added downstream by `FormatSourceWithDeclaration`, after
`CorrelateMixedSource` has already bucketed IL onto body lines. So there is no
line in the `AnnotatedSourceLine` stream to attach preamble IL to.

Making this work means the stream has to start at the declaration rather than at
the body's first statement — which is a change to what the correlation layer
considers its subject, not a lookup tweak.

## Scope beyond the base call

This is worth doing for a second reason: unanchorable IL is not rare. Measured
over `ILInspector.Decompiler.dll`:

| | |
| --- | --- |
| IL lines total | 266,961 |
| IL lines with no C# line to sit beside | 4,553 |
| Methods affected | 2,391 |

Roughly a quarter of methods omit at least one opcode from the mixed view. The
implicit base call accounts for 1,780 of those lines; the rest have other causes
worth their own triage. A declaration-line home fixes the preamble class of them
[...]

Issue #3328, opened 2026-07-27
Printer-emitted character ranges: give annotations sub-statement caret placement


richlander:
## Problem

Annotations can only ever underline a **whole statement**. `AnnotationCaret.Render`
(`src/ILInspector.Decompiler/Annotations/AnnotationCaret.cs:94`) hardcodes
`.Append('^', trimmed.Length)`, because nothing in the pipeline can tell it *which characters*
of the printed line the fact is about.

That is a correlation gap, not a data gap. The facts are already precisely located:
`AllocationOccurrence` (`src/ILInspector.Analysis/AllocationOccurrence.cs:103`) carries an exact
`int ILOffset` for each `newobj` / `box` / `newarr`. Precision is lost only in the projection
from IL to *text*.

The missing link is narrow and mechanical. `IrNode.SourceOffset`
(`src/ILInspector.Decompiler/Pipeline/Ir/IrNode.cs:29`) already survives the pass pipeline —
stamped at import, copied by `Clone`, adopted via `InheritSourceOffset` — and its own doc comment
states the intent verbatim:

> This is the bridge that lets a single annotation set key both the IL view (by offset) and the
> C# view (via the node that printed a line).

And the printer already emits provenance:
`CSharpPrinter.PrintRaised(IrFunction, out IReadOnlyDictionary<IrNode, int> statementLines)`
(`:187`), populated in `AppendStatement` (`:1697-1703`).

So the gap is two dimensions at one existing seam:

- **Granularity** — recorded only for statements. A `newobj` is an *expression* node nested inside.
- **Resolution** — the recorded value is a line index, not a character range.

## Why the printer is the only honest source

Three routes to a character range exist. Only one is sound.

1. **Re-parse the rendered C# and locate syntax.** This is what `SpanAttribution`
   (`tools/DecompilerHarness/SpanAttribution.cs:31`) does, and its header explicitly declines to
   push a body-span API into the product printer. It guesses when two `new` expressions share a line.
2. **Text-search for a detail token.** Worse.
3. **Printer-emitted provenance.** The printer is the one component that knows which characters it
   emitted for a given IR node. Once the string exists, that information is unrecoverable.

A precise-but-wrong caret is worse than a coarse-but-true one in a teaching tool, so guessing is
not an acceptable fallback. Route 3 or nothing.

## Coordinate stability is a non-issue for carets

Worth recording, because it looked like a blocker and is not. `RenderRaisedOverlay`
[...]

PR #3399, opened 2026-07-28
Record printed character ranges in the C# printer (#3328)

richlander:
Phase 1 of #3328, plus the two corrections that fell out of verifying it.

Annotations can currently only underline a *whole statement*, because nothing in
the pipeline records which characters of a printed line a fact refers to. The
progression below is the point of the work; all three views are of the same
method, `CommandCaretGestureFixture.Pump`.

### Where it started: facts as trailing comments

Every fact rides on the end of the line it belongs to. Nothing is claimed about
*where* in the statement the fact lives, so nothing is wrong — but a long fact
pushes the code off screen, and two facts on one line have to share.

```csharp
    List<object> sink = new();  // alloc.new(List<object>; alloc=System.Collections.Generic.List<System.Object>; path=straight-line; path-confidence=dominates-return; multiplicity=once; churned=System.Object[])
        // IL_0000: newobj       List<object>::.ctor()
        // IL_0005: stloc.0
    ...
        sink.Add(new object());  // alloc.new(object; alloc=System.Object; path=loop-body; path-confidence=behind-branch; multiplicity=loop)
            // IL_000A: ldloc.0
            // IL_000B: newobj       object::.ctor()
            // IL_0010: callvirt     List<object>::Add(object)
```

### Today: the caret exists, but points at the whole statement

`--focus allocation` promotes the fact to a caret gesture. The fact now has room
to wrap and read properly. But the caret spans the entire statement, which
asserts something false — and does so directly above the IL that shows the
truth:

```csharp
    List<object> sink = new();
//  ^^^^^^^^^^^^^^^^^^^^^^^^^^ alloc.new(List<object>;
//                             alloc=System.Collections.Generic.List<System.Object>;
//                             path=straight-line; path-confidence=dominates-return;
//                             multiplicity=once; churned=System.Object[])
        // IL_0000: newobj       List<object>::.ctor()
        // IL_0005: stloc.0
    ...
        sink.Add(new object());
//      ^^^^^^^^^^^^^^^^^^^^^^^ alloc.new(object; alloc=System.Object; path=loop-body;
//                              path-confidence=behind-branch; multiplicity=loop)
            // IL_000A: ldloc.0
            // IL_000B: newobj       object::.ctor()
            // IL_0010: callvirt     List<object>::Add(object)
```

The `List<object> sink =` part does not allocate; `new()` does. `sink.Add` does not
allocate; `new object()` does. In both cases the caret underlines the local, the
receiver and the method call and claims they allocate, and the reader can see
`newobj` sitting on its own line immediately below.

Both blocks above are real output. Abridged in two ways: the loop header and
its IL are elided (`...`), and the trailing `// alloc...` / `stack:` annotations
that each IL line also carries are trimmed. The C# lines, carets and fact text
are verbatim.

### The target — **not in this PR**

```csharp
    List<object> sink = new();
//                      ^^^^^ alloc.new(List<object>; ...)
        // IL_0000: newobj       List<object>::.ctor()
    ...
        sink.Add(new object());
//               ^^^^^^^^^^^^ alloc.new(object; ...)
            // IL_000B: newobj       object::.ctor()
```

This block is a mockup, not captured output. Reaching it needs three more
phases: carry the range to `AnnotatedSourceLine` (the caret renderer currently
receives no map, no offset and no column, so a range cannot reach it at all),
record expression nodes and not just statements, then select the narrowest node
covering the fact's IL offset and fall back to the whole statement rather than
guess.

### Why the printer has to be the one to do it

The precision was never missing. `AllocationOccurrence` carries an exact
`ILOffset`, and `AnnotationAnchor.Best` already resolves that offset to the
owning IR node. What is missing is the last hop: which *characters* of the
printed line that node produced. Only the printer knows, and once the string
exists the information is unrecoverable — the repo's own harness
(`tools/DecompilerHarness/SpanAttribution.cs`) re-parses decompiled C# with
Roslyn to recover even a coarse body span.

This PR captures it at the one place it is free: the builder's length either
side of a node's emission. It does **not** narrow any caret.

## What is here

**`PrintedRangeMap`** replaces the statement-line dictionary. A lookup that also
enumerates, because both patterns are real: annotation anchoring walks the parent
chain doing keyed lookups, the correlation seams enumerate to build line tables.
Lines became a projection of the range rather than a separate recording, which
removes a per-statement rescan of the whole builder — not measurable today,
but it
is what would have gone quadratic once expression nodes are recorded.

Ranges are captured by wrapping the emission body rather than recording inside it,
which is what makes the end offset correct across every exit of a 340-line method.

**Two invariants, arrived at by breaking things:**

*Every range is non-empty.* A node the printer walks but which emits nothing was
being recorded with a zero-width range at whatever position emission had
reached —
reading as "this node printed here" while resolving to the line of the *next*
statement. 890 of these across 9,114 methods, every one an implicit parameterless
`base()`. Dropping them lets a consumer slice or place a caret without a width
guard.

*A silent node keeps its position.* Dropping the range took something else
with it:
"where does this node sit in emission order" is perfectly well defined at
offset 0,
and the mixed IL view was the consumer relying on it. Its opcodes lost their only
bucket and vanished from the interleave. So the two questions are now separate
members — `TryGetRange`/`TryGetLine` for printed characters, `TryGetInsertionLine`
for emission order — and the channels are disjoint, asserted by a test. A caret
still cannot reach a node that printed nothing.

Only the IL path takes the insertion-point fallback. An annotation whose owner is
silent stays unplaced rather than claiming a line it did not print: a fact asserts
something about the line it sits on, an opcode only has to render in the right
order.

## This ends up ahead of where it started

Placing against an insertion point renders preamble IL *above* the line — for a
constructor, above the first statement of the body, which is when it runs.
`main` rendered it *below* that statement, simultaneously claiming the statement
executed the base call and showing it executing afterward:

```csharp
public Constant(object? value, TypeRef type)
{
        // IL_0000: ldarg.0
        // IL_0001: call         IrExpression::.ctor()
    this.Value = value;
        // IL_0006: ldarg.0
        // IL_0008: stfld        object Constant::<Value>k__BackingField
}
```

Where the printer hoists synthesized local declarations to the top of a block,
the insertion point falls below them rather than immediately after the opening
brace. That is what an insertion point *is* — the declarations would have printed
first — and they emit no IL of their own, so nothing executable is ever shown as
running before the base call. Called out because an earlier draft of this
description claimed "immediately after the opening brace", which is only true for
bodies with no hoisted declarations.

## Verified rather than assumed

| | |
| --- | --- |
| Renders vs `main`, 9,195 methods | 900 changed, **0 by content** (ordering + 609 stray blanks removed) |
| IL lines with no C# line to sit beside | 4,596 → **2,796** |
| Facts with no printed line | 56 → 56 — none lost |
| Range overruns of the returned string | 0 of 62,838, so the defensive clamp was dead and is deleted rather than shipped unproven |
| Print time, best-of-6 over 9,122 methods | 5,236 ms → 5,205 ms — below per-round variance, so "no change", not a win |

**Gates proven non-vacuous by breaking what they cover:** an off-by-one start fails
the text-alignment test; a zero-width range fails containment; removing the
insertion-point fallback fails the preamble test; bucketing after the line instead
of before fails it; recording insertion points for printed nodes fails disjointness.

Two of my own gates were vacuous when first written and were rebuilt — the
suppressed-chain fixture originally used an explicit `: base(seed)`, which is the
*lifted* case that is never walked at all.

## Suites

`ILInspector.Decompiler.Tests` 4,261 · `ILInspector.Research.Tests` 132 ·
`dotnet-inspect.Tests` 1,638 · `CommandExecutionTests` 595.

Seven decompiler failures reproduce on the unmodified base commit with identical
messages — three generated-fixture catalogs, three fidelity gates, and the pinned
missing-return list. None touch this code.

## Follow-ups

Issue #3338 keeps the remaining 2,796 unanchorable IL lines and the
declaration line as
the more precise home for preamble IL. The 56 homeless facts are field-initializer
stores lifted onto declarations — a separate gap.

Phases 2-4 (carry the range to `AnnotatedSourceLine`, record expression nodes,
narrow the caret) are what actually fix the render at the top of this description.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/MixedSourceRendererTests.cs
+++ b/src/ILInspector.Decompiler.Tests/MixedSourceRendererTests.cs
@@ -247,3 +247,65 @@ public class MixedSourceRendererTests
         Assert.Equal(left.GetHashCode(), right.GetHashCode());
     }
 }
+
+/// <summary>
+/// A derived constructor whose implicit <c>base()</c> targets a user-defined
+/// base. The chain call prints nothing, so it owns no characters, but its two
+/// opcodes still run — before the first statement. See
+/// <see cref="MixedPreambleTests"/>.
+/// </summary>
+public class MixedPreambleBase
+{
+    protected MixedPreambleBase() => Created = true;
+
+    public bool Created { get; }
+}
+
+public sealed class MixedPreambleDerived : MixedPreambleBase
+{
+    public MixedPreambleDerived(string name) => Name = name;
+
+    public string Name { get; }
+}
+
+public class MixedPreambleTests
+{
+    static string Render(Type type)
+    {
+        var source = MetadataSource.Open(type.Assembly.Location);
+        var result = ResearchViews.ProjectMember(new ResearchViews.MemberProjectionRequest(
+            source, type.FullName!, ".ctor", AnnotatedSource: true)).AnnotatedSource;
+        Assert.NotNull(result?.Output);
+        return result!.Output!;
+    }
+
+    [Fact]
+    public void ImplicitBaseCall_RendersAboveTheStatementItRunsBefore()
+    {
+        // The implicit base() prints no C#, so it has no range and no line, and
+        // the interleave reaches it only through its insertion point. Two things
+        // have to hold and both have been wrong before: the opcodes must appear
+        // at all (dropping the zero-width range once removed them entirely), and
+        // they must appear *above* the first statement rather than beneath it
+        // (bucketing them onto the next statement's line once claimed that
+        // statement executed them).
+        var lines = Render(typeof(MixedPreambleDerived)).Split('\n');
+
+        int baseCall = Array.FindIndex(lines, l => l.Contains($"{nameof(MixedPreambleBase)}::.ctor()"));
+        int firstStatement = Array.FindIndex(lines, l => l.Trim().StartsWith("this.Name = name;"));
+
+        Assert.True(baseCall >= 0, $"base call IL is missing from:\n{string.Join('\n', lines)}");
+        Assert.True(firstStatement >= 0, "the constructor body should print its field store");
+        Assert.True(
+            baseCall < firstStatement,
+            $"base call IL should precede the statement it runs before:\n{string.Join('\n', lines)}");
+        Assert.StartsWith("//", lines[baseCall].TrimStart());
+
+        // The invariant is "before the first statement", not "immediately after
+        // the opening brace". An insertion point is where the node's text would
+        // have gone, and the printer may hoist synthesized local declarations to
+        // the top of the block first, in which case the preamble lands below
+        // them. Those declarations emit no IL of their own, so nothing
+        // executable is ever shown as running before the base call.
+    }
+}
--- a/src/ILInspector.Decompiler.Tests/PrintedRangeMapTests.cs
+++ b/src/ILInspector.Decompiler.Tests/PrintedRangeMapTests.cs
@@ -10,13 +10,24 @@ namespace ILInspector.Decompiler.Tests;
 public class PrintedRangeMapTests
 {
     static (string Output, PrintedRangeMap Ranges) Print(string methodName)
+        => Print(typeof(AllocSampleClass), methodName);
+
+    static (string Output, PrintedRangeMap Ranges) Print(Type declaringType, string methodName)
     {
-        var source = MetadataSource.Open(typeof(AllocSampleClass).Assembly.Location);
-        var function = IrImporter.Import(source, typeof(AllocSampleClass).FullName!, methodName);
+        var source = MetadataSource.Open(declaringType.Assembly.Location);
+        var function = IrImporter.Import(source, declaringType.FullName!, methodName);
         Assert.NotNull(function);
         var result = CSharpPrinter.PrintRaised(function!, out var ranges);
         Assert.NotNull(result.Output);
         return (result.Output!, ranges);
+    }
+
+    static IrFunction Import(Type declaringType, string methodName)
+    {
+        var source = MetadataSource.Open(declaringType.Assembly.Location);
+        var function = IrImporter.Import(source, declaringType.FullName!, methodName);
+        Assert.NotNull(function);
+        return function!;
     }
 
     static (int Start, int End) Offsets(PrintedRange range, int length)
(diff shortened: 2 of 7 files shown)


richlander:
Two adversarial reviews are in. One clean, one with two Medium findings. Both
findings reproduce; neither is a regression, and one was a real defect in my
description rather than in the code.

## Finding 1 — insertion point lands after hoisted local declarations

Valid, and the reviewer is right that `MixedPreambleDerived` misses the shape.
Reproduced on `ILInspector.Research.ResearchChange..ctor`:

```csharp
{
    bool V_1;
    ResearchChange S_0;

        // IL_0000: ldarg.0
        // IL_0001: call         object::.ctor()
    ResearchSubjectKey S_256 = subject;
```

I am not changing the behaviour. An insertion point is *the offset the node's
text would have started at*, the declarations really would have printed first,
and they emit no IL of their own — so nothing executable is ever shown as
running before the base call. It still lands above the first statement, which is
the claim that matters.

What was wrong is that my commit message, this PR body and the #3338 comment all
said "immediately after the opening brace", which is only true for bodies with no
hoisted declarations. All three are corrected, and the invariant the test
actually holds — *before the first statement*, not *immediately after the brace*
— is now written down at the test.

I tried to add a locals-bearing fixture and could not construct one small enough
to keep in the suite that reproduces hoisting; the fixture I wrote inlined its
declarations instead. I also drafted assertions for "no other content precedes
the preamble IL" and dropped them: for this fixture every line between the first
IL and the base call is IL by construction, so they are tautological and the
only mutation that breaks them already breaks the existing
`baseCall < firstStatement` gate. I would rather ship a comment I can defend
than an assertion I cannot break.

## Finding 2 — IL renders out of offset order

Reproduces, and is pre-existing on `main`. Rendering the same member with the
`main` renderer and the branch renderer over identical input gives byte-identical
opcode ordering:

```
main:                          branch:
  S_256 = subject;               S_256 = subject;
    // IL_0007 ldarg.1             // IL_0007 ldarg.1
    // IL_0008 dup                 // IL_0008 dup
  S_0 = this;                    S_0 = this;
    // IL_0006 ldarg.0             // IL_0006 ldarg.0
```

The cause is upstream: the raiser prints `S_0 = this;` after
`S_256 = subject;`, and bucketing follows printed-line order. This PR only
changes where *silent* nodes' IL goes; both statements here print. Filed as
#3403.

## Re-verified after the change

`MixedPreambleTests` + `PrintedRangeMapTests` — 14 tests, 0 failures. The only
code change since the reviews is to a comment; the amended commit contains no
behavioural difference.

richlander:
Both findings are closed by the reviewer that raised them, after it reproduced
each independently rather than accepting my account.

**Finding 1 — insertion point after hoisted declarations.** Closed as correct
behaviour. The reviewer traced it: `CollectDeclarations` emits the synthetic
declarations before `AppendContainer`, the silent node is walked afterward, so
the recorded offset accurately reflects printer emission order — and no original
IL is attributed to those declarations. Its conclusion on the alternative:
moving the insertion point above the declaration block "would create a
constructor-specific exception rather than record where the node was visited."
It also independently tried and failed to build a small fixture that forces
hoisting, and agreed that not adding a vacuous one was right. The defect was my
prose, and that is corrected.

**Finding 2 — IL out of offset order.** Closed as out of scope. The reviewer ran
both harnesses itself and confirmed `main` and this branch produce the same
`IL_0007, IL_0008, IL_0006` sequence; the reordered instructions all take the
pre-existing `ilByLine` path and never touch the new insertion-point channel.
Tracked as #3403.

Reviewer's check of the amended commit: "The only code delta from the previously
reviewed HEAD is the explanatory test comment. Build, targeted tests, and
`diff --check` pass. No new significant issues found."

Two independent adversarial reviews, both clean. The second reviewer covered the
same seven attack areas and found no issues; I did not take that at face value
either, and verified its two load-bearing claims empirically over the whole
assembly — 39,980 methods, 93,668 ranges, 0 channel overlaps, 0 range overruns,
0 out-of-bounds lines.

PR #3402, opened 2026-07-28
Match SourceLink document keys by prefix, not by regex

richlander:
Stacked on #3367. GitHub will retarget this to `main` when #3367 merges.

## What is wrong

`SourceLinkResolver.ResolveUrl` built a regex from a SourceLink document key and matched with **no timeout**:

```csharp
string regexPattern = "^" + Regex.Escape(pattern).Replace("\\*", "(.*)") + "$";
var match = Regex.Match(filePath, regexPattern);
```

Those keys come from the SourceLink JSON embedded in the PDB of the assembly being inspected, so **an attacker chooses them** — squarely inside `docs/design/untrusted-data-threat-model.md`. Every `*` becomes another adjacent greedy group, and adjacent greedy groups over a separator-rich path backtrack exponentially.

Measured on the pre-fix code, matching a path of only 62 characters:

| wildcards | 5 | 7 | 9 | 11 | 12 |
| --- | --- | --- | --- | --- | --- |
| match time (ms) | 12 | 253 | 2,109 | 6,575 | **14,298** |

Roughly doubling per added wildcard, from a key inside a PDB. Real source paths are far longer than 62 characters.

Found by a Gemini Pro adversarial review of #3367. **Pre-existing** — it shipped in the `SourceLinkFetch` 0.1.0 package that `main` already consumes, so this is not a regression from adopting the source. Adopting the source is what makes it fixable.

## The fix removes the regex rather than bounding it

The SourceLink specification allows a key **at most one `*`**, and requires that `*` to be the **final character**. A conformant key is therefore a prefix, and the match is a prefix test — no regex, no backtracking, and the vulnerability becomes unrepresentable rather than merely timed out.

Testing the *first* `*` for finality rejects both violations at once: a second `*` necessarily leaves the first one non-final.

## Evidence that rejecting non-conformant keys is safe

I surveyed every Portable PDB in the local `dotnet-inspect` cache:

```
pdb files scanned : 1626
with SourceLink   : 1452
document keys     : 1584
--- key shapes ---
  1584  one '*', final char (spec-conformant)
```

**1,584 keys, 100% conformant, no exceptions.** No real producer emits a key this rejects. Rejecting one is also the only safe reading — a key we cannot honor unambiguously is exactly the key that hangs us.

## The new test is not vacuous

`AKeyWithManyWildcards_ResolvesPromptly_AndDoesNotBacktrack` asserts a loose 1-second bound. On its exact inputs:

```
AFTER  (prefix match) : 1 ms, result=null
BEFORE (regex)        : STILL RUNNING after 30s -- aborted by an artificial
                        timeout the product code did NOT have
```

Three orders of magnitude of headroom, so it fails on exponential behavior and never on a slow machine.

11 tests added covering suffix mapping, empty suffix, prefix misses, exact keys, backslash normalization, the three non-conformant key shapes, non-conformant keys not shadowing conformant ones, the denial-of-service bound, and the real-world key shape the PDB survey found.

## Validation

| Suite | Result |
| --- | --- |
| Metadata | **1233** pass (1222 + 11 new) |
| Services | **319** pass |
| CLI | **2233** pass |

## Behavior note

A key with `*` in a non-final position previously matched via regex and is now ignored. That shape is non-conformant, absent from all 1,584 surveyed keys, and is the shape that causes the hang. Called out explicitly rather than buried.

Adjacent and **not** addressed here: `src/ILInspector.Metadata/TypeMatcher.cs:367` uses the same `Regex.Escape(p).Replace("\\*", ".*")` shape with no timeout. Narrower vector — the pattern is the user's own CLI glob, though the matched text is an attacker-supplied type name.

### DIFF
--- a/src/SourceLinkFetch/SourceLinkFetch.csproj
+++ b/src/SourceLinkFetch/SourceLinkFetch.csproj
@@ -13,4 +13,8 @@
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
   </PropertyGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="ILInspector.Metadata.Tests" />
+  </ItemGroup>
+
 </Project>
--- a/src/SourceLinkFetch/SourceLinkResolver.cs
+++ b/src/SourceLinkFetch/SourceLinkResolver.cs
@@ -14,7 +14,7 @@ public class SourceLinkResolver
 
     private readonly Dictionary<string, string> _documentMappings;
 
-    private SourceLinkResolver(Dictionary<string, string> documentMappings)
+    internal SourceLinkResolver(Dictionary<string, string> documentMappings)
     {
         _documentMappings = documentMappings;
     }
@@ -45,16 +45,20 @@ public class SourceLinkResolver
 
         foreach (var (pattern, urlTemplate) in _documentMappings)
         {
-            if (pattern.Contains('*'))
+            int star = pattern.IndexOf('*');
+            if (star >= 0)
             {
-                string regexPattern = "^" + Regex.Escape(pattern).Replace("\\*", "(.*)") + "$";
-                var match = Regex.Match(filePath, regexPattern);
+                // SourceLink allows a key at most one '*', and requires it to be the final
+                // character, so a conformant key is a prefix and the match is a prefix test.
+                // Testing the first '*' for finality rejects both violations at once: a second
+                // '*' leaves the first one non-final. A key we reject is one no conformant
+                // producer emits and one we could not honor unambiguously anyway.
+                if (star != pattern.Length - 1)
+                    continue;
 
-                if (match.Success && match.Groups.Count > 1)
-                {
-                    string captured = match.Groups[1].Value;
-                    return urlTemplate.Replace("*", captured);
-                }
+                string prefix = pattern[..^1];
+                if (filePath.StartsWith(prefix, StringComparison.Ordinal))
+                    return urlTemplate.Replace("*", filePath[prefix.Length..]);
             }
             else if (filePath == pattern)
             {
--- a/tests/ILInspector.Metadata.Tests/SourceLinkUrlResolutionTests.cs
+++ b/tests/ILInspector.Metadata.Tests/SourceLinkUrlResolutionTests.cs
@@ -1,0 +1,130 @@
+using System.Diagnostics;
+using SLF = SourceLinkFetch;
+
+namespace ILInspector.Metadata.Tests;
+
+/// <summary>
+/// Covers <see cref="SLF.SourceLinkResolver.ResolveUrl"/>, which maps a PDB document path to a
+/// source URL using the SourceLink document keys embedded in that PDB. Those keys are
+/// attacker-controlled: they arrive inside the assembly under inspection, which the user does
+/// not necessarily trust (see docs/design/untrusted-data-threat-model.md).
+/// </summary>
+public class SourceLinkUrlResolutionTests
+{
+    private static SLF.SourceLinkResolver For(params (string Pattern, string Url)[] mappings)
+        => new SLF.SourceLinkResolver(mappings.ToDictionary(m => m.Pattern, m => m.Url));
+
+    [Fact]
+    public void ATrailingWildcard_MapsThePathSuffixIntoTheUrl()
+    {
+        var resolver = For(("/_/src/*", "https://raw.githubusercontent.com/o/r/abc/*"));
+
+        Assert.Equal(
+            "https://raw.githubusercontent.com/o/r/abc/System/Text/Json/Utf8JsonReader.cs",
+            resolver.ResolveUrl("/_/src/System/Text/Json/Utf8JsonReader.cs"));
+    }
+
+    [Fact]
+    public void AWildcardKey_MatchesTheEmptySuffix()
+    {
+        var resolver = For(("/_/*", "https://example.test/*"));
+
+        Assert.Equal("https://example.test/", resolver.ResolveUrl("/_/"));
+    }
+
+    [Fact]
+    public void AWildcardKey_DoesNotMatchAPathOutsideItsPrefix()
+    {
+        var resolver = For(("/_/src/*", "https://example.test/*"));
+
+        Assert.Null(resolver.ResolveUrl("/other/src/File.cs"));
+    }
+
+    [Fact]
+    public void AKeyWithoutAWildcard_MatchesOnlyThatExactPath()
+    {
+        var resolver = For(("/_/src/File.cs", "https://example.test/File.cs"));
+
+        Assert.Equal("https://example.test/File.cs", resolver.ResolveUrl("/_/src/File.cs"));
+        Assert.Null(resolver.ResolveUrl("/_/src/File.cs.bak"));
+    }
+
+    [Fact]
+    public void ABackslashPath_IsNormalizedBeforeMatching()
+    {
+        var resolver = For((@"/_/src/*", "https://example.test/*"));
+
+        Assert.Equal("https://example.test/a/b.cs", resolver.ResolveUrl(@"\_\src\a\b.cs"));
+    }
+
+    /// <summary>
+    /// SourceLink requires a key to carry at most one '*' and requires that '*' to be the final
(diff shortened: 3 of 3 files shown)


richlander:
## Independent corroboration: this repo already had a second, safe implementation

While tracing where the resolved URL goes, I found that `ILInspector.Metadata` carries its **own** live implementation of SourceLink document mapping — `SourceDocumentPathResolver.Resolve` in `src/ILInspector.Metadata/SourceDocumentPath.cs:64`, wired up from `PdbContext.cs:323` and `:1109`.

It already does exactly what this PR changes `SourceLinkFetch` to do:

```csharp
if (pattern.EndsWith('*'))
{
    string prefix = pattern[..^1];
    if (normalizedPath.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
```

**Prefix match, no regex, no backtracking** — written independently, and it is not vulnerable. That is strong corroboration that "one trailing `*` means a prefix match" is the correct reading of the spec, and it means the fix aligns the two implementations rather than inventing a new interpretation.

It also makes the acquisition-path duplication concrete for a third time in this area.

### The two implementations still disagree — follow-up, not this PR

| Behavior | `SourceDocumentPath` (Metadata) | `SourceLinkFetch` (after this PR) |
| --- | --- | --- |
| Path comparison | `OrdinalIgnoreCase` | `Ordinal` |
| Suffix | `.TrimStart('/')` | no trim |
| URL construction | `Uri.EscapeDataString` per segment | raw `urlTemplate.Replace("*", suffix)` |

I deliberately kept `Ordinal` here: the regex it replaces was case-sensitive by default, and a security fix should not quietly widen what matches. Reconciling the three differences needs its own change with its own evidence.

## On the one behavior change, which MAI-Code independently found

MAI-Code compared old and new across the full matrix and reported exactly one difference: a document path containing a **newline** previously returned `null` (regex `.` does not match `\n` without `RegexOptions.Singleline`) and now resolves.

I verified it and then checked whether it is exploitable, since a newline in a URL is a CRLF-injection shape:

```
BEFORE: success=False -> null
AFTER : https://example.test/src/a\nb.cs
```

Then, on the URL that produces:

```
Uri ctor OK   : AbsoluteUri = https://example.test/src/a%0Ab.cs
                contains raw LF: False
CRLF Uri OK   : https://example.test/a%0D%0AX-Injected:%201
                contains raw CR/LF: False
HttpRequestMessage OK: https://example.test/a%0D%0AX-Injected:%201
```

.NET's `Uri` percent-encodes control characters, so no raw CR/LF reaches the wire and `HttpRequestMessage` accepts the already-encoded form. **Not exploitable**, and the newline path is not reachable from any conformant producer anyway. Recording it rather than leaving it implicit.

Note the third row of the divergence table is the related gap: `SourceLinkFetch` relies on `Uri` to encode, while `SourceDocumentPath` encodes explicitly. Same follow-up.

richlander:
## Review reconciliation — MAI-Code and GPT-5.6

Both reviewed independently on isolated worktrees at the same head. Both recommend approve.

| Finding | Reviewer | Pre-existing? | Disposition |
| --- | --- | --- | --- |
| 1. Newline in a document path now resolves where it previously returned `null` | MAI-Code | No — introduced here | **Verified not exploitable**, documented above |
| 2. Unanchored provenance regexes allow repository spoofing | GPT-5.6 | **Yes** — reproduces at `b4f35447` | Split to #3408 |
| 3. Three negative tests are vacuous | GPT-5.6 | No — introduced here | **Fixed** in `0e03cfdd` |

### Finding 3 was right, and it was the important one

GPT checked something I did not: whether my *negative* tests actually fail against the vulnerable code. Three did not. The keys I picked were ones the replaced regex never matched to begin with, so they returned `null` before the fix and `null` after it:

```
"/_/*/src"       vs "/_/a/src/b.cs"  base=<null>  head=<null>   VACUOUS
"/*/*/*/*/*/*a"  vs "/_/a/src/b.cs"  base=<null>  head=<null>   VACUOUS
"/_/*/bad/*"     vs "/_/src/a.cs"    never shadowed anything    VACUOUS
```

I reproduced this against a faithful reconstruction of the pre-fix matcher before changing anything. Three of my five negative assertions were decorative — they would have passed against the exact bug this PR exists to fix.

Fixed by choosing keys the old regex genuinely matched and resolved:

```
"/_/*/src/*"     base="https://example.test/a"     head=<null>   discriminating
"/_/*/src/b.cs"  base="https://example.test/a"     head=<null>   discriminating
"/*/*/*/*s"      base="https://example.test/_"     head=<null>   discriminating
"/_/*/a.cs"      base="https://wrong.test/src"     shadowed      discriminating
```

Per `AGENTS.md` ("asserted properties name their gate"), I did not leave that as a comment. `AssertTheReplacedImplementationWouldHaveResolved` reconstructs the replaced matcher and requires it to match, so a future edit that swaps in a key the old code ignored **fails loudly** rather than quietly going vacuous.

Non-vacuity of the guard itself proven by tampering — restoring `"/_/*/src"`:

```
Vacuous case: the replaced implementation ignored key '/_/*/src' for path
'/_/a/src/b.cs', so asserting that the current implementation ignores it
proves nothing. Choose a key the old regex matched.
   Total: 3, Failed: 1
```

Restored: `Total: 1233, Failed: 0`.

### Finding 2 is real but not this PR's

Confirmed pre-existing — it reproduces identically at the base commit and lives outside the changed lines. Filed as #3408 with the repro and a suggested fix. Both reviewers independently agreed the ReDoS fix itself is sound; GPT additionally confirmed the remaining regexes are not backtracking-prone (40 KB hostile inputs, 0 ms) and that the corpus evidence holds — its own survey found 183 distinct keys across 3,705 SourceLink-bearing PDBs, **all** conformant, alongside my 1,584 across 1,626.

### Validation at `0e03cfdd`

```
dotnet build dotnet-inspect.slnx -c Release   0 errors
ILInspector.Metadata.Tests      1233 passed, 0 failed
DotnetInspector.Services.Tests   319 passed, 0 failed
dotnet-inspect.Tests            2233 passed, 0 failed
```

The test-only change is on top of the reviewed head. Since it narrows rather than broadens what the suite accepts, and both reviewers' substantive findings are resolved or split out, I'd like one confirming pass on `0e03cfdd` before calling this ready.

richlander:
## Confirming pass on `0e03cfdd` — both reviewers clean

Both re-reviewed the fixed head on isolated worktrees, both confirmed head `0e03cfdddc6d2f703900a1e645513e960acaf3c8` against merge-base `b4f3544760d7aabd672c3be2e20b1d9f9d0510ac`.

| Reviewer | Result |
| --- | --- |
| MAI-Code | **No findings. Approve.** |
| GPT-5.6 | **No significant issues in the delta. Approve.** |

Both independently recomputed the discrimination table rather than accepting mine, and agreed:

| Pattern | Replaced implementation | Current implementation |
| --- | --- | --- |
| `/_/*/src/*` | `https://example.test/a` | `null` |
| `/_/*/src/b.cs` | `https://example.test/a` | `null` |
| `/*/*/*/*s` | `https://example.test/_` | `null` |
| `/_/*/a.cs` | `https://wrong.test/src` | ignored; conformant key still returns `https://right.test/a.cs` |

GPT specifically closed the question I was least sure of — whether the guard's *match* implies the old code *resolved*, or only that the regex matched:

> Every guarded pattern contains `*`, so a match guarantees `Groups.Count > 1`. `urlTemplate.Replace` necessarily returns non-null; therefore matching implies the old method resolved non-null.

So the gate is as strong as claimed, not weaker. It also noted the guard omits the old backslash normalization — equivalent here because every guarded path contains only `/`, and worth knowing if a future row uses a Windows-style path.

## Both findings from the first round are resolved

- **Newline paths now resolve** (MAI, introduced here) — verified not exploitable; `Uri` percent-encodes to `%0D%0A`, documented above rather than left implicit.
- **Vacuous negative tests** (GPT, introduced here) — fixed in `0e03cfdd` and gated so they cannot silently regress.
- **Unanchored provenance regexes** (GPT) — confirmed pre-existing at the base commit, split to #3408.

## Validation at `0e03cfdd`

```
dotnet build dotnet-inspect.slnx -c Release   0 errors
ILInspector.Metadata.Tests      1233 passed, 0 failed
DotnetInspector.Services.Tests   319 passed, 0 failed
dotnet-inspect.Tests            2233 passed, 0 failed
```

Ready to merge

PR #3404, opened 2026-07-28
Address rows by their displayed number, not their list position

richlander:
Fixes the bugfix half of #3364. The range grammar (`--rows 2..10`) stays open.

## Problem

`--row` did not select the row a reader counted in the output, and the failure was silent.

Four call sites filtered rows by payload availability -- or projected a filtered subset -- and then indexed that shorter list **positionally**, while every row already carried the number it was rendered with:

| Site | Defect |
| --- | --- |
| `ApiCommand.PrintUrlProjectionAsync` | filtered on `!IsNullOrWhiteSpace(row.Url)`, then indexed by position |
| `ShapeProjectionOutput.Write` | `rows[row - 1]` while each row carries `.Row` |
| `PrintProjectionOutput.Write` | `documents[row - 1]` while each carries `.Row` |
| `ApiCommand.ProjectTypeSourceFiles` | renumbered survivors `1..N` after filtering |

With rows 1-2 carrying no URL, `--row 1` returned the row displayed as 3. It returns a *real* row, so nothing looks broken.

## Change

`RowSelector.Resolve(int count)` is replaced by `Resolve(IReadOnlyList<int> rowNumbers)`. A count cannot distinguish the Nth entry of a list from the row labelled N, so removing the overload makes the mistake **unrepresentable** rather than merely fixed -- there is no longer a way to ask the question that produced the bug.

Consequences:

- `first`/`last` resolve to the endpoints of the *rendered* sequence, not `1` and the count.
- `RowNumbering.IndexOf`/`Describe` add the lookup and the error vocabulary. Errors now name the addressable rows (`Use --row 2, 5, first, or last`) instead of quoting a `1..N` range that includes rows that cannot be selected and excludes rows that can.
- **Printability never affects numbering.** A row that declares no payload keeps its number; selecting it reports `row N has no printable document` rather than sliding to a neighbour.

## Evidence

New `RowSelectionNumberingTests` builds rows numbered **2 and 5** -- contiguous rows cannot distinguish a correct implementation from a positional one. **4 of its 5 cases fail before this change**, including the silent wrong answer:

```
Row_SelectsDisplayedNumber_NotListPosition
  Expected: "alpha"   (row 2)
  Actual:   "beta"    (row 5)
```

Full CLI suite at the head commit: **2267 tests, 1 failure**, `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`. That failure **reproduces on unmodified `origin/main` at 44f2832a** (verified in a separate clean worktree) and is unrelated to row selection.

## Compatibility

User-visible error text changed on the `--print` URL path, and four existing assertions were updated:

- `selected section has N printable rows` -> `selected section has N rows`
- `failed to fetch the document for printable row N` -> `... for row N`

The count is unchanged for `Source Files` (`TypeSourceFileRow.Url` is non-nullable, so every row is printable there). `PrintProjectionOutput`'s `printable rows` wording is deliberately **kept**: those documents are printable by construction, so the term is accurate there.

Behaviour changes only where displayed and positional numbering diverged -- i.e. only where `--row` was returning the wrong row.

### DIFF
--- a/docs/design/output-shapes.md
+++ b/docs/design/output-shapes.md
@@ -140,20 +140,30 @@ modifier changes how a selected payload is rendered.
 
 `--print` projects a selected row's declared printable payload. It is unary, but
 it does not mean "take the first row." Cardinality is resolved after section
-selection, filtering, and printable-capability filtering:
+selection and filtering:
 
-| Printable rows | `--print` | `--print --row N\|first\|last` |
+| Rendered rows | `--print` | `--print --row N\|first\|last` |
 | ---: | --- | --- |
-| 0 | Error: the selected shape is not printable. | Error. |
-| 1 | Print the one payload. | Print row `1`, `first`, or `last`; any other index is an error. |
-| More than 1 | Guidance error requiring `--row`. | Print exactly the selected printable row. |
+| 0 | Error: the selected section has no rows. | Error. |
+| 1 | Print the one payload. | Print that row by its number, `first`, or `last`; any other number is an error. |
+| More than 1 | Guidance error requiring `--row`. | Print exactly the selected row. |
 
 `--print` resolves exactly one payload. There is no fan-out gesture: printing
 more than one document at a time is not currently expressible.
 
-Numeric `--row N` is one-based and counts printable rows, not every row in the
-selected table. `first` and `last` are stable aliases for the endpoints of that
-printable-row sequence.
+Numeric `--row N` addresses the row number the reader can see in the rendered
+section. It is not a position within a filtered subsequence, and in particular
+printability does not renumber anything: a row that declares no payload still
+occupies its number, and selecting it reports that it has no document rather
+than silently sliding to a neighbour. `first` and `last` are the endpoints of
+the rendered sequence, so when a projection skips rows they resolve to the
+first and last numbers actually on screen rather than to `1` and the row count.
+
+This is the one rule that makes the ordinal trustworthy. Numbering by position
+in a filtered list is wrong in the worst way available: it returns a real row,
+so nothing looks broken, and the reader has no way to see the sequence being
+indexed. Selecting by displayed number can only ever hit the intended row or
+report a miss.
 
 Because `--print` is exactly-one, failing to acquire the selected row's payload
 is an error, not an omission: it reports the failure and exits non-zero rather
@@ -162,7 +172,7 @@ selected row; whether a section producer declares a row at all is that
 producer's concern.
 
 `-n N` / `--head N` and `--tail N` are rendered-line windows applied after
-printable-row cardinality is resolved and the payload is fetched. They do not
+row cardinality is resolved and the payload is fetched. They do not
 select rows:
 
 ```text
@@ -170,7 +180,7 @@ select rows:
   multi-row selection -> error; does not choose the first row
 
 --print --row 2 --head 20
-  select printable row 2 -> fetch one payload -> render its first 20 lines
+  select row 2 -> fetch one payload -> render its first 20 lines
 ```
 
 `--rows` changes head/tail from rendered-line windows into per-table data-row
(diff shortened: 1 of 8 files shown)


richlander:
## Review reconciliation

Two adversarial reviews at `d6c77008` (MAI-Code, Gemini Pro), both re-run against the fixed head `4dc3718e`. **Both NON-BLOCKING at this head.** They disagreed on the central premise, so that is worth recording rather than smoothing over.

### Gemini Pro — one real gap, accepted in full

Cleared attack points 2-6 (first/last semantics, the removed bound check, `ThrowIfZero` reachability, the completeness sweep, `Describe` edge cases) and filed the finding that mattered: **nothing tested the path where the addressed row exists but carries no printable document.**

That is the path this PR most changes. Before the fix such a row was filtered out and the numbering slid, so the defect was structurally invisible; after the fix the row is addressable and must *report* that it has no payload rather than silently returning a neighbour. It was untested for a structural reason -- the selection logic was inline in `PrintUrlProjectionAsync` and reachable only through a network fetch.

Fixed in `4dc3718e`: `ApiCommand.SelectPrintableRow` extracted as `internal static`, plus seven hermetic tests covering a row that carries a payload, the no-payload report, `last` landing on a valueless row, an unrendered number, the multi-row guidance error, counting *rendered* rather than printable rows, and the empty section.

On re-review Gemini confirmed the gap closed, traced that `last` locks in the position *before* payload validation (so a payload-free trailing row reports a miss rather than sliding), and found the extraction behaviour-preserving.

### MAI-Code — wording accepted, severity disputed

MAI argued the premise was unfounded: the rendered sections carry no row-number column, so "address the row the reader can see" names an identifier that does not exist.

I resolved this by running the product rather than by argument, and it split:

- **Correct about the wording.** There is no row-number column in the default Markdown table. Every "the number the reader can see" / "on screen" / "displayed number" phrasing is now narrowed to "the row's position in the rendered section".
- **Severity not accepted.** The ordinal is well-defined and addressable -- it is the position counting top-to-bottom, and it is *explicit* in structured output (`--jsonl` emits `"row":1`). Out-of-range selection gives a clean `Error: row 5 is not in this section. Use --row 1 through 2, first, or last.` with no crash. The contract was described badly, not broken.

MAI did not restate the objection on re-review, and independently ran the build and full suite.

This is also exactly why `RowNumbering.Describe` earns its place: with a gapped sequence a user cannot guess that 2 and 5 are the addressable numbers, so the error enumerates them.

### Evidence

- **Non-vacuity.** Mutating `SelectPrintableRow` back to filter-then-renumber fails **4 of 12** tests in the class, including both tests written for Gemini's gap. The tests are load-bearing.
- **Full suite: 2274 tests, 1 failure** -- `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`, which I verified reproduces on unmodified `origin/main` in a clean detached worktree, and which MAI independently reproduced.
- Gapped row numbers (2 and 5) throughout the tests: contiguous rows cannot distinguish a correct implementation from a positional one.

### Non-blocking follow-up

The redesign half of #3364 (range grammar `--rows 2..10` / `2+10`, `-r` alias, bare `--row` an error) is deliberately **not** in this PR; this is the bugfix half only.

One measured input for that work: a `ZeroOrOne` option **does** swallow the following positional -- `dotnet-inspect library -S <path.dll>` binds the path to `-S` and leaves the positional unbound. So a value-taking `--rows` needs `ExactlyOne` arity plus a value validator (the pattern already at `SharedOptions.cs:143`), which also delivers the requested "bare `--row` is an error".

richlander:
Ready to merge

All CI green at `4dc3718e` (test, build-net10, markdownlint, changes). Both required adversarial reviews -- MAI-Code and Gemini Pro -- are clean on this exact head; reconciliation is in the comment above.

Worth noting: the `test` job passes on ubuntu-26.04, so the single local failure I reported (`LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`) is environment-specific to my Windows checkout rather than a real baseline break -- it reproduces on unmodified `origin/main` there and does not occur in CI at all. Either way it is unrelated to row selection.

Scope reminder: this is the **bugfix half** of #3364 only. The range-grammar redesign is tracked separately and is non-blocking for this merge.

PR #3405, opened 2026-07-28
Project payload projections through lens render paths

### LINKED ISSUE #3395: library --il-offsets drops --count, --value, --urls, and --paths
Found by adversarial review of #3390 (Gemini 3.1 Pro).

`library --il-offsets <file>` branches early to the IL-coordinate batch path,
which renders its own table (or JSON) and returns. It never dispatches any
payload projection, so all four are accepted and silently discarded.

```console
$ dotnet-inspect library <assembly> --il-offsets coords.txt --count -S "Library Info"
## IL Coordinates

Coordinate      Label  Member                             IL Offset  Meaning  Evidence
0x06000001+0x5         AnonymousType.get_Command          IL_0005    member   ...
```

Before #3390 this exited 0 with the unprojected table. #3390 makes it exit 1
with the audit error. Reproduced for `--count`, `--value`, `--urls`, and
`--paths`. `--print` is already correctly rejected by
`ValidateLibraryPrintSelection`.

Note the section filter is also ignored in this mode -- `-S` is only needed to
satisfy the single-section validation -- which is a related silent drop worth
resolving at the same time.

Proposed: honor `--count` by counting the coordinate rows, and reject
`--value`/`--urls`/`--paths` explicitly, since the batch table exposes no url
or path values.

### LINKED ISSUE #3396: package lens modes (--versions, --layout, --tfms) drop every payload projection
Found by adversarial review of #3390.

`--versions`, `--layout`, and `--tfms` are alternate lens modes: each produces
its own payload and returns before the section pipeline runs. They ignore both
`-S/--select` and all five payload projections.

```console
$ dotnet-inspect package Newtonsoft.Json --versions 1 -S Files --count --offline
13.0.4
```

The count was accepted and discarded; the printed value is a version, not a
count. Before #3390 this exited 0. #3390 makes it exit 1 with the audit error.

A sweep over 14 sections x 3 lens modes x 5 projections shows this is one
structural defect rather than a per-section one: every combination drops.

Note that without `-S`, the same request is already rejected with
`--count requires -S/--select to match exactly one section` -- so the lens modes
are reachable only through a section filter they then ignore.

Proposed:

- `--versions --count` counts the versions listed. The version list is the
  payload in that mode, so counting it is the natural projection.
- `--tfms --count` and `--layout --count` likewise count their entries.
- Reject `--value`/`--urls`/`--paths` in these modes, or define them against the
  lens payload if that is wanted.
- Reject `-S/--select` in these modes rather than ignoring it.

### LINKED ISSUE #3398: -D/--discover drops payload projections across commands
Found by adversarial review of #3390 (Gemini 3.1 Pro).

Every command that supports `-D`/`--discover` branches early to
`DiscoverOutput.Execute` / `ExecuteEffective`, which writes the discovery
table or tree and returns. That path never dispatches a payload projection, so
all five are accepted and silently discarded.

```console
$ dotnet-inspect project -D "" --count
| Name | Kind |
| ---- | ---- |
| Skills | section |
| @All | category |
```

The count was accepted and dropped; the output is the discovery listing. Before
#3390 this exited 0. #3390 makes it exit 1 with the audit error.

Reproduced for `project -D "" --count`, `library <dll> -D "" --count -S "Library
Info"`, `find System.String --platform -D "" --count`, and `type --schema -D ""
--count`. `diff` is unaffected because it does not accept projection flags.

This is the same shape as #3395 and #3396: an early-branching render path that
skips the section pipeline and therefore skips projection dispatch. Worth fixing
together with them, since a single dispatch seam would cover all three.

Proposed: the discovery listing is a row payload, so `--count` should count the
discovered rows and the shape projections should either project a discovery
column or be rejected explicitly.

richlander:
Fixes #3395, fixes #3396, fixes #3398.

The projection audit merged in #3390 reported three routes that accepted a
payload projection and then rendered something else. This fixes all three
through one seam, plus the two more the audit found while I was there.

## What was wrong

A *lens* flag renders a payload it computes itself and returns before the
section pipeline:

| Lens | Command |
| --- | --- |
| `--versions`, `--layout`, `--tfms`, `--content`, `--readme` | `package` |
| `--il-offsets` | `library` |
| `-D` / `--discover` | every command that discovers |

None of them consulted the payload projections. `--count` and the shape
projections were accepted and dropped, and `-S` was accepted and ignored.

The interesting part is why this kept recurring rather than being a set of
independent oversights. Eight options records each redeclare the five
projection flags with no shared base, so an early-returning render path had no
typed way to ask what the caller requested. Dispatching a projection required
knowing the concrete options type, which is exactly what a shared render path
does not have. `LibraryCommand` shows the consequence: the same projection
ladder is triplicated across its three resolution branches, and two early
returns jump ahead of all three copies.

## What this does

`IProjectionOptions` supplies the missing identity. `Count` is required; the
four shape projections are default interface members returning `false`, so the
four search commands that genuinely have no such options do not need
placeholder properties. (A record that declares a matching property implements
the member and the default does not apply.)

`LensProjection` is the single dispatch seam:

- **`--count` counts the lens payload** — versions, TFMs, package files, IL
  offsets, discovered artifacts — not the lines used to render it.
- **Shape projections are refused with the reason.** `--print`/`--value`/
  `--urls`/`--paths` address a cell or column of a selected section; a lens
  payload has neither. Answering would mean inferring structure from rendered
  text.
- **An explicitly typed `-S` is refused, not ignored.** A lens and a section
  selection are competing answers to *what am I looking at*.

## Behavior change to note

`-S` in a lens mode now exits 1 instead of being silently ignored. This needed
a new `SelectExplicitlySet` captured in the parser **before** the `--path`/
`--type` sugar injects into `Select` — neither `Select` nor `IncludeSections`
can distinguish a selection the caller typed from one the CLI synthesized. I
got this wrong twice before adding the explicit signal.

## Evidence

- **The audit caught my own incomplete fix.** After the first pass,
  `package --versions 1 --count` still printed `13.0.4`: the versions block has
  three further scalar early returns (cached-package hit, known-version
  confirmation, cache-first bare `--version`) plus `--latest-version`. All four
  patched.
- **Sweep: 70 lens x projection combinations, zero drops, zero audit fires.**
- **Counts verified against their payloads, not assumed.** `--layout --count` is
  19 (files) where the tree renders 28 lines (9 directory nodes);
  `--tfms --count` is 8 matching 8 listed; `library -D "" --count -S` is 3
  matching 3 data rows.
- **Non-empty payloads throughout.** Empty result sets short-circuit ahead of
  the defective dispatch — that is what caused a wrong dismissal in an earlier
  PR in this series.
- **15-command parity sweep** against a `main`-built binary: untouched paths
  byte-identical.
- CLI 2270 pass / 0 fail / 14 skipped; services 319 pass / 0 fail; lint clean.
- 11 new regression tests via `RunAppAsync`, the harness entry that actually
  exercises the audit (`PackageVersionTests` calls `InvokeAsync` directly and
  bypasses it).

## Deliberately not in scope

`package --dependencies --count` is left as-is. It is **not** a silent drop —
it is rejected with exit 1 on both `main` and this branch — but it has three
defensible answers (22 declared deps for the selected TFM group, 16 rendered
top-level nodes after display dedup, 136 `-S Dependencies --count` rows across
all TFM groups). That is a contract decision, not a bug fix. Filing separately.


richlander:

Reviewed at `aa7cdba6` (base `44f2832a`) by **GPT-5.6** and **Gemini 3.1 Pro**,
in isolated worktrees. Fixed head is `d9ede1af`; both are re-reviewing it now.

Every finding below was reproduced against a binary built from `main` before
being acted on, to separate regressions introduced here from pre-existing gaps
the merged audit reports.

### Found by both, independently

| Finding | Verdict |
| --- | --- |
| `LibraryCommand.cs:48` static `-D --schema` discovery not wired | **Fixed.** `library <dll> --schema -D --count` = 64, equal to its rendered rows. |

Two models finding the same miss is the clearest signal in this round: my own
sweep did not cover static schema discovery.

### Found by Gemini

**`--readme --print` was deadlocked, and I had regressed it.** Verified against
base: `-S "Package Readme" --readme --print` returned exit 0 / 71 lines on
`main` and exit 1 on my head. My `-S` refusal broke it, and the
`printHandledByLens` hatch I had added for exactly this case was unreachable
because the print gate ran first.

Fixed, but not the way suggested. Guarding the gate alone would have preserved
the base incantation, so I first excused `--print` from the `-S` refusal — and
the sweep immediately caught that this let `-S Files --readme --print` silently
ignore `-S Files`, which is the defect class this PR exists to remove. The
refusal is now unconditional: `--readme --print` works with no selection, and
`-S <anything> --readme --print` exits 1. **This deliberately breaks a
base-working invocation**, and is the most debatable call in the PR.

**`-D --count` was blocked by the `--count requires -S` gate** across package,
library, api, and project, which made this PR's own `-D` fix hollow. Not folded
into `lensMode` as suggested, because `-S` *is* meaningful with `-D` — it
narrows effective discovery, and `library <dll> -D "" --count -S "Library Info"`
legitimately returns 3. There is now a separate `rendersOwnPayload` predicate
used only for the count and shape gates.

### The bug that lifting the gate exposed

This is the important one, and neither reviewer could have seen it while the
gate was in place.

`PackageCommand`'s generic `if (options.Count)` branch sits **ahead** of its
`if (effectiveDiscovery)` branch. With the gate lifted, `package <pkg> -D
--count` returned **66** — the inspection document's count — where `package
<pkg> -D` renders **14** discovered rows. It also called
`ProjectionAudit.MarkHonored` on the way out.

So it exited 0, satisfied the guardrail, and answered a different question with
a plausible-looking number. That is strictly worse than the loud drops this
series has been fixing, because nothing reports it — the only way to catch it is
to check the count against the payload it claims to summarise. The shape
projections had the same ordering. Both now defer to discovery, and a test pins
each discovery count to its rendered row count.

All eight discovery counts now match their payloads exactly: package
`-D`/`-D ''` 14, `--schema -D` 17, no-target `-D` 17, library `-D` 15,
`--schema -D` 64, type `-D` 21, project `-D` 2.

### Found by GPT

**`--content --count` is rejected** — resolved as by-design, but it changed a
different decision. `docs/design/output-shapes.md` classifies a text blob as a
**Scalar**, and `--count` collapses a Vector, so counting `--content` is not a
smaller answer but a meaningless one. Applying that rule consistently showed the
real defect was the opposite one: **my own `--readme --count` returning 1**.
Both now refuse, via a `scalarPayload` parameter.

**Empty `--all-libraries` skipped the count** — fixed, now writes `0`.
Pre-existing, but trivially correct to answer.

### Dismissed with evidence

**`--dependencies` drops the count** (GPT). Reproduced identically on base and
head (exit 1, audit fires), so it is not introduced here. It has three
defensible answers — 22 declared / 16 rendered roots / 136 section rows — which
makes it a contract decision rather than a bug fix. Filed as #3406.

**`library --il-offsets` ignores `-S`** (Gemini, non-blocking). Not fixed
deliberately: widening the `-S` refusal is precisely what caused the
`--readme --print` regression above, so extending it further inside this PR is
the riskier choice. Follow-up.

### Validation at `d9ede1af`

- Full solution build clean; markdownlint clean.
- CLI **2276 pass / 0 fail / 14 skipped**; services **319 / 0 fail**.
- 80-combination lens x projection sweep: **0 audit fires, 0 silent drops**.
  The only exit-0 long output is `--readme --print`, which is now correct.
- 15-command parity sweep against a `main`-built binary: untouched paths
  byte-identical.
- 6 new regression tests, including one pinning each discovery count to its
  rendered payload.

richlander:
## Round 3 reconciliation + main integrated

Head is now `be5cc62a`, which merges `origin/main` (`92a20510`, #3404) and addresses GPT-5.6's round-3 findings.

### GPT-5.6 round 3

**BLOCKING 1 — zero-match `--content --count` returns 1.** Already fixed in `d1c342f8`, one commit after the head GPT reviewed. Verified rather than assumed: `package Newtonsoft.Json@13.0.4 --content --path 'NO-SUCH-*.md' --count` returns `0`. Independent confirmation of a defect I had found by probing the same edge, which is a good sign for the review's coverage.

**BLOCKING 2 — `--count --out <file>` silently ignores `--out`.** Real, and mine. Fixed in `be5cc62a`.

The combination was unreachable on base (`--tfms --count` failed the section gate), so lifting that gate is what exposed it. `--out` is honored for the ordinary `--content` render — 94 lines to the file — but every lens count wrote to stdout and left the file absent. That is precisely the silently-ignored-option defect this branch exists to remove, so refusing it was not good enough; a count is the command's payload and `--out` has to apply to it.

`IProjectionOptions` now carries the destination as a default interface member, so the commands with no `--out` need no placeholder property, and `CountOutput.WriteCount` routes to the file. Verified:

```
--content --path *.md --count --out F  ->  file='2', stdout empty
--tfms   --count --out F               ->  file='8', stdout empty
--layout --count --out F               ->  file='19', stdout empty
--tfms   --count                       ->  stdout '8'   (unchanged)
```

Covered by `LensCount_WritesToTheRequestedOutputFile`, which asserts the file exists rather than only checking the value, so the regression cannot come back as a stdout write.

**NON-BLOCKING — `type -S Enums --value` renders the section then trips the audit.** Confirmed pre-existing: a `main`-built binary and this head both exit 1 with the same audit failure. Filed as #3411 rather than expanded into this PR — it is the whole-surface `type` path, which has no projection dispatch at all, and is the known remaining audit-firing route recorded when the guardrail landed in #3390.

GPT is right that this qualifies my "0 audit fires" claim, so to restate it precisely: **0 fires across the 45 lens x projection combinations this PR touches.** It was never a whole-surface claim, and the whole-surface path is exactly the one still known to fire.

### main integrated

Merged rather than rebased, since the branch is public and under review. #3404 touched `ShapeProjectionOutput`, `PrintProjectionOutput`, `RowSelection`, `ApiCommand`, and the same `output-shapes.md` section, so the overlap is real and a clean textual merge does not prove semantic correctness. Re-validated on the merged head rather than trusting it.

### Validation at `be5cc62a`

- CLI **2295 pass / 0 fail / 14 skipped**; services **319 / 0**.
- Sweep: 45 lens x projection combinations, non-empty payloads, **0 audit fires, 0 silent drops**. The one exit-0 multi-line result is `--readme --print`, where the 71-line readme *is* the requested payload.
- markdownlint clean.

Both reviewers are being sent this exact head for a fixed-head pass.

richlander:
## Gemini round 2 reconciliation — one finding, and a worse bug under it

Head is now `83b04bdd`.

### Gemini's BLOCKING: `-D` projections report a section requirement instead of the lens refusal

Correct, and half of it was still live. The **shape** gates (`--value`/`--urls`/`--paths`) had already been shielded for discovery in `921c9560`, but the **`--print`** gates had not, at `ApiCommand.cs:159`, `LibraryCommand.cs:173`, and `PackageCommand.cs:141`. On its own this is a message-quality bug — exit 1 either way — so it looked like the cheapest fix in the review.

**It was not.** Shielding those gates exposed the ordering hazard a second time:

```
package Newtonsoft.Json@13.0.4 -D "" --print
  -> exit 0, and the 71-line README
```

The grounding branch (`if (options.ShowReadme || options.Print)`) sits *ahead* of the discovery branch. Once `--print` stopped demanding `-S` in a mode that renders its own payload, discovery fell straight into grounding and printed the readme — **with the audit satisfied**, because a print writer genuinely did run, just over an unrelated payload. Same shape as the `-D --count` returning 66 instead of 14 earlier in this PR: a plausible answer to a question nobody asked, at exit 0, with nothing to report it.

Discovery now wins, and refuses `--print` with a reason naming `-D`. Covered by `Discover_Print_RefusesInsteadOfPrintingTheGroundingDocument`, which asserts the readme text is *absent* rather than only checking the exit code — the exit code was never the problem.

Both reviewers have now independently pushed me into this same hazard from different directions. That it recurred twice in one PR says the risk is structural: **a generic projection branch must never sit ahead of a branch that renders its own payload.** Both reviewers are hunting a third instance on the current head.

### Gemini's NON-BLOCKING and the two tradeoff verdicts

- `library --il-offsets` silently ignores `-S`: Gemini agrees with deferring it, since widening the `-S` refusal is precisely what caused the regression it caught. Staying filed as follow-up.
- **The unconditional `-S` refusal in lens modes** — the most debatable decision in this PR, deliberately breaking base-working `-S "Package Readme" --readme --print`. Gemini's verdict: *"absolutely the correct call. Silently ignoring a user's explicit filtering intent when they request a lens is exactly the defect class you are eliminating."* Both reviewers now agree, so I consider it settled.

### GPT round 4: `--out` to an unwritable path throws

Real, but the base comparison was not apples-to-apples — main never reaches a write there, since `--tfms --count --out` fails the section gate first. Against a main-built binary on paths that *do* write, `-S Files --out DIR`, `--readme --out DIR`, and `--content --out DIR` all crash identically. Every `--out` writer calls `File.WriteAllText` unguarded, so the new count path behaves exactly like its neighbours: newly reachable, not newly broken.

Guarding only the count would leave `--tfms --count --out DIR` clean while `-S Files --out DIR` still prints a stack trace. Filed as **#3413** for one shared write helper across all call sites.

### Also filed

**#3411** — `type -S Enums --value` renders the section then trips the audit. Confirmed pre-existing on main; it is the whole-surface path with no projection dispatch, which PR 5 addresses.

### Validation at `83b04bdd`

- CLI **2296 pass / 0 fail / 14 skipped**; services **319 / 0**; markdownlint clean.
- 50 lens/discovery x projection combinations: **0 audit fires**, and every `--count` verified **equal to its rendered payload** rather than merely present.
- Base parity: bare `--print` and `-S Grounding --print` byte-identical to main. `--readme --print` now exits 0 with 71 lines where main exited 1 — the intended fix for the regression Gemini caught.

richlander:
## GPT round 4: clean at `83b04bdd`

> "Your rebuttal is correct. I verified main crashes on a real file-writing path. The prior finding is pre-existing and should not block this PR. Fixing all writers through one shared helper in #3413 is the right scope."
>
> "**No blocking findings at `83b04bdd`.**"

That is one of the two required fixed-head reviews clean. Gemini's fixed-head pass on the same head is still running.

### Its one non-blocking finding was worth more than its label

GPT reported a *third* ordering instance: `package <pkg> --dependencies -S 'Package Readme' --print` exits 0 with the README and silently ignores `--dependencies`. Confirmed pre-existing (main is byte-identical).

Probing it showed the problem is **general, and larger than the report**. `package` has several flags that each render their own payload, and they are mutually exclusive in practice — but only one pair is actually validated:

```
--readme --layout       -> exit 0, layout tree   (--readme dropped)
--layout --tfms         -> exit 0, layout tree   (--tfms dropped)
--readme --tfms         -> exit 0, tfm list      (--readme dropped)
--versions --layout     -> exit 0, version list  (--layout dropped)
--dependencies --layout -> exit 0, layout tree   (--dependencies dropped)
--dependencies --readme -> exit 0, readme        (--dependencies dropped)

--content --readme      -> Error: --readme cannot be combined with --content.   <- the only guarded pair
```

This is #3364's defect class one level up: not a *projection* silently discarded, but a *payload mode* silently discarded, with the winner decided by `if`-branch order rather than by any contract. **`ProjectionAudit` cannot catch it**, because the audit tracks the five projection flags only — a lens losing to another lens leaves no trace. That is a real limit of the guardrail from #3390 and worth stating plainly.

Filed as **#3414** with the full matrix and a suggested fix in the shape of the existing `ValidateMultiPackageMode`. Not folded in here: it introduces *new* refusals for combinations that currently exit 0, which is a behavior break deserving its own review pass rather than a late addition to a branch that has already been through four review rounds.

### Follow-ups filed from this review cycle

| Issue | What | Status |
| --- | --- | --- |
| #3406 | `--dependencies --count` contract | needs a decision, not a fix |
| #3411 | whole-surface `type -S X --value` renders then trips the audit | pre-existing; PR 5 |
| #3413 | `--out` to an unwritable path throws | pre-existing, product-wide |
| #3414 | two lenses combined: one silently discarded | pre-existing |

All four are pre-existing on `main` and independently verified as such against a `main`-built binary, rather than assumed.

richlander:
## Both fixed-head reviews clean at `83b04bdd`

**Gemini 3.1 Pro:** *"The current head is **clean of regressions**. I found zero blocking issues. The shape gates and `--print` gates are now properly shielded across the board, and the `--content`/`--readme` count logic behaves exactly as specified."* It also agreed with the #3413 scoping.

**GPT-5.6:** *"No blocking findings at `83b04bdd`."*

Two different model families, both on the exact merge head, both clean. Every blocking finding across four rounds was either fixed or reproduced on a `main`-built binary and filed.

### What the reviews actually bought

Worth recording, because it argues for the process rather than just reporting it. The reviews found **three defects I would have shipped**, and none of them were the ones I was watching for:

1. **Gemini** caught that my new `-S` refusal broke base-working `-S "Package Readme" --readme --print`. A real regression, invisible to the test suite because no test covered it.
2. **GPT** proved `--content` is a Vector, reversing my Scalar classification — with a fixture, against my stated reasoning. I had it backwards in the shipped code.
3. **Gemini's** apparently-cosmetic finding about `-D --print` reporting the wrong error message. Fixing it exposed `package <pkg> -D --print` returning the **71-line README at exit 0 with the audit satisfied**.

That third one is the pattern of the whole series: the dangerous bugs were never the loud ones. Both times, a gate that *looked* like it was validating input was actually **masking a wrong-payload bug behind it**, and the bug only became visible once the gate was lifted. The `-D --count` returning 66 instead of 14 was found the same way.

### The limit of the guardrail, stated plainly

`ProjectionAudit` from #3390 verifies that *some* projection writer ran. It does **not** verify the writer ran over the payload the caller asked for. A projection that answers the wrong payload satisfies the audit perfectly — which is precisely what `-D --count` = 66 and `-D --print` = README did before this PR.

So the guardrail catches *dropped* projections, not *misdirected* ones. Gemini's hunt turned up more of the latter (inventoried in #3420, all pre-existing), including `type --shape -S Interfaces --count` returning `9` — identical to the same command without `--shape`, so the lens is discarded and a plausible number answers a different question. I would rather the guardrail's scope be understood than assumed.

### Validation at `83b04bdd`

- CI green: `test (ubuntu-26.04)`, `build-net10`, `markdownlint`, `changes` all pass on this exact head.
- CLI **2296 pass / 0 fail / 14 skipped**; services **319 / 0**.
- 50 lens/discovery x projection combinations: 0 audit fires, and every `--count` verified **equal to its rendered payload**, not merely present.
- Base parity: untouched paths byte-identical to a `main`-built binary.

### Follow-ups, all verified pre-existing on `main`

| Issue | What |
| --- | --- |
| #3406 | `--dependencies --count` contract — needs a decision, not a fix |
| #3411 | whole-surface `type -S X --value` renders, then trips the audit |
| #3413 | `--out` to an unwritable path throws, product-wide |
| #3414 | two lenses combined: one silently discarded |
| #3420 | inventory of generic branches ordered ahead of lens branches |

None are regressions from this PR; each was checked against a `main`-built binary rather than assumed.

**Ready to merge**

richlander:
## Conflicts resolved — and the merge reintroduced the exact bug this PR removes

Head is `8d7ea2c1`; `origin/main` (`4976db95`) is now an ancestor. Six conflicts, all the same shape: #3229 and this branch each appended an optional parameter to the same discovery signatures. `DiscoverOutput.Execute`/`ExecuteEffective` now take both `listedCategoryDoors` and `projection`, and the library call sites pass both.

**The textual resolution compiled cleanly and was wrong.**

```
library <dll> -D ""            -> 13 rendered rows (13 --jsonl rows)
library <dll> -D "" --count    -> 15          <- counts a row set the command never renders
library <dll> --schema -D "" --count -> 61 vs 58 rendered
```

The projection built its row list *without* `listedCategoryDoors` while the render passed it, so `--count` answered over an uncurated set: **exit 0, audit satisfied, wrong payload** — precisely the defect class this branch exists to remove, reintroduced by a merge, in the very code that removes it.

Worth being blunt about why it slipped: the conflict looked like a pure signature-ordering question, the build succeeded, and `package -D --count` (14 = 14) still matched, because the package path has no curated catalog. Only the *library* path drifted. Had I trusted the green build and the one spot check, this ships as a silent wrong answer.

Both call sites now build the same list, and `Discover_Count_EqualsTheRowsDiscoveryRenders` pins the count to the rendered `--jsonl` rows for both the plain and `--schema` catalogs, so a third parameter cannot silently diverge again. The test asserts the probe rendered a non-zero number of rows first — an empty payload would otherwise pass while proving nothing.

### One test was silently no longer testing anything

`AllLibraries_EmptyMatch_CountsZeroRatherThanReturningSilently` selected `-S OpenTelemetry`, a section #3229 removed. An unknown `-S` is rejected *before* the render path, so the empty-match short-circuit it was written to cover was no longer being reached. It now uses a section the package genuinely has no rows for, which is what the test always meant.

### Validation at `8d7ea2c1`

- CLI **2347 pass / 0 fail / 14 skipped** (count up from 2296 with #3229/#3415's tests plus the new theory); services **319 / 0**; markdownlint clean.
- 48 lens/discovery x projection combinations: **0 audit fires**; every discovery `--count` verified equal to its `--jsonl` row count across `package`, `library`, `library --schema`, `type`, `project`, and a narrowed `-S`.
- #3415's new `--rows` range grammar (`1..3`, `2+2`, `3..`) exercised against `--count` and `--value`: no audit fires.

Both fixed-head reviews were clean at `83b04bdd`. This head differs only by the merge and the two fixes above; given that the merge silently produced a wrong count, I would not call the prior reviews binding on it. **Ready to merge** once CI is green, but I am happy to send it back for a short re-review of the merge commit if you would prefer — the conflict resolution touched the exact seam the reviews were scrutinizing.

richlander:
## Re: "the merge introduces the bug, and then there are no commits"

Fair catch on the sequencing, and worth being precise about what happened.

The fix **was** there — it was inside the merge commit `8d7ea2c1` itself (parents `83b04bdd` + `4976db95`), not in a commit after it:

```
$ git diff 8d7ea2c1^1 8d7ea2c1 -- src/dotnet-inspect/Output/DiscoverOutput.cs
-    var projectedRows = GetDiscoveryRows(..., catalogHiddenSections);
+    var projectedRows = GetDiscoveryRows(..., catalogHiddenSections, listedCategoryDoors);
```

But **you were right to distrust it.** A semantic fix carried inside a conflict resolution is effectively invisible: GitHub does not show it in the PR's Files-changed view, so a reviewer reading the commit list sees a merge, a comment claiming a bug was fixed, and no commit to inspect. That is a bad way to present a correctness fix, regardless of whether the code is right.

## So I did the newest integration the other way

`main` moved again (six commits, including #3424's `--rows` rework and #3436's `--include-unlisted`). Two commits now, deliberately separated:

- **`7b7341fa`** — merge, **conflict resolution only**. Both sides had changed the same two single-version output blocks; the projection runs first and returns, then the render path picks its spelling.
- **`37906bfe`** — the actual fix, as a normal reviewable commit.

## What the merge exposed this time

#3436 added two listing-aware render paths that return before the section pipeline, so they dropped a requested projection exactly as the lens paths did before this branch:

```
package Newtonsoft.Json --versions --include-unlisted --count
  -> Error: '--count' was accepted but this command path produced unprojected output.
```

**The audit caught both on the first run after the merge** — the gap arrived with new code, not with this branch, and the #3390 guardrail did precisely the job it was built for on someone else's change.

Now:

| | count | payload | |
| --- | --- | --- | --- |
| `--versions --include-unlisted` | 53 | 53 | OK |
| `--versions --include-unlisted --preview` | 84 | 84 | OK |
| `--versions` | 53 | 53 | OK |
| `library -D` / `--schema -D` | 13 / 58 | 13 / 58 | OK |
| `package -D` / `type -D` | 14 / 21 | 14 / 21 | OK |
| `--tfms` / `--content` | 8 / 2 | 8 / 2 | OK |

The range rows are materialized before counting — counting a lazy sequence and then re-enumerating it for the render is how a count starts to disagree with its payload.

### One thing I checked and am *not* fixing

`package Newtonsoft.Json@[12.0,13.0] --versions` returns all 53 versions ending at `3.5.8`, so the range looks unapplied. I touched that branch, so I verified rather than assumed: a `main`-built binary returns the identical 53 rows. **Pre-existing, not from this branch.** Flagging it because it looked like mine and someone should confirm it is intended.

### Validation at `37906bfe`

CLI **2381 pass / 0 fail / 14 skipped**; services **336 / 0**; every count above verified against its rendered payload.

Both fixed-head reviews were clean at `83b04bdd`, two integrations ago. Given that each integration since has exposed a real projection gap, I would not treat those reviews as binding on this head, and a re-review of `7b7341fa..37906bfe` is cheap and worth it.

richlander:
## Retracting the earlier "Ready to merge" — it is stale

It was accurate when posted, at `83b04bdd`. The head is now `37906bfe`, two integrations later, and **the current head does not have two clean fixed-head reviews.**

| commit | reviewed |
| --- | --- |
| `83b04bdd` | GPT clean, Gemini clean |
| `8d7ea2c1` | no — merge with a fix buried inside it |
| `7b7341fa` | no — merge, conflict resolution only |
| `37906bfe` | no — the `--include-unlisted` projection fix |

The unreviewed commits are not cosmetic. Each integration exposed a real projection gap — the discovery count answering 15 for a 13-row listing, then `--include-unlisted` dropping the count entirely — in exactly the seam both reviewers were asked to attack. Treating reviews of an earlier head as covering them would be exactly the kind of unearned confidence this PR series exists to remove.

Both reviewers are being sent `37906bfe` now. I will not call this ready again until both come back clean on that exact head.

richlander:
## Round 3 reconciliation — one blocking finding, found independently by both reviewers

Both reviewers reported the **same** bug at `37906bfe`, reached from different directions. That convergence is the strongest signal this round produced.

### BLOCKING (fixed): `--il-offsets --count` ignored `--rows`

```console
$ dotnet-inspect library <dll> --il-offsets coords.txt --rows 1 --jsonl | wc -l
1
$ dotnet-inspect library <dll> --il-offsets coords.txt --rows 1 --count
2
```

Exit 0, projection audit satisfied, count describing a payload the caller never asked for — the fourth instance of this PR's core defect class, and one my own sweeps missed because I never crossed `--rows` with a lens count.

Fixed in `cb3b402e`.

**On which way to fix it.** Gemini offered two options: window the count, or refuse `--rows` on lenses. I took the first, because it is not a new contract — it is the one `main` already has:

```console
# main @ 1955b054
$ dotnet-inspect library <dll> -S Resources --rows 3 --count
3          # and the render emits exactly 3 rows
```

The ordinary section `--count` has always honored `--rows`. The lens was the outlier.

**How.** `RowWindow` gained `Apply`, which windows a materialized row list. Both the row limiters and this projection resolve through the existing `RowWindow.Resolve` against the same data-row count, so the window cannot mean one thing to the renderer and another to the count — the property that file's own doc comment already asked for.

The batch exit code still derives from every row, not the windowed ones: a coordinate that failed to resolve is a resolution result, not a display concern. Both reviewers were explicitly asked to attack that call.

**Evidence.** A theory over head, tail, `2..3`, `3..`, and a wider-than-batch window compares `--count` against `--jsonl` rows. A second test pins that `--rows 1 --head` yields `first` and `--rows 1 --tail` yields `second` — a count can match the total while describing the wrong rows, so direction is pinned separately. Reverting the one-line fix fails 4 of the 5 theory cases.

### NON-BLOCKING, pre-existing, filed

- **#3457** (Gemini): `package -S <section> --rows N --count` honors the window while the package render ignores it — the inverse direction. Byte-identical on a `main`-built binary. Cause is `WritePackageTable` not forwarding `options.Rows`, where `WriteLibraryResult` does.
- Version lenses ignore `--rows` in **both** render and count, so they agree and the invariant holds. Pre-existing, #3414 class.

### Corrections to my own earlier claims

- GPT is right that `[12.0,13.0]` is **not** this CLI's range grammar — it uses `A..B`, and `Newtonsoft.Json@12.0.1..13.0.1 --versions` renders and counts 4 rows correctly. My earlier "version range is ignored" note was wrong; withdrawn.
- Gemini confirmed the `listedCategoryDoors` merge fix is correctly integrated: both the projection and the renderer call `GetDiscoveryRows` with identical arguments.

### Also in this head

`0e140d0a` merges `origin/main` @ `adaf9414` (#3439, #3389, #3440). No conflicts; the only overlapping file was `ApiCommand.cs`, where main swapped `SourceLinkResolver.ExtractMethodBody` → `BodySlicer.ExtractMethodBody` well away from the projection dispatch. Because the last two merges each silently broke a count, I re-ran the full count-vs-payload sweep after merging rather than trusting the clean build.

### Validation at `0e140d0a`

- CLI **2392 pass / 0 fail / 14 skipped**; services **343 / 0**
- count == rendered payload: versions 53, `--include-unlisted` 53, `--preview` 54, `--tfms` 8, `package -D` 14, `library -D` 15, `library --schema -D` 58, `type -D` 5, il-offsets 3, `il-offsets --rows 2` 2, `il-offsets --rows 1 --tail` 1, `-S Resources --rows 3` 3

Fixed-head reviews at `0e140d0a` are in flight with both reviewers. Not ready to merge until both come back clean on that exact head.

richlander:
## Round 4 reconciliation — both reviews clean

### GPT: one blocking, withdrawn after rebuttal

GPT reported that `package A B --content --path LICENSE.md`, where only A has the file, renders 2 `--jsonl` rows while `--count` answers 1, and argued the count must equal rendered rows.

I did not change the count, because GPT's own two findings contradict each other and only one can hold:

- Round 2 (blocking): `--content --path 'NO-SUCH-*' --count` returning **1** is wrong; it must be **0**. Fixed in `d1c342f8`.
- Round 4 (blocking): the count must equal rendered rows — which makes that same case **1** again.

So the real question is whether a `found:false` row is payload or presentation, and the product already answered it:

- **`--skip-empty` exists to remove exactly these rows**, and the pre-existing test is named `Package_MultiplePackages_PathContent_PreservesEmptyPackageBlock`. Under `--skip-empty` the render emits 1 row and the count is 1 — they agree exactly.
- **`--bare`** never emits the placeholder and refuses outright when nothing matched.
- The placeholder is **per package, not per path**: `--path LICENSE.md --path no-such.zzz` against one package renders 1 row and counts 1, with no placeholder. Its `path` is `""` and `size` is `0`. It is not a file.
- The render is **byte-identical on a `main`-built binary**, so this PR added only the count.

**GPT's verdict:** *"Yes — the rebuttal holds. `--count` answers matched files; `found:false` is a per-package presentation marker. Counting it would regress the zero-match contract. I withdraw the blocking finding."*

**Gemini's adjudication:** *"If `--count` included placeholders, a script querying 50 packages for `LICENSE.md` would receive a count of 50 regardless of whether the file exists in 1 package or 50. That renders `--count` mathematically useless."*

No behavior changed. `96c2cf79` adds the rule to `output-shapes.md` and a network-free test that renders one selection three ways — default 2 rows, `--skip-empty` 1, `--count` 1 — so the deliberate divergence can never later be mistaken for a dropped projection. Gemini confirmed the test hard-fails if anyone "fixes" the count to `rows.Count`.

### Gemini: clean, with an independent completeness audit

Gemini audited **all 13 `LensProjection.TryProject` call sites** and confirmed `--il-offsets` was the only lens honoring `--rows` in the render but not the count — the fix is complete, not just locally correct. It also validated `RowWindow.Apply` against open ranges, out-of-bounds starts, and unlimited windows, and endorsed keeping the batch exit code unwindowed: *"ignoring an unresolvable coordinate just because it fell outside the `--head 5` view would hide a real failure."*

### Final head, and one more integration

`db7b6b18` merges `origin/main` @ `27cd64ee`. That included **#3453, which changes library section-to-scanner binding** — the same class of change that silently broke the discovery count during the #3229 merge. No conflicts, but a clean merge has twice proven insufficient here, so I re-ran the full sweep rather than trusting the build:

| check | count | rendered |
| --- | --- | --- |
| `library -D` | 15 | 15 |
| `library --schema -D` | 58 | 58 |
| `library -D @Surface` | 4 | 4 |
| `type -D` | 5 | 5 |
| `package -D` | 14 | 14 |
| `--versions` | 53 | 53 |
| `--versions --include-unlisted` | 53 | 53 |
| `--tfms` | 8 | 8 |
| `--il-offsets` | 3 | 3 |
| `--il-offsets --rows 2` | 2 | 2 |
| `-S Resources --rows 3` | 3 | 3 |

All unchanged. CLI **2393 pass / 0 fail / 14 skipped**; services **343 / 0**; markdownlint clean.

The delta from the reviewed head `96c2cf79` is that merge alone — no conflicts and no product edits of mine. Per `AGENTS.md`, a mechanical integration does not require another adversarial round; the count-vs-payload sweep above is the evidence that it stayed mechanical in behavior and not just in text.

### Follow-ups filed, all verified pre-existing against a `main` binary

#3406, #3411, #3413, #3414, #3420, #3457.

richlander:
Ready to merge

Both fixed-head adversarial reviews are clean at `96c2cf79`; GPT withdrew its one blocking finding after the rebuttal, and Gemini adjudicated the same question independently and verified the delta. The only change since is `db7b6b18`, a no-conflict integration of `origin/main` @ `27cd64ee`, re-verified with the full count-vs-payload sweep rather than a clean build.

- CI: all checks pass; `MERGEABLE` / `CLEAN`
- CLI 2393 pass / 0 fail / 14 skipped; services 343 / 0; markdownlint clean
- `origin/main` is an ancestor of the head

Non-blocking follow-ups, all verified pre-existing against a `main`-built binary and filed separately: #3406, #3411, #3413, #3414, #3420, #3457.

One note for whoever merges #3448 (package section model): it is the same class of change as #3229 and #3453, both of which passed a clean build while silently breaking a count. Whichever of #3405/#3448 lands second, please re-run the count-vs-payload sweep rather than relying on the build and the audit — `ProjectionAudit` verifies that a projection ran, not that it ran over the right payload.

Issue #3400, opened 2026-07-28
CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition fails deterministically on Windows: forwarder hop out of facade assemblies never resolves


richlander:
`CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition` fails deterministically on Windows against a shared-framework layout whose framework assemblies are pure type-forwarder facades. It is not a flake: 3/3 in isolation and every full-suite run over a long session, always with the same assertion.

```
expected at least one framework TypeRef to resolve to readable metadata
  src\ILInspector.Analysis.Tests\LibraryBodyIndexTests.cs(134,0)
```

## Where it is not

Both early stages are healthy, so this is not candidate discovery and not assembly resolution.

Probing `typeof(Console).Assembly.Location` with the same `AssemblyDependencyResolutionOptions` the test uses:

```
target: C:\Program Files\dotnet\shared\Microsoft.NETCore.App\11.0.0-preview.6.26359.118\System.Console.dll
typeRefs=123  assemblyScoped=122  frameworkRefs=122
ref assemblies: System.Memory, System.Runtime, System.Runtime.InteropServices,
                System.Text.Encoding.Extensions, System.Threading
```

All five resolve at the assembly level:

```
resolve System.Runtime                    -> ...\11.0.0-preview.6.26359.118\System.Runtime.dll
resolve System.Runtime.InteropServices    -> ...\System.Runtime.InteropServices.dll
resolve System.Threading                  -> ...\System.Threading.dll
resolve System.Memory                     -> ...\System.Memory.dll
resolve System.Text.Encoding.Extensions   -> ...\System.Text.Encoding.Extensions.dll
```

So 122 framework TypeRefs are found and every referenced assembly resolves to a real file. The loop still completes without a single `TryResolveExternalTypeDefinition` success.

## Where it is

Every one of those assemblies is a facade on this layout:

| assembly | TypeDefs | ExportedTypes |
| --- | ---: | ---: |
| `System.Runtime.dll` | 1 | 954 |
| `System.Threading.dll` | 17 | 30 |
| `System.Private.CoreLib.dll` | 2957 | 0 |

[...]

PR #3409, opened 2026-07-28
Narrow the Callers scope by direct type reference

richlander:
Closes part of #3333 (increment 2).

## The increment was redirected by a negative measurement

#3333''s increment 2 proposed narrowing candidates by **TypeRef** instead of AssemblyRef. That is
the right idea, but it has to be applied somewhere, and the two consumers of the candidate set
differ:

### Why it cannot go in the transitive graph — worked

Let `foo` live in assembly F, `bar` in B, `prebar` in A, with `prebar` -> `bar` -> `foo`.

- **B** calls `foo`, so B''s `TypeRef` table must carry a row naming `foo`''s declaring type.
  `CallerScopeTypeFilter` keeps B. Correct at depth 1.
- **A** calls `bar`, and references B. A may never mention `foo`''s declaring type at all.
  `CallerScopeTypeFilter` rules A out — so at depth 2 `prebar` disappears and the tree silently
  comes back one level short.

This is not hypothetical. For `ImmutableArray<T>.Length` the assemblies carrying an
`ImmutableArray` 1 `TypeRef` row are exactly `System.Collections.Immutable`,
`System.Diagnostics.StackTrace`, `System.Reflection.Emit` and `System.Reflection.Metadata`.
`System.Private.Xml` references one of those four but names `ImmutableArray` 1 nowhere, so it
enters the reverse-reference closure at hop 2. It is `prebar`''s assembly.

The reverse-reference closure exists precisely to cover that case, and does so correctly. That is
why this PR leaves it alone.
**It cannot be applied to the transitive `Call Graph` at all.** `CallerScopeFilter`''s own contract
says why: an assembly that never names the target can still appear in the tree by calling something
that does, so "testing a direct reference alone drops those upstream callers and silently shortens
the tree." The only *sound* place to put TypeRef information into a transitive closure is its
**seed** — and a closure''s fixpoint does not depend on its seed once the seed reaches a hub. That
is not a near-tie that measurement resolved narrowly; it is the same set by construction, hence
exactly 0%, hence increment 2b is **withdrawn**. The mechanism is traced under
[Why the win varies](#why-the-win-varies-so-much-between-cases): seeding with 4 candidates instead
of 169 still saturates at 169 by hop 3, because hop 2 lands on a corelib-canonical facade and 168
of 169 assemblies reference one.

**Where it can be applied, it is worth a lot.** `Callers` is strictly single-hop, so it has no
upstream to lose — the soundness objection above simply does not arise. This PR applies it there
and nowhere else.

What the measurement *did* find is that the `Callers` section never needed a closure at all.

## What changed

`MethodBodyInspectionSession.CallerEdges` is **strictly single-hop** — it scans each scope assembly for call sites whose callee matches the target and never asks what calls those. It was being handed `CallerScopes`, the closure computed for the transitive `Call Graph`. So rendering `Callers` on a framework scope opened **169** assemblies to answer a question that only the few naming the declaring type can possibly answer — measured, **2 to 44** of them.

`CallerScopeTypeFilter` (new, Analysis) answers the narrower question: does this candidate name the target's declaring type at all?

- It decodes the candidate's `TypeRef` rows with **the same `TypeRefDecoder`** the matcher's `TypeRef` values come from. Equivalence with `MemberPattern.MatchesCrossAssembly` is therefore **by construction**: nested naming (`Outer+Inner`), generic arity, ModuleRef/nil resolution scopes and corelib canonicalization cannot drift.
- A callee declared by a **TypeDef** carries its own assembly's name and appears in no TypeRef row, so an assembly whose canonical identity equals the target's is kept without scanning.
- Anything undecidable (unreadable image, `Unsupported` TypeRef row, no typed declaring identity) is **kept**.

`DirectCallerScopes` is a **separate path with its own cache**, deliberately. `CallerScopes`' two lenses share a cache when allocations are not requested, so narrowing in place would silently narrow the transitive graph and truncate it — the #3331 round-3 defect. When the graph lens has already opened its wider set, that set is reused as-is: narrowing only avoids *opening*, so re-deciding already-paid sessions saves nothing.

## Measured — output byte-identical in all nine cases

Base `44f2832a`. Time and peak working set.

| case | base | head |
| --- | --- | --- |
| 1 `ImmutableArray<T>.Length` | 8458 ms / 685 MB | **989 ms / 96 MB** |
| 2 corelib facade `StringBuilder.Append:1` | 8696 ms / 721 MB | 7222 ms / 628 MB |
| 3 `Regex.IsMatch:1` | 6915 ms / 680 MB | **3018 ms / 286 MB** |
| 4 `JsonSerializer.Serialize:1` (miss) | 917 ms / 98 MB | 903 ms / 98 MB |
| 5 unscoped | 3987 ms / 283 MB | 3674 ms / 283 MB |
| 6 both sections | 10692 ms / 939 MB | 17359 ms / 948 MB &dagger; |
| 7 `Call Graph` only | 14764 ms / 936 MB | 13181 ms / 951 MB &dagger; |
| 8 `List<T>.Add` | 8882 ms / 723 MB | **4499 ms / 314 MB** |
| 9 nested `List<T>+Enumerator.MoveNext` | 9684 ms / 719 MB | **5639 ms / 309 MB** |

Best case is **8.6x faster** and uses **7.1x less peak memory** (case 1: 8458 ms -> 989 ms, 685 MB -> 96 MB). Ratios below are base/head, so larger is better.

&dagger; **The peak-WS polling loop in the harness was itself distorting the near-tie rows.** Re-measured without it, 5-6 alternating reps, medians:

| case | base | head | |
| --- | --- | --- | --- |
| 5 unscoped | 3652 ms | 3706 ms | +1.5%, noise |
| 6 both sections | 9635 ms | 9967 ms | **+3.4%, real** |
| 7 `Call Graph` only | 9787 ms | 9672 ms | -1.2%, noise |

Case 6 is a genuine small cost and it is the predicted one: when `Callers` renders before `Call Graph`, it opens ~3 narrow sessions that the graph then re-opens as its own, plus 169 cheap metadata classifications — about 2% of 169 body decodes. Reported rather than smoothed away.

## Why the win varies so much between cases

The saving is the ratio of two independently varying quantities, both measured here by calling
the product''s own `CallerScopeFilter.SelectCouldReach` and the new `CallerScopeTypeFilter` over
the same 182-file scope:

| case | base opens | base MB | head opens | head MB | time | peak WS |
| --- | --- | --- | --- | --- | --- | --- |
| 1 `ImmutableArray<T>.Length` | 169 | 57 | **4** | **2** | 8.6x | 7.1x |
| 2 `StringBuilder.Append:1` | 169 | 57 | 38 | 46 | 1.2x | 1.15x |
| 3 `Regex.IsMatch:1` | 169 | 57 | 9 | 17 | 2.3x | 2.4x |
| 4 `JsonSerializer.Serialize:1` | **2** | 1 | 2 | 1 | 1.0x | 1.0x |
| 8 `List<T>.Add` | 169 | 57 | 44 | 33 | 2.0x | 2.3x |
| 9 `List<T>+Enumerator.MoveNext` | 169 | 57 | 32 | 29 | 1.7x | 2.3x |

**Base saturates because the corelib facades are a hub.** `mscorlib`, `netstandard` and
`System.Runtime.Extensions` all canonicalize to `System.Private.CoreLib`, and **168 of 169**
readable framework assemblies reference a corelib facade. So as soon as the reverse-reference
closure selects any facade, that facade republishes itself under the corelib name and the next
hop takes everything. Traced for case 1, which no facade references directly:

```text
=== System.Collections.Immutable ===
  hop 1: +4   -> 4     System.Collections.Immutable, System.Diagnostics.StackTrace,
                       System.Reflection.Emit, System.Reflection.Metadata
  hop 2: +6   -> 10    ... including two corelib-canonical assemblies
  hop 3: +159 -> 169   saturated
=== System.Text.Json ===
  hop 1: +2   -> 2     System.Net.Http.Json, System.Text.Json
  hop 2: +0   -> 2     stops
```

`System.Text.Json` postdates the netstandard surface, so nothing forwards it and no facade
references it — the only case where base was already tight, and correctly the only case with no
win. **This is also the mechanism behind the 0% result above**: seeding the closure with narrowed
candidates cannot help when the closure saturates two hops later regardless of its seed.

**Head varies with how popular the declaring type is**, which is a completely different question
from how popular its assembly is. `ImmutableArray<T>` is named by 4 assemblies; `List<T>` by 44.

**The cost driver is IL bytes decoded, not assembly count.** Case 2 drops 169 -> 38 assemblies but
only 57 MB -> 46 MB, because the survivors are the large ones (corelib itself, `System.Private.Xml`,
`System.Linq.Expressions`) — and it is the byte ratio, not the count ratio, that the observed 1.2x
matches. Case 1 is the reverse: 25x fewer bytes but 8.6x faster, because it has hit the floor.
Case 4 measures that floor at ~900 ms / ~98 MB, and case 1''s head is 989 ms / 96 MB — startup plus
the target assembly, with essentially nothing left to remove.
## The transitive graph is untouched, and that is checked directly

`-s "Call Graph"` does **not** suppress the auto-included `Callers` section, so a whole-file diff on such a run is not evidence about the graph. Extracting `## Call Graph` alone: **51 lines, 0 diffs** across base, head, **and** a deliberately over-narrowing mutant.

## Evidence

**18 new tests** (15 Analysis, 3 CLI), all on real compiled fixtures. The load-bearing ones:

- `AssemblyNamingTheTargetAssemblyButNotTheTypeIsRuledOut` — asserts **both** filters, so the disagreement between them is the claim. `CallerGraphCallerTwin` references the target *assembly* (it calls `Target.Api.Ping`) but never `Box\`1`, with zero fixture churn.
- `FrameworkTypeMatchesOnlyUnderItsOwnAssembly` — pins assembly sensitivity in both directions on a real artifact.
- `NothingRuledOutEverContributesAMatchingCallSite` — soundness over every fixture x declaring type, compared against **the matcher itself**, with two non-vacuity guards.
- `CallerEdges_AreUnchangedByNarrowing` — every method of the target fixture, un-narrowed scope as control, plus guards that at least one member actually had cross-assembly callers.

**Every new test was mutation-tested.** Six mutants, all killed:

| mutant | killed by |
| --- | --- |
| assembly-insensitive comparison | `FrameworkTypeMatchesOnlyUnderItsOwnAssembly` |
| drop the own-assembly TypeDef check | `TargetAssemblyItselfIsKept` |
| no-op filter (never rule out) | 7 tests, incl. the non-vacuity guard |
| over-narrow (rule out everything) | all 3 CLI tests |
| narrow set leaked into the shared `_callerScopes` | `CallerScopes_ForTheGraphIsNotNarrowed` |

Two of those found defects **in my own tests** rather than in the product:

1. `SameNamedTypeInAnotherAssemblyIsRuledOut` **survived** the assembly-insensitive mutant, so it does not pin assembly sensitivity as its comment claimed (the lookalike has no such TypeRef row at all). Comment corrected to say what it actually pins and to name the test that does.
2. `CallerScopes_ForTheGraphIsNotNarrowed` originally asked the **allocations** lens, which reads the *other* cache — so it could not observe a narrow set leaking into the graph, which is the single hazard it exists to gate. Switched to `includeAllocations: false` and re-verified: the leak mutant now kills it, and kills nothing else.

Two of my Analysis test failures were also the tests, not the product: `SelectCouldReach` closes over the supplied *scope* (so an indirect candidate needs its bridge present), and `List<T>` resolves to assembly **`System.Collections`**, not corelib — `System.Object` is the genuine corelib-facade case. The second became a stronger test rather than a weakened one.

## Validation

- Build `0 Error(s)`.
- Analysis **571 / 1** — only the known #3400 `CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition`.
- CLI **2262 / 1 / 10 skipped** — only the known `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`, re-confirmed with a **committed** tree (an uncommitted tree suppresses SourceLink and produces the same failure for a different reason).

### DIFF
--- a/src/ILInspector.Analysis.Tests/CallerScopeTypeFilterTests.cs
+++ b/src/ILInspector.Analysis.Tests/CallerScopeTypeFilterTests.cs
@@ -321,10 +321,18 @@ public class CallerScopeTypeFilterTests
     /// produces. Only the metadata blob is built, because <see cref="CallerScopeTypeFilter"/>
     /// decides from a <see cref="MetadataReader"/> and never needs the surrounding PE.
     ///
-    /// The assertions below pin the two properties that make this stand in for a compiled
-    /// netmodule, so the fixture cannot quietly stop being one: the reader is not an assembly, and
-    /// the decoder gives its definitions the empty assembly name. Both were confirmed against an
-    /// actual <c>csc -target:module</c> output before this test was written.
+    /// The corelib reference and <c>System.Object</c> base type are not decoration: a real
+    /// <c>csc -target:module</c> output carries exactly them, and a fixture without them would let
+    /// a negative test "rule out" <c>System.Object</c>, which no real module would allow. Verified
+    /// against an actual netmodule, whose entire TypeRef table is:
+    /// <code>
+    /// System.Runtime.CompilerServices.RefSafetyRulesAttribute -> asmref:System.Private.CoreLib
+    /// System.Object                                           -> asmref:System.Private.CoreLib
+    /// </code>
+    ///
+    /// The assertions in the tests below pin the two properties that make this stand in for a
+    /// compiled netmodule, so the fixture cannot quietly stop being one: the reader is not an
+    /// assembly, and the decoder gives its definitions the empty assembly name.
     /// </summary>
     static MetadataReaderProvider BuildModuleMetadata(string ns, string typeName)
     {
@@ -335,6 +343,18 @@ public class CallerScopeTypeFilterTests
             builder.GetOrAddGuid(Guid.NewGuid()),
             default,
             default);
+
+        var corelib = builder.AddAssemblyReference(
+            builder.GetOrAddString("System.Private.CoreLib"),
+            new Version(9, 0, 0, 0),
+            culture: default,
+            publicKeyOrToken: default,
+            flags: default,
+            hashValue: default);
+        var objectRef = builder.AddTypeReference(
+            corelib,
+            builder.GetOrAddString("System"),
+            builder.GetOrAddString("Object"));
 
         // Row 1 is always <Module>; the real type follows it, exactly as a compiler emits.
         builder.AddTypeDefinition(
@@ -348,7 +368,7 @@ public class CallerScopeTypeFilterTests
             TypeAttributes.Public,
             builder.GetOrAddString(ns),
             builder.GetOrAddString(typeName),
-            baseType: default,
+            baseType: objectRef,
             fieldList: MetadataTokens.FieldDefinitionHandle(1),
             methodList: MetadataTokens.MethodDefinitionHandle(1));
 
@@ -391,6 +411,12 @@ public class CallerScopeTypeFilterTests
     /// <summary>
     /// The same module must still be ruled out for a type it does not declare, or the test above
     /// would be satisfied by a filter that simply kept every module.
+    ///
+    /// The foreign type has to be one a real module would genuinely not name. <c>System.Object</c>
+    /// is the wrong choice and was the original mistake here: every compiled module references it
+    /// as a base type, so ruling it out would have been a claim no real netmodule could satisfy —
+    /// the assertion only held because the fixture was less faithful than the thing it stood for.
+    /// It is now asserted in the other direction, as a positive control.
     /// </summary>
     [Fact]
     public void ModuleWithoutAnAssemblyManifestIsStillRuledOutForAForeignType()
@@ -400,6 +426,14 @@ public class CallerScopeTypeFilterTests
 
         Assert.Equal(
             CallerScopeTypeFilter.TypeReferenceState.DoesNotName,
+            CallerScopeTypeFilter.Classify(
+                reader,
+                TypeRef.Definition("System.Net.Sockets", "System.Net.Sockets", "Socket")));
+
+        // Positive control: the type the module really does reference is found through the
+        // TypeRef table, so the negative above is a property of this type and not of modules.
+        Assert.Equal(
+            CallerScopeTypeFilter.TypeReferenceState.Names,
             CallerScopeTypeFilter.Classify(reader, TypeRef.CoreLib("System", "Object")));
     }
 
--- a/src/ILInspector.Analysis.Tests/CallerScopeTypeFilterTests.cs
+++ b/src/ILInspector.Analysis.Tests/CallerScopeTypeFilterTests.cs
@@ -1,0 +1,315 @@
+using System.Collections.Immutable;
+using System.Reflection.PortableExecutable;
+using DotnetInspector.Fixtures;
+using ILInspector.Analysis;
+
+namespace ILInspector.Analysis.Tests;
+
+/// <summary>
+/// <see cref="CallerScopeTypeFilter"/> decides, from an assembly's <c>TypeRef</c> table alone,
+/// whether it could contain a <em>direct</em> caller of a member declared by a given type. Its
+/// correctness obligation is the same shape as <see cref="CallerScopeFilter"/>'s and no weaker: it
+/// must never rule out an assembly whose call sites
+/// <see cref="MemberPattern.MatchesCrossAssembly"/> would have matched.
+///
+/// The tests that matter are the ones distinguishing it from its sibling. <see cref="CallerScopeFilter"/>
+/// asks whether an assembly could reach the target <em>assembly</em>, transitively; this asks
+/// whether it names the target <em>type</em>, directly. Every case below where the two disagree is
+/// a real compiled fixture, because the saving this filter exists for is exactly the size of that
+/// disagreement.
+/// </summary>
+public class CallerScopeTypeFilterTests
+{
+    static string Target => FixtureCatalog.AnalysisCallerGraphTarget.AssemblyPath();
+    static string Caller => FixtureCatalog.AnalysisCallerGraphCaller.AssemblyPath();
+    static string Twin => FixtureCatalog.AnalysisCallerGraphCallerTwin.AssemblyPath();
+    static string Indirect => FixtureCatalog.AnalysisCallerGraphIndirectCaller.AssemblyPath();
+    static string Lookalike => FixtureCatalog.AnalysisCallerGraphLookalikeCaller.AssemblyPath();
+
+    /// <summary>
+    /// The declaring type of a real method in the target fixture, taken from the index rather than
+    /// spelled by hand so the expected value is the one the matcher actually compares against.
+    /// </summary>
+    static TypeRef DeclaringTypeOf(string typeName, string methodName)
+    {
+        var index = LibraryBodyIndex.Open(Target);
+        var method = index.Methods.First(m => m.DeclaringType.Name == typeName && m.Name == methodName);
+        return GenericMemberIdentity.OpenDeclaringType(method.DeclaringType);
+    }
+
+    static CallerScopeTypeFilter.TypeReferenceState Classify(string candidatePath, TypeRef declaringType)
+        => CallerScopeTypeFilter.Classify(candidatePath, declaringType);
+
+    /// <summary>
+    /// Whether <see cref="CallerScopeFilter"/> — the transitive, assembly-level sibling — keeps a
+    /// candidate. Used to assert that this filter is strictly narrower on the same input, which is
+    /// the entire reason it exists.
+    ///
+    /// The whole scope is supplied, not just the candidate under test: the closure closes over the
+    /// scope rather than the machine, so an assembly whose only bridge to the target is another
+    /// scope member is only kept when that member is present.
+    /// </summary>
+    static bool AssemblyFilterKeeps(string candidatePath)
+    {
+        static ILInspector.Metadata.AssemblyIdentityNames Identity(string path)
+        {
+            using var stream = File.OpenRead(path);
+            using var peReader = new PEReader(stream);
+            return ILInspector.Metadata.AssemblyIdentityScanner.Scan(peReader);
+        }
+
+        string[] scope = [Caller, Twin, Indirect, Lookalike];
+        var candidates = scope
+            .Select(path =>
+            {
+                var identity = Identity(path);
+                return CallerScopeFilter.Candidate.Known(identity.Name, identity.ReferenceNames);
+            })
+            .ToArray();
+
+        var selected = CallerScopeFilter.SelectCouldReach(Identity(Target).Name, candidates);
+        return selected[Array.IndexOf(scope, candidatePath)];
+    }
+
+    // The real caller calls Target.Api.Ping, so its TypeRef table names Target.Api and it must be
+    // kept. This is the assembly the caller-edge tests prove does contribute edges.
+    [Fact]
+    public void AssemblyNamingTheDeclaringTypeIsKept()
+    {
+        Assert.Equal(
+            CallerScopeTypeFilter.TypeReferenceState.Names,
+            Classify(Caller, DeclaringTypeOf("Api", "Ping")));
+    }
+
+    // THE case this filter exists for. The twin references the target assembly and calls
+    // Target.Api.Ping, so the assembly-level filter keeps it for every target in that assembly.
+    // But it never mentions Target.Box`1, so no call site in it can produce a callee whose
+    // declaring type matches Box`1, and opening it to discover that is pure cost.
+    //
+    // The two assertions are one claim: the filters disagree, and the disagreement is the saving.
+    [Fact]
+    public void AssemblyNamingTheTargetAssemblyButNotTheTypeIsRuledOut()
+    {
+        Assert.True(AssemblyFilterKeeps(Twin));
+
+        Assert.Equal(
+            CallerScopeTypeFilter.TypeReferenceState.DoesNotName,
+            Classify(Twin, DeclaringTypeOf("Box`1", "Store")));
+    }
+
+    // Same assembly, same scan, a type it does reference: without this the test above would pass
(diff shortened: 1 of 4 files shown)


richlander:

Two independent fixed-head reviews at `79e272dc`, in isolated worktrees: **Gemini 3.1 Pro** and **GPT-5.6**. Fixes are in `4d3c29d6`. Both are now re-reviewing the fixed head.

### GPT-5.6 — Blocking: netmodule `TypeDef` contradicts the matcher &mdash; **valid, fixed**

`Classify` checked the candidate's own identity only under `if (reader.IsAssembly)`. But `TypeRefDecoder` gives definitions in an image with **no assembly manifest** the empty assembly name ([`TypeRefDecoder.cs:87-89`](../blob/79e272dc/src/ILInspector.Analysis/TypeRefDecoder.cs#L87-L89)), and `MatchesCrossAssembly` compares that name like any other. So a module was ruled out **even for a type it declares itself** — exactly the unsoundness this filter must not have.

Reproduced rather than taken on faith: restoring the old guard now fails the new test.

```
ILInspector.Analysis.Tests.CallerScopeTypeFilterTests.ModuleWithoutAnAssemblyManifestIsNotRuledOutForItsOwnTypes [FAIL]
   Total: 18, Failed: 1
```

Reachability is not merely theoretical — the product already treats a bare netmodule as real input (`OptionalArgumentFacts.cs:172`, *"a bare netmodule input (reader is not an assembly)"*).

**The root cause is worth naming: the branch was a restatement.** It recomputed by hand what the decoder already derives. That is precisely the failure mode the rest of this filter was designed to avoid — the TypeRef scan reuses `TypeRefDecoder` so it *cannot* drift, and the one place I hand-wrote the identity is the one place that drifted. The fix removes the restatement rather than widening it: the branch now asks `TypeRefDecoder.Instance.GetTypeFromDefinition` for one of the image's own definitions.

Verified against a real `csc -target:module` artifact, not just synthetic metadata:

```
PROBE IsAssembly=False
PROBE typedef .<Module>   assembly='' kind=Definition
PROBE typedef .ModTarget  assembly='' kind=Definition
PROBE filter=Names
```

### GPT-5.6 — Important: two documented branches were unguarded &mdash; **valid, both fixed**

Both were real gaps in *my* mutation coverage:

1. **`Unsupported` TypeRef row.** Changing `return Undecidable` to `continue` killed no test. Now gated by `AnUndecodableTypeReferenceRowLeavesTheCandidateUndecided`, which builds a resolution-scope chain longer than `MetadataSafetyPolicy.MaxRelationshipNodes` and asserts as a *premise* that the decoder really does return `Unsupported`, so the test cannot pass for a filter that never returns `Undecidable` at all. The `continue` mutant now kills it.
2. **Per-declaring-type cache.** Collapsing both cache keys to a constant killed no test. Now gated by `DirectCallerScopes_AreCachedPerDeclaringType`, which resolves two declaring types in **both orders** with a non-vacuity assertion that they genuinely select different assemblies. The collapsed-key mutant now kills it.

### Gemini 3.1 Pro — Clean

Graded Clean on all eight attack points. I reproduced its load-bearing result independently and it holds: in a real forwarding scenario (`Caller` &rarr; facade `A` &rarr; `B`), the Caller's `TypeRef` for the forwarded type resolves to **`A`**, the filter says `DoesNotName`, and the matcher finds **0** matches. They agree, so the filter is sound relative to the matcher.

Two caveats I'm recording rather than glossing:

- **Its forwarding "test" contained no assertions** — only `Console.WriteLine`. It could not have failed. That is why I rebuilt and re-ran the scenario myself before accepting it.
- **5 of its 8 attack points were reasoned, not executed** — and the Blocking bug lived in the area it marked clean by reasoning (attack point 5, ModuleRef/nil scopes). I closed the highest-risk reasoned-only item myself by code inspection: `LibraryBodyIndex` opens candidates with the identical `File.OpenRead` share mode (`:300`, `:1410`, `:2212`) and also returns null on `!HasMetadata` (`:2214`), so `Classify` returning `DoesNotName` for an unopenable or metadata-less image cannot drop a candidate the session could otherwise have used.

### Non-actions, stated explicitly

**Type forwarding is not resolved by the matcher either**, so forwarded callers were already invisible in `Callers` before this PR. This change neither introduces nor worsens that. I am deliberately **not** fixing it here — it is a pre-existing matcher limitation with its own blast radius, and folding it into a performance change would hide it. Filing as non-blocking follow-up.

### Validation at `4d3c29d6`

- Analysis **574 / 1**, CLI **2263 / 1 / 10 skipped** — only the two known pre-existing failures (#3400, and the SourceLink discovery test, re-confirmed against a committed tree).
- Nine-case matrix re-run at the fixed head: **all IDENTICAL**, memory wins intact (685&rarr;98 MB, 724&rarr;315 MB).
- Six mutants from round 0 plus three new ones, all killed, each by the test whose claim it is.

richlander:
## Review reconciliation — round 2, and readiness

Both reviewers are now **clean at the fixed head**.

### GPT-5.6 — Important: the netmodule fixture was not compiler-faithful &mdash; **valid, fixed**

My new negative test asserted the module was ruled out for `System.Object`. That held **only because my synthetic fixture had no TypeRef table at all** — a real module *always* references `System.Object` as a base type, so it was a claim no real netmodule could satisfy. GPT proved it by making the fixture faithful and watching the assertion flip to `Names`.

I verified the premise against the real artifact before fixing. The complete TypeRef table of an actual `csc -target:module` output:

```text
System.Runtime.CompilerServices.RefSafetyRulesAttribute -> asmref:System.Private.CoreLib
System.Object                                           -> asmref:System.Private.CoreLib
```

Fixed in `9504b2e3`:

- `BuildModuleMetadata` now carries the corelib AssemblyRef, a `System.Object` TypeRef, and uses it as the base type.
- The negative case uses `System.Net.Sockets.Socket`, which a module genuinely would not name.
- `System.Object` is asserted in the **opposite** direction as a positive control, so the negative is a property of the chosen type rather than of modules — and the control independently detects a broken TypeRef scan.
- Re-verified that restoring the `reader.IsAssembly` guard **still** kills the positive test against the now-faithful fixture, so the fidelity fix did not weaken what it pins.

GPT confirmed at `9504b2e3` with three mutants, each failing exactly one test:

| mutant | result |
| --- | --- |
| old `reader.IsAssembly` guard | own-type test dies |
| TypeRef match `return Names` &rarr; `continue` | `System.Object` control dies |
| keep every module | `Socket` negative dies |

> The fixture is faithful for all metadata relevant to these tests. It omits the compiler's `RefSafetyRulesAttribute` TypeRef, but that cannot affect own-definition, `System.Object`, or `Socket` classification.

### Gemini 3.1 Pro — clean on re-review, this time by execution

Its first pass graded Clean and **missed the Blocking bug**, having reasoned rather than executed on the adjacent attack point. Asked to execute, it produced the two results I most wanted:

- **Every definition in an image carries the same assembly identity** — decoded *every* `TypeDefinition` in all 182 framework DLLs: `Tested: 169, Match: 169, Mismatch: 0`. This is the claim that licenses decoding only the first row.
- **No regression on the common path** — old `CanonicalAssembly(assemblyDefName)` vs new `decoder(firstTypeDef).Assembly` over the same 169 managed assemblies: **169/169 exact matches**.
- Zero-TypeDef image falls through to the TypeRef scan without throwing; generic arity distinguishes `` List`1 `` from `` List`2 ``.

### Final state

`7524513d` also updates the class doc, which had described the TypeDef branch as canonicalizing an assembly's own name — precisely what was wrong with it. It now records that the identity comes from the decoder and why deriving it by hand failed.

- Analysis **574 / 1**, CLI **2263 / 1 / 10 skipped** — only the two known pre-existing failures (#3400; and the SourceLink discovery test, re-confirmed against a committed tree).
- Nine-case matrix at the fixed head: **all IDENTICAL**, memory wins intact (685&rarr;98 MB, 724&rarr;315 MB).
- **Ten mutants across three rounds, all killed**, each by the test whose claim it is.

### Non-action

Type forwarding is filed as **#3419** and deliberately not fixed here: the matcher does not resolve it either, so forwarded callers were already invisible. That issue records that **if the matcher is ever taught forwarding, this filter must widen in the same change**, or a display gap becomes a wrong answer.

Ready to merge

richlander:
## Merged `origin/main` (`a2877f1c`) into the branch — `21246cd4`

Main had advanced two commits (#3415, the `--rows` range grammar). Merged rather than rebased,
since this branch is public and reviewed.

**The merge is orthogonal to the review.** All four files this PR touches are **byte-identical**
across it:

```text
git diff --stat 7524513d HEAD -- \
  src/ILInspector.Analysis/CallerScopeTypeFilter.cs \
  src/dotnet-inspect/Inspectors/ApiMemberAnalysisInspection.cs \
  src/ILInspector.Analysis.Tests/CallerScopeTypeFilterTests.cs \
  src/dotnet-inspect.Tests/ApiMemberAnalysisInspectionTests.cs
(no output)
```

Both fixed-head reviews at `7524513d` (Gemini 3.1 Pro clean, GPT-5.6 clean) therefore remain
valid at the merged head; no re-review is required for the merge itself.

### Validation at `21246cd4`

- `dotnet build dotnet-inspect.slnx -c Release` — `Build succeeded`, `0 Error(s)`.
- `ILInspector.Analysis.Tests` — **574 / 1**, the only failure being the pre-existing
  `CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition` (#3400).
- `dotnet-inspect.Tests` — **2315 / 1 / 10 skipped**, the only failure being the pre-existing
  `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`.

Both at baseline. Still **Ready to merge**.

PR #3415, opened 2026-07-28
Add the --rows range grammar as a parsed value type

richlander:
First half of the #3364 redesign. **Grammar only** -- `RowSpec` is not referenced by any product path yet (`grep` finds it in its own file and its test file, nowhere else), so no user-visible behaviour changes. Wiring it to the flag is the next PR.

## The grammar

| Form | Meaning | Rows |
| --- | --- | --- |
| `6` | a count, like `head -n` | six rows |
| `2..10` | inclusive range | **nine** rows |
| `2+10` | start plus count | **ten** rows |
| `10..` | from row ten to the end | depends on the section |

`6` and `6..` both mention six and mean different things: the first is a quantity, the second a position. That is inherent to offering a count and a range through one flag, so the two parse into different `RowSpecKind`s rather than into one normalized shape -- the caller that must reject a direction applied to an absolute range has to be able to tell them apart.

## Two decisions worth review

**The colon form is rejected with an explanation, not a generic parse error.** `2:10` carries Python slice heritage (0-based, end-exclusive), so for identical digits it names a different set of rows than `2..10` by one row at each edge. Failing it silently, or failing it without saying why, leaves the user to rediscover that from the output.

**The grammar is closed.** Anything outside the four forms is rejected rather than guessed at, including a leading sign -- parsing `2..-1` and range-checking it later would turn a malformed range into a silently empty one. Start-plus-count is computed in `long`, so a start plus count landing past `int.MaxValue` is reported rather than wrapping into a negative end that selects nothing.

## Evidence

35 tests. They assert **resolved extents**, not merely that parsing succeeded: a parser that accepted all four forms while conflating two of them would pass a "parses without error" suite and still select the wrong rows.

Non-vacuity checked with two mutations:

| Mutation | Result |
| --- | --- |
| make `..` end-exclusive (`FromRange(start, end - 1)`) | **6 fail**, including inclusivity, single-row, same-digits, and `ToString` round-trip |
| off-by-one in `+` (`start + count` instead of `start + count - 1`) | **3 fail**, including `PlusRange_IsStartPlusCount` and same-digits |

## Suite status, including something that needs its own issue

Full suite: **2309 tests, 1 failure** -- `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`, which reproduces on unmodified `origin/main` locally and does not occur in CI at all.

Two *other* tests failed intermittently across my runs, and I could not attribute them to this change:

| Run | Failures |
| --- | --- |
| clean `main`, files removed | 1 |
| with this change | 3 (`WriteTips_WithMinimalLevel_WritesTips`, `DiffCommandTests.ExecuteAsync_AllocationFindingTransitions_ComposesJson`) |
| with this change | 3 (`WriteTips_...`, `DiffCommandTests.BuildApiDiff_TypeFilterWithNoChangedTypesAndMemberFilter_ReportsTypeFilter`) |
| with this change | 1 |

The `DiffCommandTests` failure was a **different test each time**, and both classes pass in isolation (295 tests, 0 failures). Since `RowSpec` is called by no product path, the only mechanism available to it is shifting test scheduling -- which points at pre-existing shared state or a parallelism race in the suite that adding 35 tests perturbs. I have not tried to fix that here; it wants its own issue and is not this PR's to carry.

## Not in this PR

The flag wiring, `-r`, the `--head`/`--tail` direction split, and the `output-shapes.md` rewrite. The open compatibility question -- whether bare `--rows` keeps its current meaning ("interpret `-n` as data rows", load-bearing in ~70 test call sites, `README.md`, and four skill docs) -- is settled there, not here.

### DIFF
--- a/src/dotnet-inspect.Tests/RowSpecGrammarTests.cs
+++ b/src/dotnet-inspect.Tests/RowSpecGrammarTests.cs
@@ -1,0 +1,204 @@
+using DotnetInspector.Output;
+
+namespace DotnetInspector.Tests;
+
+/// <summary>
+/// Pins the <c>--rows</c> grammar.
+///
+/// The central claim is that identical digits mean different things in different
+/// forms -- <c>2..10</c> is nine rows and <c>2+10</c> is ten -- so the tests
+/// assert the resolved extent rather than just that parsing succeeded. A parser
+/// that accepted every form but conflated two of them would satisfy a
+/// "parses without error" suite while still selecting the wrong rows, which is
+/// the failure mode this flag exists to remove.
+/// </summary>
+public class RowSpecGrammarTests
+{
+    [Fact]
+    public void BareNumber_IsACount_NotARange()
+    {
+        Assert.True(RowSpec.TryParse("6", out var spec, out var error));
+        Assert.Null(error);
+
+        Assert.Equal(RowSpecKind.Count, spec.Kind);
+        Assert.Equal(6, spec.Count);
+        Assert.False(spec.IsRange);
+
+        // A count has no absolute extent until a direction and a section supply
+        // one, so it deliberately does not answer Contains.
+        Assert.False(spec.Contains(1));
+        Assert.Null(spec.RowCount);
+    }
+
+    [Fact]
+    public void DotDotRange_IsInclusive_AtBothEnds()
+    {
+        Assert.True(RowSpec.TryParse("2..10", out var spec, out var error));
+        Assert.Null(error);
+
+        Assert.Equal(RowSpecKind.Range, spec.Kind);
+        Assert.Equal(2, spec.Start);
+        Assert.Equal(10, spec.End);
+
+        // Inclusive at both ends: 2 through 10 is nine rows, not eight.
+        Assert.Equal(9, spec.RowCount);
+        Assert.True(spec.Contains(2));
+        Assert.True(spec.Contains(10));
+        Assert.False(spec.Contains(1));
+        Assert.False(spec.Contains(11));
+    }
+
+    [Fact]
+    public void PlusRange_IsStartPlusCount()
+    {
+        Assert.True(RowSpec.TryParse("2+10", out var spec, out var error));
+        Assert.Null(error);
+
+        Assert.Equal(RowSpecKind.Range, spec.Kind);
+        Assert.Equal(2, spec.Start);
+
+        // Ten rows starting at 2 ends at 11, not 12.
+        Assert.Equal(11, spec.End);
+        Assert.Equal(10, spec.RowCount);
+        Assert.True(spec.Contains(11));
+        Assert.False(spec.Contains(12));
+    }
+
+    [Fact]
+    public void SameDigits_SelectDifferentRows_AcrossTheTwoRangeForms()
+    {
+        Assert.True(RowSpec.TryParse("2..10", out var inclusive, out _));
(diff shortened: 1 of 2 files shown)


richlander:
Filed the intermittent-failure observation as #3416 so it is tracked rather than buried in this PR's description. It is unrelated to this change and not this PR's to fix -- `RowSpec` is called by no product path, so the only mechanism it has is perturbing test scheduling. The diagnostic detail is that a **third** run of the identical binary produced 1 failure rather than 3, and a different `DiffCommandTests` test failed in each of the two 3-failure runs.

richlander:
## Review reconciliation

One review (MAI-Code), risk-sized: the change is additive and unwired — `RowSpec` is referenced by no product path — so it sits in the "more than trivial, but not high risk" tier.

### Finding: `Contains` answered a question a count spec cannot answer (Medium) — **accepted, and extended**

`Contains` returned `false` for every row of a count spec. That is wrong in the one way a caller cannot detect: `false` is exactly what a genuine miss looks like, so code filtering rows through it before checking `Kind` would silently select nothing. That is the success-shaped empty output `AGENTS.md` forbids, so I treated it as a defect rather than an API-taste question.

I applied the same argument to every kind-specific member:

- `Contains`, `Start`, `End` now throw `InvalidOperationException` for a count.
- `Count` now throws for a range.
- The message names the member, the kind required, and the kind found, and tells the caller to check `Kind` first.
- `ToString` and `IsOpenEnded` read backing fields directly, so formatting never trips a guard.

Two new tests pin the throws, and the assertion that previously pinned the trap is gone.

### `RowCount` moved the *opposite* way — deliberately

`RowCount` returned `null` for a count, which understated what a count knows: `6` selects six rows whichever direction anchors it. What a count does not know is *which* rows, not how many. So `RowCount` now reports the count and is `null` only for a genuinely open-ended `N..`.

I asked MAI to attack that asymmetry specifically — including whether reporting a count invites a caller to treat it as an absolute extent. It did not sustain an objection.

### Fixed-head re-review

`7d4f2dbf` came back **NON-BLOCKING**. It verified rather than assumed the structural risk I flagged: the compiler-synthesized equality and `GetHashCode` on the `readonly record struct` compare *fields*, not properties, so the throwing getters do not break them. It also confirmed `IsRange`/`IsOpenEnded`/`Kind`/`ToString`/factories return nothing misleading for either kind, and re-ran both grammar mutations from its first pass to confirm the parser is untouched.

### Evidence

- 37 tests in `RowSpecGrammarTests`, asserting **resolved extents** rather than parse success — `2..10` is nine rows and `2+10` is ten, from identical digits.
- Non-vacuity proven by mutation: making `..` end-exclusive fails 6 tests; an off-by-one in `+` fails 3. Both reverted.
- Full suite at this head: 2311 tests, 1 failure — `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`, which reproduces on unmodified `main` locally and **passes in CI on ubuntu-26.04**, so it is a Windows-local artifact, not a baseline break.
- The intermittent `WriteTips`/`DiffCommandTests` ordering noise I saw while validating is tracked separately as #3416 and is not caused by this PR.

### Scope boundary

This PR lands the grammar only. `--rows` still takes no value; nothing parses a `RowSpec` yet. Wiring it to the flag — including the decision that **bare `--rows` does not survive** — is the follow-up, and is a user-visible breaking change that will carry two-model review.

Ready to merge

Issue #3416, opened 2026-07-28
Suite has an order-dependent flake: WriteTips and DiffCommandTests fail intermittently, pass in isolation


richlander:
Two tests fail intermittently in full-suite runs and pass in isolation. The failing member of the pair varies between runs, which points at shared state or a parallelism race rather than at either test being individually wrong.

## Observed

Measured on Windows at `origin/main` = `92a20510` plus an additive branch (#3415) that adds 35 tests and **no product-path code** -- `RowSpec` is referenced only by its own file and its test file, so it cannot execute during any other test.

| Run | Tests | Failures |
| --- | --- | --- |
| `main`, branch files removed | 2274 | 1 |
| branch | 2309 | 3 -- `WriteTips_WithMinimalLevel_WritesTips`, `DiffCommandTests.ExecuteAsync_AllocationFindingTransitions_ComposesJson` |
| branch | 2309 | 3 -- `WriteTips_WithMinimalLevel_WritesTips`, `DiffCommandTests.BuildApiDiff_TypeFilterWithNoChangedTypesAndMemberFilter_ReportsTypeFilter` |
| branch | 2309 | 1 |

The one persistent failure is `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceAndHidden`, which is separate, reproduces on unmodified `origin/main`, and does not occur in CI.

In isolation both classes are clean:

```
dotnet run --project src\dotnet-inspect.Tests -c Release --no-build -- \
  -class DotnetInspector.Tests.CommandLineTests -class DotnetInspector.Tests.DiffCommandTests
=> Total: 295, Errors: 0, Failed: 0
```

## Why this is not caused by the added tests

The added tests call only `RowSpec.TryParse`, a pure parser over a string. There is no reference from any product path (verified by grep across `src`). The only mechanism available to them is changing how the rest of the suite is scheduled, which is what makes the third run -- 1 failure with the identical binary -- diagnostic: the same code produced 3 failures and then 1.

A **different** `DiffCommandTests` test failed in each 3-failure run, which is hard to explain by anything other than nondeterministic ordering or concurrent access to shared mutable state.

## Suspects worth checking

- `ArgumentPreprocessor` holds static mutable state (`HeadLines`, `TailLines`) with an `internal static Reset()` "used for testing" (`src/dotnet-inspect/CommandLine/Helpers/ArgumentPreprocessor.cs`). Static parse state plus xUnit's parallel collections is the classic shape for this.
[...]

PR #3424, opened 2026-07-29
Wire --rows to the range grammar; split count, unit, and direction onto three flags

### LINKED ISSUE #3364: Rework row selection: --row indexes an invisible sequence

richlander:
Closes #3364.

Wires the `RowSpec` grammar from #3415 into the renderers and splits the three
concerns that `--rows` used to carry on two flags onto three flags.

## The change

Before, `--rows` was a **bool** meaning "interpret `-n`/`--head`/`--tail`
as data rows rather than output lines". The count lived on one flag and the unit
on another, and the direction rode on the count flag. `--rows -n 20 --tail`
is three concerns braided across three tokens that do not say so.

After, each axis has exactly one flag:

| Axis | Flag |
| --- | --- |
| unit | `--rows` (data rows) vs `-n` (output lines) |
| count / selection | the `--rows` value, or the `-n` value |
| direction | `--head` / `--tail` (bools) |

`--rows` now takes a `RowSpec`, so all four grammar forms reach the
renderers:

```text
--rows 6          first six data rows      (--rows 6 --tail = last six)
--rows 2..10      rows 2 through 10 inclusive -- nine rows
--rows 2+10       ten rows starting at row 2
--rows 10..       row 10 through the last
```

## Breaking changes

These are deliberate and each is reported with an actionable message:

- **`--head N` / `--tail N` no longer carry the count.** Use `-n N --tail`
  (lines) or `--rows N --tail` (rows). A stale spelling is *detected and
  named* rather than silently binding `N` as a positional -- otherwise
  `--tail 20` goes looking for a package named "20".
- **`--head` is no longer an alias of `-n`.** It is a direction.
- **Bare `--rows` is an error.** It used to mean "interpret `-n` as rows",
  which is exactly the count-on-a-different-flag-than-the-unit problem.

## Design notes

**Why `--head`/`--tail` are bools.** An optional-valued (`ZeroOrOne`)
option swallows the following positional -- measured, not assumed:
`dotnet-inspect library -S <path.dll>` bound the DLL to `-S`. Bools consume
no tokens, so `--tail <path>` is unambiguous. Verified directly against the
built CLI.

**One interpretation point.** `RowWindow` was a positional
`(int Count, bool FromEnd)` record whose two consumers each open-coded the
skip/take. It is now private fields behind named factories (`Head`/`Tail`/
`Range`) plus a single `Resolve(int dataCount) -> (KeepStart, KeepEnd)`, so a
range cannot land in one renderer and miss the other. `Resolve` guarantees
`0 <= KeepStart <= KeepEnd <= dataCount`, so callers never re-clamp and a
start past the end yields an empty rather than an inverted window --
`Take(end - start)` would throw on an inverted pair.

**A count and a range are different kinds.** `--rows 2..10 --tail` is rejected
rather than silently resolved: a count anchors to an end, a range does not.

## Defects found and fixed while building this

- **Bare `--rows` printed a stack trace *and* exited 0** -- a failure
  invisible to any caller checking the exit code. The validator called
  `GetValue(Rows)` on a required-argument option with no value, which throws
  from inside the validator before System.CommandLine can report the missing
  argument. It now reads raw tokens via `GetResult(Rows)`.
- **`--rows --tsv` blamed `--tsv`.** SCL hands a required-argument option the
  next token even when it is plainly another option, so the row spec arrived as
  the literal `"--tsv"`. A leading-dash check now reports the missing
  selection instead.
- **The migration guard was stranded in `Program.cs`**, which
  `CommandExecutionTests.RunAppAsync` does not execute -- product behavior no
  test could reach. It moved into
  `ArgumentPreprocessor.TryGetStaleDirectionFlagError`, and both `Program.cs`
  and the harness call that one product method.

## Evidence

- New `RowWindowResolutionTests` (18 tests) gates the window semantics:
  extents, clamping, the ordering invariant, unlimited handling, and the
  count-vs-range disagreement.
- `dotnet build dotnet-inspect.slnx -c Release` clean.
- Full CLI suite: **2346 tests, 3 failures**, all accounted for:
  - `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceLinkDoor` and
    `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically` both
    **reproduce on unmodified `origin/main`** (verified in a detached
    baseline worktree at 97a5de04).
  - `PerformanceTriagePredicates_DoNotAlterDiscovery` is the order-dependent
    flake tracked by #3416; passes in isolation.
- Manual CLI verification against `System.Text.Json` of all four grammar
  forms, six error paths, the three line modes (`-n 5`, `-n 5 --tail`,
  `-5`), and the positional-binding hazard (`--tail <path>`,
  `--head <path>`, `--rows 3 <path>` all bind the positional correctly).

Docs, skills, and README migrated off the old spellings; a repo-wide grep for
`--rows -n`, `--head N`, and `--tail N` is clean apart from the migration
test and its documentation.


richlander:
## Review reconciliation

High-risk tier (user-visible breaking change, wide blast radius), so two reviews from two different model families, neither my own. Both reviewed in isolated linked worktrees detached at a fixed head, and both re-reviewed at the fixed head after the fixes.

### Findings and resolution — all in `1e19a98f`

**MAI-Code — migration guard scanned past `--` (medium).** `TryGetStaleDirectionFlagError` inspects raw tokens before parsing, and it did not stop at the end-of-options separator. `dotnet-inspect library -- --tail 5` reported `'--tail 5' is no longer valid` for a token that was never a flag. **Reproduced independently** on a clean exact-head worktree, and confirmed it was a behavior change rather than a pre-existing quirk: `origin/main` answers `Unrecognized command or argument '5'`. Fixed by stopping the scan at `--` (the `--rows` row-mode probe that selects the suggested replacement is bounded the same way). After `--` everything is positional, so the stale spelling cannot be what the user meant.

**Gemini — row limiting silently rewrote line endings (medium).** Both limiters normalized CRLF to LF and never restored it, so passing a row window changed every line ending on Windows. The unlimited fast path returns the string untouched, so identical content differed byte for byte depending on whether a window was supplied — and `--rows 1..`, which selects every row, looked like a no-op but was not.

I verified this with raw byte redirection through `cmd.exe` rather than PowerShell, since PowerShell redirection rewrites line endings and cannot measure the property. **This predates the PR** — `--rows -n 500` on `origin/main` also emits bare LF — but it lives in the code this PR rewrote and `1..` made it reachable through a spelling that reads as a no-op, so it is in scope. Fixed via `MarkdownScan.DetectNewline`; both limiters rejoin with the ending the text arrived with. `--rows 1..` is now byte-identical to passing no flag (26775 bytes, 256 CRLF, 0 bare LF, `SequenceEqual` true).

**Gemini — dead clamp made a claim it did not enforce (non-blocking).** The `Math.Max(start, end)` in `RowWindow.Resolve` was unreachable: the `Range` factory rejects `end < start`, so `_end >= _start`, and `Math.Min` is monotonic. Gemini proved it by deleting the clamp and watching the test that named the property still pass. Removed, and the doc comment now states the invariant holds by construction and names `Range_RefusesAnEndBeforeItsStart` as the actual gate — rather than implying a clamp enforced it.

Nothing was dismissed. Each fix has a gate that I confirmed non-vacuous by mutating the product and watching the named test fail:

| Fix | Gate | Fails when mutated |
| --- | --- | --- |
| `--` separator | `ValuedTailFlag_AfterEndOfOptions_IsNotReportedAsAMigration` | yes |
| line endings | `LimitRenderedTableRows_PreservesTheLineEndingsItWasGiven`, `..._MarkdownPreservesTheLineEndingsItWasGiven` | yes (both) |

### Re-review at `1e19a98f` — both clean

- **MAI-Code:** "No blocking issues found... I consider the PR clean overall." Verified by execution that the guard still fires before `--` (`--tail 20`, `--head 5`, `--rows 6 --tail 20`) and no longer fires after it, including the `--`-as-value case.
- **Gemini:** "I consider the PR **clean**." Worked through `DetectNewline` against mixed endings, a lone `\r`, empty input, and no-newline input, and walked the `TrimEnd`/rejoin path to confirm the trailing ending cannot double. Gave a proof that `KeepEnd < KeepStart` is unreachable.

Both re-ran the full suite at the fixed head: **2349 tests, 2 failures**, both of which reproduce on unmodified `origin/main` at `97a5de04` (verified in a detached baseline worktree) and both of which **pass in CI on ubuntu** — they are Windows-local environment failures. The #3416 flake did not fire in either run.

### CI

All green on `1e19a98f`: `test (ubuntu-26.04)`, `build-net10`, `markdownlint`, `changes`.

### Non-blocking follow-up (not in this PR)

- **#3416** — the order-dependent flake. Gemini checked specifically and reports this PR leaves the static `HeadLines`/`TailLines` mutation profile unchanged: neither worse nor better.
- `TrimEnd('\n')` collapses a run of trailing newlines to one. Benign today because the formatters emit at most one, and it is pre-existing behavior.

richlander:
Ready to merge

PR #3423, opened 2026-07-29
Adopt NuGetFetch source into this repository

richlander:
## Stack

Slice **1 of 2** for #3417.

| Slice | PR | Branch | Base | Claim |
| --- | --- | --- | --- | --- |
| **1** | **#3423 (this PR)** | `feature/rehome-nugetfetch` | `main` | Adopt the `NuGetFetch` source, byte-identical, no behavior change |
| 2 | #3428 | `feature/3417-search-sources` | `feature/rehome-nugetfetch` | Search authenticated and non-nuget.org feeds |

This slice stands on its own: it is an import plus one additive test trait, and it is
correct and safe whether or not #3428 ever lands. Merge this one first.

**Declared residual, and therefore scope rather than defect:** every behavior question
about the imported code — authentication, search sources, error typing, and the
duplication this import exposes — belongs to slice 2 or later. Nothing here is edited
to anticipate it.

## Conclusion

Moves the `NuGetFetch` package source into this repository as `src/NuGetFetch` and
`src/NuGetFetch.Tests`, replacing the `PackageReference` with a `ProjectReference`.
All 12 source files, both trusted-root CTL bundles, and all 10 test files are
**byte-identical** to upstream. No behavior changes.

Follows the pattern established by #3367 for `SourceLinkFetch`.

The diff reads as ~22,000 added lines, but 18,012 of those are the two PEM trusted-root
bundles (`codesignctl.pem`, `timestampctl.pem`) carried over verbatim. The actual C# is
~4,200 lines, none of it new.

## Why

`NuGetFetch` is being reshaped around this repository's package layer rather than
consumed from it, so this repository is its new home. Doing the move first, with no
edits, keeps the reshaping diff honest: any later change to these files is visibly a
change, not tangled with the import.

The concrete forcing function is #3417. Its four items do not split by bug, they split
by repository: Bug 2 (`totalHits` typed `int`) is fixable only upstream, Bug 1 (401
misreported as "not found") only here, and Bug 3 plus the credential feature need
**both** — `SearchService.SearchAsync` calls bare `client.GetStreamAsync` with no auth
header and no retry, while every auth, retry, and status facility lives in
`DotnetInspector.Packages`. While the boundary stands, authenticated search is not
expressible on either side of it alone.

## What was verified

- Upstream `github.com/richlander/nuget-fetch` HEAD is `639ab178`, tagged `v0.7.0`. The
[...]

Issue #3417, opened 2026-07-28
Authenticated NuGet feeds: 401 reported as "package not found", and package search returns empty against every Azure DevOps feed


rjmurillo:
## Ask first

I would like to add support for **authenticated NuGet feeds** (Azure Artifacts / GitHub Packages), so that `dotnet inspect package` and `dotnet inspect package search` work against a private feed without a prior `dotnet package download` or `dotnet restore` to warm the global packages folder.

Before writing any code I want scope agreement, because the work splits into **three bug fixes and one feature**, and the bugs are independently valuable. Please tell me which parts you want, and whether you want them as one PR or several.

All four items below are reproduced against `0.16.0+44f2832` on Windows using **public, credential-free feeds**, so you can verify each one without an internal account.

---

## Bug 1: an HTTP 401 from a source is reported as "package not found"

`https://pkgs.dev.azure.com/dnceng/_packaging/DoesNotExistPubliclyXYZ/nuget/v3/index.json` returns **HTTP 401 Unauthorized** (Azure DevOps answers 401 rather than 404 for a feed the caller cannot see).

```console
$ dotnet-inspect package NodaTime.Testing@3.1.9 --source https://pkgs.dev.azure.com/dnceng/_packaging/DoesNotExistPubliclyXYZ/nuget/v3/index.json
Error: Package 'nodatime.testing' not found.
(exit 1)
```

The package exists; the *feed* was not readable. The message names the wrong subject, and there is no indication that authentication was the problem, which source failed, or at which phase.

Path: `HttpRetryHelper` classifies 401/403 as non-retryable and returns `null` (`src/DotnetInspector.Packages/HttpRetryHelper.cs:118-132`); `PackageExtractor` then treats the null as absence and prints "not found" (`src/DotnetInspector.Packages/PackageExtractor.cs:224-228`, `:330-347`). Source identity and failure phase are lost on the way out because the current signatures return `string?` / `List<string>?` / `bool`.

This is the defect that makes an authenticated feed look like a typo, and it is worth fixing on its own even if you want none of the rest.

Note: `--source` itself behaves correctly. It **does** replace the defaults; the command above does not silently fall back to nuget.org. A separate, much milder observation is that a package already present in `~/.nuget/packages` is returned with `Source: NuGet` and exit 0 even when the requested `--source` is unreachable, so a cache hit hides an unusable source.

## Bug 2: `package search` returns zero results against any Azure DevOps feed

[...]

PR #3428, opened 2026-07-29
Search authenticated and non-nuget.org NuGet feeds

richlander:
## Stack

Slice **2 of 2** for #3417.

| Slice | PR | Branch | Base | Claim |
| 1 | #3423 | `feature/rehome-nugetfetch` | `main` | Adopt the `NuGetFetch` source, byte-identical, no behavior change |
| **2** | **#3428 (this PR)** | `feature/3417-search-sources` | `feature/rehome-nugetfetch` | Search authenticated and non-nuget.org feeds |

**Merge #3423 first.** This PR targets its branch, so the diff below is only this slice:
47 files, +5,626. The ~3,700-line `src/NuGetFetch/` library belongs to #3423, not here.

Closes bugs **2, 3, and 4** of #3417. Bug 1 is a declared residual — see
[Boundary](#boundary-and-non-actions).

## Conclusion

`package search` could only ever search nuget.org, and when pointed at anything else it
reported success with no results. This slice makes every source-aware path actually
traverse the sources the user configured, authenticate to them the way NuGet itself
does, and say so when it cannot.

Three things land together because they are one capability:

1. **Search reaches non-nuget.org feeds** and fails loudly when it cannot.
2. **Credentials come from a NuGet credential provider**, not just from clear text in
   `NuGet.Config` — the mechanism Microsoft actually recommends.
3. **Version queries can report provenance** (`--versions-with-feed`), so a multi-feed
   answer is attributable instead of an anonymous union.

## What was broken

| # | Defect | Effect |
| --- | --- | --- |
| 2 | `SearchResponse.TotalHits` typed `int` | nuget.org sends `totalHits` as a number, Azure DevOps sends a string. Binding the field made every ADO document fail to parse. |
| 2 | `NuGetApi.GetSearchResponseAsync` swallowed `JsonException` → `null` | `SearchService` mapped null to `[]`, so a parse failure was indistinguishable from a genuine zero-result search. |
| 3 | `package search` took no source options; `NuGetSearchService` hardcoded the nuget.org URL with no credentials | No way to search any other feed, authenticated or not. |
| 4 | Zero results always exited 0 | Scripts could not tell "nothing matched" from "nothing worked". |

Together these are a single failure mode: **a hard failure rendered as a confident empty
result.**

## What changed: source traversal

- **`totalHits` is no longer modelled.** Nothing consumed it, and ADO reports `"0"`
  alongside a populated `data` array — so it is untrustworthy even when it parses.
  `Data.Count` is the real count. This sidesteps the number-vs-string split entirely
  rather than papering over it with a custom converter.
- **Search parse failures propagate.** The service/version-index readers keep their
  existing null-on-malformed behaviour; only the search path changed, because only there
  does null mean "success, nothing found".
- **`--source`, `--add-source`, `--nugetconfig` on `package search`**, reusing
  `SharedOptions.AddNuGetOptionsTo` / `ParseNuGetSourceOptions` — the same options every
  other source-aware command already takes.
- **`SearchQueryService` discovery.** `PackageExtractor`'s service-index reader is
  generalised from its hardcoded `PackageBaseAddress` to a resource-type prefix;
  `GetSearchQueryServiceAsync` is the second caller. `SearchService` gained an
  `AuthenticationHeaderValue?` and moved from `GetStreamAsync` to `SendAsync` so
  credentials can be attached. It stays a leaf — retry and telemetry remain the caller's
  job.
- **Multiple sources** are searched in order, deduped by (id, version), capped at
  `--take`. Source precedence is pinned by tests: sources are asked in configured order
  and the first feed to answer a resolution wins.

## Credential providers

The largest addition since this PR was first opened, and the one that moves the tool off
the bottom rung of NuGet's own guidance.

Before this, credentials could only come from `<packageSourceCredentials>` in
`NuGet.Config`, as `Username` + `ClearTextPassword`. Microsoft's
[guidance for authenticated feeds](https://learn.microsoft.com/nuget/consume-packages/consuming-packages-authenticated-feeds)
treats that as the discouraged fallback and points at a **credential provider** — for
Azure DevOps, [artifacts-credprovider](https://github.com/microsoft/artifacts-credprovider).

`src/NuGetFetch/Plugins/` speaks the NuGet cross-platform plugin protocol directly
(NDJSON over stdio) rather than shelling out to any one vendor's provider, so AWS
CodeArtifact and other conforming providers are covered by the same path:

| File | Role |
| --- | --- |
| `PluginMessages.cs` | Wire contract |
| `PluginConnection.cs` | Handshake, request/response correlation, `Progress` timeout extension |
| `PluginDiscovery.cs` | All three discovery routes |
| `PluginCredentialProvider.cs` | `GetAuthenticationCredentials`, retry semantics |
| `PluginAuthenticationHandler.cs` | The 401 loop, as a `DelegatingHandler` |
| `ICredentialSource.cs` | Seam so the handler does not hard-depend on the plugin path |

Three points worth a reviewer's attention:

**Discovery has three routes, and the newest one is the easiest to miss.** Verified
empirically against a live private feed by A/B on `PATH` alone, with `~/.nuget/plugins`
absent and `NUGET_PLUGIN_PATHS` unset:

1. `NUGET_PLUGIN_PATHS` — explicit override, wins outright.
2. `~/.nuget/plugins/netcore/<Name>/<Name>.dll` — what `installcredprovider.sh` populates.
3. **A dotnet tool named `nuget-plugin-*` on `PATH`** — what
   `dotnet tool install --global Microsoft.Artifacts.CredentialProvider.NuGet.Tool`
   produces. Scanning only `~/.nuget/plugins` would have found nothing on a machine with
   a working provider.

**Credential acquisition is 401-driven, and it lives in a `DelegatingHandler`.** This
follows NuGet's own structure —
[`HttpSourceAuthenticationHandler`](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs)
and [`AmbientAuthenticationState`](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol/HttpSource/AmbientAuthenticationState.cs)
— rather than restating it. Two consequences follow from putting the loop in the
pipeline instead of at call sites: the provider's documented `-IsRetry` contract (first
attempt may return stale cached credentials; the host must detect the 401 and ask again)
is satisfiable without typed HTTP failures, and the service-index fetch in
`NuGetClient.GetPackageBaseAddressAsync` — which attached no credential at all — is
covered centrally instead of by auditing every call site for a forgotten `auth:`
argument.

The plain reason acquisition waits for a 401 is that most feeds are public and do not
need credentials; there is nothing to ask for until a feed says otherwise.

**The handler is opt-in by circumstance.** `HttpClientFactory` gained a decorator seam;
`Program.cs` installs the handler only when the run is not `--offline` *and* a plugin is
actually discoverable. A machine with no provider behaves exactly as before.

**Clear text stays.** Dropping `ClearTextPassword` would break GitHub Packages, GitLab,
Artifactory, Nexus, ProGet, Cloudsmith, and Artifact Registry outright — none of them
ship a NuGet credential provider. It is the floor, not the recommendation, and
`docs/design/nuget-authentication.md` ranks it as such.

## Credential scope

A service index names its own search endpoint. That `@id` is controlled by whoever
controls the feed, so a hostile, compromised, or merely misconfigured index can point it
at an attacker and collect the Basic credentials the user configured for the real feed.
This is the design question @rjmurillo raised in #3417.

`NuGetCredentialScope` scopes credentials to the origin the user actually configured —
scheme, host, and port of the source URL — and withholds them from anything else. The
request still goes out, so a genuinely cross-origin authenticated endpoint fails visibly
with 401 rather than silently handing over a token.

Origin comparison uses `Uri.IdnHost`, not `Uri.Host`: `Host` does not canonicalize
Unicode against punycode, so a feed configured as `https://bücher.example` whose index
advertises `https://xn--bcher-kva.example` would have had credentials withheld from the
user's own feed. `IdnHost` narrows nothing — distinct DNS names still compare distinct,
and the subdomain-suffix, scheme, and port cases are covered by theory.

## Version provenance: `--versions-with-feed`

`--versions` unions versions across every configured source and drops the provenance.
`--versions-with-feed` keeps it, emitting one row per (version, feed) pair, so a version
carried by two sources appears twice. TSV is the natural default for the shape; every
other output format is supported.

```console
$ dotnet-inspect package Markout --versions-with-feed 4 \
    --source https://api.nuget.org/v3/index.json \
    --source <private-ado-index>
Version  Feed
0.32.99  markout-azdo
0.32.0   nuget.org
0.32.0   markout-azdo
0.31.0   nuget.org
```

It is built on the same listings pipeline #3435/#3436 introduced, not beside it:
`FetchListingsPerSourceAsync` is the one per-source loop, `GetAllVersionListingsWithCacheAsync`
merges its output and `GetVersionListingsWithSourceAsync` keeps the split. Unlisted
filtering therefore applies by construction, and `--include-unlisted` adds a `Listing`
column rather than forking the query. Feed labels prefer a meaningful configured `Name`,
fall back to `nuget.org` or the URI host, and expand to the full URL on collision.

## Exit codes

The exit code follows the failure, not the result count:

| Situation | stdout/stderr | Exit |
| --- | --- | --- |
| Results found | table / JSONL | 0 |
| Genuine zero results | `No packages found for "..."` | 0 |
| Some sources unsearchable | results + `Warning: could not search <source>: <reason>` | 1 |
| Every source unsearchable | `Error: No configured NuGet source could be searched` + per-source reason | 1 |

A partial answer never reads like a complete one.

### Public Azure DevOps feed — the shape that broke parsing

`pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng`, no credentials needed. It
returns `"totalHits":"0"` with populated data:

```console
$ package search Microsoft.DotNet --source <ado-index> --take 5
3 package(s) found
  Microsoft.DotNet.Arcade.MSBuild.Xcopy       18.0.0
  Microsoft.DotNet.GitHubIssueLabeler.Assets  1.12.21
  Microsoft.DotNet.UAP.TestTools              1.0.31
exit=0

$ package search zzz-no-such-package-zzz --source <ado-index>
No packages found for "zzz-no-such-package-zzz".
exit=0

$ package search anything --source https://example.invalid/v3/index.json
Error: No configured NuGet source could be searched.
  explicit: no searchable endpoint for 'https://example.invalid/v3/index.json'
  (service index unavailable, or advertises no SearchQueryService)
exit=1
```

Raw ADO body confirming the wire shape:

```json
...,"index":"PackageIndex","totalHits":"0"}
```

Default nuget.org path unchanged (`package search Newtonsoft.Json --take 3` → 3 results,
exit 0).

### Live *private* Azure DevOps feed

A private project-scoped feed was stood up specifically to test this, publishing builds
of a real package. Subjects `Markout.Ansi.Spectre` and `MarkdownTable.Query` exist
**only** in that feed (404 on nuget.org), so a result proves private-feed access rather
than a public fallback. Cache purged between runs — a warm cache silently invalidates
this class of test.

| Setup | Result | Exit |
| --- | --- | --- |
| search, PAT via `--nugetconfig` | results | 0 |
| search, no credentials | `No configured NuGet source could be searched` | 1 |
| search, short-lived Microsoft Entra token | results | 0 |
| search, **discovered** `nuget.config` in cwd, no flags | results | 0 |
| `package`, PAT | metadata | 0 |
| `package`, credential provider on `PATH`, no credentials in config | metadata | 0 |

Row 4 is the case @rjmurillo originally reported; it works on this branch. Row 6 is the
credential-provider path end to end, with nothing secret at rest.

Multi-feed resolution was isolated by publishing a version that exists **only** in the
private feed (`Markout@0.32.99`) for a package that exists on **both**:

| Setup | Result | Exit |
| --- | --- | --- |
| `Markout@0.32.99`, nuget.org only, cold cache | `Version '0.32.99' of package 'markout' not found` | 1 |
| `Markout@0.32.99`, both feeds, **nuget.org listed first**, cold cache | resolved from the private feed | 0 |
| `Markout --versions`, both feeds | union including `0.32.99` | 0 |

Resolution does not stop at the first source that merely knows the package id.

## What review changed

The first commit was reviewed by MAI-Code and GPT independently, at a fixed head, in
isolated review worktrees, re-run at every new head. **Six rounds; thirteen blocking
defects.** The full attribution table, the two disputes and how they resolved, and the
exact validation are in the
[review reconciliation comment](https://github.com/richlander/dotnet-inspect/pull/3428#issuecomment-5113507617).

**Blast radius:** source resolution is shared by every source-aware command, not just
`package search`, and one edit touches the invocation path of every command in the tool.

The substantive changes:

- **A discovered `NuGet.Config` was ignored entirely — this was the reported bug.**
  Resolution only ran when a source *flag* was passed. Resolution now always runs, and
  the nuget.org shortcut is keyed on where resolution actually landed rather than on the
  presence of a flag.
- **Source selection was wrong in every combination but the simplest one.**
  `SourceResolver.ResolveSources` returns early on its explicit-source fast path, which
  takes a single value — so `--source A --add-source B` dropped B, and two `--source`
  values were forwarded as *additional* sources, silently searching feeds the user never
  named. Selection now happens in `NuGetSourceResolver.SelectExplicitSources`:
  `--source` replaces, `--add-source` appends.
- **Explicit sources adopt credentials from configuration, matched by endpoint** —
  without which `--source <authenticated-feed>` could never authenticate. Three matching
  bugs were found and fixed: case was folded over the whole URL (so `/FeedA` and
  `/feeda` could swap credentials — now folded only over the origin); percent-escape hex
  casing compared raw (RFC 3986 makes those digits case-insensitive, so `%2f` vs `%2F`
  silently lost credentials); and trailing-slash tolerance adopted the *first* candidate.
  **I defended the third and was wrong** — the reviewer's reproduction refuted the
  argument that origin scoping made it safe, because the adopted credential is attached
  directly and never re-gated. Slash tolerance is now a candidacy test only. On a match
  **only the credentials are adopted**; the URL stays exactly as the user spelled it.
- **The nuget.org shortcut trusted the host, not the endpoint,** so
  `--source https://api.nuget.org/definitely-not-a-service-index` was never contacted and
  search answered from the well-known endpoint at exit 0. Now keyed on the canonical V3
  service index URL.
- **An explicitly named `--nugetconfig` that cannot be used is now a failure.** Missing
  path, malformed XML, no declared sources, or unreadable all fell through to the
  nuget.org fallback and answered with unrelated packages at exit 0. That fallback is
  right for configs discovered by walking the tree and wrong for one the user named.
  Rejection happens at **parse time** via a validator on the shared option, so every
  command reports it identically — and because validators run during `Parse`, outside
  every `try`/`catch` in `Program.cs`, the validator **reports** and never throws.
- **Prefix expansion failure is a failure, not an empty answer.** The reviewers
  disagreed; resolved in favour of failure visibility while keeping the requirement that
  it not abort with a stack trace. `--package-prefix` raises `PrefixResolutionException`,
  rendered as a clean `Error:` line at exit 1. A prefix that legitimately matches nothing
  still exits 0.
- **System.CommandLine's default exception handler is disabled.**
  `parseResult.InvokeAsync()` was catching everything and printing
  `Unhandled exception: <trace>` — which is why `Program.cs`'s existing
  `catch (RowWindowValidationException)` had never fired. This is the highest-leverage
  edit in the PR.
- **Failure paths no longer skip cleanup.** `Program.cs` returned early from its catch
  blocks, skipping tail flush, `--info`, the request trace, and cache maintenance on any
  failing run. They now set `exitCode` and fall through.

## Validation

Release, at head `45b0e53bc`, which merges the updated parent branch (#3423 at
`d232ae808`, itself current with `origin/main`). Both PRs report `MERGEABLE`.

| Suite | Result |
| --- | --- |
| `NuGetFetch.Tests -trait- "Network=Live"` (offline tier) | 265 total, **0 failed**, 0 skipped |
| `NuGetFetch.Tests` (full) | 292 total, **0 failed**, 9 skipped — the live-credential tests, which skip without feed environment set |
| `DotnetInspector.Services.Tests` | 357 total, **0 failed**, 0 skipped |
| `dotnet-inspect.Tests` | 2559 total, 14 skipped, **1 failed** |

The single CLI failure is `CommandExecutionTests.Layout_Count_CountsFilesRatherThanRenderedTreeLines`,
filed as #3506 and failing identically on `main`.

**Correction to an earlier revision of this description**, which claimed 28 pre-existing
failures. That number came from running the test project directly without first building
the full solution, leaving stale assemblies in `artifacts/`. After
`dotnet build dotnet-inspect.slnx -c Release` the count is 1, here and on `main` alike. A
stale-artifact baseline inflates the noise floor, which is exactly where a real regression
would hide.

New tests in this slice, all hermetic unless noted:

| File | Covers |
| --- | --- |
| `SearchServiceTests` | Query-string shaping, auth attached/absent, the real ADO body, malformed/null/error responses throw |
| `NuGetSearchSourcesTests` | Versioned `SearchQueryService/3.5.0` discovery, feeds with no search resource, local-folder sources, all-sources-failed, partial failure alongside results, cross-source dedupe, credentials sent same-origin and **withheld** from a feed-supplied foreign endpoint, origin-comparison theory over host/scheme/port/subdomain/relative/unparseable |
| `PluginProtocolTests` | Drives a **real plugin process** — a bash script that genuinely speaks NDJSON — through handshake, request correlation, and timeout extension |
| `PluginDiscoveryTests` | All three discovery routes and their precedence |
| `PluginAuthenticationHandlerTests` | The 401 loop, retry semantics, give-up behaviour, and that a configured credential does not burn the retry budget |
| `ServiceIndexAuthenticationTests` | Service-index fetch carries credentials |
| `CredentialMechanismTests` | The `NuGet.Config` credential matrix |
| `SourcePrecedenceTests` | Per-feed listing split, configured-order precedence, and credential withholding on the discovered content host |
| `AzureDevOpsFeedTests`, `AzureDevOpsCredentialProviderTests` | Live tier, skipped without feed environment |
| `NuGetApiTests` | `MalformedJson_ReturnsNull` became `MalformedJson_Throws`; added the ADO string-`totalHits` case |

Docs: `docs/design/nuget-authentication.md` (new), `docs/design/version-resolution.md`
and `docs/workflows/core/version-queries.md` (updated). markdownlint clean.

## What the review at `1bb9ed2c` changed

An independent read-only review at a fixed head found two blocking defects. Both are fixed
in `a83043d6`, each with a test that **fails without the fix** — verified by reverting the
production change and re-running.

**The credential origin gate covered search only.** `NuGetCredentialScope.AuthFor` had
exactly one production caller. Every `PackageExtractor` fetch passed raw
`source.GetAuthHeader()`, including four whose URL is derived from the feed-controlled
service-index `@id`: the nuspec fetch, the `.nupkg` download, and the version-list and
latest-version fallbacks. A feed advertising `PackageBaseAddress` on another host was
handed the credential configured for the real feed — verbatim the attack
`NuGetCredentialScope`'s own doc comment describes. Every fetch now derives its header from
`AuthFor` against the URL actually being requested, so the rule is uniform rather than a
per-call-site judgment.

Two of those four sites hoisted the header into a local reused across both the
flat-container and the base-address URL, so the review's reading of them as "safe by
construction" did not hold; the fix covers them. Reverting the fix turns the new test red
with `Assert.Null() Failure: Value is not null` — the credential was reaching the foreign
host.

**A configured credential made the plugin fallback unreachable, expensively.**
`PluginAuthenticationHandler` cloned from the original request on every iteration and
applied an acquired credential only when the clone carried none. With a configured
credential already on the request, an acquired one could never be applied: a 401 produced
four identical failing requests and up to three `isRetry` acquisitions whose results were
discarded — the ordinary expired-PAT-plus-installed-provider case. Precedence is unchanged;
the loop is simply not entered when the caller already supplied a credential. Reverting the
fix turns the new test red with `Assert.Single() Failure: The collection contained 4 items`.

## What the second review round changed

Two further fixed-head rounds ran after `a83043d6`, from two different model families
(Gemini 3.1 Pro and MAI-Code). Both are clean at `45b0e53bc`. Three changes came out of them.

**A malformed protocol message disabled credentials for the life of the process**
(`bd064d144`). `PluginConnection.HandleLine` guarded `JsonDocument.Parse` with
`catch (JsonException)`, then read `RequestId`/`Type`/`Method` with `JsonElement.GetString()`.
Well-formed JSON carrying those names with the wrong *value types* survives the parse, and
`GetString()` on a number throws `InvalidOperationException`, which is not a `JsonException`.
It escaped `HandleLine`, was caught by the general handler in `ReadLoopAsync`, ended the read
loop, and the loop's `finally` faulted every pending request. One malformed line therefore
took the credential plugin down for the whole process — defeating the intent stated in
`HandleLine`'s own comment that non-protocol noise must not kill the conversation. The three
header fields now go through `TryReadString`, which requires `JsonValueKind.String`.

Reverting the production file turns the new test red with `Assert.NotNull() Failure: Value is
null` after **30.1s**, against 0.18s with the fix — the delay being the request timeout, which
is itself the evidence that the connection was already dead. The reviewer then audited the rest
of `src/NuGetFetch/Plugins/` for other type-assuming `JsonElement` accessors; none remain.

**Six Windows-only test failures** (`6731f72a5`), all fixture-level, no product change. The
`PluginDiscoveryTests` helpers wrote extensionless files as "executables" (not executable on
Windows, where `IsPathPluginName` requires `.exe`/`.bat`) and left `/` separators inside paths
later compared ordinally against product-built `\` paths.

**The shared-authority test was strengthened twice** (`7a2f7da2b`, `45b0e53bc`) after review
showed it proved less than claimed. It is the evidence for treating the authority-keyed cache
collision (#3540) as non-blocking, so its discriminating power is load-bearing: it now pins
that the cache exists, and that a credential learned for one authority is never offered to
another.

## Boundary and non-actions

Declared residuals — read these as scope, not as defects:

- **Bug 1 of #3417 is not fixed here.** A 401 still surfaces as `Package '<id>' not
  found` on the package path. Typed HTTP failures carrying source, phase, and status are
  their own change. The credential-provider work was originally sequenced *behind* bug 1
  on the belief that 401-driven acquisition needed status propagated to call sites; the
  `DelegatingHandler` structure showed that premise was wrong, which is why this landed
  first. Bug 1 remains worth fixing, for diagnostics.
- **The flat-container/download path keeps its current credential behaviour** on the
  `NuGet.Config` route. The plugin handler covers it where a provider exists; threading
  gated credentials through its four call sites otherwise is a behaviour change on a path
  this PR cannot exercise.
- **Per-source attribution in *search* output** is not included; `--versions-with-feed`
  covers the version path only.
- **Multi-source `--take` budgeting**: each source is asked for `take` and the merged
  list is truncated to `take`, so the first feed can consume the whole budget. Raised by
  MAI-Code; deferred as a ranking/interleaving design question coupled to per-source
  attribution.
- **`<Password>` (DPAPI-encrypted) is still not parsed** — `EncryptionUtility` has zero
  callers on `main` as well as here. `dotnet nuget add source --password` writes that
  element by default on Windows, and such a config is silently ignored.
- Duplicate same-URL `nuget.config` entries with differing credentials resolve by
  dictionary order, so which identity authenticates is nondeterministic (#3447). Not a
  leak — the request target is byte-identical to the configured entry.
- `--help` does not short-circuit option validators, so an invalid option value reports
  the error instead of showing help. Pre-existing and repo-wide (the untouched `--rows`
  validator behaves identically); raised independently by both reviewers and deferred.
- Test isolation: `ConsoleCapture` redirects the process-global `Console` (#3446), and
  two `LibraryCommand` tests fail deterministically on Windows (#3445). Both pre-existing,
  both surfaced while validating this PR.
- A source URL containing userinfo is echoed verbatim into logs and error messages
  (#3535). Pre-existing on `main`; the cited lines are untouched by this PR.

Non-blocking findings from the review at `1bb9ed2c`, deferred deliberately:

- **The plugin credential cache is keyed per authority on a process-wide client.**
  NuGet.Client builds a handler per package source and so never mixes credentials across
  sources; every Azure DevOps feed shares `pkgs.dev.azure.com`, so all orgs share one
  credential slot. Benign with org-scoped PATs, but a real divergence from NuGet's
  isolation model — worth either matching or documenting deliberately, and not while
  fixing two blocking defects.
- **Plugin discovery is eager**, running a PATH scan in the provider constructor on every
  non-offline run, including commands that never touch the network. Making it lazy is
  free and is the obvious next change.
- **No Windows test job.** This PR spawns external processes, redirects stdio, and calls
  `Kill(entireProcessTree: true)` — precisely where Windows diverges — and CI runs only
  `ubuntu-26.04`.
- **`--versions-with-feed` rows dedupe by feed label rather than source URL**, and
  `BuildFeedLabels`' URL fallback can still collide. Raised independently by two
  reviewers; the criticism holds and the fix belongs with per-source attribution in
  search output, which is already listed above.


richlander:

High-risk change (credential handling plus a CLI-wide exception-handling edit), so it took the two-model tier: **MAI-Code** and **GPT**, each in an isolated review worktree pinned to an exact head, re-run at every new head. Six rounds, thirteen blocking defects. Final reviewed head: `cfed8a01`. Both reviewers report no blocking issues there.

### Findings and disposition

| # | Found by | Finding | Disposition |
|---|---|---|---|
| 1 | both | `--source A --add-source B` dropped B; multiple `--source` re-entered config resolution and searched feeds the user never named | Fixed — `SelectExplicitSources` resolves selection directly (`d3c5195e`) |
| 2 | GPT | Missing `--nugetconfig` file silently fell back to other feeds and exited 0 | Fixed — explicit configs are validated, not defaulted (`d3c5195e`) |
| 3 | GPT | IDN/punycode host comparison | Fixed — `IdnHost` (`d3c5195e`) |
| 4 | MAI | `--nugetconfig <missing>` produced an unhandled stack trace on every source-aware command | Fixed — parse-time validator on the shared option (`54c4d175`) |
| 5 | GPT | Credential matching folded case, so `/FeedA` and `/feeda` could swap credentials | Fixed — ordinal path/query compare (`54c4d175`) |
| 6 | GPT | Well-formed but unusable config searched unrelated feeds and exited 0 | Fixed — reports and exits 1 (`54c4d175`) |
| 7 | GPT | Prefix-expansion failure exited 0 | Fixed — `PrefixResolutionException`, exit 1 (`54c4d175`) |
| 8 | GPT | A config that exists but cannot be read crashed the process | Fixed — validators report, never throw (`f9be5094`) |
| 9 | GPT | Percent-escape hex casing compared raw, silently withholding credentials | Fixed — `NormalizeEscapes` (`f9be5094`) |
| 10 | GPT | Failure paths skipped tail flush, `--info`, request trace, cache maintenance | Fixed — catch blocks set `exitCode` and fall through (`f9be5094`) |
| 11 | GPT | **A discovered `NuGet.Config` was ignored entirely** — search went to nuget.org anyway | Fixed — resolution always runs (`c97506eb`) |
| 12 | GPT | Trailing-slash tolerance could attach one entry's credential to another entry's spelling | Fixed — exact match wins; slash tolerance only when unambiguous (`c97506eb`) |
| 13 | GPT | nuget.org shortcut keyed on *host*, so a named non-index path on that host returned nuget.org's results at exit 0 | Fixed — keyed on the canonical service index (`cfed8a01`) |

**Finding 11 deserves calling out.** It was the reported symptom of #3417 itself. The first implementation only resolved sources when a source *flag* was passed, so a feed configured the ordinary way — a `NuGet.Config` discovered by walking up from the working directory, which is exactly the reported case — still went to nuget.org. Caught in round 4, not round 1.

### Disputed, and resolved

**MAI, round 5 (blocking).** Duplicate identical-URL entries let `Match` adopt the first entry's credential; MAI wanted an uncredentialed source returned instead.

> Dismissed after argument. Both entries name the byte-identical URL, so the adopted credential is one configured *for exactly the endpoint being requested* — no cross-endpoint leak, unlike finding 12 where the entries named different URLs. The suggested fix would break authentication for a feed the user did credential, i.e. reintroduce #3417. MAI withdrew explicitly after failing to construct a cross-endpoint repro, and GPT independently agreed. The residual nondeterminism — two same-URL entries with *different* credentials resolve by `Dictionary` order — is real, but it is a wrong-identity-for-the-right-endpoint wart rather than a leak. Filed as #3447.

**GPT, round 2 (blocking).** Trailing-slash tolerance in `IsSameEndpoint`.

> I defended this and I was wrong. My argument was that `AuthFor` is origin-scoped, so tolerance granted no exposure the origin gate did not already grant. GPT refuted it with a working reproduction: the adopted credential is attached to the request directly and is never re-gated by `AuthFor`. Fixed as finding 12. The doc comment that asserted the false property now states the requirement it actually places on callers, and names where that requirement is met.

**Both reviewers, non-blocking.** `--help` combined with an invalid option value shows the error instead of help.

> Deferred. Pre-existing and not introduced here — the untouched `--rows` validator behaves identically (`package --help --rows 9:2` errors, exit 1), which GPT verified at this head. Changing it is a repo-wide option-validation semantics change.

**MAI, non-blocking.** `--take` is not budgeted across multiple sources, so the first feed can consume the whole budget. Real; listed as follow-up.

### Test-suite work

An intermittent `SymbolPackageDownloaderTests` failure was traced to a `[Collection("Console")]` attribute **I had added speculatively** in round 3 to fix a different console bleed. Measured: base `f000b809` clean 6/6, head with the attribute failing ~40% of runs, head with it reverted clean 0/7. Reverted.

The underlying cause — `ConsoleCapture` redirecting the process-global `Console` while other collections run in parallel — cannot be fixed by growing that collection, because a collection cannot contain process-global state. Filed as #3446. The two Windows-only `LibraryCommand_*` failures are pre-existing and filed as #3445.

### Validation at `cfed8a01`

`dotnet build dotnet-inspect.slnx -c Release` clean. `dotnet-inspect.Tests` 2376 total, 2 failed (both #3445). `DotnetInspector.Services.Tests` 319/319. `NuGetFetch.Tests` 227/227. Both reviewers independently reproduced this.

Behavioral contract verified through the built CLI:

| Invocation | Result | Exit | Previously |
|---|---|---|---|
| discovered `NuGet.Config` naming an unreachable feed | names that feed and fails | 1 | nuget.org results, exit 0 |
| `--source https://api.nuget.org/definitely-not-a-service-index` | fails | 1 | nuget.org results, exit 0 |
| `--source https://api.nuget.org/v3/index.json` | results | 0 | unchanged |
| bare `package search Newtonsoft.Json` | results | 0 | unchanged |
| `package search <nonexistent>` | `No packages found` | 0 | unchanged |

That last row is the contract chosen for this PR: **exit 1 only for failures; a genuine zero-result search stays exit 0.**

### Correction: CI does run on this stacked PR

An earlier revision of this comment claimed CI could not run here because `.github/workflows/ci.yml` filters on `branches: [main]`. **That was wrong**, and I am correcting it rather than quietly deleting it. CI ran on this PR at the reviewed head and passed:

- run [30423773634](https://github.com/richlander/dotnet-inspect/actions/runs/30423773634), event `pull_request`, head `cfed8a01`, conclusion **success**
- `changes`, `build-net10`, `test (ubuntu-26.04, linux-x64)` all green; `pack`, `csharp-diff-smoke`, `il-diff-smoke`, `markdownlint` path-skipped

One caveat that is *not* a correction: this run validated merging into `feature/rehome-nugetfetch`, not into `main`. Merge order is still forced — **#3423 merges first** (`MERGEABLE`/`CLEAN`, 7/7 green, already marked ready) — after which GitHub retargets this PR to `main` and CI re-runs against the real merge path.

### Remaining before merge

Required review is complete and CI is green at the reviewed head. Two things still gate merge:

1. #3423 merges first, and CI re-runs green against `main`.
2. A demo posted to #3417 showing the fixed behavior against a real authenticated feed.

I will post `Ready to merge` once both are done.

richlander:
## Setup: markout, built downstream in Azure DevOps

To test this against a real private feed rather than a mock, I stood up an Azure DevOps org with
a pipeline that clones [markout](https://github.com/richlander/markout) from GitHub, builds and
packs it, and publishes to a project-scoped Azure Artifacts feed. **Upstreams are disabled on
that feed**, so a successful read proves private-feed access rather than a silent fallback to
nuget.org.

The pipeline takes a version parameter, so I built **`Markout 0.32.99`** — a version that exists
only on that feed. Not on nuget.org, not in any local cache. That is the probe used throughout
below. Findings follow; tests and docs pushed as b7f8b7c and 07088c4.

### Methodology note: the cache short-circuits the feed, by design

A version already in the cache is returned without probing any feed. That is the tool working as
intended — cache-first, and I am not reporting it as a bug. It matters here only as methodology:
every result below was taken cold, because my first two attempts were contaminated by cached
content and told me nothing.

The one case where that design does bite is multiple feeds serving **different content for the
same version**, where the cache decides which one you get. That is not a scenario this tool sets
out to support, and I would not change the behavior for it.

### Confirmed working

- **The reported bug is fixed.** Searching an authenticated non-nuget.org feed works.
- **Multi-feed resolution asks every feed.** With nuget.org listed *first* — it carries `Markout`
  but not `0.32.99` — resolution correctly continued on to the private feed. `--versions` unions
  across feeds.
- **PAT and Microsoft Entra access token behave identically**, which is expected: both are just
  the Basic password.

### 1. We support the one credential mechanism NuGet ranks last

NuGet publishes a [most-to-least-secure ordering](https://learn.microsoft.com/nuget/consume-packages/consuming-packages-authenticated-feeds#security-best-practices-for-managing-credentials)
of credential mechanisms. Lining it up against what we honor:

| NuGet's rank | Mechanism | Supported |
| --- | --- | --- |
| 1, "highly recommended" | Credential provider | **No** — no plugin invoked, no `ARTIFACTS_CREDENTIALPROVIDER_*` / `VSS_NUGET_*` read |
| 2 | Encrypted `<Password>` | **No** — never parsed; `EncryptionUtility` has zero callers here and on `main` |
| 3 | `%VAR%` macros in `nuget.config` | **No** — taken verbatim, so the literal `%VAR%` is sent as the password |
| 4 | `NuGetPackageSourceCredentials_<name>` | **No** — environment never consulted |
| 5, "only ... where no other secure option is available" | `Username` + `ClearTextPassword` | **Yes — the only one** |

So we support exactly the mechanism the docs rank last and attach a leak warning to, and ignore
all four preferred ones. Rank 1 is the expensive miss: a credential provider is what a correctly
configured CI pipeline uses, so such a pipeline still arrives unauthenticated.

Two more that aren't on NuGet's list but look plausible and also fail: `ClearTextPassword` with
no `Username` (both halves required, though AzDO ignores the username), and userinfo in the URL
(`https://user:pass@host/...`).

**Every one of these surfaces as `Package 'markout' not found`** — identical to a typo in the
package name. That is bug 1 doing real damage.

### 2. Adding mechanisms needs bug 1 fixed first

The official client is **401-driven**: it requests anonymously, and *on a challenge* searches for
credentials (env var, then `nuget.config`, then credential provider). We attach whatever we
parsed up front and treat a 401 as terminal — we never widen the search after a challenge.

That's why rank 1 isn't just a parser change. The credential provider contract expects a caller
that can detect a 401 and re-invoke with `-IsRetry`, precisely to avoid reusing invalid cached
credentials. Bug 1 is a prerequisite, not an independent fix.

If you do want the provider later, integration is cheap: standalone mode returns exactly
`{"Username","Password"}`, one-for-one with `PackageSourceCredential`. I verified end to end that
a provider-issued credential resolves `Markout 0.32.99` from the private feed. Caveat: the v2.0.2
dotnet-tool build ships no `msalruntime.so` for linux-x64, so its MSAL paths throw
`DllNotFoundException`; only the env-var paths worked here.

### 3. `NuGetClient` fetches the service index anonymously

`GetPackageBaseAddressAsync` takes no credential and issues a bare `GetStreamAsync`. Azure DevOps
authenticates its service index, so a caller that passes a perfectly good credential to
`GetVersionsAsync` still gets a 401 at discovery, before the credential is ever offered.

The CLI does **not** hit this — `PackageExtractor` has its own service-index reader that passes
`source.GetAuthHeader()`. So this is confined to the `NuGetFetch` library and affects any other
consumer of it. Pinned by a test rather than fixed here, since it is arguably a separate change.

Worth noting for bug 1: the 401 *is* available at this layer. `GetVersionsAsync` only swallows
404, so the status survives on the exception — fixing the diagnosis does not need new plumbing.

### 4. `NuGetFetch.Tests` was not run by any workflow

The project this PR adds had no gate at all. Now wired into CI, excluding the live tier.

### What landed

- `CredentialMechanismTests` — 9 hermetic tests pinning every row of the ranking table, so
  support for any of them gets added deliberately instead of by accident.
- `ServiceIndexAuthenticationTests` — 5 tests over an in-memory transport covering which request
  carries the credential and that 401 stays distinguishable from 404.
- `AzureDevOpsFeedTests` — 4 live tests, `[Trait("Network", "Live")]`, skipped unless a feed and
  token are supplied, since CI and fork PRs have neither. Token comes from the environment and is
  never written to a config file. Only a real AzDO feed exercises an authenticated service index.
- CI runs `-trait- "Network=Live"`. 224 tests green offline; all 4 live tests green against the
  real feed.
- `docs/design/nuget-authentication.md`, which links to the NuGet and Azure DevOps guidance
  rather than restating it.

### Not addressed

The serving feed is not machine-readable: `--verbose` names the feed, but the table and `--json`
both emit a fixed literal `"source": "NuGet"`. With multiple feeds configured, an agent cannot
tell where a package came from. Happy to file separately if useful.

Also: there is no client-certificate mode at the feed, and there cannot be. A certificate
authenticates a service principal to Entra ID, which returns a token; the feed only ever sees
Basic. So the PAT and Entra cases above already cover the security models.

richlander:
## Read-only review of `1bb9ed2c`

Reviewed in an isolated worktree at the exact head; no code changed. Findings are ordered by severity. Nothing here was reproduced against a live feed — items 1 and 2 are read from the code paths and are stated as such.

### 1. Blocking — the credential origin gate covers search only, not download

`NuGetCredentialScope.AuthFor` is called from exactly one place: `NuGetSourceCompat.cs:290`, the search path. Every `PackageExtractor` path passes raw `source.GetAuthHeader()` instead:

| Call site | URL derived from | Gated? |
| --- | --- | --- |
| `PackageExtractor.cs:579` (service index) | `source.Url` | safe by construction |
| `PackageExtractor.cs:790`, `:1020` (flat container) | `GetFlatContainerUrl()`, which returns non-null only for nuget.org and so is never credentialed | safe by construction |
| `PackageExtractor.cs:500` (nuspec) | `GetPackageBaseAddressAsync` → feed-controlled `@id` | **no** |
| `PackageExtractor.cs:315` (`.nupkg` download) | `GetPackageDownloadUrlAsync` → feed-controlled `@id` | **no** |

`HttpRetryHelper` attaches whatever header it is handed (`:227`, `:318`) and applies no gate of its own.

So a hostile, compromised, or merely misconfigured feed that advertises `PackageBaseAddress` on another host receives the configured credential on the nuspec and `.nupkg` requests. That is verbatim the attack `NuGetCredentialScope`'s own doc comment describes:

> a hostile, compromised, or merely misconfigured feed can name any URL it likes and collect the credentials the user configured for the real feed

The gate is correct; it is simply not on the download path. Suggest routing every credentialed `PackageExtractor` fetch through `AuthFor` rather than `GetAuthHeader`, and pinning that with a test per discovered-endpoint path, not just search.

### 2. Blocking — a configured credential makes the plugin fallback unreachable, and expensively so

`PluginAuthenticationHandler.SendAsync` clones from the **original** request on every iteration and applies the acquired credential only when the clone has none:

```csharp
using (HttpRequestMessage attempt = CloneRequest(request))
{
    if (credential is not null && attempt.Headers.Authorization is null)
    {
        attempt.Headers.Authorization = CreateBasicHeader(credential);
    }
    ...
}
```

When a static credential is already on the request — set upstream at `SearchService.cs:33` and `HttpRetryHelper.cs:227`/`:318` — and the server answers 401:

- the plugin credential is acquired and cached, but never applied, because every clone re-inherits the static header;
- the loop still runs to `MaxAuthRetries = 4`, producing four identical, identically-failing requests;
- `AcquireAsync` is called up to three more times with `isRetry: true`, which this PR's own doc comment says makes the Azure Artifacts provider acquire fresh credentials rather than serve its cache — up to three expensive, potentially interactive acquisitions whose results are discarded.

The precedence policy itself is defensible, and `ConfiguredCredentialsTakePrecedenceOverAcquiredOnes` gates it — but that test returns 200 on the first attempt, so the retry loop never engages and the cost of the policy under 401 is ungated.

This is the realistic Azure DevOps case: an expired PAT in `nuget.config` plus an installed credential provider. The user gets four failing requests and no fallback.

Suggested fix, preserving identical precedence semantics: if `request.Headers.Authorization` is non-null on entry, send once and return without entering the acquisition loop. A test with a pre-set header against a transport that always answers 401 would gate it.

### 3. Moderate — the credential cache is keyed per authority on a process-wide client

`GetAuthority` is `uri.GetLeftPart(UriPartial.Authority)` (scheme, host, port), and `HttpClientFactory.Shared` is a single process-wide client carrying one decorator. NuGet.Client builds an `HttpSource`, and therefore a handler, **per package source**, so it never mixes credentials across sources.

Every Azure DevOps feed lives under `https://pkgs.dev.azure.com`, so all organizations and feeds share one credential slot, and the plugin is asked once per host rather than once per source. Mostly benign with org-scoped PATs, but it is a real divergence from NuGet's isolation model and worth either matching or documenting deliberately.

### 4. Minor — plugin discovery is eager on every invocation

`PluginCredentialProvider`'s constructor runs `PluginDiscovery.Discover()` — a PATH scan plus directory enumeration — and `Program.cs:65` constructs the provider on every non-offline run, including `il`, `type`, and `decompile`, which never touch the network. The comment at `Program.cs:62-64` acknowledges the scan but does not defer it. Making `_executables` lazy would cost nothing and keep startup off the scan.

### 5. Process — no Windows test job

CI is green, but there is a single test job, `test (ubuntu-26.04, linux-x64)`. This PR spawns external processes, redirects stdio, and calls `Kill(entireProcessTree: true)`. Windows is precisely where that behavior diverges, and nothing covers it.

Separately, per `AGENTS.md`, a change of this risk profile (credential handling plus subprocess launch) sits in the two-model adversarial review tier.

### Credit where due

Several things here are done well and should survive any rework:

- `UseShellExecute = false`, bounded request, close, and exit timeouts, and `Kill(entireProcessTree: true)` on teardown.
- `MaxAuthRetries = 4` matches NuGet's `AmbientAuthenticationState`.
- `PromptOn403` and `AllowInteractive` both default to false, so the tool cannot hang unattended.
- Failed plugins are remembered as `null`, so a bad plugin is not relaunched.
- `PluginCredentialProvider` refuses non-`basic` `AuthenticationTypes` rather than sending a credential in an unsanctioned form.
- `SetAuthenticationDecorator` is deliberately not applied to `SharedUntrustedFetch`, so feed credentials are never offered to URLs drawn from untrusted artifacts. This is the easiest thing in the change to have gotten wrong.
- `NuGetCredentialScope.NormalizeEscapes` has correct bounds, and comparing `IdnHost` rather than `Host` is the right call with a clear justification.
- `PluginDiscovery` mirrors NuGet.Client's three-route precedence faithfully, including the easily-missed PATH scan.

richlander:
## Re-review of `2ad4a14a` (fix commit `a83043d6`)

Both blocking findings are **fixed correctly**, and the fixes are gated by tests that fail without them. Verified in an isolated worktree at the exact head. One residual, two unchanged follow-ups, and one new finding backed by a local Windows run.

### Finding 1 (credential origin gate) — resolved, and better than asked

Every credentialed fetch now derives its header from `AuthFor` against the URL actually being requested. The structural check confirms it:

```
$ git grep -n "GetAuthHeader()" -- "src/*.cs" ":!*Tests*"
src/DotnetInspector.Packages/NuGetCredentialScope.cs:135
src/NuGetFetch/PackageRecords.cs:17
```

`GetAuthHeader()` now has exactly one production caller — the gate itself. That makes the rule hold by construction rather than by per-call-site judgment, so a new fetch cannot quietly bypass it. That is a stronger fix than the one proposed.

The commit message is also right to correct the review: my "safe by construction" reading of `:790` and `:1020` did not hold, because both hoisted the header into a local reused across the flat-container *and* the base-address URL. Only the flat-container leg was nuget.org-only; the fallback leg was `@id`-derived and credentialed. Good catch.

Gating is sound. `VersionListing_WithholdsCredentialsFromForeignContentHost` and `VersionListing_SendsCredentialsToSameOriginContentHost` are a proper positive/negative pair — the second is what stops the gate from being "withhold everything," which would pass the first while breaking every private feed.

### Finding 2 (plugin retry burn) — resolved

The early return when `request.Headers.Authorization is not null` preserves precedence exactly and skips the loop. `ConfiguredCredentialThatIsRejectedDoesNotBurnTheRetryBudget` asserts `Single(transport.Requests)` and `source.Calls == 0` under a blanket 401; without the fix those would be 4 and 3. Non-vacuous.

### New — `NuGetFetch.Tests` fails 6 tests on Windows

This is finding 5 ("no Windows job") turning from speculation into evidence. Local run at this head, after `dotnet build dotnet-inspect.slnx -c Release`:

| Suite | Result |
| --- | --- |
| `DotnetInspector.Services.Tests` | Total 357, **Failed 0** — includes both new credential-scope tests |
| `NuGetFetch.Tests` | Total 290, Skipped 22, **Failed 6** |

All six are in `PluginDiscoveryTests`:

- `Path_FindsNuGetPluginPrefixedExecutables`
- `Path_IgnoresFilesWithoutThePrefix`
- `Path_ToleratesEntriesThatDoNotExist`
- `PluginPathsVariable_AcceptsSeveralEntriesAndDirectories`
- `ConventionDirectoryAndPath_AreBothSearched`
- `ConventionDirectory_RequiresEntryPointNamedAfterItsFolder`

**The product is correct here; the fixture is not.** `PluginDiscovery.IsExecutable` properly requires `.exe`/`.bat`/`.cmd` on Windows and the execute bit elsewhere. Two Unix-centric assumptions in the test helpers cause the failures:

1. `CreateExecutable` (`PluginDiscoveryTests.cs:203`) writes an **extensionless** file. On Windows that is correctly not an executable, so PATH discovery finds nothing and `Assert.Single` sees an empty collection.
2. `CreateFile` (`:196`) returns `Path.Combine(_root, "plugins/netcore/...")`, keeping forward slashes in the relative portion. The product builds its paths from `Directory.GetDirectories` plus `Path.Combine`, which are all backslashes, so the ordinal `Assert.Equal` on path strings fails even though discovery worked.

Suggested fix, both in the fixture: normalize with `relativePath.Replace('/', Path.DirectorySeparatorChar)` in `CreateFile`, and append `.exe` on Windows in `CreateExecutable`.

There is also one **vacuous pass** worth fixing at the same time. `Path_MatchesThePrefixCaseSensitively` creates `NuGet-Plugin-Uppercase` and asserts `Assert.Empty`. On Windows that file is rejected for lacking `.exe`, never for its casing, so the case-sensitivity property it names is not gated on Windows at all. Giving the helper the `.exe` extension makes it assert the thing it claims.

This matters beyond a red suite: Windows plus Azure DevOps plus the Azure Artifacts Credential Provider is the single most likely deployment of this feature, and it is the configuration with no CI coverage.

### Residual on finding 1 — `NuGetClient` has the same ungated shape

`NuGetClient.ApplyCredential` (`src/NuGetFetch/NuGetClient.cs:216`) attaches the credential unconditionally, and its two callers — `:24-30` (`GetVersionsAsync`) and `:95-101` (`DownloadAsync`) — target a base address resolved from the feed-controlled service-index `@id` via `ResolveBaseAddressAsync`. That is the same defect class just fixed in `PackageExtractor`.

**Not blocking:** `NuGetClient` has no production callers outside `NuGetFetch` today, so it is latent rather than live.

Worth noting as follow-up because the layering makes it easy to re-introduce. `NuGetCredentialScope` lives in `DotnetInspector.Packages`, which sits *above* `NuGetFetch`, so `NuGetClient` cannot call the gate even if it wanted to. Any future caller wiring a credential into `NuGetClient` gets the ungated behavior by default. Pushing origin scoping down into `NuGetFetch` — where `PackageSource` already lives — would give both layers one rule.

### Unchanged, non-blocking

- **Finding 3:** `GetAuthority` still keys on scheme/host/port (`PluginAuthenticationHandler.cs:183`) on a process-wide shared client, so all Azure DevOps orgs share one credential slot. Divergence from NuGet's per-source isolation.
- **Finding 4:** `PluginDiscovery.Discover()` still runs eagerly in the constructor (`PluginCredentialProvider.cs:38`) on every non-offline invocation.

Both are reasonable as follow-up issues rather than gates on this PR.

### Assessment

The two blocking defects are properly closed. The Windows test failures should be fixed before merge — they are a fixture change with no product risk, and leaving them red means the first Windows contributor to touch this hits a broken suite that CI says is green.

richlander:
## Follow-ups filed

The non-blocking items from the two reviews above now have homes, so they are off this PR's critical path:

| Issue | Item | Severity |
| --- | --- | --- |
| #3539 | `NuGetClient.ApplyCredential` sends credentials to service-index-derived endpoints ungated (`NuGetClient.cs:216`, callers at `:24-30` and `:95-101`) | latent — no production callers outside `NuGetFetch` |
| #3540 | Plugin credential cache keyed per authority, so all Azure DevOps orgs share one credential slot (`PluginAuthenticationHandler.cs:183`) | behavioral divergence from NuGet |
| #3541 | `PluginDiscovery.Discover()` runs eagerly on every non-offline invocation (`PluginCredentialProvider.cs:38`, `Program.cs:65`) | startup cost |

#3539 is the one I would not leave indefinitely. It is unreachable today, but `NuGetCredentialScope` lives in `DotnetInspector.Packages`, *above* `NuGetFetch`, so `NuGetClient` structurally cannot call the gate — which means any future caller wiring a credential into it gets the ungated path by default rather than by mistake.

**The only item I would still treat as blocking on this PR is the six Windows test failures in `PluginDiscoveryTests`**, detailed in the re-review above. They are a fixture change with no product risk, but leaving them red means a Windows contributor hits a broken suite that CI reports as green, plus one assertion (`Path_MatchesThePrefixCaseSensitively`) that passes for the wrong reason.

## Offer: I can run the Windows tests

I have a Windows machine and CI does not — the matrix is `test (ubuntu-26.04, linux-x64)` only. If whoever picks up the fixture fix cannot verify locally, push the change and say so here, and I will run it and report back.

What I can provide on request:

- `dotnet build dotnet-inspect.slnx -c Release` followed by any suite from the `AGENTS.md` table, on Windows.
- Full failure output — assertion messages, stack traces, and actual-vs-expected path strings, which is what makes the separator mismatch legible.
- Confirmation that the two vacuous-pass cases become genuine once `CreateExecutable` produces a real Windows executable name.
- A baseline comparison against the pre-fix head, so a Windows-only failure can be told apart from one this PR introduced.

For reference, the current numbers at `2ad4a14a` on Windows:

| Suite | Total | Failed | Skipped |
| --- | --- | --- | --- |
| `DotnetInspector.Services.Tests` | 357 | 0 | 0 |
| `NuGetFetch.Tests` | 290 | 6 | 22 |

All six failures are in `PluginDiscoveryTests`, and the product code is correct in every one of them — `PluginDiscovery.IsExecutable` handles Windows properly. The two fixture causes are extensionless files from `CreateExecutable` (`PluginDiscoveryTests.cs:203`) and forward slashes retained by `CreateFile` (`:196`) being ordinal-compared against the product's backslashes.

richlander:
## Reconciliation: re-review of `2ad4a14a`, and the Windows finding

Responding to the re-review of `2ad4a14a` (which confirmed the two blocking fixes in `a83043d6`). Head is now `6731f72a5`.

### Windows failures — fixed in `6731f72a5`

Confirmed the diagnosis by reading the product against the fixtures. `PluginDiscovery.IsPathPluginName` is correct on both platforms; the two fixture helpers were not:

- `CreateExecutable` wrote an **extensionless** file and set the owner execute bit. That is an executable on Unix and is correctly *not* one on Windows, so PATH discovery found nothing. It now appends `.exe` on Windows.
- `CreateFile` passed relative paths containing `/` straight to `Path.Combine`, which inserts a separator *between* segments but never rewrites separators *inside* one. The product builds paths from `Directory.EnumerateFiles` and `Path.Combine` — backslashes throughout on Windows — so the ordinal comparisons failed on separator alone. It now normalizes to `Path.DirectorySeparatorChar`.

Both are no-ops on Unix, which the suite confirms: `NuGetFetch.Tests` is 263 total, **0 failed** on Linux at `6731f72a5`, unchanged from before the fix.

I should be explicit about the limit of my evidence: **I have no Windows machine and did not run the suite there.** The Windows-green claim rests on your local run plus the code reading above, not on a run of mine. That gap is itself the finding — see #3542 below.

### One correction: `Path_MatchesThePrefixCaseSensitively` is not a vacuous pass

The re-review argued that on Windows this test's file "is rejected for lacking `.exe`, never for its casing." That does not hold. `IsPathPluginName` tests the prefix **before** it tests the extension:

```csharp
if (!name.StartsWith(PathPrefix, StringComparison.Ordinal))
{
    return false;          // <-- NuGet-Plugin-Uppercase exits here, on every platform
}

if (OperatingSystem.IsWindows())
{
    string extension = Path.GetExtension(name);   // never reached for that name
    ...
}
```

and `SafeEnumerateFiles` applies no glob (`Directory.EnumerateFiles(directory)`), so nothing filters by name earlier either. The casing was always the operative reason for rejection, on Windows as much as on Unix. The test was gating what it claimed.

I made the change anyway — the helper now gives it a real `.exe` — because it costs nothing and removes the need to make this argument again. But it is a robustness tidy-up, not a repair.

Also minor: the extension check accepts `.exe` and `.bat` only, not `.cmd` as the re-review stated.

### Findings converted to issues

| Finding | Issue |
| --- | --- |
| `NuGetClient.ApplyCredential` ungated on service-index-derived base address (residual on finding 1) | #3539 |
| Credential cache keyed per authority, so Azure DevOps orgs share one slot (finding 3) | #3540 |
| Plugin discovery scans PATH on every invocation (finding 4) | #3541 |
| CI has no Windows leg (finding 5) | #3542 |
| `CloneRequest` shares the original `HttpContent` | #3543 |
| Plugin request catch filter narrower than a dying plugin's failures | #3544 |
| Feed label disambiguation can still collide | #3545 |

I verified each in code before filing rather than forwarding the claim. Two notes:

- **#3539 is the sharpest of these.** The layering is what makes it dangerous: `NuGetCredentialScope` lives in `DotnetInspector.Packages`, which sits *above* `NuGetFetch`, so `NuGetClient` **cannot** call the gate even if it wanted to. Any future caller wiring a credential into `NuGetClient` gets ungated behavior by default. Confirmed `NuGetClient` has no production callers outside `NuGetFetch` today, so it is latent, not live — hence follow-up rather than blocking.
- **#3540's practical bite is worse than "divergence from NuGet".** `GetAuthority` is scheme+host+port, and every Azure DevOps organization shares the host `pkgs.dev.azure.com`, differing only by path. So the collision is not exotic; it is the default for anyone with two AzDO orgs.

### Validation at `6731f72a5`

| Suite | Result |
| --- | --- |
| `NuGetFetch.Tests` (offline) | 263 total, 0 failed |
| `DotnetInspector.Services.Tests` | 357 total, 0 failed |
| `dotnet-inspect.Tests` | 1 failed — pre-existing #3506 |

CI green at this head: `changes`, `build-net10`, `test`, `pack`, `il-diff-smoke`, `csharp-diff-smoke`, `markdownlint` all pass. `MERGEABLE`.

### Round status

This is a high-risk change, so it needs two fixed-head rounds from two different model families. At `6731f72a5` one is complete (MAI-Code, clean at `2ad4a14a` plus this fixture-only delta). A second round from a different family is running now against this exact head. Not marking ready until it returns clean.

richlander:
## Adversarial review reconciliation — second round complete at `45b0e53bc`

This is a high-risk change (credential handling), so it requires two fixed-head rounds from two different model families. Both are now clean at the shipping head.

| Round | Family | Verdict at `45b0e53bc` |
| --- | --- | --- |
| 1 | Gemini 3.1 Pro | clean |
| 2 | MAI-Code | clean |

Each ran in its own isolated linked worktree at a fixed head, with mutations prohibited.

### Round 1 opened with two blocking findings. One was real; one was not.

**Accepted and fixed — malformed protocol message killed the plugin for the life of the process** (`bd064d144`).

`HandleLine` guarded `JsonDocument.Parse` with `catch (JsonException)`, then read `RequestId`/`Type`/`Method` with `JsonElement.GetString()`. Well-formed JSON carrying those names with the *wrong value types* survives the parse, so the guard never sees it, and `GetString()` on a number throws `InvalidOperationException` — not a `JsonException`. It escaped `HandleLine`, was caught by the general handler in `ReadLoopAsync`, ended the read loop, and the loop's `finally` faulted every pending request. One malformed line disabled credentials for the whole process, defeating the intent stated in `HandleLine`'s own comment that non-protocol noise must not kill the conversation.

The three header fields now go through `TryReadString`, which requires `JsonValueKind.String`.

The gating test drives a fake plugin that emits `{"RequestId":123,...}` before answering the handshake normally. Stashing only the production file:

- with the fix: passes in **0.18s**
- without it: `Assert.NotNull() Failure: Value is null` after **30.1s**

The 30 seconds is the corroboration — it hung to the request timeout because the connection was already dead.

At my request the reviewer then audited the rest of `src/NuGetFetch/Plugins/` for other type-assuming `JsonElement` accessors, to close the class rather than the instance. None remain.

**Disputed and withdrawn — the authority-keyed cache does not fail requests.**

The finding claimed the shared `pkgs.dev.azure.com` cache slot "burns through the 4-retry budget, fails requests" in an "endless cache-thrashing loop." The mechanism is real and already tracked as #3540, but the consequence was overstated: `attempts` is scoped to a single `SendAsync` call, and re-acquisition uses the *current* request's URI, so the request recovers.

I tested it rather than argued. Four alternating requests across two orgs: **all 200 OK**, exactly **2 transport attempts each**, never approaching `MaxAuthRetries`. The reviewer withdrew the block — *"I misread the scope of the `attempts` counter... I withdraw the block."* Real cost is one wasted round trip and one extra plugin call per alternation. #3540 stays open as a follow-up.

### Round 2 twice found that test to be weaker evidence than I claimed

This was the most useful part of the second round, and both escalations were correct.

1. The original assertions would have been satisfied by a handler that **never cached at all**. Added a control: repeating the org that just ran costs exactly 1 attempt and 0 acquisitions (`7a2f7da2b`).
2. Even so, a **single global credential slot** would have satisfied every assertion, because everything targeted one host. That gap mattered beyond hygiene — "one feed's credential is never offered to a different host" is the guarantee this handler exists to keep, and nothing was watching it. The transport now records the `Authorization` offered on every attempt, and the test reaches a second authority and requires the first attempt there to carry **nothing** (`45b0e53bc`). The foreign host reuses the same `org-b` path segment, so the two differ only by authority.

Since I had used that test to argue a blocking finding down to non-blocking, its discriminating power was load-bearing. It is now pinned in both directions.

### Follow-ups filed, not fixed here

#3539, #3540, #3541, #3542, #3543, #3544, #3545, #3551. Each was verified in code before filing rather than forwarded on the reviewer's word. Two worth calling out:

- **#3539** — `NuGetClient.ApplyCredential` has the same ungated shape that was fixed in `PackageExtractor`. Latent today (no production callers outside `NuGetFetch`), but the layering makes it a trap: `NuGetCredentialScope` lives *above* `NuGetFetch`, so `NuGetClient` cannot call the gate even if it wanted to.
- **#3551** — the mirror of the bug fixed here, on inbound messages: `HandleInboundRequestAsync` omits `JsonException` from its catch filter while both branches deserialize plugin payloads. Less severe (confined to one message, process survives), but the handshake is documented as mandatory, so a throw leaves the plugin waiting and the fault unobserved. Left as follow-up because the real repair is answering the plugin with an `Error` response, which is a behavior change.

One reviewer claim I did not accept: that `Path_MatchesThePrefixCaseSensitively` was a vacuous pass on Windows. `IsPathPluginName` tests the prefix before the extension, so casing was always the operative rejection. Detail in the previous comment.

### Validation at `45b0e53bc`

| Suite | Result |
| --- | --- |
| `NuGetFetch.Tests` | 265 total, 0 failed |
| `DotnetInspector.Services.Tests` | 357 total, 0 failed |
| `dotnet-inspect.Tests` | 2559 total, 1 failed (pre-existing #3506), 14 skipped |

All 7 CI checks green at this head. `MERGEABLE`.

richlander:
Ready to merge

All merge-blocking validation, CI, and required review are complete at head `45b0e53bc`.

**Review** — high-risk tier, two fixed-head rounds from two different model families, both clean at this exact head:

| Family | Verdict |
| --- | --- |
| Gemini 3.1 Pro | clean |
| MAI-Code | clean |

**CI: 7/7 green** — `changes`, `build-net10`, `test (ubuntu-26.04, linux-x64)`, `pack`, `il-diff-smoke`, `csharp-diff-smoke`, `markdownlint`. `MERGEABLE`.

**Suites** (after a full `dotnet build dotnet-inspect.slnx -c Release`):

| Suite | Result |
| --- | --- |
| `NuGetFetch.Tests` (offline) | 265 total, 0 failed |
| `NuGetFetch.Tests` (full) | 292 total, 0 failed, 9 skipped |
| `DotnetInspector.Services.Tests` | 357 total, 0 failed |
| `dotnet-inspect.Tests` | 2559 total, 1 failed, 14 skipped |

The single CLI failure is #3506, failing identically on `main`.

**Merge order:** this is the child of the stack. #3423 merges first; it is separately marked ready and green at `d232ae808`.

**Non-blocking follow-ups**, all filed and verified in code, none gating this PR: #3539, #3540, #3541, #3542, #3543, #3544, #3545, #3551.

richlander:
## Windows verification — green on `main` @ `ec7b254e`

Closing the loop on my earlier offer to run these on Windows, since this repository has no Windows CI job and the other agent had no Windows access.

I originally reported **6 failing `PluginDiscoveryTests`** here. They are fixed on current `main`:

| run | total | failed | errors | skipped |
| `NuGetFetch.Tests.PluginDiscoveryTests` | 12 | **0** | 0 | 1 |
| full `NuGetFetch.Tests` | 292 | **0** | 0 | 23 |

The fix is the right shape rather than a blanket platform exclusion: **11 of the 12** discovery tests now genuinely execute on Windows, and the single skip states an accurate reason rather than hiding a gap —

> `Path_SkipsNonExecutableFiles` — *"The execute bit is a Unix concept; Windows filters by extension."*

That is a real platform difference, not an unimplemented behavior. The `PluginProtocolTests` skips are correct for the same kind of reason (*"the fake plugin is a shell script"*), as are the live Azure DevOps skips, which gate on `DOTNET_INSPECT_TEST_AZDO_FEED` / `DOTNET_INSPECT_TEST_AZDO_TOKEN`.

### How it was run

```
dotnet run --project src/NuGetFetch.Tests -c Release
dotnet run --project src/NuGetFetch.Tests -c Release -- -class NuGetFetch.Tests.PluginDiscoveryTests
```

Windows 11, `main` @ `ec7b254e`, Release, clean worktree with no local modifications. Also verified identical results at `0e4c303c`; the three commits in between touch no `NuGetFetch` files.

### One environment caveat worth recording

Building this repository on Windows currently requires pinning the SDK. README's `dotnetup sdk install 11.0-daily` produces `11.0.100-rc.1.26380.103`, which pins every runtime pack to its own version, while `nuget.config` is hermetic (`<clear />`, then nuget.org only) and nuget.org's newest is `preview.6`. Restore therefore fails:

```
error NU1102: Unable to find package Microsoft.NET.ILLink.Tasks with version (>= 11.0.0-rc.1.26380.103)
  - Found 99 version(s) in nuget.org [ Nearest version: 11.0.0-preview.6.26359.118 ]
```

Pinning `11.0.100-preview.6.26359.118` via `global.json` restores and builds cleanly. Not a defect in this PR, and I will file it separately against the SDK guidance.

Issue #3406, opened 2026-07-28
Decide what 'package --dependencies --count' should count


richlander:
`package --dependencies --count` currently exits 1 with:

```text
Error: --count requires -S/--select to match exactly one section.
```

This is **not** a silent projection drop — it fails loudly, and #3405 leaves it
unchanged deliberately. But the remedy the message suggests does not describe
what the caller asked for, and unlike the other package lenses (`--versions`,
`--tfms`, `--layout`, `--readme`) there is no obvious right count to give.

Three answers are each defensible, and they disagree by a lot. For
`Newtonsoft.Json@13.0.4`:

| Answer | What it counts | Value |
| --- | --- | --- |
| Declared dependencies | direct deps of the selected TFM group | 22 |
| Rendered roots | top-level tree nodes after display dedup (`globalSeen` in `ResolveDependencyTreeAsync`) | 16 |
| Section rows | `-S Dependencies --count`, all TFM groups | 136 |

The precedent set for the other lenses is that **`--count` counts the lens
payload, not the lines used to render it** — which argues for the declared
count, and against the rendered-roots count that exists only because the
display dedups across groups.

But `--dependencies` differs from the other lenses in that its payload is a
*tree*, not a flat list, so "the payload" is genuinely ambiguous in a way
`--tfms` is not: it could reasonably mean direct dependencies or the transitive
closure.

Deciding this is a contract choice, not a bug fix, so it is filed separately
rather than settled inside #3405.

richlander:
Implemented in #3430 (`5a4414d2`).

**The answer is one row per dependency edge over the whole resolved forest.**

#3405 reserved this because there were three defensible numbers. Measuring `Microsoft.Extensions.Logging@9.0.0`:

| candidate | value | what it counts |
| --- | --- | --- |
| declared dependencies | 22 | the manifest, not the lens |
| rendered roots | 16 | a slice of the **rendering** |
| `-S Dependencies` rows | 21 | a **different selection** |
| **graph edges (chosen)** | **5** | the lens payload's row lowering |

#3405's rule settles it: `--count` answers the **row lowering of the lens payload, not the lines used to render it**. That is the same rule the call-graph lens already follows.

It rejects the other two for concrete reasons, not by preference:

- **Rendered roots** counts which nodes survived `globalSeen` dedup. That is a property of the printer — change the dedup and the count changes, while the dependency graph did not. Counting it would make `--count` a fact about presentation.
[...]

PR #3429, opened 2026-07-29
Make -S Dependencies and -S References collect the data they name

richlander:
`-S Dependencies` and `-S References` reported "no data" for every assembly:

```console
$ dotnet-inspect System.Text.Json -S "Dependencies"
Note: 1 matched section has no data: Dependencies.
```

`-D` listed both sections as available, so discovery promised data that `-S`
could not deliver.

## Root cause

Three independent breaks in one path. Both sections read assembly references,
which are extracted on demand rather than always:

1. `Dependencies.ScannerKey` named a `"TransitiveRefs"` scanner that was **never
   registered** in `CreateScannerRegistry()`. Selecting the section requested a
   scanner that did not exist, which is a silent no-op.
2. `Dependencies.CanRender` additionally required `UseDependenciesView` — the
   routing flag for the separate `--dependencies` view, which `-S` cannot set.
   The section could never render through selection at all.
3. `References.ScannerKey` was `null`, which the pipeline reads as "always
   collected". That is false for references: they are extracted only when
   `IncludeReferences || IncludeDependencies || needsAuditSignals`. `-S
   References` appeared to work at `-v:d` only incidentally, because the
   audit-signals scanner forces the same extraction.

The shared bug class is a scanner key that does not describe reality — declared
but unregistered, or `null` on data that is collected conditionally.

## Change

Register the transitive-reference scanner, consult both scanner keys when
deciding whether to extract references, and gate each section on its own data
rather than on a view-routing flag.

`References` no longer blanks itself when a dependency tree exists. That mutual
exclusion existed so the `--dependencies` view would not print the flat list
under the tree — but that view returns before sections render, and once both
sections are selectable it instead suppressed `References` whenever the two were
selected together.

`Dependencies` is declared `Verbose`/`Unbounded` on the axes added in #3229 and
`ExplicitOnly` besides: the closure grows without a meaningful bound and is built
by resolving and reading every referenced assembly recursively, so no verbosity
ever auto-runs it and it stays out of `-S @All`. It is reached through a new
`@Dependencies` category door, which lists it next to `References` — the section
a reader comparing the two lenses wants beside it anyway.

## Evidence

All four now render, where each printed "no data" before:

```console
$ dotnet-inspect System.Text.Json -S Dependencies          # transitive tree
$ dotnet-inspect System.Text.Json -S References            # reference table
$ dotnet-inspect System.Text.Json -S "References,Dependencies"   # both
```

**Every automatic view is byte-identical to `main`.** Swept 5 assemblies
[...]

Issue #3427, opened 2026-07-29
Bind row projections to a graph's row lowering, not its rendered tree


richlander:
Graph-shaped sections render as a tree by default and can lower to rows. Today
the row projections (`--count`, `--row`, `--rows`, `--fields`) bind to the
*rendered tree text* rather than to the row lowering, so the same flag answers
correctly in one lowering and silently wrongly — or not at all — in another.

## The defect in one table

One graph (`Call Graph` on `JsonSerializer.Serialize:1`), one flag, two
lowerings:

| Command | Result | Correct |
| --- | --- | --- |
| `-S "Call Graph" --table --rows -n 3` | 3 rows | yes |
| `-S "Call Graph" --rows -n 3` | 25 nodes, unfiltered | **silently ignored** |
| `-S Dependencies --rows -n 3` | 27 nodes, unfiltered | **silently ignored** |

And counting, on `main` today:

| Command | `main` | Actual |
| --- | --- | --- |
| `-S "Call Graph" --count` | 0 | 25 nodes |
| `-S "Call Graph" --table --count` | 0 | 24 edges |
| `-S Dependencies --count` | 0 | 27 nodes |

The zeros are the visible half. The invisible half is that a fix which counts
the *tree text* returns 25 for `--table --count`, where the table has 24 rows —
a plausible wrong number rather than an obvious broken one. I know because I
wrote that fix while repairing an unrelated bug, and it is what prompted this
issue. Counting rendered text cannot be right, because the text is not
necessarily the shape the caller selected.

This is the same failure #3404 described for rows: *"It returns a real row, so
nothing looks broken."*

## Why this recurs

The projections are bound to output rather than to the model. #3405 already
states the rule for lens paths:

> `--count` counts the lens payload … **not the lines used to render it**

and refuses shape projections rather than "inferring structure from rendered
text." Section paths need the same rule, and graph sections are where it bites
hardest, because a graph has more than one legitimate row reading.

## Half of the machinery already exists

PR #3291 built `CallGraphProjection` (`src/ILInspector.CallGraph/CallGraphProjection.cs`),
which already guarantees exactly what row addressing requires:

- typed identity — *"hosts must not infer identity from display text"* (line 36),
  the same rule #3405 applies to counting
- a dense zero-based node id (line 31)
- an explicit edge list, `Nodes` and `Edges` (lines 101, 104)
- *"Ordering is part of the contract, not an implementation detail"* (line 87)

[...]

PR #3430, opened 2026-07-29
Lower graphs to rows: --count answers the model, not the rendering (#3427, #3406)

richlander:
> **Stacked on #3429.** Review that first; this PR's diff is the top commit only.
> Prototype for #3427 — phases 1 and 2 of four.

A graph section renders as a tree, a row list, or a diagram. `--count` answered
by scanning rendered output, so it reported `0` for every graph section, and any
fix that counts the *tree text* reports the tree's node count even when the
caller selected the edge-list lowering.

## The defect

Same graph, same flag, on `main` today:

| | result |
| --- | --- |
| `-S "Call Graph" --count` | 0 |
| `-S "Call Graph" --table --count` | 0 |
| actual `--table` data rows | 24 |

The zeros are the visible half. The invisible half is that counting the tree
text returns **25** for `--table --count` — the node count, where the table has
24 rows. I know because I wrote that fix while repairing #3429 and measured it
here; a plausible wrong number is worse than an obvious broken one. It is the
failure #3404 named: *"It returns a real row, so nothing looks broken."*

## Change

**Phase 1** gives `CallGraphProjection` a declared row unit. Which element is a
row is a property of the graph, not a global truth: a call graph is read as
"what calls what", so its rows are edges — matching the edge list `--table`
already renders. A dependency graph is read as "what does this depend on", so
its rows will be nodes.

`RowCount` is *derived* from `RowUnit` through a switch rather than restated, so
the declaration drives the count and the two cannot drift.

**Phase 2** binds `--count` to that declaration on the member path. The count is
carried on the view as `CallGraphRowCount` and answered before serialization, so
no output text is scanned.

```
-S "Call Graph" --count          0 -> 24
-S "Call Graph" --table --count  0 -> 24
actual --table data rows              24
```

The count no longer varies with the rendering the caller selected. That is the
point: `--count` asks how many rows the section has, and a graph has one row
lowering however it is displayed.

## Why this shape

This follows the direction #3405 sets for lens paths — *"`--count` counts the
lens payload … not the lines used to render it"* — and extends it to section
paths, where a graph has more than one legitimate row reading. It also supplies
what #3404 requires of a row number: one assigned by the model in a contractual
order, not re-derived from a position.

`CallGraphProjection` already guaranteed the hard parts (typed identity, dense
ids, *"ordering is part of the contract"*); this only names the row unit.

## Evidence

CLI suite: **2328 passed, 0 failed, 14 skipped**.

`Member_CallGraphCount_AnswersRowLoweringNotRenderedTree` derives its expected
value from the lowering rather than pinning `24`, so a row-unit change cannot
silently pass it. Verified non-vacuous by stashing `ApiCommand.cs`:

```
```

## Not in scope

- **Phase 3** — `--row`/`--rows` binding, so the tree honours them carrying the
  same numbers as the table. `--rows -n 3` is silently ignored on tree lowerings
  today.
- **Phase 4** — an equivalent projection for the dependency graph, whose
  `AssemblyReferenceNode` list encodes structure as a depth `int` with no ids or
  edges.

No shared graph abstraction is extracted yet. With one implementation it would
be speculative; phase 4 is when a second consumer justifies it.


richlander:
Also resolves #3406 (`5a4414d2`).

`package <id> --dependencies --count` used to exit 1. It now answers **one row per dependency edge over the resolved forest** — the same rule this PR applies to the call graph, and the rule #3405 states: a count answers the row lowering of the lens payload, not the lines used to render it.

Rationale and the measured alternatives are on [#3406](https://github.com/richlander/dotnet-inspect/issues/3406#issuecomment-5113468486). Short version: rendered roots counts `globalSeen` dedup (a printer property), and `-S Dependencies` is a different selection whose 21 rows stay unchanged.

This is the second graph type to get a row lowering, and it needed no new abstraction — `DependencyNode` is already a real recursive tree, so the count is a walk of the model. That is some evidence the "graph stays a graph, rendering is a lowering" shape generalizes.

Phase 3 (`--row`/`--rows`) remains blocked on the Markout row-identity fork recorded on #3427.

richlander:
## Phase 3 is now blocked on an upstream issue, filed: richlander/markout#175

You asked which component did the call-graph lowering to tables. I read #3280's
implementation rather than answer from memory, and the answer changed my
recommendation.

The lowering is split three ways and only the first part is domain-specific:

| Layer | Owns | Domain-specific? |
| --- | --- | --- |
| `ILInspector.CallGraph` (`CallGraphProjection`) | graph *semantics* — edge direction, node identity, ids/ordering, cycle collapse, external/truncated classification | **yes** |
| `dotnet-inspect` (`CallGraphSectionAdapter.ToGraph`) | *vocabulary* only — member spelling, `(external)`, `…`, `--fields` cues. Returns `Markout.Graph` | boundary; domain knowledge is spent here |
| **Markout** (`GraphLowering`) | tree, **edge table**, Mermaid | **no** — generic nodes + edges |

So the component that lowered the call graph to a table is **Markout**.
`CallGraphSectionAdapter.ToGraph` is the only producer of a `Markout.Graph` in
the CLI, and nothing in `src/dotnet-inspect/Output/` turns a `Graph` into rows —
I checked. `docs/design/call-graph-projection.md` says the domain layer "knows
nothing about Mermaid, Markdown, tables… takes no dependency on Markout," and
each host owns rendering.

**That argues against my earlier plan of owning the row lowering here.** It would
leave the call-graph table lowered by Markout and the dependency table lowered by
us — two implementations of one lowering, which #3439 seam rule 6 calls a defect.

### What the upstream issue actually reports

I probed `GraphLowering` on `markout@main` with a graph shaped like a real call
graph (cycle through the focus, one shared node, one isolated node). The two
lowerings do not merely order things differently — they **enumerate different
populations**:

- `ToEdgeTable` → **6 rows** = 5 edges + 1 isolated node
- `ToTree` → **8 lines** = 3 roots + 5 edges

`Focus` and `CallerA` are tree lines with **no table row at all**. Edge 1→0 is
table row `0` and tree line `6`. And neither lowering reports provenance —
`GraphEdgeTable.Rows` is `List<string[]>`, `TreeNode` carries only
`Text`/`Badge`/`Children`/`State` — so the mapping cannot be recovered even by
walking both.

That is why `--count` over a graph section has no defensible answer today: the
two lowerings disagree about what there is one of.

### Status

Phases 1–2 are done and stand on their own — that is the half that fixed #3406,
where `package --dependencies --count` now answers **5** (one row per dependency
edge over the resolved forest) instead of a printer property. Re-verified after
folding into #3405's `rendersOwnPayload`: **5 / 21 / 0**.

Phase 3 stays in draft pending markout#175. I offered three options there
including "no change upstream", in which case I keep row projections off graph
sections here rather than invent a correspondence Markout does not promise.

richlander:
Closing with the #3429 stack. The call-graph row-projection slice landed independently in #3966, binding `--count` and row windows to edge rows across renderers. The unlanded legacy `package --dependencies --count` compatibility question remains tracked by #3406; the preferred `depends --package --count` path already supports counting.

PR #3435, opened 2026-07-29
Exclude NuGet unlisted versions from enumeration and latest resolution (#3388)

### LINKED ISSUE #3388: Version enumeration/resolution includes NuGet unlisted versions (flat-container index.json has no listed flag)
## Summary

`dotnet-inspect` enumerates and resolves package versions from the NuGet
**flat-container** `index.json`, which lists **every** published version
including **unlisted** ones and carries no `listed` flag. As a result unlisted
versions surface in version listings, and on the flat-container code paths
"latest" can resolve to an unlisted version. This should be handled centrally
so every consumer (CLI and the WASM engine) shares one listing-aware policy.

## Concrete example

`Markout` has 71 published versions; exactly one — **`10.0.2`** — is unlisted
and does not appear on the NuGet gallery.

```
# registration index carries per-version listed status
$ curl -s https://api.nuget.org/v3/registration5-gz-semver2/markout/index.json | gunzip | ...
10.0.2 -> listed: False
0.30.0 -> listed: True

# flat-container index.json has NO listed flag and includes 10.0.2
$ curl -s https://api.nuget.org/v3-flatcontainer/markout/index.json
{ "versions": [ "0.1.0", ..., "0.30.0", ..., "10.0.2" ] }

# stable-latest picked from flat-container (LastOrDefault !prerelease) => UNLISTED
flatcontainer stable-latest: 10.0.2
# listing-aware search API => correct
search API latest:          0.30.0
```

## Where this happens (central code)

- `src/DotnetInspector.Packages/PackageExtractor.cs`
  - `FetchAllVersionsFromSourceAsync` / `FetchVersionListAsync` enumerate
    versions straight from flat-container `index.json` — unlisted versions are
    included with no way to distinguish them.
  - `GetLatestVersionFromSourceAsync`: the **stable** nuget.org path uses the
    search API (`azuresearch-usnc.nuget.org`), which already excludes unlisted
    versions and is correct. But the **flat-container fallback** and the
    **`includePrerelease`** path (`ParseVersionIndexAsync`) can select an
    unlisted version as "latest".
- Registration metadata is already fetched elsewhere in the product
  (`src/DotnetInspector.Services/PackageMetadataService.cs` uses
  `registration5-semver1`), so the `catalogEntry.listed` signal is available —
  it just isn't consulted during version enumeration/resolution.

The same defect exists in the WASM prototype engine
(`prototypes/inspect-web/engine/Program.cs` `ResolvePackageVersionAsync`), and
the new prototype version selector currently lists unlisted versions. Rather
than patch the prototype UI, we want the fix to live centrally.

## Impact

- Version listings show versions that are hidden on nuget.org.
- On flat-container / prerelease code paths, default/"latest" resolution can
  land on an unlisted version (e.g. `Markout` "latest" resolves to the unlisted
  `10.0.2` instead of `0.30.0`).

## Proposed central handling

- Consult the NuGet **registration** index (`catalogEntry.listed`) when
  enumerating and resolving versions, and filter out (or explicitly flag)
  unlisted versions in one shared place — ideally the version-resolution
  service described in `docs/design/version-resolution.md` — so the CLI and the
  WASM engine share a single policy.
- **Preserve explicit access**: an unlisted version requested by exact
  `Package@Version` must still resolve and load (NuGet itself allows restoring
  a known unlisted version); only *discovery* (enumeration + "latest") should
  hide it.
- Expose the `listed` bit on the version model so surfaces that want to show
  unlisted versions can mark them rather than silently including them.

## Repro environment

- Observed via the `prototype/wasm-command-ui` version selector; confirmed
  against live nuget.org registration + flat-container APIs (commands above).

richlander:
## Summary

Fixes #3388. NuGet's flat-container `index.json` lists **every** version including **unlisted** ones, with no `listed` flag. Previously, unlisted versions leaked into enumeration and — worse — could be auto-selected as "latest". This PR makes version resolution **listing-aware**: unlisted versions are excluded from enumeration and from flat-container/prerelease latest selection, using nuget.org's registration index as the authoritative source of the `listed` bit.

## Behavioral change

- **Enumeration & raw listing** (`GetVersionsAsync`, `package Name --versions`): unlisted versions are filtered out. If the nuget.org registration read fails, this path **fails open** (returns the unfiltered list) — better to show a real version than to drop it during a transient outage.
- **Auto-selecting paths** (`GetLatestVersionFromSourceAsync`, wildcard `ResolveVersionPatternAsync`): these **fail closed** (return null) when the version list is non-authoritative. Silently resolving to an unlisted version is the actual #3388 harm, so we refuse rather than guess.
- **Explicit pinned access preserved**: a fully-pinned coordinate (`pkg@X`) still resolves even if that version is unlisted — explicit access is not discovery.
- **Non-nuget.org sources unchanged**: filtering only applies where nuget.org's registration provides an authoritative `listed` bit.

## Implementation notes

- `GetAllVersionsWithCacheAsync` now returns `(List<string>? Versions, bool Authoritative)`; only authoritative (filtered) lists are cached.
- Uses the SemVer2 registration hive (`registration5-gz-semver2`) so unlisted SemVer2 prereleases are also detected.
- `CollectUnlisted` treats absent/`true` `listed` as listed, `false` as unlisted, and **throws** on any other JSON value kind (string, number, null, object, array) — a malformed/hostile feed thus fails open, and the auto-select path then fails closed.
- Version cache category bumped `versions` → `versions-v2` so pre-upgrade unfiltered entries are never trusted.

## Evidence

- Services tests: 330 pass. CLI suite: green.
- New regression tests: `ResolveVersionPattern_ReturnsNull_WhenRegistrationUnavailable`, `GetLatestVersion_ReturnsNull_WhenListedValueIsMalformed`, plus fail-open/fail-closed vectors in `UnlistedVersionTests`.
- Live-verified against System.CommandLine unlisted SemVer2 prerelease `2.0.0-beta3.22101.1`.

## Adversarial review

High-risk change; reviewed across five rounds by two model families per round (Gemini + GPT). Full reconciliation posted as a follow-up comment.

### DIFF
--- a/docs/design/version-resolution.md
+++ b/docs/design/version-resolution.md
@@ -153,6 +153,33 @@ dotnet-inspect diff --package Foo@1.4.0..1.5.0 \
 kind and producer detail establish identity, while IL offsets remain local to
 each endpoint.
 
+## Listed vs. unlisted versions
+
+NuGet lets a publisher **unlist** a version: it stays restorable by exact
+coordinate but is hidden from discovery on nuget.org. The flat-container
+`index.json` that drives version enumeration lists **every** published version
+and carries no listed flag, so it cannot distinguish an unlisted version on its
+own. Only the nuget.org **registration** index exposes the per-version
+`catalogEntry.listed` bit.
+
+Version resolution applies one shared listing-aware policy so discovery matches
+the nuget.org gallery:
+
+- **Enumeration** (`Name --versions`, wildcard resolution) and the
+  **flat-container / prerelease "latest"** paths consult the registration index
+  and drop versions whose `catalogEntry.listed` is explicitly `false`. The
+  stable "latest" path already uses the listing-aware search API and is
+  unaffected. This is nuget.org-only; other feeds have no listed concept and are
+  returned unfiltered.
+- **Explicit access is preserved.** A pinned `Name@Version` (including
+  `Name@latest` and the addressable-vector endpoints) never enumerates, so a
+  known unlisted version still resolves and loads — matching NuGet's own
+  behavior of restoring a known unlisted version.
+- **Fail-open.** If the registration index cannot be fetched or parsed, the list
+  is returned unfiltered rather than silently dropping versions; the condition
+  is logged. The filtered nuget.org list is what gets cached, so a cache hit
+  within the 1-hour TTL never re-surfaces an unlisted version.
+
 ## Cache locations
 
 | Cache | Location | TTL | Written by |
--- a/docs/design/version-resolution.md
+++ b/docs/design/version-resolution.md
@@ -160,7 +160,11 @@ coordinate but is hidden from discovery on nuget.org. The flat-container
 `index.json` that drives version enumeration lists **every** published version
 and carries no listed flag, so it cannot distinguish an unlisted version on its
 own. Only the nuget.org **registration** index exposes the per-version
-`catalogEntry.listed` bit.
+`catalogEntry.listed` bit. Resolution reads the SemVer2 registration hive
+(`registration5-gz-semver2`) rather than the SemVer1 hive, because the SemVer1
+hive omits SemVer2 versions entirely — reading it would let unlisted SemVer2
+prereleases escape filtering. That hive is gzip-encoded and transparently
+decompressed by the shared HTTP client.
 
 Version resolution applies one shared listing-aware policy so discovery matches
 the nuget.org gallery:
@@ -175,10 +179,14 @@ the nuget.org gallery:
   `Name@latest` and the addressable-vector endpoints) never enumerates, so a
   known unlisted version still resolves and loads — matching NuGet's own
   behavior of restoring a known unlisted version.
-- **Fail-open.** If the registration index cannot be fetched or parsed, the list
-  is returned unfiltered rather than silently dropping versions; the condition
-  is logged. The filtered nuget.org list is what gets cached, so a cache hit
-  within the 1-hour TTL never re-surfaces an unlisted version.
+- **Fail-open.** If the registration index cannot be fetched or parsed (network
+  failure, or a valid-JSON document whose shape defies the expected schema), the
+  list is returned unfiltered rather than silently dropping versions, and the
+  condition is logged. A fail-open (unfiltered) snapshot is **not** cached, so a
+  transient registration outage cannot re-surface unlisted versions for the
+  cache TTL; only an authoritatively filtered list is persisted. For the same
+  reason, nuget.org "latest" resolution returns no result rather than falling
+  back to an unfiltered index when the listing-aware path cannot produce a list.
 
 ## Cache locations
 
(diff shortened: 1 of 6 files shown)


richlander:

High-risk change → reviewed across **five rounds**, two model families per round (Gemini `gemini-3.1-pro-preview` + GPT `gpt-5.6-sol`), each in an isolated fixed-head worktree. Every round 2–4 surfaced a real bug; round 5 came back fully clean. Reviews held across a clean rebase onto latest `main`.

### Blocking findings — all resolved

| # | Reviewer(s) | Finding | Resolution |
|---|---|---|---|
| 1 | Gemini + GPT | `GetLatestVersionFromSourceAsync` discarded the Authoritative flag → a fail-open registration outage yielded unfiltered/unlisted "latest". | `if (listed != null && authoritative)` + regression test. Commit `f2619277`. |
| 2 | GPT | SemVer1 registration hive excluded SemVer2 unlisted prereleases (they still leaked). | Switched to `registration5-gz-semver2`. Commit `0f3039ca`. |
| 3 | Gemini | Wildcard `ResolveVersionPatternAsync` auto-select dropped authoritativeness → could resolve unlisted during outage. | Tuple return + fail-closed + test. Commit `ad708270`. |
| 4 | Gemini + GPT | Valid-but-wrong-shape JSON (`{"items":{}}`) threw `InvalidOperationException`, escaping the `catch(JsonException)` fail-open. | Widened catch to cover both; validate `ValueKind`. Commit `0f3039ca`. |
| 5 | GPT (`gpt-5.6-sol`) | `CollectUnlisted` treated a non-boolean `listed` (e.g. string `"false"`) as listed+authoritative → a hostile feed could smuggle an unlisted version as latest. | Throw on any non-boolean `listed` → fail open → auto-select fails closed. + test. Commit `9ce390ad`. |

### Non-blocking / out-of-scope — dispositioned

- **Pre-upgrade cache trusted for TTL** (Gemini, GPT): resolved by the `versions` → `versions-v2` category bump — pre-fix unfiltered entries are never read.
- **Cleanup glob leaves bare `versions/` dir** (Gemini, Medium): **wontfix** — consistent with the repo-wide `-v` prefix convention (`pkg-index-v`, `effective-v`, `packs-v` all orphan their bare predecessor); orphaned data is never *read*, only inert on disk.
- **`limit<=0` returns 1 element** (Gemini, Low) and **unparseable versions sort as newest** (Gemini, Medium): **out of scope** — both are byte-identical on `origin/main`, unchanged by #3388, and unreachable for nuget.org (all valid SemVer).

### Round-5 verdict

Both reviewers: **clean**. Confirmed the malformed-`listed` fix is correct, non-vacuous (verified against the pre-fix commit), catches the exception, and that `listed: null` correctly fails closed as designed.

richlander:
Ready to merge

All merge-blocking checks green: `test` (9m27s), `build-net10`, `markdownlint`, `changes` all pass; `csharp-diff-smoke`/`il-diff-smoke`/`pack` correctly skipped. Five rounds of two-family adversarial review are clean (reconciliation above). Local: Services 330, CLI suite green.

Issue #3426, opened 2026-07-29
Give the wasm engine a product API for call graphs: it reimplements scope selection, retention, member resolution and forwarding


richlander:
## Why this is the point of the project

The wasm site (`origin/prototype/wasm-command-ui`, `prototypes/inspect-web/engine/Program.cs`,
2868 lines) is blocked on product API, and a substantial part of that file is expected to be
**deleted** once the API exists. It is not an incidental consumer — it is the motivating one.

`dotnet-inspect` should own this work. Today the engine reimplements it against raw
`LibraryBodyIndex` / `AssemblyInspectionSession`, which is the consumer-side equivalent of the
harness-boundary rule in `AGENTS.md`: a consumer covering for missing product capability hides the
gap and creates a second implementation.

## The concrete trigger: the perf track does not reach its own consumer

```csharp
var callerScopes = workspaceAssemblies
    .Select(a => Analysis.LibraryBodyIndex.Open(a.Path, LibraryBodyAnalysisFeatures.MethodEvidence))
    .ToArray();
var callers = index.BuildCallerTree(token, maxDepth: 2, maxNodes: 30);
if (callerScopes.Length > 0)
    callers = index.BuildCallerTree(token, callerScopes, maxDepth: 2, maxNodes: 30);
```

Grep of that file for `CallerScopeFilter`, `CallerScopeTypeFilter`, `SelectCouldReach`,
`ApiMemberAnalysisInspection`, `SelectedScopePaths` returns **nothing**.

| #3333 increment | implemented in | reaches wasm? |
| --- | --- | --- |
| 1 — AssemblyRef prefilter (#3331) | `ApiMemberAnalysisInspection` (CLI) | **no** |
| 1b — scope-graph caching (#3342) | `LibraryBodyIndex` (Analysis) | yes |
| 2 — TypeRef narrowing (#3409) | `ApiMemberAnalysisInspection` (CLI) | **no** |

Every workspace assembly is opened and passed as a caller scope. In "Platform" mode the workspace
is the platform library set (`assets/platform-index.tsv`, 1760 rows) — the shape #3333 baselined at
703-965 MB.

`ProgressiveMemberCallGraph` shows the same pattern from the other side: it lives in
`src/dotnet-inspect/Inspectors/`, and repo-wide it has **no callers outside its own test file**.
The progressive seam was built, placed in the CLI layer, and the consumer that needed it wrote
`ExpandPlatformCallGraph` instead.

[...]

Issue #3419, opened 2026-07-28
Callers matcher compares TypeRef identities instead of resolved definitions


richlander:
## Summary

`MemberPattern.MatchesCrossAssembly` compares a call site''s decoded declaring `TypeRef` against the target''s. When a caller reaches a type through a **facade that type-forwards** to the defining assembly, the caller''s `TypeRef` names the *facade*, not the definer, so the comparison fails and the caller is never reported in the `Callers` table.

This is **pre-existing** and was **not** introduced by #3409. It surfaced while adversarially reviewing that PR, and is recorded here so the limitation is tracked rather than folded into an unrelated performance change.

## Reproduction

Three assemblies:

- `B` defines `public class TheType { public void Do() {} }`
- `A` references `B` and declares `[assembly: TypeForwardedTo(typeof(TheType))]`
- `Caller` references **only** `A`, and calls `new TheType().Do()`

`Caller`''s TypeRef for `TheType` has resolution scope `A`:

```text
PROBE typeref TheType -> scope A
PROBE target declaring type = TheType
PROBE filter=DoesNotName matcherMatches=0
```

The target''s declaring type decodes with assembly `B`, the call site''s with assembly `A`, so `TypeRef.Equals` is false and no edge is emitted.

## Impact

A real caller is silently missing from `Callers`. `TypeRef.CanonicalAssembly` collapses the corelib alias set (`System.Private.CoreLib` / `System.Runtime` / `mscorlib` / `netstandard` / `System.Runtime.Extensions`), which is why the common framework facades happen to work. Any forwarder outside that fixed list does not.

## Note for whoever picks this up

`CallerScopeTypeFilter` (added in #3409) is deliberately **exactly as permissive as the matcher** — it decodes the same rows with the same `TypeRefDecoder`. It currently agrees with the matcher here, ruling the caller out for the same reason the matcher misses it, so today it is sound.

**If the matcher is taught to resolve forwarding, the filter must be widened in the same change**, or it will start ruling out assemblies the matcher would now match — turning a display gap into a wrong answer. `NothingRuledOutEverContributesAMatchingCallSite` compares the filter against the matcher itself and is the test that should catch that, provided a forwarding fixture is added to its corpus.

richlander:
## Reframed: the resolution half is fixed, the comparison half is what remains

[...]

PR #3437, opened 2026-07-29
Follow type forwarders when resolving cross-assembly TypeRefs

### LINKED ISSUE #3400: CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition fails deterministically on Windows: forwarder hop out of facade assemblies never resolves

richlander:
## What

`LibraryBodyIndex.IndexBuilder` now follows type forwarders when resolving a cross-assembly `TypeRef`.

A `TypeRef` names the assembly the compiler **bound against**, which for framework references is routinely a facade that defines no types and forwards to the definer. `TryResolveTopLevelExternalType` opened only the named assembly, scanned its `TypeDefinitions`, and returned null when the type was not there. Against a reference-assembly or facade-only shared-framework layout, that is *every* framework `TypeRef`.

```
TypeRef System.Object, scope = System.Runtime
  -> open System.Runtime          (facade: zero TypeDefs)
  -> scan TypeDefinitions         (no match)
  -> return null                  <- never reads ExportedType
```

## Why it is a delegation fix, not a new capability

The general capability already exists and is correctly factored: `ILInspector.Metadata.TypeForwardResolver` is public, bounds hops, detects cycles, and splits mechanism from policy via the one-method `IAssemblyReferenceResolver` — its own doc cites `MetadataLoadContext` and ILSpy as precedent. It is already consumed by the decompiler (`MemberBodyProducer`, `CrossAssemblyTypeResolver`) and by `PlatformResolver`.

`ILInspector.Analysis` **already references `ILInspector.Metadata`** and `LibraryBodyIndex.Open` **already accepts an `IAssemblyReferenceResolver`**. It had the seam, the reference, and the interface — it just did a one-hop open. The call graph was the only major consumer not using the shared resolver.

Decoding a forwarder reuses `TypeForwardResolver.ForwardTargetAssemblyIdentity`. The traversal runs in the builder rather than calling `LocateType` because the defining `MetadataReader` must outlive the call under `_referencedAssemblyCache`'s ownership, while `LocateType` disposes every assembly it opens. `DefaultMaxHops` is now a named constant on the resolver so the two traversals cannot drift apart in value.

## Scope is carried, and never loosened

The resolution scope is computed once from the originating reference and reused for every hop, matching `LocateType`. A `Platform` reference stays `Platform`, so a forwarded framework type cannot be satisfied by a confusable local copy.

## Blast radius: none today — this is enabling work

**Corrected after review.** An earlier revision of this description claimed the change "can surface analysis that was silently empty before." That overstates it, and the correction is the most interesting thing about this PR.

`TryResolveExternalTypeDefinition` has **no product callers**. Only its own recursion through `TryResolveNestedExternalType` and four test call sites reach it.

The resolver itself is fully plumbed. `ApiAnalysisInspection.CreateReferenceResolver` builds a real `AssemblyDependencyResolver` and threads it through `MethodBodyInspectionSession.Open` into `LibraryBodyIndex.Open`, including on the `Callers` path. But nothing consumes it during indexing, and the suite already pins that: `Open_WithResolverDoesNotChangeIndexShape` asserts `Assert.Equal(0, resolver.ResolveCalls)` after a full build with a resolver supplied.

So cross-assembly type resolution is **built, plumbed, and unconsumed**. This PR makes it correct; #3419 is the change that makes something call it. That also explains the zero-regression result below rather than being contradicted by it.

This is the same pattern as #3426's root cause and as `ProgressiveMemberCallGraph` (which likewise has no callers outside its own tests): capability exists, is wired most of the way, and the consumer that needs it reaches past it.

## Evidence

`CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition` was a **deterministic pre-existing failure on `main`** (#3400). Causally confirmed both directions on this branch: with the product change stashed and the tests kept it fails with the original assertion; with the change applied it passes.

Two tests added, both against real framework artifacts:

- `..._FollowsForwardersToDefiningAssembly` **names the mechanism**. It uses a resolver that returns exactly the assembly a `TypeRef` names (no implementation-preference redirect), self-selects a `TypeRef` whose named assembly *forwards rather than defines* the type, and asserts the definition arrives from a **different** assembly that does define it. **Verified non-vacuous** — reverting the product change fails it with `System.Object is forwarded by System.Runtime and must resolve through that forwarder`.
- `..._ForwarderCycleTerminatesWithoutResolving` answers every identity with the same facade so the chain cycles, and asserts termination with an honest null. This one guards the new loop rather than gating the fix: it also passes on `main`, where resolution returns null one step earlier for a different reason.

| suite | result |
| --- | --- |
| Analysis | **576 / 0 failed** (was 574 with #3400 red; +2 new) |
| Services | 319 / 0 failed |
| Metadata | 1233 / 1 failed |
| CLI | 2323 / 2 failed / 10 skipped |
| Decompiler (`--gate fast`) | 4317 / 5 failed |

**Every remaining failure was verified pre-existing on `origin/main` (d7f38a1a)** by stashing the product change and re-running each one: 1 Metadata (`AuthoredSourceValidity`, source extraction), 2 CLI (`LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceLinkDoor`, `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically`), 5 Decompiler (`ExpressionTreeLambda`, `StringSwitchRaising`, 3x `CorpusSensorComparison`). Zero regressions.

## Not in scope

- `ForwardTargetAssemblyIdentity` handles `Implementation.Kind == AssemblyReference`. Forwarders whose implementation is a nested `ExportedType` still resolve through `TryResolveNestedExternalType`, which now benefits from forwarding on the declaring type.
- #3419 (`MemberPattern.MatchesCrossAssembly` comparing decoded `TypeRef` identities rather than resolved definitions) is the same theme one layer up, and is the consumer that will exercise this path.

Fixes #3400.

### DIFF
--- a/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
+++ b/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
@@ -135,6 +135,87 @@ public class LibraryBodyIndexTests
     }
 
     [Fact]
+    public void CrossAssemblyMetadataResolver_FollowsForwardersToDefiningAssembly()
+    {
+        // A framework TypeRef names the assembly the compiler bound against, which is a
+        // facade that defines nothing and forwards to the definer. This resolver hands back
+        // exactly the assembly a TypeRef names -- no implementation-preference redirect --
+        // so the definition is reachable only by following the facade's ExportedType
+        // forwarder. Dropping that hop makes this return null, which was the #3400 bug.
+        string frameworkDir = Path.GetDirectoryName(typeof(object).Assembly.Location)!;
+        string targetPath = typeof(Console).Assembly.Location;
+
+        using var stream = File.OpenRead(targetPath);
+        using var peReader = new PEReader(stream);
+        Assert.True(peReader.HasMetadata);
+        var reader = peReader.GetMetadataReader();
+        using var builder = new LibraryBodyIndex.IndexBuilder(
+            targetPath, reader, peReader, new FrameworkDirectoryResolver(frameworkDir));
+
+        bool exercisedForwarder = false;
+        foreach (var handle in reader.TypeReferences)
+        {
+            var typeReference = reader.GetTypeReference(handle);
+            if (typeReference.ResolutionScope.Kind != HandleKind.AssemblyReference)
+                continue;
+
+            string referencedName = reader.GetString(
+                reader.GetAssemblyReference((AssemblyReferenceHandle)typeReference.ResolutionScope).Name);
+            string ns = reader.GetString(typeReference.Namespace);
+            string name = reader.GetString(typeReference.Name);
+            string fullTypeName = ns.Length == 0 ? name : $"{ns}.{name}";
+            string namedPath = Path.Combine(frameworkDir, referencedName + ".dll");
+
+            // Only a TypeRef whose named assembly forwards rather than defines exercises the hop.
+            if (!File.Exists(namedPath) || !TypeForwardResolver.ForwardsType(namedPath, fullTypeName))
+                continue;
+
+            var definition = builder.TryResolveExternalTypeDefinition(handle);
+            Assert.True(
+                definition is not null,
+                $"{fullTypeName} is forwarded by {referencedName} and must resolve through that forwarder");
+
+            var definingReader = definition!.Value.DefiningReader;
+            string definingName = definingReader.GetString(definingReader.GetAssemblyDefinition().Name);
+            Assert.NotEqual(referencedName, definingName);
+            Assert.True(
+                TypeForwardResolver.DefinesType(Path.Combine(frameworkDir, definingName + ".dll"), fullTypeName),
+                $"{definingName} should define {fullTypeName}");
+            Assert.Equal(name, definingReader.GetString(
+                definingReader.GetTypeDefinition(definition.Value.Definition).Name));
+
+            exercisedForwarder = true;
+            break;
+        }
+
+        Assert.True(exercisedForwarder, "expected a framework TypeRef whose named assembly forwards the type");
+    }
+
+    [Fact]
+    public void CrossAssemblyMetadataResolver_ForwarderCycleTerminatesWithoutResolving()
+    {
+        // Answering every identity with the same facade turns the forwarder chain into a
+        // cycle. Resolution must terminate and report nothing rather than spin or invent
+        // a definition.
+        string frameworkDir = Path.GetDirectoryName(typeof(object).Assembly.Location)!;
+        string facade = Path.Combine(frameworkDir, "netstandard.dll");
+        Assert.True(File.Exists(facade));
+        Assert.True(TypeForwardResolver.ForwardsType(facade, "System.Object"));
+
+        string targetPath = typeof(Console).Assembly.Location;
+        using var stream = File.OpenRead(targetPath);
+        using var peReader = new PEReader(stream);
+        var reader = peReader.GetMetadataReader();
+        using var builder = new LibraryBodyIndex.IndexBuilder(
+            targetPath, reader, peReader, new ConstantResolver(facade));
+
+        var objectReference = FindExternalTypeReference(reader, "System", "Object");
+        Assert.False(objectReference.IsNil, "System.Console should reference System.Object");
+
+        Assert.True(builder.TryResolveExternalTypeDefinition(objectReference) is null);
+    }
+
+    [Fact]
     public void CrossAssemblyMetadataResolver_NullResolverFailsHonest()
     {
         string targetPath = typeof(Console).Assembly.Location;
@@ -180,6 +261,40 @@ public class LibraryBodyIndexTests
         }
 
         throw new InvalidOperationException("Expected at least one external TypeRef.");
+    }
+
+    static TypeReferenceHandle FindExternalTypeReference(MetadataReader reader, string ns, string name)
+    {
+        foreach (var handle in reader.TypeReferences)
+        {
+            var typeReference = reader.GetTypeReference(handle);
+            if (typeReference.ResolutionScope.Kind != HandleKind.AssemblyReference)
+                continue;
+            if (reader.StringComparer.Equals(typeReference.Namespace, ns)
+                && reader.StringComparer.Equals(typeReference.Name, name))
+                return handle;
+        }
+
+        return default;
+    }
+
+    /// <summary>Resolves an identity to the same-named assembly in a directory, with no redirect.</summary>
+    sealed class FrameworkDirectoryResolver(string directory) : IAssemblyReferenceResolver
+    {
+        public ResolvedAssemblyReference? Resolve(AssemblyReferenceIdentity identity, AssemblyResolutionScope scope)
+        {
+            string candidate = Path.Combine(directory, identity.Name + ".dll");
+            return File.Exists(candidate)
+                ? new ResolvedAssemblyReference(identity, candidate, () => File.OpenRead(candidate), "test")
+                : null;
+        }
+    }
+
+    /// <summary>Answers every identity with one assembly, so any forwarder chain cycles.</summary>
+    sealed class ConstantResolver(string path) : IAssemblyReferenceResolver
+    {
+        public ResolvedAssemblyReference? Resolve(AssemblyReferenceIdentity identity, AssemblyResolutionScope scope)
+            => new(identity, path, () => File.OpenRead(path), "test");
     }
 
     sealed class CountingResolver : IAssemblyReferenceResolver
--- a/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
+++ b/src/ILInspector.Analysis.Tests/LibraryBodyIndexTests.cs
@@ -178,9 +178,6 @@ public class LibraryBodyIndexTests
             var definingReader = definition!.Value.DefiningReader;
             string definingName = definingReader.GetString(definingReader.GetAssemblyDefinition().Name);
             Assert.NotEqual(referencedName, definingName);
-            Assert.True(
-                TypeForwardResolver.DefinesType(Path.Combine(frameworkDir, definingName + ".dll"), fullTypeName),
-                $"{definingName} should define {fullTypeName}");
             Assert.Equal(name, definingReader.GetString(
                 definingReader.GetTypeDefinition(definition.Value.Definition).Name));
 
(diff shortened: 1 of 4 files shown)


richlander:

Two fixed-head reviews at `2667b895`, in isolated worktrees: **MAI-Code** and **Gemini 3.1 Pro**. Findings resolved in `98120385`.

### Gemini — High/High: forwarder hops could launder a platform assembly. **Accepted and fixed.**

`ScopeForReference` was computed once from the *originating* reference and carried across every hop. So an `Any`-scoped reference that forwards into a platform assembly resolved that platform assembly under `Any`.

I reproduced the mechanism rather than accepting it on reasoning, because the finding is only real if the scope is load-bearing. It is:

- `AssemblyDependencyResolver.Resolve` (`src/DotnetInspector.Services/AssemblyDependencyResolver.cs:152-166`) rejects, under `Platform`, any candidate whose provenance is not `TrustedPlatformAssembly` or `SharedFramework`. Under `Any` it calls `ResolveAll()`, which admits sibling and package candidates.
- This repository already models the adversary: `src/ILInspector.Analysis.SpoofRuntimeFixtures` is an unsigned assembly literally named `System.Runtime`.

Fixed with `IndexBuilder.NextHopScope`, which is **monotone — it only ever tightens**: `Any` becomes `Platform` when the forwarded identity carries a framework public-key token, and `Platform` is never downgraded. Recomputing the scope purely from the target would have been the obvious symmetric fix and is wrong — it would *loosen* a `Platform` chain that forwards to a non-framework identity. That direction fails closed instead.

Gated by four `NextHopScope` tests (tighten on three framework tokens; stay `Any` for unsigned and third-party tokens; never downgrade) plus an end-to-end test that records the scope of *every* hop of a real forwarder chain and asserts all of them are `Platform`.

### Gemini — attack point 8: layout-dependent assertion. **Accepted.**

`CrossAssemblyMetadataResolver_FollowsForwardersToDefiningAssembly` asserted the defining assembly is present in `frameworkDir`. Removed; the test now asserts only what it is about.

Its point that the cycle test is not a regression gate is correct and was already disclosed in the PR description — that test passes on `main` too. It documents the bound rather than gating it.

### MAI-Code — clean on all eight points, but surfaced the observation that mattered most. **Accepted.**

It reported that `TryResolveExternalTypeDefinition` has no product callers. I verified and expanded it: the only calls are its own recursion and four test sites. The resolver *is* fully plumbed — `ApiAnalysisInspection.CreateReferenceResolver` builds a real `AssemblyDependencyResolver` and threads it through `MethodBodyInspectionSession.Open` into `LibraryBodyIndex.Open`, including on the `Callers` path — but nothing consumes it. The suite had already pinned this: `Open_WithResolverDoesNotChangeIndexShape` asserts `ResolveCalls == 0`.

**This corrected the PR description, which was wrong.** It claimed the change "can surface analysis that was silently empty before." It cannot, today. The description now states the product blast radius is zero and this is enabling work — which also explains the zero-regression result rather than sitting awkwardly beside it.

### Gemini — Medium: delete the dead method family. **Declined.**

The observation is the same true one MAI-Code made; the recommendation is the opposite of what this track needs. #3419 is the consumer, and the wasm prototype already hand-rolls its own `FindForwardTarget` because the product does not offer this. Deleting it would deepen the gap it exists to close.

### Calibration note

MAI-Code graded scope carrying explicitly safe — "consistent with the platform-scope threat model and does not appear to weaken the resolver policy." That was the one thing in the diff that was not safe. Recording it because the miss is more useful than the clean verdict.

Round 2 at the fixed head `98120385` is running now; this PR is not ready until both come back clean.

richlander:
## Adversarial review round 2 — reconciliation (head `a1033ae7`)

Both reviewers re-ran against the fixed head `98120385`. **Both are now non-blocking.** One further finding was accepted and fixed in `a1033ae7`.

### Gemini — clean, but proved a test of mine was vacuous. **Accepted and fixed.**

Gemini confirmed the security fix is complete: the first hop is still derived from full metadata identity, `AssemblyReferenceIdentity` normalizes a full public key to a lowercase-hex token so `IsFrameworkToken`'s `OrdinalIgnoreCase` set cannot mismatch on format, and tightening does not break legitimate resolution because a framework-signed assembly does not forward to unsigned or third-party targets.

It then found the thing that mattered: **`CrossAssemblyMetadataResolver_ResolvesEveryForwarderHopUnderPlatformScope` does not gate tightening at all.** `System.Console`'s TypeRef to `System.Object` carries a framework token, so `ScopeForReference` returns `Platform` at hop 0 and the chain never starts at `Any`. Gemini proved it by reverting `NextHopScope` to `=> current` and observing the test still passed.

Following that thread found something worse, which is the real reason this is a third commit. The `NextHopScope_*` unit tests exercise the function in isolation, so **deleting the `scope = NextHopScope(scope, forwarded);` call from the loop left the entire suite green.** The function was gated; the wiring was not. AGENTS.md asks for a named non-vacuity test precisely for this case, and my doc comment's "Gated by `NextHopScope_*` tests" was overclaiming.

`a1033ae7` adds `ForwarderIntoFrameworkSignedAssemblyIsResolvedUnderPlatformScope`. It emits a synthetic three-assembly chain — `Contoso.App` → unsigned `Contoso.Facade` (`Any`) → framework-signed `Fake.CoreLib` via an ExportedType forwarder — because no natural artifact pairs an `Any`-scoped TypeRef with a framework-signed forwarder target. It asserts both the premise (the facade hop really is `Any`, so the test cannot pass for an implementation that scoped everything `Platform` from hop 0) and the property (the forwarded hop is `Platform`).

Verified non-vacuous by deleting the call site: it fails that test **and only that test**, `Expected: Platform / Actual: Any`.

Gemini's layout-dependence point on `Requests.Count > 1` is also accepted — whether `System.Object` takes a forwarder hop from `System.Console` is a framework-layout detail. Replaced with a non-emptiness assertion, and the older test's doc comment now states it gates only the non-downgrade half.

### MAI-Code — raised a High/High "residual bypass", then withdrew it. **Not a defect in this PR.**

MAI-Code argued that a forwarded reference carrying no public-key token leaves the hop at `Any`, so hostile metadata could still steer a hop to a sibling spoof. I did not accept the withdrawal on authority; I put three pieces of evidence to it and asked it to rebut or withdraw:

1. **Not a regression.** `ScopeForReference` is unchanged by this PR and already keys on the inspected assembly's own untrusted PKT. An attacker needs no forwarder at all — emitting `AssemblyRef System.Runtime` with no PKT reaches the same state at hop 0 on `main`. `NextHopScope` is monotone, so no hop is ever looser than hop 0.
2. **The unsigned case is not laundering.** It is the repository's deliberate policy that strong PKT identity, not simple name, decides framework-ness — which is exactly what `SpoofRuntimeFixtures` exists to enforce. An unsigned target resolves to an unsigned assembly, and `IsFrameworkReference`/`IsFrameworkDefinition` then correctly decline to treat it as framework. Nothing platform is laundered because nothing platform is claimed. Gemini's original finding was the narrower and real one: the forwarded ref *does* claim framework identity and was nonetheless looked up unconstrained.
3. **`MatchesIdentity`** (`AssemblyDependencyResolver.cs:203-232`) independently checks name, version, culture, and PKT-when-expected under both scopes.

Its suggested fix — tighten on the *resolved* assembly's real public key via `IsFrameworkDefinition` — is circular: you must resolve before you can inspect, and resolution is the thing being constrained. Under `Any` you would already have opened the spoof before you could judge it.

MAI-Code withdrew, agreeing on all three points and confirming it could not name a concrete type whose resolution becomes incorrect. Its verdict: non-blocking, pre-existing property of `ScopeForReference` rather than a regression here.

### Validation at `a1033ae7`

Analysis **584 / 0** (583 + the new gate). The eight failures in the other suites remain individually verified pre-existing on `origin/main`.

### Net

Round 1 caught a real security defect. Round 2 caught that the fix's wiring was untested and that one of my tests claimed more than it proved. Both are fixed; no reviewer has an open blocking finding.

richlander:
Ready to merge

CI green at `a1033ae7` including the Linux test job. Both required fixed-head reviews are non-blocking: Gemini clean after its vacuity finding was fixed, MAI-Code withdrawn. Reconciliation for both rounds is above.

Follow-up, non-blocking: #3419 is the consumer that will actually exercise this resolution path, and is unblocked by this PR.

richlander:
### Final head confirmation

Gemini re-reviewed `a1033ae7` (the test-only follow-up) and returned clean. It independently reproduced the gate both ways -- reverting \NextHopScope\ to \=> current;\ **and** deleting the call site each fail the new test -- and confirmed the premise assertion, the \	dForwarder\ encoding, and portability of the synthetic emission.

Every required fixed-head review is now clean at the exact pushed head `a1033ae7`. **Ready to merge** stands.

PR #3439, opened 2026-07-29
Define inspection layers, vocabulary, and seam rules

richlander:
## What

Adds `docs/design/inspection-layers.md`, a design note that defines how the
inspection stack splits into layers so more than one consumer can sit on it,
which layer owns which noun, and the seam rules that keep the split from
eroding. Registers it in `docs/overview.md` and the `AGENTS.md` task map.

Documentation only. No code, no behavior change.

## Why

The browser/WebAssembly prototype (`prototypes/inspect-web` on
`prototype/wasm-command-ui`) is the first consumer of the inspection stack
other than the CLI. Its engine can reference the `ILInspector.*` libraries but
not the CLI project, so it re-derived package acquisition, version resolution,
TFM ranking, symbol acquisition, XML-doc lookup, and call-graph orchestration
in a single ~2870-line file.

Several re-derivations diverged from the shared rules rather than merely
duplicating them — for example `.NETCoreApp,Version=v3.1` normalizing to
`net3.1` instead of `netcoreapp3.1`, version resolution without a semver
comparer, and symbol acquisition that never verifies the PDB debug GUID. A
second implementation of a shared rule is a second place for the rule to be
wrong, so the fix is a reachable seam rather than better discipline.

## What it settles

Four nouns on four axes, each chosen against existing usage rather than
invented:

| Noun | Meaning | Owner |
| --- | --- | --- |
| Scanner | a sequential pass over metadata | `ILInspector.Metadata` |
| Query | a typed inspection request producing a typed result | L1 |
| Section | the named, selectable unit (table, fields, list, blob, tree) | L2 |
| Shape | the narrowing rung (Document/Table/Vector/Scalar) | L2, Markout-defined |
| Format | presentation (markdown/json/tsv/jsonl/plaintext) | L3 |

Notable conclusions the note records:

- **"Section" is Markout's noun, not the CLI's.** `MarkoutSection` is the
  most-used Markout type in the repository, and `ISectionDescriptor.Name` is
  already defined as matching it.
- **"Shape" must not name a layer or project.** It already carries two
  established unrelated meanings: the ladder in `output-shapes.md` and the
  metadata concept in `TypeShape`/`TypeShapeKind`/`ArrayShape`.
- **A tree is a section.** `output-shapes.md` already settles that a section may
  be a table, field set, list, blob, or tree, so no broader noun is needed.
- **JSON is a format, not a shape.** `--json`/`--tsv`/`--jsonl` are presentation
  modifiers that do not change the shape.
- **Cost and capabilities belong to the query, not the section** — they are
  properties of acquisition, which is what lets a non-CLI consumer choose eager
  vs lazy.

## Boundary / non-goals

Explicitly does not change any user-visible command, flag, section name, or
category name; `-S`, `-D`, `@Category`, and the verbosity ladder keep their
current meanings. Does not add an output format or shape rung. Does not require
a consumer to adopt L2 — consuming L1 alone is supported. Does not alter
`ILInspector.*` ownership.

## Validation

Documentation-only change making no measured behavior claim, so Markdown
validation is the appropriate evidence per `AGENTS.md`:

```text
node markdownlint.js --config .markdownlint.yaml \
  docs/design/inspection-layers.md docs/overview.md AGENTS.md
EXIT=0
```

No product build or test run, per the documentation-only rule.

### DIFF
--- a/docs/design/inspection-layers.md
+++ b/docs/design/inspection-layers.md
@@ -85,7 +85,7 @@ categories; it does not compute facts and does not decide what a section costs.
 
 ## Vocabulary
 
-Four nouns, four axes. They are not synonyms and must not be used
+Five nouns, five axes. They are not synonyms and must not be used
 interchangeably.
 
 | Noun | Meaning | Values | Owner |
@@ -125,14 +125,27 @@ graph, and all of them are still "one section". Trees and tables are siblings
 `--jsonl` are presentation modifiers: they change how a selected payload is
 rendered without changing the shape.
 
-**Query** generalizes a pattern already present — `TypeQuery`,
-`OriginalTypeQuery`, `MetadataDeclarationQuery`, `MemberSourceQuery`,
-`SourceDocumentQuery`, `ILOffsetQuery`, `PlatformPrefixQuery` are all typed
-inspection requests that produce data. Two other senses exist on different axes
-and always keep their qualifier: **schema query** (`-D` catalog discovery, see
-[schema-query.md](schema-query.md)) and **row query** (field predicates within a
-section, see [row-query-order.md](row-query-order.md)). Unqualified "query" means
-the L1 inspection query.
+**Query** has a typed precedent and two senses to keep clear of. The typed
+precedent is small but exact:
+[`SourceDocumentQuery`](../../src/ILInspector.Metadata/MetadataFindings.Source.cs)
+and `MemberSourceQuery` are records in `ILInspector.Metadata` that carry a typed
+request a producer runs — precisely the L1 shape. `MetadataDeclarationQuery` is
+the same concept in utility form. The browser prototype independently named its
+entire exported surface `Query*`.
+
+Three other uses of the word are *not* precedent and must not be confused with
+the L1 noun:
+
+- **schema query** — `-D` catalog discovery, see [schema-query.md](schema-query.md). L2.
+- **row query** — field predicates within a section, see
+  [row-query-order.md](row-query-order.md). L2.
+- **a user's search string** — CLI option names such as `OriginalTypeQuery` and
+  `PlatformPrefixQuery` in `ApiOptions`, and the `ILOffsetQuery` helper. These
+  are inputs typed by a user, not typed requests. L3.
+
+Unqualified "query" means the L1 inspection query. The other three always keep
+their qualifier, and a new L1 query type is named for what it returns, never for
+the text a user typed.
 
 **Scanner** stays with the passes that genuinely scan —
 `MethodClassificationScanner`, `AssemblyDetailScanner`, `ResourceScanner`,
--- a/docs/design/inspection-layers.md
+++ b/docs/design/inspection-layers.md
@@ -1,0 +1,206 @@
+# Inspection layers and consumer boundaries
+
+How the inspection stack is split into layers so that more than one consumer can
+sit on it, which layer owns which noun, and the seam rules that keep the split
+from eroding. This is a design note about boundaries, ownership, and vocabulary —
+not a tour of every type.
+
+See [overview.md](../overview.md) for subsystem ownership,
+[section-model.md](section-model.md) for section selection semantics, and
+[output-shapes.md](output-shapes.md) for the shape ladder this note builds on.
+
+## Purpose
+
+`dotnet-inspect` grew as a single consumer, so inspection logic accumulated
+wherever it was first needed — which was usually the CLI project. That was cheap
+while the CLI was the only caller. It stopped being cheap when a second consumer
+appeared: a browser/WebAssembly app whose engine could reference the
+`ILInspector.*` libraries but not the CLI.
+
+The result was measurable. The prototype engine re-derived package acquisition,
+version resolution, TFM ranking, symbol acquisition, XML-doc lookup, and call
+graph orchestration from scratch, and the re-derivations were not merely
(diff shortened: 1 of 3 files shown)


richlander:
## Review reconciliation

Reviewed adversarially with **MAI-Code** in an isolated read-only worktree. Per `AGENTS.md`, this PR sits in the "more than trivial, but not high risk" tier — it is documentation-only, makes no measured behavior claim, and changes no product code — so one review applies.

### Pass 1 — base `1eea7b1e` → head `9f32eed1`

The reviewer verified 10 of 11 factual claims in the doc against the sources and returned two findings, both non-blocking:

**1. `Four nouns, four axes` contradicted the table below it (low).** The vocabulary table has five rows — scanner, query, section, shape, format. Straightforward miscount.

**2. The `Query` precedent was overstated (the substantive one).** The paragraph listed seven identifiers as evidence that "query" already means a typed inspection request in this repository. I verified this independently before acting on it, and the reviewer was right:

| Identifier | What it actually is |
| --- | --- |
| `SourceDocumentQuery` | record — `src/ILInspector.Metadata/MetadataFindings.Source.cs:6` |
| `MemberSourceQuery` | record — `src/ILInspector.Metadata/MetadataFindings.Source.cs:30` |
| `MetadataDeclarationQuery` | `static class` — `src/ILInspector.Metadata/MetadataDeclarationQuery.cs:62` |
| `OriginalTypeQuery` | `string` property on `ApiOptions` — `src/dotnet-inspect/Options/ApiOptions.cs:215` |
| `PlatformPrefixQuery` | `string` property on `ApiOptions` — `src/dotnet-inspect/Options/ApiOptions.cs:216` |
| `ILOffsetQuery` | CLI helper class — `src/dotnet-inspect/Commands/ILOffsetQuery.cs:10` |
| `TypeQuery` | local variable only |

Only the first two are typed query records. The bottom three are *a user's search string* — the opposite of a typed request.

That is worth getting right, because the doc's job is to justify the name `DotnetInspector.Queries` from evidence rather than assert it. A padded citation list would have made the name look better supported than it is, and the discarded entries turned out to carry the more useful signal: they are a **fourth sense of the word** the layering has to stay clear of.

### Resolution — `07959d98`

The rewritten paragraph states the typed precedent at its real size (`SourceDocumentQuery`, `MemberSourceQuery`, `MetadataDeclarationQuery` in utility form, plus the prototype's exported `Query*` surface), and promotes the demotion into content: the disambiguation list grew from two senses to three, with "a user's search string" assigned to L3 and named with its real citations. The closing rule now says an L1 query type is named for what it returns, never for the text a user typed. Noun count fixed to five.

### Pass 2 — fixed head `07959d98`

Re-reviewed at the exact fixed head. **No remaining findings.** The reviewer re-confirmed each new characterization against the sources, checked the new relative link resolves, counted the table rows against the corrected wording, checked the rewrite introduces no contradiction with the vocabulary table / seam rules / "What must change" sections, and re-ran markdownlint clean on all three changed files.

CI is green on `07959d98` (`markdownlint` pass; product jobs correctly path-skipped for a docs-only change).

Both review worktrees will be removed once this merges.

richlander:
Ready to merge

docs/design/inspection-layers.md
```diff
@@ -0,0 +1,219 @@
...
+Markout type in the repository, and `ISectionDescriptor.Name` is defined as
+matching the `MarkoutSection` name. Since L2 is the layer integrated with Markout
+serialization, it speaks Markout's vocabulary. The historical `Views/` directory
+is the misnomer, and it is what made the layer look ambiguous.
+
+**Shape** must not be reused as a layer or project name. It already carries two
+established, unrelated meanings: the output ladder defined in
+[output-shapes.md](output-shapes.md) ("Markout defines the shapes and produces
+them"), and the metadata domain concept in `TypeShape`/`TypeShapeKind`/
+`ArrayShape`. A third meaning would make the word useless.
```
richlander:
```suggestion
`ArrayShape`.
```

richlander:
## Follow-up: diagram order corrected — `3766fd32`

@richlander caught that the bottom box of the layer diagram listed the two base library families in the wrong order. Reversed in `3766fd32`.

The diagram descends, so of two stacked lines the upper reads as the higher layer. It had `ILInspector.*` above `DotnetInspector.*`; the real edge runs the other way. Verified against the project files:

- `DotnetInspector.Services` → `ILInspector.Metadata`, `ILInspector.Text` (plus `Core`, `Packages`)
- `DotnetInspector.Packages` → `DotnetInspector.Core`
- No `ILInspector.*` project references `DotnetInspector.*` at all

`ILInspector.*` is the floor.

This is worth a commit rather than a silent fix, because **fault 3** of the analysis behind this note is an instance of precisely this inversion: `ILInspector.Metadata` carrying a `PackageReference` on `SourceLinkFetch`, i.e. network fetch underneath the lowest metadata layer. A diagram that draws the two families in the wrong order makes that class of mistake look like the intended shape — and this note is meant to be the thing people check the shape against.

**Review tier: trivial — no review.** Two lines swapped inside a fenced `text` block, no prose changed, claim verified directly against `.csproj` references. The comment column still aligns. Prose at L1 (line 54) enumerates both families as a set rather than an ordering, so it needed no change.

`markdownlint` and `changes` pass on `3766fd32`; product jobs path-skipped as before. Still **ready to merge**.

Issue #3408, opened 2026-07-28
SourceLink provenance is extracted with unanchored regexes, allowing a hostile PDB to spoof a repository


richlander:
Split out of the review of #3402 as explicitly **pre-existing and out of scope** there. Found by GPT-5.6 while reviewing that PR; independently reproduced.

## The problem

`src/SourceLinkFetch/SourceLinkResolver.cs` derives *provenance* — owner, repo, commit, and browse URL — by running unanchored regexes over the SourceLink URL template taken from the PDB:

- `:79` `https://raw\.githubusercontent\.com/([^/]+)/([^/]+)/`
- `:94` `/([0-9a-f]{40})(?:/|$)`
- `:108` `https://raw\.githubusercontent\.com/([^/]+)/([^/]+)/([^/]+)/(.+)`

Because they are unanchored, they match a GitHub URL appearing **anywhere** in the string, including inside the query string of a host the attacker controls. The PDB is untrusted input (`docs/design/untrusted-data-threat-model.md:36` names spoofed provenance as an in-scope risk).

## Reproduction

A hostile document key/value pair:

```
"/_/*" -> "https://attacker.test/serve?ref=https://raw.githubusercontent.com/dotnet/runtime/1111111111111111111111111111111111111111/*"
```

yields:

```
fetched from : https://attacker.test/serve?ref=...      <-- attacker's host
repository   : https://github.com/dotnet/runtime        <-- spoofed
commit       : 1111111111111111111111111111111111111111 <-- spoofed
browse URL   : https://github.com/dotnet/runtime/raw/1111.../src/Evil.cs
```

So the content is fetched from the attacker while the tool attributes it to `dotnet/runtime`.

## Impact and non-impact

- Impact: displayed provenance is attacker-chosen and disagrees with the actual fetch origin.
- **Not** cache poisoning: cache keys incorporate the exact fetched URL, so the entry cannot collide with the real repository's. Verified during review.

## Suggested direction

Parse with `Uri` and require scheme `https` plus `Host == "raw.githubusercontent.com"` **exactly** before extracting provenance; do not recognize GitHub by substring. The repo already has the safer habit elsewhere — `SourceDocumentPath` does structural parsing rather than substring matching.

[...]

PR #3440, opened 2026-07-29
Move the C# body slicer out of ILInspector.Metadata

richlander:
`ILInspector.Metadata` should expose PDB resources and nothing else. `SourceLinkResolver.cs` carried **~1,800 lines of C# text processing** in the same file as **~420 lines** of genuine SRM/PDB work. This moves the text half into a new `DotnetInspector.CSharpBodySlicer` library.

## The two halves shared a file and nothing else

| | |
| --- | --- |
| Text half (lines 80–1906) → SRM types | **0 references** |
| SRM half → text-half helpers | **0 calls** |
| Anything in Metadata → `ExtractMethodBody` | **none** |
| Actual consumers | `DotnetInspector.Services`, `dotnet-inspect` — both **above** Metadata |

The slicer was *parked* in Metadata, not coupled to it. Its seam was already a pure function — `ExtractMethodBody(string sourceText, int startLine, int endLine, string methodName, ...)`, text in and text out — so the move is mechanical rather than a refactor.

## Evidence that nothing changed

```
moved region vs d67bfa42:80-1906   IDENTICAL (1,827 lines)
surviving file vs d67bfa42 minus that region   IDENTICAL
```

The only edits are the enclosing type name, the two consumer call sites, and two `using`s the move left dead (`System.Text.Json`, `System.Text.RegularExpressions`; `Ecma335` stays — `MetadataTokens` is still used).

## Naming

The library is named for what it delivers and not more. It isolates one member's text given the line range a PDB reports. It is **not** a C# parser, has no syntax tree, and is heuristic at the edges — its seven pinned `_KnownGap` tests say so. A name like `CSharpSource` or `CSharpReader` would have promised a great deal more than it delivers. The internal scanner added in the follow-up will be named on the same principle.

## Corpus reduction — measured, not incidental

`AuthoredSourceValidityTests.SliceCorpus()` scans **every assembly in its own output directory**, so moving the tests changed the corpus:

```
baseline (main, ILInspector.Metadata.Tests)   4,924 slices
this branch (CSharpBodySlicer.Tests)          3,757 slices   -24%
```

Cause, from diffing the two assembly sets: the corpus includes **the test assembly itself**, and `ILInspector.Metadata.Tests.dll` (989 tests) is much larger than `DotnetInspector.CSharpBodySlicer.Tests.dll` (244). Four small metadata-only fixtures also dropped out. `DiffAsmFixtures.Target` and `ILInspector.Decompiler.Fixtures.NewUnsafe` are referenced explicitly because the corpus genuinely needs them — without the former, `LastMemberOfAType_ExtractsThroughItsOwnClosingBraceOnly` fails.

The enforced floor is 1,000 slices (cleared 3.7×) and every rate ceiling still passes. The reduction is nonetheless real, and it exposes that this corpus is defined by incidental output-directory contents. I did not redesign that here — it is not a move-PR concern — but it is worth its own change.

## Validation

```
dotnet build dotnet-inspect.slnx -c Release   0 errors

DotnetInspector.CSharpBodySlicer.Tests    244 passed, 0 failed
ILInspector.Metadata.Tests                989 passed, 0 failed   (1,233 - 244 moved)
DotnetInspector.Services.Tests            319 passed, 0 failed
dotnet-inspect.Tests                    2,353 passed, 0 failed
```

## What this does not do

Metadata still references `SourceLinkFetch` and still carries SourceLink URL mapping and `GitHubBrowseUrl` — that is the next change, and it is a real refactor (public record shapes, 59 call sites across 14 files) rather than a move. It also collapses the two divergent SourceLink resolvers behind #3408.


richlander:

Two reviews at fixed head `8ec3ad8c`, base `d67bfa42`, each in an isolated review worktree.

| Reviewer | Verdict | Findings |
| --- | --- | --- |
| MAI-Code | Approve | none |
| GPT | Approve | none |

Both confirmed `git rev-parse HEAD` == `8ec3ad8c` before reviewing.

### What both reviewers independently verified

- **Byte-identity of the move.** Both reconstructed the moved region from `d67bfa42:src/ILInspector.Metadata/SourceLinkResolver.cs` lines 80-1906 and diffed it against `BodySlicer.cs`. Identical apart from the namespace/class wrapper. Both also confirmed the surviving Metadata file is that same file with exactly that region deleted.
- **Nothing dropped or duplicated.** `LexState`, `IsDestructorSignatureLine`, `SimpleTypeName`, the private helpers, and `ExtractMethodBody` each exist exactly once in the new library and no longer appear in Metadata.
- **Seam is behaviorally neutral.** No static state introduced; call sites route through the same logic; `InternalsVisibleTo` wiring is sufficient for the moved tests.
- **Packaging.** `dotnet pack` on the new project produced no `.nupkg`, so it correctly inherits `IsPackable=false` from the root `Directory.Build.props`.
- **Layering.** New library has zero project references; Metadata does not reference it.

### Corpus reduction — the item I flagged as highest risk

I called this out to both reviewers explicitly rather than letting it pass on a green suite, because a shrinking corpus is exactly how a gate stops gating without anything turning red.

Both reproduced the measurement independently: **4,924 slices at `d67bfa42` → 3,757 on this branch (-24%)**, and both confirmed the attribution — `SliceCorpus()` enumerates every `*.dll` in `AppContext.BaseDirectory`, so it counts the test assembly itself, which changed from `ILInspector.Metadata.Tests.dll` (989 tests) to `DotnetInspector.CSharpBodySlicer.Tests.dll` (244), plus four fixture assemblies no longer landing in the output directory.

Both concluded **no assertion became vacuous or trivially satisfied**; the corpus still clears the `>=1000` / `>=10` / `>=10` breadth thresholds, and the rate ceilings remain meaningful.

I'm satisfied with that, but noting for the record: the corpus being an implicit function of the test project's reference closure is a latent trap independent of this PR. It is worth making explicit later; it is not a reason to hold this move.

### Validation at the reviewed head

Both reviewers ran the full set in Release and got identical numbers to mine:

| Suite | Result |
| --- | --- |
| `DotnetInspector.CSharpBodySlicer.Tests` | 244, 0 failed |
| `ILInspector.Metadata.Tests` | 989, 0 failed |
| `DotnetInspector.Services.Tests` | 319, 0 failed |
| `dotnet-inspect.Tests` | 2353, 0 failed |

`1233 - 244 = 989` accounts exactly for the moved tests.

### Non-actions, deliberate

`DocCommentParser.cs` and the SourceLink URL / `GitHubBrowseUrl` handling remain in `ILInspector.Metadata`. Both are out of scope here and deferred to follow-up PRs; the SourceLink eviction is the real refactor (public record shapes, 59 call sites across 14 files) and also collapses the two divergent resolvers tracked by #3408.

Ready to merge

PR #3448, opened 2026-07-29
Package section model: curated SizeClass + Cost catalog, document-kind file family, reachable nuspec

richlander:
Brings the `package` command to the same section standard as `library`, and fixes
the bug that made a NuGet package's most important document unreachable.

## What changes

**The `.nuspec` is reachable.** `PackageFileLister.IsPlumbing` classified it
alongside `_rels/`, `[Content_Types].xml`, and `.psmdcp`. It is authored content,
not OPC plumbing. Before this change every route to it returned success-shaped
empty output — even an exact-path request:

```console
$ dotnet-inspect package SkiaSharp --path SkiaSharp.nuspec
Note: 1 matched section has no data: Files.
$ echo $?
0
```

A user could not distinguish "this package has no manifest" (impossible) from
"the tool hides it".

**A file family named for document kinds, behind `@Files`.** `Package README file`,
`Package nuspec file`, and `Package skill files` — three disjoint kinds, so
`@Files` never renders the same path twice. The singular/plural form is
load-bearing: a singular name yields at most one row. The package command had
zero categories before this; `@Files` and `@SourceLink` are its first doors.
`Package files` remains the unfiltered superset and is deliberately not a family
member.

**Layout roots are not sections.** An earlier revision of this branch added
`Package library files`, `Package reference files`, and `Package runtime files`;
this PR removes them along with every alias reaching them. `lib/`, `ref/`, and
`runtimes/` are slices of one listing rather than different kinds of thing —
`--path "lib/**"` scopes `Package files`, and `Package Info`'s `Content` field
already names the roots a package ships.

**Curated catalog.** Every package section now declares a `SizeClass` (how its
row count grows across the universe of packages) and a `Cost` (what producing it
costs), and the pipeline runs with `UseCuratedCatalog`. Package descriptors had
never adopted these axes; they were still on the legacy `Info`/`ExplicitOnly`
booleans.

**`@All` and `@Default` are gone from `package`.** Not merely unlisted —
unresolvable (`WithoutComputedPoles`). Package sections are reachable by name, by
topical door, and by verbosity; a pole that renders a superset nobody asked for
is a surface no discovery output describes. `@Default` survives only as the
internal encoding of bare `-S`, short-circuited before selector resolution.
`library`, `api`, and `type` are untouched and keep both.

**`Source Files` → `SourceLink: Files`.** Confirmed to be genuinely SourceLink
data — it calls the same `SourceFileCollector.CollectFromAssemblyAsync` the
library command's `SourceLink: Files` calls.

## The behavioral claim

Bare `-S` is now the network-free **fixed** overview: the row set does not grow
with the package, so the overview cannot become long. Applicability still applies
on top, so a section's absence means "not applicable" and never "too long for
this package" — `NETStandard.Library`, which ships no README, renders four
sections rather than five.

```console
$ for p in Markout Microsoft.Data.SqlClient Microsoft.Extensions.Hosting SkiaSharp Newtonsoft.Json; do
    dotnet-inspect package $p -S | grep '^## '
  done
```

All five render `Manifest`, `Package Info`, `Package nuspec file`,
`Package README file`, `Signature`. These packages differ by an order of
magnitude in file count (Markout 22, SqlClient 62) and by structure (SkiaSharp is
`ref/`-heavy, SqlClient is the only one with `runtimes/`), which is what makes
the constant result evidence rather than coincidence.

## Two pipeline fixes this port surfaced

**`package -v:q` had silently lost its identity line.** The curated ladder
returned `false` for everything at Quiet, which cost the headless `Summary`
preamble — the `Version: … | Type: …` line. Library never caught this because it
has no `Summary` section at all; it renders identity from the view model.
`Summary` is a rendering convention, not a rung on the size/cost ladder, and
`IsHeadlessSummary` now says so.

**`Cost=Unbounded` can no longer join `@All`.** `IsAllMember` is computed from
`IsExpensive`/`ExplicitOnly` while auto-rendering is computed from `Cost`, so the
two axes could drift and admit an unbounded section into a pole that renders
every member. Registration now rejects `Cost=Unbounded` unless the section is
also `IsExpensive` or `ExplicitOnly`, making that state unrepresentable.
`SectionPipeline_Add_UnboundedCostSectionThatWouldJoinAll_Throws` is the
non-vacuity gate; disabling the check turns it red.

`Noisy` correspondingly now means "listed in the discovery catalog but not an
`@All` member". It has exactly one user (`Package files`), so the semantic change
reaches nothing else.

## For the reviewer

**Section render order is declaration order, not sort order.**
`Package_DetailedOutput_RendersSectionsAlphabetically` pins alphabetical output,
so `InspectionResultView`'s properties must be hand-ordered by section name — a
trap for the next person who adds one.

**`Moderated` is a claim about latency**, and I measured it rather than assuming:
`Signals` ~1.0s and `Statistics` ~0.25s warm on both Markout and
System.Text.Json.

## Compatibility

`Package library/reference/runtime files` and `Package markdown files` are
removed with no replacement spelling — `--path` is the replacement mechanism
(`--path "lib/**"`, `--path "*.md"`). `@All`/`@Default` no longer resolve on
`package`. Every surviving rename keeps its old spelling in
`SelectResolver.LegacySectionAliases`, including `Grounding`, which was this
section's canonical name on main rather than a nickname.

AGENTS.md is dropped from the README selection chain (now README.md >
PACKAGE.md > declared), scoped to README selection only: `--path @agents`,
`IsAgents`, `HasAgentDocumentation`, and `project --agents` are untouched.

## Validation

- `dotnet run --project src/dotnet-inspect.Tests -c Release` — **2370 passed**,
  0 failed, 14 skipped (ilasm/ildasm absent).
- `npx markdownlint-cli` clean over all changed docs.
- Two-model adversarial review; round-1 findings reconciled in a PR comment.

## Follow-ups, not blocking

- The family is still incomplete: `analyzers/`, `build/`, `tools/`, and
  `contentFiles/` have no document-kind section.
  `Microsoft.CodeAnalysis.NetAnalyzers` ships no `lib/` at all, so its entire
  payload reaches the user only through `Package files`.
- `type` still spells its section `Source Files`, now the odd one out.
- Porting `SourceLink: Availability`, `Missing Files`, and `Integrity` to
  `package`.
- `api` and `type` remain on the legacy catalog with visible `@All`/`@Default`.


richlander:
## Round-1 adversarial review reconciled

Two independent reviews of `03bcd344` — **GPT-5.6 Sol** and **Gemini 3.1 Pro**,
each in an isolated worktree at the exact head. Six findings between them; all
six reproduced, all six addressed in `41980415`.

### Actioned

**`@All` rendered a superset alongside its subsets** (both reviewers, blocking).
`Package files` contains every row the family sections contain, so `-S @All`
emitted `README.md` and every `SKILL.md` twice. My PR text asserted `Noisy` kept
it *out* of `@All`; the code said otherwise —
`IsAllMember = !IsExpensive && (!ExplicitOnly || Noisy)` pulled it *in*.

**`@All` did unbounded work it never rendered** (GPT, blocking).
`ShouldPopulatePackageSourceFiles` treated `@All` as a request for
`SourceLink: Files`, acquiring PDBs over the network to populate rows no view
emitted. Pre-existing (from #1169), but tightly coupled to the `@All` semantics
this PR changes.

Rather than patch these individually, **`package` now drops the computed `@All`
and `@Default` poles entirely** (`WithoutComputedPoles`). They are unresolvable
there, not merely unlisted. Package sections are reachable by name, by topical
door, and by verbosity; a pole that renders "everything" is a surface no
discovery output describes. `@Default` survives only as the internal encoding of
bare `-S`, short-circuited before selector resolution — bare `-S` is unaffected.
`library`, `api`, and `type` keep both poles.

**`Grounding` was a compat break** (Gemini, blocking). I mischaracterized it in
the PR description as an alias. It was the *canonical* name on main —
`PackageSections.PackageReadme = "Grounding"` — so `-S Grounding` worked and now
errored. It also contradicted this PR's own rule that every rename keeps its old
spelling. Restored in `LegacySectionAliases`; the term stays gone from the
rendered surface.

**`Cost`/`IsExpensive` had no gate** (both reviewers). Now a registration-time
check rejects `Cost=Unbounded` unless the section is also `IsExpensive` or
`ExplicitOnly`, making the drift unrepresentable rather than comment-managed.
`SectionPipeline_Add_UnboundedCostSectionThatWouldJoinAll_Throws` is the
non-vacuity gate — disabling the check turns it red, verified by tampering.

**`Package markdown files` overlapped its siblings** (GPT). It matched every
`.md`, so it was a superset of both `Package README file` and
`Package skill files`. Removed. The family is now four disjoint document kinds
and `@Files` cannot render a path twice:

```console
$ dotnet-inspect package Markout -S @Files --count
| Section              | Count |
| Package nuspec file  | 1     |
| Package README file  | 1     |
| Package skill files  | 5     |
```

Markdown at large is still reachable via `--path "*.md"`.

**Bare `-S` is 4 sections, not 5, for a package with no README** (GPT blocking,
Gemini verified-as-correct). Gemini is right: `NETStandard.Library` genuinely has
no README, and a 0-row section dropping is the intended rule. The defect was in
my wording, not the behavior. `SectionSizeClass.Fixed`'s doc no longer claims
these sections "render the same set for every target" — applicability applies on
top, so absence means "not applicable", never "too long for this package". PR
description corrected to match.

### Corrected claim

My PR text said unifying `IsExpensive` with `Cost` "would change library
behavior" because `UnsafeMembers` and `TopLeverage` are `IsExpensive=false` with
`Cost=Unbounded`. **That was wrong** — both are `ExplicitOnly=true`, so they were
never `@All` members under either rule. The reviewers were right to press on it,
and the registration gate now enforces what I had only asserted.

### Validation

Full suite green at **2370** on `41980415`. `markdownlint` clean.

Round-2 review of the fixed head is in flight with both models. Not ready to
merge until both come back clean.

richlander:
## Merged `origin/main` (#3405 landed) — resolved, with one real behavior interaction

#3405 added the lens/discovery projection dispatch and touches the same three sites in `PackageCommand` this branch changes. All three conflicts keep **both** sides.

### 1 & 3 — the two `-D` call sites

Now pass this branch's `catalogHiddenSections` and `listedCategoryDoors` **and** main's `projection: options`.

This is the one that needed care. `DiscoverOutput` builds its row list **twice** — once for the projection, once for the render — so a filter reaching only one of them makes `--count` answer a listing the command never prints, at exit 0 with the projection audit satisfied. That is precisely how the #3229 merge broke `library -D --count` (15 for a 13-row listing). I verified the auto-merged `DiscoverOutput` passes the full argument set on both paths, then checked it end to end:

| | count | rendered |
| --- | --- | --- |
| `package -D` | 12 | 12 |
| `package --schema -D` | 17 | 17 |

`package -D` moved 14 → **12** under this branch's curated catalog while `--schema -D` shows the full **17**. Both agree with their own render. Had the projection path missed `catalogHiddenSections`, the count would have read 17 against a 12-row listing — a plausible number for the wrong question, and invisible to the audit.

### 2 — the `--count` gate

Keeps main's `!rendersOwnPayload` shield (exempting the lenses and `-D`, which render their own payload) **and** this branch's `ValidateSectionsSelected` (allowing a count over several selected sections). That combination is exactly what `LibraryCommand` already does on main, so the two commands stay aligned rather than diverging through a conflict resolution.

Verified: the lens `-S` refusal, `--tfms --count`, `-D --count`, and `-D --print` refusal all still behave as #3405 intends, and `-S "Package Info" -S Dependencies --count` returns a per-section count map as this branch intends.

### One real failure the merge exposed

`Layout_Count_CountsFilesRatherThanRenderedTreeLines` expected 19 and got 20. That is **this branch working as designed**: it makes the nuspec a reachable package file, so `Newtonsoft.Json@13.0.4` now lays out 20 files, and the tree renders exactly those 20. The count followed the render; the literal was stale.

The literal was also the *only* thing that test asserted, which is why an intended change read as a regression. It now also pins that the tree renders more lines than the count — the property the test is named for — and that the nuspec is present, so a tree that grows directory nodes cannot drift into agreeing with the count by coincidence.

### Validation

- CLI **2402 pass / 0 fail / 14 skipped**; services **343 / 0**
- Count-vs-payload sweep all matching: `package -D` 12, `package --schema -D` 17, `library -D` 15, `library --schema -D` 58, `type -D` 5, `--versions` 53, `--versions --include-unlisted` 53, `--tfms` 8, `--il-offsets` 3, `--il-offsets --rows 2` 2

The merge commit `fa39416d` is the conflict resolution only; the test change is `23556f69`, so the behavioral change is reviewable on its own rather than buried in a merge diff.

richlander:
## Round 2: `--print` restructured, `--readme` removed

**GPT-5.6 Sol** found a blocking regression; **Gemini 3.1 Pro** came back clean on the merge integration and file-family consistency (independent audit of both `GetDiscoveryRows` call sites, `ExecuteEffective`'s 14-argument delegation, and `.nuspec` reachability across `--layout` / `-S Files` / `--paths` / `--content` / `--count`).

### The finding

`package Polly@7.2.4 -S 'Package nuspec file' --print` emitted XML that did not match the manifest nuget.org serves — a URL inside `<releaseNotes>` rewritten from `github.com/.../blob/...` to `raw.githubusercontent.com/...`. Exit 0, no indication. Reproduced, and confirmed to be this PR's regression: `PrintNuspec` exists on neither `main` nor the merge base.

### Why it happened

`--print` was implemented *inside* the `--readme` early exit:

```csharp
if (options.ShowReadme || (options.Print && !effectiveDiscovery))
{
    if (options.Print && options.IncludeSections?.Contains(PackageSections.FilesNuspec) == true)
        return PrintNuspec(...);          // added by this PR
    return PrintReadme(...);
}
```

Structurally, "print" on this command meant "print the README". A second printable section could not be expressed, so it had to intercept that branch and hand off to a copy of `PrintReadme`'s print slice — which inherited the README *content reader*, whose parameter is literally named `readmeFile` and which applies frontmatter scoping and blob→raw rewriting to whatever it read. The rewriter is a bare regex over the whole string, not a Markdown-link parser, so a URL inside an XML element matches.

So the corruption was a symptom. `output-shapes.md` specifies `--print` as "a selected row's **declared** printable payload" with "printability is a row capability" — hardcoding the payload kind into the print path inverts that, and makes every new printable section need a printer that re-derives a document from the command line.

### What changed

`--print` now joins `--value`/`--urls`/`--paths`/`--count` after inspection and reads the **same `InspectionResultView` rows the section renders**, so it cannot answer from a different list than the one displayed — the invariant this whole series has been about. Cardinality, `--row`, and output shape come from `PrintProjectionOutput`. `PrintNuspec` and `PrintReadme` are deleted.

Skill documents became printable by declaring rows and nothing else, which is the proof the shape is generic rather than a third special case:

```console
$ ... -S "Package skill files" --print
Error: selected section has 2 printable rows; use --row N|first|last to choose one row.
$ ... -S "Package skill files" --print --row last --bare
# Beta skill
```

**`--readme` is removed.** A flag that renders one document is a second answer to *which document*, competing with the section selection — and that disagreement is exactly how the manifest reached the README's pipeline. The replacements already existed: `-S "Package README file" --print` (with `--bare`/`--json`/`--jsonl`/`--frontmatter`/`--body`), and `--content --path @readme --frontmatter` for the multi-package survey, which the existing `--content` error text already recommended. The `InfoTracker` "readme" provenance detail — written only by the deleted printer — is replaced by the path the print payload already carries.

Markdown treatment is now applied only to Markdown documents, and a Markdown scope on a non-Markdown document is refused rather than answered with the whole document or an empty one.

### Evidence

| | before | after |
| --- | --- | --- |
| `-S 'Package nuspec file' --print` vs `api.nuget.org` | 1 line differs | **byte-identical** |
| `.props` comment URL via `--content` | rewritten | preserved |
| README via `--print` | blob→raw | blob→raw (unchanged) |
| `--print --frontmatter` on the nuspec | exit 0, empty | refused with reason |

New tests: manifest pinned byte-for-byte against the `.nuspec` entry in the package; `.props` verbatim *while the README in the same package still gets rewritten* (so it is a rule about kind, not the rewriter being off); scope refusal; skills cardinality + `--row` agreeing with the section listing.

CLI **2406 / 0**, services **343 / 0**, count-vs-payload sweep clean across 9 pairs including `-S "Package files"` 20, `-S "Package README file"` 1, `-S 'Package nuspec file'` 1.

Two commits so the restructure is reviewable on its own: `9c1f5f11` (the Markdown predicate) and `a5773dea` (the restructure). Dispatching fixed-head reviews at `a5773dea`.

richlander:
## Round 2 reconciliation — five findings actioned, head now `5daa885d`

Both reviewers ran against `d26a7556`. **GPT-5.6 Sol** returned three blocking findings, **Gemini 3.1 Pro** two blocking plus one speculative. They overlapped on exactly one (migration guidance), which is the argument for the two-model rule: each found things the other missed.

### Blocking — fixed

**1. `--content` accepted a Markdown scope and then ignored it (GPT).** Gating Markdown treatment on document kind fixed the corruption but left the scope *accepted*, so `--content --path build/X.props --frontmatter` returned the **whole document** and exited 0. That is this series' defect class reintroduced through the door I had just built. `--print` already refused it; `--content` now refuses on the same terms, at the single choke point both the single- and multi-package paths funnel through.

**2. Printing dropped a byte order mark (GPT).** `ReadAllText` consumes the preamble, so a manifest shipping one was emitted three bytes shorter than it exists in the package. EntityFramework 6.4.4 ships one.

Worse, my `EmitsTheManifestExactlyAsShipped` test compared through a `StreamReader`, which strips the mark from the *expectation* too — it would have agreed that a three-bytes-shorter document was identical. It compares bytes now, and a dedicated test uses a package that actually ships a mark, since the existing fixture writes BOM-less files and structurally could not have caught this.

```console
$ dotnet-inspect package EntityFramework@6.4.4 -S 'Package nuspec file' --print | cmp - EntityFramework.nuspec
$ echo $?
0     # 1948 bytes, sha256 5f6a6bc7…, both sides
```

**3. `--print` read every document before resolving cardinality (GPT).** An ambiguous request paid for all five Markout skill payloads and *then* refused; `--row 2` read all five to emit one. `output-shapes.md` specifies that one `--print` authorizes exactly one payload fetch. Only one document is ever emitted — the JSON shapes serialize the selected row too — so selection now resolves from row identity, which is metadata, and content is acquired for the emitted row alone. `PrintProjectionOutput` gained a lazy overload; existing document-list callers are unchanged. Pinned by counting resolver invocations for both the refusal (0 reads) and `--row` (1 read, the right one).

**4. Extensionless READMEs lost Markdown treatment (Gemini).** `ResolvePackageReadme` falls back to whatever the manifest declares, so a package can ship `<readme>README</readme>`. Keying Markdown treatment on extension refused `--frontmatter` on it and dropped blob→raw rewriting — both worked on base. The rule is now **extension decides for ordinary documents, role decides for the readme**. The nuspec in the same package still prints verbatim, so this is role rather than a relaxation of the rule that fixed the manifest in the first place.

**5. `--readme` gave no migration guidance (both reviewers).** Now answered with its replacement rather than `Unrecognized option`, matching how this repo already answers the stale `--head N` spelling. It retains no capability — only guidance — and covers all three forms that used to accept it, including the bare name that preprocesses to `router` rather than `package`. `project --readme` is a different option and is untouched.

### Not reproduced — but the code changed anyway

Gemini also reported that verbosity gating could make the empty-section refusal fire on a *populated* section. **This is not reachable**: the file listing is populated at `-v:q/m/n/d` alike, which GPT verified independently and I confirmed directly. Gemini gave no reproduction, as the prompt required.

The underlying code smell was real, though: `FamilyRows` returns `null` when the listing was never collected and I was collapsing that to `[]`, so "we never looked" and "there is nothing there" produced the same message. Those are now distinct, and an unprintable section is an explicit contradiction rather than an empty answer.

### Also fixed, found by me rather than review

- At `-T:d` the CLI still recommended `package <id> --readme`; following the tool's own suggestion produced `Unrecognized option`. Pinned by a test that **parses the emitted tip and executes it**, verified to fail both when the tip names a removed flag and when it names an unresolvable section.
- `LensProjection.TryProject`'s `scalarPayload` and `printHandledByLens` parameters had zero callers and zero coverage once `--readme` was gone — deleted.
- The empty-payload refusal had lost the specificity the bespoke printer had; it names the section again.

### Validation

CLI **2413 / 0** (14 skipped), services **343 / 0**, markdownlint clean. Mutation-checked: reverting the preamble fix fails the BOM test; reverting the tip fails the tip test.

Dispatching fixed-head reviews at `5daa885d`.

richlander:
## Round 3 reconciliation (reviewed at `5daa885d`, fixed at `cada7b6f`)

Two models re-reviewed the fixed head: **GPT-5.6 Sol** and **Gemini 3.1 Pro**, each in an isolated worktree pinned at `5daa885d`, each told which of its own earlier findings had been accepted.

Both confirmed the round-2 fixes are genuinely fixed: the `--content` scope refusal covers single-package, multi-package, mixed-match, JSONL and `--out` routes; UTF-8 BOMs survive exactly once; `--print` fetches exactly one document on success and none on a refusal; the extensionless README keeps its Markdown role.

They then found four more things. Three were real and are fixed; one is not a defect and one is intended.

### Fixed: the readme role overrode the extension (Gemini, high confidence)

A manifest can declare anything as its readme. `<readme>images/logo.png</readme>` is malformed but shippable, and the role rule made that declaration force Markdown handling onto the PNG. Reproduced:

```console
$ dotnet-inspect package Test.BinReadme.1.0.0.nupkg --content --path images/logo.png --bare | xxd
00000000: efbf bd50 4e47 0d0a 1a0a 7365 6520 6874  ...PNG....see ht
00000010: 7470 733a 2f2f 7261 772e 6769 7468 7562  tps://raw.github
00000020: 7573 6572 636f 6e74 656e 742e 636f 6d2f  usercontent.com/
```

The link was rewritten. This is the exact defect the PR opens by fixing, reintroduced through the door the role rule opened. The role now answers only where the extension is silent:

```csharp
=> MarkdownContent.IsMarkdown(path) || (isReadme && !Path.HasExtension(path));
```

An extensionless declared README is still Markdown; a declared `.png`, `.txt` or `.pdf` is not. Test `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown` asserts the bytes are unchanged and that a scope over it is refused; mutation-tested against the old rule.

(The leading `efbfbd` is `0x89` decoded as UTF-8. That is pre-existing on base -- binary content through a text pipeline is lossy either way -- and out of scope here. What this PR owned was the rewriting, and that is gone.)

### Fixed: the readme role was lost when the conventional name also existed (GPT, blocking)

`ResolvePackageReadme` prefers `README.md`, so a package shipping one *and* declaring `docs/GUIDE` left the declared file unflagged, and a scope over it was refused as non-Markdown. Which file the README section displays is a presentation choice; the manifest declaration is what makes a document a readme. Content reading now restores the role to the declared path. Test `Package_DeclaredReadme_KeepsItsRoleWhenTheConventionalNameAlsoExists`, mutation-tested.

### Fixed: the removed-flag diagnosis could not tell a flag from a value (both models)

The migration error scanned raw tokens, so `--out --readme` -- where the token is an output file name -- was refused, and a bare name routing to a platform assembly got package-specific advice for an option that route never had.

The token scan is gone. The package command already classifies unrecognized options, so the replacement is named there. The result is that a run which parses is never second-guessed:

| command | before | after |
| --- | --- | --- |
| `package X --readme` | migration error | migration error |
| `X --readme` (bare) | migration error | migration error |
| `package X --readme=true` | `Unrecognized option` | migration error |
| `package X --out --readme` | migration error | writes the file, exit 0 |
| `System.Private.CoreLib --readme` | migration error | the router's own answer |
| `package X --nosuch` | `Unrecognized option` | unchanged |

`--readme=true` is now covered for free, because the parser reports the assigned spelling too. Test `Package_ReadmeInAValuePosition_IsNotMistakenForTheRemovedFlag`.

### Not a defect: the BOM in JSON output (Gemini, high confidence -- dismissed)

Gemini read `"content":"\uFEFFHello"` as a stray non-character corrupting the payload. It is the escape that makes the payload faithful. Verified at the byte level:

```console
$ ... --content --path with-bom.txt --jsonl | xxd | tail -1
00000060: 656e 7422 3a22 5c75 4645 4646 4865 6c6c  ent":"\uFEFFHell

$ ... --jsonl | python3 -c "import sys,json; print(json.loads(sys.stdin.read())['content'].encode())"
b'\xef\xbb\xbfHello BOM\n'
```

The serializer emits the six-character escape `\uFEFF`, not a raw BOM byte inside the JSON, and a consumer that parses and re-encodes gets back the exact bytes the package shipped. Base returned a document three bytes shorter. GPT examined the same shapes and reached the same conclusion independently. U+FEFF is also still lost for UTF-16/UTF-32 files -- both models checked, both confirmed it is identical on base, so it is pre-existing and out of scope.

### Intended, with the message improved: the scope refusal is atomic (Gemini, medium -- kept)

`--path "*" --frontmatter` over a package containing an image is refused rather than scoping the Markdown files and skipping the rest. Gemini called this a workflow regression; GPT examined the same behavior and endorsed the atomic refusal. I am keeping it: a selection is one request, and answering part of it while silently dropping the rest reports success for files that were never scoped -- the defect class this series exists to close. On base the same command applied frontmatter scoping to the PNG and exited 0.

What was fair in the finding is that the refusal did not say what to do instead. It now does:

```text
Error: --frontmatter/--yaml-header and --body apply to Markdown documents;
'Icon.png' is not Markdown. Narrow the selection to Markdown, for example --path "*.md".
```

Recorded in `docs/design/output-shapes.md` so the choice is not rediscovered as a bug.

### Also verified not defects

Both models independently cleared the `InvalidOperationException` in `WritePackagePrintProjection` as genuinely unreachable (`ValidatePackagePrintSelection` admits only the three printable sections; lens and discovery modes intercept `--print` earlier), and the `ReferenceEqualityComparer` dictionary in the compatibility `Write` overload as correct for duplicate-valued rows, with `MarkHonored` unchanged and `--row first`/`last`/out-of-range at parity.

### Validation

CLI 2416/0 (14 skipped, `ilasm`/`ildasm` absent), services 343/0, solution builds clean, `markdownlint` clean. Both new behaviors mutation-tested.

Sending both models back to the fixed head now.

richlander:
## Round 4 reconciliation — both fixed-head reviews clean at `bd5c133e`

GPT-5.6 Sol and Gemini 3.1 Pro both re-reviewed the fixed head and report **no blocking findings**. This round was a single question, and it took three commits to settle because the first two answers each closed one hole and opened another.

### The question: when does the readme role decide a document's kind?

The role exists because NuGet renders whatever the manifest declares, so a package declaring `<readme>README</readme>` with no extension still ships Markdown. The danger is the mirror image: a package declaring `<readme>images/logo.png</readme>` must not have the link rewriter run over it.

Three rules were tried.

| declared readme | `Path.HasExtension` (`cada7b6f`) | `NamesASuffix` (`cb9d987c`) | `NamesAKind` (`bd5c133e`) |
| --- | --- | --- | --- |
| `README`, `docs/GUIDE` | Markdown | Markdown | Markdown |
| `README.md` | Markdown | Markdown | Markdown |
| `images/logo.png` | verbatim | verbatim | verbatim |
| `images/logo.png.` | **corrupted** | **corrupted** | verbatim |
| `.png` | verbatim | **corrupted** | verbatim |
| `.README` | scope refused | Markdown | scope refused |

GPT found `.README` against the first rule: .NET reads a whole dotfile name as an extension, so a hidden README was told it names a suffix. Both models then found that the second rule reopened `.png`, whose only dot is at index zero.

### Why it settled where it did

`.png` and `.README` are structurally identical — a leading dot and nothing else. Gemini put it plainly: *"structurally, a file named `.png` is indistinguishable from `.README`."* No reading of dot positions separates them. Only a list of known binary suffixes would, and that list goes stale and still guesses wrong at the edges.

So the tie is broken on the asymmetry of the two mistakes, not on which reading of a dot is more correct:

```csharp
private static bool NamesAKind(string path)
    => Path.GetFileName(path.AsSpan()).Contains('.');
```

Refusing a Markdown scope on `.README` is loud, names the file, exits 1, and leaves the document readable and printable without a scope. Corrupting a declared PNG returns a damaged file at exit 0. Those are not equally bad, and the silent one is the defect class this whole series exists to close.

It also closes `logo.png.`, which **neither** earlier rule covered — a real hole, present on base, that only surfaced because the first two attempts disagreed and forced the shape to be enumerated.

Both models were asked to argue the tie should break the other way, with a reproduction of harm. Neither did. GPT: *"I do not consider that a defect under the chosen rule."* Gemini: *"the asymmetry argument is robust... `bd5c133e` is strictly superior to both previous rules."*

### Independently verified by both models at this head

- `.README` refuses a scope loudly; unscoped `--content`/`--print` return it intact with its original blob URL.
- `.png`, `.logo.png`, `logo.png.`, `a.b.c`, `..README`, `...`, `docs/.hidden` all stay verbatim and reject Markdown scopes. Base rewrites `logo.png.`; head preserves it.
- Empty declarations and directory-ending paths resolve no README and never reach the predicate. Dotted directories (`docs.v1/GUIDE`) and manifest backslashes normalize correctly. A 240-character name is fine. The predicate cannot throw.
- Compliant `.md` readmes short-circuit on extension before the predicate is evaluated, across `--path @readme`, the `Package README file` section, `--print`, scoped output, link rewriting, and multi-package fan-out. NuGet mandates `.md`, so no real package shape is affected.
- Both theories are non-vacuous: the old rules fail them where they should.

### Everything else from earlier rounds, confirmed still fixed

- Declared readme keeps its role when `README.md` also exists (GPT, round 3) — path normalization, directory and missing declarations, `--path @readme`, and multi-package fan-out all checked; no inconsistency with the README section.
- Removed `--readme` diagnosed where the parser rejects it rather than by token scan — `--out --readme` writes its file at exit 0, `System.Private.CoreLib --readme` gets the router's own answer, and `package X --readme`, bare `X --readme`, `package X --readme=true`, `--json`, `--bare` and local `.nupkg` paths all keep the migration message. Gemini specifically confirmed `package search --readme` now searches for the keyword, which is correct — that subcommand never had the flag.
- BOM fidelity, one-fetch-per-`--print`, the `ReferenceEqualityComparer` mapping, `MarkHonored`, `--row` parity, and the unreachable-section throw all re-cleared.

### Validation

CLI 2421/0 (14 skipped, `ilasm`/`ildasm` absent), services 343/0, solution builds clean, `markdownlint` clean. Every behavior change in this round is mutation-tested against both superseded rules.

Unchanged and out of scope, confirmed identical on base by both models: UTF-16/UTF-32 preambles are still dropped, and true binary content through a text pipeline is still lossy at the byte level.

richlander:
Ready to merge

Head `bd5c133e`. Both fixed-head adversarial reviews clean (GPT-5.6 Sol and Gemini 3.1 Pro, four rounds). CI green: `changes`, `markdownlint`, `build-net10`, `test (ubuntu-26.04, linux-x64)`. Mergeable, state CLEAN. Locally: CLI 2421/0 (14 skipped, `ilasm`/`ildasm` absent), services 343/0, solution builds clean, `markdownlint` clean.

Non-blocking follow-ups, deliberately not in this PR:

- UTF-16/UTF-32 preambles are dropped by the text pipeline, and true binary content read through it is lossy at the byte level. Both confirmed identical on base by both reviewers.
- Whether `project --readme` shares the content reader fixed here was not checked.
- A Unicode dot homoglyph (`logo\u2024png`) bypasses `NamesAKind` and receives Markdown treatment. It states no filesystem suffix and is indistinguishable from a supported extensionless Markdown name, so it is not a defect under the chosen rule; noted for the record.

PR #3449, opened 2026-07-29
Follow ExportedType forwarders when resolving external type definitions

### LINKED ISSUE #3400: CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition fails deterministically on Windows: forwarder hop out of facade assemblies never resolves

richlander:
Fixes #3400.

## What was wrong

Cross-assembly type resolution scanned only the referenced assembly's `TypeDefinitions`. On a shared-framework (implementation) layout the referenced framework assembly is usually a pure facade — `System.Runtime.dll` carries **one** TypeDef and ~950 `ExportedType` forwarders into `System.Private.CoreLib` — so every one of its types resolved to nothing.

That is the whole of #3400: all 122 framework TypeRefs in `System.Console` found their assembly, every assembly resolved to a real file, and not one produced a definition. Both earlier stages were healthy; the forwarder hop was simply absent.

## The change

`IndexBuilder` now follows a matching top-level `ExportedType` forwarder into the implementation assembly, bounded at 8 hops so a malformed or hostile chain cannot spin the loop (real chains are one or two: `netstandard` → `System.Runtime` → `System.Private.CoreLib`).

Only forwarders implemented by an `AssemblyReference` are followed. An `ExportedType` implemented by another `ExportedType` is a nested entry and a `File` implementation is a multi-module member, so neither can satisfy a top-level lookup — those are skipped rather than guessed at.

`ResolveReferencedAssembly`/`ScopeForReference` now take the `MetadataReader` explicitly, because a hop reads the next assembly reference out of the *facade's* metadata, not the inspected assembly's. The resolution cache is keyed on `AssemblyReferenceIdentity`, which is reader-independent, so it is unaffected.

## The existing test was never a gate on this

`CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition` breaks at the first success, so on a layout where some referenced assembly still defines its own types it passes without ever crossing a facade. The issue asked whether the green runs were coverage — they were not.

`CrossAssemblyMetadataResolver_FollowsTypeForwarderOutOfFacadeAssembly` is the gate. It selects only TypeRefs that **provably** require the hop (the referenced assembly has no matching TypeDef but does have a matching forwarder), reads the referenced assembly itself as an independent oracle rather than asking the product where the type went, and asserts resolution lands in a *different* assembly than the facade. It carries a non-vacuity guard on the candidate count.

**Tamper-verified:** with the hop disabled it fails with `System.Object is a forwarder in System.Runtime.dll; the hop must follow it`. The pre-existing test does not fail in that state.

## Validation

Windows, SDK `11.0.100-preview.7`, runtime `Microsoft.NETCore.App 11.0.0-preview.6.26359.118`, Release.

- `dotnet run --project src/ILInspector.Analysis.Tests -c Release` — **575 passed, 0 failed** (was 556 with the standing 1 failure)
- `dotnet run --project src/dotnet-inspect.Tests -c Release` — 2353 total, 2 failed

Those 2 CLI failures (`LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceLinkDoor`, `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically`) are **pre-existing and unrelated**: they reproduce identically on a second worktree at `origin/main` with no product change. One is a Markdown line-ending bug fixed separately; the other is a SourceLink grouping failure that still needs triage.


richlander:
## Review reconciliation — GPT (fixed head `16bf6e8d`-era, now superseded)

Both findings were independently verified rather than accepted at face value. One is adopted, one is dismissed with its underlying fact stated honestly.

### Finding 2 — bounded-metadata-traversal contract — **accepted, fixed**

> The forwarder walk is depth-capped, but `docs/design/bounded-metadata-traversal.md` requires cycle detection by identity.

**Legitimate.** I confirmed the document is a live contract and not design history — `src/ILInspector.MetadataPrimitives` is a real project — and that rule 3 says cycles are `detected by identity, not merely stopped by a depth ceiling`. Metadata here is untrusted input, so the rule applies.

Fixed in d8583ee8. `TryResolveTopLevelTypeInAssembly` now threads a `HashSet<AssemblyReferenceIdentity>`, seeded with the assembly the TypeRef already named so a forwarder pointing back at the facade is a repeat visit rather than a fresh hop. `MaxTypeForwarderHops` stays as a second bound on chain length.

I also checked whether the substrate already owned this walk before adding a parallel one: `MetadataRelationshipTraversal.WalkExportedTypeImplementationChain` walks the *intra-assembly* nested `ExportedType` chain and stops at an `AssemblyReference`. The cross-assembly facade hop is a genuinely different traversal, so this is not a duplicated abstraction.

**Gate and tamper evidence.** `CrossAssemblyMetadataResolver_ForwarderCycleTerminatesAndFailsHonest` builds two synthetic assemblies that forward `Cyc.T` at each other and asserts the walk terminates with an honest miss. No compiler emits a forwarder cycle, so hand-built metadata is the only way to reach the state. Removing both bounds kills the test process with `STATUS_STACK_OVERFLOW` (exit `0xC00000FD`); restoring them returns 576/0.

**One thing the gate does *not* prove, stated rather than implied.** Identity detection's narrower benefit — each assembly rescanned at most once instead of up to the ceiling — is *reasoned but not gated*. `_referencedAssemblyCache` means the resolver is consulted once per identity under either bound, so resolver call counts are identical and the rescan saving is invisible to any observer outside the walk. The test's doc comment says so. Per `AGENTS.md`, unverified is acceptable when marked; unmarked would not be.

### Rule 5 (typed rejection) — **deliberate non-action**

`bounded-metadata-traversal.md` also says budget exhaustion should yield a typed rejection, never an ordinary empty result. Not adopted. `TryResolveExternalTypeDefinition` returns a nullable tuple and every sibling failure path on this seam returns `null`; converting to a typed rejection is a cross-cutting refactor of a currently-dormant seam and would be a different PR. Flagging it as a known, reasoned gap rather than silently leaving it.

### Finding 1 — resolution unreachable from product — **out of scope, but the fact is real**

> `TryResolveExternalTypeDefinition` is not called from any product path.

**Factually correct.** It is reached only from tests and from internal recursion in `TryResolveNestedExternalType`; no product path passes a resolver to `LibraryBodyIndex.Open`, and `Open_WithResolverDoesNotChangeIndexShape` asserts `ResolveCalls == 0`.

Wiring cross-assembly resolution into analysis is a separate feature, not this fix. But the honest consequence is worth stating plainly rather than leaving a reader to infer a bigger win: **the blast radius of this change is nil today.** What it fixes is a seam that was silently incapable of resolving anything on a shared-framework layout, which is what #3400 reports and what would have made that future wiring quietly return nothing.

### Why the original test never caught this

`System.Runtime.dll` on a shared-framework (implementation) layout has **1 TypeDef and ~950 ExportedType forwarders** into `System.Private.CoreLib`. Reference-assembly layouts never need the hop, which is why Linux CI stayed green. The pre-existing test also breaks at the first success, so it could never prove the hop existed — hence the added gate.

### Validation

`ILInspector.Analysis.Tests`: **576 passed, 0 failed** (Release, Windows). Was 556/1 before the fix.

Re-review of the fixed exact head is in progress; not marking ready until it is clean.

richlander:
## Closing as superseded by #3437

#3437 landed the same fix on `main` (`4920d491`) while this PR was in review, and it is a **strict superset** of this one. Closing rather than resolving the conflict, because merging this would replace a better implementation with a worse one.

Verified on Windows at `origin/main` in a clean worktree: **`ILInspector.Analysis.Tests` 584 passed, 0 failed**, including the originally failing `CrossAssemblyMetadataResolver_ResolvesFrameworkTypeDefinition`. Issue #3400 is genuinely fixed there.

### What #3437 has that this PR does not

The decisive difference is security, and it is exactly the attack point my own reviewers raised against this PR.

MAI, asked to construct the concrete attack, confirmed it: `ScopeForReference` as written here reads the public-key-token from *whichever assembly the hop is currently in*, and `FrameworkAssemblyKeys.IsFrameworkReference` gates only on that token — so an untrusted assembly can emit a reference carrying a framework token and cause a hop to resolve under `AssemblyResolutionScope.Platform` where the pre-change code used `Any`. That widens what the tool opens from disk based on untrusted metadata, against `docs/design/untrusted-data-threat-model.md`.

`main` closes that hole and I did not. `NextHopScope` enforces that a forwarder may only *tighten* scope, never widen it — `a forwarder must not launder a platform reference` — gated by `NextHopScope_AnyForwardedToFrameworkSignedAssembly_TightensToPlatform`, `NextHopScope_AnyForwardedToUnsignedOrThirdPartyAssembly_StaysAny`, `NextHopScope_PlatformIsNeverDowngraded`, `ForwarderIntoFrameworkSignedAssemblyIsResolvedUnderPlatformScope`, and `ResolvesEveryForwarderHopUnderPlatformScope`.

Beyond that, `main` iterates instead of recursing (no stack depth exposure at all), reuses `TypeForwardResolver`'s decode rather than open-coding it, and already had the identity-based cycle detection that GPT's review asked me to add here.

### Nothing left to salvage

Both gates I added exist on `main` under different names:

| This PR | `main` |
| --- | --- |
| `CrossAssemblyMetadataResolver_FollowsTypeForwarderOutOfFacadeAssembly` | `CrossAssemblyMetadataResolver_FollowsForwardersToDefiningAssembly` |
| `CrossAssemblyMetadataResolver_ForwarderCycleTerminatesAndFailsHonest` | `CrossAssemblyMetadataResolver_ForwarderCycleTerminatesWithoutResolving` |

So there is no follow-up PR to carve out of this branch.

### One thing worth keeping from the review

MAI's answer on nested forwarders is a real, still-open gap in **`main`**, not in this branch: both `LibraryBodyIndex` and `TypeForwardResolver` only follow `ExportedType` rows whose `Implementation.Kind == HandleKind.AssemblyReference`, so an `ExportedType` implemented by another `ExportedType` (a nested forwarder entry) is skipped. MAI probed the installed .NET packs on this machine and found no such row, so it is latent rather than live — and the behavior is an honest `null`, never a wrong definition. Filing that separately rather than losing it here.

Thanks to both reviewers; the scope-spoofing finding is what made the right call obvious.

PR #3450, opened 2026-07-29
Skip nested checkouts when censusing packable projects

### LINKED ISSUE #3422: `PackagingSurfaceTests` fails locally because it scans `.worktrees/` copies of the shipping projects
`PackagingSurfaceTests.OnlyTheShippingToolsOptIntoPackaging` enumerates every `*.csproj` under the repository root and asserts set equality against the shipping-project census. Its exclusion list does not cover `.worktrees/`, so the test fails for any contributor who keeps linked worktrees under the repo root — which is the workflow `AGENTS.md` prescribes.

## Reproduction

With any worktree present under `<root>/.worktrees/`:

```console
$ dotnet run --project src/dotnet-inspect.Tests -c Release -- -method "*OnlyTheShippingToolsOptIntoPackaging*"
    DotnetInspector.Tests.PackagingSurfaceTests.OnlyTheShippingToolsOptIntoPackaging [FAIL]
      Assert.Equal() Failure: Collections differ at index 0
                 ↓ (pos 0)
      Expected: "src/dotnet-inspect/dotnet-inspect.csproj"
      Actual:   ".worktrees/evil-run-0728/src/dotnet-inspect/dotnet"···
```

The worktree's own copy of `src/dotnet-inspect/dotnet-inspect.csproj` is a legitimately packable project, so it is collected and sorts ahead of the real one.

Passes in CI, which clones fresh and has no worktrees — so the failure only ever hits local runs, where it reads as a real regression in the change under test.

## Root cause

`src/dotnet-inspect.Tests/PackagingSurfaceTests.cs:98`:

```csharp
static bool IsExcluded(string root, string path) =>
    Path.GetRelativePath(root, path)
        .Split(Path.DirectorySeparatorChar)
        .Any(static segment => segment is "bin" or "obj" or ".git" or "artifacts" or "node_modules");
```

`.worktrees` is missing from the segment list.

## Suggested fix

Add `.worktrees` to the excluded segments.

This does not weaken the property the test defends. The two anti-vacuity guards are unaffected: `Assert.True(scanned > 0, ...)` still holds because the primary checkout's own projects are still scanned, and the sibling `PackagingIsOffByDefaultForEveryProject` — which is the guard that actually matters, since it fails if the `IsPackable=false` default is deleted — does not use `ScanProjects` at all. Excluding a nested checkout removes duplicates of projects that are already being censused from the real tree, rather than removing coverage.

A more thorough alternative is to skip any directory containing a `.git` *file* (as opposed to directory), which is precisely the marker of a linked worktree, and would also cover worktrees placed under the root with a different name. The segment-list fix is smaller and matches this repo's `.worktrees/` convention.

## Notes

Pre-existing; introduced with the test in #3330. Surfaced repeatedly while validating #3229, where it was the sole local failure against an otherwise green 2282-test run and had to be manually discounted on every run.

richlander:
Fixes #3422.

## What was wrong

`PackagingSurfaceTests.OnlyTheShippingToolsOptIntoPackaging` enumerates every `*.csproj` under the repository root. Its exclusion list covered build output but not a nested checkout, so it failed for any contributor whose linked worktree lives under the root — the workflow `AGENTS.md` prescribes. The worktree's own copy of `src/dotnet-inspect/dotnet-inspect.csproj` is legitimately packable, is collected, and sorts ahead of the real one.

CI clones fresh and never sees it, so the failure only ever lands locally, where it reads as a regression in whatever change is under test.

## The change

The scan prunes on the `.git` marker rather than on a directory name: a linked worktree marks itself with a `.git` **file**, a nested clone or submodule with a `.git` **directory**. This covers a nested checkout under any name instead of only the repository's `.worktrees/` convention — the more thorough of the two options the issue raised.

## This does not weaken the census

It removes duplicates of projects the real tree already contributes:

- `Assert.True(scanned > 0, ...)` still holds from the primary checkout's own projects.
- `PackagingIsOffByDefaultForEveryProject` — the guard that actually matters, since it fails if the `IsPackable=false` default is deleted — does not use `ScanProjects` at all.

## The gate

`ProjectScanSkipsNestedCheckoutsAndBuildOutput` is hermetic over a temp tree, so it holds whether or not the machine running it has a worktree present. It carries the negative case: an ordinary nested directory with **no** `.git` marker is not a checkout and stays censused, so the prune keys on the marker rather than on depth or on a name.

## Validation

Windows, Release.

**Repro confirmed live, then fixed.** With a linked worktree created at `<root>/.worktrees/repro-3422`:

- unfixed: `OnlyTheShippingToolsOptIntoPackaging` fails with `Actual: \".worktrees/repro-3422/src/dotnet-inspect/dotnet-in\"···` — byte-for-byte the symptom in the issue
- fixed, same worktree still present: `PackagingSurfaceTests` 3/3 green

Full suite: `dotnet run --project src/dotnet-inspect.Tests -c Release` — 2354 total, 2 failed. Both failures are pre-existing and unrelated (`LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceLinkDoor`, `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically`); they reproduce identically on a separate worktree carrying no test change.


richlander:
## Review reconciliation

**Tier: single review.** Test-only, one file, no product path touched — more than trivial but not high risk.

**Reviewer: MAI-Code**, isolated in a linked review worktree detached at the exact head `72783fee`.

**Result: clean — no blocking or non-blocking findings.**

What the reviewer verified independently, with real runs:

- Built the solution and ran `PackagingSurfaceTests` — 3/3 pass.
- Created a **real** linked worktree (`git worktree add --detach .worktrees\probe HEAD`) and confirmed the tests still pass with it present. This is the exact condition that fails on `main`.
- Confirmed git's linked-worktree marker format by reading `.worktrees\probe\.git`, which contained `gitdir: C:/Users/rlander/git/dotnet-inspect/.git/worktrees/probe` — i.e. the `.git`-file detection the fix keys on is the real marker, not an assumption.
- **Tamper-verified the gate:** reverting `IsExcludedDirectory` to the old name-only behavior fails *both* `OnlyTheShippingToolsOptIntoPackaging` and the new `ProjectScanSkipsNestedCheckoutsAndBuildOutput`, with the scan visibly including `.worktrees/probe/.../dotnet-inspect.csproj`. So the new test is a genuine gate, not vacuous.
- Restored the tamper and left the worktree clean.

The central question put to the reviewer — whether the prune could let a real packaging regression through — came back negative: the prune removes duplicates of projects the primary tree already contributes, and `PackagingIsOffByDefaultForEveryProject` (the guard that fails if the `IsPackable=false` default is deleted) does not use `ScanProjects` at all.

## CI

All required checks pass: `test (ubuntu-26.04, linux-x64)` and `changes`. `pack`, `build-net10`, and the smoke lanes are path-gated and correctly skipped for a test-only change.

Ready to merge

PR #3452, opened 2026-07-29
Keep the token stream the slicer's scanner already computes

richlander:
## What this changes

The slicer has always lexed each line to place braces, and has always thrown that work away at the line break. Every predicate then re-derived "am I inside a comment, a literal, or an attribute list?" from a string it was handed — and a caller that forgot to thread the carried lexical state got a *plausible wrong answer* rather than an error. That shape is the root of two of the defect families found across the ten review rounds in this series.

This keeps what the scan already works out.

**No predicate moves onto the token stream here, and no behavior changes.** The existing suites are the evidence for that. Migrating the ~16 predicates is the next PR; doing it in the same change would mean the migration was the first thing that ever looked at the stream.

## Why a sink, not a second lexer

The literal, interpolation, and raw-string rules in `Scan` took ten review rounds to settle — `$$` hole runs, verbatim vs raw quote counting, `@` before a triple quote, holes spanning lines since C# 11. Rewriting them independently would have risked regressing hard-won fixes to gain a differential oracle I do not actually need. So the token stream is emitted **from the existing loop**: same code, same rules, output retained instead of discarded.

Two branches now consume a run where they previously stepped one character at a time (plain literal content, and identifier runs). Both are behavior-identical — none of the characters consumed is a delimiter, and the loop's early-exit flags cannot change mid-run — and the reasoning is in the diff comments.

## The model

`ScanToken` carries kind, position, block depth, and bracket depth. Two decisions worth review:

- **A delimiter is placed with the text it bounds.** An opener reports the depth *outside* it, a closer the depth *inside*, so `{` and `}` describe their block rather than each other.
- **`DepthKnown` is corrected retroactively.** The scan discovers it lost its place at *end* of line — after tokens on that line were already emitted. Those tokens are rewritten rather than left reporting a depth that has since become meaningless, because a stale depth reads exactly like a real one.

## Gates

| Gate | What it holds |
| --- | --- |
| `EveryCharacterOfTheCorpus_IsAccountedFor` | Every non-whitespace character of every C# file the corpus PDBs point at falls inside exactly one token — **36,651 lines**, zero gaps |
| same test, second half | Coverage alone is also satisfied by a scanner that calls each line one token, so the stream must additionally be decomposed (>3 tokens/line) and exercise every kind the corpus can reach |
| `CoverageCheck_ReportsATokenThatWasRemoved` / `_ThatWasShortened` | Non-vacuity gate for the checker itself |
| 20 pinning tests | raw/verbatim/interpolated/`$$` literals, block and line comments, char literals, attribute lists across lines, conditional vs non-conditional directives, verbatim identifiers |

`Directive` is the one kind no corpus file contains. It is named in `KindsTheCorpusCannotReach` and compared for **set equality**, so if a corpus assembly ever gains a directive the test fails and the entry gets deleted — the entry cannot go stale silently.

**Verified by tampering, not by assertion:** suppressing the token for three-character words fails the corpus gate at `DiffAsmFixtures.Target/Api.cs` line 8 column 24.

## Validation

Release, at `c098f913` (rebased onto `adaf9414`):

| Suite | Result |
| --- | --- |
| `DotnetInspector.CSharpBodySlicer.Tests` | 277, 0 failed (244 + 33 new) |
| `ILInspector.Metadata.Tests` | 998, 0 failed |
| `DotnetInspector.Services.Tests` | 343, 0 failed |
| `dotnet-inspect.Tests` | 2361, 0 failed |

The commit message quotes the counts measured before the rebase onto `adaf9414`; the table above is the current head. The commit was not amended.

## Naming

`ScanToken`/`ScanTokens` sit in `DotnetInspector.CSharpBodySlicer` and are `internal`. They deliberately avoid `Lexer`/`Tokenizer`: this is not a general C# lexer — keywords are matched by text, numeric literals are not decomposed, and `@class` comes back as `@` plus a word — and a lexer name would invite a fidelity it does not have. #3439's reserved noun **Scanner** (a sequential pass over metadata in `ILInspector.Metadata`) is not used; no `*Scanner` type is declared here.

### DIFF
--- a/src/DotnetInspector.CSharpBodySlicer/BodySlicer.cs
+++ b/src/DotnetInspector.CSharpBodySlicer/BodySlicer.cs
@@ -1345,7 +1345,10 @@ public static class BodySlicer
 
                     if (run < frame.DollarRun)
                     {
-                        // Too short to delimit a hole.
+                        // Too short to delimit a hole. The depth passed here is inert: this
+                        // fragment always follows literal content, so coalescing fuses it into
+                        // the preceding StringLiteral, which keeps the first fragment's depth.
+                        // A wrong value here cannot reach an observer, so no test gates it.
                         Emit(depth, ScanTokenKind.StringLiteral, i, run);
                         i += run;
                         continue;
@@ -1409,7 +1412,8 @@ public static class BodySlicer
                         continue;
                     }
 
-                    // A shorter run inside a raw literal is content.
+                    // A shorter run inside a raw literal is content. As above, the depth is
+                    // inert: the fragment is always coalesced into a preceding StringLiteral.
                     Emit(depth, ScanTokenKind.StringLiteral, i, run);
                     i += run;
                     continue;
--- a/src/DotnetInspector.CSharpBodySlicer/BodySlicer.cs
+++ b/src/DotnetInspector.CSharpBodySlicer/BodySlicer.cs
@@ -1239,7 +1239,6 @@ public static class BodySlicer
         int i = start;
         bool opened = false;
         bool bracketOpened = false;
-        int entryCount = tokens?.Count ?? 0;
 
         void Emit(int atDepth, ScanTokenKind kind, int column, int length) =>
             EmitAt(atDepth, state.BracketDepth, kind, column, length);
@@ -1252,7 +1251,10 @@ public static class BodySlicer
             // Literal text arrives in fragments — an escape, a quote run, a stretch of plain
             // characters — because the scan decides what each one means separately. They are one
             // token to a caller, so adjacent fragments coalesce instead of surfacing that.
-            if (kind == ScanTokenKind.StringLiteral && tokens.Count > entryCount)
+            // Adjacency is same line and touching columns: a literal that spans lines emits one
+            // token per line, so an empty line inside a verbatim literal must not fuse the
+            // fragments on either side of it.
+            if (kind == ScanTokenKind.StringLiteral && tokens.Count > 0)
             {
                 var previous = tokens[^1];
 
--- a/src/DotnetInspector.CSharpBodySlicer/BodySlicer.cs
+++ b/src/DotnetInspector.CSharpBodySlicer/BodySlicer.cs
@@ -1586,6 +1586,15 @@ public static class BodySlicer
                         frame.HoleDepth = 0;
                         state.Replace(frame);
                         int closerAt = i;
+
+                        // Min because a longer run closes the hole with DollarRun braces and
+                        // leaves the rest as literal text. This cannot be gated through the token
+                        // stream: the leftover braces are re-consumed by the literal branch on the
+                        // next step and coalesce back into this token, so consuming one brace here
+                        // produces a byte-identical stream. Verified inert over every string up to
+                        // length 7 on the `$ " { } \ a` alphabet (335,922 inputs, 0 divergences).
+                        // It is kept because it is what the language says, and a consumer that
+                        // wants un-coalesced fragments would be able to tell.
                         i += frame.DollarRun > 1 ? Math.Min(run, frame.DollarRun) : 1;
                         EmitAt(depthBefore, bracketBefore, ScanTokenKind.StringLiteral, closerAt, i - closerAt);
                         continue;
--- a/src/DotnetInspector.CSharpBodySlicer/BodySlicer.cs
+++ b/src/DotnetInspector.CSharpBodySlicer/BodySlicer.cs
@@ -1345,10 +1345,7 @@ public static class BodySlicer
 
                     if (run < frame.DollarRun)
                     {
-                        // Too short to delimit a hole. The depth passed here is inert: this
-                        // fragment always follows literal content, so coalescing fuses it into
-                        // the preceding StringLiteral, which keeps the first fragment's depth.
-                        // A wrong value here cannot reach an observer, so no test gates it.
+                        // Too short to delimit a hole.
                         Emit(depth, ScanTokenKind.StringLiteral, i, run);
                         i += run;
                         continue;
@@ -1412,8 +1409,7 @@ public static class BodySlicer
                         continue;
                     }
 
-                    // A shorter run inside a raw literal is content. As above, the depth is
-                    // inert: the fragment is always coalesced into a preceding StringLiteral.
+                    // A shorter run inside a raw literal is content.
                     Emit(depth, ScanTokenKind.StringLiteral, i, run);
                     i += run;
                     continue;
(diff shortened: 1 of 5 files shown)


richlander:

Two reviewers, alternating on a fixed head each round: **GPT-5.6** and **Gemini 3.1 Pro**. Both are clean at `910726940`, which is also the head CI passed. Twenty-one rounds. Every round but the last found something real.

The rules were fixed throughout. **In scope:** weakening a value the test computes or supplies — an alphabet, a seed list, an input array, a theory row, a recorded value, a loop body — so the suite stays green **at an unchanged test count of 330** while a product defect survives. **Out of scope:** deleting an assertion, deleting a test or an `[InlineData]` row, or editing an assertion's expected literals. "Delete assertion X, then weaken Y" is degenerate.

### What the reviewers found

**GPT.** Line reversal; single-line stand-in; inner-loop collapse; `seconds.Clear()`; dimension trade; alphabet `{`/`\`; `seconds` contents; raw-literal erasure; the `len5` counterexample; the verbatim-interpolated seed; `QuoteRun`; joint `(DollarRun, QuoteRun)` pairing; the plain verbatim frame; record-before-scan; the two-array seam; the self-adjusting theory expectation; alphabet-pinned-by-size; opener-pin-placed-before-the-scan; unpinned sweep input sets; distinct-row substitution; parallel recording; relative-only raw depth; unpinned raw fixture input; the cross-line coalescing filter; opener-loop collapse; `(opener, tail)` joint pairing; **the detached assertion** (the asserted projection could be read off the *opener* token, which shares kind and both depths); **the post-count filter** (the cross product's expected side is built from `tails`, so narrowing `tails` after its count assertion moves both sides together); **trailing-space padding** (the exact vector characterised *trimmed images*, and trimming is lossy exactly where the attack lives); **the balanced split/merge trade** (splitting 9,855 `$@` punctuators and merging 9,855 brace pairs holds all three per-kind totals while every boundary moves); and **the `$`/`@` permutation** — the sharpest of the series, below.

**Gemini.** Cross-arm record; short-opener verbatim return; the inert `content.Length > 1`; the `tails` seed; arm trades; joint pairing found independently; the 13th `Emit` site; `DepthKnown` at 7 sites; duplicate theory rows; distinct-variant rows; sweep-bound shrink; three projection-steering fixtures; **the argument-versus-post-hoc measurement flaw** (below); **the call-site alphabet substitution** (the alphabet *constant* was pinned, but the generator is handed an *argument* and reads only its **length**, so a same-length literal erased `\` with every count identical); **the unreconstructed sweep** (three integer totals and a constant are all marginals); and **the post-scan array swap** (`TextIn` resolves text against the *caller's* array, so reconstruction pins token **extent**, not the text scanned — scanning a `\`→`a` copy and swapping the original back rebuilt everything exactly).

### The two findings that changed how I was working

**Gemini, mid-series: mutate the argument, not the emitted value.** My sweep patched `tokens[^1]` *after* each `Emit`. Gemini mutated the argument instead. These are different experiments, because `EmitAt`'s coalescing branch fuses a fragment into the previous token and **keeps the earlier token's depth, discarding the argument entirely**. A post-hoc patch survives coalescing; a wrong argument does not. My sweep reported 0 survivors; the correct sweep reported **8** — every literal-fragment site. I had reported the invalid result to both reviewers as measured.

**GPT, last round: the sweeps are exhaustive, so the input set has symmetries.** The whole-stream fingerprint was summed across scans and therefore blind to *which* input produced which stream. Rewriting `$`→`@` and back on whitespace-free lines containing `a` and `"` **permutes the swept set onto itself**: every per-scan hash reappears against a different input, the total is unchanged, and inputs are misclassified throughout. Gemini had cleared that same fingerprint one round earlier by *reasoning* that cross-scan compensation requires recognising fixtures. It does not, when the input set is closed under a symmetry of the product's own behaviour.

### The one lesson, in about thirty forms

*The test's inputs are as much a gate surface as the product.* Counts don't pin shape; shape doesn't pin diversity; a collection's count doesn't pin its contents; **marginals never pin joints** (five separate instances); pins on *recorded* inputs don't pin *scanned* inputs; a theory whose expectation derives from its parameter pins nothing; an alphabet's size doesn't pin its characters; a pin placed before the loop it protects pins nothing; a filtered assertion discards the evidence that the fixture reached its branch; pinning a collection doesn't pin that the loop body uses it; a post-hoc mutation doesn't reach the path an argument mutation does; an *exact* count of a *lossy projection* is still only a marginal on the pre-image; and an assertion that names a value but not *which object* it read can be satisfied by another object.

Adding marginals never terminated — each new one traded against the last. What terminated it was pinning the **pair**: each scanned input folded together with the token stream it produced, into one value per scan.

### Where the gates ended up

Three exhaustive sweeps, each characterising its scanned input **exactly, from literals**, by a counting argument (over an alphabet of exactly N characters there are only N^L distinct strings of length L, so observing that many *distinct* inputs at each length is observing all of them), each pinning its scan count, and each folding every token's full state — kind, line, column, length, both depths, `DepthKnown` — together with the lines that produced it. Verified deterministic across separate processes.

### Corrections, non-actions, and my own errors

- Gemini's product-side input-rewriting mutation was **rejected as an oracle attack** — the product recognising the test's input shape defeats *any* suite and so distinguishes nothing. Applied literally at the head it was reported against, it also **did not reproduce**: 2 failures, not "330 all passing". Its underlying observation was correct and led to the population pin, which GPT then broke.
- Gemini's final two replies opened with the literal word `BLOCK` above a body reading `APPROVE` with "surviving product defect: None". Read as approvals; the substance is unambiguous.
- Earlier: Gemini retracted one false BLOCK (stale build) and returned two premature APPROVEs.
- **Mine:** a corrupted mutation baseline; a wrong `len5` removal; a silent `str.replace` bug; a "no product files changed" misstatement; moving a worktree under a running agent; **reporting the invalid 22-site sweep result to both reviewers as measured**; and **reasoning that site 1377 was unreachable instead of measuring it** — it is reachable, because an unterminated plain literal carries.
- **Measured residuals, accepted:** rewriting `ScanTokens(wrapped[2..])` to consume a fabricated `bareInput` survives; and one sweep bound can be shrunk together with its derived expectation.
- The `Layout_Count_CountsFilesRatherThanRenderedTreeLines` failure I reported for several rounds as pre-existing is **gone at this head**; nothing in the repository is failing.

### Validation at `910726940`

Slicer 330 · Analysis 584 · Services 343 · Metadata 998 (5 skipped) · CLI 2460 (14 skipped) — **zero failures**. Thirty-six test and product mutations, all caught. Argument-level `Depth`, `BracketDepth` and `DepthKnown` sweeps clean at all 22 emit sites; 19-site kind sweep and 9 hand-written product mutations clean. CI green on this exact head.

richlander:
Ready to merge

Issue #3445, opened 2026-07-29
Two LibraryCommand tests fail deterministically on Windows since the curated -D catalog


richlander:
Two `CommandExecutionTests` cases fail on every Windows run of the suite, and have since the curated `-D` catalog landed (#3229). They are green on Linux CI, so the whole suite is permanently 2-red locally on Windows and contributors have to carry a known-failure list in their heads to read a test run.

Repro (Windows, Release):

```
dotnet run --project src\dotnet-inspect.Tests -c Release -- -class DotnetInspector.Tests.CommandExecutionTests
```

```
dotnet-inspect.Tests  Total: 625, Errors: 0, Failed: 2, Skipped: 0, Not Run: 0
```

### 1. `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceLinkDoor`

`CommandExecutionTests.cs:7147`

```
Assert.DoesNotContain() Failure: Sub-string found
                                       ↓ (pos 411)
String: ···"                section\r\nSourceLink: Missing Files"···
Found:  "SourceLink: Missing Files"
```

The test warms the symbol cache with an explicit `-S "SourceLink: Availability"` render, then asserts that a bare `-D` lists the SourceLink family only under the `@SourceLink` door and never at the top level. On Windows `SourceLink: Missing Files` is present at the top level.

The test's own comment says the family lists under `-D` only when a local PDB exposes a SourceLink document, so the likely difference is that the warm step leaves Windows in a state where more of the family is considered discoverable than the assertion allows. Worth confirming whether the platform difference is in PDB acquisition or in the discovery predicate, because if it is the predicate then `-D` output genuinely differs by platform for the same package.

### 2. `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically`

`CommandExecutionTests.cs:8860`

```
Assert.NotEmpty() Failure: Collection was empty
```

The test runs `library System.Text.Json -v:d` through the in-process harness and collects lines starting with `## `. It gets none.

The shipped binary does not agree. Running the same arguments through `artifacts\bin\dotnet-inspect\release\dotnet-inspect.exe` produces 16 `## ` headings and CRLF line endings, so the renderer is fine and the divergence is somewhere in the in-process `RunAppAsync` path rather than in section rendering. Note the assertion that *does* pass is `Assert.Equal(0, exit)` — the command reports success while producing output with no sections at all, which is worth treating as a finding in its own right regardless of how the test is fixed.

### Notes

[...]

PR #3453, opened 2026-07-29
Hold the library section-to-scanner binding to a real gate

richlander:
## What this changes

Two of the seventeen library scanner keys were declared as consts, referenced by a section descriptor, and **never registered**:

| Key | Declared by | Registered in `CreateScannerRegistry` |
| --- | --- | --- |
| `TransitiveRefs` | `Dependencies` | no |
| `Symbols` | `Symbols` | no |

`ScannerRegistry.RunScanners` iterates its own registrations and runs the ones the caller asked for, so a requested key with no registration is skipped **in silence** — no throw, no log.

## Why nothing was visibly broken

Neither section ever depended on a scanner:

- **Dependencies** gets its tree from `LibraryMetadataService`, built inline under `options.IncludeDependencies` *before* the scanner phase.
- **Symbols** reads `PdbContext` fields set on the always-collected path — which is why its `CanRender` is an unconditional `true`.

Both keys were pure vestige. Confirmed by output rather than by reading:

```
library <asm> -S Symbols
library <asm> --dependencies -S Dependencies
```

Both are **byte-identical** before and after (`Compare-Object` → 0 differences).

## Why they survived

The test meant to catch this was `LibraryScannerRegistry_HasAllDetailedScanners`. It asserted only `Assert.NotEmpty(detailedScanners)`, and its own comment conceded the gap:

```csharp
// Registry should handle all of them without throwing
// (we can't easily inspect the registry, but we can verify it runs)
```

It read like a registration check in the test list and was one nowhere else. Both dead keys passed it. This is the `AGENTS.md` "asserted properties name their gate" failure mode in miniature — a green suite plus a confident name is indistinguishable from a verified property.

## The gate

Expose the two halves of the binding — `ScannerRegistry.RegisteredKeys` (supply) and `SectionPipeline.DeclaredScannerKeys` (demand) — and assert **set equality**.

Set equality rather than containment, so both directions fail: a section asking for a scanner nobody registered, *and* a scanner no section can ask for. Both sides are **derived** from the pipeline and the registry rather than restated as a literal list, so adding a section or a scanner cannot drift past the check.

**Confirmed non-vacuous by tampering.** Pointing the `Symbols` descriptor at an unregistered `"TamperKey"` fails with the offending key named:

```
Assert.Equal() Failure: Collections differ at index 11
Expected: "TamperKey"
Actual:   "TopLeverage"
```

`DeclaredScannerKeys` is defined on `SectionPipeline<TModel>` generally, but only the library pipeline declares scanner keys today — every descriptor in `ApiSectionDescriptors`, `PackageSectionDescriptors`, and `DiffSections` returns `null`, and none of those pipelines has a registry to pair with. The doc comment says so rather than implying wider coverage.

## Why this first

This is the opening slice of the library strand of the layering work ([`docs/design/inspection-layers.md`](docs/design/inspection-layers.md), #3439). L1 replaces the stringly-typed `ScannerKey` with a declared query type, at which point a dead binding becomes a compile error. Until then this test is the only thing standing between a section and data that silently never arrives — and the mechanical scanner-to-query conversion that follows needs a check that fails loudly when a binding is dropped mid-move.

No behavior change. No `package` files touched.

## Validation

- `dotnet build dotnet-inspect.slnx -c Release` — 0 errors.
- `dotnet run --project src/dotnet-inspect.Tests -c Release` — 2361 total, 1 failed, 10 skipped.
- The one failure, `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically`, is **pre-existing on `origin/main`**: verified by stashing this branch's changes, rebuilding at `adaf9414`, and reproducing the identical `Assert.NotEmpty() Failure: Collection was empty`. The underlying command succeeds when run directly, so it looks like a harness/environment issue rather than a product one. Untouched here; flagging separately.
- Skips are `ildasm`/`ilasm` absence, as expected.

### DIFF
--- a/src/dotnet-inspect.Tests/SectionPipelineTests.cs
+++ b/src/dotnet-inspect.Tests/SectionPipelineTests.cs
@@ -1120,17 +1120,19 @@ public class SectionPipelineTests
     }
 
     [Fact]
-    public void LibraryScannerRegistry_HasAllDetailedScanners()
+    public void LibraryScannerRegistry_RegistrationMatchesDeclaration()
     {
+        // Set equality, not containment, so both failure directions are caught: a section
+        // declaring a key nobody registered (its data silently never collected) and a registered
+        // scanner no section asks for (dead code). Derived from the pipeline and the registry
+        // rather than restated as a literal list, so adding a section or a scanner cannot drift
+        // past this test.
         var registry = LibrarySections.CreateScannerRegistry();
         var pipeline = LibrarySections.CreatePipeline();
 
-        // All scanner keys from detailed sections should be registered
-        var detailedScanners = pipeline.GetRequiredScanners(Verbosity.Detailed);
-
-        // Registry should handle all of them without throwing
-        // (we can't easily inspect the registry, but we can verify it runs)
-        Assert.NotEmpty(detailedScanners);
+        Assert.Equal(
+            pipeline.DeclaredScannerKeys.OrderBy(k => k, StringComparer.Ordinal),
+            registry.RegisteredKeys.OrderBy(k => k, StringComparer.Ordinal));
     }
 
     // ===== Presence flag / CanRender discovery tests =====
--- a/src/dotnet-inspect/Sections/LibrarySections.cs
+++ b/src/dotnet-inspect/Sections/LibrarySections.cs
@@ -10,8 +10,9 @@ namespace DotnetInspector.Sections;
 /// </summary>
 public static class LibrarySections
 {
-    // Scanner keys identify data collection steps in LibraryMetadataService
-    public const string ScannerTransitiveRefs = "TransitiveRefs";
+    // Scanner keys identify data collection steps in LibraryMetadataService.
+    // Every key here must be registered in CreateScannerRegistry and declared by at least one
+    // section. Gate: SectionPipelineTests.LibraryScannerRegistry_RegistrationMatchesDeclaration.
     public const string ScannerExtensionMethods = "ExtensionMethods";
     public const string ScannerClassifiedMethods = "ClassifiedMethods";
     public const string ScannerResources = "Resources";
@@ -19,7 +20,6 @@ public static class LibrarySections
     public const string ScannerUnionTypes = "UnionTypes";
     public const string ScannerTypeForwarders = "TypeForwarders";
     public const string ScannerInfoCounts = "InfoCounts";
-    public const string ScannerSymbols = "Symbols";
     public const string ScannerAuditSignals = "AuditSignals";
     public const string ScannerIntegrations = LibraryIntegrationCatalog.RollupName;
     public const string ScannerIntegrationOpportunities = "IntegrationOpportunities";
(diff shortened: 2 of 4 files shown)


richlander:
Note on the single test failure in the validation section: it is already tracked, and its root cause is now identified.

- #3445 — `Two LibraryCommand tests fail deterministically on Windows since the curated -D catalog`
- #3454 — `MarkdownSectionOrderer rewrites CRLF to LF`, which is the mechanism: the test splits on `Environment.NewLine` (CRLF on Windows) while the ordered document comes back LF-only, so nothing splits and the `## ` header list is empty. That matches the observed `Assert.NotEmpty() Failure: Collection was empty` exactly, and explains why the same command run directly renders its sections fine.

So the failure is unrelated to this PR in both cause and location, and is being handled separately. Nothing in this diff touches Markdown post-processing.

richlander:
## Review reconciliation

Reviewed adversarially with **MAI-Code** in an isolated read-only worktree detached at the exact head `c5ee801d`. Per `AGENTS.md`, this sits in the "more than trivial, but not high risk" tier — small diff, no behavior change, but it removes declarations and replaces a test, so it earns one review rather than none.

**No findings. Every claim in the PR body was independently verified**, several by stronger means than I used:

| Claim | How the reviewer checked it |
| --- | --- |
| Both keys dead at base | Counted the `.Add(Scanner...)` calls at `adaf9414` — 15 registrations, neither key among them |
| `RunScanners` is silent | `ScannerRegistry.cs:93` — iterates its own entries, no throw, no diagnostic path for an unknown key |
| No behavior change | Traced both data sources: `LibraryMetadataService.cs:140` (inline, pre-scanner) and `LibraryInspectionView.cs:691` (PDB-derived) |
| `ScannerKey` doesn't gate selection | Confirmed it reaches only `GetRequiredScanners`; selection still flows through `IsRequested`/`CanRender` |
| Gate is genuine set equality | `SectionPipelineTests.cs:1123` — equality, not containment |
| Doc comments name a real gate | Named gate exists; also confirmed `ApiSectionDescriptors`, `PackageSectionDescriptors`, `DiffSections` are all `ScannerKey => null`, so "nothing gates them" is accurate |
| Failure is pre-existing | Built and ran the test **at base `adaf9414` in a separate temp clone** and reproduced it |
| Byte-identical output | Ran both commands **at base in the temp clone** and diffed against head |

The last two are better evidence than the PR body carried: I had compared against a stashed working tree, the reviewer compared against a clean base build.

### One thing the reviewer did not report on, verified by me

I asked it to check for a **comparer mismatch** between the two sets, since `DeclaredScannerKeys` builds with `StringComparer.Ordinal` while `RegisteredKeys` is a `Dictionary.KeyCollection`. It did not address that specifically, so recording it here rather than leaving it implied:

`ScannerRegistry._scanners` is `Dictionary<string, Action<ScannerContext>> = []` (`ScannerRegistry.cs:69`) — the default comparer, which for `string` is ordinal. `GetRequiredScanners` likewise returns a default-comparer `HashSet<string>`, and `RunScanners` does an ordinal `Contains`. So the whole binding is ordinal end to end and the new assertion compares like with like. Two keys differing only in case would be distinct on both sides, consistently.

### CI

Green on `c5ee801d`: `build-net10` pass, `test (ubuntu-26.04, linux-x64)` pass at 10m5s, `changes` pass. The full suite passing on Linux is independent corroboration that the one Windows failure is the CRLF defect in #3454 and not this change.

Ready to merge.

richlander:
Ready to merge

PR #3455, opened 2026-07-29
Preserve document line endings when reordering Markdown sections

### LINKED ISSUE #3454: MarkdownSectionOrderer rewrites CRLF to LF, breaking section-ordered output on Windows
`MarkdownSectionOrderer.Apply` normalizes the document to LF and then rejoins on a hardcoded `'\n'`:

```csharp
var lines = markdown.Replace(""\r\n"", ""\n"").Split('\n');
...
return string.Join('\n', output);
```

Every sibling post-processing pass preserves the document's line ending via `MarkdownScan.DetectNewline` — `MarkdownTableRowLimiter.Apply` and `OutputFormatter.LimitRenderedTableRows` both do. The orderer is the only one that does not, and `MarkdownScan.DetectNewline` documents the convention it breaks:

> The line ending the text already uses, so a pass that splits on '\n' can rejoin without changing them. Row limiting selects which rows survive; it is **not licensed to rewrite CRLF to LF**, which would make the same output differ byte for byte depending on whether a row window was supplied.

That reasoning applies verbatim to reordering: the same document differs byte for byte depending on whether a section order was supplied.

## Observable effect

Measured on Windows against `main`:

```console
$ dotnet run --project src/dotnet-inspect -c Release -- library System.Text.Json -v:d > out.txt
CRLF=1  bareLF=373
```

One CRLF, 373 bare LFs. Any consumer that splits the output on `Environment.NewLine` gets a single giant line.

## Failing test

`CommandExecutionTests.LibraryCommand_DetailedOutput_RendersSectionsAlphabetically` fails deterministically on Windows:

```
Assert.NotEmpty() Failure: Collection was empty
  src\dotnet-inspect.Tests\CommandExecutionTests.cs(8986,0)
```

The test splits captured output on `Environment.NewLine` and selects lines starting with `## `. On Windows the split yields one line — the whole document — which does not start with `## `, so the header collection is empty. Exit code is 0 and the assertion on ordering is never reached, so the failure reads as ""no sections were rendered"" when sections rendered fine.

This was one of the two standing local failures being carried as noise while validating unrelated PRs (seen during #3400 and #3422 work).

## Why the unit test did not catch it

`OutputFormatterTests.MarkdownSectionOrderer_ReordersH2SectionsAndKeepsFenceHeadings` asserts `Assert.Contains(""B\n\n## Alpha"", output)`. Its input is a raw string literal, which carries whatever ending the source file is checked out with — CRLF here. So the assertion passed **because** the orderer rewrote CRLF to LF: the test was pinned to the bug. There is no test for line-ending preservation in the orderer at all.

## Fix

Use `MarkdownScan.DetectNewline` and rejoin on it, matching the sibling passes. The adjacency assertion in the existing unit test should normalize (that claim is about which lines end up adjacent, not about line endings), and line-ending preservation needs its own gate over both `\n` and `\r\n`.

richlander:
Fixes #3454.

## What was wrong

`MarkdownSectionOrderer.Apply` normalized the document to LF and rejoined on a hardcoded `'\n'`. Every sibling post-processing pass preserves the document's line ending via `MarkdownScan.DetectNewline` — `MarkdownTableRowLimiter.Apply` and `OutputFormatter.LimitRenderedTableRows` both do. The orderer was the only one that did not, and `DetectNewline` documents the convention it broke:

> a pass that splits on '\n' can rejoin without changing them ... it is **not licensed to rewrite CRLF to LF**, which would make the same output differ byte for byte depending on whether a row window was supplied.

That reasoning applies verbatim to a section order.

## Windows symptom

`library System.Text.Json -v:d` emitted **1 CRLF and 373 bare LFs**, so any consumer splitting on `Environment.NewLine` got a single line. `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically` failed deterministically with `Assert.NotEmpty() Failure: Collection was empty` — the split yielded the whole document as one line, none of which starts with `## `, so it reported no sections for a document whose sections rendered correctly.

## The existing unit test was pinned to the bug

`MarkdownSectionOrderer_ReordersH2SectionsAndKeepsFenceHeadings` asserted `Assert.Contains(""B\n\n## Alpha"", output)`. Its input is a raw string literal, which carries whatever ending the source file is checked out with — CRLF here — so that assertion passed **because** the orderer rewrote CRLF to LF. It was guarding the defect, not the property.

That claim is about which lines end up adjacent, so it now asserts against normalized text. Spelling a bare LF against raw output silently asserts the platform rather than the shape.

## The gate

`MarkdownSectionOrderer_PreservesDocumentLineEndings` is the missing gate, a theory over both `\n` and `\r\n`. It asserts the ordering still happened, that `DetectNewline` reports the input's ending, and that the LF count equals the requested-ending count so no stray bare LF survives.

**Tamper-verified:** reverting the rejoin to a hardcoded `'\n'` fails the `\r\n` case.

## Validation

Windows, Release, full suite `dotnet run --project src/dotnet-inspect.Tests -c Release`:

- **2363 total, 0 failed, 10 skipped** (ildasm/ilasm absent)

`LibraryCommand_DetailedOutput_RendersSectionsAlphabetically` now passes. The suite also reproduces the order-dependent flake tracked in #3416 — a first run showed `PerformanceTriagePredicates_DoNotAlterDiscovery` and `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceLinkDoor` failing, both of which pass in isolation and were clean on the second full run. That is #3416's documented shape and is not touched here.

### DIFF
--- a/src/dotnet-inspect.Tests/OutputFormatterTests.cs
+++ b/src/dotnet-inspect.Tests/OutputFormatterTests.cs
@@ -1302,7 +1302,35 @@ public class OutputFormatterTests
         Assert.True(output.IndexOf("## Beta", StringComparison.Ordinal) < output.IndexOf("## Alpha", StringComparison.Ordinal));
         Assert.True(output.IndexOf("## Alpha", StringComparison.Ordinal) < output.IndexOf("## Zebra", StringComparison.Ordinal));
         Assert.Contains("## Not a section", output);
-        Assert.Contains("B\n\n## Alpha", output);
+        // This claim is about which lines end up adjacent, so it is asserted against
+        // normalized text. Line endings are the separate claim below; a raw string
+        // literal carries whatever ending this source file is checked out with, so
+        // spelling "\n" here would silently assert the platform rather than the shape.
+        Assert.Contains("B\n\n## Alpha", output.ReplaceLineEndings("\n"));
+    }
+
+    /// <summary>
+    /// This is the gate for line-ending preservation in <c>MarkdownSectionOrderer</c>.
+    /// Reordering selects the order sections appear in; it is not licensed to rewrite CRLF
+    /// to LF, for the same reason row limiting is not — the same document would otherwise
+    /// differ byte for byte depending on whether a section order was supplied. The orderer
+    /// used to rejoin on a hardcoded '\n', which on Windows silently converted the whole
+    /// document and broke every caller that split it on <see cref="Environment.NewLine"/>.
+    /// </summary>
+    [Theory]
+    [InlineData("\n")]
+    [InlineData("\r\n")]
+    public void MarkdownSectionOrderer_PreservesDocumentLineEndings(string newline)
+    {
+        string markdown = string.Join(newline, ["# Title", "", "intro", "", "## Zebra", "", "Z", "", "## Alpha", "", "A"]);
+
+        var output = MarkdownSectionOrderer.Apply(markdown, ["Alpha", "Zebra"]);
+
+        Assert.True(output.IndexOf("## Alpha", StringComparison.Ordinal) < output.IndexOf("## Zebra", StringComparison.Ordinal));
+        Assert.Equal(newline, MarkdownScan.DetectNewline(output));
+        Assert.Equal(
+            output.Split('\n').Length - 1,
+            output.Split(newline).Length - 1);
     }
 
     [Fact]
--- a/src/dotnet-inspect/Output/MarkdownSectionOrderer.cs
+++ b/src/dotnet-inspect/Output/MarkdownSectionOrderer.cs
@@ -2,6 +2,9 @@ namespace DotnetInspector.Output;
 
 /// <summary>
 /// Reorders rendered Markdown H2 sections while preserving the document preamble.
+/// Reordering selects the order sections appear in; like row limiting, it is not
+/// licensed to rewrite the document's line endings, so it rejoins on the ending the
+/// text already used (see <see cref="MarkdownScan.DetectNewline"/>).
 /// </summary>
 internal static class MarkdownSectionOrderer
 {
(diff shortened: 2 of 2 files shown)


richlander:
## Review reconciliation

Two fixed-head reviews at `16bf6e8dd955f71eea4f3543df4ea37faff699c5`, from two different model families, both isolated in separate detached review worktrees.

### GPT — clean, high confidence

Verified the contract direction (preservation, not normalization) from `MarkdownScan.DetectNewline`'s own doc comment and both sibling passes; confirmed no `.gitattributes`, `.editorconfig`, golden output, or LF-only consumer depends on the old normalization; mapped the blast radius to section-ordered Markdown only (JSON, JSONL, TSV, plaintext, `--print`, `--value`, and shape projections all bypass the orderer, and `runfaster` consumes JSONL). Tamper-checked the new theory by evaluating the old LF join against the CRLF row: `expected=CRLF actual=LF` and `expected=10 actual=0` — the gate fails as intended.

A direct Windows probe of the shipped path reported `CRLF=374, bareLF=0` across 16 H2 sections, which is the user-visible defect in #3454 inverted.

The one sharp observation: the LF-count assertion is tautological on the LF theory row. True — the CRLF row is what carries the gate, and the theory is non-vacuous as a whole.

### MAI — one non-blocking finding, not actionable here

> Mixed newline styles are still rewritten by section ordering.

**Confirmed as described, and explicitly not fixed in this PR.** `DetectNewline` picks one style for the whole document, so a document mixing LF and CRLF is homogenized to CRLF.

That is the behavior of the shared heuristic, not something this change introduces:

| Pass | Detect | Normalize | Rejoin |
| --- | --- | --- | --- |
| `MarkdownTableRowLimiter` | `:11` | `:10` | `:59` |
| `OutputFormatter.LimitRenderedTableRows` | `:63` | `:64` | `:84` |
| `MarkdownSectionOrderer` (this PR) | `:16` | `:17` | `:60` |

All three are now the same three lines. On `main` the first two already homogenize mixed input exactly as MAI describes; the orderer was the *only* pass that hardcoded `'\n'` on rejoin, which is the bug #3454 reports. This PR makes the orderer consistent with its siblings — it does not change the mixed-input policy, and could not without changing `DetectNewline` for all three callers.

Changing that heuristic to per-line preservation is a cross-cutting change to a shared helper with three consumers and its own stated contract. It is out of scope here, and generated Markout input is homogeneous, so no shipped path reaches the mixed case. Recording it as a reasoned non-action rather than a silent one.

### Validation

- `dotnet-inspect.Tests`: **2363 passed, 0 failed** (Release, Windows), second run; the first run's 2 failures were both known #3416 order-dependent flakes that pass in isolation.
- CI green on all jobs including `test (ubuntu-26.04, linux-x64)`, which independently confirms the fix is not Windows-only in effect.

Ready to merge

PR #3460, opened 2026-07-29
Guard the type-forwarder path sink, and give the path-component rule one owner (#3441)

### LINKED ISSUE #3441: Untrusted AssemblyRef name reaches Path.Combine and PDB acquisition in PdbContext.ResolveImplementationAssemblyPath
## Summary

`PdbContext.ResolveImplementationAssemblyPath` builds a filesystem path from an
assembly-reference name read out of untrusted metadata, with no validation. The
resulting path is opened as an assembly and then fed into PDB acquisition,
including its network path.

Found during adversarial review of #3429, which fixed the same bug class in
`LibraryMetadataService.BuildTransitiveReferences`. This sink is independent of
that one and is not fixed by it.

## The sink

`src/ILInspector.Metadata/PdbContext.cs:790-801`

```csharp
public string? ResolveImplementationAssemblyPath(string typeName)
{
    var targetAssemblyName = FindTypeForwarder(typeName);   // reader.GetString(assemblyRef.Name)
    if (targetAssemblyName == null)
        return null;

    var dir = Path.GetDirectoryName(_assemblyPath);
    if (dir == null)
        return null;

    var targetPath = Path.Combine(dir, targetAssemblyName + ".dll");
    return File.Exists(targetPath) ? targetPath : null;
}
```

`FindTypeForwarder` returns `reader.GetString(assemblyRef.Name)` (line 778) — a
raw `#Strings` heap value from the inspected assembly, which is attacker
controlled. `docs/design/untrusted-data-threat-model.md` states under "Derived
paths" that `Path.Combine(root, untrustedValue)` is not a containment check.

## Impact

Reachable from the CLI. `SourceFileCollector.ResolveForwardedTypeSourceAsync`
(`src/dotnet-inspect/Inspectors/SourceFileCollector.cs:94`) consumes the result:

```csharp
var implementationPath = service.ResolveImplementationAssemblyPath(typeName);
...
implementationService = SourceLinkService.Open(implementationPath, logger.Log);
await SourceEnricher.AcquirePdbAsync(
    implementationService.Context, httpClient,
    packageName: null, packageVersion: null,
    isPlatformAssembly: true, logger.Log);
```

So an `ExportedType` whose implementation `AssemblyRef` is named
`../../some/path/target` causes the tool to read an assembly outside the
inspected assembly's directory, and then to run PDB acquisition against it —
attacker-influenced network egress and attacker-chosen local reads, from merely
inspecting a file.

A reviewer demonstrated the traversal resolving to an existing DLL outside the
assembly directory.

## Proposed fix

Validate the name as a single path component before combining it, and refuse to
resolve rather than sanitize — a trimmed or rewritten name silently designates a
different assembly. Emit the forwarder as unresolved rather than dropping it, so
the refusal does not hide evidence.

The validation should be **consolidated, not copied**. There are already three
near-copies of this rejection list:

- `src/DotnetInspector.Packages/NuGetCache.cs` (`ValidatePathComponent`)
- `src/DotnetInspector.Packages/Storage/FileSystemPackageStore.cs`
- `src/dotnet-inspect/Inspectors/LibraryMetadataService.cs` (`IsSafeAssemblySimpleName`, added by #3429)

They have already drifted: #3429's review found that the newest copy accepted
`CON `, `NUL   `, `Foo.`, `System.Text.Json `, ` System.Text.Json`, `.`, and the
superscript device aliases `COM¹`/`LPT¹`/`COM²`/`COM³`. Adding a fourth copy
guarantees the same drift again.

There is a layering constraint: `PdbContext` is in `ILInspector.Metadata`, which
cannot reference the CLI project where `IsSafeAssemblySimpleName` currently
lives. The shared predicate needs to move down to a layer all consumers can see
(`ILInspector.MetadataPrimitives` is a candidate — `ILInspector.Metadata` already
references it).

## Evidence requirements

Per `AGENTS.md`, the fix should carry a real-artifact canary that exercises the
resolution walk with a hostile forwarder name, plus close negatives so the guard
cannot grow into rejecting legitimate forwarded types. The hostile name set
should include every form the #3429 review turned up, so the consolidated
predicate starts at least as strong as the strongest current copy.

richlander:
Closes #3441.

## The defect

`PdbContext.ResolveImplementationAssemblyPath` took a type-forwarder target read from the inspected assembly's metadata — untrusted input — and passed it straight to `Path.Combine`/`File.Exists`:

```csharp
var targetAssemblyName = FindTypeForwarder(typeName);   // untrusted
var targetPath = Path.Combine(dir, targetAssemblyName + ".dll");
return File.Exists(targetPath) ? targetPath : null;
```

A forwarder naming `../payload` resolves a file outside the forwarding assembly's directory. This is the same bug as #3429's, and the **worse instance**: the resolved path feeds `SourceEnricher.AcquirePdbAsync`, so traversal drives symbol acquisition and its network fetches, not merely a local read.

## Why an owner, not a fourth guard

The rule already had three near-copies, and they had **measurably diverged**. `NuGetCache.ValidatePathComponent` — which guards package ids and versions on the way into the cache — was the weakest:

| | control chars | length bound | device names | host canonicalization |
| `NuGetCache` (before) | `\0` only | none | none | none |
| CLI `IsSafeAssemblySimpleName` (#3429) | all | 256 | yes, folded | yes |

So a package coordinate could name `CON` or `COM1`. Seam rule 6 in the inspection-layers note that just landed says a second implementation of a shared rule is a defect, and that a consumer which cannot reach the owner is a seam bug. This adds the owner: `HardenedPath`.

## What review found, and why that is the argument for the PR

The PR opened claiming one sink and three copies. Three review rounds each found
one more, and none of them found the same one:

| Round | Found | Kind |
| --- | --- | --- |
| 7 (MAI) | `SourceEnricher.TryEnrichFromForwardedAssemblyAsync` did its own `Path.Combine` on the forwarder target | **sink** |
| 7 (MAI) | `CONIN$` / `CONOUT$` match no device pattern and were absent | rule gap |
| 7 (GPT) | the digit fold walked UTF-16 **code units**, so supplementary-plane digits escaped | rule gap |
| 8 (MAI) | `AssemblyDependencyResolver.CollectPackageDependencies` combines a nuspec dependency id and version, **then recurses into** the result | **sink** |
| 8 (Gemini) | `ResourceExtractor.GetSafePathComponents` was a fifth **copy** — and the only one that knew `CLOCK$` | copy |
| 8 (Gemini) | `PlatformPackService.EnsurePackAsync` combines a pack coordinate | **sink** |
| 8 (Gemini) | the fold consumed a unit after **any** high surrogate without checking for a low one | real bug in my own code |

Final tally: **four sinks, five copies**. I found one of them. That a rule with
five copies keeps producing unguarded callers, and that three independent
reviewers each found a different one, is the case for an owner more directly
than the original defect was.

Two corrections I owe:

- `ResourceExtractor`'s copy uniquely knew `CLOCK$` and the Windows-invalid
  character set. Delegating would have **silently dropped both**; they moved
  *into* the owner first. Diff the copy against the owner before absorbing it.
- Gemini called the surrogate bug a device bypass. It is not — the lone
  surrogate stays in the stem, so no spelling becomes a device name. I only
  established that because the rejection test I wrote **failed**. The bug was
  real; the severity was not.

## Where it lives, and how I got that wrong first

I put `HardenedPath` in `DotnetInspector.Core` next to `HardenedJson` and `HardenedXml`, which is where it fits *by content*. `EngineToolBoundaryTests` rejected it: `Core` is tool tier, `ILInspector.Metadata` is engine, and the arrow points **tool → engine only**.

It now lives in `ILInspector.MetadataPrimitives`, an engine leaf both tiers can reach, keeping company with `SignatureBlobGuard` and `TypeSpecGuard` — the same untrusted-metadata guard family. The gate caught the mistake before review did, which is the argument for having it.

## Evidence

- **82 `HardenedPath` theories.** Every hostile form any former copy rejected, plus what #3429 learned — device-digit folding by **code point** (superscript, fullwidth, Arabic-Indic, supplementary-plane), format characters (Trojan Source), unpaired surrogates, and host canonicalization of trailing dots and spaces — each paired with close negatives (`COM1Plus`, `CONtoso.Library`, `Contoso.V⁴`, `CONIN$Extras`, real package versions).
- **7 forwarder-resolution tests** plant a real, readable payload exactly where `../payload` lands, with a legitimate sibling as the positive control.
- **2 seam tests** (`SourceEnricherForwarderSeamTests`) pin that the enricher reaches the owner. Their doc comment says plainly that they pin the *seam*, not the guard — a canary proving a refusal cannot prove a caller arrives, and that gap is exactly how the `SourceEnricher` sink escaped the first round.
- **3 theories** on `AssemblyDependencyResolver` prove a traversing dependency id cannot escape the package root.
- **Non-vacuity checked by tampering, for every guard.** With the forwarder guard removed, 4 of the 5 hostile spellings resolve the payload. (`..\payload` resolves only on Windows, where the backslash is a separator.)
- **Over-rejection measured, not asserted.** Two reviewers swept real inputs against the strengthened rule — 136 pack coordinates + 1954 assembly names, and 157 coordinates — with **zero** rejections. The fold fires only when the *whole stem* becomes a device name, so its cost on real data is nil.
- **Metadata 1078 pass / 5 skip** (was 998). **Services 346 pass**, exercising the strengthened `NuGetCache` and resolver against real package coordinates. **CLI 2363 pass / 14 skip.**

## Scope

`ResolveImplementationAssemblyPath` gained an internal `ResolveForwardedAssemblyPath` seam so the refusal is directly observable. `SourceEnricher` no longer takes a path at all — it takes the `PdbContext` and resolves through the owner, which is how a sink stops being possible rather than being fixed.

The CLI's `IsSafeAssemblySimpleName` is the remaining copy. It lives on #3429, which is still in review, so it is deliberately **not** folded in here — whichever lands second makes it delegate to `HardenedPath`.

Two fixed-head reviews are running against `cc27619a`. Not ready to merge.

### DIFF
--- a/src/DotnetInspector.Packages/DotnetInspector.Packages.csproj
+++ b/src/DotnetInspector.Packages/DotnetInspector.Packages.csproj
@@ -15,6 +15,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\DotnetInspector.Core\DotnetInspector.Core.csproj" />
+    <ProjectReference Include="..\ILInspector.MetadataPrimitives\ILInspector.MetadataPrimitives.csproj" />
     <PackageReference Include="NuGetFetch" />
     <PackageReference Include="NuGet.Versioning" />
   </ItemGroup>
--- a/src/DotnetInspector.Packages/DotnetInspector.Packages.csproj
+++ b/src/DotnetInspector.Packages/DotnetInspector.Packages.csproj
@@ -24,4 +24,10 @@
     <PackageReference Include="System.Security.Cryptography.ProtectedData" />
   </ItemGroup>
 
+  <ItemGroup>
+    <!-- StorePath is internal; StorePathSegmentTests gates it against drifting away from
+         HardenedPath, the single owner of the path-component rule. -->
+    <InternalsVisibleTo Include="DotnetInspector.Services.Tests" />
+  </ItemGroup>
+
 </Project>
--- a/src/DotnetInspector.Packages/NuGetCache.cs
+++ b/src/DotnetInspector.Packages/NuGetCache.cs
@@ -1,3 +1,4 @@
+using ILInspector.MetadataPrimitives;
 using DotnetInspector.Core;
 using NuGet.Versioning;
 
@@ -47,25 +48,14 @@ public static class NuGetCache
         ?? throw new InvalidOperationException("NuGetCache.Initialize(appName) must be called before using app cache methods.");
 
     /// <summary>
-    /// Validates that a value is safe to use as a path component. Rejects empty
-    /// or whitespace values, traversal (<c>..</c>), separators, volume
-    /// qualifiers (<c>:</c>), null characters, and otherwise rooted values, so an
-    /// attacker-influenced package coordinate cannot escape or reset the cache
-    /// root (a legitimate package id or version contains none of these).
+    /// Validates that a value is safe to use as a path component, so an attacker-influenced
+    /// package coordinate cannot escape or reset the cache root. Delegates to
+    /// <see cref="HardenedPath.ValidatePathComponent"/>, which is the single owner of this rule;
+    /// this method's own copy of it had drifted (it accepted reserved device names, every control
+    /// character other than <c>\0</c>, and names the host rewrites before opening).
     /// </summary>
     internal static void ValidatePathComponent(string value, string name)
-    {
-        if (string.IsNullOrWhiteSpace(value)
-            || value.Contains("..")
-            || value.Contains('/')
-            || value.Contains('\\')
-            || value.Contains(':')
-            || value.Contains('\0')
-            || Path.IsPathRooted(value))
-        {
-            throw new ArgumentException($"Invalid {name}: '{value}'");
-        }
-    }
+        => HardenedPath.ValidatePathComponent(value, name);
 
     /// <summary>
     /// Gets the path to the NuGet package cache (read-only).
--- a/src/DotnetInspector.Packages/NuGetCache.cs
+++ b/src/DotnetInspector.Packages/NuGetCache.cs
@@ -120,10 +120,18 @@ public static class NuGetCache
     /// <param name="packageName">The package name (case-insensitive)</param>
     /// <param name="version">The package version</param>
     /// <returns>The path to the cached package directory, or null if not found</returns>
+    /// <remarks>
+    /// An unsafe coordinate answers null rather than throwing. No cache entry can be named by one,
+    /// so "not found" is the truthful answer, and it is the answer this method's contract promises;
+    /// throwing from a <c>Try</c> method crashed the CLI on <c>package CON@1.0.0 --version</c> and
+    /// made <see cref="PackageExtractor"/>'s nuspec probe report a hostile coordinate as an error
+    /// rather than refusing it. The refusal stays visible where it matters: every path that
+    /// acquires or extracts a package still refuses the coordinate loudly.
+    /// </remarks>
     public static string? TryGetCachedPackage(string packageName, string version)
     {
-        ValidatePathComponent(packageName, "package name");
-        ValidatePathComponent(version, "version");
+        if (!HardenedPath.IsSafePathComponent(packageName) || !HardenedPath.IsSafePathComponent(version))
+            return null;
 
         var normalizedName = packageName.ToLowerInvariant();
         var normalizedVersion = version.ToLowerInvariant();
--- a/src/DotnetInspector.Packages/NuGetCache.cs
+++ b/src/DotnetInspector.Packages/NuGetCache.cs
@@ -323,8 +323,18 @@ public static class NuGetCache
     /// Returns the newest cached version of a package from the NuGet or app cache.
     /// Pure disk I/O — never hits the network.
     /// </summary>
+    /// <remarks>
+    /// Returns <see langword="null"/> for a name that is not safe as a path component, rather than
+    /// combining it into a cache root. The guard lives here, not in the callers: six call sites
+    /// reach this method, and an earlier fix in this series guarded one route while the identical
+    /// hole survived on a sibling route with a green suite. Callers that surface the result to a
+    /// user should distinguish a refusal from a miss.
+    /// </remarks>
     public static string? TryGetLatestCachedVersion(string packageName)
     {
+        if (!HardenedPath.IsSafePathComponent(packageName))
+            return null;
+
         var normalizedName = packageName.ToLowerInvariant();
 
         // Newest non-prerelease, structurally-valid version across both caches.
--- a/src/DotnetInspector.Packages/PackageExtractor.cs
+++ b/src/DotnetInspector.Packages/PackageExtractor.cs
@@ -697,6 +697,22 @@ public static class PackageExtractor
         string normalizedName = packageName.ToLowerInvariant();
         string cacheKey = includePrerelease ? $"{normalizedName}-prerelease" : normalizedName;
 
+        // A name that cannot be a path component can never name a package we could acquire: the
+        // coordinate becomes a cache directory further down every route that consumes this
+        // version. Refuse here rather than at the callers -- PackageCommand's --versions branch
+        // already refused, so `package CON --version` was clean while `package CON` reached the
+        // cache and the network before reporting a miss. Guarding the method makes the refusal
+        // early, network-free and identical on every route into it, instead of a property of
+        // which branch the caller took.
+        if (!HardenedPath.IsSafePathComponent(normalizedName))
+        {
+            // Unconditional, matching CoreCache's cache-escape warning: a verbose-gated log would
+            // make the refusal invisible in the default view, and the caller's "not found" reads
+            // as "searched for and absent" when the name was never searched for at all.
+            Console.Error.WriteLine($"Warning: invalid package name, not searched for: '{packageName}'.");
+            return null;
+        }
+
         // Cache nuget.org results even when additional custom sources are configured.
         bool canCache = !skipCache && sources.Any(s => s.IsNuGetOrg);
 
--- a/src/DotnetInspector.Packages/PackageExtractor.cs
+++ b/src/DotnetInspector.Packages/PackageExtractor.cs
@@ -500,7 +500,7 @@ public static class PackageExtractor
         if (!HardenedPath.IsSafePathComponent(normalizedName)
             || !HardenedPath.IsSafePathComponent(normalizedVersion))
         {
-            log?.Invoke($"Refusing unsafe package coordinate '{packageId}.{version}'");
+            HardenedPath.ReportRefusal($"refusing unsafe package coordinate '{packageId}.{version}'");
             return null;
         }
 
--- a/src/DotnetInspector.Packages/PackageExtractor.cs
+++ b/src/DotnetInspector.Packages/PackageExtractor.cs
@@ -492,6 +492,18 @@ public static class PackageExtractor
         string normalizedName = packageId.ToLowerInvariant();
         string normalizedVersion = version.ToLowerInvariant();
 
+        // A dependency id comes from another package's nuspec, so it is untrusted. Refusing it has
+        // to say so: TryGetCachedPackage answers null for an unsafe coordinate, and without this
+        // the refusal became indistinguishable from an ordinary cache miss -- the dependency
+        // simply vanished from the resolved tree. Returning null is still right for the caller;
+        // being quiet about why is not.
+        if (!HardenedPath.IsSafePathComponent(normalizedName)
+            || !HardenedPath.IsSafePathComponent(normalizedVersion))
+        {
+            log?.Invoke($"Refusing unsafe package coordinate '{packageId}.{version}'");
+            return null;
+        }
+
         // Cache hit: read the nuspec straight from the already-extracted package.
         var cachedPath = NuGetCache.TryGetCachedPackage(normalizedName, normalizedVersion);
         if (cachedPath != null && NuGetCache.IsCachedPackageValid(cachedPath))
--- a/src/DotnetInspector.Packages/PackageExtractor.cs
+++ b/src/DotnetInspector.Packages/PackageExtractor.cs
@@ -4,6 +4,7 @@
 using System.IO.Compression;
 using System.Net.Http.Headers;
 using DotnetInspector.Core;
+using ILInspector.MetadataPrimitives;
 using NuGetFetch;
 using NuGetSource = NuGetFetch.PackageSource;
 
@@ -232,6 +233,20 @@ public static class PackageExtractor
         // Normalize to lowercase for NuGet API
         string normalizedName = packageName.ToLowerInvariant();
         string normalizedVersion = version.ToLowerInvariant();
+
+        // The coordinate becomes a cache directory name, so it has to be a safe path component.
+        // NuGetCache validates it too, but by throwing, which leaves the command with an unhandled
+        // ArgumentException and a stack trace. That throw used to be unreachable for an ordinary
+        // name: the local copy of the rule this PR deleted could only reject separators, traversal
+        // and volume qualifiers, all of which the argument parser rejects earlier. Tightening the
+        // rule made it reachable -- `package CON@1.0.0` printed 1846 bytes of stack trace where it
+        // previously printed `Error: Package 'con' not found.` -- so refuse it here as an ordinary
+        // acquisition failure instead.
+        if (!HardenedPath.IsSafePathComponent(normalizedName))
+            return PackageExtractionOutcome.Error($"Invalid package name: '{packageName}'.");
+
+        if (!HardenedPath.IsSafePathComponent(normalizedVersion))
+            return PackageExtractionOutcome.Error($"Invalid package version: '{version}'.");
 
         var request = new PackageAcquisitionRequest(
             Path.GetFullPath(
--- a/src/DotnetInspector.Packages/Storage/StorePath.cs
+++ b/src/DotnetInspector.Packages/Storage/StorePath.cs
@@ -1,3 +1,5 @@
+using ILInspector.MetadataPrimitives;
+
 namespace DotnetInspector.Packages;
 
 /// <summary>
@@ -46,17 +48,17 @@ internal static class StorePath
     }
 
     /// <summary>
-    /// True when <paramref name="segment"/> is a safe single path component:
-    /// non-empty, not <c>.</c> or <c>..</c>, containing no separator, volume
-    /// (<c>:</c>), or null character, and not otherwise rooted.
+    /// True when <paramref name="segment"/> is a safe single path component.
     /// </summary>
+    /// <remarks>
+    /// Delegates to <see cref="HardenedPath.IsSafePathComponent"/>, the single owner of this rule.
+    /// This was previously a separate implementation testing only for empty, <c>.</c>, <c>..</c>,
+    /// separators, <c>:</c>, NUL and rooted values, so it accepted reserved device names, other
+    /// control characters, invisible format characters, unpaired surrogates, trailing dots and
+    /// spaces, and overlong components. An untrusted CodeView PDB file name reaches this rule on
+    /// its way to filesystem-backed symbol storage (<c>SymbolPackageDownloader</c>), so those gaps
+    /// were reachable: <c>CON</c> was accepted here and refused by every other copy of the rule.
+    /// </remarks>
     public static bool IsSafeSegment(string segment)
-        => segment.Length != 0
-            && segment != "."
-            && segment != ".."
-            && !segment.Contains('/')
-            && !segment.Contains('\\')
-            && !segment.Contains(':')
-            && !segment.Contains('\0')
-            && !Path.IsPathRooted(segment);
+        => HardenedPath.IsSafePathComponent(segment);
 }
--- a/src/DotnetInspector.Services.Tests/AssemblyDependencyResolverTests.cs
+++ b/src/DotnetInspector.Services.Tests/AssemblyDependencyResolverTests.cs
@@ -371,10 +371,35 @@ public class AssemblyDependencyResolverTests
     /// id lands: on unguarded code the resolver escapes the package root, recurses into that
     /// directory, and returns its assemblies as the package's references.
     /// </summary>
+    /// <remarks>
+    /// Which spellings actually traverse is a property of the host. <c>..\escape</c> is a traversal
+    /// on Windows and an ordinary directory name on Unix, so on this machine that case exercises
+    /// the guard's literal-name branch and proves nothing about traversal.
+    /// <see cref="HostileDependencyIds_ContainACaseThatTraversesOnThisHost"/> is what stops the
+    /// whole theory degrading to literal names, which is how it would pass while the traversal
+    /// guard was gone.
+    /// </remarks>
+    private static readonly string[] HostileIds = ["../escape", "..\\escape", "CON"];
+
+    public static TheoryData<string> HostileDependencyIds => [.. HostileIds];
+
+    /// <summary>
+    /// Non-vacuity gate for the theory below: at least one hostile id must use a separator this
+    /// host recognises, or every case is a literal directory name and the theory asserts nothing
+    /// about traversal. It reads the same data the theory does, so removing the traversing
+    /// spelling fails here rather than silently weakening the theory.
+    /// </summary>
+    [Fact]
+    public void HostileDependencyIds_ContainACaseThatTraversesOnThisHost()
+    {
+        Assert.Contains(
+            HostileIds,
+            id => id.Contains(Path.DirectorySeparatorChar)
+                || id.Contains(Path.AltDirectorySeparatorChar));
+    }
+
     [Theory]
-    [InlineData("../escape")]
-    [InlineData("..\\escape")]
-    [InlineData("CON")]
+    [MemberData(nameof(HostileDependencyIds))]
     public void PackageDependencyReferencePaths_WithTraversingDependencyId_DoesNotEscapePackageRoot(string hostileId)
     {
         string sandbox = Directory.CreateTempSubdirectory("dotnet-inspect-deps-traversal-").FullName;
--- a/src/DotnetInspector.Services.Tests/AssemblyDependencyResolverTests.cs
+++ b/src/DotnetInspector.Services.Tests/AssemblyDependencyResolverTests.cs
@@ -400,7 +400,7 @@ public class AssemblyDependencyResolverTests
 
     [Theory]
     [MemberData(nameof(HostileDependencyIds))]
-    public void PackageDependencyReferencePaths_WithTraversingDependencyId_DoesNotEscapePackageRoot(string hostileId)
+    public async Task PackageDependencyReferencePaths_WithTraversingDependencyId_DoesNotEscapePackageRoot(string hostileId)
     {
         string sandbox = Directory.CreateTempSubdirectory("dotnet-inspect-deps-traversal-").FullName;
         try
@@ -439,12 +439,29 @@ public class AssemblyDependencyResolverTests
             // Positive control: the guard must refuse traversal specifically, not stop resolving.
             string legit = CreatePackageAsset(root, "Legit.Package", "1.0.0", "lib", "net8.0", "Legit.dll");
 
-            var references = AssemblyDependencyResolver.PackageDependencyReferencePaths(targetPath, [root]);
+            IReadOnlyList<string> references = [];
+            var stderr = await StderrCapture.RunAsync(() =>
+                references = AssemblyDependencyResolver.PackageDependencyReferencePaths(targetPath, [root]));
 
-            Assert.Contains(legit, references);
-            Assert.DoesNotContain(payload, references);
-            Assert.DoesNotContain(literal, references);
-            Assert.All(references, path => Assert.StartsWith(root + Path.DirectorySeparatorChar, path, StringComparison.Ordinal));
+            // The refusal has to reach the user. It was reported through the optional
+            // Action<string>? log, and no caller in the product passes one, so the message went
+            // nowhere: the dependency simply vanished from the result and the run exited 0. A
+            // refused coordinate is indistinguishable from an absent one unless it is announced.
+            Assert.Contains("refusing unsafe package coordinate", stderr, StringComparison.Ordinal);
+            Assert.Contains(hostileId, stderr, StringComparison.Ordinal);
+
+            // Canonicalise before comparing. Path.Combine does not normalise, so an escaping id
+            // yields "<root>/../escape/lib/net8.0/Payload.dll" -- a string that both StartsWith
+            // "<root>/" and differs character-for-character from the payload's own path. Asserting
+            // on the raw strings therefore passed with the guard entirely removed: only the "CON"
+            // row failed, and the traversal rows, the whole point of the theory, proved nothing.
+            var resolved = references.Select(Path.GetFullPath).ToList();
+            var rootPrefix = Path.GetFullPath(root) + Path.DirectorySeparatorChar;
+
+            Assert.Contains(Path.GetFullPath(legit), resolved);
+            Assert.DoesNotContain(Path.GetFullPath(payload), resolved);
+            Assert.DoesNotContain(Path.GetFullPath(literal), resolved);
+            Assert.All(resolved, path => Assert.StartsWith(rootPrefix, path, StringComparison.Ordinal));
         }
         finally
         {
--- a/src/DotnetInspector.Services.Tests/AssemblyDependencyResolverTests.cs
+++ b/src/DotnetInspector.Services.Tests/AssemblyDependencyResolverTests.cs
@@ -364,6 +364,69 @@ public class AssemblyDependencyResolverTests
         }
     }
 
+    /// <summary>
+    /// Artifact canary for the nuspec dependency-coordinate sink. A dependency id and version are
+    /// read from a nuspec, which is a feed artifact rather than something this process authored,
+    /// and both become path components. This plants a readable payload exactly where a traversing
+    /// id lands: on unguarded code the resolver escapes the package root, recurses into that
+    /// directory, and returns its assemblies as the package's references.
+    /// </summary>
+    [Theory]
+    [InlineData("../escape")]
+    [InlineData("..\\escape")]
+    [InlineData("CON")]
+    public void PackageDependencyReferencePaths_WithTraversingDependencyId_DoesNotEscapePackageRoot(string hostileId)
+    {
+        string sandbox = Directory.CreateTempSubdirectory("dotnet-inspect-deps-traversal-").FullName;
+        try
+        {
+            string root = Path.Combine(sandbox, "packages");
+            Directory.CreateDirectory(root);
+
+            string packageDir = Path.Combine(root, "root.package", "1.0.0");
+            string targetDir = Path.Combine(packageDir, "lib", "net8.0");
+            Directory.CreateDirectory(targetDir);
+            string targetPath = Path.Combine(targetDir, "Root.dll");
+            File.WriteAllText(targetPath, "");
+            File.WriteAllText(
+                Path.Combine(packageDir, "root.package.nuspec"),
+                $"""
+                <?xml version="1.0" encoding="utf-8"?>
+                <package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
+                  <metadata>
+                    <id>Root.Package</id>
+                    <version>1.0.0</version>
+                    <dependencies>
+                      <group targetFramework="net8.0">
+                        <dependency id="{System.Security.SecurityElement.Escape(hostileId)}" version="1.0.0" />
+                        <dependency id="Legit.Package" version="1.0.0" />
+                      </group>
+                    </dependencies>
+                  </metadata>
+                </package>
+                """);
+
+            // The payload sits one level above the package root, where "../escape" lands, and
+            // also at the literal name inside it, so a refusal cannot be mistaken for absence.
+            string payload = CreatePackageAsset(sandbox, "escape", "1.0.0", "lib", "net8.0", "Payload.dll");
+            string literal = CreatePackageAsset(root, hostileId.Replace("..", "dots"), "1.0.0", "lib", "net8.0", "Payload.dll");
+
+            // Positive control: the guard must refuse traversal specifically, not stop resolving.
+            string legit = CreatePackageAsset(root, "Legit.Package", "1.0.0", "lib", "net8.0", "Legit.dll");
+
+            var references = AssemblyDependencyResolver.PackageDependencyReferencePaths(targetPath, [root]);
+
+            Assert.Contains(legit, references);
+            Assert.DoesNotContain(payload, references);
+            Assert.DoesNotContain(literal, references);
+            Assert.All(references, path => Assert.StartsWith(root + Path.DirectorySeparatorChar, path, StringComparison.Ordinal));
+        }
+        finally
+        {
+            Directory.Delete(sandbox, recursive: true);
+        }
+    }
+
     static string CreatePackageAsset(string root, string id, string version, string assetKind, string tfm, string fileName)
     {
         string assetDir = Path.Combine(root, id.ToLowerInvariant(), version.ToLowerInvariant(), assetKind, tfm);
--- a/src/DotnetInspector.Services.Tests/DependencyRefusalVisibilityTests.cs
+++ b/src/DotnetInspector.Services.Tests/DependencyRefusalVisibilityTests.cs
@@ -21,7 +21,7 @@ public class DependencyRefusalVisibilityTests
     /// asset in the same file is the control: it must still be added.
     /// </summary>
     [Fact]
-    public void UnsafeDepsJsonAssetPath_IsRefusedAndReported()
+    public async Task UnsafeDepsJsonAssetPath_IsRefusedAndReported()
     {
         var root = Directory.CreateTempSubdirectory("di-refusal-deps-");
         try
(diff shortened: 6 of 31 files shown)


richlander:
## Review round 7 — MAI, reconciled at `6a47b7cf`

Both findings were real and both are fixed. Reviewed head was `ee7e3e06`.

**1. A third unguarded forwarder sink — `SourceEnricher` (actioned).**
`TryEnrichFromForwardedAssemblyAsync` took the forwarding assembly's path and
did its own `Path.Combine` on the forwarder target. Verified from source: the
target is a name read from untrusted metadata, and this is the *live* path —
resolution here feeds source enrichment and its network fetches, which is
precisely the egress #3441 described.

This is not a finding adjacent to the PR; it is the PR's own defect class.
Seam rule 6 in `docs/design/inspection-layers.md` — "a second implementation of
a shared rule is a defect" — is the warrant for this change, and I had counted
copies without finding this one. The method now resolves through
`PdbContext.ResolveImplementationAssemblyPath` and is handed no path to
re-derive one from.

The refusal is owned and proved by `PdbContextForwarderResolutionTests`. What
that canary cannot see is whether a caller *reaches* the owner — which is
exactly how this sink escaped it — so `SourceEnricherForwarderSeamTests` pins
the shape the fix depends on. Non-vacuity verified by tampering: reintroducing
a path parameter fails it. The test's doc comment is explicit that it pins the
seam and not the guard, so it is not mistaken for a second proof of refusal.

**2. `CONIN$` / `CONOUT$` missing from `ReservedDeviceNames` (actioned).**
Correct: Windows reserves both, and neither follows the `CON`/`COMn`/`LPTn`
patterns the list otherwise covers. Added, with theories for `CONIN$`,
`conout$`, and `CONIN$.dll`, plus `CONIN$Extras` as the close negative so the
addition cannot quietly widen into a prefix match.

**Evidence at `6a47b7cf`:** Metadata 1072 pass / 5 skip; CLI 2361 pass / 14
skip. The CLI suite matters here specifically because `SourceEnricher` sits on
the source-enrichment path.

A second review (`r7-gpt`) is still running against `ee7e3e06`. Not ready to
merge: this head is unreviewed, and I will re-review the fixed head before
claiming otherwise.

richlander:
## Review round 7 — GPT, reconciled at `582bf39a`

Reviewed head was `ee7e3e06`. Two of its findings had already been actioned
from the MAI review in the comment above; one sub-claim was new and is now
fixed.

**1. `SourceEnricher` bypasses the guard (confirmed, already fixed).** GPT found
the third sink independently, with the same reading: `FindTypeForwarder`'s
result reaches `Path.Combine` directly, so `../payload` escapes the assembly
directory. Two reviewers converging on it is worth noting — I had counted
copies of the *rejection list* rather than the *sinks*, which is why my own
survey missed it. Fixed at `6a47b7cf`.

**2. `CONIN$`/`CONOUT$` not rejected (confirmed, already fixed).** GPT correctly
declined to call it exploitable — the forwarder path appends `.dll` and NuGet
ids cannot contain `$`. It is fixed anyway, because the list claims to be the
reserved set.

**3. The digit fold walked UTF-16 code units (new — actioned at `582bf39a`).**
`char.GetNumericValue` on a high surrogate returns `-1`, so the mathematical
and enclosed digit blocks passed through unfolded. Measured directly:

```text
code-unit loop (old): -1
code-point loop (new): 1     // for "COM\U0001d7cf"
```

I am *not* claiming Windows aliases those to a device, and GPT did not claim it
either. The reason to fix it is narrower: the rule is written as a property of
the code point, and a code-unit loop made that statement false for part of its
own domain — and enumerating which encodings to fold is exactly what drifted
twice in earlier rounds. `AsciiDigitFor` now indexes the string.
Non-vacuity is the measurement above; theories added for `COM𝟏`/`LPT𝟣` as
rejections and `COM𝟏Plus`, `Contoso.𝐀ssembly`, `😀.Assembly` as close negatives
(a folded surrogate pair shortens the stem, so a non-digit pair must survive).

**Independent over-rejection evidence.** GPT scanned the real local NuGet cache
and 20 `project.assets.json` files: 95 unique coordinates, 8 prerelease, 18
mixed-case, **0 rejected**. That is a stronger measurement of the fold's cost
than I had, and it confirms the round-6 correction — the cost I once asserted
without measuring is in fact zero.

**Non-vacuity was not reached by this reviewer** (source modification was
prohibited in the review worktree). It is covered by my own tamper runs,
recorded in the PR body and in the round-7 comment above.

**Evidence at `582bf39a`:** Metadata 1077 pass / 5 skip; Services 343 pass; CLI
2361 pass / 14 skip at `6a47b7cf` (the fold commit does not touch CLI code).

Not ready to merge: this head is unreviewed. Next step is a fixed-head review
of `582bf39a` from two families.

richlander:
## Review round 8 — MAI + Gemini, reconciled at `cc27619a`

Both reviewed `582bf39a`. Five findings between them; all five actioned.

### MAI — a fourth sink: `AssemblyDependencyResolver` (High, actioned `59332208`)

`CollectPackageDependencies` reads a dependency id and version out of a nuspec
and combines both into a path with no validation. Reproduced from source: an id
of `../escape` resolves outside the package root, and the resolver then
*recurses* into that directory looking for its nuspec and records it as the
package's location, so the escape propagates into what is later read as an
assembly.

New canary plants a readable payload one level above the package root, where
`../escape` lands, plus a legitimate sibling dependency as the positive control
so a refusal cannot pass by breaking resolution generally. Non-vacuity by
tampering: with the guard deleted, `Assert.DoesNotContain() Failure: Item found
in collection`.

MAI's over-rejection scan was the largest yet — 136 coordinates and 1954
assembly names, 0 rejections.

### Gemini — a fifth *copy*, two more sinks, and a real bug in my fold

**`ResourceExtractor.GetSafePathComponents` (actioned).** A fifth
implementation, and the best of them — it was the only one that knew about
`CLOCK$` — but missing the digit fold, the format-character refusal, and the
length bound. It now delegates; `CLOCK$` and the Windows-invalid filename set
moved to the owner. Only the genuinely path-level checks stay behind (the
split, rooted, drive-qualified). This is seam rule 6 working as intended: the
merge strictly strengthened three call sites and strictly strengthened the
owner.

**`PlatformPackService.EnsurePackAsync` (actioned).** Pack name and version
become a cache directory this process extracts a downloaded package into, and
the version normally comes from `GetLatestPackVersionAsync` — a feed response.
Guarded.

**`LibraryMetadataService.BuildTransitiveReferences` (confirmed, not actioned
here).** Real on `main`, and already fixed on #3429 by
`IsSafeAssemblySimpleName`. That copy is the one to make delegate once both
land; noted as follow-up rather than duplicated into this PR.

**The fold's surrogate handling (actioned) — with a correction.** Gemini is
right that the loop consumed the code unit after *any* high surrogate without
checking it was a low surrogate, so in `COM\ud800\u00b9` the digit was copied
out untested. It read this as a device bypass. **It is not**, and I only found
that out by writing the test: the unpaired surrogate stays in the stem either
way, so no spelling becomes a device name, and my first rejection theory failed
for exactly that reason. The bug is real as a correctness defect in a guard;
the fix that removes the question entirely is to refuse malformed UTF-16
outright, which costs nothing since no real coordinate contains one.

That test is a `[Fact]`, not a `[Theory]`: xUnit's `InlineData` round-trip
replaces an unpaired surrogate with U+FFFD, so the theory version was testing a
well-formed string and would have passed for the wrong reason. Worth flagging
for anyone else pinning malformed UTF-16 here.

### On the seam pin

Both reviewers landed in the same place, and I agree with them: it gates the
wiring, not the behavior, and a caller could accept a `PdbContext` and still
build its own path. Its doc comment says so, which is what `AGENTS.md` asks
for, but "adequate" is doing real work in that sentence. Left as is and named
plainly rather than upgraded.

### The pattern is the argument

Rounds 6, 7 and 8 each found one more sink or one more copy — four sinks and
five copies now. That is not a sign the PR keeps breaking; it is the measurement
of what a rule with no owner costs. Each was findable only by looking, which is
precisely why they survived.

**Evidence at `cc27619a`:** Metadata 1078 pass / 5 skip; Services 346 pass; CLI
2363 pass / 14 skip.

Not ready to merge: this head is unreviewed.

richlander:
## Round 9 reconciliation — `cc27619a` → `d865ffdb`

Reviewers: MAI-Code and Gemini Pro, both on the fixed head in isolated worktrees.

### MAI-Code — no significant issues

Independently re-derived several things I had claimed:

- Swept Services/Packages/Metadata/Inspectors for path-building sites reachable
  from untrusted names and found **no unguarded sink** (it did not look at
  `deps.json`/`project.assets.json` parsing, which is where Gemini did).
- **4,059** real coordinates through the current rule, **0 rejections**.
- Reproduced the xUnit surrogate round-trip independently, printing
  `0043 004F 004D FFFD FFFD FFFD 00B9` — confirming `[InlineData]` rewrites an
  unpaired surrogate to `U+FFFD` and that `UnpairedSurrogate_IsRejected` has to
  be a `[Fact]`.
- Re-proved non-vacuity by deleting the guard and watching the hostile
  coordinates resolve.

### Gemini Pro — a fifth kind of sink. Actioned in `d865ffdb`.

**Finding:** `AssemblyDependencyResolver.cs:730,733` and
`ProjectAssetsParser.cs:201,497` combine names read from `.deps.json` and
`project.assets.json` with a trusted root and read the result as an assembly,
with no guard.

**Verified, and the repo had already written this down.**
`docs/design/untrusted-data-threat-model.md:37` lists
`project.assets.json`, `.deps.json`, and *"paths within those files"* as
untrusted input to the project and dependency resolvers, with **"path confusion,
unintended file reads"** as the named risk. So this is not a severity judgement
call — the finding lands on a surface the threat model already classified, and I
had not covered it.

**It is exploitable.** `Parse_WithTraversingLibraryPath_DoesNotEscapeTheNuGetCache`
computes the traversal from the real cache root to a planted readable payload
with `Path.GetRelativePath`, so the payload is genuinely reachable. With the
guard removed:

```text
Assert.DoesNotContain() Failure: Filter matched in collection
Collection: [Tuple ("/Users/rich/.nuget/packages/../../../../var/folder"…,
            "Evil.Package", "1.0.0")]
```

**The fix is a new rule on the owner, not a sixth copy.** These carry
multi-segment relative paths (`lib/net8.0/Foo.dll`), so `IsSafePathComponent`
would refuse every legitimate value for containing a separator.
`HardenedPath.IsSafeRelativePath` applies the component rule per segment, treats
both separators as separators on every host, rejects rooted values, and rejects
empty/`.` segments so a path cannot normalize past the rule.

**My first measurement of over-rejection was wrong, and I am keeping the
record.** Sweeping 1,889 `.deps.json` and 1,568 `project.assets.json` files, the
new rule rejected **5,249** paths. Every one was a `type: "project"` library,
whose `path` is legitimately `../Contoso/Contoso.csproj`. The parser already
skips those *before* reading `path`, so the product never sees them — my probe
was broader than the code. Matching the probe to actual behavior:
**268,005 paths scanned, 0 rejected.**

That skip is now load-bearing for the guard, so it carries a comment saying so
with both numbers. Moving the guard above that line would break ordinary project
references, and the next person deserves to know that before finding out.

### Gemini's non-vacuity observation — correct, and it sharpens an earlier claim

Gemini established that the `..\payload` forwarder case **passes vacuously on
macOS**: `\` is not a separator there, the literal file is never created, and it
resolves NULL guarded or not. Only 4 of the 5 hostile spellings actually prove
that guard on this platform. My PR body already said "4 of the 5", but the prose
was carrying a fact the test should state; that is a fair hit on where the claim
lives.

### State

**Head `d865ffdb`.** Metadata **1119 pass / 5 skip** (was 1078).
Services **348 pass** (was 346). CLI **2363 pass / 14 skip**.

Running tally across rounds 6–9: **five sinks, five copies**, found by four
different reviewers, none of whom found the same one. I found one.

Not ready to merge — this head has not been reviewed.

richlander:
## Round 10 — two families, two *different* missed sinks. Both real. `d865ffdb` → `f92ee885`

Rounds 6, 7, 8 and 9 each found one more sink or copy, so I said before this
round that I expected a sixth. There were two, plus the last divergent copy.

### GPT — High: `ResolvePackagePath` honored a rooted `library.path`

```csharp
return Path.IsPathRooted(normalized)
    ? Path.GetFullPath(normalized)      // attacker names the directory outright
    : Path.GetFullPath(Path.Combine(NuGetCache.GetNuGetCachePath(), normalized));
```

No traversal required. `ParsePackageFileEntries` then reads README, AGENTS and
skill content from that directory. My earlier guards covered `Parse` and
`ResolvePackageFilePath` — **different code paths**. Fixed.

Both callers reject project libraries via `IsProjectLibrary` *before* reaching
it, which is what makes refusing relative traversal correct here and unable to
refuse an ordinary `../Contoso/Contoso.csproj`. That is the same precondition as
the `Parse` guard, but enforced by a named check at each call site rather than a
loop `continue` — which answers Gemini's round-10 point 4 about that skip being
structurally fragile.

### Gemini — High: untrusted `AssemblyRef.Name` reaches the framework directory

`BuildTransitiveReferences` forwards a name read straight off the inspected PE
into `PlatformResolver.ResolveAssembly`, which ends at:

```csharp
var exactPath = Path.Combine(directory, assemblyFileName);
if (File.Exists(exactPath)) return exactPath;
```

Gemini's probe produced the escape verbatim:

```text
/usr/local/share/dotnet/shared/Microsoft.NETCore.App/11.0.0-preview.6.26359.118/../../../…/tmp/foo.dll
```

**I did not take the suggested fix.** Gemini proposed guarding
`BuildTransitiveReferences`. `ResolveAssembly` has **twelve call sites across
five files**, and guarding one caller is precisely how this rule became six
divergent copies. I checked every caller passes a simple assembly name and never
a path, then guarded the chokepoint. Twelve sites covered by one rule.

Over-rejection swept at head: **30,454 real assemblies, 736 distinct simple
names, 0 rejected.**

### GPT — Medium: the sixth divergent copy

`StorePath.IsSafeSegment` was an independent implementation. `CON` was accepted
here and refused by every other copy, and an untrusted CodeView PDB file name
reaches it through `SymbolPackageDownloader`. It now delegates to `HardenedPath`.

The gate asserts the two **agree on every probe** rather than restating the rule,
so a reintroduced local copy fails the moment it disagrees on anything:

```csharp
var disagreements = probes
    .Where(p => StorePath.IsSafeSegment(p) != HardenedPath.IsSafePathComponent(p))
    .ToArray();
Assert.Empty(disagreements);
```

### GPT — Medium: success-shaped output. Confirmed, and it was worse than described.

Refusing a package path did not just omit a dependency — `ParsePackageFileEntries`
still emitted the entry:

```text
ProjectPackageFileEntry { PackageName = Evil.Package, Path = skills/SKILL.md,
                          PackagePath = , FullPath = , … }
```

A skill reported as present that resolves nowhere. It now skips the package and
logs why, and the test asserts **the warning**, not merely the absence.

The resolver's nuspec and deps.json skips have no log channel to use; threading
one changes the resolver's shape and its callers. Filed as **#3477** with the
threat-model and `AGENTS.md` citations rather than bolted on here.

### Gemini — Medium: my test wrote into the real `~/.nuget/packages`

Correct, and it violated a rule stated in the code it was testing —
`GetNuGetCachePath` carries an explicit `// NEVER write to ~/.nuget/packages`.
It now redirects the root, in a collection that disables parallelization because
`NUGET_PACKAGES` is process-wide.

### The blast radius was the finding's best evidence

Removing the rooted branch broke **19 tests** — 2 in Services, 17 `Project_*` in
the CLI — all using an absolute `library.path` as a fixture convenience. The
rooted branch was load-bearing **for tests and nothing else**, which is the
clearest possible argument for deleting it. They now redirect the package root,
scoped *inside* the semaphore that already serializes console capture, so a
redirected run cannot be observed by a concurrent one.

### Non-vacuity: I reverted each fix and checked

**13 tests fail** with the three fixes reverted. Two of my new tests **passed**
with their fix reverted — both were vacuous and both were rewritten:

- the rooted-path test drove `Parse`, which does not route through
  `ResolvePackagePath` at all;
- a `ResolveAssembly` theory asserted only that unresolvable names fail to
  resolve, which is true without any guard.

Rechecked after rewriting: both now fail without their fix.

### Validation at `f92ee885`

| Suite | Result |
| --- | --- |
| `ILInspector.Metadata.Tests` | 1119 pass / 5 skip |
| `DotnetInspector.Services.Tests` | 373 pass / 0 fail |
| `dotnet-inspect.Tests` | 2363 pass / 14 skip |

Both round-10 review worktrees left clean. Not ready to merge — this head has
not been reviewed. `origin/main` also needs merging in first.

richlander:
## Round 11 reconciliation (GPT-5.6 + Gemini 3.1 Pro, both at `f92ee885`)

Two independent reviews. Head is now `dc9088ab`. **Four findings actioned, one disproved.**

Both reviewers independently reported the same first finding, and it is real — but it belongs to **#3429**, not here.

### Actioned

**Type-forwarder targets (GPT, High)** — `ApiServices.ResolveForwardedTypes` joined an `AssemblyRef` name read from the inspected assembly's metadata onto that assembly's directory and opened the result, merging its API surface into the output. A forwarder target of `../payload` read an assembly outside the directory being inspected. It bypassed both the `PlatformResolver` chokepoint and the guarded `PdbContext` path.

**Tool payload package ids (GPT, Medium)** — `RidPackageVerifier` and `LibraryCommand.TryFindLocalSiblingPackage` took a package id from the inspected package's `DotnetToolSettings.xml` and a version from its nuspec and built a local `.nupkg` path from them; `LibraryCommand` then extracts whatever it finds.

**Decompiler sibling resolver — the eighth sink (Gemini, High)** — `MetadataSource.SiblingAssemblyReferenceResolver.Resolve` does its own `Path.Combine(_directory, identity.Name + ".dll")`. Genuinely independent: it never calls `PlatformResolver`, so the chokepoint added earlier in this PR does not cover it.

**Full-width device names (Gemini, Medium)** — confirmed by probe against the built assembly:

```text
FF23,FF2F,FF2D,0031  safe=True  NFKC=COM1
FF23,FF2F,FF2E       safe=True  NFKC=CON
FF21,FF35,FF38       safe=True  NFKC=AUX
```

Round 6 fixed this bug class for *digits* (`COM¹`) and I stopped there. Windows best-fit maps the whole component, so full-width **letters** reach the same device, and `OrdinalIgnoreCase` does not match `\uFF23` to `C`. The device-name check now also runs over the NFKC form.

Normalization does **not** replace the numeric fold — NFKC leaves Arabic-Indic digits alone, so `COM٤` survives it unchanged. Both run, which is what catches `ＣＯＭ٤` and `ＣＯＮ．txt`, where a full-width dot hid the stem boundary from an ASCII-only split. Normalization decides device-name identity only; it never rewrites the value that gets used, and four accept cases pin that ordinary full-width and Japanese names are still accepted.

### Disproved

**Environment-variable race (Gemini, Medium)** — that concurrent test classes would observe `NUGET_PACKAGES`. Its two named classes:

- `ProjectAssetsParserCacheRootTests` is in **`DotnetInspector.Services.Tests`**, a different assembly and therefore a different process. It cannot observe this assembly's environment.
- `PackageStoreTests` is `[Collection("Console")]` — the *same* collection as `CommandExecutionTests`. xUnit serializes a collection, so they never run concurrently. The finding read the shared collection as evidence of non-isolation when it is the isolation.

`ConsoleCapture.cs` is the only file in `dotnet-inspect.Tests` that touches the variable at all.

### Already fixed in #3429 (both reviewers, High)

`LibraryMetadataService.BuildTransitiveReferences` resolves `Path.Combine(sourceDir, reference.Name + ".dll")` locally *before* reaching the guarded `PlatformResolver`, so a traversing name never reaches the chokepoint. Correct, and correctly a High. Both reviewers saw it because this branch does not carry #3429, where the guard is placed **before** the local combine and so covers both the local and platform paths. No change here; #3429 owns it.

### Non-vacuity

Every new gate was tamper-checked, not just run. My first tamper attempt *failed to compile* (`CS0162`, warnings-as-errors) and the runner silently used the stale binary — the same contaminated-baseline trap from round 10, caught this time by the compiler. Deleting the guard blocks outright:

- both hostile path tests fail, both positive controls still pass;
- removing the normalization check fails exactly the six new reject cases and leaves the four accept cases passing.

The forwarder failure is worth quoting, because it is the vulnerability rather than a proxy for it:

```text
ResolveForwardedTypes_WithTraversingTargetAssembly_DoesNotReadOutsideTheAssemblyDirectory [FAIL]
  Assert.Empty() Failure: Collection was not empty
  Collection: [ApiType { ... BaseType = "System.Object" ... }]
```

That is the planted assembly's type, read from outside the inspected directory.

### Validation at `dc9088ab`

```text
ILInspector.Metadata.Tests       Total: 1129, Failed: 0, Skipped: 5
DotnetInspector.Services.Tests   Total:  373, Failed: 0
ILInspector.Decompiler.Tests     Total: 4319, Failed: 0    (--gate fast; the eighth sink is decompiler code)
dotnet-inspect.Tests             Total: 2367, Failed: 0, Skipped: 14
```

Not yet ready to merge: this head has no clean review, and `origin/main` has moved since.

richlander:
## Round 12 reconciliation — head `5626f25d`

Two reviewers reached **opposite** conclusions on the same code, so I resolved it against an authoritative source rather than picking a side.

- **GPT (round 12):** the NFKC device check **over-rejects**.
- **Gemini (round 11):** the guard **under-rejects** fullwidth device lookalikes because Windows best-fit maps the whole component. I acted on this in `dc9088ab`.

**GPT is right; Gemini's mechanism claim was wrong, and I should not have accepted it without checking whether it applied to this code path.**

Per the [Windows path-parsing reference](https://chrisdenton.github.io/omnipath/), the device matcher uppercases ASCII letters and strips trailing dots and spaces — and does nothing else. No Unicode normalization, no best-fit. Best-fit mapping happens when a wide string is converted for an **ANSI** entry point; .NET calls the wide APIs, so a path component never passes through it.

The same reference lists `COM¹ COM² COM³ LPT¹ LPT² LPT³` as reserved names **in their own right, as literal strings** — not as fold products. That is why round 6's digit fold looked correct: it got those six right by the wrong rule, and the wrong rule also rejected `COM⁴`, `COM１`, `COM١`, astral math digits, and every fullwidth-letter spelling, none of which are devices.

`5626f25d` therefore **reverses part of round 6 and part of round 11**: the six superscript names are literal entries; the stem is split at the first dot, trailed of dots and spaces, and compared ordinal-case-insensitively. `FoldDigits`, `AsciiDigitFor`, `ContainsNonAsciiDigit`, `StemNamesReservedDevice` and the NFKC pass are deleted. The previously-rejected non-device spellings moved from the reject theory to the accept theory.

### The tamper that mattered

Deleting the six literal entries fails 7 tests. Deleting the trailing dot/space strip initially failed **nothing** — the two cases I wrote for it (`COM¹ `, `COM¹.`) are already refused by the edge-whitespace and trailing-dot rules further up, so the strip never ran. It is load-bearing only when the space *precedes* the extension, where the component does not end in a space: `COM1 .txt` and `COM1 . .ext` both reach the device. Those are now the pinned cases, and they are what the tamper fails. A third, `CON  .`, was dropped for being vacuous the same way.

`PdbContextForwarderResolutionTests` used `COM⁴` as its hostile forwarder name; since that is not a device, it is now `COM¹`.

Metadata 1130 / Services 373 / CLI 2405 — 0 failed.

### Also from round 12

GPT's `--rows`-ignored-by-IL-offset-JSON note reproduces on **merged main**, so it is out of scope here; filing separately.

richlander:
## Round 13 reconciliation — head `d8b58366`

Two reviews at `5626f25d` (Gemini, MAI). **Both independently confirmed the round-12 reversal was correct** — including Gemini, which reverses its own round-11 position. Gemini: "the removal of the fold passes is completely safe and fixes a regression of over-rejection." MAI: the implementation "matches the Windows device-name rule from the cited reference." Over-rejection sweeps: 31,259 and 2,368 components, **0 rejections**.

Three findings actioned, all about what the user *sees* when a guard fires rather than whether it fires.

### 1. The tool crashed on a refused package coordinate

Found while checking Gemini's ninth-sink claim by running hostile ids through the real CLI:

```console
$ dotnet-inspect package CON@1.0.0
Unhandled exception: System.ArgumentException: Invalid package name: 'con'
   at ILInspector.MetadataPrimitives.HardenedPath.ValidatePathComponent(...)
   ... ~1800 bytes of stack trace
```

Base prints `Error: Package 'con' not found.` in 32 bytes. Both exit 1, so the failure was never *invisible* — only unreadable.

**This is a presentation regression this PR introduced.** The throw in `NuGetCache.GetPackageCachePath` is not new, but it was unreachable for an ordinary name: the local copy of the rule this PR deleted could only reject separators, traversal and volume qualifiers, all of which the argument parser rejects earlier. Tightening the rule made it reachable. `DownloadAndExtractPackageAsync` now refuses the coordinate as an ordinary acquisition failure:

```console
$ dotnet-inspect package CON@1.0.0
Error: Invalid package name: 'con'.
```

### 2. Refusals were reported through a verbose-gated logger

Gemini flagged `RidPackageVerifier`; MAI flagged the same pattern independently. `VerboseLogger.Log` writes nothing without `-v`, so a refused name left a quietly thinner result and exit 0 — the success-shaped failure `AGENTS.md` forbids. Added `VerboseLogger.Warn`, which writes unconditionally, and routed the `ApiServices` and `RidPackageVerifier` refusals through it. `CoreCache` already wrote its cache-escape refusal this way. Refusals are rare enough not to become noise: 0 across 30,454 real assemblies.

### 3. Two hostile cases proved nothing where they sat

`"CON "` and `"NUL   "` were grouped under "names the host rewrites before opening" to cover the trailing-space rule, but both reach the *device* rule first. Regrouped; the comment now sits on `"Foo."` and `"System.Text.Json "`, which are non-devices and do cover it. Same defect as the `COM1 . .ext` case corrected in `5626f25d` — a hostile case can pass for the wrong reason.

Also documented that `Path.IsPathRooted` in `IsSafePathComponent` is redundant (a rooted value must contain a separator or volume qualifier, both rejected above) so it is not misread as the rule that stops rooted paths. Kept as a backstop; Gemini confirmed removing it fails no test.

### Not actioned

- **Gemini, High — ninth sink** (`PackageExtractor` tool-wrapper redirect reaching `Path.Combine(tempDir, $"{packageName}.{version}.nupkg")`). Already guarded: `GetPackageCachePath` validates the same coordinate earlier in `DownloadAndExtractPackageAsync` and throws before that line. Verified by running hostile ids through the CLI — which is what surfaced finding 1.
- **MAI, High — `LibraryMetadataService.BuildTransitiveReferences`.** Real, but fixed in #3429, which this branch does not carry. Third round to report it; the guard there sits before the local combine and still emits the node so evidence is not hidden.
- **Gemini, Medium — `CacheCommand.IsSafeSessionName` as a seventh copy.** Session name is a CLI argument, not inspected metadata, so it is outside this PR's threat model. Worth consolidating, not here.

MAI found no seventh surviving copy and no ninth sink.

### Validation

Non-vacuity: deleting the coordinate guard fails both new theory cases, on a build confirmed to have succeeded. Metadata **1130**, Services **373**, CLI **2407** — 0 failed.

richlander:
## Round-14 reconciliation — head `94f04afe`

Reviewed at `d8b58366` by Gemini. Four findings: **two actioned, one actioned differently than suggested, one disproved.**

### 1. Actioned (High) — `package CON@1.0.0 --version` still crashed

```
Unhandled exception: System.ArgumentException: Invalid package name: 'con'
   at NuGetCache.TryGetCachedPackage(...)
   at PackageCommand.ExecuteAsync(...)
```

Round 13 fixed this same crash in `DownloadAndExtractPackageAsync`, **and its test only ever exercised that route**, so this one kept crashing while the suite stayed green. I fixed a call site instead of the defect.

Gemini offered two fixes; I took the second. **The defect is a `Try` method that throws.** `TryGetCachedPackage` now answers `null` for an unsafe coordinate — its documented contract, and also the truth, since no cache entry can be named by one. Guarding each caller would have left the next caller to find.

That single change also closes finding 3 (`TryGetNuspecXmlAsync`, Medium), which was the same throw surfacing as `ERROR checking (CON 1.0.0)` instead of a clean refusal. Output now matches base byte-for-byte: `Error: Package 'con' not found.` The refusal stays visible on every path that actually acquires or extracts.

New test covers the `--version` route explicitly, because covering one route is what caused this.

### 2. Disproved (High) — `ProjectAssetsParser` is not a ninth sink

It is guarded, by this PR, at `ProjectAssetsParser.cs:210-211` — two lines below the line cited.

I did not stop at reading it. I built a hostile `project.assets.json` escaping to a real `/tmp/evil.dll` and ran the parser. **My first probe returned 0 and I nearly reported it as disproved — but the benign control also returned 0.** The probe was loading a stale dependency and `Parse` was swallowing the `FileNotFoundException` in its catch. Both zeros were artifacts of my harness, not the guard. That is the same vacuous-green trap this PR keeps hitting, and the control is the only reason I caught it.

With the probe fixed: guard removed → `results=1`, resolving to `/tmp/evil.dll`. Guard restored → `results=0`, refused before the `Path.Combine`.

**But the finding was still worth its weight**, because nothing in the suite contradicted it. The rooted-path test drives `library.path`; the only test that touched a `compile` key **asserted nothing at all** — no `Assert`, with a comment admitting it. A reviewer reading that code had no way to tell the guard worked. `Parse_WithTraversingCompilePath_DoesNotEscapeTheNuGetCache` now pins the `asm.Name` half specifically, with a deliberately benign `library.path` so the other half cannot mask a regression; removing only that half fails it. The assertion-free test is deleted.

### 3. Actioned (Low) — my test regrouping was factually wrong

Gemini is right, and this one stings: `char.IsWhiteSpace(value[^1])` runs **before** `IsReservedDeviceName`, so `"CON "` and `"NUL   "` are refused as trailing-whitespace names and never reach the device rule. I moved them *to* the device group in round 13 on the opposite claim. Restored to the group they belonged in.

The round-13 note warned that a hostile case can be refused by an earlier rule than the one it is filed under. I then made exactly that error while acting on it — by asserting the rule order rather than reading it.

## Validation

Metadata **1130**, Services **373**, CLI **2409** — 0 failed.

`94f04afe` is unreviewed, so this is not `Ready to merge`.

richlander:
## Round 16 reconciliation — MAI at `94f04afe`, fixed head `871a2cad`

MAI was asked specifically to falsify "the refusal stays visible", since round 15 converted `TryGetCachedPackage` from throwing to returning `null` — and `AGENTS.md` forbids turning a failure into success-shaped empty output. That was the right thing to attack, and it found one real half and one wrong half.

### Real — `TryGetNuspecXmlAsync` refused silently

The null-returning `Try` method is correct as a contract, but one caller consulted the cache for an unsafe coordinate and then reported "not found" with no indication that a *refusal* had occurred. Fixed: `TryGetNuspecXmlAsync` now emits `Refusing unsafe package coordinate '…'` through the unconditional `VerboseLogger.Warn` channel before consulting the cache, matching the pattern round 13 established for `CoreCache`'s cache-escape warning.

New `PackageExtractorCoordinateRefusalTests`: 3 hostile coordinates plus 1 positive control. Silencing the log fails all three hostile cases; the control still passes, so the test is asserting the warning rather than the absence of output.

### Disproved — the `RidPackageVerifier` half

MAI reported the same silent-refusal pattern in `RidPackageVerifier`. That class already guards **and warns at line 31**, before any nuspec path is constructed, so the refusal was never silent there. Reproduced by reading the guard and running the path; no change made.

### Standing note on the `Try` conversion

Worth restating why the null return is the fix and not a regression. Round 13 guarded `DownloadAndExtractPackageAsync` and its test covered only that route; the identical crash then survived on the `--version` route, with a green suite. The defect was never the call site — it was a `Try` method that throws. Refusals are now *quiet in control flow and loud in output*, which is the split `AGENTS.md` asks for.

**Suites:** Metadata 1130, Services 377, CLI 2443 — 0 failed.

Not ready to merge — `871a2cad` is unreviewed.

richlander:
## Convergence fix: the shared owner no longer over-rejects embedded dots — head `0a2fa8d0`

Recording this here because it resolves a divergence between this PR and #3429 that would otherwise have been reintroduced by the very cleanup this PR exists to enable.

#3429's round-16 review found that its local copy of the rule refused **any** embedded `..`, which is wrong: the value is a single path component, and a component with no separator cannot leave its directory. Probed directly —

```
Valid..Dependency.dll    -> /tmp/dotprobe/app/Valid..Dependency.dll  escapes=False
Foo..Bar..Baz.dll        -> /tmp/dotprobe/app/Foo..Bar..Baz.dll      escapes=False
..LeadingDots.dll        -> /tmp/dotprobe/app/..LeadingDots.dll      escapes=False
a/../../evil.dll         -> /tmp/dotprobe/evil.dll                   escapes=True
```

Only the spelling with separators escapes. Traversal is stopped by the separator and rooting rejections; the substring check was contributing nothing to safety and was costing real assembly names the C# compiler emits.

`HardenedPath.IsSafePathComponent` had the identical over-strict rule. Since #3429's `IsSafeAssemblySimpleName` is scheduled to delegate here once both land, leaving it would have quietly restored the over-rejection at the moment of convergence. It is narrowed here.

### The part worth reviewing carefully

Removing the check changes *which rule* refuses `.` and `..`. Every all-dot name ends in a dot, so the trailing-dot rule further down already refuses them. That matters more than it looks, because `IsSafeRelativePath` splits on separators and hands `..` to the component rule — so **relative-path traversal refusal now rests on a rule written for a different reason.**

Per `AGENTS.md` ("asserted properties name their gate"), that dependency is stated in the remarks rather than left implicit, and `TraversalSegment_IsRefusedByTheComponentRule` pins it. Deleting the trailing-dot rule fails that test **plus all seven `HostileRelativePath` traversal cases**. Before this commit it would have failed none of them, because the substring check was covering for it — the property was real but ungated, which is exactly the state `AGENTS.md` says is indistinguishable from a verified one.

Non-vacuity, both directions, on builds confirmed `Build succeeded`:

| Tamper | Result |
| --- | --- |
| restore the `..` substring check | 3 accept theories + `TraversalSegment_IsRefusedByTheComponentRule` fail |
| delete the trailing-dot rule | 11 fail, including every traversal case |

**Suites:** Metadata 1134, Services 377, CLI 2443 — 0 failed.

Round 17 review is running against this exact head. Not ready to merge.

richlander:
## Round 17 reconciliation — Gemini + MAI at `0a2fa8d0`, fixed head `64f5a914`

Three findings across two reviewers. All three reproduced before acting; all three fixed.

### Gemini, Medium — supplementary-plane format characters bypassed the Trojan Source check

The strongest finding of the round, and a genuine hole in code written specifically to close this attack.

The scan was `foreach (var c in value)`. For a character outside the BMP, `char.GetUnicodeCategory` sees only the two surrogate halves and reports `Surrogate` for each — never `Format`. The Plane 14 tag block **U+E0000–U+E007F is `Format` in its entirety** and renders as nothing. Probed directly against the built assembly:

```
BOM U+FEFF (BMP Cf)      safe=False
RLO U+202E (BMP Cf)      safe=False
TAG SPACE U+E0020        safe=True   <-- renders as "ValidDependency"
TAG LATIN A U+E0041      safe=True
LANG TAG U+E0001         safe=True
```

Now scans **runes**. The surrogate check moved *above* it and says why: `EnumerateRunes` substitutes U+FFFD for an unpaired surrogate, which is not `Format` and would sail past the scan that replaced the char loop.

Four new reject cases fail under the old char scan. Two new **accept** cases (U+1F600, U+20000) pass under both — that pairing is the point: the rule must discriminate by Unicode category, not by "outside the BMP".

### MAI, High — a ninth path sink

`NuGetCache.TryGetLatestCachedVersion` combined a caller-supplied package name into a cache root with no validation, reachable from the CLI as `package ../foo --version`. Guarded **in the method, not at the call site** — six call sites reach it, and round 13 of this series guarded one route while the identical hole survived on a sibling route with a fully green suite.

MAI also noted the user-facing message. A refusal was reported as `Package '../foo' not found on nuget.org` — i.e. *searched for and absent*, when it was never searched for at all:

```
before: Error: Package '../foo' not found on nuget.org
after:  Error: Invalid package name: '../foo'.
control: package newtonsoft.json --version -> 13.0.3
```

### Gemini, High — the resolver refused silently

An unsafe package coordinate or deps.json asset was dropped with a bare `continue`. That is indistinguishable from a dependency that legitimately did not resolve: it simply vanished from the tree. `AGENTS.md` forbids exactly this.

Added `AssemblyDependencyResolutionOptions.Log` and threaded it through the recursive collection walk and the deps.json asset groups — and through to `ProjectAssetsParser.Parse`, **which already had a `log` parameter that this caller was passing `null` to**.

New `DependencyRefusalVisibilityTests`, each with a positive control: the hostile entry must be reported *and* the benign entry in the same file must not be, so a guard that refused everything would not pass. Silencing either log call fails the test. A second test pins that the log channel stays optional.

### Verified, not taken on faith

Gemini's own tamper log independently reproduced both round-16 non-vacuity results (11 failures for the trailing-dot rule, 4 for the substring check), and both reviewers ran over-rejection sweeps with **zero** false rejections — MAI's covered 75 package ids, 130 versions, 23,952 assembly simple names, **609,389 type-forwarder targets** and 3,928 relative paths.

Neither reviewer re-litigated the settled NFKC/best-fit question.

**Suites:** Metadata 1140, Services 384, CLI 2443 — 0 failed.

Not ready to merge — `64f5a914` is unreviewed.

richlander:
## Fixed-head review reconciliation (was `22682721`, now `85f747fc`)

Reviewed by **Gemini Pro** at `22682721`. Four findings; **two actioned, one already fixed elsewhere, one incorrect**.

### 1. Refusals were invisible — **valid, fixed**

Gemini flagged refusals reported through `log?.Invoke`. On investigation it is worse than reported: **no product caller passes a logger to these methods at all**, so the message went nowhere unconditionally — not merely "verbose-gated". The rest do pass `VerboseLogger.Log`, which is silent by default. A refused coordinate was indistinguishable from an absent one.

All six sites now go through `HardenedPath.ReportRefusal`, which writes to stderr unconditionally, and lives beside the rule rather than at the six call sites — a refusal announced six different ways is exactly the drift this PR exists to end. Making it a no-op fails **8** tests.

**A second finding fell out of this.** The two existing visibility tests asserted on a log delegate **the tests themselves supplied**:

```csharp
Log = messages.Add
...
Assert.Contains(messages, m => m.Contains("Refusing unsafe deps.json", ...));
```

That proves only that the product writes to a channel the test opened. No product caller opens one, so the assertion was satisfied while real users saw nothing — green for the wrong reason, the same pattern that has dominated #3429's review. Both now assert on captured stderr.

### 2. Vacuous traversal theory — **valid, fixed**

Confirmed exactly as described. With the guard removed entirely, **only the `CON` row failed**; both traversal rows passed. `Path.Combine` does not normalise, so an escaping id yields `<root>/../escape/…` — a string that both `StartsWith` `<root>/` *and* differs character-for-character from the payload's own path, so every assertion held while the escape succeeded.

Comparisons are now canonicalised with `Path.GetFullPath`. Re-running with the guard disabled now fails the `../escape` row as well as `CON`. (`..\escape` legitimately does not traverse on macOS — that host-dependence is already documented on the theory and gated by `HostileDependencyIds_ContainACaseThatTraversesOnThisHost`.)

### 3. "Tenth unguarded sink" in `BuildTransitiveReferences` — **real, but already fixed in #3429**

Accurate against this branch, which forks from `main`. #3429 already closes it, with `BuildTransitiveReferences_TraversingName_RefusesToResolveOutsideSourceDir` planting a real payload at Gemini's exact `../payload` coordinate. It closes here when #3429 merges; not duplicated.

### 4. "Unverified guard: `value.Length > 255`" — **incorrect**

Gemini reported that deleting the length cap causes 0 test failures. It does not. `HardenedPathTests.OverlongComponent_IsRejected` exists at `22682721` — the exact head reviewed — and asserts both sides of the boundary:

```csharp
Assert.False(HardenedPath.IsSafePathComponent(new string('a', 256)));
Assert.True(HardenedPath.IsSafePathComponent(new string('a', 255)));
```

I reproduced the deletion: it fails that test. I had written a replacement before checking, found it was a duplicate, and removed it. No change made.

### Cross-PR note

#3429 independently added an identical `VerboseLogger.Warn`. I aligned #3429 to this PR's signature, prefix, and doc comment so the two do not collide on merge.

### Validation

| Suite | Result |
| --- | --- |
| `ILInspector.Metadata.Tests` | 1140 passed, 0 failed, 5 skipped |
| `DotnetInspector.Services.Tests` | 389 passed, 0 failed |
| `dotnet-inspect.Tests` | 2512 passed, 0 failed, 14 skipped |

Still blocked behind #3429 per the agreed order. Needs a fresh fixed-head review at `85f747fc`.

PR #3461, opened 2026-07-29
Character-level annotation positions, and a printed body map that can leave the process (#3328)

richlander:
Phase 2 of #3328. Phase 1 (#3399) recorded a range per statement; this makes an annotation able to point at the characters it is about, and — the larger change — makes that positional data leave the process.

## Two changes worth reading separately

**1. Positions moved from line-level to character-level.** An expression now claims the characters it printed, bound within its parent's window rather than its statement's, with siblings completed left to right. `sink.Add(new object())` can be pointed at as `new object()` instead of as a whole line.

These positions are still *inferred*: the printer composes bottom-up by returning strings, so a node's offset is recovered by searching the printed line for its text, and the search refuses when a spelling is ambiguous. That refusal is why the result is trustworthy, and it is also why this is not the end state — see "What is deliberately not here".

**2. The data can now leave the decompiler.** `PrintedBodyMap` is a payload of lines, columns, lengths and names — no object references. The printer's own map is keyed by `IrNode`, whose identity is the CLR object reference, so using it means running inside the decompiler holding its object graph. That made annotated source a vertical: grab the structured data, print the pretty result, and every other consumer re-implements the guessing.

With the payload, the gesture becomes the renderer's choice rather than something baked into the datum — which is what `AnnotationCaret` already documented ("a span-carrying datum can underline a narrower range; that is a property of the datum, not of this gesture").

## Evidence

Rendered from the serialised JSON alone, with no decompiler in scope (`probe/envelope`):

```
--- SIDE ---
return new int[n];    // alloc.array: int[]; escape=escapes-return

--- CARET (focus: Allocation) ---
return new int[n];
^^^^^^^^^^^^^^^^^^ alloc.array: int[]; escape=escapes-return
```

Doing that end to end found two defects that the unit tests could not have caught, both fixed here:

- **Every statement was unplaceable.** A statement's range runs to the end of the line it printed, newline included, and the projection treated that as crossing a line break. Since `AnnotationAnchor` binds facts to statements, *every* fact degraded to "whole line" — the payload was useless for carets until this was fixed.
- **Category was missing.** `AnnotationGestureSelector.Focus` promotes on category *or* id, so a consumer holding only the payload could honour `focus alloc.box` but not `focus allocation` — the selector's own documented input.

## A latent bug found by review

`ResearchViews` builds the same line→IL-offset table as `CSharpBodyDiff.BuildSourceLines` in two places, without the `IrExpression` guard that keeps expression nodes out of it. Measured on a frozen corpus: **7,599 of 9,218 methods differ, 38,861 line offsets**. The four overlay sections rendered across 86 members are byte-identical between arms, so the divergence does not reach what is rendered today — but the two tables must agree, and now do.

## Measurements (frozen, hash-pinned corpus, 9,218 methods)

| check | result |
| --- | --- |
| methods whose printed text changes (SHA-256) | **0** |
| base ranges lost | **0** |
| ranges recorded | 62,459 → 220,113 |
| straddling / identifier-cutting / blank / out-of-bounds | 0 |
| `Annotated Source` over 86 members (9,340 lines) | identical |
| overlay sections, 4 × 86 members (15,990 lines) | identical |
| ancestor order violations | 0 of 180,018 |
| facts that narrow today | 5,698 of 6,469 (88.1%) |
| print cost, opt-in path (best of 6, contended box) | 6,309 ms → 6,623 ms (+5%, upper bound) |

Two probe bugs produced false alarms during this work and were corrected rather than reported: keying methods by `Type|Method` collided record `.ctor` overloads and invented 1,375 phantom lost ranges, and `string.GetHashCode()` is randomised per process, which made an earlier render comparison meaningless.

## What is deliberately not here

**Oracle-provided positions.** The end state is that the printer *records* where each node's characters landed instead of searching for them, which removes the ambiguity refusals entirely. `Frag` and its interpolated-string handler are the foundation for that and land here with gates, but **no composer is converted yet** — every position in this PR is still inferred by search. `Frag` is `internal` precisely because nothing uses it.

The conversion is bounded but broad: 65 `Expression(` call sites, and spans drop at any unconverted parent, so it needs a contiguous converted chain from the statement seam down. Its measured payoff over the heuristic is **+93 facts** (5,698 → ~5,791); the reason to do it is that the printer stops guessing, not the fact count.

One trap is worth recording because it cost real time: an interpolated string is handler-converted only when the target is a *parameter* of the handler type, so `Frag M() => $"..."` binds through `string` and drops every position while still printing correct text. There is deliberately no `string`→`Frag` conversion, which makes that mistake a build error rather than a silent reversion to searching; a reflection gate fails if it is re-added.

**Caret narrowing itself.** Facts still anchor to statements, so a caret underlines a whole statement. The narrow spans already exist and are placed correctly (`NewArray [7,17)` = `new int[n]`); only the anchor is coarse.

**AOT serialisation.** `SurvivesSerialisationAndReplays` passes only because the test project permits reflection-based JSON, which is disabled repo-wide. A wasm consumer needs `[JsonSerializable(typeof(PrintedBodyMap))]` on its source-generated context. Follow-up, along with adding the IL line stream to the envelope — the join key (`SourceOffset`) already exists on both sides.

## Gates

37 test methods (42 cases with theory arguments) across `PrintedRangeMapTests` (20), `FragTests` (7), `PrintedBodyMapTests` (15). Every gate is proven non-vacuous by a mutation that breaks it and names it: 8 on the range recording, 5 on `Frag` (reinstating the string conversion, zeroing the append offset, dropping the nested offset, attributing empty text, sharing a span list across fragments), and 13 on the payload (zeroing the column, accepting a straddling range, an off-by-one length, losing the kind, emitting spans only on failure, reporting the unknown length as zero, zeroing the fact column, dropping any one of the ordering tiebreaks, dropping a fact anchored to a refused node, dropping conditionality, and pinning a refused fact to line 0). Sorting the map by position fails all three ordering gates.

Fact ordering is a total comparison, exposed so it can be tested directly: facts arrive keyed by a dictionary, whose enumeration order is not a contract, and `List.Sort` is not stable, so any pair left equal could come out in either order and the serialised payload would differ between runs over identical input.

Fast lane: 4,348 passing. The deep-inspect test lane reports 7 failures; a control run of the same lane on clean `main` reports **the same 7 by name** (#3489–#3493, which #3495 filed). They were previously invisible because the lane timed out and was cancelled rather than reporting.

## Review

Two adversarial reviews ran read-only in isolated detached worktrees (claude-opus-4.8 and gpt-5.6-sol). Four findings, all fixed in `6e9d7bc`:

- **`Frag.Attribute` mutated a shared span list** — found independently by *both* reviewers. `Spans ?? []` aliased the existing list, so attributing a node retroactively re-attributed fragments already built from the original.
- **A fact on a refused node vanished silently**, contradicting the positive-only contract. It now falls back to the nearest recorded ancestor.
- **`EverySpanSelectsExactlyTheCharactersTheNodePrinted` was vacuous** — it never read a single `PrintedNodeSpan` field; substituting wholly bogus spans still passed.
- **The enumeration contract was wrong.** It claimed siblings appear in textual order. `UnionSwitchExpression.NullValue` sits *after* the arms in `Children` but prints first (offset 111 at column 48 emitted before 222 at column 32). The earlier "0 of 45,140 sibling violations" sweep missed this because it identified siblings *by `Children` order* — the very thing that is wrong. That row has been removed from the table above and the documented contract corrected; the map was not reordered, since that would change enumeration for line-table consumers and needs its own verification sweep.

The reviews also ruled a great deal out with evidence: 215,296 placed spans all slice correctly, 223,472 ranges cut 0 identifiers, `Compare` is a valid total order over 56,623,104 triples, and the `ResearchViews` guard is complete (3 such tables in the repo, all guarded).

## Second review round

The first round reviewed the tree *before* its own fixes, so it proved nothing about them. Two fresh reviews ran against `6e9d7bcc` itself. Both found real defects; fixed in `2bd4be9`.

**The enumeration contract corrected in round one was still false.** It claimed siblings enumerate in child order. A structured statement records its **body before its condition** — the printer recurses into the block while composing the statement and records the condition afterwards.

| claim | checks | violations |
| --- | --- | --- |
| every node follows all its descendants | 434,843 | **0** |
| siblings in child order | 222,205 | **20,145** |

`IfStatement` accounts for 17,480, with `ForLoop`, `WhileLoop`, `Fixed`, `Lock`, `Switch`, `UsingStatement`, `ForeachStatement` and `CatchClause` making up the rest.

Two sweeps missed this — mine included — for the same reason: a `Block` records no range of its own, so comparing only those children that are *themselves recorded* never compares a condition against a body, and reports a clean zero. Subtree extents are what must be compared. The contract now promises descendants-before-ancestors and nothing else, and records why the weaker check misleads. The producer-side comments in `CSharpPrinter` asserted the same repudiated claim and are corrected too. The `UnionSwitchExpression` offsets I cited in round one did not reproduce; those unverified numbers are gone and only measured ones remain.

**The payload dropped `Conditionality`.** `AnnotationText` appends `cached-once` / `per-iteration` to the rendered label, so a consumer holding only the envelope would render a *different* annotation than the in-process renderer — silently promoting a cached-once allocation to an unconditional one. That defeats the point of the envelope, so it is now carried, ordered on, and round-tripped.

Both test fixes came from review:

- the sibling test asserted the false claim and passed only because its fixture was a flat expression, with no branch anywhere in the fixture set. It is replaced by two tests over subtree extents on a new branching fixture — one pinning post-order, one pinning the known inversion so it cannot be believed away again.
- the refused-node fallback test asserted only that the fact *survived*. A mutation pinning refused facts to line 0 passed against it, because its fixture was one line long and the ancestor line was genuinely 0. It now runs on a multi-line fixture and asserts the ancestor's line, and that the line is not zero.

Also ruled out with evidence this round: no residual aliasing path in `Frag` (spans are immutable value tuples with single-owner lists); the ancestor walk cannot loop and always yields an in-body line, degrading to column 0 / length -1 rather than blending coordinates; the `emitted++` pairing in the span test is sound because production and test iterate the same list under the same predicate; 544,245 span-slice checks with 0 failures.

## Third review round (post-merge)

`origin/main` is integrated (#3497, #3508, #3510, #3522) and the round ran against the merged head, per the precondition #3522 added to `AGENTS.md`: integrate before **every** round, not only the first.

Four reviews across three models. **No defects found.** Two ran against the pre-merge commit and two against the merged head:

| reviewer | model | tree | result |
| --- | --- | --- | --- |
| rev-g | claude-opus-4.8 | `2bd4be93` | clean |
| rev-h | gpt-5.6-sol | `2bd4be93` | clean |
| rev-i | gpt-5.6-sol | `d8df7a17` | clean |
| rev-j | gemini-3.1-pro | `d8df7a17` | clean |

Evidence, not just verdicts: post-order verified by subtree extent with **0 violations** on every corpus tried, up to 181 assemblies / 135,228 methods / 1,732,655 checks; containment 1,877,710 checks and token boundaries 2,945,165 checks, both 0; comparator transitivity up to 2,097,152 triples including null strings and undefined enum values; and `AnnotationConditionality` round-tripping through a **source-generated** `JsonSerializerContext`, which is the path a wasm consumer actually uses. rev-g independently mutation-proved all five new gates.

The merge itself was checked for semantic interaction: the main-side commits touch metadata and CLI projection paths only, nothing under the decompiler, IR, printer or `ResearchViews`.

Two things came out of the round that were worth fixing or recording:

- **The cited counts were not reproducible as written.** Three sweeps measured the same property and got three different denominators (434,843 vs 1,095,503 post-order checks; 20,145 vs 22,718 sibling inversions) while all agreeing on zero post-order violations. "A check" isn't defined until the pairing is, so the doc now states the pairing — and a reviewer using it reproduces 434,843/0 and 222,205/20,145 with `IfStatement` 17,480 exactly.
- **`CatchClause` inverts its filter** for the same reason `IfStatement` inverts its condition: a filter is a condition. One reviewer measured 0 of these using the recorded-children method the surrounding paragraph warns against; by subtree extent there are 19.


richlander:
## Replaying the issue's own examples

Both examples in #3328 now land exactly where the issue said they should.
Every caret below is drawn from `PrintedRangeMap`, sliced out of the printer's
own output — none of it is hand-written.

**Before** — the caret claims the whole statement, and `sink.Add` allocates
nothing. The IL two rows down carries the truth the C# view is missing:

```csharp
        sink.Add(new object());
//      ^^^^^^^^^^^^^^^^^^^^^^^ alloc.new(object; ...)
            // IL_000B: newobj object::.ctor()
```

**After:**

```csharp
        sink.Add(new object());
//               ^^^^^^^^^^^^ alloc.new(object; ...)
```

The issue's harder first example, `new()`, works too:

Before:

```csharp
List<object> sink = new();
^^^^^^^^^^^^^^^^^^^^^^^^^^ alloc.new(List<object>; ...)
```

After:

```csharp
List<object> sink = new();
                    ^^^^^ alloc.new(List<object>; ...)
```

```text
line  0 col 20 len  5 -> "new()"
line  3 col 13 len 12 -> "new object()"
```

## What the reviews caught

Two adversarial reviews disagreed on the central question — whether textual
uniqueness implies ownership. Settling it empirically decided it: a claim
**could** land inside a string literal.

Emission re-breaks a too-wide fluent chain one call per line *after* its links
were composed. The link's captured spelling no longer occurs at its own site,
while a literal spelling the same thing survives — so the literal becomes the
only match:

```text
return f
    .A()
    .B("f.A()")
    ...
  Call [26,31) "f.A()"    ← inside the string literal
```

Precise, confident, and wrong — the one outcome worse than the coarse underline
this feature exists to replace.

**Fix:** each node is searched only within the characters its parent claimed,
and a node whose parent refused refuses too. Reformatting breaks the chain at
the first ancestor that no longer occurs contiguously, and everything beneath
declines rather than guesses.

The narrower window also *reaches further* — not the reason for it, but the
larger effect. A short spelling ambiguous across a whole statement is often
unique inside its own operand: `Wrap(x) + x` could place neither `x` before.

Three more findings, all real:

| Finding | Before | After |
| --- | --- | --- |
| Nested expressions recorded parent-first, breaking the advertised completion order | 5,533 methods, 81,505 nodes | 0 |
| `new()` bypasses `Expression()`, so it claimed nothing | whole declaration | `new()` |
| Expression ranges leaked into `CSharpBodyDiff` line offsets | 215 lines, 86 methods | 0 |

The order test could not have caught the first: pointed at a fixture of leaf
expressions, it only ever compared an expression against its enclosing
*statement*. It now runs on nested expressions and asserts it found some.

The diff leak is the kind a text hash cannot see — the render is byte-identical
either way. `BuildSourceLines` is contracted to report "-1 when the line owns
no statement", and expression ranges land on continuation lines that own none.

## Measured over 9,195 methods

| Check | Result |
| --- | --- |
| Renders vs merge base | byte-identical |
| Claims | 142,566 → 159,891 |
| Blank / out-of-bounds / identifier-cutting | 0 / 0 / 0 |
| Straddling, of 40,273,930 pairs | 0 |
| Escaping an ancestor, of 179,466 pairs | 0 |
| Enumeration-order violations | 0 |
| Diff line offsets moved | 0 |
| Facts a caret can narrow | 575 → **773** of 878 |
| Underlined text | **−37.6%** |
| Print time, best-of-6 | 6,234 ms against a 5,419–6,689 ms spread — no cost established |

## Gates

Four mutations, each failing a different test:

| Mutation | Fails |
| --- | --- |
| Search the whole statement | `Ambiguity_IsJudgedWithinTheParent_...` |
| Drop refusal propagation | `ReformattedStatement_ClaimsNothing_...` |
| Record forward instead of reversed | `EnumerationOrder_IsEmissionCompletion_...` |
| Drop the `new()` capture | `TargetTypedNew_ClaimsItsOwnCharacters_...` |

18 `PrintedRangeMapTests`, 38 across the printer/range/caret/diff classes. The
full decompiler suite is running and I will post the result rather than imply I
already have it.

richlander:
## Follow-up: sibling completion order

A re-review caught that my previous fix was half right. Reversing the pre-order
walk put descendants ahead of their ancestors, but it also **reversed siblings** —
and `PrintedRangeMap`'s documented emission-completion order includes sibling
order. The printer composes a left operand before a right one, so `x` in `x + y`
completes first, and the map was recording `y` first.

Fixed in `dcda07b2a` with a genuine post-order walk: children pushed forward so
they pop right-first, then the walk reversed, yielding `(left, right, parent)`.

The order test could not have caught this. It asserted only that a node precedes
its nearest recorded *ancestor* — which **any** sibling order satisfies,
including backwards. A second test now pins sibling order directly, against a
new `TwoDistinctOperands` fixture; `TwiceTheSame` cannot serve, because
identical spellings claim nothing by design and so cannot express the question.

| Check, 9,195 methods | Result |
| --- | --- |
| Ancestor-order violations | 0 |
| Sibling-order violations, of 172,610 pairs | 0 |
| Renders vs merge base | byte-identical |
| Ancestor containment, of 179,466 pairs | 0 escapes |

Reverting to the reversed pre-order fails the new sibling test and nothing else.

### Review coverage so far

Four adversarial reviews across two rounds. Findings accepted and fixed:

| Finding | Round |
| --- | --- |
| Claims could land inside a string literal after chain wrapping | 1 |
| Nested expressions recorded parent-first, breaking completion order | 1 |
| Expression ranges leaked into `CSharpBodyDiff` line offsets | 1 |
| Target-typed `new()` bypassed capture entirely (found while replaying the issue) | 1 |
| Sibling completion order reversed | 2 |

Round 2 independently re-confirmed clean: the wrapped-literal claim is gone,
179,466 ancestor pairs with 0 escapes, 490,013 structural containment pairs with
0 violations, and no reproducible stale-parent or missing-intermediate-parent
misattribution across fluent chains, lambdas, switches, conditionals,
interpolation, or target-typed `new()`. No realistic quadratic pathology was
demonstrated either, which is why the `IndexOf` scan is left as is.

richlander:
## Correction, disclosure, and the last review finding

Three things, the first of which I need to state plainly because it invalidates numbers I already posted in this PR.

### 1. A commit on this branch contained a reviewer's mutation, and my measurements were taken against unpinned source

I ran two adversarial review agents against the same worktree I was committing from. One of them was proving my gates non-vacuous by mutating `CSharpPrinter.cs` — which is exactly what I asked it to do. While that mutation was applied, I ran `git add -A`. Commit `dcda07b2a` therefore shipped `// MUTATION: whole-statement` in `RecordExpressionRanges`, which is to say it shipped the unsound whole-statement search this PR exists to remove. That commit was pushed. Anyone who fetched it got code that fails `Ambiguity_IsJudgedWithinTheParent`.

It has been amended out, and the branch is force-pushed. But the more serious consequence is not the commit — it is that **every probe in this PR rebuilds from source on each run**, so any measurement I took while a mutating agent was live was measuring something other than what was committed. I could not tell you which of the numbers in my earlier comment were affected, so I have not tried to. I re-ran all of them in a separate worktree pinned to a *commit*, which no agent can modify:

| | earlier comment | re-measured on pinned `b20dbcb8b` |
|---|---|---|
| renders vs merge base | byte-identical | byte-identical |
| claims | 142,566 -> 159,891 | 142,566 -> 159,891 |
| ancestor-order violations | 0 | 0 |
| sibling-order violations | not measured | 0 / 172,610 |
| ancestor containment escapes | 0 / 179,466 | 0 / 179,466 |
| diff line offsets moved | 0 | 0 |
| facts narrowable | 575 -> 773 of 878 | 575 -> 773 of 878 |

The numbers survived. That is luck, not method, and the method is what I have changed: measure from a pinned commit, stage explicit paths, read `git diff --cached` before committing.

Both reviewers independently reproduced the headline numbers on the current head. One of them disclosed the same collision from its side, unprompted.

### 2. The last open finding: enumeration order named something it could not deliver

`rev2-sol` pushed on the sibling-order fix I made and found the fix's *justification* was wrong even though its behaviour was right.

The map documented enumeration as **emission-completion order**. I had taken that at face value and reasoned "the printer composes the left operand first, so the left completes first." That is not true in general:

```
// CSharpPrinter.Members.cs — CallText
string rest = Arguments(arguments.Skip(1), ...);   // arguments composed here
...                                                 // receiver composed later
```

So `sink.Add(new object())` genuinely finishes `new object()` **before** `sink`, even though `sink` prints to its left. Structural post-order and composition order disagree for every instance call in the corpus.

Worse, my own test could not have caught this, because it derived the expected order from the claimed text offsets — the same data it was checking. That is the second time in this PR I wrote an assertion that verified its input against itself. The first was a column check that read its expected column out of the offset it was validating, so a uniform shift passed it.

I fixed the prose rather than the code, and the reason matters. Composition order is an artefact of which interpolation hole a printing method fills first. It would flip under a refactor that changed nothing observable, no consumer can see it, and no test can pin it without reaching into the printer. Position is a property of the output. So the contract now names position:

> Enumeration order is post-order over the printed tree: every node follows all of its descendants, and siblings follow one another in the order their characters appear in the text.

with the `CallText` counterexample written into the doc so the next person does not re-derive the wrong justification. `1951afab2` is prose plus two test renames; the only non-comment lines in the diff are the two method names.

### 3. Where the PR stands

- 19 `PrintedRangeMapTests`, green. Five mutations proven to bite, each failing a different gate.
- Both #3328 examples now land:

```
List<object> sink = new();          sink.Add(new object());
//                  ^^^^^           //       ^^^^^^^^^^^^
```

- Full decompiler suite re-running on the final head; I will post the result rather than assert it.
- Two independent adversarial reviews, both clean on the code. The second additionally ran a sibling-disjointness sweep stronger than my own containment probe (0 overlaps) and an adversarial fixture covering wrapped chains, initializers, lambdas, switch expressions, interpolation and `nameof` — the wrapped chain correctly refuses every link, and `(x+x)*(x+x)` correctly refuses both inner operands.

richlander:
`main` is integrated (`6cd7fb508`) and the verification is now complete.

**The slow-lane failures are pre-existing, and this is now measured rather than argued.**

I dispatched `deep-inspect lane=test` on clean `main` as a control ([run 30482635035](https://github.com/richlander/dotnet-inspect/actions/runs/30482635035)), same runner, none of these changes on it. It fails the **same 7 tests, by name**:

| | `main` control | this branch |
| --- | --- | --- |
| total | 4,244 | 4,250 |
| failed | **7** | **7** |

`FidelityGateTests.PinnedFixesStayExact` (#3489), `FidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket` (#3490), `LoweredFidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket` (#3491), `ValidityCoverageReportingTests.DecompilerAssembly_MissingReturnPopulationIsPinned` (#3492), `GeneratedFixtureCatalogTests` ×3 (#3493).

All five issues were filed by #3495 itself, which also predicted this exact transition: before it, those gates timed out at 45 min each and the lane was silently *cancelled* rather than failed, so these were invisible, not absent.

**Everything in the description was re-measured after the merge**, not carried over:

- printed C# **byte-identical to `main` across all 9,218 methods** (SHA-256 per method)
- **0** ranges that `main` recorded were lost; 62,459 → 220,113
- `Annotated Source` **identical** across the arms over 86 members / 9,340 lines — so "no caret is narrowed yet" is a measured fact, not an inference
- all **8** mutations re-proven to bite, each naming the gate that catches it

**Two bugs in my own probes**, both of which had briefly shown alarming results and both now fixed: keying methods by `Type|Method` collided record `.ctor` overloads and invented 1,375 phantom "lost ranges"; and `string.GetHashCode()` is randomised per process, so my first render comparison was meaningless. Positional alignment and SHA-256 respectively.

**Where the remaining headroom actually is.** 6,469 facts, 5,698 (88.1%) narrowable here, structural ceiling 5,869. The 171-fact gap splits into two populations, and neither argues for the emitted-position rewrite:

| | facts | |
| --- | --- | --- |
| printed, but the text was ambiguous → refused | 93 | what emission would recover |
| structurally narrowable, but the only tighter node is a `Block`/`BlockContainer`/`UnsupportedNode` | 78 | mostly `alloc.statemachine` / `alloc.delegate`; narrowing to a block is not a caret win |

So the practical ceiling is much closer to 88.1% than the structural 90.7% suggests, and emitted positions stay deprioritised.

Ready to merge from my side.

PR #3462, opened 2026-07-29
Fail closed on unsupported --json projections (#3386)

### LINKED ISSUE #3386: --json still ignores --fields/--columns, and the whole-surface type path ignores projections entirely
`--json` composes with `--print`, `--value`, `--urls`, and `--paths` after #3379, but two sibling paths still silently discard a projection and exit 0.

## 1. `--fields` / `--columns` are ignored under `--json`

On both `type` and `member`, a valid field projection produces the unprojected type surface:

```console
$ dotnet-inspect type ... --fields Url --json     # byte-identical to plain --json
$ dotnet-inspect type ... --columns Url --json    # byte-identical to plain --json
```

Verified byte-identical (type: 875 bytes, sha `6306db6816d5`; member: 13,705 bytes, sha `e48eacf8c9b8`). `-D` confirms `Url` is a valid field, so this is not a rejected-input path.

`WriteJsonTypeOutput` should apply the field/column projection, or the combination should be rejected.

## 2. The whole-surface `type` listing path swallows projections entirely

`ApiCommand.cs:498` carries the same early return that #3379 fixed at `:655`, but this route has **no** projection dispatch at all:

```console
$ dotnet-inspect type --package Newtonsoft.Json@13.0.3 \
    -S Classes --fields Type --value --row 1 --json --tips q
```

exits 0 and emits a 1,012,886-byte `ApiSurface` document instead of one value.

## Why it matters

Same failure class as #3364 and #3379: a projection request is accepted and then silently dropped with a success exit code, rather than composing or being rejected. `docs/design/output-shapes.md` treats format and shape as orthogonal.

**Pre-existing, not a regression.** Both were verified byte-identical on base `729f97bf` and on the head of #3385. Found by adversarial review (GPT) of #3385.

richlander:
Fixes #3386.

## Problem

Two output paths in the `type` command silently dropped a projection request instead of honoring or rejecting it:

1. **`--fields`/`--columns` + document `--json` (no payload projection)** silently discarded the column filter and emitted the whole typed document with exit 0. Column projection has no meaning against the typed JSON document schema — table columns such as `Type` are *synthesized* from namespace+name, not document properties — and `--json` has no jq-style column-slicing facility.
2. **A payload projection (`--print`/`--value`/`--urls`/`--paths`) against the whole-surface type listing** (`type` with no type name) had no dispatch, so it dumped the full ~20 MB `ApiSurface` to stdout and *then* tripped the #3390 projection audit with a misleading "bug in dotnet-inspect; please report it" message.

## Fix (fail closed)

- `WriteFullApiOutput` now returns `int`. It rejects payload projections against the type-listing surface up front (the listing is a name table with no printable payload), and rejects `--fields`/`--columns` under `--json` when no payload projection is present. The four `TypeCommand` call sites propagate the exit code so the #3390 audit does not re-fire the double error.
- The single-type and member paths reject `--fields`/`--columns` + `--json` (no payload) as well.
- `--fields X --value --json` still composes (the column feeds `--value`); `--count`, `--tsv`/`--jsonl`/`--table`, plain `--json`, and default output are unchanged.
- Adversarial review (GPT-5.6) found that the glob and prefix-browse fallback routes never dispatched `-D` discovery — they fell through to `WriteFullApiOutput` — so a discovery request carrying `--fields --json` was rejected with a misleading message. Discovery is now dispatched on those routes before the projection guard, matching the main and platform-prefix routes.

## Why not build the projection instead

A column projection's natural JSON home is the row shape (`--jsonl`), which already exists; slicing a hierarchical document like `jq` is a capability Markout (the output library) does not have. Filed as a companion design issue: richlander/markout#173.

## Validation

- Full CLI suite green (2367 tests; 14 ilasm skips). `Package_LibrarySourceFilesSection_PreservesTypeFilterAndBlobUrls` is a pre-existing network-flaky test that passes in isolation and is unrelated to these paths.
- New tests cover single-type/member/surface rejection, the still-composing `--fields --value --json` case, the discovery-on-glob regression, and unchanged `--count`/row-format behavior; the stale #3390-era `WholeSurfaceListing_DroppedProjection_FailsLoudly` now asserts the up-front rejection.
- `docs/design/output-shapes.md` updated (markdownlint clean).

### DIFF
--- a/docs/design/output-shapes.md
+++ b/docs/design/output-shapes.md
@@ -237,6 +237,13 @@ enforced structurally rather than per command — the request is recorded from t
 parse result and the projection writers report which projection they honored, so
 a route that drops one fails loudly instead of shipping the wrong payload.
 
+The whole-surface type listing (`type` with no type name — the Classes, Structs,
+Interfaces, Enums, and Delegates sections) is a name table that exposes no
+printable payload, so `--print`/`--value`/`--urls`/`--paths` there is rejected up
+front rather than dumping the full surface and then tripping this audit. Inspect
+a single type (for example `type <Name>`) to project a member payload. `--count`
+is the one payload projection the surface does honor.
+
 Writers report *which* flag they honored rather than merely acknowledging one.
 A writer can be reached for more than one reason — the print writer also serves
 `--bare` — so an untyped signal would let it satisfy an unrelated request and let
@@ -251,7 +258,7 @@ resolved by discarding one.
 | Flag | Effect |
 | --- | --- |
 | `--markdown` | force the full Markdown Document format |
-| `--json` | render the selected shape as JSON: the whole Document when no narrower shape is selected, otherwise the projected payload (`--print`, `--value`, `--urls`, `--paths`) |
+| `--json` | render the selected shape as JSON: the whole Document when no narrower shape is selected, otherwise the projected payload (`--print`, `--value`, `--urls`, `--paths`). A column projection (`--fields`/`--columns`) does **not** compose with `--json`: JSON renders the whole document and has no column-slicing facility, so the combination is rejected rather than silently dropped — use `--tsv`/`--jsonl`/`--table` to project columns, or add `--value`/`--print` to project a payload (`--fields` then picks which column feeds it). |
 | `--tsv` / `--jsonl` | render the single selected section as TSV / JSON Lines (a Table or Vector) |
 | `--table` | render the single selected section as a space-padded pretty table |
 | `--no-header` (`--no-headers`) | drop the Table header row |
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -3675,6 +3675,21 @@ public class CommandExecutionTests
     }
 
     [Fact]
+    public async Task GlobListing_Discovery_IsHonoredNotRejected()
+    {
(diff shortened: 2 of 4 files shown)


richlander:

Two-model review at head `0d77e807`, re-reviewed at the fixed head `6a942f2d`.

**GPT-5.6 (Sol)** — one Medium/High finding:
- The glob and prefix-browse fallback routes in `type` never dispatched `-D` discovery; they fell through to `WriteFullApiOutput`, which ignored it. Once that path rejects `--fields`/`--columns` under `--json`, a discovery request such as `type 'System.Str*' -D Classes --fields Type --json` was rejected with a misleading column-projection message instead of returning the section schema. **Fixed** in `6a942f2d`: discovery is now dispatched in the not-found fallback branch before `TryWritePrefixBrowse`/glob rendering, matching the main and platform-prefix routes. Regression test `GlobListing_Discovery_IsHonoredNotRejected` added.

**Gemini 3.1 Pro** — no issues. Verified exit-code propagation across the `int?` refactor and all four `WriteFullApiOutput` call sites, the column-under-`--json` guard isolation, scalar payload composition (`--fields X --value --json`), the whole-surface payload rejection, and `--count` invariants.

**Re-review (GPT-5.6) at `6a942f2d`** — clean. Confirmed the discovery branch fires only for not-found + `-D`, no non-discovery path regresses, and the #3386 rejections still fire when `-D` is absent.

Companion design issue for a projectable JSON document capability: richlander/markout#173.

richlander:
Ready to merge

Issue #3401, opened 2026-07-28
Implement the @Metadata section-category lens on `library`


richlander:
Implement the `@Metadata` section-category lens on `library`, exposing the raw ECMA-335 metadata-table projection through the normal CLI surface. This is the last unbuilt piece of #3282 and the only remaining scope item on it.

The design is **settled, not open**. #3360 resolved every open question and recorded the answer in `docs/design/metadata-table-projection.md`. This issue is the implementation.

## Why this is mostly wiring

The library work is done and on `main`. All five gaps in #3341 landed across #3345, #3348, #3350, and #3352:

| Capability | Entry point |
| --- | --- |
| Project tables (with row window) | `MetadataTableProjector.Project(PEReader, MetadataProjectionOptions?)`; `StartRowId` + `MaxRowsPerTable` |
| Random-access row lookup | `MetadataTableProjector.ProjectRow(...)` |
| Reverse-reference search | `MetadataTableProjector.FindReferences(...)` |
| Image overview | `MetadataImageInspector.Describe(PEReader)` |
| Heap value by address | `AssemblyInspectionSession.MetadataHeapValue(...)` |

Session-shaped equivalents already exist (`MetadataTables`, `MetadataTableRow`, `MetadataReferences`, `MetadataImage`, `MetadataHeapValue`), and `DotnetInspector.MetadataRendering.MetadataProjectionRenderer` already renders all of them in Markdown and the machine formats, with caveats routed to stderr.

`src/mdi` already exercises the whole surface (`--table`, `--start-row`, `--max-rows`, `--references`, `--overview`, `--heap`, `--format`). **The projection, the rendering, and the caveat discipline are all built.** What is missing is registration in the section pipeline and the flag surface on `library`.

### 1. Sections and category

Each metadata table is **one section**, named `Metadata: <Table>` (for example `Metadata: TypeRef`). Register the group as a category exactly as `@Performance` is registered:

```csharp
.AddCategory(SectionCategoryNames.Metadata, MetadataTables.Sections)
```

Precedent verified: `LibrarySections.cs:95` does this for `@Performance`, and `SectionNames.cs:191-212` already defines colon-and-space section names such as `"Performance: Boxing"`. `-S "Performance: Boxing"` works today, so `-S "Metadata: TypeRef"` needs no parser change.

### 2. Disclosure

[...]

PR #3465, opened 2026-07-29
Add the @Metadata lens: raw ECMA-335 tables on `library`

richlander:
Adds the `@Metadata` section-category lens to `library`: the raw ECMA-335 metadata tables, reachable through the normal CLI section surface rather than a separate command or flag. Progresses #3401.

## What lands

One section per projected table (`Metadata: TypeRef`, `Metadata: MethodDef`, …) plus `Metadata: Image` for the image-level facts that are not rows, grouped under a `@Metadata` category door. The lens adds **no new shape flags** — a table is already a section, and `-S` already addresses sections — so all four rungs of the shape ladder come free:

```bash
dotnet-inspect library My.dll -S @Metadata                                 # Document
dotnet-inspect library My.dll -S "Metadata: TypeRef"                       # Table
dotnet-inspect library My.dll -S "Metadata: TypeRef" --columns Name --tsv  # Vector
dotnet-inspect library My.dll -S "Metadata: TypeDef" --count               # Scalar
```

Section names are **derived** from `MetadataTableProjector.ProjectedTables`, and column names from `MetadataTableProjector.ColumnsFor`, so a table added to or removed from the projector moves its section and its schema with it. A stale or missing entry is not expressible.

## Acceptance status against #3401

| Criterion | State |
| --- | --- |
| `-S @Metadata` selects the group | met |
| Each rung of the ladder renders in Markdown and the machine formats | met |
| No metadata section appears at any verbosity without explicit selection | met |
| `-D` lists the category | met |
| `-D` lists the `--heap` coordinate section only when the carrier is present | **not met — deferred** |
| Caveats reach stderr in machine formats, inline in Markdown | met |
| No change to typed-extractor layering | met |

Issue items **5 (`--heap` coordinate carrier)** and **6 (hex table selection, `0x02` ≡ `TypeDef`)** are not in this PR. #3401 therefore stays open; follow-ups are stacked on this branch.

## Design notes a reviewer should interrogate

**Composition, not injection.** `MarkdownSectionOrderer`, `MarkdownTableRowLimiter`, and `CountOutput` all operate on rendered Markdown text. Appending metadata Markdown *before* ordering makes `--count`, `--rows`, and section ordering work with zero metadata-specific knowledge. This is why each heading must be exactly the registered section name — the orderer reads it back out of the `##` line.

**Two-phase cost split.** The discovery scanner does only the cheap `Describe` (row counts); the expensive projection runs at render time for **selected tables only**, so `-S "Metadata: TypeRef"` costs one table rather than seventeen.

**`--rows` windows rendered output; the projection is not windowed.** Double-windowing would return empty for any range past the projection ceiling. `MaxRowsPerTable = 1_000_000` is therefore high enough that ranges stay reachable, and finite so a corrupt image cannot drive unbounded allocation.

**Column narrowing lives in the CLI, not the shared renderer.** Which columns a user asked for is a presentation concern the command owns, and `MetadataProjectionRenderer` is also the standalone inspector's renderer. A table none of whose columns were named narrows to its row ids rather than disappearing — with several tables selected, an unmatched name belongs to a sibling, and dropping the table would misreport a populated table as absent.

**Multi-assembly is rejected, not rendered.** Row ids are image-relative and section names carry no assembly, so a multi-assembly document would repeat `## Metadata: TypeDef` with rows silently belonging to different images.

## Corrections to claims in the issue and design doc

The issue and the design doc both state that `ExplicitOnly` is *the* render gate. **Mutation testing shows that is an overclaim.** `SectionCost.Unbounded` is a second, independently sufficient gate: flipping either one alone leaves raw tables suppressed, and only flipping **both** makes them render. The doc comment on `MetadataSections` and the design doc now name both, say they are redundant, and name the gate test *plus its limitation* — because the redundancy means that test cannot catch the loss of either mechanism alone.

## Incidental fixes

Three pre-existing bugs in the library command's effective-section cache, found while debugging why catalog-hidden sections were reaching the top-level `-D` catalog:

1. **CRLF poisoning.** `CacheEffective` wrote with `AppendLine` (CRLF on Windows) but the reader split on `'\n'` only, welding `\r` onto every cached section name. Such a name compares unequal to the registered name and so **silently escaped every name-keyed filter**. User-visible on `main` today for `Dependencies` and `SourceLink: Missing Files`.
2. **Cache key collision.** The key used the caller's path verbatim, so a relative path names different files from different working directories and one checkout's binary could serve another's cached catalog. Now `Path.GetFullPath`.
3. Cache version bumped `v18` → `v19` so poisoned entries are not read.

`--schema` now lists `@Metadata` among the category doors. That is a consequence of registering the sections in the schema, and it is a fix rather than a regression: the door is real and selectable, and `--schema` is defined as the whole catalog. The curated top-level `-D` is unchanged.

A layering violation was caught by `LayeringTests.Cli_DoesNotReferenceRawMetadataReaders` during development — the renderer had opened a `PEReader` in the CLI — and is fixed by routing through the sanctioned `AssemblyInspectionSession` seam.

## Validation

- `dotnet run --project src/dotnet-inspect.Tests -c Release` — **2411 tests, 1 failure**: `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically`, reproduced identically on a clean `origin/main` worktree and known to pass in ubuntu CI.
- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release` — 998 passed, 0 failed.
- 18 new gates in `MetadataLensTests`. Two are proven non-vacuous by mutating the product: the disclosure gate (both mechanisms must be removed before it fails) and the cache round-trip gate (reverting the CRLF fix fails it).

Merged `origin/main` at `1f03f4bc`; verified the new `ProjectionAudit` from #3405 does not fire on any metadata path and that `--count`/`--value`/`--print`/`--paths` all answer correctly.


richlander:
## Review reconciliation

Two independent adversarial reviews on the fixed head `a7a7e790`, each isolated in its own detached worktree.

### MAI-Code — one blocking finding, **upheld with different evidence**

> Effective-section cache can serve stale discovery results for a different/same-size file

**The claim is correct and I have fixed it. The submitted repro does not demonstrate it.**

MAI's repro used `-D @Metadata --schema`. That is the documented *static discovery* path — `LibraryCommand.cs` gates it explicitly (`Static discovery mode: -D --schema lists schema without resolving/loading the library`) — so it prints the full catalog without reading the file at all. Verified against a file the cache has never seen:

```text
$ dotnet-inspect library <never-seen-4KB-of-zeros.dll> -D "@Metadata" --schema --tsv
name    kind
Metadata: Assembly      section (opt-in)
...
$ dotnet-inspect library <same file> -D "@Metadata" --tsv
Error: Category '@Metadata' not found.
```

The identical output was the static path, not a stale entry. Without `--schema` the same file correctly reports no metadata sections.

The underlying defect is nonetheless real, and I found it on the path that *does* consult the cache. The cache is only read when `Discover` is empty (`options.Discover is { Length: 0 }`), so **bare `-D`** is the mode to attack; a category drill-in recomputes and would pass vacuously. Warming the entry with one assembly, then overwriting the path with a *different* assembly zero-padded to the identical length:

```text
STEP1 warm cache, Packages.dll at probe path: hash=1876206148
STEP2 same path+size, now Core.dll padded:    hash=1876206148   <-- stale
STEP3 TRUTH identical bytes, fresh path:      hash=1111680885
RESULT: STALE CACHE CONFIRMED
```

**Pre-existing, not introduced by this PR.** `main`'s key is `{assemblyPath}#{size}#sl{n}`; this branch had `{fullPath}#{size}#sl{n}`, and for the absolute path used above `GetFullPath` is the identity. Fixed here anyway because this branch already repairs two other defects in the same function, which makes it tightly coupled rather than unrelated cleanup.

**Fix** (`fb315de2`): the key now includes `LastWriteTimeUtc.Ticks`. With size, that is the same identity test the tool's other file caches use, and it costs one stat rather than a content hash on every discovery run. After the fix, STEP2 matches STEP3.

Gated by `LibraryCommand_DiscoverEffective_SameSizeReplacement_InvalidatesCache`, **proven non-vacuous**: reverting the key to path + size fails it (confirmed by mutating the product and re-running).

MAI additionally probed and found clean: disclosure gating, Markdown composition/ordering/`--count`/`--rows`, multi-assembly rejection, empty-table omission, control-character safety in structured output, and column narrowing.

### Gemini 3.1 Pro — zero findings

Probed and reported clean, with reproductions: the CRLF fix, the `GetFullPath` key change (noting `CoreCache.GetFilePath` SHA-256s the key, so illegal path characters are a non-issue), the v19 bump, column narrowing against differing cell counts and sibling-owned column names, `AugmentSchema` mutation safety and non-leakage into the curated `-D` catalog, `Metadata: Image` format parity (17 facts in both Markdown and TSV), the `MaxRowsPerTable` truncation caveat, zero-row table exclusion, and terminal-injection escaping via `EscapeText`/`EscapeCore`. Its worktree was verified unmodified at the reviewed head afterwards.

### State

- Head is now `fb315de2`.
- Full suite: **2412 tests, 1 failure** — `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically`, reproduced identically on clean `origin/main` and green in ubuntu CI.
- Both reviewers' single actionable finding is resolved. Re-review of the fixed head is warranted for the cache-key change before merge.

richlander:
## Fixed-head review round 2 (`fb315de2`) — reconciliation

Two reviews ran against `fb315de2`. They **disagreed**, and resolving the disagreement produced a better fix than either recommended.

| Reviewer | Verdict on `fb315de2` | Cache-key finding |
| --- | --- | --- |
| MAI-Code | 1 blocking | `LastWriteTimeUtc` still permits a stale hit |
| Gemini 3.1 Pro | Ready to merge | Same collision constructed, but dismissed as outside the threat model |

Both reviewers **built the same collision**. They split only on whether it mattered.

### The disagreement, and why I sided with MAI-Code

Gemini dismissed it on two grounds, one of which I was able to measure and refute:

> Conducting a full content hash (SHA256) on every `library -D` run would cripple performance, so the `Length` + `LastWriteTimeUtc` heuristic is the correct industry standard.

Measured, that is not true here:

| Assembly | Size | SHA-256 |
| --- | --- | --- |
| `DotnetInspector.Core.dll` | 54 KB | 1.65 ms |
| `dotnet-inspect.Tests.dll` | 1.3 MB | 4.14 ms |
| `dotnet-inspect.dll` | 3.4 MB | 9.76 ms |

Against a **2,421 ms** warm `-D` run on that same 3.4 MB assembly, hashing is **0.4%** of the work the cache exists to avoid. The performance argument for the heuristic does not survive measurement.

Gemini's other ground — that forging the collision needs prior write access — is a fair *security* judgment, but this is not primarily a security bug. It is a **correctness** bug, and the mundane trigger needs no attacker: a rebuild in place, a `Copy-Item` (which preserves the source's write time), or an archive/package extraction (whose recorded stamps are coarse and frequently pinned to a constant for reproducible packaging). PE files are padded to file alignment, so a small code change routinely produces a byte-different assembly of *identical* length.

I also measured the timestamp-tick hole directly rather than reasoning about it. On this repository's NTFS volume, **19 of 2000** back-to-back rewrites of a same-size file produced an identical `LastWriteTimeUtc` (minimum non-zero delta 5,364 ticks ≈ 0.54 ms).

### Fix (`d752076a`)

Rather than argue about how narrow the residual window is, I removed it. The key is now the resolved path plus a **SHA-256 content hash**, so different bytes cannot share a key **by construction**:

```
{fullPath}#{sha256}#sl{0|1}
```

Two further improvements fell out of it:

- **Fail-safe on unreadable bytes.** The key is now `string?`; when the content cannot be read the key is `null` and the cache is *bypassed* rather than keyed on a weaker identity. This also removes the TOCTOU throw Gemini noted on the old `FileInfo.Length` access (correctly classified as pre-existing, not a regression) — the caller reads the same file immediately afterwards and surfaces the failure with its own diagnostics.
- The file is opened `FileShare.ReadWrite | FileShare.Delete`, i.e. as permissively as the inspector about to read it.

### Gates, proven non-vacuous by mutation

`LibraryCommand_DiscoverEffective_PreservedWriteTimeReplacement_InvalidatesCache` is new and pins MAI-Code's exact scenario: same path, same length, write time explicitly restored.

| Key under test | Same-size gate | Preserved-write-time gate |
| --- | --- | --- |
| `path + size` | **FAIL** | **FAIL** |
| `path + size + write time` | pass | **FAIL** |
| `path + content hash` (shipped) | pass | pass |

The middle row is the point: the new gate catches what the old one structurally cannot.

While proving this I hit a methodology trap worth recording — rebuilding only `src/dotnet-inspect` leaves the test project's copied binary stale, so the first mutation run reported a **false pass**. The table above was produced after rebuilding `src/dotnet-inspect.Tests`.

### Test-soundness finding (raised independently by both reviewers)

Both flagged that the older gate's `if (smallBytes.Length >= largeBytes.Length) return;` would silently skip while reporting green if the fixture sizes ever inverted. Agreed and fixed — it is now a loud `Assert.True` naming both assemblies and sizes.

### Re-confirmed clean at `fb315de2`

Gemini re-verified cold/warm agreement for ordinary (non-metadata) library sections via `Compare-Object` (identical), the `effective-v19` bump (no stale entry reachable; orphaned keys exist only on this unmerged branch, so a `v20` bump would be churn), and `FileInfo` behavior on locked/symlinked paths. MAI-Code re-confirmed cold/warm agreement and explicit metadata rendering (`-S "Metadata: TypeRef" --rows 2 --tsv`).

Credit where due: Gemini's first pass marked this area clean, and when challenged it diagnosed its own miss precisely — it had verified the author's stated intent (`GetFullPath`) and implicitly trusted the pre-existing `size` component instead of asking whether size identifies a binary.

### Validation

- `dotnet build dotnet-inspect.slnx -c Release` — succeeded.
- `dotnet run --project src/dotnet-inspect.Tests -c Release` — **2413 tests, 1 failed**. The single failure is `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically`, reproduced identically on clean `origin/main` and green in Ubuntu CI.

Both reviewers have been sent a re-review at `d752076a`. Not marking ready until those come back clean.

richlander:
## Fixed-head review round 3 (`d752076a`) — reconciliation

Both reviewers converged on the **same** remaining issue and split on severity.

| Reviewer | Verdict | Remaining finding |
| --- | --- | --- |
| MAI-Code | Ready to merge | TOCTOU between hashing and inspection — "real, but narrow" |
| Gemini 3.1 Pro | 1 blocking | Same race — "undermines the cache identity" |

Everything else came back clean from both: null-key bypass (nonexistent path, directory, zero-byte file, `FileShare.None` holder all surfaced the real error and never success-shaped output), cold/warm agreement on ordinary sections, and the gates.

Both reviewers independently reproduced a row of the mutation matrix, and Gemini independently confirmed the performance numbers it had previously disputed (10.19 ms for the 3.32 MB assembly vs my 9.76 ms), retracting its "industry standard" framing.

### Adjudication: I sided with Gemini on severity

MAI-Code called it non-blocking because the window is narrow. That is true about the *probability* but understates the *consequence*, which is what decides severity here.

The catalog is derived from the bytes `InspectAsync` parsed; the key came from a **separate read issued after it**. If the assembly is replaced in between, the entry files the old catalog under the *replacement's* identity — so it is not a transient glitch that the next run repairs. It is a **persistently poisoned entry** that is thereafter served as a correct answer for bytes it never described, and content-hash keying alone cannot detect it. A rebuild racing a discovery run is an ordinary dev-loop event.

### Fix (`ce7c7127`)

Capture the content hash **once, before** the inspection; reuse it for the lookup and as the pre-inspection snapshot. `CacheEffective` compares it against a post-inspection hash and **declines to write** when they disagree.

This converts silent, persistent mislabelling into a recomputation on the next run. It also reduces hashing: a cold run goes from three hashes to two, a warm run from two to one.

### What this fix does *not* do

Gemini's recommendation — have `LibraryMetadataService`/`AssemblyInspectionSession` hash the stream it actually parses — is the correct complete fix, and I did **not** do it here. The hash and the parse remain two reads, so bytes that change and change back around the inspection still agree.

I deliberately kept that out of this PR: it is a metadata/service-layer change (the CLI is forbidden from opening `PEReader`/`MetadataReader` by `LayeringTests.Cli_DoesNotReferenceRawMetadataReaders`), it needs its own review, and the condition is **pre-existing** — the identical window existed when the key was path + size. Filed as **#3478** with the full analysis.

### Honesty about the gate

Per `AGENTS.md`, an asserted property must name its gate or be marked unverified. The decline path **has no deterministic gate** — the race cannot be driven reliably from a test — so `docs/design/section-model.md` marks that specific behavior unverified rather than implying coverage. I did not add a test that merely re-asserts the comparison operator and calls it a gate.

### Guarding against a silent regression

Changing the lookup gating risked a cache that silently *never hits* — which every existing test would still pass. Verified explicitly on a fresh probe assembly:

```
cache .tsv entries before=67 after=68 (delta 1)
cold=4600ms  warm=438ms
outputs identical: True
```

### Validation

- `dotnet build dotnet-inspect.slnx -c Release` — succeeded.
- `dotnet run --project src/dotnet-inspect.Tests -c Release` — **2413 tests, 1 failed**, the known pre-existing `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically` (reproduces on clean `origin/main`, green in Ubuntu CI).

Both reviewers sent to `ce7c7127`. Not marking ready until they return clean.

richlander:
## Fixed-head review round 4 (`ce7c7127`) — final reconciliation

**Both reviewers are clean at `ce7c7127`.**

| Reviewer | Verdict |
| --- | --- |
| Gemini 3.1 Pro | Ready to merge |
| MAI-Code | Ready to merge |

### Gemini 3.1 Pro — clean

Independently verified cold→warm cache behavior on **all three** discovery paths, which was my main worry (a cache that silently never hits would still pass every test):

| Path | Cold | Warm |
| --- | --- | --- |
| Plain file | 5340 ms | 428 ms |
| Platform assembly | 659 ms | 536 ms |
| Package (by id) | 2459 ms | 606 ms |

Also confirmed: the pre-inspection hash is the same value used for the lookup; `hasSourceLink` is still folded into the written key correctly; `inspectedContentHash: null` is reachable *only* when a snapshot is genuinely impossible (`IOException`/`UnauthorizedAccessException`), never through a logic gap; and both invalidation gates pass. It independently argued that **declining** to cache is the correct choice over writing under the new hash.

### MAI-Code — one finding, reproduced but misattributed

MAI-Code reported that local `.nupkg` discovery never warms the cache, as **blocking**, attributing it to the new `fullPath#contentHash` key.

I reproduced the behavior exactly, then tested the attribution. Clean `origin/main` at `d08132c3` — which predates every cache change in this PR — behaves identically:

```
baseline run 1:  | Cache | 0 hits, 1 miss |
baseline run 2:  | Cache | 0 hits, 1 miss |
baseline run 3:  | Cache | 0 hits, 1 miss |
```

The causal claim could not hold: the path component of the key is **not new**. `main`'s key was already `{assemblyPath}#{size}#sl{n}`, so a content hash cannot explain a miss that `main` also produces. The real cause — which MAI-Code itself identified in the same report — is that a local `.nupkg` is extracted to a **fresh temp directory every run**, so the path differs even for identical bytes.

The scope is also narrower than "package discovery". Package-by-id resolves into the stable global packages folder and warms fine on this head:

```
run 1:  | Cache | 1 hit, 1 miss |
run 2:  | Cache | 2 hits, 0 misses |
```

I gave MAI-Code the path to the already-built clean-`main` binary and invited it to falsify me. It ran the baseline itself, reproduced `0 hits, 1 miss` on both runs, and withdrew the finding:

> Yes — I agree the local-`.nupkg` cache miss is pre-existing rather than introduced by this PR. … No — it is not blocking for this PR. … Nothing else is outstanding in `ce7c7127`. Ready to merge.

Filed as **#3483** with the repro, the `origin/main` comparison, the scope limitation, and two candidate fixes.

### Follow-ups filed, not smuggled in

| Issue | Scope |
| --- | --- |
| #3467 | `--heap` coordinate carrier + bounded heap previews (carries the one unmet #3401 acceptance bullet) |
| #3468 | Hex table selection (`0x02` ≡ `TypeDef`) |
| #3478 | Derive the cache key from the bytes the inspection actually parsed (closes the residual TOCTOU) |
| #3483 | Local `.nupkg` discovery never warms the cache (pre-existing) |

#3478 is Gemini's recommended complete fix. I deliberately did not do it here: the CLI is barred from `PEReader`/`MetadataReader` by `LayeringTests.Cli_DoesNotReferenceRawMetadataReaders`, so it is a metadata/service-layer change deserving its own review, and the condition is pre-existing.

### Review summary across all four rounds

| Round | Head | Gemini | MAI-Code | Outcome |
| 1 | `a7a7e790` | clean | 1 blocking (stale cache) | Claim upheld, repro invalid; fixed |
| 2 | `fb315de2` | ready | 1 blocking (mtime insufficient) | Sided with MAI; replaced with content hash |
| 3 | `d752076a` | 1 blocking (TOCTOU) | non-blocking (same TOCTOU) | Sided with Gemini; added pre/post guard |
| 4 | `ce7c7127` | ready | ready | Clean |

Reviewers disagreed with each other in rounds 2 and 3, and I sided against a different one each time, on measured evidence rather than on which reviewer had been right previously.

### Validation

- `dotnet build dotnet-inspect.slnx -c Release` — succeeded.
- `dotnet run --project src/dotnet-inspect.Tests -c Release` — **2413 tests, 1 failed**: `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically`, which reproduces on clean `origin/main` and passes in Ubuntu CI.
- CI on `ce7c7127`: `build-net10`, `markdownlint`, `pack`, `test (ubuntu-26.04)`, `changes` all **pass**.
- Both review worktrees verified unmodified at `ce7c7127` afterwards.

Ready to merge

PR #3466, opened 2026-07-29
Route all console redirection through ConsoleCapture's lock

### LINKED ISSUE #3416: Suite has an order-dependent flake: WriteTips and DiffCommandTests fail intermittently, pass in isolation

richlander:
Fixes #3416.

## What was wrong

`ConsoleCapture` serializes `Console.Out`/`Console.Error` behind a semaphore, and 136 call sites across 28 files use it correctly. Three tests in `CommandLineTests` redirected `Console.Error` **directly**, without taking that lock.

`Console.Out`/`Error` are process-global, so those three raced every test using `ConsoleCapture`. Two things go wrong when they interleave:

- One side's writer is swapped out mid-test — it asserts on an empty writer, or on text another test wrote.
- Each side restores the writer *it* captured on entry, so an interleaved pair can leave the console pointing at a **disposed** `StringWriter` after both finish.

That is exactly the reported signature: a varying member of a pair fails, and everything passes in isolation.

The theory `WriteFileLayoutTips_NeverWritesTips` had a second process-global mutation — it cleared `DOTNET_INSPECT_TIPS`, which `OptionParsers.cs:45` reads — visible to any concurrently running command test.

## The fix

A `ConsoleCapture.RunAsync(Action)` overload for synchronous bodies, and the three tests moved onto it. The env-var mutation now happens inside the captured region, so the same lock covers it.

The overload is **async rather than blocking on purpose**, and I have measurements for that. I first wrote it as a blocking `_lock.Wait()`. xUnit runs tests on the thread pool, so blocking parks pool threads; with 24 `Task.Run` workers contending it starved the pool and **failed 29 unrelated tests**. Reusing the existing async path adds no new blocking and no new lock.

## Gates

Both halves of the invariant, because they fail independently:

| Gate | Covers |
| --- | --- |
| `ConcurrentCapturesDoNotObserveEachOthersOutput` | the lock actually excludes |
| `SyncAndAsyncCapturesExcludeEachOther` | the sync and async paths share one lock — the shape this bug took |
| `TestAssemblyRedirectsConsoleOnlyThroughConsoleCapture` | nothing bypasses the lock |
| `ConsoleRedirectionScanReadsRealSources` | the scan above is non-vacuous |

The structural gate derives its subject from the test sources on disk rather than a hardcoded list, so a newly added raw redirection fails it without anyone remembering to update anything. Its needles are composed via `nameof` so the scanner is not its own offender.

## Tamper evidence

Removing the lock reproduces **both** #3416 failure shapes exactly:

```text
worker 0 stdout: expected 'token-0', got ''
worker 1 stdout: expected 'token-1', got 'token-0t'...
```

The first is the `WriteTips_WithMinimalLevel_WritesTips` mode (empty writer); the second is the `DiffCommandTests` mode (stolen output). Adding a raw `Console.SetError` to `SkillCommandTests` fails the structural gate naming that file.

## Validation, and what is *not* fixed

Three full-suite runs, Release, Windows, after `dotnet build dotnet-inspect.slnx -c Release`: **2365 tests, 2 failures each run**. The `WriteTips`/`DiffCommandTests` pair named in the issue **never recurred**.

The two remaining failures are not this bug, and I am not claiming them:

1. `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically` fails **all three** runs deterministically. That is #3454, fixed separately by #3455.
2. A **second, distinct** varying pair remains — `PerformanceTriagePredicates_DoNotAlterDiscovery(library)` in runs 1 and 3, `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceLinkDoor` in run 2. Same varying-pair signature, different shared state; both are `LibraryCommand` discovery tests, so the console lock does not address them.

So this closes the console-redirection race that #3416 documents, and leaves a separate discovery-side flake that deserves its own investigation rather than being folded in here.

### DIFF
--- a/src/dotnet-inspect.Tests/CommandLineTests.cs
+++ b/src/dotnet-inspect.Tests/CommandLineTests.cs
@@ -130,37 +130,21 @@ public class CommandLineTests
     }
 
     [Fact]
-    public void WriteTips_WithQuietLevel_WritesNothing()
+    public async Task WriteTips_WithQuietLevel_WritesNothing()
     {
-        var originalErr = Console.Error;
-        var errWriter = new System.IO.StringWriter();
-        Console.SetError(errWriter);
-        try
-        {
-            Hints.WriteTips(TipLevel.Quiet, new Tip("package", "Foo", "inspect"));
-            Assert.Empty(errWriter.ToString());
-        }
-        finally
-        {
-            Console.SetError(originalErr);
-        }
+        var (_, error) = await ConsoleCapture.RunAsync(
+            () => Hints.WriteTips(TipLevel.Quiet, new Tip("package", "Foo", "inspect")));
+
+        Assert.Empty(error);
     }
 
     [Fact]
-    public void WriteTips_WithMinimalLevel_WritesTips()
+    public async Task WriteTips_WithMinimalLevel_WritesTips()
     {
-        var originalErr = Console.Error;
-        var errWriter = new System.IO.StringWriter();
-        Console.SetError(errWriter);
-        try
-        {
-            Hints.WriteTips(TipLevel.Minimal, new Tip("package", "Foo", "inspect"));
-            Assert.Contains("Tips:", errWriter.ToString());
-        }
-        finally
-        {
-            Console.SetError(originalErr);
-        }
+        var (_, error) = await ConsoleCapture.RunAsync(
+            () => Hints.WriteTips(TipLevel.Minimal, new Tip("package", "Foo", "inspect")));
+
+        Assert.Contains("Tips:", error);
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


richlander:
## Review reconciliation — MAI-Code

**One blocking finding, and it was right.** Fixed in `e5de48d1`.

> The new concurrency gates are vacuous and would not catch the original race.

### Reproduced before fixing

I deleted `await _lock.WaitAsync()` and `_lock.Release()` from `ConsoleCapture` entirely and re-ran:

```text
dotnet-inspect.Tests  Total: 4, Errors: 0, Failed: 0, Skipped: 0
```

All four gates passed with **no lock at all**. Exactly as reported.

### Why it happened

`RunAsync(Action)` delegates to a body returning `Task.FromResult(0)`, and `await _lock.WaitAsync()` on an uncontended semaphore also completes synchronously. So every `async` lambda ran start-to-finish without ever yielding, and `Task.WhenAll` received 16 already-completed tasks. There was no concurrency to serialize.

**How I let this through is the part worth stating plainly.** I *did* tamper-verify these gates — against the earlier implementation, which used a blocking `_lock.Wait()` with `Task.Run` workers. That version failed under tamper with the exact #3416 output shapes. I then rewrote the tests onto the async overload to avoid thread-pool starvation and **did not re-run the tamper check against the rewrite**. The evidence in my PR description was real but stale — it described a version of the test that no longer existed.

### The fix

Two changes, because either alone is insufficient:

1. Workers start with `Task.Run` and are released together through a `TaskCompletionSource`, so they are on distinct threads and genuinely contend.
2. Each action sleeps **before** writing. This is the subtle half: the damaging window is between `ConsoleCapture` redirecting the console and the action using it. A peer that redirects during that window steals this worker's writes. Sleeping *after* the write leaves the writer already populated and hides the bug — the failure mode reads as an empty writer, which only happens if the redirect is stolen before the write lands.

### Re-verified, in both directions

Lock removed — both gates now fail:

```text
DotnetInspector.Tests.ConsoleCaptureTests.SyncAndAsyncCapturesExcludeEachOther [FAIL]
DotnetInspector.Tests.ConsoleCaptureTests.ConcurrentCapturesDoNotObserveEachOthersOutput [FAIL]
   Total: 4, Errors: 0, Failed: 2
```

Lock restored — five consecutive runs, all green, each ~0.8-1.2s. That duration is itself corroboration: 8 workers holding a 25ms sleep serialize to roughly 200ms plus overhead, whereas the vacuous version finished in 0.3s.

### On the self-flake risk you were right to raise

The gates cannot starve the pool: waiting is `await _lock.WaitAsync()`, which yields rather than parking a thread, and only 8 workers are used. The 25ms hold is inside the lock, so exactly one worker sleeps at a time. The `CountdownEvent` rendezvous is bounded by an explicit 30s timeout rather than waiting forever.

### Your other checks

Thank you for independently confirming the base-commit diagnosis (three raw `Console.SetError` sites in `CommandLineTests.cs`, and no others) and that the structural gate fires when a raw redirection is added. That matches what I found, and the structural gate was the one part of this that was never vacuous.

### Full suite

Two runs after this fix, Release, Windows: **2365 tests, 2 failures each** — unchanged, and both are the disclosed unrelated failures (#3454 via #3455, and the separate discovery-side varying pair).

Re-review of the fixed head `e5de48d1` requested below before this is marked ready.

richlander:
## Fixed-head re-review — clean

MAI-Code re-reviewed the fixed head `e5de48d1` in an isolated worktree and found no issues, having independently verified the fix rather than taking my word for it:

- Tampered `ConsoleCapture.cs` by removing `WaitAsync`/`Release`, rebuilt, and confirmed **both** behavioral gates fail — the property the first round showed was missing.
- Restored, rebuilt, confirmed 4/4 pass.
- Ran the three converted `CommandLineTests` cases (`*WriteTips*`, `*WriteFileLayoutTips*`) — all pass.
- **Stress-ran `ConsoleCaptureTests` 15 consecutive times; all 15 green.** That was the review question I most wanted answered, since these gates now use real concurrency and a timed rendezvous — a gate that fixes an old flake by introducing a new one would be a bad trade.

It also independently confirmed the sleep-before-write reasoning: the dangerous interleaving is between the console redirect and the first write, so sleeping *after* the first write would not expose this race.

## CI

All jobs green, including `test (ubuntu-26.04, linux-x64)` in 10m1s. That matters here specifically — CI has a different core count than my machine, so it is an independent check that the concurrency gates do not flake under different scheduling.

## Review tier

Per `AGENTS.md`: test-only change, no product code, blast radius confined to the test assembly — more than trivial but not high risk, so a single MAI-Code review. Two rounds were run because the first found a blocking issue.

## Summary of what this closes

Fixes #3416's console-redirection race. Deliberately **not** claimed:

- `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically` — deterministic, #3454, fixed by #3455.
- The second varying pair — now filed as #3471 with narrowing evidence showing it is neither intra-class nor console-related.

Ready to merge

PR #3470, opened 2026-07-29
Return library scan results instead of mutating the aggregate

richlander:
## What

Five library scanners wrote their answers into the shared `LibraryInspection`
aggregate by side effect. Each now **returns** its result; the call site does the
assignment.

| Scanner | Was | Now |
| --- | --- | --- |
| `ScanExtensionMembers` | `void` + 1 setter call | `ExtensionMemberScan` |
| `ScanClassifiedMethods` | `void` + 4 field writes | `ClassifiedMethodScan` |
| `ScanCustomAttributes` | `void` + 1 setter call | `AssemblyAttributeScan` |
| `ScanTypeForwarders` | `void` + 1 field write | `FindingInspection<TypeForwarderInfo>` |
| `ScanResourceTriage` | `void` + 2 field writes | `ResourceTriageScan` |

Six library scanners already returned their results; this brings five more to that
shape, leaving four.

## Why

This is the shape problem that made the libraries unreachable from a second
consumer. Computing an extension-member census required constructing an
85-setter aggregate purely to receive the answer. A returning scanner can run
for a caller that has no aggregate at all — which is the precondition for moving
these below the CLI as typed inspection queries
(`docs/design/inspection-layers.md`).

Where a scanner produces several correlated values they travel together in one
record rather than as separate writes, because they are only meaningful as a
set: a census and the metadata order that census was projected from must not
drift apart, and the three per-classification projections come from one pass
over the same classified-method list — splitting them into separate scanners
would re-walk the assembly three more times.

`LibraryScanApply` holds the assignment into the aggregate. It is deliberately a
separate, named seam rather than inline field writes at each call site: it is
exactly the adapter that gets deleted when the queries move below the CLI.

## Behavior

**No output change.** Byte-identical CLI output across a full sweep:

- `library <dll> -v:d --tips q` on `ILInspector.Metadata.dll`: **byte-identical**
  (52,397 bytes both sides, matching SHA-256).
- Every affected section, in both `--table` and `--json`, across **all 54 built
  assemblies**: **756 comparisons, 0 mismatches.** Sections compared:
  `Extension Methods`, `Custom Attributes`, `Type Forwarders`, `Signals`,
  `P/Invoke Methods`, `Async Methods`, `Array Pool Escapes`.
- Each of those sections produced non-empty data in at least some runs, so the
  comparison is not comparing empty output to empty output: Custom Attributes
  104, Array Pool Escapes 56, Signals 44, Extension Methods 33, Async Methods
  24, P/Invoke Methods 5, Type Forwarders 2.

Both binaries were pointed at the **same** input assemblies (the base worktree's
build output) so the only variable is the CLI under test.

## One deliberate semantic difference

The old `ApplyClassifiedMethods` returned early when the classified list was
empty, and the exception paths only assigned `ClassifiedMethodInspection`. In
both cases `UnsafeMethods`, `PInvokeMethods` and `AsyncMethods` were left
holding whatever they already had. `LibraryScanApply.Apply(ClassifiedMethodScan)`
now assigns all four unconditionally, so those three become `null`.

This scanner is the sole writer of all three fields (verified by grepping every
assignment repo-wide), so the only value it can overwrite is its own from an
earlier run. `AuditSignalBuilder` guards its call on
`ClassifiedMethodInspection is null`; `ScanInfoCounts` and the `Verbosity.Detailed`
fallback do **not** — they re-run the scanner unconditionally. So the overwrite
does happen, and it lands in one of two states:

- **The re-run computes the same answer** (same scanner, same assembly,
  deterministic). Nulls overwrite nulls. Unobservable — this is the ordinary case.
- **The re-run fails where the first succeeded** (file deleted or replaced
  mid-command). Here the old code was actually *wrong*: it assigned the failure
  to `ClassifiedMethodInspection` while leaving the three stale success
  projections in place, producing a model that reported a failed scan alongside
  populated rows derived from the scan that failed. The new code clears them, so
  the failure stays visible rather than being masked by stale data — which is
  what "do not turn analysis failures into success-shaped output" asks for.

Making the result total is the point rather than an accident: a returned result
that omits fields cannot express "leave the previous value alone" without
knowing there *is* a previous value — which is precisely the aggregate coupling
being removed. The `ClassifiedMethodScan.FromInspectionOnly` factory names the
failure case where there is genuinely nothing to project.

## Not converted here

- `ScanIntegrations` / `ScanIntegrationOpportunities` — the latter reads the
  former's output from the shared model and re-runs it when absent. That is a
  real inter-scanner dependency, currently papered over by dictionary insertion
  order. It needs an explicit expression (a result parameter), which is a design
  decision about how the query layer expresses dependencies, not a mechanical
  return-type change.
- `ScanInfoCounts` and `AuditSignalBuilder.PopulateLibraryAudit` — these write
  scattered presence flags rather than a single result. Different problem.

No package files touched: `PackageCommand.cs` calls
`LibrarySections.CreateScannerRegistry()`, so the two share a registry, but the
registry's own signature is unchanged.

## Validation

`dotnet build dotnet-inspect.slnx -c Release` — succeeded, 0 errors. The 5
warnings are pre-existing in unrelated test projects and are present at the base
commit.

`dotnet run --project src/dotnet-inspect.Tests -c Release`:

| | Total | Failed | Skipped |
| --- | --- | --- | --- |
| Base (`27cd64ee`) | 2361 | 2 | 10 |
| This branch | 2361 | 2 | 10 |

The two failures are identical on both sides and pre-existing, confirmed by
building and running the base commit in a separate worktree on this machine:

- `LibraryCommand_DetailedOutput_RendersSectionsAlphabetically` — Windows-only,
  tracked by #3445, root-caused by #3454 (`MarkdownSectionOrderer` rewrites CRLF
  to LF while the test splits on `Environment.NewLine`).
- `LibraryCommand_DiscoverEffective_GroupsSourceLinkUnderSourceLinkDoor` —
  order-dependent on shared symbol-cache state; passes in isolation on this
  branch.

The 10 skips are `ildasm`/`ilasm` absent.

### DIFF
--- a/src/dotnet-inspect.Tests/LibraryFindingConsumerTests.cs
+++ b/src/dotnet-inspect.Tests/LibraryFindingConsumerTests.cs
@@ -28,10 +28,9 @@ public class LibraryFindingConsumerTests
     {
         var inspection = new LibraryInspection();
 
-        LibraryMetadataService.ScanClassifiedMethods(
+        inspection.Apply(LibraryMetadataService.ScanClassifiedMethods(
             typeof(SampleUnsafeClass).Assembly.Location,
-            inspection,
-            new VerboseLogger(enabled: false));
+            new VerboseLogger(enabled: false)));
 
         var finding = Assert.Single(
             inspection.ClassifiedMethodInspection.Findings(),
@@ -49,10 +48,9 @@ public class LibraryFindingConsumerTests
     {
         var inspection = new LibraryInspection();
 
-        LibraryMetadataService.ScanExtensionMembers(
+        inspection.Apply(LibraryMetadataService.ScanExtensionMembers(
             typeof(ExtensionsCommandTests).Assembly.Location,
-            inspection,
-            new VerboseLogger(enabled: false));
+            new VerboseLogger(enabled: false)));
 
         var finding = Assert.Single(
             inspection.ExtensionMemberInspection.Findings(),
@@ -170,10 +168,10 @@ public class LibraryFindingConsumerTests
             SwitchInspection = LibraryMetadataService.ScanSwitches(missingPath, logger),
         };
 
-        LibraryMetadataService.ScanClassifiedMethods(missingPath, inspection, logger);
-        LibraryMetadataService.ScanExtensionMembers(missingPath, inspection, logger);
-        LibraryMetadataService.ScanCustomAttributes(missingPath, inspection, logger);
-        LibraryMetadataService.ScanTypeForwarders(missingPath, inspection, logger);
+        inspection.Apply(LibraryMetadataService.ScanClassifiedMethods(missingPath, logger));
+        inspection.Apply(LibraryMetadataService.ScanExtensionMembers(missingPath, logger));
+        inspection.Apply(LibraryMetadataService.ScanCustomAttributes(missingPath, logger));
+        inspection.TypeForwarderInspection = LibraryMetadataService.ScanTypeForwarders(missingPath, logger);
         LibraryMetadataService.ScanIntegrations(missingPath, inspection, logger);
 
         AssertFailure(inspection.ClassifiedMethodInspection, MetadataFindings.ClassifiedMethodDescriptor);
--- a/src/dotnet-inspect/Inspectors/AuditSignalBuilder.cs
+++ b/src/dotnet-inspect/Inspectors/AuditSignalBuilder.cs
@@ -42,7 +42,8 @@ internal static class AuditSignalBuilder
 
             // Reuse the same session for the classified-methods scan rather than re-opening the file.
             if (inspection.ClassifiedMethodInspection is null)
-                LibraryMetadataService.ScanClassifiedMethods(session, assemblyPath, inspection, logger);
+                inspection.Apply(
+                    LibraryMetadataService.ScanClassifiedMethods(session, assemblyPath, logger));
         }
         catch (Exception ex)
         {
--- a/src/dotnet-inspect/Inspectors/LibraryMetadataService.cs
+++ b/src/dotnet-inspect/Inspectors/LibraryMetadataService.cs
@@ -163,12 +163,12 @@ internal static class LibraryMetadataService
                 try
                 {
                     using var session = AssemblyInspectionSession.Open(path);
-                    ScanExtensionMembers(session, path, inspection, logger);
-                    ScanClassifiedMethods(session, path, inspection, logger);
+                    inspection.Apply(ScanExtensionMembers(session, path, logger));
+                    inspection.Apply(ScanClassifiedMethods(session, path, logger));
                     inspection.ResourceInspection = ScanResources(session, path, logger);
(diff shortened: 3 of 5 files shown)


richlander:

**Tier:** more-than-trivial, not high risk (mechanical signature change, no
output change, no new logic) — one review.

**Reviewer:** MAI-Code, isolated in a detached review worktree at the exact head
`9d9d0fd7`. **Verdict: NO FINDINGS.**

The reviewer independently ran the build, the full test suite, and A/B output
comparisons rather than only reading the diff. It confirmed all three claims:

| Claim | Reviewer verdict | Reviewer evidence |
| --- | --- | --- |
| No observable output change | VERIFIED | Base vs. head binaries over the same inputs across the 7 affected sections, `-v:d`, and a failure-path input — 0 mismatches |
| No test regression | VERIFIED | 2361 total / 2 failed / 10 skipped, the same two known pre-existing failures |
| Null-overwrite unobservable | VERIFIED | Repo-wide search found no production writer of `UnsafeMethods` / `PInvokeMethods` / `AsyncMethods` other than this scanner's apply path |

It checked all seven named attack points (A–G), including setter/cache-reset
preservation (`ResetFindingProjectionCaches` still runs on both converted
setter paths), exception-path field fidelity, and guard/idempotence semantics.

### One correction I made to my own PR description

While reconciling I found an inaccuracy in my original write-up, which I have
now fixed in the body. I had written that "both `AuditSignalBuilder` and
`ScanInfoCounts` guard their call on `ClassifiedMethodInspection is null`". That
is true of `AuditSignalBuilder` only. `ScanInfoCounts`
(`LibraryMetadataService.cs:1531`) and the `Verbosity.Detailed` fallback
(`LibraryMetadataService.cs:167`) call the scanner **unconditionally** — the
null guard there covers `ExtensionMemberInspection`, not the classified methods.

The conclusion is unchanged but the reasoning had to be rebuilt, and it comes
out stronger. The overwrite does occur; it resolves to one of two cases:

- The re-run is the same scanner over the same assembly and is deterministic, so
  nulls overwrite nulls. Unobservable. This is the ordinary case, and it is what
  the 756-comparison sweep exercises.
- The re-run fails where the first succeeded. The **old** behavior was wrong
  here: it stored the failure in `ClassifiedMethodInspection` while leaving the
  three stale success projections populated, yielding a model that reported a
  failed scan next to rows derived from the scan that failed. The new code
  clears them, keeping the failure visible instead of masking it with stale
  data.

I am flagging this because the error was mine and it was load-bearing for the
"unobservable" argument, not because the reviewer raised it — it did not.

### Not addressed here (deliberate, non-blocking follow-up)

- `ScanIntegrations` / `ScanIntegrationOpportunities`: the latter reads the
  former's result off the shared model and re-runs it when absent. Converting
  it requires deciding how the query layer expresses query-to-query
  dependencies, which is a design decision rather than a mechanical change.
- `ScanInfoCounts` / `PopulateLibraryAudit`: these write scattered presence
  flags rather than a single result — a different shape problem.

Ready to merge

Issue #3459, opened 2026-07-29
Decompiler: fold object initializers used as call arguments when the enclosing call's receiver/arguments are spilled between newobj and the member stores (#3336 follow-on)


richlander:
## Summary

`ObjectInitializerPass` still leaves a distinct residue class fully lowered: an object initializer whose **use-site is a call argument**, where Roslyn spills the enclosing call's **receiver (and/or a later argument) between the `newobj` and the initializer's member store**. This is *not* the member-value spill resolved by #3336 (stages 1–3, merged in #3368 and #3451) — there the interleaved statement produced a *member value*; here the interleaved statements produce the *enclosing call's* receiver and arguments, and the folded initializer must land as a call argument rather than a `return`/store-local.

The `ForeignReadBeforeMembers_StaysLowered` close-negative deliberately declines a foreign read interleaved before the members, so these correctly stay lowered today — folding them safely needs foreign-receiver/argument-hoist reasoning (prove the interleaved receiver/argument spills can be reordered relative to the member stores without changing observable side-effect order), which is a separate capability from #3336's linear member-value-spill fold.

### Witness — `Azure.Data.Tables.TableClient.Create` (7.20.1 → azure.data.tables 12.11.0)

```bash
dnx dotnet-inspect -y -- member Azure.Data.Tables.TableClient Create --all \
  --library <cache>/azure.data.tables/12.11.0/lib/net8.0/Azure.Data.Tables.dll -S "Decompiled Source"
```

Current output (residue):

```csharp
public virtual Response<TableItem> Create(CancellationToken cancellationToken = default)
{
    // ...
    TableProperties S_256;
    TableRestClient S_257;
    // ...
    S_256 = new();
    S_257 = _tableOperations;      // <-- interleaved foreign receiver spill (for the enclosing Create call)
    S_256.TableName = Name;        // <-- initializer member store, AFTER the receiver spill
    V_2 = default;                 // <-- interleaved argument spill
    response = S_257.Create(S_256, V_2, _defaultQueryOptions, cancellationToken);  // <-- initializer use-site is a call argument
    // ...
}
```

Should fold to:

```csharp
response = _tableOperations.Create(
    new TableProperties { TableName = Name }, default, _defaultQueryOptions, cancellationToken);
```

[...]

PR #3474, opened 2026-07-29
Fold object initializer used as a call argument (#3459)

richlander:
- Fixes/advances #3459
- Folds an object initializer used as a **call argument** back to `new T { ... }` and inlines the call's reorder-safe receiver/argument spills so the render is byte-Exact
- Card revision: `6abf09ee9`

## Change

**Conclusion:** **PASS** — the Azure.Data.Tables `TableClient.Create` residue folds to the canonical `new T { ... }` call argument and recompiles byte-for-byte to the original IL (no valid/correct-but-opcode-divergent trap).

### Benchmark target

Benchmark target: `Azure.Data.Tables@12.11.0`

dotnet-inspect command:

```bash
dotnet-inspect member TableClient Create --package Azure.Data.Tables -S "Decompiled Source"
```

### Original source

```csharp
{
    scope.Start();
    try
    {
            null,
            _defaultQueryOptions,
            cancellationToken);
    }
    catch (Exception ex)
    {
        scope.Failed(ex);
        throw;
    }
}
```

### Before

```csharp
{
    Nullable<ResponseFormat> V_2;
    Response<TableItem> V_3;
    TableProperties S_256;
    TableRestClient S_257;

    {
        scope.Start();
        try
        {
            S_256 = new();
            S_257 = _tableOperations;
            S_256.TableName = Name;
            V_2 = default;
            response = S_257.Create(S_256, V_2, _defaultQueryOptions, cancellationToken);
            V_3 = Response.FromValue<TableItem>(response.get_Value(), response.GetRawResponse());
        }
        catch (Exception ex)
        {
            scope.Failed(ex);
            throw;
        }
    }
    return V_3;
}
```

- Valid: True
- Correct: True
- IL fidelity: False — the stackifier materializes the stack-resident receiver (`S_257 = _tableOperations`) and `default` argument (`V_2 = default`) as named locals; recompiling them emits extra `stloc`/`ldloc` and hoists `initobj` to the method top, diverging from the stack-only original. This is the #3127 trap: Valid + Correct yet not opcode-faithful.
- Taste applied: None
- Commit: `d08132c3d`

### After

```csharp
{
    Response<TableItem> V_3;

    {
        scope.Start();
        try
        {
            V_3 = Response.FromValue<TableItem>(response.get_Value(), response.GetRawResponse());
        }
        catch (Exception ex)
        {
            scope.Failed(ex);
            throw;
        }
    }
    return V_3;
}
```

- Valid: True
- Correct: True
- IL fidelity: True (byte-Exact) — the reorder-safe receiver and `default` spills are inlined back into the call, restoring the canonical `_tableOperations.Create(new T { ... }, default(T?), _opts, ct)`, which recompiles opcode-for-opcode to the original. Verified on the standalone witness `--fidelity-check` (see Evidence).
- Taste applied: None
- Commit: `6abf09ee9`

The `default(Nullable<ResponseFormat>)` explicit-type spelling (vs source `null`) is byte-Exact; a target-typed `default`/`null` printer choice is a separate, opcode-neutral concern.

### Residual differences from the original

The After decompilation is fully raised **for this PR's claim** — the call argument is the canonical `new T { ... }` and the render is byte-Exact. Two differences from the original source remain, neither owned by this change:

- **`Response<TableItem> V_3;` and the trailing `return V_3;`.** csc's return-accumulator lowering. Fixed by #3553, which sinks the `return` back inside the `try` for exactly this `using` + rethrowing-`catch` shape. That PR is independent of this one; with both in, the accumulator is gone.
- **`ResponseWithHeaders<TableResponse, TableCreateHeaders> response;` hoisted above the `using`,** where the original spells `var response = _tableOperations.Create(...)` at the assignment site. This is **not** #3553's doing and is not fixed by it. `CSharpPrinter.CollectDeclaringStores` merges a declaration into its store only when that store is a statement in the **entry block** — its doc comment says so explicitly ("sits at statement level in the entry block — the current emitter's merged-declaration shape"). `response`'s store is nested inside the `using`/`try`, so it fails that test and the local falls back to a bare up-front declaration.

  The scoping predicate needed to do better, `LocalReferencesStayInBlockAfterStore`, already exists in the printer but is currently wired only for `ref` locals (CS8174) and `for`-initializers. Extending it to the general case is the decl-sinking work tracked by #3165 and #3071.

  The difference is cosmetic, not semantic: the declaration is byte-neutral, since a C# local declaration emits no IL of its own.

## Taste oracle (dotnet/runtime)

Per the AGENTS taste-raise rule, both facets were consulted for this object-initializer raise:

- **Declared:** `dotnet_style_object_initializer = true:suggestion` (`dotnet/runtime` `.editorconfig`, line 97) — object-initializer syntax is the preferred form.
- **Revealed:** an object initializer passed directly as a call argument, `Throws<NotSupportedException>(new HttpRequestMessage { Version = new Version(0, 42) });` in `src/libraries/System.Net.Http/tests/FunctionalTests/SocketsHttpHandlerTest.cs:5859` (`dotnet/runtime` @ `ba10a6ee`).

The fold is byte-neutral, so no byte-divergent style lens is recorded in `-S "Applied Taste"` (verdict: None).

## Evidence

| Check | Baseline | Head |
| --- | --- | --- |
| Focused tests | `ObjectInitializerPassTests`: 38 passed | `ObjectInitializerPassTests`: 41 passed (3 new) |
| Focused tests | `RaisingPassTests`: 109 passed | `RaisingPassTests`: 109 passed |
| Decompiler fast suite | 4322, 1 failed (pre-existing) | 4322, 1 failed (pre-existing) |
| Reduced fixture fidelity | `W.Client::Create` opcode-divergent | `W.Client::Create` byte-Exact |
| Real witness | `TableClient::Create` lowered residue | `TableClient::Create` folded `new T { ... }` call arg |

The 1 fast-suite failure is `ExpressionTreeLambdaTests.LookalikeExpressionAssembly_StaysFactoryCalls`, a pre-existing environmental miss (needs `System.Linq.Expressions.dll` from the ExpressionTreeSpoof fixture via a full `dotnet build dotnet-inspect.slnx -c Release`) — identical on Baseline and Head, unrelated to this change.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- -class ILInspector.Decompiler.Tests.ObjectInitializerPassTests
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- --gate fast
```

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -6225,3 +6225,46 @@ public static class BraceBodyWrappingSamples
             FourthMeasuredProjection = fourth,
         };
 }
+
+// #3459: an object initializer used as a CALL ARGUMENT, where the enclosing call's
+// receiver (a non-volatile instance field off `this`) and a `default` struct
+// argument are the compiler's pure spills sitting on the stack beneath the dup
+// chain — the Azure.Data.Tables `TableClient.Create` shape. The importer materializes
+// them as `S = _rest;` and `V = default;` around the member store, which #3336's
+// member-value fold could not cross. The pass now skips those reorder-safe spills,
+// folds the construction into the call-argument position, and inlines each single-use
+// spill back into its operand, restoring the canonical stack-only spelling
+// `_rest.Create(new CallArgTarget { Name = Label }, default(CallArgFlag?), _options)`
+// — which recompiles byte-for-byte to the original IL. Appended at end of file so
+// these top-level types cannot shift any existing generated-code ordinal.
+public sealed class CallArgTarget
+{
+    public string? Name { get; set; }
+}
+
+public enum CallArgFlag { A, B }
+
+public sealed class CallArgRest
+{
+    public int Create(CallArgTarget target, CallArgFlag? flag, object options) => target.Name?.Length ?? 0;
+}
+
+public sealed class CallArgClient
+{
+    readonly CallArgRest _rest = new();
+    readonly object _options = new();
+    volatile CallArgRest _volatileRest = new();
+    public string Label = "n";
+
+    // Foldable: the receiver `_rest` (pure field-off-this) and the `default` struct
+    // argument are reorder-safe spills, so both are inlined into the folded call.
+    public int CreateViaField()
+        => _rest.Create(new CallArgTarget { Name = Label }, default, _options);
+
+    // Close negative for the inlining guard: a VOLATILE field receiver must NOT be
+    // inlined (reordering a volatile access is observable). The initializer still
+    // folds — the receiver ran before the `newobj`, an offset-guarded skip — but the
+    // volatile spill is left in place rather than hoisted into the call receiver.
+    public int CreateViaVolatileField()
+        => _volatileRest.Create(new CallArgTarget { Name = Label }, default, _options);
+}
--- a/src/ILInspector.Decompiler.Tests/ObjectInitializerPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ObjectInitializerPassTests.cs
@@ -727,6 +727,55 @@ public class ObjectInitializerPassTests
         function.CheckInvariant();
     }
 
+    // #3459 blocking negative (constructor crossing): a receiver field read spilled
+    // before the members is only reorder-safe to inline as the folded receiver when
+    // the construction cannot observe or mutate what it reads. A non-parameterless
+    // constructor receives a reference (here `this`) and may mutate `this.Rest`, and
+    // the use-site fold evaluates the receiver `this.Rest` before that constructor —
+    // reordering the read across the mutation. The parameterless-ctor gate must block
+    // admitting the receiver spill as a reorder-safe skip, so the dangerous use-site
+    // spelling `this.Rest.Consume(new InitTarget(this) { ... })` never appears.
+    [Fact]
+    public void ReceiverSpillWithArgumentedCtor_IsNotFoldedAcrossConstructor()
+    {
+        var function = FunctionWithReceiverSpillAndArgumentedCtor();
+
+        new ObjectInitializerPass().Run(function, PassContext.None);
+
+        // The receiver read survives as a spill store (never inlined as the folded
+        // receiver), and no initializer is hoisted into the call ahead of it.
+        Assert.Single(function.Descendants.OfType<StoreStackSlot>(), store => store.Slot == 257);
+        Assert.DoesNotContain(".Consume(new", CSharpPrinter.Print(function).Output);
+        function.CheckInvariant();
+    }
+
+    // #3459 blocking negative (argument-position soundness): a mutable-memory read
+    // spilled before the members is admitted as a reorder-safe skip, but inlining it
+    // is sound only when its load evaluates BEFORE the folded initializer in the
+    // enclosing call. Here `this.Field` feeds the SECOND argument while the
+    // initializer feeds the first, so C# evaluates the initializer (and its
+    // side-effecting member value `Mutate()`, which may write `this.Field`) first,
+    // then the second argument — inlining `this.Field` there would observe the
+    // post-mutation value. The construction still folds at the use site, but the
+    // read must stay a spill statement in its original pre-call position.
+    [Fact]
+    public void MutableReadSpillFeedingLaterArgument_IsNotInlined()
+    {
+        var function = FunctionWithMutableReadSpillFeedingLaterArgument();
+
+        new ObjectInitializerPass().Run(function, PassContext.None);
+
+        // The construction still folds into an initializer at the first argument.
+        var initializer = Assert.Single(function.Descendants.OfType<ObjectInitializerExpression>());
+        Assert.Equal(["X"], initializer.Members);
+
+        // But the mutable read (feeding the later argument) is NOT inlined: its spill
+        // store survives so the read keeps its original pre-call position.
+        Assert.Single(function.Descendants.OfType<StoreStackSlot>(), store => store.Slot == 257);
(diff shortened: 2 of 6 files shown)


richlander:
## Adversarial review reconciliation — fixed head `5866cbc25`

This change is high-risk (subtle correctness + hostile-input reasoning on a fold/inline that rewrites user-visible C#), so it took the two-model cross-family tier. Reviews were run in isolated linked worktrees pinned to the exact head, scratch confined to `/tmp`, no git mutations in review trees.

### Round 4 (head `5866cbc25`'s parent chain, pre-fix)

**Gemini (gemini-3.1-pro): CLEAR.** Validated the direct-`Object` lemma, `.cctor` scan, struct/generic instantiation, EH-region checks, and wiring. Additionally declared the deferred round-3 **named-local-hoist finding Roslyn-unreachable** — a standalone `StoreLocal`/`initobj` never interleaves in a Roslyn object-initializer dup-chain — so no SSA guard is needed under the chosen threat model.

**GPT (gpt-5.6-sol): 2 findings.**

- **Finding 1 — arg/local reads not gated (a ctor could mutate a caller arg/local via an escaped unsafe pointer, `ldarga; conv.u; stsfld` + `stind.ref`): DISMISSED with rationale.** Reaching a caller arg/local slot from a parameterless ctor requires the caller to have previously escaped that address to shared storage, which Roslyn never emits without explicit `unsafe` pointer code. This is the same accepted arbitrary-IL boundary as the null-`this` case, and the threat model for this feature is explicitly **Roslyn-faithful, not arbitrary-IL-sound**. **Both models concur** with the dismissal. The boundary is now documented inline on the `LoadArgument`/`LoadLocal` arms and on `MethodRef.ConstructorEffectFree`.

- **Finding 2 — `IsObjectConstructor` accepted a cross-assembly lookalike `System.Object`: REAL hole, FIXED.** The base-derivation check inferred core-library identity from the base-ctor call token's namespace+name *spelling* and never consulted `ResolutionScope`, so a `[Evil]System.Object` (a `MemberReference`→`TypeReference` merely spelled `System.Object`) passed. This violated the repo rule "typed identity over display text." Fixed in `5866cbc25` by anchoring the lemma on the pipeline's typed core-library identity: `MemberIdentity.IsCoreLibraryType(source.ResolveBaseType(ctor.DeclaringType), "System", "Object")`. `IsObjectConstructor` is now documented as a necessary *shape* check only, not an identity proof.

### Round 5 (verification at fixed head `5866cbc25`, post-fix) — both CLEAR

**GPT: CLEAR.** Confirmed `[Evil]System.Object` retains assembly identity `Evil` and is rejected by `IsCoreLibraryType`; a Roslyn unsafe-source probe kept the receiver *before* `newobj` so no reorder-safe spill formed (no Finding-1 counterexample reproduced); `TableClient.Create` still folds; 46/46 focused tests pass.

**Gemini: CLEAR.** Traced the fix soundness to `TypeRefDecoder`/`CanonicalReferenced`: `TypeRef.CoreLibrary` identity is granted **iff** the `AssemblyReference` carries a trusted platform public-key token verified by `PlatformKeys.IsPlatform`, which an attacker cannot forge. Verified type-forwarding, `ModuleRef`, and `TypeSpec` (`ELEMENT_TYPE_OBJECT`) edge cases all resolve correctly, and confirmed the real `TableClient.Create` witness still hoists to `new TableProperties { TableName = Name }` (no over-decline).

### Resolution

- Finding 2 → fixed in `5866cbc25`.
- Finding 1 → accepted arbitrary-IL boundary (both models concur), documented inline.
- Named-local-hoist (deferred round-3) → Roslyn-unreachable (Gemini), no code change.

Both required fixed-head reviews are clean. Validation on the fixed head: `ObjectInitializerPassTests` 46/46; `--gate fast` 4327/1 (the sole miss is the pre-existing environment-dependent `LookalikeExpressionAssembly_StaysFactoryCalls`, which requires `ilasm`/`ildasm` not present here — unrelated to this change); real `TableClient.Create`/`TableServiceClient.CreateTable` fold byte-exact.

richlander:
Ready to merge

richlander:
## Head moved: `origin/main` integrated (no content change)

New head: `0581d998a` (was `5866cbc25`, the head both round-5 reviews cleared).

The only new commit is the merge of `origin/main` (through `15153c5fe`). It is a **no-op for this PR's own files** — the merge brought in main's changes and nothing else:

```console
$ git diff --name-only origin/main...HEAD
src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
src/ILInspector.Decompiler.Tests/ObjectInitializerPassTests.cs
src/ILInspector.Decompiler/Pipeline/ConstructorConfinementFacts.cs
src/ILInspector.Decompiler/Pipeline/Ir/IrImporter.cs
src/ILInspector.Decompiler/Pipeline/Ir/IrNodes.cs
src/ILInspector.Decompiler/Pipeline/Passes/ObjectInitializerPass.cs

$ git diff 5866cbc25 0581d998a -- <those six files>
(empty)
```

So the reviewed content is byte-identical at the new head; the round-5 CLEAR verdicts from both models carry over as confirmation rather than needing a second full pass.

Re-validated at `0581d998a`:

- `dotnet build dotnet-inspect.slnx -c Release` — 0 errors (5 pre-existing warnings in `ILInspector.Metadata.Tests`, unrelated).
- `ObjectInitializerPassTests` — **46/46**.
- `--gate fast` — **4327 total, 0 failed, 0 skipped** (the previously environment-dependent `LookalikeExpressionAssembly_StaysFactoryCalls` miss is gone; this run is fully clean).
- Real witness still folds: `_tableOperations.Create(new TableProperties { TableName = Name }, default(Nullable<ResponseFormat>), _defaultQueryOptions, cancellationToken)`.

Still ready to merge.

PR #3488, opened 2026-07-29
Prune dot directories from the packable-project census (#3422 follow-up)

richlander:
## Summary

Simplify the #3450 nested-checkout prune to a dot-directory prune, per review
feedback: dot directories never hold source, so skipping the class is both
simpler and sufficient.

#3450 probed each directory for a `.git` marker -- a file for a linked
worktree, a directory for a clone or submodule. It works, but it costs a
filesystem probe per directory and a two-case marker rule to buy generality
this repository does not need.

## Change

`PackagingSurfaceTests.IsExcludedDirectory` now skips any directory whose
name starts with `.`, alongside the existing `bin` / `obj` /
`artifacts` / `node_modules` list. `.git` drops out of that list as
redundant. No filesystem probe remains.

This still covers the #3422 failure: a linked worktree under `.worktrees/`
contributes a second, legitimately packable
`src/dotnet-inspect/dotnet-inspect.csproj`, which the census would otherwise
report as an extra shipping project. It additionally covers other tool state
such as `.claude` and `.vs`.

## Boundary

The rule is by **name**, so a nested checkout parked under an ordinary
directory name (for example `vendor/other`) is no longer pruned. That is a
deliberate narrowing relative to #3450 and is stated in the
`EnumerateRepositoryProjects` doc comment rather than left implicit. The
`AGENTS.md` worktree convention is `.worktrees`, which the rule covers.

## Evidence

`dotnet run --project src/dotnet-inspect.Tests -c Release -- -class
"DotnetInspector.Tests.PackagingSurfaceTests"` -> **Total: 3, Failed: 0** on
Windows Release.

**Tamper-verified, not merely green.** Deleting the `StartsWith('.')` term
and rebuilding fails
`ProjectScanSkipsDotDirectoriesAndBuildOutput` at `PackagingSurfaceTests.cs:106`
(`Assert.DoesNotContain() Failure: Item found in collection` -- the
`.worktrees` project). Restoring the term returns 3/3.

The gate was updated to match the new rule: the `vendor/other` nested-clone
case is replaced by a `.claude` case (a dot directory that is not the
repository's own convention), and `scratch/copy` is kept as the negative
case proving the prune keys on the leading dot rather than on depth.

## Review tier

Trivial-to-low risk: a test-only helper, name-based, no product code, with a
tamper-verified hermetic gate and no behavior change to the census itself.

### DIFF
--- a/src/dotnet-inspect.Tests/PackagingSurfaceTests.cs
+++ b/src/dotnet-inspect.Tests/PackagingSurfaceTests.cs
@@ -71,7 +71,7 @@ public sealed class PackagingSurfaceTests
     }
 
     /// <summary>
-    /// The gate for the nested-checkout prune in <see cref="EnumerateRepositoryProjects"/>
+    /// The gate for the dot-directory prune in <see cref="EnumerateRepositoryProjects"/>
     /// (#3422). Without it, <see cref="OnlyTheShippingToolsOptIntoPackaging"/> fails for any
     /// contributor whose linked worktree lives under the repository root — a failure that
     /// reads as a regression in whatever change is under test, and that CI never sees because
@@ -79,7 +79,7 @@ public sealed class PackagingSurfaceTests
     /// machine running it happens to have a worktree present.
     /// </summary>
     [Fact]
-    public void ProjectScanSkipsNestedCheckoutsAndBuildOutput()
+    public void ProjectScanSkipsDotDirectoriesAndBuildOutput()
     {
         string temp = Path.Combine(Path.GetTempPath(), $"packaging-scan-{Guid.NewGuid():N}");
         try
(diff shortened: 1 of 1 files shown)


richlander:

Tier: **more than trivial, not high risk** (test-only helper, name-based rule,
no product code, narrows an exclusion rule). One review, MAI-Code, per
`AGENTS.md`.

**MAI-Code (fixed head `8db302a5`, isolated review worktree): clean.** All
seven attack points answered with commands and observed output. Two findings
worth surfacing because they strengthen the evidence beyond what I had run
myself:

- **Per-assertion non-vacuity, not just whole-gate.** I had tamper-verified
  only the `StartsWith('.')` term against the `.worktrees` assertion. The
  reviewer tampered `IsExcludedDirectory` to exclude *only* `.worktrees`
  and got a failure at `PackagingSurfaceTests.cs:108` on the `.claude/scratch`
  case -- so that case is load-bearing and not redundant coverage. Tampering
  further to also exclude `scratch` failed at
  `PackagingSurfaceTests.cs:111` on the ordinary-nested case, so the negative
  case genuinely pins that the prune keys on the leading dot rather than on
  depth. Every assertion in the gate is independently non-vacuous.
- **The coverage narrowing is currently vacuous in practice.** `git submodule
  status` is empty and `.gitmodules` does not exist, and a tree scan found no
  ordinary-name nested checkout. So the generality dropped relative to #3450 is
  not generality this repository is presently using. The boundary remains
  documented in the `EnumerateRepositoryProjects` doc comment rather than
  left implicit, in case that changes.

Also verified by the reviewer: the `bin` / `obj` / `artifacts` /
`node_modules` deny list is still live and not subsumed by the dot rule;
`Path.GetFileName` returns the leaf name for every path
`Directory.EnumerateDirectories` yields into the walk, so the empty-string
concern does not arise; and no reference anywhere in the tree survives to the
old test name `ProjectScanSkipsNestedCheckoutsAndBuildOutput`.

No blocking findings, no non-blocking findings, nothing dismissed -- so there
are no resolution commits. Head under review is the head proposed for merge.

CI: `test (ubuntu-26.04, linux-x64)` pass. `mergeStateStatus: CLEAN`.

Ready to merge

Issue #3493, opened 2026-07-29
GeneratedFixtureCatalogTests: parameter-attributes RecompileFail and two ReturnToSender record fragment losses


richlander:
## Summary

`GeneratedFixtureCatalogTests` fails 3 of 18 tests. There are two distinct
defects behind them.

## Defect 1 — `minimal.parameter-attributes` no longer recompiles

Fails `MinimalCompileBackRungs_ReportExpectedOutcomesByFixtureId` and
`CompilerLoweringFrontier_IsSelectableButNotInDefaultRun` (both render the same
ladder, so one root cause shows up twice):

```
FAIL  minimal.parameter-attributes  GeneratedFixtures.MinimalParameterAttributes.Class1::Update#0
      decompiler=Full  compile-back=RecompileFail  expected-compile-back=Exact
```

Every other row in the 41-fixture / 119-target ladder passes.

## Defect 2 — `ReturnToSender` record helpers lost expected body fragments

Fails `ReturnToSenderRecordCatalog_CoversGeneratedRecordHelpers`: 2 of 7
fixtures, 2 of 12 targets, both in bucket `target-body-fragment-missing`.

```
record.equality-operators
  RecordEqualityOperatorsRow::op_Equality#0   rts=Exact  bucket=target-body-fragment-missing
  detail: missing expected target body fragment: (object)left == (object)right

record.virtual-helpers
  RecordVirtualHelpersRow::ToString#0         rts=Exact  bucket=target-body-fragment-missing
  detail: missing expected target body fragment: new StringBuilder()
```

Note both report `rts=Exact` — the IL round-trips, but the emitted **source
text** no longer contains the fragment the catalog expects, so this is a
spelling/shape regression rather than an IL one.

One caveat worth checking before treating the `il-diff` rows as signal: the
diffs printed alongside these are **assembly-scope-only**, e.g.
`[GeneratedDecompilerFixtures]` vs `[return-to-sender]`, and
`[System.Runtime]System.Text.StringBuilder` vs
`[System.Private.CoreLib]System.Text.StringBuilder` (type forwarding). Those
look like normalization gaps, not behavior changes, and may be noise relative to
the `detail:` line that actually drives the failure.

## Why this was not caught

Class-level `[Trait("Speed", "Slow")]`
(`src/ILInspector.Decompiler.Tests/GeneratedFixtureCatalogTests.cs:16`), and
every pre-merge lane runs `-trait- "Speed=Slow"`
[...]

Issue #3492, opened 2026-07-29
Validity census: SwapIdiomPass::Mat* added a new CS0161 to the pinned missing-return population


richlander:
## Summary

`ValidityCoverageReportingTests.DecompilerAssembly_MissingReturnPopulationIsPinned`
fails: the pinned CS0161 ("not all code paths return a value") census over the
decompiler's own assembly has **gained one entry**. The Release pin expects 5
methods; the actual population is 6.

## Evidence

```
Assert.Equal() Failure: Collections differ
Expected: [..., "ILInspector.Decompiler.Pipeline.IndexFromEndPass::..."]
Actual:   [..., "ILInspector.Decompiler.Pipeline.IndexFromEndPass::...",
                "ILInspector.Decompiler.Pipeline.SwapIdiomPass::Mat..."]
                                                                    ^ (pos 5)
```

The new entry is `ILInspector.Decompiler.Pipeline.SwapIdiomPass::Mat…`
(truncated by xUnit's collection formatter). Candidates in that file are
`MatchPlace` (`SwapIdiomPass.cs:148`), `MatchStore` (`:157`), and `MatchCarrier`
(`:217`) — the repro below prints the exact name.

This means the decompiler now emits a body with a missing return for one more of
its *own* methods than the pin allows. Unlike the other gate failures this is a
**census**, so the whole-assembly sweep is intentional and correct.

Pin: `src/ILInspector.Decompiler.Tests/ValidityCoverageReportingTests.cs:68-73`
(Release arm), assertion at `:62`. The surrounding comment block records the
#2959 / #2973 / #3017 history of this population and should be extended with
whatever explains this entry.

## Why this was not caught

The failing test carries a method-level `[Trait("Speed", "Slow")]`, and every
pre-merge lane runs `-trait- "Speed=Slow"` (`.github/workflows/ci.yml:301`). The
lanes that do include it (`release.yml`, weekly Deep Inspect) were exceeding
their job timeout, and Deep Inspect's failure notifier is gated on `failure()`,
which a *cancelled* job does not satisfy. See #3432.

## Repro

```bash
dotnet build dotnet-inspect.slnx -c Release
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
  -class ILInspector.Decompiler.Tests.ValidityCoverageReportingTests
```

Note this pin is `#if DEBUG`-conditional and the two arms differ (DEBUG expects
[...]

Issue #3491, opened 2026-07-29
Lowered fidelity gate compares Roslyn local-function name ordinals, which no decompiler can match (5 of 6 diffs); one real diff needs docketing


richlander:
> [!IMPORTANT]
> **Diagnosis below is superseded.** Traced in
> [this comment](https://github.com/richlander/dotnet-inspect/issues/3491#issuecomment-5123310485).
> There is **no local-function lowering defect**, and the enum-argument lead is
> spurious. All five local-function entries have **byte-identical opcode streams**;
> the sole difference is the Roslyn name-mangling ordinal in
> `<Enclosing>g__Name|<methodOrdinal>_<K>`, which encodes the enclosing method's
> *positional index in its type*. Compile-back reconstructs `CfgSampleClass` in a
> different member order, so the ordinal cannot match — for any decompiler.
>
> Pass/fail here is coincidental: both ordinals move whenever anyone adds a method to
> the fixture file. The fix is a fourth `IlBodyDiffNormalization` flag folding
> `|N_K` to `|*_K`; it retires the five new entries **plus** the two already docketed
> (`DoubleViaLocalFunction`, `StaticLocalFunctionCalledTwice`) in both dockets — a net
> docket *reduction*.
>
> The sixth diff, `ManualConstantOnlyComparisonFactory`, is unrelated and real: it
> belongs to the already-docketed stack-slot/temporary reshaping family
> (`LoweredFidelityGateTests.cs:106-111,119-121`), whose five siblings are docketed.
>
> Original report retained below for the CI evidence.

---

## Summary

`LoweredFidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket` fails with **six**
new lowered fidelity diffs beyond the docket. Five of the six involve local
functions, which suggests a single root cause in local-function lowering rather
than six independent defects.

## New diffs

| Method | Local function involved |
| --- | --- |
| `EnumArgInInlineLocalFunction` | yes |
| `EnumArgInLocalFunctionWithLocal` | yes |
| `RecursiveLocalFunction` | yes |
| `StaticLocalFunctionWithLocal` | yes |
| `TwoLocalFunctionQuadrants` | yes |
| `ManualConstantOnlyComparisonFactory` | no |

Two of the five also involve an **enum-typed argument**
(`EnumArgInInlineLocalFunction`, `EnumArgInLocalFunctionWithLocal`), which may
narrow it further.

Full current diff set (45 entries) is in the assertion output of the run linked
[...]

Issue #3490, opened 2026-07-29
Fidelity gate: MakeConsumerWithTwoLeadingArgs has a new raised opcode diff (spurious stloc/ldloc)


richlander:
## Summary

`FidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket` fails:
`CfgSampleClass.MakeConsumerWithTwoLeadingArgs` produces a new `OpcodeDiff` that
is not in the raised gate's `KnownDiffs` docket. The recompiled body round-trips
the object-initializer receiver through a local instead of keeping it on the
stack.

## Evidence

```
MakeConsumerWithTwoLeadingArgs [OpcodeDiff]
  original  : ldarg call ldarg call newobj dup ldarg callvirt dup ldarg callvirt newobj ret
  recompiled: ldarg call ldarg call stloc ldloc newobj dup ldarg callvirt dup ldarg callvirt newobj ret
```

The inserted `stloc`/`ldloc` pair is the whole delta; opcode families otherwise
match in order.

Fixture: `src/ILInspector.Decompiler.Tests/CfgSampleClass.cs:2981`

## Note on scope

This method is *already* a documented known diff in the **lowered** gate
(`LoweredFidelityGateTests.cs:161`, attributed to #3272, "the trailing object
initializer ..."), but it is not in the **raised** gate's docket. So either the
raised path has regressed to match a shape only the lowered path was expected to
show, or the raised docket needs the same entry with a rationale. Worth deciding
which before pinning it.

## Why this was not caught

Class-level `[Trait("Speed", "Slow")]` (`FidelityGateTests.cs:14`); every
pre-merge lane runs `-trait- "Speed=Slow"` (`.github/workflows/ci.yml:301`). The
lanes that do include it (`release.yml`, weekly Deep Inspect) were exceeding
their job timeout, and Deep Inspect's failure notifier is gated on `failure()`,
which a *cancelled* job does not satisfy. See #3432.

## Repro

```bash
dotnet build dotnet-inspect.slnx -c Release
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
  -class ILInspector.Decompiler.Tests.FidelityGateTests
```

Observed on CI run 30440377772 (ubuntu-26.04, Release).

richlander:
## Triage at `e1b3cac9`: the raised gate now has **three** new diffs, not one

[...]

PR #3495, opened 2026-07-29
Give FidelityCheck.Evaluate a type filter: 7 CI timeouts to 0, 56m to 22m

richlander:
Seven decompiler fidelity gates were timing out at 45 minutes each in CI. This change gives `FidelityCheck.Evaluate(string, …)` an optional type filter and threads it to the seven test call sites that were already filtering their results *after* paying the full cost. Measured on the real workflow, the lane goes from **56m34s wall with 7 timeouts** to **22m05s wall with 0 timeouts**, and the serialized cost of all 22 classes is now ~44.8 min — finite, and small enough to gate PRs on.

## Root cause

`FidelityCheck.Evaluate(string path, …)` had no cap and no filter. It walked every `reader.TypeDefinitions` sequentially, calling `EvaluateType` — which decompiles the type and Roslyn-recompiles it — on each one.

Seven test classes passed it **their own test assembly** and then filtered the results to a single fixture type:

```csharp
var results = FidelityCheck.Evaluate(TestAssemblyPath)
    .Where(r => r.Type == FixtureType);   // after decompiling all 1,823 types
```

The test assembly measures **1,823 TypeDefs / 17,858 methods**, so each gate paid the whole assembly to look at one type. The cost is O(types in the test assembly) *per gate*, and that assembly grows with every test class anyone adds — so these gates got slower with every unrelated PR until several crossed the timeout cliff together.

The guards already existed on the sibling overload: `Evaluate(IReadOnlyList<string>, …)` has `perAssemblyCap`, `attemptCap`, and `Parallel.ForEach`. This path simply never got them.

## The change

`Evaluate` gains a trailing `Func<string, bool>? typeFilter = null` on all four `string`-path overloads, forwarded through `EvaluateType` → `CollectType` → `CollectTypeEntries`. The check sits immediately after `string fullType = reader.GetFullTypeName(typeDef);` — metadata-only work above it, render and recompile below it.

Two more instances of the same shape are fixed:

- `CollectibleFullTypeNames` rendered all 17,858 methods just to collect type *names* (`NestedTargetLookupTests:24`).
- The documented `CB_TYPE` "focus one type" developer switch was applied in `RunType` *after* `CollectType` had already rendered everything (`FidelityCheck.cs:2321` vs `:2317`). It is now hoisted into the `CollectType` filter, so the switch does what its documentation says.

## Why the result set is unchanged

By construction. Every converted call site **keeps** its existing `.Where`/`.Single` predicate and *adds* the filter admitting the same names, so the filter can only remove results the assertion already discarded.

The retained types' results are also byte-identical, for two independent reasons:

- On the normal paths, `BuildUnit` still emits the **whole module** — all 1,823 stubbed types — regardless of the filter (`FidelityCheck.cs:2592`), so a retained type's compile input does not change. `ClusterMode.ForceAll` is the exception: it prunes via `includeRoots` rather than emitting whole-module. Result preservation still holds there, because that closure is derived from complete metadata and is independent of `typeFilter` — but the whole-module property is not universal, and `ClusterCaptureTests:40` exercises the `ForceAll` path.
- `Renderer` is a lambda (`function => CSharpPrinter.PrintRaised(function, method => IrImporter.Import(source, method), options)`) over `reader`/`pe`/`references`. It is not strictly pure: `MetadataSource` holds lazy caches and `FidelityCheck` memoizes around `:3494`–`:3512`. Those are full-metadata or per-type deterministic memoization, not cross-type accumulators, so `EvaluateType` remains self-contained per type — verified empirically by matching cold filtered probes against warmed unfiltered results.

One call site needs care and is commented in place: `NestedTargetLookupTests.CorpusSweep_StillSkipsNestedTypes` asserts `DoesNotContain(r.Type == Inner)`, so both call sites deliberately admit **both** `NestedTargetFixture` and `NestedTargetFixture.Inner` rather than narrowing to `Outer`. Review established that the assertion is in fact guarded upstream of the filter — `CollectType` rejects nested types via `if (!typeDef.GetDeclaringType().IsNil) return null;` before the filter runs — so admitting `Inner` is inert rather than load-bearing. It is retained so the filter cannot become the reason the assertion passes.

## Follow-up recorded

The `Func<string, bool>` predicate is a new instance of the display-string-matching pattern that `docs/design/type-spelling-identity-display.md` names as an anti-pattern. It is defensible here because this is *selection* (the predicate admits a superset; the pre-existing `.Where` still does real selection), not identity — but it should gain a zero-match guard so a mistyped filter fails loudly instead of passing vacuously. That guard, and a type-count budget for the unfiltered path that would have caught the original growth-to-timeout drift, are filed as #3504. The type-representation question this exposes is filed separately as #3498.

## Evidence

Real workflow runs, one runner per test class, `timeout` used so an overrun is a recordable verdict rather than a lost job:

| | before (run 30425586012) | after (run 30440377772) |
| --- | --- | --- |
| wall clock | 56m34s | **22m05s** |
| classes timing out at 45m | **7** | **0** |
| serialized cost of all 22 classes | unbounded | ~44.8 min |

`NestedTargetLookupTests` is the extreme case: **45m timeout → 4.5s**.

Per-class after the fix: SkeletonEmit 630s, ClusterCapture 376s, LoweredFidelityGate 325s, PrinterPrecedence 324s, FidelityGate 302s, RangeFromGetSubArray 196s, ByteNeutrality 136s, LongLiteralFold 75s, ReturnToSenderPrototype 68s, TypeBind 59s, RuntimeAsyncAwaiter 53s, ValidityCoverageReporting 42s, GeneratedFixtureCatalog 30s, CorpusSweep 22s, LadderIterator 10s, NestedTargetLookup 4.5s, DiffFixtureFidelity 4.3s, ReturnToSenderEvidence 3.9s, MemberBodyProducerUnion 3.3s, AuthoredRebuildFidelity 2.2s, CompilerFeatureOptions 1.4s, SubstrateLeaderDifferential 0.8s.

For the `CB_TYPE` fix, the harness's own `--fidelity-timings` on the README's `CB_TYPE=CfgSampleClass` command now attributes the residual 503s run as emit 388.8s (77%), skeleton emit 56.4s, parse 37.7s, collect/render 7.1s — i.e. the collect/render phase this PR addresses is no longer the cost, and the remaining work is emit-bound.

## This surfaces pre-existing failures; it does not cause them

Running to completion instead of timing out reveals **five real decompiler regressions** that had been landing unnoticed while these gates never finished. They are filed, not fixed here:

- #3489 — `TupleRest` RecompileFail / CS0246
- #3490 — `MakeConsumerWithTwoLeadingArgs` opcode diff
- #3491 — 6 lowered diffs, 5 of them local-function
- #3492 — `SwapIdiomPass` CS0161 census
- #3493 — `GeneratedFixtureCatalogTests`, 3 failures

These are pre-existing, and provably not caused by this change: both methods in #3489/#3490 are defined on `CfgSampleClass`, which the original `.Where(r => r.Type == FixtureType)` predicate already admitted — so they were in the evaluated set before and after.

Separately, the PR fast lane (`-trait- "Speed=Slow"`) fails 4 tests **on Windows** in `StringSwitchRaisingTests` and `CorpusSensorComparisonTests` — all four are CRLF-vs-LF assertion mismatches. Verified pre-existing by running those two classes on a clean detached worktree at `origin/main` (3d0f7efd): identical 4 failures. CI is green because it runs on Linux. Untouched by this PR; filed separately as a Windows-developer papercut.

## Expected follow-on

Merging this will make the weekly Deep Inspect "Test lane" flip from silently **cancelled** at the 6h job timeout to genuinely **failed**. That is the intended outcome, not a regression: a cancelled job does not satisfy `if: failure()`, which is why `deep-inspect.yml:577`'s `Report scheduled failure` step has never fired despite consecutive 6h overruns. Once it fails properly, it will start opening tracking issues.

## Validation

- `dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -trait- "Speed=Slow"` — 4319 tests, 4 failures, all four pre-existing on `origin/main` and Windows-line-ending-only (control run above).
- All 7 converted gate classes run to completion locally and in CI, versus 7 timeouts before.

## Non-goals

- Not fixing #3489–#3493.
- Not wiring a path-gated decompiler PR lane (now feasible at ~9 min, worth a follow-up).
- Not optimizing the emit-bound residual. `EvaluateGrouped` still falls back to a per-method whole-module compile, each re-emitting all 1,823 stubbed types; `ClusterMode`/`CompileOneClustered` and `BuildUnit`'s `includeRoots` pruning already exist but only the cluster path uses them. Note `SkeletonEmitTests` deliberately asserts the *whole-module* skeleton compiles, so that path must stay.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/ClusterCaptureTests.cs
+++ b/src/ILInspector.Decompiler.Tests/ClusterCaptureTests.cs
@@ -34,10 +34,10 @@ public class ClusterCaptureTests
     {
         string assembly = typeof(CfgSampleClass).Assembly.Location;
 
-        var whole = FidelityCheck.Evaluate(assembly, lowered: false, FidelityCheck.ClusterMode.Off)
+        var whole = FidelityCheck.Evaluate(assembly, lowered: false, FidelityCheck.ClusterMode.Off, type => type == FixtureType)
             .Where(r => r.Type == FixtureType)
             .ToDictionary(r => (r.Method, r.Overload), r => r.Status);
-        var forced = FidelityCheck.Evaluate(assembly, lowered: false, FidelityCheck.ClusterMode.ForceAll)
+        var forced = FidelityCheck.Evaluate(assembly, lowered: false, FidelityCheck.ClusterMode.ForceAll, type => type == FixtureType)
             .Where(r => r.Type == FixtureType)
             .ToList();
 
--- a/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
@@ -488,7 +488,7 @@ public class FidelityGateTests
     static readonly Lazy<IReadOnlyList<FidelityCheck.CompileBackResult>> Results = new(() =>
     {
         var assembly = typeof(CfgSampleClass).Assembly.Location;
-        return FidelityCheck.Evaluate(assembly)
+        return FidelityCheck.Evaluate(assembly, type => type == FixtureType)
             .Where(r => r.Type == FixtureType)
             .ToList();
     });
--- a/src/ILInspector.Decompiler.Tests/LoweredFidelityGateTests.cs
+++ b/src/ILInspector.Decompiler.Tests/LoweredFidelityGateTests.cs
@@ -228,7 +228,7 @@ public class LoweredFidelityGateTests
     static readonly Lazy<IReadOnlyList<FidelityCheck.CompileBackResult>> Results = new(() =>
     {
         var assembly = typeof(CfgSampleClass).Assembly.Location;
-        return FidelityCheck.Evaluate(assembly, lowered: true)
+        return FidelityCheck.Evaluate(assembly, lowered: true, type => type == FixtureType)
             .Where(r => r.Type == FixtureType)
             .ToList();
     });
--- a/src/ILInspector.Decompiler.Tests/NestedTargetLookupTests.cs
+++ b/src/ILInspector.Decompiler.Tests/NestedTargetLookupTests.cs
@@ -21,7 +21,10 @@ public class NestedTargetLookupTests
     [Fact]
     public void TargetedPath_ReachesNestedTypeIdentity()
     {
-        var collectible = FidelityCheck.CollectibleFullTypeNames(FixtureAssembly);
+        // Admits both names so the assertions below still discriminate: the
+        // point of the test is that Inner is reachable, not that it was asked for.
+        var collectible = FidelityCheck.CollectibleFullTypeNames(
+            FixtureAssembly, type => type == Outer || type == Inner);
 
         Assert.Contains(Outer, collectible);
         // The fix: the nested type is now part of the identity surface a delta
@@ -36,7 +39,9 @@ public class NestedTargetLookupTests
         // Preserve existing --fidelity-check behavior: the non-targeted corpus
         // sweep roots at top-level types only, so a nested type's methods never
         // enter the broad sweep even though the targeted path can now reach them.
-        var swept = FidelityCheck.Evaluate(FixtureAssembly);
+        // The filter admits both the outer and nested names, so the nested type's
+        // absence below is CollectType's top-level rooting -- not the filter.
+        var swept = FidelityCheck.Evaluate(FixtureAssembly, type => type == Outer || type == Inner);
 
         Assert.Contains(swept, r => r.Type == Outer && r.Method == "OuterAdd");
         Assert.DoesNotContain(swept, r => r.Type == Inner);
--- a/src/ILInspector.Decompiler.Tests/PrinterPrecedenceTests.cs
+++ b/src/ILInspector.Decompiler.Tests/PrinterPrecedenceTests.cs
@@ -48,7 +48,9 @@ public class PrinterPrecedenceTests
     public void StoreElement_CompoundArrayReceiver_RecompilesExactly()
     {
         var result = Assert.Single(
-            FidelityCheck.Evaluate(typeof(CfgSampleClass).Assembly.Location),
+            FidelityCheck.Evaluate(
+                typeof(CfgSampleClass).Assembly.Location,
+                type => type == typeof(CfgSampleClass).FullName),
             r => r.Type == typeof(CfgSampleClass).FullName
                 && r.Method == nameof(CfgSampleClass.ConditionalArrayElementStore));
 
--- a/src/ILInspector.Decompiler.Tests/RangeFromGetSubArrayPassTests.cs
+++ b/src/ILInspector.Decompiler.Tests/RangeFromGetSubArrayPassTests.cs
@@ -246,7 +246,9 @@ public class RangeFromGetSubArrayPassTests
             nameof(CfgSampleClass.ArrayRangeCompoundBoth),
         };
 
-        var results = FidelityCheck.Evaluate(typeof(CfgSampleClass).Assembly.Location)
+        var results = FidelityCheck.Evaluate(
+                typeof(CfgSampleClass).Assembly.Location,
+                type => type == typeof(CfgSampleClass).FullName)
             .Where(r => r.Type == typeof(CfgSampleClass).FullName && methods.Contains(r.Method))
             .ToList();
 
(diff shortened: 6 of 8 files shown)


richlander:

Two fixed-head reviews on `2a13105f` (base `3d0f7efd`), each in its own isolated detached worktree, different model families. **Both clean — no blocking findings.**

### Findings I acted on

Both are corrections to *my PR description*, not to the code. GPT refuted two of the three legs of my "result-preserving by construction" argument **as worded**, while confirming result preservation itself. The body has been corrected.

**Leg 2 — "`BuildUnit` still emits the whole module regardless of the filter" was too broad.** True on the normal paths (`FidelityCheck.cs:2592`), but `ClusterMode.ForceAll` prunes via `includeRoots` instead. Result preservation still holds, because that closure derives from complete metadata and is independent of `typeFilter` — but the whole-module property is not universal, and `ClusterCaptureTests:40` exercises `ForceAll`. Credit: GPT.

**Leg 3 — "`Renderer` is a pure lambda over immutable state" was wrong.** `MetadataSource` holds lazy mutable caches, and `FidelityCheck` memoizes around `:3494`–`:3512`. These are full-metadata or per-type deterministic memoization rather than cross-type accumulators, so the conclusion (`EvaluateType` is self-contained per type) survives — but the stated reason did not. GPT verified the conclusion empirically by matching cold filtered probes against warmed unfiltered results. Credit: GPT.

**Vacuity guard is upstream of my filter, not provided by it.** I described admitting `NestedTargetFixture.Inner` as what keeps `CorpusSweep_StillSkipsNestedTypes` non-vacuous. In fact `CollectType` rejects nested types via `if (!typeDef.GetDeclaringType().IsNil) return null;` *before* the filter runs, so the assertion is guaranteed by top-level rooting and admitting `Inner` is inert. Retained anyway, so the filter cannot become the reason the assertion passes. Credit: Gemini, independently corroborated by GPT.

### Verified and dismissed

- **Leg 1, filter/predicate equivalence — VERIFIED by both.** Every filter admits exactly the type the pre-existing predicate retained. Both traced `TypeResolver` to confirm `GetFullTypeName` spells nested types `Outer.Inner`, matching the hardcoded constants.
- **Filter placement** — only metadata reads sit above the predicate; nothing semantically required is skipped.
- **Null-default safety** — `null` admits everything at every level, and the filter is forwarded through every added parameter.
- **Overload resolution** — no ambiguity and no changed positional binding; the `lowered: bool` flag distinguishes the footprints.
- **Nested enumeration** — recursion into `GetNestedTypes()` continues even when the outer type is filtered out; no subtree is pruned.
- **`CB_TYPE` hoist** — identical ordinal substring semantics before and after.
- **#3489–#3493 are pre-existing** — independently confirmed by both. The affected methods remain within `CfgSampleClass`, which the original `.Where(r => r.Type == FixtureType)` predicate already admitted; #3492's path is untouched; #3493 uses the unchanged default-null path.

### Evidence quality, stated plainly

The two reviews were not equal. GPT ran real tests: `NestedTargetLookupTests` (7 passed), a targeted `SkeletonCompilesPastExplicitImplAndConstEnum`, filtered-vs-unfiltered probes on a small fixture matching at 73 results, and reproduced the known 3/18 `GeneratedFixtureCatalogTests` failures. Gemini built the solution and traced code but **ran no tests**, so its verification is compile-plus-reading only. The empirical burden is carried by GPT's run plus my own: the full fast lane (4319 tests) and all seven converted gate classes to completion, locally and in CI.

### Non-actions

No code changes resulted from either review. The zero-match guard both reviews implicitly probed — a mistyped filter yielding an empty set that a negative assertion passes vacuously — is real but is **not** introduced by this PR, and is recorded as follow-up rather than fixed here, along with the type-representation question, in #3498.

### Expected after merge

The weekly Deep Inspect "Test lane" will flip from silently **cancelled** at the 6h job timeout to genuinely **failed**, and will begin opening tracking issues for #3489–#3493. That is the intended outcome: a cancelled job does not satisfy `if: failure()`, which is why `deep-inspect.yml:577`'s notifier has never fired.

CI green on `2a13105f`: `test (ubuntu-26.04, linux-x64)` passed in 8m59s.

*Posted after merge — the PR description has been corrected in place so the merge record does not carry the two overclaims above.*

richlander:
Correction to the reconciliation above: the deferred guards are filed as #3504 (zero-match guard plus a type-count budget for the unfiltered path), not #3498. #3498 owns the separate question of what shape the type predicate should take.

Issue #3468, opened 2026-07-29
Metadata lens: hex table selection (0x02 as an alias for TypeDef)


richlander:
Split out of #3401, which shipped the table lens in #3465 but not the hex table-selection spelling.

Both spellings should address the same table:

```bash
dotnet-inspect library My.dll -S "Metadata: TypeDef"
dotnet-inspect library My.dll -S "Metadata: 0x02"
```

The tool already prints hex tokens throughout its output (`token 0x02000015`, `TypeRef[16]`), so users will paste them. Requiring a mental hex-to-name translation for a value the tool itself just emitted is a needless step.

## Acceptance

- `-S "Metadata: 0x02"` selects the same section as `-S "Metadata: TypeDef"`.
- The **rendered heading is the canonical name**, not the hex spelling. Two spellings must not produce two different-looking sections, or the section orderer and `--count` would see them as distinct.
- `-D` continues to list only the canonical names; the hex form is an input alias, not a second catalog entry.
- An out-of-range or unprojected table index gives a typed rejection naming the projected tables, not an empty section.

## Notes

`MetadataSectionNames.TryGetTable` is the single lookup point today and is where the alias belongs. Section names are derived from `MetadataTableProjector.ProjectedTables`, so the alias table should be derived the same way rather than restated -- an unprojected index must not become selectable by its hex spelling.

Design: `docs/design/metadata-table-projection.md`.
Tracking: #3282. Parent: #3401. Lands on: #3465.

PR #3497, opened 2026-07-29
Add the --heap coordinate carrier and per-heap listing sections

### LINKED ISSUE #3467: Metadata lens: --heap coordinate carrier and bounded heap previews
Split out of #3401, which shipped the table lens in #3465 but not the heap coordinate carrier.

Heap **addressing** is the one place the `@Metadata` lens introduces a new currency. A heap coordinate such as `#Strings:0x1a4` is not a section name, so it needs a carrier. `--heap` is that carrier, behaving like `--il-offset`: it makes a coordinate-scoped section available and discoverable only when present. See the coordinate-carrier family in `docs/design/output-shapes.md`.

```bash
# bounded heap preview -- just a section
dotnet-inspect library My.dll -S "Metadata: #Strings"

# one address -- a coordinate
dotnet-inspect library My.dll --heap "#Strings:0x1a4"
```

## Acceptance

- `--heap "#Strings:0x1a4"` resolves one address and renders its value.
- `-D` lists the `--heap` coordinate section **only when the carrier is present**. This is the one acceptance bullet of #3401 that #3465 does not meet.
- Bounded, escaped previews for the string/blob/user-string/guid heaps as ordinary sections, never dumped by default -- they are the largest amplification surface in the projection.
- Heap enumeration is explicitly **out of scope**: SRM exposes no public way to walk a heap. The surface is overview plus random access, and must say so rather than faking a walk by scanning bytes.

## Notes

The library-side primitive already exists: `AssemblyInspectionSession.MetadataHeapValue(...)`. `mdi` already parses this coordinate form -- roughly 90 lines of coordinate parsers are `mdi`-only today and are the natural thing to promote into the product.

Design: `docs/design/metadata-table-projection.md` (Surface section, which marks this as designed-not-implemented).
Tracking: #3282. Parent: #3401. Lands on: #3465.

richlander:
Closes #3467. Carries the one #3401 acceptance bullet that #3465 did not meet, so **#3401 can close once #3468 lands**.

A heap coordinate such as `#Strings:0x1a4` is not a section name, so it needs a carrier. `--heap` is that carrier and behaves like `--il-offset`: it makes `Metadata: Heap` exist and be discoverable exactly when a coordinate is present. Four per-heap listing sections join it.

```console
$ dotnet-inspect library My.dll --heap "#Strings:0x1a4"
## Metadata: Heap

| Heap | Address | Length | Truncated | Value |
| ---- | ------- | ------ | --------- | ----- |
| #Strings | 420 | 8 | no | ToString |

$ dotnet-inspect library My.dll -D @Metadata --tsv | grep Heap    # nothing
$ dotnet-inspect library My.dll -D @Metadata --heap "#Strings:1" --tsv | grep Heap
Metadata: Heap
```

## Heaps are not tables, and the listings say so

A table has a row count and fixed-width rows. `#Strings` and `#Blob` are length-prefixed items with no index, and `System.Reflection.Metadata` exposes no walker — only random access from an address a table cell already holds. I checked this directly rather than assuming it: there is no `StringHandles`/`UserStringHandles` collection, only `GetUserString(UserStringHandle)`.

So "bounded heap preview" cannot mean what the issue's wording suggests without faking a walk by scanning bytes, which [metadata-table-projection.md](https://github.com/richlander/dotnet-inspect/blob/main/docs/design/metadata-table-projection.md) explicitly forbids. Instead each heap is listed by the strongest honest means it admits, and `MetadataHeapCoverage` makes **which one** part of the answer:

| Heap | Coverage | What is listed |
| --- | --- | --- |
| `#GUID` | `Complete` | Every entry. Records are a fixed 16 bytes, so the count is `size / 16` and each is read by index. |
| `#Strings`, `#Blob` | `ReferencedOnly` | The distinct values projected table rows point at, address-ordered, with a `Refs` count. |
| `#US` | `NotEnumerable` | Nothing. |

Every listing renders its coverage as a caveat, so a referenced-values listing never reads as a walk of the heap:

```console
$ dotnet-inspect library My.dll -S "Metadata: #US"
## Metadata: #US

| Address | Length | Truncated | Refs | Value |
| ------- | ------ | --------- | ---- | ----- |

No entry of the #US heap can be listed: no metadata table column points into it — its
references are ldstr operands inside method bodies, which this projection does not read —
and it cannot be walked. Its 65812 bytes are not empty; read a known address with
--heap "#US:<address>".
```

`#US` deliberately **keeps** its section rather than being hidden. The section is how a caller learns the heap exists, how large it is, and that `--heap "#US:<addr>"` still reads any address in it. Hiding it would be undiscoverable; rendering the empty table without the caveat would report a 65 KB heap as empty.

## Notes for review

- **Reference scanning ignores the caller's table filter** (`options with { Tables = default }`). An entry is referenced by the *image*, not by whichever subset of tables the caller is looking at, so honoring the filter would drop entries and undercount references while looking complete. A row window is a stated bound, so it *is* honored and reported as `RowsTruncated`.
- **Hex is opt-in, never inferred.** A bare `1a4` is rejected rather than read as `0x1a4`, because guessing would silently address a different entry than the one a dump printed. The coordinate splits on the **last** colon and validates each half separately, so the diagnostic names the half that is wrong.
- **The discovery cache is disabled when `--heap` is present**, exactly as it is for `--il-offset`: the coordinate changes which sections exist, so a catalog cached without it would omit `Metadata: Heap`, and one cached with it would offer a section that no longer resolves. All three `WriteEffectiveSections` call sites.
- **Selection semantics mirror `--il-offset`.** Reached through `@Metadata`, `Metadata: Heap` is silently dropped when there is no coordinate — a category selection asks for whatever applies. Named *exactly*, it is an error naming the missing carrier.
- **`mdi` behavior change:** the ~40-line coordinate parser is promoted into `DotnetInspector.MetadataRendering` so both tools share one grammar, and heap names now render as ECMA-335 stream names (`#Strings`, not `String`). That makes a rendered cell paste back as a `--heap` coordinate. Four rendering-test expectations updated.

## Evidence

New gates in `MetadataLensTests.cs`: coordinate rendering, spelling/radix equivalence through the command surface, the discoverability property, the without-carrier rejection *and* its category-door negative case, three malformed-coordinate diagnostics, the three coverage behaviors, and heap listings joining the `-v:d` / `-S @All` disclosure gate.

`MetadataLens_HeapSection_IsDiscoverableOnlyWithItsCoordinate` asserts both directions in one test because either alone passes trivially. Measured non-vacuous by mutation, rebuilding the test project each time:

| Mutation | Result |
| --- | --- |
| `IsApplicable = static model => true` | FAIL — `Assert.DoesNotContain() Failure: Item found in collection` |
| `IsApplicable = static model => false` | FAIL — `Assert.Contains() Failure: Item not found in collection` |
| restored | PASS |

Suites, Release, rebased onto `e1b3cac9`:

| Suite | Result |
| --- | --- |
| `src/dotnet-inspect.Tests` | 2476 total, 2 failed |
| `tests/ILInspector.Metadata.Tests` | 998, 0 failed |
| `tests/DotnetInspector.MetadataRendering.Tests` | 111, 0 failed |

The two failures are `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown(readmePath: "images/logo.png.")` and `Layout_Count_CountsFilesRatherThanRenderedTreeLines`. Both are **pre-existing**: confirmed reproducing in a clean detached worktree at unmodified `origin/main` (`3d0f7efd`), deterministically and in isolation. Windows-only; they pass in ubuntu CI.

## Docs

`docs/design/metadata-table-projection.md` moves the heap carrier from "designed, not yet implemented" to implemented and gains a **Heaps are not tables** section recording the coverage decision. `docs/design/output-shapes.md` promotes `--heap` from "the designed second" carrier currency to the second. `README.md` gains a **Heaps** subsection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ context for copied DLLs. A future `--deps` source can represent runtime
 | Source mapping | `type`/`library`/`package -S "Source Files"`, `member -S "Source Locations"` / `"Original Source"` | SourceLink URLs, member file/line locations, source fetching, URL verification, token+IL-offset to source-line resolution. |
 | Performance analysis *(experimental)* | `library -S @Performance` (kind sections: `"Performance: Boxing"`, `"Performance: Arrays"`, …), `type`/`member -S "Performance Triage"`, `"Top Leverage"`, `"Resource Triage"`, `"Call Graph"` | Whole-assembly call-graph leverage ranking — direct callers, root reach, fanout, depth, loop calls — with opt-in per-node cost signals (alloc, copy, unsafe, reflection, throw/exception, catch/finally), actionable rewrite-shape detection, and exception-path resource-lifecycle candidates. |
 | Decompiler *(experimental)* | `member -S @Source` (`Decompiled Source`, `Annotated Source`, `Original Source`, `Source Diff`, `IL`); `member -S "Fidelity Causes"` | Raises method bodies to C#, interleaves IL and hidden-fact annotations, diffs SourceLink-backed source against decompiled source, and exposes typed `DEC####` fidelity causes rather than emitting plausible-but-wrong source. |
-| Raw metadata | `library -S @Metadata` (table sections: `"Metadata: TypeDef"`, `"Metadata: MethodDef"`, …, plus `"Metadata: Image"`) | The ECMA-335 metadata tables of an assembly, with handles resolved to the rows they point at and heap offsets to their values. Opt-in only: the tables are unbounded, so no verbosity renders them. |
+| Raw metadata | `library -S @Metadata` (table sections: `"Metadata: TypeDef"`, `"Metadata: MethodDef"`, …, plus `"Metadata: Image"`, the heap sections, and `--heap "#Strings:0x1a4"`) | The ECMA-335 metadata tables of an assembly, with handles resolved to the rows they point at and heap offsets to their values. Opt-in only: the tables are unbounded, so no verbosity renders them. |
 | Agent-friendly output | global flags | Markdown by default, compact `--table`, normalized `--tsv`, `--jsonl`, `--plaintext`, `--json`, Mermaid diagrams, section/field projection, `--count`, table row limiting, built-in head/tail limiting. |
 
 ## Command inventory
@@ -377,6 +377,31 @@ dotnet-inspect library MyLib.dll -S "Metadata: TypeRef" --rows 20
 dotnet-inspect library MyLib.dll -S "Metadata: TypeRef" --columns Name --tsv
 dotnet-inspect library MyLib.dll -S "Metadata: MethodDef" --tsv --rows 100..199
 ```
+
+### Heaps
+
+The four ECMA-335 heaps get sections of their own (`Metadata: #Strings`,
+`Metadata: #Blob`, `Metadata: #GUID`, `Metadata: #US`), and a single address is
+addressable with the `--heap` coordinate carrier:
+
+```bash
+dotnet-inspect library MyLib.dll -S "Metadata: #Strings" --rows 20
+dotnet-inspect library MyLib.dll --heap "#Strings:0x1a4"
+```
+
+A heap name may be written with or without its `#`, and an address is decimal
+unless it carries an explicit `0x` — a bare `1a4` is rejected rather than
+guessed at.
+
+A heap is not a table, and the listings say so rather than implying a
+completeness they cannot deliver. `#Strings` and `#Blob` are length-prefixed
+byte soup with no index, so what a listing shows is the distinct values the
+projected table rows point at, address-ordered, with a `Refs` count. `#GUID` is
+listed completely, because its entries are fixed 16-byte records. `#US` lists
+nothing at all — no table column points into it, since its references are
+`ldstr` operands inside method bodies — but it keeps its section so its size
+stays visible and any address in it stays readable with `--heap`. Every listing
+renders its coverage as a caveat.
 
 ## Output and querying
 
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -217,8 +217,8 @@ session-lifetime rule in
 
 ## Surface — the `metadata` table lens
 
-**Status: implemented**, except the `--heap` coordinate carrier noted below. The
-commands and outputs below are the shipping surface.
+**Status: implemented.** The commands and outputs below are the shipping
+surface.
 
 Each metadata table is **one section**, and the tables together form a section
 category, `@Metadata`, registered the same way `@Performance` is:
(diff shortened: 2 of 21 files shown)


richlander:

Two independent adversarial reviews on `3cdccc48`, in isolated worktrees: **MAI-Code** and **Gemini 3.1 Pro**. Four findings, all actioned. Fixed head is now **`7ffcd32c`**.

### 1. Hex addresses above `0x7FFFFFFF` wrapped to negative — *found independently by both reviewers* (fixed in `51059d0e`)

`NumberStyles.AllowHexSpecifier` on a **signed** `int` wraps rather than overflows, so `int.TryParse` returned `true` with a negative value:

```console
$ dotnet-inspect library X.dll --heap "#Strings:0x80000000" --tsv
#Strings	-2147483648		no	!malformed: ... address ('-2147483648') must be a non-negative value.
exit 0
```

This is precisely the silent mis-addressing the "hex is opt-in, never inferred" rule exists to prevent, and I had missed it while writing that rule. Hex now parses through `uint` and is range-checked. Both reviewers reproduced it; MAI-Code additionally showed the `0xFFFFFFFF` → `-1` case.

Gated by `TryParseHeapLocation_NamesTheHalfThatIsWrong` (`0x80000000`, `0xFFFFFFFF`, `0x100000000`, `2147483648`) plus `TryParseHeapLocation_AcceptsTheLargestRepresentableAddress`, which pins the boundary so the rejection cannot overshoot `int.MaxValue`.

### 2. An unresolvable coordinate exited 0 — *escalated from finding 1* (fixed in `51059d0e`)

Chasing the exit code above surfaced a bigger one the reviewers had accepted. Gemini explicitly **ruled this in scope as correct** ("conforms to the projection safety rules"), and I disagree — respectfully, and with a precedent:

```console
$ dotnet-inspect library X.dll --il-offset 0x06FFFFFF+0x0
Error: Could not resolve member context for token 0x6FFFFFF.
exit 1

$ dotnet-inspect library X.dll --heap "#Strings:999999999"      # before
| #Strings | 999999999 |  | no | !malformed: ... past the end of a 230090-byte heap. |
exit 0
```

The two `!malformed` cases look alike but are not. A bad heap reference found **inside a projected table row** is a fact about the image — render it, exit 0. A **coordinate** is the caller's own input naming exactly one thing that does not exist; rendering that as a successful row exits 0 while answering nothing, and `-D` went on advertising `Metadata: Heap` as available. This PR's stated claim is that `--heap` "behaves like `--il-offset`", so it now does. Exit 1 in both the render and `-D` paths.

### 3. **The cache *read* sites were unguarded** — Gemini, High (fixed in `7ffcd32c`)

The best finding of the round, and a real defect I shipped. I guarded the three `WriteEffectiveSections` **write** sites with `!HasHeapCoordinate(options)` and never touched the three `TryGetCachedEffective` **read** guards above them. Gemini found it by inspection; I reproduced it:

```console
$ dotnet-inspect library X.dll -D --tsv     # prime
$ dotnet-inspect library X.dll -D --tsv     # warm
$ dotnet-inspect library X.dll -D --heap "#Strings:999999999" --tsv
name	kind
@Audit	category
...
exit 0        # stale catalog, and the coordinate was never resolved at all
```

Note this also **defeated fix 2**: a warm cache short-circuits the run before `PopulateMetadataHeapIfRequested` ever executes. Fixed at all three read sites, plus the matching `FilterEffective` line its IL-coordinate sibling already had.

`MetadataLens_CachedDiscovery_DoesNotBypassHeapCoordinateResolution` primes twice so it exercises a cache **hit** rather than the cold path. Measured non-vacuous: reverting only the read-site guard fails it (`Assert.Equal() Failure: Values differ`); restored, it passes.

### 4. A stray colon blamed the correct half — Gemini, Low (fixed in `7ffcd32c`)

`TryParse` split on the **last** colon, so `#Strings:0x1a4:0x1b4` reported `unknown heap '#Strings:0x1a4'` — naming the one half the caller got right. Gemini's suggested fix is right and I took it verbatim: no accepted heap spelling contains a colon, so the **first** colon is the separator, and the address parser reports the real fault:

```console
Error: invalid --heap value '#Strings:0x1a4:0x1b4': '0x1a4:0x1b4' is not a heap address.
```

### Verified and ruled out by review (no action)

Gemini independently confirmed, by reading and running: the `RowsTruncated`/`EntriesTruncated` budgets reflect reality; dedup keys on offset so two equal strings at two addresses stay two entries; `#GUID` `MaxAddress` really is `SizeInBytes / GuidSize`, making `MetadataLens_GuidHeapListing_ListsEveryEntry` mathematically sound rather than vacuous; the disclosure property holds against `-v:d` and `-S @All`; caveats route to stderr leaving TSV stdout pure (checked with `cmd > out 2> err`); and the promoted parser accepts everything `mdi`'s deleted `TryParseHeapLocation` did.

### Validation at `7ffcd32c`

| Suite | Result |
| --- | --- |
| `src/dotnet-inspect.Tests` | 2479 total, 2 failed |
| `tests/ILInspector.Metadata.Tests` | 998, 0 failed |
| `tests/DotnetInspector.MetadataRendering.Tests` | 118, 0 failed |

The 2 are the same pre-existing Windows-only failures documented above, confirmed at unmodified `origin/main`. One transient third failure (`CliDiscoverySections_AreSelectable`, "`member -S 'Source Locations'` produced no data") was a dirty-worktree build artifact from the mutation testing — assemblies built mid-mutation carry no SourceLink URLs. It passes on a clean build and is not related to this change.

Fixed-head re-review of `7ffcd32c` is running with both reviewers.

richlander:
## Fixed-head re-review — both clean

Both reviewers re-reviewed the exact fixed head **`7ffcd32c`** in their isolated worktrees, after rebuilding.

**MAI-Code:** *"No significant issues found."* Rebuilt Release, ran the heap-coordinate and parser coverage, and exercised the built CLI against `#Strings:0`, `#Blob:0`, `#GUID:0`, `#US:0`, `#Strings:999999999`, `#Strings:0x80000000`, and a warmed `-D` followed by a bad `--heap`, confirming the expected exit codes. Worktree clean.

**Gemini 3.1 Pro:** *"No significant issues found... the fixes fully resolve the findings from the first pass."* Notably, it withdrew its own first-pass position on the exit-code question — *"Your distinction between an image fact (`!malformed` row) and an invalid caller input (exit 1) is absolutely correct and matches `--il-offset` precedent"* — after testing the regression I asked it to hunt for:

- **The 0-address attack found nothing to fix.** `--heap "#Strings:0"`, `#Blob:0`, `#US:0`, and `#GUID:0` all exit 0 and render a `nil` row, because SRM represents address 0 as `IsNil`, which resolves to `MetadataValue.Nil` and never reaches the `Malformed` path. No legitimate boundary read was turned into an error.
- **It reproduced the cache-gate mutation independently** — removed the read-site guard, rebuilt, and saw `MetadataLens_CachedDiscovery_DoesNotBypassHeapCoordinateResolution` fail with exit 0 instead of 1, confirming the double-prime really does produce a cache *hit* and the test is not passing for the wrong reason.
- **`FilterEffective`'s new line cannot strip a legitimately requested section**, and the guard costs nothing on ordinary uncoordinated runs.
- **The first-colon split breaks no prior `mdi` input**: the deleted parser accepted only alphabetical heap enum names, none containing a colon.

### Validation

| Suite | Result |
| --- | --- |
| `src/dotnet-inspect.Tests` | 2479 total, 2 failed |
| `tests/ILInspector.Metadata.Tests` | 998, 0 failed |
| `tests/DotnetInspector.MetadataRendering.Tests` | 118, 0 failed |

The 2 are the pre-existing Windows-only failures confirmed at unmodified `origin/main`; both pass in the ubuntu CI lane above. CI is green on `7ffcd32c`, markdownlint included.

Non-blocking follow-up, not in scope here: #3468 (hex table selection, `0x02` ≡ `TypeDef`) is the remaining slice of #3401, which stays open until it lands.

Ready to merge

PR #3499, opened 2026-07-29
Replace scanner fan-out with declared prerequisites

richlander:
## Stack position

Slice 3 of the library section/scanner strand.

- Parent: #3470 (*Return library scan results instead of mutating the aggregate*) — **still open**. This PR targets `feature/library-scanners-return` and will be restacked onto `main` when #3470 lands.
- `origin/main` (`e1b3cac9`) is merged in, so this branch is current with the `@Metadata` lens (#3465) and #3495.

Residual left for later slices, enumerated so readiness here is unambiguous:

1. **No shared `AssemblyInspectionSession` in `ScannerContext`.** Prerequisites run the path-based scanner overloads, so each opens the file itself. Deliberate: `AssemblyInspectionSession` is `IDisposable` (unlike `MethodBodyInspectionSession`, which is why `ScannerContext` never disposes `_bodySession`), and a shared session must map one open-failure onto a `Failed` inspection *per requested scanner* — the failure contract each path overload owns today. That is a behavioral change and belongs in its own slice.
2. `Cost`/`ProbeEffectiveness` are still declared per section rather than derived from the scanner, and have drifted (the 8 `Performance: *` sections and `Array Pool Escapes` are body-index-backed but declare `NetworkFree`).
3. Discovery still ignores the requested door when choosing scanners (`LibraryCommand.cs:82`).
4. `ScannerKey` is still a `string?`; making it a typed query is the L1 extraction.

## What this changes

Three library scanners called other scanners from inside their bodies:

| Site | Fan-out |
| --- | --- |
| `ScanInfoCounts` | ran five censuses (extension members, classified methods, resources, custom attributes, type forwarders) |
| `AuditSignalBuilder.PopulateLibraryAudit` | ran `ScanClassifiedMethods` |
| `ScanIntegrationOpportunities` | re-ran `ScanIntegrations` when its output was absent |

An undeclared dependency is invisible to the registry, so it can neither order nor dedupe the work. At `-v:d` that meant `ScanCustomAttributes`, `ScanTypeForwarders` and `ScanClassifiedMethods` each ran **twice** — once via their own registry entry, once again inside `ScanInfoCounts`, which re-ran them unconditionally.

`ScannerRegistry.Add` now takes the keys a scanner reads:

```csharp
.Add(ScannerAuditSignals, ctx => AuditSignalBuilder.PopulateLibraryAudit(...),
    ScannerClassifiedMethods)
```

`RunScanners` walks that graph, running prerequisites first and each scanner at most once per context, and throws on a cycle. `ExpandRequired` exposes the transitive closure so `LibraryMetadataService.InspectAsync` expands **before** `SelectBodyAnalysisFeatures`, rather than narrowing body-analysis phases against a set that omits prerequisites.

`ScannerInfoCounts` becomes a **bundle** — a key with no body of its own that names the five scanners `Library Info` needs. Bundles exist because `ISectionDescriptor.ScannerKey` names a single key, so a section fed by several scanners has nowhere else to say so. `ScanInfoCounts` itself is deleted; it computed no counts, it only fanned out.

`PackageCommand` is untouched: it reaches the registry through `InspectAsync`, so closure expansion applies to it without a source change.

## Why behavior is unchanged

Every fan-out call site was a guarded re-run of the same deterministic scanner against the same file. Ordering them explicitly produces the same model with less work. The guards (`is null`, `??=`) become moot once the prerequisite is guaranteed to have run first.

**Evidence:** 100 comparisons across 25 assemblies × `-v:d`, `-v:d --json`, `-v:n`, and `-S "Library Info,Signals"` — **0 mismatches**, NativeAOT binaries, baseline = this branch's parent merged with `origin/main` (i.e. everything except this slice).

## Gates

`LibrarySections_RenderIdenticallyAloneAndTogether` asserts every scanner-bound section renders identically alone and in company. Asking for a section alone runs only its declared closure, so a scanner reading data it did not declare renders *less* in isolation — exactly the failure the fan-out hid.

Two properties of the gate matter:

- **Its section set derives from the pipeline, not from the prerequisite declarations**, so deleting a declaration does not also delete the coverage that catches it. The first version derived from the declarations and was silently vacuous under exactly that tamper.
- **Its assemblies are chosen, not decorative.** `System.Text.Json` reports 0 unsafe public signatures either way and cannot catch a missing audit prerequisite; `System.Runtime.InteropServices` reports 15.

Tamper-verified: removing `ScannerClassifiedMethods` from the audit registration fails the gate with `Unsafe public signatures | 15` → `| 0`. (The same tamper on `System.Private.CoreLib` moves 2256 → 0, confirming the prerequisite is load-bearing rather than defensive.)

Body-index-backed scanners are excluded from that gate for run time only; a new one would make it slower, never wrong.

Registry mechanics are covered separately in `SectionPipelineTests`: prerequisite ordering, once-per-run across multiple requesters, bundle behavior, transitive closure, and cycle detection. `LibraryScannerPrerequisites_AreAllRegisteredAndAcyclic` holds every declared prerequisite to a registration — `RunScanners` skips an unregistered key, so nothing else would notice.

## Performance

Not the headline, and stated conservatively because body-index runs are noisy on this machine. NativeAOT, interleaved A/B, min and median of 11–15 warm runs:

| Case | Base | After |
| --- | --- | --- |
| `-v:q`, `-v:m`, `-S Library Info`, `-S Signals` (Decompiler.dll) | — | within noise (±20 ms on ~300 ms) |
| `-v:n` / `-v:d` (Decompiler.dll) | 4433 / 3852 ms min | 4256 / 4012 ms min — within noise |
| CoreLib, six census-backed sections (the set that actually duplicated) | 617 min / 751 med | **576 min / 657 med** |

An earlier sequential-block measurement appeared to show −42% on `-S Library Info`; that was an ordering artifact and does not survive interleaving. The real gain is the ~7–12% on the previously-duplicated selection.

The one trade: `-S "Library Info"` alone now opens the assembly once per census instead of sharing one session, since prerequisites use the path-based overloads. Measured at or below noise, and residual 1 above is the fix.

## Validation

- `dotnet run --project src/dotnet-inspect.Tests -c Release` — **2468 total, 2 failed, 10 skipped**. Both failures (`Layout_Count_CountsFilesRatherThanRenderedTreeLines`, `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown`) reproduce identically on a clean detached worktree at `origin/main` (`e1b3cac9`); 10 skips are absent `ildasm`/`ilasm`.
- Output sweep and benchmarks above.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -6,21 +6,23 @@
 projects, platform libraries, metadata, APIs, dependencies, source provenance,
 analysis, Findings, implementation diffs, and decompilation.
 
-Read this file before doing work. Then read only the task-specific entry
-documents relevant to the change:
-
-- Read `README.md` when changing user-visible capabilities, commands, or
-  examples.
-- Read `docs/overview.md` when a change crosses subsystem ownership boundaries.
-- Read the relevant section of `docs/architecture.md` only when implementation
-  structure matters to the task.
-- Follow the task-specific entry points below and then only the links relevant
-  to the change.
+Read this file before doing work, then read only the entry documents below that
+are relevant to your change.
 
 This file is the source of truth for repository-wide engineering and workflow
 rules. Detailed design, subsystem mechanics, version requirements, and
 historical context belong with their owning code, workflow, or focused
 documentation.
+
+### Nightshift is opt-in
+
+`NIGHTSHIFT.md`, the `nightshift` skills, and the
+`nightshift`/`turnstile`/`octoshift` tools describe a separate multi-agent
+operating model with its own vocabulary and its own stricter gates. **They apply
+only when you have been explicitly told that you are working in Nightshift mode
+for this session.** Otherwise they are inapplicable: follow this file, and do
+not adopt Nightshift roles, orders, gates, or tooling merely because you noticed
+those documents exist.
 
 ## Before changing files
 
@@ -35,35 +37,38 @@ documentation.
 - Use one development worktree per PR, plus temporary worktrees for independent
   reviews. Do not reuse a worktree across unrelated changes.
 - Never amend commits; create follow-up commits.
-- Before requesting review, fetch `origin/main` and incorporate it into the
-  feature branch. Rebase only before the branch's first push. Once a branch is
-  public or under review, merge `origin/main`; never amend, rebase, or
-  force-push reviewed history. A slice in a stack is the standing exception:
-  restacking rebases and force-pushes a public branch by design — see
+- Integrate `origin/main` into the feature branch before **every** review round,
+  not only the first — see [Adversarial review](#adversarial-review). Rebase
+  only before the branch's first push. Once a branch is public or under review,
+  merge `origin/main`; never amend, rebase, or force-push reviewed history. A
+  slice in a stack is the standing exception: restacking rebases and
+  force-pushes a public branch by design — see
   [Stacked PRs for multi-slice issues](#stacked-prs-for-multi-slice-issues) for
   the discipline that replaces this rule there.
 - After updating from main or resolving conflicts, re-read `AGENTS.md` and
   task-relevant docs before continuing.
 - Do not mix unrelated changes into one commit or sweep another contributor's
   working-tree changes into your work.
-- Treat worktrees as temporary. For a PR requiring adversarial review, confirm
-  the exact reviewed head is pushed, then remove the development and review
-  worktrees with `git worktree remove <path>` as soon as every required
-  fixed-head review is clean. For a change that does not require adversarial review,
-  remove its development worktree after merge. Do not retain inactive
-  worktrees in case more work appears; recreate one for the branch if follow-up
-  work is needed.
+- Treat worktrees as temporary. Confirm the exact reviewed head is pushed, then
+  `git worktree remove <path>` as soon as every required fixed-head review is
+  clean — or after merge, for a change that needs no adversarial review. Do not
+  retain inactive worktrees in case more work appears; recreate one for the
+  branch if follow-up work is needed.
 
 ## Task-specific guidance
 
 | Area | Read first |
 | --- | --- |
+| User-visible capabilities, commands, or examples | `README.md` |
+| A change crossing subsystem ownership boundaries | `docs/overview.md` |
+| Implementation structure | the relevant section of `docs/architecture.md` |
 | Layering and consumer boundaries | `docs/design/inspection-layers.md` |
 | Command defaults and disclosure | `docs/design/progressive-disclosure.md` |
 | Output data shapes | `docs/design/output-shapes.md` |
 | Output style | `docs/design/style-guide.md` |
 | Sections and selection | `docs/design/section-model.md` |
 | Metadata and API inspection | `docs/design/assembly-inspection-query.md` |
+| Type, member, or API identity | `docs/design/type-member-api-representation.md` |
 | PDB and source acquisition | `docs/pdb-acquisition.md` |
 | Source Finding producers | `docs/design/source-finding-producers.md` |
 | Package resolution and caches | `docs/design/version-resolution.md` |
@@ -74,6 +79,7 @@ documentation.
 | IL round-trip tests | `tests/DotnetInspector.ILRoundtrip.Tests/README.md` |
 | Decompiler behavior or harnesses | `docs/decompiler-correctness-pipeline.md` |
 | Skills | `taste/skill-guidance.md` |
+| Stacked PRs and restacking | `docs/stacked-prs.md` |
 | Release and publishing | `docs/release-workflow.md` |
 
 PR templates:
@@ -89,9 +95,12 @@ current product behavior and tests over design history. When current sources
 disagree, stop and resolve which owner is authoritative rather than silently
 choosing one.
 
-When adding a focused skill, register it in `SkillCommand.Skills`. Its YAML
-frontmatter `description:` is the single source of truth for the generated
-skill listing.
+When adding a focused skill, register it in `SkillCommand.Skills` **and** add an
+`EmbeddedResource` line for it in `src/dotnet-inspect/dotnet-inspect.csproj`;
+the embeds are enumerated per skill, and no test compares them against the
+`skills/` directory, so a skill missing from either list ships as nothing with a
+green suite. Its YAML frontmatter `description:` is the single source of truth
+for the generated skill listing.
 
 ## Repository-wide engineering constraints
 
@@ -111,6 +120,23 @@ skill listing.
   presentation as separate concerns. Do not infer one from display text when a
   typed identity exists.
 
+### Output contract
+
+Commands that render sections follow this verbosity model:
+
+- `-v:q`: compact fields only; include high-value fields only.
+- `-v:m`: one section, plus an optional text line. Include all high-value
+  fields in that section.
+- `-v:n`: multiple sections are allowed; include all sections that are not
+  network-bound.
+- `-v:d`: all sections.
+
+New sections must not enter the default `-v:m` view unless they are the
+command's single high-value section. Focused flags may explicitly select a
+section and promote verbosity as needed. Keep alternate lenses, section
+selection, row queries, and rendering formats orthogonal; follow the current
+progressive-disclosure and output-shape docs for detailed behavior.
+
 ### Terminology
 
 Prefer inclusive terminology in code, identifiers, comments, output, and docs.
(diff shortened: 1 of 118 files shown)


richlander:
## Adversarial review reconciliation — fixed head `fdb2bc8f`

Two reviews, two different model families (GPT-5.6 and MAI-Code), each in its own detached review worktree at the exact head. Both returned **no blocking findings at `fdb2bc8f`**.

### Findings and resolution

| # | Reporter | Severity | Finding | Resolution |
|---|---|---|---|---|
| 1 | GPT | Blocking | Closure expansion could recurse forever on a prerequisite cycle | `d00b8fa4` — `visiting` set checked before `closure`, plus a gate |
| 2 | GPT | Blocking | Sharing one session let a scanner failure abort the whole inspection instead of mapping to a scanner-level failure | `c67373ca` — restored the `try`/`catch` around `ScanIntegrationOpportunities`' session overload |
| 3 | GPT | Should-fix | The scan-count gate could not distinguish "shared" from "reopened" | folded into the retarget gates in `fdb2bc8f` |
| 4 | GPT | **Blocking** | `-S Signals` mixed data from two different images | `ceb7c2ea` — the borrow (see below) |
| 5 | GPT + MAI (independent) | **Blocking** | A warmed `MethodBodySource` outliving its lender reads unmapped memory → uncatchable `AccessViolationException` | `fdb2bc8f` — restored lender liveness |
| 6 | GPT | Should-fix | Both retarget gates were vacuous for the three action-based scanners | `fdb2bc8f` |
| 7 | GPT | Should-fix | The failure-mapping gate was vacuous for `IntegrationOpportunities` | `fdb2bc8f` |
| 8 | MAI | Should-fix | `IntegrationOpportunities → Integrations` prerequisite unpinned | `d9b9f529` (implemented before the report; pinned in response) |
| 9 | MAI | Should-fix | A prerequisite naming an unregistered scanner was silently dropped | `ddc7a99d` |
| — | MAI | Blocking (**dismissed**) | `ScannerAuditSignals` bypasses the borrow | inverted; **withdrawn by MAI** at `fdb2bc8f` |

### On the dismissal

MAI's probe set `MetadataContext` to assembly A and `AssemblyPath` to assembly B and observed A's counts. That is the borrow *working* — reopening `AssemblyPath` would have yielded B's count of 0. The fix it proposed is what `LibrarySections.cs:150-153` already does, and the crossed fixture is a state `InspectAsync` cannot construct. MAI withdrew the finding explicitly this round after re-running the gate itself.

### Finding 4 is the substantive one

`InspectAsync` already opened the assembly once (`SourceLinkService.Open` → `PdbContext`) and already handed that context to `ScannerContext.MetadataContext`. But `ScannerContext.Session()` called `AssemblyInspectionSession.Open(AssemblyPath)` — **a second open of the same path**. Sharing one session among the scanners did not close the window; it only moved it earlier. `ceb7c2ea` borrows the already-open image instead, so there is no second open to race.

Measured with a junction-retarget harness (150 runs/arm, interleaved): **base 15 mixed-image results across 64 successful runs (23%); head 0 across 86+.**

### Finding 5 is the one I got wrong

I wrote the lender-liveness delegate, tampered it, saw the suite could not detect its absence, and **deleted it as decoration**. My tamper touched a *cold* `MethodBodies` after disposal, where the already-disposed `PEReader` throws and masks the hole. The dangerous shape is a `MethodBodySource` obtained *while the lender was alive*: it captures the reader, so the borrow's own `_disposed` flag is not the question. Both reviewers found this independently.

A tamper test that does not exercise the dangerous state proves nothing. `BorrowedSession_FailsLoudlyAfterTheLenderIsDisposed` now warms the body source first, and covers both `Open` and `OpenPrefetched`.

I then swept all 21 borrowed facets after lender disposal — including the ones that read `_image.PEReader` directly and never call `EnsureAlive`. Every one throws `ObjectDisposedException`; none returns data; the process survives. Both reviewers reproduced this independently.

### Gate vacuity, and the assertion that caught a trap

Extending the retarget gates to the action-based scanners, the new per-group `Assert.NotEqual` on the action signature alone fired immediately: audit signals are a fixed catalog, so comparing *counts* produced an identical signature for every input. It now compares signal **values**. Without that non-vacuity assertion I would have added coverage that looked real and asserted nothing — so it stays, permanently: if a future fixture pair stops distinguishing those scanners, the gate fails loudly instead of silently reverting to census-only coverage.

### Validation at `fdb2bc8f`

- `dotnet-inspect.Tests` 2500 total, 2 failed, 10 skipped — both failures reproduce at clean `origin/main` (`Layout_Count_CountsFilesRatherThanRenderedTreeLines`, `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown` / `images/logo.png.`).
- `ILInspector.Metadata.Tests` 998/0 · `ILInspector.Analysis.Tests` 584/0 · `DotnetInspector.Services.Tests` 343/0 · `SectionPipelineTests` 152/152.
- NAOT output sweep vs `05318066`: 8 targets × 9 argument sets, **72 comparisons, 0 mismatches**.
- Perf, interleaved NAOT A/B: neutral (`-S Signals` 838 ms median both arms).
- Tampers verified to fail the gate they belong to: borrow → reopen; borrow disposing the lender's reader; removing `_ensureLenderAlive?.Invoke()`; `SharedScanCount++; reopen();` in the **void** overload; deleting `MarkIntegrationFailuresIfMissing`.

### Known gap (non-blocking, follow-up)

A behavioral gate for `ScannerIntegrationOpportunities → ScannerIntegrations` needs a purpose-built fixture that both renders opportunities and carries a suppressible existing integration in the same category. The prerequisite is pinned structurally today.

### Stack note

This PR is based on #3470, not `main`, so GitHub runs no CI on it and its `CONFLICTING` flag is stale (the compare endpoint reports `behind_by: 0`). Both resolve when it is restacked onto `main` after #3470 merges.

richlander:
## Merge conflicts resolved — `e92864d3`

**Cause.** #3470 was rebased onto current `main` and force-pushed. That orphaned `9d9d0fd7`, the commit this branch was stacked on, so the same change existed twice under two different SHAs and every conflict was "the same change arriving twice."

**The check that made the resolution safe.** Before resolving anything I confirmed the rebase carried no content change:

```
git range-diff 4d44b1c8..9d9d0fd7 59f731a1..93569639
1:  9d9d0fd7 = 1:  93569639 Return library scan results instead of mutating the aggregate
```

`=` means an identical patch. Without that, taking this branch's side could have silently dropped new work done during the rebase.

**Resolution.** Four conflicting hunks in three files, every one a place where this PR deliberately supersedes its parent, so all four take this branch's side:

| File | Hunk | Why ours |
|---|---|---|
| `AuditSignalBuilder.cs` | classified-methods fan-out | this PR removes the fan-out |
| `LibrarySections.cs` | extension members / classified methods | `ctx.Scan(session, reopen)` replaces the path-only call |
| `LibrarySections.cs` | custom attrs / forwarders / counts | `AddBundle` replaces `ScanInfoCounts` |
| `LibraryMetadataService.cs` | `ScanInfoCounts` itself | deleted by this PR |

**Verification that the resolution lost nothing.** `main`'s six new commits touch **none** of the three conflicted files (`git diff --stat 4d44b1c8..59f731a1 --` on them is empty), and the merged tree is byte-identical to `fdb2bc8f` for all three (`git diff fdb2bc8f..HEAD --` on them is empty). So nothing incoming existed to lose, and this branch's side survived exactly. No conflict markers remain anywhere.

**Post-merge validation.** The merge also pulled in #3510's new metadata-lens sections and #3461's decompiler work, which is the interaction worth checking — new sections are what perturbed the render gate when #3497 landed.

- Solution builds clean; 5 warnings, all pre-existing in untouched test projects.
- `dotnet-inspect.Tests` **2523 total, 2 failed, 10 skipped**. Total is up 23 from main's new `MetadataLensTests`; the two failures are the same pre-existing pair that fails at clean `origin/main` (`Layout_Count_CountsFilesRatherThanRenderedTreeLines`, `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown` / `images/logo.png.`).
- `ILInspector.Metadata.Tests` 998/0 · `DotnetInspector.Services.Tests` 343/0 · `ILInspector.Analysis.Tests` 584/0.
- `LibrarySections_RenderIdenticallyAloneAndTogether` passes against the newly merged sections without needing a new exclusion.

The PR diff is now correctly scoped to this change alone — 11 files, +1291/−86 — and GitHub reports `MERGEABLE`. CI is running against this PR for the first time, since it previously had a stacked base that suppressed it.

richlander:
Ready to merge.

Conflicts are resolved at 92864d3 and CI is green on this PR for the first time (uild-net10 pass, 	est (ubuntu-26.04, linux-x64) pass; pack/markdownlint/diff-smokes path-skipped).

The two adversarial reviews were clean at db2bc8f, and the merge provably did not disturb what they reviewed:

- Every one of the 29 files that changed since db2bc8f is a file `main` or the rebased parent moved -- none outside.
- The three conflicted files are byte-identical to the reviewed head.
- Full local suites re-run after the merge: 2523/2 pre-existing/10 skipped, 998/0, 584/0, 343/0.

**Merge order:** this PR is based on `feature/library-scanners-return`, so #3470 merges first.

Issue #3489, opened 2026-07-29
Declaration formatter escapes unnamed keyword-typed tuple elements: (int, int) renders as (@int, @int), emitting non-compiling C#


richlander:
> [!IMPORTANT]
> **Diagnosis below is superseded.** Traced in
> [this comment](https://github.com/richlander/dotnet-inspect/issues/3489#issuecomment-5123965893).
> The cause is **not** the nested `ValueTuple<…,TRest>` spelling and is not specific
> to `TupleRest` or to arity 8. The declaration formatter `@`-escapes **any unnamed
> tuple element whose type spells as a C# keyword**, so `(int, int)` renders as
> `(@int, @int)` — a verbatim identifier, hence `CS0246`. `(int X, int Y)` and
> `(System.Guid, System.Guid)` are unaffected.
>
> This is a **shipping-product bug**, not only a gate failure: `dotnet-inspect member
> -s Methods` prints non-compiling C# for ordinary tuple signatures. Extraction is
> correct (`--json` emits `(int, int)`); the corruption is applied downstream in the
> declaration formatter.
>
> Original report retained below for the CI evidence.

---

## Summary

`FidelityGateTests.PinnedFixesStayExact` fails: `CfgSampleClass.TupleRest` has
regressed from pinned-`Exact` to `RecompileFail`. The decompiled C# no longer
compiles, so the eight-element `ValueTuple` nested-`TRest` flattening that
`PinnedExact` was created to protect is no longer verified at all.

## Evidence

```
TupleRest regressed to RecompileFail: a prior fidelity check fix no longer holds.
  original  : ldarg ldarg ldc.i4 add ldarg ldc.i4 add ldarg ldc.i4 add ldarg ldc.i4 add
              ldarg ldc.i4 add ldarg ldc.i4 add ldarg ldc.i4 add newobj newobj ret
  recompiled:
  detail    : CS0246: The type or namespace name 'int' could not be found
              (are you missing a using directive or an assembly reference?)
```

`CS0246` on `int` is the notable part: the emitted skeleton is failing to bind a
primitive, which points at the spelling of the nested `ValueTuple<...,TRest>`
construction rather than at a missing `using`.

Fixture: `src/ILInspector.Decompiler.Tests/CfgSampleClass.cs:2770`
Pin: `src/ILInspector.Decompiler.Tests/FidelityGateTests.cs:401` (`PinnedExact`),
rationale at `:271`.

## Why this was not caught

[...]

PR #3500, opened 2026-07-29
Collapse spilled receiver+coalesce in reference-type field initializers

richlander:
- Advances #3242 (spilled-temporary inlining was under-powered) — concrete evidence and fix for a reference-type field-initializer spill the pass could not collapse.
- Changes: a reference-type field initializer `this._f = x ?? new()` now renders as a clean field store instead of leaking spilled stack-slot temporaries, and the primary-constructor base-`object..ctor` residual that the leftover temps caused is elided.
- Card revision: `fe9d7381d55622a31c25c0df29dc53dfa7483c5c`

## Change

**Conclusion:** **PASS** — collapses a spilled receiver+coalesce pair that a circular-ordering deadlock left in place; both sub-changes are soundness-gated (reference-type receiver only; pure-prefix + type-witness for the value), and the full fidelity diff set is byte-identical to `origin/main`.

### Benchmark target

Benchmark target: `richlander/markout` `TemplateDemo/release/Markout.dll` (real witness `Markout.FieldWriter::.ctor`).

dotnet-inspect command:

```bash
dotnet-inspect member 'FieldWriter' '.ctor' --library Markout.dll -S "Decompiled Source"
```

### Original source

SourceLink resolves the published `Markout@0.33.0` PDB to
`richlander/markout@2636d69c211640ad051b883a1465b09536f6ada2`.
Acquired with:

```bash
dnx dotnet-inspect -y -- member FieldWriter .ctor --package Markout@0.33.0 -S "Original Source"
```

```csharp
using Markout.Formatting;

namespace Markout;

/// <summary>
/// Writes key-value fields to a TextWriter using a field formatter.
/// Handles projection (include/exclude fields) and layout variants
/// (standard, inline, bulleted, numbered). Document state is managed
/// by the caller or <see cref="MarkoutWriter"/>.
/// </summary>
public class FieldWriter(TextWriter writer, IFieldFormatter formatter, MarkoutWriterOptions? options = null)
{
    private readonly MarkoutWriterOptions _options = options ?? new();
```

The corresponding IL shows that the field stores precede the base
`object::.ctor` call at `IL_001F`, and that the coalesce lowers to
`dup/brtrue/pop/newobj` at `IL_0010..0019`:

```il
IL_0000: ldarg.0
IL_0001: ldarg.1
IL_0002: stfld        Markout.FieldWriter::<writer>P
IL_0007: ldarg.0
IL_0008: ldarg.2
IL_0009: stfld        Markout.FieldWriter::<formatter>P
IL_000E: ldarg.0
IL_000F: ldarg.3
IL_0010: dup
IL_0011: brtrue.s     IL_0019
IL_0013: pop
IL_0014: newobj       Markout.MarkoutWriterOptions::.ctor()
IL_0019: stfld        Markout.FieldWriter::_options
IL_001E: ldarg.0
IL_001F: call         System.Object::.ctor()
IL_0024: ret
```

### Before

```csharp
public FieldWriter(System.IO.TextWriter writer, Markout.Formatting.IFieldFormatter formatter, Markout.MarkoutWriterOptions? options = null)
{
    this.writer = writer;
    this.formatter = formatter;
    FieldWriter S_0 = this;
    MarkoutWriterOptions S_1 = options ?? new MarkoutWriterOptions();
    S_0._options = S_1;
    /* Unsupported IL_001F call .ctor: direct constructor call to object is not representable as a C# statement */
}
```

- Valid: False
- Correct: False (the `/* Unsupported ... */` residual is not compilable C#)
- IL fidelity: not currently checkable (invalid C#)
- Taste applied: None
- Commit: `e1b3cac9c` (origin/main)

### After

```csharp
public FieldWriter(System.IO.TextWriter writer, Markout.Formatting.IFieldFormatter formatter, Markout.MarkoutWriterOptions? options = null)
{
    this.writer = writer;
    this.formatter = formatter;
    _options = options ?? new MarkoutWriterOptions();
}
```

- Valid: True
- Correct: True
- IL fidelity: True (round-trips to the original opcodes; base call re-emitted, coalesce re-emitted)
- Taste applied: None
- Commit: `fe9d7381d55622a31c25c0df29dc53dfa7483c5c`

### Fully raised

## Evidence

| --- | --- | --- |
| Product build (`dotnet build dotnet-inspect.slnx -c Release`) | Pass | Pass |
| Focused tests (`SpilledReceiverCoalesceInliningTests`) | n/a (new) | 4 passed |
| `Area=Pass` | 1415 passed | 1419 passed |
| Decompiler fast suite (`-trait- "Speed=Slow"`) | 4319, 0 failed | 4323, 0 failed |
| Corpus gate (`Area=Corpus`) | 319, 0 failed | 319, 0 failed |
| Fidelity gate (`Area=Fidelity`) | 66, 3 failed | 66, 3 failed |
| Real witness (`Markout.FieldWriter::.ctor`) | spilled temps + Unsupported residual | clean field init |

The 3 Fidelity-gate failures (`FidelityGateTests.PinnedFixesStayExact` = #3489 TupleRest; `FidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket` = #3490 MakeConsumerWithTwoLeadingArgs et al.; `LoweredFidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket` = #3491) are **pre-existing on `origin/main`**, surfaced by #3495 completing gates that previously timed out. Proven by running both fidelity classes on a clean detached `origin/main` worktree: identical 3 failures, and the full "current diff set" is **byte-identical** (123 rows, `md5 9a0d7221…`) between baseline and head — this change introduces no new fidelity diff.

## Decompiler quality

**Conclusion:** **PASS** — the pinned Corpus gate is green (319/319); the change strictly reduces residue counts.

### Aggregate context

Static residue sizing over the local corpus:

| Metric (goal) | Baseline | PR |
| --- | ---: | ---: |
| Constructor base-call residuals (-) | 49 | 47 |
| Single-use argument spills unrescued (-) | 157 | 81 |

Both move in the intended direction; no metric regresses.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -class ILInspector.Decompiler.Tests.SpilledReceiverCoalesceInliningTests
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- -trait "Area=Corpus"
```


richlander:
## Adversarial review reconciliation (high-risk, two-model)

This inlining change touches a soundness-critical path, so it received a two-model review from two different families at exact head `fe9d7381d`: **GPT** (`gpt-5.6-sol`) and **Gemini** (`gemini-3.1-pro-preview`), each in an isolated review worktree.

### Converging finding — conditional-arm inlining (FIXED)

Both reviewers **independently found the same real bug** (GPT: High; Gemini: Critical):

> `PrecedingEvaluationIsPure` only checks the load's **left siblings**, not whether an **ancestor** evaluates the load conditionally. A non-pure spilled value could therefore be inlined into a `?:` arm, `??` fallback, `&&`/`||` right operand, `?.` member, or switch arm — moving it from **unconditional** to **conditional** execution. This is realizable in the late `slotsOnly` run after `SlotStoreDiamondPass` structures diamonds into `Conditional`.

Gemini additionally flagged **try/catch/finally regions** as a class of conditional context to cover.

**Resolution — commit `0cb59f403`:** added `LoadIsUnconditionallyEvaluated(load, next)` to the `precedingPure` gate. It walks the statement root down to the load and rejects if:

1. any expression parent evaluates the on-path child conditionally — `ChildIsUnconditional` denylists `Conditional` (only `Condition` unconditional), `Coalesce` (only `Left`), `LogicalBinary` (only `Left`), `NullConditional` (none), the three switch-expression forms (only `Value`), and `TupleSwitchExpression` (only `Components`); or
2. the path descends through **any non-`IrExpression` node** below the root statement — a generic guard that rejects try/catch/finally regions, if/loop bodies, nested statements, and switch arms **without enumerating them** (directly covers Gemini's try/catch/finally concern).

The existing `firstLeaf || pure` path is intentionally left unguarded: a pure value is effect-free (conditional evaluation is unobservable), and a non-pure `firstLeaf` value cannot sit inside a conditional arm (the condition/scrutinee would itself be the first leaf).

Regression test: `NonPureSpill_InConditionalArm_IsNotInlined`.

### Confirmed sound (no change needed)

Both reviewers independently verified as **sound**:

- **`ReceiverThisIsPure`** — treating `this` as pure/stable only for confirmed reference-type receivers (non-null `BaseType` that is not corelib `System.ValueType`/`Enum`). Generic `T:struct` decompiles with a declaring type inheriting `System.Object`; real structs/enums/ref-structs fall back to `ValueType`/`Enum`; null/unresolved base (including `System.Object` itself) stays conservative.
- **The type-witness guard** — `StoreStackSlot { Value.ResultType } == LoadStackSlot { ResultType }` prevents dropping a type reconciliation; null `ResultType` fails the `{}` pattern and closes the gate.

### Validation on the fixed head (`0cb59f403`, Release)

- fast suite **4324/4324**, Corpus **319/319**, spill tests **5/5**
- Fidelity gate: 3 failures, all **pre-existing** (#3489/#3490/#3491) — proven by a byte-identical diff set on a clean `origin/main` baseline; **zero new fidelity diffs**
- Markout witness `Markout.FieldWriter::.ctor` still renders `_options = options ?? new MarkoutWriterOptions();`

Re-review of the fixed exact head is in flight with both reviewers.

richlander:
## Second review round — reconciliation (head `e3ce63b6b`)

Both reviewers re-reviewed the first fix (`0cb59f403`) and independently found the conditional-evaluation guard was **incomplete**. Both findings are now addressed in `e3ce63b6b`.

### GPT (High) — `firstLeaf` path bypassed the guard via `??=`

> A `??=` exposes its guarded fallback as `Children[0]`, so `IsFirstEvaluatedLeaf` misreports the fallback as first-evaluated. Because the new guard only ran on the `precedingPure` (`!firstLeaf`) path, a non-pure spill inlined into the conditional fallback via the `firstLeaf` path.

**Verified real, and pre-existing.** I reproduced GPT's exact input on a clean `origin/main` build (`4d44b1c83`) — `S_0 = SideEffect(); V_0 ??= S_0;` folds to `V_0 ??= SideEffect()` there too. This is a latent bug that predates this PR, but it is the same soundness class, so it is fixed here.

**Resolution:** the unconditional-evaluation check now runs ahead of **both** paths — `if (!pure && !LoadIsUnconditionallyEvaluated(load, next)) continue;` — so a node that misreports its first-evaluated leaf can no longer bypass it. (Removed the now-redundant copy inside the `precedingPure` gate.)

### Gemini (Critical) — two `IrExpression` forms missing from the denylist

> `NullCoalescingFieldAssignmentExpression` (Value conditional) and `TupleBinaryExpression` are `IrExpression`s the generic non-expression guard does not catch, and both were absent from `ChildIsUnconditional` (default `true`).

- **`NullCoalescingFieldAssignmentExpression` — verified, fixed.** Its `Value` is genuinely conditional. Added to the denylist (receiver unconditional, value conditional).
- **`TupleBinaryExpression` — respectfully dismissed with evidence.** Gemini's claim rests on C# tuple equality short-circuiting *element evaluation*. A compiled canary shows it does not:

  ```csharp
  (Touch("L0",0), Touch("L1",99)) == (Touch("R0",1), Touch("R1",99))
  // prints: eval L0 / eval L1 / eval R0 / eval R1 / result=False
  ```

  All four elements evaluate (left-to-right) **before** any comparison; only the effect-free comparisons short-circuit. So a tuple-equality component is unconditional — adding it would wrongly decline a valid fold. Documented in `ChildIsUnconditional`, with a positive regression test (`NonPureSpill_InTupleEqualityComponent_StillInlines`).

### Completeness

I audited the full IR node catalog for conditionally-evaluating forms. The complete denylist is now: `Conditional`, `Coalesce`, `LogicalBinary`, `NullConditional`, the four switch-expression forms, and the four null-coalescing (`??=`) forms. Everything else evaluates all operands unconditionally; non-expression control boundaries (blocks, loops, try regions, switch arms, **lambda bodies** — `Lambda.Body` is a `BlockContainer`) are rejected by the generic non-expression guard; pattern nodes need no case because their only non-scrutinee children are pure `Constant`s.

### Validation on `e3ce63b6b` (Release)

- spill tests **8/8** (3 new: `??=` local, `??=` field-expression, tuple-equality positive)
- fast suite **4327/4327**, Corpus **319/319**
- `??=` local + field-expression probes now stay spilled; tuple-equality still folds
- Markout witness still renders `_options = options ?? new MarkoutWriterOptions();`

Third-round re-review at `e3ce63b6b` requested from both reviewers.

richlander:
## Third review round — reconciliation (head `ad92507f2`)

### Gemini — **CLEAN** at `e3ce63b6b`

Gemini re-reviewed and passed the previous head, independently compiling a canary that confirmed the `TupleBinaryExpression` dismissal (C# evaluates every tuple element before any comparison), and confirming the null-coalescing completeness, the universal `firstLeaf` guard, the control-flow-boundary handling, and `ReceiverThisIsPure`/type-witness.

### GPT (High) — `WithExpression` clone ordering — verified, fixed in `ad92507f2`

> `receiver with { P = value }` clones the receiver (a possibly effectful/throwing copy constructor) **before** evaluating any initializer value. The receiver is pure, so `PrecedingEvaluationIsPure` accepted folding a non-pure spill into the initializer, moving its effect past the clone.

**Verified real** — reproduced on `e3ce63b6b`: `S_0 = SideEffect(); return receiver with { P = S_0 };` folded to `return receiver with { P = SideEffect() };`. This is a **distinct class** from conditionality (the initializer *is* unconditionally evaluated) — it is a *hidden operation* that runs partway through a node's children, which `PrecedingEvaluationIsPure`'s "the node's operation runs after all its operands" model did not account for. It is reachable only via the `precedingPure` path this PR introduced, so it is a genuine regression this PR must own (not pre-existing, unlike the `??=` finding).

**Resolution:** `ChildFollowsHiddenOperation` rejects descent into any non-receiver child of a `WithExpression` (clone precedes it) and, defensively, any non-creation child of an `ObjectInitializerExpression` (the constructor precedes the initializers; the creation is already non-pure, so the left-sibling check blocked this, but reject explicitly so the guard does not depend on that). Regression test `NonPureSpill_InWithExpressionInitializer_IsNotInlined`.

### Where the soundness argument now stands

The `precedingPure` fold is safe iff, for the value moved to the load position, nothing observable is evaluated before it that was not already there. Three independent hazards, each now gated:

1. **Impure left siblings** — original `PrecedingEvaluationIsPure` sibling-purity check.
2. **Conditional evaluation** (value moves from always-run to maybe-run) — `LoadIsUnconditionallyEvaluated` + `ChildIsUnconditional` denylist: `Conditional`, `Coalesce`, `LogicalBinary`, `NullConditional`, the four switch-expression forms, the four `??=` forms; non-expression control boundaries rejected by the generic guard.
3. **Hidden operation partway through children** (value crosses a clone/construction) — `ChildFollowsHiddenOperation`: `WithExpression`, `ObjectInitializerExpression`.

### Validation on `ad92507f2` (Release)

- spill tests **9/9** (4 new across the two rounds), fast suite **4328/4328**, Corpus **319/319**
- `WithExpression` probe now retains the store; intended coalesce fold still collapses
- Markout witness still renders `_options = options ?? new MarkoutWriterOptions();`

Re-review at `ad92507f2` requested from both — with an explicit ask to challenge the **completeness** of both denylists (any remaining conditionally-evaluating or hidden-operation node type).

richlander:
## Reconciliation — round 4 (head `73916a6f2`)

**A fourth hidden-operation hazard, and a reviewer disagreement worth recording.**

**GPT (NOT CLEAN, High):** found that an interpolated string `$"{a}{b}"` lowers to a
`DefaultInterpolatedStringHandler` whose appends are **interleaved** with hole
evaluation — the append of an earlier hole runs *before* a later hole is evaluated.
Appending an earlier `IFormattable`/custom-handler value can run arbitrary user code,
so every hole after the first is preceded by an observable operation. The `precedingPure`
fold this PR introduced would move a non-pure spilled value into a later hole (its left
siblings are pure loads), reordering the spilled effect after those appends. Same
hidden-operation class as the `WithExpression` clone.

- Reproduced on `ad92507f2` with a direct-IR probe: `S_0 = SideEffect(); return $"{formatter}{S_0}";`
  folded the store into the second hole (`stores=1 → 0`). GPT's compiled canary independently
  showed `spilled=effect,format` vs `inlined=format,effect`.
- Introduced by this PR (on `origin/main` a non-first, non-pure hole was never inlined —
  it is neither the first leaf nor pure), so this PR owns it.

**Gemini (CLEAN):** its completeness audit reached the opposite conclusion on the *same node*,
explicitly classifying `InterpolatedStringExpression` as safe on the assumption that it, like
`TupleExpression`/`NewArray`, "lower[s] to constructors … after unconditionally evaluating the
arguments left to right." That assumption is the exact one the interleaved handler lowering
violates: the appends are **not** deferred to a single trailing operation. GPT is empirically
correct; I verified the interleaving with the direct-IR repro above.

**Fix (`73916a6f2`):** `ChildFollowsHiddenOperation` now rejects **every** formatted-value child
of an `InterpolatedStringExpression`. The `firstLeaf` path stays unguarded and sound — a hidden
operation runs *between* children, so it can never precede the first evaluated leaf, and the
handler construction / literal appends before the first hole are runtime plumbing with no
IR-modeled evaluation. Regression test `NonPureSpill_InInterpolatedStringHole_IsNotInlined`.

**Soundness model (unchanged shape, denylist extended):** three gated hazards for the
`precedingPure` fold —
1. impure left siblings → `PrecedingEvaluationIsPure`;
2. conditional evaluation → `LoadIsUnconditionallyEvaluated` + `ChildIsUnconditional`;
3. hidden operation partway through children → `ChildFollowsHiddenOperation`
   (`WithExpression` clone, `ObjectInitializerExpression` ctor, **`InterpolatedStringExpression`
   interleaved appends**).

**Validation on `73916a6f2` (Release):** spill tests 10/10, fast suite 4329/4329, Corpus 319/319,
Markout `FieldWriter::.ctor` witness still renders `_options = options ?? new MarkoutWriterOptions();`
with no `S_0`/`S_1`/`Unsupported` residual.

Re-engaging both reviewers at `73916a6f2`.

richlander:
## Reconciliation — round 5 (head `f6b70bf0b`)

Rounds 4-5 converged on one hazard family: a **container node performs an observable
operation before or between its children**, so folding a non-pure spilled value into an
element reorders the value across that operation. Both reviewers found instances; I've
generalized the fix rather than patch node-by-node.

**GPT (NOT CLEAN → addressed):** `$"{S_0}"` — the FIRST interpolation hole takes the
`firstLeaf` path, which my round-4 (`precedingPure`-only) guard did not cover. I initially
argued this was dismissable (result-preserving, pre-existing). **GPT was right and I was
wrong on the key point:** GPT's follow-up repro reads `ArrayPool<char>.Shared.Rent`
*identity* (not undefined buffer contents) — handler construction consumes the seeded
buffer before the inlined call, so `SideEffect()` returns a different value and the produced
string differs (`spilled=42, inlined=17`). Array identity is unspecified but
**program-observable**, so this is a genuine faithfulness violation on the firstLeaf path.

**Gemini (NOT CLEAN → addressed):** `new int[] { S_0 }` (`ArrayLiteral`) and `[first, S_0]`
(`CollectionExpression`) allocate the container (`newarr` / `newobj List`) before any
element, and a collection expression may interleave `Add` / spread `MoveNext` (user code)
between elements. (Note: `ArrayLiteral` is reachable with non-constant elements via
`ArrayLiteralFromStoresPass`, not only the constant blob path.)

**Fix (`f6b70bf0b`):** promoted the hidden-operation check to a **universal guard**
`LoadCrossesHiddenOperation`, run ahead of BOTH the `firstLeaf` and `precedingPure` paths —
mirroring the existing universal `LoadIsUnconditionallyEvaluated` conditional guard.
`ChildFollowsHiddenOperation` now rejects every element of `InterpolatedStringExpression`,
`CollectionExpression`, and `ArrayLiteral` (in addition to `WithExpression` non-receiver and
`ObjectInitializerExpression` non-creation children). The now-redundant per-path call in
`PrecedingEvaluationIsPure` is removed; that walk only proves left-sibling purity.

Rejecting the **first** element is correct, not merely conservative — the container
allocation before it is program-observable (identity / pooled buffer / exception ordering).
And it is **free**: corpus 319/319 unchanged, so no clean raise regresses.

**Soundness model — three gated hazards for a deferred non-pure fold, two now universal:**
1. impure left siblings → `PrecedingEvaluationIsPure` (precedingPure path only; a first leaf
   has no preceding siblings);
2. conditional evaluation → `LoadIsUnconditionallyEvaluated` (universal);
3. hidden operation before/between children → `LoadCrossesHiddenOperation` (universal):
   `WithExpression`, `ObjectInitializerExpression`, `InterpolatedStringExpression`,
   `CollectionExpression`, `ArrayLiteral`.

**Validation (Release):** spill tests 13/13, fast suite 4332/4332, Corpus 319/319, Markout
`FieldWriter::.ctor` witness still renders `_options = options ?? new MarkoutWriterOptions();`.

Re-engaging both reviewers at `f6b70bf0b`.

richlander:
## Round-6 reconciliation — head `7021e5fcf` (parent `f6b70bf0b`)

**GPT — NOT CLEAN at `f6b70bf0b` (6th hazard, correct):** `StackAllocArray`
with an initializer evaluates its `Count`, performs `localloc`, then evaluates
its elements. The reserved stack address is program-observable — GPT's unsafe
canary showed the address offset changing (`spilled=-64, inlined=-80`,
`store-count=0`) when an effectful spill is folded into an element, with no
uninitialized-memory read. Same hazard class as `ArrayLiteral`/`CollectionExpression`
(container allocates before elements), and reachable with non-pure initializer
elements. Confirmed, not dismissed.

**Fix (`7021e5fcf`):** `ChildFollowsHiddenOperation` now rejects every non-`Count`
child of `StackAllocArray`; the `Count` child stays foldable because it is
evaluated before the `localloc`. Matches GPT's suggested fix exactly. Regression
test `NonPureSpill_InStackAllocArrayElement_IsNotInlined`.

**Also closed proactively in the same commit — `SpanLiteral`:** `CreateSpan`
runs before its elements, so it is structurally the same class. Its elements are
always decoded blob constants today (no non-pure value can currently reach one),
but I gated it too so the container-allocates-before-elements class is complete
against future producers rather than left as a latent whack-a-mole target.

**SAFE audit recorded in the doc comment (kept the `false` default):**
`AnonymousObject` and tuple literals construct (`newobj`) after all operands;
a plain `new T[n]` / `stackalloc T[n]` without an initializer allocates after
its size child; calls / `new T(args)` / local-function invocations / indexers
invoke last. Each runs its own operation strictly after all operands.

**Validation (Release):** spill 14/14, fast 4333/4333, Corpus 319/319; Markout
`FieldWriter::.ctor` witness still renders `_options = options ?? new MarkoutWriterOptions();`.

Re-review of `7021e5fcf` requested from both reviewers.

richlander:
## Round-6 verdicts — both CLEAN at `7021e5fcf` ✅

Two-model adversarial review (high-risk) complete at the fixed head `7021e5fcf`:

- **GPT (`gpt-5.6-sol`): CLEAN** — "Hidden-operation and conditional denylists
  are complete." Verified the stackalloc element store is retained, the `Count`
  child still folds, and the intended field-initializer fold still collapses.
- **Gemini (`gemini-3.1-pro-preview`): CLEAN** — independently audited the
  `stackalloc int[Size] { … }` IL (`push Size` → `localloc` → element eval),
  confirming the refined `Count`-child exemption is sound (`Count` precedes the
  `localloc`, elements follow it → `Stores left = 1`). Confirmed the denylist is
  the complete set of C# constructs that lower to interleaved intermediate IL
  side effects.

**Two-model convergence:** both reviewers independently found the same 6th node
(`StackAllocArray`) at `f6b70bf0b`, and both cleared `7021e5fcf` after the fix.

### Full review arc (all reproduced + fixed as follow-up commits)
1. **Core fold** `fe9d7381d` — spilled receiver + coalesce collapse (`this._f = x ?? new()`).
2. **Conditional-arm** `0cb59f403` (GPT High / Gemini Critical, converged) —
   `LoadIsUnconditionallyEvaluated` guard; non-pure spill could move into a `?:`/`??`/`&&`/`||`/`?.`/switch arm.
3. **`??=` firstLeaf + null-coalescing denylist** `e3ce63b6b` (GPT High pre-existing / Gemini Critical) —
   promoted the conditional guard to universal; completed all 4 `??=` forms.
4. **WithExpression / ObjectInitializer ordering** `ad92507f2` (GPT High) — `ChildFollowsHiddenOperation` (hidden clone/ctor before initializers).
5. **Interpolated-string holes** `73916a6f2` (GPT) — `DefaultInterpolatedStringHandler` appends interleave with hole eval.
6. **Universal hidden-op guard + ArrayLiteral/CollectionExpression + first hole**
   `f6b70bf0b` (GPT #5 first-hole via ArrayPool identity / Gemini ArrayLiteral+CollectionExpression) —
   `LoadCrossesHiddenOperation` runs ahead of both firstLeaf and precedingPure; measured free on corpus.
7. **StackAllocArray + SpanLiteral (class complete)** `7021e5fcf` (GPT #6 / Gemini, converged) —
   non-`Count` children gated; `Count` stays foldable.

Soundness model = three gated hazards for a deferred non-pure fold: impure left
siblings (`PrecedingEvaluationIsPure`), conditional evaluation
(`LoadIsUnconditionallyEvaluated`), and hidden operation before/between children
(`LoadCrossesHiddenOperation`). Pure values skip all guards.

**Final validation (Release):** spill 14/14, fast 4333/4333, Corpus 319/319;
Markout `FieldWriter::.ctor` witness renders `_options = options ?? new MarkoutWriterOptions();`.

Holding `Ready to merge` until the `test` CI job (currently pending) is green.

richlander:
Ready to merge

Both two-model reviews CLEAN at `7021e5fcf` (GPT `gpt-5.6-sol` + Gemini `gemini-3.1-pro-preview`); CI `test (ubuntu-26.04, linux-x64)` PASS (9m6s), `build-net10`/`csharp-diff-smoke`/`changes` PASS, path-gated jobs skipped as expected.

richlander:
Integrated current `origin/main` at merge commit `562a9e44e` to resolve the fixture append conflict. The resolution preserves main's newly appended fixture order and appends #3500's spilled-coalesce witnesses afterward; the PR diff remains limited to its original three files.

Local validation at `562a9e44e`: Release solution build passed; `SpilledReceiverCoalesceInliningTests` passed 14/14; `Area=Pass` passed 1477/1477.

Issue #3504, opened 2026-07-29
FidelityCheck.Evaluate: guard the two silent failure modes (unbounded sweep, zero-match filter)


richlander:
`FidelityCheck.Evaluate(string, …)` has two silent failure modes. #3495 removed the pain of one of them without removing the mode itself, and deliberately deferred both.

Neither is hypothetical: mode 1 is what produced seven 45-minute CI timeouts and five decompiler regressions (#3489–#3493) that landed unnoticed over months, and mode 2 is one typo away at every call site the same PR converted.

## Mode 1 — unfiltered whole-assembly evaluation degrades silently with growth

`Evaluate(string, …)` walks every `reader.TypeDefinitions` (`FidelityCheck.cs:387`) calling `EvaluateType`, which decompiles **and** Roslyn-recompiles each type. Cost is O(types in the assembly), per gate.

The bug that bit us was not a bad call site. It was **drift**: seven test classes passed their own test assembly, which grows with every test class anyone adds. Those call sites were fine at a few hundred types and fatal at 1,823 TypeDefs / 17,858 methods, and they look *identical* in both states. No API shape catches that — the code that was correct on Monday is the code that times out in March.

The failure was also maximally quiet. The weekly Deep Inspect lane hit the 6h job timeout and was recorded as **`cancelled`**, and a cancelled job does not satisfy `if: failure()` — which is why `deep-inspect.yml:577`'s `Report scheduled failure` step never fired across consecutive overruns. Nobody was told.

## Mode 2 — a filter that matches nothing passes vacuously

#3495 added `Func<string, bool>? typeFilter`. If a predicate matches zero types, `Evaluate` returns an empty list, and every downstream negative assertion — `DoesNotContain`, `Empty`, `All(...)`, count comparisons — passes while testing nothing.

The predicate is a raw string match against a spelling with no compiler check and a known trap: nested types are spelled `Outer.Inner` on the API side (`MetadataReaderExtensions.cs:33`) and `Outer+Inner` on the body side (`LibraryBodyIndex.cs:3201`). A perfectly reasonable `type => type == typeof(NestedTargetFixture.Inner).FullName` yields `…NestedTargetFixture+Inner`, matches nothing, and goes green. #3495 had to special-case exactly this call site.

Adversarial review of #3495 (two models, both clean) confirmed no *current* call site is vacuous — existing `Assert.Single`, count equality, and presence checks happen to cover it — but that is a property of today's call sites, not of the API.

[...]

PR #2105, opened 2026-07-02
Thread assembly resolver into the Analysis metadata path (#2098 slice 1)

richlander:
## Summary

Implements slice 1 of #2098: threads an `IAssemblyReferenceResolver` into the
Analysis metadata path so cross-assembly type metadata can be read on demand. This
is the **seam only** — a shared enabler for cross-assembly allocation sizing (#2096),
escape/type backing (#2064), and referenced-enum shape. **No perf-shape behavior
change.**

Per #2066, reading a referenced assembly's *metadata* via an injected resolver is
permitted and is **not** runtime assembly loading; the product stays NativeAOT- and
Roslyn-free and never loads/executes inspected code.

## What changed

- `LibraryBodyIndex.Open(string path, IAssemblyReferenceResolver? resolver = null)`
  overload (the 1-arg overload delegates with `resolver: null`).
- Resolver-aware `IndexBuilder` with a lazy, per-identity cache of referenced
  `MetadataReader`s (`AssemblyReferenceIdentity -> ReferencedAssemblyMetadata`),
  opened via `resolver.Resolve(...)`.
- Internal `TryResolveExternalTypeDefinition(TypeReferenceHandle)` resolving a
  cross-assembly type ref to `(definingReader, TypeDefinitionHandle)`, handling
  top-level + nested types with a recursion guard. **Internal only** (test-visible
  via `InternalsVisibleTo`); it is not exposed to shape code, and slice 2 will
  replace it with typed value-extractors so no live reader escapes (see #2098).
- Product wiring: `ApiOutputFormatter` passes `ApiCommand.PlatformAssemblyResolver`.

## Lifetime model

`Build()` is a one-shot eager pass returning immutable data; `LibraryBodyIndex`
holds no live reader. All `MetadataReader`s (target and cross-assembly) are disposed
when `Open()` returns — the cross-assembly readers share the target reader's existing
lifetime exactly. The `IDisposable` builder owns and disposes every opened
`PEReader`/stream (including the failure path). Consumers extract values during Build
and store only the extracted values — the pre-existing contract.

## Fail-honest

Null resolver, an unresolvable reference, or a reference without managed metadata
all yield `null` (negative results cached, not retried) — behavior identical to the
single-reader path.

## Validation

- `dotnet build src/dotnet-inspect -c Release -p:NoWarn=NU1507` — clean.
- `dotnet run --project src/ILInspector.Analysis.Tests -c Release` — 298/298 pass,
[...]

PR #3508, opened 2026-07-29
docs: add consolidated type, member, and API representation map

### LINKED ISSUE #3498: Consolidate type, member, and API representation into one owning design doc
There is no single document that answers "how does this repository represent a type, a member, and an API surface element, and which representation do I use when?" The answer is currently distributed across at least nine design docs, and the code carries **three distinct types named `TypeRef`** plus **two distinct `public enum TypeRefKind`**. A contributor adding a type-shaped parameter today has no way to find the right answer short of reading all nine docs, and the most recent example got it wrong twice in a row (see "How this surfaced").

We should write one owning document — `docs/design/type-member-api-representation.md` — that covers the motivating scenarios, the shape census with selection guidance, and the decisions that were made *against* certain shapes, then reduce the existing docs to references into it.

## Why one doc

The existing material is individually good and collectively unusable for the question a contributor actually asks. Each doc is correct within its own frame and silent about the frames next to it:

| Doc | What it owns | What it doesn't say |
| --- | --- | --- |
| `type-spelling-identity-display.md` | `TypeNode.Render()` vs `RenderCanonical()`; identity-vs-display as a root-cause conflation | Which shape a *new* consumer should take as a parameter |
| `finding-coordinates.md` | "Why there is no generic anchor"; the coordinate axes | That this reasoning does not forbid *typed* type identity |
| `member-target-resolution.md` | selector → resolver → anchor; API vs body vocabularies | The type-level equivalent, which does not exist |
| `member-body-substrate.md` | `filter → render` producer contract; scope-per-type | That its `TypeAnchor` is a scope, not an identity |
| `bounded-metadata-traversal.md` | `GetFullTypeName` traversal | Which of the two nested-type spellings it emits, and why |
| `csharp-member-recompilation.md`, `il-diff-canonicalization.md`, `implementation-diff.md`, `source-finding-producers.md` | consume `MemberAnchor` | — |

The failure mode is not that any doc is wrong. It is that answering one question requires knowing in advance which four of them to read, and the search terms that lead there (`TypeRef`, anchor, identity) each land in the wrong one.

## Motivating scenarios the doc must answer

These are concrete questions that currently have no single addressable answer:

1. **"I need a cheap predicate over types, evaluated before expensive work."** Selection, not identity. Wants to be approximate on the admit side and precise about "matched nothing." This is the #3495 case.
2. **"I need to name the type in a result that outlives the `MetadataReader`."** Durable identity. Cannot hold a `TypeDefinitionHandle` — handles are reader-scoped and die with the `PEReader`.
3. **"I need to compare two types for equality across assemblies/versions."** Structural equality, corelib facade canonicalization. `TypeRef`'s job.
4. **"I need to display a type to a human or an agent."** Presentation-refined: NRT `?`, `dynamic`, tuple element names.
5. **"I need to look a type up in XML documentation."** DocID-shaped grammar; `XmlDocText.SimplifyTypeName` already parses `T:System.String`.
6. **"I need to round-trip a type through compile-back."** Must preserve `fnptr` / `modreq` / `modopt`, which rules out shapes that normalize them away.
7. **"I need a type name that is stable across a JSON round-trip."** The persisted-vs-live split that `type-spelling-identity-display.md` calls identity's "two entry points."

Scenarios 1 and 2 are routinely conflated — they are the input and output sides of the same operation, and they want *different* shapes. The doc should say so explicitly, because the member layer already models this correctly (`MemberTargetSelector` in, `MemberAnchor` out) and the type layer has no such split.

## Shape census the doc must contain

With, for each, its home, what it is authoritative for, and what disqualifies it elsewhere:

- `TypeNode` — `src/ILInspector.Metadata/TypeNode.cs`. The fact owner. `Render()` display, `RenderCanonical()` structural (`:50`).
- `TypeRef` — `src/ILInspector.Analysis/TypeRef.cs:26`. "Semantic type identity for IL analysis. Display names are for humans; equality is structural." Carries `TypeRefKind`, `ElementType`, `TypeArguments`, trust bits.
- `TypeRefKind` — a **second, unrelated** `public enum` at `src/ILInspector.Decompiler/Pipeline/TypeRef.cs:6`.
- `TypeRef` — a **third**, private `sealed record TypeRef(string FullName, string Namespace, string SimpleName)` at `src/ILInspector.CSharp/CSharpDeclarationWriter.cs:1783`.
- `ApiType` — the API-surface element.
- `string TypeFullName` — the untyped fallback, in `MemberAnchor` (`MemberAnchor.cs:18`), `TypeDiff` (`ApiDiffAnalyzer.cs:161`), `MetadataFindings.cs:687`.
- The **two nested-type vocabularies**: API spells `Outer.Inner` (`MetadataReaderExtensions.cs:33`), body spells `Outer+Inner` (`LibraryBodyIndex.cs:3201`). `member-target-resolution.md` documents this as deliberate; nothing states it where someone writing a type predicate would find it.

Three types sharing the name `TypeRef` across three assemblies is itself worth a paragraph. Whether they should be renamed is out of scope here, but a reader who greps `TypeRef` needs to know they will hit three unrelated things.

## Why `TypeAnchor` was not used for any of the type shapes

This is the part most likely to be re-litigated, so it should be written down once with its reasoning rather than rediscovered. Three separate decisions point the same way, and none of them is "an anchor would be bad":

**1. `TypeAnchor` was proposed, and the role it named turned out to be a scope, not an identity.** `member-body-substrate.md:209`:

> The substrate formalizes it: open a scope per type (a `TypeAnchor`), resolve each selected `MemberAnchor` to a handle within it, and import bodies through the one scope — never load the assembly per member.

That is a loading-scope concept: one PE load and one `EnsureTypeMaps` per type. The same paragraph names what already fills the role — `MetadataSource : IDisposable`, which "loads the PE once and builds its type maps once … and `Project` already reuses it across every member of a type." So `TypeAnchor` was not rejected on identity grounds; the thing it named already existed under another name and did not need a new type. The doc should say this plainly, because the name survives in prose and reads today like a missing identity primitive.

**2. A generic anchor was considered and rejected on coordinate-space grounds.** `finding-coordinates.md`, "Why there is no generic anchor":

> Flattening these into `FindingAnchor(string)` would discard type, coordinate space, and authority while duplicating data already owned by producer payloads. […] A shared anchor belongs on the leaf only after at least two producers require the same validated semantics.

The doc should be explicit that this argues against a *semantics-free* anchor, not against typed type identity — the distinction is load-bearing and easy to over-read in the direction of "no new shapes, ever."

**3. Type spelling was solved by projection instead.** `type-spelling-identity-display.md` puts both spellings on the fact owner (`TypeNode.Render()` / `RenderCanonical()`) rather than minting an identity type, on the grounds that Metadata already "builds a `TypeNode` tree, flattens it to a display string, discards the structure, then makes every downstream consumer re-parse." An anchor would have been a fourth spelling; a canonical projection retires the re-parsing instead.

The doc should also record that same document's **round-2 caveat**, which is the specific reason "just use `TypeRef`" is not the general answer:

> `TypeRef` cannot simply move below Metadata. It carries Analysis-specific trust bits and its decoder *rejects* function pointers and custom modifiers (`TypeRefDecoder` → `Unsupported`) — precisely the `fnptr`/`modreq`/`modopt` shapes this design's pin **must** preserve.

That caveat is decisive for any compile-back or decompiler-fidelity consumer, whose entire job is preserving those shapes — and it is currently buried in a "Broader opportunity" section of a doc about tuples.

## How this surfaced

#3495 added a `Func<string, bool>? typeFilter` to `FidelityCheck.Evaluate`. It is a legitimate *selection* filter and the PR stands on its own, but the parameter is a new instance of the anti-pattern `type-spelling-identity-display.md` already names:

> multiple consumers recover a **structural** fact by string-matching a **display** spelling — the same anti-pattern, each independently fragile to any presentation refinement (NRT `?`, `dynamic`, tuples)

While reviewing the design, two wrong answers were proposed in sequence before the right frame was found: first a local `FidelityCheck.TypeName(Type)` helper (forbidden outright by `member-target-resolution.md`: "Do not add local selector, canonical-signature, fingerprint, or anchor-construction helpers in producers"), then routing the predicate through Analysis `TypeRef` (ruled out by the round-2 caveat above). Both were reached by reading real docs carefully; neither doc was the one that contained the disqualifying constraint. That is the strongest argument for consolidation.

Note also that `member-body-substrate.md` already establishes `filter → render` as the universal producer strategy — "A producer takes a **filter** (which members, and — within a member — which regions) and **renders** a **projection** of the selection." A type filter on a producer is therefore squarely in-pattern; the missing piece is only what *shape* the filter should take, which is exactly what the consolidated doc would answer.

## Proposed scope

1. Write `docs/design/type-member-api-representation.md` covering: the motivating scenarios, the shape census with a "use X when Y" selection table, the selector-vs-anchor (input-vs-output) split, the two nested-type vocabularies and who owns each, and the rejected-alternatives section above.
2. Reduce the nine existing docs to short references into it, keeping each doc's own mechanics local. Do not delete history — `finding-coordinates.md` and the round-2 review reconciliation in `type-spelling-identity-display.md` are load-bearing rationale.
3. Add a task-specific entry-point row to `AGENTS.md` pointing at the new doc for "type, member, or API identity."
4. Record the open question the census exposes: whether a type-level selector/resolver counterpart to `MemberTargetSelector`/`MemberTargetResolver` should exist, with typed `MissingType`/`AmbiguousType` diagnostics mirroring `MemberTargetDiagnosticKind`.

Item 4 is genuinely open and the doc should mark it as such rather than settle it.

richlander:
Closes #3498.

## What this does

Adds `docs/design/type-member-api-representation.md`: one owning document for
"which type, member, or API identity shape do I use, and when?"

The material existed, but spread across **ten** documents. Answering the
question required reassembling the census, the input-vs-output split, and the
rejected alternatives from scratch each time — and the reassembly is
error-prone, because several of the shapes look interchangeable and are not.

This is a **map, not a replacement**. Every existing document keeps its own
mechanics and gains a four-line pointer to the map. No rationale is removed and
no existing prose is rewritten.

## What it records

**There is deliberately no single canonical type spelling.** This is the
strongest and most re-litigated conclusion in the area, established as a
blocking round-2 review finding in `type-spelling-identity-display.md`: the
XML-doc id must *erase* NRT while the Member Index digest must *preserve* it, so
one spelling for both breaks XML-doc lookup for every nullable API. Identity is
N projections with different erasure contracts. The doc tabulates the three
persisted projections so the next "let's just have one canonical name" proposal
has to answer this first.

**Why `TypeAnchor` was not used for any of the type shapes** (asked directly in
#3498). It *was* proposed, at `member-body-substrate.md:209` — but read in
context it names a **loading scope**, not an identity, and the role was already
filled by `MetadataSource : IDisposable`. So it was never rejected on identity
grounds, which is exactly why the name survives in prose and reads today like a
missing primitive. A `TypeAnchor` in the identity sense fails separately, on the
no-single-spelling finding above. `TypeAnchor` appears exactly once in all of
`docs/`; that single mention has been costing readers real time.

**`TypeRef` unification is closed, not open.** `metadata-primitives.md`
("Decision (2026-06): stop after step 3") calls it "decisively wrong" and
`architecture.md` principle 9 records it as a committed product boundary —
`ILInspector.Analysis` keeps zero project references so it can ship standalone.
The doc carries the single documented trip-wire that would reopen it, so the
residual cost is understood as a search collision rather than a design defect.

**The two nested-type vocabularies.** API identity spells `Outer.Inner`; body
identity spells `Outer+Inner`. They agree on non-nested types and diverge
*silently* on nested ones — the highest-value fact here for anyone writing a
type predicate, and the mechanism behind the vacuous-match hazard tracked in
 #3504.

**The selector-vs-anchor split.** `MemberTargetSelector` in, `MemberAnchor` out.
The type layer has only the "out" half, spelled as a bare string. That asymmetry
is the single most common source of confusion in this area; it is recorded as
open question 1 rather than silently fixed.

## Boundaries

- **Docs-only.** No product or test code changes, and the document makes no new
  behavior claim — it cites existing sources for every claim.
- **Nothing deleted.** The two load-bearing rationales most at risk in a
  consolidation — `finding-coordinates.md`'s "why there is no generic anchor"
  and `type-spelling-identity-display.md`'s round-2 reconciliation — are
  preserved in place and quoted with their scope intact. In particular the doc
  is explicit that the generic-anchor rejection targets a *semantics-free*
  anchor and "should not be cited as though" it argued against typed identity.
- **Open questions are marked open**, not resolved by assertion.

## Validation

- `markdownlint` against the repository `.markdownlint.yaml`: **clean** across
  all 11 changed files.
- Every quotation re-verified against its source at `4d44b1c8` by grep; every
  referenced document path confirmed to resolve.
- Line-number citations (`TypeNode.cs:12/41/50`, both `TypeRef.cs` declarations,
  `member-body-substrate.md:209`) re-checked in this worktree rather than
  carried over from earlier notes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -64,6 +64,7 @@ documentation.
 | Output style | `docs/design/style-guide.md` |
 | Sections and selection | `docs/design/section-model.md` |
 | Metadata and API inspection | `docs/design/assembly-inspection-query.md` |
+| Type, member, or API identity | `docs/design/type-member-api-representation.md` |
 | PDB and source acquisition | `docs/pdb-acquisition.md` |
 | Source Finding producers | `docs/design/source-finding-producers.md` |
 | Package resolution and caches | `docs/design/version-resolution.md` |
--- a/docs/design/csharp-member-recompilation.md
+++ b/docs/design/csharp-member-recompilation.md
@@ -1,5 +1,9 @@
 # C# assembly round-trip testing
 
+> **Map:** [Type, member, and API representation](type-member-api-representation.md) is the entry
+> point for choosing a type, member, or API identity shape. This document owns
+> the details below.
+
 ## Summary
 
 This document proposes a modest tools-only capability for compiling C# artifacts
--- a/docs/design/finding-coordinates.md
+++ b/docs/design/finding-coordinates.md
@@ -1,5 +1,9 @@
 # Finding Coordinates
 
+> **Map:** [Type, member, and API representation](type-member-api-representation.md) is the entry
+> point for choosing a type, member, or API identity shape. This document owns
+> the details below.
+
 Finding keeps correspondence, ordering, and provenance separate. They answer
 different questions and do not share one generic string slot.
 
--- a/docs/design/il-diff-canonicalization.md
+++ b/docs/design/il-diff-canonicalization.md
@@ -1,5 +1,9 @@
 # IL diff canonicalization boundary
 
+> **Map:** [Type, member, and API representation](type-member-api-representation.md) is the entry
+> point for choosing a type, member, or API identity shape. This document owns
+> the details below.
+
 `IlBodyDiff` is a low-level body-diff substrate in
 `ILInspector.Instructions`. It compares decoded IL operations after a small
 amount of canonicalization, then projects producer-owned rows through
--- a/docs/design/implementation-diff.md
+++ b/docs/design/implementation-diff.md
@@ -1,5 +1,9 @@
 # Implementation Diff Boundary
 
+> **Map:** [Type, member, and API representation](type-member-api-representation.md) is the entry
+> point for choosing a type, member, or API identity shape. This document owns
+> the details below.
+
 `ImplementationDiff` is the product-side decompiled C# + IL/body + authored
 Source diff projection in
 `ILInspector.Research`. It is the reusable implementation-diff component for
--- a/docs/design/member-body-substrate.md
+++ b/docs/design/member-body-substrate.md
@@ -1,5 +1,9 @@
 # Member body substrate
 
+> **Map:** [Type, member, and API representation](type-member-api-representation.md) is the entry
+> point for choosing a type, member, or API identity shape. This document owns
+> the details below.
+
 How the product renders a type and its member bodies through **one producer
 contract** repeated at each layer, so that skeleton source, full source, the
 merged IL+C# view, and the implementation diff are built from parallel
--- a/docs/design/member-target-resolution.md
+++ b/docs/design/member-target-resolution.md
@@ -1,5 +1,9 @@
 # Member target resolution
 
+> **Map:** [Type, member, and API representation](type-member-api-representation.md) is the entry
+> point for choosing a type, member, or API identity shape. This document owns
+> the details below.
+
 Member target resolution is the typed seam between user selectors, API surface
 members, durable member anchors, and physical body evidence.
 
--- a/docs/design/source-finding-producers.md
+++ b/docs/design/source-finding-producers.md
@@ -1,5 +1,9 @@
 # Source Finding producers
 
+> **Map:** [Type, member, and API representation](type-member-api-representation.md) is the entry
+> point for choosing a type, member, or API identity shape. This document owns
+> the details below.
+
 Metadata owns the local, SRM-only observations available from PE and portable
 PDB data. Network source acquisition, checksum verdicts, decompiler source
 correspondence, and old/new product interpretation remain separate consumers.
--- a/docs/design/type-member-api-representation.md
+++ b/docs/design/type-member-api-representation.md
@@ -1,0 +1,369 @@
+# Type, member, and API representation
+
+> The owning document for "how does this repository represent a type, a member,
+> and an API surface element, and which representation do I use when?"
+> Consolidates material previously spread across ten design documents
+> ([#3498](https://github.com/richlander/dotnet-inspect/issues/3498)).
+
+Each layer's mechanics stay with that layer's document. This document owns the
+**map**: what shapes exist, what each is authoritative for, what disqualifies
+each elsewhere, and which alternatives were rejected and why.
+
+## The one-paragraph answer
+
+There is no single representation, and there is deliberately no single canonical
+spelling. A type is a *structured value* inside a layer and a *string* only when
+it escapes one. Identity is not one key but several **projections**, each with
+its own erasure policy, because the same type must spell differently for a
+correspondence digest than for an XML-documentation lookup. Pick the projection
+that matches your question, and never recover a structural fact by pattern-matching
+a display string.
+
+## Motivating scenarios
+
+Find your question here; the shape census below says what to use.
+
+| # | Question | Kind | Answer |
+| --- | --- | --- | --- |
+| 1 | "Cheap predicate over types, before expensive work." | Selection | Cheapest available spelling; guard for zero matches ([#3504](https://github.com/richlander/dotnet-inspect/issues/3504)) |
+| 2 | "Name a type in a result that outlives the `MetadataReader`." | Materialization | Resolved `TypeRef` or string — never a handle |
+| 3 | "Compare two types for equality across assemblies or versions." | Structural identity | `TypeRef` (the layer's own) |
+| 4 | "Show a type to a human or an agent." | Display | `TypeNode.Render()` |
+| 5 | "Look a type up in XML documentation." | Projection | XML-doc id projection — *not* the identity digest |
+| 6 | "Round-trip a type through compile-back." | Fidelity | A shape that preserves `fnptr`/`modreq`/`modopt` |
+| 7 | "Survive a JSON round-trip." | Persistence | A persisted projection key on `ApiMember` |
+
+Scenarios 1 and 2 are the ones most often conflated. They are the **input** and
+**output** sides of one operation and want different shapes: selection may be
+approximate on the admit side but must be loud about matching nothing;
+materialization must be durable and exact. The member layer models this split
+correctly — `MemberTargetSelector` in, `MemberAnchor` out. **The type layer has
+only the output half**, spelled as a bare string. That asymmetry is the single
+most common source of confusion in this area and is the open question recorded at
+the end of this document.
+
+## The rule that generates most of the others
+
+From `docs/decompiler-ir.md:15`:
+
+> Strings end at the printers. Inside the pipeline, a type is a `TypeRef`: a
+> structured, comparable value carrying assembly identity, definition token, and
+> shape.
+
+and `docs/decompiler-ir.md:20`:
+
+> Structured type identity must survive the pipeline: the moment a type degrades
+> to a string, every downstream consumer inherits the loss.
+
+This is the general form of `AGENTS.md`'s "Do not infer one from display text
+when a typed identity exists." Strings are a boundary format, not a working
+format.
+
+The boundary is real and is also structural. `docs/decompiler-ir.md:10`:
+
+> no analysis result that escapes a `MetadataSource`'s scope may hold metadata
+> handles — escaping results must be fully materialized (resolved `TypeRef`s,
+> strings, byte arrays).
+
+A `TypeDefinitionHandle` is an index into one `MetadataReader`. It is meaningless
+across readers and dead once the `PEReader` is disposed. So any result type that
+outlives the scope **cannot** hold one, and must materialize. Strings are a
+sanctioned materialization; a resolved `TypeRef` is the better one.
+
+## Shape census
+
+### `TypeNode` — the Metadata fact owner
+
+`src/ILInspector.Metadata/TypeNode.cs:12`. Holds every discriminator
+(`IsDynamic`, `IsNullableAnnotated`, tuple elements and `TupleElementName`) and
+emits two spellings:
+
+| Method | Line | Spelling | Example |
+| --- | --- | --- | --- |
+| `Render()` | `:41` | Display, presentation-refined | `(int count, string name)`, `dynamic`, `string?` |
+| `RenderCanonical()` | `:50` | Structural, name- and presentation-insensitive | `System.ValueTuple<int, string>`, `object`, `string` |
+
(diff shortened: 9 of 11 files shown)


richlander:
## Review reconciliation

Three fixed-head review rounds by **MAI-Code**, each in an isolated read-only worktree at an exact commit. Per `AGENTS.md` this PR sits in the "more than trivial, but not high risk" tier (docs-only, no behavior claim, no compatibility surface), which calls for a single MAI-Code review; it took three rounds because each round found something real.

| Round | Head | Verdict |
| --- | --- | --- |
| 1 | `faf67901` | 2 blocking — stale citation line numbers |
| 2 | `7dd7dedd` | 1 blocking — repeated a stale claim from another doc |
| 3 | `53e3da9b` | in progress |

### Round 1 — stale citations (upheld, and wider than reported)

Reviewer flagged two citations pointing at the wrong lines.

**Upheld, and the root cause was worse than the report.** The cause was self-inflicted: this PR inserts a 4-line `Map:` blockquote at the top of nine documents, which shifted every line number the new document cites *into those same nine files*. The citations had been captured before the inserts. So the defect was systematic, not two typos — **five** citations were affected, and the reviewer found two.

Corrected in `7dd7dedd`, each re-verified against the post-change file:

| Citation | Was | Now |
| --- | --- | --- |
| `docs/design/member-body-substrate.md` | 209 | 213 |
| `docs/design/csharp-member-recompilation.md` | 310 | 313 |
| `docs/metadata-primitives.md` | 10-11 | 14-15 |
| `docs/design/implementation-diff.md` | 107-110 | 113-116 |
| `docs/design/implementation-diff.md` | 115 | 119 |

Citations into `docs/decompiler-ir.md` and `docs/architecture.md` were already correct — neither received a `Map:` insert.

To stop this recurring I added a mechanical check that extracts every `path:line` reference from the document and resolves it against the working tree. It now reports **0 unresolved**, and it is what caught the three the reviewer missed.

**Round 1 non-blocking (adopted):** use repo-relative paths. Worth more than it sounds — `MemberIdentity.cs` and `TypeRef.cs` each name **two different real files**, so a bare filename could send a reader to the wrong one. Every citation is now fully qualified.

### Round 1 finding I did *not* accept as stated

Round 1 asserted that `src/ILInspector.Analysis/TypeRefDecoder.cs:197-199` maps function pointers and custom modifiers to `Unsupported`. **That citation is wrong** — those lines are recursion-depth and signature-blob-size guards. The actual handlers are at `:232` (`GetFunctionPointerType`) and `:233-234` (`GetModifiedType`). Round 2 independently confirmed this and resolved it in favour of the corrected citation.

The reviewer's *verdict* was right and its *evidence* was wrong. The document now cites the real handlers, so the claim is stronger than before: it rests on primary source rather than only on a design-doc quote.

### Round 2 — stale premise inherited from another document (upheld)

Reviewer found the document asserting that `ILInspector.Analysis` keeps **zero project references**.

**Upheld.** `src/ILInspector.Analysis/ILInspector.Analysis.csproj:25-28` declares four `ProjectReference` entries, including `ILInspector.Metadata` — one of the very type-identity layers the prose says `Analysis` avoids depending on.

The claim was not invented here; it was quoted from `docs/architecture.md:691`. `git log -S` dates the wording to #710 and the `ILInspector.Metadata` reference to #2105, so it has been stale for a long time and this PR propagated it.

Two things followed:

1. **I did not edit `architecture.md` or `metadata-primitives.md`.** `AGENTS.md` says to resolve which owner is authoritative rather than silently choosing one, and that correction belongs to those documents' owners, not to a consolidation PR. Filed as #3512.
2. **The document now flags the premise as stale, links #3512, and tells readers not to cite it.** The *conclusion* is unaffected and is re-rested on the argument that survives: `TypeRef` non-unification is load-bearing on a capability fact, not a dependency count — `Analysis`'s decoder resolves function pointers and custom modifiers to `Unsupported` (`src/ILInspector.Analysis/TypeRefDecoder.cs:232-234`), so a shared model "would have forced `Analysis` to keep its own anyway."

This is the consolidation doing its job: pulling ten documents into one view is what made a decade-stale premise visible.

### Verified clean across rounds

Round 2 independently confirmed several central claims that round 1 had not covered — the three-row projection table cell by cell against `ApiSurface.cs` and `ApiMemberIdentity.cs`, the `MemberCanonicalSignature` `M`/`P`/`F`/`E`-with-no-`T` claim, all five anti-pattern instances, the nested-type vocabulary split, and that the `Map:` inserts are purely additive (no deletions, no reordering).

### Evidence

- `markdownlint` against `.markdownlint.yaml`: clean, 11 files — matching the CI `markdownlint` job, which passes. All code jobs correctly skip (path-gated, docs-only).
- Mechanical citation resolver: 0 unresolved.
- Every quotation re-verified in a worktree rather than carried over from notes.

I will post readiness once round 3 reports clean.

richlander:
## Round 3 clean — Ready to merge

Final MAI-Code review on the exact head `53e3da9b`: **no blocking issues found.**

It independently re-derived the round-2 fix rather than taking my word for it — confirming the four `ProjectReference` entries, the `TypeRefDecoder.cs:232-234` replacement argument, the `git log -S` provenance for both #710 and #2105, that the disagreement with `architecture.md`/`metadata-primitives.md` is labelled and tracked rather than silent, that the fix caused no collateral markdown damage, and that every `path:line` citation resolves and supports its claim.

**Merge-blocking status:**

- CI green on `53e3da9b` (run [30490515088](https://github.com/richlander/dotnet-inspect/actions/runs/30490515088)): `markdownlint` and `changes` pass; all code jobs correctly skip as path-gated for a docs-only change.
- All three fixed-head review rounds resolved, reconciled [above](https://github.com/richlander/dotnet-inspect/pull/3508#issuecomment-5123323342).
- Closes #3498.

**Ready to merge**

### Non-blocking follow-up

- #3512 — `docs/architecture.md:691` and `docs/metadata-primitives.md` still assert the stale "zero project references" claim. Owned by those documents, deliberately not fixed here; this PR only stops propagating it.
- #3504 — the zero-match and type-count-budget guards for `FidelityCheck.Evaluate`, which is the live instance of the display-string-matching anti-pattern this document catalogues.
- Open questions 1 and 2 in the new document (whether a type-level selector should exist, and whether `MemberCanonicalSignature` should gain a `T:` form) are recorded as open, not proposed here.

PR #3510, opened 2026-07-29
Address a metadata table by its ECMA-335 index (#3468)

### LINKED ISSUE #3468: Metadata lens: hex table selection (0x02 as an alias for TypeDef)

richlander:
Closes #3468.

`-S "Metadata: 0x02"` and `-S "Metadata: TypeDef"` are now the same selector. This tool's own output prints hex tokens, so a reader following a `0x02000015` reference already has the table index in hand and should not have to translate it before asking for the table.

## The design decision

**The alias rewrites the input selector; it does not register a second section.** That is what makes the two spellings *one* section rather than two that happen to render alike. A hex alias in the catalog would print its own heading, sort independently in the section order, count separately under `--count`, and appear as a second `-D` entry. Rewriting at the boundary (`NormalizeMetadataTableAliases`, alongside the existing `--il-offset` and `--heap` normalizers and before `SelectResolver`) means the orderer, the heading, `--count`, the document schema, and the effective-section cache key only ever see the canonical name -- so those failure modes are unreachable, not merely untested.

**Divergence from the issue's note.** #3468 suggested `MetadataSectionNames.TryGetTable` "is the single lookup point today and is where the alias belongs". It is not, and it stays canonical-only: by the time any lookup runs the name is already canonical. Teaching `TryGetTable` hex would put alias resolution in two places and would make `IsMetadataSection` claim a name that selection resolution -- which matches against the canonical catalog -- would still reject.

Three rules follow from treating hex as an address rather than a name:

- **Hex carries its `0x`.** A bare `02` is a table *name* position; inferring a radix would let one spelling mean two things. Matches the `--heap` address rule.
- **Only projected tables resolve.** The alias map is derived from `MetadataTableProjector.ProjectedTables`, the same array the canonical names come from, so a table the projection does not cover cannot become selectable by index. The rejection names the projected tables in *both* spellings, since a caller who pasted an index is thinking in hex.
- **A metadata token is not a table.** `0x02000015` addresses a row; its high byte is the table. The byte-width parse rejects it rather than silently returning `TypeDef`.

## One deliberate behavior divergence

A bad index fails the run **even beside a selector that does match**, where an unknown *name* is tolerated (`-S "Metadata: NoSuchTable" -S "Metadata: TypeDef"` exits 0 today).

The tolerance rule exists for names that may exist in one inspected assembly and not another. A hex index outside the projection is not that -- no image can ever supply it -- so tolerating it would silently drop a selector the caller definitely got wrong. This matches the `--heap` coordinate rule, where input naming nothing is exit 1. Confirmed with the repo owner before landing, and pinned by `MetadataLens_BadHexTable_FailsEvenBesideAMatchingSelector`.

## Evidence

8 gates in `MetadataLensTests`, **mutation-proven in both directions**:

| Mutation | Killed | Survived (correctly) |
| --- | --- | --- |
| `canonical = resolved` -> `canonical = section` (rewrite suppressed) | all 6 rewrite gates | both rejection gates |
| rejection `return false` -> `return true` (bad hex passed through) | both rejection gates | rewrite gates |

The one-section gate was **vacuous on the first pass** -- it also passed when hex was merely ignored, since one heading is one heading either way. It now asserts the hex-alone precondition first, and dies under the rewrite mutation.

Suites, all on Windows:

- CLI: `Total: 2492, Failed: 2` -- both pre-existing and Windows-only (`Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown(readmePath: "images/logo.png.")`, `Layout_Count_CountsFilesRatherThanRenderedTreeLines`), confirmed reproducing in a clean worktree at unmodified `origin/main` during #3497 and green in ubuntu CI.
- `ILInspector.Metadata.Tests`: 998/0.
- `DotnetInspector.MetadataRendering.Tests`: 118/0.

## Compatibility

Purely additive at the input boundary. Every existing selector string resolves exactly as before -- the alias passes through anything that does not start with `Metadata: 0x`. No section, schema, or output shape changes.

## Docs

`docs/design/metadata-table-projection.md` gains a **Hex table selection** section (linked from the decisions list, which already anticipated this) and `README.md` gains the hex example plus the one-section rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -390,8 +390,9 @@ The two spellings are one section, not two: the hex form is an input alias, so
 the heading, section order, `--count`, and `-D` all answer in the canonical
 name. Hex must carry its `0x` — a bare `02` is a table *name* position — and
 only projected tables resolve, so an index outside the projection is an error
-naming what is available rather than an empty section. A full metadata token
-such as `0x02000015` addresses a row, not a table, and is rejected as one.
+naming what is available rather than an empty section. A table index is one
+byte, so it is one or two hex digits; a full metadata token such as
+`0x02000015` or `0x00000001` addresses a row, not a table, and is rejected.
 
 ### Heaps
 
--- a/README.md
+++ b/README.md
@@ -378,6 +378,21 @@ dotnet-inspect library MyLib.dll -S "Metadata: TypeRef" --columns Name --tsv
 dotnet-inspect library MyLib.dll -S "Metadata: MethodDef" --tsv --rows 100..199
 ```
 
+A table can also be named by its ECMA-335 index, since this tool's own output
+prints hex tokens and a reader following a `0x02000015` reference already has
+the index in hand:
+
+```bash
+dotnet-inspect library MyLib.dll -S "Metadata: 0x02"   # same as "Metadata: TypeDef"
+```
+
+The two spellings are one section, not two: the hex form is an input alias, so
+the heading, section order, `--count`, and `-D` all answer in the canonical
+name. Hex must carry its `0x` — a bare `02` is a table *name* position — and
+only projected tables resolve, so an index outside the projection is an error
+naming what is available rather than an empty section. A full metadata token
+such as `0x02000015` addresses a row, not a table, and is rejected as one.
+
 ### Heaps
 
 The four ECMA-335 heaps get sections of their own (`Metadata: #Strings`,
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -1006,4 +1006,4 @@ Resolved:
   one. See [Heaps are not tables](#heaps-are-not-tables).
 - **Table selection grammar.** Both, since output already prints hex tokens and
   users will paste them: `TypeDef` and `0x02` address the same table. See
-  [Hex table selection](#hex-table-selection).
+  [Hex table selection](#implemented-hex-table-selection).
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -907,6 +907,49 @@ caching state and lifetime questions to the projector. If `mdi`'s default path
 becomes a memory complaint, this table says to start with `Scalar`/`Flags`
 interning and the `Nil` singleton, and the probe says how to prove it moved.
 
+## Implemented: hex table selection
+
+`-S "Metadata: 0x02"` and `-S "Metadata: TypeDef"` are the same selector. The
+motivation is mechanical: this tool's own output prints hex tokens, so a reader
+following a `0x02000015` reference already has the table index in hand and
+should not have to translate `0x02` to `TypeDef` before asking for the table.
+
+The alias rewrites the **input selector**; it does not register a second
+section. That is the whole design, and it is what makes the two spellings *one*
+section rather than two that happen to render alike. A hex alias registered in
+the catalog would print its own heading, sort independently in the section
+order, count separately under `--count`, and appear as a second entry under
+`-D`. Rewriting at the boundary means everything downstream — the orderer, the
+heading, `--count`, the document schema, the effective-section cache key — only
+ever sees the canonical name, so those failure modes are not merely untested but
+unreachable.
+
+`MetadataSectionNames.TryGetTable` therefore stays canonical-only. Teaching it
+hex as well would put alias resolution in two places, and would make
+`IsMetadataSection` claim a name that selection resolution — which matches
+against the canonical catalog — would still reject.
+
+Three rules follow from treating the hex form as an address rather than a name:
+
+- **Hex carries its `0x`.** A bare `02` is a table *name* position, and
+  inferring a radix would let one spelling mean two things. This matches the
+  `--heap` address rule.
+- **Only projected tables resolve.** The alias table is derived from
+  `MetadataTableProjector.ProjectedTables`, the same array the canonical names
+  come from, so a table the projection does not cover cannot become selectable
+  by its index. The rejection names the projected tables in both spellings,
+  because a caller who pasted an index is thinking in hex.
+- **A metadata token is not a table.** `0x02000015` addresses a *row*; its high
+  byte is the table. The byte-width parse rejects it rather than silently
+  reading the high byte and returning `TypeDef`.
+
+A bad index fails the run even beside a selector that does match — a deliberate
+divergence from the unknown-*name* rule, which tolerates a miss when something
+else matched. That tolerance exists for names that may exist in one inspected
+assembly and not another; a hex index outside the projection is not that, since
+no image can ever supply it. Tolerating it would silently drop a selector the
+caller definitely got wrong.
+
 ## Layer placement
 
 The projection lives in the **Metadata layer** (`ILInspector.Metadata`), beside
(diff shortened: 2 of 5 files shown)


richlander:

Two reviews at `c010a9dc`, both in isolated detached worktrees: **MAI-Code** and **GPT-5.6 Sol**. Two-model tier, since this parses untrusted input and diverges from an existing selection error policy. Both findings are real; both are fixed in `1f73908d`.

### 1. Static discovery bypassed the alias entirely -- Medium

*Found independently by **both** reviewers.*

`NormalizeMetadataTableAliases` sat below the early `-D`/`--schema` return in `ExecuteAsync`, so:

```
library <dll> -D "Metadata: 0x02" --schema --tsv   ->  exit 1, "Section 'Metadata: 0x02' not found."
library <dll> -D "Metadata: TypeDef" --schema --tsv ->  exit 0
```

**Verified and fixed.** I hoisted the normalizer *above* that branch rather than patching the branch, because "every consumer of a selector sees canonical names" is the invariant the whole design rests on -- patching one path would have left the next early return to rediscover this. My existing discovery gate missed it precisely because it supplied an input source and so took the later, effective-discovery path.

Two new gates: `MetadataLens_StaticSchemaDiscovery_ResolvesHexTables` and `MetadataLens_DiscoveryWithoutAnAssembly_ResolvesHexTables`.

### 2. A token whose value fits in a byte resolved as the wrong table -- Medium

*Found by **GPT-5.6 Sol**, in response to the "find a third mutation the suite misses" ask.*

`byte.TryParse` enforces numeric *range*, not textual *width*. So the eight-digit `0x00000001` -- a well-formed Module **row token** -- resolved as table `0x01`, TypeRef, at exit 0.

This is the sharper half of the finding: my close-negative case `0x02000015` was being rejected because it *overflowed a byte*, not because it was recognized as a token. It never covered the rule it was written for. I got lucky on the one case I picked.

**Verified and fixed.** Width is now checked textually -- a table index is one byte, hence one or two hex digits. Four new rejection cases whose low byte *is* a projected table: `0x00000001`, `0x02000002`, `0x80000002`, `0x0002`.

### Mutation results

GPT supplied the exact mutation its finding predicted would survive:

```csharp
string digits = suffix.Length > 4 ? suffix[^2..] : suffix[2..];
```

It passed **all 13** gates before this commit. It now **fails 4**.

Moving the normalizer back below the static-discovery branch kills exactly the two new static gates and leaves the effective-discovery gate passing -- the correct discrimination, since that gate legitimately exercises the later path.

### Attacks that found nothing

Both reviewers probed and cleared: sign prefixes, leading/trailing whitespace, unicode digits, embedded nulls, culture sensitivity, over-long inputs, `0X` casing, `0x` with no digits; whether an unprojected table can become selectable (it cannot -- `HexAliases` is derived from `MetadataTableProjector.ProjectedTables`); the `ReferenceEquals` rewrite-detection in `ResolveTableAliases`; normalizer ordering against `--il-offset` and `--heap`; and divergence between the two spellings across `--count`, `-D`, `--tsv`/`--jsonl`, `--columns`, `--rows`, section ordering, and the discovery cache key. Notably, the cache-ordering hazard that #3497's review turned up does **not** recur here, because the alias is a pure input rewrite resolved before any cache read.

### Validation at `1f73908d`

CLI `Total: 2498, Failed: 2` (both pre-existing and Windows-only, green in ubuntu CI); `ILInspector.Metadata.Tests` 998/0; `DotnetInspector.MetadataRendering.Tests` 118/0; `DotnetInspector.Services.Tests` 343/0.

Fixed-head re-review of `1f73908d` is in flight with both reviewers.

richlander:
## Fixed-head re-review: both clean

Both reviewers re-reviewed `1f73908d` in their isolated worktrees. **Neither found a remaining product issue.**

**MAI-Code** confirmed its original repro now exits 0 and matches the canonical schema output; verified the width rule accepts `0x02`/`0x2`/`0x00` and rejects all four token forms; and confirmed the hoist introduced no regression across `--il-offset`, `--heap`, bare `-S`, `-S @All`, `-S @Metadata`, `-D @Hidden`, `--schema`, and mixed combinations -- noting specifically that the IL and heap cases still fail at the coordinate-requirement stage rather than misrouting through alias handling.

**GPT-5.6 Sol** independently confirmed both fixes, cleared padded/token-width/signed/Unicode/embedded-space/over-long forms, and agreed that rejecting `0x0002` is consistent with the documented one-byte grammar rather than a lost spelling.

### Each reviewer then found one surviving mutation -- both test-only gaps, both now closed

**Three-digit width** (GPT). `digits.Length is 1 or 2` -> `1 or 2 or 3` passed all 19 gates: my padding coverage used four digits (`0x0002`) but never three. Added `0x002`; the mutation now fails.

**Case-sensitive prefix** (MAI). `OrdinalIgnoreCase` -> `Ordinal` on the prefix check passed every gate, because every case used the canonical `Metadata:` casing. That mutation would have made `metadata: 0x02` the one spelling that silently stopped working while `metadata: TypeDef` kept working -- a real asymmetry, since section matching is case-insensitive everywhere else in this CLI. Added lowercase and uppercase prefixes on both the accept and reject paths; the mutation now fails 3 cases across 2 gates.

Both are covered in `879743ed`, which touches **only** `MetadataLensTests.cs` -- `git diff 1f73908d..879743ed -- src/dotnet-inspect/` is empty, so the product code both reviewers cleared is byte-identical to what is being merged.

`d8db74a9` then fixes an MD051 link fragment CI caught: I wrote `#hex-table-selection` for a heading whose anchor is `#implemented-hex-table-selection`. Every intra-document fragment in both changed files now resolves to a real heading.

### Final validation at `d8db74a9`

CI all green: `build-net10` pass, `markdownlint` pass, `test (ubuntu-26.04, linux-x64)` pass.

Local (Windows): CLI `Total: 2502, Failed: 2` -- both pre-existing, Windows-only, and green in the ubuntu lane above. `ILInspector.Metadata.Tests` 998/0, `DotnetInspector.MetadataRendering.Tests` 118/0, `DotnetInspector.Services.Tests` 343/0.

### Mutation coverage, final

| Mutation | Result |
| --- | --- |
| `canonical = resolved` -> `canonical = section` | kills all 6 rewrite gates |
| rejection `return false` -> `return true` | kills both rejection gates |
| normalizer moved back below static discovery | kills exactly the 2 static gates, effective-discovery gate survives |
| `suffix.Length > 4 ? suffix[^2..] : suffix[2..]` | kills 4 token cases |
| `digits.Length is 1 or 2 or 3` | kills `0x002` |
| prefix `OrdinalIgnoreCase` -> `Ordinal` | kills 3 cases across 2 gates |

Ready to merge

Issue #3444, opened 2026-07-29
mdi renders untrusted metadata without the containment #3380 gives the CLI


richlander:
`src/mdi` is a second entry point that reads the same untrusted .NET metadata the
main CLI reads, and renders it to stdout and stderr. It is deliberately **out of
scope** for #3319/#3380, and this issue records that boundary so it is not
mistaken for coverage.

## Why it is excluded, mechanically

`CommandError_IsTheOnlyWriterOfTheErrorPrefix` derives its scan scope from the
CLI's transitive `ProjectReference` closure. `mdi` has its own `Program.cs` and
is not in that closure, so the gate asserts `Assert.DoesNotContain("mdi", names)`
— the exclusion is a property of the project graph, not a hand-kept deny list.
That is correct for the gate (code in `mdi` cannot reach `CommandError`), but it
means `mdi` inherits none of the containment work.

## What is unhandled

`src/mdi/MdiCommand.cs` writes to `Console.Error` at ~13 sites with raw string
interpolation, and its renderers take `TextWriter` sinks:

- Argv-derived text: `$"Error: unknown format '{formatText}'."` (line 120),
  `$"Error: {heapError}"` (167), `$"Error: {specError}"` (196).
- Metadata-derived text: the `ExecuteOverview` / `ExecuteHeapValue` /
  `ExecuteReferences` / `Execute` renderers write heap strings, table rows, and
  user-string blob contents — the exact untrusted surface #3319 is about — in
  `md`, `tsv`, and `jsonl` shapes.

A user-string heap entry carrying CR/LF, U+000B, U+001B, or a bidi override can
therefore escape its row or line in `mdi` output the same way it could in the
CLI before #3380.

## Suggested shape

Do **not** reach into `DotnetInspector.Core`'s CLI writer; that would drag the
closure boundary across a tool that is intentionally standalone. Either:

1. give `mdi` its own small stream owner with the same fold-and-contain
   contract, gated by its own ownership test scoped to `mdi`'s closure; or
2. decide `mdi` is a developer-only debug tool whose output is not a trust
   boundary, and say so in its own docs — in which case close this as by design.

Option 2 is a legitimate answer, but it should be an explicit decision rather
than an unexamined gap.

Tracked as declared-unverified boundary #6 in #3380.

richlander:
Closed by #3518 (merged as `ce9a508d`) — but the premise here was mostly wrong, and the correction is worth recording.

[...]

PR #3518, opened 2026-07-29
Contain the metadata version stamp, and gate mdi's inherited containment (#3444)

richlander:
## What this does

Fixes a terminal control injection leak in the metadata image overview, and
gates the containment property `mdi` inherits from the projection.

The PR started as tests and docs only. Adversarial review found a real
product defect that contradicted the gate's central claim, so it now changes
product code in the shared metadata layer.

## The leak

The metadata root's version stamp is a counted string read straight out of the
image -- attacker-controlled in exactly the way heap strings are. Unlike them,
it reaches presentation without passing through the row projection:
`MetadataImageInspector.Describe` reported `reader.MetadataVersion` verbatim
and `MetadataProjectionRenderer.ImageFactRows` emitted it raw.

Both `mdi --overview` and the shipping CLI's `Metadata: Image` section were
affected, in Markdown and TSV. Measured on a real assembly whose version stamp
was length-preservingly patched to `v` + `ESC [ 3 1 m` + `BEL` + `DEL` + `U+009F`:

| | raw ESC | raw BEL | raw DEL |
| --- | --- | --- | --- |
| `dotnet-inspect library <patched> -S "Metadata: Image"` before | 1 | 1 | 1 |
| after | 0 | 0 | 0 |

After the fix the row reads `| Metadata version | v\u001B[31m\u0007\u007F\u009F |`.

The fix neutralizes the stamp in the metadata layer, reusing
`MetadataTableProjector.NeutralizeControls` rather than adding a parallel
escaper, so every consumer inherits it. That method becomes `internal` for
this one caller. The budget cannot truncate a conforming value: ECMA-335
bounds the field at 255 bytes and neutralization expands a control character
to six characters.

## Why the gate exists

`mdi` performs almost no escaping of its own -- it hands every view to
`MetadataProjectionRenderer`, which renders values `MetadataTableProjector`
already neutralized. That makes `mdi` the reference example of consuming the
projection safely, and it also makes the safety invisible at the call site:
nothing in `mdi` would fail if a value reached presentation without being
projected. The version stamp was exactly that value. `MdiContainmentTests`
fails instead.

The fixture is a real assembly, not a synthetic string, per the `AGENTS.md`
evidence rule -- the claim is about what reaches a terminal when a hostile
*artifact* is inspected, so the payload travels the whole decode-project-render
path. Both splices are byte-length preserving, so no offset in the image moves.

Coverage is driven from `MetadataTableFormat` itself, so a new rendering format
is gated on arrival rather than added to a hand-written list.

## Mutation evidence

| mutation | result |
| --- | --- |
| `IsControl => false` | 6 kills (Markdown + TSV across table, heap, overview) |
| `IsControl(c) => c == '\x1B'` | 6 kills |

JSONL legitimately survives both: `MarkoutTableMode.Jsonl` escapes
independently, so the projector is pinned by Markdown and TSV and the JSONL
writer is pinned by JSONL. Two layers, each with its own gate.

## Validation

- `DotnetInspector.MetadataRendering.Tests` 131/131
- `ILInspector.Metadata.Tests` 998/998
- `dotnet-inspect.Tests` 2502 total, 29 failed -- all pre-existing: the
  missing-`ilasm` fixture set plus the known Windows-only failures. Confirmed
  identical at the unmodified previous head (`ProgressiveMemberCallGraphTests`
  10/10 failing there with a clean tree).
- `markdownlint` on the three changed Markdown files (CI).


richlander:
## Review reconciliation (head `4998d0d3` -> `5639c938`)

Two independent adversarial reviews on the fixed head `4998d0d3`, isolated in
separate review worktrees: **GPT-5.6 Sol** and **MAI-Code**. Six findings
between them; four actioned, one already satisfied, one dismissed with
reasoning.

### GPT-5.6 Sol

**High -- the overview leaks raw ESC. Confirmed, and it contradicted this PR's
own thesis.** `MetadataImageInspector.cs:45` passed `reader.MetadataVersion`
into the overview and `MetadataProjectionRenderer.ImageFactRows` emitted it
verbatim, so the version stamp reached a terminal without ever meeting the
containment the PR claimed was inherited. I reproduced it independently on a
patched real assembly and confirmed the shipping CLI shared it, not just `mdi`.
Fixed at the metadata layer in `5639c938`; before/after byte counts are in the
PR description. This is the finding that turned the PR from tests-and-docs into
a product change.

**Medium -- the gate was ESC-only.** Demonstrated by mutating
`IsControl(c) => c == '\x1B'`, which passed all thirteen tests while BEL, DEL,
and C1 controls leaked raw. Correct. The payload now spans all three ranges
`IsControl` recognizes, and that mutation kills six cases.

**Medium -- the fixture check was vacuous.** It scanned the whole image for the
payload and matched the `Payload` array initializer's own bytes in this test
assembly, so it passed against an *unpatched* fixture -- it was confirming its
own source code. Correct and cited with offsets. The check now verifies the two
exact splice offsets and asserts the unpatched assembly differs there.

I also reclassified the overview case on my own account while fixing the above:
it had been labelled a coordinates-only regression net, which was simply false,
and that mislabelling is a fair part of why the leak sat unobserved. Only
`--references` remains a net now, and the file says why.

### MAI-Code

**Medium -- `CanaryTail` hardcoded a 5-byte payload length.** Same defect GPT
flagged, found independently and by a different route (MAI changed the payload
to 6 bytes and observed the lookup throwing rather than the assertions
failing). Fixed: derived from `Payload.Length`, so the two cannot drift.

**Low -- the hostile fixture leaked into `%TEMP%` on every run.** Correct and
verified; I had deferred it pending review. Fixed: the fixture is now an
`IClassFixture` that deletes its assembly in `Dispose`. Confirmed zero
`mdi-containment-*.dll` files remain after a full run.

**High -- "the gate can be made to pass by weakening `AssertNoRawControlCharacters`."
Dismissed.** The reported experiment mutated the *test helper* while leaving the
product correct, then observed the suite still green. That is the expected
result and not evidence of a weak gate: with containment working there is no raw
ESC in the output, so removing the check for one changes nothing. The mutation
removes a check for something that is not there. A helper-weakening mutation is
survivable against *any* correct product, so it cannot distinguish a strong gate
from a weak one -- only product-side mutations can, and those are tabulated in
the PR description. The underlying suggestion, asserting the escaped form is
present rather than relying on absence alone, was already implemented at the
reviewed head and remains so: every payload-carrying case asserts all four
neutralized forms, which is what stops it passing by rendering nothing.

### Next

Both reviewers are being re-run against the fixed head `5639c938`. Not marking
ready until both are clean.

richlander:
## Review reconciliation, round 2 (head `e6376891` -> `a462a29a`)

Fixed-head re-reviews. **MAI-Code: clean, no significant issues.**
**GPT-5.6 Sol: two Medium findings, both legitimate, both fixed.**

### GPT-5.6 Sol -- oversized metadata versions are silently truncated (Medium)

Correct, and the more interesting of the two because my own justification was
the bug. `NeutralizeVersion` discarded the `truncated` flag with `out _`, and I
had defended that in a doc comment on the grounds that "the budget cannot
truncate a conforming value" -- ECMA-335 caps the field at 255 bytes and
neutralization expands a control character sixfold, so 255 * 6 exactly
accommodates any conforming stamp.

That is true and beside the point. The stamp is a counted string read straight
out of the image, so conformance is precisely what this code does not get to
assume. Discarding the flag also made it the only truncation in the layer that
is not surfaced: `MetadataValue.HeapReference.Truncated`,
`HandleRef.DisplayTruncated`, `EntriesTruncated`, and `RowsTruncated` are all
typed facts, and `FormatHandle` carries an explicit comment that a truncated
display "must always carry the ellipsis so it is never mistaken for a whole
value."

I could not initially reproduce it -- `MetadataRootBuilder` rejects an over-long
version even with `suppressValidation: true`, so no emitter can produce this
input. I asked for the fixture and the reviewer supplied a complete builder: a
conforming image is built first, then the version field is widened in place with
every position-recording structure repaired -- stream header offsets, the CLI
header's metadata size, the containing section's virtual and raw sizes, the raw
pointers of later sections, and `SizeOfCode`. I ran it and reproduced exactly:

```text
HasMetadata: True
MetadataVersion.Length: 1547
Module name: LongVersion.dll
#Strings bytes: 29
```

So the image is malformed-but-fully-parseable -- reachable in exactly the
population this code exists to survive. Fixed: `MetadataImageOverview` now
carries `MetadataVersionTruncated`, and `ImageFactRows` appends the ellipsis the
way `FormatHeap` does. Rendered stamp goes from 1530 characters unmarked to 1530
plus the marker.

The fixture is now `OversizedMetadataVersionTests`, gating both halves -- that
`Describe` sets the flag, and that the marker survives to `mdi`'s rendered
output. Forcing the flag to `false` kills both cases. This one mattered to
include rather than assert in prose: without a reachable input, the flag would
be indistinguishable from a constant `false` and no test could tell.

Thanks for writing the fixture out in full -- that turned an unprovable claim
into a gate.

### GPT-5.6 Sol -- threat model overstates containment coverage (Medium)

Correct, and mine to own. Both the threat-model row and the class doc comment
said every `mdi` view asserts every neutralized form. The `--references` view
carries no artifact text -- it renders coordinates only -- so it asserts the
absence of raw controls and nothing else. Measured: `neutralizedForms=0/4` in
all three formats.

Both now scope the claim to the three payload-carrying views (table, heap,
overview) and describe `--references` as the coordinates-only regression net it
is. I asked whether it should instead be made payload-carrying; the reviewer
recommended relabelling, on the grounds that resolving names there would change
product semantics solely to strengthen a test. Agreed, and that is what I did.

### Correction to my earlier validation claim

I previously reported ~29 pre-existing `dotnet-inspect.Tests` failures and
attributed them to a missing `ilasm`. That was wrong. They were stale fixtures
from building a dirty worktree. A clean build at this head reproduces only the
two known Windows-only failures:

- `Layout_Count_CountsFilesRatherThanRenderedTreeLines`
- `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown(readmePath: "images/logo.png.")`

### Validation at `a462a29a`

- `DotnetInspector.MetadataRendering.Tests` 137/137
- `ILInspector.Metadata.Tests` 998/998
- `dotnet-inspect.Tests` 2502 total, 2 failed (the known pair above)
- Mutations killed: `IsControl => false` (6), `IsControl => c == '\x1B'` (6),
  ellipsis removed from `ImageFactRows` (1), truncation flag forced `false` (2)

Both reviewers go back to this head. Not marking ready until both are clean.

richlander:
## Review reconciliation, round 3 (head `a462a29a` -> `43d0aa52`)

One finding from GPT-5.6 Sol, accepted and fixed. Test-only change.

### The fixture did not gate the `SizeOfCode` repair it claimed was required (Medium)

Correct, and this is the third time in this PR that a claim turned out broader
than its evidence — the pattern is worth naming.

The fixture's doc comment said that skipping any one of the repairs "yields an
image SRM refuses, which is why the repair is this specific." The reviewer
tested that directly, deleted the `SizeOfCode` update, and got:

```text
Total: 3, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0
```

All green, with the image now carrying an optional header that contradicts its
own section table. SRM never reads `SizeOfCode`, so nothing noticed. By
contrast, omitting the stream-offset repair does fail loudly:

```text
System.BadImageFormatException: Unknown tables: 0x4141414141414141
```

— the version padding being read as a table mask.

That contrast is the actual structure of the fixture, and the doc now says so:
most repairs enforce themselves because `MetadataReader` refuses the image
without them, and exactly one does not. Recording the `BadImageFormatException`
is part of the fix, because it is what makes the rest of the fixture
self-checking and shows why only one repair needed a test.

`SizeOfCode` is now gated by
`HostileImage_KeepsSizeOfCodeConsistentWithItsSectionTable`, asserting that it
equals the sum of the code sections' raw sizes. I anchored the invariant on a
compiler-produced assembly first, so it is grounded in what real toolchains emit
rather than a rule invented to suit the fixture — it holds exactly for both:

| image | `SizeOfCode` | sum of code-section raw sizes |
| --- | --- | --- |
| a compiler-produced assembly | 7680 | 7680 |
| the hostile fixture | 2560 | 2560 |

Repeating the reviewer's exact experiment now fails:

```text
OversizedMetadataVersionTests.HostileImage_KeepsSizeOfCodeConsistentWithItsSectionTable [FAIL]
Total: 4, Errors: 0, Failed: 1
```

### On the pattern

Three rounds, three findings of the same species: the code was fine, the
*claim* about the code was not. First the design doc said containment was
complete when the version stamp bypassed it. Then the threat model said every
view asserts every neutralized form when `--references` asserts one. Now a
fixture said every repair is load-bearing when one was unchecked. Each was found
by someone testing the sentence rather than the code — deleting the thing the
sentence said was necessary and watching nothing fail.

That is the repository's own rule in `AGENTS.md` — an asserted property names
the gate that enforces it, and "unverified" is an acceptable answer where an
unmarked ungated claim is not. Worth restating that a fixture's own invariants
fall under that rule too, not just product properties.

### Validation at `43d0aa52`

- `DotnetInspector.MetadataRendering.Tests` 138/138
- No product code changed since `a462a29a`, so the metadata and CLI suites are
  unaffected; their results at that head stand (998/998 and 2502 with the two
  known Windows-only failures).

Both reviewers go to this head.

richlander:
## Review reconciliation, round 4 (head `43d0aa52` -> `f24683cf`)

Both reviewers found the same gap from different directions — GPT-5.6 Sol as a
Medium, MAI-Code as a Low. Both accepted. Test-only change.

### The `SizeOfCode` gate was real but too narrow (Medium, GPT-5.6 Sol / Low, MAI-Code)

`SizeOfCode == sum of the code sections' raw sizes` is a *self-consistency*
check, and self-consistency survives coupled corruption. GPT-5.6 Sol dropped
both the section's raw-size growth and the `SizeOfCode` growth together, and the
equality still held at the old values while the image fell apart:

```text
Total: 4, Errors: 0, Failed: 0
versionChars=1547 moduleRows=1
SizeOfCode=1024 codeRawSum=1024
metadataEnd=2360 textRawEnd=1536
```

Metadata ending at 2360 inside a section whose raw data stops at 1536, and every
test green. Separately, MAI-Code deleted the later-section raw-pointer repair and
also got a clean run — a second ungated repair, which the round-3 doc comment had
implicitly claimed SRM enforced.

The fix asserts properties the corruption actually violates instead of one it can
satisfy. Two new gates:

- `HostileImage_KeepsMetadataInsideItsOwningSection` — the declared metadata
  range must lie within the raw range of the section that holds it.
- `HostileImage_KeepsSectionsAlignedAndDisjointOnDisk` — sections start on a
  file-alignment boundary, do not overlap, and do not run past the image.

Both are anchored against a compiler-produced assembly first, as the `SizeOfCode`
invariant already was.

### The doc comment, now sorted by who enforces what

The round-3 comment said SRM refuses the image "if the stream offsets or the
section geometry are wrong." That put the virtual-size repair in neither category
and overstated the rest, which is exactly how the raw-pointer repair went
unnoticed. I measured each repair rather than reasoning about it:

| repair | enforced by | how it fails |
| --- | --- | --- |
| stream offsets | SRM | `BadImageFormatException: Unknown tables: 0x4141414141414141` |
| containing section virtual size | SRM | `BadImageFormatException: Section too small.` |
| containing section raw size | test | `KeepsMetadataInsideItsOwningSection` |
| later sections' raw pointers | test | `KeepsSectionsAlignedAndDisjointOnDisk` |
| `SizeOfCode` | test | `KeepsSizeOfCodeConsistentWithItsSectionTable` |

### Mutation matrix

Each repair deleted in turn, against the new head. Both reviewer cases are killed:

| mutation | before | after |
| --- | --- | --- |
| drop raw size | — | 2 failed |
| drop `SizeOfCode` | 1 failed | 1 failed |
| drop raw size + `SizeOfCode` | **0 failed** | 1 failed |
| drop later raw pointers | **0 failed** | 1 failed |
| drop virtual size | — | 5 failed (SRM) |

### On the pattern, again

Four rounds, four findings, every one of them a claim that outran its evidence
rather than a bug in the code. This round is the sharpest version: I wrote a gate
*specifically* to stop overclaiming, and the gate itself was narrower than the
sentence it was justifying. A check that the corruption can satisfy is not a gate
for that corruption, however precisely it is described.

Worth stating plainly since `AGENTS.md` asks that an asserted property name the
gate enforcing it: naming a gate is not sufficient, the gate has to be violated by
the thing the property forbids. Both reviewers found this the same way — by
deleting the thing the prose said was necessary and watching nothing fail.

### Validation at `f24683cf`

- `DotnetInspector.MetadataRendering.Tests` 140/140
- No product code changed since `a462a29a`, so the metadata and CLI suites are
  unaffected; results at that head stand (998/998, and 2502 with the two known
  Windows-only failures).

Both reviewers go to this head.

richlander:
## Review reconciliation, round 5 (head `f24683cf` -> `59933eb4`)

Both reviewers found a distinct gap in the round-4 gates. Both accepted, both
fixed. Test-only change.

### `SizeOfRawData` alignment was unchecked (Medium, GPT-5.6 Sol)

The reviewer grew the section and `SizeOfCode` by `expansion - 1`. Every
relationship the gates checked stayed intact:

```text
fileAlignment=512 textRawSize=2559 remainder=511
SizeOfCode=2559 codeRawSum=2559 metadataEnd=2360 textRawEnd=3071
Total: 6, Errors: 0, Failed: 0
```

Metadata inside its section, sections disjoint, `SizeOfCode` matching the
section table at 2559 == 2559 — and a `.text` raw size of 2559 against a file
alignment of 512, which PE/COFF does not permit.
`AssertSectionsAreAlignedAndDisjoint` checked `PointerToRawData` alignment but
never `SizeOfRawData`. It does now.

This is the same defeat as round 4, one level up: a gate written against the
last corruption, beaten by an edit that keeps the checked relationship intact.

### `SizeOfInitializedData` was unchecked (Medium, MAI-Code)

The reviewer zeroed the field and nothing failed. I measured before fixing,
because the interesting question was whether the fixture *owes* a repair here:

```text
.text   ContainsCode, MemExecute, MemRead
.rsrc   ContainsInitializedData, MemRead
.reloc  ContainsInitializedData, MemDiscardable, MemRead
```

`ManagedPEBuilder` emits `.text` as code-only, so growing it genuinely does not
move `SizeOfInitializedData` and the fixture is correct to leave it alone. But
that premise was neither stated nor checked — were `.text` ever to gain
`ContainsInitializedData`, the fixture would silently owe a repair it does not
make. Both reviewers reached for this field independently, which is a fair sign
the assumption deserved pinning rather than remembering.

So the fix is two-part: the field is now covered by the optional-header gate,
and `HostileImage_GrowsASectionThatOnlyCountsAsCode` pins the premise that makes
leaving it alone correct. I also added `SizeOfImage`, per the suggestion.

### Changes

- `AssertSectionsAreAlignedAndDisjoint` also requires `SizeOfRawData` to be a
  whole number of file-alignment units.
- The `SizeOfCode` test becomes
  `HostileImage_KeepsOptionalHeaderSizesConsistentWithItsSectionTable`, covering
  `SizeOfInitializedData` and `SizeOfImage` alongside `SizeOfCode`.
- `HostileImage_GrowsASectionThatOnlyCountsAsCode` pins the code-only premise.
- The `Expansion` doc now says why multiple-of-alignment is load-bearing rather
  than merely tidy.

All three invariants hold exactly on compiler-produced assemblies, which is what
they are anchored against:

| image | `SizeOfCode` | `SizeOfInitializedData` | `SizeOfImage` | alignment |
| `DotnetInspector.MetadataRendering.Tests.dll` | 69120 | 2560 | 98304 | 512 |
| `System.Collections.dll` | 253952 | 28672 | 286720 | 4096 |

### Mutation matrix

| mutation | before | after |
| --- | --- | --- |
| drop raw size | 2 failed | 2 failed |
| drop `SizeOfCode` | 1 failed | 1 failed |
| drop raw size + `SizeOfCode` | 1 failed | 1 failed |
| drop later raw pointers | 1 failed | 1 failed |
| drop virtual size | 5 failed (SRM) | 6 failed (SRM) |
| off-by-one raw size + `SizeOfCode` | **0 failed** | 1 failed |
| zero `SizeOfInitializedData` | **0 failed** | 1 failed |
| corrupt `SizeOfImage` | **0 failed** | 1 failed |

The last is mine, not a reviewer's — having been beaten twice by "mutate two
things so their relationship survives," it seemed worth attacking the new gate
myself before shipping it.

### On the pattern

Five rounds. Every finding has been the same shape: a claim, or a check, that was
narrower than the thing it was supposed to guarantee. Rounds 1–3 were prose
outrunning evidence. Rounds 4 and 5 are the more interesting version — gates that
were real, non-vacuous, and killed by mutation, yet still satisfiable by a
coherent-looking edit.

What that says in general: a gate that survives *deleting* a repair does not
necessarily survive *changing* it consistently. Deletion tests are the cheap
first pass; the real question is whether an attacker with the same information
can keep every asserted relationship true while still corrupting the artifact.
That is a materially harder bar than "the mutation is killed," and it is the one
these gates are now written to.

The doc comment says explicitly that this set is not a general PE validator and
does not try to be — it covers the ways this fixture can plausibly rot.

### Validation at `59933eb4`

- `DotnetInspector.MetadataRendering.Tests` 141/141
- No product code changed since `a462a29a`; the metadata and CLI suites are
  unaffected and their results at that head stand (998/998, and 2502 with the two
  known Windows-only failures).

Both reviewers go to this head.

richlander:
## Review reconciliation, round 6 (head `59933eb4` -> `c59f347c`)

One finding from MAI-Code, accepted and fixed. Test-only change.

### Sections were gated on disk but not in memory (Medium, MAI-Code)

The round-5 gates covered file-space layout completely and memory-space layout
not at all. The reviewer raised the fixture's expansion to 8192 and deleted the
builder's RVA-overlap guard together, producing an image that satisfied every
assertion in the class:

```text
.text   VA=8192  VS=8720
.reloc  VA=16384 VS=12
Overlap? .text -> .reloc: True
Total: 141, Errors: 0, Failed: 0
```

Sections aligned and disjoint on disk, optional-header sizes matching the section
table, metadata inside its section — and two sections claiming the same memory.

The builder has always refused that expansion. The problem is that the guard was
the *only* thing holding the property up, so the property died with it. That is
the same shape as the raw-pointer repair in round 4: a correct piece of code
whose correctness nothing checked.

`HostileImage_KeepsSectionsAlignedAndDisjointInMemory` now asserts it directly —
section RVAs are multiples of `SectionAlignment`, virtual ranges do not overlap —
so the guard and the assertion fail independently of each other. Repeating the
reviewer's exact mutation now fails that one test:

```text
HostileImage_KeepsSectionsAlignedAndDisjointInMemory [FAIL]
Total: 142, Errors: 0, Failed: 1
```

I anchored this one on three compiler-produced assemblies rather than two, since
it is a layout rule and CoreLib is the most structurally different image to hand:

| image | alignment | `.text` | next | next |
| `DotnetInspector.MetadataRendering.Tests.dll` | 8192 | 8192 + 69800 | `.rsrc` 81920 | `.reloc` 90112 |
| `System.Collections.dll` | 4096 | 4096 + 251971 | `.data` 258048 | `.reloc` 282624 |
| `System.Private.CoreLib.dll` | 4096 | 4096 + 15336918 | `.data` 15343616 | `.reloc` 15958016 |

### The scope statement, made checkable

MAI-Code also suggested narrowing the doc comment so it stops implying a full PE
validator. Agreed, and worth doing properly rather than by hedging harder. The
old sentence said the set "covers the ways this fixture can plausibly rot" — which
sounds modest but is not a claim anyone can test, and this PR has now been burned
five times by claims nobody could test.

It now says something verifiable instead: every repair the builder makes is gated
by something — `MetadataReader` itself for the stream offsets and the virtual
size, a test here for the rest — and every field the builder deliberately leaves
alone has the premise that makes leaving it alone correct gated too. You check
that by reading the builder against the test list, which is the point.

### On the pattern

Six rounds. The finding has been the same shape every time and it keeps climbing
a level:

1. prose claiming a property the code did not have
2. prose claiming a gate that did not exist
3. a fixture claiming repairs that were not load-bearing
4. a gate satisfiable by mutating two things consistently
5. a gate satisfiable by an off-by-one that preserved every relationship
6. a property held up only by the single guard that implemented it

The through-line is that "I checked it" and "something would notice if it broke"
are different statements, and only the second one survives contact with a
reviewer. Each round the fix has been to move a fact from the first category to
the second.

### Validation at `c59f347c`

- `DotnetInspector.MetadataRendering.Tests` 142/142
- No product code changed since `a462a29a`; the metadata and CLI suites are
  unaffected and their results at that head stand (998/998, and 2502 with the two
  known Windows-only failures).

Both reviewers go to this head.

richlander:
## Review reconciliation, round 7 (head `c59f347c` -> `f4ebabb8`)

MAI-Code returned clean at `c59f347c`. GPT-5.6 Sol returned two findings, both
accepted and both fixed. Test-only change.

### The `SizeOfInitializedData` invariant was simply false (Medium)

I asserted `SizeOfInitializedData == sum of the initialized sections' raw sizes`
as a general PE property, anchored on three assemblies. The reviewer measured 35
of 990 repository PEs and 241 of 500 System32 PEs violating it. I scanned 2,432
shipped assemblies under the dotnet shared framework plus this repo's build
output and reproduced it — 69 counter-examples, all native:

```text
aspnetcorev2_inprocess.dll: SizeOfInitializedData=126464 rawSum=122368
```

Zero-filled virtual data counts toward the field without occupying file bytes.
The rule held for every image I happened to anchor against, which is exactly how
an invented law survives review.

Usefully, the same scan **clears every other layout invariant across all 2,432
images** — raw-pointer and raw-size alignment, raw-range disjointness, RVA
alignment, RVA disjointness, `SizeOfCode`, and `SizeOfImage`, zero violations
each. So this is confined to the one rule that was wrong, not a general retreat
from anchoring:

| invariant | violations / 2,432 |
| --- | --- |
| `PointerToRawData` and `SizeOfRawData` file-aligned | 0 |
| raw ranges disjoint | 0 |
| `VirtualAddress` section-aligned | 0 |
| virtual ranges disjoint | 0 |
| `SizeOfCode` == sum of code raw sizes | 0 |
| `SizeOfImage` == align(max(RVA + VirtualSize)) | 0 |
| **`SizeOfInitializedData` == sum of init raw sizes** | **69** |

Header sizes are now asserted against the fixture's own pre-patch baseline, which
the fixture exposes: `SizeOfCode` grows by exactly `Expansion`,
`SizeOfInitializedData` and `SizeOfImage` do not move. That asserts what the edit
did rather than what PE law is presumed to say. It is also *stronger* here — it
catches the drop-both and off-by-one mutations, which the formula did not — and
it is true everywhere.

### The fixture silently corrupts the PE entry point (Medium)

The round-6 scope claim said every unrepaired field had its premise gated. It did
not. Inserting bytes mid-`.text` moves everything after the insertion point, and
the fixture repairs only what `MetadataReader` traverses. Measuring the baseline:

```text
insertion point      = file 620
ImportAddressTable   rva=0x2000 file=512    before the cut, still valid
CorHeaderTable       rva=0x2008 file=520    before the cut, still valid
metadata root        file=592 size=232
ImportTable          rva=0x21B8 file=952    AFTER the cut, now stale
AddressOfEntryPoint  rva=0x220A file=1034   AFTER the cut, now stale
```

The reviewer's numbers reproduce exactly — the declared entry point now targets
version padding, and the real stub moved by precisely the 1,536-byte expansion.

They stay unrepaired, deliberately: this is a metadata fixture, nothing loads or
executes the image, and SRM reaches metadata through the CLI header at file 520,
which is before the cut and stays valid. But "we know what we broke" is only true
while the list is checked. `HostileImage_BreaksOnlyTheRvasItIsKnownToBreak` now
enumerates every data directory plus the entry point, classifies each against the
insertion RVA, and asserts the stale set is exactly
`{AddressOfEntryPoint, ImportTable}`. If `ManagedPEBuilder` ever emits another
`.text` directory past the metadata root, it fails and forces the choice to be
made again rather than inherited.

The class doc no longer claims PE-wide coherence. It states the image is
deliberately not loadable and points at the test that says how.

### Mutation matrix

Eleven mutations, all killed. Two are now caught by two gates each, where
previously one:

| mutation | failed | mutation | failed |
| --- | --- | --- | --- |
| drop raw size | 1 | drop CLI `MetaData.Size` | 2 |
| drop `SizeOfCode` | 1 | drop version-length | 2 |
| drop raw size + `SizeOfCode` | 2 | RVA overlap | 1 |
| drop later raw pointers | 1 | off-by-one raw size + `SizeOfCode` | 2 |
| drop virtual size | 7 | zero `SizeOfInitializedData` | 1 |
| corrupt `SizeOfImage` | 1 | | |

### On the pattern

Seven rounds. This one is the most instructive, because the defect was the
*remedy* from two rounds ago. In round 5 I was told to anchor invariants against
compiler-produced assemblies rather than invent rules — good advice, which I
followed, and which produced a false rule anyway because three assemblies is not
a sample. Anchoring is not the same as validating; it only tells you a rule is
consistent with the images you looked at.

The durable version is narrower: assert what your own code did to your own
input, and reach for a general law only when you are prepared to test it at
scale. Where I did need general laws — the layout invariants — 2,432 images is
the evidence, not three.

### Validation at `f4ebabb8`

- `DotnetInspector.MetadataRendering.Tests` 143/143
- No product code changed since `a462a29a`; metadata and CLI suite results at
  that head stand (998/998, and 2502 with the two known Windows-only failures).

Both reviewers go to this head.

richlander:
## Round 8 reconciliation — head `4d9e3778`

Both reviewers found the same defect from different directions, and it is the
one this test existed to prevent. `HostileImage_BreaksOnlyTheRvasItIsKnownToBreak`
was written in round 7 to end a pattern of claims nothing would notice breaking.
It was itself such a claim.

**@GPT-5.6 Sol — Medium, accepted.** The test read only the baseline, asking
whether each RVA sat past the insertion point. It never compared against the
patched image, so it could not distinguish a deliberately stale RVA from a
repaired one. Reproduced exactly: adding an `AddressOfEntryPoint` repair to the
builder left the class green.

**@MAI-Code — Medium, accepted.** The directory list was hand-written and short.
Reproduced exactly: planting a stale `DelayImportTable`, which the list did not
name, left 143/143 green.

**A third error neither reviewer had to find.** The list was wrong in the other
direction too. `BaseRelocationTable` (RVA `0x4000`) sits past the insertion RVA
and is entirely correct, because only `.reloc`'s raw pointer moves. Naming it
would have broken the old rule. Position in the baseline never determined
staleness — disagreement between the two images does.

### Fix

Per RVA, compute where the expansion *should* have left it — unchanged unless it
lies past the insertion point **in the same section** — and compare that to the
patched image. Directories are read from `NumberOfRvaAndSizes` rather than from a
list in the test, so an entry this fixture has never emitted is still classified.
An entry absent from the baseline must stay absent, which is MAI's case.
`CertificateTable` holds a file offset rather than an RVA and gets its own
assertion that it is empty, so it fails rather than being misfiled.

### Mutations

| mutation | result |
| --- | --- |
| repair `AddressOfEntryPoint` (GPT's case) | **1 failed** — previously survived |
| repair `ImportTable` | **1 failed** |
| plant stale `DelayImportTable` (MAI's case) | **1 failed** — previously survived |

Tests only; no product code changed since `a462a29a`. `DotnetInspector.MetadataRendering.Tests` **144/144**.

Both reviewers are being re-run at `4d9e3778`.

richlander:
## Round 9 reconciliation — head `16357a3a`

**@MAI-Code was clean at `4d9e3778`** — rebuilt, 144/144, and confirmed its own
`DelayImportTable` mutation now fails.

**@GPT-5.6 Sol — Medium, accepted.** The base relocation block encodes an RVA of
its own, pointing at the entry stub's operand. It stays `0x220C` after the stub
moves to `0x280A`, and the gate reported the image clean because it walked only
*header-declared* RVAs. My doc comment had said `.reloc` was "entirely correct" —
true of the directory, false of what it points at. Reproduced exactly.

### The class is wider than the instance

Rather than patch the one case, I walked the other RVA-bearing structures too.
The import directory and the CLI header hold **five more stale RVAs that no review
had reached**:

| stale RVA | is | should be |
| --- | --- | --- |
| `AddressOfEntryPoint` | `0x220A` | `0x280A` |
| `ImportTable` | `0x21B8` | `0x27B8` |
| `BaseRelocation[0]` | `0x220C` | `0x280C` |
| `Import[0].LookupTable` | `0x21E0` | `0x27E0` |
| `Import[0].Name` | `0x21FA` | `0x27FA` |
| `Import[0].Lookup[0].HintName` | `0x21EC` | `0x27EC` |
| `Import[0].Address[0].HintName` | `0x21EC` | `0x27EC` |
| `Cli.StrongNameSignature` | `0x2138` | `0x2738` |

Two, before this round. Eight now.

The import walk cannot follow the *patched* image's own pointers, because the
directory addressing those structures is itself one of the broken things —
following it lands in version padding, where the first "RVA" read is
`0x41414141`. Each structure is located in the baseline and its field re-read at
the offset the expansion moved that byte to. The content is intact and
unmodified; only what points at it is stale, which is the distinction this test
exists to record.

### Gating the walkers themselves

The relocation target hid because a structure went unwalked, so trusting the new
walkers to be complete would repeat the mistake one level up:

- `HostileImage_HasOnlyTheRvaBearingStructuresTheWalkersKnow` asserts which
  directories are present, so a fixture that gains a debug directory, TLS block,
  or delay-import table fails until a walker learns it.
- `HostileImage_IsPe32AsTheseWalkersAssume` pins the thunk stride and the
  directory-count offset, which differ in PE32+.

### Mutations

| mutation | tests failed |
| --- | --- |
| repair the relocation target (GPT's case) | 1 |
| repair the imported name RVA | 1 |
| repair the IAT hint/name RVA | 1 |
| repair the strong-name RVA | 1 |
| repair `AddressOfEntryPoint` | 1 |
| plant a stale `DelayImportTable` | 2 |

Tests only; no product code changed since `a462a29a`. `DotnetInspector.MetadataRendering.Tests` **146/146**.

Both reviewers are being re-run at `16357a3a`.

richlander:
## Round 10 reconciliation — head `e1e53d8e`

**@GPT-5.6 Sol — Medium, accepted.** CLI header offset 20 holds a managed
entry-point token normally, but an **RVA** when the header sets
`NativeEntryPoint`. `CliHeaderTargets` omitted the field on that unstated
premise, and neither meta-gate pinned the flag. Reproduced exactly: setting the
flag and planting a stale `0x220A` there left 146/146 green.

The field is now read as an RVA exactly when the flag says it is one, so the
premise is enforced by construction rather than assumed. Reproducing GPT's case
now fails with the value it predicted:

`Cli.NativeEntryPoint 0x220A should be 0x280A`

**Pre-empting the same class one level down.** Metadata stores RVAs too — in
`MethodDef` bodies and the `FieldRva` table. This fixture has neither, which is
why the walkers stop at the CLI header. That is a property of the fixture rather
than a law, so it is now asserted alongside the directory set instead of being
left as the next thing a reviewer finds.

**@MAI-Code** reported clean at `16357a3a` but quoted `144/144` — the total from
the *previous* head. I verified its worktree was correctly at `16357a3a` with
both new tests present, so the evidence did not match the code. I rejected the
report and asked for a fresh run; that is still outstanding, and this PR will not
be marked ready on it.

### Mutations

| mutation | tests failed |
| --- | --- |
| set `NativeEntryPoint` + plant a stale RVA (GPT's case) | 1 |
| repair the relocation target | 1 |
| repair the imported name RVA | 1 |
| repair the IAT hint/name RVA | 1 |
| repair the strong-name RVA | 1 |
| repair `AddressOfEntryPoint` | 1 |
| plant a stale `DelayImportTable` | 2 |

Tests only; no product code changed since `a462a29a`. `DotnetInspector.MetadataRendering.Tests` **146/146**.

richlander:
## Round 11 reconciliation — head `70f9c4f0`

**@MAI-Code — Medium, accepted as a documentation defect.** Its fresh run at
`e1e53d8e` reported `Total: 146`, so the earlier stale-count report is resolved.

The finding: my prose claimed more than the walk does. It said "every RVA the
image declares — optional header **and section content alike**", which reads as
coverage of all bytes in a section. MAI wrote a plausible RVA into section
padding and nothing failed.

The observation is correct. The remedy is the wording, not a scan:

- No structure points at those bytes, so **nothing in the image says they are an
  address rather than data**. A four-byte value in padding is not a stale RVA; it
  is a four-byte value.
- Searching for values that merely *resemble* RVAs would report padding, string
  bytes, and IL as findings — a guess dressed as a gate, and exactly the
  harness-side shape recognition `AGENTS.md` rules out.

So the scope is now stated as what it is: **every RVA reachable from a structure
the image declares** — the optional header, the relocation blocks, the import
directory and its thunks, and the CLI header. The claim is held up from the other
end by `HostileImage_HasOnlyTheRvaBearingStructuresTheWalkersKnow`, which fails
if the fixture gains a structure the walkers do not follow — the case that would
actually make unwalked bytes meaningful.

This is the same defect as rounds 1 and 2, arriving last: not a missing check, a
sentence that promised more than the check delivers.

Docs only within the test file; no behavior change, no product code changed since
`a462a29a`. `DotnetInspector.MetadataRendering.Tests` **146/146**.

Both reviewers are being re-run at `70f9c4f0`.

richlander:
## Round 12 reconciliation — head `8d80980d`

**@MAI-Code was clean at `70f9c4f0`** — fresh run, `Total: 146`, no findings. It
also endorsed the scope statement and the decision not to overreach into
unreferenced bytes.

**@GPT-5.6 Sol — Medium, accepted.** The scope statement claims every RVA
reachable from the optional header, and `RvaFields` omitted `BaseOfCode` and
PE32's `BaseOfData`. Planting a stale `BaseOfData` left 146/146 green.

GPT drew the right line: these are **typed address fields, not incidental bytes**,
so last round's argument for rejecting a heuristic scan of unreferenced padding
does not cover them. Both are walked now, and so are the section table's virtual
addresses — the other place an image declares RVAs in quantity.

Neither turns out to be stale: `BaseOfCode` is `0x2000` and `BaseOfData`
`0x4000`, both outside the grown region ahead of the cut, and section RVAs do not
move because the expansion moves raw pointers. The inventory stays at eight.
Reading `BaseOfData` is safe only in PE32, where the slot exists at all rather
than being the low half of `ImageBase`;
`HostileImage_IsPe32AsTheseWalkersAssume` is what makes that true rather than
assumed.

### Mutations

| mutation | tests failed | message |
| --- | --- | --- |
| plant a stale `BaseOfData` (GPT's case) | 1 | `BaseOfData 0x220A should be 0x4000` |
| plant a stale `BaseOfCode` | 1 | `BaseOfCode 0x220A should be 0x2000` |
| repair the relocation target | 1 | |
| repair the imported name RVA | 1 | |
| repair the IAT hint/name RVA | 1 | |
| repair the strong-name RVA | 1 | |
| repair `AddressOfEntryPoint` | 1 | |
| plant a stale `DelayImportTable` | 2 | |

Tests only; no product code changed since `a462a29a`. `DotnetInspector.MetadataRendering.Tests` **146/146**.

Both reviewers are being re-run at `8d80980d`.

richlander:
## Round 13 reconciliation — head `b7e7ab77`

**@MAI-Code — Medium, accepted.** Fresh run at `8d80980d` reported `Total: 146`.

The finding: round 12 added `SectionRvas` and `OptionalHeaderRvas` to the walk
but left the scope prose listing only the optional header, relocation blocks,
import directory, and CLI header. The documented scope no longer matched the
implemented one.

This is the same defect as rounds 1, 2 and 11 — a sentence out of step with its
check — and this time in the direction of **understating** coverage rather than
overstating it. Both directions are wrong for the same reason: the sentence is
what a reader checks against, so it is the thing that has to be true.

The scope now reads: the optional header's scalar address fields, the section
table, the data directories, the relocation blocks, the import directory and its
thunks, and the CLI header.

I had flagged this exact question to both reviewers when posting round 12 —
"the section table is now walked but is *not* named in that sentence; is the
sentence wrong, or merely incomplete?" — and MAI confirmed it rather than letting
it pass.

Comments only; no behavior change, no product code changed since `a462a29a`.
`DotnetInspector.MetadataRendering.Tests` **146/146**.

Both reviewers are being re-run at `b7e7ab77`.

richlander:
## Round 14 — GPT-5.6 Sol, one Medium, accepted and fixed in `6060caef`

**Finding.** The CLI header's `VTableFixups` entry points at an array of `COR_VTABLEFIXUP` records whose first field is an RVA. The walk recorded the top-level pointer but never descended, and the meta-gate did not require the structure to be absent. GPT built a reachable `VTableFixups`, repaired the CLI pointer so it classified as *correct*, and left the nested RVA stale at `0x220A`. All 146 tests passed.

**Why the existing gate missed it.** `HostileImage_HasOnlyTheRvaBearingStructuresTheWalkersKnow` pins the set of present *data directories*, on the argument that a structure the walk cannot descend into must not be present in the fixture. That argument does not stop at the directory level — the CLI header is itself a table of pointers, and two of its entries (`VTableFixups`, `ManagedNativeHeader`) point at structures containing addresses. Exhaustiveness was asserted one level up from where it was needed.

**Fix.** The present CLI sub-directories are pinned the same way as the directories: `Cli.MetaData` and `Cli.StrongNameSignature`, and nothing else.

I considered walking `COR_VTABLEFIXUP` records instead and rejected it. A `ManagedPEBuilder` image can never contain one, so the walker would be dead code — and an untested walker is a weaker guarantee than an assertion that the structure is absent. The pin deliberately also fails on harmless additions such as managed resources, whose blobs hold no addresses, because the property being defended is that *someone classifies the new structure*, not that the gate guesses correctly on their behalf.

**Mutation evidence, both directions:**

| mutation | tests failed |
| --- | --- |
| plant a **stale** `VTableFixups` RVA | 2 (inventory + meta-gate) |
| plant a **correctly-classified** `VTableFixups` RVA — review's exact mutation | **1 (meta-gate alone)** |

The second row is the one that matters: it proves the new gate is load-bearing on its own, and not riding on the staleness inventory.

**Validation at `6060caef`:**

- `DotnetInspector.MetadataRendering.Tests` — `Total: 146, Errors: 0, Failed: 0`
- `dotnet-inspect.Tests` — `Total: 2502, Errors: 0, Failed: 2`, both pre-existing Windows-only (`Layout_Count_CountsFilesRatherThanRenderedTreeLines`, `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown`), confirmed identical at an unmodified head.

Comments and tests only. **No product code has changed since `a462a29a`**, twelve rounds ago.

**Status.** MAI-Code was clean at `b7e7ab77`, but the head has moved, so both reviewers re-run at `6060caef`. Not marking ready until both are clean at the same head.

richlander:
## Round 15 — GPT-5.6 Sol, one Medium, accepted and fixed in `880b830e`

**Finding.** `ImportTargets` is driven entirely by the baseline. GPT replaced the moved null import descriptor in the *patched* image with one carrying three typed RVAs; all 146 tests passed and none of the new RVAs entered the stale inventory.

**Why.** Since round 9 the import walk cannot follow the patched image's own import directory — that directory is one of the things the expansion breaks, and following it lands in version padding where the first "RVA" read is `0x41414141`. So every structure is located in the baseline and each field re-read at the offset the expansion moved it to.

That is still correct and still necessary. What it quietly does, though, is make the **baseline the sole authority on shape**: which descriptor is the null terminator, where each thunk list ends, and which entries carry the ordinal flag. A structure present only in the patched image is not walked — it is not even reached. Nothing stated that premise and nothing checked it.

**Fix.** The premise is now asserted rather than assumed. The same traversal runs twice — once reading baseline bytes, once reading the patched bytes each site moved to — and the two shapes must agree. Only shape is compared, never an RVA value: the RVAs are *expected* to differ between the images, and classifying that difference is what `HostileImage_BreaksOnlyTheRvasItIsKnownToBreak` exists for.

**Mutation evidence, on all three dimensions the gate claims to cover:**

| mutation | tests failed |
| --- | --- |
| patched-only import descriptor — GPT's exact mutation | 1 (the new gate alone) |
| patched-only thunk | 1 (the new gate alone) |
| ordinal flag flipped in the patched image | 2 |

**Validation at `880b830e`:** `DotnetInspector.MetadataRendering.Tests` — `Total: 147, Errors: 0, Failed: 0`.

Comments and tests only. **No product code has changed since `a462a29a`**, thirteen rounds ago.

---

### Where this leaves the two-model tier

MAI-Code was clean at `6060caef`. GPT found this at the same head, so the head has moved again and both reviewers re-run at `880b830e`.

Worth naming the shape of the last few rounds, because it is converging rather than wandering. Rounds 13, 14 and 15 are the same defect at successively deeper levels of the same argument:

- **13** — the scope prose did not match the walk.
- **14** — the exhaustiveness gate was asserted one level up from where it was needed: it pinned the present *data directories*, but the CLI header is itself a table of pointers.
- **15** — the walk's own *siting strategy* carried an unstated premise: baseline-driven traversal is only sound if both images have the same shape.

Each round has moved from "what does the walk cover" to "what does the walk assume". Round 15 is the first that is about the walk's method rather than its extent, which is the last layer I can identify. Not marking ready until both reviewers are clean at the same head.

richlander:
## Round 16 — both reviewers, one seam, opposite sides. Fixed in `543d171a`

**GPT-5.6 Sol (Medium, accepted).** A data directory entry is an address *and* a size. The RVA inventory classifies the address half — which is all it can do, since a size is not an address — so **every declared extent in the image was unchecked**. GPT shrank the patched import directory's size to 20 bytes, putting its own null terminator outside the declared extent without altering a single byte the walk reads. All 147 tests passed.

That is not just bookkeeping. `ImportShape` bounds *both* of its traversals with the baseline's `ImportTableDirectory.Size`, so growing the patched extent and planting a descriptor in the newly covered region would have been **declared by the patched image and still invisible to the walk**.

**MAI-Code (Medium, half accepted).** MAI planted a descriptor-shaped run of bytes past the baseline import window and reported that the shape gate did not catch it. Reproduced and split in two:

| variant | verdict |
| --- | --- |
| grow the patched extent so the planted bytes become **declared**, then plant | **Real.** Fails the new gate. |
| plant past the extent, leave the extent unchanged | **Not a finding.** |

The second is round 11's line. Measured, the import region is **79 bytes** spanning the descriptors, the lookup table *and* the name strings; the descriptor array ends at its null terminator at index 1. Bytes past both are declared by nothing, and reporting them would mean recognising structures by their shape rather than by reachability — which is the harness-side shape recognition `AGENTS.md` forbids. The distinction is now written into the test's doc comment rather than left as a judgement call.

**Fix.** Every directory's declared extent must survive the expansion unchanged. The expansion moves bytes; it does not change how much of the image anything claims. Measured rather than assumed — all sixteen extents are identical across the patch.

| mutation | tests failed |
| --- | --- |
| patched import extent shrunk to 20 — GPT's exact mutation | 1 (new gate alone) |
| patched `CorHeaderTable` extent grown to 999 — proves it is not import-specific | 1 |
| extent grown to 119 **and** a descriptor planted inside — MAI's real variant | 1 |
| descriptor planted past the extent, extent unchanged — MAI's other variant | 0, by design |

**Validation at `543d171a`:** `DotnetInspector.MetadataRendering.Tests` — `Total: 148, Errors: 0, Failed: 0`.

I also independently checked the question I put to both reviewers about `Pair`: it already asserts length equality *and* per-index name equality, so every walker that runs independently per image (`RelocationTargets`, `CliHeaderTargets`, `DataDirectories`, `OptionalHeaderRvas`, `SectionRvas`) fails loudly on a shape difference. `ImportTargets` was the only walker taking *both* images and emitting pairs directly — which is precisely why it was the one with the hole.

Comments and tests only. **No product code has changed since `a462a29a`**, fourteen rounds ago.

Both reviewers re-run at `543d171a`.

richlander:
## Round 17 — both reviewers, same class again. Fixed in `6142aade` by changing the shape of the assertion

**The findings.** MAI-Code: `SizeOfHeaders` is unchecked — corrupt it in the patched image and no test fails. GPT-5.6 Sol: the CLI header's extents are unchecked — overstating `Cli.MetaData.Size` by 128 bytes makes the declared metadata extent cover bytes that are not metadata, and all 148 tests passed.

Both accepted. But the more useful observation is that this is the **fourth consecutive round** to find one more unchecked field, and that is the actual defect.

Every gate in this file named what it checked. A field nobody had named was therefore checked by nobody. Rounds 12, 14, 15 and 16 each added one more name — `BaseOfCode`/`BaseOfData`, the CLI sub-directories, the import shape, the directory extents — and each time the next round found the next omission. I probed ahead of review this round hoping to break the pattern and instead confirmed it: I found `SizeOfHeaders` unchecked before MAI's report arrived, which tells you the supply of unnamed fields was not close to exhausted.

**So this commit stops naming fields and asserts the complement.** The expansion touches two regions; in each, every byte *outside* the documented repairs must be byte-identical:

| region | may change |
| --- | --- |
| PE headers | `SizeOfCode`, `.text` `VirtualSize`, `.text` `SizeOfRawData`, `.reloc` `PointerToRawData` |
| CLI header | `MetaData.Size` |

A complement cannot go stale. A field added to either region tomorrow is covered the moment it exists, because it was never the naming that made it covered. This subsumes `SizeOfHeaders`, `NumberOfRvaAndSizes`, the CLI header's `cb`, the directory extents in the optional header, and every field none of us has thought of.

The CLI header needs its own test because it does **not** live in the PE header region — it sits inside `.text`, past the first section's raw data — so the header-region comparison does not reach it. That is asserted, not assumed: the test first pins that the CLI header precedes the insertion point, since otherwise the byte comparison would be comparing unrelated bytes.

Both gates also assert **non-vacuity**: each documented repair must actually have happened. Without that, an expansion that quietly stopped repairing anything would pass an empty complement.

| mutation | tests failed |
| --- | --- |
| `SizeOfHeaders` corrupted — MAI's exact mutation | 1 |
| `SizeOfCode` repair removed from the expansion — non-vacuity | 2 |
| `Cli.MetaData.Size` overstated by 128 — GPT's exact mutation | 1 |
| `Cli.Resources.Size` changed — proves it is not metadata-specific | 1 |

**Validation at `6142aade`:** `DotnetInspector.MetadataRendering.Tests` — `Total: 150, Errors: 0, Failed: 0`.

Comments and tests only. **No product code has changed since `a462a29a`**, fifteen rounds ago.

Both reviewers re-run at `6142aade`.

richlander:
## Round 18 — the complement, asserted over the whole image

**GPT-5.6 Sol** found at `6142aade` that the metadata stream headers were covered by neither complement: growing a stream's declared `Size` by four bytes while repairing its `Offset` normally left all 150 tests green. **MAI-Code** was clean at that head.

I had found the same gap independently, from the other side — flipping a byte inside the patched metadata heap also failed nothing.

### The finding is accepted, and so is what it says about round 17

Round 17 diagnosed the previous four rounds correctly: every gate named what it checked, so a field nobody had named was checked by nobody. The remedy was to assert the complement instead.

But it wrote **two** complements — the PE headers and the CLI header — and splitting the image into named regions to check reintroduces the exact problem the complement was meant to end. A region nobody named is checked by nobody. The metadata was that region. So were `.reloc`, the IL, the heaps and the tables.

### `e983b405`

There is one region now: the whole image. `HostileImage_DiffersFromItsBaselineOnlyWhereDocumented` replaces both round-17 gates, which are deleted rather than kept alongside — a weaker restatement of the same claim is a liability, not defence in depth.

The assertion is the fixture's own description of itself. The patched image is a byte-for-byte copy of the baseline with a widened version field and a handful of position repairs:

```
patched[0 .. versionStart)                   == baseline[same offsets]      except documented in-place repairs
patched[versionStart .. +newVersionLength)   == the payload
patched[insertionPoint+Expansion ..)         == baseline[insertionPoint ..]  except stream Offset fields
```

The documented in-place repairs are `SizeOfCode`, `.text VirtualSize`, `.text SizeOfRawData`, `.reloc PointerToRawData`, `Cli.MetaData.Size` and `MetadataRoot.VersionLength`. That they all precede the insertion point is asserted, not assumed. The stream-header `Offset` fields follow it and are **enumerated by walking the storage header** rather than listed, so a fixture that gains a stream is covered without anyone editing this test.

Non-vacuity is stronger than round 17's: each repair must not merely have changed, it must equal `baseline + Expansion`. A repair that fires with the wrong value now fails here rather than only in whichever gate happens to own that field.

### Mutation matrix, all against the single gate

| mutation | tests failed |
| --- | --- |
| first stream's declared size grown by 4 — GPT's round-18 mutation | 1 |
| metadata heap byte flipped (offset 2300) — found independently | 1 |
| stream offset repair skewed by 4 | 4 |
| `SizeOfHeaders` corrupted — MAI's round-17 mutation | 1 |
| `Cli.MetaData.Size` overstated by 128 — GPT's round-17 mutation | 1 |
| `Cli.Resources.Size` changed | 1 |
| import extent shrunk to 20 — GPT's round-16 mutation | 2 |
| `VTableFixups` planted — GPT's round-14 mutation | 3 |
| patched-only import descriptor — GPT's round-15 mutation | 2 |

`DotnetInspector.MetadataRendering.Tests Total: 149, Failed: 0.`

### Scope

Comments and tests only. **No product code has changed since `a462a29a`**, eighteen rounds ago — the version-stamp containment fix and its renderer change are exactly as first reviewed. Everything since has been the argument that the fix is gated.

Both reviewers are re-running at `e983b405`.

richlander:
## Round 19 — the excluded middle

**GPT-5.6 Sol** replaced the version payload's terminating NUL with `A` at `e983b405`. All 149 tests passed.

### The finding

The whole-image complement compared the bytes *before* the widened version field and the bytes *after* it, and exempted the payload itself — on the reasoning that the payload is the thing being *inserted* rather than something *preserved*, so there is no baseline counterpart to compare it against.

That reasoning is correct and the conclusion drawn from it was wrong. Nothing else in the image describes the payload's content, so exempting it in the one test that decomposes the image exempted it everywhere. **A complement is only as good as the union of what it excludes being asserted somewhere**, and the excluded middle was asserted nowhere.

This is the same defect as rounds 12 through 18, one level further in. Round 18 asserted the complement over the whole image; it did not notice that "the whole image" had a hole in the middle of it by construction.

### `9d5aafcb`

All three ranges are checked, and the **tiling is itself asserted** rather than left as arithmetic a reader has to verify:

```csharp
Assert.Equal(baseline.Length + expansion, patched.Length);
int payloadEnd = versionStart + newVersionLength;
Assert.Equal(payloadEnd, insertionPoint + expansion);   // no gap, no overlap
```

then the payload itself:

```csharp
byte[] payload = [(byte)'v', .. Enumerable.Repeat((byte)'A', newVersionLength - 2), (byte)0];
Assert.Equal(payload, patched[versionStart..payloadEnd]);
```

### Payload mutations

| mutation | tests failed |
| --- | --- |
| terminating NUL replaced with `A` — GPT's mutation | 1 |
| leading `v` replaced with `B` | 1 |
| fill one byte short | 1 |

The round-18 matrix holds unchanged: stream size grown 1; metadata heap byte flipped 1; stream offset skewed 4; `SizeOfHeaders` 1; `Cli.MetaData.Size` 1; `Cli.Resources.Size` 1; import extent shrunk 2; `VTableFixups` planted 3; patched-only descriptor 2.

`DotnetInspector.MetadataRendering.Tests Total: 149, Failed: 0.`

### Scope

Comments and tests only. **No product code has changed since `a462a29a`**, nineteen rounds ago.

GPT-5.6 Sol is re-running at `9d5aafcb`. MAI-Code was clean at `6142aade` and is mid-run at `e983b405`; it will re-run at `9d5aafcb` before this is marked ready.

richlander:
## Round 20 — a gate that could not fail

GPT-5.6 Sol reported **clean** at `9d5aafcb`, and MAI-Code reported **clean** at `e983b405`. This round exists because I checked one of the clean report's specific claims and it did not hold.

### The finding

I had asked GPT whether the payload expectation was anchored on the artifact or merely restated the fixture's arithmetic. It answered clean, stating that "the declared length matches the payload extent". That claim was false at `9d5aafcb`, and the assertion that appeared to establish it was the problem:

```csharp
int newVersionLength = checked(oldVersionLength + expansion);   // test's own arithmetic
int payloadEnd = versionStart + newVersionLength;
Assert.Equal(insertionPoint + expansion, payloadEnd);
```

`insertionPoint` is `versionStart + oldVersionLength`, so both sides expand to `versionStart + oldVersionLength + expansion`. **The two sides were the same expression; the assertion could not fail.**

That is worse than a missing check. It read exactly like the check that the metadata root's *declared* version length agrees with the payload *actually written* — and no such check existed anywhere. A gate that cannot fail is the precise failure mode this file exists to prevent, so it could not stay in the file whose entire argument is that gates must be able to fail.

### `4b2ba425`

The payload's extent is read from the patched image's own declared version length, so the tiling assertion compares a value *from the artifact* against the geometry the rest of the test assumes:

```csharp
int newVersionLength = ReadRva(patched, metadataRoot + 12);
int payloadEnd = versionStart + newVersionLength;
Assert.Equal(insertionPoint + expansion, payloadEnd);
```

The right-hand side stays anchored on the **baseline** artifact plus the compile-time expansion, so a corrupt declared length cannot shrink the expectation to match itself.

| mutation | at `4b2ba425` | at `9d5aafcb` |
| --- | --- | --- |
| declared version length 4 shorter than the payload written | **4 failed** | 0 — both sides moved together |
| declared **and written** 4 shorter (the consistent mutation) | **4 failed** | — |

The second row is the round-5 lesson applied to this change: mutating two things consistently has defeated a gate in this file before, so the anchor was checked against it explicitly rather than assumed.

Round-19 payload matrix holds: NUL replaced with `A` 1; leading `v` replaced with `B` 1; fill one byte short 1.
Round-18 matrix holds: stream size grown 1; heap byte flipped 1; stream offset skewed 4; `SizeOfHeaders` 1; `Cli.MetaData.Size` 1; `Cli.Resources.Size` 1; import extent shrunk 2; `VTableFixups` planted 3; patched-only descriptor 2.

`DotnetInspector.MetadataRendering.Tests Total: 149, Failed: 0.`

### Scope

Comments and tests only. **No product code has changed since `a462a29a`**, twenty rounds ago.

Both reviewers are re-running at `4b2ba425`, and both have been asked to verify the vacuity claim themselves rather than take the table on trust.

richlander:
## Both reviews clean at `4b2ba425`

| reviewer | head | verdict |
| --- | --- | --- |
| GPT-5.6 Sol | `4b2ba425` | no findings — independently reproduced both vacuity mutations |
| MAI-Code | `4b2ba425` | no findings — independently reproduced the vacuity mutation, and audited the test's other assertions for tautology |

Both were asked to verify the non-vacuity table themselves rather than accept it, and both did.

### What this PR actually changes

**One product defect, fixed in `a462a29a` and untouched for twenty rounds since.** `MetadataImageInspector.Describe` reported `reader.MetadataVersion` verbatim and `MetadataProjectionRenderer` emitted it raw. The metadata root's version stamp is attacker-controlled and reached presentation *without* passing through the row projection, so it inherited none of the projection's containment. It now goes through the same `NeutralizeControls` every projected value does, with a 255x6 budget and an explicit ellipsis when truncated.

That is the whole product change. Everything else is the argument that it is gated.

### Why issue #3444's premise was mostly wrong, and what was actually missing

#3444 said `mdi` renders untrusted metadata without containment. It doesn't — it inherits containment from the shared `MetadataTableProjector`, and adds no escaping of its own. That is exactly what makes it a good reference consumer, and exactly why the version stamp leaked silently: **containment is inherited only along paths that go through the projection, and an unprojected value looks identical at the call site to a projected one.**

So the real gap was not containment but a *gate* on it, per this repo's rule that an asserted safety property must name the gate that enforces it.

### Twenty rounds, one defect

Every round after the first found the same thing at a new depth: **a claim broader than its check.** The arc is worth recording because the fix kept being a smaller version of the disease.

| | finding |
| --- | --- |
| 1 | the version stamp leaked (the only product defect) |
| 2–6 | fixtures claimed load-bearing that weren't; gates defeated by mutating two things consistently |
| 7 | `SizeOfInitializedData == sum(init raw)` is false in general — 69 counter-examples in 2,432 assemblies |
| 8 | the gate read only the baseline, never comparing against the patched image |
| 9–10 | `.reloc` content, then the CLI header, held stale RVAs nobody walked |
| 11 | a "finding" in bytes nothing points at — rejected as shape recognition |
| 12 | `BaseOfCode`/`BaseOfData` unwalked for four rounds |
| 13 | the prose *understated* what the walk covered |
| 14 | the exhaustiveness gate was asserted one level up from where it was needed |
| 15 | the walk's siting strategy made the baseline the sole authority on shape |
| 16 | a directory entry is an address **and a size**; only the address half was classified |
| 17 | **naming fields is the bug** — a field nobody named was checked by nobody |
| 18 | **naming regions is the same bug** — the metadata was in neither complement |
| 19 | the complement had a hole in the middle: the payload was excluded from everything |
| 20 | the tiling assertion was a tautology — **it could not fail** |

Rounds 17–20 are one idea converging: *assert the complement over the whole artifact, and make sure the complement can fail.* A complement cannot go stale — a field added tomorrow is covered the moment it exists — but only if nothing is quietly exempted from it and its bounds are anchored outside itself.

Round 20 was not a reviewer finding. It came from checking a specific claim inside a **clean** report instead of accepting it.

### Final validation at `4b2ba425`

| suite | result |
| --- | --- |
| `DotnetInspector.MetadataRendering.Tests` | Total: 149, Failed: 0 |
| `ILInspector.Metadata.Tests` | Total: 998, Failed: 0 |
| `dotnet-inspect.Tests` | Total: 2502, Failed: 2 |

The two failures are pre-existing and Windows-only, unrelated to this change: `Layout_Count_CountsFilesRatherThanRenderedTreeLines` and `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown(readmePath: "images/logo.png.")`. They fail identically on `origin/main`.

CI green: `build-net10`, `test (ubuntu-26.04)`, `decompiler-gates`, `markdownlint`, `changes`.

### Compatibility

No public API change. `MetadataImageOverview` gains a trailing defaulted constructor parameter. `NeutralizeControls` widens from private to internal. Output changes only for images whose metadata version stamp contains control characters or exceeds the budget — which no compiler-produced assembly does.

Ready to merge

PR #3520, opened 2026-07-29
Gate the decompiler docket and byte-neutrality checks pre-merge

### LINKED ISSUE #3432: Fidelity and byte-neutrality gates claim to fail CI but are excluded from every pre-merge lane
## Problem

`FidelityGateTests`' own doc comment asserts a property that its trait, the CI
workflow, and the branch ruleset all decline to enforce:

```csharp
/// This pins the green set so a regression that turns
/// an exact method into a diff fails CI, ...
[Trait("Speed", "Slow")]
public class FidelityGateTests
```
`src/ILInspector.Decompiler.Tests/FidelityGateTests.cs:10-14`

It does not fail **PR** CI. Three independent layers each stop it:

1. **The trait is excluded on PRs.** `ci.yml` runs on `pull_request`
   (`.github/workflows/ci.yml:6`) and the decompiler lane is
   `-trait- "Speed=Slow"` (`ci.yml:301`) — the exact trait the class carries.
2. **The PR corpus sensor is compile-disabled and non-blocking.** It runs
   `--compile-cap 0 --corpus-fidelity-cap 0` under `continue-on-error: true`
   (`ci.yml:225-243`), so it neither performs compile-back nor can fail the job.
3. **No status check is required to merge.** `main`'s ruleset enforces
   `deletion`, `non_fast_forward`, and `pull_request`, and contains **no
   `required_status_checks` rule** (`gh api repos/richlander/dotnet-inspect/rules/branches/main`).

The only lane that runs fidelity is Deep Inspect, on `schedule: cron: '0 9 * * 1'`
(`.github/workflows/deep-inspect.yml:17-18`) — **weekly**, after merge.

This is not confined to one class. **8 of the 22 `Speed=Slow` classes in
`ILInspector.Decompiler.Tests` are gates or fidelity checks**, all excluded from PR CI:

```
FidelityGateTests           TypeBindGateTests
CorpusSweepGateTests        LadderIteratorGateTests
ByteNeutralityGateTests     DiffFixtureFidelityTests
LoweredFidelityGateTests    AuthoredRebuildFidelityTests
```

`ByteNeutralityGateTests` is the class `AGENTS.md` holds up as the exemplar of
declaration-driven enforcement ("derives its coverage set from the style catalog").
It does not run before merge either.

## Why this is worth a separate issue

`AGENTS.md` requires that "a doc comment or README that asserts a safety, soundness, or
faithfulness property must name the gate that enforces it, or explicitly mark the
property as unverified." `FidelityGateTests` names its gate as "CI" and CI does not
enforce it pre-merge.

The repository already has the correct pattern **one file over**.
`CorpusSweepGateTests.CoreLibSweep_MeetsHealthFloors` says so plainly:

```csharp
/// Breadth health floors — runtime-version-sensitive and drifty, so
/// <c>Speed=Slow</c> and run in the corpus gate rather than per-PR.
```
`src/ILInspector.Decompiler.Tests/CorpusSweepGateTests.cs:169-171`

That comment is accurate and self-aware. `FidelityGateTests`' is not. A reader comparing
the two would reasonably conclude fidelity is merge-blocking and breadth is not.

## Consequence

Compile-back fidelity — "valid but different IL," the failure class the gate exists to
catch and the one no parse/bind check can see — has a **detection latency of up to seven
days** and no merge-blocking authority. Any structural work (loop raising, per-region
containment, CFG model consolidation in #3242) inherits that safety net.

## Relationship to existing issues

- **#2252** item 2 is the origin: it proposed trait-splitting corpus-scale tests and
  running them "once **pre-merge or** in the daily." The trait split landed; the daily
  landed; the pre-merge half did not. This issue is the unlanded half, not a duplicate.
- **#3242 / #3240** depend on this: both propose changes to control-flow models whose
  regressions are exactly what the fidelity gate detects.

## Proposed work

Either of these closes the gap; they are not mutually exclusive.

1. **Make the comment true.** Wire a bounded fidelity slice into the PR lane — scoped to
   the methods `--emit-corpus-delta` reports as changed. The flag exists
   (`tools/DecompilerHarness/Program.cs`) and is documented
   (`docs/decompiler-corpus-delta-repro.md`) but has **no CI wiring** in any workflow.
   Raise `--corpus-fidelity-cap` above 0 for that slice and drop `continue-on-error`
   for it.
2. **Make the comment accurate.** If fidelity is deliberately post-merge, restate the
   doc comment in `CorpusSweepGateTests`' honest form — naming the weekly lane as the
   gate — so no reader infers pre-merge enforcement.

Separately, consider whether `main` should require the `test` status check. Today a red
CI run does not mechanically block merge.

## Non-goal

This is not a proposal to run the full slow suite on PRs. #2252 measured why that is the
wrong trade. The claim here is narrower: **a gate that says it fails CI should either
fail CI, or say which lane it actually gates.**

## Method

- `.github/workflows/ci.yml` lines cited above, read at `main`.
- `gh api repos/richlander/dotnet-inspect/rules/branches/main --jq '[.[].type]'`
  → `["deletion","non_fast_forward","pull_request"]`.
- Trait census: `grep -rln 'Trait("Speed", "Slow")' src/ILInspector.Decompiler.Tests/`
  → 22 files, 24 declarations, 8 matching `gate|fidelity`.

richlander:
Closes #3432.

## What this does

Runs the three decompiler docket / byte-neutrality gates on PRs, in a path-gated job, and decides the job's result by comparing the run against a pinned list of known-red tests.

**The gate goes on red, deliberately.** The three currently-open failures are pinned, not fixed. Waiting for green is what let the backlog accumulate in the first place; a gate's job is to make the *next* breakage attributable to the PR that caused it.

## Why the existing gates never fired

The gates carry doc comments asserting they fail CI. They do not, and have not.

- They are all `Speed=Slow`, and the PR `test` job runs `-trait- "Speed=Slow"` (`ci.yml:326`). No pre-merge lane ran them.
- That left `release.yml` and the weekly Deep Inspect lane. In Deep Inspect they were **exceeding the job timeout**, and a *cancelled* job does not satisfy that workflow's `if: ... && failure()` notifier — so no notification was ever sent.

Detection latency was therefore unbounded, not weekly (#3432). Five real regressions accumulated unseen: #3489, #3490, #3491, #3492, #3493.

## Design

| Piece | Role |
| --- | --- |
| `--gate pre-merge` | new preset; the three gate classes |
| `decompiler-gates` job | path-gated, own job, never serializes with the hot `test` lane |
| `eng/decompiler-gate-known-red.txt` | pinned failures, each with issue + pin date + root cause |
| `eng/check-decompiler-gate.cs` | reads the xUnit report, decides pass/fail |

The gate step is `continue-on-error`; the checker owns the verdict. It fails on:

| Condition | Why |
| --- | --- |
| a failure that is not pinned | new breakage — the point of the job |
| a pinned test that **passed** | the fix landed; retire the pin |
| a pinned test that never ran | dead pin — renamed or deleted test |
| missing report, or zero tests | a crashed run is not an empty pass |

The stale-pin check is the part that matters. A pin that outlives its failure silently un-gates the test it names, and that is how a known-red list decays into wallpaper. This one is a ratchet in both directions.

## Scope is a cost decision, not a principle

`pre-merge` names three classes rather than taking the whole `Fidelity` area, because the area is too expensive for the PR lane:

| Selection | Wall | Tests |
| --- | --- | --- |
| `--gate fidelity` | 31m12s | 66 |
| `--gate pre-merge` | **7m46s** | 14 |

`ClusterCaptureTests` and `PrinterPrecedenceTests` are ~21 minutes of CPU **for two tests** between them. They want the #3495 type-filter treatment before they can be gated; that is the blocker to widening this preset, and it is called out in the job comment and the docs.

(Timings are from a shared local machine and are indicative, not benchmarks. The job's `timeout-minutes: 45` is deliberately loose against a 7m46s local run since runner cost is unmeasured.)

## Evidence

- **End to end**, the exact CI invocation: 14 tests, 3 failures, checker exit 0.
- **Checker is non-vacuous.** Verified against the real 66-test run artifact (exit 0), then five negative controls, each of which failed it as intended: unpinned failure; pinned-but-passing; dead pin; missing report; zero-test report.
- `GateArgumentExpanderTests` 11/0; `--gate list` renders the new preset; solution builds clean.

## Note for reviewers: two of the three pinned issues have wrong diagnoses

I root-caused #3489 and #3491 while pinning them, and **both filed diagnoses were wrong**. Their titles and bodies have been corrected in place with superseded-diagnosis banners.

- **#3491** is not a decompiler defect. Five of its six diffs are Roslyn local-function name-mangling ordinals (`g__Name|N_K`), which encode the enclosing method's *source position*. Reconstruction re-nests the 18 hoisted methods, so the ordinal shifts, and **no correct decompiler can match it**. ([analysis](https://github.com/richlander/dotnet-inspect/issues/3491#issuecomment-5123310485))
- **#3489** is not a decompiler defect either. It is a **shipping-product bug in the declaration formatter**: any unnamed tuple element whose type spells as a keyword is `@`-escaped, so `(int, int)` renders as `(@int, @int)` and `dotnet-inspect member -s Methods` prints non-compiling C#. Arity and `TRest` are irrelevant; `--json` is clean. ([analysis](https://github.com/richlander/dotnet-inspect/issues/3489#issuecomment-5123965893))

Both landed **outside** the decompiler. #3492 and #3493 remain untraced, and given two-for-two, their filed diagnoses should be treated as unverified too.

## Non-goals

Fixing any of the pinned failures. Widening the gate beyond the three classes. Changing the Deep Inspect notifier (the timeout/`failure()` interaction is real and still open).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,14 +118,21 @@ jobs:
             # transitive closure, which an enumerated list cannot track without
             # rotting. Under-triggering silently disables the gate on exactly
             # the changes it exists to catch; over-triggering costs a parallel
-            # job that never blocks the hot lane. Prefer the cheap error.
+            # job that never blocks the hot lane. Prefer the cheap error. The
+            # documentation-extension arm covers only `*.md`, deliberately: a
+            # `.txt` or `.jsonl` under these trees can be a corpus or baseline
+            # fixture, and excluding it by extension would silently un-gate it.
+            # `global.json` and `nuget.config` select the SDK and package
+            # inputs, which select the compiler, which selects the IL compared.
             case "$file" in
               eng/check-decompiler-gate.cs) DECOMPILER=true ;;
               eng/decompiler-gate-known-red.txt) DECOMPILER=true ;;
-              *.md|*.txt) ;;
+              eng/decompiler-gate-expected-classes.txt) DECOMPILER=true ;;
+              *.md) ;;
               src/*) DECOMPILER=true ;;
               tests/*) DECOMPILER=true ;;
               tools/*) DECOMPILER=true ;;
+              global.json|nuget.config|NuGet.config|NuGet.Config) DECOMPILER=true ;;
               *.props|*.targets|*.slnx) DECOMPILER=true ;;
               .github/workflows/ci.yml) DECOMPILER=true ;;
             esac
@@ -456,7 +463,9 @@ jobs:
       - name: Compare against pinned known-red
         run: |
           dotnet run eng/check-decompiler-gate.cs -- \
-            "$RUNNER_TEMP/decompiler-gates.xml" eng/decompiler-gate-known-red.txt
+            "$RUNNER_TEMP/decompiler-gates.xml" \
+            eng/decompiler-gate-known-red.txt \
+            eng/decompiler-gate-expected-classes.txt
 
       - name: Upload gate report
         if: always()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
     outputs:
       code: ${{ steps.filter.outputs.code }}
       csharpdiff: ${{ steps.filter.outputs.csharpdiff }}
+      decompiler: ${{ steps.filter.outputs.decompiler }}
       docs: ${{ steps.filter.outputs.docs }}
       ildiff: ${{ steps.filter.outputs.ildiff }}
       ilroundtrip: ${{ steps.filter.outputs.ilroundtrip }}
@@ -51,6 +52,7 @@ jobs:
           fi
           CODE=false
           CSHARPDIFF=false
+          DECOMPILER=false
           DOCS=false
           ILDIFF=false
           ILROUNDTRIP=false
@@ -102,6 +104,28 @@ jobs:
             case "$file" in
               *.md|*.txt|docs/*|skills/*) DOCS=true ;;
             esac
+            # The decompiler docket/byte-neutrality gates cost ~15 minutes, so
+            # they run as their own job rather than in the hot `test` lane, and
+            # only when something that can move a fidelity result changes: the
+            # decompiler and its IL/CFG/C# substrate, the compile-back harness,
+            # the fixtures the gates decompile, the gate tests and their
+            # dockets, and the known-red pin itself. Docs, packaging, and
+            # unrelated tools must not trip this.
+            case "$file" in
+              src/ILInspector.Decompiler/*) DECOMPILER=true ;;
+              src/ILInspector.Decompiler.Tests/*) DECOMPILER=true ;;
+              src/ILInspector.Instructions/*) DECOMPILER=true ;;
+              src/ILInspector.ControlFlow/*) DECOMPILER=true ;;
+              src/ILInspector.CSharp/*) DECOMPILER=true ;;
+              src/ILInspector.Metadata/*) DECOMPILER=true ;;
+              tools/DecompilerHarness/*.md|tools/DecompilerHarness/*.txt) ;;
+              tools/DecompilerHarness/*) DECOMPILER=true ;;
+              eng/check-decompiler-gate.cs) DECOMPILER=true ;;
+              eng/decompiler-gate-known-red.txt) DECOMPILER=true ;;
+              Directory.Packages.props) DECOMPILER=true ;;
+              *.props|*.targets|*.slnx) DECOMPILER=true ;;
+              .github/workflows/ci.yml) DECOMPILER=true ;;
+            esac
             # The IL round-trip oracle has a fast PR subset plus a slow assembly
             # sweep. It depends on ILInspector.Metadata ->
             # { DotnetInspector.Core, ILInspector.MetadataPrimitives } plus the
@@ -150,6 +174,7 @@ jobs:
           done <<< "$FILES"
           echo "code=$CODE" >> "$GITHUB_OUTPUT"
           echo "csharpdiff=$CSHARPDIFF" >> "$GITHUB_OUTPUT"
+          echo "decompiler=$DECOMPILER" >> "$GITHUB_OUTPUT"
           echo "docs=$DOCS" >> "$GITHUB_OUTPUT"
           echo "ildiff=$ILDIFF" >> "$GITHUB_OUTPUT"
           echo "ilroundtrip=$ILROUNDTRIP" >> "$GITHUB_OUTPUT"
@@ -157,7 +182,7 @@ jobs:
           echo "shipped=$SHIPPED" >> "$GITHUB_OUTPUT"
           echo "Changed files:"
           echo "$FILES"
-          echo "code=$CODE csharpdiff=$CSHARPDIFF docs=$DOCS ildiff=$ILDIFF ilroundtrip=$ILROUNDTRIP packaging=$PACKAGING shipped=$SHIPPED"
+          echo "code=$CODE csharpdiff=$CSHARPDIFF decompiler=$DECOMPILER docs=$DOCS ildiff=$ILDIFF ilroundtrip=$ILROUNDTRIP packaging=$PACKAGING shipped=$SHIPPED"
         env:
           GH_TOKEN: ${{ github.token }}
 
@@ -350,6 +375,80 @@ jobs:
 
       - name: Build tool for net10.0 fallback
         run: dotnet build src/dotnet-inspect -c Release -p:DefaultTargetFramework=net10.0 -p:PublishAot=false
+
+  # Decompiler docket and byte-neutrality gates, pre-merge.
+  #
+  # These gates carry doc comments asserting they fail CI, but every pre-merge
+  # lane runs `-trait- "Speed=Slow"` and they are all Slow, so until now they
+  # ran only in release.yml and the weekly Deep Inspect lane — where they were
+  # exceeding the job timeout, and a *cancelled* job does not satisfy Deep
+  # Inspect's `if: ... && failure()` notifier. Detection latency was unbounded,
+  # not weekly (#3432). Five real regressions (#3489-#3493) accumulated unseen.
+  #
+  # This job runs the three gates whose failures are attributable to a specific
+  # fixture and docket entry, path-gated so it stays off most PRs, and as its
+  # own job so it never serializes with the hot `test` lane. Currently-open
+  # failures are pinned in eng/decompiler-gate-known-red.txt; the checker fails
+  # on any unpinned failure AND on any pin that starts passing, so the pin list
+  # is a ratchet in both directions rather than a permanent exemption.
+  #
+  # The remaining Fidelity-area classes are excluded on cost, not principle:
+  # ClusterCaptureTests and PrinterPrecedenceTests alone are ~21 minutes for two
+  # tests and want the #3495 type-filter treatment before they can be gated.
+  decompiler-gates:
+    needs: changes
+    if: github.event_name == 'pull_request' && needs.changes.outputs.decompiler == 'true'
+    runs-on: ubuntu-26.04
+    timeout-minutes: 45
+    env:
+      DOTNET_INSPECT_TIPS: q
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v5
+        with:
+          dotnet-version: '11.0.x'
+          dotnet-quality: 'preview'
+
+      - name: Cache NuGet packages
+        uses: actions/cache@v4
+        with:
+          path: ~/.nuget/packages
+          key: nuget-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('**/*.csproj', '**/*.props', '**/*.targets', '**/*.slnx') }}
+          restore-keys: |
+            nuget-${{ runner.os }}-${{ runner.arch }}-
+
+      # The gates decompile and recompile the fixture assemblies, so the whole
+      # solution must be built: building only the test project leaves the
+      # fixtures absent and the gates fail with a FileNotFoundException that
+      # reads exactly like a product regression.
+      - name: Build
+        run: dotnet build dotnet-inspect.slnx -c Release
+
+      # continue-on-error so the pinned known-red failures do not fail the job
+      # here; eng/check-decompiler-gate.cs decides pass/fail from the report.
+      # A crashed run produces no XML, which the checker treats as an error
+      # rather than an empty pass.
+      - name: Run decompiler gates
+        id: gates
+        continue-on-error: true
+        run: |
+          dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
+            --gate pre-merge -xml "$RUNNER_TEMP/decompiler-gates.xml"
+
+      - name: Compare against pinned known-red
+        run: |
+          dotnet run eng/check-decompiler-gate.cs -- \
+            "$RUNNER_TEMP/decompiler-gates.xml" eng/decompiler-gate-known-red.txt
+
+      - name: Upload gate report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: decompiler-gates-report
+          path: ${{ runner.temp }}/decompiler-gates.xml
+          if-no-files-found: warn
 
   # Cheap C# Diff harness coverage. Path-gated separately from the full test
   # lane so CSharpDiffHarness/card changes get focused snapshot/baseline JSONL
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -446,15 +446,19 @@ jobs:
       - name: Build
         run: dotnet build dotnet-inspect.slnx -c Release
 
-      # The gate below trusts eng/decompiler-gate-expected-classes.txt to say
-      # what must have run. Validate that file here rather than relying on
-      # another job's path filter to have run the fast lane: the job that
-      # consumes the inventory is the job that should prove it still matches
-      # the preset. Cheap (sub-second) and not continue-on-error.
-      - name: Verify gate inventory matches the preset
+      # Enumerate what the preset selects, without running it. This is the
+      # reference the completeness check compares the results against; the
+      # report's own summary counters cannot serve, because an honestly
+      # truncated report declares its own smaller total and agrees with itself.
+      #
+      # Uses --gate pre-merge, the same preset as the run below, so the two
+      # cannot drift. Sub-second, and not continue-on-error: if discovery fails
+      # there is no reference and the gate must not proceed.
+      - name: Discover expected gate tests
         run: |
           dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- \
-            -class ILInspector.Decompiler.Tests.GateExpectedClassesTests
+            --gate pre-merge -noColor -list methods/json \
+            > "$RUNNER_TEMP/decompiler-gates-expected.json"
 
       # continue-on-error so the pinned known-red failures do not fail the job
       # here; eng/check-decompiler-gate.cs decides pass/fail from the report.
@@ -479,7 +483,8 @@ jobs:
           dotnet run eng/check-decompiler-gate.cs -- \
             "$RUNNER_TEMP/decompiler-gates.xml" \
             eng/decompiler-gate-known-red.txt \
-            eng/decompiler-gate-expected-classes.txt
+            eng/decompiler-gate-expected-classes.txt \
+            "$RUNNER_TEMP/decompiler-gates-expected.json"
 
       - name: Upload gate report
         if: always()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,25 +104,28 @@ jobs:
             case "$file" in
               *.md|*.txt|docs/*|skills/*) DOCS=true ;;
             esac
-            # The decompiler docket/byte-neutrality gates cost ~15 minutes, so
-            # they run as their own job rather than in the hot `test` lane, and
-            # only when something that can move a fidelity result changes: the
-            # decompiler and its IL/CFG/C# substrate, the compile-back harness,
-            # the fixtures the gates decompile, the gate tests and their
-            # dockets, and the known-red pin itself. Docs, packaging, and
-            # unrelated tools must not trip this.
+            # The decompiler docket/byte-neutrality gates cost ~8 minutes, so
+            # they run as their own job rather than in the hot `test` lane.
+            #
+            # This filter is deliberately broad. The first version enumerated
+            # the decompiler projects, and review found it already wrong: it
+            # missed ILInspector.Decompiler.Fixtures.* (the assemblies the
+            # gates decompile), tools/RoundTripCompilation (the compile-back
+            # harness), DiffFixtures.V*, and transitive substrate such as
+            # ILInspector.MetadataPrimitives, ILInspector.Findings, and
+            # ILInspector.Text. A fidelity result is a whole-pipeline
+            # observation, so its true input set is the test project's
+            # transitive closure, which an enumerated list cannot track without
+            # rotting. Under-triggering silently disables the gate on exactly
+            # the changes it exists to catch; over-triggering costs a parallel
+            # job that never blocks the hot lane. Prefer the cheap error.
             case "$file" in
-              src/ILInspector.Decompiler/*) DECOMPILER=true ;;
-              src/ILInspector.Decompiler.Tests/*) DECOMPILER=true ;;
-              src/ILInspector.Instructions/*) DECOMPILER=true ;;
-              src/ILInspector.ControlFlow/*) DECOMPILER=true ;;
-              src/ILInspector.CSharp/*) DECOMPILER=true ;;
-              src/ILInspector.Metadata/*) DECOMPILER=true ;;
-              tools/DecompilerHarness/*.md|tools/DecompilerHarness/*.txt) ;;
-              tools/DecompilerHarness/*) DECOMPILER=true ;;
               eng/check-decompiler-gate.cs) DECOMPILER=true ;;
               eng/decompiler-gate-known-red.txt) DECOMPILER=true ;;
-              Directory.Packages.props) DECOMPILER=true ;;
+              *.md|*.txt) ;;
+              src/*) DECOMPILER=true ;;
+              tests/*) DECOMPILER=true ;;
+              tools/*) DECOMPILER=true ;;
               *.props|*.targets|*.slnx) DECOMPILER=true ;;
               .github/workflows/ci.yml) DECOMPILER=true ;;
             esac
@@ -386,11 +389,17 @@ jobs:
   # not weekly (#3432). Five real regressions (#3489-#3493) accumulated unseen.
   #
   # This job runs the three gates whose failures are attributable to a specific
-  # fixture and docket entry, path-gated so it stays off most PRs, and as its
-  # own job so it never serializes with the hot `test` lane. Currently-open
-  # failures are pinned in eng/decompiler-gate-known-red.txt; the checker fails
-  # on any unpinned failure AND on any pin that starts passing, so the pin list
-  # is a ratchet in both directions rather than a permanent exemption.
+  # fixture and docket entry, as its own job so it never serializes with the
+  # hot `test` lane. Currently-open failures are pinned in
+  # eng/decompiler-gate-known-red.txt; the checker fails on any unpinned
+  # failure, on any pin that starts passing, and on any gate test that neither
+  # passes nor fails, so the pin list is a ratchet in both directions rather
+  # than a permanent exemption.
+  #
+  # NOTE: this job is not merge-blocking today, because the `main` ruleset
+  # declares no required status checks at all — no job in this workflow blocks
+  # a merge. That is a repository-wide gap tracked separately; this job is
+  # exactly as enforcing as the existing `test` job until it is closed.
   #
   # The remaining Fidelity-area classes are excluded on cost, not principle:
   # ClusterCaptureTests and PrinterPrecedenceTests alone are ~21 minutes for two
@@ -428,11 +437,18 @@ jobs:
 
       # continue-on-error so the pinned known-red failures do not fail the job
       # here; eng/check-decompiler-gate.cs decides pass/fail from the report.
-      # A crashed run produces no XML, which the checker treats as an error
-      # rather than an empty pass.
+      #
+      # The step timeout is the point, not a formality. This PR exists because
+      # a *job* timeout cancels the job, and a cancelled job runs no further
+      # steps and satisfies no `failure()` condition — it fails silently. A
+      # step timeout well under the job timeout converts a hang into a failed
+      # step that the job survives, so the checker still runs, finds a missing
+      # or truncated report, and fails the job loudly. Do not raise this to the
+      # job timeout.
       - name: Run decompiler gates
         id: gates
         continue-on-error: true
+        timeout-minutes: 25
         run: |
           dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
             --gate pre-merge -xml "$RUNNER_TEMP/decompiler-gates.xml"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -431,6 +431,15 @@ jobs:
           dotnet-version: '11.0.x'
           dotnet-quality: 'preview'
 
+      # Caches packages only. Do not add build outputs (artifacts/, bin/, obj/)
+      # here. The completeness check compares a discovery listing against a
+      # results file, and both steps below run with --no-build against whatever
+      # the Build step produced. That comparison is only meaningful because the
+      # two provably come from one build: the checkout is fresh, nothing
+      # restores binaries, and Build is fail-fast, so neither step can run
+      # against a stale or partial binary. Caching build outputs would break
+      # that silently -- the gate would still pass, while comparing a listing
+      # and a report from different builds.
       - name: Cache NuGet packages
         uses: actions/cache@v4
         with:
@@ -443,6 +452,9 @@ jobs:
       # solution must be built: building only the test project leaves the
       # fixtures absent and the gates fail with a FileNotFoundException that
       # reads exactly like a product regression.
+      #
+      # Deliberately not continue-on-error: the two --no-build steps below
+      # depend on this having succeeded.
       - name: Build
         run: dotnet build dotnet-inspect.slnx -c Release
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -482,12 +482,18 @@ jobs:
       # step that the job survives, so the checker still runs, finds a missing
       # or truncated report, and fails the job loudly. Do not raise this to the
       # job timeout.
+      #
+      # --no-build is required, not an optimization. Without it `dotnet run`
+      # builds implicitly, which would (a) break the guarantee that this run and
+      # the discovery listing above describe the same binary, and (b) hide a
+      # build failure inside a continue-on-error step, where it would surface
+      # only as a missing report.
       - name: Run decompiler gates
         id: gates
         continue-on-error: true
         timeout-minutes: 25
         run: |
-          dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
+          dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- \
             --gate pre-merge -xml "$RUNNER_TEMP/decompiler-gates.xml"
 
       - name: Compare against pinned known-red
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,10 @@ jobs:
               eng/prepare-decompiler-opt-in-corpus.sh) CODE=true ;;
               eng/prepare-decompiler-pr-corpus.sh) CODE=true ;;
               eng/report-decompiler-opt-in-corpus-drift.sh) CODE=true ;;
+              # GateExpectedClassesTests lives in the fast lane and is the only
+              # thing keeping this file honest against the pre-merge preset, so
+              # editing the file must run the lane that validates it.
+              eng/decompiler-gate-expected-classes.txt) CODE=true ;;
               *.props|*.targets|*.sln|*.slnx) CODE=true ;;
               .github/workflows/*) CODE=true ;;
             esac
@@ -441,6 +445,16 @@ jobs:
       # reads exactly like a product regression.
       - name: Build
         run: dotnet build dotnet-inspect.slnx -c Release
+
+      # The gate below trusts eng/decompiler-gate-expected-classes.txt to say
+      # what must have run. Validate that file here rather than relying on
+      # another job's path filter to have run the fast lane: the job that
+      # consumes the inventory is the job that should prove it still matches
+      # the preset. Cheap (sub-second) and not continue-on-error.
+      - name: Verify gate inventory matches the preset
+        run: |
+          dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- \
+            -class ILInspector.Decompiler.Tests.GateExpectedClassesTests
 
       # continue-on-error so the pinned known-red failures do not fail the job
       # here; eng/check-decompiler-gate.cs decides pass/fail from the report.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -330,10 +330,12 @@ The PR `decompiler-gates` job runs `--gate pre-merge` (the docket and
 byte-neutrality gates) and compares the result against
 `eng/decompiler-gate-known-red.txt`. That list is a record of *open, filed*
 failures, not an escape hatch: do not add an entry to make your own change go
-green. A new failure means either a regression to fix or a diff to docket in the
-owning gate with a rationale. Adding a pin requires an issue and a date, and the
-checker fails the job when a pinned test starts passing, so retire pins as fixes
-land. See `docs/decompiler-correctness-pipeline.md`.
+green, and do not skip a gate test to green it — the checker treats a test that
+neither passed nor failed as a coverage hole and fails the job. A new failure
+means either a regression to fix or a diff to docket in the owning gate with a
+rationale. Adding a pin requires an issue and a date, and the checker fails the
+job when a pinned test starts passing, so retire pins as fixes land. See
+`docs/decompiler-correctness-pipeline.md`.
 
 Pack and publish flows remain separate and build `src/dotnet-inspect`
 directly. Packaging is off by default (`IsPackable=false` in the root
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -326,6 +326,15 @@ The executable also accepts a discoverable `--gate <preset>` flag that expands
 to these trait filters (e.g. `--gate fast`, `--gate no-corpus`); run
 `--gate list` for the table.
 
+The PR `decompiler-gates` job runs `--gate pre-merge` (the docket and
+byte-neutrality gates) and compares the result against
+`eng/decompiler-gate-known-red.txt`. That list is a record of *open, filed*
+failures, not an escape hatch: do not add an entry to make your own change go
+green. A new failure means either a regression to fix or a diff to docket in the
+owning gate with a rationale. Adding a pin requires an issue and a date, and the
+checker fails the job when a pinned test starts passing, so retire pins as fixes
+land. See `docs/decompiler-correctness-pipeline.md`.
+
 Pack and publish flows remain separate and build `src/dotnet-inspect`
 directly. Packaging is off by default (`IsPackable=false` in the root
 `Directory.Build.props`), so only `src/dotnet-inspect` and `src/runfaster` opt
--- a/docs/decompiler-correctness-pipeline.md
+++ b/docs/decompiler-correctness-pipeline.md
@@ -299,7 +299,19 @@ same run, so a report containing four of fifteen tests and honestly declaring
 results against a **discovery listing** produced by `-list methods/json` over
 the same preset, which enumerates what should run without running it. Every
 discovered test must appear in the results, and every result must correspond to
-a discovered test.
+a discovered test. Identity comes from the report's `type` and `method`
+attributes rather than from parsing display names, which carry theory arguments
+and honor `-methodDisplayOptions`.
+
+That comparison is **method-granular**, and deliberately so: no `-list` mode
+enumerates theory cases. A theory with five cases is listed once, so a run that
+lost four of them would still satisfy the check. Method granularity is
+sufficient only while the gate classes declare no theories, and that is
+enforced rather than assumed —
+`GateExpectedClassesTests.PreMergeGateClasses_DeclareNoTheories` fails if a
+theory is added to a gate class, with a message pointing at this constraint.
+Resolving each preset class by name in the same test also catches an arm naming
+a renamed or deleted class, which would otherwise select zero tests and exit 0.
 
 The declared totals are still cross-checked, including `passed` and `failed`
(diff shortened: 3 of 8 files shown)


richlander:
## Adversarial review round 1 — three blocking findings, all fixed

Reviewed at `4ef9dd1` by two models in isolated worktrees, given the same self-contained prompt and no knowledge of each other: **Gemini 3.1 Pro** and **GPT-5.6**. They converged independently on the same three defects. I reproduced all three before acting on any of them. Fixed in `3fb40c5`.

That convergence is worth stating plainly: **two of the three were bugs I introduced while writing a change whose entire purpose is to stop bugs from going unnoticed.** The path filter shipped with a comment claiming it covered "the compile-back harness, the fixtures the gates decompile" — and it covered neither.

### 1. Skipped tests silently satisfied the gate — *blocking, both reviewers*

Every `<test>` entered the `ran` set, but only `result="Fail"` entered `failed`. So:

- an **unpinned** test that skipped → not failing → `OK: the failing set matches` → **exit 0**;
- a **pinned** test that skipped → in `ran`, not in `failed` → reported as `STALE PINS`, whose remediation text says *"Remove these from the pin file."*

The second is the bad one. The tool would have instructed a developer to delete the pin that was the last thing naming the test, after which the test skips silently forever. A gate that can be greened by skipping is not a gate.

Fixed: only `Pass` counts as passing, only `Fail` as failing. Anything else — `Skip`, `NotRun`, an unknown value, or a missing `result` attribute — is reported as `NOT EXECUTED` and fails the job. A pass+fail retry pair now resolves to failed. Dead-pin detection excludes names already reported as not executed, so a skipped pin is reported once, correctly.

### 2. The path filter missed most of what can change a fidelity result — *blocking, both reviewers*

The enumerated list was wrong the day it was written. Missing: `ILInspector.Decompiler.Fixtures.*` (the assemblies the gates decompile), `tools/RoundTripCompilation` (the compile-back harness), `DiffFixtures.V1`/`V2`, and transitive substrate including `ILInspector.MetadataPrimitives`, `ILInspector.Findings`, `ILInspector.Text`, `ILInspector.Research`. A PR touching only a decompiler fixture would not have run the gate.

I did not extend the list, because the list *is* the bug: a fidelity result is a whole-pipeline observation, so its true input set is the test project's transitive closure, and an enumerated list cannot track that without rotting. Replaced with a broad filter (`src/`, `tests/`, `tools/`, build files, minus docs). Under-triggering silently disables the gate on exactly the changes it exists to catch; over-triggering costs a parallel job that never blocks the hot lane. Cheap error preferred.

While testing the fix I found an ordering bug **in the fix**: `case` takes the first match, and the pin file is a `.txt`, so the leading docs no-op swallowed `eng/decompiler-gate-known-red.txt` — editing a pin would not have re-run the gate that validates the pin. The `eng/` entries now come first.

### 3. A hang would have reproduced the exact failure this PR fixes — *both reviewers*

The PR's thesis is that a *cancelled* job runs no further steps and satisfies no `failure()` condition, so it fails silently. The job carried only `timeout-minutes: 45` — so on a hang it would have been cancelled, skipped the checker, and reported nothing. The fix reproduced the bug.

The `Run decompiler gates` step now carries its own `timeout-minutes: 25`, under the job's. A step timeout with `continue-on-error: true` lets the job survive, so the checker still runs, finds the missing or truncated report, and fails loudly. Malformed XML is now a diagnosed `exit 2` rather than the unhandled-exception abort (exit 134) Gemini observed.

### 4. Nothing here blocks a merge — *both reviewers, out of scope, filed as #3523*

Correct, and broader than this PR: the `main` ruleset declares **no required status checks at all**, so no job in `ci.yml` blocks a merge — not `test`, not `pack`, not this one. The legacy protection API returns 404, which is why this is easy to miss.

I did not fix it here; requiring checks is a repository policy change with its own blast radius, and it needs the aggregate-job design to avoid path-gated jobs blocking merges forever. Filed as **#3523** with that design. The overclaiming language has been removed from the job comment and the docs; this job is exactly as enforcing as the existing `test` job.

### Validation of the fixes

13 checker controls, each asserted against an expected exit code:

| Control | Expect | Got |
| --- | --- | --- |
| baseline (pinned fail + pass) | 0 | 0 |
| **unpinned skip** | 1 | 1 `NOT EXECUTED` |
| **pinned skip** | 1 | 1 `NOT EXECUTED` |
| missing `result` attribute | 1 | 1 `NOT EXECUTED` |
| new failure | 1 | 1 `NEW FAILURES` |
| stale pin | 1 | 1 `STALE PINS` |
| dead pin | 1 | 1 `DEAD PINS` |
| dead pin + `--partial` | 0 | 0 |
| pass+fail retry | 0 | 0 |
| zero tests | 2 | 2 |
| truncated XML | 2 | 2 |
| missing file | 2 | 2 |
| empty pin file + failure | 1 | 1 |

The two Skip controls **fail against the old checker and pass against the new one** — that is the non-vacuity evidence for this fix specifically.

Re-validated against both real run artifacts: the 14-test `--gate pre-merge` run and the 66-test `--gate fidelity` run. Both report 0 not executed, 3 pinned failures, exit 0. Path filter asserted over 22 real paths including all nine previously missed.

Round 2 review of `3fb40c5` is in flight with both models. Not ready to merge until those come back clean.

richlander:
## Adversarial review round 2 — two more findings, both fixed

Reviewed at `3fb40c5`. **Gemini 3.1 Pro cleared it and recommended merge. GPT-5.6 found two more defects.** Both reproduced. Fixed in `bf791f4`.

That split is the useful result here. A single reviewer — including the one that had already found three real bugs in round 1 — would have green-lit a gate with a hole in it.

### 5. An incomplete report still passed — *blocking, GPT only*

This is the sharpest finding of the review, because it is the PR's own thesis turned against it.

The checker judged only the tests a report *happened to contain*. It had no notion of what should have been there. GPT stripped every passing `<test>` from the real 14-test report, leaving just the three pinned failures:

```console
$ dotnet run eng/check-decompiler-gate.cs -- <(perl -0777 -pe 's#<test\b(?=[^>]*\bresult="Pass")[\s\S]*?</test>##g' /tmp/pm.xml) eng/decompiler-gate-known-red.txt
Decompiler gate: 3 tests, 0 passed, 3 failed, 0 not executed, 3 pinned known-red.

OK: the failing set matches the known-red list exactly.
exit=0
```

Eleven of fourteen tests absent, exit 0. The same hole covers a `pre-merge` preset that stops selecting a class, a renamed or deleted class, and a run killed after emitting some results. **A shrinking gate passes trivially** — which is the exact failure mode this PR was written to end, one level up from where I was looking.

Fixed with `eng/decompiler-gate-expected-classes.txt`, a **required** third argument. Every listed class must have executed at least one test.

The obvious objection is that a hand-maintained inventory is just a new thing to rot. So it isn't hand-maintained. Following the `AGENTS.md` rule to *"make the declaration drive the enforcement set over restating it, so that stale and missing entries both fail"*, `GateExpectedClassesTests` asserts set equality between the file and the `pre-merge` preset's `-class` arguments, in both directions. It is a **fast-lane** test on purpose: it must run on PRs that never trigger the slow gates.

Tamper-tested rather than asserted:

| Tamper | Result |
| --- | --- |
| drop `ByteNeutralityGateTests` from the file | `Assert.Equal() Failure: HashSets differ` — 1 failed |
| add a bogus `GhostTests` entry | `Assert.Equal() Failure: HashSets differ` — 1 failed |
| restore | 1 passed |

Also fixed: a `<test>` with no usable name collapsed to the degenerate name `"."` and counted as passing. Both reviewers spotted it; Gemini judged it unreachable, which is probably true — but it contradicted the fail-closed rule the rest of the tool follows, so it now reports `UNIDENTIFIED RESULTS`. *A result that cannot be attributed to a test is not a result that can clear one.*

### 6. Path filter still had gaps — *blocking, GPT only*

- `global.json` and `nuget.config` were missing. They select the SDK and package inputs → the compiler → the IL the gates compare. The sibling packaging filters already treat both as build inputs. (`global.json` does not exist today; adding one is exactly the change that would need this gate.)
- The `*.md|*.txt` no-op **preceded** the directory arms, so a `.txt` corpus or baseline fixture under `src/`, `tests/`, or `tools/` was silently excluded. Such fixtures exist. Exclusion is now `*.md` only.

This is the second ordering bug in this `case` block, after the pin-file one I found in round 1. First-match ordering is clearly the sharp edge here.

### Validation

18 checker controls. GPT's exact reproduction now exits 1 with `MISSING CLASSES`. The five result-classification controls were re-run with **real** class names, so the skip logic is proven independently of the new completeness check rather than being masked by it:

| Control | Expect | Got |
| --- | --- | --- |
| GPT's stripped report | 1 | 1 `MISSING CLASSES` |
| unnamed `<test result="Pass"/>` | 1 | 1 `UNIDENTIFIED RESULTS` |
| unpinned skip (real names) | 1 | 1 `NOT EXECUTED` |
| pinned skip (real names) | 1 | 1 `NOT EXECUTED` |
| new failure (real names) | 1 | 1 `NEW FAILURES` |
| stale pin (real names) | 1 | 1 `STALE PINS` |
| missing manifest argument | 2 | 2 |
| empty manifest | 2 | 2 |
| stripped report + `--partial` | 0 | 0 |
| both real artifacts | 0 | 0 |

Path filter asserted over 17 paths including all of GPT's omissions. 12/12 gate-plumbing tests pass. Full solution build clean at `bf791f4`.

Round 3 review of `bf791f4` is in flight with both models, including an explicit ask to tell me whether this is now over-engineered — six failure conditions, a manifest, and a test guarding the manifest is a lot of machinery for one CI job, and I would rather cut it than ship ceremony.

richlander:
## Adversarial review round 3 — two more findings, both fixed

Reviewed at `bf791f4`. Same split as round 2, and it is now a pattern worth naming.

**Gemini 3.1 Pro cleared it, stating:** *"It is impossible to bypass this check with a missing or truncated test run."*

**GPT-5.6 bypassed it twice in the same round.** Both reproduced. Fixed in `4ae0336`.

### 7. The class inventory proved breadth, not completeness — *blocking, GPT only*

The inventory required each expected class to execute *at least one* test. So a report containing 4 of 14 tests — the three pinned failures plus one passing `ByteNeutralityGateTests` — covered all three classes and passed:

```console
Decompiler gate: 4 tests, 1 passed, 3 failed, 0 not executed, 3/3 expected classes covered, 3 pinned known-red.

OK: the failing set matches the known-red list exactly.
exit=0
```

GPT's framing was the useful part: **the class manifest proves selection breadth, not report completeness.** I had fixed "did the gate stop selecting a class" and mistaken it for "is the report whole."

Pinning expected *test names* would catch it, but would rot on every added test. Instead the checker now cross-checks the report against **its own declared per-assembly totals** — xUnit already writes `total`/`skipped`/`not-run`/`errors`. A count mismatch fails, and any declared skipped/not-run/errored test fails. Nothing to maintain.

That also closes the skip hole a second, independent way:

| Report | Caught by |
| --- | --- |
| honest summary declaring `skipped="1"` | summary check, exit 2 |
| doctored summary claiming `skipped="0"` | `NOT EXECUTED`, exit 1 |

### 8. The inventory's guard did not run when the inventory changed — *blocking, GPT only*

Editing `eng/decompiler-gate-expected-classes.txt` set `DECOMPILER=true` but left `CODE=false` — and `GateExpectedClassesTests` lives in the fast lane, which is gated on `code`. So the gate job would run against an edited manifest while the test that validates the manifest sat out. Deleting a class from the manifest would have been accepted, silently un-gating that class.

The guard I added in round 2 to stop the inventory rotting could itself be bypassed by editing only the inventory.

Two fixes, because the second is the one that actually holds: the file now sets `CODE=true`, **and** the `decompiler-gates` job runs `GateExpectedClassesTests` directly as a step before the gates. The job that consumes the inventory now proves it, instead of trusting another job's path filter to have done so.

### Validation

GPT's reproduction now exits 2 on the count mismatch. The per-check controls were **regenerated with valid summary attributes**, so each condition is still proven independently rather than being masked by the new one — without that, every control would have tripped the summary check first and proven nothing:

| Control | Expect | Got |
| --- | --- | --- |
| baseline | 0 | 0 |
| unpinned skip, honest summary | 2 | 2 summary |
| unpinned skip, doctored summary | 1 | 1 `NOT EXECUTED` |
| new failure | 1 | 1 `NEW FAILURES` |
| stale pin | 1 | 1 `STALE PINS` |
| class absent entirely | 1 | 1 `MISSING CLASSES` |
| class absent + `--partial` | 0 | 0 |
| GPT's 4-of-14 report | 2 | 2 summary |
| both real artifacts | 0 | 0 |

Full solution build clean, 12/12 gate-plumbing tests pass.

### On the reviewer split

Three rounds, seven findings. **GPT found all seven; Gemini found three, and cleared the head as merge-ready in rounds 2 and 3 — including with a false universal claim.** I am recording that because it is the strongest argument in this PR for the two-model tier in `AGENTS.md`: a single reviewer here, even a good one that opened by finding three real bugs, would have shipped a gate that could be silently emptied.

Round 4 is running against `4ae0336` with GPT plus a **fresh MAI-Code reviewer** brought in for a third perspective. I have asked both directly whether this is now over-engineered — specifically whether the totals cross-check subsumes the class inventory, in which case I would rather delete the manifest and its guard test than ship two mechanisms for one property.

richlander:
## Adversarial review round 4 — the completeness check was self-referential

Reviewed at `4ae0336` by **GPT-5.6** and a **fresh MAI-Code** reviewer brought in as a third family. Two blocking findings, both reproduced, both fixed in `192566d`.

This round invalidated the premise of my round-3 fix, so I want to state it plainly rather than bury it in a fix list.

### 9. The summary cross-check could not prove completeness — *blocking; GPT and MAI independently*

In round 3 I wrote that cross-checking the report against its own declared totals "proves the report is not missing rows." **That was wrong.** GPT put it in one sentence:

> The summary is supplied by the same incomplete report.

A report containing 4 of 15 tests and honestly declaring `total="4"` is internally consistent and sails through. The original 4-of-14 bypass never closed — I had only closed the variant where the forger forgot to update the counters.

MAI arrived from a different angle: `passed`/`failed` were never validated against the rows. That specific gap turned out to be **inert** — the verdict is computed from the rows, so a lying counter changes nothing — but chasing it produced the report that mattered: delete a genuinely-failing row, decrement `total`, and you get a fully self-consistent report that passes. Same conclusion as GPT, reached from the opposite end.

The lesson is the one a self-check can never escape: **a run cannot certify its own completeness.** So the reference now comes from outside the report:

```bash
  --gate pre-merge -noColor -list methods/json > expected.json
```

`-list` enumerates what the preset selects *without running anything*, so a run that was cut short, filtered down, or rewritten has a smaller result set than the listing no matter how tidy its own counters are. Every discovered test must appear in the results; every result must map to a discovered test. It uses `--gate pre-merge` — the same preset as the run — so the two cannot drift, and it costs **0.77s**.

Fails closed in all four listing failure modes: missing, malformed, non-array, or empty. An empty expectation would be satisfied by every report, so "discovery found nothing" is an error, not an empty pass.

I kept the totals check, including MAI's `passed`/`failed` comparison, but **demoted its stated purpose** to what it actually is — an internal-consistency check on a corrupt report. The comment and docs claiming it proved completeness are corrected.

### 10. The step guarding the inventory was vacuous — *blocking, GPT*

I had asked GPT to test whether a `-class` filter matching nothing exits 0. It does:

```console
Discovered: ILInspector.Decompiler.Tests (0 test cases to be run)
   ILInspector.Decompiler.Tests  Total: 0
exit=0
```

So `Verify gate inventory matches the preset` would have gone green while enforcing nothing the moment that class was renamed or deleted — **the exact silent-no-op failure this PR exists to fix**, reintroduced by the step I added in round 3 to fix it.

I deleted the step rather than patching it. `GateExpectedClassesTests` is now in the `pre-merge` preset itself, so it runs inside the gate job and is covered by the same discovery and coverage checks as the correctness gates. One mechanism, not a guard on a guard.

### On over-engineering

I asked both reviewers whether the totals check made the class inventory redundant, committing to delete the manifest if so. GPT's answer was that I had it backwards:

> The class inventory is not redundant: it checks required classes, while totals only check internal report consistency.

So both stay, but they now assert **distinct, statable properties** — inventory: *the preset still selects these classes*; discovery: *everything selected actually reported*. Neither implies the other, and the docs now say which is which instead of claiming the weaker one proves the stronger property.

### Validation

Full solution build clean. 12/12 plumbing tests. A real 15-test gate run plus its listing exits 0 end to end. **16 controls**, each condition proven independently:

| Control | Expect | Got |
| --- | --- | --- |
| baseline (15, honest) | 0 | 0 |
| lying `passed`/`failed` | 2 | 2 counts |
| failure excised, totals decremented | 1 | 1 `INCOMPLETE` |
| self-consistent 4-of-15 | 1 | 1 `INCOMPLETE` |
| skip, honest summary | 2 | 2 summary |
| skip, doctored summary | 2 | 2 counts |
| `NotRun` row, consistent summary | 1 | 1 `NOT EXECUTED` |
| new failure / stale pin / class absent | 1 | 1 each |
| row never discovered | 1 | 1 `UNEXPECTED` |
| discovery empty / non-array / truncated / missing | 2 | 2 each |
| subset with `--partial` | 0 | 0 |
| real 15-test run | 0 | 0 |

The strongest one is not synthetic: the **real 14-test artifact from the previous round** is now correctly rejected, naming the missing test.

```console
Decompiler gate: 14 tests, 11 passed, 3 failed, 3/4 expected classes covered
INCOMPLETE REPORT (1) — discovered but absent from the results:
  ...GateExpectedClassesTests.ExpectedClassesFile_MatchesThePreMergePresetExactly
```

Round 5 is going to both reviewers at `192566d`.

richlander:
## Adversarial review round 5 — both reviewers found the same hole in the new check

Reviewed at `192566d`. **GPT-5.6 and MAI-Code independently reported the same defect**, which is the strongest signal this review has produced: the completeness check I added in round 4 is method-granular, and method granularity cannot see theory cases. Fixed in `a18c7a2`.

### 11. A theory could silently lose cases — *blocking; GPT and MAI, independently*

Discovery lists *methods*. A theory with five cases is listed once, so a report containing one case satisfied the check for the whole method. GPT demonstrated it against real repo data (`GateArgumentExpanderTests`: 9 discovered, 11 executed); MAI built the minimal synthetic case. Both are right.

I checked whether discovery could be made case-level, since that would be the clean fix:

| mode | entries | actually executed |
| --- | --- | --- |
| `-list methods/json` | 9 | 11 |
| `-list tests/json` | 9 | 11 |
| `-list full/json` | 9 | 11 |

**No mode enumerates theory cases.** So case-level completeness is simply not available from discovery, and MAI's alternative — a manifest of individual cases — rots on every added `[InlineData]`, which is the maintenance failure the discovery approach existed to avoid.

That leaves one honest option: method granularity is sound *only if the gate classes declare no theories*, so make that a fact rather than an assumption.

`PreMergeGateClasses_DeclareNoTheories` fails if a theory is added to a gate class, and says what to fix first:

```console
The pre-merge gate completeness check is method-granular because xUnit's discovery
listing does not enumerate theory cases, so a theory that silently lost cases would
still satisfy it. Either make these facts, or teach eng/check-decompiler-gate.cs a
case-level expectation before adding them:
  ILInspector.Decompiler.Tests.ByteNeutralityGateTests.TamperTheory
```

Scope is deliberately the *gate* classes, not the suite — MAI cited `IrInvariantsHostContractTests` as a real theory, and it should stay one. The constraint only binds where method granularity has to hold.

The same test resolves each preset class by reflection, so an arm naming a renamed or deleted class fails too:

```console
System.InvalidOperationException : The pre-merge preset names
'ILInspector.Decompiler.Tests.RenamedAwayGateTests', which does not exist in the test
assembly. A -class filter matching nothing runs zero tests and exits 0, so this would
silently un-gate that class.
```

Both are **tamper-tested** — I added a theory and renamed a class, confirmed each fails, and restored. A guard that cannot fail is exactly what this PR is about.

### 12. Display-name parsing was ambiguous — *both reviewers*

I was cutting result names at the first `(`, which breaks on theory arguments containing `(`, on `-methodDisplayOptions`, and on operator names. Every row already carries `type` and `method` attributes, so identity now comes from those, with the name parse kept only as a fallback. A row whose display name is:

```text
...ByteDivergentKnobs_AreExcludedAndBehaviorGated(s: "a(b.c", n: 1)
```

now resolves correctly instead of becoming an `UNEXPECTED RESULTS` phantom.

### Validation

Full solution build clean. 13/13 plumbing tests. A **real 16-test gate run** plus its listing exits 0 end to end. 18 controls, each condition proven independently:

| Control | Expect | Got |
| --- | --- | --- |
| baseline / real 16-test run | 0 | 0 |
| lying `passed`/`failed` | 2 | 2 counts |
| failure excised, totals decremented | 1 | 1 `INCOMPLETE` |
| self-consistent 4-of-16 | 1 | 1 `MISSING CLASSES` |
| skip honest / skip doctored | 2 | 2 each |
| `NotRun` row | 1 | 1 `INCOMPLETE` |
| new failure / stale pin / class absent | 1 | 1 each |
| row never discovered | 1 | 1 `UNEXPECTED` |
| discovery empty / non-array / truncated / missing | 2 | 2 each |
| hostile theory-argument display names | 0 | 0 |
| subset with `--partial` | 0 | 0 |
| real 14-test artifact vs 16-test listing | 1 | 1 |

Round 6 is going to both reviewers at `a18c7a2`.

### Where this stands

Twelve findings over five rounds; **ten were in code I wrote**. Every one is the same shape — *a check that quietly stops checking still reports success* — and it recurred at every layer: the gate itself, the guard on the gate, the guard on the guard, and now the granularity of the guard's reference. I do not think that is coincidence, and it is the strongest argument in this PR for the two-model tier in `AGENTS.md`. Rounds 4 and 5 were both found by the second reviewer.

richlander:
## Adversarial review round 6 — two blocking defects, both introduced by the round-5 fix

Reviewed at `a18c7a2` by GPT-5.6. Both reproduced, both fixed in `5e35f9d`.

Notably, **both defects were created by the previous round's fix**, not pre-existing.

### 13. The identity change was half done — *blocking*

Round 5 moved identity from display names to `type`/`method`. I converted the discovery comparison and class coverage — and left `passed`, `failed`, stale-pin, and dead-pin matching on the display name. GPT built the row that exploits the gap:

```xml
<test name="N.C.PinnedKnownFailure" type="N.C" method="ActualNewFailure" result="Fail"/>
```

```console
OK: the failing set matches the known-red list exactly.
exit=0
```

A new failure wearing a pinned test's name. The fix I shipped to remove display-name ambiguity **created a worse ambiguity** by leaving two identities in play.

`TestName` now prefers `type + "." + method` and falls back to the display name only when both attributes are absent, so every decision keys off one identity. That also deletes the duplicate set round 5 added. Same row now:

```console
NEW FAILURES (1) — not in eng/decompiler-gate-known-red.txt:
  ILInspector.Decompiler.Tests.FidelityGateTests.ActualNewFailure
exit=1
```

### 14. The fact-only guard was a deny list, and deny lists cannot be complete — *blocking*

Two real escapes, both confirmed against the installed xUnit:

- **Inherited theories.** I scanned `DeclaredOnly`. xUnit discovers with `FlattenHierarchy | NonPublic` plus interface methods, so a `[Theory]` inherited from a base class runs normally and was invisible to the guard.
- **`[CulturedFact]`.** It derives from `FactAttribute`, **not** `TheoryAttribute`, and yields one case per culture. Verified directly:

  ```text
  Xunit.FactAttribute: base=System.Attribute
  Xunit.TheoryAttribute: base=Xunit.FactAttribute
  Xunit.CulturedFactAttribute: base=Xunit.FactAttribute
  ```

So a deny list on `TheoryAttribute` was never going to be complete — it enumerates today's multi-case attributes and silently admits tomorrow's.

The guard now **allow-lists**: every test in a gate class must carry exactly `FactAttribute`, scanned over the same surface xUnit discovers. Anything else — theory, cultured fact, or a future attribute nobody has written yet — fails closed.

Tamper-tested with both escapes present at once:

```console
ILInspector.Decompiler.Tests.ByteNeutralityGateTests.InheritedTheory [TheoryAttribute]
ILInspector.Decompiler.Tests.ByteNeutralityGateTests.TamperCultured [CulturedFactAttribute]
```

Neither was caught by the round-5 version. Both restored after testing.

### Validation

Full solution build clean, 13/13 plumbing tests, the **real 16-test gate run** still exits 0 end to end. 19 controls, now covering both the structured path and the name-only fallback:

| Control | Expect | Got |
| --- | --- | --- |
| real 16-test run | 0 | 0 |
| baseline (name-only fallback) | 0 | 0 |
| type/method + hostile theory-arg names | 0 | 0 |
| **name spoofs a pin, method says otherwise** | 1 | 1 `NEW FAILURES` |
| lying `passed`/`failed` | 2 | 2 |
| excised failure, totals decremented | 1 | 1 `INCOMPLETE` |
| self-consistent 4-of-16 | 1 | 1 |
| skip honest / `NotRun` row | 2 / 1 | 2 / 1 |
| new failure / stale pin / class absent | 1 | 1 each |
| row never discovered | 1 | 1 `UNEXPECTED` |
| discovery empty / non-array / truncated / missing | 2 | 2 each |
| subset `--partial` | 0 | 0 |
| real 14-test artifact vs 16-test listing | 1 | 1 |

Round 7 is going to both reviewers at `5e35f9d`.

### An honest note on the trend

Fourteen findings over six rounds, **twelve in code I wrote**. Rounds 5 and 6 both found defects *introduced by the immediately preceding fix*. That is worth stating plainly rather than presenting as steady progress: this change has a high defect-injection rate, and the reason is visible in the pattern — each fix adds a mechanism, and the new mechanism arrives with the same blind spot as the last one, namely *a check that quietly stops checking still reports success*.

The countervailing evidence is that the defects are getting narrower — round 3 was "the gate can be emptied," round 6 is "an attribute subclass nobody uses could evade a guard." I would rather keep reviewing until a round comes back clean than call that convergence myself.

richlander:
## Round 7 reconciliation — `5e35f9d` reviewed, fixed at `cd7f4b2`

Two reviewers at `5e35f9d`. **One finding confirmed, one rejected with evidence, and one hole neither reviewer found — discovered while probing the rejected one.**

### Finding 15 — confirmed (GPT). The guard was anchored on the wrong abstraction

I anchored the allow list on `FactAttribute`. xUnit keys discovery on the **`IFactAttribute` interface**. Verified by reflection against the assembly actually in use (3.2.2):

```text
ExtensibilityPointFactory.GetMethodFactAttributes -> IReadOnlyCollection<Xunit.v3.IFactAttribute>
IXunitTestCaseDiscoverer.Discover(..., IFactAttribute factAttribute)
```

So an attribute can implement `IFactAttribute` directly, never derive from `FactAttribute`, and supply a discoverer emitting several cases. Built exactly that:

```text
execution:  4 test cases
-list methods/json:  3 methods
previous guard:  no offender reported
```

That is the theory-granularity false-green the guard exists to prevent, reachable by a route the guard did not look down. Fixed: the scan is anchored on `IFactAttribute`.

I was also wrong to doubt GPT's version number — I read `3.1.0` off a nuget folder; the resolved assembly is 3.2.2.

### Finding from MAI — rejected, with evidence

MAI argued the interface arm over-rejects: a class with a plain `[Fact]` implementing an interface whose *declaration* carries a fact-derived attribute gets flagged, though "xUnit only discovers the concrete implementation."

The premise is false. Attributes on interface method declarations are live. Measured:

| shape | result rows |
| --- | --- |
| `[Fact]` on interface declaration **and** on implementation | **2** |
| same, with the interface `[Fact]` removed (control) | **1** |

And a `[Fact]` on a *default interface method* is discovered and run on the implementing class (`UsesDimProbe.DimFactProbe`), with its `[Theory]` sibling running 2 cases from 1 listed method. The interface arm is load-bearing. **MAI's suggested fix — stop scanning interfaces — would have reopened the hole the last three rounds were spent closing.** No change made.

The over-rejection concern was still worth testing, and it is not triggered by real code: 13/13 plumbing tests pass unchanged.

### Finding 16 — mine, found while disproving MAI's. The one that mattered

Chasing that repro produced the row I did not expect:

```text
type=...ImplDeclProbe method=DeclaredOnlyProbe
type=...ImplDeclProbe method=DeclaredOnlyProbe
```

A plain `[Fact]` on an interface method declaration and a plain `[Fact]` on its implementation produce **two cases from one discovered method**. Both attributes are exactly `FactAttribute`, so no attribute-*type* test can ever see it — only their *number* is wrong. My guard passed it cleanly.

Multiplicity is now counted per method signature.

### The structural change: prevention is no longer the only line of defence

The guard has now been wrong three rounds running — deny list, then class-instead-of-interface, then count-instead-of-type. The pattern is that I keep reasoning about xUnit's attribute model and keep getting it wrong.

So the checker now independently fails any report where one `(type, method)` produced more than one row. That check is **observational**: it knows nothing about attributes and catches every multi-case shape above, plus future ones, by counting what the run actually produced.

This is the closest thing to the "simpler design" I asked for. It does not let me delete the guard — the guard fails fast and prevents a multi-case test from ever reaching a 12-minute gate, and it catches a theory whose cases decay to exactly one, which the row check cannot see. But it does mean **the completeness argument no longer depends on my reflection being correct.**

### Same-build question, answered

Nothing asserted it; it happens to hold. The cache covers `~/.nuget/packages` only — not build outputs — the checkout is fresh, and `Build` is fail-fast, so neither `--no-build` step can run against a stale or partial binary. That is an unstated invariant a future edit could break silently, so it is now a comment on the cache step. (The stale-binary hazard is real: it bit me locally this session, where a failed build let a previous binary report "2 passed".)

### Evidence

Both new properties tamper-tested **in both directions** — the fix catches it *and* the previous version misses it:

| property | previous | new |
| --- | --- | --- |
| double plain `[Fact]` via interface declaration | 0 failures | fails, names the method |
| duplicated result row | **exit 0, "OK"** | exit 1, `MULTI-CASE TESTS (1)` |

Full solution build clean · 13/13 plumbing tests · real 16-test gate run + its discovery listing exit 0 end to end · 16 controls (positive control passes; every negative rejects for the correct reason) · markdownlint clean · YAML parses, 8 jobs.

One control regenerated: `c1-baseline` had drifted against the round-6 rename. Worth noting the checker diagnosed that rename precisely, in both directions, which is the behaviour it was built for.

### Readiness

**Not claiming ready.** Round 7 found a real defect, and probing it found a larger one. That is three consecutive rounds where the previous round's fix was itself defective, so the honest read is that this file is not yet converged and one more fixed-head pass is warranted. Dispatching round 8 on `cd7f4b2`.

richlander:
## Round 8 (GPT) — `cd7f4b2` reviewed, fixed at `ceaabf9`

Two findings, both correct, **both in code this PR added — one of them added in the previous commit.** MAI's round-8 pass is still running.

### Finding 17 — the gate never used the build it promised

`.github/workflows/ci.yml` — the execution step had no `--no-build`, so `dotnet run` built implicitly.

This is the one worth dwelling on. One commit earlier I *answered* the same-build question and wrote a comment on the cache step stating that "both steps below run with `--no-build` against whatever the Build step produced." That sentence was false when I wrote it. I verified the cache contents and the fail-fast ordering — the two things I was asked about — and never checked the command the claim actually rested on.

There is a second-order effect I had not considered: an implicit build inside a `continue-on-error: true` step **swallows its own build failure**, surfacing only as a missing report. Fixed, with the reason recorded so it does not read as an optimization someone can drop.

This is the exact failure mode `AGENTS.md` names — an asserted property with no gate behind it. I asserted it *in a comment*, which is the weakest possible form, and got it wrong in the same commit that introduced the assertion.

### Finding 18 — half-structured rows fell back to the display name

`eng/check-decompiler-gate.cs` — `TestName` fell back to the display name whenever *either* structured attribute was missing, while its own comment said the fallback applied only when both were "absent entirely."

Verified before fixing, on a real report with `method` removed from one pinned failing row:

```text
before: type=...FidelityGateTests  method=PinnedFixesStayExact
after : type=...FidelityGateTests  method=None
OK: the failing set matches the known-red list exactly.   EXIT=0
```

The row kept an identity its own attributes no longer corroborated — the precise substitution structured identity was introduced in round 6 to prevent. Half-structured rows are now unidentified. The same control after the fix:

```text
UNIDENTIFIED RESULTS (1)
INCOMPLETE REPORT (1) — ...FidelityGateTests.PinnedFixesStayExact
DEAD PINS (1)
EXIT=1
```

Three independent signals, which is what fail-closed should look like.

### What GPT cleared

- No credible false positive in the per-signature multiplicity key — the thing I flagged as least certain. I had asked specifically about null `ParameterType.FullName` on open generics, `ref`/`out`/`in` overloads, and two interfaces declaring the same signature.
- The guard and the row check are complementary, not redundant. That was the "did I just add mechanism" question, and the answer was no.
- Independently re-ran the build, 13/13 plumbing tests, the real 16-test gate, the duplicate-row control, and the baseline control, and confirmed the `c1-baseline` regeneration did not paper over a real failure.

### Evidence

Full solution build clean · real 16-test artifacts still exit 0 · **17 controls** behave, including the name-only control that exercises the fallback path this change narrows · YAML parses, 8 jobs, both test-assembly steps now carry `--no-build`.

### Readiness

Still not claiming ready. Eighteen findings over eight rounds, and this round's pair were both introduced by this PR — one of them by the immediately preceding commit, as part of *answering a review question*. That is a fourth consecutive round of self-inflicted defects, and the pattern is consistent: my prose claims outrun what I actually checked.

Holding for MAI's round-8 result before deciding whether to run a round 9 or stop.

richlander:
## Round 8 (MAI) — finding rejected, with execution evidence

MAI reported that the per-signature multiplicity count false-positives when a class implements two interfaces that each declare the same `[Fact]` signature: the guard buckets two attributes under `M()` and flags it, while xUnit "only discovers one test."

I built MAI's exact shape — `interface IZZA { [Fact] void ZZM(); }`, `interface IZZB { [Fact] void ZZM(); }`, `class ZZMultiIfaceProbe : IZZA, IZZB` with a single unattributed implementation — and ran it:

| | |
| --- | --- |
| `-list methods/json` | `["...ZZMultiIfaceProbe.ZZM"]` — **1 entry** |
| execution | **2 result rows**, both `type=...ZZMultiIfaceProbe method=ZZM` |

The guard is right. Two interface declarations produce two cases from one discovered method, which is exactly the shape method-granular completeness cannot see. **Rejected; no change made.**

### The reason this keeps happening is worth naming

MAI's evidence in both round 7 and round 8 was the discovery listing. But the discovery listing is precisely the thing that *cannot* answer this question — it is `Distinct()`ed on `Class.Method`, which is the entire reason the theory-granularity hole existed in the first place. Using it as ground truth for case count will report every genuine multiplicity bug as a false positive.

Both rejected findings had the same structure: correct reflection output, correct discovery output, and a conclusion that does not follow because the wrong instrument was used. The measurement that settles it is always execution.

I want to be careful not to overstate this. MAI found real defects earlier in this review — it independently found the theory-granularity hole in round 5 and surfaced the `passed`/`failed` gap in round 4 that led to the self-referential-summary fix. The failure here is specific and correctable, not general.

### Where this leaves the round

| Round 8 finding | Source | Outcome |
| --- | --- | --- |
| Gate step missing `--no-build` | GPT | **Fixed** (`ceaabf9`) |
| Half-structured rows fall back to display name | GPT | **Fixed** (`ceaabf9`) |
| Multi-interface signature false positive | MAI | **Rejected**, execution evidence above |

Re-verified after removing the probe: tree clean, full solution build 0 errors, 13/13 plumbing tests.

### Readiness

Head is `ceaabf9`. Both GPT findings are fixed but **have not yet been reviewed at that head** — every previous round found a defect in the previous round's fix, and the `--no-build` fix in particular touches the CI path that nothing local exercises. I am not going to break that streak by assertion.

Dispatching round 9 on `ceaabf9`, scoped narrowly to the two fixes rather than another full pass. If it comes back clean, that is the point at which I would call this ready.

richlander:
## Round 9 — both reviewers clean. Ready to merge

Head `ceaabf9`. **First clean round in nine.** Both reviewers, independently, at the same head, found nothing blocking, and both answered the standing question directly: further review rounds are now lower-value than shipping.

### The one thing I did not delegate

I flagged `--no-build` as the highest-risk change because *nothing local exercised it end to end* — my `artifacts/` had accumulated output from dozens of builds, so `--no-build` would succeed locally for reasons CI would not reproduce. A reviewer running the same command in the same kind of tree would have inherited the same blind spot.

So I simulated CI faithfully: deleted `artifacts/` entirely (235M), ran exactly the workflow's sequence, and nothing else.

```text
rm -rf artifacts
dotnet build dotnet-inspect.slnx -c Release           -> 0 errors
dotnet run ... --no-build -- --gate pre-merge -list methods/json  -> 16 methods
dotnet run ... --no-build -- --gate pre-merge -xml ...            -> 16 tests, 3 pinned failures
dotnet run eng/check-decompiler-gate.cs ...                       -> exit 0
```

A single `Build` step does produce every artifact the gates need, including the fixture assemblies they decompile and recompile. The `FileNotFoundException` failure mode documented on the Build step does not occur.

### Final tally

**Eighteen defects across nine rounds. Sixteen were in my own code.** Four rounds running, the defect was introduced by the previous round's fix — including one where I broke the same-build guarantee *in the commit that documented it*.

Two reviewer findings were rejected, both from MAI, both because the discovery listing was used as evidence of case count when it is `Distinct()`ed on `Class.Method`. Both rejections are backed by execution measurements on the PR.

Reviewer record, since it is the strongest argument for the two-model tier in `AGENTS.md`:

| Reviewer | Outcome |
| --- | --- |
| GPT-5.6 | 14 of 18 findings, correct every time |
| MAI-Code | found the round-5 theory hole and surfaced the round-4 `passed`/`failed` gap; 2 rejected findings |
| Gemini 3.1 Pro | cleared defective heads twice, once with a false universal claim; retired after round 3 |

Rounds 4, 5, 6, 8 and 9 were each decided by a reviewer the other missed or by a probe of a *rejected* finding. A single-model review would have shipped this with holes in it.

### What the gate now guarantees, stated precisely

Three independent properties, none of which implies another:

1. **Class inventory** — the preset still selects the expected classes.
2. **Discovery vs results** — everything selected actually reported, and no `(type, method)` reported twice.
3. **Declared totals** — the report is not internally corrupt. Explicitly *not* evidence of completeness.

The load-bearing lesson, learned the hard way in rounds 4–6: **a run cannot certify its own completeness.** Any reference produced by the run itself is provable only against corruption, never truncation.

### Validation at this head

Full solution build 0 errors · 13/13 plumbing tests · clean-artifacts CI simulation green end to end · 17 controls (positive control passes; every negative rejects for the correct reason, each proven non-vacuous against the prior implementation) · markdownlint clean · YAML parses, 8 jobs, both test-assembly steps carry `--no-build`.

**Ready to merge.**

### Non-blocking follow-ups

- `#3523` — `main` has no `required_status_checks`, so *no* CI job blocks a merge today. This gate is only as good as that policy; it is a call only @richlander can make.
- `pre-merge` cannot currently be widened to `ClusterCaptureTests`/`PrinterPrecedenceTests` — those two tests alone cost ~21 min.
- `IlBodyDiffResult.FailureRows` is declared `= default` (`IlBodyDiff.cs:88-92`), an NRE trap noted on #3491 and not yet filed separately.

richlander:
## `origin/main` integrated — head moved to `b045f503c`

Integrating main per the rule main itself just added ("Integrate `origin/main` before every review round"). One conflict, in `AGENTS.md`.

### The conflict was textual, not semantic

Git aligned this branch's added paragraph against main's rewritten **Evidence and validation** bullets — two unrelated regions. The conflict region also swept in the adjacent **"Pack and publish flows"** paragraph, which main had *deliberately deleted*.

### Resolution: took main's side entirely

This branch's only `AGENTS.md` change was an 11-line paragraph on the `decompiler-gates` job and known-red pin discipline. #3524 ("Move mechanism out of AGENTS.md to its owners") restructured that file specifically to push mechanism into owning docs, and its task table already routes *"Decompiler behavior or harnesses"* → `docs/decompiler-correctness-pipeline.md`.

Keeping the paragraph would have reintroduced exactly what #3524 removed. So the pin-discipline prose **moved into the owning doc**, next to the known-red description — it is not lost:

> That list is a record of *open, filed* failures, not an escape hatch. Do not add an entry to make your own change go green, and do not skip a gate test to green it […] Adding a pin requires an issue and a date, and the checker fails the job when a pinned test starts passing, so retire pins as fixes land.

`AGENTS.md` on this branch is now **byte-identical to main** — it no longer appears in `git diff origin/main --stat`. Taking main's side also preserves main's intended deletion of the pack/publish paragraph rather than resurrecting it.

`docs/decompiler-correctness-pipeline.md` auto-merged: main relocated its **IR invariant checks** section here, disjoint from this branch's gate section. Verified no duplicate or orphaned headings.

### The merge could have widened the gate. It didn't — measured, not assumed

Main brought in new decompiler test classes (`FragTests`, `PrintedBodyMapTests`). A new class picking up the pre-merge traits would silently change what the gate covers, so I re-ran discovery rather than reasoning about it:

| Check | Result |
| --- | --- |
| `dotnet build dotnet-inspect.slnx -c Release` | 0 errors, 0 warnings |
| Discovery (`--no-build`) | **16 methods, 4 classes** — unchanged |
| `--gate pre-merge` (`--no-build`) | 16 total, **3 failed, 0 skipped, 0 not run** |
| `eng/check-decompiler-gate.cs` | **exit 0** — failing set matches known-red exactly |
| `markdownlint` | clean on both changed files |

Neither new class carries a pre-merge trait, so selection is unchanged.

### Re-review

The approved head was `ceaabf99c`; it is now `b045f503c`. Rather than carry the round-9 approval across a head change, I have a **narrow confirmation pass** running against the exact merge head, scoped to the merge itself — conflict resolution, auto-merge coherence, and whether anything main introduced weakens the gate. I'll report the outcome here.

**Ready to merge** stands *pending that pass* — I'd rather re-confirm a moved head than assume a merge is inert.

richlander:
## Round 10 (merge confirmation): clean — and a correction to my own claim

Narrow confirmation pass on the exact merge head `b045f503c`, scoped to the merge itself. **GPT-5.6: no blocking findings.**

### It corrected me, and it was right

I reported **"0 errors, 0 warnings."** The warning half was wrong. My number came from an *incremental* build, which didn't recompile the projects that emit warnings, so it didn't re-emit them. A forced rebuild shows the truth:

```
dotnet build dotnet-inspect.slnx -c Release -t:Rebuild   →   0 Error(s), 5 Warning(s)
```

All five are pre-existing and in projects **this PR does not touch** — `ILInspector.Metadata.Tests` (CS8601 ×2, CS8603, xUnit1025) and `DotnetInspector.CSharpBodySlicer.Tests` (xUnit2030). None is in this PR's 7 changed files, so nothing here regressed. But the claim as I wrote it was unearned.

This is the same failure mode as **finding 17** (the missing `--no-build`), and worth naming as such: *I stated a property my command had not actually measured.* An incremental build reports the warnings of what it rebuilt, not of the tree. Two for two now — which is the argument for the reviewer, not against it.

### Independently confirmed at the exact head

| Claim | Verified by |
| --- | --- |
| `AGENTS.md` is the exact main-parent blob | blob comparison; only the 11-line gate paragraph discarded |
| "Pack and publish flows" deletion came from main | confirmed independent of this branch; no main content lost |
| Pin discipline survived relocation | present and coherent, `docs/decompiler-correctness-pipeline.md:303-312` |
| Auto-merge coherent | no duplicate headings, no broken internal anchors |
| Gate not widened by main's new test classes | measured **16 methods / 4 classes**, no extras |
| No multi-case or non-plain facts entered gate classes | `GateExpectedClassesTests` 2 tests, 0 failures |
| `ci.yml` unaffected | main never touched it; branch blob retained exactly |
| Gate + checker | `16 total, 3 failed, 0 skipped, 0 not run`; checker `OK: the failing set matches the known-red list exactly` |
| markdownlint | exit 0 |

Also non-blocking: the reviewer started while two checks were still pending, against main's new *"do not start a round until the branch is settled"* rule. Those checks have since passed, and its findings were taken from the settled head, so the conclusion stands — but the rule is right and I should have gated the launch on it.

### Strongest evidence available

`decompiler-gates` **ran in real CI at this head and passed**, alongside all 7 other checks. The job is no longer only locally simulated. `mergeStateStatus: CLEAN`.

**Ready to merge.**

PR #3524, opened 2026-07-30
Move mechanism out of AGENTS.md to its owners; harden the review start gate

richlander:
## Conclusion

`AGENTS.md` goes from **575 lines / 30.8 KB to 439 lines / 22.8 KB** by moving
mechanism to the docs that own it, deleting one block that was already
duplicated, and tightening prose. Two rules change: the adversarial-review start
gate becomes a stop stated first, and the `gh` commands that prove a PR is green
are corrected.

Every agent reads this file before every task, so its length is a real cost, and
its own rule says where the detail belongs: *"Detailed design, subsystem
mechanics, version requirements, and historical context belong with their owning
code, workflow, or focused documentation."*

## Relocations — no rule changed

| Moved out of AGENTS.md | To | Note |
| --- | --- | --- |
| Decompiler `Speed`/`Area`/`--gate` traits | — | **Already documented** at `docs/decompiler-correctness-pipeline.md:137-215`. Deleted as duplication; one-line pointer replaces ~25 lines. |
| IR invariant essay (host contract, structural vs semantic levels, fixture rules) | `docs/decompiler-correctness-pipeline.md`, new section | The entry-gate checklist item 3 now links to it. |
| Packaging mechanics (`IsPackable`, `PackagingSurfaceTests`) | `docs/release-workflow.md` under Packages | |
| Stacked-PR mechanics (why restacking force-pushes and cascades, the `range-diff` proof, interaction with fixed-head review) | New `docs/stacked-prs.md` | AGENTS.md keeps nine binding rules + link. |
| "What to read" bullets in *Start here* | The *Task-specific guidance* table | One index instead of two. |

What AGENTS.md **keeps** from each move is the repository-wide rule the detail
illustrated — e.g. the IR invariant essay left behind "a correctness check must
not hide behind `[Conditional("DEBUG")]`, because the suite runs Release", and
the packaging block left behind "internal libraries are not an external
compatibility surface".

## Rule changes

**1. The review start gate is a stop, and it is stated first.** A round does not
begin until the PR is stable, CI-green, and free of merge conflicts — and for a
stacked PR, until *every layer* is. This was previously a paragraph in the
middle of the section; it now opens *Adversarial review* under its own heading
(*Do not start a round until the branch is settled*). It also adds "settled" to
the head requirement: finish your own edits before handing out a round, so the
branch is not moving while reviewers read it.

**2. Corrects the blocking finding from #3522's MAI-Code review, which never
landed before that PR merged.** `gh pr checks <n>` alone does not establish
green: it exits `8` while checks are pending, and reports `skipping` for
path-filtered jobs. The guidance now specifies:

- `gh pr view <n> --json mergeable,mergeStateStatus` for conflicts;
- `gh pr checks <n> --required` for the gating runs;
- explicitly, that pending is not green and a skipped path-filtered job is not a
  pass signal — wait for completion (`--watch`) rather than reading either as
  clear.

*Adversarial review* also gains three subheadings (start gate, tiers, running
the round) and the tier list becomes a table. The tiers, the roster, and the
single-vendor-harness rule are unchanged in substance.

## Compatibility / non-action boundary

- All three referenced anchors survive: `#adversarial-review` (inbound from
  `docs/decompiler-quality.md:329` and
  `docs/decompiler-correctness-pipeline.md:353`), `#building-and-testing`, and
  `#stacked-prs-for-multi-slice-issues`.
- No rule was dropped. Where a rule reads differently, it was rewritten, not
  removed — see the accounting below.
- `.claude/CLAUDE.md` still points at `AGENTS.md` as the single source of truth
  and needed no edit.
- Not attempted: further compression of *Evidence and validation*, *Harness
  boundary*, or the doc-index tables. Those are repository-wide rules or the
  index itself, which is the file's job.

## Validation

Documentation-only, making no measured behavior claim, so per `AGENTS.md` this
requires Markdown validation and not product builds or tests.

```console
$ npx markdownlint-cli AGENTS.md docs/stacked-prs.md \
    docs/decompiler-correctness-pipeline.md docs/release-workflow.md
$ echo $?
0
```

Link and anchor check across the four changed files (resolves every `](...)`
target and fragment against the headings that exist at this head):

```console
broken links: 0
```

Sentence-level accounting of everything removed from `AGENTS.md`, matched
against the three destination docs: 34 of 198 sentences unmatched, and each is a
deliberate rewrite of a retained rule (the tier bullets that became a table, the
precondition bullets that were reworded, the stacked-PR bullets that became
prose in `docs/stacked-prs.md`) — not a dropped rule. The reviewer prompt asks
for this to be re-derived independently.

## Review

High risk despite being documentation: this file governs how every agent works
in the repository, and the failure mode — a rule silently lost in a 384-line
diff — is invisible in a green build. Two reviews, two different models.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -6,16 +6,8 @@
 projects, platform libraries, metadata, APIs, dependencies, source provenance,
 analysis, Findings, implementation diffs, and decompilation.
 
-Read this file before doing work. Then read only the task-specific entry
-documents relevant to the change:
-
-- Read `README.md` when changing user-visible capabilities, commands, or
-  examples.
-- Read `docs/overview.md` when a change crosses subsystem ownership boundaries.
-- Read the relevant section of `docs/architecture.md` only when implementation
-  structure matters to the task.
-- Follow the task-specific entry points below and then only the links relevant
-  to the change.
+Read this file before doing work, then read only the entry documents below that
+are relevant to your change.
 
 This file is the source of truth for repository-wide engineering and workflow
 rules. Detailed design, subsystem mechanics, version requirements, and
@@ -57,18 +49,19 @@ those documents exist.
   task-relevant docs before continuing.
 - Do not mix unrelated changes into one commit or sweep another contributor's
   working-tree changes into your work.
-- Treat worktrees as temporary. For a PR requiring adversarial review, confirm
-  the exact reviewed head is pushed, then remove the development and review
-  worktrees with `git worktree remove <path>` as soon as every required
-  fixed-head review is clean. For a change that does not require adversarial review,
-  remove its development worktree after merge. Do not retain inactive
-  worktrees in case more work appears; recreate one for the branch if follow-up
-  work is needed.
+- Treat worktrees as temporary. Confirm the exact reviewed head is pushed, then
+  `git worktree remove <path>` as soon as every required fixed-head review is
+  clean — or after merge, for a change that needs no adversarial review. Do not
+  retain inactive worktrees in case more work appears; recreate one for the
+  branch if follow-up work is needed.
 
 ## Task-specific guidance
 
 | Area | Read first |
 | --- | --- |
+| User-visible capabilities, commands, or examples | `README.md` |
+| A change crossing subsystem ownership boundaries | `docs/overview.md` |
+| Implementation structure | the relevant section of `docs/architecture.md` |
 | Layering and consumer boundaries | `docs/design/inspection-layers.md` |
 | Command defaults and disclosure | `docs/design/progressive-disclosure.md` |
 | Output data shapes | `docs/design/output-shapes.md` |
@@ -86,6 +79,7 @@ those documents exist.
 | IL round-trip tests | `tests/DotnetInspector.ILRoundtrip.Tests/README.md` |
 | Decompiler behavior or harnesses | `docs/decompiler-correctness-pipeline.md` |
 | Skills | `taste/skill-guidance.md` |
+| Stacked PRs and restacking | `docs/stacked-prs.md` |
 | Release and publishing | `docs/release-workflow.md` |
 
 PR templates:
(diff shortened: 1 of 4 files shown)


richlander:
## Review reconciliation

Two-model tier (high risk: this file gates every other change in the repo). Reviewers: **MAI-Code** and **Gemini 3.1 Pro** — neither my own family. Four heads were reviewed; the last two are clean for both.

| Head | MAI-Code | Gemini 3.1 Pro |
| --- | --- | --- |
| `0ad80708a` | 1 non-blocking | 1 blocking, 1 non-blocking |
| `ab120283f` | clean | 1 non-blocking |
| `3b1c56353` | clean | **1 blocking** |
| `7410248d2` | **clean** | **clean** |

### Findings and disposition

**1. `skipping` semantics — blocking (Gemini), same text non-blocking (MAI). Fixed in `ab120283f`.**
The start gate said a `skipping` result "is not a pass signal" and to "wait for completion". Both halves were wrong: a path-filtered skip is terminal and never becomes a pass, so an agent would have rejected a green PR and waited for a transition that cannot happen. Verified empirically on this PR — `gh pr checks 3524` exits `0` with five `skipping` jobs present, and `--watch` returns immediately.

I did **not** adopt Gemini's suggested framing ("a skip *is* a valid pass signal") verbatim, and said so. The bullet keys off the exit code and keeps a distinction: a skip does not block, but it is not evidence either, so a skipped job is never proof a change was validated, and a job that skipped when it should have run makes the path filter the bug. Gemini re-reviewed the divergence and accepted it, noting it "correctly bounds the *evidentiary claim* rather than binding the *workflow state*".

**2. Duplicated "Stop stacking" bullet — non-blocking (Gemini). Fixed in `ab120283f`.** Confirmed at 2 occurrences; second removed. Gemini re-verified the other 8 moved bullets survived intact.

**3. Quick-read/tier loophole — non-blocking (Gemini). Fixed in `3b1c56353`.**
The new `### A quick read is not a round` section gates a *round*, but the tier table required a *review* — and a quick read is also MAI-Code. An agent could report "reviewed with MAI-Code" and an auditor reading the table would find its requirement met, silently banking a quick read as the tier's review. Adopted as suggested: the table now reads "a single review round" / "two review rounds".

**4. Start gate demanded CI a stacked slice never gets — blocking (Gemini). Fixed in `7410248d2`.**
This one was a defect I introduced in `3b1c56353` and Gemini caught the contradiction it left behind. `.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`, so a slice targeting its parent gets **no** CI runs at all — verified on the child PR #3525, where `gh run list --branch` is empty and `gh pr checks` prints "no checks reported" while the PR is MERGEABLE/CLEAN. The gate nonetheless required every layer to be "CI-green" and refused exceptions, so an agent would have held the round forever.

Fixed in guidance rather than by broadening CI to non-main PRs, per this file's own "do not broaden CI without a measured need"; Gemini was asked to push back on that tradeoff and endorsed it. I diverged from Gemini's suggested parenthetical carve-out in favour of a positive rule — "green on every check that runs for it" — and explicitly asked both reviewers to attack it as potentially more elastic. Both rejected the worry: "runs for it" is a server-determined property readable from `gh pr checks`, not an agent's judgement about relevance.

**5. Residual gating uses of "review" — non-blocking (Gemini). Fixed in `7410248d2`.** Four sites where "review" was doing gating work that "round" should do. Gemini re-verified the substitution was not over-applied — the general activity is still "review".

### Self-found, no reviewer credit

Finding 4's underlying CI-trigger gap was found by applying this PR's own stack-wide green rule to its child PR #3525 and discovering the rule was unsatisfiable. Worth recording: the rule failed the first time it was used.

### Non-actions

- CI is **not** broadened to non-main PRs. Upper slices remain unrun until they retarget `main`.
- `.claude-plugin/plugin.json` declares `0.7.2` against the csproj's `0.16.0`. Noticed, out of scope, left alone.

### Validation

markdownlint clean on all four changed files at every head; anchor check clean, including the three inbound `AGENTS.md#` links from `docs/stacked-prs.md`, `docs/decompiler-quality.md`, and `docs/decompiler-correctness-pipeline.md`. Docs-only with no measured behavior claim, so no product build or test. Gate at head: MERGEABLE/CLEAN, `gh pr checks` exit 0.

Ready to merge

richlander:
## Retracting `Ready to merge` — blocking defect found after the fact

Do not merge this yet. I found a false mechanical claim in my own text after posting readiness, on the same bullet that finding 4 added.

The bullet asserts:

> **An upper slice reports no checks at all, and that is not pending.** CI here runs only on PRs targeting `main` (`.github/workflows/ci.yml`), so a slice aimed at its parent gets no runs and `gh pr checks` says "no checks reported".

The child PR #3525 falsifies it. It targets `docs/agents-md-shrink`, not `main`, and it has a full passing CI run:

```
$ gh api repos/richlander/dotnet-inspect/actions/runs/30504539354 \
    --jq '{event, pull_requests: [.pull_requests[] | {number, base: .base.ref}]}'
{"event":"pull_request","pull_requests":[{"number":3525,"base":"docs/agents-md-shrink"}]}

$ gh run view 30504539354 --json workflowName --jq .workflowName
CI
```

So the CI workflow ran for a `pull_request` whose base is a parent slice, even though its trigger is `pull_request: branches: [main]`.

What makes this worse is that the original observation was real: earlier in this PR's life, `gh run list --branch docs/release-shipped-docs` was **empty** and `gh pr checks 3525` printed "no checks reported" while the PR was MERGEABLE/CLEAN. Both MAI-Code and I reproduced that independently. So the observable behavior changed between two heads of the same PR with the same base, and I generalized a mechanism from the first observation without establishing why it held.

Both reviewers verified the claim as I stated it and agreed with it — MAI-Code reproduced it on #3525 and Gemini read the trigger block. That is not their failure. I gave them a claim to check against a snapshot that happened to agree, and none of us tested the other direction.

### Fix

The operational advice is still right and is what an agent actually needs: absence of checks is terminal, not pending, so do not wait for runs that are not coming; and never report an unrun slice as green. That guidance is correct under **both** behaviors. The mechanism claim is what has to go — I will not assert a trigger rule I cannot reproduce on demand.

Fix and a fresh round on the new head to follow. `Ready to merge` above is withdrawn.

richlander:
## Round 5 — both reviewers clean at `0b3d8dce5`

The retraction above is resolved. Head `0b3d8dce5046043af0b34c183fc028aac2cf52ed`, MERGEABLE/CLEAN, `gh pr checks` exit 0.

**Fix.** The bullet no longer asserts any mechanism about when CI runs. It states only the observable behavior, which holds regardless of trigger semantics: no-checks-reported is terminal rather than pending; read what the PR actually reports instead of assuming in either direction; do not wait for runs that may never come; interim evidence is local validation plus the bottom slice's CI and must be stated; never report an unrun slice as green; re-check after retargeting `main`.

The gate sentence was left alone. "Green on every check that runs for it" never depended on the trigger behavior — as Gemini put it, if no checks run it is vacuously satisfied and the agent is still caught by the downstream bullet demanding local evidence. That is the argument for having written it as a positive rule rather than as the stacked-PR carve-out originally suggested.

**Both reviewers were asked to be harder this round**, specifically to re-test the *other* mechanical claims in the file rather than reason about them, since one we had all signed off on was wrong. Re-verified against live PRs:

| Claim | Result |
| --- | --- |
| `gh pr checks` exit `0` = nothing failed or outstanding | verified on #3524 and #3525 |
| exit `8` = checks still running | verified |
| `--required` reports none and exits non-zero here | verified, exit 1, both PRs |
| `skipping` is terminal | verified |

Gemini extended the sweep unprompted to other assertions in the file and confirmed them by execution rather than reading: `dotnet test src/dotnet-inspect.Tests` really does restore and run nothing (the reason this file mandates `dotnet run`), `IsPackable=false` in `Directory.Build.props` with `src/dotnet-inspect` overriding it, and the single `IrInvariants.DisableForShippedTool()` call site in `src/dotnet-inspect/Program.cs` with its host-contract guard. No further false or fragile claims found.

### What this PR ended up being

Five rounds, four blocking-or-equivalent findings, two of them in text I wrote to fix an earlier finding. Worth stating plainly in the record: the stack-wide green rule failed the first time it was used, and the mechanism I wrote to explain that failure was itself wrong. Both are now fixed, and the surviving guidance asserts only what is observable.

### Full finding ledger

| # | Finding | Raised by | Severity | Fixed in |
| 1 | `skipping` wrongly treated as non-passing, with an unactionable "wait" | Gemini blocking, MAI non-blocking | blocking | `ab120283f` |
| 2 | Duplicated "Stop stacking" bullet | Gemini | non-blocking | `ab120283f` |
| 3 | Quick read could be banked as the tier's review | Gemini | non-blocking | `3b1c56353` |
| 4 | Start gate demanded CI a stacked slice may never get | Gemini | blocking | `7410248d2` |
| 5 | Residual gating uses of "review" instead of "round" | Gemini | non-blocking | `7410248d2` |
| 6 | False CI-trigger mechanism claim | self, post-readiness | blocking | `0b3d8dce5` |

Non-actions unchanged: CI is not broadened; `.claude-plugin/plugin.json` version drift (`0.7.2` vs `0.16.0`) noticed and left alone.

Validation: markdownlint clean on all four changed files at every head; anchors clean including the three inbound `AGENTS.md#` links. Docs-only, no measured behavior claim, so no product build or test.

Ready to merge

PR #3525, opened 2026-07-30
Make README.md and SKILL.md a release prerequisite

richlander:
Both files ship, so a release publishes whatever they currently say, and neither
had a release-time checkpoint. `README.md` is the package readme
(`PackageReadmeFile` in `src/dotnet-inspect/dotnet-inspect.csproj`, packed by
the `None Include` entry beside it). Every `skills/**/SKILL.md` is an
`EmbeddedResource` in the tool binary, served by `dotnet-inspect skill`. A
version bump could therefore publish a readme describing commands the tool no
longer has, or ship a skill teaching agents a capability that moved.

This makes a `VersionPrefix` change a documentation checkpoint.

- `docs/release-workflow.md` gains a fourth prerequisite and a **Shipped
  documentation** section: why the two are release artifacts rather than
  repository notes, what to check in each, and that a skill's frontmatter
  `description:` is the single source of truth for the generated listing — so a
  stale description ships as a stale listing.
- The outcome is recorded either way. "Neither needed a change" is a valid
  answer; silence reads the same as an unchecked box.
- `AGENTS.md` carries the four-line rule and points at the checklist, extending
  the packaging paragraph rather than opening a new section.

## Stack

Slice 2 of 2. Parent: #3524 (`docs/agents-md-shrink`), which is where the
packaging paragraph this extends now lives; this PR targets that branch, so the
diff is only its own slice. Nothing remains after this — the release checklist
is the last piece of guidance this pass touches.

## Compatibility and non-action

Guidance only; no product code, no behavior claim, and no change to what the
package actually contains. It does not attempt to *enforce* the checkpoint in
CI — the drift it addresses is a judgement call about whether prose still
matches behavior, which a gate cannot make. Related drift noticed and
deliberately left alone: `.claude-plugin/plugin.json` declares `0.7.2` against
the csproj's `0.16.0`.

## Validation

`npx markdownlint-cli AGENTS.md docs/release-workflow.md` — clean.
Docs-only with no measured behavior claim, so per `AGENTS.md` no product build
or test is required.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -95,9 +95,12 @@ current product behavior and tests over design history. When current sources
 disagree, stop and resolve which owner is authoritative rather than silently
 choosing one.
 
-When adding a focused skill, register it in `SkillCommand.Skills`. Its YAML
-frontmatter `description:` is the single source of truth for the generated
-skill listing.
+When adding a focused skill, register it in `SkillCommand.Skills` **and** add an
+`EmbeddedResource` line for it in `src/dotnet-inspect/dotnet-inspect.csproj`;
+the embeds are enumerated per skill, and no test compares them against the
+`skills/` directory, so a skill missing from either list ships as nothing with a
+green suite. Its YAML frontmatter `description:` is the single source of truth
+for the generated skill listing.
 
 ## Repository-wide engineering constraints
 
@@ -203,9 +206,9 @@ public surface as an internal design constraint, not an external compatibility
 surface. `docs/release-workflow.md` owns the packaging mechanics.
 
 Changing `VersionPrefix` in `src/dotnet-inspect/dotnet-inspect.csproj` is a
-release, and `README.md` (packed as the package readme) and every
-`skills/**/SKILL.md` (embedded in the binary) ship with it. Consult both before
-the version moves and update whatever the release changed; the checklist is in
+release, and `README.md` (packed as the package readme) and the shipped
+`SKILL.md` files (embedded in the binary) ship with it. Consult both before the
+version moves and update whatever the release changed; the checklist is in
 `docs/release-workflow.md`.
 
 ### File-based apps
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -202,6 +202,12 @@ libraries carry no versioning story or API-stability commitment: treat their
 public surface as an internal design constraint, not an external compatibility
 surface. `docs/release-workflow.md` owns the packaging mechanics.
 
+Changing `VersionPrefix` in `src/dotnet-inspect/dotnet-inspect.csproj` is a
+release, and `README.md` (packed as the package readme) and every
+`skills/**/SKILL.md` (embedded in the binary) ship with it. Consult both before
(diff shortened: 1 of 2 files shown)


richlander:
## Review reconciliation — both reviewers clean at `83be037a8`

Head `83be037a841a3587503a33874ef5ceac04bae227`, MERGEABLE/CLEAN, `gh pr checks` exit 0.

**Escalated from the single-round tier to two models** (MAI-Code and Gemini 3.1 Pro), despite the slice's size. Reason: on the parent PR #3524 a mechanical claim that *both* reviewers had verified turned out to be false, and after MAI-Code's first clean pass here I added several more claims of exactly that class. Per `AGENTS.md`, when unsure of the tier, escalate.

Both reviewers were instructed to **falsify, not confirm**, and were authorized to probe empirically rather than reason from the source.

### Claims, and how each was settled

| Claim | Method | Result |
| --- | --- | --- |
| Embeds are enumerated per skill, not globbed | Gemini built the product with an untracked probe skill and ran `strings` on `dotnet-inspect.dll` | true — no trace of the probe; SDK default globs do not pick up `.md` |
| An unregistered skill is invisible to the suite and ships as nothing | **both** reviewers independently created `skills/zz-probe/SKILL.md` and ran the suite | true — `Total: 20, Errors: 0, Failed: 0` both times |
| Frontmatter `description:` is the single source of truth for the listing | read `SkillCommand.cs`; `SkillEntry` delegates to `ReadFrontmatterValue(...)`, no hardcoded descriptions | true |
| `README.md` is the packed package readme | csproj lines 33 and 64 | true |
| `VersionPrefix` change is the release trigger | `release.yml` greps it out of the csproj | true |

Both reviewers confirmed they removed their probes and left the worktree clean.

### Self-found before this round

My own first draft of this slice was wrong in two ways, corrected in `e5ea4cbad` before review: it described the embeds as a `skills/**/SKILL.md` glob (they are ten enumerated `EmbeddedResource` lines), and it framed the risk as a *stale* description shipping when the sharper risk is a *missing* skill shipping as nothing.

### Non-blocking, actioned as follow-up

MAI-Code: the new checklist is a manual process, not a repository-enforced gate. Correct, and it is the weaker form — this repository's own *Asserted properties name their gate* rule prefers making the declaration drive the enforcement set. Filed as **#3531** with the reproduction, the three sets that should be asserted equal, and the note that `PackagingSurfaceTests` already parses csproj files so the csproj half has precedent. Not fixed here: it is a product change and this slice is docs-only.

The documentation is explicit that it is standing in for a gate rather than being one, so nothing here asserts an enforced property that is not enforced.

### Stack

Slice 2 of 2. Parent #3524 is clean at `0b3d8dce5` and marked ready. This PR targets that branch; merge #3524 first, then confirm this retargets to `main` and still shows only its own diff. No residual after this.

### Validation

`npx markdownlint-cli AGENTS.md docs/release-workflow.md` clean; `#shipped-documentation` anchor resolves. Docs-only with no measured behavior claim, so no product build or test required — though both reviewers built and ran tests anyway as part of falsifying the claims above.

Ready to merge

Issue #3526, opened 2026-07-30
No .gitattributes: CRLF checkout breaks every raw-string test expectation on Windows


richlander:
## Summary

The repository has **no `.gitattributes`**. With Git for Windows'' default
`core.autocrlf=true`, every `.cs` file is checked out with CRLF line endings. Any
test whose expectation is a **raw string literal** therefore carries CRLF on
Windows and LF on Linux, while the product printers emit `\n`. Those tests pass on
Linux CI and fail on Windows, for every developer, forever.

This is the shared root cause behind #3496 and #3509 rather than a third instance
of the same bug.

## Observed

```text
ILInspector.Decompiler.Tests.StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString [FAIL]
  Assert.Equal() Failure: Strings differ
                       v (pos 10)
  Expected: "switch (s)\r\n{\r\n    case "a":\r\n        return 1;\r\n "...
  Actual:   "switch (s)\n{\n    case "a":\n        return 1;\n    c"...
                       ^ (pos 10)
```

Reproduced on a clean detached worktree at `origin/main` (15153c5f), so it is not
branch-specific:

```console
git worktree add --detach /tmp/ctrl origin/main
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- \
  -method ILInspector.Decompiler.Tests.StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString
# Total: 1, Failed: 1
```

## Mechanism

`StringSwitchRaisingTests.cs:43` normalizes only the **actual** side:

```csharp
var output = CSharpPrinter.Print(Raised(nameof(CfgSampleClass.SmallStringSwitch)))
    .Output!.ReplaceLineEndings("\n").Trim();

Assert.Equal(
    """
    switch (s)
    ...
```

`ReplaceLineEndings("\n")` correctly pins the actual side to LF. The expected side
is a raw string literal, whose line endings are whatever the file has **on disk**.
Without `.gitattributes` that is CRLF on Windows, so the two sides can never agree
there.

Note the direction is the opposite of #3509, where the expectation was an ordinary
`"\n"`-escaped literal (always LF) and the *product* emitted `Environment.NewLine`.
Both directions have the same underlying cause: line endings are not pinned
anywhere, so each test picks a side by accident.

## Scope

This is not one test. `src/ILInspector.Decompiler.Tests` alone has ~60 files
[...]

Issue #3514, opened 2026-07-29
Switch-expression default reachable from both a failing when-guard and a failing type test loses its return (CS0161)


richlander:
## Summary

When a switch-expression **default arm** is reachable from **two** edges at once — a failing `when` guard *and* a failing type test on a sibling arm — structuring duplicates the shared `return` onto the guard-failure edge but **not** onto the type-test-failure edge. The result is a body that falls off the end: invalid C# (**CS0161**).

This is a distinct trigger in the same family as #2959 / #2973 / #3017, all of which fixed *other* ways the shared return could go missing.

## Reproduction

`ILInspector.Decompiler.Pipeline.SwapIdiomPass::MatchCarrier` is a live in-repo case:

```csharp
static Carrier? MatchCarrier(IrFunction function, IrNode node) => node switch
{
    StoreStackSlot slot => new StackSlotCarrier(slot.Slot) { SavedValue = slot.Value },
    StoreLocal local when IsHiddenLocal(function, local.Index)
        => new HiddenLocalCarrier(local.Index) { SavedValue = local.Value },
    _ => null,
};
```

```bash
dotnet run --project tools/DecompilerHarness -c Release -- \
  artifacts/bin/ILInspector.Decompiler.Tests/release/ILInspector.Decompiler.dll \
  --dump "ILInspector.Decompiler.Pipeline.SwapIdiomPass::MatchCarrier"
```

Raised output (`fidelity Full`, stage 97/97):

```csharp
IrNode V_3 = node;
StoreStackSlot slot = V_3 as StoreStackSlot;
if (slot is null)
{
    if (V_3 is StoreLocal local)
    {
        if (IsHiddenLocal(function, local.Index))
        {
            return new HiddenLocalCarrier(local.Index) { SavedValue = local.Value };
        }
        return null;          // <-- guard-failure edge got the shared return
    }
                              // <-- type-test-failure edge did NOT; falls off the end
}
else
{
    return new StackSlotCarrier(slot.Slot) { SavedValue = slot.Value };
}
```

The inner `if (V_3 is StoreLocal local)` has no `else`, so the `if (slot is null)` block has a path that returns nothing.

## Why this is a clean natural experiment

`SwapIdiomPass` has three sibling matchers with the same expression-bodied `T? M(...) => x switch { ..., _ => null }` shape:

| Method | `when` guard? | Decompiles validly? |
| --- | --- | --- |
[...]

Issue #3503, opened 2026-07-29
Fidelity gate false positive: Contract V1 compares synthesized lambda-cache ordinals literally


richlander:
## Summary

`FidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket` reports
`StatementBodyLambdaInsideIf` as a new `OperandDiff` on current `main`
(`e1b3cac9`). **It is not a decompiler defect.** The recompiled IL is
behaviourally identical to the original; the gate is comparing
compiler-synthesized closure-cache *ordinals* literally.

Surfaced while triaging #3326, alongside #3490 and #3502 — but unlike those
two, this row should be fixed in the gate, not in the decompiler.

## The diff is ordinal-only

The opcode sequences are **byte-for-byte identical**:

```
original  : ldarg brfalse ldsfld dup brtrue pop ldsfld ldftn newobj dup stsfld ret ldnull ret
recompiled: ldarg brfalse ldsfld dup brtrue pop ldsfld ldftn newobj dup stsfld ret ldnull ret
```

The whole reported difference is three operands:

```
- <>c::<>9__103_0                          + <>c::<>9__128_0
- <>c::<StatementBodyLambdaInsideIf>b__103_0   + <>c::<StatementBodyLambdaInsideIf>b__128_0
- <>c::<>9__103_0  (stsfld)                + <>c::<>9__128_0  (stsfld)
```

Both names denote the same cache slot for the same lambda in the same closure
class. `ldsfld`/`ldftn`/`stsfld` all move together, so there is no
mismatched-slot possibility — only a renumbering.

## Why the number differs

`<>9__N_M` is Roslyn's lambda cache field: `N` is the containing method's
**declaration ordinal within the compilation unit**, `M` the lambda index
within that method. The ordinal is a property of the *unit being compiled*, not
of the method.

Confirmed against the real assembly with a metadata probe:

- `CfgSampleClass` declares **687** methods.
- `StatementBodyLambdaInsideIf` sits at metadata method-table index **97**.
- The synthesized member in the shipped assembly is
  `<>c::<StatementBodyLambdaInsideIf>b__103_0` — ordinal **103**.

So the ordinal is neither the metadata index nor anything the decompiler emits;
it is assigned by Roslyn from declaration order. The compile-back path
recompiles a **reconstructed skeleton**, whose member ordering and member count
are the harness's, not the original source file's — so Roslyn assigns **128**
for the same method. Nothing about the decompiled C# changed.

[...]

Issue #3502, opened 2026-07-29
Fidelity gate: ManualConstantOnlyComparisonFactory has four spurious spill ops (second unfiled row)


richlander:
## Summary

`FidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket` reports **three**
unexpected rows on current `main` (`e1b3cac9`, i.e. with #3495 merged).
#3490 covers `MakeConsumerWithTwoLeadingArgs`. This issue covers the second:
`ManualConstantOnlyComparisonFactory`.

Surfaced while triaging #3326. Like the other rows, it is pre-existing — it was
invisible only because this gate was timing out at 45 minutes before #3495.

## The diff

```
original  : ldtoken call ldstr call stloc ldc.i4 box ldtoken call call ldc.i4 box ldtoken call call call ldc.i4 newarr dup ldc.i4 ldloc stelem.ref call ret
recompiled: ldtoken call ldstr call stloc ldc.i4 box ldtoken call call ldc.i4 box ldtoken call call call stloc ldc.i4 newarr dup ldc.i4 ldloc stelem.ref stloc ldloc ldloc call ret
```

```
Added IL operation 'stloc'
Added IL operation 'stloc'
Added IL operation 'ldloc'
Added IL operation 'ldloc'
```

Four spurious spill operations: the recompiled body stores two intermediate
values to locals and reloads them, where the original keeps them on the stack.

This is the same family as #3490 (which adds a single spurious `stloc`/
`ldloc` pair), so the two may share a root cause in how the printer
materializes an expression that the original evaluated purely on the stack.
Both are expression-tree factory shapes here: the diff appears around the
`newarr`/`stelem.ref` argument-array construction, just before the final
`call`.

## Reproduction

```
dotnet build dotnet-inspect.slnx -c Release
dotnet run --project src\ILInspector.Decompiler.Tests -c Release --no-build -- \
  -class "ILInspector.Decompiler.Tests.FidelityGateTests"
```

`ManualConstantOnlyComparisonFactory` appears in the unexpected list of
`NoNewFidelityDiffsBeyondKnownDocket`.

Observed on Windows, but not Windows-specific: this gate is `Speed=Slow` and
excluded from the PR lane (`ci.yml:301`), so it does not run on Linux PR CI
either.

## Not to be confused with the third row

The third unexpected row, `StatementBodyLambdaInsideIf`, is a **different
kind of thing** — an operand-only diff in compiler-synthesized names that
[...]

PR #3528, opened 2026-07-30
Restore the decompiler gates to green: fix tuple spelling and out/in refKind, docket the rest

### LINKED ISSUE #3489: Declaration formatter escapes unnamed keyword-typed tuple elements: (int, int) renders as (@int, @int), emitting non-compiling C#

richlander:
- Fixes #3489, #3490, #3491, #3492, #3493
- Changes `dotnet-inspect member` to spell unnamed tuple types compilably, and fixes an `out`/`in` refKind misclassification; restores both fidelity gates to green
- Card revision: `7ff3f5f0`

## Change

**Conclusion:** **PASS** — two real product defects are fixed with focused regression tests; the remaining rows are docket omissions, each dated to the commit that introduced it and matched to an existing docket class.

These five issues are the entire red surface of the decompiler gates, and `release.yml:97` runs `--gate no-corpus`, so **releases cannot ship until they are green**. They accumulated unseen because these gates are `Speed=Slow`: the PR lane skips them (`ci.yml:301`), and the weekly Deep Inspect lane that did run them was being *cancelled* at its 6h timeout — a cancelled job does not satisfy `if: failure()`, so its notifier never fired (#3432). #3495 cut that lane from 56m to 22m, which is what finally made these visible.

Two are genuine product bugs; three are bookkeeping. They are separated below so the bookkeeping does not launder the bugs.

### Product defect 1 — unnamed tuple types were spelled uncompilably (#3489)

`CSharpDeclarationWriter.EscapeParameterLists` escapes keyword-named parameters. It found the parameter list by scanning for the **first** `(` in the signature. A C# tuple type is parenthesized too, so a tuple-typed *return* was consumed as if it were the parameter list, and each element''s trailing token was run through `EscapeIdentifier`.

Benchmark target: `ILInspector.Decompiler.Tests.dll` (in-repo fixture assembly `CfgSampleClass`).

```bash
dotnet-inspect member ILInspector.Decompiler.Tests.CfgSampleClass TupleRest --library artifacts/bin/ILInspector.Decompiler.Tests/release/ILInspector.Decompiler.Tests.dll -v:d
```

#### Before

```csharp
public static (@int, @int, @int, @int, @int, @int, @int, @int) TupleRest(int a)
```

- Valid: **False** — `@int` is an identifier, not the predefined type, so this is CS0246
- Correct: False (does not bind)
- IL fidelity: False — `FidelityGateTests.PinnedFixesStayExact` reported `TupleRest` as `RecompileFail`
- Taste applied: None (this is an escaping defect, not a style lens)
- Commit: `15153c5f`

#### After

```csharp
public static (int, int, int, int, int, int, int, int) TupleRest(int a)
```

- Valid: True
- Correct: True
- IL fidelity: True — `PinnedFixesStayExact` is green again
- Taste applied: None
- Commit: `7ff3f5f0`

**This is broader than the fixture that caught it.** A *named* element hid the bug: in `(int Sum, int Product)` the trailing token is the element name, which needs no escape. Only unnamed elements — where the trailing token *is* the keyword — were corrupted, which is why exactly one pinned member regressed rather than all tuple-returning members. Also affected, and now covered by tests:

| Shape | Before | After |
| --- | --- | --- |
| Unnamed tuple return | `(@int, @string) Pair(int a)` | `(int, string) Pair(int a)` |
| Nested tuple | `((int, int), @int) Nested(int a)` | `((int, int), int) Nested(int a)` |
| Constraint tuple | `where T : IComparable<(@int, @int)>` | `where T : IComparable<(int, int)>` |

A parameter list''s `(` directly follows the member name, a generic `>`, an operator token, or a close paren. A tuple type''s `(` starts the string or follows whitespace or one of `< ( , [`. The new `OpensParameterList` helper encodes exactly that, and non-parameter-list parens are now *scanned past* rather than skipped wholesale, so a tuple return still gets its real parameter list escaped.

### Product defect 2 — `[In, Out]` parameters were spelled `out` (#3493)

The repository had **four** independent implementations of "is this C# `out`?". Three test `Out && !In`; `ApiSurfaceExtractor.cs:1718` tested `Out` alone.

| Site | Logic | Correct |
| --- | --- | --- |
| `MethodDefinitionFacts.cs:145` | `Out && !In` | yes |
| `IrImporter.cs:2498` | `Out && !In` | yes |
| `MetadataDeclarationQuery.cs:686` | `isOut && !isIn` | yes |
| `ApiSurfaceExtractor.cs:1718` | `Out` only | **no** |

An interop-marshalled `[In, Out] ref int` sets **both** flags. The buggy site therefore spelled it `out int value`, and the body''s `value++` then failed definite assignment with CS0269. `IrImporter.cs:2491` already carries a comment noting that interop-marshalled `ref` parameters set the In flag; that knowledge just had not reached the fourth copy. Fixed by adding the missing `!isIn` guard.

### Bookkeeping 1 — two stale taste pins (#3493, remaining rows)

Two `GeneratedFixtures` source fragments were authored in #2279 (2026-07-03) and pinned spellings that the project has since *deliberately* changed:

- expected `new StringBuilder()`, emits `StringBuilder V_0 = new();` — target-typed `new()` landed in #3059 (2026-07-23) and is the recorded taste choice (`docs/decompiler-taste.md:122`, #3072)
- expected `(object)left == (object)right`, emits `!=` with inverted branches — the reference-null-test spelling is recorded at `docs/decompiler-taste.md:82`

Both are `rts=Exact` **spelling** mismatches, not correctness regressions. The fragments are updated with rationale comments pointing at the taste record. Diagnosing this required dumping the emitted bodies by hand, so that dump is now productized as bounded harness reporting (`TruncateBody`, 2000-char cap) — reporting only.

### Bookkeeping 2 — CS0161 census gained one row (#3492)

`SwapIdiomPass::MatchCarrier` joined the census. It is an **addition, not a regression**: `SwapIdiomPass` landed in #3234, *after* the census was last refreshed in #3003.

The defect is real and precisely characterized: the switch-expression default (`_ => null`) is reachable from **two** edges — a failing `when` guard and a failing type test — but #2959''s return-duplication covers only the guard edge, so the block falls off the end. Clean natural experiment: sibling matchers `MatchPlace` (:148) and `MatchStore` (:157) have identical shape but no `when` guard, and decompile validly. Filed as **#3514** and pinned with that analysis. The census is an exact-set `Assert.Equal`, so #3514''s fix cannot land silently — deleting the row will fail this test until the pin is updated.

### Bookkeeping 3 — docket omissions on both fidelity rails (#3490, #3491)

Nine rows, zero regressions. Each was dated to its introducing commit and matched against an existing docket class before being added:

| Row | Rail | Why it was missing |
| --- | --- | --- |
| `MakeConsumerWithTwoLeadingArgs` | sugared | #3272 added the fixture; #3290 docketed it in the **lowered** gate only. The sugared rail produces the identical residual — confirmed by diffing the two rails'' opcode streams, which differ only by the same inserted `stloc`/`ldloc`. |
| `ManualConstantOnlyComparisonFactory` | both | Arrived in #3053, *after* its five `ManualConstantOnly*` siblings were docketed as a group, so it inherited the group''s compile-back temp reshaping without inheriting its entry. #3502. |
| `StatementBodyLambdaInsideIf` | sugared | Cached-lambda ordinal moved `<>9__103_0` → `<>9__128_0` because fixtures were added after #2987 placed it; opcodes identical. Same class as `TwoLocalFunctionQuadrants`. #3503. |
| `EnumArgInInlineLocalFunction`, `EnumArgInLocalFunctionWithLocal`, `RecursiveLocalFunction`, `StaticLocalFunctionWithLocal`, `TwoLocalFunctionQuadrants` | lowered | Synthesized local-function ordinals. **All five are already docketed in the sugared gate with per-fixture rationale.** De-sugaring does not renumber local functions, so the class applies identically to both rails; the lowered docket had simply fallen behind. |

The lowered gate also now **reports what the sugared gate reports**. It printed only method names, so triaging #3491 required re-running an 8-minute gate with an ad-hoc dump just to see which opcodes moved. It now emits the original and recompiled opcode streams, the detail string, and up to six fidelity rows. Reporting only — no comparison, normalization, or fallback logic changed, so the gate fails on exactly the same set as before, per the AGENTS.md harness boundary.

## Evidence

| Check | Baseline (`15153c5f`) | Head (`7ff3f5f0`) |
| --- | --- | --- |
| `FidelityGateTests` | 5 total, **2 failed** | 5 total, 0 failed |
| `LoweredFidelityGateTests` | 4 total, **1 failed** | 4 total, 0 failed |
| `GeneratedFixtureCatalogTests` | 18 total, **3 failed** | 18 total, 0 failed |
| `ValidityCoverageReportingTests` | 3 total, **1 failed** | 3 total, 0 failed |
| The four classes above, together | **7 failed** | **30 total, 0 failed** (1126s) |
| `CSharpDeclarationWriterTests` | 46 total, 0 failed | 55 total, 0 failed (+9 new) |
| **`--gate no-corpus`** (the gate `release.yml:97` runs) | 8 failed | **4244 total, 1 failed**, 7 skipped |

The single remaining `--gate no-corpus` failure is `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString`, a **pre-existing Windows-only CRLF failure that this PR does not touch**. Rather than assume, it was reproduced on a clean detached worktree at `origin/main`:

```console
git worktree add --detach /tmp/ctrl origin/main   # 15153c5f
  -method ILInspector.Decompiler.Tests.StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString
# Total: 1, Errors: 0, Failed: 1
```

Running it down found the shared root cause of that whole family: the repository has **no `.gitattributes`**, so with Git for Windows'' default `core.autocrlf=true` every `.cs` is checked out CRLF and raw-string test expectations carry CRLF while the printers emit `\n`. Linux CI never sees it. Filed as **#3526** with the one-line fix; deliberately **not** bundled here, since renormalizing the working tree is a large, content-free diff that deserves its own PR.

The 7 skips are `ilasm not available` on this machine, which those tests skip on by design.

## Compatibility

`EscapeParameterLists` changes user-visible declaration text, so the boundary is worth stating precisely: it only ever *stops* adding `@` to tokens inside parenthesized groups that are not parameter lists. Every previously-correct escape is preserved — operators (`operator +(`), constructors, generic methods (`M<T>(`), explicit-interface implementations, and `where T : new()` all still get parameter escaping, each covered by a case in the new theory. No output that was previously valid becomes invalid; output that was previously CS0246 now binds.

## Non-actions

- **#3514, #3502, #3503 are filed, not fixed.** They are pinned or docketed with the analysis needed to fix them. Fixing them here would mix four unrelated decompiler defects into a gate-restoration PR.
- **The `Speed=Slow` PR lane is not wired up here.** Path-gating decompiler PRs into a lane that runs these gates is the obvious follow-up, but it must land *after* this PR — wiring it first would turn every decompiler PR red.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- --gate no-corpus
  -class ILInspector.Decompiler.Tests.FidelityGateTests \
  -class ILInspector.Decompiler.Tests.LoweredFidelityGateTests \
  -class ILInspector.Decompiler.Tests.ValidityCoverageReportingTests \
  -class ILInspector.Decompiler.Tests.GeneratedFixtureCatalogTests
dotnet run --project src/ILInspector.CSharp.Tests -c Release -- \
  -class ILInspector.CSharp.Tests.CSharpDeclarationWriterTests
```


richlander:
## Adversarial review round 1 — head `7ff3f5f0`

Two independent reviewers, two model families, each in its own detached review worktree at the exact head.

| --- | --- | --- |
| A | MAI-Code | BLOCK |
| B | GPT-5.6 | BLOCK |

Four findings between them. Three were accepted and fixed in `9c784bf8`; one was filed as out of scope. **Both reviewers independently found the same blocking issue**, which is the strongest signal in the round.

### 1. Whitespace before a parameter list dropped the escape — BLOCKING, accepted, fixed

Both reviewers found it. `void M (int event)` rendered `int event` unescaped, because `OpensParameterList` rejected every `(` preceded by whitespace.

Reproduced at head:

```text
void M (int event)   ->  void M (int event)      # bug
void M(int event)    ->  void M(int @event)      # correct
```

**Both reviewers'' suggested fix was rejected, and I verified the rejection rather than asserting it.** Both proposed skipping backwards over the whitespace and accepting a preceding identifier character. Simulating that rule:

```text
public static (int, int) Pair(int a)     first '(' at 14  ->  rule says parameterList=True
```

That is #3489 exactly: the character before the tuple is the `c` of `static`, so the tuple return is taken for the parameter list and mangled back to `(@int, @int)`. The whitespace is not an oversight in the guard — it is the only string-level signal separating a tuple return from a parameter list.

Resolved by the **token** instead of the character (`9c784bf8`). A tuple return follows the modifier/return-type run, so the token before it is a C# keyword; a parameter list follows the member name, which never is — a keyword-named member is already escaped to `@name` before this runs. The rule asks `CSharpKeywords.RequiresDeclarationEscape`, the existing product answer to "is this token a keyword", rather than introducing a second keyword list.

This satisfies reviewer B''s actual suggestion ("locate the parameter list relative to the known member name/operator structure, allowing intervening whitespace") without the mangling.

### 2. Tuple-returning conversion operators lost their operator syntax — accepted, fixed

Reviewer B found a **third instance of the very defect this PR exists to fix**, which neither I nor reviewer A had spotted. `FormatOperatorSignature` also took the first `(`:

```csharp
var parenStart = signature.IndexOf('(');
if (parenStart <= 0)
    return signature;      // tuple return => index 0 => bail out
```

For `(int a, int b) op_Implicit(Foo f)` that bailed out and left the raw `op_Implicit` spelling instead of `implicit operator (int a, int b)(Foo f)`. Fixed in `9c784bf8` by locating the member name first and taking the parameter list after it, with four new test cases (two tuple returns, two non-tuple returns to pin that ordinary operators are unchanged).

### 3. Interop `[In] ref` rendered as C# `in` — accepted as real, filed as #3537, out of scope

Reviewer B is correct: raw `ParameterAttributes.In` does not imply C# `in`; it is also the interop `[In]` marshalling attribute on a mutable `ref`. `IConnectionPointContainer.FindConnectionPoint([In] ref Guid riid, …)` renders as `in System.Guid riid`.

This is the mirror image of the `out` half that **is** fixed here, but it is **pre-existing and unchanged by this PR** — the `!isIn` guard only affects parameters with the `Out` flag set. Folding a fourth unrelated defect into a gate-restoration PR would work against getting releases unblocked, so it is filed rather than fixed.

### 4. "Fidelity dockets are one-way allow lists" — acknowledged, no action

Reviewer B observed that removing a genuine known diff would not fail the docket tests. True, and pre-existing. Worth noting that the CS0161 census in this PR does **not** have that weakness — it is an exact-set `Assert.Equal`, so #3514''s fix cannot land silently.

### What each reviewer verified clean

Reviewer B independently confirmed the parts of this PR most at risk of laundering a regression:

- docket chronology (each added row dated to its introducing commit)
- the taste records behind the two updated `GeneratedFixtures` fragments
- the exact-set census
- that the harness/reporting changes are reporting-only
- fidelity gates 5/5 and 4/4, generated fixtures 18/18, validity census 3/3, tuple metadata 22/22

Reviewer A confirmed the four `out`/`in` sites and found no fifth copy, and ran the requested temporary-revert proof: reverting the product hunk turns the new tuple tests red, so they are genuinely load-bearing.

## Scale evidence added in response

Because the escaper changes user-visible output product-wide, I measured base vs head across a real assembly rather than relying on fixtures. Using the product''s own `EscapeParameterName`/`SplitTopLevel` to reconstruct the base algorithm faithfully:

| System.Private.CoreLib (38,507 members) | Result |
| --- | ---: |
| Signatures changed by this PR | **1** |
| Still mangled at head | 0 |
| Keyword parameters still correctly escaped | 138 |
| Parameter lists preceded by whitespace | 0 |

The one change is the real-world canary:

```text
base: (@uint, @int) GetScalar(ref char source, int index, int length)   # CS0246
head: (uint, int) GetScalar(ref char source, int index, int length)
```

The 207 signatures that do contain `" ("` are all default-value casts *inside* a parameter list (`NumberStyles style = (NumberStyles)231`), which the guard correctly declines to treat as a new list. **These counts are identical before and after the whitespace refinement**, which is the evidence that the refinement is a robustness fix and does not move real product output.

## Validation at the fixed head `9c784bf8`

| Check | Result |
| --- | --- |
| `CSharpDeclarationWriterTests` | 60/60 pass (was 55; +4 operator, +1 whitespace) |
| `ILInspector.CSharp.Tests` | 359 total, 14 failed — all `CSharpTypePrinterTests` CRLF (#3526), unchanged |
| `ILInspector.Metadata.Tests` | 998/998 pass |
| `dotnet-inspect.Tests` | 2502 total, 2 failed — both reproduce on clean `origin/main` |
| Fidelity gates + fixtures + census | **30/30 pass** |

The 2 `dotnet-inspect.Tests` failures (`Layout_Count_CountsFilesRatherThanRenderedTreeLines`, `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown`) were confirmed pre-existing by running them on a clean detached worktree at `origin/main` (`ce20e3d1`), not assumed.

Round 2 review of `9c784bf8` is running with both models. Not marking ready until both are clean.

richlander:
## Round-2 reconciliation — both blocks fixed at `5d3e2efc`

Both reviewers were right, and both blocks were in code I added in round 1. Thanks — round 1's fix traded one narrow bug for a broader one, and neither of you let it through.

### The common cause

MAI's findings 1–2 and GPT's finding 1 are four faces of one mistake. I classified a `(` from its **preceding** context, which forces me to tell a modifier keyword apart from a member name. That is unbounded, and each of you found a different way past it:

| Counterexample | Why the keyword rule failed | Found by |
| --- | --- | --- |
| `void M<T> (int event)` | `>` is not an identifier part, so the empty token fell through the `start > end` guard and returned "keyword" | MAI |
| `partial (int, int) M(int event)` | `partial` is not in `RequiresDeclarationEscape`, so the tuple return was read as a parameter list | MAI |
| `void @event (int class)` | the extractor strips `@` and tests `event`, which *is* a keyword | GPT |
| `void event (int class)` | same, via the unescaped spelling | GPT |

Patching the keyword table would have left a fifth. Instead the rule now keys on what **follows** the closing paren: a parameter list *ends* the declaration (optionally followed by `where` constraints), whereas a parenthesized type is followed by the member name it types. That is the one structural difference between the two, so `OpensParameterList` needs no keyword table at all and `PrecedingTokenIsKeyword` is **deleted** — as is this PR's dependency on `CSharpKeywords` from that call path.

```csharp
static bool OpensParameterList(string signature, int open)
{
    int close = Matching(signature, open, '(', ')');
    if (close < 0)
        return false;

    var trailing = signature.AsSpan(close + 1).TrimStart();
    return trailing.IsEmpty || StartsWithConstraintClause(trailing);
}
```

All four counterexamples are now regression tests in `MemberDeclaration_ClassifiesParenGroupsByTrailingContext` (renamed from `…ByPrecedingToken`, since the name asserted the rule that was wrong).

### GPT finding 2 — accepted, and worse than it looked

The interim `LastIndexOf` selects a **parameter type** named `op_Implicit` over the member itself, finds no following `(`, and skips operator formatting entirely. Confirmed against your compiled fixture. The member is now located as the whole identifier token whose next non-whitespace character is `(`.

Worth recording how this was reachable at all: I scanned all **732** `op_*` members in CoreLib and the interim rule and the `main` rule pick the **same index every time** — `interim-vs-main index diffs: 0`. The defect never fires on real metadata. Only a compiled fixture could have caught it, which is exactly what you used. My CoreLib scan would never have found it, and I over-trusted that scan in round 1.

### GPT finding 3 / MAI finding 3 — non-reproduction resolved

GPT reproduced my scan exactly (38,507 members, 1 delta, 0 mangled) on .NET 9. MAI saw 51,547 members and 0 changes, which is a different runtime's CoreLib — the delta is a single .NET 9 member (`GetScalar`), so a different version legitimately shows zero. Not a discrepancy in the finding.

**Correction to my round-1 numbers:** GPT is right that the escaped keyword/contextual parameter count is **146**, not 138. 138 is `@object` alone; I quoted the largest bucket as the total. The PR body is corrected.

### Evidence at `5d3e2efc`

| Check | Result |
| --- | --- |
| `CSharpDeclarationWriterTests` | **63/63** — includes all four counterexamples + 3 `op_*`-named-type cases |
| CoreLib delta probe (9.0.17, 38,507 members) | **1 changed, 0 mangled** — byte-identical to the `9c784bf8` result GPT reproduced |
| Operator probe (732 `op_*` members) | **0 left unformatted** |
| `ILInspector.Metadata.Tests` | **998/998** |
| `ILInspector.CSharp.Tests` | 14 failures, all pre-existing CRLF (#3526), count unchanged |
| **`--gate no-corpus`** (the release gate, `release.yml:97`) | **4244 total, 1 failed** — the #3526 CRLF failure, reproduced on clean `origin/main` |

The CoreLib probe result is the load-bearing one: the redesign is **behavior-identical at scale** to the version GPT already verified, while fixing all four counterexamples. It changes which signatures are *classified*, not which are *rewritten*.

### Process note

While re-running the gates I found that one `-class` name in my own command (`GeneratedFixtureFidelityTests`) does not exist, and xUnit **silently ran zero tests** for it rather than erroring. That is the same hazard `AGENTS.md` calls out for `dotnet test`. It did not affect any result reported above — the release-gate run supersedes the per-class runs — but it is worth a follow-up so a typo'd filter cannot read as a pass.

### Not fixed here (filed)

- #3537 — `[In] ref` should render as `in`. Pre-existing, independent of this PR; GPT confirmed this PR does not change interop behavior.
- #3526 — missing `.gitattributes` + `core.autocrlf=true` is the root cause of the entire Windows CRLF failure family.

Ready for round 3 at `5d3e2efc`.

richlander:
## Evidence correction and strengthening (self-reported, before round 3 reports)

While round-3 review runs I widened my own scale check from CoreLib to the **whole shared framework**, and it both strengthens the PR and corrects two things I said earlier. Flagging it myself rather than letting a reviewer find it.

### 1. The bug was 33× more widespread than my CoreLib-only scan showed

Sweeping all 169 assemblies in `Microsoft.NETCore.App 9.0.17` (**122,084 members**):

| | |
| --- | --- |
| base-vs-head differences | **33** |
| still mangled at head | **0** |

Every one of the 33 is a correct tuple-mangling fix. Beyond the `GetScalar` case I already reported:

```
base: (@string, @string) GetDestinationAndLinkPaths(string destinationDirectoryPath)
head: (string, string)   GetDestinationAndLinkPaths(string destinationDirectoryPath)

base: System.Threading.Tasks.ValueTask<(@int, @int, @int, @bool)> ParseTextAsync(int outOrChars)
head: System.Threading.Tasks.ValueTask<(int, int, int, bool)>     ParseTextAsync(int outOrChars)
```

Note the `Task<(...)>` cases: the mangled tuple is **nested inside a generic return type**, which my earlier `(int, string) Pair(...)` framing did not cover. My CoreLib-only number of "1 changed" understated the blast radius because CoreLib happens to contain one instance.

### 2. Correction: my scale evidence does *not* exercise the escaping path

`head APPLIED an escape: 0` across all 122,084 members. `EscapeParameterLists` never fires on any framework assembly — C# source cannot declare an unescaped keyword-named parameter, so there is nothing to escape. The path only matters for IL-authored, obfuscated, or non-C#-compiled input, which is precisely the decompiler's domain but is not represented in the framework.

So the honest reading of the scale evidence is narrower than I implied in the previous comment: it is strong proof of **no over-escaping and no mangling** (122,084 members, 0 mangled), and proves nothing about the escaping path itself. That path is covered only by the unit tests in `MemberDeclaration_ClassifiesParenGroupsByTrailingContext`. Reviewers should weight it accordingly.

### 3. Correction: withdrawing the "138 / 146 escaped parameters" figure

Both numbers were mis-attributed. Those `@` tokens came from other escape paths in `RenderMemberDeclaration` (`EscapeMemberNameInSignature`, `EscapeQualifiedKeywordSegments`), not from `EscapeParameterLists`. The correct figure for the path this PR changes is **0** on the framework, per above. I should not have quoted it as evidence for this change in round 1 or round 2.

### 4. Two outliers that happen to validate the rule

The sweep surfaced the only two signatures in CoreLib that stress the new rule's edges, and both land correctly:

- `(object? Target, object? Dependent) TargetAndDependent { get; }` — a **property** whose type is a tuple. There is no parameter list at all, and the trailing-context rule correctly finds none. A "first paren wins" rule would treat the tuple as a parameter list.
- `int AddIdToGuid(System.Guid* outPtr, int whereToAddId, uint id, bool overflow = false)` — contains the substring `where` as part of a **parameter name**. This is why `StartsWithConstraintClause` requires `where` to be followed by whitespace or end-of-span rather than doing a bare `Contains("where")`; the naive check would have been wrong on a real framework member.

Probe: `$env:TEMP\probe_sweep.cs`, run as `dotnet run probe_sweep.cs "C:\Program Files\dotnet\shared\Microsoft.NETCore.App\9.0.17"`. It reconstructs the base algorithm through the product's own `EscapeParameterName`/`SplitTopLevel`/`Matching` via reflection, so only the changed decision differs.

Head under review is unchanged at `5d3e2efc` — this comment adds evidence and corrects claims, no code change.

richlander:
## Review reconciliation — six rounds, two reviewers

Both reviewers (MAI-Code and GPT-5.6-Sol) reviewed this PR independently in isolated worktrees at fixed heads. This comment attributes every finding, states what was verified or dismissed, and links resolution.

This PR is in the **high-risk** tier: subtle correctness risk in a string-scanning path that runs over every rendered member declaration. Two reviewers from two different families, per `AGENTS.md`.

### Findings adopted

| # | Round | Reporter | Finding | Resolution |
|---|---|---|---|---|
| 1 | 1 | both | Whitespace-insensitive paren classification | `9c784bf8` (superseded) |
| 2 | 2 | both | Four counterexamples to the keyword heuristic | **Redesigned** — trailing-context rule, keyword table deleted (`5d3e2efc`) |
| 3 | 2 | GPT | `LastIndexOf` picks a parameter type named `op_Implicit` over the member | `TryFindMemberNameBeforeParameterList` (`5d3e2efc`) |
| 4 | 3 | both, independently | `Matching` is literal-blind: `void M(int event = ")")` | `SkipLiteral`, used by both `Matching` and `SplitTopLevel` (`13d83434`) |
| 5 | 4 | GPT | A member *named* `where` defeats constraint detection | Require the full `where T :` shape (`6de20bc8`) |
| 6 | 4 | GPT | My own `SkipLiteral` **regressed verbatim strings** (`@"\"`) | Verbatim support (`e541c951`) |
| 7 | 4 | GPT | `IsIdentifierPart` rejects Unicode combining marks | Made constraint detection delimiter-based, so it never classifies the name (`e541c951`) |
| 8 | 5 | GPT | `$@"` worked but `@$"` did not | Scan the whole `@`/`$` prefix run; also modelled interpolation holes (`38f7d40f`) |
| 9 | 5 | GPT | Unbounded mutual recursion → stack overflow (~4,800 levels), reported **non-blocking** | Fixed anyway — capped at 32 (`49b9e6c1`) |
| 10 | 5 | MAI | Raw string literals | `SkipRawLiteral` (`49b9e6c1`) |
| 11 | 6 | GPT **and** my own probe, independently | `opening >= 3` routed **verbatim** strings into the raw scanner: `@""""` swallowed the rest of the signature. Reported **non-blocking** | Fixed anyway — `!verbatim` (`2cba9cfc`) |
| 12 | 7 | GPT | **BLOCK**: the fidelity dockets are one-directional allow lists; 25 stale rows silently permit future regressions | Reproduced, filed **#3584**, accepted non-blocking on attribution |

Findings 9 and 11 were both reported as non-blocking and both fixed regardless. Finding 9 because a stack overflow is uncatchable process death rather than a wrong answer, and the reachability argument that made it non-blocking is the one GPT correctly dismantled in round 5. Finding 11 because the fix is a single term in an existing condition, in code that already claimed to handle verbatim strings — and because I had hit it independently, which is a strong signal it is not merely theoretical.

**Finding 12 was the one genuine BLOCK of the final round, and it is a real defect.** `FidelityGateTests.cs:540` and `LoweredFidelityGateTests.cs:268` assert only that every *actual* diff appears in the docket; `known \ actual` is unconstrained. A stale row is therefore not untidiness but a blanket exemption — if `CollectionListLiteral` regresses tomorrow, the gate will not report it, which is exactly the "keep failure visible" rule inverted.

I reproduced it independently rather than taking it on report:

```
RAISED   known=83 actual=73 stale=10 missing=0
LOWERED  known=60 actual=45 stale=15 missing=0
*** rows added by THIS PR that are stale: NONE
```

Attribution settled it: this PR adds three raised and six lowered rows, **all live**, removes none, and both rails report `missing = 0`. The 25 stale rows predate the branch. Retiring them is not a one-liner either — flipping to a set-equality assert fails unless 25 rows are deleted in the same change, and each needs its own justification, since "the raise improved" and "the fixture moved so the diff relocated" both present as `Exact` and only the first warrants deletion. Filed as **#3584**; GPT reviewed the attribution and accepted it as non-blocking.

### Findings declined, with rationale

**MAI, round 3 — `i > 0` guard in `TryFindMemberNameBeforeParameterList`.** Requires an operator with no return type. Not producible: `returnType` is never empty, and `void` renders as `"void"`. Worth recording that `main` emitted `implicit operator (int value)` here — success-shaped garbage. Leaving the raw `op_*` spelling keeps the failure visible, per `AGENTS.md`.

**GPT, round 5 — comments in signatures.** Declined on a rationale that does *not* depend on reachability, since reachability proves too much here. `EscapeParameterLists` runs over a **rendered declaration**, not over C# source. Every literal form above is a legal spelling of a *parameter default* — data that flows from metadata into the rendered text. A comment is not: it is source syntax with no representation in a declaration fragment.

### Corrections I made against my own evidence

Recorded because the reviewers could not have caught either of these:

1. **Withdrawn: "138/146 escaped parameters".** Mis-attributed. Widening the probe from CoreLib to the whole shared framework showed the true figures are 33 tuple fixes and **0 escapes applied**. The scale evidence proves *non-mangling*; it does **not** show the escaping path works. Corrected on the PR at the time.
2. **A test row that is documentary, not evidential.** MAI's reported raw-string input `"""x)"""` already round-tripped *before* the fix — three quotes coincidentally pair up as three degenerate literals. It is kept and labelled as such in-source; the two adjacent rows are the ones that fail without `SkipRawLiteral`.
3. **A wrong prediction of mine, corrected against me rather than against the product.** I expected the closing quote of `""""a""", )""""` at index 13; it is at 14. The product was right. Recorded because this is exactly where an author can quietly "fix" correct code to match a wrong expectation.

I also discovered mid-review that a typo'd `-class` filter **silently runs zero tests and exits 0**, which had made one of my own earlier "passing" claims vacuous. Filed as #3546.

### The structural point the reviewers proved

Every one of findings 3–10 is the same shape: the writer **re-lexes a rendered declaration string** to recover information the extractor held as structured data moments earlier. Tuple mangling → operator name → literals → `where` → Unicode → prefix spellings → interpolation holes → raw strings. Each fix narrows the gap; none closes it.

Filed as **#3561**: escape from `ApiMember.SignatureModel` during signature construction, where a parameter name is unambiguously a parameter name. That deletes paren classification and literal scanning entirely. Out of scope here because this PR unblocks the release gates and #3561 touches the extractor/writer boundary.

### Evidence at the final head (`df180752`)

- **`--gate no-corpus` — 4291 tests, 1 failure**, that one being the known pre-existing Windows CRLF failure `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString` (#3526). This is the release-blocking check from `release.yml:97` and the reason this PR exists. **No docket row went stale across the merge of `main`.**
- `CSharpDeclarationWriterTests` — 89/89 (was 83 at branch point)
- `ILInspector.Metadata.Tests` — 998/998
- Direct `SkipLiteral` edge-case probe — 15/15, covering both interpolated-verbatim prefix orders declining the raw path, interpolated raw taking it, six-quote verbatim, unterminated forms, and the empty string
- Whole-framework sweep — 169 assemblies / 122,084 members / 33 differences / **0 mangled** / 0 escapes applied
- Non-vacuity verified for **every** fix by stashing only the product file and rebuilding

`origin/main` moved 16 commits during review, including two decompiler raising changes; it is merged in at `df180752` and the gate above was run *after* that merge.

### An anomaly reported against myself

The gate run at `2cba9cfc` failed **8**, not 1: the CRLF failure plus seven `MemberBodyProducerUnionTests`. Those seven passed 23/23 in isolation against the byte-identical build and did not reproduce at `df180752`.

Two variables changed between the runs — the 16 merged commits and the xUnit seed — so I make **no claim** that the merge fixed them, and I did not run a control at `origin/main` with the original seed. Filed as **#3582** with both runs recorded.

Why I did not treat it as blocking: this branch changes only `EscapeParameterLists` and its string-scanning helpers, which are pure static functions over strings with no shared mutable state, and so are an implausible source of cross-test interference. That is an argument rather than evidence, and both reviewers were asked to push back if they considered the control run necessary before merge.

### Issues filed rather than fixed here

#3489–#3493 (the regressions this PR closes), #3496, #3504, #3512, #3514, #3526, #3537, #3546, #3561, #3582, #3584.

### Verdicts

Seven rounds, two reviewers, every round on a fixed exact head in an isolated worktree.

| Head | MAI-Code | GPT-5.6-Sol |
| --- | --- | --- |
| `7ff3f5f0` | BLOCK | BLOCK |
| `9c784bf8` | BLOCK | BLOCK |
| `5d3e2efc` | BLOCK | BLOCK |
| `13d83434` | — | BLOCK |
| `e541c951` | BLOCK | BLOCK |
| `38f7d40f` | CLEAN | CLEAN |
| `49b9e6c1` | CLEAN | CLEAN (1 non-blocking, fixed anyway) |
| `2cba9cfc` | CLEAN | CLEAN |
| **`df180752`** | **CLEAN** | **CLEAN** |

richlander:
## Addendum: pin retirement, and CI green

Two heads were added after the reconciliation above.

### `df180752` — `origin/main` merged in

`main` moved 16 commits during review, two of them decompiler raising changes (#3553, #3459). Since this PR pins docket rows, a stale pin was a real risk that only a post-merge gate run could rule out. Merged (not rebased — the branch is public). Result: **no docket row went stale across the merge**.

### `e9725626` — three known-red pins retired

PR CI's `decompiler-gates` job failed with **all 16 of its tests passing**:

```
Decompiler gate: 16 tests, 16 passed, 0 failed, 4/4 expected classes covered, 3 pinned known-red.

STALE PINS (3) — pinned known-red but passing:
  FidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket
  FidelityGateTests.PinnedFixesStayExact
  LoweredFidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket

  Remove these from eng/decompiler-gate-known-red.txt. A pin that outlives its
  failure silently un-gates the test it names.
```

That is the ratchet working exactly as designed. Those three pins were added on 2026-07-29 to keep the lane attributable while #3489/#3490/#3491 were open; this PR fixes them, so the pins had to go. **`eng/decompiler-gate-known-red.txt` now has no entries** — the header records what was retired and why, so the file still reads as a ledger. `docs/decompiler-correctness-pipeline.md` asserted "The gate runs **red**", which is no longer true, and was corrected.

Both reviewers confirmed independently that `eng/check-decompiler-gate.cs` tolerates an empty pin list (only the expected-classes inventory is a hard failure) rather than taking my reading of it.

### This strengthens #3584 rather than closing it

The repository **already implements** the bidirectional ratchet GPT argued for — at the CI-lane level, where it caught a stale pin on this very branch within minutes of the head being pushed. #3584 is the same principle missing one level down, inside `FidelityGateTests.KnownDiffs`, where 25 stale rows currently sit unrejectable. The precedent is an argument for fixing it, and I have noted that on the issue.

### Final state

| Check | Result |
| --- | --- |
| `decompiler-gates` | **pass** (was the failing lane) |
| `test (ubuntu-26.04, linux-x64)` | pass |
| `build-net10` | pass |
| `markdownlint` | pass |
| `changes` | pass |
| Local `--gate no-corpus` | 4291 tests, 1 failure — the known Windows-only CRLF issue (#3526) |
| `CSharpDeclarationWriterTests` | 89/89 |
| Reviewer verdicts at `e9725626` | MAI-Code **CLEAN**, GPT-5.6-Sol **CLEAN** |

Ready to merge

Issue #3494, opened 2026-07-29
Make JSON a first-class output Format over the L2 section/shape model (per #3439/#3453)


richlander:
## Summary

Make **JSON a first-class output Format** — an L3 renderer over the L2
section/shape model — instead of the ~77 top-level source-generated serializers
that each re-freeze a fixed document shape. This is the L2/L3 target that the L1
typing conversion (opened by #3453, framed by #3439
[`docs/design/inspection-layers.md`](docs/design/inspection-layers.md)) is
heading toward. There is no rush: the re-architecture has not reached this seam
yet, and this issue exists so that when it does, JSON is already specified as a
Format that walks the shape ladder rather than a parallel emit path.

No user-visible flag or section name changes. This is about *where the JSON
shape decision lives*, not what it renders.

## The one defect, stated in layer terms

`output-shapes.md` already settles the model: output narrows through the **shape
ladder** Document → Table → Vector → Scalar, `--fields`/`--columns`/`--count`
are **Shape selectors**, and `--json`/`--tsv`/`--jsonl` are **Formats**,
"orthogonal to the shape ladder." `inspection-layers.md` assigns Shape/Section
to L2 and Format to L3, and states plainly: *"JSON is a format, not a shape."*

Every other Format is already an L3 renderer over the L2 section model, so it
honors the L2 knobs (`MarkoutWriterOptions.Projection`, `TableMode`, section
selection). **JSON is the exception.** It is emitted by top-level source-gen
serializers that bind a fixed shape at compile time:

- `Output/OutputFormatter.cs:179` — `return JsonSerializer.Serialize(result, JsonContext.Default.InspectionResult)` returns **before** any section/shape pass runs.
- 36 output-side `JsonSerializer.Serialize(...)` call sites across **77** `[JsonSerializable]` contexts (vs. 1 late-bound `Utf8JsonWriter` path, in a cache).
- `find`/search JSON is a second, hand-rolled emitter (`Output/FindResultWriters.cs`).

Because the JSON Format does not walk the ladder, every symptom we have been
chasing is the *same* mislayer:

| symptom | in layer terms |
| --- | --- |
| `--fields`/`--columns` silently dropped under `--json` (now fail-closed: #3462, #3472) | Shape selector not honored by the JSON Format |
[...]

Issue #3487, opened 2026-07-29
Section order should be a Markout writer option, not a post-hoc Markdown rewrite


richlander:
## Summary

`MarkdownSectionOrderer` should not exist. Section order is an **emit-time**
concern, and the tool currently expresses it by re-parsing its own already-
rendered Markdown and shuffling `## ` blocks as text. The real gap is in
[Markout](https://www.nuget.org/packages/Markout) (pinned at `0.29.0` in
`Directory.Packages.props`): `MarkoutWriterOptions` can say *which* sections to
render and *which* columns to project, but not *in what order* to render them.

Follow-up to #3455 / #3454, which fixed a Windows CRLF corruption bug in this
pass. That fix was correct for the bug, but it is upkeep on a component whose
existence is the actual defect.

## Current shape

`src/dotnet-inspect/Output/MarkdownSectionOrderer.cs` takes rendered Markdown
plus an ordered list of section *names* and rewrites the document. Every call
site sits immediately after serialization, so the tool renders and then
immediately re-reads its own output:

| Call site | Order source |
| --- | --- |
| `Output/OutputFormatter.cs:190` | `pipeline.GetAllSelectorSections(result)` |
| `Output/OutputFormatter.cs:192` | `pipeline.InfoSectionNames` |
| `Output/OutputFormatter.cs:341` | `pipeline.AlphabeticalSectionOrder` |
| `Output/OutputFormatter.cs:411` | `pipeline.AlphabeticalSectionOrder` |
| `Output/OutputFormatter.cs:434` | `pipeline.AlphabeticalSectionOrder` |
| `Commands/ApiCommand.cs:1007`, `:1012` | `ApiMemberSectionPipelines` |

In every case the ordering is known **before** a single character is written —
it comes from the section pipeline, which is the same place `IncludeSections`
comes from.

## Why the post-hoc pass is the wrong layer

1. **It is a Markdown parser on the rendering path.** It hand-tracks code
   fences (`MarkdownScan.IsCodeFence`) so a `## ` inside a fenced block is not
   mistaken for a heading. Every Markdown construct it does not model —
   indented code blocks, HTML blocks, setext headings, front matter — is a
   latent misparse in shipped output.
2. **It uses display text as an identity.** Sections are matched by heading
   text under `OrdinalIgnoreCase`. `AGENTS.md` is explicit that identifiers and
   presentation are separate concerns and that one must not be inferred from
   display text when a typed identity exists — and here it does, in the section
[...]

PR #3472, opened 2026-07-29
Fail closed on find --fields/--columns under --json (#3386 follow-up)

richlander:
## Summary

`find --fields`/`--columns` combined with `--json` **silently dropped** the column filter, emitting the full per-result JSON object with exit 0. This is the `find` analog of the two silent-projection-drop bugs fixed for the `type`/`api` surface in #3386 (merged as #3462). This PR applies the same **fail-closed** resolution to `find`: the unsupported combination is now rejected cleanly with exit 1 and a clear message.

## Why fail closed

`--fields`/`--columns` select **table columns**. `find --json` is an object/document format whose keys (`type`, `library`, `source_version`) differ from the table column vocabulary (`Type`, `Library`, `Sim`, `Match`, `Pattern`), and it has no jq-style column-slicing facility. The combination is a category error, so silently ignoring the filter hides user intent. Rejecting is consistent with the command-agnostic rule already documented in `docs/design/output-shapes.md` (no doc change needed).

## What still works (unchanged)

- Row-oriented formats still honor projection: `--jsonl`, `--tsv`, `--table`.
- `--count` reduces to a scalar and composes with `--json` → excluded from the guard.
- `-D`/discovery is dispatched **before** the guard, so discovery requests carrying `--fields`/`--json` are honored, not rejected.
- Both type-search and member-search (`--members`) paths flow through the single guard.

## Validation

- Full CLI suite green: 2405 tests, 0 failed (14 ilasm-dependent tests skip when ildasm/ilasm absent).
- 6 new tests in `CommandExecutionTests.cs` covering: `--fields`/`--columns` rejection (Theory), member-search rejection, `--jsonl` honored, `--count`+`--json` honored, discovery honored.
- Real-run probes confirm: rejected combos exit 1 with empty stdout + clear error; plain `--json`, `--jsonl`/`--tsv`/`--table` projection, `--count`+`--json`, and discovery all exit 0 with expected output.

## Adversarial review

Localized output-path rejection mirroring the already-reviewed #3386 pattern → one review (MAI-Code) per policy. Review was clean with real-run verification of the rejection, honored formats, discovery, and member-search paths.

Refs #3386.

PR #3536, opened 2026-07-30
Lower find --json to the projected view instead of rejecting it

richlander:
Lowers `find --json` to the projected view when `--fields`/`--columns` is named, instead of rejecting the combination. First slice of #3494.

## The claim

`--json --columns Type,Library` now returns the projected rows. Previously it exited 1 with "cannot be combined with --json".

The rule this slice establishes, resolved on #3494 before implementation:

| Invocation | View |
| --- | --- |
| `--json` | **pre-lowered** typed document — unchanged |
| `--json` + `--fields`/`--columns` | **lowered** display view, rendered through the Markout formatter seam |

`--fields`/`--columns` name post-lowering vocabulary. Computed columns such as `Return Type` have no counterpart in the typed object model, so naming one is definitionally an opt-in to the display view. The evidence is `CallGraphSectionAdapter.Label():82`, which already takes `requestedFields` as an argument *to the lowering function* and uses it to decide which typed facts get stringized into a display label.

Section selection (`-S`) and `--rows` stay on the pre-lowered side: they subset or truncate, they do not change vocabulary.

## Demo

Name the lowered columns you want and keep valid JSON output:

```console
$ dotnet-inspect find JsonSerializer --json --columns Type,Library --compact
{"results":[{"type":"JsonSerializer","library":"System.Text.Json"}]}
```

Before this PR, that command exited with the `--columns`/`--json` incompatibility error. Omitting `--columns` still returns the existing typed JSON array.
## How

`JsonSectionFormatter` is an `IMarkoutFormatter` — the same seam Markdown, TSV, and JSONL each render through, following the existing `RenderManifestFormatter` precedent in this repo. Markout applies `IncludeSections` and `Projection` *before* invoking a formatter, so JSON inherits projection rather than reimplementing it. That is the whole mechanism, and it is why Format-invariance holds by construction instead of by parallel maintenance.

An unmatched column fails closed exactly as it does under `--tsv` (`Error: No columns matched projection: Bogus`, exit 1) because the same Markout throw propagates — no new error convention.

## Compatibility and boundary

- Plain `--json` is untouched; the ~77 source-generated serializers still own it.
- Scope is `find` only. `type`/`member` still reject (#3386/#3472) and `package`/`library` still silently drop (#3487). Those are slices 2 and 3; the residual is deliberate, not overlooked.
- No behavior on `main` changes for any invocation that previously succeeded.

## Review

Two-model adversarial review (high risk: user-facing output contract plus new product code), **seven rounds on fixed heads**, GPT-5.6 Sol and Gemini 3.1 Pro in isolated worktrees. Nine real defects — seven from the reviewers, two from the author's own re-reading. Every fix carries a gate that was tamper-tested: revert the fix, watch the test fail.

| Finding | Reviewer | Resolution |
| --- | --- | --- |
| `--rows` silently dropped (`--rows 3` returned all 179 rows; `--tsv` returned 3) | both, independently | resolved over buffered rows via `RowWindow.Resolve` — the type's documented single interpretation point — not the line-counting `LimitRenderedTableRows`. Parity verified for head, tail, `N..M`, `N+K`, `N..`, on both search paths. |
| Key casing diverged: `--json --columns` emitted `Type` where `--jsonl` and plain `--json` emit `type` | Gemini | request Markout's JSONL vocabulary; section/field keys through `SnakeCaseLower`, the policy the pre-lowered serializers already declare |
| `--compact` ignored | GPT | threaded |
| Mixed content in one section silently destroyed (last-writer-wins `Kind`) | Gemini | `Section.Adopt`/`SetHeaders` throw and name the section; same-kind appends stay lossless, with a negative test |
| Duplicate JSON keys from the casing fix itself (`Call Graph` vs `CallGraph`) | Gemini | `RequireUniqueKey` throws; `Utf8JsonWriter` does not reject repeated names, so the parser would have kept one silently |
| `--columns Type,Type` emits the property twice per row | both, independently | rejected upstream in `BuildProjection` — see below |
| That rejection stack-traced on `package` | both, independently | moved to **parse time** — see below |
| A gate test that could not fail | author self-review | retargeted so tampering breaks it |
| A doc comment asserting something untrue | author self-review | audited the callers, corrected the claim |

The last one is worth stating precisely, because both reviewers proposed the wrong repair. They suggested guarding the JSON formatter. Checked against `main` first: `--jsonl --columns Type,Type` **already** emits `{"type":...,"type":...}` on untouched code. The lowered JSON was agreeing with JSONL exactly as intended, so guarding only JSON would have made `--json` reject a request `--jsonl` accepts — breaking the invariant this PR exists to establish. A duplicated column is a malformed request in any format; what differs is only whether the format keys its output, and so whether the loss is visible. It is now rejected once in `BuildProjection`, before any renderer runs, and all four formats fail closed identically. This changes `--tsv`/`--table` from silently-redundant to rejected, which is a deliberate and stated scope expansion.

### The `package` crash, and why the fix is not where it was reported

Rejecting a duplicate in `BuildProjection` threw from inside the System.CommandLine invocation pipeline. SCL reports such a throw cleanly only for commands that happen to catch it — `find`, `type`, `member`, and `library` have a catch-all; `package` catches only `HttpRequestException`. So `package Foo --fields Authors,Authors` printed a stack trace.

Both reviewers proposed adding a `catch` to `PackageCommand`. That fixes the command that was reported. The check now runs at **parse time**, on the `--columns`/`--fields` options themselves, which is better on two counts: it covers every command accepting those options rather than the one that happened to be probed, and it rejects before invocation, so nothing depends on a per-command catch broad enough to also swallow genuine bugs. This is the treatment `--rows` already gets in the same file, for the reason its own comment states.

`RejectDuplicates` stays as a second gate. Its comment now says accurately that it is unreachable from the CLI — every product caller passes values sourced from the validated options — and names the parse-time validator as the enforcing gate.

### Two defects the reviewers did not find

Both rounds 5 and 6 came back clean from both models. Re-reading the round-5 commit turned up two problems in it:

- **A test that could not fail.** The spelling-coverage Theory asserted through `find`, whose catch-all reports the *second* gate's throw with the same `Error: Duplicate ...` text as the parse-time gate. It passed with the validator's splitter replaced by a plain `Split(',')` — it was pinning the message, not the gate. Retargeted at `package`, which has no catch-all, so an escaped spelling surfaces as the stack trace the fix exists to remove. Re-tampering now fails three cases.
- **A doc comment claiming more than was true.** It said `BuildProjection` has callers not coming through the validated options. Audited: every product caller does; only `FindCommandTests` constructs one directly. Corrected, per the repo rule that an asserted property names its gate or is marked unverified.

Splitting on `';'` turned out to be load-bearing rather than cosmetic: `ArgumentPreprocessor.MergeRepeatedListOption` collapses repeated occurrences into one **`';'`-joined** token before parsing, so `--columns Type --columns Type` reaches the validator as `Type;Type`. A comma-only split would miss every duplicate spread across repeated occurrences — the form a scripted caller produces. Verified the merge round-trips correctly too: `--columns "Type,Kind" --columns "Namespace"` renders all three, identically to the `Type,Kind;Namespace` spelling.

The `package` gates name a package that does not exist, so their network independence is structural rather than incidental — measured 53ms with a duplicate versus 749ms and a `Package ... not found` without one.

Two reported items were **dismissed with evidence** rather than fixed:

- **`-n` corrupting multi-line JSON.** Pre-existing and not this change's. `Program.cs:139` installs a global `LineLimitingTextWriter` — a literal `head -n` over the byte stream. `type CommandExecutionTests --library <dll> --json -n 4`, on code this PR does not touch, truncates identically. Both reviewers accepted this on verification.
- **Backtick escaping.** `Utf8JsonWriter` emits `\u0060` where Markout emits a literal backtick. That is encoding, not content, and matching `Utf8JsonWriter` is what keeps `--json` consistent with its own pre-lowered output, which already escapes that way. `--jsonl` is the outlier. The parity gate compares decoded pairs for exactly this reason.

## Validation

- `dotnet run --project src/dotnet-inspect.Tests -c Release` — **2536 total, 1 failed**, 14 skipped.
- The one failure is pre-existing and environmental: `Layout_Count_CountsFilesRatherThanRenderedTreeLines` asserts Ordinal `Newtonsoft.Json.nuspec` while the local NuGet cache stores `newtonsoft.json.nuspec` lowercased. Its file count is correct, and it touches no code in this change.
- AOT publish (`-r linux-x64 -p:PublishAot=true`) — zero trim/reflection warnings; native binary verified for projection, `--rows`, `--compact`, and fail-closed.
- `markdownlint` clean on the changed doc.

The load-bearing gate is `Find_ProjectedJsonRows_CarryTheSameContentAsJsonl`: every row of the projected `--json` array must carry the same keys, in the same order, with the same values as `--jsonl` for the same query. That is the Format-invariance claim stated as a test, and it is what caught the casing drift.

Refs #3494. Related: #3386, #3472, #3487.


richlander:

Seven rounds on fixed heads, two models in isolated read-only worktrees — GPT-5.6 Sol and Gemini 3.1 Pro. High-risk tier: this changes a user-facing output contract and adds new product code.

Nine defects. Seven came from the reviewers, two from re-reading my own fix after both models had already returned clean.

| # | Finding | Found by | Head | Resolution |
| 1 | `--rows` silently dropped under projected JSON — `--rows 3` returned all 179 rows where `--tsv` returned 3 | both, independently | `a477e1e5d` | `38cc1258c` — resolved over buffered rows via `RowWindow.Resolve`, the type's documented single interpretation point, not the line-counting `LimitRenderedTableRows` |
| 2 | `--compact` ignored | GPT | `a477e1e5d` | `38cc1258c` |
| 3 | Key casing diverged — `--json --columns` emitted `Type` where `--jsonl` and plain `--json` emit `type` | Gemini | `a477e1e5d` | `38cc1258c` |
| 4 | Mixed content kinds in one section silently destroyed by last-writer-wins | Gemini | `a477e1e5d` | `38cc1258c` — `Adopt`/`SetHeaders` throw and name the section |
| 5 | The casing fix itself created duplicate-key risk (`Call Graph` and `CallGraph` both → `call_graph`) | Gemini | `38cc1258c` | `2a0302275` — `RequireUniqueKey` |
| 6 | `--columns Type,Type` emitted the property twice per row | both, independently | `2a0302275` | `7bb3c3134` — rejected upstream, not in the formatter |
| 7 | That rejection stack-traced on `package` | both, independently | `7bb3c3134` | `b85558489` — moved to parse time |
| 8 | A gate test that could not fail | **self-review** | `b85558489` | `a298174a6` |
| 9 | A doc comment asserting something untrue | **self-review** | `b85558489` | `a298174a6` |

Every fix carries a gate, and every gate was tamper-tested: revert the fix, watch the test fail. Finding 8 exists precisely because one of them did not.

### Two findings where the reviewers' diagnosis was right and their proposed repair was not

**#6 — both models proposed guarding the JSON formatter.** Checked `main` first: `--jsonl --columns Type,Type` **already** emits `{"type":…,"type":…}` on untouched code. The lowered JSON was agreeing with JSONL exactly as designed, so guarding only JSON would have made `--json` reject a request `--jsonl` accepts — breaking the invariance this PR exists to establish. Fixed once in `BuildProjection`, before any renderer runs, so all four formats fail closed identically. That makes `--tsv`/`--table` stricter than before; deliberate and stated.

**#7 — both models proposed adding a `catch` to `PackageCommand`.** That repairs the command that happened to be probed. The check now runs at parse time on the `--columns`/`--fields` options themselves, which covers every command accepting them and rejects before invocation — so nothing depends on a per-command catch broad enough to also swallow genuine bugs. This is the treatment `--rows` already gets in the same file, for the reason its own comment states.

Worth noting that #7 was **pre-existing for the same path**: `package Foo --columns NotAColumn --table` already stack-traced from Markout's `No columns matched projection` on untouched `main`. GPT surfaced that. My change added a second door into it, so it was in scope; the unmatched-column case is now fixed by the same parse-time move for duplicates, but the unmatched-column crash itself remains pre-existing and out of scope.

### Two items dismissed with evidence rather than fixed

- **`-n` corrupting multi-line JSON.** `Program.cs:139` installs a global `LineLimitingTextWriter` — a literal `head -n` over stdout. `type CommandExecutionTests --library <dll> --json -n 4`, on code this PR does not touch, truncates identically. Both reviewers accepted this on verification.
- **Backtick escaping.** `Utf8JsonWriter` emits `\u0060` where Markout emits a literal backtick. That is encoding, not content, and matching `Utf8JsonWriter` keeps `--json` consistent with its own pre-lowered output, which already escapes that way. `--jsonl` is the outlier. The parity gate compares decoded pairs for exactly this reason.

### One reviewer observation that turned out to matter more than reported

Gemini noted at `a298174a6` that the merge joins with `';'`, not `','` as my comment claimed, and judged it "an invisible implementation detail." The first half was right and is corrected. The second half is not: because `ArgumentPreprocessor.MergeRepeatedListOption` joins with `';'`, a validator splitting only on `','` would miss every duplicate spread across repeated occurrences — `--columns Type --columns Type` arrives as `Type;Type`. It is invisible for the happy path and load-bearing for the duplicate check. Tampering the splitter to `Split(',')` fails three tests, including that case.

### Validation at `cf07d4131`

- `dotnet run --project src/dotnet-inspect.Tests -c Release` — 2536 total, 1 failed, 14 skipped.
- The one failure is pre-existing and environmental: `Layout_Count_CountsFilesRatherThanRenderedTreeLines` asserts Ordinal `Newtonsoft.Json.nuspec` while the local NuGet cache stores it lowercased. Its file count is correct and it touches no code here. It passes in CI.
- AOT publish (`-r linux-x64 -p:PublishAot=true`) — zero IL2xxx/IL3xxx warnings; native binary verified for projection, `--rows`, `--compact`, and fail-closed rejection.
- CI green on this exact head.
- Both models CLEAN on the final head.

richlander:
Ready to merge

Both reviewers CLEAN on the final head `cf07d4131`, CI green on that exact SHA, mergeable state CLEAN.

Follow-up slices, non-blocking: `api`/`type`/`member` (#3386, #3472) need section-scoped `--columns` validation because those documents are multi-section and Markout applies projection globally; `package`/`library` (#3487) still silently drop projection under `--json`. `package Foo --columns NotAColumn --table` stack-tracing from Markout is pre-existing on `main` and untouched here.

richlander:
## Final review reconciliation

Reviewed fixed head: `2ed2c937557886c4ad468354300c9e392400d94b` against `151d91022449aac6ec08d6be9dd35f97e1a98dae`, using GPT-5.6 Sol and Claude Opus 5 in isolated worktrees.

- GPT repeated the zero-result JSON concern. Dismissed: empty stdout plus the no-results diagnostic is the established behavior of plain `--json`, projected JSON, JSONL, TSV, and table output; changing only projected JSON would break compatibility and format parity.
- Claude found no shipped-path blocker. Its non-blocking follow-up is that default Markdown/PlainText can still visibly repeat columns from overlapping wildcard patterns, while keyed/tabular projected formats reject them. That behavior is pre-existing, loses no data, and is outside this PR's keyed/tabular validity boundary.
- The prior blocking findings are resolved: resolved wildcard overlap now fails cleanly across projected JSON/JSONL/TSV/table; multi-assembly tabular rendering publishes atomically; and Windows tail buffering preserves CRLF without producing `\r\r\n`.

After the clean round, `main` moved from `151d9102` to `4279d718` by one commit that only renamed the progressive member-call-graph query/session API and updated its tests/docs. It touches none of this PR's 11 files and does not alter Markout, projection, output, CLI error handling, or line limiting. Per the user's decision, merge commit `b5526ef9` integrates that non-interacting range and carries the clean reviews forward.

Final-head evidence at `b5526ef98ccdcac718b96132ec891298ba6b3d01`: Release solution build passed; 204 formatter tests passed with one unavailable-ilasm skip; 20 projected-JSON/projection/multi-TFM tests passed; exact-head required CI passed on Windows and Linux in run https://github.com/richlander/dotnet-inspect/actions/runs/31337339156.

Ready to merge

richlander:
Correction: the Ready to merge status above is retracted. While final-head CI was running, main advanced to 4be245cc (#3950), which rewrites Markout row-window/section-ordering paths and directly overlaps CommandExecutionTests.cs, FindCommand.cs, OutputFormatter.cs, and progressive-disclosure.md. GitHub now reports the PR as conflicting. This range is interacting, so it must be integrated, revalidated, and reviewed at the resulting head before the PR can be ready.

richlander:
Final readiness reconciliation for `d46257af261bca449cddc63bc891103def7039f6`:

- Integrated current `main` through `c5e53f6c`. The post-review range contained the annotated-source prototype and call-graph edge-row projection. It did not touch this PR's product code or tests; the only shared file was `docs/design/output-shapes.md`, in separate non-overlapping sections. Per the maintainer's decision, the clean reviews were carried forward across that non-interacting merge.
- GPT-5.6 Sol: CLEAN at reviewed head `ef99e244`.
- Claude Opus 5: CLEAN at reviewed head `ef99e244`. Its two observations were non-blocking and outside the shipped `find` path: unnamed array composition in shared formatter infrastructure is currently unreachable, and an already-overbroad design-doc statement describes unrelated commands.
- Current behavior adopts Markout 0.35 row windows and wildcard deduplication. Exact duplicate projections remain parse-time errors. Plain `find --json` retains typed JSON; `find --json` with `--fields`/`--columns` emits lowered projected JSON.
- The earlier multi-TFM tabular buffering and Windows CRLF workaround were removed after `main` established the document-only `library --tfm all` contract: Markdown/JSON are valid, while TSV/JSONL/table fail closed.
- Exact-head CI run 31355826921 passed on rerun. The first attempt's only failure was the unrelated Windows package-acquisition timing test; the unchanged rerun passed the full Windows CLI suite and all remaining jobs.

Ready to merge

richlander:
Correction: retracting the `Ready to merge` status from my preceding comment. `main` advanced again during the final CI run; the branch remains green and mergeable, but repository policy requires a maintainer decision on carrying clean reviews across the new range before another integration.

richlander:
Ready to merge

Final candidate: `41cc42413c69a118b5eb71f734a0ad2882d5f4c9`.

Review reconciliation:
- Round 12 on `b108d428d` found three formatter-surface concerns. Two were reproduced and fixed in `15a27627a`: single-field callbacks and streaming tree-node callbacks now fail visibly instead of silently losing or mislabeling data, with real `MarkoutWriter` regression tests. The projected malformed-row observation was dismissed: Markout performs that transformation before every formatter, and the current source-generated `find` views emit fixed-width rows.
- Round 13 on `b60fd8c52` was clean from GPT-5.6 Sol and Claude Opus 5.
- After those clean reviews, `main` advanced from integrated base `c99fc505a` to `61e5ee308`. The landed range touched CI/provenance, graph/analysis, metadata, decompiler, package-path, and disjoint source-failure/project-command regions of `CommandExecutionTests.cs`; it did not touch the formatter, `find`, argument preprocessing, output-shape docs, or their tests. Per the user's decision, `main` was integrated in `41cc42413` and the clean reviews were carried forward without another round.

Current-head evidence:
- Release solution build passed.
- 311 focused formatter/`find`/command-line tests passed.
- CI run 31815196715 passed Linux, Windows, net10 build, markdownlint, and `ci-required`.
- GitHub reports `MERGEABLE` / `CLEAN` for the exact head.

Compatibility boundaries retained: plain `find --json` keeps typed arrays; named projections opt into lowered JSON; no-match behavior remains exit 0 with empty stdout and a stderr diagnostic; bare projection options remain typed-JSON no-ops.

Issue #3549, opened 2026-07-30
package tabular projection (--fields/--columns) ignores --rows while --count honors it


richlander:
Sibling of #3457, found while fixing it. #3457 fixes `OutputFormatter.WritePackageTable`; the projection branch bypasses it.

When `package` renders tabular output with a projection, `PackageCommand` renders through `OutputFormatter.RenderTable` and writes the result straight to the console so `ProjectionDiagnostics` can inspect it. That path never applies `options.Rows`, so the render disagrees with `--count`.

## Repro

```console
$ dotnet-inspect package Newtonsoft.Json@13.0.4 -S "Package Info" --rows 1..1 --count --fields Authors,Built
1
$ dotnet-inspect package Newtonsoft.Json@13.0.4 -S "Package Info" --rows 1..1 --tsv --fields Authors,Built
field	value
Authors	James Newton-King
Built	2025-09-16
```

Both exit 0. The count says one row; the render emits two.

## Cause

The `hasProjection` branch in `PackageCommand.cs` calls `OutputFormatter.RenderTable(...)` and then `Console.Out.Write(rendered)`. The non-projection sibling calls `WritePackageTable`, which #3457 teaches to forward `options.Rows`.

## Suggested fix

Window the rendered text with `OutputFormatter.LimitRenderedTableRows(rendered, options.Rows, !options.NoHeader)` before writing. Diagnostics must keep running against the *unwindowed* render, otherwise a projected field that exists but falls outside the window would be reported as missing.

Found by adversarial review (Gemini) on #3405, via #3457.

Issue #3548, opened 2026-07-30
package --jsonl on a file section ignores --rows while --count honors it


richlander:
Sibling of #3457, found while fixing it. #3457 fixes `OutputFormatter.WritePackageTable`; this path never reaches it.

`package -S Files --jsonl` short-circuits into `PackageCommand.WritePackageFilesJsonl`, which serializes every `PackageFile` row directly to the console and never consults `options.Rows`. `--count` still routes through `FormatResult`, so the two disagree.

## Repro

```console
$ dotnet-inspect package Newtonsoft.Json@13.0.4 -S Files --rows 1..2 --count
2
$ dotnet-inspect package Newtonsoft.Json@13.0.4 -S Files --rows 1..2 --jsonl | wc -l
33
```

Both exit 0. The TSV/table form of the same query is fixed by #3457 (it goes through `WritePackageTable`); only the JSONL short-circuit remains.

## Cause

`PackageCommand.cs` takes the `options.Jsonl && TryGetSingleFileSection(...) && !hasProjection` early return before the shared render path. `WritePackageFilesJsonl(InspectionResult, string)` has no row-window parameter.

## Suggested fix

Resolve `options.Rows` against the file list with `RowWindow.Resolve(files.Count)` and emit only the windowed slice, or route the path through `OutputFormatter.WriteTable` so the shared text limiter applies.

Found by adversarial review (Gemini) on #3405, via #3457.

PR #3550, opened 2026-07-30
Window the package tabular render with --rows (#3457)

### LINKED ISSUE #3457: package section render ignores --rows while --count honors it
`package -S <section> --rows <window> --count` reports the windowed count while the rendered table ignores the window entirely, so the count and the payload disagree — the inverse of the defect fixed for `--il-offsets` in #3405.

## Repro

```console
$ dotnet-inspect package Newtonsoft.Json@13.0.4 -S "Package Info" --rows 1..1 --count
1
$ dotnet-inspect package Newtonsoft.Json@13.0.4 -S "Package Info" --rows 1..1 --jsonl | wc -l
15
```

Both exit 0. The count says one row; the render emits fifteen.

## Cause

`--count` routes through `FormatResult`, which applies the row limiter. The render routes through `OutputFormatter.WritePackageTable`, which never passes `options.Rows` down to `OutputFormatter.WriteTable`. `WriteLibraryResult` does pass it, which is why `library -S <section> --rows N` windows correctly and `package` does not.

## Not a regression

Verified byte-identical on a binary built from `main` (`1955b054`), so this predates #3405.

## Suggested fix

Forward `options.Rows` from `WritePackageTable` into `WriteTable`, matching `WriteLibraryResult`.

Found by adversarial review (Gemini) on #3405.

richlander:
Fixes #3457.

`package -S <section> --rows <window> --count` reported the windowed count while the rendered table ignored the window entirely, so the count and the payload disagreed. `--count` routes through `FormatResult`, which applies the markdown row limiter; the render routed through `OutputFormatter.WritePackageTable`, which never passed `options.Rows` down to `WriteTable`. `WriteLibraryResult` does pass it, which is why `library -S <section> --rows N` already windowed correctly and `package` did not.

## Change

One call site: forward `options.Rows` from `WritePackageTable` into `WriteTable`. The caller has already narrowed to a single section for tabular output (`CheckMultiSection` collapses a multi-section selection to `Package Info`), so the rendered text is a single table and the existing text-level limiter applies unchanged — no new windowing semantics.

A null or unlimited window still takes `WriteTable`s unbuffered fast path, so output with no `--rows` is byte-identical.

## Evidence

Before:

```console
$ dotnet-inspect package Newtonsoft.Json@13.0.4 -S "Package Info" --rows 1..1 --count
1
$ dotnet-inspect package Newtonsoft.Json@13.0.4 -S "Package Info" --rows 1..1 --jsonl | wc -l
15
```

After: both report 1.

`CommandExecutionTests.PackageSection_Rows_WindowsTheTabularRenderAndAgreesWithCount` pins count/render agreement **with and without** a window (the no-window case is the negative that catches accidental windowing), and pins that an absolute `--rows 2..3` names rows 2 and 3 rather than the first two.

Non-vacuity: with the product change reverted, the test fails `Assert.Equal() Failure: Expected: 2, Actual: 15`.

Suite: `dotnet run --project src/dotnet-inspect.Tests -c Release` — 2503 total, 1 failed, 14 skipped. The single failure is `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on unmodified `origin/main` in this environment (`Assert.Contains` for `Newtonsoft.Json.nuspec`, a stale local package extract). Skips are the `ildasm`/`ilasm` and Windows-only cases.

## Residual (deliberately not in this PR)

Two sibling paths in `package` bypass `WritePackageTable` and still ignore `--rows`. Both were found while fixing this one and are filed rather than folded in, to keep this change to a single call site:

- #3548 — `package -S Files --jsonl` short-circuits into `WritePackageFilesJsonl`, which emits every row.
- #3549 — the `--fields`/`--columns` projection branch renders via `RenderTable` and writes unwindowed.

No behavior outside `package`s non-projection tabular render is touched. `library`, markdown, and JSON output paths are unchanged.

### DIFF
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -4132,6 +4132,47 @@ public partial class CommandExecutionTests
     }
 
     [Fact]
+    public async Task PackageSection_Rows_WindowsTheTabularRenderAndAgreesWithCount()
+    {
+        // Regression (#3457): --count windowed the section through the markdown limiter, but the
+        // tabular render never received options.Rows, so `--rows 1..1 --count` reported one row
+        // while `--rows 1..1 --jsonl` emitted the whole section. A count that does not describe
+        // the payload it is counting is worse than no count at all.
+        const string Package = "Newtonsoft.Json@13.0.4";
+
+        // Negative case: with no window, the count and the render already agreed, and must still.
+        var (bareCountExit, bareCountOutput, _) = await RunAppAsync(
+            "package", Package, "-S", "Package Info", "--count", "--tips", "q");
+        var (bareRowsExit, bareRowsOutput, _) = await RunAppAsync(
+            "package", Package, "-S", "Package Info", "--jsonl", "--tips", "q");
+        Assert.Equal(0, bareCountExit);
+        Assert.Equal(0, bareRowsExit);
+        var bareRows = bareRowsOutput.ReplaceLineEndings("\n").Split('\n', StringSplitOptions.RemoveEmptyEntries).Length;
+        Assert.True(bareRows > 2, "The section must have more rows than the window keeps for this test to prove anything.");
+        Assert.Equal(bareRows, int.Parse(bareCountOutput.Trim(), CultureInfo.InvariantCulture));
+
+        var (countExit, countOutput, _) = await RunAppAsync(
+            "package", Package, "-S", "Package Info", "--rows", "2..3", "--count", "--tips", "q");
+        Assert.Equal(0, countExit);
+        Assert.Equal(2, int.Parse(countOutput.Trim(), CultureInfo.InvariantCulture));
+
+        var (jsonlExit, jsonlOutput, _) = await RunAppAsync(
+            "package", Package, "-S", "Package Info", "--rows", "2..3", "--jsonl", "--tips", "q");
+        Assert.Equal(0, jsonlExit);
+        Assert.Equal(2, jsonlOutput.ReplaceLineEndings("\n").Split('\n', StringSplitOptions.RemoveEmptyEntries).Length);
+
+        // The window is absolute, so it names rows 2 and 3 of the section rather than the first
+        // two, and the header survives it.
+        var (tsvExit, tsvOutput, _) = await RunAppAsync(
+            "package", Package, "-S", "Package Info", "--rows", "2..3", "--tsv", "--tips", "q");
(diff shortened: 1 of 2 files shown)


richlander:
## Review reconciliation

**Tier:** more than trivial, not high risk — one product call site, no new windowing semantics, blast radius confined to `package`'s non-projection tabular render. One review round, with MAI-Code.

### Round 1 — MAI-Code, fixed head `613c1aa54`, isolated worktree

**No blocking findings.** Verified by real runs, not reading: built the worktree, ran the new test, and exercised the CLI with `--rows 2..3` across `--count`/`--jsonl`/`--tsv`, zero-width windows (`--rows 100..200`), and headerless TSV/JSONL forms — count and payload agreed in every case.

**Non-blocking (accepted, fixed in `4e5b8393d`):** the regression test hard-coded `Built`/`Content` from `Newtonsoft.Json@13.0.4`'s Package Info section, so a metadata change to that package would fail the test even though windowing was still correct. Now the expected slice is derived from the unwindowed TSV render — header plus rendered rows 2 and 3 — which still pins *absolute* window semantics without depending on the package's field list.

### Round 2 — MAI-Code, fixed head `4e5b8393d`, fresh isolated worktree

The head moved, so this confirmation round re-read the whole change at the new head rather than only the delta.

**Clean — no blocking, no non-blocking.** Previous finding confirmed resolved. The reviewer specifically attacked the risk that the fix *weakened* the test, and could not:

- Changed the window from `2..3` to `1..2` — the TSV assertion failed, proving it is pinned to "rows 2 and 3" and not "the first two rows".
- Removed the `options.Rows` forwarding from `WritePackageTable` — the test failed `Expected: 2, Actual: 15`, so the original regression is still exercised.
- `--rows 1..999999 --tsv` was byte-identical to the unwindowed render (`diff` reported zero lines).
- `--rows 2..3 --count` reported `2`; `--rows 2..3 --tsv --no-header` produced 2 data rows; `--rows 16..16 --tsv --no-header` (past the end of a 15-row section) produced empty output.

### CI

Green at both heads. At `4e5b8393d`: `changes`, `build-net10`, `test (ubuntu-26.04, linux-x64)`, and `decompiler-gates` all pass. `csharp-diff-smoke`, `il-diff-smoke`, `markdownlint`, and `pack` are path-filtered and skipped — not cited as evidence for anything here.

### Non-actions

#3548 and #3549 remain deliberately out of scope and are filed, not folded in. Both reviewers were told so and neither re-raised them; neither found a third `--rows`-ignoring path.

Ready to merge

PR #3553, opened 2026-07-30
Sink returns through a try/catch whose catch arm rethrows (#3552)

### LINKED ISSUE #3552: Decompiler: return accumulator survives a try/catch whose catch arm rethrows (TableClient.Create)
`ReturnSinkingPass` undoes csc's return-accumulator lowering: a `return` from inside an EH region is spilled to a synthetic local, `leave`s the region, and returns after the region closes. The pass reconstructs `try { return e; }` from `try { V = e; } finally { } return V;`.

It declines when the protected body is a `try/catch` whose catch arm rethrows, because `CollectStmtTails` demands a fall-through tail store from **every** arm and a rethrowing arm has none. That is the extremely common `using` + `try/catch`-that-rethrows shape.

## Witness

`Azure.Data.Tables.TableClient.Create` (`Azure.Data.Tables@12.11.0`):

```bash
dotnet-inspect member TableClient Create --package Azure.Data.Tables -S "Decompiled Source"
```

Today the accumulator survives:

```csharp
    Response<TableItem> V_3;
    {
        scope.Start();
        try
        {
            ...
            V_3 = Response.FromValue<TableItem>(response.get_Value(), response.GetRawResponse());
        }
        catch (Exception ex) { scope.Failed(ex); throw; }
    }
    return V_3;
```

The original source returns from inside the `try`:

```csharp
        try
        {
            var response = _tableOperations.Create(...);
        }
        catch (Exception ex) { scope.Failed(ex); throw; }
```

## Fix

An arm that cannot fall through — it rethrows, throws, or returns — never reaches the trailing return, so it contributes no tail rather than defeating the plan. This does not weaken the pass: `TryPlan` still refuses unless the plan consumes *every* store of the local, so an arm that stored the accumulator before terminating leaves that store unconsumed and the whole plan is rejected.

Note the `using` itself needs no new case: `UsingStatementPass` deliberately runs *after* `ReturnSinkingPass` (`IrPass.cs`), so a `using` reaches this pass as its underlying try/finally, which is already handled.

Out of scope (separate gaps visible in the same output): the unfolded `S_256`/`S_257` initializer spills (#3474), `response.get_Value()` (#2947), and the explicit trailing optional arguments (#3027).

richlander:
Fixes #3552.

## Claim

`ReturnSinkingPass` now sinks a return through a `try/catch` whose catch arm rethrows, so `return` lands back inside the `try` instead of leaving a synthetic accumulator behind — and the pass no longer silently deletes a loop `break` while doing so.

## Benchmark target

Benchmark target: `Azure.Data.Tables@12.11.0`

dotnet-inspect command:

```bash
dotnet-inspect member TableClient Create --package Azure.Data.Tables -S "Decompiled Source"
```

Base `59f731a1e` (the merge-base when this work started) → head `d54d2797a`.

`origin/main` has since been merged in, which brought #3474. The After below is captured at the current head, so it shows both changes.

### Original source

```csharp
{
    scope.Start();
    try
    {
            null,
            _defaultQueryOptions,
            cancellationToken);
    }
    catch (Exception ex)
    {
        scope.Failed(ex);
        throw;
    }
}
```

### Before

```csharp
{
    Nullable<ResponseFormat> V_2;
    Response<TableItem> V_3;
    TableProperties S_256;
    TableRestClient S_257;

    {
        scope.Start();
        try
        {
            S_256 = new();
            S_257 = _tableOperations;
            S_256.TableName = Name;
            V_2 = default;
            response = S_257.Create(S_256, V_2, _defaultQueryOptions, cancellationToken);
            V_3 = Response.FromValue<TableItem>(response.get_Value(), response.GetRawResponse());
        }
        catch (Exception ex)
        {
            scope.Failed(ex);
            throw;
        }
    }
    return V_3;
}
```

### After

```csharp
{

    {
        scope.Start();
        try
        {
            return Response.FromValue<TableItem>(response.get_Value(), response.GetRawResponse());
        }
        catch (Exception ex)
        {
            scope.Failed(ex);
            throw;
        }
    }
}
```

**This PR's delta is exactly the `V_3` accumulator**: its declaration, its assignment, and the trailing `return V_3;`. The `return` now sits inside the `try`, matching the original's control flow.

The `S_256`/`S_257`/`V_2` spills also disappear between Before and After, but that is #3474, merged into `main` while this PR was in review — not a claim of this change. With both in, the only remaining differences from the original source are the hoisted `response` declaration, the statement-vs-block `using`, `get_Value()`, and the explicit `default(...)` argument, each tracked separately below.

## Why it declined

The pass undoes csc's return-accumulator lowering (`try { V = e; } finally { } return V;` → `try { return e; }`). `CollectStmtTails` demanded a fall-through tail store from **every** catch arm, and a rethrowing arm has none — so the whole plan was rejected. That is the common `using` + `try/catch`-that-rethrows shape.

Note the `using` itself needed no new case. `UsingStatementPass` runs *after* `ReturnSinkingPass` by design (`IrPass.cs`: "Runs after return sinking so a `return` from inside the protected body stays inside the using body"), so a `using` arrives here as its underlying try/finally, which was already handled. I initially added a `UsingStatement` case and removed it once the ordering showed it was unreachable.

## Scope grew during review

Adversarial review found a **pre-existing fidelity bug** in the same collector, and it is fixed here because it lives in the code this PR changes.

`FallsThrough` answers "can control fall out of this arm". Its terminator set omitted `Break` and `Continue`, and `CollectStmtTails` did not consult it at all for `TryFinally.TryBody`, `Lock.Body`, `TryCatch.TryBody`, or the `if`/`else` arms. So an arm ending `accumulator = 2; break;` was collected as a foldable tail, `Apply` detached the `Break`, and a **loop `break` became a method `return`** — a silently deleted loop exit.

Reproduced through the full `IrPasses.Run` pipeline on Roslyn-compiled fixtures:

```csharp
while (x > 0)
{
    try
    {
        if (x == 1) { accumulator = 1; }
        else { accumulator = 2; break; }
    }
    catch { accumulator = 3; }
    return accumulator;
}
```

printed `else { return 2; }`. Checking out the base revision of `ReturnSinkingPass.cs` alone and re-running the identical fixtures gives byte-identical bad output, so this is not a regression introduced here.

The rule is now uniform across every arm: an arm that cannot fall through contributes no tail. `CollectFallThroughTail(BlockContainer)` and `CollectFallThroughBlockTail(Block)` return `[]` for such an arm, and `FallsThrough` gained a `Block` overload the container overload delegates to.

## Soundness boundary

Skipping an arm does **not** weaken the pass. `TryPlan` still refuses unless the plan consumes *every* store of the local, so an arm that stored the accumulator before terminating leaves that store unconsumed and the whole plan is rejected. The synthetic negative pins exactly that.

Returning `[]` ("contributes no tail") rather than `null` ("plan impossible") is deliberate, and was the highest-risk point of the final design. Both reviewers attacked it and neither produced a counterexample: an arm that transfers control away cannot reach the trailing return, so deleting that return cannot affect paths through the arm, and a store made before the transfer is still counted by the all-stores-consumed guard.

## Evidence

- Compiled fixtures (`ReturnSinkSamples`): `ReturnFromTryInsideUsing` is the witness shape; `ReturnFromUsingBody` is a control for the simple shape that already worked.
- Compiled fixtures (`ReturnSinkBreakSamples`): `CatchArmBreaksOutOfLoop`, `TryBodyBreaksOutOfLoop`, `FinallyProtectedBodyBreaksOutOfLoop`, `IfElseArmBreaksOutOfLoop`. The first three fail without the fix. The fourth already printed correctly on `main` — Roslyn's lowering does not present that shape to the pass — so it pins the uniform rule rather than witnessing a repaired regression, and its test comment says so.
- Synthetic negative `CatchArmStoringAccumulatorBeforeRethrow_StaysAccumulator` plus its positive control `RethrowingCatchArmWithNoAccumulatorStore_Sinks`.
- **Non-vacuity**: the new positives fail with the pass change reverted; the negative fails when the all-stores-consumed guard is tampered (`consumed.Count != indexStores.Count` → always false); unguarding both new helpers fails the try-body and finally-body cases.
- `ReturnSinkingPassTests` **19/19**.
- Decompiler `--gate fast`: **4356 total, 0 failed, 0 skipped** at the reviewed head; **4366 / 0 / 0** after merging `main`.
- `dotnet-inspect.Tests`: **2502 total, 0 failed, 14 skipped** at the reviewed head; **2581 / 0 / 14** after merging `main`.
- CI: `build-net10`, `test`, **`decompiler-gates`**, `csharp-diff-smoke`, `changes` all pass — the slow corpus and fidelity gates included.

## Review

Three rounds, two models, each at a fixed head in an isolated worktree. Gemini 3.1 Pro blocked twice (both findings real and accepted here) and cleared round 3. GPT-5.6-sol cleared all three rounds, but its round-2 characterization of the `break` loss as synthetic-only was incorrect and it confirmed that after seeing the compiled fixtures.

## Out of scope

Other gaps visible in the same witness, each already tracked:

- the unfolded `S_256`/`S_257` initializer spills — #3474, now **merged** and included in the After above
- `response.get_Value()` instead of `response.Value` — #2947
- explicit trailing optional arguments — #3027
- the hoisted `ResponseWithHeaders<...> response;` declaration, which the original spells `var response = ...` at the assignment site. `CSharpPrinter.CollectDeclaringStores` merges a declaration into its store only when that store is a statement in the **entry block**; `response`'s store is nested inside the `using`/`try`, so it falls back to a bare up-front declaration. Tracked by #3165 / #3071.

Follow-up filed: #3569 — `ReturnMergePass.FallsThrough` uses the same terminator idiom, also omits `Break`/`Continue`, and runs after `DoWhileLoopPass` constructs those nodes. Recorded as a latent risk; I found no counterexample and did not look hard. `EhStructuringPass.TerminatesUnconditionally` is provably unaffected.


richlander:
## Head moved to `31eacd48d` — round-1 finding addressed

### Gemini 3.1 Pro — BLOCKING: `FallsThrough` omits `Break`/`Continue`

**Accepted and fixed.** The finding is real: `FallsThrough` decided whether a `catch` arm reaches the statement after the `try/catch`, and its terminator set omitted `Break` and `Continue`. An arm ending in `accumulator = 2; break;` was therefore collected as a foldable tail — `CollectBlockTail` accepts a `StoreLocal; Break` pair, and `Apply` detaches the `Break` — rewriting a loop `break` into a method `return`.

I reproduced the counterexample exactly: `break;` printed as `return 2;`.

**The bug is pre-existing on `main`, not introduced here.** I reverted only `ReturnSinkingPass.cs` and re-ran the identical probe: byte-identical bad output. A `Break`-terminated arm reached the same tail collector before this PR as after it.

**Fixed here anyway**, because `FallsThrough` has exactly one call site — the catch-arm skip this PR introduces — so the fix is confined to the code under change rather than being an unrelated drive-by.

Commit `31eacd48d` adds `Break or Continue` to the terminator set, the compiled fixture `ReturnSinkBreakSamples.CatchArmBreaksOutOfLoop`, and a regression test asserting the `break` survives and the accumulator return is retained.

### GPT-5.6-sol — CLEAR at `ccb2edcf2`

No blocking findings. Because the head has moved, this verdict does not carry; a re-review at `31eacd48d` has been requested from both models.

### Evidence at `31eacd48d`

| Check | Result |
| --- | --- |
| `ReturnSinkingPassTests` | 16 / 16 |
| Decompiler `--gate fast` | **4353 total, 0 failed, 0 skipped** |
| `dotnet-inspect.Tests` | 2502 / 1 failed / 14 skipped |
| Non-vacuity | removing `or Break or Continue` fails `CatchArmBreakingOutOfLoop_KeepsTheBreakAndTheAccumulator` |

The single product-test failure is `CommandExecutionTests.Layout_Count_CountsFilesRatherThanRenderedTreeLines` (expected 20, actual 22). It reproduces byte-identically on a clean detached `origin/main` worktree, so it is environmental and not this PR's.

The commit also deletes stale untracked scratch that was sitting in the development worktree (a probe project plus a `partial class ObjectInitializerPassTests` test writing to a long-removed review worktree path). That scratch was the source of the "1 pre-existing failure" I reported earlier against the decompiler gate; with it gone the gate is clean at 0 failures.

### Non-blocking follow-up

`ReturnMergePass.FallsThrough` and `EhStructuringPass.TerminatesUnconditionally` use the same terminator idiom and also omit `Break`/`Continue`. Whether the misclassification is reachable through those paths is being checked as part of the re-review; if so it will be filed separately rather than expanded into this PR.

richlander:
## Round 2 reconciled — head moved to `5faac6217`

### Gemini 3.1 Pro — BLOCKING: the arm rule was only applied to catch arms

**Accepted and fixed.** `CollectStmtTails` collected a tail from
`TryFinally.TryBody`, `Lock.Body`, `TryCatch.TryBody`, and both `if/else` arms
without asking whether control could fall out of them, so the exact defect fixed
for catch arms in `31eacd48d` remained fully exposed everywhere else.

I reproduced it **through the full `IrPasses.Run` pipeline on Roslyn-compiled
fixtures**, not synthetically:

```csharp
while (x > 0)
{
    try
    {
        if (x == 1) { accumulator = 1; }
        else { accumulator = 2; break; }
    }
    catch { accumulator = 3; }
    return accumulator;
}
```

printed `else { return 2; }` — the `break` silently deleted. Same loss on a
`finally`-protected body. Both are **pre-existing on `main`**: checking out the
base revision of `ReturnSinkingPass.cs` alone and re-running the identical
fixtures gives byte-identical bad output.

Gemini's third claimed path, the `if/else` arm, **did not reproduce** with
Roslyn-compiled input — that shape already printed correctly. I guarded it for
uniformity anyway, and the test comment says exactly that rather than claiming a
repaired regression.

Fix takes Gemini's own suggested alternative: make the rule uniform instead of
adding one guard per call site. New `CollectFallThroughTail(BlockContainer)` and
`CollectFallThroughBlockTail(Block)` return `[]` for an arm that cannot fall
through, and `FallsThrough` gains a `Block` overload that the container overload
delegates to. The soundness argument is unchanged and already reviewed: skipping
an arm cannot strand a store, because `TryPlan` refuses unless the plan consumes
*every* store of the local.

### GPT-5.6-sol — CLEAR, with an incorrect non-blocking note

GPT cleared round 2 and reported the non-catch loss as synthetic-only, stating
"Roslyn-compiled probes normalized before reaching that shape". That is wrong, as
the compiled fixtures above show. Recorded here because the disagreement is the
reason this round exists: Gemini's read was correct and GPT's was not.

### Evidence at `5faac6217`

| Check | Result |
| --- | --- |
| `ReturnSinkingPassTests` | 19 / 19 |
| Decompiler `--gate fast` | **4356 total, 0 failed, 0 skipped** |
| `dotnet-inspect.Tests` | **2502 total, 0 failed, 14 skipped** |
| Non-vacuity | unguarding both helpers fails `TryBodyBreaksOutOfLoop` and `FinallyProtectedBodyBreaksOutOfLoop` |
| Over-declining | none — gate clean, witness unchanged |

The PR's actual witness still folds:

```csharp
{
    scope.Start();
    try
    {
        ...
        return Response.FromValue<TableItem>(response.get_Value(), response.GetRawResponse());
    }
    catch (Exception ex)
    {
        scope.Failed(ex);
        throw;
    }
}
```

`return` inside the `try`, no `V_3` return temp.

The `Layout_Count_CountsFilesRatherThanRenderedTreeLines` failure I reported last
round was file-count sensitivity to the scratch files I had in the worktree; with
them gone the product suite is clean at 0 failures.

Round-3 review requested from both models at fixed head `5faac6217`. Not ready to
merge until both are clean.

richlander:
## Round 3 — both reviews CLEAR at fixed head `5faac6217`

### Verdicts

| Reviewer | Round 1 | Round 2 | Round 3 |
| --- | --- | --- | --- |
| Gemini 3.1 Pro | BLOCKED | BLOCKED | **CLEAR** |
| GPT-5.6-sol | CLEAR | CLEAR (one incorrect note) | **CLEAR** |

Both reviewers ran the builds and tests themselves in isolated worktrees at the
fixed head and independently reproduced the numbers below.

### What each attacked

**`[]` versus `null` for a non-falling-through arm** — the highest-risk part of the
final design, since `[]` means "contributes no tail" while `null` means "plan
impossible". Both confirmed `[]` is sound and neither could build a counterexample.
Gemini's statement of why: an arm ending in a terminator, `break`, or `continue`
transfers control away unconditionally and cannot reach the trailing return, so
deleting that return does not affect paths through the arm; and if the skipped arm
*did* store the accumulator, that store stays in `indexStores` and is never added
to `consumed`, so `TryPlan`'s `consumed.Count == indexStores.Count` guard rejects
the plan. Gemini also noted that returning `null` would be a heavy over-decline on
`main` whenever an arm throws.

**Over-declining** — neither found a real shape that folded on `main` and no longer
folds. Gemini checked `ExpressionStatement` and `SwitchBranch`, which hit
`default: return null` before and after alike. The only newly-declined cases are
exactly the ones `main` folded erroneously.

**`TryCatch.TryBody` where the body cannot fall through but a catch arm can** —
confirmed sound by the same all-stores-consumed argument.

**`continue`** — every committed fixture uses `break`, so both built `continue`
counterexamples. Correctly classified as non-falling-through; the `continue`
survives untouched.

**Nested loops** — an inner-loop `break` inside a `try` inside an outer loop.
`FallsThrough` classifies the try body's final block (the inner loop) as falling
through, then `CollectStmtTails` hits `default` and returns `null`, declining to
fold because it does not model loops. Correct by construction.

**Completeness** — no remaining unguarded `CollectContainerTail` /
`CollectBlockTail` reachable with a `StoreLocal; Break` tail, including via the
`CollectBlockTail` → `CollectStmtTails` recursion.

GPT confirmed its round-2 "synthetic-only" characterization was incorrect.

### Evidence at `5faac6217`

| Check | Result |
| --- | --- |
| `ReturnSinkingPassTests` | 19 / 19 |
| Decompiler `--gate fast` | 4356 total, 0 failed, 0 skipped |
| `dotnet-inspect.Tests` | 2502 total, 0 failed, 14 skipped |
| CI | build-net10, test, **decompiler-gates**, csharp-diff-smoke, changes — all pass |
| Non-vacuity | unguarding both helpers fails `TryBodyBreaksOutOfLoop` and `FinallyProtectedBodyBreaksOutOfLoop` |
| Witness | `TableClient.Create` return is inside the `try`, no `V_3` temp |

`decompiler-gates` — the slow corpus and fidelity gates — passed, so the fold
behavior is confirmed against the corpus, not only the fast suite.

### Non-blocking follow-up

Filed #3569. `ReturnMergePass.FallsThrough` uses the same terminator idiom and
also omits `Break`/`Continue`, and it runs after `DoWhileLoopPass` constructs
those nodes, so they are reachable there. I did not find a counterexample and did
not look hard, so the issue records it as a latent risk rather than a confirmed
bug. `EhStructuringPass.TerminatesUnconditionally` is **not** affected: it runs
before any `Break`-constructing pass, and its use is conservative.

Ready to merge

richlander:
## Merged `origin/main` — conflict resolved, head `d54d2797a`

#3474 landed and appended its `CallArgTarget`/`CallArgClient` fixtures to the end
of `CfgSampleClass.cs`, the same place this branch appended `ReturnSinkScope`,
`ReturnSinkSamples`, and `ReturnSinkBreakSamples`. Both follow that file's
convention of adding top-level types at the end so no existing generated-code
ordinal shifts, so the conflict was **positional only**.

Resolved by keeping both sides in full. No fixture dropped, renamed, or reordered;
no pass or test logic changed. The merge commit touches only `CfgSampleClass.cs`.

### Validation after the merge

| Check | At reviewed head `5faac6217` | After merge `d54d2797a` |
| --- | --- | --- |
| Decompiler `--gate fast` | 4356 / 0 failed / 0 skipped | **4366 / 0 / 0** |
| `dotnet-inspect.Tests` | 2502 / 0 failed / 14 skipped | **2581 / 0 / 14** |

The counts rise by #3474's and other merged work's tests; nothing fails.

### The witness with both changes in

Now that #3474 is in the same tree, the accumulated result is worth seeing:

```csharp
{

    {
        scope.Start();
        try
        {
            return Response.FromValue<TableItem>(response.get_Value(), response.GetRawResponse());
        }
        catch (Exception ex)
        {
            scope.Failed(ex);
            throw;
        }
    }
}
```

against the original:

```csharp
{
    scope.Start();
    try
    {
            null,
            _defaultQueryOptions,
            cancellationToken);
    }
    catch (Exception ex)
    {
        scope.Failed(ex);
        throw;
    }
}
```

Four differences remain, none owned by this PR and each tracked: the hoisted
`response` declaration (#3165 / #3071), statement-vs-block `using` (a deliberate
non-change — `dotnet/runtime`'s `.editorconfig` sets
`csharp_prefer_simple_using_statement = false:none`, and both forms compile to
identical IL, so the author's choice is unrecoverable), `get_Value()` instead of
`.Value` (#2947), and the explicit `default(Nullable<ResponseFormat>)` argument
(#3027).

The merge is a mechanical resolution of an append-collision, so the round-3 CLEAR
verdicts from both reviewers stand on the pass and test logic they reviewed, which
this commit does not touch.

Ready to merge

PR #3557, opened 2026-07-30
Give the printed body map a producer (#3328)

richlander:
**Slice 1 of 2.** Slice 2 (#3558) wires this into the wasm browser engine.

`PrintedBodyMap` shipped in #3461 with **no production caller**. It was exercised only by tests, so the claim it was built for — that positions can leave the process and be re-rendered elsewhere — was untested in the role that motivated it. This adds the projection that produces one.

## What this adds

`ResearchViews.MemberProjectionRequest` gains an opt-in `BodyMap` flag, and the result gains a `PrintedBodyMap?`.

The projection prints the same raised body the overlay prints and stops one step earlier. `RenderRaisedOverlay` goes on to bake a chosen gesture into text, so a consumer receiving it can display only that choice; the map keeps the printer's ranges and the fact/node binding as data. That is what lets the *reader* choose a side comment, a caret, or a category focus.

Printer options are honoured, because positions measured against the shipped spelling would underline the wrong characters for a reader shown a style lens.

## The isolation, and how it was nearly fake

The projection takes its own IR import and is produced **before** the others. A byte-divergent lens rewrites the IR in place, so a map printed from the shared function would leave its rewrites on the graph the overlay and fact-row projections read.

That ordering is load-bearing, and I only know so because the test caught me:

- Written the obvious way, the map was produced **last**. The isolation test **passed under the shared-import mutation** — nothing read the graph afterwards, so corrupting it broke nothing.
- Moving the projection first was not enough either. Without a style lens there is nothing for the isolation to protect, so the mutation still passed.
- Routing `PrinterOptions` through — which the feature wanted anyway — made the hazard real, and the gate now bites.

## Evidence

| gate | mutation that breaks it |
|---|---|
| `AskingForTheMapLeavesEveryOtherProjectionIdentical` | share `imported` instead of taking an import |
| `PrinterOptionsReachTheMap` | pass `null` options to the map |
| `EveryPlacedSpanAddressesCharactersThatExistOnTheLineItNames` | shift every column by one |
| `ConditionalitySurvivesTheProjection` | drop conditionality from the payload |
| `TheProjectionIsOffUnlessAsked` | produce the map unasked |

Each mutation fails exactly one test and nothing else.

Suites: `ILInspector.Research.Tests` 138/138, `ILInspector.Decompiler.Tests` fast lane 4,348/4,348, `dotnet-inspect.Tests` 1,743/1,743.

## Honest negatives

- **`AllocationFrequency.PerIteration` is declared but never assigned anywhere in the pipeline.** My conditionality test targeted it first and failed. It pins `CachedOnce` instead — the only non-default value a producer actually emits. Pinning `PerIteration` would have gated nothing.
- **Facts still anchor to statements**, so spans are statement-wide (`col 0`, length = the statement). The narrow expression ranges from #3461 exist and place correctly; re-anchoring facts to them is separate work.
- **No consumer on `main`.** Slice 2 supplies one, in the browser engine.

### DIFF
--- a/src/ILInspector.Research.Tests/PrintedBodyMapProjectionTests.cs
+++ b/src/ILInspector.Research.Tests/PrintedBodyMapProjectionTests.cs
@@ -1,0 +1,213 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using ILInspector.Decompiler;
+using ILInspector.Decompiler.Annotations;
+using ILInspector.Decompiler.Pipeline;
+using ILInspector.Metadata;
+
+namespace ILInspector.Research.Tests;
+
+/// <summary>
+/// The projection that hands a member's positions to a consumer outside this
+/// process. <see cref="ResearchViews.MemberProjectionResult.AnnotatedSource"/>
+/// answers "what does this member look like annotated?" by baking one gesture
+/// into a string; the body map answers "where is everything?" and leaves the
+/// gesture to whoever renders. A browser needs the second, because the choice of
+/// side comment, caret, or category focus belongs to the reader.
+/// </summary>
+public class PrintedBodyMapProjectionTests
+{
+    static PrintedBodyMap Map(string method)
+    {
+        using var source = MetadataSource.Open(typeof(BodyMapFixture).Assembly.Location);
+        var projection = ResearchViews.ProjectMember(new ResearchViews.MemberProjectionRequest(
+            source,
+            typeof(BodyMapFixture).FullName!,
+            method,
+            BodyMap: true));
+        return Assert.IsType<PrintedBodyMap>(projection.BodyMap);
+    }
+
+    [Fact]
+    public void TheProjectionIsOffUnlessAsked()
+    {
+        using var source = MetadataSource.Open(typeof(BodyMapFixture).Assembly.Location);
+        var projection = ResearchViews.ProjectMember(new ResearchViews.MemberProjectionRequest(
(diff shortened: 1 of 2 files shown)


richlander:
Pausing this PR pending the outcome of #3570.

Two reasons, both from adversarial review of this PR plus follow-up measurement:

1. **An unfixed review finding.** The position test never asserts `Column >= 0`. Mutating every annotation column to `-1` still passed all six tests. That gap is real and is not addressed on this branch.
2. **The payload's shape is changing.** #3570 established that the projected ranges drop 100% of every multi-line construct, and that facts are anchored to *statements* rather than nodes — so the coordinates this PR serializes are coarser than they should be, and 6.98% of them are placed by a nearest-preceding fallback rather than by containment.

Landing this first would put a shape under test that #3570 intends to replace.

The stack now starts from #3574 on `main`. This PR will be restacked above the anchoring work and revisited once the payload settles; its projection and reflection-free serialization proof remain useful and are not being discarded.

richlander:
Closing as superseded by #3570. The current plan replaces this PR’s `PrintedBodyMap` producer contract with the portable `AnnotatedSourceLine` stream in slice 4, then adds producer/CLI wiring and harness gates in slice 5. The useful proof obligations from this PR—opt-in production, isolated imports, printer-option fidelity, and source-generated serialization—will carry forward there rather than landing the obsolete single-line payload API.

PR #3558, opened 2026-07-30
Give the printed body map a consumer: the browser engine (#3328)

richlander:
**Slice 2 of 2**, stacked on #3557 (targets `main`). This PR targets `prototype/wasm-command-ui`, so its diff also carries the `origin/main` integration that branch was missing; the slice's own change is the two files under `prototypes/inspect-web/engine/`.

Slice 1 added the projection. This is the **out-of-process consumer** that makes the payload's central claim testable at all.

## What this adds

`QueryMemberBodyMap` returns the printed body plus the character positions of every node and every fact, as `BrowserMemberBodyMap`.

`QueryMemberAnnotatedSource` already renders the same member — but it chooses a gesture in C#, bakes it into a string, and the browser can only display that one choice. Switching between side comments, carets, and a category focus costs another full decompilation round trip. The map carries positions instead, so those become re-renders of a single fetch.

The export takes `styleOptionsJson` and routes it to the projection's `PrinterOptions`, for the same reason the Source view does: the reader's taste changes the text, and positions measured against a different spelling underline the wrong characters.

## Is it actually wasm-safe?

This is the question #3461 could not answer. Its round-trip test passes **only because the test project permits reflection**, so it never demonstrated anything about a reflection-free host.

Two independent checks:

1. **The generator really covers the payload.** Building the engine emits `BrowserJsonContext.PrintedBodyMap.g.cs`, `.PrintedNodeSpan.g.cs`, `.PrintedAnnotationSpan.g.cs`, `.AnnotationConditionality.g.cs`, and the three collection type-infos — with a fast-path writer, not just metadata. `Conditionality` is written explicitly, which matters because `AnnotationText` appends `cached-once` to the label, so an envelope that dropped it would silently promote a cached allocation to an unconditional one.
2. **A reflection-free process really serialises it.** Verified with `JsonSerializerIsReflectionEnabledByDefault=false` over a real map from `System.Private.CoreLib`. The probe first asserts that a *reflection-based* serialize of the same payload **throws** in that process — otherwise the pass would prove nothing — then serialises through a source-generated context and confirms every placed span addresses characters that exist on the line it names.

Build: engine `Build succeeded`, 0 warnings. `engine.js` passes `node --check`.

## Deliberately not done

**No UI consumes this yet.** The engine boundary is what this slice claims to establish; rendering the C#/IL/annotation selector on top of it is the next piece of work, not a hidden part of this one. Flagging it as scope, not as a defect.

Also unchanged: facts still anchor to statements, so spans are statement-wide. Caret narrowing is separate.


richlander:
Pausing this PR.

The design work it rests on moved: #3570 concluded that `PrintedBodyMap` is a snapshot plus a flat coordinate sidecar rather than a portable line stream, and that its ranges drop 100% of every multi-line construct (`IfStatement` 81,432 dropped / 0 kept, likewise every loop, `try`, `using`, `switch`, `lock`, `fixed`). Building a browser consumer on that payload would harden a shape we already know we are replacing.

Also worth stating plainly for anyone reading later: **this PR has no CI.** Workflows do not trigger for PRs based on `prototype/wasm-command-ui`, so its green-looking state is an absence of signal, not a pass.

Work continues on `main` instead, starting with #3574. This will be revisited once the payload settles.

richlander:
Closing as superseded. The portable payload landed in #3790, the independent browser consumer and UI landed in #3964, and stable rendered-syntax kinds plus viewer kind filtering landed in #4029. This prototype targets an obsolete branch, conflicts with current `main`, has no CI on its base, and preserves no unique implementation that still needs to land.

PR #3559, opened 2026-07-30
Make CI change detection fail safe instead of failing the run

richlander:
## What breaks

`github.event.before` is not always the immediate parent. A push advances `main` by more than one commit whenever two PRs merge back to back or a batch lands. `fetch-depth: 2` left that commit outside the shallow clone, so the change-detection diff aborted:

```
fatal: Invalid symmetric difference expression 93d4924f...5a337055
##[error]Process completed with exit code 128
```

That is `main` at `5a337055`, where #3470 and #3499 merged in the same second. #3525 hit it earlier from the same cause. **Four of the last twenty-five `main` runs failed this way.**

## Why it is worth fixing rather than tolerating

Every downstream job is gated on this job's outputs, so a failure here does not merely lose the filter — it **skips the jobs it gates**.

On a push to `main` the cost is small and I want to be precise about it: `test`, `build-net10`, `decompiler-gates`, both diff smokes and `pack` are all deliberately `github.event_name == 'pull_request'`-only, so they never run on `main` regardless. The push-event cost is `markdownlint` plus a spuriously red `main`.

The real exposure is on **pull requests**. There `FILES` comes from `gh pr diff`, and that call was equally unguarded — a transient API failure would fail the step, skip the **entire build and test matrix**, and leave nothing but a red change-detection job to notice. A re-run of just that job, or a glance at a mostly-green checks list, would hide it. That is the footgun this closes.

## The fix

Three parts, in increasing order of importance:

1. **`fetch-depth: 0`** — the before-commit is present however far back it is. Root cause. The repo is 44 MB / ~2.6k commits, so a full fetch for this tiny job is free.
2. **Two-dot diff instead of three-dot** — a symmetric difference has to compute a merge base, which is a second way to fail and is not the question change detection is asking. `main` only takes fast-forward pushes, where the two agree.
3. **A fallback** — when the change set cannot be determined, treat every tracked file as changed and emit a `::warning::` rather than exiting non-zero. **CI over-runs instead of silently vanishing.** This now also covers the `gh pr diff` path.

Per the repo rule on asserted properties: the "this step must never fail" claim is enforced by the fallback branch itself, and the comment says so rather than implying an automated gate that does not exist.

## Evidence

Local harness against real history, running the old and new logic verbatim against a `--depth 2` clone and a full clone:

| case | logic | clone | `before` | result |
|---|---|---|---|---|
| A | old | shallow | 2 commits back | **EXIT 128 — run fails**, reproduces the production error verbatim |
| B | old | shallow | 1 commit back | ok, 11 files |
| C | new | full | 2 commits back | ok, 32 files |
| D | new | full | 1 commit back | ok, 11 files |
| E | new | full | zero SHA | ok, **fell back**, 1714 files |
| F | new | full | absent SHA | ok, **fell back**, 1714 files |
| G | new | **shallow** | 2 commits back | ok, **fell back**, 1714 files |

Two of those rows carry the argument:

- **B and D are byte-identical** (`diff` reports no difference), so ordinary single-commit pushes are completely unaffected.
- **G isolates the fallback from the `fetch-depth` change.** With the shallow clone deliberately left in place, the step still cannot fail. So the two mechanisms are independently effective, and a future regression in `fetch-depth` cannot resurrect the outage.

Workflow structure re-validated after the edit: YAML parses, all 8 `changes` outputs still declared and still written, no three-dot diff remains.

## Behaviour change worth flagging

Zero-SHA pushes (newly created or force-pushed branch) now take the fallback rather than guessing `HEAD~1`. For a branch with no prior tip, "everything changed" is the honest answer. This cannot affect `main`, which is protected.

## Not included

I did not add `always()` guards to the downstream jobs. The single-point guarantee — change detection cannot fail — is simpler and cannot drift out of sync across eight job conditions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,9 +50,10 @@ jobs:
           # This step must never fail. Every downstream job is gated on its
           # outputs, so failing here does not just lose the filter -- it SKIPS
           # the entire build and test matrix, landing an unvalidated commit on
-          # main. So when the change set cannot be determined, fall back to
-          # "everything changed": CI over-runs instead of silently vanishing.
-          # Enforced by the fallback branch below, not by an automated gate.
+          # main. So when the change set cannot be determined, force every
+          # output true: CI over-runs instead of silently vanishing. Enforced
+          # by the fallback branch below, which runs no fallible command, not
+          # by an automated gate.
           FILES=""
           RESOLVED=false
           if [ "${{ github.event_name }}" = "pull_request" ]; then
@@ -75,8 +76,18 @@ jobs:
             fi
           fi
           if [ "$RESOLVED" != "true" ]; then
-            echo "::warning title=Change detection fell back::Could not determine the changed files for ${{ github.sha }}; treating every tracked file as changed so that no validation is skipped."
-            FILES=$(git ls-files)
+            # Force every gate on and stop here. Deliberately NOT "list every
+            # tracked file and classify it": that would put another fallible
+            # git call -- and a dependency on every case pattern below matching
+            # something -- on the one path whose entire job is to be
+            # infallible. `git ls-files` exits 128 on an unreadable index.
+            # Keep this list in sync with the `outputs:` block above.
+            echo "::warning title=Change detection fell back::Could not determine the changed files for ${{ github.sha }}; running every job so that no validation is skipped."
+            for name in code csharpdiff decompiler docs ildiff ilroundtrip packaging shipped; do
+              echo "$name=true" >> "$GITHUB_OUTPUT"
+            done
+            echo "Change detection fell back; every job filter forced true."
+            exit 0
           fi
           CODE=false
           CSHARPDIFF=false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,19 +36,47 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
-          fetch-depth: 2
+          # Full history. Change detection diffs against github.event.before,
+          # which is not always the immediate parent: a push advances main by
+          # more than one commit whenever two PRs merge back to back or a
+          # batch is pushed. fetch-depth: 2 left that commit outside the
+          # shallow clone and the diff died with "Invalid symmetric difference
+          # expression", taking the whole run with it.
+          fetch-depth: 0
 
       - name: Detect changes
         id: filter
         run: |
+          # This step must never fail. Every downstream job is gated on its
+          # outputs, so failing here does not just lose the filter -- it SKIPS
+          # the entire build and test matrix, landing an unvalidated commit on
+          # main. So when the change set cannot be determined, fall back to
+          # "everything changed": CI over-runs instead of silently vanishing.
+          # Enforced by the fallback branch below, not by an automated gate.
+          FILES=""
+          RESOLVED=false
           if [ "${{ github.event_name }}" = "pull_request" ]; then
-            FILES=$(gh pr diff ${{ github.event.pull_request.number }} --name-only)
-          else
-            BEFORE=${{ github.event.before }}
-            if [ "$BEFORE" = "0000000000000000000000000000000000000000" ]; then
-              BEFORE=HEAD~1
+            if FILES=$(gh pr diff ${{ github.event.pull_request.number }} --name-only); then
+              RESOLVED=true
             fi
-            FILES=$(git diff --name-only "$BEFORE"..."${{ github.sha }}")
+          else
+            BEFORE="${{ github.event.before }}"
+            # The zero SHA means a new branch or a force push with no prior tip.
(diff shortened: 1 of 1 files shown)


richlander:
## Review reconciliation

Two-family adversarial review, escalated on blast radius: this job's outputs gate
every other job in the workflow, so a defect here is silent skipping rather than a
visible failure.

| Reviewer | Head reviewed | Verdict |
| --- | --- | --- |
| GPT-5.6 | `bf958fa0`, then `fb8e44db` | No blocking findings at `fb8e44db` |
| Gemini 3.1 Pro | `fb8e44db` | No blocking findings at `fb8e44db` |

### Findings and disposition

**1. GPT — SHOULD-FIX, accepted, fixed in `fb8e44db`.** In `bf958fa0` the fallback
was `FILES=$(git ls-files)` falling through to the normal classification loop. That
is another fallible git call on the one path whose entire purpose is to be
infallible. Reproduced:

```
$ bash -e -c 'GIT_INDEX_FILE=/dev/null; export GIT_INDEX_FILE; FILES=$(git ls-files); echo SURVIVED'
fatal: nul: index file smaller than expected
(exit 128; SURVIVED never printed)
```

`fb8e44db` writes the eight outputs directly and `exit 0`, running no fallible
command on the fallback path. This also removed an implicit dependency the earlier
version had — that every `case` pattern matches at least one tracked file — which
had only been verified empirically.

Tamper evidence for the fix: reinstating `git ls-files` in the fallback makes the
step exit 128 again. Worth noting what the tamper showed when the reinstated call
was placed *after* the output writes — the outputs were still written correctly, and
the step still exited 128, so the job still failed and every gated job still
skipped. Correct-looking outputs do not save a failing step.

**2. Gemini — NON-BLOCKING, accepted as a known limitation.** The fallback
enumerates the eight output names in a shell `for` loop; a ninth output added later
without updating that list would come back empty and silently under-run. Both
reviewers independently judged the sync comment adequate for this PR (GPT: "further
shell indirection would add complexity without eliminating YAML duplication").
Recorded as non-blocking follow-up rather than fixed here.

**3. No other findings.** Neither reviewer found a reachable path on which the step
can exit non-zero, other than an unset or unwritable `$GITHUB_OUTPUT` — runner
infrastructure failure, and outputs cannot be published without it. Both confirmed
that is the only residual.

### Independently verified by both reviewers

- The `case` classification block is **byte-for-byte identical** to `00b3c1c9`.
- Three-way set equality of the eight output names: the job's declared `outputs:`
  mapping, the normal path's writes, and the fallback's `for` loop.
- The `changes` job has exactly two steps, so `exit 0` in the second one skips
  nothing.
- Two-dot vs. three-dot is a no-op on this repo's topology: `main` only takes
  fast-forward pushes, where the merge base of `$BEFORE` and `$GITHUB_SHA` is
  `$BEFORE`, so `A..B` and `A...B` agree. The two-dot form simply removes the
  merge-base computation as a second failure mode.

### Real-run evidence

The `Detect changes` script was extracted from `ci.yml`, had its `${{ }}`
expressions substituted, and was executed under Git Bash:

| Scenario | Exit | Outputs |
| --- | --- | --- |
| ordinary push, resolvable `before` | 0 | discriminates (`docs=false`, `ilroundtrip=false`) |
| unresolvable `before` | 0 | all eight `true`, fell back |
| empty `before` | 0 | all eight `true`, fell back |
| all-zeros `before` (branch creation) | 0 | all eight `true`, fell back |
| `gh pr diff` forced to fail | 0 | all eight `true`, fell back |
| unresolvable `before` + broken git index | 0 | all eight `true`, fell back |
| resolvable `before` + broken git index | 0 | discriminates |
| **tampered** (`git ls-files` reinstated) | **128** | — |

A separate old-vs-new harness against `--depth 2` and full clones reproduces the
production failure verbatim on the old logic (`fatal: Invalid symmetric difference
expression`, exit 128, two-commit push) and shows the old-good and new-good runs
byte-identical on a one-commit push.

CI is green on `fb8e44db` (run
[30518091446](https://github.com/richlander/dotnet-inspect/actions/runs/30518091446)),
and was green on `bf958fa0` as well.

### Behavior change, flagged

Zero-SHA pushes (branch creation) now take the all-true fallback rather than the
previous `HEAD~1` guess. This cannot affect protected `main`.

Ready to merge

richlander:
Post-merge addendum: a third reviewer (MAI-Code), whose run stalled and only completed after merge, independently reproduced the `git ls-files` fallback defect at the superseded head `bf958fa0` and reported no other findings. Its trigger was distinct -- it pointed `GIT_INDEX_FILE` at a directory rather than `/dev/null` -- so it is an independent confirmation rather than a rerun of the same tamper. The defect was already fixed in `fb8e44db`, which is what merged, and which both GPT-5.6 and Gemini 3.1 Pro reviewed clean. No action needed; recording it so the finding's provenance is complete.

Issue #3542, opened 2026-07-30
CI has no Windows leg, so Windows-only breakage is only ever found by hand


richlander:
The PR test matrix (`.github/workflows/ci.yml:190-196`) declares a single include:

```yaml
matrix:
  include:
    - os: ubuntu-26.04
      rid: linux-x64
```

Every other job in `ci.yml` is a hard-coded `runs-on: ubuntu-26.04`. Nothing in CI ever runs on Windows, so a green PR says nothing about Windows.

The cost is already visible: #3496, #3509, #3519, #3526, #3445 and #3326 are all Windows-only failures found by hand, on a tree CI called green. A sixth instance was found during review of #3428 — six `PluginDiscoveryTests` cases failing on Windows because the fixtures assumed Unix executability and path separators.

Most of these are fixture-level (CRLF vs LF, separators, extensions) rather than product defects, which is exactly the class a matrix leg catches cheaply and a human catches expensively.

Worth weighing against the fixed CI cost noted in AGENTS.md. A reduced Windows leg — build plus the suites that have historically broken — would cover most of the observed cases without doubling the whole matrix.

Issue #3519, opened 2026-07-29
Two CommandExecutionTests cases fail on Windows on a clean main


richlander:
Two `CommandExecutionTests` cases fail on Windows on a clean `origin/main`. Neither is related to any in-flight change; both reproduce at `dcaed617` in a fresh detached worktree.

```
DotnetInspector.Tests.CommandExecutionTests.Layout_Count_CountsFilesRatherThanRenderedTreeLines [FAIL]
DotnetInspector.Tests.CommandExecutionTests.Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown(readmePath: "images/logo.png.") [FAIL]
   dotnet-inspect.Tests  Total: 5, Errors: 0, Failed: 2, Skipped: 0, Not Run: 0
```

Reproduce:

```
git worktree add -f ../wt-mainbase dcaed617 --detach
cd ../wt-mainbase
dotnet build dotnet-inspect.slnx -c Release
dotnet run --project src/dotnet-inspect.Tests -c Release --no-build -- `
  -method "*Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown*" `
  -method "*Layout_Count_CountsFilesRatherThanRenderedTreeLines*"
```

The `readmePath` parameter is `images/logo.png.` — a trailing dot, which Windows strips from path components. That is the likely cause for the first; the layout count case is probably a path-separator or enumeration-order difference. Both look like the same family as #3509 (Windows-specific test assumptions), not product defects on Windows.

Found while validating #3505; unrelated to that PR, which touches only the IL body diff substrate.

Issue #3509, opened 2026-07-29
Windows: three CorpusSensorComparisonTests quality-diff-card tests fail on CRLF vs LF expectations


richlander:
## Summary

Three `CorpusSensorComparisonTests` quality-diff-card tests fail on **Windows only**. They assert against string literals containing `\n` escapes, but the card is rendered through Markout, which emits `Environment.NewLine` — `\r\n` on Windows. On Linux both sides are LF and the tests pass, so CI never sees this.

## Failing tests

- `CorpusSensorComparisonTests.QualityDiffCard_RealWorld_RendersEntirelyThroughMarkout`
- `CorpusSensorComparisonTests.QualityDiffCard_StaleBaseline_LeadsWithTableThenStalenessAndAppendsCaveat`
- `CorpusSensorComparisonTests.QualityDiffCard_OptInNet11_UsesProfileHeadingAndSkipsPinnedGate`

## Observed

```text
ILInspector.Decompiler.Tests.CorpusSensorComparisonTests.QualityDiffCard_RealWorld_RendersEntirelyThroughMarkout [FAIL]
  Assert.StartsWith() Failure: String start does not match
  String:         "### Decompiler quality diff\r\n\r\n| Metric | Change |"...
  Expected start: "### Decompiler quality diff\n\n| Metric | Change |\n|"...
  src\ILInspector.Decompiler.Tests\CorpusSensorComparisonTests.cs(1670,0)
```

## Mechanism

The assertions at `src/ILInspector.Decompiler.Tests/CorpusSensorComparisonTests.cs:1670` and nearby use ordinary (non-raw) string literals with `\n` escapes, so the expectation is LF on every platform. `CorpusSensor.QualityDiffCardForTesting` renders through Markout, which uses the platform newline. The two only agree on Linux.

Note this is **not** the same bug as #3454: there the *product* (`MarkdownSectionOrderer.Apply`) was rewriting CRLF to LF and a test was pinned to that rewrite. Here the product is correct in using the platform newline and the *tests* carry the platform assumption.

## Not a regression from #3455

Verified by bisecting across the #3455 merge on Windows, Release:

| Commit | Result |
| --- | --- |
| `4039832d` (#3450, immediately before #3455) | same 3 failures |
| `c2ddd98f` (#3455) | same 3 failures |
| `4d44b1c8` (current `main`) | same 3 failures |

So this predates #3455 and is independent of it.

## Repro

```bash
[...]

Issue #3496, opened 2026-07-29
Four decompiler tests fail on Windows only: CRLF-vs-LF assertion mismatches


richlander:
Four tests in `ILInspector.Decompiler.Tests` fail on Windows and pass on Linux. All four are CRLF-vs-LF assertion mismatches: the code under test emits `Environment.NewLine` (or otherwise picks up the platform line ending) while the expected string is an LF-only C# literal.

CI never sees this because the lanes run on Linux, so it is invisible until a Windows developer runs the suite locally — where it produces a red fast lane on a clean `main` and buries any real failure a developer is actually looking for.

## Affected tests

| Test | Assertion site |
| --- | --- |
| `StringSwitchRaisingTests.SmallStringSwitch_RendersSwitchOnString` | `src/ILInspector.Decompiler.Tests/StringSwitchRaisingTests.cs:46` |
| `CorpusSensorComparisonTests.QualityDiffCard_RealWorld_RendersEntirelyThroughMarkout` | `src/ILInspector.Decompiler.Tests/CorpusSensorComparisonTests.cs:1670` |
| `CorpusSensorComparisonTests.QualityDiffCard_OptInNet11_UsesProfileHeadingAndSkipsPinnedGate` | `src/ILInspector.Decompiler.Tests/CorpusSensorComparisonTests.cs:1698` |
| `CorpusSensorComparisonTests.QualityDiffCard_StaleBaseline_LeadsWithTableThenStalenessAndAppendsCaveat` | `src/ILInspector.Decompiler.Tests/CorpusSensorComparisonTests.cs:1727` |

## Evidence

`StringSwitchRaisingTests.cs:46` — `Assert.Equal`, differing at position 10, exactly where the first newline falls:

```text
Assert.Equal() Failure: Strings differ
                     ↓ (pos 10)
Expected: "switch (s)\r\n{\r\n    case "a":\r\n        return 1;\r\n "···
Actual:   "switch (s)\n{\n    case "a":\n        return 1;\n    c"···
                     ↑ (pos 10)
```

Note the direction: here the *expected* value carries CRLF and the *actual* rendered output is LF, so this one is the mirror image of the other three — the expected string is being normalized to the platform ending (most likely a raw string literal in a CRLF-checked-out file) while the renderer emits LF.

`CorpusSensorComparisonTests.cs:1670` — `Assert.StartsWith`, opposite direction:

```text
Assert.StartsWith() Failure: String start does not match
[...]

PR #3562, opened 2026-07-30
Normalize line endings so a checkout is identical on every platform

### LINKED ISSUE #3526: No .gitattributes: CRLF checkout breaks every raw-string test expectation on Windows

richlander:
Fixes #3526.

## What this is

The repository has no `.gitattributes`. Git for Windows defaults to
`core.autocrlf=true`, so every text file is rewritten to CRLF on checkout.
Test expectations written as C# raw string literals then carry CRLF, while the
product printers emit `\n`. Those tests pass on Linux CI and fail for every
Windows developer.

## The evidence settles which side is wrong

#3496 and #3526 give opposite diagnoses of the *same* failure. #3496 says the
code under test emits `Environment.NewLine` against an LF-only literal. #3526
says the literal carries CRLF against a printer that emits `\n`. The trace,
quoted identically in both issues, is decisive:

```text
Expected: "switch (s)\r\n{\r\n    case "a":\r\n        return 1;\r\n "...
Actual:   "switch (s)\n{\n    case "a":\n        return 1;\n    c"...
```

Expected is CRLF; actual is LF. The printers are correct and the checked-out
test source is what moved, so the fix belongs in checkout, not in the printers.
Had #3496's mechanism been taken at face value, the fix would have gone into
product code that is not at fault.

## The change

`* text=auto eol=lf` pins the working-tree ending rather than only the stored
one, so literals stay LF on Windows too.

`*.sh` is called out separately because a trailing CR becomes part of the
interpreter path.

The binary list protects `fixture.nettrace`, which is the only committed binary
today, plus the assembly and package extensions this repository is most likely
to grow fixtures for. `git check-attr` confirms it resolves to `text: unset,
binary: set`, so no conversion is applied to it.

## Why this is low risk

I surveyed the committed tree before writing the file:

- Exactly **one** committed blob contains CR bytes, and it is the binary
  `fixture.nettrace`. Every other committed file is already pure LF.
- `git add --renormalize .` reports **zero** content change. The diff is the
  new file and nothing else, so no existing blob is rewritten.

Because the working tree is byte-identical on Linux, Linux behaviour cannot
change. The four tests named in #3496 were run here and pass, as they did
before.

## What still needs a Windows run

I have no Windows machine, so the part that matters is unverified by me. On a
fresh Windows clone of this branch, with `core.autocrlf` left at its default:

```console
dotnet run --project src/dotnet-inspect.Tests -c Release --no-build -- -class DotnetInspector.Tests.CommandExecutionTests
```

A **fresh clone is required**. An existing checkout keeps its CRLF working-tree
files until `git add --renormalize .` is run, so re-testing in place will
reproduce the old failures and say nothing about this change.

Expected outcome per issue, stated as a prediction to be checked rather than a
claim:

| Issue | Prediction | Reasoning |
| --- | --- | --- |
| #3526 | Fixed | This is the root cause it describes |
| #3496 | Fixed | Same trace, same mechanism |
| #3509 | Fixed | #3526 names it as the same root cause |
| #3445 | Possibly fixed | Its trace shows CRLF in the *actual* string, so it depends on whether a checked-out corpus input feeds the printer. Genuinely unsure |
| #3519 | Not fixed | Trailing-dot path stripping and enumeration order, unrelated to line endings |

## Sequencing note

#3542 proposes a Windows CI leg. It should land **after** these fixes, not
before: adding a Windows leg while these suites are still red would simply make
`main` red.


richlander:
## Windows verification run — this PR is a net regression on Windows

I offered on #3428 to be the Windows runner, so here is the run you asked for. **The headline is not the one predicted: `.gitattributes` is necessary but not sufficient, and landing it alone makes the Windows suite worse.**

### Environment

- `core.autocrlf` = `true` from **system** config, i.e. the exact Git for Windows default this PR describes.
- Two **fresh clones**, as required. No existing checkout was reused.
- Control: `main` @ `ce77becd`. Subject: this branch @ `9af6cfa0`.
- `git diff --stat origin/main...HEAD` is exactly `.gitattributes | 30 ++++`. The two trees differ in nothing else, so every difference below is attributable to line endings alone.

### The PR's own mechanical claims all check out

- Fresh clone on this branch: `AddressInValuePositionTests.cs` has **0** CR bytes. Same file in a CRLF checkout: **130**. The `eol=lf` pin works.
- `git check-attr` on `*.nettrace` resolves to `text: unset, binary: set`. No conversion applied.
- `git add --renormalize .` produces **zero** staged content change. No existing blob is rewritten, so Linux cannot move.

That part of the PR is sound. The problem is elsewhere.

### Controlled per-class comparison

`ILInspector.Decompiler.Tests`, Release, `--no-build`, same machine, same test content:

| Test class | Total | main (CRLF tree) | this branch (LF tree) | Effect |
| `StringSwitchRaisingTests` | 9 | **1 failed** | **0 failed** | fixed |
| `CorpusSensorComparisonTests` | 63 | **3 failed** | **3 failed** | unchanged |
| `MemberBodyProducerUnionTests` | 23 | **0 failed** | **10 failed** | **regressed** |

Net effect on these three classes: fixes 1, leaves 3, **breaks 10**.

Full suite on this branch: `Total: 4594, Failed: 20`.

### Why — both diagnoses were right, about different tests

The failures that remain or appear have the **opposite** direction from the one this PR's evidence quotes. On this branch:

```text
CorpusSensorComparisonTests.QualityDiffCard_RealWorld_RendersEntirelyThroughMarkout
  String:         "### Decompiler quality diff\r\n\r\n| Metric | Change |"...
  Expected start: "### Decompiler quality diff\n\n| Metric | Change |\n|"...
```

```text
MemberBodyProducerUnionTests.ClassUnionSwitchExpression_RendersTypePatternArms
  Expected: "return pet switch\n{\n    Cat cat => cat.Name,\n    D"...
  Actual:   "return pet switch\r\n{\r\n    Cat cat => cat.Name,\r\n  "...
```

Here the **expectation is LF and the product output is CRLF** — `Environment.NewLine`. That is precisely the mechanism #3496 described and this PR dismissed.

Both mechanisms are real; they affect different tests:

1. **#3526's mechanism** — CRLF literal vs LF printer. Real. `.gitattributes` fixes it. `SmallStringSwitch_RendersSwitchOnString` now passes.
2. **#3496's mechanism** — LF literal vs CRLF printer. Also real, in a larger set of tests. `.gitattributes` does not fix it, and **actively exposes it**: before this change the CRLF literal accidentally matched the CRLF product output, so those 10 union tests passed on Windows for the wrong reason. Normalizing the literals removes the accidental match.

So "the printers are correct and the checked-out test source is what moved" holds for the tests the PR examined and not as a general claim. The product genuinely emits platform-native newlines:

```console
$ git grep -c "AppendLine(" -- "src/*.cs" ":!*Tests*"
src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.cs:90
src/ILInspector.Decompiler/MemberBodyProducer.cs:32
...
```

`StringBuilder.AppendLine()` uses `Environment.NewLine`, and `CSharpPrinter.cs` and `CSharpPrinter.RaisedExpressions.cs` additionally use `Environment.NewLine` explicitly (e.g. `CSharpPrinter.cs:625`, `:2030`, `RaisedExpressions.cs:180-183`).

### Prediction scorecard

| Issue | Predicted | Observed |
| --- | --- | --- |
| #3526 | Fixed | Mechanism confirmed; fixes part of the class of failures |
| #3496 | Fixed | **Fixed** — `SmallStringSwitch_RendersSwitchOnString` passes |
| #3509 | Fixed | **Not fixed** — all 3 `QualityDiffCard_*` still fail, unchanged from main |
| #3445 | Possibly fixed | Not measured in this run; different suite |
| #3519 | Not fixed | **Correct** — `CommandExecutionTests` = 762 total, 2 failed: `Layout_Count_CountsFilesRatherThanRenderedTreeLines` and `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown(readmePath: "images/logo.png.")` |

### Recommendation

Do not land this alone. The `.gitattributes` content is correct and I would keep it exactly as written — the problem is sequencing, not the file.

Pair it with making the decompiler and Markout-facing print paths emit `\n` explicitly instead of `Environment.NewLine`/`AppendLine()`, and land the two together. Fixing the product first also independently improves the tool: today `dotnet-inspect decompile > out.cs` produces different bytes on Windows than on Linux, which is a real behavioural difference and not only a test artifact.

This also sharpens the PR's own sequencing note about #3542. A Windows CI leg added after this lands would show **more** decompiler failures than one added today, not fewer.

I have the machine and both clones set up, so iteration is cheap on my side. Happy to take the product-side newline fix and validate the combined change here — say the word.

richlander:
## Windows follow-up: the prerequisite is now open as #3595

Following up on [my earlier comment](https://github.com/richlander/dotnet-inspect/pull/3562#issuecomment-5128369147). I've now built and Windows-tested the product-side half, and it's up as **#3595**.

### Recap of the problem

This PR is correct but incomplete. There are two opposite defects, and normalizing the tree only fixes one of them:

| | mechanism | fixed by `.gitattributes`? |
| --- | --- | --- |
| #3526 | CRLF *literal* vs LF *printer* | yes |
| #3496 | LF *literal* vs CRLF *printer* (product used `Environment.NewLine`) | **no — and it gets unmasked** |

Before this PR, the CRLF literals accidentally matched CRLF product output, so ten `MemberBodyProducerUnionTests` passed *for the wrong reason*. Normalizing the tree removes that accident.

### Measured on Windows

Full `ILInspector.Decompiler.Tests`, Release, same machine, two fresh clones differing only in `.gitattributes`:

| | `main` @ `ce77becd` | this PR @ `9af6cfa0` |
| --- | --- | --- |
| Total | 4602 | 4594 |
| **Failed** | **11** | **20** |

Net **+9**: `SmallStringSwitch_RendersSwitchOnString` is fixed, and 10 `MemberBodyProducerUnionTests` regress.

Per class, including the combination with #3595:

| Test class | `main` (CRLF tree) | this PR alone | this PR + #3595 |
| --- | --- | --- | --- |
| `CorpusSensorComparisonTests` (63) | 3 | 3 | **0** |
| `MemberBodyProducerUnionTests` (23) | 0 | **10** | **0** |
| `StringSwitchRaisingTests` (9) | 1 | 0 | **0** |

### On the prediction table in the description

- **#3496 — Fixed.** Confirmed.
- **#3519 — Not fixed.** Confirmed; exactly two `CommandExecutionTests` still fail, and they are not line-ending related.
- **#3509 — predicted Fixed, but it is *not*.** All three `QualityDiffCard_*` tests still fail with this PR alone. The cause is `MarkoutTemplate.Render()`, whose writer defaults to `Environment.NewLine`; no amount of tree normalization reaches it. #3595 fixes it by rendering through the `Render(TextWriter)` overload.

### Sequencing

Neither change is green on Windows by itself:

| | expected literals | product output | union tests |
| --- | --- | --- | --- |
| today | CRLF | CRLF | pass (wrong reason) |
| this PR alone | LF | CRLF | **fail** |
| #3595 alone | CRLF | LF | **fail** |
| both | LF | LF | **pass** |

So **#3595 should land first and this PR should rebase on top of it.** #3595 is byte-neutral on Linux by construction, so it is safe to land ahead of this one.

### Mechanical claims in this PR — all verified on Windows

- Fresh clone has 0 CR bytes.
- `*.nettrace` resolves to `binary: set / text: unset`.
- `git add --renormalize .` produces zero content change, confirming the stored blobs were already LF.

No changes requested here beyond rebasing onto #3595 once it lands. Happy to re-run the full Windows suite on the combined result — the other agents on this repo have no Windows access, so I'm the runner for that.

richlander:
## Retargeted onto #3595

I have changed this PR's base from `main` to `fix/product-emits-lf` (#3595), and merged that branch in. This PR's unique diff is now **exactly `.gitattributes`, 30 lines** — every other file in the comparison comes from the branch below it.

### Why

This change is correct, but it is **not green on Windows by itself**, and neither is #3595. They fail in opposite directions, so only the pair is green.

Windows passes today for the wrong reason: CRLF literals in the working tree accidentally match CRLF product output. This PR makes literals LF while the product still emits the ambient CRLF. #3595 makes the product emit LF while a CRLF checkout still holds CRLF literals. Both halves break the same 10 tests.

| tree state | expected literals | product output | `MemberBodyProducerUnionTests` (23) |
| ---------- | ----------------- | -------------- | ----------------------------------- |
| `main` | CRLF | CRLF | 0 failed (for the wrong reason) |
| this PR alone | LF | CRLF | 10 failed |
| #3595 alone | CRLF | LF | 10 failed |
| **stacked** | **LF** | **LF** | **0 failed** |

All four rows are measured on Windows. CI has no Windows job, so it reports green for either PR alone and cannot see this.

### Correction to this PR's prediction table

The description predicts that #3509 is fixed by tree normalization. It is not. The `QualityDiffCard` failures come from `MarkoutTemplate.Render()`, whose string-returning overload inherits `Environment.NewLine`. Normalizing the tree does not touch it. That is fixed in #3595 by rendering through `Render(TextWriter)`.

### Combined result on Windows

Measured on the stacked head:

| suite | `main` | this PR alone | **stacked** |
| ----- | ------ | ------------- | ----------- |
| `ILInspector.Decompiler.Tests` | 11 failed | 20 failed | **7 failed** |
| `dotnet-inspect.Tests` | — | — | **2 failed** |

The 7 remaining decompiler failures are confirmed pre-existing on `main` by name (`ValidityCoverageReportingTests.DecompilerAssembly_MissingReturnPopulationIsPinned`, `LoweredFidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket`, `FidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket`, `FidelityGateTests.PinnedFixesStayExact`, and three `GeneratedFixtureCatalogTests`). The 2 CLI failures are pre-existing #3519.

So the stack removes exactly the 4 line-ending failures and adds none.

### Merge order

Merge bottom-up: this PR into `fix/product-emits-lf`, then #3595 into `main`. That gives one merge commit to `main` that is green on Windows, with no interval in which `main` is worse than it is today.

If the two are instead merged separately into `main`, **#3595 must go first** — this PR must never precede it.

### What I did not touch

I did not rebase or force-push this branch; the merge is an ordinary merge commit, so existing review history is intact. The `.gitattributes` content is unchanged from what you wrote.

richlander:
Ready to merge — into `fix/product-emits-lf` (#3595), not into `main`.

This PR is now the top of a two-PR stack. Its unique diff is exactly `.gitattributes`, 30 lines; every product and test change lives in #3595 below it.

**Merge this PR first**, into #3595's branch. That advances #3595's head to the combined content and triggers CI there (its base is `main`). Then merge #3595 into `main` once green.

Do not merge this PR into `main` on its own, and do not merge #3595 to `main` before this one lands on it. Normalization alone makes Windows worse: the CRLF literals it rewrites to LF currently match the product's CRLF output, so #3562 alone fails 10 `MemberBodyProducerUnionTests` that pass on `main`. #3595 supplies the other half — LF product output — and the two are green only together.

Validation, reviews, and the full evidence table are on #3595: https://github.com/richlander/dotnet-inspect/pull/3595#issuecomment-5137040337

On the stacked head `ae82b3be`, run on Windows:

| suite | result |
| --- | --- |
| `ILInspector.Decompiler.Tests` | 4620 total, 7 failed — all 7 reproduce on `main` by name |
| `dotnet-inspect.Tests` | 2581 total, 2 failed — both pre-existing (#3519) |

This PR receives no CI while stacked, because `ci.yml` triggers on `pull_request: branches: [main]`. That is benign: its blob-level diff against #3595 is only `.gitattributes`, Linux checkouts are already LF, and the bottom-up merge order restores full CI coverage on #3595 before anything reaches `main`.

Issue #310, opened 2026-04-05
`--latest-version` / `--versions` can crash when a local folder source appears before a valid HTTP NuGet feed


aynur-safin:
## Summary

`dotnet-inspect` crashes with `System.NotSupportedException: net_http_unsupported_requesturi_scheme, file` when resolving package versions if a local folder source from `NuGet.Config` is processed before a valid HTTP/HTTPS feed that contains the package.

This happens for packages that are not available on `nuget.org` but are available on another remote feed.

## Repro

My `NuGet.Config` contains a mix of sources like:

- `https://api.nuget.org/v3/index.json`
- `D:\some\local\packages`
- `C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\`
- `https://my-private-feed/.../index.json`

Now run a version query for a package that is not on `nuget.org` but is present on the private feed:

```bash
dotnet-inspect MyApi.Core --latest-version
```

or

```bash
dotnet-inspect package MyApi.Core --versions 5
```

## Actual behavior

The command crashes with:

```text
Unhandled exception: System.NotSupportedException: net_http_unsupported_requesturi_scheme, file
```

The stack trace goes through version-resolution code paths such as:

- `PackageExtractor.GetVersionsAsync`
- `FetchAllVersionsFromSourceAsync`
- `GetPackageBaseAddressAsync`
- `HttpRetryHelper.GetStringWithRetryAsync`

## Expected behavior

Local folder sources should not cause version lookup to crash.

Possible acceptable behaviors:

- skip non-HTTP/non-HTTPS sources for remote version queries
- handle local folder sources explicitly
- catch per-source failures and continue trying the remaining sources

In all cases, `dotnet-inspect` should continue to the next source instead of failing the entire command.

## Why this seems to happen

From reading the code, `--latest-version` for a bare package name appears to route through the `ListVersions` path with `Limit = 1` and `ForceLatest = true`, which eventually calls `PackageExtractor.GetVersionsAsync`.

`GetVersionsAsync` iterates over resolved sources and calls `FetchAllVersionsFromSourceAsync(...)` for each source.

For a local folder source, the code eventually falls back to `GetPackageBaseAddressAsync(...)`, which assumes the source can be queried like a V3 HTTP feed and tries to fetch an `index.json` via `HttpClient`. That becomes a `file:` URI and throws `System.NotSupportedException`.

## Notes

[...]

Issue #3540, opened 2026-07-30
Plugin credential cache is keyed per authority, so Azure DevOps orgs share one credential slot


richlander:
`PluginAuthenticationHandler` caches acquired credentials per **authority**:

```csharp
// src/NuGetFetch/Plugins/PluginAuthenticationHandler.cs:183
private static string GetAuthority(Uri uri) => uri.GetLeftPart(UriPartial.Authority);
```

That is scheme, host, and port. Combined with `HttpClientFactory.Shared` being a single process-wide client carrying one decorator, every package source on the same host shares one credential slot.

NuGet.Client does not work this way: it builds an `HttpSource`, and therefore a handler, **per package source**, so credentials never travel between sources.

### Why it matters in practice

Every Azure DevOps feed lives under `https://pkgs.dev.azure.com`, regardless of organization or feed. So:

- A credential acquired for org A's feed is attached to requests for org B's feed.
- The credential plugin is asked once per **host**, not once per **source**, so a provider that would serve different credentials per feed cannot.

Mostly benign with org-scoped PATs against a single server, and not an external disclosure — but it is a real divergence from the model users expect from NuGet, and it silently defeats per-feed credential separation.

### Options

Either key the cache by package source rather than authority, or document the divergence deliberately with the reasoning for accepting it.

Found during adversarial review of #3428 (https://github.com/richlander/dotnet-inspect/pull/3428#issuecomment-5125076109).

PR #3563, opened 2026-07-30
Report a source that refuses access as unreadable, not as a missing package

### LINKED ISSUE #3417: Authenticated NuGet feeds: 401 reported as "package not found", and package search returns empty against every Azure DevOps feed

richlander:
Part of #3635.

Fixes #3417.

This is the fix for **bug 1**. Bugs 2, 3 and 4 were open when this PR was raised but are already fixed on `main`; I verified each one end to end and posted the evidence on the issue (https://github.com/richlander/dotnet-inspect/issues/3417#issuecomment-5128146041), so bug 1 is the last one outstanding and merging this closes the issue.

## The defect

Azure DevOps answers **401**, not 404, for a feed the caller cannot see. So an unauthenticated
private feed and a mistyped package name both ended at the same message:

```console
$ dotnet-inspect package Markout --source https://pkgs.dev.azure.com/<org>/<project>/_packaging/<feed>/nuget/v3/index.json
Error: Package 'markout' not found.
```

The package may well exist. The *feed* was unreadable, and nothing said so.

## Why it matters more now than when it was filed

Before credential support merged (#3428), reaching this message meant you had configured no
credentials, and "not found" was merely unhelpful.

Now users will configure credentials — and **an expired credential is the normal end state of
every PAT**. Azure DevOps is actively shortening that lifetime; `docs/design/nuget-authentication.md`
already links the [global PAT retirement](https://devblogs.microsoft.com/devops/retirement-of-global-personal-access-tokens-in-azure-devops/)
(December 1, 2026, pushing toward short-lived Entra-backed auth). That makes this the message
users will most often see from the feature that just shipped, at the exact moment the correct
action is "refresh your credential".

I hit it myself while smoke-testing #3428 after merge, and spent several minutes assuming a tool
bug before `curl` showed a 401.

## Why the fix is shaped this way

The status was never missing. `HttpRetryHelper` classifies 401 as non-retryable and captures it
in `HttpRetryResult`. It is lost one frame later, because `GetLatestVersionAsync` returns
`string?` and cannot carry a source or a phase. #3417 anticipated this: *"I originally hoped Bug 1
was severable from the typed failure model; it is not, because the current signatures cannot
carry the source or the phase."*

Changing every signature between the two would be a large refactor of a 1,200-line file and its
callers. Instead `FeedFailureTelemetry` follows the ambient-scope shape this repository already
uses in `NetworkTelemetry`, `RequestTelemetry`, and `CacheTelemetry`: `ExtractPackageAsync` opens
one scope, nested async work records into the same collector, and the "nothing resolved" path
consults it before choosing a message.

The **phase** comes free from the ambient `NetworkTrafficKind` that the telemetry scope already
tracks, so no call site had to be taught to describe itself.

## Result

```console
Error: Package 'markout' could not be resolved because a source requires credentials.
  https://pkgs.dev.azure.com/rich0534/markout-downstream/_packaging/markout/nuget/v3/index.json — HTTP 401 Unauthorized while reading the service index
The package may exist; the source was not readable. Supply credentials for this source and retry.
```

Verified against the live Azure DevOps feed that prompted this work, which answers 401 because
PATs are disabled for the account.

## Two rules that keep the message honest

- **404 is never recorded.** It is the one status that genuinely means the package is absent, so
  a real miss still reports *not found*. A recorder that captured every non-success status would
  satisfy the positive assertions while destroying the real message, so the suite pins 404 as an
  explicit control.
- **A recorded failure is advisory, not fatal.** The collector is consulted only when the overall
  lookup produced nothing, so if one source refuses and another answers, the success stands. That
  is a concrete answer to the third open design question in #3417.

Two exits deliberately do **not** record: an unsupported URL scheme (a local-folder source being
skipped benignly, per #310) and `OfflineException` (offline mode has its own message). Recording
either would manufacture noise.

## Verification

| Case | Result |
| --- | --- |
| Live AzDO feed returning 401 | Names source, status, and phase; exit 1 |
| `ThisPackageDoesNotExistXYZ123` | Still `Error: Package 'ThisPackageDoesNotExistXYZ123' not found.`; exit 1 |
| `package Newtonsoft.Json` | Resolves 13.0.4; exit 0 |
| `dotnet-inspect.Tests` | 2586 passed, 0 failed, 14 skipped |
| `DotnetInspector.Services.Tests` | 357 passed, 0 failed |
| New `FeedFailureTelemetryTests` | 5 passed |

**The new tests gate the fix.** Stashing only `HttpRetryHelper.cs` and rebuilding turns 3 of the
5 red. The 2 that stay green are exactly the controls — the 404 case and the no-scope case —
which assert the *absence* of recording and so must pass in both worlds.

## Docs

`docs/design/nuget-authentication.md` asserted the old behaviour in two places; both are updated,
and a new *Reporting an unreadable source* section records the design and the two rules above.
Lints clean.


richlander:
Both findings verified in code and fixed in `9cd30154c`. Details, including one place I did not do what was suggested.

**1. Redaction was incomplete — confirmed, both vectors.**

I added both to the redaction theory before changing anything, and both failed:

```
ASecretInTheSourceUrlIsNeverStoredOrPrinted(url: "...?accessToken=...") [FAIL]
ASecretInTheSourceUrlIsNeverStoredOrPrinted(url: ".../F/feed/auth/sup3rs3cret/ap"...) [FAIL]
```

I had expected the path vector to be the weak half of this finding and I was wrong. `/F/<feed>/auth/<token>/api/v3/index.json` is MyGet's published service index shape, so a credential in the path is a real NuGet feed form, not a contrived one. That is the part I would have dismissed without running it.

Fixes, both in the shared helper (`NetworkTelemetry.cs`), since a leak through network telemetry is no better than one through this message:

- Query names now match on fragments (`token`, `key`, `secret`, `password`, `credential`, `auth`, `sig`) instead of an exact-name list. `accessToken`, `apiKey` and `x-api-key` were all passing before. The tradeoff is that a harmless parameter containing one of those fragments now gets redacted; that costs a little diagnostic detail, where the reverse error prints a live credential.
- The path segment following an `auth` segment is redacted.

**Where I did not follow the suggestion:** I did not redact path segments generally. An Azure DevOps organization, project and feed all live in path segments, and #3540 established that AzDO orgs share an authority — so blanket path redaction produces a message that cannot distinguish two feeds on the same host, which is exactly what this message exists to do. Targeted `auth`-segment redaction closes the demonstrated vector without that cost. Happy to reconsider if there is a second real-world path shape this misses.

**2. Redirect-loop contamination — confirmed, and worse than described.**

The scope wrapped `while (true)`, and each hop resolves a different package id. The scope now opens per hop. The gating test uses a source that refuses *only* the wrapper and returns a plain 404 for the target, so hop 2 has no failure of its own. Reverting just the scope move turns it red with the contamination in plain sight:

```
Assert.DoesNotContain() Failure: Sub-string found
String: ..."s credentials.\n  https://refusing.example/v3/flat2"...
```

Note the recorded URL carries the *wrapper's* id in its flat-container path, so the old behaviour named both a source and a package unrelated to the failure. That makes this more than cosmetic, though I agree it was right to call it non-blocking.

**Still outstanding from my original list.** Five of the six risks I raised have not been addressed in either round: `AsyncLocal` propagation correctness, cross-package contamination outside the redirect path, whether the message can be actively *wrong* when nuget.org 404s and a private source 401s, `FeedFailureCollector` thread safety under concurrent source probes, and whether a plausible wrong implementation would pass the new tests. The redirect finding was a real instance of the second one, which is why I would like the rest looked at rather than assumed clean.

Validation at `9cd30154c`: CLI 2593/0 (14 skipped), Services 357/0, NuGetFetch 292/0 (9 skipped), markdownlint clean.

One process note: my first attempt to reproduce this looked like a flaky test. It was not — the stash-gate step had left deliberately-broken binaries in place and my `--no-build` runs were reusing them. Worth flagging because the failure looked exactly like a race and I nearly treated it as one.

richlander:
## Review reconciliation — three rounds, MAI-Code

**Ready to merge.** Final verdict at fixed head `4e3b71a2f`: NOT BLOCKING.

Every round found something real, which is worth recording plainly.

| Round | Finding | Outcome |
| --- | --- | --- |
| 1 | URLs recorded and printed verbatim could leak credentials | Valid — fixed in `949a3786b` |
| 2a | Redaction incomplete: `accessToken` and a path-borne token both leaked | Valid — fixed in `9cd30154c` |
| 2b | Failure collector spanned the whole redirect traversal, so one package's refusal could explain another's absence | Valid — fixed in `9cd30154c` |
| 3a | "The message can be actively wrong when the package is genuinely absent" | Withdrawn by the reviewer after discussion |
| 3b | A wrong implementation treating a refusal as fatal would pass the suite | Valid — fixed in `4e3b71a2f` |

Two of these deserve a note.

**The path-credential vector is the one I would have missed.** I expected it to be the weak half of the round-2 finding and said so before testing it. It is not: `/F/<feed>/auth/<token>/api/v3/index.json` is MyGet's published service index shape, so a credential in the *path* is a real NuGet feed form. Running it was what settled it:

```
ASecretInTheSourceUrlIsNeverStoredOrPrinted(url: ".../F/feed/auth/sup3rs3cret/ap"...) [FAIL]
```

I still declined the proposed remedy of redacting path segments generally, because an Azure DevOps organization, project and feed all live in path segments and #3540 established that AzDO orgs share an authority — blanket redaction would produce a message unable to distinguish two feeds on one host, which is the message's entire purpose. The targeted `auth`-segment rule closes the demonstrated vector without that cost.

**Finding 3a was withdrawn on the merits, not conceded.** The claim was that a genuinely absent package plus a refusing private source now wrongly blames the feed. But the tool cannot distinguish "absent everywhere" from "present in the source I was not allowed to read" — that indistinguishability *is* bug 1, so both cases necessarily produce one message and the only question is which. "Not found" asserts a fact never established and strands the user; the refusal message asserts only what happened, hedges with *"The package may exist; the source was not readable"*, and converges — supply credentials, retry, and a genuinely absent package reports a clean "not found", which `ANotFoundIsNotRecorded_SoAGenuinelyAbsentPackageStillReportsAsNotFound` pins. The reviewer agreed and restated it as an optional wording preference, which I declined with reasons on the thread.

**Where the reviewer was most useful was 3b**, and not in the form it was raised. The suite genuinely had a hole: every test had either no successful source or no failing one, so nothing pinned the rule that a recorded failure is *advisory*. A private feed 401ing beside a public one that serves the package is the ordinary arrangement, and a "refusal is fatal" implementation would have broken it while staying green. Verified by mutation — injecting `if (FeedFailureTelemetry.Current?.HasFailures == true) outcome = Error(...)` inside the acquisition scope turns the new test red and leaves the rest of the suite passing.

**Validation at `4e3b71a2f`:** CI all green. Locally CLI 2594/0 (14 skipped), Services 357/0, NuGetFetch 292/0 (9 skipped), markdownlint clean. Every behavioural claim in this PR is gated by a test proven to fail without its fix.

**Scope note:** this closes #3417 rather than only its bug 1, because bugs 2, 3 and 4 turned out to be already fixed on `main`. I verified each end to end rather than trusting the log — including `package search` against a public Azure DevOps feed, the case that used to silently return zero results — and posted the evidence on the issue.

**Not merging:** `main` is protected and that is yours to do.

One follow-up filed, deliberately not folded in here: #3572. `package search` still reports a 401 as *"no searchable endpoint (service index unavailable, or advertises no SearchQueryService)"* — bug 1's misdiagnosis on the search path. `NuGetSourceCompat.cs:278-286` carries a comment saying it is waiting on exactly the typed-failure mechanism this PR introduces, so this PR unblocks it.

richlander:
## Correction: the elided URL was mine, not the tool's

Reviewing the PR text, the sample output read:

```
  https://pkgs.dev.azure.com/.../nuget/v3/index.json — HTTP 401 Unauthorized while reading the service index
```

That `...` was hand-elision in my prose. **Nothing in the code elides the path**, and it would be a defect if it did — this message's only job is to say *which* source refused, and on Azure DevOps the organization, project and feed name are all path segments, so two feeds in one organization differ nowhere else. I made exactly that argument when declining blanket path redaction in round 2, then undercut it by printing an abbreviated example.

What the tool actually emits, run just now against the real refusing feed:

```console
$ dotnet-inspect package markout --source https://pkgs.dev.azure.com/rich0534/markout-downstream/_packaging/markout/nuget/v3/index.json
Error: Package 'markout' could not be resolved because a source requires credentials.
  https://pkgs.dev.azure.com/rich0534/markout-downstream/_packaging/markout/nuget/v3/index.json — HTTP 401 Unauthorized while reading the service index
The package may exist; the source was not readable. Supply credentials for this source and retry.
```

Full path, org and feed intact. And with a credential in the query, only the credential goes:

```console
$ dotnet-inspect package markout --source '<same URL>?access_token=hunter2'
  https://pkgs.dev.azure.com/rich0534/markout-downstream/_packaging/markout/nuget/v3/index.json?access_token=REDACTED — HTTP 401 Unauthorized while reading the service index
```

Fixed in `d24f14f68`: both console examples in `docs/design/nuget-authentication.md` now show real output, and the redaction note no longer claims merely that "the host survives" — the whole path survives; only the credential is removed. The PR body above is corrected too.

No product code changed; the behaviour was always right, only its documentation was wrong. Still ready to merge.

richlander:
## New behavior added after the NOT BLOCKING verdict (`dc8670b93`)

Flagging this explicitly because it is a **product** change landing after the round-3 verdict at `4e3b71a2f`, so that verdict does not cover it.

### What changed

`--source` and `--add-source` now **reject** a URL carrying `<user>:<password>`:

```
$ dotnet-inspect package markout --source https://alice:t0ken@pkgs.dev.azure.com/org/_packaging/feed/nuget/v3/index.json
Error: Source URL 'https://pkgs.dev.azure.com/org/_packaging/feed/nuget/v3/index.json' embeds <user>:<password>, which NuGet does not support. Configure the credentials in a nuget.config, or use a credential provider.
```

Exit 1, before any network call, and the credential is not echoed back.

### Why

Verified while answering a question on the docs: **no NuGet client sends the userinfo component**, and dotnet-inspect has zero `UserInfo` handling anywhere. The form is a git/curl convention, which is exactly why operators reach for it. Previously it was silently ignored — the request went out anonymous and came back 401, so the user was told to supply credentials they believed they had just supplied. That is the same class of misdiagnosis this PR exists to fix, so it belongs here.

### Shape of the fix

A parse-time validator, mirroring the existing `--nugetconfig`/`DescribeConfigProblem` precedent at `SharedOptions.cs`. This matters: `Program.cs`'s generic `catch (Exception)` prints a full stack trace and is reserved for bugs, so a user-facing option problem has to be surfaced by a validator rather than thrown.

`DescribeSourceProblem` **reports rather than throws**, and rebuilds the URL through `UriBuilder` with empty userinfo so the secret never reaches the terminal.

### Relationship to the redaction work from rounds 1–2

This does not replace it. The validator covers the two options; redaction still covers a URL arriving from a `nuget.config` or a redirect. Defense in depth, and the review history for the redaction path stands unchanged.

### Tests

7 new (a 3-case rejection theory and a 4-case acceptance theory pinning that nuget.org and AzDO URLs are untouched). Docs shrank by 18 lines: they now describe only mechanisms that actually authenticate, instead of explaining a pattern the tool now refuses.

### Unrelated failure, for the record

While validating this I hit `Layout_Count_CountsFilesRatherThanRenderedTreeLines` failing locally. It **reproduces on a clean checkout of `main`** and is caused by nuspec filename casing differing between the global-packages-folder and `.nupkg` acquisition paths — it passes on CI because a fresh runner has no restored copy. Filed as #3592; not related to this PR.

richlander:
## `InertText` vs. the updated spec

Scoped to the `InertText` files only — nothing here is about the feed-failure change. Reviewed at `99d1a58e`.

The spec this tracks has moved since this was written. It now lives in `docs/design/metadata-table-projection.md` on `docs/hardened-untrusted-metadata` (#3601), at `72e28766`. Three of the five notes below are deltas from that update, not mistakes in this PR.

### What this got right

Worth saying explicitly, because these are the parts that are easy to get wrong and were gotten right:

- **The wide `Cf`.** `Cc | Cf | Cs | Zl | Zp` is exactly the set the spec settled on, including choosing `Cf` over the narrower `Bidi_Control` that `ApiOutputFormatter.IsBidiControl` implements. The spec now records that divergence and instructs unifying *upward*.
- **Surrogate pairs are not hazards.** `IsPairAt` skips well-formed pairs so only unpaired surrogates encode. A naive `foreach (char)` spells every emoji `\uD83D\uDE00`; this avoids it.
- **A real decoder.** `TryDecode` exists, which is what makes invertibility an asserted property rather than a comment.
- **The caret-collision reasoning**, captured in-code. `U+001E + 0x40 == '^'` is exactly right and is why the backslash introducer is load-bearing.

### 1. Non-BMP scalars pass through unencoded (correctness)

`Encode` and `IsInert` iterate `char` and, on a well-formed surrogate pair, append both units without consulting `IsActive`. `IsActive` takes a `char`, so it cannot classify an astral scalar at all. 127 code points in the encoded categories live above the BMP.

Measured against this branch's own code:

```text
IsInert   : True
Encoded   : U+0041 U+13430 U+0042
Unchanged : True
non-BMP scalars in encoded categories: 127
```

`U+13430` is EGYPTIAN HIEROGLYPH VERTICAL JOINER, general category `Cf`. It is in the set this class exists to encode, it is returned untouched, and `IsInert` reports the string as inert. The others are `U+110BD`, the rest of `U+13430`-`U+1343F`, and the musical-symbol format characters.

The fix is structural rather than a patch: iterate `Rune`, classify with `Rune.GetUnicodeCategory`, and add the astral spelling. `\uXXXX` is four hex digits and cannot express these, so the spec adds C#'s `\UXXXXXXXX`:

| Input | Spelling |
| --- | --- |
| C0 | `\^` + (code point + `0x40`) |
| `DEL` | `\^?` |
| other encoded scalar in the BMP | `\uXXXX` |
| other encoded scalar above the BMP | `\UXXXXXXXX` |
| literal `\` | `\\` |

`TryDecode` needs the matching `\U` arm, and the round-trip sweep should run `U+0000`-`U+10FFFF` rather than stopping at `U+FFFF` — which is why the gap survived the existing tests.

### 2. The predicate belongs to the caller (spec delta)

`IsActive` is private and fixed, which makes this a *policy* library as well as a spelling one. The updated spec separates them:

- The **predicate** answers "is this scalar permitted *here*?" It is per-sink and holds the policy. Deny-shaped for free-form text, allow-shaped for constrained identifiers.
- The **speller** answers "how do I write a scalar that is not permitted?" One function, total over Unicode, that never learns *why*.

The reason this matters is the main attack in scope. A package-id sink wants an allow list, because the central typosquatting vector is a homoglyph: Cyrillic `а` U+0430 and Latin `a` U+0061 are the same glyph and both are category `Ll`. No category rule catches that, and none should — neither character is a hazard. Only the grammar catches it. A speller with a built-in hazard set cannot serve that sink, and because the speller is total, an allow-list caller needs no hazard set of its own.

That is also what makes `--dangerously-skip-checks` useful: with the allow list as the predicate, a hijacked `Nеwtonsoft.Json` renders `N\u0435wtonsoft.Json`, so the mode that says "show me anyway" shows why the tool objected.

### 3. Report which forms were emitted (spec delta)

New rule in the spec: a mapping table earns its place by being *recognizable in the output*. Caret notation is small and has real audit value, but `\^[` is obvious only to someone who knows `cat -v` — and a spelling the reader cannot decode looks like data.

So the caller emits a legend, on **stderr**, since stdout is the data channel and a legend there would corrupt structured output. It is emitted only when something was encoded, names forms and never values, and is *derived from the forms actually emitted* rather than from a second copy of the table — a legend written independently drifts, one projected from the encoder cannot. That needs a small addition here: `Encode` should report which forms it used.

### 4. `IsInert` is a claim without a gate

Finding 1 is really this one. `IsInert` returns `True` for a string containing `U+13430`. Per `AGENTS.md`, an asserted property names the gate that enforces it; this one asserts in its name and has none behind it.

Under the split it is also the policy problem in API form — it hard-codes one answer to a question that belongs to the sink.

### 5. Naming

`InertText` is the right *word* on the wrong noun, for two reasons beyond the above.

The type no longer owns inertness once the predicate is a parameter: pass a permissive one and `Encode` returns non-inert text while the name still promises otherwise. And "inert" is scoped in a way the name cannot carry — it means no terminal interprets the text and no bidi algorithm reorders it, explicitly **not** that the text is safe to drop into a structured format. A `|` still breaks a Markdown cell, a backtick still opens a span, a `"` still terminates a JSON string; none is in any encoded category and none should be, because escaping those is the *serializer's* job. `InertText.Encode(x)` invites precisely that mistake.

Suggestion: keep `InertText` as the assembly and namespace — inertness is the goal of the effort and that is the right home for it — and name the type for what it does, e.g. `VisualEncoder`, matching the spec's "visual encoding, not neutralization".

### 6. Stale remark

The class remarks currently tell callers that a Cyrillic `а` is "confusability analysis, a different problem with a different answer, and callers who need it need something in addition to this." That was true of the older spec. It now points away from the answer: the allow-list predicate handles it, in this component. Worth updating so it does not steer the next implementer off.

### Minor

`TryDecode`'s `\u` arm accepts a lone `\uD83D` and yields an unpaired surrogate. Round-trip on encoder output is unaffected, but this is a public entry point taking untrusted input, so rejecting an unpaired result is probably worth it.

---

Nothing above blocks the feed-failure work this PR is really about. Happy to be wrong on the naming — findings 1 and 4 are the ones I would not ship without.

richlander:
## Orientation for agents planning to consume `InertText`

This branch is readable now — clean tree, pushed, merged with latest `main` at `26ed34d44`.
Posting the state plainly so you can judge what to build on and what to wait for.

### What is solid

- **The predicate/speller split.** `ScalarPolicy` decides *what is permitted here*; `VisualEncoder`
  decides *how a refused scalar is written down* and never learns why. This matches the six
  obligations in the #3601 spec and is the part least likely to move.
- **The speller is total over Unicode**, including the 127 encoded scalars above the BMP, and
  ships with its decoder. There is a whole-plane `U+0000`–`U+10FFFF` round-trip sweep.
- **Zero `ProjectReference`, by design.** `InertText` sits below everything, so you can reference
  it from any project including the dependency-free leaves. Please keep it that way.

### Two things to know before you build on it

**1. No allow-shaped policy ships yet.** `TextPolicy` exposes only `Field` and `Prose`, both
deny-shaped (`Cc | Cf | Cs | Zl | Zp`, with `Prose` exempting `CR`/`LF`/`TAB`).

If you are doing identifier work — the `NuGetCache.ValidatePathComponent` replacement, `--survey`,
`--dangerously-skip-checks`, or anything about homoglyphs — **you need to write your own
`ScalarPolicy`**, and that is the supported path rather than a gap. A deny list structurally
cannot catch a homoglyph: Cyrillic `а` and Latin `a` are both `Ll` and neither is a hazard. Because
the speller is total, an allow-list sink needs no hazard set of its own; supply the grammar and
everything outside it gets spelled by the same function.

`InertText.Tests/VisualEncoderTests.cs` has
`Policy_AllowShaped_CatchesAHomoglyphThatNoCategoryRuleCan` as a worked example.

**2. The API has had no adversarial review yet.** `InertString` is three commits old. My own
self-review pass on it found two real defects — a `Join` that reported a spelling its output did
not contain, and a nullable hole that bound to the generic `AppendFormatted` and double-encoded on
a live path. Both are fixed with regression tests, but treat the surface as movable and expect
review to shift it further.

### Where to start reading

1. `src/InertText/TextPolicy.cs` — the split, and why deny lists cannot catch homoglyphs.
2. `src/InertText/VisualEncoder.cs` — the contract: total, lossless, injective, invertible,
   policy-free.
3. `src/InertText/InertString.cs` — **only if you carry text to a sink** rather than just encoding
   at the point of print. This is the value form; there is deliberately no conversion *from*
   `string`, and a test asserts its absence.

Narrative version is in `docs/design/nuget-authentication.md`. The normative spec is #3601, which
I updated at `e6ccd412c` with a seventh obligation ("available as a value, not only as a returned
`string`"), so the spec and the implementation now agree.

### Gotchas that cost me time

- **`string.EnumerateRunes()` is the wrong tool here.** It substitutes `U+FFFD` for an unpaired
  surrogate, which makes the transform lossy on exactly the class `Cs` exists for. Use
  `Rune.DecodeFromUtf16` and check `OperationStatus`.
- **`IsPermitted` is not "would `Encode` change this."** A backslash is permitted by any sane
  policy but is always rewritten for invertibility. A check derived from the encoder would reject
  every Windows path.
- **`TryDecode` accepts an unpaired surrogate on purpose.** `Cs` is an encoded category, so
  `Encode` emits `\uD800` for one; refusing to decode it makes the encoder non-invertible. This
  looks like a hardening opportunity and is not one — there is a test pinning it.
- **A `InertString?` interpolation hole** needs the nullable `AppendFormatted` overload or it binds
  to the generic one and encodes twice. Fixed here, but the same trap applies to any wrapper you
  add.

Questions welcome on this PR. If you need an allow-shaped policy promoted into `TextPolicy` rather
than written locally, say so and I will add it here rather than have three of us write it
separately.

richlander:
Reviewed the `InertText` component at `26ed34d`, from the perspective of the SourceLink provenance work in #3577 — which is a prospective consumer, since its rejection messages quote artifact text throughout. Every finding below was reproduced by compiling the three files as-is and running them; nothing here is from reading alone.

The design holds up. Encoding per interpolation piece rather than over the assembled string is right, and round-trips: `Format(Field, $"name={piece} raw={"c\u0007d"}")` gives `name=a\u202Eb raw=c\^Gd`, which `TryDecode` recovers exactly. `ToString()` being unrestricted because the payload is already inert is the correct call, and the reasoning in the doc comment is the part I'd keep hardest.

### 1. Splicing across policies silently violates the type's stated invariant

`InertString`'s doc says "the invariant holds by construction: an `InertString` contains no scalar its policy refused." It doesn't, because `AppendFormatted(InertString)` trusts the incoming value without knowing which policy produced it:

```csharp
InertString prosePiece = InertString.Encode("line1\nline2", TextPolicy.Prose);   // LF permitted
InertString message = InertString.Format(TextPolicy.Field, $"repository {prosePiece} refused");

VisualEncoder.IsPermitted(message.ToString(), TextPolicy.Field)   // False
message.Forms                                                     // None
```

The `Field` message carries a raw `LF` — a scalar `Field` exists to refuse — and `Forms` reports `None`, so a sink asking "was anything encoded?" is told no and a legend derived from `Forms` won't mention it. `Join` has it too: `Join("; ", Field, [prosePiece, …])` produces the same unencoded `LF`.

This is the log-injection case the library is for: a newline reaching a line-oriented sink. And it defeats the type's own selling point — "a sink that accepts only `InertString` cannot be handed raw text by accident" — because a `Field` sink *can* be handed a `Prose` value, and nothing in the type records enough to notice.

`TryDecode` already makes the clean fix available: on splice, if `IsPermitted(piece, currentPolicy)` fails, decode the piece and re-encode under the current policy. That avoids the double-encoding trap the handler's doc comment correctly warns about, and it's the second thing invertibility buys you. Carrying the policy on the value and rejecting a stricter splice would work too; what doesn't work is trusting the piece.

### 2. `ScalarViolation` names the wrong code point for the one case it exists for

```csharp
VisualEncoder.IsPermitted("ab\uD800cd", TextPolicy.Field, out var v);
// v.ToString() == "U+FFFD (Surrogate) at 2"      actual scalar at 2 is U+D800
```

`Rune.DecodeFromUtf16` hands back `Rune.ReplacementChar` on failure, and `DecodeAt` returns it unchanged, so the violation reports U+FFFD. The pair is self-contradictory on its face — U+FFFD is `So`, not `Cs` — and the real value is lost. This is the type whose whole job is naming a scalar instead of printing it, so it's worth `IsPermitted` keeping `value[i]` for the surrogate arm.

### 3. `TextPolicy.IsNonGraphic`'s `Surrogate` arm is unreachable

Exhaustively: no valid scalar in `0..0x10FFFF` is categorized `Surrogate` — `Rune`'s invariant excludes `D800..DFFF`, so `Rune.GetUnicodeCategory` cannot return it. The arm is dead, and the doc's "`Cs` covers unpaired surrogates, which break UTF-8 conversion" describes protection this method can't provide; the real handling is `VisualEncoder.DecodeAt`'s `isUnpairedSurrogate` path, which bypasses policy entirely and says so.

Worth keeping the arm for totality if you like, but the comment should point at where surrogates are actually caught — a caller writing the allow-shaped policy the docs encourage would otherwise think this is their concern.

### 4. Canonicality is enforced in one decoder arm and not the other

The `\U` arm rejects the BMP range, commented "keeps the transform injective: a scalar has exactly one spelling". That rationale applies verbatim to the `\u` arm, which accepts spellings `Encode` never emits:

| input | decodes | re-encodes to |
| --- | --- | --- |
| `\u005C` | yes | `\\` |
| `\u001B` | yes | `\^[` |
| `\u007F` | yes | `\^?` |
| `\U0000202E` | **no** | — |

`TryDecode(Encode(x)) == x` still holds, so nothing is broken today. But the stated property is false, and if anything ever treats a successful decode as evidence that text came from `Encode`, two encodings map to one value. One condition on the `\u` arm closes it.

### 5. `Empty` and `Encode("", …)` are indistinguishable yet unequal

```
IsEmpty        True  / True
ToString()     ""    / ""      equal
Forms          None  / None    equal
GetHashCode()  equal
==             False
```

`Equals` compares `_text` directly, so `null` vs `""` diverges while every public accessor agrees and the hash agrees. Comparing `ToString()` — or normalizing in the constructor — fixes it.

### 6. Interpolation holes use two different cultures

Under `de-DE`, `$"{d}"` gives `1234,5` and `$"{d:F1}"` gives `1234.5`: `AppendFormatted<T>(T)` uses `value?.ToString()` (current culture) while `AppendFormatted<T>(T, string?)` passes `InvariantCulture`. Same message, two regimes, depending only on whether a hole happens to carry a specifier. `ScalarViolation.ToString()` already reaches for `InvariantCulture` explicitly, so the intent looks settled — the no-specifier overload should match.

---

Not blocking on my account — I'm not a required reviewer here — but 1 is worth resolving before this is depended on, because a consumer that splices a `Prose` value into a `Field` message gets exactly the outcome the library is meant to prevent, with the type system appearing to vouch for it.

Context on the consumer side, for when this lands: in #3577 I verified that `SourceLinkProvenanceResult.Reason` is dropped by all six of its callers today, so the artifact text its messages quote is latent rather than live. #3590 is the change that renders it, which makes #3590 the natural adoption point for this — I've noted that there. The one artifact-derived string that *is* rendered on that path, `RepositoryUrl`, turned out to be inert only incidentally (`Uri.AbsolutePath` leaves percent-escapes escaped), so it now has an explicit gate rather than an accident.

richlander:
## Re-review of `InertText` at `26ed34d4`

All five substantive findings from the last pass are addressed, and the shape is right now. Verified against this branch's own code rather than by reading:

```text
scalars encoded by Field: 238, roundtrip failures: 0
unpaired-surrogate roundtrip failures: 0
```

238 is exactly right: 65 `Cc` + 43 BMP `Cf` + 127 astral `Cf` + 1 `Zl` + 1 `Zp` + the always-encoded backslash. The astral hole is closed, and `TryDecode` rejecting `\U0000202E` because it names a BMP scalar is a nice catch I had not asked for — it keeps one scalar to one spelling. The `Rune.DecodeFromUtf16` comment explaining why `EnumerateRunes()` can't be used (silent U+FFFD substitution, lossy on exactly the input that matters) is the kind of note that stops this being re-broken later.

One real defect, and two minor ones.

### 1. A violation names the wrong code point for an unpaired surrogate

`Encode` handles unpaired surrogates correctly, using `value[i]` directly. `IsPermitted` does not — it reports the `scalar` that `DecodeAt` returned:

```csharp
if (isUnpairedSurrogate)
{
    violation = new ScalarViolation(i, scalar, UnicodeCategory.Surrogate);
    return false;
}
```

`Rune.DecodeFromUtf16` returns `OperationStatus.InvalidData` **and hands back `Rune.ReplacementChar`**, which the comment on `DecodeAt` already says. So `scalar` is U+FFFD, not the surrogate:

```text
input       : "A\uD83DB"
IsPermitted : False
violation   : U+FFFD (Surrogate) at 1
actual unit : U+D83D
encoder says: A\uD83DB
```

The encoder spells it `\uD83D`; the reporter says `U+FFFD`. The record is also internally contradictory — U+FFFD's category is `OtherSymbol`, so this pairs a symbol with `Category = Surrogate`.

This matters more than an ordinary off-by-one because naming the exact code point is the entire contract of the survey path. `ScalarViolation.ToString()`'s remarks are explicit that `U+XXXX` is reported precisely because it says strictly more than the character would. Here it says something false, and it says it identically for all 2,048 unpaired surrogates.

**The type is why this happened, so a local patch will not hold.** `ScalarViolation.Scalar` is a `Rune`, and `Rune` cannot represent a lone surrogate at all:

```text
Rune(0xD83D) throws: ArgumentOutOfRangeException
```

The correct answer is unrepresentable in the field meant to carry it, which is exactly why the code reached for the replacement character. So the fix is a typing decision, not a line change — carry the raw scalar value as an `int`, or give the record a second shape for the not-a-scalar case. `int` is probably simpler, and it matches the fact that the survey output is `U+XXXX` either way.

Existing tests do not catch this because the two that inspect violation *content* both use well-formed scalars (`U+13430` at line 88, the homoglyph at line 242), where `DecodeAt` succeeds. `RoundTrip_UnpairedSurrogates_RecoverExactly` covers the encoder, which was never wrong. The gap is precisely the intersection: violation reporting on the unpaired path.

### 2. `Empty` is not equal to an encoded empty string

```text
Empty == Encode("") : False   (IsEmpty True/True, hash equal True)
```

`Empty` is `default`, so `_text` is null, while `Encode("", ...)` stores `""`. `Equals` compares the fields and separates them; `GetHashCode` goes through `ToString()` and joins them. That is not an invariant violation — equal values still hash equally — but two values that both report `IsEmpty` and both render `""` comparing unequal will eventually surprise someone. Comparing `ToString()`, or normalising `_text` to `""` on construction, closes it.

### 3. The legend gate does not fail when a form is added

`Legend_NamesEveryFormTheOutputContains` hand-lists the five introducers against a fixed input. Add a sixth `VisualForm`, emit it, and forget the `DescribeLegend` arm: the test still passes, because the input produces the same five forms as before.

That is the drift the legend was designed to prevent, reappearing one level up in the gate. Deriving the expectation from `Enum.GetValues<VisualForm>()` minus `None` and asserting set equality makes a missing arm a failure — the same "declaration drives the enforcement set" shape `AGENTS.md` asks for, and the same argument the class remarks already make about legends written independently.

### Not findings

- Encoding literals in the handler, and the separate `InertString` / `InertString?` overloads, are right. The nullable one is a genuine trap — it would bind to the generic arm and re-encode.
- `Join` folding the separator's forms in only when a separator is actually emitted is a nice detail.
- `TryDecode` accepting spellings `Encode` never emits (`\u0041`, `\u001B`) makes decode a superset of the encoder's range. Round-trip is unaffected and I would leave it, but it is worth knowing that "well-formed" in the doc comment means "decodable", not "in `Encode`'s image".

Only finding 1 is worth blocking on.

richlander:
All six findings reproduced, and all six are fixed in 069037c6a. Thanks — the two of you converged on the two that mattered, and the first one was reachable from the API as shipped.

I verified every claim in code before touching anything, per the rule that has already caught a wrong finding on this PR. Nothing here needed to be argued down.

## The two blocking ones

**Policy splicing.** Confirmed exactly as described: a `Prose` value spliced into a `Field` message produced `IsPermitted(Field) == False`, `Forms == None`, and a raw LF in the output. That is log injection with the type appearing to vouch for it, which is worse than not having the type.

I took the decode-and-re-encode remedy rather than carrying the policy on the value. Carrying it would make `InertString` non-composable across sinks — the common case is legitimately building a piece for one sink and quoting it into another, and that should work, not throw. So `Format` and `Join` now route every spliced value through a `Conform` helper: if it already satisfies the policy in force it is appended byte-for-byte (no cost on the common path), and if it does not, it is decoded and re-spelled. There is a test asserting the fast path does not fire unnecessarily, because a repair on every splice would be its own bug.

You were right that this is the second thing invertibility buys, and I have added that to the spec: the requirement is literally unsatisfiable without a decoder, which is now a second argument against encode-without-a-decoder in #3601.

**`ScalarViolation` naming the wrong code point.** Confirmed, including the internal contradiction — U+FFFD is `OtherSymbol`, reported alongside `Category = Surrogate`.

Agreed this is a typing problem rather than a line fix, and I confirmed the root cause before choosing: `new Rune(0xD83D)` throws, so `Rune` structurally cannot hold a lone surrogate and the replacement character was not a lazy choice, it was forced. `Scalar` is now an `int` and the surrogate arm reports `value[i]`. The regression test asserts the reported code point agrees with the `\uXXXX` the encoder emits for the same input, which is the property that was actually broken.

## The other four, plus the one from the re-review

- **Dead `Surrogate` arm** — confirmed, 0 valid scalars categorize that way. Kept for switch totality, but the doc comment now says it is unreachable and why, instead of misdirecting the next person writing an allow-shaped policy.
- **Decoder canonicality** — confirmed `\u005C`, `\u007F`, `\u001B` all decoded despite the encoder spelling them `\\`, `\^?`, `\^[`. The `\u` arm now rejects C0, DEL and backslash. Note it deliberately still accepts `\u0041`: `Encode`s image is policy-dependent, so a restrictive policy is free to encode `A`, and only the scalars with a canonical *short* form are policy-independently redundant.
- **`Empty != Encode("")`** — confirmed; `IsEmpty`, `ToString` and `GetHashCode` all already disagreed with `Equals`. Now compared through `ToString`.
- **Culture** — confirmed `1234,5` vs `1234.5` under `de-DE`. Both hole overloads are invariant now.
- **Legend gate** (re-review) — confirmed the hand-listed introducers would not catch a sixth `VisualForm`. Derived from `Enum.GetValues<VisualForm>()` now, and it asserts the sample input still produces every form, so adding one fails here rather than passing quietly.

## Gating

11 new tests. I mutation-tested them by reverting all six fixes at once: exactly those 11 go red, so none of them is decorative. Full sweep is unchanged at 238 scalars encoded under `Field` with 0 round-trip failures, which is what I wanted to see after tightening the decoder.

Suites: InertText 46/0, NuGetFetch 293/0, Services 357/0, CLI 2632 with the 3 known local-only failures (#3592 plus two that self-analyze the test assembly).

#3601 updated in c2b433ae5 — the composition rule there stated the defective version, since I wrote it before your reviews landed.

richlander:
## Public surface changed since the orientation comment

Heads up to anyone who read [the orientation comment](https://github.com/richlander/dotnet-inspect/issues/3563#issuecomment-5138813630) or the [reconciliation](https://github.com/richlander/dotnet-inspect/issues/3563#issuecomment-5139100717) and started writing against `InertText`: both are now partly stale. Three breaking changes, all from design review after those were posted.

**`VisualEncoder` no longer exists.** Its members moved onto `InertString`, which is now a partial struct spanning `InertString.cs` and `InertString.Speller.cs`.

```diff
- string s = VisualEncoder.Encode(value, ScalarPolicy.Field, out VisualForm forms);
+ InertString s = InertString.Encode(value, ScalarPolicy.Field);
+ VisualForm forms = s.Forms;
```

`IsPermitted`, `TryDecode`, and `DescribeLegend` moved across unchanged, just to a different declaring type. The string-returning `Encode` is now `internal`, deliberately: a public method handing back treated text as a bare `string` strips the provenance the value type exists to carry, and it had no production callers outside `InertString`.

`ScalarPolicy`, `ScalarViolation`, and `VisualForm` are untouched and still live where they did.

**`ScalarViolation.Scalar` is `int`, not `Rune`.** This is the fix for the blocking finding that every unpaired surrogate was reported as U+FFFD. `Rune` cannot represent a lone surrogate — `new Rune(0xD83D)` throws — so the substitution was structural, not sloppy. If you were reading `.Scalar.Value`, drop the `.Value`.

**`InertString.Empty` is now a constructed value rather than `default`.** No source change needed, but the contract is now stated rather than inherited from how the CLR zeroes memory: empty text, `VisualForm.None`. `default(InertString)` remains reachable and remains equal to it — a reflection test compares all three empty states pairwise across the whole public surface.

Everything from the six-fix commit still stands: `Conform` on splice, the `\u` injectivity tightening, invariant culture in the interpolation holes.

Sorry for the churn. The reviews are what surfaced that the type was overselling itself, and the design pass that followed is what shrank it. Better now than after you had built on it.

richlander:
Now that #3380 has merged, some field evidence for this component from the adjacent work — including a concrete first consumer and one API-shaping conflict.

## A concrete first consumer

#3628: `@Metadata` table cells reach stdout carrying raw `U+202E`. The cause is a fourth hand-rolled hazard set:

```csharp
// MetadataTableProjector.cs:1439
static bool IsControl(char c) => c < ' ' || c == '\x7f' || (c >= '\x80' && c <= '\x9f');
```

Holding that fix for this PR rather than adding a fifth escaper — a `ScalarPolicy` denying `Cc | Cf | Cs | Zl | Zp` closes it by category instead of by hand-enumerated range.

## The count is the argument

The codebase now carries **four** independent "what is dangerous" predicates, each correct about a different subset:

| Predicate | Cc | Cf (bidi) | Zl/Zp |
| --- | --- | --- | --- |
| `CSharpIdentifierCore.IsRenderingHazard` | yes | yes | yes |
| `InspectionResultView.ContainProse` | yes (not `\n\r\t`) | yes | yes |
| `MetadataTableProjector.IsControl` | yes | **no** | **no** |
| `ApiSurfaceExtractor` (~1927) | yes | yes | — |

Three of four are right; the fourth is a live leak, and nothing in the build knew the fourth existed. That is the empirical version of #3601's claim that a sanitizer "is silent when wrong" — the silence here lasted from the predicate's introduction until someone happened to point a hostile assembly at that one lens.

## One conflict worth resolving before this lands

`CSharpIdentifierCore.ContainComposedName` — which #3380 spread across the C# lens — opens with:

```csharp
name.ReplaceLineEndings(" ")
```

That is **lossy and non-injective** by inspection: `"A\nB"` and `"A B"` both become `"A B"`. #3601 requires encoding that is unique and invertible, so these helpers are a **prospective `VisualEncoder` consumer, not a peer**. They should not be treated as an existing implementation of this contract, and the eventual migration is a behavior change at the C# lens (fold → encode), not a refactor. Worth stating explicitly somewhere so nobody reads the two as already-aligned.

## `InertString` vs. the analyzer

#3380 ended up at the same destination from the other direction. After the hand-rolled ownership gate proved unclosable, it was replaced with `BannedApiAnalyzers` banning `Console.Error` outright — the *analyzer-enforced* form of the guarantee `InertString` gives at the *type* level.

Having built the analyzer version, the type version is clearly stronger. The analyzer needs a per-project opt-out list (`OwnsItsOwnStderr`, currently five projects), and each entry is an unenforced hole. A sink typed `InertString` needs no list, because nothing else typechecks. Where a sink can be retyped, prefer that over banning the API.

richlander:
**Restacked, and the surface changed again — this supersedes comments 5138813630, 5139100717 and 5140000362, all of which are now stale.**

## This PR is now the NuGet work only

InertText moved to #3636 and this PR is based on it. The diff here is 16 files, none of them InertText. If you are consuming the primitive, read #3636; if you are consuming the feed-failure reporting, read this.

## Current API

`VisualEncoder` is **back**, but in namespace `InertText.Encoder`:

```csharp
using InertText;         // currency: InertString, ScalarPolicy, TextPolicy, VisualForm
using InertText.Encoder; // capability: VisualEncoder.Encode / TryDecode

InertString inert = new(value, TextPolicy.Field);   // was InertString.Encode(value, policy)
InertString msg   = InertString.Format(TextPolicy.Field, $"...{inert}...");
string spelled    = inert.ToString();
```

| was | now |
| --- | --- |
| `InertString.Encode(v, p)` | `new InertString(v, p)` |
| `InertString.TryDecode(...)` | `VisualEncoder.TryDecode(...)`, namespace `InertText.Encoder` |
| `VisualEncoder` (removed in an earlier round) | restored, in `InertText.Encoder` |

**You almost certainly do not need the second using.** Producing inert text never requires naming the capability namespace — that is now gated, and it is the whole point of the public constructor. If you find yourself importing `InertText.Encoder` in production code, that is the signal to stop and ask why you need to recover the original text.

Sorry for the churn. The shape is settled now: three mutation-proved gates hold it, so a further move would have to turn a test red.

Part of #3635.

richlander:
**Correction to the comment above.** I wrote that the audit is "no `using InertText.Encoder`". That search string is wrong in a way that fails open, and it is worth stating precisely because it is the whole point of the namespace split.

A using directive is one of *two* ways to reach a namespace. The other leaves the import block untouched:

```csharp
using InertText;                    // the only directive in the file

InertText.Encoder.VisualEncoder.TryDecode(inert.ToString(), out string? original);
```

That compiles and recovers the original. Anyone adding a decode deliberately has no reason to touch the imports, so this is the shape it would actually take — and a reviewer greping for the directive would see a clean import list and conclude the opposite of the truth.

**Grep the bare namespace: `InertText.Encoder`.** It catches both forms, because a fully-qualified call has to spell the namespace too. There is no third way in.

Fixed in the docs and test remarks on #3636 (39e284486).

Issue #3547, opened 2026-07-30
The section model from library and package is the one we want tool-wide, and main has drifted from it


richlander:
#3229 worked this model out on the `library` command. #3448 replayed it on
`package`, which is where it stopped being one command's layout and started being
a model — the second application is what proved the axes were real and not just a
description of `library`. **We want this model everywhere in the tool.**

Two adversarial review rounds on #3448 produced findings that arrived after the
merge, so the design we settled on and the code on `main` have drifted apart in
four places. This issue records the model, shows where `main` diverges with
reproductions from a build of `origin/main`, names the commands still outside it,
and folds in the ideas worth keeping from the #3429 stack before it is abandoned.

## The model

**Sections are named for kinds of thing, not for layout.** The package file family
is `Package README file`, `Package nuspec file`, and `Package skill files` — three
*disjoint* document kinds behind the `@Files` door, plus `Package files` as the
unfiltered superset that is deliberately not a member. Layout roots (`lib/`,
`ref/`, `runtimes/`) are slices of one listing, reached with `--path "lib/**"`,
not sections of their own. Singular names yield at most one row.

**A section shows only when it has rows.** Zero rows means the section does not
render. This is why `SizeClass.Fixed` cannot promise a constant *set* — it
promises the row set does not grow with the target. Absence always means "not
applicable", never "too long for this package".

**Two declared axes, not ad-hoc booleans.** `SizeClass` is row-count growth;
`Cost` is latency/network. Bare `-S` is `Fixed && NetworkFree`.

**Discovery describes exactly what you can ask for.** Every name `-D` lists must
resolve, and every name that resolves should be discoverable. The package command
therefore drops the computed `@All` and `@Default` poles entirely: sections are
reachable by name, by topical door, and by verbosity, and a pole that renders a
superset nobody asked for is a surface no discovery output describes.

**Internal encodings never reach the user.** Bare `-S` is encoded as `@Default` on
the wire, but that is an implementation detail of the parser.

## Where `main` does not satisfy it

All six reproduce against `origin/main` (`59f731a1`) built at
`artifacts/bin/dotnet-inspect/release/dotnet-inspect.dll`.

### 1. Help advertises a selector the package command rejects

`SharedOptions.cs:123` is shared across commands and still names `@All`, which no
longer resolves on `package`. The help text points users at a failing command.

```console
$ dotnet-inspect package --help | grep -F -- '--select'
[...]

PR #3566, opened 2026-07-30
Give bare -S its own marker instead of the string "@Default" (#3547)

richlander:
Bare `-S` no longer travels as the selector string `@Default`, which fixes two defects in #3547 and unblocks removing the poles.

**Stack**: slice 1 of 2 for #3547. Parent: `main`. Next slice removes `@All`/`@Default` outright (opens against this branch).

## What was wrong

Bare `-S` asks for a command's default preset. It was *encoded* as the literal selector value `@Default`, so the marker and a hand-typed pole were indistinguishable to every consumer downstream. Two user-visible defects follow directly:

- **Gap 2** — combining bare `-S` with anything that contributes its own selector pushed the internal encoding through resolution and leaked it:
  ```console
  $ dotnet-inspect package Markout -S --path '*.nuspec' --paths
  Warning: Select value '@Default' not found.
  markout.nuspec
  ```
- **Gap 3** — `package -S @Default` still resolved despite `.WithoutComputedPoles()`, because the command short-circuited on the encoding before the category map was ever consulted. `-S @All` errored; `-S @Default` rendered five sections.

## What changed

Bare `-S` travels as a `SelectDefault` flag on the options records. `ResolveSelectAsSections` contributes the info sections for it directly rather than through a `@Default` category entry, so it resolves the same way on pipelines that publish no poles.

Sites that inferred "the user asked for sections" from `Select` being non-empty now consult the flag too — tip suppression, `HasSectionQuery`, `SelectExplicitlySet`, and the performance-triage `hasExplicitSelect` gates. Without that, bare `-S` would have silently stopped counting as a selection (this is what a first pass got wrong; caught by output diffing, see below).

## Behavioral claim

`@Default` is now only a category, wherever the category map still publishes one.

| Command | Before | After |
| --- | --- | --- |
| `package -S @Default` | renders 5 sections | `Error: Select value '@Default' not found.` |
| `package -S --path '*.nuspec' --paths` | spurious `@Default` warning | no warning, same rows |
| `library -S @Default` | Library Info + Signals + Symbols | Library Info |
| `library -S` (bare) | Library Info + Signals + Symbols | unchanged |
| `package -S -D` | `Error: Select value '@Default' not found.` | lists the sections |
| `package X --library -S` | Library Info + Signals + Symbols | unchanged |
| `timeline -S` | `Error: Unknown timeline section '@Default'.` | `Error: Bare -S has no fixed sections for timeline.` |

The `package -S -D` row is the same defect as the two above it, not a separate decision: `-S <name> -D` has always worked and is byte-identical before and after, so only the *bare* form failed, and only because the marker was a string `package` does not publish after `WithoutComputedPoles()`. It clears when the encoding does.

The last two rows are review findings, both instances of one class — *code that used to receive the marker for free inside `Select` now receives nothing*. `package X --library -S` is restored to its previous output; `timeline -S` keeps refusing, since both timeline sections grow with the version range and bare `-S` denotes the fixed/bounded, non-network sections. Only the message changes, to stop spelling an internal marker at the user.

The `library -S @Default` row is the intended divergence: bare `-S` asks for the broader network-free fixed overview, the pole resolves to exactly its members. They were previously the same request only because of the encoding. Slice 2 deletes the pole.

## Evidence

Output-diffed 10 invocations covering every command that consumes `-S`, before and after, at `Release`. Seven are byte-identical; the three that differ are the table above. That diff is what caught the tips/projection/triage regressions — each one showed up as an unexpected `DIFF` and was fixed before this PR existed.

Gates, all three verified non-vacuous by reverting `src/dotnet-inspect/` while keeping the tests (`3 Failed`):

- `Package_ComputedPoles_AreNotResolvable` — extended to `-S @Default` alone
- `Package_BareSelect_CombinedWithPathSugar_DoesNotLeakThePresetMarker` — the leak
- `LibraryCommand_BareSelect_IsNotTheSameRequestAsTheDefaultPole` — the intended divergence

Added from review, each verified non-vacuous the same way:

- `PackageLibraryMode_BareSelect_MatchesStandaloneLibraryBareSelect` — asserts the nested and standalone heading lists are *equal*, rather than pinning a literal set
- `Timeline_BareSelect_IsRefusedWithoutLeakingTheMarker`
- `Package_BareSelectWithDiscover_ListsSectionsRatherThanFailing`
- `ApiShim_BareSelect_ForwardsThePresetToTypeOptions` — calls `ApiCommand.ExecuteAsync` directly, since that arm is unreachable from the CLI

The sweep for that defect class is closed by construction: the only two sites that copy `Select` into another options record both copy `SelectDefault`, and every `ParseSelect` call site has a matching `SelectDefault` assignment. A confirmation round independently checked the shapes greps miss — `with` expressions, record inheritance, and serialization paths.

Full CLI suite: `2587 total, 0 failed, 14 skipped` (skips are `ilasm`/native-PE/Windows-only).

## Residual

Both poles remain resolvable wherever the category map publishes them, and `--select`'s help text still advertises `@All` on commands that reject it (gap 1). `docs/design/info-view.md`'s stale five-section claim (gap 4), the `-S @Files --json` omission (gap 5), and `-S --count` (gap 6) are untouched. Slice 2 takes the poles and the help text.


richlander:
## Review round 1 reconciliation (GPT-5.6)

Reviewed at `4a11f7d94`. Four blocking findings; **two accepted and fixed, one accepted as intended and characterized, one dismissed**. Resolution commit: `57fc574af`.

### Accepted and fixed

**1. Nested `package X --library -S` dropped bare `-S`.** `CreateLibraryOptions` copied `Select` but not `SelectDefault`, so the nested view silently degraded to "no sections requested" and rendered less than the equivalent `library X -S`. Confirmed against a base build of `fdf24f5f7`:

```text
BASE  package Markout --library -S  ->  ## Library Info / ## Signals / ## Symbols
HEAD@4a11f7d94                      ->  ## Library Info
HEAD@57fc574af                      ->  ## Library Info / ## Signals / ## Symbols
```

Test `PackageLibraryMode_BareSelect_MatchesStandaloneLibraryBareSelect` asserts the nested and standalone heading lists are *equal* rather than pinning a literal set, so it keeps holding as the library view evolves.

**3. `timeline -S` went from a refusal to being silently ignored.** Confirmed base vs head:

```text
BASE  timeline 'Markout@0.29.0..0.33.0' Markout.MarkoutWriter -S
      Error: Unknown timeline section '@Default'. Use Evaluations or Transitions.
HEAD@4a11f7d94  (bare -S silently ignored, renders the default view)
```

`TimelineOptions` now carries `SelectDefault` and `ResolveSections` refuses it. Both timeline sections grow with the version range, so there is no fixed/bounded subset for bare `-S` to denote — the refusal is the right answer, and it is what base did. What changes is that the message no longer spells an internal marker at the user:

```text
HEAD@57fc574af
      Error: Bare -S has no fixed sections for timeline. Use -S Evaluations or -S Transitions.
```

Test: `Timeline_BareSelect_IsRefusedWithoutLeakingTheMarker`.

Both findings are the same shape, and it is the shape worth naming: **code that used to receive the bare-`-S` marker for free inside `Select` now receives nothing.** That is the cost of the decouple, and it is exactly what round 1 was for.

### Accepted as intended, now characterized

**4. `package X -S -D` now succeeds where base failed.** Not a separate behavior decision. `-S <name> -D` is a supported combination and is byte-identical on base and head:

```text
BASE and HEAD  package Markout -S 'Package Info' -D  ->  | Package Info | section |
```

Only the *bare* form failed on base, and only because the marker was the literal string `@Default`, which `package` does not publish after `WithoutComputedPoles()`. That is the same defect as declared changes 1 and 2, so it clears with them. Added `Package_BareSelectWithDiscover_ListsSectionsRatherThanFailing` so the behavior is pinned rather than incidental, and the PR body now lists it as a fourth declared change.

### Dismissed

**2. Dotted generic member selectors lose the preset.** Does not reproduce. The repro compared `member 'T.GenericChoice<T>' ... -S` against a *different invocation form* (`member -m ...`), not against base. Built `fdf24f5f7` and ran the exact repro on both:

```text
BASE  # DotnetInspector.Tests.MemberGenericSelectorFixture   EXIT:0
HEAD  # DotnetInspector.Tests.MemberGenericSelectorFixture   EXIT:0
```

Identical. The dotted form takes the `lookupResult.ImpliedMember` branch, so the `else if` re-resolution branch cited at `MemberCommand.cs:105` is never reached — on base either. The suggested gate change (`|| options.SelectDefault`) was applied and measured **vacuous**: `member <TypeOnly> --library X -S` renders `## Constructors` + `## Method Groups` identically with and without it. Reverted rather than kept as a speculative no-op.

There may still be a real latent issue in that re-resolution branch, but it predates this PR and is not part of its blast radius.

### Validation

- Full CLI suite at `57fc574af`: **2586 total, 0 failed, 14 skipped**.
- Non-vacuity: reverted the uncommitted product fixes with the test changes in place — **2 of 3 new tests fail**. The third (`Package_BareSelectWithDiscover_...`) passes without them by design; it characterizes behavior that arrived with this PR's already-committed changes.
- The three originally declared behavior changes re-verified at the new head.

A confirmation round at `57fc574af` is running, with an explicit brief to sweep for further instances of the defect class above.

richlander:
## Review round 2 reconciliation (Gemini 3.1 Pro, differential) + round 1 follow-on fix

Two more rounds landed. Head is now `76b7f2ea3`.

### Gemini 3.1 Pro — differential matrix at `4a11f7d94`

Assigned differential testing rather than code reading: build base and head, run a matrix across every `-S`-consuming command (`package`, `library`, `type`, `member`, `api`, `project`, `diff`, `timeline`) crossed with bare `-S`, `-S @Default`, `-S @All`, an exact section, a glob, `--tips d`, `--count`, `-v:q`/`-v:m`, plus `--path` and `--library`.

**It independently reproduced both real defects that GPT-5.6 found** — the nested `package --library -S` regression and the `timeline -S` change — arriving via a different method. Two reviewers converging on the same two findings, from code reading and from output diffing respectively, is good evidence the blast radius is now understood rather than merely surveyed.

Its confirmations are worth recording, because they are negative results that cost real work to establish:

- **The tip-suppression patches were exactly right and exactly enough.** It checked `type`, `member`, `project`, and `find` for the same `Select != null` gate that `PackageOptionsParser` and `DiffOptionsParser` had. Their `TipLevel` logic is not gated on `Select`, so bare `-S` never suppressed tips there on base and still does not on head.
- `-S` with `--count` identical on base and head, including that `package X -S --count` still fails with `--count requires -S/--select to match at least one section.`
- The `api` deprecation-banner short-circuit is diff-free.
- `ResolveSelectAsSections` with `selectDefault: true` and no info sections returns `Sections = null` / `Unresolved = []` cleanly, and callers render their default view rather than erroring.

**One divergence from its advice, deliberately.** Its fix had two steps; I applied step 1 (`SelectDefault` in `CreateLibraryOptions`) and not step 2 (stop passing `SelectDefault` to the nested `ResolveSelectAsSections`). After step 1 alone, all three spellings agree with base:

```text
BASE / NEW  package Markout --library -S              ->  ## Library Info ## Signals ## Symbols
BASE / NEW  package Markout --library Markout.dll -S  ->  ## Library Info ## Signals ## Symbols
BASE / NEW  library Markout -S                        ->  ## Library Info ## Signals ## Symbols
```

Step 2 appears premised on the unfixed state. I have asked Gemini to attack that specifically at the new head, since its own analysis predicted one section where three are observed — so either its reading of `InfoSectionNames` was off, or that resolve is not the path these invocations take. I would rather have that discrepancy explained than leave it as "the output looks right".

It also marked `timeline -S` **non-blocking** and suggested rendering the default set might be more correct, where GPT marked it blocking. I refused it instead — see below.

### Round 1 follow-on: a third instance of the same defect class

GPT's confirmation round at `57fc574af` swept for more instances of *"code that used to receive the bare-`-S` marker for free inside `Select` now receives nothing"* and found one:

**`ApiCommand.ExecuteAsync`'s `_ =>` compatibility arm** rebuilds a `TypeOptions` field by field and copied `Select` but not `SelectDefault`, silently downgrading bare `-S` to the default tree view. Fixed in `76b7f2ea3`.

Not reachable from the CLI — the CLI always constructs `TypeOptions` or `MemberOptions`, which match the earlier switch arms — so this affects the direct callers the shim exists to serve. `ApiShim_BareSelect_ForwardsThePresetToTypeOptions` calls `ApiCommand.ExecuteAsync` directly the way they do. Non-vacuity confirmed: reverting the one product line fails the test.

**The sweep is now closed**, by construction rather than by inspection:

- Only two sites in the product copy `Select` into another options record (`PackageCommand.cs:2288`, `ApiCommand.cs:53`). Both now copy `SelectDefault`.
- Every file calling `ParseSelect` also calls `ParseSelectDefault`, with matching per-call-site counts.

### Why `timeline -S` refuses rather than renders

The two reviewers split on this, so it needed a tiebreak from intent rather than from taste. The product owner has since stated the forward contract for bare `-S`: *print all the fixed/bounded, non-network-required sections*, where **bounded** means the section's length does not vary with the target being inspected.

Both timeline sections grow with the version range. There is no bounded subset for bare `-S` to denote on that command, so refusing is the correct answer and matches what base did. What changes is only that the message no longer spells an internal marker at the user.

Worth noting this contract is already implemented for `package` and `library` — `SectionPipeline.IsRequested` gates bare `-S` on `SizeClass == Fixed && Cost == NetworkFree`. It is *not* honored by `type`/`member`/`api`/`project`/`diff`, which still resolve bare `-S` to the `Info` flag (the `-v:m` target-section axis, unrelated to boundedness or network cost). Measured: `type -S` returns 15/103/47 rows across three targets, while `library -S` returns 50/52/54 in a constant 3 sections. That gap is follow-up work, not this PR's.

### Validation at `76b7f2ea3`

Full CLI suite: **2587 total, 0 failed, 14 skipped**.

Confirmation rounds are running at the new head with both reviewers — GPT-5.6 to verify the sweep is exhaustive against shapes its greps would miss, and Gemini to re-run its matrix and attack the step-2 discrepancy above. Not marking ready until both are clean.

richlander:
## Confirmation round clean (GPT-5.6 at `76b7f2ea3`)

The confirmation round at the current head reports **no significant issues**, and it confirmed both claims I asked it to attack rather than accept:

- **The `ApiCommand` fallback arm is unreachable from the CLI** — verified independently, so the fix is correctly scoped to direct callers of the shim and the test correctly exercises `ApiCommand.ExecuteAsync` directly.
- **The sweep is exhaustive.** I asked it specifically to look for *shapes* my greps would miss rather than re-run my greps. It checked `Select` forwarding, record construction, `with` expressions, record inheritance, parser pairings, and serialization/round-trip paths. Nothing further found.

Also reproduced independently:

- Full CLI suite: **2587 passed, 0 failed, 14 skipped**.
- `package`, `library`, nested-library, `timeline`, discovery, path-sugar, and dotted-generic behavior all rechecked against base `fdf24f5f7`; all expected results intact.

### Where this leaves the PR

|---|---|---|---|
| 1 | GPT-5.6 (code reading) | `4a11f7d94` | 4 blocking → 2 fixed, 1 accepted+tested, 1 dismissed |
| 2 | Gemini 3.1 Pro (differential) | `4a11f7d94` | 1 blocking (same defect, found independently), 1 non-blocking |
| 3 | GPT-5.6 (confirmation + sweep) | `57fc574af` | 1 blocking (`ApiCommand` shim) → fixed |
| 4 | GPT-5.6 (confirmation) | `76b7f2ea3` | **Clean** |
| 5 | Gemini 3.1 Pro (confirmation) | `76b7f2ea3` | running |

Two model families, both of which found real defects, and the last round from each is the gate. Awaiting Gemini's confirmation — it is specifically re-attacking the one place I declined its advice (the nested-library step 2), so I would rather have that answered than assume the passing output settles it.

richlander:
## Ready to merge

Both model families' final rounds are clean at `76b7f2ea3`, and CI is green.

### Gemini 3.1 Pro confirmation — clean

It re-ran its full 30+ invocation matrix across every `-S`-consuming command against base `fdf24f5f7` and head `76b7f2ea3`. **The only diffs are the five intended ones**; everything else — tip suppression, `--count` validation, `-v:q`/`-v:m` — is byte-for-byte identical.

It also answered the two things I asked it to attack rather than accept, and the answers were worth the round:

**On the nested-library step 2 I declined:** it retracted the analysis and identified its own error. `--library` routes through `ExecutePackageLibraryAsync` → `LibraryCommand.ExecuteAsync`, which consumes the marker and sets `FixedOverview = true`. Step 2 was unnecessary for that path, exactly as the output suggested.

**But it found the variant I had not exercised** — and this is the useful part. `package --all-libraries -S` renders only `Library Info`, dropping `Signals` and `Symbols`, because `ExecutePackageAllLibrariesAsync` does *not* delegate to `LibraryCommand`; it resolves sections itself and bypasses `FixedOverview`. It then checked base before calling it a regression, and it is not one:

```text
BASE and HEAD  package Markout --all-libraries -S  ->  ## Library Info (lib/net10.0/Markout.dll)
BASE and HEAD  package Markout --library -S        ->  ## Library Info ## Signals ## Symbols
```

I reproduced this independently. Pre-existing, out of this PR's blast radius, and filed as **#3573** — it is the same "two implementations of what bare `-S` selects" split that #3547's later slices exist to close, so it should fall out of that work rather than needing a separate fix.

**On the `ApiCommand` shim:** confirmed unreachable from the CLI, and with a sharper reason than mine — the `api` command is wired to `CreateDeprecatedApiCommand()`, which prints a deprecation banner and returns `1` without ever constructing an `ApiOptions`. So the shim serves direct callers only, and the test correctly exercises it that way.

It also confirmed the timeline refusal, `-S Evaluations`/`-S Transitions`/no-`-S` all byte-identical to base, `package -S -D`, and the suite at 2587/0/14.

### Round ledger

|---|---|---|---|
| 1 | GPT-5.6 (code reading) | `4a11f7d94` | 4 blocking → 2 fixed, 1 accepted+tested, 1 dismissed |
| 2 | Gemini 3.1 Pro (differential) | `4a11f7d94` | 1 blocking (same defect, independently), 1 non-blocking |
| 3 | GPT-5.6 (confirmation + sweep) | `57fc574af` | 1 blocking (`ApiCommand` shim) → fixed |
| 4 | GPT-5.6 (confirmation) | `76b7f2ea3` | **clean** |
| 5 | Gemini 3.1 Pro (confirmation) | `76b7f2ea3` | **clean** |

Two families, both of which found real defects that the other's method surfaced differently — code reading found the shim, differential testing found `--all-libraries`. Three defects fixed, all one class: *code that used to receive the bare-`-S` marker for free inside `Select` now receives nothing*. The sweep for that class is closed by construction, not by inspection.

### CI at `76b7f2ea3`

`build-net10` pass · `decompiler-gates` pass · `test (ubuntu-26.04, linux-x64)` pass · `changes` pass. The four `skipping` jobs are path-filtered and terminal; none is cited as evidence.

### Follow-up, non-blocking

- **#3573** — `package --all-libraries -S` bypasses `FixedOverview` (pre-existing).
- Slice 2 (#3567) removes `@Default`; it is restacked on this head and awaiting its own rounds.
- The `--select` help text still advertises `@All` (gap 1), and gaps 4–6 of #3547 are untouched. Slice 3 takes those.

Ready to merge.

PR #3567, opened 2026-07-30
Remove the @Default section pole (#3547)

richlander:
`@Default` is gone. It was a computed pole that restated what bare `-S` already means.

**Stack**: slice 2 of 3 for #3547. Parent: #3566 (`decouple-bare-select`). Next slice removes `@All` and fixes the `--select` help text.

> Review this against its parent. The diff shown here is this slice only; #3566 must land first.

## What changed

Slice 1 gave bare `-S` its own marker, which left `@Default` with no job. This drops it from every pipeline's category map and deletes the constants, predicates, and fallback entries that published it:

- `SectionPipeline.DefaultCategory` and `SelectResolver.InfoSelector` / `IsInfoSelector`
- the `@Default` entry in `GetCategoryMap` and in `BuildFallbackCategories` (whose now-unused `infoSections` parameter goes with it)
- `SelectResolver.IsActiveInfoSelector` loses its selector-array parameter — the preset is reachable only through bare `-S`, so there is no selector list to inspect
- `MemberCommand`'s pure-selector check and `ProjectCommand`'s category map lose their `@Default` entries

`WithoutComputedPoles` now drops only `@All`; its doc comment says so.

## Behavioral claim

`-S @Default` reports `Select value '@Default' not found.` on every command, exactly as `package` already did for `@All`.

**Bare `-S` is unchanged everywhere.** Verified per command against the pre-change build:

| Command | bare `-S` renders |
| --- | --- |
| `package Markout` | Manifest, Package Info, Package nuspec file, Package README file, Signature |
| `library System.Text.Json` | Library Info, Signals, Symbols |
| `type …JsonSerializer` | Properties, Method Groups |
| `project …csproj` | unchanged mode error |

`library -S @All` still resolves — that is slice 3's job.

## Evidence

Full CLI suite: `2584 total, 0 failed, 14 skipped`.

Gates:

- `LibraryCommand_DefaultPole_IsNotResolvable` replaces the slice 1 test that pinned the pole's divergence from bare `-S`; it now asserts the pole errors while bare `-S` still renders the full fixed overview
- `ResolveSelect_DefaultPoleSpelledOut_DoesNotResolve` pairs with `ResolveSelect_SelectDefaultFlag_ReturnsDefaultSections`, so the flag path and the retired spelling are both covered
- `Package_ComputedPoles_AreNotResolvable` keeps its slice 1 coverage of `-S @Default` alone

Docs: `docs/design/section-model.md` and `docs/design/info-view.md` updated; no `@Default` references remain under `docs/`. `markdownlint` clean on both.

## Residual

`@All` is still resolvable on every pipeline that has not called `WithoutComputedPoles`, and `--select`'s help text still advertises it on `package`, which rejects it (gap 1). Slice 3 takes both. It is a larger change because `@All` has real consumers beyond resolution — `GetAllSelectorSections` and the `MarkdownSectionOrderer` path — which is why it is not folded in here.

Gaps 4, 5, and 6 of #3547 remain untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -449,7 +449,7 @@ dotnet-inspect type JsonSerializer --platform System.Text.Json -S "Source Files"
 dotnet-inspect type JsonSerializer --platform System.Text.Json -S "Source Files" --print --row 1
 ```
 
-For target-based queries, `-D` reports the effective schema by default: only sections and columns that can actually render for that query. Add `--schema` for the static schema. Bare `-S` renders `@Default`, a curated high-density view; type/member summaries use `Method Groups`, while `member Type -m Name` uses `Methods` overload rows. Lists for `-S`, `--columns`, and `--fields` accept commas or semicolons. Use `-S @All` to select all sections; it renders the default section first, then remaining sections alphabetically. Workflow categories such as `@Source` and `@Audit` expand to scenario-focused section groups.
+For target-based queries, `-D` reports the effective schema by default: only sections and columns that can actually render for that query. Add `--schema` for the static schema. Bare `-S` renders a curated high-density default view; type/member summaries use `Method Groups`, while `member Type -m Name` uses `Methods` overload rows. Lists for `-S`, `--columns`, and `--fields` accept commas or semicolons. Use `-S @All` to select all sections; it renders the default section first, then remaining sections alphabetically. Workflow categories such as `@Source` and `@Audit` expand to scenario-focused section groups.
 
 ## Common examples
 
--- a/docs/design/info-view.md
+++ b/docs/design/info-view.md
@@ -52,6 +52,6 @@ Some commands or contexts should not get a default preset until there is a clear
 
 When adding or changing a default preset, first write the bullseye question. Then select the smallest section set that answers it. If the question needs more than three sections, the preset is probably not focused enough.
 
-Avoid adding a section just because it is cheap. `@Default` is not "all safe sections"; it is the smallest high-value bundle.
+Avoid adding a section just because it is cheap. The default preset is not "all safe sections"; it is the smallest high-value bundle.
 
 For library defaults, large local metadata lists such as `Async Methods`, `Custom Attributes`, `Extension Methods`, `Resources`, and `Type Forwarders` are intentionally represented as counts in `Library Info`. Select those sections explicitly when the list itself is the answer.
--- a/docs/design/section-model.md
+++ b/docs/design/section-model.md
@@ -154,8 +154,9 @@ requires `-v:m`; every other bounded network-free section first renders at
 
 ### Bare `-S` — the network-free fixed overview
 
-Bare `-S` (the `-S` flag with no value, which parses to the lone `@Default`
-preset) is the ergonomic **network-free fixed overview**. Instead of the `-v:m`
+Bare `-S` (the `-S` flag with no value) is the ergonomic **network-free fixed
+overview**. It is a request for the command's default preset, carried as its own
+flag rather than as a selector value, so it has no spelling a caller can type. Instead of the `-v:m`
 target-only view, the library command renders only the sections whose declared
 growth class is `Fixed` and whose cost is `NetworkFree` — today the `Library
 Info`, `Signals`, and `Symbols` fact tables. Because membership is a function of
@@ -279,7 +280,7 @@ count > 0 and drops categories that become empty. `--schema` is the exhaustive
 escape hatch — the full section graph plus every topical door, still without
 annotations. `@All`/`@Hidden` are internal computed poles: `@All` is the flat
 standalone set, `@Hidden` is its complement, and neither is a user-facing door
-(`@Default`/`@All`/`@Hidden`/`@Switches` never appear in `-D`).
+(`@All`/`@Hidden`/`@Switches` never appear in `-D`).
 
 ### Invariants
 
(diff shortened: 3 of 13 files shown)


richlander:
## Review round reconciliation (MAI-Code) + restack

### Review: clean

Reviewed at `4f3592a60`. **No significant issues found.** The reviewer built both the parent commit and the reviewed head and diffed real output rather than reading only:

- Bare `-S` byte-for-byte identical to the parent for `library System.Text.Json -S` and `package Markout -S` (empty `diff -u` for both).
- `-S @Default` now errors; `-S @All` still resolves; `type`/`member`/`project` bare `-S` unchanged.
- `-D` and `--schema` on both `library` and `package` no longer expose `@Default` anywhere.
- Full suite: 2584 total, 0 failed, 14 skipped.
- **Non-vacuity independently verified**: reintroduced `@Default` into `SelectResolver`'s fallback categories in a scratch copy and confirmed the targeted test fails.

### Restacked

Slice 1 (#3566) took a review fix, so this slice was rebased onto its new head. `range-diff` confirms the base moved and the commit did not:

```text
$ git range-diff 4a11f7d94..4f3592a60 57fc574af..HEAD
1:  4f3592a60 = 1:  5b7d2c09f Remove the @Default section pole (#3547)
```

`=` on the only row: identical content, new base. New head is `5b7d2c09f`.

Revalidated after the restack — full CLI suite **2587 total, 0 failed, 14 skipped**. (One run showed a single failure in `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which also fails on unmodified `main` here because the local NuGet cache lowercases `newtonsoft.json.nuspec`; it passed on re-run and is unrelated to this slice.)

The restack changed the head, so this slice is not being marked ready on the strength of the round above. It needs a clean read at `5b7d2c09f` — which, given the `=` range-diff, is a confirmation rather than a second full pass. Holding until #3566's confirmation round lands, since anything that moves #3566 again cascades here.

richlander:
## Review reconciliation — round 3 (Gemini 3.1 Pro, second family)

Reviewed at `ebb33bcda` in an isolated worktree. Result: **clean on product code**, one non-blocking documentation finding, now fixed.

### Verified by the reviewer

- **Differential execution matrix** across `package`, `library`, `type`, `member`, `api`, `project`, `diff`, `timeline`, `find` for `-S`, `-S @Default`, `-S @All`, `-S <exact>`, `-D`, `-D @All`, `--schema`. The only base-vs-head changes were the intended `exit 1` refusals when `@Default` is spelled out. `@All` and bare `-S` keep exact parity with the parent.
- **Honest error paths** — `SelectResolver.ResolveSelectAsSections` returns a `SelectMiss` rather than a null section list, so `PackageCommand` bounces the request through `SelectOutput.WriteUnresolved` instead of succeeding with an empty table. That is the "no success-shaped empty output" property this PR had to preserve.
- **Independent non-vacuity** — the reviewer re-injected the `@Default` pole into `SectionPipeline`'s map and confirmed the new tests catch the regression. That check was run by the reviewer, not by me.
- **`BuildFallbackCategories`** — dropping the `infoSections` parameter is inert: every product caller supplies its own category map, and the internal fallback is reached only from `SectionRegistrySpike` and unit tests.

### Finding: `@Default` still named in documentation — fixed in 5adedb216

Non-blocking, and correct. `README.md:452` still told users bare `-S` "renders `@Default`" — a selector that now exits 1.

Sweeping for the same shape turned up two more, which I fixed in the same commit because they describe a **live** contract in terms of a deleted name:

- `SectionPipeline.InfoSectionNames` and the `-D` annotation legend — "the curated @Default preset"
- `ISectionDescriptor.Info` — same wording, and it is the contract a reader consults when deciding whether to set the flag

All three now say "the curated default preset", and `Info` states explicitly that the preset has no user-spellable name and is reachable only through bare `-S`.

Docs and comments only — no product code, no behavior change.

### Round ledger

| --- | --- | --- | --- |
| 1 | MAI-Code | `4f3592a60` | clean, independent non-vacuity |
| 2 | MAI-Code | `ebb33bcda` | clean (post-restack confirmation) |
| 3 | Gemini 3.1 Pro | `ebb33bcda` | clean on product; 1 non-blocking doc finding |

The head has moved to `5adedb216`, so a confirmation round at the new head is running before this is marked ready.

richlander:
## Review reconciliation — round 4 (MAI-Code, confirmation at `5adedb216`)

Clean. The doc-only commit is confirmed inert and its claims confirmed accurate.

### Verified

- **Byte-identical output** across `library … -S`, `library … -D`, `package … -S`, `type … -S`, `type … -D`, `member … -S`, and `library --help` between `ebb33bcda` and `5adedb216`.
- **The new `ISectionDescriptor.Info` claim is true.** The comment asserts the preset "has no user-spellable name; it is reachable only through bare `-S`". The reviewer read `SelectResolver`, `SectionPipeline`, and the category maps in `ProjectCommand`, `LibrarySections`, `PackageSectionDescriptors`, and `ApiSectionDescriptors`, then tried to falsify it empirically: `@Default`, `@Info`, `Default`, and `Info` all exit 1 across `library`, `type`, and `package`, and no category expands to the preset. The preset is reachable only through the `selectDefault` path.
- **No remaining user-facing `@Default` leaks** in `README.md` or `docs/`.
- **`markdownlint` clean** on `README.md`.
- **Non-vacuity re-verified at this head.** The reviewer reintroduced an `@Default` selector path in `SelectResolver`, confirmed `ResolveSelect_DefaultPoleSpelledOut_DoesNotResolve` and `LibraryCommand_DefaultPole_IsNotResolvable` both fail, then reverted and confirmed they pass. The removal is still genuinely enforced.

### Dismissed: `--help` version banner differs

The reviewer initially filed this as blocking: `--help` prints `0.16.0+ebb33bc` on the old head and `0.16.0+5adedb2` on the new one, so the commit is "not byte-identical".

Dismissed, and the reviewer agreed on being shown why. The informational version embeds the short commit SHA, so **every** commit in this repository changes that string — including one that adds a blank line to a text file. The observation cannot discriminate an inert change from a behavioral one, so it is not evidence about this commit. The suggested fix (stop stamping the SHA) would also be a real regression: that stamp is how a user reports which build they are running.

### Full round ledger

| --- | --- | --- | --- |
| 1 | MAI-Code | `4f3592a60` | clean, independent non-vacuity |
| 2 | MAI-Code | `ebb33bcda` | clean (post-restack confirmation) |
| 3 | Gemini 3.1 Pro | `ebb33bcda` | clean on product; 1 non-blocking doc finding |
| 4 | MAI-Code | `5adedb216` | **clean**; 1 finding dismissed with reasoning |

Two model families, four rounds, every round clean at its head. MERGEABLE / CLEAN. No checks are reported on this slice branch, which is terminal for a stacked base — the evidence here is the full local CLI suite (2587–2588 total, 0 failed, 14 skipped) plus #3566's green CI; anything newly scheduled once this retargets `main` will gate then.

`Ready to merge` — after #3566, which is its base.

richlander:
Ready to merge

PR #3576, opened 2026-07-30
Add a fixed Type Info section to the type view

richlander:
Adds `Type Info` to the `type` view: an identity fact table whose size does not depend on the type being inspected. Purely additive — bare `-S` does not select it yet.

**Stack position: slice 3 of 4.** Parent is #3567 (`remove-section-poles`), whose parent is #3566 (`decouple-bare-select`). Part of #3547.

## Why

The `type` view has no bounded section. Every section on its pipeline enumerates something that scales with the target — members, method groups, methods, member index, interfaces, type parameters, IL — so bare `-S` returns wildly different amounts depending on what you point it at:

| target | rows at bare `-S` |
| --- | --- |
| `System.Text.Json.JsonSerializer` | 15 |
| `System.Text.Json.JsonSerializerOptions` | 47 |
| `System.String` | 103 |

The `library` view does not have this problem, because it has `Library Info` — a fact table that answers "what is this library" without listing anything *from* it. `library -S` is a constant three sections across every library.

The type view has no equivalent. Its identity renders as the tree header from the view model, which is not a section and therefore not selectable. This PR supplies the missing piece.

## What this adds

`Type Info` reports type, kind, modifiers, base, interfaces, library, package, version, TFM, source, and member counts. Every row is a fact *about* the type rather than an entry *from* it:

```
$ dotnet-inspect type System.DayOfWeek -S "Type Info"

## Type Info

| Field | Value |
| ----- | ----- |
| Type | System.DayOfWeek |
| Kind | enum |
| Library | System.Private.CoreLib |
| TFM | net11.0 |
| Source | Platform |
| Members | 8 |
| Fields | 1 |
```

26 lines for `System.String` (252 members), 17 for `System.DayOfWeek` (8 members). The variation is `MarkoutSkipNull` dropping facts that do not apply, not growth — the same way `Library Info` varies 50/52/54 across three libraries. That is what `SectionSizeClass.Fixed` declares, and it is why this section can serve as the bounded overview.

**The data is not new.** It is exactly what the compact `-v:q` / `--markdown` identity line already computes and prints inline, and the counts reuse the same expressions, so the section cannot disagree with the header.

## Why `ExplicitOnly`

This pipeline is not a curated catalog, so its ladder still selects by position and `IsExpensive`. Without the flag, a section at this position joins the default `-v:m` markdown view — where these facts already render as the inline identity line, making the section pure duplication. I measured this: the first draft added 20 duplicate lines to `type <T> --markdown`.

AGENTS.md forbids new sections entering the default `-v:m` view. `ExplicitOnly` is the declarative way to say that, and `Type_TypeInfoSection_DoesNotEnterTheDefaultMarkdownView` is the gate that enforces it.

## Scope boundary

Registered only on `ApiMemberSectionDescriptors.CreatePipeline()`, which serves `type`. Member-detail and overload-inventory views use separate pipelines and do not receive it — which is why `CanRender` can be unconditional rather than guessing at view mode.

**Bare `-S` deliberately does not select it yet.** Pointing bare `-S` at the fixed overview requires declaring size classes on the remaining type sections and decoupling `fixedOverview` from `_curatedCatalog`; that is the next slice. Keeping it out of this PR is what makes this one provably additive.

## Evidence

**Base-vs-head matrix, 18 invocations.** Exactly two change:

| invocation | result |
| --- | --- |
| `type System.String -S @All` | **gains** `## Type Info` |
| `type System.String -D` | **gains** `\| Type Info \| section \|` |
| default tree, `-v:q`, `-v:n`, `-v:d` | identical |
| `--markdown`, `--json` | identical |
| bare `-S` (String, JsonSerializer, `List<T>`, DayOfWeek) | identical |
| `-S Properties` | identical |
| `member … -S`, `api …`, `library … -S` | identical |

**Full CLI suite:** 2591 total, 1 failed, 14 skipped. The failure is `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on unmodified `main` (the local NuGet cache lowercases `newtonsoft.json.nuspec`).

**Non-vacuity, by targeted mutation rather than assertion:**

| mutation | tests that failed |
| --- | --- |
| `ExplicitOnly` → `false` | only `DoesNotEnterTheDefaultMarkdownView` |
| view population → `null` | only `RendersIdentityFacts` and `DoesNotGrowWithTheType` |

Each mutation is caught by exactly the test that claims it, and by no others.

`ApiMemberPipeline_HasExpectedSectionCount` moves 30 → 31. That gate caught the addition, which is what it is for.

## Residual

- Bare `-S` on `type` still resolves to the `Info` preset (`Method Groups`), unchanged by this PR — next slice.
- The remaining `type` sections still declare no `SizeClass` — next slice.
- `List<T>` shows no `Type Parameters` row because `typeParamsInline` is null outside the inline path; the arity is visible in the `Type` row (`System.Collections.Generic.List<T>`). Not addressed here.
- Commands other than `type` are untouched; each gets its own PR, per the one-command-at-a-time constraint.

### DIFF
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -2980,10 +2980,110 @@ public partial class CommandExecutionTests
         Assert.DoesNotContain("├─", output);
     }
 
+    /// <summary>
+    /// <c>Type Info</c> is the type view's identity fact table, and the only section on the type
+    /// pipeline that does not grow with the type under inspection.
+    /// </summary>
     [Fact]
-    public async Task Type_SingleType_SelectWithColumns_ProjectsColumns()
+    public async Task Type_TypeInfoSection_RendersIdentityFactsRatherThanMembers()
     {
         var options = new TypeOptions
+        {
+            PlatformAssembly = "System.Text.Json",
+            TypeName = "JsonSerializer",
+            Select = [SectionNames.TypeInfo]
+        };
+
+        var (exit, output, _) = await ConsoleCapture.RunAsync(
+            () => TypeCommand.ExecuteAsync(options));
+
+        Assert.Equal(0, exit);
+        Assert.Contains("## Type Info", output);
+        Assert.Contains("| Type | System.Text.Json.JsonSerializer |", output);
+        Assert.Contains("| Kind | class |", output);
+        Assert.Contains("| Library | System.Text.Json |", output);
+        // Identity, not inventory: the member sections stay out.
+        Assert.DoesNotContain("## Methods", output);
+        Assert.DoesNotContain("## Method Groups", output);
+    }
+
+    /// <summary>
+    /// The section is <c>ExplicitOnly</c>, so it must not join the default markdown view, where
+    /// the same facts already render as the inline identity line. This is the gate for the
+    /// "new sections do not enter the default -v:m view" rule for this section.
+    /// </summary>
+    [Fact]
+    public async Task Type_TypeInfoSection_DoesNotEnterTheDefaultMarkdownView()
+    {
+        var options = new TypeOptions
+        {
+            PlatformAssembly = "System.Text.Json",
+            TypeName = "JsonSerializer",
+            MarkdownExplicitlySet = true
+        };
+
+        var (exit, output, _) = await ConsoleCapture.RunAsync(
+            () => TypeCommand.ExecuteAsync(options));
+
+        Assert.Equal(0, exit);
+        Assert.DoesNotContain("## Type Info", output);
+        // The identity facts are present, just inline rather than as a section.
+        Assert.Contains("Kind: class", output);
+    }
+
+    /// <summary>
+    /// The boundedness claim: the row set is a function of which facts apply to the type, never of
+    /// how many members it has. A 200+ member type must not produce a materially larger section
+    /// than an 8-member enum, and every label it emits must come from the same fixed vocabulary.
+    /// </summary>
+    [Fact]
+    public async Task Type_TypeInfoSection_DoesNotGrowWithTheType()
+    {
+        var large = await RenderTypeInfoLabelsAsync("System.Private.CoreLib", "String");
+        var small = await RenderTypeInfoLabelsAsync("System.Private.CoreLib", "DayOfWeek");
+
+        Assert.NotEmpty(large);
+        Assert.NotEmpty(small);
+
+        // Both draw from one declared vocabulary; neither invents rows for members.
+        string[] vocabulary =
+        [
+            "Type", "Kind", "Modifiers", "Base", "Type Parameters", "Interfaces", "Library",
+            "Package", "Version", "TFM", "Source", "Members", "Constructors", "Finalizer",
+            "Fields", "Properties", "Methods", "Operators", "Events",
+            "Explicit Interface Implementations", "Extension Methods"
+        ];
+
+        Assert.All(large, label => Assert.Contains(label, vocabulary));
+        Assert.All(small, label => Assert.Contains(label, vocabulary));
+        Assert.True(
+            large.Count <= vocabulary.Length,
+            $"Type Info grew past its declared vocabulary: {string.Join(", ", large)}");
+    }
+
+    private static async Task<List<string>> RenderTypeInfoLabelsAsync(string assembly, string typeName)
+    {
+        var options = new TypeOptions
+        {
+            PlatformAssembly = assembly,
+            TypeName = typeName,
+            Select = [SectionNames.TypeInfo]
+        };
+
+        var (exit, output, _) = await ConsoleCapture.RunAsync(
+            () => TypeCommand.ExecuteAsync(options));
+
+        Assert.Equal(0, exit);
+
+        return output.Split('\n')
+            .Select(line => line.Trim())
+            .Where(line => line.StartsWith('|') && !line.StartsWith("| ---") && line != "| Field | Value |")
+            .Select(line => line.Split('|', StringSplitOptions.RemoveEmptyEntries)[0].Trim())
+            .ToList();
+    }
+
+    [Fact]
+    public async Task Type_SingleType_SelectWithColumns_ProjectsColumns()    {        var options = new TypeOptions
         {
             PlatformAssembly = "System.Text.Json",
             TypeName = "JsonSerializer",
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -3057,7 +3057,7 @@ public partial class CommandExecutionTests
         // property silently - including an unbounded one. Pinning the count makes any addition fail
         // here, so whoever adds a property has to state that it is a fixed fact about the type and
         // not a per-member row. Bump this only alongside that judgement.
-        Assert.Equal(21, vocabulary.Count);
+        Assert.Equal(11, vocabulary.Count);
 
         Assert.All(large, label => Assert.Contains(label, vocabulary));
         Assert.All(small, label => Assert.Contains(label, vocabulary));
@@ -3125,12 +3125,26 @@ public partial class CommandExecutionTests
     // An enum is the case that catches a census computed off a different member list than the one
     // discovery sees: DayOfWeek's only field is the compiler-generated `value__`.
     [InlineData("System.DayOfWeek")]
-    public async Task Type_TypeInfoSection_EffectiveDiscovery_ListsTheFieldsItRenders(string typeName)
+    // A readonly ref struct: BuildFilteredTypeForSections did not copy IsReadOnly/IsByRefLike, so
+    // discovery hid the Modifiers row that -S renders.
+    [InlineData("System.Span`1")]
+    [InlineData("System.DateTime")]
+    // Filters narrow the type discovery builds its manifest from. Type Info reports identity, not
+    // the filtered slice, so its field set must not move when a filter is active.
+    [InlineData("System.String", "-m", "Contains")]
+    [InlineData("System.String", "--all")]
+    [InlineData("System.String", "-m", "5")]
+    [InlineData("System.String", "-k", "property")]
+    [InlineData("System.Span`1", "--unsafe")]
+    public async Task Type_TypeInfoSection_EffectiveDiscovery_ListsTheFieldsItRenders(
+        string typeName,
+        params string[] extraArgs)
     {
-        var (discoverExit, discoverOutput, _) = await RunAppAsync(
-            "type", typeName, "-D", SectionNames.TypeInfo);
-        var (renderExit, renderOutput, _) = await RunAppAsync(
-            "type", typeName, "-S", SectionNames.TypeInfo);
+        string[] discoverArgs = ["type", typeName, .. extraArgs, "-D", SectionNames.TypeInfo];
+        string[] renderArgs = ["type", typeName, .. extraArgs, "-S", SectionNames.TypeInfo];
+
+        var (discoverExit, discoverOutput, _) = await RunAppAsync(discoverArgs);
+        var (renderExit, renderOutput, _) = await RunAppAsync(renderArgs);
 
         Assert.Equal(0, discoverExit);
         Assert.Equal(0, renderExit);
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -3082,9 +3082,73 @@ public partial class CommandExecutionTests
             .ToList();
     }
 
+    /// <summary>
+    /// Effective <c>-D</c> must list the fields <c>Type Info</c> actually renders. Two things can
+    /// break this and both did: the section is a <c>Field</c>/<c>Value</c> fact table, so matching
+    /// the schema against rendered *table columns* intersects nothing and reports the section as
+    /// having no queryable fields at all; and the discovery render manifest is built without
+    /// acquisition context, so the provenance rows are invisible to it unless that context is
+    /// threaded in. Either failure is silent — exit 0 with fields missing.
+    /// </summary>
     [Fact]
-    public async Task Type_SingleType_SelectWithColumns_ProjectsColumns()    {        var options = new TypeOptions
-        {
+    public async Task Type_TypeInfoSection_EffectiveDiscovery_ListsTheFieldsItRenders()
+    {
+        var (discoverExit, discoverOutput, _) = await RunAppAsync(
+            "type", "System.String", "-D", SectionNames.TypeInfo);
+        var (renderExit, renderOutput, _) = await RunAppAsync(
+            "type", "System.String", "-S", SectionNames.TypeInfo);
+
+        Assert.Equal(0, discoverExit);
+        Assert.Equal(0, renderExit);
+
+        var advertised = ParseFirstColumn(discoverOutput, "Name");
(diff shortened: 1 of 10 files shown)


richlander:
## Review reconciliation — round 1 (GPT-5.6, adversarial)

Two blocking findings, **both correct**, both fixed in `1d040b7e7`. Neither was caught by my original matrix, and both were silent — exit 0 with data missing.

### Finding 1: effective discovery returned success-shaped empty output — fixed

`type <T> -D "Type Info"` exited 0 and listed nothing, while `--schema` listed the fields and `-S` rendered them.

I confirmed this is specific to my section rather than pre-existing class behavior, by testing the precedent I copied:

```console
$ dotnet-inspect library System.Text.Json -D "Library Info"
| Name | Kind |
| Assembly Version | field |
...                              # 27 fields

$ dotnet-inspect type System.String -D "Type Info"
                                 # nothing, exit 0
```

Cause: `Type Info` is a `Field`/`Value` fact table, so its rendered header cells are literally "Field" and "Value". `FilterSchemaToRenderedColumns` intersects schema item names against those header cells, matches nothing, and reports the section as having nothing to query. `library` already solved this with `FilterSchemaToRenderedFields`. `FilterSchemaToRenderedColumns` now takes an optional field-layout section set and matches those on rendered field rows; the parameter defaults to null, so the package and library callers are untouched.

### Finding 1b: the fix exposed a deeper gap — also fixed

This one was not in the review, and I would not have found it without fixing the above first. With column matching repaired, `-D` still under-reported — it omitted `Library`, `TFM`, and `Source`, which `-S` visibly renders:

```console
$ dotnet-inspect type System.String -S "Type Info" | grep -E '^\| (Library|TFM|Source)'
| Library | System.Private.CoreLib |
| TFM | net11.0 |
| Source | Platform |

$ dotnet-inspect type System.String -D "Type Info" | grep -E 'Library|TFM|Source'
                                 # nothing
```

`BuildTypeRenderDocument` called `BuildTypeView(type, null, null, null, null, null, ...)` — every acquisition argument hardcoded null, so the discovery manifest could not observe any provenance fact. Latent until now: no section on this pipeline reported acquisition context, and `Type Info` is the first that does. Fixed by threading a `TypeAcquisitionContext` from the two call sites, both of which already had all five values in scope for their normal render path. `-D` now lists exactly what `-S` renders.

### Finding 2: generic type parameters never appeared — fixed

Correct, and I had understated it in the PR body as a cosmetic residual. It was worse than that: the field was permanently null outside quiet verbosity, so `--fields "Type Parameters"` reported "1 field has no data" for a fact the section declares. The summary is now computed unconditionally; only its inline placement stays quiet-only, so the header output is byte-identical.

```console
$ dotnet-inspect type 'System.Collections.Generic.List`1' -S "Type Info" --fields "Type Parameters"
| Type Parameters | T |
```

### Newly disclosed residual — #3579

Widening the matrix to 33 invocations to check the threading turned up one more, which I am **not** fixing here:

```console
$ dotnet-inspect member -m System.String.Substring -D          # lists "Type Info"
$ dotnet-inspect member System.String -m Substring -S "Type Info"
Error: Select value 'Type Info' not found.
```

`member`'s static catalog is built from a bare `MemberOptions`, which falls through to the **type** pipeline rather than the overload-inventory pipeline a real member query uses. That divergence predates this PR; `Type Info` is just the first section to make it observable. Fixing it means either changing the member catalog (which would remove sections users see today) or registering the section on the member pipeline (where the counts would reflect the filtered type). Both are `member` changes, and this PR is deliberately scoped to `type`. Filed as #3579.

### Evidence

- **Matrix widened 18 → 33 invocations.** Changes remain confined to `-S @All` and the `-D` listings gaining `| Type Info | section (opt-in) |`. `member … -D`, `member … --schema`, `type … -D <existing section>`, `--schema`, and all render paths are byte-identical.
- **Full CLI suite:** 2593 total, 1 failed, 14 skipped — the failure is the pre-existing `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on unmodified `main`.
- **Non-vacuity, three new mutations, each caught by exactly one test:**

| mutation | test that failed |
| --- | --- |
| drop the field-layout section set | only `EffectiveDiscovery_ListsTheFieldsItRenders` |
| drop the acquisition context (back to `null`s) | only `EffectiveDiscovery_ListsTheFieldsItRenders` |
| revert the type-parameter summary | only `ReportsTypeParametersForOpenGenerics` |

### Not yet addressed from the review brief

The reviewer's report did not cover points 3, 4, 6, or 8 from the brief (boundedness falsification, inline/section consistency, `--json` schema surface, Markout label and `NullIfZero` behavior). Those are going to the second-family round at the new head.

Head is now `1d040b7e7`, so the next round reviews that.

richlander:
## Round 2 reconciliation (Gemini 3.1 Pro @ `1d040b7e7`)

Both findings were real and are fixed in `5222b666a`. Credit to the reviewer on both — the enum case is one my own round-1 test missed by only exercising `System.String`.

### 1. BLOCKING — `-D` and `-S` disagreed for enums. Fixed, but not the suggested way.

Reproduced exactly as reported: `type System.DayOfWeek -S "Type Info"` rendered `| Fields | 1 |` while `-D "Type Info"` did not list `Fields`.

The suggested fix was to pass the unfiltered `apiType` to `BuildTypeRenderManifest` so discovery sees what rendering sees. I did not take it, because it makes the two agree on **the wrong number**:

```console
$ dotnet-inspect type System.DayOfWeek -S Fields
Note: section 'Fields' has no data for System.DayOfWeek.
```

`DayOfWeek`'s only field is the enum's compiler-generated `value__`. Every member section drops compiler-generated members, and `BuildFilteredTypeForSections` applies that same filter before discovery. So `Fields: 1` contradicted the `Fields` section it was summarizing, and the `Values` table listing seven entries directly above it. Rendering was the wrong side of the disagreement, not discovery.

Fixed by counting from the filtered list. `-D` and `-S` now agree on `System.DayOfWeek`, `System.String`, `System.Collections.Generic.List<T>` and `System.Text.Json.JsonSerializerOptions` — verified as set equality, not containment.

**This also corrected a pre-existing bug outside the new section, which I want to flag explicitly.** The inline identity line carried its own copy of the same ten count expressions and therefore the same overcount, so `type System.DayOfWeek --markdown` printed `Fields: 1` above a `## Values` table listing seven. Leaving it would have shipped two adjacent outputs disagreeing with each other. The duplication was the defect, so the census is now computed once (`TypeMemberCounts`) and shared. That makes this PR no longer purely additive; the full measured blast radius, base vs head over 9 types × 4 output formats plus the `member` and `api` paths, is:

- enums drop the bogus `Fields` row from the inline line
- `type <enum> --fields Fields` reports `note: 1 field has no data` instead of `1`

Nothing else changes. `member` and `api` are byte-identical.

### 2. NON-BLOCKING — weak boundedness gate. Fixed, and strengthened past the suggestion.

Agreed the hardcoded 21-label vocabulary plus `large.Count <= vocabulary.Length` did not fail when a property was added. It now derives from `TypeInfoSection`'s declaration via reflection, honoring `[MarkoutPropertyName]`.

Deriving alone does not actually close the hole the reviewer described: a reflective vocabulary *absorbs* a new property silently, including an unbounded one. So the test also pins the declared property count, which forces an addition through a deliberate judgement that the new property is a fixed fact about the type rather than a per-member row.

### Non-vacuity

Both gates were mutation-tested, and each failed only its intended test:

| Mutation | Result |
| --- | --- |
| Census stops filtering compiler-generated members | fails **only** `EffectiveDiscovery_ListsTheFieldsItRenders(typeName: "System.DayOfWeek")` |
| A property added to `TypeInfoSection` | fails **only** `DoesNotGrowWithTheType` |

### Validation

Full CLI suite: **2595 total, 1 failed, 14 skipped**. The failure is `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on unmodified `main` (local NuGet cache lowercases `newtonsoft.json.nuspec`).

Clean verdicts from round 2 that I have not disturbed: `--json` byte-identity, `--fields`/`--jsonl`, `--columns`/`--value` correctly failing on field-layout sections, `library`/`package` discovery byte-identity, `NullIfZero` + `[MarkoutSkipNull]`, and `TypeAcquisitionContext` argument order.

The head has moved to `5222b666a`, so this needs a confirmation round at the new head.

richlander:
## Round 3 reconciliation (GPT-5.6 @ `5222b666a`)

Three blocking findings, all three confirmed and all three fixed in `871395ff6`. This round changed the shape of the PR, so the summary up front: **the member census is gone, and the PR is purely additive again.**

> **Correction:** the commit message for `871395ff6` references issue #3580. The correct issue is **#3589** — it was filed after the commit was written, and I do not amend history.

### The correction I owe first

My round-2 comment claimed the blast radius was "only enums change" and that `member`/`api` were byte-identical. **That was wrong.** The reviewer measured 49 changes across 689 invocations, including `AppContext`, `ArrayPool<T>`, `JsonSerializer`, `Stopwatch` and `Environment`, and demonstrated a real `member` diff. My differential was 9 types × 4 formats and never passed `--all`, which is exactly where the change bit. The claim was too strong for the evidence behind it.

### Findings 2 and 3 — the census was the wrong thing to put in this section

Both findings are about the member counts, and they share a root cause worth stating plainly: **a census is a claim about a member list, and this command has several of them that do not agree.**

- **Filter-sensitive on one path only.** Rendering counted the full type; discovery builds its manifest from the option-filtered type. So `-m`, `-k` and `--unsafe` made `-D` and `-S` disagree. Confirmed.
- **`IsCompilerGenerated` matches the `s_` naming convention.** Those are ordinary private fields with no attributes, so my round-2 census stopped counting real members. Confirmed — `System.AppContext` lost three.
- **It disagreed with `--table`.** Census 21 against the `Members` column's 24. I had traded the inconsistency I set out to fix for a different one. Confirmed.

The reviewer's suggested fix was metadata-aware classification. I agree that is the right end state, and @richlander confirmed the intended rule: private and compiler-generated members should be treated alike — not counted or presented by default, revealed by `--all`.

But that rule is currently unimplemented on **both** halves. `--all` does not bring these members back either:

```console
$ dotnet-inspect type System.AppContext --all -S Fields
Note: section 'Fields' has no data for System.AppContext.
```

Implementing it touches the filter predicate, the section renderers' `--all` handling, and the table column — and changing `IsCompilerGenerated` moves what the `Fields` section shows for many types. That is a bigger change than this slice, with its own differential. **Filed as #3589.**

So the census comes out. `Type Info` keeps its 11 identity and provenance fields, which is what the section is named for and what its own tests already assert. The round-2 change to the inline identity line is reverted along with it.

### Finding 1 — identity facts dropped in the filtered copy. Fixed, and broadened.

Confirmed: `type 'System.Span`1' -D "Type Info"` hid the `Modifiers` row that `-S` rendered, because `BuildFilteredTypeForSections` did not copy `IsReadOnly`/`IsByRefLike`.

The reviewer suggested auditing the rest, which was worth doing — **six** properties were being dropped, not two: `Accessibility`, `Attributes`, `EnumUnderlyingType`, `IsByRefLike`, `IsReadOnly`, `SourceAssemblyPath`. All six now carry over, so the copy differs from its source only in `Members`, which is its entire purpose.

```console
$ dotnet-inspect type 'System.Span`1' -D "Type Info" | grep -i modifiers
| Modifiers | field |
```

### Validation

**Purely additive, measured this time.** 25 types × 13 flag combinations (`-v:q/n/d`, `--markdown`, `--json`, `--all`, `--all --markdown`, `--all --json`, `-m 5`, `-k property`, `--shape`, `--table`) = **325 `type` invocations, 0 diffs** against the parent. Plus **30 `member`/`api` invocations, 0 diffs** — including the exact `member System.AppContext --all -m GetData` case from the review.

Full CLI suite: **2602 total, 1 failed, 14 skipped** — the failure is `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on unmodified `main`.

Tests: the discovery theory now covers a readonly ref struct, a struct, an enum, an open generic, and three filter combinations (`-m 5`, `-k property`, `--unsafe`) — the case class that would have caught the census defect on round 2. Boundedness pin moves 21 → 11. (One test case I wrote used `--limit`, which `type` does not accept; caught locally and replaced with the real flags.)

Head is `871395ff6`. Needs a clean round at the new head.

richlander:
## Rounds 4 and 5 reconciliation — both clean at `871395ff6`

Two different model families reviewed this exact head. Both returned **CLEAN with no findings**, and both ran the full suite independently.

### Round 4 — Gemini 3.1 Pro

Verified clean, each backed by real runs:

- **Purely additive.** Its own baseline build of `remove-section-poles`, its own matrix — including types I had not covered (`System.ObsoleteAttribute`, a type from this repo's own assembly, a type with no namespace) across `-v:m/n/d`, `--json`, `--jsonl`, `--columns`, `--count`, `--schema`, `-D`, `--all`. Zero byte changes.
- **`BuildFilteredTypeForSections` aliasing.** Specifically checked the concern that `Attributes` is copied by reference: confirmed `ApiType` is treated as read-only across the render and discovery pipelines, and that consumers reconstruct lists rather than mutate. Also confirmed the method is reached only from `WarnEmptySelectedSections` and `ExecuteEffectiveDiscovery` — never the render payload — so copying the six identity facts strictly narrows the gap between what discovery sees and what rendering sees.
- **`ExplicitOnly` holds** across `--markdown`, `-v:m`, `-v:n`, `-v:d`, and `--json` (the property is `[JsonIgnore]`, so the type model is not polluted).
- **Boundedness.** Never more than 11 rows regardless of type complexity; `Interfaces` is a count, `Type Parameters` a single summary string. Judged compliant with `SizeClass.Fixed`, which constrains structural row count rather than cell length.
- **`-D`/`--schema`/`-S` agreement** on a platform type, a **NuGet-sourced type** (`Newtonsoft.Json@13.0.3`), and `--unsafe`. Notably it confirmed the interaction I was relying on but had not stated: `[MarkoutSkipNull]` drops the null `Package`/`Version` rows from the dummy render, so `-D` purges them from the advertised schema instead of advertising all 11 properties unconditionally.
- **Test non-vacuity.** Confirmed the `DeclaredTypeInfoLabels` regex matches Markout's `PascalCaseWords` for the current properties and that `[MarkoutPropertyName]` correctly takes precedence for `TFM`; confirmed `ParseFirstColumn` cannot silently vacate an assertion because `Assert.NotEmpty` guards both sides.

### Round 5 — MAI-Code

Independent second family at the same head. Compared head against base on identical invocations across rendering, section selection, discovery/schema, package-based resolution, **malformed type names, and error paths** — the areas I explicitly asked it to probe because rounds 1–4 had concentrated on success paths. No blocking issue, no regression against the purely-additive claim.

### Validation

Full CLI suite, run independently by MAI-Code and by me, with identical results:

**2602 total, 0 errors, 1 failed, 14 skipped, 0 not run.**

The single failure is `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on unmodified `main` (the local NuGet cache lowercases `newtonsoft.json.nuspec`).

### Review ledger for this PR

| Round | Model | Head | Outcome |
| --- | --- | --- | --- |
| 1 | GPT-5.6 | `b01caca65` | 2 blocking — discovery schema matching, missing acquisition context |
| 2 | Gemini 3.1 Pro | `1d040b7e7` | 1 blocking, 1 non-blocking — enum `-D`/`-S`, weak boundedness gate |
| 3 | GPT-5.6 | `5222b666a` | 3 blocking — all in the member census; census removed |
| 4 | Gemini 3.1 Pro | `871395ff6` | **CLEAN** |
| 5 | MAI-Code | `871395ff6` | **CLEAN** |

Two families clean at the final head.

Ready to merge

Issue #3245, opened 2026-07-26
No regression ratchet on the authored-corpus headline metric


richlander:
Part of #3079.

## Problem

Nothing detects a regression in the authored-corpus headline metric.

The benchmark's exit code is:

```csharp
return honest && invalid == 0 && drift == 0 && unsupported == 0 ? 0 : 1;
```

(`tools/DecompilerHarness/AuthoredCorpusBenchmark.cs:180`, and again at `:353`.)

With `invalid = 5217` on the current 12,000-row corpus, this is **permanently
1**. A gate that can only go green at perfection carries no signal: it is
identically red at 56.5% valid and at 40% valid.

The movement indicators are **presentation, not assertion**. The ✓/✗ arrows are
rendered by the history card, and the 11 tests in
`src/ILInspector.Decompiler.Tests/AuthoredCorpusHistoryCardTests.cs` assert
Markdown strings against synthetic fixture rows — they verify the card renders,
not that the metric held.

Consequence: `valid %` dropping from 56.5 to 54 would land silently, and
`history.jsonl` is appended per run by hand, so nothing forces the comparison.

## The fix already exists in this repo

`CorpusSensor` is the working model (`tools/DecompilerHarness/CorpusSensor.cs:174-201`):
compare current against a stored baseline, emit a typed regression list, and
exit non-zero only when that list is non-empty.

```csharp
var regressions = Compare(..., gateAggregateRates: ShouldGateAggregateRates(...));
...
return regressions.Length == 0 && !rtsParityRegressed ? 0 : 1;
```

## Proposed change

Apply the same shape to the authored-corpus benchmark: compare against the last
recorded `history.jsonl` row (same pool, same `sweepManifestSha256`) and fail on
regression rather than on imperfection.

Design points to settle:

- **Comparability guard.** Only compare rows with an identical pool and sweep
  manifest; a changed denominator must skip the gate loudly, not silently pass.
  Pair this with `methodologyVersion` (see #3244).
- **Which metrics ratchet.** At minimum `correct` and `productBodyDefect`.
  `valid %` mixes correct with valid-but-different, so it may want a tolerance
  band rather than a strict floor.
- **Noise.** Movement within roughly +/-0.3pp is about 36 rows on a 12,000-row
  corpus; the threshold should be justified against observed run-to-run
  variance rather than picked.

## Caveat worth recording

This gate would run against a metric whose product-defect count is a **lower
[...]

Issue #3560, opened 2026-07-30
Gate the two sweep properties that only hand-run probes cover


richlander:
Two properties of `eng/prepare-decompiler-package-sweep.cs` are currently evidenced only by hand-run probes recorded on PRs, not by an automated gate:

- **The sweep honors the pin.** A sweep must pool the exact bytes `docs/data/nuget-top-packages.lock.json` names, and fail rather than pool anything else.
- **The assembly copies land where they should.** The 91 assembly writes per sweep must put the right bytes at the right path inside the output directory, without writing through anything planted there, and must account for every package whose copy fails.

`EvilPoolPinTests` gates the pin *file* thoroughly — shape rules, encoding, read bounds, refusal reports — but it runs offline and never executes a sweep, because a sweep acquires packages. So the two properties the pinning work exists to provide are the two nothing in CI watches.

This matters more than it looks. #3434 needed sixteen adversarial review rounds, and the two most serious defects found were both on this ungated path: a poisoned cache entry satisfying a pin that described the *request* rather than the bytes, and `File.Copy` writing an assembly through a symlink planted at the destination while the sweep exited 0. Both were caught by a human-run probe that no future change will re-run.

### What a gate would need

A spike during #3434 found most of the pieces already exist:

- `CoreCache` honors `XDG_CACHE_HOME` (`src/DotnetInspector.Core/CoreCache.cs:71`) and `NuGetCache` honors `NUGET_PACKAGES` (`src/DotnetInspector.Packages/NuGetCache.cs:78`), so a test can redirect both caches into scratch and never touch the shared ones.
- `PackageExtractorOfflineTests` shows the extractor can be driven offline via `HttpClientFactory.Initialize(offline: true)` and `NuGetCache.Initialize(appName, basePath, skipNuGetCache: true)`, serving from a seeded cache.

Two pieces are missing:

1. **The sweep hardcodes its inputs.** `sourcePath` (line 159) and `pinPath` (line 161) are always `docs/data/nuget-top-packages{,.lock}.json` under the repository root. A gate needs to point the sweep at a small synthetic list and pin. Options to weigh: `--package-list <path>` / `--pin-file <path>`, or a single `--data-directory <path>`. Either is arguably a real capability rather than a test backdoor — it makes a variant pool possible — but that should be decided deliberately, not assumed.
[...]

PR #3434, opened 2026-07-29
Pin the EVIL pool so a sweep measures the same code twice

richlander:
Closes the pinning half of #3353 (follow-up to #3245 / #3349).

## The problem

`eng/prepare-decompiler-package-sweep.cs` acquired every package as `@latest`. A fresh sweep therefore measured **different code** than any previously recorded run, so its pool identity could never match a recorded baseline's. That is why the authored-corpus ratchet shipped with no caller on the weekly lane: there was nothing stable to compare against.

## What this does

`docs/data/nuget-top-packages.lock.json` records the exact version and TFM of every ranked package. The sweep honors it and **refuses** to acquire a package it cannot get as pinned.

| Flag | Meaning |
| --- | --- |
| *(none)* | honor the pin; a package that cannot be acquired as pinned fails the run |
| `--refresh-pin` | re-record the pin deliberately (merges, never replaces) |
| `--resolve-latest` | explicit escape hatch for callers that want what ships today |

The weekly `package-sweep` lane now passes `--resolve-latest`, because that lane is **current-package discovery evidence** — honoring the pin there would have quietly turned discovery into a replay. Each caller now declares which it wants.

## Three things worth review attention

**The TFM check is load-bearing, not decorative.** Repinning `newtonsoft.json` to `13.0.1` acquires cleanly and matches on version — but selects `netstandard2.0` instead of `net6.0`, a genuinely different assembly. The version check alone would have passed it.

**All 100 ranked packages are covered, including the 9 that yield no library.** Meta-packages (`xunit`, `swashbuckle.aspnetcore`) and packages with an ambiguous primary library (`nunit`, `grpc.tools`) are pinned as `no-library` and skipped without a network round trip. Without that, those nine would read as *unpinned* forever and the pool could never be clean — "known to contribute nothing" has to be distinguishable from "nobody pinned this."

**No timestamp in the file.** Re-recording an unchanged pool produces a byte-identical file, so a diff means something actually moved. #3349 spent a round on the fact that hashing a file containing `generatedAtUtc` yields an identity that never repeats.

## Evidence

Real runs against nuget.org, not reasoning:

| Case | Result |
| --- | --- |
| **Two independent 100-rank sweeps** | **byte-identical pool** — content digest `85a56d1b40ec66ba` over all 91 assemblies |
| Pinned run, ranks 1–100 | exit 0, 91 selected + 9 `no-library-by-pin` |
[...]

PR #3586, opened 2026-07-30
Cross the trend store's identity bootstrap so the weekly lane can ratchet

### LINKED ISSUE #3353: Pin the EVIL pool so the authored-corpus ratchet can run in CI
Follow-up from #3349 (which closes #3245).

## The problem

`eng/prepare-evil-corpus.sh` builds its pool from `docs/data/nuget-top-packages.json`, whose entries carry `rank`, `package`, and `downloads` — **no version**:

```json
{"rank": 1, "package": "newtonsoft.json", "downloads": 226348175}
```

So `eng/prepare-decompiler-package-sweep.cs` resolves *latest* on every run. A fresh sweep therefore measures different code than any previously recorded run, and its pool identity can never match a recorded baseline's.

The consequence for #3349's ratchet: `--ratchet-baseline` on a freshly swept pool will always find no comparable row, and a skip fails. The weekly `authored-corpus-ratchet` lane consequently ships **without** `--ratchet-baseline` — it is a measurement-integrity gate and the artifact source, and the ratchet's only caller is `AuthoredCorpusRatchetTests` on the tracked trend store at merge time.

## Why this is not new

It was already true. The ratchet's first comparability key was under-specified enough to compare anyway, which is a *false green*: a package resolving to a newer version that retains the same method identities resolves cleanly, produces no drift and no unmatched rows, and was compared against numbers measured on different code. #3349 closes that hole, which is what makes the drift visible as a refusal to compare.

## What would fix it

Pin the pool so a sweep is reproducible:

- Record resolved versions (and TFMs) alongside each package in the top-packages list, or add a separate committed lockfile the sweep consumes.
- Have `prepare-decompiler-package-sweep.cs` honor the pin instead of resolving latest.
- Record one trend-store row measured against the pinned pool, carrying both `sweepManifestSha256` (canonical, over the resolved package identities — see `AuthoredCorpusRatchet.PoolManifestDigest`) and `corpusSha256`.
- Then add `--ratchet-baseline` + `--ratchet-pool-manifest` back to the weekly lane.

Refreshing the pin then becomes a deliberate, reviewable act that records a new baseline — which is the correct shape for a trend store.

## Notes

- Both identity hashes compare symmetrically in #3349, absence included, so a partially-identified run cannot borrow a baseline's identity.
- The pool digest is taken over sorted resolved `package@version/tfm` triples, not the manifest file, because the file carries `generatedAtUtc` and per-package `fromCache`. A digest that never repeats would make the gate permanently red.

richlander:
Closes #3353. Follows #3434, which pinned the pool.

## What this does

Pinning the pool made a sweep reproducible. It did not make the trend store
*ratchetable*, because every recorded row predated run identity: comparability
compares `poolSha256` and `corpusSha256` symmetrically with **absence as a
distinct value**, so the first identified row is comparable to nothing, the
comparison skips, and a skip fails.

That can only be crossed by landing two identified runs at once, which is what
this does. The weekly `authored-corpus-ratchet` lane then drops
`--integrity-only` for `--ratchet-baseline`.

## The two rows

Both are full 12,000-target runs at `35014d91` over the same pinned pool (104
assemblies: the 91-package pinned sweep plus the real-world half) and the same
corpus. They agree exactly:

| | run 1 | run 2 |
| --- | --- | --- |
| `poolSha256` | `c0d2e0de…` | `c0d2e0de…` |
| `corpusSha256` | `f450230c…` | `f450230c…` |
| correct / invalid | 1576 / 5180 | 1576 / 5180 |
| `productBodyDefect` | 325 | 325 |
| drift / unsupported / notFull / unknownOutcome | 0 / 0 / 0 / 0 | 0 / 0 / 0 / 0 |

That agreement is worth more than the rows themselves: **it is the
reproducibility claim of #3353 discharged end to end.** Before the pin, a fresh
sweep resolved whatever was latest, so no two runs could be compared at all.
Two independent sweeps now produce a byte-identical pool identity.

The rows share a date, a commit, and their counts, which is deliberate rather
than a double-append — a trend needs a second point before it can have a
direction, and a bootstrap pair's only honest direction is flat. The README says
so, so a future reader does not read it as a mistake.

## `productBodyDefect` ratchets for the first time

No earlier pair could compare it. The newest row's baseline predated methodology
stamping, so the metric dropped out of the comparison and left a green result
that ratcheted less than it appeared to.

Verified by planting a **+20 `productBodyDefect` regression absorbed from
`unclassified`** — so all three partitions still close, which is the laundering
shape that has fooled this gate before — and confirming it fails by name:

```text
Collection: [Metric { Name = productBodyDefect, Baseline = 325, Current = 345,
             HigherIsBetter = False, Regressed = True }]
```

## The lane

`--integrity-only` → `--ratchet-baseline`. Both wrong alternatives are named in
the workflow comment so this is not re-litigated:

- **Merely dropping `--integrity-only`** selects the historical `invalid == 0`
  contract this ~5,200-invalid corpus cannot satisfy — the lane would fail every
  week and file a scheduled-failure issue each time. A permanently red lane
  reports exactly as much as a permanently green one.
- **Keeping `--integrity-only`** claims only that the measurement was sound and
  declines to judge quality at all.

The integrity half still applies underneath the ratchet: drift, unsupported
targets, unmatched or malformed rows, and a non-closing partition fail
regardless of movement, because a quality metric on a run that measured less
than it claimed is meaningless.

`--ratchet-pool-manifest`, named in #3353, no longer exists and is not restored:
pool identity is derived from the run's own inputs, because an identity that
depends on the caller remembering an argument has the same shape as the gate
nobody invoked (#3245).

## Deletion

`TrackedHistory_RecordsNoRunIdentity_SoTheNextAppendCannotRatchet` is deleted, as
its own doc comment instructed once the bootstrap was crossed. Nothing re-opens
it: a row that dropped its identity would be comparable to nothing in the same
way, and `TrackedHistory_NewestRowDoesNotRegressAgainstItsBaseline` fails on a
skip.

## Validation

- Two full benchmark runs, ~92 min each, exit 0 (integrity gate)
- `--gate fast` **4375 / 0**
- `AuthoredCorpusRatchetTests` + `AuthoredCorpusHistoryCardTests` **164 / 0**
- Regression probe fails by name; store restored byte-identical after
- `markdownlint` clean; workflow YAML parses and the lane's command is as intended
- Run JSONs and sweep manifest archived out-of-tree per the append procedure

## Non-actions

`#3560` (gating the pin's two core properties offline) stays open — unrelated to
ratcheting, and it needs product path overrides plus a synthetic `.nupkg`.


richlander:
## Review round one — one finding, actioned at `90183cc25`

Single MAI-Code review at `dd3a681a6`, isolated worktree. Tier per AGENTS.md:
more than trivial, not high risk — small blast radius (one weekly lane plus the
trend store), and the gate was already demonstrated to bite.

### Actioned (Medium): deleting the bootstrap test dropped a live assertion

`TrackedHistory_RecordsNoRunIdentity_SoTheNextAppendCannotRatchet` carried
**two** assertions. Only the first expired. The second — that an identified run
is not comparable to a baseline recording no identity, and that the mismatch
names which side failed to identify itself — is a permanent rule, and the review
was right that nothing called `RunKey.IsComparableTo` directly afterward.

I measured the gap rather than assuming it, and the result is narrower and more
interesting than "the rule is untested":

| Weakening applied to `IsComparableTo` | Caught by |
| --- | --- |
| Wave through a baseline recording **neither** digest | **nothing** in the 148-test class — the whole suite stayed green |
| Wave through an absent **pool** hash alone | `Ratchet_UnknownIdentityNeverEqualsAKnownOne(runSha: "0a7eded…", baselineSha: null)` and `Ratchet_MismatchedPoolDoesNotFallThroughToAnUnidentifiedOlderRow` |

So the pool direction was already covered. The gap was specifically the
**both-absent** case — which is exactly the shape of every row predating #3353,
so it is the one the tracked store actually contains, and the one a future
regression would meet first.

`Ratchet_IdentifiedRunWillNotBorrowABaselineThatRecordsNoIdentity` closes it,
asserting at both levels: `IsComparableTo` directly, since no other test calls
it, and through `Compare`, so the baseline walk is known to honor the rule rather
than the rule merely holding in isolation. Verified non-vacuous — it fails under
the both-absent weakening and is the only test that does.

### Verified independently, before the review returned

The review skipped question D (whether next week's live run will actually be
comparable) as too expensive. That was the finding I most wanted attacked, since
getting it wrong means the lane skips and fails **every week** — so I checked it
myself:

- Both pool halves are version-pinned to published packages. The sweep half is
  pinned by `nuget-top-packages.lock.json`; the real-world half pins explicit
  package versions and takes the self-assemblies from the published
  `dotnet-inspect.any` package at a pinned `SELF_VERSION`, not from
  `artifacts/bin`, so no local build output enters the pool.
- `PoolDigest` hashes `Path.GetFileName` plus content, never the path. **Tested,
  not just read:** staging all 104 assemblies under a completely different root
  and re-running produced the identical `poolSha256`
  (`c0d2e0de36ad631da14b0e9ef33e2daa4e7608b957d92a08c1ee015a15333b6f`), which is
  the case CI exercises, since it stages to `evil-pool` rather than my scratch
  path.
- The digest covers the 26 assemblies actually decompiled, so a 200-row corpus
  subset reproduces the full run's digest exactly — which is also why that probe
  is representative rather than a weaker proxy.

### Review's clean probes, for the record

Rows verified field-by-field against the archived run JSONs; `validPct`
consistent; all three closure levels and non-negativity hold. Tamper matrix:
worsened `productBodyDefect` (caught by name), worsened `invalid` and `correct`
(rejected as unsound), removed `poolSha256` (skip), removed `invalidBreakdown`
(skip naming the metric that could not ratchet), non-closing partition (unsound).
A bumped `methodologyVersion` correctly does **not** fail — `productBodyDefect`
is designed to drop out across a methodology change.

Environment rule held: no shared cache was touched, worktree clean.

### Validation at `90183cc25`

- `--gate fast` **4376 / 0**
- `AuthoredCorpusRatchetTests` **148 / 0**; non-vacuity of the new test confirmed
  by tampering with the product and restoring

richlander:
## Review round 2 reconciliation — `a0c05e801` → `6bdf3fd7a`

Two reviewers on the fixed head `a0c05e801`: **GPT-5.6** and **MAI-Code**. Both returned findings. Both were right, and GPT's was the serious one.

### GPT-5.6 — High: a malformed newest row fell through to a weaker older row — **fixed**

This is the one that mattered, and it was a defect I introduced in `a0c05e801` while fixing a smaller one.

I had made a malformed identity mean *untrustworthy measurement*. But that predicate's disposition is `continue` — skip the row and keep walking. So a malformed **newest** row didn't fail the gate; it silently handed the comparison to an older, laxer threshold. GPT demonstrated it end to end: a run that regressed reported `RATCHET OK` with **exit 0**, where the same run against a valid store exits 1.

That is the identical fallthrough shape this file has now been fixed for three times: *refusing one row must never be a way to select a different baseline.*

Fixed by changing the disposition, not the detection. A malformed identity is a schema error, not a measurement defect, so it condemns the **whole file**: `ParseHistory` throws at the boundary, and `Compare` refuses up front before any row is considered. `IsTrustworthy` goes back to meaning only "was this run's own measurement sound" — which was also GPT's layering point.

Verified at the process level, where the hole was demonstrated:

| Baseline | Exit | Output |
| --- | --- | --- |
| malformed | 1 | `History row records an identity no run could produce: 2026-07-30: poolSha256 ''` |
| valid, no regression | 0 | `held productBodyDefect 0 -> 0` |
| valid, with regression | 1 | `REGRESSED correct 1 -> 0` |

Non-vacuity: restoring the previous per-row disposition fails `Ratchet_MalformedNewestRowDoesNotFallThroughToAWeakerOlderRow` and nothing else.

### GPT-5.6 — Medium: README contradicted the lane — **fixed**

I updated one description of the lane and left a contradicting one three paragraphs earlier still saying it "deliberately passes `--integrity-only`". Also correct that `AuthoredCorpusHarnessProcessTests` was named for the scheduled lane while exercising flags the lane no longer passes — it now pins the real combination, with `--integrity-only` covered separately since that contract still exists.

### GPT-5.6 — Medium: three factual errors in the re-bootstrap section — **fixed**

All three were in text I wrote in `a0c05e801`, and I verified each before correcting:

- `SELF_VERSION` is in `eng/prepare-decompiler-corpus.sh`, not `prepare-evil-corpus.sh`.
- `corpusSha256` digests `corpus.jsonl` alone, so a vendor-branch commit re-opens the bootstrap only when it changes that file — not on every commit, as I'd written.
- My argument against pinning the corpus was **wrong**: a pin bump changes the corpus bytes, so it *would* force a re-baseline. Pinning is declined for a different reason — it buys nothing, since identity already detects the change, while adding a second thing to keep in sync and delaying corpus improvements.

Also documented the restore script's early exit, so a stale local worktree isn't mistaken for the current corpus when appending by hand.

### MAI-Code — Medium: a trim-based weakening survived green — **fixed**

Narrower but fair: leading/trailing whitespace cases weren't pinned, so a future refactor that trimmed before validating would not have failed anything. Space and tab cases added.

### On severity, again

Worth recording: in the previous round I noted reviewers are reliable on mechanism and unreliable on severity. This round inverts that lesson in a useful way — GPT's High was correctly rated, and it caught a regression *I introduced while fixing review feedback*. The first reviewer's finding was real, my fix for it was worse than the bug, and only a second review at the fixed head caught that. That is the argument for re-reviewing fixed heads rather than treating an actioned finding as closed.

Validation: `--gate fast` **4395 / 0** (was 4388). Re-review of `6bdf3fd7a` in progress; not marking ready until it's clean.

richlander:
## Review round 3 — `6bdf3fd7a` — both reviewers CLEAN

Two independent reviewers on the fixed head, both **CLEAN**, no findings.

**GPT-5.6** (also found the round-2 High, so it re-ran its own repro):

- Original malformed-newest repro: now exit 1.
- Malformed row in first, middle, and last position: all exit 1 at `ParseHistory`.
- Valid baseline: exit 0, `RATCHET OK`, four metrics compared.
- Real regression: exit 1, `RATCHET FAILED`.
- Absent vs different-but-valid digest: exit 1, skipped, for the right reasons.
- Ratchet class 166 passed; process class 89 passed.
- README corrections verified against workflow, scripts, and digest code.

Its one observation, which I agree with and am not treating as a finding: a *well-formed but deliberately false* copied digest is indistinguishable from any other falsely copied metric. That's authored-data provenance, not a malformed-identity bypass, and the store has never claimed to defend against an appender who fabricates values that are internally consistent.

**Gemini 3.1 Pro**, given a deliberately different angle — is the gate load-bearing at all, given this is the third PR in the "#3245 nothing invoked it / #3353 inputs weren't fixed / this one always skipped" family:

- Traced workflow → harness → exit code: `set -euo pipefail` propagates a non-zero harness exit to the step, while `always()` still uploads the JSON artifact. A real regression next week does turn CI red.
- Confirmed no dedup or unstable ordering, so the identical-row pair genuinely sets the bar.
- Independently derived that `poolSha256` is *mathematically* independent of `EVIL_PACKAGE_COUNT`, because `SelectPool` bounds the pool to the 26 assemblies matching corpus identities — a stronger statement than my empirical path-independence check.
- Judged the `ParseHistory` throw acceptable: the card catches `JsonException` and exits 1, and rendering a partial trend over a corrupt store would mask the defect.

I separately confirmed the vendor orphan branch is untouched at `ad4e21a83` locally and on origin, and that the corpus worktree is clean.

### Round summary

| Round | Head | Reviewers | Result |
| --- | --- | --- | --- |
| 1 | `dd3a681a6` | MAI-Code | 1 Medium |
| 2 | `a0c05e801` | GPT-5.6, MAI-Code | **1 High**, 3 Medium |
| 3 | `6bdf3fd7a` | GPT-5.6, Gemini 3.1 Pro | **CLEAN, CLEAN** |

The round-2 High is the one worth remembering: my fix for round 1's finding was worse than the bug it fixed. Only re-reviewing the *fixed head* with a second model family caught it. An actioned finding is not a closed one.

Validation at `6bdf3fd7a`: `--gate fast` **4395 / 0**. CI green. Review worktrees removed.

Ready to merge.

Issue #3591, opened 2026-07-30
Decompiler discards portable-PDB LocalScope ranges, hoisting narrowly-scoped locals to the method top


richlander:
## Summary

The decompiler reads the portable PDB's `LocalScope` table but keeps only the
variable *names*, discarding each scope's `StartOffset`/`EndOffset`. As a
result a local that the source declared inside a nested block is emitted as a
bare declaration hoisted to the top of the method, even though the PDB records
its true extent precisely.

This is a **raising gap driven by evidence the tool already holds**, not a
taste/style question: when a PDB is present the correct answer is knowable, so
it must not be modeled as a `StyleOptionCatalog` option.

## Witness

`Azure.Data.Tables@12.11.0`, `Azure.Data.Tables.TableClient.Create`. The
package ships a PDB and dotnet-inspect already acquires it (the `response`,
`scope`, and `ex` names in the output below are PDB-derived).

Original source declares at the assignment site, inside the `try`:

```csharp
var response = _tableOperations.Create(...);
```

Decompiled output hoists it:

```csharp
ResponseWithHeaders<TableResponse, TableCreateHeaders> response;

using (...)
{
    try
    {
        response = _tableOperations.Create(...);
```

## Minimal reproduction

Two shapes, compiled `-c Release` with `<DebugType>portable</DebugType>`:

```csharp
// Shape A: `response` is used only inside the try.
public string Create(string name, int timeout)
{
    using (Scope scope = new Scope())
    {
        try
        {
            Result response = _ops.Create(name, timeout);
            return response.Value + response.Raw;
        }
        catch (Exception ex) { new Scope().Failed(ex); throw; }
    }
}

// Shape B (control): the catch arm reads `response`, so the source
// genuinely must hoist it.
public string CreateHoisted(string name, int timeout)
{
    Result response = null;
    try
    {
        response = _ops.Create(name, timeout);
        return response.Value + response.Raw;
    }
    catch (Exception ex)
    {
        new Scope().Failed(ex);
        return response == null ? "none" : response.Value;
    }
}
```

### PDB ground truth

```
Client.Create         slot 1  response   IL_0006..IL_002D   <- the try, NOT the method
Client.CreateHoisted  slot 0  response   IL_0000..IL_004A   <- whole method
```

The PDB separates the two cases exactly. Note also that Shape A's compiler
temp (slot 2) has **no** `LocalScope` entry at all, which is a usable
synthetic-vs-source discriminator.

### Current output

[...]

PR #3593, opened 2026-07-30
Pin the PDB declaration-placement frontier with a fixture (#3591)

richlander:
Fixes the evidence gap in #3591 by pinning it. **Tests and fixtures only — no
product behavior changes.**

## Claim

A present PDB changes a local's **name** and nothing about **where it is
declared**. That is only possible because `MetadataSource.LocalNames` reads the
portable PDB's `LocalScope` table for names and drops each scope's
`StartOffset`/`EndOffset`.

Because the correct answer is PDB-knowable, this is a raising gap and must not
be modeled as a `StyleOptionCatalog` taste option.

## Benchmark target

`Azure.Data.Tables@12.11.0` — `Azure.Data.Tables.TableClient.Create`. The
package ships a PDB and dotnet-inspect already acquires it to
`~/.cache/dotnet-inspect/packages/symbols/servers/Azure.Data.Tables.pdb/`, which
is where the `response`, `scope`, and `ex` names in its output come from.

```bash
dotnet-inspect member 'Azure.Data.Tables.TableClient' -m Create \
  --package Azure.Data.Tables@12.11.0 -S "Decompiled Source"
```

Original source declares at the assignment site, inside the `try`; the output
hoists it to the method top. The two fixtures added here reproduce that shape
in-repo so it can be pinned without a network dependency.

## Fixtures

Appended to the end of `CfgSampleClass.cs`, per that file's convention, so no
generated-code ordinal shifts.

`DeclScopeClient.CreateNarrow` — the local is read twice (so it survives as a
slot rather than inlining) and never referenced outside the try:

```csharp
using (DeclScopeGuard scope = new DeclScopeGuard())
{
    try
    {
        DeclScopeResult response = _ops.Create(name, timeout);
        return response.Value + response.Raw;
    }
    catch (Exception ex) { new DeclScopeGuard().Failed(ex); throw; }
}
```

`DeclScopeClient.CreateHoisted` — the control. The catch arm reads the local, so
the source genuinely must declare it above the try.

### PDB ground truth

```
DeclScopeClient.CreateNarrow    slot 1  response   IL_0006..IL_002D   <- the try, NOT the method
DeclScopeClient.CreateHoisted   slot 0  response   IL_0000..IL_004A   <- whole method
```

The PDB separates the two cases exactly.

## Current output

`CreateNarrow` with the PDB present:

```csharp
public string CreateNarrow(string name, int timeout)
{
    DeclScopeResult response;

    using (DeclScopeGuard scope = new DeclScopeGuard())
    {
        try
        {
            response = _ops.Create(name, timeout);
            return string.Concat(response.Value, response.Raw.ToString());
        }
        catch (Exception ex)
        {
            new DeclScopeGuard().Failed(ex);
            throw;
        }
    }
}
```

With symbols withheld (`MetadataSource.OpenWithoutSymbols`) the same method
renders with `V_index` names and **identical placement**.

`CreateHoisted` round-trips correctly (`DeclScopeResult response = null;` above
the try), so the gap is one-directional: the emitter is accidentally right
whenever the source really did hoist, which is why correct output on a hoisted
method is not evidence that placement is derived.

## Tests

`DeclarationPlacementFrontierTests`, modeled on `CollectionExpressionFrontierTests`
(which already carries the `Raised` / `RaisedWithoutSymbols` pair).

| Test | Pins |
| --- | --- |
| `NarrowLocal_IsHoistedAboveTheUsing_WithPdb` | the gap, with symbols |
| `NarrowLocal_IsHoistedAboveTheUsing_WithoutPdb` | the gap, without symbols |
| `NarrowLocal_PlacementIsIdenticalWithAndWithoutPdb` | **the headline claim** |
| `HoistedLocal_RoundTripsAtMethodTop` | control — the gap is one-directional |
| `PdbScopes_DistinguishTheNarrowAndHoistedShapes` | non-vacuity: the evidence exists |
| `PdbScopes_OmitCompilerTemps` | a temp has no scope entry at all |

The headline test masks every local identifier (`V_\d+`, `response`, `scope`,
`ex`) and asserts the two renderings are then byte-identical, while also
asserting they are *not* equal unmasked. Naming is the only difference.

`PdbScopes_DistinguishTheNarrowAndHoistedShapes` is the non-vacuity gate named
per AGENTS.md: if it fails, the two fixture shapes stopped being distinguishable
in the PDB and the frontier tests would be pinning a limitation that no longer
has a knowable answer. It reads the `LocalScope` table directly rather than
through a product API, because exposing those ranges is exactly what #3591 asks
for and does not exist yet.

The class `<remarks>` states that closing #3591 must flip the two `NarrowLocal`
placement tests, so a future implementer is not left guessing whether a failure
is a regression.

## Oracle consultation

Required by AGENTS.md for anything touching style-oriented output. Recorded
here because this PR asserts the difference is **not** a style question.

**Declared facet** — `dotnet/runtime` `.editorconfig`: no key governs general
declaration placement; the nearest, `csharp_style_inlined_variable_declaration =
true:suggestion` (line 124), covers `out` parameters only. On placement the
declared oracle is **silent**. Separately `csharp_style_var_* = false` (lines
52-54), so the correct *spelling* here is the explicit type, not the original's
`var` — an already-modeled, byte-neutral axis (`var-spelling-style`), out of
scope for this PR.

**Revealed facet** — runtime declares at first use when use is confined to the
block (`HttpContent c = response.Content;`
`src/libraries/System.Net.Http/src/System/Net/Http/HttpClient.cs:199`;
`StringBuilder sb = new StringBuilder();`
`src/libraries/System.Private.CoreLib/src/System/IO/File.cs:1132`) and hoists
only when a `catch`/`finally` or later code needs the local
(`HttpResponseMessage? response = null;` `HttpClient.cs:189`; `char[]? buffer =
null;` `File.cs:1126`). The rule is *narrowest scope covering all uses*, which is
precisely what the PDB encodes.

## Non-action boundary

This PR does not plumb `LocalScope` extents through `MetadataSource`, does not
touch `CSharpPrinter.CollectDeclaringStores`, and does not change any rendered
output. It only makes the gap and its available evidence testable. The fix
direction is described in #3591.

## Validation

```bash
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --no-build -- --gate fast
```

- `ILInspector.Decompiler.Tests --gate fast`: **4382 passed, 0 failed**.
- `dotnet-inspect.Tests`: 2581 run, 14 skipped, **1 failure** —
  `CommandExecutionTests.Layout_Count_CountsFilesRatherThanRenderedTreeLines`,
  reproduced identically on a clean detached `origin/main` worktree. Pre-existing
  and unrelated: it counts files inside a downloaded `Newtonsoft.Json@13.0.4`
  package and fails on `Assert.Contains("Newtonsoft.Json.nuspec", ...)`, a cached
  package artifact.

## Review

Test-only, no product code, no output change — the *more than trivial, but not
high risk* tier. One review.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -6424,3 +6424,72 @@ public sealed class CallArgClient
     public int CreateViaVolatileField()
         => _volatileRest.Create(new CallArgTarget { Name = Label }, default, _options);
 }
+
+// Declaration placement (#3591). A local the source declared inside a nested block
+// is emitted as a bare declaration hoisted to the top of the method, because
+// MetadataSource.LocalNames reads the portable PDB's LocalScope table for names only
+// and drops each scope's StartOffset/EndOffset. These two shapes are the discriminator
+// the PDB records and the printer currently ignores: in CreateNarrow the local's scope
+// is the try block, in CreateHoisted it is the whole method, and both print the same
+// way. Modeled on Azure.Data.Tables `TableClient.Create`. Appended at end of file so
+// these top-level types cannot shift any existing generated-code ordinal.
+public sealed class DeclScopeGuard : IDisposable
+{
+    public void Failed(Exception e) { }
+
+    public void Dispose() { }
+}
+
+public sealed class DeclScopeResult
+{
+    public string Value = "v";
+    public int Raw;
+}
+
+public sealed class DeclScopeOps
+{
+    public DeclScopeResult Create(string name, int timeout) => new DeclScopeResult { Value = name, Raw = timeout };
+}
+
+public sealed class DeclScopeClient
+{
+    readonly DeclScopeOps _ops = new();
+
+    // The local is read twice (so it survives as a slot rather than inlining) and is
+    // never referenced outside the try, so the PDB scopes it to the try block alone.
+    public string CreateNarrow(string name, int timeout)
+    {
+        using (DeclScopeGuard scope = new DeclScopeGuard())
+        {
+            try
+            {
+                DeclScopeResult response = _ops.Create(name, timeout);
+                return response.Value + response.Raw;
+            }
+            catch (Exception ex)
+            {
+                new DeclScopeGuard().Failed(ex);
+                throw;
+            }
+        }
+    }
+
+    // Control for the same shape: the catch arm reads the local, so the source
+    // genuinely must declare it above the try and the PDB scopes it to the whole
+    // method. Today's hoisting emitter is accidentally right here, which is why
+    // placement alone cannot be read off the current output.
+    public string CreateHoisted(string name, int timeout)
+    {
+        DeclScopeResult? response = null;
+        try
+        {
+            response = _ops.Create(name, timeout);
+            return response.Value + response.Raw;
+        }
+        catch (Exception ex)
+        {
+            new DeclScopeGuard().Failed(ex);
+            return response is null ? "none" : response.Value;
+        }
+    }
+}
--- a/src/ILInspector.Decompiler.Tests/DeclarationPlacementFrontierTests.cs
+++ b/src/ILInspector.Decompiler.Tests/DeclarationPlacementFrontierTests.cs
@@ -1,0 +1,204 @@
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
+using System.Text.RegularExpressions;
+using ILInspector.Decompiler.Pipeline;
+using ILInspector.Metadata;
+
+namespace ILInspector.Decompiler.Tests;
+
+/// <summary>
+/// Declaration placement for a local the source declared inside a nested block
+/// (#3591). The portable PDB's <c>LocalScope</c> table records each local's true
+/// extent, but <c>MetadataSource.LocalNames</c> keeps only the variable names and
+/// drops <c>StartOffset</c>/<c>EndOffset</c>, so the printer hoists a narrowly
+/// scoped local to the top of the method. These tests pin that frontier and, just
+/// as importantly, pin the evidence that would close it: the PDB already
+/// distinguishes the two shapes that today print identically.
+/// </summary>
+/// <remarks>
+/// Closing #3591 must flip <see cref="NarrowLocal_IsHoistedAboveTheUsing_WithPdb"/>
+/// and <see cref="NarrowLocal_PlacementIsIdenticalWithAndWithoutPdb"/>: with a PDB
+/// the declaration should merge into its store inside the try, which is exactly the
+/// masked equality this class currently asserts. The remaining tests are invariants
+/// and should keep passing.
+/// </remarks>
+public class DeclarationPlacementFrontierTests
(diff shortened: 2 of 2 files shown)


richlander:
## Review reconciliation

Tier: tests and fixtures only, no product code, no output change — the *more
than trivial, but not high risk* tier, so one review with MAI-Code. Both rounds
ran in an isolated detached review worktree pinned to the exact head, with
scratch under `/tmp/` and no writes to the tree.

### Round 1 — head `31b6365` — one non-blocking finding, accepted

**`PdbScopes_OmitCompilerTemps` asserted against the wrong evidence.** It checked
`IrImporter.Import(...).LocalNames` for a `null` entry rather than the portable
PDB's `LocalScope` table its name claims to cover. The reviewer showed with a
standalone probe that the importer's view reports
`[0]=scope | [1]=response | [2]=<null> | [3]=ex` while the raw PDB has no null
slot at all — the `null` is the importer's materialization, not a PDB fact. The
test would have passed even if the PDB stopped omitting temps.

Real finding. Fixed in `768d686`: a new `PdbScopedSlots` helper reads slot
indices and names straight from the PDB, the test asserts at least one IL local
slot has no scope entry, and only then cross-checks that the importer reports
`null` for exactly those slots. The claim is now anchored in the PDB and the two
views cannot drift apart silently.

### Round 2 — head `768d686` — clean

Re-reviewed the fix and re-attacked the whole change: masking-regex vacuity, the
`Scopes` iterator's `IDisposable` lifetime across `yield return`, duplicate slot
indices across nested scopes, `LocalNames`/`Locals` length alignment, PE-vs-PDB
`MethodDefinitionHandle` validity, incidental `IndexOf` substring matches, and
fixture robustness under Release codegen.

**No blocking defects.** Verified by real runs, not reading:

- `-class ILInspector.Decompiler.Tests.DeclarationPlacementFrontierTests` — 6 passed, 0 failed
- A standalone probe that rendered `CreateNarrow` with and without symbols and
  confirmed the mask makes them identical, and independently dumped the PDB:

  ```
  CreateNarrow    slot 1  response   start=6   end=45   (narrow)
  CreateHoisted   slot 0  response   start=0   end=74   (whole method)
  ```

  `LocalNames.Length == Locals.Length == 4`, and the unscoped slot in
  `CreateNarrow` is slot 2, `<null>` — so `Assert.NotEmpty(unscoped)` has teeth
  and the cross-check is not tautological.

### Validation at the fixed head

- `ILInspector.Decompiler.Tests --gate fast`: **4382 passed, 0 failed**.
- `dotnet-inspect.Tests`: 2581 run, 14 skipped, 1 failure —
  `CommandExecutionTests.Layout_Count_CountsFilesRatherThanRenderedTreeLines`,
  reproduced identically on a clean detached `origin/main` worktree. Pre-existing
  and unrelated (it counts files in a downloaded `Newtonsoft.Json@13.0.4` package
  and fails on `Assert.Contains("Newtonsoft.Json.nuspec", ...)`).

### Non-blocking follow-up

None opened. The product fix itself is tracked in #3591 and is deliberately out
of scope here.

Ready to merge

PR #3595, opened 2026-07-30
Emit LF from printers and rendered-text buffers on every platform

richlander:
Product printers and rendered-text buffers now emit LF on every platform, instead of the ambient `Environment.NewLine`.

This is the lower half of a two-PR stack. It is a prerequisite for #3562, not an alternative to it — **neither PR is green on Windows alone**, and the two fail in opposite directions.

## Why this has to land with #3562

Windows today passes for the wrong reason: CRLF literals in the working tree accidentally match CRLF product output. Normalizing the tree (#3562) makes literals LF while the product still emits CRLF. Normalizing the product (this PR) makes output LF while a CRLF checkout still holds CRLF literals. Both halves break the same tests.

| tree state | expected literals | product output | `MemberBodyProducerUnionTests` (23) |
| ---------- | ----------------- | -------------- | ----------------------------------- |
| `main` | CRLF | CRLF | 0 failed (for the wrong reason) |
| #3562 alone | LF | CRLF | 10 failed |
| this PR alone | CRLF | LF | 10 failed |
| **stacked** | **LF** | **LF** | **0 failed** |

All four rows are measured on Windows, not predicted. #3562 is therefore stacked on this branch so the pair lands as one merge point that is green on Windows.

## What changed

Printers and rendered-text buffers:

- One shared `StringBuilderLineExtensions` in `ILInspector.Text` with `AppendLf()`, `AppendLf(string?)`, and an `AppendInterpolatedStringHandler` overload. 200 `AppendLine` call sites across 11 printer files moved to it.
- 21 `new StringWriter()` sites set `NewLine = "\n"`.
- `ResearchViews`, `ApiOutputFormatter`, `OutputFormatter`, `SourceTextDiffRenderer`, `MemberCodeProvider` moved off `Environment.NewLine`.
- `CorpusSensor.RenderQualityDiffCard` renders through `MarkoutTemplate.Render(TextWriter)` rather than the string-returning overload, which inherits `Environment.NewLine`. This — not tree normalization — is the actual cause of #3509.

Terminators, where an LF interior was followed by an ambient CRLF:

- `--bare` returns `MemberCode` content (decompiled source, IL, overlays), which this PR made LF; `sink.WriteLine` then appended the console's CRLF. Measured on `member ... -S "Decompiled Source" --bare`: CR=5/LF=5 before, **CR=0/LF=5** after.
- `SerializeLibraryMarkdown` built its shell with the string-returning `MarkoutSerializer` overload, then appended LF-joined metadata sections. Because `MarkdownSectionOrderer` rejoins on the *detected* newline, a CRLF shell also normalized the LF metadata back to CRLF.

Library output, both branches, after adversarial review caught a defect this PR introduced:

- The string-returning `MarkoutSerializer` overload returns **no** trailing newline; the `TextWriter` overload writes one. Routing the Markdown shell through an LF `StringWriter` therefore added a stray blank line to every library Markdown document — **on Linux as well as Windows** — and ended `\n\r\n` on Windows.
- The plain-text branch streamed straight to `Console.Out` (ambient CRLF) while the metadata it appends comes from `MetadataLensRenderer`, which this PR made LF. Measured: **4 CRLF against 62,374 bare LF**, where `main` is 62,378 CRLF and 0 bare LF.
- Both branches now compose through an LF writer, trim, and terminate with `WriteLfLine`. Plain text joins its metadata with a single newline, matching what the streamed original produced; the Markdown sibling keeps its blank line because Markdown sections require one.

Only one other site swapped a string-returning overload for a `TextWriter` one: `CorpusSensor.RenderQualityDiffCard`. It is unaffected because its pre-existing `TrimEnd('\n') + "\n"` normalization absorbs the difference. That audit is complete — the PR contains exactly three such changes.

## Scope boundary

`OutputFormatter.WriteLimitedMarkdown` and its six callers are **deliberately unchanged**. Those callers still receive CRLF interiors from string-returning `MarkoutSerializer.Serialize` overloads, so moving only their terminator would introduce the very mixing this PR removes elsewhere. Interior and terminator have to move together; that sweep is #3596.

## User-visible change on Windows

Not "none" — an earlier revision of this description claimed that, and it was wrong.

- `type` is byte-identical to `main` (it passes straight through `Console.Out`).
- `member StringDistance ... -s @All -v:d` goes 1621 bytes / CR=45 / LF=45 to 1577 bytes / CR=1 / LF=45. **Content is equal after normalizing CRLF to LF**; the residual single CR is `Console.WriteLine`'s own terminator.

Library output is LF-only and content-equal to `main` after normalizing endings and trimming trailing whitespace:

| command | bytes | CRLF | LF | content == `main` |
| ------- | ----- | ---- | -- | ----------------- |
| `library -S "Library Info"` | 883 | 0 | 32 | yes |
| `library -v:d` | 322561 | 0 | 1034 | yes |
| `library -S @Metadata --plaintext` | 5947168 | 0 | 62378 | yes |

Linux is a no-op by construction for the printer and buffer changes: `Environment.NewLine` is already `"\n"` there and `StringWriter.NewLine` defaults to it. That claim did **not** hold for the library path before the fix above — the stray blank line affected every platform — which is why it is called out rather than folded into the general statement.

## Validation

Run on Windows, on the stacked head. CI has no Windows job, so none of this is covered by CI.

| suite | result |
| ----- | ------ |
| `ILInspector.Decompiler.Tests` | 4620 total, **7 failed**, 7 skipped |
| `dotnet-inspect.Tests` | 2581 total, **2 failed**, 10 skipped |

`main`'s Windows baseline is 11 failures in the decompiler suite. The stack removes exactly the 4 line-ending failures (3 `QualityDiffCard`, 1 `StringSwitchRaising`) and adds none. The remaining 7 are confirmed pre-existing on `main` **by name**, not merely by count:

- `LoweredFidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket`
- `FidelityGateTests.NoNewFidelityDiffsBeyondKnownDocket`
- `FidelityGateTests.PinnedFixesStayExact`
- `GeneratedFixtureCatalogTests.MinimalCompileBackRungs_ReportExpectedOutcomesByFixtureId`
- `GeneratedFixtureCatalogTests.CompilerLoweringFrontier_IsSelectableButNotInDefaultRun`
- `GeneratedFixtureCatalogTests.ReturnToSenderRecordCatalog_CoversGeneratedRecordHelpers`

The 2 CLI failures are pre-existing #3519.

## Test-side changes

`SourceTextDiffRendererTests` built its expectations with `Environment.NewLine` against a renderer this PR made LF.

Nine `Split(Environment.NewLine)` sites in `CommandExecutionTests` now use a `SplitOutputLines` helper that splits on `'\n'` and trims `'\r'`. Only one of the nine actually failed. The other eight were **silently vacuous** on Windows: the split returned a single line, the per-line `Where(...)` filters matched nothing, and the assertions passed while checking nothing. The helper is correct for either ending, so it stays correct when #3596 moves the remaining console paths.

## Follow-up

#3596 tracks the residual `Environment.NewLine` inventory and the string-returning `MarkoutSerializer.Serialize` call sites.


richlander:
## Adversarial review reconciliation — `e0f70e73`

Two reviewers, both on the fixed exact head, each in an isolated detached review worktree. This PR is high risk (subtle correctness, cross-platform blast radius, no Windows CI), so it took the two-model tier.

| reviewer | model | first pass | final pass on `e0f70e73` / `ae82b3be` |
| --- | --- | --- | --- |
| MAI-Code | `mai-code-1-flash-picker` | 2 findings, 1 blocking | **clean** |
| GPT | `gpt-5.6-sol` | 4 findings, 1 blocking | **clean** |

### Findings, attribution, and resolution

**1. Interpolated-string fast path lost — MAI-Code (finding 1), independently found by GPT.**
The original `AppendLf(string)` forced the decompiler printer hot path through the string overload, defeating `StringBuilder.AppendInterpolatedStringHandler`. GPT measured the cost at **40 MB allocated per 1M calls**. I had self-caught this shortly before both reports landed; crediting both reviewers because each found it independently.
Resolved in a0db320d — added an `AppendLf(ref StringBuilder.AppendInterpolatedStringHandler)` overload with `[InterpolatedStringHandlerArgument(nameof(builder))]`.

**2. Three duplicated per-assembly helpers — MAI-Code (finding 2, blocking).**
`StringBuilderLineExtensions` had been copied into CSharp, Decompiler, and Instructions, and the LF change had reached only some product surfaces.
Resolved in 4c165d25 — one shared `public static class StringBuilderLineExtensions` in `src/ILInspector.Text/`, the three copies deleted, and LF extended to `ResearchViews`, `ApiOutputFormatter`, `OutputFormatter`, `SourceTextDiffRenderer`, and `MemberCodeProvider`.

**3. LF interior with a CRLF terminator — GPT.**
Three sites emitted an LF-interior payload and then terminated it with `Console.WriteLine`, producing exactly the mixing this PR exists to remove.
Resolved in 64eded18 — added `OutputFormatter.WriteLfLine` and applied it to `ApiCommand`'s `--bare` payload and rendered-Markdown path. Verified empirically: `--bare` decompiled source is **CR=0 / LF=5**, against **CR=5 / LF=5** before.

**4. Stray blank line on every platform — GPT (blocking). The most valuable finding in the review.**
Routing `SerializeLibraryMarkdown` through an LF `StringWriter` silently swapped `MarkoutSerializer.Serialize(T, context, options)` — which returns **no** trailing newline — for `Serialize(T, TextWriter, context, options)`, which **writes one**. The callers then added their own terminator on top. Every library Markdown document gained a blank line, and on Windows ended `\n\r\n`.

This was worse than the mixing it was meant to fix, and it **falsified my "Linux is a no-op by construction" claim** — the stray line appeared on Linux too. I reproduced it on a clean exact-head worktree before acting, per AGENTS.md.

While confirming it I found **a second half neither reviewer reported**: the plain-text branch of the same method streamed to `Console.Out` (ambient CRLF) while appending metadata that this branch had just moved to LF — measured **4 CRLF against 62,374 bare LF**, where `main` is 62,378 CRLF and 0 bare LF.

Both halves resolved in e0f70e73. In fixing it I introduced and self-caught a follow-on error: a `"\n\n"` join inserted a blank line before `## Metadata: Image` that `main` does not have. The tests passed either way; only a line-by-line diff against `main` caught it. The plain-text separator is now a single `"\n"` to match the streamed original, while the Markdown sibling deliberately keeps its blank line because sections require one.

**5. Overload-swap audit — asked of MAI-Code after finding 4, answered independently by both of us.**
Since this class of bug bit twice, I audited every string→`TextWriter` swap in the PR. There are exactly three: `SerializeLibraryMarkdown` (was the bug), the plain-text branch (fixed), and `CorpusSensor.RenderQualityDiffCard` — which is safe because its pre-existing `TrimEnd('\n') + "\n"` absorbs the difference. MAI-Code reached the same conclusion on the same site.

### Verified, not dismissed

- **The deferred scope boundary is real.** `WriteLimitedMarkdown` and its six callers (`ApiCommand:556`, `DependsCommand:327`, `FindCommand:166/197`, `ExtensionsCommand:247`, `ImplementsCommand:169`) stay on CRLF, tracked in #3596. I asked both reviewers to attack this. GPT verified by runtime probe that those interiors are **coherent CRLF-only, not mixed**; MAI-Code independently agreed the boundary is an intentional compromise rather than a new defect. Interior and terminator have to move together — moving only the terminator would manufacture the mixing this PR removes.
- **Multi-assembly output is not mixed.** `WriteLibraryResults` uses the string-returning overload plus `Console.WriteLine`, so both interior and terminator are CRLF. Better than a probe, this holds **by construction**: `RejectMultiAssemblyMetadataSelection` (`LibraryCommand.cs:857`) errors out when metadata sections are selected with multiple assemblies, and `MetadataLensRenderer` is the sole LF-injection source into that path. It therefore cannot mix.
- **Buffering the plain-text document is acceptable.** The plain-text branch moved from streaming to `Console.Out` to buffering a full document (~6 MB for `-S @Metadata` on a 1.7 MB assembly). I raised this against myself and asked MAI-Code to rule on it; it is a bounded tradeoff required to compose body and metadata coherently, and not a regression.
- **Nine `Split(Environment.NewLine)` sites were silently vacuous.** Against LF output on Windows those return one line, per-line `Where(...)` filters match nothing, and the assertions pass while checking nothing. Only **one of the nine** failed loudly. All nine now use an ending-agnostic `SplitOutputLines` helper, so the other eight assert something real for the first time.

### Non-actions

- No change to `MarkdownSectionOrderer.Apply`, which deliberately rejoins on the *detected* newline — its doc comment states that reordering "is not licensed to rewrite the document's line endings." Correct as-is; it was a contributing mechanism in finding 3, not a defect.
- No change to `TryExtractSectionBody`'s own split, which is scoped to already-normalized text; left with an explanatory comment.

### Evidence

All Windows runs on the stacked head `ae82b3be`, which is the combined state that actually lands. Windows has no CI job in this repository, so these are the only signal for the platform this PR is about.

| suite | result |
| --- | --- |
| `ILInspector.Decompiler.Tests` | 4620 total, **7 failed** — all 7 reproduce on `main` **by name**, not merely by count |
| `dotnet-inspect.Tests` | 2581 total, **2 failed** — both pre-existing (#3519) |

`main`'s baseline is 11 failures: these same 7, plus 3 `QualityDiffCard` and 1 `StringSwitch`. The stack removes exactly the 4 line-ending failures and adds none.

Byte-level comparison against a `main` control clone, content compared after normalizing CRLF to LF and trimming:

| command | bytes | CRLF | bare LF | content == `main` |
| `library -S "Library Info"` | 883 | 0 | 32 | yes |
| `library -v:d` | 322561 | 0 | 1034 | yes |
| `library -S @Metadata --plaintext` | 5947168 | 0 | 62378 | yes |
| `library --package Microsoft.TestPlatform.ObjectModel -v:d` | 29992 | 0 | 194 | yes |

`main` produces the same content with the endings inverted (e.g. CRLF=32/LF=32 and CRLF=195/LF=0 for the first and last rows). No output is mixed in either direction.

Linux CI is green on `e0f70e73`: `build-net10`, `test (ubuntu-26.04)`, `decompiler-gates`, `csharp-diff-smoke`, `il-diff-smoke` all pass.

richlander:
Ready to merge.

All merge-blocking validation is complete: Linux CI green on `e0f70e73`, both required fixed-head adversarial reviews clean (MAI-Code and GPT, reconciled [above](https://github.com/richlander/dotnet-inspect/pull/3595#issuecomment-5137040337)), and the Windows suites run manually since this repository has no Windows CI job.

### Merge order matters — merge #3562 into this branch first

This PR and #3562 are two halves of one fix. Neither is green on Windows alone, in *opposite* directions:

| state | expected literals | product output | `MemberBodyProducerUnionTests` (23) |
| --- | --- | --- | --- |
| `main` | CRLF | CRLF | 0 failed — passing for the wrong reason |
| #3562 alone | LF | CRLF | **10 failed** |
| this PR alone | CRLF | LF | **10 failed** (measured, not predicted) |
| stacked | LF | LF | **0 failed** |

The preferred sequence:

1. **Merge #3562 into `fix/product-emits-lf`.** #3562 is already retargeted to this branch and its unique diff is exactly `.gitattributes`, 30 lines. This advances this PR's head to the combined content and **triggers CI here**, since this PR's base is `main`. (#3562 gets no CI while stacked — `ci.yml` is `pull_request: branches: [main]`. That is benign: its blob-level diff against this branch is only `.gitattributes`, and Linux checkouts are already LF.)
2. **Merge this PR into `main`** once that run is green.

That produces one merge commit to `main` that is green on Windows, with no interval in which `main` is worse than it is today.

If the two are instead merged separately, **this PR must go first** — #3562 must never precede it, or `main` briefly carries 20 Windows failures against this PR's 18.

### Follow-up

#3596 tracks `WriteLimitedMarkdown` and its six callers, deliberately left on CRLF here because their interiors are still CRLF and moving only the terminator would manufacture mixing. Non-blocking.

PR #3505, opened 2026-07-29
Normalize synthesized-member ordinals in the compile-back fidelity oracle

richlander:
- Fixes #3503
- Fixes #3491 (five of its six rows; the sixth, `ManualConstantOnlyComparisonFactory`, is #3502 and is unaffected)
- Advances #3326
- Changes the compile-back fidelity **oracle**, not decompiler output: Contract V1 no longer treats Roslyn's renumbering of synthesized-member ordinals as a diff.

## Change

> Should we accept this change?

**Conclusion:** **PASS** — the harness was comparing a property of member ordering in the recompiled compilation unit, which is not evidence about the method body. The normalization removes exactly those six false positives while leaving every genuine diff in place.

This is not a raise, structuring, or printer change, so the template's Original source / Before / After render blocks do not apply: **no decompiled output changes**. The Before/After that matters is the oracle's verdict, recorded in the Evidence table.

### Root cause

Roslyn embeds the containing method's declaration ordinal in the names it synthesizes for closures, lambda caches, and local functions:

```text
<>9__N_M            lambda cache field
<Name>b__N_M        lambda method
<Name>g__Local|N_M  local function
```

`N` is the containing method's ordinal **in the compilation unit**. It is a fact about member ordering, not about the body.

The compile-back harness recompiles a reconstructed skeleton whose member ordering is the harness's, not the original source file's, so Roslyn assigns a different `N` to the same method. `CfgSampleClass.StatementBodyLambdaInsideIf` is the clearest witness: the shipped assembly contains `b__103_0`, the recompiled skeleton contains `b__128_0`, and Contract V1 compared the two literally and reported an OperandDiff.

`ContractV1BodyDiffNormalization` already normalizes variable layout and assembly scopes for the same reason — both are recompilation artifacts rather than body evidence. This adds the third member of that family.

### Why this is not harness compensation for a product gap

`AGENTS.md` prohibits harness-side normalization that covers for missing or incorrect product behavior. This is the opposite case: the product's decompiled C# is **correct**, and the divergence is introduced by Roslyn on the recompile leg. The harness owns the oracle, and an oracle that reports a difference the product did not cause is an oracle defect. That is why the fix sits in the shared diff substrate plus Contract V1, and not in a shape-recognizer that special-cases the failing rows.

### How the normalization is bounded

Masking is the only dangerous failure here. Normalization is many-to-one, so an over-broad rewrite can hide a real IL difference, while an under-broad or declined rewrite only costs a false positive. Every design choice below leans toward declining.

[...]

PR #3600, opened 2026-07-30
Make the fidelity dockets two-directional and retire 45 stale rows

### LINKED ISSUE #3584: Fidelity dockets are one-directional allow lists: 25 stale rows silently permit future regressions
## Summary

`FidelityGateTests.KnownDiffs` and `LoweredFidelityGateTests.KnownDiffs` are allow lists that are only ever checked in one direction. A method that stops producing a diff keeps its docket row forever, and that stale row then **silently permits a future regression in that method**.

Measured on `df180752`:

| Rail | KnownDiffs | Actual diffs | Stale | Missing |
| Raised | 83 | 73 | **10** | 0 |
| Lowered | 60 | 45 | **15** | 0 |

## Root cause

Both gates assert only that every *actual* diff appears in the docket:

```csharp
// FidelityGateTests.cs:540
var unexpected = diffResults.Where(result => !KnownDiffs.Contains(result.Method)).ToList();
```

There is no converse check, so `KnownDiffs \ actual` is unconstrained. `LoweredFidelityGateTests.cs:268` has the same shape.

## Why this matters

A docket row is a statement that a specific method is *known* to round-trip imperfectly. Once the underlying raise improves and the method becomes `Exact`, the row stops describing reality and starts acting as a blanket exemption: if `CollectionListLiteral` regresses tomorrow, the raised gate will not report it, because the row is still there.

This is the "keep failure visible" rule in `AGENTS.md` — the gate currently converts a future regression in any of these 25 methods into success-shaped output.

It also means the docket cannot be trusted as a burndown list: 10 of 83 raised rows and 15 of 60 lowered rows are already done and nobody knows it.

## Stale rows

Raised (10):

```
CallOutTarget, CapturedParamReadInOuterBody, CollectionListLiteral,
GenericRefKindCallSites, ManualParameterPlusConstantFactory, OrBoolIntMix,
OrBoolUintMix, RefKindCallSites, SelectBoolReturn, SimpleExpressionTreeLambda
```

Lowered (15):

```
CallOutTarget, CollectionListLiteral, GenericRefKindCallSites,
Issue2830_ForLoopEhCleanupContinue, Issue2830_ForLoopLeave,
Issue2830_ForLoopNestedContinue, Issue2861_ForLoopTryAndCatchContinues,
ManualParameterPlusConstantFactory, OrBoolIntMix, OrBoolUintMix,
RefKindCallSites, ReverseCopy, SelectBoolReturn, SimpleExpressionTreeLambda,
WhileNestedContinueKeepsArmExclusive
```

The `Issue2830_*` / `Issue2861_*` cluster suggests a for-loop/EH raising fix landed and its docket rows were never retired.

## Suggested fix

Assert exact set equality in both directions, so that a stale row and a missing row both fail. This is the shape `ValidityCoverageReportingTests` already uses for the CS0161 census, and the shape `AGENTS.md` recommends under *Asserted properties name their gate* — make the declaration drive the enforcement set rather than restate it.

Retiring the 25 rows should be done row by row rather than in bulk: each one needs a check of *why* it is now `Exact`, since "the raise improved" and "the fixture moved and the diff relocated" are different outcomes and only the first justifies deletion.

## Provenance

Found by the GPT-5.6-Sol reviewer during adversarial review of #3528 and independently reproduced with a reflection probe against the Release test assembly. Not caused by #3528: that PR adds three raised and six lowered rows, **all of which are live** (`stale ∩ rows-added-by-3528 = ∅`), and both rails report `missing = 0`. Filed rather than fixed there because retiring rows is a separate behavioural change to the gates.

richlander:
## What

`NoNewFidelityDiffsBeyondKnownDocket` is a **one-directional allow list**. It fails when an
undocketed method diffs, but never when a docketed method *stops* diffing. Both ways a row can
leave the diff set are silent un-gates:

| Row becomes | Meaning | Today |
| --- | --- | --- |
| `Exact` | the residual was fixed | row now allows a regression it was never meant to cover; nothing reports it |
| `RecompileFail`/`ContextFail` | decompiled C# stopped compiling | drops out of the diff set, so the allow list **swallows a real regression** |

The repo already knew about the second mode and patched it **one fixture at a time**.
`FidelityGateTests.cs` remarks, on `SwitchStoreThenUse`:

> its danger mode is a *silent* regression that produces uncompilable output: that would drop it
> out of the opcode-diff set, so `NoNewFidelityDiffsBeyondKnownDocket` would not catch it.

That is why `SwitchStoreFold_StaysCompileBackCheckable` and
`PointerStoreUsesOriginalAddress_StaysCompileBackCheckable` exist. This PR generalizes them to all
143 docket rows.

## The measurement

Measured on this PR's merge-base:

| Rail | rendered | docket | actual diffs | MISSING | STALE |
| --- | --- | --- | --- | --- | --- |
| Raised | 668 | 83 | 59 | **0** | **24** |
| Lowered | 668 | 60 | 38 | **0** | **22** |

`missing = 0` on both rails, so retiring the rows cannot mask a current diff.

This was 10 raised / 15 lowered when #3584 was filed. #3505 landed after #3528, canonicalized
synthesized-member ordinals, and took the raised rail from 10 to 24 — exactly the outcome the
`StatementBodyLambdaInsideIf` row predicted when it was filed as #3503:

> Tracked as #3503, which proposes canonicalizing synthesized ordinals in the oracle so this whole
> class of row stops needing docket entries.

The class stopped needing entries. Nothing removed them, and nothing said so. `StatementBodyLambdaInsideIf`
was added by #3528, one commit before #3505 made it stale.

## The change

1. **`DocketRowsStayCheckedDiffs`** on both rails: every `KnownDiffs` row must still be `OpcodeDiff`
   or `OperandDiff`. This matches `eng/decompiler-gate-known-red.txt`, which already fails both when
   an unpinned test fails and when a pinned test passes.
2. **45 now-`Exact` rows promoted to `PinnedExact`**, not deleted. A docket row that stops diffing
   gates nothing; a `PinnedExact` row does. Promoting converts each silent allow into a real guarantee.
3. **`KnownNotFull`** — one row (`CapturedParamReadInOuterBody`) imports below Full, so the harness
   reports `NotFull` and forms no opcode verdict. It gets a named set rather than blanket tolerance,
   so a row that *newly* drops to `NotFull` still fails as the validity regression it is.

## Evidence

- `FidelityGateTests` + `LoweredFidelityGateTests`: **11/11 green**, Release.
- **Non-vacuity**: adding `CheckedAdd` (a `PinnedExact` method) to `KnownDiffs` fails with
  `CheckedAdd: now recompiles Exact, so its docket row is stale and silently allows a future
  regression.` Reverting it passes. The gate is not vacuous.

## Validation note

These gates are `Speed=Slow`, so **PR CI does not run them** — they run in Deep Inspect / publish.
The measurement above was taken on Windows, and some docket prose attributes rows to *SDK toolchain
drift* (the expression-tree group), so the stale set could in principle be toolchain-sensitive. I am
dispatching `deep-inspect.yml` on this branch to confirm the set on Linux before this merges. If the
Linux set differs, that is itself a finding and I will adjust the rows.

Closes #3584

### DIFF
--- a/docs/decompiler-correctness-pipeline.md
+++ b/docs/decompiler-correctness-pipeline.md
@@ -586,7 +586,11 @@ Report compile-back evidence in two layers:
    covers the changed shape. Name whether the sugared gate (`FidelityGateTests`),
    lowered gate (`LoweredFidelityGateTests`), or a pass-specific test is the
    relevant guard. If a fidelity-diff docket row is fixed, shrink `KnownDiffs` and
-   add the method to `PinnedExact` in the same PR.
+   add the method to `PinnedExact` in the same PR. `DocketRowsStayCheckedDiffs`
+   (both rails) enforces this: a `KnownDiffs` row that recompiles `Exact` fails the
+   gate and names the row to promote. Before #3584 the rule was documented but
+   unenforced, and 46 of 143 rows had silently gone stale — a stale row gates
+   nothing, because the diff it allows no longer happens.
 2. **Changed-method / corpus layer** — for risky or broad changes, identify the
    methods the PR actually changed and run `--fidelity-method-delta` over that
    population when available. Treat `Exact` as checked green and `OpcodeDiff` /
--- a/docs/decompiler-quality.md
+++ b/docs/decompiler-quality.md
@@ -763,7 +763,10 @@ method (a previously-fixed one) regresses. Shrinking `KnownDiffs` and growing
 That ratchet is two-directional: `DocketRowsStayCheckedDiffs` fails when a
 `KnownDiffs` row stops being a diff, whether because it was fixed (promote it to
 `PinnedExact`) or because its C# stopped recompiling (a regression the
-diff-set-shaped assert would otherwise swallow).
+diff-set-shaped assert would otherwise swallow). The one tolerated non-diff state
+is an import below Full fidelity, which forms no opcode verdict at all; those rows
+must be named explicitly in `KnownNotFull`, so a row that *newly* drops to
+`NotFull` still fails as the validity regression it is.
 The **annotation gate** (`AnnotationGateTests`) holds annotation
 fidelity over the whole CoreLib corpus the same way — precision is absolute (a
 wrong fact always fails; it is never runtime drift), recall is held above a
--- a/docs/decompiler-quality.md
+++ b/docs/decompiler-quality.md
@@ -760,6 +760,10 @@ contract V1 body — a regression beyond the documented `KnownDiffs` docket —
 when the product body comparison is unavailable, or when a `PinnedExact`
 method (a previously-fixed one) regresses. Shrinking `KnownDiffs` and growing
 `PinnedExact` is how fidelity progress ratchets forward and cannot slip back.
+That ratchet is two-directional: `DocketRowsStayCheckedDiffs` fails when a
+`KnownDiffs` row stops being a diff, whether because it was fixed (promote it to
+`PinnedExact`) or because its C# stopped recompiling (a regression the
+diff-set-shaped assert would otherwise swallow).
 The **annotation gate** (`AnnotationGateTests`) holds annotation
 fidelity over the whole CoreLib corpus the same way — precision is absolute (a
 wrong fact always fails; it is never runtime drift), recall is held above a
--- a/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
+++ b/src/ILInspector.Decompiler.Tests/FidelityGateTests.cs
@@ -54,11 +54,6 @@ public class FidelityGateTests
         // now fully raised by ComparisonTreeBoolArmPass, but still recompiles to
         // different branch structure like the sparse-switch over-render docket.
         "ByteRangeSearchTree",
-        // CollectionListLiteral is the #1142 PDB-discriminated List<T>
-        // collection-expression raise. The original SDK csc lowering uses
-        // constant Span<T> indices, while the Roslyn compile-back package emits a
-        // hidden index temp for the same `return [a, b, 42];` source idiom.
-        "CollectionListLiteral",
         "GotoCommonExit",
         // This hand-written await-enumerator loop recompiles through the same
         // runtime-async shape but schedules the receiver load after the
@@ -70,7 +65,6 @@ public class FidelityGateTests
         // the element-ref helper. Exact recovery needs a higher-level inline-array
         // foreach/indexer raise.
         "RuntimeInlineArrayForeach",
-        "SelectBoolReturn",
         // SlotDiamondDispatch is a clustered-case switch with bool-computing arms
         // (#912): SlotDiamondPass folds each returned slot diamond so the dispatch
         // raises into nested if/else, the same honest comparison-tree over-render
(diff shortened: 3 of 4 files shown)


richlander:
## Review reconciliation

Two adversarial reviewers on isolated worktrees at fixed heads: **MAI-Code** and **GPT-5.6-Sol**.
Both returned CLEAN on the initial head and again on the fixed head.

### Round 1 — head `b3b19ea6`

I gave both reviewers six attack points, including my own top suspicion.

| Attack point | Outcome |
| --- | --- |
| **Overload semantics** — the dockets are keyed by method *name*, but `CompileBackResult` has an `Overload` field and the new test requires *every* match to be a diff. A name with one `Exact` and one `OperandDiff` overload would pass the old assert and fail mine. | **Refuted, independently by both.** Overloaded names in `CfgSampleClass` do not intersect the gate name sets, so there is no latent false positive. This was my own worry and both reviewers killed it with evidence. |
| **Rows in both sets** (contradictory: must be `Exact` *and* must be a diff) | None. Verified mechanically on both rails. |
| **Dropped/duplicated rows** during comment-laden editing | None. All 45 promotions accounted for exactly; nothing dropped or duplicated. (I had caught one accidental `TwoCaptureLambda` duplicate myself mid-edit, so this class of error was live.) |
| **Stale group prose** | Two real hits — see NITs below. |
| **`NotFull` handling** | Correct and consistent with `BodyComparisonRemainsAvailable`. |
| **Toolchain sensitivity of the expression-tree promotions** | See "Platform validation" below — settled empirically rather than by argument. |

Both reviewers also built and ran the gates themselves (MAI: 6 + 5 = 11 tests, 0 failures).

**GPT NIT 1** — the lowered `KnownDiffs` `<summary>` still described `SelectBoolReturn` and
`ReverseCopy` as docketed diffs after this PR promoted both to `PinnedExact`. **Fixed** in `6194205c`.

**GPT NIT 2** — `FidelityUnavailable` fell to the default arm, which asserted the decompiled C#
"no longer recompiles". That is false for that status: recompilation succeeded and only the body
comparison produced no verdict. **Fixed** in `6194205c` with its own arm pointing at
`BodyComparisonRemainsAvailable`. MAI judged the resulting overlap useful defense-in-depth rather
than harmful duplication.

### Round 2 — fixed head `6194205c`

Both reviewers confirmed mechanically that **no row in any of the six collections changed** between
`b3b19ea6` and `6194205c`, and that the 46-of-143 arithmetic holds (24/83 raised + 22/60 lowered).
Both CLEAN.

**GPT NIT 3** — the ratchet sentence in `docs/decompiler-quality.md` said the gate fails when a
`KnownDiffs` row stops being a diff, without naming the one tolerated exception (`KnownNotFull`).
Stating a rule without its carve-out overstates what the gate enforces. **Fixed** in `28dd2fea`.

### Platform validation

These gates are `Speed=Slow`, so **PR CI does not run them**. I flagged in the PR body that my
measurement was taken on Windows and that some docket prose blames *SDK toolchain drift*, which
would make the stale set toolchain-sensitive. Rather than argue it, I dispatched `deep-inspect.yml`
(lane `test`) on this branch to settle it on Linux:

> `ILInspector.Decompiler.Tests  Total: 4293, Errors: 0, Failed: 0, Skipped: 0`

**4293 = 4291 + 2** — the prior baseline plus exactly the two tests this PR adds — and
`FidelityGateTests.PinnedFixesStayExact` appears in the log as a long-running test, confirming the
slow gates genuinely executed rather than being filtered away. The row sets are platform-stable, and
the toolchain-sensitivity concern does not materialise.

### One thing worth recording

While updating docs I found `docs/decompiler-correctness-pipeline.md` already stated this PR's rule
as policy:

> If a fidelity-diff docket row is fixed, shrink `KnownDiffs` and add the method to `PinnedExact` in
> the same PR.

So this was **documented but never enforced** — which is exactly why 46 of 143 rows drifted. This PR
does not invent a new rule; it makes an existing one mechanical. Both docs now name
`DocketRowsStayCheckedDiffs` as the gate that enforces it.

### Non-vacuity

Adding `CheckedAdd` (a `PinnedExact`, i.e. `Exact`, method) to `KnownDiffs` fails the new test with
`CheckedAdd: now recompiles Exact, so its docket row is stale and silently allows a future
regression.` Reverting it passes. The gate is not vacuous.

richlander:
Ready to merge

Final head `28dd2fea`.

- PR CI: `decompiler-gates` pass, `test (ubuntu-26.04, linux-x64)` pass, `markdownlint` pass.
- Deep Inspect `test` lane on Linux, dispatched twice on this branch (`b3b19ea6` and `6194205c`, which is `28dd2fea` minus a docs-only commit): **`ILInspector.Decompiler.Tests  Total: 4293, Errors: 0, Failed: 0`** both times. That is the 4291 baseline plus exactly the two tests this PR adds. These gates are `Speed=Slow` and do not run in PR CI, so this was the necessary validation — and it settles the toolchain-sensitivity caveat I raised in the PR body.
- Adversarial review: MAI-Code and GPT-5.6-Sol, isolated worktrees, both **CLEAN** on the initial head and again on the fixed head. Three NITs raised, all three fixed (`6194205c`, `28dd2fea`). Reconciliation above.
- Non-vacuity demonstrated: docketing an `Exact` method fails the new gate with the intended message.

No follow-up work is required for this change. Related but non-blocking, tracked separately: #3542 (no Windows CI leg, which is why the `Speed=Slow` gates and #3526's CRLF failures are only ever observed on one platform).

PR #3601, opened 2026-07-30
docs: record the hardened untrusted-input model

richlander:
## What this is

Two documents, no product code. It writes down the untrusted-input model that
came out of the discussion following PR #3518, so the next person changing a
parser or a renderer does not have to re-derive it.

The gap this closes: `untrusted-data-threat-model.md` listed controls and
rules, but never stated what the tool is *trying to be*, how untrusted-input
work ranks against ordinary correctness, or which strategy to reach for when a
contract is violated. #3518 shipped the sanitize strategy by default largely
because nothing said otherwise.

## Threat model changes

**Scope and priority (new section).**

- In scope: harm caused by untrusted input arriving over the internet.
  Out of scope: deliberately opening artifacts you already know are hostile,
  and running elevated.
- Target is **two to three nines, not five** -- stated so the ordering below is
  a real ceiling rather than an aspiration.
- Work ranks: (1) reliability and security on **correct** binaries, (2) security
  on malformed binaries, (3) reliability against environmental patterns that
  need machine access. Correct binaries lead because rigor there is what makes
  reasoning about malformed ones possible.
- A crash on malformed input is a **reliability** defect -- still worth fixing,
  because it means the code is one step from the same input producing an effect
  that does not announce itself.
- Local machine weirdness is tier 3. A symlink in a package cache is a user
  doing user things. It would be a security issue only if a package could
  create it during extraction, and that is NuGet's restore, not this tool.

**Strategy: reject, do not sanitize (new section).** Two reasons: a sanitizer
has to be right about the full set of dangerous forms and is silent when wrong,
and where there is one mouse there are many -- a hostile field is evidence about
the artifact, not about the field. The consequence is a stack resilient to
**errors** rather than resilient to handling bad input.

It also draws the line that keeps this usable: **uniform, lossless encoding on
the way into a sink is not sanitization.** Both have to know which characters
the sink interprets -- a terminal has no formal grammar, so this is not free --
but the difference is what happens when you are wrong. Under-encode and the fix
widens one set in one place, with nothing lost and a decoder still able to
recover the original. Under-sanitize and the data is already gone and the
judgment is spread across every call site that made it.

Where the decision belongs: a type whose construction *is* the check, so
choosing the type grants the capability and auditing is a search rather than an
argument. `HardenedJson` is named as the reference shape.

**Failure messages carry no artifact data (new section).** A rejection names the
user-supplied input, the rule, and the location -- not the value. Attribution is
satisfied by naming what the user wrote. Derived-paths rule #7 matches.

**Presentation** is specified as two orthogonal axes, each with its own tiers.

Trust -- what happens when a concerning pattern is found:

| Flag | Behavior |
| --- | --- |
| *(default)* | abort at the first one |
| survey mode | keep going; report location and pattern kind, never content |
| a `dangerously`-named skip | keep going and render the values anyway |

Rendering -- how artifact text is spelled once something is printed:

| Flag | Behavior |
| --- | --- |
| *(default)* | visually encoded into an inert form |
| a `dangerously`-named raw mode | the artifact's bytes, unmodified |

The axes are independent, and that is the design. Visual encoding is the default
on **every** artifact-text path, *including underneath the trust-axis skip* --
which is precisely what makes that skip defensible: it means "do not refuse,"
not "attack my terminal." A live control character requires opting out of both.

**Precedents:**

- **BSD `vis(3)`** -- where the vocabulary and the contract come from. Encodes
  arbitrary input into graphic characters only, and pairs the encoder with a
  decoder (`unvis`) so the transform is unique and invertible.
  Encode-without-a-decoder is not this pattern.
- **Caret notation** -- `^[` for ESC, `^?` for DEL. The standard spelling,
  used by `cat -v`, `less`, and `stty`, dating to the PDP-6 up-arrow that
  the 1967 ASCII revision replaced with `^`.
- **`grep`** -- refuses binary content by default, prints `Binary file X
  matches` (location, never content), with `-a`/`--text` as the named
  opt-in, for exactly this threat.
- **`less`** -- caret notation by default, raw reserved for `-r`.
- **`rustc`** -- bidirectional control characters became a deny-by-default
  hard error after Trojan Source (CVE-2021-42574), rather than being stripped.
- **`binutils`** -- the cautionary case rather than a model: continuously
  fuzzed, with fuzzing repeatedly finding parser defects that received CVEs.

And `less`'s one mistake, not to be copied: its protection is TTY-conditional
and degrades to `cat` when stdout is redirected, where `-r` makes no
difference because nothing is being encoded in the first place. A pipe is
precisely where an agent or a log is. Measured:

```
default:  033   [   3   1   m   X  \n
-r     :  033   [   3   1   m   X  \n
```

**Open work** gains three items: adopt the strategy where the product currently
neutralizes (10), audit failure messages for artifact data (11), establish
fuzzing (12). Item 9 and the local-adversary residual-risk paragraph are marked
tier 3.

Item 11 records a live violation and three **verified** graph-resolved paths
into `NuGetCache.ValidatePathComponent`, which throws
`Invalid {name}: '{value}'`:

- `ProjectCommand` -> `ProjectAssetsParser` -> `ExtractPackageAsync`
- `NuspecParser` -> `PackageDependency.Id` ->
  `DependencyResolutionService.ResolveDependencyTreeAsync` ->
  `TryGetNuspecXmlAsync` -> `TryGetCachedPackage`
- a package-authored `DotnetToolSettings.xml` `Id` reaching acquisition and
  cache validation

The second leaks twice and reaches a package the user never named:
`DependencyResolutionService` logs `dep.Id`/`dep.Version` *before any
validation*, then catches `Exception` and logs `ex.Message`, re-emitting the
rejected value -- and returns an empty result, the success-shaped failure the
doc forbids elsewhere. `ValidatePathComponent` does not reject control
characters other than NUL, so an ESC passes it outright. This is why the doc
says printability is a function of **provenance, not content**.

## metadata-table-projection.md changes

The worked example the threat model points at: the two axes with concrete flags
(`--survey`, `--dangerously-skip-checks`, `--dangerously-print-raw-text`),
and a specified encoding.

The encoding is `vis(3)`'s, which introduces with a backslash and puts caret
notation inside it:

| Input | Spelling |
| --- | --- |
| C0 (`U+0000`-`U+001F`) | `\^` + (code point + `0x40`); ESC is `\^[` |
| DEL (`U+007F`) | `\^?` |
| C1 (`U+0080`-`U+009F`) | `\u` + four hex digits |
| literal `\` | `\\` |

The backslash introducer is the whole point, and the doc says so, because the
obvious simplification is broken: caret notation alone is **not invertible**.
`U+001E` (RS) is `0x1E + 0x40 = 0x5E`, which *is* `^`, so RS spells
`^^` -- identical to any escape of a literal caret. `cat -v` lives with
this because it never claims reversibility; `vis(3)` avoids it by introducing
with a character outside the caret image. C1 diverges from `vis(3)`'s meta
notation because a .NET `string` holds code points, not high-bit bytes, so
`\uXXXX` is the honest spelling.

Invertibility is verified, not asserted:

```
single code units: round-trip OK (65536)
hostile alphabet up to length 3: round-trip + injective OK (7239 strings, 7239 distinct encodings)
caret-introduced: U+001E -> '^^', literal '^' -> '^^'  => collide = True
vis(3)-introduced: U+001E -> '\^^', literal '^' -> '^'  => collide = False
```

The doc names that gate -- decoder, full-range round-trip, and an injectivity
sweep over the colliding alphabet -- rather than asserting invertibility bare.

The version-stamp narrative argues for what the #3518 bug actually
demonstrates, rather than for the fix it received: containment applied by
*calling a function* is containment you can forget, and `string` is the type
of both a contained and an uncontained value, so no reviewer or compiler can
see the difference.

## What is claimed

Nothing about product behavior.

The Safety section states what the code does today and names the gate for each
property: bounded traversal returns a successful projection with an explicit
`Truncation` marker (**not** a typed rejection -- the contract's rejection
model is not implemented), gated by
`RowBudget_TruncatesExplicitly_NeverSilently`,
`StringBudget_ProjectsBoundedPreviewWithExplicitTruncation`, and
`BlobBudget_ProjectsBoundedHexPreviewWithExplicitTruncation`; heap opt-in
gated by `MetadataLens_NoVerbosity_RendersAnyHeapListing`. Parse-never-load is
an architectural constraint with no single gate naming it and is **marked
unverified** rather than asserted.

A **Status** subsection separates those from the target model, and the threat
model's Presentation section opens with a Status fence pointing at open work 10.
`MdiContainmentTests` gates the *current* behavior; adopting the model changes
what it asserts, not whether the property is gated.

## Validation

Docs only, no measured behavior claim, so per AGENTS.md this needs Markdown
validation rather than product builds or tests. `npx` is unavailable here, so
`markdownlint` runs in CI; MD022/MD058/MD009/MD012/MD024/MD047 and MD051
anchor resolution were checked by hand against `.markdownlint.yaml` (MD013,
MD032, MD034, MD060 disabled in this repo). The encoding and `less` claims
were verified by measurement as shown above.

## Review

Three independent reviews on `9f406573` -- MAI-Code, Gemini 3.1 Pro, and
GPT-5.6 Sol. Reconciliation is posted as a comment below.


richlander:

Three models reviewed the fixed head, each isolated in its own worktree: **MAI-Code**, **Gemini 3.1 Pro**, and **GPT-5.6 Sol**. Everything actioned is in `f385d873`.

### Found by more than one reviewer

**`Cf` sweeps in legitimate text.** All three of us landed on this independently. The doc claimed the rule `leaves ordinary text — including CJK, diacritics, and ligatures — untouched`. Measured, `Cf` also contains `U+200C` ZWNJ, `U+200D` ZWJ, `U+2060` WORD JOINER and `U+00AD` SOFT HYPHEN, so Persian `می‌خواهم`, Devanagari `क्‍ष`, and emoji ZWJ sequences all get encoded. The example list was three cases that happen to pass, with the failing cases omitted.

GPT took it further and found the decisive fact: **`ApiOutputFormatter.IsBidiControl` already implements the narrower `Bidi_Control` set deliberately**, with the reason in a comment — *"a zero-width joiner or a BOM does not reorder its neighbors, and legitimate identifiers may contain format characters, so escaping all of Cf would corrupt ordinary names."* C# admits `Cf` in `identifier_part_character`, so that reasoning holds.

The doc no longer decides this. It states both readings — reordering and invisibility are different attacks — and defers, because one reading is already shipped and this document should not silently overrule it.

**Malformed open-work list.** Gemini and MAI-Code both caught a duplicated item 11 merged into item 10. Root cause turned out to be a bare `CR`, which folds into the next finding.

### Found by GPT-5.6 Sol alone

**Line endings — the one with the widest blast radius.** `core.autocrlf=true` plus my raw file writes had rewritten both files in CRLF against LF originals. `git diff --check` reported **1,095 whitespace errors** and the diff covered every line in both files. Normalized to LF; the true diff is **451 insertions / 36 deletions**. This also explains the malformed list — one bare `CR` survived the conversion.

**`rustc` denied nine, not twelve.** `TEXT_FLOW_CONTROL_CHARS` is `U+202A`–`U+202E` and `U+2066`–`U+2069`. The twelve I cited is Unicode's `Bidi_Control` property, which adds `U+200E`, `U+200F`, `U+061C`. I had misattributed the wider set to `rustc`. Confirmed against the Rust security advisory. This matters because the `rustc` precedent was requested specifically.

**Scalar versus UTF-16 code unit.** Every non-BMP character is two `Cs` code units, so a per-`char` category test encodes every emoji — 😀 becomes `\uD83D\uDE00`. Only *unpaired* surrogates are hazards. The gate now requires a paired-surrogate case, since this bug ships looking correct.

**"Inert" was overstated.** Visual encoding stops terminal interpretation and bidi reordering. It does not make text safe for a structured format: no encoded category contains `|`, a backtick, or a quote. Structural escaping is the serializer's obligation, composes with visual encoding, and stays mandatory in every mode — so raw mode no longer claims to emit unmodified bytes, which would produce invalid JSON.

**`HardenedJson` is not the shape I claimed.** It is a `static class` whose `Parse` returns a plain `JsonDocument` — a centralized entry point, not a type whose construction is the check. I had cited it as the reference for the stronger shape while criticizing exactly that weaker shape elsewhere in the same document. Two feed readers already bypass it: two `JsonDocument.Parse` call sites in `PackageExtractor` registration-page reading, and `NuGetFetch.NuGetApi`'s source-generated contexts. The claim that only `runfaster` was uncovered was false, and nothing gates the invariant — now open work.

**Named gates were wrong, which the repo treats as worse than naming none.** `StringBudget_ProjectsBoundedPreviewWithExplicitTruncation` asserts explicit truncation, not the bound — the bound is `StringPreview_NeverExceedsCharBudgetEvenWhenEscaped`. `MetadataLens_NoVerbosity_RendersAnyHeapListing` covers only `-v:q` and `-v:d`. `MdiContainmentTests` carries a `Cc`-only payload, permits raw `CR`/`LF`/`TAB`, and covers no delimiters. All three now say what they actually enforce.

Also actioned: survey output is bounded by the traversal budget rather than exhaustive; `EscapeCore` formats with `X4` so the collision literal is uppercase `\u001B`; and three provenance paths sat under a sentence saying two.

### Dismissed

Gemini's third finding asked for named gates on `Assemblies are parsed, never loaded`, `HardenedJson` duplicate rejection, and the SSRF client. Those sections are **pre-existing text this PR does not touch**, and the parse-never-load property is already explicitly marked unverified. Real, but out of scope for a PR that would otherwise grow without bound — worth a follow-up issue.

### Verified and held up

- Encoder injectivity. Brute-forced by all three reviewers independently: 65,536 single code units plus hostile-alphabet strings, zero collisions, full round-trip. The `vis(3)` backslash introducer correctly avoids the `U+001E` → `^^` collision that killed the earlier caret-introduced draft.
- Category arithmetic: 65 `Cc` + 43 `Cf` + 2,048 `Cs` + 1 `Zl` + 1 `Zp` = **2,158**.
- Every codebase claim about `IsControl`, `NeutralizeControls`, the truncation marker, `ValidatePathComponent`, and the `DependencyResolutionService` double-leak.
- `vis(3)`, caret notation, and the redirected-`less` behaviour.

### Earlier round, for the record

The first MAI-Code pass returned two findings and **both were wrong**. It claimed `less` protection was not TTY-conditional — disproved by running `less` and `less -r` with stdout redirected and dumping with `od -c`; both emit a raw `033`. It claimed the `ValidatePathComponent` provenance chain was unsupported — disproved by tracing call sites outward, which found a live path from a transitively-fetched nuspec. Both claims stand as written.

richlander:
## The obligation list is now incomplete: how treated text is *carried*

Implemented the predicate/speller split in #3563 (`InertText`), and one thing came out of it that
the spec does not currently say. Flagging rather than pushing to this branch, since it is actively
owned — say the word and I will commit it directly instead.

### What changed on my side

`VisualEncoder` stayed exactly as the six obligations describe it. What was added is a second API
shape on top: the result of encoding is carried as a value, `InertString`, rather than as a
`string`.

The argument is auditability, and it is narrow. The speller alone is *transactional* — a `string`
goes in, a `string` comes out — so a treated value and an untreated one have the same type.
Deciding whether a sink is safe means tracing every call path that reaches it, and redoing that
after every change. A distinct type turns the question into a type search.

That is not theoretical. Retyping one return value surfaced a live hole:

```csharp
// before: string?
lines.Add($"Package '{packageName}' could not be resolved ...");
```

`packageName` arrives from a command line or a dependency graph, and was interpolated untouched
into a printed message. Changing the return type to `InertString?` made it a build error at both
call sites. A review pass had not caught it — including my own, twice.

### Proposed delta

A seventh bullet in **"The speller belongs in a shared, dependency-free component"**:

> - **Able to be carried as a value, not only returned as a `string`.** The encoded result is
>   available as a distinct type that can only be constructed by applying a predicate. A sink
>   that accepts only that type cannot be handed untreated text by accident, which turns "is
>   this path safe" from a call-graph trace into a type search. There is no conversion *from*
>   `string`. Conversion *to* `string` is unrestricted, and safe here in a way it usually is
>   not: the customary objection is that `ToString` launders the wrapper, but that assumes the
>   payload is dangerous and the wrapper restrains it. Here the payload is already inert, so
>   losing the wrapper loses provenance, not protection.

And a note under it, because composition is where this shape usually dies:

> Composition has to be part of the contract or callers fall back to `$"...{treated}..."` and
> drop the guarantee at the moment it matters most. An interpolated string handler that applies
> the predicate to each part as it is appended covers this. It must encode *literals* as well as
> holes — an invariant with an exception in it has to be re-argued at every use — and must append
> an already-treated part as-is rather than encoding it twice.

### Two things for you to arbitrate

1. **Terminology collision.** I have been calling this the "currency" form, but this document
   already uses *currency* for section addressing (§"Heap addressing ... does introduce a new
   currency"). Worth picking a different word here, or explicitly noting the two senses.

2. **Whether this binds `--survey` / `--dangerously-skip-checks`.** Both render to a terminal, so
   both are exactly the sinks that would benefit from accepting the type rather than `string`.
   That is your call, not mine — I have only applied it to the NuGet failure path.

Not proposing any change to the six existing obligations. `InertString.Forms` satisfies "able to
report which forms it emitted" more directly than the `out` parameter did, and the policy-free
rule is untouched: the type takes the predicate as input, exactly like the speller.

richlander:
Pushed as `e6ccd412c` now that you're idle — revert or reword freely, it's your document.

Resolved the two open points myself rather than leaving them dangling:

1. **Terminology.** Avoided "currency" entirely, since this document already uses it for heap addressing. The bullet reads *"Available as a value, not only as a returned `string`"*.
2. **Scope.** Wrote it as a contract on the shared component only. I did **not** claim it binds `--survey` or `--dangerously-skip-checks` — those are your modes, and that call is still open.

The log delegate this does *not* yet cover is now #3606.

richlander:
Measured evidence for the central claim here, from #3319/#3380 and the leak it turned up (#3628).

This document argues that a sanitizer "has to be right about the full set of dangerous forms" and "is silent when wrong." The codebase now has four independent hazard predicates. Three agree on Cf and Zl/Zp; the fourth — `MetadataTableProjector.IsControl`, C0+DEL+C1 only — does not, and consequently a hostile type name reaches stdout as raw `e2 80 ae`. Nothing failed. No test, gate, or analyzer knew that predicate existed, so the silence was total until a hostile assembly happened to be pointed at that one lens.

The failure also matches the document's *reason* for preferring encoding, not just its conclusion. The three correct predicates were written by someone who had thought about bidi; the fourth was written by someone reasonably reaching for "control character." Judgment spread across four call sites produced four different answers, which is the predicted outcome — and the correction is not "teach the fourth about Cf" but "stop having four."

One caveat about existing code, also noted on #3563: `CSharpIdentifierCore.ContainComposedName` starts with `ReplaceLineEndings(" ")`, which is lossy and non-injective. It should be read as a prospective `VisualEncoder` consumer, not as an existing implementation of this contract.

PR #3602, opened 2026-07-30
Trace which scanners and sections a library run actually executes

richlander:
## What

Adds `--trace` to the `library` command. It writes a plain diagnostic report to **stderr**; stdout markout is untouched and byte-identical.

## Why

The section pipeline is demand-driven — a section declares a scanner key, the registry expands prerequisites, only the closure runs. That is the design, but nothing observed it. "The system does the correct minimum work" was a claim resting on reading the code.

The next change in this area (moving cost declaration from sections onto scanners) needs to prove a before/after difference in **what runs**, not just in what renders. This is the instrument for that.

## The report

Default `-v:m` against `ILInspector.Decompiler.dll`:

```
trace: library ILInspector.Decompiler.dll [Minimal]
  sections demanding a scanner
    Library Info -> InfoCounts
  scanners requested   InfoCounts
  added by prerequisite ClassifiedMethods, CustomAttributes, ExtensionMethods, Resources, TypeForwarders
  scanners executed
    ExtensionMethods            11.4 ms
    ClassifiedMethods           53.1 ms
    Resources                    2.1 ms
    CustomAttributes             9.1 ms
    TypeForwarders               2.0 ms
    InfoCounts                   0.0 ms  (bundle, no work of its own)
  resources acquired
    metadata session            borrowed from the command's open image
  total scanner time   77.8 ms
```

The same assembly at `-v:d`: **6,880 ms**, of which `OptimizationOpportunities` is **4,728 ms**, `ResourceTriage` 1,273 ms, `Switches` 753 ms; resources show `body index built in 2,153.4 ms` and `drill map built in 1,263.5 ms (7,463 members)`.

Those sections — the eight `Performance: *` plus `Array Pool Escapes` — declare `Cost = NetworkFree`. That is the defect the next PR fixes, and this is the measurement it will be judged against.

## Design decisions worth reviewing

**The trace is a nullable object, not a flag.** An untraced run allocates nothing and pays one null check per site. There is no second code path. `Tracing_DoesNotChangeTheWorkTheRunDoes` holds `SharedScanCount` equal with and without a trace, so the observer cannot start perturbing what it observes.

**Resources are recorded at acquisition, not at request.** Absence is the observable fact: a metadata-only scan that quietly starts building the IL body index costs seconds and changes no output, so no other gate in the suite would notice. `Trace_RecordsNoBodyIndexForAScanThatDoesNotNeedOne` is the minimum-work gate; `Trace_RecordsTheBodyIndexWhenAScannerActuallyBuildsIt` is its paired positive — without it, a recorder that never records anything passes.

**Failed acquisition is recorded** (`FAILED after X ms: <Type>`). Scanners swallow a failed body index and render an empty section, so without this line "tried and failed" is indistinguishable from "correctly never needed it".

**Bundles are marked.** `InfoCounts` does no work of its own; attributing its prerequisites' time to it would be a lie in the one place the report is meant to be trusted.

**Written in a `finally`**, so a failing run still says what it did before failing.

**Section→scanner attribution exists only on the demand side.** Downstream the registry sees a bare set of keys, and the report does not pretend otherwise.

**Scoped to the command that implements it.** `--trace` is registered on the `library` command directly, not in the shared `AddOutputOptionsTo` helper. A flag every command accepts and only one honours is worse than an unrecognized argument, which at least fails loudly. `dotnet-inspect package <id> --trace` now reports `Unrecognized option '--trace'`.

**Tests assert on the `InspectionTrace` object, not on rendered text**, so the report can be reformatted without rewriting the gates.

## Evidence

- `dotnet build dotnet-inspect.slnx -c Release` — clean, 0 errors, 6 pre-existing warnings.
- `dotnet run --project src/dotnet-inspect.Tests -c Release` — **2587 total / 2 failed / 10 skipped**. Both failures (`Layout_Count_CountsFilesRatherThanRenderedTreeLines`, `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown(readmePath: "images/logo.png.")`) reproduce at clean `origin/main`.
- stdout **SHA-256 identical** with and without `--trace` at `-v:n` (103,620 bytes).
- `--trace` rejected as unrecognized on `package`; accepted and honoured on `library`.

## Boundary

`PackageCommand` also calls `GetRequiredScanners` and is deliberately **not** wired for tracing here — it shares the registry, so extending it is additive and belongs in its own change.

### DIFF
--- a/docs/design/section-pipeline.md
+++ b/docs/design/section-pipeline.md
@@ -189,6 +189,60 @@ When the pipeline computes multiple sections and the output format is the defaul
 
 This is tracked on options objects, which distinguish the default format from an explicit tabular flag.
 
+## Tracing
+
+`--trace` writes a diagnostic report to **stderr** describing the work a run actually did. stdout is
+untouched, so a caller parsing the document is unaffected (gated by a byte-identical stdout check).
+
+The report answers "did this run do the correct minimum work?", which nothing else can:
+
+```text
+trace: library ILInspector.Decompiler.dll [Minimal]
+  sections demanding a scanner
+    Library Info -> InfoCounts
+  scanners requested   InfoCounts
+  added by prerequisite ClassifiedMethods, CustomAttributes, ExtensionMethods, Resources, TypeForwarders
+  scanners executed
+    ExtensionMethods            9.3 ms
+    ClassifiedMethods           48.6 ms
+    ...
+    InfoCounts                  0.0 ms  (bundle, no work of its own)
+  resources acquired
+    metadata session            borrowed from the command's open image
+  total scanner time           74.4 ms
+```
+
+Five things are recorded, each at the only layer that knows it:
+
+| Fact | Recorded by |
+| --- | --- |
+| Which section demanded which scanner | `SectionPipeline.GetRequiredScanners` |
+| What prerequisite expansion added | `ScannerRegistry.ExpandRequired`, via `InspectAsync` |
+| Which scanners ran, in order, with timings | `ScannerRegistry.RunScanners` |
+| Whether a scanner is a bundle (no work of its own) | `ScannerRegistry.RunScanners` |
+| Which expensive resources were built | `ScannerContext.Session`/`BodyIndex`/`DrillMap` |
+
+Section-to-scanner attribution exists only on the demand side — downstream the registry sees a set of
+keys with no memory of who asked for them — which is why it is captured there rather than
+reconstructed later.
+
+Design points worth keeping:
+
+- **The typed record is the contract, not the text.** Tests assert on `InspectionTrace`, so the report
+  can be reformatted without rewriting gates.
+- **Resources are recorded at acquisition, not at request.** A resource that never appears was never
+  built. That absence is the observable: a regression that makes a metadata-only scan open the
+  whole-assembly IL index costs seconds and changes no output, so no other test would notice it.
+- **A failed acquisition is recorded too.** Scanners swallow a failed body index and render an empty
+  section; without a `FAILED` line, a run that tried and failed would look exactly like a run that
+  correctly never needed the index.
+- **Untraced runs pay nothing.** Tracing is threaded as a nullable object rather than a flag, so an
+  untraced run allocates nothing and takes no branch beyond a null check. A gate holds the shared
+  scan count equal with and without tracing, because a diagnostic that perturbs what it measures is
+  worse than none.
+- **The report is written in a `finally`.** A failed run still reports what it had done before
+  failing, which is when the question is most worth asking.
+
 ## Design Decisions
 
 **Static abstract interfaces over instances.** Descriptors are never instantiated. The pipeline extracts static members into delegate-based `SectionEntry<T>` records at registration time. This avoids allocations and keeps metadata in a format that NativeAOT can optimize.
--- a/docs/design/section-pipeline.md
+++ b/docs/design/section-pipeline.md
@@ -235,8 +235,10 @@ Design points worth keeping:
   command asked for it directly (today only discovery mode's metadata row counts), or a declared
   prerequisite pulled it in. Collapsing the second into the third would render a prerequisite edge
   that does not exist, and send anyone chasing an unexpected scan to the wrong declaration.
-  `Trace_AttributesEveryAddedScannerToARealPrerequisiteEdge` walks the registry's declared edges and
-  fails if anything shown as expansion is not actually reachable from a requested key.
+  `Trace_ExplainsEveryScannerThatRan` seeds reachability from what the trace *claims* — the recorded
+  section and command demands — and walks the registry's declared edges to reach the closure the run
+  actually produced. That asymmetry is what makes it a gate: seeding from the requested set instead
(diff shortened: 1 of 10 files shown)


richlander:
## Review reconciliation

Reviewed by **MAI-Code**, adversarially, across three fixed heads in an isolated detached worktree. Tier: more-than-trivial, not high risk — diagnostic-only, stdout proven byte-identical — so one review. Final pass: **No blocking findings at `ea5443f8`.**

The review found two real defects. Both are fixed, and the second one was mine to be embarrassed about.

### Finding 1 — the report was lying (`2aee7b58`)

`LibraryCommand` added `ScannerMetadata` to the requested set *after* `GetRequiredScanners` had recorded the trace, so prerequisite expansion picked it up and `-D --trace` rendered:

```
added by prerequisite Metadata
```

No such edge exists. Anyone chasing an unexpected metadata scan would go looking for a declaration that isn't there — in a report whose only value is being trustworthy.

Fixed in `36515596`, structurally rather than cosmetically. The bug was possible because a caller could mutate the requested set after the one method that knows how to record it had finished, so `GetRequiredScanners` now takes a `commandDemand` parameter and the post-hoc `scanners.Add(...)` is gone. Computing and recording happen in one place; there is nothing left to forget. The report gained a `demanded by the command` bucket, shown only when non-empty.

### Finding 2 — my regression gate was a tautology (`36515596`)

The reviewer reported the new gate staying green under tamper. The reason turned out to be worse than "weak test": it seeded reachability from `trace.Requested` and walked prerequisite edges to check `trace.Closure` was explained — but `Requested` *is* what `ExpandRequired` is seeded from, and the walk *is* what `ExpandRequired` computes. The assertion reduced to `X ⊆ X` and could not fail for any input.

I wrote a fresh instance of the exact failure mode this repository keeps re-encountering, in the test whose stated purpose was guarding against one.

Fixed in `ea5443f8` by introducing an asymmetry: the closure comes from what the run *did* (`ExpandRequired` over the returned set), while reachability is seeded from what the trace *claims* (recorded section demands ∪ recorded command demands). A scanner that ran with no recorded demand explaining it is now unreachable and the gate goes red. Renamed to `Trace_ExplainsEveryScannerThatRan` — the old name described the mechanism and flattered the assertion.

The reviewer independently confirmed the gate is now non-tautological with four tampers, each producing a full-suite failure:

| Tamper | Gate result |
| --- | --- |
| Suppress `RecordCommandDemand`, keep `scanners.Add` | FAIL — `["Metadata"]` |
| Suppress `RecordDemand` in the section loop | FAIL — large scanner set |
| Record a *different* scanner name than the one added | FAIL — `["Metadata"]` |
| Inject a bogus key with no demand recorded | FAIL — `["BogusScanner"]` |

### One reviewer conclusion I corrected

The reviewer asked whether an edit to `LibraryCommand.cs` alone leaves the suite green, and reported that it did. That came from running only `-class SectionPipelineTests`. Running the **full** suite with `DiscoveryScanners` replaced by `null` fails seven existing tests: `MetadataLens_BareDiscovery_ListsDoorWithoutMembers`, `MetadataLens_DiscoveryDrillIn_ListsOnlyTablesWithRows`, `MetadataLens_Discovery_AnswersInCanonicalNamesOnly`, `MetadataLens_DiscoveryCatalog_SurvivesCacheRoundTrip`, `MetadataLens_CachedDiscovery_DoesNotBypassHeapCoordinateResolution`, `MetadataLens_Columns_NarrowsTableAndMatchesDiscovery`, `MetadataLens_HeapSection_IsDiscoverableOnlyWithItsCoordinate`.

So the command's behavior is gated by its behavior; the new gate covers the attribution, which nothing else observes. The reviewer verified this independently on the re-review.

### Cleared with evidence

Every attack point was cleared by a real run, not by inspection:

- **Observer does not perturb the observed** — `Tracing_DoesNotChangeTheWorkTheRunDoes` holds `SharedScanCount` equal traced vs. untraced.
- **stdout untouched** — SHA-256 identical with and without `--trace` at `-D` (479 B), `-v:n` (22,323 B), `-v:d` (356,509 B), and `--json` (656 B).
- **Laziness is real** — `-S "Library Info" --trace` reports no body index; `-S "Unsafe Members" --trace` reports one built.
- **Failure paths** — a throwing scanner is still recorded; `library <missing>.dll --trace` emits the original error *and* a trace, exit code unchanged, error unmasked.
- **Option scoping** — `package --trace` → `Unrecognized option '--trace'`; `--trace` appears in `library --help` only.
- **No latent hazard elsewhere** — `PackageCommand` calls `GetRequiredScanners` without post-hoc mutation.
- **Concurrency** — scanner execution is a single sequential loop; the untraced path is a null check with no allocation.

### Validation

`dotnet-inspect.Tests`: **2588 total / 2 failed / 10 skipped**. Both failures (`Layout_Count_CountsFilesRatherThanRenderedTreeLines`, `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown(readmePath: "images/logo.png.")`) reproduce at clean `origin/main`. CI green: `changes`, `markdownlint`, `test`, `build-net10`, `decompiler-gates`.

### Non-blocking follow-up

`PackageCommand` shares the scanner registry and is not traced. Extending tracing to it is additive and belongs in its own change.

Ready to merge

PR #3604, opened 2026-07-30
Gate the two sweep properties that only hand-run probes covered

### LINKED ISSUE #3560: Gate the two sweep properties that only hand-run probes cover

richlander:
The EVIL pool's version pin was gated as a *file* and never as a *run*. `EvilPoolPinTests` checks its shape, encoding, read bounds and refusal wording — none of which starts a sweep. So the two properties the pinning work exists to provide were the two nothing in CI watched:

- a sweep pools the exact bytes `nuget-top-packages.lock.json` names, and fails rather than pool anything else;
- the assembly copies put those bytes at the right path inside the output directory, without writing through whatever is already there, and every package whose copy fails is counted as failed.

Both were evidenced only by probes run by hand and recorded on #3434. That mattered: the two most serious defects sixteen review rounds found were both on this path — a pin that described the *request* rather than the bytes, so a poisoned cache entry satisfied it; and `File.Copy` writing an assembly through a symlink planted at the destination while the sweep exited 0.

`EvilPoolSweepGateTests` runs real sweeps in ~20 seconds. Seven cases, one per probe, plus an eighth that gates the isolation the other seven rest on.

## The issue's premise was wrong, and the change is smaller for it

#3560 said the sweep "hardcodes its inputs" and would need `--package-list`/`--pin-file` or `--data-directory`, flagging that as a call to make deliberately. It needs none of them: the sweep already resolves its inputs from the repository root **above its working directory**, so a scratch directory holding a `dotnet-inspect.slnx` and a `docs/data` feeds a real run a list and pin of the suite's choosing. `EvilPoolPinTests.RunSweep` has done exactly this since #3434; nobody had noticed it reaches acquisition. **No new argument surface.**

What was genuinely missing was isolation. The sweep called `HttpClientFactory.Initialize()` and `NuGetCache.Initialize("dotnet-inspect")` with no arguments, so it always used the shared caches and the real network. It now honors the three knobs the CLI has always read — `DOTNET_INSPECT_OFFLINE`, `DOTNET_INSPECT_ISOLATED`, `DOTNET_INSPECT_CACHE_DIR` — with the CLI's meanings unchanged. That is one program catching up to a convention the rest of the tool follows, not a test backdoor, and it is inert when the variables are unset, which is every real sweep.

Offline is what keeps the gate honest rather than merely fast: a case that stopped being served from the seeded cache would reach for the network and **fail**, instead of quietly passing on whatever the machine happened to have. Verified — pointed at an empty cache, the sweep reports `not available offline` and exits 1.

The precise claim is that no package the sweep pools can come from anywhere but the seeded fixture. It is not that the test opens no socket: each case launches the sweep with `dotnet run`, which restores and builds a file-based app first, and that reads the ordinary NuGet cache and may go to the network to fill it. That is true of every build in this repository and of the sweep runs in `EvilPoolPinTests`; it is the acquisition under test that is sealed.

The seven cases cannot check that isolation themselves: they all pool a synthetic package only the scratch cache holds, so they pass whatever the isolation does. Measured — deleting both knobs from the sweep (`HttpClientFactory.Initialize(false)`, `skipNuGetCache: false`) leaves **all seven green**, and the suite silently regains the shared NuGet cache and the network. The eighth case names a real package instead, `newtonsoft.json@13.0.4`, reachable by either knob's absence and by neither's presence, and asserts the sweep cannot reach it — asserting the *reason* (`not available offline`), because a sweep that found it would still exit 1 on the pin. Each knob dropped alone turns exactly that case red.

The fixture package is committed through `NuGetCache.CommitPackage`. The layout of a committed package, including the marker that makes it count as committed, is the product's; a harness that laid it out itself would be a second implementation and would keep passing after the product's moved.

## Non-vacuity

Every case was proved to bite, and to bite alone. Each tamper is one edit to the sweep; the column is which tests went red.

| Tamper on the sweep | Red |
| --- | --- |
| `File.Copy(overwrite: true)` restored (the round-sixteen defect) | destination case only |
| sha256 comparison disabled | poisoned-bytes case only |
| TFM comparison re-guarded by `pin.Tfm is not null` | TFM case only |
| failed copy counted as accounted-for | accounting case only |
| temporary cleanup disabled | cleanup case only |
| acquisition falls back to a cached version + version check off | version case only |
| acquisition ignores the pin entirely | the clean run, plus 3 others |

The last row is the positive control: a harness that could not produce a *successful* sweep would report every refusal as correct.

**A gap this found in itself.** The leftover-temporary assertion was vacuous as first written. It rode on the unwritable-directory case, which fails at the moment the temporary is created — `created` stays false, the cleanup branch is never reached, and the assertion was that a file which was never made does not exist. A directory standing at the destination is the failure that happens *after*; that case is `ASweepLeavesNoTemporaryWhenAWriteFailsAfterCreatingOne`, and disabling the cleanup turns it red.

## Evidence

- Full solution build **0 errors**.
- `Area=Corpus` **355/0**; `--gate fast` **4402/0**; new class **7/0 in 11s** (`Area=Corpus`, not `Speed=Slow`, so PR CI runs it).
- **Real sweep unchanged**: ranks 1–3 against the committed list and pin, exit 0, same three assemblies. The env change is inert when unset.
- **Shared caches untouched**: `newtonsoft.json` 13.0.4 still `a28c251d…` in both `~/.nuget/packages` and `~/.cache/dotnet-inspect`, and no `sweep.fixture` anywhere in either.

## Boundary

Not covered, deliberately: the `no-library` claim still needs the network, because it asks a real package whether it ships a library — that doc comment is left as it was rather than made to point at this gate. Symlinks planted at *parent* directories remain out of scope per the call on #3434.

Closes #3560

### DIFF
--- a/eng/prepare-decompiler-package-sweep.cs
+++ b/eng/prepare-decompiler-package-sweep.cs
@@ -390,8 +390,9 @@ catch (Exception ex) when (ex is IOException or UnauthorizedAccessException or N
 // honored here so a caller can point this sweep at a cache of its own. Without them the
 // sweep reaches the developer's shared caches and the network unconditionally, which is
 // why its two central properties -- that the pin binds, and that the copies land where
-// told -- could only ever be evidenced by hand-run probes (#3560). This is not a test
-// backdoor: it is one program catching up to a convention the rest of the tool follows,
+// told -- could only ever be evidenced by hand-run probes (#3560). Those properties are
+// now gated by EvilPoolSweepGateTests, which runs this file offline against a scratch
+// cache. This is not a test backdoor: it is one program catching up to a convention the rest of the tool follows,
 // with the CLI's meanings unchanged. An isolated session skips the shared NuGet cache
 // and, absent an explicit directory, gets its own. Reading them costs nothing when they
 // are unset, which is the case for every real sweep.
--- a/eng/prepare-decompiler-package-sweep.cs
+++ b/eng/prepare-decompiler-package-sweep.cs
@@ -1179,6 +1179,13 @@ static async Task<WriteOutcome> ReplaceOrReport(string path, Func<FileStream, Ta
     // directory can occupy first, and a write through a planted symlink lands wherever
     // that link points. CreateNew refuses to open anything that already exists, so the
     // sweep either makes this file itself or writes nothing.
+    //
+    // Reasoned, not gated, and marked as such. The destination half of this is gated --
+    // EvilPoolSweepGateTests plants a symlink at the destination and requires a real file
+    // afterwards -- but the temporary half is not, because the name is drawn inside this
+    // function and never leaves it, so nothing outside can occupy it except by racing a
+    // loop against the allocator. Treat the CreateNew above as load-bearing: it is the
+    // whole of the protection, and no test would notice it becoming Create.
     string temporary = path + $".{Path.GetRandomFileName()}.tmp";
     bool created = false;
     try
--- a/eng/prepare-decompiler-package-sweep.cs
+++ b/eng/prepare-decompiler-package-sweep.cs
@@ -1200,7 +1200,11 @@ static async Task<WriteOutcome> ReplaceOrReport(string path, Func<FileStream, Ta
         // though -- something already sitting at that name is not this sweep's to
         // remove, and deleting whatever is found there would be a worse answer than
         // the failure being reported.
-        string fate = "none";
+        // Assumed left behind the moment it exists, and downgraded only by a delete that
+        // returned. Derived the other way round -- "none" until something says otherwise --
+        // a cleanup that stopped being called would report that no temporary was ever
+        // created, which is the one answer that makes the leak invisible.
+        string fate = created ? "left-behind" : "none";
         try
         {
             if (created)
@@ -1211,7 +1215,6 @@ static async Task<WriteOutcome> ReplaceOrReport(string path, Func<FileStream, Ta
         }
         catch (Exception cleanup) when (cleanup is IOException or UnauthorizedAccessException)
         {
-            fate = "left-behind";
         }
 
         return new WriteOutcome($"Could not write '{path}': {ex.Message}", fate);
--- a/eng/prepare-decompiler-package-sweep.cs
+++ b/eng/prepare-decompiler-package-sweep.cs
@@ -759,31 +759,20 @@ if (unwritable is not null)
 // deep-inspect workflow writes its acquisition log into the output directory, beside
 // packages/ rather than inside it, and it is not this step's to remove.
 //
-// Every removal is recorded in the manifest, not merely printed. This step deletes
-// unrecorded files under packages/, and a write temporary this run leaked is an
-// unrecorded file under packages/ -- so a reconciliation that only cleaned up would
-// quietly erase the evidence of a leak, and the disk check in EvilPoolSweepGateTests
-// that used to catch one would pass over a swept pool. Measured: with the removals
-// unreported, a sweep that skipped its temporary cleanup while still reporting it as
-// "removed" left all seventeen cases green. Recording them keeps the fact the check was
-// looking for, and puts it where an operator reads it: a run that had to remove
-// anything is a run where something else went wrong.
+// How the walk itself is bounded, and why every removal is recorded, is on ReconcilePool.
 var removedFromPool = new List<string>();
+
+// Ordinal, because these are paths on a filesystem that distinguishes case, and
+// comparing them any other way would let an unrecorded file whose name differs from a
+// recorded one only in case pass for the recorded one and stay in the pool. That is
+// unverified: no case here plants such a file, because whether the assertion means
+// anything depends on the filesystem the suite happens to run on.
 var recordedAssemblies = new HashSet<string>(
     assemblies.Select(Path.GetFullPath), StringComparer.Ordinal);
 string? unreconciled = null;
 try
 {
-    foreach (string pooled in Directory.GetFiles(packageDirectory, "*", SearchOption.AllDirectories))
-    {
-        if (recordedAssemblies.Contains(Path.GetFullPath(pooled)))
-            continue;
-
-        File.Delete(pooled);
-        removedFromPool.Add(pooled);
-        Console.Error.WriteLine(
-            $"removed '{pooled}' from the pool: this sweep did not record it.");
-    }
+    ReconcilePool(packageDirectory, recordedAssemblies, removedFromPool);
 }
 catch (Exception ex) when (ex is IOException or UnauthorizedAccessException)
 {
@@ -793,6 +782,61 @@ catch (Exception ex) when (ex is IOException or UnauthorizedAccessException)
     unreconciled = $"Could not reconcile the pool under '{packageDirectory}': {ex.Message}";
     Console.Error.WriteLine(unreconciled);
     Environment.ExitCode = 1;
+}
+
+// Walks the pool without following links, and removes what the run did not record.
+//
+// The walk is written out rather than done with Directory.GetFiles(..., AllDirectories)
+// because that enumeration descends through a directory symlink and yields the paths it
+// finds on the far side. This step deletes what it is handed, so that enumeration aims a
+// deletion loop at whatever a link in the pool points at, and then reports success --
+// measured: a link planted beside packages/ had the file behind it deleted and the run
+// exited 0. A link is removed as a link instead, which is a deletion this step can bound
+// to the pool it owns: removing a symlink never touches its target.
+//
+// Every removal is recorded by the caller, not merely printed. This step deletes
+// unrecorded files under packages/, and a write temporary this run leaked is an
+// unrecorded file under packages/ -- so a reconciliation that only cleaned up would
+// quietly erase the evidence of a leak, and the disk check in EvilPoolSweepGateTests
+// that used to catch one would pass over a swept pool. Measured: with the removals
+// unreported, a sweep that skipped its temporary cleanup while still reporting it as
+// "removed" left all seventeen cases green. Recording them keeps the fact the check was
+// looking for, and puts it where an operator reads it: a run that had to remove
+// anything is a run where something else went wrong. The stderr line beside each
+// recorded removal is an operator convenience and is not itself gated.
+static void ReconcilePool(string directory, HashSet<string> recorded, List<string> removed)
+{
+    foreach (FileSystemInfo entry in new DirectoryInfo(directory).EnumerateFileSystemInfos())
+    {
+        // Non-null for a symlink alone, so a real directory is descended into and a hard
+        // link -- indistinguishable from the file it is, and whose removal frees only
+        // this name -- is treated as the file it appears to be.
+        if (entry.LinkTarget is not null)
+        {
+            entry.Delete();
+            Record(entry.FullName);
+            continue;
+        }
+
+        if (entry is DirectoryInfo child)
+        {
+            ReconcilePool(child.FullName, recorded, removed);
+            continue;
+        }
+
+        if (recorded.Contains(Path.GetFullPath(entry.FullName)))
+            continue;
+
+        entry.Delete();
+        Record(entry.FullName);
+    }
+
+    void Record(string path)
+    {
+        removed.Add(path);
+        Console.Error.WriteLine(
+            $"removed '{path}' from the pool: this sweep did not record it.");
+    }
 }
 
 removedFromPool.Sort(StringComparer.Ordinal);
--- a/eng/prepare-decompiler-package-sweep.cs
+++ b/eng/prepare-decompiler-package-sweep.cs
@@ -386,8 +386,27 @@ catch (Exception ex) when (ex is IOException or UnauthorizedAccessException or N
     return;
 }
 
-HttpClientFactory.Initialize();
-NuGetCache.Initialize("dotnet-inspect");
+// The same isolation knobs the CLI already reads (src/dotnet-inspect/Program.cs),
+// honored here so a caller can point this sweep at a cache of its own. Without them the
+// sweep reaches the developer's shared caches and the network unconditionally, which is
+// why its two central properties -- that the pin binds, and that the copies land where
+// told -- could only ever be evidenced by hand-run probes (#3560). This is not a test
+// backdoor: it is one program catching up to a convention the rest of the tool follows,
+// with the CLI's meanings unchanged. An isolated session skips the shared NuGet cache
+// and, absent an explicit directory, gets its own. Reading them costs nothing when they
+// are unset, which is the case for every real sweep.
+bool offline = string.Equals(
+    Environment.GetEnvironmentVariable("DOTNET_INSPECT_OFFLINE"), "1", StringComparison.Ordinal);
+string? sessionName = Environment.GetEnvironmentVariable("DOTNET_INSPECT_ISOLATED");
+if (string.IsNullOrWhiteSpace(sessionName))
+    sessionName = null;
+bool isolated = sessionName != null;
+string? cacheBasePath = Environment.GetEnvironmentVariable("DOTNET_INSPECT_CACHE_DIR");
+if (isolated && cacheBasePath == null)
+    cacheBasePath = Path.Combine(Path.GetTempPath(), $"dotnet-inspect-{sessionName}");
+
+HttpClientFactory.Initialize(offline);
+NuGetCache.Initialize("dotnet-inspect", cacheBasePath, skipNuGetCache: isolated);
 
 // Counted where a package reaches the outcome its mode expects, not where it fails
 // to. An incident counter has to be remembered at every failure site and silently
--- a/eng/prepare-decompiler-package-sweep.cs
+++ b/eng/prepare-decompiler-package-sweep.cs
@@ -587,7 +587,8 @@ foreach (var entry in selected)
         // reported success -- the arbitrary-file overwrite the metadata writes were
         // hardened against, still open on the ninety-one writes per sweep that carry the
         // pool itself.
-        string? uncopied = await CopyOntoOrReport(source, destination);
+        var copy = await CopyOntoOrReport(source, destination);
+        string? uncopied = copy.Error;
         if (uncopied is not null)
         {
             // Recorded as this package failing rather than thrown away. A copy that did
@@ -595,7 +596,7 @@ foreach (var entry in selected)
             // package must be short in the accounting too.
             results.Add(Failed(
                 entry, "copy-failed", uncopied, resolvedPackage, package.Version,
-                selection.Tfm, package.FromCache));
+                selection.Tfm, package.FromCache) with { WriteTemporary = copy.TemporaryFate });
             Console.Error.WriteLine($"rank {entry.Rank}: {entry.Package}: {uncopied}");
             continue;
         }
@@ -707,9 +708,9 @@ foreach (var entry in selected)
 
 assemblies.Sort(StringComparer.Ordinal);
 string assembliesPath = Path.Combine(outputDirectory, "assemblies.txt");
-string? unwritable = await ReplaceTextOrReport(
+string? unwritable = (await ReplaceTextOrReport(
     assembliesPath,
-    string.Concat(assemblies.Select(assembly => assembly + Environment.NewLine)));
+    string.Concat(assemblies.Select(assembly => assembly + Environment.NewLine)))).Error;
 if (unwritable is not null)
 {
     Console.Error.WriteLine(unwritable);
@@ -726,9 +727,9 @@ var manifest = new PackageSweepManifest(
     SelectedPackageCount: assemblies.Count,
     Packages: results);
 string manifestPath = Path.Combine(outputDirectory, "manifest.json");
-unwritable = await ReplaceTextOrReport(
+unwritable = (await ReplaceTextOrReport(
     manifestPath,
-    JsonSerializer.Serialize(manifest, jsonContext.PackageSweepManifest) + Environment.NewLine);
+    JsonSerializer.Serialize(manifest, jsonContext.PackageSweepManifest) + Environment.NewLine)).Error;
 if (unwritable is not null)
 {
     Console.Error.WriteLine(unwritable);
@@ -780,7 +781,7 @@ if (refreshPin)
     // Rank lives in nuget-top-packages.json and is deliberately not repeated here.
     // Two files stating the same rank is two things to keep in step, and the one that
     // drifts is the one nothing reads.
-    unwritable = await ReplaceTextOrReport(
+    unwritable = (await ReplaceTextOrReport(
         pinPath,
         JsonSerializer.Serialize(
             // No timestamp: the file is a pure function of the pins, so re-recording an
@@ -789,7 +790,7 @@ if (refreshPin)
             // #3349 found that hashing a file with a timestamp in it yields an identity
             // that never repeats.
             new PackagePinFile(SchemaVersion: 1, Packages: recorded),
-            jsonContext.PackagePinFile) + Environment.NewLine);
+            jsonContext.PackagePinFile) + Environment.NewLine)).Error;
     if (unwritable is not null)
     {
         // A refresh that cannot write the pin has not refreshed anything, and the file
@@ -1150,7 +1151,7 @@ static bool IsBarePackageId(string? id) =>
 /// appears, which is the hang the read path already had to grow a deadline for. A move
 /// onto the name replaces whatever is there instead of writing through it.</para>
 /// </summary>
-static async Task<string?> ReplaceTextOrReport(string path, string content) =>
+static async Task<WriteOutcome> ReplaceTextOrReport(string path, string content) =>
     await ReplaceOrReport(
         path, stream => stream.WriteAsync(Encoding.UTF8.GetBytes(content)).AsTask());
 
@@ -1158,7 +1159,7 @@ static async Task<string?> ReplaceTextOrReport(string path, string content) =>
 /// Copies <paramref name="source"/> onto <paramref name="destination"/> without opening
 /// the destination, or says why it could not.
 /// </summary>
-static async Task<string?> CopyOntoOrReport(string source, string destination) =>
+static async Task<WriteOutcome> CopyOntoOrReport(string source, string destination) =>
     await ReplaceOrReport(destination, async stream =>
     {
         await using var input = new FileStream(
@@ -1171,7 +1172,7 @@ static async Task<string?> CopyOntoOrReport(string source, string destination) =
 /// hold, then moved onto the name. Every write goes through here, because the property
 /// is not one a second implementation keeps -- File.Copy kept none of it.
 /// </summary>
-static async Task<string?> ReplaceOrReport(string path, Func<FileStream, Task> write)
+static async Task<WriteOutcome> ReplaceOrReport(string path, Func<FileStream, Task> write)
 {
     // Random and created exclusively, not a name another process can predict. A
     // temporary named after the pid is a name anything with write access to the
@@ -1190,7 +1191,7 @@ static async Task<string?> ReplaceOrReport(string path, Func<FileStream, Task> w
         }
 
         File.Move(temporary, path, overwrite: true);
-        return null;
+        return new WriteOutcome(null, "moved");
     }
     catch (Exception ex) when (ex is IOException or UnauthorizedAccessException)
     {
@@ -1199,16 +1200,21 @@ static async Task<string?> ReplaceOrReport(string path, Func<FileStream, Task> w
         // though -- something already sitting at that name is not this sweep's to
         // remove, and deleting whatever is found there would be a worse answer than
         // the failure being reported.
+        string fate = "none";
         try
         {
             if (created)
+            {
                 File.Delete(temporary);
+                fate = "removed";
+            }
         }
         catch (Exception cleanup) when (cleanup is IOException or UnauthorizedAccessException)
         {
+            fate = "left-behind";
         }
 
-        return $"Could not write '{path}': {ex.Message}";
+        return new WriteOutcome($"Could not write '{path}': {ex.Message}", fate);
     }
 }
 
@@ -1287,7 +1293,26 @@ sealed record SweepPackageResult(
     bool? FromCache,
     string? CleanupStatus = null,
     string? CleanupDetail = null,
-    string? Sha256 = null);
+    string? Sha256 = null,
+    string? WriteTemporary = null);
+
+/// <summary>
+/// What a write did, and what became of the temporary it writes through.
+///
+/// <para><c>Error</c> is the message, or null when the write landed.
+/// <c>TemporaryFate</c> is <c>moved</c> when it became the file, <c>removed</c> when the
+/// write failed after it existed and it was cleaned up, <c>left-behind</c> when that
+/// cleanup itself failed, and <c>none</c> when the failure happened before it existed.
+/// </para>
+///
+/// <para>Reported rather than inferred because the difference is not visible from
+/// outside. A failure before the temporary exists and a failure after it is cleaned up
+/// leave the same directory behind and the same message, so nothing downstream can tell
+/// a cleanup that ran from one that was never needed -- including the operator reading
+/// the manifest to find out whether a failed sweep left anything in the pool, which is
+/// the question <c>left-behind</c> exists to answer out loud.</para>
+/// </summary>
+readonly record struct WriteOutcome(string? Error, string TemporaryFate);
 
 [JsonSerializable(typeof(List<PackageListEntry>))]
 [JsonSerializable(typeof(PackagePinFile))]
--- a/eng/prepare-decompiler-package-sweep.cs
+++ b/eng/prepare-decompiler-package-sweep.cs
@@ -444,6 +444,13 @@ foreach (var entry in selected)
             // A backstop behind the version-shape check above: the extractor validates
             // its own path components and throws, and a throw here exits 134 --
             // indistinguishable from a crash, over an input the pin file supplied.
+            //
+            // Ungated, deliberately: IsBareVersion and IsBarePackageId already refuse
+            // every spelling that reaches this, so no input the sweep will accept can
+            // make the extractor throw, and no black-box case can arrange one. It stays
+            // because the two shape checks and this catch guard the same 134 from
+            // opposite sides, and the cheap side to be wrong on is this one. Measured:
+            // disabling this catch leaves every case in EvilPoolSweepGateTests green.
             results.Add(Failed(entry, "acquisition-failed", ex.Message));
             Console.Error.WriteLine(
                 $"rank {entry.Rank}: {entry.Package}: acquisition failed: {ex.Message}");
@@ -465,6 +472,13 @@ foreach (var entry in selected)
         // that matters is made against what came back: a pool entry acquired under a
         // different identity than the one selected is not the package the list ranked,
         // whatever spelling got it there.
+        //
+        // Ungated, and not gateable by EvilPoolSweepGateTests: those cases run offline
+        // against a seeded cache, and the cache-hit path echoes the requested name back
+        // as PackageName (PackageExtractor.AcquireResolvedPackageAsync), so the two sides
+        // of this comparison are the same string by construction. It has teeth only on
+        // the download path, where the name comes from what was served. Measured:
+        // deleting this check outright leaves every case in that class green.
         if (!string.Equals(resolvedPackage, entry.Package, StringComparison.OrdinalIgnoreCase))
         {
             results.Add(Failed(
@@ -722,6 +736,50 @@ if (unwritable is not null)
     Console.Error.WriteLine(unwritable);
     Environment.ExitCode = 2;
     return;
+}
+
+// The pool is made to hold what this run recorded, rather than merely being described
+// by it. Writing assemblies.txt from the in-memory list keeps a stray file out of the
+// record; nothing kept one out of the pool, and the pool is reused by design -- the
+// corpus script passes the same output directory every time so that the paths in an
+// earlier assemblies.txt stay valid. So a package that was pooled once and is refused,
+// re-versioned, or dropped from the list later left its assembly behind, unrecorded and
+// indistinguishable from a current one, for as long as the directory survived. Measured
+// before this existed: a second run whose subject could not be acquired recorded the
+// lead alone and left the first run's assembly sitting in packages/.
+//
+// Held against the record rather than by clearing the directory first. Clearing is the
+// obvious spelling and the wrong one: it also removes whatever a caller had put at a
+// destination, which is the seam two of this sweep's gates use to make a copy fail, and
+// a fix that silently disarms the tests for the code it touches is worse than the leak.
+// Reconciling afterwards touches only what the run did not record.
+//
+// Scoped to packages/, which is the sweep's own: it names every child and records what
+// it put there. Files elsewhere under the output directory belong to the caller.
+// Deletions are announced -- a leftover the sweep removes is still a fault somewhere,
+// and a silent cleanup is how one goes unnoticed for a release.
+var recordedAssemblies = new HashSet<string>(
+    assemblies.Select(Path.GetFullPath), StringComparer.Ordinal);
+try
+{
+    foreach (string pooled in Directory.GetFiles(packageDirectory, "*", SearchOption.AllDirectories))
+    {
+        if (recordedAssemblies.Contains(Path.GetFullPath(pooled)))
+            continue;
+
+        File.Delete(pooled);
+        Console.Error.WriteLine(
+            $"removed '{pooled}' from the pool: this sweep did not record it.");
+    }
+}
+catch (Exception ex) when (ex is IOException or UnauthorizedAccessException)
+{
+    // Reported, and the run fails: the pool now holds something the record does not
+    // name, which is the state this whole step exists to prevent. Exiting 0 here would
+    // hand a consumer a pool the manifest misdescribes.
+    Console.Error.WriteLine(
+        $"Could not reconcile the pool under '{packageDirectory}' with the record: {ex.Message}");
+    Environment.ExitCode = 1;
 }
 
 var manifest = new PackageSweepManifest(
--- a/eng/prepare-decompiler-package-sweep.cs
+++ b/eng/prepare-decompiler-package-sweep.cs
@@ -609,6 +609,12 @@ foreach (var entry in selected)
         // and then copying it leaves an interval a cache writer can land in, and the
         // file in the pool is the one that gets measured -- so that is the one whose
         // bytes must match.
+        //
+        // That last step is reasoning, not a gated property, and is marked so rather
+        // than left to read as covered: every case in EvilPoolSweepGateTests tampers
+        // with the cache entry, so source and destination carry the same bytes and no
+        // black-box case can tell which was measured. Hashing the source here leaves
+        // the whole class green. What is gated is that the hash is compared at all.
         string assemblySha = Convert.ToHexStringLower(
             SHA256.HashData(File.ReadAllBytes(destination)));
         if (honorPin && pin is not null
--- a/eng/prepare-decompiler-package-sweep.cs
+++ b/eng/prepare-decompiler-package-sweep.cs
@@ -755,11 +755,23 @@ if (unwritable is not null)
 // Reconciling afterwards touches only what the run did not record.
 //
 // Scoped to packages/, which is the sweep's own: it names every child and records what
-// it put there. Files elsewhere under the output directory belong to the caller.
-// Deletions are announced -- a leftover the sweep removes is still a fault somewhere,
-// and a silent cleanup is how one goes unnoticed for a release.
+// it put there. Files elsewhere under the output directory belong to the caller -- the
+// deep-inspect workflow writes its acquisition log into the output directory, beside
+// packages/ rather than inside it, and it is not this step's to remove.
+//
+// Every removal is recorded in the manifest, not merely printed. This step deletes
+// unrecorded files under packages/, and a write temporary this run leaked is an
+// unrecorded file under packages/ -- so a reconciliation that only cleaned up would
+// quietly erase the evidence of a leak, and the disk check in EvilPoolSweepGateTests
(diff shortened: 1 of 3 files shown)


richlander:
## Round one reconciled — head `6a6daf15b`

Two reviewers at `b717f3f55`: **GPT-5.6** and **Gemini 3.1 Pro**. Four findings, three distinct. All actioned; one workaround declined with reasons.

### 1. A dropped platform case reported PASS (GPT, Medium; Gemini, Medium — found independently)

Accepted as stated, and it is the worse of the two forms it looks like. `TryPlantSymbolicLink` and `TryMakeUnwritable` returned `false` when they could not arrange the scenario, and the case then `return`ed — which xUnit records as a **pass**. So on a platform where the arrangement failed, the sweep's worst historical defect would have read as *gated* by a case that ran nothing. That is this PR's own subject turned on the PR: a green result covering an empty one.

Fixed in `7d671c5d7`. Only Windows may duck these, as an explicit `Assert.Skip`; everywhere else a failure to arrange throws, because on those platforms it always works and the reason it stopped is something the run should say out loud.

The read-only case needed a second thing neither reviewer raised: it now probes that the mode bits actually **bite** rather than trusting the `chmod`. Running as root ignores them, and a case that believed the `chmod` would watch the copy succeed and then assert that a failing copy was counted as failed — passing on the opposite of its premise.

Demonstrated: with the destination made unplantable, the case now fails naming the denied path, where before it passed.

### 2. The class claimed to open no socket (GPT, Medium) — mechanism taken, remedy not

The mechanism is right and I had it wrong. Each case launches the sweep with `dotnet run`, which restores and builds a file-based app first, and that reads the ordinary NuGet cache and may go to the network to fill it. "Opens no socket" was not mine to claim.

I did not take the suggested fix — prebuilding the sweep and running it with `--no-restore`. That re-architects the harness to make a sentence true. The sentence was the defect, so the sentence is what changed: the claim is now the narrower one the gate actually supports, that **no package the sweep pools can come from anywhere but the seeded fixture**. That is the property #3560 exists for, it is unaffected by the build step, and it is the one every case runs. Corrected in `7d671c5d7` and in the PR body.

### 3. The harness reimplemented the product's cache layout (Gemini, High → I read it as loud, not silent; fixed anyway)

Correct, and sharper than the report: the offending line sat **three lines below a doc comment saying that layout belongs to the product** and that a harness laying it out itself "would keep passing after the product's own layout moved."

It would not keep passing. Seeding goes through `CommitPackage`, so the package follows the product wherever it goes, while the case that overwrites the cached assembly wrote to the abandoned path. So the failure direction is a **false RED**, not a false green — the sweep reads an untampered package, exits 0, and the case demanding exit 1 fails as though the sweep had stopped checking hashes. Loud, but blaming the wrong component, which is its own cost.

The copy also dropped the lowercasing `GetPackageCachePath` applies to name and version, and agreed only because the fixture's constants happen to be lowercase.

Fixed in `6a6daf15b`: recorded at seed time from `NuGetCache.GetPackageCachePath(...)` plus the layout of the package this test authored, with the seed asserting the fixture is really there.

Demonstrated by moving the product's layout for real — `PackageContentCategory` from `package-content-v2` to `package-content-v3`:

| Harness | Product layout | Result |
| --- | --- | --- |
| fixed (`6a6daf15b`) | moved to v3 | **7/0 green** — follows the product |
| pre-fix (`b717f3f55`) | moved to v3 | **`ASweepRefusesBytesThePinDoesNotName` FAILS** |

Both files restored by copy afterwards; `git status` clean.

### 4. Test-host global state (Gemini, Medium) — declined, documented instead

`SeedCache` calls `NuGetCache.Initialize`, which sets process-global cache state with no reset, and `Dispose` then removes the directory it points at. The mechanism is real. I am not adding a workaround, for two reasons.

There is no consumer: **no other type in this test assembly references `CoreCache` or `NuGetCache`** (grep over `src/ILInspector.Decompiler.Tests` and `src/ILInspector.Decompiler` returns only this file). And of the two available end states, the current one is the safer: an override left pointing at a removed directory fails **loudly** if a future cache user appears, whereas the obvious tidy-up — keeping a valid scratch cache alive for the process — would quietly serve that user this fixture instead. The suggested alternative, seeding through a subprocess, buys the same guarantee for a build of another program in the inner loop of a 10-second suite.

Recorded as a comment at the call site saying exactly that, so the next person meets the constraint rather than discovering it.

### Confirmed by both reviewers

Gemini independently re-ran the vacuity matrix and got the same result I did: tampering with the sweep turns exactly the corresponding case red — cleanup deletion, `ReplaceOrReport` writing directly instead of renaming, the TFM check, the SHA-256 check. It also confirmed the isolation is real (`skipNuGetCache: true` bypasses `~/.nuget/packages`, `DOTNET_INSPECT_CACHE_DIR` overrides the app cache base) and that the env-var change is inert when the variables are unset.

### Validation at `6a6daf15b`

- Build 0 errors; `EvilPoolSweepGateTests` **7/0, 0 skipped** on Linux
- `--gate fast` **4408/0**
- Shared caches untouched — `newtonsoft.json/13.0.4` still `a28c251d…`; no `sweep.fixture` anywhere outside scratch
- Tree clean

Round two is running at this head, since findings were actioned after the reviewed one.

richlander:
## Round two, and a gap it led me to

**GPT-5.6** at `b717f3f55`→`6a6daf15b`: no defects. Its evidence was better than its own summary suggested — under `strace`, the sweep made **441 file syscalls against the scratch fixture and 0 against either shared cache**, and every network connection belonged to the outer `dotnet run` rather than to the sweep. That is exactly the narrowed claim I substituted for the retired "opens no socket", independently confirmed. It also corrected a figure from my review brief (`--gate pre-merge` is 18 cases, not 16); that number appears in no shipped text, so nothing needed changing. Its review mode turned out to be read-only, so nine of its rows read "mutation not run" — I ran the two delta-specific ones myself rather than take a clean read on trust:

| mutation | result |
| --- | --- |
| force `File.CreateSymbolicLink` to fail | `ASweepReplacesWhatIsPlanted...` **FAILS** (`FileNotFoundException`) — pre-fix, this passed green |
| move `PackageContentCategory` v2→v3 | **8/0 green** — the harness follows the product |

**Gemini** was cleared before it reported. Its worktree showed it had been drafting a case asking the sweep for a *real* package — and that instinct found something neither review round stated.

### The seven cases did not gate their own isolation

They all pool a synthetic package only the scratch cache holds, so they pass regardless of what the isolation does. Measured, not reasoned — deleting both knobs from the sweep:

```
HttpClientFactory.Initialize(false);
NuGetCache.Initialize("dotnet-inspect", cacheBasePath, skipNuGetCache: false);
```

→ **all seven still green.** The suite would silently regain the developer's NuGet cache and the network, and `Run`'s own doc comment would go on asserting it had not.

That is this PR's thesis pointed back at the PR: a property asserted in prose, believed, and gated by nothing — the same shape as the hand-run probes #3560 exists to retire. My PR body was making the claim too ("pointed at an empty cache, the sweep reports `not available offline`"), from a probe I ran by hand.

`166a718a5` adds the eighth case. It names `newtonsoft.json@13.0.4` — present in the NuGet cache of any machine that has built this repository, and on nuget.org, so reachable by either knob's absence and by neither's presence — and asserts the sweep cannot reach it. It asserts the **reason**, not the exit code: a sweep that found the package would still exit 1 on the pin naming bytes no real package has, so an exit-code-only case would pass just as well with the isolation gone.

Each knob dropped alone turns exactly that case red:

| mutation | red |
| --- | --- |
| `HttpClientFactory.Initialize(false)` | `ASweepCannotReachAPackageTheSeededCacheDoesNotHold` only |
| `skipNuGetCache: false` | `ASweepCannotReachAPackageTheSeededCacheDoesNotHold` only |

`Run`'s doc comment now names that case as its gate, per the repository's rule that an asserted property names what enforces it.

### Validation at `166a718a5`

- `EvilPoolSweepGateTests` **8/0**, 0 skipped; `--gate fast` **4409/0**; build 0 errors
- CI green at the previous head `6a6daf15b` (`test` 9m42s, `decompiler-gates` 12m10s)
- Shared caches untouched — `newtonsoft.json/13.0.4` still `a28c251d…`, no `sweep.fixture` outside scratch; tree clean

Round three is running at this head against both models, with mutation testing as the assignment rather than a read.

richlander:
## Round three reconciled — head `5e97a42ae`

**GPT-5.6** and **Gemini 3.1 Pro**, both with mutation testing as the assignment this time (round two's reviewer was read-only, which is why its clean read did not count for much). They converged independently on the same two HIGH findings. Four things changed.

### 1. The isolation gate's power was half ambient — and silently so (both, HIGH)

The eighth case only bites on the shared-NuGet-cache regression where that cache actually holds `newtonsoft.json`. Somewhere it does not, a sweep with `skipNuGetCache` regressed misses the cache, falls through to a severed network, and fails **exactly as it should have** — green, gating nothing. Both demonstrated it by pointing `NUGET_PACKAGES` at a farm without the package: 8/8 with the regression in place.

Reproduced and fixed. The precondition is read through the product (`NuGetCache.GetNuGetCachePath()`, read-only) and skipped visibly when absent:

```
ASweepCannotReachAPackageTheSeededCacheDoesNotHold [SKIP]
  newtonsoft.json 13.0.4 is not in the shared NuGet cache, so this case cannot
  tell an isolated sweep from one that merely found nothing.
```

A case that looks like it covers a regression and does not is worse than one that says it is not covering it today.

### 2. The cleanup case was vacuous on any early exit (both, HIGH)

It asserted an exit code and an absence, and absence is also what a sweep that never got far enough produces. Gemini made the sweep exit 1 immediately; GPT made the failure happen before the temporary existed — both green with the cleanup deleted.

This is the *same defect this case was added to avoid*, merely moved: it was already the fix for `AssertNoTemporaryLeftBehind` riding on a case that failed before the temporary existed. It now asserts the sweep's recorded status, so the absence means the cleanup ran. Gemini's attack is closed — it now fails with `the sweep exited 1 over a run that wrote no manifest`.

### 3. The gate was coupled to product prose (GPT, Medium)

It asserted the substring `not available offline`, which is `PackageExtractor`'s wording. GPT reworded the message without changing behavior and turned the case red.

Both isolation cases now assert the sweep's own statuses — `acquisition-failed` against `pin-mismatch`, *never reached* against *reached and rejected* — which is the discrimination the case actually needs and does not move when prose does. Verified by re-running GPT's exact attack: reworded message, behavior unchanged, **suite green**.

### 4. The version case never reaches the version comparison (Gemini, confirmed by GPT)

Disabling the sweep's version comparison alone leaves the case green: acquisition has already refused, because the pinned version binds the *request*. The comparison is a backstop against an extractor answering with something other than what it was asked for, and nothing here can arrange for that.

The case now pins the enforcement point it does cover (`acquisition-failed` — never acquired, rather than acquired and then caught), and says the backstop is deliberately ungated rather than leaving it looking gated.

### What is still not gated, and why

GPT's narrower attack — a failure raised *inside* `ReplaceOrReport` before the temporary is created — reports the same `copy-failed`, and the case would pass with the cleanup deleted. I could not close it and did not pretend to.

Closing it requires the temporary to be observable, and it is deliberately not: the name is randomized precisely so nothing can predict or occupy it, which is the property the round-sixteen fix was made to gain, and it exists for the length of one write. So the limit is recorded in the case: it carries that the current implementation's post-creation failure leaves nothing behind, and a refactor moving the failure earlier would weaken it silently. Unverified and marked, rather than unmarked.

### Both reviewers also confirmed

Every other mutation caught exactly the right case — SHA-256, TFM, `File.Copy` restored in place of the rename, failed copy counted as accounted-for, cleanup deleted. Harness-boundary compliance holds: moving `PackageContentCategory` v2→v3 leaves the suite green. The env-var change is inert when unset, matches `Program.cs`, and behaves identically on partial combinations. Both finished with a clean `git status`.

### Validation at `5e97a42ae`

- `EvilPoolSweepGateTests` **8/0**, 0 skipped; `--gate fast` **4409/0**; build 0 errors
- Reviewer attacks re-run against the fixes: prose reword now green, missing-package farm now a visible skip, early exit now red
- Shared caches untouched — `newtonsoft.json/13.0.4` still `a28c251d…`; tree clean

richlander:
## Round four reconciliation — reviews at `5e97a42ae`, head now `6b4f93c99`

Reviewers: **GPT-5.6-Sol** and **Claude Opus 4.8** (rotating the second family
off Gemini). Both were invited specifically to break the claim I had made in the
cleanup case — that its remaining gap could not be closed from outside.

**They broke it, independently, the same way.** That is the round's headline: the
finding I was most confident about was the one that was wrong.

### 1. The temporary's creation is observable after all (Opus LOW, GPT Medium — both demonstrated)

I wrote that the cleanup case could not prove the temporary was ever created,
because its name is randomized and it lives for the length of one write. Both
reviewers pointed out that a `FileSystemWatcher` needs no filename, and GPT went
further and implemented it: armed on the destination's directory, it passes on
current behaviour and fails a sweep whose failure is raised *before* the file
exists.

The gap was real — reaching the copy is not the same as having created the
temporary, and a failure raised earlier reports the same `copy-failed`. The case
passed with the cleanup deleted.

Now watched, on Linux only. inotify is dependable for a file created in a watched
directory; the other backends are laxer, and a missed event would fail a run that
did nothing wrong. That is the same bargain `PlantSymbolicLink` already makes,
and CI runs Linux, so the strong form is the one that gates. The doc comment says
which form is running rather than assuming the strong one.

Replayed: a failure raised inside `ReplaceOrReport` before the `FileStream` →
**RED**, `a run that never created the temporary this case is about`.

### 2. A status is a value, not a subject (GPT Medium)

`ReportedStatus` read the manifest's one entry for its `Status` and never asked
whom the entry was about. GPT rewrote every recorded package name to
`wrong.package` and **all eight cases stayed green** — three cases assert on that
status and none of them could tell they were reading about something else.

Fixed by asserting the row names the package the world asked for, which the world
now tracks across `PinInstead`. Replayed: the same mutation is **RED in 3**.

### 3. The isolation case skipped away coverage it still had (GPT Medium)

Round three made that case skip where the shared NuGet cache does not hold its
probe, because the cache half of its power is ambient. That was right about the
cache half and wrong to take the network half with it: a sweep that regressed
`Initialize(false)` downloads the probe from nuget.org and reaches `pin-mismatch`
on any machine, cache or no cache. GPT measured it — with the precondition
bypassed on a cacheless setup the case still discriminated
(`Expected: acquisition-failed; Actual: pin-mismatch`).

The assertion is unconditional now. The ambient half is reported by its own
precondition case, so losing it is still a visible skip, and losing it no longer
costs the half that does not depend on the machine.

### 4. `Directory.Exists` is not "the cache can serve this" (GPT Medium)

The same precondition asked whether a directory existed. An **empty**
`newtonsoft.json/13.0.4/` exists and the product's cache lookup rejects it, so the
precondition declared power the case did not have and the isolation case went
green over a regression it could no longer see.

It now makes the same two calls `NuGetCache.TryGetCachedPackage` makes on its
NuGet-cache branch — `Directory.Exists` **and** `NuGetCache.IsCachedPackageValid`
— so this and the sweep cannot disagree about what is cached. Not a copy of the
rule: the product's own predicate.

Replayed against a cache holding an empty version directory: old spelling
**passes** (0 skipped), new spelling **skips** with the reason.

### 5. `NUGET_PACKAGES` breaks the subprocess's own restore (GPT Medium) — documented, not fixed

Aimed at a directory without this repository's package graph, the file-based
app's self-restore fails with NU1102 and the sweep never starts. Every case goes
red naming a package it has nothing to do with.

Not actioned as a defect: that is a broken environment reporting itself, not a
false green, and `Explain` prints the restore error that says so. What it earns is
a note at the place someone would reach for the wrong knob, because the product's
cache lookup *ignores* a non-existent `NUGET_PACKAGES` while the restore honours
it, and those two cannot be made to agree. Isolation is
`DOTNET_INSPECT_ISOLATED`'s job.

### Confirmed by both, no action

Opus re-ran the whole tamper matrix and reported the PR sound apart from (1). GPT's
table records the same: SHA, TFM, version, rename→copy, accounting, cleanup,
offline and isolation mutations each bite; the `package-content-v2`→`v3` move
still leaves 8 green, so the harness is still asking the product where the cache
is; and no shared-cache writes were observed. The version backstop remains
deliberately ungated, for the reason recorded in round three.

### Validation at `6b4f93c99`

- Full solution build, 0 errors.
- `EvilPoolSweepGateTests` **9/0, 0 skipped** (eight cases plus the precondition).
- `--gate fast` **4410/0**.
- Shared caches intact — `newtonsoft.json/13.0.4` still `a28c251d…`.
- Tree clean, pushed.

Every fix above was proved non-vacuous by replaying the reviewer's own attack, not
by re-running the suite.

richlander:
## Correction to the round-four reconciliation: the watcher was the wrong instrument

Above, I reported finding 1 as fixed with a `FileSystemWatcher`. That fix is
withdrawn, and the reason is worth recording because it is the sort of thing a
green suite would not have told me.

After integrating `origin/main` I re-ran the class and the cleanup case went
**red** — on a merge that touched neither the sweep nor `DotnetInspector.Packages`.
Three runs, three failures, each taking the full 30-second wait. The message said
the sweep never created its temporary. The sweep's own stderr, in the same
failure text, said `Could not write '...Sweep.Fixture.dll': Is a directory` —
which is `File.Move` failing, which only happens *after* the temporary exists.

A probe named it:

```text
ANY ERROR: The configured user limit (65536) on the number of inotify watches
has been reached, or the operating system failed to allocate a required resource.
```

This box has enough worktrees, editors and agents on it to exhaust the limit.
`FileSystemWatcher` does not throw for that — it reports through the `Error`
event, which nothing was subscribed to, and then simply never fires. So the case
failed a sweep that had done exactly the right thing, and it would have kept
doing so on any machine under the same pressure. It passed on CI, and it passed
here an hour earlier, which is the whole problem: a gate that is green until the
machine is busy is not a gate.

Both reviewers were right that the gap was real and right that it was reachable.
The instrument was wrong.

### What replaced it

The sweep reports it. `ReplaceOrReport` already knows whether it created a
temporary and whether it removed it, and was throwing that away; it now returns a
`WriteOutcome`, and the copy path records `WriteTemporary` in the manifest beside
the staging-directory `CleanupStatus` the manifest already carried:

- `moved` — it became the file.
- `removed` — the write failed after it existed, and it was cleaned up.
- `left-behind` — that cleanup itself failed.
- `none` — the failure happened before it existed.

The case asserts `removed`. `removed` against `none` is exactly the distinction
the watcher was reaching for, and `left-behind` is the answer an operator wants
when a failed sweep may have dropped a partial assembly in the pool — so this is
a fact worth reporting on its own terms, not instrumentation.

The value is derived from creation *downward*, not from `none` upward: a cleanup
that stopped being called must report `left-behind`, because reporting that no
temporary was ever created is the one answer that makes the leak invisible.

### Evidence

| Mutation | Before | Now |
| --- | --- | --- |
| Failure raised before the temporary exists | green | **red** — `Expected: "removed"  Actual: "none"` |
| `File.Delete(temporary)` never reached | green | **red** — `Expected: "removed"  Actual: "left-behind"` |

Deterministic on every platform, so the case no longer gates on Linux only, and
it takes 4.6s instead of 31s.

### Validation at `f5fd3a4be`

Full solution build 0 errors; gate class **9/0, 0 skipped**; `--gate fast`
**4416/0**; shared caches intact; CI green (`test` and `decompiler-gates` both
pass). Round five goes out against this head.

richlander:
## Round five reconciliation — reviews at `f5fd3a4be`, head now `6c469f544`

Reviewers: **Claude Opus 5** and **Gemini 3.1 Pro**, isolated worktrees, both
mutating. Opus ran 22 build-and-run cycles. Nine findings between them, seven
actioned.

Both models independently found the same defect as their headline, which by the
pattern of the last four rounds is the strongest signal available here.

### 1. The only filesystem assertion in the class was disarmed by a rename (both, High/Medium — both demonstrated)

`AssertNoTemporaryLeftBehind` globbed `*.tmp`: the product's naming convention,
restated in the harness. Renaming the sweep's temporary suffix **alone** left all
nine cases green with that assertion silently doing nothing, and each reviewer
composed it with a second one-line change to leave a partial assembly sitting in
`packages/` with the suite green:

- Gemini: `.tmp` → `.temp`, cleanup commented out.
- Opus: `.tmp` → `.partial`, `File.Delete(temporary)` → `File.Delete(temporary + ".gone")`.

It now compares the files under `packages/` against `assemblies.txt` as a set.
That is the sweep's own claim, written at the point it deletes an assembly that
failed its pin — *"assemblies.txt is written from the in-memory list so a stray
file cannot enter the pool"* — so the harness asserts something the product
already promises instead of something the product happens to be named. Nothing in
the test knows what a temporary is called.

Replayed, both variants: **red**, naming the leftover file.

### 2. The identity check reached three of five readers (Opus, Medium)

Round four made `ReportedStatus` ask which package the manifest row was about.
Two cases indexed `manifested["Packages"]![0]` themselves and bypassed it, so
rewriting every recorded package name to `someone.else` left them green.
`ReportedWriteTemporary` had no check either and was covered only by the accident
that its one caller happened to call `ReportedStatus` first.

There is one door now: `ReportedEntry` establishes the subject and returns the
row, and every reader goes through it.

Replayed with Opus's exact mutation: **5 red**, each naming the substitution —
including both cases he reported green.

### 3. Nothing read the `none` half of the value this PR added (Opus, Medium)

The delta's stated property is that `removed` and `none` separate a cleanup that
ran from one that was never needed. Only `removed` was asserted. Opus stuck the
`created` flag true and all nine cases passed: the sweep reported removing a
temporary it never made, because `File.Delete` swallows a missing file even in a
directory it could not have written to.

`ASweepCountsAPackageWhoseCopyFailedAsFailed` — whose write fails before the
temporary exists — now asserts `none`. Replayed: **red**,
`Expected: "none"  Actual: "removed"`.

### 4–6. Three doc claims corrected rather than defended (Opus F4/F5/F7, Gemini F2)

- **`left-behind` is unreachable from a black-box case** — it needs `File.Delete`
  to throw on a file `CreateNew` just made, and on Linux both need the same
  directory permission. It was written up as the payoff. It is now marked ungated,
  the way the version backstop already was, and the doc says what actually catches
  the leak (the set difference in 1).
- **Deriving the fate from creation downward is not a property this gate
  enforces.** Opus reverted it and every case stayed green — correctly, since the
  two derivations differ only on the `left-behind` path nothing can reach. It is a
  defensive choice for the operator reading the manifest, and now reads as one.
- **`moved` is a value the write returns and the manifest never carries**, and
  `null` — the value on every row that is not `copy-failed` — was undocumented.
  Fixed.
- **The `CreateNew` safety claim in `ReplaceOrReport` had no named gate** (Gemini).
  The destination half is gated by the symlink case; the temporary half is not,
  because the name is drawn inside the function and never leaves it. Marked, with
  a note that the `CreateNew` is load-bearing and no test would notice it becoming
  `Create`.

### Recorded, not actioned

**The three `ReplaceTextOrReport` error checks are ungated** (Opus F6). Neutering
all three lets a failed `assemblies.txt`, `manifest.json`, or pin write exit 0
with no case noticing. He verified the current code is correct — no dropped,
inverted, or misdirected check — and this is a pre-existing gap that predates the
return-type change, on paths outside the two properties #3560 names. `--refresh-pin`
likewise has no test anywhere in the repository. Both are real residual, and both
are follow-up rather than this PR's scope.

### Confirmed by mutation, no action

Opus re-ran the whole matrix. The gate catches the historical `File.Copy`
arbitrary-overwrite defect (sentinel preserved), the TFM wildcard, the accounting
defect, the sha-rejection cleanup, and the cache-isolation knob, each with a
legible message. `package-content-v2` → `v3` still leaves the class green, so the
harness still asks the product where the cache is. The documented claim that the
post-acquisition version comparison is unreachable is **accurate** — he disabled
it and the case stayed green, exactly as the comment says it would.

Both worktrees verified clean; neither reviewer wrote to the shared caches.

### Validation at `6c469f544`

- Full solution build, 0 errors.
- `EvilPoolSweepGateTests` **9/0, 0 skipped**.
- `--gate fast` **4416/0**.
- Shared caches intact — `newtonsoft.json/13.0.4` still `a28c251d…`.
- Tree clean, pushed.

Every fix was proved by replaying the reviewer's own attack.

richlander:
## Round six reconciliation — GPT-5.6 and Claude Opus 5, at `6c469f544`

Both reviewers ran mutations against the real nine-case gate. Between them they
found four green-while-broken mutations and three doc claims that overstated what
is gated. All seven are addressed in `91db51174` and `1455f87df`; three further
items are recorded as non-actions.

### Actioned — GPT-5.6

**The pool assertion covered `packages/`, not the output directory.** A cleanup
that moved its temporary one level up instead of deleting it left the file beside
the pool with all nine cases green. The sweep's output directory has exactly three
kinds of occupant — the manifest, the record, and the assemblies the record lists
— so the assertion now names all three and scans the whole directory. Replayed:
red, naming the stray file.

**A missing record read as an empty one.** Suppressing the `assemblies.txt` write
whenever the pool came out empty made "the sweep wrote no record" indistinguishable
from "the sweep recorded pooling nothing" — all nine green. The record is now
required to exist. Replayed: five cases red.

### Actioned — Claude Opus 5

**`pin-mismatch` was the one status no case asserted** — the status this whole
suite exists to produce. Relabelling both sites `library-unavailable` left all nine
green, because exit code, stderr prose and an empty pool are unchanged. That is not
cosmetic: `--refresh-pin` turns `library-unavailable` into a permanent `no-library`
pin, so a package whose cached bytes were poisoned — the exact tamper
`ASweepRefusesBytesThePinDoesNotName` replays — is pinned out of the pool for good
and the evidence goes with it. Both cases now assert the label. Replayed: two red.

**The manifest's `AssemblyPath` could say anything.** It was the one recorded field
nothing read; replacing it with a literal left all nine green. It is how an operator
finds a pooled file without walking the tree, so the success case now holds it to
the path the file is at. Replayed: one red.

**Three doc claims corrected rather than gated**, per the rule that a named gate
must enforce the property attributed to it:

- `EvilPoolPinTests` said the sweep verifying the hash *against the assembly it
  pooled* was gated. Verifying the hash is; verifying it against the pooled copy
  rather than the cache entry it came from is not, and cannot be black-box — the
  case tampers with the cache, so both files carry the same bytes. Hashing the
  source leaves all nine green. Now marked unverified, with the sweep's TOCTOU
  reasoning left standing as sound-and-ungated.
- "Every read of the manifest goes through here" was false in the commit that
  claimed it: `ReportedManifest` parsed independently. Inert today — its only
  caller reads an aggregate — but the sentence was the delta's central claim.
  `ReportedEntry` is now built on `ReportedManifest`, and each says which reads it
  owns.
- "Nothing it writes lands outside the scratch directory" was false of the
  subprocess: restoring and building a file-based app writes to the shared package
  cache and to the SDK's runfile directories. The narrow claim — nothing the sweep
  *pools or acquires* comes from outside the scratch tree — is the one that matters
  and is what the paragraph now says.

### Verified and not changed

Opus's ten mutations also confirmed the round-five fixes hold, each by a mutation
that went red: breaking cleanup while the manifest still says `removed`; a
stuck-true `created` flag (so `"none"` is load-bearing); regressing either
isolation knob; reintroducing write-through instead of replace; and a decoy
subject on every manifest row, which now reds five cases. Dotfiles, symlinks and
files under symlinked directories were probed and are all returned by the pool
scan, so a renamed or hidden temporary cannot slip past.

### Non-actions

- `CreateNew` for the temporary is still ungated — Opus confirmed all nine stay
  green when it becomes `Create`, exactly as the code comment predicts. Marked, not
  fixed.
- The three `ReplaceTextOrReport` error paths and `--refresh-pin` remain ungated;
  both are pre-existing and outside #3560's scope.
- The version comparison after acquisition remains unreachable from a black-box
  case, as documented.

### Validation at `1455f87df`

Solution build 0 errors. Gate class 9/0/0. `--gate fast` 4416/0. Shared caches
intact.

richlander:
## Round seven reconciliation — GPT-5.6 and Gemini 3.1 Pro, at `1455f87df`

Both ran mutations against the real gate. Between them, seven survived. Six are
fixed in `7dde2eae8` and `134614f51`; one was a misread, and the correction is
below. The gate is now ten cases.

### Actioned — GPT-5.6: the world held one package, so every guarantee was about the first iteration

This is the round's finding, and it subsumes four separate mutations. A real sweep
runs its loop ninety-odd times. This harness gave it one package at rank 1, so
restricting any property to `accountedFor == 0` left all nine cases green:

| Restricted to the first package | Was |
| --- | --- |
| The hash comparison | green |
| The replace-rather-than-write-through copy | green |
| Counting a failed copy as a failure | green |
| The manifest's package identity | green |

The sweep those four describe pools poisoned bytes, writes through a planted
symlink, exits 0 over a failed copy, and refreshes the wrong pin key — for packages
two through a hundred.

The world now seeds a **lead** package at rank 1 and the case's subject at rank 2,
with the run window covering both. Each mutation is now one that turns the
subject's own iteration off: **2, 2, 2 and 8 red**.

The lead earns its place twice over. A refusal has to stop the package it is about
and nothing else, so the cases whose subject is refused no longer assert an empty
pool — they assert the pool is *exactly the lead*, with the lead's own distinct
bytes. And a two-package success makes `SelectedPackageCount` meaningful: nothing
read that aggregate, so a successful pool could report zero selected. **6 red** on
it now.

Also actioned: `ASweepCountsAPackageWhoseCopyFailedAsFailed` asserted "not
selected", which a mislabel as `acquisition-failed` satisfies — sending an operator
to the network for a fault on their own disk. It asserts `copy-failed`. **1 red.**

### Actioned — Gemini 3.1 Pro

**No case had ever pinned a package `no-library`**, so that arm of the pin was
entirely unexercised. It is not a way of ignoring a package: nine of the top
hundred carry it, and a version that starts shipping an assembly starts
contributing to the pool. A tenth case now pins the fixture that way over a cache
that does hold a library.

**`FromCache`** is the manifest's own record of the isolation every other case
rests on, and nothing read it — hardcoding it `false` left every case green. The
success case reads it. **1 red.**

### Correction — the `no-library` finding was reported at the wrong severity

Gemini reported that disabling the `no-library` arm "quietly pools the
newly-arrived assembly", so the pool drifts. Measured, it does not. A `no-library`
pin **carries no hash** — there is no assembly for one to describe, which is the
invariant `EvilPoolPinTests` holds the committed file to — so with the arm
disabled, the arriving assembly is compared against no hash and refused by the hash
check a few lines later. The pool still holds the lead alone.

So the new case gates the refusal, which holds either way, and says so: the arm is
a better-diagnosed refusal in front of a backstop that catches it regardless.
Gating the arm *itself* would need a pin that is both `no-library` and carries a
matching hash — a combination the real file cannot hold — so it is marked ungated
rather than left looking covered. This is the same shape as the version comparison
already documented here.

### Non-actions

- **`--refresh-pin` remains ungated** (Gemini, High). Recorded as out of scope in
  round five and still is: it is the path that *writes* the pin, not one that
  enforces it, and this PR's claim is about enforcement. The failure mode is also
  loud rather than silent — a pin written without a hash is refused by the next
  sweep, which is the arm's documented reason for requiring one.
- **The `when (ex is IOException or UnauthorizedAccessException)` filter in
  `ReplaceOrReport`** (Gemini, Medium). Widening it to `catch (Exception)` is
  green, but no black-box case can inject an `ArgumentException` into that write,
  so this is unreachable rather than uncovered. Pre-existing and unchanged by this
  PR.
- **`Downloads`** (Gemini, Low). It is the input list echoed back, with no consumer
  and nothing to disagree with. `FromCache` and `AssemblyPath` were worth reading;
  this is not.
- `CreateNew` for the temporary remains ungated and marked, confirmed again by both
  reviewers.

### Validation at `134614f51`

Solution build 0 errors. Gate class **10/0/0**. `--gate fast` **4417/0**. Shared
caches intact. Every mutation above replayed against the fix: 2, 2, 2, 8, 1, 6, 1.

richlander:
## Round eight reconciliation — MAI-Code and Claude Opus 4.8, at `134614f51`

Both reviewers attacked the same seam from opposite ends, and both were right. Fixed
in `92a9e265e`; the gate is now eleven cases.

### Actioned — MAI-Code: moving the subject off the front left the front unwatched

Round seven put each case's subject at rank 2 so that mutations disabling the first
iteration would be caught. That made the first iteration *observable* and left it
*unasserted*: gating the hash comparison on `entry.Rank > 1` left all ten cases
green, while a poisoned rank-1 package entered the pool. MAI-Code confirmed it by
running the mutated sweep directly against a tampered first package — exit 0, package
selected.

An eleventh case tampers with the **lead** instead, so both ends of the window the
sweep runs over are subjects. It also asserts something no other case could: that the
run **carried on and pooled rank 2**. In every other case the refused package is the
last one, so a sweep that abandoned the whole list on the first bad hash would pass
all of them.

Replayed: `entry.Rank > 1` reds the new case; `accountedFor == 0` reds the old one.

### Actioned — Claude Opus 4.8: the comment claimed the loop, and the fixture cannot

Opus found the general form of the same thing, and drew the right conclusion. A
mutation keyed on the index — `accountedFor <= 1` — enforces the property for exactly
the ranks this fixture occupies and drops it for every rank beyond, with everything
green. Demonstrated on three of the four flagship properties: the byte pin, the safe
copy, and the failure accounting. **No finite fixture closes this**, because the
threshold simply moves; adding a third package would relocate the finding, not answer
it.

Opus's control run is the part that makes the diagnosis stick: the *unconditional*
unsafe-copy mutation reds 3 cases, so the property is genuinely gated at the ranks the
fixture occupies. It is the threshold that leaks, not the assertion.

So the `SweepWorld` comment is corrected rather than the fixture grown. What is gated
is the loop **body**, at both ends of the window it is run over — and a two-package
window has no interior to hide in. That the body is reached identically on the
eighty-eight iterations no fixture here reaches is **unverified**, and now says so
instead of reading as covered.

### Actioned — two more claims marked rather than gated

- **The destination-hash TOCTOU argument** (Opus, Low). Round six marked this in
  `EvilPoolPinTests` but left the security framing standing, unmarked, at the sweep's
  own comment — which is where a reader meets it. Every case tampers with the cache
  entry, so source and destination carry the same bytes and hashing the source leaves
  the class green. Marked at the comment that makes the claim.
- **`FromCache` is asserted one way only** (Opus, Low). A constant `true` passes it,
  because no case here can produce a row that legitimately reports `false`. The
  round-seven comment overstated what the assertion buys; it now says which direction
  it holds and points at the case that gates the isolation itself.

### Validation at `92a9e265e`

Solution build 0 errors. Gate class **11/0/0**. `--gate fast` **4418/0**. Shared caches
intact.

richlander:
## Round nine — Gemini 3.1 Pro and GPT-5.6, against `92a9e265e`

Six findings, all real. Four are now gated, one was a live product defect and is
fixed, and two are marked ungated because this harness cannot reach them.
Resolution is `8097f948d`; `34f682167` merges `origin/main`.

### The pool accumulated across runs — a live defect (GPT-5.6)

The only finding so far that was a defect in the sweep rather than a hole in its
gates, and GPT proved it with a probe that failed against the then-current code.

`packages/` was created and never emptied. Reuse is how this sweep is normally
run — `eng/prepare-evil-corpus.sh` passes the same `<outdir>/sweep` every time,
deliberately, so the paths in an earlier `assemblies.txt` stay valid — so a
package that was pooled once and later refused, re-versioned, or dropped from the
list left its assembly behind: unrecorded, and indistinguishable from a current
one. I reproduced it independently before acting; the second run recorded the lead
alone and left the first run's `002-sweep.fixture/1.0.0/Sweep.Fixture.dll` sitting
in the pool.

The sweep already said this was the dangerous state, in the comment on its own
record: *"a file in `packages/` that nothing recorded is one a later sweep finds,
hashes, and disagrees with the pin about, reported as the pin failing rather than
as the run that made it."* That was prose about a single run.

**Fixed by reconciling the pool against the record**, at the point the record has
just been written, announcing each removal and failing if it cannot.

Worth recording *why not* the obvious fix. I first wrote GPT's suggested remedy —
clear `packages/` before acquisition — and it took two existing gates down with
it: `ASweepCountsAPackageWhoseCopyFailedAsFailed` and
`ASweepLeavesNoTemporaryWhenAWriteFailsAfterCreatingOne` both plant an obstruction
*inside* `packages/`, and clearing removed it, so both watched the copy succeed and
went red at exit 0. A fix that silently disarms the tests for the code it touches
is worse than the leak. Reconciling afterwards touches only what the run did not
record, so both keep working.

Gated by `ASweepIntoAReusedDirectoryPoolsOnlyWhatItRecorded`. Measured: disabling
the reconciliation reddens exactly that case.

### The `!IsSelected` arm had never run (Gemini 3.1 Pro)

The strongest of the round's coverage findings. Both packages every case used ship
an assembly, so `!selection.IsSelected` was never true and the entire arm beneath
it — `library-unavailable`, and the `no-library` pin it confirms — could be deleted
outright with all eleven cases green.

Not a quiet corner: `--refresh-pin` turns a recorded `library-unavailable` into a
permanent `no-library` pin, so this is the status that decides whether a package is
in the pool at all from then on.

A third synthetic package that ships a nuspec and no library now reaches it, held
by three cases: the refusal, the confirmation of a matching `no-library` pin, and
the detail comparison that stops a wiped cache entry confirming a pin recorded over
something else. Measured: `if (false && !selection.IsSelected)` reddens 3; dropping
the detail comparison reddens exactly the third.

I did **not** take Gemini's stated consequence. It reported the mutation as
crashing the sweep at `selection.Paths[0]`, which it could not have observed —
there was no library-less package to reach it with. The green result was real; the
crash was inferred. The finding stands on the coverage gap alone.

### A planted destination was one shape out of three (GPT-5.6)

The symlink case planted a link whose target exists. GPT showed two ways around a
gate that narrow, both measured green against all eleven:

- a sweep that protects live symlinks and writes through **dangling** ones creates
  an arbitrary external file at exit 0;
- a sweep that asks only whether `LinkTarget` is set — which a **hard link** leaves
  null — writes straight through onto the file it shares an inode with.

The property was never about symlinks; it is that the sweep replaces what it finds
instead of opening it. The case is now a theory over all three shapes. Measured:
each of GPT's two mutations reddens exactly its own shape.

### Two checks are marked ungated rather than gated

Both are real gaps in coverage. Neither is reachable from this harness, so they are
marked at the code rather than papered over with a case that would prove nothing.

**The identity comparison** (Gemini). Its two sides are the same string by
construction on the path these cases take: `PackageExtractor` returns the
*requested* name as `PackageName` on a cache hit
(`AcquireResolvedPackageAsync`), and every case here runs offline against a seeded
cache. I built the fixture to try it — a package committed under one name whose
nuspec declares another — and the sweep pooled it, confirming the check never
fires offline. It has teeth only on the download path. Fixture removed, comment
marked.

**The `ArgumentException` backstop** (Gemini). `IsBareVersion` and
`IsBarePackageId` already refuse every spelling that could reach it, so no input
the sweep accepts can make the extractor throw. It stays — the shape checks and
the catch guard the same exit 134 from opposite sides — and it is now marked as
the unreachable backstop it is. Gemini rated this Medium on the strength of a
`../bad` version reaching the extractor; that version is refused at pin-read time
with exit 2.

### Validation

At `34f682167`, after merging `origin/main`:

- solution build, 0 errors
- `EvilPoolSweepGateTests` **17/0**, `EvilPoolPinTests` **10/0**
- `--gate fast` **4441/0**
- shared NuGet cache intact (`newtonsoft.json/13.0.4` → `a28c251dfe36`)

Every mutation quoted above was executed, not reasoned about.

richlander:
## Round ten — Gemini 3.1 Pro and GPT-5.6

Nine findings across the two reviewers, at `34f682167`. Both reviewers converged
independently on the same defect, and it is one I introduced in round nine.

**First, a correction.** The round-nine comment on this PR said the
reconciliation "does not weaken `AssertNoTemporaryLeftBehind`". That was wrong.
The reconciliation deletes unrecorded files under `packages/`, and a write
temporary a failed copy leaked *is* an unrecorded file under `packages/` — so it
swept away the very evidence that check was looking for. Gemini raised it as
High and GPT as Medium; both are right, and the claim should not have been made.

Worth stating plainly: I rejected GPT's round-nine remedy ("clear `packages/`
first") precisely because it disarmed the tests for the code it touched, and
then shipped a fix with the same class of flaw.

### Fixed

| --- | --- | --- | --- |
| 1 | Gemini (High), GPT (Medium) | Reconciliation erases the evidence of a leaked write temporary | Product records removals; cases assert the pool against the record |
| 2 | GPT (High) | Reconciliation deletes **outside** the pool through a directory symlink, and exits 0 | Walk written out; a link is removed as a link |
| 3 | Gemini (Medium), GPT (Medium) | Reconcile-failure exit code ungated | New case, built so the reconciliation is the only thing that can fail |
| 4 | GPT (Medium) | A walk that stops after its first removal stays green | Reused-directory case plants three leftovers |
| 5 | GPT (Medium) | A walk that considers only `*.dll` stays green | One of the three is not an assembly |

**#1** is fixed by having the product *report* the fact rather than having the
test read it off disk. The sweep records `RemovedFromPool` and `Unreconciled` in
the manifest, and `AssertPoolMatchesRecord` holds the pool to that record, so a
swept leak now shows up as a removal the case did not expect.

Gemini's stated diff would have gone red on its own — `fate` starts as
`"left-behind"`. The mutation that actually stays green keeps `fate = "removed"`
while skipping `File.Delete(temporary)`. Re-derived, replayed, and it reddens
`ASweepLeavesNoTemporaryWhenAWriteFailsAfterCreatingOne` again.

**#2** is the serious one, and I reproduced it before trusting it:
`Directory.GetFiles(dir, "*", SearchOption.AllDirectories)` follows a directory
symlink and hands back paths on the far side. This step deletes what it is
handed, so round nine's recursive deletion could be aimed at a directory the
sweep does not own — and then report success. It now walks explicitly and removes
a link as a link, which is a deletion bounded to the pool, since removing a
symlink never touches its target.

This is not a claim that anything plants links in the pool. It is that a step
which deletes recursively should be unable to delete outside the directory it
owns, whatever is in it.

**#3** needed a second attempt. My first shape made the *subject's* destination
directory unwritable — but that fails the copy, which exits 1 too, so the case
passed while the exit code under test was mutated to 0. The stray now sits in a
directory of its own and every package is acquired, copied, and recorded
normally, so the exit code can only have come from the reconciliation.

### Marked ungated

- The stderr line beside each recorded removal (GPT, Low) — the manifest is the
  gate; the print is operator convenience.
- The ordinal path comparison (GPT, Low) — whether a case-variant assertion means
  anything depends on the filesystem the suite runs on.
- `ExitCode = 2` IO paths generally (Gemini, Low) — pre-existing and outside this
  PR, which is about the two sweep properties in #3560.

### Confirmed safe

Gemini attacked path normalization from five directions — relative paths,
trailing slashes, `%20`, non-ASCII, symlinked output roots — and all normalize
correctly. GPT confirmed the same, and separately confirmed that removing `ln`
from `PATH` makes the hard-link case fail visibly rather than silently pass.

### Every mutation replayed against the fix

| Mutation | Before | After |
| --- | --- | --- |
| Skip `File.Delete(temporary)`, still report `"removed"` | 17 green | reddens 1 |
| Follow directory symlinks (round nine's spelling) | 17 green, bystander deleted | reddens 1 |
| Reconcile failure exits 0 | 17 green | reddens 1 |
| `break` after the first removal | 17 green | reddens 1 |
| Reconcile only `*.dll` | 17 green | reddens 3 |

### Validation

At `d0404984e`, with `origin/main` integrated at `81e1c069a`:

- `EvilPoolSweepGateTests` — 19 passed, 0 failed
- `--gate fast` — 4443 passed, 0 failed
- `EvilPoolPinTests` — 10 passed, 0 failed
- Shared NuGet and dotnet-inspect caches untouched

Round-ten yield: 9 findings, 5 fixed, 3 marked ungated, 1 declined as
out of scope. Two of the five were real product defects, one of which could
delete files outside the output directory.

Issue #3517, opened 2026-07-29
Expose product-owned style choices: selectability, stable choice ids, and conflict semantics are re-derived per consumer (#3486 sibling)


richlander:
## Summary

`StyleOptionCatalog` owns the taste knobs as **axes**: each
`StyleOptionDescriptor` has an `Id`, a `DefaultValue`, and a list of `Values`
with `Token`/`Title`/`OracleEndorsed`. What it does not own is the projection a
picker actually needs — **the set of things a user can select, and the stable id
of each one**. Every consumer has to derive that, and the WASM prototype derives
it twice:

```csharp
// prototypes/inspect-web/engine/Program.cs:799 (ListStyleOptions)
// A knob is an axis of values; the default/off value is not selectable taste.
var choices = descriptor.Values
    .Where(value => !string.Equals(value.Token, descriptor.DefaultValue, StringComparison.Ordinal))
    .ToArray();
var multiValue = choices.Length > 1;
...
    multiValue ? $"{descriptor.Id}:{value.Token}" : descriptor.Id,   // id
    multiValue ? descriptor.Id : null));                             // conflict group
```

```csharp
// prototypes/inspect-web/engine/Program.cs:830 (BuildPrinterOptions)
var choices = descriptor.Values
    .Where(value => !string.Equals(value.Token, descriptor.DefaultValue, StringComparison.Ordinal))
    .ToArray();
var multiValue = choices.Length > 1;
var id = multiValue ? $"{descriptor.Id}:{value.Token}" : descriptor.Id;
```

Three separate rules are invented here, none of them product-owned:

1. **Selectability** — "the default value is not selectable taste", so a knob's
   choices are its non-default values.
2. **Id composition** — a boolean knob keeps the bare `descriptor.Id`; a
   multi-value axis becomes `"{Id}:{Token}"`. The two spellings exist so that
   adding a third value to a boolean knob does not silently change the id of
   its existing choice.
3. **Conflict semantics** — a `ConflictGroup` (the descriptor id) that the
   client uses to single-select within an axis
   (`prototypes/inspect-web/src/app.js:6802`, `toggleTaste`).

The listing and the apply path each re-derive all three, so they agree only by
being kept in sync by hand. Diverge them and the failure is silent: an id the
picker offers that the applier never matches simply prints in the default style.

## Why the id matters beyond one consumer

These invented ids are **persisted**. The prototype stores the selection under
`localStorage["inspect-taste"]` as a raw id array
[...]

Issue #3516, opened 2026-07-29
wasm engine: QueryMemberAnnotatedSource discards styleOptionsJson, so Annotated Source silently ignores taste (#3191 invariant)


richlander:
## Summary

The WASM prototype's annotated-source export accepts the user's taste selection
and throws it away. `QueryMemberAnnotatedSource` opens with a discard:

```csharp
// prototypes/inspect-web/engine/Program.cs:945
    string styleOptionsJson)
{
    _ = styleOptionsJson;
```

Every sibling export threads it — `QueryMemberSource` (:913),
`QueryTypeSource` (:2315), and `QueryTypeMemberSource` (:2372) all pass
`printerOptions: BuildPrinterOptions(styleOptionsJson)`. The client is not at
fault: `loadSelectedMemberAnnotatedSource` sends the current selection like every
other source call (`prototypes/inspect-web/src/app.js:5495`,
`styleOptionsJson: JSON.stringify(state.taste)`).

So the user toggles a knob in the taste popover, Decompiled Source re-renders in
the new style, and Annotated Source re-renders in the shipped default — two
views of the same member disagreeing about how that member is spelled, with no
signal that one of them ignored the request.

## Why this is a defect and not a scoping choice

This is exactly the invariant the CLI states, and #3191 closed:

```csharp
// src/dotnet-inspect/Inspectors/MemberCodeProvider.cs:218
// Annotated Source spells the same member as Decompiled
// Source, so it renders with the same resolved style options
// -- otherwise the two views of one member disagree. The
// other projections here (cost/semantics overlays, fact rows)
// are style-invariant evidence and keep the shipped defaults.
PrinterOptions: request.AnnotatedSource ? renderOptions : null,
```

The distinction that comment draws is the right one and it survives here: the
annotated projection is a *spelling* of the member, so it takes the style; the
cost/semantics overlays and fact rows are style-invariant evidence and keep the
defaults.

## Fix

The product seam already carries this — `ResearchViews.MemberProjectionRequest`
has a `PrinterOptions` parameter (`src/ILInspector.Research/ResearchViews.cs:38`)
and `ProjectMember` forwards it (:108). The prototype's call site
(Program.cs:986) simply omits it. Drop the discard and pass
`PrinterOptions: BuildPrinterOptions(styleOptionsJson)` on the
`MemberProjectionRequest`, matching the CLI.

## Evidence to include

Toggle a byte-divergent knob (e.g. a `Lens` option) in the prototype, then
[...]

PR #3607, opened 2026-07-31
Expose a product-owned registry of style-option tiers (#3486)

### LINKED ISSUE #3486: Expose a product-owned registry of style-option categories (taste tiers) with display labels
## Summary

The decompiler style ("taste") catalog is product-owned and data-driven at the
**option** level — `StyleOptionCatalog.Options` exposes each knob's `Id`,
`Title`, `Summary`, `Tier`, `ByteDivergent`, and per-value `Token`/`Title`/
`OracleEndorsed`. But the **category** level (the `StyleOptionTier` grouping a
consumer uses to lay the catalog out) has no product-owned presentation.
`StyleOptionTier` is a bare enum
(`src/ILInspector.Decompiler/Pipeline/Ir/StyleOptionCatalog.cs:9`): it carries
a rich fidelity contract in XML docs, but no human-facing display **label**,
**description**, or explicit **display order** that a UI can render.

As a result, every consumer that groups the catalog by tier must hardcode a
display string and an ordering per enum value, restating a taxonomy the product
already owns. The WASM prototype UI is the first such consumer and does exactly
this today:

```js
// prototypes/inspect-web/src/app.js
const TASTE_TIERS = [
  ["Formatting", "Formatting"],
  ["Spelling",   "Spelling (this.)"],
  ["Lens",       "Lenses · byte-divergent"],
  ["Synthesis",  "Name synthesis"],
];
```

These four strings — and the order — are the only taste labels not sourced from
the engine. Add a knob to a *new* tier and the prototype silently drops it until
someone edits this array. That is precisely the coupling `StyleOptionCatalog`
was designed to remove at the option level.

## Ask

Expose a product-owned **registry of style-option categories (tiers)** alongside
`StyleOptionCatalog.Options`, so the taxonomy is data-driven end to end. For
each tier, at minimum:

- a **stable id** (the enum token, e.g. `Formatting`) — never localized, safe as
  a grouping key;
- a human **Title** (e.g. "Spelling (this.)", "Lenses · byte-divergent");
- an explicit **display order** (the enum is already ordered "most conservative
  contract → least"; make that order a first-class, queryable fact rather than
  relying on `Enum` ordinal);
- optionally a short **Summary** — the fidelity contract already written in the
  enum's XML docs (e.g. "byte-divergent style lens: behavior-faithful but not
  opcode-faithful") — so a UI can explain the group, not just name it.

This mirrors the existing `StyleOptionDescriptor` contract one level up, so a
consumer can render the full grouped picker (categories + options) with zero
hardcoded presentation. A natural shape is `StyleOptionCatalog.Tiers` returning
descriptors, with `StyleOptionDescriptor.Tier` referencing one by id.

## Consumers that benefit

- The WASM prototype's `ListStyleOptions` engine export
  (`prototypes/inspect-web/engine/Program.cs`) currently emits
  `descriptor.Tier.ToString()` and leaves labeling to the client; it could
  project the tier registry verbatim, retiring `TASTE_TIERS`.
- Any future CLI/`taste`-skill presentation that wants to group the catalog gets
  the same labels and order for free, keeping every surface consistent.

## Notes / constraints

- Keep it SRM-only / NativeAOT-friendly / Roslyn-free like the rest of the
  catalog — static descriptors, no reflection over the enum for display data.
- The tier **id** must stay the stable enum token so persisted taste selections
  and grouping keys are unaffected; only presentation metadata is added.
- Titles/summaries are the product's to word; the labels above are the
  prototype's current guesses and should be treated as a starting point, not a
  spec.

richlander:
Closes #3486 (product half).

## What changes

`StyleOptionCatalog` is data-driven at the **option** level but not at the
**category** level: `StyleOptionTier` is a bare enum, so every consumer that
groups the catalog by tier has to restate the taxonomy with its own labels and
ordering. This adds the missing level.

- **`StyleOptionTierDescriptor`** — the enum token as a stable `Id`, a display
  `Title`, a `Summary` stating the fidelity contract the tier's knobs honor, an
  explicit `Order` (ascending, most conservative contract first), and the
  tier-level `ByteDivergent` flag.
- **`StyleOptionCatalog.Tiers`** — every tier in display order, plus `GetTier`
  to resolve one by id.

A host now renders a grouped picker by walking `Tiers` and filtering `Options`
on `Tier`, holding no label, blurb, or ordering of its own.

`Order` is deliberately **not** the enum ordinal, which is the point of the ask:
presentation order and declaration order can move independently.

## Behavioral claim

**Additive only.** No existing type, member, output, config key, or default
changes. `PrinterOptions`, the `Options` list, the endorsement subsets, and
`ApplyFullTaste` are untouched; the new surface is two additions to a public
static class plus one new record.

## Evidence

Gates follow the declaration-drives-enforcement convention — the registry
*drives* the checks rather than being restated by them, so a stale entry and a
missing entry both fail:

- `Tiers_CoverEveryTierExactlyOnce` — set equality against
  `Enum.GetValues<StyleOptionTier>()`, plus no duplicates.
- `Tiers_AreListedInAscendingDisplayOrder_WithUniquePositions` — the list is
  already sorted by `Order`, so a host never sorts; positions are unique so no
  unstable tie-break can decide layout.
- `EveryOption_GroupsUnderARegisteredTier` — the grouped-picker walk itself: every
  knob lands in exactly one rendered group.
- `ByteDivergence_IsATierProperty` — each option's `ByteDivergent` equals its
  tier's, and the byte-divergent tier set equals the tiers of byte-divergent
  options (non-vacuity, so a tier cannot claim divergence with no knob to
  witness it).
- `GetTier_ResolvesEveryTier_AndRejectsAnUnregisteredOne` — a miss throws rather
  than degrading into an unlabeled group.

**Verified non-vacuous by tampering** (both reverted):

| Tamper | Result |
| --- | --- |
| File the byte-divergent `prefer-long-literal-suffix` under `Formatting` | 2 tests fail (`ByteDivergence_IsATierProperty`, and the pre-existing `ByteDivergent_IsExactlyTheLensTier`) |
| Delete the `Synthesis` descriptor from `Tiers` | 4 tests fail (coverage, grouping, `GetTier`, byte-divergence) |

Runs (Release, per repo policy):

```
dotnet build dotnet-inspect.slnx -c Release                       # 0 errors
dotnet run --project src/ILInspector.Decompiler.Tests -c Release --gate fast
    # 4413 total, 0 failed
dotnet run --project src/ILInspector.Decompiler.Tests -c Release \
  -class StyleOptionCatalogTests -class ByteNeutralityGateTests \
  -class ExpressionTreeLambdaTests                                # 72 / 0 failed
dotnet run --project src/dotnet-inspect.Tests -c Release \
  -class AppliedTasteSectionTests -class RenderStyleConfigTests   # 79 / 0 failed
npx markdownlint-cli docs/decompiler-taste.md                     # clean
```

(The one `--gate fast` failure on the first pass was
`ExpressionTreeLambdaTests.LookalikeExpressionAssembly_StaysFactoryCalls`
reporting a missing fixture binary because only the test project had been built;
it passes after the full-solution build the error message prescribes.)

## Scope boundary / non-action

`#3486` names the WASM prototype's `TASTE_TIERS`
(`prototypes/inspect-web/src/app.js:6746`) as the consumer to retire. That file
is **not on `main`** — it lives on `prototype/wasm-command-ui` — so this PR
delivers the product registry only, and retiring `TASTE_TIERS` plus projecting
`Tiers` from the `ListStyleOptions` engine export is a follow-up on that branch.
Worth noting for whoever picks it up: the prototype's current `"Spelling (this.)"`
label is already stale, since the `Spelling` tier now also holds the
`var-spelling-style` axis — exactly the drift this registry removes.

Also out of scope, tracked separately: #3517 (product-owned *choice* identity —
selectability, stable choice ids, conflict semantics, the level below this one)
and #3516 (the prototype's annotated-source export discarding taste).

## Review

Additive, no behavior change, gated by set-equality tests verified non-vacuous —
the single-review tier. Requesting one MAI-Code review.

### DIFF
--- a/docs/decompiler-taste.md
+++ b/docs/decompiler-taste.md
@@ -751,6 +751,21 @@ delegates that read and drive that value's backing state without reflection. The
 descriptor reads and single-selects the whole axis through
 `GetValue`/`WithValue`.
 
+The tier itself is described by the catalog too, one level above the knobs:
+`StyleOptionCatalog.Tiers` lists a `StyleOptionTierDescriptor` per
+`StyleOptionTier`, carrying the tier as its stable `Id`, a display `Title` and
+`Summary` (the fidelity contract its knobs honor), an explicit `Order`
+(ascending, most conservative contract first), and the tier-level
+`ByteDivergent` flag — with `GetTier` resolving one by id. A host renders a
+grouped picker by walking `Tiers` and filtering `Options` on `Tier`, holding no
+label, blurb, or ordering of its own, so a knob in a newly added tier surfaces
+without a consumer edit
+([#3486](https://github.com/richlander/dotnet-inspect/issues/3486)). The
+registry is exhaustive against the enum and agrees with its knobs' own
+`ByteDivergent` by gate (`StyleOptionCatalogTests`), so neither a missing tier
+nor a misfiled knob is silent. `Order` is deliberately not the enum ordinal, so
+presentation order and declaration order can move independently.
+
 `OracleEndorsed` records **declared**-oracle endorsement specifically — the value
 has a `.editorconfig` rule behind it. `CorpusEndorsed` records **revealed**
 endorsement — the runtime's own source corpus reveals a dominant practice for the
--- a/src/ILInspector.Decompiler.Tests/StyleOptionCatalogTests.cs
+++ b/src/ILInspector.Decompiler.Tests/StyleOptionCatalogTests.cs
@@ -445,4 +445,84 @@ public class StyleOptionCatalogTests
         var full = StyleOptionCatalog.ApplyFullTaste(PrinterOptions.Default);
         Assert.Equal("explicit", varStyle.GetValue(full));
     }
+
+    [Fact]
+    public void Tiers_CoverEveryTierExactlyOnce()
+    {
+        // The registry drives the grouping a host lays the catalog out with, so it
+        // has to be total: a StyleOptionTier value with no descriptor would drop
+        // its knobs out of a grouped picker silently. Set equality fails both a
+        // missing entry and a stale one.
+        var declared = Enum.GetValues<StyleOptionTier>().ToHashSet();
+        var registered = StyleOptionCatalog.Tiers.Select(t => t.Id).ToArray();
+
+        Assert.Equal(declared, registered.ToHashSet());
+        Assert.Equal(registered.Length, registered.Distinct().Count());
+    }
+
+    [Fact]
+    public void Tiers_AreListedInAscendingDisplayOrder_WithUniquePositions()
+    {
+        // Order is the product's presentation fact, so the list is already sorted
+        // by it — a host renders Tiers as-is and never sorts. Positions are unique
+        // so the layout cannot depend on an unstable tie-break.
+        var orders = StyleOptionCatalog.Tiers.Select(t => t.Order).ToArray();
+
+        Assert.Equal(orders.OrderBy(o => o).ToArray(), orders);
+        Assert.Equal(orders.Length, orders.Distinct().Count());
+    }
+
+    [Fact]
+    public void Tiers_HaveNonEmptyTitlesAndSummaries()
+    {
+        Assert.All(StyleOptionCatalog.Tiers, tier =>
+        {
+            Assert.False(string.IsNullOrWhiteSpace(tier.Title));
+            Assert.False(string.IsNullOrWhiteSpace(tier.Summary));
+        });
+    }
+
+    [Fact]
+    public void ByteDivergence_IsATierProperty()
+    {
+        // The tier states the contract its knobs honor, so a host can warn about a
+        // whole group. Per-knob agreement makes that statement enforced rather than
+        // documented: a byte-divergent knob filed under a byte-neutral tier fails
+        // here, as does the reverse.
+        Assert.All(Options, option =>
+            Assert.Equal(StyleOptionCatalog.GetTier(option.Tier).ByteDivergent, option.ByteDivergent));
+
+        // Non-vacuity in the other direction: a tier cannot claim byte-divergence
+        // with no knob to witness it, so the registry's flag stays a statement
+        // about real knobs. ByteDivergent_IsExactlyTheLensTier pins which tier
+        // that is; this pins that the registry agrees with the knobs.
+        Assert.Equal(
+            StyleOptionCatalog.Tiers.Where(t => t.ByteDivergent).Select(t => t.Id).ToHashSet(),
+            Options.Where(o => o.ByteDivergent).Select(o => o.Tier).ToHashSet());
+    }
+
+    [Fact]
+    public void GetTier_ResolvesEveryTier_AndRejectsAnUnregisteredOne()
+    {
+        foreach (var tier in Enum.GetValues<StyleOptionTier>())
+            Assert.Equal(tier, StyleOptionCatalog.GetTier(tier).Id);
+
+        // A miss is a catalog defect, so it throws instead of yielding an unlabeled
+        // group that renders as if it were fine.
+        Assert.Throws<ArgumentOutOfRangeException>(() => StyleOptionCatalog.GetTier((StyleOptionTier)(-1)));
+    }
+
+    [Fact]
+    public void EveryOption_GroupsUnderARegisteredTier()
+    {
+        // The grouped-picker walk a host performs: every knob lands in exactly one
+        // rendered group, so none is invisible to a consumer that trusts Tiers.
+        var grouped = StyleOptionCatalog.Tiers
+            .SelectMany(tier => Options.Where(o => o.Tier == tier.Id))
+            .Select(o => o.Id)
+            .ToArray();
+
+        Assert.Equal(Options.Select(o => o.Id).ToHashSet(StringComparer.Ordinal), grouped.ToHashSet(StringComparer.Ordinal));
+        Assert.Equal(Options.Count, grouped.Length);
+    }
 }
--- a/src/ILInspector.Decompiler/Pipeline/Ir/StyleOptionCatalog.cs
+++ b/src/ILInspector.Decompiler/Pipeline/Ir/StyleOptionCatalog.cs
@@ -38,6 +38,59 @@ public enum StyleOptionTier
 }
 
 /// <summary>
+/// The product-owned presentation of one <see cref="StyleOptionTier"/>: the
+/// category level of the catalog, sitting one level above
+/// <see cref="StyleOptionDescriptor"/>. A host grouping the catalog by tier reads
+/// its label, blurb, and display position from here instead of restating a
+/// taxonomy the product already owns, so a knob in a newly added tier surfaces
+/// without a consumer edit.
+///
+/// <para>The <see cref="Id"/> is the stable grouping key; <see cref="Title"/> and
+/// <see cref="Summary"/> are presentation and may be reworded. Every tier has
+/// exactly one descriptor, which
+/// <c>StyleOptionCatalogTests.Tiers_CoverEveryTierExactlyOnce</c> enforces by set
+/// equality against <see cref="StyleOptionTier"/> — a new enum value with no
+/// descriptor fails there rather than silently vanishing from a picker.</para>
+/// </summary>
+public sealed record StyleOptionTierDescriptor
+{
+    /// <summary>
+    /// The tier this describes. The enum token is the stable, never-localized
+    /// grouping key a host keys presentation and persisted selections off.
+    /// </summary>
+    public required StyleOptionTier Id { get; init; }
+
+    /// <summary>Short human-facing label for a group heading in a picker.</summary>
+    public required string Title { get; init; }
+
+    /// <summary>
+    /// One-sentence statement of the fidelity contract every knob in this tier
+    /// honors, so a host can explain a group and not merely name it.
+    /// </summary>
+    public required string Summary { get; init; }
+
+    /// <summary>
+    /// Explicit display position, ascending, most conservative contract first.
+    /// First-class rather than an <see cref="StyleOptionTier"/> ordinal so the
+    /// enum's declaration order and the presentation order can move
+    /// independently.
+    /// </summary>
+    public required int Order { get; init; }
+
+    /// <summary>
+    /// <see langword="true"/> when every knob in this tier is byte-divergent —
+    /// the tier-level statement of the contract
+    /// <see cref="StyleOptionDescriptor.ByteDivergent"/> carries per knob. A host
+    /// can warn about a whole group without inspecting its members.
+    /// <c>StyleOptionCatalogTests.ByteDivergence_IsATierProperty</c> enforces the
+    /// agreement in both directions, so
+    /// <see cref="StyleOptionTier.Lens"/> being the only byte-divergent tier is a
+    /// gated claim rather than a comment.
+    /// </summary>
+    public required bool ByteDivergent { get; init; }
+}
+
+/// <summary>
 /// One selectable value on a <see cref="StyleOptionDescriptor"/>'s axis: its
 /// stable token, optional human label, oracle endorsement, and the
 /// <c>.dotnet-inspectconfig</c> key (if any) that selects it. A boolean knob has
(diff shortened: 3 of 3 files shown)


richlander:
## Review reconciliation

**MAI-Code**, reviewed at the exact head `1f8fb61dfc97217d34d86bba8115c774a131d059`
in an isolated detached review worktree. **Verdict: CLEAN — no blocking findings.**

Per the review tiers in `AGENTS.md`, this is the single-review tier: the change is
more than trivial but not high risk — purely additive, no behavior/output/config
change, and its correctness surface is a set-equality gate rather than subtle
logic. The worktree was verified clean (`git status --short` empty) and has been
removed.

Attack points the reviewer was given and what came back:

| # | Attack point | Result |
| --- | --- | --- |
| 1 | Is it genuinely additive? Packaging/API-stability consequence of a new public record? | Additive confirmed; `ILInspector.Decompiler` verified still non-packable (no `IsPackable=true`), so no packaging surprise |
| 2 | Are the new gates non-vacuous? Find a tamper that *should* fail but doesn't | Independent tampers (flip a byte-neutral tier's `ByteDivergent`; add a duplicate tier descriptor) failed the expected gates; no missed tamper found |
| 3 | Should `GetTier` throw or return null? Linear scan acceptable? | No product call sites outside tests; throwing is right for an exhaustive-by-gate registry |
| 4 | `Order` diverges from enum declaration order (Lens/Synthesis swapped) — hazard? | Searched for anything sorting or comparing `StyleOptionTier`; no consumer depends on the ordinal, so the divergence is isolated and intended |
| 5 | Doc accuracy — do the XML-doc gate references name real tests? | Names match the tests as written |
| 6 | Overlap with pre-existing `ByteDivergent_IsExactlyTheLensTier` | Complementary, not contradictory — noted, not blocking |

Reviewer's independent real runs (Release):

- `dotnet build dotnet-inspect.slnx -c Release` — succeeded
- `StyleOptionCatalogTests` — 35 tests, 0 failures
- `AppliedTasteSectionTests` + `RenderStyleConfigTests` — 79 tests, 0 failures

Nothing was dismissed and no resolution commits were needed — the review produced
no findings, so the reviewed head is unchanged.

Non-blocking follow-ups, tracked separately and deliberately out of scope here:
retiring the prototype's `TASTE_TIERS` on `prototype/wasm-command-ui` (the
consumer half of #3486, whose file is not on `main`), #3517 (product-owned choice
identity), and #3516 (prototype annotated-source export discards taste).

richlander:
Ready to merge

All CI green at `1f8fb61` (test, decompiler-gates, build-net10, csharp-diff-smoke, markdownlint; il-diff-smoke and pack correctly path-skipped), and the required MAI-Code review is clean at that same head. No outstanding findings.

The follow-ups named above (prototype `TASTE_TIERS` retirement, #3517, #3516) are non-blocking.

richlander:
## Consumability from the browser (Wasm) host — the shape of the follow-up

This registry is exactly what the browser prototype needs: today it restates the
tier taxonomy by hand (`prototypes/inspect-web/src/app.js`, `TASTE_TIERS` — tier
labels **and** display order hardcoded in JS), which is the bespoke code this PR
exists to retire. Recording the shape of the host-half so it can land cleanly
once this merges — **no changes until then.**

### The gap

This PR (the *product half* of #3486) stops at the C# public surface. The
browser is across the Wasm boundary, and the engine bridge only projects
*options*, not the new *tiers*:

- `prototypes/inspect-web/engine/Program.cs` → `ListStyleOptions()` returns
  `BrowserStyleOption[]` (`Id, Title, Summary, Tier` *as a bare string*,
  `ByteDivergent, OracleEndorsed, ConflictGroup`).
- There is no `ListStyleTiers()` / `BrowserStyleTier`, so tier `Title` /
  `Summary` / `Order` / `ByteDivergent` never reach the client. That is why the
  JS side still hardcodes them.

So this PR is **necessary but not sufficient** for native Wasm consumption; it
needs a small host-half companion.

### The enabling follow-up (host half of #3486)

1. **Project the descriptors across the boundary.** Add a `BrowserStyleTier`
   record mirroring `StyleOptionTierDescriptor` (`Id` as the enum
   `.ToString()` grouping key, `Title`, `Summary`, `Order`, `ByteDivergent`) and
   a `[JSExport] ListStyleTiers()` that serializes `StyleOptionCatalog.Tiers`
   verbatim — the same *project-the-catalog-verbatim* pattern `ListStyleOptions`
   already follows, so a tier added to the enum surfaces in the browser with no
   further edit.
2. **Expose it in the bridge** as `inspectListStyleTiers` (alongside
   `inspectListStyleOptions`) and fetch it once into state next to
   `state.styleOptions`.
3. **Retire the bespoke array.** Replace `TASTE_TIERS` with a walk of the
   projected tiers: render groups in `Order` (never sort client-side — the
   registry is pre-sorted), filter `Options` on `Tier`, use `Title` as the group
   heading, `Summary` as a group blurb, and drive a group-level badge off the
   tier `ByteDivergent` flag (dropping the hand-rolled `"· byte-divergent"`
   label suffix).

Net effect on the site: zero locally-held label, blurb, or ordering — the
prototype renders, the product owns the taxonomy. Two visible consequences worth
noting up front: consuming `Order` **reorders** the site's groups (Synthesis
moves ahead of Lens — intended), and the group headings/summaries switch to the
product's wording.

### One request on *this* PR, because the Summaries become user copy

The moment a picker renders them, the tier `Summary` strings are **user-facing
group blurbs** (the browser would show them verbatim, HTML-escaped). Two read as
spec/dev prose rather than UI copy:

- **Spelling** — `"…such as this. qualification or var."` escapes to plain text
  and reads as a broken sentence once the `this.` / `var` code tokens lose their
  formatting.
- **Name synthesis / Lens** — `"compile-back fidelity gates"` is internal
  vocabulary; a site user won't know the term. Prefer user-facing phrasing
  ("recompiles to different bytes than the shipped output").

`Title`, `Order`, and `ByteDivergent` are exactly right as-is — the tier-level
`ByteDivergent` in particular lets the browser warn about a whole group without
inspecting its knobs.

I'll pick up the host-half work once this merges and I can integrate `main`.

richlander:
## Feedback addressed — tier summaries reworded as user-facing copy

Thanks — the point that lands is that these strings **stop being dev prose the
moment a picker renders them**, and I had written them as if only a reader of
this file would see them. Fixed in `38ad8ecbe`, head now `112a255c7`.

### What changed

All four, not just the two flagged — a mixed register across one group list
would read worse than either register alone:

| Tier | Before | After |
| --- | --- | --- |
| Formatting | …Token- and byte-identical to the shipped default. | Layout only — whitespace and line breaks. The code itself is unchanged, and compiles to identical IL. |
| Spelling | An equally faithful token choice, such as this. qualification or var. Byte-preserving (IL-identical). | An equally faithful way to spell the same code, such as qualifying a member with this or writing var. Compiles to identical IL. |
| Name synthesis | Readable synthesized names. The emitted IL is unchanged, but Annotated-IL name alignment is traded. | Readable invented names for locals that have none of their own. Compiles to identical IL, but these names no longer match the ones the IL uses. |
| Style lenses | Behavior-faithful but not opcode-faithful: …so it never feeds the compile-back fidelity gates. | Keeps what the code does, but not the exact bytes: this rendering recompiles to different IL than the shipped output, so it is excluded from byte-level fidelity checking. |

Two specific fixes behind that table:

- **Spelling** no longer names its examples as code tokens. The old text broke
  precisely because `this.` carries a trailing period that reads as a sentence
  end once formatting is stripped; the new wording spells the example as prose
  ("qualifying a member with this") so it survives escaping intact.
- **Lens and Synthesis** named the *mechanism that measures the cost*
  ("compile-back fidelity gates", "Annotated-IL name alignment") rather than the
  cost itself. Now they name what the choice costs the reader — different bytes,
  names that no longer match the IL — which is the part a site user can act on.

### And the reason it happened

The `Summary` doc comment described it as "one-sentence statement of the fidelity
contract", which invited exactly the register I used. It now says the string is
user-facing copy rendered verbatim as escaped plain text, must read as a complete
sentence without code formatting, and should name what a choice costs the reader
rather than the gate that measures it. Stating that where the strings are
authored is what stops the next tier from drifting back.

No contract, `Id`, `Order`, or `ByteDivergent` change — the reworded text is the
same claim in plain language, and nothing pins these strings.

### On the host half

Agreed on all three steps, and the `BrowserStyleTier` + `ListStyleTiers()`
shape is right — projecting `Tiers` verbatim is the same pattern
`ListStyleOptions` already follows. Two notes for when you pick it up:

- **Render `Tiers` in the order given; never sort client-side.** The list is
  pre-sorted by `Order`, and `Order` is deliberately *not* the enum ordinal — the
  enum declares `Lens` before `Synthesis`, the registry displays Synthesis third
  and Lens fourth. A client that sorts by the enum token, or trusts the token
  order, gets a different layout than the product intends.
- **The reorder you flagged is the intended one**, and the label change is worth
  more than it looks: the prototype's current `"Spelling (this.)"` is already
  stale, because the `Spelling` tier now also holds the `var-spelling-style`
  axis. That drift is the whole argument for the registry.

### Validation at the new head

`origin/main` integrated (it had moved 4 commits) and reconciled with the merge
you pushed — by merging, never force-push, so the reviewed history is intact.
`git diff origin/main...HEAD` is still exactly three files, insertions-only.

```
dotnet build dotnet-inspect.slnx -c Release                        # 0 errors
StyleOptionCatalogTests + ByteNeutralityGateTests                  # 40 / 0 failed
AppliedTasteSectionTests + RenderStyleConfigTests                  # 79 / 0 failed
```

The head moved, so the earlier clean round no longer covers it; a fresh review
round at `112a255c7` is running now and I'll post the result before calling this
ready.

richlander:
## Review round 2 — CLEAN at `112a255c7`

The head moved after the first clean round (the summary reword plus two
`origin/main` integrations), so that round no longer covered it. **MAI-Code**,
fresh isolated detached worktree at the exact head, verdict **CLEAN — no
blocking findings**. Worktree verified clean and removed.

The round was aimed at what the move could have broken, not a repeat of round 1:

| # | Attack point | Result |
| --- | --- | --- |
| A | Are the reworded summaries still **true**? Cross-check each tier's IL claim against its knobs' own `ByteDivergent` flags and the enum's contract docs | Consistent: Formatting/Spelling byte-neutral and IL-identical, Synthesis byte-neutral but name-divergent, Lens byte-divergent and gate-excluded. No stale pin of the old prose anywhere in tests, snapshots, or baselines |
| B | The expanded `Summary` doc asserts the string "must read as a complete sentence" — ungated claim needing a gate or an unverified marking? | An authoring instruction, not a safety/soundness/faithfulness property, so no gate required; non-empty is the right level for the tests to enforce |
| C | Did the merges drop, duplicate, or alter any of this PR's lines? Does anything main newly turned on (BannedSymbols, analyzers) bite? | `git diff origin/main...HEAD` still exactly three files, insertions-only; solution builds clean with main's new analyzer config |
| D | Do the gates still bite at this head? | Independent tamper (move the byte-divergent `prefer-long-literal-suffix` to `Formatting`) failed the expected two tests; restored and re-passed |
| E | Anything round 1's points would answer differently now — additive-ness, `GetTier` throwing, `Order` vs enum ordinal, new sorting in main | Unchanged; nothing merged from main sorts or compares `StyleOptionTier` |

Reviewer's independent runs at this head (Release): solution build 0 errors;
`StyleOptionCatalogTests` + `ByteNeutralityGateTests` 40/0; `AppliedTasteSectionTests`
+ `RenderStyleConfigTests` 79/0.

Worth recording that A was the point of running this round rather than waving the
reword through as cosmetic: rewriting a fidelity contract into plain language is
exactly where an inaccurate claim can slip in, and it now has an independent
check that each tier's sentence matches what its knobs actually do.

No findings, so nothing was dismissed and the head is unchanged from what was
reviewed.

richlander:
Ready to merge

Full CI green at `112a255c7`, including the `ci-required` aggregate (test, decompiler-gates, build-net10, csharp-diff-smoke, markdownlint, changes; il-diff-smoke and pack correctly path-skipped). Two clean review rounds, the second at this exact head. PR is MERGEABLE/CLEAN with no outstanding findings.

Non-blocking follow-ups, all tracked: the host half of #3486 (`BrowserStyleTier` / `ListStyleTiers()` and retiring `TASTE_TIERS`) on the wasm branch, #3517 (product-owned choice identity), and #3516 (annotated-source export discards taste).

PR #3608, opened 2026-07-31
Make CI block merges through one aggregate required check (#3523)

### LINKED ISSUE #3523: No CI job blocks a merge: the main ruleset declares no required status checks
Found during adversarial review of #3520 by two independent reviewers.

## The gap

The active `main` ruleset has `enforcement: active` and rules for `deletion`, `non_fast_forward`, and `pull_request` — but **no `required_status_checks` rule**. The legacy branch-protection API returns 404 (protection is by ruleset, not the legacy API), which makes this easy to miss.

```console
$ gh api repos/richlander/dotnet-inspect/rulesets/12507218 --jq '[.rules[].type]'
["deletion","non_fast_forward","pull_request"]

$ gh api repos/richlander/dotnet-inspect/branches/main/protection
gh: Branch not protected (HTTP 404)
```

Consequence: **no job in `ci.yml` blocks a merge.** Not `test`, not `markdownlint`, not `pack`, not the new `decompiler-gates`. A red PR is mergeable. CI is advisory, and merge safety currently rests on whoever clicks merge noticing a red check.

## Why it matters more than it looks

The repo invests heavily in gates whose value is that they *stop* something. #3432 was the same class of bug one level down: gate tests carried doc comments asserting "fails CI", but no pre-merge lane ran them, so five regressions (#3489–#3493) landed unseen. This is that bug at the workflow level — the checks run and report, and nothing consumes the report.

It also interacts badly with a failure mode that is already known here: a job that **times out** is `cancelled`, not `failed`. A cancelled check is not a failed check, so it is even less likely to be noticed, and it satisfies no `failure()` condition. That is exactly how the Deep Inspect notifier went silent.

## Two things to decide

**1. Which checks are required.** Path-gated jobs (`csharp-diff-smoke`, `il-diff-smoke`, `pack`, `decompiler-gates`) do not run on every PR, and a required check that never runs blocks merges forever. The standard fix is an aggregate job:

```yaml
  ci-required:
    needs: [changes, markdownlint, test, build-net10, decompiler-gates, csharp-diff-smoke, il-diff-smoke, pack]
    if: always()
    runs-on: ubuntu-26.04
    steps:
      - name: Verify no needed job failed or was cancelled
        run: |
          echo '${{ toJSON(needs) }}'
          ! echo '${{ toJSON(needs) }}' | grep -E '"result": *"(failure|cancelled)"'
```

Then require only `ci-required` in the ruleset. `skipped` must pass (path-gated jobs) while `cancelled` must not — that distinction is the whole point and is easy to get backwards.

**2. Whether to require it at all.** This is a policy call for @richlander, not something a PR can decide: making checks required changes the merge workflow for every contributor and agent. There is a legitimate position that a small, fast-moving repo prefers advisory CI. But if that is the position, the docs should say so, because several gate doc comments and `AGENTS.md` currently read as though CI enforces things it does not.

Note this issue is **not** a blocker for #3520 — that job is exactly as enforcing as the existing `test` job. It is the ceiling on what any gate in this repo can currently achieve.

Related: #3432.

richlander:
## The gap

The `main` ruleset has `enforcement: active` with rules for `deletion`, `non_fast_forward`, and `pull_request` — but **no `required_status_checks` rule**. The legacy branch-protection API returns 404 because protection is by ruleset, not the legacy API, which makes this easy to miss.

```console
$ gh api repos/richlander/dotnet-inspect/rulesets/12507218 --jq '[.rules[].type]'
["deletion","non_fast_forward","pull_request"]

$ gh api repos/richlander/dotnet-inspect/branches/main/protection
gh: Branch not protected (HTTP 404)
```

**No job in `ci.yml` blocks a merge.** Not `test`, not `markdownlint`, not `pack`, not `decompiler-gates`. A red PR is mergeable. Every gate this repository has built — the fidelity dockets, the byte-neutrality gate, the decompiler lane ratchet — is advisory, and merge safety rests on whoever clicks merge noticing a red check.

That is #3432 one level up. There, gate tests carried doc comments asserting "fails CI" while no pre-merge lane ran them, so five regressions (#3489–#3493) landed unseen. Here the checks run and report, and nothing consumes the report.

## Why an aggregate job

No existing job can be required directly. All of them are path-gated, and **a required check that does not run on a given PR is reported as "Expected" forever and blocks the merge permanently.** So this adds `ci-required`: it always runs, depends on every other job, and collapses them into one stable context for the ruleset to require.

Three details carry the whole contract, and each is easy to get backwards:

| Detail | Why |
| --- | --- |
| `if: always()` | Without it the job inherits the default `success()` condition and is skipped the moment a dependency fails. A skipped required check does not block a merge — the failure it exists to catch would be exactly what silences it. |
| `skipped` passes, `cancelled` fails | Path-gated jobs skip on most PRs and that is correct. A cancelled job is one that hit its timeout. This repo has been burned by that distinction twice: the Deep Inspect notifier stayed silent because a cancelled job satisfies no `failure()` condition (#3432), and seven decompiler classes sat at a 45-minute timeout unnoticed (#3489–#3493). |
| Allow list, not deny list | Only `success` and `skipped` pass. A deny list of `failure\|cancelled` would admit any result state GitHub adds later — the fail-open direction. This fails closed. |

`changes` is among the dependencies deliberately. If it fails, every downstream job is skipped, so a check reading only the gated jobs would see nothing but `skipped` and pass a PR on which no test ran at all.

## Keeping the gate from rotting

A merge gate that quietly becomes a no-op is worse than none, because it reads as protection. Two self-checks run inside the job:

- **`Verify this gate depends on every other job`** parses the workflow and asserts set equality between its job list and this job's `needs`, failing in *both* directions — a job missing from `needs`, and a `needs` entry naming a job that no longer exists. A hand-maintained `needs` list is an allow list, and an unchecked allow list goes stale silently; that is precisely the shape of #3584, where a one-directional docket un-gated 46 fixture rows.
- **`Self-test the result filter`** runs the same filter the gate uses over a synthetic payload carrying one of each result state (`success`, `skipped`, `failure`, `cancelled`, and an unknown `neutral`) and asserts it selects exactly the blocking ones. A filter that accepted everything would be indistinguishable from a green run.

Both consume the filter from a single `RESULT_FILTER` definition, so the expression proven non-vacuous is the expression that gates the merge.

## Validation

- YAML parsed with YamlDotNet; `needs` contains all 8 sibling jobs, and each `run:` block was dumped post-parse to confirm the embedded Python lands at column 0 after block-scalar indentation stripping.
- The remaining behavior is runner-side by nature, so **this PR's own CI run is the evidence**: `ci-required` must report, its two self-checks must pass, and the coverage check must agree with the workflow. Additionally I will force a real dependency failure on a scratch commit to confirm the gate *blocks* rather than merely reports, then restore the reviewed head.

## Rollout, and what it costs

**Merging this PR changes nothing by itself.** The ruleset rule is a repository-settings change applied *after* this lands, deliberately in that order: requiring `ci-required` before the job exists on `main` would block every open PR with no path forward.

Two things to know before enabling it:

- The ruleset has `bypass_actors: []` and `current_user_can_bypass: "never"`. There is no escape hatch, so the check must be correct before it is required.
- There are currently 15 open PRs. Their heads predate this job, so after the rule is enabled each will show `ci-required` as "Expected" until it merges `main` — a one-time, unavoidable cost of turning required checks on.

Closes #3523.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,8 +16,14 @@ env:
 # (test/pack are PR-only), so there is little to cancel. release.yml
 # builds all release packages fresh at publish time and does not consume CI
 # artifacts.
+#
+# The group keys on the pull request *number*, not `github.head_ref`. Branch
+# names are not unique across forks, so two PRs both branched as `patch-1`
+# would share a group and cancel each other. That was merely wasteful while CI
+# was advisory; now that `ci-required` gates merges, the loser's required check
+# would land as `cancelled` and block that PR until someone re-ran it by hand.
 concurrency:
-  group: ci-${{ github.head_ref || github.ref }}
+  group: ci-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
@@ -888,13 +894,18 @@ jobs:
       - name: Uninstall tool
         run: dotnet tool uninstall --global dotnet-inspect
 
-  # The single check the `main` ruleset requires.
+  # The one check the `main` ruleset is meant to require, and the only one that
+  # safely can be.
   #
   # Every other job here is path-gated, so none of them can be required
   # directly: a required check that does not run on a given PR is reported as
   # "Expected" forever and blocks the merge permanently. This job always runs
   # and collapses the others into one result, so the ruleset has exactly one
-  # stable context to require.
+  # stable context to name.
+  #
+  # Adding the `required_status_checks` rule is a repository-settings change
+  # applied once this job exists on `main`; requiring it earlier would block
+  # every open PR with no way forward.
   #
   # Three details carry the whole contract, and each is easy to get backwards:
   #
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -887,3 +887,118 @@ jobs:
 
       - name: Uninstall tool
         run: dotnet tool uninstall --global dotnet-inspect
+
+  # The single check the `main` ruleset requires.
+  #
+  # Every other job here is path-gated, so none of them can be required
+  # directly: a required check that does not run on a given PR is reported as
+  # "Expected" forever and blocks the merge permanently. This job always runs
+  # and collapses the others into one result, so the ruleset has exactly one
+  # stable context to require.
+  #
+  # Three details carry the whole contract, and each is easy to get backwards:
+  #
+  # 1. `if: always()` is mandatory. Without it this job inherits the default
+  #    `success()` condition, is skipped the moment any dependency fails, and a
+  #    skipped required check does not block a merge -- the failure it exists to
+  #    catch would be exactly what silences it.
+  #
+  # 2. `skipped` must pass but `cancelled` must not. Path-gated jobs are skipped
+  #    on most PRs and that is correct. A cancelled job is a job that hit its
+  #    timeout, and this repository has already been burned by that distinction
+  #    twice: the Deep Inspect notifier stayed silent because a cancelled job
+  #    satisfies neither `failure()` nor a failed conclusion (#3432), and seven
+  #    decompiler test classes sat at a 45-minute timeout unnoticed (#3489-#3493).
+  #    Treating `cancelled` as success would rebuild that hole at the merge gate.
+  #
+  # 3. The check is an allow list, not a deny list. Only `success` and `skipped`
+  #    pass; anything else blocks. A deny list of `failure|cancelled` would admit
+  #    any result state GitHub adds later, which is the fail-open direction. This
+  #    fails closed, matching the repository rule that failure stays visible.
+  #
+  # `changes` is in `needs` deliberately. If it fails, every downstream job is
+  # skipped, so a check that looked only at the gated jobs would see nothing but
+  # `skipped` and pass a PR on which no test ran at all. `changes` reporting
+  # `failure` is what closes that path.
+  ci-required:
+    needs:
+      - changes
+      - markdownlint
+      - test
+      - build-net10
+      - decompiler-gates
+      - csharp-diff-smoke
+      - il-diff-smoke
+      - pack
+    if: always()
+    runs-on: ubuntu-26.04
+    timeout-minutes: 5
+    env:
+      # Defined once and consumed by both the self-test and the real check, so
+      # the thing proven non-vacuous is the same thing that gates the merge.
+      RESULT_FILTER: >-
+        to_entries[]
+        | select(.value.result != "success" and .value.result != "skipped")
+        | "\(.key)=\(.value.result)"
+    steps:
+      - uses: actions/checkout@v6
+
+      # A hand-maintained `needs` list is an allow list, and an allow list that
+      # nothing checks goes stale silently: add a job to this workflow, forget
+      # this line, and the merge gate stops covering it while still reporting
+      # green. That is the same shape as #3584, where a one-directional docket
+      # un-gated 46 fixture rows. So the workflow's own job list drives the
+      # required set, and both directions fail -- a job missing from `needs`
+      # and a `needs` entry naming a job that no longer exists.
+      - name: Verify this gate depends on every other job
+        env:
+          NEEDS: ${{ toJSON(needs) }}
+        run: |
+          set -euo pipefail
+          python3 -c '
+          import json, os, sys, yaml
+          with open(".github/workflows/ci.yml") as f:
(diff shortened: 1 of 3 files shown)


richlander:
## Review reconciliation

Two independent adversarial reviews on isolated worktrees, re-run on each fixed head. **Both are CLEAN on `49cb47b7`.**

Reviewers: **MAI-Code** and **GPT-5.6-Sol**. Four heads reviewed: `a45d59f7` → `71f545fc` → `f5c7f7fa` → `bde5e9d1` → `49cb47b7`.

### Findings actioned

| --- | --- | --- | --- |
| 1 | GPT-5.6-Sol | `concurrency` keyed on `github.head_ref`, which is not unique across forks. Two PRs branched `patch-1` cancel each other; with a required check the loser lands as `cancelled` and is blocked until re-run by hand. | Fixed in `71f545fc` — keys on `github.event.pull_request.number`, falling back to `github.ref` for pushes to main. |
| 2 | Both, independently | The workflow comment and both docs asserted in the present tense that the `main` ruleset requires `ci-required`. It does not, and cannot until this job exists on `main`. | Fixed in `f5c7f7fa` — reworded to intent, so the wording is understated after the rule is enabled rather than false before it. |
| 3 | GPT-5.6-Sol | **The sharpest catch of the review.** `decompiler-gates` carried its *own* NOTE — "not merge-blocking today, because the `main` ruleset declares no required status checks at all" — that neither of us had cited. I had updated the parallel note in the pipeline doc and missed this one. True today, false the instant the rule is enabled. | Fixed in `bde5e9d1`. Replaced with the durable statement: this path-gated job must never be named as a required check directly, and its result reaches `ci-required`. A repo-wide grep for other claims of this shape found none. |
| 4 | GPT-5.6-Sol | Workflow comment tense did not match the doc. | Fixed in `49cb47b7`. |

### Findings refuted

**MAI-Code, blocking: "PyYAML is not preinstalled on the runner; this step will fail with `ModuleNotFoundError` and, once required, wedge every merge."**

Refuted by direct evidence. Run `30600496149` executed the step on `ubuntu-26.04` and printed its success line:

```
Gate covers every job in the workflow.
```

That line is only reachable after `import yaml` succeeds and `yaml.safe_load` parses the workflow. GPT-5.6-Sol independently verified the same run and identified the image as `20260726.86.1`. I had also found a web source claiming PyYAML was absent; the runner disagrees with it, and the runner is the oracle. Asked to re-examine with that evidence, MAI-Code withdrew the finding.

### Findings declined, with reasoning

**GPT-5.6-Sol / MAI-Code: "validate that each `skipped` job was *expected* to skip."**

Declined. Reconstructing the `changes` classification inside `ci-required` would make the gate a second implementation of the filter, and the two would drift — `AGENTS.md` explicitly prohibits harness-side logic that duplicates or compensates for product behavior. Instead the actual guarantee is now documented: a green `ci-required` means "nothing that ran went wrong", not "everything ran".

The two *concrete* instances GPT found are real and are filed as **#3614** rather than fixed here, per the standing instruction to file unrelated defects and land the fix quickly:

- `.markdownlint.yaml` matches no filter, so editing the lint config runs no lint.
- `eng/restore-ilassembler.sh` sets `ILROUNDTRIP=true`, but the only consumer is gated on `code`, so that output has no consumer at all.

## Evidence

**The gate blocks, not merely reports.** Throwaway PR #3612 (now closed, branch deleted) forced the `changes` job to `exit 1`:

```
failure  changes
skipped  build-net10, markdownlint, test, il-diff-smoke, pack, decompiler-gates, csharp-diff-smoke
failure  ci-required
```

with the log line:

```
##[error]CI is not green; these jobs did not succeed or skip:
changes=failure
```

That confirms both `if: always()` fires when a dependency fails, and that including `changes` in `needs` is what prevents a green result from a sea of `skipped`.

**Non-vacuity, on the real runner**, every run:

```
Filter discriminates correctly: [broke=failure,timeout=cancelled,future=neutral]
```

The unknown `neutral` state is selected as blocking, which is the fail-closed direction an allow list buys over a deny list.

**Coverage check runs green:** `Gate covers every job in the workflow.`

**Final head `49cb47b7`: all nine jobs green,** including `ci-required`. Earlier fully-green runs at `a45d59f7` (`30600496149`) and `71f545fc` (`30601647933`).

The `ci-required` job itself takes ~4 seconds.

## Next step after merge

Applying the `required_status_checks` rule naming `ci-required` to the `main` ruleset, authorized by @richlander. Note for anyone with an open PR: your PR will show `ci-required` as "Expected" until you merge `main` to pick up the job. That is the one-time cost of turning required checks on.

Ready to merge.

PR #3609, opened 2026-07-31
Stop two package tests from asserting host filesystem and restore layout

### LINKED ISSUE #3519: Two CommandExecutionTests cases fail on Windows on a clean main

richlander:
Fixes #3519.

Both `CommandExecutionTests` failures on Windows are the test over-specifying its environment, not product behavior.

## `Layout_Count_CountsFilesRatherThanRenderedTreeLines`

The test pinned the manifest as `Newtonsoft.Json.nuspec` with `StringComparison.Ordinal`. The nuspec filename casing is a property of the **source the package resolved from**, not of the package:

- expanded from the `.nupkg` -> authored casing (`Newtonsoft.Json.nuspec`)
- restored into the NuGet global packages folder -> normalized, lowercased id (`newtonsoft.json.nuspec`)

Instrumenting `extractPath` at the `--layout` early exit showed the resolver returning the global packages folder on this host, because `NUGET_PACKAGES=C:\Nuget` is set:

```
PROBE extractPath=C:\Nuget\newtonsoft.json\13.0.4
PROBE nuspec=C:\Nuget\newtonsoft.json\13.0.4\newtonsoft.json.nuspec
```

The rendered tree was correct; the assertion was asserting which source resolved the package. Renaming the `.nupkg`-expanded copy under `package-content-v2` (and moving that directory aside entirely) left the output unchanged, confirming that copy is not what the run reads.

The claim the assertion carries is that the nuspec is a reachable package file, so it now compares case-insensitively. The two assertions that carry the test's actual point -- `count == 20` and `renderedLines > count` -- are untouched and passed throughout.

## `Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown(readmePath: "images/logo.png.")`

Win32 strips trailing dots before a path reaches the filesystem:

```
[IO.File]::WriteAllText("\logo.png.", "x")
created: 'logo.png'
Exists('logo.png.') = True
Exists('logo.png')  = True
```

So the fixture's declared readme lands on disk as `logo.png`, the declared path names nothing the package lists, and `--content --path "images/logo.png."` exits 1.

This is not a fixture artifact that a better fixture can dodge. Staging a `.nupkg` whose zip entry really is `images/logo.png.` and running the product against it fails the same way, because the product re-extracts to disk and loses the dot identically:

```
Error: --bare found no selected package content.
exit=1
```

The case therefore cannot be staged on Windows rather than behaving differently there, so it skips with that reason via the suite's existing `Assert.SkipWhen` convention. It stays covered on `linux`/`macOS` and the other three theory cases (`images/logo.png`, `.png`, `.README`) still run on Windows.

## Validation

`dotnet run --project src/dotnet-inspect.Tests -c Release` on Windows, after a full `dotnet build dotnet-inspect.slnx -c Release`:

```
dotnet-inspect.Tests  Total: 2609, Errors: 0, Failed: 0, Skipped: 11, Not Run: 0, Time: 422.674s
```

Previously 2 failed. No product code changes; no other platform's behavior changes.

### DIFF
--- a/src/dotnet-inspect.Tests/CommandExecutionTests.cs
+++ b/src/dotnet-inspect.Tests/CommandExecutionTests.cs
@@ -4651,7 +4651,13 @@ public partial class CommandExecutionTests
         Assert.True(
             renderedLines > count,
             $"expected the tree to render more lines ({renderedLines}) than the {count} files it counts");
-        Assert.Contains("Newtonsoft.Json.nuspec", rendered, StringComparison.Ordinal);
+
+        // The manifest is a reachable package file, which is the claim; its filename casing is
+        // not. A package restored into the NuGet global packages folder carries the manifest under
+        // the normalized (lowercased) id, while one expanded from the .nupkg keeps the authored
+        // casing, so pinning "Newtonsoft.Json.nuspec" ordinally asserts which source resolved the
+        // package rather than that the nuspec is listed.
+        Assert.Contains("Newtonsoft.Json.nuspec", rendered, StringComparison.OrdinalIgnoreCase);
     }
 
     [Fact]
@@ -12171,6 +12177,16 @@ public partial class CommandExecutionTests
         // something else. The role answers a document's kind only where the name is silent;
         // letting a declaration override a stated kind would run the link rewriter over a PNG
         // and hand back a corrupted file, which is the outcome this command prevents.
+
+        // Win32 strips trailing dots from a path before it reaches the filesystem, so a package
+        // entry named "logo.png." lands on disk as "logo.png" no matter who expands it -- both
+        // this fixture and the product's own extraction. The declared path then names nothing the
+        // package lists, and the case cannot be staged on Windows at all rather than behaving
+        // differently there. It stays covered on every other platform.
+        Assert.SkipWhen(
+            OperatingSystem.IsWindows() && readmePath.EndsWith('.'),
+            "Windows cannot hold a file whose name ends in a dot: the name is normalized away before it reaches the filesystem.");
+
         var tempDir = Path.Combine(Path.GetTempPath(), $"package-test-{Guid.NewGuid():N}");
         var packageRoot = Path.Combine(tempDir, "content");
         var declared = Path.Combine(packageRoot, readmePath.Replace('/', Path.DirectorySeparatorChar));


richlander:
## Review reconciliation

Per the repository's review tiering, this is a test-only change with no product delta and a bounded blast radius, so it took a single review.

**MAI-Code, at the exact head `cbcda91a`, in an isolated detached review worktree: APPROVE, no findings.**

It independently re-derived the two factual claims rather than taking them from the description:

- `C:\Nuget` is `C:\Nuget` and `C:\Nuget\newtonsoft.json\13.0.4` holds `newtonsoft.json.nuspec`. It then ran `--layout` against an explicit local `.nupkg` and got `Newtonsoft.Json.nuspec`, and against the global-packages-resolved reference and got `newtonsoft.json.nuspec` -- confirming the casing is source-dependent by direct A/B rather than by inference.
- Its own filesystem probe reproduced the trailing-dot normalization.

It also ran the full class: `-class "*CommandExecutionTests*"` -> 783 total, 0 failed, 1 skipped.

### One thing it explicitly declined, which I then did myself

It said: *"I did not tamper with product code to perform a synthetic regression test."* That was the one attack point I most wanted covered, because relaxing `Ordinal` to `OrdinalIgnoreCase` is exactly the shape of change that quietly retires an assertion. An approval that skips it does not establish the assertion still bites.

So I tamper-tested it in a throwaway worktree at `cbcda91a`.

**First attempt, and why it was not good enough.** Dropping the nuspec from the layout listing entirely does fail the test -- but at the *count* assertion, not the one I changed:

```
Expected: 20
Actual:   19
  at CommandExecutionTests.cs(4645,0)
```

That proves the test is non-vacuous; it does not isolate `Assert.Contains`.

**Second attempt, which isolates it.** Renaming the manifest in the projection keeps the count at exactly 20 and changes only the rendered name:

```csharp
.Select(p => p.Replace(".nuspec", ".tamperedspec", StringComparison.OrdinalIgnoreCase))
```

```
Assert.Contains() Failure: Sub-string not found
String:    "├─ lib\r\n│  ├─ net20\r\n│  │  ├─ Newtonsoft.Json.dll\r"···
Not found: "Newtonsoft.Json.nuspec"
  at CommandExecutionTests.cs(4660,0)
```

Line 4660 is the relaxed assertion. It fails on its own, with the count assertion satisfied. The assertion still detects a nuspec that goes missing from the tree; what it no longer detects is which of two legitimate resolution sources served the package, which was never its claim.

The tamper worktree was reverted and removed; nothing from it is in this branch.

### Design questions the reviewer was asked to rule on

- *Should Change 1 have been a product fix?* Reviewer: no -- the product is faithfully reflecting the filename from the resolution source. I agree. Normalizing the rendered name would make `--layout` report a file the extract directory does not contain.
- *Should this PR file a follow-up for the Windows trailing-dot limitation?* Reviewer: useful but not necessary to land this. I agree it should not block, and I did not file one: the CLI's current behavior there is a visible, honest failure (`Error: --bare found no selected package content.`, exit 1), not success-shaped empty output, so it does not trip the "keep failure visible" rule. Happy to file it if you would rather have it tracked.

### Validation on the merge head

`dotnet build dotnet-inspect.slnx -c Release` then:

```
dotnet-inspect.Tests  Total: 2609, Errors: 0, Failed: 0, Skipped: 11, Not Run: 0, Time: 422.674s
```

Previously 2 failed.

richlander:
Ready to merge.

- MAI-Code review at `cbcda91a`: APPROVE, no findings. The one attack point it declined (tamper-testing the relaxed assertion) I covered myself; the isolating tamper and its output are in the reconciliation above.
- CI green, `mergeStateStatus: CLEAN`. `test (ubuntu-26.04, linux-x64)` SUCCESS confirms the `images/logo.png.` theory case still executes and passes on Linux, so the Windows skip narrows coverage on Windows only.
- Windows, Release, full project: 2609 total, 0 failed, 11 skipped (was 2 failed).

No follow-up work is outstanding for this PR. Separately from it, and non-blocking: I filed #3611 for the `.gitattributes` renormalization gap found while measuring this pass, and closed #3445, #3496 and #3509 with re-measured evidence.

richlander:
## Unblocked: merged `main` to pick up the required check

This PR was stuck at `BLOCKED` for a reason that is not visible from its own checks — they were all green or correctly skipped:

```
changes                        SUCCESS
test (ubuntu-26.04, linux-x64) SUCCESS
decompiler-gates               SUCCESS
markdownlint / build-net10 / csharp-diff-smoke / il-diff-smoke / pack   SKIPPED
```

The problem is what is *absent*. The `main` ruleset now requires exactly one context:

```
$ gh api repos/richlander/dotnet-inspect/rulesets/12507218
required_status_checks -> context: ci-required
```

`ci-required` landed on `main` in #3608 (#3523), after this branch was cut. A required check that never runs is reported as "Expected" indefinitely and blocks the merge permanently — which is precisely the hazard that job's own comments warn about, here applying to a branch that predates it rather than to a path-gated job.

Merged `origin/main` in (`5a46491e`, no conflicts) so `ci-required` actually runs. Per `AGENTS.md` this is a merge, not a rebase, because the branch is already public and reviewed.

The reviewed change is untouched by the merge — still one file, +17/−1:

```
src/dotnet-inspect.Tests/CommandExecutionTests.cs | 18 +++++++++++++++++-
```

Both hunks verified present after the merge: the `OrdinalIgnoreCase` relaxation at the nuspec assertion, and the `Assert.SkipWhen` for the trailing-dot case.

**Any other open PR branched before #3608 will be blocked the same way** and needs the same `main` merge.

### Independent confirmation that this fix is needed

The Windows CI leg in #3625 ran the CLI suite on `windows-latest` against a tree *without* this fix, and reproduced exactly the failure this PR addresses:

```
DotnetInspector.Tests.CommandExecutionTests.Package_DeclaredNonMarkdownReadme_IsStillNotMarkdown(readmePath: "images/logo.png.") [FAIL]
  Expected: 0
  Actual:   1

dotnet-inspect.Tests  Total: 2765, Errors: 0, Failed: 1, Skipped: 10
```

So the trailing-dot half is now confirmed on a clean GitHub-hosted runner, not just on my machine. The nuspec-casing half did **not** reproduce there, which matches its root cause: it depends on `NUGET_PACKAGES` pointing at the global packages folder, which the runner does not set.

#3625 is blocked on this PR — merging it first would land a red `test-windows` and a red `ci-required` on `main`.

richlander:
`ci-required` now runs and passes; `mergeStateStatus` is `CLEAN`.

```
changes                        SUCCESS
test (ubuntu-26.04, linux-x64) SUCCESS
decompiler-gates               SUCCESS
ci-required                    SUCCESS
markdownlint / build-net10 / csharp-diff-smoke / il-diff-smoke / pack   SKIPPED
```

The reviewed change is unchanged by the merge (one file, +17/-1). Review was clean at the pre-merge head, and the merge touched no product or test source.

Ready to merge.

Once this lands I will re-run #3625, which should then go green — its only failure is the trailing-dot test this PR fixes.

PR #3613, opened 2026-07-31
Point type's bare -S at the fixed overview

richlander:
`type X -S` now renders a bounded overview whose shape does not depend on which type you are looking at.

## What changed

Bare `-S` on `type` resolved to the **Info set** — sections chosen by the `-v:m` target-section axis, which says nothing about how long a section is. So bare `-S` grew with the type:

| `type X -S` | sections (base -> head) | lines (base -> head) |
| --- | --- | --- |
| `System.String` | 7 -> 1 | 125 -> 13 |
| `System.Int32` | 3 -> 1 | 156 -> 13 |
| `System.Span<T>` | 5 -> 1 | 127 -> 13 |
| `System.Exception` | 3 -> 1 | 28 -> 12 |
| `System.DayOfWeek` | 1 -> 1 | 13 -> 11 |

A ~12x line spread becomes ~1.2x. It now resolves to the **fixed overview**: sections that are both `SizeClass.Fixed` and `SectionCost.NetworkFree`. For a single type that is the `Type Info` section added in #3576.

This is the rule the curated catalog already implements for `package`/`library`; `type` is the first non-curated command to adopt it.

## Design

**`SectionPipeline.FixedOverviewSectionNames`** exposes the set. The definition of a fixed-overview member is single-sourced into `IsFixedOverviewMember`, which the existing curated `fixedOverview` branch in `IsRequested` now also calls, so the two cannot drift.

**Bare `-S` resolves to an explicit include set, not the `fixedOverview` flag.** `IsRequested` checks `ExplicitOnly` *before* the `fixedOverview` branch, and `Type Info` is `ExplicitOnly` (load-bearing — see #3576). `include` wins over everything in `IsRequested`, so neither the verbosity ladder nor `_curatedCatalog` moves. That is what keeps this slice to ~40 lines instead of the full catalog migration.

**The discriminator is the options record, not `singleTypeMode`.** `type` and `member` share `ApiCommand`'s preamble and both run in `singleTypeMode`. An earlier draft keyed off that flag and silently converted `member` too; `options is TypeOptions` scopes it to `type`. `MemberOptions` is a sibling of `TypeOptions`, not a subclass, so the check is exact.

## Non-action boundary

Two neighbours deliberately keep the Info set:

- **Type listing** (`type System.Uri`, a prefix match over 11 types) — every section it publishes is a per-kind member table that grows with the assembly, so it has no `Fixed` section to offer yet. A fixed identity section for list mode is the next slice.
- **`member`** — a separate command with its own overview (decompiled source, signature, learn order). Converted on its own PR, per the one-command-at-a-time directive.

`ApiShim_BareSelect_ForwardsThePresetToTypeOptions` changes its assertion from `## Method Groups` to `## Type Info`. That test pins *that the `SelectDefault` flag survives the shim's field-by-field copy*; the shim rebuilds a `TypeOptions`, so it renders whatever `type` renders. The intent is unchanged and `## Type Info` is the more distinctive marker.

## Evidence

**Differential, base `b1e802e1c` vs head, same invocation each side:**

- `type` outside single-type bare `-S`: **0 diffs / 151 invocations** — 9 types x 19 flag combos including `--all`, `-m`, `-k`, `--table`, `--json`, `-D`, `--schema`, `-S @All`, and `--all -S Fields`. (The `--all` and filter flags are here deliberately: #3576's round-3 review caught a false blast-radius claim from a matrix that omitted them.)
- `member`, **including bare `-S`**: 0 diffs / 32 invocations, 4 targets x 8 flag combos.
- `type` list mode, **including bare `-S`**: 0 diffs.
- Single-type bare `-S`: changes for every type, as intended.

**Tests** — 6 added, 1 assertion updated:

- `Type_BareSelect_RendersOnlyTheFixedOverview` (Theory over 5 types incl. an enum and a generic)
- `Type_BareSelect_DoesNotGrowWithTheType` — the boundedness claim itself
- `Type_ExplicitSelect_StillReachesGrowingSections` — `-S Properties` is unaffected
- `Type_Listing_BareSelect_KeepsTheInfoSet` — negative case for the mode boundary
- `Member_BareSelect_KeepsTheInfoSet` — negative case for the `TypeOptions` discriminator

**Mutation-verified, both halves:**

- Dropping `&& options is TypeOptions` -> 4 tests fail (all three pre-existing `Member_*BareSelect*` tests plus the new negative case). The type-side tests correctly stay green.
- Reverting `FixedOverviewSectionNames` -> `InfoSectionNames` -> 7 tests fail. Both boundary tests correctly stay green.

**Suite:** 2618 total, 1 failed, 14 skipped at head `14f28f780`. The one failure is `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on unmodified `main` (the local NuGet cache lowercases `newtonsoft.json.nuspec`).

Refs #3547.


richlander:
## Review reconciliation — round 1 (GPT-5.6) and round 2 (Gemini 3.1 Pro)

Both rounds ran at `14f28f780` in isolated worktrees. Fixes are in `27bd19be2`.

### Findings and disposition

| # | Reviewer | Finding | Severity | Disposition |
| 1 | Gemini | Empty fixed-overview set silently falls through to the full ladder | Blocking | **Fixed** — and it turned out to be worse than filed |
| 2 | Gemini | `README.md` / `progressive-disclosure.md` still describe `type -S` as `Method Groups` | Blocking | **Fixed** (+ a third file they missed) |
| 3 | GPT | Fixed overview grows with generic arity (492-char row at `Func<...,17>`) | Blocking | **Withdrawn by the reviewer**; split to #3616, claim now gated |
| 4 | GPT | `Type_Listing_BareSelect_KeepsTheInfoSet` is vacuous for its precise claim | Non-blocking | **Fixed** |

### 1 + 4 together produced a correction to this PR's own description

Writing the guard for finding 1 made `Type_Listing_BareSelect_KeepsTheInfoSet` fail with **exit 1**. That was not a bug in the guard — it was the guard doing its job on a case I had described wrongly.

I probed both pipelines directly:

```text
list  InfoSectionNames  = []   (count=0)
list  FixedOverview     = []
type  InfoSectionNames  = [Values, Constructors, Finalizer, Fields, Properties, Method Groups, ...]
type  FixedOverview     = [Type Info]
```

**The type-listing pipeline publishes no Info sections at all.** So `type <prefix> -S` has never rendered an "Info set" — it resolves to an empty include set and reaches the verbosity ladder through precisely the fallthrough Gemini flagged. My original PR text said listing "keeps the Info set." That was wrong, and it is corrected in the commit message and below.

This makes the finding more serious than filed (the fallthrough is load-bearing production behavior on `main`, not a latent hazard) and simultaneously constrains the fix: a global guard turns a working command into an error. The guard is therefore scoped to the fixed-overview path. Slice 4c gives listing a `Fixed` section and can adopt the guard then.

GPT's finding 4 was the same defect seen from the test side: the old assertion checked headings the ladder produces anyway, so it could not distinguish a working preset from a fallthrough. It now pins the empty declaration itself, so listing gaining an Info or `Fixed` section fails the test and forces a decision.

### 3 — withdrawn, and why I pushed back

I reproduced the 492-char row independently before reading the review, so the fact was never in dispute; the severity was. `Type Parameters` ships in merged #3576, and the row is **byte-identical base vs head** on the same invocation:

```bash
cmp <(base type 'System.Func`17' -S 'Type Info' --tips q) <(head type 'System.Func`17' -S 'Type Info' --tips q)   # identical
```

This PR changes only *which sections bare `-S` selects*. GPT verified this itself and withdrew the finding, agreeing it is a follow-up. Filed as **#3616** — including the real inconsistency underneath it, that `Interfaces` in the same section renders as a count (`| Interfaces | 9 |`) while `Type Parameters` concatenates.

Per AGENTS.md the claim now names its gate rather than being asserted: `Type_BareSelect_StaysBoundedAtWorstCaseArity` pins the **line count** at the platform's worst arity. Boundedness is a row-count property; width is #3616.

### 2 — docs

Fixed `README.md` and `docs/design/progressive-disclosure.md` as filed, plus `docs/design/info-view.md`, whose preset table had the same stale row and is the document the other two link to. `markdownlint` clean.

### What each round verified clean

GPT enumerated on request: `RunPreamble` has exactly two production callers, so `options is TypeOptions` is exact for every path including the deprecated shim; 19-invocation base/head differential all identical; `package`/`library` 16 invocations identical, confirming the `IsFixedOverviewMember` extraction is a pure refactor; `-D`/`--schema` agree with the render path; full suite 2618/1 (the known `Layout_Count_` failure).

Gemini independently confirmed boundedness holds vertically, that discovery and render converge, that the tests fail under mutation, and that no workflow is stranded (`member X -S` still yields the old tables).

### Re-validation at `27bd19be2`

- Full suite **2624 total, 1 failed** — `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on unmodified `main`.
- Platform listing bare `-S`: **byte-identical** to base.
- Prefix/list routes: 24 invocations, the only 2 diffs are `System.Collections.Generic.List` and `System.Convert`, both of which resolve to a **single type** and so are the intended change.
- `markdownlint` clean on all three docs.

### Corrected residual

- **`member`** keeps its Info set; it is a separate command and converts on its own PR.
- **Type listing** is unchanged and continues to fall through to the verbosity ladder, because it has no Info preset and no `Fixed` section. Giving it one is slice 4c.

Requesting one clean round at the new head `27bd19be2` before marking this ready.

richlander:
## Ready to merge

Both review families are clean at the final head. CI green, MERGEABLE/CLEAN at `c0c60d917`.

### Round ledger

| --- | --- | --- | --- |
| 1 | GPT-5.6 | `14f28f780` | 1 blocking — **withdrawn by the reviewer** after verifying it predates the PR |
| 2 | Gemini 3.1 Pro | `14f28f780` | 2 blocking — both fixed |
| 3 | Gemini 3.1 Pro | `27bd19be2` | **Clean** |
| 4 | GPT-5.6 | `27bd19be2` | 1 non-blocking (docs) — fixed |
| 5 | GPT-5.6 | `d641e7a42` | 1 non-blocking (docs) — fixed |
| 6 | GPT-5.6 | `c0c60d917` | **Clean** |

Rounds 5 and 6 were docs-only; product code has been unchanged since `27bd19be2`, the head both families cleared.

### The finding that mattered

Gemini's empty-set finding was filed as a latent hazard. Implementing the guard proved it is **not** latent, and corrected an error in my own PR description. Probing both pipelines directly:

```text
list  InfoSectionNames  = []   (count=0)
list  FixedOverview     = []
type  InfoSectionNames  = [Values, Constructors, Finalizer, Fields, Properties, Method Groups, ...]
type  FixedOverview     = [Type Info]
```

The type-listing pipeline publishes **no `Info` sections**, so `type <prefix> -S` has never rendered an "Info set" — it resolves to an empty include set and reaches the verbosity ladder through exactly that fallthrough. That is load-bearing behavior on `main` today.

It cut both ways: it made the finding more serious than filed, and it forbade the obvious fix. A global guard would turn a working command into an error, so the guard is scoped to the fixed-overview path. GPT independently reproduced the same fallthrough on the listing route.

### Corrections I made to my own claims

Recorded because three of the six rounds spent their findings on my inaccuracies rather than on the code:

1. **"Type listing keeps the Info set"** — false. It has no Info set. Corrected in the commit message, the residual below, and all three docs.
2. **My doc fix repeated the error it was fixing** — I collapsed "type listing" and "broad `member Type`" into one clause, so the corrected docs still claimed listing renders `Method Groups`. (GPT, round 5.)
3. **My width qualification named only `Type Parameters`** — GPT checked all 11 properties and found `Type` itself concatenates every generic parameter via `FormatGenericFullName`, and `Base` can too. `Interfaces` is the *only* counted collection. (Round 6.)

Finding 3 reshapes #3616: making `Type Parameters` a count would not make the section width-bounded while `Type` still renders `System.Func<T1, ..., TResult>` for identity. That is recorded on the issue as the open design question.

### Final state

Product change is ~50 lines across two files. Bare `-S` on a single `type` renders the fixed overview (`Type Info`), resolved as an explicit include set so `ExplicitOnly` never has to move; `IsFixedOverviewMember` single-sources the predicate the curated path already used; `HasNoBareSelectOverview` makes the empty-overview state fail loudly instead of silently rendering everything.

**Residual (corrected):**
- **`member`** keeps its Info set — separate command, converts on its own PR.
- **Type listing** is unchanged and still falls through to the ladder, because it has no `Info` and no `Fixed` section. Giving it one is the next slice, and `Type_Listing_BareSelect_IsUnchangedAndFallsThroughToTheLadder` now fails loudly when that happens rather than silently changing what bare `-S` renders.
- **Width vs count** — row count is gated by `Type_BareSelect_StaysBoundedAtWorstCaseArity` at `System.Func`17`; width is #3616.

**Validation at `c0c60d917`:** full suite 2624 total, 1 failed (`Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on unmodified `main`), 14 skipped. Differential base vs head on the same invocation: byte-identical everywhere outside single-type `type ... -S`, including all of `member`, type listing, `package`, and `library`. `markdownlint` clean.

richlander:
## Integrated `origin/main` — head moved to `991dfe344`

Not a response to a review finding: `main` moved while this sat ready, and the merge required a real code change, so the head moved and needs a fresh round.

**Merge commit `081d7a2b5`** — no conflicts, no manual resolution.

**`991dfe344`** — `main` landed #3319's stderr-ownership rule while this PR was open. `CommandError` is now the single writer of stderr, and `eng/BannedSymbols.txt` bans `Console.Error` with build-time enforcement. The empty-overview guard added in `27bd19be2` predates that rule and used `Console.Error.WriteLine`, so it **would not have compiled** on the new `main`.

Converted to the current idiom:

```csharp
CommandError.Write(
    "this view publishes no bare -S overview sections.",
    "Use -S <Section> to select one, -D to discover what is available, or -S @All for everything.");
```

This is strictly better than what it replaced, not just build-legal: `CommandError.Write` owns the `Error:` severity line and **indents** the detail, so the follow-up guidance is no longer a second unindented line on stderr — which is precisely the property #3319 exists to guarantee.

Also worth noting for anyone reading the check list: this branch was `BLOCKED` before the merge because #3608 made `ci-required` the sole required check and this branch predated it, so no such check existed on it. It now reports and passes.

### Validation at `991dfe344`

- Full suite **2780 total, 1 failed, 14 skipped** — the one failure is `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on unmodified `main`. (Test count rose from 2624; `main` brought ~156 new tests, all passing.)
- `type System.String -S` -> one section, `Type Info`.
- `type --platform System.Private.CoreLib -S` -> exit 0, unchanged fallthrough (the guard correctly does **not** fire).
- `member ... -S` -> exit 0, unchanged.
- CI green, `ci-required` passing, MERGEABLE/CLEAN.

Running one confirmation round on this delta before restoring the ready-to-merge state. The previously clean rounds at `27bd19be2` and `c0c60d917` stand for everything else; nothing in the product logic of this PR changed.

richlander:
## Round 7 — MAI-Code, clean at `991dfe344`

This round covered only the integration delta (`081d7a2b5` merge + `991dfe344`), since rounds 1–6 already cleared `c0c60d917`. The delta exists because merging `origin/main` brought in #3319's stderr-ownership rule, which bans `Console.Error` at build time — the empty-overview guard used it and would not have compiled. `991dfe344` routes it through `CommandError.Write` instead.

**Result: clean. No blocking or non-blocking findings.**

Reviewer's verification, reproduced independently:

- `dotnet build dotnet-inspect.slnx -c Release` — 0 warnings, 0 errors (i.e. the `BannedApiAnalyzers` rule is satisfied).
- Full suite — `Total: 2780, Failed: 1`, the single failure being the known pre-existing `Layout_Count_CountsFilesRatherThanRenderedTreeLines` local-cache issue.
- Focused `-method "*BareSelect*" -method "*Type_FixedOverview_IsExactlyTypeInfo"` — 26/26 pass.
- `type System.String -S` → exit 0, one section, `## Type Info`.
- `type --platform System.Private.CoreLib -S` → exit 0, list mode unchanged (`Classes`/`Structs`/`Interfaces`/`Enums`/`Delegates`).
- `member System.Text.Json.JsonSerializer --platform System.Text.Json -S` → exit 0, Info set unchanged.

Notably the reviewer **mutation-tested the guard** rather than taking its reachability on trust: forcing the fixed-overview set empty and rebuilding produced exit 1 with

```text
Error: this view publishes no bare -S overview sections.
  Use -S <Section> to select one, -D to discover what is available, or -S @All for everything.
```

which confirms the `CommandError` conversion preserved both the `Error:` prefix and the detail indentation that #3319's containment property depends on. The worktree was restored clean afterward.

### Review ledger (final)

| --- | --- | --- | --- |
| 1 | GPT-5.6 | `14f28f780` | 1 blocking — withdrawn by the reviewer |
| 2 | Gemini 3.1 Pro | `14f28f780` | 2 blocking — both fixed in `27bd19be2` |
| 3 | Gemini 3.1 Pro | `27bd19be2` | Clean |
| 4 | GPT-5.6 | `27bd19be2` | 1 non-blocking (docs) — fixed in `d641e7a42` |
| 5 | GPT-5.6 | `d641e7a42` | 1 non-blocking (docs) — fixed in `c0c60d917` |
| 6 | GPT-5.6 | `c0c60d917` | Clean |
| 7 | MAI-Code | `991dfe344` | Clean (integration delta) |

Two model families, three clean rounds, all findings resolved on the PR. CI is green at `991dfe344` (`ci-required` passing) and the PR is MERGEABLE/CLEAN.

Ready to merge

PR #3615, opened 2026-07-31
Derive local declaration placement from portable-PDB scopes (#3591)

### LINKED ISSUE #3591: Decompiler discards portable-PDB LocalScope ranges, hoisting narrowly-scoped locals to the method top

richlander:
Closes #3591.

The portable PDB's `LocalScope` table records each local's true declaration extent. `MetadataSource` read those rows for the variable *names* and threw `StartOffset`/`EndOffset` away, so a local the source declared inside a `try`, `using`, or loop was hoisted to the top of the method. #3593 pinned that gap and the evidence that closes it; this closes it.

**The headline is fidelity, not style.** The hoisted spelling was not merely uglier — it recompiled to *different IL*. Declaration order drives local-slot allocation, so the hoisted render swapped slot numbering against the original. The sunk render is byte-exact. Measured below.

**Benchmark target:** `Azure.Data.Tables@12.11.0`

**dotnet-inspect command:**

```bash
dotnet-inspect member Azure.Data.Tables.TableClient --package Azure.Data.Tables@12.11.0 -m Create -S "Decompiled Source"
```

## Original source

`-S "Original Source"`, SourceLink-backed.

```csharp
{
    scope.Start();
    try
    {
            null,
            _defaultQueryOptions,
            cancellationToken);
    }
    catch (Exception ex)
    {
        scope.Failed(ex);
        throw;
    }
}
```

## Before

```csharp
{

    {
        scope.Start();
        try
        {
            return Response.FromValue<TableItem>(response.get_Value(), response.GetRawResponse());
        }
        catch (Exception ex)
        {
            scope.Failed(ex);
            throw;
        }
    }
}
```

- Valid: True
- Correct: True
- IL fidelity: **False** — see the round-trip measurement below. The declaration order the hoist imposes reallocates local slots.
- Taste applied: none (`-S "Applied Taste"`: *No recorded style choices were applied to this member.*)
- Commit: `0e4c303ce`

## After

```csharp
{
    {
        scope.Start();
        try
        {
            ResponseWithHeaders<TableResponse, TableCreateHeaders> response = _tableOperations.Create(new TableProperties { TableName = Name }, default(Nullable<ResponseFormat>), _defaultQueryOptions, cancellationToken);
            return Response.FromValue<TableItem>(response.get_Value(), response.GetRawResponse());
        }
        catch (Exception ex)
        {
            scope.Failed(ex);
            throw;
        }
    }
}
```

- Valid: True
- Correct: True
- IL fidelity: **True** for the declaration placement itself — a minimal reproduction of this exact shape recompiles byte-for-byte after the change and did not before.
- Taste applied: none. This is derived from evidence, not chosen; there is no knob, matching the `## Names` precedent in `docs/decompiler-taste.md`.
- Commit: `d069dfdc4`

## Fully raised

The remaining deltas against Original source are all tracked separately and untouched here: `response.get_Value()` vs `.Value` (#2947), `default(Nullable<ResponseFormat>)` vs `null` (#3027), statement-`using` vs block-`using` (deliberate non-change; declared oracle `.editorconfig:92` `csharp_prefer_simple_using_statement = false:none`), and `var` spelling (#3169 family).

## Why this is a raising gap and not taste

Under the rule that we do not implement taste when the correct answer is PDB-knowable, this was never a style question. IL alone cannot express declaration position — `.locals init` is a flat slot list — but the PDB records it exactly, and dotnet-inspect had already fetched the PDB for this very package. The `response`, `scope`, and `ex` names in the Before render are themselves PDB-derived; only the scope ranges were dropped.

`docs/decompiler-taste.md:630` (`## Names`) sets the split this follows: PDB-present facts are simply *used*, with no knob; only *synthesizing* where no PDB exists is an opt-in knob. Declaration placement maps onto that exactly.

## Design

- `MetadataSource.LocalDeclarations` replaces `LocalNames` and returns names **and** scopes from one walk of the same `LocalScope` rows. Splitting them would traverse the table twice per method and let the two views disagree about which entries were skipped.
- `MethodImporter` reduces each raw scope to the one fact the IR needs — is this slot scoped narrower than the whole body — because mapping IL offsets onto IR nodes *after* passes rewrite the tree would be fragile. Metadata owns metadata facts; the importer converts them to an IR fact.
- `CSharpPrinter.CollectDeclaringStores` gains a branch that merges the declaration into its store when the flag is set.

**The safety split is the load-bearing part.** The PDB scope is evidence of *intent*; it is never evidence of *validity*. The branch only fires when the IR independently proves every reference to the local stays inside the store's block at or after the store. Both are required, and each has a dedicated negative test.

`LocalReferencesStayInBlockAfterStatement` could not serve as that proof: it walks *top-level* statements and requires each referencing one to be a **descendant** of an allowed statement. For a store nested inside a `using` or `try` the top-level statement is an **ancestor** of the reference instead, so it can never say yes. The new `LocalReferencesStayInsideStoreBlock` walks the reference nodes themselves, which is orientation-free, reusing `IrFunction`'s gate-enforced local-slot binding list rather than duplicating it.

## Evidence

Round-trip on a minimal reproduction of the witness shape (original compiled from the *sunk* source, decompiled with each build, recompiled, IL compared):

| Render | Recompiled IL identical to original |
| --- | --- |
| Before (`0e4c303ce`) | **False** — `stloc.1`/`ldloc.0` vs `stloc.0`/`ldloc.1` at 6 offsets |
| After (`ff695a216`) | **True** |

Both renders are Valid and Correct; only the After is opcode-faithful. This is the #3127 trap in reverse — the Before block's two `True`s were not enough.

| Gate | Result |
| --- | --- |
| `--gate fast` | 4409 / 0 failed / 0 skipped |
| `-trait "Area=RoundTrip"` (incl. slow compile-back) | 568 / 0 |
| `-trait "Area=Fidelity"` (slow, 32 min) | 68 / 0 |
| `-trait "Area=Corpus"` | 348 / 0 |
| `-trait "Area=Validity"` | 103 / 0 |
| `dotnet run --project src/dotnet-inspect.Tests -c Release` | 2602 / 1 failed / 14 skipped |

The single CLI failure is `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, pre-existing and reproduced byte-identically on a clean detached `origin/main` worktree.

### Adversarial shapes

Neither reviewer built counterexamples for the soundness claim, so I did. 15 real C# shapes compiled with a portable PDB, decompiled at this head, and the 13 compilable renders recompiled and executed against the original:

| Outcome | Shapes |
| --- | --- |
| Sank correctly, matching the original source | `LoopLocal`, `ForEachNested`, `PatternVar`, `GotoOut` |
| Correctly **declined** to sink | `CarryOut` (read after the loop), `BackEdge` (read before the store across a back-edge), `FinallyReads` (the `finally` reads it), `NestedEh` (the outer `catch` reads it), `TwoArms` (store in one `if` arm, read after), `OutArg`, `SiblingRead` |
| Unchanged and correct | `RefLocal` (ref local in a nested block), `SwitchSections` |
| Documented residual | `LambdaCapture`, `LocalFuncCapture` (display-class shapes) |

Behavior comparison, original assembly vs recompiled decompiled output, 22 invocations with fresh state per call: **22 identical, 0 different.**

### Blast radius

Across 4409 decompiler tests only the two #3593 named as having to flip changed behavior. That is small because most decompiler tests build IR by hand and carry no PDB flags — it is **not** evidence that the real-world reach is small. Measured over the 80 real portable PDBs in the local cache: 207,977 methods, 217,304 `LocalScope` rows, 92,269 local variables, of which **41,818 scopes start after IL offset 0**. Narrow scopes are pervasive in real code, so this changes a great deal of real output — which is what the corpus, fidelity, round-trip and validity gates are for.

That same sweep is the evidence for narrowing the `catch`: across all 207,977 methods, replaying `LocalDeclarations`' exact SRM call sequence threw **no exception of any type**, so restricting the catch to `BadImageFormatException` cannot regress well-formed input.

### Non-vacuity

Each half of the rule was tampered and the expected tests failed:

| Tamper | Failing tests |
| --- | --- |
| `IsLocalDeclaredInNestedScope` forced false | the 3 sinking positives |
| containment check removed from `LocalReferencesStayInsideStoreBlock` | `BranchedStore_StaysHoisted_DespiteNestedPdbScope` |

### Tests

`DeclarationPlacementFrontierTests` is renamed to `DeclarationPlacementTests` — it no longer pins a frontier. #3593's `<remarks>` named the two tests that closing this had to flip, and both flipped.

| Test | Claim |
| --- | --- |
| `NarrowLocal_DeclarationSinksToItsStore_WithPdb` | flipped: declaration merges into its store inside the `try` |
| `NarrowLocal_PlacementDiffersWithAndWithoutPdb` | flipped: masking local names no longer collapses the two renderings |
| `NarrowLocal_StaysHoisted_WithoutPdb` | **negative for the PDB half**: no evidence, no guess — output byte-identical to before |
| `BranchedStore_StaysHoisted_DespiteNestedPdbScope` | **negative for the IR half**: scope is equally narrow, but the store sits in one arm of an `if` and the read is after it, so sinking would not compile — the printer declines |
| `LoopBodyLocal_DeclarationSinksToItsStore` | new positive: the commonest form of the same evidence |
| `HoistedLocal_RoundTripsAtMethodTop` | control, unchanged |
| `PdbScopes_DistinguishTheNarrowAndHoistedShapes` | non-vacuity of the fixture pair |
| `PdbScopes_OmitCompilerTemps` | a compiler temp has no scope row at all |

## Style oracle

Both facets consulted, per `docs/decompiler-taste.md`.

- **Declared** — `dotnet/runtime` `.editorconfig` is **silent** on general declaration placement. L124 `csharp_style_inlined_variable_declaration = true:suggestion` covers `out` parameters only. (L52-54 `csharp_style_var_* = false` is why the render spells `ResponseWithHeaders<...> response` rather than `var`; that is #3169's business, not this PR's.)
- **Revealed** — `dotnet/runtime` declares at first use when use is confined to the block, and hoists only when a `catch`/`finally` or later code needs it: `HttpContent c = response.Content;` (`HttpClient.cs:199`) and `StringBuilder sb = new StringBuilder();` (`File.cs:1132`) against `HttpResponseMessage? response = null;` (`HttpClient.cs:189`) and `char[]? buffer = null;` (`File.cs:1126`). That is the narrowest scope covering all uses — exactly what the PDB encodes and what this PR emits.

The revealed facet agrees, but the claim here rests on the PDB, not on the oracle.

## Non-action boundary

- **No PDB, no change.** The flag is empty and output is byte-identical to before. Nothing is inferred from IL alone.
- **Lambdas and local functions.** `CSharpPrinter` builds sub-`IrFunction`s from `lambda.LocalNames` and does not carry the new flags, so locals inside them keep the hoisted shape. Not a regression; not addressed here.
- **Partial sinking is not attempted.** When the IR guard declines (`CreateBranched`), the declaration stays where it is today — at method top — rather than moving to the enclosing block the PDB names. Sinking to an intermediate block is a strictly larger change and is left out.
- **`ResetLocals` drops the flags.** A renumbering no longer names the same locals, so keeping scope evidence would misattribute it. The printer degrades to the byte-stable method-scope shape.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
+++ b/src/ILInspector.Decompiler.Tests/CfgSampleClass.cs
@@ -6493,3 +6493,38 @@ public sealed class DeclScopeClient
         }
     }
 }
+
+public sealed class DeclScopeLoopClient
+{
+    readonly DeclScopeOps _ops = new();
+
+    // The source declares the local inside the loop body and every use stays there,
+    // so the PDB scope and the IR agree and the declaration sinks to its store.
+    public int SumNarrow(int count)
+    {
+        int total = 0;
+        for (int i = 0; i < count; i++)
+        {
+            DeclScopeResult step = _ops.Create("s", i);
+            total += step.Raw + step.Value.Length;
+        }
+        return total;
+    }
+
+    // Close negative for the same PDB evidence. The scope is again nested (the using
+    // block), but the first store sits inside one arm of the if while the read is
+    // after it, so sinking the declaration onto that store would not compile. The IR
+    // guard must decline and leave the hoisted declaration alone.
+    public string CreateBranched(string name, int timeout, bool flag)
+    {
+        using (DeclScopeGuard scope = new DeclScopeGuard())
+        {
+            DeclScopeResult response;
+            if (flag)
+                response = _ops.Create(name, timeout);
+            else
+                response = _ops.Create(name, timeout + 1);
+            return response.Value + response.Raw;
+        }
+    }
+}
--- a/src/ILInspector.Decompiler.Tests/DeclarationPlacementTests.cs
+++ b/src/ILInspector.Decompiler.Tests/DeclarationPlacementTests.cs
@@ -9,65 +9,110 @@ namespace ILInspector.Decompiler.Tests;
 /// <summary>
 /// Declaration placement for a local the source declared inside a nested block
 /// (#3591). The portable PDB's <c>LocalScope</c> table records each local's true
-/// extent, but <c>MetadataSource.LocalNames</c> keeps only the variable names and
-/// drops <c>StartOffset</c>/<c>EndOffset</c>, so the printer hoists a narrowly
-/// scoped local to the top of the method. These tests pin that frontier and, just
-/// as importantly, pin the evidence that would close it: the PDB already
-/// distinguishes the two shapes that today print identically.
+/// extent; <c>MetadataSource.LocalScopes</c> reads it and the printer sinks the
+/// declaration onto its store when the IR independently proves that is valid.
 /// </summary>
 /// <remarks>
-/// Closing #3591 must flip <see cref="NarrowLocal_IsHoistedAboveTheUsing_WithPdb"/>
-/// and <see cref="NarrowLocal_PlacementIsIdenticalWithAndWithoutPdb"/>: with a PDB
-/// the declaration should merge into its store inside the try, which is exactly the
-/// masked equality this class currently asserts. The remaining tests are invariants
-/// and should keep passing.
+/// The two halves of that rule each have a dedicated negative here. Without a PDB
+/// there is no evidence of intent, so
+/// <see cref="NarrowLocal_StaysHoisted_WithoutPdb"/> pins the unchanged hoisted
+/// shape rather than a guess. With a PDB but a store the IR cannot prove dominates
+/// its uses, <see cref="BranchedStore_StaysHoisted_DespiteNestedPdbScope"/> pins the
+/// decline. This class was <c>DeclarationPlacementFrontierTests</c> in #3593, where
+/// it pinned the gap and named the two tests that closing it had to flip.
 /// </remarks>
-public class DeclarationPlacementFrontierTests
+public class DeclarationPlacementTests
 {
     static readonly IAssemblyReferenceResolver RuntimeResolver = TestAssemblyReferenceResolvers.TrustedPlatformAssemblies();
 
     /// <summary>
-    /// The PDB names a local declared in the try block. The scope range that comes
-    /// with the name is what the printer never reads.
+    /// The PDB scopes the local to the try block, and every reference stays in that
+    /// block after the store, so the declaration merges into the store.
     /// </summary>
     [Fact]
-    public void NarrowLocal_IsHoistedAboveTheUsing_WithPdb()
+    public void NarrowLocal_DeclarationSinksToItsStore_WithPdb()
     {
         var output = Print(nameof(DeclScopeClient.CreateNarrow));
 
-        Assert.Contains("DeclScopeResult response;", output);
-        Assert.Contains("response = _ops.Create(name, timeout);", output);
-        AssertDeclarationPrecedesUsing(output, "DeclScopeResult response;");
+        Assert.Contains("DeclScopeResult response = _ops.Create(name, timeout);", output);
+        Assert.DoesNotContain("DeclScopeResult response;", output);
+        Assert.True(
+            output.IndexOf("using (", StringComparison.Ordinal)
+                < output.IndexOf("DeclScopeResult response =", StringComparison.Ordinal),
+            $"Expected the declaration inside the using.{Environment.NewLine}{output}");
     }
 
     /// <summary>
-    /// Same method with symbols withheld. The local loses its source name, and
-    /// nothing else moves.
+    /// Non-vacuity gate for the PDB evidence. With symbols withheld the same IR
+    /// reaches the printer with no scope facts, and the local keeps its hoisted
+    /// declaration: absent evidence the tool does not guess.
     /// </summary>
     [Fact]
-    public void NarrowLocal_IsHoistedAboveTheUsing_WithoutPdb()
+    public void NarrowLocal_StaysHoisted_WithoutPdb()
     {
         var output = PrintWithoutSymbols(nameof(DeclScopeClient.CreateNarrow));
 
         var declaration = Assert.Single(Regex.Matches(output, @"DeclScopeResult V_\d+;")).Value;
         Assert.DoesNotContain("response", output);
-        AssertDeclarationPrecedesUsing(output, declaration);
+        int declarationIndex = output.IndexOf(declaration, StringComparison.Ordinal);
+        int usingIndex = output.IndexOf("using (", StringComparison.Ordinal);
+        Assert.True(usingIndex >= 0, $"Missing using statement.{Environment.NewLine}{output}");
+        Assert.True(
+            declarationIndex < usingIndex,
+            $"Expected `{declaration}` hoisted above the using.{Environment.NewLine}{output}");
     }
 
     /// <summary>
-    /// The headline claim of #3591: a present PDB changes the local's *name* and
-    /// nothing about where it is declared. Masking every local identifier makes the
-    /// two renderings byte-identical, which is only possible because the scope
-    /// ranges are discarded.
+    /// The headline claim of #3591, inverted. A present PDB used to change the
+    /// local's name and nothing about where it is declared — masking every local
+    /// identifier made the two renderings byte-identical. Now placement differs too,
+    /// and masking no longer collapses them.
     /// </summary>
     [Fact]
-    public void NarrowLocal_PlacementIsIdenticalWithAndWithoutPdb()
+    public void NarrowLocal_PlacementDiffersWithAndWithoutPdb()
     {
         var withPdb = Print(nameof(DeclScopeClient.CreateNarrow));
         var withoutPdb = PrintWithoutSymbols(nameof(DeclScopeClient.CreateNarrow));
 
-        Assert.NotEqual(withPdb, withoutPdb);
-        Assert.Equal(MaskLocalNames(withPdb), MaskLocalNames(withoutPdb));
+        Assert.NotEqual(MaskLocalNames(withPdb), MaskLocalNames(withoutPdb));
+    }
+
+    /// <summary>
+    /// A loop body is the same evidence in its commonest form: the source declared
+    /// the local inside the body, so it belongs at its store rather than above the
+    /// loop where every iteration would appear to share it.
+    /// </summary>
+    [Fact]
+    public void LoopBodyLocal_DeclarationSinksToItsStore()
+    {
+        var output = Print<DeclScopeLoopClient>(nameof(DeclScopeLoopClient.SumNarrow));
+
+        Assert.Contains("DeclScopeResult step = _ops.Create(\"s\", i);", output);
+        Assert.DoesNotContain("DeclScopeResult step;", output);
+    }
+
+    /// <summary>
+    /// Non-vacuity gate for the IR guard. The PDB scopes this local to the using
+    /// block just as narrowly, but the first store is inside one arm of the if while
+    /// the read is after it, so sinking the declaration onto that store would not
+    /// compile. The scope is evidence of intent, not of validity, and the printer
+    /// declines.
+    /// </summary>
+    [Fact]
+    public void BranchedStore_StaysHoisted_DespiteNestedPdbScope()
+    {
+        var scope = PdbScope<DeclScopeLoopClient>(nameof(DeclScopeLoopClient.CreateBranched), "response");
+        Assert.True(
+            scope.Start > 0 || scope.End < scope.IlLength,
+            $"Expected a nested scope, got IL_{scope.Start:X4}..IL_{scope.End:X4} of 0x{scope.IlLength:X4}.");
+
+        var output = Print<DeclScopeLoopClient>(nameof(DeclScopeLoopClient.CreateBranched));
+
+        Assert.Contains("DeclScopeResult response;", output);
+        Assert.True(
+            output.IndexOf("DeclScopeResult response;", StringComparison.Ordinal)
+                < output.IndexOf("using (", StringComparison.Ordinal),
+            $"Expected the declaration to stay above the using.{Environment.NewLine}{output}");
     }
 
     /// <summary>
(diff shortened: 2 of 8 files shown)


richlander:
## Adversarial review

Two models, two rounds, isolated detached worktrees, scratch under `/tmp/`, no writes to the trees.

| Reviewer | Round 1 head `ff695a216` | Round 2 head `d069dfdc4` |
| --- | --- | --- |
| Gemini 3.1 Pro | **BLOCKING** — 1 finding | **CLEAR** |
| GPT-5.6 Sol | **CLEAR** (with a stated limitation) | **CLEAR** |

### Finding 1 — partial scope array on PDB decode failure (Gemini, Blocking) — FIXED in `d069dfdc4`

`MetadataSource.LocalDeclarations` inherited `LocalNames`' bare `catch`, which returned whatever had been read so far. That was tolerable when the walk only produced names; it is not tolerable now that it also produces scopes, because scopes drive **where a declaration is printed**. A partial array would make output shape a function of where the corruption stopped. GPT independently identified the same mechanism and rated it non-blocking; I agree with Gemini that the mechanism is the point and fixed it.

The fix does two things:

1. Scopes are dropped **wholesale** on failure (`scopesUsable = false`), which yields exactly the documented no-PDB behavior: no evidence, no sinking, byte-stable output. Names deliberately stay partial — a missing name degrades visibly to `V_index`, and that is the pre-existing contract.
2. The `catch` is narrowed to `BadImageFormatException`, so anything that is not a decode failure propagates instead of being swallowed.

**GPT flagged the narrowing itself as the main risk of the fix** — if a malformed PDB can surface some *other* exception type here, narrowing converts a silent degradation into a crash. I measured it rather than argued it: replaying `LocalDeclarations`' exact SRM call sequence over every method of the 80 real portable PDBs in the local cache — 207,977 methods, 217,304 `LocalScope` rows, 92,269 local variables — threw **no exception of any type**. GPT then confirmed against SRM sources that malformed table bounds, invalid string offsets, and `EndOffset` overflow all surface as `BadImageFormatException`, and that other types indicate programming or lifetime failures that should propagate.

The fallback is marked **explicitly unverified** in the `<remarks>`, per AGENTS.md's rule that "unverified" is acceptable but an unmarked ungated claim is not. It is unreachable from any fixture this repo can build, measured: a truncated portable PDB throws at `MetadataReaderProvider.FromPortablePdbStream`, i.e. at open, handled by `PdbReader()`'s own pre-existing catch which disables symbols entirely; and a PDB belonging to a different assembly returns empty scopes without throwing (18,242 method handles, 0 throws). Reaching this catch needs a PDB that opens cleanly and then fails mid-walk. Both reviewers accepted the marking as honest; GPT noted a byte-mutated PDB fixture could gate it later, which I agree is worthwhile follow-up and is **non-blocking**.

### Gap in the reviews themselves, and what I did about it

Both reviewers reached attack point 1 — the soundness of the sinking predicate, which is the load-bearing claim of this PR — by **code reading only**. GPT said so explicitly and unprompted when asked to substantiate, and also disclosed that it had not reproduced the headline IL-fidelity measurement. That is the right disclosure and I am recording it rather than letting the CLEARs imply execution that did not happen.

So I built the counterexamples myself: 15 real C# shapes compiled with a portable PDB, decompiled at this head, 13 compilable renders recompiled and executed against the original. Four sank correctly; **seven correctly declined** — `CarryOut`, `BackEdge`, `FinallyReads`, `NestedEh`, `TwoArms`, `OutArg`, `SiblingRead`; `RefLocal` and `SwitchSections` were unchanged and correct; the two display-class shapes are the documented residual. Behavior comparison across 22 invocations with fresh state per call: **22 identical, 0 different.** Full table in the PR description.

Both reviewers assessed that evidence in round 2. GPT: "materially supports the containment soundness claim. No critical missing shape identified." Gemini worked three further edge cases independently — `goto` into a nested scope, `out` parameters, and `do`/`while` condition placement — and concluded the predicate is conservative in each: "when in doubt, they decline to sink."

### Verified, not just asserted

`GotoOut` renders an unreachable trailing `return "done";`. I checked this against the base build before attributing it: it is present in **both** renders, so it is a pre-existing structuring artifact and not caused by this PR.

One behavioral difference did appear in my first comparison run (`RefLocal`: 15 vs 20). It was a bug in **my harness** — `object[].Clone()` is shallow, so the first call mutated the `int[]` the second call saw. Fixed by constructing fresh state per invocation; the shape is correct.

### Real-run evidence at the fixed head `d069dfdc4`

| Gate | Author | Gemini | GPT |
| --- | --- | --- | --- |
| `dotnet build dotnet-inspect.slnx -c Release` | clean | clean | clean |
| `--gate fast` | 4409 / 0 | 4409 / 0 | 4409 / 0 |
| `DeclarationPlacementTests` | 8 / 0 | 8 / 0 | — |
| `-trait "Area=RoundTrip"` | 568 / 0 | — | 568 / 0 (at `ff695a216`) |
| `-trait "Area=Corpus"` | 348 / 0 | — | 348 / 0 (at `ff695a216`) |
| `-trait "Area=Validity"` | 103 / 0 | — | 103 / 0 (at `ff695a216`) |
| `-trait "Area=Fidelity"` | 68 / 0 (~32 min, at `ff695a216`) | — | 68 / 0 (1573s, at `ff695a216`) |
| `dotnet-inspect.Tests` | 2602 / 1 / 14 | — | — |

The slow gates were run at `ff695a216`. The delta since is confined to an unreachable `catch` in `MetadataSource`; the decompiler logic is byte-identical between the two heads, and I re-ran `--gate fast`, `Area=RoundTrip` and `Area=Corpus` at `d069dfdc4` to confirm. The one CLI failure is `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, pre-existing and reproduced on a clean detached `origin/main` worktree.

### Non-blocking follow-up

- Gate the decode fallback with a byte-mutated portable PDB fixture (GPT's suggestion).
- Carry the scope flags into lambda and local-function sub-`IrFunction`s, which today keep the hoisted shape.
- Sink a declined store to the enclosing block the PDB names, rather than leaving it at method top.

Ready to merge

richlander:
CI fully green at `d069dfdc4`, including the slow `decompiler-gates` job (14m53s) and `test` (9m45s). `build-net10`, `csharp-diff-smoke`, `changes` pass; `il-diff-smoke`, `markdownlint`, `pack` path-gated skips.

Review worktrees removed. Still `Ready to merge`.

richlander:
Merged. The three declared non-blocking follow-ups are now filed:

- #3621 — gate the PDB scope-table decode fallback with a byte-mutated portable-PDB fixture (retires the `unverified` marker on `LocalDeclarations`)
- #3622 — carry the declaration-scope flags into lambda / local-function `IrFunction`s
- #3623 — sink a *declined* declaration to the enclosing PDB scope rather than method top (`DeclScopeLoopClient.CreateBranched` is the witness)

Issue #3610, opened 2026-07-31
Caret underline is unclamped: 9.84% of caret lines exceed the 100-column budget, worst is 57,943 columns


richlander:
## What

`AnnotationCaret.Budget` is 100 columns, and the detail text is wrapped to respect it. The `^^^^` underline itself is not clamped: it is drawn at `extent.Length` characters, whatever that is.

When a fact narrows to an expression this is usually fine. When it falls back to the statement — because the facts on the line disagree, or no fact has a printed node — the underline is as wide as the statement, and a printed statement has no width limit.

## Measured

`System.Private.CoreLib` (11.0.0-preview.7.26366.102), all 31,641 rendered caret lines:

| | caret lines > 100 columns | worst |
|---|---|---|
| with expression narrowing (#3570 slice 1b) | 3,115 (9.84%) | 57,943 columns |
| without | 3,527 (11.15%) | 57,943 columns |

So narrowing removes 412 of them and does not touch the worst case, because the worst case is exactly the one narrowing declines: a long statement whose facts disagree.

## Reproducer

`System.Tuple`8.Equals` carries 16 facts on one line — each `comparer.Equals(m_ItemN, V_0.m_ItemN)` boxes both operands, so 8 items x 2 boxes:

```
return comparer.Equals(m_Item1, V_0.m_Item1) && ... && comparer.Equals(m_Rest, V_0.m_Rest);
// ^^^^^^^ ...313 carets... ^^^^^^^
//   alloc.box(T1; alloc=boxed T1; path=branch; path-confidence=behind-branch; ...)
//   alloc.box(T1; ...)
//   ... 14 more
```

## Why it is not fixed in slice 1b

It predates that change and is orthogonal to it: slice 1b decides *what* the caret points at, not how wide a caret may be drawn. Clamping is a layout decision with its own question — a truncated underline has to signal that it was truncated, or the reader will believe the fact stops where the carets do. Worth deciding on its own rather than folding into a placement fix.

## Options to weigh

- Clamp to the budget with a continuation marker (`^^^^...` / `^^^^>`), so truncation is visible.
- Clamp to the budget and move the detail to a block underneath unconditionally past some width.
- Leave long statements uncarreted and rely on the detail block alone.

Related: #3570.

PR #3617, opened 2026-07-31
Point the --focus caret at the expression a fact is about

richlander:
Slice 1b of #3570. #3574 fixed *which statement* a fact anchors to; this fixes *what the caret underlines*.

## Before / after

`--focus allocation` on a fixture that allocates at two depths:

**Before** — the caret spans the whole statement, so it says "something on this line allocates":

```csharp
    List<object> sink = new();
//  ^^^^^^^^^^^^^^^^^^^^^^^^^^ alloc.new(List<object>; ...)
        sink.Add(new object());
//      ^^^^^^^^^^^^^^^^^^^^^^^ alloc.new(object; alloc=System.Object; path=loop-body; ...)
```

**After** — it names the allocation:

```csharp
    List<object> sink = new();
//                      ^^^^^ alloc.new(List<object>; ...)
        sink.Add(new object());
//               ^^^^^^^^^^^^ alloc.new(object; alloc=System.Object; path=loop-body; ...)
```

## The density case

The interesting demonstration is the opposite one. `System.Tuple`8.Equals` carries **16 facts on a single line** — each `comparer.Equals(m_ItemN, V_0.m_ItemN)` boxes *both* operands, so 8 items x 2. Reproduced in-repo as `CommandCaretGestureFixture.DenseBoxes` and pinned by a test:

```csharp
    return comparer.Equals(a1, b1) && comparer.Equals(a2, b2) && comparer.Equals(a3, b3) && comparer.Equals(a4, b4);
//  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//   alloc.box(T1; alloc=boxed T1; path=straight-line; path-confidence=dominates-return;
//   post-dominance=return-post-dominates; multiplicity=once)
//   alloc.box(T2; alloc=boxed T2; path=branch; path-confidence=behind-branch;
//   post-dominance=return-post-dominates; multiplicity=conditional)
//   alloc.box(T3; ...)
//   alloc.box(T4; ...)
```

One underline, four facts, each with its own detail rows. The four boxes are at four distinct offsets, so no single expression is true of all of them and the statement is the honest answer — it is exactly the smallest span true of every fact on the line.

## What it does

`AnnotationAnchor.ComputeCaretExtents` recovers a character span for each fact from the printed tree: the printed node carrying that offset exactly. `AnnotationCaret` uses it when every fact on the line agrees, and falls back to the statement otherwise.

## Measured — `System.Private.CoreLib` 11.0.0-preview.7.26366.102

37,866 facts across 41,952 imported methods — 40,917 of which print a body and 18,538 of which carry at least one fact — measured **as the annotated-source view prints them** — that is, with callee bodies imported, which is the render `--focus` produces.

| | |
|---|---|
| facts with an exact-offset printed node | 33,943 of 37,800 |
| ...that survive to a delivered extent | **33,657 (89.04%)** — 283 lost to a continuation line, 3 not line-positionable |
| ...none of which span a line break | 0 of 33,657 |
| caret-bearing lines that narrow | 25,628 / 31,640 (**81.00%**) |
| lines whose facts disagree | 3,385 (10.70%) |
| lines with no fact carrying a printed node | 2,627 (8.30%) |
| narrowed extents strictly inside their statement | 25,172 (98.22%) |
| invariant violations (outside statement, or underlining whitespace) | **0** |

The render matters, and round 6 is where that was established. The C#-only overlay prints the same members *without* importing callee bodies; a few printed ranges shift, and the same probe returns **33,729** there. Both are production paths and each computes extents against its own printed body, so neither is wrong — but a figure is only meaningful against a stated render, because extents are measured in printed characters. Every number in this PR and in the shipped XML is now the annotated-source figure, and says so.

Be plain about the shape of that headline: it is carried by the common case.

| facts on line | lines | narrowed |
|---|---|---|
| 1 | 27,414 | 92.1% |
| 2 | 2,987 | 12.1% |
| 3 | 852 | 1.1% |
| 4-5 | 340 | 0% |
| 6-9 | 42 | 0% |
| 10+ | 5 | 0% |

Density and disagreement are the same phenomenon — more facts on a line means more distinct offsets on it. And the width win is modest: 1,320,017 underlined characters against 1,626,528 statement characters, 81.2%. The value here is the **correct start column**, not a dramatic narrowing.

## Two rules, both forced by the corpus rather than chosen

**Narrowing is all-or-nothing per line.** A row per distinct extent was built first, then rejected: of 5,816 extent pairs sharing a line, **3,325 nest or overlap**, so the rows would re-underline the same characters at different widths and the reader would have to count columns to tell which row named which fact. The worst line in CoreLib carries eight distinct extents. The rejection and its evidence are recorded in `AnnotationCaret.Agreed` so it is not re-litigated.

**Width is only the tie-break among real expressions.** A `LoadStackSlot` prints as `S_n` — a stand-in the raise emits where it could not recover a value — and it carries the *producing* instruction's offset. It is narrower than the `new` it sits inside, so a width-only rule underlines the argument:

```
return new ConstructorInvoker(S_0);
//                            ^^^ alloc.new(ConstructorInvoker; ...)
```

Preferring real expressions moves 50 of 10,646 `alloc.new` underlines onto the allocation: **99.52% -> 99.99%** of them now begin with `new` (51 wrong of 10,659 with the rule disabled, 1 wrong of 10,646 with it on). The single remaining case is a printer `/* Unsupported IL ... */` comment that *is* the whole statement, so it is not a mis-selection.

Two further guards, both found by corpus sweeps rather than by reading:

- An extent is produced only when the narrow node prints on the **owner statement's line** (283 facts otherwise). Narrowing across lines would underline a line the fact is not attached to.
- Extents are **trimmed onto printed characters**. A statement that is its own narrowest node has a printed range starting at the line's indent, which drew carets left of the code on 205 extents. Sweep now reports 0 extents starting or ending on whitespace.

## Tests: rewritten to the new contract, not relaxed

Six existing caret tests asserted `caret column == statement indent` and `caret count == statement length` — they pinned the old contract exactly. Loosening them would have forfeited the gate, so they assert stronger properties instead:

- the underline is **contained** in the statement it points at;
- it **starts and ends on non-whitespace**;
- **non-vacuity**: every examined caret is *strictly narrower* than its statement.

Plus new tests: `Member_AnnotatedSource_FocusWidensToTheStatementWhenFactsOnALineDisagree` (the density case above, pinning the fallback branch), `CaretExtent_PrefersTheAllocation_OverAStandInSharingItsOffset` (a `[Theory]` over both stand-in kinds, on a synthetic `return new Holder(S_0);`), `CaretExtent_TrimsAStatementWideExtentOffTheLineIndent`, and two direct `[Theory]` gates on `TryTrimToPrinted` itself — 6 shrink cases and 5 rejection cases. **`AnnotationAnchorTests` is now 13 test methods / 25 executed cases, all green.**

`Member_AnnotatedSource_FocusPromotesFactsToAlignedCaretComments` now pins the underlined **text** — `new()` and `new object()` — rather than a width, which is what makes it a placement gate.

### Mutation-verified

Every gate below was mutated, rebuilt, and re-run. **No mutation broke the build** — which is the whole point: each of these would have shipped silently.

| mutation | result |
|---|---|
| `Render(..., extents)` -> `Render(..., null)` | **6 of 7 fail**; the density test correctly still passes, because it gates the fallback branch |
| `printed.Node is LoadStackSlot or CaughtException` -> `false` | stand-in `[Theory]` **fails both cases** |
| drop `CaughtException` from that pattern | the `__exception` case **fails** |
| delete the **leading** whitespace loop | **3 fail**, incl. `CaretExtent_TrimsAStatementWideExtentOffTheLineIndent` |
| delete the **trailing** whitespace loop | **exactly the 2 trailing-padding cases fail** |
| delete the `Math.Min` clamp | **exactly the overhang case fails** |
| clamp to `lineText.Length - 1` | **4 fail**, incl. both exact-end cases |
| delete the `line != ownerLine` filter | `CaretExtent_IsWithheldWhenTheNarrowNodePrintsOnAContinuationLine` **fails** |
| `char.IsWhiteSpace` -> `== ' '` | **exactly the tab and NBSP cases fail** |
| ignore `TryTrimToPrinted`'s return value | `CaretExtent_IsWithheldWhenTheTrimRejectsTheRange` **fails** |
| ignore `TryGetLineColumn`'s failure | **no test fails — equivalent mutant**, corpus output byte-identical |
| drop the **late stand-in** ordering guard | `CaretExtent_KeepsTheRealExpression_WhenTheStandInIsRecordedAfterIt` **fails** (8 corpus extents move) |
| `best <= width` -> `best < width` (equal-width tie) | `CaretExtent_KeepsTheFirstOfTwoEqualWidthCandidates` **fails** (96 corpus extents move) |
| `Agreed`, **missing** entry: `return null` -> `continue` | `CaretWidensWhenOnlySomeFactsOnTheLineHaveAnExtent` **fails** (annotated render: **499 of 615** mixed lines would narrow wrongly; overlay: 498 of 609) |
| `Agreed`, **disagreeing** entries: `seen != extent` -> `false` | `CaretWidensWhenTwoFactsOnTheLineDisagreeAboutTheExtent` **fails** |
| `caretColumn = extent.Column + hoisted` -> `hoisted` | **6 fail** — already gated before round 5; reported as a gap, did not reproduce |

All restored and rebuilt afterwards.

## Suites

| suite | result |
|---|---|
| `ILInspector.Decompiler.Tests` | 4,692 / **0 failed** |
| `ILInspector.Research.Tests` | 132 / **0 failed** |
| `dotnet-inspect.Tests` | 2,766 / **1 failed**, 14 skipped |

The one failure is `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, **pre-existing**: verified failing on clean `main` (`7a14f1ff2`) in a separate worktree. Unrelated to this change.

Two further tests in that suite — `RouterVersionTests.Version_WithLatestTag_SkipsCacheAndQueriesVersionApi` and `PackageVersionTests.LatestVersion_AlwaysQueriesNuGet` — **query nuget.org live** and fail in an environment without network access. They pass here (two runs, `Failed: 1` both times) and neither file is touched by this PR, so the suite result is `1 failed` **with network** and `3 failed` without.

## Not fixed here

Caret width is unclamped. A user types one `--focus` at a time, so the figure is stated per focus, on the annotated-source render with `hoist: true` as `AnnotationCaret.Render` is actually called:

| `--focus` | caret rows | over 100 cols | worst |
|---|---:|---:|---:|
| `Semantics` | 13,908 | 2,350 (16.90%) | 27,676 |
| `Allocation` | 14,789 | 1,131 (7.65%) | 1,247 |
| `Cost` | 2,516 | 340 (13.51%) | 331 |
| `Lifetime` | 928 | 121 (13.04%) | **57,945** |
| `Unsafety` | 723 | 31 (4.29%) | 163 |

The worst case in the corpus is reachable with `--focus Lifetime`. Promoting *every* fact at once would give 3,739 of 31,640 rows (11.82%), but no single command produces that render, so it is recorded here only as the union bound and is not the user-facing number. It predates this change — the same worst case appears with extents disabled — and clamping is a layout decision with its own question (a truncated underline must signal that it was truncated). Filed as #3610.

Refs #3570.

## Review

Nine adversarial rounds, two models (Gemini 3.1 Pro, GPT-5.6 Sol), each in an isolated worktree pinned to the exact head. Every finding below was found by a reviewer, not by me — and every one was **a gate that passed for the wrong reason, or a number with the wrong denominator**, never a logic error in the shipped behaviour.

| round | finding | resolution |
|---|---|---|
| 1 | `TryTrimToPrinted` could be **deleted entirely** with no test failure. The non-whitespace assertion was true for the wrong reason: every CLI fixture narrows to a `new` expression, which has no leading whitespace either way. | added `CaretExtent_TrimsAStatementWideExtentOffTheLineIndent` |
| 1 | `CaughtException` (prints `__exception`) is a **second stand-in** of the same shape as `LoadStackSlot` | excluded, and generalized the test to a `[Theory]` over both |
| 2 | the **trailing** trim loop was still ungated — the round-1 fixture line ends on `;` | made the helper `internal`, added direct `[Theory]` gates for both edges |
| 3 | the `Math.Min` **clamp** was ungated | measured, then gated — see below |
| 4 | my clamp measurement counted **the wrong population**: every printed range, not the ranges the caller delivers | re-measured through the real internals; **three stale figures corrected in shipped XML** |
| 4 | the `line != ownerLine` **continuation-line filter** was ungated — 279 facts depend on it (237 later lines, 42 earlier) | added `CaretExtent_IsWithheldWhenTheNarrowNodePrintsOnAContinuationLine` |
| 4 | `char.IsWhiteSpace` could be narrowed to a literal `' '` with no test failure | added tab and no-break-space cases |
| 4 | the caller could **ignore a rejected trim** and store the raw range | added `CaretExtent_IsWithheldWhenTheTrimRejectsTheRange`, on a hand-built range map |
| 5 | `NarrowestPrintedByOffset` dropped a stand-in recorded **after** the real expression only because of the ordering guard | added `CaretExtent_KeepsTheRealExpression_WhenTheStandInIsRecordedAfterIt` |
| 5 | the **equal-width tie-break** was ungated — flipping it changed 96 corpus extents silently | added `CaretExtent_KeepsTheFirstOfTwoEqualWidthCandidates` |
| 5 | `Agreed` has **two** ungated branches — a fact with no extent, and two facts with conflicting extents. GPT found the first, Gemini the second; **neither found both** | added one test per branch |
| 5 | hoist marker `U+0011` leaks from hostile metadata | **pre-existing** — identical leak measured on clean `main`; filed as #3637 |
| 6 | the corpus figures were measured on the **C#-only overlay**, not the render `--focus` produces | every headline figure restated against the annotated-source render, and the render now named wherever a figure appears |
| 7 | round 6 restated the *headline* figures but left the **secondary** ones on the overlay path | methods 135,228 -> **41,952**; `alloc.new` denominator 10,664 -> **10,646**; extent pairs 5,848 -> **5,816**; continuation 270 -> **283** |
| 7 | the headline table **conflated two populations** — facts that *have* a printed node with facts that *survive* to a delivered extent | split into two rows (33,943 have a node, 33,657 survive) |
| 8 | the round-7 sweep fixed the PR body and `AnnotationAnchor.cs` but missed the same stale numbers in **`AnnotationCaret.cs` and test comments** | swept every remaining site |
| 8 | `3,169 nest or overlap` was not merely stale, it was **wrong for the claim it makes** | re-measured to **3,325 of 5,816** |
| 9 | four figures in the PR body were stale: over-budget carets, both suite totals, and the mutation table's mixed-line population | all four re-measured and corrected here |
| 9 | the over-budget figure was measured with **`hoist: false`** — a geometry the product never renders | re-measured under the shipped `hoist: true` call: 3,128 (9.89%) -> **3,739 (11.82%)** |

The two reviewers **disagreed** on that last one: Gemini called it a should-fix that prevents an out-of-range read for multi-line nodes; GPT called it unreachable, because `TryGetLineColumn` guarantees a line-bounded range. Rather than pick a side I measured — and **got the measurement wrong**, which Gemini then caught in round 4. See below.

Corrected: of the **33,657** ranges `ComputeCaretExtents` actually delivers to `TryTrimToPrinted` on the annotated-source path, **0 overhang** their line and **650** end exactly on its last character. (Round 4 published 33,729/652 for this, which is the overlay path; round 6 restated it against the render `--focus` actually produces.) GPT is right on reachability and Gemini's stated rationale does not hold — `TryGetLineColumn` rejects every line-crossing range, so multi-line nodes never arrive. The gate went in anyway, because the helper is now `internal` specifically so it can be tested apart from its caller, and it should not borrow that caller's bounds. The remarks say plainly that the clamp is defensive and never exercised in the corpus, so it is not mistaken later for a measured win.

Same honesty applies to `CaughtException`: **0 facts** in CoreLib share an offset with a printed `__exception`. It is excluded because it is the same shape, not because it was observed, and the XML says so.

Round 5 also produced the sharpest reviewer disagreement of the series. Gemini called the published corpus figures wrong (30,264 rather than 33,729) — but that contradicts **its own round-4 number**, and GPT reproduced 33,729 independently in rounds 4 and 5. The authoritative probe calls the production `ComputeCaretExtents` and sums `extents.Count`, replicating nothing, so a missing filter cannot be missing from it; the gap matches the ancestor-climb trap that produced my own wrong figure in round 4. Gemini also reported the caret-column arithmetic as ungated; applying that exact mutation at the head it reviewed fails four pre-existing tests. Both are answered with evidence in the [round-5 reconciliation](https://github.com/richlander/dotnet-inspect/pull/3617#issuecomment-5144766551) rather than conceded.

Reviewers approved at each fixed head after their findings were addressed.

## Correction: three published figures were stale

Gemini's round-4 review found that my clamp measurement counted every printed range `TryGetLineColumn` accepts across the whole tree, not the ranges `ComputeCaretExtents` actually hands to `TryTrimToPrinted`. Most printed nodes carry no fact and never reach the helper.

Re-measured by invoking the real private members through **reflection** rather than replicating them — my first hand-written replica silently missed the ancestor climb in `TryGetPrintedLine` and under-counted by 1,519, which is precisely the drift that produced the bad figure. Gemini's independent count of 33,729 is confirmed.

Three figures in the **shipped XML** had gone stale when the stand-in rule changed which node is narrowest, and are now corrected:

| claim | was | is |
|---|---|---|
| facts getting an extent | 34,008 (89.97%) | **33,729 (89.23%)** (overlay; the annotated figure is 33,657) |
| facts on a continuation line | 270 | 279 (overlay) / **283** (annotated) |
| trim effect | "161 of 25,682 narrowed lines" | **205 extents moved, 0 rejected** (both renders) |
| clamp population | "none of the 359,584" | **0 of 33,729 overhang; 652 end exactly at the line end** |

The conclusion is unchanged — 0 overhangs, so the clamp is defensive — but the numbers supporting it were wrong, and shipped documentation asserting a wrong number is a defect whether or not the conclusion survives.

Round 4 is also the round where the two reviewers **independently converged**: both arrived at 33,729 by different routes, and both flagged `line != ownerLine`. GPT additionally settled the CRLF question by construction — `TryGetLineColumn` strips terminal CR/LF while `Split('\n')` retains the CR, so a line is always one character *longer*, never shorter, and no overhang or off-by-one is possible.

One reported mutant was **not** a finding: ignoring `TryGetLineColumn`'s failure is *equivalent*, because that method zeroes its out parameters on every false return, so the resulting length of 0 is rejected downstream regardless. Verified by measurement — corpus output is byte-identical under the mutation — rather than by argument. The guard stays; a comment now records the equivalence so a fifth round does not re-derive it.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
@@ -201,7 +201,7 @@ public class AnnotationAnchorTests
     {
         // A statement that wraps puts its narrow sub-expression on a later line
         // than the one the fact is attached to, so narrowing there would draw
-        // the caret under a line the fact does not belong to. 279 facts in
+        // the caret under a line the fact does not belong to. 283 facts in
         // System.Private.CoreLib take this path; all keep the statement-wide
         // caret. Only the if and the allocation carry the offset -- giving it to
         // the inner statement too would make Best pick that, and the owner line
@@ -411,13 +411,13 @@ public class AnnotationAnchorTests
     [InlineData("\tx\t", 0, 3, 1, 1)]
     [InlineData("\u00a0x", 0, 2, 1, 1)]
     // Reaching exactly the last character, so a clamp that stopped one short
-    // would silently drop it. 87,400 of the corpus's printed ranges end here.
+    // would silently drop it. 650 of the extents the caller delivers end here.
     [InlineData("    new object();", 4, 13, 4, 13)]
     // Overhanging the end of the line. ComputeCaretExtents cannot deliver this
     // today -- TryGetLineColumn rejects every range that crosses a line break,
-    // 16,895 of them in the corpus, so 0 of 359,584 survivors overhang -- but
-    // the clamp is the reason an overhang is survivable at all, and an internal
-    // helper is not entitled to assume it keeps only its current caller.
+    // so 0 of the 33,657 extents the caller delivers overhang -- but the clamp
+    // is the reason an overhang is survivable at all, and an internal helper is
+    // not entitled to assume it keeps only its current caller.
     [InlineData("    x", 4, 99, 4, 1)]
     public void TryTrimToPrinted_ShrinksOntoPrintedCharactersAtBothEnds(
         string lineText, int column, int length, int expectedColumn, int expectedLength)
--- a/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
@@ -87,19 +87,24 @@ public class AnnotationAnchorTests
         Assert.Contains(statementNodes, n => n is NewArray);
     }
 
-    [Fact]
-    public void CaretExtent_PrefersTheAllocation_OverAStackSlotStandInSharingItsOffset()
+    [Theory]
+    [InlineData("S_0")]
+    [InlineData("__exception")]
+    public void CaretExtent_PrefersTheAllocation_OverAStandInSharingItsOffset(string standInText)
     {
         // `return new Holder(S_0);` where the raise could not recover the
-        // argument, so it prints a stack-slot stand-in that carries the same
-        // instruction offset as the newobj it feeds. S_0 is the narrower node,
-        // so width alone would underline the argument instead of the allocation
-        // the fact reports. Measured over System.Private.CoreLib, that mistake
-        // affected 50 of 10,664 alloc.new underlines.
+        // argument, so it prints a stand-in that carries the same instruction
+        // offset as the newobj it feeds. The stand-in is the narrower node, so
+        // width alone would underline the argument instead of the allocation the
+        // fact reports. Measured over System.Private.CoreLib, that mistake
+        // affected 50 of 10,664 alloc.new underlines for the stack-slot form;
+        // __exception has the same shape and is excluded for the same reason.
         const int Offset = 5;
         var objectType = TypeRef.CoreLib("System", "Object");
         var holder = TypeRef.CoreLib("Test", "Holder");
-        var slot = new LoadStackSlot(0, objectType);
+        IrExpression slot = standInText == "S_0"
+            ? new LoadStackSlot(0, objectType)
+            : new CaughtException(objectType);
         slot.SetSourceOffset(Offset);
         var allocation = new NewObject(
             new MethodRef(holder, ".ctor", TypeRef.CoreLib("System", "Void"), [objectType], HasThis: true),
@@ -131,8 +136,64 @@ public class AnnotationAnchorTests
 
         // The stand-in is on the line and is the narrower node, so a width-only
         // rule underlines it; this asserts the allocation wins instead.
-        Assert.Contains("S_0", line, StringComparison.Ordinal);
+        Assert.Contains(standInText, line, StringComparison.Ordinal);
         Assert.StartsWith("new ", line.Substring(extent.Column, extent.Length), StringComparison.Ordinal);
+    }
+
+    [Fact]
+    public void CaretExtent_TrimsAStatementWideExtentOffTheLineIndent()
+    {
+        // When the statement itself is the narrowest node carrying the offset,
+        // its printed range starts at the line's indent, so an untrimmed extent
+        // would draw carets left of the code. Nest the statement so the indent
+        // is non-zero and give only the statement the offset, leaving the
+        // expression inside it without one.
+        const int Offset = 7;
+        var objectType = TypeRef.CoreLib("System", "Object");
+        var holder = TypeRef.CoreLib("Test", "Holder");
+        var allocation = new NewObject(
+            new MethodRef(objectType, ".ctor", TypeRef.CoreLib("System", "Void"), [], HasThis: true),
+            []);
+        var statement = new ExpressionStatement(allocation);
+        statement.SetSourceOffset(Offset);
+
+        // A second offset inside the same block widens the if's span past the
+        // statement's, so Best picks the statement rather than the if.
+        var second = new ExpressionStatement(new NewObject(
+            new MethodRef(objectType, ".ctor", TypeRef.CoreLib("System", "Void"), [], HasThis: true),
+            []));
+        second.SetSourceOffset(Offset + 2);
+
+        var inner = new Block(1);
+        inner.Add(statement);
+        inner.Add(second);
+        var outer = new Block(0);
+        outer.Add(new IfStatement(new Constant(true, TypeRef.CoreLib("System", "Boolean")), inner, null));
+        outer.Add(new Return(null));
+        var container = new BlockContainer();
+        container.Add(outer);
+        var function = new IrFunction(
+            "M", holder,
+            new MethodSignature(TypeRef.CoreLib("System", "Void"), [], HasThis: false, GenericParameterCount: 0),
+            [], container);
+
+        CSharpPrinter.PrintRaised(function, out var ranges);
+        var annotation = new Annotation(
+            new AnnotationDescriptor("alloc.new", AnnotationCategory.Allocation, "allocation"),
+            Offset);
+
+        var extents = AnnotationAnchor.ComputeCaretExtents(
+            [annotation], AnnotationAnchor.ComputeSpans(function), ranges);
+        var extent = Assert.Contains(annotation, (IDictionary<IAnnotation, AnnotationAnchor.CaretExtent>)extents);
+
+        Assert.True(AnnotationAnchor.TryGetPrintedLine(statement, ranges, out int lineIndex));
+        string line = ranges.Output.Split('\n')[lineIndex].TrimEnd('\r');
+
+        // The line is indented, so an untrimmed extent would start at column 0.
+        Assert.NotEqual(0, line.Length - line.AsSpan().TrimStart().Length);
+        Assert.False(char.IsWhiteSpace(line[extent.Column]));
+        Assert.False(char.IsWhiteSpace(line[extent.Column + extent.Length - 1]));
+        Assert.Equal(line.Trim(), line.Substring(extent.Column, extent.Length));
     }
 
     [Fact]
--- a/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
@@ -248,6 +248,54 @@ public class AnnotationAnchorTests
         Assert.DoesNotContain(annotation, (IDictionary<IAnnotation, AnnotationAnchor.CaretExtent>)extents);
     }
 
+    [Fact]
+    public void CaretExtent_IsWithheldWhenTheTrimRejectsTheRange()
+    {
+        // The caller must honour a rejected trim rather than storing the raw
+        // range. The C# printer never emits a node as whitespace alone, so the
+        // rejection branch is unreachable through it; a hand-built range map is
+        // the only way to hold the caller to the helper's contract.
+        const int Offset = 7;
+        const string Output = "a   b\n";
+        var objectType = TypeRef.CoreLib("System", "Object");
+        var holder = TypeRef.CoreLib("Test", "Holder");
+        var allocation = new NewObject(
+            new MethodRef(objectType, ".ctor", TypeRef.CoreLib("System", "Void"), [], HasThis: true),
+            []);
+        allocation.SetSourceOffset(Offset);
+        var statement = new ExpressionStatement(allocation);
+        statement.SetSourceOffset(Offset);
+
+        var block = new Block(0);
+        block.Add(statement);
+        var container = new BlockContainer();
+        container.Add(block);
+        var function = new IrFunction(
+            "M", holder,
+            new MethodSignature(TypeRef.CoreLib("System", "Void"), [], HasThis: false, GenericParameterCount: 0),
+            [], container);
+
+        var ranges = new PrintedRangeMap();
+        ranges.Record(statement, 0, 5);
+        // Narrower than the statement and carrying the same offset, so it wins
+        // "narrowest" -- but it covers only the three spaces between a and b.
+        ranges.Record(allocation, 1, 4);
+        ranges.Complete(Output);
+
+        Assert.True(ranges.TryGetLineColumn(allocation, out int line, out int column, out int length));
+        Assert.Equal(0, line);
+        Assert.Equal("   ", Output.Substring(column, length));
+
+        var annotation = new Annotation(
+            new AnnotationDescriptor("alloc.new", AnnotationCategory.Allocation, "allocation"),
+            Offset);
+
+        var extents = AnnotationAnchor.ComputeCaretExtents(
+            [annotation], AnnotationAnchor.ComputeSpans(function), ranges);
+
+        Assert.DoesNotContain(annotation, (IDictionary<IAnnotation, AnnotationAnchor.CaretExtent>)extents);
+    }
+
     [Theory]
     // Leading indent only — what a statement's own printed range looks like.
     [InlineData("    new object();", 0, 17, 4, 13)]
@@ -258,6 +306,10 @@ public class AnnotationAnchorTests
     [InlineData("    x   ", 0, 8, 4, 1)]
     // Already tight: trimming must not move an extent that names an expression.
     [InlineData("    new object();", 4, 12, 4, 12)]
+    // Non-space whitespace. The C# printer indents with spaces, so only a
+    // direct gate can hold the helper to char.IsWhiteSpace rather than ' '.
+    [InlineData("\tx\t", 0, 3, 1, 1)]
+    [InlineData("\u00a0x", 0, 2, 1, 1)]
     // Reaching exactly the last character, so a clamp that stopped one short
     // would silently drop it. 87,400 of the corpus's printed ranges end here.
     [InlineData("    new object();", 4, 13, 4, 13)]
--- a/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
@@ -251,12 +251,13 @@ public class AnnotationAnchorTests
     [Fact]
     public void CaretExtent_KeepsTheRealExpression_WhenTheStandInIsRecordedAfterIt()
     {
-        // The printer records a parent before its children, so the existing
-        // stand-in test only ever sees the stand-in AFTER the allocation has
-        // won. The opposite order -- a stand-in arriving once a real expression
-        // already holds the offset -- is a different branch of the rule, and
-        // deleting it changes 8 extents in CoreLib. A hand-built map is the only
-        // way to pin the order.
+        // The printer records children before their parents -- verified against
+        // the real printer, which emits LoadStackSlot, then NewObject, then
+        // Return -- so the existing stand-in test only ever sees the stand-in
+        // BEFORE the allocation. The opposite order, a stand-in arriving once a
+        // real expression already holds the offset, is a different branch of the
+        // rule, and deleting it changes 8 extents in CoreLib. A hand-built map
+        // is the only way to pin the order.
         const int Offset = 5;
         const string Output = "return new Holder(S_0);\n";
         var objectType = TypeRef.CoreLib("System", "Object");
@@ -303,8 +304,8 @@ public class AnnotationAnchorTests
     public void CaretExtent_KeepsTheFirstOfTwoEqualWidthCandidates()
     {
         // Width is the tie-break, so two real expressions of equal width sharing
-        // an offset must not swap places on a later pass: the printer records
-        // outermost first, and the descendant that follows is no more precise.
+        // an offset must not swap places on a later pass: the one recorded first
+        // stays, and the equally wide candidate that follows is no more precise.
         // Loosening the comparison to a strict one moves 96 extents in CoreLib.
         const int Offset = 5;
         const string Output = "Sink(aaa, bbb);\n";
--- a/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
@@ -196,6 +196,58 @@ public class AnnotationAnchorTests
         Assert.Equal(line.Trim(), line.Substring(extent.Column, extent.Length));
     }
 
+    [Fact]
+    public void CaretExtent_IsWithheldWhenTheNarrowNodePrintsOnAContinuationLine()
+    {
+        // A statement that wraps puts its narrow sub-expression on a later line
+        // than the one the fact is attached to, so narrowing there would draw
+        // the caret under a line the fact does not belong to. 279 facts in
+        // System.Private.CoreLib take this path; all keep the statement-wide
+        // caret. Only the if and the allocation carry the offset -- giving it to
+        // the inner statement too would make Best pick that, and the owner line
+        // would agree, which is the case this test must avoid.
+        const int Offset = 7;
+        var objectType = TypeRef.CoreLib("System", "Object");
+        var holder = TypeRef.CoreLib("Test", "Holder");
+        var allocation = new NewObject(
+            new MethodRef(objectType, ".ctor", TypeRef.CoreLib("System", "Void"), [], HasThis: true),
+            []);
+        allocation.SetSourceOffset(Offset);
+
+        var inner = new Block(1);
+        inner.Add(new ExpressionStatement(allocation));
+        var conditional = new IfStatement(
+            new Constant(true, TypeRef.CoreLib("System", "Boolean")), inner, null);
+        conditional.SetSourceOffset(Offset);
+
+        var outer = new Block(0);
+        outer.Add(conditional);
+        outer.Add(new Return(null));
+        var container = new BlockContainer();
+        container.Add(outer);
+        var function = new IrFunction(
+            "M", holder,
+            new MethodSignature(TypeRef.CoreLib("System", "Void"), [], HasThis: false, GenericParameterCount: 0),
+            [], container);
+
+        CSharpPrinter.PrintRaised(function, out var ranges);
+
+        // Non-vacuity: the allocation must really print on a different line from
+        // the statement the fact anchors to, or this gates nothing.
+        Assert.True(AnnotationAnchor.TryGetPrintedLine(conditional, ranges, out int ownerLine));
+        Assert.True(ranges.TryGetLineColumn(allocation, out int nodeLine, out _, out _));
+        Assert.NotEqual(ownerLine, nodeLine);
+
+        var annotation = new Annotation(
+            new AnnotationDescriptor("alloc.new", AnnotationCategory.Allocation, "allocation"),
+            Offset);
+
+        var extents = AnnotationAnchor.ComputeCaretExtents(
+            [annotation], AnnotationAnchor.ComputeSpans(function), ranges);
+
+        Assert.DoesNotContain(annotation, (IDictionary<IAnnotation, AnnotationAnchor.CaretExtent>)extents);
+    }
+
     [Theory]
     // Leading indent only — what a statement's own printed range looks like.
     [InlineData("    new object();", 0, 17, 4, 13)]
--- a/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
@@ -196,6 +196,36 @@ public class AnnotationAnchorTests
         Assert.Equal(line.Trim(), line.Substring(extent.Column, extent.Length));
     }
 
+    [Theory]
+    // Leading indent only — what a statement's own printed range looks like.
+    [InlineData("    new object();", 0, 17, 4, 13)]
+    // Trailing padding. No printer output produces this today, which is exactly
+    // why it needs a direct gate: the corpus cannot tell whether the trailing
+    // loop still works, so a regression there would be invisible.
+    [InlineData("    x   ", 4, 4, 4, 1)]
+    [InlineData("    x   ", 0, 8, 4, 1)]
+    // Already tight: trimming must not move an extent that names an expression.
+    [InlineData("    new object();", 4, 12, 4, 12)]
+    public void TryTrimToPrinted_ShrinksOntoPrintedCharactersAtBothEnds(
+        string lineText, int column, int length, int expectedColumn, int expectedLength)
+    {
+        Assert.True(AnnotationAnchor.TryTrimToPrinted(lineText, ref column, ref length));
+        Assert.Equal(expectedColumn, column);
+        Assert.Equal(expectedLength, length);
+    }
+
+    [Theory]
+    // All whitespace, and a degenerate range: there is nothing to point at, so
+    // the caller must fall back to the statement rather than draw a caret.
+    [InlineData("        ", 0, 8)]
+    [InlineData("    x", 0, 4)]
+    [InlineData("    x", 2, 0)]
+    [InlineData("    x", -1, 3)]
+    [InlineData("    x", 9, 3)]
+    public void TryTrimToPrinted_RefusesARangeWithNoPrintedCharacters(
+        string lineText, int column, int length)
+        => Assert.False(AnnotationAnchor.TryTrimToPrinted(lineText, ref column, ref length));
+
     [Fact]
     public void ResearchRegistry_RunsAllocationProducer()
     {
--- a/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
@@ -249,6 +249,105 @@ public class AnnotationAnchorTests
     }
 
     [Fact]
+    public void CaretExtent_KeepsTheRealExpression_WhenTheStandInIsRecordedAfterIt()
+    {
+        // The printer records a parent before its children, so the existing
+        // stand-in test only ever sees the stand-in AFTER the allocation has
+        // won. The opposite order -- a stand-in arriving once a real expression
+        // already holds the offset -- is a different branch of the rule, and
+        // deleting it changes 8 extents in CoreLib. A hand-built map is the only
+        // way to pin the order.
+        const int Offset = 5;
+        const string Output = "return new Holder(S_0);\n";
+        var objectType = TypeRef.CoreLib("System", "Object");
+        var holder = TypeRef.CoreLib("Test", "Holder");
+
+        IrExpression slot = new LoadStackSlot(0, objectType);
+        slot.SetSourceOffset(Offset);
+        var allocation = new NewObject(
+            new MethodRef(holder, ".ctor", TypeRef.CoreLib("System", "Void"), [objectType], HasThis: true),
+            [slot]);
+        allocation.SetSourceOffset(Offset);
+        var statement = new Return(allocation);
+        statement.SetSourceOffset(Offset);
+
+        var block = new Block(0);
+        block.Add(statement);
+        var container = new BlockContainer();
+        container.Add(block);
+        var function = new IrFunction(
+            "M", holder,
+            new MethodSignature(holder, [], HasThis: false, GenericParameterCount: 0),
+            [], container);
+
+        var ranges = new PrintedRangeMap();
+        ranges.Record(statement, 0, 23);
+        // The allocation first, so it holds the offset; then the narrower
+        // stand-in, which must not displace it on width.
+        ranges.Record(allocation, 7, 22);
+        ranges.Record(slot, 18, 21);
+        ranges.Complete(Output);
+
+        var annotation = new Annotation(
+            new AnnotationDescriptor("alloc.new", AnnotationCategory.Allocation, "allocation"),
+            Offset);
+
+        var extents = AnnotationAnchor.ComputeCaretExtents(
+            [annotation], AnnotationAnchor.ComputeSpans(function), ranges);
+
+        var extent = Assert.Contains(annotation, (IDictionary<IAnnotation, AnnotationAnchor.CaretExtent>)extents);
+        Assert.Equal("new Holder(S_0)", Output.Substring(extent.Column, extent.Length));
+    }
+
+    [Fact]
+    public void CaretExtent_KeepsTheFirstOfTwoEqualWidthCandidates()
+    {
+        // Width is the tie-break, so two real expressions of equal width sharing
+        // an offset must not swap places on a later pass: the printer records
+        // outermost first, and the descendant that follows is no more precise.
+        // Loosening the comparison to a strict one moves 96 extents in CoreLib.
+        const int Offset = 5;
+        const string Output = "Sink(aaa, bbb);\n";
+        var objectType = TypeRef.CoreLib("System", "Object");
+        var holder = TypeRef.CoreLib("Test", "Holder");
+
+        var first = new NewObject(
+            new MethodRef(objectType, ".ctor", TypeRef.CoreLib("System", "Void"), [], HasThis: true), []);
+        first.SetSourceOffset(Offset);
+        var second = new NewObject(
+            new MethodRef(objectType, ".ctor", TypeRef.CoreLib("System", "Void"), [], HasThis: true), []);
+        second.SetSourceOffset(Offset);
+        var statement = new ExpressionStatement(first);
+        statement.SetSourceOffset(Offset);
+
+        var block = new Block(0);
+        block.Add(statement);
+        var container = new BlockContainer();
+        container.Add(block);
+        var function = new IrFunction(
+            "M", holder,
+            new MethodSignature(TypeRef.CoreLib("System", "Void"), [], HasThis: false, GenericParameterCount: 0),
+            [], container);
+
+        var ranges = new PrintedRangeMap();
+        ranges.Record(statement, 0, 15);
+        // Both three characters wide: "aaa" then "bbb".
+        ranges.Record(first, 5, 8);
+        ranges.Record(second, 10, 13);
+        ranges.Complete(Output);
+
+        var annotation = new Annotation(
+            new AnnotationDescriptor("alloc.new", AnnotationCategory.Allocation, "allocation"),
+            Offset);
+
+        var extents = AnnotationAnchor.ComputeCaretExtents(
+            [annotation], AnnotationAnchor.ComputeSpans(function), ranges);
+
+        var extent = Assert.Contains(annotation, (IDictionary<IAnnotation, AnnotationAnchor.CaretExtent>)extents);
+        Assert.Equal("aaa", Output.Substring(extent.Column, extent.Length));
+    }
+
+    [Fact]
     public void CaretExtent_IsWithheldWhenTheTrimRejectsTheRange()
     {
         // The caller must honour a rejected trim rather than storing the raw
--- a/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
@@ -206,6 +206,15 @@ public class AnnotationAnchorTests
     [InlineData("    x   ", 0, 8, 4, 1)]
     // Already tight: trimming must not move an extent that names an expression.
     [InlineData("    new object();", 4, 12, 4, 12)]
+    // Reaching exactly the last character, so a clamp that stopped one short
+    // would silently drop it. 87,400 of the corpus's printed ranges end here.
+    [InlineData("    new object();", 4, 13, 4, 13)]
+    // Overhanging the end of the line. ComputeCaretExtents cannot deliver this
+    // today -- TryGetLineColumn rejects every range that crosses a line break,
+    // 16,895 of them in the corpus, so 0 of 359,584 survivors overhang -- but
+    // the clamp is the reason an overhang is survivable at all, and an internal
+    // helper is not entitled to assume it keeps only its current caller.
+    [InlineData("    x", 4, 99, 4, 1)]
     public void TryTrimToPrinted_ShrinksOntoPrintedCharactersAtBothEnds(
         string lineText, int column, int length, int expectedColumn, int expectedLength)
     {
--- a/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationAnchorTests.cs
@@ -88,6 +88,54 @@ public class AnnotationAnchorTests
     }
 
     [Fact]
+    public void CaretExtent_PrefersTheAllocation_OverAStackSlotStandInSharingItsOffset()
+    {
+        // `return new Holder(S_0);` where the raise could not recover the
+        // argument, so it prints a stack-slot stand-in that carries the same
+        // instruction offset as the newobj it feeds. S_0 is the narrower node,
+        // so width alone would underline the argument instead of the allocation
+        // the fact reports. Measured over System.Private.CoreLib, that mistake
+        // affected 50 of 10,664 alloc.new underlines.
+        const int Offset = 5;
+        var objectType = TypeRef.CoreLib("System", "Object");
+        var holder = TypeRef.CoreLib("Test", "Holder");
+        var slot = new LoadStackSlot(0, objectType);
+        slot.SetSourceOffset(Offset);
+        var allocation = new NewObject(
+            new MethodRef(holder, ".ctor", TypeRef.CoreLib("System", "Void"), [objectType], HasThis: true),
+            [slot]);
+        allocation.SetSourceOffset(Offset);
+        var statement = new Return(allocation);
+        statement.SetSourceOffset(Offset);
+
+        var block = new Block(0);
+        block.Add(statement);
+        var container = new BlockContainer();
+        container.Add(block);
+        var function = new IrFunction(
+            "M", holder,
+            new MethodSignature(holder, [], HasThis: false, GenericParameterCount: 0),
+            [], container);
+
+        CSharpPrinter.PrintRaised(function, out var ranges);
+        var annotation = new Annotation(
+            new AnnotationDescriptor("alloc.new", AnnotationCategory.Allocation, "allocation"),
+            Offset);
+
+        var extents = AnnotationAnchor.ComputeCaretExtents(
+            [annotation], AnnotationAnchor.ComputeSpans(function), ranges);
+
+        var extent = Assert.Contains(annotation, (IDictionary<IAnnotation, AnnotationAnchor.CaretExtent>)extents);
+        Assert.True(AnnotationAnchor.TryGetPrintedLine(statement, ranges, out int lineIndex));
+        string line = ranges.Output.Split('\n')[lineIndex].TrimEnd('\r');
+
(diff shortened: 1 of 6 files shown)


richlander:

Two rounds, two families, neither my own (AGENTS.md two-model tier): **Gemini 3.1 Pro** and **GPT-5.6 Sol**. Both read `f13560b16` on an isolated worktree with a full build and their own CoreLib sweeps.

| reviewer | verdict |
|---|---|
| Gemini 3.1 Pro | **BLOCKING** (1 blocking, 1 should-fix) |
| GPT-5.6 Sol | **APPROVE WITH SHOULD-FIX** (1 should-fix) |

Fixed in `4642bfa79`.

### Finding 1 — the trim had no gate (both reviewers; Gemini blocking)

Both independently mutated away the `TryTrimToPrinted` call, confirmed the build succeeded, and found **zero tests failed**.

Gemini named the reason precisely, and it is a good lesson: the rewritten tests do assert the underline starts on a non-whitespace character, but every fixture on the CLI path narrows to a `new` expression, which has no leading whitespace either way. **The assertion was true for the wrong reason** — non-vacuous by count, vacuous for this specific invariant.

`CaretExtent_TrimsAStatementWideExtentOffTheLineIndent` builds the case directly: a statement nested one level so its line is indented, holding the only exact-offset stamp in its subtree, so it is its own narrowest node and its printed range begins at the indent. Mutation-verified — build succeeded with the trim replaced by a bare `column < 0 || length <= 0` guard, and the test fails.

### Finding 2 — a second stand-in of the same shape (Gemini, should-fix)

`CaughtException` prints as `__exception` and is stamped with the offset of the instruction that consumes it, so it is the same defect class as `LoadStackSlot`: an 11-character token that beats the statement it sits in on width and would confidently underline something the fact is not about.

I measured before acting. **0 facts in `System.Private.CoreLib` share an offset with a printed `CaughtException`** (probe swept all 37,866 facts). So this is excluded because it is the same shape, not because it was observed, and the remarks now say exactly that rather than implying a corpus win. The stand-in test is a `[Theory]` over both kinds; dropping `CaughtException` from the exclusion fails the `__exception` case and only that case.

### Cleared, with the checks that cleared them

- **`NarrowestPrintedByOffset` bookkeeping** — Gemini enumerated all 6 arrival orders of `[Slot(5), NarrowNonSlot(2), WideNonSlot(100)]` and confirmed `widths[offset]` is correctly rewritten when a non-slot displaces a slot. GPT ran its own synthetic interleavings and agreed.
- **Record value equality on the annotation dictionary** — my own biggest worry, raised explicitly in both prompts. Facts sharing a line always differ in `SourceOffset`, which is a record component, so they never collide as keys. GPT measured **zero equal annotations** across CoreLib and noted that even a synthetic collision would be harmless, since equal records necessarily share an offset and therefore an extent.
- **Hoist arithmetic** — clamping requires `caretColumn < commentColumn - 1`; on the `hoist: true` path `ResearchViews` actually uses, that needs the statement's indent to be less than the member indent minus five, which valid C# layout cannot produce. GPT: reachable on **0 / 25,682** corpus lines.
- **Same-line guard and the `Agreed` bounds backstop** — GPT: **0** bounds-invalid extents across the corpus; both traced `lineText` back to `SourceLine.Split(csResult.Output)`, the same string the extents were measured against.
- **Actively-wrong narrowing** — neither found a case beyond the two stand-ins. The one non-`new` `alloc.new` underline is a printer `/* Unsupported IL ... */` comment that *is* the whole statement, so it is not strictly narrowed and not a mis-selection.

### Independent reproduction of every published figure

GPT-5.6 Sol swept 41,952 methods with 0 failures and reproduced the table exactly: 89.97% exact, 25,682 / 31,641 lines narrowing (81.17%), 10.71% disagreement, 8.12% no extent, 25,224 strictly inside, 0 whitespace edges, 0 out-of-bounds, 10,650 / 10,651 `alloc.new` beginning with `new`, density 92.34 / 12.07 / 1.06 / 0%. Gemini's sweep agreed.

One correction worth recording: GPT notes the eligible-fact denominator is **37,800**, not 37,866 — 66 facts come from methods that produce no printed ranges at all. The percentages are unaffected; the PR body's "37,861 facts" refers to the post-zero-span-drop total and is a different denominator from the one the exact-offset rate is computed over.

### Explicit non-actions

- **Caret width is still unclamped.** Out of scope and pre-existing; filed as #3610 with its measurements.
- **`hoist: false` skew.** Gemini notes an unindented statement rendered without hoisting would skew three columns right. That path is unreachable from the CLI — `ResearchViews` always hoists — and changing the clamp is a layout decision belonging with #3610.

Re-review of the fixed head follows.

richlander:
## Round 3 reconciliation — `698f186ce`

Both reviewers ran against the exact head, in isolated worktrees.

**GPT-5.6 Sol — APPROVE.** Reproduced both per-edge mutations exactly, exhausted the remaining mutation surface (return value, rejection path, loop bounds, column/length writes, ±1 length), re-verified the stand-in rule, and re-ran the corpus sweep: 37,861 anchored facts, 25,682 narrowed lines, 0 invariant violations, 10,650/10,651 `alloc.new` extents beginning with `new`. All published figures hold.

**Gemini 3.1 Pro — APPROVE WITH SHOULD-FIX.** Reproduced the same mutations, confirmed the round-2 gates intact, and judged `internal` + `InternalsVisibleTo` the correct seam on the grounds that the trailing loop is *unreachable* through the public surface. One should-fix: the `Math.Min(column + length, lineText.Length)` clamp is ungated — removing it fails 0 tests.

### The reviewers disagreed, so I measured

Gemini called the clamp load-bearing, "critical to prevent an `IndexOutOfRangeException` for multi-line nodes." GPT had explicitly cleared the same code as "caller-unreachable defensive behavior; `TryGetLineColumn` guarantees line-bounded ranges."

That is a factual question, so I swept CoreLib rather than picking a reviewer:

```
ranges reaching TryTrimToPrinted : 359,584
  overhanging the line end       : 0
  ending exactly at the line end : 87,400
ranges rejected by TryGetLineColumn: 16,895
```

**GPT is right on reachability**, and Gemini's stated rationale does not hold: `TryGetLineColumn` refuses any range that crosses a line break — 16,895 of them — so multi-line nodes never reach the helper at all.

**The gate went in anyway.** `TryTrimToPrinted` became `internal` in the previous commit precisely so it could be tested apart from its caller; a helper in that position should not depend on the single caller that happens to exist today for its bounds. Two cases added in `2fb7b3a37`:

- `4,99` on a 5-character line — the overhang the clamp exists for.
- `4,13` on a 17-character line — reaching the last character exactly, which is where **87,400** corpus ranges end and where an off-by-one clamp would silently drop a character.

Mutation-verified, build clean under both:

| mutation | result |
|---|---|
| drop the clamp entirely | **exactly the overhang case fails** |
| clamp to `lineText.Length - 1` | **4 fail**, including both exact-end cases |

The XML remarks now state that the clamp is defensive and never exercised by the production caller, so it is not later mistaken for a corpus-driven win — the same treatment already given to the `CaughtException` exclusion (0 observed occurrences, excluded for shape).

**Gemini's NIT** — that `length <= 0` is redundant with the `end <= start` rejection — is correct. Kept as an explicit precondition of an internal helper rather than an accident of the loop bounds, with a comment saying so, so the next reader does not re-find it.

`AnnotationAnchorTests` is now 19/19. Round 4 running against `2fb7b3a37`.

richlander:
## Round 4 — Gemini 3.1 Pro: APPROVE WITH SHOULD-FIX, both findings valid

Two findings, both correct, both now fixed in `df5840975`.

### 1. My clamp measurement counted the wrong population

I published "359,584 ranges reach `TryTrimToPrinted`". That counted every printed range `TryGetLineColumn` accepts across the whole tree. `ComputeCaretExtents` is driven by the **annotation list** — most printed nodes carry no fact and never reach the helper. Gemini's count of **33,729** is right and mine was not.

Worse, my first attempt to re-measure it was *also* wrong: a hand-written replica of the internals silently missed the ancestor climb in `TryGetPrintedLine` and under-counted by 1,519 (32,210 vs 33,729). So I stopped replicating and invoked the real private members through reflection:

```
facts total                     : 37,866
facts eligible                  : 37,800
  no narrowest printed node     : 3,792
  node on a CONTINUATION line   : 279
CALLS to TryTrimToPrinted       : 33,729
  overhanging the line end      : 0
  ending exactly at the line end: 652
  REJECTED by the trim          : 0
  MOVED by the trim             : 205
facts with an extent            : 33,729  (89.23% of eligible)
```

That exposed **three further stale figures** in the shipped XML — stale since the stand-in rule changed which node is "narrowest", which I never re-measured:

| claim | was | is |
|---|---|---|
| facts getting an extent | 34,008 (89.97%) | **33,729 (89.23%)** |
| facts on a continuation line | 270 | **279** |
| trim effect | "161 of 25,682 narrowed lines" | **205 of 33,729 moved, 0 rejected** |
| clamp population | "none of the 359,584" | **0 of 33,729 overhang; 652 at the exact end** |

The **conclusion is unchanged** — 0 overhangs, the clamp is defensive, and the round-3 reasoning stands. But documentation that ships a wrong number is a defect regardless of whether the conclusion survives it, and this is the second time on this PR that a figure was right in its claim and wrong in its denominator.

### 2. The continuation-line filter was ungated

Deleting `line != ownerLine` from `ComputeCaretExtents` failed **0 tests**, despite 279 facts depending on it to avoid drawing a caret under a line the fact is not attached to. Same class as the three previous findings.

Added `CaretExtent_IsWithheldWhenTheNarrowNodePrintsOnAContinuationLine`. It asserts up front that the allocation genuinely prints on a different line from its owning statement, so it cannot pass vacuously — the failure mode that has now bitten this PR twice. Mutation-verified, build clean: deleting the filter fails exactly that test.

### Also confirmed by Gemini

- The clamp Theory cases reproduce exactly: drop the clamp → 1 failure (the overhang case); clamp to `Length - 1` → 4 failures.
- My overruling of Gemini's own round-3 rationale was correct — it verified for itself that `TryGetLineColumn` makes a line-crossing range impossible, and said so.
- Two mutants it identified as **equivalent**, correctly not reported as findings: `line >= lines.Length` (unreachable — `Split('\n')` and `BuildLineStarts` count the same newlines) and `column >= lineText.Length` (redundant with the clamp plus `end <= start`).
- Worktree confirmed clean at finish.

Suites at `df5840975`: `AnnotationAnchorTests` **20/20**, Research **132/132**, CLI fast lane **1826/0**.

GPT-5.6 Sol's round 4 is still running against `2fb7b3a37`; its findings will be reconciled against this head.

richlander:
## Round 4 — GPT-5.6 Sol: APPROVE WITH SHOULD-FIX

GPT **independently confirmed both of Gemini's findings**, having reached them by a different route, which is the strongest signal either reviewer produced this round:

- It reproduced my 359,584 / 87,400 / 16,895 figures exactly *and* showed what they actually measure — all printed ranges in fact-bearing methods — against the real call path's `actualTryTrimCalls=33729`, `actualExactEnd=652`, `actualTryGetRejected=0`. Two reviewers, two methods, same 33,729. Fixed in `df5840975`.
- It found `line != ownerLine` ungated, and added detail Gemini did not have: the 279 mismatches split **237 later lines and 42 earlier lines**. Fixed in `df5840975`.

It then found three more surviving mutants. **Two were real, and are fixed in `dbf07f561`:**

### `char.IsWhiteSpace` could be narrowed to a literal `' '`

The C# printer indents with spaces, so no corpus measurement can tell the two apart. Added tab and no-break-space cases to the shrink Theory.

### The caller could ignore `TryTrimToPrinted`'s return value

Rejection never occurs in the corpus — the printer never emits a node as whitespace alone — so this branch is unreachable through the real printer entirely. `Record`/`Complete` are `internal` and the test project already has `InternalsVisibleTo`, so `CaretExtent_IsWithheldWhenTheTrimRejectsTheRange` builds a `PrintedRangeMap` by hand: a narrow node covering only the three spaces between two tokens, carrying the statement's offset. Asserts no extent is produced.

### The third is an equivalent mutant, and is not a finding

GPT listed "ignoring `TryGetLineColumn` failure" as non-equivalent. It is not. `TryGetLineColumn` opens with `line = column = length = 0` and so zeroes its out parameters on **every** false return; the resulting `length` of 0 is then rejected downstream whatever `ownerLine` is.

I checked rather than argued. Under the mutation the corpus output is **unchanged**:

```
facts WITH an extent           : 33,729  (89.23% of eligible)
extents ending at the line end : 652
extents overhanging the line   : 0
```

Byte-identical to the unmutated run. The guard stays — relying on that zeroing is a trap — but a comment now records the equivalence at the call site, because this is the fourth consecutive round in which a reviewer has spent effort re-deriving an equivalent mutant.

### Mutation table for this commit

| mutation | build | result |
|---|---|---|
| `char.IsWhiteSpace` -> `== ' '` | clean | **exactly the tab and NBSP cases fail** |
| ignore the trim's return value | clean | `CaretExtent_IsWithheldWhenTheTrimRejectsTheRange` **fails** |
| ignore `TryGetLineColumn` failure | clean | **no test fails, and no corpus output changes** — equivalent |

### Also confirmed by GPT

- Headline figures reproduced again: 37,861 anchored facts, 25,682 narrowed lines, 0 invariant violations, 10,650/10,651 `alloc.new` extents beginning with `new`.
- The **CRLF question** settled by construction: `TryGetLineColumn` strips terminal CR/LF while `Split('\n')` retains the CR, so the line is always one character *longer*, never shorter — no overhang or off-by-one is possible. It simulated CRLF and got zero overhangs.
- Equivalent mutants correctly excluded: `length <= 0`, the upper-column guard, `end <= start` -> `end == start`, `line >= lines.Length`, and CRLF normalisation before splitting.
- Worktree clean at finish.

Suites at `dbf07f561`: `AnnotationAnchorTests` **23/23**, Research **132/132**, CLI fast lane **1826/0**.

richlander:
## Round 5 reconciliation

Reviewed at `dbf07f561`: **GPT-5.6 Sol** and **Gemini 3.1 Pro**, isolated worktrees, both pinned to that exact head. Both returned **BLOCKING**, with nine findings between them. Both were steered into `AnnotationCaret.Render`/`Agreed`, the least-scrutinised code in the change.

**Four findings were real and are now fixed. Three do not reproduce. Two are equivalent mutants or matters of taste.** Notably, the two reviewers contradict each other on three separate points, so several of these had to be settled by measurement rather than by picking a side.

Fixed head: `d467fc35b`. As before, **no finding was a logic error in shipped behaviour** -- every real one was a correct branch with no test holding it.

### Real, fixed

| # | reviewer | finding | gate | mutant |
|---|---|---|---|---|
| 1 | GPT | `NarrowestPrintedByOffset` drops a stand-in recorded *after* the real expression only because of the ordering guard. Removing it moved **8 corpus extents**, failed 0 tests. | `CaretExtent_KeepsTheRealExpression_WhenTheStandInIsRecordedAfterIt` | M13 |
| 2 | GPT | Equal-width tie-break `best <= width` → `best < width` changed **96 corpus extents**, failed 0 tests. | `CaretExtent_KeepsTheFirstOfTwoEqualWidthCandidates` | M14 |
| 3 | GPT | `Agreed`, **missing-entry** branch: `return null` → `continue` would narrow **498 of 609** mixed CoreLib lines wrongly. | `CaretWidensWhenOnlySomeFactsOnTheLineHaveAnExtent` | M15 |
| 4 | Gemini | `Agreed`, **disagreement** branch (`seen != extent`): two facts narrowing to *different* characters. A genuinely different branch from #3, which GPT did not reach. | `CaretWidensWhenTwoFactsOnTheLineDisagreeAboutTheExtent` | M16 |

Finding 4 is the clearest case this round for running two models rather than one: GPT and Gemini each found one half of `Agreed`, and neither found both.

Each new test asserts non-vacuity first -- that the inputs really do narrow on their own, to distinct columns -- so a future change that disables narrowing outright cannot leave them passing green for the wrong reason.

### Not reproducible

**Gemini, BLOCKING: "the corpus figures are wrong"** -- claimed 30,264 extents / 23,471 narrowed lines / 138 continuation / 10,265 `alloc.new`, against the published 33,729 / 25,682 / 279 / 10,650, attributing the gap to `TryGetPrintedLine` failing for statement owners.

The published figures come from a probe that calls the production `ComputeCaretExtents` and sums `extents.Count`. It replicates no internal logic, so a filter Gemini believes is missing cannot be missing from it -- if `ComputeCaretExtents` rejected those facts, the sum would already be lower. Re-run on a clean rebuild this round: **33,729, unchanged.** GPT independently reproduced 33,729 by reflection in rounds 4 *and* 5, and independently reproduced the 279 continuation figure with its 237/42 split.

The 30,264 is very likely the ancestor-climb trap. `TryGetPrintedLine` climbs to the nearest printed ancestor when the owner is an inline expression body; `TryGetLineColumn` does not. **I made this exact error myself in round 4** and got 32,210 with 1,647 owners spuriously reported as having no printed line -- which is why round 4 moved to reflection onto the real private members instead of replication. Gemini disagreeing with its own round-4 number of 33,729 is the strongest evidence that the discrepancy is in the probe, not the code.

**Gemini, BLOCKING: `caretColumn = extent.Column + hoisted` is ungated** -- claimed `caretColumn = hoisted` passes every test. Applied that exact mutation at `dbf07f561`, the head Gemini reviewed:

```
CaretUnderlinesTheTrimmedStatementAndNotItsIndent                      [FAIL]
EveryHoistedCaretPointsExactlyAtItsStatement(line: "    if (x)")        [FAIL]
EveryHoistedCaretPointsExactlyAtItsStatement(line: "  Sink(new ...);")  [FAIL]
CaretCommentSitsOnTheMemberGutterRegardlessOfNestingDepth               [FAIL]
Total: 26, Failed: 4
```

Four pre-existing tests kill it. On the fixed head it is six.

**GPT, BLOCKING: hoist marker U+0011 leaks from hostile metadata** -- real, but **pre-existing**. Built a hostile assembly with `PersistedAssemblyBuilder` carrying a type named `Evil\u0011Marker` and ran it against both builds: **2 marker-bearing lines on clean `main` @ `7a14f1ff2`, 2 on this PR** -- identical. `AnnotationText.cs` is untouched here and `HoistMarker` predates the branch. Filed as **#3637** rather than expanding this PR, as was done for #3610.

### Declined

**Gemini, SHOULD-FIX: `length <= 0` is killable.** It is an equivalent mutant. Length 0 gives `end == start`; negative length gives `end < start`; both are already rejected by the `end <= start` guard on the next line. Gemini's own evidence is that no test fails when it is removed -- which is the definition of equivalence, not of a coverage gap. This also contradicts GPT and Gemini's own round-4 NIT, which correctly called it redundant. The redundancy is documented in place so a seventh round does not re-derive it.

**Gemini, NIT: the hand-built `PrintedRangeMap` is "theatre".** GPT reviewed the same construct this round and called it legitimate. It is: `Record`/`Complete` are `internal` and the test assembly has `InternalsVisibleTo`, so the map is built through the production recording API, not faked. It is also the only way to control *record order*, which is precisely what finding #1 -- Gemini's co-reviewer's BLOCKING finding -- required a test for. M11, M13 and M14 are all killed through it.

### Mutation ledger

M13–M16 join the eleven from earlier rounds. **None of the sixteen breaks the build** -- that is the whole point of the exercise; a mutant that fails to compile proves nothing.

| # | mutation | result |
|---|---|---|
| M13 | drop the late stand-in ordering guard | ordering test fails |
| M14 | `best <= width` → `best < width` | tie test fails |
| M15 | `Agreed` missing entry: `return null` → `continue` | partial-narrowing test fails |
| M16 | `Agreed` disagreement: `seen != extent` → `false` | disagreement test fails |

Suites on the fixed head: `AnnotationAnchorTests` + `AnnotationGestureTests` **53**, Research **132**, CLI fast lane **1826** (12 skipped), 0 failed.

Round 6 to follow at `d467fc35b`, per the two-clean-rounds-on-the-fixed-head rule in `AGENTS.md`.

richlander:
## Round 6 reconciliation

Reviewed at `d467fc35b`: **GPT-5.6 Sol** (**BLOCKING**) and **Gemini 3.1 Pro** (**APPROVE**). Both were steered away from `AnnotationAnchor.cs`, which five rounds had already hammered, and into `ResearchViews.cs` integration and `AnnotationCaret.Render` geometry.

Fixed head: `577d9831b`, then merged with `main` (`ec7b254ef`).

**Round 6 finally produced a finding of a different kind.** The first five rounds each found a correct branch with no test holding it. This one found that **a published measurement described the wrong render** — and it is the one finding in the series I would have shipped without noticing.

### GPT, BLOCKING: the corpus figures were measured on the wrong path — valid, fixed

There are two production caret paths, and they print differently:

| path | `importMethodBody` | extents |
|---|---|---:|
| Annotated Source (`ResearchViews.cs:263`) — the render `--focus` produces | `ImportMethodBody` | **33,657** |
| C#-only overlay (`:445`) | `null` | 33,729 |

Both probes I published passed `importMethodBody: null`. That is the overlay, not the view this PR demonstrates. Reproduced GPT's figure exactly:

```
facts eligible                  : 37,800
CALLS to TryTrimToPrinted       : 33,657
  overhanging the line end      : 0
  ending exactly at the line end: 650
  REJECTED by the trim          : 0
  MOVED by the trim             : 205
facts with an extent            : 33,657  (89.04% of eligible)
```

and independently reproduced the narrowing ratio at **25,628 / 31,640 (81.00%)**, matching GPT's number.

**Neither path is buggy** — each computes extents against its own printed body, and GPT's probes confirmed no body/range mismatch, including under style lenses. The defect was entirely in what I published: an unqualified number, when extents are measured in *printed characters* and are therefore only meaningful against a stated render.

Corrected everywhere, with the render named in place:

| | was | now |
|---|---|---|
| facts with an extent | 33,729 (89.23%) | **33,657 of 37,800 (89.04%)** |
| narrowed caret lines | 25,682 / 31,641 (81.17%) | **25,628 / 31,640 (81.00%)** |
| continuation-line facts | 279 | **283** |
| ending at the line end | 652 | **650** |
| trim moves / rejects | 205 / 0 | **205 / 0** — unchanged |

GPT also caught that the PR body's **headline table still carried `34,008 / 89.97%`** — round 4 corrected the body's correction table but not its headline. Fixed.

The overlay figure is now published *alongside* rather than dropped, so the next person who measures that path and gets 33,729 finds it already accounted for instead of filing it as a discrepancy. This is also, in hindsight, the most likely explanation available for a round-5 dispute: **measuring a different render gives a different number, and neither party is wrong until the render is stated.**

### GPT, minor: a test comment described the printer's recording order backwards — valid, fixed

Settled by measurement rather than by reading. Reflecting into `PrintedRangeMap._ranges` after a real print:

```
[0] PrintedRange { Node = LoadStackSlot S_0, Characters = 31..34 }
[1] PrintedRange { Node = NewObject Holder,  Characters = 20..35 }
[2] PrintedRange { Node = Return,            Characters = 13..37 }
```

The printer records **children before parents**. My comment claimed the opposite, which made the new test look like it duplicated the existing one rather than covering the genuinely opposite order. The tests were right; the prose explaining them was inverted. Both comments corrected.

### Gemini, APPROVE — three equivalent mutants, correctly identified as such

Gemini probed three gates and, in each case, correctly concluded the surviving mutant was **equivalent** rather than a coverage gap:

- `Best`: `offset <= span.Max` → `<`. Rejected by `containing`, the `preceding` fallback selects the same statement.
- `Agreed`: `final.Column >= 0` → `> 0`. Rejecting column 0 returns `null`, and the statement-wide fallback for an unindented statement *also* starts at column 0.
- The gutter padding clamp `Math.Max(1, ...)` → `Math.Max(0, ...)`: a deliberate formatting preference (`// ^^^^` versus `//^^^^`), already documented in place.

That is the distinction this series has been trying to hold throughout, and it is worth noting that the round-5 disagreements about equivalence did not recur.

### Gemini: expression-bodied caret misalignment — real, pre-existing, filed

`FormatSourceWithDeclaration` appends an expression body after `=> `, so the code starts near column 19 while the caret line still carries only its 4-column hoist padding. Pre-existing **by construction**: `ApiOutputFormatter.cs` does not appear in `git diff --stat origin/main...HEAD`, and `BodyIndentWidth`, `Flatten` and the `hoist` parameter all exist on `origin/main`. This PR narrows the caret's *length*, never its *origin*. Filed as **#3638**.

That makes three pre-existing caret defects surfaced by adversarial review of a change that causes none of them: **#3610** (width unclamped), **#3637** (hoist marker leak), **#3638** (expression-bodied origin).

### Validation at the fixed head, after merging `main`

| suite | result |
|---|---|
| `AnnotationAnchorTests` + `AnnotationGestureTests` | **53 / 0 failed** |
| `ILInspector.Research.Tests` | **132 / 0 failed** |
| `dotnet-inspect.Tests` (fast lane) | **1,982 / 0 failed**, 12 skipped |

Figures re-measured after the merge and unchanged at 33,657 / 89.04%.

GPT's blocking finding is fixed, so round 7 goes to the new head per the fixed-head rule in `AGENTS.md`.

richlander:
## Round 7 reconciliation

Reviewed at `c6787538d` (fixed head plus `main` merged): **GPT-5.6 Sol** (**BLOCKING**) and **Gemini 3.1 Pro** (**APPROVE**).

Fixed head: `d657deb9a`.

Both reviewers reached the same conclusion about the code — **the shipped behaviour is correct and the design is sound** — and both said so explicitly, independent of coverage. The remaining disagreement was entirely about published numbers, and the two reviewers were **each right about the figures they measured**.

### Gemini, APPROVE — full independent reproduction

Gemini built its own reflection probe against `NarrowestPrintedByOffset`, `TryGetPrintedLine` and `TryTrimToPrinted` on the annotated-source render and reproduced **every headline figure exactly**: 33,657 extents, 283 continuation, 205 trim moves / 0 rejects, 0 overhang, 650 exact-end, 31,640 caret lines, 25,628 narrowed (81.00%), 3,385 disagreement (10.70%), 2,627 no-extent (8.30%), 25,172 strictly narrower (98.22%), 0 invariant violations.

This is the number the round-5 dispute was about, now independently confirmed by the reviewer that disputed it, on the render this PR actually produces. That closes it.

Gemini also probed the `main` merge specifically — `af6cb96c9` (escaping untrusted metadata names) and `fef762c13` (local declaration placement) both change printed character widths — and confirmed extents inherit the new widths correctly, because they are measured from the `PrintedRangeMap` *after* the render. 0 overhangs, 0 invariant violations post-merge.

### GPT, BLOCKING — the *secondary* figures were still on the overlay path

Round 6 restated the headline figures for the annotated-source render and stopped there. GPT found that the rest of the document had not moved with them. Every one reproduces:

| claim | published | measured (annotated) |
|---|---:|---:|
| imported methods | 135,228 | **41,952** (40,917 print a body, 18,538 carry a fact) |
| `alloc.new` denominator | 10,664 | **10,646** |
| extent pairs sharing a line | 5,848 | **5,816** |
| ...that nest or overlap | 3,191 | **3,169** |
| continuation-line facts (rules section) | 270 | **283** |
| trim effect | "161 lines" | **205 extents moved** |

The `135,228` was the worst of these: not a stale figure but a **wrong quantity**, near-certainly a metadata-wide method count rather than the imported bodies the sweep actually walks. All corrected.

GPT also caught that the headline table **conflated two populations**. `33,943` facts have an exact-offset printed node; `33,657` survive to a delivered extent. I had labelled the second with the first's description. Both rows are now published, because the gap between them — 283 continuation-line facts and 3 that are not line-positionable — *is* the withholding rule working, and hiding it understates what the rule does.

### The one figure that survived, and why it is now stated differently

The `50 alloc.new underlines moved` claim was re-measured by mutation rather than trusted. Disabling the stand-in rule and re-sweeping:

| | underlines | not starting with `new` |
|---|---:|---:|
| stand-in rule **off** | 10,659 | **51** |
| stand-in rule **on** | 10,646 | **1** |

51 → 1 is the 50, confirmed on the annotated path. Only the denominator was wrong. The XML and the PR body now carry the evidence rather than the bare count, so the next reader can see the rule earns its complexity instead of taking it on faith. The single remaining case is a printer `/* Unsupported IL ... */` comment that *is* the whole statement — not a mis-selection.

### What rounds 6 and 7 together established

Six rounds found eleven correct-but-untested branches. Rounds 6 and 7 found something different and, for a change whose entire justification is a measurement, more serious: **a figure without a named render is not a claim about anything.** Two production paths print the same members differently, so they yield different extents, and neither is wrong. Every number in this PR now names its render, and the overlay figures are published alongside rather than dropped so that measuring the other path reproduces a documented value instead of looking like a discrepancy.

That also retires the round-5 dispute properly. It was very likely never a disagreement about the code at all — it was two parties measuring different renders without either naming one.

### Validation at `d657deb9a`

| suite | result |
|---|---|
| `AnnotationAnchorTests` + `AnnotationGestureTests` | **53 / 0 failed** |
| `ILInspector.Research.Tests` | **132 / 0 failed** |
| `dotnet-inspect.Tests` (fast lane) | **1,982 / 0 failed**, 12 skipped |

Both reviewers judged the behaviour correct and the design sound. GPT's blocking finding was documentation-only and is fixed, so round 8 goes to the new head per the fixed-head rule in `AGENTS.md`.

richlander:
## Round 8 reconciliation

Reviewed at `d657deb9a`: **GPT-5.6 Sol** (**BLOCKING**) and **Gemini 3.1 Pro** (**BLOCKING**). Both confirmed the shipped behaviour is correct and the design sound; both blocked on figures that round 7 corrected in the PR body and `AnnotationAnchor.cs` but **missed where the same numbers also appear in `AnnotationCaret.cs` and the test comments**.

Fixed head: `f26fa8e0a`. This time I swept every numeric claim in every changed file rather than fixing the ones reported.

### Corrected

| claim | where | was | now |
|---|---|---:|---:|
| extent pairs sharing a line | `AnnotationCaret.cs:188` | 5,848 | **5,816** |
| ...that nest or overlap | `AnnotationCaret.cs:189` | 3,191 | **3,325** |
| worst line | `AnnotationCaret.cs:192` | nine extents | **eight** |
| mixed-extent lines | `AnnotationGestureTests.cs:192` | 609 / 498 | **615 / 499** |
| continuation-line facts | `AnnotationAnchorTests.cs:204` | 279 | **283** |
| test count | PR body | "19 tests" | **13 methods / 25 cases** |

### One of these is a correction of substance, not of render

GPT found that **3,169 was wrong for the claim it was making**, and it was right. The original probe bucketed pairs sharing a start column separately from the ones it called nested — 156 of them. But two extents starting at the same column plainly *do* overlap, and they are the single worst case for the design this number exists to reject: a per-extent caret row would draw two underlines from the same column at different widths.

Re-measured with the buckets merged where they belong:

```
pairs=5,816  overlapping(any)=3,325  disjoint=2,491  sameStartColumn=156
worst line: 8 distinct extents (System.SpanHelpers.NonPackedContainsValueType line 17)
```

3,169 + 156 = 3,325. Counting them correctly **strengthens** the argument rather than weakening it — 57% of co-located pairs overlap, not 54%. GPT's independent count of the worst line at **8** also stands; my earlier probe reported 9 by grouping differently, and the careful re-measurement agrees with GPT. Its 615 / 499 for mixed-extent lines reproduced exactly too.

The two reviewers disagreed here: Gemini flagged `AnnotationCaret.cs` as stale against the commit message, while GPT checked whether the *number itself* was right and found it was not. Both were useful; only one of them found the real defect.

### A stale figure I removed rather than restated

The clamp comments in `AnnotationAnchorTests.cs` quoted **87,400**, **16,895** and **359,584**. GPT offered annotated-render equivalents (87,427 / 16,945 / 359,876), but those are the same *population* — every printed range in the tree — that **round 4 established is not the population the caller delivers**. Restating them on the correct render would have preserved a denominator already refuted for a different reason.

They now cite what the caller actually hands the helper: **650** extents ending exactly at the line end, **0 of 33,657** overhanging. Fewer numbers, all of them about the thing under test.

### Verified unchanged

- Headline arithmetic reproduced by both reviewers: **33,943 − 283 − 3 = 33,657**.
- `c6787538d..d657deb9a` touched only comments — no production logic, confirmed by both.
- `f26fa8e0a` likewise changes only XML remarks and test comments.

### Validation

| suite | result |
|---|---|
| `AnnotationAnchorTests` + `AnnotationGestureTests` | **53 / 0 failed** |
| `ILInspector.Research.Tests` | **132 / 0 failed** |

### On the shape of these last three rounds

Rounds 1–5 found eleven correct-but-untested branches. Rounds 6–8 found **no defect in shipped behaviour at all** — they found that the *evidence for it* was wrong three times over: measured on the wrong render, then corrected in some places and not others, then correct everywhere but computed with a bucket in the wrong column.

For a change whose whole justification is a corpus measurement, that is not a lesser class of finding. Every number in this PR is now measured on the named render, reproduced independently, and cited where it is used. Round 9 goes to `f26fa8e0a`.

richlander:
## Round 10 reconciliation

Both reviewers returned **BLOCKING**. Two findings were real and are fixed; one is refuted by the reviewer's own run. The code at `f26fa8e0a` is unchanged — round 10 was an evidence audit, because round 9 (Gemini) approved the behaviour with full independent reproduction.

### Fixed — the over-budget population was not one a user can produce (Gemini)

Gemini confirmed from `ResearchViews.cs` that both render paths pass `hoist: true`, so the geometry restated after round 9 is right. But it rejected the *population*: "every fact promoted to the caret gesture" is a union over all `--focus` strings, and `AnnotationGestureSelector.Focus` takes one string, so **no single command renders it**. Publishing a failure rate against a render no user can generate misstates what they will see.

Correct, and the fix is to measure what a user actually runs — one focus at a time:

| `--focus` | caret rows | over 100 cols | worst |
|---|---:|---:|---:|
| `Semantics` | 13,908 | 2,350 (16.90%) | 27,676 |
| `Allocation` | 14,789 | 1,131 (7.65%) | 1,247 |
| `Cost` | 2,516 | 340 (13.51%) | 331 |
| `Lifetime` | 928 | 121 (13.04%) | **57,945** |
| `Unsafety` | 723 | 31 (4.29%) | 163 |

The union figure is retained only as a labelled bound. This also locates the #3610 pathological line precisely: it is reachable with `--focus Lifetime`.

Worth recording that this figure has now been wrong three ways in three rounds — 3,128 from an unshipped `hoist: false` geometry, then 3,739 against an unproducible population — which is the strongest argument in this PR for stating the render and the gesture next to every number.

### Fixed — the suite result depends on network (GPT)

GPT measured `Total: 2766, Failed: 3`, against the published 1. The two extra are `RouterVersionTests.Version_WithLatestTag_SkipsCacheAndQueriesVersionApi` and `PackageVersionTests.LatestVersion_AlwaysQueriesNuGet`, which query nuget.org live; neither file is in this PR's diff (`git diff --name-only origin/main...HEAD`), and both pass here across two runs (`Failed: 1` each time).

Environmental rather than a defect — but "1 failed" was stated unconditionally and is not unconditionally true, so the body now says `1 failed` **with network**, `3` without. GPT was right to refuse to assume.

### Refuted — the decompiler suite total is not stale (Gemini)

Gemini reported `Discovered: ILInspector.Decompiler.Tests (4041 test cases to be run)` against the published 4,692 and called 4,692 stale. Both numbers come from the **same run**:

```
Discovered:  ILInspector.Decompiler.Tests (4041 test cases to be run)
   ILInspector.Decompiler.Tests  Total: 4692, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0
```

4,041 is the discovery count and 4,692 the executed total; the gap is `[Theory]` cases expanded at runtime. The published figure is the executed total, which is what a suite result should report. Not actioned.

### Scope both reviewers did not reach

Neither reviewer independently re-derived the over-budget counts, and Gemini did not dynamically re-derive 499/615 or 498/609 — it confirmed the mutation fails the gate but read the counts from the test comments. Both said so explicitly rather than staying silent, which is what makes the gap actionable: those two figures currently rest on my measurement alone.

Reviewers: Gemini 3.1 Pro and GPT-5.6 Sol, isolated worktrees pinned to `f26fa8e0a`, both worktrees verified clean afterwards.

richlander:
## Round 10 — resolved, both reviewers APPROVE

Both round-10 reviewers initially returned BLOCKING. Both have now independently
re-derived every corpus figure in the body and flipped to APPROVE. Recording how,
because one of the two blocks was a real defect in *my* published numbers' provenance
and the other was a real defect in the reviewer's probe — and the distinction only
became visible by measuring rather than arguing.

### What was actioned

**The over-budget population was not producible by any single command** (Gemini).
Valid, and the more interesting of the two findings. `AnnotationGestureSelector.Focus(null)`
returns `SideOnly`, so **no carets render at all without a `--focus`**. The published
union was therefore a union over all five focus strings that no user invocation produces.
The body now leads with a per-`--focus` table and keeps the union only as a labelled bound:

| `--focus` | caret rows | over 100 cols | worst |
|---|---:|---:|---:|
| Semantics | 13,908 | 2,350 (16.90%) | 27,676 |
| Allocation | 14,789 | 1,131 (7.65%) | 1,247 |
| Cost | 2,516 | 340 (13.51%) | 331 |
| Lifetime | 928 | 121 (13.04%) | 57,945 |
| Unsafety | 723 | 31 (4.29%) | 163 |
| *(union — not a single command)* | 31,640 | 3,739 (11.82%) | 57,945 |

**The suite figure stated a condition it did not name** (GPT). The body said `1 failed`
unconditionally; two further CLI tests query nuget.org live and fail without network.
Neither file is in this diff. The body now names the condition rather than reporting the
convenient number.

### What was refuted, with the reason

**"Decompiler total 4,692 is stale, actual 4,041."** Both numbers come from the same run:
`Discovered: (4041 test cases to be run)` and `Total: 4692`. The gap is `[Theory]` expansion
at runtime. Not stale.

**"Union is 3,120 of 28,208, your figures are wrong."** Rather than overrule this I measured
where the two probes diverge, and found two defects in the verification probe — both since
confirmed by the reviewer:

1. **`ComputeSpans` ran before `PrintRaised`.** Printing raises and rewrites the IR in place,
   so pre-raise spans do not line up with the post-raise printed ranges carets are drawn
   against. Measured: **28,208** caret-bearing lines before vs **31,640** after — the "before"
   figure is *exactly* the denominator reported, which identifies the cause rather than
   merely disagreeing with the result.
2. **One `IrFunction` reused across both renders.** `PrintRaised` mutates in place, so the
   second render sees an already-raised graph. Measured: **23 of 200** members render
   differently on a second `PrintRaised`. This was also visible as a self-consistency failure
   in the reviewer's own output — annotated and overlay mixed-line counts came back *byte
   identical* (418/514), which cannot happen for two renders round 6 established differ.

With both corrected, the reviewer reproduces **499/615**, **498/609**, all five per-focus
rows, and **3,739 of 31,640** exactly.

### Independent reproduction

Both reviewers had initially reported that they had *not* re-derived the corpus counts
themselves — they had checked internal consistency. That gap is now closed. Each has since
reproduced, from scratch and independently of the other, all five per-focus rows, both
mutation-table populations, and the union bound.

Ten adversarial rounds. Every figure in the body now has two independent derivations.

richlander:
Ready to merge.

Gate verified at head `f26fa8e0a` — the exact commit both round-10 reviewers approved:

- **Required review:** round 10 clean. Gemini 3.1 Pro APPROVE, GPT-5.6 Sol APPROVE, both after independently re-deriving every corpus figure in the body from scratch.
- **CI:** `gh pr checks` exit 0. `ci-required`, `build-net10`, `decompiler-gates`, `test (ubuntu-26.04, linux-x64)`, `csharp-diff-smoke`, `changes` all pass.
- **Mergeability:** CLEAN / MERGEABLE, `origin/main` fully integrated, not a draft.
- **Head unmoved** since the reviewed commit, so the approvals apply to what would merge.

Non-blocking follow-ups, tracked separately and deliberately not in this PR:

- **#3610** — the pathological 57,945-column line reachable under `--focus Lifetime`. This PR narrows extents; it does not cap them.
- **#3637**, **#3638**.
- Caret **stacking** (the display half of `--focus`) is the next slice of #3570 and is stacked on this branch. It depends on the narrowed extents this PR lands, which is why it is not bundled here.
- The eight `alloc.box` facts on `Tuple`8.Equals` first operands still arrive without an extent. That is an anchoring gap this PR does not close; stacking makes it visible rather than causing it.

Issue #3635, opened 2026-07-31
Untrusted text containment: one mechanism, two policies (umbrella)


richlander:
## The problem

Untrusted text — metadata names, feed URLs, package identities, resource names — reaches rendered output. The repository has fixed this **six times**, each fix correct and local, and the class has survived every one of them.

This issue names the class, fixes the scope, and gives the other work a single thing to point at.

### 1. Six independent hazard sets

Every escaper in the tree carries its own idea of what is dangerous. Sweeping each predicate over the BMP:

| claims | predicate |
| ---: | --- |
| 76 | `CSharpIdentifierCore.IsRenderingHazard` |
| 78 | `CSharpIdentifierCore.RequiresLiteralEscape` |
| 76 | `InspectionResultView.ContainProse` |
| 65 | `MetadataTableProjector.IsControl` |
| 62 | `AttributeReader.cs:782` (inline) |
| 65 | `ResourceExtractor.cs:195` (inline) |
| **110** | `InertText` `TextPolicy.Field` (#3563) |

**Six distinct sets across seven predicates.** Rows 1 and 3 are the same *size* and a different *set*. `MetadataTableProjector.IsControl` is a hand-rolled range that measures byte-identical to `char.IsControl` — it buys the appearance of deliberation and nothing else.

This is exactly what #3319 was filed to end, and #3380 ended it for *four* of them. The other three were never in its scope.

### 2. The containment property is syntactic, so it must be pinned by hand

`MarkoutRowContainmentTests` does not assert that columns are contained. It asserts that **the set of columns which are not contained has not changed** — currently **359 columns across 86 row types**, tracked by #3463. Its own remarks are honest about this: *"Not a leak list."*

That is a good ratchet. But it is a ratchet, and it exists because the property being checked is *"is `Contain(...)` spelled at this site?"* — a syntactic fact re-established at every producer and forgettable at every producer.

The cost is visible in a single expression. On `main`, `ApiOutputFormatter.cs:1106`:

```csharp
new FieldSummaryRow(
    OperatorNames.FormatDisplayName(e.members[0].Name),          // uncontained
    CSharpIdentifier.ContainRenderedText(e.members[0].ReturnType ?? ""), // contained
    SignatureDecodeMarker(e.members))
```

Adjacent arguments, one contained and one not, in the same constructor call.

### 3. Every gate's domain is narrower than the set of producers

[...]

Issue #3463, opened 2026-07-29
Contain remaining view-record string members (follow-up to #3319)


richlander:
## Summary

View records expose ~290 uncontained `string` members. None leaks today, but the
property is held by *construction-site discipline*, not by the type — every
call site currently wraps metadata-derived text in `MarkoutInline.Code` or an
equivalent, and nothing fails if a future site forgets.

#3380 contained the members that a shape-derived gate proved reachable
(`LibraryInspectionView` scalars and rows, `ApiViews.UnsafeMemberRow`,
`ApiViews.TopLeverageRow`). The rest are recorded here.

## Measured bound

Running the real hostile-assembly fixture through the library view at all four
verbosities produces **0 raw rendering hazards**. So this is not a live leak;
it is an unverified invariant. The honest statement is:

> Safe by current analyzer behavior, not safe by construction.

## Uncontained string members by file

| File | Count |
| --- | --- |
| `Views/ApiViews.cs` | 161 |
| `Views/LibraryInspectionView.cs` | 59 |
| `Views/DiffViews.cs` | 28 |
| `Views/FindViews.cs` | 15 |
| others (8 files) | ~27 |

## The reusable gate

`LibraryViewShapeDerivedContainmentTests` (added in #3380,
`src/dotnet-inspect.Tests/UntrustedLibraryViewContainmentTests.cs`) fills every
`string` in a view model **by reflection** to depth 4 with a hostile marker,
renders, and asserts no hazard survives and no line splits. It derives its
coverage from the shape rather than a hand-kept list — which is exactly why it
found 7 raw fields that the hand-enumerated fixture beside it missed, in a test
whose own comment declared that hand-enumeration was what had failed.

Extending that fixture to the remaining view roots is the tractable path: point
`FillStrings` at each view model, fix what it reports, and the gate then fails
on any newly added raw member.

Note the constraint the fixture already encodes: a reflection filler must
assign every `List<>` (a null list yields a model no analyzer emits, the view
throws, and the failure says nothing about containment).

## Not in scope

`mdi` has its own entry point and cannot reach this writer — tracked separately
in #3444.

richlander:
Filed #3635 as the umbrella for the containment class.

[...]

PR #3636, opened 2026-07-31
Add InertText: make untrusted text inert at the type level

richlander:
Part of #3635. Adds the primitive only — **no production caller is migrated here.** The NuGet work that motivated it stacks on top in #3563.

## Why a type and not another escaper

The repository has fixed this class six times. Each fix was correct and local, and the class survived every one, because the property being enforced is **syntactic**: *"is `Contain(...)` spelled at this site?"* That must be re-established at every producer, is forgettable at every producer, and composition discards it silently.

The cost is measurable. `MarkoutRowContainmentTests` pins **359 columns across 86 row types** (#3463) and asserts not that they are contained but that *the set has not changed*. Its own remarks are honest about this — "Not a leak list." It is a good ratchet, and it is a ratchet.

`InertString` moves the property from the call site to the **value**. Treated text has a different type from untreated text, so it survives composition and a site that merely passes it along cannot lose it. Forgetting becomes a compile error rather than a missing line in a hand-pinned list.

## The audit boundary

Encoding is invertible — that is what lets an already-treated value be re-treated for a different sink instead of double-encoded, and lets `EnsurePermitted` repair a spliced value rather than reject it.

Invertibility is also the risk, so the decoder lives in namespace `InertText.Encoding`. Text enters through `InertString`'s constructor and leaves through `ToString` already spelled, so the currency namespace covers every ordinary use and the reversing half is visible when a file reaches for it.

**The audit is one search, and the string is `InertText.Encoding`.** Reaching the decoder means naming its namespace, and that act *is* the opt-in. Decoding is legitimate and stays public; the design does not try to prevent it, only to make it impossible to do quietly.

A namespace can be named in two places, and both are equally legitimate:

```csharp
using InertText.Encoding;            // named in the import block
VisualEncoder.TryDecode(inert.ToString(), out string? original);
```

```csharp
// no using directive of any kind in this file. InertText.Encoding here is the
// namespace; there is no InertText type and no Encoder member on InertString.
InertText.Encoding.VisualEncoder.TryDecode(inert.ToString(), out string? original);
```

Neither is evasion — the second names the namespace at the call site instead of the top of the file, in plain sight on the line that uses it. What matters is that the *audit* covers both, which is why the search is for the bare namespace and not for `using InertText.Encoding`: a fully-qualified call declares itself on its own line and needs no directive at all, so a reviewer grepping only the import block would read that file as clean and conclude it cannot decode.

There is a third way, raised in review, and unlike those two it does not name the namespace in the file that decodes at all:

```csharp
// one file, or a <Using Include="InertText.Encoding" /> item in the .csproj
global using InertText.Encoding;
```

Every other file in that project can then call `VisualEncoder.TryDecode` with no local mention of the namespace. The search still finds the import — it is still text in the repository — but it stops answering *which files* can decode and starts answering *which projects* can, and the file it points at is not the one doing the decoding. That is the one spelling where the capability is genuinely not visible where it is used.

That granularity is an invariant of the build rather than of the language, so it is gated rather than asserted: `NoProjectImportsTheCapabilityNamespaceForEveryFileAtOnce` fails on a `global using` or a `<Using>` item naming the capability namespace anywhere under `src`, and is mutation-gated both ways. Nothing needs one — production never names the namespace, and the tests that legitimately decode use ordinary per-file directives.

This is an **audit boundary, not a capability barrier**. A file can name the namespace either way, and nothing should stop it — decoding is a legitimate operation with legitimate callers. Assembly separation would not change this; `ProjectReference` is transitively compile-visible. What the boundary buys is that the reversing half cannot arrive unnoticed by a reviewer who searches correctly.

This PR migrates no caller, so its own producer count is 0 by construction. Measured on the stacked branch that does adopt it (#3563): **5 production files produce inert text, 0 name the capability namespace.**

## Three gates, each mutation-proved

| Gate | Mutation | Result |
| --- | --- | --- |
| Nothing public in the currency namespace returns text | add a `TryDecode` forwarder to `InertString` | red |
| The decoder exists, and only in the capability namespace | — pins that gate 1 cannot be satisfied by deleting invertibility | — |
| Every public creation path is self-contained | move `TextPolicy` into `InertText.Encoding` | red |
| No public member takes a delegate | add an `EnsurePermitted(Func<Rune, bool>)` overload | red |

Gate 1 enumerates and accounts for each text-returning member one by one, because the failure it guards against is an *addition* — a convenience overload handing back the decoded form — which an existence assertion would not notice.

Gate 3 is the one that makes the other two worth having, and it is why `InertString` has a public constructor rather than an encoder factory. If producing inert text required calling the encoder, every producer would name the capability namespace, the decoder would sit one member access away in exactly the files handling hostile input, and the audit search would return the whole producer set — surviving in form, worthless in practice. Without the constructor the measurement above would be 5 of 5 instead of 0 of 5.

The gate requires that *every* creation path is self-contained, not that one is: a caller uses the overload that fits its call site, so a single contaminated overload is a hole for everyone who reaches for it.

## Policy is a closed set

**This reverses the position this PR opened with.** `ScalarPolicy` began as a public delegate, on the argument that the existing predicates disagree for good reasons and should keep their own rules — `CSharpIdentifier.IsRenderingHazard` excludes most of `Cf` **on purpose**, because ZWJ is meaningful in emoji sequences and in Indic and Persian text and ECMA-334 permits `Cf` in `identifier_part`. That reasoning about `Cf` is still correct. The conclusion drawn from it was not.

An open policy parameter fails twice over, and both failures were measured rather than argued.

**It drifts.** Five predicates in this repository answer "is this scalar dangerous" and disagree by up to **48 BMP characters**:

| Predicate | BMP characters claimed |
| --- | --- |
| `CSharpIdentifierCore.IsRenderingHazard` | 76 |
| `InspectionResultView.NeedsEscape` (local function) | 76 |
| `MetadataTableProjector.IsControl` (private) | 65 |
| `TextPolicy.Field` | 110 |
| `TextPolicy.Prose` | 107 |

That drift has already shipped a defect. Calling `MetadataTableProjector.NeutralizeControls` on code identical to `origin/main`:

```text
ESC / ANSI        -> name\u001B[31mRED      escaped
Trojan Source     -> Utils<202E>gnp.exe      EMITTED VERBATIM
zero-width joiner -> Mic<200D>rosoft         EMITTED VERBATIM
BOM inline        -> Micro<FEFF>soft         EMITTED VERBATIM
line separator    -> line<2028>forged        EMITTED VERBATIM
```

The metadata layer — the *first* place hostile assembly metadata is rendered — escapes ANSI but not bidi, missing the attack class Trojan Source is named after. `InspectionResultView.NeedsEscape` is meanwhile a local copy of `IsRenderingHazard` with U+2028/2029 added, a patch that never propagated back. An open parameter does not cause this, but it is the shape in which it stays invisible: a rule written at one sink cannot be reached from the next.

**It leaks.** `EnsurePermitted` must decode before it re-encodes, or backslashes double. With a caller-supplied predicate, that decoded original is then passed to the caller, one scalar at a time:

```csharp
using InertText;                     // the only directive in the file

StringBuilder recovered = new();
value.EnsurePermitted(r => { recovered.Append(r); return false; });
// recovered now holds the hostile original, byte for byte
```

`grep -c 'InertText.Encoding'` on that file returns **0**. This is the audit boundary walked back out through a callback, and no reflection test over return types can see it, because the disclosure is an *argument* rather than a result. The leak fires when the repair path actually runs — the realistic trigger is an allow-shaped policy on a `Field` value, since `\` is not in a letters-only allow set.

`TextPolicy` is therefore a **closed enum** and `ScalarPolicy` is internal. Rules are written once and shared; no caller code runs during a repair. Adding a kind of text is an enum member and an arm in the rule table — **no new public member anywhere**, which is what makes the closed set extensible without growing the type.

`Identifier` is that member, and it is deliberately **not** in this PR: nothing here migrates `CSharpIdentifier`, and a policy with no caller is speculative surface. It arrives with the migration that needs it, and #3380's decision about `Cf` is preserved by adding it, not by overwriting `Field`.

For adoption risk, unchanged: for each existing predicate, the count of characters it claims that `Field` does not is **0**. `Field` is a strict superset, so routing a producer through this cannot lose an escape it currently makes.

### The set does not try to hold an allow-shaped rule

A homoglyph needs an allow list, and no `TextPolicy` catches one — a test pins this across every member. That is a boundary rather than a gap. **Encoding cannot serve an allow-shaped rule**, because the spelling it produces is itself outside the grammar: repairing `Nеwtonsoft.Json` to `N\u0435wtonsoft.Json` fails the same letters-only check that rejected the original. Such a rule wants **rejection**, which is a different answer than "here is a safe rendering", and the threat model already says reject rather than sanitize.

This was not reasoning ahead of the code. An earlier revision let a caller pass an allow-shaped predicate, and `EnsurePermitted` returned text its own policy rejected — silently. Adding an allow-shaped member back to the enum turns **three** tests red today, including the conformance sweep.

## Asking whether a value suits your sink

`EnsurePermitted` is public and is the only member shipped for this. It takes the policy the *sink* requires and returns a value that satisfies it, decoding first so it is idempotent.

It takes a `TextPolicy` rather than a predicate, which is what keeps the decoded original inside the library.

The direct construction a caller would otherwise reach for is a trap: `new InertString(policy, value.ToString())` re-encodes text that is already encoded, and backslashes double on every pass — `a\u202Eb` grows 8 → 9 → 11 → 15 characters across three rounds. The failure direction is over-encoding rather than a leak, so it is a legibility bug and not a security one, which is exactly why it would survive review.

Three overloads were designed and then **not** shipped — `EnsurePermitted(policy, out bool)`, and instance `IsPermitted` with and without a `ScalarViolation`. Every existing caller discards the bool, and instance `IsPermitted` has zero callers in any form. One member, three tests, no speculative surface.

### One shape at every entry point

Every public entry point reads `(TextPolicy, payload)` — the constructor, `Format`, `Join`, `IsPermitted`, `EnsurePermitted`. Named producers (`InertString.Field(value)`) were prototyped and rejected: they are noun-shaped methods for a verb-shaped job, they read asymmetrically against `EnsurePermitted(policy)`, and each new policy would force a new public member. The enum is the extension mechanism, so the type does not grow when the set does.

Naming it `TextPolicy` rather than `TextSink` also keeps the migration small: the `Format` and `Join` call sites are **character-identical**, because that call was already policy-first. Only the five `new InertString` sites in `NetworkTelemetry` swap arguments.

**`WasEncoded` does not answer this question, and is close to an anti-signal.** Proved in both directions: `Prose`-encoded `"line1\nline2"` reports `WasEncoded=false` and still **violates** `Field`; `Field`-encoded `"a\u202Eb"` reports `WasEncoded=true` and satisfies **both** policies, because its escapes are plain ASCII. Conformance is a relation between a value and a policy, not a property of the value — so it cannot be cached, and storing the producing policy would not help either, since knowing which delegate produced a value says nothing about whether it is stricter than yours. `EnsurePermitted` recomputes, and a test pins the anti-correlation so the tempting shortcut stays closed.

Today the policies barely meet: **13 `Field` sites, 1 `Prose`**, touching at one splice (`FeedFailureTelemetry`, `Field` lines joined as `Prose`) in the safe direction. The repair path has never fired in production. It is here because the splice exists, not because it currently fails.

## Why `readonly struct`

A class can enforce the invariant that a struct cannot — `default(InertString)` bypasses every constructor by CLR design — which is the stronger argument, and it lost to measurement.

`Encode` returns the **original string instance** when nothing needed encoding. Across the whole shared framework that is **471,968 of 471,968 metadata names, 181 assemblies, 0 rewrites — 100.0000% unchanged**. Containment is free precisely when nothing needed containing, and a class would put ~32 bytes of Gen0 garbage on exactly that path; `System.Private.CoreLib` alone has 144,854 name slots. `InertString?` is also 24 bytes as `Nullable<T>` against 16 for the value.

The same measurement rules out a `ReadOnlySpan<char>` overload: it would force `new string(span)` on the 100% path to save an allocation on the 0% path. The caveat is that this corpus is Microsoft-signed and therefore benign by construction — it bounds the *cost* of containment, not its value.

`_text` stays `string?` for a reason worth one line of review: a non-nullable annotation on a struct field is unverifiable — it compiles with zero warnings and still NREs on `default(T)`, because the compiler does not track `default(T)` through to fields. The null is confined to one accessor; one declaration, one assignment, one read, and exactly one `??` in the file.

## Corpus

22 named attacks — Trojan Source, tag smuggling, zero-width splitters, homoglyphs, inline BOM — asserting six claims per fixture plus injectivity. It names what it caught: mutating `Cf` out of the hazard set turns it red **by attack name**, not by code point.

## Details worth a reviewer's attention

- **Scalars, not code units.** `Rune.DecodeFromUtf16` with `OperationStatus`, not `EnumerateRunes`, which substitutes `U+FFFD` for unpaired surrogates and is therefore lossy on exactly the class `Cs` exists for. Every `char`-based predicate in the tree is structurally blind to astral payloads such as tag smuggling.
- **`IsPermitted` stays outside the capability namespace.** It returns `bool` plus a `ScalarViolation(int index, int scalar, UnicodeCategory)` — an index and a code point, never the character — so a survey can name what it refused without echoing it. Moving it would force validate-only callers to take on decode and dilute the signal.
- **Backslash is an encoder property, not a policy one.** Without rewriting it, `\u0041` in a real name is indistinguishable from an encoded escape. It is not in any deny set.
- **The decoder refuses spellings the encoder never emits** — `\U` for a BMP scalar, `\uXXXX` for a scalar with a canonical short form such as `\\`, `\^X` or `\^?`, lowercase hex in either width, and a raw unpaired surrogate. One scalar, one encoding, in both directions. A **lone** surrogate escape stays legal, and the distinction matters: `Encode` genuinely emits that form, and an unpaired surrogate has no other representation at all, so refusing the whole surrogate range would make the encoder non-invertible on the one input class that needs it most.
- **An already-inert value reached through a generic parameter or through `object` is not encoded twice.** Overload resolution binds on the *static* type of an interpolation hole, so `AppendFormatted(InertString)` is invisible from inside a generic method and `AppendFormatted<T>` wins — handing already-encoded text back to the encoder. The trap was known and closed for `InertString` and `InertString?`; neither overload can reach the generic or boxed case, and only a type test can.
- **No `ProjectReference`.** This sits below every assembly that prints artifact-derived text, including the dependency-free leaves. A reference added here is a reference added to everything.

## Evidence

`InertText.Tests`: **190 tests, 0 failures**, Release. Library builds clean with 0 warnings (`TreatWarningsAsErrors` is set in `src/Directory.Build.props`, so it covers everything under `src/`; three spoof-fixture projects opt out). CI runs the suite unconditionally rather than behind a path filter — the gates are a BMP sweep and a round-trip, and are fast.

`ci.yml` carries one unrelated fix, called out so it is not mistaken for part of the primitive: the `pull_request` base filter was `[main]`, which also decides whether a *stacked* PR runs CI at all. #3563 was retargeted onto this branch and stopped matching, so it reported **no check-runs** rather than failing ones — invisible on the PR page, and indistinguishable from "not run yet" once `ci-required` is required. Adding `feature/**` restored it; #3563 now reports 8 checks.

## Review

Two-model tier per `AGENTS.md`: security-relevant code on a default output path.

A two-model round on the previous head found five defects. Three are fixed here (the callback disclosure, the generic/boxed double-encode, the non-canonical decoder spellings) and each new test is mutation-gated — the fix is reverted and the test observed to fail. Two were closed as **not defects**, with reasons:

- **`Prose` permits a lone `CR`.** Both reviewers called this blocking; it is not, and removing it would be worse. `Join` encodes its separator on purpose — the same reason the handler encodes source literals — so refusing `CR` would render `Environment.NewLine` as `\^M` on Windows for every caller who joins with it. The two cannot both hold, and the only production `Prose` site joins `Field`-encoded parts, where widening does not un-escape.
- **A proposed fix for the surrogate spelling was wrong.** Rejecting every surrogate-range `\uXXXX` would break the lone-unpaired-surrogate round-trip the encoder actually emits.

A second two-model round on the fixed head found one more, and it was caused by the round
before it. The rule added above — refuse a decoded high surrogate *immediately followed by* a
decoded low — made a value this library had just produced fail to decode. `Join` and the
interpolation handler concatenate fragments that were each encoded alone, so a lone high
surrogate encoded in one fragment lands beside a lone low surrogate encoded in the next.
`EnsurePermitted` falls back to treating undecodable text as raw, so the repair encoded the
escapes a second time and `\uD834\uDD73` became `\\uD834\\uDD73` — which is also what the
ASCII text a user typed encodes to. Two unrelated inputs converged on one output, and the
rendered form stopped being evidence of which one arrived.

The premise of the refusal did not hold. A .NET string is UTF-16, so `"\uD83D" + "\uDE00"` *is*
`"\U0001F600"`: there is no string in which the halves stay apart, and no other text the
spelling could denote. Refusing it rejected an input rather than disambiguating two. That
spelling now decodes, and because `Encode` still only ever emits `\U` for a pair, accepting it
is an input form rather than a second output form. Three regression tests cover it, all
mutation-gated against the exact prior decoder. A 300,000-case fuzz over composed values —
alphabet loaded with backslashes, `u`, `U`, `^`, hex digits and surrogates — now reports zero
undecodable compositions, zero round-trip mismatches, and zero repairs that return text their
target policy rejects.

### Third review round

An independent review found the delegate gate weaker than its remarks claimed, and I reproduced
it before fixing: inside a loop over every exported type it enumerated
`typeof(InertString).GetMethods`, so a delegate-taking `AppendFormatted` on `InertStringHandler`
left it green. The handler is precisely where an escape-valve predicate would land. The original
mutation proof used an `EnsurePermitted` overload on `InertString` — the one type the broken scan
did cover — so the proof passed while the gate was weaker than documented. Fixed to
`type.GetMethods`, re-proven with the handler mutation.

The project-wide-import gate scanned only `src/` for `.cs`/`.csproj`/`.props`, missing the repo
root's `Directory.Build.props`/`.targets`, which flow into every project under `src/`. That was
the widest-reaching version of the hazard the gate exists for. Now scans root-level
`Directory.Build.*` and `.targets`, and matches the `global::` spelling. Both mutation-proven.

Three doc comments were stale against the shipped design (the decoder's pair-rejection claim, the
corpus's caller-chosen-predicate claim, and a renamed `paramref`). Rather than sweep them and
leave the class open, `GenerateDocumentationFile` is now on for `InertText`, making CS1574/CS1734/
CS0419 build errors — which immediately caught an ambiguous `<see cref="IsPermitted"/>` the review
had not found. Zero CS1591: every public member was already documented.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,14 @@ name: CI
 on:
   push:
     branches: [main]
+  # The base-branch filter also decides whether a *stacked* PR gets CI at all.
+  # A PR retargeted from `main` onto a long-lived feature branch silently stops
+  # matching `[main]`, so it reports no checks rather than failing ones -- and
+  # once `ci-required` is a required check, "no checks reported" is
+  # indistinguishable at a glance from "not run yet". `feature/**` keeps the
+  # lower half of a stack covered while its base is still in review.
   pull_request:
-    branches: [main]
+    branches: [main, 'feature/**']
 
 env:
   DOTNET_NOLOGO: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -384,6 +384,12 @@ jobs:
       - name: Run NuGetFetch tests (offline)
         run: dotnet run --project src/NuGetFetch.Tests -c Release -- -trait- "Network=Live"
 
+      # InertText sits below every other project, so a regression here reaches every assembly
+      # that prints artifact-derived text. The gates are pure and fast — a full BMP sweep and a
+      # round-trip — so they run unconditionally rather than behind a path filter.
+      - name: Run InertText tests
+        run: dotnet run --project src/InertText.Tests -c Release
+
       # Fast decompiler unit tests: pass logic, printer, importer facts, identity,
       # and classification regressions. Slow compile-back/recompile, corpus-sweep,
       # and fidelity tests are tagged [Trait("Speed", "Slow")] and excluded here;
--- a/docs/design/inert-text.md
+++ b/docs/design/inert-text.md
@@ -1,0 +1,186 @@
+# InertText
+
+A shared component for making untrusted text inert before it reaches a sink that
+can act on it.
+
+## The problem
+
+Text that comes out of an artifact — a package id, a type name, a file path, an
+assembly reference — is attacker-controlled. Printing it hands that control to
+whatever renders it. The attacks do not need malformed input; every payload in
+this component's test corpus is well-formed Unicode that a naive sink prints
+without complaint:
+
+- **Reordering.** `Cf` scalars drive the bidi algorithm. Trojan Source
+  (CVE-2021-42574) uses them to make source render differently from how it
+  compiles, and a right-to-left override in a name reorders the rest of the line.
+  `invoice\u202Egpj.exe` renders as `invoice.jpg`.
+- **Terminal control.** `Cc` scalars introduce escape sequences that clear the
+  screen, rewrite the window title, or make text a hyperlink to somewhere else.
+- **Line forgery.** A line terminator inside a value produces a second log line
+  that looks like the tool emitted it. `CR` alone rewinds so the truth is
+  overwritten. `NEL`, `LS` and `PS` are terminators that a `CR`/`LF` check misses.
+- **Invisibility.** Zero-width scalars, soft hyphens and word joiners split a
+  name for the eye while leaving it one token to a comparison. Tag characters
+  carry an invisible ASCII payload, which is how instructions get hidden in text
+  an agent will read.
+
+Percent-encoding, `Uri` normalization and HTML escaping each stop some of this
+and none of it. `Uri` percent-encodes `Cc` and passes `Cf` straight through.
+
+## What it does
+
+Refused scalars are rewritten in a visible spelling. The term and the contract
+come from BSD `vis(3)`: the output is **inert**, **lossless** and **invertible**.
+
+```text
+invoice\u202Egpj.exe   ->   invoice\u202Egpj.exe    (the override, spelled out)
+package<LF>Error: x    ->   package\^JError: x
+```
+
+This is not neutralization, which has none of the three properties. Dropping the
+hostile scalars would also produce safe output, but it destroys the evidence —
+and the reader is usually trying to work out what the artifact actually says.
+
+## The predicate and the speller are separate
+
+This is the central design split, and it is what lets one speller serve every
+sink:
+
+- The **predicate** (`ScalarPolicy`) answers "is this scalar permitted *here*". It
+  is per-sink and it is the caller's to choose. `TextPolicy.Field` is deny-shaped
+  and refuses `Cc`, `Cf`, `Cs`, `Zl` and `Zp`. `TextPolicy.Prose` is the same
+  minus a `CR`/`LF`/`TAB` exemption, for genuinely multi-line text.
+- The **speller** (`VisualEncoder`) answers "how is a refused scalar written
+  down". It is total over Unicode, and it never learns *why* a scalar was
+  refused.
+
+A speller with a built-in hazard set has absorbed policy and cannot serve a sink
+whose grammar is constrained. That case is not hypothetical: the central
+typosquatting vector is a homoglyph, and Cyrillic `а` and Latin `a` are the same
+glyph, both category `Ll`. No category rule catches that and none should —
+refusing every non-Latin letter would break most of the world's text. Only an
+allow-shaped policy catches it, and because the speller is total, such a sink
+needs no spelling table of its own.
+
+The speller's obligations:
+
+- **Total.** Defined on every Unicode scalar, including the 127 encoded scalars
+  above the BMP. A `\uXXXX`-only speller is neither total nor invertible on
+  exactly the inputs an attacker reaches for.
+- **Injective, in both directions.** One scalar has one spelling, so the decoder
+  refuses spellings the encoder never emits: `\U` for a BMP scalar, and `\uXXXX`
+  for a scalar with a canonical short form such as `\\`, `\^X` or `\^?`.
+- **Scalar-based, not code-unit-based.** An unpaired surrogate is not a scalar.
+  `Rune` cannot hold one and `EnumerateRunes` substitutes `U+FFFD`, so a speller
+  built on either is lossy on the one input that cannot be written any other way.
+- **Able to report what it emitted**, so a caller can print a legend without
+  keeping a second copy of the spelling table.
+- **Not responsible for structural escaping.** A `|` still breaks a Markdown
+  cell and a `"` still ends a JSON string. Neither is in an encoded category, and
+  neither should be — escaping for a grammar is the serializer's job. Visual
+  encoding and structural escaping compose; neither substitutes for the other.
+
+## Treated text is carried as a type
+
+Encoding on its own is transactional: a `string` goes in and a `string` comes
+out, so a treated value and an untreated one have the same type. Nothing stops a
+later edit from interpolating the untreated one, and nothing marks the difference
+at the sink. Auditing that means tracing every path that reaches a printer, again
+after every change.
+
+`InertString` is the currency form. It can only be built by applying a policy —
+there is no conversion *from* `string`, and a test asserts that no public entry
+point takes text without also taking a policy. A sink that accepts only this type
+cannot be handed raw text by accident, which answers **what a sink accepts** with
+a type search rather than a call-graph trace.
+
+Conversion *to* `string` is unrestricted, which is safe here in a way it usually
+is not. The customary objection to a wrapper is that `ToString` launders it, but
+that assumes the payload is dangerous and the wrapper is what holds it back. Here
+the payload is already inert. Losing the wrapper loses provenance, not
+protection.
+
+## Holding a value is not the same as being able to reverse it
+
+A type answers what a *sink* accepts. It does not answer what a *file* can do,
+and with the decoder as a member of the value type it cannot: every file holding
+a value would be one member access away from the original, so a search for the
+type name finds mentions rather than capabilities.
+
+So the decoder lives in its own namespace:
+
+| Namespace | Contains | A file that imports it can |
+| --- | --- | --- |
+| `InertText` | `InertString`, `ScalarPolicy`, `TextPolicy`, `VisualForm` | build, compose, compare and print inert text |
+| `InertText.Encoder` | `VisualEncoder` | additionally recover the original text |
+
+Text enters through `InertString`'s constructor and leaves through `ToString`
+already spelled, so the currency namespace is sufficient for every ordinary use.
+A file that imports `InertText` and not `InertText.Encoder` has no path back to
+the original of any value it handles — and that is legible in its import list.
+
+A reflection test enumerates every public member of the `InertText` namespace
+that returns text and accounts for each one: `ToString` (the encoded form),
+`DescribeLegend` (fixed strings naming spellings) and `ScalarViolation.ToString`
+(an index, a code point and a category — which is why a violation reports an
+`int` rather than the character, so a survey can name what it refused without
+echoing it). Adding a decode convenience to the currency type fails that test.
+
+**This is an audit boundary, not a capability barrier.** A file can add the
+import or write the name out in full, and nothing should stop it — decoding is a
+legitimate operation with legitimate callers. What the boundary buys is that the
+reversing half cannot arrive unnoticed.
+
+## Composition
+
+Composition has to work, or callers fall back to `$"...{treated}..."` and drop
+the guarantee at the moment it matters most. `InertString.Format` is an
+interpolated string handler that encodes each part as it is appended — holes
+because they are untrusted, and literals too, because an invariant with an
+exception in it has to be re-argued at every use.
+
+An already-inert hole is not encoded twice, but neither is it trusted. The type
+records that *a* policy was applied, not *which* one, and a value built for one
+sink is routinely spliced into a message bound for another. `Prose` permits the
+line feed that `Field` exists to remove, so splicing a `Prose` value into a
+`Field` message unexamined would put a raw newline into a single-line record and
+report no encoded forms for it — log injection, with the type appearing to vouch
+for it.
+
+Splices are therefore re-checked against the policy in force and re-spelled where
+they do not satisfy it, in `Format` and `Join` alike. That repair is the second
+thing invertibility buys, and it is worth noticing that the requirement is
+unsatisfiable without a decoder: a mismatched part can only be repaired if the
+original can be recovered exactly.
+
+The repair only ever **tightens**. A value encoded under a stricter policy keeps
+its spellings when spliced into a laxer sink, because composition making a value
+*less* inert would let a caller launder one by quoting it somewhere permissive.
+The cost is that the splice path is observable — the same source text can render
+differently depending on where it was encoded — which is a deliberate trade.
+
+## Testing
+
+Two shapes, because they fail differently.
+
+**Property tests** sweep every scalar for invertibility and injectivity. They
+prove the transform is total, and a regression reports that some scalar broke.
+
+**A corpus of named attacks** records what the properties are *for*. A regression
+there reports that Trojan Source works again, which is a sentence someone can
+act on. Each fixture records what it attacks, and six claims are asserted over
+every one of them, so adding a fixture is a single line that is immediately
+subject to all of them.
+
+The corpus also carries a payload nothing catches — the Cyrillic homoglyph — with
+both halves asserted: the category policy permits it, an allow-shaped policy
+refuses it. A corpus containing only what the default catches would quietly imply
+the default is sufficient.
+
+## Placement
+
+`InertText` sits below every other project and references nothing. Every assembly
+that prints artifact-derived text needs it, including the dependency-free leaves,
+so it has to sit below all of them. A reference added to it is a reference added
+to everything.
--- a/docs/design/inert-text.md
+++ b/docs/design/inert-text.md
@@ -279,6 +279,36 @@ its spellings when spliced into a laxer sink, because composition making a value
 The cost is that the splice path is observable — the same source text can render
 differently depending on where it was encoded — which is a deliberate trade.
 
+### Asking whether a value suits your sink
+
+That repair is exposed as `EnsurePermitted(ScalarPolicy)`, because a sink that
+accepts an `InertString` has no other correct way to make one safe for itself.
+The obvious substitute is wrong: `new InertString(value.ToString(), policy)`
+hands already-encoded text back to the encoder, so the backslashes double on
+every pass — `a\u202Eb` becomes `a\\u202Eb`, then `a\\\\u202Eb`. Because
+`EnsurePermitted` decodes before re-spelling, repeating it is a no-op.
+
+What a sink must **not** use for this is `WasEncoded` or `Forms`. Those report
+what was *done* to a value, never what it *satisfies*, and as a conformance
+check they are wrong in both directions:
+
+| Value | `WasEncoded` | Satisfies `Field` |
+| ----- | ------------ | ----------------- |
+| `Prose`-encoded `"line1\nline2"` | `false` | **no** — it carries a raw line feed |
+| `Field`-encoded `"a\u202Eb"` | `true` | yes, and `Prose` too |
+
+Encoding makes a value *more* conformant, so the flag is close to
+anti-correlated with the property a sink cares about. The underlying reason is
+that conformance is a relation between a value and a policy rather than a
+property of the value, which is why it cannot be cached on one and why
+`EnsurePermitted` has to recompute it.
+
+Storing the producing policy on the value would not help either. Knowing which
+delegate produced a value says nothing about whether it is stricter than the
+delegate you are about to apply, short of sweeping every scalar to compare their
+permitted sets — so the value would carry an extra field and still pay for the
+scan.
+
 ## Testing
 
 Two shapes, because they fail differently.
--- a/docs/design/inert-text.md
+++ b/docs/design/inert-text.md
@@ -252,22 +252,32 @@ So the decoder lives in its own namespace:
 Text enters through `InertString`'s constructor and leaves through `ToString`
 already spelled, so the currency namespace is sufficient for every ordinary use.
 
-**The audit is one search, and the string is `InertText.Encoder`.** Not
-`using InertText.Encoder` — a using directive is only one of the two ways to
-reach a namespace, and the other one leaves the import block untouched:
+**The audit is one search, and the string is `InertText.Encoder`.** Reaching the
+decoder means naming its namespace, and that act *is* the opt-in. Decoding is a
+legitimate operation with legitimate callers, so the design does not try to
+prevent it — only to make it impossible to do quietly.
+
+A namespace can be named in two places, and both are equally legitimate:
 
 ```csharp
-using InertText;                    // the only directive in the file
+using InertText.Encoder;            // named in the import block
+VisualEncoder.TryDecode(inert.ToString(), out string? original);
+```
 
+```csharp
+// no using directive of any kind in this file
 InertText.Encoder.VisualEncoder.TryDecode(inert.ToString(), out string? original);
 ```
 
-That compiles and recovers the original. A reviewer who greps for the directive
-sees a clean import list and concludes the file cannot decode, which is the
-opposite of the truth. Grepping the bare namespace catches both forms, because a
-fully-qualified call has to spell the namespace too.
+Neither is evasion. The second names the namespace at the call site rather than
+at the top of the file, in plain sight on the line that uses it. What matters is
+that the *audit* covers both, which is why the search is for the bare namespace
+and not for `using InertText.Encoder`: a fully-qualified call declares itself on
+its own line and needs no directive at all, so a reviewer grepping only the
+import block would read that file as clean and conclude it cannot decode.
 
-There is a third way, and it spells the namespace in a different file:
+There is a third way, and unlike those two it does not name the namespace in the
+file that decodes at all:
 
 ```csharp
 // one file, or a <Using Include="InertText.Encoder" /> item in the .csproj
--- a/docs/design/inert-text.md
+++ b/docs/design/inert-text.md
@@ -247,12 +247,12 @@ So the decoder lives in its own namespace:
 | Namespace | Contains | A file naming it can |
 | --- | --- | --- |
 | `InertText` | `InertString`, `TextPolicy`, `VisualForm` | build, compose, compare and print inert text |
-| `InertText.Encoder` | `VisualEncoder` | additionally recover the original text |
+| `InertText.Encoding` | `VisualEncoder` | additionally recover the original text |
 
 Text enters through `InertString`'s constructor and leaves through `ToString`
 already spelled, so the currency namespace is sufficient for every ordinary use.
 
-**The audit is one search, and the string is `InertText.Encoder`.** Reaching the
+**The audit is one search, and the string is `InertText.Encoding`.** Reaching the
 decoder means naming its namespace, and that act *is* the opt-in. Decoding is a
 legitimate operation with legitimate callers, so the design does not try to
 prevent it — only to make it impossible to do quietly.
@@ -260,19 +260,20 @@ prevent it — only to make it impossible to do quietly.
 A namespace can be named in two places, and both are equally legitimate:
 
 ```csharp
-using InertText.Encoder;            // named in the import block
+using InertText.Encoding;            // named in the import block
 VisualEncoder.TryDecode(inert.ToString(), out string? original);
 ```
 
 ```csharp
-// no using directive of any kind in this file
-InertText.Encoder.VisualEncoder.TryDecode(inert.ToString(), out string? original);
+// no using directive of any kind in this file. InertText.Encoding here is the
+// namespace; there is no InertText type and no Encoder member on InertString.
+InertText.Encoding.VisualEncoder.TryDecode(inert.ToString(), out string? original);
 ```
 
 Neither is evasion. The second names the namespace at the call site rather than
 at the top of the file, in plain sight on the line that uses it. What matters is
 that the *audit* covers both, which is why the search is for the bare namespace
-and not for `using InertText.Encoder`: a fully-qualified call declares itself on
+and not for `using InertText.Encoding`: a fully-qualified call declares itself on
 its own line and needs no directive at all, so a reviewer grepping only the
 import block would read that file as clean and conclude it cannot decode.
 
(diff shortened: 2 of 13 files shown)


richlander:
## Two-model review round at `f446837a9`

Per `AGENTS.md` this is the high-risk tier, so: two rounds, two different families,
neither of them my own. **Gemini 3.1 Pro** and **GPT-5.6 Sol**, each isolated in its own
read-only linked worktree, given the same self-contained prompt.

Scope was the **full diff against `main`**, not an incremental slice. The previous
InertText round landed at `26ed34d4`, which is *not an ancestor* of this head — the branch
was split and restacked since — so an incremental review would have been built on a review
of code that no longer exists in that shape. Base `ec7b254ef` is exactly the current tip of
`main`, so there is no merge skew. 2,698 insertions, 13 files.

**Verdict: not ready to merge.** Five distinct findings, all reproduced independently by me
on a clean worktree at the exact head before being accepted here. The suite is green
(171/171, 0 warnings) and stayed green through every one of them, which is the point.

| # | Finding | Found by | I reproduced | Severity |
| 1 | `ScalarPolicy` callback hands back the decoded original | GPT | yes | **blocking — falsifies the headline claim** |
| 2 | `EnsurePermitted` returns text its own policy rejects | GPT | yes | significant |
| 3 | Generic/boxed `InertString` double-encodes | **both** | yes | significant |
| 4 | `Prose` permits a lone CR | **both** | yes | significant, *not* blocking — see below |
| 5 | Decoder accepts four spellings the encoder never emits | **both** | yes | significant |

Both models converged independently on 3, 4 and 5. GPT alone found 1 and 2, and 1 is the
one that matters.

---

### 1. The audit boundary does not hold — the policy callback returns the plaintext

This falsifies the claim this PR is built on. `EnsurePermitted` decodes first, then passes
every recovered scalar to a **caller-supplied delegate**. So a file importing only
`InertText` recovers the original byte-for-byte without ever naming `InertText.Encoder`:

```csharp
using InertText;                      // the only directive

string original = "\u202Esecret\nline\u0007";
InertString inert = new(original, TextPolicy.Prose);

StringBuilder leaked = new();
bool first = true;
inert.EnsurePermitted(r => { if (first) { first = false; return false; } leaked.Append(r); return false; });
```

```text
original : <202E>secret<000A>line<0007>
inert    : \u202Esecret<000A>line\^G
leaked   : <202E>secret<000A>line<0007>
>>> EXACT MATCH TO ORIGINAL? True
```

The PR argues "the audit is one search, and the string is `InertText.Encoder`." It is not.
The surface-gate test at `InertStringTests.cs:536` reflects over **return and out types**,
so it is structurally incapable of seeing disclosure through an inverted callback — the gate
passes, and the property it names is false.

Worth being precise about what this is and is not. The threat model is hostile *artifact
metadata*, not hostile first-party code; nobody writes a capturing `ScalarPolicy` by
accident. So this is not "a package escapes containment." It is that **the audit search
fails open** — which is exactly the failure mode called out when the search string was
corrected earlier in this stack, and the reason the namespace split was thought to buy
anything.

It is also not a bug that can be patched at a call site. "Policy is a delegate" and "the
only way back is through the capability namespace" are in direct tension: any repair that
decodes must show the decoded scalars to the policy. That is a design decision, not a fix.

### 2. `EnsurePermitted` returns text that its own policy rejects

The contract is that the result satisfies the policy passed in. For **allow-shaped**
policies it does not — and allow-shaped is the shape `TextPolicy`'s own remarks recommend
("should supply an allow-shaped `ScalarPolicy` of its own, which is the only thing that
catches a homoglyph typosquat").

```text
Field-encoded                : <0435>vil          (Cyrillic e)
EnsurePermitted(LettersOnly) : \u0435vil
IsPermitted(result, LettersOnly)? False
```

Root cause: the escape spelling is built from `\`, `u` and hex digits, and an allow-list
that permits only letters rejects its own escape alphabet. Encoding into a form a policy
accepts requires that policy to accept the spelling alphabet — which no allow-list written
for typosquat defense will.

### 3. A generic or boxed `InertString` is double-encoded

The dedicated `AppendFormatted(InertString)` overload only binds when the *static* type is
`InertString`. Through a generic parameter or `object`, `AppendFormatted<T>` wins, calls
`ToString()` on already-encoded text and hands it back to the encoder:

```text
direct           : arg: \u202Eevil
via generic T    : arg: \\u202Eevil
via object       : arg: \\u202Eevil
```

The remarks on `AppendFormatted(InertString?)` show this trap was known and closed for
`InertString` and `InertString?` — the generic and boxed cases were missed. Failure
direction is over-encoding, so it is legibility rather than leak. Mechanical fix: a runtime
`is InertString` check in both generic overloads.

### 4. `Prose` permits a lone CR — real, but I am downgrading it from blocking

Both models called this critical. The behaviour is real: `Prose` passes raw `\r`, and a lone
CR returns the cursor to column 0, so hostile text can overwrite a trusted prefix.

It is **not reachable on any current path**, which I checked rather than assumed. The only
production `Prose` site is `Join(Environment.NewLine, Prose, lines)` in
`FeedFailureTelemetry`, where every part is already `Field`-encoded and the separator is a
developer literal. I tested the widening directly, since `EnsurePermitted` decodes first and
`Prose` is laxer:

```text
Field-encoded          : Package not found\^MSUCCESS: signature verified
EnsurePermitted(Prose) : Package not found\^MSUCCESS: signature verified
raw CR restored?       : no — stays escaped
```

It does not un-escape, because `EnsurePermitted` early-returns when the value already
satisfies the policy. So this is a **latent trap for the next adopter** who hands raw
foreign text to `Prose`, not a live hole.

Both models proposed "never permit raw CR." That has a cost neither noticed: `Join` encodes
the **developer-supplied separator**, so with CR refused, `Environment.NewLine` renders as
`\^M` on every joined line on Windows. And `ScalarPolicy` is a *per-scalar* predicate — it
cannot see context, so "permit CR only in CRLF" is not expressible. The real options are
drop CR and use `"\n"`, or stop encoding the separator and treat it as trusted structure.

### 5. The decoder accepts four spellings the encoder never emits

Documented property: "one scalar, one encoding, in both directions." All four are
non-canonical — accepted by `TryDecode`, and re-encoding does not reproduce the input:

```text
lowercase BMP hex     \u00ad        -> re-encodes to \u00AD
lowercase astral hex  \U0001f600    -> re-encodes to \U0001F600
raw unpaired surrogate <D800>       -> re-encodes to \uD800
surrogate-pair spelling \uD83D\uDE00 -> re-encodes to \U0001F600
```

**Gemini's suggested fix is wrong and I am not taking it.** It proposed rejecting `\uXXXX`
whenever the value is in the surrogate range. The encoder *does* emit exactly that form for
a lone unpaired surrogate, and it round-trips today — `VisualEncoder.cs:192-195` says so in
a comment, and I confirmed it. That fix would break invertibility on the one input class
that has no other representation. The correct rule is narrower: reject a decoded high
surrogate *immediately followed by* a decoded low surrogate, plus require uppercase hex, plus
reject raw unpaired surrogates in decoder input.

---

### What each model got wrong

Recorded because the reconciliation is only worth reading if it is honest in both directions.

- **Gemini** rated `Prose` blocking without checking whether any path reaches it, and its
  surrogate fix would have introduced a real invertibility bug. Its claim-3 and claim-5
  verifications matched mine.
- **GPT** also rated `Prose` blocking. Everything else it reported reproduced exactly,
  including both findings Gemini missed, and its sweep (1,112,064 scalars, 2,048 unpaired
  surrogates, 0 round-trip failures) matches the encoder-side claims.
- **Both** confirmed the encoder-side invariants hold: totality, `default(InertString)`
  safety, encoder round-trip and injectivity, `ci-required` job dependencies intact, and no
  `pull_request_target` exposure from the trigger widening.

### Disposition

Findings 3 and 5 are mechanical. Finding 4 is a policy decision with a known cost. Findings
1 and 2 are design questions about whether `ScalarPolicy` can be an arbitrary delegate and
still support the audit and conformance guarantees this library advertises — and finding 1
decides whether the namespace split buys what the PR says it buys.

Not marking this ready. No fixes pushed yet; the next commits will be linked here and the
fixed head re-reviewed per `AGENTS.md`.

richlander:
### Finding 1, reproduced without the contrivance

GPT's PoC forced the repair path with a flag that refuses the first scalar, which
reads as artificial. It is not necessary. The leak sits on the **documented
recommended usage**.

`TextPolicy`'s own remarks say a field whose grammar is externally defined "should
supply an allow-shaped `ScalarPolicy` of its own, which is the only thing that
catches a homoglyph typosquat." Do exactly that, on a `Field` value, and the
repair path runs on its own — because `\` is not in the allow set, so the encoded
value genuinely does not satisfy the new policy.

```csharp
using System.Text;
using InertText;               // the only InertText import in the file

InertString contained = new("Utils\u202Egnp.exe", TextPolicy.Field);

StringBuilder seen = new();
ScalarPolicy packageIdPolicy = scalar =>
{
    bool ok = scalar.Value is (>= 'a' and <= 'z') or (>= 'A' and <= 'Z')
                           or (>= '0' and <= '9') or '.' or '-' or '_';
    seen.Append(scalar);       // an ordinary "what did I inspect?" diagnostic
    return ok;
};

InertString repaired = contained.EnsurePermitted(packageIdPolicy);
```

```text
original                     : Utils<202E>gnp.exe
contained (Field)            : Utils\u202Egnp.exe
scalars handed to the policy : Utils\Utils<202E>gnp.exe
>>> did the policy see the DECODED original? True

result                                   : Utils\u202Egnp.exe
result satisfies the policy it was given? False
```

Two things in one call.

The policy is invoked twice: first over the *encoded* text, where it stops at the
backslash, and then — after `EnsurePermitted` decodes — over the **original
scalars**, handing the raw `U+202E` to caller code. `grep -c 'InertText.Encoder'`
on that file returns **0**. Its only imports are `System.Text` and `InertText`.

The same run shows finding 2: the returned value does not satisfy the policy that
was just passed in.

So the exposure is not "someone writes a deliberately capturing delegate." It is
"someone follows the guidance in `TextPolicy`'s remarks." Any cross-policy
`EnsurePermitted` where the value does not already conform will show the policy
the decoded original, and non-conformance is the only reason to call it.

For completeness, the case that does **not** leak: when the value already
satisfies the policy, `EnsurePermitted` early-returns and the delegate only ever
sees encoded text. That is why finding 4 is not live today, and it is the same
mechanism.

richlander:
## Review

**Verdict: approve after one test fix.** The library code is correct as far as I can verify it — I re-derived the decoder's canonicality rules, the surrogate handling, the splice/repair paths, and the `default(InertString)` story against the diff and found no defect. I ran the suite in an isolated worktree: **190/190 pass in 0.35s** (the PR body says 189 — trivially stale). But one of the three headline gates does not do what its remarks claim, and I mutation-proved it.

### Confirmed defect: the delegate gate doesn't scan the handler

`NoPublicMemberOfTheCurrencyNamespaceTakesADelegate` (InertStringTests.cs) loops over every exported type in the namespace but inside the loop enumerates `typeof(InertString).GetMethods(Declared)` instead of `type.GetMethods(Declared)` — so it checks `InertString`'s methods repeatedly plus each type's *constructors*, and never any other type's methods. The remarks explicitly promise the opposite: *"reintroducing a predicate anywhere — as an overload, **on the handler**, as an optional escape valve — fails here."*

Verified by mutation: adding `public void AppendFormatted(string? value, Func<Rune, bool> permitted)` to `InertStringHandler` leaves the gate **green**. That is not a hypothetical spot — the handler is precisely where an escape-valve predicate would land, and it is the type that touches decoded-adjacent text during splices. The fix is one line — `type.GetMethods(Declared)` — and I confirmed it turns the mutation red and stays green on the clean tree. Note the PR's own mutation table proved this gate red via an `EnsurePermitted` overload on `InertString`, which is the one type the buggy scan does cover — so the proof passed while the gate is weaker than documented. Suggest re-proving with the handler mutation after the fix.

### Gate scope gap: the global-using gate misses the root props file

`NoProjectImportsTheCapabilityNamespaceForEveryFileAtOnce` scans `src/**` for `.cs`, `.csproj`, `.props`. But the repo has a root `Directory.Build.props` — outside `src/`, and it applies to every project under `src/`. A `<Using Include="InertText.Encoder"/>` added there evades the gate entirely while giving every file in every project the import. Same for the `tests/` tree (defensible under the production-only claim, but the root props file reaches production projects). Suggest scanning from the repo root, or at minimum adding root-level `Directory.Build.props`/`.targets` to the candidate list. Minor variants also slip the `StartsWith` match (`global using global::InertText.Encoder;`); probably fine to accept as best-effort, but the root-props hole is worth closing since the test exists precisely because this granularity "is an invariant of the build."

### Stale documentation — three spots contradict the shipped design

1. **`VisualEncoder.TryDecode`, trailing comment** (just before `value = builder.ToString()`): *"Two escapes forming a pair are rejected in that arm, which is what keeps accepting the lone case injective."* That is the pre-fix behavior — the final review round reversed it, the `'u'` arm's own comment says pairs decode to the astral scalar, and `Decode_AcceptsASurrogatePairSpelledAsTwoEscapes` pins it. Two comments in one method now assert opposite behaviors.
2. **`AdversarialCorpus.Homoglyph` remark**: *"the fix is an allow-shaped policy … which is why the predicate is the caller's to choose."* The "Policy is a closed set" section is the reversal of this; the delegate gate exists to forbid it. Leftover from the open-predicate revision.
3. **`EnsurePermitted` summary**: `<paramref name="permits"/>` names a parameter that no longer exists (renamed to `policy`). Silent today because `GenerateDocumentationFile` is off; it would be CS1734 otherwise.

Also: the `InertTextCanBeProducedWithoutNamingTheCapabilityNamespace` remark says "four production files … would make that four out of five" while the PR body says 5 and 5-of-5 — one of them is wrong, and per the repo's present-only-docs convention the stacked-branch measurement probably belongs in the PR body rather than the source.

### Nits

- Two stray-indentation glitches: the `IsPermitted(TextPolicy, string)` doc block in InertString.cs and the `for` opening `TryDecode`'s loop are both over-indented by ~8 spaces.
- `NoPublicMemberOfTheCurrencyNamespaceReturnsText`'s `IsTextual` won't see a future `Span<char>`/`ReadOnlySpan<char>`-typed return (its generic-argument check looks only for `string`). Not a live hole; noting it since the gate's value is catching additions.

### What I verified and found sound

- **Decoder/encoder injectivity**: the leniencies I could find (raw astral `Cf` pair in decoder input; `A`-style escapes for graphic scalars) are all unreachable from library-produced values — Encode doubles backslashes and never leaves a refused scalar raw — so `EnsurePermitted`'s repair cannot be steered through them.
- **The surrogate-pair acceptance** from the last review round is the right call; I checked the convergence scenarios (fragment-composed pair vs. direct encode vs. user-typed ASCII spelling) and they stay distinct or correctly converge.
- **Handler traps**: generic/boxed/format-specifier holes all route through the type test; the `InertString?` overload is correct.
- **Perf**: clean text is zero-allocation pass-through (`ToString_DoesNotCopy` pins it), policy delegates are cached statics, the 1.1M-scalar sweep costs ~0.3s — fine to run unconditionally in CI as claimed.
- **CI change** (`feature/**` base filter) does what the body says and is properly called out.

The corpus with per-fixture attack rationale, the accounted-enumeration style of the reflection gates, and the honest "boundaries not yet fully enumerated" section of the design doc are all excellent. Fix the `type.GetMethods` line (and ideally re-prove with the handler mutation), sweep the three stale comments, and this is ready.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

richlander:
All findings reproduced before fixing, all fixed in `767164e43`. Thanks — the delegate gate finding is the good kind, because the gate was green the whole time and the PR's own mutation table "proved" it.

## The delegate gate

Confirmed exactly as described, and I reproduced your mutation rather than reading the code: `AppendFormatted(string?, Func<Rune, bool>)` on `InertStringHandler` left the gate **green**. `type.GetMethods(Declared)` turns that same mutation **red**, and the clean tree stays green.

Your closing observation is the part worth keeping: the original proof used an `EnsurePermitted` overload on `InertString`, which is the one type the broken scan did cover. So the mutation test passed *and* the gate was weaker than documented — a mutation proof only certifies the path it walks. Re-proven on the handler.

## Root props

Confirmed and closed. Also folded in two things while I was there: the scan omitted `.targets` entirely, and `global using global::InertText.Encoding;` slipped the `StartsWith` match. Both mutation-proven red now — root-props `<Using Include>` and the `global::` spelling.

## The three stale comments

All three confirmed. Rather than sweep them by hand and leave the class open, I turned on `GenerateDocumentationFile` for `InertText`, which makes CS1574/CS1734/CS0419 build errors under the repo's warnings-as-errors.

That immediately found a fourth you didn't: `<see cref="IsPermitted"/>` is **CS0419 ambiguous** between the two overloads. It would have broken the build the moment anyone enabled docs. Also worth noting: **zero CS1591**, so every public member was already documented — the file was one config line away from being gated all along.

## Counts

You're right that they disagreed, and the source comment was the wrong one twice — the measurement is **five and five-of-five**, and it is a stacked-branch fact that doesn't belong in source on a branch shipping the library alone. Replaced with the structural claim, no count. PR body corrected 189 → 190.

## Nits

Both indentation glitches fixed. `IsTextual` widened to `char[]` and `char`-argument generics — mutation-proven with a `ReadOnlySpan<char> Peek()`. You called it "not a live hole" and that's right, but the gate's whole value is catching an addition, so blindness to the spelling an addition would use is the one blindness it can't afford.

## One thing changed under you

The capability namespace was renamed `InertText.Encoder` → **`InertText.Encoding`** after your review. The old name read as member access on the currency type (`InertText` is one word from `InertString`, and `Encoder` is a plausible property), which is exactly the leak the gates disprove — in the one string the audit tells reviewers to grep. Precedent is `System.Text.Encodings.Web.TextEncoder`: encoding word on the namespace, `-er` on the type.

190/190 green, solution builds clean with docs on.

richlander:
Correction to my reply above: I wrote that `TreatWarningsAsErrors` is repo-wide. It is not — it is set in `src/Directory.Build.props`, not the root one, so it covers everything under `src/` (three spoof-fixture projects opt out). InertText is under `src/`, so the substance holds: CS1591/CS0419 really are build errors there, which is what makes `GenerateDocumentationFile` a gate rather than a suggestion. PR body corrected.

While checking that, I also verified the root-props hazard empirically rather than by reading, since `src/Directory.Build.props` shadows the root one and MSBuild stops at the first file it finds walking up. It does not shadow it in practice — line 3 imports the parent via `GetPathOfFileAbove`. Proof rather than inference: a file using `VisualEncoder` unqualified fails with **CS0103**, and adding `<Using Include="InertText.Encoding"/>` to the *root* props makes it **compile**. So your finding was right and reaches compilation, not just the grep.

richlander:
## Re-review of the response commits

**All review findings are addressed and verified.** I re-ran the suite on the current head (`767164e43` via #3660's base): **317/317 green** including the 190 here, and re-proved the delegate gate by mutation — adding `public void AppendFormatted(string? value, Func<Rune, bool> permitted)` to `InertStringHandler` now turns `NoPublicMemberOfTheCurrencyNamespaceTakesADelegate` **red**, which is the exact mutation the buggy scan missed before.

Beyond the asks:

- The global-using gate now covers `*.targets`, repo-root `Directory.Build.*`, and the `global::` spelling — all three holes I flagged, closed.
- `GenerateDocumentationFile=true` is the right systemic fix rather than just correcting the one stale `paramref`; the comment explaining *why* (the audit story lives in doc comments, so cref rot is a defect) earns its place.
- `IsTextual` extended to char-shaped returns — I called this a non-live nit and the fix is cheap; good.
- Grepped the tree for `InertText.Encoder` after the rename: zero stragglers in sources or docs.

One observation on the rename, informational only: `InertText.Encoding` makes the simple name `Encoding` bind to the child namespace for any code inside `namespace InertText` or `InertText.Tests`, shadowing `System.Text.Encoding` (namespace members win over using directives). Nothing in the tree writes bare `Encoding` today, and the failure is a compile error (CS0118), not silence — so this is a known-friction note for future edits in those namespaces, not a request for change.

Ready to merge from my side. Findings on the stacked truncation work are on #3660.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3634, opened 2026-07-31
Raised lambda and local-function bodies lose PDB local names (locals demoted to stack slots)


richlander:
## Summary

A lambda or local-function body that gets raised inline loses the PDB local names its
enclosing method would have kept. Source locals come out as `S_256`-style **stack
slots**, so the render is both less readable and unable to carry any evidence keyed on
local slots.

This is visible with no PDB feature work at all — the names are already read at import
and simply do not survive raising.

## Repro

```csharp
public static class Sample
{
    // lambda
    public static Func<int, int> FlatScopeLambda()
        => x =>
        {
            int step = x * x;
            return step + step;
        };

    // local function
    public static int NestedInIf(int x)
    {
        return F(x);
        static int F(int n)
        {
            if (n > 0)
            {
                int step = n * n;
                return step + step;
            }
            return -n;
        }
    }
}
```

Rendered:

```csharp
public static Func<int, int> FlatScopeLambda() => x =>
{
    int S_256 = x * x;      // source: step
    return S_256 + S_256;
};

public static int NestedInIf(int x)
{
    return F(x);
    static int F(int n)
    {
        int S_256;          // source: step

        if (n > 0)
        {
            S_256 = n * n;
            return S_256 + S_256;
        }
        return -n;
    }
}
```

The enclosing method's own locals render with their PDB names in the same run, so this
is specific to raised sub-bodies.

## Why it matters beyond readability

`S_` names mark **stack slots**, not local slots. That is the substantive part: the
local has not merely been renamed, it has been demoted, so anything keyed on a local
slot index cannot attach to it.

Concretely, this blocks #3622: #3615 derives declaration placement from per-local-slot
PDB scope flags, and a demoted stack slot has no slot for those flags to index. The
scope carry into raised bodies is a two-line change that provably does nothing until
this is fixed (measured in #3622).

Note the second render also shows the declaration hoisted above the `if` even though
the source declares it inside — which is exactly the #3615 shape, unreachable here for
the same reason.

## Direction

Preserve local identity across cross-method import and the raising pipeline so a raised
body keeps its local slots and `LocalNames` rather than demoting them. The names are
already imported — `Lambda` and `LocalFunctionStatement` both take and store
`LocalNames`, and both printer sites already forward them to the nested `IrFunction`.
The loss is upstream of that, in whatever demotes the slot.

Worth confirming first whether the demotion is deliberate (e.g. a nested-scope
[...]

Issue #3622, opened 2026-07-31
Carry PDB declaration-scope flags into lambda and local-function IrFunctions


richlander:
## Context

#3615 threads a per-slot `LocalDeclaredInNestedScope` flag from the portable PDB's
`LocalScope` table onto `IrFunction`, and `CSharpPrinter` uses it to declare a local
at its store site instead of hoisting it to the top of the method.

The flag is only populated for the **top-level** `IrFunction` built by
`MethodImporter` / `IrImporter`. Sub-`IrFunction`s for lambdas and local functions
are constructed separately from `lambda.LocalNames`:

- `src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.RaisedExpressions.cs:244`
- `src/ILInspector.Decompiler/Pipeline/Ir/CSharpPrinter.cs:1685`

Those call sites do not carry scope flags, so `IsLocalDeclaredInNestedScope` is
always `false` inside a lambda or local-function body and every local there keeps
the pre-#3615 hoisted shape.

## Why it matters

This is the same fidelity concern #3615 fixed for method bodies, not a cosmetic one.
Roslyn allocates local slots in source declaration order, so a hoisted declaration
inside a lambda body can round-trip to different IL than the original. #3615 measured
6 differing IL bytes for exactly this swap at method level.

Lambdas and local functions are compiled into display-class or state-machine methods,
each of which has its own `LocalScope` rows in the PDB, so the evidence is present —
it just is not being carried across the sub-function boundary.

## Ask

Populate `LocalDeclaredInNestedScope` on lambda and local-function `IrFunction`s from
the same PDB scope rows, matching how names are already carried via `LocalNames`.

Keep the #3615 safety split intact: the PDB scope is evidence of **intent**, and
`CSharpPrinter.LocalReferencesStayInsideStoreBlock` remains the independent evidence
of **validity**. Both must hold before sinking. Absent a PDB, the flags stay empty and
output must be byte-identical to today.

Add one positive fixture (a lambda whose local's PDB scope is narrower than the body)
and one close negative (a lambda-body store the IR guard must decline), mirroring
`DeclScopeClient` / `DeclScopeLoopClient` in
`src/ILInspector.Decompiler.Tests/CfgSampleClass.cs` — appended at the **end** of that
file per its append-only convention.

Raised as a declared non-blocking follow-up on #3615.

richlander:
[...]

PR #3643, opened 2026-07-31
Stop spelling an unraised local-function call with its source name (#3631)

### LINKED ISSUE #3631: Declined local-function raising emits a call to an undeclared method (invalid C#, no fidelity signal)
## Summary

When `LocalFunctionRaisingPass` **declines** to raise a local function, the
declaration is (correctly) not emitted — but the call site is still printed with the
local function's *speakable* source name. The emitted C# therefore calls a method that
is declared nowhere in the output, and does not compile (**CS0103**).

Nothing in the output says so. There is no diagnostic and no fidelity downgrade, at
any verbosity.

## Repro

Fixture (appended to `src/ILInspector.Decompiler.Tests/CfgSampleClass.cs`):

```csharp
public static class Probe3622Shapes
{
    public static int WithTry(int x)
    {
        return F(x);
        static int F(int n) { try { return n / 2; } catch { return 0; } }
    }

    public static int WithForeach(int[] x)
    {
        return F(x);
        static int F(int[] a) { int t = 0; foreach (int v in a) t += v; return t; }
    }

    public static int WithIf(int x)
    {
        return F(x);
        static int F(int n) { if (n > 0) return n; return -n; }
    }
}
```

```bash
dotnet-inspect type Probe3622Shapes --library <tests>.dll -S "Decompiled Source" -v:d
```

Output:

```csharp
public static class Probe3622Shapes
{
    public static int WithIf(int x)
    {
        return F(x);
        static int F(int n)
        {
            if (n > 0)
            {
                return n;
            }
            return -n;
        }
    }

    public static int WithTry(int x) => F(x);       // <-- F declared nowhere

    public static int WithForeach(int[] x) => F(x); // <-- F declared nowhere
}
```

`WithIf` raises correctly. `WithTry` and `WithForeach` do not, and their renders are
invalid C#. Compiling the emitted `WithTry`/`WithForeach` unit:

```
error CS0103: The name 'F' does not exist in the current context
```

A `for` loop body fails the same way. Reproduced on clean `origin/main` (not
introduced by any in-flight branch).

## Mechanism

1. `LocalFunctionRaisingPass` gates raising on `IsPrintableBody`
   (`src/ILInspector.Decompiler/Pipeline/Passes/LocalFunctionRaisingPass.cs:86`,
   helper at :305). That helper accepts only a single block of
   `ExpressionStatement` / `StoreLocal` / `StoreStackSlot` ending in a valued
   `Return`. A `try` or `foreach`/`for` body fails it, so the pass `continue`s.
2. Because it declined, it never rewrites call sites to `LocalFunctionInvocation`
   and never adds a `LocalFunctionStatement`. That part is correct.
3. The host body is left with a **direct call** to the synthesized
   `<WithTry>g__F|0_0`. The printer names it through `CSharpNaming.MethodName`
   (`src/ILInspector.Decompiler/Pipeline/Ir/CSharpNaming.cs:103`), which strips
   `>g__` and the `|N_M` ordinal and yields the bare speakable `F`.
4. The synthesized method is filtered out of the type render, so no declaration of
   `F` exists anywhere in the emitted compilation unit.

The name decode at step 3 is only sound when step 2 *succeeded*. It is applied
unconditionally.

## Why this is a rules violation, not just a defect

AGENTS.md: *"Keep failure visible. Do not turn decode, acquisition, analysis, or
rendering failures into success-shaped empty output."* A declined raise is exactly
such a failure, and it currently renders as clean, confident, plausible C#. A reader
cannot distinguish `WithIf` (genuinely recovered) from `WithTry` (silently broken)
without compiling the output.

The fidelity signal actively asserts the wrong thing.
`UnspeakableNameFidelityTests` pins a call to `<M>g__Local|0_0` as
`DecompilationFidelity.Full` (~line 204-211) precisely *because* the name decodes to a
source spelling — an assumption that only holds for a raised local function.

Note the established convention for the analogous **lambda** case in the same test
file (~line 87-90): an unraised `<M>b__0_0` keeps its unspeakable identity, escaped to
`__M_b__0_0`, and is rated `Partial`. The `>g__` path is the inconsistent one.

## Suggested direction

Make the speakable decode conditional on the raise having happened, rather than
unconditional at the naming layer:

- When a call targets a `>g__` method that was **not** raised into a
  `LocalFunctionStatement` in the enclosing print scope, print the escaped
  unspeakable form (matching the `__M_b__0_0` lambda convention) and downgrade
  fidelity to `Partial`.
- Adjust the `UnspeakableNameFidelityTests` `Full` expectation to cover the raised
  case only, and add a negative for the declined case.

That keeps the output honest without requiring `IsPrintableBody` to grow. Widening
`IsPrintableBody` to accept structured bodies is a separate, larger improvement —
worth doing, but it would only shrink this bug's surface, never close it, since some
body will always be unraisable.

## Evidence notes

- Verified with `csc` (net10.0): the emitted unit fails `CS0103`.
- `if`-bodied and expression-bodied local functions raise correctly, so the bug is
  specific to bodies `IsPrintableBody` rejects — `try`, `for`, `foreach` at minimum.
- Found while investigating #3622; see that issue for the related finding that raised
  lambda/local-function bodies also lose their PDB local *names* (rendering `S_256`
  rather than `step`).

richlander:
Fixes #3631.

## Conclusion

A call to a local function that `LocalFunctionRaisingPass` **declined** to raise was
printed with its decoded source spelling. The pass correctly emits no declaration in
that case, so the rendered C# called a method declared nowhere — **CS0103**, presented
as ordinary recovered source with no diagnostic and `DecompilationFidelity.Full` at
every verbosity.

This lands the honest signal. It does **not** make the render compile — see the
non-action boundary.

## Before / After

Fixture (`UnraisedLocalFunctionSamples`, appended to `CfgSampleClass.cs`) — two bodies
`IsPrintableBody` rejects, plus one it accepts as a control:

```csharp
public static int CallsUnraisedTry(int x)
{
    return F(x);
    static int F(int n) { try { return n / 2; } catch { return 0; } }
}

public static int CallsRaisedIf(int x)      // control
{
    return F(x);
    static int F(int n) { if (n > 0) return n; return -n; }
}
```

**Before** — indistinguishable from a correct recovery:

```csharp
public static int CallsUnraisedTry(int x) => F(x);        // F declared nowhere; Full
public static int CallsUnraisedForeach(int[] x) => F(x);  // F declared nowhere; Full
```

**After** — the compiler-generated identity is visible and fidelity degrades:

```csharp
public static int CallsUnraisedTry(int x) => __CallsUnraisedTry_g__F_0_0(x);        // Partial
public static int CallsUnraisedForeach(int[] x) => __CallsUnraisedForeach_g__F_1_0(x); // Partial
```

**Control, unchanged before and after** — raised, declared, `Full`:

```csharp
public static int CallsRaisedIf(int x)
{
    return F(x);
    static int F(int n)
    {
        if (n > 0)
        {
            return n;
        }
        return -n;
    }
}
```

## Why the discriminator is structural, not a heuristic

Every call site of a **raised** local function is rewritten to a
`LocalFunctionInvocation`, which prints its already-decoded `Name`
(`CSharpPrinter.cs:3214`) and never reaches `CSharpNaming.SourceMethodName`. A `Call`
still carrying a `>g__` name is therefore, **by construction**, one the pass declined.
No guessing is involved.

I verified all 15 `SourceMethodName` call sites are call/method-group targets; none is
a declaration header.

## Change

- **`CSharpNaming.SourceMethodName`** no longer decodes `>g__`. An unraised call now
  renders the sanitized identity, exactly matching the existing treatment of an
  unraised lambda body method `<M>b__N_M` → `__M_b__0_0`. `CSharpNaming.MethodName`
  still decodes for `LocalFunctionRaisingPass`, the one caller entitled to it.
- **`CSharpSpellability.MethodIssue`** checks the metadata name as it stands rather
  than a `>g__` decode of it.

The second half **makes existing dead code reachable**:
`DecompilerFidelityDiscriminators.LocalFunctionMethodName` was already declared and
already selected by `MethodNameDiscriminator`, but unreachable — the decode satisfied
the escapability check before the discriminator could ever be chosen. `grep` found no
test and no other consumer. The infrastructure for this fix was built and then
defeated by the decode.

The old `SourceMethodName` doc comment asserted *"The method's fidelity is still
degraded by the spellability check on the unchanged IR name"*. That was false for
`>g__`: the check applied the same decode. Both doc comments now state the invariant
they actually have.

## Evidence

| Gate | Result |
| --- | --- |
| `--gate fast` | **4430 / 0 / 0** |
| `Area=RoundTrip` | 568 / 0 |
| `Area=Validity` | 103 / 0 |
| `Area=Corpus` | 348 / 0 |
| `Area=Fidelity` | 68 / 0 (~17 min) |
| `dotnet-inspect.Tests` | 2765 / **1** / 14 — the 1 is the pre-existing `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which reproduces on clean `origin/main` |

The CLI suite matters here specifically because fidelity is computed from current IR:
it confirms the product reads fidelity **after** raising, so a method whose local
function *is* raised still reports `Full` at the CLI.

**Non-vacuity** — both halves tampered independently, each caught:

| Tamper | Result |
| --- | --- |
| restore the `>g__` decode in `SourceMethodName` | 2 tests fail |
| restore the `>g__` decode in `MethodIssue` | 3 tests fail |

**CS0103 confirmed with `csc`** (net10.0) for the emitted `Before` unit, and again for
the `After` unit — the render is invalid in both, which is the point of the
non-action boundary below.

## Tests

New `UnraisedLocalFunctionCallTests` (5 cases, compiled fixtures — real IL, not
hand-built IR):

- declined bodies emit no `F(` call and no `static int F(` declaration, and carry no
  raw `<` or `|`;
- declined bodies degrade to `Partial` with the `local-function-method-name`
  **discriminator** (asserted on the typed cause, not the `DEC` id, since the
  discriminator is the stable semantic facet policy consumers key on);
- the raised control keeps `F(`, its declaration, and `Full`.

Four existing tests pinned the old contract and were updated, each with a comment
saying why:

1. `CSharpNamingTests` — the `("<M>g__return|0_0", "@return")` row moves from the
   demangling theory to the sanitizing theory. The keyword-escape rows are untouched.
2. `UnspeakableNameFidelityTests.LocalFunctionMetadataName_StaysFull` → 
   `UnraisedLocalFunctionMethodGroup_DegradesToPartial`, now identical in shape to the
   un-raised **lambda** method-group test a few lines above it.
3. `RaisingPassTests.LocalFunctionCall_DemanglesToSourceName` → 
   `UnraisedLocalFunctionCall_RendersSanitizedNameNotSourceName`. This test used
   `CSharpPrinter.Print`, which has no cross-method resolver, so nothing was ever
   raised — it pinned the bug directly.
4. `FinallyDisposePrinterTests` — the fidelity assertion moves to **after**
   `PrintRaised`. Its local function *is* raised, so the rendered output is `Full`;
   asserting at the post-raise point describes the state the output actually has.

`LadderRung4GateTests` is the sharpest confirmation that the change is targeted: in one
test, the three **raised** local functions (`Double`, `AddSeed`, `Fact`) each assert
both the call *and* the declaration and were untouched, while only the one already
asserting `Partial` + a leaked `DisplayClass` — i.e. the unraised one — changed.

## Non-action boundary

- **The declined render still does not compile.** `__CallsUnraisedTry_g__F_0_0` is as
  undeclared as `F` was. This PR converts a silent, plausible-looking invalid render
  into a `Partial`, visibly-compiler-generated one. Emitting a declaration would mean
  fabricating one; making these bodies raise is #3634's territory and larger.
- **`IsPrintableBody` is unchanged.** Widening it to accept `try`/`for`/`foreach`
  bodies would shrink this bug's surface but never close it — some body is always
  unraisable — so it is deliberately out of scope.
- **Lambdas are unaffected**; they already behaved this way, and this change makes the
  local-function path consistent with them rather than the reverse.

## Related

Found while investigating #3622 (blocked on #3634). This PR is independent of both.


richlander:
## Corpus sensor caught a real defect in the first approach — redesigned

CI's `test` job failed on the PR-quick corpus sensor: **detected lowering residue 17.67% → 18.33% (+0.66 pp)** on the pinned subset. I reproduced it locally and it was not a flake or a tolerance quibble — it was a genuine over-report the first design caused.

### What was wrong

The first commit judged an unraised local function **by its metadata name**. That is not knowable from the name: *before* `LocalFunctionRaisingPass` runs, a local function that **will** be raised carries an identical `<Enclosing>g__Name|N_M`. So the check degraded methods whose final output is perfectly valid.

A clean A/B against a freshly generated `origin/main` baseline isolated exactly **6** `Full → Partial` methods. Four render a correctly declared local function — e.g. `NuGet.Versioning.FloatRange.TryParse`, whose output contains both the calls and `static int IndexOf(string str, char c) => …`. Calling that `Partial` is simply wrong.

The mechanism: `CorpusSensor.AnalyzeCompleteness` runs the passes with `PassContext.None` — **no cross-method seam** — so nothing can ever be raised there, and every local-function call looked "declined".

### The fix

Record the fact at its source instead of re-deriving it from display text. `LocalFunctionRaisingPass` — the only component that knows — stamps `MethodRef.LocalFunctionRaiseDeclined` on every call it considered and left without a matching declaration. Both consumers read the stamp:

- the printer, via a new `CSharpNaming.SourceMethodName(MethodRef)` overload;
- fidelity, via `CSharpSpellability.MethodIssue`.

The stamp is keyed on **whether a declaration was actually emitted**, not on the `CompilerGenerated` fact that gates raising, so a mangled call left undeclared is still spelled honestly when that metadata fact is unavailable.

**Boundary:** without the seam the pass is a documented no-op, so it stamps nothing and behaviour is byte-identical to before. That boundary is what keeps the corpus sensor honest, and it has its own gate — `WithoutTheCrossMethodSeam_NoCallIsStampedAndFidelityIsUnchanged`.

### Behaviour, unchanged in intent

| Case | Result |
| --- | --- |
| Declined (`try` body) | `__CallsUnraisedTry_g__F_0_0(x)` + `Partial` + `LocalFunctionMethodName` |
| Raised (`if` body) | `F(x)` with `static int F(int n)` declared + `Full` |
| Real-world raised (`FloatRange.TryParse`) | unchanged, `IndexOf(...)` + declaration |
| No seam | unchanged from `main` |

### Evidence at this head (`18e871f74`, `origin/main` merged)

| Gate | Result |
| --- | --- |
| PR-quick corpus sensor | **PASS** (exit 0); pinned gate **0 pp** (was +0.66 pp) |
| Clean A/B vs fresh `origin/main` baseline | **zero** `Full → Partial` among methods present in both runs |
| `--gate fast` | 4449 / 0 |
| `Area=RoundTrip` / `Validity` / `Corpus` / `Fidelity` | 568 / 103 / 348 / 68, all 0 failures |
| `dotnet-inspect.Tests` | 2765, 1 failure — pre-existing `Layout_Count_…`, reproduced on clean `main` |

The advisory (non-gating) aggregate still moves +0.20 pp; the per-method delta shows every changed row is `None → …` / `… → None` sample churn, because this change adds methods to `ILInspector.Decompiler.dll` itself and shifts its own hash-stable 100-method sample.

**Non-vacuity:** suppressing the stamp fails 7 tests; making the printer ignore it fails 4; making fidelity ignore it fails 3; removing the no-seam guard fails **exactly** the boundary test.

### Non-action boundary

- A **method group** over a local function (`AddressOfMethod` / `DelegateCreation`) is not stamped — the pass only raises `Call` nodes, and a method group may legitimately reference a local function raised through its call sites. Behaviour there is unchanged from `main`.
- The fix makes the failure **visible**, not absent: the declined render is still uncompilable (`__CallsUnraisedTry_g__F_0_0` is as undeclared as `F` was). Widening `IsPrintableBody` is #3622/#3634 territory.

Adversarial review has not started — per AGENTS.md I held it until the branch was green and settled, which it now is.

richlander:
## Review round 1 reconciliation — `18e871f74` → `7d8f4a3de`

Two adversarial reviews ran against `18e871f74` in isolated worktrees: **Gemini 3.1 Pro** and **GPT-5.6**. Between them they found **three** real holes, all of them #3631 reached by a different route — a synthesized local-function name decoded to a spelling nothing declares, reported `Full`. All three are fixed in `7d8f4a3de`; one further finding is refuted with evidence.

### 1. No seam, no stamp — *blocking, both reviewers*

The pass returned early when `context.ImportMethodBody is null`, on my premise that "the product always supplies the seam". **That premise was wrong.** `CSharpBodyDiff.cs:485` and two `ResearchViews` lenses (`:440`, `:598`) print with `importMethodBody: null`. GPT reproduced it through the shipped CLI:

```
member ILInspector.Decompiler.Tests.UnraisedLocalFunctionSamples CallsUnraisedTry:1 \
  --library ... -S "Cost Overlay,Semantics Overlay"
```

Both overlays emitted `return F(x);` with no declaration of `F`. They now emit `return __CallsUnraisedTry_g__F_0_0(x);` at `Partial`.

The guard existed only to stop the corpus sensor's residue metric from moving. **The sensor was the thing that was wrong.** It ran the pipeline with `PassContext.None`, so cross-method passes declined every candidate and it measured a pipeline strictly less raised than the product ships. #2818 had already found and fixed exactly this for classic state machines; the same seam is now wired for the default profile, with the sensor's own comment as the precedent. The pinned gate moves **−0.84 pp (17.67% → 16.83%)** — 13 methods were counted as debt purely because the sensor withheld the seam.

GPT's suggested fix (supply seams *to* `CSharpBodyDiff`/`ResearchViews`) is a reasonable enhancement but not a substitute: even with a seam, declined calls must still be stamped. Filed as follow-up, not done here.

### 2. Same-name siblings bind to the wrong function — *blocking, both reviewers, and independently reproduced here*

Keying the stamp on "was some declaration of the decoded name emitted?" is unsound: C# permits same-named local functions in disjoint scopes (`<M>g__F|0_0`, `<M>g__F|0_1`). When one raises and the other declines, the declined call went unstamped and printed `F(...)` — **binding to the wrong function**, output that compiles and silently means something else. Strictly worse than the undeclared name #3631 reported.

A surviving `Call` is by construction one raising did not rewrite, so it is now stamped unconditionally and the name set is gone. Gated by `DeclinedLocalFunction_IsNotSpelledAsARaisedSiblingOfTheSameName`.

### 3. Method groups are never stamped — *blocking, GPT only*

Neither Gemini nor I found this one. A local function converted to a delegate lowers to `ldftn` and imports as `DelegateCreation`/`AddressOfMethod` — **no `Call` node exists**, so the sweep never saw it, and the two method-group printer sites (which I had deliberately left on the string overload) decoded the raw name to `Type.F`: a member that does not exist, CS0117, at `Full`. Both nodes now carry the stamp and both sites read it.

### Refuted: orphaned nested calls — *Gemini finding 3*

Gemini predicted that a successfully raised `F0` whose body calls `F2` would leave an orphaned `Call` that my unconditional stamping would then mis-spell. The premise is false: `HasOtherLocalFunctionCall` **declines** any body that calls another local function, so a raised body structurally cannot contain one. Verified on a compiled fixture — `F2` raises and is declared, `F0` declines and prints honestly:

```csharp
return __M_g__F0_0_1(a) + F2(a);   // Partial
```

Gemini's findings 4 and 5 (`MethodRef` record-equality/cache safety, idempotency) and GPT's attack points B–D independently reached the same clear verdict I had.

### Validation at `7d8f4a3de`

| Gate | Result |
| --- | --- |
| Corpus sensor (pinned) | **PASS**, −0.84 pp |
| `--gate fast` | 4451 / 0 |
| `Area=RoundTrip` / `Validity` / `Corpus` / `Fidelity` | 568 / 103 / 367 / 68, all 0 |
| `dotnet-inspect.Tests` | 2793, 1 failure — pre-existing `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, reproduced on clean `main` |

Three new gates, each proven non-vacuous by tampering: restore the declared-name set → the sibling test fails; revert the method-group printer site → the method-group test fails; suppress the stamp → the suite fails.

The head moved, so **round 1 does not carry**. A clean round at `7d8f4a3de` is owed and is being run now.

richlander:
## Review round 2 reconciliation — `7d8f4a3de` → `200ecef66`

Clean round at `7d8f4a3de` from **Gemini 3.1 Pro** and **GPT-5.6**, isolated worktrees, both briefed that round-1 findings were fixed and told to attack the new method-group and harness surfaces. They found **two more holes, both in the method-group work I added in round 1.** Both are fixed.

### 1. Function pointers escaped the sweep entirely — *blocking, found independently by both*

`AddressOfMethod` **does not exist when the sweep runs.** `ldftn` imports as `LoadFunctionPointer` and only becomes `AddressOfMethod` in `MethodAddressPass`, registered at `IrPass.cs:361` versus this pass at `:233`. So the `AddressOfMethod` case I added was dead code, and a function-pointer reference printed:

```csharp
return &F;      // Full — and F is declared nowhere (CS0103)
```

I reproduced it **with the seam present**, which is sharper than either reviewer's repro. `LoadFunctionPointer` now carries the stamp; `MethodAddressPass` forwards `pointer.Method` into the new node, so it survives the conversion. `AddressOfMethod` stays in the sweep so the fix does not silently depend on pass ordering.

### 2. A method group over a *raised* local function was stamped declined — *blocking, GPT only*

`RaiseCalls` rewrites only `Call` nodes, so `Func<int, int> d = F;` survives the raise **even though the declaration is emitted**. My round-1 fix stamped it declined regardless, printing `Type.__M_g__F_0_0` — a member that exists nowhere — and degrading a faithful method to `Partial`.

The fix is not simply "don't stamp it". The pre-existing spelling `Type.F` is **also** wrong (CS0117): the declaration is a *local function*, not a member of the type. The reference has to be spelled bare `F`.

So the stamp is now a tri-state — `LocalFunctionRaiseState { None, Raised, Declined }`:

| State | Spelling | Fidelity |
| --- | --- | --- |
| `Declined` | sanitized metadata identity | `Partial` |
| `Raised` | source name, **unqualified** | `Full` |
| `None` | IR the pass has not judged | unchanged |

The raised set is keyed on the **full synthesized identity**, never the decoded source name — the same distinction that resolved round 1's sibling-collision finding. Before/after on GPT's fixture:

```csharp
// before
Func<int, int> V_0 = new Func<int, int>(MixedLocalFunctionUse.__M_g__F_0_0);   // Partial, CS0117
// after
Func<int, int> V_0 = new Func<int, int>(F);                                    // Full, compiles
```

### The systemic fix

Two rounds, three separate holes, all the same failure mode: **a node the sweep didn't know about falls back to decoding the raw name.** So rather than keep patching node types as reviewers find them, there is now a structural gate — every `IrNode` carrying a `MethodRef` must be listed either as *swept* or as *unable to name a local function, with a reason*, asserted by **set equality in both directions**. A new node type fails the suite instead of shipping as a silent hole, and a stale entry fails too.

Reflection found **three** `MethodRef`-bearing nodes that my source grep had missed (`DeconstructionAssignment`, `DeconstructionTarget`, `IncrementDecrement`) — which is the point.

### Clean results

Both reviewers cleared the corpus-sensor harness change (attack point C), independently confirming the `Parallel.ForEach` seam is safe by the same argument the `ClassicStateMachines` branch already relies on, and judging it a fidelity fix rather than a weakening. Both cleared `MethodRef` record equality and mutation, stamp ordering across later passes, and over-stamping/false positives.

### Validation at `200ecef66`

| Gate | Result |
| --- | --- |
| Corpus sensor (pinned) | **PASS**, −0.84 pp |
| `--gate fast` | 4454 / 0 |
| `Area=RoundTrip` / `Validity` / `Corpus` / `Fidelity` | 568 / 103 / 367 / 68, all 0 |
| `dotnet-inspect.Tests` | 2793, 1 failure — pre-existing `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, reproduced on clean `main` |

Three more gates, each proven non-vacuous by tampering: drop the `LoadFunctionPointer` case → the address test fails; force `State` to always return `Declined` → the raised-method-group test fails.

The head moved again, so **round 2 does not carry either.** Round 3 at `200ecef66` once CI is green.

richlander:
## Round-3 review reconciliation (head `200ecef66`)

Two models reviewed at `200ecef66`: **Gemini 3.1 Pro** and **GPT-5.6**. Both filed blocking findings; one was retracted by its author after challenge. Fixes are in `4ee68eabe`.

### Finding 1 — generic local functions raised into a declaration that cannot express them (both reviewers, independently) — **FIXED**

`LocalFunctionStatement` carries no type-parameter list. `Tag<T>()` called as `Tag<int>()` and `Tag<string>()` was declared `static int Tag()` with **both call sites collapsed to `Tag()`**, and a surviving method group over it printed bare `Tag`. GPT compiled the emitted C# and got `CS0411`, `CS1503`, `CS0246`; Gemini got `CS0411`. Reported `Full` throughout — the exact failure mode of #3631.

**Provenance:** I reproduced this on clean `origin/main` and it is **byte-identical** there, so the underlying bug is pre-existing, not introduced here. Gemini confirmed on request: *"it is 100% pre-existing."* I fixed it anyway rather than deferring it, because it is tightly coupled to this PR's new code — the `Raised` state is what makes the generic **method group** print bare `Tag`, a spelling that did not exist before this PR.

Generic local functions are now declined, which routes them through the honest-spelling path this PR adds: `__M_g__Tag_0_0<int>() + __M_g__Tag_0_0<string>()` at `Partial`, with the instantiations preserved and distinct.

### Finding 2 — identity keyed on rendered display text (GPT) — **FIXED**

`DeclaringType.ToDisplayString()` omits namespace and assembly, so `NsA.Owner` and `NsB.Owner` render identically. A reference to one type's local function could match a declaration raised from a different type's same-named one and **silently bind to the wrong method** — output that compiles and means something else. This is the same hazard class as the round-1 sibling-collision finding, one level up.

Now keyed on the `TypeRef` itself, whose equality is hand-written and structural (assembly, namespace, name, recursively type arguments) and so does not hit the `ImmutableArray` reference-equality trap that rules out using the `MethodRef` record. The **call grouping now uses the same helper**, so what is grouped and what is recorded as raised cannot disagree; that also fixes the same latent collision in the pre-existing `GroupBy`.

### Finding 3 — outer sweep overwrites a nested `Raised` stamp (Gemini) — **RETRACTED by its author**

I could not reproduce it: the iterator host prints `/* Unsupported ... MoveNext not reconstructed */` and the async host prints the raw builder sequence, so in neither case was a nested body spliced into the host tree for the sweep to re-stamp. On challenge Gemini retracted it in full — *"You caught me hallucinating the reproduction"* — having conflated two probes, the second of which was a capturing local function correctly declined via `HasThis`.

I also **rejected the fix Gemini proposed for it** ("only stamp `None` nodes"), because it is safe in one direction only: it would equally prevent a nested `Declined` from being upgraded when the *outer* pipeline is the one that raised the function. That case is reachable today — a self-method-group inside a raised body — and the current unconditional sweep gets it right:

```csharp
static int F(int n)
{
    Func<int, int> V_0 = new Func<int, int>(F);   // unqualified; nested Declined upgraded to Raised
    ...
}
```

Gemini agreed the proposed fix was wrong and confirmed the tree-wide scoping that makes the current behavior correct: declarations attach to the host's last top-level block, so *"if the declaration exists, all references to it in that tree MUST be `Raised`."*

### Self-directed attacks (no reviewer prompted these)

| Case | `main` | This PR |
| --- | --- | --- |
| Generic local-function call | `Tag<int>()` @ `Full`, undeclared | sanitized @ `Partial` |
| Generic method group | `Owner.Tag` @ `Full` (CS0117) | sanitized @ `Partial` |
| Capturing local fn + method group | instance receiver → declined | unchanged, honest |
| Capturing, call only | raised @ `Partial` | **byte-identical to `main`** |

### Validation at `4ee68eabe`

| Gate | Result |
| --- | --- |
| Corpus sensor (pinned) | **PASS**, −0.84 pp (17.67% → 16.83%) — unchanged, so declining generics adds no residue |
| `--gate fast` | 4457 / 0 failures |
| `Area=RoundTrip` / `Validity` / `Corpus` / `Fidelity` | 568 / 103 / 367 / 68, all 0 |
| `dotnet-inspect.Tests` | 2793, 1 failure — pre-existing `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, reproduced on clean `main` |

Both new gates proved **non-vacuous by tampering**: removing the generic decline and reverting the key to display strings fails exactly the three new tests and nothing else.

The head moved, so a clean round is owed at `4ee68eabe`; round 4 is dispatched.

richlander:
## Round-4 review reconciliation (head `4ee68eabe`)

**Gemini 3.1 Pro** and **GPT-5.6** reviewed `4ee68eabe`. Two findings, both real, both fixed in `1ff1a253c`. One is a regression introduced by round 3's own fix.

### Finding 1 — the generic decline was too broad, and regressed real framework code (GPT) — **FIXED**

Round 3 declined on `!method.TypeArguments.IsEmpty`. That is not the same question as "is this local function generic": a local function inside a generic **method** inherits that method's type parameters, so an ordinary non-generic local function carries them too.

GPT found this against real framework code rather than a fixture — `System.Runtime.Intrinsics.VectorMath.HypotSingle` lost its raised `CoreImpl` declaration and fell from `Full` to `Partial`, printing `__HypotSingle_g__CoreImpl_7_0<TVectorSingle, TVectorDouble>(...)`. That is a straight fidelity regression, and exactly the over-reporting failure mode that regressed the corpus sensor earlier in this PR's history.

The inherited parameters are already lexically in scope at the declaration site, so the discriminator is now whether a type argument is *not* one of the enclosing method's parameters:

```csharp
if (method.TypeArguments.Any(t => t.Kind != TypeRefKind.MethodGenericParameter))
    continue;
```

Verified against the same real assembly — `HypotSingle` is back to `Full` with the declaration restored:

```csharp
static TVectorDouble CoreImpl(TVectorDouble x, TVectorDouble y)
    => TVectorDouble.Sqrt(TVectorDouble.MultiplyAddEstimate(x, x, y * y));
```

The round-3 case still declines correctly: `Tag<int>()`/`Tag<string>()` print sanitized at `Partial` with the instantiations distinct.

Note Gemini assessed this same attack point as **clean**, reasoning that a local function generic only over its enclosing *type* has empty method `TypeArguments`. True, but it did not consider the enclosing *method* case. GPT's real-product reproduction settles it.

### Finding 2 — the structural gate under-collected and was passing while incomplete (both reviewers, independently) — **FIXED**

`p.PropertyType == typeof(MethodRef)` matched only members typed exactly `MethodRef`, so every collection-typed member was invisible. This is the precise vacuity the gate exists to prevent, and it is the open question I had flagged to round 3.

The query now looks through arrays, nullables, and generic arguments, and reads fields as well as properties. That surfaced **five** escaped nodes — the reviewers named three and four respectively; the widened query found `ForeachStatement` and `UsingStatement` as well:

| Node | Member |
| --- | --- |
| `ForeachStatement` | `ImmutableArray<MethodRef> ConsumedMemberRefs` |
| `UsingStatement` | `ImmutableArray<MethodRef> ConsumedMemberRefs` |
| `ObjectInitializerExpression` | `ImmutableArray<MethodRef?> ConsumedMethods` |
| `WithExpression` | `ImmutableArray<MethodRef?> ConsumedMethods` |
| `InitializerBlock` | `ImmutableArray<MethodRef?> ConsumedMethods` |

All five are now classified **unreachable with a reason**, verified by reading their consumers: they hold the members a pattern consumed — `GetEnumerator`, `MoveNext`, `Current`, `Dispose`, property setters, `<Clone>$` — as typed evidence routed to ReturnToSender via `ConsumedMemberEvidence`. They are never spelled as a callee in output, and none can be a `<M>g__F|0_0`.

### Attack points assessed clean

Both reviewers independently cleared the structural identity key (B): `TypeRef.Equals` iterates array elements structurally rather than by reference, and `GetHashCode`'s inputs are a strict subset of what `Equals` checks, so equal shapes cannot hash apart. Gemini also cleared the sweep overwrite (C) and fidelity accounting (E), confirming the nested-`Declined` → outer-`Raised` upgrade is correct because `HasOtherLocalFunctionCall` runs before the nested `IrPasses.Run`.

### Validation at `1ff1a253c`

| Gate | Result |
| --- | --- |
| `VectorMath.HypotSingle` | `Full`, `CoreImpl` raised — regression gone |
| Corpus sensor (pinned) | **PASS**, −0.84 pp (17.67% → 16.83%), unchanged |
| `--gate fast` | 4458 / 0 failures |
| `Area=RoundTrip` / `Validity` / `Corpus` / `Fidelity` | 568 / 103 / 367 / 68, all 0 |
| `dotnet-inspect.Tests` | 2793, 1 failure — pre-existing `Layout_Count_CountsFilesRatherThanRenderedTreeLines` |
| GitHub CI | all pass incl. `ci-required` |

Both fixes proved **non-vacuous by tampering**: restoring the broad decline and the narrow reflection query fails exactly the two new gates and nothing else.

The head moved again, so round 5 is dispatched at `1ff1a253c`.

richlander:
## Round-5 reconciliation — head `ce0e8b5d9`

Reviewers at fixed head `1ff1a253c`: **GPT-5.6** and **Claude Opus 4.8**. Three
blocking findings, all reproduced on a clean exact-head worktree before acting,
all fixed. Every round so far has found a real bug, including bugs the previous
round's fix introduced; this round is no exception — finding 1 is mine from round 4.

### 1. Own type parameters slipped through the round-4 narrowing — FIXED

Found independently by **GPT-5.6**, **Opus 4.8**, and by my own probe.

Round 4 narrowed the generic decline to
`TypeArguments.Any(t => t.Kind != MethodGenericParameter)`. That judges genericity
from the **call site**, and `Own<U>(U u)` called as `Own<T>(value)` inside `M<T>`
has all-`MethodGenericParameter` arguments — so it slipped through and was declared
`static int Own(U u)` with `U` bound to nothing, at `Full`. Compiling the output
back gives **CS0246 + CS1503**: #3631's exact failure mode, reintroduced by my own
round-4 fix.

Both reviewers correctly noted the shape is byte-identical on clean `main`, so it is
pre-existing there — but it is a genuine hole in the new guard, which is what the
guard exists to close.

Arity comparison does not work either: Roslyn lowers only the *used* enclosing type
parameters, so `Own<U>` inside `Mixed<T>` has lowered arity 1 == host arity 1.

**Fix**: judge from the **body**. A raised body may not declare a type-parameter
*name* the host lacks (`MethodSignature.GenericParameterNames`). Inherited parameters
keep the host's names; own parameters do not. A non-zero count with unknown names
declines — the honest answer, not a guess.

The round-4 regression witness `System.Runtime.Intrinsics.VectorMath.HypotSingle`
still raises `CoreImpl` at `Full`, verified against real CoreLib.

### 2. `&Core<T>` — CS0308 at `Full` — FIXED

Found by **GPT-5.6**.

`AddressOfMethodText`'s new `Raised` branch fixed the type qualifier but still
appended the method's `TypeArguments`. Inside a generic method those are inherited,
so it emitted `&Core<T>` against the raised, non-generic declaration
`static T Core(T v)`. Confirmed with `csc`:

```text
error CS0308: The non-generic method 'Core(T)' cannot be used with type arguments
```

**Fix**: a raised local function is spelled with no type arguments. This is sound
*precisely because* finding 1's fix always declines own-generic local functions, so
every remaining type argument belongs to the enclosing method and is already implicit
at the declaration site. The rendered output now compiles clean.

I also checked `MethodGroupText`, which GPT did not flag: it never appended type
arguments, so it was already correct. It now has its own gate so the two spellings
cannot drift apart.

### 3. Structural gate under-collected through carrier records — FIXED

Found by **GPT-5.6**.

`MentionsMethodRef` stopped at generic arguments, so a `MethodRef` reached through a
plain carrier record was invisible to the gate — it passed while incomplete, for the
third round running. It now recurses through assembly-owned carriers with a
visited-type set, stopping at `IrNode`-typed members (each node is enumerated in its
own right, so following children would make every node match).

Newly surfaced, both classified unreachable with the reason recorded in the test:

| Node | Reached through | Why unreachable |
| --- | --- | --- |
| `ChainedAssignment` | `ImmutableArray<ChainedAssignmentTarget>.Accessor` | property setter |
| `PatternSwitchExpressionArm` | `PropertySubpattern.Accessor` | property getter — `PropertyName` literally slices `get_` off it |

Both are property accessors, so both fall under the accessor rule the gate already
documents. `BindingFlags.DeclaredOnly` is also dropped (an open question **Opus**
raised), so a `MethodRef` member inherited from a base node cannot hide either.

### Non-action

**Opus 4.8** assessed the remaining attack points clean and I found nothing to add
there. Its one open question — `DeclaredOnly` — is addressed above.

### Non-vacuity

Each fix was reverted in turn and the suite re-run:

| Reverted | Fails | Anything else |
| --- | --- | --- |
| body-based generic check | exactly the 3 generic declines | no |
| address-of type-argument drop | exactly its gate | no |
| carrier recursion | the structural gate | no |

### Evidence

`--gate fast` **4461/0** · `Area=RoundTrip` **568/0** · `Area=Validity` **103/0** ·
`Area=Corpus` **367/0** · `Area=Fidelity` **68/0** · `dotnet-inspect.Tests`
**2793/1/14**, the 1 being the known pre-existing
`Layout_Count_CountsFilesRatherThanRenderedTreeLines` (reproduces on clean `main`) ·
pinned corpus sensor unchanged at **−0.84 pp** (17.67% → 16.83%), as it has been at
every revision of this PR.

Round 6 will run at this head once CI is green.

richlander:
## Round-6 reconciliation — head `56c617d90`

Reviewed at fixed head `ce0e8b5d9` by **Gemini 3.1 Pro**. (The GPT-5.6 run aborted
without reviewing — its sandbox refused to read the brief from `/tmp`. It will be
re-run at the new head, so this round counts as one model, not two, and the PR is
not ready.)

One blocking finding, which I also found independently while the round was running.
Fixed in `56c617d90`.

### Type-parameter shadowing defeats the name-membership test — FIXED

C# permits a local function to declare a type parameter whose **name** shadows one of
the enclosing method's, with only a warning:

```text
warning CS8387: Type parameter 'T' has the same name as the type parameter from
outer method 'M<T>(T)'
```

So `hostGenericNames.Contains(name)` was true while the parameter was **not** the
host's. The local function raised, its type-parameter list was dropped, and:

```csharp
static int Own(T u) => 3;
return Own(1) + Own("x");
```

```text
error CS1503: Argument 1: cannot convert from 'int' to 'T'
error CS1503: Argument 1: cannot convert from 'string' to 'T'
```

`Full`. #3631's failure mode for the third time, and a hole in the guard that exists
to close it. Gemini reproduced it correctly and independently; both of us confirmed
it is byte-identical on clean `origin/main`, so the shape is pre-existing there.

### The fix, and why not the one proposed

Neither half of the old test was sufficient, and each catches what the other misses:

| Test | Misses |
| --- | --- |
| call-site kind (round 4) | `Own<U>(U u)` called as `Own<T>(value)` — all arguments are method generic parameters |
| body-name membership (round 5) | shadowing — the name is a host name, the parameter is not |

So judge the **substitution** instead: raise only when every call site instantiates the
body's type parameters with the host's parameters **of the same name, positionally**.
Then dropping the type-parameter list changes nothing, which is the only condition
under which dropping it is sound.

That also rejects a case *neither* of the above can see, which I built while checking
whether a conjunction of the two would do: in `M<T, U>`, a local function declaring its
own `T` and called as `Own<U>(u)` has an all-generic-parameter argument list **and** a
host-known name, yet `Own(u)` against `static int Own(T x)` is CS1503.

### Correcting my own record

Gemini challenged a factual claim in my round-5 commit message — that Roslyn lowers
only the *used* enclosing type parameters, so arity comparison could not work. **It is
right and I was wrong.** Measured against the compiled fixtures:

| Source | Lowered |
| --- | --- |
| `Passthrough<T>` / `Core(T v)` | `<T>` |
| `CalledWithHostTypeArgument<T>` / `Own<U>` | `<T, U>` |
| `DifferingInstantiations<T>` / `Own<T>` | `<T, T>` |
| `ForeignHostParameter<T, U>` / `Own<T>` | `<T, U, T>` |

Roslyn prepends **all** of the host's parameters unconditionally, used or not. Arity
comparison would in fact have caught every case above.

I did not adopt it, for two reasons. It rests on a lowering *convention* rather than on
what the call sites actually do, so it is only as sound as that convention. And it
declines the identity-shadow case — a body whose own `T` is instantiated only with the
host's `T` — which raises correctly and compiles clean; that is now pinned as a
positive test precisely so this rule cannot be silently downgraded to an arity check.

### Also dropped: an ungated assertion

The separate host-membership arm is gone rather than kept as a belt. Removing it failed
**no test out of 4463**, and `AGENTS.md` is explicit that an asserted property names its
gate. A call site lies inside the host, so matching its arguments positionally already
establishes host membership.

### Gemini's clean assessments

Both other attack points **CLEAR**, with reasoning I checked and agree with: the
`AddressOfMethodText` type-argument drop is sound given the generic gate; the `visited`
set in the widened structural walk cannot produce a false negative; and
`ChainedAssignmentTarget`/`PropertySubpattern` really do carry only property accessors.

### Non-vacuity

Dropping the positional call-site loop fails **exactly the two new negatives** out of
4464, and nothing else.

### Evidence at `56c617d90`

`--gate fast` **4464/0** · `Area=RoundTrip` **568/0** · `Area=Validity` **103/0** ·
`Area=Corpus` **367/0** · `Area=Fidelity` **68/0** · `dotnet-inspect.Tests`
**2793/1/14** (the 1 is the known pre-existing `Layout_Count_...`) · pinned corpus
sensor unchanged at **−0.84 pp** · `VectorMath.HypotSingle` still raises `CoreImpl` at
`Full`.

Round 7 will run at this head with two models once CI is green.

richlander:
## Round 7 reconciliation — three independent discoveries of one hole, fixed at `ecf033a0c`

Round 7 ran at `56c617d90` with **GPT-5.6-sol** and **Gemini 3.1 Pro**. Both blocked. I had
found the same hole myself while they ran. Three independent discoveries of one root cause is
the clearest signal this PR has produced, so I am recording it plainly rather than splitting it
by reporter.

### Root cause

`TypeParametersAreTheHostsOwn` decides whether dropping a raised local function's
type-parameter list is the identity substitution. It read **only `Call` nodes**. But `RaiseCalls`
*rewrites* only calls — a method group, a `DelegateCreation`, or an `&F` survives the raise
untouched while still spelling the declaration the raise produced. Those references had no vote,
so they were free to instantiate the body differently from the calls.

### Three reachable failures, all reported `Full`

| Reference kind | Found by | Symptom |
| --- | --- | --- |
| `LoadFunctionPointer` — `&Own<int>` beside `Own<T>(value)` | me, and GPT (BLOCKING 1) | `delegate*<int, int> p = &Own;` against `static int Own(T x)` → **CS8757** |
| `DelegateCreation` — `Func<Type> a = L<int, string>` beside `L<U, V>()` | Gemini | **compiles clean, wrong answer at run time** |
| Recursive self-call — `Own<int>(1, false)` inside `Own<T>` | GPT (BLOCKING 2) | `Own(1, false)` against `static int Own(T x, …)` → **CS1503** |

**Gemini's is the worst of the family and the most valuable finding of the round.** It does not
fail to compile. Raising bound `new Func<Type>(L)` to the *host's* `U` instead of the `int` the
IL names. Gemini executed both assemblies and got `Int32` from the original against `Double`
from the decompiled output — a silent miscompile presented as `Full`. Every earlier instance of
this bug family announced itself with a compiler error; this one does not.

GPT's BLOCKING 2 is a genuinely distinct axis: the offending reference is in the **body**, not
the host. `RewriteSelfCalls` drops a self-reference's type arguments exactly as the host call
sites' are dropped, but self-references never appear in `function.Descendants`, so widening the
host sweep alone does not reach them.

Both reviewers correctly noted that clean `origin/main` also fails on their inputs, but not
byte-identically — main emits an undeclared name (CS0103/CS0308) where this PR emitted the
wrong-but-declared one. That distinction is the whole point of the finding: this PR converted a
loud failure into a quiet one, and in Gemini's case into no failure at all.

### Fix

The gate now takes `List<MethodRef>`, and **every** reference votes — all four node kinds in the
host, plus the body's own self-references, gathered through the same `LocalFunctionReference`
helper the stamping sweep uses. Any single non-identity substitution declines the whole identity.

One trap worth recording, because the first version of the self-reference fix silently did
nothing: `SelfReferences` must compare identities with `.Equals`, not `==`. `TypeRef` implements
`IEquatable<TypeRef>` but declares no `operator ==`, so `ValueTuple`'s `==` compared declaring
types **by reference** and the list came back empty on every input. The dictionary and `GroupBy`
keys were always structural because they route through `EqualityComparer<T>.Default`. This is
the same hazard the existing grouping comment warns about, reached from a different direction.

### No over-decline

Legitimate recursion still raises, generic (`Own(T x, bool)` recursing as `Own(x, false)`) and
non-generic alike, and `VectorMath.HypotSingle`'s `CoreImpl` still raises at `Full`.

### Evidence at `ecf033a0c`

Three fixtures, three gates, each **tampered and observed to fail alone** out of 4467: dropping
the reference widening, dropping the self-reference arm, and restoring the `==` comparison.

`--gate fast` 4467/0 · `Area=RoundTrip` 568/0 · `Area=Validity` 103/0 · `Area=Corpus` 367/0 ·
`Area=Fidelity` 68/0 · `dotnet-inspect.Tests` 2793/1/14 (the 1 is the pre-existing
`Layout_Count_CountsFilesRatherThanRenderedTreeLines`) · corpus sensor **matched baseline**.

### Round 7 does not count

The head moved to fix what round 7 found, so per `AGENTS.md` a moved head needs a clean round at
the new head. Round 8 will be dispatched at `ecf033a0c` once CI is green. Seven rounds have now
each found a real bug; I am not assuming the eighth will not.

richlander:
## The constraint question, answered

Two rounds flagged generic **constraints** as an untested axis of the raise and nobody had actually
measured it, so I did while round 8 runs.

**Result: the raise is sound on constraints, and needs no gate.** A `LocalFunctionStatement` has no
type-parameter list, so a raised body's own constraints are dropped by construction. That is safe
because the call site proves the implication: `Own<T>(…)` is legal C# only if the host's `T`
satisfies every constraint the local declares, so the host's constraint set implies the local's, and
anything the body does legally under the local's constraints is still legal under the host's —
which are equal or stronger. A call site that violated the local's constraints could not have been
compiled in the first place.

Three compiled fixtures, all raising correctly:

| Host | Local | Raised |
| --- | --- | --- |
| `where T : struct, IComparable<T>` | `where T : new()` | `static T Own() => Activator.CreateInstance<T>();` |
| unconstrained | none | `static string Own(T x) => x?.ToString() ?? "null";` |
| `where T : IComparable<T>` | `where T : IComparable<T>` | `static int Own(T a, T b) => a.CompareTo(b);` |

**But the probe turned up a real pre-existing bug, filed as #3664.** The decompiler never emits
`where` clauses *at all*, for any method. `Plain.Compare<T>(T a, T b) where T : IComparable<T>`
— no local function anywhere — decompiles to `public static int Compare<T>(T a, T b) => a.CompareTo(b);`,
which does not compile (CS7036, after binding to an unrelated `MemoryExtensions` extension method),
and reports `Full`. Same failure class as #3631, entirely different cause, reproduces on clean
`origin/main`.

That is **not** this PR's to fix, and it does not weaken the argument above — it means the soundness
of dropping a *local's* constraints rests on the host's being printed, which is #3664's job.

richlander:
## Round 8 reconciliation — both reviewers found the same hole again, fixed at `21ad39af7`

Round 8 ran at `ecf033a0c` with **GPT-5.6-sol** and **Gemini 3.1 Pro**. Both blocked, independently,
on the **same root cause** — as round 7's reviewers did before them, and on the same underlying
mistake.

### Root cause: a reference held by a *sibling* local function

Round 7 established that every reference must vote on whether a raised local function may drop its
type-parameter list. That fix gathered references from two places: the **host's** descendants, and
the **referee's own body**. It missed a third: a **sibling** local function's body. A sibling is a
separate `MethodDefinition`, so its IL is in neither set.

The gate that should have caught it from the other side — `HasOtherLocalFunctionCall`, which
declines a body for touching a foreign local function — read only `Call` nodes. So a sibling holding
`&A<int>` or `Func<int,int> a = F<int>;` declined nothing, and both functions raised.

| Reference kind | Reporter | Symptom |
| --- | --- | --- |
| `LoadFunctionPointer` — `&A<int>` inside sibling `B` | GPT-5.6 | **compiles, wrong answer at run time**: executed both assemblies, `Int32,Double` → `Double,Double` |
| `DelegateCreation` — `Func<int,int> a = F<int>;` inside sibling `G` | Gemini 3.1 Pro | `new Func<int,int>(F)` against `static int F(T u)` → **CS0123** |

Both correctly reported that clean `origin/main` fails on their inputs too, with a *different*
symptom (CS0308 / CS0117 — an undeclared name). That difference is the finding: this PR had
converted a loud failure into a quiet one, and in GPT's case into no failure at all. That is the
regression, not the pre-existing bug.

### Fix

`HasOtherLocalFunctionCall` now routes through the same `LocalFunctionReference` helper the stamping
sweep and the reference gate already use, so all four node kinds count. This is Gemini's suggested
fix #1, and it is sufficient: the sibling declines, so its body is never printed, which is precisely
what makes the referee's raise sound again — the referee's only non-identity reference lived in the
body that is now gone. GPT's `A` still raises correctly.

I did **not** adopt the broader suggestion of gathering references across the whole type. Declining
the sibling removes the offending reference from the output entirely, so there is nothing left for a
wider sweep to judge; a type-wide walk would add cost and a new ordering dependency to reach the
same answer.

On Gemini's fixture both functions now decline, each for its own principled reason: `G` for holding a
foreign reference, and `F` because `F<TShadow>` instantiated as `F<T>` is not the identity
substitution — the round-6 shadowing gate doing its job.

### This is the third instance of one mistake

Judging a decision on `Call` nodes when **four** node kinds can carry a `MethodRef`:

1. Round 7 — the generic gate read only calls.
2. Round 7 — self-references were gathered only from calls.
3. Round 8 — `HasOtherLocalFunctionCall` read only calls.

All three now go through a single `LocalFunctionReference` helper, so there is one place to be wrong
rather than three.

### Evidence at `21ad39af7`

Fixture and gate added, **tampered** by restoring the call-only form and observed to fail alone out
of 4468.

`--gate fast` 4468/0 · `Area=RoundTrip` 568/0 · `Area=Validity` 103/0 · `Area=Corpus` 367/0 ·
`Area=Fidelity` 68/0 · `dotnet-inspect.Tests` 2793/1/14 (the 1 is the pre-existing
`Layout_Count_CountsFilesRatherThanRenderedTreeLines`) · corpus sensor **matched baseline**.

### Round 8 does not count, and round 9 is dispatched

The head moved to fix what round 8 found. Eight rounds have now each found a real bug. I am not
assuming the ninth will not.

richlander:
## Four ordinary shapes this PR fixes, measured against clean `main`

While round 9 runs I probed the axes I asked the reviewers to attack. No blocking finding, but the
probe produced the clearest evidence yet of what this PR is actually worth — four **completely
ordinary** C# shapes that clean `origin/main` decompiles to code naming something declared nowhere,
each reported as `Full`.

| Shape | `main` output | `main` fidelity | `21ad39af7` |
| --- | --- | --- | --- |
| Non-generic local fn in a generic **type** — `Gen<T>.TypeParamOnly` | `=> Own(v);` | **Full** | sanitized, `Partial` |
| Generic type **and** generic method — `Gen<T>.Both<U>` | `=> Own<U>(t, u);` | **Full** | sanitized, `Partial` |
| **Nested** local function — `Outer` containing `Inner` | `=> Outer(v);` | **Full** | sanitized, `Partial` |
| **Mutual recursion** — `IsEven`/`IsOdd` | `=> IsEven(v);` | **Full** | sanitized, `Partial` |

In every row `main` emits a bare decoded name with no declaration anywhere in the output — CS0103 —
and asserts the decompilation is complete and faithful. Nested and mutually recursive local
functions are not exotic; neither is a local function inside a generic type. Fidelity measured
directly through `IrImporter.Import` + `CSharpPrinter.PrintRaised`, both worktrees built from source.

### The attacked axes came back clean

- **Nested and mutual local functions** decline correctly, via `HasOtherLocalFunctionCall` — each body references a foreign local function.
- **A lambda inside a local-function body** raises correctly: `static int Own(int x) => ((Func<int, int>)(y => y * 2)).Invoke(x);`
- **Generic types** decline, and importantly they decline for a *reason that predates this PR*, not because of any gate this PR added.

### One follow-up filed: #3665

Isolating that last row turned up something worth fixing separately. `GenS<T>.NoTypeParam` contains
`static int Own(int x) => x + 1;` — no type parameter of any kind, no capture — and is declined,
while the byte-identical `PlainS.Baseline` in a *non-generic* type raises. **The generic declaring
type alone is what declines it.** That is a missed opportunity across a broad slice of ordinary
code, it is pre-existing on `main`, and it is not this PR's to fix. Filed as **#3665**.

This PR's contribution to that row is to make the decline *honest*: `main` printed the undeclared
name and claimed `Full`; this prints the mangled form and reports `Partial`. #3665 is about removing
the need to decline at all.

richlander:
## Round 9 reconciliation — the first clean round, and one dismissed finding

Round 9 ran at `21ad39af7` with **GPT-5.6-sol** and **Gemini 3.1 Pro**.

### GPT-5.6: CLEAR

The first clean result this PR has received in nine rounds, and it is a substantive one. It attacked
nested and mutual local functions, lambdas, async/iterator lowering, generic types, post-gate
mutations, and `MethodRef`-carrier completeness with compiled fixtures, and it scanned **20,398
CoreLib methods and 10,420 fixture methods for surviving local-function references that carry no
raise stamp — zero found.**

It also ran a `main`-vs-head comparison over all of CoreLib looking for *lost* raises, which is the
over-decline risk every tightening round has added to. Exactly one method lost a declaration, and I
asked for the specifics and **verified it independently** (see below). It is a correct decline.

### The CoreLib witness — the best evidence this PR has produced

`System.Buffers.StringSearchValues.CreateFromNormalizedValues` is **not generic**, yet it contains
`PickAhoCorasickImplementation<TCaseSensitivity>`, called four times with four *different* concrete
instantiations. On clean `origin/main`, at **`Full`**:

```csharp
if (!ignoreCase)          return PickAhoCorasickImplementation(V_5, uniqueValues);
if (V_2) { …              return PickAhoCorasickImplementation(V_5, uniqueValues); }
if (V_1)                  return PickAhoCorasickImplementation(V_5, uniqueValues);
                          return PickAhoCorasickImplementation(V_5, uniqueValues);
static SearchValues<string> PickAhoCorasickImplementation(AhoCorasick a, HashSet<string> u)
{
    …  return new StringSearchValuesAhoCorasick<TCaseSensitivity, …>(a, u);
}
```

Two independent failures in real framework code. `TCaseSensitivity` is bound to **nothing** — the
host is non-generic and the declaration has no type-parameter list — so this is CS0246. And all four
call sites, which in the IL name `CaseSensitive`, `CaseInsensitiveUnicode`,
`CaseInsensitiveAsciiLetters` and `CaseInsensitiveAscii`, print **identically**. The distinction is
gone, silently, at `Full`.

At `21ad39af7` the raise is declined and every instantiation survives:

```csharp
return __CreateFromNormalizedValues_g__PickAhoCorasickImplementation_4_0<
    StringSearchValuesHelper.CaseInsensitiveUnicode>(V_5, uniqueValues);
```

Fidelity `Partial`. I have added this as a **real-artifact canary**
(`CoreLibNonIdentityGenericLocalFunction_IsDeclinedAndKeepsItsInstantiations`), resolved through
`typeof(object).Assembly.Location` rather than a pinned path, asserting both the decline and that all
four instantiations remain distinguishable. The mangled ordinal differs between CoreLib versions
(`_4_0` on 10.0.10, `_3_0` on 11.0-preview), so it is deliberately not pinned. Tampered by relaxing
the generic gate; it fails alongside the seven synthetic gates.

### Gemini finding 1 — NOT REPRODUCIBLE, dismissed

Gemini reported that a self-reference hidden inside a lambda escapes the generic gate because the
gate votes before `IrPasses.Run` reconstructs the lambda, yielding `new Action(A<int>).Invoke()`
against a raised `static void A()` — CS0308 at `Full`.

I compiled Gemini's **exact** fixture and ran it at this head:

```csharp
public void M<T>()
{
    __M_g__A_0_0<T, T>();
}
```

`C.M: Partial`. The raise is declined; the claimed output does not occur. On clean `origin/main` the
same input gives `A<T, T>();` at `Full` — an undeclared name, i.e. #3631, which this PR fixes.

The mechanism described is real and worth guarding against in principle — the gate does vote before
`IrPasses.Run` — but the reported consequence does not exist, because the body is declined well
before reaching that path. **Dismissed as unreproducible.** This is the second time this reviewer has
submitted a reproduction that does not reproduce; the round-7 and round-8 findings from the same
model were both real and excellent, so I am recording this narrowly rather than discounting the
source.

### Gemini finding 2 — real gap, wrong classification. Filed as #3666

Gemini reported that empty local functions (`void A() { }`) are declined because `IsPrintableBody`
rejects a zero-statement body, and labelled it a `Full` → `Partial` regression introduced by this PR.

The root cause is exactly right — `LocalFunctionRaisingPass.cs:534`, `statements.Count == 0` after
`ReturnPass` strips the trailing void return. **The classification is wrong.** Both revisions decline
it; they differ only in what the decline prints. Measured:

| | `main` | `21ad39af7` |
| --- | --- | --- |
| output | `public void M() => A();` | `public void M() => __M_g__A_0_0();` |
| fidelity | **Full** | **Partial** |

`main` emits an undeclared `A()` and calls it faithful — that is #3631, and this PR fixes it. Not a
regression, and nothing for this PR to change. The underlying missed opportunity is genuine, so it is
filed as **#3666** with credit.

### Status

`--gate fast` 4469/0 (the canary is new) · `Area=RoundTrip` 568/0 · `Area=Validity` 103/0 ·
`Area=Corpus` 367/0 · `Area=Fidelity` running · corpus sensor matched baseline at the previous head.

The canary is additive test-only coverage, so the head will move once more. I will run a final
confirmation round at that head before marking this ready.

richlander:
## Round 10 reconciliation — GPT-5.6 (CLEAR + 1 non-blocking), Gemini 3.1 Pro (1 blocking, not reproducible as a defect)

Reviewed at `f90bbfd06`, isolated worktrees, fixed head. Head is now `b2a9daf06`.

### Gemini — "the generic gate votes before the pipeline runs" — structurally right, not reachable, hardened anyway

The structural claim is exactly right, and I want to state it plainly because it is the fourth appearance of this PR's recurring mistake:

`TypeParametersAreTheHostsOwn` judges a candidate **before** `IrPasses.Run(body, …)`, while `HasOtherLocalFunctionCall` deliberately runs on **both** sides of it. `IrPasses.Run` can *add* reference nodes — `LambdaRaisingPass` imports a lambda's body and attaches it to the tree — so a self-reference written inside a lambda is not a node at all when the vote happens. `RewriteSelfCalls` runs afterwards and rewrites calls anywhere in the body. A non-identity `Own<int>(1)` inside a lambda in `Own<T>(T x)` would become `Own(1)` against `static string Own(T x)`: CS1503, at `Full`. That is #3631's failure mode, one level deeper than round 7's.

The report carried no reproduction, so I built one. **It does not reproduce**, and GPT-5.6 independently reached the same conclusion by a different route. Measured:

```
N   (non-generic host)               => lambda raised, Full
G   (generic host, cached lambda)    => new Func<string>(___c__1<T>.___9.__G_b__1_0), Partial
Gc  (generic host, capturing lambda) => new Func<string>(new ___c__DisplayClass2_0<T> { t = t }.__Gc_b__0), Partial
```

Lambdas in a generic context are never raised. A local function only has type parameters worth judging when its host is generic. So `LambdaRaisingPass` cannot materialize a reference in exactly the bodies where this gate matters.

That makes today's safety a **coincidence of #3665**, an unrelated open bug, rather than a property of this pass. Given that judging a decision on the wrong set has now produced three real bugs here, I did both things in `30ea35ba9`:

- **Re-vote after the pipeline**, symmetric with the foreign check already there. The arm is unreachable today and its comment says so.
- **Pin the premise.** `LambdasInGenericContextsAreNotRaised` asserts the three results above, so fixing #3665 fails there and points at this ordering instead of silently reopening #3631. A non-generic control keeps it from passing merely because lambda raising is broken outright.

Tampered the premise gate by inverting its generic assertion; it fails alone.

Gemini also reported that the gate "declines **all** generic local functions". That is not correct — `VectorMath.HypotSingle`'s `CoreImpl` raises at `Full`, and GPT re-confirmed it on CoreLib 10 and 11. Identity instantiations raise; only non-identity ones decline.

### GPT-5.6 — non-blocking: the CoreLib canary was weaker than it read

`"CaseInsensitiveAscii"` is a **substring** of `"CaseInsensitiveAsciiLetters"`, so the canary's four `Assert.Contains` calls still passed when only three distinct instantiations survived. GPT proved it by deleting the `CaseInsensitiveAscii` call from the output and watching the assertions pass — a real hole in the half of the canary meant to show the raise would have *collapsed* four call sites into one.

Fixed in `14f0d3b3e` by extracting the type arguments and comparing **sets**, which fails on a missing instantiation and on an unexpected extra one. The ordinal stays matched rather than pinned (`\d+_\d+`): `_4_0` on CoreLib 10.0.10, `_3_0` on 11.0-preview. I reproduced GPT's tamper against the new assertion — it now fails where the old form passed.

### Verified clean by both reviewers

All four `MethodRef`-bearing node kinds and `Descendants` reachability; `TypeRef`/tuple structural equality (no `operator ==`, the round-7 trap); `GenericParameterName` failing closed; the pre/post foreign gates; stamping propagation; `MethodAddressPass`; over-declining, with `HypotSingle` confirmed still `Full` on CoreLib 10 and 11. GPT independently re-verified the canary's CoreLib shape: host generic count `0`, local signature `1/TCaseSensitivity`, exactly four calls with the four claimed types.

I separately audited the two axes I could settle without a reviewer. All five `Identity()` sites are structural (`HashSet.Contains`/`.Add`, `GroupBy`, `.Equals`) — no `==` on the tuple anywhere. And an unpopulated `GenericParameterName` fails **safe** by construction, not merely empirically: the gate first rejects any body whose `GenericParameterNames` contains an empty entry, so `names[i]` is proven non-empty and an empty argument name can never match.

### Validation at `b2a9daf06` (after merging `origin/main`)

`--gate fast` 4492/0 · RoundTrip 568/0 · Validity 103/0 · Corpus 367/0 · Fidelity 68/0 · corpus sensor **PASS, matched baseline tolerances** (15 assemblies, 1,500 methods, `--corpus-method-cap 100`) so the re-vote costs no raises · CLI 2794/1/14, the 1 being the pre-existing `Layout_Count_CountsFilesRatherThanRenderedTreeLines`. CI green on `b2a9daf06`; MERGEABLE/CLEAN.

Round 11 is running at `b2a9daf06`, since the product changed.

richlander:
## Round 11 reconciliation — GPT-5.6 CLEAR, MAI-Code one finding (dismissed, reproduced), Gemini 3.1 Pro discounted

Reviewed at `b2a9daf06` — the settled head, CI green on that exact SHA, MERGEABLE/CLEAN. Round 11 was required because the product changed in round 10.

### GPT-5.6 — CLEAR, no findings

It went directly at my own biggest worry about `30ea35ba9`: that re-voting over a **post**-pipeline body could cause a *spurious decline*, since reconstructed nodes did not exist when the first vote ran. It found reconstructed nodes preserve their original `MethodRef` and no pass alters `TypeArguments`, and confirmed the safe cases still raise at `Full` — recursive identity call, `LoadFunctionPointer` → `AddressOfMethod`, and method-group reconstruction. It could not reach the new arm: generic cached, type-dependent, capturing, and deliberately type-independent lambdas all stayed on generic `<>c<T>`/display classes and unraised. It also confirmed `group.Key` is still structurally correct, that `SubstituteEnvironment` cannot introduce method references so the ordering of the three post-pipeline checks is safe, and that this PR's files are byte-identical across the `origin/main` merge.

### MAI-Code — one BLOCKING finding, reproduced and dismissed

Claim: the re-vote is blind to the shape it exists to catch, because "once the pipeline rewrites the self-call to a `LocalFunctionInvocation`, the new vote cannot see it."

I reproduced its fixture exactly:

```csharp
public static object Host<T>(T value)
{
    return Core(value);
    static object Core<T>(T v)
    {
        Func<object> f = () => Core<int>(1);
        return f();
    }
}
```

```
return Core(value);
static object Core(T v) => new Func<object>(___c__0<T>.___9.__Host_b__0_1).Invoke();

FIDELITY: Partial
```

Byte-for-byte the same output it reported — and it does not show what it claims. Two independent reasons:

1. **The ordering makes the mechanism impossible.** The re-vote is at `LocalFunctionRaisingPass.cs:315`; `RewriteSelfCalls` is at `:326`. No `LocalFunctionInvocation` node exists when the vote runs — self-calls are still `Call` nodes carrying their `TypeArguments`. That is the same ordering the two other reviewers independently confirmed.
2. **Its own output refutes it.** The lambda was never raised — it prints as the stamped `___c__0<T>.___9.__Host_b__0_1` — so the non-identity `Core<int>(1)` is not a node in the tree and is never printed as source. Nothing invalid is emitted, and fidelity is already `Partial`, which is the honest answer.

This is a *false pass* claim that requires a `Full`, and the result is `Partial`. Dismissed. MAI-Code's verification of (b) and (c) was correct, and it ran the focused class at 27/27.

### Gemini 3.1 Pro — CLEAR, but discounted as a round

Its analysis agreed with GPT's and with mine, and its load-bearing claim — that `RewriteSelfCalls` runs after the re-vote so `TypeArguments` are intact when the gate judges — is correct and I verified it independently at `:315`/`:326`. But it pasted run counts for `-method *CoreLibInstantiationsSurviveDecline`, a name that does not exist. That filter discovers **0** test cases and reports `Total: 0` while still exiting successfully:

```
Discovered:  ILInspector.Decompiler.Tests (0 test cases to be run)
   ILInspector.Decompiler.Tests  Total: 0
```

It reported that as `Total: 1`. This is its fourth unreliable artifact in this campaign, so I am not counting it toward the two-model requirement and am not citing its numbers. MAI-Code was run as the replacement second-family round.

### Where that leaves the tier

Two different-family rounds at the fixed head with no surviving finding: **GPT-5.6 CLEAR**, and **MAI-Code** whose single finding is dismissed with a reproduction and a code-order citation.

### Validation at `b2a9daf06`

`--gate fast` 4492/0 · RoundTrip 568/0 · Validity 103/0 · Corpus 367/0 · Fidelity 68/0 · corpus sensor **PASS, matched baseline tolerances** (15 assemblies, 1,500 methods, `--corpus-method-cap 100`) · CLI 2794/1/14, the 1 being the pre-existing `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on `main`.

### Non-blocking follow-up, not this PR

#3664 (constraint `where` clauses never emitted), #3665 (local functions in generic types never raised), #3666 (empty local functions never raised). #3665 is the bug that makes the new re-vote arm unreachable today; `LambdasInGenericContextsAreNotRaised` fails when it is fixed, which is the point.

richlander:
Ready to merge

Issue #3614, opened 2026-07-31
Two changed-file classifications skip the validation they exist to protect


richlander:
Found during adversarial review of #3608 (GPT-5.6-Sol), which made CI actually gate merges and therefore made under-triggering consequential.

Two changed-file classifications in the `changes` job of `.github/workflows/ci.yml` do not run the validation they exist to protect. Neither is introduced by #3608 — both are pre-existing — but #3608 changes their meaning: a job that skips when it should have run now produces a *green* `ci-required`, because `ci-required` deliberately treats `skipped` as passing.

## 1. `.markdownlint.yaml` matches no filter

The `markdownlint` job runs when `docs == 'true'`, which is set by `*.md`. The lint *configuration* itself is not a `.md` file, so a PR that edits `.markdownlint.yaml` sets no output at all: every job skips and CI is green without ever running markdownlint under the new configuration.

That is the wrong direction for a config file whose entire purpose is to change how the linter behaves.

Suggested fix: add `.markdownlint.yaml` (and any sibling lint config) to the `DOCS=true` case.

## 2. `eng/restore-ilassembler.sh` sets `ILROUNDTRIP=true`, but nothing consumes it alone

`ci.yml` sets `ILROUNDTRIP=true` for `eng/restore-ilassembler.sh`:

```sh
eng/restore-ilassembler.sh) ILROUNDTRIP=true ;;
```

But the IL round-trip validation steps live inside the `test` job, which is gated on `code == 'true'`. `eng/restore-ilassembler.sh` does not match any `CODE=true` pattern, so a PR touching only that script sets `ilroundtrip=true`, skips `test`, and runs no IL round-trip validation whatsoever. The output is computed and then discarded.

Suggested fix: either add the script to the `CODE=true` cases, or give the IL round-trip validation its own job gated on `ilroundtrip` so the output has a consumer. The second is cleaner and matches how `csharp-diff-smoke` and `il-diff-smoke` are already factored.

## Why these are worth fixing rather than tolerating

`ci-required` passing `skipped` is correct and deliberate — most jobs are path-gated, and a required check cannot demand that every job run. The consequence is that the honesty of the merge gate rests entirely on the accuracy of the `changes` filter. Every mis-classification is now a silent hole in the gate rather than merely a missed run.

Related: #3523, #3432.

richlander:
[...]

PR #3646, opened 2026-07-31
SCRATCH do not merge: probe for #3614

richlander:
Throwaway. Touches only .markdownlint.yaml and eng/restore-ilassembler.sh, on top of the #3614 fix. Expect docs=true, ilroundtrip=true, code=true.


richlander:
Throwaway probe for #3614. Evidence captured; closing unmerged.

PR #3647, opened 2026-07-31
SCRATCH do not merge: control for #3614

richlander:
Throwaway control. Same two files, but against unfixed main. Expect docs=false, ilroundtrip=true, code=false, and every job skipped.


richlander:
Throwaway control for #3614. Evidence captured; closing unmerged.

Issue #3616, opened 2026-07-31
Type Info: Type Parameters concatenates every parameter and constraint, while Interfaces is a count


richlander:
Split out of review on #3613 (found by the GPT-5.6 round, independently reproduced).

## The inconsistency

`TypeInfoSection` (added in #3576) has a convention for unbounded facts: **count them, do not list them.** `Interfaces` follows it:

```
| Interfaces | 9 |
```

`Type Parameters` does not — it concatenates every parameter with its constraints into one cell:

```bash
dotnet-inspect type 'System.Func`17' -S "Type Info" --tips q
```

```
| Type Parameters | in T1 : allows ref struct, in T2 : allows ref struct, ... out TResult : allows ref struct |
```

That row is **492 characters** at arity 17.

## Why this is not a #3613 defect

It reproduces identically on `main` via explicit `-S "Type Info"` — verified byte-identical base vs head on the same invocation. #3613 only changed *which sections bare `-S` selects*, so it made the row reachable via a shorter command without altering it. The reviewer who raised it withdrew the blocking finding on that basis.

The section is still **bounded in line count**, which is the property #3613 claims and gates (11-13 lines across 10 diverse types and `System.Func`17`; pinned by `Type_BareSelect_StaysBoundedAtWorstCaseArity`). This issue is about *width*, and about the section disagreeing with its own convention.

## Decision needed

Whether `Type Parameters` should follow `Interfaces` and report arity (e.g. `17`), keep names but drop constraints, or stay as-is because parameter names and constraints are genuine identity information that a count would destroy.

Not a trivial call: changing it alters the merged section for every consumer — `-S "Type Info"`, `-D "Type Info"`, `--schema`, and the JSON/table shapes — which is why it was deliberately kept out of #3613.

Refs #3547, #3576.

richlander:
Review of #3613 (GPT-5.6 round) checked all 11 `TypeInfoSection` properties, not just the two originally named. Scope for this issue is wider than filed:

- **`Interfaces`** is the *only* collection that is counted (`| Interfaces | 9 |`).
- **`Type Parameters`** concatenates every parameter with its constraints — 492 chars at arity 17.
- **`Type`** also spells out every parameter via `FormatGenericFullName`: `System.Func<T1, T2, ..., T16, TResult>`.
- **`Base`** can likewise hold a wide constructed-generic name.
- `Modifiers` joins a bounded vocabulary; the rest are scalars.

[...]

PR #3648, opened 2026-07-31
Give the type listing a bounded API Info section

richlander:
## Slice 4c of the section-catalog stack (#3547)

**Parent: #3613** (slice 4b, `type-bare-select`). This PR targets that branch, not `main`.

Stack: 4a #3576 (merged, added `Type Info`) → 4b #3613 (open, wires single-type bare `-S`) → **4c (this)** → 4d.

### The claim

The type-listing view has no bounded section. Every section its pipeline publishes is a per-kind table that grows with the assembly, and its identity facts live in an inline text line that no `-S` can select:

```text
Library: System.Text.Json.dll | Types: 89 | Methods: 643 | Properties: 217 | Source: Platform | Version: 11.0.0-preview.7.26366.102 | TFM: net11.0
```

This promotes that line into `API Info`, a `Fixed`/`NetworkFree` fact table, mirroring `Type Info` for the single-type view and `Library Info` for the library view:

```console
$ dotnet-inspect type --platform System.Text.Json -S "API Info"
## API Info

| Field | Value |
| ----- | ----- |
| Library | System.Text.Json.dll |
| Types | 89 |
| Methods | 643 |
| Properties | 217 |
| Version | 11.0.0-preview.7.26366.102 |
| TFM | net11.0 |
| Source | Platform |
```

It is populated from the same fields the inline line reads, so selecting the section and reading the line cannot give different answers. `Type_Listing_ApiInfo_RestatesTheInlineIdentityLineExactly` is the gate for that.

### Additive only

The section is `ExplicitOnly`, so it stays off the verbosity ladder, and bare `-S` is unchanged. A differential against the parent slice — three scopes, five verbosities, every per-kind section in markdown/`--tsv`/`--jsonl`/`--count`, plus `--json`, `--rows`, `--columns`, `--fields`, `-n`, `-m`, `-k`, `--urls`, `--paths`, `--row`, `--value`, `--json-array`, `@Surface`/`@Audit`, the deprecated `api` shim, and both `member` shapes — produced **9 diffs across 126 invocations, every one a pure addition** of the section under `-S @All`, `-D`, and `-D --schema`.

> **Correction.** The original text of this section claimed the matrix covered `--limit` and `--docs`. `--limit` is not an option — the flag is `-n` — so that row errored identically on both sides and scored as "no diff" while proving nothing. Caught by GPT-5.6 in review. The harness now fails any invocation whose *base* output reports an unrecognized option, which immediately caught `--plain` as a second instance. The numbers above are from the corrected run.

### One fix was required to render it correctly

`-S "API Info" --tsv` initially emitted **every type in the assembly**, exit 0.

`BuildSurfaceTableView` filters by mapping selected section names to type *kinds*. A name it cannot map leaves that filter empty, and an empty filter means "no filter". So the tabular renderer answered a different question than the one `-S` asked, and disagreed with both the markdown rendering and `--count` of the same invocation. Markdown-only testing could not see it.

Fixed by routing that one selection to the document serializer, exactly as the single-type path already does for a single selected section. Per-kind selections keep their surface projection (`Type_Listing_KindSections_StillProjectTypeRows` is the negative case). All three renderers now agree, and the section is serialized alone so the inline prose line does not leak into machine output.

### Boundedness

5–7 **fact rows** across six targets, from a 1353-type assembly to a 2-type one (8–10 nonblank Markdown lines, once the heading and the header/separator rules are counted — the original text quoted those line counts as though they were rows). The floor is the inline line's own null-skipping: a local `.dll` has no version or TFM, and the section omits exactly the fields the line omits. The three counts are counts, not enumerations — each is one row at any size.

### Non-action boundary (the enumerated residual)

- **Bare `-S` for the listing is unchanged.** It still resolves to an empty set and falls through to the ladder. Wiring it is slice 4d, and `Type_Listing_BareSelect_IsUnchangedAndFallsThroughToTheLadder` now pins `FixedOverviewSectionNames == ["API Info"]` so that wiring must be an explicit decision.
- **The prefix-match path is untouched and still broken.** For a prefix that does not resolve to a type (`type System.Coll …`), `ApiCommand` picks `singleTypeMode` from the argument shape while `TypeCommand.TryWritePrefixBrowse` later renders a *listing*, so `-D` advertises `Classes`/`Structs`/`API Info` while `-S Classes` answers `Error: Select value 'Classes' not found.` Verified byte-identical on `origin/main` — pre-existing, not caused here. Slice 4d.
- **Width is not bounded**, same as `Type Info` (#3616). Row count is.
- **`ApiInfoSection`'s four string columns join the `NotSelfContaining` pin.** They carry the identical values already pinned as `CliApiSurface.Library/Source/Tfm/Version`, so this is a second rendering of already-pinned text rather than new residual; containing only this copy would make the section disagree with the inline line one row above it. Residual tracked by #3463.

### Validation

- Full suite **2790 total, 1 failed** — the known pre-existing `Layout_Count_CountsFilesRatherThanRenderedTreeLines` local-cache failure.
- **Four mutations, each failing its own gate**: `ExplicitOnly` off (default views gain the section), a value drifted off the inline line, `SizeClass` demoted to `Terse`, and the tabular routing disabled.
- `markdownlint` clean on all three changed docs.


richlander:
## CI status: no checks reported

`gh pr checks 3648` reports `no checks reported on the 'listing-api-info' branch` while the PR is MERGEABLE/CLEAN. That is terminal, not pending — this slice targets `type-bare-select` rather than `main`, and that base schedules no runs. I am not waiting on runs that will not come, and I am not reporting this slice as CI-green.

The evidence for this slice is therefore the local validation recorded in the PR body — full suite (2788 total, 1 known pre-existing failure), the 150-invocation differential, four mutation tests, and `markdownlint` — plus the bottom slice's CI: **#3613 is green at `991dfe344` with `ci-required` passing**, and it is this PR's base.

I will re-check once this slice retargets `main` after #3613 merges, where anything newly scheduled does gate.

Two adversarial review rounds are in flight at `99172e778` (two model families, isolated worktrees).

richlander:
## Round 1 reconciliation — 2 blocking findings, both real, both fixed in `75f1cac7f`

**Reviewer: GPT-5.6, at `99172e778`.** Both findings confirmed by reproduction before acting on them.

### Finding 1 (blocking) — `--fields` produced empty output and exit 0

Real, and it is the *same defect class this branch exists to fix*, reintroduced one layer down. The tabular routing branch wrote straight to the console and returned, bypassing `ProjectionDiagnostics.DiagnoseRendered`:

```console
$ type --platform System.Text.Json -S "API Info" --tsv --fields Bogus
(no output, exit 0)
```

Fixed by using the fall-through's exact machinery — `RenderProjectedTable`, `DiagnoseRendered`, `LimitRenderedTableRows` — differing only in *what* is serialized. Verified byte-identical to `-S Classes` for both `--fields` and `--columns`:

| invocation | `-S Classes` (unchanged) | `-S "API Info"` (this PR) |
| --- | --- | --- |
| `--fields Bogus` | `Note: 1 field has no data: Bogus`, exit 0 | identical |
| `--columns FakeColumn` | `Error: No columns matched projection`, exit 1 | identical |

**The bar is parity with the listing sections beside it, not with the single-type view.** `type System.String -S "Type Info" --columns FakeColumn` additionally warns by name and exits 1, because `TypeCommand` runs `ValidateProjection` *before* rendering (`TypeCommand.cs:258-261`). The listing view has never had that pre-check, for any section — `-S Classes` behaves identically on `main`. Filed as #3651 rather than widened into this slice.

Gated by `Type_Listing_ApiInfo_ReportsUnmatchedProjectionsLikeTheRestOfTheView`, which compares against a per-kind section rather than pinning a literal, plus a non-vacuity assertion so silence on both sides cannot pass. Mutation-verified: dropping `DiagnoseRendered` fails exactly that test.

### Finding 2 (blocking) — a false claim in my commit message and PR body

Also real, and the more useful of the two. I claimed the differential matrix covered `--limit 5`. **`--limit` is not an option** — the flag is `-n`. That row errored identically on both sides, so it scored as "no diff" while proving nothing, and inflated the invocation count.

The harness now fails any invocation whose **base-side** output reports an unrecognized option. That immediately caught **`--plain`** as a second nonexistent flag I had been counting. Corrected run against the parent slice: **126 invocations, 9 diffs, 3 vacuous** — every diff a pure addition of the section under `-S @All`, `-D`, and `-D --schema`, with no `<` lines anywhere. The PR body now carries the correction inline.

Lesson worth stating plainly: a differential harness that does not flag rows failing on the base side is not measuring what it claims. Two of my flags were fictional and the harness reported them as evidence.

### Round 2 (Gemini 3.1 Pro) — discarded as unreliable, not counted

The reviewer's commands ran from the default working directory — the primary checkout, on `main` — rather than from its detached review worktree, because each shell starts fresh in the session cwd. Its report ends with `On branch main`. Its one substantive claim (that `-S "api info"` is rejected for casing) is false at this head and true on `main`, which is consistent with that. I cannot tell which of its observations came from which tree, so I am treating the round as not having happened rather than mining it selectively.

Both re-rounds are now running at `75f1cac7f` with explicit per-command `cd` instructions: GPT-5.6 confirming its own findings, and MAI-Code as an independent second family.

Full suite at this head: **2790 total, 1 failed** (`Layout_Count_CountsFilesRatherThanRenderedTreeLines`, pre-existing on unmodified `main` — a local NuGet cache lowercases `newtonsoft.json.nuspec`), 14 skipped.

richlander:
## Round 2 reconciliation — GPT-5.6 clean, MAI-Code one finding (pre-existing and documented)

Both rounds at `75f1cac7f`, isolated worktrees, with per-command `cd` instructions added after the round-1 cwd failure. Both reported clean worktrees.

### GPT-5.6 (confirmation) — **no blocking findings**

Verified both round-1 findings fixed by reproduction across `--fields`, `--columns`, valid/invalid mixes, JSON, JSONL, `--count`, row limits, and `--no-header`, plus zero-type assemblies, version-less local assemblies, and package scopes. Three non-blocking findings, all real, all acted on:

1. **The non-vacuity assertion I added last round was itself vacuous.** Confirmed by measurement: `-S Classes --tsv --fields Nonexistent` prints **54 lines on stdout**, so `Assert.NotEmpty(kindErr + kindOut)` was satisfied by class rows alone — both diagnostics could have been deleted and the test would have stayed green. That is the same empty-set-reads-as-success shape the test exists to catch, one level up, in the test. It now asserts the reference side *names* the bad projection, so the equality that follows carries weight. Fixed in `0a937f219`.
2. **Stale suite count** — 2788 → **2790**. The two-case Theory accounts for the increase. PR body corrected.
3. **"8–10 rows" counted nonblank Markdown lines, not fact rows** — it is **5–7 fact rows**. PR body corrected, with the misreading called out rather than quietly reworded.

### Following finding 1 to its end turned up a real divergence

GPT noted the Theory supplies only one flag, so `--fields`/`--columns` precedence was untested. It is also **not at parity**:

```console
-S Classes    --tsv --fields BogusA --columns BogusB  ->  Error: No columns matched projection: BogusB   exit 1
-S "API Info" --tsv --fields BogusA --columns BogusB  ->  Note: 1 field has no data: BogusA              exit 0, empty stdout
```

Each flag **alone** is at parity — that is what the test gates. Only the combination diverges, and toward exit 0 with empty output, the worst direction.

Root cause is **markout**, not this repository: `No columns matched projection` appears nowhere in our product source, and markout skips the column-match check once a field filter has emptied the row set. A two-column fact table empties; a multi-column type table narrows columns instead, so its check still runs. The existing fact tables (`library -S "Library Info"`, `type <T> -S "Type Info"`) never show it because `ValidateProjection` runs *before* rendering there — the pre-check the listing view has never had, which is #3651, now carrying this as a second symptom with full reproduction.

I did **not** compensate in `ApiCommand`. Doing so would put a second implementation of projection semantics in the command layer, and the divergence is left **unpinned** on purpose: pinning it as expected would freeze a defect and make a markout fix fight a test asserting it. It is recorded in the test's doc comment instead.

### MAI-Code (second family) — one blocking finding, **not upheld**

> `-S "API Info" --json` ignores the selector and emits the whole `ApiSurface`.

The behavior is real and correctly observed, but it is neither new nor specific to this section:

| invocation | output |
| --- | --- |
| `-S Classes --json` | whole document |
| `-S Structs --json` | whole document |
| `-S "API Info" --json` | whole document — **byte-identical to `-S Classes --json`** |
| `-S Classes --json` on the **parent slice** binary | whole document |

And it is the documented contract — `docs/design/output-shapes.md:399`: `--json` renders "the whole Document when no narrower shape is selected, otherwise the projected payload (`--print`, `--value`, `--urls`, `--paths`)". Section selection is not a payload projection. GPT observed the same behavior in the same round and recorded it as expected.

So: pre-existing, universal across sections, unchanged by this PR, and matching the written contract. Not folded into this slice. The two reviewers disagreeing about whether that contract is *right* is worth something, and I have asked GPT whether it wants to argue the contract is wrong — if so it becomes a design question of its own, not a change here.

MAI also confirmed by mutation that disabling the tabular routing fails `Type_Listing_ApiInfo_ProjectsTheSameRowsInEveryMachineMode`, and verified `ExplicitOnly` across defaults, `-v:q`, `-v:m`, `--all`, `member`, and the `api` shim. It honestly reported **not** reviewing the PR body or docs; GPT did.

### Head moved

`0a937f219` is test-file-only (the strengthened assertion and the divergence note). A confirmation round at that head is running now. Full suite there: **2790 total, 1 failed** (`Layout_Count_CountsFilesRatherThanRenderedTreeLines`, pre-existing on unmodified `main`), 14 skipped.

richlander:
## Round 5 (confirmation at `0a937f219`) — clean. Ready to merge.

GPT-5.6, `/tmp/rev-3648-e` at the exact final head: **no blocking findings, no non-blocking findings**, clean worktree.

It did not take my root-cause attribution on trust — it **decompiled markout 0.33.0** and located the mechanism more precisely than I had:

- `MarkoutWriter.WriteFieldsTable` **returns early when field projection yields zero fields**, before column resolution runs at all.
- `MarkoutProjection.ComputeColumnMap` is what raises `No columns matched projection`.

So the column projection is never *evaluated* on a fact table whose field filter matched nothing — not "evaluated and passed", which is what my commit message implied. A multi-column type table never hits that early return, which is exactly why only the two-column fact table diverges. Recorded on #3651, including the point that the pre-check masks the early return rather than repairing it, so markout is still worth fixing on its own merits.

It also agreed with both dispositions I asked it to challenge: leaving the divergence unpinned and uncompensated (pinning would freeze a dependency defect; a local special case would make `API Info` diverge from the listing-wide projection model), and treating `--json`'s whole-document output as pre-existing and contractual rather than a blocker.

### Review ledger

| --- | --- | --- | --- |
| 1 | GPT-5.6 | `99172e778` | 2 blocking — both real, both fixed |
| 2 | Gemini 3.1 Pro | `99172e778` | **Discarded** — ran in the primary checkout on `main`, not its worktree |
| 3 | GPT-5.6 | `75f1cac7f` | Clean; 3 non-blocking, all real, all acted on |
| 4 | MAI-Code | `75f1cac7f` | 1 blocking — **not upheld** (pre-existing, universal, documented) |
| 5 | GPT-5.6 | `0a937f219` | **Clean** |

Two families, three clean rounds, the final one at the exact head.

### What review actually caught

Every finding was in my claims or in a path I had not exercised — never in the core design. Three of them were the **same shape**: an empty set read as "no constraint". The tabular view's kind filter (the defect this PR fixes), the bypassed `DiagnoseRendered` (round 1), and my own non-vacuity assertion (round 3), which passed on 54 lines of class rows and would have stayed green with both diagnostics deleted. Writing an assertion that *names* what it wants, rather than one that checks something is non-empty, is the lesson I would carry forward from this slice.

### Validation at `0a937f219`

- Full suite **2790 total, 1 failed, 14 skipped** — the failure is `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on unmodified `main` (a local NuGet cache lowercases `newtonsoft.json.nuspec`).
- Differential vs the parent slice: **126 invocations, 9 diffs, 3 vacuous**; every diff a pure addition under `-S @All`, `-D`, and `-D --schema`, with no removed lines anywhere.
- Four mutations each fail their own gate: `ExplicitOnly` off, value drifted off the inline line, `SizeClass` demoted, tabular routing disabled — plus dropping `DiagnoseRendered`, which fails the new parity test.
- `markdownlint` clean on all three changed docs.

### CI

`gh pr checks 3648` reports **no checks reported**, which is terminal rather than pending for a PR based on `type-bare-select` instead of `main`. MERGEABLE / CLEAN. Evidence for this slice is the local validation above plus the parent slice's CI: **#3613 is MERGEABLE / CLEAN with `ci-required` passing**. I will re-check checks once this retargets `main`, where anything newly scheduled does gate.

**Ready to merge**, after #3613.

Non-blocking follow-ups, none gating this PR: #3651 (the listing view's missing pre-render projection validation, now carrying the both-flags divergence as a second symptom) and the markout early-return behind it. Slice 4d wires the listing's bare `-S` to this section and fixes the prefix-selection residual documented in the PR body.

Issue #3644, opened 2026-07-31
Declaration index: compute a C# file's shape once instead of scanning backward per request


richlander:
## Problem

Authored-source extraction used to recover one member's boundaries for every request by scanning
backward and forward from its portable-PDB line range. The recovery was bounded, duplicated C#
structure already available from the lexer, and accumulated special cases for constructors,
destructors, accessors, primary constructors, positional records, and malformed input.

That design produced invalid fragments in real source:

- multiline signatures and long leading trivia could exceed the backward window;
- accessor ranges could return half a property, which is never a C# declaration;
- constructor sequence points on a signature line could fall through to the enclosing type;
- guessed forward boundaries could under-capture or return malformed C#.

## Layering

`CSharpText` owns the declaration index. It is a dependency-free lexical-text layer: it recognizes
declaration shapes and spans without SRM, portable PDBs, SourceLink, Roslyn, or metadata models.

`DotnetInspector.CSharpBodySlicer` is a consumer above it. It combines a portable-PDB line supplied
by its caller with `CSharpText.DeclarationIndex`; it does not own C# declaration recognition.

This corrects the original proposal, which placed the index in the body-slicer project. The index
moved to `CSharpText` as part of the model-free C# layering work.

## Delivered substrate

PR #3652 added one forward lexical pass that emits a row for every declaration:

- declaration kind and source name;
- trivia, signature, body, and end lines;
- parent/depth;
- member-applied attribute-list ranges;
- `SpanKnown`, which refuses spans the lexical evidence cannot vouch for.

The corpus gates compare the table against Roslyn as an independent test oracle; Roslyn is not a
product dependency. PR #3680 then restored knownness after lexically decidable conditional groups,
while preserving conservative refusal inside unresolved branches.

## Final wiring

PR #3927 completes the consumer migration.

Authored-source slicing now:

1. builds `CSharpText.DeclarationIndex`;
2. selects the innermost known declaration containing the PDB start line, including signature
   lines;
3. rejects type/namespace rows and unknown spans;
4. slices the declaration's `[SignatureStartLine, EndLine]`;
5. removes non-code text before the first code token on the declaration line and dedents.

Selection is positional rather than name-based. Metadata names such as `.ctor`, `get_P`, and
`set_P` do not match their source declaration names reliably. Constructor identity is used only
as a kind discriminator: initializer sequence points belong to constructors, so a constructor
request must identify a known constructor declaration at one flattened range boundary.

### Intentional behavior

[...]

PR #3652, opened 2026-08-01
Compute a C# file's declaration shape once instead of scanning backward per request

richlander:
## What

Adds `DeclarationIndex` to `DotnetInspector.CSharpBodySlicer`: one forward pass over the scanner's existing token stream that produces one row per declaration — kind, name, trivia start, signature span, body start, end line, parent.

This is the substrate from #3644. **Nothing consumes it yet.** `BodySlicer` is untouched and all 330 existing tests are unchanged, so there is no behavior change to review — only a new, gated capability.

## Why

`BodySlicer.ExtractMethodBody` finds a member's declaration by scanning backward from its body inside a fixed 15-line window, re-deriving file structure from text on every request. That window is measurably too small: `GraphLowering.ToTree` in the peer `markout` repo has **30 lines** of XML docs above its signature, so the scan cannot reach the declaration it is looking for.

A PDB answers *where the body is* exactly, by row correspondence rather than by search. It says nothing about where the *declaration* starts, because the signature, its attributes, and its doc comments all sit above the first sequence point. Computing the file's shape once turns that second question into a column of a row.

## What it is not

Not a parser. No syntax tree, no type resolution, no semantic model, no Roslyn on any product path. It is a lexical scan plus declaration recognition and it is heuristic at the margins by construction. The value over the scan it will replace is not that the heuristics are gone, but that they run once, in one place, into an artifact whose whole-file correctness can be asserted — which is exactly what the gate below does.

Where the lexer loses its place (unterminated literal; a conditional directive whose braces may belong to a discarded branch) rows report `SpanKnown` false rather than a plausible wrong span.

## Evidence

| Gate | Claim |
| --- | --- |
| `EveryDeclarationRoslynReports_IsReportedIdenticallyByTheIndex` | kind, name, first line, last line identical to Roslyn over **124 files / 3416 declarations** of real repository source |
| `ADeclarationsTriviaStart_MatchesRoslynsLeadingTrivia` | leading trivia — the reason the change exists — matches Roslyn on every one of those declarations |
| `RowsNestWithinTheirParentAndNeverOverlapASibling` | rows nest, never escape their parent, and `Depth` agrees with the parent chain |
| `ASpanTheScanCannotVouchFor_ReportsUnknown` | an unresolvable region reports unknown rather than a guess |

Corpus is discovered the same way the existing parse-validity corpus is: every PDB beside the test binary names the source files its assembly was built from. Roslyn is a test-only oracle, as it already is for #3365.

**The decline path, stated plainly:** the oracle declines a file whose conditional compilation makes the two answer different questions — Roslyn parses with no preprocessor symbols and reports no declarations from a `#if` body, while a lexical scan indexes the text it can see. Neither is wrong. **No file in the corpus currently trips it**, so the corpus cannot distinguish a rule that declines correctly from one that declines nothing; `TheOracleDeclines_OnAConditionalDirectiveAndOnlyOnOne` is the gate that can. The 10% skip ceiling and the file/declaration floors remain as guards against an oracle that quietly starts comparing nothing.

## What building the differential found

Eight defects that eyeballing a prototype had missed. Each is fixed by a stated rule, and **every rule is mutation-checked — removing any one turns the differential red, with no survivors**:

1. A constructor initializer read as the parameter list, so `C(int x) : this(x, 0)` was named `this`.
2. A field's brace initializer read as a property body.
3. A `;` inside a lambda inside an initializer read as the declaration's terminator, losing the declaration entirely.
4. `@namespace = x;` inside a method body read as a namespace declaration — "no enclosing row" conflated file scope with an anonymous scope. The fix is the single rule that also keeps local functions and lambdas out of the index without recognizing either.
5. A `delegate*<int, int>` function pointer read as a delegate declaration.
6. A C# 14 `extension` block read as a method, which hid every extension member inside it. It is now a transparent scope, matching metadata, where extension members are emitted onto the enclosing static class.
7. Commas inside a collection initializer read as declarator boundaries, inventing fields named `0x3f`.
8. Multi-declarator fields collapsing to one row named for the *last* declarator. Metadata sees three fields for `int A, B, C;`, so the index now owes three rows — and fields are exactly the declaration-only members a name lookup exists to serve.

## Validation

- `DotnetInspector.CSharpBodySlicer.Tests` **340/0** (330 pre-existing, unchanged, plus 10 new)
- `DotnetInspector.Services.Tests` **357/0**
- `dotnet-inspect.Tests` **2768, 1 failed** — `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, the known flake, identical at a clean `main`
- `dotnet build dotnet-inspect.slnx -c Release` clean

## Residual

Wiring `ExtractMethodBody` onto the index is deliberately **not** in this PR. That is a user-visible output change — it decides whether XML docs enter a slice by default — and it deserves its own claim and its own evidence. The 15-line backward scan is untouched here.

Refs #3644


richlander:
## Round 1 reconciliation — `2184da15b` → `85a47b002`

Two-model tier. **GPT-5.6 Terra** and **Gemini 3.1 Pro**, isolated worktrees, fixed head `2184da15b`. **Both BLOCK.** Five distinct findings, **all five real, all five fixed.** No dismissals this round.

| # | Reviewer | Finding | Severity | Resolution |
| 1 | Gemini | `Action<string, int, float> A;` splits into two declarators, emitting a hallucinated `Field string` | Critical | Fixed |
| 2 | Gemini | `@class` / `@delegate` match keyword tests → a field becomes a type declaration | Critical | Fixed |
| 3 | Gemini | `@where` truncated by the constraint cut → a property is misread | High | Fixed |
| 4 | GPT | the differential's `Diff` used `Except` — a **set** difference | High | Fixed |
| 5 | both | the oracle declined files by substring-matching `#if` in raw text | Medium | Fixed |

All in `85a47b002`.

### 1–3: verbatim identifiers were read as keywords

One root cause with three symptoms. The scanner emits `@` as its **own token**, so `@class` reaches `Classify` as a `Word` whose text is `class` — indistinguishable from the keyword to every test that read only the word. Reproduced all three.

Every keyword test now routes through `IsKeyword`, which requires that no `@` precedes the word: `TypeKeywords`, `delegate`, `event`, `this`, `base`, `using`, `extern alias`, `namespace`, `operator`, and `Truncate`'s `where` cut. The name is still recovered as `class`, matching Roslyn's `Identifier.ValueText`.

Gated by `AnIdentifierThatSpellsAKeyword_IsNotThatKeyword`.

### 1: and the declarator rule I claimed did not need angle brackets

Gemini disproved a claim in my own doc comment. I wrote that the lookahead — a depth-0 comma is a boundary only when followed by a name and then `,`/`=`/end — made angle tracking unnecessary, and cited `Dictionary<string, int> map` as proof. **Two type arguments happen to survive it. Three do not.** In `Action<string, int, float>` the comma after `string` is followed by `int` and then another comma: exactly a declarator list's shape.

The fix keeps both tests, because neither suffices alone. Angle depth is tracked **only before the segment's `=`**, where a `<` is always a generic bracket and counting is exact. After the `=` a relational `<` may never close, and counting there would swallow every later comma — so the initializer is left to the lookahead, which is sound there. `>>` closing two nested lists arrives as one token, so the counter counts `<`/`>` **characters** within a punctuator.

The doc comment that asserted the opposite is replaced.

### 4: the gate compared sets, so it could not see cardinality

`Except` is a set operation. A row emitted **twice**, or dropped when a duplicate existed, was invisible to the differential — and identically spelled declarations are real: a partial method's defining and implementing halves share kind, name, and differ only in span.

`Diff` is now a multiset comparison over a count map, reporting `(xN)`. **Verified by mutation:** appending a duplicate span now fails the differential across 106 of 124 files; under `Except` it passed silently. Also gated at the index level by `TwoDeclarationsSpelledAlike_AreBothReported`.

### 5: the oracle was shrinking the corpus invisibly

Both reviewers found this. `ScanTokenTests.cs` spells `#if` **nine times** in comments and has no directive at all. The oracle now asks Roslyn for `IfDirectiveTrivia`/`ElifDirectiveTrivia`/`ElseDirectiveTrivia` via `DescendantTrivia(descendIntoTrivia: true)`.

All three previously-declined files had `#if` only in comments or strings. **The corpus grows from 121 files / 3239 declarations to 124 / 3416, declining zero.**

That creates a second-order problem the reviewers did not raise and I am recording rather than leaving implicit: with zero declines, **the corpus can no longer distinguish a rule that declines correctly from one that declines nothing.** So the rule gets its own gate — `TheOracleDeclines_OnAConditionalDirectiveAndOnlyOnOne` asserts both arms: it declines on `#if`/`#else`, and does **not** decline on `#if` in a comment, in a string literal, or on `#region`/`#pragma`/`#nullable`. The 10% skip ceiling and the file/declaration floors stay as guards from the other side.

The PR body claimed "the oracle declines 3 of 124 files" and "121 files / 3239 declarations". Both are now false; the body is corrected.

### Mutation check on the new rules

Each new rule fails a test when neutralized, individually:

| Mutation | Result |
| --- | --- |
| angle depth ignored (`angle >= 0`) | `CommasInsideATypeArgumentList_AreNotDeclaratorBoundaries` fails |
| `IsKeyword`'s verbatim guard short-circuited | `AnIdentifierThatSpellsAKeyword_IsNotThatKeyword` fails |
| `IsVerbatim` dropped from the type-keyword arm | same test fails |
| a duplicate span appended | the differential fails (and did **not** before the `Diff` fix) |

`? null` and `if (false)` mutations compile away here (CS0173, CS0162), so these are runtime-opaque.

### Validation at `85a47b002`

Slicer **344/0**, Services **357/0**, CLI **2768/1** — the 1 is the known `Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake, identical at a clean `main`. `dotnet-inspect.slnx` builds clean in Release.

**Not ready to merge.** Round 2 is owed at this head from both models.

richlander:
## Round 2 reconciliation — `85a47b002` → `e27d33700`

**GPT-5.6 Terra** and **Gemini 3.1 Pro**, isolated worktrees, fixed head `85a47b002`. **Both BLOCK.** Four findings; **three real and fixed, one dismissed by measurement.** The two reviewers answered one of them in *opposite directions*, so it is resolved on the merits below rather than by vote.

| # | Reviewer | Finding | Severity | Resolution |
| 1 | GPT | `extension<T>(...)` indexed as a method, **erasing every member inside it** | High | Fixed |
| 2 | Gemini | `new Action<int, int, int>(null), B = null` invents a `Field int` | High | Fixed |
| 3 | both, opposed | checked operator naming | Medium | Fixed — **Gemini's direction** |
| 4 | Gemini | modifiers/attributes before `extension` break the scope | High | **Dismissed by measurement** |
| 5 | *neither* — mutation battery | the speculative match carried an **unreachable** condition | — | Removed |

All in `e27d33700`.

### 1: generic extension blocks erased their own members

Reproduced. The transparent-scope check required `(` immediately after `extension`, so `extension<T>(...)` fell through to `Classify` and became `Method extension`.

The cost is not one bad row. That method-shaped row then becomes the enclosing scope, and `Allowed` rejects a member whose enclosing declaration is not a type — so **`Page` and `IsEmpty` disappeared from the index entirely**. Before/after on the same fixture:

```text
before:  Class C / Method extension          (members gone)
after:   Class C / Method Page / Property IsEmpty
```

`DeclaresAnExtensionBlock` now accepts an optional balanced type parameter list between the keyword and the receiver. Gated by `AGenericExtensionBlock_IsTransparentJustLikeAPlainOne`, which also asserts each member's parent is the enclosing class.

### 2: round 1's Critical defect, on the other side of the `=`

Reproduced. Round 1 fixed generic commas in the **declarator**; Gemini found the same class of defect in the **initializer**, where I had deliberately turned angle counting off because a relational `<` never closes.

Turning counting back on is not available. Instead a `<` that follows a name is matched **speculatively** by `TypeArgumentListEnd`, which accepts only tokens a type argument list can contain — words, `<` `>` `,` `.` `?` `[` `]` `*` `(` `)` `::` — and the matched region is skipped. **The exclusion is the discriminator:** `a < b, y = c > d` contains an `=`, and no type argument list does, so it is left alone and its comma still separates declarators.

`CommasInsideAGenericInitializer_AreNotDeclaratorBoundaries` gates both directions — the type argument list must not split, *and* the two relational rows must not be lost, which would be as wrong as inventing one.

### 3: the reviewers disagreed, so the merits decide

GPT read the **index** as wrong: it emitted `operator checked+` where the oracle said `operator +`. Gemini read the **oracle** as wrong: it dropped `checked` entirely, and the gate passed only because no corpus file declares a checked operator.

**Gemini is right.** `operator checked +` emits `op_CheckedAddition`; `operator +` emits `op_Addition`. They are two distinct metadata members and **both may be declared in the same type** (verified — the fixture compiles). A name that cannot tell them apart discards information the index already has, in the one position where it costs nothing to keep. That also matters downstream: distinguishing them is exactly selector B's job.

GPT was right about the symptom — `operator checked+` was malformed, and it diverged from the oracle *silently*, which is the defect that outlives either spelling. Both sides now derive the name **independently**: the index from the token after `operator`, the oracle from Roslyn's `CheckedKeyword`. Conversions follow the same rule (`operator checked explicit`).

The `Name` contract in `DeclarationIndex.cs` is updated, and now also states plainly that a name is **not an identity** — a conversion carries no target type, so two conversions from one type share a name, and `FindByName` returns every match.

### 4: dismissed by measurement

Gemini's fixture was `public extension MyExt for int { }`. That is not C#; it is an earlier design proposal. The compiler rejects the whole family:

| Form | Compiler |
| --- | --- |
| `public extension MyExt for int` | **CS9281** extension declarations may not have a name |
| `public extension(int r)` | **CS0106** the modifier 'public' is not valid for this item |
| `[Obsolete] extension(int r)` | **CS7014** attributes are not valid in this context |

So `extension` is always the header's first token and there is nothing for it to follow. Recorded in `DeclaresAnExtensionBlock`'s doc comment, where the assumption lives, so nobody re-derives it.

### 5: what the reviewers missed and the mutation battery caught

My fix for finding 2 shipped with a second condition — that the closing `>` not be followed by an identifier — which I justified at length in a doc comment. Neutralizing it **broke no test.**

It is unreachable. For it to change any outcome the matched region must contain a comma the lookahead accepts, which needs the shape `<a, b, c> name` — and the compiler rejects that as a syntax error (CS1003/CS1002/CS1519, measured). Removed. The measurement is recorded in the comment instead, so the next reader does not re-add it.

This is the second time in this PR that a claim in my own doc comment was the defect.

| Mutation | Result |
| --- | --- |
| generic-extension arm removed | `AGenericExtensionBlock_...` fails |
| `checked` dropped from the name | `ACheckedOperator_...` fails |
| speculative match disabled | `CommasInsideAGenericInitializer_...` fails |
| identifier-after-`>` condition removed | **nothing failed** → condition deleted |

### Validation at `e27d33700`

Slicer **347/0**, Services **357/0**, CLI **2768/1** — the known `Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake, identical at a clean `main`. `dotnet-inspect.slnx` builds clean in Release.

**Not ready to merge.** Round 3 is owed at this head from both models.

richlander:
## Round 3 reconciliation — `e27d33700` → `627835441`

**GPT-5.6 Terra** and **Claude Opus 4.8** (fresh family, deliberately rotated in), isolated worktrees, fixed head `e27d33700`. **Both BLOCK.** Four findings, **all four real and fixed** — and the fixture written to close the smallest of them uncovered two more defects nobody had reported.

| # | Reviewer | Finding | Severity | Resolution |
| 1 | **both** | file-scoped namespace `EndLine` ran to the last physical line, diverging from the oracle | Medium | Fixed |
| 2 | GPT | the same row stayed `SpanKnown` even when the scan lost its place | Medium | Fixed |
| 3 | Claude | the oracle's `EndLine` doc comment claimed an agreement it never implemented | Medium | Fixed |
| 4 | Claude | `TypeArgumentListEnd`'s allow list was ungated | Low | Fixed |
| 5 | *the fixture for 4* | `global::` broke the declarator split | High | Fixed |
| 6 | *the measurement for 5* | comments claiming `>>` arrives as one token were false | — | Fixed |

All in `627835441`.

### 1: both reviewers found it independently

A file-scoped namespace **scopes** the rest of the file, but its **declaration** ends where its last member ends. A trailing comment or blank line belongs to the file, not to the namespace, and Roslyn's span says so.

The honest part of this finding is not the off-by-N, it is that **the gate was green by corpus luck**. No file in the 124-file corpus has anything after the last member's closing brace, so `EveryDeclarationRoslynReports_...` never saw the divergence — and adding a trailing comment to any repository file would have turned CI red for a reason unrelated to the change that added it. Claude reproduced exactly that with the gate's own `Format`/`Diff`.

The row now ends at the last line of the last thing it contains, and at its own signature line when it contains nothing. `AFileScopedNamespace_EndsAtItsLastMemberNotAtTheLastLine` covers both, with the trailing comment *and* trailing blank lines the corpus lacks.

### 2: a span that reaches the whole file cannot be better known than the file

The EOF-closing special case read:

```csharp
if (!r.ClosesAtEndOfFile)
    r.SpanKnown = false;
```

The exemption is right about *why* the row is still open — the file ended, as expected — but it also skipped the unknown-span marking, so:

```text
namespace N;          Namespace N  known=True   ← the guess
#if FEATURE
class A { }           Class A      known=False
#else
class B { }           Class B      known=False
#endif
```

That is precisely the guess `SpanKnown` exists to prevent, and it is design constraint 3. The builder now tracks whether the scan ever lost its place and marks the EOF-closing row unknown when it did. `AFileScopedNamespaceOverAConditionalRegion_ReportsAnUnknownSpan` gates it.

### 3: the third false doc claim in this PR

The oracle's `EndLine` summary said the index reports EOF *"which is what the index reports too"*, and described a file-scoped-namespace compensation the oracle never implemented. Both halves were false. Replaced with what the code does.

That is now **three** false doc comments in this change — two caught by reviewers, one by the mutation battery. Every one of them sat on top of code I had reasoned about carefully. The pattern is worth naming: the comment records the reasoning I *did*, and survives the code changing underneath it.

### 4 → 5: the gate found a defect the reviewers did not

Claude showed the allow list could be shrunk to `","` and `"."` with the suite still green. Writing the fixture to gate it — one declarator list per type-argument shape — immediately failed on a shape nobody had tested:

```csharp
public object i = new Func<global::System.Guid, string, int>(null), j = null;
//                                                          ↑ invented: Field int
```

**The scanner emits one-character punctuators**, so `::` arrives as two `:` tokens. The `"::"` entry I wrote was dead, and `":"` — the token that actually appears — was missing. This is a High-severity hallucinated row on ordinary C#, reached only because the low-severity missing-gate finding was taken seriously.

`ATypeArgumentListInAnInitializer_MayContainAnyTypeSyntax` now exercises every entry: array, tuple, pointer, qualified name, alias-qualified name, nullable, and nested generic.

### 6: and the measurement falsified two more comments

Dumping the token stream to confirm how `::` is lexed also disproved the comments claiming `">>" closing two nested type argument lists arrives as one token`, which justified counting `<`/`>` *characters* inside a punctuator. Every punctuator is one character; `>>` is two tokens. The counting is replaced with plain comparisons and the claim with the measurement.

### Where the reviewers tried and failed

Worth recording, since it is evidence about the parts that did **not** move. Claude ran fixtures — all verified to parse clean — across constructor initializers, generic constraints, expression bodies vs lambda-valued fields, multi-declarator fields with generics/tuples/arrays/nullable, tuple returns, every operator form including `>>>`/`true`/`false`, indexers, destructors, static constructors, `delegate*<>`, plain and generic extension blocks, `extension` as an ordinary identifier, nested/file-scoped/block namespaces, raw and interpolated strings with code in the holes, verbatim names, attributes with `[return:]`/`[assembly:]`/arrays/parens, explicit interface implementations, records in every spelling, and field-like vs block events — with **no divergence from Roslyn**. It also could not construct a compilable initializer where a relational `<…>` is wrongly matched, and confirmed `SpanKnown` is neither over- nor under-applied around directives.

### Mutation battery

| Mutation | Result |
| --- | --- |
| namespace end back to `lines.Count` | `AFileScopedNamespace_EndsAtItsLastMember...` fails |
| `depthLost` ignored | `AFileScopedNamespaceOverAConditionalRegion_...` fails |
| allow list shrunk to `,` and `.` | `ATypeArgumentListInAnInitializer_...` fails |
| allow list with only `:` removed | `ATypeArgumentListInAnInitializer_...` fails |

### Validation at `627835441`

Slicer **350/0**, Services **357/0**, CLI **2768/1** — the known `Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake, identical at a clean `main`. `dotnet-inspect.slnx` builds clean in Release.

**Not ready to merge.** Round 4 is owed at this head.

richlander:
## Round 4 reconciliation — `627835441` → `59eb980ee`

Two models, both BLOCK. **Four findings, all four real, all four fixed.** No dismissals this round.

| # | Reviewer | Finding | Sev | Outcome |
| 1 | Gemini 3.1 Pro | a speculative type argument list could close with a `(` or `[` still open | High | fixed in `9f83b43ff` |
| 2 | Gemini 3.1 Pro | a file-scoped namespace adopted an unclosed member's guessed end and still reported `SpanKnown` | Med | fixed in `9f83b43ff` |
| 3 | Claude Opus 4.8 | `Delegate` classification, function-pointer disambiguation included, was entirely ungated | Med | fixed in `59eb980ee` |
| 4 | Claude Opus 4.8 | `Destructor` classification was entirely ungated | Med | fixed in `59eb980ee` |

Every finding was reproduced at the reviewed head before anything was changed, and every triggering
fixture was compiled first.

### 1 — an unbalanced group inside a speculative type argument list

`TypeArgumentListEnd` accepted a closing `>` without checking that the groups opened inside the
region were closed. `a < b(name: c > d)` has exactly that shape, so the caller skipped past the `(`
and then met the matching `)`, driving its own group depth to `-1` — after which **no later comma
could ever separate a declarator** and every trailing declarator vanished from the index. Three
distinct shapes reach it, all of which compile:

```csharp
static bool a = x < b(name: c > d), f = true;      // f lost
static int  g = x < p ? q : m(r > s), h = 1;       // h lost
static bool k = x < arr[c > d ? 1 : 0], n = false; // n lost
```

This is the third recurrence of the same underlying hazard — round 1 found the generic comma before
the `=`, round 2 found it after the `=`, and this is the group depth the speculative match hands
back to its caller. Requiring the groups balanced at the closing `>` also strengthens the
discriminator: a real type argument list always balances, `a < b(...)` does not.

### 2 — doubt that did not propagate

A file-scoped namespace derives its end as a maximum over the rows it encloses, so it is only as
good as the worst of them. Round 3 marked it unknown when *lexical* depth was lost; an unclosed
brace never loses lexical depth, so `namespace N; class A {` gave the namespace a measured span
built entirely from `A`'s guess. The maximum now carries the doubt of the rows it reads.

### 3 and 4 — green by corpus luck, again

Claude could not break the recognition logic — ~35 adversarial fixtures across constructor
initializers, constraints, every operator form, explicit interface implementations, function
pointers, records, raw strings and preprocessor regions all matched Roslyn — and instead attacked
the gate, which is where it found something. The 127-file corpus contains **zero delegate
declarations and zero destructors**, and there was no synthetic test for either, so
`DeclaresADelegate` and the `~` branch could each be **deleted outright with all 352 tests green**.

One fixture now carries both, with the close negatives that make them discriminators rather than
string matches: a `delegate*<int, int>` parameter and an anonymous method in a field initializer
both spell `delegate`, and neither declares a type.

### The minor observation, and what measuring it changed

Claude also noted the `extern alias` skip is removable with the suite green. That is true — and
measurement showed **it is equally true of the `using` skip**, for a reason worth writing down
rather than gating: `Allowed` already rejects any member kind whose enclosing scope is not a type,
and neither construct can appear inside one. Both skips keep `Classify` locally right without
changing any result.

So the fix here is not a test. My first attempt *was* a test, whose doc comment claimed it gated the
branch — and the mutation showed that claim was false before it was pushed. That would have been a
**fourth** false doc comment on this PR. The code comment and the test's doc comment now both say
what actually enforces the property.

### Validation at `59eb980ee`

Slicer **354/0**, Services **357/0**, CLI **2768/1** — the known
`Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake, identical at clean `main`. `slnx`
builds clean in Release.

**Seven mutations run this round**, each failing exactly the intended test: dropping the group
balance requirement; making `)` not close a group; narrowing the namespace's doubt; neutering
`DeclaresADelegate`; removing its `*` check; removing the destructor branch; and the two
skip-removals above, which are the two that stayed green and are documented as such.

Round 5 to follow at `59eb980ee`.

richlander:
## Round 5 reconciliation — `59eb980ee` → `a5d40bc0a`

**GPT-5.6 Terra: BLOCK, one finding, real, fixed. Claude Opus 4.8: APPROVE.**

Because the head moved, round 5 does not close the PR — the fixed head needs its own clean round.

### The finding — a comment attributed to the wrong line

Trivia attribution used a comment token's own line. A block comment yields **one scanner token per
line it covers**, so this:

```csharp
int A; /* trailing
comment */
int B;
```

opened on `A`'s terminator line — correctly trailing `A` — while its **second** line sat past that
terminator and became `B`'s trivia start. `B.TriviaStartLine` was 4 where Roslyn says 5, and a slice
taken from there begins inside the comment, past its `/*`, producing source that does not compile.
Severity Medium, and it is the failure mode this component exists to prevent.

The fix attributes by where the comment *opened*. That required measuring rather than reasoning:
dumping tokens shows a continuation line **may itself start with `//` or `/*`**, so the opening
token cannot be identified from text and the carried block-comment state is reconstructed instead.
`/*/` is an unterminated comment rather than a closed one, which is why the opening test carries a
length guard.

Roslyn is the oracle for the new test rather than hand-written line numbers, because the question is
what counts as leading trivia and that is Roslyn's answer to give. Nine fixtures, all compiled: the
trailing and leading multi-line forms, contiguous line comments, a block containing `//`, a block
containing `/*`, two blocks on one line, and both `/*/` positions.

**Four mutations. Two of them stayed green on the first attempt** — a block that never closes, and
dropping the `/*/` guard — because my original fixtures only made those sub-rules shift *state*
without changing an *answer*. The last two fixtures exist to close exactly that, and all four go red
now. This is the fourth round in a row where the mutation battery found that a fix I had just
written was under-gated.

### What the second reviewer confirmed

Claude re-ran the round-4 gates independently and reproduced all four as red under mutation, and for
the group-balance test confirmed **each of the three shapes independently** loses its trailing
declarator when the check is removed, so that test is not passing on the strength of one case.

It also pressed the two round-4 judgement calls rather than accepting them:

- The `extern alias` redundancy claim was traced through the mechanism, not just observed:
  `extern alias LibA;` classifies to `(Field, "LibA")` and `Allowed` rejects it because
  `IsTypeKind(Namespace)` is false and an extern alias can never sit inside a type.
- The `guessed` flag was checked for **over**-marking, which was the risk of that fix. A row can
  have `SpanKnown == false` while `depthLost == false` only via the unclosed-brace path, which means
  the file is truncated and the namespace's maximum genuinely reaches a guessed line. Verified
  empirically both ways: valid files with trailing comments still report a **known** namespace span
  ending at the last member, and the round-3 trailing-comment behavior still holds.
- The group counter rejects `group != 0` but not `group < 0`. Claude tried to reach a
  negative-then-zero false match and could not, for a structural reason: the speculative `<` is only
  entered at caller depth 0, so no top-level `)` can precede its `(`. Recorded as defensible rather
  than fixed — adding an unreachable guard is what round 2 removed.

Sixteen further adversarial fixtures against its own Roslyn oracle — tuple-typed generic fields,
`global::` in both a type argument and a relational, `x < y ? f > g : h > x`, `delegate*<>` in field,
parameter and return position, positional records with primary-constructor base lists, nested
generic `extension<T>` blocks with constraints — all agreed exactly.

### Noted, not actioned

The non-vacuity floors (`files >= 100`, `declarations >= 2500`) sit about 20% below the current
corpus. That slack is deliberate: the floors catch a corpus that *collapsed*, and the `skip <= 10%`
ceiling catches the subtler oracle-decline failure. Tightening them to the current numbers would
make the gate fail on any unrelated file deletion, which is noise rather than signal. Raised by the
reviewer explicitly as a non-finding, and recorded here as an explicit non-action.

### Validation at `a5d40bc0a`

Slicer **355/0**, Services **357/0**, CLI **2768/1** — the known
`Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake, identical at clean `main`. `slnx` builds
clean in Release.

Round 6 to follow at `a5d40bc0a`.

richlander:
## Round 6 reconciliation — `a5d40bc0a` → `1ba16ebd2`

Both models BLOCK. **Two findings, both real, both fixed.**

| # | Reviewer | Finding | Sev | Outcome |
| 1 | GPT-5.6 Terra | an `[assembly:]` / `[module:]` list became the next declaration's trivia | Med | fixed |
| 2 | Gemini 3.1 Pro | round 5's comment opening test was gated in **one direction only** | Med | fixed |

### 1 — a compilation-unit attribute is not a declaration's trivia

```csharp
using System;
[assembly: CLSCompliant(true)]
class C { }
```

`C.TriviaStartLine` was 2 where Roslyn says 3. A slice would open with an assembly attribute that
has nothing to do with the member selected.

Measuring Roslyn rather than assuming produced the part I would have got wrong: a unit attribute has
to **clear** trivia that came before it, not merely decline to extend it. Given a file header comment
above `[assembly: …]`, Roslyn puts the comment inside the *attribute list's* trivia, so the following
class's trivia starts after the list.

The discriminator is the colon, not the word: `[assembly]` applies a type called `assemblyAttribute`
and **is** ordinary trivia. Eleven fixtures, all compiled, carry both readings and the close
negatives `[Obsolete]`, `[type: Obsolete]`, and `[return: …]`, which must still be kept.

### 2 — a gate that only worked in one direction

Round 5's opening test is `StartsWith("/*") && !(Length >= 4 && EndsWith("*/"))`. Round 5 gated
treating `/*/` as *closed*. It did not gate the opposite: treating a closed one-line `/* … */` as
still **open**.

```csharp
class C {
    /* single line */ int A;
    /* next comment */
    int B;
}
```

Under that mutation the second comment reads as a continuation, `commentOpenLine` stays on line 2,
and `B` loses its trivia start entirely. One fixture closes all three mutation directions.

This is the **third** consecutive round in which a reviewer found the previous round's fix
under-gated — round 4's finding was two ungated rules, round 5's mutation battery caught two
sub-rules mid-fix, and round 6 caught the direction that battery missed. Writing the mutation is
evidently much easier than choosing an input that makes the mutation *change an answer* rather than
merely shift state, and that gap is where all three landed.

### One condition removed rather than gated

The attribute-target check originally required the token to be a `Word`. That condition is
unreachable in effect: literals, comments and directives are skipped before that point, and every
punctuator the scanner emits is one character, so no non-`Word` token can spell `assembly`. Deleted,
on the round-2 precedent — an unreachable condition that looks load-bearing is worse than none.

### Also examined and found sound

Gemini reports that C# 14 extension scopes, multi-line string literals inside attributes,
interpolated strings spanning lines, and `TypeArgumentListEnd`'s group balancing all handled its
adversarial fixtures correctly.

### Validation at `1ba16ebd2`

Slicer **356/0**, Services **357/0**, CLI **2768/1** — the known
`Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake, identical at clean `main`. `slnx` builds
clean in Release.

**Nine mutations run** across the two rules, each failing exactly the intended test. Two needed a
better fixture first: dropping the colon requirement is pre-empted by `]` in a one-token list, so it
takes a two-token list such as `[assembly()]` or `[module, Obsolete]` to make that rule's removal
observable.

Round 7 to follow at `1ba16ebd2`.

richlander:
## Round 7 reconciliation — `1ba16ebd2` → `19a2eb0b4`

Both models BLOCK. **Five findings, all real, all fixed.** This round also folds in a capability the
findings argued for: the index now reports a declaration's **attribute lists**.

| # | Reviewer | Finding | Sev | Outcome |
| 1 | Claude Opus 4.8 | a comment trailing `[assembly: X]` became the next declaration's trivia | High | fixed |
| 2 | Claude Opus 4.8 | `[@assembly: X]` was not recognized as a unit attribute | Med | fixed |
| 3 | Claude Opus 4.8 | `[Obsolete][assembly: X]` cleared the real attribute's trivia | Low | fixed |
| 4 | GPT-5.6 Terra | `union` declarations were omitted entirely | High | fixed |
| 5 | GPT-5.6 Terra | a verbatim namespace name disagreed with the oracle | Med | fixed, in the oracle |

### Attribute lists are now data, not a collapsed cell

Findings 1–3 have one root: every attribute list was collapsed into a single `triviaStart`
integer, so the only thing the pass could do with a compilation-unit list was destructively clear
that cell — and clearing it took its neighbours with it.

`DeclarationSpan.AttributeLists` now reports one `LineRange` per **member-applied** list, in source
order. `[assembly:]`/`[module:]` lists are excluded by construction. Ranges, not parsed attributes:
the names and arguments are metadata facts, and this layer reports where the authored text sits so
a caller can slice it. `LineRange` is line-granular by design, matching PDB sequence points.

That makes findings 1 and 3 structural rather than patched. A unit list is only a unit list when it
is the **first** in a run — C# binds every later list to the declaration below, so
`[Obsolete][assembly: X] class A` applies both to `A` (CS0657). And a unit list now ends where its
`]` is, so `[assembly: InternalsVisibleTo("T")] // for tests` has a comment that trails **the
list**, not one that opens `Widget`'s trivia.

Gated by `EveryAttributeListRoslynReports_IsReportedIdenticallyByTheIndex`, a corpus differential
against Roslyn's own `AttributeLists` with its own non-vacuity floor — a corpus can hold thousands
of declarations and few attributes, and a shared floor would let this pass by comparing nothing.

### Finding 4 — `union` lost the type and everything in it

```csharp
public union Result { int Ok; }
```

Roslyn reports a struct; the index reported **nothing**, and `Ok` went with it. Fixing it exposed
the trap that makes `union` different from `class`: it is contextual, so `int union;` is a field
whose *name* is the keyword. Adding it naively produced `Struct ''` and lost the field.

A type keyword must now be followed by the type's name. That also fixes the same latent defect for
**`record`**, which pre-dates this PR — `int record;` indexed as `Record ''` at every earlier head.

### Finding 5 — the disagreement was the oracle's

The index reports names without `@` everywhere. Roslyn's `Name.ToString()` reproduces the source
spelling while a token's `Text` does not, so the oracle demanded `@event.Models` from a namespace
and `class` from a type — two APIs mistaken for one rule.

The product is consistent; the **doc comment** was not. `Name` claimed "as spelled in source",
which is false, and a reader had no way to tell that from a green suite. Names are metadata-aligned
because that is what they are for — `FindByName` correlates a row with a member, and metadata never
carries the escape. Corrected, and gated.

**This is the fourth false doc comment this review has found**, and the second one that survived a
round because the code and the comment disagreed in a direction nothing tested.

### Mutations

**Twelve, in both directions.** One survived and mattered: the round-7 trailing-comment fix was
initially asserted with a formatter that omits `TriviaStartLine`, so the fix was ungated at the
moment it was written — the sixth consecutive round in which mutation caught a just-written fix
under-gated. The assertion now includes trivia.

One mutation is **equivalent**, not a gap: advancing `lastTerminatorLine` for a *non-unit* list is
unobservable, because that path always sets `triviaStart`, which the comment guard then rejects.
Recorded rather than gated with a contrived test.

### Validation at `19a2eb0b4`

`origin/main` merged first, then re-run: slicer **360/0**, Services **357/0**, CLI **2793/1** — the
known `Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake, identical at clean `main`. `slnx`
clean in Release.

Round 8 to follow. The attribute-list reporting is new code that no reviewer has seen.

richlander:
## Round 8 reconciliation — `19a2eb0b4` → `cef356263`

Both models BLOCK. **Three findings, all real, all fixed.** Two are High, and both were invisible to
the corpus gate.

| # | Reviewer | Finding | Sev | Outcome |
| 1 | Claude Opus 4.8 | every operator whose symbol contains `=` was lost and reported as `Field 'operator'` | High | fixed |
| 2 | Gemini 3.1 Pro | round 7's contextual-keyword rule was too weak; `int record, union;` hallucinated a type | High | fixed |
| 3 | Gemini 3.1 Pro | `unitTarget = false` is a dead store | Low | deleted |

### 1 — the operator's name contains an equals sign

```csharp
public static bool operator ==(V a, V b) => a.X == b.X;
```

reported `Field 'operator'`, `SpanKnown = true`, with the real method gone. The header cut that
looks for an assignment cut at the `=` **inside the operator's own name**, discarding the parameter
list — and a header without one is not recognized as an operator at all.

Block-bodied it cascaded: the truncated header made the body look like an initializer, so
`Equals` and `GetHashCode` were swallowed into the fabricated field's span. Three members lost to
one bogus row.

`==`, `!=`, `<=`, `>=` and the C# 14 compound family are affected; `<`, `>`, `+`, `>>>`,
`operator checked +` and conversions were always fine.

**Why eight rounds and ~3400 declarations missed it:** the repo's only equality operators live in
`ILInspector.Decompiler.Tests`, which is not in this test binary's dependency closure, and
`OperatorNames.cs` spells `"operator =="` only inside a string literal. The corpus has never
contained one. Gated by fixture, since the corpus cannot gate it.

### 2 — a word after the keyword is not enough

Round 7 fixed `int record;` by requiring a name to follow the keyword. Gemini showed that is the
wrong discriminator:

```csharp
public int record, union;
public int M(int record, int name) { return 0; }
```

Both put a word after the keyword. Observed `Record union`, `Record int`, `Field 0` — two
hallucinated types, every real member lost, and the statements inside `M` indexed as fields once
the fabricated type adopted `M`'s `{` as its body.

The real rule is that **everything a type declaration spells before its keyword is a modifier**.
The close positives are now every modifier a type may carry, including `record class` and
`record struct`.

**My round-7 doc comment said the name test "is what tells the two apart". That was false** — the
fifth false doc comment this review has found, and the first one that was mine. The name test
survives only as a bounds guard for the indexing below, and now says so, marked **unverified and
ungated** rather than left reading as a rule.

### Two mutation survivors became deletions, not tests

- `record` in the modifier set is **dead**: the scan reaches `record` before the `class` it
  precedes, so no header ever tests `class` with a `record` before it.
- `unitTarget = false` is a **dead store**: the first word of every attribute list assigns it, and
  it is read only after that assignment.

Both removed on the round-2 precedent — an unreachable condition that reads as a load-bearing rule
is worse than no condition. The one guard I kept despite being unreachable is labelled as such,
because deleting it would leave `words[nameAt]` unguarded.

### Also examined and found sound

Claude reports that the round-7 attribute work held under targeted `[return:]`/`[field:]`/
`[method:]`/`[param:]` and type-parameter attributes, multi-line lists, nested brackets, a `]`
inside a string, a comment between two lists, `[assembly:]`-first versus the CS0657 second position,
and positional-record attributes — all matching Roslyn exactly.

### Validation at `cef356263`

Slicer **361/0**, Services **357/0**, CLI **2793/1** — the known
`Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake, identical at clean `main`. `slnx` clean
in Release. **Seven mutations this round, both directions.**

Round 9 to follow at `cef356263`.

richlander:
## Round 9 reconciliation — reviewed at `cef356263`

| --- | --- | --- |
| `rv3652i-gpt` | GPT-5.6 Terra | BLOCK — 1 finding |
| `rv3652i-gem` | Gemini 3.1 Pro | APPROVE |

This round was aimed at one thing: the blind spot round 8 exposed, where the corpus differential
passes because the repository's own source never writes the shape that breaks the code. The brief
asked both reviewers directly what common C# this repository never writes. Both answered usefully,
in opposite directions.

### GPT — a conditional directive loses every later row, to end of file (High) — real, deferred

`LexState.Untracked` is set by any `#if`/`#elif`/`#else` and never cleared, so every token to end of
file is stamped `DepthKnown=false` and every later row reports `SpanKnown=false` — including rows
after the `#endif` that no branch could affect, and even when every branch is brace-balanced.

Reproduced directly:

```csharp
class C
{
#if DEBUG
    void Debug() { }
#endif
    void Always() { }
}
```

```text
Method Always sig=6..6 body=6 end=6 known=False
FindByBodyLine(6) -> <null>
without the #if, FindByBodyLine(3) -> Always
```

Quantified over `dotnet/runtime/src/libraries`, the corpus SourceLink member slicing targets:

| Metric | Value |
| --- | --- |
| Files scanned | 20,838 |
| Declarations | 581,298 |
| Files with a conditional directive | 1,737 (8.3%) |
| **Declarations `SpanKnown=false`** | **70,472 (12.1%)** |
| Files losing *every* declaration | 656 |
| Declarations after the last `#endif` | 36,451 |
| **…unresolvable anyway** | **36,411 (99.9%)** |

A conditional appears in 8.3% of files but costs 12.1% of declarations, because the loss runs to end
of file rather than to the `#endif`.

GPT's diagnosis of why no gate caught it is also correct, and is the round's real result: the corpus
differential *declines* every conditional file, because Roslyn parsing with no symbols defined
discards disabled branches while the index, being lexical, indexes them. The two are not comparable
as equality, so the gate opts out — and nothing else covers the collateral damage after `#endif`.

**Deferred, not fixed here, by decision.** The behavior is conservative — the index reports "do not
know", never a wrong span — and nothing consumes the index yet, so the gap costs nothing until the
wiring PR. Tracked as #3668, with the proposed fix (per-branch brace balance: if every branch of a
conditional group is individually balanced, depth after `#endif` is knowable regardless of which
branch compiles) and the fair gate for it (subset comparison rather than equality, so extra rows
from disabled branches are not failures).

What did change here, in `3a066e6c0`:

- The `SpanKnown` contract said "affected rows", which reads as the guarded region. It now states
  the real reach, gives the measured cost, and names #3668 as a known limitation rather than an
  intended property.
- `AConditionalDirective_LosesEveryLaterRowToEndOfFile` pins the current behavior and is marked as a
  characterization test, so recovering the depth is a visible change rather than a silent one.
  Verified non-vacuous: mutating the directive branch to stop setting `Untracked` fails it.

No product behavior change in this commit — doc comments and one test.

### Gemini — APPROVE, and it closed both open questions from round 8

Round 8 left two things I had flagged as unproven. Gemini attacked both directly rather than
re-reading them:

- **`TypeModifiers` completeness.** This was the one I was least sure of — a missing modifier
  silently loses a type *and every member inside it*. Gemini brute-forced every C# keyword against
  the compiler to find any that is legal as a class or struct modifier and absent from the set.
  None. It also checked `unsafe`, `file`, `new`, `partial`, `readonly`, `ref` explicitly.
- **`inOperatorSymbol` set and never cleared.** Probed with `operator` appearing in fields, lambdas,
  tuples, and explicit interface implementations, block-bodied versus expression-bodied, and
  `operator checked +=`. Cleared correctly in all.

It then built the round-9 blind-spot list into a single fixture — `unsafe` methods and structs,
`stackalloc`, `fixed` buffers, `event` declarations, multi-parameter indexers, nested generic
constraints, partial properties, `required` members, `init` accessors, `scoped` parameters,
collection expressions, explicit interface implementations — and diffed it against the oracle:
exact match, zero diffs. It independently re-ran the round-8 operator mutation and confirmed the
test fails without the fix.

So the corpus blind spot is now covered from both sides: Gemini found no missing shape, and GPT
found the one real gap, which is a capability limit rather than a wrong answer.

### Validation at `3a066e6c0`

- Slicer **362/0** (up one, the new characterization test)
- Services **357/0**
- CLI **2793**, 1 failure: `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, the known flake,
  identical at clean `main`
- `dotnet build dotnet-inspect.slnx -c Release` clean
- Mutation: forcing the conditional branch not to set `Untracked` fails the new test

### Review status

Nine rounds, 29 findings, 28 real, 27 fixed and 1 deferred with an issue and a pinning test.

`3a066e6c0` is doc comments plus one characterization test, with no product behavior change, so it
does not reopen the tier — the reviewed product code is unchanged from `cef356263`, which both
round-9 reviewers read.

richlander:
Ready to merge

Head `3a066e6c0`, MERGEABLE / CLEAN, CI green including `ci-required`.

### Review

Nine adversarial rounds across four model families (Claude Opus 4.8, Gemini 3.1 Pro, GPT-5.6 Terra),
two models per round at a fixed head. 29 findings, 28 real — 27 fixed, 1 deferred.

Round 9: Gemini APPROVE, GPT BLOCK on one finding.

### The one open finding, and why it does not block

A conditional directive marks every later declaration `SpanKnown=false`, to end of file, rather than
only within the region it guards — 12.1% of declarations on dotnet/runtime's libraries. Full
analysis in the [round 9 reconciliation](https://github.com/richlander/dotnet-inspect/pull/3652#issuecomment-5152218808).

Deferred deliberately, on three grounds:

1. **It is conservative, not wrong.** The index reports "do not know"; it never reports a bad span.
   Failure stays visible, which is the property `SpanKnown` exists to provide.
2. **Nothing consumes the index yet.** This PR adds the substrate; no caller reads it, so the gap
   costs nothing until the wiring PR, where it is the first thing that will be fixed.
3. **The fix is a real change with its own evidence needs** — per-branch brace-balance tracking, an
   added distinction on the internal `ScanToken`, and a new subset-comparison gate, since the
   existing corpus differential declines conditional files by construction. That belongs in its own
   reviewed PR, not appended to a nine-round one.

Tracked as #3668 with the proposed design and gate. The current behavior is pinned by
`AConditionalDirective_LosesEveryLaterRowToEndOfFile`, marked as a characterization test and
verified non-vacuous by mutation, so the fix will be a visible change rather than a silent one.

The contract was also corrected: `SpanKnown` said "affected rows", which reads as the guarded
region, and now states the real reach and names the limitation.

### What this PR is

`DeclarationIndex` — a whole-file lexical index of C# declarations: kind, name, trivia start,
signature range, body range, end, nesting depth, parent, attribute lists, and whether the span is
known. Three product files, no consumer changes, `BodySlicer` untouched.

It is the shared substrate for the two selectors the design settled on — PDB line containment
(`FindByBodyLine`) and canonical signature match (`FindByName`) — replacing the per-request rescan.

### Validation

- Slicer 362/0, Services 357/0, CLI 2793 with the one known pre-existing flake
  (`Layout_Count_CountsFilesRatherThanRenderedTreeLines`, identical at clean `main`)
- Whole-file correctness gated by a Roslyn differential over the real source of every PDB-bearing
  assembly beside the test binary — ~3,400 declarations, with separate gates for leading trivia,
  attribute lists, and parent/sibling nesting, each with its own non-vacuity floor
- Roslyn is a test-only oracle; the library stays Roslyn-free, SRM-only, NativeAOT-friendly

### Follow-up

- #3668 — conditional-directive depth recovery
- Wiring `ExtractMethodBody` onto the index, keeping the 61 characterization tests green
- The `MemberSource` parts API (docs, signature, body, attributes, with a printer), which needs a
  docs span on `DeclarationSpan` — the docs block is not a slice of trivia

richlander:
## Review at `3a066e6c0` — GPT-5.6 Sol — BLOCK, 2 findings, both real, both fixed

I asserted on the previous comment that `3a066e6c0` "does not reopen the tier", because its product
diff excluding `///` lines is empty. That argument covered `src/` and quietly ignored the 52 added
test lines. Reviewed at the exact head, scoped to that one commit. Both findings landed on the part
my argument skipped.

### 1. The characterization test did not gate the bodiless emit path (Medium)

The fixture used only methods with bodies. A bodiless declaration — a field, an interface method —
takes a **separate** `SpanKnown` assignment in `EmitBodiless`, and nothing exercised it.

Reproduced, and it is worse than reported: hard-coding that assignment to `true` passed **all 362
tests**, not merely the new one.

```text
mutated EmitBodiless SpanKnown -> true
DotnetInspector.CSharpBodySlicer.Tests  Total: 362, Errors: 0, Failed: 0
```

So `SpanKnown` for every field, interface method, abstract method, delegate and enum member had no
gate anywhere in the suite.

Why that path and not the other: a body-bearing row is corrected again when its closing brace is
seen (`if (!tok.DepthKnown) rows[idx].SpanKnown = false`), and since `DepthKnown` is monotonic that
correction subsumes the initial assignment. A bodiless row has no closing brace, so nothing covers
for it.

Fixed by naming `Field` and `Bodiless` in the fixture and asserting each row **by name** —
`Assert.Single` then `Assert.False` — rather than through an `Assert.All` that a vanished row would
satisfy.

Mutation, both paths:

| Mutation | Before | After |
| --- | --- | --- |
| `EmitBodiless` `SpanKnown = true` | 362 pass | **fails** |
| body-bearing `SpanKnown = true` | 362 pass | 362 pass — **equivalent**, see below |
| directive branch stops setting `Untracked` | fails | fails |

The body-bearing mutation is equivalent, not a gap: `DepthKnown` never recovers, so any row that
closes is corrected at its closing brace regardless. Recorded rather than gated with a contrived
test. Worth noting for #3668: **restoring depth after a balanced conditional makes that assignment
load-bearing again**, so that PR must add a gate for it. Added to the issue.

### 2. A false figure in the doc comment I had just corrected (Medium)

The comment said 36,411 declarations "lie entirely after the last `#endif`". Re-measured at runtime
`e614b717a9d`:

```text
unknown                                       70472 (12.123%)
unknown, SIGNATURE begins after last #endif   36411
unknown, TRIVIA    begins after last #endif   36135
  of which cross the #endif via trivia          276
```

36,411 counts declarations whose **signature** begins after the `#endif`. 276 of those start earlier
through leading trivia — an attribute or doc comment above the member — so they do not lie entirely
after anything. Entirely after, trivia included, is **36,135**.

The percentages were right; the phrase was wrong, and wrong specifically against this type's own
`TriviaStartLine` model. **This is the sixth false doc comment in this PR, and the second I wrote
myself — in the commit whose entire purpose was to correct a misleading doc comment.**

Corrected to 36,135 with "leading trivia included", the corpus revision pinned, and the figures
marked explicitly as an **ungated point-in-time measurement**: nothing re-measures them and they
will drift as that corpus moves. `AGENTS.md` requires an asserted property to name its gate or be
marked unverified; a measured number in a doc comment can only be the latter.

### Validation at `4f4e755fe`

- Slicer **362/0**; `dotnet build dotnet-inspect.slnx -c Release` clean
- Mutation table above
- Scoped deliberately: the product change is one doc comment, and the only changed test file is the
  slicer's, so the CLI and Services suites are not implicated

### On the process point

The reviewer was right to be dispatched. A "doc comment and a test" commit from the author of the
five prior false doc comments is not self-evidently the trivial tier, and treating it as one is how
the sixth got in. `4f4e755fe` is the same shape of change, so the same caveat applies to it.

richlander:
## Round 10 re-review (GPT-5.6 Sol at `4f4e755fe`) — reconciliation

**BLOCK, and it found a defect that is worse than the one it was sent to confirm.** Fixed in `fff4013ca`.

### Medium — a conditional initializer reported one branch's end as *known*

Reproduced exactly as reported:

```csharp
class C
{
    int P { get; }
#if FEATURE
    = 1;
#else
    = 2;
#endif
}
```

```text
Property P  sig=3-3  body=3-5  known=True     <- before
Property P  sig=3-3  body=3-5  known=False    <- after
```

Line 5 is the answer for a `FEATURE` build and off by two lines for every other build. Every other row in that file already reported unknown; `P` was the one that escaped.

**Root cause, and why it is the only one.** There are six places that write a row's span. Five consult the depth flag. The sixth — `if (text == ";")`'s trailing-initializer branch — *extends* a span that was measured and marked known when the accessor block closed a moment earlier, and it did not. That makes it the only path in the builder that can report a **wrong** span rather than **lose** a row, which is exactly why it is the one that falsified the PR's standing claim that an unresolvable conditional is conservative rather than wrong.

Gated by `AConditionalInitializer_ReportsUnknownRatherThanOneBranchsEnd`. It asserts the unconditional shape still resolves with `EndLine == 4`, so the fix cannot degenerate into a blunt "always unknown". Mutated **both directions**: neutralising the new check fails only the new test; forcing it always-unknown fails the new test *and* both corpus differentials, so the positive arm has real coverage too.

### Low — unpinned figures in the test comment

Taken. The test comment's 8.3%/12.1% now carry the corpus revision `e614b717a9d` and the same "ungated point-in-time measurement" qualification the production comment carries.

### The reviewer's other four answers

- Bodiless `SpanKnown=true` mutation now fails on `Field` — the round-10 fix holds.
- `FindByBodyLine(6)` and `(3)` select `Always`, not coincidentally.
- The body-bearing initial assignment **is** an equivalent mutation: closure or EOF always subsumes it. My reasoning was attacked and survived.
- 363/0, slnx clean.

### A seventh false doc comment, and it was the safety claim itself

`SpanKnown`'s doc said "This is conservative rather than wrong" and named **no gate** for it, which `AGENTS.md` forbids — and the property was in fact false. It now names the mechanism (every span write consults the depth flag) and the gate at the site where it was once absent.

Head is now `fff4013ca`; the reviewer has been sent back to it.

richlander:
## Round 10 re-review at `fff4013ca` — reconciliation

**BLOCK on one Medium, and it is a doc-accuracy finding, not a correctness one.** Fixed in `a410513ee`.

### Medium — the rewritten safety mechanism was itself false

I wrote, in the sentence whose entire purpose was to repair the previous false claim:

> every place that writes a row's span consults the depth flag

The unclosed-row sweep does not:

```csharp
if (r.EndLine < 0 && !r.ClosesAtEndOfFile)
{
    r.EndLine = lines.Count;
    r.SpanKnown = false;      // unconditional; never asks the flag
}
```

Safe, but not what the sentence said. **That is the eighth false doc comment in this PR, written inside the fix for the seventh.**

Checking the reviewer's correction, I found the claim was loose in a second way it did not report: the file-scoped-namespace opener writes `ns.EndLine = -1` as a sentinel, which neither consults the flag nor marks unknown — it fixes no span at all, and the site that resolves it later does consult. So the universal quantifier was wrong over *writes* in two different ways.

Restated over the right set:

> a site that **fixes** a row's span either consults the depth flag or sets `SpanKnown` false outright, the unclosed-row sweep being the only one of the latter kind

Docs only; no product behaviour changed. 363/0.

### What the reviewer confirmed at `fff4013ca`

Everything I asked it to attack about the fix itself held:

- **Both gate mutations fail correctly** — the new test is non-vacuous in both directions.
- **`FindByBodyLine(3)` and the plain `EndLine == 4`** assert what they appear to.
- **The fix is narrow to the conditional case** — both corpus differentials pass, so real unconditional trailing initializers in dotnet/runtime remain `SpanKnown=true`. The fix costs nothing.
- **No other known-wrong span found.** The reviewer went looking for a row reporting `SpanKnown=true` with a span that is wrong rather than merely lost, and did not find one.
- Build clean, 363/0, mutations restored.

Head is now `a410513ee`.

richlander:
## Ready to merge

Head `a410513ee`. **GPT-5.6 Sol APPROVE** at that exact head. CI green including `ci-required`; PR MERGEABLE/CLEAN.

### Round 10, in full

Four heads, because each fix was itself defective until the last:

| At | Verdict | Result |
| --- | --- | --- |
| `3a066e6c0` | BLOCK, 2 Med | ungated bodiless emit; a figure that counted signatures, not declarations |
| `4f4e755fe` | BLOCK, 1 Med + 1 Low | **a conditional initializer reported one branch's end as a _known_ span** |
| `fff4013ca` | BLOCK, 1 Med | the sentence written to repair the safety claim was itself false |
| `a410513ee` | **APPROVE** | all six span-finalizing sites enumerated; restated claim confirmed |

### The one product defect

Six places fix a row's span. Five consulted the depth flag; the trailing-initializer branch of `if (text == ";")` did not — and it is the only one that *extends* a span already measured and marked known, making it the only path in the builder that could report a **wrong** span rather than **lose** a row. It reported line 5 for `int P { get; }` followed by a conditional initializer: right for a `FEATURE` build, off by two lines otherwise. Every other row in that file already reported unknown.

Gated by `AConditionalInitializer_ReportsUnknownRatherThanOneBranchsEnd`, mutated in both directions. The fix is narrow: both corpus differentials pass, so real unconditional trailing initializers across dotnet/runtime remain `SpanKnown=true`.

### The process result worth recording

**Eight false doc comments in this PR; four of them mine, and three were consecutive attempts at the same sentence.** Four of the eight were *quantified* claims — "every", "all", a count. Each survived a build and a green suite, because prose is ungated by construction. That is precisely the failure mode `AGENTS.md` addresses by requiring an asserted property to name its gate or be marked unverified, and the `SpanKnown` safety claim had done neither.

Two heads that "only changed docs and tests" each turned out to carry a real defect. I would not have caught either without sending them back.

### Deferred, non-blocking

The set-and-never-cleared conditional loss itself — a directive costs every later row to end of file, 12.1% of declarations on dotnet/runtime's libraries. Tracked by **#3668** (lexical recovery, +25,293 members) and **#3672** (PDB branch liveness, → 99.6%). Conservative today, and now genuinely so.

Over to you to merge.

PR #3654, opened 2026-08-01
Materialize the pool walk before it deletes what it is walking

richlander:
## What this changes

`ReconcilePool` — the pool reconciliation added in #3604 — deletes entries
during a lazy `EnumerateFileSystemInfos`. This materializes the enumeration
before the loop deletes anything.

## Why

Whether an entry removed before the enumeration reaches it is still returned is
**unspecified**: POSIX permits `readdir` to skip it. A skipped entry here is an
unrecorded file left in the pool, unreported in `RemovedFromPool`, over an exit
code of 0 — which is the exact state the reconciliation exists to prevent,
arrived at silently.

The enumeration this replaced (`Directory.GetFiles(..., AllDirectories)`)
returned a materialized array, so the property held before #3604 and was lost in
that rewrite rather than never having been there. It is now stated in the code
so the next rewrite does not lose it again.

## Evidence, and its limit

Measured on ext4, 5000 entries, three trials: **no entries skipped**. So this is
not a reproduced defect on this filesystem — it is unspecified behavior that
happens to work here.

That is also why the property is **marked unverified rather than gated**: a case
asserting it would pass on ext4 whichever way the code was written, and would be
evidence of nothing. The comment says so rather than leaving the property to read
as covered.

## Validation

At `e81112b03`, based on `4b76b0973` (`main`):

- `EvilPoolSweepGateTests` — 19 passed, 0 failed
- `--gate fast` — 4443 passed, 0 failed
- Replayed #3604's symlink mutation (drop the `LinkTarget` arm) against the
  materialized walk — still reddens
  `ASweepRemovesALinkPlantedInThePoolWithoutDeletingWhatItPointsAt`, so
  materializing did not weaken the gate it was added with.

## Boundary

Behavior is otherwise unchanged: same traversal, same removal rules, same
manifest fields, same exit codes. No product code and nothing packed into
`dotnet-inspect` is touched — `eng/prepare-decompiler-package-sweep.cs` is a CI
step.

## Context

#3604 merged before its round-ten head had a clean review. Its reviewers
produced nine findings, the fixes for those landed at a head no reviewer saw, and
this is one issue found reading that head afterwards.


richlander:
## Second commit: the CI path filter was hiding this change

Worth recording how this surfaced, because it is the more consequential half.

After pushing `e81112b03`, `gh pr checks` reported `ci-required` **pass** — with
`test` and `decompiler-gates` both `skipping`. A change to
`eng/prepare-decompiler-package-sweep.cs` ran neither the lane that executes
`EvilPoolSweepGateTests` nor the decompiler gates.

The filter enumerates `eng/prepare-decompiler-*.sh` by name and never listed the
sweep. So the 19 cases written across ten review rounds specifically to hold this
script to its rules **did not run on a change to that script**. Everything
validating this PR up to that point was local.

That is precisely the failure the workflow's own comment warns about, a few lines
above the arm I edited:

> Under-triggering silently disables the gate on exactly the changes it exists to
> catch; over-triggering costs a parallel job that never blocks the hot lane.
> Prefer the cheap error.

`841397c90` adds the sweep to both the `CODE` and `DECOMPILER` arms. It also adds
`docs/data/nuget-top-packages.lock.json` and its list: the pin the sweep reads
lives under `docs/`, which no arm in the `CODE` list matches, so editing the pin
did not run `EvilPoolPinTests` either.

**Proof it took effect** — same PR, one commit later:

| Job | At `e81112b03` | At `841397c90` |
| --- | --- | --- |
| `test (ubuntu-26.04, linux-x64)` | `skipping` | runs |
| `decompiler-gates` | `skipping` | runs |

## Scope note

This is a second, separable claim in one PR. They are together because the second
was found by checking the first one's evidence, and because a fix to the sweep
that cannot be CI-validated is not worth landing on its own. If you would rather
see the filter fix land separately, say so and I will split it.

richlander:
## Review round — GPT-5.6

Six findings at `e81112b03`, in an isolated worktree. **One is a defect in the
code that merged with #3604**; the other five are properties this code's own
comments claim and nothing enforced. All six are fixed in `ed8e38b66`.

### The defect: the pool root itself could be a link

`ReconcilePool` refuses to descend through a link it finds *inside* the pool —
that is what #3604 added, and it is what bounds the deletion to the pool. But the
argument is only as good as the boundary it starts from, and
`Directory.CreateDirectory` is satisfied by an **existing symlink**.

With `<outdir>/packages` a link, the reviewer's probe against unmodified source:

```text
before: bystander exists = True
exit: 0
after: bystander exists = False
packages is link = True
```

The sweep wrote its assemblies outside the output directory, deleted a file
there, and reported success. Refused now, before anything is written. Refused
rather than replaced: removing the link would be this step deleting something the
caller put on a path the caller named, which is a larger liberty than declining to
run.

### Five ungated claims

| Claim in the code | Mutation that stayed green | Now gated by |
| --- | --- | --- |
| Scoped to `packages/`, leaves a caller's sibling files alone | Widen the walk to the output directory — deleted a planted `acquisition.log` | `ASweepLeavesWhatTheCallerPutBesideThePoolAlone` |
| Removing a leftover frees the name, not the bytes behind it | Truncate each leftover before unlinking — emptied a file through a shared inode | `ASweepRemovingALeftoverFreesTheNameAndNotTheBytesBehindIt` |
| A leftover differing only in case is a leftover | `OrdinalIgnoreCase` — plant survived in the pool, exit 0 | `ASweepRemovesALeftoverThatDiffersFromARecordedNameOnlyInCase` |
| `RemovedFromPool` survives a later failure | `removedFromPool.Clear()` in the catch | extended `ASweepThatCannotReconcileThePoolSaysSoAndFails` |
| `RemovedFromPool` names only what was removed | Record before deleting — manifest named a file still on disk | same case |

### A correction

`e81112b03` marked the ordinal comparison unverified, saying a case asserting it
"would pass on this filesystem whichever way the code was written". **That was
wrong.** Where the filesystem holds both spellings, the two files coexist and only
one is recorded, so the assertion distinguishes exactly as it should. The
reviewer's probe showed the plant surviving under `OrdinalIgnoreCase` and being
removed under `Ordinal`. It is gated now, and skipped only where the two names
cannot coexist.

This is the second time in this line of work that a property I marked "unverified"
was in fact gateable. The bar has to be a demonstration that the assertion cannot
distinguish, not an impression that it might not.

### One product change beyond the fixes

The walk now **sorts** its materialized entries. The failure case needs the walk
to reach a removable leftover before an undeletable one for its assertion to mean
the same thing on every run, and that ordering should be a fact about the pool
rather than about `readdir`. What a destructive step managed before it failed is
part of what it reports.

### Every mutation replayed against the fix

| Mutation | Before | After |
| --- | --- | --- |
| Linked `packages/` root (no mutation — current code) | deleted outside the pool, exit 0 | reddens 1 |
| Widen scope to the output directory | 19 green | reddens 1 |
| Truncate before unlinking | 19 green | reddens 1 |
| Discard removals on failure | 19 green | reddens 1 |
| Record before deleting | 19 green | reddens 1 |
| Case-insensitive recorded set | 19 green | reddens 1 |

The reviewer also confirmed eight attacks were already correctly caught,
including skipping the removal record, disabling recursion, stopping after the
first removal, swallowing the reconcile exception, exiting 0 with `Unreconciled`
set, exiting 1 with it null, and resolving and deleting a link's target.

### Validation

At `ed8e38b66`:

- `EvilPoolSweepGateTests` — **23 passed**, 0 failed, 0 skipped
- `--gate fast` — 4447 passed, 0 failed
- `EvilPoolPinTests` — 10 passed, 0 failed
- Shared NuGet and dotnet-inspect caches untouched

### Note on the baseline

The reviewer's baseline showed `TheSharedNuGetCacheHoldsWhatTheIsolationCaseNeeds`
skipping in its worktree, where mine did not skip on any run. That case is
self-describing about why it skips, so nothing is hidden — but it means the
class's effective coverage is environment-dependent in a way worth knowing. Not
addressed here.

richlander:
## Round twelve reconciliation (Gemini 3.1 Pro at `ed8e38b66`) — fixed at `615d26d8e`, head now `9f4c14b6c`

Four findings, all real, all fixed. Two of them are defects in *my own work from the previous round*, which is the part worth reading.

### F1 (High) — the gate I added last round was vacuous. Third time in this PR.

`ASweepRefusesAPoolDirectoryThatIsALinkOutOfTheOutputDirectory` asserted `!File.Exists(elsewhere/Sweep.Lead.dll)`. The sweep pools under `<rank>-<id>/<version>/`, so it writes `elsewhere/001-sweep.lead/1.0.0/Sweep.Lead.dll` and never the path the test named — under *any* circumstance, escape or not. Confirmed by removing the refusal entirely and watching the case stay green.

It now asserts the whole `elsewhere` directory still holds exactly its one planted bystander. Non-vacuity: moving the pool-root `LinkTarget` refusal to after the copies reddens it; that same mutation was green before.

This is the third time a gate I wrote in this PR asserted against a path or premise I composed rather than against reality:

1. the reconcile-failure case passed vacuously (an unwritable destination fails the *copy*, which also exits 1);
2. the ordinal comparison was left ungated on a justification that was simply false (it is perfectly gateable on a case-sensitive filesystem);
3. this one.

The pattern that keeps working is the opposite: hold a whole directory to its expected contents, or have the product *report* the fact (`RemovedFromPool`) and assert on the report. Both are things the test cannot accidentally agree with itself about.

### F2 (High) and F3 (Medium) — last round's CI filter fix was itself incomplete, in two ways.

`841397c90` was the fix for a filter that let the sweep change while `test` and `decompiler-gates` skipped and `ci-required` reported green. It was under-done:

- The pin files (`docs/data/nuget-top-packages{,.lock}.json`) reached the **code** arm but not the **decompiler** one. The pin decides which packages land in the pool, so it selects the IL the decompiler gate is run over — it could change the corpus while skipping the gate that reads it.
- `eng/prepare-evil-corpus.sh`, the script that actually *runs* the sweep, matched **no arm anywhere**. Editing it ran no lane at all.

Both added to both arms. Evidence: extracting the workflow's real `while read … case … done` loop and running paths through it verbatim.

| path | before | after |
| --- | --- | --- |
| `eng/prepare-evil-corpus.sh` | code=false decompiler=false | code=true decompiler=true |
| `docs/data/nuget-top-packages.lock.json` | code=true decompiler=**false** | code=true decompiler=true |
| `docs/data/nuget-top-packages.json` | code=true decompiler=**false** | code=true decompiler=true |

The probe is non-stuck (`README.md` gives code=false) and non-vacuous (deleting either new arm flips both outputs back to false).

**My first probe of this was itself a false negative** — it reported false for the sweep, a path I know CI matches, because my `sed` extraction grabbed only the first of six `case` blocks. Worth stating: I nearly reported a filter regression that did not exist. The fix was to stop paraphrasing the workflow and execute its own text.

### F4 (Low) — `Assert.NotNull` on `Unreconciled` passes for `""`.

A manifest reporting an empty reason satisfies a null check and tells a consumer nothing. Now `Assert.NotEmpty`. Non-vacuity: setting `unreconciled = ""` reddens the case; it was green before.

### The broader rot, named but not swept up

`eng/prepare-evil-corpus.sh` is not alone. Six further `eng/` files match no arm: `BannedSymbols.txt`, `measure-metadata-projection-allocation.cs`, `prepare-aspire-dashboard-corpus.sh`, `prepare-performance-triage-corpus.sh`, `prepare-resource-triage-corpus.cs`, `regen-inverse-ledger.sh`, `restore-authored-source-corpus.sh`. The filter enumerates by name, so this class recurs by construction, and the workflow's own comment says it prefers the cheap error (over-trigger) to the silent one.

I have **not** broadened the filter to `eng/*` here. That is a CI-cost decision across lanes this PR has no claim over, and this PR's claim is the sweep. Flagging it rather than deciding it.

### Validation at `9f4c14b6c`

`origin/main` merged (17 commits, no conflicts) before validating, so this is the head that would land.

- Full solution build: 0 errors
- `EvilPoolSweepGateTests`: **23/0**
- `EvilPoolPinTests`: **10/0**
- `--gate fast`: **4498/0** (was 4447; main added 51)
- Shared cache intact: `newtonsoft.json/13.0.4` net6.0 = `a28c251dfe36`
- `ci.yml` parses; tree clean

### Round yield, unchanged in character

`4, 0, 4, 5, 9, 8, 7, 4, 6, 9, 6, 4` — not decaying. Rounds eleven and twelve both found defects **in the previous round's fix**, which is the signal that matters more than the count.

richlander:
## Round thirteen reconciliation (GPT-5.6-Terra + MAI-Code at `9f4c14b6c`) — fixed at `ed22d09d1`

Two models, 25 + 7 executed probes. One real product defect, found after twelve rounds had missed it. Two further pre-existing defects verified and deferred to follow-ups.

### The defect, and why twelve rounds missed it

**GPT-5.6-Terra, Medium → fixed.** `ReconcilePool` descended into a directory and never removed one. So a `<rank>-<id>/<version>/` skeleton survived every sweep — wherever a package was refused, failed to copy, or was pooled by an earlier run and not this one — and the run exited **0**. To anything enumerating the pool by directory, that skeleton is a package that shipped nothing, which is exactly the state a recorded-only pool exists to make unrepresentable. The reconciliation was reconciling the pool's bytes and not its shape.

The reason it survived twelve rounds is worth more than the bug. `AssertPoolMatchesRecord` — the assertion that holds the pool to its record, the one nearly every case in this file leans on — read the pool with `Directory.GetFiles` alone:

```csharp
string[] present =
    [.. Directory.GetFiles(OutputDirectory, "*", SearchOption.AllDirectories) …];
```

**It was structurally unable to see a pool entry that holds nothing.** No mutation of the walk's directory handling could ever have been caught by it, no matter how many rounds ran. Twelve rounds of mutation testing against a blind assertion produced twelve clean results on this property.

That is a different failure from the three vacuous *cases* found earlier in this PR. Those were individual tests agreeing with themselves. This was the shared assertion having a blind spot, so the vacuity was distributed across every case that called it.

Fixed on both sides:

- **Product**: an unrecorded directory that is empty after its contents are reconciled is removed. Checked after the recursion, so a directory holding only unrecorded files leaves with them; asked of the child by its parent, so the pool root — which the sweep owns — is never a candidate. `Delete()` without recursion, so a directory that turned out not to be empty raises a reconcile failure rather than silently recursively deleting whatever raced in.
- **Harness**: the assertion now holds the pool's directories too, **derived from the sweep's own record** — the directories a pool may contain are exactly the ancestors of the files it says it pooled — rather than from a list composed in the test. Composing that list is precisely how three earlier cases here came to gate nothing.

An emptied directory is removed but **deliberately not added to `RemovedFromPool`**. That record exists so that destroyed *content* stays visible; an empty directory has no content, and the sweep creates and abandons these itself on every refusal, so reporting them would put an entry in `RemovedFromPool` on routine runs and dilute the signal it is kept for. The shape is held by the pool equalling its record instead.

**Non-vacuity**: removing the emptiness check reddens **five** cases — `ASweepCountsAPackageWhoseCopyFailedAsFailed`, `ASweepRefusesBytesThePinDoesNotName`, `ASweepRefusesBytesThePinDoesNotNameAtTheFirstRankToo`, `ASweepIntoAReusedDirectoryPoolsOnlyWhatItRecorded`, `ASweepLeavesNoTemporaryWhenAWriteFailsAfterCreatingOne`.

Also: `RestoreWritable` now tolerates a directory that is gone. Teardown made a directory unwritable to fail a copy, and such a directory now ends the run empty and removed — removing it needs write permission on its parent, not on itself.

### What came back clean

GPT ran **25** mutations; 22 GATED, including all five cases added in rounds eleven and twelve. Its verdict on those: "The five recently scrutinized cases are all non-vacuous." One SKIPPED (its own mutation did not remove the behavior — reported honestly rather than counted). One Low: `TheSharedNuGetCacheHoldsWhatTheIsolationCaseNeeds` is a precondition test, not a product gate — agreed, and the real isolation gate is `ASweepCannotReachAPackageTheSeededCacheDoesNotHold`.

MAI-Code independently re-derived the CI filter probe — including the trap I fell into, checking that its probe could distinguish `README.md` from `src/dotnet-inspect/Program.cs` before trusting it — and confirmed all four sweep-related paths now set both `code` and `decompiler`, with no further unmatched code- or decompiler-relevant path under `src/`, `tools/`, `tests/`, or `docs/data/`. The fallback branch forces every lane on when change detection fails.

### Verified, and deliberately NOT fixed here

Both are pre-existing, neither is a regression from this PR, and both are filed as follow-ups rather than bolted onto a PR that already carries two claims and thirteen rounds.

**High — a failed run leaves a stale manifest** (#3703). I confirmed the control flow myself rather than taking the report: `assemblies.txt` is written at line 751 and exits **2** on failure, while `manifest.json` is written at line 926. Between them sit the copies and the reconciliation. So a run that fails to write `assemblies.txt` has already changed the pool, and leaves the previous `manifest.json` in place describing a pool that no longer holds.

Deferred because the fix is not obvious. The sweep reuses its output directory *by design* — "the corpus script passes the same output directory every time so that the paths in an earlier assemblies.txt stay valid" — so simply clearing outputs up front fights a documented intent, and there is more than one exit-2 path after the pool has changed. That deserves its own design and its own round.

**Medium — an empty diff defeats the CI fallback** (#3704). For a `push` this is correct: nothing changed, so nothing needs validating. For a `pull_request` an empty file list is impossible by construction, so it should take the fail-safe branch rather than turning every lane off.

### Validation at `ed22d09d1`

- `EvilPoolSweepGateTests`: **23/0**
- `EvilPoolPinTests`: **10/0**
- `--gate fast`: **4498/0**
- Shared cache intact: `newtonsoft.json/13.0.4` net6.0 = `a28c251dfe36`

### Round yield

`4, 0, 4, 5, 9, 8, 7, 4, 6, 9, 6, 4, 3` — and the thirteenth round found a product defect that a blind assertion had been hiding from all twelve before it. Not decaying.

richlander:
## Round fourteen reconciliation (Gemini 3.1 Pro + GPT-5.6-Sol at `41220144f`) — fixed at `ef35d63f3` and `eb717e6e9`

Two models, 8 + 20 executed probes. The worst finding is against a change *this PR made*, and it is the same mistake the PR exists to fix.

### The DECOMPILER lane mapping I added was cargo cult

**GPT-5.6-Sol, Medium → fixed in `ef35d63f3`.** Earlier in this PR I mapped the sweep, the wrapper, and both pin files to `DECOMPILER=true`, justified in a comment reading "The sweep builds the EVIL pool this gate is run over".

Measured, and false:

```console
$ dotnet run --project src/ILInspector.Decompiler.Tests -c Release -- --gate pre-merge -list classes
ILInspector.Decompiler.Tests.ByteNeutralityGateTests
ILInspector.Decompiler.Tests.FidelityGateTests
ILInspector.Decompiler.Tests.GateExpectedClassesTests
ILInspector.Decompiler.Tests.LoweredFidelityGateTests
```

None of those read the EVIL pool, and the `decompiler-gates` job never runs `eng/prepare-evil-corpus.sh`. So editing the sweep or the pin scheduled a **fifteen-minute job that validated nothing about the file that triggered it**.

The coverage those paths actually have is the `test` job. `EvilPoolSweepGateTests` and `EvilPoolPinTests` carry no `Speed=Slow` trait, so that job's `-trait- "Speed=Slow"` selector runs all 23 and all 10 — verified by running that exact selector. `CODE=true` alone is correct and sufficient. The four `DECOMPILER` arms are removed, with the reason and the paths recorded in place should the lane ever consume the pool.

**This is the same failure this PR is about, committed by the fix for it**: an assertion about what a check covers, believed rather than run. Round eleven found a defect in round ten's fix; round twelve in round eleven's; this is round fourteen finding one in round twelve's. That pattern is the strongest argument in this PR's history for not stopping early.

Also fixed per the repository's gate-naming rule, which I had violated in comments written this PR: three comments asserting a safety property now name the gate enforcing it, and the readdir claim that rested on an unreproducible "measured over 5000 entries" is now marked plainly unverified.

### A dotfile in the pool — Gemini's finding, sharpened

**Gemini 3.1 Pro, Low → fixed in `eb717e6e9`.** Gemini reported that filtering dotfiles out of the *emptiness check* went unnoticed. Re-deriving the mutation against the *walk itself* rather than the check made it considerably worse: with dotfiles filtered from the enumeration `ReconcilePool` walks, all 23 cases stayed green while an unrecorded file remained in the pool, unreported, at exit **0** — precisely the state this step exists to prevent.

Nothing in this file had ever put a dotfile in the pool. That is not a synthetic gap: NuGet writes `.nupkg.metadata` and `.signature.p7s` into package directories, so a dotfile leftover is what a partial or abandoned run actually leaves. `ASweepIntoAReusedDirectoryPoolsOnlyWhatItRecorded` now plants a fourth leftover, `.nupkg.metadata`. The filtering mutation reddens it, and left everything green before.

Re-deriving rather than taking the reported mutation is now the third time that has changed a finding's severity in this PR. Reviewer-stated diffs have repeatedly not been the ones producing the reported result.

### What came back clean

Gemini's primary assignment was whether the **new** directory assertion is blind the way its `Directory.GetFiles` predecessor was. Its verdict, with the axes checked: symlinked directories are compared without being traversed; case differences fail under `StringComparer.Ordinal`; and the ancestor-derivation loop terminates exactly at `packages/`, never including the output directory. Its mutations confirm the strength — making the emptiness check always false reddens 5 cases, always true reddens 19, deleting before recursing reddens 5, and removing the pool-root guard reddens 4.

It also constructed two adversarial shapes I had not: `packages/` as a regular file (refused, exit 2) and `packages/` unwritable (exit 1, no crash). Both behaved correctly.

GPT confirmed no consumer depends on the empty directories now removed — `DecompilerHarness` reports are byte-identical before and after — and that `manifest.json` is archived rather than parsed by the harness.

### Filed rather than fixed here

**#3706 (High), separately verified.** `ci.yml` triggers `pull_request` only on `[main, 'feature/**', 'fix/**']`, directly below a comment stating "The list must therefore cover every prefix a *base* can have". PR **#3558**, based on `prototype/wasm-command-ui`, is `CLEAN` and mergeable with **zero checks**. This is a recurrence — the comment records `fix/**` being added after #3684 hit it. Not fixed here because it governs every lane and every PR, not this change.

Two further GPT findings are filed within that issue's family and left for follow-up: the wrapper `eng/prepare-evil-corpus.sh` has no gate that executes it, and the path classification itself has no self-test (a one-character typo in the CODE arm went unnoticed). Both argue for moving classification into an executable script with a test matrix, which is its own change.

Also noted, not fixed: a missing required pin exits **1** where the same missing pin under `--validate-pin` exits **2**. Pre-existing exit-code incoherence, unrelated to this change.

### Validation at `eb717e6e9`

- `EvilPoolSweepGateTests`: **23/0**
- `EvilPoolPinTests`: **10/0**
- `--gate fast`: **4500/0**
- `ci.yml` parses; the filter loop run verbatim reports `code` only for the four sweep paths, both lanes for `src/dotnet-inspect/Program.cs`, and neither for `README.md`
- Shared cache intact: `newtonsoft.json/13.0.4` net6.0 = `a28c251dfe36`

### Round yield

`4, 0, 4, 5, 9, 8, 7, 4, 6, 9, 6, 4, 3, 4` — fourteen rounds, still finding things, and still finding them in the previous round's fix.

richlander:
## Round fifteen — MAI-Code and GPT-5.6-Luna, both at `eb717e6e9`

Two reviewers, aimed at the two surfaces the previous fourteen rounds never attacked: the `SweepWorld` harness itself and `EvilPoolPinTests` (MAI-Code), and the acquisition / TFM-selection / hashing / copy steps that rounds thirteen and fourteen skipped (GPT-5.6-Luna, 25 mutations).

**Nothing in this round changes the branch.** Head remains `eb717e6e9`. Every real finding is against sweep steps 1-4, which this PR does not touch, so all three are filed rather than folded in: **#3709**, **#3710**, **#3711**.

### MAI-Code — two findings, neither survived re-derivation

**Claimed High: weakening `AssertOnlyTheLeadWasPooled` leaves 23/23 green.** Methodologically invalid as reported — a *test assertion* was weakened and run against a *correct product*, where passing is the expected outcome and proves nothing. Re-derived properly, with the helper stripped to its `SelectedPackageCount` check **and** the sweep's hash-mismatch refusal (line 660) disabled: `ASweepRefusesBytesThePinDoesNotName` and `...AtTheFirstRankToo` both go red. The helper's disk assertions are redundant defense-in-depth, not an ungated property.

**Claimed Medium: `EveryPinNamesAPackageAndAnExactVersion` is ungated for a missing `status`.** The mechanism is real — `ReadPins()` (`EvilPoolPinTests.cs:737`) manufactures `"pinned"` when the property is absent — but it mirrors the product's own record default (`prepare-decompiler-package-sweep.cs:1500`), and that default is the *safe* direction: `pinned` requires a sha256 by the validator's `sha` rule, while `no-library` would skip that requirement. A *typo'd* status is still caught, because `status.GetString() ?? ""` preserves it; only *absence* is laundered, and absence is legal and safe by design. The suggested fix — make `ReadPins` require the property — would make the harness stricter than the product, the divergence `AGENTS.md` forbids. Not changed. The residual observation, that the harness duplicates the default rather than exercising it, is real, small, and pre-existing.

### GPT-5.6-Luna — 25 mutations, 17 gated, 8 ungated, of which five are real

Two of the eight were **already marked unverified in the product, accurately**, with the same measurement the reviewer took:

- **F1, hashing the source rather than the destination.** The comment at lines 645-656 already says: *"every case in EvilPoolSweepGateTests tampers with the cache entry, so source and destination carry the same bytes and no black-box case can tell which was measured. Hashing the source here leaves the whole class green. What is gated is that the hash is compared at all."*
- **F3, the acquired-identity check.** Marked at lines 496-506, naming the cache-echo construction and the measurement.

An independent reviewer's mutations landing exactly on the marked spots, and the markings being accurate enough to dispose of both in one read, is the `AGENTS.md` "name your gate or mark it unverified" rule paying for itself.

One was unobservable by construction:

- **F8, removing `assemblies.Sort`.** The reviewer notes processing is already rank ordered; destinations are named `{rank:D3}-{package}`, so an ordinal sort of them *is* rank order. The sort is defensive and redundant given the naming scheme, and no fixture can separate the two. Observation, no change.

The remaining five are real, and were re-derived independently before filing:

- **#3709 (the substantial one) — continuation after a rank-1 failure is gated at one of four sites.** `continue` at 490 (acquisition failed), 610 (version/TFM mismatch) and 640 (copy failed) all become `break` with **Total: 23, Failed: 0**. Only the sha site (672) is gated, by `ASweepRefusesBytesThePinDoesNotNameAtTheFirstRankToo`. A sweep that abandons the remaining ninety packages the moment rank 1 fails is indistinguishable from a correct one to this suite — and the exit code would be explained by the rank-1 failure while the missing ninety would not be.
- **#3710 — the sha256 comparison's full length is ungated.** `assemblySha[..8]` leaves 23/23 green; a zero-length prefix is caught. The reviewer proposed a product seam on the reading that gating this needs a prefix collision; it does not. The validator's `sha` rule checks `IsSha256` — shape only — so a pin sharing the fixture hash's first 8 characters and arbitrary hex after is legal input that a full comparison refuses and any truncated one accepts. No collision search, no seam.
- **#3711 — the returned-version check is unmarked.** Deleting `package.Version` vs `pin.Version` leaves 23/23 green, for precisely the construction its neighbour's marking already names: `PackageExtractor.cs:290` echoes `packageName` **and** `version` back from the request on the cache-hit path. The identity check says so; the version check, five lines of reasoning about pin fidelity, does not, and so reads as covered. Comment-only fix. The reviewer's suggested fixture is not constructible — requesting a version the cache lacks yields `acquisition-failed`, not a successful acquisition carrying the wrong version.

### Where this leaves the round count

Yield by round: `4, 0, 4, 5, 9, 8, 7, 4, 6, 9, 6, 4, 3, 4, 5`.

After round fourteen produced no High/Medium *product* defects, the working hypothesis was that the well was running dry. Round fifteen refutes it: pointing two reviewers at surfaces the previous fourteen rounds had never touched produced a coherent and valuable class of gap — a property the suite gates at one of the four sites where it applies, with the one gated site being the only one anybody had thought to write a case for.

The pattern from earlier rounds held again in both directions. Re-derivation demoted MAI's High to a methodological error, and it upgraded GPT's F2 from "needs a product seam" to a cheap pin-side fixture. Neither reviewer's severity survived contact unchanged.

### Branch state

`eb717e6e9`, MERGEABLE/CLEAN, zero commits behind `main`, all checks passing, working tree clean and verified at 23/0 after every mutation was restored.

richlander:
## Round sixteen — Gemini 3.1 Pro and GPT-5.6-Sol, both at `eb717e6e9`

Aimed at the two surfaces fifteen prior rounds had never systematically attacked: the **input validation rule set** (Gemini, 19 mutations) and **everything after the per-package loop** (GPT-5.6-Sol, 36 mutations). Both prompts listed the eight already-documented ungated properties as do-not-report, so the round was spent on new ground rather than rediscovering marked ones.

**Nothing in this round changes the branch.** Head remains `eb717e6e9`; the tree was verified back to 23/0 and 10/0 after every mutation was restored. Every finding is filed: **#3715**, **#3716**, **#3717**, **#3718**, **#3719**, **#3720**.

### What is gated — the reassuring half, and it is substantial

- **All ten pin-file validator rules are individually gated.** `schema`, `packages`, `null-entry`, `blank-name`, `bare-id`, `version`, `status`, `sha`, `no-library-hash`, `duplicate` — muting any one produces exactly one failure, via `TheSweepRefusesEveryPinFileShapeThisSuiteRefuses`. `IsBareVersion` and `IsBarePackageId` are gated too.
- **No failure-to-success exit transition survived.** The accounting exit `1 → 0` produces 11 failures and the reconcile exit `1 → 0` produces 1. The highest-severity thing available in this subject is defended.
- **This PR's own manifest fields are gated.** Clearing `RemovedFromPool` produces 5 failures; nulling `Unreconciled` produces 1; `SelectedPackageCount` off by one produces 11; dropping a package row produces 14. The accounting check is gated in all three directions (never fires 11, always fires 10, off-by-one 20). Moving reconcile after the manifest write produces 5.
- `assemblies.txt` content is well gated: omitting a path 17 failures, relative instead of absolute paths 17, reversed order 1.

### What is not — six findings, all re-derived before filing

**#3715 (the one worth reading) — every pin rule is gated; not one ranked-list rule is.** The sweep validates two input files. The pin side is exemplary. The list side has four refusals — non-bare ids (240), duplicate ranks (249), a window short of the requested count (312), a contiguity gap (334) — and muting all four together leaves **23/0 and 10/0 green**. These are not stylistic checks: the comments above them record the corpus-integrity bugs that motivated them, including a non-bare spelling letting the same package be ranked twice and "pad the pool straight past the duplicate check". The asymmetry is the finding — same program, same cost to fixture, one file comprehensively defended and the other not at all.

**#3716 — the manifest's provenance and header are asserted by nothing.** Wrong `ResolvedPackage`, `ResolvedVersion`, `Tfm`, `SchemaVersion: 999`, wrong `Source`/`StartRank`/`RequestedPackageCount`, null failure `detail`, inverted failure-row `FromCache`: all green. I re-derived the first — a manifest naming `"wrong-package-name"` leaves 23/0. I also checked who consumes this file, since I had scoped both agents to one class and an "ungated" verdict could have been rescued elsewhere: `EvilPoolSweepGateTests` is the **only** reader of `PackageSweepManifest` in the repository. That caps today's blast radius and is simultaneously the argument for gating `SchemaVersion` before something does read it.

**#3717 — the exit 2 vs exit 1 contract is ungated at both metadata writes.** Both degrade from "refused bad input" to "ran, something failed" with 23/0 green; re-derived. Cheaply gateable by planting a directory at each path.

**#3718 — reconcile running only after the record is durable is ungated.** Moving reconcile *before* the `assemblies.txt` write is unnoticed. Reconcile is the sweep's only destructive step; ahead of the write, a failed write leaves the pool reconciled to a list that was never written, with the previous run's `assemblies.txt` describing a pool that no longer exists. Distinct from #3703, which is about the trailing write.

**#3719 — `IsSha256`'s hex check is gated by a specimen too short to reach it.** The `sha` rule is gated, but its specimen is `"NOTAHASH"` — 8 characters — so the length check short-circuits and the hex predicate is never evaluated. Weakening the helper to length alone leaves 10/0 green. This is the same vacuity shape this suite keeps producing: a specimen that trips the first of two conjoined conditions gates only the first while reading as though it gates both. Fix is one 64-character specimen. Filed alongside `SafePathSegment`, which is ungated for a legitimate reason — the shape helpers refuse every separator, so no accepted input can reach it — and needs the marking rather than a fixture.

**#3720 — `assemblies.txt` encoding and operator diagnostics have no asserted contract.** UTF-16LE with a BOM passes. Filed for a decision rather than as a defect; declining to gate the prose is an acceptable outcome, and the encoding half is the sharper one.

### Round count

Yield: `4, 0, 4, 5, 9, 8, 7, 4, 6, 9, 6, 4, 3, 4, 5, 6`.

Two rounds ago the hypothesis was that the well was running dry. Rounds fifteen and sixteen both refute it, and for the same reason: the yield tracks *which surface is attacked*, not how many rounds have run. Fourteen rounds of attacking the pooling, copying, hashing, and reconcile paths had reached genuine diminishing returns on those paths — and the first look at input validation found four undefended refusals, while the first look at the post-loop steps found the manifest saying whatever it likes about every package it pooled.

The severity-changes-on-re-derivation pattern held for a third round. Gemini's `SafePathSegment` finding is correctly *not* actionable as a fixture, and its own analysis says so; its `IsSha256` finding, which it rated Low, is the cheapest real fix in the batch.

### Branch state

`eb717e6e9`, MERGEABLE/CLEAN, zero behind `main`, all checks passing, working tree clean and verified at 23/0 and 10/0.

richlander:
Integrated current `origin/main` in merge commit `9b3ba2f7dc982898770375b166963775734126c7`.

The only merge conflict was the `CODE` path classifier in `.github/workflows/ci.yml`; it was resolved additively so this PR retains coverage for the sweep, its caller, and both pin files while preserving main’s new iltools activation/restore coverage. The sweep and `EvilPoolSweepGateTests` auto-merged cleanly, including #3736’s SHA-256 hex gate.

Validation at this head:
- Release solution build: passed
- `EvilPoolSweepGateTests` + `EvilPoolPinTests`: 33 total, 0 failed, 1 documented shared-cache skip
- decompiler `--gate fast`: 4,504 total, 0 failed, 1 same documented skip

Net diff against current `origin/main`: the same three PR-owned files, 378 additions and 9 deletions.

richlander:
Integrated current `origin/main` again in merge commit `bd104e923760eae144ce4d9ff66d2c5abf356a18`, after #3609, #3753, and #3749 landed.

#3753 touches `EvilPoolSweepGateTests.cs`, but its three rank-one continuation gates auto-merged cleanly with this PR’s pool-shape/removal gates. Both sets are present; the net PR diff remains the same three owned files.

Validation at this head:
- Release solution build: passed
- `EvilPoolSweepGateTests` + `EvilPoolPinTests`: 36 total, 0 failed, 1 documented shared-cache skip

No review round has been started; the round-17 approval request remains pending.

richlander:
Integrated current `origin/main` for the authorized round 17 in merge commit `f2c2655703b85e0c7df2443806f782d885cb6081`.

This range interacts with the PR:
- #3757 adds the record-before-reconcile durability gate from #3718;
- #3764 defines and gates BOM-less UTF-8/LF bytes for `assemblies.txt` from #3720.

Both auto-merged cleanly with this PR’s sweep and test changes. The combined source retains those two contracts plus #3654’s materialized/sorted walk, pool-root boundary, removal truthfulness, directory-shape reconciliation, case-sensitive identity, and dotfile witness.

Validation at this head:
- Release solution build: passed
- `EvilPoolSweepGateTests` + `EvilPoolPinTests`: 37 total, 0 failed, 1 documented shared-cache skip
- net diff remains the same three PR-owned files

Round 17 was explicitly approved and will begin only after current-head CI is green and the PR is mergeable.

richlander:
## Round 17 reconciliation — `f2c2655703b85e0c7df2443806f782d885cb6081`

- **GPT-5.6 Sol:** CLEAN. Release build passed; focused sweep/pin gate passed 37 total, 0 failed, 1 documented shared-cache skip. A transient `GenerateDepsFile` lock cleared on rerun; the review worktree remained clean.
- **Claude Opus 5:** CLEAN. Same build and 37-case result; the review worktree remained clean. Claude additionally mutation-tested the empty-directory removal gate and probed the destructive walk across deep empty trees, dotfiles, hidden directories, symlinks, dangling links, hard links, partial failure, and durable-record ordering. The assertions reddened when enforcement was removed and the interacting #3736/#3753/#3757/#3764 contracts remained intact.

No blocking findings were raised, so no resolution commit was needed. Both reviewers used isolated worktrees at the exact head and reviewed the exact three-file range from current integrated base `e5ca5d974d3ef5a8c4752229da8efcdf9618338d`.

Current-head CI run 30863759505 is green, GitHub reports MERGEABLE/CLEAN, and `origin/main` remains the reviewed base.

Ready to merge

PR #3656, opened 2026-08-01
Stack a caret per fact extent instead of widening the line

richlander:
Stacks a caret per fact extent instead of widening the line to the statement.
Slice of #3570. #3617 (caret extent narrowing), which this used to be stacked on, has
merged as `fbabdf6d0`; this now targets `main` directly. Design and rationale: #3657.

## The defect

`--focus` renders one caret per line. When the facts on a line pointed at
different characters, the caret widened to the whole statement: it pointed at
everything and therefore at nothing, above details with no way to tell which
detail belonged to which expression.

`System.Tuple<…>.Equals` is the extreme in CoreLib — 16 facts, one 370-column
caret, nothing attributed. The facts were right and the anchoring was right.
Only the render threw the information away.

## After

```text
    return comparer.Equals(a1, b1) && comparer.Equals(a2, b2) && comparer.Equals(a3, b3) && comparer.Equals(a4, b4);
//                       1.^^                       2.^^                       3.^^                       4.^^
//  1. alloc.box(T1; alloc=boxed T1; path=straight-line; path-confidence=dominates-return; …)
//  2. alloc.box(T2; alloc=boxed T2; path=branch; path-confidence=behind-branch; …)
//  3. alloc.box(T3; …)
//  4. alloc.box(T4; …)
```

That is the real render of `CommandCaretGestureFixture.DenseBoxes` through
`member … -S "Annotated Source" --focus allocation`, not a mockup.

Facts the anchoring layer could not place are listed under `-` rather than
widening the caret on their account. That is what lets a **mixed** line — some
facts with an extent, some without — keep the placement that was recovered:

```text
        Span<char> V_0 = stackalloc char[256];
    //                 1.^^^^^^^^^^^^^^^^^^^^
    //  1. lifetime.stack-bound(Span<char>)
    //  -  unsafe.stackalloc(byte*)
```

Both facts there are about `stackalloc char[256]`. Widening also underlined
`Span<char> V_0 = `, which neither fact claims. The `-` says the second fact has
no characters to point at — a different claim from "it is about the statement".

## The rules

1. Group facts **by extent**, not per fact — same characters, one caret, one number.
2. Order by start column, **widest first** at a tie. Extents sharing a start column
   are always a nesting and can never share a row, so the outer one goes above the
   inner one. That orders *same-start* extents only. It does **not** order row widths
   overall — a later disjoint extent can be packed onto a lower row and reach further
   right than anything above it, on 50 of the 2,842 lines — and it is the *reverse* of
   the order the printer records nesting in, which is descendants-before-ancestors.
3. Number in that order; the label sits immediately left of the trail.
4. Pack greedily onto rows.
5. Draw each trail at **true width**, clipped to `~` only where it would collide
   with the next label on its row.
6. List facts below, numbered to match, wrapped to `Budget`.
7. A fact with no extent is **listed, not drawn** — marked `-`. That holds only when
   the line stacks; a line rule 8 rejects widens and marks nothing.
8. **The gutter wins.** If any label would begin left of the comment gutter, the
   stacked layout is rejected for the whole line and it widens instead.

Width is information `ComputeCaretExtents` worked to recover, so rule 5 states it
truthfully or marks that it could not. Rule 4's two-column threshold is a *spill*
threshold, not a render floor: a one-character expression renders one caret, never
a padded two.

Design doc: `docs/design/caret-stacking.md` (separate PR).

## Why this layout and not a nicer one

The primary consumer is a terminal. A layout wider than the terminal wraps, and a
wrapped caret row lands under characters it has nothing to do with — worse than no
caret, because it actively misleads.

**Every stacked caret row lies within the code line it annotates**, and that is
*structural*, not a corpus result — an earlier revision of this section presented it as
a measurement, which is exactly backwards. `Stack` sends any extent reaching past the
line length to the unplaced list, labels grow leftward, and rule 8 bails rather than
shifting a trail right. No corpus could produce a counter-example.

The contrast with widening is narrower than two earlier revisions claimed. **No caret
glyph overhangs the code line in either render** — measured over the same 2,842 lines,
both are **0**, because the widening underline covers the trimmed statement, which ends
inside the line. What differs is the rendered *row*: widening appends the first detail
string to the caret row when it fits the inline budget, and on **all 20** of the lines
where that happens the text carries the row past the
end of the code line. Stacking never appends detail to a caret row, so it is 0. The
contrast is about where detail goes, not about bounding the underline.

All widths below are **final rendered columns**, measured after the same projection
`ApiOutputFormatter` applies — code lines receive `BodyIndentWidth`, hoisted caret
lines do not (`ApiOutputFormatter.cs:2504-2514`) — and the caret block is the
renderer's real output including detail rows, not a re-derived model. Two earlier
revisions of this section were withdrawn: one compared pre-projection code width against
post-hoist caret width and omitted detail rows; the other measured the whole population
instead of the focused one.

Whole rendered block, on the 2,842 lines this PR changes, old versus new:

| terminal | old (widen) | new (stacked) |
| --- | ---: | ---: |
| 80 cols | 23.3% | **23.4%** |
| 100 cols | 48.3% | 48.3% |
| 120 cols | 65.9% | 65.9% |
| 160 cols | 84.9% | 84.9% |

That denominator is padded, and the padding is most of it: a block can never be
narrower than its code line, so a line whose code alone overflows cannot fit under
*any* caret model. At 80 columns only **730 of the 2,842** have code that fits at all;
among those the rates are **90.5%** widening and **91.1%** stacked. At the wider
terminals the fit rate and the code-fits share converge almost exactly (100: 48.3% /
48.3%; 120: 65.9% / 66.0%; 160: 84.9% / 84.9%), which says the block width simply *is*
the code width there.

Terminal fit is unchanged, because on these dense lines the block width is set by the
code line and the wrapped detail rows (`Budget` = 100), which both renders share. Nor does the block
shrink: it is the same width on **91.8%** of these lines, narrower on **1.1%**, and
wider on **7.2%** — the numbered labels cost a few columns where the details were
already the widest thing in the block.

All three shares are padded, and three review rounds each found a different one, which
was the signal that shares were the wrong presentation. A block is never narrower than
its code line, so splitting on that floor settles it:

| population | unchanged | narrower | wider |
| --- | ---: | ---: | ---: |
| **2,612** pinned at the code width | 2,608 | — | 4 |
| **230** with headroom above it | 0 | 30 | 200 |

The "91.8% unchanged" is almost entirely the pinned population. Every line with room to
move moves, and 200 of the 230 move the wrong way. This also refuted my own earlier
correction — I had argued growth has no structural floor, and it does: only 4 of the
2,612 pinned blocks grow.

**Stacking buys attribution, and costs a little width.** Unpadded, fit goes 90.5% →
91.1% on the 730 lines that could fit. Block width moves only on the 230 with room to
move, 200 wider against 30 narrower. An
earlier revision claimed stacking "does not improve fit and does not narrow the block",
which was categorical where the data is marginal; and a revision before that claimed it
strictly improved fit, which was simply wrong.

The constraint still binds on the *alternatives*, which is why it is stated here.
Side-aligned annotations (each caret's text beside it) are the nicer picture, and were
mocked up against real CoreLib lines and rejected as the **default** purely on fit: they
overflow the code line on **97.52%** of the 29,933 lines carrying an extent, by a mean of
**81 columns** over the 29,191 that overflow, and fit **24.7%** of them at 80 columns —
**32.1%** unpadded — against **75.7%** for the bare
code.

## Measurements

CoreLib `11.0.0-preview.7.26366.102`, **annotated-source render** (the one `--focus`
produces), **summed over the five focus families** (`alloc`, `safety`, `cost`,
`lifetime`, `unsafe`).

Three hazards that silently corrupt figures here, all avoided. `ComputeSpans` must run
**after** `PrintRaised`, and an `IrFunction` must not be reused across renders (23 of
200 members render differently on a second `PrintRaised`). The third was found by
review, in an earlier revision of this very section: **the focus filter has to be
applied before anything is counted.** `ResearchViews.CollectFacts` returns every fact,
but `--focus` promotes only one family to carets, so counting pre-filter describes a
render no invocation produces. Earlier revisions of this PR body quoted that
pre-filter population (615 mixed / 499 single-extent / 3,385 stacking). Those numbers
were measured correctly and were answers to the wrong question. The figures below are
post-filter.

| | |
| --- | ---: |
| caret lines | 32,864 |
| …carrying at least one extent | 29,933 |
| …with a single distinct extent | 27,345 (91.4%) |
| …with two or more distinct extents | 2,588 (8.6%) |
| worst line | 8 distinct extents |
| mixed lines (facts with **and** without an extent) | 355 |
| …with exactly one surviving extent | 254 |

The worst line is 8 and it is reproducible by name — `System.Tuple`8.Equals` line 9
under `--focus alloc`, and `System.SpanHelpers.NonPackedContainsValueType` line 17 under
`--focus safety`. The distinct-extent distribution is 1: 27,345 · 2: 1,772 · 3: 608 ·
4: 135 · 5: 45 · 6: 19 · 7: 7 · 8: 2, and the multi-extent tail sums to 2,588.

### What this PR actually renders

The population that stacks is **2,842 lines** — the 2,588 multi-extent lines plus the
254 mixed single-extent lines that rule 7 promotes. Measured by calling
`AnnotationCaret.Render` on every one of them through the shipped geometry
(`hoist: true`, real member indent) and reading the output back:

| caret rows | lines | |
| --- | ---: | ---: |
| 1 | 2,543 | 89.5% |
| 2 | 297 | 10.5% |
| 3 | 1 | 0.0% |
| 4 | 1 | 0.0% |

That 89.5% is padded: 254 of these lines carry a single extent group and cannot occupy
two rows. Among the **2,588** lines with something to pack, **2,289 (88.4%)** take one
row.

**30,022 of the 32,864 caret lines (91.4%) keep today's geometry exactly** — the
complement of the 2,842 that stack. Two populations, unchanged for different reasons:
**27,091** narrow to one agreed extent, and **2,931** have no fact with an extent and
widen as they always did. An earlier revision quoted only the first (82.4%) as the
unchanged share, which dropped the entire no-extent population.

#### Trail widths, with the padding removed

3,884 of 6,566 trails (59.2%) render at true width. That headline is **padded by a
structural immunity** and review was right to reject it as a packing success rate: the
last trail on a row has no successor, so its clip limit is `int.MaxValue` and it renders
at true width *by construction*.

Measured rather than derived: those 2,842 lines occupy **3,144 rows**, and all **3,144
of 3,144** last-on-row trails render at true width — exactly as the derivation predicts.
Of the **3,422** trails actually exposed to a successor, 740 (21.6%) survived at true
width. Five of those are immune too — their extents are no longer than `MinTrail`, which
row admission already reserves — so among the **3,417** trails that could genuinely be
cut, **735 (21.5%)** survived and **2,682 (78.5%)** were clipped. It reconciles:
3,144 + 3,422 = 6,566 and 3,144 + 740 = 3,884.

Of those 2,682 clipped trails, the successor extent whose label does the clipping is
nested inside the trail on **1,867 (69.6%)** and wholly disjoint from it on **815
(30.4%)**. A disjoint successor still clips because its label is written two columns
left of the extent it belongs to. Clipping is concentrated at nestings but is not caused
by nesting — a trail is cut short only by the next label **on its own row**, and two
extents sharing a start column can never share a row, so a nested extent is never
positioned to clip its parent.

#### Widest-first does not order row widths

An earlier revision claimed reading down the rows was zooming in. It is not: a later
disjoint extent can be packed onto a lower row and reach further right than anything
above it. That happens on **50 of the 2,842** lines.

#### The stacked block never overhangs — structurally, not measurably

No stacked caret row is wider than the code line it annotates. This is **structural**,
so the corpus sweep confirming it could not have failed: `Stack` sends any extent with
`Column + Length > lineLength` to the unplaced list, labels grow leftward, and rule 8
bails rather than shifting a trail rightward.

The contrast with widening is narrower than it looks, and review corrected it twice.
**No caret glyph overhangs in either render** — 0 and 0 over the same 2,842 lines,
because widening underlines the trimmed statement, which ends inside the line. An
earlier revision claimed the widening underline "shifts rightward while preserving the
full extent length", which describes a mechanism that does not occur; the probe behind
it measured the whole rendered row, and on the widening path the first detail string is
appended *after* the caret, so it never measured a caret at all.

What differs is the rendered **row**. Widening appends that first detail string when it
fits the inline budget, carrying the row past the code line. Quoting this as **20 of
2,842 (0.70%)** pads the denominator with the 2,822 whose detail never goes inline and
so cannot overhang: inline detail appears on exactly **20** of these lines and **all 20**
overhang, so the rate is **20/20**. It is structural — widening underlines the whole
trimmed statement, so the caret ends where the code line ends, and anything appended
lands past it.

#### The whole-line fallback never fires — and the first probe for it was vacuous

`RenderStacked` returns null — discarding every caret on the line and widening — when an
extent starts too close to the gutter for its own label to fit.

The first probe for this **could not have detected it**. It tested for a rendered block
with no caret row, but falling back still renders the widening caret, so the reported
zero was unfalsifiable. Re-measured by testing for the absence of an `N.` label before
the caret run: **0 firings**, and the row histogram is unchanged.

The denominator was also circular in an earlier revision — "0 of the lines that stack"
is empty by construction, because a line the guard rejects does not stack. The 0 is over
the **2,842 lines that qualify to stack before the guard runs**.

Non-vacuity was then proved by mutating only the guard margin:

| margin | lines that hit the fallback |
| --- | ---: |
| `+2` (shipped) | 0 |
| `+6` | 306 |
| `+200` | 2,842 |

Graded, not binary. The guard is a real branch with a reachable boundary; this corpus
sits below it.

## Non-action boundary

- Lines whose facts **agree**, and lines carrying a single fact, keep exactly today's
  geometry including the inline-detail shortcut.
- Lines where **no** fact has an extent still widen — there is nothing to point at.
- `AnnotationAnchor` is untouched. No fact moves, no extent changes. This is a change
  to `AnnotationCaret` alone.
- The eight `alloc.box` facts on `Tuple`8.Equals`'s *first* operands still arrive
  without an extent. That is an anchoring gap this PR makes visible rather than causes.

## Validation

Five gates, each proven non-vacuous by mutation:

| gate | pins |
| --- | --- |
| `CaretsStackWhenTwoFactsOnTheLineDisagreeAboutTheExtent` | disagreement stacks rather than widens |
| `CaretIsDrawnAndFactsWithoutAnExtentAreMarkedWhenTheLineIsMixed` | the mixed line draws its caret and marks the rest `-` |
| `FactsSharingAnExtentShareOneCaretAndOneNumber` | rule 1: the unit is the extent, so two facts about the same characters get one caret and one number |
| `CaretStillWidensWhenNoFactOnTheLineHasAnExtent` | the boundary: a line with nothing placeable still widens |
| `Member_AnnotatedSource_FocusStacksACaretPerExtentWhenFactsOnALineDisagree` | end-to-end through the CLI: each caret underlines `a1`…`a4` |

Mutants, each rebuilt and re-run:

- restoring the old `Count: > 1` threshold kills **only** the mixed gate — so that gate
  pins the 254-line decision specifically and nothing else;
- disabling stacking entirely kills the first three;
- dropping `extents` at the CLI call site (`ResearchViews.cs:437`, passing `extents: null`)
  kills **only** the end-to-end gate: all 30 renderer gates stay green, because they call
  `AnnotationCaret.Render` directly and never exercise that wiring. So the CLI gate is not
  a duplicate of the renderer gates — it is the only thing pinning that the extents
  actually reach the renderer in the shipped path;
- replacing the extent-grouping lookup with a constant `-1`, so every fact gets its own
  caret, kills **only** `FactsSharingAnExtentShareOneCaretAndOneNumber`. Before that gate
  existed this mutant was **green on all 29 other tests**: rule 1 was ungated;
- relaxing the `Count: > 0` guard to `{ }` kills **only**
  `CaretStillWidensWhenNoFactOnTheLineHasAnExtent`, and only after that test was
  corrected. It previously passed an *empty* extent dictionary, which short-circuits
  inside `Stack` before the guard is reached, so it survived this mutant and gated
  nothing. It now passes the production shape — a member-wide dictionary carrying no
  entry for this line's facts.

Two of the five gates were therefore vacuous or absent until review pressed on them.

Suites (`dotnet run --project … -c Release`, since `dotnet test` runs zero tests here):

- Decompiler **4,713 / 0 failed / 0 skipped** at this head. Reproduced independently by
  a reviewer in its own worktree (4,713 / 0), and again by me after the rebase onto
  `main`. The base `fbabdf6d0` measures **4,711 / 0 failed / 0 skipped**, so the delta is
  **+2** — measured, not inferred from the diff. The figure rose from the 4,694 quoted in
  earlier revisions of this body because #3617 merged into the base and brought its own
  tests; it is not growth in this PR.
- CLI **2,794 total, 1 failed, 14 skipped** — `Layout_Count_CountsFilesRatherThanRenderedTreeLines`.
  That test lives in `CommandExecutionTests.cs`, which this PR edits, so "unrelated" was
  checked rather than asserted: `main` at `fbabdf6d0` produces **2,794 / 1 failed / 14
  skipped** with the same test failing. Identical totals, identical failure.

That +2 is exactly what the diff predicts: it adds four gates and
retires two older ones that pinned the widening behaviour it replaces
(`CaretWidensWhenTwoFactsOnTheLineDisagreeAboutTheExtent`,
`CaretWidensWhenOnlySomeFactsOnTheLineHaveAnExtent`). The CLI count is unchanged at +0 —
`Member_AnnotatedSource_FocusWidensToTheStatementWhenFactsOnALineDisagree` became
`Member_AnnotatedSource_FocusStacksACaretPerExtentWhenFactsOnALineDisagree`.

## What review changed

Ten rounds against fixed heads. The comments in this PR were wrong in eight places, and
every one was a claim about behaviour that had never been tested against the code
implementing it:

1. **Wrong population** — every figure was measured before the focus filter, describing a
   render no invocation produces (615/499/3,385 → 355/254/2,842).
2. **A vacuous gate** — the fallback probe tested for a missing caret row, but a fallback
   still renders the widening caret, so it could never fire.
3. **A second vacuous claim** — zero stacked overhang is structural, not measured.
4. **A circular denominator** — "0 of the lines that stack", when a rejected line does not
   stack by construction.
5. **A false structural claim** — widest-first does not order row widths (50 violations).
6. **A mechanism error** — clipping is caused by the next label on the row, not by nesting.
7. **A false provenance claim** — widest-first is the *reverse* of the printer's record
   order, which is descendants-before-ancestors.
8. **Padded denominators**, found in two passes — the 59.2% true-width rate counts 3,144
   trails structurally immune to clipping, and the 21.6% replacement still counted five
   more; the clip-capable rate is 21.5%. The 89.5% one-row rate counts 254 lines that
   cannot take two rows (88.4% among those that can), and the 82.43% agreement rate
   counts 29,550 single-fact lines that cannot disagree (0% among the 3,314 that can).
9. **A figure corrected into an error, twice** — the Tuple caret is **370** columns.
   It was recorded as 370, an earlier round "corrected" it to 330 on a recapture of the
   wrong render, and a falsification pass caught that and measured it back to 370. It
   survived because nothing else depended on it, so it stayed internally consistent
   while being false. Both the doc and the remark now say to re-derive it from the
   render, never from the surrounding prose.
10. **A false comparison** — no caret glyph overhangs the code line in *either* render.
   The 20 overhangs are whole-row, caused by inline detail text, and my probe had
   silently measured the detail along with the caret.

Plus two unqualified guarantees (the `-` marker holds only when the line stacks) and one
false mechanism for the widening overhang.

Two of the five gates were vacuous or absent until review pressed on them, and the
non-vacuity of the rule 8 guard is now demonstrated by margin mutation rather than
asserted.

### DIFF
--- a/src/ILInspector.Decompiler.Tests/AnnotationGestureTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationGestureTests.cs
@@ -206,10 +206,13 @@ public class AnnotationGestureTests
         // narrowing would underline an expression true of only some of the
         // facts sharing the caret. Marking the extent-less fact '-' removes the
         // ambiguity that justified widening, so the surviving extent is drawn.
-        // 615 lines of System.Private.CoreLib, as the annotated-source view
-        // prints it, mix facts with and without extents; 499 of them have
-        // exactly one surviving extent, so this is the majority shape. (The
-        // overlay render is a different population: 609 and 498.)
+        // 355 lines of System.Private.CoreLib, as the annotated-source view
+        // prints it, mix facts with and without extents; 254 of them have
+        // exactly one surviving extent, so this is the majority shape. That is
+        // summed over the five focus families: --focus promotes only one family
+        // to carets, so a count over every collected fact would describe a
+        // render no invocation produces. The two categories below are a unit
+        // convenience -- Stack is category-agnostic and never sees the filter.
         const string Line = "        Sink(new object());";
         var narrowed = Fact(Alloc, "has-extent");
         var bare = Fact(Unsafety, "no-extent");
--- a/src/ILInspector.Decompiler.Tests/AnnotationGestureTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationGestureTests.cs
@@ -206,8 +206,10 @@ public class AnnotationGestureTests
         // narrowing would underline an expression true of only some of the
         // facts sharing the caret. Marking the extent-less fact '-' removes the
         // ambiguity that justified widening, so the surviving extent is drawn.
-        // 615 CoreLib lines mix facts with and without extents; 499 of them
-        // have exactly one surviving extent, so this is the majority shape.
+        // 615 lines of System.Private.CoreLib, as the annotated-source view
+        // prints it, mix facts with and without extents; 499 of them have
+        // exactly one surviving extent, so this is the majority shape. (The
+        // overlay render is a different population: 609 and 498.)
         const string Line = "        Sink(new object());";
         var narrowed = Fact(Alloc, "has-extent");
         var bare = Fact(Unsafety, "no-extent");
--- a/src/ILInspector.Decompiler.Tests/AnnotationGestureTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationGestureTests.cs
@@ -252,9 +252,12 @@ public class AnnotationGestureTests
         var first = Fact(Alloc, "one");
         var second = Fact(Unsafety, "two");
 
-        // The dictionary is member-wide, so it is non-empty even when no fact
-        // on *this* line has an extent. Passing an empty one instead would
-        // short-circuit inside Stack and never reach the guard under test.
+        // The dictionary is member-wide in production, so it is non-empty even
+        // when no fact on *this* line has an extent. That is the shape worth
+        // gating: it makes Stack return an empty group list, which is what the
+        // Count: > 0 guard has to reject. An empty dictionary would reach the
+        // same guard today, but only because the short-circuit that used to
+        // intercept it inside Stack has since been deleted.
         var elsewhere = Fact(Alloc, "on some other line");
         var extents = new Dictionary<IAnnotation, AnnotationAnchor.CaretExtent>
         {
--- a/src/ILInspector.Decompiler.Tests/AnnotationGestureTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationGestureTests.cs
@@ -252,7 +252,15 @@ public class AnnotationGestureTests
         var first = Fact(Alloc, "one");
         var second = Fact(Unsafety, "two");
 
-        var extents = new Dictionary<IAnnotation, AnnotationAnchor.CaretExtent>();
+        // The dictionary is member-wide, so it is non-empty even when no fact
+        // on *this* line has an extent. Passing an empty one instead would
+        // short-circuit inside Stack and never reach the guard under test.
+        var elsewhere = Fact(Alloc, "on some other line");
+        var extents = new Dictionary<IAnnotation, AnnotationAnchor.CaretExtent>
+        {
+            [elsewhere] = new AnnotationAnchor.CaretExtent(0, 4),
+        };
+
         string caretLine = AnnotationCaret.Render(Line, "    ", [first, second], extents: extents)[0];
 
         int statementColumn = Line.Length - Line.TrimStart().Length;
@@ -284,8 +292,9 @@ public class AnnotationGestureTests
         var rendered = AnnotationCaret.Render(Line, "    ", [first, second, third], extents: extents);
         string caretLine = rendered[0];
 
-        // Three facts, two extents, so two carets and two numbers -- not three.
-        Assert.Equal(2, caretLine.Count(c => c == '.'));
+        // Three facts, two extents, so two numbered carets -- not three. The
+        // group count is pinned structurally by the caret runs below; this
+        // pins that the numbering agrees with it.
         Assert.Contains("1.", caretLine, StringComparison.Ordinal);
         Assert.Contains("2.", caretLine, StringComparison.Ordinal);
         Assert.DoesNotContain("3.", caretLine, StringComparison.Ordinal);
--- a/src/ILInspector.Decompiler.Tests/AnnotationGestureTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationGestureTests.cs
@@ -261,6 +261,64 @@ public class AnnotationGestureTests
     }
 
     [Fact]
+    public void FactsSharingAnExtentShareOneCaretAndOneNumber()
+    {
+        // Rule 1 of the display model: the unit is the extent, not the fact.
+        // Two facts about the same characters get one caret and one number,
+        // with both texts listed under it -- otherwise a line would grow a
+        // redundant caret per fact and re-underline the same characters.
+        const string Line = "        Sink(new object(), new int[1]);";
+        var first = Fact(Alloc, "one");
+        var second = Fact(Alloc, "two");
+        var third = Fact(Unsafety, "three");
+
+        int shared = Line.IndexOf("new object()", StringComparison.Ordinal);
+        int other = Line.IndexOf("new int[1]", StringComparison.Ordinal);
+        var extents = new Dictionary<IAnnotation, AnnotationAnchor.CaretExtent>
+        {
+            [first] = new AnnotationAnchor.CaretExtent(shared, "new object()".Length),
+            [second] = new AnnotationAnchor.CaretExtent(shared, "new object()".Length),
+            [third] = new AnnotationAnchor.CaretExtent(other, "new int[1]".Length),
+        };
+
+        var rendered = AnnotationCaret.Render(Line, "    ", [first, second, third], extents: extents);
+        string caretLine = rendered[0];
+
+        // Three facts, two extents, so two carets and two numbers -- not three.
+        Assert.Equal(2, caretLine.Count(c => c == '.'));
+        Assert.Contains("1.", caretLine, StringComparison.Ordinal);
+        Assert.Contains("2.", caretLine, StringComparison.Ordinal);
+        Assert.DoesNotContain("3.", caretLine, StringComparison.Ordinal);
+
+        // Non-vacuity: the two carets really are at the two distinct extents.
+        // The first trail collides with the second caret's label, so rule 5
+        // clips its last column to the truncation glyph rather than overrun it;
+        // the second trail is last on the row and keeps its true width.
+        Assert.Equal(shared, caretLine.IndexOf('^'));
+        Assert.Equal(1, caretLine.Count(c => c == '~'));
+
+        var runs = new List<int>();
+        for (int c = 0; c < caretLine.Length; c++)
+        {
+            if (caretLine[c] != '^' || (c > 0 && caretLine[c - 1] == '^'))
+                continue;
+            int width = caretLine.AsSpan(c).IndexOfAnyExcept('^');
+            runs.Add(width < 0 ? caretLine.Length - c : width);
+        }
+
+        Assert.Equal(2, runs.Count);
+        Assert.True(runs[0] < "new object()".Length, "the colliding trail must be clipped");
+        Assert.Equal("new int[1]".Length, runs[1]);
+
+        // Both facts sharing the extent are listed, and only the first carries
+        // the number: the second is indented to sit under it.
+        Assert.Contains(rendered, l => l.Contains("1. alloc.new(one)", StringComparison.Ordinal));
+        Assert.Contains(rendered, l => l.Contains("alloc.new(two)", StringComparison.Ordinal)
+            && !l.Contains("2.", StringComparison.Ordinal));
+        Assert.Contains(rendered, l => l.Contains("2. unsafe.stackalloc(three)", StringComparison.Ordinal));
+    }
+
+    [Fact]
     public void NothingIsRenderedForABlankLineOrNoFacts()
     {
         Assert.Empty(AnnotationCaret.Render("        Work();", "    ", []));
--- a/src/ILInspector.Decompiler.Tests/AnnotationGestureTests.cs
+++ b/src/ILInspector.Decompiler.Tests/AnnotationGestureTests.cs
@@ -153,13 +153,12 @@ public class AnnotationGestureTests
     }
 
     [Fact]
-    public void CaretWidensWhenTwoFactsOnTheLineDisagreeAboutTheExtent()
+    public void CaretsStackWhenTwoFactsOnTheLineDisagreeAboutTheExtent()
     {
-        // The other half of the all-or-nothing rule: both facts narrow, but to
-        // different characters. Stacking two carets was rejected as a design, so
-        // the only correct answer is the statement -- the smallest span true of
-        // both. Distinct from the partial case below, which is a missing entry
-        // rather than a conflicting one, and gated by a different branch.
+        // Both facts narrow, but to different characters. Widening to the
+        // statement was the old answer; it points at everything and therefore
+        // at nothing, and leaves the details below unattributable. Each extent
+        // now gets its own numbered caret. See docs/design/caret-stacking.md.
         const string Line = "        Sink(new object(), new int[1]);";
         var first = Fact(Alloc, "object");
         var second = Fact(Unsafety, "array");
@@ -177,21 +176,38 @@ public class AnnotationGestureTests
         Assert.Equal(firstColumn, AnnotationCaret.Render(Line, "    ", [first], extents: extents)[0].IndexOf('^'));
         Assert.Equal(secondColumn, AnnotationCaret.Render(Line, "    ", [second], extents: extents)[0].IndexOf('^'));
 
-        string shared = AnnotationCaret.Render(Line, "    ", [first, second], extents: extents)[0];
+        var rendered = AnnotationCaret.Render(Line, "    ", [first, second], extents: extents);
+        string caretLine = rendered[0];
+
+        // Two carets on one row, each still at the column of its own extent and
+        // at its own true width -- not one widened underline.
+        Assert.Equal(2, caretLine.Count(c => c == '.'));
+        Assert.Equal(firstColumn, caretLine.IndexOf('^'));
+        Assert.Equal(secondColumn, caretLine.LastIndexOf('^') - "new int[1]".Length + 1);
+
+        // The statement underline is precisely what must no longer appear.
         int statementColumn = Line.Length - Line.TrimStart().Length;
-        Assert.Equal(statementColumn, shared.IndexOf('^'));
-        Assert.Equal(Line.Trim().Length, shared.Count(c => c == '^'));
+        Assert.NotEqual(Line.Trim().Length, caretLine.Count(c => c == '^'));
+        Assert.NotEqual(statementColumn, caretLine.IndexOf('^'));
+
+        // Each caret is labelled, and each label has a matching numbered detail
+        // row, so a reader can attribute every fact to a specific expression.
+        Assert.Contains("1.", caretLine, StringComparison.Ordinal);
+        Assert.Contains("2.", caretLine, StringComparison.Ordinal);
+        Assert.Contains(rendered, l => l.Contains("1. alloc.new(object)", StringComparison.Ordinal));
+        Assert.Contains(rendered, l => l.Contains("2. unsafe.stackalloc(array)", StringComparison.Ordinal));
     }
 
     [Fact]
-    public void CaretWidensWhenOnlySomeFactsOnTheLineHaveAnExtent()
+    public void CaretIsDrawnAndFactsWithoutAnExtentAreMarkedWhenTheLineIsMixed()
     {
-        // The agreement rule is all-or-nothing, and the case that matters is
-        // PARTIAL: one fact narrows, another has no printed node to point at.
-        // Narrowing there would underline an expression that is true of only
-        // some of the facts sharing the caret. 615 lines in CoreLib mix facts
-        // with and without extents, 499 of them with a single surviving extent,
-        // so this is the difference between right and wrong on real output.
+        // The PARTIAL case: one fact narrows, another has no printed node to
+        // point at. This used to widen to the statement, on the reasoning that
+        // narrowing would underline an expression true of only some of the
+        // facts sharing the caret. Marking the extent-less fact '-' removes the
+        // ambiguity that justified widening, so the surviving extent is drawn.
+        // 615 CoreLib lines mix facts with and without extents; 499 of them
+        // have exactly one surviving extent, so this is the majority shape.
         const string Line = "        Sink(new object());";
         var narrowed = Fact(Alloc, "has-extent");
         var bare = Fact(Unsafety, "no-extent");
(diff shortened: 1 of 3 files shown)


richlander:
## No CI runs on this slice — read this before treating anything here as green

`.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`. This PR
targets `fix/3570-caret-extent` (#3617), so **no workflow is scheduled for it at all** —
`gh pr checks 3656` reports "no checks reported", which is terminal, not pending.
Nothing here is CI-validated, and the absence of a red mark is not a green one.

The evidence for this slice is therefore what I ran locally, plus #3617's own CI:

- Decompiler suite **4,693 passed / 0 failed** — `dotnet run --project src/ILInspector.Decompiler.Tests -c Release`
  (`dotnet test` silently runs zero tests in this repo, so it is not used).
- CLI suite **2,766, 1 failed / 14 skipped**. The failure is
  `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which is unrelated to carets
  and fails identically on the base branch.
- Full solution built via `dotnet build dotnet-inspect.slnx -c Release` — required, or
  fixture-dependent decompiler tests throw a phantom `FileNotFoundException`. One of my
  earlier runs hit exactly that and reported a false failure.
- Three gates mutation-tested: restoring the old `Count: > 1` threshold kills only the
  mixed-line gate; disabling stacking kills all three.

Once #3617 merges, this PR retargets `main` and CI becomes applicable. I will re-check
then and report the result rather than assuming it.

richlander:
## Review round 2 — two gates were not gating

Both reviewers returned at `d41a236ff`. Gemini APPROVE; GPT BLOCKING with three findings. All three verified independently, all three correct, all fixed at `e5906ec13`.

The important one is that **a gate I had claimed as mutation-proven was in fact vacuous**, and I had missed it because I mutated the wrong direction.

`CaretStillWidensWhenNoFactOnTheLineHasAnExtent` passed an **empty** extent dictionary. `Stack` returns null at its own `extents.Count == 0` check before the `Count: > 0` guard in `Render` is ever reached, so relaxing that guard to `{ }` left the test green. I had previously mutated `> 0` to `> 1` and watched the mixed gate die, and took that as coverage of the guard. It was not: it covered the *upper* boundary only, and the lower boundary — the one this test exists for — was untested.

The dictionary is member-wide in production, so the shape that actually reaches the guard is a non-empty dictionary holding no entry for *this line's* facts, which makes `Stack` return an empty group list. The test now passes that shape and dies under the `{ }` mutant, verified both ways.

That is the second ungated rule this round. The first was extent grouping — rule 1 of the display model — where replacing the group lookup with a constant `-1`, so every fact gets its own caret, was **green on all 29 other tests**.

Also fixed:

- The class remark said statement widening happens *only* when no fact on the line has an extent. `RenderStacked` widens too when a label will not fit beside the gutter. That path fires on 0 of 3,385 CoreLib lines, but it is reachable public behaviour for inputs this corpus does not contain, so "only" was a false absolute. Zero traffic is not the same as unreachable, and the doc should not conflate them.
- The validation section said "three gates". It is five, and the mutation table now names the mutant for each rather than asserting the property in aggregate.

Gemini's approve carried a fair caution: the grouping gate asserted `caretLine.Count(c => c == '.') == 2`, which would break if detail were ever trailed onto the caret row. The run-length assertions already pin the group count, so the character count was redundant and brittle; it is gone. Because that *weakened* an assertion, I re-ran the grouping mutant afterwards rather than assuming the gate survived the edit — it still dies.

**Suites at `e5906ec13`**, full runs, worktree untouched during them:

- Decompiler **4,694 / 0** (4,693 before this PR, plus the new gate)
- CLI **2,766, 1 failed / 14 skipped** — `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, unrelated and failing identically on base

An earlier full run was discarded rather than reported: I had rebuilt the test project and run two mutants in the same worktree while it was in flight, so its result described no single state of the tree.

Still no CI here — this PR targets `fix/3570-caret-extent`, so `ci.yml` (`pull_request: branches: [main]`) does not trigger. Absence of checks is terminal, not pending, until #3617 merges and this retargets `main`.

Both reviewers re-tasked at `e5906ec13`.

richlander:
## Review round 3 — head `0b9cfb0d8`

Gemini APPROVE at `e5906ec13`, having independently mutation-verified all five gates and reproduced each kill. GPT BLOCKING with four findings; all four verified, all four correct, all fixed.

**The two reviewers converged independently on the same structural point**, which is worth more than either finding alone: `Stack`'s `extents.Count == 0` early return is subsumed by the `Count: > 0` guard in `Render`. Gemini deleted it and ran the full suite — 0 failures. With an empty dictionary every fact falls through to `unplaced`, `Stack` returns an empty group list, and the guard widens the line: exactly what the early return produced. It was a branch with no behaviour of its own and no way to gate it, so it is now deleted rather than documented. `Count: > 0` is the single semantic boundary, and it is the one the tests pin.

That check is also what made the boundary gate vacuous last round — the test's empty dictionary was short-circuiting there instead of reaching the guard. Removing it means that shape can no longer hide a hole.

The other three findings:

- **Figures did not name their render.** "615 lines in the corpus" is ambiguous in a way that matters: annotated-source gives 615/499, the overlay gives 609/498. A reader reproducing the number could get a different one and be right. Every figure now names the assembly and the render.
- **`Agreed`'s remark asserted a mechanism, not a measurement**: "density and disagreement are the same phenomenon — more facts on a line means more distinct offsets on it". That is not what was measured, and not what the method does — it compares rendered extents, not offsets, and two facts can share one extent, which is precisely what the grouping gate pins. Reduced to the correlation actually measured.
- **The end-to-end gate had no named mutant.** Fixed by running one rather than asserting it: passing `extents: null` at `ResearchViews.cs:437` kills that gate while **all 30 renderer gates stay green**, because they call `AnnotationCaret.Render` directly and never exercise the wiring. So the CLI gate is not a duplicate — it is the only thing pinning that extents reach the renderer in the shipped path.
- **The base test count was wrong.** I wrote "4,693 before this PR". The base has **4,692**; 4,693 was an intermediate head of this PR. Verified by counting the gate attributes on the base file: 18 there, 20 here, so this PR adds exactly two tests.

**Suites at `0b9cfb0d8`**, worktree untouched during both runs:

- Decompiler **4,694 / 0**
- CLI **2,766, 1 failed / 14 skipped** — `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, unrelated, identical on base

Across three rounds the defect pattern here has been consistent and worth naming: **the logic was sound in every round, and the claims about it were not.** Two of five gates were vacuous or absent, one branch was untestable, and roughly a dozen prose figures were wrong, ambiguous, or asserted mechanisms nobody had measured. Every one was found by mutation or re-measurement, none by reading.

Both reviewers re-tasked at `0b9cfb0d8`.

richlander:
## Correction: my previous comment described fixes that were not in the branch

Retracting part of the round-3 comment above, and force-pushing a repaired branch. Head is now `427cc18b4`; `0b9cfb0d8` and `e25c2d10b` are gone.

**What happened.** Both review agents were working in the same worktree I was editing and building in. One of them wrote a mutation-runner script into the tree and used `git checkout` to revert files between mutants. That reverted my uncommitted prose edits mid-build, and my `git add -A` then committed the result. The consequences:

- **`e25c2d10b` contained none of the changes its message described.** It carried exactly one file: the reviewer's `run_mutants.sh`. The commit message is a detailed account of prose fixes that are not in the commit. That is the worst artifact of this whole PR, and I created it by committing with `-A` from a tree I did not control.
- **My round-3 comment claimed those prose fixes had been made.** They had not. The causal claim in `Agreed` and the unnamed-render figures were still exactly as GPT had described them. That claim is withdrawn; the fixes exist now, at `427cc18b4`, and I have verified their presence at HEAD rather than asserting it.
- **A later commit picked up two live mutations.** `extents: null` at two `AnnotationCaret.Render` call sites in `ResearchViews.cs` — the reviewer's own mutant for the CLI gate, left in the tree. **This disables caret extents in the shipped CLI path entirely.** It was staged, committed, and would have been pushed.

**Why my checks did not catch it.** After that commit I ran only the decompiler gesture class, which calls `AnnotationCaret.Render` directly and never touches the CLI wiring — so all 30 gates passed with the feature disabled downstream. The gate that would have caught it is the end-to-end CLI test, and it exists precisely because the renderer gates cannot see that wiring. I had argued that point in this PR and then failed to act on it. Running the narrow suite after a change to a wiring-adjacent tree was the mistake.

It was caught by GPT reporting the prose was unchanged at HEAD, which did not match what I believed I had committed, and by auditing `git show --stat` per commit and the full `f26fa8e0a..HEAD` diff rather than trusting the tree.

**Repair.** `e25c2d10b` dropped by rebase; `ResearchViews.cs` restored from base; prose fixes re-applied and their presence verified at HEAD; the branch diff now touches exactly three files and no others. The full diff was scanned for injected patterns (`&& false`, `extents: null`, `= -1`, stray `Console.Write`) and is clean.

**Gemini's approval above is void, through no fault of its reasoning.** It reviewed `0b9cfb0d8` but reported prose that was not in that commit — it was reading my working tree while I edited it. Its verdict describes a state that never existed as a commit, so it cannot clear this PR.

**Process change.** Reviewers now get their own detached worktrees at a fixed commit (`/home/rich/rev-stack-a`, `/home/rich/rev-stack-b`), created from the pushed head. I will not build or test in a tree a reviewer can write to, and I will not use `git add -A` while one is running. Verification at `427cc18b4` re-run from scratch, including the CLI gate.

The substance of GPT's four findings this round was correct and all four are fixed. But the ordering matters: two of them were only *reported* as fixed last round.

richlander:
## Round 4 — both reviewers clean at `8c17dcb35`

GPT APPROVE, Gemini APPROVE, both at the committed head in isolated worktrees, both stating what they checked rather than approving bare.

GPT's last finding was correct and is fixed in `8c17dcb35`: three comments still said only "CoreLib", two of them in the test files — where a reader is most likely to try reproducing a number. The commit before it had claimed *every* figure named its render, which made that message an instance of the very defect it was fixing. All three are now scoped, and I re-scanned the whole branch diff for bare "CoreLib" and found none.

### Where this ended up

| | |
| --- | --- |
| files touched | 3 (`AnnotationCaret.cs`, `AnnotationGestureTests.cs`, `CommandExecutionTests.cs`) |
| gates | 5, each with a named mutant that kills it and only it |
| decompiler | **4,694 / 0** (base 4,692 + the two gates this PR adds) |
| CLI | **2,766, 1 failed / 14 skipped** — pre-existing, identical on base |

### What the four rounds actually found

The logic was sound in every round. Almost nothing found was a bug in the rendering — it was claims about the rendering that had not been earned:

- **Two of five gates were vacuous or absent.** Extent grouping — rule 1 of the display model — had no gate at all; a constant `-1` in place of the group lookup was green on 29 tests. The boundary gate passed an empty dictionary that short-circuited inside `Stack` and never reached the guard it named, so relaxing that guard left it green.
- **One branch was untestable.** `Stack`'s `extents.Count == 0` return was subsumed by `Count: > 0`; both reviewers reached this independently, and it is now deleted rather than documented.
- **Roughly a dozen prose figures** were wrong, ambiguous about their render, or asserted mechanisms nobody had measured. 331 was 330; "the null lines are precisely the dense ones" was false in 2,156 of 6,012 cases; the terminal-fit table mixed two coordinate systems; and the A/B that should have existed from the start showed stacking does **not** improve terminal fit — a claim I had been repeating. It buys attribution and block width instead.

None of that was found by reading. All of it came from mutation or re-measurement.

### And one failure of process, not of analysis

Both reviewers were working in the tree I was editing and building in. A reviewer's mutation runner reverted my uncommitted edits mid-build; `git add -A` then committed the result. That produced a commit containing none of the changes its message described, and later swept up two live mutations that **disable caret extents in the shipped CLI path**. Committed, and nearly pushed.

It survived my checks because I ran only the renderer gates afterwards — and those call `AnnotationCaret.Render` directly, so all 30 pass happily with the CLI wiring severed. The gate that catches it is the end-to-end test, which exists for exactly that reason. I had made that argument in this PR and then failed to apply it.

Repaired by rebase, and the process changed: reviewers now get detached worktrees pinned at the pushed head, I do not build in a tree they can write to, and no more `git add -A`. Both then re-verified the repair from committed state.

### Merge

Ready, with one gate outstanding that is not mine to close: this targets `fix/3570-caret-extent`, so `ci.yml` (`pull_request: branches: [main]`) never triggers and the absence of checks here is terminal, not pending. **#3617 merges first, this retargets `main`, and CI must then be green before merge.** I will confirm that after retarget rather than assume it carries.

PR #3657, opened 2026-08-01
Specify the caret stacking display model

richlander:
Specifies **caret stacking** — the display model for `--focus`, which renders one
caret per fact extent on a line instead of widening to a statement-length caret
whenever the facts on a line disagree.

Documentation only. No code, no behaviour change. The implementation is #3656.

## Why a design doc

Caret layout is UX, and it had never been designed — it had accreted. The
existing behaviour (widen on disagreement) was a reasonable local decision that
turned out to be wrong on exactly the lines that carry the most information:
`System.Tuple<…>.Equals` renders 16 facts under one 370-column caret and
attributes none of them.

The doc locks **one** style, specified as eight numbered rules — six for the
core layout, one for mixed lines, and one for the gutter-fit fallback — and
records what was rejected and on what evidence.

## What it settles

- **The unit is the extent, not the fact.** Facts about the same characters share
  one caret and one number.
- **Ordering** is start column ascending, widest first at a tie. Extents sharing a
  start column are always a nesting and cannot share a row, so the outer one goes
  above the inner one. That orders *same-start* extents only: it does not order row
  widths overall (50 of 2,842 lines have a lower row wider than the row above), and
  it is the reverse of the printer's record order, which is
  descendants-before-ancestors.
- **True width, clipped only on collision.** Width is information the anchoring
  layer worked to recover, so the renderer states it truthfully or marks with `~`
  that it could not.
- **Mixed lines** — some facts with an extent, some without — draw the carets they
  can and mark the rest `-`. This reverses the previous rule, and the doc says why:
  the `-` marker removes the ambiguity that justified widening. The marker applies
  only when the line stacks; a line the gutter rule rejects widens and marks nothing.
- **A spill threshold is not a render floor.** A one-character expression renders
  one caret, never a padded two.

## The measurement that decided the style

The primary consumer is a terminal. A layout wider than the terminal wraps, and a
wrapped caret row lands under characters it has nothing to do with — worse than no
caret, because it actively misleads.

All widths are **final rendered columns**, after the same projection
`ApiOutputFormatter` applies: code lines receive `BodyIndentWidth`, hoisted caret
lines do not. The caret block is real rendered output, detail rows included.

**A stacked caret row never adds width the code line did not already have** — and
that is *structural*, which is the correction review forced. `Stack` sends any extent
reaching past the line length to the unplaced list, labels grow leftward, and the
gutter rule bails rather than shifting a trail right, so no corpus could produce a
counter-example. The document now derives the bound and says the observed 0 cannot
falsify it.

The contrast with widening is narrower than two revisions of the doc claimed, and the
final round corrected it: **no caret glyph overhangs the code line in either render**
(0 and 0 over the same 2,842 lines), because the widening underline covers the trimmed
statement, which ends inside the line. What differs is the rendered *row* — widening
appends the first detail string to the caret row when it fits the inline budget, and
that text carries the row past the code line. Quoting that as **20 of 2,842 (0.70%)**
pads the denominator with the 2,822 whose detail never goes inline and so cannot
overhang; inline detail appears on exactly **20** lines and **all 20** overhang, so the
rate is **20/20** and it is structural. Stacking never appends detail to a caret row.
The contrast is about detail placement, not underline bounding.

Side-aligned annotations (each caret's text beside it) are genuinely the nicer
picture: association becomes spatial, numbering disappears, and clipping disappears.
They are rejected as the **default** purely on fit — they overflow the code line on
**97.52%** of the 29,933 lines carrying an extent, by a mean of **81 columns** over the
29,191 that overflow, and fit **24.7%** of them at 80 columns against **75.7%** for the
bare code — **32.1%** once the denominator drops the lines whose bare code already
overflows.

The doc keeps side-alignment on the record as a plausible later *style*, not as a
mistake. It is rejected as the one style that has to work everywhere, not outright.

### What stacking does not buy

Whole rendered block on those 2,842 lines, widening versus stacking:

| terminal | widen (today) | stacked |
| --- | ---: | ---: |
| 80 cols | 23.3% | **23.4%** |
| 100 cols | 48.3% | 48.3% |
| 120 cols | 65.9% | 65.9% |
| 160 cols | 84.9% | 84.9% |

That denominator is padded: a block is never narrower than its code line, so a line
whose code alone overflows cannot fit under any caret model. At 80 columns only **730
of the 2,842** could fit at all; among those, 90.5% → 91.1%. The doc now discloses this
rather than quoting the padded rate bare.

Terminal fit is **unchanged**. On these dense lines the block width is set by the code
line and the wrapped detail rows, which both models share. Nor does the block shrink: same width on
**91.8%**, narrower on **1.1%**, wider on **7.2%** — but all three shares are padded,
and the doc now gives the partition instead: of **2,612** blocks pinned at the code
width, 2,608 are unchanged and 4 grow; of the **230** with headroom, **0** are unchanged,
**30** narrow and **200** widen. Stacking buys attribution and
essentially nothing else, and the doc says exactly that — an earlier revision claimed
it strictly improved fit, and the revision after that overcorrected into a categorical
denial.

## Measurements

CoreLib `11.0.0-preview.7.26366.102`, annotated-source render (the one `--focus`
produces), spans computed after `PrintRaised` with a fresh import per render, **summed
over the five focus families with the focus filter applied before counting**. That last
condition was the single largest correction in review: `CollectFacts` returns every
fact, but `--focus` promotes only one family, so a pre-filter count describes a render
no invocation produces.

| | |
| --- | ---: |
| caret-bearing lines | 32,864 |
| …carrying at least one extent | 29,933 |
| …with a single distinct extent | 27,345 (91.4%) |
| …with two or more distinct extents | 2,588 (8.6%) |
| worst line | 8 distinct extents |
| mixed lines | 355 |
| …with exactly one surviving extent | 254 |

**30,022 of the 32,864 caret-bearing lines (91.4%) keep exactly today's geometry** —
the complement of the 2,842 that stack. It splits into **27,091** lines that narrow to
one agreed extent and **2,931** with no extent at all, which widen as they always did.
An earlier revision quoted the 27,091 alone (82.4%) as the unchanged share; review
caught that it drops the whole no-extent population. The claim is about *output*: a
no-extent line now runs through `Stack` and collects every fact as unplaced before the
widening branch takes over, so the bytes match but the path does not. Note also that the 27,091 is *not*
the 27,345 single-extent count: 254 of those are mixed lines whose geometry does
change.

Applying the rules to the resulting **2,842** stacking lines puts 2,543 (89.5%) on a
single row and none above four — though that rate is padded by 254 single-group lines
that cannot take two rows; among the 2,588 lines with something to pack it is **88.4%**.

Trail widths came back through review twice. 3,884 of 6,566 trails (59.2%) render at
true width, but that rate is **padded by a structural immunity**: the last trail on a
row has no successor, so its clip limit is `int.MaxValue` and it cannot be clipped.
Measured, not derived — those 2,842 lines occupy **3,144 rows** and all 3,144
last-on-row trails render at true width. Of the **3,422** trails actually exposed to a
successor, 740 (21.6%) survived — and five of *those* are immune as well, their extents
being no longer than `MinTrail`. Among the **3,417** trails that could genuinely be cut,
**735 (21.5%)** survived and **2,682 (78.5%)** were clipped. It took two passes to strip
both layers of padding off that one rate.

## Every figure is verbatim output

Each code block in the doc is a verbatim excerpt of `Annotated Source` output,
attributed to the member and focus family that produces it, with runnable
commands in a *Reproducing these figures* section. Two blocks are declared
mockups: the 374-column `Tuple.Equals` line is elided, and the side-aligned
block depicts a layout that was never implemented.

Three blocks were corrected after review. The `Vector.IsSubnormal` example was
substantively wrong — it claimed four caret rows with ~100- and ~84-column
trails, where the renderer emits two rows and clips the first nesting pair to
11 columns and a `~`. The `stackalloc` mixed-line pair had a hand-indented
gutter and the two facts in the opposite order to the render.

The `HebrewCalendar` block's carets were correct all along; only its code line
was under-indented, which made it *look* misaligned and led two reviewers to
report a caret defect that did not exist. The doc now names all three coordinate
systems — research render, `ApiOutputFormatter` projection, and the snippet —
so the next reader does not repeat that diagnosis.

## Validation

`npx markdownlint-cli docs/design/caret-stacking.md` — clean. Index entries added to
`docs/README.md` and `docs/overview.md`, both of which had described the
annotation as sitting on the same row as its caret — the side-aligned layout this
doc rejects.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -65,7 +65,7 @@ System.Text.Json.JsonSerializer (System.Text.Json 10.0.2)
 | [Section Model](design/section-model.md) | Section selection design notes; use with [Progressive Disclosure](design/progressive-disclosure.md). |
 | [Schema Query](design/schema-query.md) | `-D`/`-S` schema/query implementation notes. |
 | [Hidden-Fact Annotations](design/hidden-fact-annotations.md) | Allocation/unsafety/lifetime annotation model and the static IL pair-agreement oracle strategy. |
-| [Caret Stacking](design/caret-stacking.md) | `--focus` display model: one caret per fact extent, its text on the same row, uncapped alignment. |
+| [Caret Stacking](design/caret-stacking.md) | `--focus` display model: one caret per fact extent, packed onto as few rows as fit, with the numbered fact texts listed below. |
 | [Decompiler Inspection & Oracle](design/decompiler-inspection-oracle.md) | Unifies single-method inspection (dump/stages) with the corpus-wide fidelity check oracle; product-vs-tool scoping. |
 | [ReturnToSender: Fact-Planned Compile-Back Harness](design/fact-planned-compile-back-harness.md) | Spec for a fresh tools-side compile-back harness with fact-planned TypeProducer/TypePrinter shells. |
 | [Method Body Inspection](design/method-body-inspection.md) | Target service seam for shared `member` and `library --il-offset` method-body facts and coordinate inspection. |
--- a/docs/README.md
+++ b/docs/README.md
@@ -65,6 +65,7 @@ System.Text.Json.JsonSerializer (System.Text.Json 10.0.2)
 | [Section Model](design/section-model.md) | Section selection design notes; use with [Progressive Disclosure](design/progressive-disclosure.md). |
 | [Schema Query](design/schema-query.md) | `-D`/`-S` schema/query implementation notes. |
 | [Hidden-Fact Annotations](design/hidden-fact-annotations.md) | Allocation/unsafety/lifetime annotation model and the static IL pair-agreement oracle strategy. |
+| [Caret Stacking](design/caret-stacking.md) | `--focus` display model: one caret per fact extent, its text on the same row, uncapped alignment. |
 | [Decompiler Inspection & Oracle](design/decompiler-inspection-oracle.md) | Unifies single-method inspection (dump/stages) with the corpus-wide fidelity check oracle; product-vs-tool scoping. |
 | [ReturnToSender: Fact-Planned Compile-Back Harness](design/fact-planned-compile-back-harness.md) | Spec for a fresh tools-side compile-back harness with fact-planned TypeProducer/TypePrinter shells. |
 | [Method Body Inspection](design/method-body-inspection.md) | Target service seam for shared `member` and `library --il-offset` method-body facts and coordinate inspection. |
--- a/docs/design/caret-stacking.md
+++ b/docs/design/caret-stacking.md
@@ -286,12 +286,19 @@ line the moment any one label would start left of the gutter, so a broad focus
 can fall back to widening on account of a single early fact while a narrower
 focus, having dropped that fact, stacks the two later extents successfully.
 Naming the guard's rate against the lines that *stack* would be circular, since
-a line rule 8 rejects does not stack by construction. Measured against the set
-that **qualifies** to stack before rule 8 runs, the guard fires on **0 of
-2,842** — so all 2,842 qualifying lines do stack, no such line exists in this
-corpus, and the 2,842 does bound every narrower argument over CoreLib. That is a
-measurement, not a consequence of subsetting, and it would have to be
-re-measured on another assembly.
+a line rule 8 rejects does not stack by construction. The measurement is against
+the set that **qualifies** to stack before rule 8 runs, and a fallback is
+detected by the shape of the row: a stacked row always carries an `N.` label
+immediately before its caret run, and the widening row never does. Detecting it
+as "the block has no caret row" would be vacuous, because a fallback still
+renders the widening caret.
+
+So counted, the guard fires on **0 of 2,842** qualifying lines: all 2,842 do
+stack. That the gate can report a non-zero was checked by mutation — raising the
+guard's margin from `commentColumn + 2` to `+ 6` makes it fire on 306 of the
+2,842, and to `+ 200` on all 2,842. The zero is a property of the corpus, not of
+the probe. It is still a measurement rather than a consequence of subsetting,
+and it would have to be re-measured on another assembly.
 
 The bound is also specific to the **stacks** column. The other columns are not
 monotone in the same direction: dropping one of two disagreeing extents turns a
@@ -340,7 +347,7 @@ exactly where extents nest. The rule 8 gutter fallback fires on **0** of the
 ### Reproducing these figures
 
 The specification above is implemented in `AnnotationCaret` and shipped in
-[#3656](https://github.com/richlander/dotnet-inspect/pull/3656) at `a14d3ddce`.
+[#3656](https://github.com/richlander/dotnet-inspect/pull/3656) at `40bb4411d`.
 
 Every code block in this document is a verbatim excerpt of `Annotated Source`
 output for the member and focus family named beside it:
--- a/docs/design/caret-stacking.md
+++ b/docs/design/caret-stacking.md
@@ -12,6 +12,11 @@ which characters a fact is about — is described in
 per-fact extents already exist and are already correct; today's renderer
 discards them.
 
+Every figure in this document comes from the corpus named under
+[Measurements](#measurements): `System.Private.CoreLib`
+`11.0.0-preview.7.26366.102`, rendered by the annotated-source view. Extents
+are measured in printed characters, so the render is part of the figure.
+
 ## The problem
 
 `AnnotationCaret.Agreed` returns an extent only when every fact on the line
@@ -21,7 +26,7 @@ several different sub-expressions, that produces a caret that points at
 everything and therefore at nothing, above a stack of details with no way to
 tell which detail belongs to which expression.
 
-`System.Tuple<…>.Equals` is the worst real case in CoreLib — 16 facts, one
+`System.Tuple<…>.Equals` is the worst real case in the corpus — 16 facts, one
 330-column caret:
 
 ```csharp
@@ -141,8 +146,9 @@ the bare code.
 ### Mixed lines
 
 A **mixed** line carries facts of both kinds: some with an extent, some
-without. There are 615 of them in CoreLib, and 499 have exactly one surviving
-extent, so this is not a corner.
+without. There are 615 of them in the corpus, and 499 have exactly one
+surviving extent, so this is not a corner. (The overlay render is a different
+population: 609 and 498.)
 
 Today they widen. The reasoning was sound while a caret was the only signal
 available: narrowing to the one surviving extent would underline an expression
--- a/docs/design/caret-stacking.md
+++ b/docs/design/caret-stacking.md
@@ -188,9 +188,14 @@ trimmed statement, which ends inside the line. The widening render has no
 explicit bound, but it does not need one to stay within the line.
 
 What differs is the rendered **row**. Widening appends the first detail string
-to the caret row when it fits the inline budget, and on **20 lines (0.70%)**
-that appended text carries the row past the end of the code line. Stacking never
-appends detail to a caret row, so it is 0. The contrast is about where detail is
+to the caret row when it fits the inline budget, and that appended text carries
+the row past the end of the code line. Quoting this as **20 of 2,842 lines
+(0.70%)** pads the denominator with the 2,822 whose detail never goes inline and
+which therefore cannot overhang at all. Inline detail appears on exactly **20**
+of these lines, and **all 20** overhang — the rate is **20/20**, and it is
+structural rather than lucky: widening underlines the whole trimmed statement,
+so the caret ends where the code line ends and anything appended after it is
+past the end. Stacking never appends detail to a caret row, so it is 0. The contrast is about where detail is
 placed, not about bounding the underline — and the earlier claim that the
 widening underline "shifts rightward while preserving the full extent length"
 described a mechanism that does not occur.
@@ -437,7 +442,7 @@ fires on **0** of the 2,842 lines qualifying to stack before it runs.
 ### Reproducing these figures
 
 The specification above is implemented in `AnnotationCaret` and shipped in
-[#3656](https://github.com/richlander/dotnet-inspect/pull/3656) at `89c56d1bd`.
+[#3656](https://github.com/richlander/dotnet-inspect/pull/3656) at `102c73f8f`.
 
 Every code block in this document is a verbatim excerpt of `Annotated Source`
 output for the member and focus family named beside it:
--- a/docs/design/caret-stacking.md
+++ b/docs/design/caret-stacking.md
@@ -81,12 +81,13 @@ depicts a layout that was never built. This one is
    for its 16 facts.
 2. **Order by start column, widest first at a tie.** Extents sharing a start
    column are always a nesting, so they cannot share a row; widest-first puts
-   the outer one first, so the inner caret is clipped into its parent's trail
-   rather than overwriting it, and it matches the order the printer records
-   nesting in. This orders *same-start* extents only and says nothing about row
-   widths overall: a later disjoint extent can be packed onto a lower row and
-   reach further right than anything above it, which happens on **50 of the
-   2,842** lines.
+   the outer one on the upper row, which matches the order the printer records
+   nesting in. This orders *same-start* extents only, and it is not what causes
+   clipping — a trail is only ever cut short by the next label **on its own
+   row** (rule 5), and a nested extent that lands on a different row is not on
+   its parent's row to cut anything. Nor does it order row widths overall: a
+   later disjoint extent can be packed onto a lower row and reach further right
+   than anything above it, which happens on **50 of the 2,842** lines.
 3. **Number in that order,** `1.` upward, and label each caret with its number
    immediately before the trail.
 4. **Pack greedily onto rows.** An extent joins the first row where the
@@ -292,7 +293,9 @@ equal here. On an assembly that produced them, `semantics` would be a strict
 superset of `safety` and would need measuring separately.
 
 Any other argument is a narrower id prefix, such as `alloc.box`, and selects a
-strict subset of one family. Dropping facts can only remove distinct extents and
+subset of one family — not necessarily a *strict* one, since `safety.callee` is
+the only `safety.*` descriptor CoreLib produces and so selects exactly what
+`safety` does. Dropping facts can only remove distinct extents and
 remove extent-less facts, and a line qualifies to stack only when it has two of
 the former or one of each — so the *qualifying* set shrinks under a subset.
 
@@ -356,14 +359,17 @@ Applying the specification to the 2,842 lines that stack:
 | 3 | 1 | 0.0% |
 | 4 | 1 | 0.0% |
 
-3,884 of 6,566 trails (59.2%) render at true width. Clipping concentrates
-exactly where extents nest. The rule 8 gutter fallback fires on **0** of the
-2,842 lines qualifying to stack before it runs.
+3,884 of 6,566 trails (59.2%) render at true width. The other 2,682 are cut
+short by the next label on their own row, which is the only thing that clips a
+trail. That successor is nested inside the trail it clips on **1,867 (69.6%)**
+of them and wholly disjoint from it on **815 (30.4%)**, so clipping is
+concentrated at nestings but is not confined to them. The rule 8 gutter fallback
+fires on **0** of the 2,842 lines qualifying to stack before it runs.
 
 ### Reproducing these figures
 
 The specification above is implemented in `AnnotationCaret` and shipped in
-[#3656](https://github.com/richlander/dotnet-inspect/pull/3656) at `3d0e2afb0`.
+[#3656](https://github.com/richlander/dotnet-inspect/pull/3656) at `8fa2bf44a`.
 
 Every code block in this document is a verbatim excerpt of `Annotated Source`
 output for the member and focus family named beside it:
--- a/docs/design/caret-stacking.md
+++ b/docs/design/caret-stacking.md
@@ -80,9 +80,13 @@ depicts a layout that was never built. This one is
    together under it. Under `--focus alloc`, `Tuple.Equals` renders 8 carets
    for its 16 facts.
 2. **Order by start column, widest first at a tie.** Extents sharing a start
-   column are always a nesting, so they cannot share a row; widest-first makes
-   each row narrower than the one above, and matches the order the printer
-   records nesting in.
+   column are always a nesting, so they cannot share a row; widest-first puts
+   the outer one first, so the inner caret is clipped into its parent's trail
+   rather than overwriting it, and it matches the order the printer records
+   nesting in. This orders *same-start* extents only and says nothing about row
+   widths overall: a later disjoint extent can be packed onto a lower row and
+   reach further right than anything above it, which happens on **50 of the
+   2,842** lines.
 3. **Number in that order,** `1.` upward, and label each caret with its number
    immediately before the trail.
 4. **Pack greedily onto rows.** An extent joins the first row where the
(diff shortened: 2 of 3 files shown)


richlander:
## Round 2 review — both findings confirmed, and they led somewhere bigger

Both reviewers returned BLOCKING with the same two findings. Both were correct, and chasing the first uncovered a defect neither reviewer caught and that invalidated **every figure in the document**.

### The real defect

`--focus` does not render every collected fact. `AnnotationGestureSelector.Focus` promotes only the requested family to a caret; `SplitByGesture` leaves the rest as side comments, so they never reach `AnnotationCaret`. My corpus probes passed the **full** fact list to `Render`. Every published number therefore described a render no invocation can produce.

Re-measured after the focus filter, summed over the five families:

| figure | was | is |
| --- | ---: | ---: |
| lines this model changes | 3,385 | **2,842** |
| mixed lines | 615 | **355** |
| old render's over-wide caret rows | 137 (4.05%) | **20 (0.70%)** |
| single row | 88.6% | **89.5%** |
| trails at true width | 56.5% | **59.2%** |
| side-alignment overflow | 97.28%, mean 81 cols | **96.61%, mean 77 cols** |

Mixed lines are not spread evenly: 298 `alloc`, 47 `safety`, 10 `cost`, **0** `unsafe`, **0** `lifetime`.

**One claim reversed.** The old figures said stacking made the rendered block 91.0% narrower. On the real population it is the **same width on 91.8%** of lines, narrower on 1.1%, wider on 7.2%. Terminal fit is still unchanged (23.3% → 23.4% at 80 columns). The document now says plainly that stacking buys attribution and **nothing else**.

**`--focus lifetime` never stacks.** No line in CoreLib carries two lifetime facts that disagree about the extent. A figure summed over all families at once hid that completely — the clearest argument for why the filter has to come first.

### The reported findings

**@gpt — `--focus lifetime` cannot produce the mixed-line example.** Confirmed. **@gemini** added the mechanism: `unsafe.stackalloc` is an *instruction* annotation printed inline, so that line carries a single caret fact and takes the single-extent shortcut. Replaced with `MemoryExtensions.ThrowNullLowHighInclusive` under `--focus alloc` — a genuine post-filter mixed line. Today's render captured from the CLI built at base `f26fa8e0a`; the stacked render from the CLI at the implementation commit.

**Both — the reproduction commands fail.** Confirmed. Without `--platform System.Private.CoreLib` the internal types do not resolve, and the flag also loads the PDB, so locals print as `tk`/`buffer` rather than `V_0`. The `ResolveSignature` block came from a probe that opened the DLL directly; it is now verbatim CLI output with 14-column carets.

### On the round-1 finding

Round 1's "misaligned carets" was refuted by measurement and remains refuted — the carets were always correct; the code line was under-indented. But that refutation is *not* what made the examples wrong. Two of them were wrong for reasons neither round found until now. Being right that a reviewer's diagnosis was mistaken is not evidence that the artifact is sound.

### Also corrected

- `Tuple.Equals` restated from the CLI: a 374-column line under a single 370-column caret above 106 wrapped detail rows, rendering as 8 numbered carets plus 8 `-` rows.
- Every code block re-verified line-for-line against CLI output. Five verbatim; the two that are not say so where they appear.
- The implementation PR #3656 carried the same wrong-population figures in its XML remarks. Corrected there too (comments only, `da519f697`), and both reviewers re-tasked at that fixed head.

CLI suite at `da519f697`: 2,766 total, 1 failed, 14 skipped — the pre-existing `Layout_Count` failure, identical on base.

Round 3 review is running at `5ab572da1`.

richlander:
## Round 3 — three defects found, all confirmed by measurement, all fixed

Head is now `4f177decf`.

**The `Tuple.Equals` detail-row count was wrong.** The document said 106 wrapped detail rows. That count had included the `IL_xxxx:` rows, which are also `//` comment lines. Recaptured from the CLI built at the base commit `f26fa8e0a`, counting only the annotation rows that precede the first IL row:

```
code cols: 374
caret rows: 1   caret run: 370
detail rows (excluding IL): 32
fact starts: 16
```

**A stale `330` survived** in the elision note. The line is **374** columns; the caret is **370**. The same stale figure had leaked into #3656's XML remarks and both PR bodies, and is corrected in all four places.

**The fit claim was categorical where the data is not.** "Does not improve fit, does not narrow the block" sat directly above measurements showing 1.1% of blocks getting narrower and 80-column fit moving 23.3% → 23.4%. Those are rounding-scale movements against a 7.2% share that gets wider, but they are not zero. Restated as no *material aggregate* improvement, naming both movements.

**`--focus semantics` is not an alias of `--focus safety` by construction.** The `Semantics` category declares `semantics.callee` as well as `safety.callee`, so `--focus semantics` selects the union of the two. The measured equality — both 17,041 — holds because CoreLib yields **zero** `semantics.callee` facts. That is a property of this corpus, not of the categories, and the document now says so.

### Also corrected in #3656

Two comment sites there still quoted the **pre-filter** population — 615 mixed / 499 single-extent, plus an overlay parenthetical — in `AnnotationCaret.Stack`'s remarks and in the PARTIAL-case test comment. `--focus` promotes only one family to carets, so those described a render no invocation produces. Post-filter, summed over the five families: **355** mixed, **254** with a single surviving extent. Comments and XML docs only; `AnnotationGestureTests` 30/30 green, build clean.

Both reviewers have been re-tasked at `4f177decf`, and the approving reviewer's round-3 APPROVE is treated as withdrawn — two of the three defects were in figures it reported having recomputed.

richlander:
## Rounds 4–6 — the gutter-fallback gate was vacuous

Head is now `ae9aeb503`; the implementation pin is `40bb4411d`.

### The serious one: a measurement that could not fail

Rule 8's `0 of 2,842` rested on a probe that detected a fallback as **a rendered block with no caret row**. Falling back still renders the widening caret, so that test could never fire. The zero was unfalsifiable.

Re-measured with a detector that separates the two shapes — a stacked row always carries an `N.` label immediately before its caret run, a widening row never does — the figure is **unchanged at 0**, and the row histogram is unchanged at 2543/297/1/1. The numbers were right; the probe had not established them.

Non-vacuity then proved by mutating only the guard margin:

| guard margin | fallbacks |
| --- | ---: |
| `commentColumn + 2` (real) | **0** of 2,842 |
| `commentColumn + 6` | **306** of 2,842 |
| `commentColumn + 200` | **2,842** of 2,842 |

Graded rather than a binary flip, so the zero is a property of the corpus, not of the measurement.

### Rule 8 is not monotone, so the upper-bound claim was wrong both ways

`RenderStacked` returns `null` for the **whole line** on the first gutter-overlapping label, so a broad focus can fall back to widening because of one early fact while a narrower focus, having dropped it, stacks the survivors. Subsetting therefore does not imply the bound. And in the other direction, `single extent` can *rise* under a narrower focus, since dropping one of two disagreeing extents converts a multi-extent line into a single-extent one. The bound now applies only to the `stacks` column, and is stated as a corpus measurement rather than a consequence.

### The guard rate had a circular denominator

A line rule 8 rejects does not stack by construction, so quoting the rate against *stacking* lines was circular. The probe had in fact measured against the set qualifying **before** rule 8, and the prose now says so.

### `Agreed()` still carried all-facts figures

Re-measured post-focus over the five families:

```
agreed:    27091 (82.43%)    facts=1:  29550 → 91.7% agree
disagree:   2842 (8.65%)     facts=2:   2282 → 0.0%
no extent:  2931 (8.92%)     facts=3:    714 → 0.0%
null lines: 5773 (2459 single-fact, no extent)   facts=4+:  318 → 0.0%
```

The *shape* changed, not just the digits: the old remark had agreement **declining** with density (12.1% at two facts); post-focus it **vanishes** — two facts of one family never share an extent here. Verified not to be an artifact of `Agreed`'s bounds test by counting all-equal multi-fact lines before bounds: **0**. Two figures tie as they must — 32,864 matches the Measurements table and 27,091 is exactly the "keeps today's geometry" count.

### Corroboration

The other reviewer's independent census, written separately, reproduced the corpus figures exactly: `stacked=2842; rows=[1:2543, 2:297, 3:1, 4:1]; true-width=3884/6566 (59.2%)` over 41,952 method bodies.

Every change to #3656 since its last approved head is **comments and XML docs only** (`git diff -U0 da519f697..40bb4411d -- '*.cs'` filtered to non-comment lines is empty); `AnnotationGestureTests` 30/30, solution builds clean.

richlander:
## Rounds 7–10 — a third vacuous measurement, and two claims that were simply false

Four more rounds against fixed heads. Seven findings, all confirmed by measurement, all fixed. The pattern from rounds 4–6 held: every defect was an assertion about behaviour that had never been tested against the code implementing it.

### Round 7 — the third vacuous measurement

The reviewer that had approved this document three times was re-tasked to *falsify* rather than verify. It then found what the approvals had missed:

> "no stacked caret row is ever wider than the code line it annotates"

is **structurally guaranteed**, not measured. `Stack` sends any extent with `Column + Length > lineLength` to `unplaced`; labels grow leftward from their extent; rule 8 bails rather than shifting a label right. No corpus could produce a counter-example, so the corpus sweep that "confirmed" it was unfalsifiable.

Restated as structural in both PRs. The informative number is the *contrast*, and it is measured: today's widening render overhangs the code line on **20 lines (0.70%)**, because

```csharp
pad = Math.Max(1, caretColumn - commentColumn - 2)
```

shifts the start column rightward while keeping the full extent length. That figure can be falsified; the old one could not.

That reviewer's earlier census had also disagreed with mine (2,525 vs 2,842) and it had attributed the gap to "environment differences". That explanation was rejected and reconciliation demanded. It reconciled to **2,842** — its own count came from a flawed manual regex script, not the environment.

### Round 8 — widest-first does not make each row narrower than the last

The document claimed the widest-first row assignment produces rows of monotonically decreasing width. A counterexample was supplied. Rather than concede on the example, it was measured across the corpus: **50 of 2,842 stacked lines have a lower row wider than the row above it.** The claim is wrong, and it is wrong at a rate worth stating rather than hand-waving.

Also in this round: the circular denominator corrected in round 5 had survived in two summary paragraphs (doc lines 112 and 354).

### Round 9 — clipping is not caused by nesting

A mechanism error, not an arithmetic one. The document attributed a shortened caret trail to a *nested* extent. The code says otherwise:

```csharp
int limit = position + 1 < row.Count
    ? LabelStart(row[position + 1]) - TrailGap - Column(index)
    : int.MaxValue;
int width = Math.Min(length, limit);
```

Only the next label **on the same row** clips. Two extents with the same start column can never share a row, so a nested extent is never positioned to clip its parent. The real split of the 2,682 clipped trails: **1,867 (69.6%) nested successor / 815 (30.4%) disjoint successor.** Replication ties exactly — 6,566 − 3,884 = 2,682.

Second finding: "a narrower id prefix selects a strict subset" is false. `safety.callee` is the only `safety.*` descriptor, so `--focus safety.callee` selects exactly what `--focus safety` selects. Subset, not strict subset.

### Round 10 — precision, and the inverse audit

One finding: the implementation remark said the **label** doing the clipping sits inside the trail, where the document correctly says the successor **extent**. Not equivalent — a disjoint successor's label reaches two columns left of its own extent, so it can clip a trail it does not overlap. Fixed in the source.

The reviewer was then asked for the inverse pass — claims stated as *contingent* that are in fact *structural*, the mirror of the round 7 defect. It returned **no remaining errors** in this document.

### Standing rule adopted from this run

For every quantitative claim: *what observation would falsify this, and could the cited evidence have produced it?* Where the claim guards a gate, prove the gate non-vacuous by mutation. The rule 8 guard was proven that way — margin `+2` → 0 firings, `+6` → 306, `+200` → 2,842. Graded, not binary, which is what a real gate looks like.

Both PRs are green. Cross-referenced implementation: #3656.

richlander:
## Round 13 — falsification pass, three findings, all confirmed by measurement

A hostile falsification pass found three defects. I measured each independently rather than accepting the report; all three held, and one of them was an error I introduced myself in an earlier round.

### 1. The Tuple caret width is 370, not 330 — and I broke it

This figure has now been wrong in **both** directions. It was originally recorded as 370, an earlier round of this review "corrected" it to 330 on a recapture of the wrong render, and this round measured it back to 370.

Rendered through the documented command and measured off the string:

| quantity | value |
| --- | ---: |
| trimmed statement | **370** cols |
| full line after body indent | **374** |
| base widening caret (`fbabdf6d0`) | **370** glyphs, cols 5–374 |
| stacked caret row (head) | 127 glyphs, ends col **372** |

The neighbouring claims all held on re-measurement: 16 facts, 32 wrapped detail rows in *both* renders, 8 numbered labels + 8 `-` markers. I added that the block does not grow — one caret row above the same 32 detail rows either way.

**Why it survived:** nothing else in the document depended on it, so 330 stayed internally consistent while being false. Both the doc and the `Stack` remark now carry a standing instruction to re-derive this from the render, never from agreement with the surrounding prose. Internal consistency is not evidence.

### 2. The terminal-fit denominator was padded — the fifth instance

A block can never be narrower than its code line, so a line whose code alone overflows cannot fit under *any* caret model. Those lines were in the denominator.

| terminal | code alone fits | of those: widen | of those: stacked |
| --- | ---: | ---: | ---: |
| 80 | **730 of 2,842** (25.7%) | 90.5% | 91.1% |
| 100 | 1,373 (48.3%) | 99.9% | 99.9% |
| 120 | 1,875 (66.0%) | 99.9% | 99.9% |
| 160 | 2,414 (84.9%) | 100.0% | 100.0% |

At 80 columns **2,112 of the 2,842 are structurally incapable of the outcome being counted**. Every previously published figure reproduced exactly, so the *comparison* was sound — it is a paired A/B and both columns shared the denominator — but the absolute rate was not meaningful as stated.

The partition also produced a sharper result than the old prose had: at 100/120/160 the code-fits share and the fit rate converge to within a tenth of a point, which says the block width simply **is** the code width there.

### 3. "Keeps today's geometry" was a claim about output stated as a claim about the path

Traced through `Render`: a no-extent line now enters `Stack`, walks every fact, fails `TryGetValue` for each, populates `unplaced`, returns an empty group list, fails the `{ Count: > 0 }` guard, and only then runs the widening branch. Same bytes, different work. The document now says so.

### Running tally

Fifteen defects across thirteen rounds. Two recurring families, both now with named countermeasures in the text:

- **Padded denominator** — five instances. Every rate must name a denominator whose members are all *capable* of the numerator's outcome.
- **Broken by correction** — two instances, and the second was a re-break of the first. Verify a revised figure against the artifact, never against the document.

Design doc `e6af0ac07`; implementation repinned to `89c56d1bd`.

richlander:
**Round 18 — the presentation was the defect, not the individual rates.**

Three consecutive rounds each found a *different* one of the three block-width shares padded (91.8% unchanged / 1.1% narrower / 7.2% wider). That pattern was the finding. Rather than unpad a fourth time, the shares are gone, replaced by a partition on the one structural floor that governs them — a block is never narrower than its code line:

| population | unchanged | narrower | wider |
| --- | ---: | ---: | ---: |
| **2,612** pinned at the code width | 2,608 | — | 4 |
| **230** with headroom above it | 0 | 30 | 200 |

This says more than any of the three rates did. Nothing happens on the pinned majority; *every* line with room to move moves, and 200 of the 230 move the wrong way. Stacking buys attribution and costs a little width — that is the honest summary, and it is now the one in the document.

It also refutes a correction I made two rounds ago. I had argued growth has no structural floor, so "any block can grow". It does have one: only 4 of the 2,612 pinned blocks grow. A review round turned a right figure into a wrong one, for the second time in this review.

**Two more corrections this round.**

*The side-align maximum.* The document asserted an annotation column of 57,942 under a gap convention it never stated. Measured, the extent ends at column 57,940 and the line is 57,940 characters; text one column past the caret starts at 57,941, two columns past at 57,942. Both numbers are right under different styles, so the document now states the fact that is not convention-dependent — the line length — and derives the rest.

*The blast-radius denominator, defended rather than conditioned.* A review round called the 8.6% change rate a tenth padded denominator, on the grounds that 27,091 single-agreed and 2,931 no-extent lines cannot change. They cannot, but that denominator is the whole shipped population **on purpose**: the claim is how much existing output this PR alters, not a rate of an outcome among cases capable of it. Conditioning it would drive it to 100% and measure nothing.

The finding did expose something better, though. 27,091 + 2,931 is exactly 30,022, so the 2,842 changed lines are precisely the complement — and `Agreed()` returns null on exactly the lines this model treats differently (any fact missing an extent, or two extents disagreeing). So **every line that can stack does**, by construction rather than coincidence. That is a mechanism, and it was derivable from arithmetic already printed in the same paragraph. That is now the fourth defect in this review recovered from text the document was already carrying.

**Scorecard for the last round:** three findings right and decisive, two half right, one wrong and contradicting the reviewer's own earlier verification. The claim that the 10.5% figure was cross-contaminated was itself an error — it reproduces exactly at 3,131, and inferring contamination from a numeric coincidence is the same reasoning mistake as trusting a figure for being internally consistent. The companion claim was right: the wrap rate is 71.9%, not "three lines in four".

Design head `e164dcbd9`. Implementation `721adb61a`, CI fully green (`decompiler-gates` 13m21s, `test` 10m32s).

richlander:
**Rounds 19–20 — the review found a bug in the measurement apparatus itself, which is the best thing it has done.**

Every side-alignment figure in this document was wrong, and not for any of the reasons the previous eighteen rounds were about. `sidealign3.cs` computed the annotation column *inside the code text* and compared it against a code width that included `BodyIndentWidth`. Two coordinate systems on either side of the same comparison. The rendered column of code character *i* is `BodyIndentWidth + i` (`AnnotationCaret.cs:156,458`), so the indent belongs on both sides or neither.

| figure | published | corrected |
| --- | ---: | ---: |
| overflows the code line | 96.61% | **97.52%** |
| mean overhang | 77 | **81**, over the 29,191 that overflow |
| fits 80 columns | 28.1% | **24.7%** |
| conditioned on bare-fitting lines | 36.4% | **32.1%** |
| maximum annotation column | 57,942 | **57,946** |

The reviewer's one-column-gap re-measurement and mine agree to the digit (97.20%, 81, 25.3%, 33.0%, 11.8%, 87.2%, 57,945), which is the check that mattered — a corrected figure I could not independently reproduce would have gone the way of the withdrawn 87.5%.

I published the **two-column** set, because the sketch in that section shows two spaces and the sketch is the specification of the style. A document should not depict one thing and measure another. The one-column set is given alongside so that the gap's irrelevance to the rejection is visible rather than asserted.

The root cause was correctly diagnosed as the *absence of a stated convention*, not the arithmetic. That section now opens by naming it outright — final rendered columns, one row per fact, two-column gap, population 29,933 — and the same figures appearing a second time earlier in the document, plus both PR bodies, carried the identical bug and are now consistent with it.

**Two more defects in the same rounds.**

*Eleventh padded denominator.* The annotation-column-over-100 rate counted 25,761 lines too short to reach column 100 under any style. Conditioned on the **4,172** that can, it is **87.6%**. Note the direction: unpadding this one made the rejected alternative look *worse*, not better. The 12.2% had been understating the objection. Unpadding is not an argument in a fixed direction, which is the point of doing it by rule rather than by taste.

*A correction I introduced in round 18.* I wrote that `Agreed` selects the stacked lines "by construction". It does not: `Agreed`-null is a superset at 5,773, the selector is `Stack(...) is { Count: > 0 }`, and the 2,931 no-extent lines are the difference. The bounds check makes the identity empirical — it rejects 0 lines measured. The XML remark on `Agreed` already said *"100% as measured rather than 100% by construction"* about a neighbouring rate, so the document had contradicted a correction the code was already carrying. Fifth defect in this review whose answer was already sitting in the repository.

**Tally across twenty rounds:** eleven padded denominators, three figures broken *by* a correction, two probes of mine measuring the wrong thing, one measured in the wrong coordinate system, one claim about an unbuilt alternative withdrawn for having no probe at all, and five defects recoverable from text the repository already contained.

Design head `40b44ee10`, **approved**. Implementation `721adb61a`, CI fully green.

Issue #3628, opened 2026-07-31
Metadata table cells escape C0 controls but not bidi overrides, leaking raw U+202E to the terminal


richlander:
## Summary

`@Metadata` table cells escape C0/DEL/C1 controls but not **Cf bidi overrides** or the **Zl/Zp separators**, so a hostile name in a `.dll` reaches the terminal as raw `U+202E` bytes on stdout. This is the #3319 class, on a path #3380 did not cover.

Reproduced on merged `main` (the metadata rendering path is byte-identical between `af6cb96c9` and the tree I measured).

## Reproduction

Build an assembly whose type, field, and method names each carry `U+202E`:

```csharp
var t = m.DefineType("N.Hostile\u202EINJECTED", TypeAttributes.Public | TypeAttributes.Class);
t.DefineField("field\u202EINJECTED", typeof(int), FieldAttributes.Public);
```

```console
$ dotnet-inspect library MetaHostile.dll -S "Metadata: TypeDef" --tips q | hexdump -C
00000010  e2 80 ae 42 20 7c 20 4e  20 7c 20 54 79 70 65 52  |...B | N | TypeR|
```

`e2 80 ae` is the raw override, on **stdout**.

Lines carrying raw `U+202E`, same assembly:

| Section | Markdown | TSV | JSON |
| --- | --- | --- | --- |
| `Metadata: TypeDef` | 1 | 1 | 0 |
| `Metadata: Field` | 1 | — | — |
| `Metadata: MethodDef` | 1 | — | — |
| `Metadata: #Strings` | 3 | — | — |

JSON is clean because the serializer escapes non-ASCII. The human- and terminal-facing modes are the ones that leak.

**Control:** the same assembly through the C# lens that #3380 hardened is clean — `dotnet-inspect type … --library MetaHostile.dll` yields **0**. The leak is specific to the metadata lens.

## Cause

`MetadataTableProjector` escapes cell text with a hand-rolled predicate:

```csharp
// src/ILInspector.Metadata/MetadataTableProjector.cs:1439
static bool IsControl(char c) => c < ' ' || c == '\x7f' || (c >= '\x80' && c <= '\x9f');
```

That is C0 + DEL + C1. `U+202E` is category **Cf**, and `U+2028`/`U+2029` are **Zl/Zp**; none are `< ' '`, so all three fall through to the `_ => 1` width and are appended unmodified.

[...]

PR #3660, opened 2026-08-01
Bound an InertString to a budget without dividing a spelling

richlander:
Adds the two members that let a sink bound an `InertString` to a length budget, stacked on #3636.

## Why this is not a substring

A spelling is between two and ten characters wide — `\\`, `\^X`, `\uXXXX`, `\UXXXXXXXX` — so an arbitrary cut can land inside one and leave `\u2` behind.

That is unsound rather than merely ugly. `EnsurePermitted` treats text it cannot decode as raw and re-encodes it, so the surviving backslash doubles and `a\u2` becomes `a\\u2` — which is exactly what the untreated literal `a\u2` encodes to. Two unrelated inputs converge on one output, and the injectivity the repair path rests on is gone. That is the same convergence failure fixed during review on #3636, arriving through a new door.

Measured on encoded `a\u202Eb\\c` (length 11): **6 of the 12 cut points are undecodable.**

## Surface

| Member | Contract |
| ------ | -------- |
| `int Length` | encoded length — what a sink emits, not what it was given |
| `int IndexOfFirstEncoded()` | position of the first spelling, `-1` when there is none |
| `InertString Truncate(int maxLength)` | the longest prefix within the budget |
| `InertString Truncate(Range)` | the largest whole window inside the range |

Both are best-effort and total: every range is answerable, including a reversed one and one that runs off either end.

Both bounds move **inward** — the start forward to the next whole spelling, the end back to the previous one — so the result is always a subset of what was asked for. That asymmetry is the whole of the safety argument: returning less than a caller asked for is something it can detect from `Length`, while returning more is not. A window holding no whole spelling is empty rather than widened to the spelling enclosing it.

Neither reports truncation separately, because `result.Length` against the width asked for already says whether anything moved.

**There is no exact counterpart that throws.** An earlier revision had `this[Range]` refuse an unusable window. It wore the syntax of ordinary slicing while throwing on bounds that look perfectly reasonable — 6 of the 12 positions above divide a spelling — so the shape invited exactly the call it refuses, and it bought nothing over the length comparison.

```csharp
InertString cell = inert.Truncate(options.MaxStringChars);
bool truncated = cell.Length < inert.Length;
```

## Details worth review

- **A surrogate pair is one thing however it is spelled.** Raw, as `\U`, or as the two `\uXXXX` escapes composition produces when the halves are encoded in separate fragments — `Join` of `"\uD83D"` and `"\uDE00"` gives `\uD83D\uDE00`, and cutting between them would leave a lone surrogate in the decoded text. That atomicity is a guarantee the escaper this replaces documented, so dropping it would be a regression.
- **`Forms` is recomputed from what is kept**, so a legend drawn from a bounded value cannot name a spelling that went with the tail.
- **The token walker lives beside `AppendSpelling`**, because its widths are that method's output read backwards; a new spelling added to one and not the other would make every boundary past it wrong. It is `internal` and adds nothing to the capability surface: it reports where text divides, never what it decodes to.
- **The walker is total**, so no caller clamps. Mutation testing found the first version's reversed-range guard and negative-budget clamp were both dead — the end walks from the start and only advances, so each already gave an empty window. A line no mutation can falsify is a line nothing holds true, so both are gone.
- **Escape widths are clamped, not trusted.** `Encode` cannot emit a truncated escape, but the internal constructor asserts the invariant rather than establishing it, so an unclamped walk would surface a malformed value as an `IndexOutOfRangeException` at a boundary check instead of where it was built.

## Naming

`Slice(int)` was considered and rejected: `Span<T>.Slice(int start)` "forms a slice … that begins at a specified index", so a one-`int` `Slice` means a **start** everywhere in .NET. `inert.Slice(1000)` would read as "skip 1000 characters" while meaning "keep at most 1000" — both compile, opposite contents.

## Testing

330 tests, up from 190. The central property is not "the result is short enough" but that the result is a **fixpoint of `EnsurePermitted`** — the same statement the encoder's injectivity rests on — checked over the whole adversarial corpus at *every* budget from 0 past the end, together with the claim that the cut is maximal. Removing the indexer meant restating maximality without it: opening a window at a position is its own boundary oracle, since the start moves forward only when it must and so keeps the whole tail exactly when the position was already whole.

The range form adds a subset property swept over every `(start, end)` pair in the corpus: what comes back must appear inside the window that was asked for.

The walker restates `AppendSpelling`'s table backwards, and the boundary assertions only cover the widths half of it. The mapping from escape to `VisualForm` had no gate at all — three of the five forms could report `None`, and `CaretDelete` could collapse into `Caret`, with everything still green. That drift breaks no boundary; it makes the value describe itself wrongly. It is now pinned by decoding each window and encoding it again from scratch, which runs `AppendSpelling` and never consults the walker. `EnsurePermitted` cannot serve as that oracle: it short-circuits on an already-permitted value and hands back the value's own metadata.

Each behavior was mutation-tested; every mutation below turned the suite red.

Two of those sweeps could not see the shape they were meant to protect. Their inputs came from the adversarial corpus, whose values are all encoded directly, and direct encoding never emits the width-12 token: `Field` permits an astral graphic scalar, so it goes raw rather than as two `\uXXXX` escapes. Only `Join` composes that shape, so the sweeps now draw from a corpus of named `InertString` *values* that includes three composed ones.

Widening the inputs was necessary but not sufficient. Each half of a divided pair is itself a well-formed `\uXXXX` escape, so a cut between them still decodes, still satisfies `IsPermitted`, and is therefore still an `EnsurePermitted` fixpoint — every property the sweeps asserted lives on the **encoded** side, while the harm, handing a sink a lone surrogate, is only visible on the **decoded** side. The guarantee is now stated where it can fail: a bound may not fall between the halves of a surrogate pair of the decoded text. That forbids only creating a lone surrogate, not carrying one — the corpus holds two on purpose. The range sweep carries it too, since the near bound exists only there while the same token walk places both.

| Mutation | Caught by |
| -------- | --------- |
| escaped surrogate pair divisible | 7 tests |
| raw surrogate pair divisible | `Truncate_KeepsARawSurrogatePairWhole` |
| cut exactly at the limit | 25 tests |
| end bound unchecked | 23 tests |
| start snapped backward instead of forward | 23 tests |
| over-conservative by one token | 31 tests |
| start or end bounds check removed | 25 tests |
| full-window fast path always taken | 53 tests |
| `Forms` not recomputed | 3 tests |
| `IndexOfFirstEncoded` constant | 3 tests |
| `Backslash` reported as `None` | 1 test |
| `AstralHex` reported as `None` | 3 tests |
| `CaretDelete` collapsed into `Caret` | 1 test |
| `BmpHex` reported as `None` | 15 tests |
| `Caret` reported as `None` | 7 tests |

## Scope

No caller yet. These exist for the #3628 metadata migration, where `EscapeText(raw, max, out bool truncated)` becomes the two lines above; that migration is a separate PR.

### DIFF
--- a/docs/design/inert-text.md
+++ b/docs/design/inert-text.md
@@ -371,6 +371,50 @@ you are about to apply, short of sweeping every scalar to compare their
 permitted sets — so the value would carry an extra field and still pay for the
 scan.
 
+## Bounding a value to a budget
+
+A sink that puts treated text in a column or a record field has a length limit,
+and applying it to the encoded text is not a substring operation. A spelling is
+between two and ten characters wide, so an arbitrary cut can land inside one and
+leave `\u2` behind.
+
+That is not merely ugly, it is unsound. `EnsurePermitted` treats text it cannot
+decode as raw and re-encodes it, so the surviving backslash doubles and `a\u2`
+becomes `a\\u2` — which is exactly what the untreated literal `a\u2` encodes to.
+Two unrelated inputs converge on one output, and the injectivity the repair path
+rests on is gone. The budget is enforced on the *encoded* length for the same
+reason it is in the escaper this replaces: encoding expands, so bounding the
+input bounds the wrong number.
+
+Two members divide a value, and they differ in what they do with a bad cut:
+
+| Member | Bad cut | For |
+| ------ | ------- | --- |
+| `Truncate(int maxLength)` | moved down to the nearest whole spelling | a budget |
+| `this[Range]` | refused, with `ArgumentException` | a cut the caller already chose |
+
+Moving the cut is right for a budget, where the number is a limit rather than a
+claim, and refusing is right for a range, where the caller has named a span and
+quietly returning a different one is the worse answer. `Truncate` therefore
+reports no truncation flag: `result.Length < Length` already answers it, and a
+flag that restates a comparison is a second thing to keep true.
+
+Neither member moves a **start**. Moving an end down returns less than was
+asked for, which a caller can detect; moving a start down returns *more*, which
+is the one direction it cannot check.
+
+A surrogate pair counts as one thing however it is spelled — raw, as `\U`, or as
+the two `\uXXXX` escapes that composition produces when the halves are encoded in
+separate fragments — so a boundary can never leave a lone surrogate behind.
+`Forms` is recomputed from what is kept, so a legend drawn from a bounded value
+cannot name a spelling that went with the tail.
+
+The walker that finds these boundaries lives beside the speller rather than on
+the currency type, because its widths are the speller's output read backwards; a
+new spelling added to one and not the other would make every boundary past it
+wrong. It is internal, and it adds nothing to the capability surface: it reports
+where text can be divided, never what any of it decodes to.
+
 ## Testing
 
 Two shapes, because they fail differently.
--- a/docs/design/inert-text.md
+++ b/docs/design/inert-text.md
@@ -386,22 +386,30 @@ rests on is gone. The budget is enforced on the *encoded* length for the same
 reason it is in the escaper this replaces: encoding expands, so bounding the
 input bounds the wrong number.
 
-Two members divide a value, and they differ in what they do with a bad cut:
+Two members divide a value, and both are best-effort:
 
-| Member | Bad cut | For |
-| ------ | ------- | --- |
-| `Truncate(int maxLength)` | moved down to the nearest whole spelling | a budget |
-| `this[Range]` | refused, with `ArgumentException` | a cut the caller already chose |
+| Member | Takes |
+| ------ | ----- |
+| `Truncate(int maxLength)` | the longest prefix within the budget |
+| `Truncate(Range)` | the largest whole window inside the range |
 
-Moving the cut is right for a budget, where the number is a limit rather than a
-claim, and refusing is right for a range, where the caller has named a span and
-quietly returning a different one is the worse answer. `Truncate` therefore
-reports no truncation flag: `result.Length < Length` already answers it, and a
-flag that restates a comparison is a second thing to keep true.
+Both bounds move **inward** — the start forward to the next whole spelling, the
+end back to the previous one — so the result is always a subset of what was
+asked for. That asymmetry is the whole of the safety argument: returning less
+than a caller asked for is something it can detect from `Length`, while
+returning more is not. A window holding no whole spelling is therefore empty
+rather than widened to the spelling enclosing it.
 
-Neither member moves a **start**. Moving an end down returns less than was
-asked for, which a caller can detect; moving a start down returns *more*, which
-is the one direction it cannot check.
+Every range is answerable, including a reversed one and one that runs off
+either end. Neither member reports truncation separately, because
+`result.Length` against the width that was asked for already says whether
+anything moved.
+
+There is deliberately **no exact counterpart** that refuses an unusable window.
+An indexer would wear the syntax of ordinary slicing while throwing on bounds
+that look perfectly reasonable — six of the twelve positions in an
+eleven-character value divide a spelling — so the shape invites exactly the call
+it refuses. It would also buy nothing over the length comparison above.
 
 A surrogate pair counts as one thing however it is spelled — raw, as `\U`, or as
 the two `\uXXXX` escapes that composition produces when the halves are encoded in
--- a/src/InertText.Tests/BoundaryCorpus.cs
+++ b/src/InertText.Tests/BoundaryCorpus.cs
@@ -1,0 +1,80 @@
+namespace InertText.Tests;
+
+/// <summary>
+/// The values the boundary sweeps run over: every <see cref="AdversarialCorpus"/> payload
+/// encoded directly, plus the values only composition can produce.
+/// </summary>
+/// <remarks>
+/// Direct encoding cannot emit two adjacent <c>\uXXXX</c> escapes that together spell one astral
+/// scalar. A paired scalar is spelled as a single <c>\U</c> escape, and the corpus's lone
+/// surrogates have no partner to pair with. That token is twelve characters wide and is the only
+/// one in the alphabet whose width is not settled by its own prefix, so sweeping directly-encoded
+/// payloads alone leaves the walker's subtlest arm to whatever hand-written cases exist —
+/// breaking pair atomicity reddens those and nothing else, while every property sweep stays
+/// green.
+///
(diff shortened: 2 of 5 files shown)


richlander:
## Review

**Verdict: the product code is sound — approve. Two findings in the tests, one verified empirically: the Forms oracle overclaims its validity, and the corpus sweeps are structurally blind to the one token class only composition produces.**

I ran the suite in an isolated worktree (**317/317 green, 0.2s**) and re-derived the walker against `AppendSpelling`'s table: token widths, greedy pair matching (including the three-escape adjacency where a lone `\uD834` precedes a `\uD834`+`\uDD73` pair, and the lone high correctly stays separate from the pair that follows), the concatenation seams composition can produce (`\\` + `u202E` stays two tokens and round-trips), inward-moving bounds over negative/reversed/past-end/from-end ranges, the identity fast path, and zero-value agreement. `Length` and `IndexOfFirstEncoded` rest on "every spelling starts with `\` and raw text can never contain one," which holds. The `Truncate`-not-`Slice` naming argument is correct, and dropping the throwing indexer was the right call. I also spot-checked the mutation table's headline row: reverting the pair width (`12` → `Math.Min(6, remaining)`) turns `Truncate_KeepsASpelledSurrogatePairWhole` red as claimed.

### Finding 1 (verified): the Forms oracle is only valid for directly-encoded values, and its remark claims more

`TruncateRange_ReportsTheSpellingsTheWalkerFindsAsTheEncoderWouldSpellThem` takes decode-then-re-encode as *"ground truth … from outside the walker."* It is not ground truth for every `InertString` — only for values produced by direct encoding. I ran this probe in the worktree:

```csharp
InertString split = InertString.Join(string.Empty, TextPolicy.Field,
    [new(TextPolicy.Field, "\uD83D"), new(TextPolicy.Field, "\uDE00")]);
InertString window = split.Truncate(0..split.Length);

VisualEncoder.TryDecode(window.ToString(), out string? original);
Assert.Equal(VisualEncoder.Encode(TextPolicy.Field, original!).Forms, window.Forms);
// FAILS: Expected None, Actual BmpHex — same result for a proper sub-window.
```

The window text `😀` decodes to a raw emoji, which re-encodes to itself with `Forms = None` — while the walker reports `BmpHex`. **The walker is the correct side.** `Forms`' contract is "the spellings emitted while producing this value; composition unions the flags," and the composed value's own `Forms` is `BmpHex` — so the oracle disagrees with composition semantics generally, not just with the walker; `DescribeLegend` on the window correctly names the `\uXXXX` spelling the reader actually sees. The sweep stays green today only because direct encoding of the corpus payloads can never emit adjacent high/low escapes. Whoever later strengthens the sweep to composed inputs — the natural move — gets a red test where the code under test is right. Suggest scoping the remark to directly-encoded values, and pinning the composed case explicitly with `BmpHex` expected so the walker's composition-consistent answer is the stated contract.

### Finding 2: every corpus sweep is blind to the escaped-pair token

All four every-budget/every-window sweeps run over `new InertString(TextPolicy.Field, payload)` — and direct encoding can never produce the `\uHIGH\uLOW` adjacency, since a paired scalar encodes as one `\U` escape and the corpus's lone surrogates have no partner. So the width-12 arm of the walker is covered by exactly the two hand-written tests. My mutation run confirms it: breaking pair atomicity turns **1 test of 317** red, while the sweeps — the tests carrying the strong properties (fixpoint at every budget, subset at every window, maximality) — all stay green. The atomicity guarantee the PR body calls a would-be regression deserves the same exhaustive treatment as everything else: add one or two composed values to the sweep inputs (the plain pair, and ideally the lone-high-then-pair adjacency — a `\uD834` fragment joined before a fragment that composes to the `\uD834`+`\uDD73` pair — whose greedy grouping is the subtlest thing the walker does). Fixing Finding 1 first is what makes this extension possible, since the form-oracle sweep is the one that would fail.

### What else I checked and found sound

- **The maximality oracle** in `Truncate_KeepsAsMuchAsTheBudgetAllows` — inferring "position divides a spelling" from "a window opened there loses text" — is airtight given start-moves-forward-only semantics, and it is a genuinely nice replacement for the deleted indexer.
- **`WindowWithin` totality**: start below zero reads as 0 without a clamp, the end walks from the start so a reversed range is empty by construction, and the width clamps convert a malformed value into a short token rather than an `IndexOutOfRangeException` — consistent with the internal constructor asserting rather than establishing the invariant.
- **The identity fast path** returns `this` (original string instance and `Forms`) exactly when the window is the whole text, so an unbinding budget is free.
- **No new capability surface**: `Length`/`IndexOfFirstEncoded`/`Truncate` return no text derived from the original and take none, so all the #3636 reflection gates remain satisfied without edits — which is itself evidence the gates were shaped right.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

richlander:
Both findings reproduced before acting; the second turned out to go deeper than the suggested fix.

**The `Forms` oracle.** Correct — decode-then-re-encode asks "what would a fresh encode spell?", not "what was this spelled with", and those diverge for composed values. A `Join`-composed `\uD83D\uDE00` is reported `BmpHex` by the walker and `None` by a re-encode, because `Field` permits the astral scalar and emits it raw. The walker is the correct side there: a reader of the encoded text does see `\uXXXX` escapes. I scoped the oracle's remark to directly-encoded values and added `TruncateRange_KeepsTheSpellingsCompositionEmitted` to pin the composed contract directly. (The asymmetry is pre-existing `Join` behaviour, not introduced here.)

**Pair atomicity — necessary but not sufficient.** The diagnosis was right: direct encoding cannot produce the width-12 token, so the sweeps had never seen it. But adding composed inputs alone still killed only **1** test.

The reason is that each half of a divided `\uD83D\uDE00` is itself a well-formed `\uXXXX` escape. So a cut between them decodes cleanly, satisfies `IsPermitted` (the text is ASCII), and is therefore still an `EnsurePermitted` fixpoint. Every property those sweeps asserted lives on the **encoded** side, and a divided pair is well-formed there. The harm — handing a sink a lone surrogate — only exists on the **decoded** side, which nothing was looking at.

So the assertion had to move rather than just widen: *a bound may not fall between the halves of a surrogate pair of the decoded text*. That forbids creating a lone surrogate without forbidding carrying one, which matters because the corpus holds two deliberately. The range sweep carries it too — the near bound exists only there, and the same token walk places both.

Collapsing the pair token to six characters now fails **7** tests rather than 1, with both sweeps among them.

330 tests, up from 317. Also rebased onto `main` now that #3636 has merged; `range-diff` reports `=` for all eight commits.

PR #3677, opened 2026-08-01
Build the Native AOT tools with OptimizationPreference=Speed

### LINKED ISSUE #3675: Native AOT: switch to OptimizationPreference=Speed; treat the x64 instruction-set baseline as a separate hardware-floor decision
## Summary

`dotnet-inspect` ships primarily as Native AOT, but the two settings that govern its
codegen have never been measured on an AOT binary. Measuring them shows the current
configuration is the worst of the six available combinations for scalar code.

Proposed, as **two independent decisions** — they were originally bundled, which was a
mistake, because only one of them is a pure win:

**A. `OptimizationPreference`: `Size` → `Speed`, for every AOT RID.**
Measured 1.5x–1.9x on scalar code on both x64 and arm64. Changes no hardware
requirement, so it carries no compatibility risk. Costs +9.5% to +12.3% binary size.
Set in `src/dotnet-inspect/dotnet-inspect.csproj`.

**B. `IlcInstructionSet`: unset (`x86-64-v2`) → `x86-64-v3`, for `win-x64`/`linux-x64` only.**
This is a **product hardware-floor decision**, not a free optimization — see
"The v3 decision is a hardware floor" below. Its measured benefit accrues only to
vectorized code. Recommend deciding it separately from A, and note the CLI has very
little vectorized code today for it to pay off on.

Arm64 needs no baseline change at all (measured below).

Cost is a 12.3% larger binary. `arm64` is not yet proposed — see "What we still need".

A JIT benchmark cannot answer this. The JIT picks vector widths at runtime from the
actual CPU; Native AOT bakes them in at compile time. On the same machine a JIT run
reported `Vector512` accelerated while the default AOT binary reported only `Vector128`.

## Measurements

Host: AMD Ryzen 9 9900X (Zen 5, AM5), Linux 6.14, .NET 11.0.100-preview.7.
Produced by `tools/AotProbe/run.sh` (branch `perf/naot-instruction-set-probe`).
Best of 9 repetitions, nanoseconds per operation.

`scan` is a scalar per-scalar walk (`InertString.IsPermitted`), representative of the
CLI's text handling. `pre` is a vectorized span search (`IndexOfAnyExceptInRange`),
representative of BCL vectorized helpers.

| config | scan typename | scan signature | scan long4k | pre typename | pre signature | pre long4k |
|---|--:|--:|--:|--:|--:|--:|
| **v2-Size (today)** | 208.7 | 446.6 | 12807 | 3.77 | 5.76 | 146.1 |
| v2-Speed | 141.2 | 295.8 | 8399 | 3.67 | 5.49 | 148.1 |
| v3-Size | 210.5 | 445.2 | 12747 | 2.51 | 3.87 | 48.6 |
| **v3-Speed (proposed)** | 139.8 | 293.2 | 8434 | 2.51 | 3.80 | 48.7 |
| v4-Size | 207.4 | 448.6 | 12869 | 2.22 | 3.17 | 26.3 |
| v4-Speed | 167.3 | 297.2 | 8361 | 2.68 | 2.97 | 27.7 |

### The two settings are orthogonal

This is the main result, and it is why both changes are proposed together rather than
traded against each other.

- `OptimizationPreference` moves **scalar** code and barely touches vectorized code:
  scan improves ~1.5x at every baseline, `pre` is unchanged.
- The instruction-set baseline moves **vectorized** code and does not touch scalar code
  **at all**: scan is 208.7 / 210.5 / 207.4 across v2 / v3 / v4. A scalar loop cannot
  spend wider vector registers.

So the current `v2 + Size` configuration gives up ~1.5x on scalar code and ~3x on
vectorized code independently, and `v3 + Speed` recovers both.

### Binary size

| config | bytes | vs today |
|---|--:|--:|
| v2-Size (today) | 1,263,600 | — |
| v3-Size | 1,263,600 | +0.0% |
| v3-Speed | 1,419,200 | **+12.3%** |
| v4-Speed | 1,423,296 | +12.6% |

Raising the baseline is free in size terms. The 12.3% is entirely `Speed`.

`OptimizationPreference=Size` was presumably a deliberate choice for a CLI tool. The
point here is only to put an exchange rate on it: 12.3% size for ~1.5x on scalar code.

## Why not v4

v4 is faster still on long inputs (26.3 vs 48.7 ns at 4 KB) and worth ~1 ns on
realistic metadata strings. It should nonetheless be rejected:

- **No Intel consumer CPU since 11th-gen Rocket Lake / Tiger Lake (2020-21) has
  AVX-512.** Alder Lake, Raptor Lake, Raptor Refresh, Meteor Lake and Arrow Lake all
  lack it, because the E-cores cannot execute it. It returns with Nova Lake, expected
  late 2026 or 2027. AMD Zen 4/5 and Intel Xeon do have it.
- **Native AOT has no fallback path.** An unsupported CPU fails at *startup* with
  "The required instruction sets are not supported by the current CPU." This is not a
  slow path; the tool does not run at all.

## The v3 decision is a hardware floor

An earlier revision of this issue claimed v3 was "a safe floor for anything running a
current .NET SDK". **That was wrong**, and the correction matters enough to state
plainly, because the same no-fallback argument used against v4 applies to v3.

Per the [.NET 11 minimum hardware
matrix](https://learn.microsoft.com/en-us/dotnet/core/compatibility/jit/11/minimum-hardware-requirements):

| | x64 JIT/AOT minimum | x64 R2R target |
| --- | --- | --- |
| .NET 10 | `x86-64-v1` | `x86-64-v1` |
| .NET 11 | **`x86-64-v2`** | `x86-64-v3` (Windows/Linux) |

`x86-64-v3` appears only as the **ReadyToRun** target — and it can be, precisely because
a v2 host that cannot use an R2R image falls back to jitting. Native AOT has no such
fallback. The supported JIT/AOT floor is `x86-64-v2`, which is what an unset baseline
already gives us.

So adopting v3 for Native AOT means **deliberately dropping supported v2-only hosts**:
pre-Haswell Intel (before 2013) and pre-Excavator AMD (before 2015). That may well be an
acceptable product decision, but it has to be made as one, on its own evidence about who
runs this tool. It cannot be derived from what the SDK supports, and this issue should
not have implied otherwise.

Worth weighing against it: the measured v2 → v3 win is entirely in the **vectorized**
column, and the CLI has almost no vectorized code today (`src/InertText/` uses
`SearchValues` and friends zero times). Adopting v3 now would buy very little until that
changes, while the compatibility cost is immediate.

## Arm64 results

Collected on an **Apple M4 Pro** (`osx-arm64`, .NET SDK `11.0.100-preview.6.26359.118`)
by running `tools/AotProbe/run.sh` at `e53ed48e`. **The hypothesis held.**

- Every baseline — `default`, `armv8.1-a`, `armv8.2-a`, `apple-m1` — reports
  `v128=1 v256=0 v512=0`.
- `pre` is unchanged across baselines: ~4.24-4.33 ns (typename), ~6.76-6.91 ns
  (signature), ~145-149 ns (4 KiB).
- **All four baseline binaries are byte-identical** within each optimization mode. This
  is the cleanest possible confirmation: on arm64 the baseline is not a tradeoff to
  evaluate, it is a no-op. (It follows from the table above — the Apple JIT/AOT minimum
  is already `Apple M1` (~`armv8.5-a`), so every value swept is at or below the existing
  default.)
- `Speed` improves `scan` at the default baseline: 204.7 → 118.8 ns (1.72x),
  400.9 → 254.9 ns (1.57x), 13,850.5 → 7,250.7 ns (1.91x) — a slightly *larger* factor
  than x64's ~1.5x.
- Size cost is smaller than x64's: 1,334,008 vs 1,218,504 bytes (**+9.5%**).

**Recommendation for the arm64 RIDs: adopt `Speed`, leave the baseline unset.**

The `linux-arm64` caution from the original request stands and is now moot in the same
direction: `armv8.2-a` is known to fail outright on Graviton2 (dotnet/runtime#89937), and
there is nothing to gain by raising it anyway.

## Scope

This is a build-configuration change, independent of any library work. The probe lands
under `tools/` as a contained evaluation and is not part of the CLI's reference closure.

Branch: `perf/naot-instruction-set-probe`

richlander:
Closes #3675.

Switches both Native AOT tools from `OptimizationPreference=Size` to `Speed`,
and fixes an MSBuild ordering bug that made the setting inert for local
publishes. Also lands `tools/AotProbe`, the instrument used to measure it.

Raising the x64 instruction-set baseline to `x86-64-v3` was measured alongside
this and **deliberately not taken**. That negative result is written up below,
because it is the more surprising half.

## The change

`OptimizationPreference=Speed` for every AOT RID, on both `dotnet-inspect` and
`mdi` — they ship from the same RID list and should not disagree about codegen.

Timed on published `linux-x64` binaries, min of 25 alternating runs, warmed:

| workload | today (`Size`) | `Speed` | |
| --- | --: | --: | --: |
| `member System.Convert` | 455.3 ms | 427.2 ms | **1.066x** |
| `member System.String` | 454.7 ms | 424.3 ms | **1.072x** |
| `member System.Math` | 449.0 ms | 422.4 ms | **1.063x** |
| `member String.Substring -S "Decompiled Source"` | 458.9 ms | 431.8 ms | **1.063x** |
| `find Stream --platform` | 40.5 ms | 38.8 ms | 1.044x |

**A consistent 6-7% on real commands.** The isolated microbenchmark shows
1.5x-1.9x on both x64 and arm64, but only part of a command's runtime is that
scalar code, so 6-7% is the honest user-visible number.

`Speed` is the setting that matters for this tool because the tool is scalar:
the ~420 ms above is ECMA-335 metadata walking — small reads, integer decoding,
dictionary lookups.

Cost is binary size: **+9.1%** on `linux-x64` (31,090,352 → 33,925,056 bytes),
+9.5% on arm64.

## Why not also raise the x64 baseline to x86-64-v3

The tempting argument: .NET 11 raised the Windows/Linux ReadyToRun target to
`x86-64-v3` ([dotnet/runtime#120826][1]), R2R only has to be conservative
because a v2 host falls back to jitting, and a developer CLI has a narrower
audience than the runtime's don't-break-anyone floor. On the isolated
benchmark, v2 → v3 is a **3.0x** win on vectorized code.

It does not survive contact with the shipped binary. Adding v3 on top of
`Speed` produced 1.086x, 1.062x, 1.065x, 1.082x against `Speed`'s 1.066x,
1.072x, 1.063x, 1.063x — scatter on both sides, not signal.

The obvious explanation is wrong, and worth recording. It is **not** that the
binary lacks vectorized code. Native AOT statically links the BCL and compiles
it at the chosen baseline — the BCL's whole premise is that it absorbs SIMD and
register-width variation on your behalf — so raising the baseline takes the
published binary from **390** to **31,025** `ymm` references. `IndexOf`,
`Contains` and `SequenceEqual` really do get wider.

The saving just scales with how much data each call scans: about **1.2 ns** on
a short type name against about **99 ns** on a 4 KB buffer. This tool handles
short identifiers and emits a few KB per command (largest output found: 5,389
bytes), so millions of calls would be needed to surface a millisecond.

So v3 would have cost pre-2013 Intel and pre-2015 AMD support — a hard startup
refusal, verified under emulation — to buy nothing measurable. The probe is
committed so the decision can be revisited if a long-buffer scan ever appears.

arm64 was never a candidate: AdvSimd is mandatory in the ARMv8-A baseline and
.NET exposes no wider vectors there, confirmed by all four arm64 baseline
binaries coming out byte-identical per optimization mode.

## The ordering bug

Both tools guarded these settings with a `PropertyGroup` conditioned on
`PublishAot` that sat **above** the line setting `PublishAot`. MSBuild evaluates
top to bottom, so for a plain `dotnet publish -r <rid>` the entire group was
silently inert:

```console
$ dotnet msbuild src/dotnet-inspect/dotnet-inspect.csproj -p:RuntimeIdentifier=linux-x64 \
    -getProperty:PublishAot -getProperty:OptimizationPreference
  "PublishAot": "true",
  "OptimizationPreference": "",
```

Official builds were unaffected, because `OfficialAotBuild` sets `PublishAot` in
`src/Directory.Build.props` first — which is why this never surfaced. Locally
published binaries were simply tuned differently from shipped ones. The groups
now sit below the property they test.

## tools/AotProbe

The instrument behind every number here. `run.sh` selects the RID and ISA list
from `uname` and sweeps ISA × Size/Speed; `Program.cs` uses adaptive iteration
counts and best-of-9, because the measurements are noisy enough that single runs
disagree. It is how the arm64 column was collected on hardware I do not have.

Its main lesson is the one above: a microbenchmark of a BCL primitive does not
predict an end-to-end gain in the tool that calls it.

[1]: https://github.com/dotnet/runtime/issues/120826

### DIFF
--- a/dotnet-inspect.slnx
+++ b/dotnet-inspect.slnx
@@ -83,6 +83,7 @@
   </Folder>
   <Folder Name="/tools/">
     <Project Path="tools/AnalysisHarness/AnalysisHarness.csproj" />
+    <Project Path="tools/AotProbe/AotProbe.csproj" />
     <Project Path="tools/CSharpDiffHarness/CSharpDiffHarness.csproj" />
     <Project Path="tools/DecompilerHarness/DecompilerHarness.csproj" />
     <Project Path="tools/DiffHarnessCommon/DiffHarnessCommon.csproj" />
--- a/src/dotnet-inspect/dotnet-inspect.csproj
+++ b/src/dotnet-inspect/dotnet-inspect.csproj
@@ -9,13 +9,6 @@
     <RollForward>LatestMajor</RollForward>
   </PropertyGroup>
 
-  <!-- Native AOT size optimizations (only apply when publishing with AOT) -->
-  <PropertyGroup Condition="'$(PublishAot)' == 'true'">
-    <UseSystemResourceKeys>true</UseSystemResourceKeys>
-    <InvariantGlobalization>true</InvariantGlobalization>
-    <OptimizationPreference>Size</OptimizationPreference>
-  </PropertyGroup>
-
   <ItemGroup>
     <InternalsVisibleTo Include="dotnet-inspect.Tests" />
   </ItemGroup>
@@ -37,6 +30,23 @@
     <!-- Keep debug symbols (.pdb/.dbg and the macOS .dSYM bundle) out of the
          published tool package. -->
     <CopyOutputSymbolsToPublishDirectory>false</CopyOutputSymbolsToPublishDirectory>
+  </PropertyGroup>
+
+  <!-- Native AOT publish settings. This group must come after PublishAot is set
+       above: MSBuild evaluates properties top to bottom, so a group conditioned on
+       PublishAot that sits earlier in the file is silently inert for a plain
+       `dotnet publish -r <rid>` (where PublishAot comes from this file rather than
+       from a global property or OfficialAotBuild). -->
+  <PropertyGroup Condition="'$(PublishAot)' == 'true'">
+    <UseSystemResourceKeys>true</UseSystemResourceKeys>
+    <InvariantGlobalization>true</InvariantGlobalization>
+    <!-- Speed, not Size. Worth a measured 6-7% on real commands for about 9% more
+         binary size (issue #3675). The setting moves scalar codegen, which is what
+         this tool is: metadata walking is small reads, integer decoding and
+         dictionary lookups. Raising the x64 instruction-set baseline was measured
+         alongside it and deliberately not taken: it moves vectorized codegen
+         only, which is a null result here and would have cost pre-2013 hardware. -->
+    <OptimizationPreference>Speed</OptimizationPreference>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(OfficialBuild)' == 'true'">
--- a/src/dotnet-inspect/release-notes.md
+++ b/src/dotnet-inspect/release-notes.md
@@ -33,6 +33,12 @@
   `Kind` column. The `type`/`member` `Performance Triage` lens is
   unchanged (#2833).
 
+### Native AOT codegen
+
+- Builds the Native AOT tools with `OptimizationPreference=Speed` instead of
+  `Size`, worth a measured 6-7% on real commands for about 9% more binary size
+  (#3675).
+
 ### Output and projections
 
 - Adds JSON array projection output and scalar URL/path shape projections,
--- a/src/mdi/mdi.csproj
+++ b/src/mdi/mdi.csproj
@@ -15,13 +15,6 @@
     <OwnsItsOwnStderr>true</OwnsItsOwnStderr>
   </PropertyGroup>
 
-  <!-- Native AOT size optimizations (only apply when publishing with AOT) -->
-  <PropertyGroup Condition="'$(PublishAot)' == 'true'">
-    <UseSystemResourceKeys>true</UseSystemResourceKeys>
-    <InvariantGlobalization>true</InvariantGlobalization>
-    <OptimizationPreference>Size</OptimizationPreference>
-  </PropertyGroup>
-
   <ItemGroup>
     <InternalsVisibleTo Include="DotnetInspector.MetadataRendering.Tests" />
   </ItemGroup>
@@ -39,6 +32,16 @@
     <PublishAot>true</PublishAot>
     <!-- Keep debug symbols out of the published tool package. -->
     <CopyOutputSymbolsToPublishDirectory>false</CopyOutputSymbolsToPublishDirectory>
+  </PropertyGroup>
+
+  <!-- Native AOT publish settings. Must come after PublishAot is set above; see the
+       matching comment in dotnet-inspect.csproj for why the ordering matters.
+       mdi ships from the same RID list as dotnet-inspect, so the two tools carry the
+       same codegen settings. Rationale: issue #3675. -->
+  <PropertyGroup Condition="'$(PublishAot)' == 'true'">
+    <UseSystemResourceKeys>true</UseSystemResourceKeys>
+    <InvariantGlobalization>true</InvariantGlobalization>
+    <OptimizationPreference>Speed</OptimizationPreference>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(OfficialBuild)' == 'true'">
--- a/tools/AotProbe/.gitignore
+++ b/tools/AotProbe/.gitignore
@@ -1,0 +1,2 @@
+# Published Native AOT output from run.sh.
+out-*/
--- a/tools/AotProbe/AotProbe.csproj
+++ b/tools/AotProbe/AotProbe.csproj
@@ -1,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <!--
+    Measures how the two Native AOT codegen knobs - the instruction-set baseline
+    (IlcInstructionSet) and OptimizationPreference - affect the two shapes of code the
+    CLI spends its text-handling time in: a scalar per-scalar walk and a vectorized
+    span search. dotnet-inspect ships primarily as Native AOT, so these are settled by
+    measurement on the AOT binary rather than by JIT benchmarks. See run.sh.
+
+    This is a contained runnable evaluation, not a shipping component.
+  -->
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
+    <LangVersion>preview</LangVersion>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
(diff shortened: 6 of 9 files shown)


richlander:
Correcting the analysis in this PR after measuring the shipped binary rather than the probe.

I had written that B (`x86-64-v3`) was a bet because the tool "has very little vectorized code today." That reasoning was wrong. Native AOT statically links the BCL and compiles it at the chosen baseline, so raising it takes the published binary from **390** to **31,025** `ymm` references — `IndexOf`, `Contains`, `SequenceEqual` and the rest really do get wider.

The conclusion survives for a different reason: the v2→v3 saving scales with input length (~1.2 ns on a short type name vs ~99 ns on a 4 KB buffer), and this tool handles short identifiers and emits a few KB per command. End to end, min of 25 alternating runs:

| workload | today (v2-Size) | +A (v2-Speed) | +A+B (v3-Speed) |
| --- | --- | --- | --- |
| `member System.Convert` | 455.3 ms | 427.2 (**1.066x**) | 419.1 (1.086x) |
| `member System.String` | 454.7 ms | 424.3 (**1.072x**) | 428.2 (1.062x) |
| `member System.Math` | 449.0 ms | 422.4 (**1.063x**) | 421.6 (1.065x) |
| `member String.Substring -S "Decompiled Source"` | 458.9 ms | 431.8 (**1.063x**) | 424.1 (1.082x) |

A is worth a consistent 6-7% on real commands. B adds nothing measurable on top — the A+B column scatters both sides of A.

The body now reflects this. A is justified on measurement; B is a forward-looking bet whose present cost is dropping pre-2013 Intel and pre-2015 AMD hardware.

richlander:
Dropped the `x86-64-v3` baseline; this PR is now `OptimizationPreference=Speed` only.

The measurement is what settled it. `Speed` is worth a consistent **6-7%** on real commands. v3 on top of it was a null result end to end, so it would have cost pre-2013 Intel and pre-2015 AMD support to buy nothing measurable.

The negative result is written up in the body rather than dropped, along with why the intuitive explanation for it is wrong: the binary is *not* short of vectorized code (v3 takes it from 390 to 31,025 `ymm` references), the saving just scales with input length and this tool only ever scans short identifiers.

`tools/AotProbe` stays so the decision can be revisited if a long-buffer scan ever lands.

Issue #3592, opened 2026-07-30
Layout_Count test is casing-dependent: fails when the package is in the global packages folder


richlander:
`CommandExecutionTests.Layout_Count_CountsFilesRatherThanRenderedTreeLines` fails reproducibly on a developer machine, and passes on CI.

```
Assert.Contains() Failure: Sub-string not found
Not found: "Newtonsoft.Json.nuspec"
```

## Cause

The assertion is `Assert.Contains("Newtonsoft.Json.nuspec", rendered, StringComparison.Ordinal)`, but the rendered name depends on **where the package came from**:

- Extracted from the `.nupkg`, the nuspec keeps the authored casing: `Newtonsoft.Json.nuspec`.
- Resolved from the **global packages folder**, NuGet has already written it lowercased: `~/.nuget/packages/newtonsoft.json/13.0.4/newtonsoft.json.nuspec`.

Observed on a clean checkout of `main` (`ce9a508d3`):

```
├─ LICENSE.md
├─ newtonsoft.json.nuspec
├─ packageIcon.png
└─ README.md
```

The `--count` half is unaffected (still 20), so only the casing-sensitive substring assertion breaks. CI passes because a fresh runner has no restored copy and takes the `.nupkg` path.

## Why it matters beyond the test

The test is the symptom; the question underneath is whether the two acquisition paths are *supposed* to render identical layouts. Right now they do not, and nothing pins that. A user comparing `--layout` output across machines sees different file names for the same package version.

## Suggested fix

Decide the intended behavior first:

1. If layout should reflect what is physically on disk, make the assertion casing-insensitive (or match on `.nuspec`) and add a test pinning that the two paths may differ.
2. If layout should be canonical regardless of source, normalize the nuspec name at the lens and pin *that* — which also makes the existing assertion correct as written.

Found while validating an unrelated PR; not caused by it, and reproducible on `main` with no local changes.

richlander:
Resolved by merged PR #3609 (commit `6cd14078af946135e531d7f9c392af9b123cb7cf`). That change made the nuspec assertion case-insensitive because casing differs legitimately between `.nupkg` extraction and the NuGet global packages folder while preserving the file-count and rendered-tree claims. Verified on current `main`: Release build passed and `Layout_Count_CountsFilesRatherThanRenderedTreeLines` passed.

PR #3679, opened 2026-08-02
Contain metadata cell text by Unicode category, not by a hand-written range

### LINKED ISSUE #3628: Metadata table cells escape C0 controls but not bidi overrides, leaking raw U+202E to the terminal

richlander:
Part of #3635. Closes #3628.

Stacked on #3660 — this uses `InertString.Truncate`, so it cannot merge before it.
The diff against that base is two product files, two test files, and two docs.

## The defect

`MetadataTableProjector` decided what to escape with a hand-rolled predicate:

```csharp
static bool IsControl(char c) => c < ' ' || c == '\x7f' || (c >= '\x80' && c <= '\x9f');
```

C0, DEL and C1. Everything else reached the terminal exactly as the artifact
spelled it. Against a purpose-built hostile assembly, on `main`:

| Section | U+202E | U+2028 | U+200B |
| --- | --- | --- | --- |
| `Metadata: TypeDef` | 1 | 1 | 0 |
| `Metadata: Field` | 1 | 0 | 1 |
| `Metadata: MethodDef` | 1 | 0 | 0 |
| `Metadata: #Strings` | 3 | 1 | 1 |

The metadata lens is the *first* place hostile assembly metadata is rendered,
and it missed the one attack class the adversarial corpus is named after.

The range was also `char`-shaped, which is a second and strictly larger hole:
a supplementary scalar arrives as two surrogates, and neither surrogate carries
the category of the scalar it spells. No `char` predicate can classify
`U+E0074` no matter which ranges it lists.

## The change

Both entry points now go through the primitive:

```csharp
internal static string ContainCellText(string value, int maxChars, out bool truncated)
{
    var contained = new InertString(TextPolicy.Field, value);
    InertString bounded = contained.Truncate(maxChars);
    truncated = bounded.Length < contained.Length;
    return bounded.ToString();
}
```

`TextPolicy.Field` classifies by Unicode general category — `Cc`, `Cf`, `Cs`,
`Zl`, `Zp` — so the bidi overrides, the separators and the zero-width
characters are covered because of *what they are*, not because someone
remembered to list them. `Truncate` supplies the emitted-length budget the old
code enforced by hand, with the same surrogate-pair atomicity.

`EscapeCore` did three jobs at once: budget truncation, structural escaping,
and visual escaping. Only the third was wrong, but all three collapse into the
one call above.

### Why structural escaping collapsed rather than composed

Layering was not available. The encoder escapes backslash *unconditionally*, so
its output is self-delimiting; putting structural escaping either side of it
double-escapes the backslashes the other introduced.

It turned out not to be needed. `Field` already refuses every member of the
structural set `\ " \n \r \t` except the quote — and the quote is not load
bearing. Cells are not quote-wrapped in Markdown or TSV, and JSON escapes
independently in the serializer, so `\"` was decoration. Dropping it *removes*
a fidelity loss: a name genuinely containing `"` used to render as `\"`,
implying a backslash the artifact did not contain.

## What changes in the output, measured

Swept the old and new escapers over **all 1,112,064 Unicode scalars**:

| | scalars |
| --- | --- |
| identical | 1,111,858 |
| **newly contained** (raw before) | **172** — 45 BMP, 127 supplementary |
| respelled (contained before and after) | 33 |
| no longer escaped | 1 — the quote |

The 172 are the fix: the full bidi set `U+202A`–`U+202E` and `U+2066`–`U+2069`,
`U+2028`/`U+2029`, ZWSP/ZWNJ/ZWJ, BOM, soft hyphen, the Arabic number signs,
and the supplementary tag block `U+E0020`–`U+E007F` that a `char` predicate
cannot see. **Nothing moved the other way** — not one scalar contained before
is raw now, apart from the quote.

The 33 are C0 and DEL changing spelling only, `\u001B` → `\^[`.

On real artifacts nothing changes at all. Byte-identical across 13 MB of
`library` metadata sections and 63 MB of `mdi` output over six framework
assemblies, seven sections, three formats and six truncation budgets — the last
of these being the non-vacuity floor for the budget half, since names are cut
mid-cell there (`ReadOnl…`). Framework metadata contains no escapable scalar at
all, so that corpus is deliberately *not* offered as evidence about the
escaping half; the sweep above covers that exhaustively instead.

## The gate could not have caught this

`MdiContainmentTests` was already a good end-to-end gate — real spliced
assembly, every format, every view — and it was green throughout. Its payload
was documented as spanning "all three ranges `MetadataTableProjector.IsControl`
recognizes", and its assertion helper said it "mirrors
`MetadataTableProjector.IsControl`".

So the payload, the oracle and the predicate all agreed with each other. A gate
in that shape can only ever confirm the predicate; it cannot discover that the
predicate is looking at the wrong set. That is the reusable lesson here, more
than the missing characters.

Both halves are widened: the payload now carries `U+202E`, `U+2028`, `U+200B`
and `U+E0074` alongside the live `ESC [ 3 1 m` sequence, and the oracle
classifies by category over scalars decoded with `Rune.DecodeFromUtf16`.

Mutation-checked in both directions the old design was blind in:

| Mutation | Result |
| --- | --- |
| restore the pre-#3628 range predicate | 9 tests red, naming `U+202E` |
| category-correct but `char`-based predicate | 9 tests red, naming `U+E0074` |

The second is the one worth having: it is what a careful fix that kept iterating
`char`s would look like, and it fails on the supplementary scalar alone.

## Also here

`EscapeText_KeepsSurrogatePairsAtomic…` reflected on the escaper by name, so
renaming it produced a `NullReferenceException` inside the test rather than a
build error. It calls the method directly now — the project already has
`InternalsVisibleTo`.

Six of the seven predicates in #3635's census remain. This PR retires one
(`MetadataTableProjector.IsControl`); `AttributeReader.cs:782` and
`ResourceExtractor.cs:195` are untouched and still open.

## Verification

`ILInspector.Metadata.Tests` 1283/0, `DotnetInspector.MetadataRendering.Tests`
149/0, `ILInspector.CSharp.Tests` 449/0, `dotnet-inspect.Tests` 2817 with the
one known local-only failure (#3592, nuspec filename casing; green in CI).


richlander:
Rebased onto `main` now that #3660 and #3601 have landed; the base branch was squashed, so the stack had to be replayed. Two conflicts, both in docs:

- **`release-notes.md`** — #3675 added a section in the same place. Both kept.
- **`untrusted-data-threat-model.md`** — #3601 had already rewritten this row, and had named the gap honestly: *"the payload is `Cc` only, so a bidi override would not be noticed, and the assertion deliberately permits raw `CR`/`LF`/`TAB`. Format delimiters are not covered by this gate at all."*

  My version fixed the first limit but had silently dropped the other two, which are still true. The resolved row keeps them, so the doc still names what the gate does **not** cover.

No product code changed in the rebase. Metadata 1283/0 and MetadataRendering 149/0 re-run against the new base.

richlander:
Folded in two commits from a parallel branch that was solving #3628 independently (`fix/3628-metadata-cell-inert`, now redundant). Heads-up to whoever owns this PR: the head moved from `52d842a4d` to `5f5b135b5`, fast-forward, no force-push.

**`d71fb1718` — bound at construction, and report it.** `ContainCellText` encoded, then bounded, then kept the intermediate value only to subtract lengths:

```csharp
var contained = new InertString(TextPolicy.Field, value);
InertString bounded = contained.Truncate(maxChars);
truncated = bounded.Length < contained.Length;
```

`InertString` now takes the budget directly, so that becomes `new InertString(TextPolicy.Field, value, maxChars)` and `contained.IsTruncated`.

The part worth reviewing is not the constructor, it's the equality change that came with it. `IsTruncated` is *derived* — from the length the value was bounded from, so `Bound` carries it forward and a second cut can't report itself as the only one. But deriving it does **not** make it agree with `Equals`, which is what I initially and wrongly assumed: the length behind it is still state the text doesn't determine, so two values could carry identical text, disagree on `IsTruncated`, and compare equal while one renders with an ellipsis and the other doesn't. Verified before fixing:

```
whole  : [a\u202Eb] IsTruncated=False
clipped: [a\u202Eb] IsTruncated=True
whole == clipped -> True
```

`Equals`/`GetHashCode` now include the flag — the *flag*, not the length, because two values bounded from different lengths to the same text do render the same. `Forms` stays out and safely so: it's a function of the text, so equal text already implies equal forms. Backward compatible — anything built without a budget reports `false` on both sides, so equality is unchanged for every existing value.

Why the flag has to exist at all: bounding in one step doesn't leave the caller holding both values, and the comparison it would reach for instead — emitted length against the raw input's — is wrong in the direction that matters, because spelling a scalar makes text longer. At budget 8, `a\u202Eb\\c` is clipped from 11 to 8, and `8 < 5` says nothing was dropped. A cell cut mid-hazard would render without the ellipsis that marks it partial.

**`5f5b135b5` — sweep instead of enumerate.** This PR's fix to `MdiContainmentTests` is right and necessary, but that gate still *enumerates*, so it still answers only for what someone thought to write down — which is precisely how it stayed green through the whole life of this defect while describing itself as mirroring `MetadataTableProjector.IsControl`.

`MetadataCellContainmentTests` asks it the other way round: for every scalar Unicode classifies as `Cc`/`Cf`/`Zl`/`Zp` (surrogates separately, since `Rune` can't hold one), embed it between ordinary text, project, and fail if the output still contains it. That can't be satisfied by widening a range. Non-vacuity is membership, not a count — a threshold is a magic number to re-guess whenever Unicode adds a format character.

It calls `ContainCellText` directly rather than by reflection, which your collapse to one `internal` method made possible.

**Evidence.** Both gates tamper-tested, and both discriminate rather than just going red:

- Narrowing `IsNonGraphic` back to `Cc` only → fails the sweep, the truncation sweep, and 6 of 8 named hazards, while **ESC and NEL keep passing**, because the old rule did cover `Cc`.
- Forcing `IsTruncated` to `false` → reddens the corpus sweep plus the equality fact.
- Reverting equality to text-only → reddens the equality fact alone.

InertText 358, Metadata 1301, MetadataRendering 149, Services 357, Analysis 584, CLI 2008 (`-trait- "Speed=Slow"`) — all 0 failed. `-t:Rebuild` reports 9 warnings, matching the `origin/main` baseline.

**Not done, and worth not duplicating a second time:** the projection model still stores text and truncation as separate fields in three places — `HeapReference(..., Text, Preview, Truncated)`, `HandleRef(..., Display, DisplayTruncated)`, `MetadataImageOverview(MetadataVersion, ..., MetadataVersionTruncated)`. If those carried `InertString`, all three flags become derivable and `ContainCellText` disappears entirely. That's a separate slice — `.Truncated` alone has 33 consumers and it changes a public model shape.

richlander:

Two rounds at head `5f5b135b5`, from two different model families (Gemini 3.1 Pro, GPT-5.6). Resolved in `3edf70099`. Head moved; confirmation rounds are running at the new head.

### Blocking finding — both reviewers, independently

`EnsurePermitted`, `Join`, and the interpolated-string handler each built a new value through the internal constructor, which defaulted truncation state to "never bounded". A truncated value passing through any of them came back reporting `IsTruncated == false` — a clipped cell would render without the ellipsis that marks it partial, a false negative in the direction that matters.

```
before-repair:  [a<LF>b]    truncated=True
after-repair:   [a\^Jb]     truncated=False
join:           [a\u202Eb]  truncated=False
format:         [a\u202Eb]  truncated=False
```

**Gemini sharpened it into a regression the equality change *caused* rather than merely missed:** `Join([clipped])` renders identically to `clipped` but no longer compared equal to it. That's the sharper statement — putting `IsTruncated` into equality is only sound if every path preserving the text also preserves the flag. It didn't, so the change meant to fix substitutability was defeated on exactly those paths.

Accepted in full. GPT also identified the interpolated-string handler as a fourth site I had missed when I found the `EnsurePermitted` and `Join` cases myself.

### Where I did not take the suggested fix

Both reviewers proposed the same repair — carry the source length through those paths (`Math.Max(_sourceLength, ...)`). **I believe that repair is unsound, and unsound in the same direction as the bug it fixes.**

A length only means something inside one encoding. Re-spelling under a stricter policy makes the text longer, so a length remembered from the old spelling gets compared against text measured in the new one. Measured:

```
prose   : len=10  IsTruncated=True   (source encoded len = 11)
respelled under Field: len=30
if _sourceLength were preserved across the re-encode:  30 < 11  =  False
```

Ten line feeds cut from eleven characters under `Prose` re-spell to thirty under `Field`, because `\n` goes from one character to `\^J` at three. The suggested repair reports that truncated value as **whole**. Credit where due: this was caught by a reviewer's question about whether the three aspects — raw length, encoding expansion, truncation — were all accounted for, not by either automated round.

So the state is now the **fact**, not the length: `private readonly bool _truncated`. A bool survives re-spelling because it doesn't measure anything. `Bound` no longer needs a length to compare against — reaching its tail means the window is a proper subset, so it or-s in the cut it just made, which is what keeps a second cut from reporting itself as the only one. `EnsurePermitted` carries the fact across the re-encode; `Join` and the handler or it across their parts. The bounding constructor now assigns nothing, since `Truncate` establishes it.

Verified after:

```
join(one)     : [Hostil] IsTruncated=True   equals-clipped=True
interpolated  : [Hostil] IsTruncated=True
conformed     : len=30   IsTruncated=True   <-- 30 chars from an 11-char source
whole         : IsTruncated=False  join=False
default==Empty: True
```

### One deliberate imprecision

`Join` propagates from **any** part, not only the last. The flag says something was dropped, not where, which is imprecise for a part cut out of the middle. The alternative is a composed value asserting it is complete when it is not, and in a hardening library the safe error is over-marking. Flagged for the confirmation round to attack.

### Evidence

Each propagation is gated by a test that fails without it, tampered one site at a time: removing it from `EnsurePermitted`, `Join`, or the handler reddens that path's test **and only that one**; reverting `Bound`'s or reddens the budget sweep, the negative-budget case, and the second-cut case. Deleting the handler's assignment outright doesn't compile — `CS0649` under warnings-as-errors — so the compiler gates that one.

**One of my tampers initially looked clean and was not.** Removing the handler's assignment failed the *build*, so no tests ran, and the absence of `[FAIL]` lines read as a pass. Redone as a change that compiles, it correctly reddens the interpolation test alone. Recording it because "no failures" from a build that never produced a test binary is indistinguishable from a real pass at a glance.

Suites: InertText **362**, Metadata **1301**, MetadataRendering **149**, Services **357**, Analysis **584**, CLI **2008** (`-trait- "Speed=Slow"`) — all 0 failed. `-t:Rebuild` reports 9 warnings, matching the `origin/main` baseline.

### Verified clean by the round, not re-litigated

Both reviewers independently confirmed end-to-end containment at byte level against hostile assemblies — Gemini built one by patching an assembly-name string in the heap directly to `U+202E U+2028 ESC` — across MDI and CLI Markdown/TSV/JSONL, at budgets `0..40`, `1024`, and `int.MaxValue`, with non-vacuous positive controls: zero raw hazard bytes. Sweep non-vacuity via the legacy predicate: **237 swept, 172 leaked**, matching my own measurement. NativeAOT publish succeeds for both binaries. No Roslyn, no inspected-assembly loading. `default == Empty` holds with matching hashes; middle and end windows report truncation correctly; negative, zero and `int.MaxValue` budgets raise nothing.

richlander:
## Adversarial review round 2 — confirmation at `3edf70099`

The round-1 fix moved the head, so both families re-reviewed at the new fixed head in isolated worktrees. **Both clean: zero blocking, zero non-blocking.**

### Both reviewers independently reproduced the counterexample

Neither was told the measurement, only asked to attack the claim that a length-based repair is unsound. Both built it themselves and got the same numbers:

```
ten LFs + 'X'                      = 11 chars
bounded to 10 under Prose          = 10 chars, IsTruncated=True
re-spelled under Field (\n -> \^J) = 30 chars, IsTruncated=True
a preserved source length would evaluate  30 < 11  ->  false
```

Gemini: *"genuinely unsound because it forgets that re-spelling text often expands it."* GPT: *"a propagated old source length would evaluate `30 < 11` as false and incorrectly report the value whole."* This is the repair both of them recommended last round, so the round-2 verdict is each reviewer overturning its own round-1 suggestion on evidence.

### What round 2 added beyond confirming

GPT enumerated the construction paths exhaustively and confirmed the set is closed — fresh encoding (`VisualEncoder.cs:93`), bounding (`InertString.cs:315`), join (391), re-spelling (515), interpolation (662), metadata containment (`MetadataTableProjector.cs:1353`) — then searched for a false negative through range truncation, repeated truncation, no-op `EnsurePermitted`, nested composition, redaction, and empty/default values, and for a false positive through whole ranges and non-binding budgets. Found neither. It also noted `Truncate(^0..)` marks a nonempty value truncated, which is right: that range returns the empty suffix, so the cut is real.

Both independently resolved the `Join` over-marking question I flagged for attack, by the same route: its only production caller is `FeedFailureTelemetry.cs:182`, joining freshly constructed unbounded values, and nothing it produces reaches the renderer that appends the ellipsis. So the imprecision has no live consumer today, and remains fail-safe if it acquires one. GPT additionally confirmed no production dictionary, set, `Distinct`, or `Contains` depends on `InertString` equality, which is what makes the equality change safe to land.

Gemini re-ran my tampers rather than taking them on trust, and confirmed the one-test-one-gate correspondence: `EnsurePermitted` returning `false` reddens only `EnsurePermitted_KeepsTruncation_AcrossARespellingThatGrowsTheText`; the handler at `|= false` reddens only `Interpolation_ReportsTruncation_WhenASplicedValueWasClipped`. GPT explicitly recorded that it did *not* do mutation testing, so that claim rests on Gemini's reproduction and mine. It also flagged that it did not hand-dump raw CLI bytes — Gemini did, at budgets 5/10/50/1000 across md/tsv/jsonl against a hostile assembly carrying `U+202E`, `U+2028`, and ESC in both `AssemblyTitleAttribute` and a `#US` literal: **zero** occurrences of `e2 80 ae`, `e2 80 a8`, or `1b`, with a positive control confirming the benign part of the payload still rendered.

Suites at this head, both reviewers: InertText 362, Metadata 1301, MetadataRendering 149, CLI 2008 — 0 failed. Full solution build 0 warnings, 0 errors. Both worktrees ended clean with no diff.

### Review ledger

| Round | Head | Gemini 3.1 Pro | GPT-5.6 Sol |
| --- | --- | --- | --- |
| 1 | `5f5b135b5` | 1 blocking (equality regression via `Join`) | 1 blocking (+ handler site) |
| 2 | `3edf70099` | clean | clean |

Two families, two rounds, high-risk tier. CI green at the reviewed head: `ci-required` pass, `test` pass, `decompiler-gates` pass, `build-net10` pass, `markdownlint` pass. `pack`, `csharp-diff-smoke`, and `il-diff-smoke` are path-gated and skipped — not evidence of anything, and this change touches none of their paths.

**Ready to merge.**

### Non-blocking follow-up, not part of this PR

The projection model still stores text and its truncation flag as two separate fields in three records — `HeapReference(..., Text, Preview, Truncated)`, `HandleRef(..., Display, DisplayTruncated)`, and `MetadataImageOverview(..., MetadataVersionTruncated)` — with 42 consumer sites between them. `InertString` is now exactly that pair, so those fields can be derived rather than stored, which would delete the parallel bookkeeping and make the flag impossible to drop on the way to a sink. Deliberately left out to keep this PR to one behavioral claim. Noting it here so it isn't independently rediscovered — this fix was already built twice in parallel.

richlander:
## Restack verification — head `3edf70099` → `852a57e79`

The stack was rebased onto current `main` and CI has run green.

The rebase **linearized** this branch: the `origin/main` merge commits from the earlier integrations are gone, leaving five commits. A `range-diff` is therefore not the right proof here, since every commit legitimately has a different parent. Tree identity is:

```
$ git worktree add --detach /tmp/verify-restack 3edf70099   # the twice-reviewed head
$ git merge origin/main
$ git diff HEAD 852a57e79
                                                             # empty
```

Replaying the reviewed head against current `main` reproduces the rebased head exactly, so the restack changed the base and nothing else. The content that carried two clean review rounds is the content on the branch.

CI: `ci-required` **pass**, with `build-net10`, `test (ubuntu-26.04, linux-x64)`, `decompiler-gates`, and `markdownlint` all passing. `csharp-diff-smoke`, `il-diff-smoke`, and `pack` report `skipping` — path-gated and terminal, not evidence.

Locally at the rebased head: `-t:Rebuild` **9 warnings, 0 errors** (baseline), `ILInspector.Metadata.Tests` 1358/0 failed, `DotnetInspector.MetadataRendering.Tests` 149/0, `InertText.Tests` 362/0. End-to-end containment re-proved against a freshly built hostile assembly — 0 raw hazard bytes across 12 outputs with positive controls. Details on #3687.

`MERGEABLE` / `CLEAN`. Still `Ready to merge`, and this is the bottom slice — merge this one first.

Issue #3672, opened 2026-08-01
Resolve conditional branch liveness from PDB sequence points


richlander:
Depends on #3668.

## Summary

Decide **which branch of a conditional directive is live** from PDB sequence points, so members
guarded by, or adjacent to, a structural `#if` become sliceable. #3668 recovers the cases decidable
from source text alone; this issue covers the rest.

**Key framing: sequence points identify a _branch_, not a _symbol_.** Reasoning in branches rather
than symbols removes the need to parse conditions, track symbol values, or special-case `#elif` —
and covers strictly more. See "Why branches, not symbols" below.

## Evidence that the PDB answers this

Fixture with a conditional in a body, around a whole member, and inside a signature, compiled twice
and read with `System.Reflection.Metadata`:

```csharp
public class C
{
    public int InBody()
    {
#if FEATURE
        int a = 1;      // line 6
        return a;       // line 7
#else
        int b = 2;      // line 9
        return b;       // line 10
#endif
    }

#if FEATURE
    public int Only() => 41;    // line 15
#endif

    public
#if FEATURE
        long            // line 19
#else
        int             // line 21
#endif
        SigSplit(int x) => x;   // line 24
}
```

Sequence point start lines:

| Member | `FEATURE` undefined | `FEATURE` defined |
| --- | --- | --- |
| `InBody` | 4, **9, 10**, 12 | 4, **6, 7**, 12 |
| `Only` (line 15) | *method absent* | 15 |
| `SigSplit` (signature 19–23, body 24) | 24 | 24 |

Three facts follow:

1. **Body branches are named exactly.** `InBody` reports 9,10 or 6,7 — never both, never neither.
2. **Conditional members resolve by presence.** `Only` simply is not in the PDB when undefined.
3. **Signature lines carry no sequence points in either build.** Lines 19–23 appear in neither.
   Sequence points exist for executable code only, so the PDB is structurally silent there.

Fact 3 is the real boundary of this technique, and it is small — see "Residual".

## Rule

For each conditional group:

- If **exactly one branch contains sequence points**, that branch is live; drop the others.
- If **no branch contains sequence points**, the PDB has no evidence. Fall back to #3668's
  structural rule: if every branch is brace-balanced the spans stay known, otherwise unknown.
- Never infer deadness from absence alone in a group where some other branch also has no sequence
  points — a live branch containing only fields or a `const` emits none.

That is the whole rule. It is a decidable check against recorded evidence, not a heuristic.

[...]

PR #3680, opened 2026-08-02
Recover the declaration index's depth after a brace-balanced conditional group

### LINKED ISSUE #3668: DeclarationIndex: recover structure across brace-balanced conditional regions (lexical, no PDB)
## Problem

`DeclarationIndex` marks **every** declaration from the first conditional directive to the end of
the file `SpanKnown=false`, even declarations that sit entirely outside the conditional region and
even when every branch is brace-balanced.

The scanner sets `LexState.Untracked` on any `#if`/`#elif`/`#else` and never clears it
(`BodySlicer.cs`, `ScanLine`'s directive branch). Every later token is stamped `DepthKnown=false`,
and `DeclarationIndexBuilder` propagates that into `SpanKnown`.

Minimal repro — `Always` is unresolvable only because of an unrelated, brace-balanced `#if`:

```csharp
class C
{
#if DEBUG
    void Debug() { }
#endif
    void Always() { }
}
```

```text
Method Always sig=6..6 body=6 end=6 known=False
FindByBodyLine(6) -> <null>
without the #if, FindByBodyLine(3) -> Always
```

Pinned by `DeclarationIndexTests.AConditionalDirective_LosesEveryLaterRowToEndOfFile`.

## Scope of this issue

**The lexical half only: recover structure where it is decidable from source text alone, with no
PDB and no knowledge of which symbols are defined.**

Deciding *which branch compiles* needs the PDB and belongs in the layer above. That is #3672, which
depends on this issue. The split is deliberate: this half needs no new inputs, no new layer
dependency, and delivers half the total win on its own.

## Measured gain

Over `dotnet/runtime/src/libraries` — 20,817 files, 501,649 members, 4,913 directive groups:

| Tier | Members resolved | Gain |
| --- | --- | --- |
| Today | 439,858 (87.7%) | — |
| **This issue (lexical)** | **465,151 (92.7%)** | **+25,293** |
| With #3672 (PDB) | 499,646 (99.6%) | +34,495 more |

Directive group shapes, which is what determines the ceiling:

| Shape | Groups | Handled here? |
| --- | --- | --- |
| body-only, brace-balanced | 3,475 (70.7%) | **yes** |
| structural, one symbol | 1,114 (22.7%) | no — #3672 |
| structural, `!SYMBOL` | 137 (2.8%) | no — #3672 |
| structural, complex expression | 141 (2.9%) | no — #3672 |
| body-only, unbalanced | 37 (0.8%) | no — genuinely undecidable |
| structural, contains `#elif` | 9 (0.2%) | no — #3672 |

**70.7% of all conditional groups never affect declaration structure at all.** They sit inside a
method body, and today they cost the entire rest of the file.

## Design

A conditional group compiles exactly one branch, or none. Let `d0` be the brace depth at the `#if`.
If **every** branch is individually brace-balanced (net depth change zero, measured from `d0`), then
depth after `#endif` is `d0` regardless of which branch the compiler keeps — so the depth is
knowable and rows after `#endif` need not be poisoned.

Sketch:

- Maintain a stack of conditional frames; push on `#if` recording `d0`.
- On `#elif`/`#else`: if depth != `d0`, mark the frame unbalanced; reset depth to `d0` so one
  branch's braces do not accumulate into the next.
- On `#endif`: if depth != `d0`, mark unbalanced; pop and restore `d0`. If the frame stayed
  balanced, knownness is restored; otherwise it stays lost.
- Propagate an unbalanced inner frame to its enclosing frame.

Note this is *branch-balance*, not whole-group balance: a group whose branches individually balance
is safe even though the group as scanned counts every branch's braces.

`DeclarationIndexBuilder` never reads `ScanToken.Depth` — it tracks its own brace structure and
consumes only `DepthKnown`. So this can live in the builder, given a way to distinguish
"unknown because conditional" from "unknown because unterminated literal"; today `DepthKnown` is a
single bool carrying both. Adding that distinction to the internal `ScanToken` leaves
`BodySlicer.Untracked` semantics — and `ExtractMethodBody`'s behavior — unchanged.

### Also: expose conditional regions as data

This is the seam #3672 needs, and it belongs here because it is a lexical fact. The index should
report the conditional groups it saw — their line ranges and branch boundaries — so the layer above
can intersect them with PDB sequence points without re-lexing. The index itself stays PDB-free and
liveness-free; it reports structure and says which text is conditional, never which text is live.

## Gating

The current corpus differential **cannot** see any of this: `RoslynDeclarations` returns null for
any file carrying a conditional directive, because Roslyn (parsing with no symbols defined) discards
disabled branches while the index, being lexical, indexes them. Neither is wrong, but they are not
comparable as equality.

A fair gate compares as a **subset**: every declaration Roslyn reports in a conditional file must
appear in the index with the same span and `SpanKnown=true`. Extra index rows from disabled branches
are expected and are not failures. That gate would have caught this defect, and would cover the
1,720 conditional files the current gate declines.

## Residual after this issue

Declarations still unresolved: structural groups (rules needing the PDB, #3672) and the 37
body-only groups whose branches do not balance. The latter are genuinely undecidable without knowing
the symbol set — a brace that opens in one branch and closes in another.

## Context

Found by adversarial review round 9 of #3652 (GPT-5.6 Terra). Deferred from that PR by decision:
the behavior is conservative and safe, and #3652's index is not yet consumed.

### Measurement method

Roslyn enumerates directive groups (so `#if` inside a comment or string is not counted) and
`DeclarationIndex` supplies the declaration inventory, so the figures line up with the 12.1%
baseline. Loss is modelled as running from the first unhandled group to end of file, matching
today's semantics — so these numbers are **conservative**; localising the loss to the group scores
higher. Brace balance in the measurement used a crude comment- and string-aware character count,
adequate for a distribution but not for the product.

richlander:
## What changes

A conditional directive no longer costs every later declaration in the file. `LexState.Untracked` was set by any `#if`/`#elif`/`#else` and never cleared, so from the first conditional to end of file every `DeclarationSpan` reported `SpanKnown=false` — including rows after the `#endif` that no branch could affect, and even when every branch was brace-balanced.

Measured over `dotnet/runtime/src/libraries` (20,838 files, 581,298 declarations, 1,737 with a conditional):

| | Declarations `SpanKnown=false` | |
| --- | --- | --- |
| Before | 70,474 | 12.12% |
| After | 6,511 | **1.12%** |

**63,963 declarations recovered.** The baseline reproduces #3652's round-9 figure (70,472) on the same corpus, which is what makes the two numbers comparable.

## The rule

A conditional group compiles exactly one branch or none. Let `d0` be the brace depth at the `#if`. If every branch individually returns to `d0`, the depth after the `#endif` is `d0` whichever branch was taken, so the scan can go on vouching for it.

That is decidable from evidence already held — no condition parsing, no symbol tracking, no heuristic, and no `#elif` special case (`#elif` ends the branch above it and starts another, exactly as `#else` does).

`LexState` gains a stack of conditional frames. `#if` pushes `d0`; a branch boundary raises an unbalanced flag if the branch did not return to `d0` and then resets the depth to `d0`; `#endif` marks, pops, propagates a failure outward to the enclosing group, and restores `d0`. A stray `#else`/`#elif`/`#endif` with no group open loses the depth outright, which also keeps the frame stack in range.

**The inside of a group stays unknown**, and not only out of caution about which branch is live. A declaration that *crosses* a directive — `int P { get; }` with `= 1;` in one branch and `= 2;` in another — occupies a different, and in one reading non-contiguous, set of lines per build, which no single line range can express. At token granularity that is indistinguishable from a declaration written wholly inside one branch, so the conservative answer covers both. This is #3652's round-10 defect, and it still holds.

## Evidence

**The gate is a subset differential, because the equality one opts out by construction.** Roslyn with no symbols defined discards the disabled branches while the lexical index keeps them, so `EveryDeclarationRoslynReports_IsReportedIdenticallyByTheIndex` declines every conditional file — which is exactly why this defect survived nine review rounds with a green suite. `InAConditionalFile_EveryDeclarationOutsideTheConditionals_IsStillVouchedFor` restricts the comparison to a population where the two *are* comparable and weakens equality to containment: a declaration lying wholly outside every conditional region is present in every build, so Roslyn reports it and the index must report it identically and vouch for its span. Extra index rows from disabled branches are expected and are not failures. **Reverting the recovery turns it red.**

Its corpus is not the PDB-discovered one, which contains **zero** conditional files — this repository has five in fourteen hundred and none is in the test binary's dependency closure. It walks the repository instead, using an `#if` text search to *select* candidates and Roslyn to confirm them; that direction is sound, because a file with a real directive always contains the text. A floor asserts the *skipped* population is non-empty too, so a change that widened the skip until nothing was compared could not pass.

**Real-artifact canary.** Serilog 4.2.0's `Logger.cs` opens `#if FEATURE_ASYNCDISPOSABLE` at line 27, so today every one of its 100 declarations is withheld. After this change 93 are vouched for and the 7 withheld are the `IAsyncDisposable` members inside the groups.

**Ten mutations, both directions.** Eight killed. The battery found two rules ungated, and the dangerous one was the flag raised at a branch boundary: the depth reset erases the discrepancy before the `#endif` sees it, so deleting the flag left the suite green while turning every such group into a false `SpanKnown` — the one outcome the index may not produce. The first two fixtures written for it *passed against the deleted rule anyway*, because the open brace wrecked the row structure and every row then reported unknown for an unrelated reason; they balance the file now so a clean row survives past the `#endif`.

Two survivors are recorded rather than gated, per `AGENTS.md`:

- **The branch depth reset is equivalent** given the flag, which condemns the group whatever the counter goes on to say. The doc comment claiming the reset was what made multi-branch groups work was false and is corrected; it is kept for the invariant, not the answer.
- **`Clone` copying the conditional frames is unobservable** from either probe site — both are backward-scan probes reading `Untracked`, `InLiteral`, `InBlockComment` and `BracketDepth`, and neither emits tokens. Labelled unverified and ungated rather than left reading as a rule.

## Boundary

- **`LexState.Untracked` keeps its old sticky meaning**, so `ExtractMethodBody`'s backward scan is byte-for-byte unaffected. Only `ScanToken.DepthKnown` — whose consumers are `DeclarationIndexBuilder` and tests — carries the sharper property. There is no user-visible output change in this PR.
- **PDB branch liveness is not here.** The index reports what is *conditional*, a lexical fact; the layer above, which holds the PDB, decides what is *live*. That is #3672, and it takes the residual from 1.12% toward zero. Putting liveness in the index would reinvert what #3323 set out to fix.
- **The conditional-region data seam is deferred to #3672**, where it has a consumer. Adding API with no consumer is what this repository's guidance tells me not to do.
- **The absence-reason misattribution is not fixed here.** The CLI renders one message for every null slice, so a withheld span claims the member is "declared by its type's header". That null is unreachable on this branch, since `ExtractMethodBody` still uses the backward scan — it becomes reachable, and testable, only once the slicer is wired onto the index, so the fix belongs to that PR.

## Validation

`DotnetInspector.CSharpBodySlicer.Tests` 375/0. `DotnetInspector.Services.Tests` 357/0. `ILInspector.Metadata.Tests` 1283/0. `dotnet-inspect.Tests` 2817 with the one recorded `Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake. `dotnet-inspect.slnx` builds clean in Release.

Fixes #3668


richlander:

Head `95a62a470`. **GPT-5.6 Sol: BLOCK. Gemini 3.1 Pro: BLOCK.** Two-model tier
(this change is high risk: it converts a conservative refusal into a positive
safety claim about spans, and the index is about to become the slicing
substrate). Both reviewers ran in isolated worktrees at a fixed head.

Five findings. **Three are real product defects, one is a real docs defect, one
is dismissed by measurement.** Fixed in `a205670a3`.

| # | Reviewer | Finding | Sev | Outcome |
| 1 | GPT | a `#if` hidden in a comment left a group open, so a later `#endif` closed the wrong one and **vouched a wrong span** | High | fixed |
| 2 | GPT + Gemini | a branch closing a scope its group did not open changes the vouched **parent** between builds | Med | fixed |
| 3 | GPT | `LoseDepth`'s `Untracked = true` survives mutation | Low | **real, and not a dead store** — gated |
| 4 | GPT + Gemini | `SpanKnown` / `DepthKnown` still document poison-to-EOF and cite a deleted test | Low | fixed |
| 5 | Gemini | the `depth != BaseDepth` check in `CloseConditional` is ungated | Med | **dismissed by measurement** |

### Finding 1 — disabled text is not lexed as code

The one that matters, because it produces a **wrong** span rather than losing a
row, which is the exact property this PR claims to establish.

In a branch the compiler drops, `/*` opens no comment and a quote opens no
string — but directives are still recognized and still nest. So a `#if` this
scan skipped because it believed itself inside a comment left a group open, and
a later `#endif` closed the wrong group and restored knownness early. GPT's
fixture vouched `class C` as ending line 8 where the real answer is line 10.

A conditional directive found inside a comment or literal now loses the depth —
but **only while a group is open**. Outside one the text cannot be disabled, so
`#if` in a comment is unambiguously prose, and this repository's own sources
write it that way about nine times in `ScanTokenTests.cs`. Gating that negative
case is part of the fix.

### Finding 2 — equal depth is not an equal enclosing declaration

Found independently from opposite ends: GPT from the rule's statement, Gemini
from a fixture (`class D { #if DEBUG } class E { #else } class F { #endif`).

Balance is measured in depth, and depth is not identity. A branch that closes a
brace its group did not open is closing a scope from **outside** the group, so
two branches can agree on the depth after the `#endif` while disagreeing about
which declaration encloses it. A group may now not reach below its own opening
depth.

Cost on `dotnet/runtime/src/libraries`: **1.12% → 1.14% unknown**, i.e. 122
declarations of 581,298. The headline gain is unchanged.

### Finding 3 — the reviewer was right that it was ungated, and wrong that it was dead

`LoseDepth` sets both `literalDepthLost` (read by the index) and `Untracked`
(read only by the legacy backward scan). Disabling `Untracked` left all 375
tests green, so nothing distinguished it from a dead store.

My own first probe could not reproduce a difference, so I asked GPT for its
exact reproduction rather than guessing — and the detail I had wrong is the
whole finding. A constructor spelled `public C(int x)` is recognized as a
declaration on sight, so backward scanning never reaches the literal. A bare
`C()` requested as `.ctor` forces recovery **through** that line, and there
`Untracked` is what tells the scan its brace depth is meaningless. Without it
the slicer returns `null` instead of the captured text.

Now gated by `AConstructorRecoveredPastAnUnterminatedLiteral_StillCapturesItsText`,
and both assignments in `LoseDepth` are mutation-killed by distinct tests.

### Finding 5 — dismissed, with the mechanism named

Gemini mutated the check to `group.Unbalanced || false` and saw green. A
runtime-opaque mutation of the same term fails 4 tests. I believe Gemini hit the
stale-assembly hazard that this repository has recorded before: `dotnet run`
reuses a stale build unless you `dotnet build` and then run `--no-build`.

Its **narrower** point held and is worth recording: the first of the three
fixtures in `AnUnbalancedConditional_StillLosesEveryLaterRow` passes for another
reason, because both branches deviate identically, so the flag rule catches it
before the closing check is consulted.

### Also fixed

A duplicated `<summary>` tag on `StructuralDepthKnown`, left by one of my own
edits (Gemini). And a second copy of a doc comment claiming the per-branch depth
reset is "essential" — it is **equivalent**, established by case analysis and
probe before this review, and this was the second place it was written down.

### Mutation

Both new rules, in both directions, plus both assignments in `LoseDepth`:

| Mutation | Result |
| --- | --- |
| directive-in-comment loss removed | killed |
| loss applies with no group open | killed (3, incl. both corpus differentials) |
| opening-depth floor removed | killed |
| floor uses `<=` rather than `<` | killed (5) |
| `NoteDepth` call removed | killed |
| `Untracked = true` disabled | killed |
| `literalDepthLost = true` disabled | killed (5) |

Two of my four review fixtures initially passed **against the deleted rule
anyway** — an open brace wrecked the row structure, so every row reported
unknown for an unrelated reason. Rebalanced until a clean row survives past the
`#endif`.

### Validation at `a205670a3`

`dotnet-inspect.slnx` Release clean. Slicer **378/0**, Services **357/0**,
Metadata **1283/0**, CLI **2817/1** — the 1 being the recorded
`Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake, which fails
identically on a clean `main`.

Round 2 will run at this head once CI is green.

richlander:
## Adversarial review round 2 — reconciliation

Head `a205670a3`. **GPT-5.6 Sol: BLOCK. Gemini 3.1 Pro: BLOCK.** Fresh worktrees,
fixed head. **Six findings, every one real and reproduced.** Fixed in `1065113cb`.

| # | Reviewer | Finding | Sev | Outcome |
| 1 | GPT | a doc comment inside a group gives a vouched row a **branch-dependent trivia start** | High | fixed |
| 2 | GPT + Gemini | **file-scoped namespaces** open a scope with no brace, so neither the balance rule nor the floor sees them | High | fixed |
| 3 | Gemini | the round-1 ambiguity rule asks *whether* a line is a directive, not *which* — `#pragma` in a literal poisons the file | High | fixed |
| 4 | GPT | the `InLiteral` arm of that rule is **ungated**; only the comment arm has a fixture | Low | fixed |
| 5 | GPT | `SpanKnown`'s XML is malformed — a stray `</para>` | Low | fixed |
| 6 | Gemini | three doc comments assert properties without naming a gate | Med | fixed |

Round 1 fixed three defects and **round 2 found three more of the same kind**,
two of them in code round 1 had just written. That is the shape of this problem:
"the branches agree on brace depth" keeps turning out to be weaker than "the
branches mean the same thing".

### Findings 1 and 2 — two more ways to agree on depth and disagree on meaning

`SpanKnown` is documented as a claim about the row's **lines**, and trivia opens
a row's span. But comment and attribute-list tokens never reach the
pending-signature list, so **neither** `SpanKnown` expression consulted them:

```csharp
#if X
/// X docs
#else
/// Y docs
#endif
class C { }        // vouched, trivia=2 — a line only the X build compiles
```

Only the token that *opens* the trivia matters: a later comment inside a group
cannot move the recorded start, and every line it occupies already falls inside
the row's range. Attribute lists are the same defect through a second door.

A **file-scoped namespace** is the one scope opener in C# that uses no brace:

```csharp
#if X
namespace B;
#else
namespace C;
#endif
class D { }        // vouched with parent=C — but it is B in the X build
```

The group opens and closes at depth 0, no branch drops below the floor, so it is
judged balanced — while the enclosing declaration of everything below differs by
build. The scope runs to end of file, so no `#endif` repairs it. The rest of the
file is now refused. This is why the rule is stated as a **refusal** rather than
a depth correction.

I had written that "where the branches of a group disagree about which
declaration encloses the text after the `#endif`, the group is refused outright
— that is what the opening-depth floor is for." Finding 2 is a direct
counterexample to my own sentence, written one round earlier.

### Finding 3 — my round-1 fix over-refused

The ambiguity guard called `IsDirective(line, out _)`, discarding which
directive it found. A skipped section is the one place the compiler does **not**
process `#pragma`, `#region`, `#nullable` or `#line` at all, so such a line
inside a literal is text whichever way the branch falls and cannot open, close
or renumber a group. Refusing it poisoned a whole file for a directive that
changes nothing. The rule now tests the kind — which is both narrower and more
faithful to what the compiler does.

### Cost, and the headline

| | unknown | % |
| --- | --- | --- |
| Pre-recovery baseline | 70,474 | 12.12 |
| Round 1 | 6,633 | 1.14 |
| **Round 2 (this head)** | **7,208** | **1.24** |

Round 2's refusals cost 575 declarations of 581,298; finding 3's relaxation gives
some back. The branch still moves **12.12% → 1.24%**. Serilog 4.2.0 `Logger.cs`
unchanged at **93 of 100** vouched, `Logger.Write` resolving.

### Mutation

Ten mutations across the new rules, both directions. **Nine killed first time;
one survived**, and it is the same lesson as round 1: both trivia fixtures used
`class C { }`, so they only reached the braced emit site, leaving the *bodiless*
site's own `SpanKnown` expression ungated. A field fixture kills it. All ten now
killed by a named test.

### Note on finding 6, and what is deliberately not fixed

Two claims now name gates. `NextBranch`'s depth reset does not, and is now marked
**unverified and ungated** — it is an equivalent mutation, since any deviating
branch raises the unbalanced flag in the same breath. Marking it is the honest
answer under `AGENTS.md`, not gating it.

Enabling `GenerateDocumentationFile` on this project surfaces **pre-existing**
errors on `main` (`cb7b422f33`), including a second orphaned `<summary>` above
`IsDirective`. Those are unrelated to this change and are left alone; worth a
follow-up issue rather than a mixed commit.

### Validation at `1065113cb`

`dotnet-inspect.slnx` Release clean, 0 warnings. Slicer **383/0**, Services
**357/0**, Metadata **1283/0**, MetadataRendering **149/0**, CLI **2817/1** —
the 1 being the recorded `Layout_Count_CountsFilesRatherThanRenderedTreeLines`
flake.

Round 3 will run at this head once CI is green. Given that rounds 1 and 2 each
found real defects in the previous round's fixes, I am not treating round 2 as
the last one on the strength of a clean local suite.

richlander:
## Adversarial review round 3 — reconciliation

Head `1065113cb`. **GPT-5.6 Sol: BLOCK. Gemini 3.1 Pro: BLOCK.** Seven findings.
Fixed in `02fe35235`.

**The important result: neither reviewer reproduced a false `SpanKnown` from the
round-2 rules.** GPT states it explicitly ("I reproduced no current false
`SpanKnown`"). Rounds 1 and 2 each found High-severity wrong-span defects; round
3 found none. What remained were two narrower over-vouches, three ungated rules,
and three false doc comments.

| # | Reviewer | Finding | Sev | Outcome |
| 1 | Gemini | `#endif_foo` read as `#endif`, closing a group Roslyn leaves open | Med | fixed |
| 2 | Gemini | a conditional attribute list **after** an unconditional one is reported with no knownness signal | Med | fixed |
| 3 | GPT | the namespace refusal's start index is **ungated** (a two-namespace fixture masks an off-by-one) | Low | fixed |
| 4 | GPT | the unit-attribute trivia reset is **ungated** | Low | fixed |
| 5 | GPT | the `conditionalDepthLost` gate citation is **false** | Low | fixed |
| 6 | GPT + Gemini | two more stale/imprecise doc comments | Low | fixed |
| 7 | Gemini | a conditional comment after an unconditional one over-vouches | High | **dismissed by measurement** |

### Finding 1 — a directive name is an identifier

`#endif_foo` spells the single identifier `endif_foo`, which is not the `#endif`
directive. Roslyn reports **CS1024 and CS1027 in every symbol configuration** and
leaves the group open; we closed it, recovered the depth, and vouched for what
followed. `char.IsLetterOrDigit` misses underscore, and that is the entire gap —
I checked the neighbours rather than tightening blindly, because over-tightening
here costs real recovery:

| closer | Roslyn | now |
| --- | --- | --- |
| `#endif` | accepted | closes |
| `#endif//note` | accepted | closes |
| `#endif /* note */` | accepted | closes |
| `#endif-` | CS1025, still closes | closes |
| `#endif_foo` | **CS1024/CS1027, stays open** | **stays open** |

All five are asserted.

### Finding 2 vs finding 7 — why one is fixed and the other is not

They look like the same shape and are not. I rendered both builds with line
numbers preserved:

```text
                    trivia  end   known  attrs
attrs  TRUE on         1      5    True   [1..1, 3..3]
attrs  TRUE off        1      5    True   [1..1]        <- differs
comments TRUE on       1      5    True   []
comments TRUE off      1      5    True   []            <- identical
```

A conditional **comment** below an unconditional one changes no line the row
reports; `SpanKnown` is a claim about the row's lines, so refusing would cost
recall for nothing. **Finding 7 dismissed**, and pinned by a test that says so.

A conditional **attribute list** is different in kind: it is not merely a line
inside the range, it is a claim about what is *applied to* the declaration, and
it was reported identically in both builds with no knownness signal. Knownness
is now intersected over every list, not taken from the one that opened the
trivia.

### Finding 5 — I have now cited the same test falsely twice

`conditionalDepthLost` claimed to be gated by
`AnUnbalancedConditional_StillLosesEveryLaterRow`. Round 3 showed it is not. My
round-3 *fix* was to add a fourth fixture to that test — and the stickiness
mutation was killed by the **hidden-directive** tests, not by the new fixture.
So the replacement citation was false too.

It fails for the reason that test's own comment warns about: an unbalanced group
mangles the brace structure enough that its rows are unknown for other reasons,
so the fixture passes whether or not a balanced close clears the flag. The two
hidden-directive tests are the real gate — each sets the flag inside a group that
then closes balanced. Cited correctly now, and the added fixture is labelled as
**coverage, not a gate**.

That makes twelve false doc comments in this program, and this one was written
while fixing a finding about doc comments that assert properties without gates.

### Cost

| | unknown | % |
| --- | --- | --- |
| Pre-recovery baseline | 70,474 | 12.12 |
| Round 1 | 6,633 | 1.14 |
| Round 2 | 7,208 | 1.24 |
| **Round 3 (this head)** | **7,531** | **1.30** |

Still **12.12% → 1.30%**. Serilog `Logger.cs` unchanged at **93 of 100**.

### Mutation

Seven across the new rules, both directions, all killed by a named test —
including both directions of the underscore rule (removing it, and over-refusing
with it).

### Validation at `02fe35235`

`dotnet-inspect.slnx` Release clean. `GenerateDocumentationFile` reports no
errors in `DeclarationIndex.cs`. Slicer **388/0**, Services **357/0**, Metadata
**1283/0**, CLI **2817/1** — the recorded `Layout_Count` flake.

Round 4 at this head once CI is green. The trend is the right one — High
severity in rounds 1 and 2, none in round 3 — but three rounds running have
found real defects in the previous round's fixes, and round 3 found two of my
own gate citations false, so I am not calling it on a green suite.

richlander:
## Adversarial review round 4 — reconciliation

Head `fcaf16253` (round-3 fixes plus an `origin/main` integration merge).
**GPT-5.6 Terra: BLOCK. Gemini 3.1 Pro: BLOCK.** Two findings — one real and
High, one declined on its premise. Fixed in `3c28d1aa6`.

The brief asked the reviewers the one open question directly: **is there a fifth
way for the branches to agree on brace depth and disagree on meaning?** There
is, and GPT found it.

### GPT — a conditional attribute *target* (High, fixed)

```csharp
[
#if X
assembly:
#endif
System.CLSCompliant(true)]
class C { }
```

| | Roslyn | index |
| --- | --- | --- |
| `X` **off** | `C` spans **1..6**, one attribute list 1..5 | — |
| `X` **on** | `C` spans **6..6**, no attributes; the list is file-level | — |
| reported | | `trivia=6 sig=6..6 attrs=[] **known=True**` |

Confirmed against Roslyn in both symbol configurations before touching anything.
A wrong span, vouched for — the outcome the contract forbids.

Round 3 already refuses a branch-dependent `AttributeLists` set. This slips past
it because the list **crosses** the group rather than sitting inside it:
knownness was sampled at the `[`, which is on line 1, outside the group, and
known. The tokens that decide whether the list binds to `C` at all — `assembly`
and `:` — are inside it.

Knownness is now intersected over every token of the list. Both `SpanKnown` emit
sites are gated, with a field fixture for the bodiless one, because a round-2 fix
shipped with that site ungated for exactly this reason.

**Cost: one declaration.** 7,531 → 7,532 unknown of 581,298 on
`dotnet/runtime/src/libraries`. Serilog `Logger.cs` unchanged at 93 of 100.

### Gemini — directives inside an interpolated string hole (declined)

Gemini reported the hidden-directive guard as over-refusing inside an
interpolation hole, and proposed **relaxing the guard** to `state.InLiteral &&
!state.Top.InHole`, on the premise that

> the compiler processes conditional directives inside a hole as directives,
> regardless of whether the surrounding branch is active or dropped.

That premise is false. Roslyn on Gemini's own fixture:

```text
symbols []      -> no errors                (the # lines ARE directives)
symbols [A]     -> CS1073 Unexpected token '#'
                   CS1056 Unexpected character '#'  (x2)
symbols [A,B]   -> same
```

A `#` inside an interpolation hole is not a directive at all — it is a lexical
error. So the two builds disagree about whether that text is a directive, which
is precisely the asymmetry the guard exists for. The proposed relaxation would
vouch based on one build's reading, and it is also ungated: it "passes all 388
tests" because no test covers it.

Declined. The row loss is real but is the safe side of a conservative contract.

### Two of my own gates failed their own mutation test

Six mutations, both directions. On the first pass **two survived**:

- `AConditionalCommentInsideAnAttributeList_LosesTheRowBelowIt` gated nothing.
  Its fixture also contained `assembly:`, so the punctuator path killed the
  mutation and the comment token was never load-bearing. **Deleted.**
- The `[` seed is subsumed by the accumulation: the closing `]` is accumulated
  too, so for the seed to decide the outcome the group would have to close
  between the `[` and the rest of the list — a file whose other build has no `[`
  at all. Kept as the conservative initialization and **marked unverified**,
  following the `NextBranch` precedent, rather than given a citation it cannot
  support.

Replacing the deleted test with a *literal* inside a crossing list then
**falsified my own fix**: it failed, because my first placement consumed comment
and literal tokens too, and both builds of that fixture report identical lines.
That is an over-refusal by the same line-scoped principle that dismissed round
3's comment finding. The placement is now narrowed, and both directions are
killed — M1 (no-op) by the target tests, M2 (broaden) by the literal test.

Final: 5 of 6 killed, M6 documented as subsumed.

### Validation at `3c28d1aa6`

`dotnet-inspect.slnx` Release, 0 errors. Slicer **392/0**, Services **357/0**,
Metadata **1314/0**, MetadataRendering **149/0**, CLI **2817/1** — the recorded
`Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake, which fails
identically on clean `main`.

Round 5 at this head once CI is green. Four of six permitted rounds used.

richlander:
## Round 5 reconciliation — `f4f4d82d7` → `8c57a5906`

GPT-5.6 Sol and Gemini 3.1 Pro, both **BLOCK**. Six findings real, one dismissed by measurement.

### There is a sixth way

The brief asked both reviewers the one open question: is there a **sixth** way for two branches to agree on brace depth and disagree on meaning? GPT found one.

**A terminator in one branch discarding trivia recorded in another.** `ResetHeader` cleared the recorded trivia start and restored `triviaKnown = true` unconditionally. So a comment that documents the next declaration in one build and documents nothing in the other was vouched for. `ResetHeader` now takes the discarding token's `DepthKnown`, and the comment handler intersects (`&=`) rather than assigns, so a later record on a known line cannot overwrite the poison.

That is the sixth consecutive round in which the reviewers found a wrong span. Rounds 3, 4 and 5 each falsified the previous round's read that the rule set had converged; I am no longer treating "no reviewer reproduced a wrong span" as evidence of convergence.

### Findings

| # | Reviewer | Finding | Sev | Outcome |
| 1 | GPT | a terminator in one branch discards trivia recorded in another | High | fixed |
| 2 | GPT | a UTF-8 BOM before `#if` hid the group entirely | High | fixed |
| 3 | GPT | `#line` remapping breaks PDB-line ↔ row correspondence | Med | documented limitation |
| 4 | GPT | the corpus gate skipped exactly the rows recovery protects | Med | fixed |
| 5 | GPT | `IsIdentifierPart` missed Unicode Mn/Mc/Pc/Cf | Med | fixed |
| 6 | GPT | some over-refusals remain | Low | accepted |
| 7 | Gemini | a multi-line literal in a group renumbers later lines | **Critical** | **dismissed by measurement** |
| 8 | Gemini | the `conditionalDepthLost` citation is false | Med | fixed |
| 9 | Gemini | `NoteDepth` / `CloseConditional` name no gate | Low | fixed |

### Finding 2: `TrimStart()` does not remove U+FEFF

A UTF-8 byte order mark is not whitespace, so the BOM sat in front of the `#` of a file's first `#if` and the directive scanned as ordinary code — the whole group went unseen, in the one position (a file's first line) where a BOM actually occurs. `IsDirective` now strips a leading `\uFEFF` before the test.

### Finding 5, checked against Roslyn rather than assumed

A directive name is an **identifier**, and identifier continuation is more than `IsLetterOrDigit || '_'`. Verified in both symbol configurations: `#endif\u0301` (Mn), `#endif\u203F` (Pc) and `#endif\u200C` (Cf) are all **CS1024 + CS1027** — invalid, and the group stays open. We were closing it. The existing accepted spellings (`#endif`, `#endif//note`, `#endif /* note */`, `#endif-`) still close.

### Finding 4 is the round's methodological result

The corpus gate skipped every declaration that **overlaps** a conditional region. That is precisely the set of enclosing rows whose recovered closing line the whole change exists to produce — so a mutation reporting a containing type's end one line short passed the gate that was written to catch it. The gate now skips only declarations that **start** inside a region and compares the containing ones, with a `containingVouched` non-vacuity floor, plus a fixture pinning the same property. Both verified by mutating the closing-line fix.

### Finding 7 dismissed by measurement

Gemini's Critical claim was that a multi-line string literal inside a group renumbers the lines below it. **Directives do not renumber physical lines.** Rendering both builds of its own fixture reports identical line numbers for every declaration below. `SpanKnown` is a claim about the row's lines, so there is nothing to refuse.

### Finding 3: a real limitation, documented rather than fixed here

Measured against a real build: `#line 500 "elsewhere.cs"` above a method makes **every** one of its sequence points read `elsewhere.cs:500`, while a physical-line index reports the line it occupies. `DeclarationSpan`'s doc comment claimed its lines "match portable-PDB sequence points", which is false in that presence — so this was a doc defect as well as a correspondence one.

It is **not affected by conditional recovery**: recovery changes which rows are vouched for, never which lines they name. And it is rare — 4 of 20,838 files in `dotnet/runtime/src/libraries` carry a renumbering `#line`, and **none of those four is a conditional file**, so this change does not widen the exposure. Recorded on the type as a known, ungated limitation belonging to whichever layer performs the PDB-to-row match, rather than silently fixed in a PR about something else.

### The third consecutive false citation of one field

`conditionalDepthLost` has now had three successive doc comments claiming a gate it does not have. Round 3 found the original false; **the correction I wrote for it was false in turn.** `AnUnbalancedConditional_StillLosesEveryLaterRow` gates the **assignment**; the **stickiness** is gated by the two hidden-directive tests. Both halves now name their own gate separately, and the comment records that the correction was itself falsified, because that is the part a future reader needs.

`NoteDepth` and the inner-propagation rule now cite `ABranchThatClosesAScopeItsGroupDidNotOpen_LosesTheDepth` and `AnUnbalancedInnerConditional_PoisonsTheGroupAroundIt`; both citations were verified by mutation rather than by reading.

### Cost

| | unknown | % |
| --- | --- | --- |
| Pre-recovery baseline | 70,474 | 12.12 |
| Round 3 | 7,531 | 1.30 |
| Round 4 | 7,532 | 1.30 |
| **Round 5** (`8c57a5906`) | **7,602** | **1.31** |

71 declarations. Serilog `Logger.cs` unchanged at **93 of 100**, `Logger.Write` still resolving.

The `ResetHeader` poison is deliberately **broader than strictly necessary**: without branch identity on tokens it cannot distinguish "trivia and terminator in different branches" from "both in the same branch". That is the refuse-rather-than-guess side of the trade, and it is where most of the 71 went.

### Mutation

Eight across the new and corrected rules, all killed. **Two survived the first pass and both were informative** — one was a genuine reachable defect that needed the new intersection test written for it, and one exposed a malformed mutation of mine that changed no behavior and therefore asserted nothing.

### Validation at `8c57a5906`

`dotnet-inspect.slnx` Release clean. Slicer **398/0**, Services **357/0**, Metadata **1339/0**, MetadataRendering **149/0**, CLI **2817/1** — the 1 being the recorded `Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake, which fails identically on a clean `main`.

richlander:
## Round 6 reconciliation — `8c57a5906` → `b8173e1fd`

GPT-5.6 Sol and Claude Opus 4.8, both **BLOCK**. Two more ways a vouched span can be wrong — and, more importantly, the first *mechanised* way of looking for the next one.

### The fuzzer

Opus did not hand-craft its finding. It built a **differential fuzzer**: generate a small C# file sprinkled with conditional groups, run `DeclarationIndex.Build` once, run Roslyn **four** times as an independent oracle (`{}`, `{X}`, `{Y}`, `{X,Y}`), and flag any row the index **vouches for** whose lines or attribute lists differ between two builds that both parse. Against the round-5 head it reported **3,146 flags** on a single seed.

Five rounds of hand fixtures had found five ways. One fuzzer run found two more in an afternoon. It is now **in the repository**: `ConditionalRecoveryFuzzTests` runs five fixed seeds at 4,000 cases each (~0.5s), and `eng/conditional-recovery-fuzz.cs` drives deep runs through **the same generator and oracle**, so a sweep and the gate cannot drift apart.

Its header documents the **fair-case** exclusion set, which is the part a future reader most easily gets wrong: a file counts only when at least two configurations parse cleanly, and a row is compared only where a config supplies exactly one match for its (Kind, Name). Every flag is a real over-vouch; **a clean run is evidence, not proof** — it cannot flag a shape the generator never spells, and the header lists what it does not spell.

### The seventh way — knownness was *assigned*, so a later clean reset restored it

```csharp
#if X
// doc
#else
struct s { }
#endif
class Tail { }
```

The comment is discarded by `struct s {`, which poisons. Then `}` resets **again** with nothing recorded, and the assignment declared the header clean — while still inside the group. With `X` there is no `struct s` to have eaten the comment, so it is `Tail`'s documentation: Roslyn reports trivia line **2** with `X` and line **6** without, both configurations parsing with zero errors.

Round 5's regression test exercised a *single* reset (`using System;`, no braces), so it never saw the second, `}`-driven one. Inside an unresolved group a reset may now only take knownness away, never restore it.

**GPT found a narrower instance of the same class independently**: a terminator in one branch discarding a bare `public` recorded in another — moving the row's **signature** start rather than its trivia start, which round 5's trivia-only rule could not see at all.

### Sections, and why not `DepthKnown`

Testing this with `DepthKnown` would condemn every declaration below every group containing a statement — tolerable while only trivia consulted it, catastrophic once signature tokens do. Tokens now carry a **`Section` ordinal**: a counter bumped at every `#if`, `#elif`, `#else` and `#endif`.

Two tokens sharing a value are **certainly** in the same compiled branch. Different values **may** still share one. So it is conservative in the safe direction only, and it answers the question the defect is actually about: was the header, and the terminator that ate it, written in the same branch? That is what distinguishes a header **crossing** a group from one written entirely inside a single branch — which brace depth alone cannot do.

### The eighth way — found by the fuzzer *after* the seventh was fixed

The `ResetHeader` fix cut the flag count from 3,146 to **6**, and every survivor was one site:

```csharp
#if Y
#else
[System.Obsolete]
#endif
#if X
class t1 { }
#endif
[assembly: System.CLSCompliant(true)]
class Tail { }
```

An `[assembly:]` list **legitimately** discharges a *trivia* poison — such a list ends the trivia run above it in every build, which is exactly what round 3's `AUnitAttributeAfterConditionalTrivia_StillVouchesForWhatFollows` pins. It must **not** discharge an unconsumed conditional **attribute list**, which still binds to whatever follows: without `Y`, Roslyn binds the line-3 list to `Tail` *and* the unit list with it (CS0657), so `Tail`'s trivia is line 3 and it carries two lists; with `Y` its trivia is line 9 and it carries none.

One flag was carrying both claims. My first two attempts here were wrong in opposite directions — intersecting broke round 3's guard, and a `DepthKnown` guard did nothing because the list sits *outside* the group. The flag is now split: **`triviaKnown`** for the trivia start line, **`headerKnown`** for lists, modifiers, and headers a branch-dependent terminator threw away.

### Also fixed

GPT showed the **initializer reset site was ungated** — neutralizing it left all 398 tests passing — while its doc comment claimed otherwise. That is the fourteenth false doc comment in this program.

The `#endif` section increment is recorded as **unverified, conservative and ungated**, with the argument for why no answer depends on it, rather than gated by a contrived fixture.

### Cost

| | unknown | % |
| --- | --- | --- |
| Pre-recovery baseline | 70,474 | 12.12 |
| Round 5 | 7,602 | 1.31 |
| **Round 6** (`b8173e1fd`) | **8,334** | **1.43** |

**732 declarations, all of which were vouched-for wrong answers.** Serilog `Logger.cs` 93 → **91** of 100.

### Validation at `b8173e1fd`

Slicer **410/0**, Services **357/0**, Metadata **1339/0**, MetadataRendering **149/0**, CLI **2817/1** (the recorded `Layout_Count` flake, identical on clean `main`). `dotnet-inspect.slnx` Release clean.

Eight mutations across the new rules, **all killed by the intended gate**. Reverting the stickiness fails **all five fuzz seeds** plus three hand fixtures, so the new gate is not vacuous. Fuzzer clean at **140,000 cases over seven 20,000-case seeds** plus a fresh 8,000-case sweep, against 3,146 flags on seed 12345 before the fix.

### Status

`AGENTS.md` permits six review rounds without explicit approval, and **all six are now used — and round 6 was not clean.** Holding here for direction rather than opening a seventh.

richlander:
## Adversarial review round 7 — reconciliation

Reviewed at `3f27e1fd9` by **Claude Opus 5** and **Gemini 3.1 Pro**, each in an isolated worktree. Both returned BLOCK. Fixed in `8e8491fff`.

Both reviewers independently found the same defect from opposite directions, which is the strongest signal this program has produced: Gemini by enumerating carried state and noticing `lastClosed` is written inside a group but read outside it; Opus by asking what its widened fuzzer could reach that the shipped one could not.

### Actioned

**The ninth way — an initializer reaching *back* through a group.** (Gemini Q1, Opus Finding 1.) Which declaration `= 1;` extends is branch-dependent:

```csharp
class C {
    public int P { get; }
#if X
    public int Q { get; }
#endif
    = 1;
}
```

Roslyn: `P` is lines 2-2 with `X`, 2-6 without; both configurations parse with zero errors. The product vouched for `P` at 2-2 unconditionally. Every rule refused in rounds 1-8 asks whether a header written *before* a group survives it; this is the reverse, so nothing in place could see it.

It turned out to have **three** shapes, only the first of which either reviewer reported:

1. The initializer follows the group (both reviewers).
2. The initializer sits *inside* the group and the sibling branch has already consumed the row it would extend, so `lastClosed` is `-1` and the existing guard is never entered (Opus repro B; found independently by the widened generator).
3. The `=` is not first in the pending run, because a modifier left behind by another branch precedes it — `#if X / public / #else / = 1; / #endif` (found only by the widened generator, after shapes 1 and 2 were fixed).

The vouch now survives only when the `=`, the `;` and the `}` that produced the target are provably in one branch. Otherwise every declaration under the same parent is a candidate — an `#elif` chain offers several — and the refusal takes the whole run rather than guessing which.

**The fuzzer's structural hole.** (Gemini Q2, Opus Finding 4.) The round-6 generator emitted every group at file scope, where the only rows Roslyn and the product both produce are types and namespaces. Across 140,000 clean cases it had **never once compared a Method, Property, Field, Event or EnumMember row** — and `EmitBodiless`, the enum terminator paths and the initializer path are all member-only code. A clean run was evidence about part of the product, presented as evidence about all of it. Opus's widened generator is adopted wholesale: groups inside type bodies, declarations split across group boundaries, nested groups, `#elif` chains, negated and compound conditions, directive text hidden in verbatim and raw literals, and `SignatureEndLine`/`BodyStartLine` added to the comparison.

**Two ungated load-bearing guards.** (Opus Findings 2 and 3.) Both `EmitBodiless` terms — the terminator's depth flag and the pending run's — could be deleted with the entire 417-test suite staying green, fuzzer included. Each now has a fixture, and each is killed by all five fuzz seeds.

**Two false doc comments.** (Opus Finding 5, Gemini Q4.) `SpanKnown`'s remarks named one test as covering "the only path that can report a wrong span" — the exact path fixed here, by two routes that never reach the named gate. Restated as what the differential actually enforces. Separately, `ResetHeader`'s trivia-poison discharge was ungated: all 410 tests passed with the assignment neutralised. Now gated. That makes **seventeen** false doc comments this program has caught by mutation.

### Dismissed, with evidence

**Gemini Q3** — that `AUnitAttributeAfterADiscardedAttribute_DoesNotRestoreTheVouch` passes for the wrong reason. The mutation used (deleting `headerKnown &= attributeKnown;`) makes the code *stricter*, not laxer: it leaves `headerKnown` untouched, so no test asserting a refusal could possibly fail. The mutation that actually reintroduces the eighth way is `headerKnown = true;`, and under it that test **does** fail, together with `AConditionalAttributeTarget_LosesTheRowBelowIt` and all five fuzz seeds. The gate is real.

Opus separately re-confirmed the `#endif` section increment (documented as unverified and ungated) as corroborated — clean across 5 seeds × 4,000 cases of the widened generator with it neutralised — and re-confirmed ways 1-8 live by mutation.

### Cost and validation

139 declarations of 581,298 over `dotnet/runtime`'s libraries — 1.43% → **1.46%** unknown, against 12.12% before recovery. Serilog's `Logger.cs` unchanged at 91 of 100.

Slicer **417/0** (was 410), Services 357/0, Metadata 1358/0, MetadataRendering 149/0, CLI 2817 with only the pre-existing `Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake (fails identically on clean `main`). Deep fuzz sweep of three fresh seeds at 20,000 cases each: 29,624 fair, 0 flags.

Non-vacuity: the pre-fix builder and both `EmitBodiless` mutants each fail **all five fuzz seeds plus their named gate**. Before this commit, both `EmitBodiless` mutants were entirely green.

Round 7 was explicitly approved after round 6 came back dirty. Rounds 3-7 have each falsified the previous round's convergence read, so I am not claiming convergence here either — but the shape of round 7 differs: the widened fuzzer now reaches the code the defects were hiding in, and the two independent reviewers converged on one defect rather than diverging onto new ones.

richlander:
## Round 8 reconciliation

Reviewers: **GPT-5.6 Sol** and **Gemini 3.1 Pro**, both at `b15c1e251` in isolated worktrees. Claude Opus 5 was deliberately excluded this round: it authored the widened fuzzer that round 7 shipped as a permanent gate, and asking it to attack its own artifact would have produced the weakest possible answer on the round's highest-value question.

Both returned **BLOCK**, and — as in round 7 — they converged on the same *family* while finding **two different defects in it**. Fixed in `0188922b3`.

### Actioned

**The tenth way — an initializer masked by another branch's initializer (GPT-5.6 Sol).** The scan took the first `=` in the pending run and disqualified it if any earlier token shared its section. A branch carrying its own complete declaration puts an ordinary `=` in front of the bare tail beside it, so the tail was never examined:

```csharp
int P { get; }
#if X
    int Q = 0
#else
    = 1
#endif
    ;
```

Verified independently against Roslyn: `P` is lines 3..3 with `X`, 3..9 without, **both zero errors**; the product reported `3..3 SpanKnown=True`. The scan now takes the first `=` that *qualifies* rather than the first that exists. Worth noting that the comment above this code already described the correct rule — the code simply did not implement it.

**The eleventh way — enum member initializers (Gemini 3.1 Pro).** The refusal was reachable only from the `;` path, and an enum member terminates with `,` or `}`. The identical reaching-back initializer ran unchecked through both enum emit sites, which had already emitted and vouched the preceding member before the `=` was read:

```csharp
enum E {
    A
#if X
    , B
#endif
    = 1
}
```

Verified independently: `A` is lines 2..2 with `X`, 2..6 without, both zero errors; product said `2..2 SpanKnown=True`.

Both fixes share one qualification test (`ReachingBackEquals`), and both cost the corpus **nothing** — unknown declarations stay at 8,473 / 581,298 (1.46%), Serilog `Logger.cs` stays at 91/100. That is the point of testing for a reaching-back `=` rather than for a conditional comma: the blunt rule passes the same tests and would have cost every conditionally-extended enum in the corpus. `AConditionalEnumMemberWithNoInitializer_StillVouchesForTheMemberBeforeIt` pins that distinction.

**Both fuzzer blind spots (Gemini 3.1 Pro, GPT-5.6 Sol independently).** Two structural limits, not case-count limits:

- `Group` spelled both branches from a single body, so **asymmetric branches were unreachable at any case count** — which is exactly where the tenth way lives. `Group2` spells them independently.
- The pools contained enums only as complete one-line declarations, so no enum interior was ever conditional. A 20,000-case sweep now compares **12 row kinds including 2,098 `EnumMember` rows, against zero before**.

**The missing half of the gate (Gemini 3.1 Pro).** `NoVouchedRowMovesBetweenBuilds` compares the *builds against each other* and never reads the product's own numbers, and is silent on any row present in only one build. Added `EveryVouchedRowMatchesRoslynInEveryBuildItExistsIn`, which checks the product's numbers against every build a row exists in. It was already implemented as `product` mode but ran only from `eng/`, never in CI.

### Dismissed, with evidence

Nothing was dismissed this round.

### Withdrawn by me

I wrote a mirror-ordering test whose comment claimed it pinned the *direction* of the `=` scan. The mutation that would prove that — taking the last qualifying `=` rather than the first — kills neither it nor its partner, because each source contains exactly one qualifying `=`. The test is kept as coverage and its comment now says so. Flagging it because an unearned "this gates X" comment is the failure mode this PR has already hit seventeen times.

### Evidence

- Slicer **428/0**; Services 360/0, Metadata 1397/0, MetadataRendering 187/0, CLI 2843 with only the pre-existing `Layout_Count` flake.
- Fresh-seed deep sweeps: **3 × 20,000 cases in each mode, 0 flags**.
- **Every widening mutation-proved independently.** Against the pre-round-8 builder, all five seeds fail in both modes. With *only* the tenth-way fix applied, the enum widening still fails all twelve selected tests — so neither widening is riding on the other.
- Confirmed the round-7 `EmitBodiless` gates are still honest (Gemini re-ran the mutation).

### Assessment

Rounds 3–8 have each falsified the previous round's convergence read, and round 8 falsified the specific hope that round 7's widening had made the fuzzer complete. The pattern is now unambiguous: **every round that has attacked the generator has found something, and the generator's limits have been structural every time — never a matter of case count.** Two rounds running, the fuzzer's blind spot has been the defect's hiding place.

That is an argument for one more round rather than against it, but the decision is @richlander's; round 8 was the approved scope.

richlander:
## Round 9 reconciliation — `0188922b3` → `353e9a049`

Reviewers: **Claude Opus 5** (`/home/rich/rr9-a`) and **GPT-5.6 Sol** (`/home/rich/rr9-b`), both at `0188922b3`. Both **BLOCK**. Every finding actioned; none dismissed.

The brief made one thing Priority 1: **attack the fuzzer's reach, not its case count.** That is now three rounds running where doing so found something.

### GPT-5.6 Sol — two harness findings, both real (fixed in `0eca4554a`)

| # | Finding | Sev |
| --- | --- | --- |
| 1 | `DifferFromProduct` compared 4 of 6 span fields, omitting `SignatureEndLine` and `BodyStartLine` | High |
| 2 | Constructors were never compared — zero across 50,000 cases | High |

**Finding 1 is a gate I added in round 8 vouching for fields it never read.** Verified by mutation: setting `BodyStartLine = -123` on every product row left all five product seeds green.

**Finding 2 has a subtle mechanism worth recording.** Adding an unsplit constructor to the pool changed nothing, because rows are matched by `(Kind, Name)` and a constructor whose name differs from its enclosing type is a `Method` to the product but still a `ConstructorDeclaration` to Roslyn's parser. The mismatch produces **zero matches and a silent skip, not a flag** — and the case still counts as "fair", because the harness reads *parse* diagnostics and CS1520 is bind-time. Fixed by plumbing `curType` so generated constructors are named for their type.

Constructor comparisons went 0 → 202, Destructor 0 → 230, EnumMember 0 → 1,817.

### Claude Opus 5 — the twelfth way (fixed in `353e9a049`)

The brief asked whether a **third direction** exists. It does.

```csharp
class A
{
}
#if Y
;
#endif
```

A type or namespace may be followed by an optional `;`, and the builder did not model it at all. So a declaration already closed, measured and vouched for reaches **forward** to claim a terminator written after it. Ways 1–8 are a header written *before* a group surviving it; ways 9–11 are a tail reaching *back* through one. This is the first that moves forward, which is why nothing in place could see it.

Reproduced independently: `A` is 4..10 without `Y` and 4..6 with it, all four symbol configurations parse cleanly so no oracle declines, and the product vouched 4..6.

**This is the only defect in the series that this PR introduces.** Opus proved it: at merge base `0334a2de5` the leading group poisoned the file, so the row was declined for an unrelated reason. Underneath it sits a **pre-existing plain wrong span** — `class A {\n}\n;` reports 1..3 where Roslyn says 1..4 — which the same fix corrects while keeping the row vouched.

The rule mirrors the reaching-back one, and reuses its shape: on a `;` following a closed type/namespace row with nothing pending, extend only if the `;` is in the same section with known depth; otherwise refuse the row rather than extend it. Verified across all six kinds (class/struct/interface/record/enum/namespace), nested and file-scoped, plus the symmetric both-branches form.

**Generator cases 19–21 are the point.** No composition of the existing pools could place a bare `;` after a `}` that closed a type, so the shape was unreachable at any case count. With them in place, neutralizing the fix fails **13** tests — 3 named gates, 5 diff seeds, 5 product seeds — so the generator catches it unaided.

### Opus finding 2(c) — it corrected a rationale of mine

My round-8 justification for the `product`-mode gate said it covers rows that exist in only one build. **That cannot happen.** Vouching requires `pending.All(t => t.DepthKnown)`, so a vouched row lies outside every group and its text is identical in both builds; `cmp1 == cmp2` by construction. The gate's real value is different and still worth having: `diff` mode never reads the product's own numbers, so a product that is consistently wrong in both builds passes it. Corrected in the doc.

### Opus M5 — marked, not fixed

Opus flagged the third conjunct of the `oneBranch` condition as possibly redundant, and explicitly **declined to report it as a vacuous gate**, on the grounds that the mutation which would prove it makes the code stricter and so no refusal test could fail. That is the right standard — *a surviving mutant only means something if the mutation could have made the product wrong.* The conjunct is now annotated as conservative and **UNVERIFIED**, with the reachability argument and an explicit "do not cite it as gated".

### The standing methodological result, now three for three

| Round | What the attack on the generator found |
| --- | --- |
| 7 | every group emitted at file scope → 140,000 clean cases had compared **zero** member rows |
| 8 | `Group` spelled both branches from one body → asymmetric branches unreachable at any count; enums only as one-liners |
| 9 | `DifferFromProduct` silently compared 4 of 6 fields; constructors emitted only in never-vouched forms; no bare `;` after a type's `}` |

Three rounds, three structural blind spots, **never a case-count problem**. The practical check is now written into the fuzzer's header: read the `cmp1/<Kind>` and `cmp2/<Kind>` buckets before citing any zero.

### Cost

| | unknown | % |
| --- | --- | --- |
| Round 7 | 8,473 | 1.46 |
| Round 8 | 8,473 | 1.46 |
| **Round 9** | **8,473** | **1.46** |

Serilog `Logger.cs` **91 of 100**, unchanged since round 6. The round-8 and round-9 fixes have each cost exactly zero declarations.

### Validation at `353e9a049`

`dotnet-inspect.slnx` Release clean. Slicer **431/0**, Services **360/0**, Metadata **1397/0**, MetadataRendering **187/0**, CLI **2843/1** (the recorded `Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake, identical on clean `main`). Deep sweeps of three fresh seeds × 20,000 cases in **both** `diff` and `product` modes: 30,452 fair, 0 flagged. Comparison buckets cover 14 kinds.

The fixes again landed after the reviews, so this head is unreviewed.

richlander:
## Round 10 reconciliation — `353e9a049` → `c2c0e5225`

Reviewers: **Gemini 3.1 Pro** (`/home/rich/rr10-a`) and **Claude Opus 5** (`/home/rich/rr10-b`), both at `353e9a049`. Both **BLOCK**. Three wrong vouches found, all fixed; every finding actioned, none dismissed.

*(GPT-5.6 Sol was dispatched first and returned an empty response twice, including after an explicit step-by-step re-prompt. It was dropped rather than re-launched a third time, and Opus 5 took the second slot. Recording it because "a reviewer returned nothing" should not read as "a reviewer found nothing".)*

The brief made the fuzzer's structural **reach** Priority 1 for the fourth round running. For the fourth round running, that is where the defects were.

### Gemini 3.1 Pro — the thirteenth way, a shield

The trailing-`;` rule added last round requires an empty pending run, because a declaration under construction owns the terminator it reaches. But the scan lexes every branch, so a declaration written in a branch this build discards is **still pending**, and hides the `;` behind itself:

```csharp
class C {
    class Sy { }
#if X
    int Field = 1
#endif
    ;
}
```

With `X` the `;` terminates `Field` and `Sy` ends on line 2; without `X` there is no `Field`, the `;` is `Sy`'s own optional trailer, and `Sy` ends on line 6. Reproduced independently — both builds parse with **zero errors**, and the product vouched `2..2`. A file-scope `delegate` form does the same thing through a different classification path.

**This is not a fourth direction.** It is the same *masking* shape as the tenth and eleventh ways, now applied to the forward reach — which is the argument that **masking, not direction, is the axis worth attacking next.**

The refusal needs two terms and neither implies the other, so each has its own named recall test:

| term | what it protects | mutation |
| --- | --- | --- |
| every pending token is inside a group | a member whose *header* is unconditional keeps its terminator in every build | drop it → `AConditionalDeclaratorOnAnUnconditionalMember` fails |
| the `;` is in none of their sections | a declaration and its terminator in **one** branch vanish together | drop it → `AConditionalMemberCarryingItsOwnTerminator` fails |

The second is the one that matters for recall: `#if X int F = 1; #endif` after a class is the overwhelmingly common shape, and refusing it would have cost far more than the defect is worth. Neutralizing the whole rule fails **13** tests.

### Claude Opus 5 — two holes in the twelfth way's own refusal set

Opus was reviewing someone else's implementation of *its own* round-9 finding, and attacked the fix rather than defending the finding. Both holes are in the refusal set, which I had borrowed from the backward case — the brief's attack point 4, and the answer was **too narrow**.

**(a) A brace-less scope opener re-parents the row the `;` follows.**

```csharp
class A { }
#if Y
namespace NS;
class B { }
#endif
;
```

Without `Y` this is `class A { }\n;` — a legal program in which `A` ends at the `;`. But `lastClosed` is `B`, whose parent is the *conditional* namespace, so a walk over one parent never visits `A` at the outer scope. The refusal now walks **outward through branch-dependent parents, stopping at a vouched one** — a vouched parent exists in every build, so the scope it opens exists in every build and no terminator can escape it. A block-namespace control confirms the defect is specific to the brace-less opener.

**(b) The same hole one step further out — and this is the better story.** A brace-less declaration ends *without closing a block*, so it leaves `lastClosed` at `-1` and the rule does not run at all. Worse, when the namespace is inside a type it is **not an allowed row**, so it leaves no trace among the rows either. What it does leave is a terminator, so the test became whether *that terminator* was written in this branch.

Opus explicitly predicted this arm and reported that it could not build a parsing case for it. **The widened generator built one within 5,000 cases** — which is the case for widening generators rather than hand-writing fixtures.

### Which gate can see what — and a correction to a correction

Both of Opus's cases parse in **only one configuration**. `NoVouchedRowMovesBetweenBuilds` needs *two* configs each supplying a unique match, so it is blind to them however wrong the product is. Only the product-mode gate failed.

That matters because this is the third pass over that test's rationale. Round 8 justified it as covering "rows present in only one build". Round 9 showed that set is empty by construction — a vouched row lies outside every group, so `cmp1 == cmp2` always. Round 10 shows the **instinct was right one level up**: not a row present in one *build*, but a file with one parsing *configuration*. The doc now records all three justifications and says which one round 8 stated.

I also removed a completeness claim Opus falsified — the comment asserting that "every sibling the `;` could have attached to is a candidate", which is exactly what a brace-less opener breaks. That is the nineteenth false doc comment found in this program, and the fourth that was a quantified claim.

### Opus's second finding (Low) and negative results

`0eca4554a`'s `curType` fix reached `MemberPool` but not `EmitSplit` case 9, so the only *conditional* constructor shape still goes uncompared. Opus verified by hand that the product is correct there, and that repointing it produces `ambig/Constructor` instead — the shape needs two same-named constructors, and the `(Kind, Name)` matcher cannot express overloads. Recorded as a **known structural limit of the matcher**, not fixed, since fixing it as written just moves the silent skip.

Opus also added six probe cases (non-`class` kinds, block namespaces, block-namespace interiors, split delegates, split operators, `lastClosed` reset) and ran 4 seeds × 8,000 in both modes: **clean**. I adopted the three that reach product code with no generator path. Two structural observations worth carrying forward: `Differ` never compares `ParentIndex` or `Depth`, so a group that changes a vouched row's **parent** while leaving every line identical is invisible to the fuzzer by construction; and `Generate` picks `inType` once per case, so a group never appears inside a *nested* type body.

### Cost — the first non-zero one in four rounds

| | unknown | % |
| --- | --- | --- |
| rounds 7–9 | 8,473 | 1.46 |
| **round 10** | **8,539** | **1.47** |

Sixty-six declarations is the price of not reporting a wrong span. Serilog `Logger.cs` unchanged at **91 of 100**.

### Validation at `c2c0e5225`

`dotnet-inspect.slnx` Release clean. Slicer **441/0** (ten new gates), Services **360/0**, Metadata **1397/0**, MetadataRendering **187/0**, CLI **2843/1** (recorded `Layout_Count` flake, identical on clean `main`). Three fresh seeds × 20,000 cases in both `diff` and `product` mode after the fixes: 29,241 fair, **0 flagged**. Each of the three fixes independently mutation-proved (13, 7 and 7 failing tests).

The fixes again landed after the reviews, so this head is unreviewed.

richlander:
## Round 11 reconciliation

Reviewed at `c2c0e5225` (MERGEABLE/CLEAN, CI green) in isolated worktrees.
Fixes are in `ff0d7c367`.

**Reviewers: Gemini 3.1 Pro (BLOCK), GPT-5.6 Terra (APPROVE).**

GPT-5.6 **Sol** was deliberately not used: it returned two empty responses in
round 10. That is a harness failure and is not evidence that it found nothing.

### Gemini 3.1 Pro — three findings, all reproduced, all actioned

| # | Finding | Sev | Outcome |
| --- | --- | --- | --- |
| 1 | Trailing `;` masked after a brace-less opener bypasses **both** round-10 arms | Critical | **Fixed** in `ff0d7c367` |
| 2 | A vouched row's parent differs between builds with identical lines | Critical | **Split to #3725** — see scope note |
| 3 | `PendingCanVanishBefore` doc claims `DepthKnown` is false "exactly" inside a group | Medium | **Fixed** in `ff0d7c367` |

**Finding 1 — the fourteenth way.** Round 10 added two rules for a trailing `;`
and this masks one with the other's shape. The brace-less arm exists precisely
because `namespace Nr;` leaves `lastClosed` at -1, but it demanded an *empty*
pending run, so a second group carrying a pending declaration bypasses it — and
the thirteenth-way rule cannot cover, because it needs `lastClosed >= 0`. With
`{!X,!Y}` the file is `class C { class Sr { } ; }` and `Sr` ends at line 9; with
`{!X,Y}` the `;` is the field's and `Sr` ends at line 2. Both parse; `Sr` was
vouched at 2..2. The arm now tests the pending run the same way the
thirteenth-way rule does. Verified refused, control still refused, enclosing
type still vouched. **Corpus cost zero** — the shape needs a brace-less scope
opener inside a group.

**Finding 3 — false doc comment #20**, and the fifth quantified one. A group
that never balances makes the depth unknown for the rest of the file, so a token
outside every group can be unknown too; witness gated by
`AfterAnUnbalancedGroup_EvenARowOutsideEveryGroup_IsNotVouchedFor`. The rule
needs only the converse, which holds, and the extra unknown tokens can only add
refusals.

**Finding 2 — split, not dismissed.** `SpanKnown` is a claim about lines and the
contract says `Depth`/`ParentIndex` are not covered by it. But the same
paragraph asserts such groups are "refused outright", and that is false:
`#if A class X #else class Y #endif { int dummy; class Nested { } }` vouches
`Nested` with parent `X` when half the builds say `Y`, identical lines. This is
way 7's assign-vs-intersect at a new site, needs its own product fix, and is
tracked in **#3725** with full evidence.

### GPT-5.6 Terra — APPROVE, with a caveat I want on the record

Its first pass ran only the existing gates and did not touch the generator,
which the brief made Priority 1. I pushed back; its second pass built a
*separate* probe rather than closing the gap in the real harness, so the blind
spot stayed open in the committed fuzzer. Its durable contribution is a
soundness argument I verified independently: re-parenting requires either
leaving a brace open (fails the depth return) or closing one (breaches the depth
floor), so only a brace-less opener escapes both. My probes A–M confirm every
valid conditional re-parenting shape is otherwise refused, including
close-and-reopen at net-zero depth.

### The generator-reach result now reads five for five

Every round that attacked the fuzzer's structural reach found a defect; no round
ever found one by raising the case count. Round 11 continues it from the other
side: closing the `ParentIndex`/`Depth` gap produced 23 flags immediately, of
which 7 survive in compilable sources. **16 were a harness artifact** — the
generator can emit `public namespace N;` and `[Obsolete] namespace N;`, which
Roslyn *parses* but the language rejects (CS1671), so nesting has no answer to
be right about there. Recorded in #3725 so the follow-up does not rediscover it.

### Validation at `ff0d7c367`

slnx Release clean; slicer **445/0**, Services **360/0**, Metadata **1397/0**,
MetadataRendering **187/0**. Three seeds x 20,000 in both `diff` and `product`
modes: **29,311 fair, 0 flagged**. Corpus **8,539 / 1.47%**, unchanged from
round 10. Serilog canary **91/100**, unchanged.

Mutation proofs: removing `PendingCanVanishBefore` from the widened arm fails
the new gate; removing the empty-run term fails eight.

richlander:
## Round 12 reconciliation

Reviewed at `ff0d7c367` (MERGEABLE/CLEAN, CI green) in isolated worktrees.
Fixes are in `f62450b3f`.

**Reviewers: Claude Opus 5 (BLOCK), Gemini 3.1 Pro (APPROVE).**

### Claude Opus 5 — two new over-vouches, both reproduced and fixed

| # | Finding | Sev | Outcome |
| --- | --- | --- | --- |
| 1 | Brace-bodied non-type member masks **all three** trailing-`;` arms | Critical | **Fixed** — the fifteenth way |
| 2 | Extension block close makes the refusal start one scope too high | High | **Fixed**, with +1 corpus recall |
| 3 | Duplicated/truncated round-10 comment sentence | Info | **Fixed** as part of the rewrite |

**Finding 1 — the fifteenth way, and the requested three-way mask.**

```csharp
class C
{
    class Sh { }
#if X
    void Mh() { }
    int Fh = 1
#endif
    ;
}
```

All four X/Y configurations compile. With X the `;` terminates `Fh` and `Sh`
ends at line 3; without X both members vanish and the `;` is `Sh`'s optional
trailer on line 8. Product: `SpanKnown=true`, line 3.

All three arms skipped simultaneously: `lastClosed` names `Mh` (Method), so
`trailerTarget == false` defeats the twelfth/thirteenth arms; `lastClosed >= 0`
defeats round 11's brace-less arm. The fix replaces the two special
masked-trailer arms with one `predecessorCanVanish` rule: the relevant fact is
not the remembered row's kind but whether that row (or its remembered
terminator) and the pending run can both vanish before this `;`.

**Finding 2 — the correct refusal target was below `lastClosed`.** C# 14
extension blocks carry their enclosing type's row index so members remain
parented to that type, but the closing-brace path treated the carried index as
the row the brace closed. It rewrote and declined `C`, then began the refusal at
C's siblings and never visited its child `Sy`, which is the row whose end moves.
Scope entries now carry `(RowIndex, OwnsRow)`: extension blocks remain
transparent for parenting, but closing one no longer masquerades as closing its
parent.

This is a recall fix too: corpus unknown **8,539 -> 8,538** of 581,298 (still
1.47%). Serilog remains 91/100.

### Gemini 3.1 Pro — APPROVE, explicitly superseded

Gemini stated that the three-way mask could not be constructed. Opus supplied
one whose four configurations all pass both parse and compilation diagnostics,
and I reproduced it independently. Gemini also reasoned that a refusal target
could not live below `lastClosed`; Opus's extension-block fixture is exactly
that case. The APPROVE is therefore superseded by two concrete counterexamples,
not merely outweighed by a stricter reviewer.

### Generator reach — six for six

Opus found both defects by widening `EmitSplit`. Before round 12, no case put a
brace-bodied **non-type** member before a trailing `;`, and no case put a C# 14
extension block there. Both are now permanent generator cases. This is the
sixth consecutive round to find a product defect in structural reach; no round
has found one by raising case count.

The contract and fuzzer header also now say what the harness actually proves:
the fair-case gate checks **parse** diagnostics, not compilation diagnostics.
Round 11's `public namespace N;` / CS1671 artifact is why that distinction must
be explicit when the compared fact is structural rather than a line.

### Validation at `f62450b3f`

slnx Release clean; slicer **448/0**, Services **360/0**, Metadata **1397/0**,
MetadataRendering **187/0**, CLI **2843/1** (known
`Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake). Three seeds x
20,000 in both `diff` and `product`: **29,167 fair, 0 flagged**. Corpus
**8,538 / 1.47%**; Serilog **91/100**.

Mutation proofs on final code: restoring the non-type kind dependency fails
**12** tests; making an extension scope own its carried parent row fails **11**.

richlander:
## Round 13 reconciliation

Reviewed at `f62450b3f` (exact pushed head, `ci-required` pass,
MERGEABLE/CLEAN) in isolated worktrees. Fixes are in `a184c947a`.

**Reviewers: Claude Opus 5 (BLOCK), MAI-Code (APPROVE).**

### Claude Opus 5 — Critical, reproduced and fixed

**The seventeenth wrong-span mechanism: a brace-less scope steals an enclosing
namespace's physical close.**

```csharp
namespace Outer
{
    namespace Mid
    {
        class Sy { }
#if X
        namespace Inner;
#endif
    }
}
```

The `{}` and `{Y}` configurations have zero parse and compilation diagnostics;
Roslyn puts `Mid` at lines 3..9. Product: `SpanKnown=true`, lines 3..10, where
line 10 is `Outer`'s brace.

A file-scoped namespace owns a row and opens a scope, but no brace closes it.
The typed stack introduced in round 12 captured only `(RowIndex, OwnsRow)`, so
`Inner`'s stranded entry consumed `Mid`'s `}` and shifted every outer close by
one. The existing `namespaceScopeLostFrom` refusal starts at `Inner`'s row and
cannot repair enclosing rows emitted earlier. `origin/main` computes the same
wrong number but declines `Mid`; balanced recovery promoted a safe decline into
an over-vouch.

Scope entries now also carry `ClosesWithBrace`. Before a `}`, brace-less entries
are discarded so the physical brace closes its actual owner. The fix restores
exact Roslyn lines and vouches the enclosing namespaces rather than merely
broadening refusal.

**False comment #22 (if round 12's parse/compile wording is counted as #21).**
The unified-rule comment said ordinary unconditional text shares the
terminator's section and is unaffected. An empty group changes the section and
conservatively declines the row despite all builds agreeing. Corrected without
locking that recall loss into a gate.

### Generator reach — seven for seven

The shipped generator never wrapped top-level generation in a block namespace,
so `TopPool`'s file-scoped namespace always landed at file scope. There was no
physical namespace `}` for its brace-less entry to steal. Opus added a 0-2
namespace wrapper and immediately got **216 / 230 flags per 20,000**, all in
product mode; diff mode is blind because the valid builds agree and only the
product's number is wrong. The wrapper is now permanent.

Namespace comparisons rose roughly 16x (about 211 to 3,377 in Opus's seed),
which is the evidence that the widening changed reach rather than case count.
This is the seventh consecutive review round to find a product defect in
structural reach. No round has found one by raising case count.

### MAI-Code — APPROVE, superseded by the missing wrapper

MAI ran the shipped suite, four shipped-generator sweeps, and ad hoc
file-scoped-namespace probes, all clean. Its namespace buckets were only
**33-52** per 5,000 cases; it did not place a file-scoped namespace inside a
block namespace. Opus's three-line wrapper exposes 216-230 failures per 20,000.
The approval therefore describes the generator's old reach, not the failing
family, and is superseded by the reproduced fixture.

### Validation at `a184c947a`

slnx Release clean; slicer **450/0**, Services **360/0**, Metadata **1397/0**,
MetadataRendering **187/0**, CLI **2843/1** (known
`Layout_Count_CountsFilesRatherThanRenderedTreeLines` flake). Three seeds x
20,000 in both `diff` and `product`: **28,973 fair, 0 flagged**. Corpus
**8,538 / 1.47%** and Serilog **91/100**, both unchanged.

Mutation proof: making a file-scoped namespace claim `ClosesWithBrace` again
fails **6** tests.

richlander:
## Round 14 fixed-head convergence review

Reviewed exact settled head `a184c947a` in isolated worktrees.

**Claude Opus 5: APPROVE. GPT-5.6 Sol: APPROVE.**

### GPT-5.6 Sol

Attacked 17 scope-transition variants: nested and sibling conditional
namespaces, types, extension blocks, initializer and anonymous scopes, trailing
semicolons, malformed closes, and EOF. It checked parse and compilation
diagnostics and found no vouched-line mismatch.

Evidence:

- build clean; slicer 450/0
- seed 3680 x 20,000: 9,566 fair, 0 flagged
- Namespace `cmp1/cmp2 = 4,031/4,031`
- changing the `while` to remove only one brace-less entry passed both direct
  gates but the widened fuzzer caught wrong namespace ends, proving both the
  loop and the permanent generator reach are load-bearing

### Claude Opus 5

Established the fix's core invariant: no compilable C# configuration lets a
physical `}` close a file-scoped namespace. A brace-less row's end is resolved
from rows at EOF, not from the scope stack, so discarding its stack entry cannot
lose its end line; the only exposed scope that can compile is the physical
brace owner below it.

Attacked 42 hand-built variants across the complete scope-transition matrix,
plus a purpose-built scope-model fuzzer and repository fuzz:

- 40,000 product cases: 19,232 fair, Namespace 7,790/7,790, 0 flagged
- 30,000 product cases: 14,427 fair, Namespace 6,167/6,167, 0 flagged
- 30,000 diff cases: 14,496 fair, Namespace 5,942/5,942, 0 flagged
- scope-model fuzzer: 11,237 fair, 5,034 file-scoped namespace cases, 2,942
  stranded-entry cases, 0 flagged

Mutations killed:

- file-scoped namespace claims `ClosesWithBrace=true`: 6 tests
- `while` becomes `if`: all five fuzz seeds
- discard only at EOF: equivalent to full revert, killed
- extension block marked brace-less: 10+ tests

One redundant `lastClosed` mutation survived but could not produce wrong
output; Opus explained why every meaningful path overwrites it.

Opus found one real but delta-independent trailing-semicolon defect. It
reproduces byte-identically with the round-13 fix reverted, so it is filed as
follow-up **#3741** rather than moving this reviewed head.

The two independent reviews make the narrow `ClosesWithBrace` delta
persuasively clean.

Ready to merge

Issue #3651, opened 2026-08-01
Type listing view lacks the projection pre-check the single-type view has


richlander:
## Behavior

The single-type `type` view validates a projection against the section schema *before* rendering, so an unmatched field or column gets a named, actionable diagnostic:

```console
$ dotnet-inspect type System.String -S "Type Info" --tsv --columns FakeColumn
Warning: column 'FakeColumn' not found in section 'Type Info' Run -D "Type Info" to list available columns.
Error: No columns matched projection: FakeColumn
```

The **listing** view (`type --platform X`, `type <prefix>`) has no such pre-check. It only reports what the post-render diagnostic can infer:

```console
$ dotnet-inspect type --platform System.Text.Json -S Classes --tsv --columns FakeColumn
Error: No columns matched projection: FakeColumn

$ dotnet-inspect type --platform System.Text.Json -S Classes --tsv --fields Bogus
Note: 1 field has no data: Bogus
```

Two gaps follow:

1. No `Warning: column 'X' not found in section 'Y'`, and so no pointer at `-D` to discover the real names.
2. `--fields` on a nonexistent field is a `Note` with **exit 0**, where the single-type path treats it as an error with exit 1. A machine consumer piping `--tsv` cannot distinguish "this field is genuinely empty" from "you misspelled it".

## Cause

`TypeCommand` runs `ProjectionDiagnostics.ValidateProjection` against the rendered schema before rendering (`src/dotnet-inspect/Commands/TypeCommand.cs:258-261`). The listing path in `ApiCommand.WriteFullApiOutput` only calls `ProjectionDiagnostics.DiagnoseRendered` afterwards.

## Scope

Pre-existing and listing-wide — it applies to every section the listing view publishes, not to any one section. Confirmed identical on `main` (`ec7b254ef`) for `-S Classes`.

Found while reviewing #3648, where the new `API Info` fact table was held to parity with the per-kind sections beside it rather than to the stronger single-type behavior. Raising the listing view to the single-type standard is a separate change with its own blast radius across every listing section and both machine modes.

richlander:
## Second symptom: a fact table on the listing path can exit 0 with empty output

Found while re-reviewing #3648. Supplying `--fields` and `--columns` **together**, both unmatched:

```console
$ dotnet-inspect type --platform System.Text.Json -S Classes --tsv --fields BogusA --columns BogusB
Error: No columns matched projection: BogusB          # stderr
(no stdout, exit 1)

[...]

PR #3683, opened 2026-08-02
Wire the type listing's bare -S to its fixed overview

richlander:
Bare `-S` on a `type` listing now renders the bounded `API Info` section instead of the verbosity ladder's five growing per-kind tables, and the inline compact fields list becomes the `-v:q` view.

Slice 4d of #3547. `type` only; one command per PR.

## What changed and why

Bare `-S` on a listing was never a preset. The listing pipeline published no `Info` sections, so it resolved to an **empty include set**, and `IsRequested`'s `include is { Count: > 0 }` read empty as *"no filter"* — falling through to `Classes`, `Structs`, `Interfaces`, `Enums`, `Delegates`, every one of which grows with the assembly. `type --platform System.Private.CoreLib -S` printed 1353 type rows.

#3648 gave the listing a bounded `Fixed` section. This slice points bare `-S` at it, so the listing answers the same directive every other migrated command does: **print the fixed, bounded, network-free sections**.

That made the identity facts appear twice on any view showing both — as the inline line and as the section restating it. So the compact line becomes the `-v:q` view: `-v:q` renders no sections, so the line is the only identity that view has; everywhere else `API Info` carries it.

| Invocation | Before | After |
| --- | --- | --- |
| `-v:q` | compact line | **unchanged, byte-identical** |
| default (`-v:m`) | compact line + `Classes` | `Classes` |
| `-S` | 5 growing tables | `## API Info` |
| `-S -D` | advertises all 5 kinds | advertises `API Info` |
| `--fields Types` | `Types: 89` | **unchanged, byte-identical** |
| `--columns Type` | title + line + table | **unchanged, byte-identical** |
| `-S --fields <unmatched>` | **nothing at all**, exit 0 | named error, **exit 1** |
| `-S Classes --fields <unmatched>` | renders Classes, exit 0 | **unchanged** |

## Two defects the first commit introduced, both found by differential

Neither was caught by tests; both were caught by comparing 213 invocations against the base.

1. **The prefix-browse path rendered an empty document.** A dotted prefix that doesn't resolve enters `RunPreamble` looking like a single type — so sections resolve against the *single-type* pipeline — but `TypeCommand` renders a **listing**. Bare `-S` resolved to `Type Info`, which that view cannot render. The compact line had been masking it. `TryWritePrefixBrowse` now re-resolves `IncludeSections` against the pipeline that actually renders them.

2. **`--fields Types` lost its target.** Those seven scalars *are* the document fields `--fields` names, so emptying them turned an explicit projection into the `Classes` table with no title — a projection surface silently losing what it asked for. Suppression is a decision about the **default view**, so naming the fields opts out of it. `-S` still wins, because there the section already carries the facts.

## Non-action boundary

- `-S Classes` on the unresolved-prefix path is still rejected (the preamble rejects it before `TypeCommand` runs). **Pre-existing and byte-identical on the base** — verified — and left to its own slice under #3547.
- `member`, the deprecated `api` shim, `library`, `package`, and single-`type` are untouched.
- **`type <unresolved-prefix> -v:q` no longer prints the identity line** (found by MAI-Code). That path floors verbosity to Minimal — pre-existing, three sites in `TypeCommand`, present on base — so the line is suppressed there at every verbosity. The floor exists so the "best-effort prefix matches" the stderr note promises are actually shown; removing it would make quiet print identity and no candidates, contradicting the note. Identity stays reachable on that path via bare `-S`, which renders `API Info`. Disclosed rather than silently accepted.
- `--fields <unmatched>` **without** `-S` still renders the ladder and exits 0, because the ladder still renders. The gate fires on an empty render, not on an unmatched name, so it says nothing whenever output was produced.
- The gate is **not** name validation, deliberately. Two reviewers found a name-validating pre-check rejecting three valid projections (`-S "API Info" --columns Field`/`Value`, `-S Classes --fields Types`) whose names no single section schema lists. The same latent flaw exists on `main` in the single-type path — `type System.String -S "Type Info" --columns Field` is rejected there today — and is **not** fixed here; it is out of this slice and unchanged.
- `-S "API Info" --tsv --fields <unmatched>` printed nothing and exited 0 **on base too**. Fixed here as pre-existing collateral, because this PR's own parity test compares against that path. Same for `--plaintext`, which bypassed the gate entirely by serializing straight to the console.
- An empty render is **only** a failure when a projection is active AND no projected name matches, **anywhere in the document, as the kind being projected**. Matching is done by markout's own `ValidateProjection` over a kind-filtered throwaway schema, because projection names may be **wildcards** -- `--fields "Ver*"` selects `Version`, and set membership rejected it (found by GPT-5.6). Reimplementing the glob would be a second matcher that could drift from the one performing the projection. Internal `StableName`s are excluded: markout does not project by them, so `--fields Assembly` must fail (found by MAI-Code) -- `--fields` only by a `field`, `--columns` only by a `column`. Matching the kind is what makes "resolves somewhere" mean something: `Type` is a `Classes` **column** and a field nowhere, so without it `-S "API Info" --fields Type` printed nothing at exit 0 (found by GPT-5.6 in round 6). Resolving against only the *selected* sections was wrong: `Version` is document-level and advertised under `API Info`, so `-t "NoSuchType*" -S Classes --fields Version --tsv` failed a valid name once the zero-row filter collapsed the render (found by GPT-5.6 in round 5). A known field holding no value (`-S "API Info" --fields Version` against a local `.dll`) and an empty section with no projection (`-S Interfaces`) both stay exit 0, byte-identical to base.
- **`--fields <valid> --columns <bogus>` silently ignores the unmatched column and exits 0.** Identical on base, so not a regression. markout's `WriteFieldsTable` returns early once field projection empties the row set, before `ComputeColumnMap` resolves columns. Recorded on #3651 and deliberately left to markout rather than covered by a fifth product-side check.
- `-S "API Info" --fields Assembly` now exits 1. `Assembly` is the schema's internal stable name for `Library` and markout does not project by it. Of the whole schema only `Library`/`Assembly`, `TFM`/`Tfm` (case-only, inert) and `Target Library`/`TargetLibrary` (unreachable section) differ at all.
- `-S "API Info" --fields Type` now exits 1. `Type` is a `Classes` **column**; `-D` reports a `Kind` per item and the gate requires it to match the flag. Base printed nothing and exited 0 -- a gap this PR previously failed to close, not one it opened.
- The synthesized `Field`/`Value` columns are listed in **no** schema and would fail a kind match. They survive because emptiness is checked before names and they always render. Verified rather than assumed: `-t`, `-m`, and `--rows 0` crossed with `--columns Field`/`Value`, 3 targets x 5 formats, **60 combinations, all byte-identical to base**.
- `-S "API Info" --fields Field` now exits 1. `Field` is the fact table's synthesized **column** header, not a field: `-D "API Info"` advertises `Library`, `Types`, `Methods`, `Properties`, `Version`, `TFM`, `Source`. Base printed nothing and exited 0, which is the defect. `--columns Field`/`Value` stay byte-identical.
- `-S --columns Type` now exits 1 with markout's own `No columns matched projection: Type`, because bare `-S` selects `API Info`, which has no `Type` column. Direct consequence of the slice; fails loudly.

## Evidence

**Differential vs the current base `f42d8e586`, widened after review: 275 identical / 154 diffs / 7 vacuous**, across 6 libraries plus `--package` and local-`.dll` targets. Every diff is the `type` listing, its `--package`/local-dll equivalents, or its prefix path. **Single-`type` is 0 diffs.** On the listing itself these remain byte-identical to base: `-v:q`, `--json`, `--tsv`, `--jsonl`, `--count`, `-D`, `-D --schema`, `--columns`, `--fields`.

The one hit outside those surfaces is `package --columns Name`, which crashes identically on both binaries and differs only by worktree path inside the stack trace — pre-existing, filed as #3686.

The harness fails any row whose **base** side rejects a flag, so a fictional-flag row scores as VACUOUS rather than as coverage; 7 rows scored that way (`--all` and `--docs` on commands that do not take them).

**Seven mutations, each failing at least the gate that covers it.** (An earlier revision of this body said "only" that gate; GPT-5.6 correctly pointed out that the `showCompactFields` mutations also fail `Type_Listing_ApiInfo_RestatesTheInlineIdentityLineExactly`, which reads the same line. The claim is corrected here.)

| Mutation | Fails |
| --- | --- |
| `usesFixedOverview` back to the single-type form | `Type_Listing_BareSelect_RendersOnlyTheFixedOverview` |
| `showCompactFields` forced `true` | `Type_Listing_CompactFields_AppearAtQuietAndNowhereElse`, `..._RestatesTheInlineIdentityLineExactly` |
| `showCompactFields` forced `false` | same two, the first on its non-vacuity assertion |
| re-resolution neutered | `Type_PrefixBrowse_BareSelect_ResolvesAgainstTheListingPipeline` |
| `--fields` carve-out dropped | `Type_Listing_CompactFields_AppearAtQuietAndNowhereElse` |
| `--columns` dropped from the carve-out | `Type_Listing_CompactFields_AppearAtQuietAndNowhereElse` |
| empty-render gate weakened to a null test | `Type_Listing_UnmatchedProjection_FailsByNameRatherThanRenderingNothing`, `..._ReportsUnmatchedProjectionsLikeTheRestOfTheView` |
| no-projection guard removed | all 3 `Type_Listing_EmptySectionWithNoProjection_StaysSuccessful` |
| name-resolution narrowing removed | all 3 `Type_Listing_EmptyResultWithoutAnUnmatchedName_StaysSuccessful` |
| empty-render gate made to reject any active projection | all 4 `Type_Listing_LegitimateProjections_SurviveTheEmptyRenderGate` cases |
| name resolution narrowed back to the **selected** sections | exactly the 2 new `Type_Listing_EmptyResultWithoutAnUnmatchedName_StaysSuccessful` cases (`--tsv`, `--jsonl`) |
| kind filter relaxed to accept every kind | exactly the 4 `Type_Listing_ProjectionMatchingTheWrongKind_FailsLikeAnUnknownName` cases |
| markout matcher reverted to set membership | exactly the 3 positive `Type_Listing_WildcardProjection_MatchesTheWayMarkoutMatches` cases |
| `StableName` re-added to the candidate set | `Type_Listing_StableNameProjection_IsNotAValidProjectionName` |

**Suite:** 2843 total, 14 skipped, 1 failure — `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails identically on unmodified `main` (local NuGet cache lowercases `newtonsoft.json.nuspec`).

Docs updated: `README.md`, `docs/design/info-view.md`, `docs/design/progressive-disclosure.md`. `markdownlint` clean.


richlander:
## Review reconciliation — round 1 (GPT-5.6 and MAI-Code, both @ `444c34adf`)

Three findings, **all three real and all three acted on**. Fixed head is `357bcc4d1`.

### BLOCKING (GPT-5.6) — bare `-S` could return success-shaped empty output — **FIXED**

`type --platform System.Net.Http -S --fields Bogus` printed **nothing at all** and exited 0, where base printed the Classes tables. Reproduced exactly as reported.

Root cause was deeper than the projection: bare `-S` selects `API Info`, a two-column fact table, so an unmatched field empties it — and the Markdown fall-through in `WriteFullApiOutput` had **no projection diagnostics at all**, unlike both tabular branches. With the document scalars suppressed there was then no document field left, and markout drops the H1 once a projection leaves nothing renderable, so the output was not thin but empty.

Fixed by adding the pre-render `ProjectionDiagnostics.ValidateProjection` that the single-type path (`TypeCommand.cs:257`) and the package path already had. Unmatched names are now rejected **by name with exit 1**.

This also closes the `--fields`+`--columns` divergence recorded on #3651 for this view — the pre-check runs before markout sees either projection, so the fact table and the per-kind sections can no longer disagree. `Type_Listing_ApiInfo_ReportsUnmatchedProjectionsLikeTheRestOfTheView` is retargeted from the weaker parity it pinned (Note, exit 0) to the stronger one, and now pins the formerly-divergent combination.

**Disclosed consequence:** `-S <section> --fields <unmatched>` now exits 1 where it exited 0 and rendered the section unprojected.

### BLOCKING (GPT-5.6) — `--columns` lost the document title; my byte-identity claim was false — **FIXED, claim corrected**

`type --platform System.Net.Http --columns Type` lost both the H1 and the identity line. The finding is correct **and so is the charge against the claim**: the PR body listed `--columns` as byte-identical on the strength of a `--columns Field` row, which matched nothing on either side and therefore proved nothing. `--columns Type` is the row that would have caught it — the same too-narrow-matrix mistake that has now cost this stack four rounds.

The `--fields` carve-out is widened to cover `--columns`. `--columns Type`, `--columns Members`, and `--columns Type,Members` are byte-identical to base again, and the corrected claim is in the body.

### BLOCKING (MAI-Code) — `-v:q` on the unresolved-prefix path loses the identity line — **UPHELD as a real behavior change; disclosed, not reverted**

Reproduced. The finding is right that the output changes and right that my `-v:q` claim did not distinguish this path.

The mechanism is a **pre-existing** verbosity floor — `options.Verbosity < Verbosity.Minimal ? Verbosity.Minimal : options.Verbosity`, three sites in `TypeCommand`, present unchanged on base — so `-v:q` never actually reaches this view as Quiet. Base masked it because the line rendered at Minimal too.

Not reverted, and the reason is the stderr note: the floor exists so the "best-effort prefix matches" it promises are actually shown. Removing it makes `-v:q` print library identity and **zero candidates** while the note claims candidates are being shown. Identity remains reachable on that path via bare `-S`, which now renders `API Info` — verified. Recorded in the non-action boundary.

### NON-BLOCKING (GPT-5.6) — "each mutation fails only one gate" was false — **CORRECTED**

Correct, and verified by reading the assertions: both `showCompactFields` mutations also fail `Type_Listing_ApiInfo_RestatesTheInlineIdentityLineExactly`, which reads the same line. The body now says "at least" and names the second test. The mutation table also grew to seven.

### Also found while re-measuring

`package --columns Name` prints a correct `Error: No columns matched projection` and **then crashes** with the underlying `InvalidOperationException` from `MarkoutProjection.ComputeColumnMap`. Identical on both binaries, differing only by worktree path — pre-existing, not this PR. Filed as **#3686**.

### Re-validation at `357bcc4d1`

- **Suite:** 2822 total, 14 skipped, 1 failure — the known `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails on unmodified `main` here.
- **Two new gates mutation-verified:** neutering the pre-check fails all four unmatched-projection cases; dropping `--columns` from the carve-out fails the compact-fields gate.
- **Widened differential:** 269 identical / 160 diffs / 7 vacuous, across 6 libraries plus `--package` and local-`.dll` targets, and including every flag combination both reviewers used. Every diff is the `type` listing, its `--package`/local-dll equivalents, or its prefix path. **Single-`type` is 0 diffs.**

Both reviewers ran their commands from their own worktrees and both reported the correct head — thank you, that was the failure mode that cost this stack a discarded round.

Requesting a fresh round at `357bcc4d1`; not marking ready until it is clean.

richlander:
## Review reconciliation — round 2 (GPT-5.6 and MAI-Code, both @ `357bcc4d1`)

The confirmation round **failed**, and it failed on the fix rather than on the original findings. Both reviewers independently found the same class of regression. New head is `f63b21d8e`.

### The fix from round 1 was the wrong mechanism — **replaced**

Between them the two reviewers found **three valid projections that my pre-check rejected**, each of which worked before this PR:

| Invocation | Base | `357bcc4d1` | `f63b21d8e` |
| --- | --- | --- | --- |
| `-S "API Info" --columns Field` | renders | **exit 1** | renders, byte-identical |
| `-S "API Info" --columns Value` | renders | **exit 1** | renders, byte-identical |
| `-S Classes --fields Types` | renders | **exit 1** | renders, byte-identical |

Found by MAI-Code (the first two) and GPT-5.6 (all three). Both reproduced exactly as reported.

The cause is that **the set of legitimately projectable names is wider than any one section's schema**. `Field` and `Value` are synthesized by the fact-table renderer and appear in no schema; `Types` is a document-level field that survives whichever section is selected. A name-validating pre-check cannot see either, so it reports a false negative — and it did, on three invocations, in a commit whose entire purpose was to make failures visible.

**Replaced with a check of the rendered result**: if a projection leaves nothing to print, fail by name; otherwise say nothing. That has no blind spot, because it asks the question the ban on success-shaped empty output actually asks — *did anything come out?* — rather than trying to predict what could have. It also needs no per-section knowledge, so it holds for sections added later.

I want to name the mistake plainly: I reached for name validation because the single-type path does it, and I did not check whether that precedent was itself sound. It is not — `type System.String -S "Type Info" --columns Field` is rejected on `main` today for exactly the same reason. I propagated a latent bug into a path where it was reachable.

### `--count` false zero (both reviewers) — **FIXED**

`-S "API Info" --count --fields Bogus` printed `0` and exited 0 while the section has seven rows. GPT-5.6 additionally showed `-S --count --fields Bogus` regressing from base's exit 1 to exit 0. Both now exit 1 with a named error; the count path runs the same gate.

### One case fixed beyond what was reported

`-S "API Info" --tsv --fields Nonexistent` printed nothing and exited 0 — **on base as well**, so pre-existing rather than a regression. It is the same banned shape in the very path this PR's parity test compares against, so leaving it would have made that test assert parity between a fixed path and a broken one. The gate now covers all three paths that can print a projected result.

### Two round-1 behavior changes revert to base, and both reverts are correct

`-S Classes --fields Bogus` renders Classes and exits 0 again — the Classes table is not field-projected, so the request genuinely does produce output. Same for its `--tsv` form. **The disclosed exit-code change now applies only where the render is actually empty**, which is a much narrower and much better-justified boundary than the one I disclosed in round 1.

### Test changes

`Type_Listing_ApiInfo_ReportsUnmatchedProjectionsLikeTheRestOfTheView` no longer asserts equal exit codes. The two sections legitimately differ — an unmatched `--fields` empties the fact table but not the Classes table — so demanding identical codes would force one of them to lie. It now asserts the invariant that binds both: the projection is named on stderr, and empty output never accompanies exit 0. New `Type_Listing_LegitimateProjections_SurviveTheEmptyRenderGate` pins all four false positives as a `[Theory]`.

### Re-validation at `f63b21d8e`

- **Both gates mutation-verified.** Weakening the emptiness test to a null test fails 2 tests; making it reject any active projection fails all 4 false-positive cases.
- **Suite:** 2825 total, 14 skipped, 1 failure — the known `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, which fails on unmodified `main`.
- **Differential:** 270 identical / 159 diffs / 7 vacuous. Single-`type` **0 diffs**. The only hit outside the listing and its prefix path is the pre-existing `package --columns Name` crash (#3686), differing only by worktree path.

One remaining intended diff worth naming: `-S --columns Type` now exits 1 with markout's own `No columns matched projection: Type`, because bare `-S` selects `API Info` and that section has no `Type` column. That follows directly from the slice's core change and fails loudly rather than silently.

### Also verified from round 2

GPT-5.6 independently confirmed the `package --columns Name` crash is present on both binaries, and reported it did not reconstruct the differential totals — noted, so that number rests on my run alone.

Requesting a third round at `f63b21d8e`. Not marking ready until it is clean.

richlander:
## Review reconciliation — round 3 (GPT-5.6 and MAI-Code, both @ `f63b21d8e`)

Round 3 **failed too**. Four findings, all real, all fixed. New head is `43ad3103c`.

I want to be direct about the pattern: this is the third consecutive round where my fix was wrong, and each time it was wrong in the same direction — a check that fires more widely than the defect it was written for. Round 1 validated names and rejected valid ones. Round 2 gated emptiness and rejected honest empty answers. Both times the tests I wrote passed, and both times two independent reviewers found it in minutes.

### BLOCKING (both reviewers) — `--plaintext` bypassed the gate entirely — **FIXED**

`type --platform X -S --plaintext --fields Bogus`: base rendered the full plaintext tables, head printed **nothing on either stream and exited 0**. That branch serialized straight to `Console.Out`, so the gate never saw it. It now renders to a buffer and goes through the same check.

This is the **third instance** of the same bypass shape — a path that skips a shared check because it renders differently, not because it should behave differently. The fact-table routing in #3648 was the first. Worth stating as a rule: **a new output format is a new place for every existing check to be forgotten.**

### BLOCKING (GPT-5.6) — a valid field holding no value was reported as unmatched — **FIXED**

```text
type --library <local>.dll -S "API Info" --fields Version --tsv
  base: Note: 1 field has no data: Version           exit 0
  head: Note: ... + Error: No fields matched         exit 1
```

`Version` is a field `-D "API Info"` advertises; the local assembly simply has none. Emptiness alone cannot tell an unknown name from a known field with no data — a distinction my round-2 rationale explicitly claimed was unnecessary. It was not.

### BLOCKING (GPT-5.6) — an empty section with no projection at all became an error — **FIXED**

```text
type --platform System.Net.Http -S Interfaces --tsv
  base: empty, exit 0
  head: Error: Nothing to render for the requested selection.   exit 1
```

No `--fields`, no `--columns` — System.Net.Http just has no interfaces. Worse, it failed **only in the tabular formats**, so the same query's success depended on the output format. That message is now unreachable and has been deleted.

### The corrected gate

An empty render has three causes, and only one is a failure:

| Cause | Verdict |
| --- | --- |
| Nothing requested that exists | **fail** |
| A known field holding no value | honest empty answer |
| No projection requested at all | honest empty answer |

The gate now requires an active projection **and** that every projected name resolves in no selected section. The name check is a **narrowing condition on an already-empty render, never a pre-check** — the schema blind spots that sank rounds 1 and 2 (`--columns Field`, `--fields Types`) all *render*, so ordering emptiness first puts them permanently out of reach. Both conditions are documented in the method's remarks with the false positive that motivated each.

### One of my new tests was vacuous, and I caught it by mutation

`Type_Listing_EmptySectionWithNoProjection_StaysSuccessful` originally targeted the test assembly — which **has** interfaces, so the section rendered rows and the case proved nothing. It passed under the exact mutation it was written to catch. Retargeted to a library with genuinely zero interfaces, it now fails 3/3 when the guard is removed. Flagging it because "I added a test" was not evidence here, and the mutation run is the only reason I know that.

### Re-validation at `43ad3103c`

- **Both narrowing conditions mutation-verified**, each against tests corrected first: removing the no-projection guard fails 3; removing the name narrowing fails 3.
- **Suite:** 2831 total, 14 skipped, 1 known pre-existing failure.
- **Differential:** 275 identical / 154 diffs / 7 vacuous — **five fewer diffs** than the previous head, which are exactly the five false positives reverting to base.
- All four round-2 false positives and both round-3 ones confirmed byte-identical to base.

### Noted, not acted on

Both reviewers said plainly they did not independently reconstruct the differential totals. That number rests on my run alone, and I am stating that rather than letting it read as reviewed.

Requesting a fourth round at `43ad3103c`.

richlander:
## Round 5 reconciliation — head `6f10c566e`

Two reviewers at `b933f27d6`. **One blocking finding upheld and fixed; one not upheld; one disclosed non-action.**

### Upheld — GPT-5.6: document-level fields escape a selected-section name check

```
type --platform System.Text.Json -t "NoSuchType*" -S Classes --fields Version --tsv
```

Base `f42d8e586` exits 0; `b933f27d6` exited 1 with `No fields matched projection: Version`.

`Version` is a document-level field, advertised under `API Info`, not under `Classes`. `TryReportEmptyProjection` resolved projected names only against `options.IncludeSections`, so it could not see it.

My round-3 rationale claimed this blind spot was unreachable: a document-level field keeps the render non-empty, and the gate checks emptiness *before* it checks names. **That reasoning was wrong in one case I had not tried** — `-t "NoSuchType*"` filters the selected table to zero rows, the render collapses anyway, and the name check runs after all.

Fixed in `6f10c566e` by resolving against `schema.SectionNames` — every section in the document. The cost is only the ability to fail a name that is valid elsewhere in the same document, which is the conservative direction for a gate that has now fired too widely in four consecutive rounds. Two `InlineData` cases pin it (`--tsv`, `--jsonl`); narrowing the loop back fails exactly those two and nothing else.

### Not upheld — MAI-Code: `-S "API Info" --fields Field`

```
type --platform System.Text.Json -S "API Info" --fields Field
```

Base: empty stdout, exit 0. Head: exit 1, `Error: No fields matched projection: Field`.

Reported as a false positive on a legitimate projection. It is not one — `Field` is not a field. `-D "API Info"` advertises exactly seven names, all of kind `field`:

```
Library | Types | Methods | Properties | Version | TFM | Source
```

`Field` and `Value` are the fact table's synthesized **column** headers, which is why they are legitimate for `--columns` and not for `--fields`. Both `--columns` forms — MAI's own round-3 findings — remain byte-identical to base at this head:

| Invocation | Base | Head |
| --- | --- | --- |
| `-S "API Info" --columns Field` | exit 0, 280 B | exit 0, 280 B, **identical** |
| `-S "API Info" --columns Value` | exit 0, 304 B | exit 0, 304 B, **identical** |
| `-S "API Info" --fields Version` | exit 0, 146 B | exit 0, 146 B, **identical** |
| `-S "API Info" --fields Field` | exit 0, **0 B** | exit 1, error by name |

The base behavior in the last row — a projection that matched nothing producing empty stdout at exit 0 — is the exact defect this PR exists to fix, and it is the same shape as round 1's first finding. Erroring by name is the intended change, not a regression.

### Non-action, disclosed — `--fields <valid> --columns <bogus>`

```
type --platform System.Text.Json --fields Version --columns Bogus
```

Both binaries exit 0 and the unmatched column is silently ignored. **Identical on base**, so not a regression. Root cause is upstream in markout: `MarkoutWriter.WriteFieldsTable` returns early once field projection yields zero fields, before `ComputeColumnMap` runs, so the column is never resolved. Recorded on #3651.

Deliberately **not** patched here. Four rounds of false positives came from adding checks that guessed at name validity; adding a fifth to cover an upstream early return is the same mistake again. It belongs in markout.

### Claims restated with the correct base

`main` moved mid-round, so the earlier `275 / 154 / 7` figure was measured against the retired base `72a6cfb45`. Re-measured at this head against the current base **`f42d8e586`**: **275 identical / 154 diffs / 7 vacuous**, single-`type` 0 diffs. Suite **2833**, 1 known pre-existing failure, 14 skipped.

### The gate, four revisions on

| Round | Mechanism | How it was wrong |
| --- | --- | --- |
| 1 | Name pre-check copied from the single-type path | Rejected `--columns Field`/`Value`, `--fields Types` |
| 2 | Emptiness of the render | Rejected valid-but-empty fields, and empty sections with no projection |
| 3 | Emptiness + active projection + names unresolved in **selected** sections | Document-level fields, once the selected table filters to zero rows |
| 4 | …resolved against **all** sections | current |

Every revision failed the same way — firing too widely — and each was found by a reviewer rather than by the suite. Round 6 should assume a fifth.

richlander:
## Round 6 reconciliation — head `fedb068a3`

**GPT-5.6's finding is upheld and fixed. MAI-Code has not reported yet;** this comment will be amended if its round adds anything.

### Upheld — resolving across all sections was too weak without matching the kind

```
type --platform System.Net.Http -S "API Info" --fields Type --tsv
```

Empty stdout, exit 0 — on **both** binaries. `Type` is not a field; it is a **column** of the `Classes` table, and round 5's widening let it validate a field projection against a section that has no such field.

GPT is right that this is the defect the gate exists to prevent, and right that my round-5 widening is what left the door open. Worth being precise about the status, because it changes nothing about whether to fix it but does change what the PR may claim: this case is **byte-identical on base**, so it is a gap this PR failed to close rather than one it opened. The PR previously implied the gate caught every unmatched projection. It did not.

**The two kinds are distinguishable, which I had assumed they were not.** `-D` reports a `Kind` per item:

```
-D "API Info"  ->  Library, Types, Methods, Properties, Version, TFM, Source   (all kind=field)
-D Classes     ->  Kind, Type, Members, Description                            (all kind=column)
```

So `fedb068a3` resolves projected names across every section **and** requires the item's kind to match the flag: `--fields` only by a `field`, `--columns` only by a `column`. That keeps round 5's fix (`Version` is a field in `API Info`, so `-t "NoSuchType*" -S Classes --fields Version` still passes) while closing round 6's hole.

| Invocation | Base | Head `fedb068a3` |
| --- | --- | --- |
| `-S "API Info" --fields Type` (`--tsv`/`--jsonl`/`--plaintext`/`--count`) | exit 0, empty | **exit 1, error by name** |
| `-S Classes --columns Type --tsv` | renders | identical |
| `-t "NoSuchType*" -S Classes --fields Version --tsv`/`--jsonl` | exit 0 | identical |
| `-S "API Info" --columns Field` / `--columns Value` | renders | identical |
| `-S Classes --fields Types` | renders | identical |
| `-S "API Info" --fields Version` | renders | identical |
| `-S Interfaces` on a library with none | exit 0 | identical |

### The risk this created, and why it is not reachable

Kind matching would reject the synthesized `Field`/`Value` columns, which **no schema lists**. They survive only because emptiness is checked before names and they always render. Rather than assert that, I tried to break it: `-t`, `-m`, and `--rows 0` crossed with `--columns Field`/`Value`, three targets, five formats — **60 combinations, all byte-identical to base**, no empty render reached. Documented in the method's `<remarks>` as the reason the ordering is load-bearing.

### Evidence

- Suite **2837**, 1 known pre-existing failure, 14 skipped.
- Differential against base `f42d8e586`: **275 identical / 154 diffs / 7 vacuous** — unchanged, so the narrowing introduced no new diffs.
- **Mutation:** relaxing the filter to accept every kind fails exactly the four new cases and nothing else.
- Four new `InlineData` cases, each also asserting `-S Classes --columns Type` still renders, so the rule reads as "wrong kind" rather than "banned name".

### GPT's other verifications

Both zero-row-filter cases confirmed non-vacuous and confirmed to reach the gate; ten focused gate tests passing; `-D "API Info"` confirmed to advertise exactly the seven claimed names. GPT stated plainly that it did **not** independently reconstruct the 275/154/7 differential totals.

### Running total

The gate has now been corrected in five consecutive rounds — four times for firing **too widely**, and this round, for the first time, for firing **too narrowly**:

| Round | Mechanism | How it was wrong |
| --- | --- | --- |
| 1 | Name pre-check copied from the single-type path | Rejected `--columns Field`/`Value`, `--fields Types` |
| 2 | Emptiness of the render | Rejected valid-but-empty fields, and empty sections with no projection |
| 3 | Emptiness + active projection + names unresolved in **selected** sections | Document-level fields, once the selected table filters to zero rows |
| 4 | …resolved across **all** sections | Any section's column validated any other section's field projection |
| 5 | …**and** the kind must match | current |

richlander:
## Round 6 addendum — MAI-Code

**Not upheld as blocking.** It is a real and fair observation, but it is a pre-existing gap of a different shape than this gate, and acting on it here would break behavior this PR already pins as correct.

**The report:**

```
type --platform System.Text.Json -S Classes --fields Source --tsv
```

renders the normal `Classes` table and exits 0, with the `Source` projection silently ignored — `Source` being advertised by `-D "API Info"` and not by `-D Classes`.

**Verified, with three corrections to the framing.**

1. **Byte-identical on base and head** — 3180 bytes, exit 0, on both binaries. MAI states this itself. Not a regression, and not something round 5's widening or round 6's kind matching changed in either direction.

2. **`Source` is not special.** Any `--fields` name behaves identically here, including one that exists nowhere in the document:

   | Invocation | Base | Head |
   | --- | --- | --- |
   | `-S Classes --fields Source --tsv` | exit 0, 3180 B | identical |
   | `-S Classes --fields Version --tsv` | exit 0, 3215 B | identical |
   | `-S Classes --fields NoSuchFieldAtAll --tsv` | exit 0, 3224 B | identical |

   The cause is not name validity at all: the `Classes` table is **column**-projected, so a `--fields` projection has nothing to apply to and the table renders unchanged. That is the same upstream shape as the `--fields <valid> --columns <bogus>` non-action already disclosed on this PR.

3. **It is not silent.** markout emits `Note: 1 field has no data: NoSuchFieldAtAll` on stdout in every one of these cases.

**Why not fix it here.** This gate fires on an **empty render**, deliberately and by construction — that ordering is what keeps the schema's blind spots (`--columns Field`/`Value`, synthesized and listed in no schema) out of reach, and it is what four earlier rounds of false positives established. What MAI is asking for is a different check: *did the projection actually apply?* That is not a narrowing of this one, it is a general capability that has to come from markout reporting per-section whether projection was honored, and it would apply to every command rather than to the `type` listing.

It would also contradict a case this PR already pins as legitimate, on the reviewers' own findings from round 3:

```csharp
// Unmatched against the section, but the section's own table is not field-projected, so this
// renders exactly as it did before and must keep exiting 0.
[InlineData(new[] { "-S", "Classes", "--fields", "NoSuchField" }, "## Classes")]
```

Rolled into **#3651**, which already tracks the unscoped `--fields` case and markout's `WriteFieldsTable` early return — the same root cause. Not patched here, for the same reason as the round-5 non-action: five rounds of corrections to this gate came from making it guess at name validity, and a sixth guess to cover an upstream projection-application gap is that mistake again.

### Round 6 status

- **GPT-5.6 — upheld and fixed** in `fedb068a3` (kind matching). See the reconciliation above.
- **MAI-Code — not upheld**, disposed above and tracked on #3651.

Head is now `fedb068a3`, which neither reviewer has seen. Round 7 will run there once CI is green.

richlander:
## Round 7 reconciliation — head `7062a9cb5`

**Both reviewers found a real defect, and both were in round 6's kind-matching rewrite rather than in the slice itself.** Both are fixed.

### GPT-5.6 — wildcard projections were rejected

```
type --library <local.dll> -S "API Info" --fields 'Ver*' --tsv
type --platform System.Net.Http -S Classes -t 'NoSuchType*' --columns 'Ty*' --jsonl
```

Base exits 0; head exited 1 with `No fields matched projection: Ver*`. **A genuine regression**, and the most serious finding since round 1 — `Ver*` selects `Version`, and `README.md:433` documents wildcard field projections.

The cause is precise and worth naming: markout's `ValidateProjection` **is wildcard-aware**. Round 5 used it and was correct here. Round 6 replaced it with `Discover` + `HashSet.Contains` to get at `SchemaItem.Kind`, and silently dropped wildcard support in the process. Verified directly:

```
ValidateProjection("API Info", ["Ver*"])   -> Resolved=[Ver*]   Unresolved=[]
ValidateProjection("API Info", ["Bogus*"]) -> Resolved=[]       Unresolved=[Bogus*]
```

**Fixed without reimplementing the glob.** GPT suggested using the renderer's wildcard-aware matcher while keeping the kind constraint, which is exactly right. The wanted-kind names are collected into a throwaway single-section `DocumentSchema` and matched with `ValidateProjection`, so markout stays the only matcher and cannot drift from the one that actually performs the projection.

### MAI-Code — stable names are not projection names

```
type --platform System.Text.Json -S "API Info" --fields Assembly --count
```

`Assembly` is the stable name of `Library`. The gate accepted `item.StableName` alongside `item.Name`, but markout does not project by stable names — `--fields Assembly` renders nothing on base and on head. Byte-identical on base, so a hole this PR opened in round 6 and failed to close, not a regression against `main`; it was `known.Add(item.StableName)`, defensive over-inclusion of exactly the kind that has gone wrong in every round.

Dumped the whole schema to size it. Only three items differ from their stable name at all:

| Section | Name | StableName |
| --- | --- | --- |
| API Info | `Library` | `Assembly` |
| API Info | `TFM` | `Tfm` |
| Type Forwarders | `Target Library` | `TargetLibrary` |

`TFM`/`Tfm` differ only by case and the comparison is case-insensitive, so it is inert. `Type Forwarders` is not a selectable section in this view. **`Assembly` was the only reachable one.** Dropped.

### Verification — all 21 cases raised across rounds 1–7, base vs head

| Group | Result |
| --- | --- |
| Wildcards `Ver*`, `Ty*`, `*` (5 forms, incl. GPT's two exact repros) | **identical to base** |
| Wildcards matching nothing (`Bogus*`, `Zzz*`) | correctly exit 1 |
| Stable name `Assembly` (3 forms) | correctly exit 1 |
| Wrong kind `--fields Type` (`--tsv`, `--plaintext`) | correctly exit 1 |
| Rounds 1–5 false positives (`--columns Field`/`Value`, `--fields Types`/`Version`/`Library`, `-S Interfaces`, both zero-row-filter cases) | **identical to base** |
| `-S --fields Bogus` | correctly exit 1 |

- Suite **2843**, 1 known pre-existing failure, 14 skipped.
- Differential against base `f42d8e586`: **275 / 154 / 7 — unchanged.**
- **Mutation:** reverting to set membership fails exactly the three positive wildcard cases; re-adding `StableName` fails exactly the stable-name test.

### One of my own new tests was nearly vacuous, again

I first pinned the wildcard fix with `--fields *`. That case passes for the wrong reason: `*` matches fields that **have values**, so the render is non-empty and the gate is never consulted. The mutation run exposed it — reverting to set membership failed `Ver*` but not `*`. Replaced with `TF*` and `?ersion`, both of which render empty against the test assembly and so actually reach the gate, and `?ersion` additionally covers the `?` metacharacter that set membership also broke. This is the second time in this stack that a test I wrote was satisfied by an unrelated cause, and both times only mutation testing found it.

### GPT's other checks

No synthesized `Field`/`Value` empty-render case found independently, matching my 60-combination sweep. All projectable leaf schema items are `field` or `column`; `@All` is a `category` and sections are `section`, so no projectable item carries an unmatched kind. All listing write paths confirmed gated; `--json` rejects field/column projections before writing. The four wrong-kind tests confirmed non-vacuous. GPT agreed dropping `StableName` is correct. It did **not** reconstruct the differential or the mutation claim.

**Note on process:** GPT's first round-7 attempt returned an empty response. I did not count it as clean and re-ran the round; the findings above are from the re-run.

### Running total

| Round | Mechanism | How it was wrong |
| --- | --- | --- |
| 1 | Name pre-check copied from the single-type path | Rejected `--columns Field`/`Value`, `--fields Types` |
| 2 | Emptiness of the render | Rejected valid-but-empty fields, and empty sections with no projection |
| 3 | Emptiness + active projection + names unresolved in **selected** sections | Document-level fields, once the selected table filters to zero rows |
| 4 | …resolved across **all** sections | Any section's column validated any other section's field projection |
| 5 | …**and** the kind must match, via `Discover` + set membership | Dropped wildcard support; accepted internal stable names |
| 6 | …matched by markout's own `ValidateProjection` over a kind-filtered schema | current |

richlander:
## Round 8 — clean from both reviewers at `7062a9cb5`

First clean round in eight. Both reviewers initially returned a bare "No significant issues found", which after seven consecutive rounds of real findings — and one completely empty response in round 7 — is indistinguishable from a round that did not run. I asked both to substantiate before acting on it. Both did, at the correct head.

### GPT-5.6 — substantiated, fuller pass

Confirmed `7062a9cb5` in `/tmp/rev-4d-gpt`. Independently ran the **non-vacuity check** I most needed, since I had shipped a nearly-vacuous case the round before:

```
'Ver*'     exit 0, stdout ''   Note: 1 field has no data: Ver*
'TF*'      exit 0, stdout ''   Note: 1 field has no data: TF*
'?ersion'  exit 0, stdout ''   Note: 1 field has no data: ?ersion
'Bogus*'   exit 1              Error: No fields matched projection
'Zzz*'     exit 1              Error: No fields matched projection
```

All three positive cases genuinely render empty against `TestAssemblyPath`, so each actually reaches the gate rather than passing for an unrelated reason. Also ran the focused tests (10 passed) and the **full suite independently: 2843 total, 1 failure, 14 skipped**, confirming the sole failure is the known `newtonsoft.json.nuspec` casing one. Explicitly did not: reconstruct the differential, run the mutations, or repeat every earlier target/format combination.

### MAI-Code — substantiated, lighter pass

Confirmed `7062a9cb5` in `/tmp/rev-4d-mai`. Independently ran the same three empty-render checks via `--plaintext` (all exit 0, empty stdout) and the base-vs-head mixed-projection comparison. Stated plainly that it did **not** run the full suite, the broader format/target matrix, or the stable-name regression — a lighter pass than its previous rounds, which it said so rather than implying otherwise. GPT's pass and my own sweep cover those gaps.

### "Some" vs "all" — both reviewers agree, and it matches base

```
type --library <test.dll> -S "API Info" --fields Version,Bogus --count
```

Base and head both exit 0 and print `0`. `Resolved.Length > 0` is the right rule: a projection with one valid and one garbage name must not fail, because the valid name was honored. Failing only when **nothing** resolves is also what the existing diagnostics contract does. Confirmed identical to base in three orderings (`Version,Bogus`, `Bogus,Version`, `Ver*,Bogus`), and the all-invalid case `Bogus*,Zzz*` correctly exits 1 at head where base exited 0.

### My own round-8 sweep, beyond what the reviewers ran

Attacking the throwaway-schema probe on the axes I flagged as riskiest:

| Probe | Result |
| --- | --- |
| Names with spaces — `--columns "Target Library"` | **identical to base**, exit 0. The space does not break `Add`. |
| `--fields "Target Library"` | exits 1 — correct: it is a **column**, so wrong kind, same class as `--fields Type` |
| Character classes `[Vv]ersion`, `V[a-z]rsion` | exit 1 — **correct**: markout supports only `*` and `?`, so these match nothing even against a target that *has* a version |
| Backslash escape `Ver\*` | exit 1 — correct, markout does not support escaping |
| `?` alone | exit 1 — correct, no field has a one-character name |
| `**`, `V*n`, `*rsio*`, `Ver?*`, `Sourc?`, `Metho*`, `Prop*`, `Lib*` | **all identical to base** |
| Whitespace forms — `" Version"`, `"Version   "`, `"   "`, `""` | **all identical to base** |
| Wrong-kind wildcards `Descript*`, `Membe*`, `Kin*` (all `Classes` columns) | exit 1 — correct, consistent with `--fields Type` |

The character-class result is worth stating explicitly because it looks like a regression and is not: I verified against `--platform System.Text.Json`, which **does** have a version, and `[Vv]ersion` still matches nothing there while `Ver*` renders normally. markout's matcher supports `*` and `?` only, so those patterns were never projecting anything — base simply printed nothing and exited 0, which is the defect.

### Final state

| | |
| --- | --- |
| Head | `7062a9cb5` |
| CI | green, `ci-required` passing, MERGEABLE/CLEAN |
| Suite | **2843**, 1 known pre-existing failure, 14 skipped — reproduced independently by GPT-5.6 |
| Differential vs base `f42d8e586` | **275 identical / 154 diffs / 7 vacuous**, single-`type` 0 diffs |
| Review | Two families clean at a fixed head: **GPT-5.6** and **MAI-Code** |

### Non-actions carried to follow-up, unchanged

- `--fields <valid> --columns <bogus>` silently ignores the unmatched column. Identical on base; markout's `WriteFieldsTable` returns early before `ComputeColumnMap`. **#3651.**
- `-S Classes --fields <anything>` renders the table and exits 0 with the field projection ignored, because that table is column-projected. Identical on base; not specific to any name. **#3651.**
- `package --columns Name` crashes identically on both binaries. Pre-existing. **#3686.**
- `type <unresolved-prefix> -v:q` prints no identity line — pre-existing verbosity floor, three sites in `TypeCommand`, present on base. Disclosed, deliberately not reverted.

richlander:
Ready to merge.

Eight review rounds, seven of which found a real defect. Head `7062a9cb5`, CI green, MERGEABLE/CLEAN, and clean at this fixed head from two different model families (GPT-5.6 and MAI-Code), both substantiated with commands and observed output.

Suite 2843 with 1 known pre-existing failure — reproduced independently by GPT-5.6. Differential against base `f42d8e586`: 275 identical / 154 diffs / 7 vacuous, single-`type` 0 diffs.

Everything listed under non-actions above is pre-existing, byte-identical on base, and tracked on #3651 / #3686 as non-blocking follow-up.

PR #3684, opened 2026-08-02
Add trust and rendering axes to mdi for artifact text

richlander:
Builds the trust and rendering axes that `docs/design/untrusted-data-threat-model.md` has specified since #3636 and recorded as unbuilt.

Stacked on #3679 — review that first; this targets its branch and will retarget to `main` once it merges.

## The default refuses

`mdi` rendered everything it found, neutralizing control characters and continuing. That is sanitizing, and the same document opens with **"reject, do not sanitize."** Now, when an assembly carries text a terminal would act on, `mdi` exits non-zero:

```text
Error: '/tmp/MetaHostile.dll' contains text that is not safe to render as it is.
  #Strings:0x2d carries U+202E (Format) at index 7.

  --show-untrusted-text     render it inertly; every such scalar is spelled, never emitted
  --show-untrusted-text --dangerously-print-raw
                            render it verbatim; unsafe, and best redirected to a file
```

The message never contains the text. The diagnostic goes to stderr, which is read on a terminal and is almost never piped through whatever treatment stdout got — quoting the characters would deliver the payload by the one route the user cannot redirect, in the mode chosen specifically to avoid that. The coordinate is spelled the way `--heap` accepts it, so the way forward is a command to paste rather than assemble; a test pulls the coordinate out of the diagnostic and feeds it back in.

## The flags compose

They are two axes, not a three-way choice.

| | trust | rendering | result |
| --- | --- | --- | --- |
| *(default)* | refuse | — | exits 1, names the coordinate |
| `--show-untrusted-text` | continue | encoded | today's output, byte for byte |
| both flags | continue | raw | the artifact's own spelling |
| `--dangerously-print-raw` alone | — | — | **rejected** |

Raw output costs two separately named mistakes. The raw flag alone is rejected rather than silently ignored: refusal comes first, so it would otherwise change nothing while appearing to.

An earlier draft of this PR made the flags *mutually exclusive* — the exact inverse — and the design doc caught it.

## Nothing changes for ordinary assemblies

Refusal is checked on the raw text before encoding, and when it succeeds it is byte-for-byte identical to containment; the modes differ only in whether they can fail. Default output equals `--show-untrusted-text` output on every clean assembly measured, including `System.Private.CoreLib` at 5.4 MB of output. That equivalence is gated.

`dotnet-inspect` is unaffected. The library default stays `Contain`: containment is a safety property `ILInspector.Metadata` owes every caller, refusing is a policy only a caller can choose, and an internal projector scan must never refuse.

## What raw actually promises

That `mdi` adds no encoding — not that every scalar reaches the stream, because the format still keeps itself well formed. Measured:

| format | behavior under `--dangerously-print-raw` |
| --- | --- |
| md | carries everything |
| tsv | replaces U+2028/U+2029, which it cannot carry in a record |
| jsonl | escapes scalars below U+0020 per RFC 8259 — these **decode back**, so they are raw |

The gate decodes per format before asserting. It has to: `mdi` spells U+009F as `\u009F` and JSON escapes a real U+009F to the same six characters, so on the wire the tiers are indistinguishable and differ only in what they parse back to.

## Tests

35 new cases driving `MdiCommand.Invoke(string[], TextWriter?, TextWriter?)` with an argv, because the mode is chosen by option parsing and the CLI default deliberately differs from the library's — nothing below the argv layer can observe either. `Invoke` gained optional writers so this needs no `Console` redirection, which is process-global and would make parallel execution unsound.

Five mutations, each caught by a different case for a different reason:

| mutation | failures | reason |
| --- | --- | --- |
| default no longer refuses | 9 | exit code |
| raw falls back to containment | 7 | `still contained \^[` |
| diagnostic echoes the scalar | 6 | `echoed U+001B raw` |
| axes collapsed (trust opt-out implies raw) | 9 | `doing the rendering axis's job as well as its own` |
| raw flag alone silently accepted | 1 | not rejected |

`IsNonGraphic` moved onto `HostileAssemblyFixture` so one definition of "dangerous" serves every gate over that payload. Two copies would be two oracles, which is the failure this corpus exists to prevent.

Suites: MetadataRendering 184/0, Metadata 1283/0, InertText 330/0, CLI 2817 with only the known local #3592.

## Still unbuilt

`--survey` — refusal stops at the first violation rather than reporting all of them, though `InertString.IsPermitted` already returns a `ScalarViolation` shaped for it. Both status sections say so.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,10 +7,14 @@ on:
   # A PR retargeted from `main` onto a long-lived feature branch silently stops
   # matching `[main]`, so it reports no checks rather than failing ones -- and
   # once `ci-required` is a required check, "no checks reported" is
-  # indistinguishable at a glance from "not run yet". `feature/**` keeps the
-  # lower half of a stack covered while its base is still in review.
+  # indistinguishable at a glance from "not run yet".
+  #
+  # The list must therefore cover every prefix a *base* can have, which is every
+  # prefix a branch can have, since any branch under review can be stacked on.
+  # `feature/**` alone left `fix/**` uncovered, and that is not a hypothetical
+  # gap: #3684 opened onto `fix/3628-...` and reported no checks at all.
   pull_request:
-    branches: [main, 'feature/**']
+    branches: [main, 'feature/**', 'fix/**']
 
 env:
   DOTNET_NOLOGO: true
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -847,12 +847,17 @@ channel the check just closed.
 ### Status
 
 The bounded-traversal budgets, execution-incapable parsing, and opt-in heaps are
-implemented, with their gates named above. The identifier allow lists, the
-trust and rendering axes, `--survey`, the visual-encoding spelling and its
-decoder, and the failure-message rule are the **target model** and are not yet
-implemented; today the projector neutralizes control characters unconditionally
-and continues, and nothing validates an identifier's grammar at all. See the
-threat model's open work.
+implemented, with their gates named above. So are the visual-encoding spelling
+(#3636, extended to Unicode general categories in #3628), the failure-message
+rule, and both the trust and rendering axes, which `mdi` exposes as
+`--show-untrusted-text` and `--dangerously-print-raw`.
+
+The identifier allow lists and `--survey` remain the **target model**: nothing
+validates an identifier's grammar, and refusal stops at the first violation
+rather than surveying them all. The projector itself still encodes and
+continues by default — the axes are a property of the command line, not of the
+library, because containment is a guarantee every caller gets while refusing is
+a policy only a caller can choose. See the threat model's open work.
 
 ## Prior art: `mdv` / `MetadataVisualizer`
 
--- a/docs/design/metadata-table-projection.md
+++ b/docs/design/metadata-table-projection.md
@@ -991,6 +991,22 @@ raw `string` — the shape `HardenedJson` already uses, where the guarantee come
 from choosing the type rather than from remembering the call. Auditing then
 becomes a search for a type rather than an argument about coverage.
 
+That durable fix shipped in #3687: `HeapReference.Text`/`.Preview`,
+`HandleRef.Display`, `MetadataImageOverview.MetadataVersion` and
+`Malformed.Detail` are `InertString`, and there is no conversion admitting a
+`string` into any of them. The enforcement is the compiler rather than a gate.
+
+It also decides where the rendering axis can act. Because the model cannot hold
+untreated text, `--dangerously-print-raw` cannot be served by carrying a second
+raw copy through the projection; raw output is produced at the sink by running
+the encoder backwards, which the encoding supports by construction — it is
+lossless and invertible, and a literal backslash is always rewritten so the
+inverse is unique. The trust axis is unaffected and still acts upstream, inside
+`ContainCellText`, because refusal asks about the raw text and must therefore
+see it before anything is spelled. That is why `ContainCellText` still takes a
+`string` while returning an `InertString`: the parameter is the last place the
+untreated value legitimately exists.
+
 That is also why the property needs its own gate rather than a comment.
 `MdiContainmentTests` splices a payload spanning all three control ranges the
 projector recognizes — a live `ESC [ 3 1 m` sequence, `BEL`, `DEL`, and a C1
--- a/docs/design/untrusted-data-threat-model.md
+++ b/docs/design/untrusted-data-threat-model.md
@@ -768,6 +768,32 @@ and TSV replaces the line and paragraph separators, which it cannot carry in a
 record. Markdown carries everything. So the raw mode promises that `mdi` adds
 no encoding of its own, not that every scalar reaches the stream.
 
+**Raw output is produced by the decoder, at the sink.** Since #3687 the
+projection cannot hold untreated text at all: its text-bearing fields are
+`InertString`, and no conversion admits a `string` into one. That closes off the
+obvious implementation — keeping a second, raw copy of every value beside the
+contained one — and forces the honest one, which is to run the encoding
+backwards at the moment of printing. This is the `vis`/`unvis` pairing named
+below rather than a workaround for it: the encoding is lossless and invertible
+precisely so that a decoder can exist, and having the decoder is what makes raw
+output a *rendering* choice instead of a property of the model. A literal
+backslash is always rewritten on the way in, which is what keeps the inverse
+unique. Refusal is unaffected and still happens upstream, against the raw text,
+because the question it asks — does this artifact carry something concerning —
+is about the artifact rather than about the spelling.
+
+Placing the decode after the character budget also buys a property the earlier
+implementation lacked: **both modes cut the same value at the same point.**
+Bounding raw text separately, in its own units, made the rendering axis quietly
+a *content* axis too, so that asking for a different spelling changed how much
+of the value you saw — a subtler form of exactly the collapse this section
+warns about. `MdiUntrustedTextModeTests.RawAndEncodedRenderingsShowTheSamePrefix`
+gates it by re-encoding the raw cell and comparing it to the encoded one,
+running the encoder forward so the assertion is not a restatement of the
+decode. On an artifact carrying nothing that needs containment the three
+tiers are byte-identical; measured over a 2,031,325-byte assembly, all three
+agree exactly.
+
 The axes are independent, and that is the design. Visual encoding is the
 default on **every** artifact-text path, including underneath the trust-axis
 skip — which is precisely what makes that skip defensible: it means "do not
--- a/docs/design/untrusted-data-threat-model.md
+++ b/docs/design/untrusted-data-threat-model.md
@@ -720,14 +720,20 @@ Artifact text can contain Markdown delimiters, newlines, terminal control
 characters, URLs, or prompt-like instructions. Renderers must preserve output
 structure and must not interpret inspected text as authority.
 
-> **Status.** The two axes below are the **target model**, not current
-> behavior. Today the metadata projector neutralizes control characters
-> unconditionally and continues, with no flags on either axis. See open work
-> item 10, and
-> [metadata-table-projection.md](metadata-table-projection.md#status).
+> **Status.** Both axes are built in `mdi`, which is the reference consumer:
+> the default refuses, `--show-untrusted-text` opts out of the trust axis, and
+> `--dangerously-print-raw` additionally opts out of the rendering axis. The
+> rendering axis rests on `InertText.InertString` (#3636), extended to Unicode
+> general categories in #3628.
 >
-> The rendering axis has a primitive as of #3636: `InertText.InertString`
-> supplies the encoded default. The flags, and the trust axis, remain unbuilt.
+> Two things remain unbuilt. **Survey mode** is not implemented: refusal stops
+> at the first violation rather than reporting every one, though
+> `InertString.IsPermitted` already returns a `ScalarViolation` shaped for it.
+> And **`dotnet-inspect` has neither flag**; the library default is to encode
+> and continue, so its behavior is the middle row of both tables. That default
+> is deliberate — containment is a safety property the library owes every
+> caller, while refusing is a policy only a caller can choose — but it means the
+> trust axis currently exists only where a command line can express it.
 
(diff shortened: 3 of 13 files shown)


richlander:
Rebased onto #3679's new head (`3edf70099`), which changed `ContainCellText` underneath this branch.

One conflict, in `ContainCellText` itself, since both sides rewrote it. Resolved toward theirs: bounding now happens at construction via `new InertString(TextPolicy.Field, value, maxChars)` and truncation is read from `IsTruncated` rather than derived by comparing lengths. That is the more correct half — a length is only meaningful within one encoding, so the comparison this branch had been using could report a value clipped mid-spelling as whole. The mode and origin parameters sit in front of it unchanged.

The raw path does not go through `InertString` at all, so it reports its own cut from `BoundRaw`. Noted on the `truncated` parameter: nothing is spelled under raw, so the divergence their doc comment describes cannot arise there.

Re-verified against the new base rather than assumed:

- MetadataRendering 184/0, Metadata 1301/0, InertText 362/0.
- All four command-line tiers behave as before, checked against a spliced assembly.
- Default output is still byte-identical to `--show-untrusted-text` on a clean assembly (`System.Text.Json`, 1,972,357 bytes).
- Truncation still fires in both modes, and the two bound different things correctly: at `--max-chars 12` containment cuts to `Hostile…` rather than divide the 6-character `\\u202E` spelling, while raw fits `Hostile<U+202E>INJE…` because the scalar costs one character there.

richlander:
Two interactions with #3679's ready-to-merge state, recorded here so neither is rediscovered.

**1. The "zero raw bytes" evidence becomes mode-conditional once this lands.**

#3679's confirmation round verified zero occurrences of `e2 80 ae`, `e2 80 a8`, or `1b` in CLI output at budgets 5/10/50/1000 across md/tsv/jsonl. That result stays true here for the default and for `--show-untrusted-text` — the containment gate in this PR asserts it independently — but it is no longer an invariant of the tool. With `--show-untrusted-text --dangerously-print-raw` those bytes are emitted, deliberately, and this PR's raw gate asserts they *are* present, because a flag that silently fell back to containment would satisfy every other assertion in the file.

So the property after this merges is "zero by default, and reachable only through two separately named opt-outs", not "zero". Anything re-running that byte sweep as a regression check needs to pass no flags, or it will measure the wrong claim.

**2. The projection-model follow-up has to keep refusal somewhere that still sees the raw string.**

The non-blocking follow-up — deriving `HeapReference.Text/Truncated`, `HandleRef.Display/DisplayTruncated` and `MetadataImageOverview.MetadataVersion/…Truncated` from a stored `InertString`, deleting `ContainCellText` — is a good slice and I am not touching it. But this PR puts the refusal check inside `ContainCellText`, ahead of encoding:

```csharp
if (mode is UntrustedTextMode.Refuse
    && !InertString.IsPermitted(TextPolicy.Field, value, out ScalarViolation? violation))
    throw new UntrustedTextException(origin, violation.Value.Index, violation.Value.Scalar, violation.Value.Category);
```

A constructed `InertString` can answer *whether* something was contained — `WasEncoded` — but it cannot reproduce the diagnostic. `IndexOfFirstEncoded()` is `Text.IndexOf('\\')`, an offset into the **encoded** text, and it carries no scalar and no category; the `ScalarViolation` that supplies `U+202E (Format) at index 7` comes from asking `IsPermitted` about the **raw** string. Deriving the fields from a stored `InertString` is therefore compatible with refusal, but only if the raw text is still reachable at the construction site — otherwise the refusal message degrades from a heap coordinate plus code point plus category to "something was encoded", which is most of what makes it actionable without echoing the text.

Rebase onto `3edf70099` is already pushed; this branch is on the reviewed head.

richlander:
Rebased onto `main` now that #3679 and #3687 are in, and resolved the raw-mode collision I flagged on #3687. I said I would come back with an implementation rather than three options; the answer turned out to be a fourth one, and the threat model had already specified it.

## What changed

#3687 makes the projection unable to hold untreated text — no conversion admits a `string` into an `InertString` — which is exactly the property `--dangerously-print-raw` needed to violate. Rather than weaken it, **raw output is now produced at the sink by running the encoding backwards**, and `BoundRaw` is gone.

That is not a way around the new typing. It is what the threat model already asked for:

> **BSD `vis(3)`** is where the vocabulary and the contract come from. It visually encodes arbitrary input into graphic characters only, and pairs the encoder with a decoder (`unvis`) so the transform is unique and invertible. **Encode-without-a-decoder is not this pattern.**

The decoder exists (`VisualEncoder.TryDecode`, public), the encoding is invertible by construction — a literal backslash is always rewritten, which is what makes the inverse unique — and `InertString.EnsurePermitted` already depends on the round trip, with a gate test asserting it holds. So raw is the decoder half of a transform the library was built to have, and it is a *rendering* choice rather than a property of the model. The doc's other cited precedent points the same way: `less -r` applies to the whole output, not to one addressed value.

Refusal is unchanged and still acts upstream against the raw text, because it asks about the artifact rather than the spelling. `ContainCellText` therefore keeps its `string` parameter while returning an `InertString` — that parameter is the last place the untreated value legitimately exists.

## It also fixed a flaw this PR had

Raw text used to be bounded in *its own units*. So at `--max-chars 12` the encoded rendering gave `Hostile…` while raw gave `Hostile␮INJE…`: the rendering axis was quietly changing **how much of the value you saw**, not just how it was written. That is a subtler form of the axis collapse the threat model calls a design error, and I had not noticed it — I described it in the original PR as a feature.

Deciding the spelling *after* the budget fixes it. Both modes now clip the same value at the same point:

```
--max-chars 12, encoded : Hostile\u2026
--max-chars 12, raw     : Hostile\u2026     (identical bytes)
```

`RawAndEncodedRenderingsShowTheSamePrefix` gates it by re-encoding the raw cell and comparing it to the encoded cell — the encoder run **forward**, so the assertion is not a restatement of the decode.

## And a vacuous test

`RawTextIsNeverCutThroughASurrogatePair` passed `--dangerously-print-raw` **alone**. That combination is rejected, so the command wrote nothing to stdout and the test swept forty budgets across an empty string, asserting nothing. It passed for the wrong reason through review, mutation testing and CI. It now passes both flags and asserts the output is non-empty.

While confirming the property still holds I checked what the decode does with surrogates, since `TryDecode` reconstructs whatever the encoder spelled:

```
in=0041 D800 0042  encoded=A\uD800B  decodeOk=True  roundTrips=True  loneSurrogateOut=True
in=006F 006B D83D DE00 ...  encoded=ok😀end  roundTrips=True  loneSurrogateOut=False
```

So raw output can reproduce a lone surrogate the *artifact* carried. That is not a regression — `BoundRaw` returned the original string and did the same — and it is faithful, which is what the tier is for. Worth recording rather than leaving implied.

## Evidence

- Three tiers against the hostile image: default refuses (exit 1, `#Strings:0x2d carries U+202E (Format) at index 7`), `--show-untrusted-text` gives `Hostile\u202EINJECTED`, both flags emit `e2 80 ae`.
- **All three tiers byte-identical** on a clean 2,031,325-byte assembly.
- Two mutations, each caught by a different set: dropping the decode reddens the raw-tier tests across all three views and formats; letting raw skip the truncation marker reddens only the same-prefix gate.
- Suites: MetadataRendering **187**, Metadata **1358**, InertText **362**, CLI **2817** with only the known local #3592.

The `ci.yml` commit adding `'fix/**'` stays, even though this PR now targets `main` and no longer needs it — it is the gap that left #3687 with no checks at all.

richlander:
Reviewed at `4b245f4a1` in an isolated worktree, with emphasis on the interaction with #3687 (which I authored and which this now sits on top of). **No blocking findings.** Two non-blocking, one of which I think is worth acting on before merge.

## The load-bearing property holds

The design decision I most wanted to attack is producing raw output by *running the encoder backwards* rather than keeping an untreated copy. That makes invertibility a security property, not just an aesthetic one: if the encoding were ambiguous, `--dangerously-print-raw` would not be showing the artifact's spelling, it would be **synthesizing** scalars the artifact never contained.

The specific attack is an artifact whose text *literally contains the escape syntax*. I built one — a type named with a real U+202E, another named with the seven ASCII characters `\u202E`, and a third with literal backslashes:

| artifact contains | `--show-untrusted-text` | `--dangerously-print-raw` |
| --- | --- | --- |
| real U+202E | `Real\u202EOverride` | `M-bM-^@M-.` — real UTF-8 `E2 80 AE` |
| literal `\u202E` | `Literal\\u202EText` | `Literal\u202EText` — **not** promoted to a real override |
| literal `\\` | `Back\\\\slash` | `Back\\slash` |

`VisualEncoder` escapes its own escape character (`value[i] == \ || !permits(scalar)`, VisualEncoder.cs:67), so the round-trip is total and the two cases stay distinguishable in both directions. This is the thing that makes the decoder approach sound instead of clever, and it is worth the doc comment it already has.

The converse also holds — refusal does **not** false-positive on a name that merely needs escaping. An assembly whose only unusual characters are `\`, `~`, and `^` renders at exit 0. "Needs escaping to round-trip" and "impermissible" are correctly distinct predicates.

## Other claims, verified

- **Refusal precedes containment.** `ContainCellText` scans raw `value` via `IsPermitted` before `new InertString(...)`. The ordering the doc claims is the ordering the code has.
- **Refusal is not weakened by the budget.** With `--max-chars 2` against a hazard at index 4 — where the clipped output could never have shown it — it still refuses. Correct: the question is what the artifact carries, not what would have been printed.
- **Byte-for-byte equivalence on clean artifacts.** `dotnet-inspect.dll`, all three formats, default vs `--show-untrusted-text`: identical, exit 0.
- **Per-format raw semantics are exactly as documented.** md carries real bytes; tsv substitutes U+2028; jsonl JSON-escapes. I checked the jsonl claim at parse level, since that is the one place contained and raw look alike on the wire:

  ```
  RAW       -> "Esc\u001BHere"   parses to  U+0045 U+0073 U+0063 U+001B ...
  CONTAINED -> "Esc\\^[Here"     parses to  U+0045 U+0073 U+0063 U+005C U+005E U+005B ...
  ```

  Raw decodes back to a real U+001B; contained decodes to the literal spelling. The claim that jsonl escapes "are raw" is true in the only sense that matters.
- **The diagnostic locates without quoting**, and its coordinate is directly usable: `#Strings:0x21` from the error pasted into `--heap` resolves to the right value.
- **Coverage of artifact-text paths.** String, UserString, handle display, and the version stamp all pass the mode. Guid and blob hex keep the default, correctly — both spellings are always graphic and cannot carry a hazard.
- **Library default stays `Contain`**, via optional parameters, so `dotnet-inspect` and existing callers are unaffected.

Suites at this head: MetadataRendering **187/0**, Metadata **1358/0**, InertText **362/0**.

## Non-blocking 1 — the CI fix does not implement its own stated principle

The comment added to `ci.yml` is right about the failure mode, and states the rule correctly:

> The list must therefore cover every prefix a *base* can have, which is every prefix a branch can have, since any branch under review can be stacked on.

It then adds `fix/**` only. `origin` currently carries **15+ distinct prefixes**:

```
47 feature/**   23 fix/**   6 docs/**   3 prototype/**   3 feat/**
 2 vendor/**     2 release/**   2 raise/**   1 scratch/**  1 review/**
 1 reports/**    1 proto/**     1 integrate/**  1 experiment/**  1 design/**
```

plus several with no prefix at all. So the same bug is one stacked PR away — and not hypothetically: **#3690 is on `docs/…` right now.** Anything opened against it reports no checks, which is the exact state this change exists to eliminate, and the state `AGENTS.md` calls terminal rather than pending.

Since the principle is "every prefix", `branches: ['**']` states it directly and cannot go stale. A prefix allow list has to be *maintained*, and the cost of a miss is silent.

## Non-blocking 2 — `Malformed.Detail` is contained but never refused

All 14 `Malformed` producers build their text with `InertString.Format(TextPolicy.Field, …)`, five of which splice a third party's `ex.Message`. That is contained, so there is no terminal exposure — but it never passes through `IsPermitted`, so `Refuse` cannot fail on it.

Low severity, and I would not hold the PR for it: SRM's messages are resource strings, and containment already covers the dangerous half. But `Refuse`'s whole value is completeness — "hear that an artifact carries a bidi override rather than receive a faithful rendering of one" — so the gap is worth either closing or naming in the status section next to `--survey`.

## Nit

The body's suite counts (MetadataRendering 184, Metadata 1283, InertText 330) predate #3687; actuals are 187, 1358, 362.

## Verdict

The design is sound and the risky part — reconstructing raw text by decoding — is safe for a reason the code establishes rather than assumes. I would land it after widening the CI filter, since that one is cheap and its failure mode is invisible.

Issue #3606, opened 2026-07-31
Log delegate carries raw URLs: retype Action<string> log to a treated-text value


richlander:
## Problem

`DotnetInspector.Packages` threads an `Action<string>? log` delegate through the package layer.
Roughly a dozen call sites interpolate a **raw** URL into it, indistinguishably from a treated one:

- `src/DotnetInspector.Packages/HttpRetryHelper.cs:133,138,151,174,180`
- `src/DotnetInspector.Packages/PackageExtractor.cs:572,587`

Example:

```csharp
log?.Invoke($"HTTP {methodName} {(int)statusCode} (not retryable): {url}");
```

`url` here is the untreated source URL. A source URL is untrusted input — it can come from a
`nuget.config` in a repository the user only cloned — and the log sink is a terminal.

This is **pre-existing on `main`**, not introduced by #3563. Verified on a clean checkout.

## Why it is filed separately

#3563 introduced `InertString`, a value that can only be built by applying a policy, and converted
the redaction chokepoint (`RedactUrl`, `RedactSensitiveUrlText`, `FeedFailure.Url`,
`DescribeFailure`, the unsupported-source message) to it.

Retyping the log delegate would make the compiler enumerate every one of these sites, exactly as
it did for `DescribeFailure` — which is how the untreated `packageName` hole was found there.

It was left out of #3563 because `Action<string> log` is declared in **27 files** across
`DotnetInspector.Packages`, `DotnetInspector.Services`, `dotnet-inspect`, `NuGetFetch` and
`ILInspector.Metadata`, over ~215 call sites. That is a mechanical but wide change, and it does
not belong in a fix for feed-failure reporting.

## Suggested approach

1. Introduce a log delegate that takes `InertString` rather than `string`.
2. Let the compiler enumerate the call sites.
3. At each, decide deliberately: `InertString.Format(TextPolicy.Prose, $"...")` for messages
   carrying untrusted data, and confirm the trusted-only ones really are trusted.

Step 3 is the value of the exercise. Most sites interpolate only trusted data (`methodName`,
`retryCount`, a `TimeSpan`), and the point is that today nothing distinguishes those from the ones
that do not.

## Related

- #3563 — introduced `InertString` and converted the redaction chokepoint
- #3601 — the spec; the "available as a value" obligation

PR #3687, opened 2026-08-02
Carry metadata text as InertString so a sink cannot be handed raw text

richlander:
**Stack position: slice 2 of 2.** Parent: #3679 (targets `main`). This PR targets `fix/3628-metadata-lens-inert-text` and will retarget `main` when #3679 merges.

## What changes

#3679 fixed the leak: metadata cells are contained by category through `InertText` instead of a hand-written control-character range. This slice makes that containment a property of the model rather than a discipline each producer keeps.

Every field of the metadata projection that can hold artifact-derived text was a `string`, and the containment helper took an `InertString` apart one line after building it:

```csharp
internal static string ContainCellText(string value, int maxChars, out bool truncated)
{
    var contained = new InertString(TextPolicy.Field, value, maxChars);
    truncated = contained.IsTruncated;
    return contained.ToString();
}
```

Its only real content is the choice of `TextPolicy.Field`; the rest exists to split the value into a string and a flag that three records then stored separately. Now:

```csharp
internal static InertString ContainCellText(string value, int maxChars)
    => new(TextPolicy.Field, value, maxChars);
```

| Field | Was | Now |
| --- | --- | --- |
| `HeapReference.Text` / `.Preview` | `string?` / `string` | `InertString?` / `InertString` |
| `HandleRef.Display` | `string?` + `DisplayTruncated` | `InertString?` (flag deleted) |
| `MetadataImageOverview.MetadataVersion` | `string` + `MetadataVersionTruncated` | `InertString` (flag deleted) |
| `Malformed.Detail` | `string` | `InertString` |

`Malformed.Detail` is included because it splices a third party's exception message into a diagnostic produced *only* while describing an image that has already proved malformed, and the renderer prints it raw as `!malformed: {Detail}`. Whether SRM ever puts artifact bytes into a message is not a property this repository can pin; typing the field means it does not have to.

`Scalar.Display` and `Flags.Decoded` stay `string` — formatted numbers and BCL enum names, never heap content.

**The enforcement is now the compiler.** There is no conversion admitting a `string` into any of those fields, which is a stronger gate than any test for "no untreated metadata text reaches a sink".

## The one flag that survives, and why

The obvious next step is to delete every truncation field and read `IsTruncated` off the text. That is right for `HandleRef` and `MetadataImageOverview`, where a character budget is the only way a value can become partial. It is **wrong** for `HeapReference`:

```csharp
int take = Math.Min(length, Math.Max(0, options.MaxPreviewBytes));
byte[] bytes = blobReader.ReadBytes(take);
```

A blob preview is bounded in *bytes*, before there is any text at all, and `Convert.ToHexString` then produces a **complete** spelling of the bytes that were read. So `Preview.IsTruncated` is `false` for a blob that lost most of its content, and deriving the flag would report those blobs as whole. Confirmed live:

```
--max-bytes 2   | method\u202EINJECTED | 2000…  |
--max-bytes 64  | method\u202EINJECTED | 200001 |
```

`HeapReference.Truncated` therefore stays stored, meaning "this projected value is not the whole value", with two causes feeding it and only one visible in the text.

This is the same shape as the `_sourceLength` finding in #3679: a length or a flag is only meaningful in the units that produced it.

## Evidence

`BlobTruncation_IsInvisibleToTheHexItProduces` gates the distinction, and names itself as that gate in its doc comment. Tampering `BlobCell` to derive the flag from the hex **compiles** and reddens it plus `BlobBudget_ProjectsBoundedHexPreviewWithExplicitTruncation` and `ReadHeapValue_HonoursTheBlobPreviewBudget`.

End-to-end against an assembly carrying `U+202E`, `U+2028` and ESC in type, field and method names — **0** raw hazard bytes across Markdown, TSV and JSONL for TypeDef, Field, MethodDef and #Strings. Positive control: the payload renders as `Hostile\u202EINJECTED\u2028NEXT\^[ESC`. Truncation intact through every retyped path — heap text `Hostile…`, handle display `TypeRef[1] (Syst…)`, blob `2000…`.

Suites: Metadata **1302**, MetadataRendering **149**, InertText **362**, Services **357**, Analysis **584**, CLI **2008** — 0 failed. `-t:Rebuild` 9 warnings, matching baseline.

## Two things the retype surfaced

**A pinned invariant dissolved.** `MetadataImageOverview` was on the positional-record property-order deny list, with a comment explaining that `MetadataVersionTruncated` had to be declared next to `MetadataVersion` while being optional forced it last in the constructor — "the two orders cannot agree". They cannot agree *while the pair is split*. Uniting them removes the divergence, and the set-equality gate correctly failed until the entry was removed.

**A fixture was asserting an impossible state.** A renderer test built `Display: "", DisplayTruncated: true` — empty text claiming to have been cut. With `InertString` that cannot be written; the fixture now bounds real text to a real budget, so it exercises a value the projector can actually produce.

## Residual — deliberately not in this PR

- Containment boundaries beyond the metadata lens remain un-enumerated (`docs/design/inert-text.md` tracks 359 row-type columns via #3463 and 97 `Action<string>` sites via #3606).
- `MetadataRowReferences.ColumnName` stays `string`: column names are product-authored, not artifact-derived.

Refs #3628.

### DIFF
--- a/docs/design/inert-text.md
+++ b/docs/design/inert-text.md
@@ -235,6 +235,104 @@ text. Establishing that list is a prerequisite for the guarantee, not a detail
 of it — a chokepoint you have not finished enumerating is a chokepoint you
 cannot claim.
 
+## Carrying a contained value to the sink
+
+Containing text at the boundary answers *whether* a value was treated. It does
+not, on its own, get the value to the printer intact. The metadata table
+projection is the worked example, and it started out doing this:
+
+```csharp
+internal static string ContainCellText(string value, int maxChars, out bool truncated)
+{
+    var contained = new InertString(TextPolicy.Field, value, maxChars);
+    truncated = contained.IsTruncated;
+    return contained.ToString();
+}
+```
+
+That takes an `InertString` apart one line after building it, and it is worth
+being precise about the cost, because "it still returns treated text" is true and
+is not the point.
+
+**The type stops guarding the moment it is unwrapped.** A `string` field cannot
+say whether it was treated, so every later edit that assigns to it is
+unconstrained, and the only way to audit the field is to re-trace its producers.
+Carrying `InertString` instead makes that a compiler question: there is no
+conversion into the type that does not apply a policy, so a field of that type
+has no untreated inhabitants.
+
+**Splitting the pair invites the halves to disagree.** Truncation is not
+decoration on the text, it is the difference between a prefix and a whole value,
+and a sink that loses it renders a clipped name as though it were complete. Two
+fields can drift; one value cannot. `HandleRef` carried a `Display` and a
+`DisplayTruncated` that could be set independently — and a test constructed
+`Display: "", DisplayTruncated: true`, a state the projector cannot produce.
+Once `Display` is an `InertString`, that fixture has to be written as bounding
+real text to a real budget, because a value that was never cut cannot claim it
+was.
+
+So the rule is: **contain once, carry the contained value, and un-pair only at
+the sink.** The producer names the policy and nothing else; the sink decides what
+a partial value looks like, because notation is the sink's business and no
+producer should be guessing at an ellipsis.
+
+```csharp
+internal static InertString ContainCellText(string value, int maxChars)
+    => new(TextPolicy.Field, value, maxChars);
+```
+
+```csharp
+InertString text = ContainCellText(raw, options.MaxStringChars);
+return new MetadataValue.HeapReference(
+    HeapKind.String, HeapOffset(handle), raw.Length, text, text, text.IsTruncated);
+```
+
+### A partiality flag survives only where the text cannot know
+
+The tempting next step is to delete every truncation field and read
+`IsTruncated` off the text. That is right wherever the character budget is the
+only way a value can become partial, which is why `HandleRef.DisplayTruncated`
+and `MetadataImageOverview.MetadataVersionTruncated` are gone.
+
+It is wrong for `HeapReference`. A Blob preview is bounded by a **byte** budget,
+upstream of any text at all:
+
+```csharp
+int take = Math.Min(length, Math.Max(0, options.MaxPreviewBytes));
+byte[] bytes = blobReader.ReadBytes(take);
+```
+
+The hex that follows is a *complete* spelling of the bytes that were read, so its
+`IsTruncated` is `false` for a blob that lost most of its content. Deriving the
+flag there would report those blobs as whole. `HeapReference.Truncated` therefore
+stays a stored field meaning "this projected value is not the whole value", with
+two causes feeding it and only one of them visible in the text.
+
+The general form: a length or a flag is only meaningful in the units that
+produced it. `InertString` refuses to remember a source *length* for the same
+reason — re-spelling under a stricter policy makes text longer, so a length
+carried across a re-encode is compared against a different unit.
+
+### What this buys, and what still carries `string`
+
+Every field of the metadata projection that can hold artifact-derived text is
+now `InertString`: `HeapReference.Text` and `.Preview`, `HandleRef.Display`,
+`MetadataImageOverview.MetadataVersion`, and `Malformed.Detail`. The last one is
+included because it splices a third party's exception message into a diagnostic
+about an image that has already proved malformed; whether SRM ever puts artifact
+bytes into a message is not a property this repository can pin, and typing the
+field means it does not have to.
+
+`Scalar.Display` and `Flags.Decoded` remain `string` on purpose. They hold
+formatted numbers and BCL enum names — `0x{rva:X8}`, `TypeCode.Boolean` — and
+never heap content, so containing them would add ceremony without removing a
+hazard.
+
+The gate for "no untreated metadata text reaches a sink" is therefore the
+compiler: there is no conversion that would let a `string` into any of those
+fields. That is a stronger enforcement than a test, and it is the reason to
+prefer carrying the type over re-checking the text.
+
 ## Holding a value is not the same as being able to reverse it
 
 A type answers what a *sink* accepts. It does not answer what a *file* can do,
--- a/src/DotnetInspector.MetadataRendering/MetadataProjectionRenderer.cs
+++ b/src/DotnetInspector.MetadataRendering/MetadataProjectionRenderer.cs
@@ -1,5 +1,6 @@
 using System.Reflection.Metadata.Ecma335;
 using ILInspector.Metadata;
+using InertText;
 using Markout;
 
 namespace DotnetInspector.MetadataRendering;
@@ -334,11 +335,7 @@ public static class MetadataProjectionRenderer
         // A truncated stamp must carry the ellipsis for the same reason a
         // truncated handle display does: a prefix must never be mistaken for the
         // whole value.
-        Add(
-            "Metadata version",
-            overview.MetadataVersionTruncated
-                ? overview.MetadataVersion + Ellipsis
-                : overview.MetadataVersion);
+        Add("Metadata version", Display(overview.MetadataVersion));
         Add("Metadata kind", overview.Kind.ToString());
         Add("Has assembly manifest", overview.IsAssembly ? "yes" : "no");
         Add("Metadata offset", overview.MetadataOffset.ToString());
@@ -856,11 +853,28 @@ public static class MetadataProjectionRenderer
         _ => throw new InvalidOperationException($"Unhandled metadata value: {value.GetType().Name}"),
     };
 
+    /// <summary>
+    /// Turns a contained value into display text, marking a partial one so it can never
+    /// be read as whole.
+    /// </summary>
+    /// <remarks>
+    /// This is the boundary the projection carries <see cref="InertString"/> up to. The
+    /// model keeps the text and its truncation together precisely so that the decision
+    /// made here — what a clipped value looks like — is made once, by the sink that knows
+    /// its own notation, rather than by every producer guessing at it.
+    /// </remarks>
+    static string Display(InertString text, bool truncated)
+        => truncated ? text + Ellipsis : text.ToString();
+
+    /// <inheritdoc cref="Display(InertString, bool)"/>
+    static string Display(InertString text) => Display(text, text.IsTruncated);
+
     static string FormatHeap(MetadataValue.HeapReference heap)
     {
         // The Blob heap carries no decoded text; its bounded hex preview stands in.
-        string body = heap.Text ?? heap.Preview;
-        return heap.Truncated ? body + Ellipsis : body;
+        // Truncation comes off the reference rather than the text because a blob's
+        // preview is bounded in bytes, which the hex spelling cannot report.
+        return Display(heap.Text ?? heap.Preview, heap.Truncated);
     }
 
     static string FormatHandle(HandleRef reference)
@@ -870,17 +884,20 @@ public static class MetadataProjectionRenderer
 
         string target = $"{reference.TargetTable}[{reference.TargetRowId}]";
 
+        if (reference.Display is not { } display)
+            return target;
+
         // A truncated display must always carry the ellipsis so it is never
         // mistaken for a whole value — even when the budget clipped it to empty,
         // which must still render as "(…)" rather than a bare target that looks
         // like an unavailable display.
-        if (reference.DisplayTruncated)
-            return $"{target} ({reference.Display}{Ellipsis})";
+        if (display.IsTruncated)
+            return $"{target} ({Display(display)})";
 
-        if (string.IsNullOrEmpty(reference.Display))
+        if (display.IsEmpty)
             return target;
 
-        return $"{target} ({reference.Display})";
+        return $"{target} ({display})";
     }
 
     static string FormatRange(HandleRange range)
--- a/src/ILInspector.Metadata/ILInspector.Metadata.csproj
+++ b/src/ILInspector.Metadata/ILInspector.Metadata.csproj
@@ -16,6 +16,7 @@
   <ItemGroup>
     <InternalsVisibleTo Include="dotnet-inspect.Tests" />
     <InternalsVisibleTo Include="ILInspector.Metadata.Tests" />
+    <InternalsVisibleTo Include="DotnetInspector.MetadataRendering.Tests" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/ILInspector.Metadata/MetadataImageInspector.cs
+++ b/src/ILInspector.Metadata/MetadataImageInspector.cs
@@ -73,10 +73,11 @@ public static class MetadataImageInspector
     /// the stamp is a counted string read straight out of the image and this code
     /// does not get to assume the image conforms.
     ///
-    /// When the budget does bind, the overview reports it via
-    /// <see cref="MetadataImageOverview.MetadataVersionTruncated"/> rather than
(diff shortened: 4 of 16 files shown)


richlander:

Two full rounds at `9a5782225`, then a confirmation round at the moved head `11c65aff1`.

### Round 1 — two models, at `9a5782225`

| Reviewer | Blocking | Non-blocking | Central claim |
| --- | --- | --- | --- |
| GPT-5.6 Sol | 0 | 0 | Confirmed |
| Gemini 3.1 Pro | 0 | 1 | Confirmed |

Both reviewers were pointed directly at the load-bearing claim — that `HeapReference.Truncated` cannot be derived from the text while the other two flags could be deleted — and both attacked it independently rather than taking it on description.

**GPT-5.6 Sol** reproduced the divergence through the CLI:

```text
--heap #Blob:1 --max-bytes 2           Length: 3  Truncated: yes  Value: 2000…
--heap #Blob:1 --max-bytes 2147483647  Length: 3  Truncated: no   Value: 200001
```

and then closed the converse by walking every other heap: String/UserString take the full decoded value through exactly one character budget, GUID is a fixed complete spelling, handle display and the metadata version stamp each pass exactly one budget. It also checked something I had not: that `ILTokenResolver` can degrade a malformed method display to a best-effort base name — and correctly ruled it out, since that is not character clipping and was never what the deleted flags represented. Beyond the claim it verified all 14 `Malformed` producers use `InertString.Format`, found no production record-equality, dictionary, set, or dedup dependency on the retyped records, confirmed `default(InertString) == InertString.Empty`, and published both binaries NativeAOT.

**Gemini 3.1 Pro** ran the tamper. Replacing `Truncated: take < length` with `Truncated: preview.IsTruncated` builds cleanly and reddens exactly the three tests named in the PR body:

- `ReadHeapValue_HonoursTheBlobPreviewBudget`
- `BlobBudget_ProjectsBoundedHexPreviewWithExplicitTruncation`
- `BlobTruncation_IsInvisibleToTheHexItProduces`

That is the evidence that matters most here, because it demonstrates the gate is non-vacuous rather than merely present. It separately built a hostile assembly via `Reflection.Emit` carrying U+202E, U+2028, and ESC in types, fields, methods, and a `#US` literal, and hexdumped `mdi` output across `md`/`tsv`/`jsonl` to confirm zero raw hazard bytes with positive controls.

### The one finding, and what it exposed

Gemini found a `<see cref="MetadataImageOverview.MetadataVersionTruncated"/>` in `MetadataImageInspector.cs` still pointing at a member this slice deleted. Real, and fixed in `11c65aff1` along with two test doc comments in `OversizedMetadataVersionTests.cs` that still described the flag as the thing downstream keys off. The test bodies were already correct; only the prose was stale.

The interesting part is why a dangling `cref` survived a full `-t:Rebuild`, two review rounds, and CI. Roslyn only resolves `cref`s when it is emitting a documentation file, and `ILInspector.Metadata` does not enable one — so `CS1574` was never going to fire. Filed as #3688 rather than folded in: it is a `Directory.Build.props` change affecting every project, the 10 pre-existing violations it surfaces live in `ILInspector.Findings` and `SourceLinkFetch`, and it needs its own review round.

Worth stating plainly, because it is the same failure this stack exists to fix: a comment asserting an invariant, with nothing checking it.

### Round 2 — confirmation at `11c65aff1` (MAI-Code)

The head moved, so the new head got its own round. Verdict: the delta is genuinely comment-only, and the new prose is accurate. It re-derived the accuracy claim from source rather than accepting it — confirming `Describe` routes the stamp through `ContainCellText`, that `ContainCellText` is `new InertString(TextPolicy.Field, value, maxChars)`, and therefore that the character budget really is the stamp's only path to partiality. It confirmed the new `<see cref="InertString"/>` resolves under forced XML-doc generation, matched the updated comments against the assertions in both test bodies, and swept the tree for other stale references — finding only the deliberate past-tense mentions in `docs/design/inert-text.md`.

**And it corrected me.** My commit message for `11c65aff1` claims "no project in this repo sets `GenerateDocumentationFile`." That is false. Two do:

- `src/InertText/InertText.csproj:13`
- `src/DotnetInspector.HostileNameFixtures/DotnetInspector.HostileNameFixtures.csproj:30`

I had grepped only `Directory.Build.props` and `ILInspector.Metadata.csproj` and generalized from an absence I had not actually looked for. The mechanism for *this* bug is unaffected — the project holding the dangling `cref` still does not emit docs — but the blanket claim was wrong, and the correct version is sharper: the containment primitive itself has its doc comments checked, while every consumer that applies it does not. The gate exists where the design is authored and not where it is used. #3688 has been rewritten to say that; the commit message stands uncorrected in history, which is what this comment is for.

### Validation at `11c65aff1`

| Suite | Total | Failed |
| --- | ---: | ---: |
| `ILInspector.Metadata.Tests` | 1302 | 0 |
| `DotnetInspector.MetadataRendering.Tests` | 149 | 0 |
| `InertText.Tests` | 362 | 0 |
| `DotnetInspector.Services.Tests` | 357 | 0 |
| `ILInspector.Analysis.Tests` | 584 | 0 |
| `dotnet-inspect.Tests` (excl. `Speed=Slow`) | 2008 | 0 |

`dotnet build dotnet-inspect.slnx -c Release -t:Rebuild` — **9 warnings, 0 errors**, matching baseline. Independently reproduced by two reviewers.

### CI status — stated precisely

This PR reports **no checks at all** (`gh pr checks 3687` → "no checks reported"), a consequence of its base being #3679's branch rather than `main`. Per `AGENTS.md` that is terminal, not pending, so I am not waiting on it and **I am not claiming this slice is green**. Its evidence is the local validation above plus the parent's CI, which is green. I will re-check after this retargets `main` on #3679's merge, where anything newly scheduled does gate.

Both PRs are currently `MERGEABLE` / `CLEAN`.

### Residual

Unchanged from the PR body: containment boundaries beyond the metadata lens remain un-enumerated — 359 row-type columns via #3463, 97 `Action<string>` sites via #3606. Doc-comment validation now tracked by #3688.

richlander:
Ready to merge

Merge bottom-up: #3679 first, then confirm this slice retargets `main`, re-check `gh pr checks 3687` for anything newly scheduled, then merge.

richlander (commented):
Reviewed by reproducing the claims rather than reading them. Build clean (9 warnings, matching baseline), Metadata **1302**/0 and MetadataRendering **149**/0 locally. The design argument holds up and the retype is the right shape. One coordination problem, below, which is not about this PR being wrong.

## What I verified

**The blob distinction, by mutation.** Replacing `Truncated: take < length` with `Truncated: preview.IsTruncated` in `ReadBlobValue` **compiles** — so the type system does not protect this — and reddens exactly the three tests named in the description:

```
MetadataTableProjectionTests.BlobTruncation_IsInvisibleToTheHexItProduces      [FAIL]
MetadataTableProjectionTests.BlobBudget_ProjectsBoundedHexPreviewWithExplicit… [FAIL]
MetadataImageOverviewTests.ReadHeapValue_HonoursTheBlobPreviewBudget           [FAIL]
```

The asymmetry is real and the argument for keeping `HeapReference.Truncated` stored while deleting `HandleRef.DisplayTruncated` is correct: `Convert.ToHexString` is a complete spelling of the bytes that were read, so the flag cannot survive being derived.

I also checked the converse, since `HeapReference.Truncated` and `Text.IsTruncated` are now two copies of one fact on the String path — the shape the `HandleRef` remarks give as the reason to delete a flag. Setting `Truncated: false` there compiles too, but is caught by `StringBudget_ProjectsBoundedPreviewWithExplicitTruncation` and `ProjectRow_IgnoresTheWindowStart_ButKeepsTheCellBudgets`. So the duplication is covered; disregard.

**The compiler-gate claim.** `InertString` declares no conversion operators at all, and every public entry point — both constructors, `Format`, `Join` — routes through `VisualEncoder.Encode`. `TextPolicy` has exactly two members, `Field` and `Prose`, neither a passthrough. There is genuinely no way to seat a raw `string` in these fields.

**`int.MaxValue` as the budget** for the GUID and hex paths is safe and free: `Bound` calls `WindowWithin`, then returns `this` unchanged when `from == 0 && to == text.Length`, so the unbounded case allocates nothing and cannot report truncation.

**End-to-end.** My first sweep found zero hazard bytes *and* zero contained spellings, which proves nothing — the payload wasn't in the output. With a real positive control (`mdi <asm> -t TypeDef,MethodDef,Field` and `--heap '#Strings:0x2d'`, all three formats, against an image carrying `U+202E` in type, field and method names):

```
positive control: Hostile\u202EINJECTED, field\u202EINJECTED, method\u202EINJECTED
raw e2 80 ae: 0    raw e2 80 a8: 0    raw 1b: 0
```

`Malformed.Detail` has exactly one sink (`$"!malformed: {malformed.Detail}"`), which interpolates through `ToString()`, so the retype covers it.

## The coordination problem

This PR and #3684 are **siblings**, not a stack: both target `fix/3628-metadata-lens-inert-text`, both call themselves slice 2 of 2, and both rewrite `ContainCellText`. Merging one into the other conflicts in `MetadataTableProjector.cs` and `MetadataImageInspector.cs`.

The textual conflict is trivial. The design one is not, and it is the point I raised on #3684 arriving from the other side:

```csharp
internal static InertString ContainCellText(string value, int maxChars)
```

#3684 adds a third treatment in which the projected text is the artifact's **own** spelling, reached only through `--show-untrusted-text --dangerously-print-raw`. The return type now forbids expressing it — which is the PR working as designed, since "there is no conversion admitting a `string` into any of those fields" is exactly the property raw mode has to violate. There are only three ways out, and they should be chosen deliberately:

1. **A passthrough `TextPolicy`.** Cheapest, and it guts the guarantee: the compiler gate becomes vacuous for anyone who can name the member.
2. **Raw stops at the sink.** Correct in principle, but the projection is what reaches the sink, so it would have to carry the raw text — reintroducing the field this PR removes.
3. **Raw is not a projection mode.** `--dangerously-print-raw` re-reads from the image at the reported coordinate rather than routing through the projection. This preserves the invariant unconditionally and matches the flag's framing as an escape hatch rather than a rendering option. It is the most work.

Refusal itself is fine either way — the parameter is still `string value`, so `IsPermitted` can still see raw text and produce the `U+202E (Format) at index 7` coordinate. It is only the raw *rendering* that this typing closes off.

My suggestion: land this one first, since it is self-contained and #3684 is the one that has to bend. I will rebase #3684 onto it and bring the raw-mode question back with an implementation rather than three options.

## Nits

- `Display(InertString text, bool truncated) => truncated ? text + Ellipsis : …` binds `string operator +(object, string)`, boxing the struct. Only on truncated cells, so it is noise — `string.Concat(text.ToString(), Ellipsis)` if you are touching the line anyway.
- `InternalsVisibleTo DotnetInspector.MetadataRendering.Tests` is a new widening that the description does not mention. Worth a word on which internal the rendering tests now need, since it is the kind of grant that outlives its reason.

richlander:
One more, and it is the reason this looks fine at a glance:

```
$ gh pr checks 3687
no checks reported on the 'feat/3628-carry-inertstring' branch
```

**This PR has no CI at all.** `.github/workflows/ci.yml` filters `pull_request` on the **base** branch:

```yaml
pull_request:
  branches: [main, 'feature/**']
```

The base here is `fix/3628-metadata-lens-inert-text`, which matches neither. The comment directly above that line already predicts this exact failure — "a PR retargeted from `main` onto a long-lived feature branch silently stops matching, so it reports no checks rather than failing ones" — but the pattern it added covers `feature/**` only, and #3628's base branch is named `fix/…`.

I hit the same thing on #3684 and fixed it there by extending the list to `[main, 'feature/**', 'fix/**']`; checks appeared on the next push and went 9/9. That fix lives on `feature/mdi-untrusted-text-modes`, so it does not help this PR, and whichever of the two lands first should carry it.

Worth doing before merge rather than after: the local suites I ran cover Metadata and MetadataRendering, but nothing here has been through `decompiler-gates`, `pack`, `markdownlint`, or the full `test` job — and this PR changes public record shapes across four assemblies plus 98 lines of markdown. Once `ci-required` is enforced, "no checks reported" is exactly the state the comment warns is indistinguishable from "not run yet".

richlander:
## Restack verification — head `11c65aff1` → `53c99e880`

The stack was rebased and CI has now run. This slice reports checks for the first time, which retires the caveat in my previous comment.

### The restack changed the base and nothing else

```
$ git range-diff 3edf70099..11c65aff1  852a57e79..53c99e880
1:  9a5782225 = 1:  d71215a8e Carry metadata text as InertString so a sink cannot be handed raw text
2:  11c65aff1 = 2:  53c99e880 Fix stale doc comments referencing the deleted truncation flags
```

`=` on both rows: the patches are identical to the twice-reviewed ones. Nothing was rewritten, dropped, or folded.

The parent slice was *linearized* by the rebase — its `main`-merge commits are gone — so a range-diff alone would not settle it there. Proved by tree identity instead: replaying the old reviewed head against current `main` reproduces the rebased head exactly.

```
$ git merge origin/main          # onto old head 3edf70099
$ git diff HEAD 852a57e79
                                  # empty
```

### CI — both slices green

`ci-required` **pass** on #3679 and on #3687, along with `build-net10`, `test (ubuntu-26.04, linux-x64)`, `decompiler-gates`, and `markdownlint`. `csharp-diff-smoke`, `il-diff-smoke`, and `pack` report `skipping`, which is terminal and path-gated — I am not citing them as evidence of anything.

### What the rebase actually pulled in, and why it does not widen this hole

`main` added **1845 lines to `src/ILInspector.Metadata`** since the reviewed base — `AssemblyAcquisition`, `InspectionAcquisitionPlan`, `MetadataTypeDeclarationProbe`, `MetadataTypeDefinitionName`, `TypeDeclaration`, and others. That is the same layer this stack contains, and several of those read type names straight out of untrusted images, so a clean merge is not by itself reassuring.

None of them touch the projection model:

```
$ grep -c "MetadataValue\.\|HeapReference\|MetadataImageOverview\|HandleRef\|MetadataTableProject" <each new file>
0   (all six)
```

They sit on the assembly-inspection path, whose containment is the residual this PR already declares (#3463, #3606) — not a regression the rebase introduced, and not something the compiler gate was circumvented to avoid. Worth being explicit that this PR does **not** contain them.

### Re-validated at `53c99e880`

Because the merged combination is new even though the diff is not:

| Suite | Total | Failed |
| --- | ---: | ---: |
| `ILInspector.Metadata.Tests` | 1358 | 0 |
| `DotnetInspector.MetadataRendering.Tests` | 149 | 0 |
| `InertText.Tests` | 362 | 0 |

The metadata suite grew 1302 → 1358 on main's new tests; all pass. `-t:Rebuild`: **9 warnings, 0 errors**, baseline unchanged despite the 1845 new lines.

End-to-end against a freshly built hostile assembly (U+202E, U+2028, ESC in type, field, and method names), 12 outputs across `markdown`/`tsv`/`jsonl` for TypeDef, Field, MethodDef, and `#Strings`:

- **0 raw hazard bytes**
- Positive control, so zero does not mean absent: `Hostile\u202EINJECTED\u2028NEXT\^[ESC`, `field\u202EFlip`, `Method\u2028Break` (JSONL double-escapes correctly)

The central claim still holds at the new head:

```
--max-bytes 2           | 3 | yes | 2000…
--max-bytes 2147483647  | 3 | no  | 200001
```

and the empty-but-truncated ordering survives: `--max-chars 0` renders `AssemblyRef[1] (…)`, not a bare `AssemblyRef[1]`.

Still `Ready to merge`.

PR #3690, opened 2026-08-02
Document why a renderer test assembly sees Metadata internals

richlander:
Follow-up to #3687, raised from review feedback that the `InternalsVisibleTo` it added was an **undocumented widening**. That is a fair reading of the file:

```xml
<InternalsVisibleTo Include="dotnet-inspect.Tests" />
<InternalsVisibleTo Include="ILInspector.Metadata.Tests" />
<InternalsVisibleTo Include="DotnetInspector.MetadataRendering.Tests" />   <!-- why? -->
```

The first two are this project's own natural test assemblies. The third crosses a subsystem boundary, and it was the only entry with no justification — while the `ItemGroup` immediately below it does document its unusual cross-project `Compile Include`. The reason existed, but in the PR description, which is not where a reader of the csproj is standing.

## Why the widening is the right call

It grants exactly one thing: `MetadataTableProjector.ContainCellText`, the product's own containment factory, used by two renderer test files to build projection fixtures.

```
tests/DotnetInspector.MetadataRendering.Tests/MetadataProjectionRendererTests.cs:23
tests/DotnetInspector.MetadataRendering.Tests/MetadataImageOverviewRendererTests.cs:29
    => MetadataTableProjector.ContainCellText(value, maxChars);
```

This is the harness-boundary rule in `AGENTS.md`: a harness must exercise the product-owned capability rather than reconstructing it. The alternative is `new InertString(TextPolicy.Field, …)` spelled in fixture code — which pins the policy that was current when the fixture was written. Change `TextPolicy.Field` and every such fixture keeps asserting the old policy while the product moves. They stay green and mean less. The factory is the single place the policy is chosen, so fixtures should be made to go through it.

The other alternative is making `ContainCellText` public. That is a *larger* surface, not a smaller one — `InternalsVisibleTo` to the single assembly that needs it is the narrower option.

## Change

- Comment at the site in `src/ILInspector.Metadata/ILInspector.Metadata.csproj`, saying it crosses a boundary and why that is preferable to a public factory.
- A paragraph in `docs/design/inert-text.md`, in the "Carrying a contained value to the sink" section, since that section is written to read as the practice rather than as a changelog.

## Evidence

No behavior change — an XML comment and prose.

- `dotnet build dotnet-inspect.slnx -c Release` → **0 errors**
- `DotnetInspector.MetadataRendering.Tests` → **149 total, 0 failed** (the suite that would break if the grant stopped working)
- `npx markdownlint-cli docs/design/inert-text.md` → clean

## Review

**Trivial tier, no review round requested.** The diff is one XML comment and one documentation paragraph; there is no executable change, and the one thing that could plausibly break — the `InternalsVisibleTo` grant itself — is proven still working by the renderer suite compiling and passing.

### DIFF
--- a/docs/design/inert-text.md
+++ b/docs/design/inert-text.md
@@ -287,6 +287,17 @@ return new MetadataValue.HeapReference(
     HeapKind.String, HeapOffset(handle), raw.Length, text, text, text.IsTruncated);
 ```
 
+One consequence is worth stating, because it looks like an over-share until the
+alternative is written down. `ContainCellText` is `internal`, and the renderer's
+own test assembly reaches it through `InternalsVisibleTo` rather than
+constructing `new InertString(TextPolicy.Field, …)` in fixture code. A fixture
+spelled the second way pins the policy that was current when it was written, so
+changing `TextPolicy.Field` would leave every such fixture asserting the old
+policy while the product moved — the tests would stay green and mean less. The
+factory is the single place the policy is chosen, so tests should be made to go
+through it. Widening `InternalsVisibleTo` to the one assembly that needs it is a
+smaller surface than making the factory public.
+
 ### A partiality flag survives only where the text cannot know
 
 The tempting next step is to delete every truncation field and read
--- a/src/ILInspector.Metadata/ILInspector.Metadata.csproj
+++ b/src/ILInspector.Metadata/ILInspector.Metadata.csproj
@@ -16,6 +16,14 @@
   <ItemGroup>
     <InternalsVisibleTo Include="dotnet-inspect.Tests" />
     <InternalsVisibleTo Include="ILInspector.Metadata.Tests" />
+    <!-- Crosses a subsystem boundary, unlike the two above, so it is worth
+         saying why. The renderer tests build projection fixtures, and a fixture
+         must be contained by the same policy the projector applies rather than
+         by a hand-rolled TextPolicy.Field in test code — otherwise the policy
+         could change and the fixtures would keep asserting the old one. This
+         grants exactly that: MetadataTableProjector.ContainCellText, the
+         product's own containment factory. Prefer widening this to making that
+         factory public; the alternative surface is larger, not smaller. -->
     <InternalsVisibleTo Include="DotnetInspector.MetadataRendering.Tests" />
   </ItemGroup>
 


richlander:
Ready to merge

`ci-required` **pass**, with `build-net10`, `test (ubuntu-26.04, linux-x64)`, `decompiler-gates`, and `markdownlint` all green. `csharp-diff-smoke`, `il-diff-smoke`, and `pack` report `skipping` — path-gated and terminal, not evidence. `MERGEABLE` / `CLEAN`.

Trivial tier, so no review round: the diff is one XML comment and one documentation paragraph, with no executable change. The only thing that could plausibly regress is the `InternalsVisibleTo` grant itself, and that is proven intact by `DotnetInspector.MetadataRendering.Tests` compiling and passing 149/0 — that suite does not build if the grant stops working.

richlander:
Verified the factual claim rather than taking it on the diff's word: both `Cell(...)` helpers — `MetadataProjectionRendererTests.cs:23` and `MetadataImageOverviewRendererTests.cs:29` — are `static InertString Cell(string value, int maxChars) => MetadataTableProjector.ContainCellText(value, maxChars)`. So the tests do route through the product factory, and the guidance describes what the code already does. Approving.

One refinement worth folding in, because the rule as written is unconditional and there is a case it should not claim.

**The argument holds for fixtures, not for oracles.**

A *fixture* is an input — a value handed to the code under test. There the factory is plainly right: the fixture should be whatever the product would actually have produced, so a policy change flows into the test instead of leaving it asserting against a spelling the product no longer emits.

An *oracle* is the expected value on the other side of an assertion. There, routing through the same factory the product uses can be the weaker choice: both sides move together under a mutation to `ContainCellText`, and the assertion goes quiet. An independently spelled oracle stays sensitive to exactly the mutation the test exists to catch.

**Which one applies depends on whether the claim is absolute or relational**, and I got this wrong in my own code in the course of checking yours.

`MdiUntrustedTextModeTests.RawAndEncodedRenderingsShowTheSamePrefix` in #3684 asserts that the raw and contained renderings of a clipped cell describe the same value, differently spelled. That claim is *relational*: it says nothing about which policy is in force, so if the policy changed, both spellings should change together and the test should stay green. Hand-rolling `new InertString(TextPolicy.Field, ...)` as the oracle pinned a policy the claim does not depend on, which would have reddened it spuriously. I have switched that oracle to `ContainCellText` — your rule, for your reason.

So the sharper statement is: **use the factory unless the test's claim is that a specific spelling is produced.** For an absolute claim, spell it independently; for an input or a relational claim, use the factory.

Worth a sentence, because the case that motivates the exception is rare enough that the unconditional rule will be right nearly every time — and a reader who hits the exception with only the unconditional rule in hand will make the test weaker while believing they are following guidance.

---

Unrelated to the text, and the reason this was worth checking twice: chasing the above turned up a genuinely vacuous test of mine. That gate anchored the payload cell by searching the output for `"Hostile"`, and since the fixture assembly *is* the test assembly, the first match was the innocent `HostileAssemblyFixture` type name, not the spliced canary. Both renderings returned the same plain ASCII, so the comparison was between a string and itself. Dropping the decode from raw rendering left all three cases green. It now anchors on the canary name's untouched head (exposed from the fixture so the two cannot drift), and that same mutation reddens all three.

The general form is worth having in mind wherever a test locates something by a literal: **an anchor that merely looks distinctive is not the same as one that is**, and when it collides the test does not fail — it silently addresses the wrong thing and passes. Deriving the anchor from the fixture that created the target is what makes it a lookup rather than a guess.

PR #3702, opened 2026-08-02
Acquire ilasm/ildasm and mdv from one script instead of three CI copies

richlander:
## Conclusion

Optional external-oracle tools now have a single, scripted acquisition path. Restoring them unmasked **35 tests that had been silently skipping** — all pass.

## Problem

`ilasm`/`ildasm` acquisition was an inline bash block duplicated **verbatim in three workflows** (`ci.yml`, `deep-inspect.yml`, `release.yml`), with **no local equivalent**. A version bump meant editing three places, and a developer running the suites got a green result whose oracle-backed tests never ran.

`mdv` was worse: no automation anywhere, and the skip message named `Microsoft.Metadata.Visualizer`, **which does not exist on any feed under that id**. `mdv` ships prerelease-only from the dnceng `dotnet-tools` feed as package id `mdv`. Anyone following the message would have failed to install it, so the metadata-projection cross-implementation oracle had apparently never run.

## Change

`eng/restore-iltools.sh` restores the pinned `ilasm`/`ildasm` packages into `artifacts/iltools` (gitignored, so a re-run is a no-op) and prints the directories to put on PATH, one per line, diagnostics on stderr:

```bash
# CI
eng/restore-iltools.sh --rid linux-x64 >> "$GITHUB_PATH"
# local
export PATH="$(eng/restore-iltools.sh --mdv | tr '\n' ':')$PATH"
```

The three workflows now call it. `--mdv` additionally installs the `mdv` global tool, and the two wrong/vague skip messages now name the script.

The script **verifies** the binaries are executable after restore rather than trusting the package layout — a moved layout fails loudly instead of putting a non-existent directory on PATH and degrading every dependent test to a skip.

## Evidence

Release build, this branch's head, tools on PATH:

| Suite | Result | Skips before → after |
| --- | --- | --- |
| `dotnet-inspect.Tests` | 2843 passed, 0 failed | 14 → 4 |
| `ILInspector.Decompiler.Tests` | 4742 passed, 0 failed | 8 → 1 |
| `ILInspector.Metadata.Tests` | 1397 passed, 0 failed | 5 → 0 |

The 35 newly-running tests all pass — including the `mdv` row-count/MVID/AssemblyRef oracle comparisons and the 7 hand-authored-IL `ReturnToSender` regressions. Remaining skips are genuinely environmental (Windows-only P/Invoke, native PE image, NuGet-cache isolation), not tool acquisition.

Also: script is idempotent (second run restores nothing, emits only PATH lines); `markdownlint` clean on `AGENTS.md`; all three workflows parse as YAML; committed mode is `100755`.

Baseline for comparison: the same suites on `main` (`251bd3e0`) — all green, but with those 27 tool-gated skips.

## Non-action boundary

The workflow steps stay `continue-on-error: true`, so **CI still degrades to skips when acquisition fails** rather than failing the run. That pre-existing behavior is deliberately unchanged here to keep this PR's blast radius to acquisition mechanics; `AGENTS.md` now warns readers to check the step log before trusting a green decompiler or IL-diff leg. Narrowing it is a reasonable follow-up.

No product code changes — workflows, one script, one doc section, and two test skip messages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,6 +124,11 @@ jobs:
               eng/prepare-decompiler-opt-in-corpus.sh) CODE=true ;;
               eng/prepare-decompiler-pr-corpus.sh) CODE=true ;;
               eng/report-decompiler-opt-in-corpus-drift.sh) CODE=true ;;
+              # The test lane is what actually runs this script (the
+              # `Install ilasm/ildasm` step), so it is the only thing that can
+              # catch a break in it. Without this, a PR touching only the
+              # script skips that lane and `ci-required` passes on a `skipped`.
+              eng/restore-iltools.sh) CODE=true ;;
               # GateExpectedClassesTests lives in the fast lane and is the only
               # thing keeping this file honest against the pre-merge preset, so
               # editing the file must run the lane that validates it.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -388,10 +388,15 @@ jobs:
       # --mdv matters here and nowhere else: this is the only workflow that runs
       # tests/ILInspector.Metadata.Tests, whose cross-implementation oracle
       # skips without it -- and a skipped oracle is a green leg that compared
-      # nothing. continue-on-error keeps a feed outage degrading to those skips
-      # rather than failing the lane, so read the step log before treating this
-      # leg as oracle-backed.
+      # nothing.
+      #
+      # continue-on-error lets the rest of the lane still report, so a feed
+      # outage does not cost every other result in the job. It does not make
+      # the failure survivable: `Check ilasm/ildasm/mdv result` below fails the
+      # lane once the tests have run. Losing oracle coverage must be red, not a
+      # quietly shorter skip list.
       - name: Install ilasm/ildasm/mdv
+        id: iltools
         continue-on-error: true
         shell: bash
         run: eng/restore-iltools.sh --rid ${{ matrix.rid }} --mdv >> "$GITHUB_PATH"
@@ -450,6 +455,16 @@ jobs:
       - name: Run IL round-trip tests (fast)
         if: matrix.rid == 'linux-x64' && needs.changes.outputs.ilroundtrip == 'true'
         run: dotnet run --project tests/DotnetInspector.ILRoundtrip.Tests -c Release -- -trait- "Speed=Slow"
+
+      # Deliberately last: the suites above still report their results when
+      # acquisition fails, but the lane must not go green having silently traded
+      # its oracle comparisons for skips.
+      - name: Check ilasm/ildasm/mdv result
+        if: steps.iltools.outcome == 'failure'
+        run: |
+          echo "::error::Restoring ilasm/ildasm/mdv failed, so every test that uses them skipped." >&2
+          echo "This lane cannot certify oracle-backed results. See the 'Install ilasm/ildasm/mdv' step." >&2
+          exit 1
 
   # net10.0 fallback build coverage. The official non-AOT "any" package targets
   # the net10.0 LTS floor (Directory.Build.props), but the full test lane builds
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,11 +124,13 @@ jobs:
               eng/prepare-decompiler-opt-in-corpus.sh) CODE=true ;;
               eng/prepare-decompiler-pr-corpus.sh) CODE=true ;;
               eng/report-decompiler-opt-in-corpus-drift.sh) CODE=true ;;
-              # The test lane is what actually runs this script (the
-              # `Install ilasm/ildasm` step), so it is the only thing that can
-              # catch a break in it. Without this, a PR touching only the
-              # script skips that lane and `ci-required` passes on a `skipped`.
+              # The test lane is what actually runs these scripts (the
+              # `Install ilasm/ildasm` step and IlToolsActivationTests), so it
+              # is the only thing that can catch a break in them. Without this,
+              # a PR touching only a script skips that lane and `ci-required`
+              # passes on a `skipped`.
               eng/restore-iltools.sh) CODE=true ;;
+              eng/activate-iltools.sh) CODE=true ;;
               # GateExpectedClassesTests lives in the fast lane and is the only
               # thing keeping this file honest against the pre-merge preset, so
               # editing the file must run the lane that validates it.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -381,26 +381,7 @@ jobs:
       - name: Install ilasm/ildasm
         continue-on-error: true
         shell: bash
-        run: |
-          RID=${{ matrix.rid }}
-          VERSION=11.0.0-preview.1.26104.118
-          PACKAGES_DIR=$(mktemp -d)
-          PROJ_DIR=$(mktemp -d)
-          cat > "$PROJ_DIR/iltools.csproj" <<EOF
-          <Project Sdk="Microsoft.NET.Sdk">
-            <PropertyGroup><TargetFramework>net11.0</TargetFramework></PropertyGroup>
-            <ItemGroup>
-              <PackageDownload Include="runtime.${RID}.Microsoft.NETCore.ILAsm" Version="[${VERSION}]" />
-              <PackageDownload Include="runtime.${RID}.Microsoft.NETCore.ILDAsm" Version="[${VERSION}]" />
-            </ItemGroup>
-          </Project>
-          EOF
-          dotnet restore "$PROJ_DIR/iltools.csproj" --packages "$PACKAGES_DIR" --source https://dnceng.pkgs.visualstudio.com/public/_packaging/dotnet11/nuget/v3/index.json --source https://api.nuget.org/v3/index.json
-          ILASM_DIR="$PACKAGES_DIR/runtime.$RID.microsoft.netcore.ilasm/$VERSION/runtimes/$RID/native"
-          ILDASM_DIR="$PACKAGES_DIR/runtime.$RID.microsoft.netcore.ildasm/$VERSION/runtimes/$RID/native"
-          chmod +x "$ILASM_DIR"/* "$ILDASM_DIR"/* 2>/dev/null || true
-          echo "$ILASM_DIR" >> "$GITHUB_PATH"
-          echo "$ILDASM_DIR" >> "$GITHUB_PATH"
+        run: eng/restore-iltools.sh --rid ${{ matrix.rid }} >> "$GITHUB_PATH"
 
       - name: Run CLI tests (fast)
         run: dotnet run --project src/dotnet-inspect.Tests -c Release -- -trait- "Speed=Slow"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,10 +125,10 @@ jobs:
               eng/prepare-decompiler-pr-corpus.sh) CODE=true ;;
               eng/report-decompiler-opt-in-corpus-drift.sh) CODE=true ;;
               # The test lane is what actually runs these scripts (the
-              # `Install ilasm/ildasm` step and IlToolsActivationTests), so it
-              # is the only thing that can catch a break in them. Without this,
-              # a PR touching only a script skips that lane and `ci-required`
-              # passes on a `skipped`.
+              # `Install ilasm/ildasm/mdv` step and IlToolsActivationTests), so
+              # it is the only thing that can catch a break in them. Without
+              # this, a PR touching only a script skips that lane and
+              # `ci-required` passes on a `skipped`.
               eng/restore-iltools.sh) CODE=true ;;
               eng/activate-iltools.sh) CODE=true ;;
               # GateExpectedClassesTests lives in the fast lane and is the only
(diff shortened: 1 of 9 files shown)


richlander:

**Tier:** medium risk — CI acquisition mechanics touching three workflows (including `release.yml`), no product code. One reviewer, GPT (`gpt-5.6-sol`, xhigh), in an isolated review worktree at a fixed head each round.

Three rounds. Every finding was reproduced before acting on it, and each fix landed as a follow-up commit (no amends, no force-pushes).

### Round 1 — head `72010598` — 2 findings, both valid, both fixed in `b183eb1c`

| Finding | Verdict | Resolution |
| --- | --- | --- |
| `--rid win-x64` restored successfully then **exited 1** verifying its own output: the Windows package ships `ilasm.exe`, the script looked for `ilasm`. | **Valid.** Reproduced: exit 1 with the package present. | The payload extension follows the *RID*, so the RID now selects `tool_ext`. The `mdv` shim follows the *host*, so it is accepted under either name. |
| `DOTNET_TOOLS_PATH` is not honored by `dotnet tool install --global`, so setting it made the script fail *after* a successful install. | **Valid, and I had invented the variable.** | Now `"${DOTNET_CLI_HOME:-$HOME}/.dotnet/tools`. Confirmed by probe: `DOTNET_CLI_HOME=/tmp/clihome-probe` puts the shim at `/tmp/clihome-probe/.dotnet/tools/mdv`. |

### Round 2 — head `b183eb1c` — 1 finding, valid, fixed in `c10c285b`

| Finding | Verdict | Resolution |
| --- | --- | --- |
| Host RID auto-detection retained the `\r` from Windows CRLF `dotnet --info`, building `runtime.win-x64\r.Microsoft.NETCore.ILAsm` — rejected with NU1017. Hit the **documented default invocation** (no `--rid`), the one case where auto-detection is load-bearing. | **Valid.** Reviewer reproduced the NU1017 with the real byte sequence. | RID parse now strips CR/whitespace. Contrast under `cat -A`: old `[win-x64^M]`, new `[win-x64]`. Linux still resolves `linux-x64`. |

Both round-1 findings were re-confirmed fixed in round 2 by re-running the original reproductions, not by reading the diff.

### Round 3 — head `c10c285b` — clean

No significant issues. Confirmed the CRLF fix, probed the new parse line for defects the fix itself might have introduced (over-broad `tr` deletion, `[:space:]` portability), and re-ran the stdout-only contract across `--rid linux-x64`, `--rid win-x64`, `--mdv` cache-hit, `--help`, and the invalid-argument path.

### CI

Green on `72010598` and again on `c10c285b` (run [30766071288](https://github.com/richlander/dotnet-inspect/actions/runs/30766071288)) — `ci-required`, `test`, `decompiler-gates`, `pack`, `build-net10`, `csharp-diff-smoke`, `il-diff-smoke`, `markdownlint` all pass.

Because these steps are `continue-on-error`, green alone would not prove acquisition worked, so I checked the step log directly: `Install ilasm/ildasm` ran the script, reported `Restoring ilasm/ildasm 11.0.0-preview.1.26104.118 for linux-x64`, and the `test` job log contains **zero** "ilasm not available" / "ildasm/ilasm not found" skips.

### Local evidence at the final head

`tests/ILInspector.Metadata.Tests` (Release, tools on PATH via the script): **1397 passed, 0 failed, 0 skipped** — the `mdv` oracle, which the broken install instructions meant had never run here.

richlander:
Ready to merge

richlander:
## Adversarial review, rounds 4–6 (GPT 5.6 Sol, fresh unprimed pass)

Round 3 had closed clean, but rounds 1–3 were confirmation-shaped — each reviewer was primed with my prior findings. Round 4 was therefore run **unprimed against the complete diff**, at a head with `origin/main` freshly merged. It reset the clock.

### Round 4 — head `3efd2267` — 3 findings, all valid, fixed in `d32a643b`

| Finding | Why it mattered | Resolution |
| --- | --- | --- |
| The `changes` job enumerates `eng/*.sh` individually and omitted `eng/restore-iltools.sh`, so a PR touching **only the script** would skip the `test` lane — the one lane that executes it — and `ci-required` passes a `skipped` job. | This PR ran CI only incidentally, because it also edits `.github/workflows/*`. | Added the filter arm. Verified by applying the workflow's own `case` expressions. |
| `export PATH="$(script \| tr ...)"` reports **`export`'s** status, not the script's, so a failed restore returned success. | The documented command defeated the entire purpose of the change. | Documented form now assigns first. |
| On Git Bash, `git rev-parse --show-toplevel` returns `C:/...`, so the drive colon reads as a PATH separator and splits each directory into `C` and `/...`. | Windows developers get tools that are never found. | Emitted paths go through `cygpath -u` where present. |

**Self-found while fixing:** splitting assign from export introduced a bug the original one-liner did not have — command substitution strips the trailing newline, so the last directory got glued onto the first pre-existing PATH entry. It surfaced as `command -v mdv` finding nothing while `ilasm` resolved. Fixed with `printf '%s\n'`.

### Round 5 — head `d32a643b` — 1 finding, valid, fixed in `969112c5`

Splitting the assignment only helped **under `set -e`**. Interactively, a failed assignment does not stop the following `export`, which returns 0 — and empty output prepended an empty PATH entry, i.e. **the current directory**. Reproduced without `set -e` against a stub exiting 19: before, status 0 and first PATH entry `[]`; after, status 19 and PATH untouched. Fixed by `&&`-chaining with an `-n` guard, plus `emit_path` refusing to emit a blank line at the source.

### Round 6 — head `969112c5` — 1 finding, valid, fixed in `c7f754d0`

`[ -n "$emitted" ]` accepted a **whitespace-only** conversion, so a `cygpath` returning `" \t "` let the script exit 0 while emitting unusable entries — reaching the silent-degradation failure *through the guard meant to prevent it*. `emit_path` now requires a non-whitespace character. Reproduced: script exits 1, documented setup reports status 1, PATH untouched, `command -v ilasm` empty.

### Status

CI green on every head, including `c7f754d0` ([run 30772020749](https://github.com/richlander/dotnet-inspect/actions/runs/30772020749)). At `c7f754d0`, using the documented command verbatim: `ilasm`, `ildasm`, `mdv` all resolve, PATH has zero empty entries, `ILInspector.Metadata.Tests` reports 1397 passed / 0 failed / **0 skipped**.

**`c7f754d0` has not itself been reviewed.** That is round 7, which `AGENTS.md` gates behind explicit approval. Not marking ready until that is resolved.

richlander:
## Round 7 reconciliation — `48ba4ff5` … `4bba0ab2`

Rounds 4–6 each found a defect in the *same* artifact: the hand-written PATH incantation in the `AGENTS.md` snippet. That artifact was the only part of this change with no test and no CI coverage, which is why it kept producing findings — each fix could only be checked by reading it. Per the six-round cap, I stopped and got explicit approval for the architectural remedy plus a round 7 on the result.

**The remedy (`48ba4ff5`, `ed3f6a12`).** The assembly moved out of prose and into `eng/activate-iltools.sh`, a sourceable file, behind a real gate (`IlToolsActivationTests`, 25 tests, running in the `test` lane). `ed3f6a12` additionally added `--mdv` to `ci.yml` after I found CI had been restoring `ilasm`/`ildasm` but never `mdv` — the metadata oracle had been skipping on every CI run.

**Round 7** ran two reviewers from two families, unprimed, on isolated worktrees at `ed3f6a12`: **GPT 5.6 Sol** and **Gemini 3.1 Pro**. Eight distinct findings; all valid, all reproduced, all fixed in `4bba0ab2`.

| # | Finding | Found by | Resolution |
|---|---|---|---|
| 1 | A produced directory containing `:` split into several PATH elements; a colon-only line produced empty ones (= current directory) | GPT | Refuse — such a directory cannot be represented in PATH at all |
| 2 | Skipping an already-present directory left a stale copy earlier in PATH shadowing the restored one | Gemini | Remove and re-prepend; other elements preserved verbatim and in order |
| 3 | A caller defining a helper name had it destroyed, or (readonly) had its own function run in our place and got success back | GPT | Refuse on namespace collision |
| 4 | `${BASH_SOURCE[0]:-}` aborts dash with `Bad substitution` before the "needs bash" guard can print | both | `${BASH_VERSION:-}` |
| 5 | Inherited positional parameters; `--help` leaking fails with no output rather than a usage error | both | Empty args dropped (`source … ""` is the deterministic no-arg form); the trap is named on any argument-induced failure |
| 6 | `continue-on-error` let a failed acquisition report green with a quietly shorter skip list | Gemini | `Check ilasm/ildasm/mdv result` runs after the suites and fails the lane |
| 7 | The `AGENTS.md` gate only rejected `export PATH`/`tr '\n'`, so `PATH="$(eng/restore-iltools.sh):$PATH"` — the original defect restated — sailed through | GPT | Rejects any fenced snippet invoking the producer directly |
| 8 | `Activate_TreatsAnEmptyArgumentAsNoArguments` was vacuous — `"$*"` cannot distinguish no args from one empty arg | self-found via mutation testing | Stub reports `$#`; mutation now caught |

**Explicitly not actioned.** GPT suggested requiring a sentinel (`source … --activate`) on every invocation to make inherited `$@` unmistakable. Declined: it makes the common interactive case worse to close a failure that is already loud, leaves PATH unchanged, and now names itself. The `""` escape hatch covers the script case deterministically.

**Evidence.** Every reviewer repro command was re-run against the fix. The gate is mutation-tested: all ten defects this file has ever had now fail it individually — blank-line filter, exit-status masking, unconditional `":$PATH"`, dedupe, colon rejection, shadowing, namespace guard, empty-argument handling, POSIX guard, argument hint — and reverting each restores green. Suites: `IlToolsActivationTests` 25/0 skipped; `src/dotnet-inspect.Tests` 2868 passed / 4 skipped (unchanged environmental); `tests/ILInspector.Metadata.Tests` 1397 passed / **0** skipped, both with the wrapper sourced. `markdownlint AGENTS.md` clean; `ci.yml` parses and step ordering is asserted.

`4bba0ab2` is itself unreviewed. Not marking ready. The earlier "Ready to merge" comment is stale and superseded.

richlander:
## Round 8 reconciliation

Reviewed at `4bba0ab2` by **GPT 5.6 Sol** (xhigh) and **Gemini 3.1 Pro** (high), unprimed, on separate isolated worktrees. Four distinct findings; CDPATH and the newline defect were found independently by both. All four are fixed in `3726c092`, with a test-harness correction in `e0ab43e3`.

### 1. Tools were never actually run — GPT, blocking. Fixed.

`restore-iltools.sh` checked the executable bit and called that verification, but never invoked anything. A `dotnet tool` shim with no runtime to roll forward to is present, `+x`, and inert — so the oracle skipped and the lane went green. That is precisely the failure this PR exists to close, still open inside it.

Each tool is now invoked and must emit a marker from its own help text. Neither obvious signal works: exit status carries none (`ilasm` exits 1 on no arguments, `ildasm` and `mdv` exit 0), and the tool *name* is worse than useless, because the host's startup failure prints `App: /path/to/mdv`.

```
unrunnable mdv (DOTNET_ROLL_FORWARD=Disable DOTNET_ROOT=/nonexistent): restore_rc=1   # was 0
healthy environment:                                                  restore_rc=0
```

### 2. A newline in a caller's `PATH` entry introduced an empty element — both reviewers, blocking. Fixed.

`PATH` was split by translating colons to newlines and reading lines, so an entry containing a newline was torn in two and rejoined with a colon; a *trailing* newline left the second half empty. An empty `PATH` element means the current directory — the exact hazard this file promises never to introduce, and it reported success while doing it. Now walked colon by colon with parameter expansion only.

### 3. `CDPATH` broke self-location — both reviewers, non-blocking. Fixed.

With `CDPATH` set, `cd` prints the directory it resolved, and command substitution captured that ahead of `pwd`. Sourcing then failed on a machine where nothing was wrong. Fixed with `CDPATH= cd --`. Failed safe (`PATH` untouched), hence non-blocking.

### 4. MSYS paths into `$GITHUB_PATH` — Gemini, reported blocking. **Downgraded, and guarded.**

Not reproducible today: the `test` matrix is `linux-x64` only, and `deep-inspect.yml`/`release.yml` hardcode `--rid linux-x64`, so no lane reaches it. Gemini argued this one rather than running it, and said so.

It is a real trap for whoever adds a Windows lane, and the failure mode is silent. Rather than ship an unexercised native-path mode, the producer now refuses `cygpath`-present + `$GITHUB_PATH`-set loudly, naming what to implement.

### Self-found: my first CDPATH test was vacuous

It sourced by an absolute path, and `cd` consults `CDPATH` only for a relative target. Mutation testing caught it — reverting the fix alone left the suite green. Rewritten to mirror the documented invocation; reverting now fails that test and nothing else.

I also removed a newline-safety claim from `__iltools_path_has`. Its haystack is always `$joined`, built from the producer's line-oriented output, so it can contain neither a newline nor a colon and no test can reach it. The colon walk stays for consistency; the comment now says it is hardening, not a gated property.

### Mutation results

| Reverted fix | Suite |
| --- | --- |
| Tool smoke invocation | fails (1) |
| `PATH` rebuild colon walk | fails (2) |
| `CDPATH= cd --` | fails (1) |
| `$GITHUB_PATH` guard | fails (1) |
| `__iltools_path_has` colon walk | **green — unreachable, disclosed above** |

### Validation

29 activation tests (was 25). CLI `2041 total / 0 failed / 2 skipped`; metadata `1397 / 0 failed / 0 skipped`. Locally, activating the tools moves the CLI suite from 12 skips to 2, confirming the 10 oracle tests genuinely run. All 9 checks green on `e0ab43e3`, including `ci-required`.

richlander:
## Ready to merge

This supersedes the earlier readiness comment, which was stale from rounds 7 and 8.

Head `e0ab43e3` is `MERGEABLE` / `CLEAN`, with all 9 checks green including `ci-required`. `origin/main` is fully integrated (0 commits behind).

**Review.** Eight adversarial rounds. Rounds 7 and 8 were unprimed, two reviewers from two model families each (GPT 5.6 Sol, Gemini 3.1 Pro), on isolated fixed-head worktrees. All 12 findings across those two rounds are fixed, reproduced with the reviewers' own commands, and reconciled above.

**Closing judgement on the round-8 fixes.** They are accepted without a further round. Each is small, and each is pinned by a test that fails when the fix alone is reverted:

| Reverted fix | Suite |
| --- | --- |
| Tool smoke invocation | fails (1) |
| `PATH` rebuild colon walk | fails (2) |
| `CDPATH= cd --` | fails (1) |
| `$GITHUB_PATH` guard | fails (1) |

The one behaviour deliberately left ungated — the colon walk in `__iltools_path_has` — is disclosed as unreachable rather than claimed as verified.

**What this changes.** Oracle tools are acquired by one script instead of a bash block duplicated across three workflows; CI installs `mdv`, which it never did before, so the metadata oracle actually compares (`1397 / 0 skipped`, previously skipping on every run); a tool that installs but cannot run now fails the lane instead of quietly shortening the skip list; and the `PATH` assembly is executable, gated code rather than a doc snippet.

**Non-action boundary.** No Windows CI lane is added. The producer refuses to emit MSYS paths into `$GITHUB_PATH` rather than shipping an unexercised native-path mode, so whoever adds that lane gets a loud message instead of a silently uncompared run. `cygpath` handling remains verified only through stubs; no Windows host was available. The two remaining CLI skips are environmental (a Windows-only P/Invoke and a native PE image), not tool-related.

PR #3705, opened 2026-08-02
Scope the package content cache to the source that filled it

### LINKED ISSUE #3699: Package content cache is keyed by id and version only, so one feed's bytes are served for another feed's request
## Summary

The package content cache is keyed by package id and version with no source component, so bytes fetched from one feed are served to a later request that resolved against a different feed. Content is always source-specific: the same `id@version` can be published to two feeds with different contents.

## Evidence

`NuGetCache.TryGetCachedPackage(string packageName, string version)` takes no source, and both lookups it performs are name/version only:

```csharp
var nugetPackageDir = Path.Combine(nugetCachePath, normalizedName, normalizedVersion);
...
var appPackageDir = Path.Combine(appCachePath, normalizedName, normalizedVersion);
```

The commit marker that validates an entry carries the same two components and nothing else:

```csharp
private static string GetCommitMarkerContent(string packageName, string version)
    => $"{PackageContentCategory}:{packageName}@{version}";
```

Reproduced against a live private Azure DevOps feed carrying `Markout 0.32.99`, a version that does not exist on nuget.org:

1. Resolve `Markout@0.32.99` with the feed configured — succeeds, and populates `~/.cache/dotnet-inspect/package-content-v2/markout/0.32.99/`.
2. Resolve `Markout@0.32.99` again with a **nuget.org-only** `nuget.config` — prints the full package report, exit 0, despite nuget.org not having that version.
3. `rm -rf` that one cache directory and repeat step 2 — now correctly fails with `Error: Version '0.32.99' of package 'markout' not found`, exit 1.

## Why TTL and a config digest do not address this

TTL answers staleness, and category versioning (`package-content-v2`) answers schema change. Neither answers *which feed vouched for these bytes*.

A digest of `nuget.config` also does not help, because the collision occurs **within a single unchanged config** that lists both nuget.org and a private feed. That file's digest never changes across the runs above, so it would offer no protection.

## Proposed change

Record the resolved source identity in the commit marker alongside id and version, and treat an entry whose marker names a different source as a miss. `GetCommitMarkerContent` and `IsCommittedPackageValid` are the existing seam; `TryGetCachedPackage` and the commit path need the source threaded through.

## Tradeoff

For disjoint feeds — the common case, where a given `id@version` exists on exactly one configured source — there is no drawback at all.

The cost falls only on deliberately overlapping feeds, such as an Azure DevOps feed with upstream sources enabled that mirrors nuget.org. There the same `id@version` is legitimately available from two sources and would be downloaded and stored once per source. That is duplicated bytes, not a wrong answer, and it could be made opt-in for those setups.

## Scope note

Only the app-owned content cache is in scope. `TryGetCachedPackage` also consults the global NuGet packages folder (`~/.nuget/packages`), which is source-blind by NuGet's own design and is not ours to re-key; `_skipNuGetCache` already exists for isolated runs.

richlander:
Fixes #3699.

## The defect

The package content cache was keyed by package id and version only. Bytes downloaded from one feed were served to answer a request made under a `nuget.config` that does not list that feed.

Reproduced on `main` before any change, as a three-step A/B/A:

| Step | Config | Result on `main` |
| --- | --- | --- |
| 1 | private feed only | exit 0, populates cache |
| 2 | nuget.org only, warm cache | **exit 0, full report** — the bug |
| 3 | nuget.org only, entry deleted | exit 1 |

Making the source parameter required rather than defaulted turned every call site into a compile error, which surfaced a second instance beyond the issue text: `package <id> --version` against a warm cache printed a private-feed-only version as the answer and returned 0.

## The fix

The source is part of the cache **path**: `{cache}/{id}/{version}/{sourceKey}`. Category bumped to `package-content-v4`, which retires v2 and v3 entries.

Recording the source *inside* a shared entry was tried first and is wrong — it fixes reads and breaks writes. One slot cannot hold one coordinate from two feeds: the second must either overwrite bytes it is not entitled to serve, or fail to cache what it just downloaded. That version threw `InvalidDataException` on a package it had downloaded successfully. Both reviewers caught it; the history is in the commits.

A read happens before the tool knows which source will serve the package, so the question it asks is "is this slot's source one this caller is configured to read from", consulted in configured order so the feed a cold run would have used answers first.

Source identity is a digest of the URL canonicalized by `NuGetCredentialScope.CanonicalizeEndpoint`, extracted from the existing `IsSameEndpoint` so credential scoping and cache identity cannot disagree about what one feed is. Scheme, host, default port, percent-escape casing and a trailing path slash fold; path and query case do not, because `/FeedA` and `/feeda` can be different feeds. Local folder paths preserve case on every platform.

In-flight acquisition is scoped too: `PackageAcquisitionRequest` carries the ordered source scope, so callers with different configurations cannot join one download.

## Scope of the guarantee

Strict conformance applies to content **dotnet-inspect fetched**. The NuGet global folder is deliberately outside it: it holds what the user restored, not what this tool downloaded, and binding to it eagerly is what makes an inspection agree with the bytes a build actually consumed. That is now stated plainly in the docs, including the consequence that a private package sitting there will answer a nuget.org-only run, with `--no-nuget-cache` as the escape hatch.

An earlier draft of these docs justified this with a false claim — that NuGet records no source in the global folder. It does, in `.nupkg.metadata`. The claim was written rather than checked, and is corrected in `1bfd9cd4f`.

## Verification

Live A/B/A against a real private feed passes, plus an offline control (`DOTNET_INSPECT_OFFLINE=1`) proving the entitled config is still served entirely from cache — distinguishing scoping from disabling. The two-feed case produces two sibling slots and no crash. `--no-nuget-cache` verified by A/B: offline, the same coordinate is served from the global folder without it and not found with it.

Mutation-gated, each caught by a different test: source-blind reads; a shared cache path; whole-URL lowercasing; an unscoped flight key; trimming the recombined URL; `Host` for `IdnHost`; lowercased local paths; reversed key list; reversed probe loop.

Suites: `dotnet-inspect.Tests` 2866 with 1 known local-only failure (#3592, confirmed across four runs), `DotnetInspector.Services.Tests` 360/0, `NuGetFetch.Tests` 295/0, `EvilPoolSweepGateTests` 19/0. markdownlint clean.

## Deliberately out of scope

- `<packageSourceMapping>` is not consulted — #3722.
- Precedence between two entitled sources when the higher-precedence one has no cached slot; closing it needs a network probe per request. Documented under "Known deviation" and tracked with the broader source-support design in #3723.

### DIFF
--- a/docs/design/cache-concurrency.md
+++ b/docs/design/cache-concurrency.md
@@ -47,15 +47,19 @@ of the data.
 
 ### Known deviation: precedence between two entitled sources
 
-Slots are consulted in configured order, so when two configured feeds both hold
-a coordinate the higher-precedence one answers, as it would on a cold run. The
-case that still differs is when the higher-precedence feed holds the package but
-has no cached slot: a cold run downloads from it, a warm run answers from the
-lower-precedence feed's slot. Both are entitlement-correct — the caller reads
-both feeds — so this is precedence, not source confusion. Closing it requires
-probing the network on every request, which would defeat cache-first and offline
-operation, so it is left open deliberately and tracked as part of the broader
-source-support design.
+Slots are consulted in configured order, so when two configured feeds both have
+the coordinate cached, the higher-precedence one answers, as it would on a cold
+run. The case that still differs is when the higher-precedence feed holds the
+package but has no cached slot: a cold run downloads from it, a warm run answers
+from the lower-precedence feed's slot. Both are entitlement-correct — the caller
+reads both feeds — so this is precedence, not source confusion. Closing it
+requires probing the network on every request, which would defeat cache-first and
+offline operation, so it is left open deliberately and tracked as part of the
+broader source-support design.
+
+"Entitled" here means the source appears in the caller's configured sources.
+dotnet-inspect does not yet consult `<packageSourceMapping>`, which would narrow
+entitlement further, to the feeds mapped to a particular package id.
 
 A source is identified by a digest of its canonical URL, and canonicalization is
 shared with the credential scope's `IsSameEndpoint` rather than reimplemented, so
--- a/docs/design/cache-concurrency.md
+++ b/docs/design/cache-concurrency.md
@@ -45,6 +45,18 @@ folder from lookup. Note that NuGet does record an installed package's source in
 `.nupkg.metadata`; the folder is treated as source-blind by choice, not for want
 of the data.
 
+### Known deviation: precedence between two entitled sources
+
+Slots are consulted in configured order, so when two configured feeds both hold
+a coordinate the higher-precedence one answers, as it would on a cold run. The
+case that still differs is when the higher-precedence feed holds the package but
+has no cached slot: a cold run downloads from it, a warm run answers from the
+lower-precedence feed's slot. Both are entitlement-correct — the caller reads
+both feeds — so this is precedence, not source confusion. Closing it requires
+probing the network on every request, which would defeat cache-first and offline
+operation, so it is left open deliberately and tracked as part of the broader
+source-support design.
+
 A source is identified by a digest of its canonical URL, and canonicalization is
 shared with the credential scope's `IsSameEndpoint` rather than reimplemented, so
 one URL cannot mean two things in one tool. Scheme, host, default port,
--- a/docs/design/cache-concurrency.md
+++ b/docs/design/cache-concurrency.md
@@ -7,10 +7,10 @@ owns how content for an exact coordinate becomes visible safely.
 
 ## Source conformance
 
-Cached content is scoped to the source that supplied it. A request is served
-only the exact bytes downloaded from a source the current configuration lists.
-Content obtained from any other source is invisible to it, even for an
-identical package id and version.
+Content that dotnet-inspect downloaded is scoped to the source that supplied it.
+A request is served only the exact bytes downloaded from a source the current
+configuration lists. Content obtained from any other source is invisible to it,
+even for an identical package id and version.
 
 This is a correctness boundary, not an optimization. Two feeds may publish the
 same coordinate with different content, so serving one feed's bytes for another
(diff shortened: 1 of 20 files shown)


richlander:
Converting to draft. Post-hoc review found two blocking defects, both reproduced in running code.

**1. A second source now hard-fails instead of working (regression).**

Content still lives at one slot, `{id}/{version}`, with the source recorded in the marker. So when feed B requests a package feed A already cached, B correctly misses, downloads, and then `CommitPackage` finds the directory already present and throws:

```
InvalidDataException: Package cache entry '.../package-content-v3/pkg/1.0.0' is incomplete or corrupt. Clear the cache before retrying.
```

This happens even when both feeds serve identical bytes. Before this change that user got a (wrong) cache hit; now they get a crash. That is worse than the bug being fixed for anyone whose package is legitimately available from two configured feeds.

The marker design is the cause: one slot cannot hold the same coordinate from two sources. The fix is a source-scoped **path** rather than a marker on a shared path.

**2. `GetSourceKey` lowercases the whole URL, aliasing distinct feeds.**

`https://h/FeedA/v3/index.json` and `https://h/feeda/v3/index.json` produce the same key. This repo already documents why that is wrong, in the same file I edited — `NuGetSourceCompat.cs`:

> Comparing whole URLs case-insensitively would alias `/FeedA` and `/feeda`, which are different feeds on servers with case-sensitive paths, and would hand one feed's credentials to the other.

Only scheme and host are case-insensitive; path and query must be compared ordinally on their escaped form.

Also to address before this is ready:

- In-flight acquisition (`PackageAcquisitionRequest` / `AsyncCache`) is keyed by cache path only, so a B-only caller can join an A-only download and receive A's bytes.
- The security claim should be narrowed. An unsalted digest of a low-entropy feed URL is opacity, not confidentiality, and `NuGetFetch/SourceResolver.cs` already rejects userinfo in source URLs — so the stated rationale does not match the real flow.

Reworking on the branch.

richlander:
## Review record

Three rounds, two reviewers (`code-review` on claude-opus-4.8, `rubber-duck` on gpt-5.6-sol). Both are clean as of `6c6dacc4b`.

**Round one — two blocking defects, both mine, both reproduced before acting.** The first design put the source in the commit marker rather than the path. That fixes reads and breaks writes: a second feed correctly missed, downloaded, then threw `InvalidDataException` committing a package it had just downloaded successfully. Both reviewers found it independently. The same round caught that I had lowercased whole URLs, aliasing `/FeedA` and `/feeda` — contradicting reasoning already documented in the file I was editing.

**Round two — reviewer clean, duck blocking on four.** Confirmed in running code: query-vs-path trailing-slash aliasing (`?token=abc/` collided with `?token=abc` *while* `/feed/?q=` failed to fold — one cause, trimming the recombined URL); undefined slot precedence from a `HashSet`; the global folder; `packageSourceMapping`.

**Round three — both clean.** The reviewer replicated the old `IsSameEndpoint` and diffed it against the extracted `CanonicalizeEndpoint` across a 529-pair matrix: zero behavioural differences, and the credential-*sending* path was never touched.

## A correction worth recording

An earlier revision of these docs justified the global-folder gap by claiming NuGet records no source there. It does — `.nupkg.metadata` has carried a `source` field since format version 2, and 953 of 960 entries in a working global folder have one. That claim was written rather than checked. Corrected in `1bfd9cd4f`, which states the real reason instead.

## Deferred deliberately

A fix making configured source order structural in `BuildConfiguredSources` was written and then **reverted**, because landing new logic after two clean reviews would have invalidated them. Filed as #3724 with the analysis. Current risk is latent, not live: `MergeConfigFile` only calls `Clear()` and the indexer, never `Remove`, and warm and cold paths consume the same list, so cache warmth cannot change which feed answers regardless of that order.

Also filed: #3722 (`packageSourceMapping`), #3723 (source-support design), #3726 (a pre-existing intermittent test failure, confirmed on unmodified `main` and unrelated to this change).

## Verification

Nine mutations gated, each caught by a different test, covering reads, writes, canonicalization and ordering. `dotnet-inspect.Tests` 2866 with one known local-only failure (#3592, reproduced on `main` in all three control runs); `DotnetInspector.Services.Tests` 360/0; `NuGetFetch.Tests` 295/0. Live A/B/A against a real private feed, plus an offline control proving the entitled config is still served from cache rather than the cache simply being disabled.

Issue #3645, opened 2026-07-31
NormalizeSynthesizedMemberOrdinals folds two overloads' local functions into one operand, masking a real difference


richlander:
`NormalizeSynthesizedMemberOrdinals` (landed in #3505 for #3503) masks a real difference between the local functions of two overloads. Its own remarks name this case as a known limitation, then conclude with a claim that is false for it:

> Known limitation, deliberately not addressed: overloads share a containing-method name and are told apart only by `N`, so this option conflates the closures of two overloads with the same lambda index. [...] **Each of these costs a false positive, never a masked difference.**

The first item costs a **masked difference**, not a false positive. Conflating two overloads' closures relates two members that genuinely differ, which is a false `Exact` — the one direction this normalization was designed never to take.

### Measured, on real Roslyn output

```csharp
public static class C
{
    public static int M(int x)    { return a(); static int a() => 1; }
    public static int M(string x) { return a(); static int a() => 2; }
}
```

Roslyn emits `<M>g__a|0_0` and `<M>g__a|1_0` — same declaring type, same signature `int32()`, genuinely different code (one returns 1, the other 2). Comparing the two overload bodies with `IlBodyDiff.Compare(reader, b0, reader, b1, normalization)`:

| normalization | outcome |
| --- | --- |
| `None` | `OperandDiff` — `call int32 [Ovl]C::<M>g__a\|0_0()` vs `call int32 [Ovl]C::<M>g__a\|1_0()` |
| `NormalizeSynthesizedMemberOrdinals` | **`Exact`** |

Both names normalize to `<M>g__a|#_0`, and nothing else in the operand distinguishes them.

### Why this matters beyond the doc comment

`FidelityCheck.cs:52` puts this flag in the normalization set the compile-back fidelity oracle runs with. So the oracle can report `Exact` for a decompilation that binds a call to the *wrong overload's* local function — which is exactly the kind of misattribution the oracle exists to detect. The failure is invisible: it produces a success-shaped result, not a diagnostic.

No gate covers it. `IlBodyDiffNormalizationTests` has no test mentioning overloads; the sole match for "overload" in that file is unrelated prose. The limitation was documented and then not pinned, so nothing fails today and nothing will fail if it widens.

### Root cause is the design, not the parser

[...]

PR #3580, opened 2026-07-30
Compare Roslyn generated-member ordinals through a correspondence (fidelity contract v2)

richlander:
Folds Roslyn's compiler-generated member ordinals in the compile-back fidelity comparison, retiring **23 docketed methods with 0 added**. Adopted as fidelity **contract v2**.

Closes #3491.

## The problem

Roslyn names a local function `<M>g__L|N_K` and an iterator or async state machine `<M>d__N`, where `N` is the ordinal of a member of the **containing type** — counting fields, properties, and nested types, not just methods.

Compile-back fidelity recompiles a decompiled method inside a *reconstructed* type skeleton. That skeleton necessarily has a different member population, so `N` differs. Two bodies that are otherwise byte-identical reported an operand difference for that reason alone.

The drift is **permanent by construction** and not recoverable arithmetically — measured deltas across the corpus were non-uniform and signed (−28, +7, +30).

## The fix

`CompilerGeneratedOrdinalCorrespondence` builds a **two-sided** correspondence from both metadata readers together:

- The key is the canonicalized declaring-type path plus the ordinal-free name (`<M>g__L|#_0`), with enclosing generated type names elided too, so a state machine nested in a renumbered type keys the same on both sides.
- A member folds **only if** that key resolves to exactly one eligible member on **both** sides.
- Eligibility requires `CompilerGeneratedAttribute`, not name shape.
- The method **signature** is deliberately excluded from the key, so same-key overloads simply do not fold.

Two-sidedness is the load-bearing property, and it is not expressible as a per-side rewrite. A per-side eligibility test is resolver-local: a member unique on one side but ambiguous on the other would fold on one side only, turning an exact comparison into an operand difference — *manufacturing* a difference neither assembly contains. Requiring uniqueness on both sides makes folding symmetric, so the worst case is that a pair does not fold and the original mangled names are compared, which is exactly the un-normalized behaviour.

## Measured effect

Base-vs-head set difference on the sugared fidelity gate, comparing the exact set of differing methods:

| | Methods differing |
| --- | --- |
| base `35014d91c` | 73 |
| head | 50 |

**23 retired, 0 added.** The `0 added` is the safety result — a bad correspondence manufactures differences. Retirements are 7 `g__` local functions and 16 `d__` iterators.

All 23 were then **measured** to be genuinely `Exact` (not merely below a threshold) and added to `PinnedExact`, because the gates assert only `actual ⊆ KnownDiffs` — removing a docket entry on its own proves nothing.

[...]

Issue #3661, opened 2026-08-01
IlBodyDiff renders operands non-injectively: fidelity oracle reports Exact for genuinely different bodies


richlander:
`IlBodyDiff` renders IL operands non-injectively, so the fidelity oracle reports `Exact` for bodies that genuinely differ.

## Summary

`IlBodyDiff.Compare` decides equality by rendering each operand to text and comparing the
text. Several IL distinctions never reach that text, so two bodies that differ in real IL
semantics compare `Exact`. This is the oracle that certifies decompiler correctness, so
each of these is a permanently invisible class of defect.

Found during round-15 adversarial review of #3580 (credit: GPT reviewer). **These are not
caused by #3580.** Every case below was measured with **identical member names on both
sides** and **no ordinal normalization of any kind** — only `NormalizeCurrentAssemblyScope`
— so they reproduce on `main` today, independent of both #3580's correspondence and
#3505's per-side rewrite.

## Measured

Two images differing only in the callee's signature blob; the caller's body is
`ldnull; call; ret` in both. Result is `IsExact=True` in every row.

| Class | Old signature | New signature |
| --- | --- | --- |
| `class` vs `valuetype` | `00 01 01 12 05` | `00 01 01 11 05` |
| generic flag, arity 0 | `10 00 00 01` | `00 00 01` |
| SZARRAY vs MD-ARRAY rank 1 | `00 01 01 1D 12 05` | `00 01 01 14 12 05 01 00 00` |
| MD-array size 3 vs 4 | `… 14 12 05 01 01 03 00` | `… 14 12 05 01 01 04 00` |
| vararg sentinel vs none | `00 01 01 41 08` | `00 01 01 08` |

```text
class vs valuetype param     noNormalization IsExact=True
generic arity 0 flag         noNormalization IsExact=True
szarray vs md-array r1       noNormalization IsExact=True
md-array size 3 vs 4         noNormalization IsExact=True
sentinel vs none             noNormalization IsExact=True
```

The GPT reviewer additionally reported, and I have not independently re-measured:

- **`ModuleReference` scope** — two identical `N.T` references scoped to different
  netmodules render alike.
- **Operand text is not injectively framed.** Rendered identity segments are joined with
  `::` and `.`, which are legal characters *inside* a metadata name, so the join is
  ambiguous. Type `C` + member `A::B` and type `C::A` + member `B` both render `C::A::B`.
[...]

PR #3707, opened 2026-08-02
Own Roslyn lambda methods in the compiler-generated correspondence (#3645 slice 1/3)

richlander:
Slice 1 of 3 for #3645. Moves the Roslyn lambda form `<M>b__N_K` from the per-side
rewrite onto the two-sided correspondence, which fixes a masked difference on real
compiler output at zero measured corpus cost.

## The defect, measured on Roslyn output

The per-side rewrite (`NormalizeSynthesizedMemberOrdinals`) rewrites each side's names
without seeing the other side, so it cannot tell a renumbered lambda from a genuinely
different one. Compiling

```csharp
public static Func<int> M(int x)    => () => 1;
public static Func<int> M(string x) => () => 2;
```

puts `<M>b__0_0` and `<M>b__1_0` on `<>c`. Both elide to `<M>b__#_0`, and the rewrite
folds them. Under the composed contract `FidelityCheck` ships, measured at
`0b2834ba4` versus this branch:

| shipped composed contract | main | this branch |
| --- | --- | --- |
| lambda overloads `<M>b__0_0` vs `<M>b__1_0` | **`Exact` — masked** | `OperandDiff` |
| local-function overloads `<L>g__a\|2_0` vs `\|3_0` | `OperandDiff` | `OperandDiff` (unchanged) |

The correspondence folds only where the ordinal-free key is one-to-one on **both**
sides, so it refuses this pair as ambiguous. `IlBodyDiff` already hands owned names to
the correspondence alone, so moving the form does not merely relocate the fold.

## Owning the form requires inheriting the declaring type's mark

Measured on a Release build, this is not symmetric between the two method forms:

```
TYPE C: []
  METHOD <L>g__a|2_0      [CompilerGeneratedAttribute]
TYPE <>c: [CompilerGeneratedAttribute]
  METHOD <M>b__0_0        []
  FIELD  <>9__0_0         []
```

Roslyn marks a generated *type* and leaves its members unmarked; a local function sits
on the user's own unmarked type and carries the mark itself. Asking only the member
declines every lambda; asking only the type declines every local function. Eligibility
now accepts either, computed on demand so the malformed-metadata exposure does not
widen. Types still answer for themselves — every type call site passes `false`, and the
parameter has no default so the field index the next slice adds must state its answer.

## Evidence

Tamper table, every entry measured, `.bak` restores (never `git checkout --`):

| tamper | tests that fail |
| --- | --- |
| remove the `b__` branch | `LambdaShape_Folds`, `UnmarkedMemberOfAGeneratedType_Folds`, the nesting composition test |
| drop the inherited mark | **exactly** `UnmarkedMemberOfAGeneratedType_Folds` |
| never consult the lazy mark | **exactly** `UnmarkedMemberOfAGeneratedType_Folds` |
| treat every type as generated | `UnmarkedMemberOfAnUnmarkedType_DoesNotFold` + 6 pre-existing |
| drop the attribute requirement | `UnmarkedMemberOfAnUnmarkedType_DoesNotFold` + 7 pre-existing |
| drop the canonical ordinal check | 3 of 5 tail rows + 6 pre-existing |
| require no trailing `_K` | **exactly** the no-separator tail row |
| elide the slot ordinal too | `LambdaSlotOrdinals_StayDistinct` + 3 pre-existing |
| drop the containing name from the key | **exactly** `DistinctLambdaContainingNames_StayDistinct` |
| remove both ambiguity checks | `RealLambdaOverloadShape_IsRefusedAsAmbiguous` + 4 pre-existing |
| remove the on-demand short circuit | nothing — confirming it is a pure optimization |

## Two claims in existing comments were wrong, and are corrected

- A control asserted `<<Run>b__103_0>b__104_1` is "a shape Roslyn does emit" and pinned
  that it must fold. Compiled it: a lambda inside a lambda is named after the
  **outermost** method — `<Run>b__0_0`, `<Run>b__0_1`, flat. The nested form is
  hand-written metadata only, so declining it costs no Roslyn-emitted row.
- The class remarks said local-function, state-machine and display-class members "all
  carry the attribute". False for the form this slice adds, as the dump above shows.

One new test row was **weaker than it looked**: `<M>b___0` vs `<M>b___1` passed even
with the empty-scope check removed, because the slot difference carried it. Re-paired
against a well-formed scope; it is now load-bearing, and its remaining double-guarding
is recorded in its own remarks rather than left for a reader to find.

## Non-action boundary

- `<>9__` cache fields stay on the per-side rewrite. Owning them needs a field index the
  correspondence does not have, plus sibling-pairing to the lambda method — measured, a
  naive `<>9__#_M` key folds 1 of 5 cache fields in a five-lambda `<>c` where pairing
  folds 5 of 5. That is slice 2.
- Deleting `NormalizeSynthesizedMemberOrdinals` and bumping the fidelity contract to v3
  is slice 3, so the stack pays one contract bump.
- A `b__` name reached through a `MemberReference` is now declined by both options,
  extending #3580's documented accepted cost. It is a false positive, never a masked
  difference, and it costs zero corpus rows.

## Validation

Fidelity **68/0** — unchanged, so the 23 retirements still measure `Exact` and this
change costs nothing measurable. Decompiler fast 4496/0, Instructions 293/0.
Full suite results in a follow-up comment.


richlander:
## Validation at the reviewed head `47203fc2a`

`origin/main` (`0334a2de5`) is merged in. Its four commits touch NuGet feed resolution,
the type listing's `-S` handling, and docs — no overlap with any file this PR changes.

| gate | result |
| --- | --- |
| build `dotnet-inspect.slnx -c Release` | 0 errors |
| `ILInspector.Instructions.Tests` | **293 / 0** |
| `ILInspector.Analysis.Tests` | 584 / 0 |
| `DotnetInspector.Services.Tests` | 360 / 0 |
| `ILInspector.Metadata.Tests` | 1397 / 0 (5 skipped) |
| `DotnetInspector.MetadataRendering.Tests` | 187 / 0 |
| `dotnet-inspect.Tests` | 2843 / **1** — `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, pre-existing on `main` |
| decompiler `--gate fast` | **4496 / 0** |
| decompiler `--gate fidelity` | **68 / 0** |

CI is green, `ci-required` included.

Fidelity 68/0 is the load-bearing number: it is unchanged from `main`, so the 23
retirements #3580 bought still measure `Exact` and owning the lambda form costs zero
corpus rows. The `MemberReference` false positive described in the PR body is therefore
real but currently unrealised by any fixture in the corpus.

### One observation I cannot explain, recorded rather than dismissed

An earlier run of the CLI suite at `d037514c0` reported **2** failures. That run's harness
captured only the count, not the names. Three subsequent runs at that head and one at the
merged head all report exactly **1** (`Layout_Count`), so I cannot name the second.

I am not claiming it away. The argument that it is unrelated to this change is that the
change is deterministic and confined to `CompilerGeneratedOrdinals.cs`, and a
deterministic change cannot produce a failure that appears in one run of five; the suites
that actually exercise the changed code — Instructions, Analysis, decompiler fast and
fidelity — are 0-failure in every run. The CLI suite also contains network-touching areas
(`SymbolPackageDownloader`, `PackageAcquisitionConcurrency`, `NuGetSearchSources`), and
that run overlapped my own `git push` and `gh pr create`. That is a plausible explanation,
not a measurement, and I am flagging it as such.

Adversarial review is in flight at this head. Not ready to merge.

richlander:
## Review reconciliation — round 1, two models, head `47203fc2a`

Reviewed by **GPT (`gpt-5.6-sol`, xhigh)** and **Gemini (`gemini-3.1-pro-preview`, high)**,
each in its own isolated detached worktree pinned at `47203fc2a`. Three findings between
them, two of which are the same finding. I reproduced all three before acting.

### Headline: both models independently produced a false `Exact`, and both misattributed it

GPT built two lambda methods whose returns differ as `SZARRAY int32` vs rank-1
`ARRAY int32` and measured `composed normalization: Exact`. Gemini built two types with
the identical name `<M>d__3` but `GenericParam` counts 1 and 2 and measured `IsExact=True`.
Both are real. Neither is caused by this PR, and neither is caused by name folding at all.

The test that separates the two hypotheses is to re-run the same pair with normalization
switched off. I did that:

```text
PROBE3 identical author-written name, SZARRAY vs rank-1 ARRAY, NO normalization: outcome=Exact
PROBE4 old[name='<M>d__3' genericParams=1] new[name='<M>d__3' genericParams=2]
         => noNormalization=Exact  ordinals=Exact
```

GPT's case reproduces with an **author-written, non-generated, identical** method name and
**no normalization whatsoever** — no generated name, nothing to fold, no mechanism
running. Gemini's reproduces with normalization disabled too. `IlBodyDiff.cs:1606-1607`
is the actual cause: `GetSZArrayType` returns `int32[]`, and `GetArrayType` at rank 1
returns `int32[` + `new string(',', 0)` + `]`, which is also `int32[]`.

Gemini's causal story was explicitly that the correspondence's *refusal* to fold forces a
fallback to raw names which then compare equal. That story is wrong: the comparison is
`Exact` when the correspondence is not running.

This is #3661, already filed, whose table already lists `SZARRAY vs MD-ARRAY rank 1`.
I have [added the evidence and a new row](https://github.com/richlander/dotnet-inspect/issues/3661#issuecomment-5160892351)
for Gemini's type-arity case, which #3661 did **not** previously cover.

**Did this PR widen the exposure?** No — measured, not argued. The same array-kind pair,
at the base commit `0b2834ba4` and at this head:

| | base `0b2834ba4` | head `47203fc2a` |
| --- | --- | --- |
| `<M>b__3_0` vs `<M>b__7_0` raw | `OperandDiff` | `OperandDiff` |
| rewrite only | `Exact` | `Exact` |
| correspondence only | `OperandDiff` | `Exact` |
| **composed (what ships)** | **`Exact`** | **`Exact`** |

The composed outcome is identical. On base the per-side rewrite already folded the pair —
and it does so regardless of attributes, which I checked separately, so GPT's
inherited-mark fixture composed to `Exact` on base as well:

```text
PROBE2 unmarked members on a [CompilerGenerated] type:
  raw=OperandDiff rewriteOnly=Exact corrOnly=Exact composed=Exact
```

Note the `g__` row in the same probe: the correspondence alone reported `Exact` for the
array-kind pair **at base**, before this PR existed. The behaviour is inherited from
#3580, not introduced here.

Not blocking. No change to this PR.

### Accepted and fixed: the on-demand declaring-type mark asserted a property with no gate

Both reviewers found this, by the same tamper — replace the lazy `bool? typeIsGenerated`
with an eager unconditional read and the suite stays green at 293/0. They are right, and
this one is mine: my own tamper table already recorded "remove the short circuit → fails
nothing" and I read that as evidence the short circuit was a pure optimization, when it
equally showed that the *safety* sentence next to it was ungated. AGENTS.md is explicit
that a comment asserting a safety property must name its gate or be marked unverified.

Fixed in `bb59fcc30` by marking it `UNVERIFIED` in the source and saying exactly why the
gate does not exist: the gate needs an image whose only malformed row is a custom
attribute on a type carrying no ordinal-bearing member name, and `BuildImage` shares one
attribute constructor across every attributed entity, so corrupting it also breaks the
members whose fold the test must observe. The existing `CorruptUnrelatedTypeName` helper
cannot substitute — it corrupts a type *name*, which every type path reads
unconditionally, so it fails closed identically under both shapes. #3708 tracks building
the fixture properly, and requires asserting both directions so it cannot pass vacuously.

I chose the marker over inventing a fixture inside this slice: the fixture needs
test-infra changes that deserve their own review, and a fixture that looks like it gates
this but does not would be worse than an honest marker.

### Confirmed clean by both reviewers

- `RealLambdaOverloadShape_IsRefusedAsAmbiguous`, the PR's central canary, is non-vacuous
  — GPT removed both ambiguity checks and observed it fail `Expected: False; Actual: True`.
- The `MemberReference` asymmetry is a false positive, as the PR body claims, not a masked
  difference — verified independently by both.
- `ElideScopeOrdinal` factoring changed nothing about what the `g__` path accepts or
  rejects (Gemini, attack point 2).
- The eligibility widening does not let an author-written member fold into a false `Exact`
  via a generated type's mark (Gemini, attack point 1).

### Status

Validation at `47203fc2a` stands; `bb59fcc30` is comment-only plus this reconciliation,
re-validated at Instructions 293/0 and a clean full-solution Release build. Awaiting CI on
the new head. Round 1 produced one accepted fix and no blocking finding against the
change itself.

richlander:
Ready to merge

Round 1 is closed. Both reviewers ran a confirmation pass at the final head `bb59fcc30`, independently verified that `47203fc2a..bb59fcc30` changes nothing executable, and reported Instructions 293/0 from their own worktrees.

- **GPT (`gpt-5.6-sol`)** — "Confirmed and I agree with the correction: the false `Exact` is pre-existing renderer behavior, not widened by this PR."
- **Gemini (`gemini-3.1-pro-preview`)** — "I agree with the resolution. The worktree is clean and the new head is verified."

One finding was accepted and fixed (the ungated safety claim, `bb59fcc30`, tracked by #3708). One finding was real but pre-existing and misattributed by both models; evidence added to #3661, and measured not to be widened by this change. No blocking finding stands against the change itself.

CI green at `bb59fcc30`, `ci-required` included. Local at this head: Instructions 293/0, full Release solution build clean; at `47203fc2a` the decisive gates were decompiler fast 4496/0 and **fidelity 68/0, unchanged from main**.

Non-blocking follow-up, not part of this slice: #3708 (the malformed-attribute fixture), #3661 (renderer injectivity), and slices 2 and 3 of #3645 (`<>9__` cache fields, then deleting `NormalizeSynthesizedMemberOrdinals` and bumping the fidelity contract to v3).

richlander:
## Retracting "Ready to merge" — I do not have two clean fixed-head reviews

Asked directly whether I had two clean reviews, the honest answer is no, and my previous
comment overstated the position. Correcting it.

What I actually have:

| | |
| --- | --- |
| Two different models, neither my own | yes — GPT `gpt-5.6-sol`, Gemini `gemini-3.1-pro-preview` |
| Two full adversarial rounds at a fixed head | yes, at `47203fc2a` |
| Either of those rounds **clean** | **no** — each produced findings |
| Two **clean** reviews at the final head `bb59fcc30` | **no** |

What I have at `bb59fcc30` is two *confirmation passes*: each reviewer verified that
`47203fc2a..bb59fcc30` changes nothing executable, rebuilt, re-ran Instructions 293/0, and
endorsed my reconciliation. That is a delta check, not an adversarial round. AGENTS.md
says a moved head needs a clean round at the new head, and that the PR is not ready until
every required fixed-head review is clean. Neither condition is met.

There is a defensible argument that the delta is trivial — provably non-executable, and
independently verified as such by both reviewers rather than taken on my word — and that
the round at `47203fc2a` therefore carries. I think that argument is genuinely strong. It
is not strong enough for me to have written "Ready to merge" without stating it, which is
the part I got wrong.

### The substantive gap, which is not procedural

The finding I accepted was *"a comment asserts a property that nothing backs."* My fix was
**another comment**, which makes a new factual claim about the test infrastructure:

> `BuildImage` shares one attribute constructor across every attributed entity, so
> corrupting that constructor also breaks the members the fold is supposed to prove

Neither reviewer audited that claim. Both confirmed the diff was comment-only and moved
on — which is exactly what I asked them to do, so this is my error in scoping the
confirmation, not theirs. I had reproduced every finding of theirs before acting, and then
shipped a claim of my own with less scrutiny than I demanded of them.

I have since checked it myself, by reading rather than assuming
(`CompilerGeneratedOrdinalTests.cs:1665-1686`): `attributeCtors` is a dictionary keyed by
`(Namespace, Name)`, and `compilerGeneratedCtor` resolves through it to a single
`System.Runtime.CompilerServices.CompilerGeneratedAttribute` constructor that both the
type path and every attributed member share. The claim holds. But self-verification of a
claim is precisely what the review round exists to catch, so it does not close the gap on
its own.

### What I am doing

Running a real adversarial round at `bb59fcc30` with both models — not a delta check —
with the accuracy of the new comment and of #3708 as an explicit attack point. I will
report the result here, clean or not, before saying anything about readiness again.

richlander:
## Round 3 reconciliation — head `060dd6c6f`, fixed at `04620db9f`

Two reviewers, isolated worktrees, fixed head `060dd6c6f`. Priority attack point was the
one I could not audit myself: find a tamper that breaks the on-demand declaring-type mark
but that the gate I added in `060dd6c6f` does **not** catch.

Both delivered. Three findings; two actioned, one dismissed with evidence and refiled.

### 1. Per-type cache scope was ungated — GPT — **real, fixed in `04620db9f`**

Exactly the hole I asked for. `bool? typeIsGenerated` is declared inside the type loop.
Hoisting the declaration out of that loop lets a type that earns the mark lend it to the
next type indexed, so an unmarked type's members inherit an ownership they did not earn
and fold. **That tamper passes all 294 tests**, including the malformed-row test I added
in `060dd6c6f` specifically to gate this variable — because with the cache shared, the
first type resolves it and the malformed row is never reached.

Reproduced, then fixed with `TheDeclaringTypeMark_DoesNotCarryToTheNextType`: a marked
type with an unmarked ordinal-bearing member sets the cache; a later **unmarked** type
with an unmarked ordinal-bearing member must stay out of the index. It asserts through
`CompilerGeneratedOrdinalCorrespondence.Build` rather than `IlBodyDiff.Compare`, so it
does not depend on which method the caller targets, and it reads the fixture's attribute
rows and method names back through `MetadataReader` first, so it cannot pass by never
presenting the case.

`BuildImage` hardcoded `"M"` for generated types' methods, which is why no existing
fixture could express the second type; it now takes `generatedTypeMethodNames`.

Tamper table:

| tamper | fails |
| --- | --- |
| hoist the declaration out of the type loop | `TheDeclaringTypeMark_DoesNotCarryToTheNextType`, and nothing else |
| make the attribute read unconditional | `MalformedAttributeOnUnrelatedType_IsSkippedByTheOnDemandRead`, and nothing else |

Two properties, two tampers, two gates, disjoint. That is the point: my `060dd6c6f`
comment claimed the malformed-row test "fails if this read moves out of the loop", and
the hoist tamper disproves it. The comment now names one gate per property and says
explicitly that the malformed-row test cannot see the scope hole.

### 2. The `??=` reuse is ungated — Gemini — **real, comment corrected**

Dropping the caching and reading per member leaves the suite green at 294/0. Verified.
The prose claimed more than the gate held, again.

Widening the gate is not the right answer here, and this is the one case in this PR
where it isn't: the read is pure over the same metadata rows, so reading twice returns
the same answer and throws in the same cases. There is no behavior to gate. The comment
now says so, and marks the reuse as an optimization that is not load-bearing for either
safety property.

### 3. Function-pointer generic arity — GPT — **real defect, wrong PR**

`GetFunctionPointerType` (`IlBodyDiff.cs:1617-1621`) omits
`MethodSignature.GenericParameterCount`, so two function pointers differing only in
generic arity render identically. Reproduced.

Not attributable to this PR, and the decisive measurement is the one the report did not
run — the identical pair with **`IlBodyDiffNormalization.None`**:

```
identical author-written, non-generated name, NO normalization: Exact
```

No name folded, nothing normalized, still `Exact`. Composed is `Exact` at base
`0b2834ba4` as well. The report's evidence — that removing the `b__` branch restores
`OperandDiff` — is true only for **correspondence-only** normalization, which is not a
configuration anything ships. Filed as a new row on #3661
(`#issuecomment-5162775078`) with the cross-head table.

### The pattern I want on the record

Across three rounds, **five of the seven findings were against my own claims, not the
product**. Round 1: my comment asserted an ungated safety property. Round 2: my
explanation of why I could not build the fixture was wrong — I had reasoned about the
wrong metadata column. Round 3: my comment named a gate that does not hold the claim,
and my second comment overstated again.

Every one of my *reasoned* claims that reviewers attacked has failed. Every one of my
*measured* claims has held. The `IlBodyDiffNormalization.None` re-run has now overturned
four false-`Exact` findings; measuring at the base commit has now correctly attributed
two. Neither is clever — they are just the difference between checking and arguing.

### Status

Three rounds, three rounds with findings. **No clean round exists yet**, so this is not
ready to merge. Round 4 runs at `04620db9f`.

### Validation at `04620db9f` (full — product changed, nothing carried over)

| suite | result |
| --- | --- |
| Instructions | 295 / 0 |
| Analysis | 584 / 0 |
| Services | 360 / 0 |
| Metadata | 1397 / 0 (5 skipped) |
| MetadataRendering | 187 / 0 |
| Decompiler `--gate fast` | 4496 / 0 |
| **Decompiler `Area=Fidelity`** | **68 / 0 — unchanged from main** |

Build: 0 errors.

richlander:
## Round 4 reconciliation — head `04620db9f`, unmoved

Two reviewers, isolated worktrees, fixed head `04620db9f`. The brief pushed the majority
of effort onto **product soundness** — construct a false `Exact` where a real
semantically distinct target is hidden by this PR's `b__` ownership — on the reasoning
that three consecutive rounds spent on whether comments name gates that hold is exactly
the condition under which a genuine product defect survives behind clean coverage.

**No actionable finding survived verification. The head did not move.**

### Section A — product soundness: two attempts, neither stands

**GPT — `SZARRAY` vs rank-one `ARRAY` folding to `Exact`.** Real defect, already filed,
not this PR's. GPT did run the mandated `IlBodyDiffNormalization.None` control, but ran
it on a **differing-name** pair (`<M>b__0_0` vs `<M>b__1_0`), so its `None=OperandDiff`
measured the *name* difference and read as evidence the arrays were distinguishable. Run
with identical names at this head:

| pair | `None` | shipped |
| --- | --- | --- |
| identical author-written name `Plain` | **Exact** | Exact |
| identical generated name `<M>b__0_0` | **Exact** | Exact |
| GPT's differing names | OperandDiff | Exact |

Rows one and two are the control: nothing normalized, nothing to fold, still `Exact`.
The renderer cannot tell the two array types apart. Shipped is `Exact` at base
`0b2834ba4` too, which GPT measured itself. It is #3661, filed there with the table.

This is a new failure mode worth naming: the diagnostic was *run*, but on the wrong
pair. "Re-run the identical pair" was not tight enough wording, and the brief now says
the names must be identical.

**Gemini — cross-type lambda swap (`<>c` <-> `<M>d__1`) folding to `Exact`.**
**Retracted.** Reported as High/measured, but the mechanism was internally inconsistent:
it described the diff formatting `call <>c::<M>b__#_0()` and `call <M>d__#::<M>b__#_0()`
"identically", and those are two different strings. Asked for the verbatim probe and the
literal operands, Gemini replied that it "conflated two separate probes ... and
hallucinated a cross-type collision that never actually executed", and withdrew it.

I did not take the retraction on trust either — the scenario is worth an answer on its
own merits. Measured at `04620db9f`, printing the literal rendered operands:

| case | `None` | shipped | rendered |
| --- | --- | --- | --- |
| `<>c` -> `<M>d__1`, same lambda name | OperandDiff | **OperandDiff** | `[<current>]<>c::<M>b__1_0` vs `[<current>]<M>d__1::<M>b__1_0` |
| `<>c` -> `<M>d__1`, differing ordinal | OperandDiff | **OperandDiff** | `<>c::<M>b__1_0` vs `<M>d__1::<M>b__7_0` |
| same `<>c`, ordinal 1 -> 7 | OperandDiff | Exact | — |
| `<M>d__1` -> `<M>d__2`, same lambda name | OperandDiff | Exact | — |

The declaring type is present in the rendered operand, so a lambda that moves between
generated types fails closed. Rows three and four are the intended contract — a scope
ordinal and a `d__` ordinal are precisely what shifts when unrelated code changes, the
containing method name is preserved, and each side has exactly one candidate.

### Section B — audit of the round-3 fix: clean, and the two reviewers agree

Both independently confirmed, by reading the emitted metadata back through
`MetadataReader` rather than trusting a tamper:

- The fixture presents the case it claims: `C` marked, `Target` **not** marked,
  `<M>b__3_0` genuinely owned by `Target`. It cannot pass for an unrelated reason.
- Hoisting the declaration out of the type loop fails **exactly**
  `TheDeclaringTypeMark_DoesNotCarryToTheNextType`.
- Making the read unconditional fails **exactly**
  `MalformedAttributeOnUnrelatedType_IsSkippedByTheOnDemandRead`.

Two properties, two tampers, two gates, disjoint — which is what the corrected comment
now claims, and no wider.

GPT additionally established non-vacuity beyond what was asked: removing both method
ambiguity checks fails 5 tests including both real overload canaries, and a
488,280-name local-function parser fuzz across base and head found zero behavioral
differences. It also measured that tampering the renderer to spell rank-one `ARRAY` as
`int32[*]` fails **only** its own probe out of 295 — so fixing #3661 has no existing
coverage holding the broken spelling in place. Posted to that issue.

The `??=` purity claim remains **reasoned, not measured**, by both reviewers and by me.
It is marked in source as an ungated optimization, which is the honest state.

### Status

| round | head | outcome |
| --- | --- | --- |
| 1 | `47203fc2a` | findings — one real (my ungated comment), one reattributed |
| 2 | `bb59fcc30` | findings — my explanation refuted, real gate built |
| 3 | `060dd6c6f` | findings — real scope hole, my comment corrected |
| **4** | **`04620db9f`** | **clean — nothing actionable, head unmoved** |

Round 4 is the first clean round: two models, one fixed head, no code change required.
Validation at this head was posted above (fidelity 68/0, unchanged from main); CI is
green including `ci-required`, and the PR is MERGEABLE/CLEAN.

Across four rounds, **five of eight findings were against my own claims rather than the
product**, and the product itself survived a round aimed squarely at breaking it. Every
reasoned claim of mine that a reviewer attacked failed; every measured one held.

richlander:
Ready to merge

Both fixed-head reviews at `04620db9f` are clean — GPT and Gemini, isolated worktrees,
same head, no code change required by either. Round 4 was aimed squarely at product
soundness and did not break it; both Section A attempts dissolved under verification,
one reattributed to #3661 and one retracted by its author and then independently
re-measured by me.

**Blocking work is complete:**

- Validation at this head: Instructions 295/0, Analysis 584/0, Services 360/0, Metadata
  1397/0 (5 skipped), MetadataRendering 187/0, decompiler `--gate fast` 4496/0,
  **`Area=Fidelity` 68/0 — unchanged from main**.
- CI green including `ci-required`; PR MERGEABLE/CLEAN.
- Four review rounds; the last is clean at an unmoved head.

**Non-blocking follow-up, all filed:**

- #3661 — renderer non-injectivity. Four rows now, three added during these rounds
  (rank-one `ARRAY` vs `SZARRAY`, type `GenericParam` count, function-pointer generic
  arity), plus the measurement that fixing the array row breaks no existing test.
- The `??=` reuse above `bool? typeIsGenerated` is an ungated optimization. Marked as
  such in source with the reason it needs no gate; not a deferred defect.

**Enumerated residual — this is slice 1 of 3 on #3645:**

- Slice 2: `<>9__N_M` fields, keyed by sibling-pairing to `<Name>b__N_M`. Will move
  fidelity rows.
- Slice 3: delete `NormalizeSynthesizedMemberOrdinals` and the per-side rewrite, bump
  `FidelityCheck.CurrentContractVersion` to 3, regenerate the `contractVersion:1`
  baselines.

Neither is required for this slice to stand on its own: the behavioral claim here is
that lambda overloads stop reporting `Exact` on real Roslyn output, and that is measured
and gated independently of both.

Issue #3590, opened 2026-07-30
Malformed SourceLink is invisible in output: rejected keys and parse errors are recorded but never reported


richlander:
## What happens

An assembly whose SourceLink map is malformed is indistinguishable in output from one whose map is healthy.

Reproduced against a real assembly built with `/p:SourceLink=` pointing at `{"documents":{"/_/*":42}}`:

```text
| Provenance | SourceLink | Present | SourceLink data found in PDB from standalone |
```

That is the same row a working map produces. The signal reports that the SourceLink *blob exists*, not that it maps anything.

## What the library layer already knows

`SourceLinkFetch.SourceLinkResolver` gets this right. For the same PDB:

```text
map present: True
parseError : <null>
rejected   : [/_/*]
isEmpty    : True
established: False  reason: 'no document resolves to a source URL'
```

## Where it is lost

- `SourceLinkProvenanceResult.Reason` is discarded at every product call site — `AssemblyInspector.cs:261`, `AssemblyInspector.cs:315`, and `ILInspector.Metadata/SourceLinkResolver.cs:179` all take `.Origin?.RepositoryUrl` and drop the reason.
- `SourceLinkResolver.RejectedKeys` has **no product consumer at all**. Its doc comment says "Recorded so a rejected key is reportable rather than silently absent" — it is reportable, and never reported.
- `SourceLinkResolver.ParseError` reaches only the discarded `Reason` and `AssemblyInspector.CheckSourceLinkPaths`, where it collapses into "not normalized".

## Why it matters

`AGENTS.md`: "Keep failure visible. Do not turn decode, acquisition, analysis, or rendering failures into success-shaped empty output." A broken SourceLink map currently produces exactly that — the user sees no source and no reason, and cannot tell a broken map from an assembly that ships none.

There is already a natural home for this: `library` emits a **Non-normalized Paths** section from `CheckSourceLinkPaths`, so SourceLink map health is an established output concern.

## Scope

Needs a decision on output shape rather than a mechanical fix: whether rejected keys become a section, a signal value (`Present` vs something like `Present (unusable)`), or both, and where they sit under the verbosity model. Follow-up to #3577, which fixed the resolver so the failure is no longer swallowed at the library layer but did not add reporting.

richlander:
Coupling worth recording before this is picked up: **the strings this issue asks us to report quote artifact text, so surfacing them as-is would create a terminal-injection path that does not exist today.**

[...]

PR #3577, opened 2026-07-30
Collapse SourceLink onto one spec-correct matcher and one provenance owner

richlander:
## What this changes

`ILInspector.Metadata` carried its own SourceLink mapping implementation and its
own provenance extractor, both divergent from `SourceLinkFetch`'s. This collapses
each onto a single owner and fixes what the collapse exposed.

**Neither matcher was correct.** A differential probe over 9 inputs found the two
implementations disagreeing on 6. Checked against the specification
(`dotnet/designs` `accepted/2020/diagnostics/source-link.md`) and the reference
consumer (`Microsoft.SourceLink.Tools.SourceLinkMap`), the most consequential
divergence was that a mid-URL `*` — the shape Azure DevOps emits — produced a URL
containing a literal asterisk, so AzDO-hosted assemblies could not resolve source
at all.

**Provenance was read off the map's text, not off the URL source is fetched
from.** `AssemblyInspector` reported the repository from the *first* `documents`
entry while resolution took the *most specific* match, so a well-formed map could
resolve source from one origin while provenance named another. The unanchored
`GitHubRawUrl()` regex behind it (#3408) is deleted.

## The rule, and three deliberate deviations from the reference

The reference sorts by stripped-key length with an unstable `List<T>.Sort`, and its
own comment states an intent it does not achieve: *"absolute paths will be checked
before a wildcard path with a matching base."* We implement a **total order**, so
resolution cannot depend on enumeration order. We also normalize separators on both
sides, reject a non-conformant entry **individually** into `RejectedKeys` rather
than invalidating the whole map (required by `ANonConformantKey_DoesNotShadowA-
ConformantOne` from #3402), and enforce rule 2's second direction — a wildcard key
paired with a constant URL is rejected, because honouring it serves one file's
content as the source of every document in a subtree. Each deviation is documented
where it is implemented.

## The provenance invariant

> Reported provenance must describe the origin that source content is actually
> fetched from, for every document the assembly resolves. When that cannot be
> established for all of them, report no repository.

`SourceLinkFetch.SourceLinkProvenance` owns it. It resolves every declared document
and reads the origin off each **final resolved URL**, after substitution, escaping,
and `System.Uri` canonicalization — never off the mapping text or prefix. Agreement
is required on the whole `(host, organization, repository, revision)` tuple, because
`raw.githubusercontent.com` serves any revision reachable in a repository, including
the head of an unmerged pull request.

Note the consequence: if *every* document traverses consistently to `attacker/evil`,
[...]

Issue #3391, opened 2026-07-28
SourceLink provenance can name a different origin than the resolved source


richlander:
Found during adversarial review of #3389 by two independent reviewers, and reproduced on `origin/main` (`b07a3ba0`) as well as on that PR head.

## Problem

Two readers select from the same attacker-controlled SourceLink `documents` map using different rules:

- `AssemblyInspector.ExtractRepositoryUrlFromSourceLink` enumerates `documents` and unconditionally `break`s after the **first** entry.
- `SourceDocumentPathResolver.Resolve` orders mappings by **descending pattern length** and takes the first one that actually prefixes the file path.

A well-formed map with distinct keys therefore reports provenance from one origin while fetching source from another. Duplicate-property rejection does **not** address this: for a duplicated key the sort is stable, so both readers land on the same first entry.

## Reproduction

Map:

```json
{"documents":{
  "/_/*":"https://dev.azure.com/good/project/_apis/git/repositories/repo/items?path=*",
  "/_/src/*":"https://evil.example/raw/*"
}}
```

Resolving `/_/src/Program.cs`:

```text
reported repo : https://dev.azure.com/good/project
resolved src  : https://evil.example/raw/Program.cs
```

## Proposed fix

Report no repository when `documents` entries disagree on origin (fail-closed and unspoofable), rather than binding the first entry. Cover both readers in one test, since the existing tests exercise only resolution.

Tracked as open work item 1 in `docs/design/untrusted-data-threat-model.md`.

richlander:
Refining the proposed fix. The original suggestion — report no repository when entries disagree on **origin** — is not sufficient, per adversarial review.

`AssemblyInspector.ExtractRepositoryUrlFromSourceLink` derives provenance as:

```csharp
return $"https://github.com/{match.Groups[1].Value}/{match.Groups[2].Value}";
```

which discards the commit. And `raw.githubusercontent.com` serves **any commit reachable in the repository**, including the head of an unmerged pull request — verified:

```console
$ curl -o /dev/null -w "%{http_code}" https://raw.githubusercontent.com/richlander/dotnet-inspect/025d14ad56527f75cd6187c0f011243e57f9b887/README.md
200
```

Anyone able to open a PR against a public repository can therefore host content under that repository's raw URL. Two `documents` entries naming the same `owner/repo` at **different commits** would "agree on origin" and pass the proposed check, while the resolver still selects the attacker's longer-matching pattern and fetches their commit.

[...]

PR #3712, opened 2026-08-03
Refuse a SourceLink entry whose substitution its host cannot read (#3599)

### LINKED ISSUE #3599: SourceLink: a wildcard confined to a component the host ignores serves one file for every document
## What

`SourceLinkResolver`'s two-probe check (`ProducesAFetchableRequest`) establishes that the document path **changes the request**. It cannot establish that the server **uses** the change, and no host-agnostic rule can, because the same URL shape means opposite things on the two hosts SourceLink maps actually name.

Measured:

| Request | Result |
| --- | --- |
| `raw.githubusercontent.com/o/r/<sha>/One.cs?document=A.cs` | same bytes |
| `raw.githubusercontent.com/o/r/<sha>/One.cs?document=B.cs` | same bytes |
| `raw.githubusercontent.com/o/r/<sha>/One.cs` | same bytes — query ignored |
| `dev.azure.com/.../items?...&path=/README.md` | 200 |
| `dev.azure.com/.../items?...&path=/nope` | 404 — query selects |

So this map is accepted, matches the whole subtree, outranks a valid less-specific entry, and presents one file as the source of every document in it:

```json
{"documents":{
  "/_/src/*":"https://raw.githubusercontent.com/o/r/<sha>/One.cs?document=*",
  "/_/*":"https://raw.githubusercontent.com/o/r/<sha>/*"
}}
```

Refusing the shape is not an option: a wildcard confined to the query is precisely what `Microsoft.SourceLink.AzureRepos.Git` generates, and refusing it would reintroduce the Azure Repos failure the matcher collapse was done to fix.

## Scope

This is **pre-existing**, not a regression. `main` substitutes with a bare `urlTemplate.Replace("*", …)` and performs no URL validation at all.

It is a **correspondence** gap, not a **provenance** one. The origin reported for the map above is `o/r@<sha>`, which is genuinely where the bytes are served from, so the provenance invariant holds. What fails is the correspondence between a document and the content shown for it.

## Fix

Give the known hosts a content selector, alongside the host grammars already in `SourceLinkProvenance` (`raw.githubusercontent.com` → the path; Azure Repos `items` → the `path=` parameter), and require the substitution to reach it for hosts we claim to understand. Unknown hosts keep today's behavior, so a self-hosted SourceLink server is not refused.

Note that a refusal must reach **resolution**, not just attribution: fetching is not gated on provenance, so refusing to attribute would suppress a true statement while leaving the wrong fetch in place.

## Gate

The boundary is currently written down as a decision, with the measurement above, by `SourceLinkMapConformanceTests.AWildcardConfinedToTheQuery_IsAcceptedBecauseOnlyTheHostKnowsWhetherItSelects`.

Found by adversarial review of #3577.

richlander:
Closes #3599.

## What changed

A SourceLink map whose wildcard lands in a component the serving host ignores fetches **one file for every document it matches**. Provenance already declined to attribute such a map. Resolution still fetched and displayed that one file — the half #3599 was explicitly left open for, and which the issue warned about directly:

> a refusal must reach **resolution**, not just attribution: fetching is not gated on provenance, so refusing to attribute would suppress a true statement while leaving the wrong fetch in place.

The matcher now refuses the entry outright, so it lands in `RejectedKeys` and stops shadowing a valid less-specific entry.

On the issue's own map:

| | before | after |
| --- | --- | --- |
| `/_/src/Program.cs` | `One.cs?document=src/Program.cs` | `.../src/Program.cs` |
| `/_/src/Other.cs` | `One.cs?document=src/Other.cs` | `.../src/Other.cs` |
| attribution | refused | `https://github.com/o/r` |

Both documents used to resolve to the same file with no origin. They now resolve to their own files and attribute cleanly, because the refused entry no longer outranks the valid fallback.

## The part that is not the obvious fix

The issue's stated direction — require the substitution to reach the host's content selector — reads as "refuse resolution wherever attribution refuses". **That is wrong, and it fails safely-looking.** Measured over six shapes, attribution refuses five, and only two of those fetch the wrong file:

| Shape | Fetches right file? | Attribution | Resolution now |
| --- | --- | --- | --- |
| GitHub query-wildcard | ❌ one file for all | refuses | **refused** |
| Azure inert-parameter wildcard | ❌ one file for all | refuses | **refused** |
| GitHub path-wildcard + inert query | ✅ | refuses | resolves |
| GitHub branch ref (`/o/r/main/*`) | ✅ | refuses | resolves |
| Self-hosted host | ✅ | refuses | resolves |
| Azure `path=/*` (generated shape) | ✅ | attributes | resolves |

Collapsing the two predicates would have stopped resolving source for **every SourceLink deployment outside the two hosts whose grammar is written down here**, plus branch-based GitHub maps — while looking correct against the defect. So resolution asks the narrower question: an unrecognized host keeps today's behavior, and only a host this reader can speak for, whose selector the substitution demonstrably misses, is refused.

`SourceLinkProvenance.CanSelectContent` is that narrower predicate; `TryCheckSubstitutionSelectsContent` now keys on host rather than on a full origin so both halves share one copy of the grammar.

## Evidence

- `dotnet run --project tests/ILInspector.Metadata.Tests -c Release` — **1403 passed**, 0 failed (was 1397; +6 net)
- `dotnet run --project src/DotnetInspector.Services.Tests -c Release` — 360 passed
- `dotnet run --project tests/DotnetInspector.MetadataRendering.Tests -c Release` — 187 passed
- `dotnet run --project src/dotnet-inspect.Tests -c Release` — 2843 total, 1 failed: `Layout_Count_CountsFilesRatherThanRenderedTreeLines`, **reproduced on clean `origin/main` at `0334a2de5`** in a detached worktree, so it is pre-existing and unrelated (package layout/`--count`, no SourceLink involvement)
- `dotnet build dotnet-inspect.slnx -c Release` — 0 errors; warnings 9 → 7, because splitting the repeated-selector theory removed a duplicate `InlineData` (xUnit1025)

## Gates

- `OnlyAnEntryThatCannotSelectContent_IsRefusedResolution` — the six shapes above. Its three unattributable-but-correct accept rows are the non-vacuity: collapsing the two predicates fails it, and the self-hosted row is the one that fails loudest.
- `AWildcardConfinedToTheQuery_IsRefusedOnlyOnAHostKnownNotToReadIt` — renamed from `…_IsAcceptedBecauseOnlyTheHostKnowsWhetherItSelects`, which pinned the old boundary. It still asserts the deferral that matters (the matcher does not decide this alone; it asks the host grammar), and now also asserts the shadowing recovery.
- `ARepeatedContentSelectorWhoseWildcardIsNeverRead_IsRefusedResolution` — split out of `ARepeatedContentSelector_IsNotAttributable`. Azure serves the *first* occurrence, so a wildcard in a later one is unresolvable while a wildcard in the served one stays merely unattributable. Splitting keeps a future change from moving a row between layers unnoticed.

## Boundary

- No output, section, or CLI surface changes.
- Unrecognized hosts are untouched by design.
- Does **not** address the reporting gap in #3590 — a rejected key is still never surfaced to the user, so this refusal is currently visible only as absent source. That is #3590's scope and is the natural follow-up.

## Review

More than trivial and security-adjacent, with a blast radius over any assembly carrying a SourceLink map — treating as **high risk: two rounds, two different models**, per `AGENTS.md`.


richlander:
## Two bypasses found in review, both fixed in `7998beed1`

Both rounds found real ways through the content-selector check added in `8a5bc3de9`. Both worked the same way: spell a URL component so that **this reader and the host disagree about what it says**, putting the substitution back somewhere the host never reads — #3599's defect again.

### 1. Trailing-dot host — found independently by **GPT-5.6-sol** and **Gemini 3.1 Pro**

`Uri.Host` preserves the DNS root label, so `raw.githubusercontent.com.` compares unequal to `raw.githubusercontent.com` while naming and being served by it. `IsRecognizedSourceHost` read it as an unknown host and stood aside, and the map went back to resolving every document to one file.

Reproduced on a clean exact-head worktree:

```
raw.githubusercontent.com      rejected=YES
raw.githubusercontent.com.     rejected=NO   <-- resolves
dev.azure.com.                 rejected=NO   <-- resolves
```

GPT confirmed with `curl` that GitHub serves the fully-qualified host and returns identical bytes for two different query values. Gemini noted Azure answers 400 for the same spelling, so **only the GitHub row was a live fetch** — the shape was accepted for both.

**Fix:** hosts are canonicalized at the single point *both* readers derive one from a URL, so attribution and resolution cannot drift apart about which host a URL names. That drift is #3391, fixed here once already, and a host spelling is exactly where it would come back. The test asserts the agreement, not just the refusal.

### 2. Percent-encoded parameter name — found by **Gemini 3.1 Pro**

Parameter names were compared against the URL's raw text while the host decodes them, so in `%70ath=/fixed.cs&path=/*` the first pair was invisible here and served there.

I did not take this on the reviewer's reasoning; my first measurement was **inconclusive** (every row returned the same bytes, including the control, which is equally consistent with the server ignoring an unknown parameter). The discriminating experiment against `dev.azure.com/dnceng-public/public`, repo `dotnet-public-wiki` @ `af56d96fdbd7c26e9fc94336b6f50dcc6ceff484`:

| Request | Result |
| --- | --- |
| `%70ath=/README.md` | 200, 985 bytes — **the name is decoded** |
| `%70ath=/README.md&path=/nope.txt` | 200, 985 bytes, **not** 404 — **first occurrence wins** |

Both halves are needed, and both hold. The bypass is real.

**Fix:** names are compared decoded, during the comparison rather than into a buffer, so a name cannot be read one way here and another way by the caller.

### Neither fix is a blanket refusal

A fully-qualified host and a lone `%70ath=/*` are both *working* maps — the host reads the wildcard either carries — and both have accept rows. That is the same discriminator this PR is about: refuse where the host will not read the substitution, and only there.

### Validation

Metadata **1409** (was 1403) · Services 360 · MetadataRendering 187 · CLI 2843 with the one pre-existing `Layout_Count` failure that also fails at `origin/main`. Solution: 0 errors, no new warnings.

Tamper-verified — reverting the canonicalization fails 3 tests, reverting the decoding fails 2.

Not marking ready: fixed-head re-review at `7998beed1` is required and has not run.

richlander:
## Round two: two more bypasses, both the same defect. Fixed in `178c8ad8d`

**Gemini 3.1 Pro** found two more. **GPT-5.6-sol**'s report was honest but **incomplete** — it explicitly did not complete the IDN/Unicode-dot measurements, which is precisely where the blocking finding was, so I am not counting it as a clean round.

### 3. Unicode label separators — `TrimEnd('.')` was not enough

`Uri.Host` preserves Unicode full stops as well as the root dot, so `raw.githubusercontent.com` + U+3002 was read as a host this code had never heard of, and the check stood aside. Gemini reproduced it end-to-end.

I measured how far it goes before acting on it — requesting `dotnet/core`'s README from GitHub by each spelling:

| Spelling | Result |
| --- | --- |
| `raw.githubusercontent.com` | 200, 2389 bytes, `59E1A6989F35` |
| trailing dot | 200, 2389, `59E1A6989F35` |
| U+3002 / U+FF61 | 200, 2389, `59E1A6989F35` |
| **fullwidth `ｒ`** | 200, 2389, `59E1A6989F35` |
| **soft hyphen** U+00AD | 200, 2389, `59E1A6989F35` |
| **zero-width space** U+200B | 200, 2389, `59E1A6989F35` |
| U+2066 | `UriFormatException` — cannot be requested |

So this is not just the separator: **every IDNA mapping is live**, because the request goes out in the IDN form. Folding the four separator code points would have left the fullwidth and zero-width spellings open. The host is now read as `Uri.IdnHost`, which *is* the form that goes on the wire.

**That fix creates a second hazard, which the existing suite caught.** IDNA *deletes* a soft hyphen and a zero-width space rather than refusing them — so a host checked only after canonicalization gets **sanitized into a clean one and attributed**, silently reversing the standing round-17 rule that a live format character in a host label is *refused*. `docs/design/untrusted-data-threat-model.md` settles on reject-don't-sanitize. The inertness check is now split out and applied to the host **as written**, before canonicalization, so `IdnHost` cannot launder a hostile host past it.

**And a crash.** `IdnHost` throws for a scalar IDNA forbids outright (U+2066). `UriFormatException` derives from `FormatException`, *not* `ArgumentException` — my first guard caught the wrong type, so hostile input threw an unhandled exception. Caught properly now, falling back to the literal host, which cannot be requested anyway.

### 4. Valueless parameter — `&path&path=/*`

The scan looked for `=` within the pair and **skipped** a pair that had none, so the wildcard sat in a `path` the host never reads. Gemini measured Azure serving **425 bytes of root listing** for that shape, not the 985-byte file the second pair names. A valueless pair now binds its name and selects the empty span, which no substitution can land inside.

### Still not a blanket refusal

A fully-qualified host, a Unicode-separator host, a lone `%70ath=/*`, and an unrelated valueless parameter beside a real selector are all *working* maps, and all have accept rows. That discriminator is what this PR is about.

### Useful negative results from GPT-5.6-sol

Measured live against Azure and found **not** to shadow `path`: `+` names, double-encoding `%2570`, overlong UTF-8 `%C0%B0`, `;` separators, `%3D`, `%26`, a repeated `?`, and parameters after `#`. Also confirmed `VisualStudioHostSuffix` is `".visualstudio.com"` — leading-dot anchored, so `xvisualstudio.com` cannot match.

### Validation

Metadata **1419** (was 1409) · Services 360 · MetadataRendering 187 · CLI 2843 with the one pre-existing `Layout_Count` failure that also fails at `origin/main`. Solution: 0 errors, no new warnings.

Head has moved to `178c8ad8d`, so a fresh round is owed at the new head. Not ready.

richlander:
## Round three reconciliation — `09ec10a97` → `437602edc`

Four reviewers at `09ec10a97`: **Claude Opus 4.8**, **GPT-5.6-sol**, **MAI-Code**. Gemini 3.1 Pro refused the round outright (safety filter on the adversarial framing) after producing the best findings in rounds one and two, so GPT and MAI-Code were added to keep two non-Opus families on it.

### Blocking — fixed

**`path[]` binds as `path` on Azure DevOps** — GPT-5.6-sol. Confirmed by my own measurement against `dev.azure.com/dnceng-public/public`, repository `dotnet-public-wiki` at `af56d96fdbd7c26e9fc94336b6f50dcc6ceff484`. A hostile map spells the shadowing pair `path[]=/One.cs`, this reader saw only the later `path=/*`, accepted the entry, and the host then served the one named file for every document. #3599 exactly, through a fifth spelling.

The measurement that settles it is the **discriminating control** — give the alias a missing file and let a real `path` follow with a present one, so 404 means the alias bound *and* won:

```
path[]=/nope.txt&path=/README.md      404   binds, and wins
path%5B%5D=/nope.txt&path=/README.md  404   same after decoding
path[0]  path[1]  path%5B0%5D         200   ignored
path.    path.x   [0].path   pathX    200   ignored
```

Without that control the reading inverts: `path[]=/README.md` alone returns the file, which looks like binding but is equally consistent with the parameter being dropped — except that *no* path at all returns the 425-byte root listing, not the file.

The fold lives in `DecodedNameMatches` after decoding, so it is a property of the name and not of `path`: one place covers `path[]`, `path%5B%5D`, `PATH[]`, `%70ath[]` and `scopePath[]`. The spellings the host ignores are deliberately **not** folded — over-refusal is a real defect here, and `path[0]` is now a pinned accept row for that reason. Tamper-verified both ways: dropping the fold fails 4 tests, accepting any suffix fails the `path[0]` accept row.

### Non-blocking / out of scope

**Bidi controls in the host are sanitized rather than refused** — GPT-5.6-sol. Real observation, existing documented non-action. `Uri` itself removes U+202A–U+202E from an authority *before any of this code runs* (measured: `Uri.Host` returns the clean host for all five). `ALiveFormatCharacterInAHostLabel_IsNotAttributable` already excludes U+202E from its refused rows and says why — pinning it would pin the BCL's behaviour and start failing if that normalization changed. Nothing hazardous reaches a sink, because what `Uri` hands back carries no such scalar.

It did catch a real over-claim in my own wording, now corrected: the host is judged **as `Uri` reports it**, not "as written".

**`ContainsEncodedSeparatorOrDotSegment`** — MAI-Code. Verified out of scope: no hits in this PR's diff, introduced by #3577, and it concerns attribution rather than the resolution predicate. Worth an issue on its own terms, not here.

**Two asymmetries** — Opus 4.8, both failing safe, now recorded in `AValuelessContentSelector_IsTheOccurrenceTheHostServes`'s doc comment rather than left implicit: `scopePath=<value>&path=/*` resolves but the host answers 400 for every document, and `path&scopePath=/*` is refused where the host would have selected via `scopePath`.

### Clean

Opus 4.8, live-measured and thorough. Its structural confirmation is the one worth recording: the fetch path derives `IdnHost` the same way `CanonicalHost` does, so reader and fetcher cannot disagree, and the `IdnHost`-throws fallback is unfetchable because `HttpClient` throws identically.

### The pattern this PR keeps confirming

All five bypasses across three rounds are one defect: **spell a URL component so this reader and the host disagree about what it says.** Host spelling (twice), parameter-name encoding, valueless pair, model-binding alias. Assume a sixth exists.

### Validation at `437602edc`

Metadata **1424/0** (1397 at base), Services 360/0, MetadataRendering 187/0, CLI 2843 with the one pre-existing `Layout_Count` failure that reproduces on `origin/main`. Solution builds 0 errors, same 9 pre-existing warnings. `origin/main` integrated.

**Not ready** — the head moved, so round four at `437602edc` is required before that changes.

richlander:
## Round four — `437602edc`, two blocking findings, both fixed at `e1cc93f81`

**GPT-5.6-sol** and **Gemini 3.1 Pro**, both live-measured. They converged on the same bypass independently; GPT found a second one in the other direction. Gemini engaged normally this round with correctness-oriented framing, after refusing round three.

### 1. The bracket rule is wider than one trailing group — BLOCKING, fixed

Both reviewers measured `path[][]` binding as `path`, which the previous commit's single-group fold missed. Sweeping the family showed the host's binder strips **any number of empty groups from either side** — so the **prefix** form binds too, and that is the part the previous fix was furthest from covering and that neither reviewer found:

```
binds:   path[]  path[][]  path[][][]  path[]%5B%5D  path%5B%5D%5B%5D
         []path  [][]path  []path[]  %5B%5Dpath  []PATH  []%70ath
ignored: path[0]  path[1]  path[a]  path[][0]  path[0][]  path[]x  path[]0
         [0]path  [a]path  [].path  [0].path  x[]path  []pathx
         path]  path{}  path()  path%255B%255D  path.  path.x  pathX
400:     path[
```

`DecodedNameMatches` now strips empty groups on both sides of the decoded name. The ignored spellings stay unfolded deliberately — folding them would refuse maps that resolve correctly.

### 2. `path&scopePath=/*` was over-refused — BLOCKING, fixed

GPT-5.6-sol. The selector search took whichever name it found first, so a valueless `path` matched and `scopePath` was never consulted — refusing a map the host resolves correctly.

**I had recorded this as a non-blocking boundary note last round, on Opus 4.8's report. That was wrong.** Over-refusal is a real defect in this predicate, because the only reason it is weaker than attribution is that refusing a working map strands deployments; grading it as a boundary contradicted the PR's own central constraint. GPT was right to call it blocking. The note is corrected rather than left standing.

Nine measured shapes, one model accounting for all nine — first occurrence of each name binds, an empty value is not a selection, `path` outranks `scopePath` when both select, both selecting at once is an error:

```
path&scopePath=/README.md                200 985   empty path falls through
path=&scopePath=/README.md               200 985   explicit '=' is the same
path&scopePath=/nope.txt                 404       and it really selects
path&scopePath=/README.md&path=/nope.txt 200 985   first occurrence, empty
scopePath&path=/README.md                200 985   symmetric
path&path=/README.md                     200 425   root listing
path&scopePath&path=/README.md           200 425   both empty, root listing
scopePath=/README.md                     200 985
path=/README.md&scopePath=/nope.txt      400
```

The fall-through does not become blanket acceptance: `path&path=/*` and `path&scopePath&path=/*` are still refused, because with nothing to fall through to the host serves its root listing for every document.

### Confirmed negatives — Gemini

Measured and correctly **not** folded: double-encoding `%255B%255D`; URL fragments (`HttpClient` strips them before transmission, so this reader's view matches the wire); `path%20` and `path%00`; the near-miss names `itemPath`, `filePath`, `fileName`; and `;` as a separator, which this host does not honour.

### Tamper

Verified separately, not as one run: reverting to a single trailing group fails 3 tests; restoring the first-name-wins short-circuit fails 2.

### Validation at `e1cc93f81`

Metadata **1429/0**, Services 360/0, MetadataRendering 187/0, CLI 2843 with the one pre-existing `Layout_Count` failure that reproduces on `origin/main`. Build 0 errors, same 9 pre-existing warnings. `origin/main` integrated.

### Status

**Not ready.** Six bypasses now, across four rounds, every one the same defect — spell a URL component so this reader and the host disagree about what it says. Round four moved the head and changed the predicate's shape, so a clean round five at `e1cc93f81` is required before this is ready.

richlander:
## Round five — `e1cc93f81`, three blocking findings, all fixed at `cc484cfca`

**Gemini 3.1 Pro** and **GPT-5.6-sol**. Two of the three are bypasses, and one of them is worse than the reviewer who raised it graded it.

### 1. Brackets are deleted from *anywhere* in the name — BLOCKING, too permissive

Gemini measured `p[]ath` binding as `path`. Sweeping further, so do `pa[]th`, `pat[]h`, `p[]a[]t[]h`, `p[][]ath`. Round four's edges-only rule missed **every one of them**.

The deletion is a **single left-to-right pass**, and the host does not rescan what it brings together:

```
p[]ath  pa[]th  pat[]h  p[]a[]t[]h  p[][]ath  p%5B%5Dath   404   bind and win
p[[]]ath  pa[]]th  pa[[]th  p[[]]a[]th                     200   ignored
```

`p[[]]ath` collapsing to `p[]ath` and staying unbound is now the close negative that stops this becoming a blanket bracket strip.

### 2. A non-ASCII case fold invented a selector — BLOCKING, too permissive

`char.ToUpperInvariant('ſ')` is `'S'`, so `ſcopePath` compared equal to `scopePath` here while the host ignores it outright.

GPT-5.6-sol raised this as a lead in the **over-refusal** direction and did not promote it to a finding. Measured the other way round, it is worse than that — put the wildcard in the spelling only this reader believes in:

```
ſcopePath=/README.md                      200 425   ignored, so root listing
ſcopePath=/nope.txt&scopePath=/README.md  200 985   does not bind or win
ſcopePath=/A.cs&scopePath=/README.md      200 985   one file for
ſcopePath=/B.cs&scopePath=/README.md      200 985   every document
PATH=/nope.txt&path=/README.md            404       ASCII still folds
```

Two different documents, one identical file — **issue #3599 again**, not an over-refusal: this reader found its wildcard and accepted, and the host served one file under every document's name. The fold is now ASCII-only, leaving the host's real case-insensitivity intact.

### 3. Blank is decided after decoding — BLOCKING, too strict

Both reviewers. `path=%20`, `path=+`, `path=%09`, `path=%0a`, `path=%0d` and `path=%C2%A0` all fall through to `scopePath` exactly as an absent value does, and `path=%20` alone answers with the root listing. Comparing raw span lengths refused maps the host resolves.

It is emptiness and **not** trimming — `path=%20/README.md` answers 404 rather than serving the file — so a value that merely *contains* blank text still selects. Pinned in both directions.

### Tamper

Verified separately, each caught by a focused test **and** a conformance row: removing the bracket deletion fails 7 tests; restoring the Unicode fold fails 2; comparing raw lengths for blankness fails 2.

### Reviewer notes

GPT-5.6-sol's report was thin this round — its probes did not compile, it ran 2 tests rather than the suite, and it completed no tamper. It still surfaced the `ſ` lead, which was the most serious finding of the round. Gemini's was thorough and correctly measured. Both again returned findings on the newest code rather than the code that had already been reviewed.

### Validation at `cc484cfca`

Metadata **1435/0** (1397 at base), Services 360/0, MetadataRendering 187/0, CLI 2843 with the one pre-existing `Layout_Count` failure that reproduces on `origin/main`. Build 0 errors, same 9 pre-existing warnings. `origin/main` integrated.

### Status

**Not ready.** Nine bypasses now, across five rounds, every one the same defect — spell a URL component so this reader and the host disagree about what it says. Round five moved the head and changed the name comparison, the case rule and the emptiness rule, so a clean round six at `cc484cfca` is required.

richlander:
Integrated `origin/main` (5c6ca5f93, generic constraints in the single-member decompiled view) at merge commit `ad23f33e3`. Clean merge; it touches `CSharpDeclarationWriter` and its tests, nothing this PR owns. Re-ran the build (0 errors) and `ILInspector.Metadata.Tests` (1435 passed, 0 failed) at the merge head. The head moved, so round six reads `ad23f33e3`, not `cc484cfca`.

richlander:
Integrated current `origin/main` at merge commit `b40936c17` before round six. The incoming commits concern C# type-name spelling and switch/decompiler behavior; the merge is clean and does not touch this PR’s four SourceLink files. At the merged head, the solution builds with 0 errors and `ILInspector.Metadata.Tests` passes 1440/0. Round six will review exact head `b40936c17` after current-head CI is green.

richlander:
## Round six — one blocking over-refusal, fixed in `a51ecf903`

**Gemini 3.1 Pro: CLEAN.** It ran the 1440-test metadata suite and live-probed bracket deletion/no-rescan, ASCII-only name folding, all 25 `char.IsWhiteSpace` scalars, malformed escapes, selector precedence, fragments, IDNA, and the resolution call site.

**GPT-5.6 Sol: BLOCKING (too strict).** It found that valued `path` and `scopePath` were handled asymmetrically. Azure returns 400 and 369 bytes whenever both first-occurrence selectors are valued, regardless of which carries the wildcard:

```
path=/README.md&scopePath=/A.cs  400 369
path=/README.md&scopePath=/B.cs  400 369
scopePath=/README.md&path=/A.cs  400 369
scopePath=/README.md&path=/B.cs  400 369
```

So neither direction can serve one fixed file, and both must be left to resolve so the fetch failure stays visible. Exact head `b40936c17` instead rejected `path=/README.md&scopePath=/*` while accepting `scopePath=/README.md&path=/*`. I reproduced both the live host behavior and the product asymmetry independently.

**Fix:** inspect both selectors before choosing a span. If both select, classify the request as always failing and allow resolution regardless of which value contains the substitution. Added both directions to the resolution conformance table.

**Evidence:** metadata/SourceLink 1442/0; solution build 0 errors. Tamper-verified: applying only the two new rows to the old production code fails exactly the `scopePath`-wildcard direction (1 failure).

The head moved to `a51ecf903`, so a fresh fixed-head round is required. Not ready.

richlander:
Integrated current `origin/main` at merge commit `594b1354b` before the user-approved seventh round. The merge is clean; the PR diff remains exactly the four SourceLink files. The only incoming SourceLink-adjacent change adds `SourceLinkService.OpenMetadataOnly` and does not alter map parsing, resolution, host grammar, or fetching. At the merged head, the solution builds with 0 errors and `ILInspector.Metadata.Tests` passes 1432/0. Round seven will review exact head `594b1354b` after current-head CI is green.

richlander:
Before round seven started, `origin/main` moved once more by `30aa2cf4a` (`--out` for package counts). It touches no PR files and no SourceLink behavior. Integrated cleanly at `677346363`; the PR diff remains the same four SourceLink files. Build: 0 errors. Metadata/SourceLink: 1432/0. The exact review range is now `30aa2cf4a..677346363`, pending current-head CI.

richlander:
## Round seven — one blocking bypass fixed; one raw-curl lead dismissed

**GPT-5.6 Sol: BLOCKING (too permissive).** Entry validation substituted a nonblank `a`, so valued `path` plus valued `scopePath` looked guaranteed to return 400. A real wildcard match can have an empty or whitespace-only remainder, making that selector blank; Azure then falls through and serves the other fixed selector.

Exact-head, end-to-end reproduction in both directions:

```
path=/README.md&scopePath=        200 985
scopePath=/README.md&path=        200 985
path=/README.md&scopePath=%20     200 985
```

All three bodies matched README SHA-256 `0129277c5fd5e35a1b9436a65e76efa257eef51926d5a437ac96b75eceba6e85`. The maps were accepted and fetched README for a document whose wildcard remainder was empty/blank. Attribution declined, but resolution and fetching are independent, so that did not close #3599.

**Fix (`288cb7301`):** keep the entry-level nonblank probe, then validate the concrete substituted URL in `TryResolve`. An unsafe specific match now yields to a valid less-specific entry. Nonblank dual selectors still resolve to the host’s visible 400 rather than being over-refused.

**Gemini 3.1 Pro:** raised `%uXXXX` legacy decoding as a critical bypass based on raw curl. The host observation is real—raw `%u0070ath` binds as `path`—but it is **not reachable through the product transport**. `System.Uri`/`HttpClient` sends the percent as `%25`: the exact-head resolved URL containing `%u0070ath` was transmitted as `%25u0070ath`, Azure ignored it, and the real `path=nope.txt` returned 404/524. Likewise `path=%u0020` was transmitted as `%25u0020` and returned 400/369 rather than falling through. No product change made.

**Evidence:** metadata/SourceLink 1434/0; solution build 0 errors. Tamper-verified: applying only the new outcome test to reviewed head `677346363` fails both selector directions (2 failures). The test also pins that empty/blank remainders fall through, nonblank remainders preserve the 400-producing URL, and the entry is not blanket-rejected.

Head moved to `288cb7301`; another fixed-head round requires fresh user approval under the after-six-round gate. Not ready.

richlander:
Integrated current `origin/main` at `1e2f3a285` for the user-approved eighth round. No PR-file overlap. The incoming HTTP-timeout change affects client duration/configuration, not URI canonicalization or SourceLink map semantics; untrusted fetches retain their fixed cap. The PR diff remains the same four SourceLink files. Build: 0 errors. Metadata/SourceLink: 1330/0. Exact proposed review range: `de56a0e11..1e2f3a285`, pending current-head CI.

richlander:
Before round eight started, `origin/main` moved by `c4db43c69` (obsolete repository NuGet config removal). No overlap or SourceLink/fetch interaction. Integrated cleanly at `991aa2fd7`; the PR diff remains the same four SourceLink files. Build: 0 errors. Metadata/SourceLink: 1330/0. Exact proposed review range is now `c4db43c69..991aa2fd7`, pending current-head CI.

richlander:
Round eight at `7ca8c7e892e83b101b6abeea13b65f574c8ebd14`:

- Gemini Pro: clean after auditing concrete substitution, blank-selector fallback, dual-selector 400 preservation, parameter normalization, and transport canonicalization; focused Release suite passed.
- GPT-5.6 Sol: found that both Azure attribution query readers used `TrimStart('?')`. A URL beginning `items??versionType=commit&version=<40-hex>&path=/*` was resolved correctly by `path`, but attribution discarded both delimiters, read `versionType=commit`, and reported the 40-hex value as a commit while Azure treats the first parameter as `?versionType` and defaults the value to a branch.

Reproduced on the untouched exact-head worktree: `System.Uri.Query` and `PathAndQuery` preserve `??`; the product resolved the URL and established the supplied SHA as provenance. Fixed in `65a8762c064e1cb8d2d859a013fc4777f4522e08` by removing exactly one query delimiter through one shared reader. The new outcome gate requires the URL to remain resolvable (because `path` still selects the requested document) but unattributable; mutation-restoring `TrimStart` fails the gate. The threat model records the parser disagreement and resolution boundary.

Validation at the fixed head: Release solution build, 0 errors; Metadata/SourceLink suite, 1,333 passed; markdownlint clean. Fixed-head re-review remains pending current-head CI.

richlander:
Round eight fixed-head re-review is clean at `65a8762c064e1cb8d2d859a013fc4777f4522e08`.

- GPT-5.6 Sol: no significant issues found across the full PR diff after the query-delimiter fix.
- Gemini Pro: clean after verifying exact-one-delimiter parsing, preservation of `path`-based resolution, empty-name/fragment boundaries, transport canonicalization, and the full Metadata/SourceLink Release suite.

Both reviewers evaluated the same exact base/head in isolated worktrees. The prior `??versionType` blocker is closed without widening resolution refusal: the URL remains resolvable when `path` selects the requested document, but cannot establish commit provenance when Azure reads the first name as `?versionType`.

richlander:
Ready to merge

Issue #3658, opened 2026-08-01
SourceLink silently emits nothing when the current branch's ref is packed


richlander:
Found while investigating a test failure that looked like a product regression and was not.

## Symptom

`CliDiscoverySections_AreSelectable` fails with *"member -S 'Source Locations' produced no data after being listed by -D"*. More generally, **any test that depends on SourceLink data in the test assembly's PDB fails**, because the build emits no SourceLink at all:

```
_GenerateSourceLinkFile:
  Source Link is empty, file '.../ILInspector.Text.sourcelink.json' does not exist.
```

There is **no warning** — `0 Warning(s)`. The build is green and the PDB is simply missing its SourceLink blob.

## Root cause

`Microsoft.Build.Tasks.Git` does not resolve a HEAD symref whose target lives only in `.git/packed-refs`.

`git gc` / `git pack-refs` moves loose refs into `packed-refs` over time. Once your branch's ref is packed, `LocateRepository` produces no `SourceRoot.RevisionId`, `_InitializeGitHubSourceLinkUrl` maps nothing, and `@(SourceLink)` is empty.

## Evidence

Measured matrix, same repo, SDK `11.0.100-preview.7.26366.102`:

| Checkout | HEAD | Ref storage | SourceLink |
| --- | --- | --- | --- |
| `dotnet-inspect` (primary) | `main` | loose | ✅ Updating Source Link file |
| `di-3631` (linked worktree) | `fix/3631-…` | loose | ✅ |
| `di-json-lowered` (linked worktree) | `feat/json-lowered-view` | **packed only** | ❌ Source Link is empty |
| `di-markout-adopt` (linked worktree) | `feat/adopt-markout-windowing` | **packed only** | ❌ |
| `di-markout-adopt` | **detached** at same commit | n/a (SHA in HEAD) | ✅ |
| `di-baseline` (linked worktree) | detached | n/a | ✅ |

The decisive A/B: in **one** worktree, with the **same** working tree and the **same** uncommitted changes, `git checkout --detach` flipped the result from "Source Link is empty" to "Updating Source Link file", and checking out a freshly-created (therefore loose) branch flipped it back to working. Nothing but ref storage changed.

Note this is *not* a worktree bug, which is what it first looked like — `di-3631` is a linked worktree on a branch and works fine. Linked worktrees are only over-represented because their branches are long-lived enough to get packed.

## Why it matters

[...]

PR #3713, opened 2026-08-03
Honor unprobed sections in discovery selection tests

richlander:
`CliDiscoverySections_AreSelectable` now tests the actual `-D`/`-S` contract: every discovered section must be selectable, while only sections whose effectiveness was probed must return data.

## Behavioral claim

A section declared with `ProbeEffectiveness=false` may be listed structurally by `-D` and validly render no data under `-S`. The test now derives those exemptions from the exact type-detail and member-overload pipelines it exercises, while preserving its non-empty guard for probed sections.

No product code changes. No SourceLink-specific behavior or workaround is added.

## Why #3658 exposed the test bug

The .NET 11 SourceLink task can silently omit the SourceLink blob in a linked worktree because it looks for packed refs and reftable state under `$GIT_DIR` rather than `$GIT_COMMON_DIR`. That makes `Source Locations` unavailable, exactly as when SourceLink is disabled or the inspected artifact simply has no SourceLink data.

The product already handles that input correctly:

- `-D` lists `Source Locations` structurally because its descriptor sets `ProbeEffectiveness=false`.
- `-S "Source Locations"` succeeds and renders no data because no locations are available.

`ISectionDescriptor.ProbeEffectiveness` explicitly documents this tradeoff: structurally listed unprobed sections may render empty when queried. The test contradicted that contract by requiring non-empty data from every section in two selected command shapes, making valid SourceLink unavailability look like a product regression.

## Design

The non-empty guard now takes the section name and excludes the sets returned by:

```csharp
ApiMemberSectionDescriptors.CreatePipeline().GetUnprobedSections()
ApiMemberOverloadSectionDescriptors.CreatePipeline().GetUnprobedSections()
```

The declarations therefore drive the enforcement set: a future `ProbeEffectiveness` change automatically changes what the test requires, with no SourceLink allow list or duplicated section names.

The test still asserts that every section reported by `-D` is selectable by `-S`. The existing `SourceLink: Integrity` status-result handling remains unchanged.

## Evidence

Both states were exercised on the integrated head:

| Test assembly build | `-D` lists `Source Locations` | `-S` exit | selected output | test |
| normal | yes | 0 | rows | passes |
| `/p:EnableSourceLink=false` | yes | 0 | empty | passes |

The disabled build is the packed-ref-equivalent canary: it proves the test accepts SourceLink unavailability without hiding a selection failure.

After merging current `origin/main` (`684f947c`) and activating the repository's IL oracles:

```text
dotnet run --project src/dotnet-inspect.Tests -c Release
Total: 2899, Errors: 0, Failed: 0, Skipped: 4, Not Run: 0
```

## Non-action boundary

- No product behavior changes: unavailable SourceLink remains a valid no-data result.
- No SDK or Git workaround is embedded in the test.
- No test is skipped; `-D`/`-S` selectability remains enforced for every discovered section.
- The upstream SDK defect remains documented on #3658 and dotnet/sourcelink#1290/#1657.

Refs #3658


richlander:
## Review round 1 — GPT-5.6 (max effort), reconciled

Reviewed at `319458d2` in an isolated detached worktree. Three Medium findings, **all three reproduced independently before acting**, all three fixed in `4b6baa4e`. No finding dismissed.

### 1. The probe accepted a PDB belonging to another assembly — fixed

The sharpest finding: `SourceLinkReader` does not validate PDB identity, so a stale or foreign PDB carrying its own SourceLink map satisfied the probe while the product (`PdbContext.PdbMatchesAssembly`) correctly rejected it. The hint therefore went silent in exactly the case that needed it.

Reproduced by copying `dotnet-inspect.pdb` over `dotnet-inspect.Tests.pdb`:

```
member -S 'Source Locations' produced no data after being listed by -D.
```

— the bare pre-PR message. Fixed by probing through `PdbContext.Open`, which is both identity-validating and the same question the product asks. It now reports:

> the adjacent PDB is not its own (Portable PDB identity mismatch: dotnet-inspect.Tests.pdb does not match dotnet-inspect.Tests.dll)

### 2. Packed refs were asserted as the cause without being verified — fixed

Correct, and it was the same class of error this PR exists to prevent. The probe establishes only that SourceLink is absent; the hint asserted *why* and prescribed a git repair for every cause, including a deliberately disabled build that no ref surgery can fix.

The hint now states the observation as fact, offers #3658 as the *known candidate* with a confirmation check, and names the causes needing no repair. The reviewer's `$GIT_COMMON_DIR` correction is folded in — a linked worktree's `packed-refs` is in the common dir, so the check is now `test -f "$(git rev-parse --git-common-dir)/refs/heads/$(git rev-parse --abbrev-ref HEAD)"`.

### 3. `Context: Source Location` is not SourceLink-backed — fixed

Verified with a direct A/B on token `0x060007A2`:

| Blob | Result |
| --- | --- |
| present | method, token, file, line, **URL** |
| absent | method, token, file, line — exit 0 |

It resolves from sequence points and only loses the URL, so hinting there would misattribute an unrelated token or PDB failure to the environment. Removed. The predicate now also requires the command to inspect the test assembly, per the reviewer's point that `package` and `diff` reach artifacts whose SourceLink is independent of this build.

### 4. Non-vacuity gate — added (was flagged non-blocking)

`TestAssemblySourceLinkTests` asserts the probe reports availability under a normal build. Both failure modes were silent: a probe stuck at "available" never fires and leaves #3658's misdiagnosis in place; one stuck at "unavailable" smears an irrelevant note across unrelated failures. Nothing else in the suite would have noticed either way. This is the gate that `AGENTS.md` § "Asserted properties name their gate" asks for, and the class doc names it as such.

### Confirmations from the round

Points 2 (prefix literal), 3 (thread-safety, disposal, exception handling), 5 (fail-not-skip), and 7 (no other affected tests) came back clean. The reviewer independently confirmed the full suite without SourceLink yields exactly one failure — `CliDiscoverySections_AreSelectable` — matching the author's evidence.

### Validation at the new head

Three states, each with the reason it should produce:

| State | Probe reason | Result |
| --- | --- | --- |
| normal build | none | ✅ both tests pass |
| foreign PDB | identity mismatch | ❌ fails, names the mismatch |
| `/p:EnableSourceLink=false` | no blob | ❌ fails, names the missing blob |

Full CLI suite after `dotnet build dotnet-inspect.slnx -c Release`: **2844 tests, 0 failed, 14 platform skips.**

Head moved to `4b6baa4e`; re-review of that head is pending.

richlander:
## Review round 2 — GPT-5.6 (max effort), reconciled

Reviewed at `4b6baa4e`. Round 1's findings 1 and 3 confirmed **correctly and completely fixed**. Two new Medium findings, both valid, both fixed in `ffcf88e7`. Neither dismissed.

### 1. The confirmation check misdiagnosed detached and unrelated states — fixed

The reviewer is right, and this was the same failure mode the PR exists to prevent: a diagnostic that misleads. `test -f refs/heads/$(git rev-parse --abbrev-ref HEAD)` **succeeds in the healthy loose-ref state** and fails in states unrelated to #3658, so its polarity was ambiguous. In a detached worktree it tested `refs/heads/HEAD` — and the prescribed repair would then have created a branch literally named `HEAD`.

The check now requires all three conditions that actually define the defect: HEAD on a branch, a linked worktree, and no loose ref under the common dir. Verified across four checkout shapes:

| Checkout | Reports #3658? |
| --- | --- |
| primary, on a branch | no ✅ |
| linked worktree, loose ref | no ✅ |
| linked worktree, **detached** | no ✅ |
| linked worktree, **packed ref** | **yes** ✅ |

### 2. The repair was also wrong — and so is the workaround in issue #3658

Chasing the above surfaced something worth flagging on its own. **`git update-ref <branch> HEAD` does not work.** `update-ref` to an unchanged value is a no-op, so the ref stays packed and SourceLink stays missing. That is the workaround recorded in #3658, and I have now measured it failing. `git branch -f` refuses outright for a branch checked out in a worktree.

What does work, verified with an uncommitted file present to prove it is non-destructive — HEAD stays attached, working tree untouched, same SHA:

```bash
B=$(git symbolic-ref -q HEAD); S=$(git rev-parse HEAD)
git update-ref -d "$B" && git update-ref "$B" "$S"
```

The hint prescribes this and calls out the no-op explicitly, so the obvious command is not retried. I will correct #3658's Workarounds section separately.

### 3. The gate only tested one branch — fixed

Correct, and it was the mutation that matters: a probe hard-wired to "available" would have passed the gate *and* the full suite while silently retiring the diagnostic, restoring the exact investigation this PR exists to prevent.

`DescribeUnavailability` now takes a path, and the gate drives it over a copy with no adjacent PDB and a copy carrying another assembly's SourceLink-bearing PDB. Both mutations injected and confirmed killed:

| Injected mutation | Result |
| --- | --- |
| `DescribeUnavailability => null` | 2 tests fail ✅ |
| `DescribeUnavailability => "…"` | 1 test fails ✅ |

The class doc now claims only what the gate enforces, per `AGENTS.md` § "Asserted properties name their gate".

### On the reviewer's supported-scenarios question

Explicitly checked and agreed with the reviewer's own conclusion: a source archive with no `.git` and `/p:EnableSourceLink=false` already fail the full suite through `CliDiscoverySections_AreSelectable`, and CI uses `actions/checkout`. The gate adds no new red scenario.

### Validation at the new head

- Full CLI suite: **2846 tests, 0 failed, 14 platform skips.**
- Gate: 3/3 pass on a normal build; both mutants killed.
- `/p:EnableSourceLink=false` end-to-end: still reports the missing blob, now carrying the corrected check and repair.

Head moved to `ffcf88e7`; re-review of that head is pending.

richlander:
## Review round 3 — GPT-5.6 (max effort), reconciled

Reviewed at `ffcf88e7`. Finding 5 (gate non-vacuity) confirmed **fixed** — the reviewer independently injected both mutations and reproduced 2 kills and 1 kill, and confirmed the foreign-PDB test **ran rather than skipped in CI**. Findings 1–3 remain fixed. The public correction on #3658 about `git update-ref` was independently **confirmed correct**.

Two new Medium findings, both valid and both reproduced, fixed in `4ad94d52`.

### 1. The confirmation check false-positived in a primary checkout — fixed

The sharper of the two, and the reviewer predicted the mechanism before testing it. Git answers `--git-common-dir` *relatively* when invoked below the repository root while `--git-dir` stays absolute:

```
$ cd dotnet-inspect/src
git-dir:        /home/rich/git/dotnet-inspect/.git
git-common-dir: ../.git          # <- string comparison says "linked worktree"
```

So a developer pasting the check from any subdirectory of a primary checkout would be told they had #3658 and sent into ref surgery on a tree that emits SourceLink correctly. `--path-format=absolute` on both sides fixes it; a `show-ref --verify` clause was added alongside so a branch whose ref is simply *gone* is not reported as merely packed.

### 2. The repair discarded the reflog and had an unsafe window — fixed

Also correct. Deleting the current branch drops its reflog, and between the delete and the recreate symbolic HEAD points at a missing ref — interrupt there and `git rev-parse HEAD` fails outright. The reviewer's suggested rename-and-back is strictly better and is what now ships: HEAD follows each rename and stays resolvable throughout, so an interruption leaves the branch under its temporary name rather than gone.

### Validation — shipped strings, run verbatim

Both commands were extracted from the *emitted failure message*, not retyped, and run as-is. The check reports a match in exactly one of eight checkout shapes:

| Shape | Reports #3658? |
| --- | --- |
| primary, root, loose | no ✅ |
| primary, subdir, loose | no ✅ |
| primary, root, packed | no ✅ |
| primary, **subdir, packed** (was the false positive) | no ✅ |
| linked worktree, loose | no ✅ |
| linked worktree, detached | no ✅ |
| linked worktree, ref deleted entirely | no ✅ |
| **linked worktree, packed ref** | **yes** ✅ |

The repair clears it while preserving HEAD attachment, the reflog (1 → 3 entries), upstream tracking config, and an uncommitted file.

### End-to-end closure

On a scratch repository with a real project, a linked worktree with a packed ref builds:

```
_GenerateSourceLinkFile: Source Link is empty, file 'lib.sourcelink.json' does not exist.
```

After the shipped repair, the same build reports:

```
_GenerateSourceLinkFile: Updating Source Link file 'obj/Release/net10.0/lib.sourcelink.json'.
```

And `gc.packRefs false` keeps the ref loose across `git gc --prune=now`. **Every claim the hint makes is now measured**, not asserted.

Full CLI suite: **2846 tests, 0 failed, 14 platform skips.**

Head moved to `4ad94d52`; re-review of that head is pending.

richlander:
## Review round 4 — GPT-5.6 (max effort), reconciled

Reviewed at `4ad94d52`. Findings 6 and 7 confirmed **fixed**, with the reviewer independently reproducing the mutation kills (2 and 1) and the full suite (2846 / 0 failed / 14 skipped). Two new Medium findings, both valid, both reproduced, fixed in `05626ff4`.

### 8. The repair could not work on a reftable repository — fixed

The best finding of the review, and it came from decompiling the installed task rather than reasoning from the issue text. The reviewer noticed the upstream fix title says *"Use common dir to locate packed-refs **and reftable**"*, then confirmed the shipped .NET 11 task loads `_gitDirectory` on its reftable path too.

That breaks the framing this hint inherited from #3658. A reftable repository keeps a **per-worktree ref stack**, so a linked worktree is affected *whether or not anything was ever packed* — there is nothing to pack. Measured on a scratch reftable repo with a real project:

| | build output |
| --- | --- |
| linked worktree | `Source Link is empty, file 'lib.sourcelink.json' does not exist.` |
| primary checkout | `Updating Source Link file 'obj/Release/net10.0/lib.sourcelink.json'.` |

The **check was right** — it flags these correctly. The **repair was not**: `git branch -m` writes to the reftable stack and produces no loose ref, so the check still matches afterward. `gc.packRefs` is meaningless there, and the obvious escape is closed too:

```
$ git refs migrate --ref-format=files
error: migrating repositories with worktrees is not supported yet
```

So a developer under reftable would have run the repair, seen no change, and had nowhere to go. The hint now splits on `git rev-parse --show-ref-format`: `files` keeps the rename repair and `gc.packRefs false`; `reftable` says plainly that no ref-layout repair applies and points at the primary checkout or a fixed SDK (10.0.302).

### 9. Claims asserted properties without naming a gate — fixed

Correct on the rule. `AGENTS.md` requires a doc comment asserting a soundness property to **name the gate that enforces it, or mark the property unverified**, and these did neither: the "only the real defect reports a match" and "both established by A/B-ing" claims were manual measurements written as if enforced. `TestAssemblySourceLinkTests` gates `DescribeUnavailability` — it says nothing about the shell strings or the section set.

I took the "mark unverified" branch rather than building gates. A test that shells out to construct eight checkout shapes, or that A/Bs a `/p:EnableSourceLink=false` build graph, would cost more than the diagnostic it protects. Both sites now say `UNVERIFIED BY ANY GATE`, name what was measured by hand, and tell the next editor to re-run it.

### Validation

Command strings extracted from the **built assembly** (`strings -el` on `dotnet-inspect.Tests.dll`), not retyped, and run verbatim. Sixteen shapes, two ref formats:

| Shape | files | reftable |
| --- | --- | --- |
| primary root / subdir, loose | no ✅ | no ✅ |
| primary root / subdir, packed | no ✅ | no ✅ |
| linked worktree, unpacked | no ✅ | **MATCH** ✅ |
| linked worktree, packed | **MATCH** ✅ | **MATCH** ✅ |
| linked worktree, detached | no ✅ | no ✅ |
| linked worktree, ref deleted | no ✅ | no ✅ |

Every match is a **true** positive — the reftable worktree rows are exactly the state proven broken by the build above. No false positives in either format.

Repair re-verified under `files` on a **slash-containing** branch: check cleared, HEAD still attached, reflog 1 → 3, `branch.feat/x.remote` intact, uncommitted file untouched. Collision case fails safely — `fatal: a branch named 'feat/x.unpack-tmp' already exists`, exit 128, HEAD unchanged, nothing mutated.

Full CLI suite: **2846 tests, 0 failed, 14 platform skips.**

Head moved to `05626ff4`; re-review of that head is pending. This will be round 5 of the 6 permitted.

richlander:
## Review round 5 — GPT-5.6 (max effort), reconciled

Reviewed at `05626ff4`. Findings 8 and 9 confirmed **fixed** — the reviewer verified the ref-format split by decompiling both SDKs (10.0.302 reads `packed-refs` and reftable from `commonDirectory`; the selected .NET 11 SDK reads them from `gitDirectory`), confirmed `git refs migrate` refuses with worktrees present, and agreed the unverified marking is the right call rather than an abuse of the hatch — a command-matrix gate would need backend/version/shell integration infrastructure, and the predicate gate would need isolated dual builds. It also re-ran the mutation kills (2 and 1), the full suite, and confirmed the probe returns reasons rather than throwing for zero-length, corrupt, directory, dangling-symlink, and unreadable PDBs.

One new Medium finding, valid and reproduced, fixed in `67f4f71d`.

### 10. `gc.packRefs false` does not actually prevent recurrence — fixed

The recurrence advice was the one part of the hint never tested against Git's *other* repacking path. `gc.packRefs` governs `git gc` alone; the `pack-refs` maintenance task runs `pack-refs --all --prune` unconditionally — the reviewer found `UNUSED struct gc_config *cfg` in Git 2.53's `maintenance_task_pack_refs`. So a repository with background maintenance enabled repairs itself, gets told it is protected, and silently re-breaks on the weekly schedule.

Reproduced directly, with `gc.packRefs=false` set throughout:

| step | loose ref survives? |
| --- | --- |
| `git gc --prune=now` | yes |
| `git maintenance run --task=pack-refs` | **no — packed** |

And with the new setting added, after re-repairing:

| step | loose ref survives? |
| --- | --- |
| `git gc --prune=now` + `git maintenance run` | **yes** |
| explicit `git maintenance run --task=pack-refs` | no |

That last row is why the wording changed too. Neither setting constrains an *explicit* `git pack-refs`, so the hint now claims only that it stops **automatic** repacking rather than implying recurrence is impossible — the same discipline applied to the diagnosis in round 1.

Shipped settings, extracted from the built assembly and run verbatim on a scratch worktree, survive `git gc --prune=now` followed by a full `git maintenance run`:

```
git config gc.packRefs false && git config maintenance.pack-refs.enabled false
```

Full CLI suite: **2846 tests, 0 failed, 14 platform skips.**

Head moved to `67f4f71d`; re-review of that head is pending. This will be round 6, the last permitted without fresh approval.

richlander:
## Review round 6 — two reviewers, two families, reconciled

Round 6 ran **GPT-5.6 (max effort)** for continuity and **Gemini 3.1 Pro** as a fresh family, both at `67f4f71d`. They disagreed, and the disagreement was the useful part.

GPT confirmed finding 10 fixed, re-ran the suite and the SourceLink-disabled A/B, and recommended **merge** — explicitly judging that "the inline guidance now appears converged; shortening it is not warranted."

Gemini recommended **do not merge**, on two findings. I reproduced both; both are right, and both are fixed in `3086d5a5`.

### 11. An unreachable clause in the predicate — fixed

Concrete and checkable, and neither I nor five GPT rounds caught it. `IsSourceLinkBackedSection` matched `section.StartsWith("SourceLink:")`, but those sections sit behind the `@SourceLink` **category door**, which `-D` reports as a category rather than expanding. The discovery loop therefore never selects one. Measured against the test assembly:

| command | section rows | `SourceLink:` rows |
| --- | --- | --- |
| `library` | 11 | **0** |
| `type` | 17 | **0** |

So the clause was decoration that read as protection. `Source Locations`, on `member` commands, is the sole live path — which the method's own remarks already said. Clause removed, and the remarks now record *why* the family is absent so it does not get "helpfully" re-added.

### 12. The embedded shell advice should not be there at all — fixed

The design finding, and the one I think is correct despite GPT's contrary verdict. The record is hard to argue with: **six rounds produced six defects in that text and none in the surrounding code.**

| round | defect | all in the hint text |
| --- | --- | --- |
| 2 | check succeeded in the healthy state; could create a branch named `HEAD` | ✔ |
| 3 | false-positived in a primary checkout below the root | ✔ |
| 3 | repair destroyed the reflog, left HEAD unresolvable if interrupted | ✔ |
| 4 | repair could not work under `reftable` | ✔ |
| 5 | `gc.packRefs false` did not prevent recurrence | ✔ |
| 6 | — | |

That is not a text that is one fix away from converged; it is an unbounded surface. Correct advice varies by ref-storage backend, by whether git is invoked at the repository root, and by which of `gc.packRefs` and `maintenance.pack-refs.enabled` a repository honours — and **no test can execute a shell string**, which is exactly why round 4 forced it to be marked `UNVERIFIED BY ANY GATE`. Shipping unverifiable instructions that drift silently is a poor trade for a diagnostic whose job is to stop a developer misreading an environment defect as a regression.

The hint now keeps what the assertion actually needs — what is missing, that the environment rather than the product is implicated, and the one-line mechanism — and points at #3658 for confirmation and repair. Everything removed **already lives in that issue**, with the sixteen-shape matrix, both backends, and the maintenance caveat, where it can stay current as git and the SDK move. This also retires the `UNVERIFIED BY ANY GATE` marking on the command strings rather than merely disclosing it.

### Validation

`EnableSourceLink=false` A/B, run after the change, shows the trimmed hint on the one live path and nowhere else:

```
member -S 'Source Locations' produced no data after being listed by -D. NOTE: this assertion
needs SourceLink in the test assembly's PDB, and its PDB carries no SourceLink blob. That
implicates the build environment rather than a product regression, though this check does not
establish which cause applies. The known one under a .NET 11 SDK is issue #3658: in a linked
worktree, `Microsoft.Build.Tasks.Git` looks for `packed-refs` and the reftable stack in
`$GIT_DIR` while both live in `$GIT_COMMON_DIR`, resolves no revision, and the build emits no
SourceLink without warning. See .../issues/3658 to confirm whether that is the cause here and
to repair it. Other causes — a source archive with no `.git`, or SourceLink explicitly
disabled — need no repair.
```

That single failure is also the proof that removing the predicate clause changed no behavior. Build graph restored afterwards (`rm -rf artifacts/obj` + full rebuild); full CLI suite green before and after: **2846 tests, 0 failed, 14 platform skips.**

Head moved to `3086d5a5`. Round 6 is the last permitted without fresh approval, so I am not opening a seventh; @richlander, the changes since the reviewed head are two deletions — the embedded shell paragraph and a clause proven unreachable — plus the remarks explaining both.

richlander:
## Direction corrected: test the section contract, not the build environment

The prior diagnostic approach is superseded and deleted in `c063f4af`.

The key distinction is now explicit: the packed-ref defect is indistinguishable at the artifact boundary from SourceLink simply not being enabled. In both cases the assembly and portable PDB are valid, sequence points remain available, and only the SourceLink blob is absent. Reporting no data from a SourceLink-backed section is therefore correct product behavior.

The bug was in the test expectation. `Source Locations` declares `ProbeEffectiveness=false`, whose contract says `-D` lists it structurally and accepts that `-S` may render empty. `CliDiscoverySections_AreSelectable` correctly asserted selection success, then incorrectly required non-empty output for this command shape.

The replacement derives the non-empty exemptions from the two pipelines under test via `GetUnprobedSections()`. It keeps the real contract strict:

- every section listed by `-D` must be selectable by `-S`;
- probed sections in the guard scenarios must still render data;
- unprobed sections may validly render empty, exactly as their descriptor declares.

All SourceLink-specific probe, hint, repair commands, and diagnostic tests are gone. The net PR diff from `origin/main` is one test file: **19 insertions, 5 deletions**.

Current `origin/main` (`684f947c`) was merged as `4134687c`. On that integrated head:

- normal build: discovery test passes;
- `/p:EnableSourceLink=false`: `-D` lists `Source Locations`, `-S` exits 0 with zero output, discovery test passes;
- restored Release build with IL oracles: **2899 total, 0 failed, 4 platform skips**.

The earlier review findings remain useful documentation on #3658, but none of their shell-guidance surface remains in this PR.

richlander:
## Review round 7 — corrected design, clean

User-approved GPT-5.6 max-effort review of the exact green head `4134687c36577b109edd0d7ecb5017ab0f620f1b` found **no significant issues** and recommends **MERGE**.

The reviewer independently traced and exercised the contract:

- **Pipeline mapping is exact.** `type EmptyDiscoveryFixture` uses the type-detail pipeline; `member MemberCallsFixture Overloaded` uses the overload-inventory pipeline. None of the test's supplied options switch either path.
- **The exemption follows from the declaration.** `ProbeEffectiveness=false` explicitly permits structural discovery followed by an empty render; only the non-empty assertion is exempted.
- **Selectability remains unconditional.** A nonzero unprobed selection still fails. The library-only `SourceLink: Integrity` status result is unrelated and remains correctly scoped.
- **Both guards remain non-vacuous.** Empty type discovers `{Type Info}`, leaving probed residual `{Type Info}`. Multi-overload member discovers `{Methods, Member Index, Source Locations}`, leaving probed residual `{Methods, Member Index}` after exempting `Source Locations`.
- **Real disabled-SourceLink canary:** build succeeded; `-D` listed `Source Locations`; `-S` exited 0 with zero stdout; targeted discovery test passed.
- **Full suite:** 2899 total, 0 failed, 4 platform skips.
- **Review worktree:** clean at the exact head.

This closes the review loop on the replacement implementation; the prior six rounds applied to the deleted diagnostic approach.

richlander:
Ready to merge

